Issue #1969, opened 2019-08-03
`$file->changeSort()` doesn't have permission setting


texnixe:
**Describe the bug**
While `$page->changeNum()` requires authentication, `$file->changeSort()` doesn't and there also doesn't seem to be a permission setting for this. Is there are reason why?

**To Reproduce**
1. In the album template, add `$page->file()->changeSort(5);
2. Open the `animal` page or another sibling in a browser.
3. See how there is no error message and the file is updated.

**Expected behavior**
I'd argue that there should also be permission to prevent sorting on a per user basis and required authentication from the front-end.

Kirby 3.2.3-rc1

lukasbestle:
`$page->changeNum(5)` in the template doesn't require auth for me either... 🤔

texnixe:
Oh, ok, then I mixed that up, but nevertheless, the question is then why both actions do not require authentication when called from a template/script.

There seems to be a `sort` permission for pages (mentioned here: https://getkirby.com/docs/guide/users/permissions#role-based-permissions-in-user-blueprints) but at least according to the docs it doesn't have an equivalent in the page blueprint options or this option is just missing from the docs and should probably be called `changeNum` anyway.

I was just wondering if this is intended and if so, why.

lukasbestle:
Yes, it's a bit strange. I'd expect all methods to check for permissions.

bastianallgeier:
Hm, changeNum and changeSort should indeed be protected by permissions. We need to check this.

afbora:
I was thinking about this issue. A great and simple solution came to my mind 💡 

In the Kirby system, the permissions are managed by `templates`, not by page or file directly. So instead of looking at all the pages, if we check the templates in the collection by caching, the process is completed very quickly. So even if there are 1000 pages, if all of them belong to a single template, the process will be completed as a result of one check.

afbora:
As I was dealing with this again, I realized that `changeNum` and `changeStatus` are linked. If user have the `changeStatus` permission and haven't `changeNum` permission, when user try to update the page status as a `listed`, it will be updated as a `listed`, but the page `num` cannot be set 🤔

lukasbestle:
You are right – `changeStatus = true` implies (or should imply) `changeNum = true` as well. A user cannot really have the `changeStatus` permission without `changeNum`.

distantnative:
[...]

PR #2043, opened 2019-08-28
Fix $page->changeNum() and $file->changeSort() permissions

afbora:
## Describe the PR
Fixed the issue with adding new permissions that `changeNum` for pages and `changeSort` for files

## Related issues

- Fixes https://github.com/getkirby/kirby/issues/1969

## Todos

- [x] Pass all unit tests
- [ ] If needeed, in-code documentation (DocBlocks etc.)


bastianallgeier:
I remember again, why I didn't implement the permission check here. It's significantly slowing down sorting for pages and files. I know this is a bad reason, but we somehow need a way to batch process multiple pages and files without checking permissions over and over again for each individual one.

bastianallgeier:
There's also another issue with this. If a permission check fails while sorting, the entire sort order will be screwed. This is really tough.

bastianallgeier:
I'm closing this for now. We don't have a proper solution ready.

Issue #1935, opened 2019-07-21
More meaningful error message, if fields keyword is missing in a fields section


trych:
If I have a blueprint file containing a section that is missing the `fields` keyword like this:

```yaml
sections:
  content:
    type: fields
      title:
        type: text
      text:
        type: textarea
```

then I get the following error in a panel page that contains the erronous type of page:

```text
The section "projects" could not be loaded: Illegal offset type in isset or empty
```

This would require a more meaningful error message for a user to understand what actually causes the error.

Edit: [problem discussed in the forums](https://forum.getkirby.com/t/section-could-not-be-loaded/15014)

bastianallgeier:
✅

PR #2051, opened 2019-09-01
Checks section fields to prevent error

afbora:
## Describe the PR
If `fields` property not exists, panel was throw `PHP error` as `Illegal offset type in isset or empty`. 

Because when `fields` property does not exist, the `$type` variable returns an `array` instead of a `string` (returns `fields` as array). As you know, when `$type` is an `array`, an error occurs because it searching for an array in the array index.

So a new control was written to ensure that the `$type` variable is `string` and `not empty` before the if condition of the error line (This implement of 613. line)

Btw, with this fix, non string field types will not work.
I'm not sure if the integer is a field type like `type: 123`

## Related issues

- Fixes https://github.com/getkirby/kirby/issues/1935

## Todos
If you feel like you can help to check off the following tasks, that'd be great. If not, don't worry - we will take care of it.
- [x] Pass all unit tests
- [ ] If needeed, in-code documentation (DocBlocks etc.)


texnixe:
Hm, but now you get the same error message, no matter if a correct type is set or not. I think if the `fields` property is missing, it should say so, otherwise the error message is misleading.

afbora:
Not exactly i think, it was a PHP error, not a defect in the message. 
This fix checks if both fields are missing and the type is correct.

If fields are missing or not defined correctly, the message indicates which section has an error. 
This allows the user to check and correct the corresponding section.

afbora:
**Before**

FireShot Capture 003 - Kirby Panel - localhost

**After fix**

FireShot Capture 004 - Exploring the universe - Mægazine - localhost


Certainly clearer than before. But it could be a more descriptive sentence :))

>Invalid section type for ("SECTION NAME")

I think the important thing is to fix the PHP error.

bastianallgeier (approved):

bastianallgeier:
I think it's indeed a lot better!

Issue #2046, opened 2019-08-29
$field->toStructure() throws an exception when the field value is invalid


yandodov:
The error I'm getting is:

>Argument 1 passed to Kirby\Cms\StructureObject::setContent() must be of the type array or null, string given, called in C:\xampp\htdocs\ikarov\kirby\src\Toolkit\Properties.php on line 138

In the debugger, I saw this came from a `toStructure()` call:

![image](https://user-images.githubusercontent.com/5570098/63957759-b529e080-ca91-11e9-9204-b8189a50b64f.png)

This was very tricky to debug because `toStructure()` is called by a plugin. When you save the default translation, the plugin syncs the structure data with all other translations. Each entry had an `options` field of type `tags` which is a comma-separated string. When I changed the blueprint of that field to `structure`, this error began to appear because the saved value was not a YAML-encoded structure.

**To Reproduce**
1. Have a _site.txt_ with arbitrary string as content.
2. Run `$site->title()->toStructure()`

**Expected behavior**
Whenever `toStructure()` is called with invalid data, it should return an empty structure. Then, when the value is saved, whatever invalid value existed before would be replaced with a valid one.

**Kirby Version**
3.2.3

bastianallgeier:
✅

PR #2053, opened 2019-09-04
Fix invalid toStructure() data with validate

afbora:
## Describe the PR
Since the Kirby flat file CMS, the data in the txt file may sometimes be invalid. This can be a manual modification in the txt file or an incorrect update with the hook. So I think we need to be sure of the source of structured data.

This PR validating that each value in the data is an `array`.

Actually, we're doing this check elsewhere. Inspired from there, I carried it here.
https://github.com/getkirby/kirby/blob/master/config/fields/structure.php#L128-L141

I open it as a `draft` because it's just a suggestion (_especially about looping data_).
I'll update it `ready `if it's considered appropriate by you.

## Related issues

- Fixes https://github.com/getkirby/kirby/issues/2046

## Todos

- [x] Pass all unit tests

### DIFF
--- a/config/methods.php
+++ b/config/methods.php
@@ -216,14 +216,12 @@ return function (App $app) {
             $rows = Yaml::decode($field->value);
             $data = [];
 
-            if (is_array($rows) === true) {
-                foreach ($rows as $index => $row) {
-                    if (is_array($row) === false) {
-                        continue;
-                    }
-
-                    $data[] = $row;
+            foreach ($rows as $index => $row) {
+                if (is_array($row) === false) {
+                    continue;
                 }
+
+                $data[] = $row;
             }
 
             return new Structure($data, $field->parent());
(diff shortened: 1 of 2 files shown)


distantnative:
I haven't thought about the implementation yet, more a general thought before:

I am not sure that if I have invalid data, I want the system to just skip it instead of throwing an error. I can imagine people having no clue why their one structure field row is not appearing and searching for answers for hours, not realizing it could be that the data is invalid. They see, the data is there, so they search in other places (templates, controllers, models).

afbora:
I am convinced :))

Maybe a method can be written to check if the data is valid like that `isStructure()`. 
The user can write a custom method or we can add it to the core if necessary 👍

bastianallgeier:
I think the Structure class should throw a good exception with a proper message when it finds a value that's not an array. This is probably the most user-friendly way of pointing to the issue.

afbora:
Closing the PR to better implementation

distantnative:
@afbora Is the new PR already somewhere? Do you need help?

afbora:
@distantnative I haven't started the new PR yet.
I close this PR because there might be confusion as development will go through different class.

PR #2056, opened 2019-09-05
Fix missing file in api response and new F::relativepath() method (#1937)

### LINKED ISSUE #1937: API error response "file" property is getting lost
**Describe the bug**
The `file` specified in the JSON response is `""`:
image

**Expected behavior**
The problematic file should be specified in the error response.

**Screenshots**
image

**Kirby Version**
3.2.0

**Additional context**
As you can see in the debug console (screenshot), `$e->getFile()` correctly returns the file where the error occurred. However, after the `DOCUMENT_ROOT` check, `$file` is now `""`. Here's some evaluated code in the debug console after this check:

```
$file
""
$e->getFile()
"C:\xampp\htdocs\kirby\site\plugins\exporter\src\Formatter.php"
$_SERVER['DOCUMENT_ROOT']
"C:/xampp/htdocs"
```

bastianallgeier:
## Describe the PR
Makes sure to always include a proper relative file path in API responses. `F::relativepath()` is new to help with the task of creating reliable relative file paths. 

## Related issues
- Fixes #1937

## Todos
- [x] Pass all unit tests

### DIFF
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -540,50 +540,19 @@ class Api
         try {
             $result = $this->call($path, $method, $requestData);
         } catch (Throwable $e) {
-            if (is_a($e, 'Kirby\Exception\Exception') === true) {
-                $result = [
-                    'status' => 'error',
-                    'route'  => ($this->route)? $this->route->pattern() : null
-                ] + $e->toArray();
-            } else {
-                // remove the document root from the file path
-                $file = $e->getFile();
-                if (empty($_SERVER['DOCUMENT_ROOT']) === false) {
-                    $file = ltrim(Str::after($file, $_SERVER['DOCUMENT_ROOT']), '/');
-                }
-
-                $result = [
-                    'status'    => 'error',
-                    'exception' => get_class($e),
-                    'message'   => $e->getMessage(),
-                    'file'      => $file,
-                    'line'      => $e->getLine(),
-                    'code'      => empty($e->getCode()) === false ? $e->getCode() : 500,
-                    'route'     => $this->route ? $this->route->pattern() : null
-                ];
-            }
+            $result = $this->responseForException($e);
         }
 
-        if ($result === null) {
-            $result = [
-                'status'  => 'error',
-                'message' => 'not found',
-                'code'    => 404,
-            ];
-        }
-
-        if ($result === true) {
-            $result = [
-                'status' => 'ok',
-            ];
-        }
-
-        if ($result === false) {
-            $result = [
-                'status'  => 'error',
-                'message' => 'bad request',
-                'code'    => 400,
-            ];
+        switch (true) {
+            case $result === null:
+                $result = $this->responseFor404();
+                break;
+            case $result === false:
+                $result = $this->responseFor400();
+                break;
+            case $result === true:
+                $result = $this->responseFor200();
+                break;
         }
 
         if (is_array($result) === false) {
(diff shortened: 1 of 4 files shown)


src/Toolkit/F.php
```diff
@@ -573,6 +573,31 @@ public static function realpath(string $file, string $in = null)
...
 
+    /**
+     * Returns the relative path of the file
+     * starting after $in
+     *
+     * @param string $file
+     * @param string $in
+     * @return string
+     */
+    public static function relativepath(string $file, string $in = null): string
```
lukasbestle, line 584: I think the second param shouldn't be optional as the whole method doesn't make any sense without it (it will always return the first param). Or is there any use-case I've missed?

src/Toolkit/F.php
```diff
@@ -573,6 +573,31 @@ public static function realpath(string $file, string $in = null)
...
+        if (empty($in) === true) {
+            return $file;
+        }
+
+        // windows
+        $file = str_replace('\\', '/', $file);
+        $in   = str_replace('\\', '/', $in);
+
+        if (Str::contains($file, $in) === false) {
+            return $file;
```
lukasbestle: I'm not sure about this fallback. Maybe the method should instead return `null` in this case. Otherwise there's the issue of leaking the full file path where a relative path was expected. That could be a security issue depending on how the method is used.

src/Api/Api.php
```diff
@@ -540,50 +540,19 @@ public function render(string $path, $method = 'GET', array $requestData = [])
...
-            ];
-        }
-
-        if ($result === false) {
-            $result = [
-                'status'  => 'error',
-                'message' => 'bad request',
-                'code'    => 400,
-            ];
+        switch (true) {
```
lukasbestle: That `switch` looks a bit weird. Why not `if elseif elseif`?

lukasbestle (changes requested):

src/Toolkit/F.php
bastianallgeier: When the second parameter is required it is also required to be a string. You can't pass null, which is sometimes really handy. I.e. in this case:

```php
F::relativepath('/something/something/something.php', Server::get('DOCUMENT_ROOT'));
```

src/Toolkit/F.php
bastianallgeier: In our case, we really need some valuable alternative. Maybe we could return just the filename in this case? I'm not sure. Both solutions seem to be based on assumptions.

src/Api/Api.php
bastianallgeier: I found this to be quite elegant, because it can be extended so easily. To me switches are always easier to read than endless if elseif clauses, but that might be personal.

bastianallgeier: Unfortunately switches don't do strict comparisons, otherwise it would be even cleaner.

src/Toolkit/F.php
lukasbestle: I see. :)

src/Api/Api.php
lukasbestle: As you say, it's pretty subjective. To me the special `switch` looks like a hack. A working one, but still a hack. `if elseif elseif` would be a bit less code overall and harder to screw up (you can't forget a `break`). But you can leave it like this if you want.

src/Toolkit/F.php
lukasbestle: I think just the filename would be a good alternative. Better than nothing. :)

src/Api/Api.php
bastianallgeier: I gave in and replaced it :)

lukasbestle (approved):

Issue #1781, opened 2019-05-16
[Time Field] Selecting First Hour/Minute Always Selects 12:00 AM


neildaniels:
**Describe the bug**
When setting the initial value of a `time` Blueprint Field, it always ends up showing `12:00 AM`.

**To Reproduce**
1. Have a `time` field in a Blueprint
```yaml
      scheduledTime:
        label: Scheduled Time
        type: time
        notation: 12
        interval: 30
        width: 1/3
        help: scheduledTime
```
2. Attempt to select something like the hour `06`

**Expected behavior**
The hour `06` should be displayed.

**Screenshots**
<img width="337" alt="Screen Shot 2019-05-16 at 3 33 22 PM" src="https://user-images.githubusercontent.com/417182/57891400-18984700-77f0-11e9-9df4-a4dd0b0c0611.png">
<img width="318" alt="Screen Shot 2019-05-16 at 3 33 26 PM" src="https://user-images.githubusercontent.com/417182/57891406-1c2bce00-77f0-11e9-9755-39bbca41d752.png">

**Kirby Version**
`3.1.3` and `3.1.4-rc.1`

**Console output**
If you are working with the panel, please check the browser console for errors and send the output.

**Desktop (please complete the following information):**
 - OS: macOS 10.14.5 (18F132)
 - Safari 12.1.1 (14607.2.6.1.1)
 - Chrome 74.0.3729.157

bastianallgeier:
✅

PR #2058, opened 2019-09-07
Fix time field notation

afbora:
## Describe the PR
    
Following method return `NULL` as `STRING` type:
https://github.com/getkirby/kirby/blob/master/panel/src/components/Forms/Input/TimeInput.vue#L139

    String(this.meridiem).toUpperCase()

So it is not `null` and `undefined` type, it is detected as an `invalid date` (like that `03:00 NULL`) and returns 12 by default on first selection.
`null` control of `this.meridiem` variable is required.

Note that: If the variable `this.meridiem` is always uppercase, PR should be like that:

    const a = this.meridiem || "AM";

Tested on 3.2.4

## Related issues

- Fixes https://github.com/getkirby/kirby/issues/1781

## Todos

- [x] Pass all unit tests


distantnative:
Would `String(this.meridiem || "AM").toUpperCase()` work as well?

afbora:
Yes, working perfect! 👍

bastianallgeier (approved):

PR #2059, opened 2019-09-07
Only run APCu tests when available

distantnative:
Let me end my frustrations 😂


tests/Cache/ApcuCacheTest.php
```diff
@@ -9,6 +9,14 @@
  */
 class ApcuCacheTest extends TestCase
 {
+    public function setUp(): void
+    {
+        if (extension_loaded('apc') === false || (bool)ini_get('apc.enabled') === false) {
```
lukasbestle: The tests are currently even skipped on Travis. This should probably be `extension_loaded('apcu')`. The `ini_get()` check is fine though as those options are still called `apc.*`.

lukasbestle (changes requested):

tests/Cache/ApcuCacheTest.php
bastianallgeier: I'd check for apuc_store to find out if APCU works properly.

bastianallgeier (approved):

Issue #1445, opened 2019-02-04
[Panel] Range field does not show default value on load


fabianmichael:
**Describe the bug**
When a range field is added to a blueprint, the slider element reflects its current value correctly, but the right-sided label only displays the value after the slider knob has been moved.

**To Reproduce**
1. Add a range slider to any existing blueprint. Here are the settings, that I used:
```
      perPage:
        label: Einträge pro Seite
        type: range
        step: 3
        min: 3
        max: 30
        default: 12
        required: true
        translate: false
```
2. Open the corresponding page in the panel. The slider has a red border, saying that the field is invalid and the default value right-handed to the slider does not appear.

**Expected behavior**
If a default value is set, the slider should behave like that number has been set automatically and not complain about an invalid field. I am aware, that this would mean that a permanent save bar will be displayed after a blueprint has changed, but I find it rather less intuitive if there is a default value, that requires user interaction for becoming saveable.

**Screenshots**
![bildschirmfoto 2019-02-04 um 13 00 50](https://user-images.githubusercontent.com/395617/52207654-59da9b80-287e-11e9-90e7-e9a36234f312.png)

**Kirby Version**
3.0.0

**Console output**
-

**Desktop (please complete the following information):**
 - OS: macOS 10.14.2
 - Browser: Firefox
 - Version: 65.0

fabianmichael:
I used the range slider in another project and found some more related glitches:

- If used inside of a structure field, default values work correctly, as long as the `default` parameter is not `0`.
- If the default value is `0`, the range field will display a dash `-` as value when used outside of a structure field. When used inside a structure field, the default value is not set and the right-sided label displays `NaN`.

bastianallgeier:
✅

musterknabe:
I still have the same behavior in Kirby 3.8.3:

```
BackgroundGradientAngle:
  label: Angle
  type: range
  default: 90
  step: 1
  min: 0
  max: 359
  tooltip:
    after: degree
```
<img width="500" alt="Screenshot 2022-12-14 at 13 29 47" src="https://user-images.githubusercontent.com/362391/207595786-f4c02bb2-131a-4967-b870-8575cefd8b91.png">

distantnative:
@musterknabe did you add the range field/its default in the blueprint after the page got created? Defaults are only applied during the creation of a page.

musterknabe:
Yes I did. I was building my blueprint, while reloading the already created page.

[...]

PR #2066, opened 2019-09-10
Fix range field default value

afbora:
## Describe the PR

@distantnative Need your review for this simple PR :))

**Fixed Preview Field**

```
fields:
  perPage:
    label: Einträge pro Seite
    type: range
    step: 3
    min: 3
    max: 30
    default: 12
    required: true
    translate: false
```

FireShot Capture 746 - Test - Mægazine - http___localhost_test_kirby_1445_panel_pages_notes+test

## Related issues
- Fixes https://github.com/getkirby/kirby/issues/1445

## Todos

- [ ] Pass all unit tests


panel/src/components/Forms/Input/RangeInput.vue
```diff
@@ -35,6 +35,7 @@ export default {
     autofocus: Boolean,
     disabled: Boolean,
     id: [String, Number],
+    default: [Number, String],
```
afbora, line 38: `this.default` was returning `undefined` without this prop.

panel/src/components/Forms/Input/RangeInput.vue
```diff
@@ -70,14 +71,14 @@ export default {
   },
   computed: {
     label() {
-      return this.value !== null ? this.format(this.value) : "–";
+      return this.required || this.value ? this.format(this.position) : "–";
```
afbora, line 74: Only if required and value is not empty, the position is written to the label.

panel/src/components/Forms/Input/RangeInput.vue
```diff
@@ -70,14 +71,14 @@ export default {
...
     },
     center() {
       const middle = (this.max - this.min) / 2 + this.min;
       return Math.ceil(middle / this.step) * this.step;
     },
     position() {
-      return this.value !== null ? this.value : this.center;
+      return this.value || this.default || this.center;
```
afbora, line 81: It always shows the correct position, whether it is required field or not.

distantnative:
It looks good to me but I haven't been able to test it myself.

bastianallgeier:
I made a few small changes. The computed position value should be used in all places. Then it works as expected.

plan44:
Sorry for posting here but it seems closely related:
For a range with min=0 and max=5, when the stored value is 0, it does not display the value zero (neither on the slider nor in the label) but the default value.

It seems to me that this is a JS zero/null confusion caused by the code changes above.

The original line 80 in RangeInput.vue explicitly compared with null, where the change now only uses `this.value` which is falseish also when it contains the number zero.

I tried to test/fix it in my kirby installation but apparently the vue sources are not directly used but need some build step I have no idea how to perform.

afbora:
Thank you @plan44, i have fixed the issue on #2549

PR #2073, opened 2019-09-11
Better range input baseline #2044

### LINKED ISSUE #2044: Range Field Has Unexpected Display with Missing Value
**Describe the bug**
In the Panel, the `range` field surprised me with the way it handles having no value set (like if creating a new page without a default value, or adding the field to an existing page).

For example,
```yaml
ratingValue:
  label: Rating Value
  type: range
  min: 0
  max: 5
  tooltip:
    after: '★'
  width: 1/3
```

Looks like this when there's no saved value:
Screen Shot 2019-08-28 at 11 19 59 AM

**Expected behavior**
If `0` is part of the possible range of values, I would expect the range slider to _appear_ to be positioned where 0 would be (but still display the `-` dash).

My rationale: I assume most uses of a range value involve `->toInt()` or `->toFloat()` and an empty value would mean it's returning `0` or `0.0`. In the example above, it looks like the default value is "3".

However, if `0` isn't within the allowed range, then I suppose sticking the slider in the exact middle is as valid as any other place.

Regardless of placement: maybe the knob itself should be semi-transparent to indicate that it's not "really" set to the value at that spot.

**Kirby Version**
`3.2.3`

 - OS: macOS 10.15 Beta (19A536g)
 - Safari 13.0 (15608.2.2)

distantnative:
## Describe the PR
Trying to be less clever with the range input baseline (the position where the knob is displayed when no value is yet present):

- If the minimum is below 0, the baseline should be placed at 0.
- Otherwise place the baseline at the minimum

## Related issues
- Fixes #2044

### DIFF
--- a/panel/src/components/Forms/Input/RangeInput.vue
+++ b/panel/src/components/Forms/Input/RangeInput.vue
@@ -70,15 +70,16 @@ export default {
     };
   },
   computed: {
+    baseline() {
+      // If the minimum is below 0, the baseline should be placed at 0.
+      // Otherwise place the baseline at the minimum
+      return this.min < 0 ? 0 : this.min;  
+    },
     label() {
       return this.required || this.value ? this.format(this.position) : "–";
     },
-    center() {
-      const middle = (this.max - this.min) / 2 + this.min;
-      return Math.ceil(middle / this.step) * this.step;
-    },
     position() {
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):
LGTM

PR #2074, opened 2019-09-11
Fix structure preview for files/images

### LINKED ISSUE #2039: Image Preview in Structure Field uses full image size
When using images in a structure the preview uses the original file instead of the resized image (eg. used in the files list). So creating structures with a some large images (eg. 10mb, 4000x6000 pixels each) will make the panel very slow.

Maybe I haven’t found the right option but event settings the `image` property with the option ` size: tiny` for this column the panel still uses the large image.

### LINKED ISSUE #2040: Video Thumbnail in Structure field actually plays video
**Describe the bug**
When having a `files` field within a `structure` field, and then select a (mp4) video, the video will actually play inside the field preview of the closed structure field. This slows down the performance of the panel.

**To Reproduce**
1. Create a `structure` field like this
```yml
structureField:
  type: structure
  label: Structure
  fields:

    filesField:
      type: files
      label: Files
      image: false
```
2. Upload a mp4 video file and select it from `filesField`
3. Save the page

**Expected behavior**
Instead, the `file-video` icon should be shown, I guess. Or a thumbnail can be queried using the fields `image` property.

**Screenshots**
Icon is shown when editing the structure entry:
icon is shown when editing the structure entry

Video playback inside the thumbnail, when structure field is closed after editing:
Video playback inside the thumbnail

**Kirby Version**
3.2.3

distantnative:
## Describe the PR
- Show icons for non-images (prevent videos playing)
- Use srcset for images
- Move code to previewThumb helper to DRY

## Related issues
- Fixes #2039
- Fixes #2040

### DIFF
--- a/panel/src/components/Forms/Previews/FilesFieldPreview.vue
+++ b/panel/src/components/Forms/Previews/FilesFieldPreview.vue
@@ -2,7 +2,7 @@
   <ul v-if="value" class="k-files-field-preview">
     <li v-for="file in value" :key="file.url">
       <k-link :title="file.filename" :to="file.link" @click.native.stop>
-        <k-image v-if="file.type === 'image'" :src="file.url" v-bind="image" />
+        <k-image v-if="file.type === 'image'" v-bind="imageOptions(file)" />
         <k-icon v-else v-bind="file.icon" />
       </k-link>
     </li>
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):
LGTM

Issue #2076, opened 2019-09-12
Unable to return custom response in the API


yandodov:
**Describe the bug**
I want to return an empty response along with a message, but I can't find a way. If I `return` a new `Response` object, I get an error.

**To Reproduce**

1. Add the following route:

```php
[
    'pattern' => 'data',
    'method' => 'GET',
    'action' => function () {
        return new Response('empty body', null, 204);
    }
]
```

2. Open this route in the browser
3. Error:

```json
{
    "status": "error",
    "route": "export",
    "exception": "Kirby\\Exception\\NotFoundException",
    "message": "The object \"response\" cannot be resolved",
    "key": "error.notFound",
    "file": "C:\\xampp\\htdocs\\kirby\\kirby\\src\\Api\\Api.php",
    "line": 410,
    "details": [],
    "code": 404
}
```

**Expected behavior**
There should be a way to easily return all HTTP response codes from a custom route.

**Kirby Version**
3.2.3

**Additional context**
I read [this forum post](https://forum.getkirby.com/t/return-error-response-from-custom-api-endpoint/14538) and it doesn't work for me. If I `echo` the `Response`, I get:

```json
{
    "status": "error",
    "message": "not found",
    "code": 404
}
```

Because the return value form the route is `null`. If I return `true`, I get status OK and response code `200`, i.e. the echoed `Response` is ignored.

I think two things should change:

1. You should be able to `return new Response()` in the route and have that work correctly
2. The `Response` object should be able to handle a `message` parameter so you can easily send a message along with the status code like this:

```php
return new Response([
    'code' => 204,
    'message' => 'That object is empty.'
]);
```

...instead of:

```php
return new Response([
    'code' => 204,
    'body' => json_encode([
        'status' => 'ok',
        'message' => 'That object is empty.',
        'code' => 204
    ])
]);
```

lukasbestle:
I agree on the first part (support for returning response objects from API routes), that would probably be rather simple to add. It's a new feature however and not a bug.

I'm not so sure about the second part though. Response objects are pretty raw data objects. You can throw any kind of response structure at them, so I think that it shouldn't care whether there is a message key in a JSON array etc. If you wanted a JSON array, you would return it directly from the route. If you want full flexibility, you can return a completely custom response with the first feature you proposed.

yandodov:
>If you wanted a JSON array, you would return it directly from the route.

Yes, but you wouldn't be able to set the status code? I can do this:

```php
[...]

PR #2080, opened 2019-09-13
Support for returning Response obj from API route

lukasbestle:
## Describe the PR

It is now possible to return a completely custom `Response` object from custom API routes. This enables custom HTTP status codes, headers and response bodies.

## Related issues

Implements #2076.

## Todos

- [x] Pass all unit tests

### DIFF
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -174,7 +174,7 @@ class Api
 
         $output = $this->route->action()->call($this, ...$this->route->arguments());
 
-        if (is_object($output) === true) {
+        if (is_object($output) === true && is_a($output, 'Kirby\\Http\\Response') !== true) {
             return $this->resolve($output)->toResponse();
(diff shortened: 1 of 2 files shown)


bastianallgeier (changes requested):
I think this will clash in the Api::render() method. We don't handle Response objects there.

distantnative:
@bastianallgeier Would have thought that `Api::render` does with these lines: https://github.com/getkirby/kirby/blob/0e682d5045e91ddbea109d0fb1a64416c566d973/src/Api/Api.php#L554-L556

lukasbestle:
Yep, I have tested it and it works perfectly. :)

bastianallgeier:
@distantnative ha, I completely missed that line. Of course it should work. Awesome!

bastianallgeier (approved):

PR #2081, opened 2019-09-13
Non-admin user cannot create a new user

### LINKED ISSUE #2038: Non-admin user cannot create any other users anymore if there is no other role but admin role
**Describe the bug**
When creating a new Editor role (but no other roles) without any settings, this user cannot create new users.

In this case the role radio does not show up and trying to create a user fails. I assume that the default role in this case falls back to admin, making it impossible to create new users. When I add a second role, the role selector reappears and creating users with any of those roles becomes possible again.

**To Reproduce**
1. Create a new Editor role without any settings
2. Create a new user with this role.
3. Log in to the Panel with this user
4. Try to create a new user
5. See error message ("You are not allowed to create this user").

**Expected behavior**
While it makes sense that the role selector is not shown when there is only one non-admin role, the role should be set to this non-admin role by default and the user created.

Bildschirmfoto 2019-08-25 um 21 28 45

**Kirby Version**
Tested with 3.2.3 and 3.2.4-rc1

**Console output**

Bildschirmfoto 2019-08-25 um 21 29 10

**Additional context**

bastianallgeier:
## Describe the PR
A non-admin user was not able to create a new user of the same role, if just one additional role to admin existed. This is now fixed. 

## Related issues
- Fixes #2038


distantnative (dismissed):

distantnative:
LGTM

PR #2082, opened 2019-09-13
Added $pages->notTemplate() method

distantnative:
By @neildaniels (original PR got automatically closed when we decided to remove the `features` branch and work only with `develop`)

## Describe the PR
Adds `$pages->notTemplate()` to exclude specific templates from a Pages collection.

## Related issues
- Implements https://github.com/getkirby/ideas/issues/392

## Todos
- [x] Pass all unit tests
- [x] If needeed, in-code documentation (DocBlocks etc.)

### DIFF
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -429,6 +429,27 @@ class Pages extends Collection
     }
 
     /**
+     * Filter all pages by excluding the given template
+     *
+     * @param string|array $templates
+     * @return Kirby\Cms\Pages
+     */
+    public function notTemplate($templates)
+    {
+        if (empty($templates) === true) {
+            return $this;
+        }
+
+        if (is_array($templates) === false) {
+            $templates = [$templates];
+        }
+
+        return $this->filter(function ($page) use ($templates) {
+            return !in_array($page->intendedTemplate()->name(), $templates);
+        });
+    }
+
+    /**
      * Returns an array with all page numbers
      *
      * @return array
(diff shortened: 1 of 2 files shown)


lukasbestle (approved):

bastianallgeier:
@lukasbestle @distantnative @neildaniels what do you think about this instead: 

```php
$pages->template('!project')
```

That's the syntax I want to introduce for blueprints as well and I think it would be good to keep this consistent. 

```yaml
sections: 
  pages: 
    type: pages
    template: 
      - !projects
      - !blog
```

neildaniels:
I’m happy with that syntax, especially for Blueprints.

I do think it’s slightly quirky that there wouldn’t be a `->notTemplate()` companion method, but I think the other `->notX()` don’t take any inputs, so maybe that’s distinct enough to not warrant the `->notTemplate()` version.

lukasbestle:
I'm not sure. On one hand I like it because it makes the syntax consistent, on the other hand I think that it's too much added complexity for this simple method. It suddenly needs to get "smart".

neildaniels:
I was about to say "one benefit of this is that you can combine something like `->template('foo')->notTemplate('bar')` into a single `->template(['foo', '!bar'])` call and it shouldn't have to loop though twice."

But then I realized, there should literally never be a place where you actually want to mix a whitelist and blacklist. And that makes me thing this syntax is actually unnecessary.

The moment you decide to use the `->template()` method, you necessarily will be deciding if _everything_ in the list is a whitelist or blacklist. So in that sense, I think `->template()` and `->notTemplate()` probably make more sense.

lukasbestle:
Exactly. The only case where mixing would make sense is if it allowed wildcards. But otherwise you can only use one or the other.

bastianallgeier:
@neildaniels ok, that last comment convinced me :) I guess we still need the syntax for the blueprints then. Or we decide to use a notTemplate option there as well.

PR #2083, opened 2019-09-13
Editor preview

distantnative:
**Requires first:** Content Vuex store refactoring

-----

Screenshot 2019-05-19 at 13 23 26

My main trouble at the moment is the placement of the iframe (`k-page-preview`). It seems to make sense to place it in the PageView. However, it's a child of `k-panel-view` then, which applies `overflow-y: scroll;`. With that, the iframe cannot be placed fixed to make it scrollable independent from the main Panel content.

- [ ] @bastianallgeier: see if there is a better way than my hackish CSS implementation
- [ ] How to deal with CSS media queries (breakpoints) when preview is shown
- choose a more cryptic query variable name

### DIFF
--- a/panel/src/App.vue
+++ b/panel/src/App.vue
@@ -1,6 +1,7 @@
 <template>
+  <k-page-preview v-if="$store.state.preview.active" />
   <div
-    v-if="!$store.state.system.info.isBroken"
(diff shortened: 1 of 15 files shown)


distantnative:
Will be reopened once we get there.

PR #2084, opened 2019-09-13
Don't save untranslatable fields

### LINKED ISSUE #2001: Bug with initial translated content
**Describe the bug**
Under certain conditions translated content files contain not translatable fields.

**To Reproduce**
1. Install the startkit
2. Add a second language
3. Make some fields not translatable (e.g. the `date` and `author` field in the `note` blueprint)
4. Navigate to a note in the panel and switch to the secondary language
5. Edit the `title` **or** change the `slug`
6. The created content text file will include both the `date` and `author` field

As far as I can tell this bug only occurs if you just change the `title` or `slug` of a not yet created translation. As soon as you save any "real" fields the content file gets sanatized as expected.

**Expected behavior**
The `date` and `author` field should not end up in the translated content file.

**Kirby Version**
3.2.3

 - Windows 10

bastianallgeier:
## Describe the PR
The ModelWithContent::saveTranslation method didn't check for untranslatable fields. This led to unwanted fields being stored when using Page::changeSlug and Page::changeTitle. The method now unsets untranslatable fields correctly. 

## Related issues
- Fixes #2001

## Todos
- [x] Pass all unit tests


src/Cms/ModelWithContent.php
```diff
@@ -492,8 +492,18 @@ protected function saveTranslation(array $data = null, string $languageCode = nu
             throw new InvalidArgumentException('Invalid language: ' . $languageCode);
         }
 
+        // get the content to store
+        $content = $translation->update($data, $overwrite)->content();
```
lukasbestle, line 496: Is it needed to run the loop over all fields (even the ones that won't be modified by the current save operation)? It means that existing fields that have become untranslatable in the meantime will be kicked out. It's probably useful to do that, but we should be sure that this is what we want. Besides that the changes LGTM.

bastianallgeier: I think it's better to do it that way and not to be too smart about the fields, which should be skipped or not.

PR #2087, opened 2019-09-13
Update user store

### LINKED ISSUE #1957: Updating user language in Panel: store doesn't get updated
After changing my account language in the Panel, `$store.state.user.current.language` remains unchanged (only after reloading browser).

bastianallgeier:
## Describe the PR
The current user store was not updated correctly when the email, language, role or name of the current user has been changed. 

## Related issues
- Fixes #1957


panel/src/store/modules/user.js
```diff
@@ -28,6 +28,12 @@ export default {
     current(context, user) {
       context.commit("SET_CURRENT", user);
     },
+    email(context, email) {
+      context.commit("SET_CURRENT", {
+        email: email,
+        ...context.state.current
```
distantnative: Just reading, not testing... but don't these two lines need to be switched? I'd assume there is a `email` key in `context.state.current` as well, so it would overwrite line 33 directly. Or?

bastianallgeier: You are totally right. I just fixed it.

distantnative (approved):

PR #2088, opened 2019-09-13
Calendar: use “current month” #1736

### LINKED ISSUE #1736: [Panel] Date Field "Today" Button Doesn't Actually Set Date/Time To Now
**Describe the bug**
When clicking on the "Today" button in a `date` field calendar, I fully expected the value in the field to be changed to today (or now, if using a time field as well).

Is this intentional? I can't really seem to think of a compelling use case for wanting this behavior.

**To Reproduce**
1. Open a date picker on a panel page
2. Select some date a couple months from now
3. Open up the calendar picker and click "Today"

**Expected behavior**
I would expect the calendar view and value to be set to today. Instead, it currently just changes the calendar view to focus on this month.

**Screenshots**
Screen Shot 2019-05-02 at 12 47 43 PM

**Kirby Version**
`3.1.3`

 - OS: macOS 10.14.4 (18E226)
 - Chrome 74.0.3729.131
 - Safari 12.1 (14607.1.40.1.4)

distantnative:
## Describe the PR
As we had quite a few discussion about the behavior of the "Today" button in the calendar dropdown - whether it should navigate to the current month as it does now, or rather directly update the value to the current date, I think the straightforward solution is to rename the button to "current month" and stick with the current behavior.

## Related issues
- Fixes #1736


i18n/translations/en.json
```diff
@@ -391,7 +392,6 @@
   "sort": "Sort",
   "title": "Title",
   "template": "Template",
-  "today": "Today",
```
lukasbestle: Have you checked if we use that string anywhere else? It could potentially become useful in the future as well, so I'm not sure if we should remove it.

panel/src/components/Forms/Calendar.vue
```diff
@@ -40,7 +40,7 @@
       <tfoot>
         <tr>
           <td class="k-calendar-today" colspan="7">
-            <k-button @click="go('today')">{{ $t("today") }}</k-button>
+            <k-button @click="go('today')">{{ $t("field.date.month.current") }}</k-button>
```
lukasbestle: I think we should rename the special `today` date in the `go()` method to `current-month` as well to be consistent.

i18n/translations/en.json
distantnative: We don't use it elsewhere. But you are right, let's not throw away the translations.

lukasbestle (dismissed):

lukasbestle (approved):

bastianallgeier:
This feels like a compromise to me. I ran into the same weird feeling that the Today button should directly set the date when clicking the other day and I suggest to give it a try like that as a real alternative to what we have. I just added a commit for it to the develop branch and we can give it try during the RC.

afbora:
Will translations be updated?

bastianallgeier:
No, because "Today" as label still works.

PR #2090, opened 2019-09-14
Use absolute path for dragText with parent #2003

### LINKED ISSUE #2003: Image tag doesn't include path to file when uploading to a different page in textarea
**Describe the bug**
Consider a textarea definition like this:

```
text:
  type: textarea
  uploads:
    parent: page.parent
```

The files are uploaded to the parent page correctly, but the resulting image tag only contains the filename, not the needed path to the file, resulting in wrong image URLs.

**To Reproduce**
1. Add the above field to a subpage blueprint, e.g. `album.yml` in the Starterkit.
2. Open a page with this blueprint in the Panel
3. In the textarea field, upload an image.
4. See how the image tag only contains the filename.

The same happens when selecting a file from another page, while the path is inserted correctly when dragging & dropping a file from a files section.

**Expected behavior**
The image tag should include the path to the file.

**Kirby Version**
Tested with 3.2.3

distantnative:
## Describe the PR
When using the `parent` option for uploads in textareas, the inserted `dragText` did not include an absolute path. It now does.

```
text:
  type: textarea
  uploads:
    parent: page.parent
```

It also moves the kirbytext/markdown check inside of the `dragText` method to keep code DRY.

## Related issues
- Fixes #2003

## Todos
- [x] Pass all unit tests


bastianallgeier (commented):
I love the general refactoring, but by removing the first argument of the dragText method, this becomes a breaking change that could affect plugins and other stuff.

distantnative:
It's an internal method with little documentation... I would argue keeping the code DRY and less error prone outweights the risk. But of course we can keep it and move the option check into all 3 fields/sections.

bastianallgeier:
I think I have a good solution that makes the best out of your refactoring without introducing a massive breaking change. I've brought back the first arg, but with a new 'auto' option that loads the type from config. This also keeps the advantage of being able to use the dragText method independently from what is set in the config.

bastianallgeier (approved):

bastianallgeier (approved):

distantnative:
Great solution!

PR #2091, opened 2019-09-14
Allow changeRole for non-admin users #1919

### LINKED ISSUE #1919: [Panel] A non-admin user cannot change the role of another user of the same or different role
**Describe the bug**
If I create for example an Editor user (with no permission settings set or with explicit settings to allow changing roles), I get an error message when trying to change the role of another user of the same or a different (non-admin) role. But the option is nevertheless available via the context menu.

> You are not allowed to change the role for the user "test@user.com"
```
title: Editor

options:
  changeRole:
    editor: true
    manager: true
permissions:
  users:
    changeRole: true
```

**To Reproduce**
1. Create an Editor blueprint like the above or empty
2. Create other users with the same or a different non-admin role
3. Try to change the role of these users via the context menu or in the user view. The change role dialog pops up (even with the option to change to the admin role which should not be the case, but that is another issue).
4. See error message

**Expected behavior**
A non-admin user role should still be able to change other users roles (apart from creating admins) if it is allowed via the role blueprint (assuming if there are no settings that everything is allowed).

**Screenshots**

**Kirby Version**
3.2.2

**Additional context**
I think the user permissions are a bit messed up in general atm.

https://github.com/getkirby/kirby/issues/1917

distantnative:
## Describe the PR
Line 73 prevents any non-admin user to change the role of any other user so far - no matter if the user has permissions or not (which we check later on line 89).

Instead this PR changes the check to make sure that only admin users can change the role of another admin user. That check ensures that no non-admin user is taking admin rights away from an admin.

## Related issues
- Fixes #1919


bastianallgeier (commented):
The biggest issue here is the situation when someone is able to create a new user and then turn that user into an admin. Afterwards they can login with that new account and be an admin.

distantnative:
That's why we need https://github.com/getkirby/kirby/pull/2092 too

distantnative:
I think the combination of these two PRs makes more sense than the current implementation.

PR #2099, opened 2019-09-14
Section parent: only allow objects #1938

### LINKED ISSUE #1938: Error needed when parent in files section is a collection
I think it is great, that the `files` section can display not only files from one single `page`, but also all the files from multiple `pages`. Like:
```
all_uploads:
  type: files
  headline: All uploads
  parent: site.index
```

The problem appears when uploading a file to a section like that: `No route found for path: "files" and request method: "POST"`, probably because it doesn’t know where to upload the new `file`.

A solution could be to:
1. add a `upload` option, that can be set to `false` to disable the upload functionality to this section. Or:
2. add a `upload` option where a single(!) `page` can be specified, to where a newly uploaded file should be moved to

distantnative:
## Describe the PR
Do not allow collections as parent of a files or pages section.

## Related issues
- Fixes #1938


lukasbestle (dismissed):

config/sections/mixins/parent.php
```diff
@@ -22,6 +20,14 @@
                 if (!$parent) {
                     throw new Exception('The parent for the query "' . $query . '" cannot be found in the section "' . $this->name() . '"');
                 }
+
+                if (
+                    is_a($parent, 'Kirby\Cms\Page') === false &&
+                    is_a($parent, 'Kirby\Cms\Site') === false &&
+                    is_a($parent, 'Kirby\Cms\User') === false
```
bastianallgeier, line 28: I think we need to include File here as well. You can add a files section to a file blueprint for example and the files section will automatically use the siblings.

bastianallgeier (changes requested):

config/sections/mixins/parent.php
distantnative: But would the File be the object in the parent prop then?

bastianallgeier: You are right, it wouldn't. But it could still be useful I guess. Would it do any harm to check for the file here as well?

distantnative: Probably not... then I would check for ModelWithContent?

bastianallgeier (approved):

PR #2100, opened 2019-09-14
Add $pagination->hasCurrentPage() method #1887

### LINKED ISSUE #1887: Pagination doesn't redirect to error page, shows last page instead.
**Describe the bug**
The current pagination, instead of returning an error page when trying to visit a page number that doesn't exist (for example visiting `page:50` in a collection that has only 10 pages) it's returning the last page of the pagination.

**To Reproduce**
1. In the starterkit homepage template, add a `paginate()` in the foreach loop. For example `foreach (page('photography')->children()->listed()->paginate(2)`
2. Try open a page with a stupidly high page number

**Expected behavior**
You should be redirected to the error page.

**Screenshots**
screenshot_2019-06-28_at_12 27 45

**Kirby Version**
This is happening in the current 3.2.0 but @texnixe has tested this and told me it goes back to 3.1.0

lukasbestle:
## Describe the PR

New `$pagination->hasCurrentPage()` method that can be used in the controller to check if an overflow happened and to react accordingly (e.g. redirect to error page, render custom error, redirect to last existing page).

## Related issues

- Fixes #1887

## Todos

- [x] Pass all unit tests

### DIFF
--- a/src/Toolkit/Pagination.php
+++ b/src/Toolkit/Pagination.php
@@ -123,11 +123,11 @@ class Pagination
     {
         if ($page === null) {
             if ($this->page > $this->pages()) {
-                $this->page = $this->lastPage();
+                return $this->lastPage();
             }
 
             if ($this->page < 1) {
-                $this->page = $this->firstPage();
+                return $this->firstPage();
             }
 
             return $this->page;
--- a/src/Toolkit/Pagination.php
+++ b/src/Toolkit/Pagination.php
@@ -254,6 +254,18 @@ class Pagination
     }
 
     /**
+     * Checks if the current page exists
+     * (returns false if the provided page has
+     * overflowed the range of pages)
+     *
+     * @return boolean
+     */
+    public function hasCurrentPage(): bool
+    {
+        return $this->hasPage($this->page);
+    }
+
+    /**
      * Checks if the given page exists
      *
      * @return boolean
--- a/tests/Toolkit/PaginationTest.php
+++ b/tests/Toolkit/PaginationTest.php
@@ -181,6 +181,30 @@ class PaginationTest extends TestCase
         $this->assertEquals(10, $pagination->offset());
     }
 
+    public function testHasCurrentPage()
+    {
+        $pagination = new Pagination([
+            'page'  => 5,
+            'limit' => 1,
+            'total' => 10
+        ]);
+        $this->assertTrue($pagination->hasCurrentPage());
+
+        $pagination = new Pagination([
+            'page'  => 42,
+            'limit' => 1,
+            'total' => 10
+        ]);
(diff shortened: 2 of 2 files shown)


afbora:
This will not work, `hasCurrentPage()` method returns always `true` because we overwrite it on line:

https://github.com/getkirby/kirby/blob/develop/src/Toolkit/Pagination.php#L125-L127

lukasbestle:
That's true. Once the `$pagination->page()` method gets called before, it will break. I checked it and thought "well, that's only when getting the value", but you are right that it gets overwritten.

afbora:
But I didn't understand how the test passed.

Maybe we should use `Pagination::for()` method for test like using in `Collection` class

https://github.com/getkirby/kirby/blob/develop/src/Cms/Collection.php#L224

lukasbestle:
I have fixed it by removing the cache in the `page()` method. See the commit description for my reasoning.

bastianallgeier:
The method is a good idea, but I think it doesn't really solve the general problem, because you still need to handle this manually. 

In theory we could just throw an exception when there's an overflow and this would work great for the templates, but we don't want the same behavior for the API I think. Maybe we need to be able to set the Pagination class into some sort of failure mode.

lukasbestle:
Throwing an Exception is actually a pretty great idea.

I was thinking how we could build an error page redirect without manually putting one in (because that wouldn't be flexible enough and also wouldn't allow overriding it). But exceptions can be caught and therefore a custom behavior is always possible.
I actually think that an Exception would also be fine for the API – after all the pagination page doesn't exist and shouldn't be accessible. Right now it overflows in the same way so an API client could think that there are the same items again on the next page. It won't know that the page has overflowed without checking the pagination page value in the response.

The only issue I'm having right now is that I can't get the `Kirby\Exception\NotFoundException` to work. Wasn't there a way to throw one from a template for example and get an error page rendered? Then the implementation for the pagination class would be super simple.

Moving this to the next release because we won't be able to implement it in time for 3.2.5.

afbora:
I've tried `NotFoundException` before on this issue, i could not.

I agree with @lukasbestle. `NotFoundException` with error page if we can somehow integrate, I think it will work in many places not only pagination.

bastianallgeier:
I was wrong by thinking that the exception is caught and leads to the error page. If that was the case the debugger would never really work. 

But we could introduce a `ErrorPageException` or an `errorPage()` helper for such occasions. I think this would be super useful for controllers anyway. Such a special exception could be caught when trying to render the page and could then actually load the error page instead. What do you think?

distantnative:
Sounds good to me!

lukasbestle:
Yep, that would make a ton of sense. :)

lukasbestle:
Replaced by #2147.

PR #2101, opened 2019-09-15
Refine PHP CS fixer rules

distantnative:
## Describe the PR

**First merge all other 3.2.5 PRs**

Add some more useful rules next to PSR2 to PHP CS Fixer setup. Helps us keeping our code style consistent and remove clutter.

All changes have been made through running the composer script - all automated.


lukasbestle (commented):
Looks good to me!

I have scanned the list of PHP-CS-Fixer options and found the following ones that might also make sense for us:

- `'align_multiline_comment' => ['comment_type' => 'phpdocs_like']`
  Fix indentation of multiline comments with asterisks.
- `'array_indentation' => true`
  Each element of an array must be indented exactly once.
- `'cast_spaces' => ['space' => 'none']`
  No space should be between cast and variable.
- `'class_keyword_remove' => true`
  Converts ::class keywords to FQCN strings.
- `'combine_nested_dirname' => true`
  Replace multiple nested calls of dirname by only one call with second $level parameter. Requires PHP >= 7.0.
- `'dir_constant' => true`
  Replaces dirname(__FILE__) expression with equivalent __DIR__ constant.
- `'function_typehint_space' => true`
  Ensure single space between function's argument and its typehint.
- `'include' => true`
  Include/Require and file path should be divided with a single space. File path should not be placed under brackets.
- `'logical_operators' => true`
  Use && and || logical operators instead of and and or.
  **Note:** We need to check each occurrence that gets fixed by this and verify if the behavior changed.
- `'lowercase_cast' => true`
  Cast should be written in lower case.
- `'lowercase_static_reference' => true`
  Class static references self, static and parent MUST be in lower case.
- `'magic_constant_casing' => true`
  Magic constants should be referred to using the correct casing.
- `'magic_method_casing' => true`
  Magic method definitions and calls must be using the correct casing.
- `'method_chaining_indentation' => true`
  Method chaining MUST be properly indented. Method chaining with different levels of indentation is not supported.
- `'modernize_types_casting' => true`
  Replaces intval, floatval, doubleval, strval and boolval function calls with according type casting operator.
- `'multiline_comment_opening_closing' => true`
  DocBlocks must start with two asterisks, multiline comments must start with a single asterisk, after the opening slash. Both must end with a single asterisk before the closing slash.
- `'native_function_casing' => true`
  Function defined by PHP should be called using the correct casing.
- `'native_function_type_declaration_casing' => true`
  Native type hints for functions should use the correct case.
- `'new_with_braces' => true`
  All instances created with new keyword must be followed by braces.
- `'no_blank_lines_after_class_opening' => true`
  There should be no empty lines after class opening brace.
- `'no_blank_lines_after_phpdoc' => true`
  There should not be blank lines between docblock and the documented element.
- `'no_empty_comment' => true`
  There should not be any empty comments.
- `'no_empty_phpdoc' => true`
  There should not be empty PHPDoc blocks.
- `'no_empty_statement' => true`
  Remove useless semicolon statements.
- `'no_leading_namespace_whitespace' => true`
  The namespace declaration line shouldn't contain leading whitespace.
- `'no_mixed_echo_print' => ['use' => 'echo']`
  Either language construct print or echo should be used.
- `'no_unneeded_control_parentheses' => true`
  Removes unneeded parentheses around control statements.
- `'no_useless_else' => true`
  There should not be useless else cases.
- `'ordered_imports' => ['sort_algorithm' => 'alpha']`
  Ordering use statements.
- `'phpdoc_add_missing_param_annotation' => ['only_untyped' => false]`
  PHPDoc should contain @param for all params.
- `'phpdoc_align' => ['align' => 'left']`
  All items of the given phpdoc tags must be left-aligned.
- `'phpdoc_indent' => true`
  Docblocks should have the same indentation as the documented subject.
- `'phpdoc_order' => true`
  Annotations in PHPDoc should be ordered so that @param annotations come first, then @throws annotations, then @return annotations.
- `'phpdoc_scalar' => true`
  Scalar types should always be written in the same form. int not integer, bool not boolean, float not real or double.
- `'phpdoc_separation' => true`
  Annotations in PHPDoc should be grouped together so that annotations of the same type immediately follow each other, and annotations of a different type are separated by a single blank line.
- `'phpdoc_trim' => true`
  PHPDoc should start and end with content, excluding the very first and last line of the docblocks.
- `'short_scalar_cast' => true`
  Cast (boolean) and (integer) should be written as (bool) and (int), (double) and (real) as (float), (binary) as (string).
- `'simplified_null_return' => true`
  A return statement wishing to return void should not return null.
- `'single_line_comment_style' => true`
  Single-line comments and multi-line comments with only one line of actual content should use the // syntax.
- `'single_quote' => true`
  Convert double quotes to single quotes for simple strings.
- `'ternary_to_null_coalescing' => true`
  Use null coalescing operator ?? where possible. Requires PHP >= 7.0.
- `'whitespace_after_comma_in_array' => true`
  In array declaration, there MUST be a whitespace after each comma.

What do you think? Quite a lot of rules, but the risk should be low (unlike with some other rules I have skipped). Of course we need to check the code that gets actually changed by this, but that's a one time process.

---

**IMPORTANT**: We should *not* merge this until all other 3.2.5 PRs are merged, otherwise we are likely going to have merge conflicts. After everything is merged, we should rebase this PR against `develop`, rerun the fixer and then merge this PR.

distantnative:
I will try them out - will be quite massive.
Why would you include this?

'class_keyword_remove' => true
Converts ::class keywords to FQCN strings.

I actually like the `File::class` way :D

lukasbestle:
> I will try them out - will be quite massive.

That's definitely true. I'm happy to review the changes though as it's an investment for the future.

> Why would you include this?
>
> 'class_keyword_remove' => true
> Converts ::class keywords to FQCN strings.
>
> I actually like the File::class way :D

I do too, but @bastianallgeier has pointed out that this syntax triggers the autoloader even if the class is actually never used in the current request.

distantnative:
Added them. I think we need to review carefully. Especially the `no useless else` rule moves code around quite a bit.

lukasbestle:
That's true, some of those aren't actually that useless.

I will play around with it locally and remove some rules that sounded better than they are. :)

distantnative:
@lukasbestle Wondering if we should do this in two steps to easier review: First all the actual code rules. And ones we have reviewed them add the docblock rules.

lukasbestle:
I think I will actually remove some of the docblock rules again. Especially the one that groups the at-rules. Doesn't make too much sense.

Let's see how many changes are still left after that. ;)

distantnative:
Which one I already don't like:

`phpdoc_separation`
It's too heavy especially on class blocks - we don't need empty lines between author, link etc.

distantnative:
Only pushed the config files changes for the moment. So this PR is readable :D

lukasbestle:
I have removed some rules that either don't make sense for us or that cause unwanted side-effects (I commented those out with an explanation so that we don't add them back by accident).

The two remaining phpdoc rules can only be added once the other PRs are merged – otherwise they will cause merge conflicts.

I have reviewed all changes (and manually reworked them where needed) from the first set of rules.

src/Cms/App.php
```diff
@@ -1196,9 +1196,9 @@ public function smartypants(string $text = null): string
      * and return a template snippet
      *
      * @internal
-     * @return string
      * @param mixed $name
      * @param array $data
+     * @return string
```
distantnative: Should be string|null?

lukasbestle: Yes, that's right. But I think those are errors we need to fix manually when we see them as it was like this before as well. Not going to read through all doc blocks, just the ones that got changed by the fixer. ;)

I'm planning to continue with general refactoring for 3.3, so I will fix issues like this over time.

distantnative:
@lukasbestle About your manual fixes: are they done in a way that they won't get reverted (or rather lead to an error) when the CI runs it? I assume since the checks here are all passed, but just wanted to make sure before manual work gets lost :)

bastianallgeier:
IMHO this PR is to massive for a patch release. We should think about individual PRs for each new option.

distantnative:
We could certainly split it into a few PRs. Wouldn't do one for each option, but maybe group things like docblocks, `use`, array notation etc. in individual ones. If that helps you.

I think it only looks massive cause so many files got touched. In the end it is very minimal in cosequence..

lukasbestle:
> @distantnative About your manual fixes: are they done in a way that they won't get reverted (or rather lead to an error) when the CI runs it? I assume since the checks here are all passed, but just wanted to make sure before manual work gets lost :)

Yes, they won't be corrected back. The issues I had to fix were mainly places where the library didn't clean up the code properly after doing its thing (for example removing all `use` declarations but not the empty line after the list of `use` declarations or adding `@param` declarations at the end of the docblock and not where they belonged) as well as issues the library could detect but not properly fix. I verified that the library "accepts" those changes by running it again after committing the manual changes.

> @bastianallgeier IMHO this PR is to massive for a patch release. We should think about individual PRs for each new option.

I knew you would say that. :)
And I understand you – there are many changed files and a lot of changes.

But this time I really have to say: It would be *a lot* more work to split this up into separate PRs (even if we group the rules). The rules sometimes work together, so the same lines would get changed multiple times. That's even more work to review. And in the end it makes the whole set of changes harder to understand when they are split over countless commits. Imagine what the file history view would look like. ;)

And given that I've reviewed all changes manually and verified that no logic was changed whatsoever (= it doesn't matter if it's a patch or feature release), I think we are save to merge this. :)

The remaining two rules are really something for a separate PR though as they only modify docblocks, but *massively* modify them. That's easier to review in isolation.

distantnative:
I think the only real call in this PR is if we want to change the class notation style from `Page::class` to `'Kirby\Cms\Page'` - e.g. like https://github.com/getkirby/kirby/pull/2101/files#diff-f3681b227d1a9ee8e9f258315d9d13abR46.

Everything else really seems more decorative and I think since Lukas reviewed everything, it'd be safe to merge.

bastianallgeier:
I scrolled through most of the changes and they all look useful and cleaner than before. So let's merge it!

PR #2104, opened 2019-09-15
Refactor number field #1748

### LINKED ISSUE #1748: [Panel] Number field removes decimal values when they are 0.
**Describe the bug**
When a number field is set to decimal values using the step option:

```
fields:
  price:
    label: Price
    type: number
    step: .01
```
and you select a number where the decimals are zero, e.g. 15,00, the zeros are removed when saving or reloading, also the number is stored without the decimals.

**To Reproduce**
1. Add above field in any blueprint
2. Visit page in panel and add number,  e.g 15,00
3. Save and reload
4. See zeros vanishing. They are also not stored.

**Expected behavior**
Zeros should be stored in the content file and not disappear.

**Kirby Version**
3.1.3

https://forum.getkirby.com/t/decimal-number-fields-dont-allow-00-anymore/14066

distantnative:
## Describe the PR
Problem with the existing number field:
Having a step like `0.01` would result in omitted zeros e.g. at `0.1` or `0.2` instead of `0.10` and `0.20`- both in the UI as well as the content file. Especially when using the number field for currencies that behavior is not great.

I could not get any solution working with the HTML5 number input.
So I thought, why not implement it as a normal text field instead with support for arrow keys etc.
Arrow keys get emitted directly, manual input on blur (otherwise user could never finish typing).
Also added rounding to next step.
 

## Related issues
- Fixes #1748

### DIFF
--- a/config/fields/number.php
+++ b/config/fields/number.php
@@ -24,7 +24,7 @@ return [
          * Allowed incremental steps between numbers (i.e `0.5`)
          */
         'step' => function ($step = 1) {
-            return $this->toNumber($step);
+            return (float)$this->toNumber($step);
         },
         'value' => function ($value = null) {
             return $this->toNumber($value);
@@ -36,7 +36,7 @@ return [
                 return null;
             }
 
-            return (float)Str::float($value);
+            return Str::float($value);
         }
     ],
     'validations' => [
--- a/panel/src/components/Forms/Field/NumberField.vue
+++ b/panel/src/components/Forms/Field/NumberField.vue
@@ -20,7 +20,11 @@ export default {
   props: {
     ...Field.props,
     ...Input.props,
-    ...NumberInput.props
+    ...NumberInput.props,
+    icon: {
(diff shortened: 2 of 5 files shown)


config/fields/number.php
```diff
@@ -36,7 +36,7 @@
                 return null;
             }
 
-            return (float)Str::float($value);
+            return Str::float($value);
```
distantnative, line 39: Just thinking myself... Maybe we should rename this methods to `toValue` and add another `toNumber` one that just does `return (float)$this->toValue($value)`, @bastianallgeier?

bastianallgeier (commented):
To be honest, this is a step in the wrong direction in my opinion. We shouldn't move away from the number input. It has too many benefits to stick with the native input. Especially with all the built in events and input mode for mobile devices.

distantnative:
You are right. I’m just afraid that with the native field we will never be able to close the issue.

PR #2105, opened 2019-09-15
Api: user language as locale #1872

### LINKED ISSUE #1872: strftime in the panel ignore local setting with toDate()
**Describe the bug**
With the `strftime` date handler, the date is not translated when used in the panel.

**To Reproduce**
1. Clone [the starterkit](https://github.com/getkirby/starterkit)
2. Change the date handler in the config for strftime
```
'date' => [
    'handler' => 'strftime'
],
```
3. Edit the `info` key in the `sections/notes.yml` blueprint for:
`info: "{{ page.date.toDate('%A %d %B %Y') }}"`

4. Go to the panel and install it in French.

**Expected behavior**
Date should be translated in French, like Mercredi 31 octobre 2018, Vendredi 05 octobre 2018, Mercredi 26 Septembre 2018…

**Screenshots**
The notes list in the Kirby panel

**Kirby Version**
Tested on 3.1.3 and 3.1.4.

**Additional context**
The date handler with the same `toDate()` I used work well in a template.

distantnative:
## Describe the PR
Set API locale based on the user language

## Related issues
- Fixes #1872


src/Api/Api.php
```diff
@@ -169,7 +169,11 @@ public function call(string $path = null, string $method = 'GET', array $request
         $auth   = $this->route->attributes()['auth'] ?? true;
 
         if ($auth !== false) {
-            $this->authenticate();
+            $user = $this->authenticate();
+            
+            if (is_a($user, 'Kirby\Cms\User') === true) {
+                setlocale(LC_ALL, $user->language());
```
lukasbestle: I'm not sure if `setlocale()` accepts simple language codes as locales. But I also don't know how to do this in a better way.

Maybe we could check if the language code contains an underscore (like `pt_BR`). If not, we would do `$user->language() . '_' . strtoupper($user->language())`. Then again it's unclear whether we should add system fallback locales like `.ISO8859-1` or `.UTF-8` or `.UTF8`.

distantnative: I honestly do know too little about locales in PHP to chime in here.

lukasbestle:
I have implemented it like I suggested yesterday and it seems to work pretty great. It doesn't work for all Panel translations (because the hack only works for some locales, maybe also because my Mac doesn't seem to have all locales we have translations for), but in those cases it will fall back to English like before – so not a huge deal.

translation

The manual reload is necessary because the Panel only reloads the UI and not the content when the translation is changed. But because the content is automatically reloaded when switching pages, it would only be an issue in user blueprints.

lukasbestle (approved):

bastianallgeier (approved):

PR #2124, opened 2019-09-18
Update issue template

distantnative:
@lukasbestle I saw your changes for the PR template. Would these apply here as well in your view?


bastianallgeier (approved):
I think they are super useful here as well.

PR #2131, opened 2019-09-20
Fix "Class 'I18n' not found" error

### LINKED ISSUE #2128: [Panel] Class 'I18n' not found
**Describe the bug**
Sections and some fields throw error following:

> "albums" bölümü yüklenemedi: Class 'I18n' not found

**As english translation**

> The section "albums" could not be loaded: Class 'I18n' not found

Error in "website" field
Error in "author" field
Error in "gallery" field
...
...
> Class 'I18n' not found

**To Reproduce**

1. Go to panel users
2. Edit own account
3. Change to language as `Turkish`
4. Try these actions: Go to dashboard, edit any page, edit any user
5. See error

**Kirby Version**
3.2.5-RC.2

**Console output**
```
code: 500
details: []
exception: "Error"
file: "/kirby/config/sections/mixins/empty.php"
key: null
line: 9
message: "Class 'I18n' not found"
route: "site/sections/([a-zA-Z0-9\.\-_%= \+\@\(\)]+)"
status: "error"
```

**Desktop:**
 - Centos 6
 - Chrome 77

**Additional context**
This issue is about locale `LC_CTYPE` constant (starting file/class names with `I` letter) and it was working by default in 3.2.4.

When we want to use `locale` as turkish, we use it like following, not just for Kirby.
https://getkirby.com/docs/reference/system/options/locale#detailed-settings

Even if we want to use custom locale, the error continues.
I would expect it to work without error, even if we didn't use custom locale as before versions, 

**Note**:

When i add `use Kirby\Toolkit\I18n;` to top of files that have error fields/sections, error gone 🤔

lukasbestle:
## Describe the PR

Fixes errors that occur because of the wrong locale set for `LC_CTYPE` (introduced in 3.2.5-rc.1).

Also two other minor improvements to the code:

- Also pass language code as locale fallback for the new `Api` locale feature
- Add missing `use` declaration in the file that caused the error originally

## Related issues

- Fixes #2128

## Todos

- [x] Pass all unit tests

### DIFF
--- a/config/sections/mixins/empty.php
+++ b/config/sections/mixins/empty.php
@@ -1,5 +1,7 @@
 <?php
 
+use Kirby\Toolkit\I18n;
+
 return [
     'props' => [
         /**
--- a/config/api/authentication.php
+++ b/config/api/authentication.php
@@ -1,6 +1,5 @@
 <?php
 
-use Kirby\Cms\Auth;
 use Kirby\Exception\PermissionException;
 
 return function () {
--- a/config/api/collections.php
+++ b/config/api/collections.php
@@ -1,12 +1,5 @@
 <?php
 
-use Kirby\Cms\Files;
-use Kirby\Cms\Languages;
-use Kirby\Cms\Pages;
-use Kirby\Cms\Roles;
-use Kirby\Cms\Translations;
-use Kirby\Cms\Users;
-
 /**
  * Api Collection Definitions
  */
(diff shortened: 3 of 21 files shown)


config/sections/mixins/empty.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use Kirby\Toolkit\I18n;
```
afbora, line 3: Now it works without this line because of other updates. Main issue was about `LC_CTYPE`

lukasbestle: Awesome! Thanks for verifying. 👍

bastianallgeier (approved):

PR #2139, opened 2019-09-26
Fix invalid regex with enhance escape method

### LINKED ISSUE #2138: Panel: "Invalid regular expression: nothing to repeat" with Custom Slug Rules Containing `+` (3.2.5 Regression)
**Describe the bug**  
When creating a new page in the Panel, as you type in the Title field, an error is displayed that says "Invalid regular expression: nothing to repeat"

It appears the regression has to do with a custom slug override that includes a `+`. This was previously fixed with https://github.com/getkirby/kirby/issues/1946, but seems to have been reverted?

In my `config.php`, for example:
```php
Str::$language = [
    '+' => '-plus',
];
```

**To Reproduce**  
1. Have custom slugs defined like above
2. Click on a "Create" button in a Pages section
3. Type a character of a title

**Screenshots**  
Screen Shot 2019-09-25 at 11 54 49 AM

**Kirby Version**  
`3.2.5`

**Console output**  
None

 - OS: macOS
 - Browser: Safari
 - Version: Version 13.0.1

**Additional context**  
No issues on `3.2.4`

afbora:
## Describe the PR

I think I've found the right escape method after hundreds of tests on 3.2.5.

**Informations**:
- Learned that the `u` flag should be used for unicode property `\p`
- `/gu` flags together was not working in this case
- Special characters (`-\/\\^$*+?.()|[\]{}`) were not escaped somehow (main reason of this issue) because we added `\p` property in our last commit
- Added condition that escape unicode or special characters to work with unicode and special characters (I'd appreciate it if you have a better regex syntax)
- `\p{L}` mean is escape only unicode letters as `{L}`

In summary, the PR does: escaping only unicode letters and defined special characters.

## Related issues

- Fixes #2138 

## Ready?

- [ ] Passing all unit tests

### DIFF
--- a/panel/src/helpers/regex.js
+++ b/panel/src/helpers/regex.js
@@ -1,1 +1,1 @@
-RegExp.escape = s => s.replace(/\p[-/\\^$*+?.()|[\]{}]/g, '\\$&');
+RegExp.escape = s => s.replace(/[\p{L}]|[-\/\\^$*+?.()|[\]{}]+/u, '\\$&');


afbora:
@bastianallgeier  Could you add unit test for plus sign, please?

bastianallgeier:
I just added a unit test for the plus sign issue and it works perfectly. Thank you so much for debugging this. 

What do you think about joining our organisation on Github? We could massively simplify the PR process.

afbora:
> What do you think about joining our organisation on Github? We could massively simplify the PR process.

I've always been happy to contribute to the Kirby and be a member of the family ❤️

PR #2142, opened 2019-09-27
Add margin to all dropdowns #1563

### LINKED ISSUE #1563: Annoyance with a multiselect dropdown at the end of a page
**Describe the bug**
An edge case for having a not native dropdown at the end of the page. There are actually 4 items that could be selected but he save bar overlaps the last item.

bug

**Expected behavior**
Be able to select all items 😉 

**Kirby Version**
3.0.3

**Desktop**
 - Windows 10
 - Firefox 66

**Additional context**
It's not that big of deal. More like an annoyance. The problem also only occurs if you try to add more than one item in one go. Chrome kinda works as the last item is at least visible.

grafik

bastianallgeier:
## Describe the PR

Dropdowns don't have enough space when they are at the bottom of a view. They get clipped by the form button bar. This PR adds a bottom margin to all dropdowns, which will automatically increase the scroll height of the view. 

## Related issues

- Fixes #1563

### DIFF
--- a/panel/src/components/Navigation/DropdownContent.vue
+++ b/panel/src/components/Navigation/DropdownContent.vue
@@ -97,47 +97,47 @@ export default {
           break;
         case "ArrowUp":
           e.preventDefault();
-          
+
           while (true) {
             this.current--;
-            
+
             if (this.current < 0) {
               this.close();
               this.$emit("leave", e.code);
               break;
             }
-             
+
             if (
-              this.$children[this.current] && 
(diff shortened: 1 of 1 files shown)


lukasbestle (commented):
Does this make the page jump when opening/closing a dropdown at the bottom of the page? If it doesn't, then the solution is 👍.

bastianallgeier:
@lukasbestle nope, it doesn't jump

PR #2145, opened 2019-09-27
API: Respect configured locale on single-lang site

lukasbestle:
The `$kirby->setCurrentLanguage()` method not only sets the active language, but also the active PHP locale. It should be called for every API request, no matter if a language is specified or not.

The method will automatically fall back to the default language and its locale – or on single-lang sites to the configured `locale` option.

---

Forum thread: https://forum.getkirby.com/t/using-sortby-with-special-language-specific-characters/15781?u=lukasbestle


bastianallgeier (commented):
I think it would be good to have a single language and multi language test case for that.

bastianallgeier (approved):

PR #2147, opened 2019-09-29
Add ErrorPageException, fix pagination overflow

### LINKED ISSUE #1887: Pagination doesn't redirect to error page, shows last page instead.

lukasbestle:
## Describe the PR

- Add new `ErrorPageException` that can be thrown while a page is rendered (e.g. from the template or controller) to trigger the error page
- Replace the single setters in the `Pagination` class with a new `setProperties()` method that validates the properties when setting
- Trigger the error page if the pagination page is out of bounds

## Related issues

- Fixes #1887

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/sections/mixins/pagination.php
+++ b/config/sections/mixins/pagination.php
@@ -14,7 +14,7 @@ return [
          * Sets the default page for the pagination. This will overwrite default pagination.
          */
         'page' => function (int $page = null) {
-            return get('page', $page ?? 1);
+            return get('page', $page);
         },
     ],
     'methods' => [
--- a/src/Toolkit/Pagination.php
+++ b/src/Toolkit/Pagination.php
@@ -2,7 +2,8 @@
 
 namespace Kirby\Toolkit;
 
-use Exception;
+use Kirby\Exception\ErrorPageException;
+use Kirby\Exception\Exception;
 
 /**
  * Basic pagination handling
@@ -27,14 +28,14 @@ class Pagination
      *
      * @var int
      */
-    protected $total;
+    protected $total = 0;
 
     /**
      * The number of items per page
      *
      * @var int
      */
-    protected $limit;
+    protected $limit = 20;
 
     /**
      * Creates a new pagination object
@@ -44,9 +45,7 @@ class Pagination
      */
     public function __construct(array $params = [])
     {
-        $this->page($params['page'] ?? 1);
-        $this->limit($params['limit'] ?? 20);
-        $this->total($params['total'] ?? 0);
+        $this->setProperties($params);
     }
 
     /**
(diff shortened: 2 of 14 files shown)


afbora:
FireShot Capture 772 - Kirby CMS Debugger - http___localhost_test_kirby_2147_page;2

Didn't work for me when i try to access second page `/page;2`. Am I wrong?

I checked the `$params['page']` variable, it has value `2` as `string` not `int`.

I think you may have passed the test because you tested directly with an integer value in the unit test.

lukasbestle:
You are right, the validation was too strict. I have fixed it now.

afbora:
Works great now! 👍 

Btw do we need to check the float numbers? 
I think it will always work because we cast to int, but I wanted to mention it anyway. 
For ex: `/page;3.14`

lukasbestle:
There is unfortunately no native PHP function to check for strings that may only contain ints. Options are:

```php
(string)(int)$value === (string)$value
ctype_digit((string)$value) === true
```

Both are very clumsy and long, therefore I figured that the check for `is_numeric()` should be enough here. As you wrote, the value is casted to int anyway, so it doesn't really matter.

config/sections/mixins/pagination.php
```diff
@@ -14,7 +14,7 @@
          * Sets the default page for the pagination. This will overwrite default pagination.
          */
         'page' => function (int $page = null) {
-            return get('page', $page ?? 1);
+            return get('page', $page);
```
bastianallgeier, line 17: Is there a special reason not to fall back to 1 as the default page?

bastianallgeier (commented):
Couldn't we still keep the single setters by using the setProperties method there as well?

Otherwise we should think about refactoring it completely in the same style we handle immutable objects in other places and provide a clone method instead of setProperties.

config/sections/mixins/pagination.php
lukasbestle: The automatic fallback now happens directly in the `setProperties()` method. The reason we need to do it there is that it would otherwise throw an Exception if the default page of 1 is passed for a collection with 0 items (= 0 pages).

lukasbestle:
@bastianallgeier The question is why we had the setters in the first place. I think an immutable object with a clone function would indeed be the cleanest way.

In any case we need to remove the individual setters, otherwise the validation won‘t kick in.

bastianallgeier:
@lukasbestle let's use this chance to move to our normal pattern with private setters that get called  once in the constructor and a clone method.

lukasbestle:
Alright, will do that soon.

src/Toolkit/Pagination.php
```diff
@@ -397,6 +381,86 @@ public function rangeEnd(int $range = 5): int
...
+        // generate "default page" based on other params
+        if ($this->page === null) {
+            $this->page = $this->firstPage();
+        }
+
+        // validate page based on all params
+        $min = $this->firstPage();
+        $max = $this->pages();
+        if ($this->page < $min || $this->page > $max) {
+            throw new ErrorPageException('Pagination page ' . $this->page . ' is out of bounds, expected ' . $min . '-' . $max);
```
bastianallgeier, line 405: I think we need to make this translatable and also make it less technical. This will be passed to the error page as $errorMessage variable and can be used to display a custom error message there. So I think it should be a `human` as possible.

bastianallgeier: I take my comment back. I think it's fine. We can't translate it in the toolkit class anyway.

bastianallgeier (dismissed):

PR #2148, opened 2019-09-29
Add $visitor->preferredMimeType() and $visitor->prefersJson(), remove deprecated methods

lukasbestle:
This PR **removes**:

1. the methods that have been renamed in 3.0.0 and have been deprecated since,
2. the `$request->ajax()` method (deprecated in 3.1.0, can be replaced with `$visitor->prefersJson()`) and
3. the `translations` root that has been renamed to `i18n:translations` in 3.2.0.

It also **adds** new `$visitor->preferredMimeType()` and `$visitor->prefersJson()` methods as a replacement for the removed AJAX check.

We should explicitly name all removed methods (and the respective new methods) in the changelog so that users that haven't migrated to the new names yet can do so without much work.

### DIFF
--- a/config/roots.php
+++ b/config/roots.php
@@ -11,17 +11,10 @@ return [
         return $roots['kirby'] . '/i18n';
     },
     'i18n:translations' => function (array $roots) {
-        return $roots['translations'];
+        return $roots['i18n'] . '/translations';
     },
     'i18n:rules' => function (array $roots) {
         return $roots['i18n'] . '/rules';
-    },
-    /**
-     * @deprecated 3.2.0 Use `i18n:translations` instead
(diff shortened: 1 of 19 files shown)


bastianallgeier (commented):
TBH I would prefer to have separate PRs for the different tasks in here. 

We should really try to split PRs into independent chunks depending on the issue they solve or feature they implement. In this case: 

- root changes
- deprecated methods
- mime & visitor class changes

The main reason here is that I cannot really approve the removal of all deprecated methods. We should discuss them some more first. The mime & visitor changes look good but they are hard review properly with all the other changed files. The root changes are also completely separate for me.

lukasbestle:
You are right. :)

PR #2151, opened 2019-09-29
assertEquals() -> assertSame() for the first tests

lukasbestle:
I changed all tests I have already refactored because I know that those won't break because of the change. The others will be migrated over time when I refactor them.


bastianallgeier (approved):
Awesome!

PR #2153, opened 2019-09-30
Fix broken replace action for files #1951

### LINKED ISSUE #1951: Files Section "Replace" Menu Item Broken in Safari
**Describe the bug**
Within a Files section, if you use the context menu on a file, you cannot replace a file in Safari. Works fine in Chrome. You click it and nothing happens.

**Screenshots**
2019-07-26 at 3 33 PM

**Kirby Version**
`3.2.2`

**Console output**
None

 - OS: macOS 10.15 Beta (19A512f)
 - Browser: Safari 13.0 (15608.1.34.1.1)

**Additional context**

bastianallgeier:
## Describe the PR

Content locking added a new layer of requests between the called action in dropdowns and the actual execution. This was no issue for all regular actions, but opening a window or opening the uploader was blocked by this in Safari and Firefox. This is now fixed by removing the additional API request to get the lock. The lock information is now used in the options API calls to disable options right away when the page, file or user is locked. This is a far better approach IMHO, as it makes the option API a lot more useful for other calls as well. 

## Related issues

- Fixes #1951

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/api/models/File.php
+++ b/config/api/models/File.php
@@ -54,7 +54,16 @@ return [
             return $files->nth($index + 1);
         },
         'options' => function (File $file) {
-            return $file->permissions()->toArray();
+            $options = $file->permissions()->toArray();
+            $lock    = $file->lock();
+
+            if ($lock && $lock->isLocked()) {
+                foreach ($options as $key => $value) {
+                    $options[$key] = false;
+                }
+            }
+
+            return $options;
         },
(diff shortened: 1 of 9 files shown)


lukasbestle (commented):
I think we should have a common helper for the "set all options to `false` that are not in the whitelist" code, otherwise we will have duplicated code.

PR #2154, opened 2019-09-30
Store User ID in Users field instead of email #1885

### LINKED ISSUE #1885: [Panel] Users Field Should Save Persistent ID, Not Email Address
**Describe the bug**
We just ran into an unexpected issue today, if you change the email address of a user, pages that had a Users field are no longer linked to the affected user.

I genuinely thought the Users field would save a list of persistent User IDs (i.e. the folder name that contains an individual user's files). If it had done this, renaming an email address wouldn't break existing pages (it breaks now because `$field->toUsers()` finds no matching email address.

**What's Actually Saved in a Content File**
```yaml
Authors: 

- foo@example.com
```

**What I Would Expect to be Saved**
```yaml
Authors: 

- By4XbnMB
```

**To Reproduce**
1. Have a user with an email `foo@example.com` ("Example User")
2. Go to a page with a Users field, select "Example User"
3. Change the email address for "Example User" to `bar@example.com`
4. Attempt to do `$field->toUsers()` on the original page

**Expected behavior**
The user should still be accessible.

**Kirby Version**
`3.1.x` and `3.2.0`

bastianallgeier:
## Describe the PR

Store User ID in Users field instead of email

## Related issues

- Fixes #1885

## Ready?
- [x] Passing all unit tests

### DIFF
--- a/config/fields/users.php
+++ b/config/fields/users.php
@@ -85,7 +85,7 @@ return [
         ];
     },
     'save' => function ($value = null) {
-        return A::pluck($value, 'email');
+        return A::pluck($value, 'id');
     },
     'validations' => [
         'max',


afbora:
> 3.3.0 is a good point to introduce this and it‘s luckily 100% backward compatible.

That mean will `email` data be automatically migrated as `id` on a site upgraded from version 3.2.x to 3.3.0?

bastianallgeier:
@afbora no, that would be too complex. But every time you save the users field again, the emails will be replaced.

lukasbestle (approved):
Awesome that our `toUser()` field method already supports both. So this change is no problem. 👍

PR #2162, opened 2019-10-01
Add smartypants for languages

### LINKED ISSUE #2037: Option to define smartypants on a per language basis missing?
**Describe the bug**
I would have expected that it is possible to set smartypants on a per language basis in the language configuration files:

```
<?php

return [
    'code' => 'de',
    'default' => false,
    'direction' => 'ltr',
    'locale' => [
        'de_DE'
    ],
    'name' => 'Deutsch',
    'translations' => [

    ],
    'url' => NULL,
    'smartypants' => [
        'doublequote.open' => "start",
        'doublequote.close' => 'stop',
    ],
];
```

But this doesn't have any effect, it only works when configuring smartypants in the config, but then obviously without language specific settings.

**Expected behavior**
It should be possible to define language-specific smartypants options. Wrong syntax or not possible. Studying the source code it looks as if this feature is not implemented.

**Kirby Version**
Tested with 3.2.3

afbora:
## Describe the PR

If smartypants are defined in the languages, it merges with default smartypants. If there is no default smartypants, it must be enabled as `true` in the config.

In addition, when `smartypants` were disabled as `false`, would give an error when called `smartypants()`, this error is also fixed.

### Usage

````php
<?php

return [
    'code' => 'en',
    'default' => true,
    'direction' => 'ltr',
    'locale' => 'en_US',
    'name' => 'English',
    'translations' => [],
    'url' => NULL,
    'smartypants' => [
        'doublequote.open'  => '<',
        'doublequote.close' => '>'
    ],
];
````

## Related issues

- Closes #2037 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1187,7 +1187,15 @@ class App
         if ($options === true) {
             $options = [];
         }
-
+        
+        if ($this->multilang() === true) {
+            $languageSmartypants = $this->language()->smartypants() ?? [];
+        
+            if (empty($languageSmartypants) === false) {
+                $options = array_merge($options, $languageSmartypants);
+            }
+        }
+        
         return $this->component('smartypants')($this, $text, $options);
     }
 
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1184,18 +1184,20 @@ class App
     {
         $options = $this->option('smartypants', []);
 
-        if ($options === true) {
+        if ($options === false) {
+            return $text;
+        } elseif (is_array($options) === false) {
             $options = [];
         }
-        
+
         if ($this->multilang() === true) {
             $languageSmartypants = $this->language()->smartypants() ?? [];
-        
+
             if (empty($languageSmartypants) === false) {
                 $options = array_merge($options, $languageSmartypants);
             }
         }
-        
+
         return $this->component('smartypants')($this, $text, $options);
     }
 
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -57,7 +57,7 @@ class Language extends Model
      * @var array|null
      */
     protected $slugs;
-    
+
     /**
      * @var array|null
      */
@@ -493,7 +493,7 @@ class Language extends Model
         $this->slugs = $slugs ?? [];
         return $this;
     }
-    
+
     /**
      * @param array $smartypants
      * @return self
@@ -533,9 +533,9 @@ class Language extends Model
     {
         return $this->slugs;
     }
-    
+
     /**
-     * Returns the custom smartypants for this language
+     * Returns the custom SmartyPants options for this language
      *
      * @return array
      */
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/Cms/App.php [hunk 1]
```diff
@@ -1187,7 +1187,15 @@ public function smartypants(string $text = null): string
         if ($options === true) {
             $options = [];
         }
-
+        
+        if ($this->multilang() === true) {
+            $languageSmartypants = $this->language()->smartypants() ?? [];
+        
+            if (empty($languageSmartypants) === false) {
+                $options = array_merge($options, $languageSmartypants);
```
lukasbestle, line 1197: This operation will fail if the `smartypants` config option is set to `false` (you can't merge a boolean with an array).

Now that I think about it: What happens currently if the `$kirby->smartypants()` method is called with SmartyPants disabled in the config? Probably breaks somewhere else anyway. Maybe we should change line `1187` to check for `is_array($options) !== true` instead to make sure that `$options` is always an array, even if SmartyPants is disabled in the config.

src/Cms/Language.php
```diff
@@ -517,6 +533,16 @@ public function slugs(): array
     {
         return $this->slugs;
     }
+    
+    /**
+     * Returns the custom smartypants for this language
```
lukasbestle: I think this should be "custom SmartyPants *options*" to make clearer what the property does.

lukasbestle (changes requested):
It would be great to have these two tests additionally:

- Multilang test where the SmartyPants options are *not* configured in the language config (language config without the `smartypants` array – the parser should use the default rules then.
- General test where the `smartypants` config option is set to `false` – the parser should then also use the default rules.

src/Cms/App.php
afbora: You're right! An error occurs when `smartypants` option set to `false` while `smartypants()` called.

> Argument 3 passed to Kirby\Cms\App::{closure}() must be of the type array, boolean given, called in ...\kirby\src\Cms\App.php on line 1191

We will solve this unpredictable error, thanks to you.

bastianallgeier:
It's great to include this as option in the language files!

bastianallgeier (approved):

afbora:
@bastianallgeier Now i realized that `smartypants` option default is disabled in `kirbytext()` method, enabled in `smartypants()` method. I think both should have the same default setting 🤔 

**kirbytext() method**
https://github.com/getkirby/kirby/blob/develop/src/Cms/App.php#L669-L671

**smartypants() method**
https://github.com/getkirby/kirby/blob/develop/src/Cms/App.php#L1185

lukasbestle:
@afbora No, I think the current behavior is fine:

- When calling the `smartypants()` method directly, you explicitly tell Kirby that you want to have SmartyPants rendered. So if there's no contradictory option set, it should run the parser.
- When calling the `kirbytext()` method on the other hand, SmartyPants should only be rendered automatically if the user has enabled that feature in the Kirby config. Otherwise it would be a breaking change.

PR #2166, opened 2019-10-03
PHPStan configuration and first fixes

### LINKED ISSUE #1990: Use PHPStan for static analysis
Hello!

Many PHPDoc blocks have wanna-be absolute classes `Kirby\Cms\File`
they should really be `\Kirby\Cms\File`

```bash
grep -rFl '@return Kirby' config/ src/ | xargs -t -L1 sed -e 's#@return Kirby#@return \\Kirby#' -i

grep -rFl '@param Kirby' config/ src/ | xargs -t -L1 sed -e 's#@param Kirby#@param \\Kirby#' -i

grep -rFl '@var Kirby' config/ src/ | xargs -t -L1 sed -e 's#@var Kirby#@var \\Kirby#' -i

grep -rFl '|Kirby\' config/ src/ | xargs -t -L1 sed -e 's#|Kirby\\#|\\Kirby\\#' -i
```

After that you may gain the huge benefit of @phpstan as this tool works based on types -  although it takes couple of minutes to fix.

Its simplest form is `phpstan analyze -l 0 src/`

Levels go up to 7.

An example: Cms/PageBlueprint.php:79 Array has 2 duplicate keys with value 0 (0, '0').

Another two: Cms/ModelWithContent.php:399 Access to undefined constant Kirby\Cms\ModelWithContent::CLASS_ALIAS.

Cms/FileVersion.php:92 Kirby\Cms\FileVersion::toArray() calls parent::toArray() but Kirby\Cms\FileVersion does not extend any class.

lukasbestle:
## Describe the PR

I've added a basic PHPStan config file and also some first fixes based on PHPStan level 0.

Remaining issues:

- Analyzing the `tests` folder doesn't work because PHPStan doesn't know about the PHPUnit `TestCase` class. The reason for this is that PHPUnit is not installed directly to the project with Composer to avoid having the whole PHPUnit code in our `vendor` dir in the repo. I couldn't find a way though to pass two autoloader files (the one in the repo as well as the one in the global Composer directory, which would be matched with `%rootDir%/../../autoload.php`) to PHPStan.
- There are a lot of false-positive issues. I've added an `ignoreErrors` rule for one of them, but it's all pretty annoying.

@szepeviktor Do you have experience with PHPStan and could help us out with the issues? Fixing those would be a requirement for adding PHPStan to the Travis test suite and also for upgrading the PHPStan level to 1 (skipping over all errors we can't fix every time is annoying).

## Related issues

- Fixes #1990


szepeviktor:
> Analyzing the tests folder doesn't work because PHPStan doesn't know about the PHPUnit

:) https://github.com/phpstan/phpstan-phpunit

and please see https://github.com/phpstan/phpstan/blob/master/README.md#extensibility

szepeviktor:
> I couldn't find a way though to pass two autoloader files

it is `autoload_files:` in the config file

szepeviktor:
Please add each individual main directory instead of including root and excluding others.

E.g. I think /config should not be analyzed.

szepeviktor:
one more thing: pipe signs should be escapes `\|`

szepeviktor:
`.phpstan.neon` is called `phpstan.neon.dist`

lukasbestle:
Thanks for your replies! 👍

> https://github.com/phpstan/phpstan-phpunit

How can we use that without having to `require` the extension as a Composer dependency? As I wrote above, we can't do that.

> it is autoload_files: in the config file

Ah, thanks! I overlooked that in the README. It doesn't fix the issue with PHPUnit though. As we only really need it to work at all and don't need support for mock classes etc., I would prefer a native solution over the extension (see above). Could you please check why defining the autoloader doesn't fix the `Class PHPUnit\Framework\TestCase not found and could not be autoloaded.` errors?

> Please add each individual main directory instead of including root and excluding others.

Is there any specific reason for that or is it just convention and/or personal preference?

> E.g. I think /config should not be analyzed.

Why that?

szepeviktor:
> How can we use that without having to `require` the extension as a Composer dependency? As I wrote above, we can't do that.

Download the phar file https://github.com/phpstan/phpstan/releases
or [use phive](https://phar.io/)!

szepeviktor:
> > Please add each individual main directory instead of including root and excluding others.
> 
> Is there any specific reason for that or is it just convention and/or personal preference?

In computer science it is lazy to be implicit and beneficial to be explicit.

szepeviktor:
> > E.g. I think /config should not be analyzed.
> 
> Why that?

Basically /config contains hard-coded settings: advanced coding style check and syntax check will discover 99% of the errors.

szepeviktor:
Working on `Class PHPUnit\Framework\TestCase not found` ...

szepeviktor:
:-D :-D :-D PHPUnit needs to installed to make those errors disappear.

`composer require "phpunit/phpunit:^7"`

lukasbestle:
> Download the phat file https://github.com/phpstan/phpstan/releases

You mean the PHAR archive, right? I already do that with `phpstan-shim`, but what about the PHPUnit extension? You linked to the releases of the PHPStan tool itself.

> In computer science it is lazy to be implicit and beneficial to be explicit.

To be honest I don't think this is always the best approach.

If the goal is to catch errors in the code, it can be harmful to accidentally miss parts of the code because you explicitly listed code and at a later point forgot to update the list when the code structure changes.

> Basically /config contains hard-coded settings

It doesn't. Take a look. :)

> :-D :-D :-D PHPUnit needs to installed to make those errors disappear.

Well, I know that. :P
PHPUnit is installed globally (= in `~/.composer/vendor/`), so it should be picked up by the `autoload_files` directive I've configured (because PHPStan is also installed globally). The question is why it still doesn't work with that configured.

---

PS: Could you please reply in one comment and not one per reply? Having that many comments in the issue/PR makes it quite hard to keep an overview.

szepeviktor:
> Could you please reply in one comment

Okay.

I think the phpunit extension and phpunit itself could not be installed globally, phpstan needs the class files.

What is blocking the usage of
`composer require --dev "phpunit/phpunit:^7" "phpstan/phpstan-phpunit" "phpstan/phpstan-shim"` in CI?

lukasbestle:
> I think the phpunit extension and phpunit itself could not be installed globally, phpstan needs the class files.

My mistake, I forgot that my PHPUnit is actually globally installed as PHAR and not via Composer because it conflicts with Valet being installed via Composer. 😠

I'd imagine that it actually already works in CI with the current setup because we install PHPUnit via Composer there (= the PHPUnit source files are available via the autoloader).

---

The remaining problems are the false-positives. Here's a collection of some:

```
 ------ --------------------------------------------------
  Line   config/fields/structure.php
 ------ --------------------------------------------------
  112    Access to an undefined property object::$fields.
 ------ --------------------------------------------------

 ------ ---------------------------------------------------------------------------------------------------------------
  Line   src/Cms/FileVersion.php
 ------ ---------------------------------------------------------------------------------------------------------------
  92     Kirby\Cms\FileVersion::toArray() calls parent::toArray() but Kirby\Cms\FileVersion does not extend any class.
 ------ ---------------------------------------------------------------------------------------------------------------

 ------ ------------------------------------------------------------
  Line   src/Form/Field.php
 ------ ------------------------------------------------------------
  186    Using $this in static method Kirby\Form\Field::defaults().
  187    Using $this in static method Kirby\Form\Field::defaults().
  187    Using $this in static method Kirby\Form\Field::defaults().
  191    Using $this in static method Kirby\Form\Field::defaults().
  192    Using $this in static method Kirby\Form\Field::defaults().
  192    Using $this in static method Kirby\Form\Field::defaults().
  196    Using $this in static method Kirby\Form\Field::defaults().
  200    Using $this in static method Kirby\Form\Field::defaults().
  201    Using $this in static method Kirby\Form\Field::defaults().
  204    Using $this in static method Kirby\Form\Field::defaults().
  204    Using $this in static method Kirby\Form\Field::defaults().
  207    Using $this in static method Kirby\Form\Field::defaults().
  208    Using $this in static method Kirby\Form\Field::defaults().
  208    Using $this in static method Kirby\Form\Field::defaults().
  212    Using $this in static method Kirby\Form\Field::defaults().
  213    Using $this in static method Kirby\Form\Field::defaults().
  213    Using $this in static method Kirby\Form\Field::defaults().
 ------ ------------------------------------------------------------

 ------ ----------------------------------------------
  Line   tests/Cms/Emails/fixtures/emails/contact.php
 ------ ----------------------------------------------
  1      Undefined variable: $name
 ------ ----------------------------------------------
```

Do you know how to fix those?

szepeviktor:
> Do you know how to fix those?

In descending order.

1. Contribute to @phpstan
1. Prefer 100% testable code
1. Add `/** @var ... */` comments (currently not working)
1. Populate ignores with a **comment** :)

lukasbestle:
Well, I'm afraid that doesn't help much.

We should probably first merge this first set of fixes as it's better than nothing. But I hope you understand that we can't upgrade the level or even run PHPStan in our CI suite as long as there are false-positives we cannot fix (creating a thousand-line error ignore list doesn't count). :)

szepeviktor:
e.g. `tests/Cms/Emails/fixtures/emails/contact.php` could simply be excluded as a file - globals/raw PHP templates (`$name`) are not testable

and `views/panel.php`

temporary we may add

```yaml
        - '#^Access to an undefined property Kirby\\#'
```

minus 100 errors

szepeviktor:
You may signify directories/ by appending a slash.

distantnative:
Is this correct? I still get 740 errors running `composer run-script analyze`.

szepeviktor:
Yes, the count is 740. With 10 minutes of highly efficient work (mainly defining properties) you may reduce it to under 100.

szepeviktor:
BTW `Kirby\Toolkit\Component` does need a phpstan-kirby extension as it is full of magic methods.

distantnative:
> does need a phpstan-kirby extension as it is full of magic methods 

That sounds complicated. Wondering if phpstan will really benefit us enough there to balance of the time investment to write such an extension.

lukasbestle:
> With 10 minutes of highly efficient work (mainly defining properties) you may reduce it to under 100.

@szepeviktor If it's 10 minutes for you, we would be grateful for a PR against the `feature/phpstan` branch (the branch of this PR).

> Wondering if phpstan will really benefit us enough there to balance of the time investment to write such an extension.

I agree. From what I saw until now, PHPStan currently can't deal with several patterns we use, so the amount of false-positives is really too high. And that's just with error level 0, the lowest available.

It does catch some actual mistakes, but having to review the list of errors manually or add them to the ignore list (which isn't great as that will also prevent those errors from showing up when they are correct) just isn't going to work for us.

distantnative:
I agree, if we can't find an easy way to remove the false positives, we shouldn't include this in our CI. More a helper tool to once go through the list and see where it found actual bugs - as a one-time investment.

szepeviktor:
> @szepeviktor If it's 10 minutes for you, we would be grateful for a PR

Okay, I contribute that 10 minutes.

Really you have to pay phpstan some time on Levels 0-2 to benefit from Levels 4-6.

lukasbestle:
> Really you have to pay phpstan some time on Levels 0-2 to benefit from Levels 4-6.

That's true of course and we are happy to invest that time. But unless we get the false-positives out, we can't add PHPStan to CI (as @distantnative explained), so it all doesn't make too much sense for us to be honest.

szepeviktor:
Here you go #2173 - I am totally tired :)

szepeviktor:
To get phpstan fast up and running I suggest you a work-around: test only `/src` on Level 2.

e.g. while waiting for phpstan-kirby, `src/Toolkit/V.php` could be shimmed in class comment block with

```php
 /**
  * @method static bool email($value)
  */
```

Try
```bash
sed -n -e "s#^\\s*'\\(\\S\\+\\)' => function \\((.*)\\): \\(\\S\\+\\).*# * @method static \\3 \\1\\2#p" src/Toolkit/V.php
sed -n -e "s#^\\s*'\\(\\S\\+\\)' => function \\((.*)\\).*# * @method static \\1\\2#p" src/Form/Field.php
```

I feel I am not enough for a full-blown CMS' code base :)

lukasbestle:
> I feel I am not enough for a full-blown CMS' code base :)

We aren't either. To be honest, I feel like we should invest our time into changes that actually benefit our users and not into workarounds for a tool that is not working properly.

So I think we should merge this PR with the basic fixes but otherwise wait until PHPStan works properly without hacks.

szepeviktor:
> without hacks

Excuse me: https://github.com/laravel/framework/blob/6.x/src/Illuminate/Support/Facades/App.php#L5-L43

lukasbestle:
What do you want to tell me with that link?

szepeviktor:
A proper code base should declare magic methods

- for IDEs
- for static analysis
- for human readers

e.g. the `V` class

szepeviktor:
And a proper code base must not have anything else in a class file other than `class { ... }`

e.g. not 2 classes, no side effects `V::$validators = [ ... ];`

lukasbestle:
I have now fixed the remaining PHPStan errors we can fix, which leaves us with 62 errors.

szepeviktor:
Very nice :) now we need

- apcu extension
- memcached ext.
- add an ignore
- fix the remaining ~8 errors

```yaml
        - message: '#^Using \$this in static method Kirby\\Form\\Field::defaults\(\)\.$#'
          path: %currentWorkingDirectory%/src/Form/Field.php
```

tests/Toolkit/EscapeTest.php
```diff
@@ -29,10 +29,6 @@ class EscapeTest extends TestCase
 
     protected $htmlAttrSpecialChars = [
         '\''    => '&#x27;',
-        '"'     => '&quot;',
-        '<'     => '&lt;',
-        '>'     => '&gt;',
-        '&'     => '&amp;',
```
bastianallgeier, line 35: Why are those tests removed?

lukasbestle: They are duplicated a few lines below.

bastianallgeier: LOL! I completely missed this :)

bastianallgeier (approved):

tests/Toolkit/EscapeTest.php
distantnative: XD

PR #2167, opened 2019-10-03
Including subpages feature on pages field

afbora:
## Describe the PR

Allows you to include sub-pages with using query or not

I didn't want to separate options as `subpages` and `querySubpages` as @lukasbestle suggested. This could create a bit of confusion for users and code
I wanted to make a simpler and more comprehensive (with query or not) implement with one option as `subpages`
Including subpages default option is enabled and to disable set as `subpages: false`

**Usage**:

```yaml
fields:
    testWithDefault:
        type: pages
        multiple: true
        
    testWithoutSubpages:
        type: pages
        multiple: true
        subpages: false
        
    testWithQuerySubpages:
        type: pages
        multiple: true
        query: site.find("test").children
```

**Note**:
@lukasbestle I know i can't write a very effective unit test. I think a different method needs to be applied for child elements. Therefore, it may be necessary to review the `testPageChildren()` test 😇 

**Additional idea**: 
In addition, you can select pages from specific levels by adding depth option. If approval is given, I can implement this idea. For ex:

```yaml
fields:
    testWithDepth:
        type: pages
        multiple: true
        depth: 2

```

## Related issues

- Closes https://github.com/getkirby/ideas/issues/217

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/fields/mixins/pagepicker.php
+++ b/config/fields/mixins/pagepicker.php
@@ -7,34 +7,38 @@ return [
             $subpages = $params['subpages'] ?? true;
             $model    = $this->model();
             $site     = $this->kirby()->site();
-
+            $root     = $site;
+            
             if (empty($query) === false) {
                 if ($subpages === true) {
+                    $root  = $model->query($query, 'Kirby\Cms\Pages');
+                    
                     if ($parent = $site->find($params['parent'] ?? null)) {
                         $pages = $parent->children();
                     } else {
-                        $pages  = $model->query($query, 'Kirby\Cms\Pages');
+                        $pages  = $root;
                         $parent = $pages->parent();
                     }
                 } else {
-                    $pages = $model->query($query, 'Kirby\Cms\Pages');
+                    $pages  = $model->query($query, 'Kirby\Cms\Pages');
+                    $parent = null;
                 }
             } else {
                 if (!$parent = $site->find($params['parent'] ?? null)) {
-                    $parent = $site;
+                    $parent = $root;
                 }
                 
                 $pages = $parent->children();
             }
             
-            if ($subpages === true) {
+            if ($subpages === true && $parent !== null) {
                 $self  = [
-                    'id'     => (empty($parent->id()) === true || empty($params['parent']) === true) ? null : $parent->id(),
+                    'id'     => (empty($parent->id()) === true || $parent->id() === $root->parent()->id()) ? null : $parent->id(),
                     'parent' => is_a($parent->parent(), 'Kirby\Cms\Page') === true ? $parent->parent()->id() : null,
                     'title'  => $parent->title()->value(),
                 ];
             }
-
+            
             $children = [];
 
             foreach ($pages as $index => $page) {
--- a/config/fields/mixins/pagepicker.php
+++ b/config/fields/mixins/pagepicker.php
@@ -3,42 +3,53 @@
 return [
     'methods' => [
         'pagepicker' => function (array $params = []) {
-            $query    = $params['query'] ?? null;
-            $subpages = $params['subpages'] ?? true;
-            $model    = $this->model();
-            $site     = $this->kirby()->site();
-            $root     = $site;
-            
-            if (empty($query) === false) {
-                if ($subpages === true) {
-                    $root  = $model->query($query, 'Kirby\Cms\Pages');
-                    
-                    if ($parent = $site->find($params['parent'] ?? null)) {
+
+            // default params
+            $params = array_merge([
+                'image'    => [],
+                'info'     => false,
+                'map'      => null,
+                'parent'   => null,
+                'query'    => null,
+                'subpages' => true,
+                'text'     => null
+            ], $params);
+
+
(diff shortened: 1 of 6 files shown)


config/fields/mixins/pagepicker.php
```diff
@@ -3,23 +3,35 @@
...
+            $model    = $this->model();
+            $site     = $this->kirby()->site();
 
-            if ($query) {
-                $pages = $model->query($query, 'Kirby\Cms\Pages');
-                $self  = null;
+            if (empty($query) === false) {
+                if ($subpages === true) {
+                    if ($parent = $site->find($params['parent'] ?? null)) {
+                        $pages = $parent->children();
```
lukasbestle: Why the check for the `parent` option here? Shouldn't the query always be executed if it's given? I'd say this is a breaking change otherwise.

config/fields/mixins/pagepicker.php
```diff
@@ -3,23 +3,35 @@
...
+                    } else {
+                        $pages  = $model->query($query, 'Kirby\Cms\Pages');
+                        $parent = $pages->parent();
+                    }
+                } else {
+                    $pages = $model->query($query, 'Kirby\Cms\Pages');
```
lukasbestle: `$parent` should explicitly be set to `null` here to ensure that the variable is set in all cases of the condition tree.

config/fields/mixins/pagepicker.php
```diff
@@ -3,23 +3,35 @@
...
             } else {
                 if (!$parent = $site->find($params['parent'] ?? null)) {
                     $parent = $site;
                 }
-
+                
                 $pages = $parent->children();
+            }
+            
+            if ($subpages === true) {
```
lukasbestle: This should then check for `$subpages === true && $parent !== null` to make sure that the following code never runs without a valid parent object.

config/fields/mixins/pagepicker.php
```diff
@@ -3,23 +3,35 @@
...
                 $self  = [
-                    'id'     => $parent->id() == '' ? null : $parent->id(),
-                    'title'  => $parent->title()->value(),
+                    'id'     => (empty($parent->id()) === true || empty($params['parent']) === true) ? null : $parent->id(),
```
lukasbestle: Why the check for the `$params['parent']` value?

config/fields/pages.php
```diff
@@ -42,6 +42,13 @@
         'size' => function (string $size = 'auto') {
             return $size;
         },
+        
+        /**
+         * Optional includes subpages of pages
```
lukasbestle: "Optionally include subpages of pages"

lukasbestle (dismissed):

config/fields/mixins/pagepicker.php
afbora: FireShot Capture 006 - Sandbox - Mægazine - localhost

This is for hide back button for root.

config/fields/mixins/pagepicker.php
afbora: In the first query/load, `parent` index is `null`. When navigating to the parent and subpages, the `parent` index is the `id` of the clicked page.

config/fields/mixins/pagepicker.php
```diff
@@ -3,26 +3,42 @@
...
-            $site  = $this->kirby()->site();
-
-            if ($query) {
-                $pages = $model->query($query, 'Kirby\Cms\Pages');
-                $self  = null;
+            $query    = $params['query'] ?? null;
+            $subpages = $params['subpages'] ?? true;
+            $model    = $this->model();
+            $site     = $this->kirby()->site();
+            $root     = $site;
```
afbora: FireShot Capture 006 - Sandbox - Mægazine - localhost

This page is root but showing back button.
Therefore keeping `root` page for show/hide back button.

config/fields/mixins/pagepicker.php
```diff
@@ -3,26 +3,42 @@
...
-
+                
                 $pages = $parent->children();
+            }
+            
+            if ($subpages === true && $parent !== null) {
                 $self  = [
-                    'id'     => $parent->id() == '' ? null : $parent->id(),
-                    'title'  => $parent->title()->value(),
+                    'id'     => (empty($parent->id()) === true || $parent->id() === $root->parent()->id()) ? null : $parent->id(),
```
afbora: This is for show/hide back button.

lukasbestle (commented):
The logic is quite complex, which makes it hard to review if the code works for all edge-cases. I'd say this needs a unit test for each branch in the `if` tree.

Also this is probably something for @bastianallgeier to review as he wrote the original code. It's quite hard for me to follow the logic.

bastianallgeier:
I'm on it. Maybe we can refactor it a bit to be more readable. It has been my spaghetti code for a long time.

bastianallgeier:
I moved the logic for the picker into its own class now. This makes it more future-proof. Especially when we want to work on pagination and search as a next step.

PR #2168, opened 2019-10-03
New Visitor class methods for preferred MIME type

lukasbestle:
This is the first set of changes from #2148. Removing the deprecated features requires these changes, so I will send the other two PRs once this one is merged.

### DIFF
--- a/src/Toolkit/Mime.php
+++ b/src/Toolkit/Mime.php
@@ -6,7 +6,7 @@ use SimpleXMLElement;
 
 /**
  * The `Mime` class provides method
- * for mime type detection or guessing
+ * for MIME type detection or guessing
  * from different criteria like
  * extensions etc.
  *
@@ -19,7 +19,7 @@ use SimpleXMLElement;
 class Mime
 {
     /**
-     * Extension to mime type map
+     * Extension to MIME type map
      *
      * @var array
      */
@@ -112,7 +112,7 @@ class Mime
     ];
 
     /**
-     * Fixes an invalid mime type guess for the given file
+     * Fixes an invalid MIME type guess for the given file
      *
      * @param string $file
      * @param string $mime
@@ -153,7 +153,7 @@ class Mime
     }
 
     /**
-     * Guesses a mime type by extension
+     * Guesses a MIME type by extension
      *
      * @param string $extension
      * @return string|null
@@ -165,7 +165,7 @@ class Mime
     }
 
     /**
-     * Returns the mime type of a file
+     * Returns the MIME type of a file
      *
      * @param string $file
      * @return string|false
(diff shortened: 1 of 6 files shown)


distantnative:
I think it might be fine for this PR, but the removal of lines like

```php
$this->assertTrue($page->hasVisibleChildren());
```

should go into it's own PR to not distract while reading the changes and new implementations.

Also the name of the PR seems off since the main part is around preferred MIME type.

lukasbestle:
@distantnative It's very difficult to structure those PRs properly. The more we split it up, the longer it takes because I can only create the next PR once the previous one has been merged. Removing the tests for example is required for the next PR that removes the deprecated methods, otherwise the tests won't pass.

Regarding the naming: The preferred MIME type feature has only been implemented so that we can migrate from the old `$request->ajax()` method that has been deprecated. So same thing here: If I'd split that up, that would be another PR that would need to be merged before this change could be made. It's already three PRs (two of them being lined up) and I really don't want it to be even more confusing for us all by splitting it up more than it needs to be.

distantnative:
Not sure if I follow your steps correctly:
Why can't we bundle removing 

```php
$this->assertTrue($page->hasVisibleChildren());
```

with removing the actual deprecated methods in one PR? The tests should pass when you remove them in the same PR. And bundling would make it much more sensible to me.

I get that splitting PRs can be annoying for the one working on them - trust me @bastianallgeier has pushed me to it over and over cause I tend to make huge PRs XD But from the perspective of one reviewing a PR it's really confusing to see code changes that don't really belong to the rest of the PR.

lukasbestle:
Oh, right! I actually meant to do that but it looks like I moved the changes to the wrong branch when splitting the PRs. Will fix that now.

src/Toolkit/Mime.php
```diff
@@ -239,7 +240,20 @@ public static function isAccepted(string $mime, string $pattern): bool
...
+     * matches a given type
+     *
+     * @param string $wildcard
+     * @param string $test
+     * @return bool
+     */
+    public static function matchWildcard(string $wildcard, string $test): bool
+    {
+        return fnmatch($wildcard, $test, FNM_PATHNAME) === true;
+    }
```
bastianallgeier: I'd like to restructure this a bit. 

```php
public static function matches(string $mime, string $wildcard): bool
{
    return fnmatch($wildcard, $mime, FNM_PATHNAME) === true;
}
```

In my opinion, it fits better to the way we normally handle arguments in such static methods. i.e. see `Mime::isAccepted` or something like `Str::contains`. We always pass the "haystack" first and then the "needle" second. 

```
Mime::matches('text/html', 'text/*');
```

bastianallgeier (changes requested):

src/Toolkit/Mime.php
bastianallgeier: The rest is great!

bastianallgeier (approved):
I just made the changes to the Mime class. The rest is perfect!

PR #2169, opened 2019-10-03
Upgrade Composer usage

szepeviktor:
## Describe the PR

- Travis images include self-upgrade
- Require two packages at once
- Always validate config file

## Ready?

- [x] The lock file needs synchronization


.travis.yml
```diff
@@ -15,10 +15,9 @@ before_install:
 
 install:
   # Install composer packages, will also trigger dump-autoload
-  - composer self-update
-  - composer require phpunit/phpunit ^7
-  - composer require friendsofphp/php-cs-fixer
-  - composer install --no-interaction --ignore-platform-reqs
+  - composer validate
```
lukasbestle, line 18: Why did you move the `composer validate` command to the `install` step? It's part of the actual test suite, so I think it should run later as a `script`.

I agree on the other changes.

lukasbestle (changes requested):

.travis.yml
szepeviktor: I think and invalid composer config is a road block, so it should be number 1.

lukasbestle: That's right. Our Composer config is rarely completely invalid, but when it is it should fail immediately.

lukasbestle (dismissed):

lukasbestle:
Hm, what I don't like about `composer validate` being part of the install steps is that the output gets collapsed if the step succeeded (= warnings aren't visible). But well, it all has pros and cons.

szepeviktor:
add `--strict` and it will: _Return a non-zero exit code for warnings as well as errors_

lukasbestle:
We don't want that either. As you can see, we currently have warnings and we deliberately want to ignore them. But we want to see when new warnings get output (for example when the Composer team changes anything in the config format) so that we can review the warnings manually.

szepeviktor:
Please clean up all warnings. Benefit is hidden inside each of them, although it looks like difficulty, something you don't want.

lukasbestle:
We can't do that – while we could switch to less specific version constraints for the dependencies (and indeed probably should @bastianallgeier), the other warning about the `version` field in `composer.json` can't be fixed as we rely on that internally.

bastianallgeier (approved):

szepeviktor:
I'm thankful for becoming a Kirby contributor.

PR #2174, opened 2019-10-06
Allow same chars on file rename as on upload #2164

### LINKED ISSUE #2164: Underscores in file names are not sanitize on upload but on renaming
**Describe the bug**  
File with an underscore like "image_02.png" can be uploaded without being sanitized. When renaming them underscores are replaced with a dash.

**Expected behavior**  
The same rules should apply to file names when uploading and renaming. So either "image_02.png" should be changed to "image-02.png" on upload or underscores should be allowed in filenames when renaming them.

**Screenshots**  
image

**Kirby Version**  
3.2.5

distantnative:
## Describe the PR
On upload: https://github.com/getkirby/kirby/blob/master/src/Toolkit/F.php#L652
`a-z0-9` is automatically allowed by the JS slug helper, so we only need to add the rest.

## Related issues
- Fixes #2164

### DIFF
--- a/panel/src/components/Dialogs/FileRenameDialog.vue
+++ b/panel/src/components/Dialogs/FileRenameDialog.vue
@@ -68,7 +68,7 @@ export default {
         });
     },
     sluggify(input) {
-      return this.$helper.slug(input, [this.slugs, this.system.ascii], ".");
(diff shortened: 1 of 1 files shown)


lukasbestle (approved):
👍

bastianallgeier (approved):

PR #2178, opened 2019-10-07
Remove deprecated methods

lukasbestle:
I'm open for discussion which methods we should remove and which ones we should keep for now.

I think that the `$request->ajax()` method should definitely be removed as it's not really useful anyway and not recommended.

We could probably keep the v2 alias methods, but then we should trigger deprecation warnings when they are called. We could then finally remove them in 3.5.0 or something.

### DIFF
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -363,6 +363,8 @@ class File extends ModelWithContent
      */
     public function meta()
     {
+        deprecated('$file->meta() is deprecated, use $file->content() instead. $file->meta() will be removed in Kirby 3.5.0.');
+
         return $this->content();
     }
 
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -122,6 +122,21 @@ function css($url, $options = null): ?string
 }
 
 /**
+ * Triggers a deprecation warning if debug mode is active
+ *
+ * @param string $message
+ * @return bool Whether the warning was triggered
+ */
+function deprecated(string $message): bool
+{
+    if (App::instance()->option('debug') === true) {
+        return trigger_error($message, E_USER_DEPRECATED) === true;
+    }
+
+    return false;
+}
+
+/**
  * Simple object and variable dumper
  * to help with debugging.
  *
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -256,6 +256,8 @@ trait FileActions
      */
     public function rename(string $name, bool $sanitize = true)
     {
+        deprecated('$file->rename() is deprecated, use $file->changeName() instead. $file->rename() will be removed in Kirby 3.5.0.');
+
         return $this->changeName($name, $sanitize);
     }
 
--- a/src/Cms/HasChildren.php
+++ b/src/Cms/HasChildren.php
@@ -175,11 +175,13 @@ trait HasChildren
     }
 
     /**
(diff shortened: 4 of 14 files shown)


bastianallgeier:
I'm really hesitant with removing those. Especially all visible/invisible. They are still used in so many projects and templates and will make the upgrade extremely painful.

lukasbestle:
I agree on that. That‘s why I think it would make sense to add PHP deprecation warnings. Those will only be visible in development and would make it easier to migrate. We can then remove the methods once people had time to migrate.

bastianallgeier:
Great idea! How would you solve that?

lukasbestle:
With:

```php
trigger_error('The xyz method is deprecated, use xyz instead. The deprecated method will be removed in Kirby 3.5.0.', E_USER_DEPRECATED);
```

Would that be alright? I'd then update the PR with this for the deprecated core methods. I'd still remove `$request->ajax()` though for the reasons mentioned in the original post.

bastianallgeier:
@lukasbestle sounds good! Maybe we could create a new `deprecated($message)` helper for this, which automatically checks for an enabled debug mode and triggers the error.

lukasbestle:
I noticed that Whoops also activates for warnings with the current config. That's a bit unfortunate, but maybe not even bad in the end as it will show the devs directly what to fix. It's the debug mode after all. :)

lukasbestle:
@bastianallgeier Ready for review.

src/Cms/File.php
```diff
@@ -357,12 +357,14 @@ public function mediaUrl(): string
     }
 
     /**
-     * @deprecated 3.0.0 Use `File::content()` instead
+     * @deprecated 3.0.0 Use `$file->content()` instead
```
distantnative: Please don't change the notation here (and everywhere else), this will break the automatic linking in the Reference.

lukasbestle: Oh, I didn't know we had that for the `@deprecated` markers. Fixed now. 👍

distantnative: Just for reference where this "magic" happens: https://github.com/getkirby/getkirby.com/blob/76de1d2e9ba132e604bc8b47b3b82cb096190eaa/site/plugins/site/helpers.php#L184-L210

bastianallgeier (approved):

RobertCordes:
Hi. I just needed to use the ajax() method, when I stumbled upon this pull request. Is there any other recommended way to find out if it's an ajax request? I'm using the old implementation until then…

The documentation still has an example with the method by the way ;-)

texnixe:
> The documentation still has an example with the method by the way 
@RobertCordes  The Kirby 3 docs? Where?

afbora:
@texnixe https://getkirby.com/docs/reference/objects/request

lukasbestle:
@RobertCordes Kirby 3.3.0 has a new [`$kirby->visitor()->prefersJson()`](https://getkirby.com/docs/reference/@/classes/http/visitor/prefers-json) method that checks for the `Accept` request header. That works if you need to decide whether you want to return HTML or JSON. If that‘s not your use-case, let me know. :)

@texnixe It‘s also in this recipe: https://getkirby.com/docs/cookbook/templating/generating-json

texnixe:
Ah, ok, then we should remove that. I'll create an issue in the getkirby.com repo.

RobertCordes:
Thanks @lukasbestle, that works for me.

PR #2179, opened 2019-10-08
Allow name phpmailer functions

afbora:
## Describe the PR

This PR allow name on phpmailer functions that `from`, `replyTo`, `to`, `cc`, `bcc`.

The current syntax supported only one layer (just emails).
As for the name option, it changed the definition syntax. 
I think this implement is the correct syntax usage. 
If you're thinking of a better syntax, we can close the PR. 
Therefore, it was opened as a draft because of this is just an idea 👍 

### Syntax

**1. Single email**

    'from' => 'no-reply@supercompany.com',

**2. Single email with name**

    'from' => ['no-reply@supercompany.com', 'Super Company'],

**3. Multiple email**

    'to' => [
        'john@doe.com', 
        'mark@otto.com'
    ],

**4. Multiple email with name**

    'to' => [
        ['john@doe.com', 'John Doe'],
        ['mark@otto.com', 'Mark Otto']
    ],

### Usage

````php
$kirby->email([
    'from' => ['no-reply@supercompany.com', 'Super Company'],
    'replyTo' => ['no-reply@supercompany.com', 'Super Company'],
    'to' => [
        ['john@doe.com', 'John Doe'],
        ['mark@otto.com', 'Mark Otto']
    ],
    'cc' => [
        ['anotherone@gmail.com', 'Another One']
    ],
    'bcc' => [
        ['secret@foo.com', 'Secret Foo'],
        ['secret@bar.com', 'Secret Bar']
    ],
    'subject' => 'Welcome!',
    'body'=> 'It\'s great to have you with us',
]);
````

## Related issues

- Closes https://github.com/getkirby/ideas/issues/42

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/Email.php
+++ b/src/Cms/Email.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Cms;
 
+use Kirby\Exception\InvalidArgumentException;
 use Kirby\Exception\NotFoundException;
 
 /**
@@ -22,14 +23,6 @@ class Email
     protected $preset;
     protected $props;
 
-    protected static $transform = [
-        'from'        => 'user',
-        'replyTo'     => 'user',
-        'to'          => 'user',
-        'cc'          => 'user',
-        'bcc'         => 'user',
-        'attachments' => 'file'
-    ];
 
     public function __construct($preset = [], array $props = [])
     {
@@ -45,9 +38,12 @@ class Email
         }
 
         // transform model objects to values
-        foreach (static::$transform as $prop => $model) {
-            $this->transformProp($prop, $model);
-        }
+        $this->transformUserSingle('from', 'fromName');
+        $this->transformUserSingle('replyTo', 'replyToName');
+        $this->transformUserMultiple('to');
+        $this->transformUserMultiple('cc');
+        $this->transformUserMultiple('bcc');
+        $this->transformFile('attachments');
 
         // load template for body text
         $this->template();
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by lukasbestle** [comment]
To be honest, I think that the array syntax is a bit clunky and not easy to understand.

I'd prefer the standard email syntax of `Name <email@domain.com>`. This format could be contained in a simple string like before.
PHPMailer provides a [`parseAddresses()` method](http://phpmailer.github.io/PHPMailer/classes/PHPMailer.PHPMailer.PHPMailer.html#method_parseAddresses) we could probably use for this.

**Review by lukasbestle — COMMENTED** [review]
To be honest, I think that the array syntax is a bit clunky and not easy to understand.

I'd prefer the standard email syntax of `Name <email@domain.com>`. This format could be contained in a simple string like before.
PHPMailer provides a [`parseAddresses()` method](http://phpmailer.github.io/PHPMailer/classes/PHPMailer.PHPMailer.PHPMailer.html#method_parseAddresses) we could probably use for this. We could at the same type support multiple email addresses in the same string with this:

```php
$kirby->email([
    'from' => 'Super Company <no-reply@supercompany.com>',
    'replyTo' => 'Super Company <no-reply@supercompany.com>',
    'to' => 'John Doe <john@doe.com>, Mark Otto <mark@otto.com>'
    ...
]);
```

Disadvantage: The dev needs to ensure that the format passed to the `email()` method is valid. Could be an issue if dynamic addresses are used but not properly escaped.

---

Now that I think about it: Probably not a good idea. Maybe we can find another syntax that is clean but also secure.

lukasbestle:
What about an associative array? Each email address can only be exactly one time in each list of email addresses, so it won't collide with anything:

```php
$kirby->email([
    'from' => [
        'no-reply@supercompany.com' => 'Super Company'
    ],
    'replyTo' => [
        'no-reply@supercompany.com' => 'Super Company'
    ],
    'to' => [
        'john@doe.com'  => 'John Doe',
        'mark@otto.com' => 'Mark Otto'
    ],
    'cc' => [
        'anotherone@gmail.com' => 'Another One'
    ],
    'bcc' => [
        'secret@foo.com', // without name
        'secret@bar.com' => 'Secret Bar'
    ],
    'subject' => 'Welcome!',
    'body'=> 'It\'s great to have you with us',
]);
```

It's also backwards-compatible as array items with a numeric key (= no manual key set) would automatically be converted (`'secret@foo.com'` would become `'secret@foo.com' => null`). Simple strings would also be converted to an array with one item without name.

afbora:
@lukasbestle Actually, I should have mentioned it in my first message. My first attempt was already with associative array that exactly like your sample. But the associative array is transformed to a numeric array after following line. So I tried a different syntax and you're right that it's not readable.

https://github.com/getkirby/kirby/blob/develop/src/Cms/Email.php#L48-L50

In code document says on line 141.

> value is an array or collection, call transform on each item

**Input**: 

    array(1) {
      ["no-reply@supercompany.com"]=>
      string(13) "Super Company"
    }


**Output after transform**:

    array(1) {
        [0]=>
        string(13) "Super Company"
    }

afbora:
> Disadvantage: The dev needs to ensure that the format passed to the `email()` method is valid. Could be an issue if dynamic addresses are used but not properly escaped.

Hmm, how about we add a new email validation that working with `V::email()` something like that?

````php
'emailParse' => function ($value): bool {
    $addresses  = imap_rfc822_parse_adrlist($value, '');
    
    if (is_array($addresses) === false || count($addresses) < 1) {
        return false;
    }
    
    foreach ($addresses as $address) {
        if ('.SYNTAX-ERROR.' === $address->host || empty($address->host) === true) {
            return false;
        }
        
        $email = $address->mailbox . '@' . $address->host;
        
        if (V::email($email) === false) {
            return false;
        }
    }

    return true;
},
````

**Usage**

    V::emailParse('Super Company <no-reply@supercompany.com>');
    V::emailParse('no-reply@supercompany.com');

Of course, it may be a more descriptive method name 🙂

**Comment by bastianallgeier** [comment]
@afbora I like that! What about `V::emailRecipient()`?

**Comment by afbora** [comment]
@bastianallgeier I think it's a much better name 👍 What do you think about new validation? @lukasbestle

**Comment by lukasbestle** [comment]
The validator is definitely great (no matter if we decide to use it for this feature), feel free to send a PR for that. I like the name Bastian proposed, but maybe it should be `emailRecipients` as it also validates if multiple addresses are passed in the input string.

We could also have validators `emailRecipientsMin` and `emailRecipientsMax` that would check if the number of recipients in the input matches the expected number. The actual logic for parsing the recipients would need to be moved to the `Toolkit\Str` class then to avoid code duplication.

---

Regarding this PR: I'm still not sure if it's a good idea to use that format for the `$kirby->email()` method. Let's say the dev uses the following code:

```php
$kirby->email([
  'to' => $name . ' <' . $email . '>'
  ...
]);
```

An attacker now enters `Test1 <test@example.com>, Test2` into the `$name` field. Boom, now you have an "email address injection". We can't catch that with validation alone.

So to avoid security issues, I think the safest route would be the associative array. As you wrote, that would require some changes to the `Email` class to support that structure, but I'd say it's worth it.

bastianallgeier:
@lukasbestle ugh, you are right! I didn't consider that. I'm all in for the associative array then. 

But this means that we still need to sanitize/validate the names, right? Such an injection could still happen with the name being manipulated.

afbora:
We have already have a sanitize method for HTTP requests. In fact, it would also be nice to have a method that only sanitizes for strings, something like that: `Str::sanitize($value)`

https://github.com/getkirby/kirby/blob/develop/src/Http/Server.php#L93-L112

afbora:
I think the current PR will not work for us because of many incompatibility and security issues. 
I think the updates should be implement by @lukasbestle better. 
We can close the PR if you want.

**PR summary**:

1. Use associative array to support names `[email => name]`
2. Refactor `Cms\Email` class to support structure that associative array 
3. Add new `emailRecipients` validation
4. Add new `Str::sanitize()` helper method to use on prevent injection names

lukasbestle:
> @bastianallgeier But this means that we still need to sanitize/validate the names, right? Such an injection could still happen with the name being manipulated.

PHPMailer already handles the escaping for us. I tried the following attack:

```php
$mailer->addAddress($to, 'Test1" <test@example.com>, "Test2');
```

It got escaped to this email header:

```
To: "Test1\" <test@example.com>, \"Test2" <actual@recipient.com>
```

So everything is fine. :)

> @afbora We have already have a sanitize method for HTTP requests. In fact, it would also be nice to have a method that only sanitizes for strings, something like that: Str::sanitize($value)

What would that method do exactly? It'd need to sanitize for a specific format (like HTTP headers in the other example), just "sanitizing strings" isn't going to work.

---

Let's keep this PR open, I will reuse the PR branch for a new implementation. I will have time for that this weekend. :)

afbora:
> What would that method do exactly? It'd need to sanitize for a specific format (like HTTP headers in the other example), just "sanitizing strings" isn't going to work.

I meant a filter like that, but of course it's much simpler.

https://docs.phalcon.io/3.4/en/filter

> Let's keep this PR open, I will reuse the PR branch for a new implementation. I will have time for that this weekend. :)

Thank you for taking over the PRs I've been working on 🙏 I'm giving you intensity 😅

bastianallgeier:
@lukasbestle if PHPMailer already takes care of that, we don't really need the associative array, right? we can then pass a simple array of email address and full recipients and validate those first. Or am I missing something?

lukasbestle:
@bastianallgeier PHPMailer takes name and email address as two separate params, so we’d need to split the input into its parts in our code. That would be pretty error-prone. It’s better to keep the two parts separate all across the chain.

bnomei:
german umlauts in from name might need some special treatment: https://stackoverflow.com/a/7267426

 `$headers.= "From: =?utf-8?b?".base64_encode($from_name)."?= <".$from_a.">\r\n";`

Edited: this alone does not solve the issue. i also tried setting the encoding from 8bit to base64 but no luck. https://stackoverflow.com/a/2493009

lukasbestle:
@bnomei Could you please open a separate issue for that with more details (example code to reproduce the issue with the `email()` method etc.)? Thanks!

distantnative:
@afbora @lukasbestle 
Any chance this could be ready for 3.3.2 - or shall we rather aim for 3.4?

afbora:
@distantnative @lukasbestle has knowledge now and he will take care of it :innocent:

lukasbestle:
Yep, I'm still aiming for 3.3.2. :)

lukasbestle:
The `to`, `cc` and `bcc` props now accept key-value arrays with `'email' => 'name'` while being backwards-compatible to simple arrays with just email addresses or mixed arrays. Simple strings with just an email are still supported as well. If a user object or an array/a collection of user objects is passed, the `email` and `name` are both automatically extracted from the object(s).

The `from` and `replyTo` props work a little bit differently as they only support one email address, not an array: The email address can be passed to those props like before and there are new props `fromName` and `replyToName` for the names. If a user object is passed to `from` or `replyTo`, the names are populated automatically but can also be overridden with different values.

So all in all pretty flexible I'd say. :)

bastianallgeier (approved):

PR #2182, opened 2019-10-08
Fix/pagination exception

### LINKED ISSUE #2180: [Panel] Empty sections and pagination conflict
**Describe the bug**  
Empty sections throw this error using latest `develop` code:

>  Pagination page 1 is out of bounds, expected 0-0 

**To Reproduce**  
Update to `develop`.

**Expected behavior**  
No error, displaying an empty section.

**Screenshots**  
Bildschirmfoto am 2019-10-08 um 16 54 14

lukasbestle:
## Describe the PR

Fixes the validation Exception that was thrown when a pagination page of 1 was passed even though there are no pages. I've also reworded the error message to be more human-readable as proposed by @bastianallgeier and @nilshoerrmann.

## Related issues

- Fixes #2180

## Ready?

- [x] Passing all unit tests


src/Toolkit/Pagination.php
```diff
@@ -398,11 +398,17 @@ protected function setProperties(array $props)
             $this->page = $this->firstPage();
         }
 
-        // validate page based on all params
+        // allow a page value of 1 even if there are no pages;
+        // otherwise the exception will get thrown for this pretty common case
         $min = $this->firstPage();
         $max = $this->pages();
+        if ($this->page === 1 && $max === 0) {
```
afbora, line 405: > The section "notes" could not be loaded: Pagination page 2 does not exist, expected 0-0

This will give the error if the items in the section are reset by deleting or changing the status of the listed pages while outside the first page. Could something like this be more stable?

    if ($max === 0) {

or 

    if ($this->page > 0 && $max === 0) {

afbora:
> The section "notes" could not be loaded: Pagination page 2 does not exist, expected 1-1

You'll above error when you delete or change status (not listed status on the section) the last remaining page on second page). 

**To reproduce**: 

**Total**: 6 pages
**Limit**: 5 pages

1. Go to second page
2. Delete 6. item
3. See error

bastianallgeier:
@lukasbestle will you take care of fixing @afbora's issue? Otherwise I can have a look at it so we can merge this.

lukasbestle:
@bastianallgeier Yes, it’s still on my list, I will take a look at it this evening.

lukasbestle:
@afbora @bastianallgeier To be honest, I'm not sure if that's the correct solution. The new error message we added was meant to detect pagination page overflow. If there really is overflow because the number of items changed, that should rightly trigger the error.

Especially the second error is exactly the case the error was meant for. Your proposed solution doesn't fix that one as `$max` will be greater than zero.

Regarding the first error: One could argue that the error should be completely ignored if there are no pagination pages at all (like with your proposed solution), but that will mean that the user can enter any pagination page and will get the same response of no results without an error message. That could be an issue for page caching as well as SEO (duplicate content).

So in the end I think that this is a bug in the Panel. If the number of items changes (more specific: gets lower), the Panel needs to check if the current pagination page still exists. Otherwise it should request the next lower page.

@bastianallgeier I think we can merge this PR without further changes for now to fix the error reported by @nilshoerrmann. We can discuss this additional issue afterwards.

afbora:
@lukasbestle Yes, you're right, it's a panel error.
The pagination class should give this error and the panel should display it correctly as you said.

---

On panel side after the PR merged;

After the overflow error, even if you refresh the page continues to error. It's not even enough to logged out. So if the current page is greater than the total page, the panel should ignore this error and display the last page or the first page.

src/Toolkit/Pagination.php
afbora: If the max/total is zero, why do we check the page number? So if the max/total is zero, shouldn't the page number always be zero? 🤔

lukasbestle:
The most reliable solution would be a meta request to the API that just returns the pagination info (limit, number of pages etc.). The Panel would then know which pagination page to request.

Alternatively we could handle this in the API backend by automatically checking if the requested pagination page exists. If not, it would get limited to the min and max bounds (basically exactly like it was before the pagination error was added). Because the API response contains the pagination page that was returned (which would be the correct one, not the requested one!), this is not a huge deal – unlike frontend paginations in templates.

src/Toolkit/Pagination.php
lukasbestle: Yes, it should be. But if we silently fix the page number without throwing an Exception, we‘ll have exactly the same situation as before where an overflow was ignored.

afbora: I'm trying to ignore the error, and you're trying to throw it away 😅

afbora (approved):

bastianallgeier (approved):

PR #2184, opened 2019-10-09
Escape HTML in blueprint options before it gets sent to the API

bastianallgeier:


afbora:
@bastianallgeier Can we no longer use HTML tags? It will also not work properly with Markdown. The PR is not complete as draft, but I can share the Markdown results if you want.

distantnative:
I think quite a few developers are deliberately using HTML tags in these...

bastianallgeier:
We could allow a predefined set of HTML tags in there. But we must control what's ending up in the panel.

lukasbestle:
I don't think we need to escape the info and help texts completely. Input from the blueprints can be trusted and I think we should allow custom HTML there as that can be useful for some applications.

The issue is just that untrusted data from outside the blueprint might get displayed in an info field/section. As far as I can tell, the only such source of info text that doesn't come from blueprint config is render errors from thrown Exceptions.

So I think we should only escape error messages that are displayed in an info field/section. Calling `document.createTextNode(message)` and fetching the escaped text from that should probably work pretty well already without much frontend code.

bastianallgeier:
I changed it quite a bit. The help and text options are no longer stripped and you can still use full html tags in there. But anything that comes from an exception is already escaped in the backend. 

This PR does not only fix the issue with unescaped HTML but also refactores the options in sections and fields a bit. It turns out using Kirbytext in the help text of sections wasn't even possible before and in fields there was some redundant code.

src/Form/Form.php
```diff
@@ -127,6 +120,18 @@ public function errors(): array
         return $this->errors;
     }
 
+    public static function exceptionField(Throwable $exception, array $props = [])
+    {
+        $props = array_merge($props, [
+            'name'  => $props['name'],
```
afbora: It's not important, but do you think this line is necessary? 🤔

bastianallgeier: Hahah, good one! I have no idea what I was thinking here. I just removed it.

afbora (approved):

Issue #1632, opened 2019-02-21
Search field and pagination in pages/files/users field


mcbrwr:
It would be great for larger sites to be able to search items with a search field. Preferably with live search like the global panel searchbox.

<img width="521" alt="screenshot 2019-02-21 at 14 03 54" src="https://user-images.githubusercontent.com/1662038/53170899-8d892580-35e1-11e9-8570-01a2baca28c8.png">

MathiasGmeiner:
@bastianallgeier is this something you could do for the file field as well?

distantnative:
@MathiasGmeiner definitely - when it comes to the pages field, it will also come to the files and users fields.

afbora:
Can we also add paging feature with `limit` property?

moritzebeling:
Maybe there could also be the option for the `info` text that we know from the `pages` section. That could add a little more context to just the page name / file name itself.

bastianallgeier:
# Pages Field
- [x] Pagination
- [x] Search 

# Files Field
- [x] Pagination
- [x] Search

# Users Field
- [x] Pagination
- [x] Search

![search](https://user-images.githubusercontent.com/24532/66666603-4ecbdc80-ec51-11e9-9bbe-330e38f93abe.gif)

afbora:
https://github.com/getkirby/kirby/pull/2193 ?

PR #2188, opened 2019-10-10
Add pagination to the pages picker

bastianallgeier:
## Describe the PR

This PR adds pagination handling to the pages picker to make it usable with larger lists of pages

## Related issues

- #1632


panel/src/components/Dialogs/PagesDialog.vue
```diff
@@ -21,42 +21,51 @@
...
-              :disabled="!page.hasChildren"
-              :tooltip="$t('open')"
-              icon="angle-right"
-              @click.stop="go(page)"
-            />
-          </template>
-        </k-list-item>
-      </k-list>
+      <template v-if="models.length">
+        <k-list v-if="models.length">
```
distantnative: If not needed, already in previous line

afbora (commented):
> Pagination page 3 is out of bounds, expected 1-2

I guess we have general `pagination` issue and related #2182 

Close dialog when i stay on third page
Open it again dialog when last single item on last page deleted/change status as draft
See error

How about comparing the current page and the total page when opening the dialog?

afbora:
@bastianallgeier This is related to other PR (merged new pagepicker implement) but when each opening the selection dialog, I noticed that `PagePicker::pagesForQuery()` method works twice. (_on pages() and start() method_) Do we need to cache?

bastianallgeier:
@afbora The dialog needs to reset the page when it is being opened. That's an easy fix!

bastianallgeier:
@afbora I just reset the page when the dialog closes and also cached the result of pagesForQuery.

afbora (approved):
I think the pages field is great now 🎉

PR #2191, opened 2019-10-11
Decimals are now always correctly displayed when leaving the field #1748

### LINKED ISSUE #1748: [Panel] Number field removes decimal values when they are 0.

bastianallgeier:
## Describe the PR

I found a way to display decimals correctly but at the same time still keeping the number input type. 

## Related issues

- Fixes #1748

### DIFF
--- a/panel/src/components/Forms/Input/NumberInput.vue
+++ b/panel/src/components/Forms/Input/NumberInput.vue
@@ -80,21 +80,13 @@ export default {
       return this.step.toString().split(".")[1].length || 0;
     },
     format(value) {
-      // don't convert empty values or the beginning of
-      // a negative number to a Number
-      if (value !== null && value !== "" && value !== "-" && value !== "-0") {
-        value = Number(value);
-      }
-
       const decimals = this.decimals();
 
       if (decimals) {
-        value = parseFloat(value).toFixed(decimals);
-      } else {
-        value = parseInt(value);
+        return parseFloat(value).toFixed(decimals);
       }
 
-      return value;
+      return parseInt(value);
     },
     focus() {
       this.$refs.input.focus();
--- a/panel/src/components/Forms/Input/NumberInput.vue
+++ b/panel/src/components/Forms/Input/NumberInput.vue
@@ -10,9 +10,9 @@
       name,
       placeholder,
       required,
-      step,
-      value
+      step
     }"
+    :value="number"
     class="k-number-input"
     type="number"
     v-on="listeners"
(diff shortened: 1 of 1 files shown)


panel/src/components/Forms/Input/NumberInput.vue
```diff
@@ -69,20 +72,43 @@ export default {
...
+      }
+
+      return this.step.toString().split(".")[1].length || 0;
     },
-    onInput(value) {
+    format(value) {
       // don't convert empty values or the beginning of
       // a negative number to a Number
       if (value !== null && value !== "" && value !== "-" && value !== "-0") {
         value = Number(value);
```
distantnative: I would write this with inmediate return statements.

panel/src/components/Forms/Input/NumberInput.vue
```diff
@@ -69,20 +72,43 @@ export default {
...
       }
 
-      this.$emit("input", value);
+      const decimals = this.decimals();
+
+      if (decimals) {
+        value = parseFloat(value).toFixed(decimals);
```
distantnative: same here in if and else

distantnative: And in fact we can save the else statement and just return the normal one last.

bastianallgeier: I didn't realize that the format method can be a lot simpler when we only use it on change.

panel/src/components/Forms/Input/NumberInput.vue
```diff
@@ -80,21 +80,13 @@ export default {
       return this.step.toString().split(".")[1].length || 0;
     },
     format(value) {
-      // don't convert empty values or the beginning of
```
distantnative: what happens to those string now that start with a minus sign or a zero? od't we have to treat them in a different way? or is that solved too with the new implementaition?

bastianallgeier: They were only treated in the method so far, because the method was executed on input and that would stop them from being entered entirely without this exception. Now the transformation takes place on change and it's no longer a problem to sanitize them with parseInt or parseFloat.

distantnative (approved):

distantnative:
LGTM

PR #2192, opened 2019-10-11
Add pagination to all pickers #1632

bastianallgeier:
## Describe the PR

Pagination is now implemented for all pickers (pages, users, files) and all pickers have new underlying PHP classes to refactor their spaghetti code from the fields definitions.

## Related issues

- #1632

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/FilePicker.php
+++ b/src/Cms/FilePicker.php
@@ -90,11 +90,11 @@ class FilePicker
         // by converting site and page objects to proper
         // pages by returning their children
         if (is_a($files, 'Kirby\Cms\Site') === true) {
-            $files = $files->children();
+            $files = $files->files();
         } elseif (is_a($files, 'Kirby\Cms\Page') === true) {
-            $files = $files->children();
+            $files = $files->files();
         } elseif (is_a($files, 'Kirby\Cms\User') === true) {
-            $files = $files->children();
+            $files = $files->files();
         } elseif (is_a($files, 'Kirby\Cms\Files') === false) {
             throw new InvalidArgumentException('Your query must return a set of files');
         }
--- a/src/Cms/FilePicker.php
+++ b/src/Cms/FilePicker.php
@@ -116,7 +116,7 @@ class FilePicker
      * @param \Kirby\Cms\Files|null $files
      * @return array
      */
-    public function filesToArray($files): array
+    public function filesToArray($files = null): array
     {
         if ($files === null) {
             return [];
--- a/src/Cms/FilePicker.php
+++ b/src/Cms/FilePicker.php
@@ -1,0 +1,175 @@
+<?php
+
+namespace Kirby\Cms;
+
+use Kirby\Exception\InvalidArgumentException;
+
+/**
+ * The FilePicker class helps to
+ * fetch the right files for the API calls
+ * for the file picker component in the panel.
+ *
+ * @package   Kirby Cms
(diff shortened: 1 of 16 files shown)


src/Cms/FilePicker.php
```diff
@@ -0,0 +1,175 @@
...
+        } elseif (is_a($model, 'Kirby\Cms\File') === true) {
+            $query = 'file.siblings';
+        } else {
+            $query = $model::CLASS_ALIAS . '.files';
+        }
+
+        // fetch all files for the picker
+        $files = $model->query($query);
+
+        // help mitigate some typical query usage issues
```
distantnative, line 89: Does this make sense in the context of the FilePicker? Shouldn't it convert to files not children?

src/Cms/FilePicker.php
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * Converts all given files to an associative
+     * array that is already optimized for the
+     * panel picker component.
+     *
+     * @param \Kirby\Cms\Files|null $files
+     * @return array
+     */
+    public function filesToArray($files): array
```
distantnative: `$files=null`?

src/Cms/UserPicker.php
```diff
@@ -0,0 +1,157 @@
...
+     *
+     * @param array $params
+     */
+    public function __construct(array $params = [])
+    {
+        // default params
+        $defaults = [
+            // image settings (ratio, cover, etc.)
+            'image' => [],
+            // query template for the page info field
```
distantnative: `user info field`

src/Cms/UserPicker.php
```diff
@@ -0,0 +1,157 @@
...
+            'info' => false,
+            // number of pages displayed per pagination page
```
distantnative: `number of user`

distantnative:
Added some small fixes and one review comment - other than that great!

src/Cms/FilePicker.php
```diff
@@ -0,0 +1,175 @@
...
+        // pages by returning their children
+        if (is_a($files, 'Kirby\Cms\Site') === true) {
+            $files = $files->children();
+        } elseif (is_a($files, 'Kirby\Cms\Page') === true) {
+            $files = $files->children();
+        } elseif (is_a($files, 'Kirby\Cms\User') === true) {
+            $files = $files->children();
+        } elseif (is_a($files, 'Kirby\Cms\Files') === false) {
+            throw new InvalidArgumentException('Your query must return a set of files');
+        }
```
afbora: Repeated `$files = $files->children();` 🤔 

I think there's a mistake here. If the query is one of these, the picker will list the pages.

distantnative: @afbora I think we both found the same bug in this :)

afbora: You right! When I focused on the bug, i missed your review 👍

distantnative:
But I am confused why our PHPUnit test is failing because of the `escape` helper.

bastianallgeier:
I think the PHPunit escape issue comes from another commit.

distantnative:
@bastianallgeier yes Lukas mentioned that. So only the section @afbora and I marked as weird

src/Cms/FilePicker.php
bastianallgeier: Ok, that was stupid! It took me a while to find what you mean. I was totally blind :)

PR #2196, opened 2019-10-11
Fix opening multiselect via keys #2187

### LINKED ISSUE #2187: Enter key conflicting with open field and select item
Related PR: #2137

distantnative:
## Describe the PR
I think this solves it. Opening works now with enter and space, selecting an item still works with enter and space as well.

## Related issues
- Fixes #2187


afbora (approved):
I love it ❤️ Great works both!

PR #2197, opened 2019-10-11
Respect changeTemplate order #2161

### LINKED ISSUE #2161: order of list items of changeTemplate is ignored
**Describe the bug**  
```
options:
  changeTemplate:
    - b_template
    - a_template
    - c_template
```

The ordering in the template selection dropdown is alphabetical instead of the defined order. 

**To Reproduce**  
1. Add a blueprint
2. Define changeTemplate with at least two different templates in a specific order (not alphabetical)
3. Add a page, set a template
4. Change template and notice, that the order in the blueprint is ignored

**Expected behavior**  
Order of templates should be as defined in the blueprint.

**Screenshots**
Bildschirmfoto 2019-10-01 um 13 48 56

**Kirby Version**  
3.2.5

distantnative:
## Describe the PR
Respects the order from the `changeTemplate` option and prepends the current blueprint.

## Related issues
- Fixes #2161

### DIFF
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -253,14 +253,13 @@ class Page extends ModelWithContent
             $templates = [];
         }
 
-        // add the current template to the array
-        $templates[] = $currentTemplate;
+        // add the current template to the array if it's not already there
+        if (in_array($currentTemplate, $templates) === false) {
+            array_unshift($templates, $currentTemplate);
+        }
 
         // make sure every template is only included once
         $templates = array_unique($templates);
-
-        // sort the templates
-        asort($templates);
 
         foreach ($templates as $template) {
             try {


lukasbestle (commented):
But won't the current template now always be at the top even if the dev has manually sorted the list in the blueprint and the current template was already contained in the list? `array_unique()` will always keep the first occurrence (= the one your implementation adds at the top) and throw out all remaining occurrences.

distantnative:
I missed that. Then let's add it to the bottom: if the developer has listed it, it will be at the position as chosen by the developer. If not, at the very end.

lukasbestle (dismissed):
What we could do if we want to prepend the template is to only prepend it if it's not already in the list:

```php
// add the current template to the array if it's not already there
if (in_array($currentTemplate, $templates) === false) {
    array_unshift($templates, $currentTemplate);
}
```

I think both solutions are fine. The current version also gets my approval.

lukasbestle (approved):

bastianallgeier (approved):

Issue #2113, opened 2019-09-17
[Panel] Textarea shortcuts not working on Windows


afbora:
**Describe the bug**
Textarea shortcuts not working on Windows

**To Reproduce**

1. Go to panel on Windows
2. Edit any page that have textarea field
3. Try to use shortcuts like that: `CTRL + B`, `CTRL + I`
4. See nothing happened

**Expected behavior**
Shortcuts should work as works in MacOS

@texnixe Tested on MacOS and working
https://forum.getkirby.com/t/textarea-shortcuts-not-working-on-windows/15665

**Kirby Version**
3.2.4

**Console output**
No output

**Desktop**
 - Windows 10
 - Chrome, Firefox, Edge (Latest)

distantnative:
I am seeing we are using the Vue key `meta` modifier: https://github.com/getkirby/kirby/blob/master/panel/src/components/Forms/Input/TextareaInput.vue#L37

On a mac that is the comand key, on windows this maps to the windows key.

afbora:
When pressing `Win + L` as  always use in texarea, windows locking 😱

distantnative:
Yea, I am not sure having the windows key involved is such a great idea

afbora:
Please have a look: https://github.com/vuejs/vue/issues/4843

distantnative:
Just read it, sadly no solution.

afbora:
Worked with `CTRL + WIN + B` to bold, not at all useful 😞 

Can't we add a property like [checkCmdOrCtrl()](https://github.com/vuejs/vue/issues/4843#issuecomment-277030811)? 
I don't think users can use `CTRL + WIN + X` because I see it for the first time.

distantnative:
I think we had them first on ctrl as well but that was blocking something else - not sure. @bastianallgeier might remember. Otherwise yea, we could bind both `meta` and `ctrl`.

afbora:
Let's solve it or not, it would be nice to put a link to the bottom of the textarea or panel that opens the relevant or all shortcuts dialog for end users.

bastianallgeier:
Related: https://github.com/getkirby/kirby/issues/1898

bastianallgeier:
Maybe it's better to bind them to CTRL only?

lukasbestle:
I think `Cmd` makes sense on macOS – it's convention and as long as we use the keys that are used in other apps for the purpose, it's probably the best solution. `Ctrl` would be very confusing.

[...]

PR #2198, opened 2019-10-11
Allow ctrl as meta for textarea shortcuts #2113

distantnative:


afbora (approved):
Awesome, thank you so much 👍 

--- 

Do you need to add the CTRL key here? Not that I know, but there is a meta key in the method, so I thought maybe you should add the ctrl key. It's just a detail I've seen when testing.

https://github.com/getkirby/kirby/blob/fix/2113-textarea-shortcuts-windows/panel/src/components/Forms/Input/TextareaInput.vue#L227

distantnative:
@afbora ahh indeed yes I think. Can you add it? Just on my phone now.

PR #2199, opened 2019-10-11
Login: display errors to user #2161

distantnative:
@bastianallgeier I'll leave the final styling to you as usual :D

### DIFF
--- a/panel/src/components/Forms/Login.vue
+++ b/panel/src/components/Forms/Login.vue
@@ -1,13 +1,14 @@
 <template>
   <form :data-invalid="invalid" class="k-login-form" @submit.prevent="login">
     <h1 class="k-offscreen">{{ $t('login') }}</h1>
-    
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):

panel/src/components/Forms/Login.vue
```diff
@@ -78,23 +79,10 @@ export default {
         })
         .catch(response => {
           this.invalid   = true;
-          this.issue     = response.message;
+          this.issue     = this.$t("error.access.login");
```
distantnative, line 82: @bastianallgeier but this prevents displaying the actual error message which was the main concern of the issue, doesnt it?

bastianallgeier: The original error message is still visible in the console in debug mode. In my opinion it should never end up in the alert.

PR #2201, opened 2019-10-12
Improve a11n of status title in ListItem #566

### LINKED ISSUE #566: Buttons: `pointer-events: none` prevents `title` from being shown
The new status-icons and their colors are a real improvement over the Alpha! I tried to hover them on the start page of the plainkit-panel but I don't get a tooltip with their `title` attribute and this is because they are set to `pointer-events: none`. You could prevent this with a wrapper like described [here](https://stackoverflow.com/a/47412524/1381666).

I know this is a minor one but it could be frustrating for new panel-user on their first visit when they try to guess the meaning of these icons.

Additionally the title could be changed from "Public" to "Public (Disabled)" in this case.

distantnative:
## Describe the PR
The title of the status button was not shown when disabled due to `pointer-events: none`. Solved by wrapping with another element. Tbd if ideal because of additional markup or if we decide not to fix.

## Related issues
- Fixes #566

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -23,6 +23,7 @@
 
   "delete": "Delete",
   "dimensions": "Dimensions",
+  "disabled": "Disabled",
   "discard": "Discard",
   "download": "Download",
   "duplicate": "Duplicate",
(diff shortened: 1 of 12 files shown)


bastianallgeier:
@distantnative I refactored the button component to solve this once and for all buttons. Can you have a look and let me know what you think about it?

distantnative:
@bastianallgeier LGTM 

Only you might still have to delete `@/directives/tab.js` if I read the PR correctly here on GitHub.

Issue #2205, opened 2019-10-12
[Panel] Pagination errors


afbora:
### Known Issues

**1. On files section upload**

> The section "files" could not be loaded: Pagination page 0 does not exist, expected 1-1

**To Reproduce**  
1. Go to panel
2. Edit any page
3. Upload to files section that empty

**2. Deleting/change status of item**

> The section "listed" could not be loaded: Pagination page 2 does not exist, expected 1-1

**To Reproduce**  
1. Go to panel
2. Delete or change status last remaing item on pages section of last page (pagination)
3. See error

**3. On pickers**

Will be fixed with current PRs.

**Kirby Version**  
Develop branch of 3.3.0

lukasbestle:
These are Panel issues, see #2182.

afbora:
@distantnative Please note that this fix is related to the panel ui.

distantnative:
Still needs to be fixed though? 🤨

afbora:
The bug has been fixed on the PHP side by @lukasbestle , but when the error is thrown on the panel, you cannot access the related field again even if refresh the page and logged out.

So when an error is thrown on panel side, we must ignore it or display the error without affecting the access the field/section.

distantnative:
I will try to fix them. For the future, I would prefer it if we only merge such changes when backend and Panel problems are fixed that are caused by these changes. Having them merged without dealing with the Panel makes it more difficult to stay on top of the changes and resulting bugs in my view.

distantnative:
If I read the code correctly the first error is not caused by the Panel but by line 406 of the new backend implementation - so I need your help to understand that @lukasbestle:

```php
        // allow a page value of 1 even if there are no pages;
        // otherwise the exception will get thrown for this pretty common case
        $min = $this->firstPage();
        $max = $this->pages();
        if ($this->page === 1 && $max === 0) {
            $this->page = 0;
        }
```

lukasbestle:
The issue is that the API and Panel have handled the pagination pages in a quite lax way before. It didn't cause any issues as the Pagination class silently ignored pagination pages that were out of bounds. Now that we validate that (which we need to do because of #1887), we suddenly see those bugs in the Panel implementation.

For some possible solutions, see my posts in #2182.

> For the future, I would prefer it if we only merge such changes when backend and Panel problems are fixed that are caused by these changes. Having them merged without dealing with the Panel makes it more difficult to stay on top of the changes and resulting bugs in my view.

[...]

PR #2207, opened 2019-10-12
Fix pagination after section remove action #2205

distantnative:


distantnative:
Closing this in favor of discussed backend changes.

PR #2210, opened 2019-10-14
Introducing the first css variables 

bastianallgeier:
## Describe the PR

It's a pain for plugin developers to work with existing colors, font sizes, etc. because you need to reverse-engineer them all from our styles. CSS vars are perfect to fix this, because they don't need an extra build process step to be used. 

I renamed a few of the vars from our main.scss in a way that feels cleaner. We might want to change the names in our styles as well. But I will wait for your input.

### DIFF
--- a/panel/src/App.vue
+++ b/panel/src/App.vue
@@ -101,6 +101,43 @@ export default {
   box-sizing: border-box;
 }
 
+:root {
+  /** Colors **/
+  --color-backdrop: #{$color-backdrop};
+  --color-background: #{$color-background};
+  --color-border: #{$color-border};
+  --color-focus: #{$color-focus};
+  --color-focus-light: #{$color-focus-on-dark};
+  --color-focus-outline: #{$color-focus-outline};
+  --color-negative: #{$color-negative};
+  --color-negative-light: #{$color-negative-on-dark};
+  --color-negative-outline: #{$color-negative-outline};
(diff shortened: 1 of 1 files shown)


distantnative:
As a developer I’d wonder what’s the difference between backdrop and background. Also I’d expect a notice outline one as well.

But other than that looks good to me

distantnative:
And then we should think about aliasing a few key parts where we use these variables...
What I mean adding a variable e.g. for the top bar background in between. So someone could change that easily with a custom CSS without replacing the black altogether.

bastianallgeier:
@distantnative the backdrop color is a translucent black which is used for the backdrop of our dialogs. 

I agree about the notice outline despite we never use it anywhere :)

Nice idea to have a variable for the top bar. We should be careful though. I don't have any experience with the performance of css variables. I read somewhere that you should use them with caution and not too excessive. That's why I'd rather start small and add more later.

distantnative:
@bastianallgeier i know, I’m just saying that for a third party the difference between backdrop and background isn’t very obvious :D

bastianallgeier:
Yep, that's true. But we need to document the colors anyway.

distantnative (approved):

afbora:
@bastianallgeier https://github.com/getkirby/ideas/issues/87

PR #2211, opened 2019-10-14
Add conditional clauses on collections

afbora:
## Describe the PR

Added `when()` method that @distantnative solution to write less and more legible code when filtering collections.

### Usage

**Simple**

````php
$page->children()
    ->when($sortBy, function ($sortBy) {
        return $this->sortBy($sortBy);
    });
````

**With fallback**

````php
$page->children()
    ->when($sortBy, function ($sortBy) {
        return $this->sortBy($sortBy);
    }, function () {
        return $this->sortBy('title');
    });
````

## Related idea

- Closes https://github.com/getkirby/ideas/issues/349

## Ready?

- [x] Passing all unit tests

### CODE REVIEW

**Comment by distantnative** [comment]
@afbora Your example are just the other way around ;) (simple vs. with fallback)

I still like the implementation (obviously :P) - wondering if the name is 100% spot on.

**Comment by afbora** [comment]
@distantnative Ahh that's correct, i missed it 🙂 Do you have a name in mind?

File: src/Toolkit/Collection.php [hunk 1]
```diff
@@ -1055,6 +1055,30 @@ public function values(): array
...
+     *
+     * @param mixed $condition
+     * @param Closure $callback
+     * @param Closure $fallback
+     * @return Collection
+     */
+    public function when($condition, Closure $callback, Closure $fallback = null)
+    {
+        if ($condition) {
+            return $callback->call($this, $condition) ?: $this;
```
distantnative: I think we usually write `??` instead - just for consistency.

src/Toolkit/Collection.php
```diff
@@ -1055,6 +1055,30 @@ public function values(): array
...
+    /**
+     * The when method only executes the given Closure when the first parameter
+     * is true. If the first parameter is false, the Closure will not be executed.
+     * You may pass another Closure as the third parameter to the when method.
+     * This Closure will execute if the first parameter evaluates as false
+     *
+     * @param mixed $condition
+     * @param Closure $callback
+     * @param Closure $fallback
+     * @return Collection
```
lukasbestle: This should be `@return mixed` as the return value depends on the return value of the Closure.

src/Toolkit/Collection.php
```diff
@@ -1055,6 +1055,30 @@ public function values(): array
...
+     */
+    public function when($condition, Closure $callback, Closure $fallback = null)
+    {
+        if ($condition) {
+            return $callback->call($this, $condition) ?? $this;
```
lukasbestle: I'm not sure if we need the `?? $this` here. If the Closure decides that it wants to return `null`, that's what I'd expect to be the result. Or is there any use-case I'm missing @distantnative?

Same below.

afbora: In fact this is for without return but need `Callable` instead of `Closure`?

````php
->when($sortBy, function ($sortBy) {
    $this->sortBy($sortBy);
});
````

src/Toolkit/Collection.php
afbora: You right, any method in the collection can be used 👍

lukasbestle (approved):
I've combined the tests into one method and added tests for a few more edge-cases. Looks great now!

bastianallgeier (approved):

PR #2215, opened 2019-10-15
Fix csp issues by introducing a new $kirby->nonce() method #1527

### LINKED ISSUE #1527: [Panel] CSP errors
**Describe the bug**
The panel requires CSP with inline styles and inline eval. this would require a rule to allow inline js eval for the complete site just to make the panel work. please find a more secure way.

the following examples are just for the panel login page. there might be more issues past the login.

> Content Security Policy: The page’s settings blocked the loading of a resource at inline (“style-src”).
```
<body>
<svg aria-hidden="true" style="position:absolute;width:0;height:0" xmlns="http://www.w3.org/2000/svg" overflow="hidden">
```
solution: just add a class instead, right?

> Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”).
```
<script>window.panel = {"url":"https://DOMAIN/panel","site":"https://DOMAIN","api":"https://DOMAIN/api","csrf":"ade9273632029b912a42a7aa9f54b85d7626a24ddd8b97177d05a7eb94ed3116","translation":"en","debug":true,"search":{"limit":10}}</script>
```
solution: register a nonce using http header and add a nonce for the script element. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_inline_script

**To Reproduce**
setup some kind of secure CSPs (aka not inline, not eval!) a quick test could be done with my plugin https://github.com/bnomei/kirby3-security-headers
please note that i am well aware that in almost all cases the default CSP my plugin provide must be extended. i just want to point out the panel could be made more security friendly in not using inline styles and eval.

**Expected behavior**
I expeced the panel to not use inline styles and eval because its unsave to do so. I understand kirby should not define csp for the fronend itself, but it should at least make the panel save.

**Kirby Version**
3.x

bastianallgeier:
## Describe the PR

The Kirby class now has a new `$kirby->nonce()` which returns the nonce used for all external and inline scripts + svg + stylesheets). You can use this to set your own CSP and the panel will still work

## Related issues

- Fixes #1527

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -57,6 +57,7 @@ class App
     protected $languages;
     protected $locks;
     protected $multilang;
+    protected $nonce;
     protected $options;
     protected $path;
     protected $request;
@@ -775,6 +776,17 @@ class App
         }
 
         return $this->multilang = $this->languages()->count() !== 0;
+    }
+
+    /**
+     * Returns the nonce, which is used
+     * in the panel for inline scripts
+     *
+     * @return string
+     */
+    public function nonce(): string
+    {
+        return $this->nonce = $this->nonce ?? base64_encode(random_bytes(20));
     }
 
     /**
--- a/src/Cms/Panel.php
+++ b/src/Cms/Panel.php
@@ -107,6 +107,7 @@ class Panel
             'pluginCss' => $plugins->url('css'),
             'pluginJs'  => $plugins->url('js'),
             'panelUrl'  => $uri->path()->toString(true) . '/',
+            'nonce'     => $kirby->nonce(),
             'options'   => [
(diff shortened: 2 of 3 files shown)


distantnative:
@bastianallgeier don't we need to add it to our headers/CSP for it to have any effect?

bastianallgeier:
I thought the last decision was to not set any CSP but implement the nonce in case you want to roll your own CSP.

distantnative:
Understood. And a bit embarrassed cause in the link issue my post is the one summarizing what to do as our collective decision exactly as you implemented it XD

distantnative (approved):

PR #2216, opened 2019-10-15
Improve collection unset

### LINKED ISSUE #1232: Programmatically deleting drafts in hook throws error
**Describe the bug**
Trying to programmatically delete subpage `drafts` within a `page.changeStatus:after` hook, throws the following error: `SyntaxError: Unexpected token T in JSON at position 0`

3/5 pages get deleted, 2/5 (identical) don’t.

Here is the `config.php`

```
'page.changeStatus:after' => function ($newPage, $oldPage) {
            if($newPage->intendedTemplate() == 'project' && $newPage->status() == 'unlisted') {
                foreach($newPage->drafts() as $phase) {
                    try {
                        $phase->delete();
                        echo 'The page has been deleted';
                    } catch(Exception $e) {
                        echo $e->getMessage();
                    }
                }   
            }
        }
```

**To Reproduce**
1. Add subpages to parent page
2. Change status from `listed` to `unlisted` for parent page through the panel
3. See error

**Expected behavior**
All pages should be deleted.

**Kirby Version**
3.0.0-beta-6.22

 - OS: OSX
 - Browser Chrome
 - Version 70

afbora:
## Describe the PR

Let me introduce you to the method I saw for the first time: [ArrayIterator::offsetUnset](https://www.php.net/manual/en/arrayiterator.offsetunset.php) 🎉 

It is doing exactly what we need. Very simple but effective. 
I solved it so simple, and i tested it over and over again to make sure 😅 

## Related issues

- Fixes #1232 

## Ready?

- [ ] Passing all unit tests

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -6,6 +6,7 @@ use Kirby\Cms\Filename;
 use Kirby\Cms\FileVersion;
 use Kirby\Cms\Template;
 use Kirby\Data\Data;
+use Kirby\Http\Server;
 use Kirby\Image\Darkroom;
 use Kirby\Text\Markdown;
 use Kirby\Text\SmartyPants;
@@ -24,6 +25,30 @@ return [
      */
     'css' => function (App $kirby, string $url, $options = null): string {
         return $url;
+    },
+
+
+    /**
+     * Object and variable dumper
+     * to help with debugging.
+     *
+     * @param \Kirby\Cms\App $kirby Kirby instance
+     * @param mixed $variable
+     * @param bool $echo
+     * @return string
+     */
+    'dump' => function (App $kirby, $variable, bool $echo = true) {
+        if (Server::cli() === true) {
+            $output = print_r($variable, true) . PHP_EOL;
+        } else {
+            $output = '<pre>' . print_r($variable, true) . '</pre>';
+        }
+
+        if ($echo === true) {
+            echo $output;
+        }
+
+        return $output;
     },
 
     /**
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -5,7 +5,6 @@ use Kirby\Cms\Asset;
 use Kirby\Cms\Html;
 use Kirby\Cms\Response;
 use Kirby\Cms\Url;
-use Kirby\Http\Server;
 use Kirby\Toolkit\Escape;
 use Kirby\Toolkit\F;
 use Kirby\Toolkit\I18n;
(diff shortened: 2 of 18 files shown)


distantnative:
Mindblown if that really is the solution!
I think we should add unit tests for the reported bug and see if it works.

bastianallgeier:
OMG, it would be sooooo cool to finally fix this. @afbora you are my hero at the moment! I thought this is unsolvable tbh.

afbora:
😊 Thank you so much @bastianallgeier
I found it while researching with `ArrayIterator`
Please add 3 points there for me 🤣 

I couldn't write unit tests because I wasn't good at  btw 🙏

bastianallgeier:
@afbora I will add a unit test now

bastianallgeier:
@afbora unfortunately it doesn't work for me :(

Consider the following test case in PHPUnit in the Kirby\Toolkit\CollectionTest class: 

```php
    public function testRemoveMultiple()
    {

        $collection = new Collection();

        $collection->set('a', 'A');
        $collection->set('b', 'B');
        $collection->set('c', 'C');

        $this->assertCount(3, $collection);

        foreach ($collection as $key => $item) {
            $collection->remove($key);
        }

        $this->assertCount(0, $collection);
    }
```

This fails because the collection still contains 1 item after the loop when it should actually be empty afterwards. Do you have any idea? Can you replicate this? It seems to be a bug in PHP: https://stackoverflow.com/questions/38527123/php-always-losing-item-1-in-arrayiterator-with-using-arrayiteratoroffsetunse

afbora:
@bastianallgeier Actually, I've done so many tries about page deletion, I was wrong sorry..
But there are working examples that i know and I'll research them 👀 

Now I'm a bad hero 😢

afbora:
@bastianallgeier Now finally fixed the issue 🎉

afbora:
CS Fixer gives error on different classes after the test (100%), I couldn't fix it btw

bastianallgeier:
Holy shit! It really works 😳I have to google now what's the difference between Iterator and IteratorAggregate :)

bastianallgeier (approved):

bastianallgeier:
It's a bit of black magic to me because I don't really understand the IteratorAggregate concept, but all our tests pass fine and there don't seem to be any issues with the switch. I'm really excited about this!

afbora:
> I have to google now what's the difference between Iterator and IteratorAggregate :)

To be honest, I didn't know either. 
I researched the working samples (especially Laravel) and carried them to our system. 
It was a great pleasure to see it work ☺️

PR #2217, opened 2019-10-15
Prevent $pages->add() from silently breaking when adding invalid objects #1890

### LINKED ISSUE #1890: Prevent $pages->add() from silently breaking when adding invalid entries
**Describe the bug**
I want to create a pages collection containing ModelWithContent entries. However, if I add the `$site`, I run into problems, probably because it doesn't have an ID.

I know I could use `Kirby\Cms\Collection`, but I need the `index()` method which is only present in `Kirby\Cms\Pages`.

**To Reproduce**
To reproduce the behavior:

```php
$models = new Kirby\Cms\Pages();
$models->add(site());
```

**Expected behavior**
The collection should contain the Site.

**Kirby Version**
3.2.0

**Additional context**
I get that this is kind of an edge case, but Kirby is all about flexibility. Since both Site and Page extend ModelWithContent, it is expected that you can create a group of ModelWithContent that can contain both of these child classes.

My use case is for a plugin - the user can provide a Pages object and anything inside is exported. He might want to include the Site data there or he might not.

**Note:** It works OK if I use `prepend()`/`append()` instead of `add()`. Judging by the source code, it may be because Site doesn't have an ID. If that's the problem, then my proposal is to give it an ID of `<site>` or something similar.

bastianallgeier:
## Related issues

- Fixes #1890

### DIFF
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -2,6 +2,8 @@
 
 namespace Kirby\Cms;
 
+use Kirby\Exception\InvalidArgumentException;
+
 /**
  * The `$pages` object refers to a
  * collection of pages. The pages in this
@@ -55,6 +57,10 @@ class Pages extends Collection
         // add a page object
         } elseif (is_a($object, 'Kirby\Cms\Page') === true) {
             $this->__set($object->id(), $object);
+
+        // give a useful error message on invalid objects
+        } elseif (is_object($object) === true) {
+            throw new InvalidArgumentException('You must pass a Page object to the Pages collection');
         }
 
         return $this;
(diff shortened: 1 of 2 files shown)


distantnative:
What about arrays as input?

bastianallgeier:
@distantnative they weren't handled so far either.

distantnative:
@bastianallgeier but shouldn’t we throw an exception then same as non-Page objects?

bastianallgeier:
@distantnative you are right. I just fixed it. Now, only false, null and true get ignored. The rest throws an exception.

distantnative (approved):

PR #2220, opened 2019-10-15
Moving pages

distantnative:
WIP

Would be great to tap your mind, @bastianallgeier 

https://github.com/getkirby/ideas/issues/134

### DIFF
--- a/config/api/routes/pages.php
+++ b/config/api/routes/pages.php
@@ -71,6 +71,28 @@ return [
         }
     ],
     [
+        'pattern' => 'pages/(:any)/move',
+        'method'  => 'GET',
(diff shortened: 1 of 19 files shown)


bastianallgeier:
I will have a look at it later today! Thanks for starting this.

distantnative:
I thought I'd just get your opinion, @bastianallgeier XD But looks like I triggered you to jump right in :D

bastianallgeier:
When moving pages around the store somehow messed up and the ids contain an old id which does no longer have a matching model.

bastianallgeier:
I'm currently working on an idea for the modal. I think I need a few more hours this morning and then we could have something that works.

PR #2221, opened 2019-10-15
Optimize Travis CI

afbora:
## Describe the PR

Used PCOV driver (CodeCoverage compatible driver for PHP)

**Reference**:
https://kizu514.com/blog/pcov-is-better-than-phpdbg-and-xdebug-for-code-coverage/

**Driver Repositories**

- https://github.com/krakjoe/pcov
- https://github.com/krakjoe/pcov-clobber

With @lukasbestle improvements and new driver are now tests completed in approximately 2:00 - 2:15 minutes. Fixed PHP 7.1 tests btw.

This PR is meant to give you an idea therefore opened as draft.

Related Issue: https://github.com/sebastianbergmann/php-code-coverage/issues/702

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,14 +11,18 @@ php:
   - 7.3
 
 before_install:
-  - echo yes | pecl install -f apcu
-  - echo yes | pecl install -f memcached
-  - phpenv config-rm xdebug.ini || true
+  # set up PHP config
+  - phpenv config-rm xdebug.ini || true # code coverage is generated with PCOV
   - phpenv config-add tests/php.ini
 
-install:
   # ensure that the Composer config is valid
   - composer validate
+
+install:
+  # install PHP extensions
+  - echo yes | pecl install -f apcu
+  - echo yes | pecl install -f memcached
+  - pecl install pcov
 
   # install Composer packages, will also trigger dump-autoload
   - composer require --no-interaction --ignore-platform-reqs "phpunit/phpunit:^7" "friendsofphp/php-cs-fixer" "pcov/clobber"
@@ -30,19 +34,17 @@ install:
   - php php-coveralls.phar --version
 
 before_script:
-  - pecl install pcov
   - mkdir -p build/logs
   - ls -al
 
 script:
-  - php -dpcov.enabled=1 -dpcov.directory=. -dpcov.exclude="~vendor~" ./vendor/bin/phpunit --coverage-clover build/logs/clover.xml
+  - ./vendor/bin/phpunit --coverage-clover build/logs/clover.xml
   - PHP_CS_FIXER_IGNORE_ENV=1 ./vendor/bin/php-cs-fixer fix --config .php_cs --verbose --diff --dry-run
 
 after_success:
   - travis_retry php php-coveralls.phar -v
 
 cache:
-  apt: true
   directories:
     - $HOME/.composer/cache
     - $HOME/.php-cs-fixer
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ before_script:
   - ls -al
 
 script:
-  - ./vendor/bin/pcov clobber; ./vendor/bin/phpunit --coverage-clover build/logs/clover.xml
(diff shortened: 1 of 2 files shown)


lukasbestle:
Thanks a lot for the PR. I still think that it’s an issue unrelated to PHPUnit/Travis, but those optimizations are good nonetheless.

Unfortunately, PCOV doesn‘t mark any file inside the `config` dir as covered and I don’t know why.

lukasbestle:
Ah, I know why. PCOV [only covers the `src` dir by default if it finds one](https://github.com/krakjoe/pcov#configuration). That should be rather easy to fix.

bastianallgeier:
So good! Thank you for taking care of this! What was the PHP 7.1 issue?

lukasbestle:
@bastianallgeier The reason why only PHP 7.1 failed was because PHP 7.1 didn‘t respect the increased memory limit of 1 GB and used 512 MB instead, which was not enough.

This PR doesn’t fix the underlying issue, but since PCOV is much more memory-efficient, it doesn‘t break the memory limit.

We still need to find out which change in our code caused all this, otherwise Kirby 3.3.0 will possibly be slower in real-world use. I have one idea for further tests (writing a script that compares the test run times between the last working commit and `develop`). I will write that script in the next few days, but we can still merge this PR once it is ready. I will review the PR this evening.

bastianallgeier:
@lukasbestle Stupid question: isn't it possible that we just went over the 512MB mark with one of the new tests? In this case it would not necessarily be an indicator for slower performance, but just more required MBs for test code. PHP 7.1 is less effective than the newer versions so it would make sense.

lukasbestle:
It‘s not about PHP 7.1. As I wrote, it just doesn‘t respect the memory limit setting. PHP 7.2 and 7.3 test with 1 GB and 7.1 with 512 MB.

It is definitely possible that we went over the limit because of new tests, but that wouldn‘t explain why Travis runs suddenly took over 30 minutes (instead of 4) starting on Friday.

If I can confirm that it’s nothing in particular using my script, it‘s even better. But I think we need that test, otherwise we won‘t be sure that it wasn’t caused by new code.

afbora:
Now better 👍

lukasbestle:
@afbora Thanks! I think we need to fix this a bit differently, but that‘s no problem. I will make the required changes this evening.

afbora:
Thank you @lukasbestle Even so i gave you the working version 🙂 You can close the PR if you want btw. Like I said, I opened the PR to just give you ideas 👍

lukasbestle:
Let‘s keep it open. The changes you made are pretty good already, it just needs a bit of cleanup.

afbora:
Just out of curiosity. Will you continue to use `PCOV`?

lukasbestle:
Yes, of course!

afbora:
In the meantime there is a really serious performance difference and i like that 🚀 

### Before @lukasbestle Optimize

**Tests**
Time: 2.67 minutes, Memory: 420.00 MB

**Total Time**:
4 min 48 sec

### After @lukasbestle Optimize

**Tests**
Time: 1.52 minutes, Memory: 469.00 MB

**Total Time**:
2 min 55 sec

### This Optimize

**Tests**
Time: 41.58 seconds, Memory: 461.00 MB

**Total Time**:
2 min 25 sec

Maybe with the new optimizations that @lukasbestle will make, it can fall below 2 minutes 👍 

_* Benchmark for PHP 7.3_

lukasbestle:
> Maybe with the new optimizations that @lukasbestle will make, it can fall below 2 minutes 👍

✅With a bit of extra dependency caching, everything is possible. 😄

Bildschirmfoto 2019-10-16 um 20 54 42

I think this is at a point now where it can't really get any faster. 70 seconds of the 110 seconds are the pure testing steps and another three seconds are used for submitting coverage data to Coveralls. That leaves us with just 37 seconds of VM setup, dependency installation etc. Pretty much unbeatable.

---

@afbora Thanks a lot again for your help, the testing setup is really great now. ❤️

lukasbestle (approved):

distantnative:
Big props to you two!

afbora:
Thanks @lukasbestle It was under 2 minutes and an amazing update 💪

PR #2222, opened 2019-10-16
Add event to reload Panel

distantnative:
Fixes https://github.com/getkirby/ideas/issues/212


bastianallgeier:
Did you check if this actually reloads a view and retriggers all the events? Last time I checked it didn't really work, but that was quite some time ago.

distantnative:
Tried it in a custom view, but better check it again from a field or so in PageView etc. - you are right!

bastianallgeier:
Ah, it turns out this triggers a window reload. So pretty much the same as calling window.reload(). I think we need a more Vue-ish way of reloading stuff.

PR #2223, opened 2019-10-17
Add locking feature to structure field

afbora:
## Describe the PR

If the structure field data is being loaded by the hook, sometimes we don't want to add new rows or remove existing rows by the end users. The locking feature does exactly that. Locking allows editing rows only, unlike the disabled property.

### Usage

````yaml
fields:
  addresses:
    label: Addresses
    type: structure
    lock: true
    fields:
      street:
        label: Street
        type: text
      zip:
        label: ZIP
        type: text
      city:
        label: City
        type: text
````

## Related issues

- Closes https://github.com/getkirby/ideas/issues/312

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/fields/structure.php
+++ b/config/fields/structure.php
@@ -51,6 +51,12 @@ return [
             return $limit;
         },
         /**
+         * Locks adding and removing entries.
+         */
+        'lock' => function (bool $lock = false) {
+            return $lock;
+        },
+        /**
          * Maximum allowed entries in the structure. Afterwards the "Add" button will be switched off.
          */
         'max' => function (int $max = null) {
--- a/panel/src/components/Forms/Field/StructureField.vue
+++ b/panel/src/components/Forms/Field/StructureField.vue
@@ -4,7 +4,7 @@
     <!-- Add button -->
     <template slot="options">
       <k-button
-        v-if="more && currentIndex === null"
+        v-if="more && currentIndex === null && !lock"
         ref="add"
         :id="_uid"
         icon="add"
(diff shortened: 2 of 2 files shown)


distantnative:
I'm not happy with the name as it suggests a connection to content locking of pages. I feel this needs a different term.

afbora:
@distantnative Hmm, yes, you're right, it might cause confusion, I didn't think of it.
Do you have a different and more descriptive name in mind?

The other names that comes to my mind: `editOnly: true`,  `updateOnly: true` 🤔

lukasbestle (commented):
I think `updateOnly` is pretty good already if we don't find a better name.

Besides that: I'm wondering if we need backend validation here. Only disabling/removing the buttons won't prevent changes via the API for example.

nilshoerrmann:
What about `fixed: true`?

afbora:
> Besides that: I'm wondering if we need backend validation here. Only disabling/removing the buttons won't prevent changes via the API for example.

@distantnative ?

bastianallgeier:
I think fixed is pretty cool. `updateOnly` is more clear, but also a bit clunky. I take both. 
Backend validation: yes, absolutely!

texnixe:
+1 for `updateOnly: true`

nilshoerrmann:
Additional proposals: `add: false` or `new: false`.

afbora:
Sorry, this validation was more difficult than I expected. 
Because I need to compare the size of old data with the size of new data. 
For ex: 2 rows (old data) === 2 rows (new data)
The Field class does not contains to the old data, so comparison and validation cannot be performed.

Maybe not just for this feature, but a method that retains old data can be added to the Field class. Like that: `$field->old()`

As a result, we won't be able to add this feature right now 👎 (_Unless you have a different idea of validation_)

afbora:
I'm closing for now, if I can't find a more stable solution, I'll delete the branch.

Issue #339, opened 2018-03-30
[Cms] Implement custom Exceptions in Kirby\Cms


distantnative:
- Convert all existing Exceptions to our custom Exceptions (see https://github.com/k-next/kirby/pull/71)
  - [x] Mixins
  - [x] Tags
  - [x] App.php
  - [x] AppCaches.php
  - [x] AppHooks.php
  - [x] AppOptions.php
  - [x] AppPlugins.php
  - [x] Avatar.php
  - [x] AvatarActions.php
  - [x] AvatarRules.php
  - [x] AvatarStore.php
  - [x] AvatarStoreDefault.php
  - [x] Base.php
  - [x] Blueprint.php
  - [x] BlueprintCollection.php
  - [x] BlueprintConverter.php
  - [x] BlueprintFieldsSection.php
  - [x] BlueprintFilesSection.php
  - [x] BlueprintObject.php
  - [x] BlueprintOptions.php
  - [x] BlueprintPagesSection.php
  - [x] BlueprintSection.php
  - [x] BlueprintTabs.php
  - [x] Children.php
  - [x] ChildrenFinder.php
  - [x] Collection.php
  - [x] Collections.php
  - [x] Component.php
  - [x] Config.php
  - [x] Content.php
  - [x] ContentField.php
  - [x] Email.php
  - [x] EmailTemplate.php
  - [x] Extend.php
  - [x] File.php
  - [x] FileActions.php
  - [x] FileBlueprint.php
  - [x] FileBlueprintOptions.php
  - [x] Filename.php
  - [x] FileRules.php
  - [x] Files.php
  - [x] FilesFinder.php
  - [x] FileStore.php
  - [x] FileStoreDefault.php
  - [x] Form.php
  - [x] HasChildren.php
  - [x] HasContent.php
  - [x] HasFiles.php
  - [x] HasModel.php
  - [x] HasSiblings.php
  - [x] HasSingleton.php
  - [x] HasStore.php
  - [x] HasTemplate.php
  - [x] HasThumbs.php
  - [x] HasUnknownProperties.php
  - [x] Hooks.php
  - [x] Ingredients.php
  - [x] Locale.php
  - [x] Locales.php
  - [x] Media.php
  - [x] Model.php
  - [x] Nest.php
  - [x] NestCollection.php
  - [x] NestObject.php
  - [x] Page.php
  - [x] PageActions.php
  - [x] PageBlueprint.php
  - [x] PageBlueprintOptions.php
[...]

PR #2224, opened 2019-10-17
Check if dump function already exists

lemmon:
## Describe the PR

Check whether the dump helper function already exists. Helpful when using Kirby with composer installation + symfony/var-dumper. Vendor extensions get loaded first thus Kirby fails to initialize. This patch fixes this error and adds ability to replace the dump helper with different possible dump utilities.

### DIFF
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -129,19 +129,21 @@ function css($url, $options = null)
  * @param boolean $echo
  * @return string
  */
-function dump($variable, bool $echo = true): string
-{
-    if (Server::cli() === true) {
-        $output = print_r($variable, true) . PHP_EOL;
-    } else {
-        $output = '<pre>' . print_r($variable, true) . '</pre>';
-    }
+if (!function_exists('dump')) {
+    function dump($variable, bool $echo = true): string
+    {
+        if (Server::cli() === true) {
+            $output = print_r($variable, true) . PHP_EOL;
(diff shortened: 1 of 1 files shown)


lukasbestle:
See https://github.com/getkirby/ideas/issues/339.

disjfa:
Yes, it was said it was in 3.3.0 but we are in 3.3.1 and it still breaks sadly, well ok.

robmeek:
I’d love this very simple PR to be reconsidered and merged.
Having looked at #339 and #2341 I can’t see a good reason for not merging.

In my use case I do a composer install of both kirby and a set of symfony components (not explicitly the var dumper – it’s pulled in as a dependency of other libraries) – and as others have reported, this breaks the app – “cannot redeclare dump” etc.

I wish that no library declared global functions, but very occasionally it’s done, and the symfony var dumper component does the right thing which is to check for the function’s existence before declaring it. 

I think Kirby should do the same – what is the harm? It’s seems like the responsible thing for do, especially considering that this is not a function that you would expect to ever use in production. 

Note that I personally do not want to use the symfony var dumper in my projects – that is not the problem in my case. But the lack of this pre-existence check means that, currently, no-one can easily integrate Kirby and many standard Symfony components with a composer install.     

PS: I love Kirby and Symfony. They work very well together (Once you’ve got them installed ;-)

PR #2226, opened 2019-10-19
Remove the name field from the installer to set the focus on the email address #1802

### LINKED ISSUE #1802: Improve usability for first time usage - registration process is misinterpretable
My first Kirby 3 installation was a short time fail because I did a major mistake: When I followed the admin registration process, I interpreted "name" as the username and typed in some random email address because it was a local set up for testing. Of course, afterwards I wasn't able to log in because there was no way to sign in via name but only via email address. It should be made more clearer that the email address is used for the sign in and that you have to remember it/store it somewhere safely. Otherwise you have to delete the admin account as like as I had to do.

bastianallgeier:
## Describe the PR

I understand that the name field in the installer can be mistaken as a username field, which can get quite confusing. IMHO the most simple first step is to remove it entirely from the installer. It's not required to get up and running and I think less fields == less friction. How do you feel about it? 

## Related issues

- Fixes #1802

### DIFF
--- a/panel/src/components/Views/InstallationView.vue
+++ b/panel/src/components/Views/InstallationView.vue
@@ -101,12 +101,6 @@ export default {
     },
     fields() {
       return {
-        name: {
-          label: this.$t("name"),
-          type: "text",
-          icon: "user",
-          autofocus: true
-        },
         email: {
           label: this.$t("email"),
           type: "email",


afbora:
I really liked it ❤️

afbora (approved):

lukasbestle (approved):

distantnative (approved):

distantnative:
I feel this is something that we in the team all agree on but maybe our developer community might have different thoughts?

bastianallgeier:
Well, let's just find out :)

PR #2252, opened 2019-10-28
Correct return type for js helper function

matteodem:
Analog to css()

## Describe the PR

Simple phpdoc adjustment 😄


afbora:
@matteodem Thank you for contributing. This is already fixed on 3.3.0.

https://github.com/getkirby/kirby/blob/release/3.3.0/config/helpers.php#L351-L385

PR #2275, opened 2019-11-04
Fixed that language base URLs should be copared to the kirby base URL.

tobiastom:
## Describe the PR

I think the `LanguageRoutes` change introduced a bug that installations inside subdirectories won't work.

Previously the `$baseUrl` variable would contain e.g. `http://127.0.0.1/~tobias/kirby/public`, that should be compared to the base URL of the language, which is just `http://127.0.0.1`. 
The result would be that the languages could not be loaded.

This pull request makes sure the base URL is `http://127.0.0.1`.

## Ready?

- [x] Passing all unit tests


afbora:
@tobiastom Could you test https://github.com/getkirby/kirby/pull/2411 PR please?

afbora:
I think this problem has been solved with #2411 regarding not working in subdirectories. If you are still experiencing this issue, you are free to reopen it.

tobiastom:
It seems to be gone indeed. Thank you.

PR #2286, opened 2019-11-06
Fix toDate method to work with unicode chars

### LINKED ISSUE #2185: toDate in blueprints using strftime breaks panel requests
**Describe the bug**  
Using toDate in blueprints with `date.handler` option set to `strftime` breaks requests response in the Panel, it throws `The JSON response from the API could not be parsed. Please check your API connection.`

**To Reproduce**  
- Set the Kirby option `data.handler` to `strftime`
- Create a `blog` blueprint using toDate, for example:
```yaml
title: Blog

sections:
  published:
    headline: Published articles
    type: pages
    status: listed
    templates: article
    sortBy: date desc
    info: "{{ page.date.toDate('%d %B %G') }}"
```
- Create an `article` blueprint:
```yaml
title: Article
preset: page

fields:
  date:
    icon: calendar
    label:
      en: Date
      fr: Date
    type: date
    placeholder: Select a date…
    default: today
    required: true
    width: 1/3
    translate: false
```
- In the panel create a page of type `blog` and at least one published `article` page
- Go to the `blog` page
- The page throws `The JSON response from the API could not be parsed. Please check your API connection.` under the `published` section

**Expected behavior**  
The panel page should properly list the published articles with the date displayed in the `info` field

**Kirby Version**  
3.2.5

**Desktop:**  
 - OS: macOS
 - Browser: Firefox
 - Version: 69.0.2

**Additional context**  
The fetch request to grab the articles in the Panel returns an empty response with code 200. I don't have any error in my logs.

The problem doesn't occur in my local setup but does on the remote server. Could it be a difference in the server configuration? Both are using Apache and PHP 7.3

This is similar to https://forum.getkirby.com/t/i-dont-see-my-blogposts-via-the-panel/12968/20?u=leben but in my case using the `date` format just outputs `Y.m.d` as is; it does prevent the error though. Could it be caused by the `%` in the strftime format?

afbora:
## Describe the PR

Fixed the issue by converting unicode characters using with `Str::convert()` method

## Related issues

- Fixes #2185

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/methods.php
+++ b/config/methods.php
@@ -106,7 +106,7 @@ return function (App $app) {
                 return $time;
             }
 
-            return $app->option('date.handler', 'date')($format, $time);
+            return Str::convert($app->option('date.handler', 'date')($format, $time), 'utf-8');
         },
 
         /**


lukasbestle (commented):
The issue and the solution are both pretty weird to be honest. Have you been able to track down *why* this solution works? :)

afbora:
@lukasbestle I tracked down and this issue related to @distantnative PR that #2105 

This ensures that the PR output is always UTF-8 encoding.

**Before this PR**:

    26 Eyl�l 2018

**After this PR**:

    26 Eylül 2018

Any ideas?

lukasbestle (commented):
Ah, yes! Another locale edge case.

I think this is a good first step. However we additionally need to do the following:

- [ ] Check which other methods return strings that depend on the PHP locale (which can either be ISO-8859-1 or UTF-8); all of those should return normalized UTF-8 strings.
- [ ] Fix the root of the original issue: Why do API responses completely break with invalid chars? The API should return a generally valid response with the invalid chars in it.

afbora:
Will fix with #2298

PR #2298, opened 2019-11-10
Set API locale using a new key in the translation file #2233

### LINKED ISSUE #2185: toDate in blueprints using strftime breaks panel requests

### LINKED ISSUE #2233: Setting API locale does not work for many locales
This https://github.com/getkirby/kirby/blob/master/src/Api/Api.php#L178-L199 does not work for locales without `_` where the first and second part are not equal, e.g. `en_US` from `en`.

Maybe we should add something like:
https://github.com/peterkahl/country-to-locale

lukasbestle:
## Describe the PR

The API locale was generated dynamically, which failed for some languages. The locale is now set in the translation file, so it can be customized for each translation.

## Related issues

- Fixes #2233
- Fixes #2185 

## Ready?

- [x] Passing all unit tests


afbora (commented):
Perfect and simple solution, I love it ❤️ 

But the issue #2185 continues. I tracked down and found that the reason was the locale orders in lines:

https://github.com/getkirby/kirby/blob/fix/2233-api-locales/src/Api/Api.php#L185-L192

**My locale output:**
````php
var_dump(setlocale(LC_ALL, 0));
````

> LC_COLLATE=en_US.utf-8;LC_CTYPE=en_US.utf-8;LC_MONETARY=tr_TR;LC_NUMERIC=tr_TR;LC_TIME=tr_TR

Should be like that `tr_TR.utf-8`, instead of `tr_TR`.

I changed the order like following (utf8 take to the top) and working but i'm sure you'll find a better solution.

````php
$locales = [
    $locale . '.UTF-8',
    $locale . '.UTF8',
    $locale . '.ISO8859-1',
    $locale,
    $language,
    setlocale(LC_ALL, 0)
];
````

**Now locale output is**:

> LC_COLLATE=en_US.utf-8;LC_CTYPE=en_US.utf-8;LC_MONETARY=tr_TR.UTF-8;LC_NUMERIC=tr_TR.UTF-8;LC_TIME=tr_TR.UTF-8

lukasbestle:
One fix for two issues, awesome! You are of course completely right about the fallback order – fixed now.

afbora:
@lukasbestle I think we're facing another problem 😭 What if the server does not support UTF8 or can't find it (utf8 encoding) like following example? Unfortunately, it didn't work on my other pc localhost (Windows):

I'm getting always without utf8 encoding:

> LC_COLLATE=en_US.utf-8;LC_CTYPE=C;LC_MONETARY=tr;LC_NUMERIC=tr;LC_TIME=tr

I tried to change in translation file but didn't work too:

    "translation.locale": "tr_TR.utf-8",

I don't quite understand why the problem comes from, but is this Kirby's or user's problem? 🤔

lukasbestle:
As your locales are set to the language code without full country locale, it looks like that PC doesn‘t have a Turkish locale at all. It could also be that it‘s because it‘s a Windows machine, so it may need a different locale syntax.

We can‘t really do anything about those edge-cases though. 🤷‍♂️

afbora:
I understand. I think it would be useful to add this to the requirements and troubleshooting section of the documents 👍 What do you think? @texnixe

afbora (approved):

lukasbestle:
We have already:

> The server OS must be POSIX compatible.

bastianallgeier (approved):

PR #2309, opened 2019-11-15
Str::template(): Keep placeholders in the result string if no value or fallback was passed

### LINKED ISSUE #2307: Str::template() Removes Token-Like Fields That Don't Have Mappings
**Describe the bug**  
When using the `Str::template()` method, if the source string has "token"-like parts to it (but are not meant to be replaced) then they will be unexpectedly removed.

```php
Str::template('From {{ b }} to {{ a }}, maybe {{ c }}.', [
    'a' => 'there',
    'b' => 'here'
]);
//   Actual: From here to there, maybe .
// Expected: From here to there, maybe {{ c }}.
```

This actual behavior differs from K2-behavior, and to me it seems like a bug.

**Kirby Version**  
`3.3.0`

lukasbestle:
## Describe the PR

If a placeholder is used in the input template but not defined in the data array and if there is also no fallback defined, the `Str::template()` method now keeps the original placeholder in the output as if it wasn't a placeholder.

**Open question:** This fix is a breaking-change. It's a small change and I checked all occurrences where the method is currently used in the core, but it might make sense to include this in 3.4.0 just to be on the safe side.

## Related issues

- Fixes #2307 

## Ready?

- [x] Passing all unit tests


distantnative:
I agree. This is a breaking change, if only a small one. And it’s a feature that isn’t so crucial to directly include in the next release. I’d feel better to hold on to this for 3.4.

bastianallgeier (approved):

PR #2310, opened 2019-11-16
Add missing since information

texnixe:
## Describe the PR

Adds missing `@since` info to new 3.3.0 methods

## Related issues

- Fixes #

## Ready?

- [ ] Passing all unit tests

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -784,6 +784,7 @@ class App
     /**
      * Returns the nonce, which is used
      * in the panel for inline scripts
+     * @since 3.3.0
      *
      * @return string
      */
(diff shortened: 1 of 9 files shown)


src/Cms/App.php
```diff
@@ -848,6 +849,7 @@ protected function optionsFromProps(array $options = []): array
 
     /**
      * Merge last-minute options from ready callback
+     * @since 3.3.0
```
lukasbestle: Do we need that one as that's a protected method?

texnixe: I wondered as well, but then thought that this type if information is generally useful

lukasbestle: I'm not sure. Protected methods are internal implementation details, so it shouldn't matter when they were changed/added. It would be a lot of work to add that information everywhere, so I think that it isn't worth it for the limited use.

👍 for all other changes though!

lukasbestle (approved):

Issue #2288, opened 2019-11-06
[Panel] Number Field Does Not Save Empty Value After Losing Focus (3.3.0 Regression)


neildaniels:
**Describe the bug**  
If you attempt to clear a `number` field in the Panel, it will only save the "empty" value if you save the page while the number field still has typing focus. If you click on another field, the "empty" value will never actually be saved (even if you click back into the original field).

I suspect there's something wrong with savings values in general, but can only reliably produce this "empty value" scenario as described.

**To Reproduce**  
1. Have the following Blueprint `test-template.yml`
```yaml
title: Test Template

fields:
  foo:
    type: number
    width: 1/2
  bar:
    type: number
    width: 1/2
  text:
    type: textarea
    width: 1/2
```
2. Have the following template `test-template.php`
```php
<html>
    <body>

Foo: <?= $page->foo()->html() ?>

<br><br>

Bar: <?= $page->bar()->html() ?>

<br><br>

Text: <?= $page->text()->html() ?>
```
3. Create a page in the Panel with this template type
4. Fill in values such as `2`, `4`, and `some text`.
<img width="1020" alt="Screen Shot 2019-11-06 at 2 27 14 PM" src="https://user-images.githubusercontent.com/417182/68343390-889dd080-00a1-11ea-8345-aad65e637852.png">

5. Save that in the Panel and view the page.
<img width="167" alt="Screen Shot 2019-11-06 at 2 26 49 PM" src="https://user-images.githubusercontent.com/417182/68343376-7a4fb480-00a1-11ea-91dc-4628dbb85a9e.png">

6. Go back to the Panel page and clear both number field _then_ change the text field to `other text`. **Save the page with Command-S**

<img width="1018" alt="Screen Shot 2019-11-06 at 2 28 03 PM" src="https://user-images.githubusercontent.com/417182/68343437-a66b3580-00a1-11ea-8202-dba6a3cf3025.png">
7. Re-load the rendered page.
<img width="155" alt="Screen Shot 2019-11-06 at 2 28 35 PM" src="https://user-images.githubusercontent.com/417182/68343473-b8e56f00-00a1-11ea-902f-b65d0b9f0cfc.png">
8. Go back to the Panel and re-load the page. You'll see that the original `2` and `4` values return.

**Kirby Version**  
`3.3.0`

**Desktop (please complete the following information):**  
 - OS: macOS
 - Safari 13.0.3 (15608.3.10.1.4)
 - Chrome 78.0.3904.87

[...]

PR #2311, opened 2019-11-18
Fix number field issues #2306

### LINKED ISSUE #2306: Number field auto-removes decimals if step option is not set.
**Describe the bug**  
If you enter a decimal value into a number field that has no decimals defined, as soon as the field is no longer focused, the decimals are cut off without an further hint that there was something wrong with the data. Shouldn't the field validate the entered value?

**To Reproduce**  
1. Define a number field without the step option
2. Go to the Panel and enter a number with decimals and save.
3. See as when you leave the field, the decimals are removed without any explanation, but the save bar pops up again.

**Expected behavior**  
Maybe the field should come up with a warning rather than popping up the save bar and auto-removing the decimals as the current behavior is somewhat confusing.

**Kirby Version**  
Tested with Kirby 3.3.0.

bastianallgeier:
## Describe the PR

- Decimals are no longer removed by default
- Empty values are stored correctly again

## Related issues

- Fixes #2306 & #2288

### DIFF
--- a/config/fields/number.php
+++ b/config/fields/number.php
@@ -25,7 +25,7 @@ return [
         /**
          * Allowed incremental steps between numbers (i.e `0.5`)
          */
-        'step' => function ($step = 1) {
+        'step' => function ($step = null) {
             return $this->toNumber($step);
         },
         'value' => function ($value = null) {
--- a/panel/src/components/Forms/Input/NumberInput.vue
+++ b/panel/src/components/Forms/Input/NumberInput.vue
@@ -76,11 +76,13 @@ export default {
   },
   methods: {
     decimals() {
-      if (Math.floor(this.step) === this.step) {
+      const step = Number(this.step || 0);
+
+      if (Math.floor(step) === step) {
         return 0;
       }
 
-      return this.step.toString().split(".")[1].length || 0;
+      return step.toString().split(".")[1].length || 0;
     },
     format(value) {
       const decimals = this.decimals();
@@ -89,7 +91,11 @@ export default {
         return parseFloat(value).toFixed(decimals);
       }
 
-      return parseInt(value);
+      if (Number.isInteger(this.step)) {
+        return parseInt(value);
+      }
+
+      return parseFloat(value);
     },
     focus() {
       this.$refs.input.focus();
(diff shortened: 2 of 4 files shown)


afbora (commented):
After I finish writing the number, if I click outside the input after saving the page while still focus on number input, the value changes (with step option) and the savebar appears (always).

number

afbora:
https://github.com/getkirby/kirby/blob/fix/2306-keep-decimals/panel/src/components/Forms/Input/NumberInput.vue#L111

@bastianallgeier I guess, is that supposed to be like that?

````js
this.number = this.format(value);
````

bastianallgeier:
@afbora yes! Otherwise the formatting would be happening instantly when you type and strange side effects happen.

afbora:
> this.number = this.format(value);

@bastianallgeier New commit?

bastianallgeier:
We cannot solve the problem of the save bar in any other way. The forced formatting fixes for the number field have to happen once you focus on another field. They cannot happen when typing in the field. That's why it can happen that the value changes again. But I think this is fine.

afbora:
Hmm.. I'm not sure but how about we check that the value really changes in this way? Seems to be working 🤔 

````js
onChange(value) {
  this.number = this.format(value);
  
  if (value !== this.number) {
    this.$emit("input", this.number);
  }
},
````

bastianallgeier:
@afbora I don't really understand the idea behind it

afbora:
Since `onChange()` run when we focus out the input after saving the value, let `$emit()` run if the value in the input has actually changed.

Now, whether the value changes or remains the same bar is constantly coming out after focus out.

bastianallgeier:
Ok, I think we are overthinking this issue for the benefit of a small visual improvement. I reverted everything to let the HTML number input decide. Decimals are not shown when they are just .00 and I think this is actually fine. Whenever we tried to "improve" the number input it went down the drain and we introduce bugs that were completely unnecessary. 

How do you feel about this @distantnative, @lukasbestle, @texnixe?

bastianallgeier:
I take everything back. I think I have a solution now that's somewhere between what I came up with first and what @afbora suggested. Could you have a look at it again @distantnative @lukasbestle @texnixe

bastianallgeier:
I decided to merge this now. It's a lot better than the solution in 3.3.0 and if we find additional issues in the RC we can still fix it.

PR #2319, opened 2019-11-23
Use default lang for page num query #2256

### LINKED ISSUE #2256: Content sorting by date fails on multilingual site
**Describe the bug**  
We have a multilingual site (German + English) with texts that should be sorted by date. This is defined in our blueprint as follows:

```yml
title: Text
num: '{{ page.date.toDate("%Y%m%d") }}'
```

The date field does not allow translations and is populated in the main language only (German):

```yml
date:
  type: date
  min: 2000-01-01
  label: Datum
  translate: false
  default: today
```

Thus there is never a date stored inside the English content file, only inside the German one. Everything is fine as long as we edit the German version, all pages get their date prepended to their folder an sorting works as expected:

> `20191029_my-text`

As soon as we switch to English and do any content edits, the page folder gets a zero prefix on save. My guess is that this is because Kirby tries to look up the date inside the English content file and doesn't default to the German one where the correct date can be found:

> `0_my-text`

As soon a we edit the German version again, the folder changes back to the desired sorting prefix:

> `20191029_my-text`

**Expected behavior**  
Kirby should correctly look up the date in the main language file and stick to the correctly prefixed folder name.

**Kirby Version**  
3.3.0 RC 2 – I can't remember that this happened before so I guess this is a regression between ~3.1 and now.

distantnative:
## Describe the PR

Resolve the query for a page's `num` option always with the default language.

## Related issues

- Fixes #2256

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -506,10 +506,14 @@ trait PageActions
 
                 return $num;
             default:
+                // get instance with default lanuage
+                $app = $this->kirby()->clone();
+                $app->setCurrentLanguage();
+
                 $template = Str::template($mode, [
-                    'kirby' => $this->kirby(),
-                    'page'  => $this,
-                    'site'  => $this->site(),
+                    'kirby' => $app,
+                    'page'  => $app->page($this->id()),
+                    'site'  => $app->site(),
                 ]);
 
                 return (int)$template;
--- a/tests/Cms/Pages/PageSortTest.php
+++ b/tests/Cms/Pages/PageSortTest.php
@@ -380,27 +380,89 @@ class PageSortTest extends TestCase
     public function testCreateCustomNum()
     {
         // valid
-        $page = new Page([
-            'slug' => 'test',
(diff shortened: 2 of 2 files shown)


lukasbestle (approved):

afbora (approved):

src/Cms/PageActions.php
```diff
@@ -506,10 +506,14 @@ public function createNum(int $num = null): int
 
                 return $num;
             default:
+                // get instance with default lanuage
+                $app = $this->kirby()->clone();
+                $app->setCurrentLanguage();
```
distantnative, line 511: @lukasbestle just wondering: this does set the default language as current, but should we make it even more explicit (actually passing the default language)?

lukasbestle: I thought the same when reviewing, but in the end it should be fine. Maybe we could explicitly pass `null` to make it clear that the default language is intended.

PR #2320, opened 2019-11-23
Disable base64 encoding for `File::dataUri` #2276

### LINKED ISSUE #2276: [Toolkit\File] Option to disable base64 encoding in `dataUri()` method
The `dataUri` method is quite handy, especially for inlining small resources like icons etc. The `dataUri` method always uses Base64 encoding, even if the file’s format is a plaintext format (e. g. SVG).

In those cases, encoding the file using Base64 wouldn’t be necessary. That’s not a severe problem, but it’s inefficient. Encoding using Base64 adds an [overhead of ~33 percent in file size](https://stackoverflow.com/questions/4715415/base64-what-is-the-worst-possible-increase-in-space-usage). Additionally, Base64-encoded data might [reduce the efficiency of gzip compression](https://csswizardry.com/2017/02/base64-encoding-and-performance/#gzip-saves).

I’d propose to disable Base64 encoding for plaintext formats or, alterantively, add an option to the `dataUri` method to disable it manually. As there are [some edge cases](https://codepen.io/tigt/post/optimizing-svgs-in-data-uris) where it might be required or otherwise sensible to encode plaintext data, I think the latter would be best.

distantnative:
## Describe the PR

Add a parameter to `File::dataUri`, allowing to disable base64 encoding.

## Related issues

- Fixes #2276

## Ready?

- [x] Passing all unit tests


src/Toolkit/File.php
```diff
@@ -72,11 +72,16 @@ public function copy(string $target, bool $force = false)
...
      */
-    public function dataUri(): string
+    public function dataUri(bool $base64 = true): string
     {
-        return 'data:' . $this->mime() . ';base64,' . $this->base64();
+        if ($base64 === true) {
+            return 'data:' . $this->mime() . ';base64,' . $this->base64();
+        }
+
+        return 'data:' . $this->mime() . ',' . $this->read();
```
lukasbestle: Wouldn't this need a run through `rawurlencode()` to produce valid output?

distantnative: I'm afraid you have more knowledge here than me... so not sure.

lukasbestle (approved):
I've changed it to use URL-escaping and also moved the unit tests to the class they actually belong to. No idea why we had the test in the `ImageTest` class so far. 🤷‍♂

PR #1881, opened 2019-06-26
Feature: Integrate Symfony VarDumper component

ceesvanegmond:
## Describe the PR
This pull requests integrates the famout Symfony VarDumper component in Kirby. I find it very handy in comparison the `dump` function in Kirby itself. I added it as a dev component only.
For more info about this component: https://symfony.com/doc/current/components/var_dumper.html

Please let me know what you think about this enhancement.

## Todos
If you feel like you can help to check off the following tasks, that'd be great. If not, don't worry - we will take care of it.
- [ ] Add docs

PR #2341, opened 2019-12-06
Allow symfony var dumper to be used

disjfa:
## Describe the PR

I like the symfony var dumper, but it breaks the base kirby helper. If you add a check if the function exists it can be used in stead of the Simple object and variable dumper.

## Related issues

None, better development

## Ready?

Yes

- [ ] Passing all unit tests

### DIFF
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -136,18 +136,20 @@ function deprecated(string $message): bool
     return false;
 }
 
-/**
- * Simple object and variable dumper
- * to help with debugging.
- *
- * @param mixed $variable
- * @param bool $echo
- * @return string
- */
-function dump($variable, bool $echo = true): string
-{
-    $kirby = App::instance();
-    return $kirby->component('dump')($kirby, $variable, $echo);
+if (false === function_exists('dump')) {
+    /**
+     * Simple object and variable dumper
+     * to help with debugging.
+     *
+     * @param mixed $variable
+     * @param bool $echo
+     * @return string
+     */
+    function dump($variable, bool $echo = true): string
+    {
+        $kirby = App::instance();
+        return $kirby->component('dump')($kirby, $variable, $echo);
+    }
 }
 
 /**


afbora:
Duplicated #2224

disjfa:
I know, but it is still broken in 3.3.1 sadly. Also the patch mentioned is not in the develop branch, so maybe it is better to implement.

afbora:
Check out dump component:
https://getkirby.com/docs/reference/plugins/components/dump

And check this comment: https://github.com/getkirby/kirby/pull/1881#issuecomment-507326564

lemmon:
I completelly agree with the commet https://github.com/getkirby/kirby/pull/1881#issuecomment-507326564 but I don't think anybody wants to integrate the symfony/var-dumer into Kirby's core.

However, it would be nice if you could extend the dump functionality. You are not able to do that now when you are using composer. Both Kirby and VarDumper are fighting over the `dump()` function. Symfony does check for existence of the `dump()` function. I think that Kirby should do the same and all will be resolved.

afbora:
@lemmon Same issue with `e()` helper on `Illuminate\View` of Laravel repository. But a stable solution has not been found yet.

lukasbestle:
It's unfortunate that `symfony/var-dumper` always registers the global `dump()` function. There doesn't seem to be a way to just autoload the classes without the `dump()` function. If there was, you could use the new Kirby `dump` component to set your own `dump()` output.

To be honest I'm not sure if it's a good idea to fix this with a conditional helper though. The two implementations will then fight and the one that gets loaded first will win. Not particularly predictable.

Assuming that you installed both Kirby and `symfony/var-dumper` using Composer: What you could do is to manually `require_once(__DIR__ . '/kirby/config/setup.php')` from your `index.php` before the Kirby bootstrapper is called. Alternatively, you could add that file to your Composer autoloader of the site and then run `composer update` to re-generate the autoloader. Both methods should ensure that the Kirby helper is loaded first. You can then register the `VarDumper` class as the Kirby `dump` component and you should be good to go.

If that doesn't work, please let me know. :)

lemmon:
@lukasbestle this seems like a way to go. Should work. But as @afbora mentioned the same thing might happen with another functions. Using same hack you will end up with too much hacking around and messy code.

Namespacing Kirby, maybe?

lukasbestle:
I agree that it's not optimal, but global helpers are unfortunately a feature that is difficult to get right. Let me explain the thoughts behind our implementation:

The helpers we ship with Kirby are an integral part of the CMS. We use them all across our documentation and people expect them to always work as described in the docs. The `dump()` helper is a bit of an exception as it has no clearly defined behavior, meaning that the exact output may vary. The public API of our helpers should still be consistent though, otherwise example code from the docs, the cookbook and from third-party sites and themes won't work reliably.

While this again doesn't apply to the `dump()` helper (which isn't commonly used in production), there's another argument: We can only provide support in the forum and on GitHub if the code that ships with Kirby is what we expect. Imagine that someone changes the code, reports an issue and we only find out about the changed code after hours of debugging. That happened to me before when supporting a different project and it was pretty bad for both the user (who would have gotten their solution much quicker) and me (who needed a lot of time I could have used for other support requests or development).

We have opted to build a `dump` component so that people can replace the internal behavior without changing the public API. This however only works for parts of Kirby where it makes sense to replace the core code without changing the behavior on the outside.

Making the helpers optional by checking if they are already defined definitely isn't going to work as I explained above: Imagine a plugin requires a third-party library that requires another library that defines such a helper. You would again be debugging for hours until you find out that the helper you are calling is not the Kirby one but the third party one.

Regarding your other suggestion:

> Namespacing Kirby, maybe?

I agree that this could have been a useful solution if we had decided on that early on. Template code is however sometimes ported from Kirby 1 or Kirby 2 sites or from third-party themes (that have a widely varying code quality). We would make all that code incompatible and couldn't even provide a fallback (as that would make the step to namespace all helpers completely useless).

This is something we could keep in mind for Kirby 4 though. Maybe we can then make the switch from `helper()` to `Kirby\helper()`. I have created an idea issue here: https://github.com/getkirby/ideas/issues/450

lemmon:
The problem happens when composer extensions are loaded before Kirby internals. Maybe Kirby should load first and only after that the composer autoloader is called. Or does Kirby use composer autoloader for its internals as well and the order is out of your hands @lukasbestle ?

lukasbestle:
> Or does Kirby use composer autoloader for its internals as well and the order is out of your hands

Yes. Switching to manually loading the helpers would break edge-case setups that rely on the autoloader.

disjfa:
Thanks for the update, still a bit baffled by the solution. But the inclusion of the `require_once(__DIR__ . '/kirby/config/setup.php')` adds a notice because it tries to define `DS` twice. In you append the `require_once(__DIR__ . '/kirby/config/helpers.php')` only it seems to be fixed. In the var dumper there also is a `dd()` function that can be used.

Thanks for the tips and tricks.

PR #2347, opened 2019-12-11
Fetch the last insert ID only with INSERT statements.

tobiastom:
## Describe the PR

I did implement a Postgresql Adapter for Kirby. PostgreSQL does not seem to like when the last insert ID is requested, but it was for example a `SELECT` query.

This PULL request changes the call to `lastInsertId` on `INSERT` queries.

I'm aware that this is not a nice solution, but based on the context that is available at that given point, it seems to be the only one.

## Ready?

- [ ] Passing all unit tests

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -369,7 +369,7 @@ class Database
             $this->statement->execute($bindings);
 
             $this->affected  = $this->statement->rowCount();
-            $this->lastId    = $this->connection->lastInsertId();
+            $this->lastId    = substr($query, 0, 7 ) == 'INSERT ' ? $this->connection->lastInsertId() : null;
             $this->lastError = null;
 
             // store the final sql to add it to the trace later


lukasbestle:
Thanks for the PR. I actually think that the solution is fine if it’s necessary. However we should probably use case-insensitive matching with `Str::startsWith($query, 'insert ', true)`.

Out of interest: Which kind of error did you get?

We’d also be happy to include the Postgres adapter in the core if you don’t mind sharing it.

tobiastom:
@lukasbestle I'll get back to you as soon as possible. I'm quite busy right now, sorry. :)

lukasbestle:
As we are refactoring the Database classes right now, I have added this fix to the current refactoring state as well. The refactored version will hopefully be in v3.4.0.

Thanks again for your PR! 👍

tobiastom:
Thanks for integrating it and sorry that I did not find the time to report back the error. We switched to the native `PDO` classes inside the project where I needed this fix.

PR #2349, opened 2019-12-12
Fix timezone bug in date field

### LINKED ISSUE #1813: Wrong date gets stored when in a different time zone
**Describe the bug**
I have a date field without time set up in a blueprint. When I input a date, it correctly gets stored in both the panel as well as in the txt file. However when my client (based in the UK) enters a date, the day after gets stored in the txt file and therefore in the panel as well. If we change his browser/system timezone to Europe/Berlin, this issue disappears.

**To Reproduce**
1. Add a date field to a content type
2. Change your system/browser timezone to UK or USA
3. Enter a date
4. Find that the date was changed to the day after

**Expected behavior**
The correct date should be visible.

**Kirby Version**
3.1.4

Happens on different browsers, systems and even mobile devices, if they are set to a different time zone.

**Additional context**
Seems to be a similar issue: https://forum.getkirby.com/t/date-field-shows-wrong-date-for-some-users/14167/11

pedroborges:
This issue has been very annoying for quite some time. I'm working on an event calendar plugin and it will be almost impossible for it to work correctly without fixing this issue.

The cause for this bug is: let's say a date field in the content file contains `2019-12-09 10:00`. PHP assumes that the server timezone is UTC and sends `2019-12-09T10:00:00+00:00` to the browser. Because dayjs receives the date including a timezone, it tries to be smart by offsetting the date with my system's timezone. Because I'm in Brazil (GMT-03:00) I see time as 07:00 in the date field. In some cases, you even see a different day. If I set the time to 15:00, It is saved correctly. When I refresh the page it shows 12:00 because dayjs offsets the date once again.

If I set the server timezone to be the same as in my system, it works fine. The problem would be solved if panel users were always in the same timezone. But we can do better and not try to guess panel users' timezone.

## Describe the PR

The solution is to send dates without timezone to the browser in the [`/config/fields/date.php`](https://github.com/getkirby/kirby/blob/21ebebda6bdde9aee62140f81f658f0852ea4e60/config/fields/date.php#L70) file. This changes makes the browser displays the exact same datetime as in the content file.

No change is required in the front-end because the date I see when hitting "save" is being stored correctly to the content file.

## Related issues

- Fixes #1813 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/fields/date.php
+++ b/config/fields/date.php
@@ -67,7 +67,7 @@ return [
     'methods' => [
         'toDate' => function ($value) {
             if ($timestamp = timestamp($value, $this->time['step'] ?? 5)) {
-                return date(DATE_W3C, $timestamp);
+                return date('Y-m-d H:i:s', $timestamp);
             }
 
             return null;
--- a/tests/Form/Fields/DateFieldTest.php
+++ b/tests/Form/Fields/DateFieldTest.php
@@ -30,9 +30,9 @@ class DateFieldTest extends TestCase
     public function valueProvider()
     {
         return [
-            ['12.12.2012', date(DATE_W3C, strtotime('2012-12-12'))],
-            ['2016-11-21', date(DATE_W3C, strtotime('2016-11-21'))],
-            ['2016-11-21 12:12:12', date(DATE_W3C, strtotime('2016-11-21 12:10:00')), 5],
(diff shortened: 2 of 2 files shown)


distantnative:
@pedroborges You are a champ for figuring this out (and me being mislead by the buggy timezone Chrome extension)!

PR #2352, opened 2019-12-14
Add translation CSS selector to Panel

distantnative:
## Describe the PR

```css
.k-panel[data-translation=en] {}
.k-panel[data-translation-default] {}
.k-panel:not([data-translation-default]) {}
.k-panel:not([data-translation-default]) .k-field:not([data-translate]) {}
```

## Related issues

Partially solves https://github.com/getkirby/ideas/issues/232


nilshoerrmann:
For custom styling of the panel, it would be interesting to determine whether a field is non-translatable. Therefore the language name alone might not be enough. It might also be good to have more abstract names like `primary-language` and `secondary-language` (refering to all translations that are not primary). This would ensure portability.

Maybe the language itself could be defined using the `lang` attribute instead of a custom data attribute? (Not sure if that makes sense though because the label language and the content language might defer.)

distantnative:
> Not sure if that makes sense though because the label language and the content language might defer
I stayed away from the `lang` attribute for that reason.

How about adding another `data-translation-default` attribute when it is the default translation language? Then you could do

```css
.k-panel[data-translation-default]
.k-panel:not([data-translation-default])
```

The non-translatable we would have to add on the field.

bastianallgeier (approved):

PR #2359, opened 2019-12-15
Uploads with different mime, same extension #1783

### LINKED ISSUE #1783: Cannot Replace a File With The Same Extension But Different MIME Type
**Describe the bug**
A "JPEG 2000" image was accidentally uploaded to the Panel, but it the original file had the extension `.jpg` instead of the more appropriate `.jp2`. However, once it was uploaded, the Panel would _only_ accept uploads that had the MIME type of `image/jp2`. The file picker wouldn't allow a replacement `.jpg` file, even though the original file had a `.jpg` extension.

I think the Panel should be a bit more generous in files that it allows to be uploaded: it should allow the same MIME type _and/or_ the original extension.

**To Reproduce**
1. Export out a JPEG 2000 image, rename the extension from `.jp2` to `.jpg`
2. Upload the `.jpg` to the Panel
3. Attempt to replace the file with a totally normal `.jpg` file

**Expected behavior**
The file picker should allow either `.jp2` _or_ `.jpg` files.

**Kirby Version**
`3.1.4-rc.1`

distantnative:
## Describe the PR
- the upload input only allows you to pick a file with the same extension OR with the same mime type
- the backend only throws an exception when extension and mime are different

## Related issues

- Fixes #1783

## Ready?

- [x] Passing all unit tests


bastianallgeier:
I wonder if this is a good idea on the PHP side. What happens if someone uploads a PHP file camouflaged as .jpg?

distantnative:
> A "JPEG 2000" image was accidentally uploaded to the Panel, but it the original file had the extension .jpg instead of the more appropriate .jp2. However, once it was uploaded, the Panel would only accept uploads that had the MIME type of image/jp2. The file picker wouldn't allow a replacement .jpg file, even though the original file had a .jpg extension.

To prevent such an issue in another way, we would have to forbid uploads where extension doesn't match the mime type. Not sure if that's much better.

lukasbestle:
If the user uploads a PHP file with a .jpg extension, that shouldn’t be an issue. The web server will determine whether to run the PHP code based on the extension, not based on the MIME type of the file contents.

The result will be that there‘s a broken JPG, but it‘s not a security issue. Or isn‘t that what you meant @bastianallgeier?

bastianallgeier:
@lukasbestle you are right. I was thinking too complicated.

bastianallgeier (approved):

PR #2363, opened 2019-12-15
Add language icon

distantnative:
## Describe the PR
Apparently quite a few developers already do this via a workaround, I think it would be nice to support it in the core. 

Screen Shot 2019-12-15 at 18 23 32

Screen Shot 2019-12-15 at 18 23 42


panel/src/components/Views/SettingsView.vue
```diff
@@ -112,7 +112,11 @@ export default {
             return {
               id: language.code,
               default: language.default,
-              icon: { type: "globe", back: "black" },
+              icon: {
+                type: language.icon || "globe",
+                back: "black",
+                emoji: this.$helper.isEmoji(language.icon)
```
distantnative, line 118: Actually this line is not needed anymore...

bastianallgeier:
To be honest with you, this PR is really difficult for me. On one hand this is quite useful and not a massive change, but on the other hand it's such bad practice to use flags to represent languages.

distantnative:
To be honest, this one is quite easy for me: I don't like it anymore. Let's throw it away.

distantnative:
@bastianallgeier only maybe rescue the isEmoji js helper from this?

PR #2376, opened 2019-12-22
Update PHP version support checks

lukasbestle:
Kirby 3.4.0+ supports PHP 7.2 to 7.4. Versions below 7.2 or above 7.4 (= 7.5.0 or greater) are not supported.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -2,9 +2,12 @@
 
 /**
  * Validate the PHP version to already
- * stop at older versions
+ * stop at older or too recent versions
  */
-if (version_compare(phpversion(), '7.1.0', '>=') === false) {
+if (
+    version_compare(PHP_VERSION, '7.2.0', '>=') === false ||
+    version_compare(PHP_VERSION, '7.5.0', '<')  === false
+) {
     die(include __DIR__ . '/views/php.php');
 }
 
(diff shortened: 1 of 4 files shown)


afbora:
Not related to the php version, but shouldn't we add the curl plugin while composer.json is updating as `ext-curl`? CURL using on license registration, right? Btw we need add gd library as `ext-gd`?

lukasbestle:
@afbora Kirby also requires other extensions like mbstring. However I think we shouldn‘t list them there for two reasons:

- The required extensions depend on the feature set that is being used. If the user doesn’t use a Kirby feature that requires an extension, it shouldn‘t be required.
- Sometimes the PHP installation on the command line differs from the one used for the webserver. We would have quite a lot of reports of installation errors because of that.

afbora:
@lukasbestle I understand. In fact, i realized missing extensions because of that `mbstring` and `ctype` already listed in composer file.

lukasbestle:
You are right, I forgot that we had those in there. It looks like they didn‘t cause issues so far, so we can probably leave them there. They are also the two extensions that are pretty much always required for Kirby to work.

Maybe we could add additional extensions as `suggest` dependencies. I‘m not sure if Composer prints those suggestions if the extension is available though (which would be pretty annoying).

bastianallgeier:
Is there a reason why this is targeted at the master branch?

lukasbestle:
@bastianallgeier There's always a reason. In this case it's that I didn't check when creating the PR. 😛
Fixed now.

PR #2392, opened 2020-01-13
New User::roles() method and API route for better user role dialogs

### LINKED ISSUE #2380: If there is only one admin user, this admin cannot change the role of non-admin users to admin
**Describe the bug**  
As it says in the title.

**To Reproduce**  
1. Create a new user role, e.g. client.
2. Create a user with this role.
3. Then, try to change the role of this user to `admin`.
4. The `changeRole` dialog in this case doesn't offer any other roles to change to apart from the current role.

**Expected behavior**  
It should be possible to change the role of an existing user to `admin`. 

**Additional context**
Note that it is possible to create new admin users. It is also possible to change the role of a non-admin user to `admin` if you create a second admin user first.

**Kirby Version**  
Tested with 3.2.2

bastianallgeier:
## Describe the PR

This PR introduces a new `User::roles()` method, which returns all available roles for the user, depending on the authenticated user and their privileges. This is used in the user role dialog to get role options more reliably. 

## Related issues

- Fixes #2380 

## Ready?

- [x] Passing all unit tests


panel/src/components/Dialogs/UserRoleDialog.vue
```diff
@@ -47,9 +47,11 @@ export default {
 
       this.$api.users.get(id)
         .then(user => {
-          this.$api.roles.options({ canBe: "changed" }).then(roles => {
+          this.$api.users.roles(id).then(roles => {
             this.roles = roles;
 
+            console.log(roles);
```
distantnative: Remove before merging

bastianallgeier: Ugh, thanks for spotting this 🙈

distantnative:
Otherwise looks good to me

distantnative (approved):

afbora:
@bastianallgeier Everything works great except role `permissions`.

I've created two new roles as `Client` as `Editor` with following permissions.

````yaml
permissions:
    user: true
    users: true
````

Clients and Editors can't change own role (user: true) or another user role (users: true) now.
Changing own role working in last version as 3.3.2 but changing another users role not working too.

distantnative:
@afbora Good point!

If we have the permission, we need to allow this actually.

lukasbestle:
Now that I think about it, we should probably warn about the `changeRole` permission. If enabled for a role that is not `admin`, users with that role will be able to circumvent all permissions by creating a new user and changing its role to a role with more permissions or even `admin`.

But yes: If we decide to keep this permission setting, it should be supported by the role changing logic.

bastianallgeier:
This is the reason, why I ditched the permission at all for now and I really think we should probably keep it like this. Otherwise there's room for some really dangerous mistakes.

distantnative:
@bastianallgeier Actually I agree that I would prefer it the way of this PR and only allow admins to change roles. However, maybe we need to delay this PR to 3.4 then, since it might break existing setups where others can change roles too?

texnixe:
What I would like to see:

1. Users should never be able to change their own roles
2. Non-admin users should be able to change other user's roles, but the options should not be `true/false`, but `false` by default and instead of `true` an explicit list of roles they can change (and change to), always excluding admin role.

afbora:
I agree with @texnixe . The process of changing the user's own role did not make me very functional and logical. But I think the user should have the authority to change the role of other users if have the permission.

bastianallgeier:
I decided to merge this as a first step for 3.4 and then later we can improve this further.

PR #2399, opened 2020-01-18
Clear lock if user not found for locked page #2333

### LINKED ISSUE #2333: .lock files sometimes do not get removed resulting in editors being unable to edit pages
**Describe the bug**  
Just posting this here for reference because not sure if this is a bug or some handling error.

There have been several  user reports, where editors were no longer able to edit individual pages due to `.lock` files not being removed by the system. The result was that all characters entered into a field were immediately removed again.

**To Reproduce**  
I wasn't able to reproduce such a situation myself.

**Kirby Version**  
?

**Console output**  
See link below.

**Additional context**  
I guess in some cases, lock files might have been deployed by accident, but we have also seen this happening on localhost.

Last issue report on forum: https://forum.getkirby.com/t/i-cant-add-any-text-or-file/16510

afbora:
## Describe the PR

I think the issue is because the page locked by another user that cannot find the user.
If you think the issue  is not caused by this, I must say that this PR definitely resolves a bug :))

Fixed the issue with clearing lock data if user not found for that page. 
I could not see the appropriate method for remove/clear all lock data and created a new `clear` method that remove `lock` data without any conditions.

## Related issues

- Fixes #2333 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/ContentLock.php
+++ b/src/Cms/ContentLock.php
@@ -45,7 +45,7 @@ class ContentLock
      *
      * @return bool
      */
-    public function clear(): bool
+    protected function clearLock(): bool
     {
         // if no lock exists, skip
         if (isset($this->data['lock']) === false) {
@@ -105,7 +105,7 @@ class ContentLock
             }
 
             // clear lock if user not found
-            $this->clear();
+            $this->clearLock();
         }
 
         return false;
@@ -169,10 +169,7 @@ class ContentLock
             ]);
         }
 
-        // remove lock
-        unset($this->data['lock']);
-
-        return $this->kirby()->locks()->set($this->model, $this->data);
+        return $this->clearLock();
     }
 
     /**
@@ -212,10 +209,7 @@ class ContentLock
         $this->data['unlock']   = $this->data['unlock'] ?? [];
         $this->data['unlock'][] = $this->data['lock']['user'];
 
-        // remove lock
-        unset($this->data['lock']);
-
-        return $this->kirby()->locks()->set($this->model, $this->data);
+        return $this->clearLock();
     }
 
     /**
--- a/src/Cms/ContentLock.php
+++ b/src/Cms/ContentLock.php
@@ -41,7 +41,7 @@ class ContentLock
     }
 
     /**
-     * Clear lock
+     * Clears the lock unconditionally
      *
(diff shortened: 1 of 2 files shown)


src/Cms/ContentLock.php
```diff
@@ -40,6 +40,24 @@ public function __construct(ModelWithContent $model)
         $this->data  = $this->kirby()->locks()->get($model);
     }
 
+    /**
+     * Clear lock
+     *
+     * @return bool
+     */
+    public function clear(): bool
```
lukasbestle: - This should be a protected method as it should never be possible to clear a lock unconditionally from the outside. Only this class may do that.
- The `remove()` and `unlock()` methods should use this method internally to avoid code duplication.
- It may make sense to rename this to `clearLock()` to be more ... clear. 😛

lukasbestle (commented):
That's a very helpful fix – thanks a lot for debugging this!

It only needs a few minor changes and a unit test in `tests/Cms/Content/ContentLockTest.php`.

src/Cms/ContentLock.php
afbora: Finding the right name for methods kills me :)) 

It was my hesitation when I wrote, thank you, I will resolve it.

afbora: @lukasbestle Changes done but i cant add unit test with success. I'm getting following error always. Could you help about that please 😇 

> Kirby\Cms\ContentLockTest::testClearLock
> Kirby\Exception\DuplicateException: /test is already locked

Tried following steps on test:

1. impersonate A
2. create lock
3. get current lock
4. remove user A
5. impersonate B
6. get lock to clear
7. create lock
8. get current lock

lukasbestle: I've simplified the test a bit and now it works. :)

lukasbestle (approved):

src/Cms/ContentLock.php
afbora: You're great, thank you so much 👏 
My unit test looks like spaghetti pasta next to yours 😅

lukasbestle: Thank you for your help! ❤️

bastianallgeier (approved):

PR #2404, opened 2020-01-21
Fix deleting tag with backspace

### LINKED ISSUE #2403: Backspace key in empty tags field makes it lose focus
**Describe the bug**  

Within a tag field, hitting backspace when the field is empty de-focuses it.

**To Reproduce**  
1. Create a tags field.
2. Add some tags. (optional)
3. Press backspace until the field is empty and then once more.
4. Field loses focus.

**Expected behavior**  
Expected behaviour would be the field remains focused. AFAIK no other field behaves like that. I found myself having to switch to mouse to focus the field again many times. 

**Kirby Version**  
Version
3.3.1

 - OS: linux mint cinnamon 19.2
 - Browser: chrome
 - Version: Versión 79.0.3945.130 (Build oficial) (64 bits)

afbora:
## Describe the PR

Fix backspace and randomly (not in order) deleting  issue with checking tags length.

## Related issues

- Fixes #2403

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/panel/src/components/Forms/Input/TagsInput.vue
+++ b/panel/src/components/Forms/Input/TagsInput.vue
@@ -228,10 +228,21 @@ export default {
 
       switch (position) {
         case "prev":
+          if (!this.selected) return;
+
+          currIndex = this.index(this.selected);
+          nextIndex = currIndex - 1;
+
+          if (nextIndex < 0) return;
+          break;
+
         case "next":
           if (!this.selected) return;
+
           currIndex = this.index(this.selected);
-          nextIndex = position === "prev" ? currIndex - 1 : currIndex + 1;
+          nextIndex = currIndex + 1;
+
+          if (nextIndex >= this.tags.length) return;
           break;
 
         case "first":
@@ -275,22 +286,22 @@ export default {
     leaveInput(e) {
       if (
         e.target.selectionStart === 0 &&
-        e.target.selectionStart === e.target.selectionEnd
+        e.target.selectionStart === e.target.selectionEnd &&
+        this.tags.length !== 0
       ) {
-        this.navigate("last");
         this.$refs.autocomplete.close();
+        this.navigate("last");
         e.preventDefault();
-        e.target.blur();
       }
     },
     navigate(position) {
       var result = this.get(position);
       if (result) {
         result.ref.focus();
-        this.selected = result.tag;
+        this.selectTag(result.tag);
       } else if (position === "next") {
         this.$refs.input.focus();
-        this.selected = null;
+        this.selectTag(null);
       }
     },
     prepareTags(value) {
@@ -319,10 +330,12 @@ export default {
       this.onInput();
 
       if (prev) {
+        this.selectTag(prev.tag);
         prev.ref.focus();
       } else if (next) {
-        next.ref.focus();
+        this.selectTag(next.tag);
       } else {
+        this.selectTag(null);
         this.$refs.input.focus();
       }
     },


panel/src/components/Forms/Input/TagsInput.vue
```diff
@@ -247,6 +247,10 @@ export default {
           break;
       }
 
+      if (nextIndex < 0) {
+        nextIndex = 0;
+      }
```
afbora: Make sure not negative index

panel/src/components/Forms/Input/TagsInput.vue
```diff
@@ -280,7 +284,10 @@ export default {
         this.navigate("last");
         this.$refs.autocomplete.close();
         e.preventDefault();
-        e.target.blur();
+
+        if (this.tags.length !== 0) {
+          e.target.blur();
+        }
```
afbora: Solution by @distantnative to prevent losing focus

panel/src/components/Forms/Input/TagsInput.vue
```diff
@@ -318,7 +325,9 @@ export default {
       this.tags.splice(this.index(tag), 1);
       this.onInput();
 
-      if (prev) {
+      if (this.tags.length === 0) {
+        this.$refs.input.focus();
+      } else if (prev) {
```
afbora: Focus input if no tags remain

distantnative: Mmm something else is weird then. We should not need this. We test first if there is a previous tag, if there is a next tag and the last else would already focus the input. If no tags remain, `prev` and `next` should already not match and the `else` part should be applied.

panel/src/components/Forms/Input/TagsInput.vue
distantnative: I'm wondering if we need this at all... `this.navigate("last")` should already focus the last existing tag... and therefore the input should be blurred already anyways. Maybe we could remove that line completely? We would need to test it.

panel/src/components/Forms/Input/TagsInput.vue
distantnative: Maybe we should add a similar check for not getting an index higher than the length.

panel/src/components/Forms/Input/TagsInput.vue
afbora: Because we get the previous and next neighbors before deleting tag, we have to make sure that no tags remain after deleting:

https://github.com/getkirby/kirby/blob/73de9f5e087157f992f2ca777feebe17f451c0dc/panel/src/components/Forms/Input/TagsInput.vue#L321-L322

distantnative: Yes, but if there was a previous tag before deleting, then the length can't be 0 (cause that prev tag should still be there, right).

afbora: No, if there is a previous neighbor, there are already 2 tags. So length after delete is `1`  :) If there is only one tag, there is no neighbor.

distantnative: Exactly. And we already have

```js
if (pref) {
  …
} else if (next) {
  …
} else {
 // focus on input 
}
```

So I don't think we need your extra lines. These should already cover the case: If one is before, if one is after, else (none before, none after = 0 in length).

If this is not working, then something is wrong with getting `prev` and `next`

afbora: I understand you're right. But there was an error in that part but I will test it again and share the details with you.

afbora:
@distantnative The indexes seem to interlock when removing but I couldn't solve. For example, it should take the first index as `0`, but it takes another index.

https://github.com/getkirby/kirby/blob/3.3.3/panel/src/components/Forms/Input/TagsInput.vue#L312-L328

distantnative:
I dug a little deeper and came up with slightly different solutions. Would be great if you could help testing @afbora - so easy to overlook cases.

afbora:
@distantnative It works a little better than perfect 😄 👍 Thank you so much. Please ignore my soup codes 😇

PR #2405, opened 2020-01-22
Cache blueprint extends

### LINKED ISSUE #2364: Blueprint extends with caching?
If many fields are extended by blueprint extends (as is often the case with the builder field by Tim Ötting) the Kirby backend becomes very slow. Even if no data is called up and is available and the field is completely empty, the process takes from about 5 seconds to the server timeout. Bastian talks about a way to cache the extends.

More information about the problem here:

https://forum.getkirby.com/t/asynchronous-files-access-for-blueprint-extension/16520

afbora:
## Describe the PR

Blueprint extends are now cached by their name.

However, since the `Blueprint` class is statically called in the tests (_tests/Cms/Blueprints/BlueprintFieldTest.php_), it does not match the data in the cache and the test fails. But in my local test on panel ui works great! Should we refactor the test class (BlueprintFieldTest) or the cache system I wrote was faulty 😇?
I'm creating thr PR as a draft because I'm not sure and waiting your feedback please.

## Related issues

- Fixes #2364 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/Blueprint.php
+++ b/src/Cms/Blueprint.php
@@ -198,18 +198,11 @@ class Blueprint
             return $props;
         }
 
-        $mixin = static::find($extends);
-
-        if ($mixin !== null) {
-            if (is_array($mixin) === true) {
-                $props = A::merge($mixin, $props, A::MERGE_REPLACE);
-            } else {
-                try {
-                    $props = A::merge(Data::read($mixin), $props, A::MERGE_REPLACE);
-                } catch (Exception $e) {
-                    $props = $props;
-                }
-            }
+        try {
+            $mixin = static::find($extends);
+            $props = A::merge($mixin, $props, A::MERGE_REPLACE);
+        } catch (Exception $e) {
+            // keep the props unextended if the snippet wasn't found
         }
 
         // remove the extends flag
@@ -271,7 +264,7 @@ class Blueprint
      * @param string $name
      * @return array
      */
-    public static function find(string $name)
+    public static function find(string $name): array
     {
         if (isset(static::$loaded[$name]) === true) {
             return static::$loaded[$name];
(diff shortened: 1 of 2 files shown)


lukasbestle:
To be honest, I think we should refactor the loading code to avoid having to keep two separate caches (`$loaded` and `$extends`):

- `Blueprint::find()` should always return the blueprint data, not either the data or the file path string. Basically the case where the file exists in the `blueprints` dir should run `Data::read()` and return the array as well.
- With that simplified, the two only callers of that method I can find (`Blueprint::load()` and `Blueprint::extend()`) can be massively simplified to only care for array data. The `if`s for array vs string are no longer required.
- The existing in-memory caching inside `Blueprint::load()` can then be moved to `Blueprint::find()` so that all loaded blueprint files and mixins are cached in the same `$loaded` cache.

lukasbestle (changes requested):
To be honest, I think we should refactor the loading code to avoid having to keep two separate caches (`$loaded` and `$extends`):

- `Blueprint::find()` should always return the blueprint data, not either the data or the file path string. Basically the case where the file exists in the `blueprints` dir should run `Data::read()` and return the array as well.
- With that simplified, the two only callers of that method I can find (`Blueprint::load()` and `Blueprint::extend()`) can be massively simplified to only care for array data. The `if`s for array vs string are no longer required.
- The existing in-memory caching inside `Blueprint::load()` can then be moved to `Blueprint::find()` so that all loaded blueprint files and mixins are cached in the same `$loaded` cache.

afbora:
@lukasbestle I made a few attempts, but even though the panel was working, it started to give a lot of errors in the tests. I think I make a mistake in the data types returned. Can I ask you for help 😇 ? If you want, you can delete this PR and open it yourself. Sorry for not being able to apply the changes 🙏

lukasbestle:
Feel free to push your current state to this PR and I will take over. :)

afbora:
Thank you so much. That's it from me 😅

lukasbestle (approved):

lukasbestle:
I love it when you can fix bugs by removing more code than adding code. 🎉

afbora:
@lukasbestle I admire you to simplify and solve problems :heart:

afbora:
@lukasbestle  I was curious about something, I noticed that while testing cache on the final codes. The cache works great, but when I tried to access a page, I saw 4 different requests for that page. And since each request is independent of each other, the cache process is valid only for that request.

Can we make the cache process more efficient? Like writing ready blueprint props as json to the `/cache` folder 💡 

Is this a minor performance concern? 🤔 

**Requests on page**

1. /api/pages/sandbox?view=panel
2. /api/pages/sandbox/sections/pages?page=1
3. /api/pages/sandbox/sections/main-col-0-fields
4. /api/pages/sandbox/sections/files?page=1

lukasbestle:
Caching the props to file won't increase performance unfortunately as the operation that uses the most time is loading the blueprints from file, not processing/parsing them. If we would cache the data to a JSON file, we would need to find, open and parse that as well, which wouldn't be any faster.

The only thing we could do is to cache fully extended blueprints, which would increase performance for blueprints with a lot of different extends (fewer separate files to open and parse). However that would substantially increase complexity as the cache would need to be invalidated if any of the blueprint files changes – which needs more code and also takes processing time to check. So I don't think this would be worth it.

afbora:
I understood properly now, thank you. You are right, even if we write to the cache folder, we will do the same operations on `cache` files instead of `yml` files 👍 OK, the present way is fine 😉

bastianallgeier (approved):

PR #2406, opened 2020-01-22
Accessing with default language slug on multilang

### LINKED ISSUE #2257: In a multi-language setting, a non-default page version can be accessed both with the default language slug and its own translated slug.
**Describe the bug**  
Already done in the title

**To Reproduce**  
1. Create a multi-language Starterkit with two languages
2. Change the slug of a page in the non-default language
3. Access the page on the frontend.
4. Then change the page slug to the original default language slug in the address bar
5. Instead of showing the error page, the same page is still shown.

**Expected behavior**  
The same page should not be available under two URLs.

**Screenshots**  

**Kirby Version**  
Tested with 3.3.0-rc-3 but also happens in 3.2.5 (other versions not tested yet, didn't happen in older versions, though, as far as I remember).

afbora:
## Describe the PR

Fixed the issue with separated finding page in collection based on multilingual option.

## Related issues

- Fixes #2257 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -243,11 +243,11 @@ class Pages extends Collection
         $query      = $startAt;
 
         foreach ($path as $key) {
-            $query = ltrim($query . '/' . $key, '/');
-            $item  = $collection->get($query) ?? null;
-
-            if ($item === null && $multiLang === true) {
+            if ($multiLang === true) {
                 $item = $collection->findBy('slug', $key);
+            } else {
+                $query = ltrim($query . '/' . $key, '/');
+                $item  = $collection->get($query);
             }
 
             if ($item === null) {


lukasbestle (dismissed):
Awesome fix!

distantnative:
Haven't tested it myself yet, but from reading the code: What happens if we are in a multilang setup, but there is not `slug` entry in the content file cause the slug has not been translated at all?

afbora:
@distantnative Ahh, good point! Opening default page if page not translated yet. The bigger problem is that you cannot access the page of that language on the panel. I'll take a look again.

afbora:
I have done a few tests now. Since the front side and the panel use the same methods, if we prevent it from opening at the front, it does not open at the panel. We may need to find a way to separate the two sides 🤔

afbora:
Now we know the issue better so i'm closing for now to find stable solution.

PR #2410, opened 2020-01-25
Fix decimals float on number field #2372

### LINKED ISSUE #2372: Number field cuts of decimal places
To my surprise the number field doesn't work any more if more then 4 decimal places are used.

**Working:**

```
                lat:
                    label:        Breitengrad
                    type:         number
                    step:         0.0001
```

**Not working anymore:**

```
                lat:
                    label:        Breitengrad
                    type:         number
                    step:         0.00001
```

If you try with 5 or more decimal places, all decimal places are silently dropped (you loose everything following the comma, making the field an integer-only field).

If this is expected, it doesn't make sense to me, because decimal numbers are not limited to 4 decimal places. In some situations - like with latitudes or longitudes - we need more then 4 decimal places in the numbers. What is the best way to store such numbers, if not in the numbers field?

afbora:
## Describe the PR

This issue was about casting specific float number types (like 1.0E-6) to string. The problem was difficult to find, but easy to solve 😅 

## Related issues

- Fixes #2372 

## Ready?

- [x] Passing all unit tests


lukasbestle (commented):
This doesn't fix the original issue for me – the float is later converted to string, which will again result in the exponential syntax that won't be understood by the browser.

afbora:
@lukasbestle Original issue? I tested on panel with `step: 0.00001` option before create PR and worked properly. I have tested now and step option works great 🤔

afbora:
Btw scientific notation is supported all major browsers

https://caniuse.com/#feat=mdn-css_types_number_scientific_notation

afbora:
As I mentioned in my first message, the source of the error was due to the insertion of the scientific notation type into the `str_replace()` method in `Str::float()` method.

lukasbestle (dismissed):
You are right, it *does* fix the original issue. Seems like I tested with a few too many zeroes last time, because `0.0000001` still results in the following error in the Panel:

```
Cannot read property '$invalid' of null
```

But that seems to be unrelated to this issue as the API response for such a `step` definition is correct and the error occurs only on the frontend.

afbora:
Yes, it gives this error in much smaller numbers. Actually similar errors. The error we solved was related to the backend, this error is on the javasript side. It seems like I have solved the error on the javascript side, should we continue under this PR?

Issue about javascript side about casting string:
https://github.com/getkirby/kirby/blob/3.3.3/panel/src/components/Forms/Input/NumberInput.vue#L87

lukasbestle:
Oh yes, that's exactly the same issue in the frontend.

> It seems like I have solved the error on the javascript side, should we continue under this PR?

If you have an easy fix for this: Sure!

afbora:
@lukasbestle I have fixed the issue on javascript side. It works successfully in my tests.

@distantnative @bastianallgeier There may be more descriptive method and variable names 😇  I may have made incorrect or incomplete use in `number`, `value`, and `step` assignments. I would appreciate if you fix it or make improvements. Thank you 🙏 

**Reference solution**
http://blog.davidjs.com/2018/07/convert-exponential-numbers-to-decimal-in-javascript/

bastianallgeier:
I can't believe how much pain a simple number field already cost us. The convert method seems quite extensive :( I wish there was a way to solve this with less code.

afbora:
@bastianallgeier What do you think now?

afbora:
I'll do some improvements

afbora:
@bastianallgeier Ready to review 👍

panel/src/components/Forms/Input/NumberInput.vue
```diff
@@ -84,6 +85,10 @@ export default {
         return 0;
       }
 
+      if (step.toString().indexOf('e') !== -1) {
+        return parseInt(step.toFixed(16).split(".")[1].split("").reverse().join("")).toString().length;
+      }
```
afbora, line 90: Steps of method that i found:

1. Check float: 1e-6
2. Fix with 16 decimals as `0.0000010000000000`
3. Get only decimals as `0000010000000000`
4. Reverse the number: `0000000000100000`
5. ParseInt as `100000`
6. Get length as `6`

Wohooo 😄

bastianallgeier (approved):

PR #2414, opened 2020-01-27
Update dependencies and avoid exact version constraints

neildaniels:
## Describe the PR

Kirby's `composer.json` file currently lists exact versions of multiple dependencies, instead of more forgiving version constraints. Because Kirby expects an _exact_ version of dependency, if some other required package insists on a newer version there will be an unsatisfiable conflict.


lukasbestle:
The reason why we use exact dependency versions is that we have made the experience that some libraries don‘t follow semantic versioning and may therefore include breaking changes in patch releases. For a CMS like Kirby this means that installations of a specific Kirby release (which are expected to work for a longer time period) may break in the future. Debugging this would be hard as we wouldn‘t know which exact library versions the user has unless they tell us that when reporting a bug/seeking for help in the forum.

We update the library versions regularly and in the process do the necessary tests to verify that everything still works.

I understand that this causes issues like in your case with another dependency that requires a later version, but we unfortunately can‘t really fix this in general.

neildaniels:
I’d understand if `composer.lock` and `vendor` we’re reverted. That be the “QA’d" version that’s non-Composer installations would get.

But I think it would be exceptionally helpful if `composer.json` remained flexible. I assume that most people installing Kirby via Composer expect this flexibility and accept its inherent “risk.”

It’s odd to me that Composer support has basically been reduced to act like primitive git submodules.

If there are actually dependencies that are reckless with versioning, then they probably should be reconsidered for inclusion.

lukasbestle:
Sorry for the late reply!

> I assume that most people installing Kirby via Composer expect this flexibility and accept its inherent “risk.”

I'm not sure about that. Advanced users most likely are, but not everyone who uses Kirby with Composer is an advanced user.

The issue is: If something goes wrong, we are the ones who will do the support and who need to debug such issues. If Kirby was a hobby-type open-source project, we could say "not our problem, use this at your own peril", but that's not what we aspire to. As Kirby is a professional tool with professional support, we cannot afford that attitude and therefore need to make sure that Kirby is as robust as possible.

> It’s odd to me that Composer support has basically been reduced to act like primitive git submodules.

The Composer philosophy differs a lot from ours and our requirements (some of which I explained above) vastly differ from a typical PHP project that either is such a hobby-type open-source project or the other extreme, a professional product that runs in a controlled environment.

We therefore already had to make quite a few trade-offs when implementing Composer support in Kirby 3 (like including the vendor folder in the repo, using a custom installation path etc.). We did it anyway, as Composer has several benefits for those who want to use it.

But you are right: The trade-offs mean that we can't use the full potential of Composer. I assure you that we did the best we could do given the constraints we have to work with.

> If there are actually dependencies that are reckless with versioning, then they probably should be reconsidered for inclusion.

Life is not always that simple:

- You never know in advance which dependency will misbehave in the future. You will only know it when it's too late.
- Often there are not too many alternatives and they each have their own disadvantages.

We carefully choose our dependencies and also try to use as few of them as possible in general. This still leaves the remaining risk of a future compatibility issue.

---

In conclusion: I understand your issue and I'm sorry that we cannot "just" merge this PR. I hope that my explanation could help you understand why. :)

lukasbestle:
See #2417.

Issue #1308, opened 2019-01-02
[Panel] Textarea in structure field: Spaces are deleted on save.


tobiasfabian:
**Describe the bug**
(Indenting) spaces (or tabs) at the beginning or the end of a line are deleted on save in textareas within structure fields.

If the textarea is not within a structure field spaces are preserved.

I have this structure field.
```yaml
webComponents:
  type: structure
  fields:
    html:
      label: HTML
      type: textarea
      buttons: false
```

This is my input:
```
<div class="text-box">
  <div class="text-box__heading">
    <h2>Headline</h2>
  </div>
  <div class="text-box__content">
    {{ Content }}
  </div>
</div>
```

This is how it’s saved (`default.txt`)
```text
----

Webcomponents: 

- 
  html: |
    <div class="text-box">
    <div class="text-box__heading">
    <h2>Headline</h2>
    </div>
    <div class="text-box__content">
    {{ Content }}
    </div>
    </div>
```
**Expected behavior**
It should be saved like this:
```txt
----

Webcomponents: 

- 
  html: |
    <div class="text-box">
      <div class="text-box__heading">
        <h2>Headline</h2>
      </div>
      <div class="text-box__content">
        {{ Content }}
      </div>
    </div>
```

**Kirby Version**
3.0.0-RC-3.0

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Safari
 - Version 12.0.2

tobiasfabian:
I did some research and found out that this is a ['bug'](https://github.com/mustangostang/spyc/issues/54) in [spyc](https://github.com/mustangostang/spyc).

[symfony/yaml](https://github.com/symfony/yaml) could be an alternative.

jenstornell:
Or maybe have as a component. Then it can be replaced by any library.

bastianallgeier:
I worked with symfony/yaml at the beginning of v3 and it had many other issues because of being a lot more strict that spyc.

Wemago:
Have you seen / tried [Dipper](https://github.com/secondparty/dipper) ?
The only problem is that it isn't maintained for a while.

TimOetting:
I give a +1 for this one. I have the same problem when I have a textarea inside a block of my [Kirby Builder](https://github.com/TimOetting/kirby-builder) plugin. I am sure that this is related. 

I would like to use the Kirby Builder for different segments in my pages that also contain code samples. Because of this bug these code samples lose their indentation 😕

[...]

PR #2417, opened 2020-02-01
Update dependencies

lukasbestle:
I have verified with the changelogs that there shouldn't be any breaking-changes in the updated dependencies. It's probably still better to ship this change with Kirby 3.4.0.


afbora:
For information purposes: note that we will remove the Spyc repository in version 3.4.0 #1308

PR #2425, opened 2020-02-04
Pass $used flag to route:after hook #2418

### LINKED ISSUE #2418: `route:after` should not be called on `$route→next()`
I just encountered the problem while debugging Retour (https://github.com/distantnative/retour-for-kirby/issues/138) that the `route:after` hook is a bit problematic in multilang setups.

I would argue the most common use case to hook into a successful routing.
However, the hook is actually called on every route lookup - also if the route does `$this→next()`, which seems counter intuitive as the first route explicitly says "not me" by calling `$this→next()`
https://github.com/getkirby/kirby/blob/master/src/Http/Router.php#L115-L117

My suggestion would be to move the hook call out of the loop and right before the return. To really only call the hook on routes that do not skip to the next route.
What do you think @bastianallgeier and @lukasbestle?

distantnative:
## Describe the PR

Add `$used` flag as new last parameter to the `route:after` hook that represents whether the route was actually used or `$this→next()` was called inside the route.

## Related issues

- Fixes #2418

## Ready?

- [x] Passing all unit tests


distantnative:
Damn, something broke...

distantnative:
@lukasbestle We actually have a small breaking change:
We can't add the new parameter as last one, as `->apply()` always modifies/returns the last parameter.

lukasbestle:
Hm, that's unfortunate.

Maybe we should instead use the opportunity to replace all those hook arguments (`$path, $method, $used` and maybe even `$route`) with one `RouteRequest` object that carries all the information. That would probably be cleaner in the long run.

Of course this would be something for 3.4 then, but I'd say it would be better to have this breaking-change than to change the param order. The issue with the new param order is that it won't make existing code fail, it would only make it not work properly. If we instead cut down on the arguments, existing hooks will fail hard because they aren't passed enough arguments.

bastianallgeier:
I changed the milestone

lukasbestle:
Requires #2541 for a proper implementation.

bastianallgeier:
What's the state of this @distantnative & @lukasbestle? Is it compatible with #2541 ?

lukasbestle:
@bastianallgeier Now that you have merged the other PR, we can update this to work with the changed structure. I will finalize this PR soon.

lukasbestle:
I have now rebased this PR against `develop` to pull in the changes from #2541.

Because the new `$app->apply()` implementation now allows modifying an argument that is not the last one, I was able to move the new `$used` argument to last position for maximum backwards-compatibility.

But because hooks can now request just the arguments they need, the order doesn't matter in the end – it's just relevant for the underlying `Router::$afterEach` callback where the argument order *does* matter.

distantnative:
❤️

I started calling it `$used` myself, but maybe `$final` would be the better name.

lukasbestle:
@distantnative I agree. Feel free to change it. :)

bastianallgeier:
@distantnative @lukasbestle $final sounds more intuitive to me as well.

lukasbestle:
@distantnative Could you please make the change? As I'm not the author of the PR, I don't have full access to the Travis build etc.

distantnative:
Will do

distantnative:
Done, ready.

lukasbestle (approved):
👍

bastianallgeier (approved):

PR #2446, opened 2020-02-12
Remove buttons from upload dialog #2327

### LINKED ISSUE #2327: File Upload modal: "Ok" button without a function
**Describe the bug**  
When uploading a file to a files section, a modal window pops up showing an "Ok" button – but clicking "Ok" doesn't do anything.

**To Reproduce**  
1. Go to a files section
2. Upload a file
3. Click on "Ok"

**Expected behavior**  
I would expect the modal to disappear when clicking "Ok", performing the upload in the background while I continue my work. If this is not possible, I would expect "Cancel" to be the only available option (and omitting the "Ok" button).

**Screenshots**  

grafik

**Kirby Version**  
3.3.0

**Console output**  
–

 - OS: macOS Catalina
 - Browser: Firefox
 - Version: 70.0.1

distantnative:
## Related issues

- Fixes #2327

### DIFF
--- a/panel/src/components/Forms/Upload.vue
+++ b/panel/src/components/Forms/Upload.vue
@@ -31,14 +31,17 @@
           </li>
         </ul>
       </template>
+
       <template slot="footer">
         <template v-if="errors.length > 0">
           <k-button-group>
+            <k-button :disabled="true" />
             <k-button icon="check" @click="$refs.dialog.close()">
               {{ $t("confirm") }}
             </k-button>
           </k-button-group>
         </template>
+        <div v-else />
       </template>
     </k-dialog>
 
@@ -95,6 +98,7 @@ export default {
     },
     upload(files) {
       this.$refs.dialog.open();
+      this.$store.dispatch("isLoading", true);
 
       this.files = [...files];
       this.completed = {};
@@ -134,6 +138,7 @@ export default {
       if (Object.keys(this.completed).length == this.total) {
         // remove the selected file
         this.$refs.input.value = "";
+        this.$store.dispatch("isLoading", false);
 
         if (this.errors.length > 0) {
           this.$forceUpdate();
--- a/panel/src/components/Navigation/Topbar.vue
+++ b/panel/src/components/Navigation/Topbar.vue
@@ -94,7 +94,7 @@
           <template v-if="notification">
             <k-button
               class="k-topbar-notification k-topbar-signals-button"
-              theme="positive"
+              :theme="notification.type"
               @click="$store.dispatch('notification/close')"
             >
               {{ notification.message }}
@@ -370,19 +370,10 @@ export default {
   font-weight: $font-weight-bold;
   line-height: 1;
   display: flex;
+  color: $color-light-grey;
 }
-.k-topbar .k-button[data-theme="positive"] {
+.k-topbar .k-button[data-theme="success"] {
   color: $color-positive-on-dark;
-}
-.k-topbar .k-button[data-theme="negative"] {
-  color: $color-negative-on-dark;
-}
-.k-topbar .k-button[data-theme="negative"] .k-button-text {
-  display: none;
-
-  @media screen and (min-width: $breakpoint-small) {
-    display: inline;
-  }
 }
 .k-topbar .k-button[data-theme] .k-button-text {
   opacity: 1;


panel/src/components/Forms/Upload.vue
```diff
@@ -95,6 +95,8 @@ export default {
     },
     upload(files) {
       this.$refs.dialog.open();
+      this.$store.dispatch("isLoading", true);
+      this.$store.dispatch("notification/success", "Uploading...");
```
afbora: Using `upload.progress` translation instead of static text?

distantnative: Of course! Good catch

afbora (commented):
FireShot Capture 028 - Ocean - Mægazine - localhost

Dialog not closing for me 🤔

distantnative:
@afbora are you running it with `npm run dev`? I haven't updated the dist files.

distantnative:
Oh and what did you click? `Ok`? I realize that `Cancel` might not do anything either.

afbora:
Running with `npm run build` directly, i'll test again.

afbora:
@distantnative when I click `Cancel` or `OK`, the dialog closes but I see that it continues to upload in the top right.

distantnative:
@afbora That's fine for `Ok` - that's how it should be. For `Cancel` that seems odd haha.

afbora:
@distantnative I'll say another issue :)) 

If the upload fails, no notification appears after the dialog has closed. The dialog needs to come back or another notification should appear.

afbora:
> Dialog not closing for me 

By the way, I understood that the dialog closes automatically while the upload continues 😅 Sorry, i got it wrong. 

So OK button works as it should 👍 
Only the cancel button does not work and there is no warning when upload fails.

distantnative:
Oh yes, we probably need to open the dialog here again: https://github.com/getkirby/kirby/blob/a8f0ec4e3fcac995faed3c34ef8f83b6ca9f957a/panel/src/components/Forms/Upload.vue#L146

distantnative:
With the cancel button I am not sure... maybe we should just remove it.
Cause it could become a bit messy to stop uploads, but also then revert/delete uploads that have already finished...

afbora:
If we can simply cancel it, I think it can remain like below sample 🤔 
If unforeseen issues will arise, I think we can remove them.

````js
// Can only be canceled if the request is in progress
if (this.uploadRequest) {
  this.uploadRequest.abort();
}
````

**Comment by distantnative** [comment]
Where did you get that piece of code from?

**Comment by afbora** [comment]
@distantnative Just found on internet. Do you think it is possible?

1. https://github.com/pagekit/vue-resource/blob/master/docs/recipes.md#abort-a-request
2. https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort

**Comment by distantnative** [comment]
@afbora Ah no, that would involve so much rewriting. we send a request out for each individual file. You would have to cache those request to abort them, but also check which ones have already finish and then revert/delete those ones...

To me that sounds more of a source for bugs than a benefit. Let's just run the upload without cancellation.

**Comment by distantnative** [comment]
✅

**Comment by afbora** [comment]
@distantnative It works fine now 👍  But there was a new error that we could not anticipate. 

An error occur when you click OK button (mean continues upload progress after dialog close) after switching page. For example, change the page after close the dialog while uploading progress.

**Here console log:**

> app.js:4349 Uncaught TypeError: Cannot set property 'value' of undefined
    at a.complete (app.js:4349)
    at Object.error (app.js:4341)
    at XMLHttpRequest.<anonymous> (app.js:15976)


Here beautified dist files to trace error lines: [dist.zip](https://github.com/getkirby/kirby/files/4203010/dist.zip)

**Comment by distantnative** [comment]
Oh yes switching pages can be an issue of course...
Maybe we cannot allow closing the dialog at all?

**Comment by afbora** [comment]
I actually liked that it continued to upload in the background 😍 
Is it difficult for us to understand if it is on the uploaded page?

**Comment by distantnative** [comment]
I think the main problem is that we could navigate somewhere without any upload component, which also contains the dialog. So if there are any errors, we also couldn’t open/show the dialog anymore cause it doesn’t exist in the DOM anymore.

**Comment by afbora** [comment]
I am not very knowledgeable and experienced in this regard. If we use a global error area like right top in bar, will it be solved? 🤔

**Comment by distantnative** [comment]
I also liked the uploading in the background, but there are actually issues not just with errors but also with uploading when we navigate away from the view with the upload component.

Thus, I decided, to simply remove all buttons from the upload dialog. It should block interaction until all uploads are finished, to avoid navigation.

**Comment by afbora** [comment]
@distantnative Sorry. Github Desktop did something stupid, it didn't break PR, right?

**Comment by afbora** [comment]
@distantnative Meanwhile, the OK button still appears 🤔

**Comment by distantnative** [comment]
@afbora Seems like it indeed overwrote mine with all the old commits.

**Comment by afbora** [comment]
@distantnative Really sorry, could you revert? 😞

**Comment by distantnative** [comment]
Just did. Better delete your local branch and clone it again.

**Comment by afbora** [comment]
Buttons are gone well, but you should also avoid clicking out of the dialog 🙈

**Comment by afbora** [comment]
In the meantime, why we use dialog instead of using the uploaded field area?

**Comment by bastianallgeier** [comment]
I've refactored the dialog component and you can now switch off the cancel and submit button: 

```
:submit-button="false" cancel-button="false"
```

PR #2451, opened 2020-02-13
Format numbers when translating counts

neildaniels:
## Describe the PR
The current `I18n::translateCount()` function does not format numbers, so something like "1,234,567 cars" is merely formatted as `1234567 cars`. This PR corrects this in a locale-correct way (example: German uses a period instead of a comma as a thousands separator.)

## Related issues

- Fixes https://github.com/getkirby/ideas/issues/474

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Toolkit/I18n.php
+++ b/src/Toolkit/I18n.php
@@ -231,7 +231,7 @@ class I18n
      * @param bool $formatNumber
      * @return mixed
      */
-    public static function translateCount(string $key, int $count, string $locale = null)
+    public static function translateCount(string $key, int $count, string $locale = null, bool $formatNumber = true)
     {
         $locale = $locale ?? static::locale();
 
@@ -255,8 +255,12 @@ class I18n
             }
         }
 
-        $formatter = static::decimalNumberFormatter($locale);
-        $count = $formatter->format($count);
+        if ($formatNumber) {
+            $formatter = static::decimalNumberFormatter($locale);
+            if ($formatter !== null) {
+                $count = $formatter->format($count);
+            }
+        }
 
         return str_replace('{{ count }}', $count, $message);
     }
--- a/src/Toolkit/I18n.php
+++ b/src/Toolkit/I18n.php
@@ -87,6 +87,25 @@ class I18n
     }
 
     /**
+     * Format a number
+     *
+     * @param int|float $number
+     * @param string $locale
+     * @return string
+     */
+    public static function formatNumber($number, string $locale = null): string
+    {
+        $locale = $locale ?? static::locale();
+
+        $formatter = static::decimalNumberFormatter($locale);
+        if ($formatter !== null) {
+            $number = $formatter->format($number);
+        }
+
+        return (string)$number;
+    }
+
+    /**
      * Returns the locale code
      *
      * @return string
@@ -256,10 +275,7 @@ class I18n
         }
 
         if ($formatNumber) {
-            $formatter = static::decimalNumberFormatter($locale);
-            if ($formatter !== null) {
-                $count = $formatter->format($count);
-            }
+            $count = static::formatNumber($count, $locale);
         }
 
         return str_replace('{{ count }}', $count, $message);
--- a/src/Toolkit/I18n.php
+++ b/src/Toolkit/I18n.php
@@ -233,6 +233,8 @@ class I18n
      */
     public static function translateCount(string $key, int $count, string $locale = null)
     {
+        $locale = $locale ?? static::locale();
+
         $translation = static::translate($key, null, $locale);
 
         if ($translation === null) {
@@ -251,10 +253,6 @@ class I18n
             } else {
                 $message = end($translation);
             }
-        }
-
-        if (!$locale) {
-            $locale = static::locale();
         }
 
         $formatter = static::decimalNumberFormatter($locale);
--- a/src/Toolkit/I18n.php
+++ b/src/Toolkit/I18n.php
@@ -45,7 +45,7 @@ class I18n
     public static $fallback = 'en';
 
     /**
-     * Decimal number formatters by locale
+     * Cache of `NumberFormatter` objects by locale
      *
      * @var array
      */
--- a/src/Toolkit/I18n.php
+++ b/src/Toolkit/I18n.php
@@ -3,6 +3,7 @@
 namespace Kirby\Toolkit;
 
 use Closure;
+use NumberFormatter;
 
 /**
  * Localization class, roughly inspired by VueI18n
@@ -42,6 +43,13 @@ class I18n
      * @var string
      */
     public static $fallback = 'en';
+
+    /**
+     * Decimal number formatters by locale
+     *
+     * @var array
+     */
+    protected static $decimalNumberFormatters = [];
 
     /**
      * Returns the fallback code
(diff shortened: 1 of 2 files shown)


afbora:
Note that this is breaking change due to `intl` extension required.

neildaniels:
Good point. Thanks for pointing that out.

Also, now I’m wondering if this should use `$kirby->language()->locale(LC_NUMERIC)` to determine the locale for the formatter.

nilshoerrmann:
While I like the idea of better typographic formatting, I'm not sure this should be merged with the core as is: with this change, a function that is supposed to find the correct declination based on the amount is also applying formatting I haven't asked for.

Thinking of German for example, yes, we can separate thousands with a dot but it's not required and often no separator is added for the lower thousands, e. g. it's more common to write `1234` than `1.234`. There might even be styleguides that advise not to add dots to numbers.

If this is added to the core, the function should have a flag `$format` which defaults to `false` to keep this feature optional.

neildaniels:
Yes, this is _a change_. But I think the benefits outweigh cases where you disagree with locale-aware number formatting.

In particular, this now correctly formats numbers that aren't even in the Latin script. For example, if you're using an Arabic string, the digits would also be in Arabic `٠‎ ١‎ ٢‎ ٣‎ ٤‎ ٥‎ ٦‎ ٧‎ ٨‎ ٩`. That being said, if you did want to mix and match styles, I think this should be modified to use the `LC_NUMERIC` locale property of a Kirby language; in most cases.

I genuinely have a hard time disregarding what a locale-aware number formatter and a wealth of localization documentation suggests:
- https://docs.microsoft.com/en-us/globalization/locale/number-formatting
  > In the United States, this character is a comma (,). In Germany, it is a period (.). Thus one thousand and twenty-five is displayed as 1,025 in the United States and 1.025 in Germany. In Sweden, the thousands separator is a space.

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat
  > ```
  > // German uses comma as decimal separator and period for thousands
  > console.log(new Intl.NumberFormat('de-DE').format(number));
  > // → 123.456,789
  > ```

- https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/InternationalizingLocaleData/InternationalizingLocaleData.html

nilshoerrmann:
I‘m not against the enhanced formatting, I just think it should not be forced. So there needs to be a switch.

afbora:
Agree with @nilshoerrmann I wouldn't want it to format all numbers.

This is not the task of the method. I think the `translateCount()` method should get a formatted number.

neildaniels:
I very much disagree. `translateCount()` should not accept pre-formatted numbers—trying to select the correct translation string to use will be a disaster (have fun parsing a formatted number like `۱٬۲۳۴٬۵۶۷٫۸۹۱`). I also don't think it would be acceptable to have the function accept both an integer and a formatted string—that places too much of a burden on the caller of the function.

My primary motivation is that numbers get run through a formatter so they make the most sense in the languages being displayed. For example, for a Persian site, the numbers displayed would almost certainly be using the digits `۰ ۱ ۲ ۳ ۴ ۵ ۶ ۷ ۸ ۹`, not `0 1 2 3 4 5 6 7 8 9`. (As I previously said though, this should probably be determined by `$kirby->language()->locale(LC_NUMERIC)`.

Is the only objection to the use of grouping separators? Is that something you want want disabled on a per-language basis, or on specific string translations?

neildaniels:
If https://github.com/getkirby/kirby/pull/2432/files gets merged in, then decimals will now be accepted as part of the `translateCount()`. That makes the need for a number formatter even more relevant.

src/Toolkit/I18n.php
```diff
@@ -190,6 +198,20 @@ public static function translations(): array
         return static::$translations;
     }
 
+    /**
+     * Returns (and creates) a decimal number formatter for a given locale
+     *
+     * @return array
```
lukasbestle:
```suggestion
     * @return \NumberFormatter|null
```

src/Toolkit/I18n.php
```diff
@@ -44,6 +45,13 @@ class I18n
      */
     public static $fallback = 'en';
 
+    /**
+     * Decimal number formatters by locale
```
lukasbestle:
```suggestion
     * Cache of `NumberFormatter` objects by locale
```

src/Toolkit/I18n.php
```diff
@@ -190,6 +198,20 @@ public static function translations(): array
         return static::$translations;
     }
 
+    /**
+     * Returns (and creates) a decimal number formatter for a given locale
+     *
+     * @return array
+     */
+    protected static function decimalNumberFormatter(string $locale): NumberFormatter
```
lukasbestle:
```suggestion
    protected static function decimalNumberFormatter(string $locale): ?NumberFormatter
```

src/Toolkit/I18n.php
```diff
@@ -190,6 +198,20 @@ public static function translations(): array
...
+    {
+        if (isset(static::$decimalNumberFormatters[$locale])) {
+            return static::$decimalNumberFormatters[$locale];
+        }
+
+        return static::$decimalNumberFormatters[$locale] = new NumberFormatter($locale, NumberFormatter::DECIMAL);
```
lukasbestle, line 234: There should be a check for the existence of the `NumberFormatter` class here. If the class is not defined (= if the `Intl` extension is not loaded), `null` should be returned.

src/Toolkit/I18n.php
```diff
@@ -225,6 +247,13 @@ public static function translateCount(string $key, int $count, string $locale =
                 $message = $translation[2];
         }
 
+        if (!$locale) {
+            $locale = static::locale();
+        }
```
lukasbestle: This should be moved to the beginning of the method for consistency and written like this:

```php
$locale = $locale ?? static::locale();
```

src/Toolkit/I18n.php
```diff
@@ -225,6 +247,13 @@ public static function translateCount(string $key, int $count, string $locale =
...
+
+        $formatter = static::decimalNumberFormatter($locale);
+        $count = $formatter->format($count);
```
lukasbestle: Because `$formatter` might be `null` according to my suggestion above, there should be a check for `null` here. The number should only be formatted if the `$formatter !== null`.

lukasbestle (changes requested):
I agree with @nilshoerrmann and @afbora: It should be possible to disable formatting, even if that means that there is a fourth argument to the `translateCount` method (`bool $formatNumber = true`).

Also there should be an additional `formatNumber` method that can be used if just a plain number needs to be formatted. This method can then also be used by `translateCount`.

src/Toolkit/I18n.php
distantnative: Should we not remove the type hint in general as we didn't want to hint for objects/classes?

lukasbestle: In this case it's fine as it's a protected static method inside a class that doesn't get extended. Object type hints are only a problem in inheritance.

I thought about it as well, but I'd leave the hinting in as we'll be able to use object type hints everywhere in PHP 8. So starting with it in places like this where it already works well will reduce the migration effort later.

lukasbestle:
@neildaniels Thanks a lot for the changes, the implementation is very clean now! And sorry that it took so long. :)

Because I don't have push access to the PR branch in your org repo, I merged the PR manually.

PR #2453, opened 2020-02-15
Switch to Symfony Yaml #2452

### LINKED ISSUE #1308: [Panel] Textarea in structure field: Spaces are deleted on save.

### LINKED ISSUE #2452: Switch to Symfony Yaml
Because of issues like https://github.com/getkirby/kirby/issues/1308
(see also there for the discussion).

- [x] Add additional unit tests for YAML
- [x] Switch out libraries
- [x] Test Symfony Yaml against all unit tests
- [x] Make sure Symfony Yaml resolves #1308

distantnative:
## Related issues

- Fixes #2452 
- Fixes #1308

## Ready?

- [x] Passing all unit tests


afbora:
I just think if this PR would be more stable in the 3.4.0 milestone because of a key component in the content management system is changing 🤔

afbora (commented):
Me again 😅 

The first error I saw; If there is an `@` sign in the blueprint file, it gives an error.

````yaml
twitter:
  label: Twitter
  type: text
  icon: twitter
  placeholder: @username
````

> The reserved indicator "@" cannot start a plain scalar; you need to quote the scalar at line 51 (near "placeholder: @username").

**Comment by distantnative** [comment]
@afbora Regarding 3.4 I first thought so as well.. then again I would hope that this is a non-breaking change, so no one would notice. For that of course we need to do a lot of testing, as you just found one case where it breaks at the moment.

File: src/Data/Yaml.php [hunk 1]
```diff
@@ -30,8 +30,8 @@ public static function encode($data): string
         // change to english numerics to avoid issues with floats
         setlocale(LC_NUMERIC, 'C');
```
lukasbestle, line 31: We can probably remove this locale stuff around the parse command. It was only added because Spyc is so buggy.

src/Data/Yaml.php
```diff
@@ -57,14 +57,11 @@ public static function decode($yaml): array
...
 
-        if (is_array($result)) {
-            return $result;
-        } else {
-            // apparently Spyc always returns an array, even for invalid YAML syntax
-            // so this Exception should currently never be thrown
-            throw new Exception('YAML string is invalid'); // @codeCoverageIgnore
-        }
+        // ensure that single string is wrapped in array
+        $result = A::wrap($result);
```
lukasbestle, line 63: I'm not sure if we should wrap the result if it's not already an array. While this can be convenient, it could also cause issues when the wrong kind of data is parsed by accident.

The Exception we had in the old implementation was there on purpose, it was just apparently never triggered because Spyc didn't allow it.

lukasbestle (changes requested):
I'm also for moving this to 3.4.0. The migration could cause issues we are not yet aware of and it's important that patch releases can be upgraded without thinking about it.

src/Data/Yaml.php
distantnative: @lukasbestle I put it in cause everything crashes when a single string is returned instead of an array with a single string inside. Spyc always returned arrays.

src/Data/Yaml.php
distantnative: Are you sure? Wasn’t that added over the whole comma vs. dot separator issues?

lukasbestle: Yes, but that should ideally be handled directly in the library. It was a bug in Spyc that Spyc didn't care about the locale. We have unit tests for this case and they should still pass with Symfony Yaml without the locale hack, otherwise Symfony Yaml is also buggy.

src/Data/Yaml.php
lukasbestle: Yes, but as I said: That's on purpose. The `Data` classes are always expected to deal with arrays, not scalar values. So if you pass a YAML string that only contains a string or number or whatever, that's not a valid input for the `Data` classes. Instead an Exception should be thrown like in the `JSON` handler.

Or is there code already that relies on this behavior that Spyc always returned an array?

distantnative: Yes, our code relies on that. If I put it in without the array wrapper, tests start failing left and right.

lukasbestle: Oh noes. That‘s bad. 😕

lukasbestle: I just tested it. Only three tests fail for the three field methods `toFile`, `toPage` and `toUser`. They fail because the current implementations of the respective plural methods support fields with a single string instead of an array of strings. We could fix those methods by checking if the field value begins with `-`. If it doesn't, there's only one name in the field and we can use it without trying to parse it as YAML.

distantnative:
@lukasbestle Do you have an idea for the problem that @afbora found? The Symfony library doesn’t accept values that aren’t wrapped as strings but start with an @ or so. This really seem to me like a big breaking change that I’m not sure how to get around... Spyc seemed to be much looser about these things.

lukasbestle:
Looks like we need to bite that bullet. Symfony Yaml seems to follow the spec much more closely. Fortunately it produces usable error messages, so it's at least pretty easy to fix such bugs in the YAML files.

src/Data/Yaml.php
distantnative: I'm not sure if we need to make it that complicated. The library is much more rigid with wrong data and actually throws exceptions itself - not like Spyc. And I feel these are much more helpful to use directly than covering up with any generic exception from us.

So it's really just about the case where a single string, int or so is returned instead of wrapped in an array.

distantnative:
@lukasbestle Could there be an update script we offer that parses all blueprints with Spyc but rewrites them with Symphony Yaml?

lukasbestle:
The update script is a nice idea!

src/Data/Yaml.php
lukasbestle: > So it's really just about the case where a single string, int or so is returned instead of wrapped in an array.

Yes, it is! That's what we are discussing about, aren't we?

The previous implementation has such a check that validates whether we got an array. If yes, the array is returned, otherwise an Exception is thrown. If there's a parsing error in the YAML input, the code won't even be run as the Exception that is thrown by Symfony YAML will break out of the `Yaml::decode()` method.

So we actually only need to revert these lines to the previous/current implementation and only need to remove the code coverage ignore marker and the comment.

distantnative: I read our conversation again but I still don't quite get your point.

SPYC would always return an array for valid YAML. 
So checking if the result is an array, if so returning it, if not throwing an Exception - that made sense.

The new library will return either an array or a single scalar for valid YAML.
If we keep the old code, we would throw an Exception for the latter cases, which would be a false negative.
The new library also throws reliably exceptions for invalid YAML (which SPYC didn't). So we can rely on this, I think.

distantnative:
I tried to come up with an update script:

```php
<?php

use Kirby\Toolkit\F;
use Spyc;
use Symfony\Component\Yaml\Yaml as Parser;

require 'vendor/autoload.php';

$kirby = kirby();

$blueprints = glob($kirby->root('blueprints') . '/*.yml') +
              glob($kirby->root('blueprints') . '/*/*.yml');

foreach ($blueprints as $blueprint) {
    $yaml = F::read($blueprint);
    $data = Spyc::YAMLLoadString($yaml);
    $yaml = Parser::dump($data, 99999, 2);
    F::write($blueprint, $yaml);
}
```

**Issues:**
- It will wrap most strings in single quotes. But doesn't seem to like if they already had single quotes inside them (see the two single quotes instead of a double quote):
```diff
- parent: kirby.page('photography').children.first
+ parent: 'kirby.page(''photography'').children.first'
```
- All comments get erased (of course, but still)

src/Data/Yaml.php
lukasbestle: The point is: With SPYC, the exception was never thrown, not even for invalid YAML or any kind of input – SPYC *always* returns an array. However, the exception *should* have been thrown for non-array values and that's the reason why there currently is the comment above that tells the reader that this is a known issue.

The reason why non-arrays should trigger an Exception is the following: The `Handler::decode()` method, which `Yaml::decode()` inherits from, only allows array return values, never scalar values (because the `Data` classes are designed to always deal with array data).

So if a non-array value is parsed by the decoder, that should trigger a clear error as the user tried to parse data that the `Data` class is not designed to work with. If we don't throw an exception, the method will try to return a non-array value from a method that must only return arrays. This will trigger a PHP Fatal Error, which should be avoided.

I suggest the following implementation:

```php
public static function decode($yaml): array
{
    if ($yaml === null) {
        return [];
    }

    if (is_array($yaml) === true) {
        return $yaml;
    }

    $result = Parser::parse($yaml);

    if (is_array($result)) {
        return $result;
    } else {
        throw new Exception('YAML string is invalid');
    }
}
```

Again: It's *not* about YAML that is syntactically invalid. Those cases will throw an Exception inside Symfony Yaml, so the `if (is_array())` code won't even be run. This is just about the case of non-array YAML data. Also see the `Data\Json` class with the same implementation.

distantnative: But that’s why I put the A::wrap in there: we know that the YAML is valid because the new parser didn’t throw an exception. To comply with our return types, we wrap it in an array.

afbora:
@distantnative What if we got caught with Regex?

````php
$yaml = F::read($blueprint);

// catches starts with @
// twitter: @username > twitter: '@username'
$yaml = preg_replace("/[a-z0-9_]+(?:\s+)?(@.*)/", "'\\1'", $yaml);

// catches and replaces that contains single quote
// parent: kirby.page('photography').children.first > parent: 'kirby.page("photography").children.first'
$yaml = preg_replace_callback('/[a-z0-9_]+(?:\s+)?(.*)/', function ($match) {
    return str_replace("'", '"', $match[0]);
}, $yaml);

$data = Spyc::YAMLLoadString($yaml);
````

src/Data/Yaml.php
lukasbestle: Looks like we can discuss forever. We have exchanged those same arguments several times now. :)

I still don't think `A::wrap()` is a good idea because lax parser handling can cause unexpected issues down the road, but if you want to do it that way, sure.

distantnative: Yea I guess we have opposing opinions here.
I think it’s bad to throw an exception ourselves mentioning invalid YAML when the YAML is actually valid.

bastianallgeier:
Looking at the survey results and the number of people who already hate blueprints, I think we are going to piss a lot of people off if we introduce such a massive change now. Just imagine you have to keep 10 client projects updated and for 10 client you need to run the script to fix yaml bugs just to upgrade from 3.3.x to 3.4. Afterwards you have to relearn that all you've learned about blueprints so far has changed and there are now new rules about how strings have to be written if they contain special characters. This sucks big time if you ask me.

bastianallgeier:
Just a terrible idea out of the blue. What about moving to PHP for blueprints? With the old buggy spyc yaml as fallback? With PHP we could maybe introduce entire new ways to tackle the complexity issues.

bastianallgeier:
It would probably also solve a lot of performance problems.

afbora:
It doesn't matter if it is php or yaml for me but it would be great if it was a builder that create blueprints from panel easily.

distantnative:
I actually haven't gone ahead with this cause whatever I tried I think there is no good transition, not even an easy update script (and already those aren't ideal). It would be a lot of manual fixing for people... which is a shame cause I actually would like this new library because it works so much more reliable.

Supporting PHP blueprints would be great - in fact we already do if you register a blueprint via a plugin. On the other hand, I think those people who complained about the YAML won't be much happier with PHP syntax for blueprints. Cause let's be honest all those arrows, brackets and quotation marks make it less pleasant to read and much more tiresome to write.

I would believe that the majority sticks around with YAML, PHP would rather be like the expert way instead of the new default.

@afbora The builder side is a whole other topic

lukasbestle:
I agree with @distantnative. PHP blueprints are a nice feature for advanced users (and we should probably add that feature anyway, because it's quite simple to do). But making everyone switch to it, including changing the docs all over, would also be unpopular. Of course we could make the switch optional by keeping support for YAML blueprints forever, but it would always feel like YAML blueprints are the "old, bad way".

Also it's not only a blueprint issue. For example https://github.com/getkirby/kirby/issues/1308 was about structure field data in content files. We can't solve this in another way than to switch the YAML library.

So I see basically three solutions:

1. Ship Kirby with both libraries and make the library to use configurable. Maybe we could somehow detect if a site was newly created after the switch and then enable the new library, otherwise keep the old one. Disadvantage: Complicates debugging as it increases the number of variables in a Kirby installation. Also Symfony Yaml is quite strict, which could annoy users in general (and unfortunately not modular, so it's not even possible to disable features we don't need).
2. Fork Spyc and try to fix the actual issues we are having ourselves while keeping the quirks and lax parsing in general. Problem: Potentially hard to do and a lot of effort.
3. Make the switch anyway, either now or at a later point with as much support from us as possible (detailed docs about the switch, update script etc.).

We have the same issue with the Markdown library right now, which also bites us in exactly the same way. :(

bastianallgeier:
After letting this lay around some more I think we should close it. We can't solve it right now without massive pain for all users. We plan to have a blueprint builder at some point anyway and then we can solve this in a completely new way. Until then I'm afraid we are stuck with the old lib.

lukasbestle:
See https://github.com/getkirby/kirby/pull/2453#issuecomment-648087004.

PR #2454, opened 2020-02-15
Fix dialog width on mobile #2415

### LINKED ISSUE #2415: Pages dialog/modal breaking layout
**Describe the bug**  
The pages dialog/modal layout breaks when you have longer page titles or use some kind of `info` text. Especially on smaller screens this issue is easy to encounter.

**To Reproduce**  
1. Add a `pages` field to your blueprint
2. Either use long page titles or for example display the page template in the `info`
3. Open the modal by clicking on the "Select" button
4. Resize your window if necessary

**Expected behavior**  
The modal should be max 100% in width so it fits into the viewport.

**Screenshots**  
modal

**Kirby Version**  
3.3.3

 - OS: macOS
 - Browser: Chrome, Mobile Safari (probably every browser)

**Smartphone (please complete the following information):**  
 - Device: iPhone 7
 - OS: iOS13.3
 - Browser: Mobile Safari
 - Version: 

**Additional context**  
I don't understand why the width of `k-dialog-box` is set to a fixed width. Just use `width: 100%` and a `max-width` and it's fixed. Ideally something wider than 30rem.

distantnative:


panel/src/components/Dialogs/Dialog.vue
```diff
@@ -186,16 +189,28 @@
...
 
-.k-dialog-box[data-size="large"] {
-  width: 40rem;
+@media screen and (min-width: 30rem) {
+  .k-dialog-box[data-size="medium"] {
+    width: 30rem;
+  }
+}
+
+@media screen and (min-width: 30rem) {
```
medienbaecker: I think you meant `min-width: 40rem`. Otherwise the 10rem are cut off.

distantnative: Yup! 😅

afbora:
Just an idea came to my mind. This is not the subject of PR, but what would it be if we showed the entire title with the `title` tag when it hovered over in long sentences? (Especially desktop)

afbora (approved):
Great works 👍

bastianallgeier (approved):

PR #2457, opened 2020-02-17
Validate empty locale in language editor

### LINKED ISSUE #2455: Language editor in the Panel: Empty locale not validated
Originally reported by @afbora in #2435:

---

Saving empty locale issue as a second problem steps to reproduce:

1. Add a language
2. Edit language
3. Leave blank locale input
4. Save language
5. See issue as language config like:

````php
'locale' => [
    'LC_ALL' => ''
]
````

You can also enter a space from panel and config saving like that:

```php
'locale' => [
    'LC_ALL' => '    '
]
```

---

I think we should fix this by:

- removing surrounding whitespace from the input by trimming
- setting the locale to the language code if empty (like the Panel already does it when creating languages)

afbora:
## Describe the PR

If the `locale` option is string, spaces are deleted with `trim()`, and if empty, the code value is assigned to.

@lukasbestle I fixed it in the language editor for panel, but do we have to do this in the language class? I mean that:

````php
} elseif (is_string($locale)) {
    $locale = trim($locale);
    $locale = empty($locale) === false ? $locale : $this->code;
    $this->locale = [LC_ALL => $locale];
}
````

https://github.com/getkirby/kirby/blob/3.3.3/src/Cms/Language.php#L504-L505

## Related issues

- Fixes #2455

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/panel/src/components/Dialogs/LanguageCreateDialog.vue
+++ b/panel/src/components/Dialogs/LanguageCreateDialog.vue
@@ -95,7 +95,7 @@ export default {
     },
     submit() {
       if (this.language.locale) {
-        this.language.locale = this.language.locale.trim();
+        this.language.locale = this.language.locale.trim() || null;
       }
 
       this.$api
@@ -103,7 +103,7 @@ export default {
           name: this.language.name,
           code: this.language.code,
           direction: this.language.direction,
-          locale: this.language.locale || this.language.code
+          locale: this.language.locale
         })
         .then(() => {
           this.$store.dispatch("languages/load");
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
Now that I think about it: We already have [code in the Language class](https://github.com/getkirby/kirby/blob/3.3.4/src/Cms/Language.php#L506-L507) that automatically does the conversion in the backend if the passed locale is `null`.

So maybe we can get away with returning `null` from the dialogs when the trimmed value is an empty string. That should trigger the conversion in the backend.

afbora:
@lukasbestle Now locale is return always `null`, if null or trimmed locale is empty.

lukasbestle:
I can unfortunately not test this at the moment as my local `node-sass` is broken and I can't build the dist files. 😡
But if this works, then 👍 from me.

bastianallgeier (approved):

PR #2459, opened 2020-02-17
Add support for late binding in FileActions to enable subclasses of File

weese:
## Describe the PR

I have a small fix for late binding in FileActions.php. FileActions is a trait and should not make assumptions on the class it is used in, so that it can be used in other File classes too. However, FileTraits explicitly instantiates a File object in `factory()`. I changed it to `static` to support late bind, so it simply instantiates the class it is used in
For Kirby this PR won't affect anything but for others (like me 😉) who develop their own File class. This change is required to not simply copy'n'paste the nice `FileActions` trait.

## Ready?

- [ ] Passing all unit tests


lukasbestle (approved):
I have not tested this locally, but I don't see why this could be harmful. So 👍.

PR #2465, opened 2020-02-18
Support numeric parts in query

### LINKED ISSUE #2462: Query syntax in blueprints ignores nested config variable when key contains a number
**Describe the bug**  
From within my blueprint I would like to access a variable from a config array for a label. This does not work, if the array key contains a number. It then simply shows nothing as a label.

**To Reproduce**  
Given the following config.php:
```
return [
    'places' => [
        "place1" => [
            'name' => 'Place 1'
        ],
        "placeTwo" => [
            'name' => 'Place 2'
        ]
    ]
]
```
the blueprint
```
title: My Places
fields:
  text:
    type: text
    label: "{{kirby.option("places").place1.name}}"
```
does not show anything as a label.

`label: "{{kirby.option("places").placeTwo.name}}"`, however, would work.

**Kirby Version**  
3.3.4

afbora:
## Describe the PR

Now query parts support `numeric` values with added `0-9` to regex.

## Related issues

- Fixes #2462 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Toolkit/Query.php
+++ b/src/Toolkit/Query.php
@@ -15,7 +15,7 @@ namespace Kirby\Toolkit;
  */
 class Query
 {
-    const PARTS      = '!([a-zA-Z_]*(\(.*?\))?)\.|' . self::SKIP . '!';
+    const PARTS      = '!([a-zA-Z0-9_]*(\(.*?\))?)\.|' . self::SKIP . '!';
     const METHOD     = '!\((.*)\)!';
     const PARAMETERS = '!,|' . self::SKIP . '!';
 
--- a/src/Toolkit/Query.php
+++ b/src/Toolkit/Query.php
@@ -15,7 +15,7 @@ namespace Kirby\Toolkit;
  */
 class Query
 {
-    const PARTS      = '!([a-zA-Z0-9_]*(\(.*?\))?)\.|' . self::SKIP . '!';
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
Maybe we should restrict it so that numbers may not be at the beginning of each part component (= the first character of each part must be a letter or underscore, the second and all following chars may also be a number). Numbers at the beginning of a string sometimes cause weird issues when code tries to check if the string is numeric etc.

Unfortunately it's going to be a bit difficult to express that as regex as a part can currently be empty (without any character of the first character list). Why is that BTW? @bastianallgeier

distantnative:
It’s because the refer became a monster and very hard to overlook :D 
Feel free to add changes, we just need to have sufficient tests to make sure there are no other problems introduced by the fix.

afbora:
> Unfortunately it's going to be a bit difficult to express that as regex as a part can currently be empty (without any character of the first character list).

@lukasbestle Since I could not fully understand the problem here, I could not apply it 😇, but I restricted the parts that started with the number.

lukasbestle (approved):
Looks good!

@afbora What I meant is that the RegEx now requires at least one letter or underscore in each component to match. Before it just used the `*` quantifier, which would have also permitted zero characters. But your solution looks good as I don't think there is any use-case where a valid part could be empty.

bastianallgeier (approved):

PR #2472, opened 2020-02-25
Fix loading template controllers on representations

### LINKED ISSUE #2471: Controller not passing data to content representation if php template is not present
**Describe the bug**  
I use Kirby as a headless CMS and am not interested in using the normal templates, only their content representation version, e.g. json. But the controller doesnt pass data to the content representation version of a template if its normal template is not present as well.

**To Reproduce**  
1. Go to `site/templates` and remove all templates, leave only `site/templates/default.php`
2. Add a json content representation template for some page that exists and uses this template, e.g. `site/templates/album.json.php` (in starterkit)
3. Add a controller for this representation, i.e. `site/controllers/album.json.php` and try to pass some variable to the json template.
4. Doesnt work. But if you create a file for the normal template, i.e. `site/templates/album.php`and even leave it empty, suddenly the variable from the controller gets passed to the json template.

**Expected behavior**  
The controller for the json template passes data to the json template even without having to have a normal (non-json) template present.

**Kirby Version**  
3.3.4

afbora:
## Describe the PR

The issue had two sources:

1. Since the `representation()` method doesn't update the `template` variable of class, if there is no template file, it always returns the default template.
2. Since the controller call process is related to the template, the order of setting the template and then calling the controller has been adjusted.

Maybe a better development can be done or you may want to refactor. So I open it as a draft.

## Related issues

- Fixes #2471 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -1134,13 +1134,13 @@ class Page extends ModelWithContent
                 $template = $this->representation($contentType);
             }
 
-            $kirby->data = $this->controller($data, $contentType);
-
             if ($template->exists() === false) {
                 throw new NotFoundException([
                     'key' => 'template.default.notFound'
                 ]);
             }
+
+            $kirby->data = $this->controller($data, $contentType);
 
             // render the page
             $html = $template->render($kirby->data);
@@ -1172,7 +1172,7 @@ class Page extends ModelWithContent
         $representation = $kirby->template($template->name(), $type);
 
         if ($representation->exists() === true) {
-            return $this->template = $representation;
+            return $representation;
         }
 
         throw new NotFoundException('The content representation cannot be found');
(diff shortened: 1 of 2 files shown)


src/Cms/Page.php
```diff
@@ -1172,7 +1172,7 @@ public function representation($type)
         $representation = $kirby->template($template->name(), $type);
 
         if ($representation->exists() === true) {
-            return $representation;
+            return $this->template = $representation;
```
afbora: @lukasbestle I am not sure if there is a need to undo this change?

afbora:
@lukasbestle Can you make the last touch please? 😇

lukasbestle (approved):

bastianallgeier (approved):

PR #2478, opened 2020-03-02
Fix dialog with sticky columns

### LINKED ISSUE #2477: Modal page creation appear under a page section with a sticky column
When I click on a page section which is inside a column with a sticky position to add a page, the modal for create the page appear under another column's page section.

Capture d’écran 2020-03-01 à 22 44 53

The column that is sticky is the left one.

Capture d’écran 2020-03-01 à 22 45 20

There is no problem without the sticky options.

afbora:
## Describe the PR

Fixed the issue with added `z-index` property as `2` value. The value of `z-index: 1` doesn't fully fix it. Should be min value as `z-index: 2`

## Related issues

- Fixes #2477 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/panel/src/components/Layout/Column.vue
+++ b/panel/src/components/Layout/Column.vue
@@ -22,6 +22,7 @@ export default {
 .k-column[data-sticky] > div {
   position: sticky;
   top: 4vh;
+  z-index: 2;
 }
 
 @media screen and (min-width: $breakpoint-medium) {


distantnative:
Does it really fix it? Now it seems that we raise all div containers in a column to Z index of 2...
My initial thought was rather that my CSS definition for the sticky columns was too broad and needs to exclude dialogs rather.

afbora:
@distantnative Yes fixed, but you're right, this was a simple solution, but can it lead to a different issue because we've increased the value to 2? 🤔 

Yes, we are having this issue because the dialog codes is inside the sticky container.  As you said, if the dialog was independent, this issue would not be experienced.

**Before PR**

FireShot Capture 043 - Notes - Mægazine - localhost

**After PR**

FireShot Capture 044 - Notes - Mægazine - localhost

afbora:
@distantnative Would you like me to close the PR?

distantnative:
Nono, let's explore this

bastianallgeier:
Oh no, z-index issues are the worst!

distantnative:
Had a closer look at this. I think the solution is the best/only viable. 
z-index within a sticky positioned element is ignored (the dialog). Either we would hav etc change the HTML structure or make sure that the sticky element is layered above others. I think the latter (as in this PR used) works here.

distantnative (approved):

PR #2479, opened 2020-03-02
Fix kirbytag helper passing data

### LINKED ISSUE #2474: Kirbytag helper function not working as expected
**Describe the bug**  
Right now it is not possible to pass a page (`parent`) reference to the helper function.

```php
// does not work
echo kirbytag('image', 'some-image.jpg', [
  'parent' => $page,
]);

// using the kirbytags helper works though
echo kirbytags('(image: some-image.jpg)', [
  'parent' => $page,
]);
```

After some digging I probably found the issue. I can't tell if this is by design or just an oversight but the `Kirby\Cms\App` uses the `Kirby\Text\KirbyTag` class instead of the extended `Kirby\Cms\KirbyTag` one. I think that this is indeed a bug and not a feature 🙂  
So deleting this line fixes the issue:

https://github.com/getkirby/kirby/blob/3.3.4/src/Cms/App.php#L15

**Kirby Version**  
3.3.3

afbora:
## Describe the PR

Applied changes found by @lukaskleinschmidt with added missing `data` attribute.

## Related issues

- Fixes #2474 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -412,10 +412,10 @@ function kirbytag($type, string $value = null, array $attr = [], array $data = [
         $value    = current($kirbytag);
         $attr     = $kirbytag;
 
-        // check data attribute and remove from array if exists
-        if (isset($kirbytag['data']) === true) {
-            $data = $kirbytag['data'];
-            unset($kirbytag['data']);
+        // check data attribute and separate from attr data if exists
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):

bastianallgeier:
Really helpful! I ran into this just recently.

PR #2485, opened 2020-03-04
Supports brackets in kirbytags

afbora:
## Describe the PR

I got professional support on Regex 😇 

Use a positive lookahead to match sure the match starts with that pattern and used a `(?1)` subroutine instead of `(?R)` to only recurse that pattern, not the whole regex.

See demo: https://regex101.com/r/1Tti14/2

## Related issues

- Closes https://github.com/getkirby/ideas/issues/10

## Ready?

- [x] Passing all unit tests


tests/Text/KirbyTagsTest.php
```diff
@@ -66,4 +66,19 @@ public function testParseWithException()
...
+            'test' => [
+                'html' => function ($tag) {
+                    return $tag->value;
+                }
+            ]
+        ];
+
+        $this->assertSame('foo(bar)', KirbyTags::parse('(test: foo(bar))'));
+        $this->assertSame('foo(bar)', KirbyTags::parse('(Test: foo(bar))'));
+        $this->assertSame('foo(bar)', KirbyTags::parse('(TEST: foo(bar))'));
```
lukasbestle: More tests would be great (e.g. with another argument following the main tag value, with text following the bracketed part inside the same argument, with multiple nested brackets inside the same value, with a value where opening and closing brackets don't match).

distantnative: I'm really hoping it will pass those tests - would be so cool! 😁

afbora: Now all good 🚀

afbora:
I have added more unit tests 👍 

Actually, there's something I'm curious about. 
How about we get inside the parentheses on regex? 
Thus, we do not deal with deleting the parentheses at the beginning and at the end.

Currently we match with `(test: foo)` and try to delete start and end brackets.
What about match with `test: foo` directly without start and end brackets? 🤔

distantnative:
I tried the regex on regex101 with some examples. Where it broke for me:

```
This in an example (link: http:///www.google.com text: Soo evil :( I don't like.) tag
```

In general, a single bracket seems to be an issue.

The PR would still be better than the current implementation. But maybe we can make even that case work?

afbora:
> In general, a single bracket seems to be an issue.

https://github.com/getkirby/kirby/blob/feature/10-brackets-in-kirbytags/tests/Text/KirbyTagsTest.php#L93-L94

Yes, it will not match naturally when there are missing brackets. But there may be a problem with the use of emoji. I don't think it's very important, but we can still try to solve it.

distantnative:
I guess it might be too hard to find a regex that can handle single brackets.
As I said, this PR is already better than what we have had so far.

afbora:
@distantnative You are right. Yes, when I think again, it seems almost impossible for us to understand whether this is a missing brace or an use like an emoji.

distantnative (dismissed):

src/Text/KirbyTags.php
```diff
@@ -22,7 +22,7 @@ class KirbyTags
 
     public static function parse(string $text = null, array $data = [], array $options = []): string
     {
-        return preg_replace_callback('!(?=[^\]])\([a-z0-9_-]+:.*?\)!is', function ($match) use ($data, $options) {
+        return preg_replace_callback('!(?=[^\]])(?=\([a-z0-9_-]+:)(\((?:[^()]+|(?1))*+\))!is', function ($match) use ($data, $options) {
```
lukasbestle: It would be nice to move the RegEx into a separate string and document it with comments like this:

```php
$regex = '!
            (?=[^\]])          # What this part does
            (?=\([a-z0-9_-]+:) # And this one
            ...
          !isx';

**Review by lukasbestle — COMMENTED** [review]
The ideal solution for the single-bracket case would be to make brackets inside Kirbytags escapeable with a backslash. But that would probably make the RegEx even longer. So I agree with Nico that your implementation is already a lot better than the current one and therefore worth merging.

**Comment by distantnative** [comment]
@bastianallgeier I think we could put this in 3.3.5 – it's not a breaking change, right?

File: src/Text/KirbyTags.php [hunk 3]
```diff
@@ -22,7 +22,15 @@ class KirbyTags
 
     public static function parse(string $text = null, array $data = [], array $options = []): string
     {
-        return preg_replace_callback('!(?=[^\]])\([a-z0-9_-]+:.*?\)!is', function ($match) use ($data, $options) {
+        $regex = '!
+            (?=[^\]])               # positive lookahead that matches a group after the main expression without including ] in the result
```
afbora, line 26: I'm not sure if we really need this. What exactly is it here to prevent?

bastianallgeier: I'm really sorry, but I can't really help with this one. @lukasbestle @distantnative ?

bastianallgeier:
We've been waiting for this for soooooo long! It finally seems to be a good solution. The sad emoji :( is the only breaking change, but I think that's fine. Awesome work Ahmet!!

afbora:
You're wellcome ❤️ @bastianallgeier 

Where will the break change be? 
Brackets now work more stable. 
All legacy tags will continue to work 🤔

PR #2487, opened 2020-03-05
Update cached auth user data on update

### LINKED ISSUE #2486: User.txt session refresh issue after $user->update
Passing $user = $user->update… to a template is working as expected but
when a user edit his data (user.txt) in the frontend, change stored (in a session?) aren't updated (snippet for my case).

CASE SCENARIO:
LoggedIn user can subscribe (true) or unsubscribe (false) to a newsletter on a toggle basis:
```
$user = $kirby->user('user@email.com');
$currentStatus = $user->newsletter();
echo 'Old value: '.$currentStatus.'<br>';
$variable = ($currentStatus == 'true' ? 'false' : 'true');
$user->update(['subscribe' => $variable]);
echo 'New value: '.$user->newsletter();
```
RESULT:
Old value: false
New value: false 

EXPECTED:
Old value: false
New value: **true**

Did I miss Something to update the User globally or it is a bug?
Hope it helps making Kirby even better

afbora:
## Describe the PR

Update cached auth user data on update only if the current user is this user.

@lukasbestle Could you help me about unit test? 😇 I'm getting error like following:

> Kirby\Cms\UserActionsTest::testUpdateWithAuthUser
> Exception: The directory "/dev/null" cannot be created

## Related issues

- Fixes #2486 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/UserActions.php
+++ b/src/Cms/UserActions.php
@@ -278,6 +278,26 @@ trait UserActions
     }
 
     /**
+     * Updates the user data
+     *
+     * @param array $input
+     * @param string $language
+     * @param bool $validate
+     * @return self
+     */
+    public function update(array $input = null, string $language = null, bool $validate = false)
+    {
+        $user = parent::update($input, $language, $validate);
+
+        // set auth user data only if the current user is this user
+        if ($user->isLoggedIn() === true) {
+            $this->kirby()->auth()->setUser($user);
+        }
+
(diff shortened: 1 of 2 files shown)


afbora:
I have fixed the unit test different way.
If there is a deficiency or an error in the test, we can fix it.

bastianallgeier (approved):

PR #2490, opened 2020-03-05
`language` config option

distantnative:
## Describe the PR

This PR adds a new `language` config option which is used to define a language locale and/or slug rules for a single-language site.

Can be used either as 
```php
'language' => 'de'
```
or
```php
'language' => [
    'code' => 'de',
    'slugs' => [...]
]
```

The previous `slugs` option is still supported, but should be marked as deprecated and removed in an upcoming version.

@lukasbestle @texnixe @nilshoerrmann 
I am still working on this but since all three of you have been active in the idea issue, I would like to hear your thoughts on the general implementation idea.

## Related issues

https://github.com/getkirby/ideas/issues/459


nilshoerrmann:
I like it. Not sure though if `locale` should be `code` in the Kirby context – but you‘ll know better.

distantnative:
No, you are right, it should to be consistent 👍

bastianallgeier (changes requested):
I really think this one needs tests. But it looks great otherwise!

distantnative:
That's true!

distantnative:
I also think this has an issue still of not loading translations for a single language. Moving it to 3.5 to spend some more time on this.

distantnative:
Will create new PR once ready

PR #2492, opened 2020-03-06
Improve multiselect performance with new limits

### LINKED ISSUE #1671: Multiselect field: don't render all options at once for really long lists of options
**Describe the bug**
When using a `multiselect` with nearly ~2300 items, the performance of displaying/filtering results is nearly unusable.

**To Reproduce**
1. Have a multiselect field with like 2000 options in it (either manually or with a query like index)
2. Click on it to display

**Expected behavior**
I would have expected the dialog to display relatively quickly. It currently takes about 3 or 4 seconds.

**Kirby Version**
`3.1.2`

 - OS: macOS 10.14.4 (18E226)
 - Chrome 73.0.3683.103
 - Safari 12.1 (14607.1.40.1.4)

**Additional context**
The `tags` field is much more performant—it still fills _slightly_ laggy, but is under 1/2 second.

afbora:
## Describe the PR

As @bastianallgeier suggests, I can't do any major development on the API side, but I think I have made a simple and effective development.

Alternatively, we can use the current `search` option instead of `minSearch` like that `search: bool|int`

Since the proposal is a PR, I opened it as a draft.

**Sample usage of PR:**

````yaml
fields:
  categories:
    label: Categories
    type: multiselect
    minSearch: 2
    display: 5
    options:
      panel: Panel
      templating: Templating
      seo: SEO
      security: Security
      performance: Performance
      analytics: Analytics
      assets: Assets
      text: Text
      forms: Forms
      utilities: Utilities
      integrations: Integrations
      social: Social
````

**Screencast**

multiselect

## Related issues

- Fixes #1671 

## Ready?

- [x] Passing all unit tests

### CODE REVIEW

**Comment by distantnative** [comment]
Thinking out loud about a syntax:

```yaml
search: true

# or

search:
  min: 2
  display: 5
```

Since the `display` option doesn't relate to the whole multiselect field, but just the search. What do you think, @afbora?

And are you sure it fixes the issue referenced? I would assume that the HTML DOM nodes are still added (and thus with 2000+ options the browser gets really slow).

distantnative:
Edit: Just saw, that you filter the options in general if `min` is not `0` - so maybe it does work :D

afbora:
@distantnative I like the idea because of you have improved using an existing option 👍 

We cut the options with `slice()`. Doesn't this work?

afbora:
@distantnative Now I checked double. It is rendered only as much as the specified/limited result in the browser.

afbora:
@neildaniels Can we get your opinions too? Don't be too hard (ideas), please 😄

afbora:
You can test with updated dist files:  [dist.zip](https://github.com/getkirby/kirby/files/4298797/dist.zip)

neildaniels:
This seems like an improvement to the performance aspect, but I don't like that it makes `multiselect` fields feel more like a `tags` field.

Specifically: I _like_ that the `multiselect` currently shows options before you've typed anything—you can get a sense of what options are available without having to know what to type.

I think a simpler solution to #1671 would be to show a few options and then have a "Show All" button. My original concern was the delay it takes in actually showing the options (just by clicking on the field and while filtering). I suspect only showing a few options solves that issue.

**Mockup:**
Screen Shot 2020-03-06 at 9 57 18 AM

If a "Show All" button speeds up filtering, I don't think I would personally use the minimum character count feature.

distantnative:
@neildaniels 
You don't have to set the minimum character count (or rather you can set it to `0`). Setting a display count is sufficient. The "Show all" would definitely have a delay. We won't get around that if not asynchronously loading parts of the list - which probably won't happen soon.

neildaniels:
Understood, but there's currently no way to "show all" results if they exceed the display count, right? To me, that's the UI incorrectly suggesting there are no more results and there's no need to type more characters to find the item you want.

I'm okay with a delay after pressing "Show All"—I'm mentally prepared that there could be delay after clicking the equivalent of a "load more" button. I just think it's extremely annoying to have there be a multi-second lag when simply clicking into the field.

I definitely understand that there won't be an async loading component to this soon—and that's fine. I assume the delay is actually creating the DOM elements to show, and that's acceptable to me as long as there's some kind of secondary interaction to indicate that I want those options displayed. To me, a "Show All" button accomplishes that.

afbora:
Our goal here is to force the user to perform a more specific search, as there are already too many results. Maybe `show more` can be used instead of `show all`. If the display limit is `20`, it will bring `20` more results per click 🤔

neildaniels:
“Show More” seems totally reasonable.

afbora:
@distantnative What do you think about `show more` solution?

distantnative:
Could be good... on the other hand could be annoying if someone really wants to see everything as @neildaniels pointed out - and then he/she has to click a dozen times on "show more".

afbora:
Hmm, set the number of more shows manually?

````yaml
search:
  min: 3
  display: 50
  more: 150
````

distantnative:
No, to me that's an unnecessary option - too many options aren't good either.
We should just decide together whether to go the "Show more" or the "Show all" way.

afbora:
So i would have preferred the option to `Show More`.

The first sentence of the issue is:
> When using a multiselect with nearly ~2300 items, the performance of displaying/filtering results is nearly unusable.

Therefore, showing all of items will not be a solution for this PR. Even the `minimum search` option will only partially increase this performance. For example, if there are 2300 results and a minimum of 3 characters, maybe there will be less than 500-1000, maybe under 100 results instead of 2300.

neildaniels:
I'd be fine with the "Show More" count being the same number as the displayed number. Thus, if you have the initial cap set to 20 items, the more button would show an additional 20.

Being able to know/show _more_ items is more important to me than showing _all_ items.

(This is definitely overkill, but if I was making a native app, I'd make holding down a modifier key (e.g. Option/Shift) swap "Show More" with "Show All".)

afbora:
So let's look at what we can do with the `show more` option. If I can't, I can ask you for help @distantnative 😇

distantnative:
Let's go for "Show more" then

afbora:
Hi @distantnative 😇 

It originally required a simple vue for me, but now it exceeds me a bit. 
I'm stuck on two issues, it would be better for you to do it please 🙏 

1. Because we use `slice()`, when I try to get the total number of options, it counts all the results every time. When I use `splice()`, I cannot count the total result.

2. I tried load other options when clicked, but when I first arrived and when I clicked I was not able to get results based on the current search.

distantnative:
I'll have a look!

afbora:
@distantnative ready for now to test?

distantnative:
Yes!

afbora:
@distantnative You solved the reload issue while panel login on develop branch. Can't login right now. Can you add patch it to this PR please?

distantnative:
@afbora Just did

afbora:
@distantnative Thank you. There are a few missing, can I push it?

distantnative:
A few missing? I think this PR has all commits from `develop`. You might have to remove and re-add this branch locally. I remember you had problems once when a branch gets rebase (which I just did).

afbora:
@distantnative Missing about multiselect feature :))

distantnative (approved):

bastianallgeier:
When I keep on searching with different queries, the list is shown entirely. Should we rest this once the search changes?

bastianallgeier:
There are also some type check issues

afbora:
> When I keep on searching with different queries, the list is shown entirely. Should we rest this once the search changes?

@bastianallgeier Could you be using the wrong format? It is not a format like the first message. Final syntax:

````yaml
search:
    min: 3
    display: 10
````

afbora:
@bastianallgeier missing something?

PR #2493, opened 2020-03-06
Better img `sizes`: make cards self-aware #2488

### LINKED ISSUE #2488: Panel loads huge images in cards layout
**Describe the bug**  
Currently, when using a card layout in the panel, even though there's both an `srcset` and a `sizes` attribute set, the browser is still loading the biggest possible crop rather than one at a smaller size.

This was supposed to be fixed in #1807 but the bug is still there.

Unless my understanding of sizes and srcset is completely wrong, the problem with the current implementation is mostly a matter of css/html.

Right now, there's 3 crops sizes and 3 sizes set up

`srcset="img-352x.jpg 352w, img-864x.jpg 864w, img-1408x.jpg 1408w"`
`sizes="(min-width: 30em) and (max-width: 65em) 59em, (min-width: 65em) 88em, 27em"`

The problem with this is that the current sizes attribute is telling the browser that when the browser window is at least 65em wide (so 1040px) the space available for the image is 88em.

Again, unless I got this completely wrong, sizes defines how big the image in the browser actually is displayed and with that information the browser then selects the appropriate image from the ones available inside srcset.

**To Reproduce**  
1. Create a Pages section
2. Set layout to cards and size to small or tiny
3. Open dev tools and check which image has been loaded by the browser

**Screenshots**  
75891226-c16f6b80-5e2f-11ea-8e33-07fd72c0566d

distantnative:
## Describe the PR

Current bus had been that in a `1/1` column with several cards side by side, the largest image would always be loaded, even if the actual width of the card is only 200px.

Instead of guessing wild complicated `size` configurations for all columns, sizes, nestedness... how about just providing a single `sizes` value - the width of the card itself?

This PR does this by letting Vue measure the width of the card on the fly.

## Feedback wanted

@manuelmoreale 
@nilshoerrmann (since you were so much involved in the initial discussions around `secret` and `sizes` in the Panel

## Related issues

- Fixes #2488

### DIFF
--- a/panel/src/components/Layout/Card.vue
+++ b/panel/src/components/Layout/Card.vue
@@ -3,12 +3,12 @@
     <k-sort-handle v-if="sortable" />
 
     <component :is="wrapper" :to="link" :target="target">
-      <k-image
-        v-if="imageOptions"
-        v-bind="imageOptions"
-        class="k-card-image"
-      />
-      <span v-else :style="'padding-bottom:' + ratioPadding" class="k-card-icon">
+      <k-image v-if="imageOptions" v-bind="imageOptions" class="k-card-image" />
+      <span
+        v-else
+        :style="'padding-bottom:' + ratioPadding"
+        class="k-card-icon"
+      >
         <k-icon v-bind="icon" />
       </span>
       <figcaption class="k-card-content">
@@ -41,7 +41,6 @@
         />
       </slot>
     </nav>
-
   </figure>
 </template>
 
@@ -68,28 +67,25 @@ export default {
     target: String,
     text: String
   },
+  data() {
+    return {
+      width: null
+    };
+  },
   computed: {
-    wrapper() {
-      return this.link ? "k-link" : "div";
-    },
-    ratioPadding() {
-      return this.icon && this.icon.ratio
-        ? this.$helper.ratio(this.icon.ratio)
-        : this.$helper.ratio("3/2");
-    },
     imageOptions() {
       if (!this.image) {
         return false;
       }
 
-      let src    = null;
+      let src = null;
       let srcset = null;
 
       if (this.image.cards) {
-        src    = this.image.cards.url;
+        src = this.image.cards.url;
         srcset = this.image.cards.srcset;
       } else {
-        src    = this.image.url;
+        src = this.image.url;
         srcset = this.image.srcset;
       }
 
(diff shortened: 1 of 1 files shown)


bastianallgeier:
I wonder about the performance implications when every card registers its own window.resize event.

distantnative:
Yea, I was worried about this too and still aren't too happy solving this with JS. So far I didn't notice any performance issues.

I can't think of another way though to solve this issue. Currently the behavior is still very broken, loading huge images when displaying only 100-200px wide.

nilshoerrmann:
Check out https://tornis.robbowen.digital/ or something similar if you need a general register for resize events.

distantnative:
@nilshoerrmann Does it make a difference in terms of performance whether each images watch the window resize event or each image watches a central store that watch sthe window resize event?

nilshoerrmann:
The issues are not the number of listeners, DOM interactions that require size calculations are a problem, especially if they happen in quick sequences. If you are using something like Tornis, the window dimensions are calculated once per frame, so the process is throttled but you don't miss a frame. You are not doing the calculation for each listener but only once in the library. So the heavy calculations are moved outside the listeners  – and that helps performance.

distantnative:
@nilshoerrmann But as far as I can see, Tornis doesn't give me an element's width.

WebMechanic:
Please refactor to use the [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) instead of `window.resize` for performance where supported. 
That's everyone except Safari (for now) but WebKit implementatioin is in the making (SF13 TP).
- https://caniuse.com/#feat=resizeobserver

distantnative:
@WebMechanic As long as it’s not supported by Safari we cannot use it

nilshoerrmann:
@distantnative: No, in this case it‘s only the window context. But that‘s fine because your own checks would only be required once per frame and only if the window was resized. The important part is not to rely on the `resize` event while also reducing the amount of size calculations.

By the way, I‘m not sure you should _use_ Tornis. But you should _adapt_ the concept.

distantnative:
Maybe we could simply throttle it like https://stackoverflow.com/questions/5489946/how-to-wait-for-the-end-of-resize-event-and-only-then-perform-an-action

nilshoerrmann:
The simplified idea behind Tornis is this:

```js
let windowWidth;

function watch() {
    let currentWidth = window.innerWidth;

    if (currentWidth !== windowWidth) {
        windowWidth = currentWidth;
        // window was resized, do your magic here
    }

    window.requestAnimationFrame(watch);
}

watch();
```

`requestAnimationFrame` is taking care of the throttling. From my experience the result are better than with the `resize` event which is fired highly inconsistent across browsers.

manuelmoreale:
[lodash](https://lodash.com/) has a _.throttle() function built in. From what I can see, can be bundled together with Vue. Or you can obviously go with something custom using request animation frame.

WebMechanic:
> @distantnative As long as it’s not supported by Safari we cannot use it

You sure can, 'cos it has been **polyfilled** long ago (early 2018) when it was first implemented in Chrome 64. Just pick your poison.

`ResizeObserver` was designed to outperform and replace the resize event for precisely such use cases.
WebKit TP (SF13.1) has **already** [picked up](https://webkit.org/blog/9997/resizeobserver-in-webkit/), so _the native API will kick in_ in the very near future for the general public. You then have the most performant solution you can get _for any browser_, for good. 
Meanwhile a large userbase may already benefit from this native API starting today.

RO is flexible and eventually needs less code to do the observation. You may as well combine it with `matchMedia` (another native API) to handle vertical viewport changes, let Vue write some fancy data attributes, and let CSS handle the rest.

Eventually the polyfill will no longer be needed at all. 
I highly object to add a dependency on yet another random 3rd party library, let alone "wasting" your energy and time with some RAF wizzardry.

bastianallgeier:
I'm closing this. We need to fix this on a more global level at a later point.

PR #2494, opened 2020-03-06
Fix cards overflowing in narrow columns #1666

### LINKED ISSUE #1666: Card Layout in column breaks site width
**Describe the bug**
Setting a file sections with large card layout enabled, inside a 1/3 right hand column breaks the width of the panel.

**To Reproduce**
1. Set two columns in a blue print, left 2/3, right 1/3
2. Set a files section in the right  hand column
3. Set layout to card, ratio to 16/9

My files section for reference:

```
heroimagefiles:
  type:                   files
  template:               imagefocus
  headline:               Images
  layout:                 cards
  size:                   large
  image:
    cover:                true
    ratio:                16/9
    back:                 black
```

**Expected behavior**
Card fills the column width.

**Screenshots**

card-bug

**Kirby Version**
Kirby 3.1.1

**Desktop**
 - OS: macOS High Sierra
 - Opera Browser 58.0.3135.127

distantnative:
## Describe the PR

Source of the bug:
Whenever we do for `.k-cards`

```css
grid-template-columns: repeat(auto-fill, minmax(32rem, 1fr));
```

but the container itself is narrower than `32rem`, the card will actually break out of the container.
To avoid that we can use `min()`:

```css
grid-template-columns: repeat(auto-fill, minmax(min(32rem, 100%), 1fr));
```

With this, `32rem` is only chosen as minimum, if `100%` is more than `32rem`.
Unfortunately, the support for `min()` is not yet everywhere (damn you, Firefox): https://developer.mozilla.org/en-US/docs/Web/CSS/min

However, if we add both lines, older browsers will fallback to the old (buggy) version, but more modern browsers get the fix:

```css
grid-template-columns: repeat(auto-fill, minmax(32rem, 1fr));
grid-template-columns: repeat(auto-fill, minmax(min(32rem, 100%), 1fr));
```

Then there is one more thing: since SASS once had its own `min()` function and still applies it, we need to use the unquote syntax to make sure the CSS `min()` function is used (otherwise we will also get an error message. So we end up with:

```css
grid-template-columns: repeat(auto-fill, minmax(32rem, 1fr));
grid-template-columns: repeat(auto-fill, minmax(#{"min(32rem, 100%)"}, 1fr));
```

I added extensive, annoying comments, so we can make sure to remove the old stuff once LibSASS is fixed and/or `min()` is supported by all browsers.

## Related issues

- Fixes #1666

### DIFF
--- a/panel/src/components/Layout/Cards.vue
+++ b/panel/src/components/Layout/Cards.vue
@@ -16,7 +16,7 @@ export default {
   props: {
     cards: Array
   }
-}
+};
 </script>
 
 <style lang="scss">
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):
Nice one!

Issue #2501, opened 2020-03-07
Database joins error


ronaldtorres:
**Describe the bug**  
When you use innerJoin, leftJoin or rightJoin methods on a query, you will get an `InvalidArgumentException` with the message 'Invalid join type INNER|LEFT|RIGHT'.

**To Reproduce**  
1. Execute a query as:
   ```php
   Db::table('my_table')->select('*')
        ->innerJoin('other_table', 'my_table.id = other_table.id')
        ->all();
   ```
2. See error

**Expected behavior**  
The query result(s)

**Kirby Version**  
3.3.3

**Desktop (please complete the following information):**  
 - OS: Mac OS
 - Browser: All browsers
 - Version: All

distantnative:
Proposed solution from @ronaldtorres: https://github.com/getkirby/kirby/pull/2502

bastianallgeier:
✅

PR #2502, opened 2020-03-07
Fix left, right and inner join methods

ronaldtorres:
Added the missing 'join' word at the end of the join type in order to avoid the join type validation error on the sql class at https://github.com/rtorresn10/kirby/blob/0330c0ae4214cde349710b099a770fffd0971b4e/src/Database/Sql.php#L453

## Describe the PR

This PR fix the error by adding the join missing word to the variable type.

## Related issues

#2501 

- [ ] Passing all unit tests

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -295,6 +295,10 @@ class Query
      */
     public function join(string $table, string $on, string $type = 'JOIN')
     {
+        if (strtolower($type) != 'join') {
+            $type .= ' join';
+        }
+        
         $join = [
             'table' => $table,
             'on'    => $on,


src/Database/Query.php
```diff
@@ -295,6 +295,10 @@ public function select($select)
      */
     public function join(string $table, string $on, string $type = 'JOIN')
     {
+        if (strtolower($type) != 'join') {
+            $type .= ' join';
+        }
```
afbora, line 300: I would prefer like following to support different uses as `xxx join` (already includes `join` word) on `join()` method:

```suggestion
        if (strpos(strtolower($type), 'join') === false) {
            $type .= ' join';
        }
```

lukasbestle:
Thanks, I will keep this in mind when I get to refactoring the `Query` class. Leaving this open until then.

bastianallgeier:
@lukasbestle is this now solved with your db refactor?

lukasbestle:
@bastianallgeier No, this will be part of the second refactoring round.

distantnative:
@lukasbestle Closing this PR (to get a better overview of current things in this repo again). Though made an explicit reference to this solution in the corresponding issue.

PR #2522, opened 2020-03-14
Allow required field when not meeting conditions

afbora:
## Describe the PR

I have implemented the feature with simplest way:

- The syntax of use has not changed
- When the fields used are shown only `when` options match in the panel, I did not make any changes on the panel/vue side.
- When validating while saving, we needed other fields data to check `when` fields.

````yaml
fields:
  foo:
    ...
  bar:
    ...
  headline:
    type: text
    required: true
    when:
        foo: design
        bar: architecture
````

## Related issues

- Closes https://github.com/getkirby/ideas/issues/234

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Form/Field.php
+++ b/src/Form/Field.php
@@ -24,6 +24,20 @@ use Kirby\Toolkit\V;
 class Field extends Component
 {
     /**
+     * An array of all found errors
+     *
+     * @var array|null
+     */
+    protected $errors;
+
+    /**
+     * Parent collection with all fields of the current form
+     *
+     * @var \Kirby\Form\Fields|null
+     */
+    protected $formFields;
+
+    /**
      * Registry for all component mixins
      *
      * @var array
@@ -37,14 +51,7 @@ class Field extends Component
      */
     public static $types = [];
 
-    /**
-     * An array of all found errors
-     *
-     * @var array
-     */
-    protected $errors = [];
-
-    public function __construct(string $type, array $attrs = [])
+    public function __construct(string $type, array $attrs = [], ?Fields $formFields = null)
     {
         if (isset(static::$types[$type]) === false) {
             throw new InvalidArgumentException('The field type "' . $type . '" does not exist');
@@ -54,13 +61,13 @@ class Field extends Component
             throw new InvalidArgumentException('Field requires a model');
         }
 
+        $this->formFields = $formFields;
+
         // use the type as fallback for the name
         $attrs['name'] = $attrs['name'] ?? $type;
         $attrs['type'] = $type;
 
         parent::__construct($type, $attrs);
-
-        $this->validate();
     }
 
     /**
@@ -225,8 +232,17 @@ class Field extends Component
         ];
     }
 
+    public function formFields(): ?Fields
+    {
+        return $this->formFields;
+    }
+
     public function errors(): array
     {
+        if ($this->errors === null) {
+            $this->validate();
+        }
+
         return $this->errors;
     }
 
@@ -247,7 +263,7 @@ class Field extends Component
 
     public function isInvalid(): bool
     {
-        return empty($this->errors) === false;
+        return empty($this->errors()) === false;
     }
 
     public function isRequired(): bool
@@ -257,7 +273,7 @@ class Field extends Component
 
     public function isValid(): bool
     {
-        return empty($this->errors) === true;
+        return empty($this->errors()) === true;
     }
 
     /**
@@ -271,6 +287,46 @@ class Field extends Component
     public function model()
     {
         return $this->model;
+    }
+
+    /**
+     * Checks if the field needs a value before being saved;
+     * this is the case if all of the following requirements are met:
+     * - The field is saveable
+     * - The field is required
+     * - The field is currently empty
+     * - The field is not currently inactive because of a `when` rule
+     *
+     * @return bool
+     */
+    protected function needsValue(): bool
+    {
+        // check simple conditions first
+        if ($this->save() === false || $this->isRequired() === false || $this->isEmpty() === false) {
+            return false;
+        }
+
+        // check the data of the relevant fields if there is a `when` option
+        if (empty($this->when) === false && is_array($this->when) === true) {
+            $formFields = $this->formFields();
+
+            if ($formFields !== null) {
+                foreach ($this->when as $field => $value) {
+                    $field      = $formFields->get($field);
+                    $inputValue = $field !== null ? $field->value() : '';
+
+                    // if the input data doesn't match the requested `when` value,
+                    // that means that this field is not required and can be saved
+                    // (*all* `when` conditions must be met for this field to be required)
+                    if ($inputValue !== $value) {
+                        return false;
+                    }
+                }
+            }
+        }
+
+        // either there was no `when` condition or all conditions matched
+        return true;
     }
 
     public function save(): bool
@@ -302,7 +358,7 @@ class Field extends Component
         $this->errors = [];
 
         // validate required values
-        if ($this->isRequired() === true && $this->save() === true && $this->isEmpty() === true) {
+        if ($this->needsValue() === true) {
             $this->errors['required'] = I18n::translate('error.validation.required');
         }
 
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -59,7 +59,7 @@ class Form
             }
 
             try {
-                $field = new Field($props['type'], $props);
+                $field = new Field($props['type'], $props, $this->fields);
             } catch (Throwable $e) {
                 $field = static::exceptionField($e, $props);
             }
--- a/tests/Form/FieldTest.php
+++ b/tests/Form/FieldTest.php
@@ -804,4 +804,107 @@ class FieldTest extends TestCase
 
         $this->assertEquals($expected, $field->errors());
     }
+
+    public function testWhenRequired()
+    {
+        $page = new Page(['slug' => 'test']);
+
+        Field::$types = [
+            'foo' => [],
+            'bar' => [],
+            'baz' => [],
+        ];
+
+        $fields = new Fields([
+            'foo' => [
+                'type'  => 'foo',
+                'model' => $page,
+                'value' => 'a'
+            ],
+            'bar' => [
+                'type'  => 'bar',
+                'model' => $page,
+                'value' => 'b'
+            ],
+            'baz' => [
+                'type'  => 'baz',
+                'model' => $page,
+                'value' => 'c'
+            ]
+        ]);
+
+        // default
+        $field = new Field('foo', [
+            'model' => $page,
+        ]);
+
+        $this->assertSame([], $field->errors());
+
+        // passed (simple)
+        // 'bar' is required if 'foo' value is 'x'
+        $field = new Field('bar', [
+            'model' => $page,
+            'required' => true,
+            'when' => [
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by afbora** [comment]
@distantnative @lukasbestle Have you ever had a chance to look at this?

**Comment by distantnative** [comment]
Just had a quick look, no testing myself. But I would say this is addressing the problem where it should be, so that is good. `isSavable` is a bit misleading in my opinion since there might be more validations that are not met and so the form is actually not savable after all.

**Comment by afbora** [comment]
Thank you for looking quickly @distantnative. To improve the method, can you give an example/condition that can give wrong results in this method please?

**Comment by afbora** [comment]
> might be more validations that are not met and so the form is actually not savable after all.

@distantnative I couldn't fully understand. But if any of the conditions are not met, the `isSaveable()` method will return `true` and the form will be saved.

https://github.com/getkirby/kirby/blob/73e3ad35345dfad46e4d057e032c01f6f5b855f0/src/Form/Field.php#L268-L270

**Comment by afbora** [comment]
If there is no point I missed, I can say that it works properly in single and multiple conditions 👍

**Comment by afbora** [comment]
I think it would be great if we could bring up the highest rated feature in the ideas repo to version 3.4 🎉 Last train 🚋 😅

**Comment by distantnative** [comment]
@afbora my comment was only in regard to the name. If I understand it correctly `isSavable` can return `true` but some validations might still not be met. The code still works as intended (showing errors) but my concern is if the name “is savable” fits - cause the field might still be not savable when invalid.

**Comment by afbora** [comment]
@distantnative Do you have any suggestions about name or process for better understanding please?

**Comment by afbora** [comment]
Did you have a chance to look at this PR please?

**Comment by distantnative** [comment]
I finally had the chance to take a look at it.

1. I rest my case, `isSavable` seems like a good name to me now too :D 
2. I think we should add one more test, where `when` condition is something like "a must be foo-a" and "b must be foo-b" and only one of them is matched
3. Still wondering if injecting the `input` now into every field could become a problem performance-wise. @lukasbestle, you usually have a much better knowledge in that field than me.

**Comment by afbora** [comment]
@distantnative Thank you for reviews 👍 

I understand your concern. Do you think it would work if we remove it again in `destructor` like that in `src/Form/Field.php` class?

````php
public function __destruct()
{
    if (isset($this->attrs['input']) === true) {
        unset($this->attrs['input']);
    }
}

**Comment by afbora** [comment]
So should we convert this to `when > then`? @distantnative

File: src/Form/Field.php [hunk 1]
```diff
@@ -255,6 +255,30 @@ public function isRequired(): bool
...
+                            return true;
+                        }
+                    }
+                }
+            }
+
+            return false;
+        }
+
+        return true;
```
lukasbestle: As the `if` above needs all three conditions to be `true`, this line will make the method return `true` if `$this->save()` is set to `false` – in which case the field is definitely not saveable. But we actually don't want to know if the field can be saved, but if it fulfills the criteria for the `required` option.

What about `needsValue()` as the method name? The return values would of course need to be swapped.

Could you please add a docblock to this method as well? I know the other methods don't have them, but the goal is to add them to all methods over time.

src/Form/Field.php
```diff
@@ -255,6 +255,30 @@ public function isRequired(): bool
...
+        if ($this->isRequired() === true && $this->save() === true && $this->isEmpty() === true) {
+            // check the data of the relevant fields if there is a when option
+            if (empty($this->when) === false && is_array($this->when) === true) {
+                $input = $this->attrs['input'] ?? [];
+
+                if (empty($input) === false) {
+                    foreach ($this->when as $field => $value) {
+                        // if the input data doesn't have data for when fields
+                        // or doesn't match, that means is not required and return true
+                        if (($input[$field] ?? '') !== $value) {
```
lukasbestle: This is a clever hack, but I'm not sure if it's a good idea to rely on that logic to be true. Whether a value is in the `$input` is just an implementation detail of the `Form` class in the end.

I'd say the most solid solution would be to pass the `$fields` collection to the `Field` constructor so that each field has access to all other fields directly. It can then ask the other fields whether their value matches.

src/Form/Field.php
```diff
@@ -255,6 +255,30 @@ public function isRequired(): bool
...
+                            return true;
```
lukasbestle: Doesn't this early return make this comparison an `OR` comparison (= only one of the `when` fields needs to match)? Currently our `when` conditions are `AND` conditions, aren't they?

lukasbestle:
> @distantnative 3\. Still wondering if injecting the `input` now into every field could become a problem performance-wise. @lukasbestle, you usually have a much better knowledge in that field than me.

In that *field*? 😝

As scalar values like strings get passed around by value and not by reference, this could indeed become a memory issue when the field contents are very very large. But I wouldn't see this as a practical issue as a bit of text won't break the memory limit on its own. But as wrote above, we should check if we can find a completely different solution for this comparison problem that won't rely on the `$input` array.

src/Form/Field.php
distantnative: Good idea! Probably way more robust, but for sure easier to understand later on.

distantnative:
> In that field? 😝

Unintended pun by myself. But I take the credit 😝

src/Form/Field.php
afbora: Yes, `when` the option works with the AND operator.

This line means that if one of conditions doesn't match, field is not required and can be saved. Because all conditions must be met.

src/Form/Field.php
afbora: `needsValue()` is more accurate name 👍

afbora:
Do we continue the same way or are we considering adding the `then` option?

We have to decide this now. Because if a user starts using `required` and `when` together with this PR merged, break change will happen when we add `then` feature.

Also if we are going to add `then` in this PR, I will need to ask @distantnative for help in the panel ui side 😅

src/Form/Field.php
afbora: I liked the fiction if I understood correctly, but the point that stuck in my head is this:

We will send all the fields into the constructor of the field, but the first field will not know about the next field. Because the second field has not yet been added/appended in the `$fields` parameter sent to the first field.

Perhaps there may be issues with strict comparisons, but it was more accurate to do business on `input` values.

I'm thinking wrong?

lukasbestle: The `$fields` variable is an object, which is always passed by reference by PHP. So all properties like the array of fields inside the `$fields` object are only stored in one place and not copied. This means that all fields will have access to the full collection of fields as long as the *access* inside the field happens after all fields have been initialized (= accessing other fields inside the `Field` constructor will indeed not work, but in other methods that are called later it will).

src/Form/Field.php
lukasbestle: Right, I forgot to think around two corners here. 😄

Could you add an explicit comment about this to the code to make it obvious when we read this code in the future?

lukasbestle:
> We have to decide this now. Because if a user starts using `required` and `when` together with this PR merged, break change will happen when we add `then` feature.

Why is that?

afbora:
@lukasbestle Let me explain with sample

User can use `required` and `when` together after this PR:

````yaml
fields:
  foo:
    ...
  bar:
    ...
  headline:
    type: text
    required: true
    when:
        foo: design
        bar: architecture
````

Then it will be used separately with `then`:

````yaml
fields:
  foo:
    ...
  bar:
    ...
  headline:
    type: text
    when:
        foo: design
        bar: architecture
    then:
        required: true
````

If we don't keep the old functioning (this PR after `then` feature), it won't work properly. It can also cause confusion in use 🤔

File: src/Form/Field.php (hunk 2 above)
Comment (afbora) [reply src/Form/Field.php]: But our `validate()`  method that  works within the constructor. I tried but couldn't be successful. I'm still not sure 🤔

https://github.com/getkirby/kirby/blob/3.4.0/src/Form/Field.php#L63

Can I leave this implement to you, after last commit in my local? @lukasbestle

Comment (lukasbestle) [reply src/Form/Field.php]: We could likely move this `validate` call to the `errors` method, which would lazily call it if the errors are not already defined.

**Comment by lukasbestle** [comment]
@afbora Good point! Depending on how the when/then feature will be implemented, we could still have this implementation around in parallel. Maybe it could even be the default behavior for `required` when only a `when` without `then` is used. Otherwise the `then` would be expected to define the behavior.

**Comment by afbora** [comment]
How does it look now after your valuable feedbacks? ❤️

File: src/Form/Field.php [hunk 4]
```diff
@@ -247,7 +263,52 @@ public function isEmpty(...$args): bool
...
+                if (empty($input) === false) {
+                    foreach ($this->when as $field => $value) {
+                        $formFields = $this->formFields();
+
+                        // check fields value if exist, otherwise use direct input data
+                        if ($formFields && $formFields->{$field}()->exists()) {
+                            $inputValue = $formFields->{$field}()->value();
+                        } else {
+                            $inputValue = $input[$field] ?? '';
+                        }
```
afbora, line 295: Or we can write. If there are form fields, don't check the field existence.

````
if ($formFields !== null) {
    $inputValue = $formFields->{$field}()->exists() ? $formFields->{$field}()->value() : '';
} else {
    $inputValue = $input[$field] ?? '';
}
````

afbora:
> Depending on how the when/then feature will be implemented, we could still have this implementation around in parallel. 

@lukasbestle I have no idea what way i should take to go 🙏

afbora:
Here we go 🚀 

I think I took care of the first round of the https://github.com/getkirby/ideas/issues/257 `then` feature. This was also necessary for the current PR to avoid break change in future. We'll support current `when` option syntax that have only fields.
In the second round, combining and displaying other features defined in `then` and `otherwise` on the UI (width, placeholder, etc..)

I'm aware that it is a little complicated. I just wonder that can we normalize the `when` option more simply with your valuable feedbacks?

**Sample new usage:**

````yaml
fields:
  foo:
    label: Foo
    type: select
    width: 1/2
    required: true
    options:
      design: Design
      architecture: Architecture
      photography: Photography
      3d: 3D
      web: Web

  bar:
    label: Bar
    type: select
    width: 1/2
    translate: false
    options:
      design: Design
      architecture: Architecture
      photography: Photography
      3d: 3D
      web: Web

  headline:
    type: text
    when:
      - test:
          foo: design
          bar: architecture
        comparison: or
        then:
          show: false
        otherwise:
          required: true
          show: true
````

lukasbestle:
@afbora Thank you!

To be honest, this is a pretty massive feature now, so maybe we should move this to 3.5.0 and then implement the whole thing at once.

The alternative would be to just implement the `required` fix now without the `then` feature and split off the other changes from this PR to a new PR that will then be part of the 3.5.0 feature. As I wrote above, it should be possible to implement the `then` feature without a breaking-change later by supporting the `required` behavior as a default when `then` is not used.

afbora:
I agree with everything you say @lukasbestle, thank you. I would have preferred to publish it to 3.4.x first, rather than storing it to 3.5.0. 

PR seems ready 🚀

lukasbestle (approved):
I have made some minor changes to this PR and squashed all commits to one.

Also I have made a backup of the when/then implementation and I'm currently creating a new branch for that.

afbora:
This is like the finishing touch of an artist. You're great, thank you so much @lukasbestle ❤️

bastianallgeier:
Great work on this one guys!

PR #2523, opened 2020-03-14
Add to top feature for structure field

afbora:
## Describe the PR

When `unshift` is enabled, the adding item is always added to the top of items. 

Option name is inspired by the native method in php and javascript language. Do you have a better and more understandable name idea instead of  `unshift`? 😅 

````yaml
fields:
  social:
    label: On the web
    type: structure
    unshift: true
    fields:
      platform:
        label: Platform
        type: text
        width: 1/2
      url:
        label: URL
        type: url
        width: 1/2
````

## Related issues

- Closes https://github.com/getkirby/ideas/issues/203

## Ready?

- [x] Passing all unit tests


lukasbestle (commented):
I like the feature, but I think the option name is indeed not optimal. `unshift()` is pretty uncreative and confusing naming in PHP and JS (which likely inspired each other), but I don't think we need to continue that trend. ;)

Maybe `newPosition: top|bottom`?

afbora:
Boolean seems more logical to me if it won't be more than 2 options. How is following options?

````yaml
addTop: true   // default: false
addToTop: true // default: false
prepend: true  // default: false
append: false  // default: true (more logical)
````

bastianallgeier:
I have to say that I like "prepend: true" the most so far.

afbora:
I have changed option name `unshift` to `prepend`.

lukasbestle (approved):
I like `prepend: true` as well. 👍

bastianallgeier (approved):

PR #2524, opened 2020-03-15
Add panel shortcuts for page actions

afbora:
### WIP PR

I would open it as a draft, but when editing the PR description, I opened it right away by mistake.

Can I get your feedback? What do you think? @getkirby/kirby-staff 

Dist files for test: [dist.zip](https://github.com/getkirby/kirby/files/4334677/dist.zip)

| Key  | Action |
| ------------- | ------------- |
| CTRL + ALT + X  | Change Status (`S` conflicting with browser) |
| CTRL + ALT + U | Change URL |
| CTRL + ALT + R | Rename |
| CTRL + ALT + C | Copy/Duplicate |
| CTRL + ALT + T  | Change Template |
| CTRL + ALT + D | Delete/Remove |
| CTRL + ALT + L | Switch Language |
| CTRL + ALT + O | Open/Preview |
| CTRL + ALT + N | Next Page |
| CTRL + ALT + P | Previous Page |

## Related issues

- Closes https://github.com/getkirby/ideas/issues/508

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/panel/src/components/Views/PageView.vue
+++ b/panel/src/components/Views/PageView.vue
@@ -44,7 +44,7 @@
           <k-dropdown-content ref="settings" :options="options" @action="action" />
         </k-dropdown>
 
-        <k-languages-dropdown />
+        <k-languages-dropdown ref="languages" />
 
       </k-button-group>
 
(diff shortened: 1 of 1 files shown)


bastianallgeier:
For some reason none of the shortcuts works on macOS

afbora:
Hmm, i can't test on macOs. Do we need to change the shortcut keys? `CTRL + SHIFT` may be 🤔

distantnative:
We should look for meta and ctrl like we do in the textarea https://github.com/getkirby/kirby/blob/master/panel/src/components/Forms/Input/TextareaInput.vue#L226-L227

afbora:
@distantnative I couldn't bind them in `created()` method or do like in textarea toolbar. Is there a trick?

distantnative:
I think we actually need to do some thinking about the shortcuts:

**Next/Previous page:**
Usually you would like to use them to cycle through the pages. The letters N and P are far apart. And I am wondering how many people would think of it like that. On the other hand the arrow keys might already be used by the OS itself in that combination..
CTRL + ALT + N	Next Page
CTRL + ALT + P	Previous Page

Often this combination is used for a full browser reload.
CTRL + ALT + R	Rename

Also a lot lot of these combinations are usually used for special characters such as ≈¨ç†∂@ø®~π, which we would be blocking to use in the Panel then.

Not sure how to deal with these issues or if it's ok to ignore them.

afbora:
These shortcuts make things very fast and enjoyable for people who are constantly making sites with Kirby like us.

I strongly agree that it should not overlap, but as far as I know, browsers generally do not use the combination `CTRL + ALT + *`. I know its usually use the combination of `CTRL + *` like that `CTRL + R Reload`,` CTRL + T New Tab`, `CTRL + N New Window`,` CTRL + P Print`.

We don't have to do `CTRL + ALT`. If you have a more useful idea of a key combination, why not.

lukasbestle:
We could also make the shortcuts a two-step process: Basically there would be one shortcut that enables "keyboard power mode" (like <kbd>Ctrl/Meta-Alt-K</kbd>), which would then sort-of fade out the UI and display a letter on each important UI element. The user could then press that keyboard button to trigger the action behind the UI element.

Advantage besides blocking fewer keyboard shortcuts: Makes it easy to learn the shortcuts because you can always look them up. And if you know exactly what you want to do, you can still rapidly press <kbd>Ctrl/Meta-Alt-K</kbd> and then <kbd>u</kbd> for example to change the URL. Not much slower than pressing <kbd>Ctrl/Meta-Alt-U</kbd> in the first place.

Inside the "keyboard power mode", the <kbd>Esc</kbd> key would exit the mode again.

afbora:
@lukasbestle I liked this idea if it would be a problem to apply in one step.

lukasbestle:
We could definitely make it work with direct shortcuts somehow, but I think that adding a dozen new shortcuts may cause issues later that we don't yet anticipate. Every OS and browser has its different set of reserved shortcuts, so it's probably best to use as few of them as possible for maximum compatibility.

distantnative:
While I like the "keyboard power mode" suggested by you @lukasbestle, I'm a bit afraid how big of a beast it would be to add this UI.

afbora:
I would like to do, but @lukasbestle idea exceeds me

afbora:
Let's close this PR for now? 🤔

PR #2531, opened 2020-03-19
Filter by template for files field

afbora:
## Describe the PR

Now you can filter files by `template` with easily for files field.

**Single template**
````yaml
fields:
  cover:
    type: files
    template: cover
````

**Multiple templates**
````yaml
fields:
  cover:
    type: files
    template: 
      - image
      - cover
````

## Related issues

- Closes https://github.com/getkirby/ideas/issues/495

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/fields/files.php
+++ b/config/fields/files.php
@@ -48,17 +48,6 @@ return [
             return $template;
         },
 
-        /**
-         * If template options defined for filter, disable uploads
-         */
-        'uploads' => function () {
-            if (empty($this->template) === false) {
-                return false;
-            }
-
-            return $this->uploads;
-        },
-
         'value' => function ($value = null) {
             return $value;
         }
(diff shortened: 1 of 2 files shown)


bastianallgeier:
What happens when a new file is uploaded to such a section? Which template is used?

afbora:
You are right, it will be better if the `template` and `uploads.template` option works synchronously.

afbora:
@bastianallgeier When I think again, it would be possible to work together but it would be complicated. Therefore, disabling the uploads when the template was defined seemed the most logical solution.

If it doesn't make sense, we can close the PR 👍

nilshoerrmann:
> What happens when a new file is uploaded to such a section? Which template is used?

Couldn't you open a dialog for the user to choose the template from? Similar to the pop-up when creating a new page?

afbora:
@nilshoerrmann Possible but this may be a little complicated. Is it really worth it? I just wanted to bring the ease of uploads in pages section to pages field.

nilshoerrmann:
@afbora So far, most requests regarding fine-grained filtering using queries or different templates or different parents were always rejected due to the "where will new items reside" question. If a modal enables us to solve this limitation, yes, then it's worth it ;)

afbora:
@nilshoerrmann So let's try

afbora:
I tried to implement and got close to finishing but I think I need help because it requires some advanced Vue 🙏

afbora:
I'm closing the PR due to adding to the core will increase complexity.

PR #2534, opened 2020-03-21
New option prevent redirect to new page for pages section

afbora:
## Describe the PR

Now we can prevent the redirect to the created page with new `redirect` option.

**Sample Usage**

````yaml
sections:
  drafts:
    headline: Drafts
    type: pages
    status: draft
    templates: article
    redirect: false
````

## Related issues

- Closes https://github.com/getkirby/ideas/issues/185

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/panel/src/components/Dialogs/PageCreateDialog.vue
+++ b/panel/src/components/Dialogs/PageCreateDialog.vue
@@ -3,6 +3,7 @@
     ref="dialog"
     :button="$t('page.draft.create')"
     :notification="notification"
+    :data-redirect="redirect"
     size="medium"
     theme="positive"
     @submit="$refs.form.submit()"
@@ -23,6 +24,14 @@ import DialogMixin from "@/mixins/dialog.js";
 
 export default {
   mixins: [DialogMixin],
+  props: {
+    redirect: {
+      type: Boolean,
+      default() {
+        return true;
+      }
+    }
+  },
   data() {
     return {
       notification: null,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: panel/src/components/Dialogs/PageCreateDialog.vue [hunk 1]
```diff
@@ -3,6 +3,7 @@
     ref="dialog"
     :button="$t('page.draft.create')"
     :notification="notification"
+    :data-redirect="redirect"
```
distantnative: What is this line for?

afbora: That's a great question. I put it to get the `redirect` parameter from `PagesSection.vue`, but I guess it has nothing to do with 😄 

If there is no big problem, I would be glad if you can make the last touch 🙏

bastianallgeier:
While I really love this, I have slightly different plans here. I think what we should come up with is a way to customize the "Create Page" model based on page types. I really want a way to add some additional fields to it, customize the slug handling, etc. and one part of this could be the redirect behaviour. 

I was thinking about something like this: 

```
create: 
  fields: 
    customField: 
      label: This can already be filled in when the page is created
      type: text
  slug: "{{ some }}-custom-slug-format-based-on-{{ queries }}"
  redirect: false
```

afbora:
This would be great 👍 I think a page-level blueprint setting is being considered like you've shared there before? https://github.com/getkirby/ideas/issues/119#issuecomment-470523321

Actually, I thought this at the page level first, but I thought it would be more useful as there would be different template options in the section.

So if this feature is not considered for 3.4.0, we can close the PR

PR #2556, opened 2020-04-03
Fix conflicting pagination

afbora:
## Describe the PR

Fixed the issue with separate the `variable` param that more specificed.

## Related issues

- Related https://github.com/medienbaecker/kirby-modules/issues/10

## Ready?

- [x] Passing all unit tests


lukasbestle (commented):
Thanks @afbora for researching this. I was trying to wrap my head around this earlier, but with your PR I now understand what the problem is.

I have implemented this a bit differently while keeping your original commit in this PR in case we need to go back to your implementation.

My implementation disables the `Pagination` class getting its page number from the URL itself for the sections as the page numbers are provided by the section implementations anyway. I have verified that the pagination in the Panel still works with this implementation, but maybe this needs additional tests.

afbora:
@lukasbestle Anyway, my first goal with this PR was to draw attention to the source of the issue. Your solution is more stable than ever 👍 Thank you. I will try to add a test.

afbora:
@lukasbestle Is this test sufficient?

lukasbestle (approved):
I have the feeling that this change could cause issues in the Panel, but from my manual tests with paginating the sections everything seems to work just fine. So either my feeling is wrong or we'll see when the RC is being tested by more people.

bastianallgeier:
I'm afraid there are quite a few more places in the API where this would need to be added.

afbora:
I looked but I couldn't find related usage. Can you help please? @lukasbestle

lukasbestle:
@bastianallgeier I'd suggest to start with these two to see if that breaks anything else. Also the API routes and collections don't have the issue of a conflicting pagination page param as the API is generally called directly and can't conflict "by accident" just because a blueprint is loaded from the frontend (which was the original issue here). So leaving them as they are won't have practical implications. As far as I know, the `pages` and `files` section are the only two blueprint-related objects that have pagination.

lukasbestle:
@bastianallgeier Ping

bastianallgeier (approved):

bastianallgeier:
@lukasbestle sounds like a good suggestion!

PR #2560, opened 2020-04-04
Options API: Use cURL instead of file_get_contents

### LINKED ISSUE #2330: Multiselect: better failure handling for options from API
**Describe the bug**  
We've got a multiselect field that fetches its options via an API call. This stopped working.

**To Reproduce**  
This is our field definition:

```yml
years:
  type: multiselect
  label: Zeitraum
  sort: true
  options: api
  api:
    url: '{{ kirby.url }}/hananils/years'
    fetch: years
  translate: false
```

If called directly, the API route returns this JSON:

```json
{"years":{"2019":"2019","2018":"2018","2017":"2017","2016":"2016","2015":"2015","2014":"2014","2013":"2013","2012":"2012","2011":"2011","2010":"2010","2009":"2009","2008":"2008"}}
```

This is the underlying route in our config:

```php
    'routes' => [
        [
            'pattern' => 'hananils/years',
            'method' => 'GET',
            'action' => function () {
                $years = [];
                $year = date('Y');
                $start = 2008;

                while ($year >= $start) {
                    $years[$year] = (string) $year;
                    $year--;
                };

                return \Kirby\Http\Response::json(['years' => $years]);
            }
        ]
    ]
```

Hard-coding the URL (without the `kirby.url` part) doesn't change anything. This worked in earlier versions. 

**Expected behavior**  
The provided years should be shown as option when using the field.

**Kirby Version**  
3.3.1 but I don't think it's a regression in this specific version.

lukasbestle:
## Describe the PR

`file_get_contents()` is not really meant for HTTP requests and therefore unreliable with some remote servers. Switching to cURL with our `Remote` class should make the requests more reliable.

Thanks for the suggestion @rasteiner.

## Related issues

- Fixes #2330

## Ready?

- [x] Passing all unit tests


src/Form/OptionsApi.php
```diff
@@ -56,13 +58,19 @@ public function options(): array
             return $this->options;
         }
 
-        $content = @file_get_contents($this->url());
+        if (Url::isAbsolute($this->url()) === true) {
+            // URL, request via cURL
+            $data = Remote::get($this->url())->json();
+        } else {
+            // probably a local file
+            $content = @file_get_contents($this->url());
```
nilshoerrmann, line 74: It would be great, if errors weren't supressed anymore because otherwise debugging will still be very hard, see https://github.com/getkirby/kirby/issues/2330#issuecomment-559009300.

afbora:
I agreed with @nilshoerrmann. What if we put into try catch block?

src/Form/OptionsApi.php
lukasbestle: Unfortunately `file_get_contents()` doesn't throw Exceptions, but only triggers warnings, which can only be handled with `set_error_handler()`. But that's quite a hack and therefore not really worth it. Not suppressing the warnings on the other hand will cause extra output in the API responses, which will break the Panel ("JSON could not be parsed"), so it's not useful either.

With the new implementation, most use-cases for the options API will use the `Remote` class and therefore cURL, which throws Exceptions instead that are properly displayed in the Panel. `file_get_contents()` is only kept for backwards-compatibility for use-cases where local file paths were used.

However you are right that no errors at all are bad, so I added two Exceptions manually to cover the most common error cases for local files.

lukasbestle:
> @afbora I agreed with @nilshoerrmann. What if we put into try catch block?

The example you posted in https://github.com/getkirby/kirby/issues/2330#issuecomment-559009300 doesn't do anything unfortunately as it interrupts the Exception chain only to start a new one. `try {} catch {}` is for handling errors (basically for stopping the Exceptions from bubbling up), not for triggering errors.

Also, as I explained in my answer to Nils' comment above, `file_get_contents()` only triggers PHP warnings, which can't be caught by `try {} catch {}`.

But I think my new solution works quite well to ensure that the Panel shows errors for common issues.

bastianallgeier (approved):

PR #2561, opened 2020-04-05
Don't fetch children if there are no more path keys to check

neildaniels:
## Describe the PR

When looking finding a page with an id such as `united-states/california`, the Pages class currently lookups of the children of `united-states` _and_ `california`, even though California's children are irrelevant.

If `california` has a lot of subpages, it's a lot of disk activity and Page Model creations that might ultimately go unused. 

## Ready?

- [x] Passing all unit tests


lukasbestle (approved):
That's a simple but effective improvement, thanks! 👍

PR #2562, opened 2020-04-05
Use create() on the child Page class that is being created

neildaniels:
## Describe the PR

Currently, when a child page is created, it calls the static function `create()` on the class of the parent, not the class of the page being created. This means the create action cannot be overridden by children classes (the only case where the correct override happens is if the child is the same class as the parent).

## Ready?

- [x] Passing all unit tests


lukasbestle (commented):
Good catch! Would you mind adding a unit test for this that verifies that the correct classes are used both for pages with a page model as well as for pages without a page model?

neildaniels:
@lukasbestle Added a test.

(As of writing this, `develop` branch is currently failing Travis, but my test appears to have passed.)

lukasbestle (changes requested):
Thanks! I made a few changes to your test, but unfortunately you have disabled the "allow edits by maintainers" checkbox.

Here's a patch you can commit yourself:

```patch
---
 tests/Cms/Pages/PageCreateTest.php | 6 ++++--
 1 file changed, 4 insertions(+), 2 deletions(-)

diff --git a/tests/Cms/Pages/PageCreateTest.php b/tests/Cms/Pages/PageCreateTest.php
index 84a731d8..a61e2cb6 100644
--- a/tests/Cms/Pages/PageCreateTest.php
+++ b/tests/Cms/Pages/PageCreateTest.php
@@ -9,7 +9,7 @@ class UncreatablePage extends Page
 {
     public static function create(array $props)
     {
-        return null;
+        return 'the model was used';
     }
 }
 
@@ -38,6 +38,8 @@ class PageCreateTest extends TestCase
     public function tearDown(): void
     {
         Dir::remove($this->fixtures);
+
+        Page::$models = [];
     }
 
     public function testCreateDraft()
@@ -164,7 +166,7 @@ class PageCreateTest extends TestCase
             'template' => 'uncreatable-page'
         ]);
 
-        $this->assertNull($child);
+        $this->assertSame('the model was used', $child);
         $this->assertTrue($mother->drafts()->isEmpty());
     }
 
-- 
2.26.0
```

The failing unit tests are indeed not caused by your PR, something went wrong on `develop` today, we are already on it.

neildaniels:
Changes made. Thanks!

I was surprised that the "allow edits" wasn't checked. But, I now learned that PRs from _organization_ repositories do not have that checkbox and cannot enable it. (That's a seemingly reasonable limitation—that could cause deploys to happen, etc.)

lukasbestle (approved):

lukasbestle:
That does make sense. But organization repos are probably not meant for sending PRs anyway.  At least it works. ;)

bastianallgeier (approved):

PR #2563, opened 2020-04-05
New and improved F::load*() methods, prevent access to internal resources by included files

### LINKED ISSUE #2394: No error thrown when accessing private and protected members of Kirby\Cms\App in a plugin
**Describe the bug**  
In a plugin, you can access `kirby()->roots` and no error will be thrown. We had a discussion about that with @distantnative [here](https://forum.getkirby.com/t/workaround-for-a-plugin-setting-before-the-plugin-is-initialized/16846/7?u=hdodov).

**To Reproduce**  
1. Clone the plainkit
2. Create a plugin
3. Put `var_dump(kirby()->roots)` in the plugin _index.php_ file
4. No error thrown

**Expected behavior**  
An error should be thrown.

**Kirby Version**  
3.3.2

**Additional context**  
If you do `var_dump(kirby()->roots)` in a _template_, you correctly get the error:

>Cannot access protected property Kirby\Cms\App::$roots

However, doing the same thing in a plugin works, and it shouldn't. @distantnative [mentioned](https://forum.getkirby.com/t/workaround-for-a-plugin-setting-before-the-plugin-is-initialized/16846/8) this might be due to the way hooks are triggered:

https://github.com/getkirby/kirby/blob/c1ff870e4f13b647f57f99b3e963340c3e0a286c/src/Cms/App.php#L1292

lukasbestle:
## Describe the PR

- New `F::loadOnce()` method for safely including PHP files using `include_once()`
- `F::load()` now supports passing variables to the included file with a new `$data` argument
- Files included with `F::load()` can no longer overwrite the provided `$fallback` variable
- PHP files included by Kirby (e.g. from plugins, models, controllers etc.) can no longer access `protected` methods of Kirby classes

**Note:** The fourth change is a breaking-change.

## Related issues

- Fixes #2394

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -546,7 +546,7 @@ trait AppPlugins
             $class = str_replace(['.', '-', '_'], '', $name) . 'Page';
 
             // load the model class
-            include_once $model;
+            F::loadOnce($model);
 
             if (class_exists($class) === true) {
                 $models[$name] = $class;
@@ -777,7 +777,7 @@ trait AppPlugins
                 continue;
             }
 
-            include_once $entry;
+            F::loadOnce($entry);
 
             $loaded[] = $dir;
         }
--- a/src/Cms/Collections.php
+++ b/src/Cms/Collections.php
@@ -4,6 +4,7 @@ namespace Kirby\Cms;
 
 use Kirby\Exception\NotFoundException;
 use Kirby\Toolkit\Controller;
+use Kirby\Toolkit\F;
 
 /**
  * Manages and loads all collections
@@ -119,8 +120,8 @@ class Collections
         // first check for collection file
         $file = $kirby->root('collections') . '/' . $name . '.php';
 
-        if (file_exists($file)) {
-            $collection = require $file;
+        if (is_file($file) === true) {
+            $collection = F::load($file);
 
             if (is_a($collection, 'Closure')) {
                 return $collection;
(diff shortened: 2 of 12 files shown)


bastianallgeier (commented):
@lukasbestle I like `F::loadOnce`, but `F::loadNative` sounds unintuitive to me. What about `F::loadScoped` or `F::loadIsolated` or another argument for `F::load`?

lukasbestle:
@bastianallgeier I have renamed the method, but after pushing that you got me thinking: Maybe we should indeed merge `F::load()` with `F::loadIsolated()` into one method. I wasn't doing that at first because of your comment in https://github.com/getkirby/kirby/issues/2394#issuecomment-604444878, but in the end it would probably make more sense to have one method that can handle both fallbacks and data.

The method signature would look like this:

```php
public function load(string $file, $fallback = null, array $data = [])
```

What do you think?

bastianallgeier:
I agree that it would be cool to bundle it into one. But I think it would be good if it could keep its current behaviour by default and by injecting data it would become isolated or something like that.

lukasbestle:
It's isolated in all cases because it's in this separate class. Or what do you mean exactly?

bastianallgeier:
Of course it is! 🙈 I just completely mixed this up with the issue of using require or include instead. Forget everything I wrote before 😂

I love the idea of optionally injecting data as third argument! Much cleaner as an additional method.

bastianallgeier (approved):

PR #2578, opened 2020-04-17
Improved handling of key capitalization differences when updating content

### LINKED ISSUE #2577: Untranslatable fields are overly sensitive to Blueprint field's key capitalization
**Describe the bug**  
Programmatically updating a model with untranslatable fields can lead to unwanted results if your code specifies a different key case from those in a Blueprint.

**To Reproduce**  
1. Start with a Blueprint with a field like:
```yaml
SOMEID:
  type: text
  translate: false
```
2. Programmatically create a page with content resembling:
```php
[
  'someid' => 'myidentifer'
]
```
3. Programmatically update the content for a non-default language
```php
$page->update(`['foo' => 'bar'], 'fr-fr);
```
4. The resulting translation's content will resemble:
```php
[
  'someid' => 'myidentifier',
  'foo' => 'bar',
  'SOMEID' => null,
]
```

**Expected behavior** 
I would expect the resulting content to _not_ include `'myidentifier'`. 
```php
[
  'someid' => null,
  'foo' => 'bar',
]
```

**Kirby Version**  
`3.3.5`

neildaniels:
## Describe the PR
Whereas most of model's content usage is independent of key case, updating content is extremely sensitive to case differences in keys. This forces all content keys to be lowercase when updating.

This is _technically_ a breaking change, if you're code relies on specific keys in a ContentTranslation's `->data()` value. I would assume most do not.

Open to other suggestions on approaching this.

## Related issues

- Fixes #2577, #2789

## Ready?

- [x] Passing all unit tests

I didn't change the behavior of the

### DIFF
--- a/src/Cms/ModelWithContent.php
+++ b/src/Cms/ModelWithContent.php
@@ -544,6 +544,7 @@ abstract class ModelWithContent extends Model
         }
 
         // get the content to store
+        $data         = array_change_key_case($data);
         $content      = $translation->update($data, $overwrite)->content();
         $kirby        = $this->kirby();
         $languageCode = $kirby->languageCode($languageCode);
@@ -552,7 +553,7 @@ abstract class ModelWithContent extends Model
         if ($languageCode !== $kirby->defaultLanguage()->code()) {
             foreach ($this->blueprint()->fields() as $field) {
                 if (($field['translate'] ?? true) === false) {
-                    $content[$field['name']] = null;
+                    $content[strtolower($field['name'])] = null;
                 }
             }
 
--- a/tests/Cms/Pages/PageTranslationsTest.php
+++ b/tests/Cms/Pages/PageTranslationsTest.php
@@ -233,6 +233,13 @@ class PageTranslationsTest extends TestCase
                     'b' => [
                         'type' => 'text',
                         'translate' => false
+                    ],
+                    'CAPITALIZED' => [
+                        'type' => 'text',
+                        'translate' => false
+                    ],
+                    'dDdDdD' => [
+                        'type' => 'text',
                     ]
                 ]
             ]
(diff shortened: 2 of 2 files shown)


lukasbestle:
This is difficult.

My first feeling is that the case should always be the one defined in the blueprint, but as we use camelCase there, the case differs from the content files anyway (which use Upper-Dashed-Case).

So maybe the content file case should always be forced to be Upper-Dashed-Case and the blueprint keys should always be expected to be camelCase? In this case we could always map the keys 1:1 with a fixed ruleset.

But: What happens if the blueprint doesn't use camelCase? Should it be converted when parsing?

neildaniels:
@lukasbestle Content files don’t use dashes and only capitalize the first character.

E.g.:
```
Mygreatfield: foo
```

lukasbestle:
Oh, interesting. They did in Kirby 2, so we must have changed that...

Well, lowercase fields make sense then as we don't need camelCase for the dashes when converting from blueprint name to content name. And in the other direction, any non-alphanumerical character would be ignored.

However this probably requires much more deep changes so that all parsing of blueprints and content files is case-insensitive.

neildaniels:
For sure. There very well might be a few other places where this affects, but I think the bulk of this is already handled by the `Kirby\Cms\Content` class:

https://github.com/getkirby/kirby/blob/a7f2f9695509ad6f7e407c5214b0a7aff0b99039/src/Cms/Content.php#L152-L169

It already forces requested methods like `$page->myGreatMethod()` or `$page->content()->get('myGreatMethod')` to lowercase _and_ forces the content's data array to lowercase keys. (Thus, everything is effectively `mygreatmethod` at call-time.)

But, would definitely be good to have feedback if anywhere else would need tweaking. There might be an even better place to adjust the capitalization so it doesn't have to be done at every `content()->get()` call.

lukasbestle:
I will get back to this for a future release, probably 3.5.0.

lukasbestle:
@neildaniels Thanks for your PR and the helpful discussion!

I decided to implement the case-insensitive behavior right in the `Content` class. It already sort-of behaved that way, but not all methods fully did. My implementation is now in https://github.com/getkirby/kirby/pull/2945. I'm therefore closing this PR.

Your unit tests and a part of the implementation made it into the new version, so I added you as a co-author to my commit. Again, thanks for your help! ❤️

PR #2582, opened 2020-04-20
Add 'cache.pages.flush:before' and 'cache.pages.flush:after' hooks

sylvainjule:
## Describe the PR

This PR adds a `cache.pages.flush:before` and `cache.page.flush:before` hook.
I'm actually not sure it's useful to have both `:before` and `:after` in this case but I added them for consistency with other hooks.

It's currently quite tedious to mimick the pages flush behaviour for custom global caches. One has to register all events one by one. This would allow us to register:

```php
'hooks' => [
    'cache.pages.flush:after' => function() {
        kirby()->cache('sylvainjule.my-cache')->flush();
    }
]
```

## Related issues

Haven't opened one.

## Ready?

- [ ] Passing all unit tests

### DIFF
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -115,7 +115,9 @@ trait FileActions
         $kirby->trigger('file.' . $action . ':before', ...$arguments);
         $result = $callback(...$arguments);
         $kirby->trigger('file.' . $action . ':after', $result, $old);
+        $kirby->trigger('cache.pages.flush:before');
         $kirby->cache('pages')->flush();
+        $kirby->trigger('cache.pages.flush:after');
         return $result;
     }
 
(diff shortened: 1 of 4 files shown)


neildaniels:
I really like this, but I think the wildcard hooks provided by https://github.com/getkirby/kirby/pull/2541 will be enough for most use cases.

```php
'hooks' => [
    'page.*:after' => function ($page, $event) {
        // a page modification of any kind was made
        // do additional cache clearing here
    }
]
```

sylvainjule:
Oh you're right, I didn't see this.

My only issue is that instead of:
```php
'hooks' => [
    'cache.pages.flush:after' => ...
]
```

One would need to duplicate code and register separately:
```php
'hooks' => [
    'page.*:after' => ...,
    'file.*:after' => ...,
    'site.*:after' => ...
]
```

This would be solved by the ability to register an array of hooks though:

```php
'hooks' => [
    ['page.*:after', 'file.*:after', 'site.*:after'] => ...
]
```

And this would also be useful for other purposes. So I guess that's another ticket to open and this PR to close.

lukasbestle:
@sylvainjule You can't have an array as the key inside of an array. The only valid key types in PHP are [`integer` and `string`](https://www.php.net/manual/en/language.types.array.php).

However you will also be able to register a `*:after` hook and then check inside of the closure if `$event->type()` is one of `site`, `page` or `file`.

PR #2587, opened 2020-04-24
Form\Field: Support for string validate rule

### LINKED ISSUE #2586: Field validation doesn't support string syntax
The [Kirby Guide](https://getkirby.com/docs/guide/blueprints/fields#validating-fields__using-validators) shows that a field's `validate` property should be an array, like:
```yaml
text:
  type: text
  validate:
    ip
```

However, the examples for the validators themselves show it being a string. This is for the [integer validator](https://getkirby.com/docs/reference/system/validators/ip):
```yaml
fields:
  example:
    label: Example field
    type: text
    validate: ip
```

And it seems like Kirby (3.3.5) refuses any value that isn't an array.
> Argument 2 passed to Kirby\Toolkit\V::errors() must be of the type array, string given, called in /var/www/html/kirby/src/Form/Field.php on line 328

Screen Shot 2020-04-23 at 10 17 04 AM

Also affects _most_ of the other validators, such as [integer](https://getkirby.com/docs/reference/system/validators/integer),  [date](https://getkirby.com/docs/reference/system/validators/date), [email](https://getkirby.com/docs/reference/system/validators/email), etc.

lukasbestle:
## Describe the PR

Our reference has described that it should be possible to use a string as the `validate` rule for a field:

```yaml
fields:
  example:
    label: Example field
    type: text
    validate: ip
```

However this was previously not actually supported. This PR adds support for the string syntax.

## Related issues

- Fixes #2586

## Ready?

- [x] Passing all unit tests


bastianallgeier (approved):

bastianallgeier:
Good catch!

PR #2593, opened 2020-04-28
First round of Database refactoring

### LINKED ISSUE #2234: Sql.php creates invalid SQLite query when using primary key
**Describe the bug**  
I want to create a table using the SQLite functions shipped with kirby. The problem I am experiencing is, it creates invalid queries.

**To Reproduce**  

```php
$_CHARGE = [
  "id"   => [ "name" => "id" ,  "type" => "text", "null" => false, "key" => "PRIMARY KEY" ],
  "time" => [ "name" => "time", "type" => "int",  "null" => false ],
  "card" => [ "name" => "card", "type" => "text", "null" => false ],
];

$db = new Database(array(
  'type'     => 'sqlite',
  'database' => './logs.sqlite'
));

$result = $db->createTable($tableName, $_CHARGE);
```

**Versions**  
PHP version: 7.3.7
SQLite: 3.24.0
Kirby: 3.2.3

**Actual output**  

This is the query that is created:
```sql
CREATE TABLE "transactions" (
  "id" TEXT NOT NULL PRIMARY KEY,
  "time" INTEGER NOT NULL,
  "card" TEXT NOT NULL,
  PRIMARY KEY ("id")
)
```
Which causes an error executing in SQLite:
`Query 1 ERROR: table "transactions" has more than one primary key`

The primary key can not be defined with the column and at the end of the query.

lukasbestle:
## Describe the PR

This is the first round of the Database refactoring. The classes are quite complex, so I didn't want to do everything at once. The changes here are the ones where methods interconnect on the basic/root level. The rest of the refactoring will happen in 3.5.0.

While I was at it, I've also added a few small features and fixed a few bugs.

## List of changes for the changelog

**Features**

- The `$database->createTable()` method now supports columns with a `UNIQUE` constraint
- The `Db::column()` and `$query->column()` database methods now support custom `ORDER` clauses so that you can order the results by any column and not just by the primary key
- Database results can now be fetched using a custom callback:

```php
// prepare a query with the options you need
$query = $database->table('simpsons')->fetch(function ($data, $key) {
    return new Simpson($data);
});

// this line now executes the query and returns an
// array with the return values from your callback
$results = $query->all();
```

**Bug fixes**

- The creation of database tables with primary keys using `$database->createTable()` now works reliably with SQLite #2234
- The Database `fail` mode is now no longer reset after each query, which makes it more predictable as you can't know how many queries an operation needs internally. You can now enable the `fail` mode once for the database connection and it will stay that way until it is manually disabled again.
- Database: The insert ID is now only fetched from the database when an `INSERT` query was executed to improve compatibility with some database servers #2347
- Database: The internal cache of valid table names is now immediately updated after tables are created or dropped, which enables further queries in the same request

**Refactoring**

- The `Database\Db` class and parts of the `Database\Sql` classes have been refactored, which has changed some internal method signatures.
- We have added a lot of unit tests to the Database classes and will further refactor and improve the classes with 3.5.0

## Related issues

- Fixes #2234

## Ready?

- [x] Passing all unit tests


bastianallgeier (approved):

bastianallgeier:
Awesome stuff!

PR #2603, opened 2020-05-06
$field->callback() method

distantnative:
## Describe the PR

Adds `$field->callback(function ($field) {...})`

## Related issues

https://github.com/getkirby/ideas/issues/475

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/tests/Cms/Fields/FieldMethodsTest.php
+++ b/tests/Cms/Fields/FieldMethodsTest.php
@@ -25,21 +25,21 @@ class FieldMethodsTest extends TestCase
     public function testFieldMethodCaseInsensitivity()
     {
         $field = $this->field('test');
-        $this->assertEquals('TEST', $field->upper());
-        $this->assertEquals('TEST', $field->UPPER());
+        $this->assertSame('TEST', $field->upper()->value());
+        $this->assertSame('TEST', $field->UPPER()->value());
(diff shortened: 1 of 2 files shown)


tests/Cms/Fields/FieldMethodsTest.php
```diff
@@ -458,6 +458,16 @@ public function testLength()
...
     }
 
+    public function testCallback()
+    {
+        $field  = $this->field('Hello world');
+        $result = $field->callback(function ($field) {
+            $field->value = 'foo';
+            return $field;
+        });
+        $this->assertEquals('foo', $result);
```
lukasbestle: New tests should use `$this->assertSame('foo', $result->value());`

lukasbestle (changes requested):
In the future, please use one commit per change. Cleanup changes (for unrelated parts of the code!) should be in a second commit in the same PR. This way, it's much easier to track down changes later in the commit history (the cleanup changes can be confusing when using `git blame` or when taking a look at a single commit).

lukasbestle (approved):
Awesome! I have just added the issue reference to the first commit message and added a `@since` marker to the new method. LGTM now. 👍

bastianallgeier (approved):

PR #2605, opened 2020-05-06
$page->go()

distantnative:
## Describe the PR

Implements `$page->go($options, 301)` as shorthand for `go($page->url()`

## Related issues

https://github.com/getkirby/ideas/issues/511

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -454,6 +454,20 @@ class Page extends ModelWithContent
     }
 
     /**
+     * Redirects to this page,
+     * wrapper for the `go()` helper
+     *
+     * @since 3.4.0
+     *
+     * @param array $options Options for `Kirby\Http\Uri` to create URL parts
+     * @param int $code HTTP status code
+     */
+    public function go(array $options = [], int $code = 302)
+    {
+        go($this->url($options), $code);
+    }
+
+    /**
      * Checks if the intended template
      * for the page exists.
      *
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -245,12 +245,12 @@ class Response
      * @param int $code
      * @return self
      */
-    public static function redirect(?string $location = null, ?int $code = null)
+    public static function redirect(string $location = '/', int $code = 302)
     {
         return new static([
-            'code' => $code ?? 302,
+            'code' => $code,
             'headers' => [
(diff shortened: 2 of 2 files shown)


config/helpers.php
```diff
@@ -215,7 +215,7 @@ function gist(string $url, string $file = null): string
  * @param int $code
  * @return void
  */
-function go(string $url = null, int $code = 302)
+function go(string $url = null, ?int $code = 302)
```
lukasbestle: To be honest, I have never understood why `$code` can be `null` in the `Response::redirect()` method. It doesn't really make sense semantically: What's a code of `null`?

So I'd suggest that we should at least keep the stricter function signature here, otherwise we'd cement this weird `null` param even more. The `$page->go()` method would then also accept an integer only with the default value of `302`.

src/Cms/Page.php
```diff
@@ -453,6 +453,18 @@ public static function factory($props)
         return new static($props);
     }
 
+    /**
+     * Redirects to this page,
+     * wrapper for the `go()` helper
+     *
```
afbora, line 459:
```suggestion
     * @since 3.4.0
     *
```

config/helpers.php
distantnative: Hmm... I don't know. I understand that it's a bit weird that I accepts `null` in the `go()` helper. On the other hand, I think the default value should only be defined once and not three times all over the code. So I think the default should be defined in `Response::redirect()` (though it could be defined as parameter default, not in the method body. But for that default to be applied, the other methods need to pass on `null`, if I'm not totally mistaken.

lukasbestle: OK, that probably explains why we did it that way in the first place. Thanks for the explanation!

I see both points and their advantages, but to be honest I still think we should explicitly declare the default value in all places. Otherwise the user will take at `$page->go()` and think "what the heck is the default value?". Finding that out would require following the whole call chain to `Response::redirect()`, where the user will finally find what the default value actually is.

I think there is no need to change this default value ever (which would be a breaking-change anyway), so we might as well declare it everywhere to make it obvious.

distantnative: Ok, good point

lukasbestle (approved):

afbora (approved):

PR #2613, opened 2020-05-09
Collection: Add case-sensitive mode

lukasbestle:
## Describe the PR

In plugins, it can sometimes make sense to have a case-sensitive collection (i.e. a collection where the collection keys are not converted to lowercase automatically). This PR adds such a feature in a backwards-compatible way.

It can be used in plugins like this:

```php
use Kirby\Toolkit\Collection;

class MyCollection extends Collection
{
    public function __construct(array $data)
    {
        parent::__construct($data, true);
    }
}
```

## Ready?

- [x] Passing all unit tests


distantnative:
@lukasbestle won‘t PHP complain if the construct method doesn’t feature the same parameters as in the parent class?

lukasbestle:
@distantnative No, not when it's the constructor. PHP only checks the method compatibility for methods that can be called on an instance, not for the constructor. As you can see, all unit tests pass.

distantnative:
👍

bastianallgeier:
What do you think about the performance implications? Adding additional two additional if clauses for getters and setters in case of hundreds of entries could be an issue. Wouldn't it make more sense to let plugin devs know how to easily overwrite __set and __get if this is needed?

distantnative:
@bastianallgeier has a point - especially considering what a rare use case this probably is for plugin developers. Those who would encounter the option and make use of it certainly would be able to be taught how to overwrite the methods according to their needs.

lukasbestle:
I don't think that an `if` clause that checks an instance property only (without any calculations or even IO) will be noticeable at all. PHP is very well optimized and those instructions will only delay execution by nanoseconds anyway. 

To be honest, I think it's weird that the `Collection` class is case-insensitive by default. I had a strange bug in a plugin I was developing and found out after half an hour of debugging that it was because of the `strtolower()` calls in the `Collection` class. If we wouldn't need to be careful about backwards-compatibility, that's something I would change to be the other way around.

Before I wrote this PR, I actually created a `CaseSensitiveCollection` class for my plugin, but I think that this feature should be part of the core as I imagine that it's not as much of an edge-case actually.

If you disagree, feel free to close this PR though.

bastianallgeier (approved):
You are probably right about the marginal performance hit and I agree that it's a useful feature. Maybe we can even use this to phase-out the case-insensitive version some day.

PR #2614, opened 2020-05-09
Move from App::$root to App::kirbyRoot()

lukasbestle:
Otherwise the static `App::version()` method won't have access to the root if no Kirby instance has been created before `App::version()` is called.

The performance should be unaffected as `dirname()` doesn't need to do any IO calls, so even calling it many times won't make Kirby slower.


distantnative:
I find the name a bit odd. Especially since I’m the class context we talk about app

lukasbestle:
It's not about the `App` class itself, but about the root of the `kirby` directory. So I think the name fits quite well.

bastianallgeier:
I'm with @distantnative here. I'd prefer Kirby::root() here.

lukasbestle:
@bastianallgeier That's already taken.

distantnative:
@lukasbestle but just with passing a parameter, it could return your solution if no parameter is passed, couldn’t it?

lukasbestle:
@distantnative Unfortunately not, because it's an instance method and can't be called statically.

But that gives me an idea:

- We already have a `kirby` root that will be returned by `$app->root('kirby')`. This can be used by all methods that operate on an instance (which are all but `Kirby::version()`). Using the `kirby` root in these places will actually allow customizing the path to system components. We already use the root in three places, so it would make sense anyway to use it everywhere for consistency.
- In `Kirby::version()`, we could basically hardcode the root, which would allow getting rid of the separate `Kirby::kirbyRoot()` method entirely.

What do you think?

bastianallgeier:
@lukasbestle if we can really get rid of it and don't need it anywhere else, it would be amazing!

lukasbestle:
The property we had previously was protected, so it wasn’t used anywhere else.

This PR is ready with the suggested changes.

PR #2618, opened 2020-05-12
Don't allow impersonation in API by default #2616

### LINKED ISSUE #2616: route:before hook login automaticly
**Bug/Feature?**
Use of impersonate() in route:before hook leads to autologin via /panel. 

```
  'hooks' => [
    'route:before' => function ($route, $path, $method) {
      $kirby = $this;
      $this->impersonate('kirby');
    }
  ],
```
After this configuration, anyone can login via /panel without user/password. In theory this should be correct, but i think this is dangerously because of hijacking the admin user? Is this a bug or an intended feature? 
  
-  Kirby 3.3.4
 - OS: osx 10.15.3
 - Browser:  chrome 81 mac

lukasbestle:
## Describe the PR

- The API now ignores users set with `$kirby->impersonate()` by default, which protects it against unintentionally overwriting the current user globally.

## Related issues

- Fixes #2616

## Ready?

- [x] Passing all unit tests


distantnative:
I’m still a bit sceptical whether it’s better to magically take care of this with a (rather complex) setup. Or if we shouldn’t rather actively teach people what it means to use impersonation and that they need to unset it again in a range of use cases. Wondering if it could be counterproductive when users see that they don’t have to take care of inserting it in situations with the API.. but then they don’t think about inserting it for their front end controllers or so. Just thinking out loud here, having firmed a fixed opinion on this.

findthebug:
i think the option is great for the api, but there should also be a clear description in the documentation how to use impersonation. i was running into this and just stumble upon the panel issue. there are many examples on getkirby.com which are using impersonation for page creation and other stuff. it make sense to add a section on the docs / cookbook with one or two good code/api examples how to use it. e.g. in my case i was able so solve the panel issue with setting the user to null after the task is finished.

lukasbestle:
@distantnative We don’t have to explicitly point out in the docs that the API is protected. This makes it a safeguard for those who have opened up the API access by accident (despite the warnings in the docs, which we need to add anyway no matter if we merge this or not), not for those who have read the docs but did it anyway. So I still think that this makes sense, it’s actually not that complex of a setup.

But I just realized that `$api->allowImpersonation` should probably an option instead of a property of the Api object. It’s only internal, but that would be consistent with other stuff. If you agree, I can modify this PR to use an option.

@findthebug Definitely! It’s already on our list to document this more clearly.

bastianallgeier:
This is a pretty massive change. Wouldn't it be possible to set a global const and then block the impersonate method instead?

lukasbestle:
@bastianallgeier To be honest I don't really see which part of this makes this a "massive change". The actual fix is implemented with 12 changed lines of code. The rest is just plumbing because we rely on impersonation for our unit tests. Could you please elaborate what you don't like about the implementation?

> Wouldn't it be possible to set a global const and then block the impersonate method instead?

What do you mean exactly? A global const that defines whether the impersonated user is respected by the API? If so: Where would the const be set? And why a const and not an option like our already existing `api.basicAuth` option?

rasteiner:
not sure if this is the right-place™ for discussion, but what do you people think of an API like this:

```php
<?php

//In user land
$updatedPage = $kirby->doAs('kirby', function() use ($page) {
  return $page->update(['hello' => 'world']);
});


// .... In Kirby App class
public function doAs($user, $action) {
  $this->impersonate($user);
  $returnValue = null;
  try {
    $returnValue = $action();
  } catch(Exception $e) {
    throw $e;
  } finally {
    $this->impersonate(null);
  }
  return $returnValue;
}
```
?

In this way it should be impossible to "forget" the `$this->impersonate(null);` call. I know it's a bit *javascripty*, but I think the idea of "do this as someone else" represents most use cases better than "I am now someone else and then I do this".

lukasbestle:
@rasteiner Very nice idea! Because we need to keep backwards-compatibility, we still need to have support for `impersonate()` (and therefore this PR), but your suggestion is something we can implement in addition. I will do so soon.

distantnative:
How about just allowing the callback as additional second parameter of impersonate?

lukasbestle:
@bastianallgeier I have now changed this PR to use an option instead of the `Api` property, which allowed reducing the complexity in the tests.

The callback feature for `$kirby->impersonate()` is now implemented in https://github.com/getkirby/kirby/pull/2620.

bastianallgeier (approved):

PR #2621, opened 2020-05-18
Allow content field access through model properties

qwerdee:
## Describe the PR

As described in the Issue below this PR tries give a improved way to access field content

## Related issues

https://github.com/getkirby/ideas/issues/552

## Ready?

I think so

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/ModelWithContent.php
+++ b/src/Cms/ModelWithContent.php
@@ -40,6 +40,17 @@ abstract class ModelWithContent extends Model
     public $translations;
 
     /**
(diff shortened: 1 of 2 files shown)


neildaniels:
This seems like it'd be too easy to write code against non-existent properties and it'll never truly fail.

bastianallgeier:
I have to say that I'm not a fan of this solution. I would go for `$model->content()->field()` instead. We never access properties directly and it doesn't really fit the style.

PR #2626, opened 2020-05-27
Bubble exceptions in kirbytags when in debug mode

### LINKED ISSUE #2625: Exceptions in kirbytext rendering are catched but not handled
**Describe the bug**  
Kirbytags only are rendered when there is now error thrown inside. All Errors that occure are caught but not handled. While it is good to ensure editors are not able to break the page with kirbytags, this makes it really hard to develop and debug kirbytags.

https://github.com/getkirby/kirby/blob/21ebebda6bdde9aee62140f81f658f0852ea4e60/src/Text/KirbyTags.php#L26-L30

**To Reproduce**  

Throw an exception:
```php
Kirby::plugin('my/custom-tag', [
    'tags' => [
        'test' => [
            'html' => function (KirbyTag $tag) {
                throw new Exception('something happened');
            },
        ],
    ],
]);
```

In Content:
```
(test: 123)
```

**Expected behavior**  

Perhaps bubble exceptions when debug is set to true.

**Kirby Version**  
3.3.6

qwerdee:
## Describe the PR

Currently kirbytag rendering fails silently when an Error occurs. This makes them difficult to debug. With this PR, Exceptions bubble up when debug is set to true.

## Related issues

- Fixes #2625

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Text/KirbyTags.php
+++ b/src/Text/KirbyTags.php
@@ -3,6 +3,8 @@
 namespace Kirby\Text;
 
 use Exception;
+use Kirby\Exception\InvalidArgumentException;
+use Kirby\Toolkit\Str;
 
 /**
  * Parses and converts custom kirbytags in any
@@ -31,9 +33,22 @@ class KirbyTags
         !isx';
 
         return preg_replace_callback($regex, function ($match) use ($data, $options) {
+            $debug = $options['debug'] ?? false;
+
             try {
                 return static::$tagClass::parse($match[0], $data, $options)->render();
+            } catch (InvalidArgumentException $e) {
+                // stay silent in production and ignore non-existing tags
+                if ($debug !== true || Str::startsWith($e->getMessage(), 'Undefined tag type:') === true) {
+                    return $match[0];
+                }
+
+                throw $e;
             } catch (Exception $e) {
+                if ($debug === true) {
+                    throw $e;
+                }
+
                 return $match[0];
             }
         }, $text);
--- a/tests/Text/KirbyTagsTest.php
+++ b/tests/Text/KirbyTagsTest.php
@@ -3,6 +3,7 @@
 namespace Kirby\Text;
 
 use Exception;
+use Kirby\Exception\InvalidArgumentException;
 use PHPUnit\Framework\TestCase;
 
 class KirbyTagsTest extends TestCase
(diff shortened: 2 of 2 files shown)


src/Text/KirbyTags.php
```diff
@@ -33,7 +34,12 @@ public static function parse(string $text = null, array $data = [], array $optio
         return preg_replace_callback($regex, function ($match) use ($data, $options) {
             try {
                 return static::$tagClass::parse($match[0], $data, $options)->render();
+            } catch (InvalidArgumentException $e) {
+                return $match[0];
             } catch (Exception $e) {
+                if (option('debug') === true) {
```
lukasbestle: To increase portability, this should use the provided `$options` array:

```suggestion
                if (($options['debug'] ?? false) === true) {
```

src/Text/KirbyTags.php
```diff
@@ -33,7 +34,12 @@ public static function parse(string $text = null, array $data = [], array $optio
...
+                    throw $e;
+                }
```
lukasbestle, line 50: There should be an empty line after this closing brace.

lukasbestle (commented):
Thanks for the issue and the PR! I fully agree on the issue and I like your fix. 👍

I've added two minor comments. We also need a unit test for this. Would you mind adding one? If not: No problem, I can pick the PR up from here.

src/Text/KirbyTags.php
```diff
@@ -33,7 +34,12 @@ public static function parse(string $text = null, array $data = [], array $optio
         return preg_replace_callback($regex, function ($match) use ($data, $options) {
             try {
                 return static::$tagClass::parse($match[0], $data, $options)->render();
+            } catch (InvalidArgumentException $e) {
```
lukasbestle, line 40: Why is the InvalidArgumentException in particular still swallowed? I think it could be useful to throw all Exceptions in debug mode to make them easy to debug. But maybe I just don’t see the use-case for this exception (pun intended). 😅

qwerdee: I was trying hard, to avoid this pun.. glad you did it tho 😂
About the Exception and its use case is: It is not unlikely to have potential KirbyTags in your content, which are not defined and therefore should be just ignored. This currently works, by throwing an InvalidArgumentException: https://github.com/getkirby/kirby/blob/21ebebda6bdde9aee62140f81f658f0852ea4e60/src/Text/KirbyTag.php#L42
It might be pretty distracting to have those errors be thrown, even in debug mode, because it is expected behavior. On the other hand, this Exception is widely used throughout the app and it is indeed not ideal to swallow them all. Might be a good idea to a new specific Exception Class here instead, what do you think?

qwerdee:
Thanks for looking into this and your feedback. I agree and made a new commit. 
I haven't had a look into the unit test structure here so I would appreciate if you could take over on this.

src/Text/KirbyTags.php
lukasbestle: That makes sense! 👍

I don’t think we should add another Exception class as we would end up with potentially hundreds of classes for different cases.

What we could so is to check for the Exception message and only swallow that exact Exception.

As I’m going to work on the tests anyway, I will implement that today or tomorrow.

lukasbestle (approved):
Thank you for your contribution! 👍

bastianallgeier (approved):

PR #2627, opened 2020-05-29
Updated supported version in security.md

bastianallgeier:
## Describe the PR

We should keep the supported version up to date more often. This fixes it.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,16 +1,3 @@
 # Security Policy
 
-## Supported Versions
-
-| Version  | Supported          |
-| -------  | ------------------ |
-| 3.2.3+   | :white_check_mark: |
-| 2.5.12+  | :white_check_mark: |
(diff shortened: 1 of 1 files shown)


lukasbestle (commented):
Maybe we should add „until 31.12.2020“ to v2 while we are at it.

lukasbestle:
Oh, and we could use a yellow warning emoji instead of a checkmark for v2.

lukasbestle:
💡 Idea: Because we also need access to the supported version from the planned update checker, it‘s probably better to create a page “security support” on getkirby.com. We can then link to that from security.md. I will create the page today or tomorrow.

lukasbestle:
@bastianallgeier @distantnative I have updated the `SECURITY.md` with a link to the security page.

distantnative (approved):

PR #2630, opened 2020-06-02
Improvements to the Query class

### LINKED ISSUE #2624: Queries fail silently when a page is missing
**Describe the bug**  
When you query a page that doesn't exist, the rest of the instructions carry on using the previous result.

**To Reproduce**  
1. Create a pages field with the query `site.page("foo").children.flip`
2. Make sure a page `foo` doesn't exist
3. You will see the _site_ children flipped

**Expected behavior**  
Since the page doesn't exist, it's expected that the query fails and throws an error. If I create the page `foo`, it works as expected. However, if that page doesn't exist, the query is interpreted as `site.children.flip`, which can lead to unexpected results.

**Kirby Version**  
3.3.6

lukasbestle:
## Describe the PR

To make sure that the fix for #2624 doesn't cause issues in other situations, I took the time to refactor the `Query` class, and while I did that, I have also added a few useful features and improvements:

### Features

- Query: Support for escaping quotes with `\"` and `\'`:

```
{{ site.title.or("This is a default value with \"quotes\"") }}
```

- Query: Support for calling closures in arrays:

```php
$query = new Query('closure(123)', [
    'closure' => function ($argument) {
        return $argument;
    }
]);
echo $query->result(); // 123
```

- Query: Support for accessing properties in objects (used if no method with the same name exists and only if no arguments are passed in the query)

### Enhancements

- Errors in queries are now clearly thrown as Exceptions instead of being swallowed

### Fixes

- If a part of a query returns `null` or a scalar value (number, string or boolean), Kirby now correctly blocks further calls on the return value instead of using the return value of the previous part #2624
- The query language now blocks several edge-cases where the query didn't fully match the underlying data structure (for example calling an array element with arguments even though it is not callable)

## Related issues

- Fixes #2624 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Toolkit/Query.php
+++ b/src/Toolkit/Query.php
@@ -23,8 +23,8 @@ class Query
 
     const NO_PNTH = '\([^\(]+\)(*SKIP)(*FAIL)';
     const NO_SQBR = '\[[^]]+\](*SKIP)(*FAIL)';
-    const NO_DLQU = '\"[^"]+\"(*SKIP)(*FAIL)';
-    const NO_SLQU = '\'[^\']+\'(*SKIP)(*FAIL)';
+    const NO_DLQU = '\"(?:[^"\\\\]|\\\\.)*\"(*SKIP)(*FAIL)';  // allow \" escaping inside string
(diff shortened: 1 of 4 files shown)


bastianallgeier (approved):
That's awesome!! <3

PR #2632, opened 2020-06-05
Replaced YAML en-/decoding with Data equivalents.

tobiastom:
## Describe the PR

This pull request replaces all occurrences of direct calls to `YAML::encode` and `YAML::decode` with their `Data`-class equivalents.

This changes make sure the correct data handler is used for every YAML related conversion. Previously the handler could be changed, but it was not used constantly in every place of Kirby.

## Ready?

- [x] Passing all unit tests


tobiastom:
I think there is nothing I can do agains the Pull Request  build failure. Looks like a timing issue inside the `ApcuCacheTest` test.

lukasbestle (approved):
Thank you, that's a nice idea! 👍

distantnative:
@lukasbestle what do you think about this: instead of putting If-Statements all over the place, couldn’t the Data class check whether the parameter is already an array and in that case just reiten it immediately?

lukasbestle:
@distantnative `Data::decode()` only accepts strings and I think it should stay that way:

If we’d allow an array argument there, we’d also need to allow it for all handlers, otherwise it would be more confusing than before. The YAML handler already supports that for reasons I don’t know, but I don’t think we should copy that historical behavior to all handlers (which includes handlers defined in plugins that would also need to support that).

distantnative:
I just hate to see the same if-Statement all over the place, seems like an overlook by the method if you need to wrap the method call pretty much everywhere in the same If-statement. Would love to keep this more DRY somehow.

lukasbestle:
To be honest I think this is a non-issue. We are talking about six `if` clauses without any complex logic.

The `Data` methods do exactly what they are supposed to and including the check there would not make sense semantically. Having the `if`s right where they are needed makes the methods easier to read because it's clear that the `Data::decode()` method only gets called for non-arrays. Plus the reasons I have listed above.

distantnative:
🤷‍♂️ agree to disagree

tobiastom:
I was also not very happy with the if statements all over the place, but I also thought it would be better to keep the `Data` class type safe.

tobiastom:
I there anything I can help to finish the last failing check?

lukasbestle:
@tobiastom All checks pass. The remaining branch check will never run for this PR as you are an external contributor. But the PR check has succeeded and that‘s what counts.

This PR is ready to merge in my opinion, but we‘ll have to wait what Bastian thinks about this.

tobiastom:
Ok, thank you for the update @lukasbestle.

bastianallgeier:
I'm torn here. I understand both sides, but I really wonder if the behaviour of YAML::decode isn't the more Kirby-ish way. It's a balance between type safety and comfort and checking for arrays for the umteenth time just to keep type safety in one method also seems strange to me. It's also very inconsistent right now.

lukasbestle:
I agree that we have an inconsistency issue here. So we will likely have to decide for one way, otherwise that will continue to haunt us. And if we do decide on one way, it would be the `Yaml::decode()` way with the built-in check, otherwise it would be a breaking-change for those who rely on that.

The issue I see is just that it's not just one method, but an abstract class with several classes that depend on that. All of these would have to be changed (OK, not really a huge task) and the question is if third-party `Data` classes in plugins have also depended on the previous behavior. Those would need to be changed as well.

If you all agree to do it that way anyway, let's do it.

tobiastom:
If it would help, I can offer to contribute to refactor the `Data` classes to also accept an array. I personally don’t like it, but if it helps to get this pull request merged, I would do it.

lukasbestle:
@bastianallgeier I‘d say this is something for you to decide.

bastianallgeier:
I haven't thought about it, but it would indeed introduce breaking changes in both ways. Maybe it's too much, but what about introducing a new method? Something like `Data::decodeAny()`? This new method could also take care of undecodable values.

lukasbestle:
You are right, a separate method would be a more robust and clean solution. And it would only have to be implemented in the abstract `Handler` class, not in the different handlers. At the same time, we could deprecate the current method signature of the `Yaml::decode()` method (so that it will only accept a string in the future like the other handlers).

Regarding naming (which, as we all know, is one of the two hard problems besides cache invalidation and off-by-one errors): What about `decodeAuto`? Basically the method *auto*matically turns the input value into the decoded form.

bastianallgeier:
To be quite honest, neither decodeAny or decodeAuto seems super awesome and the right solution, but I also don't really care that much in this case as long as it is solving our problem in a good way and isn't too far off. So both versions would work for me unless anyone else has a fantastic third method name suggestion :)

distantnative:
If you aren’t happy with it, wondering if anything in the `toXXX` family could work.

tobiastom:
I think the best of both worlds might be `decode`, which accepts anything (so remove the type hint) and add `decodeString`, which is the static typed version. In the end `Data` implementations would only need to implement that one, and use a trait, or an common abstract class, to reuse the `decode` method.

bastianallgeier:
@tobiastom the point was that we can avoid a breaking change when we implement it the other way around.

tobiastom:
Sorry, I somehow overlook that.

lukasbestle:
I also don't really care about the name for this to be honest as there's no "obvious" name that would fit perfectly.

However I think that the method name should start with `decode` to keep it consistent with the other methods of the `Data` classes. `to*` is completely different from what we already have.

So I guess that @tobiastom now has the honor to decide which of `decodeAny` or `decodeAuto` (or another alternative) he thinks fits best. :)

bastianallgeier:
I will take over and fix this until the RC tomorrow.

bastianallgeier:
I've created my own PR for this: https://github.com/getkirby/kirby/pull/2663

After checking out the handler code some more, I realized that our discussion was pretty useless. The abstract Handler::decode method actually has never been type safe and thinking some more about it it makes sense. Depending on the data type it might make sense that a decoder or encoder is able to handle other types of data, i.e. objects and decide for itself if it can handle that data type or not. 

That's why I decided to move forward differently and handle the type checks manually within each handler child class. The YAML child class already had that before, but all the others didn't That's quite inconsistent. Now they all take care of it and `Data::decode()` can simply pass down whatever it gets. I think this is the best fix for this issue, without risking any breaking changes. The worst-case scenario is that a custom handler doesn't do those checks yet and might run into unexpected issues, but it has been the same as before this PR. 

@lukasbestle @tobiastom could you have another look at it today?

tobiastom:
@bastianallgeier Your pull request looks good to me. Thank you for taking the time to implement it and hopefully ship it with the next release.

PR #2633, opened 2020-06-06
Update check (backend part)

lukasbestle:
### Already implemented functionality

- Kirby will by default request the security status from the Kirby site and cache the data for three days.
- The check can be disabled with the `update.kirby` option (there could be `update.plugins` later) or with the global `update` option. Both options can either be `notify` (check automatically and notify in the Panel; default), `manual` (display check button on the Settings page in the Panel) or `false` (completely disabled).
- The resulting data is provided to the Panel via the API. The `GET /api/system` route includes a new `updateStatus` key (which will be an array if the option is set to `notify` or `null` if `manual` or disabled). For manual checking, there's a new `GET /api/system/update` route that will always return the information.
- The data structure from the API is the following:

```
{
  "updated": <timestamp of last cache update>,
  "status": "ok | outdated | at-risk",
  "severity": "minor | notable | major", // when status=at-risk
  "current": "x.y.z",
  "latest": "x.y.z",
  "incidents": [...] // only the ones that affect the currently installed version
}
```

### Panel features that can be implemented based on this PR

- When the Panel loads, it should check for the `system.updateStatus.status` value. If `outdated`, the Panel should display an update notification. If `at-risk`, the notification should be displayed more prominently (red, with warning icon?). If `system.updateStatus` is `null`, no notification should be displayed.
- There should be a "Check for updates" button on the Settings page in the Panel if the `update.kirby` or `update` option is not set to `false`. Hitting the button will request `GET /api/system/update`.
- The Settings page in the Panel can use the `updateStatus` data to display the version information and the incidents that affect the currently installed version.

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/api/models/System.php
+++ b/config/api/models/System.php
@@ -1,6 +1,7 @@
 <?php
 
 use Kirby\Cms\System;
+use Kirby\Exception\PermissionException;
 use Kirby\Toolkit\Str;
 
 /**
@@ -67,6 +68,13 @@ return [
         'kirbytext' => function () {
             return $this->kirby()->option('panel.kirbytext') ?? true;
         },
+        'updateStatus' => function (System $system) {
+            try {
+                return $system->updateStatus();
+            } catch (PermissionException $e) {
+                return false;
+            }
+        },
         'user' => function () {
             return $this->user();
         },
@@ -112,6 +120,7 @@ return [
             'slugs',
             'title',
             'translation',
+            'updateStatus',
             'user' => 'auth',
             'version'
         ]
--- a/config/api/routes/system.php
+++ b/config/api/routes/system.php
@@ -2,6 +2,7 @@
 
 use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
+use Kirby\Exception\LogicException;
 
 /**
  * System Routes
@@ -30,6 +31,19 @@ return [
                     'type'   => 'model'
                 ];
             }
+        }
+    ],
+    [
+        'pattern' => 'system/update',
+        'method'  => 'GET',
+        'action'  => function () {
+            $data = $this->kirby()->system()->updateStatus(true);
+
+            if ($data === false) {
+                throw new LogicException('Checking for updates is disabled in the config');
+            }
+
+            return $data;
         }
     ],
     [
--- a/src/Cms/AppCaches.php
+++ b/src/Cms/AppCaches.php
@@ -73,7 +73,8 @@ trait AppCaches
      */
     protected function cacheOptions(string $key): array
     {
-        $options = $this->option($cacheKey = $this->cacheOptionsKey($key), false);
+        $default = $key === 'system' ? true : false;
+        $options = $this->option($cacheKey = $this->cacheOptionsKey($key), $default);
 
         if ($options === false) {
             return [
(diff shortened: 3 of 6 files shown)


distantnative:
That’s awesome! Serves very well what’s needed for the Panel implementation. Just for my understanding- what results in which `severity` value and how should that be treated differently?

lukasbestle:
@distantnative The `at-risk` status is generated when the current version is affected by one or more security incidents that are listed on https://getkirby.com/security. The incidents can have the severities `minor`, `notable` and `major` (which we can manually assign on the security page). The `severity` on the top-level of the update status object is the highest severity value of all incidents that affect the current version.

It could for example be used to display the update warning differently based on the severity (for example a notification for an available update, a highlighted notification for a minor incident and a modal dialog for a notable or major incident).

distantnative:
Awesome! I'll try to create something already on a merge branch with this and the current `storybook` branch

distantnative:
If the option is disabled, can we provide that information to the Panel (distinct form manual, at the moment both seem to provide `null`)? That way, I could directly show/hide the section on the settings page.

lukasbestle:
@distantnative Doesn't the Panel have access to the options from the config? If not: What about `updateStatus: false` for disabled and `updateStatus: null` for `manual`?

distantnative:
No, not automatically: https://github.com/getkirby/kirby/blob/master/src/Cms/Panel.php#L101-L122

I think `false` for disabled would be good.

lukasbestle:
✅

distantnative:
Current state for settings (WIP - but I'd appreciate your input):
Screen Shot 2020-06-07 at 12 26 40

lukasbestle:
Looks great! ❤️

distantnative:
While implementing the notifications, I thought these should only be shown to admins (or maybe those who have access to the settings view). I would implement this on the Panel side, but thought if it's maybe a bit securer to also prevent the backend from spilling that info out to regular users. What do you think?

lukasbestle:
Good point! I'd say the notifications should be displayed to everyone who has access to the settings view as that can be configured for each role.

I have implemented it that way now, but I feel like the implementation of the permission check is not as clean as it should be. Feedback welcome.

distantnative:
How about actually putting it into the `updateStatus` method instead of the routes? If you don’t have the permission, you only get `false` as if it has been deactivated.

lukasbestle:
Of course. 🤦‍♂️
I've now implemented it as a `PermissionException` that automatically gets converted to a `false` value for the `GET /api/system` route.

lukasbestle:
@distantnative I just realized: The `GET /api/system` route currently also leaks the license key and the current version number to every user, even those without access to the settings view.

Maybe we should add a new `panel-settings` view to the system model that will have all information and a reduced `panel` view without `version` and `updateStatus` and where `license` is just a boolean (registered `true` or `false`).

I guess this will require changes in the frontend and it's completely separate from this, but what do you think in general? If you agree, it would probably make sense to make the changes together with the update check as both rely on the `system` data.

distantnative:
@lukasbestle Given the UI will only work with storybook, let’s also move the backend part to 3.5 - what do you think?

lukasbestle:
@distantnative Yep, sure!

distantnative:
@lukasbestle I have stored a local copy. I would suggest you do the same and we close this PR and branch until we have brought the actual code base to the state where this becomes relevant again. What do you think?

PR #2635, opened 2020-06-08
User: Invalidate OPcache after updating user data

### LINKED ISSUE #2634: User credential changes are cached by OPcache
**Describe the bug**  
User credentials (email, language, name, role) are store in an `.php` file in the accounts directory. Because they're stored as a PHP file, they can fall under the control of OPcache.

If `opcache.validate_timestamps` is set to true, this won't cause an issue.

However, in my Production setup, I _disable_ `opcache.validate_timestamps` for performance reasons. The result of this is: changes to user credentials are silently "ignored" until I reset the server process.

**To Reproduce**  
1. Have `opcache.validate_timestamps = 0` in your `php.ini` file.
2. Go to an existing user's page in the Panel
3. Change this existing user's email address (note: changing it via the Panel or directly on disk are somewhat distinct here; see below)
4. Refresh the Panel page

**Expected behavior**  
The existing user's page should always reflect what is on disk.

In the situation of using the Panel, I would expect that `writeCredentials()` to follow with an `opcache_invalidate()` call. Thus, something like:
```php
    protected function writeCredentials(array $credentials): bool
    {
        $path = $this->root() . '/index.php';
        if ($result = Data::write(path, $credentials)) {
            opcache_invalidate($path, true);
        }
        return $result;
    }
```

However, the above solution will be ineffective if you make a change directly on disk—or, in my situation, you have multiple Docker instances sharing the same `accounts` directory. (In this situation, only the server that processed the API request to change the credentials would have it's cached; there other instances would never know of the invalidation.)

You _could_ move the `opcache_invalidate` to `readCredentials()` method, and call it every time before the `require`. This seems a bit silly, but should work. (I suspect _manually_ reading the PHP file (i.e. not using require/include) would work, but then you're left with the task of parsing a PHP file.)

I think a more tenable (albeit breaking) solution would be to move away from using a PHP file to store credentials. You can keep an empty file to prevent directory sniffing, but move the actual contents to a separate yaml/json file.

**Kirby Version**  
`3.3.5`

**Additional context**

lukasbestle:
## Related issues

- Fixes #2634.

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/UserActions.php
+++ b/src/Cms/UserActions.php
@@ -331,7 +331,13 @@ trait UserActions
      */
     protected function writeCredentials(array $credentials): bool
     {
-        return Data::write($this->root() . '/index.php', $credentials);
+        $path = $this->root() . '/index.php';
+
+        if ($success = Data::write($path, $credentials)) {
+            F::invalidateOpcodeCache($path);
+        }
+
+        return $success;
     }
 
     /**


neildaniels:
Actually, does this end up being unnecessary. I didn't realize the PHP data handler invalidates the cache on write anyways.

https://github.com/getkirby/kirby/blob/3175afeac359524da0896c4671cc9cffb08eb3e7/src/Data/PHP.php#L81-L92

lukasbestle:
Yes, it does! I forgot about that.

PR #2636, opened 2020-06-08
System: Protect version and license key

lukasbestle:
## Describe the PR

The information should not be leaked to users who don't have access to the settings view in the Panel (where both are displayed). If the user doesn't have access, there is no need to return the sensitive information.

This may require changes in the frontend, I can't assess that.

## Ready?

- [x] Passing all unit tests


bastianallgeier:
I agree that the API System model should take care of that, but I don't think it should also be implemented in the System class.

lukasbestle:
We (@distantnative and I) designed this together in a discussion. In the first version I added all checks directly to the API models, but Nico pointed out that a) the other API models also have their permission checks handled in the underlying model classes and that b) it can make sense to limit access from backend code as well (which is only guaranteed if the model class itself checks for permissions).

The only reason why we couldn't also do it that way with the version is that `$app->version()` is a static method (which doesn't have access to user permissions).

bastianallgeier:
@lukasbestle ok, that makes sense.

bastianallgeier (approved):

PR #2612, opened 2020-05-08
More complex draft token with new content.salt option

bastianallgeier:
## Describe the PR

This PR adds a salt to draft tokens, which are used for preview URLs. The salt can be configured with the new `content.salt` option. 

## Ready?

- [x] Added unit tests for fixed bug/feature
- [x] Passing all unit tests
- [x] Fixed code style issues with CS fixer and `composer fix`
- [x] Added in-code documentation (if needed)

PR #2641, opened 2020-06-15
Use salted content token for media directories

lukasbestle:
## Describe the PR

The new `content.salt` option (https://github.com/getkirby/kirby/pull/2612) is now also used for the `$file->mediaHash()`. This makes sure that media URLs are never guessable, not even if the modification date and filename are known (which can sometimes both be public, for example in deployment setups where all files get written at once and therefore have the same modification date).

Kirby will automatically delete folders with the old names and create new folders with the new names when each file is first accessed.

## Related issues

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -386,6 +386,31 @@ class App
     }
 
     /**
+     * Generates a non-guessable token based on model
+     * data and a configured salt
+     *
+     * @param mixed $model Object to pass to the salt callback if configured
+     * @param string $value Model data to include in the generated token
+     * @return string
+     */
+    public function contentToken($model, string $value): string
+    {
+        if (method_exists($model, 'root') === true) {
+            $default = $model->root();
+        } else {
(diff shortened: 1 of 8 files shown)


bastianallgeier:
Doesn't this add a performance overhead? Let's say you create a page with a gallery of 20 images or more and for all those the hash has to be calculated. I'm not entirely sure how fast sha1 hash calculation is though.

lukasbestle:
It's not a password hash that gets calculated in many layers, it's only a single hash. So the performance impact is negligible.

To verify this, I have run a `for` loop with 10 million calls to `hash_hmac()` and it consistently took about 19.5 seconds on my machine. That's 0.0000195 seconds (or 0.0195 milliseconds or 1.95 microseconds) per execution.

In these performance regions, security > performance IMO.

bastianallgeier (approved):

PR #2644, opened 2020-06-17
New files.read permission

lukasbestle:
## Describe the PR

It is now possible to hide files of a specific template from users with a specific role (or to hide files in general). This PR is required for the new media protection feature in #2646.

**Open question:** The `ModelWithContent::panelImage()` method currently does not respect the new `files.read` permission. So if the configured (or default) query returns an image that is not readable for the current user, the image is displayed anyway. The question is: Should the Panel not display an image at all in this situation (which would be easy to implement in two lines of code) or should it try to search for an alternative image (which would be hard to implement with custom queries)?

## Related issues

## Ready?

- [x] Passing all unit tests


bastianallgeier (approved):

bastianallgeier:
Awesome! About the open question: I don't think we can handle all those edge cases in queries at the moment without adding too much complexity. I would leave it like this for now.

PR #2646, opened 2020-06-17
createMedia and protect options for file blueprints

lukasbestle:
## Describe the PR

There are two new options for file blueprints:

```yaml
title: Large File
createMedia: false
```

This will prevent Kirby from copying the files to the `media` folder. The file will instead be passed through the router on each request. Please note that this is much less efficient than than letting the web server handle the requests (which only works when the files are copied to the `media` folder). Also it is not possible to create thumbs from images that are not in the `media` folder.

```yaml
title: Confidential File
protect: true
```

Protected files are never copied to the `media` folder (`protect` implies `createMedia: false`), additionally Kirby will check user permissions whenever the file is accessed. The file will only be downloaded if the current user has the `files.read` permission on that file template.

## Related issues

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -75,7 +75,10 @@ return [
      * @return \Kirby\Cms\File|\Kirby\Cms\FileVersion
      */
     'file::version' => function (App $kirby, $file, array $options = []) {
-        if ($file->isResizable() === false) {
+        if (
+            $file->isResizable() === false ||
+            $file->blueprint()->createMedia() === false
+        ) {
             return $file;
         }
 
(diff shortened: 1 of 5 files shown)


distantnative:
I think the general idea and implementation is valid.
But I have a bad feeling with the option names: our blueprints are becoming more and more complicated, users don't have a grasp of all the options, especially with more and more options on the root level.

I know this isn't a problem specific to this PR but it's why I'm not feeling super comfortable with it.

lukasbestle:
I agree! With blueprints it's super hard to find the right place for options and even then they get more and more complex.

Unfortunately I don't see a better way. Do you have an idea? Not implementing features at all just because blueprints would get more complex isn't really viable either.

bastianallgeier:
To be honest, I'd like to move the decision for this to another release again. I just can't find the time to have a proper look at it right now.

bastianallgeier:
Don't be mad at me, but I would rather close this PR and discuss a new implementation at some point. In this form, I don't feel good about adding another option to the blueprints like this.

PR #2648, opened 2020-06-18
New date comparison validator, new date collection filters, alias `..` for the `between` filter

lukasbestle:
## Describe the PR

There are now date-specific collection filters that can be used like this:

```php
$collection->filterBy('date', 'date >=', 'today');
```

The advantage of these filters is that all date formats that can be parsed by PHP are supported (both in the fields and in the test string).

The added filters are:

- `date ==`
- `date !=`
- `date >`
- `date >=`
- `date <`
- `date <=`
- `date between` (or `date ..`)

## Related issues

- Implements https://github.com/getkirby/ideas/issues/563

## Ready?

- [x] Passing all unit tests


afbora:
I love this feature ❤️ 

I just wonder that do we need to check the string? For ex:

````php
$date1 = strtotime('today');
$date2 = strtotime('wrong_string'); // return false, so it's zero

var_dump($date1 >= $date2); // return true
````

lukasbestle:
@afbora Thanks, you are completely right! Fixed. 👍

distantnative:
Wondering whether we could DRY the code a bit and maybe move that part to the `V` class.

nilshoerrmann:
May I suggest an additional range filter which checks if a date is within an given range?

    $pages->filterBy("date", "date ..", ["-1 month", "tomorrow"])

nilshoerrmann:
Could also be interesting for numeric fields as ".." (without the `date` prefix).

distantnative:
@nilshoerrmann For numeric fields this already exists with the keyword `between` instead of `..` - but I would like this as an alias.

I have added a PR to this PR which moves some of the logic to the `V` class to have it consistent with other collection filters. And it also adds a `date between`/`date ..` filter.

https://github.com/getkirby/kirby/pull/2650

nilshoerrmann:
@lukasbestle & @distantnative Thanks for all these quick additions!

bastianallgeier:
@lukasbestle I'm a bit confused. What is the difference between this and the other PR?

lukasbestle:
This was the original PR and @distantnative made improvements to it in the other one, which merged back to this PR.

bastianallgeier (approved):

nilshoerrmann:
Thanks everyone for this quick enhancement!

PR #2649, opened 2020-06-18
Keep user email as original

### LINKED ISSUE #2600: Umlauts in email field cause 'The JSON response from the API could not be parsed'
**Describe the bug**  
When you try to enter a domain with umlauts in the user email field, it will throw an error.

**To Reproduce**  
1. Go to panel
2. edit an user
3. set its email address to one with an umlaut like `test@exämple.com`
4. you get an error on saving:

```
The JSON response from the API could not be parsed. Please check your API connection.
```

**Expected behavior**  
Changes saved without an error.

**Kirby Version**  
3.3.6

**Console output**  
grafik

 - OS: Windows 10
 - Browser: Firefox
 - Version: 75

**Additional context**  
Related issue which was fixed in the past: #1895

afbora:
## Describe the PR

Decode email before creating user and changing email address like [email validator](https://github.com/getkirby/kirby/commit/03658fa58ae38cefb81a4b309410a4c4f0afe230).

## Related issues

- Fixes #2600

## Ready?

- [x] Passing all unit tests


lukasbestle (commented):
I think this is rather a character encoding issue than a general incompatibility with IDN domains. While your suggestion will fix both potential error sources, it adds complexity and also it would be better to store the email addresses in their original representation and not as IDN (which would be a breaking-change).

The bug that @GiantCrocodile has reported actually sounds more like an UTF-8 vs. ISO-8859-1 issue to me. An indication for this is that it works once the email address is manually entered in the file (= in an editor that probably uses UTF-8).

afbora:
I thought it would be safer to keep non-ascii characters in domain names as ascii. 

Yes this is breaking change that users who have an non-ascii domain email will not be able to login.

afbora:
I'll check again 🖐️

afbora:
@lukasbestle @distantnative  I guess, found the source of issue:

1. I think browsers convert email inputs (type=email) automaticly.

For example login process for `test@exämple.com` email (currently failed with user not found)

**Input type email request header:**

````
array() {
  ["body"]=>
  array(3) {
    ["long"]=>
    bool(false)
    ["email"]=>
    string(23) "test@xn--exmple-cua.com"
    ["password"]=>
    string(8) "********"
  }
}
````

**Input type text request header:**
````
array() {
  ["body"]=>
  array(3) {
    ["long"]=>
    bool(false)
    ["email"]=>
    string(17) "test@exämple.com"
    ["password"]=>
    string(8) "********"
  }
}
````

2. You know that `strtolower` doesn't work with non-ascii chars. Shoud use `mb_strtolower`.

https://github.com/getkirby/kirby/blob/3.3.6/src/Cms/Users.php#L89

---

As a result, login success if use text input and `mb_strtolower` for login. Do you have an idea?

I continue to track down for other places that fail.

**Comment by afbora** [comment]
There is a similar error in the changing email of user, `strtolower` again. `mb_strtolower` works properly! `strtolower` function run after user email change process and broke the non-ascii chars.

https://github.com/getkirby/kirby/blob/3.3.6/src/Cms/User.php#L731

**Comment by lukasbestle** [comment]
@afbora Thanks for tracking this down! ❤️

Regarding `strtolower()`: You are absolutely right, it should be `Str::lower()`, which uses `mb_strtolower()`.

Regarding the automatic conversion by the browser: That's interesting, I hadn't thought of this. I'm not sure if it's a good idea to switch to a `text` field though. @distantnative?

**Comment by afbora** [comment]
> Regarding the automatic conversion by the browser: That's interesting, I hadn't thought of this. I'm not sure if it's a good idea to switch to a text field though. @distantnative?

@lukasbestle I didn't say to switch it, I just said how it works to give you idea 💡

**Comment by afbora** [comment]
@lukasbestle Btw i have learned that this issue was related with HTML not browsers:

https://html.spec.whatwg.org/multipage/input.html#e-mail-state-(type=email)

In general, the solution seems to convert back (test@exämple.com) the punycoded email (test@xn--exmple-cua.com) and using `Str::lower()`.

**Comment by lukasbestle** [comment]
What we could indeed do is to check if the domain of the entered domain starts with `xn---` (in all places where we handle user input, i.e. the registration and login process and when the user email is changed) and normalize the email address to Unicode.

Waiting for @distantnative's opinion on this.

**Comment by afbora** [comment]
I agree! Exactly the opposite of what I applied in PR 😅

**Comment by afbora** [comment]
Now decoding user email instead of encoding. Kirby will save the user email with original always.

File: src/Cms/User.php [hunk 1]
```diff
@@ -244,6 +262,24 @@ public function email(): ?string
         return $this->email = $this->email ?? $this->credentials()['email'] ?? null;
     }
 
+    /**
+     * Encode a user email address
+     *
+     * @param string $email
+     * @return string
+     */
+    public static function encodeEmail(string $email)
```
lukasbestle: This should also be moved to the `Idn` class. It's no longer required for this bugfix, but because you have already implemented it, we might as well ship it for other uses.

src/Cms/User.php
```diff
@@ -234,6 +234,24 @@ protected function credentials(): array
         return $this->credentials = $this->credentials ?? $this->readCredentials();
     }
 
+    /**
+     * Decode a user email address
+     *
+     * @param string $email
+     * @return string
+     */
+    public static function decodeEmail(string $email)
```
lukasbestle: I think this method should be moved to `Kirby\Http\Idn::decodeEmail()`.

tests/Cms/Users/UserTest.php
```diff
@@ -345,4 +345,14 @@ public function testPanelOptionsWithLockedUser()
...
+
+    public function testDecodeEmail()
+    {
+        $this->assertEquals('test@exämple.com', User::decodeEmail('test@xn--exmple-cua.com'));
+    }
+
+    public function testEncodeEmail()
+    {
+        $this->assertEquals('test@xn--exmple-cua.com', User::encodeEmail('test@exämple.com'));
+    }
```
lukasbestle: Needs to be moved to the test class for `Http\Idn`.

tests/Cms/Users/UserActionsTest.php
```diff
@@ -96,6 +104,19 @@ public function testCreateAdmin()
...
+        $user = User::create([
+            'email' => 'test@exämple.com',
+            'role'  => 'admin',
+        ]);
+
+        $this->assertTrue($user->exists());
+
+        $this->assertEquals('test@exämple.com', $user->email());
+        $this->assertEquals('admin', $user->role());
+    }
```
lukasbestle: See the comment regarding the other test in line 54.

tests/Cms/Users/UserActionsTest.php
```diff
@@ -51,6 +51,14 @@ public function testChangeEmail()
...
     }
 
+    public function testChangeEmailContainsUmlaut()
+    {
+        $user = $this->app->user('editor@domain.com');
+        $user = $user->changeEmail('test@exämple.com');
+
+        $this->assertEquals('test@exämple.com', $user->email());
+    }
+
```
lukasbestle: There should also be an assertion for the case where the input is a Punycode string:

```php
$user = $user->changeEmail('test@xn--tst-qla.com');
$this->assertSame('test@täst.com', $user->email());
```

Please also use `$this->assertSame()` for new tests. We are in the process of migrating `assertEquals()` to it.

src/Cms/Auth.php
```diff
@@ -221,7 +221,7 @@ public function login(string $email, string $password, bool $long = false)
         ];
 
         // validate the user and log in to the session
-        $user = $this->validatePassword($email, $password);
+        $user = $this->validatePassword(User::decodeEmail($email), $password);
```
lukasbestle: I think the conversion should be moved to the `validatePassword()` method so that it applies to all ways of using the class.

The `validatePassword()` method then also needs a test for both cases (input is a Unicode string or a Punycode string).

lukasbestle (dismissed):
Thanks for your help, that looks really good already. I only have a few minor comments.

Please also squash your commits into one when you are done and then force-push the new commit to the PR branch. Otherwise the two different implementations may be confusing in the commit history later.

lukasbestle (approved):
I've made [a few minor improvements](https://github.com/getkirby/kirby/compare/b5b2e5811a6c66a8890021aa109952416d377261..a220304303d8fe095466f42bc6f8be8865004639) and squashed the changes into three new commits that are broken up by sub-feature.

Looks great now! 👍

bastianallgeier (approved):

PR #2650, opened 2020-06-19
Date collection filters: use V class + add between filter

distantnative:


src/Toolkit/V.php
```diff
@@ -248,6 +248,21 @@ public static function __callStatic(string $method, array $arguments): bool
                 $date['warning_count'] === 0;
     },
 
+    /**
+     * Parses the first two parameters as dates and then
+     * validates them as specified by the third parameter
+     */
+    'dates' => function ($value, $test, $validator): bool {
```
lukasbestle: The name is very similar to `date`, which just validates a date value. Also the name doesn't really tell what the validator does (could also be a validator that checks that each value in an array is a valid date).

Also a few minor other suggestions:

```suggestion
    /**
     * Compares the two provided values as dates in any format
     * PHP can parse using the provided comparison operator
     */
    'dateCompare' => function ($value, $test, $operator): bool {
```

I'm not 100 % happy with my name suggestion either as the other validators have names such as `startsWith`, so this validator would need to be called something like `dateComparesAgainst`, but that's long.

src/Toolkit/V.php
```diff
@@ -248,6 +248,21 @@ public static function __callStatic(string $method, array $arguments): bool
...
+        $value = strtotime($value);
+        $test  = strtotime($test);
+
+        if (is_int($value) === true && is_int($test) === true) {
+            return V::$validator($value, $test);
```
lukasbestle: The call stack that is behind this line can become quite deep (e.g. `max` calls `V::__call()`, which runs the `size` validator, which calls `is_numeric`) considering that the validator is going to be run on every collection element and considering that the actual comparison is very simple.

So my suggestion would be to use a simple `switch` here with `$operator` options like `>=` (like in the `size` validator) and direct comparison without calling other methods. Another advantage besides being faster: Much, much easier to read and to understand without having to follow the call stack. And because the code is so simple, the benefit of making it DRY is negligible compared to the benefits of using a `switch` directly.

lukasbestle (commented):
Thanks for implementing the improvements. Looks great in general! 👍

Besides the comments below: Could you please move the `..` alias for the `between` filter to a separate commit to keep the two unrelated changes at least a bit separated?

src/Toolkit/V.php
distantnative: I know you like very explicit method names, which I don't always see really fitting the overall naming style of Kirby, e.g. `dates` I think is very much in line with `V::errors`. What about `assertDates`?

[removed comment about `$operator` change - only saw your other comment after initially writing this one]

src/Toolkit/V.php
distantnative: Ok sorry, now I see why you changed it to `$operator` :D

Hmm, makes the method a bit less flexible, but on the other hand it would be in line with the `size` validator, I like that.

src/Toolkit/V.php
```diff
@@ -248,6 +249,36 @@ public static function __callStatic(string $method, array $arguments): bool
...
+        $value = strtotime($value);
+        $test  = strtotime($test);
+
+        if (is_int($value) !== true || is_int($test) !== true) {
+            return false;
+        }
+
+        switch ($operator) {
+            case '==':
+                return $value == $test;
```
distantnative: Just thinking, @lukasbestle - we could use strict comparisons here, couldn't we?

lukasbestle: Definitely!

lukasbestle: The `$operator` values would still be unchanged for consistency with the collection filters, but the comparison itself can be strict.

src/Toolkit/V.php
lukasbestle: I like the idea!

What about `assertDate`? The validator validates that the `$value` validates against an assertion with another date. Together these are two dates, yes, but all other validators are named from the perspective of `$value`. So I think `assertDate` would fit quite well.

bastianallgeier:
I'm just throwing in another idea for naming the validators. If we want to keep them consistent with the rest, we would probably need to split them and use something like: 

```php
// date1 < date2
V::olderDate($date1, $date2) 
// date1 > date2
V::newerDate($date1, $date2)  
// date1 <= date2
V::maxDate($date1, $date2)
// date 1 >= date2
V::minDate($date1, $date2) 
// date 1 == date 2
V::sameDate($date1, $date1) 
```

distantnative:
@bastianallgeier the idea was exactly not to have pretty much the same code again and again 😅
Also looks a lot like `V::size` to me

bastianallgeier:
Ok, I get that. Stupid second idea. How about making `V::date()` more powerful when there's more than just one argument? In this case I would change the order of the arguments though: 

```php
V::date($date1, '<', $date2);
V::date($date1, '<=', $date2);
V::date($date1, '==', $date2);
V::date($date1, '>', $date2);
V::date($date1, '>=', $date2);
```

I think this is super readable and the `V::date` method is a bit underwhelming right now anyway.

distantnative:
I like it

bastianallgeier:
@distantnative yes, me too. The more I look at it, the more intuitive it feels.

distantnative:
@bastianallgeier let‘s do it. Would you do the changes? I came down with a bad stomachache bug/cramps last night and follow today on my phone but don’t really feel ready to sit at a table and open my laptop

bastianallgeier:
@distantnative yep, I will take over

distantnative:
@bastianallgeier thank you! Before merging we should take Lukas’ comment into account and make the checks strict. What do you think?

src/Toolkit/V.php
```diff
@@ -239,13 +239,45 @@ public static function __callStatic(string $method, array $arguments): bool
...
     /**
-     * Checks for a valid date
-     */
-    'date' => function ($value): bool {
-        $date = date_parse($value);
-        return $date !== false &&
-                $date['error_count'] === 0 &&
-                $date['warning_count'] === 0;
+     * Checks for a valid date and is able to compare
+     * dates against each other.
```
lukasbestle:
```suggestion
     * Checks for a valid date or compares two
     * dates with each other.
```

src/Toolkit/V.php
```diff
@@ -239,13 +239,45 @@ public static function __callStatic(string $method, array $arguments): bool
...
+     *
+     * Pass only the first argument to check for a valid date.
+     * Pass an operator as second argument and another date as
+     * third argument to compare them.
+     */
+    'date' => function ($value, string $operator = null, $test = null): bool {
```
lukasbestle: Both `date_parse()` and `strtotime()` only support strings as arguments, so we can be strict here like in some other validators:

```suggestion
    'date' => function (string $value, string $operator = null, string $test = null): bool {
```

src/Toolkit/V.php
```diff
@@ -239,13 +239,45 @@ public static function __callStatic(string $method, array $arguments): bool
...
+            case '<':
+                return $value < $test;
+            case '>':
+                return $value > $test;
+            case '<=':
+                return $value <= $test;
+            case '>=':
+                return $value >= $test;
+            default:
+                return $value == $test;
```
lukasbestle: I wouldn't use the `switch` default branch to be honest. Like in Nico's implementation, the validator should throw an Exception if the operator is invalid. Swallowing such an error and just doing an arbitrary comparison is going to lead to bugs that are hard to debug.

src/Toolkit/V.php
```diff
@@ -239,13 +239,45 @@ public static function __callStatic(string $method, array $arguments): bool
...
+        $value = strtotime($value);
+        $test  = strtotime($test);
+
+        if (is_int($value) !== true || is_int($test) !== true) {
+            return false;
+        }
+
+        switch ($operator) {
+            case '!=':
+                return $value != $test;
```
lukasbestle:
```suggestion
                return $value !== $test;
```

The same applies to `$value == $test`, which would need to be `$value === $test`.

src/Toolkit/V.php
bastianallgeier: You are right. I fixed it.

bastianallgeier (approved):

lukasbestle:
@bastianallgeier Looks great! I have only made two minor changes:

- If an argument may be `null`, the method signature should be `?string $value` and not `string $value = null`. The first example means that the argument needs to be passed, but can be `null`, and the second example means that the argument is `null` by default (which is not what we want here considering the other validators).
- I've changed the `InvalidArgumentException::class` line in the unit test to an explicit string with the class name for consistency with other parts of the code that we have also migrated to this syntax.

lukasbestle (approved):

PR #2651, opened 2020-06-19
UI for update check in Panel

distantnative:


lukasbestle:
@distantnative Not 3.4.0?

distantnative:
@lukasbestle let’s see. It’s already a monster we are having a hard time getting together without breaking things. Not sure if we should throw him just another moving part (that’s for the Panel side). But I’ll leave that call to Bastian and once the refactoring dust as settled.

lukasbestle:
Let‘s move this to 3.4.0 so that it will show up in the milestone list for us to remember. We can later move it to 3.5.0 if needed.

distantnative:
Actually, maybe we should convert this PR to a draft until the storybook branch is merged - and we can be sure that the UI in this PR actually works with the final Panel. What do you think?

lukasbestle:
Definitely. To be honest, I first thought that you have opened this PR by accident. 🙂

distantnative:
I'll store this locally and close it for now until the pre-requisites are in place.
(mainly to better keep an overview in this repo what is currently relevant)

PR #2652, opened 2020-06-19
Merge controller data with original data safely

### LINKED ISSUE #2559: Objects updated from controller are outdated in templates/snippets
**Describe the bug**  

If an object is updated from a controller:

```php
return function($kirby, $page) {
    $kirby->impersonate('kirby');
    $page = $page->update(['test' => uniqid()]);

    return [
        'page' => $page
    ];
};
```

the updated `$page` is not available in the template and in snippets as Kirby [overrides the returned object with the original `$page`](https://github.com/getkirby/kirby/blob/3.3.5/src/Cms/Page.php#L347).

**Expected behavior**  

The updated `$page` should be provided to the template and snippet. It would be awesome if this would automatically happen even if `$page` is not manually returned from the controller.

**Kirby Version**  

3.3.5

**Additional context**  

Original discussion in https://github.com/getkirby/kirby/issues/2555.

afbora:
## Describe the PR

I have an idea about merge controller data safely with original data. Review please @lukasbestle @distantnative 

## Related issues

- Fixes #2559 

## Ready?

- [x] Passing all unit tests


lukasbestle (commented):
The implementation is good if we keep the status quo that all objects are immutable and therefore not updated in global state. The only thing I would add is an `Exception` if the class doesn't match (swallowed error cases are hard to debug).

But the general question is if we can find a way to combine immutable objects and global state. Basically the data of an updated page should be immediately available to other parts of the code (which is basically the opposite of what the point of an immutable object is). Very hard to find a balance here. 🤯

afbora:
@lukasbestle Yeah that looks really hard. So, is it possible/logical that Kirby objects are mutable (critical properties are immutable)? 🤔 

And how would we apply this PR to keep the current status quo for now?

lukasbestle:
To be honest I don't really have an idea for a good solution. Maybe @bastianallgeier can add his opinion to the discussion as the designer of the immutable object structure. I don't want to overturn any of this without his opinion.

Until we have discussed this: Let's wait. In any case, thank you for your PR!

bastianallgeier:
I think we need to separate this. The PR is good because it adds more type safety to the merge. It doesn't fix the issue of immutable objects, but maybe it really doesn't need to do this for now. I would agree with @lukasbestle though that an Exception would be great. @afbora could you take care of that? 

The state of objects in Kirby has always been a pain. I hoped to solve this with the immutable objects and it does make some parts clearer and other parts harder. I don't have a golden solution to that I'm afraid.

src/Cms/Page.php
```diff
@@ -344,7 +345,31 @@ public function controller($data = [], $contentType = 'html'): array
...
+                'pages' => 'Kirby\Cms\Pages',
+                'page'  => 'Kirby\Cms\Page'
+            ];
+
+            foreach ($controllerData as $key => $value) {
+                if (array_key_exists($key, $classes) === true) {
+                    if (is_a($value, $classes[$key]) === true) {
+                        $data[$key] = $value;
+                    } else {
+                        throw new InvalidArgumentException('Passing "' . $key . '" data must be instance of ' . $classes[$key] . ' in controller');
```
lukasbestle:
```suggestion
                        throw new InvalidArgumentException('The returned variable "' . $key . '" from the controller "' . $this->template()->name() . '" is not of the required type "' . $classes[$key] . '"');
```

tests/Cms/Pages/PageTest.php
```diff
@@ -1111,4 +1111,69 @@ public function testPageModel()
...
+                            'title' => 'Test Title',
+                        ]
+                    ]
+                ]
+            ],
+            'controllers' => [
+                'default' => function ($page) {
+                    $page = $page->changeTitle('New Title');
+
+                    return ['page' => $page];
```
lukasbestle:
```suggestion
                    return compact('page');
```

tests/Cms/Pages/PageTest.php
```diff
@@ -1111,4 +1111,69 @@ public function testPageModel()
 
         Page::$models = [];
     }
+
+    public function testPageController()
```
lukasbestle: The test method should be called like the tested method:

```suggestion
    public function testController()
```

tests/Cms/Pages/PageTest.php
```diff
@@ -1111,4 +1111,69 @@ public function testPageModel()
...
+            ]
+        ]);
+
+        $app->impersonate('kirby');
+        $page = $app->page('test');
+        $data = $page->controller();
+
+        $this->assertCount(4, $data);
+        $this->assertInstanceOf('Kirby\Cms\Page', $data['page']);
+        $this->assertSame('New Title', $data['page']->title()->value());
```
lukasbestle: The test should also test the default objects:

```suggestion
        $this->assertSame($app, $data['kirby']);
        $this->assertSame($app->site(), $data['site']);
        $this->assertSame($app->site()->children(), $data['pages']);
        $this->assertInstanceOf('Kirby\Cms\Page', $data['page']);
        $this->assertSame('New Title', $data['page']->title()->value());
```

tests/Cms/Pages/PageTest.php
```diff
@@ -1111,4 +1111,69 @@ public function testPageModel()
...
+    }
+
+    public function testPageControllerInvalid()
```
lukasbestle:
```suggestion
    public function testControllerInvalid()
```

**Note:** Maybe this test could be combined with the `testController` by adding a second page and controller to that test and adding the test code to the end of that test. This would reduce the duplication of the boilerplate code.

tests/Cms/Pages/PageTest.php
```diff
@@ -1111,4 +1111,69 @@ public function testPageModel()
...
+    {
+        $this->expectException('Kirby\Exception\InvalidArgumentException');
+        $this->expectExceptionMessage('Passing "page" data must be instance of Kirby\Cms\Page in controller');
```
lukasbestle:
```suggestion
        $this->expectExceptionMessage('The returned variable "page" from the controller "default" is not of the required type "Kirby\Cms\Page"');
```

lukasbestle (changes requested):

afbora:
I think, the PR is ready to merge now. Thanks for suggestions @lukasbestle 👍

bastianallgeier (approved):

PR #2654, opened 2020-06-22
Allow Symfony Server to run Kirby

sanderdlm:
## Describe the PR

When developing locally I like to use [Symfony Server](https://symfony.com/doc/current/setup/symfony_server.html) as a substitute for the built-in PHP server. The two major features I like about it is the local https and the ability to specify a PHP version per project with a simple file.

Out of the box, Kirby doesn't allow the Symfony server. I've circumvented this by going into the core of my Kirby projects and modifying the allowed server list. I've done this for months now with no issues. Maybe it would be a good idea to have this in the official core so other developers can also use this.

## Related issues

None.

## Ready?

- [ ] Passing all unit tests

### DIFF
--- a/src/Cms/System.php
+++ b/src/Cms/System.php
@@ -413,6 +413,7 @@ class System
             'caddy',
             'litespeed',
             'nginx',
+            'symfony',
             'php'
         ];
 


lukasbestle:
Thanks for your contribution! You are the first user I'm aware of who uses Symfony Server with Kirby, which makes this PR a bit hard to review. I myself have never used Symfony Server either, so I can't really say whether this fully works. I trust you that it does, but it would be great nonetheless to get additional confirmations from other users.

I'm also not 100 % sure if we can include support in the core as it would mean that we would need to support the environment. However we don't actively test with LiteSpeed either and yet it is in the list. So maybe we can add Symfony Server to the list anyway if a few people can confirm that it works reliably.

bastianallgeier:
I wonder if we should turn this into a different type of solution. From time to time someone asks us how to unblock a certain server because Kirby should theoretically work with it. Let's make this a config option and then every user can unblock their favourite server at their own risk.

bastianallgeier:
I've created a new PR which introduces a new server option to overwrite the check: https://github.com/getkirby/kirby/pull/2656

bezin:
I was about to open a new issue requesting to include Symfony Server in the Kirby defaults, but then found this so I add my thoughts here: We use the Symfony Server for every Kirby installation we setup for the last 2 1/2 years and it works flawlessly. The config options works well of course, but I would appreciate if Symfony Server were added to the defaults.

lukasbestle:
@bezin This will get lost here in the closed PR. Could you please create a new suggestion over at https://feedback.getkirby.com? I think it's worth discussing to get rid of the allowlist entirely instead of adding more and more servers to it.

bezin:
Done, see https://kirby.nolt.io/567

PR #2655, opened 2020-06-22
set CURLOPT_SSL_VERIFYPEER to true

### LINKED ISSUE #2653: CURLOPT_SSL_VERIFYPEER disabled in Remote::fetch()
Hello,

I just took a look at `Remote::fetch()` due to some forum question and I noticed that `CURLOPT_SSL_VERIFYPEER` is set to `FALSE` by default. Is there a reason why this security feature is disabled by default? I would have expected SSL cert validation to be enabled by default or to find some information why this is disabled but I found nothing about this.

See this line: https://github.com/getkirby/kirby/blob/6bd14bc8099d6c9dbe67ef0582b8c0a2d4c0888c/src/Http/Remote.php#L152

I'm sorry if this should have been posted in the forum or if it's an obvious misunderstanding on my side.

Best Regards,

GiantCrocodile:
Fixes #2653.

### DIFF
--- a/src/Http/Remote.php
+++ b/src/Http/Remote.php
@@ -149,7 +149,7 @@ class Remote
             CURLOPT_RETURNTRANSFER   => $this->options['body'],
             CURLOPT_FOLLOWLOCATION   => true,
             CURLOPT_MAXREDIRS        => 10,
-            CURLOPT_SSL_VERIFYPEER   => false,
+            CURLOPT_SSL_VERIFYPEER   => true,
             CURLOPT_HEADER           => false,
             CURLOPT_HEADERFUNCTION   => function ($curl, $header) {
                 $parts = Str::split($header, ':');


afbora:
Thanks for the PR. As far as I know, `CURLOPT_SSL_VERIFYPEER` will not work with just enabled. `CURLOPT_CAINFO` or `CURLOPT_CAPATH` must be set for this.

GiantCrocodile:
I'm not entirely sure about this @afbora. According to this https://curl.haxx.se/libcurl/c/CURLOPT_SSL_VERIFYPEER.html it sounds optional to me but I could be wrong. To me it sounds like you can define `CURLOPT_CAINFO` additionally to `CURLOPT_CAPATH` while `CURLOPT_CAPATH` will be defined on building cURL and thus should/must be set already.

afbora:
> When CURLOPT_SSL_VERIFYPEER is enabled, and the verification fails to prove that the certificate is authentic, the connection fails. When the option is zero, the peer certificate verification succeeds regardless.

So `CURLOPT_SSL_VERIFYPEER` disabled default in Kirby i think. 

May be it would be better if these settings can be override from config.

GiantCrocodile:
According to the linked page (official docs) and Debian wiki the CA bundles used to verify the cert are depending on the distribution used and can be quite large. The option is active by default according to these two sentences:

`By default, curl assumes a value of 1.`

`This option determines whether curl verifies the authenticity of the peer's certificate. A value of 1 means curl verifies; 0 (zero) means it doesn't.`

I think this issue has three tasks:

1. Document that this option is internally used in Kirby 3 and what it does
2. Enable this by default (as stated and expected in the cURL docs)
3. Make it easier to find out and more non-function specific used (I'm not sure why it's set specific for `fetch` method but not e. g. `get` method)

For task 3, it could mean we have global cURL options like these security ones respected in whole Kirby 3 core.

lukasbestle (commented):
You are absolutely right – it's not good practice to disable certificate validation, so we should indeed make sure to enable it by default.

As @afbora has pointed out, cURL needs a CA bundle, which is present on some servers, but some have a very outdated one or none at all.

Because Kirby needs to work on all kinds of servers, we cannot rely on the system CA bundle – which is probably the reason why `VERIFYPEER` is currently disabled by default. However we could ship Kirby with the [cURL CA bundle as packaged by the Composer team](https://github.com/composer/ca-bundle). This library has about 240 KB of unzipped data (130 KB zipped) and I think this is worth it for the increase in security.

My suggestion is: Kirby should try to use the system CA bundle by default (with `getSystemCaRootBundlePath`, which falls back to the included bundle automatically). There should also be an option though to always use the included bundle and ignore the system bundle completely ( `getBundledCaBundlePath`) for servers with outdated CA bundles.

I don't think it should be possible to disable the verification completely – at least for now (let's see if someone actually runs into issues that cannot be fixed by using the included up-to-date CA bundle).

> @GiantCrocodile 3. Make it easier to find out and more non-function specific used (I'm not sure why it's set specific for fetch method but not e. g. get method)

The `get()` method creates a new `Remote` instance, which will always call the `fetch()` method. So if we change it in `fetch()`, we change it everywhere.

bastianallgeier:
If we can solve it like @lukasbestle suggested, I'm all up for it. But I feel like the wrong one to implement it. This is definitely not my comfort zone as I have way too little knowledge about SSL certificates and certificate validation.

lukasbestle:
@GiantCrocodile If you like you can have a go, otherwise I will take over with further development.

GiantCrocodile:
Hi @lukasbestle! Feel free to work on this one as I'm busy for the next few weeks with other incidents and work-related tasks. I don't want to hack something together at a security relevant spot just to have something quickly done. If you want to, I can try to test or review it. That's less critical and time consuming. Sorry for that; I didn't expect it to end in a bigger task like this :/.

GiantCrocodile:
@lukasbestle Just something to keep in mind: It would be nice if it's possible to query these security settings via a protected API. E. g. if you want to integrate Kirby 3 in IT security solutions, it would be good if the API exposes this information to make sure security settings are fine and can be checked for (unwanted) changes. This might be something to keep in mind in general for other features of Kirby 3 too.

afbora:
I just wonder that will `CURLOPT_SSL_VERIFYPEER` option be disabled optionally after enabled feature?

lukasbestle:
This PR is replaced by https://github.com/getkirby/kirby/pull/2661.

PR #2656, opened 2020-06-23
New servers option to overwrite server check

bastianallgeier:
## Describe the PR

This PR adds a new "server" option, which can be used to overwrite our internal server check. This way it is now possible to test Kirby with a server setup that is not officially supported. 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/tests/Cms/System/SystemTest.php
+++ b/tests/Cms/System/SystemTest.php
@@ -56,6 +56,22 @@ class SystemTest extends TestCase
         $this->assertEquals($expected, $server);
     }
 
+    public function testCustomServer()
+    {
+        $_SERVER['SERVER_SOFTWARE'] = 'symfony';
+
+        $app = $this->app->clone([
+            'options' => [
+                'server' => 'symfony'
+            ]
+        ]);
+
+        $system = new System($app);
+        $server = $system->server();
+
+        $this->assertTrue($server);
+    }
+
     public function serverNameProvider()
     {
         return [
--- a/tests/Cms/System/SystemTest.php
+++ b/tests/Cms/System/SystemTest.php
@@ -56,10 +56,11 @@ class SystemTest extends TestCase
         $this->assertEquals($expected, $server);
     }
 
-    public function testCustomServer()
+    public function testServerOverwrite()
     {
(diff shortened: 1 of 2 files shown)


afbora (commented):
Something tiny, it seems to me that the plural is more understandable since the option itself is an array 🤔 

    $this->app->option('servers');

bastianallgeier:
@afbora you are right. That makes more sense! I just changed it.

afbora (dismissed):

tests/Cms/System/SystemTest.php
```diff
@@ -56,6 +56,38 @@ public function testServer($software, $expected)
...
+            'options' => [
+                'servers' => ['symfony', 'apache']
+            ]
+        ]);
+
+        $system = new System($app);
+        $server = $system->server();
+
+        $this->assertTrue($server);
+    }
```
lukasbestle: I'd change two things about the tests:

- The names of the test methods should always be sortable (from left to right) and therefore start with the method name that is tested, otherwise it can be very hard to find the relevant test methods. And if we later decide to enable the PHP-CS-Fixer feature to sort methods alphabetically, the test methods would be shuffled completely. So the test method should be called `testServerCustom`.
- In the core classes I've refactored so far, I have merged the test methods as much as possible (meaning: a single unit of code (here: method in the `System` class) should have its own test methods, but as few test methods as possible). Of course merging is not always possible, for example when multiple Exceptions are being tested (only a single Exception can be tested per test method) or when parts of the test use a provider (like the `testServer` method above) and parts don't (like `testServerCustom`). In these situations it's completely fine to use multiple test methods, but in this example there is no need to have both `testServerCustom` and `testServersCustom`. The two tests duplicate code with the setup of setting the `$_SERVER` variable, which is not necessary. Of course there are much much worse tests where it actually really matters performance-wise, but my suggestion is to keep it consistent across the code base. Which would mean: Use fewer separate test methods in general. If we stick to that rule, we can decrease the duration of our test runs by a lot.

bastianallgeier: Makes sense! I've adjusted the tests.

afbora (approved):

PR #2658, opened 2020-06-24
Improve validate on empty field

### LINKED ISSUE #2592: Blueprint fields with `validate` can't be left blank
**Describe the bug**  
I have a few Blueprint fields with a `validate` options. I would have expected that _empty_ fields should skip validation (assuming they're not set to `required: true`). However, empty fields still seem to go through a validation check.

**To Reproduce**  
1. Add a field to a page blueprint, such as:
```yaml
      someId:
        label: Some ID
        type: text
        validate:
          integer
        translate: false
```
or
```yaml
      someCode:
        label: Some Code
        type: text
        minlength: 2
        maxlength: 2
        validate:
          alpha
        translate: false
```
2. Leave these field blank, but make a change to some other field (to force the Save bar to appear).
3.  Try to save the page in the Panel.
4. Get error saying you need to "enter a valid integer" or "only enter characters between a-z".

**Expected behavior**  
If a field is left empty, then validation should be skipped. 

**Kirby Version**  
`3.3.5`

 - OS: macOS
 - Browser: Safari
 - Version: 13.1 (15609.1.20.111.8)

**Additional context**  
According to @texnixe, [in February](https://forum.getkirby.com/t/does-an-email-validator-need-a-required-validator/17116/8):
> A field is only validated with a validator if it has input. To enforce input, you have to require it, because otherwise it can be left empty.

It seems like genuinely empty fields (and not required) _should_ skip validation.

afbora:
## Describe the PR

Now `validate` run only when the field is is not empty or required.

## Related issues

- Fixes #2592 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Form/Field.php
+++ b/src/Form/Field.php
@@ -325,9 +325,14 @@ class Field extends Component
             }
         }
 
-        if (empty($this->validate) === false) {
-            $rules  = A::wrap($this->validate);
-            $errors = V::errors($this->value(), $rules);
+        if (
+            empty($this->validate) === false &&
+            (
+                $this->isEmpty() === false ||
+                ($this->isEmpty() === true && $this->isRequired() === true)
+            )
+        ) {
+            $errors = V::errors($this->value(), $this->validate);
 
             if (empty($errors) === false) {
                 $this->errors = array_merge($this->errors, $errors);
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
Ahhh, that's great!

PR #2660, opened 2020-06-28
Remote: Support setting defaults with App options

lukasbestle:
## Describe the PR

Defaults for the `Http\Remote` class can now be configured in the `site/config.php`:

```php
<?php

return [
    'remote.agent' => 'My Awesome Site'
];
```

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -577,12 +577,17 @@ class App
      * Returns the current App instance
      *
      * @param \Kirby\Cms\App $instance
-     * @return self
+     * @param bool $lazy If `true`, the instance is only returned if already existing
+     * @return self|null
      */
-    public static function instance(self $instance = null)
+    public static function instance(self $instance = null, bool $lazy = false)
     {
         if ($instance === null) {
-            return static::$instance ?? new static();
+            if ($lazy === true) {
+                return static::$instance;
+            } else {
+                return static::$instance ?? new static();
+            }
         }
 
         return static::$instance = $instance;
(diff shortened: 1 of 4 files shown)


bastianallgeier (commented):
I'm not really happy how this integrates the App class. I always hoped to keep each package as separate as possible and it works out quite well except for the Form package, which should probably be in Cms anyway. 

I think a better pattern would be to have a public static Remote::$options property, which could then be set/extended in the App constructor once the config has been loaded. 

Something like

```php
Remote::$options = array_merge(Remote::$options, $this->option('remote', []));
```

bastianallgeier:
Or maybe we are more radical and overwrite Remote::$options entirely when the App is initialized to avoid weird side-effects.

lukasbestle:
I thought about that as well.

We already have `Remote::$defaults` actually, but it‘s global state. As soon as you have multiple different App instances, you will never know which instance overwrites what. So I thought it‘s more reliable to get the options at run-time from the current instance.

The separation of ”Cms and everything else“ made a lot of sense while we had a different Toolkit repo, but that more and more fades together now that we only have one repo.

Getting the App instance lazily avoids side-effects without making it impossible to use the classes outside of the CMS.

So to be honest I think that we should rather migrate our code to this dynamic pattern over time instead of adding new dependencies with the old pattern.

bastianallgeier:
Looking at it again, you are right. We've already optimized a lot of our classes for our use-case and it will probably never end up in an independent framework anyway. But this means that we should actively search for other places where this pattern could help to clean up our classes. I was immediately thinking about the two Url classes or the Str class.

bastianallgeier (approved):

PR #2661, opened 2020-06-28
Remote: Validate TLS certs by default

### LINKED ISSUE #2653: CURLOPT_SSL_VERIFYPEER disabled in Remote::fetch()

lukasbestle:
## Describe the PR

**Note:** This PR depends on #2660.

The `Http\Remote` class now validates the TLS certificate of the remote server by default. Kirby ships with a CA bundle based on the Mozilla CA list. We keep that CA bundle up-to-date.

It is also possible to use a different CA bundle by configuring it in `site/config.php`:

```php
<?php

return [
    'remote.ca' => ...
];
```

Options are:

- An absolute path to a CA file in `.pem` format
- An absolute path to a CA directory with certificate files that cURL can understand
- The special value `Remote::CA_INTERNAL` for Kirby's internal CA (default)
- The special value `Remote::CA_SYSTEM` to use the CA bundle that is configured in `php.ini`
- `false` to disable validation completely (not recommended)

## Related issues

- Fixes #2653

## Ready?

- [x] Passing all unit tests


GiantCrocodile:
Is it possible to combine these options @lukasbestle? It might make sense to have e. g. system's CA bundle + Kirby 3's or one from an URL + Kirby 3's. This would make sure you don't have to build custom CA bundle packages if you need a combination.

lukasbestle:
@GiantCrocodile cURL does not support multiple CA files or directories, there can only be one of each type. So the use-case is pretty limited.

GiantCrocodile:
I see. You are right. It's a limitation in openSSL library used by cURL: https://curl.haxx.se/mail/lib-2012-07/0214.html.

Thank you for the improvement :)!

lukasbestle:
For most use-cases the new implementation should be fine as the bundle that now ships with Kirby includes all CAs that are also supported by browsers. So basically every public domain should be covered. If you do need to make a request to an internal host that uses a self-signed or mTLS certificate, you can pass the correct `ca` option to the `Remote` class just for that request.

GiantCrocodile:
True, excellent argument @lukasbestle. I'm very eager to see 3.4.0 released; although some fixes to the editor plug-in would be nice too :).

bastianallgeier (approved):

PR #2672, opened 2020-07-01
Fix broken toString method in ModelWithContent after new Str::template behaviour 

### LINKED ISSUE #2664: 3.4.0-rc.1: Pages section "info" line handles query parameters differently
**Describe the bug**  
For some pages sections, I have a an `info` line that formats a field as a date. In `3.3.x`, if the underlying field was empty, the resulting info string would be empty. In `3.4.0-rc.1`, it now falls back to showing the literal query string.

```yaml
type: pages
info: '{{ page.created.toDate("M j, Y") }}'
```

Screen Shot 2020-06-30 at 3 34 33 PM

This might be caused by https://github.com/getkirby/kirby/pull/2309 or https://github.com/getkirby/kirby/pull/2639.

If it's https://github.com/getkirby/kirby/pull/2309, I don't think there's any way to specify a "fallback" parameter (of an empty string) from the Blueprint level, but I think `info` and `text` would be good candidates to opt into the prior behavior.

**Kirby Version**  
`3.4.0-rc.1`

bastianallgeier:
## Describe the PR

`ModelWithContent::toString()` doesn't really work well with the new behaviour in Str::template() which leaves the placeholder if the value is empty. This PR fixes this. 

## Related issues

- Fixes #2664

### DIFF
--- a/config/methods.php
+++ b/config/methods.php
@@ -455,9 +455,10 @@ return function (App $app) {
          *
          * @param \Kirby\Cms\Field $field
          * @param array $data
+         * @param string $fallback Fallback for tokens in the template that cannot be replaced
          * @return \Kirby\Cms\Field
          */
-        'replace' => function (Field $field, array $data = []) use ($app) {
+        'replace' => function (Field $field, array $data = [], string $fallback = '') use ($app) {
             if ($parent = $field->parent()) {
                 $field->value = $field->parent()->toString($field->value, $data);
             } else {
@@ -465,7 +466,7 @@ return function (App $app) {
                     'kirby' => $app,
                     'site'  => $app->site(),
                     'page'  => $app->page()
-                ], $data));
+                ], $data), $fallback);
             }
 
             return $field;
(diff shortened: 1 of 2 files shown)


afbora (commented):
I think this is overlooked and this PR is fixes the error 👍 

Well, we use `Str::template()` in different places. 
Do you need fallback in the following places to avoid being overlooked?

https://github.com/getkirby/kirby/blob/release/3.4.0/config/methods.php#L464
https://github.com/getkirby/kirby/blob/release/3.4.0/src/Form/OptionsQuery.php#L64
https://github.com/getkirby/kirby/blob/release/3.4.0/src/Form/OptionsApi.php#L53
https://github.com/getkirby/kirby/blob/release/3.4.0/src/Form/OptionsApi.php#L148

lukasbestle (commented):
I think the most flexible way would be to add a third `$fallback` param to `ModelWithContent::toString()` that defaults to an empty string. The same applies to the `replace` field method. I can't really say much about the `Form` classes as I don't have a full understanding of these classes at the moment.

bastianallgeier:
@afbora I don't think we should replace the behaviour in those places as well. I looked through them and I think the new behaviour does make sense for all of them. But if it ever leads to an unexpected bug we can still fix it then. 

@lukasbestle that makes sense. I will update that.

lukasbestle:
I think the fallback param should be added to the replace field method as well. Besides that this PR is ready to merge.

PR #2677, opened 2020-07-02
Introduce new $go method to handle routing more safely #2665

### LINKED ISSUE #2665: [3.4.0-rc.1] NavigationDuplicated Issue
**To Reproduce**  
1. Go to panel
2. Click to title on top bar menu (for ex: Mægazine for site dashboard)
3. Check browser console

**Screenshots**  

NavigationDuplicated

**Kirby Version**  
3.4.0-rc.1

**Console output**  
````json
{
  _name: "NavigationDuplicated",
  name: "NavigationDuplicated",
  message: "Navigating to current location ("/users") is not allowed",
  stack: "Error↵    at new e (http://localhost/test/kirby/3.…73a113207524b12ced5e281e78/js/vendor.js:23:13600)"
}
````

**Desktop:**  
 - OS: Windows 10
 - Browser: Chrome 83.0.4103.116

### PR DESCRIPTION
**Pull request by bastianallgeier**
## Describe the PR

This PR introduces a new Vue.$go method, which handles routing more safely and catches unnecessary and annoying NavigationDuplicate errors. This new behaviour in the Vue router doesn't seem to be fixable in any other way, which is kind of a shame, but at the same time the $go method feels very Kirbyish and is shorter to write and easier to understand than this.$router.push (push is a weird term in this case anyway)

## Related issues

- Fixes #2665

### DIFF
--- a/panel/src/components/Navigation/Search.vue
+++ b/panel/src/components/Navigation/Search.vue
@@ -150,9 +150,6 @@ export default {
       };
     },
     navigate(item) {
-
-      console.log(item);
-
       this.$go(item.link);
       this.close();
     },
--- a/panel/src/components/Navigation/Search.vue
+++ b/panel/src/components/Navigation/Search.vue
@@ -41,7 +41,7 @@
           :data-selected="selected === itemIndex"
           @mouseover="selected = itemIndex"
         >
-          <k-link :to="item.link" @click="click(itemIndex)">
+          <k-link :to="item.link" @click="close">
             <strong>{{ item.title }}</strong>
             <small>{{ item.info }}</small>
           </k-link>
@@ -150,6 +150,9 @@ export default {
       };
     },
     navigate(item) {
+
+      console.log(item);
+
       this.$go(item.link);
       this.close();
     },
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by afbora** [comment]
> Uncaught (in promise) Error: Avoided redundant navigation to current location: "/site".

I'm getting this new error on console.

image (1)

**Comment by afbora** [comment]
> Uncaught (in promise) Error: Navigation cancelled from "/users" to "/users/w4VKDeaC" with a new navigation.

Getting when search user and click to any result.

**Comment by bastianallgeier** [comment]
Ah, I'm glad you spotted that. There was indeed yet another issue in the search component. I think I fixed it.

File: panel/src/components/Navigation/Search.vue [hunk 1]
```diff
@@ -150,6 +150,9 @@ export default {
       };
     },
     navigate(item) {
+
+      console.log(item);
```
distantnative: This shouldn’t be here :)

bastianallgeier: Argh.

bastianallgeier: Sometimes I wish there was a "stupidity check" before you can push

distantnative (approved):

PR #2693, opened 2020-07-06
Supports accessible phpMailer instance

afbora:
## Describe the PR

The `phpMailer` instance can now be accessed via `callback` option. Also `Exception` text may be more understandable 😅 

### Sample Usage

````php
$kirby->email([
    'from' => 'no-reply@supercompany.com',
    'to' => 'someone@gmail.com',
    'subject' => 'Thank you for your contact request',
    'body' => 'We will never reply',
    'transport' => [
        'type' => 'smtp',
        'host' => 'mail.getkirby.com',
        'port' => 465,
        'security' => true,
        'auth' => true,
        'username' => 'test@test.com',
        'password' => 'randomString',
    ],
    // new feature
    'beforeSend' =>function ($mailer) {
        $mailer->SMTPOptions = [
            'ssl' => [
                'verify_peer' => false,
                'verify_peer_name' => false,
                'allow_self_signed' => true
            ]
        ];

        return $mailer;
    }
]);
````

## Related issues

- Closes https://github.com/getkirby/ideas/issues/121

## Ready?

- [x] Passing all unit tests

### CODE REVIEW

**Comment by afbora** [comment]
I should have made it accessible without using smtp!

**Comment by afbora** [comment]
Waiting discussion about improvements on getkirby/ideas#121 🗣️

File: src/Email/PHPMailer.php [hunk 1]
```diff
@@ -67,6 +68,17 @@ public function send(bool $debug = false): bool
...
 
+        // accessible phpMailer instance
+        $callback = $this->callback();
+
+        if (empty($callback) === false && is_a($callback, 'Closure')) {
+            $mailer = $callback->call($this, $mailer);
+
+            if (empty($mailer) === true || is_a($mailer, 'PHPMailer\PHPMailer\PHPMailer') === false) {
+                throw new InvalidArgumentException('"callback" option return should be instance of PHPMailer\PHPMailer\PHPMailer');
+            }
```
lukasbestle: As the `$mailer` object is passed by reference to the function, the return value should be optional:

- If the return value is `null`, the original `$mailer` value should be used.
- If the return value is anything else, the object type should be validated like in your implementation.

src/Email/Email.php
```diff
@@ -159,6 +166,12 @@ protected function setBcc($bcc = null)
         return $this;
     }
 
+    protected function setCallback(Closure $callback = null)
```
lukasbestle:
```suggestion
    protected function setCallback(?Closure $callback = null)
```

lukasbestle (commented):
Besides the option naming I have two suggestions.

But we should probably wait for @bastianallgeier's feedback. :)

afbora:
Thanks for reviews @lukasbestle I applied the your suggestions and now we can listen to the feedback of @bastianallgeier

bastianallgeier:
I have to withdraw my suggestions. The "beforeSend" option is perfect and the advantage of being able to add this to the config is so much better than my callback function idea. This is really good!

bastianallgeier (approved):

PR #2701, opened 2020-07-10
Fix error page on multilang

### LINKED ISSUE #2609: Error page always rendered in default language
**Describe the bug**
Error pages will always render in the default language.
Only when directly accessed by e.g. `example.com/de/error` it will be in another language.

**To Reproduce**  
1. Enable ~~multitasking~~ multilanguage.
2. Considering `de` is a non-default language, visit a non-existing page like `example.com/de/waldo`.
3. The error page will be in the default language.

**Expected behavior**  
The error page should be translated.

**Kirby Version**  
3.3.5

afbora:
## Describe the PR

First of all, it was really hard to debug this issue. 
The source of the issue is if fails when checked for all languages for the requested page, the default language is called. 
Instead of the default language, first fallback with the current language was tried, if it fails, the default language will work.

There might be another point I missed. I opened PR only for suggestion. If the code is not suitable, it can be closed.

I tried to add an unit test, but I couldn't write a unit test that checked it correctly.

I'm waiting for your reviews. @bastianallgeier @lukasbestle @distantnative 

## Related issues

- Fixes #2609 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/tests/Cms/Languages/LanguageRoutesTest.php
+++ b/tests/Cms/Languages/LanguageRoutesTest.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Kirby\Cms;
+
+use PHPUnit\Framework\TestCase;
+
+class LanguageRoutesTest extends TestCase
+{
+    protected $app;
+
+    public function setUp(): void
+    {
+        App::destroy();
+
+        $this->app = new App([
+            'roots' => [
+                'index' => '/dev/null'
+            ],
+            'languages' => [
+                [
+                    'code'    => 'en',
+                    'name'    => 'English',
+                    'default' => true,
+                    'locale'  => 'en_US.UTF-8',
+                    'url'     => '/',
+                ],
+                [
+                    'code'    => 'de',
+                    'name'    => 'Deutsch',
+                    'locale'  => 'de_AT.UTF-8',
+                    'url'     => '/de',
+                ],
+            ]
+        ]);
+    }
+
+    public function testFallback()
+    {
+        $app = $this->app->clone([
+            'site' => [
+                'children' => [
+                    [
+                        'slug'     => 'error',
+                        'template' => 'error'
+                    ]
+                ]
+            ]
+        ]);
+
+        $app->call('notes');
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
I've done some tests locally and your change successfully fixes the issue for me. It turns out that even this seems to work:

```php
return $kirby->language()->router()->call($path);
```

as Exceptions are caught inside the `LanguageRouter` anyway. Also the `$kirby->language()` method returns the default language if no other language was set.

However it looks like we don't currently have *any* automated tests for the language routes. So it's very difficult to check if we missed any edge case. As I haven't created this part of the code, I don't have a deep understanding of the inner workings of the language routes, so I unfortunately cannot say if this would break something else.

afbora:
@lukasbestle Yes you are right, seems only this line even works since `$kirby->language()` return default language as fallback 👍 

````php
return $kirby->language()->router()->call($path);
````

Then let's ask to @bastianallgeier if we need an `Exception` here or is there anything we missed 👀

**Comment by bastianallgeier** [comment]
The exception is not needed here in my opinion. `App::language()` already handles all the edge cases. If there's a deeper issue here and there's no language available at this point, this part should actually fail and not stay silent. 

P.S. we really need tests for the multi-language routes. I really dread writing them. Multi-language is such a beast.

**Comment by afbora** [comment]
I managed to write a unit test 🎉 I think the PR is ready 🚀

File: tests/Cms/Languages/LanguageRoutesTest.php [hunk 1]
```diff
@@ -0,0 +1,56 @@
...
+                    ]
+                ]
+            ]
+        ]);
+
+        $app->call('notes');
+        $this->assertSame($app->language()->code(), 'en');
+
+        $app->call('de/notes');
+        $this->assertSame($app->language()->code(), 'de');
```
afbora, line 54: FYI: `$app->language()->code()` was return `en` language always on `defaultLanguage()` fallback. Now works great!

bastianallgeier (approved):

PR #2708, opened 2020-07-14
Add new content() method to Form class

### LINKED ISSUE #1602: `translate: false` in a child of a structure field is saved as `null` on secondary language
**Describe the bug**
I was eager to test this feature of Kirby 3.1.0:
>translate: false is no longer ignored in structure fields

But it seems there's a problem with it. When you modify a structure entry that has a `translate: false` field in a secondary language, that field's value is not saved.

**To Reproduce**
1. Create a page with the following template:
```yaml
title: Test Page
preset: page
fields:
  title:
    type: text
  mydata:
    type: structure
    fields:
      foo:
        type: text
      bar:
        type: text
        translate: false
```
2. In the panel, create a new page using the new template.
3. In the default language, add a new entry in the structure.
4. Save your changes.
5. Switch to a secondary language.
6. Open the newly added structure field entry. The `translate: false` field should correctly be disabled and have the default language value, while the other field is editable.
7. Change the editable field.
8. Save your changes.
9. The `translate: false` field of the entry is suddenly empty and if you open the secondary language's txt file, you can see that the field's value is `null`. The other field (the translatable one) is alright, though. However, since the field has `translate: false` and is disabled, you can no longer give it a value via the panel.

**Expected behavior**
When you click save in step 8, it should either:
1. Copy the default language's untranslatable fields instead of setting `null` for their values.
2. Save only the translatable fields and merge the default language's data with the translated data.

**Kirby Version**
3.1.0

 - OS: Windows
 - Browser Chrome
 - Version 73.0.3683.75 (Official Build) (64-bit)

### LINKED ISSUE #2702: Structure Field saves horizontal line in content file
**Describe the bug**  
Using the horizontal line to separate fields from each other in a structure field results in `line: null` being saved in the content file.

**To Reproduce**  
1. Use a horizontal line in a structure field
```yaml
lineTest:
  type: structure
  fields:
    title:
      type: text
    line-a:
      type: line
    description:
      type: text
```
2. add a new entry
3. check the content file
```
Linetest:

- 
  title: test
  description: test
  line-a: null
```
4. Also the structure field preview treats the line as a field

**Screenshots**
lineok

line

**Expected behavior**  
The line not present in the structure field

**Kirby Version**  
3.4.0

afbora:
## Describe the PR

Added new `$form->content()` method to `Form` class. I preferred to add shortcuts instead of always sending double parameters like `$form->data(false, false)`, of course the final decision is yours.

Also fixes about showing unsaveable fields in panel ui.

## Related issues

- Fixes #2702 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/fields/structure.php
+++ b/config/fields/structure.php
@@ -112,7 +112,7 @@ return [
 
                     // Skip hidden and unsaveable fields
                     // They should never be included as column
-                    if ($field['type'] === 'hidden' || $field['save'] === false) {
+                    if ($field['type'] === 'hidden' || $field['saveable'] === false) {
                         continue;
                     }
 
@@ -129,7 +129,7 @@ return [
 
                     $field = $this->fields[$columnName] ?? null;
 
-                    if (empty($field) === true || $field['save'] === false) {
+                    if (empty($field) === true || $field['saveable'] === false) {
                         continue;
                     }
 
--- a/config/fields/structure.php
+++ b/config/fields/structure.php
@@ -110,9 +110,9 @@ return [
             if (empty($this->columns)) {
                 foreach ($this->fields as $field) {
 
-                    // Skip hidden fields.
+                    // Skip hidden and unsaveable fields
                     // They should never be included as column
-                    if ($field['type'] === 'hidden') {
+                    if ($field['type'] === 'hidden' || $field['save'] === false) {
                         continue;
                     }
 
@@ -129,7 +129,7 @@ return [
 
                     $field = $this->fields[$columnName] ?? null;
 
-                    if (empty($field) === true) {
+                    if (empty($field) === true || $field['save'] === false) {
                         continue;
                     }
 
@@ -181,7 +181,7 @@ return [
         $data = [];
 
         foreach ($value as $row) {
-            $data[] = $this->form($row)->data();
+            $data[] = $this->form($row)->content();
         }
 
         return $data;
--- a/src/Cms/ModelWithContent.php
+++ b/src/Cms/ModelWithContent.php
@@ -702,7 +702,7 @@ abstract class ModelWithContent extends Model
             }
         }
 
-        $arguments = [static::CLASS_ALIAS => $this, 'values' => $form->content(), 'strings' => $form->strings(), 'languageCode' => $languageCode];
+        $arguments = [static::CLASS_ALIAS => $this, 'values' => $form->data(), 'strings' => $form->strings(), 'languageCode' => $languageCode];
         return $this->commit('update', $arguments, function ($model, $values, $strings, $languageCode) {
             // save updated values
             $model = $model->save($strings, $languageCode, true);
--- a/src/Cms/ModelWithContent.php
+++ b/src/Cms/ModelWithContent.php
@@ -702,7 +702,7 @@ abstract class ModelWithContent extends Model
             }
         }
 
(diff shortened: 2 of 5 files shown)


config/fields/structure.php
```diff
@@ -110,9 +110,9 @@
             if (empty($this->columns)) {
                 foreach ($this->fields as $field) {
 
-                    // Skip hidden fields.
+                    // Skip hidden and unsaveable fields
                     // They should never be included as column
-                    if ($field['type'] === 'hidden') {
+                    if ($field['type'] === 'hidden' || $field['save'] === false) {
```
afbora: Hides unsaveable fields in the panel ui

config/fields/structure.php
```diff
@@ -129,7 +129,7 @@
 
                     $field = $this->fields[$columnName] ?? null;
 
-                    if (empty($field) === true) {
+                    if (empty($field) === true || $field['save'] === false) {
```
afbora: Hides unsaveable fields in the panel ui

src/Form/Field.php
```diff
@@ -284,8 +284,9 @@ public function toArray(): array
 
         unset($array['model']);
 
-        $array['invalid']   = $this->isInvalid();
         $array['errors']    = $this->errors();
+        $array['invalid']   = $this->isInvalid();
+        $array['save']      = $this->save();
```
afbora: To access saveable fields from fields config

bastianallgeier: I would prefer "saveable" as array key here.

src/Cms/ModelWithContent.php
```diff
@@ -702,7 +702,7 @@ public function update(array $input = null, string $languageCode = null, bool $v
             }
         }
 
-        $arguments = [static::CLASS_ALIAS => $this, 'values' => $form->data(), 'strings' => $form->strings(), 'languageCode' => $languageCode];
+        $arguments = [static::CLASS_ALIAS => $this, 'values' => $form->content(), 'strings' => $form->strings(), 'languageCode' => $languageCode];
```
bastianallgeier: By using `$form->content()` here, this becomes a breaking change, right? What was the idea behind using it here? I'm not entirely sure I can follow.

afbora: I don't think it's a breaking change, actually `\Kirby\Data\Txt` handler is already deleting `null` values (other handlers don't). So we send to handler cleared data from `null` values.

If we think there will be a problem, `$form->content()` method will be sufficient to use it only in the `structure` field to clear null values from structure fields data.

bastianallgeier: The $form->content() method does no longer contain the null values though, right? It's the one that returns the array with deleted keys for fields that should not be stored. That's why I think it would be a breaking change unless I missed something.

afbora: If I understand it correctly;

If a field is deleted from blueprint, its data continues to be kept in the txt file. This PR only concerns fields that cannot be saved as `save: false`.

As I said, it will be sufficient to use this method only in the structure field if you want.

bastianallgeier: I think I would be better to focus on the structure field for now in this PR.

afbora:
It's ready 🚀

bastianallgeier (approved):

Issue #2697, opened 2020-07-09
Avoid dot notation for setting configuration


yandodov:
This issue builds upon getkirby/kirby#2343 and [a discussion we had in Slack](https://getkirby.slack.com/archives/C2UTAQAN4/p1594207467116700). [The docs](https://getkirby.com/docs/reference/plugins/extensions/options#registering-options) were updated to reflect the new changes to dot notation in the new [3.4.0 release](https://github.com/getkirby/kirby/releases/tag/3.4.0).

As I said in the mentioned issue, I think dot notation is absolutely useful, but only for _getting_ values, not for setting. I still think that's the case and here are my arguments.

# Problems

## 1. Possible misuse

The updated docs say (emphasis mine):

>**Always** register your options with the dot notation (like in the example above), **never as nested arrays** (meaning: each separate option should be defined on the top-level of the options array).

And in Slack, @lukasbestle gave the example of a plugin option with key `image.jpg`. Since users are encouraged to **always** use dot notation, you could end up with this:

```php
'options' => [
    'foo.image.jpg' => 'bar'
]
```

...which will be expanded to:

```php
'foo' => [
    'image' => [
        'jpg' => 'bar'
    ]
]
```

...and not the expected:

```php
'foo' => [
    'image.jpg' => 'baz'
]
```

## 2. Mental overhead

We can specify options as nested arrays in `config.php` and in the `App` constructor, but not in plugins. What about the `ready` callback, though? In a `ready` callback in `config.php` - with nested arrays and in `ready` callback of a plugin - with dot notation? Or we can use both? Or always with dot notation? Then, quoting the docs, there's this:

>Please also note that it's not possible to define both a plugin option without a dot and a plugin option that starts with the same name but then continues with a dot.

There are all those things you have to keep in mind for no apparent reason. It goes against the simplicity and intuitiveness of Kirby.

## 3. It's not a pattern

Where have you seen this pattern in other software? Have you ever seen something like this:

```js
let thing = new Thing({
    foo: 'bar',
    'my.nested.key': {
        'one': 1,
        'two': 2
    }
})
```

...in the JavaScript world, for example? Even the PHP one, for that matter. I haven't seen it and my guess for the reason why is ambiguity. You can do the same things in so many ways:

```php
'my' => [
    'nested' => [
        'key' => 42
    ]
]
```

```php
'my' => [
    'nested.key' => 42
]
```

```php
'my.nested' => [
    'key' => 42
]
```

```php
[...]

PR #2709, opened 2020-07-14
Add new "slugs.maxlength" option

### LINKED ISSUE #1821: Page title doesn't take OS' filepath length limit into account
You can create a page title with a very long name and this will result in an `mkdir(): No such file or directory` error. Obviously the maximum filepath length limit isn't taken into account while checking the title of a page.

Not sure how relevant this issue is - maybe it's relevant for API requests or virtual pages. I just spotted it by chance. Watch out for the length limits of Windows 10 and Linux, as they are different as you most likely know.

afbora:
## Describe the PR

Added new `slugs.maxlength` ~~content.slug.maxlength~~ option as @lukasbestle idea.
We can get your better ideas about option name and error message if you have.

## Related issues

- Fixes #1821 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/PageRules.php
+++ b/src/Cms/PageRules.php
@@ -39,7 +39,7 @@ class PageRules
             ]);
         }
 
-        self::slugMaxlength($page, $slug);
+        self::validateSlugLength($slug);
 
         $siblings = $page->parentModel()->children();
         $drafts   = $page->parentModel()->drafts();
@@ -213,7 +213,7 @@ class PageRules
             ]);
         }
 
-        self::slugMaxlength($page);
+        self::validateSlugLength($page->slug());
 
         if ($page->exists() === true) {
             throw new DuplicateException([
@@ -297,20 +297,26 @@ class PageRules
             ]);
         }
 
-        self::slugMaxlength($page);
-
         return true;
     }
 
-    protected static function slugMaxlength(Page $page, string $slug = null): void
+    /**
+     * Ensures that the slug doesn't exceed the maximum length to make
+     * sure that the directory name will be accepted by the filesystem
+     *
+     * @param string $slug New slug to check
+     * @return void
+     *
+     * @throws \Kirby\Exception\InvalidArgumentException If the slug is too long
+     */
+    protected static function validateSlugLength(string $slug): void
     {
-        if ($slugsMaxlength = option('slugs.maxlength')) {
-            $slug      = $slug ?? $page->slug();
+        if ($slugsMaxlength = option('slugs.maxlength', 255)) {
             $maxlength = (int)$slugsMaxlength;
 
             if (Str::length($slug) > $maxlength) {
                 throw new InvalidArgumentException([
-                    'key' => 'page.slug.maxlength',
+                    'key'  => 'page.slug.maxlength',
                     'data' => [
                         'length' => $maxlength
                     ]
--- a/src/Cms/PageRules.php
+++ b/src/Cms/PageRules.php
@@ -39,6 +39,8 @@ class PageRules
             ]);
         }
 
+        self::slugMaxlength($page, $slug);
+
         $siblings = $page->parentModel()->children();
         $drafts   = $page->parentModel()->drafts();
 
@@ -211,6 +213,8 @@ class PageRules
             ]);
         }
 
+        self::slugMaxlength($page);
+
         if ($page->exists() === true) {
             throw new DuplicateException([
                 'key'  => 'page.draft.duplicate',
@@ -293,6 +297,25 @@ class PageRules
             ]);
         }
 
+        self::slugMaxlength($page);
+
         return true;
+    }
+
+    protected static function slugMaxlength(Page $page, string $slug = null): void
+    {
+        if ($slugMaxlength = option('content.slug.maxlength')) {
+            $slug      = $slug ?? $page->slug();
+            $maxlength = (int)$slugMaxlength;
+
+            if (Str::length($slug) > $maxlength) {
+                throw new InvalidArgumentException([
+                    'key' => 'page.slug.maxlength',
+                    'data' => [
+                        'length' => $maxlength
+                    ]
+                ]);
+            }
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


bastianallgeier:
This is great. I just wonder if we could somehow find a cleaner option name for it. We already have the slugs option. Ideally we would end up with something like: 

```php
return [
  'slugs' => [
    'language' => 'de',
    'maxlength' => 120
  ]
];
```

https://getkirby.com/docs/reference/system/options/slugs

What do you think?

afbora:
Actually, the first thing that came to my mind was the `slugs` option, but I did not want to change the current process of `slugs` option. I like the `slugs.maxlength` option 👍

bastianallgeier:
@afbora we would need to make sure though that `slugs => "en"` could still work.

lukasbestle:
I agree with @bastianallgeier.

afbora:
Great, done! 🎉 

I just wonder that following `slugs` option related with this? Please help me about that because i have no idea about system model:

https://github.com/getkirby/kirby/blob/master/config/api/models/System.php#L48-L50

lukasbestle:
> I just wonder that following `slugs` option related with this? Please help me about that because i have no idea about system model:

No, that's just a value for the API, it's unrelated to the option.

src/Cms/PageRules.php
```diff
@@ -293,6 +297,25 @@ public static function update(Page $page, array $content = []): bool
             ]);
         }
 
+        self::slugMaxlength($page);
+
         return true;
     }
+
+    protected static function slugMaxlength(Page $page, string $slug = null): void
```
lukasbestle: ## 1. Naming

I think the name of the method should be `validateSlugLength` to make it clearer what it does.

## 2. Method signature

The second argument should be explicitly nullable:

```suggestion
    protected static function validateSlugLength(Page $page, ?string $slug = null): void
```

As the method doesn't always need the page (only if the slug is not given), the cleanest way would be to have only one argument, but type hinting both a page and a string is not possible until PHP 8.

Alternative: Only ever accept a string with the slug and make the other methods pass that explicitly:

```suggestion
    protected static function validateSlugLength(string $slug): void
```

I prefer the second alternative, but that's just personal preference.

## 3. Documentation

Could you also add a method docblock above it? All methods will have one at some point, so it's best to add them to new methods already while we still now what the method was supposed to do.

```php
/**
 * Ensures that the slug doesn't exceed the maximum length to make
 * sure that the directory name will be accepted by the filesystem
 *
 * @param string $slug New slug to check
 * @return void
 *
 * @throws \Kirby\Exception\InvalidArgumentException If the slug is too long 
 */
```

src/Cms/PageRules.php
```diff
@@ -293,6 +297,25 @@ public static function update(Page $page, array $content = []): bool
             ]);
         }
 
+        self::slugMaxlength($page);
+
```
lukasbestle: If I understand the logic correctly, updating the content won't change the slug, so this method wouldn't need to validate the slug length. Or have I missed something?

src/Cms/PageRules.php
```diff
@@ -293,6 +297,25 @@ public static function update(Page $page, array $content = []): bool
...
         return true;
     }
+
+    protected static function slugMaxlength(Page $page, string $slug = null): void
+    {
+        if ($slugsMaxlength = option('slugs.maxlength')) {
```
lukasbestle: I think there should be a default value, e.g. 150 characters. This default value should be used if the configured option is not set or if it's not already an integer.

So the `if` is not required, the check should always be made, either with the configured value or with the default.

tests/Cms/App/AppTest.php
```diff
@@ -709,4 +709,45 @@ public function testVersionHash()
...
+
+        $this->assertSame('AE', Str::$language['Ä']);
+
+        // string option with dot notation
+        new App([
+            'roots' => [
+                'index' => '/dev/null'
+            ],
+            'options' => [
+                'slugs.language' => 'de'
```
lukasbestle: This second test should use a different language code to make sure that the first test isn't the reason why this passes.

src/Cms/PageRules.php
afbora: You right! I thought slug could be updated with the update() method, but this is not possible yes.

afbora:
@lukasbestle Thank you for really helpful feedbacks. Implemented with default value is `255`

````php
$this->option('slugs', $this->option('slugs.language'))
````

I have to use above format to getting `slugs.language` option.
I know it's a bit complicated but you know that the user can make this definition in 3 different ways and we have a lack about dot notation options.

````php
return [
  'slugs' => 'de'
];

return [
  'slugs.language' => 'de'
];

return [
  'slugs' => [
    'language' => 'de'
  ]
];
````

afbora:
@lukasbestle I used the `A::nest()` method in `App` class with 2 new places to access the different syntax settings defined at the init. Now what you think about that?

lukasbestle:
I think we should wait for #2697. It’s likely that the way options are defined will be changed again.

I’m working on that issue and hope that I will have a solution this weekend.

distantnative:
Could we add a few comments, especially to 
src/Cms/App.php 
src/Cms/AppTranslations.php 

Those parts never have been easy to read and understand. And these changes make it even more complex and difficult. Some comments would help a lot.
And also making some lines more explicit even if it makes the code overall a bit longer- but also a lot more easily to read (e.g. src/Cms/AppTranslations.php line 69 I think).

What do you think about this?

lukasbestle:
@distantnative I have discussed with @afbora that we will add comments when reviewing this PR. I will take a look at the PR in detail in the next weeks.

afbora:
@distantnative As you said, i added some annotations. Also phpDoc blocks have been improved in the classes i'm working on. @lukasbestle will go over 👍

afbora:
Updated doc blocks according to @bastianallgeier says. Unit tests broken due to unneccasery space that related another PR and i've fixed on other PR.

bastianallgeier (approved):
LGTM!

PR #2721, opened 2020-07-19
Gap field

distantnative:
```yaml
fields:
  name:
    type: text
    width: 1/2
  gap:
    type: gap
    width: 1/2
  text:
    type: textarea
```

```yaml
fields:
  hasHeadline:
    type: radio
    options:
      - yes
      - no
  headline:
    type: text
    width: 1/2
    when:
      hasHeadline: yes
  gap:
    type: gap
    width: 1/2
    when:
      hasHeadline: no
  text:
    type: textarea
```


afbora (approved):
Loved ❤️

afbora:
Related idea https://github.com/getkirby/ideas/issues/332

lukasbestle (approved):
That is an incredible simple implementation for such a useful field. 👍

distantnative:
Pssssssst... it was a highly complex endaveour otherwise there would be no explanation why it only is going to be implemented now 😬

distantnative:
I think the main question is whether we add it to 3.4.1 or 3.5.0 - tempted to say 3.4.1 even though it is a feature, just to not overload 3.5.0 in the end again.

lukasbestle:
I agree! It‘s not a breaking-change so we can add it to 3.4.1 to make it available as soon as possible.

bastianallgeier (approved):

Issue #1514, opened 2019-02-28
[UrlField] debounce validation


nilshoerrmann:
**Describe the bug**
In a URL field, when adding content the field is validated `onInput` causing an error message until the user finished typing a valid link.

**To Reproduce**
1. Add a URL field to a page.
2. Start typing.

**Expected behavior**
Validation should happen on `blur` or `save`. Or – if a more immediate feedback is desired – it should be debounced giving the user time to finish writing.

**Kirby Version / Browser**
3.0.3-rc.1 using Firefox 65.0.1 on macOS Mojave

distantnative:
Moved to https://kirby.nolt.io/127

Issue #2270, opened 2019-11-02
[Panel] Conflicting JS and PHP validation on URL field


afbora:
**Describe the bug**  
We added localhost and ip address support starting with 127 to url validation. 
Since the JS and PHP validation part are different in the url field, when the localhost address is entered, the JS validation is shown as an error while it passes PHP validation as reported by @GiantCrocodile. So @bastianallgeier's suggestion is to remove instant JS validation.

**To Reproduce**  
1. Go to panel
2. Edit any page that have URL field
3. Enter localhost address: `http://localhost/kirby/panel/pages/blog+cdwcwd`
4. See red error highlight (JS Validation)
5. Saving the page is success (PHP Validation)

**Expected behavior**  
JS and PHP validation should use the same rules or JS validation should be removed.

**Screenshots**  
![grafik](https://user-images.githubusercontent.com/6271057/67640353-792cc380-f8fa-11e9-8252-108ef2b81199.png)

**Related issue:** #2130

**Kirby Version**  
3.3.0

GiantCrocodile:
Maybe the js frontend validation can be triggered by the PHP validation. If PHP validation fails, add the red error highlight. Of course this is no validation prior sending it to the backend but it still helps the user by highlighting the field with invalid data.

Thank you for creating the issue @afbora!

distantnative:
I would suggest to close this issue - we won't fix it in the current JS and PHP setup. But I think eventually it will get solved indirectly when we switch to do all validation only via the PHP backend.

afbora:
@distantnative Input still highlighted as red. I guess this issue still continue. Because this is related JS validation. Form already saveable before 3.6. But red highlight confusing.

distantnative:
@afbora as I wrote, this isn't fixed but we won't solve the JS validation as we plan to switch away from it anyway. That's why Bastian closed it.

PR #2726, opened 2020-07-19
Panel: remove inline validation

distantnative:


afbora:
Related issues to close #1514 #2270

bastianallgeier:
I wonder if we should release this in a 3.4.x release or if it is something for 3.5. What do you think?

distantnative:
Yea, tricky. Maybe it is a bigger change, cause a feature is leaving - could it be a breaking one where someone relies on inline validation? Probably no cause the backend validation would still prevent values from getting saved. On the other hand, it also fixes some bugs. Maybe a topic to test the waters with the community?

bastianallgeier:
It can't really break plugins as far as I know. We haven't created a way to use the existing validators in plugins so far. That's why field plugins have no way of reusing our inline validation code. But without communicating it clearly, it could feel like we simply broke validation entirely with this change.

distantnative:
That's true and attention will be much better for the 3.5.0 release notes than 3.4.2

PR #2734, opened 2020-07-21
Spacing specificity

lukaskleinschmidt:
## Describe the PR
I just skimmed through some files and I think the way the spacing helpers (margin and padding) are implemented could be improved. Right now something like the following example won't work as expected (at least how I would expect it to work 😬). You will always end up with the spacing of `4` on all sides because higher spacings will always have precedence over the smaller ones in the current implementation. I would expect the  `padding-bottom` to win over the more general `padding`.
```html
<div class="p-4 pb-0"></div>
```
In addition here is a quick [codepen](https://codepen.io/lukaskleinschmidt/pen/yLeGryQ) to illustrate the PR.

## Ready?

- [ ] Passing all unit tests

### DIFF
--- a/panel/src/ui/css/utilities/margin.scss
+++ b/panel/src/ui/css/utilities/margin.scss
@@ -1,32 +1,39 @@
 /* Margins */
-.mx-auto,
-.mr-auto {
-  margin-right: auto;
+@each $index, $size in $margin {
+  .m-#{$index} {
+    margin: $size;
+  }
 }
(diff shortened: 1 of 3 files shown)


distantnative:
I like it and can understand the argument.
@bastianallgeier since it’s your original code

bastianallgeier:
I in fact already ran into this issues myself. Thanks for the fix!

PR #2735, opened 2020-07-22
Add support for custom file drag texts

pwaldhauer:
## Describe the PR

This pull request adds support for defining a custom function that returns the text that should be inserted into a textarea when a file is dragged onto it (or inserted via the toolbar). 

My motivation for this change was to support my custom kirby tags:

- .heic images should be inserted as (image: xxx)
- movies should be inserted as (video: xxx)

The function in the config.php could like like this:

```
  'fileDragText' => function(\Kirby\Cms\File $file, $url) {
        if($file->extension() === 'heic') {
            return sprintf('(image: %s)', $url);
        }

        if($file->type() === 'video') {
            return sprintf('(video: %s loop:true)', $url);
        }

        return null;
    },
```

## Related issues

None

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -255,7 +255,7 @@ class File extends ModelWithContent
                 // no break
             default:
                 $fileDragTextFn = option('fileDragText');
-                if(!empty($fileDragTextFn) && is_callable($fileDragTextFn) && !empty($text = $fileDragTextFn($this, $url))) {
+                if (!empty($fileDragTextFn) && is_callable($fileDragTextFn) && !empty($text = $fileDragTextFn($this, $url))) {
                     return $text;
                 }
 
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -254,6 +254,11 @@ class File extends ModelWithContent
                 }
                 // no break
             default:
+                $fileDragTextFn = option('fileDragText');
+                if(!empty($fileDragTextFn) && is_callable($fileDragTextFn) && !empty($text = $fileDragTextFn($this))) {
+                    return $text;
+                }
+                
                 if ($this->type() === 'image') {
                     return '(image: ' . $url . ')';
                 } else {
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -245,13 +245,14 @@ class File extends ModelWithContent
 
         $url = $absolute ? $this->id() : $this->filename();
 
+        $dragTextCallback = option('panel.' . $type . '.fileDragText');
+
+        if (empty($dragTextCallback) === false && is_a($dragTextCallback, 'Closure') === true && ($dragText = $dragTextCallback($this, $url)) !== null) {
+            return $dragText;
+        }
+
         switch ($type) {
             case 'markdown':
-                $dragTextFn = option('panel.markdown.fileDragText');
-                if (!empty($dragTextFn) && is_callable($dragTextFn) && !empty($text = $dragTextFn($this, $url))) {
-                    return $text;
-                }
-
                 if ($this->type() === 'image') {
                     return '![' . $this->alt() . '](' . $url . ')';
                 } else {
@@ -259,11 +260,6 @@ class File extends ModelWithContent
                 }
                 // no break
             default:
-                $dragTextFn = option('panel.kirbytext.fileDragText');
-                if (!empty($dragTextFn) && is_callable($dragTextFn) && !empty($text = $dragTextFn($this, $url))) {
-                    return $text;
-                }
-
                 if ($this->type() === 'image') {
                     return '(image: ' . $url . ')';
                 } else {
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -247,6 +247,11 @@ class File extends ModelWithContent
 
         switch ($type) {
             case 'markdown':
+                $dragTextFn = option('panel.markdown.fileDragText');
+                if (!empty($dragTextFn) && is_callable($dragTextFn) && !empty($text = $dragTextFn($this, $url))) {
+                    return $text;
+                }
+
                 if ($this->type() === 'image') {
                     return '![' . $this->alt() . '](' . $url . ')';
                 } else {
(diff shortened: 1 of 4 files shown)


afbora:
As @distantnative  mentioned, i think it would be more stable to be done over class `F` 🤔 

1. Putting a new option that extends file types from config. Thus, it can be used in different places.
2. `dragText()` method returns by looking at `F::$types` on auto mode.

If our goal is a custom return on `dragText`, it would be correct to have a callback option as you did.

pwaldhauer:
> As @distantnative mentioned, i think it would be more stable to be done over class `F` 🤔
> 
> 1. Putting a new option that extends file types from config. Thus, it can be used in different places.
> 2. `dragText()` method returns by looking at `F::$types` on auto mode.

But this would only fix the first problem, I guess?

I don't see how I could build something like my second `if` (the video part) using the F::$types.

afbora:
New option that extending file types:

````php
return [
    'file' => [
        'types' => [
            'image' => [
                'heic'
            ],
            'image' => [
                'm4p'
            ]
        ]
    ]
];
````

I'm not sure but why don't we put the video type in the condition?

````php
if ($this->type() === 'image') {
    return '(image: ' . $url . ')';
} elseif ($this->type() === 'video') {
    return '(video: ' . $url . ')';
} else {
    return '(file: ' . $url . ')';
}
````

**Comment by distantnative** [comment]
Thanks Philipp for the PR.

I think there are two separate discussion:
1. Should there be a way to customize the drag text via an option (as included in this PR)? Maybe instead of an option, one could make it a core component. But I have no fixed opinion here yet.
2. Are there other, better, additional ways to solve Philipp's concrete problems.

The latter brings up a few idea:
```php
F::$types['images][] = 'heic';
```

And I totally agreed that the `(video: )` tag should probably become a core default. So we could attack that for sure.

But I don't want the specific solutions to drown Philipp's initial idea of making this extensible, which is a valid suggestion I think. @bastianallgeier, thoughts?

pwaldhauer:
While I understand that there are easier solutions to fix exactly these problems (heic, video), I used this approach because I can think of some use cases where those easy solutions don't apply and my more complex way would be needed. 

For example something like "if the video is < 3 seconds, automatically include "loop: true" in the kirby tag."

So I would really like having the option to do something like that.

afbora:
Hmm, i just wonder that [overriding default kirbytags](https://getkirby.com/docs/reference/plugins/extensions/kirbytags#overriding-default-kirbytags) doesn't meet your needs? 🤔

pwaldhauer:
I already extended the image tag to support heic images. I dont really see how to implement my other features by creating custom tags. 

My goal is to make the editing process more comfortable by prefilling the kirby tags with the right values in the editor. 

(I could overwrite the file tag, but I think its less expressive than having multiple tags for different files?)

afbora:
Now i also realized that we cannot show local videos with `video` tag like `(video: sample.mp4)`. Only video providers supports like youtube, vimeo 😞

afbora:
Related video tag idea https://github.com/getkirby/ideas/issues/166

bastianallgeier:
Here are my two cents :) 

1.) I like the idea to offer this as a simple option. It's super effective and I love that you added a test to your PR <3 
2.) I'm not entirely sure about the naming for the option. If we introduce this option we probably should also add another one for the page drag text and maybe even user drag text. We already have the panel namespace and I think that it should be there. There's also panel.kirbytext and maybe it should even be a child of that: 

```php
return [
    'panel' => [
        'kirbytext' => [
            'fileDragText' => function () {

            },
            'pageDragText' => function () {

            },
            'userDragText' => function () {

            }
        ]
    ]
];
```

This would mean that we could have the same for markdown at some point: 

```php
return [
    'panel' => [
        'markdown' => [
            'fileDragText' => function () {

            },
            'pageDragText' => function () {

            },
            'userDragText' => function () {

            }
        ]
    ]
];
```

3.) We should add heic to the default list of images
4.) There should be a way to add your own file type definitions and I really like @afbora's approach

pwaldhauer:
Hej, thanks for your input!

1) Thanks!
2) Makes sense. I could — if it increases the chance of this PR getting merged — implement it that way and add the other drag texts, too. 
3) I think the browser support for heic images is not really usable right now. I currently modified my image tag to automatically convert heics to jpegs using heic-convert from https://github.com/strukturag/libheif
4) I too think its a good idea. Maybe @afbora has the time to create a PR?

afbora:
@pwaldhauer Do you have time to take care of @bastianallgeier ideas? Otherwise we can take care of it 👍

pwaldhauer:
@afbora I'll push my changes that bastian recommended tomorrow! :)

pwaldhauer:
- I renamed the option to the format @bastianallgeier suggested
- I also added support for the markdown options
- I added tests
- I added support for the pageDragText

Only thing missing is the User dragText? I did not find this in the source code, can someone give me a pointer?

afbora (commented):
I would prefer using before `switch case` to avoid repetition. Also a short description would be nice.

````php
// check dragText option for markdown and kirbytext
$dragTextCallback = option('panel.' . $type . '.fileDragText');

if (empty($dragTextCallback) === false && is_a($dragTextCallback, 'Closure') === true) {
    return $dragTextCallback($this, $url);
}

switch ($type) {
...
}
````

---

IMO Also, i'm not sure but if the user has defined a setting for `dragText` option, I don't think we need to check if it is empty. If callback returns empty, the user may want it to return empty, or if it is an error, it is necessary for the user to realize that this is an error or something wrong.

Can be use in option to default return:

````php
return $file->dragText($file, $url);
````

**Comment by pwaldhauer** [comment]
@afbora Thanks, I optimized the code as you suggested. I just blindly copied the code in the commit before, sorry, that was not very efficient.

Still, I think it's helpful to be able to return `null` in the closure to have the possibility to run the default behaviour of 
the dragText function, e.g. if you only want to adjust the drag text in a specific case and fall back to the default, if the case is not met.

```php
'fileDragText' => function (\Kirby\Cms\File $file, string $url) {
                            if ($file->extension() === 'heic') {
                                return sprintf('(image: %s)', $url);
                            }
                    
                            return null;
                        },
```

I changed it to an explicit check on `null`, tho.

afbora:
I'm still not sure that returning `null` is very effective. Team members will give a better idea while review the PR. Thanks for commits, looks good 👍 

Btw i think @bastianallgeier said the `userDragText` feature can be used when it comes to the [users section feature](https://github.com/getkirby/ideas/issues/422). So I don't think it's necessary to implement right now.

bastianallgeier:
I finally merged this after a few small changes. I had to do it manually though. You can find it on the develop branch. 🎉

pwaldhauer:
@bastianallgeier nice, thanks! Any idea when this will ship? I could provide a fitting cookbook article to document this rather niche feature.

afbora:
@pwaldhauer The 3.4.3 RC version is scheduled to be released today as Tuesday, September 8th.

PR #2748, opened 2020-07-28
Styling: new SCSS variables

distantnative:
As minimal as it can be without breaking things:

- Took the new SCSS variables from `storybook`
- Replaced old SCSS variables in components, trying to match look 1:1
=> this should only introduce minimal visible changes (e.g. `$color-dark` being replaced by `$color-gray-900`)

Actual styling changes should be happening later when refactoring each component

### DIFF
--- a/panel/src/main.scss
+++ b/panel/src/main.scss
@@ -6,11 +6,12 @@
  * file is imported. Use only for mixins, variables and functions.
  */
 
-$layers: (
-  loader: 900,
-  notification: 800,
-  dropdown: 700,
-  dialog: 600,
+ $z-index: (
+  loader: 1000,
+  notification: 900,
+  dropdown: 800,
+  dialog: 700,
+  drawer: 600,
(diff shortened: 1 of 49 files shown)


panel/src/main.scss
```diff
@@ -19,119 +20,149 @@ $layers: (
...
-$list-item-info-font-size: $font-size-tiny;
+$font-normal: 400;
+$font-bold: 600;
 
 /**
- * Misc
+ * Border radius
  */
-$border-radius: 1px;
+$rounded-xs: 1px;
```
yoanmalie, line 137: Why this one is in `px` while other are `rem` ?

distantnative: @yoanmalie Cause 1px should stay 1px no matter what the root font size is. And it's also much easier o read than any weird `rem` fraction in this case.

bastianallgeier (approved):

Issue #2743, opened 2020-07-27
Unexpected behavior creating textarea with Html::tag 


LeBenLeBen:
**Describe the bug**  

Since version 3.4, when you create a `textarea` element with the Html::tag method like this:

```php
Html::tag('textarea', null);
```

The output is:

```html
<textarea>
```

and therefor not valid HTML. Everything following this tag ends as the textarea content.

**Expected behavior**  

A textarea without content should generate a closing tag as well:

```html
<textarea></textarea>
```

**Kirby Version**  
3.4.0

**Workaround**

Pass an empty string as the second argument of the `Html::tag` method.

lukasbestle:
The previous behavior of the `Html` class was pretty inconsistent and buggy, so we have refactored it in 3.4.0.

If you set the second argument to `null`, you will [now get a self-closing element](https://getkirby.com/docs/reference/tools/html/tag). If you want an empty `<textarea>`, use an empty string like so:

```php
Html::tag('textarea', '');
```

bastianallgeier:
@lukasbestle sorry to intervene here, but I missed this change to be honest. Why would passing null create a self-closing element? There's a fixed list of self-closing HTML elements that we can use to do this automatically.

LeBenLeBen:
I agree with @bastianallgeier, I didn't find a way to create just a closing element, so I don't think creating an "invalid" opening-only element should be allowed.

lukasbestle:
Sorry for closing this so early. I should have explained in more detail why this is the case, which I now want to catch up on:

The `Html::tag()` method has been documented to return a self-closing tag for `$content === null` [since Kirby 3.0.0](https://github.com/getkirby/kirby/blob/3.0.0/src/Toolkit/Html.php#L348) (possibly even earlier). Unfortunately the method didn't actually do what the documentation said. Instead it just used the fixed list of void tags (`Html::isVoid()`) to determine whether to use a self-closing tag or not.

The refactored implementation in 3.4.0 now does what the documentation has said from the beginning. To be honest, I just didn't expect while refactoring that people explicitly used `null` as the second argument before Kirby 3.4.0 as the documentation said to use an empty string for an empty non-self-closing tag. Also, `null` was the default value anyway before Kirby 3.4.0, which made me believe even more that explicitly setting `null` is pretty unlikely. The new default value is an empty string, which matches the documentation and prevents a breaking-change both for those who left out the second argument in this case and for those who have used an empty string all along.

[...]

PR #2753, opened 2020-07-30
Html::isVoid(): Allow modifying the void list

lukasbestle:
## Describe the PR

The list of self-closing tags for the `Html::isVoid()` method can now be modified with `Html::$voidList`.

## Related issues

- See https://github.com/getkirby/kirby/issues/2743#issuecomment-665908758.

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Toolkit/Html.php
+++ b/src/Toolkit/Html.php
@@ -36,6 +36,30 @@ class Html extends Xml
     public static $void = '>';
 
     /**
+     * List of HTML tags that are considered to be self-closing
+     *
+     * @var array
+     */
+    public static $voidList = [
+        'area',
+        'base',
+        'br',
+        'col',
+        'command',
+        'embed',
+        'hr',
+        'img',
+        'input',
+        'keygen',
+        'link',
+        'meta',
+        'param',
+        'source',
+        'track',
+        'wbr'
+    ];
+
+    /**
      * Generic HTML tag generator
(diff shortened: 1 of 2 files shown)


afbora:
Didn't you add the `menuitem` tag because it is deprecated?

distantnative:
LGTM – thought ran to my head whether it should be called `Html::$voidTags` but honestly I have no strong feelings about that whatsoever.

lukasbestle:
To be honest I just moved the existing list over. :)

The [HTML spec](https://html.spec.whatwg.org/#void-elements) doesn't list `<menuitem>` as void, but maybe that's because it's deprecated. On MDN it's indeed listed as being a void element.

Have you been able to find a reliable list of all elements that have at some point in the past been void tags? I found dozens of blogs with various different lists, but it would be good to have one that's directly from a W3C spec.

lukasbestle:
> LGTM – thought ran to my head whether it should be called `Html::$voidTags` but honestly I have no strong feelings about that whatsoever.

Me neither, your suggestion sounds good too. @bastianallgeier may decide. :)

afbora:
Frankly, I could not find a reference resource. I think it's better to stay like this :+1:

https://github.com/w3c/html/issues/274

bastianallgeier (approved):

bastianallgeier:
I'd also say to keep it like this. If someone actually needs to include it, it's now easy to do anyway.

PR #2754, opened 2020-07-31
Increase unit test coverage

distantnative:


tests/Toolkit/VTest.php
```diff
@@ -190,10 +190,12 @@ public function testEmail()
         $this->assertTrue(V::email('bastian-v3@getkirby.com'));
         $this->assertTrue(V::email('bastian.allgeier@getkirby.com'));
         $this->assertTrue(V::email('bastian@getkürby.com'));
+        $this->assertTrue(V::email('bastian@getkürby.com'));
```
lukasbestle: To be honest I don't see a difference between this line and the one above. Am I blind or was this a copy-paste mistake?

src/Toolkit/Str.php
```diff
@@ -179,7 +179,7 @@ public static function after(string $string, string $needle, bool $caseInsensiti
         $position = static::position($string, $needle, $caseInsensitive);
 
         if ($position === false) {
-            return false;
+            return '';
```
lukasbestle, line 182: I'm thinking if these methods should rather return `null` in these cases so that code that uses these methods can differentiate between a non-existing needle and a needle at the end/beginning of the string. The return value hint would then be `?string`.

But that would be a breaking-change. So I'm not sure.

src/Toolkit/Str.php
```diff
@@ -748,8 +748,12 @@ public static function rtrim(string $string, string $trim = ' '): string
      *                         string is too long. Ellipsis is the default.
      * @return string The shortened string
      */
-    public static function short(string $string = null, int $length = 0, string $appendix = '…'): ?string
+    public static function short(string $string = null, int $length = 0, string $appendix = '…'): string
```
lukasbestle, line 751: For consistency:

```suggestion
    public static function short(?string $string = null, int $length = 0, string $appendix = '…'): string
```

The same applies to other methods in this class that accept a string that can be null.

tests/Toolkit/StrTest.php
```diff
@@ -8,50 +8,50 @@ class StrTest extends TestCase
...
     {
         $string = 'Hellö Wörld';
 
         // case sensitive
-        $this->assertEquals(' Wörld', Str::after($string, 'ö'));
-        $this->assertEquals(false, Str::after($string, 'Ö'));
-        $this->assertEquals(false, Str::after($string, 'x'));
+        $this->assertSame(' Wörld', Str::after($string, 'ö'));
+        $this->assertSame('', Str::after($string, 'Ö'));
+        $this->assertSame('', Str::after($string, 'x'));
```
lukasbestle, line 24: I wonder why `false` worked before. Was it only because `assertEquals()` wasn't strict enough or did the methods actually return `false`.

lukasbestle (commented):
Awesome!

I'm thinking if we should use the opportunity to refactor the tests as well (with `@covers` annotations where possible and in some situations with a corrected test ordering), but I think this PR is a good first step. We can further improve the tests later. So I only have a few minor comments.

src/Toolkit/Str.php
distantnative: I’m yea, I was discussing this as well. Also in terms whether to actually use `false`. In the end, I came out on the side that I don’t think it’s worth to risk breaking things. But then the code can actually follow what the type hunting will turn it into anyways.

tests/Toolkit/StrTest.php
distantnative: As far as I can tell, assertEqual just was fine with it. The return type hinting turned the Boolean into an empty string.

src/Toolkit/Str.php
lukasbestle: Yes, if we decide that we don't want a breaking-change here, your change to explicitly return an empty string is definitely the way to go.

Pinging @bastianallgeier regarding the question whether we should add support for `null` here.

tests/Toolkit/StrTest.php
lukasbestle: 👌

src/Toolkit/Str.php
bastianallgeier: I like the idea to go for real strings in this step and then think about returning null in a next step that would then follow in 3.5 or 3.6.

src/Toolkit/Str.php
bastianallgeier: I think we should do this in a separate PR for all the places where we allow null. Right now it's totally mixed in multiple classes.

bastianallgeier (approved):

PR #2761, opened 2020-08-07
Redirects user who doesn't have site access to account page

### LINKED ISSUE #2760: Permission access.site causes the panel to crash
**Describe the bug**  
I'm finally getting into user blueprints and at some point, the panel started crashing. After digging into it, the cause appears to be the `site: false` access setting:

```yml
title: Editor
permissions:
  access:
    site: false
```

This causes the whole browser tab to become unresponsive. Perhaps it falls in some infinite loop?

**Kirby Version**  
3.4.0

afbora:
## Describe the PR

As a simple solution, redirects user who doesn't have site access to `account` page. Because it is the page that the user always has access to.

A separate error page can be made as an alternative solution 💡 

## Related issues

- Fixes #2760 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/panel/src/config/auth.js
+++ b/panel/src/config/auth.js
@@ -28,7 +28,7 @@ export default function(to, from, next) {
         message: Vue.i18n.translate("error.access.view")
       });
 
-      return next("/");
+      return next(access.site === false ? "/account" : "/");
     }
 
     next();


bastianallgeier:
I merged this manually

PR #2763, opened 2020-08-08
Pages: Only use findBy('slug') on non-default languages

neildaniels:
## Describe the PR

In a multi-language setup, finding a Page in a Pages collection by slug is unnecessary for the default language. The existing behavior means that nonexistent ids/slugs incur a full loop iteration that will never actually return a result.

Because the `id` for any page _is_ constructed from default-language slugs, the initial lookup with `->get` should be sufficient.

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -1376,7 +1376,7 @@ class Page extends ModelWithContent
 
             $defaultLanguageCode = $this->kirby()->defaultLanguage()->code();
 
-            if ($languageCode != $defaultLanguageCode && $translation = $this->translations()->find($languageCode)) {
+            if ($languageCode !== $defaultLanguageCode && $translation = $this->translations()->find($languageCode)) {
                 return $translation->slug() ?? $this->slug;
             }
         }
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -1374,7 +1374,9 @@ class Page extends ModelWithContent
                 $languageCode = $this->kirby()->languageCode();
             }
 
-            if ($translation = $this->translations()->find($languageCode)) {
+            $defaultLanguageCode = $this->kirby()->defaultLanguage()->code();
+
+            if ($languageCode != $defaultLanguageCode && $translation = $this->translations()->find($languageCode)) {
                 return $translation->slug() ?? $this->slug;
             }
         }
--- a/tests/Cms/Pages/PagesTest.php
+++ b/tests/Cms/Pages/PagesTest.php
@@ -483,6 +483,109 @@ class PagesTest extends TestCase
         $this->assertIsPage($pages->findById('kind'), 'grandma/mother/child');
     }
 
+    public function testFindByIdWithSwappedSlugsTranslated()
+    {
+        $app = new App([
+            'roots' => [
+                'index' => '/dev/null'
+            ],
+            'languages' => [
+                [
+                    'code' => 'en',
+                ],
+                [
+                    'code' => 'de',
+                ],
+            ],
+            'site' => [
+                'children' => [
+                    [
+                        'slug' => 'aaa',
+                        'translations' => [
+                            [
+                                'code' => 'en',
+                            ],
+                            [
+                                'code' => 'de',
+                                'slug' => 'zzz',
+                            ],
+                        ],
+                        'children' => [
+                            [
+                                'slug' => 'bbb',
+                                'translations' => [
+                                    [
+                                        'code' => 'en',
+                                    ],
+                                    [
+                                        'code' => 'de',
+                                        'slug' => 'yyy'
+                                    ],
+                                ],
+                            ],
+                        ],
+                    ],
+                    [
+                        'slug' => 'zzz',
+                        'translations' => [
+                            [
+                                'code' => 'en',
+                            ],
+                            [
+                                'code' => 'de',
+                                'slug' => 'aaa',
+                            ],
+                        ],
+                        'children' => [
+                            [
+                                'slug' => 'yyy',
+                                'translations' => [
+                                    [
+                                        'code' => 'en',
+                                    ],
+                                    [
+                                        'code' => 'de',
+                                        'slug' => 'bbb'
+                                    ],
+                                ],
+                            ],
+                        ],
+                    ],
+                ],
+            ],
+        ]);
+
+        $site = $app->site();
+
+        $this->assertIsPage($site->children()->findById('aaa'), 'aaa');
+        $this->assertIsPage($site->children()->findById('aaa/bbb'), 'aaa/bbb');
+        $this->assertIsPage($site->children()->findById('aaa')->children()->findById('bbb'), 'aaa/bbb');
+        $this->assertIsPage($site->children()->findById('zzz'), 'zzz');
+        $this->assertIsPage($site->children()->findById('zzz/yyy'), 'zzz/yyy');
+        $this->assertIsPage($site->children()->findById('zzz')->children()->findById('yyy'), 'zzz/yyy');
+
+        $pages = new Pages($site->children()->find('aaa', 'aaa/bbb', 'zzz', 'zzz/yyy'));
+        $this->assertIsPage($pages->findById('aaa'), 'aaa');
+        $this->assertIsPage($pages->findById('aaa/bbb'), 'aaa/bbb');
+        $this->assertIsPage($pages->findById('zzz'), 'zzz');
+        $this->assertIsPage($pages->findById('zzz/yyy'), 'zzz/yyy');
+
+//         $app->setCurrentLanguage('de');
+
+//         $this->assertIsPage($site->children()->findById('aaa'), 'zzz');
+//         $this->assertIsPage($site->children()->findById('aaa/bbb'), 'zzz/yyy');
+//         $this->assertIsPage($site->children()->findById('aaa')->children()->findById('bbb'), 'zzz/yyy');
+//         $this->assertIsPage($site->children()->findById('zzz'), 'aaa');
+//         $this->assertIsPage($site->children()->findById('zzz/yyy'), 'aaa/bbb');
+//         $this->assertIsPage($site->children()->findById('zzz')->children()->findById('yyy'), 'aaa/bbb');
+
+//         $pages = new Pages($site->children()->find('aaa', 'aaa/bbb', 'zzz', 'zzz/yyy'));
+//         $this->assertIsPage($pages->findById('aaa'), 'zzz');
+//         $this->assertIsPage($pages->findById('aaa/bbb'), 'zzz/yyy');
+//         $this->assertIsPage($pages->findById('zzz'), 'aaa');
+//         $this->assertIsPage($pages->findById('zzz/yyy'), 'aaa/bbb');
+    }
+
     public function testFindMultiple()
     {
         $pages = Pages::factory([
--- a/tests/Cms/Pages/PagesTest.php
+++ b/tests/Cms/Pages/PagesTest.php
@@ -301,6 +301,8 @@ class PagesTest extends TestCase
         $this->assertIsPage($site->children()->findById('grandma/mother.json'), 'grandma/mother');
         $this->assertIsPage($site->children()->findById('grandma')->children()->findById('mother'), 'grandma/mother');
         $this->assertIsPage($site->children()->findById('grandma')->children()->findById('grandma/mother'), 'grandma/mother');
+//         $this->assertIsPage($site->children()->findById('mother'), 'grandma/mother');
+//         $this->assertIsPage($site->children()->findByUri('mother'), 'grandma/mother');
         $this->assertIsPage($site->children()->findById('grandma/mother/child'), 'grandma/mother/child');
         $this->assertIsPage($site->children()->findById('grandma/mother/child/'), 'grandma/mother/child');
         $this->assertIsPage($site->children()->findByUri('grandma/mother/child'), 'grandma/mother/child');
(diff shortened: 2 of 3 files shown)


lukasbestle:
Thanks for the PR! This would be a nice performance improvement.

Unfortunately your assumption that default language slugs are always from the directory name is not true. You *can* put a `Slug` field into the text file of the default language and it will be used as the slug of the default language.

This leaves us with the question whether this *should* work. I'd say it's completely unnecessary to support the `Slug` field for the default language, so we could also remove support for it so we can incorporate this performance improvement. However this would then be a breaking-change.

PS: Don't be surprised about all the commits listed in the PR. We had to rebase `develop` as our `master` branch got out of sync during the 3.4.1 release. You can fix this by rebasing your PR branch against our `develop` branch.

neildaniels:
@lukasbestle I accidentally mucked up my rebase and it automatically closed this PR. Can you re-open, or should I submit a new PR? I fixed my branch, but the closed PR doesn't pick up those changes. https://github.com/neildaniels/kirby/tree/neildaniels-pages-findby-slug

As to `slug` on default languages: ah, thanks for pointing that out.

neildaniels:
I agree with @lukasbestle that the performance improvement only makes sense if default language slugs aren't "translatable."

I've updated this PR to modify the behavior of `$page->slug()`, effectively removing the ability to use a "translatable" slug for the default language.

src/Cms/Page.php
```diff
@@ -1401,7 +1401,9 @@ public function slug(string $languageCode = null): string
                 $languageCode = $this->kirby()->languageCode();
             }
 
-            if ($translation = $this->translations()->find($languageCode)) {
+            $defaultLanguageCode = $this->kirby()->defaultLanguage()->code();
+
+            if ($languageCode != $defaultLanguageCode && $translation = $this->translations()->find($languageCode)) {
```
afbora:
```suggestion
            if ($languageCode !== $defaultLanguageCode && $translation = $this->translations()->find($languageCode)) {
```

One more unnecessary query, great! Using identical operator will be more stable 👍

distantnative:
> I've updated this PR to modify the behavior of $page->slug(), effectively removing the ability to use a "translatable" slug for the default language.

We need to be careful with this. It is a breaking change. I would agree that setting the slug for the default language via content file isn't a good practice, nevertheless there could be people out there relying on this behavior - as it has been supported for a long time.

afbora:
@distantnative I didn't understand why it was breaking change. If it's the default language, it doesn't look at translations. Because the slug value for a page's default language is the name of the directory already. Am I wrong?

distantnative:
@afbora Lukas did report that even in the default language so far it has actually looked at the `Slug` content field as well. Ig we remove that, it's a breaking change.

I am not saying we shouldn't do it. I agree that it doesn't make sense for the default language. But given how recent breaking changes have bitten us quite a bit, let's be really careful with anything that is a breaking change.

neildaniels:
The performance boost from this change has been **massive** for our site. Anyone on a multi-lang setup that have "dense" top-level pages should get a huge speedup on requests for the default language.

For examples, we have a `/news` section with 4000+ articles in it. Previously, even a request for `/sports` would trigger `/news` to be inventoried for its "final" slug. Repeat for inventorying almost every directory at the same level.

In our case, our server latency went down about 450ms, which is amazing to me.

lukasbestle:
@neildaniels For my understanding: Does this performance improvement affect all languages or "just" the default language? I.e. if the user requests a page in a non-default language, does this change matter? Or only for requests of pages in the default language?

neildaniels:
@lukasbestle As far as I can reason about it: this _only_ affects requests for the default language.

The only time I can imagine non-default language performance improving is when code was **explicitly** asking a `$page` for its default's language slug.

Contrived example:
```php
$page->translation('non-default-lang')->slug() ?: $page->slug(kirby()->defaultLanguage()->code());
```

But, the Kirby core _doesn't_ explicitly ask for a default slug: it _always_ falls back to `$page->slug` (the property, not the function). In other words: the fallback slug is never pulled from the default language's content file.

https://github.com/getkirby/kirby/blob/1f3f0855438ce9e4724b3866921589fcb7ab40ab/src/Cms/Page.php#L1405

lukasbestle:
That's what I suspected. While such a massive improvement for the default language requests is awesome, the question is if it's worth it to introduce a breaking change "just" for default language requests.

Don't get me wrong: Improving the performance of Kirby is a very important goal, but in the end we need to balance this goal with reliability like Nico wrote above.

neildaniels:
Looking through my own site's code, there is a template-level benefit to this for non-default languages: we have a list of language's each page is available in. This is an example where explicitly asking for the default language's slug actually happens:

```php
<?php foreach ($languages as $language): ?>
    <a href="<?= $page->url($language->code()) ?>">Language Name</a>
<?php endforeach ?>
```

Yeah, I understand the balance of trying to avoid breaking changes. I'm probably stating the obvious, but "custom" default-language slugs certainly aren't well-supported as of today. They happen to work in some places, but in some places they're literally programmer errors.

PageActions's `changeSlug()` states that slugs for non-default languages go in the file:
https://github.com/getkirby/kirby/blob/194fe296a45328b9e45eeb7bd8a956689963306e/src/Cms/PageActions.php#L85-L92

And `changeSlugForLanguage()` refuses to update the slug in the default language's content file:
https://github.com/getkirby/kirby/blob/194fe296a45328b9e45eeb7bd8a956689963306e/src/Cms/PageActions.php#L151-L153

In that sense, they're more of a bug rather than a feature.

lukasbestle:
You are right that the Panel never does this, but not everyone uses the Panel and there could well be sites where someone has used this feature manually and now forgot about it. These sites would break if we change the behavior.

distantnative:
I think the benefits and risks are clear by now.
Let's wait for @bastianallgeier to be back from holidays and see what his take on the trade-off is.

neildaniels:
I ended up refining this more and was able to remove `findBy('slug')` from `$pages->findById()` altogether. The use of `findBy('slug')` in `$pages->findByIdRecursive()` should be sufficient.

In situations where a given page id (or slug) doesn't actually exist, `findBy('slug')` would be called twice. Once in `findById()` and then again in `findByIdRecursive()`.  **This slight improvement applies to non-default languages.**

Admittedly, this PR certainly _feels_ like it might have a significant behavior change for `$pages->findById()`. I've added a variety of test scenarios to try and cover all the bases. The tests pass on both unchanged `3.4.2` code and the code in this PR.

Some of these optimizations (and certainly the tests) will still be relevant even if there's a decision to continue to allow default language slug translations.

tests/Cms/Pages/PagesTest.php
```diff
@@ -299,11 +299,188 @@ public function testFindByIdAndUri()
...
         $this->assertIsPage($site->children()->findByUri('grandma/mother/child.json'), 'grandma/mother/child');
+        $this->assertIsPage($site->children()->findById('grandma/mother')->children()->findById('child'), 'grandma/mother/child');
+        $this->assertIsPage($site->children()->findById('grandma/mother')->children()->findById('grandma/mother/child'), 'grandma/mother/child');
+//         $this->assertIsPage($site->children()->findById('child'), 'grandma/mother/child');
+//         $this->assertIsPage($site->children()->findByUri('child'), 'grandma/mother/child');
+
+        $pages = new Pages($site->children()->find('grandma', 'grandma/mother', 'grandma/mother/child'));
+        $this->assertIsPage($pages->findById('grandma'), 'grandma');
+        $this->assertIsPage($pages->findById('grandma/mother'), 'grandma/mother');
+//         $this->assertIsPage($pages->findById('mother'), 'grandma/mother');
```
neildaniels: This passes in a multilang environment. https://github.com/getkirby/kirby/blob/add854e1b9879a61fe64009992d979cb490545a5/tests/Cms/Pages/PagesTest.php#L417

tests/Cms/Pages/PagesTest.php
```diff
@@ -299,11 +299,188 @@ public function testFindByIdAndUri()
...
+        $this->assertIsPage($pages->findById('grandma/mother/child'), 'grandma/mother/child');
+//         $this->assertIsPage($pages->findById('child'), 'grandma/mother/child');
```
neildaniels: This passes in a multilang environment. https://github.com/getkirby/kirby/blob/add854e1b9879a61fe64009992d979cb490545a5/tests/Cms/Pages/PagesTest.php#L419

neildaniels:
I added more tests, mostly dealing with a `Pages()` collection that doesn't have a `$parent` assigned to it.

There are a variety of tests that I left commented out, because they fail on _both_ `3.4.2` and this branch. If anyone sees a test they think should pass (or, think should be codified as a failure case) I think this might be a good point to discuss. At a minimum, I think it helps to illustrate there shouldn't be much of a logic change. Otherwise, I can remove the commented out tests.

I commented on two specific cases that—to my surprise—fail in a single language setup, but pass on a multilang environment.

Personally, I think it's odd that they fail in the single language environment. But, this was existing behavior and I didn't try to reconcile the differences.

neildaniels:
I added another unit test to try and formalize the behavior of pages that have "swapped" slugs.

For example: imagine a site has two subpages, `aaa` and `zzz`—but a non-default translation has custom slugs that effectively swap the slugs, so it's `zzz` and `aaa`.

Somewhat unexpectedly, the current `3.4.2` behavior basically prohibits accessing a page with the translated slugs. Following the existing code, it's clear that lookups by `id` (the collection key) happen before any attempt to using translated slugs. I personally think this is slightly less "accurate" and muddles the illusion that `$pages->findById()` is fully translation-aware. The method name itself makes it seem like it only matches on `$page->id()` but in reality it _effectively_ uses `$page->uri()`. But, as this test case shows, it tends to prefer the `id()` version.

Regardless, this PR maintains the existing behavior of "swapped" slugs.

distantnative:
Ok everyone, @getkirby/kirby-staff @neildaniels,
This PR has been sitting around for way too long. We need a decision - either do it (in 3.6.0) or not and close the PR. Letting it sit isn't helping anyone (not Neil since not getting it, nor us for feeling guilty every time seeing it).

I try to summarize...

## What's suggested
- In the default language, `Page::slug()` should always be deprived from the folder name, never from the content file
- `Pages::findByID()`: remove first `->findBy('slug, $id)` call and rely completely on `->findByIdRecursive()`
- `Pages::findByIdRecursive()`: for the default language `->get()` should already return the correct item (or not), so running another `->findBy` won't yield any different result and thus should be skipped 
  - not sure if this assumption is true in regard to `->get()` - @neildaniels, see my comment to your failed tests

## Why
Performance improvement for default language in multilanguage setups

## Breaking changes
- Slugs defined in content files for the default language won't work anymore
- more?


Please let me know if I get this all right.

Personally, I am leaning towards implementing it. Having the slug come from the content field for the default language isn't something we have really considered, and I believe it isn't well supported. For the default language, the slug should come from the folder name. I don't see a benefit of offering the content file as alternative for the default language. So the performance increase might outweigh the breaking change for a few users.

tests/Cms/Pages/PagesTest.php
distantnative: Could it be that this is, because in non-default language we end up with a `findBy('slug', $id)` call that just looks for a match of the slug itself. But with the default language, we are now stuck with `->get()` that needs the full ID, not just the slug?

Not sure here myself, my head doesn't wrap fully around this yet, but a hunch I have.

neildaniels: Yes, that is accurate. What was confusing to me was why it was happening in the `en` language (which should be the default in the test case).

Turns out, in the test cases, `en` _wasn't_ actually being treated as a default language, which threw off the logic.

I've added `'default' => true,` which makes the multi-lang version constant to the single-language version.

neildaniels:
@distantnative Thank you for the excellent summarization. I'd also like to emphasize this PR should have no behavior changes (and no breaking changes) on single language setups.

Those 3 suggestions are accurate. (Your comment in regards to failed tests reflected a testing configuration bug—now fixed.)

There is an additional suggestion (which is also a breaking change): 
- In multi-lang setups, calling `$pages->findById()` on a non-children collection with an id that is actually slug no longer works. This makes the Pages class more consistent between single and multi language setups.

For example: 
```php
 // not a children collection
$pages = new Pages($site->children()->find('grandma', 'grandma/mother', 'grandma/mother/child'));

$pages->findById('grandma'); // continues to work
$pages->findById('mother'); // no longer works, but used to work in multi-lang setups
$pages->findById('child'); // no longer works, but used to work in multi-lang setups

// children collections continue to work
$site->children()->findById('grandma')->children()->findById('mother')->children()->findById('child');
```

The behavior of getting a page from `$pages->findById('mother')` or `$pages->findById('child')` in the example **never worked in a single-language setup**, so I considered this a quirk of multi-language setups. If similar behavior is desired, `$pages->findBy('slug', $slug)` should be used instead.

lukasbestle:
I agree that we should do this. Overall it's very worth it with a low risk of affecting setups of our users. Maybe we can provide a checker plugin with the release that recursively checks in the site index whether any default language file contains a `Slug` field. The plugin could even automatically rename these pages.

distantnative:
@neildaniels agreed with this, just one question: would this still work?

```php
 // not a children collection
$pages = new Pages($site->children()->find('grandma', 'grandma/mother', 'grandma/mother/child'));

$pages->findById('grandma'); // continues to work
$pages->findById('mother'); // no longer works
$pages->findById('grandma/mother'); // still works?
```

neildaniels:
@distantnative Yes, `$pages->findById('grandma/mother')` would still work and there's test case for that exact scenario in single-language, multi-lang for default language, and multi-lang for non-default language.

bastianallgeier:
@neildaniels I really apologise for the super long delay. You've all done a great job improving this PR and summarising the possible side-effects. 

3.6 will be quite a big update anyway and we decided to run an extensive beta for it during this summer. Therefor I believe that this release is perfect to get this PR implemented. We have a lot of time in July, August and September to get feedback from various users with multi-language sites. Let's finally merge this!

bastianallgeier (approved):

PR #2764, opened 2020-08-09
Panel: Prevent adding double-slashes to API request paths #2740

### LINKED ISSUE #2740: Kirby Panel: Some API request URLs have double-slashes in their path
### Describe the bug

When saving changes to pages in the Panel, some of API request URLs have double-slashes in their path (e.g. ` https://www.example.com/api//pages/test-page`). This isn’t a problem in itself, but I’ve set up Apache to clean up any URLs with extraneous double-slashes with a 301 redirect (mainly for SEO purposes), which unfortunately causes these requests to be stripped of their bodies and not update anything.

I’ve managed to fix this by changing [line 35](https://github.com/getkirby/kirby/blob/3.4.0/panel/src/api/request.js#L35) of _panel/src/api/request.js_ to:

````
return fetch(api.config.endpoint + (path.charAt(0) === "/" ? '' : “/") + path, options)
````

Another solution would be to use [308 redirects](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308) instead, but I was wondering if it’s worth accounting for this in the Panel codebase to make things a little easier for those who run into a similar situation in the future? I’m happy to create to a pull-request.

I understand that this is a bit of an edge-case, so please feel free to ignore this!

### To Reproduce
1. Edit a page in Panel
2. Save changes
3. Open the _Network_ tab in your browser’s web developer tools and check for any requests made to the API (e.g. `https://www.example.com/api//pages/test-page`).

### Kirby Version
3.4.0

### Desktop:
 - OS: macOS
 - Browser: Firefox
 - Version: 78.0.2

### PR DESCRIPTION
**Pull request by damiendart**
## Describe the PR

Fixes instances where the Panel would create API requests that have double-slashes in their path (e.g. `https://www.example.com/api//pages/test-page`).

## Related issues

- Fixes #2740 

## Ready?

- [ ] Passing all unit tests

### DIFF
--- a/panel/src/api/request.js
+++ b/panel/src/api/request.js
@@ -33,7 +33,12 @@ export default (config) => {
       this.running++;
 
       // fetch the resquest's response
-      const response = await fetch(config.endpoint + "/" + path, options);
+      const response = await fetch(
+        [config.endpoint, path].join(
+          (config.endpoint.endsWith("/") || path.startsWith("/")) ? "" : "/"
+        ),
+        options
+      );
       const text     = await response.text();
 
       try {

### CODE REVIEW

File: panel/src/api/request.js [hunk 1]
```diff
@@ -33,7 +33,12 @@ export default (config) => {
       this.running++;
 
       // fetch the resquest's response
-      const response = await fetch(config.endpoint + "/" + path, options);
+      const response = await fetch(
+        [config.endpoint, path].join(
+          (config.endpoint.endsWith("/") || path.startsWith("/")) ? "" : "/"
+        ),
+        options
+      );
```
afbora, line 41:
```suggestion
    const url = config.endpoint.replace(/\/$/, "") + "/" + path.replace(/^\//, "");
    const response = await fetch(url, options);
```

What would it be like if we wrote it like this? 🤔

distantnative: @afbora I think your solution would add a slash even if no path exists. I think that's why Bastian went for the `[].join()` way initially.

afbora: Ahh you are right.  I couldn't think it would be empty.

bastianallgeier (approved):

bastianallgeier:
I merged this manually. Thanks for the PR!

PR #2770, opened 2020-08-10
Supports query language in email templates

afbora:
## Describe the PR

Email templates now support kirby query language to avoid swallowing newlines.

### Sample `template.text.php` template usage

````
Hey {{ name }}

Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum scelerisque fringilla convallis.

Cheers
{{ site.title }}
````

## Related idea

- Closes https://github.com/getkirby/ideas/issues/589

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/Email.php
+++ b/src/Cms/Email.php
@@ -104,42 +104,36 @@ class Email
             // prepare data to be passed to template
             $data = $this->props['data'] ?? [];
 
-            // check if html/kirby/text templates exist
-            $htmlTemplate  = $this->getTemplate($this->props['template'], 'html');
-            $kirbyTemplate = $this->getTemplate($this->props['template'], 'kirby');
-            $textTemplate  = $this->getTemplate($this->props['template'], 'text');
+            // check if html/text templates exist
+            $html = $this->getTemplate($this->props['template'], 'html');
+            $text = $this->getTemplate($this->props['template'], 'text');
 
-            // according to the following templates order
-            // whichever one finds first, that template works
-            // if none is found, throws an NotFoundException
-            // - kirby template (.kirby.php extension)
-            // - html template (.html.php extension)
-            // - text template (.text.php extension)
-            // - default php template (.php extension)
-            if ($kirbyTemplate->exists() === true) {
-                $render = $kirbyTemplate->render($data);
-                $html = Str::template($render, array_merge($data, [
-                    'kirby' => App::instance(),
-                    'site'  => App::instance()->site(),
-                ]), $render);
+            // merge kirby and site objects with passed data
+            $templateData = array_merge($data, [
+                'kirby' => App::instance(),
+                'site'  => App::instance()->site(),
+            ]);
+
+            // render text template first to use in html or text body
+            if ($text->exists() === true) {
+                $textRender = $text->render($data);
+
+                $textTemplate = Str::template($textRender, $templateData, $textRender);
+            }
+
+            if ($html->exists() === true) {
+                $htmlRender = $html->render($data);
+
+                $htmlTemplate = Str::template($htmlRender, $templateData, $htmlRender);
 
                 $this->props['body'] = [
-                    'html' => $html
+                    'html' => $htmlTemplate,
+                    'text' => $textTemplate ?? null
                 ];
 
-                if ($textTemplate->exists() === true) {
-                    $this->props['body']['text'] = $textTemplate->render($data);
-                }
-            } elseif ($htmlTemplate->exists() === true) {
-                $this->props['body'] = [
-                    'html' => $htmlTemplate->render($data)
-                ];
-
-                if ($textTemplate->exists() === true) {
-                    $this->props['body']['text'] = $textTemplate->render($data);
-                }
-            } elseif ($textTemplate->exists() === true) {
-                $this->props['body'] = $textTemplate->render($data);
+            // fallback to single email text template
+            } elseif ($text->exists() === true) {
+                $this->props['body'] = $textTemplate ?? null;
             } else {
                 throw new NotFoundException('The email template "' . $this->props['template'] . '" cannot be found');
             }
--- a/src/Cms/Email.php
+++ b/src/Cms/Email.php
@@ -116,20 +116,21 @@ class Email
 
             // render text template first to use in html or text body
             if ($text->exists() === true) {
-                $textRender = $text->render($data);
-
+                $textRender   = $text->render($data);
                 $textTemplate = Str::template($textRender, $templateData, $textRender);
             }
 
             if ($html->exists() === true) {
-                $htmlRender = $html->render($data);
-
+                $htmlRender   = $html->render($data);
                 $htmlTemplate = Str::template($htmlRender, $templateData, $htmlRender);
 
                 $this->props['body'] = [
-                    'html' => $htmlTemplate,
-                    'text' => $textTemplate ?? null
+                    'html' => $htmlTemplate
                 ];
+
+                if ($text->exists() === true) {
+                    $this->props['body']['text'] = $textTemplate ?? null;
+                }
 
             // fallback to single email text template
             } elseif ($text->exists() === true) {
--- a/src/Cms/Email.php
+++ b/src/Cms/Email.php
@@ -132,7 +132,7 @@ class Email
                     $this->props['body']['text'] = $textTemplate ?? null;
                 }
 
-            // fallback to single email text template
+                // fallback to single email text template
             } elseif ($text->exists() === true) {
                 $this->props['body'] = $textTemplate ?? null;
             } else {
--- a/src/Cms/Email.php
+++ b/src/Cms/Email.php
@@ -116,28 +116,29 @@ class Email
             // - html template (.html.php extension)
             // - text template (.text.php extension)
             // - default php template (.php extension)
-            if ($kirbyTemplate->exists()) {
-                $html = Str::template($kirbyTemplate->render($data), array_merge($data, [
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/Cms/Email.php [hunk 1]
```diff
@@ -103,22 +104,41 @@ protected function template(): void
...
+            // if none is found, throws an NotFoundException
+            // - kirby template (.kirby.php extension)
+            // - html template (.html.php extension)
+            // - text template (.text.php extension)
+            // - default php template (.php extension)
+            if ($kirbyTemplate->exists()) {
+                $html = Str::template($kirbyTemplate->render($data), array_merge($data, [
+                    'kirby' => App::instance(),
+                    'site'  => App::instance()->site(),
+                ]), '');
```
afbora: May be fallback should be `$kirbyTemplate->render($data)`? 🤔

lukasbestle (commented):
To be honest I'm not sure if the `.kirby.php` templates are a semantic and easy to understand implementation.

What if you want to use the query language both for your HTML template as well as for your plain text template? Or even just for a plain text email?
Currently you could only use it for HTML templates.

Suggestion: Use the `.txt` extension for query language templates. So there could then be just `.txt` for text-only emails or `.html.txt` and `.text.txt` for HTML emails.

Of course this would mean that the template fetching code would need to be changed to support files with the `.txt` extension as well. But I think it's more useful in the long run.

afbora:
@lukasbestle I was inspired by the Laravel blade template engine that have extension is: `.blade.php`

Yes, you right! I didn't think for plain text. But frankly, I don't want it to be a `.txt` file. Because also i would like to use `php` tags.

Maybe too long but how would that look like this?

- TEXT: media.text.kirby.php
- HTML: media.html.kirby.php

Or as a better idea; let Kirby query language available by default for all regardless of extension?

lukasbestle:
> Because also i would like to use `php` tags.

I'm not sure about this. I'd say either you use PHP templating or you use the query language. Both in the same file is confusing and also wouldn't solve the issue with the line breaks that get swallowed by PHP.

> Maybe too long but how would that look like this?
> 
> * TEXT: media.text.kirby.php
> * HTML: media.html.kirby.php

That could work, but is indeed quite long.

> Or as a better idea; let Kirby query language available by default for all regardless of extension?

This would be a breaking-change. Once a user has `{{` somewhere inside the template file, Kirby would try to parse what follows as a query. Also, as I wrote: I wouldn't mix PHP and the query language to be honest.

afbora:
Do you think we can write the entire email template with just kirby query language? Loops, if else, functions, etc.. If our answer is no, i think, we should be able to use both `php` and `kirby query language`.

lukasbestle:
Of course the query language doesn't support logic like `if`, loops etc. But I'd say there are many use-cases where logic inside the email template isn't required. If it's a fixed text that only needs a few placeholders, just the query language is enough.

If you need logic, PHP templates are the better way. But if you use PHP anyway, I don't think you really need the query language on top.

afbora:
Yes, if we will not have our own template engine, which does not exist, it would be unreasonable to support both php and query languages for using only a few usage 👍  

I guess I thought and imagined it like template engine.

distantnative:
Wouldn't it be simpler to support custom template engines? Then, whoever encounters the EOL issue and is so bothered with it, can switch to Twig or so?

lukasbestle:
That would be an option, but I'd say Kirby should then choose the template engine based on the file extension.

If you want to use the different engine just for emails, you could then do that via the file extension.

afbora:
I thought a little bit about that:

1. Users will be confused when there is a `.txt` extension. It will be difficult for them to tell the difference between `.txt` and `.text.php`. Also, writing HTML in `.html.txt` won't make much sense.

2. `.kirby.php` would be acceptable, but it doesn't support plain text. (In addition to this, I have a suggestion below.)

3. In addition to the first item, I thought of `.txt` for plain text and `.html` for HTML, but again users would be confused.

For these reasons, I think the most acceptable solution is `.text.kirby.php` and `.html.kirby.php`, albeit long.
Because it shows that it is a special file type for kirby with the extension `.kirby`. So when users see this file type, they will know that it contains a special language.

---

Also i have a new suggestion. 
What do you think about that?

- `.kirby.php` for HTML
- `.kirby.txt` for Plain Text

lukasbestle:
> For these reasons, I think the most acceptable solution is `.text.kirby.php` and `.html.kirby.php`, albeit long.
> Because it shows that it is a special file type for kirby with the extension `.kirby`. So when users see this file type, they will know that it contains a special language.

I agree that `.kirby` fits better. But to be honest I don't think `.php` should be in the filename as the query language template does not actually support any PHP syntax.

So the cleanest name would be just `.kirby` without the added `.php`, but that would make it harder to open the files in text editors as `.kirby` is not common.

> Also i have a new suggestion.
> What do you think about that?
> 
> * `.kirby.php` for HTML
> * `.kirby.txt` for Plain Text

The suggestion makes a lot of sense on its own, but doesn't fit very well with our content representation syntax where the representation type is *before* the file extension.

To be honest I'm not sure what to do. I really like your suggestions and just can't decide what's best for the future. 😢

@bastianallgeier @distantnative?

distantnative:
I'm not sure if this is getting more complicated that it's actually solving. Cause I agree those extensions are getting really confusing and long.

How big do we think is the need for only Kirby query templates in emails? I would tend to think maybe not that big because many want to send HTML emails, emails using if/else structures etc.

But as far as I understand it, if we're mixing it with PHP, we are still facing the initial problem of swallowed line breaks still. 

So I don't know if this would really solve as much worth introducing a whole new format etc.

Most sense to me would be that Kirby templates are just supported in our regular text and HTML templates. This is a breaking change though. But anything else feels quite odd to me - too complicated to make sense to the user in a way.

How bad would we think that breaking change could be? How many people might use {{ }} syntax in their email templates that needs to stay? And couldn't we make it that it is kept in if no Kirby query is matched? So that really the only breaking change would be people using {{ page.title }} or so in their templates.

lukasbestle:
You are right – integrating this into our normal templates is the cleanest and also easiest solution.

The `Str::template()` implementation already falls back to the input string if no match was found, so that part is already done.

What we would need is a syntax like `\{{ ... }}` to explicitly escape query strings. That would be needed if someone wants to use a query string inside a template that should never be matched. But I'd say it should be rather simple to add to `Str::template()` with a negative lookbehind assertion.

I agree that it's very unlikely that someone already uses the double-curly syntax in templates. We could build a simple migration checker tool for those who are not sure. Or in any case it's as simple as a global file search for `{{`.

afbora:
As far as I know, the `{{ .. }}` syntax can only be encountered in those who use javascript templates or especially those who use angularjs. Laravel Blade and some template engines also use `{{ .. }}` but the kirby template component will not be affected anyway.

distantnative:
I like adding a `\{{}}` in general as a negative look behind.

lukasbestle:
So then the only open question is whether we can afford the breaking change. For users who use JS templating inside their Kirby templates (as @afbora has explained), it might be unwelcome that Kirby suddenly tries to parse these template strings.

distantnative:
@lukasbestle not if we only replace them if they are a valid Kirby query (e.g. something with `page`). Then we are only breaking those who send emails out with content about Kirby queries. I would imagine this to be so small that it is acceptable.

To be clear: my assumption is that we only do the Kirby query replacement for email templates, not normal page templates. I don't think we should introduce them there. If someone wants the `{{}}` syntax in all their templates, they should switch the template engine. Introducing Kirby queries to email templates makes sense to fix the PHP line break issue.

afbora:
In fact, when i think again, I do not think that javascript or javascript templating will be used in email templates, right? 🤔

distantnative:
Only if someone would write about these templates in e.g. a Javascript template mailing. So the only real breaking change would be for people writing about Kirby query templates.

lukasbestle:
Ah, I thought you meant that we should introduce the query language to all templates.

If it stays limited to email templates: Sure! I don't see a problem there at all.

distantnative:
I think for all templates it's too risky, cause `{{ }}` syntax happens too often in Vue, alpineJS etc. parts.

With email templates, I think we can safely make a few assumptions that we would not have such blocking problems there. I also see a real use case for email templates (preserved line break), while for page templates those don't apply - and then it would make more sense to really switch to a different template engine (as plugin), I would say.

lukasbestle:
I fully agree!

afbora:
Is the final decision to apply the kirby query language to all existing email templates without any new extension?

distantnative:
Yes, I think this needs two steps:
- [ ] Add negative look-behind to regex in `Str::template()` - to not match `\{{ something }}`
- [ ] Run email template output through `Str::template()` with passing `$kirby`, `$site` and anything passed in `data` to the template anyways.

afbora:
I've done changes about supporting query language in email templates but i leave last touch and `negative look behind` to you. And please squash all with your commits 😅

bastianallgeier:
I have to admit that I don't feel good about this one. As much as I love the idea to solve the line break issues with this and also introduce simple template placeholders based on our queries, I feel like we are opening pandora's box with this. Why would the email templates have different features than all the other PHP templates. I think this is way more related to the general discussion about template engines and shouldn't be solved this way. 

I would actually prefer to close it. I'm sorry.

afbora:
Actually, my goal was to bring the `.kirby.php` file extension to email templates first and then apply it to all templates. But it didn't make sense to me to add a new template language for swallowing issue that have simple solution.

PR #2772, opened 2020-08-10
Fix pages section in user pages

### LINKED ISSUE #2771: [Panel] User page section can't create new page (no route)
**Describe the bug**  
When adding a pages section to a user blueprint, adding a new page using the add button results in the following error:

> No route found for path: "users/VDjIQ5/children/blueprints" and request method: "GET"

**To Reproduce**  
1. Add a pages section to a user blueprint
2. Open the user page in the panel
3. Click on the +

**Expected behavior**  
That page creation works from within the user area in the panel.

**Screenshots**  
image

**Kirby Version**  
3.4.2

**Console output**  
"No route found for path: "users/hoa3biYu/children/blueprints" and request method: "GET""

**Additional context**  
Initially reported in the forums on [May 5th](https://forum.getkirby.com/t/blueprint-section-in-user-page/18136) by user Perry and reproduced by texnixe.

afbora:
## Describe the PR

The reason why page sections don't work on the user page is that there is no route defined as `/users/ID/children/blueprints`. I've also made some fixes while fixing this issue.

1. `users/(:any)/blueprints` route added
2. Add missing routes
3. Site, user and page `/blueprints` endpoints were made common. Removed `/children` segment. I don't think `/children` segment is necessary, **but this could be a break change, i'm not sure**.

## Related issues

- Fixes #2771 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/api/routes/pages.php
+++ b/config/api/routes/pages.php
@@ -41,13 +41,6 @@ return [
         }
     ],
     [
-        'pattern' => 'pages/(:any)/children/blueprints',
-        'method'  => 'GET',
-        'action'  => function (string $id) {
-            return $this->page($id)->blueprints($this->requestQuery('section'));
-        }
-    ],
-    [
         'pattern' => 'pages/(:any)/children/search',
         'method'  => 'GET|POST',
         'action'  => function (string $id) {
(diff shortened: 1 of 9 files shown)


bastianallgeier (approved):
I made a few changes and kept it backward compatible. I think we shouldn't add throws to every method, but only to those that actually throw Exceptions themselves. It's otherwise a nightmare to keep this updated.

afbora:
Thank you for the additional enhancements.

> only to those that actually throw Exceptions themselves. It's otherwise a nightmare to keep this updated.

@bastianallgeier I understand correctly. I will pay attention from now on 👍

PR #2777, opened 2020-08-12
Support extending custom file and mime types

### LINKED ISSUE #2870: Extending $isViewable (inside FileFoundation)
As another follow-up to #2840, shouldn't we at least include some new formats as being viewable in the panel? Or rather as a question: what does that mean, being viewable - isn't it more of a browser thing?

Cheers,
S1SYPHOS

afbora:
## Describe the PR

Users can register their own custom file types and mimes with the new `fileTypes` extension.

### Sample usage in plugin

````php
    'fileTypes' => [
        'm4p' => [
            'mime' => 'video/m4p',
            'type' => 'video',
        ],
        'heif' => [
            'mime' => ['image/heic', 'image/heif'],
            'type' => 'image',
            'resizable' => true,
            'viewable' => true,
        ],
        'test' => [
            'extension' => 'kql',
            'type' => 'code'
        ],
        'midi' => [
            'mime' => 'audio/x-midi'
        ],
        'ttf' => [
            'type' => 'font'
        ]
    ]
]);
````

### Features

- New `fileTypes` plugin extension to extend custom file and mime types

### Enhancements

- The `avif` file extension is now considered viewable, so it can be displayed in the Panel #2877

### Sample usage in plugin

````php
    'fileTypes' => [
        'm4p' => [
            'mime' => 'video/m4p',
            'type' => 'video',
        ],
        'heif' => [
            'mime' => ['image/heic', 'image/heif'],
            'type' => 'image',
            'resizable' => true,
            'viewable' => true,
        ]
    ]
]);
````

## Breaking changes

None

## Related issues/ideas

#2870

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -244,7 +244,7 @@ trait AppPlugins
             $resizable = $file['resizable'] ?? false;
             $viewable  = $file['viewable'] ?? false;
 
-            if ($type !== null) {
+            if (is_string($type) === true) {
                 if (isset(F::$types[$type]) === false) {
                     F::$types[$type] = [];
                 }
@@ -256,6 +256,8 @@ trait AppPlugins
 
             if ($mime !== null) {
                 if (array_key_exists($extension, Mime::$types) === true) {
+                    // if `Mime::$types[$extension]` is not already an array, make it one
+                    // and append the new MIME type unless it's already in the list
                     Mime::$types[$extension] = array_unique(array_merge((array)Mime::$types[$extension], (array)$mime));
                 } else {
                     Mime::$types[$extension] = $mime;
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -10,6 +10,7 @@ use Kirby\Toolkit\A;
 use Kirby\Toolkit\Collection as ToolkitCollection;
 use Kirby\Toolkit\Dir;
 use Kirby\Toolkit\F;
+use Kirby\Toolkit\Mime;
 use Kirby\Toolkit\V;
 
 /**
@@ -58,6 +59,7 @@ trait AppPlugins
         'collectionMethods' => [],
         'fieldMethods' => [],
         'fileMethods' => [],
+        'fileTypes' => [],
         'filesMethods' => [],
         'fields' => [],
         'hooks' => [],
@@ -223,6 +225,63 @@ trait AppPlugins
     protected function extendFileMethods(array $methods): array
     {
         return $this->extensions['fileMethods'] = File::$methods = array_merge(File::$methods, $methods);
+    }
+
+    /**
+     * Registers additional custom file types and mimes
+     *
+     * @param array $fileTypes
+     * @return array
+     */
+    protected function extendFileTypes(array $fileTypes): array
+    {
+        $extends = [
+            'type' => [],
+            'mime' => [],
+            'resizable' => [],
+            'viewable' => []
+        ];
+
+        // normalize array
+        foreach ($fileTypes as $ext => $file) {
+            $extension = $file['extension'] ?? $ext;
+
+            if (isset($file['type']) === true) {
+                $extends['type'][$file['type']][] = $extension;
+            }
+
+            if (isset($file['mime']) === true) {
+                $extends['mime'][$extension] = $file['mime'];
+            }
+
+            if (($file['resizable'] ?? false) === true) {
+                $extends['resizable'][] = $extension;
+            }
+
+            if (($file['viewable'] ?? false) === true) {
+                $extends['viewable'][] = $extension;
+            }
+        }
+
+        // inject extended extensions
+        foreach ($extends as $type => &$values) {
+            switch ($type) {
+                case 'type':
+                    $values = F::$types = array_merge_recursive(F::$types, $values);
+                    break;
+                case 'mime':
+                    $values = Mime::$types = array_merge(Mime::$types, $values);
+                    break;
+                case 'resizable':
+                    $values = F::$resizable = array_merge(F::$resizable, $values);
+                    break;
+                case 'viewable':
+                    $values = F::$viewable = array_merge(F::$viewable, $values);
+                    break;
+            }
+        }
+
+        return $this->extensions['fileTypes'] = $extends;
     }
 
     /**
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -244,32 +244,38 @@ trait AppPlugins
             $resizable = $file['resizable'] ?? false;
             $viewable  = $file['viewable'] ?? false;
 
-            if ($type !== null && in_array($extension, F::$types[$type]) === false) {
-                F::$types[$type][] = $extension;
+            if ($type !== null) {
+                if (isset(F::$types[$type]) === false) {
+                    F::$types[$type] = [];
+                }
+
+                if (in_array($extension, F::$types[$type]) === false) {
+                    F::$types[$type][] = $extension;
+                }
             }
 
             if ($mime !== null) {
-                if (array_key_exists($extension, Mime::$types) == true) {
-                    Mime::$types[$extension] = array_merge((array)Mime::$types[$extension], (array)$mime);
+                if (array_key_exists($extension, Mime::$types) === true) {
+                    Mime::$types[$extension] = array_unique(array_merge((array)Mime::$types[$extension], (array)$mime));
                 } else {
                     Mime::$types[$extension] = $mime;
                 }
             }
 
-            if ($resizable === true && in_array($extension, Image::$resizable) === false) {
-                Image::$resizable[] = $extension;
+            if ($resizable === true && in_array($extension, Image::$resizableTypes) === false) {
+                Image::$resizableTypes[] = $extension;
             }
 
-            if ($viewable === true && in_array($extension, Image::$viewable) === false) {
-                Image::$viewable[] = $extension;
+            if ($viewable === true && in_array($extension, Image::$viewableTypes) === false) {
+                Image::$viewableTypes[] = $extension;
             }
         }
 
         return $this->extensions['fileTypes'] = [
-            'type' => F::$types,
-            'mime' => Mime::$types,
-            'resizable' => Image::$resizable,
-            'viewable' => Image::$viewable
+            'type'      => F::$types,
+            'mime'      => Mime::$types,
+            'resizable' => Image::$resizableTypes,
+            'viewable'  => Image::$viewableTypes
         ];
     }
 
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -5,6 +5,7 @@ namespace Kirby\Cms;
 use Closure;
 use Kirby\Exception\DuplicateException;
 use Kirby\Form\Field as FormField;
+use Kirby\Image\Image;
 use Kirby\Text\KirbyTag;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\Collection as ToolkitCollection;
@@ -235,53 +236,41 @@ trait AppPlugins
      */
     protected function extendFileTypes(array $fileTypes): array
     {
-        $extends = [
-            'type' => [],
-            'mime' => [],
-            'resizable' => [],
-            'viewable' => []
-        ];
-
         // normalize array
         foreach ($fileTypes as $ext => $file) {
             $extension = $file['extension'] ?? $ext;
+            $type      = $file['type'] ?? null;
+            $mime      = $file['mime'] ?? null;
+            $resizable = $file['resizable'] ?? false;
+            $viewable  = $file['viewable'] ?? false;
 
-            if (isset($file['type']) === true) {
-                $extends['type'][$file['type']][] = $extension;
+            if ($type !== null && in_array($extension, F::$types[$type]) === false) {
+                F::$types[$type][] = $extension;
             }
 
-            if (isset($file['mime']) === true) {
-                $extends['mime'][$extension] = $file['mime'];
+            if ($mime !== null) {
+                if (array_key_exists($extension, Mime::$types) == true) {
+                    Mime::$types[$extension] = array_merge((array)Mime::$types[$extension], (array)$mime);
+                } else {
+                    Mime::$types[$extension] = $mime;
+                }
             }
 
-            if (($file['resizable'] ?? false) === true) {
-                $extends['resizable'][] = $extension;
+            if ($resizable === true && in_array($extension, Image::$resizable) === false) {
+                Image::$resizable[] = $extension;
             }
 
-            if (($file['viewable'] ?? false) === true) {
-                $extends['viewable'][] = $extension;
-            }
-        }
-
-        // inject extended extensions
-        foreach ($extends as $type => &$values) {
-            switch ($type) {
-                case 'type':
-                    $values = F::$types = array_merge_recursive(F::$types, $values);
-                    break;
-                case 'mime':
-                    $values = Mime::$types = array_merge(Mime::$types, $values);
-                    break;
-                case 'resizable':
-                    $values = F::$resizable = array_merge(F::$resizable, $values);
-                    break;
-                case 'viewable':
-                    $values = F::$viewable = array_merge(F::$viewable, $values);
-                    break;
+            if ($viewable === true && in_array($extension, Image::$viewable) === false) {
+                Image::$viewable[] = $extension;
             }
         }
 
-        return $this->extensions['fileTypes'] = $extends;
+        return $this->extensions['fileTypes'] = [
+            'type' => F::$types,
+            'mime' => Mime::$types,
+            'resizable' => Image::$resizable,
+            'viewable' => Image::$viewable
+        ];
     }
 
     /**
--- a/src/Image/Image.php
+++ b/src/Image/Image.php
@@ -3,7 +3,6 @@
 namespace Kirby\Image;
 
 use Kirby\Filesystem\File;
-use Kirby\Toolkit\F;
 use Kirby\Toolkit\Html;
 
 /**
@@ -32,6 +31,30 @@ class Image extends File
      * @var \Kirby\Image\Dimensions|null
      */
     protected $dimensions;
+
+    /**
+     * @var array
+     */
+    public static $resizable = [
+        'jpg',
+        'jpeg',
+        'gif',
+        'png',
+        'webp'
+    ];
+
+    /**
+     * @var array
+     */
+    public static $viewable = [
+        'avif',
+        'jpg',
+        'jpeg',
+        'gif',
+        'png',
+        'svg',
+        'webp'
+    ];
 
     /**
      * Validation rules to be used for `::match()`
@@ -164,7 +187,7 @@ class Image extends File
      */
     public function isResizable(): bool
     {
-        return in_array($this->extension(), F::$resizable) === true;
+        return in_array($this->extension(), static::$resizable) === true;
     }
 
     /**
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: tests/Toolkit/VTest.php [hunk 1]
```diff
@@ -514,7 +534,7 @@ public function inputProvider()
                     ]
                 ],
                 false,
-                'The "email" validation failed for field "email"',
+                'Please enter a valid email address for field "email"',
```
afbora, line 537: Validation message corrected

afbora: Validation message corrected

tests/Toolkit/VTest.php
```diff
@@ -558,9 +578,9 @@ public function testValue()
     public function testValueFails()
     {
         $this->expectException('Exception');
-        $this->expectExceptionMessage('The "same" validation failed');
+        $this->expectExceptionMessage('Please enter "b"');
```
afbora: Validation message corrected

tests/Toolkit/VTest.php
```diff
@@ -40,6 +40,26 @@ public function testCustomValidator()
...
+        $this->assertTrue($result);
+
+        $this->expectException('Exception');
+        $this->expectExceptionMessage('The "me" validation failed');
+
+        V::value('you', [
+            'me' => ['I', 'me', 'myself']
+        ]);
+    }
+
```
afbora: Add unit test for custom validation

afbora:
I've reworked the PR. I think ready to merge.

⚠️ Note: I've routed all calls to the `::types()` method so that I can do the extend. Since `static::$types` is now protected, there will be a break change for those using it. 

I could not find a better way. We can move this feature to 3.6. If you think it's too complicated, we can close it. If you have a better solution, I am listening to you.

We can continue using native way like with closing this PR:

````php
Mime::$types['heic'] = ['image/heic, image/heif'];
````

**Comment by bastianallgeier** [comment]
Coming back to this after quite a while I wonder if we should design the options in a way that they are more understandable for users instead of following our class names.

I.e. 

```php
<?php

return [
    'files' => [
        'mimes' => [
           'heic' => ['image/heic', 'image/heif'],
           'm4p'  => ['video/m4p']
        ],
        'types' => [
            'image' => ['heic'],
            'video' => ['m4p']
        ]
    ],
];
```

Pinging @texnixe, @distantnative, @lukasbestle here as well for input

bastianallgeier:
Or maybe even better: 

```php
return [
    'files' => [
        'types' => [
            'm4p' => [
                'extension' => 'm4p', // can be omitted. The key would be used instead
                'mime' => 'video/m4p',
                'type' => 'video', 
            ],
            'heic' => [
                'extension' => 'heic',
                'mime' => ['image/heic', 'image/heif'],
                'type' => 'image', 
                'processing' => true
            ],
        ]
    ]
];
```

bastianallgeier:
This would leave room for more detailed file type setups in the future.

afbora:
@bastianallgeier So what do you think about switching from static `$types` variable to static `types()` method to handle extends and caching. Because I was most confused there.

Should we do the extendion before? Any ideas?

bastianallgeier:
@afbora we could do that, but it would also be a breaking change that might not be necessary right now.

lukasbestle:
@afbora I think we should implement it like the `validator` extension that [currently directly sets validators in `V::$validators`](https://github.com/getkirby/kirby/blob/3.5.5/src/Cms/AppPlugins.php#L519). Making the `$types` property `protected` would be an unnecessary breaking change and also inconsistent with validators, collection filters, model methods...

Validators are however registered as plugin extensions while this PR uses a Kirby option. To be honest I feel like the file types should also be a plugin extension and not an option – this increases the use-cases for this feature and is again more consistent.

Regarding the registration syntax: I like the second syntax suggested by Bastian as it's the most flexible and easiest to extend later.

afbora:
@lukasbestle It makes more sense to extend it like in Validator 👍

afbora:
@lukasbestle I applied new idea:

````php
<?php

Kirby::plugin('my/filetypes', [
    'fileTypes' => [
        'm4p' => [
            'mime' => 'video/m4p',
            'type' => 'video',
        ],
        'heic' => [
            'mime' => ['image/heic', 'image/heif'],
            'type' => 'image',
            'resizable' => true,
            'viewable' => true,
        ],
        'test' => [
            'extension' => 'kirby',
            'type' => 'code'
        ]
    ]
];
````

Since we extended 4 different pieces of information (type, mime, resizable, viewable), I had to normalize the array unlike the others. 

---

Or we can make 4 different extensions like following:

````php
<?php

Kirby::plugin('my/filetypes', [
    'files' => [
        'types' => [],
        'mimes' => [],
        'resizable' => [],
        'viewable' => []
    ]
];
````

What do you think?

File: src/Toolkit/F.php [hunk 4]
```diff
@@ -105,8 +108,35 @@ class F
...
+    /**
+     * @var array
+     */
     public static $units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
 
+    /**
+     * @var array
+     */
+    public static $viewable = [
+        'avif',
```
afbora, line 131: `AVIF` added to viewable as default. Should it be in the core or extended? #2877

src/Toolkit/F.php
```diff
@@ -105,8 +108,35 @@ class F
         ],
     ];
 
+    /**
```
distantnative, line 111: Since `$resizable` and `$viewable` are directly connected (and only used by) `Kirby\Image|Image`, I think they could make more sense in that class than in `F`. But curious to hear other opinions first before changing this in your PR already. @lukasbestle maybe?

afbora:
CI tests fixed. Ready to review again.

src/Cms/AppPlugins.php
```diff
@@ -225,6 +227,63 @@ protected function extendFileMethods(array $methods): array
...
+            if (($file['viewable'] ?? false) === true) {
+                $extends['viewable'][] = $extension;
+            }
+        }
+
+        // inject extended extensions
+        foreach ($extends as $type => &$values) {
+            switch ($type) {
+                case 'type':
+                    $values = F::$types = array_merge_recursive(F::$types, $values);
```
afbora: Who has better suggestions than `array_merge_recursive()`?

lukasbestle: I think it's fine for the purpose. An alternative could be `array_replace_recursive()`, but it does the same for our array structure as far as I remember.

Do you see any downsides to using this function?

Another alternative could be to restructure the method like this:

- The `$extends` variable at the beginning is removed
- Instead, the loop through `$fileTypes` updates the properties of the different classes directly
- The `$extends` variable is then constructed *afterwards* with the values from the class properties

This would also allow us to get rid of the second `foreach` with `switch` that is currently a bit hard to understand.

src/Toolkit/F.php
lukasbestle: I agree with Nico.

src/Cms/AppPlugins.php
afbora: > Do you see any downsides to using this function?

I had a doubt in my mind but false alarm, it's fine. But as you said, I think it is more correct to use `array_replace_recursive()` to avoid duplicate entry.

> Instead, the loop through $fileTypes updates the properties of the different classes directly

You mean like following?

````php
// normalize array
foreach ($fileTypes as $ext => $file) {
    $extension = $file['extension'] ?? $ext;

    if (isset($file['type']) === true) {
        // inject to `F::$types` each fileType entry ??
        F::$types = array_replace_recursive(F::$types, [$file['type'] => [$extension]]);
    }
    
    ...
    ...
    ...
}
````

As I have concern in above code;
Since using `array_merge()` each time while normalizing each type creates a performance problem, I first put it into a temporary array and then merged it. Am I wrong?

File: src/Toolkit/F.php (hunk 5 above)
Comment (afbora) [reply src/Toolkit/F.php:111]: Thank you both for reviewing 👍 I'll implement after final review.

File: src/Cms/AppPlugins.php (hunk 6 above)
Comment (lukasbestle) [reply src/Cms/AppPlugins.php]: Yep, pretty much. But wouldn't a simple `F::$types[$file['type']][] = $extension` work as well? The only downside would be that an extension could end up in the array multiple times. We can solve this by first checking if the extension is not already defined.

File: src/Toolkit/F.php (hunk 5 above)
Comment (afbora) [reply src/Toolkit/F.php:111]: Done

File: src/Cms/AppPlugins.php (hunk 6 above)
Comment (afbora) [reply src/Cms/AppPlugins.php]: Sometimes I overthink 😅

Comment (afbora) [reply src/Cms/AppPlugins.php]: How does it look now?

File: src/Cms/AppPlugins.php [hunk 7]
```diff
@@ -225,6 +228,51 @@ protected function extendFileMethods(array $methods): array
...
+            $resizable = $file['resizable'] ?? false;
+            $viewable  = $file['viewable'] ?? false;
+
+            if ($type !== null && in_array($extension, F::$types[$type]) === false) {
+                F::$types[$type][] = $extension;
+            }
+
+            if ($mime !== null) {
+                if (array_key_exists($extension, Mime::$types) == true) {
+                    Mime::$types[$extension] = array_merge((array)Mime::$types[$extension], (array)$mime);
```
afbora: Since we do not know whether the current value and the sent values are strings or arrays, we cast both variables to the array.

src/Cms/AppPlugins.php
```diff
@@ -225,6 +228,51 @@ protected function extendFileMethods(array $methods): array
...
+        foreach ($fileTypes as $ext => $file) {
+            $extension = $file['extension'] ?? $ext;
+            $type      = $file['type'] ?? null;
+            $mime      = $file['mime'] ?? null;
+            $resizable = $file['resizable'] ?? false;
+            $viewable  = $file['viewable'] ?? false;
+
+            if ($type !== null && in_array($extension, F::$types[$type]) === false) {
+                F::$types[$type][] = $extension;
+            }
```
lukasbestle: To avoid errors for types that don't already exist:

```suggestion
            if ($type !== null) {
                if (isset(F::$types[$type]) !== true) {
                    F::$types[$type] = [];
                }

                if (in_array($extension, F::$types[$type]) === false) {
                    F::$types[$type][] = $extension;
                }
            }
```

src/Cms/AppPlugins.php
```diff
@@ -225,6 +228,51 @@ protected function extendFileMethods(array $methods): array
...
+
+            if ($mime !== null) {
+                if (array_key_exists($extension, Mime::$types) == true) {
```
lukasbestle:
```suggestion
                if (array_key_exists($extension, Mime::$types) === true) {
```

src/Cms/AppPlugins.php
lukasbestle: > Since we do not know whether the current value and the sent values are strings or arrays, we cast both variables to the array.

Good point!

One minor change to ensure that the same type is only included once:

```suggestion
                    Mime::$types[$extension] = array_unique(array_merge((array)Mime::$types[$extension], (array)$mime));
```

src/Image/Image.php
```diff
@@ -32,6 +32,30 @@ class Image extends File
      */
     protected $dimensions;
 
+    /**
+     * @var array
+     */
+    public static $resizable = [
```
lukasbestle: Idea for a more speaking name that can also avoid conflicts later if we ever need to have instance properties with the same name:

```suggestion
    public static $resizableTypes = [
```

src/Image/Image.php
```diff
@@ -32,6 +32,30 @@ class Image extends File
...
+        'jpeg',
+        'gif',
+        'png',
+        'webp'
+    ];
+
+    /**
+     * @var array
+     */
+    public static $viewable = [
```
```suggestion
    public static $viewableTypes = [
```

src/Image/Image.php
```diff
@@ -163,15 +187,7 @@ public function isSquare(): bool
...
-        $resizable = [
-            'jpg',
-            'jpeg',
-            'gif',
-            'png',
-            'webp'
-        ];
-
-        return in_array($this->extension(), $resizable) === true;
+        return in_array($this->extension(), static::$resizable) === true;
```
```suggestion
        return in_array($this->extension(), static::$resizableTypes) === true;
```

src/Image/Image.php
```diff
@@ -182,16 +198,7 @@ public function isResizable(): bool
...
-            'jpg',
-            'jpeg',
-            'gif',
-            'png',
-            'svg',
-            'webp'
-        ];
-
-        return in_array($this->extension(), $viewable) === true;
+        return in_array($this->extension(), static::$viewable) === true;
```
```suggestion
        return in_array($this->extension(), static::$viewableTypes) === true;
```

src/Cms/AppPlugins.php
```diff
@@ -225,6 +228,51 @@ protected function extendFileMethods(array $methods): array
...
+                Image::$viewable[] = $extension;
+            }
+        }
+
+        return $this->extensions['fileTypes'] = [
+            'type' => F::$types,
+            'mime' => Mime::$types,
+            'resizable' => Image::$resizable,
+            'viewable' => Image::$viewable
+        ];
```
lukasbestle:
```suggestion
            if ($resizable === true && in_array($extension, Image::$resizableTypes) === false) {
                Image::$resizableTypes[] = $extension;
            }

            if ($viewable === true && in_array($extension, Image::$viewableTypes) === false) {
                Image::$viewableTypes[] = $extension;
            }
        }

        return $this->extensions['fileTypes'] = [
            'type' => F::$types,
            'mime' => Mime::$types,
            'resizable' => Image::$resizableTypes,
            'viewable' => Image::$viewableTypes
        ];
```

tests/Cms/App/AppPluginsTest.php
```diff
@@ -982,4 +988,60 @@ public function testNativeComponents()
...
+                    'mime' => 'audio/x-midi'
+                ]
+            ]
+        ]);
+
+        $fileTypes = $kirby->extensions('fileTypes');
+        $this->assertSame($fileTypes['type'], F::$types);
+        $this->assertSame($fileTypes['mime'], Mime::$types);
+        $this->assertSame($fileTypes['resizable'], Image::$resizable);
+        $this->assertSame($fileTypes['viewable'], Image::$viewable);
```
lukasbestle:
```suggestion
        $this->assertSame($fileTypes['resizable'], Image::$resizableTypes);
        $this->assertSame($fileTypes['viewable'], Image::$viewableTypes);
```

tests/Cms/App/AppPluginsTest.php
```diff
@@ -982,4 +988,60 @@ public function testNativeComponents()
...
+
+        $this->assertContains('m4p', F::$types['video']);
+        $this->assertArrayHasKey('m4p', Mime::$types);
+        $this->assertSame('video/m4p', Mime::$types['m4p']);
+        $this->assertNotContains('m4p', Image::$resizable);
+        $this->assertNotContains('m4p', Image::$viewable);
```
lukasbestle:
```suggestion
        $this->assertNotContains('m4p', Image::$resizableTypes);
        $this->assertNotContains('m4p', Image::$viewableTypes);
```

tests/Cms/App/AppPluginsTest.php
```diff
@@ -982,4 +988,60 @@ public function testNativeComponents()
...
+
+        $this->assertContains('heif', F::$types['image']);
+        $this->assertArrayHasKey('heif', Mime::$types);
+        $this->assertSame(['image/heic', 'image/heif'], Mime::$types['heif']);
+        $this->assertContains('heif', Image::$resizable);
+        $this->assertContains('heif', Image::$viewable);
```
lukasbestle:
```suggestion
        $this->assertContains('heif', Image::$resizableTypes);
        $this->assertContains('heif', Image::$viewableTypes);
```

tests/Cms/App/AppPluginsTest.php
```diff
@@ -982,4 +988,60 @@ public function testNativeComponents()
...
+
+        $this->assertContains('kql', F::$types['code']);
+        $this->assertNotContains('kql', Image::$resizable);
+        $this->assertNotContains('kql', Image::$viewable);
```
lukasbestle:
```suggestion
        $this->assertNotContains('kql', Image::$resizableTypes);
        $this->assertNotContains('kql', Image::$viewableTypes);
```

lukasbestle (commented):
It's so great already! I really love how simple it is now. ❤️

I went over the code changes and added some suggestions. They appear to be more than they actually are because of the suggested property renaming. I wanted to allow committing them directly from the GitHub UI. You can add all of them to the batch and commit them at once if you agree with the suggestions.

src/Cms/AppPlugins.php
```diff
@@ -225,6 +228,57 @@ protected function extendFileMethods(array $methods): array
...
+    {
+        // normalize array
+        foreach ($fileTypes as $ext => $file) {
+            $extension = $file['extension'] ?? $ext;
+            $type      = $file['type'] ?? null;
+            $mime      = $file['mime'] ?? null;
+            $resizable = $file['resizable'] ?? false;
+            $viewable  = $file['viewable'] ?? false;
+
+            if ($type !== null) {
```
lukasbestle:
```suggestion
            if (is_string($type) === true) {
```

src/Cms/AppPlugins.php
```diff
@@ -225,6 +228,57 @@ protected function extendFileMethods(array $methods): array
...
+                if (isset(F::$types[$type]) !== true) {
+                    F::$types[$type] = [];
+                }
+
+                if (in_array($extension, F::$types[$type]) === false) {
+                    F::$types[$type][] = $extension;
+                }
+            }
+
+            if ($mime !== null) {
```
lukasbestle, line 257:
```suggestion
            if (is_string($mime) === true) {
```

src/Cms/AppPlugins.php
```diff
@@ -225,6 +228,57 @@ protected function extendFileMethods(array $methods): array
...
+                if (array_key_exists($extension, Mime::$types) === true) {
+                    Mime::$types[$extension] = array_unique(array_merge((array)Mime::$types[$extension], (array)$mime));
```
lukasbestle, line 261:
```suggestion
                    // if `Mime::$types[$extension]` is not already an array, make it one
                    // and append the new MIME type unless it's already in the list
                    Mime::$types[$extension] = array_unique(array_merge((array)Mime::$types[$extension], (array)$mime));
```

tests/Cms/App/AppPluginsTest.php
```diff
@@ -982,4 +988,72 @@ public function testNativeComponents()
...
+        $this->assertArrayHasKey('heif', Mime::$types);
+        $this->assertSame(['image/heic', 'image/heif'], Mime::$types['heif']);
+        $this->assertContains('heif', Image::$resizableTypes);
+        $this->assertContains('heif', Image::$viewableTypes);
+
+        $this->assertContains('kql', F::$types['code']);
+        $this->assertNotContains('kql', Image::$resizableTypes);
+        $this->assertNotContains('kql', Image::$viewableTypes);
+        $this->assertNotContains('kql', Image::$resizableTypes);
+        $this->assertNotContains('kql', Image::$viewableTypes);
```
lukasbestle: This is duplicated:

```suggestion
```

lukasbestle (commented):
Thanks for adding the test for the unknown file type.

While reading through the code again I noticed four minor things.
Apart from this the PR is ready for merging.

src/Cms/AppPlugins.php
afbora: @lukasbestle I had to revert this change because of mime can be string or array.

lukasbestle (approved):
Oops, you are right, the MIME allows arrays. 🤦

Awesome work, it's really great now! 👍
Could you please update the PR description? Feel free to merge the PR then when you are ready.

PR #2779, opened 2020-08-12
New `navigation` option for prev/next pages in Panel

afbora:
## Describe the PR

New `navigation` option for prev/next pages.

### 🎉  Features
- New `navigation` option for prev/next pages in Panel

### Sample Usages

````yaml

title: Simple Page

# with all
navigation:
    status: all
    template: all
  
# with specific selects
navigation:
    status:
       - draft
       - unlisted
    template:
       - album
       - default

# with single select and custom sort
navigation:
    status: listed
    template: default
    sortBy: title asc
```` 

### Notes
 -  If the `navigation` contains two different templates, each must be defined separately
 - If no template or status is specified, the template and status of the selected/entered page are taken as a basis

## Breaking changes

None

## Related issues/ideas

- https://kirby.nolt.io/171

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/api/models/Page.php
+++ b/config/api/models/Page.php
@@ -44,6 +44,12 @@ return [
         'isSortable' => function (Page $page) {
             return $page->isSortable();
         },
+        /**
+         * @deprecated 3.6.0
+         * @todo Throw deprecated warning in 3.7.0
+         * @todo Remove in 3.8.0
+         * @codeCoverageIgnore
+         */
         'next' => function (Page $page) {
             return $page
                 ->nextAll()
@@ -58,12 +64,13 @@ return [
         'options' => function (Page $page) {
             return $page->panel()->options(['preview']);
         },
+        /**
+         * @todo Remove in 3.7.0
+         * @codeCoverageIgnore
+         */
         'panelIcon' => function (Page $page) {
-            // TODO: remove in 3.7.0
-            // @codeCoverageIgnoreStart
             deprecated('The API field page.panelIcon has been deprecated and will be removed in 3.7.0. Use page.panelImage instead');
             return $page->panel()->image();
-        // @codeCoverageIgnoreEnd
         },
         'panelImage' => function (Page $page) {
             return $page->panel()->image();
@@ -74,6 +81,12 @@ return [
         'parents' => function (Page $page) {
             return $page->parents()->flip();
         },
+        /**
+         * @deprecated 3.6.0
+         * @todo Throw deprecated warning in 3.7.0
+         * @todo Remove in 3.8.0
+         * @codeCoverageIgnore
+         */
         'prev' => function (Page $page) {
             return $page
                 ->prevAll()
--- a/src/Panel/Page.php
+++ b/src/Panel/Page.php
@@ -2,6 +2,8 @@
 
 namespace Kirby\Panel;
 
+use Kirby\Cms\Collection;
+
 /**
  * Provides information about the page model for the Panel
  * @since 3.6.0
@@ -127,6 +129,77 @@ class Page extends Model
     }
 
     /**
+     * Returns navigation array with
+     * previous and next page
+     * based on blueprint definition
+     *
+     * @internal
+     *
+     * @return array
+     */
+    public function prevNext(): array
+    {
+        $page = $this->model;
+
+        // create siblings collection based on
+        // blueprint navigation
+        $siblings = function (string $direction) use ($page) {
+            $navigation = $page->blueprint()->navigation();
+            $sortBy     = $navigation['sortBy'] ?? null;
+            $status     = $navigation['status'] ?? null;
+            $template   = $navigation['template'] ?? null;
+            $direction  = $direction === 'prev' ? 'prev' : 'next';
+
+            // if status is defined in navigation,
+            // all items in the collection are used
+            // (drafts, listed and unlisted) otherwise
+            // it depends on the status of the page
+            $siblings = $status !== null ? $page->parentModel()->childrenAndDrafts() : $page->siblings();
+
+            // sort the collection if custom sortBy
+            // defined in navigation otherwise
+            // default sorting will apply
+            if ($sortBy !== null) {
+                $siblings = $siblings->sort(...$siblings::sortArgs($sortBy));
+            }
+
+            $siblings = $page->{$direction . 'All'}($siblings);
+
+            if (empty($navigation) === false) {
+                $statuses  = (array)($status ?? $page->status());
+                $templates = (array)($template ?? $page->intendedTemplate());
+
+                // do not filter if template navigation is all
+                if (in_array('all', $templates) === false) {
+                    $siblings = $siblings->filter('intendedTemplate', 'in', $templates);
+                }
+
+                // do not filter if status navigation is all
+                if (in_array('all', $statuses) === false) {
+                    $siblings = $siblings->filter('status', 'in', $statuses);
+                }
+            } else {
+                $siblings = $siblings
+                    ->filter('intendedTemplate', $page->intendedTemplate())
+                    ->filter('status', $page->status());
+            }
+
+            return $siblings->filter('isReadable', true);
+        };
+
+        return [
+            'next' => function () use ($siblings) {
+                $next = $siblings('next')->first();
+                return $next ? $next->panel()->toLink('title') : null;
+            },
+            'prev'   => function () use ($siblings) {
+                $prev = $siblings('prev')->last();
+                return $prev ? $prev->panel()->toLink('title') : null;
+            }
+        ];
+    }
+
+    /**
      * Returns the data array for the
      * view's component props
      *
@@ -138,39 +211,25 @@ class Page extends Model
     {
         $page = $this->model;
 
-        return array_merge(parent::props(), [
-            'model' => [
-                'content'    => $this->content(),
-                'id'         => $page->id(),
-                'parent'     => $page->parentModel()->panel()->url(true),
-                'previewUrl' => $page->previewUrl(),
-                'status'     => $page->status(),
-                'title'      => $page->title()->toString(),
-            ],
-            'next' => function () use ($page) {
-                $next = $page
-                    ->nextAll()
-                    ->filterBy('intendedTemplate', $page->intendedTemplate())
-                    ->filterBy('status', $page->status())
-                    ->filterBy('isReadable', true)
-                    ->first();
-                return $next ? $next->panel()->prevnext('title') : null;
-            },
-            'prev'   => function () use ($page) {
-                $prev = $page
-                    ->prevAll()
-                    ->filterBy('intendedTemplate', $page->intendedTemplate())
-                    ->filterBy('status', $page->status())
-                    ->filterBy('isReadable', true)
-                    ->last();
-                return $prev ? $prev->panel()->prevnext('title') : null;
-            },
-            'status' => function () use ($page) {
-                if ($status = $page->status()) {
-                    return $page->blueprint()->status()[$status] ?? null;
-                }
-            },
-        ]);
+        return array_merge(
+            parent::props(),
+            $this->prevNext(),
+            [
+                'model' => [
+                    'content'    => $this->content(),
+                    'id'         => $page->id(),
+                    'parent'     => $page->parentModel()->panel()->url(true),
+                    'previewUrl' => $page->previewUrl(),
+                    'status'     => $page->status(),
+                    'title'      => $page->title()->toString(),
+                ],
+                'status' => function () use ($page) {
+                    if ($status = $page->status()) {
+                        return $page->blueprint()->status()[$status] ?? null;
+                    }
+                },
+            ]
+        );
     }
 
     /**
--- a/src/Panel/File.php
+++ b/src/Panel/File.php
@@ -251,31 +251,57 @@ class File extends Model
             'asc'
         );
 
-        return array_merge(parent::props(), [
-            'model' => [
-                'content'    => $this->content(),
-                'dimensions' => $file->dimensions()->toArray(),
-                'extension'  => $file->extension(),
-                'filename'   => $file->filename(),
-                'id'         => $file->id(),
-                'mime'       => $file->mime(),
-                'niceSize'   => $file->niceSize(),
-                'parent'     => $file->parent()->panel()->path(),
-                'panelImage' => $this->image([], 'cards'),
-                'previewUrl' => $file->previewUrl(),
-                'url'        => $file->url(),
-                'template'   => $file->template(),
-                'type'       => $file->type(),
-            ],
+        return array_merge(
+            parent::props(),
+            $this->prevNext(),
+            [
+                'model' => [
+                    'content'    => $this->content(),
+                    'dimensions' => $file->dimensions()->toArray(),
+                    'extension'  => $file->extension(),
+                    'filename'   => $file->filename(),
+                    'id'         => $file->id(),
+                    'mime'       => $file->mime(),
+                    'niceSize'   => $file->niceSize(),
+                    'parent'     => $file->parent()->panel()->path(),
+                    'panelImage' => $this->image([], 'cards'),
+                    'previewUrl' => $file->previewUrl(),
+                    'url'        => $file->url(),
+                    'template'   => $file->template(),
+                    'type'       => $file->type(),
+                ]
+            ]
+        );
+    }
+
+    /**
+     * Returns navigation array with
+     * previous and next file
+     *
+     * @internal
+     *
+     * @return array
+     */
+    public function prevNext(): array
+    {
+        $file     = $this->model;
+        $siblings = $file->templateSiblings()->sortBy(
+            'sort',
+            'asc',
+            'filename',
+            'asc'
+        );
+
+        return [
             'next' => function () use ($file, $siblings): ?array {
                 $next = $siblings->nth($siblings->indexOf($file) + 1);
-                return $next ? $next->panel()->prevnext('filename') : null;
+                return $next ? $next->panel()->toLink('filename') : null;
             },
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

**Comment by afbora** [comment]
Currently cycle definition can be defined separately for the pages we browse.

**Current PR implement** 

 - blog.yml
   - post.yml (cycle)
   - article.yml (cycle)

It may also be more accurate to define the cycle definition for its parent. I may have misunderstood the idea, i am not sure.

**Parent implement idea** 
 - blog.yml (cycle)
   - post.yml
   - article.yml

**Comment by bastianallgeier** [comment]
I think this is brilliant and it's also way better to specify this on a page level and not on a parent level.

**Review by lukasbestle — COMMENTED** [review]
What about the following scenario: Template A allows cycling between all siblings and template B only allows cycling between pages of template B.

If you now open a page of template A and click the "next" arrow to cycle to a page of template B, you cannot go back with the "prev" arrow.

This issue is basically impossible to solve, but we should keep it in mind.

**Comment by afbora** [comment]
@lukasbestle that's why I shared "Parent implement idea"  in above comment.

**Comment by bastianallgeier** [comment]
As much as I love this PR, we won't be able to add it to 3.5. Either 3.6 or a 3.5.x release.

**Comment by afbora** [comment]
PR is ready for 3.6 but what do you think? Do you have a different idea in mind?

**Comment by bastianallgeier** [comment]
I think there are two open questions

- Naming: I'm not 100% happy with cycle. I think it's not immediately clear what it does. Maybe pagination?
- What about the edge case that @lukasbestle described. Do we simply accept that this is possible? I cannot think of an ideal situation how the parent blueprint would be able to handle it.

**Comment by afbora** [comment]
1. Yes, I have concerns about the name, too. A name that gives more insight into its function would be great. The name `pagination` seems a bit lacking. Because that's not exactly what it does. What would the name `navigate` be like?

2. The scenario that Lukas mentioned could come true. If there are different templates under a parent, they must copy the same setting to each template blueprints. I know, this is not very useful. But the parental setting is easy to set up and solves the possible scenario. Because the setting is always done in one place on the parent blueprint and affects all primary children.

**Comment by bastianallgeier** [comment]
I thought some more about the parent solution and it's not going to work. Just imagine a dashboard (site.yml) with multiple page types like in the starterkit. How would the site.yml properly set navigation rules for all those different page types. I still think it needs to happen per page type and such edge cases should be avoided by devs.

**Comment by afbora** [comment]
An example cycle for `site.yml`. Do you think it wouldn't work?

````yaml
# site.yml cycle definition with all
cycle:
    status: all
    template: all
       
# site.yml cycle definition with specific templates
cycle:
    status: all
    template:
       - album
       - blog
       - about
       - default
````


We are reading the cycle setting of the current page type. Instead, we will read the cycle setting of the parent page.

````php
$cycle = $this->parent()->blueprint()->cycle() ?? [];
````

**Comment by afbora** [comment]
I guess I found a cycle bug. I think there are some edge cases for both ideas. 
If the cycle set for a template is under another parent, it will mix. 
The both ideas seems unable to fix this bug.

**For example:**

````yaml
# default.yml
cycle:
  status: all
  template:
    - default
    - "post" or "service" ambiguous ??
````

**Content structure**

````
- /blog
- - /post-1
- - /post-2
- - /default-1
- /services
- - /service-1
- - /service-2
- - /default-1
````

**Comment by bastianallgeier** [comment]
Yes, exactly. That's why I think it should still be on the individual page. 

I like `navigation` as term for the option.

**Comment by afbora** [comment]
You mean specific page by individual page, right? Because a setting dependent on `templates/blueprint` types will not work.

For example, this might work (uri based), but how useful it is is debatable:

````yaml
# default.yml
navigation:
  - uri: /blog
    status: all
    template:
      - default
      - post
  - uri: /services
    status: all
    template:
      - default
      - service

**Comment by lukasbestle** [comment]
I agree, it's literally impossible to fix this on our end. The question is if this means we shouldn't implement this feature at all or if it just means it's the dev's responsibility to use it correctly. I think the latter.

Currently not sure about the naming either.

**Comment by bastianallgeier** [comment]
Ok, we need a decision here. 

- I'm voting for a placement in the page blueprint - not the parent blueprint
- I'm also voting for navigation.

**Comment by distantnative** [comment]
+1

Page itself makes most sense to me. Logical issues aren't avoidable, developer has to make sure to set it up properly.

Fine with `navigation`.

**Comment by afbora** [comment]
@distantnative @bastianallgeier Let's say you have a setup like below. You must apply this setting to the blueprint file of all specified templates. Can I explain the scenario correctly?

````yaml
navigation:
  status: all
  template:
    - album
    - blog
    - about
    - default
````

**Comment by distantnative** [comment]
Yes. But it doesn't look to me like a common scenario that you want to cycle through pages of all statuses in a folder that hosts pages with 5+ different templates but you really want to cycle only through 4 of them.

A more likely scenario I'm just thinking of and probably not covered yet:
I have a blog with draft and published posts. All have the same template. If I'm viewing a draft I want to cycle through all other drafts. If I'm on a published post, I want to cycle through all other published posts.

**Comment by afbora** [comment]
> Yes. But it doesn't look to me like a common scenario

Yes, this was one of those rare scenarios. I just wanted to show the use cases.

In your example, it would be sufficient to add `status: all` to only the blueprint file of the relevant template in order to navigate between the pages of all the statuses in the same template.

**Comment by distantnative** [comment]
But the blog posts would all share the same template/blueprint.

**Comment by afbora** [comment]
I understand the scenario better now. For this, we can add a property like `status: current`.

**Comment by distantnative** [comment]
Or we add query support? Then you could do `status: page.status`

**Comment by afbora** [comment]
### Summarize

- Rebased
- Actually kept the initial state of functioning in PR
- Navigation continues to work on a page-based basis
- `cycle` name changed to `navigation`
- Unit tests will be add after reviewing
- The `status: page.status` is no longer needed because if no template or status is specified, it is based on the information of the selected/entered page.

**Review by bastianallgeier — DISMISSED** [review]
I think it's really great. I would just suggest one more change. I think $page->nextNavigation() should be $page->panel()->next() or $page->panel()->nextNavigation(). @distantnative what do you think?

**Comment by distantnative** [comment]
I agree, this shouldn't be in the Cm/s classes but Panel classes. I was wondering in general if we could condense it into 1-2 methods, currently it bloats up the classes quite a lot for just a side aspect.

Also, i don't think we should change the API responses (rather we should deprecate those anyways), but we need to change the props in the Panel model methods used via Fiber.

**Comment by afbora** [comment]
### Summarize

- Moved `navigation` feature from `\Kirby\Cms` to `\Kirby\Panel` namespace
- Uncovered parts unit tests will be add after review
- ⚠️ Critical update access modifiers on `PageSiblings::siblingsCollection()`, See review: https://github.com/getkirby/kirby/pull/2779/files#r648544846
- I couldn't compress the methods too much as Nico mentioned. I believe we can do this better in the second round with your opinions 😇

File: src/Cms/PageSiblings.php [hunk 1]
```diff
@@ -118,7 +118,7 @@ public function prevUnlisted($collection = null)
      *
      * @return \Kirby\Cms\Collection
      */
-    protected function siblingsCollection()
+    public function siblingsCollection()
```
afbora: ⚠️ Updated access modifier to access from panel namespace. Need your thoughts.

src/Panel/Page.php
```diff
@@ -194,4 +186,83 @@ public function route(): array
...
+    {
+        $page       = $this->model;
+        $model      = $page->parentModel();
+        $navigation = $page->blueprint()->navigation();
+        $sortBy     = $navigation['sortBy'] ?? null;
+        $status     = $navigation['status'] ?? null;
+
+        // if status is defined in navigation, all items in the collection are used (drafts, listed and unlisted)
+        // otherwise it depends on the status of the page
+        $collection = $status !== null ? $model->childrenAndDrafts() : $page->siblingsCollection();
```
bastianallgeier: Wouldn't it be possible to simply use `$page->siblings()` here instead to avoid the change to siblingsCollection?

config/api/models/Page.php
```diff
@@ -45,12 +45,7 @@
             return $page->isSortable();
         },
         'next' => function (Page $page) {
-            return $page
-                ->nextAll()
-                ->filter('intendedTemplate', $page->intendedTemplate())
-                ->filter('status', $page->status())
-                ->filter('isReadable', true)
-                ->first();
+            return $page->panel()->next();
```
distantnative: Should we actually change these? I think I would rather deprecate them, they don't seem to make much sense for the API now that the Panel isn't using this anymore.

src/Panel/Page.php
```diff
@@ -194,4 +186,83 @@ public function route(): array
             'title'      => $page->title()->toString(),
         ];
     }
+
+    /**
+     * Returns the next page in defined navigation
+     *
+     * @return \Kirby\Cms\Collection
+     */
+    public function next()
```
distantnative: I think we need to consider names here since we also have `Model::prevnext()` - so to see what they actually do. Also since `next()` and `prev()` are so similar, maybe it could be one method that receives the direction as parameter.

distantnative: And to bundle all the functionality, maybe it makes sense to have one method that combines `next()` and `prev()` but also integrates the code from `filter()` and `siblings()`

src/Panel/Page.php
afbora: Enlightening 😅 Thank you so much 🙏

src/Panel/Page.php
afbora: Firstly I'm thinking of changing about `prevnext()` method name. This method doesn't return previous or next collection. Only returns link and tooltip of current model, right?. May be `info` name more clear or something like that.

src/Cms/PageSiblings.php
afbora: Reverted.

src/Panel/Page.php
afbora: Also I would like `prev()` and `next()` to be separate methods and to be consistent with what we do in other classes.

config/api/models/Page.php
afbora: I'm asking because I'm not entirely sure: Except of the Panel, isn't this used in API requests?

afbora: Also, even if we deprecate it, I don't see any harm in changing it. Because it is one line and does the same job.

bastianallgeier: I would deprecate them as well, but keep them for now with the added deprecate() call and remove the methods in 3.7.

src/Panel/Page.php
bastianallgeier: Wouldn't it make sense to keep `prev()` and `next()` as separate methods and rename `prevnext()` into `navigation()`. Then it would be more in line with the blueprint settings. Just an idea.

afbora: `navigation()` name more clear/accurate than `prevnext()` 👍

afbora:
### Summary

- Keep `prev` and `next` method as separated
- `prevnext` renamed to `navigation`
- Rebased 
- Unit tests added
- ⚠️ I wasn't able to compress the method Nico mentioned. Can anyone share an idea about how to write the same with less code or method?

Ready to review 🚀

distantnative:
> prevnext renamed to navigation

That doesn't make much sense to me. Sounds like that method would actually return the navigation (which rather `->prev()` and `->next()` separately do) but instead that method is only used for providing label/link for what you pass?

I feel like we are adding currently quite a lot of methods to the models that all only serve this one feature. Feels quite scattered and thus bloated at the moment. Can't we rather merge them into one method? Can't `->siblings()` and `->filter()` be merged to one at least?

afbora:
> That doesn't make much sense to me. Sounds like that method would actually return the navigation (which rather ->prev() and ->next() separately do) but instead that method is only used for providing label/link for what you pass?

Let's give it a more specific name then like `prevNextInfo()`, `navigationInfo()`, `tooltip()`, etc..

> Can't we rather merge them into one method? Can't ->siblings() and ->filter() be merged to one at least?

To reduce it to a single method, we will need an additional parameter for the prev or next collection and we need a method like below. Any better idea?

````php
protected function filterCollection(string $prevNext) 
{
    if ($prevNext === 'prev') {
        $collection = $this->model->prevAll(..);
    } else {
        $collection = $this->model->nextAll(..);
    }
}
````

or 

````php
protected function filterCollection(bool $isPrev = false) 
{
    if (isPrev === true) {
        $collection = $this->model->prevAll(..);
    } else {
        $collection = $this->model->nextAll(..);
    }
}
````

**Comment by bastianallgeier** [comment]
This is turning into the mostly discussed PR ever :) 

Ok, let's approach this from a Fiber response standpoint. We want a complete array with the navigation setup for a model. 

`$page->panel()->navigation()` should return everything we need. 

```
[
  'prev' => [
    'link' => '/something',
    'tooltip' => 'Something'
  ],
  'next' => [
    'link' => '/something',
    'tooltip' => 'Something'
  ]
]
```

Is that what you meant @distantnative?

**Comment by afbora** [comment]
> $page->panel()->navigation() should return everything we need.

`navigation()` method is belong to abstract `Model` class.
Should it be overridden as there will be no prev and next for the `Site` model?

**Comment by distantnative** [comment]
Yes. That's why I mean. I would find it weird to have `Model::filter()` that only relates to prevnext functionality

**Comment by bastianallgeier** [comment]
@afbora I think it would be fine if the navigation option for site would return ['prev' => false, 'next' => false]

**Comment by afbora** [comment]
### Summary

- Used `prevnext` name again *
- Rebased
- Still there are two additional internal method: [see comment](https://github.com/getkirby/kirby/pull/2779#issuecomment-862506665)

Actually, the function of the `prevnext()` method is not exactly the subject of this PR.

\* When I started creating the new `navigation()` method we were considering replacing `prevnext()`, the codes started to repeat and get confused. The first build @distantnative created is pretty simple and flexible 😇 I still have no negative arguments except for the name.

**Comment by distantnative** [comment]
I think maybe it's too hard to explain what I meant. I will make a suggestion tonight as a PR to this PR.

**Comment by distantnative** [comment]
How about:

- Former `Mode::prevnext()` is renamed to `Mode::toLink()` (better describing what it does)
- We introduce a new `Page::prevnext()` that merges all of @afbora's code for the navigation into a single method:

```php
public function prevnext(): array
{
    $page = $this->model;

    // create siblings collection based on
    // blueprint navigation
    $siblings = function (string $direction) use ($page) {
        $navigation = $page->blueprint()->navigation();
        $sortBy     = $navigation['sortBy'] ?? null;
        $status     = $navigation['status'] ?? null;

        // if status is defined in navigation,
        // all items in the collection are used
        // (drafts, listed and unlisted) otherwise
        // it depends on the status of the page
        $siblings = $status !== null ? $page->parentModel()->childrenAndDrafts() : $page->siblings();

        // sort the collection if custom sortBy
        // defined in navigation otherwise
        // default sorting will apply
        if ($sortBy !== null) {
            $siblings = $siblings->sort(...$siblings::sortArgs($sortBy));
        }

        $siblings = $page->{$direction . 'All'}($siblings);

        if (empty($navigation) === false) {
            $statuses  = (array)($navigation['status'] ?? $page->status());
            $templates = (array)($navigation['template'] ?? $page->intendedTemplate());

            // do not filter if template navigation is all
            if (in_array('all', $templates) === false) {
                $siblings = $siblings->filter('intendedTemplate', 'in', $templates);
            }

            // do not filter if status navigation is all
            if (in_array('all', $statuses) === false) {
                $siblings = $siblings->filter('status', 'in', $statuses);
            }
        } else {
            $siblings = $siblings
                ->filter('intendedTemplate', $page->intendedTemplate())
                ->filter('status', $page->status());
        }

        return $siblings->filter('isReadable', true);
    };

    return [
        'next' => function () use ($page, $siblings) {
            $next = $siblings('next')->first();
            return $next ? $next->panel()->toLink('title') : null;
        },
        'prev'   => function () use ($page, $siblings) {
            $prev = $siblings('prev')->last();
            return $prev ? $prev->panel()->toLink('title') : null;
        }
    ];
}
```
- And use that one to merge this array into the `Panel::props()` array


I am pushing so much for this, because I think a `Page::siblings()` or a `Page::filter()` sound really general but then you find out that they are super tightly tied to the blueprint `navigation` option.

afbora:
@bastianallgeier @distantnative I've implement the Nico's suggestion. What do you think?

IMO, although the new method was quite busy, it was nice to get rid of the excess. It will help us in the future if we can explain the method better.

bastianallgeier:
I like it a lot! It's definitely more compact and understandable that way.

config/api/models/Page.php
bastianallgeier: Is this fallback still correct after the last changes?

afbora: Ahh, yes i've to revert this changes 👍

distantnative (dismissed):

config/api/models/Page.php
```diff
@@ -74,7 +81,14 @@
...
             return $page->parents()->flip();
         },
+        /**
+         * @deprecated 3.6.0
+         * @todo Throw deprecated warning in 3.7.0
+         * @todo Remove in 3.8.0
+         * @codeCoverageIgnore
+         */
         'prev' => function (Page $page) {
+            deprecated('The `prev` field has been deprecated and will be removed in 3.7.0');
```
afbora: @distantnative You removed `deprecated()` method from `next` field. This line should be removed?

distantnative: Oh yes, good catch! I think we are in no rush to remove them. So to make it easier for users, I think it's better to only mark them as deprecated now, then start throwing warning in 3.7 and finally remove it in 3.8

afbora:
Do you want me to squash all commits?

distantnative:
Would probably be good to squash them.

bastianallgeier (approved):

bastianallgeier:
Wow, this was a tough one. Thanks for your patience @afbora!! I'm really happy about the result.

PR #2780, opened 2020-08-13
WIP: Add missing unit test to increase code coverage

afbora:
### WIP

I have added missing unit tests to increase code coverage and will continue to add. There are many missing tests, but I'm trying to complete what I could do.

- [x] Overall [%84.028 > %90.291]
- [x] Api [%72.06 > %90.82]
- [x] Cms [%84.68 > %88.92]
- [x] Database [%21.92 > %87.99]
- [x] Image [%76.75 > %92.36]

### Sepated PRs of Namespaces

- [x] Api #2782
- [x] Cms #2798 
- [x] Database #2795 
- [x] Image #2796
- [x] Other #2800 

## Ready?

- [x] Passing all unit tests


afbora:
I've done what I can do. I may have made a mistake in some tests. 

Btw, there are still missing tests. We can go through this PR or split it into PR. How you see fit.

src/Cms/Roles.php
```diff
@@ -46,7 +47,8 @@ public function canBeChanged()
      * roles that can be created by the
      * current user
      *
-     * @return self
+     * @return \Kirby\Toolkit\Collection
```
distantnative: I don't think these changes are quite right (also in other places). The method returns `$this`, so not just any collection, but `Kirby\Cms\Roles`. Or what was the reason for this change?

afbora: Because `$this->filter()` method in `canBeChanged()`, returns `\Kirby\Toolkit\Collection`, am I wrong?

afbora: There may be some wrong ones, but I made these changes by checking them

distantnative: I am not a 100% sure.
But I would have thought that the `filter` method (from Kirby\Toolkit\Collection) gets inherited way down to Kirby\Cms\Roles. And the filter method clones the object itself returns it. So within Kirby\Cms\Roles, I would have thought that it clones Kirby\Cms\Roles

afbora: Hmm, I understand better now. Yes you are right. My bad. I'll check all 👍

src/Database/Sql.php
```diff
@@ -797,8 +811,7 @@ public function update(array $params = []): array
      * @param string $table
      * @param string $column
      * @return bool
-     *
-     * @throws \Kirby\Exception\InvalidArgumentException If the column is invalid
+     * @throws \Kirby\Exception\InvalidArgumentException|\Throwable If the column is invalid
```
distantnative, line 814: Where does `Throwable` come from? I only see `InvalidArgumentException` in the method.

src/Database/Database.php
```diff
@@ -590,7 +599,7 @@ public function dropTable($table): bool
         }
 
         // update cache
-        $key = array_search($this->tableWhitelist, $table);
+        $key = array_search($table, $this->tableWhitelist ?? []);
```
distantnative, line 602: Is this a bugfix? If so, it really should have its own PR.
We need to keep actual changes to the code separate, so it's easier to review them and see if it's a breaking change.

afbora: An overlooked bugfix, yes.

src/Database/Sql.php
afbora: As there are interlinked methods, you can follow the sequence below for `validateColumn` method:

> public function validateColumn() > validateColumn() > validateTable() > query() > hit()

distantnative:
A general note: maybe we could split those up by package (e.g. Api, Cms...). I think that would already make it a lot easier to review this.

I tend myself to make such big PRs - but it's true that those are really hard to review.

src/Database/Database.php
afbora: @distantnative So, i can't test methods when i revert back.  I don't think there will be a break change because it's a critical error. It wasn't working already 😅 

> array_search() expects parameter 2 to be array, string given

afbora:
@lukasbestle suggested the same. I was expecting an answer from him. I thought it would be easier to separate after completing it.

afbora:
In general, if there is no critical error/mistake in PR or you don't have a general recommendation, I can start separating. We can do the detailed reviews after separate.

lukasbestle:
It's best to separate the PRs before we start reviewing. That's the whole point of separating. 😃

lukasbestle:
But maybe we should start with one package first and then you can create the other PRs after that first one is approved (so that 2-3 PRs are open simultaneously, but not more). Some general feedback might also affect changes in other packages.

afbora:
So let's start to  separate with the `Api` namespace, OK? Also, would this PR be open for backup?

lukasbestle:
👍

src/Database/Database.php
afbora: Separated https://github.com/getkirby/kirby/pull/2788

afbora:
Refactoring of four namespaces and deprecated methods for code coverage completed successfully with separated PRs. Coverage increased `84.028%` to ` 90.269%` 🚀

bastianallgeier:
This has been a huge bump in test coverage! Amazing work! The last few percent are so hard to complete and I really appreciate any help with this.

afbora:
Thank you, i did it with pleasure 👍 Is there any namespaces you think missing tests or should be coverage?

bastianallgeier:
Coveralls gives a great overview where we can still improve: https://coveralls.io/github/getkirby/kirby?branch=develop

I think a lot of code in the config folder is still not entirely tested or testable.

afbora:
I already chose four namespaces by looking there. Then I'll complete the rest in 3.4.4 release.

Some parts of the config folder are really difficult to test or untestable (unreasonable like aliases), so I'm thinking of leaving the config folder to you 😅

PR #2782, opened 2020-08-15
[Coverage] Api unit tests

afbora:
## Describe the PR

 `Api` namespace unit tests to increase code coverage and this PR separated from #2780

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -67,7 +67,7 @@ class Api
     /**
      * The current route
      *
-     * @var Route
+     * @var \Kirby\Http\Route
      */
     protected $route;
 
@@ -107,6 +107,7 @@ class Api
      * @param string $method
      * @param array $args
      * @return mixed
+     * @throws \Kirby\Exception\NotFoundException
      */
     public function __call(string $method, array $args = [])
     {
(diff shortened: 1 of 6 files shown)


src/Api/Api.php
```diff
@@ -710,14 +716,17 @@ public function responseForException($e): array
     /**
      * Upload helper method
      *
+     * move_uploaded_file() not working with unit test
+     * Added debug parameter for testing purposes as we did in the Email class
```
afbora, line 720: Have an opinion on this matter?

src/Api/Collection.php
```diff
@@ -19,12 +19,39 @@
  */
 class Collection
 {
+    /**
+     * @var Api
```
bastianallgeier: This should probably be \Kirby\Api\Api, right? There are a few more here, which probably need to have absolute namespaces. I.e. \Exception. Etc.

src/Api/Api.php
bastianallgeier: I like it!

afbora:
@bastianallgeier DocBlocks updated, i think it's ready.

bastianallgeier (approved):

PR #2785, opened 2020-08-18
Fix/progress bar styling

### LINKED ISSUE #2784: Progress bar styling inconsistent
**Describe the bug**  
The progress bar component is not styled in firefox. In Chrome and Safari I noticed that if the width of the progress indicator is less then it's height the border radius gets smaller and it overflows the bar.

**Screenshots**

Firefox:
Bildschirmfoto 2020-08-18 um 23 15 54

Chrome:
Bildschirmfoto 2020-08-18 um 23 15 59

**Kirby Version**  
3.4.2

qwerdee:
## Describe the PR

This PR fixes styling Issues of the progress bar component. Styles for firefox are added and overflowing the bar is now prevented.

## Related issues

- Fixes #2784 

## Ready?

- [ ] Passing all unit tests

### DIFF
--- a/panel/src/components/Misc/Progress.vue
+++ b/panel/src/components/Misc/Progress.vue
@@ -31,6 +31,8 @@ export default {
   width: 100%;
   height: 0.5rem;
   border-radius: 5rem;
+  background: $color-border;
+  border: none;
 }
 
 .k-progress::-webkit-progress-bar {
@@ -41,7 +43,13 @@ export default {
 }
 
 .k-progress::-webkit-progress-value {
-  border-radius: 20px;
+  border-radius: inherit;
+  background: $color-focus;
+  transition: width 0.3s;
+}
+
+.k-progress::-moz-progress-bar {
+  border-radius: inherit;
   background: $color-focus;
   transition: width 0.3s;
 }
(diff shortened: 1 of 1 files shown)


panel/src/components/Misc/Progress.vue
```diff
@@ -41,7 +44,13 @@ export default {
 }
 
 .k-progress::-webkit-progress-value {
-  border-radius: 20px;
+  border-radius: inherit;
+  background: $color-focus;
+  transition: width 0.3s;
+}
+
```
distantnative, line 51: Let’s keep it DRY and combine both rule sets

qwerdee: That doesn't work in this case since these are vendor specific selectors. In all cases one of them is not recognized, so the whole block is ignored

qwerdee: https://stackoverflow.com/questions/50580986/why-cant-i-combine-pseudo-element-selectors/50581522#50581522

distantnative: I see. Good to know!

bastianallgeier (approved):

PR #2794, opened 2020-08-31
Implement fuzzy/similar search

afbora:
## Describe the PR

Similar searches feature added to collections 🎉 

Since the [similar_text()](https://www.php.net/manual/en/function.similar-text.php) method doesn't support multibytes, I used a [different method](https://github.com/antalaron/mb-similar-text). 

Similar search is disabled by default. For the panel UI, I added a button that can be changed instantly and easily.

Feel free to close the PR if a different implementation is being considered 👍 

**Screencast from Panel**

fuzzy

## Related issues

- Closes https://github.com/getkirby/ideas/issues/140

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/api/routes/files.php
+++ b/config/api/routes/files.php
@@ -114,7 +114,9 @@ return [
                 ->files();
 
             if ($this->requestMethod() === 'GET') {
-                return $files->search($this->requestQuery('q'));
+                return $files->search($this->requestQuery('q'), [
+                    'similar' => filter_var($this->requestQuery('similar'), FILTER_VALIDATE_BOOLEAN)
+                ]);
             } else {
                 return $files->query($this->requestBody());
             }
--- a/config/api/routes/site.php
+++ b/config/api/routes/site.php
@@ -70,7 +70,9 @@ return [
                 ->filterBy('isReadable', true);
 
             if ($this->requestMethod() === 'GET') {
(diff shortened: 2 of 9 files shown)


lukasbestle (commented):
I haven't tested this yet, but the general approach looks great! ❤️

However we cannot copy the code of the library as this would be a copyright violation. We either need to include the [full license text](https://github.com/antalaron/mb-similar-text/blob/master/LICENSE) in a code comment or require the library directly with Composer.

afbora:
@lukasbestle You are right! I added the author name, copyright and license information.

bastianallgeier:
Hey @afbora, 

I'm really sorry, that your PRs get stuck while we are working on other features. Fuzzy search is awesome! I have a few things though that I'd like to change. 

I would call the method `Str::similarity()` 

Would it be possible to split this into two tasks. I think we could instantly introduce your changes in PHP. But I'm not convinced about the implementation in the panel yet. Maybe we should start thinking about a general dropdown with search options instead of just one button for the similar search. Could you extract the PHP changes into its own PR? I'd love to merge that for 3.5 and then we can work on the interface in the next step.

afbora:
@bastianallgeier I have extracted the PHP side to #2936 Actually you can close the PR if you want. Because the changes you make will be quite different from mine 😅

PR #2815, opened 2020-09-09
Support query templates for fields default values

### LINKED ISSUE #2814: Query templates do not work for default values in non-textarea fields
**Describe the bug**  
According to the[ release notes of Kirby 3.2.5](https://github.com/getkirby/kirby/releases/tag/3.2.5), it should be possible to use query templates in blueprints in labels, default etc. However, turns out that trying this doesn't really work, at least not in current versions, see this issue in the forum for details: https://forum.getkirby.com/t/query-number-field-of-parent/19303/4

**Expected behavior**  
Query templates should work as promised ;)

**Screenshots**  

**Kirby Version**  
I actually tested with 3.3.6, but should be the same in recent versions.

afbora:
## Describe the PR

Added `Kirby\Form\Field::toString()` method for fields to support query templates. Fields can do their own operations after using this method first.

### Enhancements

- Added support for query template to default prop/value of all fields

### Fixes

- Fixed query templates for default values in non-textarea fields

## Breaking changes

None

## Related issues/ideas

- Fixes #2814 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/fields/checkboxes.php
+++ b/config/fields/checkboxes.php
@@ -24,7 +24,7 @@ return [
          * Default value for the field, which will be used when a page/file/user is created
          */
         'default' => function ($default = null) {
-            return Str::split($default, ',');
+            return $default;
         },
         /**
          * Maximum number of checked boxes
@@ -44,7 +44,9 @@ return [
     ],
     'computed' => [
         'default' => function () {
-            return $this->sanitizeOptions($this->default);
+            $default = Str::split($this->toString($this->default), ',');
+
+            return $this->sanitizeOptions($default);
         },
         'value' => function () {
             return $this->sanitizeOptions($this->value);
--- a/config/fields/checkboxes.php
+++ b/config/fields/checkboxes.php
@@ -21,12 +21,6 @@ return [
             return $columns;
         },
         /**
-         * Default value for the field, which will be used when a page/file/user is created
-         */
-        'default' => function ($default = null) {
-            return $default;
-        },
-        /**
          * Maximum number of checked boxes
          */
         'max' => function (int $max = null) {
--- a/config/fields/checkboxes.php
+++ b/config/fields/checkboxes.php
@@ -31,10 +31,7 @@ return [
          */
         'min' => function (int $min = null) {
             return $min;
-        },
-        'value' => function ($value = null) {
-            return Str::split($value, ',');
-        },
+        }
     ],
     'computed' => [
         'default' => function () {
@@ -43,7 +40,7 @@ return [
             return $this->sanitizeOptions($default);
         },
         'value' => function () {
-            return $this->sanitizeOptions($this->value);
+            return $this->sanitizeOptions(Str::split($this->value, ','));
         },
     ],
     'save' => function ($value): string {
--- a/src/Form/Field.php
+++ b/src/Form/Field.php
@@ -213,47 +213,43 @@ class Field extends Component
             'computed' => [
                 'after' => function () {
                     /** @var \Kirby\Form\Field $this */
-                    if ($this->after !== null) {
-                        return $this->model()->toString($this->after);
-                    }
+                    return $this->toString($this->after);
                 },
                 'before' => function () {
                     /** @var \Kirby\Form\Field $this */
-                    if ($this->before !== null) {
-                        return $this->model()->toString($this->before);
-                    }
+                    return $this->toString($this->before);
                 },
                 'default' => function () {
                     /** @var \Kirby\Form\Field $this */
-                    if ($this->default === null) {
-                        return;
-                    }
-
-                    if (is_string($this->default) === false) {
-                        return $this->default;
-                    }
-
-                    return $this->model()->toString($this->default);
+                    return $this->toString($this->default);
                 },
                 'help' => function () {
                     /** @var \Kirby\Form\Field $this */
-                    if ($this->help) {
-                        $help = $this->model()->toString($this->help);
-                        $help = $this->kirby()->kirbytext($help);
-                        return $help;
+                    if ($text = $this->toString($this->help)) {
+                        return $this->kirby()->kirbytext($text);
                     }
                 },
                 'label' => function () {
                     /** @var \Kirby\Form\Field $this */
-                    if ($this->label !== null) {
-                        return $this->model()->toString($this->label);
-                    }
+                    return $this->toString($this->label);
                 },
                 'placeholder' => function () {
                     /** @var \Kirby\Form\Field $this */
-                    if ($this->placeholder !== null) {
-                        return $this->model()->toString($this->placeholder);
+                    return $this->toString($this->placeholder);
+                }
+            ],
+            'methods' => [
+                'toString' => function ($value = null) {
+                    if ($value === null) {
+                        return;
                     }
+
+                    if (is_string($value) === false) {
+                        return $value;
+                    }
+
+                    /** @var \Kirby\Form\Field $this */
+                    return $this->model()->toString($value);
                 }
             ]
         ];
--- a/src/Form/Field.php
+++ b/src/Form/Field.php
@@ -237,20 +237,6 @@ class Field extends Component
                     /** @var \Kirby\Form\Field $this */
                     return $this->toString($this->placeholder);
                 }
-            ],
-            'methods' => [
-                'toString' => function ($value = null) {
-                    if ($value === null) {
-                        return;
-                    }
-
-                    if (is_string($value) === false) {
-                        return $value;
-                    }
-
-                    /** @var \Kirby\Form\Field $this */
-                    return $this->model()->toString($value);
-                }
             ]
         ];
     }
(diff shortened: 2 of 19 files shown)


afbora:
I've reworked it and it's ready to be review 🚀

config/fields/checkboxes.php
```diff
@@ -24,7 +24,7 @@
          * Default value for the field, which will be used when a page/file/user is created
          */
         'default' => function ($default = null) {
-            return Str::split($default, ',');
+            return $default;
```
lukasbestle: I think we can remove the entire `default` prop here as the implementation is the same as the default in the `Form\Field` class.

src/Form/Field.php
```diff
@@ -213,47 +213,43 @@ public static function defaults(): array
...
                 },
                 'placeholder' => function () {
                     /** @var \Kirby\Form\Field $this */
-                    if ($this->placeholder !== null) {
-                        return $this->model()->toString($this->placeholder);
+                    return $this->toString($this->placeholder);
+                }
+            ],
+            'methods' => [
+                'toString' => function ($value = null) {
```
lukasbestle: Can't we move this method to a real instance method of the `Field` class?

lukasbestle (commented):
I looked through all fields and noticed the `placeholder` computed in the `select` field. Maybe that one should have support for query templates as well?

There are also a few others like the `display` format for date and time, but I'd say these don't really need query templates.

src/Form/Field.php
afbora: We can move, but will this method;

- works for the field value?

````php
public function toString()
{
    $value = $field->value;
    //
}
````

- or as an helper method for passing values?


````php
public function toString($value = null)
{
    //
}
````

It's a little confusing 🤔

**Comment by afbora** [comment]
> There are also a few others like the display format for date and time, but I'd say these don't really need query templates.

Actually I agree. But I think it can be used, albeit very rarely. At least we have a standard use for all fields?

File: tests/Form/Fields/PagesFieldTest.php [hunk 3]
```diff
@@ -162,4 +162,29 @@ public function testRequiredValid()
...
+                        ]
+                    ]
+                ]
+            ]
+        ]);
+
+        $app->impersonate('kirby');
+        $response = $app->api()->call('pages/a/fields/foo');
+
+        $this->assertCount(2, $response['data']);
```
afbora: @lukasbestle Its (files/pages/users field api tests) passed when run one by one, but fails when running all tests. Do you have an idea?

afbora: I've fixed with separate model templates.

afbora:
I've no idea about failing `Kirby\Toolkit\LocaleTest::testExport` unit test 🤷‍♂️

src/Form/Field.php
lukasbestle: As far as I understand it, the `toString()` method as it's implemented in this PR is only meant as an internal helper method for fields to use in other methods like `default` etc.

So IMO it should be a method with the following signature inspired by `ModelWithContent::toString()`:

```php
protected function toString(string $template, array $data = [], string $fallback = ''): string
```

This can then be used with any kind of property.

lukasbestle:
I've rebased the branch against `develop`. Let's see where the test failure is coming from (weirdly it's working just fine locally...).

lukasbestle:
Very weird, I have no idea why this PR caused that test to fail. But now I fixed it forever. :)

src/Form/Field.php
afbora: @lukasbestle Before push the commit, what do you think about following implementation for `Field::toString()`?

````php
/**
 * String template builder
 *
 * @param string|null $template
 * @param array $data
 * @param string $fallback Fallback for tokens in the template that cannot be replaced
 * @return string|null
 */
protected function toString(string $template = null, array $data = [], string $fallback = ''): ?string
{
    if ($template === null) {
        return null;
    }

    if (is_string($template) === false) {
        return $template;
    }

    return $this->model()->toString($template, $data, $fallback);
}
````

Comment (lukasbestle) [reply src/Form/Field.php]: A few comments:

- What would be the use-case for passing `null` as the `$template`? Or is that just for compatibility to avoid fatal errors if `null` is passed through? If so: Can we safely assume that every use of this method can deal with a returned `null` value?
- If we decide to allow `null`, the argument should be hinted `?string $template = null`
- If the first argument is hinted as `?string $template = null`, it can never be anything else than `null` or `string`, so the second `if` doesn't do anything.

Comment (afbora) [reply src/Form/Field.php]: - If the `toString()` method doesn't return `null` (because the value of the undefined prop is `null`), the props of all fields will be empty string (`''`) instead of `null` (For ex: `before`, `after`, `default`, etc..). This is undesirable.
- I agree with the other 2 comments, but first we need to find a solution to the first comment about passing/returning `null`.

Comment (lukasbestle) [reply src/Form/Field.php]: > If the `toString()` method doesn't return `null` (because the value of the undefined prop is `null`), the props of all fields will be empty string (`''`) instead of `null` (For ex: `before`, `after`, `default`, etc..). This is undesirable.

Very good point, I fully agree!

Comment (lukasbestle) [reply src/Form/Field.php]: I think there should be a code comment that explains the reasoning:

```php
if ($template === null) {
    // if the input template prop is not defined, the output
    // of the prop should also be `null` and not an empty string
    return null;
}
```

src/Form/Field.php
```diff
@@ -213,47 +213,43 @@ public static function defaults(): array
...
+                    if ($value === null) {
+                        return;
                     }
+
+                    if (is_string($value) === false) {
+                        return $value;
+                    }
```
afbora: @lukasbestle I think we have an issue here too. For non-string values, right? Will it solve if we just remove the return type of method?

lukasbestle: Can you point me to a use-case where we need this behavior? If possible, I'd like to avoid a `toString()` method that doesn't always return strings. That would be quite confusing.

afbora: For example, `default` prop can return `array` for structure field or `bool` for toggle field.

lukasbestle: Ah, right!

Hm, to be honest I think these cases should be handled by the affected fields then. So they should only call `toString()` if their prop value is already a string. Then we could also make the `$template` argument accept a string only.

But I think this is subjective. @bastianallgeier @distantnative What do you think? Also see the [open discussion above](https://github.com/getkirby/kirby/pull/2815#discussion_r630119036) for context.

afbora: Pinging @bastianallgeier @distantnative to review

afbora:
I applied the points we discussed with Lukas. Check out the parts we discussed with Lukas and let us know if there are any points we missed 👍 

In summary, the default props of all fields have been overhauled and all are connected to the new `Kirby\Form\Field::toString()` method. Thus, the default value of all fields now supports query templates.

distantnative:
Currently, the PR removes the `default` prop from a lot of fields. This will remove it from documentation as well, I fear. So the default props should rather not be removed.

afbora:
@distantnative Are we sure? I think it will get it from the parent that the `Kirby\Form\Field` class. Don't we just override the props in their own config files?

lukasbestle:
@distantnative Ping regarding Ahmet's question.

distantnative:
Finally could check: @afbora you are right, the `Kirby\Form\Field` definition will still show up in the docs. All good.

src/Form/Field.php
distantnative: I thinking solving this in the fields themselves would require us to define the `default` prop in all these fields explicitly, while @afbora tried to rather use a generally defined one from `Kirby\Form\Field`. To me both would be fine, handling this in the fields, but I also wouldn't mind if `toString()` includes a check... could be a bit DRYer.

afbora:
Any open questions left?

pReya:
@bastianallgeier Any explanation why this feature was cancelled? Too complex?

afbora:
Actually no canceled, just delayed. As it affects all fields, we will consider it again in the coming days/releases.

texnixe:
@bastianallgeier @afbora Is there any news regarding this feature? Was it abandoned or just forgotten after this was closed?

lukasbestle:
It's in the 3.7.x milestone, which means we want to pick it up again. But it turned out to be quite complex, so it's nothing we can do without a lot of focus. Might also make sense to wait until the blueprint refactoring is done.

afbora:
I didn't take care again because of this PR and issue related refactoring of panel forms: https://github.com/getkirby/backlog/issues/76

ovenum:
Is this still being worked on? It’s causing some confusion [from time](https://forum.getkirby.com/t/using-page-function-for-the-default-value-of-a-text-field/29186/4) to time if query templates are supported for a fields default value or not : )

The [Kirby Guide article](https://getkirby.com/docs/guide/blueprints/query-language) describing also does not clarify that queries can or cannot be used for field default values. 

The link to the blocking issue in /backlog is a 404.

bastianallgeier:
@ovenum this will finally be fixed in v5-alpha.5

PR #2818, opened 2020-09-11
Check if `dump` function exists before defining

### LINKED ISSUE #2816: Cannot redeclare dump()
Using [Psy Shell](https://psysh.org/) with Kirby is proving super useful for admin tasks, migrations and general debugging.

However because psy defines it's own function named `dump` (used for inspecting variables/objects in the REPL), we run into an error when trying to bootstrap kirby inside the shell.

```
Psy Shell v0.10.4 (PHP 7.4.10 — cli) by Justin Hileman
>>> require 'kirby/bootstrap.php'
PHP Fatal error:  Cannot redeclare dump() (previously declared in phar:///usr/bin/psysh/vendor/autoload.php:21) in <snipped>/kirby/config/helpers.php on line 147
```

I can get around this by commenting out the `dump` function declaration in `/kirby/config/helpers.php` temporarily. But I wonder if this could be solved by wrapping that declaration in `if(!function_exists('dump')) {}`.

Is this something worth changing? `dump` is the only globally defined function that causes such a conflict.

mattlenz:
## Describe the PR

Some other PHP utilities sometimes to define their own `dump` util function globally. Kirby can't be loaded in these cases because it attempts to redefine this function. This PR adds a check to ensure the function is not already defined.

## Related issues

- Fixes #2816


bastianallgeier:
I merged this manually! Thank you for the PR 👏

PR #2819, opened 2020-09-11
Merge default value for structure field

### LINKED ISSUE #2790: Translations should not remove non-translatable fields in structures
This issue is a continuation of #1571. @afbora [asked me](https://github.com/getkirby/kirby/issues/1571#issuecomment-680003303) to open a new issue.

**Describe the bug**  
When you save a translation, all non-translatable fields should be skipped. This was fixed in #1571. But now there's a problem in structures - there, non-translatable fields should **not** be skipped because it leads to unexpected behavior.

**To Reproduce**
I have the following `site.yml`:

```yml
fields:
  text:
    type: text
  textConst:
    type: text
    translate: false
  test:
    type: structure
    fields:
      text:
        type: text
      textConst:
        type: text
        translate: false
```

...and the following `site.en.txt`:

```
Text: Foo

----

Textconst: Bar

----

Test:

- 
  text: foo
  textconst: bar
```

Then, I update the Bulgarian translation with the values `Text = FooBG` and `Test/text = fooBG`. That's the resulting `site.bg.txt`:

```
Text: FooBG

----

Test:

- 
  text: fooBG
```

As you can see, the `Textconst` field is missing, and it should be - this way Kirby will use the default translation value. However, the `textconst` field _in the structure_ is also missing. This means that I get the value `null` for it in the template:

```php
var_dump($site->test()->toStructure()->first()->text()->value()); // "fooBG"
var_dump($site->test()->toStructure()->first()->textConst()->value()); // null
```

At the same time, that value is grayed out in the panel and you can't fill it:

image

What's interesting is that even if I add `textconst` in `site.bg.txt` manually with my text editor, refresh the panel, and update the translation, it gets removed. (1)

**Expected behavior**  
I see two solutions for that:

1. In the `toStructure()` call, Kirby also loads the default translation values and does `array_replace_recursive()` with the current translation values
2. Kirby should leave values in structures as they are, so that (1) doesn't happen and plugins can handle the synchronization of values across translations

I'm currently working on a plugin that traverses the page blueprints and synchronizes structures according to them. For the most part, it works great, but it fails when it tries to copy values from the default translation - they are lost in the `$page->update()` call. If you implement the second solution I proposed, my plugin would be able to solve the problem.

**Kirby Version**  
3.4.2

afbora:
## Describe the PR

Since the structure field doesn't keep the data of the untranslatable fields, now it merges structure field values with default language field data in multi-language applications like we doing in `ContentTranslation` class.

## Related issues

- Fixes #2790 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/fields/structure.php
+++ b/config/fields/structure.php
@@ -1,7 +1,7 @@
 <?php
 
 use Kirby\Cms\Form;
-use Kirby\Data\Data;
+use Kirby\Cms\Structure;
 use Kirby\Toolkit\I18n;
 
 return [
@@ -145,9 +145,9 @@ return [
     ],
     'methods' => [
         'rows' => function ($value) {
-            $rows  = Data::decode($value, 'yaml');
-            $value = [];
+            $rows = Structure::toData($this->name(), $value, $this->model);
 
+            $value = [];
             foreach ($rows as $index => $row) {
                 if (is_array($row) === false) {
                     continue;
--- a/config/methods.php
+++ b/config/methods.php
@@ -250,7 +250,8 @@ return function (App $app) {
          */
         'toStructure' => function (Field $field) {
             try {
-                return new Structure(Data::decode($field->value, 'yaml'), $field->parent());
+                $data = Structure::toData($field->key(), $field->value, $field->parent());
+                return new Structure($data, $field->parent());
             } catch (Exception $e) {
                 if ($field->parent() === null) {
                     $message = 'Invalid structure data for "' . $field->key() . '" field';
--- a/src/Cms/Structure.php
+++ b/src/Cms/Structure.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Cms;
 
+use Kirby\Data\Data;
 use Kirby\Exception\InvalidArgumentException;
 
 /**
@@ -60,5 +61,36 @@ class Structure extends Collection
         }
 
         return parent::__set($object->id(), $object);
+    }
+
+    /**
+     * Converts structure value to array
+     * and merges default language data for all languages
+     * for non-translatable fields
+     *
+     * @param string $fieldName
+     * @param mixed $value
+     * @param object|null $parent
+     * @return array
+     */
+    public static function toData(string $fieldName, $value, $parent = null): array
+    {
+        $data  = Data::decode($value, 'yaml');
+        $kirby = is_a($parent, '\Kirby\Cms\Model') === true ? $parent->kirby() : App::instance();
+
+        // merge with the default content
+        if ($parent !== null && $kirby->multilang() === true) {
+            $language        = $kirby->language();
+            $defaultLanguage = $kirby->defaultLanguage();
+
+            if ($language->code() !== $defaultLanguage->code()) {
+                if ($content = $parent->content($defaultLanguage->code())->get($fieldName)) {
+                    $default = Data::decode($content->value(), 'yaml');
+                    $data    = array_replace_recursive($default, $data);
+                }
+            }
+        }
+
+        return $data;
     }
 }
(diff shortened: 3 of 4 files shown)


bastianallgeier:
I'm kind of "scared" of this one to be honest. I don't fully understand it from looking at it without really diving deeper. Do you think we could run into any unwanted side effects? I.e. performance issues or unexpected data?

afbora:
We can get the opinions of @texnixe @distantnative @lukasbestle 

Since this is not working properly, it will no longer return the expected/old data, but the data that should be. 
Therefore, we can say that it contains breaking change. There is no problem with 3.5.x or waiting until 3.6.

I'm not sure there will be a performance problem, @lukasbestle can guide us on this.

lukasbestle:
Can we safely assume that the structure items are in sync between languages? There could be use-cases with completely different structure items for each translation. I think the solution from this PR will then mix up the content. On the other hand, using `translate: false` could be interpreted as saying "my use-case has the structure items in sync between languages". But it wouldn't be obvious behavior.

I think there should be some way to link up translations of the same structure item, for example with UUIDs. This would solve so many issues at once. But it should be configurable:

- In one mode, adding or deleting a structure item in one translation would modify it in all translations. This mode would use the UUIDs and could be the default.
- In the other mode, there should be no UUIDs and each translation would be completely separate. This would also mean that the structure items of the default language would *not* be automatically copied to the other languages.

I'm not 100 % sure about the implications of such a behavior though. Especially as it would be a breaking-change.

afbora:
> Can we safely assume that the structure items are in sync between languages?

For this PR, unfortunately no. If the content in the two languages are different, the contents will be mixed as your said.

It makes sense to add mode to the structure field. Because as you said, there may be different uses. The two language contents can be completely different or synchronized. For example: `sync: true` as default and `sync: false`

By the way, the same problem exists in the blocks field.

When this PR doesn't provide a complete solution, we can close.

distantnative:
I think we had similar discussions before. From them:

- `translate: true`: structure field content is completed independent between languages
- `translate: "sync"`: some UUID is used to sync adding, removing, resorting of structure elements, however the other field contents can be translated
- `translate: false`: all languages use structure field content from default language, not translation

lukasbestle:
@distantnative I think the sync mode should be the new default in the long run. It's probably hard to implement this in a backwards-compatible manner, but the value `translate: "sync"` is weird for a default value when there is also `true`, which does something else.

distantnative:
@lukasbestle I agree from the perspective of what makes the most sense to be the default. But indeed it will be hard to migrate to this without breaking a lot. And I understand that `translate: "sync"` as default is a bit weird if there is also `true` and `false` - I mainly chose this naming as example to avoid creating another option `sync`, I think we need to make sur not to add too many options, it just gets confusing.

lukasbestle:
Yep, 💯. It's quite hard to make the configuration clean but also backwards-compatible, I think this needs some more thoughts.

afbora:
It seems difficult to smooth transition without additional options. How would it be like this?

v3.6: default mode is `sync: false`

````yaml
translate: true
sync: true # default `false`, manually enable
````

v3.7: default mode is `sync: true`

````yaml
translate: true
sync: false # default `true`, manually disable
````

lukasbestle:
I don't have a better idea at the moment. First feeling: It would move the breaking-change to 3.7. It would give users time to explicitly define `false` for fields they don't want synced, but what if you forget a field? It would then break with 3.7.

afbora:
Yes, it will break if `sync` is to default and is forgotten.
Anyway, let's think about it for a while, as you said.

distantnative:
I really would like to avoid another option. Especially when it's tied to the existing option anyways.

And I don't think a new option helps us in regard to the tricky question what should be default. 

But I agree, I think we should first implement the functionality as optional (leave the current `translate: true` as default as it is. And then down the road we can also see how adoption plays out and contemplate if/how to make the new functionality default.

Maybe we could create a `compat` option with suboptimal that easily allow turning back the old behavior. Do if a user wants to stick with the old behavior, it's an easy one-line way to do so.

PR #2826, opened 2020-09-15
[Coverage] Form unit tests

afbora:
## Describe the PR
Missing unit tests were added for the `\Kirby\Form` namespace and doc blocks were updated. `\Kirby\Form` namespace coverage increased from `92.08%` to `97.98%`

There may be incomplete or incorrect descriptions in the doc blocks. I'll be glad if you take a look.

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Form/OptionsApi.php
+++ b/src/Form/OptionsApi.php
@@ -25,34 +25,78 @@ class OptionsApi
 {
     use Properties;
 
+    /**
+     * @var
+     */
     protected $data;
+
+    /**
+     * @var
+     */
     protected $fetch;
+
+    /**
+     * @var
+     */
     protected $options;
+
+    /**
+     * @var string
+     */
     protected $text = '{{ item.value }}';
+
+    /**
+     * @var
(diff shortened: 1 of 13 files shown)


src/Form/OptionsApi.php
```diff
@@ -103,51 +147,94 @@ public function options(): array
...
 
+    /**
+     * @param $options
+     * @return self
+     */
+    protected function setOptions($options = null)
+    {
+        $this->options = $options;
+        return $this;
+    }
```
afbora, line 178: Your attention! Added new method. Because of following line:
https://github.com/getkirby/kirby/blob/3.4.3/src/Form/OptionsApi.php#L58-L60

`$this->options` are not set anywhere. I did the same for the `OptionsQuery` class.

bastianallgeier (approved):
Wow, this is again a massive update! GREAT WORK! We can still add more descriptions later. You already documented the most important methods very well.

PR #2833, opened 2020-09-17
Adds info text for users dialog

### LINKED ISSUE #2832: Users blueprint field main text attribute not working in the modal add dialog
When you add a `users`-field to your blueprint, you can use the `text:`-attribute to configure what the main text for each item should like like. For example, you can say `text: {{ user.name }} {{ user.email }}` to show both values, name and email, as the item's text. So far, so good. 

What is not working: When you click the Add-Link and the modal dialog appears to select an user item, it doesn't show you the item text as specified by the `text:`-attribute. In fact, the `text:`-attribute is ignored, and just the user's name is displayed. This is not consistent and from my point of view, a bug. 

In my case i have a site where there are 1.000+ users and we have the first users with identical (common) names. I need to use the `text:`-attribute to be able to see who is who.

Example of the Add dialog:

image

Please, make the add dialog use the same setting from the `text:`-attribute.

Trying with the `info:`-attribute is the same problem. In the main view `text:` and `info:` is working, but both are ignored in the add-view.

afbora:
## Describe the PR

Adds `info` text for users dialog and visible only not equal with `text` attribute.

## Related issues

- Fixes #2832 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/panel/src/components/Dialogs/UsersDialog.vue
+++ b/panel/src/components/Dialogs/UsersDialog.vue
@@ -28,7 +28,8 @@
           <k-list-item
             v-for="user in models"
             :key="user.email"
-            :text="user.username"
+            :text="user.text"
+            :info="user.info !== user.text ? user.info : null"
             :image="user.image"
             :icon="user.icon"
             @click="toggle(user)"


texnixe:
@afbora Would it make sense to make users, pages and files field dialogs work in the same way? Info is not show there either, if I'm not mistaken.

The dialog is probably a bit to narrow to display additional information, so actually using the overwritten text property would maybe make sense?

afbora:
I understand but `PagesDialog` and `FilesDialog` already have `info` property 🤔 

**PagesDialog** 
https://github.com/getkirby/kirby/blob/3.4.3/panel/src/components/Dialogs/PagesDialog.vue#L39-L40

**FilesDialog**
https://github.com/getkirby/kirby/blob/3.4.3/panel/src/components/Dialogs/FilesDialog.vue#L32-L33

texnixe:
Oh, I must have overlooked that although I can't see it show up in the dialog. Maybe I'm making a mistake where to define this. Sorry.

afbora:
Also i realized that dialog more wider:

screen-capture-282-Exploring-the-universe---Mægazine-localhost

bastianallgeier (approved):

PR #2838, opened 2020-09-21
[Coverage] CMS namespace step two

afbora:
## Describe the PR

Some of the missing unit tests in the CMS namespace have been completed. **I mentioned 2 critical updates in the relevant lines.**

## Ready?

- [x] Passing all unit tests


src/Cms/AppErrors.php
```diff
@@ -111,10 +111,14 @@ protected function handleJsonErrors(): void
...
                 $details  = $exception->getDetails();
-            } else {
+            } elseif (is_a($exception, '\Exception') === true) {
                 $httpCode = 500;
                 $code     = $exception->getCode();
                 $details  = null;
+            } else {
+                $httpCode = 500;
+                $code     = 500;
+                $details  = null;
```
afbora: It was giving an error in tests without an `Exception` instance. I made an improvement. I want you to check it out.

src/Cms/Media.php
```diff
@@ -100,7 +100,7 @@ public static function thumb($model, string $hash, string $filename)
             $root = $kirby->root('media') . '/assets/' . $model . '/' . $hash;
         } else {
             // model files
-            $root = $model->mediaRoot() . '/' . $hash;
+            $root = dirname($model->mediaRoot());
```
afbora, line 103: `mediaRoot` already contains both the filename and the hash.

**Current Path**
    /media/pages/photography/animals/22b8cf85b7-1600414901/peter-fox-38x38.jpg/22b8cf85b7-1600414901/.jobs/peter-fox-38x38.jpg.json

**Improved Path**
    /media/pages/photography/animals/22b8cf85b7-1600414901/.jobs/peter-fox-38x38.jpg.json

afbora:
I noticed it while writing the tests. Control should be `false`, right?

https://github.com/getkirby/kirby/blob/3.4.3/src/Cms/UserPermissions.php#L52-L55

Like that:

````php
if ($this->model->isAdmin() === false) {
   return false;
}
````

File: tests/Cms/Models/ModelWithContentTest.php [hunk 3]
```diff
@@ -38,8 +38,81 @@ public function root(): ?string
...
+                ]
+            ]
+        ]);
+
+        return [
+            [$app->site()],
+            [$app->page('foo')],
+            // [$app->site()->files()->first()], // not testable yet but should be
+            // [$app->user('test@getkirby.com')] // not testable yet but should be
+        ];
```
afbora: I couldn't write unit test for `File` and `User` model 😞

src/Cms/Media.php
bastianallgeier: Are you sure about this? I checked the mediaRoot method for the Page class again for example and it does not include the hash. So it would be wrong for page files for example.

tests/Cms/Models/ModelWithContentTest.php
bastianallgeier: What's the problem with those files?

afbora: You will see better the error if you remove the comments and run the test.

src/Cms/AppErrors.php
bastianallgeier: I've also caught all Throwables here

bastianallgeier:
> I noticed it while writing the tests. Control should be `false`, right?
> 
> https://github.com/getkirby/kirby/blob/3.4.3/src/Cms/UserPermissions.php#L52-L55
> 
> Like that:
> 
> ```
> if ($this->model->isAdmin() === false) {
>    return false;
> }
> ```

Yes, you are right 🙈

tests/Cms/Models/ModelWithContentTest.php
bastianallgeier: Ugh, I had a look at it, but honestly I have no clue what's going on here.

bastianallgeier: Ha, I solved it. You've setup pages sections for the test and they don't have a parent option. But pages sections in file or user views have to have a parent to be able to fetch pages correctly. This is what caused the error. I will commit a fix in a second.

src/Cms/Media.php
afbora: Hash is included for files, check out please:

https://github.com/getkirby/kirby/blob/3.4.3/src/Cms/File.php#L360-L363

afbora: Job file path for files like following, is that normal?

    /tests/Cms/Media/fixtures/MediaTest/media/site/c87dff63b0-1600766405/test.jpg/c87dff63b0-1600766405/.jobs/test.jpg.json

afbora: @bastianallgeier What do you think about that?

bastianallgeier: I think you might be right. The weird thing is that I wasn't able to find any place where we use Media::thumb with a File model as first parameter. Did you find a place for that?

afbora: I could not find it either. It doesn't need hash in the path of other objects but `File` do. I guess that's why it contains hashes.

How about if we write it like this for `Media::thumb`? Or if you have a better idea?

````php
if (is_string($model) === true) {
    // assets
    $root = $kirby->root('media') . '/assets/' . $model . '/' . $hash;
} elseif (is_a($model, '\Kirby\Cms\File')) {
    // model files for file model that already included hash
    $root = dirname($model->mediaRoot());
} else {
    // model files
    $root = $model->mediaRoot() . '/' . $hash;
}
````

bastianallgeier: I like that a lot! Could you make the change?

afbora: Great! Yes, i'll.

afbora:
@bastianallgeier It's ready 🚀

bastianallgeier (approved):

bastianallgeier:
Awesome job!

PR #2840, opened 2020-09-23
Adding various image formats

S1SYPHOS:
## Describe the PR
This PR adds various image formats ~~to method `isResizable` in `FileFoundation`, their classification as image~~ to class `F` and their respective MIME types to class `Mime`, that is

- `.avif` ([Wiki](https://en.wikipedia.org/wiki/AV1#AV1_Image_File_Format_(AVIF)) & [Abstract](https://aomediacodec.github.io/av1-avif)),
- `.j2k` & `.jp2` ([Wiki](https://en.wikipedia.org/wiki/JPEG_2000) & [Abstract](https://tools.ietf.org/html/rfc3745))

## Related issues
When working with eg `avif`, a next-gen image format, Kirby won't resize it because the internal check doesn't recognize it as 'resizable'.

Furthermore, it is unclear why Kirby, by itself not involved in the resize process (GD & Imagick are), deems some formats `resizable`, hardcoding the decision. For now, there's no way to hook into that.

IMHO, removing hardcoded resize definitions altogether might be something to consider. Otherwise, we should at least be able to extend it.

For more information, see [this forum thread](https://forum.getkirby.com/t/extending-trait-filefoundation-isresizable/19425/5).

## Ready?
- [x] Passing all unit tests


S1SYPHOS:
Sadly, the underlying PHP function `getimagesize` doesn't recognize next-gen image formats like `avif`, so [this line](https://github.com/getkirby/kirby/blob/master/src/Image/Image.php#L90) in `Kirby\Image` may remain unchanged for now, as the MIME check ultimately prevents invalid formats being passed to said function ..

bastianallgeier:
The problem is unfortunately way more complex than to change the resizable list. Imagemagick and GDLib don't necessarily support those formats as well. It's tricky. But I agree that we need a better way to extend such image types. 

As a first step it would be great to merge your PR, but only the changes to the F and Mime classes. Would you mind changing that? 

It would be even better if you could post something here in addition: https://feedback.getkirby.com to keep track of the idea to have a plugin API for new image types.

S1SYPHOS:
Thanks for your feedback. I'm totally okay changing my PR - but only if you keep up your great work :fox_face: 

That being said, determining whether an image format is resizable or not should be part of the `Darkroom` implementation (so basically GD & Imagick) - other thumb drivers might be able to deal with formats they are not (eg [my Kirby v3 wrapper](https://github.com/fundevogel/kirby3-colorist) for [`colorist`](https://github.com/joedrago/colorist)). I can't figure out a clever way to make `isResizable` a prop of the `new Filename` being called by Kirby's `thumb` method, effectively passing the information from the thumb driver to the `$file` - that would be totally sufficient. Otherwise, we might just outsource this completely to the `Darkroom` implementation, simply by resizing or not, based on the file format, or maybe `throw`ing stuff.

For now, I can implement my own `thumb` method (like @texnixe suggested on the forum), but this approach cancels all advantages of providing a real thumb driver: compatibility with other plugins that **are replacing / have to replace** the thumb method.

// Edit:
For now, maybe wrap [this line's](https://github.com/getkirby/kirby/blob/master/config/components.php#L78) check in an option, so we can disable it .. this is hacky and not desirable, but I'm desperate, as I don't want to maintain a custom Kirby version :bear:

S1SYPHOS:
I added my [feedback](https://kirby.nolt.io/110). Thanks for your time and effort!

S1SYPHOS:
Hey @bastianallgeier, one question  if I might: how do you get debugging information from inside the thumb driver (the ones being called by `Darkroom::factory()`)? No matter what I do, `var_dump()`, `throw`ing `Exception`s, nothing gets back when calling `thumb()` .. I guess I'm overlooking something ☹️

bastianallgeier (approved):

bastianallgeier:
First: thanks for changing the PR. 

TBH I think the entire Darkroom/Thumb architecture isn't ideal and as easy to extend as I wish. Whenever I want to change something or answer such a question I have to dig deep to understand myself how it works. 

The `File::isResizable()` method is especially tricky. It cannot really be connected with the Darkroom implementation. The Darkroom lib only takes care of processing images as a base layer of our async media API. But if you replace our media API with something like Cloudinary, the Cloudinary plugin must be able what's resizable. So we need a central point where resizable image formats can be stored and changed by plugins and I'm not 100% sure yet where the best place would be.

bastianallgeier:
Maybe another component? `file::isResizable`? Not sure.

lukasbestle:
Couldn't the Darkroom classes have a static method where the implementations can define which formats they can resize? Based on the config, Kirby could then call that.

S1SYPHOS:
> Couldn't the Darkroom classes have a static method where the implementations can define which formats they can resize? Based on the config, Kirby could then call that.

True. Since we implement one class (extending `Darkroom`) for each external library, those should be holding the information about their library's capabilities.

From a user perspective, I don't see why anyone would ever check for resizability by hand, so `$file` objects don't have to 'know' if they are resizable or not.. and I guess Bastian is right about its potential for improvement.

bastianallgeier:
Yeah, but if you build a plugin for an external image resizing API (like Cloudinary, or KeyCDN) you never really work with Darkroom. That's what I meant above. That's why the Darkroom class is the wrong place to hold that information either.

S1SYPHOS:
Just for clarification: wouldn't you call an external API from inside the `process()` method?

I'm standing on the hose, if you know what I mean!

bastianallgeier:
No, for image processing services like Cloudinary, it's enough to overwrite the file::version. You basically create a File version with a Cloudinary URL that contains all the resizing options. There are no API calls happening at all.

S1SYPHOS:
Thank you .. I guess helping with that is out of my scope then.

S1SYPHOS:
Sorry to bring this up again: wouldn't the cloud converter / third-party API respond with some error when sending files that aren't resizable / can't be converted?

Handling problems when processing might be handled where it occurs - for an external API, that's the `$response` object, for local thumb drivers it's their `Darkroom` implementation, no?

Sorry for bothering you.

S1SYPHOS:
Update:
ImageMagick already converts to AVIF (according to their issue tracker), so ultimately it depends on the IM versions used .. therefore, a general statement about formats being resizable or not simply isn't possible ..

On my library [`kirby3-colorist`](https://github.com/Fundevogel/kirby3-colorist), I simply replaced the `file::version` component and it works like a charm:

```php
'file::version' => function (App $kirby, $file, array $options = []) {
    // This replaces the `$isResizable` check
    $resizable = [
        'avif',
        'bmp',
        'gif',
        'j2k',
        'jp2',
        'jpeg',
        'jpg',
        'png',
        'tiff',
        'webp'
    ];

    if (in_array($file->extension(), $resizable) === false) {
        return $file;
    }

	// the rest is copy & paste from the original method
}
```

PR #2843, opened 2020-09-24
Search in Panel: new layout and no results text

distantnative:
## Describe the PR
This PR improves the search modal in the Panel:
- New layout incl. preview image/icon based on out `storybook` ideas
- Refactored component (e.g. type definitions from outside the component) based on `storybook`
- "No results" message
- Loading icon during API call
- Fixed link to files

Screen Shot 2020-09-24 at 09 21 37
Screen Shot 2020-09-24 at 09 21 52
Screen Shot 2020-09-24 at 09 14 32

## Related issues
- Implements https://kirby.nolt.io/100

### DIFF
--- a/panel/package-lock.json
+++ b/panel/package-lock.json
@@ -2493,6 +2493,14 @@
         "postcss": "^7.0.0"
       }
     },
+    "@linusborg/vue-simple-portal": {
+      "version": "0.1.4",
+      "resolved": "https://registry.npmjs.org/@linusborg/vue-simple-portal/-/vue-simple-portal-0.1.4.tgz",
+      "integrity": "sha512-vYeWYIT9RQ+Lyg6SECPf5+rMeqni7QfXVT/srPkK5NqbpKhppLtkBsPo+VlP61ZecsS1fQQxlfu6Fng/eNukQQ==",
+      "requires": {
+        "nanoid": "^2.0.1"
+      }
+    },
     "@mrmlnc/readdir-enhanced": {
       "version": "2.2.1",
       "resolved": "https://registry.npmjs.org/@mrmlnc/readdir-enhanced/-/readdir-enhanced-2.2.1.tgz",
@@ -6063,7 +6071,7 @@
           "dependencies": {
             "pify": {
               "version": "2.3.0",
-              "resolved": "http://registry.npmjs.org/pify/-/pify-2.3.0.tgz",
+              "resolved": "https://registry.npmjs.org/pify/-/pify-2.3.0.tgz",
               "integrity": "sha1-7RQaasBDqEnqWISY59yosVMw6Qw=",
               "dev": true
             }
@@ -10145,6 +10153,11 @@
       "resolved": "https://registry.npmjs.org/nan/-/nan-2.14.0.tgz",
       "integrity": "sha512-INOFj37C7k3AfaNTtX8RhsTw7qRy7eLET14cROi9+5HAVbbHuIWUHEauBv5qT4Av2tWasiTY1Jw6puUNqRJXQg==",
       "dev": true
+    },
+    "nanoid": {
+      "version": "2.1.11",
+      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-2.1.11.tgz",
+      "integrity": "sha512-s/snB+WGm6uwi0WjsZdaVcuf3KJXlfGl2LcxgwkEwJF0D/BWzVWAZW/XY4bFaiR7s0Jk3FPvlnepg1H1b1UwlA=="
     },
     "nanomatch": {
       "version": "1.2.13",
--- a/panel/package.json
+++ b/panel/package.json
@@ -12,6 +12,7 @@
     "test": "cypress run"
   },
   "dependencies": {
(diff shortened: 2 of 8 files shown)


nilshoerrmann:
This is really great! Thank you! :+1:

I've got two additional requests:

1. This is already an issue in the current implementation, but it's easy to overlook that the pages "label" is actually a select to switch search context. Would it be possible make this look like a select or at lease add the down facing arrow to make the interaction clear?
2. And for my inner designer balance, would you mind aligning the select and the input field's baselines?

grafik

nilshoerrmann:
Actually, looking at the screenshot showing the search result page, it's also not very clear that there is an input at all. 🤔
But I have to admit that I have this issue with macOS' search bar, too (or apps like Alfred).

distantnative:
@nilshoerrmann thanks for catching that. The wrong alignment was caused by the icon missing for the current type. With it, it's aligned again:
Screen Shot 2020-09-24 at 10 51 51

Regarding the input being recognizable, I know what you are referring too - but I think the input being focused on open with "Search..." as placeholder and users being trained from all kinds of systems towards this behavior makes it relatively clear that you have to start typing.

Regarding the select field, I see the issue. But haven't found a good solution yet. The arrow down looks weird (but maybe just cause I'm not used to it?):
Screen Shot 2020-09-24 at 10 55 01

bastianallgeier:
I agree with @distantnative. When you open the search for the first time, the input feels natural and once you start typing you are already in the input and it's a no-brainer that it exists. I like the minimal look of it though, especially in combination with the nice list of results. 

About the dropdown. We could maybe add a little arrow after the label.

nilshoerrmann:
Yes, I agree, it might be more obvious when used interactively.

Regarding the arrow: that's much better in my eyes! And yes, I guess, it looks strange because you are used to have the form without it before.

nilshoerrmann:
Another thing: while the more visual search results look lovely in this example, there are case where we don't have images or icons. Pages sections now allow for the preview being disabled (no image, no icon). 

How does the search handle this?

nilshoerrmann:
Regarding the arrow position:

I think using is as a prefix is good. But the distance between the arrow an the option value is wider than the distance between option and input. So it might help to either move the select value to the left or to widen the margin between the two fields.

afbora:
I'm getting following different errors in console when i click any result from different pages/models.

> vendor.js:17 Uncaught (in promise) Error: Navigation cancelled from "/site" to "/pages/photography+trees" with a new navigation.

> vendor.js:17 Uncaught (in promise) Error: Navigation cancelled from "/pages/photography+trees" to "/users/BddbqhRK" with a new navigation.

> Uncaught (in promise) Error: Navigation cancelled from "/users/BddbqhRK" to "/pages/photography+trees/files/stay-in-the-car.jpg" with a new navigation.

distantnative:
@afbora Just pushed a commit that I think fixes the problem. Can you confirm that on your end?

afbora:
@distantnative Sadly no, still same.

afbora:
@distantnative Also could you fix the padding issue that we discussed with @bastianallgeier 

Current PR:

screen-capture-294-Mægazine-localhost

Equal padding with header would be nice:

screen-capture-291-Mægazine-localhost

_* Screenshots taken before last commit :)_

afbora:
@distantnative FYI Bastian fixed before same issue on this commit https://github.com/getkirby/kirby/pull/2677/commits/b3d7017e5de04ca344ee98a2801257479db09446

distantnative:
@afbora tried to incorporate this

afbora:
@distantnative Could you able login panel? Login page is refreshing endless loop 🤔

distantnative:
Wasn't a problem when already logged in (like me), I think I fixed that too now.

afbora:
> {status: "error", message: "The file "stay-in-the-car.jpg" cannot be found", code: 404, exception: "Kirby\Exception\NotFoundException", key: "error.file.notFound", …}

Steps:
1. Enter a file detail page
2. Try to search any files and click the any result
3. See console error

GiantCrocodile:
It would be great to have a colon after the string 'User', 'Page' etc. to have a clear separation of what is content (variable title) and what's a fixed string to indicate what you look for/what was returned. It's a tiny detail of course.

distantnative:
@afbora I cannot reproduce it, maybe it's something else involved or different steps as described:

1.
Screen Shot 2020-09-24 at 20 17 24
2.
Screen Shot 2020-09-24 at 20 17 32
3.
Screen Shot 2020-09-24 at 20 17 38


@GiantCrocodile We had that before and it was good. Went missing in the changes. Will add it again.

afbora:
@distantnative Here steps screencast:

distantnative:
@afbora Thanks, could reproduce that now as well with your steps. I checked it with the current `develop` branch as well and the bug is present there already. So I'll open a separate issue as it's not caused by this PR.

https://github.com/getkirby/kirby/issues/2849

afbora:
Then there is no problem, works perfectly 👍

afbora (dismissed):

distantnative:
@bastianallgeier rebased and ready

PR #2844, opened 2020-09-24
Check isEmoji in Panel not backend

distantnative:
## Describe the PR

Instead of checking for emojis through the backend and passing that info in various places of the Panel, let's make the Icon component check itself whether the passed type is an emoji.


bastianallgeier:
I haven't read the entire article, but I saw that it is from 2016. Does this include newer emojis as well?

distantnative:
It seems to cover them even though the article is a bit older. Tested randomly a few emojis from 12.1 (newest one available on my Mac) and those worked. Also tested the ones that are reported as not working for this library https://github.com/mathiasbynens/emoji-regex/issues and those worked with that regex as well.

Really haven't found any that didn't work (but of course also didn't test all of them).

bastianallgeier (dismissed):

bastianallgeier (approved):

bastianallgeier:
Don't you think we could put this on the develop branch?

distantnative:
@bastianallgeier Sure

distantnative:
Against `develop`: https://github.com/getkirby/kirby/pull/2848

PR #2847, opened 2020-09-25
Builder: move blocks between fields

distantnative:
Each builder field has to define the same `group` option to allow blocks to be moved between those fields. 
Moving a block is only allowed if
- the target field has a fieldset with the same name as the block's type
- the target field is not full (`max` option)

### DIFF
--- a/panel/src/components/Forms/Field/BuilderField.vue
+++ b/panel/src/components/Forms/Field/BuilderField.vue
@@ -172,7 +172,7 @@ export default {
   computed: {
     draggableOptions() {
       return {
-        id: Math.random(),
+        id: this._uid,
         handle: true,
         list: this.blocks,
         move: this.move,
--- a/panel/src/components/Forms/Field/BuilderField.vue
+++ b/panel/src/components/Forms/Field/BuilderField.vue
@@ -19,88 +19,92 @@
       </k-dropdown-content>
     </k-dropdown>
 
-    <template v-if="blocks.length === 0">
-      <k-empty icon="box" @click="select(blocks.length)">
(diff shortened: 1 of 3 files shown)


panel/src/components/Forms/Field/BuilderField.vue
```diff
@@ -165,6 +170,21 @@ export default {
     };
   },
   computed: {
+    draggableOptions() {
+      return {
+        id: Math.random(),
```
distantnative: @afbora I copied this from your draft. I am wondering now whether we could find a cleaner solution for this than a random number.

afbora: Definitely, I just needed a unique id. I was surprised that it even got this far 😅

distantnative: Could we just use the field's `name`? Cause a field name should be unique for a page (view), right?

afbora: Actually it was the field name that came to my mind, but if there is another builder with the same name in the nested field, I thought it would conflict.

distantnative: Always those nested fields...

distantnative: Maybe silly idea, but we could use `endpoints.field` - that should be unique even for nested fields.

distantnative: I went for `this._uid`

bastianallgeier (approved):
Awesome!

PR #2850, opened 2020-09-25
FileView: fix updating parent #2849

### LINKED ISSUE #2849: FileView: search results from other parent model lead to console error
**Describe the bug**  
If I am on a FilewView with parent page A and I search for files and click on a file with parent page B, I'll get a console error

**Screenshots**  

**Kirby Version**  
3.4.3

**Console output**  
Failed GET to http://localhost:8080/api/pages/portfolio+projects+ocean/files/desert-tree.jpg/sections/content 404 (Not Found)

» the model id in this URL is already of the target page, the file id still of the previous/current file

distantnative:
## Describe the PR

The previous line https://github.com/getkirby/kirby/commit/4efb37cf5b775152bd8abfa92d6ba9499e4283a3#diff-9d0814f74112ef41fe0eb1f977d12451L51 would already update when `path` was updated, triggering the Tabs-Component to update, trickling down to the sections. All that before `file.filename` was updated in the `fetch()` method.

That way sections would first get an incorrect endpoint (new parent/path, old filename) and only later in a second update run the functioning one. This would only happen when navigating from one FileView to another FileView (mostly through search).

Adding a new data entry `parent` and setting it within `fetch()` allows to delay the full update cycle until both `path` and `file.filename` have been updated.

## Related issues

- Fixes #2849


bastianallgeier (approved):
Looks a lot cleaner!

PR #2854, opened 2020-09-26
Fields: indicator for unsaved changes

distantnative:
## Describe the PR

Screen Shot 2020-09-26 at 08 14 27

## Related issues

- https://kirby.nolt.io/49

### DIFF
--- a/panel/src/components/Forms/Field.vue
+++ b/panel/src/components/Forms/Field.vue
@@ -2,7 +2,6 @@
   <div
     :data-disabled="disabled"
     :data-translate="translate"
-    :data-hasChanges="hasChanges"
     :class="'k-field k-field-name-' + name"
     @focusin="$emit('focus', $event)"
     @focusout="$emit('blur', $event)"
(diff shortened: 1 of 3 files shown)


distantnative:
@bastianallgeier having the badges on tabs for unchanged changes, something like this (styling up for discussion) kinda needs to be in as well ;)

distantnative:
@bastianallgeier 
I am pretty torn here:
On the one hand, I would approach this as we did a lot on `storybook` - make the lower level components dumb. In that direction, `k-field` and the label shouldn't have none of the logic but only receive `hasChanges` and `hasError` props and display a UI accordingly. Setting and passing down the props would take place in `k-fields-section`.
But I am afraid whether this makes it more difficult for custom fields. Will all plugin developers pass down the props to `k-field`?

bastianallgeier:
Argh, this is one of the PRs that seems really simple but is adding a lot of questions once you look closer. Can we move it to 3.6 or a 3.5.x release please? I think the tab badges still add value without the labels

distantnative:
Sure I agree we have open questions to answer first here. Let's move it (or even to a 3.5.x release)

distantnative:
I think the real question here is: Where in our component tree do we want to add the logic of getting the value from the store? In the label as it is now, it doesn't feel right indeed. Field itself should also be decoupled from this.

Either we go all the way up to FieldsSection and pass down a prop (first an array of field names, then just a boolean) - or we jump in somewhere in the middle, e.g. Fieldset. Cleanest would probablyy be FieldsSection, but also passing a lot down.

PR #2855, opened 2020-09-26
Merge page rename and page url dialog

distantnative:
- Rename and change slug are combined into one dialog
- Dropdown options rename, both open the new combined dialog
- Depending which dropdown action opens the dialog, the corresponding field is preselected

Screen Shot 2020-10-16 at 19 36 17

## Related issues

- https://kirby.nolt.io/26


afbora:
I had to show my skills and I tried hard to find any bugs but couldn't find it 😅 Looks awesome! I like this feature so much, it is very practical 🎉 

---

It's not a bug, but I have an improvement request.

````yaml
title: Editor

permissions:
    pages:
        changeTitle: true
        changeSlug: false
````

If a user does not have permission to change the title or slug, input that not allowed comes active in dialog window. 
Even if user tries to update, the system does not allow it, but I think it will be clearer for users if the input is disabled 🤔 

screen-capture-309-Test - Mægazine-localhost

**Comment by distantnative** [comment]
@afbora I added something, is that what you meant?

And good caught. When they were separate dialogs, the menu options would be disabled and you couldn't open the dialogs in the first place. But of course now you might be able to open the dialog if you have one permission, but not the other.

Maybe a question to discuss: **Should those two permissions be merged as well?**

**Comment by afbora** [comment]
> you might be able to open the dialog if you have one permission, but not the other.

Yes, i meant this case.

> I added something, is that what you meant?

Although `Change Slug` is disabled in the menu, the slug input is still active.

> Should those two permissions be merged as well?

I'm in favor of not being merged. I think the current situation is good. Because since there is access to the page with slug, I can let it change the title while not allowing the slug.

**Comment by distantnative** [comment]
@afbora I think now I got the permissions right

**Comment by afbora** [comment]
@distantnative Yes it's perfect now 🎉

File: panel/src/components/Sections/PagesSection.vue [hunk 1]
```diff
@@ -119,11 +118,12 @@ export default {
           break;
         }
         case "rename": {
-          this.$refs.rename.open(page.id);
+          console.log(page);
```
afbora: :point_up:

panel/src/components/Views/PageView.vue
```diff
@@ -182,38 +181,35 @@ export default {
...
+
+      } catch (error) {
+        this.issue = error;
+      }
     },
     onTab(tab) {
       this.tab = tab;
     },
     update() {
+      console.log("updating!")
```
afbora: :point_up:

afbora:
I just wonder why can't we get this to 3.5?

distantnative:
There is already so much in line for 3.5 - I don't want to overbloat it, which always has the danger that we might oversee bugs, regressions. That's why I wanted to space it out a little.

bastianallgeier:
I love it! I just improved auto-selection a bit.

bastianallgeier (approved):

PR #2857, opened 2020-09-26
Overlay component

distantnative:
Introduces `k-overlay` as wrapper for overlaid elements (dialogs, search...) with/without backdrop and loading state.

Requires https://github.com/getkirby/kirby/pull/2843


distantnative:
@bastianallgeier how do you feel about this one for 3.5.0? Otherwise 3.5.1 could be an option, since it's only internal.

bastianallgeier:
Let me have a closer look at it, but in general I think it's a good idea to include it.

bastianallgeier:
Argh, this leads down a rabbit hole. It almost works perfectly, until you realise that autofocus and the focus trap is broken in dialogs. I remembered that I solved this in storybook and checked out how, just to find that it only works there because of the additional modal component. 🙈

distantnative:
@bastianallgeier true! but does it have to be a rabbit hole? Gave it a first try - needs more testing to see if that's already the solution but looks to me like this could be solvable without all the storybook overload around it.

bastianallgeier:
Yay, you are right. This works like a charm 🎉

bastianallgeier (approved):

PR #2871, opened 2020-10-03
Quick fix for replacing Panel views #2506

### LINKED ISSUE #2506: Overwrite Panel views broken
```js
panel.plugin("my/view-plugin", {
  views: {
    users: {
      icon: "heart",
      menu: true,
      component: {
        template: `
          <k-view class="k-todos-view">
            <k-header>
                Todos
            </k-header>

            <!-- your app goes here -->

          </k-view>
        `
      }
    }
  }
});
```

distantnative:
## Describe the PR

Panel views can be replaced by custom views again:

```js
panel.plugin('texnixe/users-view', {
  components: {
    "k-users-view": {
      template: "<p>This works!</p>"
    }
  }
});
```

Credit to @iskrisis!

It's not the prettiest fixes - but that is due to our main setup of loading config parts etc. isn't clean either right now. That will be a separate refactoring task. But until then this fix should work.

## Related issues

- Fixes #2506


panel/src/config/routes.js
```diff
@@ -1,6 +1,7 @@
 import Vue from "vue";
 import auth from "./auth.js";
 import store from "@/store/store.js";
+import components from "./plugins.js";
```
lukasbestle: I think there should be a code comment here that explains why this was added.

iskrisis:
It's stupid fix to be honest. It just imports useless variable - i just picked "components" because it was already there but it would work with any variable. There must be some better mechanism to do this.

I think we all are used to php way of imports which would be similar to require() in JS but ES6 modules as i undestand it puts all imports on top of the file and does some magic afterwards. https://github.com/webpack/webpack/issues/1343#issuecomment-263281890

I am a bit worried that this change might break something that relied on the old order. 

Also that confusion might lead to bugs in future so somebody will probably have to think about it a bit deeper.

distantnative:
As I wrote, this mainly originates in the fact that our main setup of loaded config, plugins, extensions isn't ideal yet.
Refactoring this though is a bigger task that shouldn't block this fix.

distantnative:
I simplified this a bit more. A simple unnamed `import "./plugins.js"` works too.

lukasbestle (dismissed):
I haven't tested it, but with the added code comment it looks good to me. 👍

bastianallgeier (dismissed):

bastianallgeier:
Looks good to me too, but I would prefer not to add it to the develop branch before 3.4.4 is out.

PR #2872, opened 2020-10-03
Prevent concurrent sort API calls #2863

### LINKED ISSUE #2863: Sorting pages can lead to duplicates in the file system
**Describe the bug**  
Sorting within a larger number of pages – in my case 20 – can take some time on the server side, and you can continue sorting the pages even though the previous operation is not complete. The operations start to interfere with each other, which can lead to duplicate folders with different numerical prefixes in the file system.

The latency is only noticeable when using plugins that use the changeNum:after hook, like "AutoID" or "Git Commit And Push Content", because their operations are a bit more time consuming.

**To Reproduce**  
1. Install https://getkirby.com/plugins/blankogmbh/git-commit-and-push-content and https://getkirby.com/plugins/bnomei/autoid
2. Create multiple listed Pages ( >= 20 should do it)
3. Move the last page to the top and then immediately move another page from the end of the list to a new position.
4. Take a look at your content folder, watch how the folders get new names one after another and wait for duplicate folder names

**Expected behavior**  
A modal indicator should be displayed while the operation is not finished

**Kirby Version**  
3.4.2, 3.4.3, not sure about previous versions

distantnative:
## Describe the PR

If sort API calls would take long (for whatever reasons) and the user would quickly sort two files, this could result in duplicates or other troubles.

PR adds to pages and files sections:
- `isProcessing` property
- disables mouse interactions while `isProcessing`
- disables sorting in `k-draggable` while `isProcessing`

## Related issues

- Fixes #2863


bastianallgeier (dismissed):

bastianallgeier (approved):

Sija:
@distantnative IIUC there's still an edge-case when you'd execute sorting in two (or more) open windows in the same time, is that correct?

Issue #1663, opened 2019-01-11
[Panel] Sorting across pagination divide


johannahoerrmann:
I'm revisiting a "complaint" I had back in June, regarding the ability to sort entries in a paginated section, see https://github.com/k-next/kirby/issues/528.

The reason is that this has come up as a question from our client _twice_ since November.

Here's the first time, also mentioned here https://github.com/k-next/kirby/issues/528#issuecomment-440183572:

> Another weird one - [we] had added a bunch of awards in draft, […] and I have re-ordered the ones that I could, but, a few need to be moved to the 21-40 page - how do you do that - you can’t seem to drag it down and trigger the next page to drop it. See screen shot.

We answered this of course and told her to set the position using the status interface, but now we were actually asked _again_ – showing it is just not self-expanatory.

> I would like to move some of the [entries] to their correct places, but how do you do that if it is not the list being displayed in kirby (as only 20 at a time)?

In the issue referenced above, @nilshoerrmann actually proposed this solution:

> This issue could be solved by making the pagination a drop area saying "move to page 21–40". When moving a page to this zone, the panel could load the next page giving the user the option to integrate the selected page in the new context. Quick mockup:

![](https://user-images.githubusercontent.com/25466/48760076-44a19800-eca5-11e8-9876-0888e7daeef5.png)

ghost:
Also need a fix for this 👍

nilshoerrmann:
Besides the more complex drag and drop solution suggested above (I still like the idea), having an UI to change the maximum number of items displayed per page would be an intermediate solution. It would require some thoughts on persistence though: does the limit only change temporarly in the browser or is it stored per section and used across accounts and devices?

johannahoerrmann:
> does the limit only change temporarly in the browser or is it stored per section and used across accounts and devices?

I'm not sure how this could/would be implemented, but I would find it weird to work across accounts. So if storing it per section across devices would mean for all accounts, I'd say no. If it doesn't, sure!

nilshoerrmann:
I don't know how the Panel deals with this kind of stuff but I see a few options:

1. I can change the limit locally, the setting is lost on reload (no storage).
2. I can change the limit locally and it's stored in `localStorage` and is reapplied in my browser after refreshing.
3. I can change the limit and it's stored globally on the server.

I'd prefer 2, maybe with the option to reset the limit to the default setting.

johannahoerrmann:
I also prefer 2.

distantnative:
[...]

PR #2874, opened 2020-10-03
Page/file sort dialog

distantnative:
## Describe the PR

Adds a "change position" dialog to the options dropdown of pages and files sections (not page/file view though). The dialog reuses the select from the existing PageStatusDialog.

Screen Shot 2020-10-16 at 19 17 22

Screen Shot 2020-10-16 at 19 17 31

Screen Shot 2020-10-16 at 19 26 48

Screen Shot 2020-10-16 at 19 26 53

## Related issues

- Implements #1663


distantnative:
Though I just had the idea, if a move page/file dialog could solve this too:
The move dialog would be a pages field (to select the parent page) + position select (if sortable/if sorted by num).
It would open with the current parent and position preselected.

So if one would just want to resort it, they would only change the position and leave the parent unchanged.
But maybe this PR is a first step and this new idea could extend/replace it in 3.6, 3.7 ...

afbora (commented):
When set `sortable: false` from section, `Change position` disabled as expected in three dots menu. But position dropdown is still appears and works in `Select a new status` dialog. Is this bug or related with this PR?

distantnative:
@afbora that sounds like it could be a bug already in Kirby, not new with this PR.

afbora:
Yes I tested it now, this issue already existed (https://github.com/getkirby/kirby/issues/2901). So PR looks great 👍

bastianallgeier (approved):

nilshoerrmann:
This looks great and will be very helpful. Thanks 👍

johannahoerrmann:
Fantastic! This will help a lot.

PR #2877, opened 2020-10-03
Adding AVIF to `$isViewable`

S1SYPHOS:
See #2870 

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/FileFoundation.php
+++ b/src/Cms/FileFoundation.php
@@ -142,6 +142,7 @@ trait FileFoundation
     public function isViewable(): bool
     {
         $viewable = [
+            'avif',
             'jpg',
             'jpeg',
             'gif',


distantnative:
@afbora I might of your PR https://github.com/getkirby/kirby/pull/2777, do you think we should make the `isViewable` check extendable as well?

S1SYPHOS:
Quick question: With the idea behind @afbora's PR, we could also set custom formats from plugins, right? Before the `Kirby::plugin` call ofc? This should then be extended to `$isReszable`, as well (see the discussion over at #2840) ..

distantnative:
> Before the Kirby::plugin call ofc?

No, @afbora's PR would allow to set this in the config or within the `Kirby::plugin()` call.

S1SYPHOS:
Thanks for clarifying. Sounds wonderful! Right now I have to overwrite `file::version` to get past the `$isResizable` check. Again, I absolutely positively realise these are edge cases - but making things configurable (for the developer at least) is a plus.

afbora:
@distantnative If the custom file extension is `isViewable` or `isResizable`, yes IMO it should be extendable to work correctly in other parts of Kirby. Also what do you think about that? @lukasbestle 

May be sample new usage for `isViewable` and `isResizable` based on #2777 PR? 

````php
return [
    'debug' => true,
    'file' => [
        'types' => [
            'image' => [
                'heic'
            ],
            'video' => [
                'm4p'
            ]
        ],
        'isViewable' => [
            'avif'
        ],
        'isResizable' => [
            'svg'
        ]
    ]
    'mime' => [
        'types' => [
            'heic' => $mime = ['image/heic, image/heif']
        ]
    ]
];
````

lukasbestle:
@S1SYPHOS We will include `avif` in the list of viewable file types in Kirby 3.6.0. We have also added a plugin extension to register custom file types: #2777

PR #2880, opened 2020-10-03
Fix i18n rules loading #2230

### LINKED ISSUE #2230: Issue about loading translation rules
This bug details are in comments after this: https://github.com/getkirby/kirby/issues/1903#issuecomment-541344389

```
<?php
echo Str::slug("Компания");
echo t('something');
echo Str::slug("Компания");
?>
```

In RU language I get:
```
kompanija
translation of something
kompaniya
```

See difference `j` to `y`

distantnative:
## Describe the PR

So far the `I18n::$load` callback ran only when using a `translate()` helper (or anything else that would call a `I18n::` method). However, `I18n::$load` also sets the language rules for the `Str` class.

To make sure the `Str` class has all the rules for the current language, even before calling any `I18n::` method, we need to run `I18n::$load` once on loading a page.

## Related issues

- Fixes #2230


afbora:
@distantnative For example; it works in Russian as expected now. But other languages seem to be inconsistent? I'm not sure.

> kompaniya > ru > kompaniya
> kompanija  > en > kompaniya

Shouldn't it be like below for English? For a language that does not contain the relevant rule, I think the before and after should be the same.
> kompanija  > en > kompanija  

Code Snippet
````php
echo Str::slug("Компания") . " > "; // before calling translate method
t('something'); // call translate method
echo $kirby->languageCode() . " > ";
echo Str::slug("Компания"); // after calling translate method
````

distantnative:
@afbora does it make a difference when you call `t()` with an existing language string, e.g. `t('files.empty')`?

For me it does because `I18n::tranlaste()` (called by `t()`) tries the fallback with default language.
But that might be a good argument, that we currently set `Str::$language` in the wrong place. Cause it shouldn't be constantly changed around by these calls.

afbora:
Yes you are right. If the language string is exists, the inconsistency issue doesn't appear 👍

distantnative:
Great! At least that. But I just realized that we are doing it not in the right place so far. Have to dig a bit more.

distantnative:
I think I found a better setup. Just trying to see still that this doesn't cause any regressions.

afbora (dismissed):
It's definitely much cleaner and stable, love it ❤️

bastianallgeier (dismissed):

bastianallgeier (approved):

PR #2881, opened 2020-10-04
Deprecate $page->sort()

### LINKED ISSUE #2475: sortBy('sort', 'desc') from template re-orders content
**Describe the bug**  
using `->sortBy('sort', 'desc')` in a template on a collection of pages causes them to be re-numbered in the content folder.

Example code:
```
<?php foreach ($site->children()->listed()->sortBy('sort', 'desc') as $p) : ?>
  <h2><?= $p->title()->html()?></h2>
<?php endforeach ?>
```

I realize this is not the correct way to change the order of the pages, but the result seems like a bug that should be fixed.

**To Reproduce**  
1. download example of the bug added to the plainkit here: https://github.com/maxackerman/plainkit
2. create a user and log in
3. vist the home page in your browser
4. refresh to see the order change, and look at folders in /content

**Expected behavior**  
Template code should not change the content folder

**Screenshots**  

**Kirby Version**  
3.3.4

distantnative:
**WIP to fix #2475**

- [x] Mark `$page->sort()` as deprecated in 3.5
- [x] Add `$page->changeSort()` as public alias for `$page->changeStatusToListed()`


distantnative:
@lukasbestle when I want to add the following to `Kirby\Cms\Toolkit` I am running in loads oh unit test error and failures:

```php
   /**
     * Sorts the elements by any number of fields.
     * Calls `Kirby\Toolkit\Collection::sortBy` without any permissions.
     *
     * @param string|callable $field Field name or value callback to sort by
     * @param string $direction asc or desc
     * @param int $method The sort flag, SORT_REGULAR, SORT_NUMERIC etc.
     * @return \Kirby\Cms\Collection
     */
    public function sortBy()
    {
        $args = func_get_args();
        return kirby()->impersonate(null, function () use ($args) {
            return parent::sortBy(...$args);
        });
    }
``` 

I don't get it quite, cause the errors are also things like:
```
25) Kirby\Toolkit\FTest::testNiceSize
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'4.1 KB'
+'4,1 KB'
```

Which confuses me how my change would cause that... 
Since you know the impersonate part of Kirby much better, help! :D

src/Cms/PageActions.php
```diff
@@ -247,6 +247,15 @@ protected function changeStatusToUnlisted()
...
 
+    /**
+     * @param null $position
+     * @return self
+     */
+    public function changeSort($position = null)
+    {
+        return $this->changeStatus('listed', $position);
+    }
+
```
lukasbestle: As I explained in https://github.com/getkirby/kirby/issues/2475#issuecomment-593088270, I'm not 100 % happy about introducing a `changeSort()` method. Also see my comment below.

src/Cms/PageActions.php
```diff
@@ -768,11 +777,15 @@ public function resortSiblingsAfterUnlisting(): bool
     }
 
     /**
+     * @deprecated 3.5.0 Use `Page::changeSort()` instead
```
lukasbestle, line 788: We already have `$page->changeStatusToListed($position)`, which does exactly what the `changeSort()` method would do and would only need to be made public. So I'd point directly to this method here.

```suggestion
     * @deprecated 3.5.0 Use `Page::changeStatusToListed()` instead
```

Or we could keep the `changeStatusToListed()` method protected and point to `changeStatus('listed', $position)`.

src/Cms/PageActions.php
```diff
@@ -768,11 +777,15 @@ public function resortSiblingsAfterUnlisting(): bool
...
+     *
      * @param null $position
      * @return self
      */
     public function sort($position = null)
     {
+        deprecated('$page->sort() is deprecated, use $page->changeSort() instead. $page->sort() will be removed in Kirby 3.6.0.');
```
lukasbestle, line 795: Same as above.

```suggestion
        deprecated('$page->sort() is deprecated, use $page->changeStatusToListed() instead. $page->sort() will be removed in Kirby 3.6.0.');
```

lukasbestle (commented):
I'm looking into the `impersonate()` issue. A few comments on the changes so far:

src/Cms/PageActions.php
distantnative: I didn't quite understand your concerns to be honest. Having `changeNum` and `changeSort`, which do the same, doesn't sound too confusing to me. In fact, probably instead of doing `$this->changeStatus('listed', $position)` in `changeSort` it should straightforward be a simple alias. I think it will help people, when they can use `changeSort` both on `$page` and `$file` - instead of remembering that the one needs `changeNum`, the other `changeSort`.

src/Cms/PageActions.php
distantnative: I would keep the `changeStatusTo...` methods protected to have consistency. Those are not meant to be called directly, instead use `$this->changeStatus('listed', $position)` then.

src/Cms/PageActions.php
distantnative: Although using `changeNum` wouldn't resort the siblings, right? Should `changeNum` be even public then?

lukasbestle: You are right – I didn't consider that `$file->changeSort()` already exists.

> Although using `changeNum` wouldn't resort the siblings, right?

Correct!

> Should `changeNum` be even public then?

There are probably use-cases where this behavior is required, for example if many pages need to be sorted in different positions in one go. It would be very slow to resort every sibling on each sort operation as each page would be sorted multiple times. So this "manual" method is useful for these use-cases.

Maybe we could make the warning in the docblock clearer though.

src/Cms/PageActions.php
```diff
@@ -247,6 +247,15 @@ protected function changeStatusToUnlisted()
         return $page;
     }
 
+    /**
+     * @param null $position
+     * @return self
+     */
+    public function changeSort($position = null)
```
lukasbestle: This should have a type hint:

```suggestion
    /**
     * @param int|null $position
     * @return self
     */
    public function changeSort(?int $position = null)
```

lukasbestle (commented):
I will open a second PR for the `impersonate()` change to keep the overview.

src/Cms/PageActions.php
```diff
@@ -247,6 +251,19 @@ protected function changeStatusToUnlisted()
         return $page;
     }
 
+    /**
+     * Change the position of the page in its siblings
+     * collection. Siblings will be resorted. If the page
+     * status isn't yet `listed`, it will changed to it.
```
lukasbestle:
```suggestion
     * status isn't yet `listed`, it will be changed to it.
```

src/Cms/PageActions.php
```diff
@@ -247,6 +251,19 @@ protected function changeStatusToUnlisted()
...
+     *
+     * @param int|null $position
+     * @return self
+     */
+    public function changeSort(int $position = null)
```
lukasbestle, line 262: As this can be `null`, it should be made explicit for consistency:

```suggestion
    public function changeSort(?int $position = null)
```

lukasbestle (commented):
Only two typos, besides that 👍.

src/Cms/PageActions.php
distantnative: But... the consistent way would be without the `?` as we don't use it in most places (since it's a more recent feature compared to most of our code base).

lukasbestle (dismissed):

lukasbestle (dismissed):

bastianallgeier (dismissed):

bastianallgeier (approved):

PR #2882, opened 2020-10-04
Prevent accidental writes from collection methods

### LINKED ISSUE #2475: sortBy('sort', 'desc') from template re-orders content

lukasbestle:
## Describe the PR

By disabling logged-in users while running collection methods like `sortBy()`, `filterBy()` and `groupBy()`, accidental calls to methods that cause write operations (like `$page->sort()`) are prevented.

## Related issues

- Fixes a part of #2475 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/Collection.php
+++ b/src/Cms/Collection.php
@@ -4,8 +4,10 @@ namespace Kirby\Cms;
 
 use Closure;
 use Kirby\Exception\InvalidArgumentException;
+use Kirby\Exception\PermissionException;
 use Kirby\Toolkit\Collection as BaseCollection;
 use Kirby\Toolkit\Str;
+use Throwable;
 
 /**
  * The Collection class serves as foundation
@@ -120,6 +122,32 @@ class Collection extends BaseCollection
     }
 
     /**
+     * Filters elements by one of the
+     * predefined filter methods, by a
+     * custom filter function or an array of filters
+     *
+     * @param string|array|\Closure $field
+     * @param array ...$args
+     * @return \Kirby\Cms\Collection
+     */
+    public function filter($field, ...$args)
+    {
+        return $this->callProtected('filter', func_get_args());
+    }
+
+    /**
+     * Find a single element by an attribute and its value
+     *
+     * @param string $attribute
+     * @param mixed $value
+     * @return mixed|null
+     */
+    public function findBy(string $attribute, $value)
+    {
+        return $this->callProtected('findBy', func_get_args());
+    }
+
+    /**
      * Groups the items by a given field or callback. Returns a collection
      * with an item for each group and a collection for each group.
      *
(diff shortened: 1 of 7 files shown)


src/Cms/Collection.php
```diff
@@ -321,6 +322,32 @@ public function search(string $query = null, $params = [])
...
+    public function sortBy()
+    {
+        $auth = kirby()->auth();
+        $userBefore = $auth->currentUserFromImpersonation();
+
+        try {
+            $auth->impersonate('nobody');
+            return parent::sortBy(...func_get_args());
+        } catch (Throwable $e) {
+            throw $e;
```
distantnative, line 425: Maybe we should provide a custom error message if it is a PermissionException? Instead of something like "insufficient permissions" we could let the user know that they are falsely using the name of an altering method instead of the field they likely wanted to use.

distantnative:
Looks good to me! Just the one idea.

distantnative:
@lukasbestle LGTM but it'd need to be rebased first to match the general collection methods changes.

src/Cms/Collection.php
```diff
@@ -337,4 +339,34 @@ public function toArray(Closure $map = null): array
...
+     * @return mixed
+     */
+    protected function getAttributeFromObject($object, string $attribute)
+    {
+        $auth = kirby()->auth();
+        $userBefore = $auth->currentUserFromImpersonation();
+
+        try {
+            $auth->impersonate('nobody');
+            return $object->{$attribute}();
```
distantnative: Just to make changes easier down the road: I'd actually call the parent method here. So if we make changes to that, we don't need to make them twice ideally.

distantnative:
Sorry, I totally missed that you had rebased it again already.
Just had one more thought on it :)

lukasbestle:
@distantnative Thanks for your feedback! Should be ready now.

bastianallgeier (commented):
I was just wondering about the performance implications of this change. The getAttributeFromObject method is executed for every item when sorting or filtering. Wouldn't it make more sense to wrap sortBy and filterBy in those permission checks to only run the impersonation code once?

lukasbestle:
@bastianallgeier I had it that way first (see my first force push to this PR), but the issue is that we have quite a few methods that use `getAttributeFromObject()` internally, so we would need to duplicate the code everywhere. My concern is that we might forget updating a method and also there could be plugin methods that would need to be updated. Fixing it right at this central location is the most DRY and safe way.

You are right that we should be careful about performance. The impersonation doesn't require any IO access though, just in-memory object manipulation. My assumption is that the performance hit should be minimal, but we should test that with a large collection.

bastianallgeier:
@lukasbestle if it doesn't have any noticeable performance implications your solution is definitely more reliable! That's for sure.

lukasbestle:
I tested sorting 50000 objects in a collection and the difference was truly surprising. On the `features` branch it took around 0.35 seconds, with the changes in this PR it took over 6 seconds. So it would have been a truly massive performance issue. Glad you spotted it @bastianallgeier! 😌

I have now wrapped all collection methods that currently use `getAttribute()` with a protective layer. I also realized: As `getAttribute()` is a protected method, plugins cannot call it, so they cannot have been affected by this change anyway. So we should now be safe for every core collection method and every plugin method that only accesses public collection methods internally.

bastianallgeier:
Phew, I'm really glad you tested this! Did you run the same test after the changes as well? I don't think there will be any significant hit afterwards, but just to be sure. 

I think we should try to spend some time in the coming months to do more profiling around potential bottlenecks. I was really astonished by the impact of the stuff that @hdodov spotted in the Page class and the fixes are often not that difficult.

lukasbestle:
@bastianallgeier Yes, with the new implementation the performance is not measurably different than before.

My plan is to do automated tests in Travis builds. I have an idea for that, but didn‘t find the time so far. After 3.5 is out, this could be my next project.

bastianallgeier (approved):

bastianallgeier:
Urghs, you know it's Monday morning when you accidentally hit "close PR" instead of "merge PR" 🙈

PR #2883, opened 2020-10-04
Callback support for Collection::filterBy

distantnative:
## Describe the PR

- Support passing Closure or array to `Collection::filterBy` (so far supported in `Collection::filter`)
- Use `Collection::filter` as simple alias for `Collection::filterBy`
- Marked `Collection::filter` as deprecated, no warning yet though - I'd suggest to add the warning in 3.6 or 3.7 and eventually remove it in 3.8 or 3.9

### DIFF
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -237,18 +237,15 @@ class Collection extends Iterator implements Countable
     }
 
     /**
-     * Filters elements by a custom
-     * filter function or an array of filters
+     * Shorthand/alias for static::filterBy()
      *
-     * @deprecated 3.5.0 Use `Kirby\Toolkit\Collection::filterBy` instead
-     *
-     * @param array|\Closure $filter
-     * @return self
-     * @throws \Exception if $filter is neither a closure nor an array
+     * @param string|array|\Closure $field
+     * @param array ...$args
+     * @return \Kirby\Toolkit\Collection
      */
-    public function filter($filter)
+    public function filter($field, ...$args)
     {
-        return $this->filterBy($filter);
+        return $this->filterBy($field, ...$args);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Toolkit/Collection.php
```diff
@@ -238,50 +238,63 @@ public function extend($items)
...
-
-            foreach ($filter as $arguments) {
-                $collection = $collection->filterBy(...$arguments);
-            }
-
-            return $collection;
-        }
-
-        throw new Exception('The filter method needs either an array of filterBy rules or a closure function to be passed as parameter.');
+        return $this->filterBy($filter);
```
lukasbestle: Please add a code comment above this line that the deprecation warning was not added on purpose for now and that it should be added in one of the next minor releases. Otherwise we might forget about it and will need to search for the discussion in this PR to find out what the plan was.

src/Toolkit/Collection.php
```diff
@@ -238,50 +238,63 @@ public function extend($items)
...
     }
 
     /**
      * Filters elements by one of the
-     * predefined filter methods.
+     * predefined filter methods, by a
+     * custom filter function or an array of filters
      *
-     * @param string $field
+     * @param string|Closure $field
```
lukasbestle:
```suggestion
     * @param string|array|Closure $field
```

lukasbestle (commented):
Besides the two comments I like the changes!

src/Toolkit/Collection.php
bastianallgeier: I don't think we should fully deprecate this method in v3 at all and keep it as shorthand/alias.

bastianallgeier:
I removed the deprecation warning for now to merge this. If we decided to add it later, we can still do that.

bastianallgeier (approved):

distantnative:
Arg, sorry I was just working still on the changes discussed. Cause we said we actually wanna use `filter` and make `filterBy` the alias

distantnative:
@bastianallgeier and just a quick note - let's use full classnames instead of `static::filterBy()` - so that the reference on getkriby.com can parse it and link it properly.

distantnative:
Moved the additional changes to https://github.com/getkirby/kirby/pull/2885

PR #2885, opened 2020-10-07
Unify collection methods

distantnative:
@lukasbestle Is that the direction we discussed (from your view)?
Still running into 2 unit tests failing.

We also talked about `find` and `findBy` but those don't seem as straightforward to me.

### DIFF
--- a/config/api/models/File.php
+++ b/config/api/models/File.php
@@ -48,7 +48,7 @@ return [
             return $file->next();
         },
         'nextWithTemplate' => function (File $file) {
-            $files = $file->templateSiblings()->sortBy('sort', 'asc', 'filename', 'asc');
+            $files = $file->templateSiblings()->sort('sort', 'asc', 'filename', 'asc');
             $index = $files->indexOf($file);
 
             return $files->nth($index + 1);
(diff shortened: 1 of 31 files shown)


lukasbestle:
The general direction (code in `filter/group/sort` and aliases as `filterBy/groupBy/sortBy`) is what we discussed. I agree on `find/findBy` – with their similar argument syntax it will likely not be possible to merge them. But I'd say that's fine. Better to unify the other methods than none.

I haven't reviewed the code in detail yet in the draft state. Feel free to request a review once the PR is ready, also let me know in case you want any help. :)

distantnative:
@lukasbestle I don't quite get why `phpunit tests/Toolkit/CollectionTest` gives me

```
1) Kirby\Toolkit\CollectionTest::testGroupByWithInvalidKey
Failed asserting that exception of type "Exception" is thrown.
```

but `phpunit tests/Toolkit/CollectionTest --filter=testGroupWithInvalidKey` runs without failures.

afbora:
@distantnative First, you actually wrote wrong the method name on filter, so missing `By` and should be like following to get failure 😅

    phpunit tests/Toolkit/CollectionTest --filter=testGroupByWithInvalidKey

Second, as you know you may forgot to throwing an `Exception` for a non-string parameter like in current `groupBy` method:

````php
if (is_string($field) === false) {
    throw new Exception('Cannot group by non-string values. Did you mean to call group()?');
}

distantnative:
🙈 stupid haha. thanks for pointing out the typo.

afbora:
@distantnative Missing one more `sortBy()` method:
https://github.com/getkirby/kirby/blob/refactor/collection-methods/src/Toolkit/Collection.php#L864

bastianallgeier (dismissed):

bastianallgeier (approved):

PR #2889, opened 2020-10-08
Tailwind-like utility classes

distantnative:
@bastianallgeier while I think this is ready as a PR as is, we probably should only add this to a version where we actually refactor (at least some of) our components to use it. So maybe 3.5.1


bastianallgeier:
Sorry to move this back again. But since we started working on the website and we are going to use utilities there as well, it's a good idea IMHO to review them during this process.

distantnative:
Totally fine. In fact, I will close this PR, store it locally and create a new/adapted one when the time is right. I know, it helps you to get PRs that don't require attention our of sight.

PR #2811, opened 2020-09-08
Additional icons

distantnative:
## Describe the PR

More from the `storybook` branch:

- Added icons:
  - blank
  - cancel-small
  - circle-half
  - circle-nested
  - live
  - road-sign

PR #2890, opened 2020-10-09
New page status icons

distantnative:
Another bit from `storybook`:

Screen Shot 2020-10-09 at 15 01 50

### DIFF
--- a/panel/src/components/Layout/Card.vue
+++ b/panel/src/components/Layout/Card.vue
@@ -18,11 +18,10 @@
     </component>
 
     <nav class="k-card-options">
-      <k-button
+      <k-status-icon
         v-if="flag"
         v-bind="flag"
         class="k-card-options-button"
-        @click="flag.click"
       />
       <slot name="options">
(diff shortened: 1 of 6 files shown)


bastianallgeier:
What needs to be done to move this out of draft mode?

distantnative:
The other icons PR, cause we need those icons.

However, I just thought this morning that the other PR is taking the wrong direction. With Inertia we would rather work with the SVG file not Icons.Vue.

So maybe better to really just add the new icons in both places for the moment.

bastianallgeier:
Yes, I agree! Putting the icons into a Vue component was really just a developer experience fix, but it's blowing up the JS bundle unnecessarily. Would you mind adding them?

distantnative:
@bastianallgeier first merge https://github.com/getkirby/kirby/pull/2811, than this

distantnative:
@bastianallgeier ready

bastianallgeier (approved):

PR #2892, opened 2020-10-10
Date & time fields

distantnative:
- Select inputs have been removed, instead a text input will be parsed by tokenized patterns based on `display` option
- `display` option to define format displayed in Panel, based on dayjs tokens - but we should only promise support on `YYYY`, `YY`, `M` `MM`, `DD`, `d`, `HH`, `hh`, `mm`, `ss`, `a`
- Calendar truly supports `min` and `max` dates
- `step` option and `timestamp` helper support rounding not just to minute, but also year, month, day, hour and second.

**Breaking changes**
- So far the `format` option defined the format how the date is stored in the content file. Now we always store the date as `Y-m-d H:i:s` - which should be no problem for everyone using `->toDate()`. The `format` option has been removed.

## To do's
- [ ] Date field with time: inputs are aligned with so much space between
- [x] Inline validations missing for `min` and `max` (or leave them out for the meantime?)
- [x] ~Date field with time. correct value gets emitted from date field (with time), ends up in content file without time~

## Implements
- https://kirby.nolt.io/96
- https://kirby.nolt.io/72
- https://kirby.nolt.io/71

### DIFF
--- a/config/fields/date.php
+++ b/config/fields/date.php
@@ -35,13 +35,13 @@ return [
         },
 
         /**
-         * Youngest date, which can be selected/saved (Y-m-d)
+         * Latest date, which can be selected/saved (Y-m-d)
          */
         'max' => function (string $max = null) {
             return $this->toDatetime($max);
         },
         /**
-         * Oldest date, which can be selected/saved (Y-m-d)
+         * Earliest date, which can be selected/saved (Y-m-d)
          */
         'min' => function (string $min = null) {
             return $this->toDatetime($min);
--- a/config/fields/date.php
+++ b/config/fields/date.php
@@ -1,21 +1,30 @@
 <?php
 
 use Kirby\Exception\Exception;
+use Kirby\Toolkit\Str;
 
 return [
+    'mixins' => ['datetime'],
     'props' => [
+        /**
+         * Unset inherited props
+         */
+        'placeholder' => null,
+
+
         /**
          * Default date when a new page/file/user gets created
          */
-        'default' => function ($default = null) {
+        'default' => function (string $default = null) {
             return $default;
         },
 
         /**
-         * Defines a custom format that is used when the field is saved
+         * Custom format (dayjs tokens) that is used to display
+         * the field in the Panel
          */
-        'format' => function (string $format = null) {
-            return $format;
+        'display' => function (string $display = 'YYYY-MM-DD') {
+            return $display;
         },
 
         /**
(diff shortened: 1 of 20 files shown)


afbora:
You may still be working on some issues, I'm not quite sure.  But I'm sharing the issues I see. 
You specified min-max as validation in progress, but I thought the issues i saw about min-max weren't validation related. If I got it wrong, it means it's okay.

I'll take a look at the timezone issue separately.

1. ~Calendar header working on click calendar icon, but not for click on date directly~ (working, great! The last commits were missing)
2. ~When min set 2000 or newer, getting error alert: `t.$el.querySelector is not a function`~
3. ~For ex: when set max 2030, should be included 2030 i think, but current end of range is 2029 (meaning <=)~
4. ~For ex: when set max 2021, you can't select, OK. But you can go to next years from calendar arrow navigations.~
5. ~For ex: when set min 1992, cant select choose the previous month and day for 1992.~
6. ~Time select would be great~ (will be later version)

_I wrongly closed the PR, sorry._

afbora:
I should have write full dates instead of just year for `min` and `max` options. So everything looks great 🎉

bastianallgeier:
@distantnative can we already merge this or do you want to fix the todos first? We could also move them to separate PRs if this helps to get this done.

distantnative:
@bastianallgeier I think with the alignment I'll need your wisdom.

Validation: probably need to add it since we won't remove inline validation in 3.5, will we? And then not having it would be a regression, so better to add it in this PR first (so we don't merge things that introduce regressions).

bastianallgeier:
@distantnative ok, I need to finish this blocks monster first and then I will look into the alignment part. Could you take care about validation?

distantnative:
I think I managed to add proper `min` and `max` inline validations for date and date time.
@afbora Always appreciate if you throw all your bug-finding skills at it.

I am not too happy with the name of the new `display` option. Format would be much nicer, but that will break with old setups unfortunately.

distantnative:
Ok I think I'm really done from my side so far.
Both feel free to take it for a test ride.

afbora:
I can no longer find errors. This is great news 😅 and PR looks great 🎉 

---

Btw i guess, min-max options don't include `time`, right? For ex: `max: 2020-10-26 15:00`

distantnative:
@afbora 
Now they can properly. Of course only makes sense if `time` is activated.

```
date:
  type: date
  default: now
  time: true
  min: 2020-10-01 10:00
  max: 2020-10-31 21:00
```

How it works actually is that min/max get compared at the level of the step unit.
With this PR, `step` can be more than a simple number for minutes, it can be

```
step:
  size: 2
  unit: hour
```

With that, we have the hours increase/decrease/rounded in 2-hour steps.
And also min/max will then evaluate up to the hour level.

trych:
I am having an issue with the time field (I am using the `blocks` branch, which seems to include the date changes already).
I set a time and am able to save it just fine (img 1), but then when I reload the panel page, the current date is listed together with the time (img 2) and when I try to edit it, the current year of the date (2020), seems to be loaded in the hour field somehow (img 3) and the original time seems to be gone.

20201101-160337_Screenshot_GoogleChrome

20201101-160346_Screenshot_GoogleChrome

20201101-160353_Screenshot_GoogleChrome

I am using these fields within a structure field, but I don't know if this is related to the issue. Let me know, if I should open a dedicated issue for that, I thought I would put it here first, as the feature is not officially released for now.

distantnative:
@trych a new issue would be great, maybe prefix the title with 3.5-beta or so.

And your blueprint would be great. Cause I think the structure field could be key in this bug.

trych:
[Done](https://github.com/getkirby/kirby/issues/2911).

PR #2894, opened 2020-10-11
Improvements to `$app->clone()`

### LINKED ISSUE #2555: Updating a page on the fly results in undefined variable page
**Describe the bug**  
When I update a page on the fly, snippets will report `page` and `site` as undefined.

**To Reproduce**  
Controller (simplified)
```php
return function($site, $page, $kirby) {
  if($kirby->request()->is('post') === true){
    $kirby->impersonate('kirby');
    $participants = $page->participants()->yaml();

    $participants[] = [
      'name' => get('name'),
      'email' => get('email'),
    ];

    $page = $page->update([
      'participants' => $participants
    ]);
  }
};
```

Template:
```php
snippet('header');
```

Snippet header:
```php
<!DOCTYPE html>
<html data-template="<?= $page->template() ?>">
```

**Expected behavior**  
The page updates itself on the fly and renders correctly.

**Screenshots**  
Google Chrome 2020-04-03 at 11 56 01

**Kirby Version**  
3.35

lukasbestle:
### Features

- New `$setInstance` argument for `new App()` that allows to create a local Kirby instance without setting it globally

### Bug fixes

- The template data is now copied over to the new `$app` object when cloning the object
- Fixed an "Undefined variable" error in snippets when a string in Kirby's query language was used as the page numbering template

## Missing discussion

My first thought was to set the `$setInstance` argument to `false` by default (meaning: Cloning the `$app` object would by default not set it globally). This would probably be the cleaner solution, but unfortunately hundreds of unit tests failed with this change because a lot of our code currently relies on the `App::instance()` method and therefore on global state.

My current implementation fixes the specific bug in #2555, but doesn't fix the general issue of breaking snippets when the app object is cloned from a controller. I searched the code for other places where the app object is cloned and didn't find any, but of course plugin code could still do this and would cause the same bug.

The question is: Should we merge this as is to fix the specific bug or should we rather rewrite the places in Kirby that rely on the global app object to fix it generally?

My opinion is that we should merge this for now and reconsider the global app object issue (together with the immutable object issue) in Kirby 4.

## Related issues

- Fixes #2555

Thanks a lot to @afbora for the amazing debugging insights that helped a lot tracking down the source of the issue.

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -79,8 +79,9 @@ class App
      * Creates a new App instance
      *
      * @param array $props
+     * @param bool $setInstance If false, the instance won't be set globally
      */
-    public function __construct(array $props = [])
+    public function __construct(array $props = [], bool $setInstance = true)
     {
         // register all roots to be able to load stuff afterwards
         $this->bakeRoots($props['roots'] ?? []);
@@ -110,7 +111,9 @@ class App
         ]);
 
         // set the singleton
-        Model::$kirby = static::$instance = $this;
+        if (static::$instance === null || $setInstance === true) {
+            Model::$kirby = static::$instance = $this;
+        }
 
         // setup the I18n class with the translation loader
         $this->i18n();
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -337,6 +337,24 @@ class App
     }
 
     /**
+     * Creates an instance with the same
+     * initial properties
+     *
+     * @param array $props
+     * @param bool $setInstance If false, the instance won't be set globally
+     * @return self
+     */
+    public function clone(array $props = [], bool $setInstance = true)
+    {
+        $props = array_replace_recursive($this->propertyData, $props);
+
+        $clone = new static($props, $setInstance);
+        $clone->data = $this->data;
+
+        return $clone;
+    }
+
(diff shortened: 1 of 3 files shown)


nilshoerrmann:
> I searched the code for other places where the app object is cloned and didn't find any, but of course plugin code could still do this and would cause the same bug.

Would it be possible to provide documention for plugin devs how to prevent this issue on their end?

lukasbestle:
> Would it be possible to provide documention for plugin devs how to prevent this issue on their end?

The solution is "don't use `$app->clone()` and if you absolutely have to, set the second argument to `false` to get a local clone".

The question is where we could put this information in the docs so that plugin devs can find it. We could definitely add it to the reference of `$app->clone()`.

bastianallgeier:
TBH, I avoided reviewing this PR for a long time, because this is just such a mind-f***

I think the way you handled it is great. I don't feel good about completely changing the entire way we handle this right now and I would also redo this concept in v4 – whenever that will be. 

I don't think that we will have a lot of plugins that clone the app instance anyway and adding an additional notice to the docs sounds good to me.

bastianallgeier (approved):

PR #2895, opened 2020-10-11
Locale option: Support for all locale formats (array/string)

### LINKED ISSUE #2893: Locale option doesn't work if array and key string for single language
**Describe the bug**  
`locale` option doesn't work if array and key string for single language.

I don't think this is a bug. Because it works without string (`LC_ALL` constant works great). This place doesn't seem to work because we support strings (with auto convert) in the other part that `Language` class. We can also enhance this part.

**To Reproduce** 

Single language app

````php
// /site/config.php
'locale' => [
    'LC_ALL' => 'tr'
]
````

But same `locale` option works for multilingual app

````php
// /site/languages/tr.php
'locale' => [
    'LC_ALL' => 'tr'
]
````

**Screenshots**  
screen-capture-384-Kirby CMS Debugger-localhost

**Kirby Version**  
3.3.3+ & 3.4.x

### PR DESCRIPTION
**Pull request by lukasbestle**
## Describe the PR

The `locale` option now supports setting the locale with named string keys (like `['LC_ALL' => 'en_US']`), which has previously been possible in language files already.

## Related issues

- Fixes #2893 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/AppTranslations.php
+++ b/src/Cms/AppTranslations.php
@@ -4,6 +4,7 @@ namespace Kirby\Cms;
 
 use Kirby\Data\Data;
 use Kirby\Toolkit\I18n;
+use Kirby\Toolkit\Locale;
 use Kirby\Toolkit\Str;
 
 /**
@@ -119,18 +120,13 @@ trait AppTranslations
     /**
      * Set locale settings
      *
-     * @internal
+     * @deprecated 3.5.0 Use \Kirby\Toolkit\Locale::set() instead
+     *
      * @param string|array $locale
      */
     public function setLocale($locale): void
     {
-        if (is_array($locale) === true) {
-            foreach ($locale as $key => $value) {
-                setlocale($key, $value);
-            }
-        } else {
-            setlocale(LC_ALL, $locale);
-        }
+        Locale::set($locale);
     }
 
     /**
--- a/src/Cms/AppTranslations.php
+++ b/src/Cms/AppTranslations.php
@@ -85,7 +85,7 @@ trait AppTranslations
     public function setCurrentLanguage(string $languageCode = null)
     {
         if ($this->multilang() === false) {
-            $this->setLocale($this->option('locale', 'en_US.utf-8'));
+            Locale::set($this->option('locale', 'en_US.utf-8'));
             return $this->language = null;
         }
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/Cms/AppTranslations.php [hunk 1]
```diff
@@ -119,18 +120,13 @@ public function setCurrentTranslation(string $translationCode = null): void
...
     public function setLocale($locale): void
     {
-        if (is_array($locale) === true) {
-            foreach ($locale as $key => $value) {
-                setlocale($key, $value);
-            }
-        } else {
-            setlocale(LC_ALL, $locale);
-        }
+        Locale::set($locale);
```
distantnative, line 129: To be honest, I wouldn't deprecate this one. It's an unnecessary breaking change. I'd leave the method and just call `Locale::set()` inside as you do. And then I also would leave the user calls in this class/trait to `$this->setLocale()`. In that way we only have to change something in one place (if we have ever changes again here), instead of three places.

distantnative:
I added one comment. Other than that it looks good to me. I like bundling this into it's own Toolkit class!

src/Cms/AppTranslations.php
lukasbestle: A deprecation isn't a breaking-change though. I also wouldn't add a deprecation warning to this, it's just a reminder for us to clean this up at some point.

In my opinion, keeping the method around forever will be confusing, especially given that the `App` class has so many methods already.

distantnative: Well, eventually it is a breaking change - deprecating in the end means we will remove it at some point. But yes, at the end it's probably a matter of taste whether to reduce the number of methods on the `App` class here.

bastianallgeier:
@lukasbestle @distantnative what's the state of this? Can we merge it?

lukasbestle:
It's waiting for final review. If you approve it, feel free to merge it. :)

src/Cms/AppTranslations.php
bastianallgeier: I think it's good to keep it as a reminder that we shouldn't use it anymore. Nothing stops us from keeping it around until 4.0

bastianallgeier (approved):
I really like that this is such a simple class now.

PR #2899, opened 2020-10-17
Page create dialog: hide single template select

distantnative:
## Describe the PR

**Page create dialog:**
- Template select is hidden if only one option available (except in debug mode)

Screen Shot 2020-10-17 at 15 10 11

## Related issues

- https://kirby.nolt.io/103


afbora:
Can we change the `Create {template} draft` phrase to make it clearer and match with singular, plural (may be in different languages) or varied template names? 🤔 

For example: `Create Default draft`
Could this statement be confusing for the end user? I'm not sure.

Just an idea: `Create draft: {template}`

distantnative:
To be honest, for me `Create draft: Project` would be more confusing, as it's missing the grammatical connection that the draft is of type `Project`.

But - I think that really calls for more perspectives, opinions etc. to find the ideal way.

afbora:
@distantnative I understand very well. So my english is really bad 😅 

Actually, what I wanted to explain was to highlight the template name a little more. It can also be to enclose the template name in bold or quotes.

Other than that, the dynamic label feature is both minimal and cool ❤️

distantnative:
I think it's more about taste and personal opinion than what is factually right or wrong, better or worse English etc. So we need more opinions and ideas to find something that works best for most people :)

nilshoerrmann:
Question: is is important to the user to actually know the template name at all? The users know that they'd like to create a new child in the current context. So couldn't it just read "Create draft"?

bastianallgeier:
I agree with @nilshoerrmann here. I would keep it simple in this case. The biggest problem I see is the button length, which is suddenly totally out of control.

distantnative:
Ok convinced, removed that part - so the PR is really just about hiding the template select if only one option is available.

bastianallgeier (approved):

PR #2906, opened 2020-10-29
dayjs: load locale for UI translation (code splitting)

distantnative:
## Status: on hold ⏸
We should first clarify if we switch to `vite` https://github.com/getkirby/kirby/pull/3190

## What it does 
- Adds a new Panel route, so webpack can load additional JS files without knowing the current version hash (files get hashed in their filename itself)
- For the current UI translation the dayjs locale is loaded (code splitting) and set
- Calendar works now with months and day names from this locale info
- Calendar shows first day of the week according to locale info

## Breaking changes
- Removes unused translation strings (though... plugins might rely on them?)


bastianallgeier:
What's the state of this one? Do you think we can get it done for 3.5?

distantnative:
I'll clean it up over the weekend a little more - I think functional-wise it's ready.

distantnative:
@bastianallgeier ready. I also removed the list files, so it's easier for you to read and understand all parts.

bastianallgeier:
@distantnative It's looking good. Just one question. Do you know if the chunks are still cached properly once they went through the PHP route? We can still optimise this later if it doesn't work yet. We just have to be aware of it.

distantnative:
@bastianallgeier I think so, since web pack adds the hashes for caching. But I would need to test exactly whether it does work.

How about we move this to 3.5.1 and maybe add in a bit more code splitting of other things as well?

bastianallgeier:
@distantnative yes, that sounds like a good plan.

distantnative:
This isn't as easily achievable with vita anymore (well, let's be honest, it was far from easy with vie-cli already).
Closing it until I have found a new solution.

PR #2908, opened 2020-11-01
Prefix core session keys with kirby.

lukasbestle:
For the upcoming password reset/2FA login feature, we need additional session keys in the core. This is a good opportunity to rename the existing session keys with a prefix that will separate them from custom session data.


bastianallgeier (approved):

bastianallgeier:
That's a lot cleaner!

PR #2909, opened 2020-11-01
Email: Allow body as fallback for template

lukasbestle:
## Describe the PR

It is now possible to set both a `body` and a `template` for `$kirby->email()`. The `body` becomes the fallback in case the snippet that was defined with `template` does not exist.

*Note:* This change is needed for the password reset feature as it will be possible to create completely custom email templates. If the custom email template does not exist, Kirby will use the standard text that comes from a translation variable.

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/Email.php
+++ b/src/Cms/Email.php
@@ -126,7 +126,7 @@ class Email
                 // fallback to single email text template
             } elseif ($text->exists()) {
                 $this->props['body'] = $text->render($data);
-            } else {
+            } elseif (isset($this->props['body']) !== true) {
                 throw new NotFoundException('The email template "' . $this->props['template'] . '" cannot be found');
             }
         }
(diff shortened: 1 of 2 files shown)


distantnative:
Wondering if that could become too confusing for people facing two options that kinda do the same with names that are kinda referring to the same.

What would speak against simply defining the template you need for the email reset feature and when a developer adds his own this default gets overwritten. That would be the same as we do it for block snippets then.

lukasbestle:
> @distantnative What would speak against simply defining the template you need for the email reset feature and when a developer adds his own this default gets overwritten.

As I wrote, the text comes from a translation variable that would be translated via Transifex.

What we could do: Register a snippet that in turn calls `I18n::translate()`. Sounds a bit like a hack to me, but it would work if we decide we don‘t want to merge this PR.

distantnative:
For me, that would be the cleaner and easier to understand approach: we have a default template, the template uses a translation string.

Having a second option for the same purpose (content of the email) that then still works in a different way seems more a hack to me.

lukasbestle:
Alright! As far as I can tell, we currently don't have any core snippets, so I will look into extending the core to recognize these.

PR #2910, opened 2020-11-02
New $user->nameOrEmail() method

lukasbestle:
## Describe the PR

New `$user->nameOrEmail()` method that returns the field for the user's name or, if empty, their email address. The method can be used when the user's identity needs to be printed, but the optional name field may not be filled.

*Note:* Requirement for the password reset feature.

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/User.php
+++ b/src/Cms/User.php
@@ -540,6 +540,18 @@ class User extends ModelWithContent
     }
 
     /**
+     * Returns the user's name or,
+     * if empty, the email address
+     *
+     * @return \Kirby\Cms\Field
+     */
+    public function nameOrEmail()
+    {
+        $name = $this->name();
+        return $name->isNotEmpty() ? $name : new Field($this, 'email', $this->email());
+    }
+
+    /**
      * Create a dummy nobody
      *
      * @internal
--- a/tests/Cms/Users/UserTest.php
+++ b/tests/Cms/Users/UserTest.php
@@ -90,6 +90,25 @@ class UserTest extends TestCase
         $this->assertEquals('Homer Simpson', $user->name());
     }
 
+    public function testNameOrEmail()
+    {
+        $user = new User([
+            'email' => $email = 'homer@simpsons.com',
+            'name'  => $name = 'Homer Simpson',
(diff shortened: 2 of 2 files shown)


bastianallgeier (commented):
`User::username()` already does the exact same thing :)

lukasbestle:
Oh, interesting! I looked through the method list in VS Code and didn't realize that it would do that. Do we have it for backwards-compatibility with v2 or where does the name come from?

There is also one difference: The `username()` method returns just the value, while the `nameOrEmail()` method returns a field that includes the information which type of value it contains.

I think that returning a field isn't as important, so what if we create an alias `nameOrEmail()` that calls the `username()` method?

bastianallgeier:
Yes, it's for backwards-compatibility but also because it's needed by the panel. I honestly also didn't find a better name :) I like the idea to create an alias for it. nameOrEmail is way more clear than username. The username method also returns a field btw. The or method replaces the field value. It just doesn't replace the key as in your method.

lukasbestle:
Are you sure? It uses `->value()` at the end.

bastianallgeier:
Oh, you are right! I missed that one. What do you think? Do we return a field or a plain value in the new method?

lukasbestle:
If we alias it to `->username()`, we would need to return just a string for backwards-compatibility. Returning a field would be a breaking-change.

The alternative is to add the `nameOrEmail()` method with the code from this PR and deprecate the `username()` method in its docblock (*without* warning for now, we can add a warning in v4).

I'd be fine with either alternative. If you agree with the second alternative, please merge my PR and I will add the deprecation comment to `username()` afterwards.

My tendency is to go with the field as it's more flexible and consistent with the `name()` method, which also returns a field.

bastianallgeier:
Yep, I also tend to return a field here. I probably wouldn't deprecate the username for now. Maybe we could reuse it later to re-enable authentication via username. There have been quite a few requests for that in the past. I think it doesn't hurt to just keep it hanging around.

bastianallgeier (approved):

PR #2912, opened 2020-11-02
Support for system templates

lukasbestle:
This is in preparation for the core email templates for password reset.

I first had an implementation that dynamically built the list using `Dir::index()`, but then decided to have a fixed list instead to increase performance on every request and also reliability as only the files will be registered that we explicitly added to the list.

I will put the email templates into a new `config/templates` folder, the list will then look something like this:

```php
<?php

return [
    'emails/name-of-the-template' => __DIR__ . '/templates/emails/name-of-the-template.php'
];
```

Also see the discussion in https://github.com/getkirby/kirby/pull/2909.

### DIFF
--- a/config/templates.php
+++ b/config/templates.php
@@ -1,0 +1,7 @@
+<?php
+
+// @codeCoverageIgnoreStart
+return [
+
+];
+// @codeCoverageIgnoreEnd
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):

bastianallgeier (approved):

bastianallgeier:
I think it would be cool at a later point to have a dynamic loader for system fallbacks. I ran into the same issues with default snippets for blocks. It's quite annoying to define them by hand. It would be way cooler to point an auto-loader to directories and if a template or snippet cannot be found in the site folder it will do a direct check in those system folders. This way we don't have to run a Dir::index but we also don't have to hard-code every template, snippet and blueprint.

lukasbestle:
That's a good idea. We would need to rewrite the loading logic though as it currently relies on all system files being registered in `AppPlugins`. Maybe something for v4.

PR #2916, opened 2020-11-04
Reset password view + k-user-info component

lukasbestle:
The `ResetPasswordView` is automatically visited after login using the password reset feature. It is also accessible via URL for plugin use-cases, but not listed in the menu (changing the password is still possible via the account view).

Bildschirmfoto 2020-11-04 um 10 51 49

At the top of the view, it uses the new `k-user-info` component, which can display the name or email of the user as well as the avatar or a user icon (if the user doesn't have an avatar).

Bildschirmfoto 2020-11-04 um 10 51 02
Bildschirmfoto 2020-11-04 um 10 51 40


bastianallgeier (approved):
This is great! I only made a few small design changes to the user info box. As a general advice for any new components, they should always be independent in their design from their surrounding. In this case the user-info component set a padding that was useful for the reset password view, but it's better to leave this up to the parent view or component and keep such specific styles out of the component itself. Other than that the view implementation is super clean! Great job!

PR #2918, opened 2020-11-08
Email: Support global debugging

lukasbestle:
If the email object cannot be controlled by the test as the email is sent by a core method, the debug mode can now be enabled globally, which will capture all sent emails.


bastianallgeier (approved):

bastianallgeier:
I just wanted to merge this, but rebasing is disabled again. @lukasbestle did you change the settings?

PR #2920, opened 2020-11-09
Change throw exception for toggle field on required fields

### LINKED ISSUE #2914: [Panel] A non-admin user cannot create new users because no role can be specified
**Describe the bug**
When trying to create a new user while logged in with a non-admin user who has the permission to create users, no roles appear for selection in the user create dialog. Clicking `Create` even though no role is selected fails due to the following error: `error.user.create.permission`.

**To Reproduce**

1. Setup the [plainkit](https://github.com/getkirby/plainkit)
2. Create a new user role with `users.create` permissions
3. Create a new user with this role and log in to the panel with this user
4. Try to create a new user via the panel

**Expected behavior**
All roles except admin should be selectable for a non-admin user when creating a new user.

**Probable cause**
The method [`Roles->canBeCreated()`](https://github.com/getkirby/kirby/blob/6112f1cb8ebf9c36e8a9f97ae08acc910ddbc792/src/Cms/Roles.php#L53) only returns the admin role if a non-admin user is logged in, because `$newUser->permissions()->can('create')` returns false for all other roles. This happens because of the following code in [`UserPermissions->canCreate()`](https://github.com/getkirby/kirby/blob/6112f1cb8ebf9c36e8a9f97ae08acc910ddbc792/src/Cms/UserPermissions.php#L45)

```php
// users who are not admins cannot create admins
if ($this->model->isAdmin() === false) {
    return false;
}
```

with the following changes everything works as expected

```php
// users who are not admins cannot create admins
if ($this->model->isAdmin() === true) {
    return false;
}
```

This has been changed with d83a9582a0463acd8d13b2e028e60cef55ccf43e. Was this accidental or intended and I missed something?

**Screenshots**

**Kirby Version**
3.4.4

- OS: Windows 10 Pro 19042.572
- Browser: Firefox
- Version: 82.0.2 (64-Bit)

### LINKED ISSUE #2917: Bad validation error message for toggle fields
## Describe the bug
When a `toggle` field is defined as `required: true` and the user tries to save a page with the toggle set to `off`, the page is not saved and the user is shown an odd validation error message.

## To Reproduce
- Grab the starterkit
- Edit the `about.yml` blueprint and add the lines from below
- Open the `About us` page in the panel
- Toggle the new toggle field on and off again
- Click save the page
- The form validation fails, as the only valid state of a toggle in Kirby 3, is the active state.

```yml
      certain:
        label: Are you sure?
        type: toggle
        default: false
        required: true
        text:
          - "No"
          - "Yes"
```

## Expected behavior
There are multiple options to deal with this issue:

- Show a more descriptive validation error, explaining which field it relates to and what value it expects
- Remove the option to set the `toggle` field as required, because what's exactly the point of requiring a toggle field, if the only valid state is to have it set to `on`?
- Change the validation rule, but not exactly sure to what?

## Screenshots
image

## Kirby Version
3.4.4

## Console output
HTTP 400 (Bad Request) response from the API endpoint

```json
{
    "status": "error",
    "message": "Invalid form with errors",
    "code": 400,
    "exception": "Kirby\\Exception\\InvalidArgumentException",
    "key": "error.invalidArgument",
    "file": "\/kirby\/src\/Cms\/ModelWithContent.php",
    "line": 771,
    "details": {
        "certain": {
            "label": "Are you sure?",
            "message": {
                "required": "Invalid argument \"-\" in method \"-\""
            }
        }
    },
    "route": "pages\/([a-zA-Z0-9\\.\\-_%= \\+\\@\\(\\)]+)"
}
```

## Desktop (unrelated)
 - OS: Windows
 - Browser: Chrome
 - Version: 86.0.4240.183

## Additional Information
Related [forum post](https://forum.getkirby.com/t/bad-error-when-saving-required-toggle-field/19950)

afbora:
## Describe the PR

Updated translation key for required field validation message and included quick small fix with reverting a commit.

## Related issues

- Fixes #2917 
- Fixes #2914 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/fields/toggle.php
+++ b/config/fields/toggle.php
@@ -59,9 +59,7 @@ return [
         'boolean',
         'required' => function ($value) {
             if ($this->isRequired() && ($value === false || $this->isEmpty($value))) {
-                throw new InvalidArgumentException([
-                    'key' => 'form.field.required'
-                ]);
+                throw new InvalidArgumentException(I18n::translate('field.required'));
             }
         },
     ]
--- a/i18n/translations/bg.json
+++ b/i18n/translations/bg.json
@@ -72,7 +72,6 @@
   "error.file.type.forbidden": "Не е позволен ъплоуда на файлове от тип {type}",
   "error.file.undefined": "\u0424\u0430\u0439\u043b\u044a\u0442 \u043d\u0435 \u043c\u043e\u0436\u0435 \u0434\u0430 \u0431\u044a\u0434\u0435 \u043d\u0430\u043c\u0435\u0440\u0435\u043d",
 
-  "error.form.field.required": "The field is required",
   "error.form.incomplete": "Моля коригирайте всички грешки във формата...",
   "error.form.notSaved": "Формата не може да бъде запазена",
 
@@ -188,6 +187,7 @@
   "error.validation.time": "Моля въведете валидно време",
   "error.validation.url": "Моля въведете валиден URL",
 
+  "field.required": "The field is required",
   "field.files.empty": "Все още не са избрани файлове",
   "field.pages.empty": "Все още не са избрани страници",
   "field.structure.delete.confirm": "Сигурни ли сте, че искате да изтриете това вписване?",
--- a/i18n/translations/bg.json
+++ b/i18n/translations/bg.json
@@ -72,6 +72,7 @@
   "error.file.type.forbidden": "Не е позволен ъплоуда на файлове от тип {type}",
   "error.file.undefined": "\u0424\u0430\u0439\u043b\u044a\u0442 \u043d\u0435 \u043c\u043e\u0436\u0435 \u0434\u0430 \u0431\u044a\u0434\u0435 \u043d\u0430\u043c\u0435\u0440\u0435\u043d",
 
+  "error.form.field.required": "The field is required",
   "error.form.incomplete": "Моля коригирайте всички грешки във формата...",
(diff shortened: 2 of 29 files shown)


bastianallgeier:
I'm really sorry @afbora. I made a mistake. Your PR is great, but it will lead to a lot of problems with Transifex. We cannot simply reimport the translation files there. The regular workflow is to update the en.json as source file and the Transifex will take care of updating the translations. But when we do that and rename the string, Transifex will remove all translations and replace it with the English version. I haven't thought about that when we talked about it. 

I think the most backward compatible way, would be to leave the translation files as they are and change how the exception is thrown. Sorry :(

afbora:
@bastianallgeier I found this to be the simplest solution. Otherwise, I would have to change the `Exception` class 🤔

bastianallgeier (approved):
Yep, that's the perfect solution.

eXpl0it3r:
Was that change in the `src/Cms/UserPermissions.php` intentional? Looks like it was added due to a commit revert...

afbora:
@eXpl0it3r We rolled back the previous update because it was incorrect. Yes it is intentional.

eXpl0it3r:
Ah okay, I was just confused seeing such an unrelated commit on this PR and since it was permission related, I thought, I'd rather ask. 😄

afbora:
Yes you are right 🙈 , I included it in this PR as it is a small update. It is already written in the issue description.

PR #2927, opened 2020-11-15
Mock usleep() function for faster PHPUnit tests

lukasbestle:
With the new additions to the `Auth` class in #2923, the `usleep()` function gets called a lot in tests.

This PR makes the tests run significantly faster as the sleep times are skipped during tests. This saves around 20 seconds, which is about 40 % of the entire PHPUnit run time.

### DIFF
--- a/tests/Cms/mocks.php
+++ b/tests/Cms/mocks.php
@@ -22,3 +22,19 @@ class MockTime
 {
     public static $time = 1337000000;
 }
+
+/**
+ * Mock for the PHP usleep() function to skip over
+ * waiting times while testing
+ *
+ * @param int $microSeconds
+ * @return void
+ */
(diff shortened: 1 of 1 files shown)


bastianallgeier:
Don't we use usleep in some tests to actually sleep to test file modification differences?

lukasbestle:
I did a global file search on `sleep(` and didn't find other occurrences. I think we replaced that test logic a while ago as it was unreliable.

lukasbestle:
Ah, yes, I remember now: The tests now use `touch()` with a timestamp in the past to fake passing time. As the tests all still pass, I think it's safe to merge this PR.

bastianallgeier:
@lukasbestle ok, perfect!

Issue #2930, opened 2020-11-15
Update npm dependencies and fix ESLint warnings


lukasbestle:
## Fix ESLint warnings

1. Checkout `features/npm-update-dev`
2. `npm update`
3. `node_modules/.bin/vue-cli-service lint` and keep the terminal window open
4. Discard all changes in Tower to avoid conflicts (the automatic fixes and the lockfile update are done when the release is ready)
5. [ ] Manually fix the 35 ESLint warnings from step 3 and commit the fixes to `features/npm-update-dev`

If a warning is not correct, please use an `eslint-disable-next-line <name of the rule>` comment so that the warning is silenced.

## Code dependencies

- [x] `@linusborg/vue-simple-portal`: `0.1.4` (no new release)
- [x] `autosize`: `4.0.2` (no new release)  
- [ ] `dayjs`: `1.9.1` → `1.9.6`
- [ ] `npm`: `6.14.7` → `6.14.8`
- [x] `vue`: `2.6.11` → `2.6.12`
- [ ] `vue-router`: `3.1.6` → `3.4.9`
- [ ] `vuedraggable`: `2.23.2` → `2.24.3`
- [ ] `vuelidate`: `0.6.2` → `0.7.6`
- [ ] `vuex`: `3.1.3` → `3.5.1`
- [ ] `vuex-i18n`: `1.13.1` → `8.22.1`

**Steps for each dependency**:

1. Review the changelogs since the current version for relevant changes
2. Add the new package version to `package.json` with a prepended `^` sign to avoid pinning the version
3. Run `npm update` locally (*don't* commit the lockfile for now because of conflicts)
4. Test that the affected parts of the Panel still work

## Remaining dev dependencies with breaking-changes

- [ ] Upgrade to Cypress 5: [Migration Guide](https://on.cypress.io/migration-guide)

lukasbestle:
Migrated the TODOs to the PR #2937.

PR #2931, opened 2020-11-15
Update npm dev dependencies

lukasbestle:
- Temporarily pinned the code dependency versions to their current versions so that they can be safely updated one by one, see https://github.com/getkirby/kirby/issues/2930.
- Updated all npm dev dependencies (except Cypress) to their latest versions
  - `node-sass` is deprecated and was replaced with `dart-sass` (the `sass` npm package)
  - `babel-core` was renamed to `@babel/core`
  - The upgrades mean that our setup is now compatible with Node v15
- The Vue CLI ESLint setup changed quite a bit. Might be that our old setup was broken or just that the new setup has new rules, but there were suddenly hundreds of errors and warnings. I have run the auto-fixer and also made manual changes to fix the fatal errors. 35 warnings still remain in `vue-cli-service lint` that we should review one by one.

**TODO:**

- [ ] Rebase against `features` when all other PRs are merged
- [ ] Run `npm update`
- [ ] Run `node_modules/.bin/vue-cli-service lint --fix`

### DIFF
--- a/panel/package.json
+++ b/panel/package.json
@@ -16,7 +16,7 @@
     "autosize": "^4.0.2",
     "dayjs": "1.9.1",
     "npm": "6.14.7",
-    "vue": "2.6.11",
+    "vue": "^2.6.12",
     "vue-router": "3.1.6",
     "vuedraggable": "2.23.2",
     "vuelidate": "0.6.2",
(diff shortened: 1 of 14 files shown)


bastianallgeier:
I wonder if we should take the eslint fixes out of this and move them to the very end before we release the first rc for 3.5. Otherwise those linting changes might make it super complicated to merge features because of dozens of conflicts.

lukasbestle:
That's true. In this case we should also wait with the dependency updates as the errors fixed in this PR are fatal errors. Vue CLI won't build without fixing them. So let's wait until all other PRs are merged and then I'll rebase this.

lukasbestle:
Meanwhile we could already start reviewing the other package updates (#2930) and add the changes to this PR. Then we won't have to do that so close to the release. @bastianallgeier @distantnative

lukasbestle:
I have removed the automatic ESLint fixes and the lockfile changes from this PR as these will definitely collect merge conflicts over time. I added a TODO list to the first post of this PR.

bastianallgeier (approved):

PR #2936, opened 2020-11-20
Add similarity search for Str class

afbora:
## Describe the PR

Add similarity search feature for Str class and extracted PHP side from #2794

**Simple usage**

    Str::similarity('foo', 'fooBar', $percent);
    echo $percent; // 66.67%
    
    

**Case Sensitive**

    Str::similarity('Kirby', 'KIRBY', $percent, true);
    echo $percent; // 20%

## Related issues

- Closes getkirby/ideas#140

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -804,6 +804,81 @@ class Str
     }
 
     /**
+     * Calculate the similarity between two strings with multibyte support
+     *
+     * @author Antal Áron
+     * @copyright Copyright (c) 2017, Antal Áron
+     * @license https://github.com/antalaron/mb-similar-text/blob/master/LICENSE MIT License
+     * @param string $first
+     * @param string $second
+     * @param float|null $percent Optional variable passed by reference for the similarity in percent
+     * @param bool $caseSensitive If true, compare strings with case sensitive
+     * @return int Number of matching chars in both strings
+     */
+    public static function similarity(string $first, string $second, float &$percent = null, bool $caseSensitive = false): int
+    {
+        if ($caseSensitive === false) {
+            $first  = static::lower($first);
+            $second = static::lower($second);
+        }
+
+        if (static::length($first) + static::length($second) === 0) {
+            $percent = 0.0;
+
+            return 0;
+        }
+
+        $pos1 = $pos2 = $max = 0;
+        $len1 = static::length($first);
+        $len2 = static::length($second);
+
+        for ($p = 0; $p < $len1; ++$p) {
+            for ($q = 0; $q < $len2; ++$q) {
+                for (
+                    $l = 0;
+                    ($p + $l < $len1) && ($q + $l < $len2) &&
+                    static::substr($first, $p + $l, 1) === static::substr($second, $q + $l, 1);
+                    ++$l
+                ) {
+                    // nothing to do
+                }
+
+                if ($l > $max) {
+                    $max  = $l;
+                    $pos1 = $p;
+                    $pos2 = $q;
+                }
+            }
+        }
+
+        $similarity = $max;
+
+        if ($similarity) {
+            if ($pos1 && $pos2) {
+                $similarity += static::similarity(
+                    static::substr($first, 0, $pos1),
+                    static::substr($second, 0, $pos2),
+                    $percent,
+                    $caseSensitive
+                );
+            }
+
+            if (($pos1 + $max < $len1) && ($pos2 + $max < $len2)) {
+                $similarity += static::similarity(
+                    static::substr($first, $pos1 + $max, $len1 - $pos1 - $max),
+                    static::substr($second, $pos2 + $max, $len2 - $pos2 - $max),
+                    $percent,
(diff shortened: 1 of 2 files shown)


src/Toolkit/Str.php
```diff
@@ -803,6 +803,63 @@ public static function slug(string $string = null, string $separator = null, str
...
+    /**
+     * Calculate the similarity between two strings with multibyte support
+     *
+     * @author Antal Áron
+     * @copyright Copyright (c) 2017, Antal Áron
+     * @license https://github.com/antalaron/mb-similar-text/blob/master/LICENSE MIT License
+     * @param string $first
+     * @param string $second
+     * @param float|null $percent
+     * @return int
```
```suggestion
     * @param float|null $percent Optional variable passed by reference for the similarity in percent
     * @return int Number of matching chars in both strings
```

config/components.php
```diff
@@ -231,6 +232,19 @@
...
+
+                // check for similar matches
+                if ($options['words'] === false && $options['similar'] === true) {
+                    $matches = Str::similarity($lowerQuery, $lowerValue, $similarPercent);
+
+                    if ($matches > 0) {
+                        // calculate score based on percentage
+                        // for example 66.6% result: (66.6 / 10) * $score, final result is (7 * $score)
+                        // the coefficient can be a value between 0 and 10
+                        $item->searchScore += round($similarPercent / 10) * $score;
```
lukasbestle: Have you tested the real-world behavior of the fuzzy search? I.e. does it sort the results in a good and expected order?

In my opinion a multiplier of up to 10 is a bit high compared to the multipliers above. Even an exact beginning match only gets a score of `8 * $score`. I haven't tested this, but my guess would be that the similar matches end up too high in the results list.

lukasbestle (commented):
I really like the clean implementation and only have two comments below.

A general question: The algorithm seems quite expensive with the nested `for` loops. Is it still fast enough for long strings? Imagine a site with hundreds of long blog articles.

config/components.php
afbora: I will do some tests and share the results.

bastianallgeier:
In order to finally get this merged, I would suggest to only use the Str::similarity method and remove the implementation in the Collection search. We can still bring this back later. Str::similarity is a great method and addition on its own.

afbora:
Since I accidentally rebase with develop, I changed the main branch from features to develop. Is it a problem?

bastianallgeier (dismissed):

bastianallgeier:
@afbora no, looks good to me! Thanks for the changes.

bastianallgeier:
BTW: I wouldn't worry about the cost of the algo for long strings. For the search, this is relevant in the end and we should have a closer look at that when working on it, but isolated, the similarity feature is IMHO super useful for the comparison of simple strings – like the "did you mean" feature in Google and it's definitely fine for that already.

bastianallgeier:
@afbora do you think you can fix the drop in code coverage?

afbora:
Since I will look but cannot see the report, I must first authorize.

afbora:
@bastianallgeier I had an idea while I was correcting the coverage error. Now this is case sensitive checking. What if we add a case-sensitive default parameter disabled? Because now `Kirby` and `KIRBY` match 20%. Actually I have completed the development in my local, would you like to have a look?

New parameter like that:

Default disabled

    Str::similarity('Kirby', 'KIRBY', $percent); // %100
    
Enabled

    Str::similarity('Kirby', 'KIRBY', $percent, true); // %20

bastianallgeier:
@afbora sounds good to me to introduce a caseSensitive param and disable it by default. Feel free to add it to the PR.

bastianallgeier (approved):

lukasbestle:
@afbora A bit late to the party, but a suggestion before this lands in the release tomorrow:

I think the `$caseSensitive` param should be before the `$percent` param, so basically the last two params should be swapped. Variables that get assigned a value by reference are most of the time at the end of the param list in PHP as far as I can tell.

afbora:
@lukasbestle I deliberately used the `caseSensitive` parameter at the end. Because if we take the percent parameter forward, we will always have to send the `caseSensitive` parameter to get the `percent` value. Isn't there a more useful way? 😅

lukasbestle:
That's true! I'm not 110 % happy with it, but you are right that it's probably the best compromise.
With PHP 8 it's all going to be a non-issue anyway with named arguments. ;)

afbora:
As a result, should it stay the this same?

Also just an another idea; Does it make sense to return the array?

````php
return [
  'percent' => $percent,
  'similarity' => $similarity
];
````

**Comment by lukasbestle** [comment]
@afbora I like that idea! More Kirby-like and less weird.

While we are at it, what about this:

```php
return [
  'percent' => $percent,
  'matchingChars' => $similarity
];
```

That describes more clearly what the value is supposed to mean.

afbora:
Then I'll create a PR for this. So what would `matches` be like instead of `matchingChars`? 🤔

PR #2937, opened 2020-11-20
ESLint fixes & npm dependency updates

lukasbestle:
## Fix ESLint warnings

1. Checkout `feature/npm-updates`
2. `npm update`
3. `node_modules/.bin/vue-cli-service lint` and keep the terminal window open
4. Discard all changes in Tower to avoid conflicts (the automatic fixes and the lockfile update are done when the release is ready)
5. [x] Manually fix the 35 ESLint warnings from step 3 and commit and push the fixes to this PR
6. [x] Fix remaining 4 ESLint warnings for `v-html`: https://github.com/getkirby/kirby/pull/2937#issuecomment-731585262

If a warning is not correct, please use an `eslint-disable-next-line <name of the rule>` comment so that the warning is silenced.

## Upgrade code dependencies

- [x] `@linusborg/vue-simple-portal`: `0.1.4` (no new release)
- [x] `autosize`: `4.0.2` (no new release)  
- [x] `dayjs`: `1.9.1` → `1.9.6`
- [x] ~~`npm`: `6.14.7` → `6.14.8`~~ (not needed as dependency)
- [x] `vue`: `2.6.11` → `2.6.12`
- [x] `vue-router`: `3.1.6` → `3.4.9`
- [x] `vuedraggable`: `2.23.2` → `2.24.3`
- [x] `vuelidate`: `0.6.2` → `0.7.6`
- [x] `vuex`: `3.1.3` → `3.5.1`
- [x] `vuex-i18n`: `1.13.1` (no new release)

**Steps for each dependency**:

1. Review the changelogs since the current version for relevant changes
2. Add the new package version to `package.json` with a prepended `^` sign to avoid pinning the version
3. Run `npm update` locally (*don't* commit the lockfile for now because of conflicts)
4. Test that the affected parts of the Panel still work
5. Commit and push the change to `package.json` only

## Upgrade dev dependencies

- [x] Build setup and all not listed dev dependencies
- [x] Upgrade to Cypress 5: [Migration Guide](https://on.cypress.io/migration-guide)

## TODOs before merging

- [x] Rebase against `features` when all other PRs are merged
- [x] Run `npm update`
- [x] Run `node_modules/.bin/vue-cli-service lint --fix`


distantnative:
@bastianallgeier @lukasbestle 
30+ of the lint errors are `'v-html' directive can lead to XSS attack` – however, since we want to offer the chance for users to style, e.g. a field's help text, with or add links etc., I think there is no way around `v-html`. I think we should just check again that all contents fro `v-html` come from blueprints and not user input. And then disable that lint rule.

lukasbestle:
@distantnative I agree. However I would disable the rule for each verified line separately so that it can notify us about newly added instances. If we disable it entirely, we might forget about it.

distantnative:
@lukasbestle As much as I dislike those inline comments for parsers/linters... yes, we should do it rather that way.

distantnative:
Your list confuses `vuex-i18n` with `vue-i18n`. We use the first one, which is still stuck at 1.13.1 - but wondering if we should look into switching here since it isn't maintained anymore.

lukasbestle:
> Your list confuses vuex-i18n with vue-i18n. We use the first one, which is still stuck at 1.13.1

Are you sure? The current `package.json` already contains `vuex-i18n`.

lukasbestle:
@distantnative You are right! I mixed that up somehow.

distantnative:
I'd suggest moving a potential switch f libraries to 3.5.1 - it's not 100% trivial. So, I'd mark the library as "updated" (still same version) on this PR.

distantnative:
There are indeed a few usages of `v-html` that could be problematic. In all those fields where we use it to render options that could come from the blueprint, but also could come from an uncontrolled API. 
I know we use it in the multi select field cause we want to highlight the matches of the search in the option strings. I don't quite understand why we use it checkboxes field or so (`v-text` should work there as well cause I don't think we actually support HTML formatting in the option label).
Or do you both think that it's fine to use it here as well?

lukasbestle:
@distantnative Could you please post the list of the remaining warnings so we can take a look at them together in detail?

distantnative:
Remaining lint errors:

```
warning: 'v-html' directive can lead to XSS attack (vue/no-v-html) at src/components/Forms/Input/CheckboxInput.vue:27:42:
  25 |       </svg>
  26 |     </span>
> 27 |     <span class="k-checkbox-input-label" v-html="label" />
     |                                          ^
  28 |   </label>
  29 | </template>
  30 | 


warning: 'v-html' directive can lead to XSS attack (vue/no-v-html) at src/components/Forms/Input/MultiselectInput.vue:59:17:
  57 |           @keydown.native.space.prevent.stop="select(option)"
  58 |         >
> 59 |           <span v-html="option.display" />
     |                 ^
  60 |           <span class="k-multiselect-value" v-html="option.info" />
  61 |         </k-dropdown-item>
  62 | 


warning: 'v-html' directive can lead to XSS attack (vue/no-v-html) at src/components/Forms/Input/MultiselectInput.vue:60:45:
  58 |         >
  59 |           <span v-html="option.display" />
> 60 |           <span class="k-multiselect-value" v-html="option.info" />
     |                                             ^
  61 |         </k-dropdown-item>
  62 | 
  63 |         <k-dropdown-item


warning: 'v-html' directive can lead to XSS attack (vue/no-v-html) at src/components/Forms/Input/ToggleInput.vue:12:40:
  10 |       @change="onInput($event.target.checked)"
  11 |     >
> 12 |     <span class="k-toggle-input-label" v-html="label" />
     |                                        ^
  13 |   </label>
  14 | </template>
  15 | 
```

I'm not sure if Checkboxes and Toggle really need `v-html` and not just `v-text` (or `{{ }}`). 
Multiselect I know why we did it (wrapping search matches with `<b></b>`) - but maybe we need to make sure no other HTML is injected first from an API.

lukasbestle:
I fully agree! `v-text` for the labels is very likely fine and the multiselect results should filter out HTML before adding in the `<b>` tags.

lukasbestle:
@distantnative What do we use `npm` for as a code dependency? I don't see any mention of it in our code.

distantnative:
@lukasbestle i asked myself the same thing earlier, likely just a install on it should actually be in devDependencies?

Edit: or maybe not a dependency at all - it's dependency globally to run it, but not for the project itself a dependency.

distantnative:
@lukasbestle I think I got all the updates in - so I guess we'll wait for all other 3.5 PRs to be merged first?

lukasbestle:
Awesome! Yes, that would be the next step.

PR #2945, opened 2020-11-23
Handle content case-insensitively

### LINKED ISSUE #2577: Untranslatable fields are overly sensitive to Blueprint field's key capitalization

### LINKED ISSUE #2789: Editing title before translation file has been created causes untranslatable fields to be copied to translation file
**Describe the bug**  
When defining field names with uppercase letters in a blueprint and setting `translate: false` for given field, it will be saved to the translation file, if the page title was edited before any other content has been edited for a translation.

**To Reproduce**  

1. Use a Kirby installation with at least 2 languages (DE and EN in this example). Include the non-translatable fields in a blueprint (`pages/default.yml`), e.g.:

```yaml
  datestart: # <= will work as expected
    type: date
    label: Start date
    translate: false
  dateEnd: # <= will be copied to translation file
    type: date
    label: End date
    translate: false
```

(BTW: The error is not limited to the `date` field.)

2. Create a new page with this blueprint while having the default language (DE) selected in the panel.
3. Fill out the fields `datestart` and `dateEnd`.
4. Make sure, that Kirby has not created a translation file `default.en.txt` yet. If you have not worked on the English translation before, the file should not exist.
5. Switch to EN translation or the page.
6. Click the page title and change it to some arbitrary value
7. Review the newly-created translation file `default.en.txt`: The value of `dateEnd` is copied to the English translation (unexpected behavior), whereas the value of `datestart` is not copied (expected behavior).

**Expected behavior**  
Fields which use the `translate: false` option, should never be copied to a translation file, because they are not editable from the panel any longer after that and values in the translation file are also never updated, if field values change in the primary language. This can easily lead to inconsistent content files.

**Kirby Version**  
3.4.2

 - OS: macOS 10.15.6
 - Browser: Mozilla Firefox
 - Version: 79.0

### LINKED ISSUE #4142: Fields in mixed case in virtual pages can't be searched
## Description

Suppose you have a virtual pages setup like described on [this page](https://getkirby.com/docs/guide/virtual-pages/content-from-api) with these fields:

```
[...]
$pages[] = [
                'slug'     => Str::slug($review->display_title),
                'num'      => $key+1,
                'template' => 'review',
                'model'    => 'review',
                'content'  => [
                    'title'    => $review->display_title,
                    'headline' => $review->headline,
                    'linkText' => $review->link->suggested_link_text,
                    'FooBar'   => $review->foobar,
                ]
            ];
[...]
```
If you now run a search for any field with uppercase letters (`linkText` or `FooBar`), e.g.

```
$site->search("foobar", "FooBar");
```

or even

```
$site->search("foobar", [
    [ 'fields' => ['FooBar'], ]
  ]);
```
it does not deliver any results. The only way to search for these fields is to leave the search fields empty.

The reason being [`config/components.php:215`](https://github.com/getkirby/kirby/blob/main/config/components.php#L220) which states `$fields = array_map('strtolower', $options['fields']);`, ie. that converts that field to `foobar` which then of course does not exist in the `$keys` array (still containing `FooBar`). The following array intersection doesn't yield any results of course.

Interestingly this does *not* happen with regular pages, where you can use any case in your field names.

**Expected behavior**  
 Virtual page data fields/keys should be lowercased either in seach component or should be normalized before. 

## Your setup
**Kirby Version**  
3.6.2

lukasbestle:
### Refactoring

- Content field names are now always handled case-insensitively.

### Fixes
 
- Fixed a bug where field content was copied to secondary languages even though `translate: false` was set #2577 #2789
- Fixed a bug where fields with uppercase field names in virtual pages could not be searched #4142 

### Breaking changes

- Field names of virtual pages' content are now converted to lowercase. If your virtual page has two fields that only differ in capitalization, only the last defined one will be available.

## Additional notes

I've also rewritten the tests for the `Content` class to test more edge-cases. The old/previous tests also pass, so I hope that we haven't broken anything else. The content fields were mostly case-insensitive anyway, now they are reliably handled that way in all methods.

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/Content.php
+++ b/src/Cms/Content.php
@@ -60,9 +60,14 @@ class Content
      *
      * @param array|null $data
      * @param object|null $parent
+     * @param bool $normalize Set to `false` if the input field keys are already lowercase
      */
-    public function __construct(array $data = [], $parent = null)
+    public function __construct(array $data = [], $parent = null, bool $normalize = true)
     {
+        if ($normalize === true) {
+            $data = array_change_key_case($data, CASE_LOWER);
+        }
+
         $this->data   = $data;
         $this->parent = $parent;
     }
@@ -163,9 +168,7 @@ class Content
             return $this->fields[$key];
         }
 
-        // fetch the value no matter the case
-        $data  = $this->data();
-        $value = $data[$key] ?? array_change_key_case($data)[$key] ?? null;
+        $value = $this->data()[$key] ?? null;
 
         return $this->fields[$key] = new Field($this->parent, $key, $value);
     }
@@ -178,10 +181,7 @@ class Content
      */
     public function has(string $key): bool
     {
-        $key  = strtolower($key);
-        $data = array_change_key_case($this->data);
-
-        return isset($data[$key]) === true;
+        return isset($this->data[strtolower($key)]) === true;
     }
 
     /**
@@ -208,7 +208,7 @@ class Content
         $copy->fields = null;
 
         foreach ($keys as $key) {
-            unset($copy->data[$key]);
+            unset($copy->data[strtolower($key)]);
         }
 
         return $copy;
@@ -258,7 +258,8 @@ class Content
      */
     public function update(array $content = null, bool $overwrite = false)
     {
-        $this->data = $overwrite === true ? (array)$content : array_merge($this->data, (array)$content);
+        $content = array_change_key_case((array)$content, CASE_LOWER);
+        $this->data = $overwrite === true ? $content : array_merge($this->data, $content);
 
         // clear cache of Field objects
         $this->fields = [];
--- a/src/Cms/ContentTranslation.php
+++ b/src/Cms/ContentTranslation.php
@@ -177,7 +177,12 @@ class ContentTranslation
      */
     protected function setContent(array $content = null)
     {
-        $this->content = $content;
+        if ($content !== null) {
+            $this->content = array_change_key_case($content);
+        } else {
+            $this->content = null;
+        }
+
         return $this;
     }
 
@@ -220,7 +225,8 @@ class ContentTranslation
      */
     public function update(array $data = null, bool $overwrite = false)
     {
-        $this->content = $overwrite === true ? (array)$data : array_merge($this->content(), (array)$data);
+        $data = array_change_key_case((array)$data);
+        $this->content = $overwrite === true ? $data : array_merge($this->content(), $data);
         return $this;
     }
 
--- a/src/Cms/ModelWithContent.php
+++ b/src/Cms/ModelWithContent.php
@@ -90,7 +90,8 @@ abstract class ModelWithContent extends Model
                 return $this->content;
             }
 
-            return $this->setContent($this->readContent())->content;
+            // don't normalize field keys (already handled by the `Data` class)
+            return $this->content = new Content($this->readContent(), $this, false);
 
         // multi language support
         } else {
@@ -102,7 +103,8 @@ abstract class ModelWithContent extends Model
 
             // get the translation by code
             if ($translation = $this->translation($languageCode)) {
-                $content = new Content($translation->content(), $this);
+                // don't normalize field keys (already handled by the `ContentTranslation` class)
+                $content = new Content($translation->content(), $this, false);
             } else {
                 throw new InvalidArgumentException('Invalid language: ' . $languageCode);
             }
@@ -453,7 +455,7 @@ abstract class ModelWithContent extends Model
         if ($languageCode !== $kirby->defaultLanguage()->code()) {
             foreach ($this->blueprint()->fields() as $field) {
                 if (($field['translate'] ?? true) === false) {
-                    $content[$field['name']] = null;
+                    $content[strtolower($field['name'])] = null;
                 }
             }
 
--- a/tests/Cms/Content/ContentTest.php
+++ b/tests/Cms/Content/ContentTest.php
@@ -2,160 +2,203 @@
 
 namespace Kirby\Cms;
 
+/**
+ * @coversDefaultClass Kirby\Cms\Content
+ */
 class ContentTest extends TestCase
 {
-    protected function mockData(): array
+    protected $content;
+    protected $parent;
+
+    public function setUp(): void
     {
-        return [
-            'title' => 'Test Content',
-            'text'  => 'Lorem ipsum'
-        ];
+        $this->parent  = new Page(['slug' => 'test']);
+        $this->content = new Content([
+            'a' => 'A',
+            'B' => 'B',
+            'MiXeD' => 'mixed',
+            'mIXeD' => 'MIXED'
+        ], $this->parent);
     }
 
-    protected function mockObject()
+    /**
+     * @covers ::__call
+     */
+    public function testCall()
     {
-        return new Content($this->mockData());
+        $this->assertSame('a', $this->content->a()->key());
+        $this->assertSame('A', $this->content->a()->value());
+        $this->assertSame('mixed', $this->content->mixed()->key());
+        $this->assertSame('MIXED', $this->content->mixed()->value());
+        $this->assertSame('mixed', $this->content->mIXEd()->key());
+        $this->assertSame('MIXED', $this->content->mIXEd()->value());
     }
 
+    /**
+     * @covers ::__construct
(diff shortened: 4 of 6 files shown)


bastianallgeier:
I really would love to fix this in 3.5, but I think we need more time for this. We have to keep in mind that field fetching is super critical in Kirby. This happens in large collection filters, in sorting, basically everywhere and we should be very thoughtful about any performance hits. Let's move this to another release.

lukasbestle:
@bastianallgeier Performance is actually improved by the change as the key case is only converted once when updating the content and not every time when getting it like before. But it might cause regressions, so it's a good idea to test it some more.

I have added the PR and the two issues to the 3.6.0 milestone to make sure that we won't forget about it. The underlying issue has been reported several times already, so we should solve it as soon as possible. But not before 3.6.0 as such sensitive changes shouldn't be in patch releases.

distantnative:
Is it really a breaking change or could we include it in 3.5.1?
Weren't we always saying content names are case-insensitive. So the current state was more like a bug?

lukasbestle:
As I wrote, it's a very sensitive issue with a considerable risk of a regression. In theory it should be fine and my tests tell the same, but IMO it's better to avoid potential regressions in patch releases. We will only know afterwards whether this causes any.

src/Cms/Content.php
```diff
@@ -256,7 +251,8 @@ public function toArray(): array
      */
     public function update(array $content = null, bool $overwrite = false)
     {
-        $this->data = $overwrite === true ? (array)$content : array_merge($this->data, (array)$content);
+        $content = array_change_key_case((array)$content);
```
distantnative: Does it need the `(array)` casting here? Wouldn't it be easier to turn the parameter then into `array $content = []`?

src/Cms/ContentTranslation.php
```diff
@@ -177,7 +177,12 @@ protected function setCode(string $code)
      */
     protected function setContent(array $content = null)
     {
-        $this->content = $content;
+        if ($content !== null) {
+            $this->content = array_change_key_case((array)$content);
```
distantnative: Similarly here, if `$content` isn't `null`, it has to be already an array based on the parameter type hinting, right?

src/Cms/Content.php
lukasbestle: That would be a breaking-change as you could previously explicitly pass `null`. I wonder why we allowed `null` in the first place as it looks like a pretty deliberate and explicit feature.

@bastianallgeier Do you remember?

src/Cms/ContentTranslation.php
lukasbestle: In this case, yes!

src/Cms/Content.php
bastianallgeier: I have no idea why the default is null. It makes no sense while looking at it. Do we break any of our tests when we change it? I really wouldn't mind going for the more useful default here.

lukasbestle: Maybe it was another case of dealing with weird user input (e.g. from `get()`). But in this case I think it's pretty unlikely that anyone might have used it. It would still be a breaking change.

If I remember correctly, our tests would still pass, but of course that's not a guarantee. What do you think: Should we risk it?

bastianallgeier: Would it be a big deal just to keep it as it is to avoid yet another breaking change?

bastianallgeier: We keep stacking them up all over the place. On the other hand it's another "a lot of pain once vs. smaller portions of pains in the next releases" kind of situation.

distantnative: I'd ask ourselves how much this change helps us? How much delaying this change until 3.7 would block us in other aspects? And on that basis decide. If it helps us to move forward on other aspects and features, let's already include it. If it's unrelated and should just be changed eventually, maybe it can wait until 3.7

distantnative:
@bastianallgeier @lukasbestle 
Let's come back to this. It is a breaking change, but how big?
Do I understand it correctly that the breaking change is only in a scenario where one content file would host two fields with the same name where only the capitalisation differs e.g. `text` vs. `Text`?

I would say that shouldn't be allowed, supported and is a breaking change that we should do now.

src/Cms/ContentTranslation.php
lukasbestle: ✅

lukasbestle:
@distantnative I think we are mixing up two different breaking changes: The general change by this PR (case-insensitive content) and the minor one about the `(array)` casting and the `array $content = null` argument (discussed above).

Regarding the minor one I agree with Bastian: We really don't need to touch that now. It makes the code cleaner, but doesn't have any user benefits.

Regarding the PR overall: As far as I can tell you are right that this would be the only breaking-change. But since we partly already treated content as case-insensitive, such fields were never fully supported anyway, so this PR makes the behavior a lot more consistent.

I checked performance again: It looks like performance slightly decreases because all fields are converted to lowercase when the content is loaded, even if they are already lowercase. Maybe we can make the `Content` constructor not do the conversion on load. I wrote this code so long ago that I don't fully remember the connections. But I can take a look again if we decide we want this in 3.6.

distantnative:
Agreed, let's not do the smaller one then.

I would want the general in 3.6. Decreased performance of course sounds problematic, especially for probably one of the most common operations in Kirby. How bad is it?

lukasbestle:
Almost negligible, but of course we should avoid any performance regressions if we can avoid them. I'll take a look.

bastianallgeier:
@lukasbestle do you have a local backup? I would like to close this and store a reference in the project.

lukasbestle:
Yep, thanks! 👍

distantnative:
@lukasbestle @bastianallgeier do we revive this and get it into 3.7? Seems pretty much ready.

lukasbestle:
@distantnative Yeah, it's on the list of things I want to work on for 3.7 over the next days (see Discord).

lukasbestle:
I've fixed the performance issue by skipping the capitalization of field names when the content is loaded from a text file (since the `Data\Txt` class already handles that). This leaves virtual pages, but as we saw in https://github.com/getkirby/kirby/issues/4142 we actually need to do the normalization for them.

I've also rebased the PR. Now finally ready for review. :)

distantnative:
Looks good to me and seems to work fine.

Only keep thinking about https://github.com/getkirby/kirby/pull/2945/files#diff-b046cf57a461e98333de92de1539cc6b080eb857ab217d344228aa5dc33f6ac7R63 - somehow it twists my brain that you're supposed to set something if the state already is fine. Keep thinking if it should be `$normalized = false` as default, and one needs to set it to `true` if normalisation should be skipped.

lukasbestle:
I don't have a strong preference for either way, but I feel like the `$normalize = true` default is more in line with other Kirby methods (the action you expect the method to do, not the state of the input data). E.g. `Str::excerpt()` with `$strip = true`.

distantnative:
True so maybe let's keep it like you have it in the PR.

afbora:
It seems to be working great 🎉 

Just why for the `array_change_key_case` function some use the default parameter `CASE_LOWER` while others do not? Is it for faster review when reading the code?

lukasbestle:
> Just why for the `array_change_key_case` function some use the default parameter `CASE_LOWER` while others do not? Is it for faster review when reading the code?

Yes. The function name doesn't describe the action well, so I always have to look up what it does (upper or lowercase).

afbora (approved):
I couldn't find any issue. LGTM 👍

PR #2953, opened 2020-11-27
Updates PHP dependencies and set PHP minimum requirement to 7.3

bastianallgeier:
## Describe the PR

See title

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ dist: bionic
 
 language: php
 php:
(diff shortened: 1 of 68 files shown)


lukasbestle (approved):
I haven't checked the code changes in the dependencies, but the changes to our code look good.

We also need to update the version check in the PHP code and the PHP version view though. But I can also do that in the PHP 8 PR.

PR #2957, opened 2020-11-28
PHP 8.0 compatibility

### LINKED ISSUE #2956: PHP8: Deprecation warning for required parameter after optional parameters
- **Kirby version:** 3.5.0-rc.2
- **PHP version:** 8.0.0 (stable)

When rendering a page with Kirby, a few warnings get printed out in the HTML output:

```
Deprecated: Required parameter $type follows optional parameter $string in vendor/getkirby/cms/src/Toolkit/Str.php on line 960
Deprecated: Required parameter $key follows optional parameter $parent in vendor/getkirby/cms/src/Cms/Field.php on line 99
Deprecated: Required parameter $value follows optional parameter $parent in vendor/getkirby/cms/src/Cms/Field.php on line 99
Deprecated: Required parameter $type follows optional parameter $string in vendor/getkirby/cms/src/Data/Data.php on line 82
Deprecated: Required parameter $type follows optional parameter $data in vendor/getkirby/cms/src/Data/Data.php on line 94
Deprecated: Required parameter $hook follows optional parameter $bind in vendor/getkirby/cms/src/Cms/Event.php on line 160
Deprecated: Required parameter $action follows optional parameter $method in vendor/getkirby/cms/src/Http/Route.php on line 96
```

On a second page load, these warnings disappear.

I’m not sure why they’re disappearing. Is that a feature of PHP? Or were the warnings due to some in-memory cache of the PHP files that got updated after that page load (and thus not actually an issue in 3.5.0-rc.2)? Or something else.

The relevant change in PHP8 seems to be:
https://github.com/php/php-src/pull/5067

I‘m not sure what PHP linters exist that could find all the affected functions.
Could be that the list above is complete for `getkirby/cms`, according [to this article](https://php.watch/versions/8.0/deprecate-required-param-after-optional):

> The deprecation notice is triggered at the compile time, and it will be raised even if it is not called.

lukasbestle:
## Describe the PR

Kirby now supports PHP 8.0. We recommend to use PHP 7.4 in production for now.

**Notes:**

- Travis doesn't support PHP 8.0 yet, so I have run the PHPUnit tests locally and they all still pass.
- PHP 8.0 has [many many breaking changes](https://www.php.net/manual/en/migration80.incompatible.php). A lot of them cannot be spotted using a global file search, so I wasn't able to check all breaking changes. All code that is tested by unit tests is safe, the small remaining rest may have bugs in PHP 8. I think there is no other way than to fix these bugs when they appear later. That's why I would recommend users to use PHP 7.4 for now unless they want to be on the bleeding edge.
- The optional method arguments in the middle of the argument list (e.g. `$arg1, $arg2 = null, $arg3`) [have never been supported by PHP and were ignored before](https://www.php.net/manual/en/migration80.deprecated.php#migration80.deprecated.core). So it's safe to remove them like I did it.

## Related issues

- Fixes #2956


distantnative:
Looks good to me from reading.
I seem to have problems getting PHP running on Apple Silicon 😅 So can't test it myself locally.

src/Data/Data.php
```diff
@@ -79,7 +79,7 @@ public static function handler(string $type)
      * @param string $type
      * @return array
      */
-    public static function decode($string = null, string $type): array
+    public static function decode($string, string $type): array
```
bastianallgeier, line 82: Not allowing null as value has bitten us a few times already. Are you sure this is not leading to any unwanted side effects?

distantnative: I think Lukas changed this cause PHP doesn't allow optional parameters ahead of a required one

afbora:
@lukasbestle Why don't we use `8.0snapshot` or `nightly` for PHP8 at Travis? I just tested your PR and all passed 👍 

https://travis-ci.com/github/afbora/kirby/builds/205915774

````yaml
php:
  - 7.3
  - 7.4
  - 8.0snapshot
````

src/Data/Data.php
bastianallgeier: Ah, right. Now I got it.

bastianallgeier:
I just added the snapshot test. I hope Travis will add the stable release asap. I also rebased the branch and fixed the composer.lock issue.

bastianallgeier (approved):

PR #2959, opened 2020-11-30
Add validations for layout field

### LINKED ISSUE #2951: Validation in layout field is not working yet
In the layouter you can save even if a block has required fields. As far as I know, this only works for pages in draft mode, but in the layouter this works for all statuses.

required

afbora:
## Describe the PR

Added layout field validations for settings and blocks. Inspired from blocks field 🙈 

## Related issues

- Fixes #2951 

## Ready?

- [x] Passing all unit tests


bastianallgeier (approved):
Thanks for taking care of this!

PR #2967, opened 2020-12-03
Add more unit tests for layout and blocks field

afbora:
## Describe the PR

I tried to add the missing unit tests for layout and blocks field.
I couldn't test all of them (100% coverage) because testing routes is really difficult. May be test via API 🤔 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Form/Field/BlocksField.php
+++ b/src/Form/Field/BlocksField.php
@@ -222,7 +222,12 @@ class BlocksField extends FieldClass
                     $index++;
                     $blockType = $block['type'];
 
-                    $blockFields = $fields[$blockType] ?? $this->fields($blockType) ?? [];
+                    try {
+                        $blockFields = $fields[$blockType] ?? $this->fields($blockType) ?? [];
+                    } catch (Throwable $e) {
+                        // skip invalid blocks
+                        continue;
+                    }
 
                     // store the fields for the next round
                     $fields[$blockType] = $blockFields;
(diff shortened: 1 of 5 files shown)


src/Form/Field/BlocksField.php
```diff
@@ -222,10 +222,6 @@ public function validations(): array
                     $index++;
                     $blockType = $block['type'];
 
-                    if (isset($fields[$blockType]) === false) {
-                        continue;
-                    }
-
```
afbora: This lines was preventing the bottom lines from running. Tell anyone who knows what it does 😅

bastianallgeier: It skips validation for blocks with unknown block types. This will make sure that the data for such unknown blocks will still be stored. Otherwise migrated blocks from builder and editor would create a validation error and also simply be removed after saving the blocks field for the first time.

afbora: Are we sure? `$fields` setted empty array. It will always be empty.

https://github.com/getkirby/kirby/blob/features/src/Form/Field/BlocksField.php#L218

bastianallgeier: Yikes. You are right. The check needs to be further down below after fetching the fields. 🙈

afbora: I have updated for layout and blocks validation. Used try-catch due to `fields()` method throw exception. Is it right now?

bastianallgeier (approved):

Issue #2940, opened 2020-11-21
Switch to vue-i18n


distantnative:
https://github.com/kazupon/vue-i18n

PR #2969, opened 2020-12-03
Remove vuex-i18n

distantnative:
## Status: ready for review 👀

- `vuex-i18n` isn't maintained anymore. 
- However, the popular `vue-i18n` is not great for how we use i18n in our setup. It's also rather bloated. 
- How about just creating our own minimal version? We already have almost everything in place and this little change seems to work just fine.

Saves us
- 8.15 kb
- 2.41 kb Gzipped

## Related issues
- Fixes https://github.com/getkirby/kirby/issues/2940

## Breaking changes 
- No `$tlang` anymore
- No `$i18n.localeExists` or `$i18n.keyExists`
- Translation strings cannot be added dynamically anymore only through our official plugin extension

## Ready?

## When merging


bastianallgeier (dismissed):
This looks really cool!

bastianallgeier:
Just a thought: I think the lib also had something like $tc for counts. Do we need to polyfill this as well if a plugin uses that?

distantnative:
@bastianallgeier I don't think so, can't find anything in their docs about a `$tc`. What they have is `$tlang` - I don't think this would be used by anyone in the context of the Panel. 

Of course, there could be some other breaking changes, if someone actually used some of their other methods, e.g. `$i18n.localeExists` or `$i18n.keyExists`. Or feeding in new strings - though plugin developer should always use our extension type for that. I don't think there is a lot of value in adding polyfills for these. I think it's ok to have that as small breaking change. But maybe that means we need to wait for 3.6 for this.

bastianallgeier:
I think it's a good idea to move this to 3.6

bastianallgeier:
@distantnative can you look into the conflicts after the vite merge?

bastianallgeier:
@distantnative can you look into the conflicts after the vite merge?

distantnative:
@bastianallgeier will do: better skip alls Vue-related PRs for today, so I can rebase them all properly

distantnative:
Done

bastianallgeier (approved):
Works like a charm ❤️

PR #2970, opened 2020-12-04
Use PHP 8.0.0 workaround on Travis CI

afbora:
## Describe the PR

The Travis CI team suggested that we use `8.0.0` temporarily.

**Discussion details with Travis team**
https://travis-ci.community/t/unable-to-load-dynamic-library-apcu-so-error-on-php-8-nightly/10754

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: php
 php:
   - 7.3
   - 7.4
-  - 8.0
+  - 8.0.0
 services:
   - memcached
 before_install:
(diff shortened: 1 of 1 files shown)


lukasbestle (approved):
@afbora Thank you for digging into it! I have decided to remove the `allow_failures` line as it would be risky if we miss a bug in Kirby on PHP 8 because of it. Currently the tests all pass and if they would start failing again (which I don't hope), we can add the allowed failure back in.

PR #2973, opened 2020-12-06
More detailed `accept` options for file blueprints; limit file types by default

lukasbestle:
## Describe the PR

Previously, file blueprints could be configured to only allow specific MIME types to be uploaded.

- It is now possible to limit the uploaded files based on the file extension and/or the file type.

```yaml
accept:
  extension: jpg, png
```

```yaml
accept:
  type: image, audio, video
```

- These restrictions can be combined with a custom set of MIME types. This makes it possible to define MIME types that are compatible with all browsers while still protecting the uploads based on the extension and/or file type:

```yaml
accept:
  mime: image/*
  type: image
```

In this example, Kirby will generate a file upload input with an `accept="image/*"` attribute and will then validate both the MIME type and the detected file type on the server after the file is uploaded.

- If no restrictions are defined in the file blueprint, Kirby will now limit the accepted uploads to the types image, document, archive, audio and video by default. This protects your sites against unexpected uploads like code files that could be used to attack the server or the visitor's browsers.

- You can disable the automatic types and allow all uploads like this:

```yaml
accept: true

# or:
accept:
  mime: null
  orientation: square
  ...
```

- New `F::typeToExtensions()` method

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/fields/mixins/upload.php
+++ b/config/fields/mixins/upload.php
@@ -30,7 +30,7 @@ return [
                     'template' => $template
                 ]);
 
-                $uploads['accept'] = $file->blueprint()->accept()['mime'] ?? '*';
+                $uploads['accept'] = $file->blueprint()->acceptMime();
             } else {
                 $uploads['accept'] = '*';
             }
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -72,7 +72,7 @@ return [
                     'template' => $this->template
                 ]);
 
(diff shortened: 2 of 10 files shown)


distantnative:
I would not make new additions/changes to the 3.5.0-RC but rather introduce this with 3.5.1

bastianallgeier (approved):

PR #2984, opened 2020-12-09
Create FUNDING.yml

bastianallgeier:
I think the Sponsor Button with a link to our buy page is something we should add to all our repos.


distantnative:
Does it work with leaving those unused lines in there?

bastianallgeier:
Yes, it works. I took it directly from the editor plugin.

lukasbestle:
But do we need these extra lines? I'm for making the file a one-liner for cleanliness.

bastianallgeier:
I just removed the parts that we don't need.

lukasbestle (approved):

PR #2985, opened 2020-12-09
Graceful fallback for Parsley when xml extension is missing

bastianallgeier:
## Describe the PR

When the xml extension is not loaded, Parsley will return a single markdown block with the HTML from the original value. 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/tests/Parsley/ParsleyTest.php
+++ b/tests/Parsley/ParsleyTest.php
@@ -19,7 +19,7 @@ class ParsleyTest extends TestCase
             $parser = new Parsley($input, new Blocks());
             $output = $parser->blocks();
 
-            $this->assertEquals($output, $expected, basename($example));
+            $this->assertSame($expected, $output, basename($example));
         }
     }
 
@@ -38,7 +38,7 @@ class ParsleyTest extends TestCase
             ]
         ];
 
-        $this->assertEquals($output, $expected);
+        $this->assertSame($output, $expected);
 
         // revert the global change
         Parsley::$documentClass = 'DOMDocument';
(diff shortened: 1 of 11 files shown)


tests/Parsley/ParsleyTest.php
```diff
@@ -19,7 +19,28 @@ public function testParsley()
             $parser = new Parsley($input, new Blocks());
             $output = $parser->blocks();
 
-            $this->assertEquals($expected, $output, basename($example));
+            $this->assertEquals($output, $expected, basename($example));
```
lukasbestle: Why that change? The old order was correct. Also please use `assertSame` for new tests.

lukasbestle (commented):
I think the method should actually use the defined class when initializing the document. Otherwise it‘s inconsistent and just there for the test.

I have also added a minor comment.

lukasbestle:
@bastianallgeier Now that I think about it: Maybe it should instead be `static::$useXmlExtension`. Otherwise people might try to do strange things with the property and we won‘t be able to change it later without breaking change.

tests/Parsley/ParsleyTest.php
bastianallgeier: I just fixed it. It actually helped to find a few things in block parsers that can be better.

bastianallgeier:
> @bastianallgeier Now that I think about it: Maybe it should instead be `static::$useXmlExtension`. Otherwise people might try to do strange things with the property and we won‘t be able to change it later without breaking change.

Ahh, that's much better. Could you check out if my implementation is what you had in mind?

lukasbestle (dismissed):

PR #3019, opened 2020-12-19
F::niceSize(): Typography and locale improvements

lukasbestle:
## Describe the PR

- `F::niceSize()` now respects the locale by default and can also be passed a custom locale or `false` to disable locale-aware formatting.
- `F::niceSize()` now places a no-break space between the number and the unit.

*Missing discussion in the [idea on Nolt](https://kirby.nolt.io/178).*

## Related issues

- Implements https://kirby.nolt.io/178

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Toolkit/F.php
+++ b/src/Toolkit/F.php
@@ -542,7 +542,7 @@ class F
 
         // avoid errors for invalid sizes
         if ($size <= 0) {
-            return '0 KB';
+            return '0 KB';
         }
 
         // the math magic
@@ -553,7 +553,7 @@ class F
             $size = I18n::formatNumber($size, $locale);
         }
 
-        return $size . ' ' . static::$units[$unit];
+        return $size . ' ' . static::$units[$unit];
     }
 
     /**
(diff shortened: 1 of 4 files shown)


bastianallgeier (dismissed):

bastianallgeier:
I would not introduce a non-breaking change in our PHP code. If this is really needed you can solve it via CSS on the parent element. I think this is taking too many "design decisions" into the backend code. I like it as it is.

bastianallgeier:
@lukasbestle can you resolve the conflicts. Afterwards it's ready to be merged in my opinion.

afbora (approved):
I've resolved the conflicts.

lukasbestle:
> @bastianallgeier I would not introduce a non-breaking change in our PHP code. If this is really needed you can solve it via CSS on the parent element. I think this is taking too many "design decisions" into the backend code. I like it as it is.

The PR *does* print no-break spaces right from PHP. You are right that this is maybe better solved in CSS.
So maybe we should revert this part of the PR but keep the locale-aware formatting?

PR #3020, opened 2020-12-20
Adds blocks navigation

afbora:
## Describe the PR

Adds blocks navigation that also support layout and nested blocks 🎉 

There was a minor focus issue with the PR that I thought I had completed before. 
We solved it with the help of @distantnative, thank you so much. 
I hope it's an acceptable PR.

blocks-navigation

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/panel/src/components/Blocks/Block.vue
+++ b/panel/src/components/Blocks/Block.vue
@@ -52,11 +52,13 @@
         />
         <k-button
           :disabled="!prev"
+          class="k-drawer-option"
           icon="angle-left"
           @click.prevent.stop="goTo(prev)"
         />
         <k-button
           :disabled="!next"
+          class="k-drawer-option"
           icon="angle-right"
           @click.prevent.stop="goTo(next)"
         />
@@ -253,5 +255,9 @@ export default {
 }
 .k-block-container[data-hidden] .k-block {
   opacity: .25;
+}
+.k-drawer-options .k-button[data-disabled] {
+  vertical-align: middle;
+  display: inline-grid;
 }
 </style>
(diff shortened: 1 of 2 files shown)


afbora:
By the way, if there is a point you see missing in PR, it is enough to give feedback, I can be interested 👍

bastianallgeier:
I love the navigation arrows. The PR is well done. I sat at this briefly today and was wondering about the position of the arrows. In general they should get a bit more padding to make it easier to click them. But I'm also not entirely sure yet if they should be in the header.

afbora:
Yes, a little more padding can be given. But I couldn't imagine a position other than the header. The common area of the blocks is the header and I think this is the ideal place I can think of. If you don't have a very different navigation style in mind, we can enhance it in the future versions.

afbora:
screen-capture-584-Mægazine-localhost

I've improved the look a bit more.

panel/src/components/Blocks/Block.vue
```diff
@@ -220,4 +256,8 @@ export default {
 .k-block-container[data-hidden] .k-block {
   opacity: .25;
 }
+.k-drawer-options .k-button[data-disabled] {
+  vertical-align: middle;
+  display: inline-grid;
+}
```
afbora, line 262: When disabled, the button was converted to span and the appearance was broken. Fixes this issue.

bastianallgeier: I was tripping over the same problem. Great catch!

bastianallgeier (approved):

PR #3021, opened 2020-12-20
I18n: Support for multiple fallback locales #3006

### LINKED ISSUE #3006: Entering wrong password on Panel Login screen gives says "Invalid argument - in method -"
**Describe the bug**  
The error reporting for the Panel login does not give a reasonable error message when entering in the wrong password for an actual Panel user.

**To Reproduce**  
1. Enable **Debug Mode**
2. Log out of Panel
3. Enter email address of extant Panel user
4. Enter incorrect password
5. Click "Login"

**Expected behavior**  
Says something like "wrong password"

**Screenshots**  
Screen Shot 2020-12-14 at 3 30 48 PM

**Kirby Version**  
`3.5.0-rc.7`

**Console output**  
```
[Error] Object

code: 400

details: [] (0)

file: "User.php"

key: "error.user.password.notSame"

line: 945

message: "Invalid argument \"-\" in method \"-\""

route: "auth/login"

status: "error"

Object Prototype

	onError (app.js:1:22156)
	(anonymous function) (app.js:1:17805)
	asyncFunctionResume
	(anonymous function)
	promiseReactionJob
```

 - OS: macOS
 - Browser: Safari 14.0.2 (16610.3.7.1.9)

lukasbestle:
### Features

- I18n: Support for multiple fallback locales

### Bug fixes

- Kirby now always falls back to the `en` translation if no translation for the site's default language exists.

### Deprecations

- `I18n::fallback()` -> `I18n::fallbacks()`

## Related issues

- Fixes #3006

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Cms/AppTranslations.php
+++ b/src/Cms/AppTranslations.php
@@ -46,6 +46,7 @@ trait AppTranslations
             return $data;
         };
 
+        // the actual locale is set using $app->setCurrentTranslation()
         I18n::$locale = function (): string {
             if ($this->multilang() === true) {
                 return $this->defaultLanguage()->code();
@@ -54,11 +55,11 @@ trait AppTranslations
             }
         };
 
-        I18n::$fallback = function (): string {
+        I18n::$fallback = function (): array {
             if ($this->multilang() === true) {
-                return $this->defaultLanguage()->code();
+                return [$this->defaultLanguage()->code(), 'en'];
             } else {
-                return 'en';
+                return ['en'];
             }
         };
 
--- a/src/Cms/AppTranslations.php
+++ b/src/Cms/AppTranslations.php
@@ -133,10 +133,10 @@ trait AppTranslations
     /**
      * Load a specific translation by locale
      *
-     * @param string|null $locale
+     * @param string|null $locale Locale name or `null` for the current locale
      * @return \Kirby\Cms\Translation|null
      */
-    public function translation(string $locale = null)
+    public function translation(?string $locale = null)
     {
         $locale = $locale ?? I18n::locale();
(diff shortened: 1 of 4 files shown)


afbora:
I have questions to understand better. 
Although the language code is `tr-tr` and the panel language is `tr`, error message is still in English. 

1. Can't we extract the fallback language from `tr-tr`?
2. Can we use panel language for panel side?
3. What if the fallback language can be set from config.php?

lukasbestle:
> 1. Can't we extract the fallback language from `tr-tr`?

@afbora You are right, if the code of the default language follows this pattern (something like `/^([a-z]+)-[a-z]+$/i` in RegEx), we could indeed extract the first part and add it as another fallback in the list before the final `en` fallback.

> 2. Can we use panel language for panel side?

Do you mean the `panel.language` option? To be honest, I don't have memorized how that works and where in the code it affects the language handling. We need to look into that.

> 3. What if the fallback language can be set from config.php?

Hm, not sure. I'd rather say that it should be configurable for each language in the language config. So that the `code` is used by default but can be overridden with another code just for translation handling if needed. Maybe a prop called `translationCode`?

But probably we are already good to go with feature 1 from your list.

afbora:
Anyway, all three were independent ideas I was thinking about to get the right fallback language. Yes, the solution you suggested for the first one looks good. Thank you 👍

bastianallgeier (approved):

afbora:
>  But probably we are already good to go with feature 1 from your list.

Will we add @lukasbestle suggestion later?

bastianallgeier:
@afbora let's implement that in a separate PR.

PR #3025, opened 2020-12-21
Improved translation fallback handling

### LINKED ISSUE #3006: Entering wrong password on Panel Login screen gives says "Invalid argument - in method -"

lukasbestle:
## Describe the PR

- The Kirby translation language is now also correctly detected if the configured default site language includes a country code (e.g. `en-us`).
- If `panel.language` is not configured, Kirby now falls back to the default site language before using `en` as a last resort.
- If no user is logged in, the translation in the Panel and API is now consistently determined based on the `panel.language` option or the default site language.

---

Open issue: If a *secondary* language (let's say `de-de`) includes a country code, the chain will now be `de-de`, `en-us`, `en`. Meaning: `de` is not considered by default. I think this would require deeper changes to the code and therefore I left it out for now. Maybe we can rewrite the `I18n` class to use a single `$locale` array instead of a `$locale` string and a `$fallback` array in Kirby 4.

## Related issues

- Fixes #3006
- See the discussion in https://github.com/getkirby/kirby/pull/3021

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/api/models/System.php
+++ b/config/api/models/System.php
@@ -15,7 +15,7 @@ return [
             return $this->kirby()->auth()->status()->toArray();
         },
         'defaultLanguage' => function () {
-            return $this->kirby()->option('panel.language', 'en');
+            return $this->kirby()->panelLanguage();
         },
         'isOk' => function (System $system) {
             return $system->isOk();
@@ -61,7 +61,7 @@ return [
             if ($user = $this->user()) {
                 $translationCode = $user->language();
             } else {
-                $translationCode = $this->kirby()->option('panel.language', 'en');
+                $translationCode = $this->kirby()->panelLanguage();
             }
 
             if ($translation = $this->kirby()->translation($translationCode)) {
--- a/src/Cms/Api.php
+++ b/src/Cms/Api.php
@@ -41,8 +41,11 @@ class Api extends BaseApi
 
         $allowImpersonation = $this->kirby()->option('api.allowImpersonation', false);
         if ($user = $this->kirby->user(null, $allowImpersonation)) {
-            $this->kirby->setCurrentTranslation($user->language());
+            $translation = $user->language();
+        } else {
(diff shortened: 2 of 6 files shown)


afbora:
I did a few tests:

1. Works perfect for hyphenated language codes 👍 

2. You have made an enhancement with `panel.language` related. Although I have specified `panel.language`, is it normal to see error messages in English on login page? So does this setting have any effect on the fallback language? Because when the default language and panel language are different in my tests on the login screen, get confused. I do not know if there is such a use. I'm just testing it.

3. In Regex, I think it would be better if we limited 2 characters to get the correct language code `{2}`, right?

lukasbestle:
@afbora Thanks for reviewing! Both are very good points, I have fixed them now.

afbora (approved):
It works great in all combinations 👍

bastianallgeier (approved):

PR #3029, opened 2020-12-22
Date field: fix missing `format` option #3026

### LINKED ISSUE #3026: date field's `format` property missing with v3.5.0 update (non-documented breaking change)
**Describe the bug**  
As [previously submitted to the forum](https://forum.getkirby.com/t/3-5-0-date-fields-format-property-removed/20473), the `format` property of the date field was removed without further notice in the official 3.5.0 update or the 3.5.0 changelog itself, which makes it a breaking change in that matter. The current format of the date field therefore falls back to always writing ISO 9075 format without the possibility to change that is in pre-3.5.0 versions. Depending on the use of Kirby, this can lead to unexpected, broken websites, e.g. when retrieving the dates via API, which now will respond with ISO 9075, where in a previous case, the developer might have set a specific, more fitting format for working with JavaScript, like ISO 8601.

**To Reproduce**  
1. Set `format` on a date field in a Kirby instance older than 3.5.0, e.g. `format: "Y-m-d\TH:i:s"` (ISO 8601).
2. Update to 3.5.0
3. The content files might now show ISO 9075. The `format` property has no effect any more.

**Expected behavior**  
The `format` property should still work and alternatively be deprecated and removed in the next major bump (even though I find this property useful, but not necessary). The documentation should still feature it, but with a deprecation or removal warning.

Alternatively, the format property should be added again and kept if there is no conflict or other reason any more why it should be removed.

**Screenshots**  
–

**Kirby Version**  

3.5.0

**Console output**  
–

**Additional context**  

My assumption is, this generally might not have a big effect on most developers unless you actually used Kirby headless, which in my case, would have broken all websites I currently maintain if the build system wouldn't have thrown an error.

More context by @distantnative 

> Indeed a breaking change, one that we/I thought was necessary when I worked on the field weeks ago - if I recall correctly, because of a way they started to conflict with the new display option. However, looking back now, I am not so sure anymore this breaking change was even necessary.

> Can we move this discussion to a Github issue (https://github.com/getkirby/kirby/issues 4) or are you both not on Github? I would like to spend some time thinking what would be the best way forward.

> And apologies for not finding out about this the proper way. The original PR (https://github.com/getkirby/kirby/pull/2892) actually still had the info about this breaking change, but we lost it on our way to the release notes. Which of course should not have happened.

distantnative:
## Describe the PR

During the development the development of the new date fields, I got to a point where the old `format` option conflicted with the new `display` option. Therefore, we removed `format` from 3.5.0 as breaking change. But upon reviewing the final version of the new date field, it has become clear to me that the breaking change is actually not necessary anymore.

This PR re-implements the `format` option.

Thanks to @renestalder for pointing this out.

## Related issues

- Fixes #3026

## Ready?

- [ ] Passing all unit tests

### DIFF
--- a/config/fields/date.php
+++ b/config/fields/date.php
@@ -106,6 +106,9 @@ return [
                 return Str::upper($this->display);
             }
         },
+        'format' => function () {
+            return $this->props['format'] ?? ($this->time === false ? 'Y-m-d' : 'Y-m-d H:i:s');
+        },
         'time' => function () {
             if ($this->time === false) {
                 return false;
@@ -127,10 +130,6 @@ return [
             return $this->toDatetime($this->value);
         },
     ],
-    'save' => function ($value) {
-        $format = $this->time === false ? 'Y-m-d' : 'Y-m-d H:i:s';
-        return $this->toContent($value, $format);
-    },
     'validations' => [
         'date',
         'minMax' => function ($value) {
--- a/config/fields/mixins/datetime.php
+++ b/config/fields/mixins/datetime.php
@@ -1,6 +1,14 @@
 <?php
 
 return [
+    'props' => [
+        /**
+         * Defines a custom format that is used when the field is saved
+         */
+        'format' => function (string $format = null) {
+            return $format;
+        }
+    ],
     'methods' => [
         'toDatetime' => function ($value, string $format = 'Y-m-d H:i:s') {
             if ($timestamp = timestamp($value, $this->step)) {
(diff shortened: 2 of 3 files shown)


distantnative:
@afbora as always your testing would be very welcome. I'm also struggling with PHPUnit in my local setup at the moment which is why I haven't written a unit test for this yet.

afbora:
I've tried many different formats for date and time field. It works great 👍 

It is just a problem with some different mixed formats. 
I'm sharing formats that I have tested and not working. 
I don't think there is such a use 😅 

- l jS \of F Y
- M j D, Y

bastianallgeier:
@distantnative LGTM. Can we mark this for review or do you still work on it?

distantnative:
@bastianallgeier now ready

bastianallgeier (approved):

PR #3035, opened 2020-12-27
Static code analysis with Psalm

lukasbestle:
Maybe you remember that we tried out the PHPStan static analyzer tool a few years ago and weren't happy with its many false-positives.

Static analysis in general is really powerful though and so I used the opportunity of switching to GitHub Actions to take a look at the topic again. PHPStan still doesn't really fit for our type of code-base and I couldn't even get it to work anymore. That left Phan and Psalm. Phan looks quite good too, but Psalm has really nice documentation and works quite well out of the box, so I integrated it into our CI process.

This PR fixes the code issues we need to fix to get to Psalm level 7 (the second-lowest in a scale where level 1 is strictest and level 8 is most lenient). Levels 6 and up need a lot more changes to our code, but I think it's good to already start somewhere.

I have split the different types of changes into separate commits and added an explanation in the commit message where necessary.


src/Form/Field.php
```diff
@@ -207,16 +207,19 @@ public static function defaults(): array
             ],
             'computed' => [
                 'after' => function () {
+                    /** @var \Kirby\Toolkit\Component $this */
```
distantnative: Wouldn't this (and the following) be rather `\Kirby\From\Field`?

distantnative:
Looks good to me. Added one comment.
Not the biggest fan of all the redundant `@var` declarations, but I don't think there is a better way. And in the end this PR already shows how Psalm helps us to catch some bugs.

distantnative:
And should we add `psalm` to our `composer.json`? As dev dependency and a script? Then local testing would be possible. Or I vaguely recall that adding dependencies messes with developer setups that actually use Kirby with composer? I'm also fine if `psalm` has to be installed globally.

Also currently our composer still features a `phpstan` script. Let's throw that out.

src/Form/Field.php
lukasbestle: You are right! When I wrote the hints I looked up the calling code and saw `$this`, but `Form\Field` extends `Toolkit\Component`. Now fixed. :)

lukasbestle:
> Not the biggest fan of all the redundant `@var` declarations, but I don't think there is a better way.

I don't think they are redundant. When I work on some of this code, I myself often don't know what `$this` or any other variable is referring to and need to go hunting in the unknown depths of our code to find out what the variable is. So I think the `@var` annotations are going to help us as well in the long run.

**Open question to everyone in the dev team @bastianallgeier @distantnative @afbora:** I have currently disabled the `$this` checks for the `config` folder. With this check enabled, there would be 269 more of these annotations to make. However it's not a good idea to keep this disabled as Psalm won't be able to find out what `$this` is, so follow-up checks (like if called methods exist) cannot be run. We have two options going forward:

- Either we add all the 269 annotations. That would be my preferred solution to be honest for the reason I mentioned above (it would help us when developing at the same time).
- Or we write a Psalm plugin that will tell Psalm about the types based on the path of the file that gets analyzed (for example basically everything inside `config/api` gets `Kirby\Cms\Api`). Disadvantage: I'm not sure if it's fully reliable and also it would be a bit of work to write the plugin.

I wanted to ask you before I invest the time for the 269 annotations. As I wrote, I think it is worth it though. What do you think?

> And should we add psalm to our composer.json? As dev dependency and a script? Then local testing would be possible. Or I vaguely recall that adding dependencies messes with developer setups that actually use Kirby with composer?

Even worse: If we add it as a dependency to `composer.json`, we will commit it and its dependencies to the repo inside the `vendor` folder, which will be an issue for everyone who installs Kirby manually or via submodule. It's the same reason why we also don't require PHPUnit or PHP-CS-Fixer that way. Composer installation is actually unaffected if we added it as a dev dependency.

Installing Psalm globally is quite simple: `composer global require vimeo/psalm`. Then you only need to run `psalm` inside the `kirby` dir and it will do its thing.

> Also currently our composer still features a phpstan script. Let's throw that out.

You are right, done. ✅
I have replaced it with one for Psalm.

distantnative:
Ah yes, that was the problem with Composer. As said, global works for me, was just wondering.

Redundant `@var`: I think in many of the current changes it's totally fine, where I felt it is redundant/annoying little are the defaults in the `Fields` class. Which would be the same with all the declarations in `config/` - I think the readability of the file could suffer quite a bit if we have a `@var` line for each and every field prop, field computed, field method...

Any chance we could define this once per file?

Also, I might not understand Psalm enough, so please correct me, but how much will this help us? I would assume we're just running in the next even bigger round of warnings cause pretty much all that uses `$this->` in `config/` is based on a magic caller or so. Will Psalm understand that e.g. `$this->min` then refers to the callback in the same file defined in the `props` array?

Just want to avoid all the work for `$this` if it only creates more warnings and no real benefit (in the case of `config/fields/` etc.).

lukasbestle:
> Any chance we could define this once per file?

I don't think so, as Psalm doesn't know that we are referring to the variables inside functions.

We could maybe create a hybrid plugin that will accept the annotations at the top of the file and pass them to Psalm as annotations for every single function internally. But I don't have experience with writing Psalm plugins.

Before we do anything, let's wait for the others' opinions.

> Will Psalm understand that e.g. $this->min then refers to the callback in the same file defined in the props array?

Psalm sees that there is a magic `__call()` method in the respective class, so it won't flag the method calls. The opposite actually: It won't know on its own which method names exist and will therefore not be able to check much after that for now. In higher Psalm levels it will warn us about that and we can then fix it by annotating the `__call()` methods.

However it *will* be able to run additional checks for all methods that are explicitly defined, which are a lot.

bastianallgeier:
The long-term goal should be to be less clever with our $this binding and pass dependencies as parameters instead. It seemed super cool when I used it for the first time, but it really isn't that great in many situations. 

For now, I think the `@var` annotations would indeed make the files more chaotic, but I also agree that they help us to know what we mean with $this. Just as an experiment I jumped into a random file in config (config/api/authentication.php) and it's a perfect example where it's totally unclear what $this is.

Wouldn't this also help to improve navigation in IDEs? This could be at least another positive side-effect for all the work.

distantnative:
@lukasbestle in `src/Form/Field.php` the annotation is currently

```
 /** @psalm-scope-this Kirby\Form\Field */
 ```
 
 Would is also work with this?
 
 ```
 /** @var \Kirby\Form\Field $this */
 ```
 
 I think it would be preferable to use this neutral declaration (which I think indeed is used by IDEs) instead of the psalm-specific one - if we can.

**Comment by lukasbestle** [comment]
@distantnative For functions inside classes (like that `Form\Field` example) we could do that, but for the files in `config` we can't (Psalm will then throw errors that `$this` is used outside a class). What we could do is to use both in parallel. Or we check if we can find a way to make Psalm happy with `@var`.

**Comment by lukasbestle** [comment]
I have tracked down the issue I mentioned above. It seems to be a bug/missing feature in Psalm, so I opened an issue here: https://github.com/vimeo/psalm/issues/4916

Once Psalm supports `@var` with `$this` outside classes, we can use `@var` declarations everywhere inside the `config` dir and enable the rule in `psalm.xml` again. But until then we can already merge this PR.

**Comment by lukasbestle** [comment]
I have also added configuration for GitHub's [code scanning alerts](https://github.com/getkirby/kirby/security/code-scanning) to this PR. In the code scanning UI we can visually see all issues that Psalm detects. The issues that are not yet matched by our configured Psalm level are added as notes only.

**Comment by distantnative** [comment]
👍 I'd be fine with merging the PR then. With the `config/` folder I would wait then and see if it gets fixed in Psalm before we add all the comment lines.

One other question: seeing a lot of checks on this PR now (18). I understand that a PR will always have both checks for push and PR. But does it make sense to have Coveralls run separately for each PHP version?

And what do you think about limiting the workflows a bit, e.g. only running all the PHP jobs if any `*.php` file is changed?

**Comment by lukasbestle** [comment]
> But does it make sense to have Coveralls run separately for each PHP version?

Yes! We have (or at least [have had](https://github.com/getkirby/kirby/pull/2957/files#diff-fafec9f34f80157359d475da6ee6928a4ef3c12cffbf822cff0d4c329110c1bbL61)) places in the code that were only called for specific PHP versions, so the overall coverage can only be calculated correctly if the coverage results of all PHP version tests are added. Also it's not actually using any more resources as the Coveralls submission is done inside the same PHP version workflow runs.

However there's indeed one issue: We really don't need Coveralls to report the coverage for each version as a check. I have already disabled that feature in the Coveralls settings ("only report aggregate results" or something along these lines), but for some reason Coveralls still adds four checks instead of just one.

That's all unrelated to this PR though.

> And what do you think about limiting the workflows a bit, e.g. only running all the PHP jobs if any *.php file is changed?

Yes, that's a good idea! We could at least do that for the PR builds. IMO the branch builds should always be complete to avoid that we miss issues because a test didn't run. But I'd also do that separately once this PR is merged.

**Comment by lukasbestle** [comment]
@bastianallgeier @distantnative This PR is now ready for review. Could I have a second pair of eyes of one of you to check if all changes in this PR are correct and don't introduce other issues? Feel free to merge the PR directly afterwards.

File: src/Cms/Item.php [hunk 2]
```diff
@@ -109,7 +109,7 @@ public function parent()
      * Returns the sibling collection
      * This is required by the HasSiblings trait
      *
-     * @return \Kirby\Editor\Blocks
+     * @return self::ITEMS_CLASS
```
distantnative: I think this will break our automatic reference docs. I couldn't find any actual information whether this is a supported format for phpdocumentor - what insight have you found there?

distantnative: On line 37 we define `\Kirby\Cms\Items` as type of the `siblings` property. So maybe we should just hint that here?

src/Toolkit/Html.php
```diff
@@ -175,6 +175,8 @@ public static function email(string $email, $text = null, array $attr = []): str
      * @param string|null $string
      * @param bool $keepTags If true, existing tags won't be escaped
      * @return string The HTML string
+     *
+     * @psalm-suppress ParamNameMismatch
```
distantnative, line 179: Should we add a `@todo` tag here?

distantnative:
Went through all the commits and they're looking really good. I had two more questions - see comments.

One separate thought: I think it would be goof if we agree as a team on a way for marking to-do's in comments. Currently there is `// TODO: ` vs. `@todo `. I like the first cause my IDE has bright yellow highlighting for it (currently we use it in 9 spots in the code). But I wouldn't mind switching either. Just think it would be good to all use a consistent style.

src/Toolkit/Html.php
lukasbestle: I explicitly left it out here as the param name mismatch is a non-issue for public static methods in static-only classes like `Html` and `Xml`. You wouldn't depend on `Html` being 100 % compatible with `Xml` as the classes have different use-cases. The only reason the classes use inheritance is that it reduced code duplication.

src/Cms/Item.php
lukasbestle: It's a protected method, so it's not going to be in the reference, is it? But in general you are right, the format is a bit non-standard. Psalm supports its own annotations that override the default `@return` annotation, so I now used that one so Psalm is happy.

The issue with just using `\Kirby\Cms\Items` would be that it wouldn't be correct for classes that extend the `Item` class. That's why we have the constant `ITEMS_CLASS` after all. For example the `Block` class sets `\Kirby\Cms\Blocks`. If we fixed it for the parent class, Psalm wouldn't know the right type for further checks of the children classes and their dependents.

If you agree with my solution with `@psalm-return`, please resolve this review comment. Otherwise please let me know.

lukasbestle:
@distantnative ✅

> One separate thought: I think it would be goof if we agree as a team on a way for marking to-do's in comments. Currently there is `// TODO:` vs. `@todo`.

I agree that it should be consistent! I don't know about the JS world, but for PHP code I'd definitely prefer `@todo` as it's [standard](https://docs.phpdoc.org/latest/guide/references/phpdoc/tags/todo.html). We can discuss this separately in another issue, can you make one?

distantnative (approved):

Issue #3040, opened 2021-01-03
Consistent TODO comments


distantnative:
Currently we use both `TODO:` and `@todo`.
Would be good if we could agree on one style. Or at least one style per language (if we think we need to have different ones for JS and PHP).

Lukas is voting for `@todo` (at least for PHP).
I have been using `TODO:` but open for change.

@afbora @bastianallgeier what would be your preferences?

afbora:
I think both are the correct use. `TODO` in comment lines, `@todo` in docblocks.

````php
/**
 * @todo The locale magic should no longer be
 */
public static function encode($data): string
{
    // TODO: The locale magic should no longer be
}
````

If we only choose one, I would write in detail what to do in the method in `@todo` in docblock.

Also `PhpStorm IDE` default pattern finds both uses.

lukasbestle:
I really like @afbora's proposal to differentiate between docblocks and single-line comments. That totally makes sense.

distantnative:
I agree, that sounds good. https://github.com/getkirby/kirby/pull/3042

PR #3042, opened 2021-01-03
Consistent TODO comments #3040

distantnative:


lukasbestle:
@distantnative The changes you made are in docblocks, so the lines were already correct. I think we don't need to change anything at this point and can close this PR and #3040.

distantnative:
Those aren't proper do blocks annotating methods or classes as far as I can see but just multiline comments inside methods.

lukasbestle:
I'd still see them as docblocks because of the leading `*` in each line. Also it's a bit weird to use `TODO: ` immediately followed by a PHPDoc tag in the next line.

@afbora What do you think Should we use `@todo` or `TODO: ` in multiline comments inside methods?

afbora:
Yes, as @lukasbestle said, it seems better to use `@todo` as we are still using docblock.

Open new question: Is it correct to use docblock on inline lines in anywhere? I do not know the answer of question, but by some standards, `if/condition` does not appear on this list.

https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#3-definitions

lukasbestle:
@afbora Not according to that specification, but Psalm needs it anyway. :)
@distantnative If you agree with us, please close this PR and the issue. Otherwise let us know.

distantnative:
I don't quite understand how this isn't an inline comment as discussed in the issue. Just because of the comment style - `*` vs `//`?

In that case, shouldn't we be consistent and only use `//` inline (inside methods vs. documenting methods and classes with docblocks)?

lukasbestle:
> I don't quite understand how this isn't an inline comment as discussed in the issue.

Because it uses the `@psalm-suppress` docblock tag.

> In that case, shouldn't we be consistent and only use // inline (inside methods vs. documenting methods and classes with docblocks)?

Yes, but when we need to tell Psalm about something, we can't – it only parses `/**` comments, not `//` comments. That's why these occurrences are using docblocks.

To be honest I don't want to discuss this further. It's a tiny detail that will be gone at some point once the `TODO` is resolved. In general I fully agree with you that we should be using `//` comments inside methods if possible.

PR #3046, opened 2021-01-04
Improves `Cookie::forever()` method for 32-bit servers #2968

### LINKED ISSUE #2968: Create lifetime cookie returns a error
I have made a Cookie consent solution by my own with Kirby Cookies. However, when I use this:
```php
<?php Cookie::forever('consent', 'all'); ?>
```
It works on localhost, but not on the hosting for the customer (Host Europe), I get this error:
```
Argument 1 passed to Kirby\Http\Cookie::lifetime() must be of the type int, float given, called in .../kirby/src/Http/Cookie.php on line 45
```
But this works.
```php
<?php Cookie::set('consent', 'all'); ?>
```
As reference, please check this thread at the forum. Is there a way to avoid this? Not sure if this is a bug or only a slow hosting problem. However, if I use the forever function, the site will not work.

https://forum.getkirby.com/t/create-lifetime-cookie-fails/20228/4

afbora:
## Describe the PR

Since we do not have a test facility on a 32-bit server, I had to do the test without using the `forever()` method.

## Related issues

- Fixes #2968

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Http/Cookie.php
+++ b/src/Http/Cookie.php
@@ -98,7 +98,8 @@ class Cookie
      */
     public static function forever(string $key, string $value, array $options = []): bool
     {
-        $options['lifetime'] = 253402214400; // 9999-12-31
+        // 9999-12-31 if supported (lower on 32-bit servers)
+        $options['lifetime'] = min(253402214400, PHP_INT_MAX);
         return static::set($key, $value, $options);
     }
 
--- a/tests/Http/CookieTest.php
+++ b/tests/Http/CookieTest.php
@@ -8,28 +8,28 @@ class CookieTest extends TestCase
 {
     public function testKey()
     {
-        $this->assertEquals('KirbyHttpCookieKey', Cookie::$key);
+        $this->assertSame('KirbyHttpCookieKey', Cookie::$key);
         Cookie::$key = 'KirbyToolkitCookieKey';
-        $this->assertEquals('KirbyToolkitCookieKey', Cookie::$key);
+        $this->assertSame('KirbyToolkitCookieKey', Cookie::$key);
     }
 
     public function testLifetime()
     {
-        $this->assertEquals(12345678901, Cookie::lifetime(12345678901));
-        $this->assertEquals((600 + time()), Cookie::lifetime(10));
-        $this->assertEquals(0, Cookie::lifetime(-10));
+        $this->assertSame(253402214400, Cookie::lifetime(253402214400));
+        $this->assertSame((600 + time()), Cookie::lifetime(10));
+        $this->assertSame(0, Cookie::lifetime(-10));
     }
 
     public function testSet()
     {
         Cookie::set('foo', 'bar');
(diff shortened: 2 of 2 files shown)


src/Http/Cookie.php
```diff
@@ -63,11 +63,15 @@ public static function set(string $key, string $value, array $options = []): boo
     /**
      * Calculates the lifetime for a cookie
      *
-     * @param int $minutes Number of minutes or timestamp
+     * @param int|float $minutes Number of minutes or timestamp
      * @return int
      */
-    public static function lifetime(int $minutes): int
+    public static function lifetime($minutes): int
```
afbora: Removed `int` type because it converts to float if the server doesn't support it.

distantnative: What converts to float?

afbora: a number greater than the maximum integer, for 32-bit server on:

    Cookie::lifetime(253402214400); // you'll get type error

lukasbestle: To be honest I wouldn't allow floats, it would create a type mess that's very hard to test for edge-cases.

Have you tried the [solution I proposed here](https://github.com/getkirby/kirby/issues/2968#issuecomment-738266254) directly in the `forever()` method?

afbora: It works, but I wanted to combine it because the `lifetime()` method has the same issue. Of course, `forever()` Kirby originated, `lifetime()` is a user issue.

tests/Http/CookieTest.php
```diff
@@ -40,6 +40,23 @@ public function testRemove()
...
+     * - 64-bit 253402214400
+     */
+    public function testForeverOn32BitServer()
+    {
+        // to be tested manually by sending the max integer on a 32-bit server
+        Cookie::set('forever', 'bar', ['lifetime' => 2147483647]);
+
+        $this->assertSame('703a07dc4edca348cb92d9fcb7da1b3931de0a85+bar', $_COOKIE['forever']);
+        $this->assertTrue(Cookie::exists('forever'));
+    }
```
afbora: You know, it's not actually a 32-bit test 🙈 Let's remove it if you wish.

lukasbestle: Hm, good point. 32-bit is not really testable without a 32-bit PHP binary. But I'd say it's fine as the year 2038 is coming closer and closer. 😜
Just kidding: It's an edge-case and we have now verified it manually and fixed it. So we don't really need a test if it's not viable (which it isn't).

src/Http/Cookie.php
lukasbestle: That's a good point! I'm kind of split here – your original solution definitely fixes the issue well, but I'm not sure if we should weaken our type hints just for that edge-case. If we could already use union types (`int|float`) in PHP 7.3, it would be a non-issue.

lukasbestle (approved):

PR #3048, opened 2021-01-04
Fix types of `page/s()` helper and `$kirby->page()`

distantnative:
## Status: ready for review 👀

- Fixes types for `page()`/`pages()` helpers
- Prepares to make their functionality much stricter

## Breaking changes 
- `page()` will now throw a deprecation warning if multiple IDs get passed (suggests to use `pages()`)
- `pages()` will now throw a deprecation warning if single ID get passed (suggests to use `page()`)
- Actual removal of functionality planned for 3.7.0

## When merging

### DIFF
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -551,12 +551,17 @@ function option(string $key, $default = null)
  * id or the current page when no id is specified
  *
  * @param string|array ...$id
- * @return \Kirby\Cms\Page|null
+ * @return \Kirby\Cms\Page|\Kirby\Cms\Pages|null
+ * @todo reduce to one parameter in 3.7.0 (also change return type)
  */
 function page(...$id)
 {
     if (empty($id) === true) {
         return App::instance()->site()->page();
+    }
+
+    if (count($id) > 1) {
+        deprecated('Passing multiple parameters to the `page()` helper has been deprecated. Please use the `pages()` helper instead.');
     }
 
     return App::instance()->site()->find(...$id);
@@ -566,7 +571,7 @@ function page(...$id)
  * Helper to build page collections
  *
  * @param string|array ...$id
- * @return \Kirby\Cms\Pages
+ * @return \Kirby\Cms\Page|\Kirby\Cms\Pages|null
  */
 function pages(...$id)
 {
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -552,7 +552,7 @@ function option(string $key, $default = null)
  *
  * @param string|array ...$id
  * @return \Kirby\Cms\Page|\Kirby\Cms\Pages|null
- * @todo reduce to one parameter in 3.7.0 (also change return type)
+ * @todo reduce to one parameter in 3.7.0 (also change return and return type)
  */
 function page(...$id)
 {
@@ -561,7 +561,9 @@ function page(...$id)
     }
 
     if (count($id) > 1) {
+        // @codeCoverageIgnoreStart
         deprecated('Passing multiple parameters to the `page()` helper has been deprecated. Please use the `pages()` helper instead.');
+        // @codeCoverageIgnoreEnd
     }
 
     return App::instance()->site()->find(...$id);
@@ -572,9 +574,16 @@ function page(...$id)
  *
  * @param string|array ...$id
  * @return \Kirby\Cms\Page|\Kirby\Cms\Pages|null
+ * @todo return only Pages|null in 3.7.0, wrap in Pages for single passed id
  */
 function pages(...$id)
 {
+    if (count($id) === 1) {
+        // @codeCoverageIgnoreStart
+        deprecated('Passing a single id to the `pages()` helper will return a Kirby\Cms\Pages collection with a single element instead of the single Kirby\Cms\Page object itself - starting in 3.7.0.');
+        // @codeCoverageIgnoreEnd
+    }
+
     return App::instance()->site()->find(...$id);
 }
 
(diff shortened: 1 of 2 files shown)


src/Cms/App.php
```diff
@@ -984,7 +984,7 @@ protected function optionsFromReadyCallback(): array
      * @param string|null $id
      * @param \Kirby\Cms\Page|\Kirby\Cms\Site|null $parent
      * @param bool $drafts
-     * @return \Kirby\Cms\Page|null
+     * @return \Kirby\Cms\Page|\Kirby\Cms\Pages|null
```
lukasbestle: How can this method return a `Pages` object? I'm sure it can somehow, otherwise Psalm wouldn't have complained, but should it? I think we should make sure that this does not happen, looks like a bug to me.

lukasbestle (commented):
What happens if `pages()` gets called without argument? That would now return the current page, which is weird.

Also: `pages()` should always return a collection IMO. So it should first check that `$id` is not empty and then `A::wrap()` it. This would be a breaking-change, but everything else is confusing.

distantnative:
Is it much weirder than the current behavior of also just passing one id to `pages()` and getting one page? I agree that it's not quite logical, more convenient - same that `page()` also works for multiple pages. But is there really the need to introduce breaking changes? Tried to rather keep the current behavior.

src/Cms/App.php
distantnative: My mistake, I though it's the same as `page()` - but actually here `$id` can only be one page. I think Psalm gets confused cause `->find()` could return also `Pages` but only if multiple parameters/arguments passed. How can we make Psalm understand that since we only pass a single argument, the return will for sure only be of type `Page`?

lukasbestle: Ah, sure, that will be it! Maybe this can be solved with conditional types: https://psalm.dev/docs/annotating_code/type_syntax/conditional_types/

lukasbestle:
I'm split on this as well. On one hand, fewer breaking changes are always better, but if we now start to document it as an alias, we will never be able to get rid of this impurity again. It's really weird and confusing and I think it shouldn't be that way.

Another argument in favor of introducing a breaking-change: I never knew that `pages()` could be used to get a single page only as it was never documented. So assuming that no user knew that by digging around in the source code, the `pages()` helper could only have been used this way by accident. So the breaking-change is less critical.

We could also first deprecate this behavior in 3.6 with a deprecation warning and then change the implementation in 3.7.

texnixe:
> I never knew that pages() could be used to get a single page only as it was never documented.

Nor did I. I would find two helpers that do exactly the same also more confusing.

On the other hand I never know you could pass multiple IDs to the `page()` helper to get a collection...

I would actually prefer a `page()` helper that returns a single page, and a `pages()` helper that returns a pages collection.

src/Cms/App.php
distantnative: That could work. Have to test if we can make the required `@template` syntax work for getkirby.com auto docs

distantnative:
> I would find two helpers that do exactly the same also more confusing.

That's pretty much what we have had so far already (except the difference when passing nothing). But ok then we do the breaking changes but only in 3.6

src/Cms/App.php
distantnative: @lukasbestle It seems like Psalm conditions work for types but not number of parameters

lukasbestle: Maybe it would be possible with multiple template types and `@psalm-param` for "faked" params, but that's probably way too complex.

Alternative: In line 998 before the `return $page` line, there could be `/** @var \Kirby\Cms\Page $page */`, which would tell Psalm the actual returned value.

distantnative:
@lukasbestle @texnixe 
OK, I rewrote this:
Added return types as the helpers work right now.
Also added a deprecation warning to `page()` so we can convert the helper to only return a single page in 3.7.0

config/helpers.php
```diff
@@ -548,22 +548,27 @@ function option(string $key, $default = null)
...
+
     return App::instance()->site()->find(...$id);
 }
 
 /**
  * Helper to build page collections
  *
  * @param string|array ...$id
- * @return \Kirby\Cms\Pages
+ * @return \Kirby\Cms\Page|\Kirby\Cms\Pages|null
```
lukasbestle, line 576: I think the return value here should be consistent as well. So when passing a single ID, I'd expect to get a `Pages` collection with a single element.

See https://github.com/getkirby/kirby/pull/3048#pullrequestreview-561332134 for how I'd implement it.

distantnative: That would be another breaking change. So how do we do that to offer a smooth transition?

lukasbestle: See https://github.com/getkirby/kirby/pull/3048#issuecomment-754237313. As it was never documented, it's technically not a breaking-change.

Of course it's not optimal and a few people may run into this, but I think it's worth it to clean this up. If we documented the `Page` return type explicitly, it would make the situation worse as we would never be able to clean this up.

lukasbestle: What we could do for the transition: Throw a deprecation warning if the `pages()` helper is called with only one argument and then change the behavior in 3.7.0.

distantnative: I think I'd prefer the latter. This isn't a crucial thing to change for us. So I would make it as smooth as possible.

lukasbestle: Yes, I agree. We definitely shouldn't break sites if we can avoid it.

distantnative:
Updated the initial opening post to reflect the current state

bastianallgeier (approved):
That's a really good step forward to get this mess sorted.

PR #3050, opened 2021-01-05
Avoid invalid content types for missing representations

### LINKED ISSUE #2924: Using .php in URLs can lead to unwanted content representations
**Describe the bug**  
When you add .php to the path of a page and the slug is the same as the template (i.e. /blog.php and the template is blog.php) Kirby will load the HTML code of the blog.php template and send a content type that will force the browser to show the result. 

We should disable content representations for .php entirely.

bastianallgeier:
## Describe the PR

When a content representation cannot be found, the App::resolve method still set the wrong content type according to the given extension. Of course this can lead to unwanted side-effects. (see issue) 

## Related issues

- Fixes #2924

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/tests/Cms/Pages/PageTemplateTest.php
+++ b/tests/Cms/Pages/PageTemplateTest.php
@@ -106,4 +106,27 @@ class PageTemplateTest extends TestCase
         $page = $this->app->page('with-template');
         $page->representation('xml');
     }
+
+    public function testRepresentationErrorType()
+    {
+        $this->app = new App([
+            'templates' => [
+                'blog' => __DIR__ . '/fixtures/PageTemplateTest/template.php',
+            ],
+            'site' => [
+                'children' => [
+                    [
+                        'slug' => 'blog',
+                        'template' => 'blog'
+                    ]
+                ]
+            ]
+        ]);
+
+        $this->app->resolve('blog.php');
+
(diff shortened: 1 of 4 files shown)


distantnative (dismissed):
LGTM

tests/Cms/Pages/PageTemplateTest.php
```diff
@@ -106,4 +106,27 @@ public function testRepresentationError()
...
+                        'template' => 'blog'
+                    ]
+                ]
+            ]
+        ]);
+
+        $this->app->resolve('blog.php');
+
+        // there must be no forced php response type if the
+        // represenation cannot be found
```
lukasbestle:
```suggestion
        // representation cannot be found
```

lukasbestle (commented):
The fix is great and simple!

IMO the test should be in the `AppResolveTest` class to keep it with the other tests for the tested method. It would be good to also check the return value of `$app->resolve()`, which should be `null`.

bastianallgeier:
@lukasbestle You are right! I just moved the test, added the assertion and fixed the typo

lukasbestle (approved):

afbora (approved):
I really like the new `php` presentation with this fix 🎉

PR #3055, opened 2021-01-05
Activate and fix PHPMD LongVariable rule

distantnative:
Since it's a protected static property, this shouldn't be a breaking change, right?


bastianallgeier (approved):
I'd say it's fine and I prefer the var name, but I wait for @lukasbestle's opinion

lukasbestle (approved):

PR #3058, opened 2021-01-05
Fix ->call() only with Closure

distantnative:
`->call()` only works with Closures so we need to explicitly check for them not just for `is_callable`

### DIFF
--- a/src/Form/Field.php
+++ b/src/Form/Field.php
@@ -84,7 +84,10 @@ class Field extends Component
      */
     public function api()
     {
-        if (isset($this->options['api']) === true && is_callable($this->options['api']) === true) {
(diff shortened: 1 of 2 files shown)


bastianallgeier (commented):
We use is_a for every instance check so far. I'm not sure if there's a difference, but it feels like we are using a new style here.

distantnative:
True, found `instanceof` as the common way to check for `Closure`, but indeed we are using `is_a` with it in other places already. Will adapt the PR.

distantnative:
Done.

bastianallgeier (approved):

lukasbestle (approved):

PR #3059, opened 2021-01-05
Fix `I18n::template` docblock parameter types

distantnative:


src/Toolkit/I18n.php
```diff
@@ -189,12 +189,12 @@ public static function translate($key, $fallback = null, string $locale = null)
...
-     * @param string $fallback
-     * @param array $replace
-     * @param string $locale
+     * @param string|array|null $fallback
+     * @param array|null $replace
+     * @param string|null $locale
      * @return string
      */
-    public static function template(string $key, $fallback = null, array $replace = null, string $locale = null)
+    public static function template(string $key, $fallback = null, ?array $replace = null, ?string $locale = null)
```
afbora:
```suggestion
    public static function template(string $key, $fallback = null, ?array $replace = null, ?string $locale = null): string
```

Also adds `string` return type 👍

distantnative: Good catch!

afbora (approved):

bastianallgeier (approved):

PR #3061, opened 2021-01-05
Fix some return types

distantnative:
- `PageActions:: createNum()` should always return `int`, so we need to type cast the result of `date()`
- `Location::num()` should also be type casted to a `float` if only one part exist
- `F:: niceSize()` never returns an `int`, always just `string`
- `Iterator::indexOf()`always returns the `int` index position (or false) but never a string key of an associative array, since we use `array_key()`. This has been the behavior so far, so we're only fixing the description.
- Pagination: `ceil()` still returns a `float`, so we need to cast it to an `int`
- `Str::isUrl()`: `filter_var` will return `false` or the value - so we need to explicitly check that the return is not `false`

### DIFF
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -533,6 +533,8 @@ trait PageActions
                 $lang   = $this->kirby()->defaultLanguage() ?? null;
                 $field  = $this->content($lang)->get('date');
                 $date   = $field->isEmpty() ? 'now' : $field;
+                // TODO: in 3.6.0 throw an error if date() doesn't
+                // return a number, see https://github.com/getkirby/kirby/pull/3061#discussion_r552783943
                 return (int)date($format, strtotime($date));
                 break;
             case 'default':
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -533,7 +533,7 @@ trait PageActions
                 $lang   = $this->kirby()->defaultLanguage() ?? null;
                 $field  = $this->content($lang)->get('date');
                 $date   = $field->isEmpty() ? 'now' : $field;
-                return date($format, strtotime($date));
+                return (int)date($format, strtotime($date));
                 break;
             case 'default':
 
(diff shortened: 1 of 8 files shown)


src/Cms/PageActions.php
```diff
@@ -533,7 +533,7 @@ public function createNum(int $num = null): int
                 $lang   = $this->kirby()->defaultLanguage() ?? null;
                 $field  = $this->content($lang)->get('date');
                 $date   = $field->isEmpty() ? 'now' : $field;
-                return date($format, strtotime($date));
+                return (int)date($format, strtotime($date));
```
lukasbestle, line 538: I'm not sure about this one. `date()` returns a string, so it could be possible that the format produces an output that cannot be reliably converted to int. I think there should be additional validation that the return value of `date()` only contains digits.

afbora: Will the [idate()](https://www.php.net/manual/en/function.idate.php) function work?

distantnative: That could be a breaking change, right? Cause I think currently the casting to int already happens through the return type hint of the method. Adding it in there only makes Psalm happier but doesn't change how it works. Throwing an error would - though it could still be a good idea, but then for 3.6

lukasbestle: @afbora Unfortunately not, it only supports a single-char format, not `Ymd` for example.

@distantnative It would be, right. I think we have two options: Either we leave the change out for now, or (my preferred solution) we change it with `(int)`, but add a TODO comment to it with a reference to this discussion.

bastianallgeier:
What's the state on this? @distantnative should we move it to the next release or 3.6?

distantnative:
I just need to add one comment with a todo for 3.6 over the weekend. Then it's good to go. Think 3.5.1 still works

bastianallgeier:
@distantnative any chance to get this fixed and merged quickly?

distantnative:
@bastianallgeier sorry totally forgot about this. added the TODO message

bastianallgeier (approved):

PR #3065, opened 2021-01-06
Automatically generates language locale code #2427

### LINKED ISSUE #2427: Panel: Creating a Language Does Not Use Displayed Placeholder "PHP locale string"
**Describe the bug**  
When creating a language in the Panel, the dialog gives the very strong impression that if the "PHP locale string" is left unfilled, it will use `en_US` (not sure if this displayed value is dependent on some other configuration).

However, the language file that is created will actually use the value supplied for "Code". 

**To Reproduce**  
1. Create a language in the Panel with values such as:
Screen Shot 2020-02-04 at 3 32 31 PM

2. Inspect the created language file

**Expected behavior**  
The `locale` should be `en_US` as suggested by the dialog. Instead, the value will look like:
```php
<?php

return [
    'code' => 'en-us',
    'default' => false,
    'direction' => 'ltr',
    'locale' => [
        6 => 'en-us'
    ],
    'name' => 'English (US)',
    'translations' => [

    ],
    'url' => NULL
];
```

**Kirby Version**  
`3.3.3`

 - OS: macOS
 - Browser: Safari
 - Version: 13.0.5 (15608.5.11)

afbora:
## Describe the PR

Instead of helps text, I tried to find and set the locale value automatically. 
If it is acceptable, we can complete a few missings and merge the PR.

**Setting Locale Process**
 - If it contains hyphens, divide it by two and capitalize after the hyphen.
 - Otherwise, match it with the local values defined in the system with new `locales` system info.

What do yo think? @bastianallgeier @texnixe @distantnative @lukasbestle 

## Screencast

locale

## Related issues

- Fixes #2427 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/config/api/models/System.php
+++ b/config/api/models/System.php
@@ -38,6 +38,14 @@ return [
         'license' => function (System $system) {
             return $system->license();
         },
+        'locales' => function () {
+            $locales = [];
+            $translations = $this->kirby()->translations();
+            foreach ($translations as $translation) {
+                $locales[$translation->code()] = $translation->locale();
+            }
+            return $locales;
+        },
         'loginMethods' => function (System $system) {
             return array_keys($system->loginMethods());
         },
@@ -114,6 +122,7 @@ return [
             'kirbytext',
             'languages',
             'license',
+            'locales',
             'multilang',
             'requirements',
             'site',
--- a/panel/src/components/Dialogs/LanguageCreateDialog.vue
+++ b/panel/src/components/Dialogs/LanguageCreateDialog.vue
@@ -15,6 +15,7 @@ export default {
   mixins: [DialogMixin],
   data() {
     return {
+      languages: [],
       language: {
         name: "",
         code: "",
@@ -53,8 +54,7 @@ export default {
         },
         locale: {
           label: this.$t("language.locale"),
-          type: "text",
-          placeholder: "en_US"
+          type: "text"
         },
       };
     },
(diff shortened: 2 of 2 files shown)


lukasbestle (approved):
I really love the UX of this! ❤️

afbora:
@lukasbestle This is great, thank you 🥳 

I was also thinking about the language update dialog but do you think we should apply it to the language update dialog?

bastianallgeier:
I love it as well, but what happens if you enter "Türkçe" for example. Most of the times the name will be the full language name, right?

afbora:
### Default behaviours
Only name entered, no manual update on codes or locale.

**English**
screen-capture-601-Settings - Mægazine-localhost

 **Türkçe**

screen-capture-604-Settings - Mægazine-localhost

**Deutsch**

screen-capture-607-Settings - Mægazine-localhost

lukasbestle:
> I love it as well, but what happens if you enter "Türkçe" for example. Most of the times the name will be the full language name, right?

@bastianallgeier The code will be incorrect by default like before, but once you correct the code, the locale will update again automatically.

lukasbestle:
> I was also thinking about the language update dialog but do you think we should apply it to the language update dialog?

@afbora As it's not possible to change the language code in the update dialog, I don't think the locale should update automatically either. So I think it's fine to limit this to the create dialog.

afbora:
@lukasbestle told exactly what I was trying to write 😅 

When I write "Türkçe/Turkish"  in current Kirby versions, I have to change the code `tr` from `tu`. But `locale` was still empty.
Now when I change the code `tr`, `locale` will be updated automatically with `tr_TR`.

afbora:
> As it's not possible to change the language code in the update dialog, I don't think the locale should update automatically either. So I think it's fine to limit this to the create dialog.

Right! I forgot that. Then PR is ready 🚀

bastianallgeier:
That's just awesome!

PR #3072, opened 2021-01-07
Automatic fallback for email security = true #2896

### LINKED ISSUE #2896: Email: `security: true` doesn't work since Kirby 3.4.0
Sending email to a remote server (not localhost) fails since version 3.4.0 and above. Kirby 3.3.6 is working with the same settings without any problems.

Settings:
```
return [
  'email' => [
    'transport' => [
      'type' => 'smtp',
      'host' => 'remote.server.de',
      'port' => 465,
      'security' => true,
      'auth' => true,
      'username' => 'theUsername',
      'password' => 'secretpassword',
    ]
  ]
];
```

When git submodule switching between 3.3.6 (working) and 3.4.0 up to current 3.4.4. you can see the difference.

The error is:

After a long timeout (between 1 or 2 minutes) the following error occurs: 
```
PHPMailer \ PHPMailer \ Exception
SMTP Error: Could not authenticate.
```

in 3.4.4:
```
PHPMailer\PHPMailer\Exception 
…/vendor/phpmailer/phpmailer/src/PHPMailer.php2035
```
image

Switching back to 3.3.6 fixes the problem and the mail is sent without any delays.

Sending mails to smtp localhost is no problem in any version. So it seems that only sending mails to a remote smtp server is buggy starting after 3.3.6.

lukasbestle:
## Describe the PR

Fixed a breaking-change in the PHPMailer dependency that was introduced with Kirby 3.4.0: The `email.transport.security = true` option is now automatically converted to `'tls'` or `'ssl'` depending on the configured port. If no port is given and secure transport is enabled, the port is set to 587 (the common port for SMTP over TLS).

## Related issues

- Fixes #2896

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/src/Email/PHPMailer.php
+++ b/src/Email/PHPMailer.php
@@ -73,6 +73,24 @@ class PHPMailer extends Email
             $mailer->Password   = $this->transport()['password'] ?? null;
             $mailer->SMTPSecure = $this->transport()['security'] ?? 'ssl';
             $mailer->Port       = $this->transport()['port'] ?? null;
+
+            if ($mailer->SMTPSecure === true) {
+                switch ($mailer->Port) {
+                    case null:
+                    case 587:
+                        $mailer->SMTPSecure = 'tls';
+                        $mailer->Port = 587;
+                        break;
+                    case 465:
+                        $mailer->SMTPSecure = 'ssl';
+                        break;
+                    default:
+                        throw new InvalidArgumentException(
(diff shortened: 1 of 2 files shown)


afbora (approved):
Just like I imagined ❤️

lukasbestle:
@afbora I literally copied your code and adapted it to the code around it. 😂

bastianallgeier (approved):

PR #3074, opened 2021-01-07
Revamped video URL parsing for PHP and JS #2955

### LINKED ISSUE #2955: Video Block Error
If you paste this type of youtube url in the video block it will result in an error.

example url: https://www.youtube.com/watch?list=PL7CcGwsqRpSM3w9BT_21tUU8JN2SnyckR&v=21HuwjmuS7A&index=1

Exception
Invalid YouTube source

bastianallgeier:
## Describe the PR

The parser for video URLs now uses the URI class in PHP and the URL parser in JS. They are actually pretty close, which is very cool. The code is almost identical that way and we can simply adapt it on both ends if needed. Tests have also been ported to JS to test the same URL patterns on both sides. 

## Related issues

- Fixes #2955

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/panel/src/helpers/embed.js
+++ b/panel/src/helpers/embed.js
@@ -119,7 +119,7 @@ export default {
         break;
     }
 
-    if (!id.match(/^[0-9]*$/)) {
+    if (!id || !id.match(/^[0-9]*$/)) {
       return false;
     }
 
--- a/panel/src/helpers/embed.js
+++ b/panel/src/helpers/embed.js
@@ -1,0 +1,131 @@
+export default {
+  youtube(url) {
+
+    if (!url.match("youtu")) {
+      return false;
+    }
+
+    let uri = null;
+
+    try {
+      uri = new URL(url);
+    } catch (e) {
+      return false;
+    }
+
+    const path = uri.pathname.split("/").filter(item => item !== "");
+    const first = path[0];
+    const second = path[1];
+    const host = "https://" + uri.host + "/embed";
+
+    const isYoutubeId = (id) => {
+      return id.match(/^[a-zA-Z0-9_-]+$/) !== null;
+    };
+
+    let query = uri.searchParams;
+    let src = null;
+
+    switch (path.join("/")) {
+      case "embed/videos":
+      case "playlist":
+        if (isYoutubeId(query.get("list"))) {
+          src = host + "/videoseries";
+        }
+        break;
+      case "watch":
+        if (isYoutubeId(query.get("v"))) {
+          src = host + "/" + query.get("v");
+
+          if (query.has("t")) {
+            query.set("start", query.get("t"));
+          }
+
+          query.delete("v");
+          query.delete("t");
+        }
+
+        break;
+      default:
+        // short URLs
+        if (uri.host.includes("youtu.be")) {
+          src = 'https://www.youtube.com/embed/' + first;
+
+          if (query.has("t")) {
+            query.set("start", query.get("t"));
+          }
+
+          query.delete("t");
+        } else if (first === "embed" && isYoutubeId(second)) {
+          src = host + "/" + second;
+        }
+    }
+
+    if (!src) {
+      return false;
+    }
+
+    const queryString = query.toString();
+
+    if (queryString.length) {
+      src += "?" + queryString.replace("&", "&amp;");
+    }
+
+    return src;
+
+  },
+  video(url) {
+
+    // YouTube video
+    if (url.includes("youtu")) {
+      return this.youtube(url);
+    }
+
+    // Vimeo video
+    if (url.includes("vimeo")) {
+      return this.vimeo(url);
+    }
+
+    return false;
+
+  },
+  vimeo(url) {
+
+    let uri = null;
+
+    try {
+      uri = new URL(url);
+    } catch (e) {
+      return false;
+    }
+
+    const path = uri.pathname.split("/").filter(item => item !== "");
+
+    let query = uri.searchParams;
+    let id = null;
+
+    switch (uri.host) {
+      case "vimeo.com":
+      case "www.vimeo.com":
+        id = path[0];
+        break;
+      case "player.vimeo.com":
+        id = path[1];
+        break;
+    }
+
+    if (!id.match(/^[0-9]*$/)) {
+      return false;
+    }
+
+    let src = "https://player.vimeo.com/video/" + id;
+
+    const queryString = query.toString();
+
+    if (queryString.length) {
+      src += "?" + queryString.replace("&", "&amp;");
+    }
+
+    return src;
+
+  }
+};
--- a/panel/src/helpers/embed.js
+++ b/panel/src/helpers/embed.js
@@ -19,6 +19,10 @@ export default {
     const host = "https://" + uri.host + "/embed";
 
     const isYoutubeId = (id) => {
+      if (!id) {
+        return false;
+      }
+
       return id.match(/^[a-zA-Z0-9_-]+$/) !== null;
     };
 
@@ -28,7 +32,7 @@ export default {
     // the query params are appended below
     let src = null;
     switch (path.join("/")) {
-      case "embed/videos":
+      case "embed/videoseries":
       case "playlist":
         if (isYoutubeId(query.get("list"))) {
           src = host + "/videoseries";
@@ -49,7 +53,7 @@ export default {
         break;
       default:
         // short URLs
-        if (uri.host.includes("youtu.be")) {
+        if (uri.host.includes("youtu.be") && isYoutubeId(first)) {
           src = 'https://www.youtube.com/embed/' + first;
 
           if (query.has("t")) {
--- a/panel/src/helpers/embed.js
+++ b/panel/src/helpers/embed.js
@@ -69,7 +69,7 @@ export default {
     const queryString = query.toString();
 
     if (queryString.length) {
-      src += "?" + queryString.replace("&", "&amp;");
+      src += "?" + queryString;
     }
 
     return src;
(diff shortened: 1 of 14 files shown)


panel/src/helpers/embed.js
```diff
@@ -0,0 +1,131 @@
...
+    const host = "https://" + uri.host + "/embed";
+
+    const isYoutubeId = (id) => {
+      return id.match(/^[a-zA-Z0-9_-]+$/) !== null;
+    };
+
+    let query = uri.searchParams;
+    let src = null;
+
+    switch (path.join("/")) {
```
lukasbestle: When reading the code, I was confused what happens with the query params in the resulting URL, because I didn't yet know they were appended at the end.

Suggestion: A quick comment above this that explains what happens:

```suggestion

    // build the correct base URL for the embed,
    // the query params are appended below
    let src = null;
    switch (path.join("/")) {
```

panel/src/helpers/embed.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    if (!src) {
+      return false;
+    }
+
+    const queryString = query.toString();
+
+    if (queryString.length) {
+      src += "?" + queryString.replace("&", "&amp;");
```
lukasbestle: For cleanliness I think this helper shouldn't escape the string as the returned URL is not a valid URL anymore. Does anything speak against escaping when the URL is used in the component?

src/Toolkit/Html.php
```diff
@@ -446,128 +447,110 @@ public static function video(string $url, array $options = [], array $attr = [])
...
      * Embeds a YouTube video by URL in an `<iframe>`
      *
      * @param string $url YouTube video URL
      * @param array $options Query params for the embed URL
-     * @param array $attr Additional attributes for the `<iframe>` tag
+     * @param array $attrs Additional attributes for the `<iframe>` tag
      * @return string The generated HTML
      */
-    public static function youtube(string $url, array $options = [], array $attr = []): string
+    public static function youtube(string $url, array $options = [], array $attrs = []): string
```
lukasbestle: Renaming the param would be a breaking-change because of PHP 8's named arguments. Also it's only there for the YouTube method, not for Vimeo. Is the change really necessary?

src/Toolkit/Html.php
```diff
@@ -446,128 +447,110 @@ public static function video(string $url, array $options = [], array $attr = [])
...
+            case 'vimeo.com':
+            case 'www.vimeo.com':
+                $id = $path->first();
+                break;
+            case 'player.vimeo.com':
+                $id = $path->nth(1);
+                break;
+        }
+
+        if (!preg_match('!^[0-9]*$!', $id)) {
```
lukasbestle:
```suggestion
        if (preg_match('!^[0-9]*$!', $id) !== 1) {
```

src/Toolkit/Html.php
```diff
@@ -446,128 +447,110 @@ public static function video(string $url, array $options = [], array $attr = [])
...
+        $uri    = new Uri($url);
+        $path   = $uri->path();
+        $query  = $uri->query();
+        $first  = $path->first();
+        $second = $path->nth(1);
+        $host   = 'https://' . $uri->host() . '/embed';
+        $src    = null;
+
+        $isYoutubeId = function (string $id): bool {
+            return preg_match('!^[a-zA-Z0-9_-]+$!', $id);
```
lukasbestle, line 540: `preg_match()` does not return a bool, but `1|0|false`. So Psalm will complain about this on higher levels. Fix:

```suggestion
            return preg_match('!^[a-zA-Z0-9_-]+$!', $id) === 1;
```

src/Toolkit/Html.php
```diff
@@ -446,128 +447,110 @@ public static function video(string $url, array $options = [], array $attr = [])
...
+                break;
+
+            // regular video URLs
+            case 'watch':
+                if ($isYoutubeId($query->v) === true) {
+                    $src = $host . '/' . $query->v;
+
+                    $query->start = $query->t;
+
+                    unset($query->v, $query->t);
```
lukasbestle:
```suggestion
                    unset($query->v, $query->t);
```

src/Toolkit/Html.php
```diff
@@ -446,128 +447,110 @@ public static function video(string $url, array $options = [], array $attr = [])
...
-        } else {
-            $query = '';
+        // append all query parameters
+        foreach ($options as $key => $value) {
+            $query->$key = $value;
         }
 
-        $url = 'https://' . $domain . '/' . $uri . $id . $query;
+        // allow fullscreen mode by default
+        $attrs = array_merge(['allowfullscreen' => true], $attrs);
```
lukasbestle: This change (extracting the `array_merge()` into its own line) is not mirrored in the `vimeo()` method. In contrast, the `vimeo()` method already appends the query string. I think the implementation should be mostly consistent for better readability.

lukasbestle (commented):
It would be great to also support other video platforms/embed types in the future, see https://kirby.nolt.io/170. I'm torn if we should use this opportunity to already offer this feature. But in the end we can still support it later, there's nothing that really stops us from fixing it first and then improving it again later.

My comments on the YouTube JS method also apply to the Vimeo one that uses identical code.

panel/src/helpers/embed.js
bastianallgeier: The only reason for escaping it here is to keep the results identical with the ones in PHP.

src/Toolkit/Html.php
bastianallgeier: Ah, I still have to get used to the PHP 8 stuff. No, it's not necessary. I will revert it.

panel/src/helpers/embed.js
bastianallgeier: I just fixed it. I think we should also stop encoding ampersands in PHP. I'm not sure why we are doing it anyway. But this should be done in a different PR. I don't know how deep that problem will go.

bastianallgeier:
> It would be great to also support other video platforms/embed types in the future, see https://kirby.nolt.io/170. I'm torn if we should use this opportunity to already offer this feature. But in the end we can still support it later, there's nothing that really stops us from fixing it first and then improving it again later.
> 
> My comments on the YouTube JS method also apply to the Vimeo one that uses identical code.

Yes, it would definitely be cool to support other platforms. But that should happen in a different PR and version.

afbora:
> Cannot read property 'match' of null

Related line for youtube:
https://github.com/getkirby/kirby/blob/fix/2955-video-block-error/panel/src/helpers/embed.js#L22

Steps for youtube:
1. Go to panel
2. Enter invalid video url that `https://www.youtube.com/watch?list=zv=21HuwjmuS7A&index=1`
3. See error

Steps for vimeo:
1. Go to panel
2. Enter invalid video url that `https://vimeo.com./115805751`
3. See error

afbora:
Doesn't work for all vimeo links: `groups`, `channels`, `album`

- https://vimeo.com/channels/covid19stories/495450370
- https://vimeo.com/groups/animation/videos/499212458
- https://vimeo.com/album/2222222/video/11111111 (I couldn't find a working video but it is said that there are albums)

bastianallgeier:
@afbora I just fixed the invalid ID check. 

The other Vimeo URLs need to be supported in a different PR.

afbora:
> I just fixed the invalid ID check.

Error still continue.

> The other Vimeo URLs need to be supported in a different PR.

OK 👍

bastianallgeier:
@afbora I forgot the same check for Vimeo. Should be fixed now.

afbora:
> Block error: "Kirby\Toolkit\Html::Kirby\Toolkit\{closure}(): Argument #1 ($id) must be of type string, null given, called in \htdocs\test\kirby\2955\kirby\src\Toolkit\Html.php on line 525" in block type: "video"

When the video link is invalid, the panel works correctly now 👍 but a similar error continues on the frontend.

distantnative:
Since this seems to include breaking changes, I think we should push it to 3.6

bastianallgeier:
@distantnative yep, I'm not getting it ready today anyway. I'm constantly falling through the test coverage checks.

afbora:
I have added the required test for Coverage. If there's nothing missing, I think it's ready 🚀

fabianmichael:
Sorry, I’m pretty late to the party … now that this has basically been resolved for now. I wrote down some thoughts in Nolt about using oEmbed for handling this in the future – could make our lives easier, because we could possibly avoid to write additional code for every new provider. See: https://kirby.nolt.io/170

afbora:
@fabianmichael I think new providers definitely need to be supported.. But at first the issue was related to parsing the video url, then we realized that PHP and JS were not working with the same validation. This PR was to solve these two issues radically. I'm sure the new video providers will be the next step. See @lukasbestle [comment](https://github.com/getkirby/kirby/pull/3074#pullrequestreview-564811527) and after @bastianallgeier [comment](https://github.com/getkirby/kirby/pull/3074#issuecomment-757889675).

fabianmichael:
@afbora it’s not something, that I would personally require at the moment. Just wanted to give a hint to the oEmbed solution, which might be a way of making this easier in the future.

afbora:
@fabianmichael I understood properly now. Yes you right, it will definitely provide much convenience for the future 👍

bastianallgeier:
I've rebased and adjusted it again, after the last updates on the video tag.

afbora:
Will there be any improvements to the `oEmbed` mentioned by Fabian or should we test the PR as it is?

bastianallgeier:
@afbora no, I don't want to add oEmbed to it in this PR. It's a completely different story with two calls for each URL. It's a fantastic solution, but I think we should still move forward with this and add that later.

distantnative (approved):
LGTM

PR #3075, opened 2021-01-07
Use local time instead UTC getting today #3005

### LINKED ISSUE #3005: Date Field w/ Time: "Today" button inserts from time (maybe Timezone related)
**Describe the bug**  
My PHP server and locale machine are both in the Pacific time zone (`America/Los_Angeles`). My computer's clock says `12:14 PM`, but when I press the "Today" button on the Date field, it inserts `8:14 pm`.

I suspect it's using the UTC time (which would be `8:14 pm`).

Kirby 3.4.5 would have filled the expected `12:14 pm` time in.

```yaml
          created:
            label: Published Date
            type: date
            translate: false
            time:
              step: 1
              notation: 12
            width: 1/3
```

**To Reproduce**  
1. Use above date field
2. Click "Today" button (also occurs if you pick any specific day)
3. If you save, it will be committed as the `8:14 pm` time (in Pacific timezone).

**Expected behavior**  
Time should reflect the "current" time. For Kirby 3.4.5, this was always from the perspective of the server. (I don't think it tried to adjust for the browser's time zone.)

**Screenshots**  
Screen Shot 2020-12-14 at 12 14 15 PM

**Kirby Version**  
`3.5.0-rc.7`

 - OS: macOS
 - Browser: Safari 14.0.1 (16610.2.11.51.8)

afbora:
## Describe the PR

I'm not sure the solution is correct but since we do not use the timezone plugin in dayjs, we are using UTC. Therefore, when merging the UTC time hour in the following method, it takes UTC time. 

https://github.com/getkirby/kirby/blob/develop/panel/src/components/Forms/Calendar.vue#L246-L250

For ex:
 - Local Time     : 2021-01-08 01:55:00
 - UTC Time       : 2021-01-07 22:55:00
 - Merged Time : 2021-01-08 22:55:00

For this reason, I preferred to use local time instead of UTC when getting today. 

## Related issues

- Fixes #3005 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/panel/src/components/Forms/Calendar.vue
+++ b/panel/src/components/Forms/Calendar.vue
@@ -233,7 +233,7 @@ export default {
       return this.datetimes.some(current => date.isSame(current, "day"));
     },
     isToday(day) {
-      return this.toDate(day).isSame(this.$library.dayjs.utc(), "day");
+      return this.toDate(day).local().isSame(this.toToday(), "day");
     },
     next() {
       let next = this.viewDt.clone().add(1, "month");
--- a/panel/src/components/Forms/Calendar.vue
+++ b/panel/src/components/Forms/Calendar.vue
@@ -314,7 +314,7 @@ export default {
       return dt.format("YYYY-MM-DD HH:mm:ss");
     },
     toToday() {
-      return this.$library.dayjs.utc();
+      return this.$library.dayjs.utc().local();
     },
   }
 };


bastianallgeier:
LGTM, but I guess @distantnative should give the final ok for this.

bastianallgeier (dismissed):

distantnative:
This looks like it could solve it, but it's really hard to test for me - essentially why I went for UTC everywhere was to avoid all the timezone issues we have seen before. I'm afraid that they'll be back when mixing UTC and local. But of course it's also wrong to show a UTC time, when clicking on today when you're not in UTC yourself :/

afbora:
So, do you think we can solve the issue permanently if we put UTC/Local mode?

````yaml
date:
  type: date
  offset: local # default mode
  

date:
  type: date
  offset: utc
  
date:
  type: date
  offset: +3
````

distantnative:
I wouldn't introduce a new option. The stored content of the date field is timezone-less. So the interface also has to behave agnostic to a timezone. Pinning a specific timezone as an extra option is probably even more confusing for developers and users.

distantnative:
I tested it a bit and at least in my setup it seems to work. So maybe we can give it a try and merge it - I don't think it would break more than the existing bug. So either it is actually fixed (hooray!) or we have the same bug remaining just caused differently.

distantnative (approved):

PR #3078, opened 2021-01-08
Closes calendar when click outside of it #3037

### LINKED ISSUE #3037: Panel: Date field not possible to close within structure
**Describe the bug**  
Normally, when opening the date picker in a date field, a mouse click outside the date picker or focussing on another field closes it. This is not possible within a structure field (only way to close the date picker without picking a date is to close the structure item entirely).

**To Reproduce**  
1. Create a blueprint with a structure field that contains a date field and some other (e.g. text) field
2. In the panel, choose "Add" for the structure
3. Click on the date field; the date picker opens
4. Now click on the text field or elsewhere within the structure area
5. The date picker remains open and cannot be closed

**Expected behavior**  
The date picker should close when moving focus away from it.

**Kirby Version**  
3.5.0, but this also applies on 3.4.5

**Console output**  
No console output.

**Desktop:**  
 - OS: Linux Mint 19.3
 - Browser: Firefox
 - Version: 84.0.1

**Additional context**  
Potentially related to #3004 ?

afbora:
## Describe the PR

I tried to do it by triggering `focusout` and `blur` events but it didn't work. Because when blur, the `update` event is triggered. So I did it a little hard way but still wish it was an easier technique.

## Related issues

- Fixes #3037 

## Ready?

- [x] Passing all unit tests

### DIFF
--- a/panel/src/components/Forms/Field/DateField.vue
+++ b/panel/src/components/Forms/Field/DateField.vue
@@ -76,12 +76,28 @@ export default {
       this.datetime = value;
     }
   },
+  created() {
+    // binds click event that clicking out of the calendar closes it
+    this.outsideClick = (event) => {
+      const calendarInput = this.$el.querySelector(".k-calendar-input");
+      if (calendarInput && this.$el.contains(event.target) === false) {
+        this.close();
+      }
+    };
+
+    document.addEventListener("click", this.outsideClick, true);
+  },
+  destroyed() {
+    document.removeEventListener("click", this.outsideClick);
+  },
   methods: {
+    close() {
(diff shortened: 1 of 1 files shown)


distantnative:
To be honest, that looks like too much of an hack for me. I have no better solution yet, but I think there needs to be a cleaner way to do this - maybe we need to rewrite how blur triggers update etc.

afbora:
Got it, it should definitely be solved in an easier way. Either merge it or move it to the next milestone 👍

distantnative:
I think I found a solution. Would be great if you could give it a try: https://github.com/getkirby/kirby/pull/3079

PR #3079, opened 2021-01-10
Date field: close calendar on blur #3037

### LINKED ISSUE #3037: Panel: Date field not possible to close within structure

distantnative:
## Describe the PR

- The input now emit a `blur` event separately from `update` (for date time input it only gets emitted when the focus is outside the input (to avoid emitting it when switching between the two child inputs)
- The field closes the calendar (if existing) on the `blur` event

## Related issues

- Fixes #3037

### DIFF
--- a/panel/src/components/Forms/Field/DateField.vue
+++ b/panel/src/components/Forms/Field/DateField.vue
@@ -64,6 +64,7 @@ export default {
     listeners() {
       return {
         ...this.$listeners,
+        blur: this.onBlur,
         enter: this.onSelect,
         focus: this.onFocus,
         input: this.onInput,
@@ -79,6 +80,16 @@ export default {
   methods: {
     focus() {
       this.$refs.input.focus();
+    },
+    onBlur(event) {
+      // close the calendar dropdown on blur on the input
+      // but make sure the click target wasn't the calendar itself
+      if (
+        this.$refs.calendar &&
+        this.$el.contains(event.relatedTarget) === false
+      ) {
+        this.$refs.calendar.close();
+      }
     },
     onUpdate(value) {
       this.$emit("input", value);
--- a/panel/src/components/Forms/Input/DateInput.vue
+++ b/panel/src/components/Forms/Input/DateInput.vue
@@ -215,7 +215,9 @@ export default {
         this.select();
       });
     },
-    onBlur() {
+    onBlur(event) {
+      this.$emit("blur", event);
+
       if (!this.parsed) {
         this.input = null;
       }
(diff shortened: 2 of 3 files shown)


afbora:
Yeah it closes well but I can't click anymore in calendar 🙈

distantnative:
Indeed for pure date field, there's a bug. Does it work for you for date with time?

distantnative:
Pushed another attempt.

afbora:
Clickable yes but there was a problem I encountered while trying to solve it. I couldn't figure it out. 

After opening the calendar, if you click anywhere without a button, it closes immediately.

close-calendar-1

distantnative:
Interestingly it seems that when I first click the arrows to go to a different month, I can also click the empty space and it doesn't close.

afbora:
I know why. We get the `relatedTarget` from the blur event instead of the click event. Since there is no element to focus on where we clicked, it does not know that it is in the calendar. That's why I binded it to the click event.

https://github.com/getkirby/kirby/blob/d093f9a55b15034089b8525cab7ca2401f325532/panel/src/components/Forms/Field/DateField.vue#L79-L89

panel/src/components/Forms/Field/DateField.vue
```diff
@@ -80,6 +81,16 @@ export default {
     focus() {
       this.$refs.input.focus();
     },
+    onBlur(event) {
+      // close the calendar dropdown on blur on the input
+      // but make sure the click target wasn't the calendar itself
+      if (
+        this.$refs.calendar &&
+        this.$el.contains(event.relatedTarget) === false
```
afbora, line 89:
```suggestion
        this.$el.contains(event.target) === false
```
Can you try this Like it works for me.

afbora:
I couldn't figure it out without hacking for all click scenarios 🙄

bastianallgeier:
Ok, let's move this to the next release.

distantnative:
I will start over rather

PR #3081, opened 2021-01-12
3.5.1-rc.1

bastianallgeier:


texnixe (approved):

lukasbestle:
@bastianallgeier Don't we still use release branches until the final release?

bastianallgeier:
@lukasbestle normally yes, but I was a bit confused with the more strict PR workflow yesterday. I think we need an additional checklist for pre-releases. But as soon as we need to make changes for the pre-release, we should add the branch.

lukasbestle:
@bastianallgeier I think the following workflow should work until we have the automated GitHub Actions release workflow:

1. Push the current state of `develop` to `release/3.5.1`
2. Then create a new temporary feature branch like `feature/release-3.5.1-rc.1`
3. Create a PR from the feature branch to the release branch

bastianallgeier:
It probably even makes sense to use `release/3.5.1-rc.1` for the PR. Then we create a branch for the pre-release which points to the release branch. That's something that feels pretty logic to me.

lukasbestle:
I think this won't work as `release/*` branches are protected as well, so they can't be pushed to directly with new commits.

bastianallgeier:
Ah, ok. Then maybe `rc/3.5.1-rc.1`

lukasbestle:
Sure, why not. Once we have the automatic release workflow, we won't need this anyway. :)

PR #3091, opened 2021-01-15
CI: Unpin Psalm version again

lukasbestle:
The CI issue we were having yesterday [only occurs on PHP < 8.0](https://github.com/vimeo/psalm/issues/5007#issuecomment-761189112). On PHP 8.0 everything works fine, so we can now use the latest Psalm version in CI again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
 
     runs-on: ubuntu-latest
     env:
-      php: "7.4"
+      php: "8.0"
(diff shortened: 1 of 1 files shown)


afbora (approved):
I just wonder that when the psalm issue is fixed, will we go back to php 7.4?

lukasbestle:
> I just wonder that when the psalm issue is fixed, will we go back to php 7.4?

I don't think we need to. I used PHP 7.4 in the beginning because I expected it to run more reliably, but now that PHP 8.0 has "settled" a few weeks after its release, I think we can continue to use PHP 8.0.

PR #3108, opened 2021-01-25
fix: make argument for Response::redirect() nullable

pReya:
Makes the second parameter (HTTP status code) to `redirect()` in the `class Response` nullable. It already has a default parameter set to `302`, which currently will only work if it's being called with `redirect('bla', 0)`, but not with `redirect('bla')` – which is a little odd in my opinion. The default parameter should also work, when no second parameter is given.

Also, in [`kirby/src/Cms/Response.php` (line 26)](https://github.com/getkirby/kirby/blob/551e2f427eaf18e471df1c304552ab3de35e37c6/src/Cms/Response.php#L26) it is already nullable. So calling this child method with `redirect('bla')` will crash, because the parent class does not allow it.

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -254,7 +254,7 @@ class Response
      * @param int $code
      * @return self
      */
-    public static function redirect(string $location = '/', int $code = 302)
+    public static function redirect(string $location = '/', ?int $code = 302)
     {
         return new static([
             'code' => $code,


bastianallgeier (approved):

bastianallgeier:
Thanks for the fix!

lukasbestle:
@bastianallgeier @pReya This reverts commit https://github.com/getkirby/kirby/commit/521e2b63b664d1db1950d0caa6b31155fc601601 where we have explicitly made the param non-nullable.

To be honest, I think it should be the other way around: `Kirby\Cms\Response::redirect()` should be non-nullable and have a default param of `302` like the class it extends. Otherwise the code would become more messy and not less messy over time. I mean: What is a `$code` of `null` supposed to be? `302` is much more clear and also avoids bugs in the future.

lukasbestle:
@bastianallgeier @afbora I think we should fix or revert this before the RC. Fixing this later would be an unnecessary breaking-change. I can prepare a PR, but I'm not sure if I can finish it before this evening because I need to go soon.

pReya:
> @bastianallgeier @pReya This reverts commit [521e2b6](https://github.com/getkirby/kirby/commit/521e2b63b664d1db1950d0caa6b31155fc601601) where we have explicitly made the param non-nullable.
> 
> To be honest, I think it should be the other way around: `Kirby\Cms\Response::redirect()` should be non-nullable and have a default param of `302` like the class it extends. Otherwise the code would become more messy and not less messy over time. I mean: What is a `$code` of `null` supposed to be? `302` is much more clear and also avoids bugs in the future.

I have no feelings about this. Either way is fine. I just wanted to make sure, both function signatures match, b/c I ran into an error due to this inconsistency between the two calls.

PR #3110, opened 2021-01-26
Separates caches for draft mode #3082

### LINKED ISSUE #3082: site()->index(true) returns no drafts in specific circumstances
**Describe the bug**  
The collection returned by `site()->index(true)` does not include drafts (as it should as per specification) if it is called by a `page.update:after` hook for a page whose blueprint includes an option query that makes use of `index`.

**To Reproduce**  
1. Install a plain Starterkit
2. Add the following to `site/config.php`:

```php
'hooks' => [
        print_r(site()->index(true));
        die;
    }
]
```

3. Open a note in the panel and edit some text.
4. After pressing "Save", the JSON debugger screen shows a Pages Object with 22 pages (incl. `notes/in-the-jungle-of-sumatra` which is a draft page). So up to here, everything works as it should.
5. In `site/blueprints/note.yml`, replace `tags: true` with:

```yml
tags:
  type: tags
  label: Tags
  options: query
  query: site.index.pluck("tags", ",", true)
```

6. Open the same note as before in the panel and edit some text.
7. After pressing "Save", the JSON debugger screen again shows a Pages Object, but only with 20 pages (two drafts, incl. `notes/in-the-jungle-of-sumatra`, are no longer returned). 
8. Replace `query: site.index.pluck("tags", ",", true)` with `query: site.children.pluck("tags", ",", true)` or anything else that does not contain `index` and try again -- now the returned collection contains drafts again.

**Expected behavior**  
`site()->index(true)` should always return a Pages Collection that contains all pages incl. all drafts of the entire site.

**Kirby Version**  
3.5.0

**Additional context**  
Documentation of the original discovery of the bug here: https://forum.getkirby.com/t/in-some-contexts-site-index-true-returns-no-drafts/20776

afbora:
## Describe the PR

As @lukasbestle mentioned [in this comment](https://github.com/getkirby/kirby/issues/3082#issuecomment-766176307), the source of the error was caching. Again I applied the solution suggested by @lukasbestle.

_About `indexWithDrafts` variable name, I would appreciate if you have a more descriptive variable name suggestion._

## Related issues

- Fixes #3082 

## Ready?

### DIFF
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -23,11 +23,18 @@ use Kirby\Exception\InvalidArgumentException;
 class Pages extends Collection
 {
     /**
-     * Cache for the index
+     * Cache for the index only listed and unlisted pages
      *
      * @var \Kirby\Cms\Pages|null
      */
     protected $index = null;
+
+    /**
+     * Cache for the index all statuses also including drafts
+     *
+     * @var \Kirby\Cms\Pages|null
+     */
+    protected $indexWithDrafts = null;
 
     /**
      * All registered pages methods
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
LGTM

PR #3113, opened 2021-01-27
CI: Use PHP 7.4 for analysis for now

lukasbestle:
The error message from PHP-CS-Fixer shouldn't stop execution as we use the `PHP_CS_FIXER_IGNORE_ENV` option, so that's a bug in PHP-CS-Fixer. I will send a PR to them, but this change is the fastest fix until then.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -268,7 +268,8 @@ jobs:
 
     runs-on: ubuntu-latest
     env:
-      php: "8.0"
+      # TODO: update to PHP 8.0 when PHP-CS-Fixer fully supports it
+      php: "7.4"
 
     steps:
       - name: Checkout Kirby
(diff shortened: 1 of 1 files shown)


afbora (dismissed):

afbora (approved):

lukasbestle:
@afbora Sorry for having to approve this twice. The new setup should work more reliably until PHP-CS-Fixer officially supports PHP 8.

Next time this happens, we should pin all the tool versions so that they don't break all the time. My hope is that such issues won't happen often, but in the last few weeks we already had two tool updates break things. :(

afbora:
@lukasbestle No problem 👍 

Or should we run other tests for a while at 7.4 instead of 8.0? (Frontend E2E and Backend Analysis)

PR #3118, opened 2021-01-28
Avoid calling global functions in filter()

### LINKED ISSUE #3117: $collection->filter('asin', 'bar') can be misinterpreted to use global `asin()` function
**Describe the bug**  
In Kirby 3.5.0, `$collection->filterBy()` was consolidated with `$collection->filter()` which makes the following scenario more obvious:

If you want to filter by a field name that _also_ happens to be the name of a global function, the `$collection->filter()` method will prefer the global function instead of filtering by the field name.

**To Reproduce**  
1. Attempt to filter a collection with something like `$collection->filter('asin', '!=', '')`

Because `asin()` exists as a global PHP function, you'll receive an error like:
> asin() expects parameter 1 to be float, object given

**Expected behavior**  
Should filter items that have a non-empty `asin` field value.

**Screenshots**  
Screen Shot 2021-01-27 at 5 22 47 PM

**Kirby Version**  
`3.5.1`

neildaniels:
## Describe the PR
The `$collection->filter()` method can erroneously call global functions instead of filtering by fields.

## Related issues
- Fixes #3117

## Ready?

- [ ] CI passes (runs automatically when the PR is created or run `composer ci` locally)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,8 @@ jobs:
     steps:
       - name: Checkout Kirby
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
 
       - name: Install memcached
         uses: niden/actions-memcached@v7
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -252,7 +252,7 @@ class Collection extends Iterator implements Countable
         $split    = $args[1] ?? false;
 
         // filter by custom filter function
-        if (is_callable($field) === true) {
+        if (is_string($field) === false && is_callable($field) === true) {
             $collection = clone $this;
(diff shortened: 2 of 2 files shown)


bastianallgeier (dismissed):
I think this is a great solution.

bastianallgeier:
Sorry, we are only trying to figure out why our test process is hanging. Afterwards we will merge this. 🙈

bastianallgeier (approved):

afbora:
@bastianallgeier Is it a situation that needs to be revert?

https://forum.getkirby.com/t/custom-filter-function-does-not-work-in-3-5-3/21242

bastianallgeier:
@afbora I've replied in the forum.

PR #3135, opened 2021-02-01
Use PHP_EOL for CS-Fixer setup

afbora:
## Describe the PR
When I run the `composer fix` command in Windows, it is trying to fix all the files because it converts end of lines automatically. I used `PHP_EOL` to make it work correctly on all servers/devices.

## Ready?

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -55,4 +55,5 @@ return PhpCsFixer\Config::create()
         'whitespace_after_comma_in_array' => true
     ])
     ->setRiskyAllowed(true)
-    ->setFinder($finder);
+    ->setFinder($finder)
+    ->setLineEnding(PHP_EOL);


lukasbestle (commented):
Shouldn't it always be `"\n"` (which seems to be the default)? If I understand it correctly, setting the line ending to `PHP_EOL` will mean that PHP-CS-Fixer will convert everything to Windows line endings when run on Windows. That's not the intended behavior.

I think this is rather something you should fix in your local setup. As far as I know, Git for Windows automatically converts line endings internally, but maybe you need to configure your editor/IDE to always use Unix line endings in the first place. Then PHP-CS-Fixer won't convert anything either.

afbora:
> setting the line ending to PHP_EOL will mean that PHP-CS-Fixer will convert everything to Windows line endings when run on Windows

@lukasbestle I'm not sure about that.

https://github.com/php/php-src/blob/PHP-7.4.15/main/php.h#L60-L78

Few tests for you:

````php
->setLineEnding("\n")    // Fixed 730 files
->setLineEnding("\r\n")  // Works properly, no files fixed
->setLineEnding(PHP_EOL) // Works properly, no files fixed
````

lukasbestle:
@afbora That seems to be because of your local Git config. Git by default checks out files with `\r\n` on Windows, which is what PHP-CS-Fixer sees and fixes. But in the repo, all files have `\n`. So I think you should configure Git to use `\n` and then check out Kirby again.

This change in the PHP-CS-Fixer config has the risk that PHP-CS-Fixer will actually change the files for people having Git configured with `\n` on Windows.

afbora:
@lukasbestle You're right about that, I've run following command and now works as expected. Thank you so much 👍 

````
git config --global core.autocrlf false
git config --global core.eol lf
````

PR #3143, opened 2021-02-06
Use `basename()` method to avoid relative path in cache key

afbora:
## Describe the PR

Used `basename()` method to avoid relative path in cache key.

## Ready?


bastianallgeier (commented):
I would only use `basename()` to get rid of relative paths and don't worry about sanitizing filenames further. I think it's too risky to create unwanted collisions. @lukasbestle what do you think?

afbora:
@bastianallgeier I think you're right. My implementation was a bit unnecessary.

bastianallgeier (approved):

PR #3116, opened 2021-01-27
Translatable sections info and text props #3109

afbora:
## Describe the PR
Now pages/files section info and text prop translatable.

## Related issues

- Fixes #3109 

## Ready?

- [x] Added unit tests for fixed bug/feature
- [x] Added in-code documentation (if needed)
- [x] CI passes (runs automatically when the PR is created or run `composer ci` locally)
  Running locally requires PHPUnit, PHP-CS-Fixer, Psalm, PHPCPD and PHPMD.

- [x] Checked whether the PR needs documentation, if needed added to the [release docs checklist](https://github.com/getkirby/getkirby.com/pulls)

PR #3152, opened 2021-02-11
Reverts default info texts for pages/files sections

afbora:
## Describe the PR

With 3.5.2 (https://github.com/getkirby/kirby/pull/3116), default info texts for pages sections (slug) and files sections (file id) started to appear. This PR reverts the changes.

## Ready?

### DIFF
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -123,7 +123,7 @@ return [
                     'id'       => $file->id(),
                     'icon'     => $file->panelIcon($image),
                     'image'    => $image,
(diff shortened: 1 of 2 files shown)


bastianallgeier (commented):
~~I think we should actually fix this in the props definition. Otherwise the info prop is already wrongly defined and this could cause other issues.~~

Ignore my comment. Your fix is correct.

bastianallgeier (approved):

afbora:
Actually I just reverted the change. I think let's do it (props defination) in 3.5.4.

PR #3158, opened 2021-02-15
Fix regression in FileCache keys that caused overlap of cache keys #3157

### LINKED ISSUE #3157: Routing problems since 3.5.2
I'm encountering this issue since v3.5.2.
If a page has the same `id` as another page (but of course not the same parent), the displayed page can be the wrong one (so it doesn't match with the route / url).

The problem seems to be related to this line : https://github.com/getkirby/kirby/blob/dbd8d2216eeec429d0f19ebea430de8347e7f073/src/Cache/FileCache.php#L71

Which has been modified by this pull request : #3143

afbora:
## Related issues

- Fixes #3157 

## Ready?

### DIFF
--- a/src/Cache/FileCache.php
+++ b/src/Cache/FileCache.php
@@ -4,6 +4,7 @@ namespace Kirby\Cache;
 
 use Kirby\Toolkit\Dir;
 use Kirby\Toolkit\F;
+use Kirby\Toolkit\Str;
 
 /**
  * File System Cache Driver
@@ -68,7 +69,41 @@ class FileCache extends Cache
      */
     protected function file(string $key): string
     {
-        $file = $this->root . '/' . basename($key);
+        // strip out invalid characters in each path segment
+        // split by slash or backslash
+        $keyParts = [];
+        foreach (preg_split('#([\/\\\\])#', $key, 0, PREG_SPLIT_DELIM_CAPTURE) as $part) {
+            switch ($part) {
+                // forward slashes don't need special treatment
+                case '/':
+                    break;
+                
+                // backslashes get their own marker in the path
+                // to differentiate the cache key from one with forward slashes
+                case '\\':
+                    $keyParts[] = '_backslash';
+                    break;
+                
+                // empty part means two slashes in a row;
+                // special marker like for backslashes
+                case '':
+                    $keyParts[] = '_empty';
+                    break;
+                
+                // an actual path segment
+                default:
+                    // check if the segment only contains safe characters;
+                    // underscores are *not* safe to guarantee uniqueness
+                    // as they are used in the special cases
+                    if (preg_match('/^[a-zA-Z0-9-]+$/', $part) === 1) {
+                        $keyParts[] = $part;
+                    } else {
+                        $keyParts[] = Str::slug($part) . '_' . sha1($part);
+                    }
+            }
+        }
+
+        $file = $this->root . '/' . implode('/', $keyParts);
 
         if (isset($this->options['extension'])) {
             return $file . '.' . $this->options['extension'];
--- a/tests/Cache/FileCacheTest.php
+++ b/tests/Cache/FileCacheTest.php
@@ -82,27 +82,104 @@ class FileCacheTest extends TestCase
         $cache = new FileCache([
             'root' => $root = __DIR__ . '/fixtures/file',
         ]);
-        $this->assertSame($root . '/test', $method->invoke($cache, '/test'));
+        $this->assertSame($root . '/_empty/test', $method->invoke($cache, '/test'));
 
         $cache = new FileCache([
             'root' => $root = __DIR__ . '/fixtures/file',
         ]);
-        $this->assertSame($root . '/test', $method->invoke($cache, '../../test'));
+        $this->assertSame($root . '/test/_empty', $method->invoke($cache, 'test/'));
 
         $cache = new FileCache([
             'root' => $root = __DIR__ . '/fixtures/file',
         ]);
-        $this->assertSame($root . '/hype-test', $method->invoke($cache, './hype-test'));
+        $this->assertSame($root . '/test/_backslash/foo/bar', $method->invoke($cache, 'test\\foo/bar'));
 
         $cache = new FileCache([
             'root' => $root = __DIR__ . '/fixtures/file',
         ]);
-        $this->assertSame($root . '/dash.test', $method->invoke($cache, './dash.test'));
+        $this->assertSame($root . '/test/_backslash/_empty/foo/_backslash/bar', $method->invoke($cache, 'test\\/foo\\bar'));
 
         $cache = new FileCache([
             'root' => $root = __DIR__ . '/fixtures/file',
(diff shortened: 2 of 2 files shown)


lukasbestle (commented):
This solution won't work on Windows servers. If the cache key contains a backslash, Windows will happily traverse the file system again. Also there could be other weird edge cases.

Maybe we can solve it like this:

```php
$key = substr(sha1($key), 0, 10) . '_' . basename($key);

$file = $this->root . '/' . $key;
```

The disadvantage is that all cache files would be in one flat directory. It would be the safest solution, but the least practical.

The alternative would be an actual root check using `realpath()`. Basically the `$file` path would be checked against `$this->root`. If the `realpath()` of `$file` does not start with `$this->root`, an Exception is thrown.

Unfortunately this won't work out of the box: `realpath()` only works with existing files, but the cache files likely don't already exist. So we'd need a solid implementation of `realpath()` that works with "virtual" paths.

These are all ideas and I'm honestly not sure what would be the best way over all. My code example above would be the absolute safest and easiest though, with the mentioned UX problem.

afbora:
Thank you for your great ideas and suggestions. I'm always grateful 🙏 

Actually, yes it would be great if the `realpath()` function worked on files that do not exist. That would be the solution that should have been.

There does not seem to be any other way to clear the relative path than I did. Except for the disadvantage, I think key encryption is an appropriate solution and it sure looks safer against clearing manuel way.

Meanwhile we are cleaning the windows backslashes in several places within Kirby. Isn't it a reference to us?

lukasbestle:
I found another edge-case that will circumvent the manual replacement:

```
....//your-file.txt

will become:

../your-file.txt
```

I think there is no replacement/path validation solution that will completely prevent any kind of directory traversal. It would be a game of cat and mouse.

The only solutions I can think of are either allow-listing the character set (`a-zA-Z0-9-/` or something, which would be a breaking-change though) or stripping out everything besides the `basename()` and adding a hash of the full cache key (as in my example above with the mentioned disadvantage).

An alternative could be to replace every invalid character (= every character that is not in the allow-list) with an underscore for example. Then all cache keys would still work, but there could be overlaps if multiple cache keys only differ in their non-allowed characters. For our page cache we should be fine though.

afbora:
What do you think of my first commit that sanitize cache key I got reverted? https://github.com/getkirby/kirby/pull/3143/commits/90e319c6c4cc0c201bfc0a8d7f1cdd3d0448ab13

distantnative:
It probably still shares some of the problems @lukasbestle raised. But maybe it would work to prevent directory traversal to replace any multiple dots (`..` or `...` or `....`) with a single dot. Might have the least cases where this causes a breaking change (it's still one for these few)?

afbora:
We can remove all dots about relative easily with regex? 🤔 

````js
Regex
[.]{1,}\/

Examples
./
../
.../
..../
...../

````

**Comment by lukasbestle** [comment]
I'm not comfortable with any solution that involves blocking specific characters. Attackers would always be able to circumvent it by finding malicious characters that aren't yet blocked (be it null bytes, weird Unicode characters etc.) and combining them in weird ways.

That's why I think we need to limit the character set. Either with an allow-list that converts every other byte with an underscore or by hashing the full cache key and only allowing safe characters in the human-readable suffix.

**Comment by afbora** [comment]
Frankly, I don't know which one is the right way. Allowing characters or hashing.. 🤷‍♂️

**Comment by lukasbestle** [comment]
@bastianallgeier What do you think?

I'm currently leaning towards the allow-list with automatic replacement of invalid characters. It could create an overlap of cache keys, but only for cache keys with invalid characters. The more "common" cache keys would be unaffected.

Advantage: We could still keep the directory structure that is easy to understand and makes it easy to selectively clear parts of the cache by prefix.

**Comment by bastianallgeier** [comment]
I agree with the allow list for characters. We need to make 100% sure though that we don't shoot ourselves in the foot with the page cache. This is the top priority. Page IDs must fully work.

**Comment by afbora** [comment]
@lukasbestle if you have time to look at this, can you taken care of it?

**Comment by lukasbestle** [comment]
@bastianallgeier One issue came to my mind: Can we safely assume that page IDs only ever use "sluggified" characters? For example if someone manually uses Unicode URIs with Arabic or Chinese text, all characters will be replaced with underscores and it will again not work.

If a 100 % working page cache is the top priority, the allow-list is likely going to cause problems in such cases.

**Comment by bastianallgeier** [comment]
Maybe this is a stupid idea, but what if we split the cache ID by `/`, hash every element in the array and join them again by `/`. We would end up with cache keys that will not collide, still could make use of nested folders, but could never step up in the filesystem.

**Comment by lukasbestle** [comment]
That's awesome! We could even combine the two ideas: We split by `/` and then check each part against the allow-list. If the part contains any invalid chars, that part is hashed, otherwise it is left as is.

**Comment by bastianallgeier** [comment]
@lukasbestle what about doing this for each part:  

```php
Str::slug($part) . '-' . sha1($part)
```

The keys would get a bit longer, but they would stay unique and readable at the same time.

bastianallgeier:
We should probably also filter parts which end up being empty after basename. ../ and ./ would automatically be removed that way.

lukasbestle:
Yep, that's great! And again we only need to do this for parts with invalid chars.

lukasbestle:
I will continue working on this in this PR. Please leave it open for now.

lukasbestle:
There were two more edge-cases, but on the reliability side, not on the security side:

- If there is a double slash in the cache key like `my//awesome/cache/key`, the `FileCache` would have previously used the same file as for `my/awesome/cache/key`. This is now fixed with a new `_empty` placeholder.
- Backslashes were the same as slashes. There is now a `_backslash` placeholder for these cases.

I think we should be good to go now. :)

afbora:
I think I have one more bad news.
I hope this is expected. I've reliazed that relative path works for `prefix` too 😱

````php
return [
    'cache' => [
        'pages' => [
            'active' => true,
            'prefix' => '../../kirby'
        ]
    ]
];

lukasbestle:
The question is: Is the prefix ever user-provided or only ever set in the configuration?

If it can be user-provided, we should fix it as well, but my current understanding is that the prefix cannot be manipulated by attackers.

afbora:
I know `prefix` option can only be changed in config.

lukasbestle:
I just looked into it as well. Users *could* manually create a `FileCache` instance with unsafe arguments, but I'd say that's not something that's common and therefore not worth the additional complexity in the `FileCache` implementation.

afbora:
We can move this process that clears the relative path to the `F` or `File` toolkit class as a helper method.

lukasbestle:
Do you mean something like `F::safePath()`? I think that's a good idea in general! However it's pretty specific to such file system storage use-cases, so maybe we could move it out once we need it somewhere else. Otherwise the `F` class gets too long for no reason.

afbora:
> Do you mean something like F::safePath()?

Yes, i meant this 👍

bastianallgeier (approved):

PR #3163, opened 2021-02-19
Add email mark to writer field

sylvainjule:
## Describe the PR
This PR adds an `email` mark to the Writer field, that I have found myself to need on almost every websites I have used the writer field on.

It is completely unopinionated as of how to obfuscate the address. I find it more convenient to preg_replace them to obfuscate them how I need afterwards, but I don't know how you'd like to handle it.

## Related issues
[kirby.nolt.io/185](https://kirby.nolt.io/185)

## Ready?
- [ ] CI passes (runs automatically when the PR is created or run `composer ci` locally)

### DIFF
--- a/panel/src/components/Writer/Writer.vue
+++ b/panel/src/components/Writer/Writer.vue
@@ -65,11 +65,6 @@ import Toolbar from "./Extensions/Toolbar.js";
 import ToolbarComponent from "./Toolbar.vue";
 
 export const props = {
-  components: {
-    "k-writer-link-dialog": LinkDialog,
-    "k-writer-email-dialog": EmailDialog,
-    "k-writer-toolbar": ToolbarComponent,
-  },
   props: {
     breaks: Boolean,
     code: Boolean,
@@ -114,6 +109,7 @@ export const props = {
 
 export default {
   components: {
+    "k-writer-email-dialog": EmailDialog,
     "k-writer-link-dialog": LinkDialog,
     "k-writer-toolbar": ToolbarComponent,
   },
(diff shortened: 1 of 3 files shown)


mrflix:
I'd love to see this in the core. The writer field is really nice inside – especially in blocks. I just built a simple website for a client and I had two places where they fill in emails: a team page and an address block in the footer.

Btw this is what it looks like:
Screenshot 2021-03-16 at 16 29 21

And the dialog:
Screenshot 2021-03-16 at 16 30 15

In my case the email should be visible so I filled in the email in both the mail and title field. Turns out it I wouldn't have had to do it. A hint below the title field explaining that its optional or that it falls back to the email would be nice.

bastianallgeier:
I agree! We should definitely add this to the core.

distantnative:
Switched to draft until we tackle the targeted release (helps us to keep a better overview of PRs). Will activate it as soon as 3.6. is on the map (after 3.5.5).

afbora (commented):
Can you replace hard coded texts (`Email`, `Title`) with translation keys?

panel/src/components/Writer/Writer.vue
```diff
@@ -58,6 +65,11 @@ import Toolbar from "./Extensions/Toolbar.js";
 import ToolbarComponent from "./Toolbar.vue";
 
 export const props = {
+  components: {
+    "k-writer-link-dialog": LinkDialog,
+    "k-writer-email-dialog": EmailDialog,
+    "k-writer-toolbar": ToolbarComponent,
+  },
```
afbora: @distantnative no longer needed these in the `develop` branch?

bastianallgeier: Why do you think they are no longer needed?

afbora: A secondary `export` came in the develop branch, I moved it to the wrong place while rebase, I fixed it.

bastianallgeier (approved):

PR #3166, opened 2021-02-21
Deprecated: improve docs and warnings

distantnative:
- `@deprecated` messages need to wrap alternative class/methods in ` for automatic linking on the website
- Add `@todo` messages for when to remove deprecated code
-  Use `@deprecated` more instead of other comments
- Add more warnings via `deprecated()` helper

Fixes https://github.com/getkirby/getkirby.com/issues/1222

### DIFF
--- a/config/api/routes/pages.php
+++ b/config/api/routes/pages.php
@@ -36,7 +36,10 @@ return [
     [
         'pattern' => [
             'pages/(:any)/blueprints',
-            // Deprecated: remove in 3.6.0
+            /**
+             * @deprecated
+             * @todo remove in 3.6.0
+             */
             'pages/(:any)/children/blueprints',
         ],
         'method'  => 'GET',
--- a/config/api/routes/site.php
+++ b/config/api/routes/site.php
@@ -49,7 +49,10 @@ return [
     [
         'pattern' => [
             'site/blueprints',
-            // Deprecated: remove in 3.6.0
+            /**
+             * @deprecated
+             * @todo remove in 3.6.0
+             */
             'site/children/blueprints',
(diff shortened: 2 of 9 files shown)


distantnative:
@lukasbestle the codecov patch check also seems to have an issue. Or I don't get what I'm doing wrong

afbora:
I guess it's due to the newly added `deprecated()` helper lines in `src/Cms/Block.php` and `src/Toolkit/I18n.php`. May be we can use `// @codeCoverageIgnore` in that lines.

lukasbestle:
It's the `I18n::fallback()` method in particular, which is marked with a [relative coverage of 0.00 %](https://codecov.io/gh/getkirby/kirby/commit/2823930d8759ea6407effbff436c14b78aa25a59). You can read more about this in their [docs](https://docs.codecov.io/docs/codecov-delta).

However I don't know why that happens. We have `ignoreDeprecatedCodeUnits` enabled in PHPUnit, so it shouldn't report these lines.

Unrelated: I think we should not add deprecation warnings to these methods for now. Adding the warning will be a breaking-change, so maybe the methods that don't already have warnings should get them in 3.6.0 and then be removed in 3.7.0. It could look like this:

```
* @deprecated 3.5.1 Use `\Kirby\Toolkit\I18n::fallbacks()` instead
* @todo Add warning in 3.6.0
* @todo Remove in 3.7.0
```

distantnative:
@lukasbestle Yes, that makes a lot of sense!

distantnative:
Changed it.

lukasbestle (approved):

PR #3167, opened 2021-02-21
Fix return types: $this, static and self

distantnative:
Fixes https://github.com/getkirby/getkirby.com/issues/1223

### DIFF
--- a/src/Cms/Blueprint.php
+++ b/src/Cms/Blueprint.php
@@ -219,7 +219,7 @@ class Blueprint
      * @param string $name
      * @param string|null $fallback
      * @param \Kirby\Cms\Model $model
-     * @return static
+     * @return static|null
      */
     public static function factory(string $name, string $fallback = null, Model $model)
     {
--- a/src/Cms/Blueprint.php
+++ b/src/Cms/Blueprint.php
@@ -219,7 +219,7 @@ class Blueprint
      * @param string $name
      * @param string|null $fallback
      * @param \Kirby\Cms\Model $model
-     * @return self
+     * @return static
      */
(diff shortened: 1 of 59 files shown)


distantnative:
@lukasbestle If I see this correctly, the checks are failing because traits themselves can't really return `static`? Which I would kinda get but am surprised `self` worked before.

lukasbestle:
Seems to be a bug in Psalm that only happens because of the `parent::update()` calls in these methods: https://github.com/vimeo/psalm/issues/5264

I think we should revert the changes in the two affected classes and add a TODO marker to them with a reference to the Psalm issue. Then we can already merge the rest.

src/Cms/Blueprint.php
```diff
@@ -219,7 +219,7 @@ public static function extend($props): array
      * @param string $name
      * @param string|null $fallback
      * @param \Kirby\Cms\Model $model
-     * @return self
+     * @return static
```
lukasbestle:
```suggestion
     * @return static|null
```

src/Form/Fields.php
```diff
@@ -23,7 +23,7 @@ class Fields extends Collection
      *
      * @param string $name
      * @param object|array $field
-     * @return self
+     * @return static
```
lukasbestle: The return value is inherited from `Toolkit\Collection::__set()`

```suggestion
     * @return $this
```

lukasbestle (commented):
Nice work, that turned out to be quite a long list of changes! 👍🙇

I only found two errors, the rest looks good.

lukasbestle (approved):

PR #3172, opened 2021-02-23
Removes `<p>` and `</p>` tags from value

### LINKED ISSUE #3086: List Items get saved with p-Tags in List-Block
When I Input the following list into a list field:

```
- One
- Two
  - Sub Two
- Three
```

the following get's saved into the content file:

```
<ul><li><p>One<\/p><\/li><li><p>Two<\/p><ul><li><p>Sub Two<\/p><\/li><\/ul><\/li><li><p>Three<\/p><\/li><\/ul>
```

The p-tags in every listitem result in the following output with a prose class to style the list:

99947162-d7370f00-2d77-11eb-9d17-1a6941f7a9ec

Is this the expected behaviour to generate those p-Tags?

I think, that lists with no p-Tags are easier to style and that this is also the behaviour of markdown lists getting converted to html.

It seems that this is a prosemirror bug and I add it here to track the status: 
https://discuss.prosemirror.net/t/removing-the-default-paragraph-p-inside-a-list-item-li/2745/8

### LINKED ISSUE #3168: List Block: Cursor jumps to end of block after each keystroke
**Describe the bug**  
In list blocks in the panel blocks field the cursor jumps to the end of the block after each typed character.
Additionally, a "space" is not accepted as the last character in the block, which makes writing separate words impossible. You can add space in the middle of the text, though (with the jumping cursor issue as described).

**To Reproduce**  
Add a blocks field to a page with the list block enabled.
Open the page in the panel, add a list block and write some text in the block.

**Expected behavior**  
Cursor does not jump to the end of the block and white space at the end of the block is accepted.

**Kirby Version**  
3.5.3 (currently latest)

**Console output**  
None

 - OS: Linux Mint 20.1
 - Browser: Firefox
 - Version: 85.0.1

afbora:
## Describe the PR

~~Since the writer field constantly converts the `value` within itself, I solved the `paragraph` cleaning only while saving the data.~~

[See below comment](https://github.com/getkirby/kirby/pull/3172#issuecomment-786699449)

## Related issues

- Fixes #3086 

## Ready?

### DIFF
--- a/panel/src/components/Forms/Input/ListInput.vue
+++ b/panel/src/components/Forms/Input/ListInput.vue
@@ -1,6 +1,7 @@
 <template>
   <k-writer
     ref="input"
+    v-model="list"
     v-bind="$props"
     :extensions="extensions"
     :nodes="['bulletList', 'orderedList']"
@@ -21,9 +22,16 @@ export default {
     },
     value: String
   },
+  data() {
+    return {
+      list: this.value
+    };
+  },
   computed: {
     extensions() {
-      return [new ListDoc()];
+      return [new ListDoc({
+        inline: true
+      })];
     }
   },
   methods: {
@@ -35,21 +43,22 @@ export default {
       let list = dom.querySelector('ul, ol');
 
       if (!list) {
-        this.$emit("input", "");
+        this.$emit("input", this.list = "");
         return;
       }
 
       let text = list.textContent.trim();
 
       if (text.length === 0) {
-        this.$emit("input", "");
+        this.$emit("input", this.list = "");
         return;
       }
 
-      // emits only when there is a change
-      if (html !== this.value) {
-        this.$emit("input", html);
-      }
+      // updates `list` data with raw html
+      this.list = html;
+
+      // emit value with removes `<p>` and `</p>` tags from html value
+      this.$emit("input", html.replace(/(<p>|<\/p>)/gi, ""));
     }
   }
 };
--- a/panel/src/components/Forms/Input/ListInput.vue
+++ b/panel/src/components/Forms/Input/ListInput.vue
@@ -46,7 +46,10 @@ export default {
         return;
       }
 
-      this.$emit("input", html);
+      // emits only when there is a change
+      if (html !== this.value) {
+        this.$emit("input", html);
+      }
     }
   }
 };
--- a/config/fields/list.php
+++ b/config/fields/list.php
@@ -8,5 +8,9 @@ return [
         'marks' => function ($marks = true) {
             return $marks;
         }
-    ]
+    ],
+    'save' => function ($value): string {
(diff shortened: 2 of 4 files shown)


afbora:
screen-capture-826-Test - Mægazine-localhost

If we remove the paragraph, we will prevent the use of more than one paragraph in the list item (sample usage like screenshot). Therefore, if we remove the `<p>` tags, we need to activate the inline mode by default (using `<br>` instead). What do you think? @bastianallgeier

lukasbestle:
I think `<br>` inside list items make sense. That's what I would use in this case.

bastianallgeier:
I agree. Inline mode with br makes sense for me too.

bastianallgeier:
The deluxe version would be to check for multiple paragraphs and in that case keep them and only remove single paragraphs. But I'm not sure if that's actually wanted by users.

afbora:
Clearing the data while saving did not work either. Because the data when the page was first loaded and the data in the editor did not match (parsed with paragraphs), the savebar always appeared.

So I tried to resolve by separating the data used by the editor (`list`) and the saved data (`value`). Javascript codes can be written better using better methods.

panel/src/components/Forms/Input/ListInput.vue
```diff
@@ -35,18 +43,22 @@ export default {
       let list = dom.querySelector('ul, ol');
 
       if (!list) {
-        this.$emit("input", "");
+        this.$emit("input", this.list = "");
```
distantnative, line 46: We might be able to skip this change and the one below, when we add a `watch` on `value` and update `this.list` whenever the `value` prop changes. We might need to do this anyways to avoid unforeseen states.

afbora: If we bind the `value` in watch to `list`, it will not work, due to following line:

https://github.com/getkirby/kirby/blob/7a5707ff2426b843ff40084e4d2c99a94bda418c/panel/src/components/Forms/Input/ListInput.vue#L60-L61

For better understanding:

`list` prop: raw html that includes `p` tags
`value` prop: cleared html that doesn't includes `p` tags

bastianallgeier (dismissed):
It works really well for me. Do you think we need to test it more properly before we merge it or can we move more tests to the pre-release phase?

afbora:
> Do you think we need to test it more properly before we merge it

Tests would always be great. Can you take care of JS tests? 😇

bastianallgeier:
I've added basic list block tests, but it's not possible so far to test the result of the input even in Cypress.

bastianallgeier (approved):

lukasbestle:
This PR broke our CI after merging. A test is failing.
@afbora Could you please take a look and send a PR?

afbora:
@lukasbestle Fix PR #3276

PR #3174, opened 2021-02-23
Fix fatal dialog

bastianallgeier:
## Describe the PR

The fatal iframe for API errors could cause issues itself when it's not accessible. This additional error handling fix will at least catch that and send the error to the console instead.

### DIFF
--- a/panel/src/App.vue
+++ b/panel/src/App.vue
@@ -122,11 +122,13 @@ export default {
     fatal(html) {
       if (html !== null) {
         this.$nextTick(() => {
-          if (this.$refs.fatal && this.$refs.contentWindow) {
+          try {
             let doc = this.$refs.fatal.contentWindow.document;
             doc.open();
             doc.write(html);
             doc.close();
+          } catch (e) {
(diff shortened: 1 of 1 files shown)


panel/src/App.vue
```diff
@@ -122,10 +122,14 @@ export default {
...
-          doc.open();
-          doc.write(html);
-          doc.close();
+          try {
+            let doc = this.$refs.fatal.contentWindow.document;
+            doc.open();
+            doc.write(html);
+            doc.close();
+          } catch (e) {
+            console.error(e);
```
lukasbestle, line 131: Shouldn't the original error also be printed to the console in this case so that the issue can be debugged even if the iframe doesn't work? Or would that be too spammy?

bastianallgeier: I think it's too verbose when you send the entire HTML to the console. In a worst case scenario it's a full HTML document.

lukasbestle: 👍

lukasbestle (approved):

PR #3175, opened 2021-02-24
In-code UI component documentation

distantnative:
## Status: ready for review 👀

- Parses all JSDoc inline documentation from our components into `dist/ui.json` 
- JSON file can then can be used to generate virtual pages on getkirby.com
- `ui.json` is not included in git to not clutter the repo (needs to be separately transferred to website server)


panel/scripts/docs.js
```diff
@@ -0,0 +1,24 @@
...
+  const parsing = files.map(async file => {
+    let data = await docgen.parse(file, {
+      alias: { '@': path.resolve(__dirname, '../src/') },
+    });
+    data.srcFile = file;
+    return data;
+  });
+
+  const components = await Promise.all(parsing);
+  fs.writeFileSync("dist/components.json", JSON.stringify(components));
```
lukasbestle: I think this should use an absolute path based on `__dirname` as well to avoid strange bugs down the line when the file is updated in the wrong place.

distantnative:
@lukasbestle 
I think ideally, a GitHub action would run the npm script and then upload the .json file to our website server. That way we don't need to include it in any repo/download or so.

To be sure that it matches the installed Kirby version for getkirby.com, maybe we could think of another action as well that updates the getkirby.com repo automatically on release of a new Kirby version. And then run those two actions subsequently.

lukasbestle:
> That way we don't need to include it in any repo/download or so.

To be honest I think we should do exactly that. Otherwise the getkirby.com site cannot be used offline, which many users do.
If we want to keep the files out of the repo, the alternative would be to automatically download them lazily from getkirby.com if the current domain is not already getkirby.com.

bastianallgeier (approved):
The amount of work in this is incredible! I love how much it improves our JS code. Having docs in there now as well is just great and will force us to take much more care of good inline docs.

bastianallgeier:
@lukasbestle @distantnative I still think we should make sure to keep it out of the repo. Imagine how big this file will get when we really document every tiny bit like we do in PHP. 

What about a simple download/cache action that would still work when you use the repo offline?

distantnative:
I have no strong feeling about either way. So whichever you two prefer.

I think we can merge this already without a final decision, that decision is rather about how to include it in the website. The PR as is seems fine, adding the in-code comments, adding the generator script and excluding the file form this repo (I think at least that's fore sure, that it shouldn't be in his one).

lukasbestle:
I agree, downloading the file on the fly from getkirby.com and then caching it locally would be the cleanest solution. It would also be great for the planned reference version switcher that requires another data file for each Kirby version with all the backend data.

PR #3144, opened 2021-02-06
Added missing nodes prop for writer field #3125

afbora:
## Describe the PR
`nodes` prop missing from [the writer field document](https://getkirby.com/docs/reference/panel/fields/writer#field-properties) was added.

## Related issues

- Fixes #3125 

## Ready?

- [ ] Added unit tests for fixed bug/feature
- [x] Added in-code documentation (if needed)
- [x] CI passes (runs automatically when the PR is created or run `composer ci` locally)
  Running locally requires PHPUnit, PHP-CS-Fixer, Psalm, PHPCPD and PHPMD.

- [x] Checked whether the PR needs documentation, if needed added to the [release docs checklist](https://github.com/getkirby/getkirby.com/pulls)

PR #3182, opened 2021-02-25
Casts to string writer&list field value with always trimming

### LINKED ISSUE #3126: Writer- & List-Field:  Impossible to revert after focusing a new field
**Describe the bug**  
The saving dialogue immediately appears, when focusing a writer- or list field that hasn't been saved before.
It's impossible to revert.

**To Reproduce**  
1. Add a writer- or list field to an existing blueprint
2. Focus the field (the saving dialogue pops up)
3. Try delete all changes (not possible)

**Expected behavior**  
The saving dialogue should only appear, after a value has been entered
or reverting should be possible.

**Kirby Version**  
3.5.1

 - OS: 11.1
 - Browser: Chrome
 - Version: 88

 **Additional context**  
 It's not of a big deal, but consider you have already made changes to another field 
 that you don't want to keep.
  
- Probably related to [issue#3115](https://github.com/getkirby/kirby/issues/3115).
- The same behaviour seems to occur with the [list-field](https://github.com/getkirby/kirby/issues/3122#issuecomment-769901705)

afbora:
## Describe the PR
It always `trim` the value to string as we did in the `textarea` field. It also corrects the `null` data to allow revert changes. Because if the field did not exist before, its data could never be reverted because it was `null`.

## Related issues

- Fixes #3126 

## Ready?

### DIFF
--- a/config/fields/list.php
+++ b/config/fields/list.php
@@ -8,5 +8,10 @@ return [
         'marks' => function ($marks = true) {
             return $marks;
         }
+    ],
+    'computed' => [
+        'value' => function () {
+            return trim($this->value);
+        }
     ]
 ];
--- a/config/fields/writer.php
+++ b/config/fields/writer.php
@@ -16,5 +16,10 @@ return [
         'marks' => function ($marks = true) {
             return $marks;
         }
-    ]
+    ],
+    'computed' => [
+        'value' => function () {
+            return trim($this->value);
+        }
+    ],
 ];


bastianallgeier (commented):
I think we should really add and document all the props that those fields really have. It looks like I totally stuck with the MVP from the development phase here. I think the trimming could also be done in the props method for the value then.

afbora:
This is not a missing prop. It needed a method like `textarea` field that converts to string and clears spaces.

Can you explain a little more what to do? 🙈 Also we have https://github.com/getkirby/kirby/pull/3144

bastianallgeier:
@afbora forget about my comments :) This works great.

bastianallgeier (approved):

PR #3190, opened 2021-03-03
Migrate from Vue CLI to Vite

distantnative:
## Status:  ready for review 👀

- `npm run dev` in ~200 ms vs. ~9s (!)
- `npm run build` in ~8 s vs. ~10s

> dist/css/index.css   82.17kb 
dist/js/index.js     360.83kb
dist/js/vendor.js    415.68kb

vs.

> dist/js/chunk-vendors.a23f8c94.js    464.42 KiB 
  dist/js/app.js                       349.19 KiB   
  dist/css/app.css                     82.48 KiB   

- clean(er) migrate path to Vue 3

## Breaking changes
- Requires in production browsers with native ESM dynamic import support: https://caniuse.com/es6-module » we would need to raise our requirements to Firefox 60+, Chrome 61+, Opera 48+, macOS Safari 11+, Android Browser 81+ and Chrome for Android 88+. or use https://github.com/vitejs/vite/tree/main/packages/plugin-legacy
- Plugin assets aren't loaded when running `npm run dev` but have to be tested on an actual instance (we should only test plugins anyways with the dist files)

## Known problems
- [ ] Sandbox broken (raw loading not supported https://github.com/underfin/vite-plugin-vue2/issues/71 / https://github.com/underfin/vite-plugin-vue2/pull/72) - IMO this shouldn't block us from merging

## Ready?

- [x] CI checks pass

## When merging


bastianallgeier (commented):
I realised a weird issue with plugins. It seems like the plugin components no longer get registered in the Vite version.

distantnative:
@bastianallgeier will look into it - that's running `dev` or with built files?

bastianallgeier:
@distantnative dev

distantnative:
Ah ok, as I wrote in the initial post, `dev` doesn't load the plugin asset files anymore - Vite was acting up and when asking on Discord, Johann brought up that it might be a good idea to just remove them and always only test plugins with dist files.

bastianallgeier:
Ah, sorry. I completely missed that.

.github/workflows/ci.yml
```diff
@@ -311,8 +311,7 @@ jobs:
         if: always() && steps.finishPrepare.outcome == 'success'
         # Use the --no-fix flag in push builds to get a failed CI status
         run: >
-          npm run lint -- --max-warnings 0 --format stylish
-          ${{ github.event_name != 'pull_request' && '--no-fix' || '' }}
```
lukasbestle: Why has this line been removed? It's important for generating a failed CI status (see the comment above). Also the `stylish` format is needed for GitHub's automatic problem matchers.

lukasbestle (commented):
I love how the setup gets simpler by this!

Two general notes:

- I think we should carefully consider each case of suppressing `vue/no-mutating-props` as this Vue feature is deprecated: https://vuejs.org/v2/guide/migration.html#Prop-Mutation-deprecated
- The `vite.config.js` needs a way to inject custom config from another file like the old Vue-CLI config had. Otherwise my local setup will break as I need to define custom TLS certs for the embedded dev server. I haven't looked into how to configure this with Vite yet, but my previous `vue.config.custom.js` looked like this:

```js
const fs = require("fs");

process.env.VUE_APP_DEV_SERVER = "https://sandbox.kirby.test";

module.exports = {
  host: "sandbox.kirby.test",
  public: "https://sandbox.kirby.test:8080",
  https: {
    key: fs.readFileSync('/path/to/sandbox.kirby.test/sandbox.kirby.test.key'),
    cert: fs.readFileSync('/path/to/sandbox.kirby.test/sandbox.kirby.test.crt')
  }
};
```

.github/workflows/ci.yml
distantnative: It caused the script to fail for me and when I looked into it, I believe I found that `stylish` was deprecated? But this was some weeks ago, so I only remember vaguely. Will try to bring it back and see if it works now.

distantnative:
- `vue/no-mutating-props` indeed is a bad practice but one we introduced 2 years ago in some places. Only with the new setup, it would fail without ignoring it explicitly. I tried to fix it quickly, but wasn't able - this will require some deeper changes. Since it's nothing the PR introduced, I thought it's better to ignore the error messages for now to not clutter the PR with a whole other topic. But I have created an issue now for this: https://github.com/getkirby/kirby/issues/3299

- I'll try to make the vite config file configurable as well.

distantnative:
@lukasbestle Would this work for you?

```js
let custom;
try {
  custom = require("./vite.config.custom.js");
} catch (e) {
  custom = {};
}

const path = require('path');
const proxy = {
  target: process.env.VUE_APP_DEV_SERVER || "http://sandbox.test",
  changeOrigin: true,
  secure: false,
  ...custom
};
```

Available options are https://github.com/http-party/node-http-proxy#options

.github/workflows/ci.yml
distantnative: Works 👍

lukasbestle:
I just tested it and it turns out I needed to extend the `server` config. Works great now and I have pushed the necessary code change. 🎉 

In my initial tests the proxy performance of the Vite dev server seems to be a bit slower than Vue CLI, but with the much better build performance that is fine.

---

New question: I'm getting the following warning:

```
npm WARN deprecated babel-eslint@10.1.0: babel-eslint is now @babel/eslint-parser. This package will no longer receive updates.
```

Can we switch to the new [`@babel/eslint-parser`](https://www.npmjs.com/package/@babel/eslint-parser)?

distantnative:
@lukasbestle `@babel/eslint-parser` works for me, added it

distantnative:
@lukasbestle I added `@babel/eslint-parser` but getting a lot of warnings, additional plugins required...

distantnative:
@lukasbestle seems like we can live without `babel-eslint`?

lukasbestle:
Well, that's a hundred times better. 🎉

lukasbestle:
Regarding the open issues:

- Browser version change: After looking at the "usage relative" view on caniuse.com I see raising the requirements as a non-issue.
- Plugins in `dev` mode: Not optimal, but I'd say it's OK.

> Sandbox broken (raw loading not supported underfin/vite-plugin-vue2#71 / underfin/vite-plugin-vue2#72) - IMO this shouldn't block us from merging

What are you referring to with sandbox here?

If that's not critical, I think we can already merge this PR after another quick code review. 🎉

distantnative:
@lukasbestle Sandbox is a little tool Bastian built for building and testing Vue components. You can find it when running `dev` at /panel/sandbox.

I linked to the PR that I proposed to fix the vite plugin. Not sure though how actively maintained that one is. If we wanted we could use my fork instead.

Either way, it's only affecting a small thing of a dev tool that we don't use a lot currently. Do either way, merging should be fine. But would be good to hear from @bastianallgeier and @afbora if this new setup works for them as well since it's such a big change.

afbora:
@distantnative I guess, works perfect and pretty fast for me 🎉 

````sh
$ npm update
$ npm run build

> kirby@ build \GitHub\kirby\panel
> vite build && rm dist/index.html

vite v2.2.3 building for production...
transforming...
✓ 692 modules transformed.
rendering chunks...
dist/index.html      41.08kb
dist/css/index.css   82.17kb / brotli: 12.80kb
dist/js/index.js     360.83kb / brotli: 65.72kb
dist/js/vendor.js    416.36kb / brotli: 112.20kb

````

bastianallgeier:
Just a few first impressions after returning to this PR after what feels like years

- the dev performance is still unbelievable. This will be so much fun while working on new components. 
- the better build performance is nice too
- I don't mind the broken plugins and agree that we should only test them in dist scenarios
- I don't mind the new browser requirements either. That's what a 3.6 release is for. We only might need to test that when the panel is loaded and redirect to the browser support page and adjust that. 
- The reload speed is worrying me a bit. Whenever we need to work with dom events, you often need to make a full reload in order to get the events registered properly. This seems quite annoying with the longer load times now. But I take that as an acceptable trade-off in comparison to the benefits. 
- I think it makes a lot of sense to merge this as the first next step before moving on with other 3.6 tasks. This way we can detect additional issues with the new setup early.

bastianallgeier (approved):
Let's move ahead with this. 
We know that we want to go this way. The PR works great already. If we find issues we can fix them during the dev process for 3.6

PR #3194, opened 2021-03-06
Replace SCSS with vanilla CSS & vars

distantnative:
## Status: ready for review 👀

- Freeing us from SCSS compilation and rather using CSS vars - which are available for all developers as well.
- Eat our own dog food (same vars as plugin developers)
- The CSS could be optimized in places. (especially with the many nested `[dir=ltr]/[dir=rtl]` parts
- To not increase the load of this PR more though, I kept it to converting SCSS to CSS 1:1 (or as close as possible).

## Breaking changes 
- When developing we now need to use CSS vars instead of SCSS vars

## When merging


bastianallgeier (commented):
It blows up the CSS file size quite a bit, but I think it's worth it.

distantnative:
I think we need to look into ways to have the RTL/LTR parts cleaner now without sass. But I'd leave this for a separate PR

bastianallgeier (approved):

PR #3199, opened 2021-03-10
FileCache: Clean up empty directories on remove() #3162

### LINKED ISSUE #3162: FileCache: Clean up empty directories on remove()
When a cache item is removed, it may leave behind an empty directory.

The `FileCache::remove()` method should check if the directory the item was in is now empty. If it is, it should delete the empty directory. Then it should check the same for the next level up until the root is reached (which should *not* be deleted).

afbora:
## Describe the PR

When the cache is removed, it moves to the root directory, removing empty directories.

## Related issues

- Fixes #3162 

## Ready?

### DIFF
--- a/src/Cache/FileCache.php
+++ b/src/Cache/FileCache.php
@@ -189,6 +189,7 @@ class FileCache extends Cache
 
     /**
      * Removes empty directories safely by checking each directory
+     * up to the root directory
      *
      * @param string $dir
      * @return void
--- a/src/Cache/FileCache.php
+++ b/src/Cache/FileCache.php
@@ -188,7 +188,7 @@ class FileCache extends Cache
     }
 
     /**
-     * Removes empty directories safely by moving to the root directory.
+     * Removes empty directories safely by checking each directory
      *
      * @param string $dir
      * @return void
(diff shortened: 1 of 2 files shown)


src/Cache/FileCache.php
```diff
@@ -178,10 +179,34 @@ public function remove(string $key): bool
...
+        if (is_file($file) === true && F::remove($file) === true) {
+            $this->removeEmptyDirectories(dirname($file));
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
+     * Removes empty directories safely by moving to the root directory.
```
lukasbestle:
```suggestion
     * Removes empty directories safely by checking each directory
     * up to the root directory
```

src/Cache/FileCache.php
```diff
@@ -178,10 +179,34 @@ public function remove(string $key): bool
...
+     */
+    protected function removeEmptyDirectories(string $dir): void
+    {
+        try {
+            if (is_dir($dir) === true && $dir !== $this->root()) {
+                $files = array_diff(scandir($dir) ?? [], ['.', '..']);
+
+                if (empty($files) === true) {
+                    if (Dir::remove($dir) === true) {
+                        $this->removeEmptyDirectories(dirname($dir));
```
lukasbestle: The recursive approach will work quite well for shallow nesting levels (structures just a few levels deep). But if cache IDs contain potentially dozens of slashes (for whatever reason), this implementation will require too much memory for its purpose IMO.

We can fix this by using an iterative approach instead. Untested example code you could try:

```php
// ensure the path doesn't end with a slash for the next comparison
$dir = rtrim($dir, '/\/');

while (Str::startsWith($dir, $this->root()) === true && $dir !== $this->root()) {
    $files = array_diff(scandir($dir) ?? [], ['.', '..']);

    if (empty($files) === true && Dir::remove($dir) === true) {
        // continue with the next level up
        $dir = dirname($dir);
    }

    // no need to continue with the next level up as $dir was not deleted
    return;
}
```

afbora: It was a 99% working code. Copy & paste & test 😅 Thank you 👍

lukasbestle (approved):
Awesome! 👍

PR #3202, opened 2021-03-11
Render the error page on invalid media URLs

lukasbestle:
## Describe the PR

More consistent behavior that also allows overriding the error behavior with custom code on the error page.

## Related issues

- Fixes https://forum.getkirby.com/t/adjusting-the-media-hash-generation/21456/14?u=lukasbestle.

## Ready?

### DIFF
--- a/src/Cms/Media.php
+++ b/src/Cms/Media.php
@@ -48,8 +48,8 @@ class Media
                 if (Str::startsWith($hash, $file->mediaToken() . '-') === true) {
                     return Response::redirect($file->mediaUrl(), 307);
                 } else {
-                    // don't leak the correct token
-                    return new Response('Not Found', 'text/plain', 404);
+                    // don't leak the correct token, render the error page
+                    return false;
                 }
             }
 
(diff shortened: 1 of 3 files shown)


distantnative:
Wondering whether this could be a breaking change for some setups - (for whatever reason) checking for text/plain on error responses but now getting HTML.

lukasbestle:
I don‘t think so. We previously had `HTTP 404` already, so that would have been much more reliable to check for. But to be honest, I don‘t see many use-cases where you would need to check the response anyway (as the media URLs are mainly used for linking to files). Do you have an example use-case?

distantnative:
Having looked at it again, you're right: I think it's a good change that won't really cause issues anywhere.

Questions is if this is urgent or could wait until 3.6.0 since strictly it's a breaking change (public method returns something different now) - one that won't break many (if any) actual setups though.

lukasbestle:
You are right, it's probably better to move this to 3.6.0.

Marking it as draft to avoid that we merge it by accident. The code is already done.

distantnative (approved):

PR #3207, opened 2021-03-14
Use `Wrong password` message on login instead `notSame`

### LINKED ISSUE #3165: Inadequate error message "Passwords do no match" when logging in with wrong password in debug mode
**Describe the bug**  
Since 3.5, the error messages for user login have changed when in debug mode.

But the error message for a wrong password now is the same as for `v::same`, i.e. `error.user.password.notSame` while it should rather be along the lines of "Incorrect or wrong password".

**To Reproduce**  
1. Enable debug mode
2. Log in to Panel with wrong password
3. See error message "Passwords do not match"

**Expected behavior**  
Error message should be "Wrong password" or similar.

**Kirby Version**  
3.5.x

afbora:
## Describe the PR

Use `Wrong password` message on login instead `The passwords do not match`.

screen-capture-848-Login - Mægazine-localhost

## Related issues

- Fixes #3165 

## Ready?

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -163,6 +163,7 @@
   "error.user.password.invalid": "Please enter a valid password. Passwords must be at least 8 characters long.",
   "error.user.password.notSame": "The passwords do not match",
   "error.user.password.undefined": "The user does not have a password",
+  "error.user.password.wrong": "Wrong password",
   "error.user.role.invalid": "Please enter a valid role",
   "error.user.update.permission": "You are not allowed to update the user \"{name}\"",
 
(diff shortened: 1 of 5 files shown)


nilshoerrmann:
Question: Is it a good idea to display this message publicly? In case of an attack, this will confirm that a user with the given address exists. So shouldn‘t this message be more arbitrary?

texnixe:
@nilshoerrmann This is only in debug mode. If debugging is off, you get a general message like before (invalid login data or something like that)

texnixe (approved):

nilshoerrmann:
Ah, sorry, I‘ve missed that 👍

lukasbestle (approved):

bastianallgeier (approved):

PR #3215, opened 2021-03-18
Checks slug length on page duplicate

### LINKED ISSUE #3214: Duplicate a page with empty URL appendix returns 500 error
**Describe the bug**  
When I duplicate a page and URL appendix is empty then 500 error is returned.

**To Reproduce**  
1. Go to the Panel
2. Click duplicate a page
3. Empty the URL appendix
5. This is when duplicate a top level page
    image
6. And this is when try duplicate a sub-page
    image

**Expected behavior**  
A proper error message that says that field is required.

**Kirby Version**  
3.5.3.1

**Additional context**  
The 500 error also happens on trykirby.com but showing different message.

image

afbora:
## Describe the PR

Fixed by checking slug in `PageRules::duplicate()` method.

## Related issues

- Fixes #3214 

## Ready?

### DIFF
--- a/src/Cms/PageRules.php
+++ b/src/Cms/PageRules.php
@@ -292,6 +292,8 @@ class PageRules
             ]);
         }
 
+        self::validateSlugLength($page->slug());
+
         if ($page->exists() === true) {
             throw new DuplicateException([
                 'key'  => 'page.draft.duplicate',
@@ -300,8 +302,6 @@ class PageRules
                 ]
             ]);
         }
-
-        self::validateSlugLength($page->slug());
 
         $siblings = $page->parentModel()->children();
         $drafts   = $page->parentModel()->drafts();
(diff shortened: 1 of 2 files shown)


lukasbestle (dismissed):
We have the same issue with `changeSlug()`. There we only catch this case in the frontend at the moment, the API happily accepts empty slugs.

I think we should move the minimum length check to the `validateSlugLength()` method to keep the code DRY. Then we can also remove the check from the `create()` method. Another advantage: The `validateSlugLength()` method then only needs to run `Str::length()` once and can use the return value for both checks (min and max length).

afbora:
@lukasbestle I have a question; should the `permissions` be checked first or the `slug length` in a rule? I think that permissions should be checked first, but different uses have been made in the class.

lukasbestle:
@afbora I agree, permissions should always be checked first.

bastianallgeier (approved):
I've switched the order in the create rule to check for the length first before checking for duplicates. Otherwise we might make a duplicate check for a slug that's too long or too short.

PR #3223, opened 2021-03-28
Fix and extend doc blocks and parameter hinting

distantnative:


src/Cms/Site.php
```diff
@@ -360,11 +360,15 @@ public function mediaUrl(): string
      *
      * @param string|null $format
      * @param string|null $handler
-     * @return mixed
+     * @return int|string
```
afbora, line 363: `Dir::modified()` returns `int`?

https://github.com/getkirby/kirby/blob/3.5.3.1/src/Toolkit/Dir.php#L257

lukasbestle: That type hint is incorrect as well. If a format is used, the return value is indeed `string`.

lukasbestle (commented):
I agree on all changes. However I think we should wait for 3.6.0 before we merge this as some type hints make the behavior stricter.

src/Cms/Site.php
lukasbestle: That type hint is incorrect as well. If a format is used, the return value is indeed `string`.

distantnative:
@lukasbestle I will move all changes that make the behavior stricter to a separate PR. That way we can get the other doc fixes faster into 3.5.4 already.

lukasbestle (approved):

Issue #3177, opened 2021-02-24
Panel: use mixins instead of extending other components' props


distantnative:
Cannot be parsed by JSDoc and cannot be used by plugin developers (mixing can!)

bastianallgeier:
✅

PR #3224, opened 2021-03-28
Fields: use mixins instead of destructing props

distantnative:
## Status: ready for review 👀
- Creates `mixins/props` for common props to DRY code (especially with in-line docs)
- Some core elements are already importing such props definition mixins
- Components that have been previously imported by other components to destruct their `props` are now exporting a mixin including the `props` object
- Instead of destructing, the other components now import such `props` object and use it as mixin
- `vuelidate` imports have been renamed to not clash with prop names

## Why?
- Object destructors are not parseable for automated docs, mixins are
- Will provide a first step to make these mixins available to plugin developers as well ("eat your own dog food")

## Future plans
- Import common exported props (from input/fields) into mixins (e.g. `mixins/forms/TextInput,js`) that can easily be exposed to plugins

## Related issues
- Closes https://github.com/getkirby/kirby/issues/3177
- 
## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/panel/src/mixins/forms/counter.js
+++ b/panel/src/mixins/forms/counter.js
@@ -12,8 +12,17 @@ export default {
         return false;
       }
 
+      let count = 0;
+
+      if (this.value) {
+        if (Array.isArray(this.value)) {
+          count = this.value.length;
+        } else {
+          count = String(this.value).length;
+        }
+      }
       return {
-        count: this.value && Array.isArray(this.value) ? this.value.length : 0,
+        count: count,
         min: this.min,
         max: this.max,
       };
--- a/panel/src/mixins/forms/counter.js
+++ b/panel/src/mixins/forms/counter.js
@@ -1,0 +1,22 @@
+
+export default {
+  props: {
+    counter: {
+      type: Boolean,
+      default: true
+    }
+  },
+  computed: {
+    counterOptions() {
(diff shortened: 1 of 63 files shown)


distantnative:
@afbora If you ever are looking for something to help on, just testing if fields and inputs still proper function with these changes would be a great help! 🙏

panel/src/mixins/forms/counter.js
```diff
@@ -0,0 +1,22 @@
...
+    }
+  },
+  computed: {
+    counterOptions() {
+      if (this.value === null || this.disabled || this.counter === false) {
+        return false;
+      }
+
+      return {
+        count: this.value && Array.isArray(this.value) ? this.value.length : 0,
```
afbora: Value is `string` for `password`, `text` and `textarea` fields. 
I think there should be `string` or `array` control here. 
Am I wrong?

distantnative: Good catch! Added a fix to check for array/string.

afbora:
When I tested it the other day, there was no issue other than `counter.js` (you've fixed). I forgot to write.

afbora (approved):
LGTM 👍

bastianallgeier:
A nice side effect seems to be that the build process is able to compress it more than the objects before. 🔥

PR #3225, opened 2021-03-28
Panel: refactor component loading

distantnative:
## Status: on hold ⏸

⚠️ needs first https://github.com/getkirby/kirby/pull/3224

- Core components as well as plugin components now get loaded and registered via the same script
- Goodbye long old list of manually registering core components
- Implements register queue that takes several rounds if a parent component isn"t found (when extending) to make sure it is actually missing and not just provided by another plugin further down the line

## Why?
- more streamlined registration - both core and plugin components are treated in the same way
- eat your own dog food: extend via string

## TODO
- [ ] Revert extending via name string: won't be parseable JSDocs for UI Kit docs


distantnative:
Closing this to reduce clutter from the PR view. Will keep working on it locally and reopen a PR once it's ready for the rest of the team to join in.

PR #3232, opened 2021-04-07
More specific type hints

distantnative:


bastianallgeier (approved):

lukasbestle:
@bastianallgeier This was for 3.6.0 and shouldn't be in 3.5.4. I think we should revert it.

PR #3237, opened 2021-04-09
Whoops: Allow Exceptions to send custom HTTP code

lukasbestle:
## Describe the PR

Whoops overrides any sent HTTP code with its own `HTTP 500` unless disabled. This meant that any JSON error response would have `HTTP 500` even if the `Exception` has defined a different code.

It's still the same with the HTML output, but here it's more difficult to set a custom code as we don't have dynamic access to the handled `Exception`.

## Ready?

- [x] ~Added unit tests for fixed bug/feature~ Automated Whoops tests are not possible

### DIFF
--- a/src/Cms/AppErrors.php
+++ b/src/Cms/AppErrors.php
@@ -144,6 +144,7 @@ trait AppErrors
         });
 
         $this->setWhoopsHandler($handler);
+        $this->whoops()->sendHttpCode(false);
     }
 
     /**


bastianallgeier (commented):
Will our error response code be even sent before this?

bastianallgeier (approved):
Sorry, that was a stupid question. Our Response class is taking care of it in the handler. Should have checked out the entire code of the file.

PR #3238, opened 2021-04-11
New `Panel` namespace

distantnative:
## Status: ready fro review 👀

- Pulls all Panel-related methods out of main model classes in their own classes in new `Kirby\Panel` namespace
  - `::dragText()`
  - `::panelIcon()`
  - `::panelImage()`
  - `::panelImageSource()`
  - `::panelPath()`
  - `::panelPickerData()`
  - `::panelOptions()`
  - `::panelUrl()`
- Adds a new `::panel()` method to all models that returns object with access to the separated methods, e.g. `$page->panel()->image()` for `$page->panelImage()`

## Breaking changes 
- Old methods are marked as deprecated in doc block. 
- We should add `deprecated()` helper warnings in 3.7.0 
- And remove the methods earliest in 3.8.0.

## Ready?

- [x] ~~Checked whether the PR needs documentation, if needed added to the [release docs checklist](https://github.com/getkirby/getkirby.com/pulls)~~

⚠️ Codecov thresholds fail but actual test coverage stayed the same

## When merging


distantnative:
@lukasbestle any way we can tell codecov that this is actually fine and allow merging?

lukasbestle:
@distantnative No, I don't think so. Maybe we can override the checks once this PR is approved, but probably we disabled that as well in the branch protection config.

Is this the PR you were talking about with the untested code that was simply moved? If so, I think the most solid solution would be to add tests for the methods that don't already have them. We will need them anyway in the long run.

distantnative:
It is. But the code that's been moved is even tested. Just since it's moved out of the classes, the percentage of tested code for that file decreases and trips codecov up.

I wouldn't add tests in this PR for other yet untested methods in the same class. We shouldn't start adding unrelated things just to please that algorithm. I think, if codecov can't be smart enough to cover this itself, we need to make the checks only a notice, not blocking merging.

lukasbestle:
There are several uncovered lines in the diff (the ones where the line number has a dark red background): https://app.codecov.io/gh/getkirby/kirby/compare/3238/diff
You can find the affected files with the "relative" number in the Codecov delta (the percentage in the middle). If that's < 100%, then not all changed lines are covered.

As far as I understand it, the "patch" check that currently says "93.91 %" only checks if all lines that have been added in the diff are covered, so in this case 93.91 % of all lines you touched are covered. I don't think the other methods in the same file matter, they are not part of that percentage calculation.

lukasbestle:
https://docs.codecov.io/docs/codecov-delta

distantnative:
@lukasbestle I added more tests etc. but it's still failing
https://app.codecov.io/gh/getkirby/kirby/compare/3238/diff

Some examples:
- for src/Cms/File it shows -3.34%: if you look in the details, a lot of code simply has been removed from the class, all code that has been touched and remains is covered. It shows a dark red background on a line that's gone, doesn't make sense
- src/Cms/Site it shows -0.23%: the line it complains about (dark red background) hasn't been touched at all by this PR
- config/fields/files.php is at 50% while config/fields/pages.php at 100% - the same change happened there

This is turning into an incredibly frustrating developing experience. To be honest, with this current setup, I think we are doing us a disservice where it only demotivates development but doesn't actually lead to better test coverage. I understand that there should be some checks but this just seems to create extra work for no benefit, just barriers.

lukasbestle:
@distantnative The diff takes some getting used to. Let me explain:

The line numbers are displayed in two different columns. The left column displays the coverage colors before the PR (which is why line 507 of `src/Cms/File.php` is marked as red in the left column, as it wasn't covered before). The right column displays the coverage colors of the code after the changes from the PR. In the `src/Cms/File.php` example you can see that the right line number column is missing, which makes sense considering the code was removed. These lines can no longer be covered as they don't exist.

If you take a look at the Codecov Delta of `src/Cms/File.php`, it says `89.52% < 100.00% > (-3.34%)`. This means that the file is now 89.52 % covered overall with 100% of the changed code being covered. The third number means that the PR reduced coverage of this file by 3.34 %. There are more examples on the Codecov Delta in the docs I linked to above. The short summary for our purposes: The last number doesn't matter at all for our checks, it's purely informative. What matters is the second number. It's 100 % so all required tests for this file are present. Nothing more to do here.

The same for `src/Cms/Site.php`: The second number is 100%, you can ignore the third percentage.

Regarding `config/fields/files.php`: You can see that the second changed line was uncovered before (red in the left column) and is still uncovered (red in the right column). As one of the two changed lines in this file is not covered, the percentage is 50%. So there needs to be a test that covers that line 120.

distantnative:
Could we please set it up in a way that lines that were uncovered before are not required to be covered after? New lines yes, removing coverage from previously covered lines yes, but as this example really highlights (we are just changing how the method is called) adding tests for those whole fields would take this PR into whole different territory. That's what I meant with that the pure algorithm shouldn't dictate and force us into directions that don't make sense for our development.

bastianallgeier (approved):
The PR looks good to me. I'm still a bit lost about the Codecov issue. Testing that newly detected part in the files.php is super hard and well outside of the scope of this PR.

distantnative:
Deactivated the branch protection rules to merge this manually. Reactivated them.

PR #3239, opened 2021-04-11
`Kirby\File` package

distantnative:
## What it does
New setup for handling file assets in Kirby:

- `Toolkit\File` takes care of everything CMS-unrelated. Gains a few more methods (e.g. `download`, `headers`) that couple it a bit more with other packages (Http)
- New `Kirby\File` package
- `Kirby\File\File` extends `Kirby\Toolkit\File` with some Cms-related functions (e.g. `modified` with handler, `isResizable`)
-  `Kirby\File\Image` extends `Kirby\File\File` in turn and adds image-related methods (e.g. `dimensions`, `exif`)
- Both replace `Kirby\Image\Image`
- The trait `Kirby\File\hasFile` adds method-proving to a class - either to `Kirby\File\Image` or `Kirby\File\File` depending on the file type of the `$root` property
- `Kirby\Cms\FileFoundation` is removed - it gets replaced by `Kirby\File\hasFile` and all proxied methods of `Kirby\File\File`/`Kirby\File\Image` which reach down to  `Kirby\Toolkit\File` as well
- `Kirby\Cms\Asset` » `Kirby\File\Asset`
- `Kirby\Cms\Filename` » `Kirby\File\Name`
- There are aliases for mapping old class names with new ones

## Breaking changes
- `Kirby\File\Image` returns an `` path when echoed as string. `Kirby\Image\Image` used to echo the file path.

## Todo
- Clean up old tests (already passing, but might create cruft)


distantnative:
@lukasbestle `psalm` still tries to load `kirby\cms\asset` -  is that due to caching of Github Actions?

lukasbestle:
Have you tried running Psalm locally? I don't think this kind of information is cached.

I suspect this could be because of the double alias in line 6 of `config/aliases.php`.

src/Cms/FileVersion.php
```diff
@@ -132,7 +133,7 @@ protected function setOriginal($original)
      */
     public function toArray(): array
     {
-        $array = array_merge($this->parentToArray(), [
+        $array = array_merge($this->asset()->ToArray(), [
```
nilshoerrmann: Are you sure this shouldn't be `$this->asset()->toArray()` (with a small `t`)?

distantnative: Thanks for the catch

distantnative:
@lukasbestle Locally psalm is once again acting up about APCU and Memcached 🙄

PR #3241, opened 2021-04-12
More specific type hints

lukasbestle:
## Status: ready for review 👀

- Add type hints/stricter type hints for 
  - `Page::controller()`
  - `Site::setUrl()`
  - `Site::url()`

## Breaking changes
_Needs to be reviewed_

## Original PR
#3232


bastianallgeier (approved):
I don't see any reason why this should cause a breaking change.

PR #3246, opened 2021-04-16
Allow setting the page cache expiry from controllers

lukasbestle:
## Describe the PR

- It is now possible to set the page cache expiry timestamp from controllers or templates:

```php
$kirby->response()->expires(1234567890); // timestamp
$kirby->response()->expires('2021-12-31');
```

- When setting cache values, you can now define an absolute expiry timestamp instead of the number of minutes. The number of minutes is still supported and auto-detected.

## Ready?

### DIFF
--- a/src/Cache/Value.php
+++ b/src/Cache/Value.php
@@ -25,6 +25,8 @@ class Value
 
     /**
      * the number of minutes until the value expires
+     * @todo Rename this property to $expiry to reflect
+     *       both minutes and absolute timestamps
      * @var int
      */
     protected $minutes;
@@ -40,7 +42,8 @@ class Value
      *
      * @param mixed $value
      * @param int $minutes the number of minutes until the value expires
-     * @param int $created the unix timestamp when the value has been created
+     *                     or an absolute UNIX timestamp
+     * @param int $created the UNIX timestamp when the value has been created
      */
     public function __construct($value, int $minutes = 0, int $created = null)
     {
@@ -70,6 +73,11 @@ class Value
         // 0 = keep forever
         if ($this->minutes === 0) {
             return null;
+        }
+
+        if ($this->minutes > 1000000000) {
+            // absolute timestamp
+            return $this->minutes;
         }
 
         return $this->created + ($this->minutes * 60);
--- a/src/Cms/Responder.php
+++ b/src/Cms/Responder.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Cms;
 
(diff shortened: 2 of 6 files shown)


afbora (dismissed):
Looks great like the https://github.com/getkirby/getkirby.com/pull/1304 PR 👍

src/Cache/Value.php
```diff
@@ -25,6 +25,8 @@ class Value
 
     /**
      * the number of minutes until the value expires
+     * @todo Rename this property to $expiry to reflect
```
distantnative, line 28: Can't we do this directly? Since it's a protected property

src/Cms/Responder.php
```diff
@@ -70,6 +79,50 @@ public function body(string $body = null)
...
 
+    /**
+     * Setter and getter for the cache expiry
+     * timestamp for Kirby's cache
+     *
+     * @param int|string|null $cacheExpiry Timestamp or time string to parse
+     * @param bool $override If `true`, the already defined timestamp will be overridden
+     * @return int|null|self
+     */
+    public function cacheExpiry($cacheExpiry = null, bool $override = false)
```
distantnative: Should we call the method and prop rather just `expires`? I feel that would be more in style with many other Kirby methods.

distantnative:
Two smaller comments. Otherwise looks good, it's a cool feature that makes caching more flexible.

src/Cache/Value.php
lukasbestle: For consistency it would also require renaming the `$minutes` argument in the constructor, which would be a breaking-change in PHP 8 (named arguments). That's why I left it out for now.

src/Cms/Responder.php
lukasbestle: All other methods in this class are nouns, so `expires` wouldn't be in line with them as the only verb. I'm split on this though, I think we need more opinions.

bastianallgeier: I'd vote for `$kirby->response()->expires()` as well. To me it feels closer to a command like `$kirby->response()->redirect()`

bastianallgeier:
I really like this! Makes caching so much more flexible.

bastianallgeier:
@lukasbestle I was about to change the method name, but for some reason there's a local issue in my tests with timestamps.

lukasbestle:
I've changed the method name and also added the feature to pass the number of minutes to the `expires()` method for consistency with the `Cache` classes.

bastianallgeier (approved):

PR #3248, opened 2021-04-17
Consistent `::add()` for model collections

distantnative:
## Status: ready for review 👀

- Adds exception message to `Files::add()` and `Users::add()` to be consistent with `Pages::add()`
- Fixes exception message of `Pages::add()`
- More precise parameter type in DocBlock for all `add()` methods

## Breaking changes

- Before this PR, one could pass something else than `Files` object, `File` object, `null`, `false`, `true` to the `Files::add()` method and it would simply be ignored (and the original collection returned). Now, an error is thrown.
- Same for `Users::add()`

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Cms/Files.php
+++ b/src/Cms/Files.php
@@ -2,6 +2,8 @@
 
 namespace Kirby\Cms;
 
+use Kirby\Exception\InvalidArgumentException;
+
 /**
  * The `$files` object extends the general
  * `Collection` class and refers to a
@@ -30,12 +32,13 @@ class Files extends Collection
      * an entire second collection to the
      * current collection
      *
-     * @param mixed $object
+     * @param \Kirby\Cms\Files|\Kirby\Cms\File|string $object
      * @return $this
+     * @throws \Kirby\Exception\InvalidArgumentException
      */
     public function add($object)
     {
-        // add a page collection
+        // add a files collection
         if (is_a($object, self::class) === true) {
             $this->data = array_merge($this->data, $object->data);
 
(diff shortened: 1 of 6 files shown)


lukasbestle (commented):
The changes look good. I wonder why `null`, `false` and `true` are silently ignored (already in the `Pages` class), but I guess it has historic reasons.

However in any case we can't include the exceptions in 3.5.4 as they make the behavior stricter. I think we should move the changes to 3.6.0 and only include the docblock changes in 3.5.4.

distantnative:
I vaguely recall that we tried it without ignoring them but this broke a lot of setups, where people would pass the result of other methods to `::add()` which could be `null`.

And good catch, I think it's fine if we move all to 3.6 then.

lukasbestle (dismissed):

src/Cms/Files.php
```diff
@@ -46,6 +49,10 @@ public function add($object)
         // add a file object
         } elseif (is_a($object, 'Kirby\Cms\File') === true) {
             $this->__set($object->id(), $object);
+
+        // give a useful error message on invalid input
```
lukasbestle:
```suggestion
        // give a useful error message on invalid input;
        // silently ignore "empty" values for compatibility with existing setups
```

src/Cms/Pages.php
```diff
@@ -68,7 +68,7 @@ public function add($object)
 
         // give a useful error message on invalid input
```
lukasbestle:
```suggestion
        // give a useful error message on invalid input;
        // silently ignore "empty" values for compatibility with existing setups
```

src/Cms/Users.php
```diff
@@ -53,6 +55,10 @@ public function add($object)
         // add a user object
         } elseif (is_a($object, 'Kirby\Cms\User') === true) {
             $this->__set($object->id(), $object);
+
+        // give a useful error message on invalid input
```
lukasbestle:
```suggestion
        // give a useful error message on invalid input;
        // silently ignore "empty" values for compatibility with existing setups
```

lukasbestle (commented):
Besides the comment changes I think this is great and ready to merge! 👍

lukasbestle (approved):

PR #3250, opened 2021-04-17
Use `Properties` for `Http\Request`

distantnative:
## Status: on hold ⏸

Will help us with the getkirby.com reference.
Wait until `Properties` has been refactored.

### DIFF
--- a/src/Http/Request.php
+++ b/src/Http/Request.php
@@ -2,13 +2,12 @@
 
 namespace Kirby\Http;
 
-use Kirby\Http\Request\Auth\BasicAuth;
-use Kirby\Http\Request\Auth\BearerAuth;
 use Kirby\Http\Request\Body;
 use Kirby\Http\Request\Files;
 use Kirby\Http\Request\Method;
 use Kirby\Http\Request\Query;
 use Kirby\Toolkit\A;
+use Kirby\Toolkit\Properties;
 use Kirby\Toolkit\Str;
 
 /**
@@ -24,10 +23,12 @@ use Kirby\Toolkit\Str;
  */
 class Request
 {
+    use Properties;
+
     /**
      * The auth object if available
      *
-     * @var BearerAuth|BasicAuth|false|null
+     * @var \Kirby\Http\Request\Auth\BasicAuth|\Kirby\Http\Request\Auth\BearerAuth|false
      */
     protected $auth;
 
@@ -41,7 +42,7 @@ class Request
      *
      * `$request->body()->get('foo')`
      *
-     * @var Body
+     * @var \Kirby\Http\Request\Body
      */
     protected $body;
 
@@ -56,7 +57,7 @@ class Request
      * `$request->files()->get('upload')['size']`
      * `$request->file('upload')['size']`
      *
-     * @var Files
+     * @var \Kirby\Http\Request\Files
      */
     protected $files;
 
@@ -68,14 +69,6 @@ class Request
     protected $method;
 
     /**
-     * All options that have been passed to
-     * the request in the constructor
-     *
-     * @var array
-     */
-    protected $options;
-
-    /**
      * The Query object is a wrapper around
      * the URL query string, which parses the
      * string and provides a clean API to fetch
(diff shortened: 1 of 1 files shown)


afbora:
Asking with not sure: Shouldn't it be processed when `body`, `files` and `query` options are `null`? 🤔

distantnative:
If I remove the `if` statements in the `set` methods, a lot of our unit tests start failing

src/Http/Request.php
```diff
@@ -383,6 +361,69 @@ public function query()
...
+
+    /**
+     * Sets the method
+     *
+     * @param string|null $method
+     * @return void
+     */
+    protected function setMethod(?string $method = null)
+    {
+        $this->method  = $this->detectRequestMethod($method);
```
lukasbestle:
```suggestion
        $this->method = $this->detectRequestMethod($method);
```

lukasbestle (commented):
It's correct to only initialize the child objects if the input is not `null`. That was the previous behavior as well (`isset()` returns `false` for `null` values).

Some general comments:

- The `auth()` method still tries to access the `$options` property. There should be a setter `setAuth` as well.
- The setters should always handle the `null` case by setting the property to null. Could work like this: `$this->body = $body ? new Body($body) : null;`
- In the other classes that use the `Properties` trait, the setters return `$this`. We should do the same here for consistency.

distantnative:
I was so sure I had done the `auth` part. And then I didn't.

- Moved most of the auth logic to a new `::setAuth`
- "The setters should always handle the null case by setting the property to null. " I actually realized that in the old code the getters would set the properties if they weren't set yet with empty objects. I removed that now, I think all of this should happen in the setters now
- Returning `$this` now
- Running into unit test errors still. Not sure yet whether valid errors or issues caused by objects not properly resetting between tests.

bastianallgeier:
To be honest, I'm not really sure about this at all. It would help the docs, but the Properties class is a concept that I thought is clever in the past, but I would rather get rid of in the future. It's built around Reflections and they are a performance bottleneck. I think Dodov made a few tests by replacing the properties in the Page class with manually assigning object props and the performance benefit was quite significant.

distantnative:
@bastianallgeier we should make a quick decision here then cause a lot of work on the website has gone in the direction of using Properties 🙈 and thus of course I wanted to get the remaining pieces on board as well... but now sounds like we should ditch that whole thing then. 

Not having these $props, $options .. parameters documented properly is one of the biggest weaknesses of our docs, I think (in terms of a weakness that a lot of developers keep running into). We would need another solution then before switching to something that destroys all these efforts but doesn't offer an equivalent solution.

distantnative:
Sorry. I think you're totally right that we should rather ditch `Properties` then. Tbh, I think we should make it a major effort for 3.6.0 then to remove it from wherever possible. Feels weird to have different philosophies in the classes.

I just wished I had known the info on performance some weeks ago before building a big feature of the docs on this.

bastianallgeier:
I'm really sorry that a lot of hard work went into this and I come up with this now. TBH I did just trust you with the auto-docs for the props and didn't even bother to think about the implications while working on the rest of the site. I don't think all of your work is lost at all. Moving away from Properties is going to be a LOT of work and it's not going to happen in just one release or in a matter of weeks or months. This will take a while and it's good to have a solid doc solution in the meantime.

distantnative:
True. Thanks for putting it in perspective :D

distantnative:
But would refactoring classes to move away from `Properties` be a good tasks for those moments where creativity is lacking (on me) but work could still be done?

Of course, we'd have to see that it's done without breaking changes wherever possible

bastianallgeier:
Yes, absolutely! It's just going to be quite tricky to not miss any of them and find a good alternative to keep the constructors clean. But I'd be super happy if you find a proper solution.

lukasbestle:
I've reopened this PR based on our discussion to consider an optimized `Properties` solution.

distantnative:
@lukasbestle would you pursue with these changes ahead of the `Properties` refactoring or only afterwards?

lukasbestle:
@distantnative I don't see a problem with implementing this already as the plan for the Properties refactoring is to keep backwards-compatibility in any case.

distantnative:
Closing this to remove clutter in the PR list. I think focusing on Inertia and then at some point on the `Properties`refactoring is more important. Will get back to this idea once the other things are in place.

PR #3268, opened 2021-04-26
F::relativepath() when an unrelated base is given, prefix with ../ for relative travel path

### LINKED ISSUE #3267: API debug exception error only returns filename (not file path) when DOCUMENT_ROOT differs from site folder
**Describe the bug**  
In situations where an installation's `$_SERVER['DOCUMENT_ROOT']` is set to a secluded folder that is separate from the Kirby installation (and `site` folder), API exceptions do not return a helpful file path.

**To Reproduce**  
1. Set document root to something like `/var/www/html/website/public`
2. Have a plugin in a folder like `/var/www/html/website/site/plugins/example-plugin/index.php`
3. Enable debug mode.
4. Do something in a hook that would throw an exception.

**Expected behavior**  
The JSON API error would say a helpful file path where the error occurred. Instead, you only get the filename.

_Actual_
```json
{
    "status": "error",
    "message": "Kirby\\Cms\\Site::TheStreamable\\ExamplePlugin\\{closure}(): Argument #1 ($foo) must be of type Object, string given",
    "code": 500,
    "exception": "TypeError",
    "key": null,
    "file": "index.php",
    "line": 122,
    "details": [],
    "route": "pages/([a-zA-Z0-9\\.\\-_%= \\+\\@\\(\\)]+)"
}
```

_Expected_
```json
{
    "status": "error",
    "message": "Kirby\\Cms\\Site::TheStreamable\\ExamplePlugin\\{closure}(): Argument #1 ($foo) must be of type Object, string given",
    "code": 500,
    "exception": "TypeError",
    "key": null,
    "file": "/var/www/html/website/site/plugins/example-plugin/index.php",
    "line": 122,
    "details": [],
    "route": "pages/([a-zA-Z0-9\\.\\-_%= \\+\\@\\(\\)]+)"
}
```

(The expected file path could be truly relative, with something like `../site/plugins/example-plugin/index.php`, but I don't really care much either way.)

**Kirby Version**  
`3.5.3`

**Additional context**  
Relavant Kirby line:
https://github.com/getkirby/kirby/blob/38afb2d650f3764d23d732b23ae8f04ab9ffba69/src/Api/Api.php#L689

It seems that `F:: relativepath()` handles the scenario that `$file` is not a descendant of `$in` by just returning the filename, which seems like an odd choice to me.

https://github.com/getkirby/kirby/blob/38afb2d650f3764d23d732b23ae8f04ab9ffba69/src/Toolkit/F.php#L646-L661

neildaniels:
## Describe the PR
When an unrelated base path is given to `F::relativepath()`, it would simply return the filename. This changes causes it to return the full path.

A more sophisticated solution could prepend relative tokens like `../../`, but the change as initially drafted still seems more helpful than existing behavior.

## Related issues
- Fixes #3267

## Ready?

### DIFF
--- a/src/Toolkit/F.php
+++ b/src/Toolkit/F.php
@@ -657,8 +657,8 @@ class F
         if (Str::contains($file, $in) === false) {
             // make the paths relative by stripping what they have
             // in common and adding `../` tokens at the start
-            $fileParts = explode('/', $file);
-            $inParts = explode('/', $in);
+            $fileParts = explode('/', rtrim($file, '/'));
+            $inParts = explode('/',  rtrim($in, '/'));
             while (count($fileParts) && count($inParts) && ($fileParts[0] === $inParts[0])) {
                 array_shift($fileParts);
                 array_shift($inParts);
--- a/src/Toolkit/F.php
+++ b/src/Toolkit/F.php
@@ -658,7 +658,7 @@ class F
             // make the paths relative by stripping what they have
             // in common and adding `../` tokens at the start
             $fileParts = explode('/', rtrim($file, '/'));
-            $inParts = explode('/',  rtrim($in, '/'));
+            $inParts = explode('/', rtrim($in, '/'));
             while (count($fileParts) && count($inParts) && ($fileParts[0] === $inParts[0])) {
                 array_shift($fileParts);
                 array_shift($inParts);
--- a/src/Toolkit/F.php
+++ b/src/Toolkit/F.php
@@ -653,7 +653,15 @@ class F
         $in   = str_replace('\\', '/', $in);
 
         if (Str::contains($file, $in) === false) {
-            return $file;
+            $fileParts = explode('/', $file);
+            $inParts = explode('/', $in);
+            while (count($fileParts) && count($inParts) && ($fileParts[0] == $inParts[0])) {
+                array_shift($fileParts);
+                array_shift($inParts);
+            }
+            
+            $parentToken = '../';
+            return str_pad('', count($inParts) * strlen($parentToken), $parentToken) . implode('/', $fileParts);
         }
 
         return Str::after($file, $in);
--- a/src/Toolkit/F.php
+++ b/src/Toolkit/F.php
@@ -654,11 +654,15 @@ class F
         $file = str_replace('\\', '/', $file);
         $in   = str_replace('\\', '/', $in);
 
-        if (Str::contains($file, $in) === false) {
+        // trim trailing slashes
+        $file = rtrim($file, '/');
+        $in   = rtrim($in, '/');
+
+        if (Str::contains($file, $in . '/') === false) {
             // make the paths relative by stripping what they have
             // in common and adding `../` tokens at the start
-            $fileParts = explode('/', rtrim($file, '/'));
-            $inParts = explode('/', rtrim($in, '/'));
+            $fileParts = explode('/', $file);
+            $inParts = explode('/', $in);
             while (count($fileParts) && count($inParts) && ($fileParts[0] === $inParts[0])) {
                 array_shift($fileParts);
                 array_shift($inParts);
@@ -668,7 +672,7 @@ class F
             return str_pad('', count($inParts) * strlen($parentToken), $parentToken) . implode('/', $fileParts);
         }
 
-        return Str::after($file, $in);
+        return '/' . Str::after($file, $in . '/');
     }
 
     /**
--- a/src/Toolkit/F.php
+++ b/src/Toolkit/F.php
@@ -655,9 +655,11 @@ class F
         $in   = str_replace('\\', '/', $in);
 
         if (Str::contains($file, $in) === false) {
+            // make the paths relative by stripping what they have
+            // in common and adding `../` tokens at the start
             $fileParts = explode('/', $file);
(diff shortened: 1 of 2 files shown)


distantnative:
While I understand the issue you are tying to solve with this change, having a method called `F::relativepath()` return an absolute path now doesn't seem to me very logical. It would be a breaking change at a very low-level toolkit method as well.

Wondering if we should rather fix this at https://github.com/getkirby/kirby/blob/38afb2d650f3764d23d732b23ae8f04ab9ffba69/src/Api/Api.php#L689

neildaniels:
@distantnative You're right. It is odd that this would start returning absolute paths. (Though, I do think it was a bug to merely return a filename in the unrelated base situation).

This PR has been updated to add `../` as necessary to have a complete travel path.

neildaniels:
PHPMD complained about using `count()` in the `while` loop, but this seems fine to me.

I added a suppression statement for that warning, but feel free to suggest an alternative.

src/Toolkit/F.php
```diff
@@ -653,7 +655,15 @@ public static function relativepath(string $file, string $in = null): string
         $in   = str_replace('\\', '/', $in);
 
         if (Str::contains($file, $in) === false) {
-            return basename($file);
+            $fileParts = explode('/', $file);
+            $inParts = explode('/', $in);
+            while (count($fileParts) && count($inParts) && ($fileParts[0] == $inParts[0])) {
```
lukasbestle:
```suggestion
            while (count($fileParts) && count($inParts) && ($fileParts[0] === $inParts[0])) {
```

src/Toolkit/F.php
```diff
@@ -653,7 +655,15 @@ public static function relativepath(string $file, string $in = null): string
         $in   = str_replace('\\', '/', $in);
 
         if (Str::contains($file, $in) === false) {
-            return basename($file);
+            $fileParts = explode('/', $file);
```
lukasbestle, line 664:
```suggestion
            // make the paths relative by stripping what they have
            // in common and adding `../` tokens at the start
            $fileParts = explode('/', $file);
```

lukasbestle (changes requested):
Thanks for your PR! I have a few code suggestions below and also a general request: Could you please add tests for when `$in` ends with a slash? I think we need `rtrim($in, '/')`, otherwise your counting algorithm would not work correctly.

A comment on the history of the method:

> Though, I do think it was a bug to merely return a filename in the unrelated base situation

The reason why the filename gets returned in this case is to avoid leaking the full path when the developer expects to securely get a relative path. So it's a security feature.

distantnative:
Have been wondering if a `for`-loop would make PHPMD happier.

lukasbestle:
Thank you! I wonder if we should always `rtrim()` the slashes, even if the path *does* start with `$in`. It would kind of be a breaking change, but much more consistent. What do you think @distantnative?

neildaniels:
@lukasbestle @distantnative Here's what always trimming slashes would look like: https://github.com/neildaniels/kirby/pull/1

I agree that it would be more consistent. If this looks desired, I can merge it in before this PR is merged.

lukasbestle:
@neildaniels Thank you! I'm not sure about the change that the return value longer begins with `/` though. Being a relative path it makes much more sense that way, but it would be yet another breaking change.
@distantnative What do you think about the breaking changes and the optimal behavior of this method?

distantnative:
I'm not sure removing the slash has a lot of benefit - probably not enough to justify the breaking change. I would leave it in. Otherwise looks good IMO.

lukasbestle:
@distantnative Do you mean just the slash at the beginning or also the `rtrim()`ed slashes at the end? I think the latter is really useful for consistency of this method.

distantnative:
I meant only the one at the beginning.

lukasbestle:
Agreed, then we are on the same page.

neildaniels:
@lukasbestle Restored the leading slash. I went ahead and merged in that branch into this PR.

bastianallgeier (dismissed):
It's looking good to me. @lukasbestle @distantnative final thoughts?

distantnative:
Looks fine to me too

lukasbestle (commented):
I really like the implementation, thanks @neildaniels! 👍
One question, besides that it's ready for merging:

src/Toolkit/F.php
```diff
@@ -652,11 +654,25 @@ public static function relativepath(string $file, string $in = null): string
...
+            // in common and adding `../` tokens at the start
+            $fileParts = explode('/', $file);
+            $inParts = explode('/', $in);
+            while (count($fileParts) && count($inParts) && ($fileParts[0] === $inParts[0])) {
+                array_shift($fileParts);
+                array_shift($inParts);
+            }
+            
+            $parentToken = '../';
+            return str_pad('', count($inParts) * strlen($parentToken), $parentToken) . implode('/', $fileParts);
```
lukasbestle: Could this line be replaced with this or have I misunderstood it?

```suggestion
            return str_repeat($parentToken, count($inParts)) . implode('/', $fileParts);
```

neildaniels: Absolutely, that's much clearer.

lukasbestle (approved):

PR #3271, opened 2021-04-27
Use Toolkit\Obj instead Email\Body

distantnative:
I think we can get rid of `Kirby\Email\Body` as `Kirby\Toolkit\Obj` just works fine here.


lukasbestle:
@distantnative Why was this closed? I think it makes sense to make this simpler.

distantnative:
I was torn, might make the setup harder to understand - since `Obj` is quite magic. If we change it, we could also simply keep in array in the `Email` class.

But if more think this is a good change, let's do it. I closed it mainly because I wasn't 100% sure myself anymore. And thought I'll revisit when looking at the email namespace in general.

lukasbestle:
I often think it's quite sad when you already invested time into a feature and then scrap it again later. Maybe ask us first what to do with it or comment why you close it so we know why and can give feedback on that. :)

I'm torn as well about this. On one hand the `Body` class is so incredibly simple and basically useless, but on the other hand it gives the objects a clear structure with free documentation. A pure array wouldn't be an alternative (different access, which would be a breaking change), but `Obj` could work and is much simpler (minus the automatic documentation).

distantnative:
> I often think it's quite sad when you already invested time into a feature and then scrap it again later. Maybe ask us first what to do with it or comment why you close it so we know why and can give feedback on that. :)

Thanks :D I think a comment could work. But I have also made the experience that sometimes it needs a few approaches to really hit the nail. So I don't mind trying something and throwing it away when it doesn't feel right anymore - still learned something and often gravitating to the better solution in the end.

Opening to get opinions from the others :)

lukasbestle:
> So I don't mind trying something and throwing it away when it doesn't feel right anymore - still learned something and often gravitating to the better solution in the end.

Makes sense. 🙂

For these PRs where you are not sure yet yourself if they are the right way I have one suggestion: Maybe only commit them to a local branch first and then make a PR once they feel ready for you? This would reduce the notifications in our inbox and we wouldn't need to think about the reasons why a PR was closed.

distantnative:
Closing this to remove some clutter from GitHub. Will re-evaluate this idea after the refactoring of `Properties`.

PR #3277, opened 2021-04-29
`video` kirbytag supports local and remote videos #3104

### LINKED ISSUE #3104: Video kirbytag supporting local and remote video
Proof of concept plugin: https://github.com/getkirby/getkirby.com/pull/1188

afbora:
## Describe the PR

This PR implements the https://github.com/getkirby/getkirby.com/pull/1188 PR with some improvements, bug fixes and unit tests.

## 🎉 Features

Supported local and remote video files by extending video tag that supporting only youtube and vimeo videos. Youtube and vimeo features have not been changed.

### New Attributes:
- autoplay
- controls (default: true)
- loop
- muted *
- poster
- preload (auto, metadata, none)

_* muted will enabled if autoplay enabled and muted is not defined_

Usages:

````markdown
# local
(video: local-video.mp4)

# remote
(video: https://www.getkirby.com/sample-video.mp4)

# example 1
(video: local-video.mp4  autoplay: true)

# example 2
(video: local-video.mp4 controls: false autoplay: true loop: true)

# example 3
(video: local-video.mp4 poster: cover.jpg)

# example 4
(video: local-video.mp4 preload: auto)

# example 5
(video: https://www.getkirby.com/sample-video.mp4 muted: true controls: false autoplay: true)

# example 6
(video: local-video.mp4 poster: https://www.getkirby.com/sample-cover.jpg)

````

## Related issues

- Fixes #3104 

## Ready?

### DIFF
--- a/config/tags.php
+++ b/config/tags.php
@@ -2,6 +2,9 @@
 
 use Kirby\Cms\Html;
 use Kirby\Cms\Url;
+use Kirby\Text\KirbyTag;
+use Kirby\Toolkit\F;
+use Kirby\Toolkit\Str;
 
 /**
  * Default KirbyTags definition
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/Cms/File.php [hunk 1]
```diff
@@ -248,12 +248,16 @@ public function dragText(string $type = null, bool $absolute = false): string
         if ($type === 'markdown') {
             if ($this->type() === 'image') {
                 return '![' . $this->alt() . '](' . $url . ')';
+            } elseif ($this->type() === 'video') {
+                return '[![ALT TEXT](IMAGE URL)](VIDEO URL)';
```
afbora: Any idea about that?

bastianallgeier: Hm, this is tricky. There's no video tag in markdown. We could only inject an html video tag instead.

afbora: @bastianallgeier You mean like following. 
So actually rendered `video` tag in directly panel text editor?

**Write output to editor;**

If local or remote video

    <figure class="video"><video controls><source src="LOCAL-REMOTE-VIDEO-URL" type="video/mp4"></video></figure>

If youtube or vimeo video
    
    <figure class="video"><iframe allowfullscreen src="YOUTUBE-VIMEO-URL"></iframe></figure>

afbora: Or should we just link to the file (local video) or video url (youtube, vimeo, remote video)?

bastianallgeier: It's definitely not ideal. Another option could be to add a simple markdown link to the video instead. I don't know what makes more sense from a user perspective. All HTML is quite annoying in the context of markdown, but the link is just a very bad compromise either. Maybe keep it simple for this first implementation and only render the markdown link. Then we wait for feedback.

distantnative: I would go for the link. There is no video tag in Markdown that's the reality. Putting a lot of HTML in it should be a conscious decision by the user

afbora:
Now ready to review 🚀

config/tags.php
```diff
@@ -236,23 +239,117 @@
...
+                    (isset($tag->{$key}) === false || $tag->{$key} === null)
+                ) {
+                    $tag->{$key} = $value;
+                }
+            }
+
+            // generates tag attributes helper inline function
+            // for ex: `autoplay` will not work if `false` is a `string` instead of a `boolean`
+            // this is why the type converter is used in function
+            $generateTagAttributes = function ($tag) {
```
distantnative: Do we need an inline function here if we only call it once? Could be just the code, without putting it in a function, or?

afbora: You are absolutely right. Previously, I was calling this inline function from two places. So I needed an inline function. After improvement I used it in one place but forgot it there 🙈

afbora: Done

bastianallgeier (approved):
I love this! I'm sure this is going to make a lot of users happy.

Issue #777, opened 2018-07-30
[Cms] Less strict plugin name validation


bastianallgeier:

bastianallgeier:
✅

PR #3279, opened 2021-04-29
Supports HTML for toggle and checkbox field labels

### LINKED ISSUE #3278: Blueprint > query > text - Does not interpret html markup anymore?
**When using HTML markup in text query option in a Blueprint, HTML is not interpreted anymore**  

**Example:**  
```
evaluatorProposals:
    label: Proposals to evaluate
    type: checkboxes
    width: 3/4
    options: query
    query: 
      fetch: page.parent.children.unlisted.template('proposal')
      text: '{{ page.title }} <span style="width:125px;text-align:right;font-size:0.75em;color:#777;float:right">{{ page.nbrEvaluators }}</span>'
```

**Expected behavior**  
The HTML markup should be interpreted in the panel.

**Kirby Version**  
Work in Kirby 3.3.6 for instance but not in 3.5.3.1

 - OS: Linux
 - Browser: Firefox
 - Version: 88.0 (64 bits)

afbora:
## Describe the PR

Escaped values with the `Str::template()` method developed for the reliability of the data coming from the API.
Parts that previously supported HTML have been removed. 
It was brought back again with disabled `eslint` checks.

### Enhancements
- `$fallback` param in `Str::template()` renamed to `$options` that accept string or array that can be contains `fallback`, `callback`, `start`, `end`.

### Fixes
- Toggle and checkbox field labels supports HTML again

## Breaking changes

None

## Related issues/ideas

- Fixes #3278 

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Form/OptionsApi.php
+++ b/src/Form/OptionsApi.php
@@ -86,10 +86,12 @@ class OptionsApi
      * @param array $data
      * @return string
      */
-    protected function field(string $field, array $data)
+    protected function field(string $field, array $data): string
     {
         $value = $this->$field();
-        return Str::template($value, $data);
+        return Str::template($value, $data, null, '{{', '}}', function ($result) {
+            return strip_tags($result);
+        });
     }
 
     /**
--- a/src/Form/OptionsApi.php
+++ b/src/Form/OptionsApi.php
@@ -7,6 +7,7 @@ use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Http\Remote;
 use Kirby\Http\Url;
+use Kirby\Toolkit\Escape;
 use Kirby\Toolkit\Properties;
 use Kirby\Toolkit\Query;
 use Kirby\Toolkit\Str;
@@ -89,8 +90,8 @@ class OptionsApi
     protected function field(string $field, array $data): string
     {
         $value = $this->$field();
-        return Str::template($value, $data, null, '{{', '}}', function ($result) {
-            return strip_tags($result);
+        return Str::template($value, $data, function ($result) {
+            return Escape::html($result);
         });
     }
 
--- a/src/Form/OptionsApi.php
+++ b/src/Form/OptionsApi.php
@@ -90,9 +90,11 @@ class OptionsApi
     protected function field(string $field, array $data): string
     {
         $value = $this->$field();
-        return Str::template($value, $data, function ($result) {
-            return Escape::html($result);
-        });
+        return Str::template($value, $data, [
+            'callback' => function ($result) {
+                return Escape::html($result);
+            }
+        ]);
     }
 
     /**
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -968,14 +968,31 @@ class Str
      * @param string|null $string The string with placeholders
      * @param array $data Associative array with placeholders as
      *                    keys and replacements as values
-     * @param string|null $fallback A fallback if a token does not have any matches
+     * @param string|array|\Closure|null $fallback A fallback if a token does not have any matches
      * @param string $start Placeholder start characters
      * @param string $end Placeholder end characters
-     * @param \Closure|null $callback A callback to be able to modify each matching result
      * @return string The filled-in string
      */
-    public static function template(string $string = null, array $data = [], string $fallback = null, string $start = '{{', string $end = '}}', Closure $callback = null): string
+    public static function template(string $string = null, array $data = [], $fallback = null, string $start = '{{', string $end = '}}'): string
     {
+        // handle fallback that can be string, closure, array or null
+        if ($fallback !== null) {
+            if (is_array($fallback) === true) {
+                $callback = $fallback['callback'] ?? null;
+                $fallback = $fallback['fallback'] ?? null;
+            } elseif (is_a($fallback, 'Closure') === true) {
+                $callback = $fallback;
+                $fallback = null;
+            } elseif (is_string($fallback) === true) {
+                $callback = null;
+            } else {
+                $callback = null;
+                $fallback = null;
+            }
+        } else {
+            $callback = null;
+        }
+
         return preg_replace_callback('!' . $start . '(.*?)' . $end . '!', function ($match) use ($data, $fallback, $callback) {
             $query = trim($match[1]);
 
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Toolkit;
 
+use Closure;
 use Exception;
 
 /**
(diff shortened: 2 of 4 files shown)


bastianallgeier:
This has been removed to avoid XSS issues as far as I remember. We shouldn't simply revert it.

afbora:
Hmm I get it. Since I don't know, I ask: we use it in other fields as well. Is there a specific reason we can't use it here?

bastianallgeier:
I think the main reason was that the options could come from an API or other external URL. But to be honest, I don't fully remember the entire conversation about it anymore. I just want us to be careful.

distantnative:
Yes, I think that was the reason that those options could come from something not in control e.g. by the admin, while blueprints (for where it is used in other places) usually are

afbora:
I understood properly now. If there is no other solution, can you mark the problem as unsolvable and explain in #3278 please?

bastianallgeier:
I don't think we should discard this. We should just look into it more carefully and see if we can fix it in any other way.

distantnative:
I think we could activate it on Vue side if at the same time we make sure in our options classes to strip any tags from API call results.

Of course, a developer could use a page method/model as well and in there use external sources - but here I would argue it's in the hands of the developer to make sure to escape everything then.

afbora:
> we make sure in our options classes to strip any tags from API call results.

This makes sense, great 👍 
But if I understand it correctly, If we strip all tags, how are we going to output html as in the issue?
Or how do we determine `$allowed_tags`?

    '{{ page.title }} <span style="width:125px;text-align:right;font-size:0.75em;color:#777;float:right">{{ page.nbrEvaluators }}</span>'

nilshoerrmann:
Please consider cases where either an API is well under control by the developer or the input from a Kirby page was initially created using a frontend form. Both sides can either be compromised or secure. It's not black and white and Kirby cannot determine the status of the content by source.

I'm not sure if it's technically feasible (thinking of complexity, performance …) but would it be possible to have something like guards for dynamic content that could be activated in the blueprints. E. g. by default there is a guard that removes all markup but it can be actively deactivated, like `guard: false`. There could be custom guards as well where a developer allows for certain tags but not all. So Kirby could be strict by default but leave the final responsibility to the developer. Just as an idea.

distantnative:
Is it really a thing that we want HTML tags rendered in the Panel coming from an API? I would have rather considered the case that you get data from an API but in the blueprint option you define some HTML how the option should be rendered.

afbora:
You can see my idea for filtering the data coming from the API [in this commit.](https://github.com/getkirby/kirby/pull/3279/commits/a98c60a9612e6d1823838cc779f674d2b3b8e9f8) Do you think it makes sense and does it work?

distantnative:
@afbora I guess I wouldn't add another parameter to the Str method. Rather make it two steps.

```
$foo = Str::template();
$foo = strip_tags($foo);
```

afbora:
No, your method will corrupt the entire HTML output.

Template
````
<strong>{{ item.name }}</strong>
````

API Input
````
Apple <script>alert('Hi!')</script>
````

Output
````
Apple alert('Hi!')
````

---

My method just clears the value returning and matching from the API.

Template
````
<strong>{{ item.name }}</strong>
````

API Input
````
Apple <script>alert('Hi!')</script>
````

Output
````
<strong>Apple alert('Hi!')</strong>
````

**Comment by afbora** [comment]
### Summarize

- Initial post updated
- Rebased

Since the data returned from the API is processed in the `Str::template()` method, we can act on each item that matches the new `callback` method.

File: src/Form/OptionsApi.php [hunk 1]
```diff
@@ -86,10 +86,12 @@ public function fetch()
      * @param array $data
      * @return string
      */
-    protected function field(string $field, array $data)
+    protected function field(string $field, array $data): string
     {
         $value = $this->$field();
-        return Str::template($value, $data);
+        return Str::template($value, $data, null, '{{', '}}', function ($result) {
+            return strip_tags($result);
```
lukasbestle: `strip_tags()` does not protect against XSS attacks. Also it would remove tag-like text completely. For backwards compatibility and security it would be better if the API return values were escaped with `Toolkit\Escape::html()` (that's what `v-text` does in Vue as well).

src/Toolkit/Str.php
```diff
@@ -964,17 +965,18 @@ public static function substr(string $string = null, int $start = 0, int $length
...
      *                    keys and replacements as values
-     * @param string $fallback A fallback if a token does not have any matches
+     * @param string|null $fallback A fallback if a token does not have any matches
      * @param string $start Placeholder start characters
      * @param string $end Placeholder end characters
+     * @param \Closure|null $callback A callback to be able to modify each matching result
      * @return string The filled-in string
      */
-    public static function template(string $string = null, array $data = [], string $fallback = null, string $start = '{{', string $end = '}}'): string
+    public static function template(string $string = null, array $data = [], string $fallback = null, string $start = '{{', string $end = '}}', Closure $callback = null): string
```
lukasbestle: I agree with Nico that the additional argument makes the method signature pretty long and unwieldy.

Idea: What if the current `$fallback` argument supported an `$options` array alternatively? Then the method could be used like this:

```php
Str::template('my {{ template }}', $data, [
    'fallback' => '...',
    'callback' => function() { ... }
]);
```

The argument would still be called `$fallback` for now (renaming it would be a breaking change because of PHP 8's named arguments). Only the `string` type hinting would be removed.

We could then deprecate the previous call signature, add a deprecation warning in 3.6.0 and remove support in 3.7.0.

I think this would be much cleaner in the long run (like `Cookie::set()` where we also introduced `$options` in that way if I remember correctly).

afbora: I thought and worked on `$fallback` but I gave up because it would be a confusing. The idea of a `$fallback`  that can be a array looks great. I will try.

afbora: @lukasbestle I've applied your idea. What do you think?

lukasbestle: The direct support for the closure is pretty cool for our use case, but I'm not sure if it's good for the future. One argument that can do several things is quite confusing.

What I meant is that `$fallback` gets the double role only temporarily until we have migrated to a `array $options` argument. This would then also support `start` and `end` keys, so in 3.7 the method signature could look like this:

```php
public static function template(?string $string, array $data = [], array $options = []): string
```

Nothing carved in stone though. What do you think @distantnative @bastianallgeier?

afbora: Yes, it's easier to understand if it's just `string` and `array` for now, and it'll make it even easier to transition in the future.

afbora: I applied it because we can easily revert.  Also, by following the commits and the final version, maybe we can end the discussion faster.

afbora:
## Summarize

Since the data returned from the API is processed in the `Str::template()` method, we can act on each item that matches the new callback method.

Instead of adding a new parameter, it is both backward compatible and forward-looking, by changing the `$fallback` parameter to `$options` as suggested by Nico and Lukas.

@distantnative @bastianallgeier Open question for you https://github.com/getkirby/kirby/pull/3279#discussion_r647570116

bastianallgeier (approved):
After finally fully understanding the new $options array, I'm really happy with it! The plan to deprecate the string argument also makes a lot of sense. Great work!

bastianallgeier:
Shit. I think I just made a mistake. Wouldn't this PR already be a great fix for 3.5.7? What do you think?

afbora:
Yes, I think it would be great to have it in 3.5.7, as we can count security enhancements as well. @lukasbestle and @distantnative's thoughts would be great.

PR #3284, opened 2021-04-30
Fixes writer field preview props in blocks field #3145

### LINKED ISSUE #3145: Blocks writer preview doesn't apply field props
**Describe the bug**  
Props applied to the field are not applied to the block preview.

**To Reproduce**  
1. Go to panel
2. Add new block type that included writer field like following sample
3. Disable `marks` and `nodes` prop
4. You'll see nodes and marks on preview but not in drawer (as expected).

````yaml
name: Test
wysiwyg: true
preview: text
fields:
  text:
    type: writer
    nodes: false
    marks: false
````

**Expected behavior**  
The field props should also be applied to field preview.

**Kirby Version**  
3.5.2.RC-1

 - OS: Windows 10
 - Browser: Chrome latest

afbora:
## Describe the PR

The missing props were added for blocks field text preview component.

**Side note:**

There are a few missing props in the other block type previews as well but I didn't want to fix it in this PR. 
Should I create new PR for them too or will we fix it later?

## Related issues

- Fixes #3145 

## Ready?

### DIFF
--- a/panel/src/components/Blocks/Types/Text.vue
+++ b/panel/src/components/Blocks/Types/Text.vue
@@ -1,8 +1,10 @@
 <template>
   <k-writer
     ref="input"
-    :nodes="false"
-    :placeholder="placeholder"
+    :inline="textField.inline"
+    :marks="textField.marks"
+    :nodes="textField.nodes"
+    :placeholder="textField.placeholder"
     :value="content.text"
     class="k-block-type-text-input"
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):

bastianallgeier:
@afbora yes, additional PRs would be great. But they don't have to be finished for 3.5.5

Issue #3233, opened 2021-04-07
Move file methods to the classes they belong to


lukasbestle:
From a [discussion in Discord](https://discord.com/channels/525634039965679616/827138504646721586/829321125707513886) with @distantnative:

> All files are Kirby\Cms\File, no matter if PDFs, images... Kirby\Image\Image is a class to take care of image conversions etc. Will need some more time thinking why these methods ended up there.
>My gut feeling is that it actually should be three things: Kirby\Cms\File for file objects in the Kirby sense, Kirby\Something\File for everything that is related to it as an actual file in the filesystem and Kirby\Image\Image for everything specific to images.

> That Kirby\Something\File already exists as Kirby\Toolkit\File. I fully agree that we should move methods around between the classes at some point.

> Indeed, we should proxy then first Kirby\Toolkit\File for all files and only for images Kirby\Image\Image as well. Or maybe it would be even better to convert the methods from Kirby\Toolkit\File into a trait that can be added directly to Kirby\Cms\File as well (since it already has to extend Kirby\Cms\ModelWithContent).

distantnative:
✅

PR #3285, opened 2021-05-01
New `Filesystem` namespace

distantnative:
## Status: ready for review 👀

New `Kirby\File` namespace featuring a new classes setup for handling file assets in Kirby:

#### `Kirby\Toolkit`
- `Kirby\Toolkit\File` takes care of everything CMS-unrelated. Gains a few more methods (e.g. `download`, `headers`) that couple it a bit more with other packages (e.g. `Kirby\Http`)

#### `Kirby\Filesystem`
- New namespace bundling several file-relates classes
- New class `Kirby\Filesystem\File` extends `Kirby\Toolkit\File` with some Cms-related functions (e.g. `::modified()` with handler, `::isResizable()`) - takes over a lot of functionality that was previously in `Kirby\Image\Image` even for non-images

- New trait `Kirby\Filesystem\IsFile` adds method-proxying to an asset object - either to `Kirby\Image\Image` or `Kirby\Filesystem\File` depending on the file type of the `$root` property
- `Kirby\Cms\FileFoundation` is removed - it gets replaced by `Kirby\Filesystem\IsFile` and all proxied methods of `Kirby\Filesystem\File`/`Kirby\Image\Image` which reach down to  `Kirby\Toolkit\File` as well

- `Kirby\Cms\Asset` has been moved to the new namespace as `Kirby\Filesystem\Asset`
- `Kirby\Cms\Filename` has been moved to the new namespace as `Kirby\Filesystem\Filename`

#### `Kirby\Image`
-  `Kirby\Image\Image` now extends `Kirby\Filesystem\File` and adds image-related methods (e.g. `dimensions`, `exif`)

## Breaking changes
- There are aliases for mapping old class names with new ones
- `Kirby\Image\Image::__toString()` returns an `` tag instead the file path.
- File hooks: the `$upload` parameter is a `Kirby\Filesystem\File` object, before `Kirby\Image\Image`
- `Kirby\Toolkit\File::header()` now returns `void` instead of an empty `string` by default
- The `Asset::alt()` method was removed as it didn't have a function on its own

## Related issues
- Closes https://github.com/getkirby/kirby/issues/3233

## Ready?

## When merging


bastianallgeier:
Ok, before we move on with this, how about introducing a Kirby\Filesystem namespace instead and we move Dir and F there as well?

distantnative:
@bastianallgeier I wouldn't move Toolkit classes into other namespaces... I feel like Toolkit classes are somehow special and (mostly) universal - not bound to the Cms? But maybe that thinking is false.

bastianallgeier:
To be honest, the toolkit is more like an "I don't know where to put it" namespace 🤣  Otherwise some HTTP stuff and other classes could also be in the toolkit.

distantnative:
True. Ok, fine with that. I like `Filesystem` as it makes the distinction to the model clearer.

Aaaaand... breaking up Toolkit: so many more possibilities now! :D

bastianallgeier:
Hahaha, keep in mind that we don't need to achieve all of this in 3.6

distantnative:
We don't need to, but...

I have changed the name to `Filesystem`, I would leave moving `F`, `Dir` etc. for a separate PR.

config/components.php
```diff
@@ -305,15 +305,15 @@
      * Add your own thumb generator
      *
      * @param \Kirby\Cms\App $kirby Kirby instance
-     * @param string $src The root of the original file
-     * @param string $template The template for the root to the desired destination
+     * @param string $src root of the original file
+     * @param string $dst template string for the root to the desired destination
```
lukasbestle:
```suggestion
     * @param string $src Root of the original file
     * @param string $dst Template string for the root to the desired destination
```

I also feel like the param should be named `$destination`, not `$dst` (it's not really Kirby-like to abbreviate param names).

src/Cms/File.php
```diff
@@ -106,12 +97,12 @@ class File extends ModelWithContent
...
     public function __call(string $method, array $arguments = [])
     {
-        // public property access
+        // Public property access
         if (isset($this->$method) === true) {
             return $this->$method;
         }
 
-        // asset method proxy
+        // Asset method proxy
```
lukasbestle: Why these changes? In our PHP code we use lowercase line comments throughout.

src/Filesystem/File.php
```diff
@@ -0,0 +1,255 @@
...
+use Kirby\Cms\App;
+use Kirby\Exception\Exception;
+use Kirby\Toolkit\File as BaseFile;
+use Kirby\Toolkit\Html;
+use Kirby\Toolkit\Mime;
+use Kirby\Toolkit\Properties;
+use Kirby\Toolkit\V;
+
+/**
+ * A representation of an file in the filesystem.
```
```suggestion
 * A representation of a file in the filesystem.
```

src/Filesystem/File.php
```diff
@@ -0,0 +1,255 @@
...
+     *
+     * @var string|null
+     */
+    protected $url;
+
+    /**
+     * Constructor sets all file properties
+     *
+     * @param string|array|null $props
+     * @param string|null $url
```
```suggestion
     * @param array|string|null $props Properties or deprecated `$root` string
     * @param string|null $url Deprecated argument, use `$props['url']` instead
```

src/Filesystem/File.php
```diff
@@ -0,0 +1,255 @@
...
+     *
+     * @param array $rules
+     * @return bool
+     * @throws \Kirby\Exception\Exception
+     */
+    public function match(array $rules): bool
+    {
+        $rules = array_change_key_case($rules);
+
+
```
```suggestion

```

src/Filesystem/File.php
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * Runs a set of validations on the file object
+     * (mainly for images).
+     *
+     * @param array $rules
+     * @return bool
+     * @throws \Kirby\Exception\Exception
+     */
+    public function match(array $rules): bool
```
lukasbestle: I think this method could be part of `Toolkit\File` as it's not CMS-specific.

src/Filesystem/File.php
```diff
@@ -0,0 +1,255 @@
...
+                $validator = $arguments[1];
+
+                if (V::$validator($this->$property(), $rule) === false) {
+                    throw new Exception([
+                        'key'  => 'file.' . $key,
+                        'data' => [$property => $rule]
+                    ]);
+                }
+            }
+        }
```
lukasbestle: This part should be in the `Image` class IMO.

src/Filesystem/File.php
```diff
@@ -0,0 +1,255 @@
...
+
+        return true;
+    }
+
+    /**
+     * Get the file's last modification time.
```
```suggestion
     * Returns the file's last modification time
```

src/Filesystem/File.php
```diff
@@ -0,0 +1,255 @@
...
+     *
+     * @return string|null
+     */
+    public function url(): ?string
+    {
+        return $this->url;
+    }
+
+    /**
+     * Convert the object to an array
```
```suggestion
     * Converts the object to an array
```

src/Filesystem/File.php
```diff
@@ -0,0 +1,255 @@
...
+        ]);
+
+        ksort($array);
+
+        return $array;
+    }
+
+
+    /**
+     * Return the url for the file object
```
```suggestion
     * Returns the URL for the file object
```

src/Filesystem/IsFile.php
```diff
@@ -0,0 +1,159 @@
...
+    protected $asset;
+
+    /**
+     * Absolute file path
+     *
+     * @var string|null
+     */
+    protected $root;
+
+
```
lukasbestle:
```suggestion

```

src/Filesystem/IsFile.php
```diff
@@ -0,0 +1,159 @@
...
+     * Magic caller for asset methods
+     *
+     * @param string $method
+     * @param array $arguments
+     * @return mixed
+     * @throws \Kirby\Exception\BadMethodCallException
+     */
+    public function __call(string $method, array $arguments = [])
+    {
+        // Public property access
```
lukasbestle:
```suggestion
        // public property access
```

src/Filesystem/IsFile.php
```diff
@@ -0,0 +1,159 @@
...
+        if (isset($this->$method) === true) {
+            return $this->$method;
+        }
+
+        // Asset method proxy
```
lukasbestle:
```suggestion
        // asset method proxy
```

src/Filesystem/Name.php
```diff
@@ -1,16 +1,16 @@
 <?php
 
-namespace Kirby\Cms;
+namespace Kirby\Filesystem;
 
 use Kirby\Toolkit\Str;
 
 /**
- * The Filename class handles complex
+ * The `Name` class handles complex
```
lukasbestle: If we name the package `Filesystem` (👍), I think the `Name` class should still be named `Filename`, otherwise the mental connection is lost. With `File\Name` it made sense, but `Filesystem\Name` is weird: What kind of name?

src/Image/Image.php
```diff
@@ -293,38 +191,30 @@ public function orientation(): string
     }
 
     /**
-     * Converts the media object to a
-     * plain PHP array
+     * Convert the object to an array
```
```suggestion
     * Converts the object to an array
```

src/Image/Image.php
```diff
@@ -293,38 +191,30 @@ public function orientation(): string
...
-    {
-        return json_encode($this->toArray());
+        ksort($array);
+
+        return $array;
     }
 
     /**
-     * Returns the url
+     * Return the url for the file object
```
```suggestion
     * Returns the URL for the file object
```

src/Toolkit/File.php
```diff
@@ -99,6 +100,18 @@ public function delete(): bool
         return true;
     }
 
+    /*
+     * Automatically sends all needed headers for the file to be downloaded
+     * and echos the file's content
+     *
+     * @param  string|null $filename  Optional filename for the download
```
lukasbestle:
```suggestion
     * @param string|null $filename Optional filename for the download
```

src/Toolkit/File.php
```diff
@@ -139,6 +152,18 @@ public function hash(): string
...
+     * Sends an appropriate header for the asset
+     *
+     * @param bool $send
+     * @return \Kirby\Http\Response|string
+     */
+    public function header(bool $send = true)
+    {
+        $response = new Response('', $this->mime());
+        return $send === true ? $response->send() : $response;
+    }
```
lukasbestle: I feel like we should remove the `string` return value and return `null` if `$send` is `true`. The returned string is always empty because the body is set to an empty string the line above.

This is a breaking change, but IMO there's no point returning an empty string here. It's only confusing.

tests/Cms/Files/FileVersionTest.php
```diff
@@ -39,4 +39,22 @@ public function testConstruct()
...
+        $original = new File([
+            'filename' => 'test.jpg',
+            'content' => [
+                'title' => 'Test Title'
+            ]
+        ]);
+
+        $version = new FileVersion([
+            'original'      => $original,
+            'root'          => __DIR__ . '/fixtures/files/test.jpg'
```
lukasbestle:
```suggestion
            'original' => $original,
            'root'     => __DIR__ . '/fixtures/files/test.jpg'
```

lukasbestle (commented):
I really really love how clean this all became immediately. ❤️ 
I agree we should move the other filesystem-related classes as well, but we can do it in another PR.

A general comment: I feel like the `Filesystem` package should be MIT-licensed as it's not CMS specific.

distantnative:
Great comments! Will go through them once I feel a bit more alive again

config/components.php
distantnative: @lukasbestle I thought it goes together well with `$src`. And that we use it the same way e.g. in the `App` class.

lukasbestle: Right, makes sense.

src/Toolkit/File.php
distantnative: Would mean the code would need to look something like

```php
$response = new Response('', $this->mime());

if ($send !== true) {
    return $response;
}

$result = $response->send();
return empty($result) !== true ? $result : null;
```

IMO I am not sure if this is such a good change since it differs how other parts like `Response::send` work. In general, it makes sense but we probably should change it in more places then (and on a separate PR).

lukasbestle:
I've rebased the PR against `develop` as somehow a lot of unrelated commits ended up in here. Also I've bundled all suggestion fixes into one single commit to make the commit history less confusing.

src/Toolkit/File.php
lukasbestle: The `$result` is always empty as it's the body that we explicitly pass as `''` in the first line. So the code would be this:

```php
$response = new Response('', $this->mime());

if ($send !== true) {
    return $response;
}

$response->send();
```

lukasbestle: I think this is the only place where `Response` is only used to send headers. So I don't think this behavior would interfere with other uses.

But I'm open about leaving this as it is.

distantnative:
@lukasbestle Do you know a way to tell Github "commit all these selected suggested changes with a single commit"?

lukasbestle:
Yes, there is an "add to batch" button right next to the "commit" button.

distantnative:
Another open discussion point: Should this be `MIT` license?
as we are moving more and more code out of `Kirby\Cms` into dedicated namespaces, I am not sure if we should all this `MIT`.

lukasbestle:
I looked through the `Filesystem` code again and think that the classes we have in it at the moment are so CMS-specific that they can well stay proprietary.

If we were to move `F` and `Dir`, I think these should stay MIT though.
But then the new question would be: What happens with `Toolkit\File`? Merge it with the current `Filesystem\File`?

distantnative:
> But then the new question would be: What happens with Toolkit\File? Merge it with the current Filesystem\File?

That's mainly why I said, let's do that in a separate PR. That's where the headaches start :D

I wouldn't mind having different licenses within a package (per class).

lukasbestle:
Me neither. So I think the licenses of the code in the current PR are fine.

Now the `File::header()` discussion seems to be the only open one.

src/Toolkit/File.php
distantnative: @lukasbestle So is your suggestion to return `null` or nothing (void)? Your initial comment and your code differ there. I'm fine with either.

lukasbestle: Definitely `void`. My comment to "return `null`" was misunderstandable. I just meant that we shouldn't return anything in this case (= `void`), which automatically results in a `null` value.

I've changed it. :)

lukasbestle (dismissed):

lukasbestle:
It's really hard to track all changes across this PR, but I don't see any obvious issues. So let's roll with it. Again, it's really awesome how clean this all became. 🎉

src/Filesystem/Asset.php
```diff
@@ -1,54 +1,42 @@
...
+ * @package   Kirby Filesystem
  * @author    Bastian Allgeier <bastian@getkirby.com>
  * @link      https://getkirby.com
  * @copyright Bastian Allgeier GmbH
  * @license   https://getkirby.com/license
  */
 class Asset
 {
-    use FileFoundation;
-    use FileModifications;
```
lukasbestle, line 23: Oh, one thing I just noticed: Why was `FileModifications` removed?

distantnative: Good catch. I think reviewing all the code, I was really convinced that `FileModifications` in there was a mistake and wouldn't actually work. Now looking at it a week later, I think it does work and I was mistaken. We should add it back in then.

lukasbestle: What I think is also a bit weird is that the `Asset` class gets an `asset()` method from `IsFile`. But as that's always been the case, I think we should leave it in to avoid even more breaking changes.

distantnative: Yea, I have also moved in circles with the naming here. You quickly end up with having a `file()` method in `File` etc.

distantnative: @lukasbestle have added it again

src/Filesystem/Asset.php
```diff
@@ -1,54 +1,45 @@
...
-    }
-
-    /**
-     * Returns the alternative text for the asset
-     *
-     * @return null
-     */
-    public function alt()
-    {
-        return null;
```
lukasbestle, line 51: This method was basically useless before, but I wonder why it was there. Probably for compatibility, but with what? Do we need a replacement for it?

src/Image/Image.php
```diff
@@ -293,38 +206,30 @@ public function orientation(): string
...
-    {
-        return json_encode($this->toArray());
+        ksort($array);
+
+        return $array;
     }
 
     /**
-     * Returns the url
+     * Returns the URL for the file object
```
lukasbestle: It's not the URL but the HTML `` code. I'm thinking: Do we really need and want this breaking change?

tests/Filesystem/FilenameTest.php
```diff
@@ -1,12 +1,14 @@
 <?php
 
-namespace Kirby\Cms;
+namespace Kirby\Filesystem;
 
-class FilenameTest extends TestCase
+use PHPUnit\Framework\TestCase as TestCase;
+
+class NameTest extends TestCase
```
lukasbestle:
```suggestion
class FilenameTest extends TestCase
```

lukasbestle (commented):
Last round of comments after another detailed review. 😅
After that we can finally merge it. 🎉

src/Filesystem/Asset.php
bastianallgeier: If I remember correctly this is making sure that the img html can still be created correctly. For CMS files we have the alt attribute coming from the content file, but for assets, such a related content object does not exist.

lukasbestle: Right! `FileFoundation` called the `alt()` method from its `html()` method. `IsFile` doesn't have that method, instead it uses the one from `Filesystem\File` that no longer relies on `alt()` as that use-case was moved to `Image\Image`.

Removing the `alt()` method should therefore be safe IMO, but it's still a breaking-change.

src/Image/Image.php
distantnative: Changed it back to URL

src/Filesystem/Asset.php
distantnative: Exactly, I think it should be fine.

lukasbestle: I've added the breaking change to the first post.

lukasbestle (approved):

distantnative:
@bastianallgeier I'll let you have a final look to decide if we should merge now. Lukas and I are a bit anxious cause it is a big one but at the same time we think it's ready and shouldn't contain any bigger regressions anymore.

bastianallgeier:
It all looks very good to me. I think we still need to work on the F and Dir classes and especially on the CMS file stuff next to make it all fit together, but we are already quite deep in heavy changes that need careful launch preparations, once this is all finished. We should write down very clear instructions for everyone who might run into issues and we need a long and active test phase for this release.

bastianallgeier (approved):

distantnative:
@bastianallgeier moving F and Dir is quite straight forward. Why we have shied away from it so far is more `Toolkit\File` - that's one to really think about

lukasbestle:
I'm pretty confident we have collected all breaking changes that are relevant, so IMO we are on a good path.

PR #3287, opened 2021-05-02
Merge `Kirby\Cms\Api` into `Kirby\Api\Api`

distantnative:
## Status: in progress 👷

- Merges all additions from `Kirby\Cms\Api` into `Kirby\Api\Api`
- Simplifies our codebase (not duplicate `Api` classes only for strict package separation)
- `Kirby\Cms\Api` gets removed

## TODO
- [ ] Fix locale-related failing unit test 
- [x] Add class alias for `Kirby\Cms\Api` → `Kirby\Api\Api`

## Breaking changes

- The `kirby` prop isn't required anymore, if not passed `App::instance()` is used

Ideally no other, but needs to be reviewed again.

## Ready?

- [ ] CI checks pass

## When merging


distantnative:
@lukasbestle Maybe you can help me fixing this failing unit test? The test which is failing belonged already to the `Kirby\Api\Api` class. It must be tripped up by this code added from the former `Kirby\Cms\Api` class: 
https://github.com/getkirby/kirby/pull/3287/files#diff-47ba08c45090f63957cdc3a151f1ea22ea13529491b23e79af8b2215198ca4a0R182-R193

Not sure if that code needs to be added differently or if the test needs to be adapted (testing for a non-existing use case?). I believe this locale code is yours?

lukasbestle:
I will take a look once my todo list cleared up a bit.

lukasbestle:
I've rebased the PR against `develop` to fix the merge conflict. Weirdly the tests run just fine on my local machine. I'm currently trying to find out why they fail in CI.

lukasbestle:
@distantnative I still can't get my local PHPUnit to fail that test (??), but at least it's clear now where the test failure comes from:

In [the old `Api\Api` implementation](https://github.com/getkirby/kirby/pull/3287/files#diff-47ba08c45090f63957cdc3a151f1ea22ea13529491b23e79af8b2215198ca4a0R219) we are explicitly *not* setting `LC_CTYPE` because that caused issues before: https://github.com/getkirby/kirby/issues/2128

What's weird is that we have always been calling `$app->setCurrentLanguage()` in the `Api\Cms` implementation. This method also sets the PHP locales, but it uses the `LC_ALL` locale and therefore also overrides `LC_CTYPE` (???).

But it does so using the language provided in the request while the locale implementation in `Api\Api` uses the user language (????).

The amount of question marks in my head is just way too high. 😰

Unfortunately our locale and translation language handling is just a mess at the moment and I feel like we would need to clean that up first before we can merge the two classes.

distantnative:
@lukasbestle Since you probably know the locale setup best of us, would you want to give it a try cleaning that up?

lukasbestle:
I think we need a brainstorming session in the team first to find out what the desired result should be. I will think about it and make an initial suggestion.

PR #3288, opened 2021-05-02
Deprecate `Kirby\Toolkit\Facade`

distantnative:
## Status: needs discussion 🗣

To further simplify our code base I would suggest to deprecate and remove `Kirby\Toolkit\Facade`:
- Currently 3 classes are built on top of `Facade`
  - `R` which provides access to `Kirby\Http\Request`, though this class has no static methods
  - `S` which provides access to `Kirby\Session\Session`, which only has a single static method `::timeToTimestamp()`
  - `Visitor` which provides access to `Kirby\Http\Visitor`, though this class has no static methods
- Seems to me this is a v2 relict which hasn't proper use anymore

The current commit removes them directly, but writing this PR of course it's clear to me that we first should mark these as deprecated with `deprecated()` warnings and then remove in 3.7.0.

But I first wanted to open this up for discussion - if we all agree, I would prepare the PR for this first step and mark this PR here for 3.7.0.


bastianallgeier:
I think it's a bit confusing what the Facades really do, but they are not just useful for static methods. They are basically providing shorthands for any public instance method. 

I.e. `S::get()` instead of `kirby()->session()->get()` or `R::method()` I'm not sure how often those are used, but I wonder if it really makes sense to deprecate the class. It's super small.

distantnative:
Ah ok, sorry my understanding was that `__callStatic` only applies to static methods. My mistake then.
The idea wasn't per se about the size of each class, but just to reduce clutter if it isn't needed. We have a lot of classes - the more abstract names and concepts float around, the more confusing they get - and if we could ditch some that wouldn't feel bad.

bastianallgeier:
Yeah, that's true. Can we ignore the facades entirely in our docs and replace any instance of such a facade method call with the proper method chain? I think this would be a really good first step to let them vanish and then we can maybe deprecate them in 3.7 or 3.8.

distantnative:
The small benefits of this PR probably don't justify the mental headspace this PR takes and docs efforts needed. We can revisit this at a later time, but for now this distracts our focus too much for something that helps us little.

PR #3289, opened 2021-05-02
Keeps `cover` option in `ModelWithContent::panelImage()` method

### LINKED ISSUE #3226: Files section and field ignore image preview settings
**Describe the bug**  
The `files` section and field, both ignore the `image.cover: false` option. 

**To Reproduce**  
1. Create a page with a files section and a files field, e.g.
```yaml
sections:

  uploads:
    type: files
    image:
      cover: false

  fields:
    type: fields
    fields:

      gallery:
        type: files
        image:
          cover: false
```
2. Upload and select some images
3. You will see, that both the section and the field have cropped images

**Expected behavior**  
The cropping or fitting of the image should follow the `image.cover` option.

This topic is important for image editors to instantly see the image orientation and aspect ratio.

**Screenshots**  
Bildschirmfoto 2021-03-30 um 09 16 28

**Kirby Version**  
3.5.3.1

**Device**  
macOS, safari

afbora:
## Describe the PR

The `cover` option in the `list` layout in has been made available again. See issue details [here](https://github.com/getkirby/kirby/issues/3226#issuecomment-830785708)

## ⚠️ Breaking Change

Since 3.2, the `cover` option (should was work) was not working in the list layout and this feature will come back with PR.

## Related issues

- Fixes #3226 

### Fixes

- `cover` option fixed for all models and list layout in section. It can now be applied.

### Breaking changes

- If the `cover` option is not defined in the sections list layout, it is disabled by default

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Panel/Model.php
+++ b/src/Panel/Model.php
@@ -139,21 +139,31 @@ abstract class Model
                 ];
 
                 // for lists
-                $settings['list'] = [
-                    'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
-                    'srcset' => $image->srcset([
-                        '1x' => [
-                            'width' => 38,
-                            'height' => 38,
-                            'crop' => 'center'
-                        ],
-                        '2x' => [
-                            'width' => 76,
-                            'height' => 76,
-                            'crop' => 'center'
-                        ],
-                    ])
-                ];
+                if (($settings['cover'] ?? false) === false) {
+                    $settings['list'] = [
+                        'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
+                        'srcset' => $image->srcset([
+                            38,
+                            76
+                        ])
+                    ];
+                } else {
+                    $settings['list'] = [
+                        'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
+                        'srcset' => $image->srcset([
+                            '1x' => [
+                                'width' => 38,
+                                'height' => 38,
+                                'crop' => 'center'
+                            ],
+                            '2x' => [
+                                'width' => 76,
+                                'height' => 76,
+                                'crop' => 'center'
+                            ],
+                        ])
+                    ];
+                }
             }
 
             unset($settings['query']);
--- a/src/Panel/Model.php
+++ b/src/Panel/Model.php
@@ -128,9 +128,12 @@ abstract class Model
             // only create srcsets for actual File objects
             if (is_a($image, 'Kirby\Cms\File') === true) {
 
+                // pixelated transparent image placeholder
+                $placeholder = 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw';
+
                 // for cards
                 $settings['cards'] = [
-                    'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
+                    'url' => $placeholder,
                     'srcset' => $image->srcset([
                         352,
                         864,
@@ -141,7 +144,7 @@ abstract class Model
                 // for lists
                 if (($settings['cover'] ?? false) === false) {
                     $settings['list'] = [
-                        'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
+                        'url' => $placeholder,
                         'srcset' => $image->srcset([
                             38,
                             76
@@ -149,7 +152,7 @@ abstract class Model
                     ];
                 } else {
                     $settings['list'] = [
-                        'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
+                        'url' => $placeholder,
                         'srcset' => $image->srcset([
                             '1x' => [
                                 'width' => 38,
--- a/tests/Panel/FileTest.php
+++ b/tests/Panel/FileTest.php
@@ -265,6 +265,7 @@ class FileTest extends TestCase
         $panel = new File($file);
 
         $hash = $file->mediaHash();
+        $imagePlaceholder = 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw';
 
         // cover disabled as default
         $this->assertSame([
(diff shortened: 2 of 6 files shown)


lukasbestle:
I don't think it's a breaking change: If a blueprint defines `cover: false`, the intended behavior is not to crop the images, so the new behavior does exactly what it should. It's a regression fix.

bastianallgeier:
I know it's technically not a breaking change, but it could be received like one. Every list that does not define `cover:true` would look different afterwards, right? That's why I tend to move it to 3.6

distantnative:
Maybe the right way would be to treat `cover: true` for lists as default?

bastianallgeier:
> Maybe the right way would be to treat `cover: true` for lists as default?

Yes and no. For the current situation it would fix the breaking change issue, but it would also be a more confusing general solution that needs additional warnings in the docs. It feels inconsistent that way.

afbora:
If we make `cover: true` default, all card layouts will be broken. That would be a huge breaking change.

distantnative:
I meant only for list layouts. But @bastianallgeier is right, forget about my comment :D

bastianallgeier:
Let's move this to 3.6. There's no need to solve this in such a point release.

afbora:
Ready for merge.

lukasbestle:
Oh, I thought `cover: true` has been the default already. Then it really is a breaking-change.

afbora:
Refactored according to the new `Panel` namespace. But it is not the same as the previous `srcset` output after rebase. It is passing on local, but failed the GitHub tests (first commit was passing on GitHub tests before rebase). The last 2 commits are related to this. (weirdly, current test fails for my local but passes for GitHub)

I checked it but could not find what exactly changed. Any ideas?

lukasbestle:
@afbora In which way did the output change in your local setup? I just ran the PHPUnit tests on my machine and all tests pass like in GitHub Actions.

tests/Panel/PageTest.php
```diff
@@ -276,6 +276,55 @@ public function testImage()
...
+
+        // cover enabled
+        $this->assertSame([
+            'ratio' => '3/2',
+            'back' => 'pattern',
+            'cover' => true,
+            'url' => '/media/pages/test/' . $hash . '/test.jpg',
+            'cards' => [
+                'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
+                'srcset' => '/media/pages/test/' . $hash . '/test.jpg 352w, /media/pages/test/' . $hash . '/test.jpg 864w, /media/pages/test/' . $hash . '/test.jpg 1408w'
```
afbora: @lukasbestle Doesn't it seem strange to you that there is no width and height information at the end of the file? I can see it at the end of the image files when I enter the panel. 

For example srcset: `/media/pages/photography/sky/1969c671f4-1617089282/dark-forest-352x.jpg 352w`

afbora:
>  In which way did the output change in your local setup? I just ran the PHPUnit tests on my machine and all tests pass like in GitHub Actions.

I don't know how it has changed either. Unit test passes local, but cannot pass GitHub tests. Either I'm making a mistake in the test or there's a problem somewhere.

afbora:
## Summarize

- Since the `cover` option has not worked since version 3.2, it was considered a `breaking change` as the team, although it was a regression ⚠️
- Implemented new `Panel` namespace
- The problem in unit tests was solved thanks to Lukas 🙏 
- Ready to review 🚀

src/Panel/Model.php
```diff
@@ -139,21 +139,31 @@ public function image($settings = null): ?array
...
+                    $settings['list'] = [
+                        'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
+                        'srcset' => $image->srcset([
+                            38,
+                            76
+                        ])
+                    ];
+                } else {
+                    $settings['list'] = [
+                        'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
```
bastianallgeier: I know it's a minor detail, but can we store the placeholder gif in a variable to not repeat it?

tests/Panel/FileTest.php
```diff
@@ -236,6 +242,63 @@ public function testImage()
...
+        $hash = $file->mediaHash();
+
+        // cover disabled as default
+        $this->assertSame([
+            'ratio' => '3/2',
+            'back' => 'pattern',
+            'cover' => false,
+            'url' => '/media/site/' . $hash . '/test.jpg',
+            'cards' => [
+                'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
```
bastianallgeier: Same here. The redundancy looks a bit scary to me 🤣

tests/Panel/PageTest.php
```diff
@@ -276,6 +282,69 @@ public function testImage()
...
+        $mediaUrl = $page->mediaUrl() . '/' . $hash;
+
+        // cover disabled as default
+        $this->assertSame([
+            'ratio' => '3/2',
+            'back' => 'pattern',
+            'cover' => false,
+            'url' => $mediaUrl . '/test.jpg',
+            'cards' => [
+                'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
```
bastianallgeier: see above

tests/Panel/SiteTest.php
```diff
@@ -38,4 +45,80 @@ public function testUrl()
...
+        $mediaUrl = $site->mediaUrl() . '/' . $hash;
+
+        // cover disabled as default
+        $this->assertSame([
+            'ratio' => '3/2',
+            'back' => 'pattern',
+            'cover' => false,
+            'url' => $mediaUrl . '/test.jpg',
+            'cards' => [
+                'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
```
bastianallgeier: and once more

tests/Panel/UserTest.php
```diff
@@ -70,6 +77,70 @@ public function testImageStringQuery()
...
+        $mediaUrl = $user->mediaUrl() . '/' . $hash;
+
+        // cover disabled as default
+        $this->assertSame([
+            'ratio' => '3/2',
+            'back' => 'pattern',
+            'cover' => false,
+            'url' => $mediaUrl . '/test.jpg',
+            'cards' => [
+                'url' => 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw',
```
bastianallgeier: +1

bastianallgeier (dismissed):
It's looking good to me except the redundant placeholder strings.

src/Panel/Model.php
afbora: @bastianallgeier You're right. How about solving it with new `->panel()->imagePlaceholder()` method for all usages?

bastianallgeier: I like that! @distantnative what do you think?

distantnative: Do we need a method for that? Wouldn't a simple variable be fine?

distantnative: Ah seeing it pops up in various files. Ok sure, but I think we could make it a protected method to not clutter the reference

afbora: If it is `protected`, we will not be able to access it from models. So let's just continue with variables, as Bastian said.

afbora:
Nothing changed in last summary: https://github.com/getkirby/kirby/pull/3289#issuecomment-841718996

Ready to merge.

bastianallgeier (approved):

PR #3290, opened 2021-05-02
Merge `Kirby\Cms\Dir` into `Kirby\Toolkit\Dir`

distantnative:
## Status: ready for review 👀

Simplifying our class structure.
I would argue it is fine to have the inventory methods in the toolkit when we flag them as Cms-specific. They are rather separated from the rest, all the rest can be used without them as before.


bastianallgeier (dismissed):

distantnative:
@bastianallgeier added a class alias to really have no breaking changes. needs another review now to be merged

afbora (approved):

PR #3292, opened 2021-05-02
Disable cache on active sessions

### LINKED ISSUE #3242: Session: Disable cache on active sessions
Kirby should automatically disable the built-in page cache if the session is accessed from the controller or template. Also Kirby should set appropriate HTTP response headers to prevent caching by intermediate caches.

I‘m thinking of a new option `session.cacheHeaders` that accepts either an array of headers or a callback function that returns such an array. If not defined, it would have a sane default [like PHP](http://www.php.net/manual/en/function.session-cache-limiter.php).

lukasbestle:
### Features

- New `$kirby->response()->cache(false)` method that can be used to completely disable Kirby's page cache from controllers or templates
- The `$kirby->response()->header()` method now accepts a new third `$lazy` param. If set to `true`, an already set header is not overridden.

### Enhancements

- All pages that rely on Kirby's session are now automatically excluded from Kirby's page cache and from the browser and intermediary caches

## Related issues

- Fixes #3242 

## Ready?

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1226,6 +1226,10 @@ class App
      */
     public function session(array $options = [])
     {
+        // never cache responses that depend on the session
+        $this->response()->cache(false);
+        $this->response()->header('Cache-Control', 'no-store', true);
+
         return $this->sessionHandler()->get($options);
     }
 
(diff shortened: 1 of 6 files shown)


bastianallgeier (approved):
This is great!

PR #3294, opened 2021-05-03
Improvements to EmailChallenge

### LINKED ISSUE #3185: EmailChallenge: Invalid default sender address for sites in subdirs
See https://forum.getkirby.com/t/email-sending-address-for-2fa-and-login-methods/21376.

The error is introduced by this line:

https://github.com/getkirby/kirby/blob/0a4a11405079b1e37adaa70db7fd621b42e2e964/src/Cms/Auth/EmailChallenge.php#L59

lukasbestle:
### Enhancements

- The email auth challenge now uses the user's language, the configured `panel.language` or site default language for the email text (in that order)

### Bug fixes

- Fixed the default sender domain for the email auth challenge if the site URL contains a path

## Related issues

- Fixes #3185
- Implements https://kirby.nolt.io/276

## Ready?

⚠️ The failing Codecov check is because the email template files can not be covered by PHPUnit as they are no functions or methods. The check should be ignored.

### DIFF
--- a/config/templates/emails/auth/login.php
+++ b/config/templates/emails/auth/login.php
@@ -5,4 +5,4 @@
  * @var string $code
  * @var int $timeout
  */
-echo I18n::template('login.email.login.body', null, compact('user', 'code', 'timeout'));
+echo I18n::template('login.email.login.body', null, compact('user', 'code', 'timeout'), $user->language());
--- a/config/templates/emails/auth/password-reset.php
+++ b/config/templates/emails/auth/password-reset.php
@@ -5,4 +5,4 @@
  * @var string $code
  * @var int $timeout
  */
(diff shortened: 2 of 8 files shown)


afbora:
@lukasbestle How about using `exclude` element for phpunit? Does it work?

````xml
<exclude>
  <directory suffix=".php">./config/templates</directory>
</exclude>

bastianallgeier (dismissed):
I love the object option for the URL method!

bastianallgeier:
@lukasbestle how can we ignore the checks? Does @afbora's solution work?

lukasbestle:
Yeah, I think that could work. 👍

@afbora Could you please push that change to this PR branch? Then we can directly see if it works.

bastianallgeier (approved):
Yay, the exclude rule works like a charm.

PR #3295, opened 2021-05-03
Update composer dependencies

### LINKED ISSUE #3280: Update composer dependencies
So that we don't forget.

bastianallgeier:


afbora:
Since we are using specific versions, it has not been upgraded to the latest versions of the repos. Will it be upgraded at 3.6?

bastianallgeier:
Sorry, my mistake. We should upgrade at least PHP Mailer. Let me check again.

distantnative:
@lukasbestle had pinned exact versions (for a good reason). Do indeed we need to go through the dependencies one by one and read there release notes to not run into breaking changes. And only when we are sure, manually raising the pinned version.

bastianallgeier:
@distantnative I just did that and fixed the releases again.

Upgraded: 

- SimpleImage 
- our composer installer
- PHPMailer
- Whoops

PHPMailer has a security release, which we should definitely upgrade to. 
The others don't mention any breaking changes and the changes overview also looks good to me.

distantnative (approved):

PR #3296, opened 2021-05-03
Merge `Cms\KirbyTag/s` into `Text\KirbyTag/s`

distantnative:
## Status: ready for review 👀

- Simplifies our code base further

## Breaking changes
- KirbyTag class cannot be set anymore via `static::$tagClass` - I don't think this would affect any users as the KirbyTag classes cannot be replaced anyways (and thus extending the KirbyTags class yourself really doesn't seem like a thing). But open to add it back if you think it's necessary

## Ready?

- [x] CI checks pass

## When merging


bastianallgeier (dismissed):

src/Text/KirbyTags.php
```diff
@@ -20,23 +21,25 @@
  */
 class KirbyTags
 {
-    protected static $tagClass = 'Kirby\Text\KirbyTag';
-
-    public static function parse(string $text = null, array $data = [], array $options = []): string
+    public static function parse(string $text = null, array $data = [], array $options = [], ?App $app = null): string
     {
+        if ($app !== null) {
```
bastianallgeier: I think we should handle the hooks in the App class in the kirbytags method instead. We already do that in the kirbytext method. I'm not sure why we went this way in the extended class in the first place.

distantnative:
@bastianallgeier now thinking again about it, if we remove `static::$tagClass` as mentioned in the breaking changes, the class cannot really be extended anymore. I think this could be ok, but if we go that way, I would actually mark the class as final.

src/Text/KirbyTags.php
distantnative: moved them

bastianallgeier (approved):
Perfect!

PR #3297, opened 2021-05-03
Don't pin Composer installer version

lukasbestle:
The Composer installer version is the only one that's not pinned on purpose. This allows us to fix bugs in the installer separately from the CMS version. In the regular CMS usage this doesn't matter because the installer is not actually used by Kirby itself.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     "ext-mbstring": "*",
     "claviska/simpleimage": "3.6.3",
     "filp/whoops": "2.12.1",
-    "getkirby/composer-installer": "1.2.1",
+    "getkirby/composer-installer": "^1.2.0",
     "laminas/laminas-escaper": "2.7.0",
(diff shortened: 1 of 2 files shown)


distantnative:
@lukasbestle learning question for me: what's the advantage over doing `^1.2.1`?

lukasbestle:
The difference is indeed not that large, so `^1.2.1` would be fine as well.

In theoretical terms: `^1.2.0` would allow a pinned `1.2.0` version in the site package that requires Kirby, whereas `^1.2.1` would cause a fatal error message as that constraint cannot be fulfilled. So the goal for maximum compatibility with these constraints is always to define the minimum release that's supported.

In practice, no one would pin the installer to `1.2.0`, so it doesn't really matter here. It's just the rule of thumb that I always use. :)

distantnative (approved):

PR #3298, opened 2021-05-03
pre-commit script for static analysis

lukasbestle:
I've been using this script for a few months now and feel it could be useful for the rest of the team. It introduces a delay into the commit process, but makes sure that we don't commit something that will fail in CI.

In any case this script is completely voluntary as it needs to be manually enabled as described in the docblock.

The script is configured to use the dry run feature of all tools, so it does not automatically fix anything. This is by design as otherwise the staged changes wouldn't match anymore. So if the commit fails, check why in Tower's output, run the appropriate tool to fix the error, stage the fix and then commit again.

### DIFF
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+
+##########
+## Pre-commit hook script for Git
+## Runs static analysis on all staged files before they are committed
+##
+## ------
+## Usage:
+## Create an executable file `.git/hooks/pre-commit` with the following contents:
+##
+##   #!/bin/sh
+##   exec scripts/pre-commit
+##
+## You may need to modify the `$PATH` variable so Git can find your tools;
+## if the script above triggers an error, use a line like this above `exec`:
+##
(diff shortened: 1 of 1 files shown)


distantnative:
I somehow only have a `.git` folder in the `sandbox` folder but not inside `kirby`. Inside `kirby` there is only a file `.git` (which then prevents creating a folder named `.git`. Is that because I use `kirby` as submodule of `sandbox`?

Cause I think this would be super helpful!

bastianallgeier (approved):

bastianallgeier:
Let's just merge this and figure out the submodule issues later 😂

lukasbestle:
> Is that because I use `kirby` as submodule of `sandbox`?

@distantnative Yes! Git stores the `.git` directories of all submodules inside the central `.git` directory of your main repo. If you open the `.git` file in the `kirby` dir, you will see the relative path of the `.git` directory of that specific submodule. Inside that directory you will find the same directory structure. Be careful to add the hook to the submodule's `.git` directory and not to the one that belongs to the main repo, otherwise it won't work.

Issue #3189, opened 2021-03-03
Prepare Vue 3 breaking changes


distantnative:
While migrating to Vue 3 still is very tricky (probably too tricky), we should start avoiding patterns that will cause difficult breaking changes for Vue 3, e.g. `$listeners`. And where possible already adapt parts of the code to migrate more easily later on.

https://v3.vuejs.org/guide/migration/introduction.html#breaking-changes

### Changes we can already prepare
- [ ] Using the app instance instead of global Vue https://v3.vuejs.org/guide/migration/global-api.html
- [ ] Deprecated `$listeners` https://v3.vuejs.org/guide/migration/listeners-removed.html
- [ ] Deprecated event bus
- [ ] Avoid using v-if and v-for on same element https://v3.vuejs.org/guide/migration/v-if-v-for.html
- [ ] New slots syntax https://v3.vuejs.org/guide/migration/slots-unification.html
- [ ] Props default factory function no longer has access to this context https://v3.vuejs.org/guide/migration/props-default-this.html
- [ ] Removed filters

### Changes when migrating to Vue 3
- Global Vue API https://v3.vuejs.org/guide/migration/global-api.html
- new `v-model` usage
- `key` attribute https://v3.vuejs.org/guide/migration/key-attribute.html
- `.native` event modifier removed https://v3.vuejs.org/guide/migration/v-on-native-modifier-removed.html#_2-x-syntax
- `$attrs` includes class and style https://v3.vuejs.org/guide/migration/attrs-includes-class-style.html
- The destroyed lifecycle option has been renamed to unmounted
- portal replaced with teleport

trych:
Is there any specific time plan, when the switch is going to happen? The change would require the rewrite of any panel plugins that deal with Vue as well, correct?

lukasbestle:
Since when does Cmd-Enter in the comment field close the issue? 🤨

distantnative:
@trych what Lukas said. I opened this issue mainly cause I realized how many breaking changes Vue 3 comes with. And that migrating in one step will probably be way too much work. So my idea is/was if we can already tackle some parts in the next weeks by refactoring how we use Vue 2.

PR #3304, opened 2021-05-05
Remove $t() from props default

distantnative:
## Describe the PR

In Vue 3, prop default factory functions are losing access to `this`. As a first step to prepare for the change, this PR removes all `this.$t()` calls from the default functions.

## Breaking changes

None

## Related issues/ideas

- Helps with https://github.com/getkirby/kirby/issues/3189

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Dialogs/RemoveDialog.vue
+++ b/panel/src/components/Dialogs/RemoveDialog.vue
@@ -5,6 +5,7 @@
 </template>
 
 <script>
+import { TranslationString } from '@/config/i18n.js'
 import DialogMixin from "@/mixins/dialog.js";
 
 export default {
@@ -15,9 +16,9 @@ export default {
       default: "trash"
     },
     submitButton: {
-      type: [String, Boolean],
-      default() {
-        return this.$t("delete");
+      type: [String, TranslationString, Boolean],
+      default() { 
+        return new TranslationString("delete")
       }
     },
     text: String,
--- a/panel/src/components/Dialogs/RemoveDialog.vue
+++ b/panel/src/components/Dialogs/RemoveDialog.vue
@@ -16,8 +16,8 @@ export default {
       default: "trash"
     },
     submitButton: {
-      type: [String, TranslationString, Boolean],
-      default() { 
+      type: [String, Boolean],
+      default() {
         return new TranslationString("delete")
       }
     },
--- a/panel/src/components/Forms/Input/TextInput.vue
+++ b/panel/src/components/Forms/Input/TextInput.vue
@@ -9,13 +9,13 @@
       minlength,
       name,
       pattern,
-      placeholder,
       required,
       spellcheck,
       type,
       value
     }"
     :dir="direction"
+    :placeholder="placeholderLabel"
     class="k-text-input"
     v-on="listeners"
   >
@@ -87,6 +87,9 @@ export default {
   computed: {
     direction() {
       return direction(this);
+    },
+    placeholderLabel() {
+      return this.placeholder ? this.placeholder.toString(this) : null;
     }
   },
   watch: {
--- a/panel/src/components/Forms/Input/TextInput.vue
+++ b/panel/src/components/Forms/Input/TextInput.vue
@@ -9,19 +9,20 @@
       minlength,
       name,
       pattern,
+      placeholder,
       required,
       spellcheck,
       type,
       value
     }"
     :dir="direction"
-    :placeholder="placeholderLabel"
     class="k-text-input"
     v-on="listeners"
   >
 </template>
 
 <script>
+import { TranslationString } from '@/config/i18n.js'
 import direction from "@/helpers/direction.js";
 
 import {
@@ -56,7 +57,7 @@ export const props = {
     maxlength: Number,
     minlength: Number,
     pattern: String,
-    placeholder: String,
+    placeholder: [String, TranslationString],
     preselect: Boolean,
     spellcheck: {
       type: [Boolean, String],
@@ -87,9 +88,6 @@ export default {
   computed: {
     direction() {
       return direction(this);
-    },
-    placeholderLabel() {
-      return this.placeholder ? this.placeholder.toString(this) : null;
     }
   },
   watch: {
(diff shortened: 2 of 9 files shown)


lukasbestle (commented):
I think the new behavior could lead to unexpected results in weird edge-cases.

Alternative idea: We could create a new type like `TranslationString` that takes the translation key. All code that interprets the prop value can then know whether it's a normal or translation string. Not sure of the implications, just throwing it into the room. :)

distantnative:
I like that idea. Let's see if Vue does too

distantnative:
Ok maybe not - cause all our props are meant to be strings. Users would usually pass strings not translation keys. Only our defaults use translation keys

lukasbestle:
But couldn't the props look something like this:

```js
placeholder: {
  type: [String, TranslationString],
  default: new TranslationString("email.placeholder")
}
```

distantnative:
And then we could use `TranslationString.toString()` to make the conversion

lukasbestle:
Yep, that sounds good. It could be `this.prop.toString(this)` so the `TranslationString` has access to `this.$t`. If `this.prop` is already a string, I think the passed object for `.toString()` is just ignored by JS.

distantnative:
@lukasbestle Implemented as discussed.

Still feels a bit clumsy in places. Mainly about passing `this` to `toString`. But not sure if we can make this better.

panel/src/components/Forms/Input/TextInput.vue
```diff
@@ -87,6 +87,9 @@ export default {
   computed: {
     direction() {
       return direction(this);
+    },
+    placeholderLabel() {
+      return this.placeholder ? this.placeholder.toString(this) : null;
```
lukasbestle: The `placeholder` prop in this component is defined as `String` only. Have I misunderstood something?

panel/src/config/i18n.js
```diff
@@ -1,9 +1,29 @@
...
+ * TranslationString prop type
+ */
+export class TranslationString {
+  constructor(key) {
+    this.key = key;
+  }
+  toString(app) {
+    return app.$t(this.key);
+  }
+}
```
lukasbestle: For maximum compatibility (e.g. `.length`):

```suggestion
export class TranslationString extends String {
  toString(app) {
    const value = this.valueOf();
    return app ? app.$t(value) : value;
  }
}
```

distantnative:
@lukasbestle I had another idea: use `window.panel.app` to avoid having to call the TranslationString differently than normal strings. Seems to work in my setup - can you confirm?

panel/src/config/i18n.js
```diff
@@ -1,9 +1,28 @@
 
 import str from "../helpers/string.js";
 
+/**
+ * TranslationString prop type
+ */
+ export class TranslationString extends String {
```
lukasbestle:
```suggestion
export class TranslationString extends String {
```

panel/src/components/Dialogs/RemoveDialog.vue
```diff
@@ -15,9 +16,9 @@ export default {
       default: "trash"
     },
     submitButton: {
-      type: [String, Boolean],
-      default() {
-        return this.$t("delete");
+      type: [String, TranslationString, Boolean],
+      default() {
```
lukasbestle:
```suggestion
      default() {
```

lukasbestle (commented):
That's even better, yes!

I think we can even get rid of the `[String, TranslationString]` type definitions for the props now. If Vue checks the types using `prop instanceof String`, it should still work with a simple `String` because our new class extends that. I haven't tested how Vue behaves, but it would make it all even simpler.

Besides that, two minor whitespace issues. Otherwise the PR is great! 😍

distantnative:
@lukasbestle that does seem to work!

Love PRs like this where we start at "works but meh" and add onto each other's ideas to end at "that's quite a neat solution".

lukasbestle (approved):

lukasbestle:
> Love PRs like this where we start at "works but meh" and add onto each other's ideas to end at "that's quite a neat solution".

💯

bastianallgeier:
Maybe I'm missing something obvious, but we still need to import the TranslationString class in every instance that needs translated props. IMHO this is the biggest issue with this rewrite and the changes in Vue 3. Components in plugins cannot simply import it and can no longer have translated props out of the box. 

I'm also not 100% sure I understand why you use the TranslationString thingy instead of importing Vue and using Vue.$t()

distantnative:
I think you're getting it a bit wrong: you only need to import TranslationString when you cannot use `this.$t` - which really only concern factory function or prop default values. Passing prop values will be still fine with `this.$t` wherever `this` is available in context. Just default factory functions won't have that anymore.

The alternative is to create a lot of computed properties that check if the prop value is empty and either pass the prop value or the default instead. We went for the TranslationString way since it doesn't create so much more markup.

Importing Vue isn't a thing anymore in Vue 3. There is no global Vue. So since this change is mainly to fix/prepare one aspect for Vue 3, of course I didn't went for another way not supported by Vue 3.

bastianallgeier:
ok, but this still means that a plugin dev cannot provide translated default values for props anymore. When we start trying to think more like plugin devs and their major problems in components, we need a fix that might work for them too. They won't be able to import TranslationString at all.

distantnative:
I don't think blowing up our code instead would be a solution for them either. It's a really annoying change of Vue 3 - one of many - that we all will have to adapt to. We could try to expose TranslationString somehow to plugins developers but that will be rather difficult I am afraid since plugins run first before all of our code.

lukasbestle:
Plugin devs have multiple alternative options:

- They can use props that are just the translation key. Often times, it's not necessary to have the flexibility of a fixed or translation string in plugins as they have much more control about their configuration.
- Or they can use a computed prop that automatically fills in a translation string default value if the prop is not set.
- Or they can design their props so that they always take a fixed string that can then be dynamically populated with a translation string from the parent component.

I agree with Nico that there's likely no good general solution that we can implement for plugin devs. We don't have the Vue 3 changes under control, so it requires that everyone adapts.

bastianallgeier:
Ok, I'm convinced :) We also have a nice grace period if plugin devs still need it.

PR #3307, opened 2021-05-08
Use email as lowercase to consistently on `User::changeEmail()` and `System::register()`

### LINKED ISSUE #3305: User account email addresses with uppercase letters cause issues
**Describe the bug**  
If I change the email address of a user account to contain uppercase letters (e. g. Test@example.com) I have observed two issues:

1. this user account will not be saved in a users field anymore
2. login to the panel with this user account isn’t possible anymore (says user account doesn’t exist)

**To Reproduce**  

For issue 1:
1. Create a page blueprint with a users field
2. Go to “Accounts” in the admin panel
3. Click on a user account
4. Change account email address to e. g. Test@example.com
5. Go to the page with the users field
6. Add the user whose email has been changed in step 4
7. Save the page
8. Reload the page
9. User disappeared from the users field

For issue 2:
1. Follow steps 1 to 4 from the previous list
2. Log out
3. Log in as the user with email Test@example.com
4. Login is not possible

**Expected behavior**  

1. The user should be saved properly and still be visible after reloading the page
2. It should be possible to log in

**Kirby Version**  

3.5.4

 - OS: macOS Catalina
 - Browser: Firefox
 - Version: 88.0.1

afbora:
## Describe the PR

Used email as lowercase to consistently on `User::changeEmail()` and `System::register()`.

### Side Note

I don't think we have a chance to test the change (registering license) in the `System::register()` method. For this reason I had to ignore coverage. Do you have an opinion on this matter?

## Related issues

- Fixes #3305 

## Ready?

### DIFF
--- a/src/Cms/System.php
+++ b/src/Cms/System.php
@@ -448,7 +448,7 @@ class System
         $response = Remote::get('https://licenses.getkirby.com/register', [
             'data' => [
                 'license' => $license,
-                'email'   => $email,
+                'email'   => Str::lower(trim($email)),
                 'domain'  => $this->indexUrl()
             ]
         ]);
--- a/src/Cms/System.php
+++ b/src/Cms/System.php
@@ -445,6 +445,7 @@ class System
             ]);
         }
 
+        // @codeCoverageIgnoreStart
         $response = Remote::get('https://licenses.getkirby.com/register', [
             'data' => [
                 'license' => $license,
@@ -474,6 +475,7 @@ class System
                 'key' => 'license.verification'
             ]);
         }
+        // @codeCoverageIgnoreEnd
 
         return true;
     }
--- a/src/Cms/UserActions.php
+++ b/src/Cms/UserActions.php
@@ -30,10 +30,7 @@ trait UserActions
      */
     public function changeEmail(string $email)
     {
-        $email = Str::lower(trim($email));
-        $email = Idn::decodeEmail($email);
-
-        return $this->commit('changeEmail', ['user' => $this, 'email' => $email], function ($user, $email) {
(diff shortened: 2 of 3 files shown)


lukasbestle (commented):
Two questions:

- If I understood it correctly, the licensing server *is* case-sensitive. If that's true, we shouldn't lowercase the license email in the core and pass it as provided. @bastianallgeier?
- @afbora I compared the `changeEmail` method to `create`. The latter doesn't seem to lowercase the email address either. Is this done at a different location in the code?

afbora:
1. Hmm yes I got it wrong I guess. While entering the license from the panel, I thought it should turn into lowercase. Anyway, this issue can be solved on the email address requested in the license server, right? Let's wait for Bastian's confirmation.

2. The user follows the following path while creating and lowercased in the `setEmail()` method.

`User::create()` > `User::factory()` > `new static()` > `Properties::setProperties()` > `User::setEmail()`

lukasbestle:
@afbora Thanks, that was really helpful for my understanding. I have moved the lowercasing to the low-level  `updateCredentials()` method to ensure that future changes to other methods don't break the behavior again.

lukasbestle:
Now we only need @bastianallgeier's opinion on how to proceed with the `System::register()` method.

bastianallgeier:
I think it would make a lot of sense to finally get rid of the case sensitivity in our registration as well. This has been an issue for so many people and the fix is simple enough. I'd go forward with the all lower case approach.

bastianallgeier (approved):

afbora:
To sure; When users purchase a license, their e-mail addresses are always converted to lowercase letters or the license register query is checked based on lowercase letters in license server, right?

bastianallgeier:
Yes, exactly.

PR #3308, opened 2021-05-09
Vue: fix deprecated slot syntax

distantnative:
## Describe the PR
https://vuejs.org/v2/guide/components-slots.html#Deprecated-Syntax

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Forms/Field/PasswordField.vue
+++ b/panel/src/components/Forms/Field/PasswordField.vue
@@ -5,7 +5,9 @@
     v-bind="$props"
     class="k-password-field"
   >
-    <slot slot="options" name="options" />
+    <template #options>
+      <slot name="options" />
+    </template>
     <k-input
       :id="_uid"
       ref="input"
--- a/panel/src/components/Forms/Field/TextField.vue
+++ b/panel/src/components/Forms/Field/TextField.vue
@@ -5,7 +5,9 @@
     v-bind="$props"
     class="k-text-field"
   >
-    <slot slot="options" name="options" />
+    <template #options>
+      <slot name="options" />
+    </template>
     <k-input
       :id="_uid"
       ref="input"
--- a/panel/src/components/Layout/Header.vue
+++ b/panel/src/components/Layout/Header.vue
@@ -9,10 +9,14 @@
       <slot v-else />
     </k-headline>
     <k-bar v-if="$slots.left || $slots.right" class="k-header-buttons">
-      <!-- @slot buttons on the left -->
-      <slot slot="left" name="left" class="k-header-left" />
-      <!-- @slot buttons on the right -->
-      <slot slot="right" name="right" class="k-header-right" />
+      <template #left>
+        <!-- @slot buttons on the left -->
+        <slot name="left" class="k-header-left" />
+      </template>
+      <template #right>
+        <!-- @slot buttons on the right -->
+        <slot name="right" class="k-header-right" />
+      </template>
     </k-bar>
 
     <k-tabs :tab="tab" :tabs="tabsWithBadges" theme="notice" />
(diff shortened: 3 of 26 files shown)


panel/src/components/Forms/Field/PasswordField.vue
```diff
@@ -5,7 +5,9 @@
     v-bind="$props"
     class="k-password-field"
   >
-    <slot slot="options" name="options" />
+    <template #options>
+      <slot name="options" />
+    </template>
```
lukasbestle, line 10: Is this correct? I haven't looked this up in the Vue docs, but I assume the `slot` attribute on the `<slot>` element just didn't do anything until now. If that's the case, we should just remove that attribute.

panel/src/components/Forms/Field/TextField.vue
```diff
@@ -5,7 +5,9 @@
     v-bind="$props"
     class="k-text-field"
   >
-    <slot slot="options" name="options" />
+    <template #options>
+      <slot name="options" />
+    </template>
```
lukasbestle, line 10: Same here.

panel/src/components/Layout/Header.vue
```diff
@@ -9,10 +9,14 @@
...
-      <!-- @slot buttons on the right -->
-      <slot slot="right" name="right" class="k-header-right" />
+      <template #left>
+        <!-- @slot buttons on the left -->
+        <slot name="left" class="k-header-left" />
+      </template>
+      <template #right>
+        <!-- @slot buttons on the right -->
+        <slot name="right" class="k-header-right" />
+      </template>
```
lukasbestle, line 19: ...and here.

panel/src/components/Misc/Draggable.vue
```diff
@@ -9,7 +9,9 @@
     v-on="listeners"
   >
     <slot />
-    <slot slot="footer" name="footer" />
+    <template #footer>
+      <slot name="footer" />
+    </template>
```
lukasbestle, line 14: ...and here.

panel/src/components/Forms/Field/PasswordField.vue
distantnative: Let's check indeed, but I assumed the old syntax was doing what it was supposed to do: just pass on its own slot to the child component's slot with the same name.

distantnative:
@lukasbestle 
Just checked again and works fine with old and new version.

I think the `Header.vue` component is the easiest to explain/understand what's happening there:
We are creating a slot called `left` so that we can pass code to the header component to be displayed in this slot position. At the same time though, we actually pass that content to our `Bar` child component which also has a slot called `left.

Before this PR, we could do this in a one liner:
```html
<slot slot="left" name="left" class="k-header-left" />
```
The first attribute tells Vue that this element is supposed to be rendered in the slot `left` position of the current context. The second prop (with the element name) tells Vue that we want to define this as new slot that can receive content from a parent.

Vue has deprecated the syntax of `slot="name"` and that's why this has to be turned into
```html
<template #left>
  <slot name="left" class="k-header-left" />
</template>
```
Outer tag tells Vue that all this is supposed to go into the slot `left` of the current context. Inner tag tells Vue that this is a new slot that can receive content from a parent.

lukasbestle:
Ah, so these `slot` arguments did do something. TBH the old syntax was really confusing, the new one is much better and more explicit.

lukasbestle (approved):

lukasbestle:
@distantnative Have you added this to the changelog already?

distantnative:
I thought about it, but actually I don't think there's anything to report here?

lukasbestle:
It's not really relevant for users, but for devs it might be interesting to see what we refactored.

PR #3309, opened 2021-05-09
Remove vuelidate

distantnative:
## Status: on hold ⏸

As a first step to refactor validations via backend calls, a clean removal of `vuelidate` is necessary. And easier to review than bundling all in one huge PR. So this PR tries to do this first step.

We might still only merge this, when the 2nd step (new implementation with backend validations) is ready as well.

- [ ] Remove `validate` helpers


distantnative:
Closing this and keeping a local copy to declutter PR. Will open it on a more complete PR once the setup is ready for it (e.g. Inertia merged).

PR #3310, opened 2021-05-09
FliesSection.vue: remove if from attribute

distantnative:
Sooner or later vite will complain about this if statement inline in the attribute.
It's not needed since the `upload` method checks this itself already.

### DIFF
--- a/panel/src/components/Sections/FilesSection.vue
+++ b/panel/src/components/Sections/FilesSection.vue
@@ -44,7 +44,7 @@
             :layout="options.layout"
             :data-invalid="isInvalid"
             icon="image"
-            @click="if (add) upload()"
+            @clik="upload"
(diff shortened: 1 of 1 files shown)


panel/src/components/Sections/FilesSection.vue
```diff
@@ -44,7 +44,7 @@
             :layout="options.layout"
             :data-invalid="isInvalid"
             icon="image"
-            @click="if (add) upload()"
+            @clik="upload"
```
lukasbestle:
```suggestion
            @click="upload"
```

distantnative: 🙈🙈🙈

lukasbestle (approved):

PR #3311, opened 2021-05-09
Vue 3

distantnative:
## Status: experimental ⚗️

This PR is less intended to be merged soon but to further explore what benefits Vue 3 would have for us and what roadblocks are preventing us from moving there if we would want.

## Roadblocks to be solved before pursuing Vue 3
- [ ] Get rid of `vuelidate` https://github.com/getkirby/kirby/pull/3309
- [ ] Switch away from `vue-router`
- [ ] Simplify `vuex` usage
- [x] Fix deprecated slot syntax https://github.com/getkirby/kirby/pull/3308
- [x] No `this` in props' default https://github.com/getkirby/kirby/pull/3304

+ all the warnings still output by the compat build

## Breaking changes to be discussed
- How would we handle plugins that are based on Vue 2? Could we ship Vue's compat version in one release? Would it show warnings to plugin developers? And then assume with the next feature release to drop to the standard Vue 3 version? This alone might define Vue 3 as a breaking change for Kirby 4.


distantnative:
This isn't close to ready - to clear up PRs, I'm closing this and will open a new PR once this seems more realistic.

PR #3312, opened 2021-05-09
Merge `Kirby\Cms\Form` into `Kirby\Form\Form`

distantnative:
## Status: ready for review 👀

- Merges `Kirby\Cms\Form` into `Kirby\Form\Form`

## Breaking changes

None. Class alias has been added.

## Ready?

- [x] CI checks pass

## When merging


src/Form/Form.php
```diff
@@ -224,6 +310,40 @@ public function isValid(): bool
...
+     * translated to be disabled when in multilang setup
+     * and not in the default language
+     *
+     * @param array $fields
+     * @param string|null $language
+     * @return array
+     */
+    protected static function prepareFieldsForLanguage(array $fields, ?string $language = null): array
+    {
+        $kirby = App::instance();
```
lukasbestle: This should use the `$lazy` argument to ensure that the class can still be used without the CMS:

```suggestion
        $kirby = App::instance(null, true);
```

src/Form/Form.php
```diff
@@ -224,6 +310,40 @@ public function isValid(): bool
...
+
+        if ($kirby->multilang() === false) {
```
lukasbestle:
```suggestion
        // only modify the fields if we have a valid Kirby multilang instance
        if (!$kirby || $kirby->multilang() === false) {
```

src/Form/Form.php
```diff
@@ -224,6 +310,40 @@ public function isValid(): bool
         return empty($this->errors()) === true;
     }
 
+    /**
+     * Updates the field props of fields that must no be
+     * translated to be disabled when in multilang setup
+     * and not in the default language
```
lukasbestle:
```suggestion
     * Disables fields in secondary languages when
     * they are configured to be untranslatable
```

bastianallgeier:
Do we really want to keep it usable outside of the Kirby app context? Even when someone uses the form class in a template, it's still connected to the App object. I think when we couple classes more to the CMS, we should really go all in. Otherwise we need to keep having those edge cases outside an app context in mind in tests and everywhere else and we are back to the same kind of complexity

distantnative:
I would agree that we should rather make the breaking changes we need to make our setup easier - that's why we are moving these classes around in the first place.

lukasbestle:
The `Form` namespace is currently published under the terms of the MIT license, so it may well be that people are using these classes for use-cases that don't involve the CMS at all.

If we force the `App` object to initialize even though it's not required for this use-case (I mean it really isn't considering the code changes in this PR and the small scope inside that new protected method), we would instantly block all such use-cases for no reason.

IMO we don't need to add additional tests for this as the behavior doesn't depend on the app instance. Also we could argue that every use outside the CMS is not our main priority and therefore not supported. If it works, great, but we don't need to make any guarantees.

src/Form/Form.php
distantnative: If we really want to make those changes, we need more inline comments. These things make understanding the class so much more complex and honestly even now that I've worked on this for a bit, I don't quite understand why this change should be/what it means (I assume it is correct, my mind just doesn't get why): why should we also make this changes to the prop values if the Kirby object doesn't exists?

lukasbestle: Well, no, it's the other way around: We make no changes to the props if we don't have a Kirby object.
I've updated the code suggestion.

distantnative: Sometimes Github's code view just fools me 🙈👍

bastianallgeier:
I think I would rather remove the MIT license to not set any expectations and further optimise our form classes for our use cases. I really see a problem here with the complexity of the forms. They are already super super hard and the form classes are quite a hot mess. One big mistake was the idea to make them available for the frontend and the backend. The result is something that is impossible to use properly in a frontend scenario without many issues and at the same time they are also a lot harder to properly use for our backend tasks. Apart from merging the namespaces, I really think that we need to refactor the entire classes sooner or later. That's why I would really prefer to take some bigger steps now already.

lukasbestle:
Making the `Form` classes easier to use for our backend use-cases would probably require a much deeper integration into the CMS. But what would we do with the existing frontend use-cases then? To be honest I don't have a good overview at the moment if there are plugins that rely on it or if anyone is actually using the classes. But if there are users, we'd need to be very careful with such backend-optimized breaking changes. In this case I'd actually prefer the other way around: keep the CMS-specific stuff separate so that we can do with it whatever we want. Alternatively we could deprecate the classes entirely for frontend use-cases and then refactor them in the next release.

It's a tough question and I'm not really qualified to assess this – you have much more experience with these classes than I have.

bastianallgeier:
You are right. If there's someone who's actually using this, it would suck. I always thought that those classes are so weird, mostly undocumented and tied to the backend that there shouldn't be any users out there, but of course it's just a guess.

distantnative:
@bastianallgeier @lukasbestle How about we make those changes as suggested by Lukas for now to reduce breaking changes. They don't add a lot more complexity right now.

But then agreed, we should refactor Form and Fields even more over the next months. And I would then also make the case for breaking more compromises we made in the past but rather hinder us.

bastianallgeier:
Sounds good to me. This way we can get it ready for 3.6 but still have options to optimise it later.

lukasbestle:
I think the question we need to answer now is what the future refactoring could look like.

If we were to create a new CMS-specific implementation, then it doesn't make sense to merge the `Cms\Form` class into `Form\Form` now. But if you see a way to refactor `Form\Form` so that it becomes both more useful for us and for frontend use-cases, then I fully agree we should merge this now with the compatibility changes I proposed.

distantnative:
I think merging makes sense no matter how we proceed - it's not viable for us long-term to maintain two classes of so many things, in this case Form.

And over time, I think we will refactor `Form\Form` in a way that makes it more useful for our CMS, less complex. we will still think about not creating unnecessary breaking changes but we will create breaking changes where needed to push our platform forward. This will mean that some developers (I think we can be quite certain that not many, but if none or some we don't know) will have to make changes. But IMO this is ok, breaking changes are necessary if we want not to stay at status quo but improve certain parts where we know better now.

distantnative:
I have rebased the PR and added the suggested changes by @lukasbestle (see last commit).

lukasbestle:
I agree, the goal should be to refactor `Form\Form` so it's both more reliable for us and also more useful for devs. Breaking changes can't be avoided and they are worth it if the quality is much improved afterwards.

lukasbestle (approved):

PR #3313, opened 2021-05-09
Refactor `Properties`

distantnative:
## Status: in progress 👷‍♂️

Currently still failing unit tests left and right. But would be great to get some first other eyes on this, especially on the `Properties` class itself.

Current assumptions
- Properties-setting has to be manually triggered in the constructor via `->setProperties($props, ['propNameA', 'propNameB'])`
- It doesn't matter whether required or optional, we will let PHP TypeErrors resulting from the set methods parameter definition handle required props that don't get passed a value
- Every prop has also a corresponding `setPropA` method - must be present, we can rely on this
- If prop data gets passed, pass it to the set method. If null is passed as prop data, don't pass it to not run into PHP errors due to the parameter definition (e.g. `string $foo = 'bar'`)

## Breaking changes

For performance reasons we made the following changes to the `Properties` class. These only affect you if you used `Properties` in a custom plugin class:

### Setting properties

- The `Properties::setProperties($data, $names)` method now requires passing an array with all property names as they are no longer autodetected.
- The `Properties::setProperties()` now expects a setter method `set*` for every passed property name.
- `default*()` methods are no longer supported. Integrate these into your custom setter methods instead.
- The error message when a required property is not set is now a PHP `ArgumentCountError`/`TypeError` based on the missing/incorrect arguments passed to the property setter.

### Getting properties

- The `Properties::propertiesToArray()` method was removed. If you relied on that in a `toArray()` method or similar, you now need to manually return the fields you need. Please note that `propertiesToArray()` also respected custom getter methods as well as `convert*ToArray()` methods, so both of these need to be integrated into your `toArray()` process.

### DIFF
--- a/src/Toolkit/Properties.php
+++ b/src/Toolkit/Properties.php
@@ -13,6 +13,12 @@ namespace Kirby\Toolkit;
  */
 trait Properties
 {
+    /**
+     * Original or default values for properties
+     * before sett methods have been called
+     *
+     * @var array
+     */
     protected $propertyData = [];
 
     /**
@@ -37,20 +43,41 @@ trait Properties
     public function hardcopy()
     {
         $clone = $this->clone();
-        $clone->toArray();
+        foreach (array_keys($this->propertyData) as $prop) {
+            $this->{$prop}();
+        }
         return $clone;
     }
 
-    protected function setProperties(array $data, array $names)
+    /**
+     * Sets the data for all named properties
+     * by calling the dedicated prop setter method
+     *
+     * @param array $data
+     * @param array $props
+     * @return $this
+     */
+    protected function setProperties(array $data, array $props)
     {
-        foreach ($names as $name) {
-            $this->propertyData[$name] = $data[$name] ?? $this->$name ?? null;
+        // loop through all the prop names
+        // that have been passed to be se
+        foreach ($props as $prop) {
 
-            if (isset($data[$name]) === true) {
-                $this->{'set' . $name}($this->propertyData[$name]);
+            // store the data value for the property;
+            // if none has been passed, use the property
+            // default and `null` as fallbacks
+            $this->propertyData[$prop] = $data[$prop] ?? $this->$prop ?? null;
+
+            // call the setter method for the prop:
+            // if a data value for the prop has been passed,
+            // pass this value to the setter method
+            if (isset($data[$prop]) === true) {
+                $this->{'set' . $prop}($data[$prop]);
             } else {
-                $this->{'set' . $name}();
+                $this->{'set' . $prop}();
             }
         }
+
+        return $this;
     }
 }
--- a/src/Toolkit/Properties.php
+++ b/src/Toolkit/Properties.php
@@ -2,9 +2,6 @@
 
 namespace Kirby\Toolkit;
 
-use Exception;
-use ReflectionMethod;
-
 /**
  * Properties
  *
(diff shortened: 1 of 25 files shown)


src/Toolkit/Properties.php
```diff
@@ -31,121 +28,25 @@ public function clone(array $props = [])
     }
 
     /**
-     * Creates a clone and fetches all
-     * lazy-loaded getters to get a full copy
-     *
+     * @deprecated Use `::clone()` instead
+     * @todo Remove in 3.7.0
```
lukasbestle: We use `hardcopy()` in the `commit()` methods of our models at the moment.
The question is: What did this method actually do? The return value of `propertiesToArray()` isn't used, so does it entirely rely on possible side-effects of getters? If so, do we use this behavior anywhere at the moment?

lukasbestle:
I've added the breaking changes to the first post. I hope I didn't forget any (except `hardcopy()`, which is still up for discussion).

src/Toolkit/Properties.php
distantnative: @bastianallgeier will need to say where this behaviour was actually used. I added another approach that might work in the same way though.

distantnative:
Currently trying to get to the bottom of errors such as

```
Kirby\Form\Fields\BlocksFieldTest::testTranslateField
TypeError: Argument 1 passed to Kirby\Cms\App::setRequest() must be of the type array or null, object given, called in /Users/nhoffmann/Sites/kirby/sandbox/kirby/src/Toolkit/Properties.php on line 75

/Users/nhoffmann/Sites/kirby/sandbox/kirby/src/Cms/App.php:1299
/Users/nhoffmann/Sites/kirby/sandbox/kirby/src/Toolkit/Properties.php:75
/Users/nhoffmann/Sites/kirby/sandbox/kirby/src/Cms/App.php:108
/Users/nhoffmann/Sites/kirby/sandbox/kirby/src/Cms/App.php:354
/Users/nhoffmann/Sites/kirby/sandbox/kirby/tests/Form/Fields/BlocksFieldTest.php:280
```

distantnative:
@lukasbestle I think you description under "Getting properties" isn't correct. `$this->propertyData` contains the prop data before the sett methods, `propertiesToArray()` used the get methods which would provide values already affected by setter and getter methods (aside the other features you mention). So developers shouldn't use `$this->propertyData` but rather manually define their `toArray()` methods, e.g. like this https://github.com/getkirby/kirby/pull/3313/files#diff-d2099b42425af7f39a2fd55d3f40d2e8e194a7be2c5ed60ad3a8c8d092028c46L105-L108

lukasbestle:
@distantnative I thought about the breaking changes in `Properties` again. IMO we should make sure that the transition for plugin devs is as smooth as possible. While the current radical approach makes a lot of sense to see which Kirby classes need to be migrated, I think we should bring back the removed features in the release version with a `deprecated()` warning.

distantnative:
@lukasbestle Shall we have a look if it's actually used by plugin developers? I'd be rather surprised since it's a very complex tool that we have explained very little.

lukasbestle:
I've used it in plugins before. Well, I'm not representative, but given that we can't know whether *someone* used it in *some* plugin (be it an internal, private one) I think we should be careful. Especially because 3.6 will already have lots of breaking changes, so if each of those breaking changes is as small and manageable as possible, the migration for our users is going to be a lot smoother.

distantnative:
Ok, let's look once we have our radical change actually working (far from it 😂) how/where we can add the old things bridging over

distantnative:
Will reopen it when working on 3.7.0

PR #3318, opened 2021-05-11
Adds missing translations for some labels in panel

### LINKED ISSUE #3316: Link overlay of writer field not being translated
**Describe the bug**  

When adding or editing a link in a writer field, the labels in its overlay are not being translated.

**To Reproduce**  
1. Set up a writer field.
2. Add a link.
3. Look at the overlay.

**Expected behavior**  

The overlay being translated to the current panel language.

**Screenshots**  

Bildschirmfoto 2021-05-10 um 18 37 49

**Kirby Version**  

3.5.5-rc.1

**Console output**  

No errors in the console.

 - OS:  macOS
 - Browser:  Chrome
 - Version:  90.0.4430.93

afbora:
## Describe the PR

Added missing translations (hard coded) for some labels in panel

## Related issues/ideas

- Fixes #3316 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -372,7 +372,7 @@
   "more": "More",
   "name": "Name",
   "next": "Next",
-  "no": "No",
+  "no": "no",
   "off": "off",
   "on": "on",
   "open": "Open",
@@ -523,5 +523,5 @@
 
   "welcome": "Welcome",
   "year": "Year",
-  "yes": "Yes"
+  "yes": "yes"
 }
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -372,9 +372,11 @@
   "more": "More",
   "name": "Name",
   "next": "Next",
+  "no": "No",
   "off": "off",
   "on": "on",
   "open": "Open",
+  "open.newWindow": "Open in new window",
   "options": "Options",
   "options.none": "No options",
 
@@ -466,7 +468,9 @@
   "toolbar.button.file.select": "Select a file",
   "toolbar.button.file.upload": "Upload a file",
   "toolbar.button.link": "Link",
+  "toolbar.button.strike": "Strike-through",
   "toolbar.button.ol": "Ordered list",
+  "toolbar.button.underline": "Underline",
   "toolbar.button.ul": "Bullet list",
 
   "translation.author": "Kirby Team",
@@ -518,5 +522,6 @@
   "view.users": "Users",
 
   "welcome": "Welcome",
-  "year": "Year"
+  "year": "Year",
+  "yes": "Yes"
 }
--- a/panel/src/components/Writer/Marks/Bold.js
+++ b/panel/src/components/Writer/Marks/Bold.js
@@ -1,11 +1,12 @@
 import Mark from "../Mark"
+import Vue from "vue";
 
 export default class Bold extends Mark {
 
   get button() {
     return {
       icon: "bold",
-      label: "Bold"
+      label: Vue.$t("toolbar.button.bold")
     };
   }
 
--- a/panel/src/components/Writer/Marks/Bold.js
+++ b/panel/src/components/Writer/Marks/Bold.js
@@ -6,6 +6,9 @@ export default class Bold extends Mark {
   get button() {
     return {
       icon: "bold",
+      /**
+       * @todo replace with `window.panel.$t()` after merging Inertia
+       */
       label: Vue.$t("toolbar.button.bold")
     };
   }
--- a/panel/src/components/Writer/Dialogs/LinkDialog.vue
+++ b/panel/src/components/Writer/Dialogs/LinkDialog.vue
@@ -25,19 +25,22 @@ export default {
     fields() {
       return {
         href: {
-          label: "URL",
+          label: this.$t("url"),
           type: "text",
           icon: "url"
         },
         title: {
-          label: "Title",
+          label: this.$t("title"),
           type: "text",
           icon: "title"
         },
         target: {
-          label: "Open in new window",
+          label: this.$t("open.newWindow"),
           type: "toggle",
-          text: ["no", "yes"]
+          text: [
+            this.$t("no"),
+            this.$t("yes")
+          ]
         }
       };
     }
--- a/panel/src/components/Writer/Marks/Code.js
+++ b/panel/src/components/Writer/Marks/Code.js
@@ -1,11 +1,12 @@
 import Mark from "../Mark";
+import Vue from "vue";
 
 export default class Code extends Mark {
 
   get button() {
     return {
       icon: "code",
-      label: "Code"
+      label: Vue.$t("toolbar.button.code")
     };
   }
 
--- a/panel/src/components/Writer/Marks/Code.js
+++ b/panel/src/components/Writer/Marks/Code.js
@@ -6,6 +6,9 @@ export default class Code extends Mark {
   get button() {
     return {
       icon: "code",
+      /**
+       * @todo replace with `window.panel.$t()` after merging Inertia
+       */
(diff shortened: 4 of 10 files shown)


i18n/translations/en.json
```diff
@@ -372,9 +372,11 @@
   "more": "More",
   "name": "Name",
   "next": "Next",
+  "no": "No",
```
lukasbestle: For consistency with the "on/off" strings:

```suggestion
  "no": "no",
```

i18n/translations/en.json
```diff
@@ -518,5 +522,6 @@
   "view.users": "Users",
 
   "welcome": "Welcome",
-  "year": "Year"
+  "year": "Year",
+  "yes": "Yes"
```
lukasbestle:
```suggestion
  "yes": "yes"
```

panel/src/components/Writer/Marks/Bold.js
```diff
@@ -1,11 +1,12 @@
 import Mark from "../Mark"
+import Vue from "vue";
```
lukasbestle, line 2: We cannot import `Vue` as this would break when upgrading to Vue 3 if I understood @distantnative correctly.

@distantnative What should we do here? Use the `TranslationString` when that PR is merged? Or is there a better alternative for this use-case?

Same for the following files.

i18n/translations/en.json
afbora: Hmm I don't think we just think of this for the `toggle` field. What if we convert into lowercase in dialog? Would different languages be a problem?

panel/src/components/Writer/Marks/Bold.js
afbora: Apart from this PR, there are also other files that import Vue for translation.

i18n/translations/en.json
lukasbestle: Yeah, I don't think we can generalize this for all languages, so IMO we need to decide on one spelling that will be most useful for us. Lowercase "no/yes" would be consistent with the existing "off/on" toggles, but you are right that we might need the other spelling in other places.

I am split on this. @bastianallgeier @distantnative?

panel/src/components/Writer/Marks/Bold.js
lukasbestle: Good point, never noticed that before. Let's see what Nico says.

distantnative: If `window.panel.app.$t()` works, that would be preferred, I think.

distantnative: Importing Vue still works in Vue 2 and we do it in many places. But we can see that it's considered as quite the anti-pattern by the Vue team and Vue 3 doesn't support global Vue anymore, only concrete instances. So we better start finding ways not to import Vue.

afbora: As a result, which one should we continue with? Can we use `TranslationString` instead `Vue` in this PR?

distantnative: Can we use `window.panel.app.$t()`?

lukasbestle: It looks like all uses are in functions that are called after the app has been initialized, so probably yes.

But wouldn't it make sense to use `TranslationString` for cleanliness and to keep the code DRY?

distantnative: I thought calling `window.panel.app.$t()` would be DRYer than importing `TranslationString` in each of these files first

lukasbestle: It's not DRYer (it repeats the same code we already use in `TranslationString`), but the complexity argument (unneeded import) convinces me. :)

distantnative: Well, whether we repeat `window.panel.app.$t(` or `TranslationString(` is probably pretty much the same level, so it's really about the additional imports in all the files. I actually was thinking, whether we should scrap TranslationString and just go with the same thing for the prop defaults. But would first need to test whether the instance is really already available there.

afbora: > If window.panel.app.$t() works, that would be preferred, I think.

Works.

I didn't know but just wondering isn't there a shorter way to this job? Like `window.t('key')`

lukasbestle: To be honest `window.panel.app.$t` feels like a hack to me. It works in most cases, but it's not clean. Ahmet's question if there is a shorter way shows that directly.

Using `TranslationString` definitely doesn't violate DRY. DRY is all about not repeating yourself, so abstracting code away so changes would only need to be made in one place. We created `TranslationString` for exactly that reason – if we ever needed to change or replace how translation strings are handled, there is exactly one place that would need to be changed.

If we used `window.panel.app.$t` everywhere, we would always need to ask ourselves "is the app even loaded at this point". That's why I think it's a hack and not a peace of mind solution that "just works"™.

distantnative: `window.panel.app.$t` has the advantage that plugin developer can use it as well - @bastianallgeier's concerns still echoes with me. One could put the logic of `TranslationString` in a new method `window.panel.$t` or so.

lukasbestle: Wouldn't we have the same problem that we wouldn't know whether the app is already loaded?

I think we can see the plugin use-case separately from our own code – here we *can* import the class.

i18n/translations/en.json
bastianallgeier: Let's keep it like it is. When we change it in the original source, they need to be changed for every other language in Transifex.

panel/src/components/Writer/Marks/Bold.js
bastianallgeier: The problem with translations not being loaded would completely vanish with the switch to Inertia. Another reason to move that forward.

i18n/translations/en.json
lukasbestle: @bastianallgeier `no` and `yes` are new translation strings.

panel/src/components/Writer/Marks/Bold.js
bastianallgeier: Does anything speak against keeping importing Vue for this PR and fix that later, once we have Inertia running? I think it does not make sense to keep this blocked just because of that, right?

distantnative: Let's wait for Inertia - I think that one will solve it quite nicely

lukasbestle: I think it makes sense then to already merge this PR to avoid merge conflicts with Inertia later.
Only the discussion above is still open.

i18n/translations/en.json
bastianallgeier: Oh, I thought we already had them. I only saw the diff. If on/off is lowercase, I would also go for lowercase. There are often no super elegant solutions with the translation strings that work everywhere.

panel/src/components/Writer/Marks/Bold.js
bastianallgeier: Just replied to the answer above. 

Ok, what about merging it right now and already adding a todo to the inertia PR? We have way too many open PRs, so being able to close a PR feels like a good choice to me :)

distantnative: I think we should add TODO tags then to all the places, otherwise it's too easy to forget about things after Inertia.

afbora: `@todo` added for all `Vue.$t()` in this PR

bastianallgeier (approved):

PR #3326, opened 2021-05-12
Clean up Html::video() changes #3074

lukasbestle:
## Describe the PR

The new video implementation is really awesome! Especially how simple it has become and that it's the same in the backend and frontend. 😍

I mainly made two fixes to the tests that I find really important in general @bastianallgeier:

- All new tests should have `@covers` marks and especially existing marks shouldn't be removed. The more tests we clean up this way, the easier it will become to find really untested code in the end.
- I think we should stay with the pattern that all test methods are named like `test<Method><Suffix>` (so the method name should always be first and written exactly like in the tested class). The reason for this is that this will allow us to sort the methods by name using PHP-CS-Fixer once all test classes are cleaned up that way. Many test classes do not use this pattern at the moment, but I think we should at least start using it in the test classes that are already refactored (like `Html`). These classes can be detected by their general use of `@covers`.

## Breaking changes

None

## Ready?

- [x] CI checks pass

## When merging

**Not** needed in the changelog (refactoring before the release)

### DIFF
--- a/src/Toolkit/Html.php
+++ b/src/Toolkit/Html.php
@@ -423,18 +423,17 @@ class Html extends Xml
      */
     public static function video(string $url, array $options = [], array $attr = []): ?string
     {
-
         // YouTube video
-        if (preg_match('!youtu!i', $url) === 1) {
+        if (Str::contains($url, 'youtu', true) === true) {
             return static::youtube($url, $options['youtube'] ?? [], $attr);
         }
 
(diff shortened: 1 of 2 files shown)


distantnative (approved):

afbora (approved):

bastianallgeier (approved):
Thank you for the clean up. That's great!

Issue #3216, opened 2021-03-18
[Panel] Browser back button updates URL, but not view


renestalder:
**Describe the bug**  
I use the browser back button quite often. And I noticed that every time I want to go back to the parent page by using the browser back, the URL in the panel changes, but not the view.

**To Reproduce**  

I can reproduce it on at least two Kirby CMS installs

1. Go to the panel
2. Click on a page
3. Hit the browser back button

**Expected behavior**  
The Browser correctly navigates to the last page in the browsing history.

**Screenshots**  
–

**Kirby Version**  
3.5.3.1

**Console output**  
There is no output in the console

**Desktop (please complete the following information):**  
 - OS: Ubuntu
 - Browser: Firefox
 - Version: 86.0.1

**Smartphone (please complete the following information):**  
Not tested

**Additional context**  
Can't tell if that is always or was always the case.

afbora:
I can confirm that. Does not happen everytime. I did not create an issue because I could not find the exact conditions under which this error occurred.

renestalder:
Yep, was the same for me. First it seemed like occasionally when changing to a page or image that was referenced in a page field, image field. But then I opened another CMS instance and could immediately reproduce it after every click.

bastianallgeier:
I wonder if this is an issue in Vue router or in our implementation. We basically just follow their examples how to use routes and I don't really know where we could even be doing something wrong here.

afbora:
I've done few tests and this issue appeared with 3.5.0. No issue for 3.4.5.

I also found the conditions for you to test, follow these steps:

1. Login to panel
2. Clear Local Storage & Session Storage & Cookies (from application tab in browser console )
3. Get back to panel, you'll redirect to login when you refresh
4. Login panel again
5. Click any page and try to go back with "browser back button"
6. See issue

lukasbestle:
In 3.5.0 we updated `vue-router` from `v3.1.6` to `v3.4.9`. I found this in their issue tracker: https://github.com/vuejs/vue-router/issues/3245

igregson:
I've noticed this, too. 

Specifically, I see it after logging into the panel. Once I do a full page refresh while being logged in (and on a panel page) the issue is gone (i.e. the back button then works as expected). This makes me wonder if the issue is connected to the redirect method used after logging in.

alancwoo:
I've written in the forum: https://forum.getkirby.com/t/panel-back-doesnt-work/21947/5

[...]

PR #3327, opened 2021-05-13
Panel navigation via Fiber

### LINKED ISSUE #3337: [3.6.0] `k-dropdown-content` background color broken
**Describe the bug**  

`k-dropdown-content` background color broken and might be related new `.k-dropdown-content[data-align=right]` class.

**To Reproduce**  
1. Go to panel
2. Click any three dots to open dropdown content
3. See error

**Screenshots**  

screen-capture-1067-Test - Mægazine-localhost

**Kirby Version**  
develop branch of 3.6.0

distantnative:
## Status: ready for review 👀

1. Add `'panel' => ['dev' => true]` to your config
2. run `npm run dev` as usual 
3. but then access `sandbox.test/panel` and not `localhost:3000`

## Benchmark

**Before PR**
```
dist/css/index.css   93.94kb / brotli: 13.85kb
dist/js/index.js     360.09kb / brotli: 65.87kb
dist/js/vendor.js    407.78kb / brotli: 110.02kb
```

**After PR**
```
dist/css/style.css   94.01kb / brotli: 13.89kb
dist/js/index.js     350.58kb / brotli: 64.19kb
```

###  Features
- Panel navigation now runs on Fiber - a new framework that allows us to handle routing, state data and much more from the backend. More exciting features and improvements built with Fiber to come.
- The Panel logic is now structured with a new feature called Panel areas. An area defines basic information e.g. for the menu entry as well as Panel routes that return dedicated Panel views or that can host other logic (e.g. our `logout` route which doesn't return a view but ends the user session and redirects to the `login` route). Our default areas are defined in `kirby/config/areas`. This feature is directly connected with the introduction of Fiber.
- New PHP plugin extension `areas` that allows defining custom Panel areas
- A new `router()` helper that takes a set of routes and request data and directly executes the matching routing action
- New `Panel` class alias for `Kirby\Panel\Panel`

### Enhancements
- The license code is now only displayed to admins, even if other users have access to the settings view in the Panel

### Refactoring and development features
- Local Panel development can now be done directly on the site domain when the [Kirby Sandbox](https://github.com/getkirby/sandbox) is used. Simply run Vite inside `kirby/panel` with `npm run dev` and open the Panel as normal.
- New Panel frontend method `this.$reload()` which refreshes all view data via Fiber. Supports partial refreshes via `this.$reload({ only: "$user" })`
- The Vue `k-app` component is now `k-inside` for creating the Panel wrapper setup when the user is logged in. Use `k-inside` - or new `k-outside` - to wrap views accordingly. 
- New `$store.dispatch("content/clear")` Vuex action
- New `k-login-alert` Vue component
- New JS helper to deep-merge objects (`helpers/object.js`)
- New `window.deprecated()` helper for the Panel frontend
- Our CI uses dynamically built production files for frontend testing now
- Upgraded `cypress` to 7.4.0
- Upgraded `vite` to 2.2.3

### Breaking changes

- Using custom Panel views (incl. routing, breadcrumbs...) in plugins has changed fundamentally. The old way of registering might still partly work but has shortcomings and will be completely removed in 3.7.0. Use Panel areas instead.
- `vue-router` has been completely removed, use Panel areas to define custom Panel routes
- Changes in the Vuex store and store modules
  - The `content` module doesn't handle content locking information anymore. This in now provided directly to the model views via Fiber.
  - The `language` module has been removed, use `this.$language` and `this.$languages` instead.
  - The `system` module has been removed, use `this.$system` instead.
  - The `translation` module has been removed, use `this.$$translation` instead.
  - The `user` module has been removed, use `this.$user` instead.
  - `$store.dispatch('breadcrumb')` has been deprecated and removed. Use Panel areas instead to define the view breadcrumb.
  - `$store.dispatch('title')` has been deprecated and removed. Use Panel areas instead to define the view title.
- Direct URLs for tabs in the Panel have changed, e.g. now `/panel/site/?tab=yourTabName`
- Direct URLs for the users view in the Panel filtered by a specific role have changed, e.g. now `/panel/users/?role=yourRole`
- All API endpoints for content locking/unlocking don't display an exception anymore when content locking isn't supported.
- The former development `panel/index.html` file was removed
- `k-browser-view` and `k-custom-view` Vue components have been removed
- The internal Panel sandbox has been removed

### Deprecations
- `GET (:all)/lock` and `GET (:all)/unlock` API endpoints have been deprecated and throw a warning in debug mode. They will be removed in 3.7.0.
- The frontend extension `panel.plugin({ view: [...] })` has been deprecated and will be removed in 3.7.0 - there are already some functionalities (e.g. menu entry) that aren't fully supported anymore. Use Panel areas instead.
- The `heartbeat` Vuex module has been deprecated and will be removed in 3.7.0

##  Related issues
- #3216 Browser back button updates URL, but not view
- #3337 `k-dropdown-content` background color broken

_* After this PR is ready to merge, issues may need to be retested._

## Ready?

- [x] CI checks pass

## When merging


config/views/login.php
```diff
@@ -0,0 +1,18 @@
...
+    return [
+        'component' => 'LoginView',
+        'props' => [
+            'methods' => array_keys($kirby->system()->loginMethods()),
+            'pending' => [
+                'email'     => null,
+                'challenge' => null
+            ]
+
+        ],
```
lukasbestle: I haven't tested it yet, but this should work:

```suggestion
    $status = $kirby->auth()->status();

    return [
        'component' => 'LoginView',
        'props' => [
            'methods' => array_keys($kirby->system()->loginMethods()),
            'pending' => [
                'email'     => $status->email(),
                'challenge' => $status->challenge()
            ]
        ],
```

bastianallgeier:
I wondered why the vendor file is so much bigger until I remembered that Inertia uses Axios. We used inertia-light in our last try: https://www.npmjs.com/package/inertia-light

distantnative:
@bastianallgeier Exactly. That's why it's on the to do list to create a new lightweight version. Inertia has had a lot of changes/updates since then so the old lightweight is sadly quite outdated

bastianallgeier:
LOL, it's even the first point in the list. I'm terrible. 🤣

bastianallgeier:
Too bad that Pedro's version is outdated.

afbora:
@distantnative How about adding the bug issues to be solved to the PR description?

distantnative:
@afbora which ones do you have in mind?

afbora:
As Lukas said, if we remove the `vue-router`, the #3216 problem will be fixed.

distantnative:
Oh I didn't see that. Yes, please feel free to add it to the post on top. We can keep that one updated collectively :)

distantnative:
@bastianallgeier can't edit the table on mobile but before I forget: the Panel routes currently use the go helper with fixed path for panel. But it could be someone else if someone used the option to set the Panel location/path

bastianallgeier:
@distantnative just added it to the table.

bastianallgeier:
@distantnative there are currently three issues that drive me insane and maybe you've already got an idea.

- A lot of the model (file, page, user) tests in the backend have a weird issue that seems to be related to the $this->panel()->url() method, but I cannot really figure it out. 
- The back button situation is just weird. For some reason Inertia does not seem to reload props on going back and it's causing all kinds of issues. This is a behaviour that seems to be exclusive to our Panel and I wonder if it's related to Vite.
- I cannot really test it though because whenever I build the dist files, I run into a third weird issue with the CSRF token not being used in Api calls and I cannot even login to test. 

I'm just leaving this here. Maybe you have ideas. Or maybe @afbora can help.

afbora:
> A lot of the model (file, page, user) tests in the backend have a weird issue that seems to be related to the $this->panel()->url() method, but I cannot really figure it out.

This is a really weird issue. Since the [`$blueprint`](https://github.com/getkirby/kirby/blob/feature/inertia/src/Cms/File.php#L45) variable is defined first during the [`setProperties()`](https://github.com/getkirby/kirby/blob/feature/inertia/src/Cms/File.php#L127) method, `filename` has not been set yet.

In order for you to understand the error better, not the solution; 
If you define the `$filename` variable in the [File class](https://github.com/getkirby/kirby/blob/feature/inertia/src/Cms/File.php) before `$blueprint`, the error will go away 😅 

The new line that caused the error:
https://github.com/getkirby/kirby/blob/feature/inertia/src/Cms/Blueprint.php#L704

bastianallgeier:
@afbora ahhh, that already helps a lot. So it's basically a dependency issue. The filename must be defined before the blueprint, but this is not happening because of the alphabetic order of the properties. Ok. I think that's fixable at least.

afbora:
> * The back button situation is just weird. For some reason Inertia does not seem to reload props on going back and it's causing all kinds of issues. This is a behaviour that seems to be exclusive to our Panel and I wonder if it's related to Vite.

I had looked at this before but could not find any clue. This part was about history mode on `vue-router`, but I couldn't see anything like that on `vite` 🤷‍♂️

bastianallgeier:
It could still be our setup, which might have remaining parts that work with vue-router but not great with Inertia yet. I will do more tests.

bastianallgeier:
The deprecation warning about promises is a false positive in my opinion. We often use it in try/catch blocks like here: https://github.com/getkirby/kirby/blob/97edc9fd95455dab8f2b842c929b2cd142f9c962/panel/src/components/Dialogs/UserRenameDialog.vue#L49 (the success method is calling the inertia visit method under the hood) and Inertia seems to be looking for try/catch blocks as indicator for promises. It's a bit weird but I think we can ignore it. Especially when we are going to have our own library.

distantnative:
@bastianallgeier I think clearing local storage isn't fully solved yet. It works for manual logouts but not for when a session timed out.

bastianallgeier:
@distantnative that's true. I thought it's not possible. But maybe we can listen to the inertia visit event and clear storage whenever the login or logout route is visited. We don't really have any other chance as long as we are not storing changes in the session or in a file.

bastianallgeier:
@distantnative cleaning it in the login view is actually the easiest way.

bastianallgeier:
@distantnative I just wanted to work on the plugin views, but it seems like we are now using the Vue version without template compiler. Do you know how we get it up and running again?

bastianallgeier:
@distantnative ha, I found it. It's working again.

distantnative:
@bastianallgeier so for me `dist` files are working - but the back button issue remains

afbora:
Does the `/panel/account` page work for you? I only see a blank white page.

afbora:
An issue that you can reproduce when edit any page:

````
code: 400
details: []
file: "/test/kirby/inertia/kirby/src/Cms/Api.php"
key: "error.invalidArgument"
line: 142
message: "Invalid model type: kirby/inertia/panel/pages"
route: "(.*)/lock"
status: "error"
__proto__: Object
````

---

Issue Track Down:

https://github.com/getkirby/kirby/blob/feature/inertia/panel/src/components/Forms/FormButtons.vue#L89-L90

Previously Usage
````js
this.$route.path // output: /pages/home
````

Now Usage

````js
this.$view.path // output: /subfolder/panel/pages/home
````

**Comment by distantnative** [comment]
@afbora how does your URL look like that you use locally when in the Panel and running `npm run dev`?

For me it's working. I'm using `http://sandbox.test/panel`. Probably line in question is https://github.com/getkirby/kirby/blob/feature/inertia/src/Panel/Panel.php#L514

**Comment by afbora** [comment]
- Account view fixed 🎉 
- Lock issue continue (even if `content.locking` disabled)

@distantnative I've tested via `npm run build` and my local url is `http://localhost/test/kirby/inertia/`.

**Comment by afbora** [comment]
@distantnative In terms of giving you an idea; 
When I update the line you specified as below, it gives the correct `path` and locking works.
The request path contains all subfolders but the `$kirby->path()` contains only the folders after installation.

```php
$kirby->request()->url(['path' => $kirby->path()])->path()->slice(1)->toString(); // outputs: pages/home
```

https://github.com/getkirby/kirby/blob/feature/inertia/src/Panel/Panel.php#L514

distantnative:
@afbora does this work too?

```php
Str::after($kirby->path(), '/')
```

afbora:
@distantnative Yes, works great. But error is changed:

```
code: 501
details: []
exception: "Kirby\\Exception\\Exception"
file: "/test/kirby/inertia/kirby/config/api/routes/lock.php"
key: "error.lock.notImplemented"
line: 48
message: "An error occurred"
route: "(.*)/lock"
status: "error"
__proto__: Object
```

afbora:
I've notice `error.lock.notImplemented` error arise when `content.locking` disabled (even `debug` disabled)

distantnative:
The error should exist, but it actually should be caught in FormButtons.vue

afbora:
Is it normal to see this error in console even though debug is disabled?

distantnative:
@afbora doesn't sound right. how is it behaving on the `develop` branch?

afbora:
@distantnative locking works as expected with enabled or disabled on develop branch.

config/areas/settings.php
```diff
@@ -0,0 +1,35 @@
...
+            [
+                'pattern' => 'settings',
+                'action'  => function () use ($kirby) {
+                    return [
+                        'component' => 'k-settings-view',
+                        'props'     => [
+                            'languages' => $kirby->languages()->values(function ($language) {
+                                return [
+                                    'default' => $language->isDefault(),
+                                    'icon' => [
```
distantnative: @bastianallgeier Should we have that `icon` definition actually in the backend? (same with `image`). I guess that should rather be added with the Vue component view (defining it once instead adding it to each language and increasing the Inertia call load size).

bastianallgeier: That's true!

panel/src/config/fiber.js
```diff
@@ -0,0 +1,373 @@
...
+    });
+
+    // set the lang attribute according to the current translation
+    if (data.$translation) {
+      document.querySelector("html").setAttribute("lang", data.$translation.code);
+    }
+
+    // set the document title according to $view.title
+    if (data.$view.title) {
+      document.title = data.$view.title + " | Kirby Panel";
```
distantnative: So far the site title has been used here instead of the hardcoded "Kirby Panel" - I think we should bring that back.

config/api/routes/lock.php
```diff
@@ -9,7 +8,13 @@
     [
         'pattern' => '(:all)/lock',
         'method'  => 'GET',
+        /**
+         * @deprecated 3.6.0
+         * @todo Remove in 3.7.0
+         */
         'action'  => function (string $path) {
+            deprecated('The `GET (:all)/lock` API endpoint has been deprecated and will be removed in 3.7.0');
```
lukasbestle, line 16: Why the deprecation of the lock/unlock getters? These API routes might still make sense for custom applications and integrations.

config/areas/account.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
```
lukasbestle, line 2: For better performance the `Panel` class should be explicitly imported:

```suggestion

use Kirby\Panel\Panel;

```

config/areas/login.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
```
lukasbestle, line 2:
```suggestion

use Kirby\Panel\Panel;

```

config/areas/site.php
```diff
@@ -0,0 +1,60 @@
...
+            return $kirby->site()->title()->or(t('view.site'))->toString();
+        },
+        'icon'   => 'home',
+        'label'  => t('view.site'),
+        'menu'   => true,
+        'routes' => [
+            [
+                'pattern' => 'pages/(:any)',
+                'action'  => function (string $path) use ($kirby) {
+                    if (!$page = $kirby->page(str_replace('+', '/', $path))) {
```
lukasbestle:
```suggestion
                    $page = $kirby->page(str_replace('+', '/', $path));

                    if (!$page) {
```

config/areas/site.php
```diff
@@ -0,0 +1,60 @@
...
+                    return $page->panel()->route();
+                }
+            ],
+            [
+                'pattern' => 'pages/(:any)/files/(:any)',
+                'action'  => function (string $id, string $filename) use ($kirby) {
+                    $id       = str_replace('+', '/', $id);
+                    $filename = urldecode($filename);
+
+                    if (!$page = $kirby->page($id)) {
```
lukasbestle:
```suggestion
                    $page = $kirby->page($id);
                    if (!$page) {
```

config/areas/site.php
```diff
@@ -0,0 +1,60 @@
...
+                        return t('error.page.undefined');
+                    }
+
+                    if (!$file = $page->file($filename)) {
```
lukasbestle:
```suggestion
                    $file = $page->file($filename);
                    if (!$file) {
```

config/areas/site.php
```diff
@@ -0,0 +1,60 @@
...
+                'action'  => function () use ($kirby) {
+                    return $kirby->site()->panel()->route();
+                }
+            ],
+            [
+                'pattern' => 'site/files/(:any)',
+                'action'  => function (string $filename) use ($kirby) {
+                    $filename = urldecode($filename);
+
+                    if (!$file = $kirby->site()->file($filename)) {
```
lukasbestle:
```suggestion
                    $filename = urldecode($filename);
                    $file = $kirby->site()->file($filename);

                    if (!$file) {
```

config/areas/users.php
```diff
@@ -0,0 +1,91 @@
...
+                                ];
+                            },
+                        ]
+                    ];
+                }
+            ],
+            [
+                'pattern' => 'users/(:any)',
+                'action'  => function (string $id) use ($kirby) {
+                    if (!$user = $kirby->user($id)) {
```
lukasbestle:
```suggestion
                    $user = $kirby->user($id);

                    if (!$user) {
```

config/areas/users.php
```diff
@@ -0,0 +1,91 @@
...
+
+                    return $user->panel()->route();
+                }
+            ],
+            [
+                'pattern' => 'users/(:any)/files/(:any)',
+                'action'  => function (string $id, string $filename) use ($kirby) {
+                    $filename = urldecode($filename);
+
+                    if (!$user = $kirby->user($id)) {
```
lukasbestle:
```suggestion
                    $user = $kirby->user($id);
                    if (!$user) {
```

config/areas/users.php
```diff
@@ -0,0 +1,91 @@
...
+                        return t('error.user.undefined');
+                    }
+
+                    if (!$file = $user->file($filename)) {
```
lukasbestle, line 77:
```suggestion
                    $file = $user->file($filename);
                    if (!$file) {
```

config/helpers.php
```diff
@@ -622,6 +623,21 @@ function r($condition, $value, $alternative = null)
     return $condition ? $value : $alternative;
 }
 
+/**
+ * Create a micro-router
+ * and execute the routing action
+ * immediately
```
lukasbestle:
```suggestion
 * Creates a micro-router and executes
 * the routing action immediately
```

panel/src/api/request.js
```diff
@@ -1,3 +1,18 @@
+
+export async function toJson(response) {
+  const text = await response.text();
+  let data;
+
+  try {
+    data = JSON.parse(text);
+  } catch (e) {
+    window.panel.$vue.$store.dispatch("fatal", text);
+    throw new Error("The JSON response of the API could not be parsed");
```
lukasbestle: This duplicates `Vue.$api.onParserError()`. Is there any reason why we cannot reuse this method? This would reduce the risk of later regressions when we forget to change the code in one of the two places.

lukasbestle (commented):
A few suggestions and comments:

panel/src/api/request.js
distantnative: Oh true, I didn't realise that with my change (basically combining the same code from API and Fiber in this `toJSON` method), I rendered `Vue.$api.onParserError()` as dead code. It never gets called actually anymore. We could either remove it or call it from `toJSON` - a bit weird cause then it gets called for FIber request failures as well, but it wouldn't really matter.

config/api/routes/lock.php
distantnative: Our current content locking isn't great - I think we need to move rather soon a version based on storing unsaved changes not he server etc. - That will likely change a lot, which is why I would really like to discourage the use of these endpoints already by removing them.

And let's face it, any discussion here is pretty sure a theoretical one - we've never documented this properly, I don't think the content locking API has been useable at all for developers so far.

config/areas/site.php
distantnative: If we rewrite these, we should actually make the if statement consistent with our code styles, e.g. explicitly check for `$page === null` in this case then.

config/api/routes/lock.php
lukasbestle: Makes sense 👍

config/areas/site.php
lukasbestle: Being less strict makes sense here – any falsy value should lead to an error, not just `null`. But I‘m fine with either solution.

panel/src/api/request.js
lukasbestle: TBH I don‘t fully understand how it all plays together. So I‘m fine with whatever makes the most sense for you.

panel/src/config/fiber.js
```diff
@@ -0,0 +1,373 @@
...
+    url = this.toUrl(url, { data: data || {}, hash: false });
+
+    try {
+      // fetch the response (only GET request supported)
+      const response = await fetch(url, {
+        method: "get",
+        headers: {
+          ...headers,
+          "Accept": "text/html, application/xhtml+xml",
+          "X-Requested-With": "XMLHttpRequest",
```
lukasbestle: I guess these lines are from Inertia. Do we really need them for our backend logic?

panel/src/store/modules/system.js
```diff
@@ -1,76 +1,19 @@
...
-        return context.state.info;
-
-      } catch (error) {
-        context.commit("SET_INFO", {
-          isBroken: true,
-          error: error.message
-        });
-      }
+    async load() {
+      window.panel.deprecated("The $store.system module has been deprecated and removed. Use this.$reload('$sysytem') instead.");
```
lukasbestle:
```suggestion
      window.panel.deprecated("The $store.system module has been deprecated and removed. Use this.$reload('$system') instead.");
```

panel/src/store/modules/user.js
```diff
@@ -1,92 +1,39 @@
...
-      });
+    language() {
+      window.panel.deprecated("The $store.user module has been deprecated and removed. Use this.$user and this.$language instead.");
     },
-    async load(context) {
-      const user = await Vue.$api.auth.user();
-      context.commit("SET_CURRENT", user);
-      return user;
+    async load() {
+      window.panel.deprecated("The $store.user module has been deprecated and removed. Use this.$user instead.");
```
lukasbestle:
```suggestion
      window.panel.deprecated("The $store.user module has been deprecated and removed. Use this.$reload('$user') instead.");
```

panel/src/store/modules/user.js
```diff
@@ -1,92 +1,39 @@
...
-      }
-
-      try {
-        await Vue.$api.auth.logout();
-
-      } finally {
-        Vue.prototype.$go("/login");
-      }
+    async logout() {
+      window.panel.deprecated("The $store.user module has been deprecated and removed. Use this.$user instead.");
```
lukasbestle:
```suggestion
      window.panel.deprecated("The $store.user module has been deprecated and removed. Use this.$go('logout') instead.");
```

src/Cms/AppPlugins.php
```diff
@@ -129,6 +130,17 @@ protected function extendApi($api): array
         }
     }
 
+    /**
+     * Registers additional custom panel areas
```
lukasbestle:
```suggestion
     * Registers additional custom Panel areas
```

src/Panel/Model.php
```diff
@@ -195,22 +212,39 @@ protected function imageSource(?string $query = null)
...
         ) {
-            $image = null;
+            return $image;
         }
 
-        // fallback for files
-        if (
-            $image === null &&
-            is_a($this->model, 'Kirby\Cms\File') === true &&
-            $this->model->isViewable() === true
```
lukasbestle, line 208: What happened to this viewable check? Has it been moved elsewhere or is it no longer needed?

src/Panel/Model.php
```diff
@@ -245,7 +279,7 @@ public function options(array $unlock = []): array
      */
     abstract public function path(): string;
 
-    /**
+    /*
```
lukasbestle:
```suggestion
    /**
```

src/Panel/Page.php
```diff
@@ -111,4 +127,73 @@ public function pickerData(array $params = []): array
...
+     * @return array
+     */
+    public function route(): array
+    {
+        $page = $this->model;
+
+        return [
+            'breadcrumb' => function () use ($page) {
+                return $page->panel()->breadcrumb();
+            },
```
lukasbestle: Why do we need a callback here? Can't it be `$this->breadcrumb()`?

src/Panel/Redirect.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Kirby\Panel;
+
+use Exception;
+
+class Redirect extends Exception
```
lukasbestle, line 7: I feel like this class should be part of `Kirby\Exception`. But I'm not 100% sure.

src/Panel/Panel.php
```diff
@@ -45,7 +199,264 @@ public static function customCss(App $kirby)
...
+                }
+
+                return [];
+            },
+            '$permissions' => function () use ($kirby) {
+                if ($user = $kirby->user()) {
+                    return $user->role()->permissions()->toArray();
+                }
+            },
+            '$license' => $kirby->system()->license(),
```
lukasbestle: Should this be passed with every response, even if the user is not admin?

panel/src/config/fiber.js
bastianallgeier: I was wondering about it as well. I think we can remove them. Let's keep it as simple as possible.

config/areas/users.php
bastianallgeier: I really love the oneliner. Why would you change it?

src/Panel/Page.php
bastianallgeier: Yep, can just be the simple method call

src/Panel/Redirect.php
bastianallgeier: I was thinking about it as well, but we have this weird custom exception handling format in our exceptions and I wasn't sure if it fits in there well enough.

src/Panel/Panel.php
bastianallgeier: No, it shouldn't!! We should convert this to a boolean. The license key is only needed in the settings view and we can check for admin rights there before we send it.

src/Panel/Page.php
distantnative: Isn't the more we put into callbacks the better? Since then this won't processed in partial reloads that don't need this key

config/areas/users.php
lukasbestle: Setting variables in `if` blocks is [discouraged](https://phpmd.org/rules/cleancode.html#ifstatementassignment) because it can cause unexpected issues that are hard to spot. So far we mostly use it like `if ($var = $object->method())` (which is fine in the cases we use it in IMO) but with the additional `!` thrown into the mix this gets quite complex and I fear that it might break in future PHP releases. In any case this is quite hard to parse mentally.

src/Panel/Page.php
lukasbestle: Ah, if that's the case, then yes.

But we shouldn't do premature optimization. If we have a use-case that actually needs page data without the breadcrumb, then it's great, but otherwise we should IMO keep it as simple as possible.

src/Panel/Redirect.php
lukasbestle: We also have `ErrorPageException` there. But you are right, maybe it's too Panel specific.
We should rename the class to `RedirectException` though (common naming scheme for all kinds of exceptions).

src/Panel/Page.php
distantnative: The use case is any partial request that doesn't asks for the breadcrumb. With closures, we can first slim town the whole data array to only contain the keys required for the partial request before we actually resolve the closures to actual values. If we don't use closures, we always already resolve all the data before we slim down the array to just the partial data. Which would reduce the meaningfulness of partial requests quite a bit. That's why we should probably use as many closures as possible in the Panel data arrays.

src/Panel/Panel.php
```diff
@@ -45,7 +199,264 @@ public static function customCss(App $kirby)
...
+                    'locales' => $locales,
+                    'slugs'   => Str::$language,
+                    'title'   => $kirby->site()->title()->or('Kirby Panel')->toString()
+                ];
+            },
+            '$translation' => function () use ($kirby) {
+                if ($user = $kirby->user()) {
+                    $translation = $kirby->translation($user->language());
+                } else {
+                    $translation = $kirby->translation($kirby->option('panel.language', 'en'));
```
lukasbestle:
```suggestion
                    $translation = $kirby->translation($kirby->panelLanguage());
```

I think in general the current translation should be used and not dynamically determined. Line 382 above also does it manually, which duplicates the logic over and over.

src/Panel/Panel.php
```diff
@@ -89,14 +516,80 @@ public static function link(App $kirby): bool
...
+     */
+    public static function partial(App $kirby, string $component, array $data): array
+    {
+        // is it a partial request?
+        $request          = $kirby->request();
+        $include          = $request->header('X-Fiber-Include')   ?? get('_include');
+        $requestComponent = $request->header('X-Fiber-Component') ?? get('_component');
+
+        // split include string into an array of fields
+        $include = Str::split($include);
```
lukasbestle: For a better overview and to avoid future regressions:

```suggestion
        $include = Str::split($include, ',');
```

src/Panel/Panel.php
```diff
@@ -89,14 +516,80 @@ public static function link(App $kirby): bool
...
+        if (empty($requestComponent) === false && $requestComponent !== $component) {
+            return $data;
+        }
+
+        // otherwise filter data based on
+        // dot notation, e.g. `$props.tab.columns`
+        $partial = [];
+
+        // check if globals are requested and need to be merged
+        if (Str::contains(implode($include), '$')) {
```
lukasbestle: For better performance the original header/request value before splitting should be preserved in a variable. Then we don't need to implode it again here.

src/Panel/Panel.php
```diff
@@ -89,14 +516,80 @@ public static function link(App $kirby): bool
...
+
+        // if a new component is requested, the include
+        // parameter must be ignored and all data returned
+        if (empty($requestComponent) === false && $requestComponent !== $component) {
+            return $data;
+        }
+
+        // otherwise filter data based on
+        // dot notation, e.g. `$props.tab.columns`
+        $partial = [];
```
lukasbestle: This variable is overwritten in the `foreach` loop below, but until then it is not used.

Edit: Ah, it's just a typo:

```suggestion
        $partials = [];
```

src/Panel/Panel.php
```diff
@@ -89,14 +516,80 @@ public static function link(App $kirby): bool
...
+    public static function render(App $kirby, string $component, array $data)
     {
+        // get $fiber response array
+        $fiber = static::fiber($kirby, $component, $data);
+
+        // if requested, send $fiber data as JSON
+        if (static::isFiberRequest($kirby->request()) === true) {
+            return Response::json($fiber, null, get('_pretty'), [
+                'Vary'      => 'Accept',
+                'X-Fiber' => 'true'
```
lukasbestle:
```suggestion
                'Vary'    => 'Accept',
                'X-Fiber' => 'true'
```

src/Panel/Panel.php
```diff
@@ -109,32 +602,260 @@ public static function render(App $kirby)
...
+     * @return \Kirby\Http\Response|false
+     */
+    public static function router(App $kirby, string $path = null)
+    {
+        if ($kirby->option('panel') === false) {
+            return false;
+        }
+
+        // switch off pagination exceptions
+        Pagination::$validate = false;
```
lukasbestle: Why do we need this here? I'm sure there is a reason, a comment that explains it would be great (the current one is stating the obvious).

src/Panel/Panel.php
```diff
@@ -109,32 +602,260 @@ public static function render(App $kirby)
...
+            }
+        }
+
+        // redirect routes
+        $routes[] = [
+            'pattern' => [
+                '/',
+                'installation',
+                'login',
+            ],
```
lukasbestle: Is this route acting as a fallback for the defined patterns if the areas that normally define these routes are not active? So basically "if the user is already logged in, but visits `/login`, use this fallback"? That's how I understand it. If that's correct, we should add a comment that explains this behavior so we still know what this code does in the future.

src/Panel/Panel.php
```diff
@@ -109,32 +602,260 @@ public static function render(App $kirby)
...
+        $view['path']       = Str::after($kirby->path(), '/');
+        $view['props']      = $view['props'] ?? [];
+        $view['search']     = $view['search'] ?? $kirby->option('panel.search.type', 'pages');
+
+        // make sure that routes are gone
+        unset($view['routes']);
 
-        return new Response($view->render());
+        // resolve lazy props
+        return $view;
```
lukasbestle: Was `A::apply()` used here before? That's what the comment suggests. If so, we can just remove the comment.

src/Panel/Page.php
lukasbestle: Yes, I understand that.

The question is: Do we *have* partial requests that specifically don't need the breadcrumb? I think we should first assume that everything is needed (which allows for the simplest syntax without the closures) and then we should optimize the props that are not always needed.

I disagree with "we should probably use as many closures as possible": Calling all these closures takes time as well, so it also slightly reduces the performance – which is bad if it turns out that a prop is always needed in our frontend implementation.

Don't get me wrong: I'm 100% for optimizing and really like that the closures are lazy. We should just not use them everywhere without considering if they are needed.

distantnative: We do, e.g. every few seconds for content locking. And we should make partial requests work best also for cases not yet used,  but e.g. used by developers. But agreed "as many closures as possible" is a bad extreme as well 😅 If it would be a straightforward return of a value, a closure is probably worse. But given that for this call we would merge flip and merge a collection and then loop across the values - and that with a page these partial content lock requests are happening every few seconds - I think here it would be worth it to keep it wrapped in a closure.

src/Panel/Panel.php
distantnative: I believe so, must be a leftover. Let's remove it.

src/Panel/Panel.php
distantnative: You understand it correctly. Let me add a comment.

src/Panel/Model.php
distantnative: It's still here https://github.com/getkirby/kirby/blob/feature/inertia/src/Panel/File.php#L155 - more specific to the `File` object. The previous implementation was a bit redundant.

src/Panel/Panel.php
distantnative: Tbh, this originated even from the 2020 Inertia trial. I just tried without it and all pagination seems still to work. Let's remove it and if we actually encounter a reason bring it back with good documentation. Or if @bastianallgeier remembers what it was for, please yell :D

distantnative:
@lukasbestle Implemented most of your comments. If @bastianallgeier and you can decide on the exception and the if-statements, we could wrap this up fully.

config/areas/settings.php
```diff
@@ -20,6 +20,7 @@
                                     'text'    => $language->name(),
                                 ];
                             }),
+                            'license' => $kirby->user()->isAdmin() ? $kirby->system()->license() : 'Kirby 3',
```
distantnative: @bastianallgeier Isn't `$kirby->system()->license()` already handling this? Only based on the `access.settings` permission instead of admins. I think we shouldn't duplicate this logic, even worse have two different ones.

src/Panel/Page.php
lukasbestle: I agree! :)

config/areas/settings.php
lukasbestle: I think Bastian's implementation is fine. Even a user who has access to the settings view should only be able to see the actual license code if they are admin. But you are right – maybe we should change this in the method itself instead.

In any case the fallback `'Kirby 3'` value should only be printed if a license is active (otherwise non-admins would still see a valid license even though the site is not registered):

```php
$license = $kirby->system()->license();
if ($license === true) {
    $license = 'Kirby 3';
} elseif ($license === false) {
    $license = null;
}
```

This would also make the prop value more consistent so we don't need [this change](https://github.com/getkirby/kirby/pull/3327/commits/f556c65e931291049f0f6a3b6b650f708ae91a2d).

src/Panel/Redirect.php
lukasbestle: Idea: What if we move the exception to `Kirby\Exception\RedirectException` and also add support for this to our CMS router (like `ErrorPageException`)? Then it's actually generally useful.

lukasbestle: The custom exception handling code can actually be useful as we can use it to store the HTTP code of the redirect (301 or 302).

src/Cms/Api.php
```diff
@@ -81,19 +79,7 @@ public function fieldApi($model, string $name, string $path = null)
...
-            return $file;
-        }
-
-        throw new NotFoundException([
-            'key'  => 'file.notFound',
-            'data' => [
-                'filename' => $filename
-            ]
-        ]);
+        return Panel::file($path, $filename);
```
lukasbestle, line 82: To be honest I feel like this is a bit of a weird mix. The API suddenly depends on the Panel, even though it has been the other way around before and partly still is. The added methods in the `Panel` class seem generally useful, so maybe they could get a new home in the CMS namespace.

Unrelated, but still kind of related: I feel like the `Panel\Panel` class is a bit chaotic from an architectural standpoint. It only has static methods even though basically all methods rely on the `$kirby` object, some even implicitly (using the global `kirby()` helper). I think a singleton class that gets instantiated with the `$kirby` object would be a lot cleaner and more future-proof.

config/areas/settings.php
distantnative: I would solve this within `System::license()` indeed. It's just weird if in that method we look at one permission just to check for another permission when we want to use the method.

lukasbestle: Then let's change `System::license()` to use the admin property instead of the access permission. I still think it should return `true` if the user is not admin. In the area, the code I posted above can then be used to convert that `true` value into a human-readable "Kirby 3" string.

panel/src/components/Navigation/Topbar.vue
```diff
@@ -144,27 +144,18 @@
 </template>
 
 <script>
-import views from "@/config/views.js";
-
 export default {
+  props: {
+    areas: Object,
+    breadcrumb: Array,
+    license: [Boolean, String],
```
lukasbestle: This should always be `Boolean` because of Bastian's change. We can also simplify the `unregistered()` computed value or remove it and use `!this.license` directly.

src/Panel/Panel.php
```diff
@@ -45,7 +200,290 @@ public static function customCss(App $kirby)
...
+    {
+        if ($js = $kirby->option('panel.js')) {
+            $asset = asset($js);
+
+            if ($asset->exists() === true) {
+                return $asset->url() . '?' . $asset->modified();
+            }
+        }
+
+        return false;
```
lukasbestle: While we have the chance to change the return signature, I think we should change this and the `customCss()` method to return `null` in this case. Then we can type-hint the return value as `?string`.

src/Panel/Panel.php
```diff
@@ -45,7 +200,290 @@ public static function customCss(App $kirby)
...
+     *
+     * @param \Kirby\Cms\User|null $user
+     * @param string|null $areaId
+     * @return bool
+     */
+    public static function firewall(?User $user = null, ?string $areaId = null): bool
+    {
+        // a user has to be logged in
+        if ($user === null) {
+            throw new PermissionException(t('error.access.panel'));
```
lukasbestle: The translation keys are handled by the exception classes, so this should be:

```suggestion
            throw new PermissionException(['key' => 'access.panel']);
```

The same applies to other uses below.

src/Panel/Panel.php
```diff
@@ -45,7 +200,290 @@ public static function customCss(App $kirby)
...
+
+    /**
+     * Creates global data for the Panel.
+     * This will be injected in the full Panel
+     * view via the script tag.
+     *
+     * @param \Kirby\Cms\App $kirby
+     * @return array
+     */
+    public static function globals(App $kirby): array
```
lukasbestle, line 401: Why do we have both the shared data in the `data()` method and this `globals()` method?

src/Panel/Panel.php
```diff
@@ -89,14 +543,160 @@ public static function link(App $kirby): bool
...
+            'users'   => 'user',
+            'pages'   => 'page',
+            'account' => 'account'
+        ];
+
+        $modelName = $modelTypes[$modelType] ?? null;
+
+        if (Str::endsWith($modelType, '/files') === true) {
+            $modelName = 'file';
+        }
```
lukasbestle, line 595: For easier understanding:

```suggestion
        $modelType = trim(dirname($path), '/');
        if (in_array($path, ['site', 'account']) === true) {
            $modelName = $path;
        } elseif (Str::endsWith($modelType, '/files') === true) {
            $modelName = 'file';
        } else {
            $modelTypes = [
                'users' => 'user',
                'pages' => 'page'
            ];

            $modelName = $modelTypes[$modelType] ?? null;
        }
```

src/Panel/Panel.php
```diff
@@ -89,14 +543,160 @@ public static function link(App $kirby): bool
...
+        switch ($modelName) {
+            case 'site':
+                $model = $kirby->site();
+                break;
+            case 'account':
+                $model = $kirby->user(null, $kirby->option('api.allowImpersonation', false));
+                break;
+            case 'page':
+                $id    = str_replace(['+', ' '], '/', basename($path));
+                $model = $kirby->page($id);
```
lukasbestle: This kind of duplicates `Panel::page()`, but with a slightly different behavior (it also converts spaces to slashes for some reason).

src/Panel/Panel.php
```diff
@@ -89,14 +543,160 @@ public static function link(App $kirby): bool
...
+        }
+
+        $kirby = kirby();
+
+        switch ($modelName) {
+            case 'site':
+                $model = $kirby->site();
+                break;
+            case 'account':
+                $model = $kirby->user(null, $kirby->option('api.allowImpersonation', false));
```
lukasbestle:
```suggestion
                $model = static::user();
```

src/Panel/Panel.php
```diff
@@ -109,32 +709,289 @@ public static function render(App $kirby)
...
+     */
+    public static function setLanguage(App $kirby): ?string
+    {
+        // language switcher
+        if ($kirby->options('languages')) {
+            $session  = $kirby->session();
+            $language = get('language') ?? $session->get('panel.language', 'en');
+
+            // keep the language for the next visit
+            $session->set('panel.language', $language);
```
lukasbestle: For best performance, `$session->set()` should only be called if `$language` is not equals to the value that's already in the session:

```suggestion
            $sessionLanguage = $session->get('panel.language', 'en');
            $language = get('language') ?? $sessionLanguage;

            // keep the language for the next visit
            if ($language !== $sessionLanguage) {
                $session->set('panel.language', $language);
            }
```

tests/Cms/Routes/RouterTest.php
```diff
@@ -294,8 +294,8 @@ public function testDisabledPanel()
             ]
         ]);
 
-        $this->assertNull($app->call('panel'));
-        $this->assertNull($app->call('panel/something'));
+        $this->assertFalse($app->call('panel'));
+        $this->assertFalse($app->call('panel/something'));
```
lukasbestle: This comes from `Panel::router()`, right? Is there anything that speaks against returning `null` in the error case like before?

tests/Panel/FileTest.php
```diff
@@ -21,14 +23,101 @@ public function isLocked(): bool
...
 class FileTest extends TestCase
 {
+    protected $app;
+    protected $fixtures;
+
+    public function setUp(): void
+    {
+        $this->app = new App([
+            'roots' => [
+                'index' => $this->fixtures = __DIR__ . '/fixtures/FileTest',
```
lukasbestle: Temporary files in test shouldn't be stored in `/fixtures`, but in `/tmp`. This allows us to easily differentiate between files that should be in the repo (= fixtures) and files that shouldn't (= tmp).

Also applies to the other `Panel` test classes.

src/Panel/Panel.php
distantnative: To be honest, I find yours much harder to read than Bastian's 🙈

src/Panel/Panel.php
distantnative: Sounds good 👍

src/Panel/Panel.php
distantnative: Some shared data should be renewed on navigation, some globale can stay stable for the whole session (only passed on hard reload).

lukasbestle: Ah, I see. Maybe it's already enough to update the method description for `globals()` to explain this.

src/Cms/Api.php
bastianallgeier: Yeah, I was thinking about that as well. Maybe we could introduce a new Kirby\Cms\Finder class with methods like:`Finder::file()`

The architecture of the Panel class really grew a lot. There's definitely room for improvements. The singleton could be a good idea. But I wouldn't solve this in this PR.

src/Panel/Panel.php
bastianallgeier: I extended the method description

src/Panel/Panel.php
bastianallgeier: To be honest, I think both ways are not easy to read. Let's just keep it as it is.

src/Panel/Panel.php
bastianallgeier: You are totally right. I replaced it with Panel::page. I have no idea why we want to support spaces here but not in Panel::page. This is probably just some weird artefact.

tests/Cms/Routes/RouterTest.php
bastianallgeier: ✅

tests/Panel/FileTest.php
bastianallgeier: ✅

src/Cms/Api.php
lukasbestle: I've created an issue: https://github.com/getkirby/kirby/issues/3404

lukasbestle (dismissed):

lukasbestle:
Wohoo, I think this PR is finally ready for merging now! 🎉

I have created three issues for the remaining tasks, but they aren't blockers. So unless you see open tasks that need to be tackled before the merge, I think we should go ahead with the merge now to avoid further merge conflicts.

**IMPORTANT:** Please merge with GitHub's "Squash and merge" mode, otherwise we'll have 230 commits on `develop`.

lukasbestle (approved):

PR #3328, opened 2021-05-13
Update PR template

lukasbestle:
## Describe the PR

Mainly added the "Release notes" section as discussed on Discord.

⚠️ PR against `master`. I think it makes most sense to use it already for 3.6 PRs, which only works on `master`.

## Release notes

Not necessary (internal change)

## Breaking changes

None

## Ready?

- [x] CI checks pass

## When merging

- [x] ~Add changes to release notes draft in Notion~

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,38 +1,51 @@
 ## Describe the PR
-<!-- 
-A clear and concise description of the bug the PR fixes or the feature the PR introduces. 
-We may use this for the changelog and/or documentation. 
+<!--
+A clear and concise description of the bug the PR fixes or the feature the PR introduces.
+Use this section for review hints and explanations for easier code understanding if necessary.
 -->
+
+
+
+## Release notes
+<!--
+A concise list of changes to be used in the version release notes.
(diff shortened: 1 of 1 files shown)


distantnative:
Should we maybe also add a "benchmark" section where we add some performance-related info (if it makes sense for the PR), eg. asset file size for the Panel or once we have the speed tests in those oft the backend?

lukasbestle:
The backend performance tests will be automated and run in CI. Asset file sizes could make sense, but I wouldn't make such information a requirement for every PR. If we think it's relevant and needs to be discussed, we can still add such a section manually, but for most PRs and especially PRs from the community it would only take a lot of effort and time.

afbora:
I agree with Lukas. Anyway, if we think there may be a performance issue in PR, we state it in PR description.

afbora (approved):

distantnative:
Got me it's the same as the breaking changes. We could've listed this on all our PRs that affect size/performance... but have we? Rarely, I'd say.

distantnative (approved):

bastianallgeier (approved):

bastianallgeier:
There's no harm merging it on master already. That's why I merged it. I agree that we should already use it right now.

PR #3330, opened 2021-05-15
Inertia: Automatic Panel dev mode

lukasbestle:
## Describe the PR

Allows setting the `panel.dev` option permanently to `true`. The actual dev mode will only be triggered if Vite is currently running.

If `panel.dev` is not set at all, Kirby will not check for the Vite status for best production performance.

## Release notes

None (internal change to the Inertia PR)

## Breaking changes

None

## Ready?

- [x] ~CI checks pass~ (Inertia not ready)

## When merging

- [x] ~Add changes to release notes draft in Notion~

### DIFF
--- a/panel/.gitignore
+++ b/panel/.gitignore
@@ -5,6 +5,7 @@ node_modules
 .env
 .env.local
 .env.*.local
+.vite-running
 vite.config.custom.js
 
 # Log files
--- a/panel/vite.config.js
+++ b/panel/vite.config.js
@@ -1,10 +1,11 @@
-import { defineConfig } from "vite";
 import { createVuePlugin } from "vite-plugin-vue2";
 import pluginRewriteAll from "vite-plugin-rewrite-all";
 import postcssLogical from "postcss-logical";
 import postcssDirPseudoClass from "postcss-dir-pseudo-class";
 
-const path = require("path");
+const fs      = require("fs");
+const path    = require("path");
+const process = require("process");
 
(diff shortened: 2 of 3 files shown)


distantnative:
Does the file also get deleted when vite crashes/exits due to an error?

lukasbestle:
Good point, now fixed.
I also made a change so that `npm run build` doesn't create the file.

distantnative:
I think we need to adapt the if-statement on line 89 of `Panel.php` as well.
Currently, if I am not running dev but trying to access the Panel with just the build-files, I get the error

```
GET http://sandbox.test/media/panel/e87f0cefd7af496385034f7c7a96884d/src/index.js net::ERR_ABORTED 404 (Not Found)
GET http://sandbox.test/media/panel/e87f0cefd7af496385034f7c7a96884d/@vite/client net::ERR_ABORTED 404 (Not Found)
```

And maybe we should add some inline comments. Hasn't been very self-explanatory so far and with the additional changes, I think we should help our future self to understand all of this :)

lukasbestle:
👍✅

distantnative:
Does it work for you not running `npm run dev` but with build files? I always get redirects to `sandbox.test/login` (not even `/panel/login`)

lukasbestle:
That's because the current dist files are not compatible with the Inertia backend changes. I get the same on the `feature/inertia` branch.

distantnative (approved):

PR #3332, opened 2021-05-15
`Collection::values()` with mapping

distantnative:
## Describe the PR

- Adds support to `Toolkit\Collection::values()` for a mapping `Closure`
- Adds unit tests for `Collection::values()`
- Switches to `assertSame` for all unit tests for `Collection `
- Assigns `@covers` annotation to all unit tests for `Collection `

### Features
- `Kirby\Toolkit\Collection::values($map)` now accepts a mapping `Closure` as first parameter. All items are resolved with this `Closure`.

## Breaking changes

None

## Related PR

- Will be used in https://github.com/getkirby/kirby/pull/3327

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -1135,14 +1135,33 @@ class Collection extends Iterator implements Countable
     }
 
     /**
-     * Returns an non-associative array
-     * with all values
+     * Returns a non-associative array
+     * with all values. If a mapping
+     * Closure is passed, all values
+     * are processed by the Closure.
      *
+     * If pagination options are passed,
+     * the values are paginated and wrapped
+     * in an array as `data` key, with the
+     * pagination info as `pagination` key.
+     *
+     * @param Closure|null $map
+     * @param array|null $paginate
      * @return array
      */
-    public function values(): array
+    public function values(Closure $map = null, ?array $paginate = null): array
     {
-        return array_values($this->data);
+        if (empty($paginate) === false) {
+            $collection = $this->paginate($paginate);
+
+            return [
+                'data'       => $collection->values($map),
+                'pagination' => $collection->pagination()->toArray()
+            ];
+        }
+
+        $data = $map === null ? $this->data : $this->toArray($map);
+        return array_values($data);
     }
 
     /**
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -1140,27 +1140,12 @@ class Collection extends Iterator implements Countable
      * Closure is passed, all values
      * are processed by the Closure.
      *
-     * If pagination options are passed,
-     * the values are paginated and wrapped
-     * in an array as `data` key, with the
-     * pagination info as `pagination` key.
-     *
      * @param Closure|null $map
-     * @param array|null $paginate
      * @return array
      */
-    public function values(Closure $map = null, ?array $paginate = null): array
+    public function values(Closure $map = null): array
     {
-        if (empty($paginate) === false) {
-            $collection = $this->paginate($paginate);
-
-            return [
-                'data'       => $collection->values($map),
-                'pagination' => $collection->pagination()->toArray()
-            ];
-        }
-
-        $data = $map === null ? $this->data : $this->toArray($map);
+        $data = $map === null ? $this->data : array_map($map, $this->data);
         return array_values($data);
     }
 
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
I think the `$map` attribute is really useful in general (nice addition to `toArray()`, which returns the keys as well).

However I feel like `$pagination` is really specific given that the method is called `values()`. The return value is no longer an array of values, but a completely custom structure. What about a new method like `paginatedData` that can be used like this:

```php
$collection->paginate($options)->paginatedData();
```

If both are needed together:

```php
$collection->map($closure)->paginate($options)->paginatedData();
```

Advantage: This new `paginatedData` method could be used in a template/snippet/view without knowledge if the collection is paginated at all. If it isn't, the same array structure would be returned, but with the `pagination` key set to `null`.

distantnative:
I don't know - to be honest I don't think adding more and more super specialised methods will be very beneficial. It's just getting harder for people to find the right method, discover it. I feel like `->values()` is quite the right spot where someone would look for this functionality - they want the pure values of the collection, whether all or paginated.

lukasbestle:
Hm, I get your point but I disagree. I think the raw number of methods is not a problem, more important is that the methods have fitting names and do what they say.

A separate method would make the behavior a lot more flexible as it can be used together with other collection methods (like the Unix philosophy: "everything should do one thing and do it well"). Your current implementation is quite fixed in its features and quite specialized for our own Panel use-case.

I'm not 100% happy with `paginatedData()` yet, but a connection to the term "pagination" makes sense in my opinion. It's a method to get the data together with pagination information.

bastianallgeier:
I love the the new map argument but I'm also torn about the pagination parameter. If we really introduce it like this, it should also be added to toArray for consistency. Not sure. The paginatedData alternative is also not super convincing. For our inertia stuff, this solution is definitely the most handy one.

distantnative:
If we really consider this more a Panel feature, I'd rather add a `Panel::paginate($collection, $options)`.

lukasbestle:
@distantnative That's an awesome and simple solution! :+1: for that.
We can still add the mapping parameter to `$collection->values()`, I think this part is really useful in general.

bastianallgeier:
Yep, that's perfect. And I agree that we should still have the $map parameter. It's super useful!

bastianallgeier:
I've removed the paginate param and also went straight for array_map instead of using toArray. I think we can already merge this and solve the Panel::paginate thingy on the Inertia branch.

bastianallgeier (dismissed):

src/Toolkit/Collection.php
```diff
@@ -1135,14 +1135,18 @@ public function toString(): string
     }
 
     /**
-     * Returns an non-associative array
-     * with all values
+     * Returns a non-associative array
+     * with all values. If a mapping
+     * Closure is passed, all values
+     * are processed by the Closure.
```
afbora:
```suggestion
     * are processed by the Closure.
     * @since 3.6.0 This also accepts callback parameter to mapping
```

Using more `@since` tags, we can read the code better in the future.

bastianallgeier: I've added it to the description.

distantnative:
@bastianallgeier i don't think putting the description inline with the since tqg will work for our Reference. IMO it needs to go into the general description

bastianallgeier:
Ahh, to bad. But then it doesn't make a lot of sense to add the since tag in the first place, right?

distantnative:
True. I think we have to live with these little inconsistencies until we have an actual version switcher on the website

bastianallgeier (approved):

lukasbestle:
@distantnative I've converted this PR to a feature. IMO enhancements are changes to the behavior of existing code, but a new argument allows completely new use-cases. It's a tiny feature, but still a feature. ;)
@bastianallgeier I've also added this to the release notes in Notion and updated the PR title and first post based on the changes. I know the new process takes some getting used to, so don't worry. :)

distantnative:
@lukasbestle Ok with me, but not sure if we have enhancements at all then cause what you're describing as enhancement would rather get our refactoring label, wouldn't it?

lukasbestle:
@distantnative It certainly depends on how we define it. For me a refactoring is purely internal without any behavior changes whatsoever. E.g. https://github.com/getkirby/kirby/issues/3343 is a good example for an enhancement.

PR #3334, opened 2021-05-15
Fix deprecated Panel method calls

distantnative:


src/Cms/Picker.php
```diff
@@ -98,7 +98,7 @@ public function itemsToArray($items = null): array
             if (empty($this->options['map']) === false) {
                 $result[] = $this->options['map']($item);
             } else {
-                $result[] = $item->panelPickerData([
+                $result[] = $item->panel()->pickerData([
```
lukasbestle, line 101: If I understand it correctly, `$item` can in theory be any model. If this is the case, the `Panel\Model` class should have an abstract definition of `pickerData` to avoid failures down the road.

lukasbestle (commented):
Were these just forgotten in the initial PR or how did it happen? I'm curious because depending on how it happened we might need to do additional checks and/or improve our PR process.

src/Cms/Picker.php
bastianallgeier: I agree that this would help. But I would handle that in another PR.

bastianallgeier (approved):

bastianallgeier:
> Were these just forgotten in the initial PR or how did it happen? I'm curious because depending on how it happened we might need to do additional checks and/or improve our PR process.

I think they were forgotten because the old methods still work until they are officially deprecated.

src/Cms/Picker.php
lukasbestle: I've opened an issue here so that we don't forget about it: https://github.com/getkirby/kirby/issues/3340

Issue #3347, opened 2021-05-22
[3.6.0] Xdebug has detected a possible infinite loop


distantnative:
> Xdebug has detected a possible infinite loop, and aborted your script with a stack depth of '256' frames

When uploading an avatar in the Panel.
Likely due to `Filesystem` changes

lukasbestle:
I can reproduce this. The error doesn't happen on upload directly, but once a user has an avatar. I could track the issue down to `Cms\FileVersion::__call()`.

lukasbestle:
Ah, I was able to track down why: `FileFoundation` had an `exists()` method that was removed in `IsFile`. So [this line](https://github.com/getkirby/kirby/blob/2970a28/src/Cms/FileVersion.php#L40) now triggers a call to `__call()` again, which results in an infinite loop.

lukasbestle:
✅

PR #3349, opened 2021-05-22
Fix infinite loop: FileVersion asset methods #3347

distantnative:
## Describe the PR

Fixes regression introduced by the `Filesystem` PR. Infinite loop would appear e.g. after uploading an avatar. This line https://github.com/getkirby/kirby/blob/2970a28/src/Cms/FileVersion.php#L40 would cause the invite loop to `FileVersion::__call()`. Adding `IsFile::exists()` will prevent this.

## Related issues/ideas

- Fixes https://github.com/getkirby/kirby/issues/3347

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Filesystem/IsFile.php
+++ b/src/Filesystem/IsFile.php
@@ -103,6 +103,17 @@ trait IsFile
     }
 
     /**
+     * Checks if the file exists on disk
+     *
+     * @return bool
+     */
+    public function exists(): bool
+    {
+        return file_exists($this->root()) === true;
+    }
+
+
+    /**
      * Returns the app instance
      *
      * @return \Kirby\Cms\App
--- a/tests/Cms/Files/FileRulesTest.php
+++ b/tests/Cms/Files/FileRulesTest.php
@@ -91,7 +91,7 @@ class FileRulesTest extends TestCase
     {
         $file = $this->createMock(File::class);
         $file->method('filename')->willReturn('test.jpg');
-        $file->method('__call')->with('exists')->willReturn(true);
+        $file->method('exists')->willReturn(true);
 
         $this->expectException('Kirby\Exception\DuplicateException');
         $this->expectExceptionMessage('The file exists and cannot be overwritten');
--- a/tests/Cms/Files/FileVersionTest.php
+++ b/tests/Cms/Files/FileVersionTest.php
@@ -15,6 +15,11 @@ class FileVersionTest extends TestCase
         ]);
     }
 
+    public function tearDown(): void
+    {
+        Dir::remove(__DIR__ . '/tmp');
+    }
+
     public function testConstruct()
     {
         $original = new File([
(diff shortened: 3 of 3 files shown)


lukasbestle (commented):
We could alternatively fix this with `$method !== 'exists' && $this->exists() === false` in https://github.com/getkirby/kirby/blob/2970a28/src/Cms/FileVersion.php#L40. Then we wouldn't need to add another method. Both have advantages and disadvantages. What do you think?

distantnative:
I think I find adding the method a bit cleaner. The other one seems a bit like a hack and probably a bit confusing to read down the road.

lukasbestle:
OK, but I think there should be a test for the infinite loop then to avoid that we remove the method later by accident because we see it as an unneeded duplicate.

distantnative:
So I would've thought this is a proper test:

```php
    public function testExists()
    {
        $original = new File([
            'filename' => 'test.jpg',
            'root'     => __DIR__ . '/fixtures/files/test.jpg'
        ]);

        $version = new FileVersion([
            'original' => $original,
            'root'     => __DIR__ . '/fixtures/files/test-version.jpg',
            'modifications' => [
                'width' => 50,
            ]
        ]);

        $this->assertFalse($version->exists());

        // calling an asset method will trigger
        // `FileVersion::save()` if it doesn't
        // exist yet
        $this->assertSame(128, $version->width());

        $this->assertTrue($version->exists());
    }
```

But it fails with

```
Exception: File not found: kirby/tests/Cms/Files/fixtures/files/test-version.jpg
```

If I set `root` of `$version` to `__DIR__ . '/fixtures/files/test.jpg'`, the first assert fails, cause the file already exists. So I am thinking whether we have a logical problem here: We call `FileVersion::save()` in https://github.com/getkirby/kirby/blob/2970a28/src/Cms/FileVersion.php#L40 when the `root` does not exist, but `$kirby->thumb()` which is called in `FileVersion::save()` seems to require that `root` exists.

lukasbestle:
@distantnative The `Darkroom` thumb engine overwrites thumbs in place, so it expects that the file already exists. This is done inside the `thumb` component with `F::copy()`.

The issue is that `Cms\File` has a `setRoot()` method that is useless on purpose. So the `root` is actually not set, which results in a root `/dev/null/content/test.jpg`, which then means that the file cannot be copied.

This works:

```php
    public function testExists()
    {
        $page = new Page([
            'root' => __DIR__ . '/fixtures/files',
            'slug' => 'files'
        ]);

        $original = new File([
            'filename' => 'test.jpg',
            'parent'   => $page
        ]);

        $version = new FileVersion([
            'original' => $original,
            'root'     => __DIR__ . '/fixtures/files/test-version.jpg',
            'modifications' => [
                'width' => 50,
            ]
        ]);

        $this->assertFalse($version->exists());

        // calling an asset method will trigger
        // `FileVersion::save()` if it doesn't
        // exist yet
        $this->assertSame(50, $version->width());

        $this->assertTrue($version->exists());
    }
```

But please put the `test-version.jpg` file into `tmp` instead of `fixtures/files`. Then the `tearDown()` method of the test class can just `Dir::remove(__DIR__ . '/tmp')`.

distantnative:
✅

lukasbestle (approved):

PR #3350, opened 2021-05-22
Fix `FileVersion::__toString()` #3333

### LINKED ISSUE #3333: [3.6.0] Object of class Kirby\Cms\FileVersion could not be converted to string
Screen Shot 2021-05-15 at 16 23 22

distantnative:
## Describe the PR

- We were missing `IsFile::__toString()` (formerly in `FileFoundation` - https://github.com/getkirby/kirby/blob/master/src/Cms/FileFoundation.php#L64)
- Adding this method but actually making use of `File::__toString()`/`Image::__toString()`
- Changing `Image::__toString()` to return HTML to prevent the bigger breaking change (`FileVersion::__toString()` of an image only returns URL not html) although this itself is a small breaking change (`Image::__toString()` used to return the root - unclear if this ever was used)

## Discussion points

- [ ] `File::__toString()` would return the URL. The method must return a string. However, the URL is optional. Does it make sense to fall back to an empty string if no URL is set?

## Breaking changes

- `Kirby\Image\Image::__toString()` now returns the HTML `` tag instead of the root string

## Related issues/ideas

- Fixes #3333

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Cms/FileVersion.php
+++ b/src/Cms/FileVersion.php
@@ -44,8 +44,8 @@ class FileVersion
             return $this->asset()->$method(...$arguments);
         }
 
+        // content fields
         if (is_a($this->original(), 'Kirby\Cms\File') === true) {
-            // content fields
             return $this->original()->content()->get($method, $arguments);
         }
     }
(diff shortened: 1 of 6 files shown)


bastianallgeier (approved):
I think it makes sense that Image::__toString() returns the img tag. It's way more useful than the old behaviour. 

The discussion point is fine for me as well.

PR #3351, opened 2021-05-23
Merge blocks module into general Vuex store

distantnative:
## Describe the PR

Simplifying our setup a bit.
Work done on the Inertia branch that I'm trying to decouple from that PR.

## Breaking changes

- The `blocks` Vuex module has been merged into the general store. For the state, instead `this.$store.state.blocks.current` use `this.$store.state.blocks`; for the action instead `this.$store.dispatch("blocks/current", block)` use `this.$store.dispatch("block", block)`.

## Ready?

- [x] CI checks pass

## When merging


lukasbestle (commented):
Is it possible to alias the old store module until 3.7 to make the migration easier for plugin devs? We already have so many breaking changes in 3.6 that I think we need to check if each one is really necessary and unavoidable.

distantnative:
I understand your concern, but on this specific case I don't think it's necessary. This module hasn't been documented or mentioned anywhere at all. I don't think anyone besides Bastian really has ever thought about it and there is no real use for plugin developers to use it.

lukasbestle (approved):
OK, then it's fine. I think @bastianallgeier should do a final review though.

Maybe we can differentiate between internal and actual breaking changes in the release notes and/or sort the breaking changes by relevance for users and devs.

distantnative:
True. Kinda from "very relevant for many" to "only if you've been really hacking the panel internals without any available docs".

bastianallgeier:
I'm not sure about this PR to be honest. With blocks we still have a couple tasks ahead that might need more store logic. I.e. more selection stuff, copy&paste, history, etc. That's why I think the module is the better place to solve this. 

About the breaking changes. We should really have a very careful look at the breaking changes we are listing for 3.6 and maybe try to patch some more of them in the end. The more annoying this upgrade will get, the harder it will be to sell the new features and inertia setup as a positive change.

distantnative:
Ok true, convinced, let's keep it (will make sure to revert it on the Inertia PR as well).

PR #3352, opened 2021-05-23
Remove deprecated code

distantnative:
## Breaking changes

- Deprecated `Kirby\Cms\Page::sort()` has been removed. Use `Kirby\Cms\Page::changeSort()` instead.

## Deprecations

- Deprecated API endpoint `pages/(:any)/children/blueprints` is now throwing deprecation warnings. Will be removed in 3.7.0. Use `pages/(:any)/blueprints` instead.
- Deprecated API endpoint `site/children/blueprints`  is now throwing deprecation warnings. Will be removed in 3.7.0. Use `site/blueprints` instead.
- Deprecated `Kirby\Cms\App::setLocale()`  is now throwing deprecation warnings. Will be removed in 3.7.0. Use `Kirby\Toolkit\Locale::set()` instead.
- Deprecated `Kirby\Cms\Block::_key()` is now throwing deprecation warnings. Will be removed in 3.7.0. Use `Kirby\Cms\Block::type()` instead.
- Deprecated `Kirby\Cms\Block::_uid()` is now throwing deprecation warnings. Will be removed in 3.7.0. Use `Kirby\Cms\Block::id()` instead.
- Deprecated `Kirby\Toolkit\I18n::fallback()` is now throwing deprecation warnings. Will be removed in 3.7.0. Use `Kirby\Toolkit\I18n::fallbacks()[0]` instead.

### DIFF
--- a/config/api/routes/site.php
+++ b/config/api/routes/site.php
@@ -58,7 +58,7 @@ return [
         'method'  => 'GET',
         'action'  => function () {
             // @codeCoverageIgnoreStart
-            if ($this->pattern() === 'site/children/blueprints') {
+            if ($this->route->pattern() === 'site/children/blueprints') {
                 deprecated('`GET site/children/blueprints` API endpoint has been deprecated and will be removed in 3.7.0. Use `GET site/blueprints` instead.');
             }
             // @codeCoverageIgnoreEnd
--- a/config/api/routes/site.php
+++ b/config/api/routes/site.php
@@ -57,6 +57,10 @@ return [
         ],
         'method'  => 'GET',
         'action'  => function () {
+            if ($this->pattern() === 'site/children/blueprints') {
+                deprecated('`GET site/children/blueprints` API endpoint has been deprecated and will be removed in 3.7.0. Use `GET site/blueprints');
+            }
+
             return $this->site()->blueprints($this->requestQuery('section'));
         }
     ],
--- a/config/api/routes/site.php
+++ b/config/api/routes/site.php
@@ -51,16 +51,17 @@ return [
             'site/blueprints',
             /**
              * @deprecated
-             * @todo remove in 3.6.0
+             * @todo remove in 3.7.0
              */
             'site/children/blueprints',
         ],
         'method'  => 'GET',
         'action'  => function () {
+            // @codeCoverageIgnoreStart
             if ($this->pattern() === 'site/children/blueprints') {
-                deprecated('`GET site/children/blueprints` API endpoint has been deprecated and will be removed in 3.7.0. Use `GET site/blueprints');
+                deprecated('`GET site/children/blueprints` API endpoint has been deprecated and will be removed in 3.7.0. Use `GET site/blueprints` instead.');
             }
-
+            // @codeCoverageIgnoreEnd
             return $this->site()->blueprints($this->requestQuery('section'));
         }
     ],
--- a/config/api/routes/pages.php
+++ b/config/api/routes/pages.php
@@ -38,12 +38,16 @@ return [
             'pages/(:any)/blueprints',
             /**
              * @deprecated
-             * @todo remove in 3.6.0
+             * @todo remove in 3.7.0
              */
             'pages/(:any)/children/blueprints',
         ],
         'method'  => 'GET',
         'action'  => function (string $id) {
+            if ($this->pattern() === 'pages/(:any)/children/blueprints') {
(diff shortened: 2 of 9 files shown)


bastianallgeier (dismissed):

bastianallgeier:
The big question is if we want to enforce all those todos now that we already create so many breaking changes in other places. When we move the deprecation into the future, we don't let the list grow so long. But of course we are also not cleaning up as intended. @lukasbestle what do you think?

lukasbestle:
I think these deprecations aren't time-critical, so there's no real disadvantage with moving them to 3.7. Since 3.6 not only already includes many breaking changes but also very deep refactoring that may cause regressions, I agree it's better to keep as many additional breaking changes out of the release.

distantnative:
Do you really think these are critical not to include in 3.6? I agree that we need to see to have not too many breaking changes but with this attitude I feel we keep postponing changes we already decided to do, announced to do, some already have been throwing warnings with 3.6 as date to be removed. I wouldn't avoid these changes that we planned for and commuted to. If we want to have less breaking changes, we should rather talk about those new ones we are actually making.

bastianallgeier:
How about a compromise: we only deprecate those who threw a warning already? With breaking routes and other features that don't use the deprecate() method, users don't know about them so far – unless they read the changelog really carefully.

bastianallgeier:
I think we all share the attitude to keep the code base clean, but at the same time I agree with @lukasbestle that this is not time-critical.

lukasbestle:
I agree with the compromise proposed by Bastian.

distantnative:
I have made the changes to only remove the one method that already was throwing deprecation warnings. All other deprecations are now throwing warnings, so we can remove them in 3.7.0

config/api/routes/site.php
```diff
@@ -57,6 +57,10 @@
         ],
         'method'  => 'GET',
         'action'  => function () {
+            if ($this->pattern() === 'site/children/blueprints') {
+                deprecated('`GET site/children/blueprints` API endpoint has been deprecated and will be removed in 3.7.0. Use `GET site/blueprints');
```
lukasbestle:
```suggestion
                deprecated('`GET site/children/blueprints` API endpoint has been deprecated and will be removed in 3.7.0. Use `GET site/blueprints` instead');
```

The `@todo` tag above should also be changed to 3.7.0. Not really critical, but better for our overview.

lukasbestle (commented):
@distantnative Looks good to me besides one comment.

Have you checked why the Cypress E2E tests fail? One of the test cases fails with this PR for some reason.
Also we need `@ignoreCoverage` tags for three of the changes: https://app.codecov.io/gh/getkirby/kirby/compare/3352/diff#D5L60

lukasbestle (approved):
Oh, looks like this is the first time our minimal E2E tests have saved us from creating a regression. 😅
Have you checked if the `if`s in the routes do what they say in real life? If so, then this PR is ready for merging!

PS: You can ignore full methods from coverage by putting `@codeCoverageIgnore` in the docblock, then you don't need start and end tags. But it doesn't matter for this PR, we'll remove these parts of the code in 3.7 anyway.

bastianallgeier (approved):

lukasbestle:
@distantnative Open question:

> Have you checked if the `if`s in the routes do what they say in real life? If so, then this PR is ready for merging!

distantnative:
Sorry, yesterday wasn't my best day. I shouldn't have worked on this and hoped it wouldn't get merged yet. But also I should've flagged it more clearly. I think the if statement actually never gets entered, since `$this->route->pattern()` returns the URL pattern where the placeholders have already been resolved to they regex equivalents. Thinking of a better way to fix this...

distantnative:
@lukasbestle or do you think it's fine to just compare against the resolved pattern, e.g. `pages/([a-zA-Z0-9\.\-_%= \+\@\(\)]+)/children/blueprints`?

lukasbestle:
> I think the if statement actually never gets entered, since `$this->route->pattern()` returns the URL pattern where the placeholders have already been resolved to they regex equivalents.

Does it? To be honest I'm having a hard time tracking down what `$this` is referring to inside the route. If I read it correctly, it should already be the route object, but then I have no idea why the code works at all. 🤔

`$route->pattern()` should then return the original pattern as that is passed when the route is constructed, not when it's executed. I could be wrong though.

distantnative:
Not if you look at the Route code you'll see that the pattern gets turned into the regex form on Route object construction

bastianallgeier:
Argh, you are right. I totally missed this myself. The regex is completely useless for the Router in its original form. That's why it gets immediately replaced. This is indeed annoying in this situation. We could only store an additional originalPattern in the Route object. Would probably be useful for other situations as well.

Issue #3362, opened 2021-05-25
[3.6.0] Content locking more integrate into Inertia


distantnative:

distantnative:
✅

PR #3366, opened 2021-05-25
Feature/inertia content locking

distantnative:
## Describe the PR

Content locking gets fixed by providing the lock state always via Inertia prop

1. Each model in `Kirby\Panel` now passes a `lock` entry  in their props array:
  - `false` for when content locking isn't supported
  - `['state' => null]` for when no lock/unlock exists
  - `['state' => 'lock', 'data' => [...]]` for when content is locked
  - `['state' => 'unlock']` for when a content lock got unlocked
2. `FormButtons` component is reloading the `lock` prop every 10 seconds via an Inertia partial request
3. When the model has unsaved changes, `FormButtons` also sends an API request every 30 seconds to write the content lock
4. When there are no unsaved changes anymore (saved or reverted), an API request is send to remove the content lock
5. When another user is already doing edits, all field sections will be disabled and a warning bar shown
6. If the other user saves or reverts changes (no more unsaved changes on their side), the lock is lifted and the model fully reloaded incl. content that might have changed (so that user A doesn't overwrite the changes of user B when making their own edits)
7. While user B is editing, user A cannot do anything. When use B stops editing (navigating away from the model),a 60 seconds grace period starts. After that user A can unlock the content lock.
8. Clicking the Unlock button sends an API request that removes the lock and writes an unlock
9. When user B now returns, they find a warning that the content has been unlocked
10. They can download a text file with their old unsaved changes (as backup) and resolve the unlock warning (which sends an API request to remove the unlock and also reverts all unsaved changes from the content store to have a clean slate again)

**Edge cases**
- If the state changes to `lock` for user A (user B edited something), we revert all unsaved changes of user A. This can happen when both start editing pretty much at the same time and user A hasn't gotten the updated state that user B is already editing. Since we update the state every 10 seconds, this should only result in loss of very few changes.

## Performance

**Before PR**
```
dist/css/style.css   94.01kb / brotli: 13.89kb
dist/js/index.js     347.30kb / brotli: 63.53kb
dist/js/vendor.js    407.46kb / brotli: 111.32kb
```

**After PR**
```
dist/css/style.css   94.01kb / brotli: 13.89kb
dist/js/index.js     346.01kb / brotli: 63.36kb
dist/js/vendor.js    407.46kb / brotli: 111.32kb
```

## Breaking changes

- GET routes for content locking info have been removed from API
- The other content locking API routes don't throw an exception anymore when content locking isn't supported but just don't do anything
- `this.$store.state.content.status.enabled` is now `this.$store.state.content.enabled` 

## Related PR / issues

- Merges into https://github.com/getkirby/kirby/pull/3327
- Closes https://github.com/getkirby/kirby/issues/3362

### DIFF
--- a/config/api/routes/lock.php
+++ b/config/api/routes/lock.php
@@ -8,33 +8,11 @@ use Kirby\Exception\Exception;
 return [
     [
         'pattern' => '(:all)/lock',
-        'method'  => 'GET',
-        'action'  => function (string $path) {
-            if ($lock = $this->parent($path)->lock()) {
-                return [
-                    'supported' => true,
-                    'locked'    => $lock->get()
-                ];
-            }
-
-            return [
-                'supported' => false,
-                'locked'    => null
-            ];
-        }
-    ],
-    [
-        'pattern' => '(:all)/lock',
         'method'  => 'PATCH',
         'action'  => function (string $path) {
             if ($lock = $this->parent($path)->lock()) {
                 return $lock->create();
             }
-
-            throw new Exception([
-                'key'      => 'lock.notImplemented',
-                'httpCode' => 501
-            ]);
         }
     ],
     [
@@ -44,28 +22,6 @@ return [
             if ($lock = $this->parent($path)->lock()) {
                 return $lock->remove();
             }
-
-            throw new Exception([
-                'key'      => 'lock.notImplemented',
-                'httpCode' => 501
-            ]);
-        }
-    ],
-    [
-        'pattern' => '(:all)/unlock',
-        'method'  => 'GET',
-        'action'  => function (string $path) {
-            if ($lock = $this->parent($path)->lock()) {
-                return [
-                    'supported' => true,
-                    'unlocked'  => $lock->isUnlocked()
-                ];
-            }
-
-            return [
-                'supported' => false,
-                'unlocked'  => null
-            ];
         }
     ],
     [
@@ -75,11 +31,6 @@ return [
             if ($lock = $this->parent($path)->lock()) {
                 return $lock->unlock();
             }
-
-            throw new Exception([
-                'key'      => 'lock.notImplemented',
-                'httpCode' => 501
-            ]);
         }
     ],
     [
@@ -89,11 +40,6 @@ return [
             if ($lock = $this->parent($path)->lock()) {
                 return $lock->resolve();
             }
-
-            throw new Exception([
-                'key'      => 'lock.notImplemented',
-                'httpCode' => 501
-            ]);
         }
     ],
 ];
--- a/config/api/routes/lock.php
+++ b/config/api/routes/lock.php
@@ -8,6 +8,29 @@ use Kirby\Exception\Exception;
 return [
     [
         'pattern' => '(:all)/lock',
+        'method'  => 'GET',
+        /**
+         * @deprecated 3.6.0
+         * @todo Remove in 3.7.0
+         */
+        'action'  => function (string $path) {
+            deprecated('The `GET (:all)/lock` API endpoint has been deprecated and will be removed in 3.7.0');
+
+            if ($lock = $this->parent($path)->lock()) {
+                return [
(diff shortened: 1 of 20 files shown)


config/api/routes/lock.php
```diff
@@ -6,35 +6,13 @@
  * Content Lock Routes
  */
 return [
```
bastianallgeier: Let's say you want to build a mobile app that still supports content locking. This endpoint would still be needed, right?

config/api/routes/lock.php
```diff
@@ -44,28 +22,6 @@
...
             }
-
-            throw new Exception([
-                'key'      => 'lock.notImplemented',
-                'httpCode' => 501
-            ]);
-        }
-    ],
-    [
-        'pattern' => '(:all)/unlock',
```
bastianallgeier, line 55: Same as above. I wonder if we should actually deprecate those. We haven't documented them at all right now and thus we could say we deprecate them without harm, but the big question remains if we want to support locking via the API in the future or not, or if we keep this a private API.

panel/src/components/Layout/Inside.vue
```diff
@@ -66,6 +66,9 @@ export default {
     "k-registration": Registration
   },
   inheritAttrs: false,
+  props: {
```
bastianallgeier, line 69: Is this still used somewhere?

panel/src/components/Views/FileView.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-  <k-inside>
+  <k-inside :lock="lock">
```
bastianallgeier, line 2: Same here. I'm not sure I fully understand how locking is passed down.

panel/src/components/Views/PageView.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <k-inside>
+  <k-inside :lock="lock">
```
bastianallgeier, line 2: see above

panel/src/components/Views/SiteView.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <k-inside>
+  <k-inside :lock="lock">
```
bastianallgeier, line 2: +1

panel/src/components/Views/UserView.vue
```diff
@@ -1,9 +1,9 @@
 <template>
-  <k-inside>
+  <k-inside :lock="lock">
```
bastianallgeier, line 2: +1

bastianallgeier (commented):
In general, it really looks great. There are quite a few parts that I can not fully review without spending lots of time on reconstructing it in my head. But I think it all boils down to proper testing in the end.

panel/src/components/Views/FileView.vue
distantnative: The view receives the `lock` prop (defined in ModelView) from Inertia, passes it down to Inside which passes it down to FormButtons.

config/api/routes/lock.php
distantnative: We can do that. Tbh, I think this is a very theoretical discussion - I don't think anyone has ever implemented content locking for virtual pages due to the state of our code and docs. I'd also bet that there's no API use besides us here. But sure, we can be on the safe side.

distantnative:
Same with me, takes a lot of mental capacity to reconstruct the scenarios. And also quite challenging to test them all. But I don't know how we could do this better.

panel/src/components/Views/FileView.vue
bastianallgeier: Ok, that makes sense, but somehow the form button component does not receive the lock prop in inside.

distantnative:
Fixed the `lock` prop and brought back the API routes (though as deprecated with warning).

bastianallgeier:
Once you've fixed the merge conflicts, I think we should merge this asap.

PR #3369, opened 2021-05-26
Globals for Fiber

distantnative:


panel/src/fiber/app.js
```diff
@@ -14,24 +14,32 @@ export default {
     Fiber.init({
       page: window.fiber,
       component: (name) => Vue.component(name),
-      props: (props) => {
+      props: (data) => {
+
+        /** Set globals */
+        ["$config", "$system", "$translation", "$urls"].forEach(key => {
```
distantnative: Why are we setting these differently than the others?
I don't think you'll need the else-statement here. Our `protocol.js` keeps the globals from the initial request around and adds them after every partial request anyways. So this method sees always a full `data` object.

bastianallgeier: I didn't get it running without this if clause.

src/Panel/Panel.php
```diff
@@ -228,10 +228,11 @@ public static function customJs(App $kirby)
     public static function data(App $kirby, string $component, array $data = []): array
     {
         // shared default data
+        $multilang = $kirby->option('languages', false) !== false;
```
afbora, line 231: If I haven't missed anything, why don't we use the `$kirby->multilang()` method? 🤔

distantnative: I think the definition is a bit different: `$kirby->multilang()` checks if there actually has been at least one language defined. While we here are checking if the config allows defining languages - no matter if the first one has been defined or not.

afbora: Ah right 👍

PR #3370, opened 2021-05-26
Add `panel.route:before` hook

distantnative:


lukasbestle:
Is there a specific reason why there is no corresponding `after` hook?

distantnative:
I just couldn't think of any use case for that

lukasbestle:
We have `route:after` though, so I think we should also have `panel.route:after` for consistency. Our users are very creative with finding use-cases. ;)

distantnative:
@lukasbestle I have added a suggestion - would there be more/other/less parameters you would pass to it?

lukasbestle:
The `route:after` hook also provides `method` and `final`, I think these should be supported as well (if they make sense).

distantnative:
Method is possible, final doesn't exists/play a role here

bastianallgeier (approved):

PR #3371, opened 2021-05-26
Tags field: adding missing layout prop

distantnative:


config/fields/tags.php
```diff
@@ -27,6 +27,12 @@
         'icon' => function ($icon = 'tag') {
             return $icon;
         },
+        /**
+         * Set to `list` to show each tag as 100% width
+         */
+        'layout' => function ($layout = null) {
```
afbora: No needed the type parameter for `data-layout` attribute? `string $layout = null`

distantnative: Oh Yes, type hinting totally makes sense there!

distantnative: Have added it

afbora (approved):

config/fields/tags.php
```diff
@@ -27,6 +27,12 @@
         'icon' => function ($icon = 'tag') {
             return $icon;
         },
+        /**
+         * Set to `list` to show each tag as 100% width
+         */
+        'layout' => function (?string $layout = null) {
```
lukasbestle, line 34: In Vue the prop type is `String`, so I think this shouldn't be nullable, but instead something like:

```suggestion
        'layout' => function (string $layout = 'inline') {
```

The default layout name is just an idea, so far we don't seem to have a name for it.

distantnative: As far as I know, Vue accepts `null` as well there. Otherwise our setup so far would've been broken. And wouldn't this add the `data-layout` attribute now to all tags, while so far it only was added when explicitly setting the layout? I think that would be less desirable to be honest.

afbora: Yes, if value is `null`, `undefined` or `false`, attribute will not render.

lukasbestle: Ah, OK, that makes sense. Then I think we only need to describe the default value in the docblock like so:

```
Set to `list` to display each tag with 100% width,
otherwise the tags are displayed inline
```

bastianallgeier: I just added the comment

bastianallgeier (approved):

lukasbestle (approved):

PR #3372, opened 2021-05-26
Deprecate optional `parent` prop in `Cms\File`

distantnative:
### Deprecated
- Creating a `Kirby\Cms\File` without a `parent` property has been deprecated and throws a warning. Starting in 3.7.0 the property will be required and cause a breaking error if not passed.

## Initial post

This https://github.com/getkirby/getkirby.com/issues/1348 issue for the website pointed out that the  `parent` prop should be listed as required, since the `File::create()` method will throw an exception if it is not passed.

However, looking at the code I could not figure out why it should be required. We only seem to pass it to `File::factory()` which creates a new `File` object. There the property setter `File::setParent()` states the property as optional.

@lukasbestle @bastianallgeier 
Do you remember maybe why this Exception was put there to require the `parent` prop?

### DIFF
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -520,7 +520,7 @@ class File extends ModelWithContent
     {
         // @codeCoverageIgnoreStart
         if ($parent === null) {
-            deprecated('You are creating a `Kirby\Cms\File` object without passing a `parent` prop. Whule unsupported, this hasn\'t thrown any direct errors so far. to fix inconsistencies, the `parent` property will be required for creating a `Kirby\Cms\File` object with 3.7.0 and higher. Not passing one will start throwing a breaking error');
+            deprecated('You are creating a `Kirby\Cms\File` object without passing the `parent` property. While unsupported, this hasn\'t caused any direct errors so far. To fix inconsistencies, the `parent` property will be required when creating a `Kirby\Cms\File` object in Kirby 3.7.0 and higher. Not passing this property will start throwing a breaking error.');
         }
         // @codeCoverageIgnoreEnd
 
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -511,7 +511,9 @@ class File extends ModelWithContent
     }
 
     /**
-     * Sets the parent model object
+     * Sets the parent model object;
+     * this property is required for `File::create()` and
+     * will be generally required starting with Kirby 3.7.0
      *
      * @param \Kirby\Cms\Model|null $parent
      * @return $this
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -514,9 +514,16 @@ class File extends ModelWithContent
      *
      * @param \Kirby\Cms\Model|null $parent
      * @return $this
+     * @todo make property required in 3.7.0
      */
     protected function setParent(Model $parent = null)
     {
+        // @codeCoverageIgnoreStart
+        if ($parent === null) {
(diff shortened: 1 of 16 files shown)


lukasbestle:
Yes, because several methods like `File::root()`, `apiUrl()` and `mediaRoot()` rely on the parent object.

distantnative:
But is then the mistake that `File::setParent` treats it as optional?

lukasbestle:
Yes, I think so. Either `setParent` should require the property or all methods that rely on it need to have a fallback if it isn't defined. I think the first option would be the easiest, but I'm not sure about possible breaking changes.

bastianallgeier:
I also think that it's a mistake in setParent. We don't have any isolated files that don't have a parent. The only ones are assets and they are handled by a different class.

distantnative:
@lukasbestle @bastianallgeier I have tried that (see new commit), but at least in a larger number of our tests, `parent` is also treated as optional. So not sure if this won't be a big breaking change for others as well.

lukasbestle:
Are there use-cases where it would make sense to manually instantiate a `File` object from scratch besides our tests? As long as the files are only accessed from the site tree, the `parent` is set automatically anyway.

distantnative:
Virtual files?

lukasbestle:
Yeah, you are right. These do have parents (their virtual or real page), but it could well be that users are manually instantiating the files without passing the `parent` property explicitly.

distantnative:
It's really tricky cause I feel either way it would be a breaking change. But the current state - `File::create()` treats it as required, `File::setParent()` not - seems like a bug as well.

lukasbestle:
Yep, currently a file object without parent is of limited use as many methods will not work if the parent is not defined.

Maybe we should deprecate a missing parent property in 3.6.0 (with deprecation warning) and then enforce it in 3.7.0.

distantnative:
@lukasbestle I have adapted the PR as suggested

src/Cms/File.php
```diff
@@ -514,9 +514,16 @@ protected function setFilename(string $filename)
      *
      * @param \Kirby\Cms\Model|null $parent
      * @return $this
+     * @todo make property required in 3.7.0
      */
     protected function setParent(Model $parent = null)
     {
+        // @codeCoverageIgnoreStart
+        if ($parent === null) {
+            deprecated('You are creating a `Kirby\Cms\File` object without passing a `parent` prop. Whule unsupported, this hasn\'t thrown any direct errors so far. to fix inconsistencies, the `parent` property will be required for creating a `Kirby\Cms\File` object with 3.7.0 and higher. Not passing one will start throwing a breaking error');
```
lukasbestle:
```suggestion
            deprecated('You are creating a `Kirby\Cms\File` object without passing the `parent` property. While unsupported, this hasn\'t caused any direct errors so far. To fix inconsistencies, the `parent` property will be required when creating a `Kirby\Cms\File` object in Kirby 3.7.0 and higher. Not passing this property will start throwing a breaking error.');
```

lukasbestle (commented):
Looks good! I've fixed a few typos and made some changes to make the message more clear.

Could you please check why the E2E tests started failing? In one of the test cases an error component pops up, so it looks like we use the deprecated behavior ourselves somewhere.

distantnative:
👍

lukasbestle (dismissed):
I've also updated the backend tests to pass the parent prop so we don't have this work when we make it required in 3.7.0.

lukasbestle:
I've rebased this PR against `develop` to incorporate the CI Codecov fix. Besides the updated `setParent()` docblock there are no other changes.

lukasbestle (approved):

distantnative:
👍 lgtm

PR #3373, opened 2021-05-26
Select field: remove undocumented `empty` prop

distantnative:
## Status: wait for https://github.com/getkirby/kirby/pull/2815 ⏸

- Removes the undocumented `empty` prop
- `empty = false` prop would remove the empty option, thus acting the same way as `required` but without actually enforcing it
- doesn't seem to me a functional logic - we should rely on `required` instead

After this PR:

```
mySelect:
  type: select
  options:
    - A
    - B
    - C
```

Screen Shot 2021-05-26 at 21 04 38
Screen Shot 2021-05-26 at 21 04 33

```
mySelect:
  type: select
  required: true
  options:
    - A
    - B
    - C
```

Screen Shot 2021-05-26 at 21 05 30
Screen Shot 2021-05-26 at 21 04 38

```
mySelect:
  type: select
  required: true
  default: A
  options:
    - A
    - B
    - C
```

Screen Shot 2021-05-26 at 21 06 40

## Breaking changes

- Select field: the `empty` option has been removed. Pleas use `required` instead.

_If someone has used `empty: false` so far and upgrades without making changes, the only effect is that now an empty option shows up. Saving and validation is just as before (not enforced)._

## Related issues/ideas

- Fixes https://github.com/getkirby/getkirby.com/issues/1006

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/config/fields/select.php
+++ b/config/fields/select.php
@@ -15,7 +15,7 @@ return [
             return $icon;
         },
         /**
-         * Custom placeholder string for empty option.
+         * Text shown when no option is selected yet
          */
         'placeholder' => function (string $placeholder = '—') {
             return $placeholder;
--- a/panel/src/components/Forms/Input/SelectInput.vue
+++ b/panel/src/components/Forms/Input/SelectInput.vue
@@ -55,14 +55,7 @@ export const props = {
     ariaLabel: String,
     default: String,
     /**
-     * The text, that is shown as the first empty option, when the field is not required.
-     */
-    empty: {
-      type: [Boolean, String],
-      default: true
-    },
-    /**
-     * The text, that is shown when no option is selected yet.
+     * Text shown when no option is selected yet
      */
     placeholder: String,
     options: {
@@ -97,10 +90,6 @@ export default {
       return this.placeholder || "—";
     },
     hasEmptyOption() {
-      if (this.empty === false) {
-        return false;
-      }
-
       return !(this.required && this.default);
     },
     label() {


lukasbestle:
Maybe we should ask our users on Discord if anyone has used this option so far. And if so, what the use-case was. I can't think of one either, but maybe we are missing something critical.

medienbaecker:
I don't like this change to be honest.
How would you remove the empty option for select fields with dynamic options (like a query)?

distantnative:
@medienbaecker why can't you set required to true?

medienbaecker:
Looking at the last example it looks like you need a default to remove the empty option. You can't do that with dynamic options. Or am I overlooking something?

distantnative:
Yes, but isn't that behavior logical? Without a default, you need a state for when the field value is still not set. With required true that empty option is listed but cannot be selected again, once selecting a value.

It's just that once I tried documenting the empty option, I found it very weird to describe the current behavior: we create a state where we remove the empty select option but telling the field to still accept an empty value. So if the field has no value yet but also the empty option gets removed, we're creating kind of a limbo state that's pretty confusing in my view.

medienbaecker:
I, as a Kirby user, would think the first option is the default automatically. If I remove the empty option the first real option would be the default.

Sometimes I also like to give the first option something like a "No XY selected" text so it's clearer than just "–".

distantnative:
@medienbaecker But that's what `placeholder` would be perfect for (adapting the label of the empty value). Sounds to me that you're removing the empty option just to recreate it?

distantnative:
And sounds like your other quirk would be resolved if `default` would allow queries, so you could say "always default to the first option"?

medienbaecker:
Yes, that would be the perfect solution in my opinion.

The placeholder works fine too 👍

iocouto:
Agree that the behaviour of 'required' here, together with 'default', is what developers expect logically - and seems to be more inline with the way configuration works for other fields.

distantnative:
Converting to PR draft until I've added the necessary `default` support to make this PR meaningful.

medienbaecker:
> @medienbaecker But that's what `placeholder` would be perfect for (adapting the label of the empty value). Sounds to me that you're removing the empty option just to recreate it?

@distantnative I finally remembered why I do the funny `placeholder` recreation.

Let's say I have this field setup:

```yml
link_type:
  label: Link Type
  type: select
  width: 1/3
  placeholder: No Link
  options:
    internal: Internal Link
    external: External Link
no_link:
  label: No Link
  type: info
  width: 2/3
  when:
    link_type: ""
  text: Please select a link type
internal_link:
  label: Internal Link
  type: pages
  multiple: false
  width: 2/3
  when:
    link_type: internal
external_link:
  label: External Link
  type: url
  width: 2/3
  when:
    link_type: external
```

empty-select

While this works _outside_ of structure fields it does not work _inside_ of one. Somehow the `link_type` does not have a value, not even `""` like the condition for `no_link` expects. If I add an "empty" default value and remove the empty option by adding `empty: false` it does work. This might be a different topic more closely related to field conditions, but I wanted to let you know nevertheless.

distantnative:
Closing for now, we'll revisit for future release.

PR #3374, opened 2021-05-26
Merge 3.5.6 changes back into develop

lukasbestle:
I already created the PR now because the merge needed to be done manually (the cherry-picking causes merge conflicts in some files). We still need to merge `master` into `develop` after the release, but at least that will then hopefully work without merge conflicts.


lukasbestle:
@distantnative Could you please check why Cypress is failing? My local Cypress installation doesn't work at the moment.

distantnative:
I won't be able to do so before tonight unfortunately

lukasbestle:
Don't worry, it's nothing urgent.

lukasbestle:
Maybe it's related to #3376.

distantnative:
That's the error cypress sees
Screen Shot 2021-05-27 at 19 04 33

Haven't been able to reproduce it myself manually in the Panel though.

lukasbestle:
That's... weird. Given the tiny set of changes to frontend code in this PR I cannot explain that at all and don't even know where to start searching.

distantnative:
Now started to pop up locally as well. A search revealed that it is apparently a more common issue with vuedraggable and vite at the moment. Suggested was replacing the import path (we use in `Draggable.vue`) with `vuedraggable/src/vuedraggable`. That seems to work for me. For you too?

lukasbestle:
@distantnative Thanks for tracking this down. Do you also see the issue on `develop`? Because I don't understand why this issue suddenly pops up in this PR. Or is it because of a dependency update in `package-lock.json` in this PR?

distantnative:
I've seen this once, then again not, then again, now not... can't really figured out where exactly this is coming from. Last time it was on the inertia branch. But if you google the error, you'll find quite a few other cases having the same issue

afbora:
I don't know exactly where the error is, but if we move the last PR #3380 to this PR, will it be fixed?

lukasbestle:
It's a Vite issue: https://github.com/vitejs/vite/issues/3356
I've fixed it by downgrading to Vite 2.2.3 for now, the version we've been using previously.

I've also merged in the latest state from `release/3.5.6` and already reverted the translation fix for 3.6.0 so it won't cause merge conflicts later.

Given that CI passes now, I think this PR is ready for merging.

afbora:
I guess we have to wait for 3.5.6 to release stable.

lukasbestle:
No, we don't need to wait. The whole point of this PR is to make the merge on release more flawless. That merge will then only contain changes to the version number and translations. :)

afbora:
I just meant the possibility of a regression still occurring in RC.

lukasbestle:
Most of the changes are already on develop as we cherry-picked from there. So merging this back into develop doesn‘t do much except merge Git history after the cherry-picks.

afbora (approved):

PR #3384, opened 2021-05-29
Supports countable field objects

### LINKED ISSUE #3382: $value is an uncountable object error in V::size validator when calling filterBy with `between`
**Describe the bug**  

Example code that causes the error `$value is an uncountable object in kirby/src/Toolkit/V.php on line 478`.
```
<?php dump($page->children()->filterBy('height', 'between', [2, 7])); ?>
```
The error occurs because `$value` is a field object in this case.

**Kirby Version**  
Tested with 3.5.5 and 3.5.6-RC-1

afbora:
## Describe the PR

Field object converted to readable value and now field objects countable.

### Fixes
- Filtering based on field value sizes issue fixed

## Breaking changes

None

## Related issues/ideas

- Fixes #3382 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Toolkit/V.php
+++ b/src/Toolkit/V.php
@@ -463,6 +463,12 @@ V::$validators = [
      * Checks that the value has the given size
      */
     'size' => function ($value, $size, $operator = '=='): bool {
+        // if value is field object, first convert it to a readable value
+        // it is important to check at the beginning as the value can be string or numeric
+        if (is_a($value, '\Kirby\Cms\Field') === true) {
+            $value = $value->value();
+        }
+
         if (is_numeric($value) === true) {
             $count = $value;
         } elseif (is_string($value) === true) {
--- a/tests/Toolkit/VTest.php
+++ b/tests/Toolkit/VTest.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Toolkit;
 
+use Kirby\Cms\Page;
 use PHPUnit\Framework\TestCase;
 
 class CanBeCounted implements \Countable
@@ -388,6 +389,20 @@ class VTest extends TestCase
         $this->assertFalse(V::size(new HasCount(), 8));
         $this->assertFalse(V::size(5, 7, '>'));
         $this->assertFalse(V::size(7, 5, '<'));
+
+        $page = new Page(['slug' => 'a', 'content' => ['foo' => 2]]);
+        $this->assertTrue(V::size($page->foo(), 2));
+        $this->assertTrue(V::size($page->foo(), 3, '<'));
+        $this->assertTrue(V::size($page->foo(), 1, '>'));
+        $this->assertFalse(V::size($page->foo(), 1, '<'));
+        $this->assertFalse(V::size($page->foo(), 3, '>'));
+
+        $page = new Page(['slug' => 'a', 'content' => ['foo' => 'hello']]);
+        $this->assertTrue(V::size($page->foo(), 5));
+        $this->assertTrue(V::size($page->foo(), 6, '<'));
+        $this->assertTrue(V::size($page->foo(), 4, '>'));
+        $this->assertFalse(V::size($page->foo(), 4, '<'));
+        $this->assertFalse(V::size($page->foo(), 6, '>'));
     }
 
     public function testSizeInvalid()


lukasbestle (commented):
Including this change directly in the validator would couple it too much with the CMS code and also we might still have the same issue with other validators or collection filters.

I feel like this should be handled in a new method `Kirby\Cms\Collection::getAttribute()` that overrides the one from `Toolkit\Collection` to convert fields into their values. Then the filters would already get the field values.

I'm not sure about possible breaking changes though, it's really hard to analyze the consequences.

distantnative:
I think including it in the validator would actually be the safest for breaking changes? Cause so far passing a Field object would cause an error, after this it would work. All other things would still behave the same. Whereas trying to solve this elsewhere might have big (breaking) side effects.

And I think it is okay to couple this tighter. We have said that we think the too strict separation has caused us more effort and troubles than actually benefits for others. And you could still use the validator class without the CMS - this change doesn't break that use case either.

distantnative:
Or how about we check here https://github.com/getkirby/kirby/pull/3384/files#diff-e99ecf0b5e40afa7f2f62a7c0c7aa3924e72e7cd19146f7c315045a5f0873700L477 for a method called `value` and if the object has one, pass that value again to `V::size()`? That would be a more abstract way of handling it.

afbora:
> Or how about we check here https://github.com/getkirby/kirby/pull/3384/files#diff-e99ecf0b5e40afa7f2f62a7c0c7aa3924e72e7cd19146f7c315045a5f0873700L477 for a method called `value` and if the object has one, pass that value again to `V::size()`? That would be a more abstract way of handling it.

I thought there, but not nested calling. Yes this seems like the right way to go.

Also I have no idea about Lukas suggestion. 
I'm not sure what breakage it could cause 🤔

lukasbestle:
@distantnative Your code link doesn't work for me. Which file and line do you mean?

afbora:
@lukasbestle Following line and i guess something like that

```php
if ($value instanceof \Countable) {
    $count = count($value);
} elseif (method_exists($value, 'count') === true) {
    $count = $value->count();
} elseif (is_a($value, '\Kirby\Cms\Field') === true) {
    return static::size($value->value(), $size, $operator);
} else {
    throw new Exception('$value is an uncountable object');
}
```

https://github.com/getkirby/kirby/blob/fix/3382-countable-field-object/src/Toolkit/V.php#L478-L486

afbora:
How about checking `__toString()` method for global solution?

lukasbestle (dismissed):
You are right. Considering all options I think the current implementation in this PR is the safest and most robust. A check for `value()` and `__toString()` methods would be inconsistent (given that objects are generally counted and an exception is thrown if they are not), so just handling the special case of field objects is useful and won't cause unexpected regressions.

distantnative:
Why would checking for a value method be inconsistent? Currently we check for already for a count method, I think if an object offers a `value()` method, it would make a lot of sense to run `V::size` on the result of that method (given that no explicit `count()` method exists). That would be much more in line with the general approach this method tries to take than just defining one exception case.

lukasbestle:
@distantnative So far this case of a `value` and/or `__toString` method was just not considered. So if we suddenly support that, it would be a breaking change. Imagine a user wants to validate the count of an object. A non-countable object would be considered invalid, but now the validator would silently use the value or string representation. I think it would be fine to go that route in the future, but we would need to mark the breaking change and keep an eye out for possible regressions.

One thing we need to be careful about is performance: `method_exists()` on every field three times (once for count, once for value, once for `__toString`) is pretty expensive. I think if we decide to do it, we should only check for `__toString` (not for value) as this check could later be replaced with the higher-performance `instanceof Stringable` interface check that was introduced with PHP 8.

distantnative:
I'd actually stay away from `__toString` as many more objects implement that with all sorts of representations (HTML tags, whole URLs). `value()` to me would be one options as it likely represents the actual value of an object. But if this is too debated, I am fine with just `Kirby\Cms\Field` check - only that I think I'd find https://github.com/getkirby/kirby/pull/3384#issuecomment-850832990 the prettier version

lukasbestle:
`value()` is used for scalar values in Kirby classes, but I don't think we can assume that for all objects. `__toString()` would at least be standardized in its behavior. Are there any Kirby classes where using the return value of `__toString()` would lead to weird results?

> only that I think I'd find [#3384 (comment)](https://github.com/getkirby/kirby/pull/3384#issuecomment-850832990) the prettier version

To be honest I feel like the current implementation is very solid as it's sequential and doesn't rely on recursion (which can be quite a bit slower and also confusing as the logic needs to be considered in two layers).

afbora:
I think the ideas you have come up with are great 👍 

If it will not cause a problem in terms of performance, I would prefer the following usage. It would be compatible for all classes to be added later.

```php
if ($value instanceof \Countable) {
    $count = count($value);
} elseif (method_exists($value, 'count') === true) {
    $count = $value->count();
} elseif (method_exists($value, '__toString') === true) {
    return static::size($value->__toString(), $size, $operator);
} else {
    throw new Exception('Cannot determine count or value from object $value');
}
```

lukasbestle:
I'm not 100 % happy with the recursion, but this implementation would be fine for me.

afbora:
If you think the recursion might be the problem later, we can take it to the beginning like the current implementation 🤔

lukasbestle:
Given that `__toString()` by definition always returns a string, we might not actually need recursion here. Can't we just always use the length of the returned string?

**Edit:** Ah, no. Because the string might be numeric. Another alternative would be to use the same code inside. Not 100 % DRY, but this has much higher performance than the recursive solution and is easier to keep the overview:

```php
if ($value instanceof \Countable) {
    $count = count($value);
} elseif (method_exists($value, 'count') === true) {
    $count = $value->count();

// @todo replace with `$value instanceof \Stringable` once support for PHP 7 is dropped
} elseif (method_exists($value, '__toString') === true) {
    $count = (string)$value;

    if (is_numeric($count) === false) {
        $count = Str::length(trim($count));
    }
} else {
    throw new Exception('Cannot determine count or value from object $value');
}
```

afbora:
I think it makes sense 👍

But wouldn't it be better if we call the `->__toString()` method directly instead of casting to string? This solution may be correct only for the `Field` object, but I think it may produce different results for another objects.

```php
// @todo replace with `$value instanceof \Stringable` once support for PHP 7 is dropped
} elseif (method_exists($value, '__toString') === true) {
    $count = $value->__toString();

    if (is_numeric($count) === false) {
        $count = Str::length(trim($count));
    }
}
````

lukasbestle:
Either way is fine. Casting an object to string calls `__toString()` by definition, so both variants do exactly the same thing for any kind of object. `(string)$value` is IMO the cleaner variant.

afbora:
Oh you right 🙈 I can implement after @distantnative review.

distantnative:
Tbh, I still don't get why we are going with `__toString()` and not `value()`. I don't think the string representation of an object is a good base for measuring the size of an object.

lukasbestle:
`value()` is not standardized, so we cannot assume that each object uses this method name as a getter for a numeric or string value. It could really do and return anything unless we depend on an interface (not very Kirby-like).

distantnative:
> we cannot assume that each object uses this method name as a getter for a numeric or string value

I'd say that argument is just as true for `__toString`. IMO the name `value()` would be clear enough to expect all objects to return their true value (whatever that is), but if that's too risky for us, I'd argue we need to stick to only `Kirby\Cms\Field` (though here again, I'd argue `$field->value()` would make more sense to measure the size than `$field->__toString()`.

lukasbestle:
> > we cannot assume that each object uses this method name as a getter for a numeric or string value
> 
> I'd say that argument is just as true for `__toString`.

`__toString()` is always a getter for a string value, so it's really well defined in the interface sense. You can't suddenly run into cases where calling it has side effects or returns another object, `bool` or similar. `__toString()` is always expected to not have side effects *and* to return a string value.

Before we start splitting hairs, let's summarize:

- `value()` is potentially risky and calling `method_exists()` for every object is quite slow
- `__toString()` is not always suitable for validation (depending on the class)

I agree the best option for now would be to only support fields here as we discussed previously. In this case I think we can approve and merge this PR as it is.

The best long-term solution would be to convert fields into their values already in the `Collection::getAttribute()` method, but the regression potential is high, so it's nothing we should do at the moment.

distantnative:
Rebased and changed base to `release/3.5.7`

distantnative (approved):

PR #3391, opened 2021-06-02
Deprecate `Str::isUrl()` #3386

### LINKED ISSUE #3386: Inconsistency: `Str::isUrl()` and `V::url()`
Why aren't we using `V::url()` in `Str::isUrl()` - or in general: why do we need two different methods which say to do the same but do it differently?

distantnative:
### Deprecations
- `Kirby\Toolkit\Str::isUrl()` has been marked as deprecated and will be removed in 3.8.0

## Related issues/ideas

- Fixes #3386

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -414,7 +414,7 @@ class Str
      */
     public static function isURL(?string $string = null): bool
     {
-        return $string !== null ? V::url($string) : false;
+        return filter_var($string, FILTER_VALIDATE_URL) !== false;
     }
 
     /**
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -407,10 +407,13 @@ class Str
      *
      * @param string|null $string
      * @return bool
+     * @deprecated 3.6.0 use `Kirby\Toolkit\V::url()` instead
+     * @todo Throw deprecation warning in 3.7.0
+     * @todo Remove in 3.8.0
      */
-    public static function isURL(string $string = null): bool
+    public static function isURL(?string $string = null): bool
     {
-        return filter_var($string, FILTER_VALIDATE_URL) !== false;
+        return $string !== null ? V::url($string) : false;
     }
 
     /**
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -410,6 +410,7 @@ class Str
      * @deprecated 3.6.0 use `Kirby\Toolkit\V::url()` instead
      * @todo Throw deprecation warning in 3.7.0
      * @todo Remove in 3.8.0
+     * @codeCoverageIgnore
      */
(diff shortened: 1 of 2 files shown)


src/Toolkit/Str.php
```diff
@@ -407,10 +407,13 @@ public static function from(string $string, string $needle, bool $caseInsensitiv
...
      * @param string|null $string
      * @return bool
+     * @deprecated 3.6.0 use `Kirby\Toolkit\V::url()` instead
+     * @todo Throw deprecation warning in 3.7.0
+     * @todo Remove in 3.8.0
      */
     public static function isURL(string $string = null): bool
     {
-        return filter_var($string, FILTER_VALIDATE_URL) !== false;
+        return $string ? V::url($string) : false;
```
afbora: to standardize

````php
public static function isURL(?string $string = null): bool
{
    return $string !== null ? V::url($string) : false;
}
````

afbora: `nullable` type? `?string` :)

afbora:
Also I guess we need to add `@codeCoverageIgnore` to docblock.

afbora (dismissed):

lukasbestle:
`V::url()` behaves very differently from `FILTER_VALIDATE_URL`. So this would be a breaking change. Would it be a problem to keep the original implementation for now and deprecate the method as it is?

afbora:
Yes it is a breaking change for those using `Str::isUrl()`. We were using regex in `V::url()` because `FILTER_VALIDATE_URL` didn't work for some URLs.

I think @lukasbestle suggestion is more appropriate. We are creating unnecessary breaking changes in the method we will remove anyway. So we can remove it in 3.7.0 instead of 3.8.0.

lukasbestle:
I think Nico's plan was to deprecate the method in 3.6 without a deprecation warning for now to give users a bit more time. This means that the warning will be in 3.7, so the method can only be removed in 3.8.

Fine in my opinion, but we should still not modify the behavior. Also the only existing use in `Html::link()` should already be migrated to `V::url()`.

bastianallgeier (dismissed):

bastianallgeier:
After reading it again, I agree with @lukasbestle. Deprecating is fine, but we should not modify it yet.

distantnative:
Agreed as well. Will change it in a second - but @bastianallgeier don't add your approval if the current state isn't actually what we want to merge :D

distantnative:
I've left `Str::isUrl()` unchanged now and also switched `Html::link()` over to using `V::url()` instead.
I would still stick to the time plan (3.7 warnings, 3.8 removed). As Lukas keeps posting out, 3.6 has already a lot of breaking changes/things to pay attention to as a developer. And this one here I really don't feel urgent about.

lukasbestle (approved):

PR #3397, opened 2021-06-06
`k-items` and cardlets 

distantnative:
## Status: in progress 👷

- Adds `k-items` and `k-item`
- Adds `k-options-dropdown`
- Adds `cardlets` layout option for sections and picker fields
- Removes `k-cards`, `k-card`, `k-list`, `k-list-item`
- Removes `previewThumb` helper
- Use `k-items` for search results

## Related Feedback

- https://kirby.nolt.io/91

### Issues
- [x] Dropdown entries cannot be clicked (instead link of item is targeted)
- [x] User placeholder icons have the wrong color

#### Before PR
dist/css/style.css   93.54kb / brotli: 13.86kb
dist/js/index.js     353.33kb / brotli: 64.64kb

#### After PR
dist/css/style.css   95.12kb / brotli: 14.04kb
dist/js/index.js     351.35kb / brotli: 64.59kb


config/api/models/File.php
```diff
@@ -59,9 +59,6 @@
         'options' => function (File $file) {
             return $file->panel()->options();
         },
-        'panelIcon' => function (File $file) {
```
distantnative: Probably better to deprecate it and return the full `panel()->image()` information for the time being - that would at least include all previous keys/values of `->icon()`

bastianallgeier:
I cannot estimate how much work it is, but do you think it would be possible to move the new panel->image() changes on the backend into its own PR? I feel like we could more easily get this discussed and integrated if it wasn't combined with all the component work on the frontend. It's only a suggestion though. If it takes forever to separate it, we better fight through this.

distantnative:
No I think it makes total sense. I had the idea anyway to move some parts to their own PR (e.g. OptionsDropdown). Open question before I do that: are we keeping `image()` or going with a new term that includes icon?

bastianallgeier:
That's great! I think image works fine. The only other option would be preview, but that could collide with the page preview settings or maybe some live preview feature later.

bastianallgeier:
Hm, thinking about it: maybe `visual` could be a more neutral alternative though?

distantnative:
Yea and no. I think it's a bit more neutral but not as much. Etter that it really make sense to change the name everywhere and add more breaking changes. IMO

bastianallgeier:
You are right. Let's stay with image

distantnative:
I'll try extracting separate PRs tonight

distantnative:
@bastianallgeier I think I need your help on the CSS issue for the dropdown. -somehow the dropdown renders below other elements and the clicks are caught by the item not the dropdown entry. So the dropdown links aren't clickable.

Other than that I haven't encountered any remaining issues so far.

bastianallgeier:
@distantnative ok, I will have a look. Do you have a chance to fix the frontend tests?

distantnative:
@bastianallgeier fixed some of the test, I think the last failing one is due to the CSS issue as well

bastianallgeier:
Yay, I made a couple fixes and now tests pass. I'm really happy with this PR.

bastianallgeier (approved):

bastianallgeier:
@distantnative would you mind adding release notes for it to Notion?

PR #3403, opened 2021-06-07
PHPUnit tests: Always pass roots to app instances

lukasbestle:
Avoids polluting the sandbox installation with temporary test files.


distantnative:
That's where those came from!

bastianallgeier (approved):
Ahhhh, nice one! This was so annoying.

PR #3409, opened 2021-06-08
Better image and icon defaults for panel models and fixed user avatars #3402

### LINKED ISSUE #3402: [3.6.0 – Fiber] User avatars not displayed in the users overview
**To Reproduce**  

1. Upload an avatar for a user
2. Go back to the users overview
3. The user is displayed with the default user icon, not the avatar

bastianallgeier:
## Describe the PR
See title

## Related issues/ideas
- Fixes #3402

## Ready?
- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Panel/Model.php
+++ b/src/Panel/Model.php
@@ -95,14 +95,22 @@ abstract class Model
      */
     public function icon(array $params = null): array
     {
-        $defaults = [
+        return array_merge($this->iconDefaults(), $params ?? []);
+    }
+
+    /**
+     * Default settings for icons
+     *
+     * @return array
+     */
+    public function iconDefaults(): array
+    {
+        return [
             'type'  => 'page',
             'ratio' => null,
             'back'  => 'pattern',
             'color' => '#c5c9c6',
         ];
-
-        return array_merge($defaults, $params ?? []);
     }
 
     /**
@@ -115,17 +123,14 @@ abstract class Model
      */
     public function image($settings = null): ?array
     {
-        $defaults = [
-            'ratio' => '3/2',
-            'back'  => 'pattern',
-            'cover' => false
-        ];
+        $defaults = $this->imageDefaults();
 
         // switch the image off
         if ($settings === false) {
             return null;
         }
 
+        // convert string settings to proper array
         if (is_string($settings) === true) {
             // use defined icon in blueprint
             if ($settings === 'icon') {
@@ -136,6 +141,9 @@ abstract class Model
                 'query' => $settings
             ];
         }
+
+        // merge defaults with given settings
+        $settings = array_merge($defaults, (array)$settings);
 
         if ($image = $this->imageSource($settings['query'] ?? null)) {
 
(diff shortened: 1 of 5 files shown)


distantnative:
Not too happy with it as it will create quite the big merge conflicts with https://github.com/getkirby/kirby/pull/3397
But in case we cannot wait for that, might be something we have to do :/

src/Panel/Model.php
```diff
@@ -95,14 +95,22 @@ public function dragTextType(string $type = null): string
...
-        $defaults = [
+        return array_merge($this->iconDefaults(), $params ?? []);
+    }
+
+    /**
+     * Default settings for icons
+     *
+     * @return array
+     */
+    public function iconDefaults(): array
```
lukasbestle, line 106: The return value is pretty specific to pages. I feel like we should make this method abstract and define it in `Panel\Page` instead. The disadvantage would be that we would need to define it in `Panel\Site` and `Panel\File` as well even though they use a completely different implementation for the icon (none or more complex respectively).

src/Panel/User.php
```diff
@@ -30,15 +30,32 @@ public function breadcrumb(): array
...
-    public function icon(array $params = null): array
+    public function iconDefaults(): array
     {
-        $params['type'] = 'user';
-        return parent::icon($params);
+        return [
+            'type'  => 'user',
+            'ratio' => '1/1',
+            'back'  => 'black',
+            'color' => '#c5c9c6',
```
lukasbestle, line 43: Is it possible to use our CSS vars for the colors? This would allow changing these colors with the general color scheme in a custom `panel.css`.

lukasbestle (commented):
It looks like Nico made similar but different changes to the icon and image settings. I don't fully understand how all the code connects, so maybe it's better if you two first discuss the best way forward.

Two comments based on the current code in case we go this route:

bastianallgeier:
I haven't looked into the item PR yet. If you managed it differently there, we should definitely find a common way forward. I will give your solution another look tomorrow and try to adapt my PR to it. 

But in the "old" state on develop, it was really hard to extend the image and icon methods properly.

src/Panel/User.php
distantnative: With the changes on my PR this would actually work. With this PR alone, no.

bastianallgeier:
@distantnative does it even make sense to keep this PR up at all? Or would it be better to get yours done first?

distantnative:
I think a main share of the mess and confusion is cause by gathering different arrays for icon and image and then trying to combine them in Vue somehow. On the `k-items` PR I have started to integrate the icon data rather in the `image()` return. This produces much more reliable and easier to handle results in my opinion. It's sadly not as clean as it could be no the PR as I tried to avoid bigger breaking changes - that's why the method than is still called `image()` and not `thumb()` or `preview()`. Same the key for the icon is still called `type` and not `icon` - that way everyone who used `->icon()` before could easily switch to `->image()` and still get the right data for same keys.

All in all, I would rather invest work in finishing and polishing `k-items`. Cause this PR would definitely cause merge conflicts and we would have to ask the same question again in a way.

bastianallgeier:
That sounds great. I will definitely look closer into it tomorrow then.

lukasbestle:
@distantnative Since the `Panel` namespace is completely new in 3.6, is there really a breaking change if we use a different name?

distantnative:
@lukasbestle well it connects through to the deprecated `$model->panelIcon()` as well as the API models' `panelIcon` key. So in a way yes - but as I said I tried it to make as little disruptive as possible. Giving it all different names would probably be quite disruptive though

lukasbestle:
The API model is indeed an issue, but I don't think we would need to change the name of the deprecated methods. They can still alias to the new name.

bastianallgeier:
I'm closing this in favour of @distantnative's PR.

PR #3414, opened 2021-06-09
Refactor Panel class 

### LINKED ISSUE #3404: [3.6.0] Refactor `Panel\Panel` class
As discussed in https://github.com/getkirby/kirby/pull/3327#pullrequestreview-676808165:

- Move `page`, `file` and `user` methods to a new class (maybe `Cms\Finder`?)
- Convert `Panel` class to a singleton that takes the `$kirby` object

Should be done in Kirby 3.6.0 as the new foundation can then be built upon in future releases. If we only do this later, it would be a breaking change.

bastianallgeier:
## Describe the PR

`Panel::page()`, `Panel::user()`, `Panel::parent()` and `Panel::file()` didn't properly fit in there, as they are also used by the API routes. They are now included in a new `Kirby\Cms\Find` class

## Related issues/ideas

- Fixes #3404 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/areas/site.php
+++ b/config/areas/site.php
@@ -1,6 +1,6 @@
 <?php
 
-use Kirby\Panel\Panel;
+use Kirby\Cms\Find;
 
 return function ($kirby) {
     return [
@@ -13,26 +13,26 @@ return function ($kirby) {
         'routes' => [
             [
                 'pattern' => 'pages/(:any)',
-                'action'  => function (string $path) use ($kirby) {
-                    return Panel::page($path)->panel()->route();
+                'action'  => function (string $path) {
+                    return Find::page($path)->panel()->route();
                 }
             ],
             [
                 'pattern' => 'pages/(:any)/files/(:any)',
-                'action'  => function (string $id, string $filename) use ($kirby) {
-                    return Panel::file('pages/' . $id, $filename)->panel()->route();
+                'action'  => function (string $id, string $filename) {
+                    return Find::file('pages/' . $id, $filename)->panel()->route();
                 }
             ],
             [
                 'pattern' => 'site',
-                'action'  => function () use ($kirby) {
-                    return $kirby->site()->panel()->route();
+                'action'  => function () {
+                    return Find::site()->panel()->route();
                 }
             ],
             [
                 'pattern' => 'site/files/(:any)',
-                'action'  => function (string $filename) use ($kirby) {
-                    return Panel::file('site', $filename)->panel()->route();
+                'action'  => function (string $filename) {
+                    return Find::file('site', $filename)->panel()->route();
                 }
             ],
 
--- a/config/areas/site.php
+++ b/config/areas/site.php
@@ -26,7 +26,7 @@ return function ($kirby) {
             [
                 'pattern' => 'site',
                 'action'  => function () {
-                    return Find::site()->panel()->route();
+                    return site()->panel()->route();
                 }
             ],
             [
--- a/src/Cms/Find.php
+++ b/src/Cms/Find.php
@@ -149,7 +149,13 @@ class Find
 
         // get the authenticated user
         if ($id === null) {
-            return $kirby->user(null, $kirby->option('api.allowImpersonation', false));
+            if ($user = $kirby->user(null, $kirby->option('api.allowImpersonation', false))) {
+                return $user;
+            }
+
+            throw new NotFoundException([
+                'key' => 'user.undefined'
+            ]);
         }
 
         // get a specific user by id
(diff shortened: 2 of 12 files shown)


bastianallgeier:
The code coverage report is pretty weird again. Why does removing parts from the Panel class suddenly trigger new detected lines?

distantnative:
I think that's not the issue but the two uncovered lines in `Find` (122-123)

bastianallgeier:
I'm obviously still too stupid to read the reports correctly.

config/areas/site.php
```diff
@@ -13,26 +13,26 @@
...
+                'action'  => function (string $id, string $filename) {
+                    return Find::file('pages/' . $id, $filename)->panel()->route();
                 }
             ],
             [
                 'pattern' => 'site',
-                'action'  => function () use ($kirby) {
-                    return $kirby->site()->panel()->route();
+                'action'  => function () {
+                    return Find::site()->panel()->route();
```
distantnative: I think this is unnecessary - it communicates that we're doing something special here while the old version would just work fine. We aren't finding anything, just using the site object as is.

src/Cms/Find.php
```diff
@@ -0,0 +1,173 @@
...
+ * the Panel to find models and parents
+ * based on request paths
+ *
+ * @package   Kirby Cms
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier GmbH
+ * @license   https://getkirby.com/license
+ */
+class Find
```
distantnative, line 20: Maybe the name should be less generic and more tied to the objects it treats, e.g. `ModelFind` or so

afbora: Like `Finder`?

distantnative: No, `Finder` is just as generic as `Find` - I think it would be good to create a connection to models with the name. Having `Kirby\Cms\Find` next to `Kirby\Cms\Search` is just so similar.

bastianallgeier: I wanted a name that is "fun" to use in Area and API routes. I'm not 100% sure about Find but at least it really looks nice and readable in routes. We will have those methods all over the place. They will be in the dialogs as well.

bastianallgeier: What about `Access` to convey the concept of trying to access those models in routes?

config/areas/site.php
bastianallgeier: Right now, it's totally useless. But I was thinking about the case of improved permissions. Maybe we want to disable read access for the site to some users or modify model access in a different way. But we could of course leave this for now until we really need it.

distantnative: But this shouldn't be solved in the method getting the site object, right? but then in the `Panel\Site::route` method or so

src/Cms/Find.php
distantnative: `Access` is worse cause it's all about getting the model not what access we have to them. `find` is good as the concept what it is achieving. And I get the part of super short and fun names to use, but with so many classes we have I'm also concerned about just looking at a list of classes and knowing what each one does.

config/areas/site.php
bastianallgeier: It depends. Right now we already check for readable pages and files in `Find::page` and `Find::file` This is quite important because we can already pretend they are not even there and throw a proper NotFoundException. But for more detailed permissions checks we would indeed probably go a different way. I will remove this method for now. It's adding more confusion than actual benefits.

src/Cms/Find.php
distantnative: I guess it'd be easier if we had a `Kirby\Model` namespace :D

bastianallgeier: @distantnative I agree, but we open yet another one of Pandora's boxes with that. I have no better idea to be honest. `ModelFinder::page()` is really something I don't want to use in the routes. In the API we already solve this by binding the API instance to routes and then routes can access models with $this->page(), $this->file() etc. But I have to say that this obscure binding of instances to $this doesn't feel very good either. I don't want to follow the same path for Fiber routes and also would love to move away from that in Api routes some day.

distantnative: Ok, I won't block this over the name 😅😉

distantnative (dismissed):

src/Cms/Find.php
bastianallgeier: Thanks :) Maybe our Greenland shark @lukasbestle should still have a look at it and give his opinion. The biggest problem is still that we are stuck with the name once we introduce the class.

bastianallgeier:
Sorry, I just fucked this PR up

PR #3423, opened 2021-06-10
`$helper.color` JS helper for using CSS color vars

distantnative:
## Describe the PR

Helper that allows us to support in Vue components color props that accept
- HEX codes
- full `var(-- )` declarations
- `pattern` string that gets converted to our background pattern
- strings that represent our color vars, e.g. `red-200` gets transformed to `var(--color-red-200)`

### Features
- New `$helper.color` helper method to turn a string into a valid CSS color value (e.g. hex code or CSS variable)

### DIFF
--- a/panel/src/components/Misc/Icon.vue
+++ b/panel/src/components/Misc/Icon.vue
@@ -6,15 +6,18 @@
     :data-back="back"
     :data-size="size"
     :class="'k-icon k-icon-' + type"
+    :style="{ background: toColor(back) }"
   >
     <span v-if="isEmoji" class="k-icon-emoji">{{ type }}</span>
-    <svg v-else :style="{ color: color }" viewBox="0 0 16 16">
+    <svg v-else :style="{ color: toColor(color) }" viewBox="0 0 16 16">
       <use :xlink:href="'#icon-' + type" />
     </svg>
   </span>
 </template>
 
 <script>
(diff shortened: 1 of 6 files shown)


panel/src/components/Misc/Icon.vue
```diff
@@ -6,15 +6,18 @@
     :data-back="back"
     :data-size="size"
     :class="'k-icon k-icon-' + type"
+    :style="{ background: toColor(back) }"
```
bastianallgeier: Wouldn't it make sense to just use `$helper.color(back)` here?

bastianallgeier (changes requested):

bastianallgeier:
This is great!

panel/src/components/Misc/Icon.vue
distantnative: You mean not putting the helper in a `css` namespace/object and actually registering it as a global helper?

distantnative: ok yes

bastianallgeier: Do you think we need to put it in a css namespace at all? Anyway, I really think it would be good to have it as a global helper. This is super useful for plugin devs as well.

distantnative:
@bastianallgeier changed it

bastianallgeier (approved):

PR #3425, opened 2021-06-10
Unify `panel()->image()`

distantnative:
## Describe the PR

-Merges `Panel\Model::icon()` into `Panel\Model::image()`
- Adds `Panel\Model::imageDefaults()`, `Panel\Model::imageColor()`, `Panel\Model::imageIcon()`
- Depends on https://github.com/getkirby/kirby/pull/3423 to actually show the colors in the Panel

### Enhancements
- `->panel()->image()` now supports a second parameter to specify layout (default is `cards`)

### Breaking changes

- Use `$model->panel()->image()` instead `$model->panel()->icon()`
- Removed `icon` prop from `k-card` and `k-list-item`. Pass data as part of the `image` prop instead.
- `->panel()->image()` doesn't return `cards` and `list` keys with data for both sizes but instead only `src` and `srcset` keys at the top level that correspondent to the `layout` parameter

### Deprecations

- The API field `page.panelIcon` has been deprecated and will be removed in 3.7.0. Use `page.panelImage` instead
- The API field `file.panelIcon` has been deprecated and will be removed in 3.7.0. Use `file.panelImage` instead

## Ready?

- [ ] CI checks pass

## When merging


bastianallgeier (dismissed):
This looks so much cleaner. Do you take care of the missing code coverage or should I have a look at it?

distantnative:
I'm tempted to wrap these uncovered lines in code coverage ignore comments. They are all part of the picker fields and still seem very temporary there - they aren't such clean solutions but I needed to draw a line there so the PR wouldn't blow up. 

But I know Lukas probably wouldn't like that :D

bastianallgeier:
Ok, then @lukasbestle should have a look at it :)

distantnative:
Ok, I would really argue for merging this as is:
- the uncovered lines have not been covered before this PR either
- it only affects the same lines in all 3 picker fields
- we can confirm whether the change works manually (it does for me, e.g. switching the `cover` field of the `album` blueprint to `layout: cards` does load the images in the right resolution)
- the picker field code isn't great and needs to be refactored soon anyways, I'd argue it makes more sense to write proper unit sense then instead of spending that time now on testing bad code (considering that the 3 points above that still should make sure the change works)

bastianallgeier:
I've covered the panelImage part in the API user model now and I really agree that everything else is not directly related and should be ignored. @lukasbestle what's the best way to "force-merge" this PR?

afbora:
I've added unit tests of uncovered lines. We can merge this PR after CI workflow issue fixed

afbora (approved):

panel/src/config/search.js
```diff
@@ -1,71 +1,71 @@
...
       q: query,
       limit: limit || 10,
-      select: ["id", ...fields, "panelIcon", "panelImage"]
+      select: ["id", ...fields, "panelImage"]
     });
 
+    console.log(["id", ...fields, "panelImage"]);
+
     return response.data.map(item => {
+      console.log(item);
```
lukasbestle, line 12: @distantnative Were these logs intended? Or just used while debugging?

src/Cms/Page.php
```diff
@@ -1484,24 +1484,6 @@ public function dragText(string $type = null): string
...
-     * @internal
-     * @param array|null $params
-     * @return array
-     * @codeCoverageIgnore
-     */
-    public function panelIcon(array $params = null): array
-    {
-        return $this->panel()->icon($params);
-    }
-
```
lukasbestle, line 1504: Can we completely remove this method without prior deprecation?

panel/src/config/search.js
distantnative: Only for debugging -  but this has been fixed already quite some time ago. And by now `search.js` doesn't exist anymore (got replaced by Fiber implementation).

src/Cms/Page.php
distantnative: The deprecated method is still inherited from `ModelWithConent`

PR #3426, opened 2021-06-10
`Str::contains`: use `strpos` instead

distantnative:
According to https://www.php.net/manual/en/function.strstr.php

> If you only want to determine if a particular needle occurs within haystack, use the faster and less memory intensive function strpos() instead.

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -250,7 +250,7 @@ class Str
      */
     public static function contains(string $string = null, string $needle, bool $caseInsensitive = false): bool
     {
-        return call_user_func($caseInsensitive === true ? 'stristr' : 'strstr', $string, $needle) !== false;
+        return call_user_func($caseInsensitive === true ? 'stripos' : 'strpos', $string, $needle) !== false;
     }
 
     /**


src/Toolkit/Str.php
```diff
@@ -250,7 +250,7 @@ public static function between(string $string = null, string $start, string $end
      */
     public static function contains(string $string = null, string $needle, bool $caseInsensitive = false): bool
     {
-        return call_user_func($caseInsensitive === true ? 'stristr' : 'strstr', $string, $needle) !== false;
+        return call_user_func($caseInsensitive === true ? 'stripos' : 'strpos', $string, $needle) !== false;
```
afbora, line 253:
```suggestion
        return call_user_func($caseInsensitive === true ? 'mb_stripos' : 'mb_strpos', $string, $needle) !== false;
```

Good catch! Also just would be great with multi-byte.

bastianallgeier: @distantnative @afbora is right. This would be even better. Do you see any problems? Could of course even decrease performance in general. I'm not sure how well the mb versions perform in comparison to the regular checks.

distantnative: That was my only concern - otherwise of course agreed with @afbora

distantnative: Reading posts like this https://stackoverflow.com/questions/24344491/why-is-mb-strpos-so-considerably-slower-than-strpos it looks like the mb versions will be a lot slower. But on the other hand non-supporting these characters could also be considered a bug.

bastianallgeier: Ok, I think we should be pragmatic about it. So far, we've not offered mb support in this method and after thinking about it again, we should first focus on the PR. The PR is improving the method as it is. Adding mb support would be a completely separate task with such additional questions. Let's merge the PR as it is and keep mb support in mind.

bastianallgeier (approved):

PR #3427, opened 2021-06-11
Silent Fiber requests

### LINKED ISSUE #3422: [3.6.0] content locking API requests shouldn't show loading indicator in topbar

distantnative:
## Describe the PR

Makes sure that the content locking requests via Fiber also happen silently again (no loading indicator).

## Related issues/ideas

- Fixes #3422

### DIFF
--- a/panel/src/config/fiber.js
+++ b/panel/src/config/fiber.js
@@ -272,7 +272,11 @@ const Fiber = {
     this.saveScroll();
 
     if (!silent) {
-      document.dispatchEvent(new Event("fiber:start"));
+      document.dispatchEvent(
+        new CustomEvent("fiber:start", {
+          detail: { silent }
+        })
+      );
     }
 
     // make sure only is an array
--- a/panel/src/config/fiber.js
+++ b/panel/src/config/fiber.js
@@ -105,7 +105,7 @@ const Fiber = {
       "$translation",
       "$urls",
       "$user",
-      "$view",
+      "$view"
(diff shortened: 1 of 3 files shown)


bastianallgeier (commented):
This is great, but I would not omit the fiber:start event. Instead I would pass the silent option to the event and then check in our Fiber:start event listener if it's a silent request.

distantnative:
Aren't we using that event only for showing the loading indicator?

bastianallgeier:
Yes, but from an event standpoint the event is still happening and you might still want to listen to it for other cases later.

distantnative:
@bastianallgeier ✅

bastianallgeier (approved):
Perfect!

panel/src/config/fiber.js
```diff
@@ -258,19 +258,26 @@ const Fiber = {
...
+      replace = false,
+      silent = false
     } = {}
   ) {
     // save the current scrolling positions
     // for all scroll regions
     this.saveScroll();
 
-    document.dispatchEvent(new Event("fiber:start"));
+    if (!silent) {
```
lukasbestle, line 274: Doesn't this mean that the event is never dispatched if `silent` is `true` as Bastian suggested? Or is the event dispatched from elsewhere in this case?

distantnative: https://github.com/getkirby/kirby/pull/3453

PR #3429, opened 2021-06-11
Simplified Panel class methods (Panel class refactoring part 2)

### LINKED ISSUE #3404: [3.6.0] Refactor `Panel\Panel` class

bastianallgeier:
## Describe the PR

The Panel class methods are no longer passing down the $kirby instance. Each method that needs the $kirby instance will now get it from `App::instance()` instead. This simplifies the method calls quite drastically and doesn't hurt the testability. 

After @lukasbestle suggested to switch to non-static methods and a singleton for the Panel class, I made a few tests and decided that sticking with the static method calls keeps the Panel class a lot easier to use in my opinion. We can discuss this though. 

## Related issues/ideas

- Fixes #3404

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Panel/Panel.php
+++ b/src/Panel/Panel.php
@@ -3,7 +3,6 @@
 namespace Kirby\Panel;
 
 use Exception;
-use Kirby\Cms\App;
 use Kirby\Cms\User;
 use Kirby\Exception\PermissionException;
 use Kirby\Http\Request;
@@ -32,6 +31,13 @@ use Throwable;
 class Panel
 {
     /**
+     * Kirby singleton
+     *
+     * @var \Kirby\Cms\App
+     */
+    public static $kirby;
+
+    /**
      * Normalize a panel area
      *
      * @param string $id
@@ -55,11 +61,11 @@ class Panel
     /**
      * Collect all registered areas
      *
-     * @param \Kirby\Cms\App $kirby
      * @return array
      */
-    public static function areas(App $kirby): array
+    public static function areas(): array
     {
+        $kirby  = kirby();
         $root   = $kirby->root('kirby') . '/config/areas';
         $system = $kirby->system();
         $user   = $kirby->user();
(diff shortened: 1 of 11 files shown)


src/Panel/Panel.php
```diff
@@ -544,12 +557,6 @@ public static function partial(string $component, array $data): array
             return $data;
         }
 
-        // if a new component is requested, the include
```
distantnative, line 653: Aren't these quite important lines? Can't find them anymore elsewhere

tests/Panel/PanelTest.php
```diff
@@ -771,85 +794,62 @@ public function testPartialWithComponentComparison(): void
...
-        $result = Panel::partial($this->app, 'k-test-view', $data);
-
-        $this->assertSame(['a' => 'A'], $result);
-
-        // via headers
-        $this->app = $this->app->clone([
-            'request' => [
-                'headers' => [
-                    'X-Fiber-Include' => 'a',
-                    'X-Fiber-Component' => 'k-test-view'
```
distantnative, line 783: Is it intentional that we aren't testing for headers anymore?

bastianallgeier: I thought about it some more while working on the clean up and I cannot see a single point where this is really needed in our setup. 

I only see two use cases for partial reloads

1) to refresh global data
2) to refresh parts of the current view

The $reload method reload is already perfect for both cases and prevents errors because you can only request the same view again. 

The $go method should not allow requesting partial content at all. It simply doesn't make sense. But instead of adding this weird check in the backend, we could simply fix this in the client implementation by throwing out the only option in $go. 

This has the added benefit of simplifying the backend calls – not just for views, but also for dialogs. In dialogs there's no need for all of this.

distantnative: ok 👍 only concern (but I cannot think it through at the moment) would be partial reloads that still alter the current URL (e.g. partial reload of a section because pagination changed and thus URL with page parameter needs to change too)

bastianallgeier: I'd suggest to look into it again if that should ever get us in trouble.

afbora (approved):
If Nico is OK, LGTM too 🙈 Also everything looks great in my local tests 👍

PR #3434, opened 2021-06-14
New `A::map` method

bastianallgeier:
## Describe the PR

Finally a wrapper for array_map with the array as the first argument. 

## Release notes

- New `A::map($array, $callback)` method

## Related issues/ideas

- Fixes #

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -376,6 +376,19 @@ class A
     }
 
     /**
+     * A simple wrapper around array_map
+     * with a sane argument order
+     *
+     * @param array $array
+     * @param \Closure $map
+     * @return array
+     */
+    public static function map(array $array, Closure $map): array
+    {
+        return array_map($map, $array);
+    }
+
+    /**
      * Move an array item to a new index
      *
      * @param array $array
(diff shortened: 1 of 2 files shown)


afbora:
Should we replace this method with the `array_map` functions we use in other classes?

bastianallgeier:
@afbora eventually yes, but I didn't want to add it to this PR.

src/Toolkit/A.php
```diff
@@ -375,6 +375,19 @@ public static function fill(array $array, int $limit, $fill = 'placeholder'): ar
...
 
+    /**
+     * A simple wrapper around array_map
+     * with a sane argument order
+     *
+     * @param array $array
+     * @param \Closure $map
+     * @return array
+     */
+    public static function map(array $array, Closure $map): array
```
afbora:
```suggestion
     * @param callable $map
     * @return array
     */
    public static function map(array $array, callable $map): array
```

Should be `callable` to work correctly. For ex: `A::map($array, 'trim');` Also it would be nice if there was a test for the different map argument.

bastianallgeier: ✅

afbora (approved):

PR #3438, opened 2021-06-15
Render dialogs using Fiber

### LINKED ISSUE #3361: [3.6.0] Dialogs via Fiber

bastianallgeier:
## Describe the PR

This PR introduces the new Fiber Dialogs. It's a pretty slow process to recreate and test them all, but it's so worth it. We will now be able to create a indefinite amount of dialogs and drawers right from the backend. No more JS code needed. 

This PR draft is here to document the ongoing process

### Backend

- [x] FileRemoveDialog
- [x] FileRenameDialog
- [x] FileSortDialog
- [x] LanguageCreateDialog
- [x] LanguageRemoveDialog
- [x] LanguageUpdateDialog
- [x] PageCreateDialog
- [x] PageDuplicateDialog
- [x] PageRemoveDialog
- [x] PageRenameDialog
- [x] PageSortDialog
- [x] PageStatusDialog
- [x] PageTemplateDialog
- [x] RegistrationDialog
- [x] SiteRenameDialog
- [x] UserCreateDialog
- [x] UserEmailDialog
- [x] UserLanguageDialog
- [x] UserPasswordDialog
- [x] UserRemoveDialog
- [x] UserRenameDialog
- [x] UserRoleDialog
- [x] Solve redirect after page has been deleted
- [x] Solve redirect after page has been renamed
- [x] Solve redirect after file has been deleted
- [x] Solve redirect after file has been renamed 
- [x] Solve redirect after user has been deleted
- [x] Solve redirect after user has been renamed
- [x] unit tests for file sort dialog submit event
- [x] Filter dropdown in Pages- and Files section to exclude the delete option if the min requirement is reached. 

### Frontend

- [x] Implement store dispatcher in Dialog submit event
- [x] Implement optional redirects in submit event
- [x] New slug field
- [x] New Vue.$dialog method
- [x] Unified Vue.$request for all Fiber requests
- [x] Replace / remove outdated Dialog components
- [x] Better Panel class setup 

## Possible performance benefits

With current dialog implementation on develop: 

```
dist/css/style.css   93.54kb / brotli: 13.86kb
dist/js/index.js     353.53kb / brotli: 64.75kb
```

Afterwards without dialog components: 

```
dist/css/style.css   95.29kb / brotli: 14.05kb
dist/js/index.js     317.82kb / brotli: 60.59kb
```

## Fixes

- https://github.com/getkirby/kirby/issues/3462
- https://github.com/getkirby/kirby/issues/3469

## Breaking changes

The following components are no longer available: 

- k-file-remove-dialog
- k-file-rename-dialog
- k-file-sort-dialog
- k-language-create-dialog
- k-language-remove-dialog
- k-language-update-dialog
- k-page-create-dialog
- k-page-duplicate-dialog
- k-page-remove-dialog
- k-page-rename-dialog
- k-page-sort-dialog
- k-page-status-dialog
- k-page-template-dialog
- k-registration-dialog
- k-site-rename-dialog
- k-user-create-dialog
- k-user-email-dialog
- k-user-language-dialog
- k-user-password-dialog
- k-user-remove-dialog
- k-user-rename-dialog
- k-user-role-dialog

## Related issues/ideas

- Fixes #

## Ready?

- [x] CI checks pass

## When merging


config/areas/site/dialogs.php
```diff
@@ -0,0 +1,458 @@
...
+                if (Panel::referrer() === $oldUrl && $oldUrl !== $newUrl) {
+                    $response['redirect'] = $newUrl;
+                }
+            }
+
+            return $response;
+        }
+    ],
+
+    // change create
```
distantnative: probably supposed to just be `create`?

panel/package-lock.json
```diff
@@ -10936,6 +11043,18 @@
           "integrity": "sha1-M7QHd3VMZDJXPBIMw4CLvRDUfwQ=",
           "dev": true
         },
+        "vue": {
+          "version": "3.0.11",
```
distantnative, line 11047: we're upgrading to Vue 3?

panel/src/components/Dialogs/Dialog.vue
```diff
@@ -147,12 +148,27 @@ export default {
...
+      this.$emit("close");
+      this.$events.$off("keydown.esc", this.close);
+      this.$store.dispatch("dialog", false);
+    },
     /**
      * Opens the dialog and triggers the `@open` event
      * @public
      */
     open() {
-      this.$store.dispatch("dialog", true);
```
distantnative, line 155: We are not setting it to open anymore but above still setting the store when closing?

panel/src/components/Dialogs/FormDialog.vue
```diff
@@ -53,6 +59,28 @@ export default {
...
+  },
+  data() {
+    return {
+      model: this.value
+    }
+  },
+  computed: {
+    listeners() {
+      return {
+        ...this.$listeners,
```
distantnative: This is a dark pattern that Vue 3 won't support anymore, so maybe at least for all new code we can already try to find a better implementation?

panel/src/components/Sections/FieldsSection.vue
```diff
@@ -36,6 +36,11 @@ export default {
       return this.$store.getters["content/values"]();
     }
   },
+  watch: {
+    timestamp() {
```
distantnative, line 42: Since I was quite irritated where this is coming from... maybe we should in general add a comment what `timestamp` is/does?

panel/src/config/fiber.js
```diff
@@ -260,6 +260,7 @@ const Fiber = {
     {
       data = {},
       headers = {},
+      include = [],
```
distantnative: We now have `only` and `include` - to me that totally sound the same. What's the difference? Or can we merge them?

src/Panel/Field.php
```diff
@@ -0,0 +1,264 @@
...
+
+namespace Kirby\Panel;
+
+use Kirby\Cms\File;
+use Kirby\Cms\Page;
+
+/**
+ * Provides common field prop definitons
+ * for dialogs and other places
+ *
```
distantnative, line 11: Tbh this feels like we are adding just another place for field-related code (after `config/field`, `Kirby\Cms` and `Kirby\Form`). Not sure we are doing ourselves a service with that.

src/Panel/Panel.php
```diff
@@ -89,7 +90,9 @@ public static function areas(): array
         // load plugins
         foreach ($kirby->extensions('areas') as $id => $area) {
             if (is_a($area, 'Closure') === true) {
-                $areas[$id] = static::area($id, (array)$area($kirby));
+                if (isset($areas[$id]) === false) {
```
distantnative: This will prevent replacing areas via plugins

src/Panel/Panel.php
```diff
@@ -276,18 +279,50 @@ public static function data(array $data = []): array
             }
         ];
 
+        // check for explicitly requested globals to be included
```
distantnative: As mentioned above, the duality of `include` and `only` isn't really clear to me - they seem to do different, but I don't get exactly what. And their names are too equal to carry meaning for differentiation.

Code here looks like a lot of duplicate code from the `only` implementation, would be better to have this DRY.

panel/package-lock.json
bastianallgeier: How is that even possible? I ran an npm audit fix today. But this should never bump us to v3. How is it even possible that the Panel is still working like that.

panel/src/components/Dialogs/Dialog.vue
bastianallgeier: Let me give that a closer look again.

panel/src/components/Dialogs/FormDialog.vue
bastianallgeier: What is the suggested alternative?

panel/src/components/Sections/FieldsSection.vue
bastianallgeier: There's one huge benefit of the Fiber setup that we didn't fully use yet. Sections don't really need to watch for certain model changes anymore. I.e. when dialogs close. In such a case we always have a reloaded view. The easiest way to tell the sections about a reloaded view is actually such an updated timestamp that's coming from the backend with every view response. So basically any component can listen for those timestamp updates and then knows that the route/model has been updated. This will even be easier when we refactor sections and send them directly on view load, but for now this is a really great way to keep everything in sync.

panel/src/config/fiber.js
bastianallgeier: Here's the explanation: 

only: is needed in cases when you want to reload a single part of the view response (i.e. the lock info) You send the fields that you want to get back and they response is merged with what's already there. 

include: is needed when you want to include globals in the view response. I.e. to update the $translation. In this case you get a full view response back, but you also include a global key that would normally not be included. I'm definitely open for better names. I liked "include" as we use it in a similar way for API calls.

src/Panel/Field.php
bastianallgeier: Yes, definitely. As I wrote in Discord, it's a shortcut because our form code is such a mess. I wish it was completely reusable here, but the complexity of our form implementation did not let me find a solution that was fast enough for the alpha to get this done.

src/Panel/Panel.php
bastianallgeier: Yes, and there's a good reason for that. Right now you could easily ruin the entire Panel with a single plugin, including quite some security relevant stuff. We need to develop a standard way for plugins to replace parts of our main areas without completely fucking everything up.

config/areas/site/dialogs.php
bastianallgeier: yep

panel/src/components/Dialogs/FormDialog.vue
distantnative: You could try binding `$attr` instead in Vue 3: https://v3.vuejs.org/guide/migration/listeners-removed.html#_2-x-syntax

But I think the actual reading is that one should clearly define what events and listeners are supported

panel/src/config/fiber.js
distantnative: Ok I understand it now, but I'd say we definitely need different names. Cause implicitly they both are saying "only include"/"include only". Why don't we just call them `props` and `globals`. If you don't specify any, you get all. As soon as you specify them, you only get these. And by the name, you already know what this refers to.

panel/src/components/Dialogs/FormDialog.vue
bastianallgeier: Yeah, it's probably the best idea to be more explicit.

panel/src/config/fiber.js
bastianallgeier: I like that a lot.

panel/src/components/Dialogs/FormDialog.vue
distantnative: Will also help to understand our code better: what events do we actually want to support there?

panel/package-lock.json
bastianallgeier: For some reason this is only in the package-lock.

panel/src/components/Dialogs/FormDialog.vue
bastianallgeier: ✅

config/areas/site/dialogs.php
bastianallgeier: ✅

panel/src/components/Dialogs/Dialog.vue
bastianallgeier: ✅

panel/src/components/Sections/FieldsSection.vue
bastianallgeier: ✅

panel/src/config/fiber.js
bastianallgeier: After rethinking it some more, I think it's wrong to split it into props and globals. In practice, it's more about the task. 

You want to load `only` one item or you want to `replace` items in the response/state.
Items can either be globals or props. 

That's why I renamed the include parameter into replace. I think this makes it a lot clearer what's the expected outcome.

I.e.

```js
this.$reload({
  replace: '$translation'
});
```

vs. 

```js
this.$reload({
  only: '$view.props.lock'
});
```

In general, replace is the way to go. It will refresh the rest of the view state as well and state changes will always be complete. 

Only is "only" needed for stuff like content locking when you really just need that one piece of the state to be refreshed.

bastianallgeier: Damn, directly after writing the comment I'm not so sure again about `replace`. The intention of the call is to reload the view, but also include a global. So it should actually probably be: 

```js
this.$reload({
  globals: '$translation'
});
```

vs. 

```js
this.$reload({
  only: '$view.props.lock'
});
```
 
The first one will return the entire view response + the requested globals. 
The second one will only return a particular prop. That can be a shared prop or a global prop. I.e. this would also work

```js
this.$reload({
  only: '$translation'
});
```

But in this case the response will only contain the $translation and nothing else.

bastianallgeier: "globals" and "only" it is!

src/Panel/Field.php
bastianallgeier: Do you think it hurts when we keep this for the alpha and try to find a better solution in the next weeks after the launch?

distantnative: I think that would be fine

afbora:
@bastianallgeier About failing on two PHP unit tests;
When i debug the tests while all running, i'm getting following error and this can be give you an idea:

> Undefined component type: info

bastianallgeier:
I already started debugging this and the error is caused by a missing info field, which is needed in one of the dialogs. 

It seems like one of the prior tests deletes or overwrites the info field. All fields are only loaded once in the Kirby\App class and then cached. That's why the dialog tests cannot pass afterwards without a working field. I could simply not find which test deletes the info field though.

bastianallgeier:
@afbora the culprit is somewhere in the tests/Cms folder. I'm currently going through all of the tests to isolate the issue

afbora:
I'm so close. This issue about `\tests\Cms\Sections\SectionTest` tests.

bastianallgeier:
Oh, you are right! I think I know how to fix it.

afbora:
I found the issue.  Do you want to me push the commit?

**Solution for all overwrites**

````php
Section::$types = ['test' => []];
````

to

````php
Section::$types['test'] = [];
````

**Comment by bastianallgeier** [comment]
I might have a better solution. Wait a second.

**Comment by bastianallgeier** [comment]
I think it's cleaner to register on setup which sections exist and restore them on tear down. I'm currently checking if this works.

**Comment by bastianallgeier** [comment]
Solved it with a mixture of your version and mine. Thank you so much for finding this. It really drove me nuts :)

**Comment by lukasbestle** [comment]
I've verified that the memory limit is indeed too low. I was concerned that our test code might be leaking memory somewhere, but it turns out the tests without coverage collection run just fine. Just the coverage collection needs too much memory (which is likely caused by our growing test and code base and no issue by itself).

So the solution is indeed to increase the memory limit. The issue is: We fix it to `512M` in `Image\Darkroom\GDLib`, which overwrites any value we set in our PHPUnit config. So to solve this, the `GDLib` file either needs to not set the memory limit at all or it should check if the memory limit is already higher than `512M` (which is unfortunately difficult as the `memory_limit` setting uses a weird format with `K`, `M` and `G` at the end, just the number of bytes or `-1` for disabled).

**Comment by afbora** [comment]
How about `ini_get()` with `Str::toBytes()`? 

Of course, it would be better not to fix it there. Maybe a new option, a constant or a global static variable 🤔 

_Side note: By the way, is it really necessary to use 512M memory? If necessary, why is this information not included in the requirements document?_

**Comment by lukasbestle** [comment]
I feel like it would be best to remove [this line from `GDLib.php`](https://github.com/getkirby/kirby/blob/3.5.7-rc.1/src/Image/Darkroom/GdLib.php#L5) all together. It's a weird side-effect and shouldn't be in a random class file. I think it's a good idea to rather add this information to the requirements section (like "to use the Kirby thumb engine with GDLib (default), a PHP `memory_limit` of 512 MB or more is recommended").

Removing the memory limit would be a breaking change, but this is 3.6 anyway.
I don't think we need an option or similar as the memory limit can be set directly in the `php.ini` or in the config file with `ini_set()`.

What do you think?

File: config/areas/settings/dialogs.php [hunk 10]
```diff
@@ -0,0 +1,171 @@
...
+                        'name'      => '',
+                        'code'      => '',
+                        'direction' => 'ltr',
+                        'locale'    => ''
+                    ]
+                ]
+            ];
+        },
+        'submit' => function () {
+            kirby()->languages()->create(get());
```
lukasbestle: I feel like the props should be allowlisted to avoid weird prop injections:

```suggestion
        'submit' => function () use ($languageDialogFields) {
            $props = array_intersect_key(get(), $languageDialogFields);
            kirby()->languages()->create($props);
```

config/areas/settings/dialogs.php
```diff
@@ -0,0 +1,171 @@
...
+                        'code'      => $language->code(),
+                        'direction' => $language->direction(),
+                        'locale'    => $locale,
+                        'rules'     => $language->rules(),
+                    ]
+                ]
+            ];
+        },
+        'submit' => function (string $id) {
+            $language = Find::language($id)->update(get());
```
lukasbestle: Same here, but this should also take the allowed fields into account as defined by the `load` callback above.

config/areas/site/dialogs.php
```diff
@@ -0,0 +1,458 @@
...
+    ],
+
+    // delete page
+    'pages/(:any)/delete' => [
+        'load' => function (string $id) {
+            $page        = Find::page($id);
+            $hasSubpages = $page->childrenAndDrafts()->count();
+            $text        = tt('page.delete.confirm', ['title' => $page->title()->value()]);
+
+            if ($hasSubpages) {
```
lukasbestle:
```suggestion
            $page         = Find::page($id);
            $subpageCount = $page->childrenAndDrafts()->count();
            $text         = tt('page.delete.confirm', ['title' => $page->title()->value()]);

            if ($subpageCount > 0) {
```

config/areas/site/dialogs.php
```diff
@@ -0,0 +1,458 @@
...
+            if ($hasFiles) {
+                $fields['files'] = [
+                    'label'    => t('page.duplicate.files'),
+                    'type'     => 'toggle',
+                    'required' => true,
+                    'width'    => $hasChildren ? '1/2' : '1/1'
+                ];
+            }
+
+            if ($hasChildren) {
```
lukasbestle:
```suggestion
            if ($hasFiles === true) {
                $fields['files'] = [
                    'label'    => t('page.duplicate.files'),
                    'type'     => 'toggle',
                    'required' => true,
                    'width'    => $hasChildren ? '1/2' : '1/1'
                ];
            }

            if ($hasChildren === true) {
```

config/areas/site/dialogs.php
```diff
@@ -0,0 +1,458 @@
...
+                        'files'    => false,
+                        'slug'     => $page->slug() . '-' . Str::slug(t('page.duplicate.appendix'))
+                    ]
+                ]
+            ];
+        },
+        'submit' => function (string $id) {
+            $newPage = Find::page($id)->duplicate(get('slug'), [
+                'children' => get('children'),
+                'files'    => get('files'),
```
lukasbestle:
```suggestion
                'children' => (bool)get('children'),
                'files'    => (bool)get('files'),
```

config/areas/site/dialogs.php
```diff
@@ -0,0 +1,458 @@
...
+
+    // change page position
+    'pages/(:any)/changeSort' => [
+        'load' => function (string $id) {
+            $page     = Find::page($id);
+            $position = null;
+
+            if ($page->blueprint()->num() !== 'default') {
+                // TODO: make translatable
+                throw new Exception('You cannot change the position of this page manually');
```
lukasbestle: Is there anything that stops us from making it translatable immediately?

config/areas/users/dialogs.php
```diff
@@ -0,0 +1,263 @@
...
+                        'email'    => '',
+                        'password' => '',
+                        'language' => $kirby->panelLanguage(),
+                        'role'     => $kirby->user()->role()->name()
+                    ]
+                ]
+            ];
+        },
+        'submit' => function () {
+            kirby()->users()->create(get());
```
lukasbestle:
```suggestion
            $props = array_intersect_key(get(), [
                'name'     => true,
                'email'    => true,
                'password' => true,
                'language' => true,
                'role'     => true
            ]);
            kirby()->users()->create($props);
```

panel/src/components/Dialogs/FormDialog.vue
```diff
@@ -2,12 +2,18 @@
...
-    v-on="$listeners"
+    @cancel="$emit('cancel')"
+    @close="$emit('close')"
+    @ready="$emit('ready')"
+    @submit="$refs.form.submit()"
   >
-    <!-- eslint-disable vue/no-mutating-props -->
+    <template v-if="text">
+      <!-- eslint-disable-next-line vue/no-v-html -->
+      <k-text v-html="text" />
```
lukasbestle, line 12: Where can the text come from? Is it protected from XSS?

src/Cms/App.php
```diff
@@ -825,12 +825,13 @@ public function languageCode(string $languageCode = null): ?string
     /**
      * Returns all available site languages
      *
+     * @param bool
```
lukasbestle, line 828:
```suggestion
     * @param bool $clone
```

In which use-cases do we need the clone behavior? Maybe we could deprecate it in the future (in this case a simple TODO comment is enough for now).

src/Cms/Language.php
```diff
@@ -234,19 +237,19 @@ public static function create(array $props)
      */
     public function delete(): bool
     {
-        if ($this->exists() === false) {
-            return true;
-        }
-
```
lukasbestle, line 240: Why was this removed?

src/Cms/SiteActions.php
```diff
@@ -55,7 +55,13 @@ protected function commit(string $action, array $arguments, Closure $callback)
      */
     public function changeTitle(string $title, string $languageCode = null)
     {
-        $arguments = ['site' => $this, 'title' => $title, 'languageCode' => $languageCode];
+        $title     = trim($title);
+        $arguments = [
+            'site'         => $this,
+            'title'        => $title,
+            'languageCode' => $languageCode
+        ];
```
lukasbestle:
```suggestion
        $site     = $this;
        $title     = trim($title);
        $arguments = compact('site', 'title', 'languageCode');
```

tests/Panel/Areas/FileDialogsTest.php
```diff
@@ -0,0 +1,379 @@
...
+class FileDialogsTest extends AreaTestCase
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->install();
+        $this->login();
+    }
+
+    public function createPageFile(): void
```
lukasbestle:
```suggestion
    protected function createPageFile(): void
```

tests/Panel/Areas/FileDialogsTest.php
```diff
@@ -0,0 +1,379 @@
...
+            ]
+        ]);
+
+        // pretend the file exists
+        F::write($this->app->page('test')->file('test.jpg')->root(), '');
+
+        $this->login();
+    }
+
+    public function createSiteFile(): void
```
lukasbestle:
```suggestion
    protected function createSiteFile(): void
```

tests/Panel/Areas/FileDialogsTest.php
```diff
@@ -0,0 +1,379 @@
...
+            ]
+        ]);
+
+        // pretend the file exists
+        F::write($this->app->site()->file('test.jpg')->root(), '');
+
+        $this->login();
+    }
+
+    public function createUserFile(): void
```
lukasbestle:
```suggestion
    protected function createUserFile(): void
```

tests/Panel/PanelTest.php
```diff
@@ -635,6 +687,52 @@ public function testGlobalsWithUser(): void
...
+     * @covers ::go
+     */
+    public function testGo()
+    {
+        try {
+            Panel::go('test');
+        } catch (Redirect $r) {
+            $this->assertSame('/panel/test', $r->getMessage());
+            $this->assertSame(302, $r->getCode());
+        }
```
lukasbestle: To ensure that the exception is thrown at all:

```suggestion
        $thrown = false;

        try {
            Panel::go('test');
        } catch (Redirect $r) {
            $thrown = true;
            $this->assertSame('/panel/test', $r->getMessage());
            $this->assertSame(302, $r->getCode());
        }

        $this->assertTrue($thrown);
```

tests/Panel/PanelTest.php
```diff
@@ -635,6 +687,52 @@ public function testGlobalsWithUser(): void
...
+    /**
+     * @covers ::go
+     */
+    public function testGoWithCustomCode()
+    {
+        try {
+            Panel::go('test', 301);
+        } catch (Redirect $r) {
+            $this->assertSame(301, $r->getCode());
+        }
```
lukasbestle, line 270:
```suggestion
        $thrown = false;

        try {
            Panel::go('test', 301);
        } catch (Redirect $r) {
            $thrown = true;
            $this->assertSame(301, $r->getCode());
        }

        $this->assertTrue($thrown);
```

tests/Panel/PanelTest.php
```diff
@@ -635,6 +687,52 @@ public function testGlobalsWithUser(): void
...
+                ]
+            ]
+        ]);
+
+        try {
+            Panel::go('test');
+        } catch (Redirect $r) {
+            $this->assertSame('/foo/test', $r->getMessage());
+            $this->assertSame(302, $r->getCode());
+        }
```
lukasbestle, line 291:
```suggestion
        $thrown = false;

        try {
            Panel::go('test');
        } catch (Redirect $r) {
            $thrown = true;
            $this->assertSame('/foo/test', $r->getMessage());
            $this->assertSame(302, $r->getCode());
        }

        $this->assertTrue($thrown);
```

panel/src/config/fiber.js
```diff
@@ -56,43 +63,57 @@ const Fiber = {
...
+   * Prepares a set of values to
+   * be included in a comma-separated list
+   * in a URL query (see include & only params)
+   *
+   * @param {string|Array} array
+   * @returns Array
+   */
+  arrayToString(array) {
+    if (Array.isArray(array) === false) {
+      array = [array];
```
lukasbestle: `array.join()` on an array with a single value is just the value converted to string:

```suggestion
      return String(array);
```

panel/src/config/fiber.js
```diff
@@ -217,137 +405,81 @@ const Fiber = {
...
+  /**
+   * Builds a full URL object based on the
+   * given path or another URL object and query data
+   *
+   * @param {string|URL} url
+   * @param {Object} query
+   * @returns
+   */
+  url(url = "", query = {}) {
+    if (typeof url === "string" && url.startsWith("http") === false) {
```
lukasbestle: What about a plugin view with a name that starts with `http`? Safer:

```suggestion
    if (typeof url === "string" && url.match(/^https?:\/\//) === null) {
```

src/Panel/Field.php
```diff
@@ -0,0 +1,264 @@
...
+    /**
+     * User role radio buttons
+     *
+     * @param array $props
+     * @return array
+     */
+    public static function role(array $props = []): array
+    {
+        // TODO: exclude the admin role, if the user
+        // is not allowed to change role to admin
```
lukasbestle: I think this is rather important for a good UX.

src/Panel/Panel.php
```diff
@@ -883,12 +1055,15 @@ public static function view(array $view = [], ?array $area = null): array
         $view = array_replace_recursive($area ?? [], $view);
 
         $view['breadcrumb'] = $view['breadcrumb'] ?? [];
+        $view['code']       = $view['code'] ?? 200;
         $view['path']       = Str::after($kirby->path(), '/');
+        $view['timestamp']  = time();
```
lukasbestle: Is the granularity of one second enough? We could use:

```suggestion
        $view['timestamp']  = (int)(microtime(true) * 1000);
```

for milliseconds.

config/areas/site/dialogs.php
bastianallgeier: Looking at it again, the variable is not needed at all.

panel/src/components/Dialogs/FormDialog.vue
bastianallgeier: The text comes from our fiber dialog definition. It's not protected from XSS by default if you pass a field or something. But this is always used for translations in our own dialogs.

src/Cms/App.php
bastianallgeier: We used clone for App::languages() to protect the main languages collection from being accidentally overwritten, which would lead to many unwanted side-effects. But in all language actions, we actually need to overwrite it to make sure it's up to date. I.e. when a language gets deleted or updated. This is why I introduced a new clone prop. This way, I can get the uncloned instance of the collection and modify that. To be honest, I'm not 100% happy with this approach. But I didn't have an idea for a better solution. It's the typical dilemma that we have with our weird immutable objects everywhere.

src/Cms/Language.php
bastianallgeier: This simply checked for an existing language file. But languages could also be defined via App constructor. In this case, they don't physically exist, but the rest of the delete method should still be executed.

bastianallgeier:
@lukasbestle thanks for digging through the tests. The memory limit should really not be set in the GDLib class. That's a very ancient artefact :) I just removed it.

afbora:
Few spots i can see. Please ignore those not related to this PR.

- [x] LanguageUpdateDialog

PHP locale string automatic update when name changed (before was not trigger, should trigger only on create)

- [x] SettingsView

Languages list icon disappeared and vertical align broken

- [x] Page Update Title/Slug Dialog

`Create from title` wizard disappeared

- [x] Select a new status Dialog

Only `1` menu position appears always when try to update draft page to listed

- [x] Pages ListItem Dropdown Delete Button

Delete button still enabled even if user has no permission

bastianallgeier:
@afbora thanks for your tests. They help a lot!

afbora:
I can see `Sandbox` in dropdown like listed page even if it is unlisted page when i try to update status/num of the draft page.

image (18)

bastianallgeier:
@afbora ✅

afbora:
Thanks 👍 At least for me everything seems to be working great now  🎉

bastianallgeier:
@distantnative @lukasbestle @afbora I feel like this PR is really ready now. What do you think?

afbora:
Sorry, I don't have the capacity/knowledge to review this 😅 

screen-capture-1100-Render dialogs using Fiber by bastianallgeier · Pull Request #3438 · getki_-github com

distantnative:
If you are done, @bastianallgeier, I think we can merge it. Three might be still smaller bugs here and there, but we will probably find them better merged on `develop` and heading towards alpha.

bastianallgeier:
@distantnative I'm done and I agree. It's better to merge it now and improve things when they turn up. I just need an approval :) 

@afbora don't worry! You've already been a huge help

distantnative (approved):

PR #3441, opened 2021-06-16
Store empty string when blocks and layout field empty

### LINKED ISSUE #3153: Block field content empty array not treated as empty by isEmpty() method
**Describe the bug**  
Block field content empty array not treated as empty by isEmpty() method

**To Reproduce**  
save a page with an empty block field, write condition $field->isEmpty() 

**Expected behavior**  
field method isEmpty should return 1/true, but returns ''/false because content of field is string of empty array `[]`

**Kirby Version**  
3.5.2RC1

afbora:
## Describe the PR

Store empty string (`''`) instead `[]` when blocks and layout field empty to consistency work with `$field->isEmpty()`.

### ⚠️ Notes for @bastianallgeier

I also asked @lukasbestle where it would be better to do this empty check/control, but we could not decide exactly. Since you created these classes, we think you will know the right place. I think the best place is the `store()` method like current implementation. It could also be done in `FieldClass::valueToJson()` but it would not be consistent if the empty string `''` is not valid JSON.

### Enhancements
- Improved block and layout field value for `isEmpty()` control #3153 

## Breaking changes

None

## Related issues/ideas

- Fixes #3153 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Form/Field/BlocksField.php
+++ b/src/Form/Field/BlocksField.php
@@ -181,6 +181,13 @@ class BlocksField extends FieldClass
     public function store($value)
     {
         $blocks = $this->blocksToValues((array)$value, 'content');
+
+        // returns empty string to avoid storing empty array as string `[]`
+        // and to consistency work with `$field->isEmpty()`
+        if (empty($blocks) === true) {
+            return '';
+        }
+
         return $this->valueToJson($blocks, $this->pretty());
     }
 
--- a/src/Form/Field/LayoutField.php
+++ b/src/Form/Field/LayoutField.php
@@ -146,6 +146,12 @@ class LayoutField extends BlocksField
     {
         $value = Layouts::factory($value, ['parent' => $this->model])->toArray();
 
+        // returns empty string to avoid storing empty array as string `[]`
+        // and to consistency work with `$field->isEmpty()`
+        if (empty($value) === true) {
+            return '';
+        }
+
         foreach ($value as $layoutIndex => $layout) {
             if ($this->settings !== null) {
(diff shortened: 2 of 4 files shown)


distantnative:
Wouldn't it be easier to support an empty array in `isEmpty()`?

afbora:
@distantnative It would certainly be easier and more accurate, but since we don't know the type of the field, we don't know what type of data it stores. If we could bring the data type constant to the fields, we could do some operations accordingly.

https://github.com/getkirby/kirby/blob/release/3.5.7/src/Cms/Field.php#L144-L147

distantnative:
The method can't check for either empty string or empty array no matter what field type?

bastianallgeier (approved):
I think it's perfect like that. So far we always tried to handle panel field specific stuff in the panel field and not the content field implementation.

afbora:
> The method can't check for either empty string or empty array no matter what field type?

Blocks and layout fields store empty data as `[]` string. It is not possible to know if it is the text `[]` or an empty array.

> So far we always tried to handle panel field specific stuff in the panel field and not the content field implementation.

By the way, it would be great if we could refactor the field classes.

For example, the isEmpty() method and the new FieldClass class only work for the block and layout field:

**\Kirby\Cms\Field::isEmpty()**
https://github.com/getkirby/kirby/blob/master/src/Cms/Field.php#L146

**\Kirby\Form\Field::isEmpty()**
https://github.com/getkirby/kirby/blob/master/src/Form/Field.php#L324

**\Kirby\Form\FieldClass::isEmptyValue()**
https://github.com/getkirby/kirby/blob/master/src/Form/FieldClass.php#L178

PR #3449, opened 2021-06-19
Reveal the heading node for writer field

afbora:
# Status: 🚀 Ready for Review

## Describe the PR

- Reveals heading node
- New heading levels h4, h5, h6
- Using new heading icons
- Improved active node
- Improved switching between nodes
- New `paragraph` icon

image (26)

### Features
- Heading node for writer field with new levels (h4, h5, h6)
- New `paragraph` icon

## Breaking changes

None

## Related issues/ideas

- Implements https://kirby.nolt.io/255

## Ready?

- [x] CI checks pass

## When merging


panel/src/components/Writer/Toolbar.vue
```diff
@@ -1,9 +1,12 @@
 <template>
   <div class="k-writer-toolbar">
-    <k-dropdown v-if="Object.keys(nodeButtons).length > 1 && activeNode" @mousedown.native.prevent>
+    <k-dropdown v-if="Object.keys(nodeButtons).length" @mousedown.native.prevent>
       <k-button
-        :icon="activeNode.icon"
-        class="k-writer-toolbar-button k-writer-toolbar-nodes"
+        :icon="activeNode.icon || 'wand'"
```
distantnative: Maybe we use https://getkirby.com/docs/reference/panel/icons/title as fallback here as well?

afbora: I thought title, but it can be assumed that the active node is heading. That's why I wanted it to be a different icon.  May be `headline`?

distantnative: I like `headline`!

distantnative:
Looks good to me. But will leave the Gina review to @bastianallgeier since I'm not super familiar with the writer field and might overlook something

bastianallgeier:
This is fantastic, but wouldn't we also need a "Text" or "Paragraph" item in the dropdown to be able to revert back to a regular paragraph?

bastianallgeier:
Ahh, now I see. When you click the format again, it get's removed. That's nice.

bastianallgeier:
I just tested it on a local installation, but I don't think it's fully ready for 3.5.7 yet. 

There are a few things we should work out first: 

- After testing it, I really think we need a paragraph button
- Switching from list item to headline and vice versa is not possible. This is some weird ProseMirror logic and I think we need to fix this some other way. Otherwise it feels completely broken. 
- I'm not happy with the T icon yet. Let's find something better.

afbora:
> After testing it, I really think we need a paragraph button

As far as I can see, converting to paragraph in ProseMirror is clicking the same button again.

> Switching from list item to headline and vice versa is not possible. This is some weird ProseMirror logic and I think we need to fix this some other way. Otherwise it feels completely broken.

I don't think this is very important because converting the list element into heading didn't make much sense to me 🙂 There is already the same situation in ProseMirror examples. You cannot convert to heading while the list item is selected.

You can test from below:
https://prosemirror.net/examples/basic/

> I'm not happy with the T icon yet. Let's find something better.

Now that I've looked at our icons again, I think this is the most appropriate icon we have. Or we can add a new icon set.

> but I don't think it's fully ready for 3.5.7 yet.

I know there are deficiencies, but I think these deficiencies are caused by ProseMirror. 

@bastianallgeier You know better, since the implementation of these codes belongs to you first. Feel free to move 3.6 👍

afbora:
The PR will be rebase when 3.5.7 merged into develop

distantnative:
@afbora Would be great if you could make these status updates in the initial post. That way one can always see at a glance the current state of the PE, without scrolling through all comments

afbora:
I've rebased into develop and ready review again. 
Bastian had some thoughts about the feature. 
Let's clear it up and move on.

afbora:
Moving next release because there is no feedback yet.

afbora:
### Summarize

- Reveals heading node
- New heading levels h4, h5, h6
- Using new heading icons
- Improved active node

_Note: Solution for switching between nodes: Hide list items when heading node is selected and hide heading items when list node is selected._

bastianallgeier:
After checking out the Tiptap examples as a reference, I realized that they have the exact same problem there. The main difference is that they use a different interface. 

By listing the different formats / node types in the dropdown, each item in the dropdown seems to be a valid alternative for the others. That's why it feels like a bug when you want to switch from headline to list and it doesn't work. Maybe we need to work on the toolbar instead to solve this. 

In Tiptap's and Prosemirror's examples you have something like 

```
Paragraph | H1 H2 H3 | UL OL
```

In this case the list icons + paragraph icons would be active when you are in a list. 

I think it's still not perfect. I much prefer the dropdown with all format types. But it does at least feel less broken.

afbora:
Do you mean like that? 👇 

**When empty or paragraph selected**

Dropdown options:
- paragraph
- headings
- list items


**When headings selected**

Dropdown options:
- paragraph
- headings

**When list items selected**

Dropdown options:
- paragraph
- list items

bastianallgeier:
Oh, now I understand what you meant before. That could indeed be another way to solve it!! Sounds actually pretty good to me.

bastianallgeier:
We could also think about always listing all formats, but disabling those which are not available. This could be a stronger indicator that i.e. converting a list to a headline isn't supported because it doesn't make sense. Then it gets clearer that you need to convert it to a paragraph first.

afbora:
I like the disabled node idea, this seems more clear and directive from the user's point of view.

Can I get your opinion on new h4, h5, h6 levels and icons?
Also, should we add a new P icon for the paragraph?

bastianallgeier:
The styling for h4, h5 and h6 need to be improved, but that's just a minor detail. I think it's good to include them all. 

A P icon makes sense. I tried with the text icon as a first solution, but it doesn't look right. The P for paragraph is needed for a better understanding

bastianallgeier:
Screenshot 2021-09-20 at 13 25 40

```
<path d="M6,0V0A4,4,0,0,0,6,8v8H8V2h3V16h2V2h2V0Z" />
```

afbora:
### Summarize

- Reveals heading node
- New heading levels h4, h5, h6
- Using new heading icons
- Improved active node
- Improved switching between nodes
- New `paragraph` icon

image (26)

afbora:
Should we add a divider between paragraph, headings and lists?

bastianallgeier:
@afbora NICE! 

yes, a divider would be good.

afbora:
@bastianallgeier Divider added. I would like to add the divider in a more systematic way, like a button, but I couldn't.

bastianallgeier:
It's working really well now. I think your divider code is totally fine. There's just one small issue. When you don't define the heading nodes or list nodes, the dividers still stay there and don't make sense.
Screenshot 2021-09-23 at 12 12 39

afbora:
You're right. I've implemented.

bastianallgeier (approved):

bastianallgeier:
This was a super complicated project. Thanks for your patience and all the hard work ❤️

MaluNoPeleke:
Looking forward to testing it in the next beta!
Will the headings also be recognized automatically by copy & paste (if I copy a paragraph with headings they will keep the formatting after pasting it)?
This isn't the case with beta-1

afbora:
Yes, headings will be recognized in copy paste.

PR #3450, opened 2021-06-19
Support alternative thumb file formats

distantnative:
### Feature
- Support for alternative thumb file formats (e.g. `webp` or `avif`)
- Either set as global option `'thumbs.format' => 'webp'` or pass as option:
```php
$image->thumb(['width' => 200, 'format' => 'webp'])

$image->srcset([
  ['width' => 200, 'format' => 'webp'],
  ['width' => 200, 'format' => 'jpg']
])
```

- GDLib/SimpleImage driver currently only supports `webp`, ImageMagick supports `webp` and `avif`

## Related issues/ideas

- Implements https://kirby.nolt.io/160

## Ready?

- [x] CI checks pass

## When merging


src/Cms/FileModifications.php
```diff
@@ -191,6 +191,13 @@ public function thumb($options = null)
             return $this;
         }
 
+        // fallback to global config options
```
distantnative, line 194: I'm unsure myself if this is at the right location.
First I had it only in the `thumb` core component, but then the `file::version` component creates a wrong filename with wrong extension. So we somehow need to pass that option to `file::version` as well.

bastianallgeier (commented):
It looks really good in general. But it also shows how weird this entire setup is. The elephant in the room is probably if we should rip this entire shit out and redo it from scratch, or if we should just get this feature done.

distantnative:
I would redo it eventually, but not sure if 3.6 is feasible for that.

Given how little changes are needed with the current setup, if we can't do the big step in 3.6, I would still do this smaller one given the demand for this support.

bastianallgeier (dismissed):
If we redo it later anyway, I wouldn't bother about the position of the format fallback in your code. I think the rest really looks good to me. Could you take care of fixing the tests?

bastianallgeier (dismissed):

distantnative:
Hmmm... not sure why my one test is actually not asserting anything

distantnative:
@bastianallgeier got it, but need another approval

bastianallgeier (approved):

PR #3451, opened 2021-06-20
New vertical alignment of the dropdown menu: `dropup` 

### LINKED ISSUE #3080: [Blocks] UI: Context menu obscured by Revert/Save bar
**Describe the bug**  
The orange revert/save bar overlaps the context menu for the last block element on a page. This makes it impossible to delete the block without saving or reverting changes. 

**To Reproduce**  
1. Add a bunch of blocks to a page
2. Make a change so the orange revert/save bar appears at the bottom of the page
2. Click on the ··· context menu icon of the last block item

**Expected behavior**  
Expected behaviour would be a bottom margin that prevents the context menu from sitting underneath the bar so the delete/trash icon is accessible.

**Screenshot**  
image

**Kirby Version**  
3.5.0

afbora:
## Describe the PR

I fixed the issue with a new javascript based feature that named `dropup` as @bastianallgeier [mentioned in the forum](https://forum.getkirby.com/t/bug-field-s-action-menu-items-can-be-partially-hidden/22512/5?u=byybora).

### How does dropup work?
Dropup is activated if the last item of the dropdown menu overflows the visible screen.

dropup

### Enhancements
- New vertical alignment of the dropdown menu: `dropup` 

### Fixes
- Fixed overflowing dropdown menu #3080 

## Breaking changes

None

## Related issues/ideas

- Fixes #3080 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Navigation/DropdownContent.vue
+++ b/panel/src/components/Navigation/DropdownContent.vue
@@ -1,5 +1,10 @@
 <template>
-  <div v-if="isOpen" :data-align="align" class="k-dropdown-content">
+  <div
+    v-if="isOpen"
+    :data-align="align"
+    :data-dropup="dropup"
+    class="k-dropdown-content"
+  >
     <!-- @slot Content of the dropdown -->
     <slot>
       <template v-for="(option, index) in items">
@@ -29,7 +34,7 @@ export default {
   props: {
     options: [Array, Function],
     /**
-     * Aligment of the dropdown items
+     * Alignment of the dropdown items
      * @values left, right
      */
     align: {
@@ -39,9 +44,10 @@ export default {
   },
   data() {
     return {
-      items: [],
       current: -1,
-      isOpen: false
+      dropup: false,
+      isOpen: false,
+      items: []
     };
   },
   methods: {
@@ -81,6 +87,7 @@ export default {
         this.items = items;
         this.isOpen = true;
         OpenDropdown = this;
+        this.onOpen();
         /**
          * When the dropdown content is opened
          * @event open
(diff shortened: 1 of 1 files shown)


panel/src/components/Navigation/DropdownContent.vue
```diff
@@ -29,16 +34,25 @@ export default {
...
      */
     align: {
       type: String,
       default: "left"
+    },
+    /**
+     * Vertical alignment of the dropdown menu
+     * @values true, false
+     */
+    dropup: {
```
distantnative: Does a prop make a lot of sense? Since the component is now "magically" taking care of this anyways, I don't think we need a prop to set the behaviour from outside the component, do we?

afbora: I thought users might want to enable and use dropup by default.  Something like a footer menu.  I think it's something extreme 🙂

lukasbestle: Maybe it should then be possible to have `true|false|null` to be able to explicitly disable the dropup as well. `null` would be the default (for auto).

afbora: I've removed the excess prop.

afbora: > Maybe it should then be possible to have true|false|null to be able to explicitly disable the dropup as well. null would be the default (for auto).

@lukasbestle Sorry, I saw your comment now. Should I revert it?

lukasbestle: I'm fine with either implementation, but I agree with you that the prop to control the behavior could be useful. Maybe we could go ahead without the prop for now and then see if we need it later. Or what do you think @distantnative?

bastianallgeier: I think it's better to keep this as a data key and not reintroduce a prop. Aligning the dropdown left right or center makes sense as a manual decision. But I agree with @distantnative that the vertical direction should be handled automatically.

bastianallgeier (approved):

PR #3463, opened 2021-06-26
Throw exception for non-Closure Panel area

distantnative:


src/Panel/Panel.php
```diff
@@ -88,9 +88,11 @@ public static function areas(): array
 
         // load plugins
         foreach ($kirby->extensions('areas') as $id => $area) {
-            if (is_a($area, 'Closure') === true) {
-                $areas[$id] = static::area($id, (array)$area($kirby));
+            if (is_a($area, 'Closure') === false) {
+                throw new Exception(sprintf('Panel area "%s" must be defined as a Closure', $id));
```
lukasbestle: Why do we use both the `\Exception` and `Kirby\Exception\*` in this class? I think this should be an `Kirby\Exception\InvalidArgumentException`.

The second use of `\Exception` is in the `link()` method. Is there a specific reason for this?

bastianallgeier: I agree with @lukasbestle that we should use our own exceptions. The path to translated strings is easier that way. I think the main reason is simplicity while working on new features. Our Exceptions are a bit more challenging. I always need to remind myself again how I setup the exception correctly.

bastianallgeier: ✅

bastianallgeier (approved):

Issue #3197, opened 2021-03-07
Missing <hr> separator block


yandodov:
**Describe the bug**  
I'm migrating from Kirby Editor to Blocks and there's no horizontal rule block. After the data is converted to Blocks JSON, the `hr` block is still there:

![image](https://user-images.githubusercontent.com/5570098/110235542-3d1b6100-7f39-11eb-84e1-1d904db5e049.png)

...but there's no default snippet for it and it also doesn't render as a separator in the panel:

![image](https://user-images.githubusercontent.com/5570098/110235555-5b815c80-7f39-11eb-8d05-37b18f47dc57.png)

**To Reproduce**  
1. Create a page with the Editor field
2. Put an `hr` block in that field
3. Convert the Editor field to Blocks field as described in the [docs](https://getkirby.com/docs/reference/panel/fields/blocks#editor-migration)

**Expected behavior**  

1. You should see a separator in the Kirby Panel
2. There should be an `<hr />` tag in the rendered HTML

**Kirby Version**  
3.5.3.1

afbora:
@bastianallgeier how do you think the solution should be? Should we create a new block type named `hr/gap/separator/line` or put `<hr>` inside the text block? By the way, `<hr>` does not work in the text block (writer field) for some reason, it also throws it into the paragraph `<p><hr></p>`. Alternatively, we can use the `markdown` block. What do you think?

bastianallgeier:
I think it would make sense to add a new line block. This would open new options to add custom styling for such gaps/lines.

bastianallgeier:
✅

PR #3476, opened 2021-07-02
New `line` block

distantnative:
I am pretty sure I haven't found all the places where support needs to be added, but maybe this is a start.

## Issues
- [x] Suppress drawer: can we have field-less blocks?

## Related issues/ideas

- Implements #3197


afbora:
Is that same with #3197?

distantnative:
Yes, haven't tested if that concrete migration already works. But that's the goal.

bastianallgeier (approved):

PR #3480, opened 2021-07-03
Add `use` statements for `F` class

distantnative:


afbora (commented):
Also need for `\Kirby\Toolkit\Tpl` class.

lukasbestle (approved):

lukasbestle:
I wonder if there's a way to automatically add these imports using a static analyzer or at least flag the missing imports for us. I'm pretty sure we have more of them (also for other classes), but it's a pain to track them down manually.

afbora:
> I wonder if there's a way to automatically add these imports using a static analyzer or at least flag the missing imports for us. I'm pretty sure we have more of them (also for other classes), but it's a pain to track them down manually.

PHPStorm can do this with `Inspect Code`. I'll check.

PR #3492, opened 2021-07-05
Clean up `k-inside`

distantnative:
## Describe the PR

- Don't use scrolling div for Panel, but rather fixed topper (https://kirby.nolt.io/265)
- Fix z-index order: dialogs above dropdowns
- Move fatal dialog to own component (`Misc/Fatal.vue`)
- Move CSS to dedicated components
  - FormButtons
  - Icons
- Move Panel-specific CSS variables to variables CSS file
- Remove unused `dragging` data key (we use the store instead)
- Clean up remaining CSS

## TODO 🚜
- [x] Scroll lock for background when dialog is open is broken now
- [x] Page section: options dropdown dots on top of dialogs

## Related idea
https://kirby.nolt.io/265

### DIFF
--- a/panel/src/components/Forms/FormButtons.vue
+++ b/panel/src/components/Forms/FormButtons.vue
@@ -271,6 +271,13 @@ export default {
 </script>
 
 <style>
+.k-form-buttons {
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  z-index: var(--z-navigation);
+}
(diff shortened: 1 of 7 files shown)


bastianallgeier (approved):

bastianallgeier:
Just saw the milestone. But isn't it fine to put it in the next release? Should I revert it?

distantnative:
Haha only wanted to take out the time pressure for reviewing here. Fine if already included in next alpha.

PR #3497, opened 2021-07-05
Remove forgotten deprecated code from Vuex store

distantnative:
We decided to remove all other old Vuex code but forgot these few lines


afbora:
@distantnative So also should we remove the `$store.heartbeat` lines from [/panel/src/store/modules/heartbeat.js](https://github.com/getkirby/kirby/blob/fix/deprecated-cruft/panel/src/store/modules/heartbeat.js) module?

distantnative:
Im actually considering removing the heartbeat store already in 3.6

afbora (approved):

PR #3502, opened 2021-07-07
3.5.7.1 changes -> `develop`

afbora:
## Describe the PR

Only version (composer) change commit is not included from https://github.com/getkirby/kirby/pull/3500


lukasbestle:
@afbora I think you did it via cherry-picking, correct? The disadvantage is that it could cause merge conflicts down the road as the original commits from `3.5.7.1` are not actually contained in `develop`, just copies of them.

So I instead merged `master` into this branch. This caused merge conflicts immediately, so I could then fix or discard the changes to adapt to `develop`. The end result is the same, but it tells Git more about the merge.

lukasbestle (approved):

afbora:
Yes, I moved it with cherry-picking. I didn't think it would be a problem as the develop branch would merge into all kinds of master when it was done. Understood 👍 

So where (in which scenarios) should cherry-picking be used to avoid problems?

lukasbestle:
@afbora I use cherry-picking mostly for these two scenarios:

- For creating hotfix releases. E.g. a PR was already merged into `develop` but there needs to be a new release based off `master` immediately. Then the commit from `develop` can be cherry-picked onto a new branch.
- If a too large PR gets closed, but some of the commits should still be used. These can then be cherry-picked into a new PR branch.

afbora:
Thank you. I understood properly now. I think I did it in the opposite case of the first scenario 🙈

PR #3504, opened 2021-07-07
More code quality fixes

afbora:
# Status: Ready for Review 🚀 

## Describe the PR

Some of the quality guidelines from the PHPStorm IDE's report.

### Enhancements

- Improved code quality based on PhpStorm Inspect Code Report

## Breaking changes

None

## Related issues/ideas

None

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -27,12 +27,12 @@ function asset(string $path)
 /**
  * Generates a list of HTML attributes
  *
- * @param array $attr A list of attributes as key/value array
- * @param string $before An optional string that will be prepended if the result is not empty
- * @param string $after An optional string that will be appended if the result is not empty
- * @return string
+ * @param array|null $attr A list of attributes as key/value array
+ * @param string|null $before An optional string that will be prepended if the result is not empty
+ * @param string|null $after An optional string that will be appended if the result is not empty
+ * @return string|null
  */
-function attr(array $attr = null, $before = null, $after = null)
+function attr(?array $attr = null, ?string $before = null, ?string $after = null): ?string
 {
     if ($attrs = Html::attr($attr)) {
         return $before . $attrs . $after;
@@ -55,10 +55,10 @@ function collection(string $name)
 /**
  * Checks / returns a CSRF token
  *
- * @param string $check Pass a token here to compare it to the one in the session
+ * @param string|null $check Pass a token here to compare it to the one in the session
  * @return string|bool Either the token or a boolean check result
  */
-function csrf(string $check = null)
+function csrf(?string $check = null)
 {
     $session = App::instance()->session();
 
@@ -175,7 +175,7 @@ if (function_exists('e') === false) {
  * @param string $context Location of output (`html`, `attr`, `js`, `css`, `url` or `xml`)
  * @return string Escaped data
  */
-function esc($string, $context = 'html')
+function esc(string $string, string $context = 'html'): string
 {
     if (method_exists('Kirby\Toolkit\Escape', $context) === true) {
         return Escape::$context($string);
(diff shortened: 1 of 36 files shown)


src/Panel/Panel.php
```diff
@@ -450,7 +450,7 @@ public static function setLanguage(): ?string
         $kirby = kirby();
 
         // language switcher
-        if ($kirby->options('languages')) {
+        if ($kirby->option('languages')) {
```
afbora: @distantnative This change weirdly broke the tests. `options()` method doesn't accept any params and I fixed it. But i have no idea why fails? Do you have an idea?

distantnative: Maybe we should move this one to its own PR as it seems to be about an actual bug and not just code quality? Would be easier to look/work on it in its own PR.

afbora: I've created own PR to look closer #3505

config/helpers.php
```diff
@@ -55,7 +55,7 @@ function collection(string $name)
 /**
  * Checks / returns a CSRF token
  *
- * @param string $check Pass a token here to compare it to the one in the session
+ * @param string|null $check Pass a token here to compare it to the one in the session
  * @return string|bool Either the token or a boolean check result
  */
 function csrf(string $check = null)
```
bastianallgeier: Shouldn't it be …

```suggestion
function csrf(?string $check = null)
```

I'm not sure though if the func_get_args method will still work properly.

config/helpers.php
```diff
@@ -876,23 +878,23 @@ function twitter(string $username, string $text = null, string $title = null, st
 /**
  * Shortcut for url()
  *
- * @param string $path
+ * @param string|null $path
  * @param array|string|null $options
  * @return string
  */
-function u(string $path = null, $options = null): string
+function u(?string $path = null, $options = null): string
```
bastianallgeier, line 885: Why not? …

```suggestion
function u(?string $path = null, ?$options = null): string
```

config/helpers.php
```diff
@@ -876,23 +878,23 @@ function twitter(string $username, string $text = null, string $title = null, st
...
 /**
  * Builds an absolute URL for a given path
  *
- * @param string $path
+ * @param string|null $path
  * @param array|string|null $options
  * @return string
  */
-function url(string $path = null, $options = null): string
+function url(?string $path = null, $options = null): string
```
bastianallgeier, line 897: Same here …

```suggestion
function url(?string $path = null, ?$options = null): string
```

config/helpers.php
afbora: You right. I've tested and works same. `func_num_args()` returns still `0` when no argument passed.

afbora: ✅

src/Database/Query.php
```diff
@@ -710,7 +710,7 @@ protected function execute($sql, array $params = [])
             $this->database->fail();
         }
 
-        $result = $this->database->execute($sql['query'], $sql['bindings'], $params);
```
bastianallgeier, line 713: Why is this being removed?

bastianallgeier (commented):
There are tons of really nice enhancements in here. Awesome! If we can discuss the last remaining ones, I'm happy. But maybe @lukasbestle and @distantnative should have a quick look at it as well.

src/Database/Query.php
afbora: Because `execute()` method accepts only two arguments, am i wrong about called `execute()` method?

https://github.com/getkirby/kirby/blob/develop/src/Database/Database.php#L467

config/helpers.php
afbora: Is this valid syntax, so without type?

bastianallgeier: Oh wow,  you are completely right. Of course not. Forget my comment :)

src/Database/Query.php
bastianallgeier: You are right again!

bastianallgeier (approved):

distantnative (approved):
LGTM as well!

PR #3505, opened 2021-07-08
Fix setting language method for panel

afbora:
## Describe the PR

Fixes setting language method for panel.
Also we need to track down the failing tests.

## Release notes

None

## Breaking changes

None

## Related issues/ideas

None

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Panel/Panel.php
+++ b/src/Panel/Panel.php
@@ -447,13 +447,13 @@ class Panel
      */
     public static function setLanguage(): ?string
     {
-        $kirby = kirby();
+        $kirby   = kirby();
+        $session = $kirby->session();
 
         // language switcher
-        if ($kirby->options('languages')) {
-            $session  = $kirby->session();
+        if ($kirby->option('languages')) {
             $sessionLanguage = $session->get('panel.language', 'en');
-            $language = get('language') ?? $sessionLanguage;
+            $language        = get('language') ?? $sessionLanguage;
 
             // keep the language for the next visit
             if ($language !== $sessionLanguage) {
--- a/src/Panel/Panel.php
+++ b/src/Panel/Panel.php
@@ -447,11 +447,11 @@ class Panel
      */
     public static function setLanguage(): ?string
     {
-        $kirby   = kirby();
-        $session = $kirby->session();
+        $kirby = kirby();
 
         // language switcher
         if ($kirby->option('languages')) {
+            $session         = $kirby->session();
             $sessionLanguage = $session->get('panel.language', 'en');
             $language        = get('language') ?? $sessionLanguage;
 
--- a/src/Cms/System.php
+++ b/src/Cms/System.php
@@ -132,6 +132,13 @@ class System
             throw new PermissionException('The accounts directory could not be created');
         }
 
+        // init /site/sessions
+        try {
+            Dir::make($this->app->root('sessions'));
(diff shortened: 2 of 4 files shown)


afbora:
@distantnative I've fixed the issue with separate panel language setting in https://github.com/getkirby/kirby/pull/3505/commits/a4046cc0861b3c7f8da77e8cc4ab1b7fdd683d24. Because the panel language must be set whether it is multilingual or not. I'm not sure about fixing and need your review.

distantnative:
I'll loop in @bastianallgeier since he wrote that code part:
As we discussed lately, I am really unsure what language this method is supposed to set. The language of the translated content or of the UI interface elements?

bastianallgeier:
@distantnative it's supposed to set the content language. The setter for the interface translation is setTranslation.

afbora:
@distantnative What should we do?

distantnative:
OK, then I don't understand what bug this is fixing. Cause if `setLanguage` is about the content language, we only should set it if `languages` option is active, right?

src/Panel/Panel.php
```diff
@@ -450,19 +450,18 @@ public static function routesForViews(string $areaId, array $area): array
      */
     public static function setLanguage(): ?string
     {
-        $kirby = kirby();
+        $kirby           = kirby();
+        $session         = $kirby->session();
```
afbora: I found a new clue regarding the issue. Maybe this will give you an idea. Actually all the error is related to this line. Just executing this line in the method is enough (outside the if condition). When I remove this line, the tests fail. What happens when the `$kirby->session()` method is called?

afbora: It's not a solution but you can see how it works. 
`$session = $kirby->session();` must be outside the if condition in this case (Related commit https://github.com/getkirby/kirby/pull/3505/commits/5fce251c24c7cbcd524991fabfbf0004111b89e6)
So why? I don't know 🙂

bastianallgeier: How weird is that? @lukasbestle do you have any idea?

afbora: Pinging @lukasbestle to get any idea 🙂

lukasbestle: I tracked it down to the `tmp/site/sessions` directory being created by `$kirby->session()`. If I put the following in [this line](https://github.com/getkirby/kirby/blob/develop/tests/Panel/Areas/AreaTestCase.php#L179), then everything works again:

```php
Dir::make($this->tmp . '/site/sessions', true);
```

No idea what in the area tests depends on an empty session directory though. 🤯

afbora: @lukasbestle I guess found the issue.

Follow/track `$system->isOk()` in following line. This is about writable `sessions` directory. Otherwise redirecting `installation` path and test fail.

https://github.com/getkirby/kirby/blob/fix/panel-setting-language/src/Panel/Panel.php#L65

lukasbestle: Ah, of course! I think we should create the session dir in `$system->init()` like we do with `accounts`.

lukasbestle: ✅

lukasbestle (approved):
@afbora Thanks for rebasing and squashing!

distantnative (approved):

PR #3512, opened 2021-07-15
HTML tidy

bastianallgeier:
### Todos for this PR

- [x] @bastianallgeier Review TODO comments
  - [x] ["Escape XSS in query params"](https://github.com/getkirby/kirby/blob/feature/tidy/src/Toolkit/Dom.php#L572): Not needed
  - [x] [Attribute value escaping](https://github.com/getkirby/kirby/blob/feature/tidy/src/Toolkit/Dom.php#L587-L588): Not needed
- [x] @lukasbestle Clean up commit history
- [x] @lukasbestle Complete release notes

### Todos that can be tackled after merging

- [x] @bastianallgeier HTML sanitization in the writer field (on save)
  Must-have for 3.6.0
- [ ] @bastianallgeier Customizable sanitization/validation for any field in the blueprint (with configurable `Sane` type, `html` being the default)
  Would be great; if we do it, we should ideally do it in beta 2
- [ ] @bastianallgeier File uploads with sanitization instead of validation
  Optional, can be added later
  The idea is: In the file blueprint, there could be a new option like `sane: sanitize|validate|false`, probably with a better name. Or alternatively (or in addition) there could be a `config.php` option to define a global default for all file uploads. `validate` should always be the default if no other value is configured (`sanitize` can remove too much from files, so validation is much safer).
- [x] @lukasbestle More tests for `Sane\Html` and for `Toolkit\Dom`

### Breaking changes (core)

- The public static configuration attributes of the `Sane` classes have been renamed for consistency with the new `Toolkit\Dom` class
- The `Sane\Svg::$allowedNamespaces` attribute now expects namespace reference names as the array keys instead of full attribute names (so `xmlns:xlink` should become just `xlink` and `xmlns` should become an empty string)
- The `Sane\Svg::$allowedTags` attribute (before `$allowedElements`) now is an associative array with the allowed attributes for this tag as the value

### Features

- New `Toolkit\Dom` class that allows to parse, modify and sanitize DOM-based documents (XML and HTML)
- The `Sane` classes now have new `sanitize()` and `sanitizeFile()` methods that will apply the same rules as `validate()` but directly modify the input string/file so it is valid afterwards
- New `Sane\Html` class for validating and sanitizing HTML files or HTML snippets
- New `$file->sanitizeContents()` method

### Enhancements

- Added server-side sanitization in the writer field on field changes for extra security against XSS attacks
- Improved error messages for file validation/sanitization errors (`Sane` classes)
- The `Sane` classes now have proper XML namespace support so that custom namespace prefixes are properly detected


src/Sane/Html.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+class Html extends Handler
+{
+    /**
+     * HTML sanitization
+     *
+     * @param string $string
+     * @return string
+     */
+    public static function tidy(string $string): string
```
distantnative: It's nit picking, but wouldn't `sanitize()` be the more fitting name next to `validate()`?

lukasbestle: `Sane::sanitize()` sounds a bit duplicated though. Another alternative: `purify()`. But I'd be happy with either one of these three.

lukasbestle (commented):
Unfortunately I didn't get to do a full review yet, will do so tomorrow.

First thoughts related to the general structure:
- I think the `DOM` classes should use default values in `public static` props just like the other `Sane` classes. `array_merge` has the huge disadvantage that it's impossible to pass completely different values (each one needs to be set to `false` manually). Also users cannot globally configure the allowed and disallowed elements like with the other classes.
- `denied` -> `disallowed` for consistency
- IMO the `DOM` classes should be named `Dom` for consistency with `Toolkit\Xml` and `Toolkit\Html`.
- The `tidyFile()` method should write the tidied data back to the file and not just return it. Otherwise one could use `tidy(file_get_contents())`. Not doing it in-place feels weird and unexpected to me.

src/Sane/Html.php
distantnative: But it would be `Sane\Html::sanitize()`. Which I think reads actually quite nice. Of course there'd be `Sane\Sane::sanitize()` which is a bit quirky but so is `Sane\Sane` on its own (we have that in a few places - `Data\Data`, `Image\Image` - I think it's fine). But if you both prefer `tidy` it's fine. Was just a thought.

bastianallgeier: I had sanitize first and then went for tidy, exactly for that reason. `Sane\Html::sanitize()` would be nice indeed, but we never use it that way.

lukasbestle: Now that I think about it again, I do like `sanitize()`. The name is more clear as `tidy` could also be just an indentation fixer or minifier. `sanitize` has a clear connection to security.

lukasbestle:
@bastianallgeier What do you think about my general thoughts? If you agree, I can continue working on this over the next few days.

src/Sane/Html.php
bastianallgeier: TBH I don't mind enough :) Both names work.

bastianallgeier:
> I think the DOM classes should use default values in public static props just like the other Sane classes. array_merge has the huge disadvantage that it's impossible to pass completely different values (each one needs to be set to false manually). Also users cannot globally configure the allowed and disallowed elements like with the other classes.

That makes sense. 

> denied -> disallowed for consistency

+1 

> IMO the DOM classes should be named Dom for consistency with Toolkit\Xml and Toolkit\Html.

+1

> The tidyFile() method should write the tidied data back to the file and not just return it. Otherwise one could use tidy(file_get_contents()). Not doing it in-place feels weird and unexpected to me.

Not entirely sure about that. It feels a bit aggressive to overwrite it. Maybe we could have an overwrite param?

lukasbestle:
> Not entirely sure about that. It feels a bit aggressive to overwrite it. Maybe we could have an overwrite param?

To be honest I can't think of a use case where you wouldn't want to overwrite the file contents. E.g. on upload you definitely want to make sure that the uploaded file is safe, so overwriting in-place is the only way to ensure that.

Since we already have the `tidy()` method that returns the sanitized string, I feel like even the cases where overwriting is *not* wanted are already taken care of. If the dev doesn't want to overwrite for some reason, they can get the file contents and pass them to `tidy()` to get the sanitized string. It's two steps, but also much more explicit about what it does.

For me, that only leaves the case of an input file that should be left untouched with a separate sanitized output file. The method could look like this:

```php
public function tidyFile(string $file, $typeLazy = false, ?string $output = null)
```

Basically the method would write the sanitized contents to `$output` if given, otherwise it would overwrite the input `$file`.

Do you see a use case where returning the sanitized contents would be useful? If so, we could extend that `$output` argument to use `null` for returning, `true` for overwriting (default) and `string` for a custom path. But that's quite a complex method signature, so I'd love to avoid that if possible.

bastianallgeier:
I think I had a completely wrong view on the method. I was coming from my own test scenarios in which I want to load HTML and tidy it to display it afterwards. But when we want to keep it as close as possible to the Sane::validateFile method, you are right. Then we should overwrite it. I don't really think we should offer any additional option in this case at all.

lukasbestle:
OK, I agree. Simpler is better. :)

src/Sane/Html.php
distantnative: https://github.com/getkirby/kirby/pull/3524

lukasbestle:
I've implemented most of my feedback. Regarding the `public static` props I wasn't sure. Two options:

- Either we have the `public static` props in `Sane\Dom` directly.
- Or we have them in `Sane\Html` and pass them as an `$options` array to the `Sane\Dom` constructor. In this case the properties of `Sane\Dom` would be `protected`. We can remove the default value merging as now all options are always passed in from the handler class.

I like the second option best as it would also allow reusing the `Sane\Dom` class for SVG/XML sanitization and validation later.

lukasbestle:
Idea regarding the “use all applying handlers” feature:

We do that for validation, e.g. if you upload a `.xml` file that contains an SVG. What `Sane` does is to run both the XML and SVG validator. I think that still makes sense as it prevents attacks based on browsers autodetecting XML namespaces and actually treating such an XML file as SVG.

But for sanitization it would mean that both handlers remove parts that they don’t allow. In the XML/SVG case that would be fine, but with the combination SVG/HTML not much of the document would be left (if any).

So the idea is: In `sanitizeFile` we could still call `handlersForFile`. But then we check the count of the returned array. If it’s exactly one, we run that handler. If it’s zero, we return silently. But if it’s greater than one, we throw an Exception like `Cannot sanitize file as more than one handler applies: Kirby\Sane\Svg, Kirby\Sane\Html`.

bastianallgeier:
@lukasbestle I also like the second option! Sounds perfect to me. The concept for the sanitizeFile method also sounds good.

lukasbestle:
I've worked a lot on the `Toolkit\Dom` class over the weekend. Main changes:

- The `Sane\Dom` class is now integrated into `Toolkit\Dom` (we merged several classes that looked like this, so I figured it makes sense to build this one as one class from the start).
- The sanitization options are extended and are now passed directly to the `sanitize` method.
- Sanitization now collects all error messages in an array and returns that in the end. This way, the sanitization code can be used 1:1 for validation (see the `Sane\DomHandler::validate()` method).
- There's a new `DomHandler` class that can be used as a base class for all DOM-based `Sane` handlers. It should be possible to do most of the validation/sanitization directly in `Dom` by passing the correct options to it. This is also why there are several new options (from code that was moved from the `Sane\Xml` class). This makes the actual `Sane` classes really simple (see `Sane\Xml`).

The `SVG`/`SVGZ` handlers are not yet ported and the `Html` class still needs some love (should also extend `DomHandler`). I think the goal should be to move everything over to `DomHandler` so that all formats will support both sanitization and validation.

There are also three `TODO`s in `Toolkit\Dom` where your opinion @bastianallgeier would be really helpful. IMO the two existing lines with `TODO`s (`javascript:` and attr escaping) are not necessary and can be removed. Regarding the XSS in query strings I need input on the attack vector we need to protect against (I can't think of one at the moment).

Also the tests are still missing. The existing ones of the `Xml` class all pass (the `Svg` ones will also pass after the port), but we need new tests for `Html` as well.

All in all this seems to be a quite big project the more I look into it and I'm starting to think whether it makes sense to go back to a simple implementation for 3.6 (that does just the `Html` sanitization and leaves everything else unchanged). Then we (I) could pick up this rewrite afterwards in 3.6.x. But this would mean that the new `Html` class would do several things differently than the `Xml` and `Svg` classes. Also it wouldn't be possible to validate using the `Html` class, so the planned feature that will allow devs to choose between validation and sanitization of fields would not work yet.

I need to think about all of this myself. Just leaving this here so you can also think about it already.

src/Sane/Sane.php
```diff
@@ -109,8 +111,23 @@ public static function sanitizeFile(string $file, $typeLazy = false): void
...
+            case 1:
+                $handlers[0]->sanitizeFile($file);
+                break;
+            default:
+                // more than one matching handler;
+                // sanitizing with all handlers will not leave much in the output
+                $handlerNames = array_map('get_class', $handlers);
+                throw new LogicException(
+                    'Cannot sanitize file as more than one handler applies: ' .
+                    $handlerNames
```
distantnative: @lukasbestle isn't `$handlerNames` still an array here?

bastianallgeier:
@lukasbestle just a quick first thought. I agree that this is a big project and I don't think there's any shame in solving it as pragmatically as possible for 3.6 and go back to make various things better later.

lukasbestle:
@bastianallgeier What are your thoughts on the general code structure? I.e. "if we manage to do this in 3.6 (or in the future), should we even do it like that or go a different route?".

bastianallgeier:
I really like the class setup now. 

We should also base Parsely DOM handling on the Toolkit class in the next steps, but I'm quite happy with this setup already.

src/Sane/Sane.php
lukasbestle: Good catch! ✅

lukasbestle:
@bastianallgeier Status update:

- All `Sane` classes are now adapted to the new `Dom` class. XML and SVG sanitization fully works and has plenty of unit tests. 🎉
- The `Sane\Html` class seems to work fine in general, but misses the test cases from DOMPurify. I think the missing tests shouldn't be a blocker for merging this PR.
- The `Toolkit\Dom` class doesn't have its own tests so far, but is well tested by all the classes that depend on it. So IMO not a blocker either.
- I've updated the PR description with a todo list and a first set of release notes.

lukasbestle:
@bastianallgeier Is there a way to make a unit test for the writer field behavior? I think we should ensure that the sanitization can never get lost in the future by accident.

Aside: I think we shouldn't rely on aliases in our own code and always import the classes explicitly.

bastianallgeier:
@lukasbestle I've fixed/added both.

lukasbestle:
The Codecov check doesn't pass because of the missing tests for the `Dom` class. If we want a beta 2 soon, we can merge this PR with the current state. The class is well tested by the `Sane` classes. If there's time before beta 2, I'll add the tests today/tomorrow.

bastianallgeier:
Feel free to merge it like that and add the tests later if you think it's well covered otherwise.

lukasbestle (approved):
I'll approve this PR for now so it can be merged if needed. I didn't finish the tests yesterday, but I don't consider them a blocker. So IMO the PR can be merged with the failing Codecov check (temporarily disable branch protection rule).

Issue #3456, opened 2021-06-21
Finish work on new `Str::template()` implementation


lukasbestle:
**Can only be done once the `3.5.7` branch is merged back into `develop` (avoids merge conflicts because of changes in the implementation).**

- Add deprecation warning for `$fallback` string, `$start` and `$end` ✅
- Add tests for callback feature as well as `OptionsApi` escaping inside queries ✅

Related PRs: https://github.com/getkirby/kirby/pull/3279 https://github.com/getkirby/kirby/pull/3454

afbora:
✅

PR #3521, opened 2021-07-16
`Str::template()`: deprecate parameters

distantnative:
### Deprecations
- `Str::template()`: the parameters `$fallback`, `$start` and `$end` have been deprecated and throw a deprecation warning. Use instead an `$options` array with `fallback`, `start` and/or `end` keys as third parameter.

## Related issues

- https://github.com/getkirby/kirby/issues/3456 - but only one part

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -973,16 +973,25 @@ class Str
      *                                    - start: start placeholder
      *                                    - end: end placeholder
      *                                    A simple fallback string is supported for compatibility (but deprecated).
-     * @param string $start Placeholder start characters
-     * @param string $end Placeholder end characters
+     * @param string $start Placeholder start characters (deprecated)
+     * @param string $end Placeholder end characters (deprecated)
      *
-     * @todo Deprecate `string $fallback` and `$start`/`$end` arguments with warning in 3.6.0
      * @todo Remove `$start` and `$end` parameters, rename `$fallback` to `$options` and only support `array` type for `$options` in 3.7.0
      *
      * @return string The filled-in string
      */
     public static function template(string $string = null, array $data = [], $fallback = null, string $start = '{{', string $end = '}}'): string
     {
+        // @codeCoverageIgnoreStart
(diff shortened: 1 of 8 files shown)


src/Toolkit/Str.php
```diff
@@ -973,16 +973,25 @@ public static function substr(string $string = null, int $start = 0, int $length
...
      */
     public static function template(string $string = null, array $data = [], $fallback = null, string $start = '{{', string $end = '}}'): string
     {
+        // @codeCoverageIgnoreStart
+        if (
+            is_string($fallback) === true ||
+            $start !== '{{' ||
+            $end !== '}}'
+        ) {
+            deprecated('Str::template(): The $fallback, $start and $end parameters have been deprecated. Please pass an araay to the $options parameter instead with `fallback`, `start` or `end` keys: Str::template($string, $data, $options)');
```
afbora: I guess text should be improved:

- `araay` typo
- we don't have `$options` parameter yet (will in 3.7.0)

distantnative: Fixed the typo. On `$options`, I think we need to call it that already. In 3.7 we will officially rename it in the code, but we are treating it already like that. And would be weird to read "$fallback parameter has been deprecated, please use $fallback instead"

afbora: Not important. I just thought it would be confusing if the developers couldn't find the `$options` parameter.

afbora:
Do you have an idea about failing frontend tests?

afbora (approved):
LGTM

lukasbestle (approved):

lukasbestle:
@afbora Could you please work on the remaining todo in #3456:

> Add tests for callback feature as well as `OptionsApi` escaping inside queries

Doesn't have to be in this PR, but we shouldn't forget about it.

afbora:
@lukasbestle I was writing the remaining unit test exactly right now 🙂

PR #3522, opened 2021-07-17
`v-html` fixes

distantnative:
# Status: ready for review 🚀

- [x] New `Str::safeTemplate()` method (@distantnative)
- [x] First draft of `$model->toSafeString()` (@distantnative)
- [x] Review all `v-html` usage again (@lukasbestle)
- [x] Use `Str::safeTemplate()`/`->toSafeString()` wherever values end up in `v-html`
- [x] Finalize `$model->toSafeString()`
- [x] Review further fixes listed in Notion

# Release notes

## Breaking changes
- Several blueprint options that use the query syntax were updated to escape the placeholder values against raw HTML output that may lead to XSS attacks. HTML code directly in the query (like `This is <strong>{{ page.important }}</strong>`) still works as normal. If placeholders need to return HTML, you can use the new `{< site.myMethodWithHtml >}` syntax. In this case you need to ensure manually that the returned HTML code is safe. With the `{{ }}` syntax, Kirby performs the escaping for you.
- The `<k-box text="..." />` property is now rendered as plain text by default instead of as HTML code. For the previous behavior use `<k-box text="..." :html="true" />`.

## Features
- New `Str::safeTemplate()` method that escapes HTML from all regular query placeholders, allows HTML with new `{< site.myMethodWithHtml >}` syntax
- New `$model->toSafeString()` method that uses `Str::safeTemplate()` instead of `Str::template()`

## Enhancements
- The `toggle` field now supports the query syntax in the `text` property
- The `RadioInput` component now supports (escaped) HTML in the `info` property for radio options.

### DIFF
--- a/src/Form/OptionsApi.php
+++ b/src/Form/OptionsApi.php
@@ -7,7 +7,6 @@ use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Http\Remote;
 use Kirby\Http\Url;
-use Kirby\Toolkit\Escape;
 use Kirby\Toolkit\Properties;
 use Kirby\Toolkit\Query;
 use Kirby\Toolkit\Str;
@@ -90,11 +89,7 @@ class OptionsApi
     protected function field(string $field, array $data): string
     {
         $value = $this->$field();
-        return Str::template($value, $data, [
-            'callback' => function ($result) {
-                return Escape::html($result);
-            }
-        ]);
+        return Str::safeTemplate($value, $data);
     }
 
     /**
--- a/config/areas/files/dialogs.php
+++ b/config/areas/files/dialogs.php
@@ -3,6 +3,7 @@
 use Kirby\Cms\Find;
 use Kirby\Panel\Field;
 use Kirby\Panel\Panel;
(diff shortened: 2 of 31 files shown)


src/Form/OptionsApi.php
```diff
@@ -234,7 +229,7 @@ public function toArray(): array
      */
     public function url(): string
     {
-        return Str::template($this->url, $this->data());
+        return Str::safeTemplate($this->url, $this->data());
```
lukasbestle: @distantnative As far as I can tell, the `url()` method is only ever called from the `options()` method and there only ever for the URL or file to request the options from. So the URL doesn't seem to end up in the HTML, which means it probably (?) shouldn't be escaped for the HTML context. Maybe I'm missing something here.

distantnative: True, I think this was a too greedy Cmd+F for `Str::template`

lukasbestle:
@distantnative Is there anything else you had in mind regarding "Finalize `$model->toSafeString()`"? I've added support for the additional arguments, maybe you thought of something else?

distantnative:
Looks good to me! - but can't approve my own PR.

lukasbestle (approved):

Issue #3406, opened 2021-06-07
[3.6.0] Unit tests for the remaining Fiber changes


lukasbestle:
From the Codecov diff in #3327:

- [ ] `src/Panel/Panel.php` (at 97.82 %)
- [x] `config/areas/settings.php` (at 100 %)
- [ ] Verify that the whole Panel namespace has specific unit tests by only running these test classes (not the areas, which also cover the `Panel` namespace)

afbora:
#3390 outdated?

distantnative:
I think this one has the newer status/info, so I closed mine.

afbora:
The license line in the `config/areas/settings.php`  file actually looks untestable becasue of need an actual license. So I guess it should be ignored: `@codeCoverageIgnore`

https://github.com/getkirby/kirby/blob/develop/config/areas/settings.php#L15

bastianallgeier:
@afbora I agree. I was thinking if we could fake the license somehow. But this would mean that it's also fakeable in a real installation.

bastianallgeier:
✅

PR #3523, opened 2021-07-17
Increase `Kirby\Panel` unit test coverage

distantnative:
Contributes to https://github.com/getkirby/kirby/issues/3406

### DIFF
--- a/tests/Panel/FieldTest.php
+++ b/tests/Panel/FieldTest.php
@@ -283,6 +283,28 @@ class FieldTest extends TestCase
     }
 
     /**
+     * @covers ::slug
+     */
+    public function testSlug(): void
+    {
+        // default
+        $field = Field::slug();
+        $expected = [
+            'label' => 'URL appendix',
+            'type'  => 'slug',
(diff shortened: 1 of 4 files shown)


afbora:
Is it possible to cover following remaining 3 lines in `src/Panel/Panel.php`?

- https://app.codecov.io/gh/getkirby/kirby/compare/3523/tree/src/Panel/Panel.php#D1L371
- https://app.codecov.io/gh/getkirby/kirby/compare/3523/tree/src/Panel/Panel.php#D1L401
- https://app.codecov.io/gh/getkirby/kirby/compare/3523/tree/src/Panel/Panel.php#D1L412

distantnative:
I think there are more in the other classes as well. That's why this PR is only there to add some, wasn't able to write unit tests for all.

bastianallgeier:
I think it's fine when we break down such coverage bumps into smaller PRs and don't get too stressed about always covering 100%. Every additional covered line counts.

bastianallgeier (approved):

PR #3525, opened 2021-07-17
UI docs: explicitly document components

distantnative:
## Status: in progress 👷

- To be listed without warning in the docs, components have to be flagged explicitly as `@public`.
- Components flagged with `@internal` will be excluded completely from the docs.

### Fixes
- `k-progress`: fixed not updating when `value` prop changed

### Deprecated
- `<k-text align="right">` has been deprecated, use `<k-text align="end">` instead

### Refactoring
- Moved `panel/src/component/Blocks` to `panel/src/component/Forms/Blocks`
- Moved `panel/src/component/Writer` to `panel/src/component/Forms/Writer `


distantnative:
@bastianallgeier should we merge this for 3.6?

bastianallgeier:
@distantnative yep, looks good to me!

bastianallgeier (approved):

PR #3527, opened 2021-07-17
Automatic `api.csrf` via `panel.dev` option

### LINKED ISSUE #3363: [3.6.0] automatically set `auth.csrf` when `panel.dev` is set

distantnative:
## Status: ready for review 👀

 If `panel.dev` is set, sets `api.csrf` automatically to `dev`

## Related issues/ideas

- Fixes #3363

### DIFF
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -62,21 +62,26 @@ function csrf(?string $check = null)
 {
     $session = App::instance()->session();
 
-    // check explicitly if there have been no arguments at all;
+    // no arguments, generate/return a token
+    // (check explicitly if there have been no arguments at all;
     // checking for null introduces a security issue because null could come
-    // from user input or bugs in the calling code!
+    // from user input or bugs in the calling code!)
     if (func_num_args() === 0) {
-        // no arguments, generate/return a token
-
         $token = $session->get('kirby.csrf');
+
         if (is_string($token) !== true) {
             $token = bin2hex(random_bytes(32));
             $session->set('kirby.csrf', $token);
         }
 
         return $token;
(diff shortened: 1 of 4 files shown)


lukasbestle (commented):
To be honest I think the `$validate` argument is a bit weird. Now the method suddenly does two things. OK, it did it before, but IMO we should rather clean this up than add another argument.

As far as I can tell, we only ever use `$auth->csrf() === false` in our code. We do have two tests that check the string value, but why do we have these tests if we don't use this in practice so far? Very weird...

distantnative:
Not sure if I can follow you what exactly you suggest to do, I find `$auth->csrf() === false` in
- config/api/authentication.php (1x)
- config/api/routes/auth.php (2x)
- config/api/routes/system.php (1x)

We could of course move this part
```php
// check for a predefined csrf or use the one from session
$isDev       = $this->kirby->option('panel.dev', false) !== false;
$fromSession = $this->kirby->option('api.csrf', $isDev ? 'dev' : csrf());
```

to its own method. I don't have a good idea for a method name right now though.

lukasbestle:
I didn't make a suggestion so far, at least that's what I think. :D
What I meant: We never use the string return value of `$api->csrf()` anywhere, only `false` (= validate mode). So the string return value is something that was silently supported before but was never used anywhere.

distantnative:
@lukasbestle what do you think of putting it into this new method?

distantnative:
@lukasbestle ping :)

lukasbestle (approved):
That looks very simple and effective. 👍
Sorry for the huge delay reviewing this.

Issue #3514, opened 2021-07-15
PHP `t()` helper inconsistent with JS `$t()`


distantnative:
The JS helper returns the key if the i18n string cannot be found, the PHP helper just null.

afbora:
Which one is correct? Returning `null` or key?

bastianallgeier:
I tend to prefer returning null. But we have to be careful in both ways that we don't break something important. At least in the backend there are parts that depend on t() returning null.

afbora:
If we'll go for `null`, php `js()` helper doesn't affected but vue i18n `$t()` helper will affected. We can add new fallback param to `$t()` helper.

distantnative:
✅

PR #3528, opened 2021-07-17
JS `$t()`: return `null` when key doesn't exist

distantnative:
Fixes https://github.com/getkirby/kirby/issues/3514

### DIFF
--- a/panel/src/config/i18n.js
+++ b/panel/src/config/i18n.js
@@ -2,15 +2,15 @@ import { template } from "../helpers/string.js";
 
 export default {
   install(app) {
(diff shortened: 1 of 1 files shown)


afbora:
How about the new fallback parameter to be even more compatible and consistent with PHP `t()`?

distantnative:
Question is if the existing `data` parameter is already a difference that `fallback` as third parameter would still feel unfamiliar.

afbora:
Yes you are right then. However, it may be more useful to return fallback rather than return the always `null` as default.. If the use of fallback in the panel is rare, it may not be necessary. As it stands, I'm ok.

afbora (approved):

PR #3530, opened 2021-07-17
`file::version` component: wrong param type

distantnative:
`FileModification` is a trait and can therefore not be a type.
It is used by `Kirby\Cms\File` and `Kirby\Cms\Asset` - so these could be parameter types here.

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -70,7 +70,7 @@ return [
      * Adapt file characteristics
      *
      * @param \Kirby\Cms\App $kirby Kirby instance
-     * @param \Kirby\Cms\File|\Kirby\Cms\FileModifications $file The file object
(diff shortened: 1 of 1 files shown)


config/components.php
```diff
@@ -70,7 +70,7 @@
      * Adapt file characteristics
      *
      * @param \Kirby\Cms\App $kirby Kirby instance
-     * @param \Kirby\Cms\File|\Kirby\Cms\FileModifications $file The file object
+     * @param \Kirby\Cms\File|\Kirby\Cms\Asset $file The file object
```
afbora:
```suggestion
     * @param \Kirby\Cms\File|\Kirby\Filesystem\Asset $file The file object
```

afbora (approved):

PR #3531, opened 2021-07-17
DRY `Kirby\Cms\Auth`

### LINKED ISSUE #3498: Auth: duplicate code block
I think we could DRY this:
- https://github.com/getkirby/kirby/blob/develop/src/Cms/Auth.php#L98-L112
- https://github.com/getkirby/kirby/blob/develop/src/Cms/Auth.php#L509-L524

distantnative:
## Describe the PR

- Eliminates duplicate code blocks by introducing `Kirby\Cms\Auth::validateEmail()`

## Related issues/ideas

- Fixes #3498

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Cms/Auth.php
+++ b/src/Cms/Auth.php
@@ -95,21 +95,7 @@ class Auth
      */
     public function createChallenge(string $email, bool $long = false, string $mode = 'login')
     {
-        // ensure that email addresses with IDN domains are in Unicode format
-        $email = Idn::decodeEmail($email);
-
-        if ($this->isBlocked($email) === true) {
-            $this->kirby->trigger('user.login:failed', compact('email'));
-
-            if ($this->kirby->option('debug') === true) {
-                $message = 'Rate limit exceeded';
-            } else {
-                // avoid leaking security-relevant information
-                $message = ['key' => 'access.login'];
-            }
-
-            throw new PermissionException($message);
-        }
+        $email = $this->validateEmail($email);
 
         // rate-limit the number of challenges for DoS/DDoS protection
         $this->track($email, false);
@@ -493,6 +479,36 @@ class Auth
     }
 
     /**
+     * Ensure that email addresses with IDN domains are in Unicode format
+     * and that rate limit was not exceeded
+     *
+     * @param string $email
+     * @return string
+     *
+     * @throws \Kirby\Exception\PermissionException If the rate limit was exceeded
+     */
+    public function validateEmail(string $email): string
+    {
+        // ensure that email addresses with IDN domains are in Unicode format
+        $email = Idn::decodeEmail($email);
+
+        if ($this->isBlocked($email) === true) {
+            $this->kirby->trigger('user.login:failed', compact('email'));
+
+            if ($this->kirby->option('debug') === true) {
+                $message = 'Rate limit exceeded';
+            } else {
+                // avoid leaking security-relevant information
+                $message = ['key' => 'access.login'];
+            }
+
+            throw new PermissionException($message);
+        }
+
+        return $email;
+    }
+
+    /**
      * Validates the user credentials and returns the user object on success;
      * otherwise logs the failed attempt
      *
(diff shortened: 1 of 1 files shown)


src/Cms/Auth.php
```diff
@@ -493,23 +479,19 @@ public function status($session = null, bool $allowImpersonation = true)
...
-     * @return \Kirby\Cms\User
+     * @return string
      *
-     * @throws \Kirby\Exception\PermissionException If the rate limit was exceeded or if any other error occurred with debug mode off
-     * @throws \Kirby\Exception\NotFoundException If the email was invalid
-     * @throws \Kirby\Exception\InvalidArgumentException If the password is not valid (via `$user->login()`)
+     * @throws \Kirby\Exception\PermissionException If the rate limit was exceeded
      */
-    public function validatePassword(string $email, string $password)
+    public function validateEmail(string $email): string
```
afbora: Should we make this method `protected` since it is login specific? 🤔

distantnative: I actually thought about that - and probably yes. In the end, I followed `validatePassword` but maybe there is a reason for this one needed to be public?

distantnative: @lukasbestle will know best

src/Cms/Auth.php
```diff
@@ -493,23 +479,19 @@ public function status($session = null, bool $allowImpersonation = true)
     }
 
     /**
-     * Validates the user credentials and returns the user object on success;
-     * otherwise logs the failed attempt
+     * Ensure that email addresses with IDN domains are in Unicode format
+     * and that rate limit was not exceeded
```
lukasbestle:
```suggestion
     * Ensures that email addresses with IDN domains are in Unicode format
     * and that the rate limit was not exceeded
```

src/Cms/Auth.php
```diff
@@ -493,23 +479,19 @@ public function status($session = null, bool $allowImpersonation = true)
...
      *
      * @param string $email
-     * @param string $password
-     * @return \Kirby\Cms\User
+     * @return string
```
lukasbestle:
```suggestion
     * @return string The normalized Unicode email address
```

src/Cms/Auth.php
```diff
@@ -493,23 +479,19 @@ public function status($session = null, bool $allowImpersonation = true)
...
-     * @throws \Kirby\Exception\InvalidArgumentException If the password is not valid (via `$user->login()`)
+     * @throws \Kirby\Exception\PermissionException If the rate limit was exceeded
      */
-    public function validatePassword(string $email, string $password)
+    public function validateEmail(string $email): string
     {
         // ensure that email addresses with IDN domains are in Unicode format
         $email = Idn::decodeEmail($email);
 
-        // check for blocked ips
```
lukasbestle: I think this comment is quite helpful. Probably was removed because the first occurrence of this duplicated block didn't contain the comment for whatever reason.

src/Cms/Auth.php
lukasbestle: I agree we should make it `protected`. It's quite specific and of little use for devs as it doesn't fully validate the email (e.g. it doesn't check whether the user exists).

`validatePassword` is `public` because of use cases where you want to check a password without logging in (e.g. for logging in to third-party sites).

src/Cms/Auth.php
distantnative: Oh yes, good catch, will add it back

src/Cms/Auth.php
```diff
@@ -523,6 +506,25 @@ public function validatePassword(string $email, string $password)
...
+     * Validates the user credentials and returns the user object on success;
+     * otherwise logs the failed attempt
+     *
+     * @param string $email
+     * @param string $password
+     * @return \Kirby\Cms\User
+     *
+     * @throws \Kirby\Exception\PermissionException If the rate limit was exceeded or if any other error occurred with debug mode off
+     * @throws \Kirby\Exception\NotFoundException If the email was invalid
+     * @throws \Kirby\Exception\InvalidArgumentException If the password is not valid (via `$user->login()`)
```
afbora, line 522: It's not very important, but while you're in it; 
Do we need this? As standard, we only add Exceptions thrown from within the method to the doc blocks. 

For example, same for `createChallenge()` method and `PermissionException`..

lukasbestle: Ideally all exceptions should be documented, no matter where in the chain they are thrown. After all the user of this method doesn't know which methods are called internally.

The only issue so far has been that it's incredibly hard to find all exceptions and keep them up to date. That's why we generally don't do it at the moment, but I wouldn't remove already existing hints.

lukasbestle (approved):

PR #3535, opened 2021-07-17
Fix `null.yml` for blueprint in file view

### LINKED ISSUE #3533: [3.6.0] Blueprint placeholder shows `null.yml`
I think it's the case when a file doesn't have a template

Screen Shot 2021-07-17 at 20 00 48

distantnative:
Fixes #3533

### DIFF
--- a/src/Panel/File.php
+++ b/src/Panel/File.php
@@ -268,7 +268,7 @@ class File extends Model
             parent::props(),
             $this->prevNext(),
             [
-                'blueprint' => $this->model->template(),
+                'blueprint' => $this->model->template() ?? 'default',
                 'model' => [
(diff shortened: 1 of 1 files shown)


afbora:
Few notes:

1. I can't reproduce the issue of this PR for file or user even if deleted all file and user blueprints.
2. May be unrelated with this PR but is this normal that PDF and default file targeting `image.yml`

screen-capture-1122-sample-file pdf - Mægazine-localhost

distantnative:
Have you looked into the content `.txt` of the files that there is no `template`? Might have been added by the file section you used to upload?

afbora (approved):
Just like you said 💯

Issue #3343, opened 2021-05-22
Password reset: email should include site name not "Kirby Panel"


distantnative:
Currently the email reads

> You recently requested a password reset code for the Kirby Panel.

I think most users won't know what Kirby is. So making it `sitename Panel` or something like this would be better.

lukasbestle:
Yep, sounds good. I think we should move the site title logic from here to a method in `Site` or `System`:

https://github.com/getkirby/kirby/blob/b6cad90/config/api/models/System.php#L55-L61

Then we can use the same logic for the email text for password reset and login codes (both are affected the same way).

distantnative:
@lukasbestle sounds good, probably `System::site()` could work well? Do you want to get onto it since you're also more familiar with all the email templates for those auth methods?

lukasbestle:
What about `System::siteName()`? I think `System::site()` is too unspecific.

distantnative:
Then it should be `siteTitle` or even just `title`

bastianallgeier:
I like `System::title()`

bastianallgeier:
✅

PR #3537, opened 2021-07-17
`System::title()` and use it for email challenges

distantnative:
## Status: ready for review 👀

- The implementation of `System::title()` differs from https://github.com/getkirby/kirby/blob/b6cad90/config/api/models/System.php#L55-L61 as that code seemed flawed: the `catch` part could never be reached since `$site->blueprint()` always adds a fallback that is loaded with a title. Even more, the content value should be more important than the blueprint. That's fixed now.

### Enhancements
- Login emails (code and password reset) now include the site's title instead of a reference to Kirby

## Related issues/ideas

- Fixes https://github.com/getkirby/kirby/issues/3343

## Ready?

- [x] CI checks pass

## When merging


config/api/models/System.php
```diff
@@ -52,12 +52,8 @@
...
             return $system->toArray();
         },
-        'site' => function () {
-            try {
-                return $this->site()->blueprint()->title();
-            } catch (Throwable $e) {
-                return $this->site()->title()->value();
-            }
+        'site' => function (System $system) {
+            return $system->title();
```
lukasbestle, line 56: I think it's weird that we now have a `site` field that returns the content title with blueprint fallback and a `title` field that only returns the content title. Can we unify this somehow?

To be honest I don't fully understand what the `site` and `title` are used for respectively.

src/Cms/System.php
```diff
@@ -514,6 +514,18 @@ public function sessions(): bool
         return is_writable($this->app->root('sessions'));
     }
 
+    /**
+     * Return's the site's title as defined in the
+     * context file or `site.yml` blueprint
```
```suggestion
     * Returns the site's title as defined in the
     * content file or `site.yml` blueprint
```

config/api/models/System.php
distantnative: I think we used it in different ways in the Panel. I don't think we use these anymore with Fiber,  but I also didn't want to remove them from the API as another breaking change.

lukasbestle: IMO we should check this and deprecate one of them, ideally `site`. Keeping both around is really confusing.

distantnative: As far as I can tell, on the JS we only use `$api.system.install()` from the system API  - `get()` and `register()` are completely unused. So from the Panel's side, we can do pretty much what we want in the API System model.

distantnative: We might even move the `install()` call to Fiber as well and consider deprecating the whole System part of the API.

But maybe we move that to its own issue and leave it here as is.

lukasbestle: Your call, it's only something I noticed and where I feel like we shouldn't leave it like that in the long term.

distantnative: https://github.com/getkirby/kirby/issues/3541

src/Cms/System.php
```diff
@@ -514,6 +514,18 @@ public function sessions(): bool
...
+    /**
+     * Returns the site's title as defined in the
+     * content file or `site.yml` blueprint
+     *
+     * @return string
+     */
+    public function title(): string
+    {
+        $blueprint = $this->app->site()->blueprint()->title();
+        return $this->app->site()->title()->or($blueprint)->value();
```
bastianallgeier: It's probably never going to be performance-relevant, but I would still improve it a bit. 

```suggestion
        $site = $this->app->site();

        if ($site->title()->isNotEmpty()) {
          return $site->title()->value();
        }
        
        return $site->blueprint()->title();
```

bastianallgeier (commented):
The rest looks good to me

distantnative:
👍

bastianallgeier (dismissed):

bastianallgeier (dismissed):

bastianallgeier (approved):

PR #3539, opened 2021-07-18
Layouts: Smaller min-height

### LINKED ISSUE #3381: Panel: Layout row min-height consumes unnecessary space
**Describe the bug**  

This is not essential, but also rather trivial.

I will use the term "layout row" for a layout block.

When an empty layout row is added, a minimal space of `6rem` is reserved on `.k-layout-column` for (probably) easier interaction with the newly created row.

Screenshot from 2021-05-28 22-41-16

Now as this is defined per `min-height` and not updated, no matter what component is placed in the row/column, if a component is added to the layout that takes less space than the reserved space, we get this unnecessary space to the next layout row.

This creates a slight visual distortion which, in some cases, can be confusing for content editors. As an example, here I add a `h3` to a layout row, with two a row above and below featuring different content. The title now is closer to the layout section above than below, which in a non-significant way can create confusion, but even more just looks strange.

Screenshot from 2021-05-28 22-45-05

It's not the title field here that claims this space, but the layout row.

_(Mind, #3071 applies on the screenshots, cutting of my title texts. It's clearly written "Überschrift" and not "Uberschrift")_

**To Reproduce**  
Steps to reproduce the behaviour:
1. Go to a panel page that has a layout field
2. Create a layout block with one column
3. Add a title field set in the previously created layout block
4. Add other layout components below.

**Expected behaviour**  
After adding a block inside a layout that takes less vertical space than the minimal height of an empty layout block, the layout block should remove the `min-height`. In short: The `min-height` on `.k-layout-column` should only be set when all columns in a row are empty.

Probably the minimum you can reduce to still have space for the layout controls on the right. There are probably additional controls on the layout row on the right in specific situations that would reason for the min height of 6rem. But as long as they are not visible, the space isn't necessary.

Screenshot from 2021-05-28 23-08-25

In the Best case, the drag control for moving the rows up and down would be placed outside the layout row, similar to how sorting with collection lists work, to save additional vertical space. But that's probably a bit of a stretch in terms of scope.

**Screenshots**  
See description.

**Kirby Version**  
3.5.5

**Console output**  
–

 - OS: Pop_OS/Ubuntu
 - Browser: Firefox
 - Version: 89.0b15 (64-bit)

**Smartphone (please complete the following information):**  
Untested

**Additional context**

distantnative:
## Describe the PR

- Applies `min-height: 6rem` only to empty layout columns, `min-height: 4rem` to filled columns

### Fix
- Improved `min-height` for layout columns

## Related issues/ideas

- Fixes #3381

### DIFF
--- a/panel/src/components/Layouter/Column.vue
+++ b/panel/src/components/Layouter/Column.vue
@@ -44,7 +44,6 @@ export default {
   display: flex;
   flex-direction: column;
   background: var(--color-white);
-  min-height: 6rem;
 }
 .k-layout-column:focus {
   outline: 0;
@@ -55,7 +54,12 @@ export default {
   padding: 0;
   height: 100%;
   background: var(--color-white);
+  min-height: 4rem;
 }
+.k-layout-column .k-blocks[data-empty="true"] {
+  min-height: 6rem;
+}
+
 .k-layout-column .k-blocks-list {
   display: flex;
   flex-direction: column;


lukasbestle:
I think it's a bit weird that the dropdown button is still in the toolbar but the drag handle is on the outside. Alternatives:

- Get rid of the drag handle all-together and make the whole toolbar draggable
- Only display the drag handle if there is enough room in the layout
- Don't fix this at all (only affects single-line layouts; larger layouts with two or more lines have enough room anyway)

All of these have huge drawbacks, maybe there are better solutions we haven't thought of so far.

distantnative:
I think the least fix we could do is reducing the `min-height` to `4rem` when the column has any blocks.

lukasbestle:
That would indeed be the least controversial fix. :)

bastianallgeier:
I agree that we should only fix the min height and keep the drag handle where it was.

distantnative:
👍

distantnative:
Done

bastianallgeier (approved):

PR #3540, opened 2021-07-18
Assets: fix media thumbs for unprocessed asset files

### LINKED ISSUE #2467: Asset::mediaUrl() is broken
## Kirby Team Summary (22.12.2021)

Assets are correctly modified and moved to the media folder when you use one of the modification methods (resize or crop). This has been fixed after the issue has been posted. The original issue is therefore mostly resolved.

But the original is not moved to the media folder though. This is not a problem when you use `$asset->url()`. It will always point to the original URL. But `$asset->mediaUrl()` is broken. It tries to load the asset from the media folder instead of the original location. But our image API does not move originals into the media folder. 

## Proposed solution

- [ ] Fix `Asset::mediaUrl()` Review @distantnative's PR as a possible fix again: https://github.com/getkirby/kirby/pull/3540

----

## Original Issue by @rzschoch

**Describe the bug**  
The urls for assets generated via the asset helper are broken and not copied into the media folder. The issue also happens with file versions of an asset. 

**To Reproduce**  
```php
<img src="<?= asset('assets/speakers/speaker.jpg')->mediaUrl() ?>">
// image source url: http://localhost:8080/media/assets/assets/speakers/555072720-1582211687/speaker.jpg
// 404 response
```

or

```php
<img src="<?= asset('assets/speakers/speaker.jpg')->resize(600)->url() ?>">
// image source url: http://localhost:8080/media/assets/assets/speakers/555072720-1582211687/speaker-600x.jpg
// 404 response
```

**Expected behavior**  
I would expect that the asset is copied to the media folder as soon as it is requested for the first time via its media url.

**Kirby Version**  
3.3.4

**Additional context**  
This seems to be related to: https://github.com/getkirby/kirby/issues/2314 and https://github.com/getkirby/kirby/issues/1728

distantnative:
## Describe the PR

- `Media::thumb()` doesn't give up anymore if no job file for an asset is found but instead copies over the full source file
- `FileVersion->mediaUrl()` returns its `->url()` now since they are the same 

### Fixed
- Media files for assets (using `asset()`) without processing are working now

## Breaking changes

None to my knowledge

## Related issues/ideas

- Fixes #2467

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Cms/Media.php
+++ b/src/Cms/Media.php
@@ -131,7 +131,7 @@ class Media
                 $options = Data::read($job);
             } catch (Throwable $e) {
                 // no job file found, still run the job
-                // for publically accesible assets
+                // for publicly accessible assets
                 if (is_string($model) === true) {
                     $options = [
                         'filename' => $filename
--- a/src/Cms/Media.php
+++ b/src/Cms/Media.php
@@ -85,6 +85,30 @@ class Media
     }
 
     /**
+     * Returns the root path for the model
+     *
+     * @param \Kirby\Cms\App
+     * @param \Kirby\Cms\Model|string $model
+     * @param string $hash
+     * @return string
+     */
+    public static function root($kirby, $model, string $hash): string
+    {
+        // assets
+        if (is_string($model) === true) {
+            return $kirby->root('media') . '/assets/' . $model . '/' . $hash;
+        }
+
+        // parent files for file model that already included hash
+        if (is_a($model, '\Kirby\Cms\File')) {
+            return dirname($model->mediaRoot());
+        }
+
+        // model files
+        return $model->mediaRoot() . '/' . $hash;
+    }
+
+    /**
      * Tries to find a job file for the
      * given filename and then calls the thumb
      * component to create a thumbnail accordingly
@@ -97,22 +121,19 @@ class Media
     public static function thumb($model, string $hash, string $filename)
     {
         $kirby = App::instance();
-
-        // assets
-        if (is_string($model) === true) {
-            $root = $kirby->root('media') . '/assets/' . $model . '/' . $hash;
-        // parent files for file model that already included hash
-        } elseif (is_a($model, '\Kirby\Cms\File')) {
-            $root = dirname($model->mediaRoot());
-        // model files
-        } else {
-            $root = $model->mediaRoot() . '/' . $hash;
-        }
+        $root  = static::root($kirby, $model, $hash);
 
         try {
             $thumb   = $root . '/' . $filename;
             $job     = $root . '/.jobs/' . $filename . '.json';
-            $options = Data::read($job);
+
+            try {
+                $options = Data::read($job);
+            } catch (Throwable $e) {
+                $options = [
+                    'filename' => $filename
+                ];
+            }
 
             if (empty($options) === true) {
                 return false;
--- a/src/Cms/Media.php
+++ b/src/Cms/Media.php
@@ -130,9 +130,13 @@ class Media
             try {
                 $options = Data::read($job);
             } catch (Throwable $e) {
-                $options = [
-                    'filename' => $filename
-                ];
+                // no job file found, still run the job
+                // for publically accesible assets
+                if (is_string($model) === true) {
+                    $options = [
+                        'filename' => $filename
+                    ];
+                }
             }
 
             if (empty($options) === true) {
--- a/tests/Cms/Media/MediaTest.php
+++ b/tests/Cms/Media/MediaTest.php
@@ -139,7 +139,7 @@ class MediaTest extends TestCase
 
         // string model
         $hash = '12345678';
-        $this->assertStringEndsWith('tests/Cms/Media/tmp/media/assets/assets/' . $hash, Media::root($kirby, 'assets', $hash));
+        $this->assertStringEndsWith('tests/Cms/Media/tmp/media/assets/files' . $hash, Media::root($kirby, 'files', $hash));
 
         // file object
         $file  = $kirby->file('test.jpg');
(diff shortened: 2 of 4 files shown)


lukasbestle (commented):
Just a thought, I haven't looked into this in detail yet and haven't tested it:

Won't the fallback for missing `.jobs` files be a security issue as it would now be possible to access any file even if the site dev has not explicitly exposed it?

distantnative:
That could be true... 
Maybe we can differentiate further around `Media::thumb` whether it's used for an asset (where this wouldn't be an issue, they are accessible anyways) or a model/from content folder.

lukasbestle:
Since all asset URLs start with `media/assets`, we can probably differentiate this way.

src/Cms/Media.php
```diff
@@ -97,22 +121,23 @@ public static function publish(File $file, string $dest): bool
...
         try {
             $thumb   = $root . '/' . $filename;
             $job     = $root . '/.jobs/' . $filename . '.json';
-            $options = Data::read($job);
+
+            try {
+                $options = Data::read($job);
+            } catch (Throwable $e) {
+                // no job file found, still run the job
+                // for publically accesible assets
```
lukasbestle:
```suggestion
                // for publicly accessible assets
```

tests/Cms/Media/MediaTest.php
```diff
@@ -106,13 +126,144 @@ public function testPublish()
...
+        // copy test image to content
+        F::copy($this->fixtures . '/../files/test.jpg', $this->tmp . '/content/test.jpg');
+
+        // get file object
+        $file  = $this->app->file('test.jpg');
+        Dir::make(dirname($file->mediaRoot()));
+        $this->assertInstanceOf('\Kirby\Cms\File', $file);
+
+        // invalid with empty job file
+        F::write(dirname($file->mediaRoot()) . '/.jobs/' . $file->filename() . '.json', '{}');
```
lukasbestle, line 171: Why was the test with a missing job file removed?

tests/Cms/Media/MediaTest.php
```diff
@@ -106,13 +126,144 @@ public function testPublish()
...
+    {
+        // copy test image to content
+        Dir::make($this->tmp . '/content');
+        F::copy($this->fixtures . '/../files/test.jpg', $this->tmp . '/content/test.jpg');
+
+        $kirby = $this->app;
+
+        // string model
+        $hash = '12345678';
+        $this->assertStringEndsWith('tests/Cms/Media/tmp/media/assets/assets/' . $hash, Media::root($kirby, 'assets', $hash));
```
lukasbestle: To ensure that `assets` doesn't end up there by accident:

```suggestion
        $this->assertStringEndsWith('tests/Cms/Media/tmp/media/assets/files' . $hash, Media::root($kirby, 'files', $hash));
```

tests/Cms/Media/MediaTest.php
```diff
@@ -106,13 +126,144 @@ public function testPublish()
...
+        $this->assertFalse($thumb);
+
+        // create job file
+        $jobString = '{"width":64,"height":64,"quality":null,"crop":"center","filename":"test.jpg"}';
+        F::write(dirname($file->mediaRoot()) . '/.jobs/' . $file->filename() . '.json', $jobString);
+
+        // invalid with file not found
+        $thumb = Media::thumb($file, $file->mediaHash(), $file->filename());
+        $this->assertInstanceOf('Kirby\Cms\Response', $thumb);
+        $this->assertSame('', $thumb->body());
```
lukasbestle, line 182: I know this comes from the existing test, but I wonder: Why does this example lead to an empty response?

lukasbestle (commented):
General question: Why do we use `Media::thumb()` for assets in the first place? Shouldn't `Media::link()` be used? That doesn't support model-less files at the moment, but I feel like it would make more sense to extend that. Creating a thumb without options breaks with non-image files, doesn't it?

distantnative:
Would a media url for a non-image file make sense/be supported? If you e.g. have a PDF, one should rather use `->url()` directly.

lukasbestle:
Now that I think about it: Why would you use `mediaUrl()` on an asset at all? Since the file is already public, there is no point in copying it 1:1 to the media folder. So shouldn't we rather make `$asset->mediaUrl()` return the original URL?

distantnative:
Actually true... and then `FileVersion` also shouldn't have a `mediaUrl()` - we should discourage that usage.
It's either `$asset->url()` (URL to assets folder) or `$asset->thumb()->url()` (URL to modified thumb in media folder).

lukasbestle:
Yep. I think `mediaUrl()` is more of an internal method anyway. Or is there a use case where you would want to access it directly?

distantnative:
Somehow I took it as a given that there is, based on the issues mentioning it. But I cannot think of any actually

distantnative:
> So shouldn't we rather make $asset->mediaUrl() return the original URL?

That's not so easy sadly due to this line https://github.com/getkirby/kirby/blob/master/config/components.php#L103
Because of that, even generated thumbs would point to http://sandbox.test/assets/speaker-blur20.jpg

lukasbestle:
OK, makes sense.

I think we should check all issues again and ask the authors why `mediaUrl()` was used.

distantnative:
Collecting responses:

>  I wanted to be able to safely apply cache headers to our assets by taking advantage of the hash segment in the -> mediaUrl() for correct invalidation.

lukasbestle:
Ah, that's actually a useful case for `mediaUrl`. Then we should really support copying the file IMO, but that should also work for non-images like PDFs.

distantnative:
Will reopen for 3.6.x or 3.7

PR #3547, opened 2021-07-19
Fix `V::size` unit test for Field objects 

### LINKED ISSUE #3544: Running unit tests only for `Toolkit` namespace fails
While they pass when we run all unit tests

afbora:
## Related issues/ideas

- Fixes #3544

### DIFF
--- a/tests/Toolkit/VTest.php
+++ b/tests/Toolkit/VTest.php
@@ -2,7 +2,7 @@
 
 namespace Kirby\Toolkit;
 
-use Kirby\Cms\Page;
+use Kirby\Cms\Field;
 use PHPUnit\Framework\TestCase;
 
 class CanBeCounted implements \Countable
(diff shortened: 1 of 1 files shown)


distantnative:
```
composer run-script test tests/Toolkit
```

still fails for me

distantnative:
Is it working for you, @afbora? Only running tests for the `Toolkit` namespace?

afbora:
It gave an error when I ran only `VTest`. This was fixed with PR. But as you said, now when I run the entire Toolkit namespace, the same error continues 🙁

afbora:
By the way, I realized that it is not fully related to `Supports countable field objects`. Only when VTest runs alone I was giving the error but this PR fixes it. We can merge this PR without closing the related issue. I'll keep tracking down for running entire Toolkit tests fails.

distantnative (approved):

PR #3549, opened 2021-07-19
ItemImage's `sizes` via ResizeObserver

distantnative:
Similar attempt as https://github.com/getkirby/kirby/pull/2493 to fix https://github.com/getkirby/kirby/issues/2488 but this time with `ResizeObserver `

### DIFF
--- a/panel/src/components/Layout/Item.vue
+++ b/panel/src/components/Layout/Item.vue
@@ -15,7 +15,6 @@
     <k-item-image
       :image="image"
       :layout="layout"
-      :width="width"
     />
 
     <!-- Sort handle -->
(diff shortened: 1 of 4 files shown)


bastianallgeier:
I had a look at it, but I have to say that I don't understand what's happening. Could you explain it a bit?

distantnative:
Defining `sizes` for our item image is tricky because of the different layouts, different column sizes etc.
The best way for the browser to pic the right source from `srcset` would one to tell it the exact size the image is displayed at in the `sizes` attribute.

Which is what we do here:
We create at `window.panel.$imgsizes` a `ResizeObserver` if it doesn't already exists and each `ItemImage` adds its `` to it via `.observe()`. Whenever the ResizeObeserver runs (also initially) it loops through all elements it is observing and adds the content rect width as `sizes` value.

And on component destruction, we make sure to unobserved its `` tag

distantnative:
let's bring this back with 3.6.x

PR #3556, opened 2021-07-21
Fixes blocks field sort handle on hover #3538

### LINKED ISSUE #3538: Blocks: drag handle on hover shows pointer cursor instead of dragging hand

afbora:
## Describe the PR

Fixed blocks field sort handle on hover with using common `.k-sort-handle` class.

⚠️ Is the change of the handle class a breaking change?

### Fixes

- Fixed blocks field sort handle shows pointer cursor on hover `#3538`

## Breaking changes

None

## Related issues/ideas

- Fixes #3538 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Blocks/BlockOptions.vue
+++ b/panel/src/components/Blocks/BlockOptions.vue
@@ -37,7 +37,7 @@
       />
       <k-button
         :tooltip="$t('sort')"
-        class="k-block-options-button k-block-handle"
+        class="k-block-options-button k-sort-handle"
         icon="sort"
         @keydown.up.prevent="$emit('sortUp')"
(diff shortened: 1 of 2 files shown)


distantnative:
It is a small breaking change (in case a developer uses the class to restyle the handle). But I think it's one we could do - we just should reference it in the release notes.

If we want to fully avoid it, I would think that keeping `k-block-handle` but adding also `k-sort-handle` class could fix the bug as well.

afbora:
I thought of copying the class, but I preferred the `.k-sort-handle` class because no code duplication and also use the `.k-sort-handle` class in the layout field handle. Which one do you think would be better to go with?

bastianallgeier (approved):

PR #3557, opened 2021-07-21
Fixes drag handle for picker fields #3553

### LINKED ISSUE #3553: [3.6.0-alpha.2] Pages, files and users field not sortable, no drag handle
**Describe the bug**  
While pages **sections** have a drag handle, I can not sort the selected pages in a pages **field**.

**To Reproduce**  
1. Add a files field
2. Select files and try to sort them

**Expected behavior**  
A drag handle so I can sort the pages.

**Kirby Version**  
3.6.0-alpha.2

 - OS: macOS Big Sur 11.4
 - Browser: Chrome 
 - Version: 91.0.4472.114

afbora:
## Describe the PR

`item.sortable` is related for sections (model permissions). I've fixed the issue with checking `sortable` definition. Page/User/File fields items no need sort permission.

### Regression Fixes

- Fixed drag handle for picker fields `#3553`

## Breaking changes

None

## Related issues/ideas

- Fixes #3553 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Layout/Items.vue
+++ b/panel/src/components/Layout/Items.vue
@@ -15,7 +15,7 @@
         v-for="(item, itemIndex) in items"
         :key="item.id || itemIndex"
         v-bind="item"
-        :class="{'k-draggable-item': item.sortable}"
+        :class="{'k-draggable-item': sortable && item.sortable}"
         :image="imageOptions(item)"
         :layout="layout"
         :sortable="sortable && item.sortable"
--- a/src/Panel/Model.php
+++ b/src/Panel/Model.php
@@ -295,14 +295,15 @@ abstract class Model
         //       no longer needed in 3.6
 
         return [
-            'id'    => $this->model->id(),
-            'image' => $this->image(
+            'id'       => $this->model->id(),
+            'image'    => $this->image(
                 $params['image'] ?? [],
                 $params['layout'] ?? 'list'
             ),
-            'info'  => $this->model->toString($params['info'] ?? false),
-            'link'  => $this->url(true),
-            'text'  => $this->model->toString($params['text'] ?? false),
+            'info'     => $this->model->toString($params['info'] ?? false),
+            'link'     => $this->url(true),
+            'sortable' => true,
+            'text'     => $this->model->toString($params['text'] ?? false)
         ];
     }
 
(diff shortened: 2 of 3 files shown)


panel/src/components/Layout/Items.vue
```diff
@@ -15,10 +15,10 @@
         v-for="(item, itemIndex) in items"
         :key="item.id || itemIndex"
         v-bind="item"
-        :class="{'k-draggable-item': item.sortable}"
+        :class="{'k-draggable-item': sortable && (item.sortable !== undefined ? item.sortable : true)}"
```
distantnative:
```suggestion
        :class="{'k-draggable-item': sortable && (item.sortable || true)}"
```

Would this work as well?

afbora: I thought so too at first but `false || true` returns `true`. We need `item.sortable ?? true` but our panel still doesn't support:
 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

distantnative: Ahh of course, you are right!

afbora: Why `Nullish coalescing operator` doesn't working. I can test on JSFiddle but in our panel throwing error on `npm run build`. Do you have any idea?

bastianallgeier: We should fix this in way that item.sortable is already a boolean and the check is not needed.

afbora: @bastianallgeier `item.sortable` is `boolean` on sections but `undefined` on pages/files/users field.

bastianallgeier: Then let's fix it on the PHP side

afbora: With setting `item.sortable = true` as default on PHP for picker fields?

afbora: @bastianallgeier I've push commit https://github.com/getkirby/kirby/pull/3557/commits/6ec420d3a2cb58a00bcaee10f6130591c57617b6. Like that?

bastianallgeier:
@afbora yep, that's great!

bastianallgeier (approved):

PR #3483, opened 2021-07-04
Fixes layouts field parsing empty value

afbora:
## Describe the PR

We parse it in the Blocks field and it returns empty array even if there is an error, but we did not do it for the Layouts field. I've fixed the issue with parsing layout value that returns empty array if field value empty or an error occured.

### Fixes

- Fixed layout field empty when use `->toLayouts()` method

## Breaking changes

None

## Related issues/ideas

- Fixes #3482 

## Ready?

- [x] Unit tests for fixed bug/feature
- [x] In-code documentation (wherever needed)
- [x] CI checks pass

## When merging

- [x] Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)
- [ ] Add changes to release notes draft in Notion

PR #3565, opened 2021-07-22
`Data\Json` and `Data\Xml`: Decode empty strings

lukasbestle:
## Describe the PR

The `Txt` and `Yaml` classes already accepted empty strings, so I think it makes sense to have the same behavior for the other formats.

### Enhancement

- The `Data\Json::decode()` and `Data\Xml::decode()` methods now accept empty strings and treat them as an empty data set (empty array) for consistency with `Data\Txt::decode()` and `Data\Yaml::decode()`

## Breaking changes

The `Data\Json::decode()` and `Data\Xml::decode()` methods no longer throw exceptions if an empty string is passed. If a differentiation between an empty string and an empty data set is required, you need to check for empty strings before calling `decode()`.

## Related issues/ideas

- See https://github.com/getkirby/kirby/pull/3483#issuecomment-873722279

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Data/Json.php
+++ b/src/Data/Json.php
@@ -34,7 +34,7 @@ class Json extends Handler
      */
     public static function decode($string): array
     {
-        if ($string === null) {
+        if ($string === null || $string === '') {
             return [];
         }
 
--- a/src/Data/Txt.php
+++ b/src/Data/Txt.php
@@ -93,7 +93,7 @@ class Txt extends Handler
      */
     public static function decode($string): array
     {
-        if ($string === null) {
+        if ($string === null || $string === '') {
             return [];
         }
 
--- a/src/Data/Xml.php
+++ b/src/Data/Xml.php
@@ -35,7 +35,7 @@ class Xml extends Handler
      */
     public static function decode($string): array
     {
-        if ($string === null) {
+        if ($string === null || $string === '') {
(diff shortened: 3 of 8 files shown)


distantnative:
But `Yaml::decode()` and `Txt::decode()` right now wouldn't follow the same behaviour?

afbora:
Empty array as `[]` returns when empty string as `''` send to `Yaml::decode()` and `Txt::decode()` methods, but `Json::decode()` and `Xml::decode()` were throwing an error. This PR is resolving it for consistency.

distantnative:
But shouldn't we add the same if statements for consistency? Would be easier to understand if they all check for it and sparing to run the parsers on an empty string might also be good.

afbora:
I understand 👍  Yes, if we add the same if condition, there will be consistency as a general view. (Although it now returns the correct output). Of course, as you said, the parser does not need to run when empty string sends. So yes, it would be nice to easily see by looking at the code that they all exhibit the same behavior across all handlers.

lukasbestle:
Very good points! I will add the remaining lines.

afbora (approved):
I've added remaining lines.

bastianallgeier (approved):

PR #3568, opened 2021-07-23
`$direction` via Fiber

distantnative:


panel/src/helpers/direction.js
```diff
@@ -1,29 +0,0 @@
...
-   * - Current editing language exists
-   * - Input is not disabled
-   * - Editing language direction not equal with default language direction or
-   *   user language not equal with editing language
-   *
-   */
-  if (
-      multilang &&
-      language &&
-      app.disabled === false &&
```
afbora, line 21: This line means: When field disabled like `translation: false`, direction should keep as default language. 

How can we handle right now?

src/Panel/View.php
```diff
@@ -148,10 +148,28 @@ public static function data(array $view = [], array $options = []): array
...
-                if ($multilang === true && $language = $kirby->language()) {
+            '$direction' => function () use ($kirby, $multilang, $language, $user) {
+                // only if mulitlang and current content language exists
+                if ($multilang === true && $language) {
+                    // content language direction is not
+                    // the same as default language direction
+                    // and user language is not the same as
+                    // content language
+                    if (
+                        $language->direction() !== $kirby->defaultLanguage()->code()||
```
afbora:
```suggestion
                        $language->direction() !== $kirby->defaultLanguage()->direction() ||
```

panel/src/helpers/direction.js
distantnative: So it relied on that the component would have a `disabled` prop/computed prop?
We might have to do that in the components then:

```
:dir="direction"
```

to 

```
:dir="!disabled ? direction : null"
```

distantnative: We might even use a Vue directive to not repeat ourselves? https://v3.vuejs.org/guide/migration/custom-directives.html#_3-x-syntax

afbora: Let me give you an example:

Default language: EN
Secondary language: AR (RTL)
Text field: `translate: false`

When you switch to secondary language, you'll see still EN content. So direction should be enabled only field editable.

afbora:
Works perfect!. There is only the issue of coverage diff and conflicts, that's all.

distantnative:
Passes unit tests/coverage now!

bastianallgeier (approved):

PR #3577, opened 2021-07-24
`k-file-preview`: clean CSS, details from backend

distantnative:
```
BEFORE
dist/css/style.css   107.11kb / brotli: 14.94kb
dist/js/index.js     316.29kb / brotli: 60.18kb

AFTER
dist/css/style.css   106.27kb / brotli: 14.87kb
dist/js/index.js     315.29kb / brotli: 60.03kb
```

### DIFF
--- a/panel/src/components/Layout/FilePreview.vue
+++ b/panel/src/components/Layout/FilePreview.vue
@@ -113,6 +113,7 @@ export default {
 @media screen and (min-width: 65em) {
   .k-file-preview-layout {
     grid-template-columns: 33.333% auto;
+    align-items: center;
   }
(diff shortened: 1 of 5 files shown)


bastianallgeier (dismissed):

distantnative:
@bastianallgeier 
We're using now `dimension()->toString()`, so the result is `933 x 1400 Pixels` instead of before `933x1400 Pixels`. Is that fine? Otherwise, we should change it in the PHP class.

bastianallgeier:
@distantnative I think it looks even better.

bastianallgeier (approved):

PR #3584, opened 2021-07-24
Fix tabbed link styling

### LINKED ISSUE #3576: [3.6.0] Topbar breadcrumb: no tabbed styling

distantnative:
## Describe the PR

We made a small mistake when moving from SCSS to CSS. See https://github.com/getkirby/kirby/blob/38afb2d650f3764d23d732b23ae8f04ab9ffba69/panel/src/main.scss#L168-L174

### Fixed regressions from alpha.2
- Styling for tabbed link highlighting works again

## Related issues/ideas

- Fixes #3576

### DIFF
--- a/panel/src/components/Navigation/Link.vue
+++ b/panel/src/components/Navigation/Link.vue
@@ -118,7 +118,6 @@ export default {
   outline: none;
 }
 .k-link[data-tabbed] {
-  outline: none;
   box-shadow: var(--shadow-outline);
 }
 </style>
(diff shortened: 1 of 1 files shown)


panel/src/components/Navigation/Link.vue
```diff
@@ -117,7 +117,7 @@ export default {
 .k-link {
   outline: none;
 }
-.k-link [data-tabbed] {
+.k-link[data-tabbed] {
   outline: none;
```
lukasbestle: This line can also be removed as it's duplicated from above:

```suggestion
```

lukasbestle (approved):

PR #3586, opened 2021-07-24
Injects pages from pages extension into site inventory

texnixe:
## Describe the PR
I'm not sure if this is the right way to go about this. But one thing I have been missing is that pages from the `pages` extension don't become part of the site object and thus are not first class children of `$site`. This PR aims to fix this, but feel free to throw it away if this is the wrong approach.

https://forum.getkirby.com/t/how-to-add-virtual-pages-to-kirbys-index/19569/4

## Release notes
- Injects pages from pages extension into `$site->inventory()`

## Breaking changes
none

## Ready?

- [ ] CI checks pass

## When merging


lukasbestle (commented):
I agree that it's a bug that the pages from the `pages` extension are not considered "real" children.

My first thoughts regarding the implementation:

- I feel like `inventory()` is the wrong place for this as it's loading the actual files and pages from the content dir. But on the other hand this change would really make the virtual pages first-class citizens. Not sure about the implications either way, I don't have a good overview about this part of the system at the moment.
- What if a virtual page is not at the root level? Shouldn't it be injected into the inventory/children list of its parent instead of into the site?

texnixe:
> What if a virtual page is not at the root level? Shouldn't it be injected into the inventory/children list of its parent instead of into the site?

Good point, I was so focused on the first level pages, I didn't even think about the fact that you could create children of other pages via the pages extension.

texnixe:
Is it even possible to add non-first-level pages via the pages extension? That doesn't seem to work, unless I'm doing something wrong. Or rather, the route works, but they are not added to the inventory at all nor are any of the others in that case.

afbora:
As far as I understand, virtual pages are only accessible, not indexed in inventory.

Because when `page('test-page')` or `site()->find("test-page")` called, running [Pages::get()](https://github.com/getkirby/kirby/blob/master/src/Cms/Pages.php#L309-L320) that find via `->extension('pages')`. But when `site()->index()` called, no page searches/finds are performed. All pages are loaded in the inventory.

I'm not sure too but it seems like the right place (at least with current codes) where @texnixe injected. May be refactoring needs 🤔

lukasbestle:
Probably we can remove the custom behavior in `Pages::get()` and instead add support to `Site::inventory()` and `Page::inventory()`. Each `inventory()` method would check for each page extension whether its parent is the current object, then I think it will work globally.

texnixe:
The weird thing is that `$kirby->extensions('pages')` returns an empty array inside the `$site->inventory()` method as soon as I add a page with a parent in the pages extension.

```
<?php

Kirby::plugin('cookbook/pages', [
  'pages' => [
    'sub-page' => new Page([
      'slug' => 'sub-page',
      'template' => 'note',
      'parent' => page('notes'),
      'content' => [
        'title' => 'New Subpage',
      ],
    ]),
    'new-page' => new Page([
      'slug' => 'new-page',
      'template' => 'default',
      'content' => [
        'title' => 'New Page',
      ],
    ]),
  ],
]);
```

In the page inventory, this still works fine, but not in the site inventory.
Unfortunately, I haven't succeeded debugging this issue 😢. Wonder if this is some sort of deferred loading.

lukasbestle:
Yeah, plugins must not call the `page()` helper. That initializes the `$site` object, so the inventory will already be built and cached.

Sorry, my suggestion was a bit unclear. I didn‘t mean the actual parent, but the one generated from the page ID. Not sure if that even works right now though.

texnixe:
That's what I thought, but the parent property seems to expect a page object…

That's why I thought that the pages extension only worked for first level pages.

lukasbestle:
You are right. I thought we had a `setId` setter that would allow setting a full id like `projects/project-a`, but we don't. The ID is always generated dynamically using the `parent`. So it really looks like we only have support for root-level pages at the moment.

In this case it's probably fine to do it in `$site->inventory()`, but we should get rid of the `Pages::get()` method then and verify that everything still works with the new setup.

PR #3593, opened 2021-07-27
Set files `sort` attribute on upload 

### LINKED ISSUE #2886: Sortable Files Section Should Set "Sort" Automatically for New Files
**Describe the bug**  
When I files section is sortable, I find it odd that a "sort" num isn't automatically set for files on upload. 

**To Reproduce**  
1. Have a sortable files section in a Blueprint
2. Upload a file called `foo.jpg`
3. Upload a file called `bar.jpg`

**Expected behavior**  
The upload files have no `sort` value, so the Files section reverts to the sorting them by filename. Thus, the order of files in the section shows as:
1. `bar.jpg`
2. `foo.jpg`

I would have expected the first file (`foo.jpg`) to be first.

**Kirby Version**  
`3.4.4`

**Additional context**  
If you already have pre-sorted files, newly uploaded files appear at the beginning of the files section (effectively like their "sort" num is `0`). If you uploaded multiple new files, they'll be sorted by filename (as described above), but still before previously sorted files.

I can certainly see some people prefer "new" files to be added at the beginning of a files section, but that should be accomplished by a setting that forces new files to get a `1` sort value (and re-numbering the other files). I would expect the default behavior to to be: new files receive a sort num of 1 _higher_ than the previous highest sorted file.

afbora:
## Describe the PR

If files section sortable, sets `sort` attribute as last item.
Otherwise no sort attribute set or write to file content.
Feel free to close the PR if not right implementation.

### Fixes

- Sets automatically `sort` attribute on upload if section sortable `#2886`

## Breaking changes

None

## Related issues/ideas

- Fixes #2886

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/api/routes/files.php
+++ b/config/api/routes/files.php
@@ -34,13 +34,19 @@ return [
         'pattern' => '(:all)/files',
         'method'  => 'POST',
         'action'  => function (string $path) {
+            // move_uploaded_file() not working with unit test
+            // @codeCoverageIgnoreStart
             return $this->upload(function ($source, $filename) use ($path) {
                 return $this->parent($path)->createFile([
+                    'content' => [
+                        'sort' => $this->requestBody('sort')
+                    ],
                     'source'   => $source,
                     'template' => $this->requestBody('template'),
                     'filename' => $filename
                 ]);
             });
+            // @codeCoverageIgnoreEnd
         }
     ],
     [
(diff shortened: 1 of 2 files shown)


afbora:
I need to write real upload test for missing/diff unit test. Can we ignore it?

bastianallgeier (approved):

PR #3596, opened 2021-07-30
Privacy friendly Video.vue

doriansimeha:
A proposition to be more GDPR compatible

### DIFF
--- a/panel/src/components/Blocks/Types/Video.vue
+++ b/panel/src/components/Blocks/Types/Video.vue
@@ -36,14 +36,14 @@ export default {
       var youtubeMatch = url.match(youtubePattern);
 
       if (youtubeMatch) {
-        return "https://www.youtube.com/embed/" + youtubeMatch[2];
+        return "https://www.youtube-nocookie.com/embed/" + youtubeMatch[2];
       }
 
       var vimeoPattern = /vimeo\.com\/([0-9]+)/;
       var vimeoMatch = url.match(vimeoPattern);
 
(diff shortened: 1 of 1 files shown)


texnixe:
Good idea. Note that you would have to change related tests, otherwise they would fail.

lukasbestle:
To be honest I wonder why the URLs are constructed manually in the `Video` block component since we have our very flexible embed helpers.

Idea: We could extend the [embed helpers](https://github.com/getkirby/kirby/blob/develop/panel/src/helpers/embed.js) with a new flag like `dnt` that will always enforce an embed URL without tracking when set to `true` (default would be auto-detection based on the provided URL like in the current implementation). Then the embed helper can be used in the block component without additional logic.

texnixe:
Why not make `dnt` the default?

lukasbestle:
The idea with the embed helpers is to auto-generate a matching embed URL for the provided video URL. We already support DNT there by using the `youtube-nocookie` domain in the video URL. So if we do it by default, users can no longer choose which behavior they want.

The disadvantage of the DNT versions is that you cannot add videos to a playlist (like the watch later playlist). This can be useful sometimes.

So while I fully agree that the blocks preview should use DNT embeds all the time, the actual embeds should be customizable IMO.

bastianallgeier:
Thanks for your PR! I've implemented it in the way @lukasbestle suggested: https://github.com/getkirby/kirby/pull/3673

PR #3601, opened 2021-08-03
Load Vue components automatically

distantnative:
💥 Really brings up JS file size - because of the additional lines for the loader but probably even more due to less tree-shaking without manual import statements?

## Performance

```
BEFORE
dist/css/style.css   101.03kb / brotli: 14.59kb
dist/js/index.js     295.58kb / brotli: 57.01kb
dist/js/vendor.js    383.11kb / brotli: 105.08kb

AFTER
dist/css/style.css   101.03kb / brotli: 14.55kb
dist/js/index.js     309.76kb / brotli: 58.39kb
dist/js/vendor.js    383.09kb / brotli: 104.99kb
```


distantnative:
@bastianallgeier given the increase in JS file size, I am leaning towards abandoning this project and stick to manual import lines - no matter how tedious they are. What do you think?

lukasbestle:
How often do we need to change the imports? IMO only when we add new components or remove components. In both cases we can't forget about it (in the first case the feature won't work at all, in the second case the build tool will throw an error). So without even considering bundle file size, is this increase in complexity worth it? For me it's an open question (not rhetorical), just something to think about.

distantnative:
The main driver behind this probably by now year-old idea was to treat core and plugin components the same in terms of registering, resolving mixing etc. 

E.g. the idea was to use mixing in the same way plugins could via string reference. But sadly, this also conflicts with parsing our core components for the Reference.

I wanted to make the experiment actually work, which it does now. But in the end I think we should stick with the current setup, no matter what monster that long list is. Actually probably even more, we should turn the existing dynamic blocks import into a manually defined list.

lukasbestle:
I think you two should decide. Both variants are fine for me, I'm leaning towards the explicit list.

bastianallgeier:
Looking at the massive changes that we already introduce in 3.6, we should abandon this for now. I think it's time to focus on getting it done without too many additional experiments.

distantnative:
@bastianallgeier i didn't consider this for 3.6 but later. Though, I'd prefer if we make an actual decision to throw this idea out (or continue it later) instead of just postponing it.

bastianallgeier:
I wonder if this will get in the way when we *finally* get into code splitting. Otherwise I would keep it.

distantnative:
It probably will. Plus the overhead because tree shaking seems not to work as well with these... let's just throw it out the window.

Issue #1372, opened 2019-01-18
Reintroduce 'options: hide' for page blueprints


SirNovi:
I assumed the 'read' option was the new 'hide' equivalent but this doesn't seem to be the case, therefore I would really appreciate it if we could get the old 'hide' option back.

bastianallgeier:
options: read should in fact solve this. What's not working for you?

SirNovi:
Hypothetical setup: 

```
sections:
  pages:
    type: pages
    parent: site.find("hiddenpage")
``` 
The bluprint for hiddenpage uses `read: false`

With this setup the "hiddenpage" does not appear in the panel and is indeed hidden. The pages section does still work as it will show all children of "hiddenpage". But adding new pages via this section will throw an error. (The page "hiddenpage" cannot be found)

Now, I must admit that I'm not sure the hide option in v2 actually behaved differently. But I do think that there should be an option that makes pages _only_ hidden for the panel user and not for the panel itself as seems to be the case with the `read` option.

medienbaecker:
@bastianallgeier This is also a deal breaker for a current project of mine. Maybe my use case helps you understand what this issue is about:

Since the builder field still has some issues, you have to use subpages for a "one-pager" setup.
Most of the time you need separate sections (or modules) **and** subpages, though. Therefore the sections should live in a hidden `sections` folder to have separate sorting numbers. Similar to the `_modules` folder with the modules plugin in Kirby 2.

It's not possible to hide the `sections` folder without setting `read` to `false`. But that means I can't add pages to the pages section:

![sections](https://user-images.githubusercontent.com/7975568/53085844-12057680-3504-11e9-96dc-ebbc6785e4ca.gif)

I'd suggest adding a `hide` option that simply hides the page from pages sections and the breadcrumb navigation.

medienbaecker:
@bastianallgeier Is there anything I can do to help you remove the "missing information" label?

silllli:
Also when setting `read` to `false` the page can't be fetched via the API anymore (accessing it with a template still works).

distantnative:
https://kirby.nolt.io/124

PR #3621, opened 2021-08-08
New `access`  and `show` page permissions and deprecate `read` option

afbora:
## Describe the PR

Brings the `hide` option in Kirby 2 to Kirby 3 as `show` so that the pages are accessible, ie viewable and updatable, but not visible.

- `read` is deprecated
- replaced/separated by `access` and `show` permissions
- `access` option sets accessibility and visibility
- `show` option sets only visibility

### Sample usage
````yaml
# Orders page is no accessible and no visible for all roles except admins.

title: Orders

options:
  access:
    *: false
    admin: true
  show: 
    *: false
    admin: true
````

### Features
- Reintroduce `options: show` for page blueprints
### Fixes

 - Fixed searching unreadable pages `#3650`

## Breaking changes

None

## Related issues/ideas

- Implements https://kirby.nolt.io/124

## Ready?

- [x] CI checks pass

## When merging


distantnative:
Isn't this very confusing to have a `read` and `show` option now in blueprints?

afbora:
We can give it a different name, but the function of the both is different. We have 3 options:

1. Using show/read props as in PR
2. Changing show/read prop names as more clear (It's pretty clear to me now)
3. Update the read option as writable but keep invisible (currently no writable and no visible and no need additional prop like show)

afbora:
Perhaps a fourth option would be more explanatory.

4. New `write` option. `read` is readable, `write` is writable. (contains breaking change)

medienbaecker:
> Isn't this very confusing to have a `read` and `show` option now in blueprints?

There is a huge difference actually. Especially for container pages. You can't even access/move/edit children of pages that have `read: false`. In some other circumstances it makes sense to actually disable a page so both options have their raison d'être.

I think option 1 is the way to go. The options are self-explanatory in my opinion.

distantnative:
Yes, I wasn't questioning the functionality - I think both have their use cases.

But I would disagree that the difference between `show` and `read` is obvious, rather the contrary: if you don't know yet what the difference could be, hard to understand how they differ.

afbora:
> I would disagree that the difference between show and read is obvious

What about `read` and `write`. Because currently `read` represents two permissions that readable and writable. It would be better if we separate it. What do you think?

lukasbestle:
As far as I can tell, the issue in https://github.com/getkirby/kirby/issues/1372 basically just comes from our implementation and wasn't intended: To create a new page, the backend needs to be able to instantiate the `Page` object of the parent. But if the `read` option of that parent is set to `false`, it looks like the page doesn't exist. So if instantiating the parent `Page` wasn't an implementation necessity to create new pages, we wouldn't even have this issue.

I haven't looked at the code in detail, but considering that fact I think there can't be such thing as a `write` option. Only a `access this page to create children` option (which is exactly what `read` does internally for the mentioned reason).

I really like the `show` name. It makes clear that it's just about whether this particular page is displayed in lists.

I'm not so sure about `read`. With its implications on the rest of the system (especially the "create child page in a section" example) it easily causes edge cases. What are examples where you actually want that behavior that the whole page is inaccessible?

afbora:
> What are examples where you actually want that behavior that the whole page is inaccessible?

I haven't needed the inaccessibility feature yet, but it might be needed if the Kirby content structure is used like a database.

Actually I agree that it's confusing because `read` also includes the `show` feature. 
`read: false` means no visible and no accessible. `show: false` means no visible only.

Thinking out loud; This confusion can go away if we split the read option into two. 

````
read: show & access *
show: visibility
access: accessibility
````

Of course, the developer can also enable show and disable access. In this case, the user cannot access the item seen.

_* Maybe we can deprecate the read option in the next major version._

lukasbestle:
I really like the idea to create a new `access` permission and deprecate `read`. That's much more clear in the long run.

Two implications we need to think about:

- We also have permissions `pages.read` and `files.read`. For consistency these would have to be renamed as well.
- What would actually happen when `access` is `false` but `show` is `true`? In theory it would lead to a read-only access where the page is listed in sections but can't be opened in the Panel. We need to check if that behavior is possible or if `access = false` would make `show = true` impossible because of our permission checks.

afbora:
I think we're going to need something like this:

````
access = permission.read === false ? permission.read : permission.access
show   = permission.read === false ? permission.read : (permission.access === true && permission.show === true)
````

lukasbestle:
That would be another alternative, yes. So it would not be possible to define `show = true` when `access = false`. `show` implies `access`.

If I understand it correctly, we would "just" need to rename `read` to `access`, correct?

afbora:
Yes, if we're going to remove the `read` option in 3.6. There will be only `access` and `show`. It would probably be something like this:

````php
public function isAccessible(): bool
{
    ...
    return $this->permissions()->can('access');
}

// `show` implies `access`
public function isVisible(): bool
{
    ...
    return $this->isAccessible() === true && $this->permissions()->can('show');
}

lukasbestle:
We definitely need to deprecate `read` first, but we can remove it in 3.7.

afbora:
If everyone agrees on the deprecation of `read`, the new `access` and `show` options, I can apply it for first review.

bastianallgeier:
Access sounds like a good alternative for read. I just wonder if we mix things here. 

`read` or `access` is a real global permission option in my opinion.`show` sounds more like a general visibility flag that does not really come with real permission implications. I might be wrong, but in this case it sounds more like you want to hide a certain page from the panel without any other implications.

afbora:
The purpose is to hide certain pages as default in the entire panel without the need for any additional queries. Hidden pages are accessible, updatable, but hidden.

````yaml
access: false # no accessible, no visible

show: false   # accessible, no visible
````

bastianallgeier:
@afbora yes, I understood that. I was just wondering if it makes sense to mix that with the other permission options, or if this should be a general flag like icon, title, etc.

afbora:
I better understand what you mean. It started to make more sense. But the `show` option depends on the `read/access` option. In other words, in order for the page to be hidden, it depends on the enabled `access` permission. Is this a problem or confusion? I mean the mixing of flags and permissions.

lukasbestle:
This discussion is already about the blueprint options for page blueprints. But if we change it there, IMO we should also change the name of the permissions that do the same thing from the role perspective. I agree we don't need a "show" permission, but we do have a "read" permission that would also need to be renamed to "access".

afbora:
There may be a point I missed. We've always thought of the show option as a flag, it's true. But it can also be used as a permission. E.g; I want the orders page to be accessible but also viewable only by admin users. What do you think about the case?

````yaml
# /site/blueprints/orders.yml
options: 
  show: 
    admin: true
    *: false
````

afbora:
Any feedback? 👀

bastianallgeier:
TBH I find the decision really hard. It feels like we have to mess around with permissions quite a bit for it. My mental capacity to get this done or at least make preparations for 3.6 is kind of overloaded.

afbora:
### Summarize

- `read` is deprecated
- replaced/separated by `access` and `show` permissions
- `access` option sets accessibility and visibility
- `show` option sets only visibility

### Sample usage
````yaml
# Orders page is no accessible and no visible for all roles except admins.

title: Orders

options:
  access:
    *: false
    admin: true
  show: 
    *: false
    admin: true
````

\* Coverage will fix after reviews done

PR #3623, opened 2021-08-10
Form performance

distantnative:


config/fields/structure.php
```diff
@@ -181,7 +186,10 @@
         $data = [];
 
         foreach ($value as $row) {
-            $data[] = $this->form($row)->content();
+            $form = $this->form;
```
afbora, line 189: I guess this should be defined without going into the loop.

config/fields/structure.php
```diff
@@ -153,26 +159,25 @@
                     continue;
                 }
 
-                $value[] = $this->form($row)->values();
+                $form = $this->form;
```
afbora, line 162: I guess this should be defined without going into the loop.

config/fields/structure.php
```diff
@@ -91,18 +91,24 @@
...
-        'default' => function () {
-            return $this->rows($this->default);
-        },
-        'value' => function () {
-            return $this->rows($this->value);
+        'form' => function () {
+            return new Form([
+                'fields' => $this->fields,
+                'model'  => $this->model
+            ]);
```
afbora, line 98: I have an idea about failing tests. It can give you an idea. Because it is fixed when I move it from `computed` to `methods`.

I guess calling `$form->fill()` not affects to computed `$this->form` variable.

bastianallgeier: Ah, that makes a lot of sense!

distantnative:
@afbora @bastianallgeier what was the status of this? I recall we were running into an issue, but not exactly what it was anymore.

bastianallgeier:
TBH, I don't fully understand what caused the problem. Feels like we need to dive much deeper into the weird internals of forms and fields to get this right and I'm not sure if we find the time for that for the beta.

bastianallgeier:
Should we close this and come back to it in 3.7?

distantnative:
Yes, let's just make sure to store the code locally

bastianallgeier:
I have a local branch. If you keep one as well, it'd be great.

distantnative:
Done

distantnative:
@lukasbestle @bastianallgeier I don't think this one would suit a before-summer release. Better to move to 3.8?

lukasbestle:
I don't know the background of this PR, but it looks like a pretty large refactoring. So in this case, it may indeed not be suited for 3.7.

I think the question should be: What is the plan with this in general? If it's still relevant, I think we should move it to 3.8 but make it a roadblock and not move it further. Otherwise we should close it.

distantnative:
I think this should be considered in the larger form refactoring, whenever that takes place

distantnative:
Solving this differently with blueprint refactoring.

Issue #2537, opened 2020-03-24
Thumbnails don’t include ICC color profile


fabianmichael:
**Describe the bug**  
When generating thumbnails from images, the color profile is stripped from the file. This causes problems on wide gamut-monitors such as highend graphic design monitors, but also more consumer-grade devices like Apple products with P3 color gamut (e.g. some iPad models, 5k iMac, MacBook Pro …), as colors of thumbnails appear over-saturated in Mozilla Firefox.

For the ImageMagick driver, there's a hardcoded method, that always adds the `-strip` parameter to the imagemagick cli command.
https://github.com/getkirby/kirby/blob/master/src/Image/Darkroom/ImageMagick.php#L136

This does not only remove unwanted metadata, but also any color profile attached to the image. Some browsers like Safari do color correction on wide gamut displays by assuming, that image files without profiles are in sRGB color space, but Firefox just displays the image in the full gamut. I can understand the decision why this is present, because it makes thumbnail files a bit smaller, but it produces weird results, especially when you use the original image file in a layout, because its thumbnail will show a noticable color shift.

The simplest solution would be to not strip any metadata or at least providing an option to configure this. A more advanced solution would be, to ensure that every thumbnail includes a color profile. According to [this post](https://phabricator.wikimedia.org/T100976), the default sRGB profile adds about 2.5 kB to the thumbnail file.

The more advanced solution, as used by large image hosters (e.g. Facebook) would be to accept a small amount of inaccuracy by replacing the original profile data with a minimal sRGB profile, that only weights around 0.5 kB (e.g. https://pippin.gimp.org/sRGBz/). However, I don’t know if the latter approach would be worth it to be included into core.

**To Reproduce**  
1. Upload an image file (I tested with a JPEG) with an sRGB profile included to any page. I used the ImageMagick driver for this test.
2. In the corresponding template file, fetch the image via the API and embed the original file and a thumbnail of that file next to each other.
3. Compare the 2 images, the thumbnail should have over-saturated colors. This effect becomes way more noticable on wide gamut displays, such as the Apple iMac 5k.

**Expected behavior**  
The thumbnail’s colors should match the color values of the original file in all browsers, that support color management in some way.

**Screenshots**  
![spaghetti](https://user-images.githubusercontent.com/395617/77417144-457c9980-6dc5-11ea-94fc-b6b09d1bdeb1.png)
[...]

PR #3648, opened 2021-08-25
Keep icc color profiles and add test. #2537

fabianmichael:


fabianmichael:
Sorry, something was wrong on my side.

PR #3649, opened 2021-08-25
Keep icc color profiles

### LINKED ISSUE #2537: Thumbnails don’t include ICC color profile

fabianmichael:
Fixes #2537.

### Enhancements

- Thumbs created with the `ImageMagick` driver now include the ICC color profile for better color accuracy. Note that this currently does *not* work for PNG images.


lukasbestle:
Thank you for the PR!

I've looked at the ImageMagick docs again and found the following list of [what `-strip` removes](https://imagemagick.org/script/command-line-options.php#strip):

```
Strip the image of any profiles, comments or these PNG chunks:
bKGD,cHRM,EXIF,gAMA,iCCP,iTXt,sRGB,tEXt,zCCP,zTXt,date.
```

So the question is: Are comments and the mentioned PNG chunks also a profile in the ImageMagick sense? In https://github.com/getkirby/kirby/issues/2537#issuecomment-882381102 you wrote that any kind of meta information is a profile. I think we need a unit test for at least comments and `EXIF`, `iTXt`, `tEXt`, `zTXt` and `date` data in PNGs. If a user relies on our thumb engine that it strips this kind of information and for some reason it isn't stripped anymore, this could be a security issue, for example with sensitive EXIF location information.

fabianmichael:
@lukasbestle I added a test PNG with profile and added GPS data using Exiftool to both the PNG and the JPEG image. Imagemagick removes GPS data reliably from the JPEG, but not the PNG, unless I use `-strip`. But that also removes the ICC profile.

I’m kinda stuck here, to be honest. It seems, like there are several ways of storing GPS and EXIF data in PNG, as that is standardized in newer versions of the spec. Maybe IM’s implementation has a bug here or I’m getting it all wrong. There’s a ton of fine-grained options for defining what the PNG encode should keep, but none of these does what I expect. PNG images can also contain additional information, like e.g. what background color a viewer application should use, rendering intent and gamma correction – currently we just strip all of these (and I’m not eager to deal with all of these).

The safest way would be to extract the ICC profile to a separate file, use `-strip` to remove all meta data and the `-profile path/to/profile.icc` options together, because that ensures that all other meta data is cleared. But that would require us to call `convert` twice and would require a few more changes to the ImageMagick driver. Should I go down that route, what do you think?

lukasbestle:
I think all image-related PNG chunks (background color etc.) can still be in the resulting image (at least they wouldn't harm). I'm only worried about the "metadata" chunks that could have privacy and/or security implications.

I googled together the following command that might (?) work for PNGs and JPEGs:

```sh
# delete all profiles except ICC, delete comment, delete PNG chunks (except image-related ones)
convert +profile "!icc,*" +set comment -define png:include-chunk=none,bKGD,cHRM,gAMA,iCCP,sRGB,tRNS,zCCP
```

Haven't tested it though.

fabianmichael:
@lukasbestle Thanks for your suggestion, but this also fails. And I think, I finally found the reason: ImageMagick’s `png:include/exclude-chunk` allows us to define which chunks we want in the final image. We have the `iCCp`-header for this, but profile description, copyright and actual data are stored as `tEXt`/`zTXt` data by IM, so are EXIF and GPS data. It seems, there is no way to tell IM to remove certain text chunks, but keep others.

Weird, because the IM docs are mentioning an EXIF chunk (`eXIf`), but the tool I used for adding artificial GPS data (`exiftool`) has saved everything as text chunks, which seems to collide with the ICC profile. A standardized way of saving EXIF data to PNG files has only been added quite lately (https://ftp-osl.osuosl.org/pub/libpng/documents/pngext-1.5.0.html#C.eXIf), which explains why there seems to be a ton of different ways of saving metadata to PNG files.

I tried to export a smartphone photograph with GPS data as PNG from Capture One 21, but that saves everything in a custom `tXMP` chunk. What a mess … 🤯

lukasbestle:
Well, the goal of my command was exactly that: Remove all the text chunks. There's no need to have description, copyright and other information in a thumb. After all these chunks were already removed with `-strip` before. So we don't need to differentiate what kind of text is stored (which is impossible because the contents of text chunks is not standardized anyway).

fabianmichael:
@lukasbestle Okay then … what if we only keep ICC profiles for JPEGs then (and maybe WebP as well), where we know that this works reliably because metadata is more standardized? JPEGs are also more likely to contain color-critical data. Maybe I should also open a bug in IM’s tracker and et them know, that it is currently not possible to strip text chunks while keeping the profile?

lukasbestle:
I think we are talking past each other. 😄

As far as I understood the PNG spec, there are two ways to embed ICC profiles in PNGs: There's the special `iCCP` chunk and there is the option to store the profile in a text chunk (for which ImageMagick has the special name `zCCP`). Both of these are allowlisted in the command I posted. `EXIF` data is not important for our use case.

Chunks we do definitely want in PNGs: `iCCP`, `zCCP`
Chunks that are nice to have: `bKGD`, `cHRM`, `gAMA`, `sRGB`, `tRNS`
Chunks that need to be removed (privacy and/or security): `eXIf`, `tEXt`, `zTXt`, `date` (and any other chunk not mentioned above; we need an allowlist approach)

Isn't this exactly what the command I posted is supposed to do? You wrote that it fails: In which way?

**Edit:** Ah, I think I understand what you mean. Seems like ICC profiles are sometimes stored in other chunk types besides `iCCP` and standard `tEXt`/`zTXt` chunks. IMO it's better to preserve *some* ICC profiles (those that are stored in the three recognized chunk types) than to remove *all* ICC profiles (which was happening before with `-strip`). We need to strip all actual metadata for privacy and security (see my first post in this PR), so if there is at least a route that preserves some useful color information, I think we should rather go that route than to strip all optional PNG chunks.

fabianmichael:
@lukasbestle I added a solution, which unfortunately adds more complexity than I would like. The ImageMagick driver now extracts the ICC profile to a separate file before processing the actual image, but only for PNG files (for JPEG/WebP, this way is not feasible, because they would be encoded twice). It also ensures, that date/time chunks are not added by IM. Also added a WebP image with metadata for an additional test. As far as my understanding goes, IM writes its own chunks after and unfortunately, they have decided to use text chunks for storing ICC profile data.

You just run the PHPUnit tests to validate, if this works. I’m not sure about Windows, because I had to redirect the output of `convert` by appending `2>/dev/null`, otherwise PHP would echo IM’s error message, when an input image has no profile. We could get around this by using `identify` instead of `convert`, but Kirby does not (and should not have) a dedicated setting for that.

BTW, if you don’t want to run the whole test suite just to check this, I’ve added a test scenario locally that only executes the image tests. Just add these lines to the `<testsuites>` in `phpunit.yml.dist`:

```yml
<testsuite name="Images">
    <directory>./tests/Image/</directory>
</testsuite>
```

lukasbestle:
To be honest I feel like two passes of ImageMagick are too much just for ICC profiles, even if it's only about PNGs.

What do you think about my thoughts in my previous post?

fabianmichael:
@lukasbestle I agree, but for this I have to find a way of creating a test PNG where the ICC profiles are stored in the "correct" way. IM even has a parameter to preserve original ICC data. That means, I have to test the PNG output of different applications against `pngcheck` and/or `exiftool` to figure out, how they store profile data. These can then be used as inputs for Kirby/IM and after that I can run Kirby’s tests to figure out, whether IM has preserved the original ICC data. Oh man, this is really getting exhausting. Who could have known, that PNG is such a messed-up format … 🙄

lukasbestle:
Don't worry – if it's difficult to generate a test case for ICC in PNG, then we can leave out this test for now. Most important is that comments, EXIF data and other metadata (in PNG text chunks) is still removed as before. As long as there's no regression, we are fine and can still improve the ICC handling in the future. Then we'll at least have proper ICC handling for JPEGs and still some improvement for PNGs immediately.

In any case I think we should go back to just one call to `convert`, otherwise it gets too complex.

fabianmichael:
@lukasbestle Okay, made a quick test again. I’m pretty confident to say now, that keeping the ICC profile as is while stripping all other metadata. Although there is a `-define png:preserve-iCCP=true` option, it does not really work. IM will always save the ICC profile data to a text chunk or discard it completely.

I ended up with stripping all profiles from PNGs using `-strip`, but keeping the ICC profile for all other file types. That’s the best we can do for now, unless we want to call `convert`/`identify` multiple time and use a temporarly ICC file. It’s not ideal, but assuming that most PNGs are in sRGB colorspace anyway, it should not hurt too much in practical use.

lukasbestle:
@fabianmichael Sorry for the huge delay in reviewing this, the last weeks have been pretty busy. I have not forgotten about your PR and will pick it up as soon as possible after the 3.6 release for 3.6.1.

manspeterson:
I'm just a layperson when it comes to this, and are probably missing something.

However, wouldn't you accomplish the same thing by just removing the `-strip` step and substituting `-resize` with `-thumbnail`, yet in a more clean way? Or do the `-thumbnail` command do something more than what's described in the documentation?

https://imagemagick.org/script/command-line-options.php#thumbnail

bastianallgeier:
@lukasbestle @fabianmichael what can we do to get this reviewed and merged? I'm afraid I didn't have time to read through any of the comments so far. Is there a summary?

lukasbestle:
It's assigned to me, I'm on it. Might just take a few more days until I have caught up. :)

fabianmichael:
@bastianallgeier @lukasbestle I created an automated test and everything should be fine, unless you can spot anything that looks wrong/asks for trouble in the future/etc.

lukasbestle:
@manspeterson Thanks for your very helpful comment. It looks to be exactly what we need, unfortunately it doesn't work properly for PNGs again. For JPEG and WebP inputs everything works perfectly, but in PNGs it doesn't remove GPS and other metadata. :(

But for other file types it seems like a perfect solution and it's definitely better to have ICC support for JPEGs than no ICC support at all. I've implemented the `-thumbnail` option and also cleaned up the test code. The tests still pass, so it seems to be fine as far as I can tell.

@fabianmichael Could you please take a look at the changes and check if everything works as you expect? From my side it's ready to merge now for the 3.6.1 release on Tuesday.

lukasbestle (approved):

lukasbestle:
Hm, GitHub Actions seems to be having problems again. The PHP 7.4 test run has locked up and can't even be cancelled and restarted...

bastianallgeier (approved):

PR #3654, opened 2021-08-26
Add loading as an allowed attribute to the video helper

MaluNoPeleke:
## Describe the PR
Added the loading attribute to the video helper
https://web.dev/lazy-loading-video/

### Enhancements
Added the loading attribute to the video helper

## Breaking changes
None

## Related issues/ideas

- Fixes #

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Toolkit/Html.php
+++ b/src/Toolkit/Html.php
@@ -465,6 +465,7 @@ class Html extends Xml
             // allow fullscreen mode by default
             'allowfullscreen' => $attr['allowfullscreen'] ?? true,
             'class'           => $attr['class'] ?? null,
+            'loading'         => $attr['loading'] ?? null,
             'height'          => $attr['height'] ?? null,
             'width'           => $attr['width'] ?? null,
         ];


afbora:
I didn't know there was a `loading` attribute. Frankly, I couldn't see it on the page you mentioned or in other sources. Could the attribute you mentioned be `preload`? 

I applied some attributes to the video tag:
https://github.com/getkirby/kirby/blob/efe24a6ef4ad61010428b1cb725f8f97a633ff82/config/tags.php#L293-L302

MaluNoPeleke:
Sorry, I used the wrong URL as a source. Please take a look at this one: https://web.dev/iframe-lazy-loading/

rasteiner:
@afbora https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-loading :)

It's experimental and works in chromium based browsers. But there seems to be general consensus on the feature.
Here's [Webkit](https://bugs.webkit.org/show_bug.cgi?id=196698) and [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1622090)

The attrs you added are for the `<video>` tag. But kirby uses an iFrame for the `video()` helper. 
Since we're already here, I'd add also "referrerpolicy", "allow" and "csp" to the list

lukasbestle:
> Since we're already here, I'd add also "referrerpolicy", "allow" and "csp" to the list

Are these really relevant for video iframes? The only relevant `allow` value that comes to my mind is `fullscreen`, which is already covered. `referrerpolicy` could make sense for additional privacy. For `csp` I don't really see a use case for video.

afbora:
Why don't we allow custom attributes? If we combine basic and custom attributes, it becomes a flexible structure allowing all kinds of different outputs.

For examples:

````html
<!-- UIkit -->
<iframe src="https://www.youtube.com/embed/a8BykUxjFZY" uk-video="autoplay: false" width="1920" height="1080" frameborder="0" allowfullscreen uk-cover></iframe>

<!-- YouTube Default Embed Code -->
<iframe width="560" height="315" src="https://www.youtube.com/embed/a8BykUxjFZY" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
````

**Comment by lukasbestle** [comment]
I wonder why we have the `videoAttr` method in the first place. In its docblock it says "Generates a list of allowed attributes for video iframes", so it seems to be some kind of security allowlist. But on the other hand the `$attr` array is passed manually by the developer, so I don't really see a security relevance. Maybe I'm missing something.

If there is indeed no risk, we could make the `videoAttr` method use `array_merge()` so that it only provides the defaults (for `allowfullscreen` in this case).

**Comment by afbora** [comment]
Definitely. What is allowed may mean supported attributes.

**Comment by lukasbestle** [comment]
Good point. Then I think it should be this unless @bastianallgeier has concerns about it:

```php
    /**
     * Generates a list of attributes
     * for video iframes
     *
     * @param array $attr
     * @return array
     */
    public static function videoAttr($attr): array
    {
        // allow fullscreen mode by default
        $attr['allowfullscreen'] = $attr['allowfullscreen'] ?? true;

        return $attr;
    }
```

We could also use the opportunity to convert `allowfullscreen` to `allow="fullscreen"` as `allowfullscreen` is considered a legacy attribute. Given that `allow` wasn't allowed (😛) before, I don't think there can be any special cases here. Basically if ` allow` is not configured, the value of `allowfullscreen` (or `true` as a fallback) would be used to determine the value of the `allow` attribute.

bastianallgeier:
It's one of those parts in our code, where I don't remember the decision behind the allow list anymore 🙈  I think it's a good idea to sanitise the allow option though.

afbora:
I've tried to implement but I couldn't find solution about escaping/encoding attribute value.

Input:
````php
$attr = ['allow' => 'camera \'none\'; microphone \'none\''];
````

Output:
````html
<iframe allow="camera &#039;none&#039;; microphone &#039;none&#039;">
````

https://github.com/getkirby/kirby/blob/develop/src/Toolkit/Xml.php#L114

lukasbestle:
Is it a problem? As far as I can tell, the browser still parses that correctly. It just doesn't look so clean, but it should work fine.

afbora:
Merged more flexible way on https://github.com/getkirby/kirby/pull/3680

PR #3656, opened 2021-08-29
Fix conditional sections with instant fields 

### LINKED ISSUE #3636: [3.6.0-alpha.4] Conditional sections with extending fields doesn't working
**Describe the bug**  

Following sample working on 3.5.7.1 but not for 3.6.0-alpha.4.

````yaml
columns:
  - width: 1/3
    fields:
      test:
        type: toggle
      
  - width: 2/3
    sections:
      test:
        extends: fields/test # includes multiple fields
        when:
          test: true
````

**fields/test.yml**

````yaml
type: fields
fields:
  foo:
    type: text
````

**Kirby Version**  

3.6.0-alpha.4

afbora:
## Describe the PR

If the section is conditional, it will not load instantly.

### Fixes

- Fix conditional sections with instant fields `#3636`

## Breaking changes

None

## Related issues/ideas

- Fixes #3636

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Panel/Model.php
+++ b/src/Panel/Model.php
@@ -348,7 +348,11 @@ abstract class Model
         if ($tab) {
             foreach ($tab['columns'] as $columnIndex => $column) {
                 foreach ($column['sections'] as $sectionIndex => $section) {
-                    if (in_array($section['type'], ['info', 'fields']) === true) {
+                    // makes sections ready if type is info or fields and conditions are empty
+                    if (
+                        in_array($section['type'], ['info', 'fields']) === true &&
+                        empty($section['when']) === true
+                    ) {
                         $tab['columns'][$columnIndex]['sections'][$sectionIndex] = $blueprint->section($sectionIndex)->toResponse();
                     }
                 }
--- a/tests/Panel/ModelTest.php
+++ b/tests/Panel/ModelTest.php
@@ -414,6 +414,26 @@ class ModelTest extends TestCase
                     [
                         'width'    => '2/3',
                         'sections' => []
+                    ],
+                    [
+                        'width'    => '1/1',
+                        'sections' => [
+                            'foo' => [
+                                'name'   => 'foo',
+                                'type'   => 'fields',
+                                'fields' => [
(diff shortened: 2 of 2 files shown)


bastianallgeier:
Looking at this I really start to wonder if we should revert the synchronous loading of the fields and info section for now. It seems that it has quite some complicated implications. I also found myself in situations where the lazy loaded fields really helped with performance – i.e. when there's a big blocks field that takes a while to load. 

Maybe we want too much at once here. 

@lukasbestle @distantnative what do you think?

lukasbestle:
I haven't worked a lot with the new field setup so far and I especially didn't make an AB test with an extreme setup. So I don't have a clear opinion to be honest.

I feel like the synchronous loading can be really great for many simple setups as it makes the Panel much snappier. But of course it shouldn't cause problems in more complex setups.

One idea could be to measure the field loading time on the first render and cache this information. On subsequent renders, the Panel could then dynamically decide to render the slow fields lazily. The end result of this would be awesome, but I'm not sure about the added complexity.

In any case: If you feel the synchronous rendering is not ready for prime time yet, then do revert it. We can still add it later.

bastianallgeier:
I agree that the synchronous loading of sections has a huge benefit. But such a performance check would be quite a hack in my opinion. When we want to turn this into a reliable solution, we would need to work on faster fields first and also rethink the conditional field loading logic. Especially with fields like blocks, layouts or structure, we currently solve their potential performance hits with lots of data, by loading all fields asynchronously. Unless we either find ways to fix the performance hits or load those individual fields asynchronously, it's not fully reliable to switch the entire fields section to synchronous loading. Combined with the issues of conditional fields, I agree that it's not ready for prime time yet. 

I will revert the synchronous loading.

distantnative:
I would think we also need to move the logic for conditional fields to PHP and rather solve updates of their visibility through Fiber (as we plan to do validation in PHP only as well, those calls would need to happen anyways).

distantnative:
Should we close this one because of https://github.com/getkirby/kirby/pull/3684?

PR #3657, opened 2021-08-30
Fix overlay scroll when destroyed not closed #3651

### LINKED ISSUE #3651: [3.6.0-alpha.4] Drawer sets body to `overflow: hidden`, doesn't get removed on navigation
**Describe the bug**  
When opening a drawer, the body gets an inline style for `overflow: hidden` to prevent scrolling the content behind the drawer. This style does only get removed when closing the drawer. The issue is you can navigate via the browser back buttons or click on links like files/pages in files/pages fields and it won't be removed.

**To Reproduce**  
1. Open a drawer
2. Click on e.g. a file/page in a files/pages field or use the browser back button
3. Try to scroll

**Expected behavior**  
I'd expect the `overflow: hidden` to be cleared on navigation.

**Kirby Version**  
3.6.0-alpha.4

distantnative:
## Describe the PR

There are situations where the Overlay component disappears without calling the `close` method (e.g. navigation away from view via back button). This PR ensures that the scroll lock is removed when the overlay component is destroyed.

### Fixes
- Overlay scroll lock is properly removed in certain situations

## Breaking changes

None

## Related issues/ideas

- Fixes #3651

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Layout/Overlay.vue
+++ b/panel/src/components/Layout/Overlay.vue
@@ -51,16 +51,26 @@ export default {
       scrollTop: 0
     };
   },
+  destroyed() {
+    // make sure scrolling is enabled again
+    // when overlay is destroyed without closing it
+    this.close(true);
+  },
   methods: {
-    close() {
+    close(destroyed = false) {
       // it makes it run once
       if (this.isOpen === false) {
         return;
       }
 
       this.isOpen = false;
-      this.$emit("close");
-      this.restoreScrollPosition();
+
+      // no need for event and restoring scroll
+      // when closed due to navigating away from view
+      if (destroyed === false) {
+        this.$emit("close");
+        this.restoreScrollPosition();
+      }
 
       // enable scrolling of background
       document.documentElement.style.overflow = "visible";
--- a/panel/src/fiber/app.js
+++ b/panel/src/fiber/app.js
@@ -18,6 +18,7 @@ export default {
         this.component = component;
         this.page = page;
         this.key = preserveState ? this.key : Date.now();
+        this.$store.dispatch("dialog", false);
       }
     });
   },


panel/src/components/Layout/Overlay.vue
```diff
@@ -51,16 +51,26 @@ export default {
...
       }
 
       this.isOpen = false;
-      this.$emit("close");
-      this.restoreScrollPosition();
+
+      // no need for event and restoring scroll
+      // when closed due to navigating away from view
+      if (destroyed === false) {
+        this.$emit("close");
```
lukasbestle, line 71: What could be listening for the `close` event? Could there be a case where the listening component needs to know about the closed overlay even when the component was destroyed?

distantnative: When testing my experience was the other way around: if the event was emitted, the form drawer was listening to it and wanted to focus some input that didn't exist anymore either. I think, the overlay really only gets destroyed without being closed when a whole view changes or similar, in which case all other components get destroyed as well. 

Or we would need to add more if statements to the listening components to make sure that their reactions might not be always possible anymore.

lukasbestle: To be honest I don't grasp the whole event setup around the `Overlay`. As far as I can tell, the only components that listen to the `close` event directly are `Dialog` and `Drawer`. But since events bubble up to the topmost component, I'm not sure about possible edge cases. It probably doesn't even make sense that this `close` event bubbles up, maybe we should explicitly stop it in the direct parent component. Given that `Dialog` and `Drawer` both emit another `close` event in their handlers for `Overlay`, it's all even more complicated. :(

bastianallgeier:
I just tried a few things and I think we have a bigger problem with dialogs. When I open a dialog and I navigate back with the back button, the dialog stays open and afterwards all kinds of weird issues happen – even with this pr. I think we need some form of solution that automatically closes all dialogs whenever you navigate in the browser. This should theoretically also solve this problem without the additional logic in the destroy method.

bastianallgeier:
Ah, it's indeed a combination of both. I just made a commit to show what I mean.

distantnative:
Could we store the instance/s of open dialogs in the store and not just set the store to false on navigation but actually call the close method on them? That way the destroy method hack might not be needed.

bastianallgeier:
Is it possible in Vue to store a reference to a component in the store?

distantnative:
That's what I was wondering. Need to try later on.

distantnative:
@bastianallgeier would it makes sense for this to combine the `dialog` Vuex store entry and the `drawers` Vuex module into something like `overlays`?

bastianallgeier:
@distantnative yes, definitely! Do you think we can find a solution for this quickly though? Otherwise, we might need to find a quick first solution for the beta and fix it afterwards.

distantnative:
Ok I think the solutions aren't even that complicated, we just do need to untangle them a bit and see which way we like most:

- Vuex store is currently a bit weird: we have a `drawers` module and a `dialog` property in the main store. However, the root of this issue is overlay instances that affect both. Question is whether we could either unite both current store implementations into one, we create a new `overlays` module or something else
- Your commit with clearing dialog store on component Fiber swap is good, but probably the drawers store should get the same treatment? Maybe we need a general "reset the store" for Fiber navigations. some parts of the store of course should keep their state.

PR #3658, opened 2021-08-31
Fix `timestamp()` helper step calculation

### LINKED ISSUE #3642: Helper function `timestamp` returns invalid dates.
When converting a date with the `timestamp` helper function (in kirby/config/helpers.php) and using the rounder feature,  the returned date becomes invalid.

**To Reproduce**  
echo timestamp('2021-08-18 10:59:00', 5); // => 2021-08-18 10:60:00 and strtotime('2021-08-18 10:60:00') becomes false

**Expected behavior**  
output should be: 2021-08-18 11:00:00

**Kirby Version**  
3.5.7.1

distantnative:
## Describe the PR

Looking at #3642, I realized that the `timestamp()` helper at the moment has quite a lot of flaws/non-functional cases.

The helper now properly calculates the nearest step corresponding to the step's unit and choose sensible reference points to do so (e.g. current century for years).

### TODO
- [x] phpmd issue blocking merging https://github.com/phpmd/phpmd/issues/910 (@lukasbestle)
- [ ] Ensure that JS implementation has no conflict with this change
- [ ] test case: 31.12.2021 23:59 rounding by 1 month (overflow into new year)

Thx to @nilshoerrmann for the ideas and input.

### Fixes
- `timestamp()` helper now rounds correctly to the nearest step 
- `timestamp()` helper allows passing a custom `base` reference for step rounding

### Breaking changes

_None_

## Related issues/ideas

- Fixes #3642

## Ready?

- [ ] CI checks pass

## When merging


config/helpers.php
```diff
@@ -800,50 +800,64 @@ function timestamp(?string $date = null, $step = null): ?int
...
+            $date = strtotime(date('Y-m-d H:00:00', $date));
+            break;
+        case 'day':
+            $size = 60 * 60 * 24;
+            $date = round($date / $size) * $size;
+            $date = strtotime(date('Y-m-d 00:00:00', $date));
+            break;
+        case 'month':
+            if ((int)date('d', $date) >= 15) {
+                $date  = strtotime('+20 day', $date);
```
afbora: Could use `next month` instead adding day

config/helpers.php
```diff
@@ -800,50 +800,64 @@ function timestamp(?string $date = null, $step = null): ?int
...
+            }
+            $date = strtotime(date('Y-m-01 00:00:00', $date));
+            break;
+        case 'year':
+            if ((int)date('m', $date) >= 7) {
+                $date  = strtotime('+7 month', $date);
```
afbora: Could use `next year` instead adding month

config/helpers.php
```diff
@@ -800,50 +800,64 @@ function timestamp(?string $date = null, $step = null): ?int
...
+            $date = strtotime(date('Y-m-d 00:00:00', $date));
+            break;
+        case 'month':
+            if ((int)date('d', $date) >= 15) {
+                $date  = strtotime('+20 day', $date);
+            }
+            $date = strtotime(date('Y-m-01 00:00:00', $date));
+            break;
+        case 'year':
+            if ((int)date('m', $date) >= 7) {
```
afbora: Could use `idate('n', $date)` instead casting.

config/helpers.php
```diff
@@ -800,50 +800,64 @@ function timestamp(?string $date = null, $step = null): ?int
...
+            $date = round($date / $size) * $size;
+            $date = strtotime(date('Y-m-d H:00:00', $date));
+            break;
+        case 'day':
+            $size = 60 * 60 * 24;
+            $date = round($date / $size) * $size;
+            $date = strtotime(date('Y-m-d 00:00:00', $date));
+            break;
+        case 'month':
+            if ((int)date('d', $date) >= 15) {
```
afbora: Could use `idate('j', $date)` instead casting

config/helpers.php
distantnative: Sadly not. Next month seems to add a fixed numbers of dates. E.g. if the date is 2020-08-31, it jumped to October not September when using next month.

config/helpers.php
distantnative: Most times true in this case. I'd assume that would only fail in a leap year on the 366 day... it's a tiny edge case but maybe better to stick to the current (though less elegant) solution?

config/helpers.php
afbora: Ohh, I didn't know this bug. So your solution solves this problem with a single line of code 👍

config/helpers.php
afbora: I didn't fully understand the edge cases. 2020 leap year and my tests:

- `next year` for `2020-02-29 23:59:59` output is `2021-03-01 23:59:59`
- `next year` for `2020-12-31 23:59:59` output is `2021-12-31 23:59:59`
- `next year` for `2020-01-01 00:00:00` output is `2021-01-01 00:00:00`

nilshoerrmann:
We recently built a frontend calendar and used `DateInterval` and `DatePeriod` to create recurring events in Kirby. We stored a starting date and then created an array of occurences for the calendar boundaries.

Couldn‘t something like this be helpful in this case? If you store a reference date and set an interval based on your step settings, you could create a list of valid dates and then choose the closest one.

The reference date could be relative like “first day in month” or alike which would give you a more meaningful starting point then “1970-01-01”. And the good thing about using intervals is that PHP will handle all the edge cases for you.

nilshoerrmann:
In case you are interested in this approach, this is how it would work for minutes:

```php
<?php

$date = new DateTimeImmutable('2021-08-18 10:59:00');
$step = new DateInterval('PT5M'); // 5 minute interval

$start = $date->modify('midnight'); // reference point to start steps from, 00:00 in this case
$end = $date->add($step);

$period = new DatePeriod($start, $step, $end);
$normalized = null;
foreach ($period as $occurence) {
    if (
        !$normalized ||
        abs($date->getTimestamp() - $occurence->getTimestamp()) <
            abs($date->getTimestamp() - $normalized->getTimestamp())
    ) {
        $normalized = $occurence;
    }
}

print_r($normalized); // 2021-08-18 11:00:00.000000
```

You would only have to define a different interval for the other units, adding the step value dynamically, the rest should work for all units. The reference point could be an additional method parameter.

E. g. using a 1 month interval `P1M` with the reference point `first day of this month` would return `2021-09-01 10:59:00.000000` which is the closest first day of a month. Using a 10 year interval `P10Y` with `2005-01-01` as reference point would return `2025-01-01 10:59:00.000000`.

distantnative:
Started something in pseudocode, just to try to think through if this approach would make sense:

```php

function normalize($value, $step, $base = null) {

  if ($base === null) {
    switch ($step['unit']) {
      case 'second':
        $base = $value->modify('beginning of minute');
        break;
      case 'minute':
        $base = $value->modify('beginning of hour');
        break;
      case 'hour':
        $base = $value->modify('beginning of day');
        break;
      case 'day':
        $base = $value->modify('beginning of month');
        break;
      case 'month':
      case 'year':
        $base = $value->modify('beginning of year');
        break;
    }
  }

  if ($value > $base) {
    $start = $base;
    $end   = $value->add($step);
  } else if ($value < $base) {
    $start = $value->minus($step);
    $end   = $base;
  } else {
    return $value;
  }

  $options = new DatePeriod($start, $step, $end);

  foreach ($options as $option) {
    if (abs($value - $options) <= $step/2) {
      return $value;
    }
  }

  return null;
}
```

nilshoerrmann:
I think the reference points for the steps have to be decoupled from the actual given dates:

- counting years usually starts with the beginning of the century,
- counting months starts on January 1st,
- counting days on the first of the month,
- counting times at midnight, 0:00.

So the base needs to be shifted to the prior occurence in relation to the current date.

distantnative:
But these are the fallbacks for when no reference date is given. 

If someone sets a step of 7 years, I'd think they rather expect 2014, 2021, 2028 instead of 1970 .., 2019, 2026, 2034.

distantnative:
Ah sorry, misread yours. I think the pseudopods above
Does this already except for years. Yes there maybe beginning of century would be an option.

nilshoerrmann:
Yes! 

And you are right that 1970-01-01 never makes sense in this context. It‘s just something programmers are familiar with – not a point of reference in real life.

nilshoerrmann:
I added a `dateRounded` helper to our date methods plugin based on the above which:
https://github.com/hananils/kirby-date-methods/blob/master/index.php#L322-L359

It uses PHP intervals which allows for "uneven" steps like `1 year, 2 months, 5 minutes` (if this might ever be needed). But the more easy-to-read array syntax used by Kirby could be transliterated easily.

Thinking of real-life usage in the frontend, it would be interesting to also have helpers like `dateCeil` or `dateFloor` if you think of rounding dates to decades or centuries (useful in archival contexts).

## Rounding Examples

Given date: `2021-08-18 10:59`

- `P1Y` (one year steps): `2022-01-01 00:00`
- `P10Y` (10 year steps): `2020-01-01 00:00`
- `P100Y` (100 year steps): `2000-01-01 00:00`
- `P1M` (1 month steps): `2021-09-01 00:00`
- `P1D` (1 day steps): `2021-08-18 00:00`
- `PT1H` (1 hour steps): `2021-08-18 11:00`
- `PT5M` (5 minute steps): `2021-08-18 11:00`

distantnative:
@nilshoerrmann Thanks for the pointers and massive help. I just pushed an updated version, still needs polish and more tests, but I think we are getting somewhere.

afbora:
When unit invalid passed like `['unit' => 'not-exists']`, getting following error:

> `Whoops\Exception\ErrorException : Undefined variable: interval`

We can throw a more descriptive exception when a disallowed unit is sent.

config/helpers.php
```diff
@@ -790,60 +790,87 @@ function tc(string $key, int $count)
...
-        str_pad($parts['month'], 2, 0, STR_PAD_LEFT) . '-' .
-        str_pad($parts['day'], 2, 0, STR_PAD_LEFT) . ' ' .
-        str_pad($parts['hour'], 2, 0, STR_PAD_LEFT) . ':' .
-        str_pad($parts['minute'], 2, 0, STR_PAD_LEFT) . ':' .
-        str_pad($parts['second'], 2, 0, STR_PAD_LEFT)
-    );
-
-    // on error, convert `false` into `null`
-    return $timestamp ?? null;
+    return null;
```
afbora: In what case does this line work? 🤯 😅

config/helpers.php
```diff
@@ -790,60 +790,87 @@ function tc(string $key, int $count)
  */
 function timestamp(?string $date = null, $step = null): ?int
 {
+    // only allow valid dates as input
     if (V::date($date) === false) {
         return null;
     }
 
+    // convert date to timestamp
     $date = strtotime($date);
```
nilshoerrmann, line 799: What happens if `$date` cannot be interpreted by `strtotime`?

nilshoerrmann:
@distantnative: That's looking great! For full flexibility, you might even add `base` to the steps array to allow developers to set their own reference point.

config/helpers.php
afbora: A few lines up `V::date($date)` takes care of it.

config/helpers.php
nilshoerrmann: This return statement is the fallback if the period `$options` are empty if I understand it correctly.

config/helpers.php
afbora: @nilshoerrmann You may be right 🤔 

````php
$date = 'january';

var_dump(strtotime($date));
var_dump(date_parse($date));
````

**Output**

````
int(1611183600)

array(12) {
  ["year"]=>
  bool(false)
  ["month"]=>
  int(1)
  ["day"]=>
  bool(false)
  ["hour"]=>
  bool(false)
  ["minute"]=>
  bool(false)
  ["second"]=>
  bool(false)
  ["fraction"]=>
  bool(false)
  ["warning_count"]=>
  int(1)
  ["warnings"]=>
  array(1) {
    [8]=>
    string(27) "The parsed date was invalid"
  }
  ["error_count"]=>
  int(0)
  ["errors"]=>
  array(0) {
  }
  ["is_localtime"]=>
  bool(false)
}
````

config/helpers.php
distantnative: I felt like there should be an ultimate return in case nothing before matched. But I agree that I with the current setup, I think this line can't ever really been reached. 🤔

config/helpers.php
distantnative: @afbora but that looks correct - we wouldn't want to allow just `january` as input and `V::date()` will not allow it - or am I reading this wrong?

afbora: You read it right, I thought wrong 🙂 The both return different results, but `V::date()` works first.

lukasbestle:
I think it might make sense to disable PHPMD until the bug is fixed. Waiting until the fix doesn't really make sense. What do you think?

distantnative:
@lukasbestle that would be good, can you do that?

bastianallgeier:
@distantnative is this ready to be merged now?

distantnative:
@bastianallgeier i would appreciate if someone else first really looks at the tests, if the results all make sense or if we are missing test cases

So it's ready for review - without just trusting the tests.

distantnative:
Oh but actually also one vector I haven't tested myself yet (but der needs to): does this different rounding logic clash with the date field and its current logic in JS

Issue #3644, opened 2021-08-23
When debug is set to false and there is only one blueprint available, Kirby uses the default blueprint instead of the defined blueprint


texnixe:
**Describe the bug**  
If a section only allows one template and the `debug` option is set to false in config, the default template is used instead of the defined template in Kirby 3.6.0-alpha.4 , because `$template` is then set to null in `dialogs.php` instead of being assigned the template name.

**To Reproduce**  
1. In a 3.6-alpha-4 StarterKit, set `'debug'  => false` in config
2. Open the Panel and create a new `photography` or `notes` subpage.
3. Notice how it is created with the default template instead of the `album` or `note` templates.

**Expected behavior**  
The defined blueprint should be used.

**Kirby Version**  
3.6.0-alpha.4

lukasbestle:
Affected lines:

https://github.com/getkirby/kirby/blob/a337eb974b2515016cce0103203a7df68aa24aab/config/areas/site/dialogs.php#L276-L283

To be honest I don't understand that `if`.

distantnative:
The line is there to hide the template select field if only one template option is available anyways (except when in debug mode, then always show the field).

But sounds like something in the logic of assigning what the one-and-only available value should be is off.

lukasbestle:
OK, yeah, that makes sense.

Then the issue is in this line:

https://github.com/getkirby/kirby/blob/a337eb974b2515016cce0103203a7df68aa24aab/config/areas/site/dialogs.php#L311

There should be logic here to use the configured template if there was no `template` field in the form. Probably it's enough to just have a fallback to the first available template if `get('template')` is empty. Then we don't need to duplicate the entire logic from the `load` method.

BTW: Where do we validate that the `get('template')` value is valid and allowed? Wouldn't it be possible at the moment to pass a completely different non-allowed template in the request?

bastianallgeier:
✅

PR #3659, opened 2021-09-01
Enhancements to page position UI

### LINKED ISSUE #3609: Page status dialog: pre-select position 1 for only public page
More detailed description: https://kirby.nolt.io/290

distantnative:
### Enhancements
- Page position field is hidden from dialogs when only one option available (except when in `debug` mode)
- Sort page is disabled in page dropdown when only one option would be available

## Breaking changes

None

## Related issues/ideas

- Fixes #3609

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Panel/Field.php
+++ b/src/Panel/Field.php
@@ -122,11 +122,8 @@ class Field
 
         // if only one available option,
         // hide field when not in debug mode
-        if (count($options) < 2 && option('debug') !== true) {
-            return array_merge([
-                'type'   => 'hidden',
-                'value'  => $options[0],
-            ], $props);
+        if (count($options) < 2) {
+            return static::hidden();
         }
 
         return array_merge([
--- a/src/Panel/Field.php
+++ b/src/Panel/Field.php
@@ -120,11 +120,20 @@ class Field
             'text'  => $index
         ];
 
+        // if only one available option,
+        // hide field when not in debug mode
+        if (count($options) < 2 && option('debug') !== true) {
+            return array_merge([
+                'type'   => 'hidden',
+                'value'  => $options[0],
+            ], $props);
+        }
+
         return array_merge([
-            'label'   => t('page.changeStatus.position'),
-            'type'    => 'select',
-            'empty'   => false,
-            'options' => $options
+            'label'    => t('page.changeStatus.position'),
+            'type'     => 'select',
+            'empty'    => false,
+            'options'  => $options,
         ], $props);
     }
 
--- a/src/Panel/Page.php
+++ b/src/Panel/Page.php
@@ -121,11 +121,13 @@ class Page extends Model
         ];
 
         if ($view === 'list') {
+            $siblings = $page->parentModel()->children()->listed()->not($page);
+
             $result[] = [
                 'dialog'   => $url . '/changeSort',
                 'icon'     => 'sort',
                 'text'     => t('page.sort'),
-                'disabled' => $this->isDisabledDropdownOption('sort', $options, $permissions)
+                'disabled' => $siblings->count() < 2 || $this->isDisabledDropdownOption('sort', $options, $permissions)
             ];
         }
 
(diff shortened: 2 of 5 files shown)


lukasbestle:
To be honest I'm not sure about this. If we hide the field completely in these cases, users could be confused as they expect that field to be normally there.

Possible alternative: Make the field disabled instead.

distantnative:
We have the same behavior with the template select on page creation (hiding field if only one option is available)

lukasbestle:
Yes, we do. I was thinking about this when I brainstormed about https://github.com/getkirby/kirby/issues/3644. But I came to the conclusion that it makes sense there as the template is just an "implementation detail" that isn't as relevant for the user. For example when you create a new blog post as the child of a blog page, you don't think about which template you need to choose. It's pretty obvious. So it makes some amount of sense to hide the selector in this case.

Now that I think about the same question here, I feel like it's more or less the same. It definitely makes sense to disable that menu option to change the position. Regarding the field in the status change dialog I also tend to agree with you now.

lukasbestle:
But I think we should think about the different display in `debug` mode again. It can cause confusion like "why does it look different in production": https://github.com/getkirby/kirby/issues/3663

bastianallgeier:
I agree with @lukasbestle that we don't really need the debug toggle in this case. For the template, you might want to actually "debug" which template is used. For the position it's not relevant.

distantnative:
> I agree with @lukasbestle that we don't really need the debug toggle in this case. 

Maybe I understood Lukas wrong, but I think his argument was more that our UI should look different for the `debug` mode for both use cases/dialogs. Not that one should be shown, the other always hidden.

bastianallgeier:
I would love to see a more prominent debug mode visual in the panel somewhere. I think this could also help to avoid unwanted debug mode in production. But in this case I would simply not make a difference between the hidden field in debug mode or prod mode.

src/Panel/Page.php
```diff
@@ -121,11 +121,13 @@ public function dropdown(array $options = []): array
         ];
 
         if ($view === 'list') {
+            $siblings = $page->parentModel()->children()->listed()->not($page);
```
bastianallgeier: Can't we use `$page->siblings()` here?

src/Panel/Field.php
```diff
@@ -120,11 +120,20 @@ public static function pagePosition(Page $page, array $props = []): array
             'text'  => $index
         ];
 
+        // if only one available option,
+        // hide field when not in debug mode
+        if (count($options) < 2 && option('debug') !== true) {
```
bastianallgeier: I'd simply remove the second check for the debug option and merge this thing.

src/Panel/Page.php
bastianallgeier: Forget about it. It's needed this way.

src/Panel/Field.php
```diff
@@ -120,11 +120,20 @@ public static function pagePosition(Page $page, array $props = []): array
...
+            return array_merge([
+                'type'   => 'hidden',
+                'value'  => $options[0],
```
bastianallgeier: I think this is wrong. It should probably be $options[0]['value']

bastianallgeier:
I made a few changes. Have a look at them and let me know what you think about it. We might want to add tests, but otherwise I'd be happy with this PR

src/Panel/Page.php
```diff
@@ -120,16 +120,14 @@ public function dropdown(array $options = []): array
             'disabled' => $this->isDisabledDropdownOption('changeStatus', $options, $permissions)
         ];
 
-        if ($view === 'list') {
```
distantnative, line 123: You want to show this option now as well on the page itself, not just in list view?

bastianallgeier: Yesterday when I tested it, I tried it in list view and then switched to the page view and wondered why the option isn't there at all. It felt like a bug. In this case the status dialog was disabled via permissions but sorting was still possible. So there was no way to change the position of the page via the page view at all.

distantnative: 👍

bastianallgeier (approved):

PR #3664, opened 2021-09-07
Change default YouTube video embed domain

yandodov:
## Describe the PR
YouTube redirects all non-www URLs to their www counterparts. Therefore, Kirby should use `www.youtube.com` by default to avoid unnecessary redirects which lead to poor SEO.

### DIFF
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -321,8 +321,8 @@ class Collection extends Iterator implements Countable
     /**
      * Alias for `Kirby\Toolkit\Collection::filter`
      *
-     * @param string|Closure $field
-     * @param array ...$args
+     * @param string|array|\Closure $field
+     * @param mixed ...$args
      * @return static
      */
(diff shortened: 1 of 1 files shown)


yandodov:
It looks like this is fixed on the `develop` branch, so it should be working in the next Kirby release.

afbora:
I guess it's not fixed. Whatever the URL is sent, it is returned as the host:

https://github.com/getkirby/kirby/blob/develop/src/Toolkit/Html.php#L532

yandodov:
@afbora oh, I searched for a hardcoded `"youtube.com"` (as it was before), didn't find one, and assumed it should be fine, for some reason...

Well, it just looks like this:

```php
$host   = 'https://' . $uri->host() . '/embed';
```

...should be constant:

```php
$host   = 'https://www.youtube.com/embed';
```

From what I've seen, YouTube pushes the www version as its main domain, so I think it's safe to assume that everything else will be redirected.

PR #3666, opened 2021-09-08
Use model id for model views

### LINKED ISSUE #3655: [3.6.0-alpha.4] Storing content in the accounts area fails with message "No route found for path account..."
**Describe the bug**  
When accessing an account via the account view (as opposed to via `users/userid` and try to save content, a dialog pops up with the error message "The form could not be saved. Exception: undefined No route found for path: "account" and request method: "PATCH" .

**To Reproduce**  
1. In a Starterkit, go to Panel => Your account
2. Make some content changes
3. Try to save changes

**Expected behavior**  
The content should be stored without error.

**Screenshots**  
Bildschirmfoto 2021-08-28 um 10 45 10

**Kirby Version**  
3.6.0-alpha.4 

**Console output**  

MacOs with Firefox

**Additional context**  
User can be updated without issues from `panel/users/:userid`

bastianallgeier:
## Describe the PR

To use the path for API calls was a very unstable solution. The model id is more reliable as it will always follow the same pattern as the API route. 

I.e. 

```
site
pages/blog+something
users/peter
```

## Related issues/ideas
- Fixes https://github.com/getkirby/kirby/issues/3655

### DIFF
--- a/panel/src/components/Views/ModelView.vue
+++ b/panel/src/components/Views/ModelView.vue
@@ -64,7 +64,7 @@ export default {
     content() {
       this.$store.dispatch("content/create", {
         id: this.id,
-        api: this.$view.path,
+        api: this.id,
         content: this.model.content
       });
     },


afbora (approved):
Perfect! Oh, was it that easy? 😅

bastianallgeier:
@afbora yes, it seems so :)

PR #3668, opened 2021-09-09
Filter unreadable pages in search

### LINKED ISSUE #3650: [3.6.0-alpha.4] Fiber search does not respect `read: false`
**Describe the bug**  
The new Fiber search does not seem to respect setting `read: false` in page blueprints. The respective pages still appear.

**To Reproduce**  
1. Add `read: false` to a page blueprint
2. Search for the page in the Panel

**Expected behavior**  
I expect the page to not appear in the search. This works fine fine in 3.5.7.1.

**Screenshots**  
Screenshot-25-08-2021_17-06-12@2x
```yml
title: Startseite
options:
  read: false
```

**Kirby Version**  
3.6.0-alpha.4

bastianallgeier:


afbora:
@bastianallgeier https://github.com/getkirby/kirby/pull/3621/commits/91f266fd8b4b6c0a5fdc677ed4ee0daee4d55fe1 ?

bastianallgeier:
@afbora since your PR is pretty extensive and we are still discussing the details, we might want to get this fix in first and replace it by your more global solution later.

afbora:
Actually, I wanted to point out that it should be fixed with `filter()` to standartize: `->filter('isReadable', true)`

bastianallgeier:
@afbora I've rewritten it.

afbora:
Shall we take the `filter` first? (filter() > search())

bastianallgeier:
@afbora I've thought about it, but I think it makes more sense this way. Filtering by permissions is quite costly, so I think it's better to search first and then filter. But I haven't done any performance tests.

afbora (approved):

PR #3672, opened 2021-09-09
Return plugin asset response instead of redirect

### LINKED ISSUE #2422: Plugin Asset Resolver Always Redirects to Same URL on First Call (Creating symlink)
**Describe the bug**  
I noticed that when trying to load plugin assets, such as `http://localhost/media/plugins/my-test-org/test-plugin/foo.js`, the first response from the server is `302` redirect to the same URL. The second time the URL is hit, the file is returned.

https://github.com/getkirby/kirby/blob/4c5e2c1feb1b9052902f8ad7d6e8248dc1eaafb7/src/Cms/PluginAssets.php#L59-L80

What seems to happen is: the symlink is created, and a redirect is returned to (presumably) the same file. It's then up to the server (Apache or NGINX) to supply the file contents.

It totally makes sense for the symlink to happen and have NGINX pick up on that for subsequent calls, but it means the _first_ request for the plugin asset never returns the actual file.

Would it make more sense to avoid having a redirect altogether and let Kirby return the file contents itself? It just seems odd to have a server return a 302 redirect to the same URL it's requesting.

**Kirby Version**  
`3.3.3`

bastianallgeier:
## Describe the PR

Returning a plugin asset directly after requesting it makes a lot more sense than creating a redirect.

### DIFF
--- a/src/Cms/PluginAssets.php
+++ b/src/Cms/PluginAssets.php
@@ -67,9 +67,6 @@ class PluginAssets
 
                 $target = $plugin->mediaRoot() . '/' . $filename;
 
-                // create the plugin directory first
-                Dir::make($plugin->mediaRoot(), true);
-
                 // create a symlink if possible
                 F::link($source, $target, 'symlink');
 
--- a/src/Cms/PluginAssets.php
+++ b/src/Cms/PluginAssets.php
@@ -66,15 +66,14 @@ class PluginAssets
                 static::clean($pluginName);
 
                 $target = $plugin->mediaRoot() . '/' . $filename;
-                $url    = $plugin->mediaUrl() . '/' . $filename;
 
                 // create the plugin directory first
                 Dir::make($plugin->mediaRoot(), true);
 
-                if (F::link($source, $target, 'symlink') === true) {
-                    return Response::redirect($url);
-                }
+                // create a symlink if possible
+                F::link($source, $target, 'symlink');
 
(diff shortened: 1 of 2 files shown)


bastianallgeier:
This one should be really simple to review: @getkirby/kirby-staff

distantnative:
I can do this evening.

afbora:
I was able to test it. It works flawlessly 👍 

Minor changes that no longer need `$url` variable:
https://github.com/getkirby/kirby/blob/fix/2422-direct-plugin-asset-response/src/Cms/PluginAssets.php#L69

src/Cms/PluginAssets.php
```diff
@@ -66,15 +66,14 @@ public static function resolve(string $pluginName, string $filename)
                 static::clean($pluginName);
 
                 $target = $plugin->mediaRoot() . '/' . $filename;
-                $url    = $plugin->mediaUrl() . '/' . $filename;
 
                 // create the plugin directory first
                 Dir::make($plugin->mediaRoot(), true);
```
afbora: `F::link()` already create dir for target in above symlink line. I think we no longer need this line.

afbora: Done https://github.com/getkirby/kirby/pull/3672/commits/c268f54ed39f3bb593c87c0d4a2271afa5e9d597

afbora (approved):

Issue #3257, opened 2021-04-19
Video block does not show domain-restricted clips in Panel


sebastiangreger:
As the Panel sets a `same-origin` referrer policy for privacy/infosec reasons,

https://github.com/getkirby/kirby/blob/38afb2d650f3764d23d732b23ae8f04ab9ffba69/views/panel.php#L20

the iframe embed of the video block correctly does not communicate the referrer domain to Vimeo and the embedded page hence shows a *"Sorry, Because of its privacy settings, this video cannot be played here."* error rather than the video (as [brought up on Discord](https://discord.com/channels/525634039965679616/761608160565133343/833639343347859456)).

This should be possible to fix — analogous to the solution [discussed on the forum](https://forum.getkirby.com/t/vimeo-embed-in-panel/21226) and [in the related kirby-embed plugin's issue](https://github.com/sylvainjule/kirby-embed/issues/5) — by adding a `referrerpolicy="strict-origin-when-cross-origin"` attribute to the iframe code at:

https://github.com/getkirby/kirby/blob/38afb2d650f3764d23d732b23ae8f04ab9ffba69/panel/src/components/Blocks/Types/Video.vue#L11

bastianallgeier:
✅

PR #3677, opened 2021-09-10
Add referrerpolicy to the video block

bastianallgeier:
## Describe the PR

The Video block does not show domain-restricted clips in Panel. With this fix we can solve the problem without compromising the privacy/security of Panel users. 

## Related issue

https://github.com/getkirby/kirby/issues/3257

### DIFF
--- a/panel/src/components/Blocks/Types/Video.vue
+++ b/panel/src/components/Blocks/Types/Video.vue
@@ -9,7 +9,7 @@
     @update="update"
   >
     <k-aspect-ratio ratio="16/9">
-      <iframe v-if="video" :src="video" />
+      <iframe v-if="video" :src="video" referrerpolicy="strict-origin-when-cross-origin" />
     </k-aspect-ratio>
   </k-block-figure>
 </template>


panel/src/components/Blocks/Types/Video.vue
```diff
@@ -9,7 +9,7 @@
     @update="update"
   >
     <k-aspect-ratio ratio="16/9">
-      <iframe v-if="video" :src="video" />
+      <iframe v-if="video" :src="video" referrerpolicy="strict-origin-when-cross-origin" />
```
afbora, line 12:
```suggestion
      <iframe v-if="video" :src="video" referrerpolicy />
```

`strict-origin-when-cross-origin` value is already default for `referrerpolicy`:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#directives

afbora:
`strict-origin-when-cross-origin` value is already default for `referrerpolicy`:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#directives

I don't know very well, but are we sure it works exactly? When I add video in the panel, it already shows as `Referrer Policy: strict-origin-when-cross-origin` in network tab of devtools 🤔

bastianallgeier:
As far as I understood it, it's not yet implemented in every browser as the default. Maybe @sebastiangreger has more input.

sebastiangreger:
Looking at the [implementation status](https://caniuse.com/mdn-http_headers_referrer-policy_default_strict-origin-when-cross-origin) of that default across different browsers it is no surprise IE did not receive this Nov 2020 spec update any more, but Safari indeed only [closed that bug](https://bugs.webkit.org/show_bug.cgi?id=218909) a few weeks ago (not sure is the compat table out of date or has this indeed not been rolled out in Safari yet). Globally, 27% of users still use browsers that use the old default of `no-referrer-when-downgrade`

However, if I understand this PR correctly, it addresses the issue that the `<meta name="referrer" content="same-origin">` tag in the panel's `head` section ([/views/panel.php#L20](https://github.com/getkirby/kirby/blob/c77ccb82944b5fa0e3a453b4e203bd697e96330d/views/panel.php#L20)) alters that default to a stricter policy for the Kirby Panel at large …which in consequence led to problems when embedding Vimeo videos that may require a `referer` header if restricted to embedding in certain domains ([Forum thread](https://forum.getkirby.com/t/vimeo-embed-in-panel/21226)).

Setting `referrerpolicy="strict-origin-when-cross-origin"` on an `<iframe>` element, as this PR does, overrides the page-wide default from the `<meta>` tag so that the browser applies the more permissive policy to the iframe and its contents. Now Vimeo would receive the domain information in a `referer` header and could provide embedding for a domain-restricted video.

Personally, I strongly believe that `same-origin` remains the most sensible default for a CMS backend. The old browser default of `no-referrer-when-downgrade` would potentially leak URLs of unlisted pages if an admin clicks an outgoing link to a third party site; hence I suggested this addition back in the days. It could of course be debated whether `strict-origin-when-cross-origin` might be a sufficient level of limitation to set in that meta tag for the panel in general – it would avoid similar issues as you are now fixing for Kirby's own video block for others dealing with such embeds in their custom blocks etc. However, there might still be a risk involved when revealing referrers to a third party, even if all they see is the domain and no longer the full path as it was before.

bastianallgeier:
@sebastiangreger that's how I understood it as well. I think we should keep the strict setting in the meta tag. It's pretty simple to soften the permissions on an element basis afterwards and I feel better with a strong default.

afbora (approved):

PR #3680, opened 2021-09-12
Allow custom attributes on video helper

afbora:
## Describe the PR

Custom attributes can now be added to iframes for video helper. For fullscreen mode, the deprecated `allowfullscreen` attribute has been changed to the new `allow` attribute.

### Enhancements

- Custom iframe attributes allowed on video helper

## Breaking changes

None

## Related issues/ideas

None

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/tags.php
+++ b/config/tags.php
@@ -314,6 +314,21 @@ return [
                     $video = Html::tag('video', [$source], $attrs);
                 }
             } else {
+                // removes attributes that providers (youtube and vimeo) don't support in iframe
+                if (
+                    Str::contains($tag->value, 'youtu', true) === true ||
+                    Str::contains($tag->value, 'vimeo', true) === true
+                ) {
+                    unset(
+                        $attrs['autoplay'],
+                        $attrs['controls'],
+                        $attrs['loop'],
+                        $attrs['muted'],
+                        $attrs['poster'],
+                        $attrs['preload']
+                    );
+                }
+
                 $video = Html::video(
                     $tag->value,
                     $options,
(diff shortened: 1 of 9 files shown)


config/tags.php
```diff
@@ -314,6 +314,21 @@
...
+                    unset(
+                        $attrs['autoplay'],
+                        $attrs['controls'],
+                        $attrs['loop'],
+                        $attrs['muted'],
+                        $attrs['poster'],
+                        $attrs['preload']
+                    );
+                }
+
```
afbora: I had to remove attributes in video kirbytag that iframe doesn't support.

bastianallgeier: This is probably exactly the reason why I added the allow list instead.

afbora: I guess, yes 😅

afbora:
So what would be an option for user defined allowed attributes and defaults like in `kirbytext.video` instead of allowing everything?

bastianallgeier:
I think the main problem here is that we could easily define an allow list based on standard attributes, but we also want to allow custom attributes like those for the UI framework. In this case we cannot really use an allow list or we simply allow any attribute with a prefix, which is pretty complicated again. 

I have to say that I'm not really happy about the massive unset.

afbora:
What do you think splitting attributes like in https://github.com/getkirby/kirby/pull/3680/commits/2fc1866f02c8a0b3a1875a416e711a2cbeb2f0d1 ?

bastianallgeier:
That's pretty clever!

bastianallgeier (approved):

PR #3683, opened 2021-09-14
New icons

### LINKED ISSUE #3493: [3.6.0] add a few more icons
- [ ] export (https://kirby.nolt.io/231)

Maybe we should collect some urgent ideas on Discord. We shouldn't add too many more but another 5-10 crucial ones maybe?

bastianallgeier:
## Describe the PR
I've collected the icons from the icon wishlist on Discord that seemed really useful for a wider audience. 

Screenshot 2021-09-16 at 13 01 29

New icon names: 

```
- audio
- crop
- display
- flag
- food
- grid
- grid-bottom
- grid-bottom-left
- grid-bottom-right
- grid-full
- grid-left
- grid-right
- grid-top
- grid-top-right
- grid-top-left
- h1
- h2
- h3
- h4
- h5
- h6
- heart-outline
- info
- images
- mobile
- order-alpha-asc
- order-alpha-desc
- order-num-asc
- order-num-desc
- paypal
- pin
- pipette
- play
- print
- qr-code
- server
- shuffle
- sitemap
- star-outline
- sun
- table
- tablet
- text-center
- text-justify
- text-left
- text-right
- tiktok
```

Aliases: 

```
- text => text-left
- filter => funnel
```


lukasbestle (commented):
Can we also have a reversed order icon (Z to A)? The names could be `order-asc` and `order-desc`.

Question: Was it intended that the dist files are removed?

afbora:
Marked done but not exists

- analytics / trend graph
- photo
- filter

The 3 icons below are pretty generic, can't we add them?

- pdf
- images
- table

screen-capture-1159-IcoMoon App - Icon Font, SVG, PDF   PNG Generator-icomoon io

bastianallgeier:
@lukasbestle I only added native Nucleo icons so far. For a reversed order I will need to modify that order icon. No big deal though.

The removed dist files where a mistake. I've just added them. 

@afbora I've marked them as done, because they already exist in the old icon set. 

- https://getkirby.com/docs/reference/panel/icons/image
- https://getkirby.com/docs/reference/panel/icons/chart
- https://getkirby.com/docs/reference/panel/icons/funnel

For the other three I wasn't able to find good ones in the Nucleo icon set. I don't want to mix icon sets. I already hate that we have the file icons with thinner borders in there.

afbora:
Single and multiple usage areas can be different. Like cover/featured image(image) - slider/gallery(images), product item (tag) - products (tags), user - users.

bastianallgeier:
I've created new order icons …

Screenshot 2021-09-14 at 16 10 40

lukasbestle:
Good idea to also make numeric ones! But shouldn't the arrow also change direction (pointing up on desc)?

afbora:
> But shouldn't the arrow also change direction

Because 9 → 1

lukasbestle:
Yes, but to be honest I've only ever seen descending icons with an arrow pointing in the opposite direction. It would also make the two icons easier to distinguish (without having to actually read the letters/numbers). I'm not 100% sure which of the icons should have the arrow pointing down as I've seen both variants in the wild.

afbora:
> It would also make the two icons easier to distinguish (without having to actually read the letters/numbers)

Yes, it has both uses. But I agree with you about distinguish easily.

bastianallgeier:
@lukasbestle @afbora it's difficult. I feel like the current version is more logical, but you might be right that the arrows are visually stronger and therefor easier to read …

Screenshot 2021-09-15 at 09 48 19

bastianallgeier:
Upper row or lower row?

afbora:
Its small appearance is also important.

screen-capture-1166-133392661-b0e53d4e-6fa7-411d-833d-797d137fac1d png (737×454)_-user-images githubusercontent com

nilshoerrmann:
The lower row is more common and has a more distinct look.

afbora:
> Upper row or lower row?

I guess, lower row +1

> Single and multiple usage areas can be different. Like cover/featured image(image) - slider/gallery(images), product item (tag) - products (tags), user - users.

@bastianallgeier Also what do you think about `images` icon?

lukasbestle:
What if the letters/numbers are also flipped?

So basically:

```
A |   Z ^
Z v   A |

1 |   9 ^
9 v   1 |
```

If you feel that's too confusing, then 👍 for your suggestion in your second row.

bastianallgeier:
@lukasbestle Isn't it technically incorrect then? This is seriously messing with my brain 🤣

nilshoerrmann:
Maybe you have to consider two different ways to look at this: the icons either represents a status (this is how the data is currently served) or they represent an action (this is how the data will be served). In the first case, Bastian's first row icons make sense as the letter order is compared to the reading direction, in the second case the second row icons suit better because they show that the reading direction is going to change.

Lukas' suggestions mixes both points of view which is logically incorrect in my eyes.

lukasbestle:
I see my suggestion more like this: Both the letters/numbers *and* the arrows show the actual order (the arrow always pointing from the lowest to the highest value just like in the original icon). For me the letters/numbers are there to define the order and the arrows are there support this information visually.

I wouldn't say it's about status vs. action. You can still use the same icon for both: In a button it would be an action and in a table header it would be a status.

Unfortunately this is one of these very subjective things we could discuss for hours. 😄

This is what Icomoon does BTW (= Bastian's original suggestion):

Icomoon

And FontAwesome has loads of them (which confirms that the topic is very very subjective 🤣):

FontAwesome

bastianallgeier:
@lukasbestle hahaha, I love how FontAwesome is just using the "We are not going to discuss this shit" solution. 

I agree that it's in general very subjective. I cannot really make up my mind which version I like best. But I'd say we should start with the lower row and see if there's any additional feedback afterwards.

bastianallgeier:
@afbora I've just added a new images icon

afbora:
I think it's been great, thanks 🎉 

I have a little question about current `text` icon, just idea.
Now we have 3 new text icons. 

- text-center
- text-justify
- text-right

But missing `text-left` icon (actually it is `text` icon).

1. Can we reference the `text-left` icon as `text`?
2. Or we can update the text icon to text-left and assign a new icon to the text icon.

bastianallgeier:
@afbora there are a few icons I'd love to have aliases for. But I'm not sure how we could solve that. Do you know about a way to create aliases in SVG?

afbora:
@bastianallgeier I found a solution with using `xlink:href`. I've tested and working perfect!

````xml
<symbol id="icon-text-left" viewBox="0 0 16 16">
  <path d="M0 1h16v2H0V1zM0 5h10v2H0V5zM0 9h16v2H0V9zM0 13h10v2H0v-2z" />
</symbol>
<symbol id="icon-text" viewBox="0 0 16 16">
  <use xlink:href="#icon-text-left" />
</symbol>
````

nilshoerrmann:
Just so you know, `xlink:href` is deprecated in favour of `href`:  
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href

bastianallgeier:
I've created two aliases

text = text-left
filter = funnel

If you find more good aliases, please let me know. But I think it would be great if we could already merge this to get the icons done for now.

distantnative (approved):

afbora (approved):

PR #3685, opened 2021-09-16
Extendable areas

bastianallgeier:
## Describe the PR
This PR introduces a new architecture for core parts of Kirby. So far, we mostly use PHP files in the `/kirby/config` folder to define lists of methods, areas, blueprints, etc and we load them in the AppPlugins trait. This was fine, but it was really hard to access core parts later when you need them to fall back to core methods in plugins. You would need to reload the same definition files again and go through arrays and method, which is pretty wasteful. 

There's now a new `Kirby\Cms\Core` method instead which lists all core parts of Kirby in simple class methods. This is a lot cleaner to read and reduces the number of files we load from the config folder. But most importantly, all of those core parts and lists can be access again later through the new `$kirby->core()` method. So for example you can now easily get a list of all kirbytags by simply running `$kirby->core()->kirbytags()`

While the Core class only returns arrays with paths to files, lists of aliases or class names, you still need a way to actually load the wanted code for a kirbytag or a panel area somehow. This is now handled by a new `Kirby\Cms\Loader` class. 

So for example you can now do `$kirby->load()->area('site')` to get the full definition area for the site area. This will give you the site area, including possible extensions that might come from plugins. But sometimes you actually need to access core code without extensions, especially while extending core stuff in plugins. 

That's why the Core class also has a load method: `$kirby->core()->load()->area('site')` does the same as the load method above, but it will return the pristine core code. Those two new ways to load parts of Kirby are perfect ways to access either the possibly extended version or the native version at anytime, anywhere in the code.

The entire change was necessary to make extendable areas possible and provide a useful way to create dialogs, dropdowns, searches, etc. that might still fall back to their core definition if needed. 

Here are the examples I've posted in our Discord team channel: 

### extended search

```php
<?php

Kirby::plugin('example/search', [
    'areas' => [
        // extending a core search
        'site' => function ($kirby) {
            return [
                'searches' => [
                    'pages' => [
                        'query' => function (string $query) {
                            // run your own search here
                            return [
                                [
                                    'image' => [
                                        'src'   => 'https://source.unsplash.com/random',
                                        'cover' => true,
                                        'back'  => 'white'
                                    ],
                                    'text'  => 'Test Result',
                                    'link'  => '/site',
                                    'info'  => 'test info'
                                ]
                            ];
                        }
                    ]
                ]
            ];
        },
    ]
]);
```

### extended dropdown

```php
<?php

Kirby::plugin('example/dropdown', [
    'areas' => [
        'site' => function ($kirby) {
            return [
                'dropdowns' => [
                    'pages/(:any)' => function (string $id) {
                        $page     = Find::page($id);
                        $dropdown = $page->panel()->dropdown([
                            'view'   => get('view'),
                            'sort'   => get('sort'),
                            'delete' => get('delete')
                        ]);

                        return [
                            $dropdown['preview'],
                            '-',
                            $dropdown['changeTitle'],
                            $dropdown['duplicate'],
                            '-',
                            $dropdown['changeSlug'],
                            $dropdown['changeStatus'],
                            $dropdown['changeSort'],
                            $dropdown['changeTemplate'],
                            '-',
                            [
                                'icon'   => 'share',
                                'text'   => 'Publish on Netlify',
                                'dialog' => 'publish'
                            ],
                            '-',
                            $dropdown['delete']
                        ];
                    }
                ]
            ];
        },
    ]
]);
```

### extended dialog

```php
<?php

Kirby::plugin('example/dialog', [
    'areas' => [
        'site' => function ($kirby) {
            return [
                'dialogs' => [
                    'pages/(:any)/changeStatus' => 
                        'load' => function (string $id) {
                            $page = Find::page($id);

                            if ($page->id() === 'blog') {
                                return [
                                    // some custom shit.
                                ]
                            }

                            return $kirby->core()->area('site')['dialogs']['pages/(:any)/changeStatus']['load']($id);
                        },
                        'submit' => function (string $id) {
                            $page = Find::page($id);
                            $status = get('status');

                            // some custom submit action

                            return true;
                        }
                    ]
                ]
            ];
        },
    ]
]);
```

The last example also shows how to access the original code for the dialog via the `$kirby->core()->area()` method (a shortcut for `$kirby->core()->load()->area()` if you want to fall back to it. 

## Breaking changes
None

## Related issues/ideas

- Fixes https://github.com/getkirby/kirby/issues/3486

## Ready?

- [ ] CI checks pass

## When merging


src/Cms/App.php
```diff
@@ -836,6 +847,16 @@ public function languages(bool $clone = true)
         return $this->languages = Languages::load();
     }
 
+    /**
+     * Load a part of Kirby
```
distantnative: Seems to be a bit misleadingly phrased. Maybe more `Access Kirby's part loader`?

src/Panel/Panel.php
```diff
@@ -54,45 +53,51 @@ public static function area(string $id, array $area): array
...
+            'settings',
+            'users'
+        ]);
+
+        $result = [];
 
-        // load plugins
-        foreach ($kirby->extensions('areas') as $id => $area) {
-            if (is_a($area, 'Closure') === false) {
-                throw new InvalidArgumentException(sprintf('Panel area "%s" must be defined as a Closure', $id));
```
distantnative, line 86: We don't need this check anymore?

bastianallgeier: After thinking about it, there's actually no harm in defining an area or an extension of a part of an area as a simple array when you don't need access to the $kirby var.

distantnative: So far there was a harm because a simple array wouldn't work and cause problems. If now a simple array is supported, indeed we should allow that.

distantnative:
Looking at this

```php
                        $page->panel()->dropdown([
                            'view'   => get('view'),
                            'sort'   => get('sort'),
                            'delete' => get('delete')
                        ]);
```

Should we bake `get('view')` etc. as defaults into the `dropdown()` method?

src/Panel/Panel.php
distantnative: Tested and works now. 👍

distantnative (dismissed):

bastianallgeier:
@distantnative I normally try to avoid putting query param getters directly into such methods, but I think in this case it would massively simplify possible plugins.

distantnative:
True. I mean, if it's a security concern, maybe query params are only used as default when passing an explicit boolean, e.g.

```
$page->panel()->dropdown(true)
```

Still takes a lot more mindspace from developers to understand what they have to do, but is less to write and maybe safer?

bastianallgeier:
I don't think it needs to be a security concern. There's nothing that can be injected here as far as I can see. I think an additional "allow query params" flag would be more confusing than just using the query params as default and make them overwriteable via the argument array.

distantnative (approved):

PR #3694, opened 2021-09-21
Refactor async JS code

distantnative:


afbora:
I don't know if it's important, but I think await is also missing in some async api methods (`/panel/src/api`). Because some of them use await while others don't. For example, await was used when creating the language, but not when creating the page or user.

bastianallgeier:
@afbora's right. There's a lot of inconsistency in api methods. Sometimes we wait, sometimes we don't. @distantnative I assume that waiting is not really necessary if you return a method that already awaits. But I'm not 100% sure.

distantnative:
I think if a method is async, there's no difference whether you do normal return or return await.

bastianallgeier:
Then we should remove some unnecessary awaits.

bastianallgeier (dismissed):

afbora:
Just one more: https://github.com/getkirby/kirby/blob/fix/async-code/panel/src/fiber/index.js#L296

bastianallgeier (approved):

PR #3695, opened 2021-09-21
`Obj::set()` method

distantnative:
## Enhancement
- New `Obj::set()` method

## Breaking changes

None

## Related issues/ideas

- Fixes #

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Toolkit/Obj.php
+++ b/src/Toolkit/Obj.php
@@ -23,8 +23,8 @@ class Obj extends stdClass
      */
     public function __construct(array $data = [])
     {
-        foreach ($data as $key => $val) {
-            $this->$key = $val;
+        foreach ($data as $key => $value) {
+            $this->set($key, $value);
         }
     }
 
@@ -54,7 +54,7 @@ class Obj extends stdClass
      * Magic property getter
      *
      * @param string $property
-     * @return mixed
+     * @return null
      */
     public function __get(string $property)
     {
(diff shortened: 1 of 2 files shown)


afbora:
There is a magic property getter and returns `null` but no setter. Is it normal?

distantnative:
I was also a bit confused why the magic setter returns null... but when I change that, everything blows up :D 
The magic setter is inherited from `stdClass` I think.

bastianallgeier:
The magic getter is only taking affect when the property does not exist. The concept behind this is that `$obj->doesNotExist` no longer throws an Exception with the magic getter but returns null instead. 

The logic within the getter is a bit weird though. It would totally be fine to write: 

```php
public function __get() {
  return null;
}
```

There's no magic setter because setting was handled via $obj->newProp = 'something'

To be honest, I'm still not sure if we actually need a set method here. The Obj class is intended to be the most basic upgrade for stdClass. When we introduce a set method, we also need to make sure to use that in a new magic setter and that there's a get method. Otherwise it wouldn't make sense in my opinion.

afbora:
Hmm, why we don't use like that? Doesn't it make more sense than returning `null`?

````php
public function __get(string $property)
{
    return $this->get($property);
}
````

bastianallgeier:
The magic getter is never called when the property exists.

distantnative:
You're right. Somehow my mind thought there was a `get()` method and thus `set()` was missing. But let's stay without both of them.

Issue #3140, opened 2021-02-05
Inconsistent return types between Query class insert method and shortcut Db::$queries['insert'] 


texnixe:
**Describe the bug**  
Not sure if this is already part of the ongoing refactoring, but while the method in the Query class returns either the ID of the inserted row or false, the shortcut always returns an integer. Is that on purpose?

**To Reproduce**  
```
$wrongData= [
  'name'    => 'moe',
  'lastnam' => 'moellers'
];

// returns 0
$id = Db::insert('users', $wrongData );

// returns false
$id = Db::table('users')->insert($wrongData);
```
**Expected behavior**  
The return values should be consistent.

**Kirby Version**  
3.5.1

bastianallgeier:
✅

Issue #2856, opened 2020-09-26
XHR: 400 Bad Request error on log-in attempt with invalid credentials


GiantCrocodile:
**Describe the bug**  
You get two errors during a log-in attempt with invalid credentials in the browser console.

1. a XHR 400 Bad Request request
2. a code 400 error because the password doesn't match:

```
code: 400
details: Array []
exception: "Kirby\\Exception\\InvalidArgumentException"
file: "/kirby/kirby/src/Cms/User.php"
key: "error.user.password.notSame"
line: 926
message: "Die Passwörter stimmen nicht überein"
route: "auth/login"
status: "error"
```

The second error is fine but I think the first one shouldn't happen or has a wrong error code.

**To Reproduce**  
1. Go to your panel to log in
2. Type in an existing email address
3. Type in a wrong password
4. Confirm the dialog
5. Check your browser console

**Expected behavior**  
Just one proper error message.

**Kirby Version**  
3.4.3

**Desktop (please complete the following information):**  
Win 10, latest Firefox, WAMP setup

**Additional context**  
- multilang website with just 1 lang
- if you do wait for the page to refresh, log in with other data (invalid user+pw), you get an 404 error from the API. Not sure if this is a different non-related issue.

distantnative:
> The second error is fine but I think the first one shouldn't happen

Why not? Isn't the POST request you sent (with the invalid login details) a request that fails?

GiantCrocodile:
According to my understanding the HTTP code 400 is used for invalid requests from clients where the request itself is malformed, e. g. missing parameters or invalid parameter values which are never valid for the given request/API. In this case a invalid password is still a valid request regarding its syntax and context. I think there is no other place where you get a HTTP error and a Kirby 3 error at the same time? If the intention was to signalize an invalid logon attempt, then a 403 or 401 might fit better, see https://stackoverflow.com/a/6937030 for further details.

distantnative:
@bastianallgeier @lukasbestle  I think `401` would be the fitting one then for this case. Agreed?

lukasbestle:
I agree!

bastianallgeier:
✅

PR #3697, opened 2021-09-22
Quick wins

bastianallgeier:
## Describe the PR
This PR closes three simple issues at once. I combined them although they are not related at all. But the fixes are so simple, that separate PRs felt too much.

- #2856
- #3140

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Database/Db.php
+++ b/src/Database/Db.php
@@ -186,7 +186,7 @@ Db::$queries['column'] = function (string $table, string $column, $where = null,
  *
  * @param string $table The name of the table which should be queried
  * @param array $values An array of values which should be inserted
- * @return int ID of the inserted row
+ * @return mixed Returns the last inserted id on success or false
  */
 Db::$queries['insert'] = function (string $table, array $values) {
     return Db::table($table)->insert($values);
--- a/src/Database/Db.php
+++ b/src/Database/Db.php
@@ -134,9 +134,10 @@ class Db
     }
 }
 
+// @codeCoverageIgnoreStart
+
 /**
  * Shortcut for SELECT clauses
- * @codeCoverageIgnore
  *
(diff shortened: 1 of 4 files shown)


bastianallgeier:
@getkirby/kirby-staff help. Codecov has a new interface again, or something's broken. Do you know how to see now where the missing coverage is?

src/Database/Db.php
```diff
@@ -190,7 +190,7 @@ public static function __callStatic(string $method, $arguments)
  * @param array $values An array of values which should be inserted
  * @return int ID of the inserted row
  */
-Db::$queries['insert'] = function (string $table, array $values): int {
+Db::$queries['insert'] = function (string $table, array $values) {
```
afbora, line 191: This line uncovered.

bastianallgeier: I don't really understand why there's a DbTest::testInsert method with a @coversNothing flag. In theory the line is definitely covered.

afbora: I guess we need `@codeCoverageIgnoreStart` and `@codeCoverageIgnoreEnd` for inline codes 🤔

distantnative: I think Ahmet is right, the `@codeCoverageIgnore` seems to not be applied at all to these inline closures, @lukasbestle .

distantnative:
@bastianallgeier I think removing the default `autocomplete=off` and `spellcheck=off` needs to be considered a breaking change as it can result in quite a different user experience using the Panel if the developer just upgrades and doesn't set it up according to their needs.

bastianallgeier:
@distantnative you are right. I haven't thought about it this way.

bastianallgeier:
I've reverted the TextInput changes

distantnative (approved):

PR #3698, opened 2021-09-22
`page.create` dialog: fix single template value

### LINKED ISSUE #3644: When debug is set to false and there is only one blueprint available, Kirby uses the default blueprint instead of the defined blueprint

distantnative:
### Fixes
- Page create dialog: correct template used when not in `debug` mode and only one template is allowed

## Breaking changes

None

## Related issues/ideas

- Fixes #3644

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/config/areas/site/dialogs.php
+++ b/config/areas/site/dialogs.php
@@ -263,7 +263,7 @@ return [
             $blueprints = $view->blueprints($section);
 
             // the pre-selected template
-            $template = null;
+            $template = $blueprints[0]['name'] ?? $blueprints[0]['value'] ?? null;
 
             $fields = [
                 'parent' => Field::hidden(),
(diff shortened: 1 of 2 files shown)


distantnative:
Can't figure out where `Page` is coming from
Screen Shot 2021-09-22 at 15 21 39

afbora:
`Page` is default template name and should be like that:

````php
$this->assertSame('Page', $props['value']['template']);
````

File: config/areas/site/dialogs.php [hunk 1]
```diff
@@ -263,7 +263,7 @@
             $blueprints = $view->blueprints($section);
 
             // the pre-selected template
-            $template = null;
+            $template = $blueprints[0]['name']  ?? $blueprints[0]['value'] ?? null;
```
afbora:
```suggestion
            $template = $blueprints[0]['name'] ?? $blueprints[0]['value'] ?? null;
```

distantnative:
@bastianallgeier why are we using `name` here as well and not just `value`? https://github.com/getkirby/kirby/blob/develop/src/Panel/Field.php#L209

bastianallgeier:
> @bastianallgeier why are we using `name` here as well and not just `value`? https://github.com/getkirby/kirby/blob/develop/src/Panel/Field.php#L209

It's the same as in the line above for the title. The blueprints already have a title and name prop and this way you don't have to convert the blueprints array to options before passing it to the field.

bastianallgeier (approved):

PR #3702, opened 2021-09-22
`toggle` field: fix for missing `text` prop

### LINKED ISSUE #3696: [3.6.0-beta.1] Toggle field error in the panel
The Toggle field shows an error in the panel in 3.6.0-beta.1

screenshot-localhost-2021 09 22-12_22_49

Localhost, Windows 10

distantnative:
### Fixes
- Toggle field: show on/off as defaults when no `text` prop is defined

### Fixed regression
- Toggle field without `text` prop on `site` working again

## Related issues/ideas

- Fixes #3696

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/fields/toggle.php
+++ b/config/fields/toggle.php
@@ -35,7 +35,11 @@ return [
                 return $value;
             }
 
-            return $model->toSafeString(I18n::translate($value, $value));
+            if (empty($value) === false) {
+                return $model->toSafeString(I18n::translate($value, $value));
+            }
+
+            return $value;
         },
     ],
     'computed' => [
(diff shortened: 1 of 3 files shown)


afbora:
This off-on is already default in frontend. Is there any other way? Text prop of toggle field should return `null` to defaults.

https://github.com/getkirby/kirby/blob/develop/panel/src/components/Forms/Input/ToggleInput.vue#L48

distantnative:
The issue is that our `toString`/`toSafeString` will return the id when `text` value is `null`. So the original `null` is not reaching The Vue component.

afbora:
What do you think about following return:

````php
'text' => function ($value = null) {
    $model = $this->model();

    if (is_array($value) === true) {
        ...
    }

    if (empty($value) === false) {
        return $model->toSafeString(I18n::translate($value, $value));
    }

    return $value;
},
````

distantnative:
I like

afbora (approved):

PR #3707, opened 2021-09-22
Additional Fiber unit tests

distantnative:
I think with these we could close https://github.com/getkirby/kirby/issues/3406

### DIFF
--- a/config/areas/settings/dialogs.php
+++ b/config/areas/settings/dialogs.php
@@ -174,15 +174,14 @@ return [
                 ]
             ];
         },
-        /**
-         * @codeCoverageIgnore
-         */
         'submit' => function () {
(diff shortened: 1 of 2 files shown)


afbora:
- [ ] `src/Panel/Document.php` 95.83%
- [ ] `src/Panel/Panel.php` 98.48%

distantnative:
I've looked at these. I think it's fine to leave them as is for now.

bastianallgeier (approved):

bastianallgeier:
@distantnative I agree. I think the coverage is already pretty awesome and we have to keep in mind that we bumped the overall coverage for Panel features quite dramatically with our Fiber changes. There's so much stuff tested now, which was not when everything was JS based.

PR #3714, opened 2021-09-24
New Panel wrapper for better composition options

bastianallgeier:
## Describe the PR

With 3.6 we offer new k-inside and k-outside wrappers for plugin views to decide what kind of view they want to build. This is great in general, but the k-inside wrapper always comes with the topbar and styling that does not work when you want to build a fullscreen view for logged in users. 

With the new k-panel wrapper, it's now possible to do one of the three: 

```html
<k-panel>
   <!-- full screen view code -->
</k-panel>
```

```html
<k-inside>
   <!-- inside view with topbar -->
</k-inside>
```

```html
<k-outside>
   <!-- outside view without everything -->
</k-outside>
```

k-inside and k-outside use k-panel as outer wrapper.


distantnative:
Do we really need an additional component? More components doing very similar things are more and more confusing for developers. I'd think a `fullscreen` prop for `k-inside` might be the better approach.

bastianallgeier:
I wanted to implement the fullscreen option first, but then I saw that there's no good reason that the outside component does not have the offline handling, error dialog, drag data attribute and a few more things.

distantnative:
Maybe then other way around: do we need inside and outside or couldn't Kirby figure this out? Or be a prop? I just find it relatively a lot of new markup that developers probably won't directly understand why, what for ...

bastianallgeier:
I think we should be careful to not make a single component too smart with lots of ifs. In this case I really prefer clear composition. But the question is if we really need to separate it into outside and inside after all. k-panel and k-outside are now mostly identical except the css class. Inside is only different by the added topbar and form buttons. 

Maybe we need to take a step back and think about the various use cases. What do you want to build as a plugin developer? 

1.) for unauthenticated views, you want a way to create simple screens like login and installation without any specific layout or interface elements. 

2.) most authenticated views will need the topbar

3.) most authenticated views will also need a header with a big title and centered content

4.) some authenticated views might want to go full screen like the blueprint editor without the toolbar or anything else. 

How can we make sure that all four situations are doable without much extra effort? Ideally it should be instantly visible what kind of view you are building.

distantnative:
Based on your 4 requirements, I could see two components

- `k-panel` as general wrapper for everything. 
- additionally `k-inside` if you need authentication. Prop to disable topper and header for those that want to go with a more custom UI (such as fullscreen - I honestly think a fullscreen without any topbar will be rather rare).

bastianallgeier:
Ok, here are a few things I've fixed in the last commit: 

- `k-outside` no longer exists. `k-panel` can be used instead.
- `k-form-buttons` are no longer placed in the `k-inside` component. Model views that need the form buttons have to include them. Too many views don't need the form buttons at all. 
- There are no additional props to remove the topbar. If you don't need the topbar, you can simply use `k-panel` as wrapper. The topbar + a main view wrapper for padding is pretty much the only difference now in `k-inside` 
- I've fixed a padding issue in the k-view[data-center] variant. The padding is no longer needed. 
- The k-search overlay is now directly placed in the topbar. This makes it easier to use the topbar separately if needed.

distantnative (approved):
Looks good to me.

To summarise:

```html
<k-inside>
   <!-- when you want the full experience with topbar -->
</k-inside>
```

```html
<k-panel>
   <!-- for everything else, no matter if authenticated or not -->
</k-panel>
```

PR #3715, opened 2021-09-24
Introduce more reliable file route pattern

### LINKED ISSUE #3187: Panel Bug with "Files" URL page
**To Reproduce**  
1. Create any page with "Files" URL
2. Try to edit any text field on this page
3. All typed data will dissapear

Video:
Mar-02-2021 12-53-54

**Kirby Version**  
3.5.3.1

**Console output**  
It seems Kirby is trying to lock/unlock "Files" page, but returns errors:

[Error] Failed to load resource: the server responded with a status of 400 () (lock, line 0)
[Error] Object
code: 400
details: [] (0)
file: "/kirby/src/Cms/Api.php"
key: "error.invalidArgument"
line: 141
message: "Invalid model type: ."
route: "(.*)/files/([a-zA-Z0-9\\.\\-_%= \\+\\@\\(\\)]+)"
status: "error"
Object Prototype
	onError (app.js:1:22082)
	(anonymous function) (app.js:1:17731)
	asyncFunctionResume
	(anonymous function)

bastianallgeier:
## Describe the PR

The file routes in the API were too greedy and caught everything with /files at the end. This led to a really weird issue when a page had the slug files. 

## Related issues/ideas

- Fixes #3187

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/api/routes/files.php
+++ b/config/api/routes/files.php
@@ -1,7 +1,7 @@
 <?php
 
 // routing pattern to match all models with files
-$pattern = '(site|users/.*?|pages/.*?)';
+$pattern = '(site|users/[^/]+|pages/[^/]+)';
 
 /**
  * Files Routes
--- a/config/api/routes/files.php
+++ b/config/api/routes/files.php
@@ -113,7 +113,7 @@ return [
         }
     ],
     [
-        'pattern' => $pattern . '/search',
+        'pattern' => 'files/search',
         'method'  => 'GET|POST',
         'action'  => function () {
             $files = $this
--- a/config/api/routes/files.php
+++ b/config/api/routes/files.php
@@ -1,12 +1,15 @@
 <?php
 
+// routing pattern to match all models with files
+$pattern = '(site|users/.*?|pages/.*?)';
+
 /**
  * Files Routes
  */
 return [
 
     [
-        'pattern' => '(:all)/files/(:any)/sections/(:any)',
+        'pattern' => $pattern . '/files/(:any)/sections/(:any)',
         'method'  => 'GET',
         'action'  => function (string $path, string $filename, string $sectionName) {
             if ($section = $this->file($path, $filename)->blueprint()->section($sectionName)) {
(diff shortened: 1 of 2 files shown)


distantnative:
It's a bit weird cause `/files` shouldn't have matched before with a path that would rather be `foo+files`

bastianallgeier:
I guess our all pattern really takes just everything.

distantnative:
But why does `(:all)/files/(:any)` match something like `pages/my-download-page+files`? The pattern says that `files` would need to be preceded by a slash. Is there rather something wrong in our router pattern matching?

config/api/routes/files.php
```diff
@@ -96,21 +99,21 @@
...
-        'pattern' => '(:all)/files/(:any)/name',
+        'pattern' => $pattern . '/files/(:any)/name',
         'method'  => 'PATCH',
         'action'  => function (string $path, string $filename) {
             return $this->file($path, $filename)->changeName($this->requestBody('name'));
         }
     ],
     [
-        'pattern' => 'files/search',
+        'pattern' => $pattern . '/search',
```
afbora: `files/search` doesn't match with new pattern, am I wrong?

bastianallgeier: Ok, that's a stupid mistake.

bastianallgeier:
@distantnative turns out the (:all) pattern is just a placeholder for (.*) and that's really taking everything. This means that it's quite dangerous to use it with anything that might come afterwards. It should probably only be used at the end of route patterns unless we find a way to stop at whatever might come afterwards.

distantnative:
But even if it takes everything, the following / would have to match, right? But that wouldn't be in the panel path as we use +

bastianallgeier:
I think we should really dig deeper why the (:all) pattern is behaving so weirdly. But in the meantime it would be great if we could approve this to fix the problem.

distantnative (dismissed):

config/api/routes/files.php
```diff
@@ -1,12 +1,15 @@
 <?php
 
+// routing pattern to match all models with files
+$pattern = '(site|users/.*?|pages/.*?)';
```
lukasbestle: I wasn't able to dive deeper into this, but as far as I understand it, user and page URIs never contain slashes (instead the `+` sign). If we want to make sure, we could use the following pattern:

```suggestion
$pattern = '(site|users/[^/]+|pages/[^/]+)';
```

The `+` instead of the `*` also makes sure that the user/page URI cannot be empty. The question mark at the end is not needed when the character group cannot include slashes.

I also don't really get why this PR fixes the issue as even `.*/files` would only ever match if `/files` follows after the page URI. Very weird.

bastianallgeier:
@lukasbestle you are right. I've added your pattern and also wrote another unit test.

afbora (approved):

PR #3717, opened 2021-09-26
Validate empty password

### LINKED ISSUE #3716: Confirming empty password dialog throws error
**Describe the bug**  
If you change the password of an account and don't provide any input in the dialog, submitting the dialog causes an exemption to be thrown.

**Expected behavior**  
Input properly validated and an information to the user shown.

**Kirby Version**  
v3.6 beta 1

**Console output**  
Kirby\Cms\UserRules::validPassword(): Argument #2 ($password) must be of type string, null given, called in [...]\kirby\config\areas\users\dialogs.php on line 181

Windows 10, Firefox

afbora:
## Describe the PR

`UserRules::validPassword` method doesn't accept `null` password. Fixed with passing empty string if `null`.

### 💡 Idea

Why don't we accept `null` in all methods in `UserRules` class. After all, if it's a validating class and `null` is invalid, so we can accept a `null` parameter and say it is invalid. 

Why do we use these methods when calling?  `$email ?? ''` or `$password ?? ''` Wouldn't it be better/stabil if we send `null` directly?

### Fixes

- Throw correct error message again when the password is empty `#3716`

## Breaking changes

None

## Related issues/ideas

- Fixes #3716 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/areas/users/dialogs.php
+++ b/config/areas/users/dialogs.php
@@ -178,7 +178,7 @@ return [
             $passwordConfirmation = get('passwordConfirmation');
 
             // validate the password
-            UserRules::validPassword($user, $password);
+            UserRules::validPassword($user, $password ?? '');
 
             // compare passwords
             if ($password !== $passwordConfirmation) {


bastianallgeier (approved):

bastianallgeier:
I agree that we should rethink null as possible value. I would handle that in a separate refactoring task for all rule classes though.

PR #3720, opened 2021-09-27
Fix inner join queries #2501

bastianallgeier:
We've never managed to finish our database refactoring and the fix for this is super simple. That's why I thought it'd be only fair to finally fix it. 

- Fix https://github.com/getkirby/kirby/issues/2501


distantnative (approved):
👍

PR #3721, opened 2021-09-27
Todos on copy/paste

afbora:
### Todos on #3712

- [x] Paste does not work if the block field is not empty. (only through the add dialog)
- [x] Unsupported block types can be pasted
- [x] Paste doesn't respect max option
- [ ] If only one block type available, it cannot be pasted because the selection dialog will not open.
- [ ] Copy any block that is not a code block. Then try to copy the code block as single or multiple. It is not copied.

### Possible features

- [x] Copy all button

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -9,6 +9,7 @@
   "collapse": "Collapse",
   "collapse.all": "Collapse All",
   "copy": "Copy",
+  "copy.all": "Copy all",
   "create": "Create",
 
   "date": "Date",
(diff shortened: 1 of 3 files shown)


afbora:
Ready to review, we'll look at the remaining two tasks in the coming days.

MaluNoPeleke:
I was wondering why it was always hard to paste something into list blocks (in my attempts it always required 2-3 tries and I couldn't figure out the "right" way).
Should this be fixed by this PR as well?

afbora:
@MaluNoPeleke I guess this is a different issue. Can you open a new issue about this?

MaluNoPeleke:
> @MaluNoPeleke I guess this is a different issue. Can you open a new issue about this?

Done: #3728

PR #3729, opened 2021-09-28
Feature/copy and paste exec

distantnative:
- [x] Translation strings
- [ ] Better icons
- [x] helper library for clipboard tasks

Implements https://kirby.nolt.io/254


panel/src/components/Forms/Blocks/Block.vue
```diff
@@ -205,6 +207,27 @@ export default {
         });
       }
     },
+    isInputEvent(e) {
```
distantnative: I think we should move this somewhere into a helper library

i18n/translations/en.json
```diff
@@ -9,6 +9,7 @@
   "collapse": "Collapse",
   "collapse.all": "Collapse All",
   "copy": "Copy",
+  "copy.all": "Copy all",
```
afbora, line 12: I guess we don't need this anymore?

bastianallgeier: Yes, I think we can remove it.

distantnative: I think we should use it in the field dropdown. Only "copy" doesn't help that much if it's not tied to a specific block

afbora: Since it is in the options of the block field, I think we can only use `copy`, but of course `copy all` is more understandable.

bastianallgeier: Ok, let's keep Copy all in the main dropdown and Copy in the block dropdown.

panel/src/components/Forms/Blocks/Blocks.vue
```diff
@@ -225,11 +225,21 @@ export default {
...
+        // filters only supported blocks
+        const availableFieldsets = Object.keys(this.fieldsets);
+        blocks = blocks.filter(block => availableFieldsets.includes(block.type));
+
+        // don't add blocks that exceed the maximum limit
+        if (this.max) {
+          const max = this.max - this.blocks.length;
+          blocks = blocks.slice(0, max);
+        }
+
```
afbora, line 225: I didn't intentionally put it in the append method so that it is checked only when paste is done. It will do a redundant check every time when adding normal blocks.

bastianallgeier: Ah, sorry. That makes sense.

bastianallgeier: After checking it out again, I really think it makes sense to put it into the append method. We normally pass a string to the append method when a new block type is created and in this case the filter is not applied. So we are fine for new blocks, but for every other task to append blocks, I think it's good to have this check.

panel/src/components/Forms/Blocks/Blocks.vue
```diff
@@ -238,8 +287,73 @@ export default {
...
+          blocks.push(block);
+        }
+      });
+
+      // don't copy if no blocks could be found
+      if (blocks.length === 0) {
+        return false;
+      }
+
+      const json = JSON.stringify(blocks, null, 2);
```
distantnative: I think moving the actual clipboard logic to a helper would be good, see https://github.com/getkirby/kirby/tree/feature/clipboard-helper

And also adding support for iOS devices (see branch)

panel/src/components/Forms/Blocks/Blocks.vue
```diff
@@ -358,6 +495,62 @@ export default {
         this.$refs["block-" + block.id][0].open();
       }
     },
+    async paste(e) {
+      let html = e;
+
+      // get the content from the clipboard if an event is given
+      if (e instanceof ClipboardEvent) {
```
distantnative: same here, I would move this to a clipboard paste helper, allows us later to extend the clipboard js functionality without having to jump into the component's code

panel/src/components/Forms/Blocks/BlockOptions.vue
```diff
@@ -60,6 +66,13 @@
           {{ $t("field.blocks.changeType") }}
         </k-dropdown-item>
         <hr>
+        <k-dropdown-item icon="template" @click="$emit('copy')">
+          {{ $t("copy") }}
+        </k-dropdown-item>
+        <k-dropdown-item icon="download" @click="$emit('paste')">
+          {{ $t("paste") }}
```
distantnative: I think this should be better "Paste after" to communicate where the element would be inserted

bastianallgeier:
I've created new helpers for the clipboard and also for keyboard to detect if the meta key is ctrl or cmd. I've fixed an issue with pasting in the block selector and also added translations. 

The only thing left now is better icons, but I think this will need to wait until beta.3

@afbora @distantnative @lukasbestle could you review again?

bastianallgeier:
Shit. There's still an issue with too many paste events being registered when you open the layout field. This needs to be solved first.

afbora:
I think it works great 🎉 By the way, I did not see any apparent copy/paste problem in the layout field.

bastianallgeier:
I've just refactored the event handling to register all document events only once and handle the rest with our $event bus. For some reason, the paste event wasn't properly removed when the component got destroyed. This leads to a very nasty effect when you go to a big layout field first, copy a block and then go to a single block field on another page and you try to paste that block there. The paste event suddenly gets fired multiple times and takes forever. 

With the refactored event handling, events seem to get unregistered properly now. 

I also converted all bus functions from anonymous functions. I read on Stackoverflow that anonymous functions cannot get removed properly at all.

bastianallgeier (approved):

PR #3730, opened 2021-09-29
Fix empty data value on frontend template parser

afbora:
screen-capture-1176-Test - Mægazine-localhost

Fixed above issue on https://github.com/getkirby/kirby/pull/3669

### DIFF
--- a/panel/src/helpers/string.js
+++ b/panel/src/helpers/string.js
@@ -80,7 +80,7 @@ export function stripHTML(string) {
 export function template(string, values = {}) {
   const resolve = function (parts, data = {}) {
     const part  = escapeHTML(parts.shift());
-    const value = data[part] || null;
+    const value = data[part] ?? null;
 
     if (value === null) {
       return Object.prototype.hasOwnProperty.call(data, part) || "…";
--- a/panel/src/helpers/string.template.spec.js
+++ b/panel/src/helpers/string.template.spec.js
@@ -12,7 +12,8 @@ describe("$helper.string.template", () => {
     user: {
       email: "bastian@getkirby.com",
       username: "bastian"
-    }
+    },
+    info: "",
   };
 
   it("should insert values", () => {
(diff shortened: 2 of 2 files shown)


panel/src/helpers/string.js
```diff
@@ -80,7 +80,7 @@ export function stripHTML(string) {
 export function template(string, values = {}) {
   const resolve = function (parts, data = {}) {
     const part  = escapeHTML(parts.shift());
-    const value = data[part] || null;
+    const value = data[part] !== undefined ? data[part] : null;
```
johannschopplich: A simpler solution: `data[part] ?? null`. Vite will transform optional chaining for us.

afbora: Yes you are right, when I used null coalescing operator before, npm run was giving an error when doing build. No error now. I think it's starting to fully support it.

panel/src/helpers/string.js
```diff
@@ -80,7 +80,7 @@ export function stripHTML(string) {
 export function template(string, values = {}) {
   const resolve = function (parts, data = {}) {
     const part  = escapeHTML(parts.shift());
-    const value = data[part] || null;
+    const value = data[part] ?? null;
```
bastianallgeier, line 83: Wait, since when is ?? supported in JS?

afbora: Yep! While working in JSFiddle, it was not working in our panel, but it seems to be working now. I guess it happened by updating our dependencies.

https://caniuse.com/mdn-javascript_operators_nullish_coalescing

johannschopplich: Since ES2020. 🙂 See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

There's also optional chaining, which might peak your interest as well!
```js
const kirby = {
  creator: {
    name: 'Bastian'
  }
}

console.log(kirby.creator?.nonExistingKey)
// expected output: undefined

console.log(kirby.someNonExistentMethod?.())
// expected output: undefined

``

panel/src/helpers/string.js
johannschopplich: Maybe prior to Vite? Vite will run the code through esbuild, which supports the nullish coalescing operator.

bastianallgeier (approved):

PR #3744, opened 2021-10-02
Upgrade npm dependencies & update Vite config

johannschopplich:
## Describe the PR

Probably a good idea while being in beta stage?

## Release notes

The following packages where upgraded: 

```
cypress                     ^8.2.0  →   ^8.5.0     
eslint-plugin-cypress      ^2.11.3  →  ^2.12.1     
eslint-plugin-vue          ^7.15.1  →  ^7.18.0     
postcss-dir-pseudo-class    ^5.0.0  →   ^6.0.0     
postcss-logical             ^4.0.2  →   ^5.0.0     
vite                        ^2.4.4  →   ^2.6.2     
vite-plugin-vue2            ^1.7.2  →   ^1.8.2     
vue-docgen-api             ^4.40.0  →  ^4.41.2     
dayjs                      ^1.10.6  →  ^1.10.7     
prosemirror-history         ^1.1.3  →   ^1.2.0     
prosemirror-schema-list     ^1.1.5  →   ^1.1.6     
prosemirror-view          ^1.18.11  →  ^1.20.1     
vue                        ^2.6.12  →  ^2.6.14     
```

Only two dependencies are available in a new major version. No breaking changes for this setup.
- postcss-dir-pseudo-class
- postcss-logical  

### Before

```
dist/css/style.css   102.64kb / brotli: 14.82kb
dist/js/index.js     304.16kb / brotli: 58.71kb
dist/js/vendor.js    383.11kb / brotli: 105.08kb
```

### After

> ℹ️ Vite shows gzipped output from 2.6 onwards, not brotli like before.

```
dist/css/style.css   102.46 KiB / gzip: 17.17 KiB
dist/js/index.js     310.19 KiB / gzip: 74.06 KiB
dist/js/vendor.js    385.04 KiB / gzip: 121.10 KiB
```

## Breaking changes

None found.

## Ready?

- [ ] CI checks pass

## When merging


distantnative:
Would be good to measure the list sizes before and after - sometimes nice to see if there is a significant effect, in whichever direction.

johannschopplich:
@distantnative Posted above. 

Notes:
- Vite uses esbuild as default minifier from 2.6+, since it is much more performant. Terser produces smaller bundles, thus use it moving forward.
- `require` doesn't need to be used in the Vite config, since it gets transpiled by esbuild to CJS before Vite actually loads the config.
- Use Vite's `defineConfig` to leverage IDE's intellisense and therefore TypeScript declarations when hovering over Vite config options.

distantnative:
> Vite uses esbuild as default minifier from 2.6+, thus lets use terser again, since it produces smaller bundles.

But according to the stats you post above, the bundle actually increases by about 8 kb with this PR. Do we know why?

johannschopplich:
> But according to the stats you post above, the bundle actually increases by about 8 kb with this PR. Do we know why?

Unfortunately, I don't… Probably some internals, before the minification actually kicks in. I will try to figure that one out. Nevertheless, I think upgrading to the latest dependencies is the way to go even if the bundle sizes increased for now.

distantnative:
I went digging...

- `prosemirror` upgrades add about 1.6 kb in vendor
- `vite-plugin-vue2` adds about 6 KB to our main JS

There I stopped digging. I looked into recent commits of `vite-plugin-vue2` but couldn't pin point what's going wrong. I am not sure if we should update this under these circumstances.

distantnative:
On the other hand, especially 1.8.2 contains quite a number of bug fixes...

johannschopplich:
I have glanced at the massive `vite-plugin-vue2` changelog of fixes as well. I'd consider upgrading, even if that means some extra kilobytes. Stability over byte size imho. You have done quite a lot through areas already. 😉

lukasbestle:
I fully agree with Johann. After all we cannot freeze the dependency versions forever just because there's slightly more code.

bastianallgeier:
I also agree that we should do this even if it adds a few more kb.

bastianallgeier (approved):

bastianallgeier:
@johannschopplich thanks for your help with this!

panel/vite.config.js
```diff
@@ -1,29 +1,28 @@
...
+import { defineConfig } from "vite";
 import { createVuePlugin } from "vite-plugin-vue2";
 import pluginRewriteAll from "vite-plugin-rewrite-all";
 import postcssCsso from "postcss-csso";
 import postcssLogical from "postcss-logical";
 import postcssDirPseudoClass from "postcss-dir-pseudo-class";
 
-const fs      = require("fs");
-const path    = require("path");
-const process = require("process");
```
distantnative, line 9: Isn't this a problem? Don't we need this?

panel/vite.config.js
```diff
@@ -1,29 +1,28 @@
+import fs from "fs";
```
johannschopplich, line 1: @distantnative It's here.

panel/vite.config.js
johannschopplich: Theoretically yes, for consistency, no, since we use ESM imports at the top of the file. Vite will transpile the ESM config to CJS via esbuild before it actually loads it. Thus, let's import it the ESM way just like all the other modules (which also get transpiled to `require` statements). I have moved the imports to the top of the file.

`process` is like `window` in the browser or `import.meta` in ES modules. Available globally, thus no importing necessary.

distantnative: 👍

PR #3750, opened 2021-10-04
Refactor Fiber internals to fix history

### LINKED ISSUE #3719: [3.6.0] broken forward button
The forward button is deactivated when navigating through the panel.

bastianallgeier:
While looking for the issue with the back button, I realised that we are still carrying around quite a lot of the Inertia concepts. I always found the way Inertia combines class methods to be very hard to understand. I tried to clean it up quite a bit more and also decouple it more from our app.js. 

I think it's a lot easier to understand now, but I'd love to hear your feedback. 

I removed the entire scroll storage. It wasn't working properly anyway and so far we don't need it. I think we should build our own implementation as soon as we need it. 

## Fixes

- #3719

### DIFF
--- a/panel/src/components/Forms/FormButtons.vue
+++ b/panel/src/components/Forms/FormButtons.vue
@@ -162,7 +162,11 @@ export default {
   },
   methods: {
     check() {
-      this.$reload({ only: "$view.props.lock", silent: true });
+      this.$reload({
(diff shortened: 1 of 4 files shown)


distantnative (approved):

distantnative:
I like!

Issue #3735, opened 2021-09-30
[3.6.0-beta.2] Strange behavior with pasting HTML to blocks


MaluNoPeleke:
**Describe the bug**  
I have discovered several strange behaviors when I tried to copy & paste content from HTML into blocks.

**To Reproduce bug 1**  
1. Go to https://loremipsum.de/
2. Copy the second block including the headline "Warum Lorem Ipsum?"
3. Paste it as a block
4. The headline is missing

**Expected behavior**  
Headline is included

**To Reproduce bug 2**  
1. Go to https://jsfiddle.net/m0681dr2/
2. Copy the content of the output window
3. Paste it as a block
4. Line break for the last sentence is missing
5. Copy only the paragraph part without the headline
6. Line break shows up

**Expected behavior**  
Line break shows up always

**Screenshots**  

**Kirby Version**  
Kirby 3.6.0-beta.2

**Console output**  
None

**Desktop (please complete the following information):**  
 - OS: macOS
 - Browser: Chrome
 - Version: 94 

**Additional context**

afbora:
I cannot reproduce the first issue, but I can see the second.

MaluNoPeleke:
I mean this block:
![image](https://user-images.githubusercontent.com/1446541/135484660-4bb5c0df-2039-4597-9011-2d3b08f6c900.png)
It never pastes the headline afterwards.

btw: have you seen my email regarding another strange issue @afbora ?

afbora:
Make sure `headings` block allowed.

![paste-2](https://user-images.githubusercontent.com/3393422/135485722-57847121-a30e-435e-96d4-795eab025405.gif)

afbora:
@MaluNoPeleke Your email has been marked as spam, I just got it. I'll reply after check.

afbora:
@bastianallgeier The break line works when we add a new mark to parsley but is this a correct solution?

https://github.com/getkirby/kirby/blob/develop/src/Parsley/Schema/Blocks.php#L72

MaluNoPeleke:
> Make sure `headings` block allowed.
> 
> ![paste-2](https://user-images.githubusercontent.com/3393422/135485722-57847121-a30e-435e-96d4-795eab025405.gif)

Headings block is allowed

afbora:
Can you share blueprint of blocks field?

MaluNoPeleke:
I just tested it again on a Windows laptop and there it works as well.

afbora:
[...]

Issue #3736, opened 2021-09-30
[3.6.0-beta.2] Stuck in an unsaved changes loop


MaluNoPeleke:
**Describe the bug**  
After updating to Kirby 3.6.0-beta.2 I am stuck in a loop of unsaved changes for articles, no matter how often I press on discard, save or close. 

**To Reproduce**  
1. Open any  (?) page in the panel
2. Get an unsaved changes notification
3. Click on Discard or Save
4. Go to 2.

**Expected behavior**  
If there is no change Kirby should not show unsaved changes.

**Screenshots**  
![screencast-dev peleke de-2021 09 30-20_23_18](https://user-images.githubusercontent.com/1446541/135510332-4c5bacf3-7e44-4339-887b-1bcc6698a79b.gif)

**Kirby Version**  
Kirby 3.6.0-beta.2

**Console output**  
None

**Desktop (please complete the following information):**  
 - OS: macOS / Windows
 - Browser: Chrome
 - Version: 94

**Additional context**

afbora:
I found the source of the problem. When we copy and paste, it transfers the links to the panel as they are. But when you focus on the input, the Kirby automatically adds the rel attribute to the links: `rel="noopener noreferrer nofollow"`.

Related line: https://github.com/getkirby/kirby/blob/develop/panel/src/components/Forms/Writer/Marks/Link.js#L96-L99

I'm not sure, but maybe we can solve it by defining the paste rule.

bastianallgeier:
I wonder if we should remove the rel attribute from the Mark in JS. We already have the referrer policy in the head so we don't need it for security reasons.

afbora:
I'm not very familiar with these topics, but after a bit of google I read in a few places that noopener and noreferrer provide extra security 🤷‍♂️

afbora:
https://github.com/getkirby/kirby/issues/3008 I think this is the same issue.

afbora:
Maybe @sebastiangreger can give us some insight on [this](https://github.com/getkirby/kirby/issues/3736#issuecomment-934308998)?

sebastiangreger:
[...]

PR #3751, opened 2021-10-05
Fix parsing logic for entire documents and breaks

bastianallgeier:
## Describe the PR
Parsley had some weird issues when HTML got pasted that does not have a wrapping body element. It also didn't handle inline breaks very well. This PR fixes both issues. 

## Fixes
- https://github.com/getkirby/kirby/issues/3735

### DIFF
--- a/src/Parsley/Schema/Blocks.php
+++ b/src/Parsley/Schema/Blocks.php
@@ -75,9 +75,6 @@ class Blocks extends Plain
             [
                 'tag' => 'a',
                 'attrs' => ['href', 'target', 'title'],
-                'defaults' => [
-                    'rel' => 'noopener noreferrer nofollow'
-                ]
             ],
             [
                 'tag' => 'abbr',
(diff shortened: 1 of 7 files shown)


src/Parsley/Schema/Blocks.php
```diff
@@ -75,13 +75,19 @@ public function marks(): array
             [
                 'tag' => 'a',
                 'attrs' => ['href', 'target', 'title'],
+                'defaults' => [
+                    'rel' => 'noopener noreferrer nofollow'
+                ]
```
afbora: Does this line have anything to do with fixing this issue? https://github.com/getkirby/kirby/issues/3736#issuecomment-931592854

bastianallgeier: Ah, damn. That was just a test. I wanted to remove it again. We need to solve it differently. Thanks for finding it.

afbora (approved):

PR #3755, opened 2021-10-05
Reload translation after login and installation

### LINKED ISSUE #3725: [3.6.0-alpha.1] Weird translation & language behaviors
language

I tested this issue with this PR, and the issue I'm having seems to be the subject of this PR as well.

**Steps to reproduce**

1. TR default, EN, secondary, EN user language
2. You'll see login screen as default language
3. Log-in panel
4. You will see some parts in Turkish and some parts in English
5. Logout
6. You will see login screen as English
7. When you refresh login screen, you'll see Turkish again

_Originally posted by @afbora in https://github.com/getkirby/kirby/issues/3713#issuecomment-927366963_

bastianallgeier:
## Describe the PR

The translation global needs to be updated after the installation and login. Both views will start with the default panel language which might be different from the actual user language. Only after the login or installation was successful, the user will be authenticated and the correct language will be loaded. 

### Fixes

- Fixed translation bug after login and installation [3725](https://github.com/getkirby/kirby/issues/3725)

## Breaking changes

None

## Related issues/ideas

- Fixes https://github.com/getkirby/kirby/issues/3725

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Forms/Login.vue
+++ b/panel/src/components/Forms/Login.vue
@@ -160,7 +160,10 @@ export default {
 
       try {
         await this.$api.auth.login(user);
-        this.$reload();
+
+        this.$reload({
+          globals: ["$system", "$translation"]
+        });
 
       } catch (error) {
         this.issue = error.message;
--- a/panel/src/components/Views/InstallationView.vue
+++ b/panel/src/components/Views/InstallationView.vue
@@ -143,9 +143,11 @@ export default {
     async install() {
       try {
         await this.$api.system.install(this.user);
-        this.$store.dispatch("notification/success", this.$t("welcome") + "!");
-        this.$go("/");
+        await this.$reload({
+          globals: ["$system", "$translation"]
+        });
 
+        this.$store.dispatch("notification/success", this.$t("welcome") + "!");
       } catch (error) {
         this.$store.dispatch("notification/error", error);
       }


afbora:
There is another small problem with login. When I log out, the page shows EN, but when I refresh the page, it switches back to the primary language (TR).

https://user-images.githubusercontent.com/3393422/136038992-846ae4ea-2420-4411-b9de-d5ec87690d56.mp4

bastianallgeier:
@afbora I think this is the correct behaviour though. 

When you logout, the Panel still knows your language because of the session. Once you are logged out, it no longer knows about you and will load the default language again. 

This could be only fixed when we keep the user language in an additional cookie somewhere, but I think we should not do that.

afbora:
But there is no `panel.language` setting in config and user's language is EN, why does login page show primary language as TR (for my setup)?

bastianallgeier:
Ok, now I'm lost. I have no idea to be honest.

bastianallgeier:
Oh, that's actually correct as well. You can check it out here: https://github.com/getkirby/kirby/blob/develop/src/Cms/AppTranslations.php#L100

The logic is basically like this: 

1. take the user language if the user is authenticated
2. otherwise take `panel.language` from the config if set
3. otherwise take the primary content language if it is a multi language installation
4. otherwise take English

afbora (approved):
Then the problem seems to be solved.

PR #3759, opened 2021-10-06
Canonical language URLs in the Panel

bastianallgeier:
## Describe the PR

With 3.6, the current content language is stored in the session. This works great, but also blocks the option to work on different languages in multiple tabs. With this PR, the language is now always appended to every Panel URL. 

It seems to work great in first tests, but could come with unexpected issues. 

@nilshoerrmann you seem to be the right person to test, if this implementation works as you expect it :)


nilshoerrmann:
> @nilshoerrmann you seem to be the right person to test, if this implementation works as you expect it :)

Yes, I'll test this over the next days :)

nilshoerrmann:
So, we've got two languages: German and English, with German being the default one.

- If I open the panel in two tabs and switch languages to English in the second one, the first will display German content and the second will display English content. ✅
- If I now reload the first one, content switches two English. My guess is because first tab doesn't have the language set in the URL. ❎ (I think this behaviour is unexpected but this happened in previous Kirby version as well)
- If I switch back to German in the first tab, both display correctly again, even on reload (because there is the URL parameter). ✅
- As soon as I click on another page, content will switch to German in both cases because the language is not promoted to links within sections. ❌ (this did not occur previously, Kirby sticked to the previously displayed language)

What I'd have expected is that a tab or window sticks to the language it is currently displaying unless I actively switch the language. It feels a bit strange that languages are synchronized across instances, especially because you might not realise this when editing a secondary language because Kirby copies the primary language on first edit and thus both languages will look the same. (It often happend to us before that we started editing the wrong language because of this.)

As we have different slugs for the different languages, wouldn't it make sense to use this in the panel as well – like `/panel/pages/projekte+mein-projekt` and `/panel/pages/en/projects+my-project` – and derive the current display language from the URL?

nilshoerrmann:
> As we have different slugs for the different languages, wouldn't it make sense to use this in the panel as well – like /panel/pages/projekte+mein-projekt and /panel/pages/en/projects+my-project – and derive the current display language from the URL?

To clarify this comment: I'd suggest something along these lines for a future release that focuses on languages. Working with languages and translations can be very painful in Kirby so I personally hope for a more general update in this area.

For Kirby 3.6, fixing promotion of the current language to the inline links should do the trick.

lukasbestle:
I think the new frontend implementation is somehow fighting with the `language` session variable we set on the backend. I literally have no idea how it all fits together, but my feeling is we need to get rid of that session variable and handle everything with URLs.

nilshoerrmann:
Handling everything in the URL has the advantages that 1) it's very clear where you are and 2) usually front- and backend will behave the same :+1: 

Currently, the issue is that you'll never know in which language you will land when navigating the panel. This is of course only true if you open your site in multiple windows or tabs but this is a common thing for people to do, I guess.

nilshoerrmann:
Am I correct in assuming that with this change `this.$store.state.languages.default` and `this.$store.state.languages.current` will be removed? What's the recommended alternative to get the current language in a panel component?

afbora:
@nilshoerrmann New syntax on 3.6 and [here document](https://getkirby.com/releases/3.6/breaking-changes#panel__helpers-libraries)

````js
const currentLanguage = this.$language;
const defaultLanguage = this.$languages.filter(language => language.default === true);
````

**Comment by nilshoerrmann** [comment]
Great, thanks!

**Comment by bastianallgeier** [comment]
@nilshoerrmann I cannot really reproduce your issue with links in sections. Is it a particular section type? Maybe a plugin? In A single method is responsible for routing any internal link and it is taking care of creating the correct query, so it is almost impossible that any internal link would end up without it. 

@lukasbestle To be honest, I also don't think that the session is causing additional problems. It's a good fallback whenever a panel view is loaded without query. When we really make sure that all internal links get created correctly, the session should not interfere with that and just add some additional comfort.

**Comment by nilshoerrmann** [comment]
These are two example sections from our site:

```yaml
        sections:
          projects:
            type: pages
            headline: Projekte
            parent: kirby.page("projekte")
            limit: 30
            layout: cards
            status: published
            info: false
            image:
              query: page.previewImage(page.basecolor)
              cover: true
          texts:
            type: pages
            headline: Texte
            parent: kirby.page("texte")
            status: published
            info: '{{ page.date.toDate("%m.%d.%Y") }}'
            image:
              query: page.previewImage(page.basecolor)
              cover: true
```

I'm going to restart and pull all changes from this PR again to make sure everything is really up-to-date.

nilshoerrmann:
I explicitely set the language by choosing German from the menu and now have the following URL:
`http://localhost:8888/hananils/panel/site?language=de`

But all linked pages are still missing the language, e. g.:
`http://localhost:8888/hananils/panel/pages/projekte+cahiers-de-mayence`

The last commit I have is 677c68aa2afc22cb6feda58f1fb52c6a579f9e31. Is there anything that might interfer with the link generation? We have custom page models but none of them touches page URLs.

nilshoerrmann:
Okay, I think I got it working: something was wrong with my panel assets. Rebuilding them again finally did the trick and I now get the language query appended to all links :+1:

Sorry for the inconvenience, I'm not sure what happend there in the first place.

bastianallgeier:
@nilshoerrmann sorry, it's my fault. I didn't add the latest dist files. You are probably still testing beta.2 🙈

nilshoerrmann:
All good, it's working now :) 
Looking at the functionality, switching languages now works like in previous versions.

bastianallgeier:
There's a major difference now. There was no way to link to a particular language in the previous versions. So it's at least a bit of an improvement :)

PR #3760, opened 2021-10-06
Select blocks in the correct order

bastianallgeier:
Blocks are now always selected in the original order, no matter in which order they have been selected.

### DIFF
--- a/panel/src/components/Forms/Blocks/Blocks.vue
+++ b/panel/src/components/Forms/Blocks/Blocks.vue
@@ -295,9 +295,8 @@ export default {
 
       let blocks = [];
 
-      this.selectedOrBatched.forEach(id => {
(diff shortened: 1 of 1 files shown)


panel/src/components/Forms/Blocks/Blocks.vue
```diff
@@ -295,9 +295,8 @@ export default {
 
       let blocks = [];
 
-      this.selectedOrBatched.forEach(id => {
-        const block = this.find(id);
-        if (block) {
+      this.blocks.forEach(block => {
+        if (this.selectedOrBatched.includes(block.id)) {
```
afbora, line 299: Nice trick 😉

afbora (approved):

PR #3761, opened 2021-10-06
Fix empty box layout for cardlets

bastianallgeier:
The styling of the empty box was broken when the cardlets layout was selected.


afbora:
I'm not sure, but it seems to me that the empty `cardlets` design should be like `cards` not `list` 🙈

bastianallgeier:
@afbora you are totally right. Should be fixed.

afbora (approved):

PR #3769, opened 2021-10-07
Accept emails and phone numbers in the Dom sanitizer

### LINKED ISSUE #3754: Writer field: no href attribute rendered when link does not contain protocol
**Describe the bug**  
The `writer` field occasionally does not render an `href` attribute in its links. I believe this happens when the entered link does not contain a protocol. Unfortunately, no warning is given and it seems like this also translates to the new `email` button.

**To Reproduce**  
1. Setup a `writer` field in any blueprint.
2. Add a link without a protocol* OR use the new `email` button.
3. Render the link in a template.
4. The link will not contain an `href` attribute.

\* It seems that adding `http://`works but the protocol is removed in the `href` attribute. `https://` works AND will be included in the `href`.

**Expected behavior**  
1. The writer field should either not let the user submit a link without protocol (like the `url` field) or render the `href` attribute as it was entered.
2. The `href` attribute should be rendered when using the `email` button.

**Kirby Version**  
3.6.0-beta.2

bastianallgeier:
## Describe the PR

The new DOM sanitizer class killed all email address and telephone numbers. I've added additional checks for such addresses and wrote tests. 

@lukasbestle could you have a look if this is how you would have done it? 

## Related issues

- Fixes: https://github.com/getkirby/kirby/issues/3754

### DIFF
--- a/src/Toolkit/Dom.php
+++ b/src/Toolkit/Dom.php
@@ -295,6 +295,28 @@ class Dom
             return 'Invalid data URI';
         }
 
+        // allow valid email addresses
+        if (Str::startsWith($url, 'mailto:') === true) {
+            $address = Str::after($url, 'mailto:');
+
+            if (empty($address) || V::email($address) === true) {
+                return true;
+            }
+
+            return 'Invalid email address';
+        }
+
+        // allow valid telephone numbers
+        if (Str::startsWith($url, 'tel:') === true) {
+            $address = Str::after($url, 'tel:');
+
+            if (empty($address) || preg_match('!^(\+|)[0-9]+$!', $address)) {
+                return true;
+            }
+
+            return 'Invalid telephone number';
+        }
+
         return 'Unknown URL type';
     }
 
--- a/src/Toolkit/Dom.php
+++ b/src/Toolkit/Dom.php
@@ -310,7 +310,7 @@ class Dom
         if (Str::startsWith($url, 'tel:') === true) {
             $address = Str::after($url, 'tel:');
 
-            if (empty($address) || preg_match('!^(\+|)[0-9]+$!', $address)) {
+            if (empty($address) || preg_match('!^[+]?[0-9]+$!', $address)) {
                 return true;
             }
 
--- a/tests/Toolkit/DomTest.php
+++ b/tests/Toolkit/DomTest.php
@@ -60,6 +60,11 @@ class DomTest extends TestCase
             ['data:image/jpeg;base64,test', 'Invalid data URI', [
                 'allowedDataUris' => []
             ]],
+            
+            // forbidden data uri
+            ['data:image/png;base64,test', 'Invalid data URI', [
+                'allowedDataUris' => ['data:image/jpeg;base64']
+            ]],
 
             // forbidden URL when no domains are accepted
             ['https://getkirby.com', 'The hostname "getkirby.com" is not allowed', [
@@ -87,6 +92,9 @@ class DomTest extends TestCase
 
             // forbidden URL type
             ['ftp:test', 'Unknown URL type'],
+            
+            // forbidden URL type
+            ['ftp://test', 'Unknown URL type'],
         ];
     }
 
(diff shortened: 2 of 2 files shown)


afbora:
Unimportant but not to be overlooked, I just wonder following urls:

- **FTP**: ftp://example.com
- **Relatives**:
  - ./any-page 
  - ./subfolder
  - sibling-page

afbora (dismissed):

src/Toolkit/Dom.php
```diff
@@ -295,6 +295,28 @@ public static function isAllowedUrl(string $url, array $options)
...
+            }
+
+            return 'Invalid email address';
+        }
+
+        // allow valid telephone numbers
+        if (Str::startsWith($url, 'tel:') === true) {
+            $address = Str::after($url, 'tel:');
+
+            if (empty($address) || preg_match('!^(\+|)[0-9]+$!', $address)) {
```
lukasbestle: I believe this can be written like this for better readability.

```suggestion
            if (empty($address) || preg_match('!^[+]?[0-9]+$!', $address)) {
```

In theory `\+?` should also be possible, but it's also hard to read.

tests/Toolkit/DomTest.php
```diff
@@ -0,0 +1,102 @@
...
+            // allowed empty phone number
+            ['tel:', true],
+
+            // allowed phone number
+            ['tel:+491122334455', true],
+
+            // forbidden data uri
+            ['data:image/jpeg;base64,test', 'Invalid data URI', [
+                'allowedDataUris' => []
+            ]],
```
lukasbestle, line 62: Another test case:

```suggestion
            // forbidden data uri
            ['data:image/jpeg;base64,test', 'Invalid data URI', [
                'allowedDataUris' => []
            ]],
            
            // forbidden data uri
            ['data:image/png;base64,test', 'Invalid data URI', [
                'allowedDataUris' => ['data:image/jpeg;base64']
            ]],
```

tests/Toolkit/DomTest.php
```diff
@@ -0,0 +1,102 @@
...
+            ['tel:test', 'Invalid telephone number'],
+
+            // forbidden phone numbers - too much formatting
+            ['tel:+49 (0) 1234 5678', 'Invalid telephone number'],
+
+            // forbidden phone numbers - invalid plus sign position
+            ['tel:491234+5678', 'Invalid telephone number'],
+
+            // forbidden URL type
+            ['ftp:test', 'Unknown URL type'],
```
lukasbestle, line 94: And also another test case:

```suggestion
            // forbidden URL type
            ['ftp:test', 'Unknown URL type'],
            
            // forbidden URL type
            ['ftp://test', 'Unknown URL type'],
```

tests/Toolkit/DomTest.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * @dataProvider urlProvider
+     * @covers ::isAllowedUrl
+     */
+    public function testIsAllowedUrl(string $url, $expected, array $options = [])
+    {
+        $dom = new Dom('Test');
+        $this->assertSame($expected, $dom->isAllowedUrl($url, $options));
```
lukasbestle: The method is static, so this also works:

```suggestion
        $this->assertSame($expected, Dom::isAllowedUrl($url, $options));
```

lukasbestle (commented):
👍❤️

Or as [Jeremy Keith has said](https://soundcloud.com/baldower/this-is-what-the-web-is-all) it:

> It's awesome. I mean I just absolutely love what you've done.

lukasbestle:
@afbora Regarding FTP I feel like we shouldn't support these URLs by default. There are basically no actual use cases for FTP anymore and if anyone has one, they can let us know and we can still add it. The relative URLs could be supported, but it could be risky as the syntax of these is more complex. E.g. you would need to check that the first part doesn't resemble a protocol in the end.

afbora:
Yes, it seems complicated to implement the relative path. We can take care if there is a need in the future 👍

PR #3770, opened 2021-10-07
Use Dom class in Parsley and improve parsing

bastianallgeier:
Improves parsing Word HTML quite a bit and also reuses the new Dom class instead of reinventing the wheel. 

This PR also bumps the test coverage for the Parsley package to 100% 🎉

We should make sure to squash all commits 🙈

### DIFF
--- a/src/Parsley/Parsley.php
+++ b/src/Parsley/Parsley.php
@@ -2,25 +2,77 @@
 
 namespace Kirby\Parsley;
 
-use DOMDocument;
-use DOMXPath;
+use DOMNode;
 use Kirby\Parsley\Schema\Plain;
+use Kirby\Toolkit\Dom;
 
+/**
+ * HTML parser to extract the best possible blocks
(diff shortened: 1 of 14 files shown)


src/Parsley/Parsley.php
```diff
@@ -216,19 +322,19 @@ public function parseNode($element)
...
+     * @return bool
+     */
     public function useXmlExtension(): bool
     {
         if (static::$useXmlExtension !== true) {
             return false;
         }
 
-        return class_exists('DOMDocument') === true;
+        return $this->dom->isSupported();
```
lukasbestle: Hm, how does this check work? If there's already a `Dom` instance, then the `DOMDocument` class was already used in `Dom::__construct()`.

`Dom::isSupported()` should IMO be a static method and used before `new Dom()`.

bastianallgeier: You are completely right. I just fixed it.

PR #3775, opened 2021-10-08
Implement system.exception hook

thathoff:
## Describe the PR
This is an idea how to implement a new hook for `system.exception`. It will register an additional Whoops callback handler to trigger the hook which makes it possible to implement plugins for custom error monitoring (eg. Sentry).

## Release notes
Feature: add hook for system.exception

## Ready?
- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Cms/AppErrors.php
+++ b/src/Cms/AppErrors.php
@@ -158,7 +158,21 @@ trait AppErrors
         $whoops = $this->whoops();
         $whoops->clearHandlers();
         $whoops->pushHandler($handler);
+        $whoops->pushHandler($this->getExceptionHookWhoopsHandler());
         $whoops->register(); // will only do something if not already registered
+    }
+
+    /**
+     * Initializes a callback handler for triggering the `system.exception` hook
+     *
+     * @return \Whoops\Handler\CallbackHandler
+     */
+    protected function getExceptionHookWhoopsHandler(): CallbackHandler
+    {
+        return new CallbackHandler(function ($exception, $inspector, $run) {
+            $this->trigger('system.exception', compact('exception'));
+            return Handler::DONE;
+        });
     }
 
     /**
--- a/tests/Cms/App/AppErrorsTest.php
+++ b/tests/Cms/App/AppErrorsTest.php
@@ -40,8 +40,9 @@ class AppErrorsTest extends TestCase
 
         $testMethod->invoke($app);
         $handlers = $whoops->getHandlers();
-        $this->assertCount(1, $handlers);
+        $this->assertCount(2, $handlers);
         $this->assertInstanceOf('Whoops\Handler\PlainTextHandler', $handlers[0]);
+        $this->assertInstanceOf('Whoops\Handler\CallbackHandler', $handlers[1]);
     }
 
     /**
@@ -66,8 +67,9 @@ class AppErrorsTest extends TestCase
 
         $testMethod->invoke($app);
         $handlers = $whoops->getHandlers();
-        $this->assertCount(1, $handlers);
+        $this->assertCount(2, $handlers);
         $this->assertInstanceOf('Whoops\Handler\PlainTextHandler', $handlers[0]);
+        $this->assertInstanceOf('Whoops\Handler\CallbackHandler', $handlers[1]);
 
         // JSON
         Server::$cli = false;
@@ -75,8 +77,9 @@ class AppErrorsTest extends TestCase
 
         $testMethod->invoke($app);
         $handlers = $whoops->getHandlers();
-        $this->assertCount(1, $handlers);
+        $this->assertCount(2, $handlers);
         $this->assertInstanceOf('Whoops\Handler\CallbackHandler', $handlers[0]);
+        $this->assertInstanceOf('Whoops\Handler\CallbackHandler', $handlers[1]);
 
         // HTML
         Server::$cli = false;
@@ -93,8 +96,9 @@ class AppErrorsTest extends TestCase
 
         $testMethod->invoke($app);
         $handlers = $whoops->getHandlers();
-        $this->assertCount(1, $handlers);
+        $this->assertCount(2, $handlers);
         $this->assertInstanceOf('Whoops\Handler\PrettyPageHandler', $handlers[0]);
+        $this->assertInstanceOf('Whoops\Handler\CallbackHandler', $handlers[1]);
 
         // reset global state
         Server::$cli            = $oldCli;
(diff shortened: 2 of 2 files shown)


lukasbestle:
I like the idea in general, but I wonder if it should rather be a plugin extension, i.e. a full Whoops handler that can also control the return value of the callback. Basically each plugin would be able to register a handler and all configured handlers would be pushed to the handler list in Whoops.

Whether a hook or a plugin extension makes sense depends on the use case of course. I think we should go with the most versatile implementation.

@thathoff What do you think about the plugin extension idea? Does anything speak against that for your use case?

thathoff:
Thanks for your feedback. We wanted to rework my [old Sentry plugin](https://github.com/thathoff/kirby-sentry) for Kirby 3. Since it’s not possible to catch exceptions with the current hooks we came up with this idea.

Having a real Whoops handler interface would be the better solution, right. We actually thought about this too but we thought this would require a bit more rework in the AppError trait, so we used that simpler solution to discuss the idea.

So from the plugin interface it would look like this?

```php
use Whoops\Handler\Handler;

Kirby::plugin('example/myplugin', [
    'whoopsHandler' => function ($exception, $inspector, $run) {
        // do something with the exception

        return Handler::DONE;
    }
]);
```

Am I right that the approach would be the following: 
- decide a name for the new extension type (proposal: whoopsHandler)
- extend AppPlugins trait to include the new extension type
- rework the AppErrors trait to register all handlers of all plugins

I’m wondering if we should call it `whoopsHandlers` (plural) in the plugin config which is more inline with the other plugin extensions but might imply that you can set multiple handlers (which would be possible to implement) or if we stay with whoopsHandler and only allow to set one Whoops Handler per plugin. What do you think?

lukasbestle:
Good question. I think given that multiple plugins can register handlers, it should definitely be plural, but then each plugin must also be able to register multiple handlers. But actually: Why not? Might be useful for more complex plugins. :)

Another thing comes to my mind: If we call it `whoopsHandlers` and also use the function signature of Whoops, we wouldn't be able to switch to a different library later. So maybe we should actually go with the simpler hook implementation as it's a lot more portable. I'm not sure...

If we go for the hook implementation: We would need to modify it slightly as it currently only registers the custom hook in debug mode. But especially error reporting plugins may want to know about exceptions in production as well (or even primarily there).

thathoff:
> Good question. I think given that multiple plugins can register handlers, it should definitely be plural, but then each plugin must also be able to register multiple handlers. But actually: Why not? Might be useful for more complex plugins. :)

I would say: If we go with a hook it’s a single function but if we go with a new extension type it should ne plural and can be an array or a single callback.

> Another thing comes to my mind: If we call it whoopsHandlers and also use the function signature of Whoops, we wouldn't be able to switch to a different library later. So maybe we should actually go with the simpler hook implementation as it's a lot more portable. I'm not sure...

Yeah, that’s a really good point. I would vote for not bind this to Whoops. In this case it also does not make sense to support custom return values. Or we need to build a custom KirbyErrorHandler class to make it replaceable afterwards. I think the hook seems a lot simpler then.

To think in another way: From my point of view there are two use cases which are possible with this:

1) Catch exceptions and log them to some service/log/whatever
This is possible with both the hook and the custom Whoops handler implementation.

2) Display a custom error page
This would be possible by using `exit/die` inside the callback but you might kill stop execution of all other exception handlers afterwards. But that would also happen if return Handler::QUIT from a custom Whoops handler. So there is no real advantage here i think.

Option 2 is also possible with the `fatal` config option so you can even combine the custom log with a custom error page.

I think we should go with the hook or a non-Whoops specific callback implementation. What do you think?

> We would need to modify it slightly as it currently only registers the custom hook in debug mode. But especially error reporting plugins may want to know about exceptions in production as well (or even primarily there).

Actually that was the main reason for this pull request because it was not possible to register error handlers when debug was disabled. We’ve tested this with a little dummy implementation and the hook is called no matter if whoops is disabled or debug is disabled. At least I think we did… ;-)

lukasbestle:
After thinking about it again, I feel pretty confident we should go with the hook. Sorry for the detour. 😅

Regarding debug mode you are right, I misread the code before. The only edge case is when `debug` is `true`, but `whoops` is `false`. But in this case the question is whether plugin hooks should actually be called given that Whoops is explicitly disabled.

Apart from that we only need updated unit tests. It's hard or even impossible to test the actual behavior, but at least the existing tests should be adapted so that they pass. :)

thathoff:
Haha, that's totally fine. I've expected a bit discussion about this so I did not checked the unit tests before... 😇

I'm also pretty sure the hook is also triggered when `whoops` is set to `false`, the only change the whoops option does is to enable the nice debug view when debug is already enabled. Or do I get the code in `AppError::handleHtmlErrors()` wrong?

I'll try to update the unit tests in the next days and recheck the thing with `whoops` set to `false`.

bastianallgeier:
Good discussion :) I totally agree with the result of it. I also don't think that the whoops option is an issue. It only affects the html handler.

thathoff:
@lukasbestle you were right, when `debug` is `true` and `whoops` is `false` the hook is not called. This seems to be quite hard to implement this another way without reworking the whole error handling and Whoops initialisation or do you have an elegant solution to this?

I think the case that someone has `debug` set to `true` and `whoops` to `false` would be a quite rare case so I vote to ignore this and explicitly mention this in the docs.

thathoff:
Just added the missing unit tests. Any ideas why the PHP tests are failing with `Allowed memory size exhausted`?

src/Cms/AppErrors.php
```diff
@@ -158,9 +158,27 @@ protected function setWhoopsHandler($handler): void
...
         $whoops->clearHandlers();
         $whoops->pushHandler($handler);
+        $whoops->pushHandler($this->getExceptionHookWhoopsHandler());
         $whoops->register(); // will only do something if not already registered
     }
 
+    /**
+     * Initialize a callback handler for triggering the system.exception hook
+     *
+     * @return CallbackHandler
```
lukasbestle: Minor consistency fixes:

```suggestion
     * Initializes a callback handler for triggering the `system.exception` hook
     *
     * @return \Whoops\Handler\CallbackHandler
```

src/Cms/AppErrors.php
```diff
@@ -158,9 +158,27 @@ protected function setWhoopsHandler($handler): void
...
+     */
+    protected function getExceptionHookWhoopsHandler(): CallbackHandler
+    {
+        return new CallbackHandler(function ($exception, $inspector, $run) {
+            $this->trigger('system.exception', [
+                'exception' => $exception,
+                'kirby' => $this
+            ]);
```
lukasbestle: The `$kirby` argument is not needed as the Kirby object is bound to `$this` inside the hook callback.

```suggestion
            $this->trigger('system.exception', compact('exception'));
```

lukasbestle (commented):
The memory limit error is very weird indeed. According to the error message, the memory limit is 512 MB, but we have configured it to be 2048 MB in our PHPUnit config. I have no idea why it occurs after these simple changes...

I have restarted CI, but it's still the same. I may need to look into this more closely.

Apart from that, two minor comments:

thathoff:
Just updated the commit accordingly. The memory limit seems somehow related to code coverage, I think. Quite weird…

bastianallgeier:
Sorry for the delay, but we want to pick this up again.

bastianallgeier (commented):
The implementation looks good to me. If we somehow get rid of the memory issue, we can merge this. Could you try to rebase it first?

distantnative:
Rebased version https://github.com/getkirby/kirby/pull/3952

PR #3790, opened 2021-10-11
Proposal: Use Prettier to format Panel source files

johannschopplich:
## Describe the PR

The formatting of the Panel sources varies a bit from file to file. One example: double quotes are used most of the time, but sometimes single quotes sneak in there. Thus, I propose Prettier with the most commonly used formatting in the Panel. 

Configuration:
- Tab width of 2 spaces
- Semicolons
- Double quotes
- No trailing commas

Further notes:

- Prettier plays nicely with ESLint, when configured correctly. 
- VSCode default rules for a standardized development setup were added as well.

I have created snackable commits for you to easily grasp how Prettier was integrated.

## Release notes

None.

## Breaking changes

None.

## Related issues/ideas

None.

## Ready?

- [ ] CI checks pass

## When merging


panel/package.json
```diff
@@ -7,7 +7,9 @@
     "cy:run": "cypress run",
     "cy:open": "cypress open",
     "dev": "vite",
-    "lint": "eslint --ext .js,.vue src",
+    "lint": "eslint \"src/**/*.{js,vue}\"",
+    "lint:fix": "npm run lint -- --fix",
+    "format": "prettier --write \"src/**/*.{js,vue}\"",
```
lukasbestle, line 12: Doesn't the glob exclude `panel/src/index.js`? Not a big deal, but it may be confusing if there's ever an issue in this file and we don't remember that this file is neither linted nor formatted.

.vscode/extensions.json
```diff
@@ -0,0 +1,6 @@
+{
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
+    "esbenp.prettier-vscode"
+  ]
+}
```
lukasbestle, line 6: While I think that the recommendations are useful, I feel like we shouldn't remove the `.vscode` dir from `.gitignore`. We have ignored it on purpose. To be honest I don't fully remember why though. :)

lukasbestle (commented):
Thank you, I think it's a really useful addition. I have two comments below.

Note to myself: Before merging, we also need to verify the automated changes. E.g. I have seen some changes to whitespace inside tags. We need to check if that breaks styling etc.

Also we need to add prettier to the CI workflow and to `scripts/pre-commit`. Then we know for sure that the formatting happens, no matter how the local editor is configured.

panel/.eslintrc.js
```diff
@@ -2,29 +2,20 @@ module.exports = {
...
-    "plugin:vue/recommended"
+    "plugin:vue/recommended",
+    "prettier"
   ],
   rules: {
     "vue/component-definition-name-casing": "off",
     "vue/require-default-prop": "off",
     "vue/attributes-order": "error",
     "vue/require-prop-types": "error",
-    "vue/max-attributes-per-line": [
```
distantnative, line 12: Why removing this one?

panel/package.json
johannschopplich: The globbed files include `panel/src/index.js` as well, since `src/**` includes `src/` itself as well.

.vscode/extensions.json
johannschopplich: It was for a long time and still is common practice to not track editor config files. However, VSCode provides useful additions like extension recommendations etc., I see more and more repos tracking the `.vscode` folder for commonly used, project-specific settings which every developer should adhere to. Prominently, the Vue repo itself: https://github.com/vuejs/vue-next/

Just an idea. Open for debate. 🙂

panel/.eslintrc.js
johannschopplich: Because this rule unfortunately conflicts with Prettier. I have looked up two issues, both resolved, but unable to provide a fix or workaround. If we use Prettier, this rule won't work.

Prettier uses 4 max attributes and _also_ takes line length into consideration, which results in a much more readable attributes formatting than just sticking to a hard-coded number. The deleted rule defines 3 max attributes. I think it's a plausible new approach.

distantnative: As long as it doesn't lead to one long line with all attributes, I'm fine. If Prettier has another sensible default, I'm fine with it :)

bastianallgeier:
I'd love to get this done, but I wonder if it's a good idea to do this during the beta or we should wait until after 3.6? I'm open for discussions. Maybe the beta is the perfect timing for it. I'm just a bit hesitant.

johannschopplich:
@bastianallgeier I understand your hesitation, the overall changes, albeit small, are plenty. Strictly speaking, the code logic shouldn't be affected, since only formatting is done. Prettier is well tested, so I expect full functionality just like before. I have reviewed whitespace-related changes and fixed leftovers from _pre-formatted_ code to _formatted_ code.

I can only see one potentially risky spot, which is negligible, since the Vue compiler will preserve white spaces (at least for Vue 2, from Vue 3 onwards they will be collapsed). Example formatted code:

```diff
- <template v-if="isLoading">
-   …
- </template>
+ <template v-if="isLoading">…</template>
```

Since the `<template>` is on a new line, the string before will be appended by a trailing white space. Resulting in `Demo message …` like before formatting (note the space before the `…`).

Scrolling through the formatted code, I couldn't find any further potential risks. Hopefully there are none. Maybe @distantnative can take a look at the formatted code as well.

Because Kirby 3.6 is a big milestone in refactoring and also heavily tested during beta, I think formatting the Panel files now is most appropriate.

johannschopplich:
Note: Whitespace-related data binding isn't changed by Prettier. `<div>{{ $t('some.message') }}</div>` will stay exactly the same. I'm confident the formatting doesn't introduce erroneous spaces, nor deletes needed ones.

distantnative:
Something is off with the PR, it somehow includes changes to the CI pinned versions and also a change in a PHP file - seems unintended.

johannschopplich:
Dang. Probably since my rebase yesterday. I will try to fix it. Don't know how to yet tho... 

@distantnative Could you help me with this? I don't know how to clean the history and properly rebase on develop.

distantnative:
I think I got it clean over there: https://github.com/getkirby/kirby/pull/3812

bastianallgeier:
Let's move on over here: https://github.com/getkirby/kirby/pull/3812

PR #3792, opened 2021-10-11
Add option 'languages.fallback' for autodetection

mrflix:
## Describe the PR
You might want another language as default for editing in the panel than the one used for auto-detection on the website. This pull request introduces an option to set the fallback language that's being used for auto-detecting the preferred language of the visitor so that you're free to choose which language should be the default for your editors.

## Release notes
New option 'languages.fallback' if you want the language auto-detection to fallback to another language than your default language.

## Breaking changes
None

## Related issues/ideas
- https://kirby.nolt.io/253
- https://github.com/getkirby/ideas/issues/143

## Ready?
- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -562,7 +562,11 @@ class App
             }
         }
 
-        return $this->defaultLanguage();
+        if ($fallbackLanguage = $languages->findBy('code', $this->option('languages.fallback'))) {
+            return $fallbackLanguage;
+        } else {
+            return $this->defaultLanguage();
+        }
     }
 
     /**


afbora (commented):
How about new `fallbackLanguage()` method?

````php
/**
 * Detect the preferred language from the visitor object
 *
 * @return \Kirby\Cms\Language
 */
public function detectedLanguage()
{
    ...

    return $this->fallbackLanguage() ?? $this->defaultLanguage();
}

/**
 * Returns the fallback language object
 *
 * @return \Kirby\Cms\Language|null
 */
public function fallbackLanguage()
{
    if ($fallbackLanguage = $this->option('languages.fallback')) {
        return $this->languages()->findBy('code', $fallbackLanguage);
    }

    return null;
}
````

mrflix:
Looks great too!

I'm questioning right now if the problem is a real-world problem or just a theoretical one. In discussion with @medienbaecker, he brought up that his browsers accepted languages are "de_de", "de" and "en". That made me think that "en" might be set as the default fallback language in all browsers. I checked in one of mine who's config I never adjusted and there was also "en" as the last fallback option behind my OS language.

I only ever us the fallback option to set the fallback language to English if the default language is something else.

So the fallback option might not be needed if thats already solved one layer above – in the accept language header.

distantnative:
I'm wondering whether this is a bit too over engineered/specific. Or also causes more problems later on.
IMO a site should have one default language - and optional translations.

I could see problems arising when we start trying to differentiate between one language being the default when using the backend, another language though when using the frontend.

mrflix:
I'm totally fine with closing this pull request. For a long time I set English as the primary language because I feared that German as the primary language would lead to non english-speaking visitors getting served German content. But it turns out that browsers have English set as their fallback language and that gets picked up by Kirby.

distantnative:
As agreed, closing this for now

PR #3797, opened 2021-10-12
Add “crop” to the filename

### LINKED ISSUE #3418: Cropping an image fails, if there is already a resized image with the crop dimensions
**Describe the bug**  
I was just trying to resize and to crop some images and ran into this bug:

Let's say I have an image of these dimensions: `529 × 705`
Now, if I first resize this image with something like: `$img->resize(512, 288)` then it will create a resized image with the dimensions `216 × 288` and save it in the media folder with a name like `myimg-512x288.png`.
Now, if I next try to crop the image with the same numbers, `$img->crop(512, 288)` it does *not* create a new, cropped image as Kirby seems to assume that the image already exists and it takes the previously resized image (which is wrong).

So my guess is that the names collide as the media folder does not create different names for cropped or resized images. Now that I investigated this issue, I am fairly certain that I have run into this issue in the past already and thought the crop command is broken somehow.

**Expected behavior**  
Crop the image as expected, even if there already is a resized image with the same arguments.

**Kirby Version**  
3.5.3

 - OS: macOS 10.13.6
 - Browser: Chrome
 - Version: 91.0.4472.77

bastianallgeier:
## Describe the PR

There was a potential filename collision when dimensions of cropped and resized images matched. This PR fixes the issue by adding the "crop" keyword to the filename, even if the image is only cropped from the center. 

### Fixes

- Fixed filename collision for cropped image files #3418 

## Breaking changes

- The crop keyword is now always present in filenames for cropped files. All center cropped files need to be regenerated. 

## Related issues/ideas

- Fixes #3418 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Filesystem/Filename.php
+++ b/src/Filesystem/Filename.php
@@ -139,7 +139,7 @@ class Filename
                     $result[] = $value;
                     break;
                 case 'crop':
-                    $result[] = ($value === 'center') ? null : $key . '-' . $value;
+                    $result[] = ($value === 'center') ? 'crop' : $key . '-' . $value;
                     break;
                 default:
                     $result[] = $key . $value;
--- a/tests/Filesystem/FilenameTest.php
+++ b/tests/Filesystem/FilenameTest.php
@@ -62,7 +62,7 @@ class FilenameTest extends TestCase
                 ]
             ],
             [
-                '',
+                '-crop',
                 [
                     'crop' => 'center',
                 ]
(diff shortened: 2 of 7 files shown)


bastianallgeier:
@distantnative @afbora @lukasbestle do you have a chance to take a look at this?

afbora (approved):
Sorry for delay. Works great 👍

PR #3811, opened 2021-10-15
Set 5 minutes timeout for GitHub actions

afbora:
## Describe the PR

The default timeout is 6 hours when an issue occurs in GitHub actions. In this PR, the time is set to 5 minutes.

## Release notes

None

## Breaking changes

None


lukasbestle (commented):
What about the other jobs for analysis and coding style? I think we could add the timeout there as well.

lukasbestle (approved):

PR #3812, opened 2021-10-17
Use Prettier for consistent JS formatting

distantnative:


johannschopplich:
You are a hero. Thank you!

panel/src/api/pages.js
```diff
@@ -69,22 +69,31 @@ export default (api) => {
...
+        query
+      );
     },
     async update(id, data) {
       return api.patch("pages/" + this.id(id), data);
     },
     url(id, path) {
-      let url = id === null ? "pages" : "pages/" + String(id).replace(/\//g, "+");
+      let url =
+        id === null ? "pages" : "pages/" + String(id).replace(/\//g, "+");
```
distantnative, line 90: To be honest, this syntax (line break) looks very weird to me. But I guess we would have to accept such side effects?

panel/src/components/Dialogs/Dialog.vue
```diff
@@ -294,7 +297,7 @@ export default {
 }
 
 .k-dialog-notification {
-  padding: .75rem 1.5rem;
+  padding: 0.75rem 1.5rem;
```
distantnative, line 300: Leading zeros mandatory?

panel/src/components/Dialogs/UsersDialog.vue
```diff
@@ -22,17 +22,9 @@
       />
 
       <template v-if="items.length">
-        <k-items
-          :items="items"
-          layout="list"
-          :sortable="false"
-          @item="toggle"
-        >
+        <k-items :items="items" layout="list" :sortable="false" @item="toggle">
```
distantnative, line 25: I thought 4 arguments would be broken down into separate lines? To be honest, moving this into one line that needs to wrap makes me doubt the usefulness of Prettier.

panel/src/api/pages.js
johannschopplich: I feel you. Some changes might look odd. I got used to it. It enforces consistency in line width, which I prefer.

Three options:
- Accept it (the easiest one).
- Add `// prettier-ignore` above (not recommended).
- Change default print width of 80 chars (not recommended as well).

panel/src/components/Drawers/FormDrawer.vue
```diff
@@ -78,11 +78,14 @@ export default {
 
       if (focus !== false) {
         setTimeout(() => {
-          let autofocus = Object.values(this.fields).filter(field => field.autofocus === true)[0] || null;
+          let autofocus =
+            Object.values(this.fields).filter(
+              (field) => field.autofocus === true
+            )[0] || null;
```
distantnative, line 86: I find the formatting here a bit distracting

panel/src/components/Dialogs/Dialog.vue
johannschopplich: They are (and will be for the bundled CSS output). Again, Prettier adds consistency to every unit.

panel/src/components/Dialogs/UsersDialog.vue
johannschopplich: Sorry, I meant > 4, if they don't intervene with the print width of 80 chars. In this case they stay on one line.

distantnative:
I couldn't spot any whitespace changes that looked problematic to me.

Though the syntax/formatting rubs me in the wrong places in some parts. Especially moving tags with 4 arguments into a single line that then starts to wrap. If it wraps, should be per argument.

johannschopplich:
Vue component props wrapping respects not only the count, but print width as well. If 4 props fits into < 80 chars, it will be put on a single line. 

Interesting note: Deno's default formatter `deno fmt` (which is baked right in) formats JS exactly the same Prettier does. Not that one has to follow those conventions. Those formatting rules are pretty common by now. I like that they are that strict. Lets you care less about how to add readability.

The deal with Prettier is that besides quotes, trailing commas, and some other rules, most of the formatting is strict. I didn't like the style at first when starting developing with Deno. Then, I got used to it. By now I prefer it. Maybe you will feel the same way. Maybe not. Debatable topic. 

There are edge cases in which formatting might look odd. I completely get that. In my opinion, consistent formatting is better than having none at all.

distantnative:
To be honest, I don't think our formatting was that inconsistent. Of course this PR makes it look like it, but in reality some single/double quotes and missing trailing commas aren't a real problem.

Consistency would be great. But I am really on the edge here - while Prettier is strict and makes it consistent, some of it rules seem very random and really not helpful at all. So code becomes consistent, but also harder to read - which in the end I think is a bigger problem than not having 100% consistency.

bastianallgeier:
My biggest problem right now is the mental load for 3.6. I think I could get used to most of the changes for the sake of consistency, but not at this stage. It'd make a lot of sense to get it done before we have to wait for another big release. But to me this seems like a change that the entire team needs to get used to over time. I think especially @distantnative and I felt pretty comfortable with the way we write our JS code, so it does not seem like a pressing issue to us. That makes it even harder to argue for such a big PR right now. I hope you understand that it feels like a really useful step in the long run, but the timing isn't ideal if you ask me.

johannschopplich:
It's your codebase, you need to be comfortable with it. I understand the pressing issue of far more important construction zones for the new release. Personally, I love 100% reproducible code accuracy, but it's just a preference. The PR is an idea, if it doesn't fit, that's OK. Thanks for considering it. Feel free to revisit any time. 😊

bastianallgeier:
I'm 100% with you about the reproducible code accuracy. We already saw how much it helped to agree to alphabetical sorting for methods, vars and other stuff and how tools like eslint and phpmd increased our code quality. All those changes took time to get used to and now they feel natural. 

I would suggest to get back to this somewhen during 3.6.x When the stakes are less high, I believe that we have more head space to submit to the new style. @distantnative what do you think?

distantnative:
I agree and I probably could even get around to using this even if it runs me in wrong ways right now.

But I think we don't have to wait for a big release again, this is a minor change (or let's say a thousand tiny ones) that could happen in 3.6.1 or 3.6.2

distantnative:
Rebased it on 3.6.0 release.

johannschopplich:
@distantnative Thanks!

bastianallgeier (approved):

PR #3819, opened 2021-10-18
New `size` option for line field

afbora:
## Describe the PR

You may have noticed that there is a lot of space when using line to separate the two fields. Now this size can be adjusted.

````yaml
fields:
  line:
    type: line
    size: small # `small`, `medium`, `large`, `huge`
````

### Features

- New `size` option for line field

## Breaking changes

None

## Related issues/ideas

None

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/fields/line.php
+++ b/config/fields/line.php
@@ -1,5 +1,28 @@
 <?php
 
 return [
-    'save' => false
+    'save' => false,
+    'props' => [
+        /**
+         * Unset inherited props
+         */
+        'after'       => null,
+        'autofocus'   => null,
+        'before'      => null,
+        'default'     => null,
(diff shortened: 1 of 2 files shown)


afbora:
Another prop name can be considered, for ex: space, height, etc..

distantnative:
Is this really needed? I feel we are getting a bit too much into tiny tweaks - we don't offer sizes for headlines either etc.

Maybe for the few use cases where this is really needed, a custom CSS would just be fine.

afbora:
I thought of custom CSS but then you can't give different sizes on each page. 
But I think you're right, we don't want the separator at different height on each page.

PR #3821, opened 2021-10-19
System & Language views

bastianallgeier:
- [x] Double-check for settings permissions
- [x] Handle disabled menu items separately 
- [x] Better link handling for plugins


distantnative:
We should either add some mapping for `settings` permissions for some versions. Or throw a warning if a `settings` permissions is defined to change this to the new permission names.

distantnative (dismissed):
LGTM

afbora (commented):
~Plugin informations can't fetch when registering and directory name doesn't match. For example:~

~I unzipped the plugin to ` /site/plugins/kirby-blade-1.9.1` (via github release download) and register with `afbora/kirby-blade`~

### Edited Review

~I found weird issue. When I add `homepage` prop to composer.json, informations gone~ 🤷‍♂️ 

✅ I guess, it was cached issue.

afbora:
Some texts needs translate:

- Debugging
- Server
- Plugins

afbora (approved):
Loved ❤️

PR #3833, opened 2021-10-21
Fixes tabs badge position

### LINKED ISSUE #3822: Tab in front of k-tabs-badge
**Describe the bug**  
If tab A has a badge and tab B is clicked, tab B will be in front of the badge.

**To Reproduce**  
1. Click on a page with tabs
2. Edit some text so the unsaved badge appears
3. Click on a different tab

**Expected behavior**  
The badge shouldn't be blocked by the other tab.

**Screenshots**  
image

**Kirby Version**  
3.6.0-rc.1

 - OS: Windows 10
 - Browser: Chrome
 - Version: 94

afbora:
## Describe the PR

Designed by measuring every pixel for @medienbaecker 😅 

Top/Right/Padding values have been adjusted for cases with long tab labels.

screen-capture-1214-Test - Mægazine-localhost

### Fixes regressions from 3.6.0-rc.1

- Fixed tabs badge position #3822

## Breaking changes

None

## Related issues/ideas

- Fixes #3822

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Forms/FormButtons.vue
+++ b/panel/src/components/Forms/FormButtons.vue
@@ -117,7 +117,7 @@ export default {
         return "negative";
       }
       if (this.mode === "unlock") {
-        return "focus";
+        return "info";
       }
 
       return "notice";
--- a/panel/src/components/Layout/Tabs.vue
+++ b/panel/src/components/Layout/Tabs.vue
@@ -219,11 +219,12 @@ export default {
   position: absolute;
   top: 3px;
   font-variant-numeric: tabular-nums;
-  inset-inline-end: 3px;
+  inset-inline-end: 2px;
   line-height: 1.5;
   padding: 0 .25rem;
   border-radius: 2px;
   font-size: 10px;
+  box-shadow: var(--shadow-md);
 }
 .k-tabs[data-theme="notice"] .k-tabs-badge {
   background: var(--theme-light);
--- a/panel/src/components/Layout/Tabs.vue
+++ b/panel/src/components/Layout/Tabs.vue
@@ -217,11 +217,11 @@ export default {
 }
 .k-tabs-badge {
   position: absolute;
-  top: -.5rem;
+  top: .325rem;
   font-variant-numeric: tabular-nums;
-  inset-inline-end: -.5rem;
-  line-height: var(--leading-normal);
-  padding: 0rem .375rem;
+  inset-inline-end: .325rem;
+  line-height: var(--leading-tight);
+  padding: 0 .25rem;
   border-radius: var(--rounded);
 }
 .k-tabs[data-theme="notice"] .k-tabs-badge {
--- a/panel/src/components/Layout/Tabs.vue
+++ b/panel/src/components/Layout/Tabs.vue
@@ -217,12 +217,13 @@ export default {
 }
 .k-tabs-badge {
   position: absolute;
-  top: .325rem;
+  top: 3px;
   font-variant-numeric: tabular-nums;
-  inset-inline-end: .325rem;
-  line-height: var(--leading-tight);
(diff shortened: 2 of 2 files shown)


distantnative:
I find this a bit too tight compared to all other visual elements. I think the padding could remain `.325rem` and the offset to the right should be still `.5rem`

afbora:
@distantnative Preview of `inset-inline-end: .5rem;`

screen-capture-1215-Test - Mægazine-localhost

distantnative:
Ok but that's an issue we would need to deal with the other version too, right? I'd think the PR faces the same problem if you change the label to end with `webs` not `web`?

afbora:
No, I already adjusted this PR for the edges cases:

screen-capture-1217-Test - Mægazine-localhost

afbora:
I know it's very tight, but I think we have two options. Narrowing the text area a little more or placing the badge in the narrow area.

afbora:
I have a third idea but I'm not sure: `tab icon area`; If there is a change, the badge will appear instead of the tab icon. What do you think?

screen-capture-1218-Test - Mægazine-localhost

bastianallgeier:
Uhh, I love the third idea!! I think that's a great solution and looks much better than the rest.

nilshoerrmann:
Have you thought about just adding a small colored dot to the icon, like an app notification (not sure if it actually needs a distinct count)? Hiding the icon might be confusing to clients who need these visual clues to find the right tab.

bastianallgeier:
That would work well too if all tabs had icons 🙈

nilshoerrmann:
Ah, damn, I totally missed that fact!

afbora:
@bastianallgeier Could it be confusing to hide icons (replace with counter) on changes like @nilshoerrmann said? How will we proceed?

bastianallgeier:
This is turning into a quite complex task for a rather simple UI elements 🙈🤣

I start to wonder if we should go back to simple numbers and make sure they stay in line with the text and break accordingly.

distantnative:
I would love if tab badges still could have a count, also for custom implementations of the tabs component.

afbora:
## Last shots 😅 

I think both seems cool.

### Alternative 1
screen-capture-1221-Test - Mægazine-localhost

### Alternative 2
screen-capture-1220-Test - Mægazine-localhost

bastianallgeier:
@afbora I like alternative 2

afbora:
@bastianallgeier Done for alt 2

screen-capture-1224-Test - Mægazine-localhost

distantnative:
Mmm to me it looks odd that it's slightly lifted from the normal text line

bastianallgeier:
I think the position is really good now. I just made the badges a tiny bit smaller and went for 10px as font size. That's super small, but after all it's just a very tiny visual hint for changes. I also moved the badge a tiny bit closer into the corner. This does no longer mess with the visual alignment of the text. I reduced the border radius a tiny bit as well to make it fit better into the corner and also added a subtle shadow to make the orange stand out a bit better against the background. 

Screenshot 2021-10-27 at 10 03 24

I'm pretty happy with it now. What do you think?

bastianallgeier:
I just saw small glitch in the form bar. I fixed it here to avoid yet another PR for such a small change.

distantnative:
I think it looks super squeezed and odd but that's personal taste. We can go with it. Maybe the drop shadow on it doesn't make sense with this positioning?

bastianallgeier:
I think the main prio for rc.2 is to fix the bug. I personally like the position now and don't think it looks too squeezed. The drop shadow also makes it more badgy in my eyes. But I'm super happy about more suggestions later. It's definitely not my most favourite UI element in the Panel :)

Issue #3827, opened 2021-10-20
[3.6.0-rc.1] When user with a limited areas access logs in, system shouldn't open the site


adamkiss:
Scenario:
I have a user setup with lesser permissions, allowing them only access to custom areas for instance:

```
permissions:
  access:
    panel: true
    site: false
    settings: false
    users: false
    custom-area1: true
    custom-area2: true
```

Situation:
When the user logs in, first thing they'll see is the Site view with a  "You've no access to this resource" message.

Expected:
- go to first available area to the user (not sure how ordering works here, though)
- allow per-role based post login urls - which I feel is fairly elegant solution

Dream:
I just can imagine doing something like:
```
panel:
  post-login-url: "/panel/custom-area/project/{{ user.project.id }}"
```

bastianallgeier:
✅

PR #3834, opened 2021-10-21
Custom redirect after login

bastianallgeier:
When a user logs in that does not have access to the site view, they end up on the "This view does not exist" error page, which is quite nasty. 

This PR solves this problem in two steps: 

1. The login redirect is now a lot smarter and checks the menu first to find the next best accessible view instead
2. There's a new `home` option for User blueprints. This can be used to set a custom path or even use a string query to redirect to a page view, a file view, a user view or any other URL

### Simple paths
```yaml
title: Editor
home: /account
```

### Queries to panel views
```yaml
title: Editor
home: "{{ site.find('blog').children.first.panel.url }}" 
```

### Queries to pages
```yaml
title: Editor
home: "{{ site.find('blog').children.first.url }}" ​
```

## Todos

- [x] Check for valid URLs in home method
- [x] Remove any queries or other parts of the URL that could be used as an attack
- [x] Only allow URLs on the same domain

## Fixes

- https://github.com/getkirby/kirby/issues/3827

## Implements

- https://kirby.nolt.io/10


src/Panel/Panel.php
```diff
@@ -153,11 +155,21 @@ public static function firewall(?User $user = null, ?string $areaId = null): boo
...
+    /**
+     * Handles redirection to the correct area
+     * after login or when visiting the login or installation
+     * views after being already authenticated
+     *
+     * @return void
+     */
+    public static function goHome(): void
+    {
+        static::go(Home::url());
```
distantnative: Do we need this new method? Now that we have the Home class, we could just put that code in the action callback below.

bastianallgeier: I was thinking about it as well. Could easily just be removed.

bastianallgeier: On the other hand, it's easier to test the results that way. The callback functions in routes are always a PITA to test.

distantnative: But what results? It's a whole method with one line inside that replaces a one line in the call-back function. In terms of test coverage, we'd still have to test now the function call to `goHome` in the callback function. Then we rather could test directly for `static::go(Home::url())`

bastianallgeier: @distantnative I changed it. It makes a lot more sense this way.

src/Panel/Home.php
```diff
@@ -0,0 +1,276 @@
...
+        $uri->query  = '';
+
+        // get a clean version of the URL
+        $url = $uri->toString();
+
+        // Don't further inspect URLs outside of the Panel
+        if (static::isPanelUrl($url) === false) {
+            return $url;
+        }
+
```
distantnative, line 242: Shouldn't we already check here if the user has Panel access in general? And not in `static::alternative()`

bastianallgeier: You are totally right. I looked at the entire class again after your comment and I think we can be a lot stricter with user checks in the other methods as well. I refactored it a bit and it's IMHO a lot cleaner now.

PR #3836, opened 2021-10-22
Update removed `settings` permission with `system`

afbora:
Since the `settings` permission has been removed, only those with `system` permission can now see the Kirby version.


bastianallgeier (approved):
I'm really glad you spotted this!

PR #3837, opened 2021-10-22
Return type declarations Cms folder

modufolio:
## Describe the PR

Added Return type declarations to the functions in the Cms folder

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
     "michelf/php-smartypants": "1.8.1",
     "mustangostang/spyc": "0.6.3",
     "phpmailer/phpmailer": "6.5.0",
-    "true/punycode": "2.1.1"
+    "true/punycode": "2.1.1",
+    "vimeo/psalm:": "3.*"
   },
   "config": {
     "optimize-autoloader": true,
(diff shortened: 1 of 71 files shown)


modufolio:
This only works in 8.0 because of union types and static type

distantnative:
We have stayed away from declaring specific classes as return types as this has been causing problems with how PHP handles these type requirements when working with class inheritance etc.

I think PHP handles this better in more recent releases (not sure if it was PHP 8 but likely) but as long as we support older PHP versions, I don't think this would be an option.

@lukasbestle will know more

modufolio:
Indeed I think we should wait with this when we won't support php7.* anymore.

lukasbestle:
Full covariance and contravariance support was added in PHP 7.4. Security support for PHP 7.3 will end on 6 Dec 2021, so the return type hints would actually be viable soon. But `static` and union types will indeed have to wait for another year.

distantnative:
That's great.

@modufolio definitely thank you for the effort!

I think then it makes sense to split this PR:
- normal return types for end of this year
- union types and `static` for end of 2022

modufolio:
@distantnative Understood! I will create PR's for each src folder with normal return types.

afbora:
@modufolio So can we close this since you'll create a separate PR?

PR #3848, opened 2021-10-25
Prevent layout shift when overlays open

### LINKED ISSUE #3823: Layout shift when dialog or drawer is opened
**Describe the bug**  
Opening a dialog (such as rename) or block drawer causes a layout shift due to the scroll bar disappearing.

**Expected behavior**  
There should not be any layout shift

**Kirby Version**  
3.6.0-rc.1

 - OS: Windows 10
 - Browser: Chrome
 - Version: 94

### LINKED ISSUE #3923: Removing layout from layout field causes scrollbar disappear
**Describe the bug**  
After the layout is deleted from layout field the scroll bar disappears.

**To Reproduce**  
1. Add new layout to layout field.
2. Remove the layout.
3. The page scrollbar disappears (it seems HTML element gets *overflow: hidden*).

**Kirby Version**  
3.6.0-rc.4

**Desktop:**  
 - OS: Windows 10
 - Browser: Mozilla Firefox Developer, Google Chrome
 - Version: 95.0b4, 95.0.4638.69

bastianallgeier:
## Describe the PR

Scrolling is now disabled with a new `preventScrolling` store action that activates a new data-prevent-scrolling attribute on the panel component. This makes it possible to use simple CSS rules for the styling. To prevent scrolling the body is now fixed instead of hiding the overflow. This will no longer lead to a layout shift due to a missing scrollbar. 

## Fixes 

- Fixed layout shift when opening overlays #3823
- Fixed dialog reset when dialogs are not closed properly https://github.com/getkirby/kirby/issues/3923

## Breaking changes

- None 

## Related issues/ideas

- Fixes #3823

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Layout/Overlay.vue
+++ b/panel/src/components/Layout/Overlay.vue
@@ -60,10 +60,11 @@ export default {
 
       this.isOpen = false;
       this.$emit("close");
-      this.restoreScrollPosition();
 
-      // enable scrolling of background
-      this.$store.dispatch("preventScrolling", false);
+      // restore scrolling
+      document.body.style.position = "static";
+      document.body.style.top = 0;
+      this.restoreScrollPosition();
 
       // unbind events
       this.$events.$off("keydown.esc", this.close);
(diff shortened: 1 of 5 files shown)


bastianallgeier:
Damn, it's not a solution. The scroll position is broken with the new CSS styles :(

distantnative:
Also, I'd suggest to use a shorter name for the store. Just activating/deactivating `scroll`?

afbora:
Double scrolling when 1024x768px screens

image (31)

bastianallgeier:
I think we will need to keep this issue open for after the release. It's simply too complex in comparison to the result.

distantnative:
Not sure if we can solve the double scrolling bar with the current setup.
Since we are scrolling now on body for the main content and not a separate div, the scrolling of the drawer and the scrolling of the main content are very disconnected.

For not having a separate scroll bar in the drawer, we would somehow make `body`'s height the height needed for the drawer content, scroll though that via the main body scroll bar, but adjust the top position of the content outside the overlay to remain in place (to lock scroll). I don't think this is feasible in any sensible way.

I would think the two scroll bars are a given by the new layout structure.

I was wondering what benefit it had to use a scrollbar on body and not a scrollable div. I know we had a reason to switch that. But I can't remember which one. Would be maybe worth it to compare that benefit with these pain points. And then decide whether to stick with this (Bastian's solution seems to work for me other than the double scroll bar in drawers with too little height) - or to return to a scrollable div for the main content.

lukasbestle:
I wonder if it's really necessary to prevent scrolling on the `<body>` while an overlay is open. Sometimes it might even make sense to scroll the background, e.g. when you want to check something further down or up the page while you are entering data in the overlay. Wouldn't this make the setup a lot easier because the scrollbar of `<body>` could just be enabled no matter if the overlay is open or not? Or will there still be layout shifts?

PR #3850, opened 2021-10-26
Improve permssion handling for changing roles

### LINKED ISSUE #2919: Non-admin user cannot change role of another user
I have a project running on Kirby 3.4.4 and I'm still facing this issue.
The bug seems to emerge from [`UserPermissions->canChangeRole()`](https://github.com/getkirby/kirby/blob/6112f1cb8ebf9c36e8a9f97ae08acc910ddbc792/src/Cms/UserPermissions.php#L37), where the number of available user roles is checked with [`User->roles()`](https://github.com/getkirby/kirby/blob/6112f1cb8ebf9c36e8a9f97ae08acc910ddbc792/src/Cms/User.php#L685), which always returns only the role of the specified user, unless the logged in user is an admin.

_Originally posted by @PaulRaschke in https://github.com/getkirby/kirby/issues/1919#issuecomment-721215752_

bastianallgeier:
## Describe the PR

The changeRole permissions are pretty complex and so far they haven't been working great at all. This PR is fixing this. 

I'm trying my best to describe all the cases. 

### Admins

Admins can change the roles of all other users, no matter what their permission setup looks like. They got super powers. But an admin cannot change its own role if they are the last admin. 

### Other roles

Every other role can be setup to be able to change other roles, except admins. No matter what you do, other roles can never make themselves or others admin and they cannot change admin roles. 

Basic changeRole permissions follow the same procedure as always. You can set the general rule in the user blueprint in the permissions array. Changing roles is enabled by default. To disable it you can do this: 

```yaml
# users/editor.yml
permissions: 
  # other users
  users: 
    changeRole: false
  # own account
  user: 
     changeRole: false
```

Those global permission settings can then be overwritten individually with the option array. 

```yaml
# users/client.yml
options: 
  changeRole: 
    editor: true
```    

The editor role is now able to change the role of clients, but not of any other user.

### DIFF
--- a/config/areas/users/dialogs.php
+++ b/config/areas/users/dialogs.php
@@ -29,7 +29,7 @@ return [
                         'language' => Field::translation([
                             'required' => true
                         ]),
-                        'role' => Field::role([
+                        'role' => Field::role($kirby->roles()->canBeCreated(), [
                             'required' => true
                         ])
                     ],
@@ -206,14 +206,14 @@ return [
                 'component' => 'k-form-dialog',
                 'props' => [
                     'fields' => [
-                        'role' => Field::role([
+                        'role' => Field::role($user->roles(), [
                             'label'    => t('user.changeRole.select'),
                             'required' => true,
                         ])
                     ],
                     'submitButton' => t('user.changeRole'),
                     'value' => [
-                        'role' => $user->role()->name()
+                        'role' => $user->role()->id()
                     ]
                 ]
             ];
--- a/src/Cms/User.php
+++ b/src/Cms/User.php
@@ -646,24 +646,38 @@ class User extends ModelWithContent
         // a collection with just the one role of the user
         $myRole = $roles->filter('id', $this->role()->id());
 
-        // if there's an authenticated user …
-        if ($user = $kirby->user()) {
-
-            // admin users can select pretty much any role
-            if ($user->isAdmin() === true) {
-                // except if the user is the last admin
-                if ($this->isLastAdmin() === true) {
-                    // in which case they have to stay admin
-                    return $myRole;
-                }
-
-                // return all roles for mighty admins
-                return $roles;
-            }
+        // if there's no authenticated user
+        // only the already assigned role is available
+        if (!$authUser = $kirby->user()) {
+            return $myRole;
         }
 
-        // any other user can only keep their role
-        return $myRole;
+        // admin users can select pretty much any role
+        if ($authUser->isAdmin() === true) {
+            // except if the user is the last admin
+            if ($this->isLastAdmin() === true) {
+                // in which case they have to stay admin
+                return $myRole;
+            }
+
+            // return all roles for mighty admins
+            return $roles;
+        }
+
+        // any other role cannot change the roles for admins
+        if ($this->isAdmin() === true) {
+            return $myRole;
+        }
+
+        // any other role cannot promote a user to admin, even
+        // if they have full changeRole permissions
+        $roles = $roles->filter('id', '!=', 'admin');
+
+        if ($this->permissions()->changeRole() === false) {
+            return $myRole;
+        }
+
+        return $roles;
     }
 
     /**
--- a/src/Cms/UserPermissions.php
+++ b/src/Cms/UserPermissions.php
@@ -19,11 +19,16 @@ class UserPermissions extends ModelPermissions
     protected $category = 'users';
 
     /**
+     * @var \Kirby\Cms\User
+     */
+    protected $model;
+
+    /**
      * UserPermissions constructor
      *
-     * @param \Kirby\Cms\Model $model
+     * @param \Kirby\Cms\User $model
      */
-    public function __construct(Model $model)
+    public function __construct(User $model)
     {
         parent::__construct($model);
 
@@ -36,7 +41,30 @@ class UserPermissions extends ModelPermissions
      */
     protected function canChangeRole(): bool
     {
-        return $this->model->roles()->count() > 1;
+        $roles = $this->model->kirby()->roles();
+
+        // authenticated admins can always change the role of
+        // a user unless the user is the last admin
+        if ($this->user->isAdmin() === true) {
+            if ($this->model->isLastAdmin() === true) {
+                return false;
+            }
+
+            return $roles->count() > 1;
+        }
+
+        // if the user is an admin and the authenticated
+        // user isn't, the authenticated user cannot
+        // change the role. No matter how the permissions
+        // are set.
+        if ($this->model->isAdmin() === true) {
+            return false;
+        }
+
+        // non-admins cannot promote users to admins
+        $roles = $roles->filter('id', '!=', 'admin');
+
+        return $roles->count() > 1;
     }
 
     /**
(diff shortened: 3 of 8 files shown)


afbora:
@bastianallgeier Still not possible following syntax right?

````yaml
# specify allowed roles
permissions:
  users:
    changeRole: 
      - editor
      - customer
````

**Comment by bastianallgeier** [comment]
@afbora no, that would be a more drastic change. You can achieve the same with the options settings for those roles.

**Comment by afbora** [comment]
I don't want an editor to assign itself to the customer role. It can assign in the following setup but cannot change it.

````yaml
# accounting.yml
permissions: 
  user: 
     changeRole: true
     
# editor.yml
permissions: 
  user: 
     changeRole: true
     
# customer.yml
permissions: 
  user: 
     changeRole: true

options: 
  changeRole: 
    editor: false
````

**Comment by bastianallgeier** [comment]
@afbora could you describe some more what's working and what not?

**Comment by afbora** [comment]
I want to set the allowed roles on a per role basis. I don't want all roles to appear like below. E.g; I want the editor to be only editor or customer.

screen-capture-1225-Your account - Mægazine-localhost

**Comment by afbora** [comment]
`changeRole` partially works, but it doesn't make sense to see the role I don't allow in the list.

**Comment by bastianallgeier** [comment]
Let me get that right. 

- You are logged in with the editor role
- You are trying to change your own role
- You only want Customer and Editor as options

Right?

**Comment by afbora** [comment]
Exactly.

**Comment by afbora** [comment]
You can see the problem here. When you do not allow the accounting role; You can switch to the accounting role, but you can no longer change it.

````yaml
# accounting.yml
permissions: 
  user: 
    changeRole: true

options: 
  changeRole: 
    editor: false
    customer: false

**Comment by bastianallgeier** [comment]
Ok, let me try to get the blueprint setup right for this. 

```yaml
# accounting.yml
permissions: 
  user: 
     changeRole: true

options: 
  changeRole: 
    editor: false
     
# editor.yml
permissions: 
  user: 
     changeRole: true
     
# customer.yml
permissions: 
  user: 
     changeRole: true

```

With this setup, all three roles are allowed to change the role. But the editor is not allowed the change to assign the accounting role.

bastianallgeier:
Let's get this straight first before we release it. It's still too much of a brain-bending monster.

afbora:
Still I can see all roles:

screen-capture-1226-Your account - Mægazine-localhost

bastianallgeier:
I'll give it another look after the RC today.

distantnative:
My theory, but maybe I am wrong with this.... but putting it out here, so @bastianallgeier can see if he understands what I mean:

In https://github.com/getkirby/kirby/pull/3850/files#diff-36731f6d2b3a5a5b67e878c0d65b00a17b8a4eb30fdb79567ab7479d04264fd8L209 we use `$user->roles()` as options for the change role dialog. 

`User::roles()` has been modified quite a bit https://github.com/getkirby/kirby/pull/3850/files#diff-3f1aec908dc4b4086b4a9283ad0ef0eb28e2579e5097434d44a4b255443d363aL641.

However, almost all of the logic is only based on the authenticated user and only very little logic on the user to-be-modified (only if that user is the last admin). 

I feel there is logic missing whether the to-be-modified user is allowed to take on a specific role (not just whether the authenticated user is allowed to assign the role).

Moreover, I think your example is hitting a wall:
```yaml
# accounting.yml
permissions: 
  user: 
     changeRole: true

options: 
  changeRole: 
    editor: false
```

If we would call `$user->permissions()->can()`, we would never reach the option `changeRole`. The code https://github.com/getkirby/kirby/blob/develop/src/Cms/ModelPermissions.php#L61 checks for a method `canChangeRole` - which `UserPermissions` class has. But that one is a simple `$this->model->roles()->count() > 1` evaluation and doesn't take the `options` option into account.


Overall, the system feels really heard, I know @bastianallgeier you aren't happy with it either. Maybe we need to live with that secondary bug for 3.6 and rather work on refactoring the system overall. Being based on `permissions` and `options` across multiple blueprints is hard for the proper logic but even for the users to make sense of it.

lukasbestle:
I think the syntax proposed by Bastian (boolean permission with extended configuration in `options`) makes a lot of sense. It avoids that issue with unclear value types in the permission array.

IMO it‘s correct that the permissions are based on the current user, not the user you want to change the role of. That fits very well with the rest of our permissions, which all define what the current user should have power for. In this case it‘s about which roles the current user can manage (e.g. a content director may be able to create or promote editors and translators while a HR manager may be able to create employees without or with limited Panel access).

There is only one edge case: Let‘s say there is a user with a more privileged role with a lot of permissions (let‘s call it `manager`). If the current user opens the change role dialog for that user, they will only see the roles they can manage (let‘s say `editor` and `translator`). The current user would now be able to “downgrade” the other user but not able to revert their mistake as they are not allowed to manage the `manager` role.

One way to fix this edge case would be to disallow access to users in roles that aren’t listed in the current user’s blueprint. Either completely (probably useful anyway in larger companies due to privacy concerns) or at least access to the change role
dialog.

We could rename the `changeRole` option (*not* the permission) to something like `managedRoles`. This would then have an effect on accessing users, creating users and changing the role of existing ones. If not set, it would still have the basic defaults regarding admin users.

lukasbestle:
Oh, I think I got the current implementation the wrong way round. Seems like the options are currently defined in the blueprint of the user that is changed, not in the one of the current user, right?

Shouldn’t it be the other way around so that the options are the actual *options* that the current user will have? I.e. it would just be a list of roles the current user has access to in the dialog in line with my suggestion above.

teichsta:
@bastianallgeier i am not quite sure about the status of this PR. The PR is closed but you postponed the milestone und the issue #2919 is also still open. Could you shed a little light on the status, please? Thanks :)

teichsta:
Hi @bastianallgeier @distantnative @lukasbestle, is this PR (or any successor) going to find it's way into a kirby release? Would be great if you could give a short update notice on this. Thanks, Thomas E.-E.

teichsta:
@bastianallgeier @distantnative @lukasbestle any news if and if yes, how you are going to tackle this issue? As we are eagerly waiting to delegate changing of to non-admin users it would be great to learn about your plans. Thanks in advance, Thomas

distantnative:
@teichsta It's definitely a thing we would like to solve but so far has proved rather tricky or overtaken by other priorities. My honest estimate would be that it won't be in 3.8.0, maybe in 3.8.x. I know it's not the answer you wish to hear or we would wish to deliver, but trying to reflect on the realities as they are.

teichsta:
@distantnative or maybe 3.9.3 ;-) ?

jaro-io:
@distantnative @lukasbestle @bastianallgeier @afbora

any update on this?

🌳🍃

samyouel:
@distantnative @lukasbestle @bastianallgeier @afbora

just throwing my hat in the ring here, is there anything planned in this area to fine tune the user permissions for being able to view and edit other users based on their roles?

That would be a great feature to have. Thank you so much :)

👍🏼

lukasbestle:
@samyouel With [user options](https://getkirby.com/docs/guide/users/permissions#specific-permissions-in-page-file-user-or-site-options__user-options) you can customize all [user permissions](https://getkirby.com/docs/reference/panel/blueprints/user#permissions__users) based on the combination of the acting user role and target user role.

In case you are missing permission settings, please create a suggestion over at https://feedback.getkirby.com. Closed PRs are not a good place to discuss ideas because they don't show up in any overviews of open tasks.

PR #3855, opened 2021-10-27
Dynamic file instance 

### LINKED ISSUE #3849: [3.6.0-rc.1 ] minwidth in file blueprint not working
**Describe the bug**  
The minwidth property in file blueprint doesnt lead to upload permission and error message.
Instead smaller files can regulary be uploaded.

**To Reproduce**  
1. Check out kirby 3.6
2. set minwidth property in file blueprint
3. upload smaller file in panel
File is being uploaded

**Expected behavior**  
File upload is permitted

**Screenshots**  

**Kirby Version**  
3.6.0-rc-1

afbora:
## Describe the PR

See the discussion: https://github.com/getkirby/kirby/issues/3849#issuecomment-952619922

### Fixed regressions from 3.6.0-rc.2
- Fixed upload validations #3849 

## Breaking changes

None

## Related issues/ideas

- Fixes #3849 

## Ready?

- [x] CI checks pass

## When merging


distantnative:
Should we create new method that returns the proper object based on the type - to keep the code DRY?

afbora:
I'm not very good at method naming. If you have a better name suggestion than `::typeModel()`, please let us know.

bastianallgeier (commented):
The implementation looks good in general. I just wonder if this should be a static method instead and somehow clarify what happens underneath. My first thought was something like: 

```php
$file = File::detect($source);
```

distantnative:
I just realized that we already had `File::asset()` which also would be the proper name with our properties etc. - we would only need to change that you can pass props to the method.

afbora:
> The implementation looks good in general. I just wonder if this should be a static method instead and somehow clarify what happens underneath. My first thought was something like:

@bastianallgeier If we use static we always need to send the file instance or type as the first parameter. That's why I used non static  method.

> I just realized that we already had File::asset() which also would be the proper name with our properties etc. - we would only need to change that you can pass props to the method.

@distantnative I thought of this at first but didn't use this method as it sets and caches `self::$asset` property. Do you mind?

distantnative:
How about something like this:

```php
    public function asset(?array $props = null)
    { 

        $defaults = [
             'root' => $this->root(),
             'url'  => $this->url()
         ];

        // only used cached object or default values
        // when no props passed to the method
        if ($props === null) {
            if ($this->asset !== null) {
                return $this->asset;
            }

            $props = $defaults;
         }

         // create asset object based on file type
         switch ($this->type()) {
             case 'image':
                 $asset = new Image($props);
             default:
                 $asset = new File($props);
         }

        // only cache object for default props
        if ($props === $defaults) {
            $this->asset = $asset;
        }

         return $asset;
     }
```

Maybe we can make it even cleaner.

afbora:
This method seems to have two different functions. I think it would be better to have separate methods in terms of clean code.

afbora:
I would use @bastianallgeier's method name idea: `::detect()` as non-static.

distantnative:
For me, `detect` as name doesn't make a lot of sense. It implies that the method is trying to figure out the type... but that's what `::type()` does. We want a method that returns an object - which we so far refer to as `asset` in this context - based on the type.

afbora:
I think it is confusing that it sets the property and reads it from the cache, depending on whether the prop is sent or not.

bastianallgeier:
Wait a second. When asset already returns the correct object, do we have to implement any of this at all? 

Wouldn't this be enough in case of the FileAction::create method 

```php
$file = static::factory($props);
$upload = $file->asset();
```

And in case of the replace action we could create a dummy File object and do the same: 

```php
$upload = static::factory(['root' => $source])->asset();
```

Am I missing something?

afbora:
@bastianallgeier I tried and 18 tests failed of 35 tests for `FileActionsTest` test..

afbora:
I tried things, but it seems to work like this. I don't know if it makes sense, what do you think? 🤔

bastianallgeier:
@afbora let's go for it. We can still change that later if we encounter more problems.

bastianallgeier (approved):

afbora:
@getkirby/kirby-staff Regardless of this PR, why don't we move the `\Kirby\Image` namespace under `\Kirby\Filesystem`? It would be great if we collect other file classes here in the future.

- /Filesystem
  - /Image (Image helper classes)
  - /Video (Video helper classes)
  - File.php
  - Image.php
  - Video.php

image (32)

bastianallgeier:
@afbora yes, we should definitely do that. But before that we urgently need to unfuck the image processing logic. Those Darkroom classes are a nightmare.

PR #3858, opened 2021-10-28
Move changes dropdown to the backend to fix broken links and other is…

### LINKED ISSUE #3854: [3.6.0-rc.2] Files name and link broken in unsaved changes
**Describe the bug**  

**To Reproduce**  
1. Go to panel (demokit)
2. Go to file detail (`/pages/portfolio+projects+sky/files/tent-in-the-woods.jpg`)
3. You'll see unsaved changes
4. See file name missing
5. Nevertheless click it, you'll see broken link that missing `files` segment (`/panel/pages/portfolio+projects+sky+tent-in-the-woods.jpg`)

**Kirby Version**  
3.6.0-rc.2

bastianallgeier:
## Describe the PR

The changes dropdown was always a bit hacky. We had to send a single request for every model in the changes store and then build the options in JS. After the move to Fiber the links and labels were broken. I decided to not fix it in JS, but move to a Fiber dropdown instead. This works better than expected but also requires a couple changes throughout multiple classes. It's all tested and should not be a big deal, but it would be good if @distantnative and @afbora could have a closer look. 

## Related issues/ideas

- Fixes #3854

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Forms/FormIndicator.vue
+++ b/panel/src/components/Forms/FormIndicator.vue
@@ -1,9 +1,10 @@
 <template>
   <k-dropdown v-if="hasChanges" class="k-form-indicator">
-    <k-button class="k-topbar-button" @click="toggle">
-      <k-icon type="edit" class="k-form-indicator-icon" />
-    </k-button>
-
+    <k-button
+      class="k-form-indicator-toggle k-topbar-button"
+      icon="edit"
+      @click="toggle"
+    />
     <k-dropdown-content
       ref="list"
       align="right"
@@ -14,12 +15,11 @@
       </p>
       <hr>
       <k-dropdown-item
-        v-for="entry in entries"
-        :key="entry.id"
-        :icon="entry.icon"
-        @click.native.stop="go(entry.target)"
+        v-for="option in options"
+        :key="option.id"
+        v-bind="option"
       >
-        {{ entry.label }}
+        {{ option.text }}
       </k-dropdown-item>
     </k-dropdown-content>
   </k-dropdown>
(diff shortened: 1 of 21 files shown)


panel/src/components/Forms/FormIndicator.vue
```diff
@@ -40,10 +40,10 @@ export default {
     ids() {
       return Object
         .entries(this.store)
-        .filter(([id, model]) => {
+        .filter(([, model]) => {
           return Object.keys(model.changes).length > 0;
         })
-        .map(([id, model]) => {
+        .map(([id]) => {
```
distantnative: Looking at this... wouldn't it make more sense to go with `Object.keys` in line 42 then?

bastianallgeier: I thought it's quite nice to get access to the model and the id like this because the model is needed in 44

distantnative: I mean, it's no dealbreaker, but I jus thought that this is shorter and still not so bad on line 44?

```js
return Object
  .keys(this.store)
  .filter((id) => {
    return Object.keys(this.store[id].changes).length > 0;
  })
  ```

distantnative:
I like this a lot - and seems to work fine for me so far.
It's a good step - also for when we will eventually move unsaved changes to the server (and thus the backend anyways).

panel/src/components/Forms/FormIndicator.vue
bastianallgeier: You are right. I updated it.

bastianallgeier:
@distantnative I agree. It helps quite a lot to think about a backend implementation already.

distantnative (approved):
We should add a note to the final release notes that all unsaved changes should be cleared before upgrading to a new major version.

PR #3862, opened 2021-10-28
Alias for minlength & maxlength in counter mixin

### LINKED ISSUE #3861: [3.6.0-rc.2] Missing max/min length indicator Textarea
**Describe the bug** 
Textareas with length validation (min- or maxlength) are missing the indicator on how the Characters are limited

**To Reproduce**  
1. Add length validation to Textarea in Blueprint

```
intro:
  label: Intro
  type: textarea
  maxlength: 160

```

**Screenshots**  
textarea

**Kirby Version**  
3.6.0-rc.2

bastianallgeier:
Too short for a description :)

### DIFF
--- a/panel/src/mixins/forms/counter.js
+++ b/panel/src/mixins/forms/counter.js
@@ -23,8 +23,8 @@ export default {
       }
       return {
         count: count,
-        min: this.min,
-        max: this.max,
+        min: this.min ?? this.minlength,
+        max: this.max ?? this.maxlength,
       };
     }
   },


distantnative:
This works as a quick fix, but shouldn't we (maybe later) rather alias `min/minlength` and `max/maxlength` on the backend field props level?

bastianallgeier:
Yep, that makes sense as well. Maybe we should add a ticket for it to the backlog.

distantnative (approved):

afbora (approved):

PR #3863, opened 2021-10-28
Add missing `@since` tags

distantnative:


afbora:
New fiber classes?

distantnative:
Probably too, yes. I ran into these just when working on the docs, so I put them in a PR. Didn't go through all systematically

bastianallgeier (approved):

bastianallgeier:
I'd say we already merge this and add additional since tags later whenever it fits. Otherwise this will quickly collect too many changes.

PR #3867, opened 2021-10-29
Undo breaking change in FieldClass::store

### LINKED ISSUE #3866: Undo breaking change on `Kirby\Form\FieldClass`
> Declaration of Kirby\Form\Field\CustomField::store($value) must be compatible with Kirby\Form\FieldClass::store($value = null)

Existing plugins started giving compatibility issues. This breaking change should be documented.

bastianallgeier:
## Describe the PR

The `FieldClass::store` method introduced an unnecessary breaking change. This PR reverts that change.

### DIFF
--- a/src/Form/Field/BlocksField.php
+++ b/src/Form/Field/BlocksField.php
@@ -187,7 +187,7 @@ class BlocksField extends FieldClass
         ];
     }
 
-    public function store($value = null)
+    public function store($value)
     {
         $blocks = $this->blocksToValues((array)$value, 'content');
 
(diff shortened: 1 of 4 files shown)


afbora (approved):

afbora:
There will be a break change but shouldn't we also remove `null` for `fill()`? It would be illogical to call the `->fill()` method without parameter/data.

bastianallgeier:
@afbora that's true, but I don't want to do that now with all the other stuff on the list.

Issue #3860, opened 2021-10-28
[3.6.0-rc.2] Permissions issue in user blueprint


vauvarin:
**Describe the bug**  
If I set permissions in a user blueprint like this:
```
permissions:
    access:
        users: false
        ...
```
The "users" item menu is hidden but when the user go to its "Your account" panel page and try to change its (email, status, language, parameters) by clicking on these links, no pop-up appear.

**To Reproduce**  
Set permissions like this in a user blueprint:
```
permissions:
    access:
        users: false
        ...
```

**Expected behavior**  
The "users" item menu is hidden but the user can edit its own account informations (emails, status, languages, parameters)

**Kirby Version**  
3.6.0-rc.2

bastianallgeier:
✅

Issue #2580, opened 2020-04-18
Expose user files via the account route


texnixe:
**Describe the bug**  
Currently, if a user role doesn't have access to the `users` (permissions => access => `users: false`) view, they cannot access files uploaded to their account, because these files live under the `panel/users/userid/files/` URL, not the `account` URL.

**To Reproduce**  
1. Create a user role files with access to users view limited and a files section
2. Go to user account and upload a file
3. Click on file to access it
4. See `You are not allowed to access this part of the panel` error message

**Expected behavior**  
Users should be able to view files uploaded to their account even if they don't have general access to the Users view.

**Kirby Version**  
3.3.5

I thought I'd come across this before, but couldn't find any related issue, so forgive me if this is a duplicate.

bastianallgeier:
✅

PR #3868, opened 2021-10-29
Account permissions issues

bastianallgeier:
The mixture between user administration and account administration led to problems before, but with the new area setup it created one really big issue. When access to the users area is disabled, the account view and all subviews are also broken. 

This is only fixable in a clean way by creating an autonomous account area. Sounds simple but needs quite a couple redundant pieces (dialogs, dropdowns, api methods) To avoid duplicating code, I tried to re-use the logic for users area pieces as much as possible. The easiest way to implement all of this is to create separate routes for the account. I.e. `/panel/dialogs/account/changeName` or `/api/account` Then we can also change the `Kirby\Panel\User::url()` method to always return account URLs for the logged in user. This leads to consistent results for every piece of the panel. Dialogs, dropdowns, fields, sections, views. It took a lot of changes and I hope the PR is not too terrible to review, but I think it's the best way forward. 

It also solves one major issue that has been lying around with a "difficult" label for a while. @distantnative already tried to solve it, but I think we can hit two birds with one stone with this one. 

## Todos

- [x] Fix unit tests
- [x] Write new unit tests for all the new account routes

## Fixes

- https://github.com/getkirby/kirby/issues/3860
- https://github.com/getkirby/kirby/issues/2580


config/areas/account/dialogs.php
```diff
@@ -0,0 +1,70 @@
...
+
+return [
+
+    // change email
+    'account.changeEmail' => [
+        'pattern' => '(account)/changeEmail',
+        'load'    => $dialogs['user.changeEmail']['load'],
+        'submit'  => $dialogs['user.changeEmail']['submit'],
+    ],
+
```
distantnative, line 13: Could these be shorter like this?

```php
'account.changeEmail' => $dialogs['user.changeEmail'] + [
    'pattern' => '(account)/changeEmail'
]
```

i18n/translations/en.json
```diff
@@ -1,4 +1,8 @@
 {
+  "account.changeName": "Change your name",
+  "account.delete": "Delete your account",
+  "account.delete.confirm": "Do you really want to delete your account? You will be logged out immediately. Your account cannot be recovered.",
+
```
distantnative, line 5: To simplify it a bit more again: wouldn't it be fine to use the normal `users` i18n strings for the account view as well?

src/Cms/Find.php
```diff
@@ -157,6 +157,12 @@ public static function parent(string $path)
      */
     public static function user(string $id = null)
     {
+        // account is a reserved word to find the current
+        // user. It's used in various API and area routes.
```
distantnative, line 161: Since we are making `account` a reserved word - do we need to ensure that it doesn't get by (slim) chance generated as ID for a user?

src/Panel/User.php
```diff
@@ -151,6 +156,11 @@ protected function imageSource(string $query = null)
      */
     public function path(): string
     {
+        // path to your own account
+        if ($this->model->is($this->model->kirby()->user()) === true) {
```
distantnative: Can we use `$this->model->isLoggedIn()` here instead?

distantnative:
I left a few comments but looks good in general.
CI is failing because the new route in `config/api/routes/users.php`. Since all other routes in there are also untested so far - maybe something we can live this for now?

i18n/translations/en.json
bastianallgeier: When working on this, it became clear how weird the translations are when you work on your own account. Of course we could keep it, but I wonder if it's making this PR that much simpler compared to the enhanced UX when you read: "Delete your account" instead of "Delete this user"

src/Cms/Find.php
bastianallgeier: Yep, that should probably be a part of the UserRule class. I will implement that.

src/Panel/User.php
bastianallgeier: Yes, definitely!

i18n/translations/en.json
distantnative: Ok, fine with that :)

src/Cms/UserRules.php
```diff
@@ -299,6 +299,10 @@ public static function validEmail(User $user, string $email, bool $strict = fals
      */
     public static function validId(User $user, string $id): bool
```
distantnative, line 300: `UserRules::validId` should be used in `UserActions::createId()` instead of only `$this->kirby()->users()->has($id)`.

config/areas/account/dialogs.php
bastianallgeier: I tried and it has to be the other way around: 

```php
'account.changeEmail' => [
    'pattern' => '(account)/changeEmail'
] + $dialogs['user.changeEmail']
```

That's exactly the reason why I almost never use that syntax. I often don't remember the correct order to overwrite keys and I think it's less readable this way.

distantnative: Ok, fair 👍

src/Cms/UserRules.php
bastianallgeier: Absolutely! But how do you write a try/catch in a while clause condition?

distantnative: Something like that? 

```php
while(true) {
    try {
        if (UserRules::validId($id) === true) {
            break;
        }
    } catch (Throwable e) {
        continue;
    }
}
```

distantnative: Not the prettiest, but could do the job.

bastianallgeier: Ah, I found it.

distantnative:
Could it be just 

```php
    public function createId(): string
    {
        $length = 8;

        do {
            try {
                $id = Str::random($length);
                if (UserRules::validId($this, $id) === true) {
                    return $id;
                }
            } catch (Throwable $e) {
                $length++;
            }
        } while (true);
    }
 ```
 
 Removing some unnecessary lines.
 
 (last one, promise!)

distantnative (dismissed):
LGTM

distantnative (dismissed):

PR #3874, opened 2021-10-29
Replace md5/crc32 hashes

distantnative:
Fixes https://github.com/getkirby/backlog/issues/26

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1576,7 +1576,7 @@ class App
      */
     public static function versionHash(): string
     {
-        return md5(static::version());
+        return hash('sha256', static::version());
     }
 
     /**
--- a/src/Cms/Media.php
+++ b/src/Cms/Media.php
@@ -154,7 +154,7 @@ class Media
 
         // get both old and new versions (pre and post Kirby 3.4.0)
         $versions = array_merge(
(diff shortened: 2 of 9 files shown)


bastianallgeier:
I don't think we can solve it by just using sha256 everywhere. 

I think the hashes in file urls and paths should still be short. crc32 still seems like a good option. We just need to double-check if the hashed content could be less verbose 

The hased version should be replaced by sha256 for now but also deprecated. We need to replace our own media folder logic for the panel with hashes from vite. 

 The hashed form field signatures should be removed entirely IMHO.

distantnative:
Ok, URLs still looked pretty ok length to me.

lukasbestle:
We could truncate the SHA256 hashes and use e.g. the first 32 or 16 characters. The SHA spec explicitly allows this with the caveat that collision resistance is reduced. But for things that aren't security-critical like the media hashes the truncated SHA256 hash is still much more secure than CRC32.

lukasbestle:
On the other hand changing it would be a breaking change. Given that CRC32 does exactly what we want it to in this use case, do we really need to change it? I wouldn't do it just to make automated tools happy.

bastianallgeier:
I agree that we shouldn't do it for the machines. I think the only thing that's really worth changing/removing is the version hash.

PR #3877, opened 2021-10-30
Add PHP 8.1 support

distantnative:


lukasbestle:
@distantnative Updating PHPUnit to 9.5.10 fixed the fatal issue. Now we can see the actual code issues in the test output. So far `strip_tags()` argument deprecations and a return type compatibility issue.

We can already start fixing these, but I suggest we wait until the PHP 8.1 release so we don't have to do multiple fixing rounds and also to avoid merge conflicts.

PR #3880, opened 2021-10-30
CI: Update tool versions

lukasbestle:
## Describe the PR

I've updated all CI tools to their latest version so that we can run CI with PHP 8.1 in a next step.

Psalm has some new features and improvements that catch more issues, so I fixed them in this PR as well.

## Release notes

None

## Breaking changes

The changes to argument names are technically breaking changes as they will break PHP 8.0 named arguments. But I think the probability that anyone used named arguments for methods that have just one argument is pretty pretty low.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
           extensions: ${{ env.extensions }}
           ini-values: ${{ env.ini }}
           coverage: pcov
-          tools: phpunit:9.5.4, psalm:4.7.1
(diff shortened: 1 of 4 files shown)


distantnative:
Should we check if we can include some of these changes in 3.6.0? As breaking changes (though agreed, mainly on paper than reality)

lukasbestle:
That's exactly why I created a separate PR with milestone 3.6.0. 🤔

bastianallgeier (approved):
LGTM

distantnative (approved):
Seems like I am somewhat confused today 😆 LGTM too

PR #3886, opened 2021-10-31
PHP 7.4 feature: arrow functions

distantnative:
## Describe the PR

- [x] Requires dropped PHP 7.3 support

Some first refactoring to use arrow functions e.g. for `array_map`

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -28,9 +28,7 @@ return [
      * @param string $url Relative or absolute URL
      * @param string|array $options An array of attributes for the link tag or a media attribute string
      */
-    'css' => function (App $kirby, string $url, $options = null): string {
-        return $url;
-    },
+    'css' => fn (App $kirby, string $url, $options = null): string => $url,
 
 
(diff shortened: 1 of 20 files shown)


bastianallgeier (approved):
I went through it and couldn't find any issues, but it would be good if a second pair of eyes could have a look at it. cc @lukasbestle @afbora 

BTW I really love that we can get rid of the use statements with the new syntax.

texnixe:
The only thing I noticed is that some instances of `array_map` were replaced with `A::map()` but not all. Wouldn't it make sense to replace them all for consistency?

bastianallgeier:
@texnixe I guess it was only applied when the A class is already defined. I think it would be good to decide if we want to use it everywhere or if we stick to the native array_map method.

distantnative:
I threw it in where it felt right to me but I didn't go through systematically to replace the native method.

PR #3889, opened 2021-11-01
Dom: Allow relative URLs

### LINKED ISSUE #3754: Writer field: no href attribute rendered when link does not contain protocol

lukasbestle:
## Describe the PR

Truly relative URLs are now supported in `Dom` and therefore also in the `writer` field.

We need to be super careful that URLs that point to external domains are never allowed. This would allow attacks where external resources are embedded or users are tricked into clicking on links to harmful external sites.

As far as I can tell, the new rule for relative URLs ("URL does not contain colon or the part before the colon contains a slash and is therefore not a valid scheme") protects us from that, but I'm not 100 % sure.

It would be super helpful if you could carefully read through and follow the logic of the [specification](https://url.spec.whatwg.org/#url-writing) as well. Four or six eyes see more than two. 🙂

### Fixed regressions

- The `writer` field allows relative URLs again.

## Breaking changes

None

## Related issues/ideas

- Fixes #3754

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Toolkit/Dom.php
+++ b/src/Toolkit/Dom.php
@@ -353,7 +353,7 @@ class Dom
         if (Str::startsWith($url, 'mailto:') === true) {
             $address = Str::after($url, 'mailto:');
 
-            if (empty($address) || V::email($address) === true) {
+            if (empty($address) === true || V::email($address) === true) {
                 return true;
             }
 
@@ -364,7 +364,10 @@ class Dom
         if (Str::startsWith($url, 'tel:') === true) {
             $address = Str::after($url, 'tel:');
 
-            if (empty($address) || preg_match('!^[+]?[0-9]+$!', $address)) {
+            if (
+                empty($address) === true ||
+                preg_match('!^[+]?[0-9]+$!', $address) === 1
+            ) {
                 return true;
             }
 
--- a/src/Toolkit/Dom.php
+++ b/src/Toolkit/Dom.php
@@ -258,8 +258,12 @@ class Dom
         }
 
         // allow URLs that point to fragments inside the file
-        // as well as site-internal URLs
-        if (in_array(mb_substr($url, 0, 1), ['#', '/']) === true) {
+        if (mb_substr($url, 0, 1) === '#') {
+            return true;
+        }
+
+        // allow site-internal URLs (but not protocol-relative URLs)
+        if (mb_substr($url, 0, 1) === '/' && mb_substr($url, 1, 1) !== '/') {
             return true;
         }
 
--- a/src/Toolkit/Dom.php
+++ b/src/Toolkit/Dom.php
@@ -10,6 +10,7 @@ use DOMElement;
 use DOMNode;
 use DOMProcessingInstruction;
 use DOMXPath;
+use Kirby\Cms\App;
 use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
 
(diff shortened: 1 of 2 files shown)


bastianallgeier:
It looks good to me. I wonder if we could get into trouble with upwards navigation via ../../../../something-hacky similar to the nasty filesystem exploits. I'm not sure if this is possible on the URL level at all. The webroot would always be the final point. The only way I could think of would be a subfolder installation where the URL is pointing to another subfolder installation, but that's very theoretical. I just wanted to bring it up as an option.

lukasbestle:
It's a good point! We have the same issue with domain-relative URLs (`/some/path`) already, but that doesn't make it better.

Do we have a way to prevent such attacks?

bastianallgeier:
@lukasbestle we could build a similar check like this: https://github.com/getkirby/kirby/blob/88cbd9be7011df59e7b789f8d5125ecf158bc6ca/src/Filesystem/F.php#L652 if a relative path is still below the index URL.

lukasbestle:
We don't know where the sanitized HTML will be printed, so the number of allowed levels will change depending on the level of the current page.

What we could do:

- Block `../` in URLs completely (at the beginning or in the middle).
- Validate URLs starting with `/` against Kirby's index URL. This would add a dependency on the CMS classes, but we could make it a soft dependency (i.e. the check is only made if an `App` object exists).

bastianallgeier:
Ah, shit. You are right. This seems to be getting a bit out of hand. One other option could be to allow relative paths only in links and block them anywhere else.

lukasbestle:
To be honest that's even more complex to implement as we'd need more configuration options to tell `Dom` which attribute contains links.

Do you see a disadvantage with the changes I proposed?

bastianallgeier:
I think it's an incomplete workaround that still takes quite a lot of effort. I guess it's just a matter of time until someone will afterwards post an issue that ../ in links are not possible.

lukasbestle:
It would be pretty quick to implement as we would only need to check if the URL contains the string `../` or `..\`.

This sequence in the middle of a URL has no actual purpose. At the beginning it *may* be useful sometimes, but I'd much rather have optimal security than a very complex solution that only fixes the issue for URLs that are not links. Also you may want to use relative URLs for image targets etc. as well, which wouldn't be possible if we only allow relative URLs for links.

bastianallgeier:
Ok convinced! Then let's do it this way.

bastianallgeier (approved):
Very nice!!

lukasbestle:
@bastianallgeier Implemented as discussed.

There is one edge case I realized while implementing: With `../` sequences in domain-relative URLs (`/some/../path`) you could "escape" from the site index URL. I added a TODO comment for possible future refactoring, but I think this case is not relevant in practice as there is no need to put `../` sequences in the middle of URLs. In case anyone ever asks for this feature, we can still add it. For now I decided to go the stricter but safer route.

PR #3891, opened 2021-11-02
Fix multilanguage inconsistency

afbora:
## Describe the PR

Now the `Kirby::multilang()` method also checks the `languages` option. What do you think?

### Enhancements

- `Kirby::multilang()` method made more consistent

## Breaking changes

- Multi-language feature cannot be used by adding language manually without activating `languages` option.

## Related issues/ideas

None

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -900,7 +900,10 @@ class App
             return $this->multilang;
         }
 
-        return $this->multilang = $this->languages()->count() !== 0;
+        return $this->multilang = (
+            $this->option('languages') === true &&
+            $this->languages()->count() !== 0
+        );
     }
 
     /**
--- a/tests/Cms/Api/ApiTest.php
+++ b/tests/Cms/Api/ApiTest.php
@@ -104,6 +104,9 @@ class ApiTest extends TestCase
         $this->assertSame('C', setlocale(LC_ALL, 0));
 
         $this->app = $this->app->clone([
+            'options' => [
+                'languages' => true
+            ],
             'languages' => [
                 [
                     'code'    => 'en',
@@ -132,6 +135,9 @@ class ApiTest extends TestCase
         $this->assertSame('C', setlocale(LC_ALL, 0));
 
         $this->app = $this->app->clone([
+            'options' => [
+                'languages'   => true
+            ],
             'languages' => [
                 [
                     'code'    => 'en',
@@ -189,6 +195,7 @@ class ApiTest extends TestCase
                 ]
             ],
             'options' => [
+                'languages'      => true,
                 'panel.language' => 'de'
             ]
         ]);
@@ -204,6 +211,9 @@ class ApiTest extends TestCase
                 [
                     'email' => 'homer@simpsons.com'
                 ]
+            ],
+            'options' => [
+                'languages'   => true,
             ],
             'languages' => [
                 [
@@ -241,6 +251,7 @@ class ApiTest extends TestCase
                 ]
             ],
             'options' => [
+                'languages'      => true,
                 'panel.language' => 'de'
             ]
         ]);
@@ -251,6 +262,9 @@ class ApiTest extends TestCase
 
         // without logged in user without Panel language
         $app = $this->app->clone([
+            'options' => [
+                'languages'   => true
+            ],
             'languages' => [
                 [
                     'code'    => 'it-it',
--- a/tests/Cms/App/AppLanguagesTest.php
+++ b/tests/Cms/App/AppLanguagesTest.php
@@ -9,6 +9,9 @@ class AppLanguagesTest extends TestCase
     public function testLanguages()
     {
         $app = new App([
+            'options' => [
+                'languages' => true
+            ],
             'languages' => [
                 [
                     'code'    => 'en',
@@ -24,12 +27,15 @@ class AppLanguagesTest extends TestCase
 
         $this->assertTrue($app->multilang());
         $this->assertCount(2, $app->languages());
-        $this->assertEquals('en', $app->languageCode());
+        $this->assertSame('en', $app->languageCode());
     }
 
     public function testLanguageCode()
     {
         $app = new App([
+            'options' => [
+                'languages' => true
+            ],
             'languages' => [
                 [
                     'code'    => 'en',
@@ -43,9 +49,9 @@ class AppLanguagesTest extends TestCase
             ]
         ]);
 
-        $this->assertEquals('de', $app->languageCode('de'));
-        $this->assertEquals('en', $app->languageCode('en'));
-        $this->assertEquals('en', $app->languageCode());
-        $this->assertEquals(null, $app->languageCode('fr'));
+        $this->assertSame('de', $app->languageCode('de'));
+        $this->assertSame('en', $app->languageCode('en'));
+        $this->assertSame('en', $app->languageCode());
+        $this->assertSame(null, $app->languageCode('fr'));
     }
 }
--- a/tests/Cms/App/AppTest.php
+++ b/tests/Cms/App/AppTest.php
@@ -967,4 +967,125 @@ class AppTest extends TestCase
 
         $this->assertSame(['title' => 'Site'], $app->controller('none', [], 'json'));
     }
+
+    /**
+     * @covers ::multilang
+     */
+    public function testMultilang()
+    {
+        $app = new App([
+            'roots' => [
+                'index' => '/dev/null'
+            ],
+            'options' => [
+                'languages'   => true
+            ],
+            'languages' => [
+                [
+                    'code'    => 'en',
+                    'name'    => 'English',
+                    'default' => true
+                ],
+                [
+                    'code'    => 'de',
+                    'name'    => 'Deutsch'
+                ]
+            ]
+        ]);
+
+        $this->assertTrue($app->multilang());
+        $this->assertCount(2, $app->languages());
+    }
+
+    /**
+     * @covers ::multilang
+     */
+    public function testMultilangDefault()
+    {
+        $app = new App([
+            'roots' => [
+                'index' => '/dev/null'
+            ]
+        ]);
+
+        $this->assertFalse($app->multilang());
+        $this->assertCount(0, $app->languages());
+    }
+
+    /**
+     * @covers ::multilang
+     */
+    public function testMultilangWithOption()
+    {
+        $app = new App([
+            'roots' => [
+                'index' => '/dev/null'
+            ],
+            'options' => [
+                'languages'   => true
+            ],
+            'languages' => [
+                [
+                    'code'    => 'en',
+                    'name'    => 'English',
+                    'default' => true
+                ],
+                [
+                    'code'    => 'de',
+                    'name'    => 'Deutsch'
+                ]
+            ]
+        ]);
+
+        $this->assertTrue($app->multilang());
+        $this->assertCount(2, $app->languages());
+    }
+
+    /**
+     * @covers ::multilang
+     */
+    public function testMultilangEnabledOptionWithoutLanguages()
+    {
+        $app = new App([
+            'roots' => [
+                'index' => '/dev/null'
+            ],
+            'options' => [
+                'languages' => true
+            ],
+            'languages' => []
+        ]);
+
+        $this->assertFalse($app->multilang());
+        $this->assertCount(0, $app->languages());
+    }
+
+    /**
+     * @covers ::multilang
+     */
+    public function testMultilangDisabledOptionWithLanguages()
+    {
+        $app = new App([
+            'roots' => [
+                'index' => '/dev/null'
+            ],
+            'options' => [
+                'languages'   => false
+            ],
+            'languages' => [
+                [
+                    'code'    => 'en',
+                    'name'    => 'English',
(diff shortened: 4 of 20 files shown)


bastianallgeier:
I wonder if we should actually do this. While it's a lot cleaner indeed, it's quite a massive change so shortly before 3.6. @distantnative @lukasbestle thoughts?

afbora:
It's actually not big. Just 2 lines. Others are just about tests 🙂

bastianallgeier:
That's what I mean! It has a big effect.

lukasbestle:
I'm a bit split on this one to be honest. The change definitely makes a lot of sense, so in the long run we should definitely make it.

I'm just not sure about the breaking change: It's another thing to keep in mind, even for users who don't use the Panel. But on the other hand we already have the Panel breaking change in 3.6, so this PR would make the behavior consistent across the whole CMS.

I think the options are:

1. Merge this PR now for best consistency
2. Merge this PR in 3.7.0 to ease the transition
3. Revert the breaking change in the Panel (so that the `languages` option is not needed in general) and change the behavior both in the Panel and the backend in 3.7.0

IMO option 2 is the worst solution as the breaking change should at least be consistent.

afbora:
While the 2nd option is a breaking change for those who use panels, it is not for those who do not use panels. Options 1 and 3 seem closer to me.

IMHO, On rethink, there are enough breaking changes in 3.6 and we can move it to 3.7. So I suggest option 3.

bastianallgeier:
I'm going to close this. We could pick this up later again.

lukasbestle:
Thinking about this again, I feel like we shouldn't make this change.

**Status quo:**

- The "Languages" area with the configuration UI in the Panel is only displayed if the `languages` option is enabled.
- If languages are already created in `site/languages`, Kirby otherwise behaves exactly the same no matter if the option is enabled or not.
  **Question:** Is this true? While testing I couldn't find other issues, but maybe I missed something.

**Use case 1: Languages created from the Panel:**

- The option becomes a toggle whether the "Languages" view should be displayed or not.
- So it can for example by used to disable the view on a production installation to prevent users from messing with it.

**Use case 2: Manually created languages:**

If the status quo I wrote above is correct, the `languages` option is not needed at all. So having to set it would only be a nuisance for these users.

---

**Conclusion:** We should verify if the multilang setup really works with the option disabled or if there are (edge) cases where Kirby won't behave correctly. If the Panel UI is really the only difference between the option being enabled or disabled, I think we should leave the behavior as it is.

distantnative:
I think the competing interests are the following:
- Setting the `languages` option to `false` should disable multilang mode. No matter if there are files in `site/languages` or not.
- If we would merge the PR as is, installations that have files in `site/languages` and rely on working in multilang mode, but haven't set the `languages` option explicitly, would suddenly turn single language. That's harmful.

So maybe the way forward would be to implement this PR as suggested by @afbora but add here https://github.com/getkirby/kirby/pull/3891/files#diff-3233f117f569d3cc70bcc07dab84edd172cd906b60ef5f57ca385d0f95855752R904 `true` as fallback value for the option, so that if no option has been set, the previous behaviour doesn't change.

afbora:
Users should enable the `languages` option even if the panel is not used. This is definitely consistent behavior.

And I like this idea of @distantnative . In 3.7 we can throw deprecate warning, in 3.8 we remove fallback `true` param.

lukasbestle:
I'm curious: What would the use case be to disable multilang via a config option?

I mean you have your content files with language code extensions. If you disabled multilang for one domain, but not for another, Kirby wouldn't work on one of these domains because of the mismatch between content files and Kirby behavior, no?

distantnative:
@lukasbestle I'm less thinking of the complex use cases than the most common one: Someone with language files that switches their existing `languages` option from `true` to `false` to disable multilang. And then wondering why it's not working. It's weird and hard to understand that a boolean config option just works one way.

lukasbestle:
Yes, that's exactly what I'm also thinking, but in the other direction. If you disable the option, Kirby will try to render a single-lang site, but with multilang content files. That's not going to work without converting the content folder back to single-language files. So IMO the `languages` option should just control the Panel UI, not the behavior of the core.

distantnative:
But the other way around (activating the option, moving from single to multilang) you also have to change content file names by hand, if you aren't using the Panel. So that'd be very consistent.

lukasbestle:
If you do everything by hand (writing the language config files and renaming the content files), you don't need the `languages` option at all. So there's no need for any consistency in this use case.

distantnative:
You might not need it, but we have it and just by name it's rather clear what one would expect setting it to false. And I think it's hard to understand that there is a `languages` option that we support, but only in one direction. I'm not sure I get your case for not supporting `languages: false` fully.

lukasbestle:
To be honest I'd then rather remove the `languages` option entirely in its current form. For manual use cases it's not needed and for Panel use cases the extra Panel view does no harm (and can be disabled with permissions if desired).

texnixe:
I would actually prefer what @lukasbestle suggested in the last post. It's in fact rather annoying that you have to enable multilang in config first when you want to switch. Maybe optionally disable it for any role but admin by default, but permissions actually make sense.

bastianallgeier:
I would also agree with @lukasbestle last comment. With every other view or panel feature that can be switched off, we use permissions. It would make a lot more sense to also do the same here and keep the Kirby::multilang method as it is. 

I'm not sure about limiting it to admins only by default though. It sounds good on first sight but we don't have a clear plan so far for admin-only features. I think we need a clear definition for features that are not available for all users by default first.

distantnative:
@bastianallgeier but managing languages shouldn't be a common-user feature anyways.

Agreed with the last comments. And it probably isn't even a breaking change if we just remove it. As multilang was active anyways if language files are present and turning off wasn't supported either.

lukasbestle:
We already have the `access.languages` permission, shouldn't it cover this?

Regarding admin vs. other roles: I agree with Bastian. So far we don't have any permissions that are admin-only except in user management. I think we should leave the permission as it is.

This leaves only removing support for the `languages` option. In the future Kirby would always behave as if the option was set to `true` right now. That's technically a breaking change for single-language sites: In those the devs need to disable the `access.languages` permission if they don't want users to enable multilang. If we don't want the breaking change, we could deprecate the option in 3.7 and finally remove it in 3.8.

afbora:
I just wonder that; the languages view in the panel depends on the `languages => true` option. If we remove this option, what should the user do to enable the languages for panel ui?

distantnative:
@afbora It would always be enabled for all users that have the specific Panel permission for languages.

distantnative:
@bastianallgeier @lukasbestle while this is kinda true we also have no other permission/part yet where user's would be able to modify the system's config considerably via the Panel. Languages bears that risk, so maybe the default shouldn't be "let everyone do this" - especially as this would be a huge breaking change for current single lang setups.

In fact, the one other similar thing (license) we even hardcore to admins.

texnixe:
The reason I thought it would make sense to enable this permission only for admins was that adding languages usually needs additional work asides from adding languages via the Panel. You typically need to add a language switch, decide which fields should be translatable etc., maybe even filter content on the frontend, probably exactly the reasons why languages were enabled via a config setting?

lukasbestle:
Right, I forgot about the system view. I think the default for the languages permission should be false then. That's also a breaking change, but it's consistent with the system view.

afbora:
Do you mean: We don't remove the `languages` option. We only grant access to administrators by default. Right?

lukasbestle:
No, I meant we remove the option but also change the default for the permission to `false`.

afbora:
~🤔 What if I don't want to show languages for admins as well? I want to make a single-language website to the client.~

I knew that all permissions to administrators are always enabled, but even an administrators can be disabled.

````yaml
title: Admin
permissions:
    access:
      languages: false
````

afbora:
Frankly, I'm a little confused 😕 How to enable, how to disable and how to deny? Can one of you summarize the process?

I should be able to enable or disable multi-language for a website (not for admins). But it didn't make sense to me to do this from the role blueprint and not the config.

lukasbestle:
It's a different approach to the multilang feature:

**Current behavior:**

You as the dev decide that you want the site to have the multilang feature, so you enable the `languages` option. Only then the Panel displays the `languages` area at all.

If you don't want your client to play around with languages on a single-lang site, you just don't enable the option.

But: If you don't even need the Panel, the option has no purpose. And also the same behavior (displaying/hiding the `languages` area) is controlled both by the option and the permission, which is confusing.

**Proposed new behavior:**

The multilang feature is always active (just like every other Panel feature). If you want to use it, you can just visit the `languages` area in the Panel and create a secondary language.

If you want to prevent your client from playing with languages, you can disable the permission or create a non-admin user for the client.

**Summary:**

By getting rid of the `languages` option we could make the multilang feature more approachable and also reduce confusion. At the same time the setup would be less complex and users who don't need the Panel don't need to think about another option.

On the other hand some may need to disable the area for their clients.

The questions are:

- Do clients commonly get an admin account? Or just an editor account where nothing would change?
- Which goal is more important to us? The reduction of general complexity or the "disable the area for single-lang sites" feature?

afbora:
Thank you very much @lukasbestle for the detailed explanation.

To be honest, I think the new behavior you suggested is more approachable but more complex. A website is multilingual or not. It makes perfect sense to manage this from a setting/config. I would prefer a setting that can be set once to be simpler rather than more approachable.

For a single-language site; 
- According to the old behavior, I don't need to do anything. 
- According to the new behavior I need to disable/hide it from roles.

In short, IMHO the behavior you suggested didn't make sense to me.

lukasbestle:
@afbora It's a good point that a site is either multilang or not.

I think we have three options now:

1. **Get rid of the `languages` option**
  Advantages as described above. Disadvantage: Does not fit with the "site is either multilang or not" paradigm.
2. **Enforce the option for the core as well** (this PR)
  TBH I still think we shouldn't go this route. It would be a breaking change for users who didn't use the Panel to set up their languages. There is also no known use case to disable the option again later (which would then break the site because of the incompatible content files with language extensions). I don't see an advantage for our users.
3. **Leave the option as it is**
  This would mean that the behavior fits the "site is either multilang or not" paradigm. At the same time we wouldn't break existing setups.

I think I prefer option 3.

afbora:
If we disagree on the 2. option, I would prefer the 3. option too.

afbora:
To continue, pinging @bastianallgeier @distantnative @texnixe to vote for https://github.com/getkirby/kirby/pull/3891#issuecomment-1141921773 comment.

distantnative:
I feel like number 2 is a bit mischaracterised in Lukas' description. It's the one that would make our behavior consistent and,if we implement it as suggested earlier (with true as a default at the specific pint), it wouldn't be a breaking change actually.

But if that's nothing we can agree on, then we have to leave it as is. Even when easily perceived as broken.

lukasbestle:
To be honest I don't understand the point on consistency. What would be consistent with what?

Independent from that, let's make this a fourth and fifth option for discussion:

4. **Enforce the option in the core, but with a default value of `true`**
  If I understand the suggestion correctly, this would have the following implications for different user groups:
    - On single-lang sites, the Panel would now display the "Languages" area by default (small breaking change). It can be disabled by explicitly setting the option to `false`.
    - For multi-lang sites that already have the `languages` option enabled, nothing would change.
    - On multi-lang sites where the language files were created manually and the option is currently not set at all, the Panel would display the "Languages" area (again a small breaking change). It can only be disabled by blocking the `access.languages` permission for all user roles.
    - Multi-lang sites where the `languages` option was explicitly set to `false` would break completely as the core would now operate in single-lang mode. This is technically a breaking change, but I think this scenario is not common in the real world.
    - Users who want to get started with a new multi-lang site and want to create the language configuration files by hand (or copy them from another project) can do this without having to set the option (as the default is `true`).
5. **Enforce the option in the core with a smart default**
  Smart default = Base the default value on the existence of language configuration in `site/languages`. If something is configured, assume `true`, otherwise assume `false`. Implications:
    - On single-lang sites, nothing would change (the default would still be `false`).
    - For multi-lang sites that already have the `languages` option enabled: Same as above (no change)
    - Multi-lang sites where the language files were created manually and the option is currently not set at all: Same as above ("Languages" area is displayed)
    - Multi-lang sites where the `languages` option was explicitly set to `false`: Same as above (would break completely)
    - New multi-lang sites with manual configuration: Same as above (the smart default would be `true` because of the language config files)

Disadvantages that are shared by both option 4 and 5:

- If you set the `languages` option to `true` so that the Panel displays the "Languages" area, the core would already behave like in a multi-lang installation even though languages are not yet configured. Basically a hen-egg problem. You can't configure languages without the option, but you can't enable the option if you don't have languages configured.
- If you disable the option, the site will break because the content files have language extensions. The same happens in reverse when the option is enabled on a site with single-lang content. It can be confusing when the user expects Kirby to magically convert the content. IMO it can be even more confusing when the "wrong" option value is set by accident (e.g. from a domain-specific config file).

Does this represent the suggestions better? Have I missed something?

afbora:
@getkirby/kirby-staff I think it took too much of our time. I suggest closing PR and running the system as is. At least for a long time 🙂

PR #3895, opened 2021-11-02
3.6.0-rc.3

bastianallgeier:


afbora (approved):
Looks great 🎉

lukasbestle:
I wonder why the Composer package `psr/log` was updated. The new version is only supported in PHP `>=8.0.0`, so we can't ship it with Kirby. Composer shouldn't even have updated it as it conflicts with the version requirement in our `composer.json`.

@bastianallgeier Which Composer version are you using locally?

bastianallgeier:
Ugh, I didn't see that. I'm on Composer 2.0.13

afbora:
@lukasbestle I found why `psr/log` updated to v3, `filp/whoops@2.14.4` : https://github.com/filp/whoops/commit/af5ed434c5fed261aa32c60805c5e7a8fcccb1f4

I guess we need to pin manually v1 from our composer, right? `"psr/log": "1.1.4"`

lukasbestle:
I tested it again, it's not caused by the Composer version but by the current PHP version. Seems like Composer ignores the PHP version constraint and just cares about the current version.

So because Composer was run with PHP 8.0, it installed the latest versions that are compatible with PHP 8.0.

Since we don't rely on `psr/log`, it is thankfully not critical. For now @afbora's suggestion will solve it. But I'll see if the dependency can instead be removed entirely by moving it to `suggest` in Whoops.

Issue #3275, opened 2021-04-29
Huge images are loaded in image/gallery block, thumb/srcset missing


moritzebeling:
**Describe the bug**  
When working with images and the blocks field, the panel quickly gets overloaded and laggy when there are more than a few large images. The problem seems to be, that the image block preview doesn’t use a reduced thumbnail or srcset, but requests the original file.

With about 10 5MB images, the panel will soon break...

**Expected behavior**  
The `image` and `gallery` blocks need to call reduced thumbnails and should use a srcset.
This technique is already implemented in other panel image views: https://github.com/getkirby/kirby/blob/3d580d869e44496c8b145835d4bbbbe95e7559d8/panel/src/components/Misc/Image.vue#L15
The same should be applied here:
- https://github.com/getkirby/kirby/blob/3d580d869e44496c8b145835d4bbbbe95e7559d8/panel/src/components/Blocks/Types/Image.vue#L13
- https://github.com/getkirby/kirby/blob/3d580d869e44496c8b145835d4bbbbe95e7559d8/panel/src/components/Blocks/Types/Gallery.vue#L12

Or maybe a universal `Image` component can help? Then this problem would be solved once and for all times.

**Kirby Version**  
3.5.4

**Additional context**  
This once (https://github.com/getkirby/kirby/issues/2488) was also a problem with the `files` section, but seems to be resolved as of right now.

distantnative:
Unfortunately, I don't think it's that easy. The preview component receives the values for the fields from the drawer - so also for the file field. Which is the standard value for a file field, in that case for a file field in list layout. The only srcset info included there is for the list view preview. 

Rethinking how the previews receive different values, including custom srcset will be not as trivial sadly.

mrflix:
First time using layout & image blocks to let a client build an image grid and unfortunately ran into this problem: the panel page gets very sluggish because of the lack of srcsets. I advice my clients to upload images that are 6000px wide or more so that I can create srcset sizes for up to 5128px width (iMac Retina in fullscreen). Not a good match in this case 😬

I was hyped when I read that srcsets for image blocks get added with 3.6.1. But it turns out that its only for the gallery block and not the image block. The image block is missing this feature!

mrflix:
I think it's a mistake to close this panel-usability issue.

silllli:
[...]

PR #3907, opened 2021-11-03
Use srcset for gallery block preview

rasteiner:
## Describe the PR
Adds the `:srcset` attribute to the `img` tag in Gallery Block preview. 
Otherwise the whole original full resolution image is always used. That can make the panel very unresponsive with large galleries.

## Release notes
 - Use srcset for gallery block preview

## Breaking changes
None

## Ready?

- [ ] CI checks pass

## When merging


afbora:
@distantnative Was it that easy? 😅 Related Issue: https://github.com/getkirby/kirby/issues/3275

rasteiner:
Oh, I hadn't seen the issue. Maybe the srcset that comes in the field data is new in 3.6? 
Or maybe it isn't always there. Don't know, but in all cases, SOMETIMES it is definitely there and I think it might be worth just for those :)

afbora:
This is great! If it is always there we should make the same change for the `Image` block.

rasteiner:
I've only tried with the Gallery block, which by default comes with a "cards" layout files field. Maybe other blocks use a "list" layout and therefore only get a srcset with very small images. I dunno

afbora:
Yes for some reason it didn't work in the Image block 🤷‍♂️

afbora:
Yay! 🎉 I can access to `srcset` from `Image` block, you can implement like that:

````js
if (this.content?.image[0]?.image?.srcset) {
    return this.content.image[0].image.srcset;
}

**Comment by rasteiner** [comment]
Yup, the problem is that (probably this is what @distantnative ment) I get `srcset`s of like "38w" and "76w". Which, especially for the Image Block, is a bit too small. 

For the Image block we'd have to update the `image` function in the `/src/Panel/Model.php` file. 
Like, adding something like this to the function works:

```php
                //preview thumb, reasonable size (used in image and gallery blocks)
                $settings['previewSrc'] = $image->resize(1408)->url();
```

But I don't know if this should be in the same PR. What do you say?

afbora:
Actually, we can solve this issue with little playing `layout: cards`, `size: medium` and `image` props for `Image` block. Seems working..

````yaml
image:
  label: field.blocks.image.name
  type: files
  multiple: false
  layout: cards
  size: medium
  image:
    cover: false
    ratio: 21/9
    back: black
  uploads:
    template: blocks/image
  when:
    location: kirby
````

rasteiner:
I don't know, personally I feel like the block preview shouldn't depend on the layout settings of the files field. 
Relying on setting a *working default* for the default blocks looks a bit frail to me and also like a missed opportunity to fix the thing the correct way. 
Ideally the image size should be determined by the preview. Since the previews only live on the frontend, the panel should either offer an authenticated API that lets us resize images from the frontend, or just always provide a standard set of resized images (so the frontend just chooses which one to use).
That would be better than srcset anyway, since the js panel app can actually choose the thumb version better than a browser. The browser (at least chrome) seems to base its decision solely on viewport width, that means that even if we show a gallery (which offers srcset: 352w, 864w, 1408w) with rendered 120px images, the browser still chooses the 1408w variant on desktop, while the 352w version would be more than enough (sure it's still better than the 5000w raw file uploaded by the client, but it's not ideal). 

That's why this PR just does what it does. It's not ideal, it doesn't fix everything, but it helps and crucially its completely transparent: it doesn't change the behavior of the API or the layout of fields in blocks. I think this is a condition that allows the PR to be squished into 3.6.0, but leaves enough room to make a thought out API change for 3.6.1. 

Not sure if I explained myself clearly :/

rasteiner:
related: https://github.com/rasteiner/k3-reasonable-gallery-preview

distantnative (approved):
I think it's a hack and not a good "real" solution. But such a thought-through one will need more time. And the proposed hack is so minimal and looks to me quite harmless, which is why I'd vote for merging.

bastianallgeier:
@distantnative do you mean you vote for merging with @afbora's suggestion or as it is?

distantnative:
Without the blueprint suggestion, I think that gets into the more complex solution already. But I would support the temporary hack of just adding `:srcset="image.image.srcset"` to the gallery block - as this PR does.

silllli:
I just encountered this problem, a very unresponsive, slow loading panel. The problem in my case is not the amount, but the image resolution. The added `srcset` doesn’t really help since there is no `sizes` attribute provided. Without a `sizes` attribute the browser assumes the `size` to be `100vw` which in this case (my display resolution) means it will download the full version as the next matching size.

PR #3908, opened 2021-11-03
Use optional chaining `(?.)` in frontend

afbora:
## Describe the PR

Use optional chaining `(?.)`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

All browsers supported
https://caniuse.com/mdn-javascript_operators_optional_chaining

### Refactoring
- Switched to `Optional chaining operator` on Frontend

## Breaking changes

None

## Related issues/ideas

None

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/panel/src/api/request.js
+++ b/panel/src/api/request.js
@@ -65,13 +65,13 @@ export default (config) => {
         }
 
         // look for an error status
-        if (json.status && json.status === "error") {
+        if (json.status === "error") {
           throw json;
         }
 
         let data = json;
 
-        if (json.data && json.type && json.type === "model") {
+        if (json.data && json.type === "model") {
           data = json.data;
         }
 
--- a/panel/src/components/Dialogs/Dialog.vue
+++ b/panel/src/components/Dialogs/Dialog.vue
@@ -214,10 +214,10 @@ export default {
       this.close();
     },
     focus() {
-      if (this.$refs.dialog && this.$refs.dialog.querySelector) {
+      if (this.$refs.dialog?.querySelector) {
         const btn = this.$refs.dialog.querySelector(".k-dialog-button-cancel");
 
-        if (btn && typeof btn.focus === "function") {
+        if (typeof btn?.focus === "function") {
           btn.focus();
         }
       }
--- a/panel/src/components/Dialogs/LanguageDialog.vue
+++ b/panel/src/components/Dialogs/LanguageDialog.vue
@@ -38,7 +38,7 @@ export default {
           // if the entered language code exists
           // matches the locale values in the languages defined in the system
           let locales = this.$system.locales || [];
-          if (locales && locales[code]) {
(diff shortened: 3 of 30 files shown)


johannschopplich:
I love minor refactors like this. Beautiful. 🤭☺️

bastianallgeier:
I like it a lot. We should just make sure to test this properly to not overlook any unwanted side-effects. I can't see any from scanning the files, but we should still check.

gearsdigital:
Not sure if you have some static code analysis somewhere but you should consider adding it because it can help to prevent unwanted side-effects.

https://eslint.org/docs/rules/no-unsafe-optional-chaining

*edit You have! And that's awesome 😎

distantnative:
Don't wanna ruin the party, but technically it would be a breaking change as it raises the browser requirements of our Panel.

Current requirements:
- Chrome 63+
- Firefox 67+
- Safari 11.1+
- Edge 79+
- Opera 50+
- Mobile Safari 11.1+
- Android Browser 92+
- Chrome for Android 92+

With this change:
- Chrome 80+
- Firefox 74+
- Safari 13.1+
- Edge 80+
- Opera 67+
- Mobile Safari 13.4+
- Android Browser 95+
- Chrome for Android 95+

The new requirements would cover 91.81% of all users according to https://caniuse.com/mdn-javascript_operators_optional_chaining. Our current requirements 94.24%.

distantnative (changes requested):
Don't wanna ruin the party, but technically it would be a breaking change as it raises the browser requirements of our Panel.

Current requirements:
- Chrome 63+
- Firefox 67+
- Safari 11.1+
- Edge 79+
- Opera 50+
- Mobile Safari 11.1+
- Android Browser 92+
- Chrome for Android 92+

With this change:
- Chrome 80+
- Firefox 74+
- Safari 13.1+
- Edge 80+
- Opera 67+
- Mobile Safari 13.4+
- Android Browser 95+
- Chrome for Android 95+

The new requirements would cover 91.81% of all users according to https://caniuse.com/mdn-javascript_operators_optional_chaining. Our current requirements 94.24%.

It's no big break at all, but I would still argue that users shouldn't be forced to communicate new browser requirements to clients on minor updates. That's why I think we would need to move this to 3.7

johannschopplich:
@distantnative Are you sure? I don't trink so. Since the `target` for Vite stays untouched, all optional chaining occurences will get transpiled by esbuild.

This can easily get shipped in 3.6.1. You won't break the browser requirements. They are defined by [Vite's default browser target](https://vitejs.dev/guide/build.html#browser-compatibility).

distantnative:
Got you, sorry somehow thoughts this needs to be supported natively.

distantnative (approved):
Given that, the changes itself look good to me.

PR #3909, opened 2021-11-04
Refactor/fiber error handling

bastianallgeier:
## Describe the PR

Error handling was very chaotic across Fiber requests. There wasn't a unified way to deal with JSON parsing errors, particular errors in the request response and invalid response objects. I've cleaned it all up and also refactored our Fatal dialog. It's now a lot more usable and works especially well when you want to quickly debug something with dump() in Fiber PHP code. 

You can do 

```php
dump($something)
exit();
```

Pretty much anywhere now and the Panel will show the Fatal dialog with the output. This makes it even easier to work with those request. 

I tried to be very careful with the changes and split everything in logical commits, but we need to double-check all of it again. 

### Refactored
- Fiber handling and dialogs for Fiber

### Enhancements
- New CSS properties
  - `--color-dark: #313740;`
  - `--z-fatal`

### DIFF
--- a/panel/src/api/request.js
+++ b/panel/src/api/request.js
@@ -7,6 +7,7 @@ export async function toJson(response) {
     data = JSON.parse(text);
   } catch (e) {
     window.panel.$vue.$api.onParserError(text);
+    return false;
   }
 
   return data;
--- a/panel/src/api/request.js
+++ b/panel/src/api/request.js
@@ -6,7 +6,7 @@ export async function toJson(response) {
   try {
     data = JSON.parse(text);
   } catch (e) {
-    window.panel.$vue.$api.onParserError(text);
+    window.panel.$vue.$api.onParserError({ html: text });
     return false;
   }
 
--- a/panel/src/components/Dialogs/FiberDialog.vue
+++ b/panel/src/components/Dialogs/FiberDialog.vue
@@ -21,24 +21,19 @@ export default {
     async onSubmit(value) {
 
       try {
-        const response = await this.$request(this.path, {
+        const dialog = await this.$request(this.path, {
           body: value,
           method: "POST",
+          type: "$dialog",
           headers: {
             "X-Fiber-Referrer": this.referrer
           }
         });
 
-        const dialog = response.$dialog;
-
-        // something's wrong with the response
-        if (!dialog) {
-          throw `The dialog could not be submitted`;
-        }
-
-        // there's an error message from the server
-        if (dialog.error) {
-          throw dialog.error;
+        // json parsing failed and
+        // the fatal dialog is taking over
+        if (dialog === false) {
+          return false;
         }
 
         // everything went fine. We can close the dialog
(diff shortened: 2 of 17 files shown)


lukasbestle (commented):
I can't really review the code itself as I don't understand the inner workings of Fiber enough yet, but I like the approach to streamline the error handling into one output channel.

One question: What happens if debug mode is disabled? If I understand it correctly, `debug` only controls whether the error is printed to the console. I'm not sure if it should be the other way around: If debugging, the error is displayed very explicitly in the dialog, but otherwise it's just printed to the console and the user-facing error is something more user-readable.

bastianallgeier:
The error notification dialog prints the error message of the thrown exception. Those should always be as human readable as possible. The entire idea of the dialog is to give a useful information about the error to "regular" folks. When debug mode is activated, the console will have additional information about the issue for developers. 

This PR is really just a first step to improve the overall architecture for error handling in the frontend. Ideally we should start working with custom error classes in JS, very much like we do on the server. I haven't touched this yet, because it's a lot of work and we need to make a good concept for it. 

But once this would be established our notification/error action could become a lot smarter and decide based on the type of error if the dialog should be shown, what the content of the dialog should be, styling etc. We can create a lot more useful errors for non-developers that way. 

But this PR is already a very good step in that direction.

bastianallgeier:
To give a bit more background about Fiber: 

There are three different types of errors that need to be handled when a Fiber request is made: 

## 1. Broken response

The server is sending back something completely unreadable (JSON parsing fails) This should always lead to the Fatal dialog, which shows the server response in a separate iframe. This didn't work properly before this PR and is a big deal, because when you dump something in your PHP code or you make a mistake that leads to a syntax error, it's really hard to debug that if the Fatal dialog with the iframe does not work. The only way is to go through the network tab in the console and while this works, it's really not very developer-friendly. With the iframe, you can really use dump in the same way you would use it in simple PHP templates 

## 2. Unexpected response

The server is sending back a proper response, but it does not match what is expected. We have four types of Fiber responses: `view`, `dialog`, `dropdown`, `search` (more might follow) 

A Fiber response object always looks like this: 

```json
"$dialog": {
  "code": 200,
  "component": "k-form-dialog",
  "props": { }  
}
```

This way, we can make sure that the response will be compatible with the sent request. If the request expects a dialog response and a view will be returned, we can easily stop at that point and throw an error

## 3. PHP Exception response

Whenever an exception is thrown in PHP, the response will include the error message and status code for the exception and the `Fiber.request()` method can then throw a matching error. 

```json
"$dialog": {
  "code": 404,
  "error": "The page could not be found"
}
```

bastianallgeier:
OK, I'm making a decision and will merge this. The worst that could happen is that we have small bugs connected to error handling. I'm not worried about this at all. 

On the positive side, this PR drastically improves the architecture and also fixes the fatal dialog, which is really needed IMHO.

distantnative (approved):
Looks good to me!

Only thought, but a bit unrelated: since we're adding `--color-dark`, maybe we want to use that for `k-file-preview` as background?

PR #3910, opened 2021-11-04
Only show plugins section if there are any plugins installed

bastianallgeier:
Title says it all. It's a pretty simple fix.


distantnative:
Why not using the `link` prop of `k-headline`?

bastianallgeier:
@distantnative our link component only handles external links correctly when the target is set. That's actually shit and we need to fix it. I wonder if I should do it in this PR.

bastianallgeier:
@distantnative ok, I fixed it :)

PR #3912, opened 2021-11-05
Pin psr/log to v1.1.4

lukasbestle:
## Describe the PR

Ensures compatibility with PHP 7.4.

Kirby does not actually need `psr/log`, but because an interface is referenced from the Whoops code, we need to ship with it to avoid type errors. My plan to remove the dependency from `filp/whoops` entirely didn't work out as Whoops really needs the interface to conform with the PSR-3 logging standard.

I wonder if there's a way to mark the `psr/log` dependency as "remove when support for PHP 7 is dropped". Unfortunately JSON doesn't support comments. :(

## Release notes

None

## Breaking changes

None

## Related issues/ideas

- Fixes getkirby/backlog#35


bastianallgeier (approved):

afbora:
@lukasbestle I guess we can use `_comment` property...

PR #3934, opened 2021-11-16
3.6.0

bastianallgeier:


views/php.php
```diff
@@ -5,7 +5,7 @@
   </p>
   <p class="admin-advice">
     Advice for developers and administrators:<br>
-    Change the PHP version to 7.4 or 8.0 (PHP 8.0 is recommended)
+    Upgrade your PHP version
```
lukasbestle: Not sure about this IMO. The view will also be displayed if the PHP version is too high (e.g. 8.1). This message won't fit then.

Suggestion:

```
Change the PHP version to one supported by your version of Kirby
```

lukasbestle (changes requested):

lukasbestle (approved):

PR #3939, opened 2021-11-17
Don't throw pages() behavior deprecation warning if single item is an array

### LINKED ISSUE #3938: Single-item pages() deprecation warning should be silenced when explicitly passing an array
**Describe the bug**  
When passing an array with a single item to the `pages()` helper, the should deprecation warning be silenced.

**To Reproduce**  

```php
$pages = pages(['example-page']);
```
returns a Pages collection on Kirby 3.5.x, there should be no need to warn about behavior change in this situation

**Expected behavior**  
Throw no exception.

**Screenshots**  

Screen Shot 2021-11-16 at 3 56 47 PM

**Kirby Version**  
`3.6.0`

neildaniels:
## Describe the PR
This changes the deprecation warning to not be thrown if the single item passed to the `pages()` is an array.

### Fixed
- `pages()` helper does not throw deprecation warning for arrays as single parameter

## Breaking changes

None

## Related issues/ideas

- Fixes #3938 

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -584,7 +584,7 @@ function page(...$id)
  */
 function pages(...$id)
 {
-    if (count($id) === 1) {
+    if (count($id) === 1 && is_array($id[0]) === false) {
         // @codeCoverageIgnoreStart
         deprecated('Passing a single id to the `pages()` helper will return a Kirby\Cms\Pages collection with a single element instead of the single Kirby\Cms\Page object itself - starting in 3.7.0.');
         // @codeCoverageIgnoreEnd


distantnative:
Wouldn't an array with one ID still only return a `Kirby\Cms\Page` object, no `Kirby\Cms\Pages` collection?

bastianallgeier:
An array with one ID would still return a collection with a single element, but I think the expectations are very different when you pass an array. 

I.e. 

```php 
pages('something')
```
Even though I don't really understand why we ever decided to implement it that way in previous versions, I can still see at least some sense in the idea that only one page is returned in this case. It makes sense to send the deprecation warning there.

But in this case …

```php 
pages(['something'])
```
… I think the expectations were already clear and you already expected a collection in previous versions. That's why I think it makes sense to not throw a deprecation warning here.

distantnative:
Agreed with not throwing an exception. I just had the impression while looking at the code that even an array with one element would be unwrapped and end up as `Kirby\Cms\Page` not as `Kirby\Cms\Pages` with one element.

bastianallgeier:
Oh that would be not good indeed

distantnative (approved):
Nope, tested. it does correctly return a `Pages` collection with one element.

PR #3945, opened 2021-11-18
Add a static cache for resolved queries of blueprint fields with options

### LINKED ISSUE #3746: Form Class is loading queries (too?) eagerly causing (unnecessary) performance bottleneck
**Describe the bug**  
Form class loads data even if it may be never used. it dataset is very big this causes a hugh delay for on each request. This is a common scenario in both autoid and boost plugin and currently make adding pages beyond 20k index impossible within 30sec timeout.

https://github.com/getkirby/kirby/blob/efe24a6ef4ad61010428b1cb725f8f97a633ff82/src/Cms/PageActions.php#L459

**To Reproduce**  
1) create a lot of pages of template `human`

2) create a collection that does return a site->index()
```php
<?php

return function () {
    return site()->index()->filterBy('template', 'human');
};
```

3) create a page blueprint with a field with a multiselect query
```
# other fields
human:
  type: multiselect
  options: query
  max: 1
  query:
    fetch: kirby.collection('humans') # this does a site index -> filter
    text: "{{ page.title }}"
    value: "{{ page.id }}"
```

4) create a single or a batch of new `human` page programmatically.
5) the query will be executed in php code even it is most likely only used in api/kql/panel.

**Expected behavior**  
the query should only be executed in context of api/kql/panel or when called explicitly in php.

**Kirby Version**  
3.5.7 and 3.6.0-beta.1

bnomei:
This does greatly improve performance when options are fetched. It's mainly needed inside the panel to speed up complex many-to-many relations of pages.

closes #3746

## Describe the PR
I added a simple static cache to keep resolved options in memory and prevent them from being resolved again.  

This does improve performance tested on a rather complex blueprint setup. Together with @waffl from the forum I have used xdebug to profile a 6000 pages setup with multiple and different many-to-many relations between pages, sometimes even in structures.

The cache is helpful since Kirby will resolve the query foreach loaded page with the same blueprint again and again. Once per field. This means once per row in a structure. We have seen between 400.000 and 2.000.000 requests to query->resolve on a base of 20 different fetch statements. A lot of these resolves are triggered by Kirby resolving the blueprints of pages related to the query, like siblings.

- The loading time of the a content in the panel with the static cache in place was improved from 14.000 millisec to 400 millisec.
- The loading time of the same content in frontend is around 120 millisec with and without the cache. That is because the blueprint does not need to be resolved here.

FORUM:
https://forum.getkirby.com/t/help-needed-troubleshooting-slow-panel-query-multiselect-field/23685/9

RELATED ISSUES:
https://github.com/getkirby/kirby/issues/3746

### DIFF
--- a/src/Form/OptionsQuery.php
+++ b/src/Form/OptionsQuery.php
@@ -141,13 +141,13 @@ class OptionsQuery
 
         preg_match('/(\bblock|file|page|user)\.+/i', $cacheKey, $matches);
         foreach($matches as $match) {
-            if ($model = A::get(strtolower($match[1]), $data)) {
+            if ($model = A::get($data, strtolower($match[1]))) {
                 $cacheKey .= '[' . $model->id() . ']';
             }
         }
 
         // if a cached collection exist return that
-        if ($cachedOptions = A::get($cacheKey, static::$cache)) {
+        if ($cachedOptions = A::get(static::$cache, $cacheKey)) {
             return $cachedOptions;
         }
         
--- a/src/Form/OptionsQuery.php
+++ b/src/Form/OptionsQuery.php
@@ -134,9 +134,9 @@ class OptionsQuery
         // "{{ kirby.collection('customCollection') }}"
         // "{{ site.customMethod }}"
 
-        preg_match('/(?:\bblock|file|page|user)\.+/i', $cacheKey, $matches);
+        preg_match('/(\bblock|file|page|user)\.+/i', $cacheKey, $matches);
         foreach($matches as $match) {
-            if ($model = A::get($match[1], $data)) {
+            if ($model = A::get(strtolower($match[1]), $data)) {
                 $cacheKey .= '[' . $model->id() . ']';
             }
         }
--- a/src/Form/OptionsQuery.php
+++ b/src/Form/OptionsQuery.php
@@ -119,12 +119,34 @@ class OptionsQuery
             return $this->options;
         }
 
-        if ($cachedOptions = A::get($this->query(), static::$cache)) {
-            return $cachedOptions;
+        $data    = $this->data();
+        $cacheKey = $this->query();
+
+        // If the query string contains any model with an id
+        // then append that id to $cacheKey to make it unique.
+        //
+        // examples: 
+        // "{{ file.parent.images }}"
+        // "{{ page.parent.files }}"
+        // "{{ user.customMethod }}"
+        // 
+        // this is not needed for queries like:
+        // "{{ kirby.collection('customCollection') }}"
+        // "{{ site.customMethod }}"
+
+        preg_match('/(?:\bblock|file|page|user)\.+/i', $cacheKey, $matches);
+        foreach($matches as $match) {
+            if ($model = A::get($match[1], $data)) {
+                $cacheKey .= '[' . $model->id() . ']';
+            }
         }
 
-        $data    = $this->data();
-        $query   = new Query($this->query(), $data);
+        // if a cached collection exist return that
+        if ($cachedOptions = A::get($cacheKey, static::$cache)) {
+            return $cachedOptions;
+        }
+        
+        $query   = new Query($cacheKey, $data);
         $result  = $query->result();
         $result  = $this->resultToCollection($result);
         $options = [];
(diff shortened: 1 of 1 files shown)


bnomei:
Instead of the automatic detection via the model it might prove simpler to let the user define the cache id in the blueprint and only then do caching.

```yml
fetch:
  query: "{{ page.customMethod }}"
  cache: "{{ page.id }}.customMethod"
```

distantnative:
In general, I think caching options could be very beneficial indeed. The query syntax is very powerful, I am a bit afraid that we overlook something crucial very easily. I am not very fond of pushing it on the user to define a cache ID. At the same time, automatically trying to generate one looks like it could be fragile.

More in general I am asking myself, if this should be solved at the level of OptionsQuery or rather more abstract for all options, also coming from an API etc. - but what if any of these (API, query...) involves a changing aspect (random or other)?

rasteiner:
can't caching happen in userland? 

like, if the query is:
```yaml
fetch:
  query: "{{ page.customMethod }}"
```
I think `pagemodel::customMethod()` should be able to do the caching and therefore also decide what kind of cache: memory for a single request, apcu, files, whatever... Maybe kirby could provide some utilities to make caching easier, instead of integrating one into `OptionsQuery`. Maybe even something nutty that allows an API like `page.memoize.customMethod` - actually that might also be done in userland like a plugin...

distantnative:
@rasteiner I think @bnomei's intention are not just custom methods but similarly queries like `{{ kirby.collection('customCollection') }}` or even `{{ page.children.listed.filterBy('a', 'b').sortBy(...) }}` on a large number of pages.

rasteiner:
> `{{ page.children.listed.filterBy('a', 'b').sortBy(...) }}`

```js
page.memoize.children.listed.filterBy('a', 'b').sortBy(...)
```
:)

when I get some time I'll do a plugin to show what I mean. 
An advantage would also be that the same could then be used in templates:
```php
$page->memoize()->children()->listed()->filterBy('a', 'b')->sortBy(...);
```


The only downside I see is loosing some "ease of use" (not all queries are always cached, but you need to enable it), and that the query parsing itself isn't cached. But I think (haven't measured) that query parse time isn't the problem - and if it is, there are ways to optimize that without the "risks" involved in data caching.

PS: don't get me wrong. I'm **all in** for caching too, for example the fact that section data doesn't come with the initial fiber request (for performance reasons), is blocking a plugin idea I have... I'm just trying to contribute to the discourse by brainstorming a bit, in the hopes to get something going :)

bnomei:
if you do not cache the result and thus avoid the repeated resolution of the query string you end up with n-time more calls. like in my example above up to millions instead of a few hundred. this issue only appears if within the data of the query further or the same query has to be resolved. any resolution of "relations" like with autoid/boost plugin might trigger this.

bnomei:
@rasteiner the collection used when tracking this issue already returned a static cached collection of an array of primitive values. we even tried with a plain array but the issue is really with the repeated resolution of the querystring.

rasteiner:
> @rasteiner the collection used when tracking this issue already returned a static cached collection of an array of primitive values. we even tried with a plain array but the issue is really with the repeated resolution of the querystring.

do you have a test setup laying around? I'd like to play with it :)
Otherwise I'll create one myself

bnomei:
@rasteiner sadly i can not share that setup. but it would be great if you created one. post it here and i will be able to compare and give feedback.

- template: relation-type-a (autoid and title)
- template: relation-type-b
- ...
- template: relation-type-n
- template: page-a, has an autoid, has a structure that has fields (autoid ref multiselect, filtered by page-a relation types)
- template: page-b, had an autoid, has structure but different set of types
- static collection: for all pages (autoid and title as array)
- static collections: one each for filtered relations-types for page-a and page-b

rasteiner:
As a basis for tests: https://github.com/rasteiner/kirby-query-stress

have not added any solutions yet.

distantnative:
@bnomei @rasteiner We are currently revamping our blueprint classes, so also the options classes. I don't think the PR will be applicable afterwards. Revisiting the topic, I also think that @rasteiner is quite right that this is nothing the OptionsQuery class should handle but either the called methods itself or some caching instance one splices into the query.

lukasbestle:
I agree with Nico. Especially the automatic cache key generation could make this quite complex and error-prone.

bnomei:
@distantnative i found an easier solution in the meantime. one can use programmable blueprints (even for like single multiselect fields) and add static and regular cache via php. personally i ended up doing that for all but the most simple queries.

https://getkirby.com/docs/cookbook/panel/programmable-blueprints

PR #3950, opened 2021-11-21
Call `route` hooks only for core router

distantnative:
## Describe the PR

In 3.6.0 we introduced a new `router()` helper and use a separate `Router` instance for our Panel routes. However, the `route:before` and `route:after` hooks get called for any call of any `Router` object - instead of only our core Kirby router, which I think is the expected behaviour.

### Fixes
- `route:before` and `route:after` hooks only get called for core routing calls

## Related

- Fixes https://github.com/getkirby/kirby/issues/3951

## Breaking changes

Previous static `Http\Router::$beforeEach` and `Http\Router::$afterEach` closures cannot be set statically anymore but need to be passed to `Router::__construct()` as 2nd and 3rd parameters.

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -328,17 +328,9 @@ class App
      */
     public function call(string $path = null, string $method = null)
     {
-        $router = $this->router();
-
-        $router::$beforeEach = function ($route, $path, $method) {
-            $this->trigger('route:before', compact('route', 'path', 'method'));
-        };
-
-        $router::$afterEach = function ($route, $path, $method, $result, $final) {
-            return $this->apply('route:after', compact('route', 'path', 'method', 'result', 'final'), 'result');
-        };
-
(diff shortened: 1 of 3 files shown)


bastianallgeier:
Couldn't we keep the static methods and set them differently in the Fiber context?

distantnative:
But it's not only the Fiber context. If we set them statically, any use of `router()` would still trigger them, which I think would be very unexpected.

bastianallgeier:
Damn, you are right. The only hack right now we be to reset them when route() is called and restore them directly afterwards. I just wonder if we could have a less aggressive fix for now.

distantnative:
I mean so far we set them each time in `App::call()` which is pretty insufficient in its own. But a minimal fix would be to under them again right after

distantnative:
Closing in favor of https://github.com/getkirby/kirby/pull/3958

PR #3959, opened 2021-11-23
Support i18n keys for form options

### LINKED ISSUE #3955: Translation failed on select fields
External translation not work on select fields:
```
foo:
  label: path.to.translation.label
  type: select
  options:
    option1: path.to.translation.option1
    option2: path.to.translation.option2
```
The panel shows the path in the dropdown.

distantnative:
## Describe the PR

Allows to use i18n keys as options

```yaml
mySelectField:
  label: myfield.label
  type: select
  options:
    option1: myfield.option1
    option2: myfield.option2

```

### Enhancements
- Support for translation keys as option text

## Breaking changes

⚠️ Option text that is the same as an existing i18n key now gets replaced by the translation string.

## Related issues/ideas

- Fixes #3955

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Form/Options.php
+++ b/src/Form/Options.php
@@ -137,7 +137,10 @@ class Options
         $result = [];
 
         foreach ($options as $key => $option) {
-            if (is_array($option) === false || isset($option['value']) === false) {
+            if (
+                is_array($option) === false ||
+                isset($option['value']) === false
+            ) {
                 $option = [
                     'value' => is_int($key) ? $option : $key,
                     'text'  => $option
@@ -145,9 +148,7 @@ class Options
             }
 
             // translate the option text
-            if (is_array($option['text']) === true) {
-                $option['text'] = I18n::translate($option['text'], $option['text']);
-            }
+            $option['text'] = I18n::translate($option['text'], $option['text']);
 
             // add the option to the list
             $result[] = $option;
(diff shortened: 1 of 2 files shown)


bastianallgeier:
Can we risk that in 3.6.1? 

To expand that a bit more. There are quite a few generic English translation keys that could cause unwanted side effects. I don't think that you would ever name something error.page.changeSlug.permission accidentally, but a value called "copy" or "date" or a few more could really cause weird confusing results.

distantnative:
No I don't think we should do it in 3.6.1 due to the potential breaking change. But I only realized when having had this finished.

Maybe we can already make a decision if we would want to support this or if this would always be too weird and unforeseen. Merging rather in 3.7 (so after the discussion I would take it down on hold until then).

bastianallgeier:
Agreed! We could also think about using the query syntax instead to solve this. This reminds me that we should introduce certain helpers into the query syntax at some point. I'm already doing that in KQL. There you can do `{{ page('something).title }}` for example. In this case we could have `{{ t('some.key') }}`

distantnative:
Indeed, I like `"{{ t('some.key') }}"`. We should convert existing places to that syntax as well. e.g. I think (not sure) that `label` currently allows direct i18n keys. Would be better to move these to explicit syntax as well. That would be clear 3.7 then. If we agree, I'll create a backlog issue with the plan (don't think it needs a separate Notion page) and plan it for 3.7.

bastianallgeier:
Yes, agreed! Let's move it to 3.7

PR #3963, opened 2021-11-24
Fix document root of router for php built-in webserver

thewizardofbureaucracy:
## Describe the PR

Currently the router.php has a hardcoded document root, namely the parent directory of the kirby directory. This PR changes that and instead uses `$_SERVER["DOCUMENT_ROOT"]`, which is either the directory the built-in webserver is started from or the directory manually set with the `-t <docroot>` parameter.

### Fixes

- `router.php` now allows the document root to be set manually

## Breaking changes

`php -S localhost:8000 kirby/router.php` as described in multiple locations of the docs will now have to be run from the directory containing the index.php or alternatively you'll have to set the `-t <docroot>` parameter to wherever the index.php is, i. e. if you use a public directory, that one. It probably isn't really "breaking", but you won't be able to run the command with router.php as target from anywhere anymore, because the root isn't hardcoded anymore.

## Related issues/ideas

n/a

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/router.php
+++ b/router.php
@@ -1,12 +1,14 @@
 <?php
 
-$root = dirname(__DIR__);
+$uri = urldecode(
+    parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH)
+);
 
-// https://yourdomain.com/media/super/nice.jpg
-if (file_exists($root . '/' . $_SERVER['SCRIPT_NAME'])) {
-    return false; // serve the requested resource as-is.
+// Emulate Apache's `mod_rewrite` functionality
+if ($uri !== '/' && file_exists($_SERVER['DOCUMENT_ROOT'] . '/' . $uri)) {
+    return false;
 }
 
-$_SERVER['SCRIPT_NAME'] = str_replace($_SERVER['DOCUMENT_ROOT'], '', $index = $root . '/index.php');
+$_SERVER['SCRIPT_NAME'] = '/index.php';
 
-include $index;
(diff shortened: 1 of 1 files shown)


johannschopplich:
Nice one! Would be great to have this in the core. At the moment, I'm always putting a `server.php` into my projects.

Suggestion to use `$_SERVER['REQUEST_URI']`:

```php
<?php

$uri = urldecode(
    parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH)
);

// Emulate Apache's `mod_rewrite` functionality
if ($uri !== '/' && file_exists($_SERVER['DOCUMENT_ROOT'] . '/' . $uri)) {
    return false;
}

$_SERVER['SCRIPT_NAME'] = '/index.php';

include $_SERVER['DOCUMENT_ROOT'] . '/' . $_SERVER['SCRIPT_NAME'];

```

thewizardofbureaucracy:
@johannschopplich Hi! Thanks for the suggestion!

I am not entirely sure whether or not there is any meaningful difference for this particular use case (as in the PHP built-in webserver), because there is no redirection akin to .htaccess rewrite rules going on, so the `PHP_URL_PATH` in `$_SERVER['REQUEST_URI']` and `$_SERVER['SCRIPT_NAME']` should be the same.

I'll still change it. A few sites including the Google Cloud docs seem to recommend this or a similar approach, so there might be something to it.

bastianallgeier (approved):
Works like a charm 🙌

PR #3969, opened 2021-11-25
`Str` methods: handle empty `$needle`

### LINKED ISSUE #3459: Str::contains() does not accept an empty needle
Possibly more of an improvement than a bug, but it would make sense in my eyes. If you use the `Str::contains()` method to query a field and it is empty, you get a PHP error with `strstr(): Empty needle`.

We use for example quite often:

```php
if ( Str::contains('Buzzword',$page->myfield()) ) {
  echo 'found';
}
```

But since `$page->myfield()` is for example a checkbox field it can also be empty, so the method fails. Therefore it would make sense if it would also work with an empty string.

distantnative:
## Describe the PR

Aligns `Str` methods in their handling of an empty `$needle` string.

### Fixed
- `Str` methods handle an empty `$needle` string

## Breaking changes

Instead of PHP errors in PHP 7.4, most methods now throw a proper `Kirby\Exception\InvalidArgumentException`.
`Str::contains()` returns `true` when empty `$needle` is passed (instead PHP error).

## Related issues/ideas

- Fixes #3459

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -3,6 +3,7 @@
 namespace Kirby\Toolkit;
 
 use Exception;
+use Kirby\Exception\InvalidArgumentException;
 
 /**
  * The String class provides a set
@@ -180,9 +181,9 @@ class Str
 
         if ($position === false) {
             return '';
-        } else {
-            return static::substr($string, $position + static::length($needle));
         }
+
+        return static::substr($string, $position + static::length($needle));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


bastianallgeier (commented):
Can't we save a lot of those Exceptions in cases where Str::position is used anyway?

distantnative:
Oh of course, pushed an updated commit. Left the unit tests for all methods in there still.

bastianallgeier (approved):

PR #3979, opened 2021-11-30
Add PHP 8.1 support

distantnative:
## Describe the PR

Adds support for PHP 8.1

### Dependencies
- [x] ~~Punycode: https://github.com/true/php-punycode/issues/35 (use alternative)~~
- [x] ~~SimpleImage: https://github.com/claviska/SimpleImage/issues/287~~

### Splitting into separate PRs (⚠️ merge these first)
- [x] https://github.com/getkirby/kirby/pull/3988
- [x] https://github.com/getkirby/kirby/pull/3989
- [x] https://github.com/getkirby/kirby/pull/3990
- [x] https://github.com/getkirby/kirby/pull/3991
- [x] https://github.com/getkirby/kirby/pull/3992

### TODO
- [x] ~~Double check breaking changes https://www.php.net/manual/de/migration81.incompatible.php~~
  - [x] ~~htmlspecialchars(), htmlentities(), htmlspecialchars_decode(), html_entity_decode(), and get_html_translation_table() now use ENT_QUOTES | ENT_SUBSTITUTE rather than ENT_COMPAT by default.~~
- [x] ~~Double check deprecations https://www.php.net/manual/de/migration81.deprecated.php~~
- [x] ~~Fix unit test failure around database query return types~~
- [x] ~~Fix calling `trim()` with `null` in page dialogs submit callbacks~~
- [x] ~~`tests/Filesystem/fixtures/f/test.txt` keeps getting deleted during unit tests~~

### Feature
- Supports PHP 8.1

## Breaking changes

None

## Related issues/ideas

- https://kirby.nolt.io/381

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -214,7 +214,9 @@ class Database
 
         // TODO: behavior without this attribute would be preferrable
         // (actual types instead of all strings) but would be a breaking change
-        $this->connection->setAttribute(PDO::ATTR_STRINGIFY_FETCHES, true);
+        if ($this->type === 'sqlite') {
+            $this->connection->setAttribute(PDO::ATTR_STRINGIFY_FETCHES, true);
+        }
 
         // store the connection
         static::$connections[$this->id] = $this;
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -212,6 +212,10 @@ class Database
         $this->connection->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
         $this->connection->setAttribute(PDO::ATTR_EMULATE_PREPARES, false);
 
+        // TODO: behavior without this attribute would be preferrable
+        // (actual types instead of all strings) but would be a breaking change
+        $this->connection->setAttribute(PDO::ATTR_STRINGIFY_FETCHES, true);
+
         // store the connection
         static::$connections[$this->id] = $this;
 
--- a/src/Toolkit/Xml.php
+++ b/src/Toolkit/Xml.php
@@ -421,7 +421,7 @@ class Xml
             return $value;
         }
 
-        $encoded = htmlentities($value);
+        $encoded = htmlentities($value, ENT_COMPAT);
         if ($encoded === $value) {
             // no CDATA block needed
             return $value;
--- a/src/Toolkit/Xml.php
+++ b/src/Toolkit/Xml.php
@@ -421,7 +421,8 @@ class Xml
             return $value;
         }
 
-        $encoded = htmlentities($value, ENT_NOQUOTES | ENT_XML1);
+        // TODO: in 3.7.0 use ENT_NOQUOTES | ENT_XML1 instead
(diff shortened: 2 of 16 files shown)


bastianallgeier:
It's great to see that simpleimage already got patched. But I'm afraid we won't have the same luck with punycode. The latest activity on that repo is from 2016. We might need to fork it

distantnative:
@bastianallgeier they recommend  https://github.com/symfony/polyfill-intl-idn - haven't looked into it yet

bastianallgeier:
Ah, nice. The Symfony stuff is mostly pretty well built and maintained.

bastianallgeier:
@distantnative I've merged all the required PRs. Can you have a look at the merge conflicts?

distantnative:
Will do and double check if 8.1 is fully covered now.

src/Database/Database.php
```diff
@@ -212,6 +212,10 @@ public function connect(array $params = null)
         $this->connection->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
         $this->connection->setAttribute(PDO::ATTR_EMULATE_PREPARES, false);
 
+        // TODO: behavior without this attribute would be preferrable
+        // (actual types instead of all strings) but would be a breaking change
+        $this->connection->setAttribute(PDO::ATTR_STRINGIFY_FETCHES, true);
```
lukasbestle: I think this only affects SQLite. MySQL is only affected with emulated prepared statements, which we AFAIK don't use.

src/Toolkit/Xml.php
```diff
@@ -421,7 +421,7 @@ public static function value($value): ?string
             return $value;
         }
 
-        $encoded = htmlentities($value);
+        $encoded = htmlentities($value, ENT_COMPAT);
```
lukasbestle, line 425: I think this can be:

```suggestion
        $encoded = htmlentities($value, ENT_NOQUOTES | ENT_XML1);
```

as quotes inside XML text are fine, just `<`, `>` and `&` are special.

distantnative: My understanding was that COMPAT was the old default, so making that explicit seemed the easiest/least risky.

lukasbestle: Good thinking, but I think `NOQUOTES` is the better choice in the long run. Otherwise we will stumble across this later and wonder why we went with `ENT_COMPAT`.

src/Database/Database.php
distantnative: Means we need to move it elsewhere? Or just context info?

lukasbestle: Not sure to be honest. For MySQL the current change is a breaking change as the previous behavior was to return native types, now we always enforce strings. For SQLite the change *prevents* a breaking change. So if we want to avoid the breaking change for both, we should only set the attribute for SQLite connections.

lukasbestle (approved):

lukasbestle:
I haven't been able to test the changes myself, but from the diff I can't see anything that could cause problems.

PR #3988, opened 2021-12-04
Replace unmaintained `true/punycode` dependency

distantnative:
## Describe the PR

- `true/punycode` isn't maintained anymore
- `true/punycode` causes issues with PHP 8.1
- `symfony/polyfill-intl-idn` is the recommended alternative and seems to be maintained, solid and an easy replacement

Ignoring `symfony/polyfill-intl-idn` when running `composer unused` as long as https://github.com/composer-unused/symbol-parser/issues/2 isn't implemented.

### Refactoring
- Replaced `true/punycode` with `symfony/polyfill-intl-idn` dependency

## Breaking changes

None

## Ready?

- [x] CI checks pass

## When merging


bastianallgeier (approved):

lukasbestle:
@distantnative This removes the `symfony/polyfill-mbstring` dependency because it was only an indirect dependency of the old punycode library. We have listed `ext-mbstring` in our requirements, but it could be that users have ignored that requirement and still got along because we shipped the polyfill. I wonder whether we should add it back, at least until 3.7.

PR #4009, opened 2021-12-07
Fix `plugins.js` regression

### LINKED ISSUE #1062: [Panel] Error when loading sections data (this.load is not a function)
Following the docs, calling `this.load` on `created` in a section component throws an error.

```javascript
created() {

  // Doesn't work
  // throws: this.load is not a function
  this.load().then(response => {
      this.test = response.test
  })

  // Hackish way that works
  this.$api
        .get(this.parent + "/sections/" + this.name)
        .then(response => {
            this.test = response.test
        })

}

distantnative:
Would be really good to get some others testing this as well

### DIFF
--- a/panel/src/config/plugins.js
+++ b/panel/src/config/plugins.js
@@ -39,19 +39,20 @@ export default {
           // if component doesn't exist, don't extend
           options.extends = null;
         }
+      }
 
-        if (options.mixins) {
-          options.mixins = options.mixins.map((mixin) => {
-            return typeof mixin === "string" ? mixins[mixin] : mixin;
-          });
-        }
+      if (options.template) {
+        options.render = null;
+      }
 
-        if (components[name]) {
-          window.console.warn(`Plugin is replacing "${name}"`);
-        }
+      if (options.mixins) {
+        options.mixins = options.mixins.map((mixin) => {
+          return typeof mixin === "string" ? mixins[mixin] : mixin;
+        });
+      }
 
-        app.component(name, options);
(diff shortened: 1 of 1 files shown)


MaluNoPeleke:
Any way to implement this manually on a Kirby 3.6.1 environment? The file doesn't exist there.

distantnative:
@MaluNoPeleke It needs more testing, but if you'd be up to help with testing, you can replace the `panel/dist` files with this: 
[dist.zip](https://github.com/getkirby/kirby/files/7676436/dist.zip)

MaluNoPeleke:
Thanks, I just did and the this.load error still exists so it doesn't seem to solve the issue yet.

distantnative:
@MaluNoPeleke have you made sure to clear the media folder?

MaluNoPeleke:
Thanks, I didn't think about that. After doing that it works again. Any way to do this automatically with the update from 3.6.1 to 3.6.1.1 to avoid further confusion?

distantnative:
I think between versions it works mostly automatically, as Kirby requests Panel assets from the media folder with a hash of the version included. New version, new hash. However, we do advise to delete the media folder for updates: https://getkirby.com/docs/guide/quickstart#updates

MaluNoPeleke:
Thanks, I will try to find out how to do this with the general composer update command (which could update other packages as well) automatically.

bastianallgeier (approved):

bastianallgeier:
Works great!

PR #4010, opened 2021-12-07
Upgrade `cypress`

distantnative:


lukasbestle (approved):
Nice that there isn't a huge list of breaking changes in the Cypress release notes for a change. ;)

distantnative:
The upside is that "only" our test setup could break, nothing in our actual system. And so far they haven't managed to break the tests 😅

Issue #3940, opened 2021-11-17
Slug sync each time reloading the Panel


plain-solutions-gmbh:
There is a strange behavior with field 'Slug'. To reproduce it, do the following:

1. create a blueprint with a label and a slug field (sync slug with label).
2. go to the panel and set the label
3. change the value of the slug as you like
4. save and reload the panel

**To expect:**
The slug should keep its value (no matter what the label is).

**What happens:**
The slug resynchronizes with the label

distantnative:
If I understand you correctly, the field just does what it's supposed to do:
A slug field with `sync: label` will always sync itself automatically with the value of the `label` field.

If you only want to generate the slug value from the label field at some point in times, but edit it on others, the `sync` option doesn't seem like a good fit for you. But rather try the `wizard` option that adds a button for one-click generation of the slug value from the other field - https://getkirby.com/docs/reference/panel/fields/slug

```yaml
wizard:
  text: "Generate from label"
  field: label
```

Does this solve your question? Or did I misunderstand your general problem?

plain-solutions-gmbh:
Hm. that's right. You can use the wizard-option indeed.
But then i missed the idea behind of the sync-option.
In this case i could remove the slug-field and convert the label-field programmatically.

distantnative:
@bastianallgeier might correct me, since he built the field:

I think the `sync` option does exactly what the word says, it keeps the value of the referenced field (label) always in sync with the value of the slug field.

Not sure why the wizard option wouldn't be the solution for your use case.

plain-solutions-gmbh:
The wizard where an alternative solution. It's more about the understanding of the `sync` option.
It's pretty confusing if you change a value and after saving, there's a different value.

bastianallgeier:
This is indeed a bug. The slug field should indeed keep the value in sync, but once a custom slug has been stored it should only resync again once the value of the synced field actively changes. This should not happen when the form is reloaded, but only when you start typing again.

plain-solutions-gmbh:
But then the `wizard` option would not make sense. 🙈
I thought the Wizard option would be available on every field.

Which brings me to the thought: What a huge sensation it would be if `wizard` would be available on every field. Especially if you could use it to execute methods within the field. Returning the content of the field.

This as a thought experiment. But that would be another topic...

plain-solutions-gmbh:
> But then the `wizard` option would not make sense. 🙈
Okay, maybe in case, the slug-field is not required... 🙊

[...]

PR #4015, opened 2021-12-08
Fix sync bug

plain-solutions-gmbh:
## Describe the PR

#3940 Fix bug in slug field: Prevent from synching when Panel reopens.

## Release notes

Fixes:
- Slug sync each time reloading the Panel
- Remove wizard props (is obsolet now)

## Breaking changes

None

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/config/fields/slug.php
+++ b/config/fields/slug.php
@@ -38,14 +38,6 @@ return [
          */
         'sync'  => function (string $sync = null) {
             return $sync;
-        },
-
-        /**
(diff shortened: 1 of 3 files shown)


bastianallgeier:
I'm sorry, but I think we have to close this. It solves the problem partially, but the field is still a problem in regular forms. Right now there's not a proper solution that we could think of without storing something like: 

```
slug: 
  value: foo
  synced: true
```

In the content file.

PR #4018, opened 2021-12-08
Reduce global helper functions

distantnative:
### Enhancements
- Helpers can be deactivated by setting a global constant named `KIRBY_HELPER_*`, e.g. `KIRBY_HELPER_DUMP` to `false`

  In some cases, our global helper functions would collide with other tools' global functions as they would use the same name, e.g. `dump()`. If you are affected by this problem, you can now set the matching constant, e.g. `KIRBY_HELPER_DUMP`, to `false` and Kirby will not register its global helper.

  To ensure that Kirby still functions properly, we moved all functionality from the helper functions to class methods (and the global helper functions are only aliases now). We recommend plugin developers to also start using these class methods to ensure that your code always uses Kirby's core helpers:

  - New `Cms\App::csrf()` method with the functionality of the `csrf()` helper
  - New `Cms\App::image()` method with the functionality of the `image()` helper
  - New `Cms\Helpers::deprecated()` method with the functionality of the `deprecated()` helper
  - New `Cms\Helpers::size()` method with the functionality of the `size()` helper
  - New `Cms\Helpers::dump()` method with the functionality of the `dump()` helper
  - New `Cms\Html::css()` method with the functionality of the `css()` helper
  - New `Cms\Html::js()` method with the functionality of the `js()` helper
  - New `Cms\Html::svg()` method with the functionality of the `svg()` helper
  - New `Filesystem\F::loadClasses()` method with the functionality of the `load()` helper
  - New `Http\Router::execute()` method with the functionality of the `router()` helper
  - New `Http\Response::go()` method with the functionality of the `go()` helper
  https://kirby.nolt.io/369
  - New `Toolkit\Date::roundedTimestamp()` method with the functionality of the `timestamp()` helper
  - New `Toolkit\Str::esc()` method with the functionality of the `esc()` helper
  - New `Toolkit\Str::uuid()` method with the functionality of the `uuid()` helper
  - New `Toolkit\V::invalid()` method with the functionality of the `invalid()` helper
  - Extended `Cms\App::kirbytag()` method with the functionality of the `kirbytag()` helper. Allows to pass an array as first parameter.
  - Extended `Cms\App::snippet()` method with the functionality of the `snippet()` helper.
  - Extended `Toolkit\Html::attr()` method with the functionality of the `attr()` helper. Allows to pass `$before` and `$after` string.

### Refactoring

- The Kirby core no longer uses Kirby's helper functions internally but calls the underlying methods directly.

### Deprecated

- The `dump` component was deprecated and superseded by the new feature to replace helpers (`KIRBY_HELPER_DUMP` constant). The component will be removed in Kirby 3.8.0.

### Breaking changes

- `Toolkit\Html::attr([], null)` now consistently returns `null` instead of an empty string.

## Related backlog issues

https://github.com/getkirby/backlog/issues/39
https://github.com/getkirby/backlog/issues/38
https://github.com/getkirby/backlog/issues/18

## Ready?

### For review team


lukasbestle:
To be honest I'm not sure if we should move the helpers to the classes by topic:

- It will make the migration for plugin devs much harder as the devs need to check where each helper was moved to. It's not a simple copy and paste operation (like to add `Helper::` before every function call). Especially considering that there is also a small breaking change.
- For example the `App` class is already such a monster. It collects more and more methods where we don't know a better place. I feel like we should be careful with adding even more methods to it.

On the other hand the grouping by topic is useful as it makes more sense in the long run to keep everything where it belongs. Difficult...

distantnative:
I mainly ended up there as half the helpers are already just calling/aliasing a method in the topic-related class. Creating another `Helper::` method for them seemed a bit ridiculous (as the helper would alias the `Helper::` method would alias the real method).

So then I started thinking where the code of those helper, that actually have their own logic, would make most sense. Which worked really well for, let's say, half of them. 

And then I went and tried to find places for the others. Here I agree, not all locations I'm super happy with. I would still say it's the right direction to find dedicated places based on topic (and it won't be so hard to migrate, we'll have a table in the release notes, they can always still check `config/helpers.php` where they'll see what class method is behind the helper). But very open to suggestions for which classes and what method names.

lukasbestle:
Good point. I agree now that the general direction makes sense. Naming is always the most difficult part. :D

distantnative:
I think next step is to determine a milestone with the team where we want to tackle the naming part together.

lukasbestle:
Suggestions:

- `deprecated()`, `dump()` and `size()` are really generic helpers. I think they really fit into a `Toolkit\Helpers` class.
- The `dump` component should IMO be deprecated. The whole reason we created that is that helper name conflict issue. But the component didn't solve that issue at all. If one wants a different `dump()` implementation, they can now define the constant to override Kirby's helper completely.
- `go()` could go in `Http\Response`.
- I find `image()` oddly specific. Could we deprecate that in favor of the already existing `$app->file()`? Only difference I can see is that `image()` uses the current page by default while `$app->file()` uses the `$site` unless a `$parent` is passed.
- For `param()` and `params()` I feel like we don't need another alias in `App`. The helpers already only have one line.
- `Router::factory()` could be renamed `Router::execute()`. I think it's clearer.
- `Classes::load()` could be placed in `F::loadClasses()`. We already have other `require`-related methods there.
- `Toolkit\Date::stamp()` could be named `roundedTimestamp()`. I think it also makes clearer what it does compared to just `timestamp()`.
- `invalid()` seems to be very very similar to the existing `V::input()` method. Maybe we can add a new `V::form()` method with the best of both worlds and deprecate the `V::input()` method (very weird name IMO). I think the whole `V` class needs refactoring. I find it very confusing while reading through the code.

distantnative:
> * `deprecated()`, `dump()` and `size()` are really generic helpers. I think they really fit into a `Toolkit\Helpers` class.

✅ Although `Cms\Helpers` as dumps is really CMS-specific


> * The `dump` component should IMO be deprecated. The whole reason we created that is that helper name conflict issue. But the component didn't solve that issue at all. If one wants a different `dump()` implementation, they can now define the constant to override Kirby's helper completely.

✅ marked as deprecated, added TODO to move/remove


> * `go()` could go in `Http\Response`.

✅ 


> * I find `image()` oddly specific. Could we deprecate that in favor of the already existing `$app->file()`? Only difference I can see is that `image()` uses the current page by default while `$app->file()` uses the `$site` unless a `$parent` is passed.

I tried to use `App::file()` - let's see what unit tests say, worried about breaking changes here.


> * For `param()` and `params()` I feel like we don't need another alias in `App`. The helpers already only have one line.

✅ 


> * `Router::factory()` could be renamed `Router::execute()`. I think it's clearer.

✅ 


> * `Classes::load()` could be placed in `F::loadClasses()`. We already have other `require`-related methods there.

✅ 


> * `Toolkit\Date::stamp()` could be named `roundedTimestamp()`. I think it also makes clearer what it does compared to just `timestamp()`.

✅ 


> * `invalid()` seems to be very very similar to the existing `V::input()` method. Maybe we can add a new `V::form()` method with the best of both worlds and deprecate the `V::input()` method (very weird name IMO). I think the whole `V` class needs refactoring. I find it very confusing while reading through the code.

Agreed that this deserves some refactoring, I would leave it as it is in the PR for now and touch it when we refactor the whole class.

lukasbestle:
I'm also worried about the breaking change with `image()`/`$app->file()`. I think we need to find a temporary home for `image()` (maybe just `$app->image()`) and deprecate it over a longer time span. But the question is if there is really no functionality in `image()` that cannot be achieved with `$app->file()`. Haven't checked that yet. If `image()` can't easily be replaced, there is no point deprecating it.

distantnative:
I've brought back `App::image()` for the `image()` helper - we can see whether we can deprecate it later on, but let's not risk it in this one.

lukasbestle (commented):
I am thinking we could call `kirby()` instead of `App::instance()` in all the helpers that depend on the `App` instance (`collection()`, `kirbytext()`, `option()`...). This would allow users to override the single `kirby()` helper to change the `App` object for all headers at once.

src/Cms/App.php
```diff
@@ -529,6 +529,43 @@ public function core()
...
     }
 
+    /**
+     * Checks/returns a CSRF token
+     * @since 3.7.0
+     *
+     * @param string|null $check Pass a token here to compare it to the one in the session
+     * @return string|bool Either the token or a boolean check result
+     */
+    public static function csrf(?string $check = null)
```
lukasbestle: We could make this an instance method to be in line with `session()` and all the others. Then we wouldn't need to use `static::instance()` inside but just `$this`.

src/Cms/App.php
```diff
@@ -770,14 +853,27 @@ public function io($input)
      * Renders a single KirbyTag with the given attributes
      *
      * @internal
-     * @param string $type
+     * @param string|array $type
```
```suggestion
     * @param string|array $type Tag type or array with all tag arguments
                                 (the key of the first element becomes the type)
```

src/Cms/App.php
```diff
@@ -1506,11 +1602,22 @@ public function smartypants(string $text = null): string
      * @internal
      * @param mixed $name
      * @param array $data
+     * @param bool $return
```
```suggestion
     * @param array|object $data Variables or an object that becomes `$item`
     * @param bool $return On `false`, directly echo the snippet
```

src/Cms/App.php
```diff
@@ -1506,11 +1602,22 @@ public function smartypants(string $text = null): string
...
      * @return string|null
      */
-    public function snippet($name, array $data = []): ?string
+    public function snippet($name, array $data = [], bool $return = true): ?string
```
```suggestion
    public function snippet($name, $data = [], bool $return = true): ?string
```

src/Cms/Auth.php
```diff
@@ -195,7 +195,7 @@ public function csrf()
     public function csrfFromSession(): string
     {
         $isDev = $this->kirby->option('panel.dev', false) !== false;
-        return $this->kirby->option('api.csrf', $isDev ? 'dev' : csrf());
+        return $this->kirby->option('api.csrf', $isDev ? 'dev' : App::csrf());
```
lukasbestle:
```suggestion
        return $this->kirby->option('api.csrf', $isDev ? 'dev' : $this->kirby->csrf());
```

src/Cms/Html.php
```diff
@@ -15,6 +18,50 @@
...
+     * @since 3.7.0
+     *
+     * @param string|array $url Relative or absolute URLs, an array of URLs or `@auto` for automatic template css loading
+     * @param string|array $options Pass an array of attributes for the link tag or a media attribute string
+     * @return string|null
+     */
+    public static function css($url, $options = null): ?string
+    {
+        if (is_array($url) === true) {
+            $links = A::map($url, fn ($url) => css($url, $options));
```
lukasbestle:
```suggestion
            $links = A::map($url, fn ($url) => static::css($url, $options));
```

src/Toolkit/Html.php
```diff
@@ -133,9 +133,11 @@ public static function a(string $href, $text = null, array $attr = []): string
      *                           Key-value array: A list of attributes will be generated. Don't pass a second argument in that case.
      * @param mixed $value If used with a `$name` string, pass the value of the attribute here.
      *                     If used with a `$name` array, this can be set to `false` to disable attribute sorting.
+     *                     * @param string|null $before An optional string that will be prepended if the result is not empty
```
lukasbestle: With these `$before` and `$after` arguments I'm not sure if we should really support `null`. I know we already do for the `attr()` helper, but maybe that should instead use `$before ?? ''` when calling this method. Otherwise we would cement those sloppy `null` arguments even more.

If we decide we do want to keep `null` support, that `?? ''` needs to happen inside this method to avoid PHP warnings in the future.

distantnative: I don't really have an opinion here to be honest.

distantnative: @bastianallgeier do you have an opinion? Guess this is the last outstanding change (plus your review then)

distantnative:
@lukasbestle please correct me if I'm missing anything

## Summary: what's left before merging?
- Decision on https://github.com/getkirby/kirby/pull/4018#discussion_r859965785
- Pass unit test coverage check
- Confirm breaking change (that we are ok with this one)

src/Toolkit/Html.php
lukasbestle: Thinking about it again, I'm fine with either solution. Only supporting `string` would be cleaner, but inconsistent. So maybe we should leave it basically like it is. With one small change: [This line here](https://github.com/getkirby/kirby/pull/4018/files#diff-9541b56c0625a5edd10a99c7249273140bb08e4b53d3cf4ce1f696034c997c78R161) should use `?? ''` to ensure that always a string is concatenated.

lukasbestle:
@distantnative

> Decision on https://github.com/getkirby/kirby/pull/4018/files/1d56f42c70767065e57e7af1d9aa5009d6f5b030#diff-9541b56c0625a5edd10a99c7249273140bb08e4b53d3cf4ce1f696034c997c78

See my reply above.

> Pass unit test coverage check

I'd be fine with merging this PR without the passing coverage check. We are already covering the most relevant parts and the overall coverage no longer drops with the tests we added.

> Confirm breaking change (that we are ok with this one)

I am OK with it. I couldn't find any use of `Html::attr()` in the core that actually differentiates between different falsy return values (`''` vs `null`). They all just check for `if ($attr)`. I feel it's going to be fine and will be much cleaner in the long run compared to the current inconsistent behavior.

src/Toolkit/Html.php
distantnative: Let's leave it then. I would also leave the mentioned line as is to make it easier to read. `null` will always converted to an empty string in PHP. https://www.php.net/manual/en/language.types.string.php#language.types.string.casting

distantnative:
Ok, sounds like we are ready to merge then?

lukasbestle (commented):
Going through the use of helpers in the core, the following question just popped up:

Where do we draw the line in reuse of custom helpers in the core? Examples:

- If the `kirbytext()` helper is overridden, the `kt()` helper will automatically use it as well.
- If the `kirby()` helper is overridden, most other helpers will use the return value of this helper at the moment. But other parts of the core use `App::instance()`, which wouldn't be affected by the custom helper.

I think the first example is fine, but I'm not sure about the second example. We could either use `App::instance()` in all helpers or we could use `kirby()` throughout the whole core.

The same question for many of the other helpers. Should users actually be able to modify core behavior with custom helpers or do we want to keep the core's behavior fully predictable?

Once we have decided that, we need to check again for helper usage in the core. I stopped at the `get()` helper that's e.g. used in the Fiber callbacks a lot.

All the rest looks great!

distantnative:
I think modifying helpers should not be done to modify core behavior. It would also be pretty hard for a user to overlook what his change does. We allow modifying behavior through extensions and exposed core components - that's the way we should stick to/change to.

distantnative:
But maybe we do the adjustment in a separate PR? I think that issue would apply same to the current state before this PR.

lukasbestle:
> I think modifying helpers should not be done to modify core behavior. It would also be pretty hard for a user to overlook what his change does. We allow modifying behavior through extensions and exposed core components - that's the way we should stick to/change to.

I agree. 👍

At first I thought about [the cookbook recipe on extending the `App` class](https://getkirby.com/docs/cookbook/extensions/replacing-core-classes), but this already fully works with the current setup.

> But maybe we do the adjustment in a separate PR? I think that issue would apply same to the current state before this PR.

Before this PR the helpers could not be replaced, so the core could rely on the functionality of the helpers. I think the feature only makes sense with these updates. I'll take a look at uses of the remaining helpers and push them to this PR.

distantnative:
Thinking about it, maybe we should make some helpers not replaceable. At least `kirby()`.

lukasbestle:
> Thinking about it, maybe we should make some helpers not replaceable. At least `kirby()`.

Why do you think so?

distantnative:
Because replacing it will have very likely huge side effects no matter what we do - the helper will be used in templates, controllers...

And this PR isn't really about making helpers replaceable/extendable but just creating the option to disable some in case of conflicts with other libraries.

I feel like we shouldn't allow a conflicting Kirby function to exist. Just so much potential for problems.

distantnative:
Not saying that your replacements aren't still valid. I am all for them. Just don't see a real use case for replacing the kirby helper, while I see a lot of problems that could be caused by it.

config/helpers.php
```diff
@@ -1,154 +1,136 @@
...
-    // argument has been passed, check the token
-    if (
-        is_string($check) === true &&
-        is_string($session->get('kirby.csrf')) === true
-    ) {
-        return hash_equals($session->get('kirby.csrf'), $check) === true;
+// @codeCoverageIgnoreStart
+if (
+    defined('KIRBY_HELPER_ASSET') === false ||
+    constant('KIRBY_HELPER_ASSET') !== false
```
afbora: What do you think about moving these repeats into `Helpers` class to avoid repeats and improve readability? Something like that:

````php
if (Helpers::isHelperDefined('asset') === false) {
  // codes
}
````

### Helpers::isHelperDefined()

````php
public static function isHelperDefined(string $name) : bool 
{
    $name = 'KIRBY_HELPER_' . strtoupper($name);
    return defined($name) === true || constant($name) === false
}
````

Comment (lukasbestle) [reply config/helpers.php]: Great idea! 👍

**Comment by lukasbestle** [comment]
> Because replacing it will have very likely huge side effects no matter what we do - the helper will be used in templates, controllers...

The only way the helpers can be disabled is by defining the respective constant before Kirby's autoloader is run. So essentially the helpers can only be disabled from the site's `index.php`, meaning by the site devs themselves.

I think the whole feature to disable helpers is an advanced feature that not many users will need. But those who do might have a reason to replace helpers like `kirby()` as well.

I don't think we should artificially limit which helpers can be disabled and which ones can't be, otherwise you would have to look up each time which constants are supported.

Since the core uses none of the helpers now, nothing can break except site code in the responsibility of the dev. I think we can solve this with a clear warning in the docs that all site code needs to be adapted when disabling or replacing helpers.

File: config/helpers.php (hunk 8 above)
Comment (distantnative) [reply config/helpers.php]: I didn't do it in the first place because I think it makes it much harder for people who face conflicts with helpers what to do. Whereas when they would see the explicit if check for the constant, it would be much more self- explanatory.

**Comment by distantnative** [comment]
Ok

File: config/helpers.php (hunk 8 above)
Comment (lukasbestle) [reply config/helpers.php]: Hm, that's a good point. It's certainly more explicit and requires less digging. On the other hand I feel it's really much cleaner with the helper method. I'd be fine with either solution.

**Comment by lukasbestle** [comment]
Remaining todos before we can finally merge this PR: 😌

- [x] @distantnative Could you please look through the changes in [this commit](https://github.com/getkirby/kirby/pull/4018/commits/8e005bc19dc6e4142ee48dd78af816cae1b91b6c)? I've written a Psalm plugin to find every helper use, so no use should be missing. But it could be that the new helper-less code has typos. It was quite a crunch, so I might have copy-pasted incorrectly somewhere.
- [x] Discussion above on the helper method `Helpers::hasOverride()`

**Comment by afbora** [comment]
Overriding is a rarely used and often advanced method/need. I think the developer who comes to `helpers.php` can reach the `hasOverride` method with one more click. IMHO, I would prefer the less repetitive one. It's not a very important/critical subject, of course 👍

**Comment by distantnative** [comment]
I'm ok with `Helpers::hasOverride()` if you two prefer it. Only stated why I didn't go that way :)

Going through the commit now, @lukasbestle.

File: config/areas/users/dialogs.php [hunk 9]
```diff
@@ -45,13 +47,16 @@
...
-                'role'     => get('role')
+            $kirby = App::instance();
+
+            $kirby->users()->create([
+                'name'     => $kirby->request()->get('name'),
+                'email'    => $kirby->request()->get('email'),
+                'password' => $kirby->request()->get('password'),
+                'language' => $kirby->request()->get('translation'),
+                'role'     => $kirby->request()->get('role')
             ]);
```
distantnative, line 58: Would be great to use your "trick" here as well to define a `$data` array with just one `$kirby->request()->get()` call for all 5 keys.

src/Panel/File.php
```diff
@@ -104,14 +105,16 @@ public function dragText(string $type = null, bool $absolute = false): string
      */
     public function dropdown(array $options = []): array
     {
+        $file    = $this->model;
+        $request = $file->kirby()->request();
```
distantnative: Maybe here too


```php
$defaults = $file->kirby()->request()->get(['view', 'update', 'delete']);
```

src/Panel/Page.php
```diff
@@ -66,14 +68,16 @@ public function dragText(string $type = null): string
      */
     public function dropdown(array $options = []): array
     {
+        $page    = $this->model;
+        $request = $page->kirby()->request();
```
distantnative:
```php
$defaults = $page->kirby()->request()->get(['view', 'sort', 'delete']);
```

distantnative (commented):
Looks really good, just 3 suggestions as I really liked your combined get method calls, @lukasbestle.

config/areas/users/dialogs.php
lukasbestle: Here the query keys don't match (`language` vs. `translation`).

afbora:
Now I checked all the helpers. I haven't found any usage except in unit tests 👍

config/helpers.php
```diff
@@ -174,782 +127,648 @@ function dump($variable, bool $echo = true): string
...
+    /**
+     * A super simple class autoloader
+     *
+     * @param array $classmap
+     * @param string|null $base
+     * @return void
+     */
+    function load(array $classmap, ?string $base = null)
+    {
+        return F::loadClasses($classmap, $base);
```
afbora, line 381:
```suggestion
        F::loadClasses($classmap, $base);
```

config/helpers.php
```diff
@@ -174,782 +127,648 @@ function dump($variable, bool $echo = true): string
...
+    }
```
afbora:
```suggestion
    function load(array $classmap, ?string $base = null): void
    {
        F::loadClasses($classmap, $base);
    }
```

config/helpers.php
```diff
@@ -174,782 +127,648 @@ function dump($variable, bool $echo = true): string
...
+}
+
+if (Helpers::hasOverride('markdown') === false) { // @codeCoverageIgnore
+    /**
+     * Parses markdown in the given string. Shortcut for
+     * `$kirby->markdown($text)`
+     *
+     * @param array $data
+     * @return string
```
afbora:
```suggestion
     * @param string|null $text
     * @param array $data
     * @return string
```

config/helpers.php
```diff
@@ -174,782 +127,648 @@ function dump($variable, bool $echo = true): string
...
-    $kirby = App::instance();
+if (Helpers::hasOverride('h') === false) { // @codeCoverageIgnore
+    /**
+     * Shortcut for html()
+     *
+     * @param string|null $string unencoded text
+     * @param bool $keepTags
+     * @return string
+     */
+    function h(?string $string, bool $keepTags = false)
```
afbora:
```suggestion
    function h(?string $string, bool $keepTags = false): string
```

config/helpers.php
```diff
@@ -174,782 +127,648 @@ function dump($variable, bool $echo = true): string
...
-        }
+if (Helpers::hasOverride('html') === false) { // @codeCoverageIgnore
+    /**
+     * Creates safe html by encoding special characters
+     *
+     * @param string|null $string unencoded text
+     * @param bool $keepTags
+     * @return string
+     */
+    function html(?string $string, bool $keepTags = false)
```
afbora:
```suggestion
    function html(?string $string, bool $keepTags = false): string
```

config/helpers.php
```diff
@@ -174,782 +127,648 @@ function dump($variable, bool $echo = true): string
...
+if (Helpers::hasOverride('snippet') === false) { // @codeCoverageIgnore
+    /**
+     * Embeds a snippet from the snippet folder
+     *
+     * @param string|array $name
+     * @param array|object $data
+     * @param bool $return
+     * @return string
+     */
+    function snippet($name, $data = [], bool $return = false)
```
afbora:
```suggestion
     * @return string|null
     */
    function snippet($name, $data = [], bool $return = false): ?string
```

config/helpers.php
```diff
@@ -174,782 +127,648 @@ function dump($variable, bool $echo = true): string
...
+     * Translate by key and then replace
+     * placeholders in the text
+     *
+     * @param string $key
+     * @param string|array|null $fallback
+     * @param array|null $replace
+     * @param string|null $locale
+     * @return string
+     */
+    function tt(string $key, $fallback = null, ?array $replace = null, ?string $locale = null)
```
afbora:
```suggestion
    function tt(string $key, $fallback = null, ?array $replace = null, ?string $locale = null): string
```

lukasbestle (approved):

distantnative:
@lukasbestle @afbora based on your latest cahnges/contributions, anything that needs to be changed in the release notes in the first post?

afbora:
Not for me 👍

afbora (approved):
Looks and works great 🎉

lukasbestle:
I went through the changes again and added a few remaining items to the list. Should be good to go now.

PR #4028, opened 2021-12-09
More PHP arrow functions

distantnative:
## Describe the PR

Use more PHP arrow functions where it makes sense.

## Ready?

- [x] CI checks pass

## When merging


afbora:
I've seen places where many more arrow functions can be done. Especially fields. Is the reason why you didn't enter the fields something to do with automatic documentation?

distantnative:
There are surely more, I went for the ones that made sense to me. I think it gets weird in parts where, e.g. half props would be arrow function, half props would be normal callbacks (as they do more than a simple return).

afbora:
So let's continue with that for now.  
Btw about code formatting, my phpstorm ide suggest without space like that `fn() =>` not like yours `fn () =>`.

bastianallgeier:
@afbora I like that `fn ()` matches `function ()` so I'd keep it this way tbh.

bastianallgeier (approved):
Couldn't find anything weird. I love that the test coverage increases :)

PR #4030, opened 2021-12-09
PHP 8.1 deprecation fixes (null passing, implicit float to int conversion)

neildaniels:
## Describe the PR

There are few PHP 8.1 changes that will throw behavior deprecation warnings.

## Release notes

Fixes:
- No longer throws deprecation warning passing null to `Str` functions
- No longer throws deprecation warning implicit float to int conversion when scaling dimensions for an SVG

## Breaking changes

None

## Ready?

- [ ] CI checks pass

## When merging


distantnative (dismissed):

src/Image/Dimensions.php
```diff
@@ -294,7 +294,7 @@ public static function forSvg(string $root)
             }
         }
 
-        return new static($width, $height);
+        return new static((int)$width, (int)$height);
```
lukasbestle: I wonder why we have the `(float)` casts above. Does anything speak against changing them to `(int)` together with `0.0` to `0` in line 290?

afbora: I guess float needed for `->ratio()` method. But `$width` and `$height` property referenced `int` in docblocks.

neildaniels: I think it’d technically be a breaking change.

If an SVG has a width of `0.5`, it currently would end up with a width of `0`; if this was changed, it’d start returning whatever the viewBox width is.

But, I’m not sure if the current behavior of returning a 0-dimension is particularly useful.

neildaniels: The constructor for Dimensions has also already been casting to `int` anyways, so ratio has already been operating on ints anyways.

lukasbestle: That's what I mean. If we cast the value to float and then to int, the result is the same if we cast to int in the first place. It's not a breaking change, it just makes explicit what was already the case before.

bastianallgeier: I also agree. We never really supported floats for width and height. But I'd also say that we should do the (int) casts instead of the (float) casts in the code above. The rest looks good to me.

neildaniels: I've made these changes.

bastianallgeier (approved):

bastianallgeier:
Thank you for your help!

Issue #3842, opened 2021-10-23
`default: now` on date field with `time: true` doesn't return proper time


GiantCrocodile:
**Describe the bug**  

**To Reproduce**  
1. Create a field like this:
```
          test_date:
            label: Test Date Field
            type: date
            time: true
            default: now
            disabled: true
            required: true
```
2. Create a page with this blueprint field
3. Check the content of the field (it's `2021-10-23 00:00`)

**Expected behavior** 
Field should contain `2021-10-23 23:54` and not `2021-10-23 00:00`

Expected behavior is described here: https://getkirby.com/docs/reference/panel/samples/blog#single-article__result

**Kirby Version**  
v3.6 rc1

**Desktop (please complete the following information):**  
Windows 10, latest Firefox

**Additional context**  
multilang site

GiantCrocodile:
Could be the same as https://github.com/getkirby/kirby/issues/3106 although no step attribute maintained here; so could be 2 different issues.

bastianallgeier:
It's the same.

Issue #3106, opened 2021-01-24
Time in date field with default: now is always 00:00


sebastiangreger:
**Describe the bug**  
When creating a new page in the Panel, a date field with `default` set to `now` always shows `00:00` as time.

**To Reproduce**  
1. Install a fresh Starterkit and log in to the panel
2. Click "Add" for Notes to create a new note
3. See error: the time in the date field shows `00:00` (despite `default` set to `now` in the `note.yml` blueprint)

The error also persists when altering the blueprint as follows:
```
date:
  type: date
  time:
    step: 1
    display: HH:mm
    default: now
  default: now
```

**Expected behavior**  
A time field with `time: true` and `default: now` would use the current time, either by default or at least when the time value is also set to `default: now`.

**Kirby Version**  
3.5.1

**Console output**  
No console output.

**Desktop:**  
 - OS: Linux Mint 19.3
 - Browser: Firefox
 - Version: 84.0.2

distantnative:
Just on my phone, so apologies that I can't check myself right now. But I think it's because your blueprint defines the step as 1 day and everything gets rounded to the nearest day therefore.

moevbiz:
I'm having the same issue with every date field, without having set `step` at all. 

```yml
date:
  type: date
  time: true
  default: today
  # same with default: now
  display: DD.MM.YYYY
```
See also this thread in the forum: https://forum.getkirby.com/t/panel-time-defaults-to-00-00/20896?u=bruno

afbora:
> But I think it's because your blueprint defines the step as 1 day and everything gets rounded to the nearest day therefore.

@distantnative But step is set for `time` prop. Shouldn't it be 1 minute?

distantnative:
That happens when I read on the phone 🙈
I would imagine all of it is related to the step mechanism but I also agree if it's date and time the default step should be based on minute, if only date then on day

CrisGraphics:
```yaml
date:
  type: date
  label: "Date"
  time:
    display: HH:mm:ss
    step:
      unit: second
      step: 1
```

I'm also having issues with the combined date + time field.
With this setup, seconds in- or decrease by 5 instead of 1.

moevbiz:
so what's the definite solution to store date & time with a correct time default? it's always showing `00:00`, no matter what I do...

MaluNoPeleke:
Would also be nice to have a time picker in the popup as well.

distantnative:
[...]

PR #4040, opened 2021-12-13
New Kirby\Toolkit\Date class

bastianallgeier:
## Features

- New `Kirby\Toolkit\Date` class as extension for PHP's DateTime class. This new class helps with all kinds of date and time related tasks and is now used by our date and time fields and the `timestamp()` helper.

## Fixes

- https://github.com/getkirby/kirby/issues/3642
- https://github.com/getkirby/kirby/issues/3106
- https://github.com/getkirby/kirby/issues/3842

## Ready?
- [x] CI checks pass

## When merging


config/helpers.php
```diff
@@ -785,60 +786,11 @@ function tc(string $key, int $count)
...
-        $parts['year'] . '-' .
-        str_pad($parts['month'], 2, 0, STR_PAD_LEFT) . '-' .
-        str_pad($parts['day'], 2, 0, STR_PAD_LEFT) . ' ' .
-        str_pad($parts['hour'], 2, 0, STR_PAD_LEFT) . ':' .
-        str_pad($parts['minute'], 2, 0, STR_PAD_LEFT) . ':' .
-        str_pad($parts['second'], 2, 0, STR_PAD_LEFT)
-    );
+    $date = Date::optional($date);
+    $step = Date::stepConfig($step);
+    $date->round($step['unit'], $step['size']);
```
distantnative: A few potential breaking changes:
- doesn't check explicitly anymore if the string is a valid date
- always does rounding, event when no step is passed

config/fields/date.php
```diff
@@ -126,51 +105,44 @@
 
             return $this->time['step'];
```
distantnative: I think we have an issue here (also had before) that we never teat the sub options of `time` (e.g. `step`) as we prepare them on the top-level

config/fields/date.php
```diff
@@ -46,42 +46,24 @@
...
-                return array_merge($default, ['size' => $step]);
-            }
-
-            if (is_string($step) === true) {
-                return array_merge($default, ['unit' => strtolower($step)]);
-            }
+            return Date::stepConfig($step, [
+                'size' => 5,
+                'unit' => 'minute',
+            ]);
```
distantnative: Default step for the date field should be 1 day. Only when `time` option is active, default step should be 5 minutes.

config/fields/date.php
bastianallgeier: Sorry, I'm lost here. What do you mean exactly :)

distantnative: In this computed property we either use the main step option or if a time option is defined, we use the step sub-option of the time option.

But while we would run the main step option through the Date class method, we just use the step sub-option as is.

bastianallgeier: Ah, I think I got it.

distantnative (approved):

Issue #3946, opened 2021-11-19
TypeError: Cannot read property '0' of null on Date field within Structure


plagasul:
**Describe the bug**  

When accessing a date field within a structure field, and clicking on the prev/next arrows of the calendar visualization I am getting a TypeError: Cannot read property '0' of null

**To Reproduce**  
1. You need a date field within a structure like this

```
      texts:
        type: structure
        fields:
          author:
            type: text
          title: 
            type: text
          date:
            type: date
            display: YYYY
          pdf:
            type: files
            layout: list
            max: 1
            uploads: document
          url:
            type: url
```

2. Add an entry, save it.
3. Access the added entry, click on the date field and then on the prev/next arrows of the calendar visualization

**Expected behavior**  
 
No error

**Screenshots**  

https://user-images.githubusercontent.com/13966673/142623411-44f58fbe-fa6a-4fdd-b4e3-3073265d80f6.mp4

**Kirby Version**  
 
3.6

**Console output**  

```
index.js:1 TypeError: Cannot read property '0' of null
    at a.separator (index.js:1)
    at vn.get (vendor.js:6)
    at vn.evaluate (vendor.js:6)
    at a.separator (vendor.js:6)
    at index.js:1
    at Array.map (<anonymous>)
    at a.patterns (index.js:1)
    at vn.get (vendor.js:6)
    at vn.evaluate (vendor.js:6)
    at a.patterns (vendor.js:6)
errorCaptured @ index.js:1
Bt @ vendor.js:6
Vt @ vendor.js:6
n @ vendor.js:6
i._wrapper @ vendor.js:6
```

**Desktop (please complete the following information):**  
 - OS: Linux Mint 20.1 Cinnamon
 - Browser: Chrome 
 - Version:  92.0.4515.159 (Build oficial) (64 bits)

**Additional context**  

The problem may be inconsistently happening, at one point I believe it did not happen without seeing a difference in the code or browser context, but I am unsure, after reloading it did happen consistently, at least until now.

afbora:
I guess duplicated with #3234

plagasul:
@afbora 

The other bug mentions not being able to input the date manually. I confirm this too.

But while s/he mentions the calendar widget being functional, I say that is not the case, at least not if you try to change the month via pre/next arrows.

**Please note that when the field is empty, this does not happen**. Only if there is something on the field.

So it works both when the date is empty OR if we first delete the field THEN try to use the calendar or input a new date.

[...]

PR #4044, opened 2021-12-14
Fix computed seperator

bastianallgeier:
## Fixes

- https://github.com/getkirby/kirby/issues/3946

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Forms/Input/DateInput.vue
+++ b/panel/src/components/Forms/Input/DateInput.vue
@@ -149,7 +149,7 @@ export default {
      * Separator from `display` format
      */
     separator() {
-      return this.display.match(/[\W]/)[0];
+      return this.display.match(/[\W]/)?.[0] ?? "";
     },
     /**
      * Array of used dayjs format tokens


panel/src/components/Forms/Input/DateInput.vue
```diff
@@ -149,7 +149,7 @@ export default {
      * Separator from `display` format
      */
     separator() {
-      return this.display.match(/[\W]/)[0];
+      return (this.display.match(/[\W]/) ?? [])[0] ?? "";
```
johannschopplich: `this.display.match(/[\W]/)?.[0] ?? ""` should work as well. 👀

afbora: Yes, this is more cleaner 👍

afbora (approved):

panel/src/components/Forms/Input/DateInput.vue
bastianallgeier: @johannschopplich that looks a lot better indeed. I'm still not super familiar with the optional operator possibilities.

johannschopplich: @bastianallgeier That's why I wanted to gently point this out to you when I happened to look in the commit. Glad it to hear it looks good to you.

distantnative:
Let's make sure to not do changes to Vue components while we are reworking them on another branch. Any PR should check if the change would be necessary there as well. I think we're good for this one, but just please keep in mind (otherwise we'll be reverting them unknowingly).

afbora:
Is this a new issue? Previously it was only a problem when there was a conflict. I don't understand why the files we haven't changed are overwritten 🤔

lukasbestle:
If the same line is changed in both PRs, Git will automatically trigger a conflict. Only issue is that it's then on us to resolve that conflict correctly. ;)

distantnative:
The whole files will change pretty much, lines like these may move to totally different files. So well for sure have merge conflicts but it'd be very easy to overlook it then how to solve these properly. Better now, to remind ourselves to either not touch the Vue components at the moment at all or really we need to evaluate twice where to fix this in the refactored components too.

lukasbestle:
Definitely best to avoid working on the same files in parallel if we can. 💯

Issue #3388, opened 2021-05-31
Panel: Date / Time fields in nested structure don't save their updated content when I hit save


trych:
**Describe the bug**  
In a structure field that is in a block (not sure if this goes for all nested structure fields, i.e. also for those that are nested within another structure field), when I have a date field or a time field and type something into this type field and then hit save (cmd + S), as I typically do to save my changes, the structure field will be closed as expected, but the value jumps back to what it previously was. If I instead click away from the field first or click on the OK button this does not happen.

**To Reproduce**  
1. Create a block that contains a structure field that contains a date and/or time field.
2. Go to the panel to add such a block.
3. Edit the date / time field of the block by typing into the text field.
4. Hit cmd + S to save.
5. Check the fields value, it should be set to what it previously (before editing) was (most likely back to empty with a new block).

**Expected behavior**  
Hitting save should save the newly entered value, as it does within date fields in non-nested structure fields (I just double checked, this works for me).

**Kirby Version**  
3.5.5

**Desktop (please complete the following information):**  
 - OS: macOS 10.13.6
 - Browser: Chrome 90.0 (most recent one)

**Additional context**  
Possibly related to #3037.

distantnative:
✅ Will be fixed in upcoming 3.6.2

Issue #3266, opened 2021-04-24
Date field with array for time throws error in plugin Vue component


sebastiangreger:
**Describe the bug**  
When we created the Commentions plugin about a year ago (i.e. before the update of the date/time field components), the `date` field with a timestamp in one-minute steps in the "Edit comment" dialog ([source code](https://github.com/sebastiangreger/kirby3-commentions/blob/de8e85bb1ee24e96e524bd3c6aa434b4eefb719f/src/components/Dialogs/EditDialog.vue#L60)) worked fine; users could change the date and time of a submitted comment.

Recently, this field does no longer appear to work; instead the following error is displayed:

![Screenshot from 2021-03-05 09-03-45](https://user-images.githubusercontent.com/6355217/110085862-ea537500-7d91-11eb-8ed9-044d8a84f516.png)

Above screenshot is from Firefox; in Chromium, the error message is **"Cannot read property '$invalid' of null"**.

Did the API change, requiring changes in our Vue component, or is this a bug?

**To Reproduce**  
1. Create a Vue component that includes the following:
```
  computed: {
    fields() {
      return {
        timestamp: {
          label: "Date and time",
          type: "date",
          required: true,
          time: {step: 1},
        },
```
2. Open the view in the panel and see the error
3. Remove the line `time: {step: 1},` and the field is shown as intended, but with the default 5-minute precision

**Expected behavior**  
The Panel view would display an input field for date and time, with the ability to enter the time with one-minute precision.

**Kirby Version**  
3.5.3

**Console output** 

in Chromium:
```
app.js:1 TypeError: this[w.p(...)] is not a function
    at p.d.get (vendor.js:12)
    at a.toNearest (app.js:1)
    at a.toFormat (app.js:1)
    at a.data (app.js:1)
    at a.<anonymous> (vendor.js:23)
    at lr (vendor.js:23)
    at ur (vendor.js:23)
    at sr (vendor.js:23)
    at a.t._init (vendor.js:23)
    at new a (vendor.js:23)
errorCaptured @ app.js:1
app.js:1 TypeError: Cannot read property '$invalid' of null
    at a.onInvalid (app.js:1)
    at a.mounted (app.js:1)
    at oe (vendor.js:23)
    at Bn (vendor.js:23)
    at Object.insert (vendor.js:23)
    at T (vendor.js:23)
    at a.__patch__ (vendor.js:23)
    at a.t._update (vendor.js:23)
    at a.r (vendor.js:23)
[...]

Issue #3234, opened 2021-04-08
Date field bug when display: YYYY


possible-is:
If i add display: YYYY to a date field and then try to enter a year manually (not via the date picker) into the date field in the panel, i get this response:
"Cannot read property '0' of null"

As a suggestion i would recommend to just automatically pick the date 01.01. of that year. 

If i change the display: YYYY to MM.YYYY the manual typing works again.

My version of kirby: 3.5.3

sboegel:
same here! It would be great to have a date picker to select the year only for this purpose.
My field definition:
```
fields:
   year:
        label: Jahr
        type: date
        display: 'YYYY'
        required: true
        calendar: true
        format: 'Y'
        translate: false
        step: year
        default: today
```

distantnative:
✅ cannot reproduce this anymore with 3.6.2 (upcoming release)

Issue #3107, opened 2021-01-25
Seconds in date field get not saved correctly


bastianallgeier:
**Describe the bug**  
When setting up the time display to also include seconds, seconds don't trigger a change and also don't get stored correctly in the text files.

distantnative:
✅ Cannot reproduce this with 3.6.2 anymore (upcoming release).
If you still encounter the issue, please provide a reproducible setup/blueprint. Thank you :)

crisgfx:
Sorry to bring this up again,
but I still can't enter seconds in the time field.
Well, I can, but the value resets back to 00 as soon as the field loses focus.

```yaml
time_info:
  type: date
  label: Ende
  display: DD.MM.YYYY
  time:
    display: "HH:mm:ss"
    step: 1
```



https://user-images.githubusercontent.com/16305374/159676819-39fe4dc4-87b4-4335-a75f-75589514e805.mp4



Kirby Version: 3.6.3 (also happenes with 3.6.2)

bastianallgeier:
@CrisGraphics that's expected in this case. The unit is minute by default. Try:
```yaml
time_info:
  type: date
  label: Ende
  display: DD.MM.YYYY
  time:
    display: "HH:mm:ss"
    step: 1
    unit: second
```

crisgfx:
Unfortuntely that doesn't work. I still get the same result.
The docs also don't mention unit `second`

Edit: sorry, they do. Was looking at the Date reference.

crisgfx:
Ok, I got it – had some wrong indentation.
This works. Sorry again for the inconvenience!

```yaml
time_info:
  type: date
  label: Ende
  display: DD.MM.YYYY
  time:
    display: "HH:mm:ss"
    step:
      unit: second
      size: 1
```

PR #4046, opened 2021-12-14
Refactor date field Vue components

distantnative:
## Describe the PR

- Use dayjs libraries for date/time inputs and fields
- Use native `input` instead of `k-text-input` (reduce one layer)

### Issues

- [x] When removing a date from the input by selecting everything and hitting the delete key, the value is not updated
- [x] When focusing on a date field for the first time, the value updates somehow and the changes bar shows up. 
- [x] Date steps are not working
- [x] Saving with CMD+S does not work when the Calendar is open

### To be solved later
- [ ] Solving https://github.com/getkirby/kirby/issues/2992

### Fixes
- Date field value gets properly saved inside structure fields when hitting Cmd+S
- Calendar dropdown of date fields closes properly inside a structure field when focussing another field
- Throws proper error when passing not an object as `step` prop to inputs
- Fixed issue when only displaying year in date field
- Fixed issue with storing seconds in date/time fields
- Fixed timezone handling, removed introduction of UTC timezone pinning

## Breaking changes

- Cannot assign arbitrary `$listeners` anymore to `k-date-field` and `k-time-field`

## Related issues/ideas

- Fixes https://github.com/getkirby/backlog/issues/25
  - https://github.com/getkirby/kirby/issues/3266
  - https://github.com/getkirby/kirby/issues/3234
  - https://github.com/getkirby/kirby/issues/3107
  - https://github.com/getkirby/kirby/issues/3388
  - https://github.com/getkirby/kirby/issues/3037
  - https://github.com/getkirby/kirby/issues/3089
  - https://github.com/getkirby/kirby/issues/3094
  - https://github.com/getkirby/kirby/issues/3946

## Ready?

- [x] CI checks pass

## When merging


bastianallgeier:
Just a thought. I know we said that it's probably easier to split date and datetime inputs. But wouldn't it be easier for 3.6.2 to keep the current fields and still try to refactor them without creating breaking changes. We could still make the split later, but maybe it helps at first to just focus on the dayjs plugins and refactoring of the backend stuff. It could at least mean less pressure for the release.

distantnative:
One main headache and source of errors is the current datetime input and how it has to merge and distribute data in synching two sub inputs (date and time). So either way we go, we have to pretty much fully refactor datetime.

distantnative:
Ok, let's try to keep the DatetimeInput for now. I think, what irritated me was 

> maybe it helps at first to just focus on the dayjs plugins and refactoring of the backend stuff

That's just not possible. The key part of the whole thing is refactoring the Vue components. The dayjs plugins on their own do nothing.

bastianallgeier:
I meant that we refactor the components with the new dayjs libs but in a way that tries to avoid any breaking changes.

distantnative:
@bastianallgeier Looking good so far. Last issue that remains is https://github.com/getkirby/kirby/issues/2992 - not sure if we can solve that at all.

bastianallgeier:
It starts to feel pretty stable, but I'm a bit lost with the last two issues (see my todos above)

1. not sure why deleting dates does not immediately fire an update or often fires no update at all
2. Date rounding feels completely off now. I.e. I cannot round years to let's say decades.

panel/src/libraries/dayjs-pattern.test.js
```diff
@@ -69,26 +69,30 @@ describe("dayjs.pattern.format()", () => {
...
+      pattern: "HH:mm:ss",
       inputs: ["19:03:22", "19.3.22"],
       result: "19:03:22"
     }
   };
 
   for (const test in data) {
     it(test, () => {
-      const pattern = dayjs.pattern(test);
+      const pattern = dayjs.pattern(data[test].pattern);
```
distantnative, line 95: @bastianallgeier what's the purpose here not to use the key as pattern but add the same as sub-key?

bastianallgeier:
Looking at it again after our discussions on Discord, I removed the debounce. The issue with instant updates in the Time field is gone and I really suspect that it was the problem with extend instead of using a mixin. Now it works great on blur and I think it is indeed the best option for now. 

I made a couple more tests and I think it is really a big step forward. I think we should ask the creators of the date issues to give it another try during the RC, but I feel like we can merge it like this.

panel/src/libraries/dayjs-pattern.test.js
bastianallgeier: I wanted to add another test for YYYY-MM-DD but with a different result. This is not possible so far. 

I didn't add the test, because the interpretation isn't able to handle it, but I still wanted to prepare it for more such tests. 

The case here is: 

```js
    "YYYY-MM-DD with just a year": {
      pattern: "YYYY-MM-DD",
      inputs: ["2012"],
      result: "2012-01-01"
    },
```

distantnative:
You had this nice styling for the date time input with separate time dropdown. But probably better to do a separate PR for that afterwards.

panel/src/libraries/dayjs-pattern.test.js
distantnative: ok :)

bastianallgeier:
@distantnative yeah, that's not doable without a big delay for 3.6.2. Splitting the inputs is unfortunately quite difficult.

bastianallgeier (approved):

PR #4049, opened 2021-12-15
refactor: more optional chaining

johannschopplich:
## Describe the PR

Just a minor refactor to give you a taste for it. 😋

Some `typeof` checks are now skipped, which you might not prefer. Feel free to only adapt, what you like.
One day the Panel will be written in TS I guess, then it will get easier. 😇 

## Related issues/ideas

- Fixes #

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Dialogs/Dialog.vue
+++ b/panel/src/components/Dialogs/Dialog.vue
@@ -219,10 +219,7 @@ export default {
     focus() {
       if (this.$refs.dialog?.querySelector) {
         const btn = this.$refs.dialog.querySelector(".k-dialog-button-cancel");
-
-        if (typeof btn?.focus === "function") {
-          btn.focus();
-        }
+        btn?.focus?.();
       }
     },
     /**
--- a/panel/src/components/Forms/Blocks/BlockSelector.vue
+++ b/panel/src/components/Forms/Blocks/BlockSelector.vue
@@ -114,7 +114,7 @@ export default {
     },
     navigate(index) {
       const ref = this.$refs["fieldset-" + index];
-      ref?.[0].focus?.();
+      ref?.[0]?.focus?.();
     },
     onClose() {
       this.dialogIsOpen = false;
--- a/panel/src/components/Forms/Blocks/BlockSelector.vue
+++ b/panel/src/components/Forms/Blocks/BlockSelector.vue
@@ -114,10 +114,7 @@ export default {
     },
     navigate(index) {
       const ref = this.$refs["fieldset-" + index];
-
-      if (ref?.[0]) {
-        ref[0].focus();
-      }
+      ref?.[0].focus?.();
     },
     onClose() {
       this.dialogIsOpen = false;
--- a/panel/src/components/Drawers/FormDrawer.vue
+++ b/panel/src/components/Drawers/FormDrawer.vue
@@ -70,9 +70,7 @@ export default {
       this.$refs.drawer.close();
     },
     focus(name) {
-      if (typeof this.$refs.form?.focus === "function") {
-        this.$refs.form.focus(name);
-      }
+      this.$refs.form?.focus?.(name);
     },
     open(tab, focus = true) {
       this.$refs.drawer.open();
--- a/panel/src/components/Forms/Field/SlugField.vue
+++ b/panel/src/components/Forms/Field/SlugField.vue
@@ -68,7 +68,7 @@ export default {
       this.$refs.input.focus();
     },
     onWizard() {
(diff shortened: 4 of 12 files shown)


panel/src/components/Dialogs/Dialog.vue
```diff
@@ -219,10 +219,7 @@ export default {
     focus() {
       if (this.$refs.dialog?.querySelector) {
         const btn = this.$refs.dialog.querySelector(".k-dialog-button-cancel");
-
-        if (typeof btn?.focus === "function") {
-          btn.focus();
-        }
+        btn?.focus?.();
```
afbora, line 222: I'm not familiar with javascript but what if `btn.focus` is not a `function`? 🤔

johannschopplich: That's why TypeScript is recommended for Kirby 4. 😜
The `querySelector` will return an `HTMLElement`, which has the focus method built-in: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus

afbora: OK for `querySelector`. What about other optional chaining changes where you made for function types?

johannschopplich: If it's a `ref`, then it's a `HTMLElement` as well.

johannschopplich: On the other elements where `selector[method]` applies, only `select`, `blur` and `focus` are the values used. Thus, we can skip the function check. If this method exists on the element, then it is intrinsically a function. Correct me if I'm wrong, of course. 🙂

distantnative: Personally, I find `focus?.()` a bit weird. I know we need it in some places (e.g. I think in the overlay it's not clear what element we are calling this on), but in most of these places we should know what element we are referring/referencing and we should know if it has a focus method or not.

panel/src/components/Forms/Blocks/BlockSelector.vue
```diff
@@ -114,10 +114,7 @@ export default {
     },
     navigate(index) {
       const ref = this.$refs["fieldset-" + index];
-
-      if (ref?.[0]) {
-        ref[0].focus();
-      }
+      ref?.[0].focus?.();
```
distantnative: Isn't this wrong? Question is not whether `ref` exists, but if it's empty or has actually a first element

johannschopplich: Absolutely! Great catch. Should be:

```js
ref?.[0]?.focus?.();
```

johannschopplich: As for other parts of the code: the optional call on `focus()` is probably redundant. However, as long as we don't use TypeScript we'd better be safe...

panel/src/components/Dialogs/Dialog.vue
johannschopplich: Weird it may be, but safer it is. 😉 TypeScript will help you from Kirby 4 onwards. 😇

distantnative: As far as I can see, Typescript isn't really an option with Vue 2

johannschopplich: Kirby 4 would be based on Vue 3, I guess... 👀

distantnative: Yea, just thinking out loud here: I don't think we will ever be really able to do the big lift as from v2 to v3, rewriting everything for more than a year. So I'm still looking at ways to rather gradually introduce and move from Vue 2 to Vue 3 or from JS to TypeScript. But every time it feels more like they want us to pull the rug out beneath our feet.

johannschopplich: Thanks for sharing your thoughts. Maybe rewriting using TS and Composition API won't be important and is optional. But one day, the jump to Vue 3 (without rewriting the entire panel, "only" the necessary parts) will probably make sense. Simpler plugins may just have to be recompiled with kirbyup (would then introduce a v3 option). More complex ones would probably still be tricky to migrate.

nilshoerrmann:
Just a note from the outside: If you want to keep the core easily readable and understandable for others, it might be good not to use optional chaining everywhere you could. It raises complexity in some cases where an if statement might explain the intention of the code more directly.

distantnative:
I moved the changes that I think we can add here: https://github.com/getkirby/kirby/pull/4133
Not removing the checks for functions for now.

PR #4051, opened 2021-12-15
Fixes more possible deprecated warnings when `null` passing for PHP 8.1

### LINKED ISSUE #4047: Calling `url()` helper throws deprecation warning with PHP 8.1
## Description
Deprecated: substr(): Passing null to parameter #1 ($string) of type string is deprecated in /kirby/config/components.php on line 378

## Your setup
**Kirby Version**  
3.6.1.1

afbora:
## Describe the PR

Fixed more possible deprecated warnings when `null` passing for PHP 8.1.

### Fixes

- Fixed calling `url()` helper throws deprecation warning with PHP 8.1 #4047 

## Breaking changes

None

## Related issues/ideas

- Fixes #4047 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Http/Url.php
+++ b/src/Http/Url.php
@@ -77,13 +77,13 @@ class Url
     /**
      * Tries to fix a broken url without protocol
      *
-     * @param string $url
+     * @param string|null $url
      * @return string
      */
     public static function fix(string $url = null): string
     {
         // make sure to not touch absolute urls
-        return (!preg_match('!^(https|http|ftp)\:\/\/!i', $url)) ? 'http://' . $url : $url;
+        return (!preg_match('!^(https|http|ftp)\:\/\/!i', $url ?? '')) ? 'http://' . $url : $url;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Http/Url.php
```diff
@@ -77,11 +77,15 @@ public static function currentDir(): string
...
-     * @param string $url
+     * @param string|null $url
      * @return string
      */
     public static function fix(string $url = null): string
     {
+        if (empty($url) === true) {
+            return '';
+        }
+
```
afbora: Possible breaking change?

bastianallgeier: Why do you think so?

afbora: Current

    Url::fix(); // Returns: http://

Now

    Url::fix(); // Returns: ''
    
If the variable is empty, there is nothing to fix.

bastianallgeier: Oh, I see. I think the old version is completely buggy though. Do we see this as a breaking change or as a fix?

afbora: Since this usage does not make sense, we can call it a bug-fix.

distantnative: I wouldn't check for empty, but rather pass `$url ?? ''` to `preg_match`

bastianallgeier (approved):

PR #4058, opened 2021-12-19
Additional YAML handlers

distantnative:
## Describe the PR

- Adds `yaml` config option (with current default `spyc`, `symfony` as alternative)
- Adds `symfony/yaml` as alternative YAML handler

### TODO
If someone wants to help, some things that are still missing:
- [ ] Add PECL handler
- [ ] Mirror unit tests for `spyc` handler
- [ ] Add additional unit tests for `symfony` handler 
- [ ] Test string escaping issue/handling with Symfony
- [ ] Test performance

### Features
- Added new YAML handler based on Symfony library
- Added new `yaml` config option to choose YAML handler (`spyc` default, `symfony` alternative)

### Notes
- We are planning to switch to the `symfony` YAML handler in the long run. Switching the default for the config option will happen in a future major release (3.7 or 3.8).

## Breaking changes

None

## Related issues/ideas

- Contributes to https://github.com/getkirby/backlog/issues/70
- Fixes https://github.com/getkirby/kirby/issues/4284
- Fixes https://github.com/getkirby/kirby/issues/4493

## Ready?

- [ ] CI checks pass

## When merging


src/Data/Yaml.php
```diff
@@ -24,22 +26,33 @@ class Yaml extends Handler
      */
     public static function encode($data): string
     {
-        // TODO: The locale magic should no longer be
-        //       necessary when support for PHP 7.x is dropped
+        // get YAML handler from config option
+        $handler = kirby()->option('yaml', 'spyc');
```
afbora, line 30: What do you think about `yaml.handler` option name to standart instead just `yaml`? Like `date.handler`. Not important just idea..

distantnative:
Closing the PR for now until we're closer to have this ready.

src/Data/Yaml.php
lukasbestle: I'm thinking if we should make the YAML handler a component just like `markdown`, `smartypants` etc.

**Advantages:**

- Would be more consistent in the long run.
- No additional config option needed.

**Disadvantages:**

- During the migration phase, it wouldn't be as simple to use a specific handler.
- The `PECL` handler would not be part of the core as there can only be one core component. But we could make it an official plugin.

**Possible compromise:**

We could make this a component, but also introduce a config option in 3.7 that controls which of the two default components (spyc or symfony) is activated (if not set, it would default to `spyc`). In 3.8 we would remove that option again and always use `symfony`. If users need an entirely different handler, they could register the component in a plugin. We could then provide an official plugin for Spyc for those who still need it.

PR #4064, opened 2021-12-20
Vue component tests via `vitest`

distantnative:


bastianallgeier:
What are your first impressions comparing Peeky and Vitest?

distantnative:
I had the feeling peeky might be a bit faster, but I got component tests properly working with vitest so quickly, code coverage also seems more built already (though still getting some odd results. And development speed at vitest looks better, mainly because of more developers. peeky has this UI but it isn't actually that super helpful, vitest console output is much better and kinda removes the need for a UI solution.

bastianallgeier:
Ok, that sounds like a lot of good points for Vitest. I will merge the peeky test PR first and then rebase this one.

bastianallgeier (dismissed):

bastianallgeier (approved):

PR #4069, opened 2021-12-22
`k-prev-next` component tests

distantnative:


panel/src/components/Navigation/__snapshots__/PrevNext.test.js.snap
```diff
@@ -0,0 +1,8 @@
+// Vitest Snapshot v1
+
+exports[`PrevNext.vue > has buttons 1`] = `
+<k-button-group-stub
+  buttons="[object Object],[object Object]"
```
bastianallgeier, line 5: Do snapshots really look like this?

distantnative: Ok, I think I should've added explanations: Our issue is that we use all our components as global components (also helpers etc.). While our tests only know about one component. We could register all components for all tests, but I am afraid that could end in bad performance and also adds dependencies between components.

However, with component/units tests we actually want to test the component in isolation. Only want to test what the component does itself. Not the children. That's why we stub `k-button-group` here, becoming `k-button-group-stub` and also being rendered like this in the snapshot.

Whether the button group component resolves this to the right result is something we need to make sure in tests for button group. We define an interface (props, methods) and we can trust in other components like prev-next that the child components will behave accordingly when using the interface as intended.

I know this doesn't always feel 100% right. But the other approach (registering everything globally, resolving all child components till the end) doesn't feel either.

bastianallgeier (commented):
It all looks good to me and I can make the Preview test run and pass when I open the file, but it's not included in the original list of tests when I run vitetest for the first time. Do you get that result as well? It's a bit weird.

bastianallgeier (approved):
Doesn't hurt to already merge it and it's great to have it as a template for more component tests.

PR #4070, opened 2021-12-22
Fix vite rollupOptions input path

afbora:
Vite rollupOptions input path fixed based on the document: https://rollupjs.org/guide/en/#input

Fixed the related error:

````
vite v2.7.1 building for production...
✓ 0 modules transformed.
Could not resolve entry module (......\src\index.js).
error during build:
Error: Could not resolve entry module (......\src\index.js).
````

### DIFF
--- a/panel/vite.config.js
+++ b/panel/vite.config.js
@@ -55,7 +55,7 @@ export default defineConfig(({ command }) => {
       minify: "terser",
       cssCodeSplit: false,
       rollupOptions: {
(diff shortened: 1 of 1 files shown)


distantnative:
I think `./` would be even safer

afbora:
You know better. I just preferred it as it is always used without slash in the document.

distantnative (approved):

PR #4077, opened 2022-01-01
fix: #4054 fiber handling of redirects

pReya:
## Describe the PR
I need to use the `home:` property in the user blueprint to redirect to a frontend page, so I fixed issue #4054. Unfortunately this was more complicated than expected, so let me explain a little:

It seems that #4054 is a regression from #3909.

When the `home:` property in the blueprint is supposed to redirect to a frontend page, the API will return a 302 response with the `Location:` header set to the proper target. However, the panel does not handle these redirect responses, yet. So it failed while trying to parse the response HTML from the redirected target.

It seems simple to fix: Just check the `Location:` header and redirect in JS to the given, correct location. However, it is not possible to get the content of the `Location` header from a redirected request via Ajax (supposedly this is a security thing). Even though there is a `redirect:` option in the `fetch()` API, which looks promising and can be set to `manual` – it clears all headers.

So I used a little workaround, which I found [here](https://github.com/axios/axios/issues/932#issuecomment-533408837): By checking for the response type, we can determine, whether the request was redirected. If that's the case, we can just point the browser to `response.url` (which is not the same as the URL in the `Location` header – but is still the initially requested URL e.g. /panel/login). However when the browser makes this request to the API (in contrast to an Ajax request), it will properly follow the request to the URL given in the `Location` header.

I hope this was understandable.

### Fixes
- Fixed redirects to frontend urls in `home:` property in user blueprint #4054 

## Related issues/ideas

- Fixes #4054

## Ready?
This could probably use a proper test. I can still add this, but I wanted to get your opinion on this solution, before I put time onto creating a test.

- [ ] Unit tests for fixed bug/feature
- [ ] In-code documentation (wherever needed)
- [ ] CI checks pass

## When merging

- [ ] Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)
- [ ] Add changes to release notes draft in Notion

PR #4079, opened 2022-01-02
fix: #4056 panel redirect without panel access

### LINKED ISSUE #4056: User blueprint: `home` option redirect fails for users w/out panel access
## Description
Setting a frontend URL in the `home` option of a user blueprint (e.g. `home: "{{ site.find('notes').children.first.url }}"`) leads to a pop-up with an error "You are not allowed to access the panel" if `panel: false` is set for that user.

**Expected behavior**  
Forwarding a user without panel access to a frontend URL should not lead to an error. This is a precondition to making use of the Kirby core's login flows for frontend-only users.

As discussed in [this Discord thread](https://discord.com/channels/525634039965679616/921888554521616444).

## To reproduce
1. Install a fresh Starterkit and create an admin user
2. Add `panel: false` and `home: "{{ site.find('notes').children.first.url }}"` to the `user/default.yml` blueprint
3. Try to log in

## Your setup
**Kirby Version**  
3.6.1

 - Device: PC
 - OS: Linux Mint
 - Browser: Firefox
 - Version: 95

**Additional context**  
Possibly related to #4054 (probably need to be fixed first, before this will work)

pReya:
## Describe the PR
This closes #4056. With 3.6 the permission check has been moved into `src/Panel/Home.php` when `Home::url()` is being called, so it needs to be disabled in the router via the `auth => false` property in the route definition. Otherwise it will fail before `Home::url()` is even called.

### Fixes
- Redirect for `home:` redirects from the user blueprint, when the user has no panel access #4056 

## Related issues/ideas

- Fixes #4056 

## Ready?

#4077 needs to be merged before. If you agree with this approach I will try to come up with a test case for this.

- [ ] CI checks pass

## When merging


afbora:
This is not valid solution. Seems working when user already logged. But doesn't work if user try to login (new session) with your `auth = false` solution. You'll get `The JSON response could not be parsed` error.

pReya:
> This is not valid solution. Seems working when user already logged. But doesn't work if user try to login (new session) with your `auth = false` solution. You'll get `The JSON response could not be parsed` error.

@afbora This is another bug. I've created another PR for this problem: #4077. The other PR will need to be merged first.

afbora:
Ohh, sorry I missed that. I'll check that 👍 

Btw you need to change base branch `main` to `develop` for both PR to safely merge.

bastianallgeier (dismissed):

bastianallgeier (approved):

bastianallgeier:
Argh, I didn't properly read the required redirect changes before merging. I'll have a look at that asap.

PR #4081, opened 2022-01-05
Make the JSON unexpected error message translatable

damiendart:
## Describe the PR

This PR makes the JSON unexpected error message translatable. From the [Kirby Feedback](https://kirby.nolt.io/386) site:

> When debug mode is off, the message that is included with JSON error responses is a hard-coded message in English. Would it be possible to make that string translatable?
> 
> As well as supporting other languages, you could then potentially override that message to display something a little more appropriate for non-developers (e.g. “An unexpected error occurred. An administrator has been notified.”).

A couple of notes:

- I haven't written any tests yet as I'm a little unsure on how to go about testing this change.
- I've added the English translation string to `i18n/translations/en.json` manually. I'm aware that translations are managed with Transifex, so if required I can sign up and make sure the translation is added there.

## Release notes

Enhancement: Make the JSON unexpected error message translatable

## Breaking changes

None

## Related issues/ideas

- Implements https://kirby.nolt.io/386

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -157,6 +157,8 @@
 
   "error.template.default.notFound": "The default template does not exist",
 
+  "error.unexpected": "An unexpected error occurred! Enable debug mode for more info: https://getkirby.com/docs/reference/system/options/debug",
+
   "error.user.changeEmail.permission": "You are not allowed to change the email for the user \"{name}\"",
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
This is a great PR. Simple to review and effective in its solution. I don't think we need additional tests for this. Adding the translation to the en.json is correct. We will upload this to Transifex before the release so other languages can translate that string too.

PR #4085, opened 2022-01-11
Redirects to non-fiber resources

### LINKED ISSUE #4084: Conceptual oversight in 3.6? Redirects and Fiber
## Description

I think I have found a rather big conceptual oversight in Kirby 3.6. With the Fiber architecture, most requests in the Panel are made via AJAX. However this will create unexpected behavior (the resulting behavior is described in #4054) , when the Kirby Backend returns a HTTP Redirect (Responses with a 302 code and a `Location` header). Even though an AJAX request will by default follow the HTTP redirect, the page which is the target of the redirect, will not be displayed (in contrast to a "proper" browser request, where the browser will automatically display the target of a redirect). So the result of these requests will just be shown in the little error modal with a JSON parsing error (because the response is a full HTML page, not just JSON).

I thought this would be easy to fix. Just check in JS if the response is a 302/redirect, and then manually navigate to the target given in the `Location:` header. However you can't access the `Location:` header via JS in an AJAX request (it's a security feature).

So I think the only way to fix this, will be to include some Location/Target property in the body of redirect responses, which can then be accessed in JS and manually redirected to.

Would love to have somebody from the Kirby team confirm this problem/oversight and/or mention some other solutions for this problem.

I thought I had a fix for this problem with #4077, but it's really not a proper solution.

bastianallgeier:
## Describe the PR

When firing a redirect to a non-fiber page in routes, the redirect would lead to non-parseable output and the request would fail. This PR fixes this issue.  

### Fixes

- Fixed redirects in fetch requests to non-fiber resources: https://github.com/getkirby/kirby/issues/4084

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/panel/src/fiber/index.js
+++ b/panel/src/fiber/index.js
@@ -202,7 +202,7 @@ export default {
       // redirect to non-fiber resources
       if (response.headers.has("X-Fiber") === false) {
         window.location.href = response.url;
-        return new Promise(() => {});
+        return false;
       }
 
       const text = await response.text();
--- a/panel/src/fiber/index.js
+++ b/panel/src/fiber/index.js
@@ -199,6 +199,12 @@ export default {
         }
       });
 
+      // redirect to non-fiber resources
+      if (response.headers.has("X-Fiber") === false) {
+        window.location.href = response.url;
+        return new Promise(() => {});
+      }
+
       const text = await response.text();
       let json;
 
--- a/src/Panel/View.php
+++ b/src/Panel/View.php
@@ -397,10 +397,6 @@ class View
      */
     public static function response($data, array $options = [])
     {
-        $kirby = kirby();
-        $area  = $options['area']  ?? [];
-        $areas = $options['areas'] ?? [];
-
         // handle redirects
         if (is_a($data, 'Kirby\Panel\Redirect') === true) {
             return Response::redirect($data->location(), $data->code());


pReya (changes requested):
EDIT: Nevermind, I was stupid. This works as expected.

Also, this approach will create unnecessary requests. Because the frontend will follow the redirect and load the `Location:` target into the AJAX response body (which is not being used at all). And then, after the first request, the browser will load the redirect target a second time, when `window.Location.replace` is being called.

bastianallgeier:
Not sure I'm able to follow to be honest. 

According to MDN response.url is the final URL after any redirects, so it's pretty perfect for this case: https://developer.mozilla.org/en-US/docs/Web/API/Response/url

pReya:
> Not sure I'm able to follow to be honest.
> 
> According to MDN response.url is the final URL after any redirects, so it's pretty perfect for this case: [developer.mozilla.org/en-US/docs/Web/API/Response/url](https://developer.mozilla.org/en-US/docs/Web/API/Response/url)

Sorry, I was confused. It works as expected, but it still creates an additional/unnecessary request as described above. But I guess there's no way to prevent that?

bastianallgeier:
I think it's not possible without it. There are two kind of request types in the Panel. 

1. A regular good old HTTP GET request that would normally return the full document for a view. If a view issues a redirect, that GET request will be redirected just as expected. 

2. A JS fetch request. This fetch request expects proper JSON as result to be able to update the state of the view properly. If the response is a redirect there are two possible results once the redirect is finished: 

a. The result is still a proper Panel view that sends back JSON. The view can be rendered and all is fine. 
b. The result is an HTML page or another resource outside the Panel. In this case, the Panel cannot simply update the Vue app with a new state because it does not receive any usable data. The only other option would be to replace the entire HTML document with the result. But this seems very hacky to me and would only work for plain text or HTML results. Redirects to images or other resources would need to be treated differently. That's why reloading the final destination again seems the most stable choice to me.

rasteiner:
Actually the resource isn't probably even downloaded (not in its entirety, at least). The promise returned by `fetch` resolves as soon as there are response headers (that's why `response.text()` is another promise), and this code does its `window.location.href = response.url; return false;` immediately after that. Probably if the resource is small it fits into the first chunks and is downloaded, but that isn't really a problem. If the resource is large (and would therefore be a problem) the download is (read: should be) aborted shorty after the redirect.

lukasbestle (approved):
I think it's a nice and elegant solution. If we wanted to optimize for large files, we could first send a `HEAD` request, but that would bring a huge overhead to every request. And given that 99.9 % of responses are Fiber responses, it would very likely decrease performance overall. This solution covers the edge case and it does so well.

PR #4091, opened 2022-01-15
Add Inuktitut syllabics

PaulMorel:
## Describe the PR

Adds Unified Canadian Aboriginal Syllabics for Inuktitut for slug transliterations.

I'm working on a website that has some page titles in Inuktitut, so I added this new slug rule.

Example:

**ᐊᓚᒃᑳᔪᑦ**  becomes  **Alakkaajut**
**ᓄᓇᕕᒃ** becomes **Nunavik**
**ᓇᑦᓯᓕᖕᒥᐅᑦ** becomes **Natsilingmiut**

Personally, I don't know the language. It's a good start but could probably be improved into the future.

## References 

- https://tusaalanga.ca/node/2506
- https://r12a.github.io/scripts/cans/
- http://tiro.com/Ross/Canadian_Native_Languages_V1.2.pdf
- https://en.wikipedia.org/wiki/Inuktitut_syllabics


lukasbestle (commented):
I have read about Inuktitut before, but I can't read it either. I agree that your version is a good start. So far you'll probably be the only one who will use the rules, but it's not a problem to include them in Kirby IMO.

One thing I noticed: The rules are used by `str_replace()` internally (in `Str::ascii()`). I think the longer combinations of characters need to be further up before their parts, otherwise a part of one syllabic will be already converted to an invalid value by the time PHP reaches the bottom of the list.

PaulMorel:
> So far you'll probably be the only one who will use the rules,

Ah, maybe! I work with a lot of artist-run centers and cultural organizations here in Canada and they often present and collaborate with [First Nations](https://en.wikipedia.org/wiki/First_Nations) and Inuit artists. So I've encountered Inuktitut in my Kirby sites before, but this is the first time I had syllabics in page titles.

> I think the longer combinations of characters need to be further up before their parts.

I think you're right. I noticed this in the Japanese `ja.json` rules. If I recall the only character combinations is the `qq-` cluster. So these should most likely come before the `q-` cluster.  Unless there are some grapheme clusters that I'm not aware. 

I'll reorder them and double check the rest.

lukasbestle:
I'm pretty sure all these characters use multiple UTF-8 bytes. So we also need to be careful about the encoding – `str_replace()` will just replace bytes, not characters.

PaulMorel:
Oh I see what you mean. I though you mean multiple codepoints or characters, not multiple bytes. 

You're right, each character is 3 bytes. I'll do some tests see how the order affects the `str_replace()` step.

lukasbestle:
Actually I really did mean multiple characters at first, it only occurred to me later that we also have the additional issue of `str_replace()` not respecting multibyte strings.

Please let me know once the PR is ready for review, then I'll test it as well.

PaulMorel:
@lukasbestle I did a bunch of tests with `str_replace()` after moving some clusters around and I didn't have any problems with any strings.

---

As for multibyte strings and `str_replace()` [apparently it's fine with UTF-8 in normal circumstances.](https://stackoverflow.com/questions/2652193/can-str-replace-be-safely-used-on-a-utf-8-encoded-string-if-its-only-given-vali).

Another comment on this that I found:
>  So long as all arguments used are well formed UTF-8, no problems.
>
> This works because every complete character sequence in a UTF-8 string is unique (cannot be mistaken as part of a longer sequence).

[Source](https://web.archive.org/web/20120614062652/http://www.phpwact.org/php/i18n/utf-8)

lukasbestle (approved):
Thank you very much! I've done some tests with strings of Inuktitut I could find online and all were transliterated correctly.

PR #4093, opened 2022-01-17
Support single string as `panel.favicon` value #4092

### LINKED ISSUE #4092: Passing string to `panel.favicon` config option throws error
## Description
[According to the docs](https://getkirby.com/docs/reference/system/options/panel#custom-panel-favicon), it should be possible to pass a path to the `panel.favicon` config option.

```php
return [
  'panel' => [
    'favicon' => 'assets/favicon.ico'
  ]
];
```

However, this throws an error on line 34 in `/kirby/views/panel.php` where an array is expected (foreach() argument must be of type array|object, string given). The tests also only test with an array.

**Expected behavior**  
The option should work as advertised in the docs.

## Your setup
**Kirby Version**  
3.6.6.1

distantnative:
## Fixes 
- `panel.favicon` option now supports single string as value

```php
return [
  'panel' => [
    'favicon' => 'assets/favicon.ico'
  ]
];
```

## Breaking changes

_None_

## Related issues/ideas

- Fixes #4092

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Panel/Document.php
+++ b/src/Panel/Document.php
@@ -3,6 +3,7 @@
 namespace Kirby\Panel;
 
 use Kirby\Exception\Exception;
+use Kirby\Exception\InvalidArgumentException;
 use Kirby\Filesystem\Dir;
 use Kirby\Filesystem\F;
 use Kirby\Http\Response;
@@ -174,7 +175,8 @@ class Document
      */
     public static function favicon(string $url = ''): array
     {
-        $icons = kirby()->option('panel.favicon', [
+        $kirby = kirby();
+        $icons = $kirby->option('panel.favicon', [
             'apple-touch-icon' => [
                 'type' => 'image/png',
                 'url'  => $url . '/apple-touch-icon.png',
@@ -189,17 +191,21 @@ class Document
             ]
         ]);
 
+        if (is_array($icons) === true) {
+            return $icons;
+        }
+
         // make sure to convert favicon string to array
         if (is_string($icons) === true) {
-            $icons = [
+            return [
                 'shortcut icon' => [
-                    'type' => 'image/svg+xml',
+                    'type' => F::mime($icons),
                     'url'  => $icons,
                 ]
             ];
         }
 
-        return $icons;
+        throw new InvalidArgumentException('Invalid panel.favicon option');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Panel/Document.php
```diff
@@ -158,22 +147,59 @@ public static function customCss(): ?string
...
+                'type' => 'image/png',
+                'url'  => $url . '/favicon.png',
+            ]
+        ]);
+
+        // make sure to convert favicon string to array
+        if (is_string($icons) === true) {
+            $icons = [
+                'shortcut icon' => [
+                    'type' => 'image/svg+xml',
```
afbora: Does it have to be the default SVG? What if we set the mime value dynamically?

````php
'type' => F::mime($icons),
````

Comment (distantnative) [reply src/Panel/Document.php]: True.... but I think then we need an absolute path for `$icons`, right?

Comment (afbora) [reply src/Panel/Document.php]: I'm not sure but yes. We can prefix kirby root path:

````php
'type' => F::mime($kirby->root('index') . '/' . ltrim($icons, '/')),
````

Comment (distantnative) [reply src/Panel/Document.php]: Would 

```php
return [
  'panel' => [
    'favicon' => 'assets/favicon.ico'
  ]
];
```

work anyways later on or do we always need an absolute path?

afbora: I'm really not sure of that either. Because it may not work without absolute path in multiple or custom domain setups? May be @lukasbestle have an idea about that.

distantnative: I just pushed a new version that I think works

src/Panel/Document.php
```diff
@@ -139,14 +127,16 @@ public static function assets(): array
...
      *
+     * @param string $option asset option name
      * @return string|null
      */
-    public static function customCss(): ?string
+    public static function customAsset(string $option): ?string
     {
-        if ($css = kirby()->option('panel.css')) {
+        if ($css = kirby()->option($option)) {
             $asset = asset($css);
```
afbora: It's not a big deal, but we'd better change the name of the `$css` variables. Because it can be confusing.

bastianallgeier (dismissed):
I really like how clean this is now.

afbora (approved):

PR #4095, opened 2022-01-17
Safe mode for markdown

bastianallgeier:
## Describe the PR

This PR adds access to the Parsedown safeMode option and cleans up the method calls for markdown and kirbytext methods. 

Safe mode can now be enabled like this: 

```php 
// helpers
kirbytext($text, [
  'markdown' => [
    'safe' => true
  ]
]);

kirbytextInline($text, [
  'markdown' => [
    'safe' => true
  ]
]);

markdown($text, [
   'safe' => true
]);

// field methods
$page->text()->kirbytext([
  'markdown' => [
    'safe' => true
  ]
]);

$page->text()->kirbytextInline([
  'markdown' => [
    'safe' => true
  ]
]);

$page->text()->markdown([
   'safe' => true
]);


```

The safe mode is now also available as global markdown setting: 

```php
<?php 

// config.php
return [
  'markdown' => [
    'safe' => true
  ]
];
```

## Breaking changes
None

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -138,6 +138,8 @@ return [
      * @param array $options Markdown options
      * @param bool $inline Whether to wrap the text in `<p>` tags (deprecated: set via $options['inline'] instead)
      * @return string
+     * @todo add deprecation warning for $inline parameter in 3.7.0
+     * @todo remove $inline parameter in in 3.8.0
      */
     'markdown' => function (App $kirby, string $text = null, array $options = [], bool $inline = false): string {
         static $markdown;
--- a/config/components.php
+++ b/config/components.php
@@ -153,7 +153,7 @@ return [
             $config   = $options;
         }
 
-        return $markdown->parse($text, $options['inline'] ?? false);
+        return $markdown->parse($text, $options['inline']);
     },
 
     /**
--- a/config/components.php
+++ b/config/components.php
@@ -136,12 +136,15 @@ return [
      * @param \Kirby\Cms\App $kirby Kirby instance
      * @param string $text Text to parse
      * @param array $options Markdown options
-     * @param bool $inline Whether to wrap the text in `<p>` tags
+     * @param bool $inline Whether to wrap the text in `<p>` tags (deprecated: set via $options['inline'] instead)
      * @return string
      */
     'markdown' => function (App $kirby, string $text = null, array $options = [], bool $inline = false): string {
         static $markdown;
         static $config;
+
+        // support for the deprecated fourth argument
+        $options['inline'] ??= $inline;
 
         // if the config options have changed or the component is called for the first time,
         // (re-)initialize the parser object
@@ -150,7 +153,7 @@ return [
             $config   = $options;
         }
 
-        return $markdown->parse($text, $inline);
+        return $markdown->parse($text, $options['inline'] ?? false);
     },
 
     /**
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -449,12 +449,12 @@ function kirbytags(?string $text = null, array $data = []): string
  * given string. Shortcut for `$kirby->kirbytext()`
  *
  * @param string|null $text
- * @param array $options
+ * @param array $data
  * @return string
  */
-function kirbytext(?string $text = null, array $options = []): string
+function kirbytext(?string $text = null, array $data = []): string
 {
-    return App::instance()->kirbytext($text, $options);
+    return App::instance()->kirbytext($text, $data);
 }
 
 /**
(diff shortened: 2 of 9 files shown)


config/components.php
```diff
@@ -136,21 +136,24 @@
      * @param \Kirby\Cms\App $kirby Kirby instance
      * @param string $text Text to parse
      * @param array $options Markdown options
-     * @param bool $inline Whether to wrap the text in `<p>` tags
+     * @param bool $inline Whether to wrap the text in `<p>` tags (deprecated: set via $options['inline'] instead)
```
distantnative, line 139: We should add `@todo` tags for adding deprecation warning in 3.7.0 and removing the parameter in 3.8.0

config/helpers.php
```diff
@@ -449,12 +449,12 @@ function kirbytags(?string $text = null, array $data = []): string
  * given string. Shortcut for `$kirby->kirbytext()`
  *
  * @param string|null $text
- * @param array $data
+ * @param array $options
```
distantnative: Wondering if we need to consider this a breaking change with named parameters in PHP 8.0 and higher?

src/Cms/App.php
```diff
@@ -788,15 +789,21 @@ public function kirbytags(string $text = null, array $data = []): string
      *
      * @internal
      * @param string|null $text
-     * @param array $data
-     * @param bool $inline
+     * @param array $options
+     * @param bool $inline (deprecated: use $options['markdown']['inline'] instead)
```
distantnative: `@todo` tags here as well

src/Cms/App.php
```diff
@@ -890,12 +897,22 @@ public function locks(): ContentLocks
...
      * @param string|null $text
-     * @param bool $inline
+     * @param bool|array $options
      * @return string
      */
-    public function markdown(string $text = null, bool $inline = false): string
+    public function markdown(string $text = null, $options = null): string
     {
-        return ($this->component('markdown'))($this, $text, $this->options['markdown'] ?? [], $inline);
+        // support for the old syntax to enable inline mode as second argument
```
distantnative, line 914: Add `// TODO` comment

distantnative:
Looks good to me in general. We should add some todo comments, so that we can remember to make the deprecation changes in 3.7.0 etc. more easily. And not sure if renaming parameters should be considered a breaking change.

config/helpers.php
bastianallgeier: I reverted those. They are not necessary in a minor release.

config/components.php
bastianallgeier: ✅

src/Cms/App.php
bastianallgeier: ✅

src/Cms/App.php
bastianallgeier: ✅

config/components.php
```diff
@@ -136,21 +136,24 @@
...
+
         // if the config options have changed or the component is called for the first time,
         // (re-)initialize the parser object
         if ($config !== $options) {
             $markdown = new Markdown($options);
             $config   = $options;
         }
 
-        return $markdown->parse($text, $inline);
+        return $markdown->parse($text, $options['inline'] ?? false);
```
afbora: No need null coalescing operator, because of `$options['inline']` already set.

```suggestion
        return $markdown->parse($text, $options['inline']);
```

distantnative:
There are still a few spots where parameter gets renamed or missing todo. Should I push to this PR?

distantnative (dismissed):
LGTM

tests/Cms/App/AppComponentsTest.php
```diff
@@ -76,6 +76,46 @@ public function testJsPlugin()
...
+        $text     = 'Test';
+        $expected = '<p>Test</p>';
+
+        $this->assertEquals($expected, $this->kirby->kirbytext($text));
+    }
+
+    public function testKirbytextWithSafeMode()
+    {
+        $text     = '<h1>Test</h1>';
+        $expected = '&lt;h1&gt;Test&lt;/h1&gt;';
```
lukasbestle: To make sure that Markdown is still parsed:

```suggestion
        $text     = '<h1>**Test**</h1>';
        $expected = '&lt;h1&gt;<strong>Test</strong>&lt;/h1&gt;';
```

lukasbestle (commented):
I love how clean the method signatures become with the `$options` arrays. Fully 👍 on the deprecation plan to avoid immediate breaking changes.

Two minor comments about the tests:
- The safe mode tests should still test Markdown parsing, see below. This also affects the other occurrences of this test string.
- In some new tests `assertEquals()` is used. For strings there is no difference, but it would be cleaner with `assertSame()` to keep it consistent.

None of these need to be done before the release if there is no time. The code itself looks great!

PR #4099, opened 2022-01-18
Allow `rel="alternate stylesheet"` in css() helper

sebastiangreger:
## Describe the PR
The `css()` helper currently does not allow setting the `rel` attribute to `alternate stylesheet`; always defaults to `stylesheet`.
- via https://forum.getkirby.com/t/alternate-stylesheet-with-css-helper-not-possible/24510/3
- MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/Alternative_style_sheets

This fix enables setting `rel="alternate stylesheet"`, but no other `rel` values, and only if the `title` attribute contains a valid string.

NB. This PR has a tiny loophole (actually might be a useful feature, even?) if the helper is used with an array of CSS paths; they would all receive the same `rel` and `title` attributes. Since that is still valid HTML (browsers simply ignore what does not make sense), I don't see a direct need to prevent that in the helper code?

### Fixes
`css()` helper now accepts `rel="alternate stylesheet"`

## Breaking changes
None

## Ready?
- [x] CI checks pass

## When merging

### DIFF
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -115,13 +115,13 @@ function css($url, $options = null): ?string
         }
     }
 
-	// only valid value for 'rel' is 'alternate stylesheet', if 'title' is given as well
+    // only valid value for 'rel' is 'alternate stylesheet', if 'title' is given as well
     if (
-		($options['rel'] ?? '') !== 'alternate stylesheet' ||
-		($options['title'] ?? '') === ''
-	) {
-		$options['rel'] = 'stylesheet';
-	}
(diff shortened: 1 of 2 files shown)


lukasbestle (approved):
Thank you very much for your contribution and the added unit tests. 💛

Since the solution is so simple and doesn't break existing behavior, I think we can safely include it in Kirby 3.6.2.

Regarding the loophole with an array of paths: I agree with you that it isn't an issue per se. The expected usage should use just one path per call of the `css()` helper, each time with a unique `title` value. But there is no need to prevent other usage as it would only make the implementation more complex.

PR #4106, opened 2022-01-23
Add JPEG size hint to ImageMagick command

### LINKED ISSUE #3023: IM thumb engine fails at processing large image files
Kirby 3.4.3 (maybe 3.4.x) reintroduced the issue #1788 and reported by @silllli

silllli:
## Describe the PR

One step in enhancing Kirby’s thumbnail creation, especially with large images in resource-restricted server environments. It adds a “size hint” (see [ImageMagick docs](https://imagemagick.org/script/defines.php)) when calling ImageMagick with a JPEG file which can reduce the CPU and memory usage drastically—most notably when the output file is a lot smaller than the original.

### Enhancements
* JPEG size hint added to the ImageMagick thumb driver to increase performance with large source images

## Breaking changes

None

## Related issues/ideas

- Partly fixes #3023 

## Ready?

- [ ] ~Unit tests for fixed bug/feature~ (hard if not impossible to test because it's an internal working of IM)
- [x] CI checks pass

## When merging

- [ ] ~Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)~

### DIFF
--- a/src/Image/Darkroom/ImageMagick.php
+++ b/src/Image/Darkroom/ImageMagick.php
@@ -5,6 +5,7 @@ namespace Kirby\Image\Darkroom;
 use Exception;
 use Kirby\Filesystem\F;
 use Kirby\Image\Darkroom;
+use Kirby\Image\Image;
 
 /**
  * ImageMagick
@@ -73,6 +74,16 @@ class ImageMagick extends Darkroom
 
         // limit to single-threading to keep CPU usage sane
         $command .= ' -limit thread 1';
+
+        // add JPEG size hint to optimize CPU and memory usage
+        if (F::mime($file) === 'image/jpeg') {
+            $image = new Image($file);
+
+            // add hint only when downscaling
+            if ($options['width'] < $image->width() && $options['height'] < $image->height()) {
+                $command .= sprintf(' -define jpeg:size=%dx%d', $options['width'], $options['height']);
+            }
+        }
 
         // append input file
         return $command . ' ' . escapeshellarg($file);
--- a/src/Image/Darkroom/ImageMagick.php
+++ b/src/Image/Darkroom/ImageMagick.php
@@ -5,7 +5,6 @@ namespace Kirby\Image\Darkroom;
 use Exception;
 use Kirby\Filesystem\F;
 use Kirby\Image\Darkroom;
-use Kirby\Image\Image;
 
 /**
  * ImageMagick
@@ -77,10 +76,8 @@ class ImageMagick extends Darkroom
 
         // add JPEG size hint to optimize CPU and memory usage
         if (F::mime($file) === 'image/jpeg') {
-            $image = new Image($file);
-
             // add hint only when downscaling
-            if ($options['width'] < $image->width() && $options['height'] < $image->height()) {
+            if ($options['scaleWidth'] < 1 && $options['scaleHeight'] < 1) {
                 $command .= ' -define ' . escapeshellarg(sprintf('jpeg:size=%dx%d', $options['width'], $options['height']));
             }
         }
--- a/src/Image/Darkroom/ImageMagick.php
+++ b/src/Image/Darkroom/ImageMagick.php
@@ -81,7 +81,7 @@ class ImageMagick extends Darkroom
 
             // add hint only when downscaling
             if ($options['width'] < $image->width() && $options['height'] < $image->height()) {
-                $command .= sprintf(' -define jpeg:size=%dx%d', $options['width'], $options['height']);
+                $command .= ' -define ' . escapeshellarg(sprintf('jpeg:size=%dx%d', $options['width'], $options['height']));
             }
         }
 
--- a/src/Image/Darkroom.php
+++ b/src/Image/Darkroom.php
@@ -129,16 +129,18 @@ class Darkroom
         $options = $this->options($options);
         $image   = new Image($file);
 
+        $dimensions      = $image->dimensions();
+        $thumbDimensions = $dimensions->thumb($options);
+
         $sourceWidth  = $image->width();
         $sourceHeight = $image->height();
 
-        $dimensions = $image->dimensions()->thumb($options);
+        $options['width']  = $thumbDimensions->width();
+        $options['height'] = $thumbDimensions->height();
 
-        $options['width']  = $dimensions->width();
-        $options['height'] = $dimensions->height();
-
-        $options['scaleWidth']  = $options['width'] / $sourceWidth;
-        $options['scaleHeight'] = $options['height'] / $sourceHeight;
+        // scale ratio compared to the source dimensions
+        $options['scaleWidth']  = $sourceWidth ? $options['width'] / $sourceWidth : null;
+        $options['scaleHeight'] = $sourceHeight ? $options['height'] / $sourceHeight : null;
 
         return $options;
     }
--- a/src/Image/Darkroom.php
+++ b/src/Image/Darkroom.php
@@ -63,14 +63,16 @@ class Darkroom
     protected function defaults(): array
     {
         return [
-            'autoOrient' => true,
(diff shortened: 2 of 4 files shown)


src/Image/Darkroom/ImageMagick.php
```diff
@@ -69,7 +70,21 @@ protected function coalesce(string $file, array $options)
...
-        return sprintf($options['bin'] . ' "%s"', $file);
+        $command = $options['bin'];
+
+        // add size hint to optimize CPU and memory usage
+        if (F::mime($file) === 'image/jpeg') {
+            $image = new Image($file);
+
+            // add hint only when downscaling
+            if ($options['width'] < $image->width() && $options['height'] < $image->height()) {
+                $command .= sprintf(' -define jpeg:size=%dx%d', $options['width'], $options['height']);
```
lukasbestle:
```suggestion
                $command .= ' -define ' . escapeshellarg(sprintf('jpeg:size=%dx%d', $options['width'], $options['height']));
```

lukasbestle (commented):
Thank you for your contribution, the change makes a lot of sense. 👍

Unfortunately I saw your PR after I created #4109 from your code comment. The two PRs conflict due to changes in the same file, so I suggest we first wait until the other PR is merged. I'll then rebase this PR.

I have two suggestions for your code:

- To increase performance, the fact that the image is being scaled down could be stored in the `$options` array in `Image\Darkroom::preprocess()` as this method already has the `$image` object at hand. There could be new float `$option['scaleWidth']` and `$option['scaleHeight']` options for this (defined as the scale ratio between the thumb and the original width and height). These options should be set to `null` by default in the `defaults()` method.
- The new command arguments should use `escapeshellarg()`, see below.

silllli:
> Unfortunately I saw your PR after I created https://github.com/getkirby/kirby/pull/4109 from your code comment. The two PRs conflict due to changes in the same file, so I suggest we first wait until the other PR is merged. I'll then rebase this PR.

No problem, could you rebase the PR now?

Also: as I have suggested in the issue, I have some more resource optimization ideas. Should I add them in this PR or should I stick to the size hint with this one and create another one for the memory `-limit` option?

src/Image/Darkroom/ImageMagick.php
silllli: > Unfortunately I saw your PR after I created https://github.com/getkirby/kirby/pull/4109 from your code comment. The two PRs conflict due to changes in the same file, so I suggest we first wait until the other PR is merged. I'll then rebase this PR.

No problem, could you rebase the PR now?

Also: as I have suggested in the issue, I have some more resource optimization ideas. Should I add them in this PR or should I stick to the size hint with this one and create another one for the memory `-limit` option?

distantnative: @silllli I don't think we can rebase this branch cause it's based on your fork. You could rebase it from `develop` branch or I copy your branch etc. and rebase it - but this would then require closing this PR and open a new one from the copied branch.

And I think separate PR would be helpful. Always easier to review one thing at a time.

distantnative: @silllli which one do you prefer (you rebasing your PR or me opening a new PR)?

silllli: I will rebase it 👍

silllli: I’ve rebased the commit and added @lukasbestle’s suggestions.

(By the way I noticed that `preprocess()` is called twice during thumb generation. Not sure if that is intended and / or needed or not.)

silllli: Oops, I’m not a git genius. All these commits that happened in the meantime shouldn’t be part of the PR I guess. I will try to fix it.

silllli: Seems to look good now, after some git black magic…

lukasbestle (dismissed):
Thank you very much for your work and sorry for the huge delay. In the last weeks I was busy with other things and only now came around to review your PR again.

I like the implementation and think it's already a great improvement.

Regarding `preprocess()` being called multiple times: One time it is called from the `thumb` component to determine the options to build the filename, and then the driver itself also calls it. With the current setup there is no way around that as the options are not cached inside `$darkroom`. But it could be improved in a later step.

lukasbestle:
Oh, the unit tests fail. I will fix them tomorrow evening.

mrflix:
@silllli I'm a bit late with commenting here: the ImageMagick docs suggest that the `jpg:size` dimensions should be double the size of the final image to prevent antialiasing effects:

> Note that this modifier causes the JPEG library to skip the reading of whole columns and rows of pixels. As such it will produce effects much like the [Sampling Resize Operator](https://imagemagick.org/Usage/resize/#sample), including its strong [Aliasing Artefacts](https://imagemagick.org/Usage/filter/#aliasing).
**Because of this is it recommended that you specify at least double the final 'resize' of the image, to avoid this problem, just as shown in the above example.**

https://imagemagick.org/Usage/formats/#jpg_read

silllli:
> @silllli I'm a bit late with commenting here: the ImageMagick docs suggest that the `jpg:size` dimensions should be double the size of the final image to prevent antialiasing effects:
> 
> > Note that this modifier causes the JPEG library to skip the reading of whole columns and rows of pixels. As such it will produce effects much like the [Sampling Resize Operator](https://imagemagick.org/Usage/resize/#sample), including its strong [Aliasing Artefacts](https://imagemagick.org/Usage/filter/#aliasing).
> > **Because of this is it recommended that you specify at least double the final 'resize' of the image, to avoid this problem, just as shown in the above example.**
> 
> https://imagemagick.org/Usage/formats/#jpg_read

I think I remember reading that and tested it myself with some images where I didn’t notice the effect so I dismissed it. I also remember the IM docs as seeming partially outdated so I figured it might not be a problem anymore. I had not seen their examples where the effect is visible. Sorry!

mrflix:
Don't say sorry! You might as well be right. I think the examples are examples of the effect itself (sampling & antialiasing), not results of the `jpeg:size` setting. I tried to understand it's [ImageMagick code](https://github.com/ImageMagick/ImageMagick/blob/acaffeb8e496ab23f6b56f3411ff8e69c7ff2fe3/coders/jpeg.c#L1201), but I don't fully understand it. In the docs it says:
> Remember this is only a hint as the size of the image wanted, you are not guaranteed to get this size, just something close-to but larger than that size. **Typically you will get something that is between this size and one twice that big, while preserving the images aspect ratio.** 

[Source](https://imagemagick.org/Usage/formats/#jpg_read)

This might have to do with libjpeg, which is used under the hood, only supporting scaling down the image by certain factors:
```
unsigned int scale_num, scale_denom
        Scale the image by the fraction scale_num/scale_denom.  Default is
        1/1, or no scaling.  Currently, the only supported scaling ratios
        are M/8 with all M from 1 to 16, or any reduced fraction thereof (such
        as 1/2, 3/4, etc.)  (The library design allows for arbitrary
        scaling ratios but this is not likely to be implemented any time soon.)
        Smaller scaling ratios permit significantly faster decoding since
        fewer pixels need be processed and a simpler IDCT method can be used.
```
[Source](https://github.com/winlibs/libjpeg/blob/6904e8cd02785f4faa500714d4d1e150a3497200/libjpeg.txt#L1224)

[Here](https://stackoverflow.com/a/9878900/731172) it says that that's true for libjpeg 8c+ and older libjpeg versions only support coarser downscaling.

My example image in #5855 has the dimensions 3543x2362. I want to crop it to 512x683. With the setting `-define 'jpeg:size=512x683'`, ImageMagick calculates a scaling factor of 0.2891617273. Which is the longer side of the crop divided by the shorter side of the image 683 / 2362. So libjpeg will use a factor of 3/8 = 0.375. Maybe that's good enough.

PR #4108, opened 2022-01-23
Vitest optimizations and CI integration

lukasbestle:
## Describe the PR

- Vue development mode warnings are silenced during testing
- Test methods are imported explicitly according to the Vitest docs
- Coverage is now collected in CI, but not submitted to Codecov yet due to the following issues:
  - `.vue` files are for some reason not included in the coverage report even though they are added to the list in the config. Seems to be a bug in Vitest or `c8`. Peeky was able to collect coverage from Vue files.
  - The covered/uncovered lines are incorrect due to a bug in Vitest (vitest-dev/vitest#375).
  - Our frontend tests are not complete enough to make coverage a valuable metric yet.

## Release notes

None (Vitest is already included in 3.6.2 release notes)

## Breaking changes

None

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,8 +125,16 @@ jobs:
 
       - name: Run JavaScript unit tests
         if: always() && steps.finishPrepare.outcome == 'success'
-        run: npm run test:unit
+        run: npm run test:unit:coverage
         working-directory: panel
+      
+      - name: Upload coverage results to Codecov
+        uses: codecov/codecov-action@f32b3a3741e1053eb607407145bc9619351dc93b # pin@v2
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }} # for better reliability if the GitHub API is down
+          fail_ci_if_error: true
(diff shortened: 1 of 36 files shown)


distantnative:
Do the reported uncovered lines make sense for you? When I tried to setup coverage collection some days ago, I stopped cause the c8 reports actually didn't make sense (pointing to lines with just comments etc.).

distantnative:
Looks like the same problem here, e.g. `panel/src/helpers/string.js` is showing line 233 as uncovered. Line 233 is simply

```
/**
```

distantnative:
Potentially relates to https://github.com/vitest-dev/vitest/issues/375

lukasbestle:
You are right, coverage support doesn't seem to be useful for us yet. I've prepared everything so we can easily switch it on when Vitest and we are ready. But for now I have disabled sending the report to Codecov from CI.

The remaining enhancements (especially the warning silencing) are still useful immediately.

distantnative (approved):

Issue #4129, opened 2022-02-03
[Panel] Cannot copy multiple blocks


rasteiner:
## Description
The button to copy multiple blocks does not work when an input element is focused. 

**Expected behavior**  
The focused input should "blur" before the copy operation is tried or the copy operation should work "anyway" when started from the button click.

## To reproduce
1. In a blocks field, create at least 1 text block
2. Select the block via Ctrl/Cmd + Click
3. Notice that the writer instance in the text block did not loose focus
4. Click on the copy button

Nothing is copied.

## Your setup
**Kirby Version** : 3.6.2, but also previous versions. 

**Console output**  
No console output

**Your system (please complete the following information)**
 - Device: PC
 - OS: Win 10
 - Browser: Chrome
 - Version: 97

**Additional context**  
The execution of the copy function in "Blocks.vue" is halted here:
```js
// don't copy if an input is focused
if (this.isInputEvent(e) === true) {
  return false;
}
```

## Quick fix
In **BlockOptions.vue**, the click on the copy (and also delete) button is handled by a `mousedown` event. This event fires before the browser has a chance to change focus from Input element to Button. As I understand it, the `mousedown` event was chosen, instead of a normal `click` event, because there is also a global "onOutsideFocus" that fires when a button is clicked that "deselects the blocks", and therefore the copy event has to fire before this "focus" event. 
The quickest fix would be to manually `blur()` the `activeElement` when the copy button receives "mouseDown".

distantnative:
@rasteiner could this actually be related to https://github.com/getkirby/kirby/issues/3941?

rasteiner:
> @rasteiner could this actually be related to #3941?

Oh haven't see that. Yes, it looks like the same issue

bastianallgeier:
✅

PR #4130, opened 2022-02-03
Blur activeElement before trying copy

rasteiner:
## Describe the PR
Makes it so that blocks can be copied via the dedicated Copy button, even when an input element has a focus. It achieves this by removing the focus from the activeElement before emitting the "copy" event, just like normal button presses would do, but before. 

## Breaking changes
None

## Related issues/ideas
- Fixes https://github.com/getkirby/kirby/issues/4129

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/panel/src/components/Forms/Blocks/BlockOptions.vue
+++ b/panel/src/components/Forms/Blocks/BlockOptions.vue
@@ -5,7 +5,7 @@
         :tooltip="$t('copy')"
         class="k-block-options-button"
         icon="template"
-        @mousedown.native.prevent="$emit('copy')"
+        @mousedown.native.prevent="copy()"
       />
       <k-button
         :tooltip="$t('remove')"
@@ -114,6 +114,16 @@ export default {
   methods: {
     open() {
       this.$refs.options.open();
+    },
+    blurActiveElement() {
+      const activeElement = document.activeElement;
+      if (activeElement && typeof activeElement.blur === "function") {
+        activeElement.blur();
+      }
+    },
+    copy() {
+      this.blurActiveElement();
+      this.$emit("copy");
     }
   }
 };


distantnative:
Hi @rasteiner, thanks for tying to fix this.
I was just looking at your fix. I don't think handling the blur should be done in `BlockOptions` though. As the actual is also not happening there, we just pass an event up, that the button was clicked. The actual copying is happening in `Blocks`, so that's probably where we should be focusing on (otherwise trying to understand the code later on is becoming really confusing, if arts of the logic are happening in different places).

But when I look at `Blocks`, it seems like copying was deliberately disabled when focusing an input: https://github.com/getkirby/kirby/blob/main/panel/src/components/Forms/Blocks/Blocks.vue#L290-L293

@bastianallgeier do you recall why you added these highlighted lines?

distantnative:
@rasteiner - sorry only now saw that in the issue you point to those lines yourself.

distantnative:
Ok, so my suspicion why we have that check in there, is that the `copy()` method can also be triggered by a keyboard event. And of course we want to still allow copying e.g. text from a text input. That's why we check there. 

But then my suggestion would be to try if we can check in `isInputEvent()` in `Blocks.vue` whether the event is a click-triggered event (which would return false).

rasteiner:
That's another approach that could work (I actually first solved it in a very similar way).

But then, my logic of putting this into the BlockOptions was to replicate what a normal button would do. If you click a button, it normally receives focus: which means "blur()" then "focus()". But since the button seems to need to work on "mouseDown", I anticipate the blur a moment. 

There's actually also a problem with the delete button (see other issue https://github.com/getkirby/kirby/issues/4131). But that seems to be more related to the general functioning of "onOutsideFocus" which would probably also solve this issue when fixed (since we could then simply use a click handler in BlockOptions that would do the blurring part automatically).

rasteiner:
Superseded by https://github.com/getkirby/kirby/pull/4132

PR #4135, opened 2022-02-04
Fix date JS unit tests

distantnative:
- Upgrades `vitest`
- Uses a pinned date (15 Jan 2022) for date interpret tests (as otherwise tests would fail every new month)


bastianallgeier (approved):
Uhhh, nice hack :)

PR #4144, opened 2022-02-09
Use headers from global response for `Response` objects

### LINKED ISSUE #4101: Panel shows old content after re-opening tab
## Description
This is a bit of an odd one, but I'll try to explain as good as I can. I first noticed this with the MarkdownField plugin, but ran some tests and it can be replicated with the Textarea field as well.

When editing text in a textarea/markdown field and saving, then closing the browser tab (Ctrl+W) and reopening it with Ctrl+Shift+T, the textarea/markdown field shows an outdated version of the content (the version that was initially loaded when first visiting that Panel page).

I have also been able to replicate this by visiting another site like `http://example.com` in the same tab and then going back with the browser's "back" button, but that has not been quite so consistent to replicate (not sure what affects it).

**Expected behavior**  
When closing and reopening a tab with the panel (or coming back to a Panel URL using the browser's "back" button after opening another URL in the same tab), I'd expect to see the content as I last saved it.

## To reproduce
1. Create a page blueprint that contains a textarea field and set up a pages section for that page type
2. Use "Add" in the panel to create a new page of that blueprint
3. Write `My first line.` into the textarea
4. Save with Ctrl+S
5. Close the browser tab with Ctrl+W
6. Reopen the tab with Ctrl+Shift+T
7. Now write `My second line.` into the textarea, right below the first
8. Save with Ctrl+S
9. Close the browser tab with Ctrl+W
10. Reopen the tab with Ctrl+Shift+T
11. See error: the second line is not shown, even though the page had been saved before closing the tab; there is no indication in the UI that this is an outdated version
12. Reloading the Panel page with F5 reveals the content as saved, i.e. with both lines of text

## Your setup
3.6.2-rc.1 (issue exists with 3.6.1 as well, but I am unable to say whether this is a new thing since 3.6.x or if I simply never noticed before)

**Console output**  
Nope.

 - Device: PC
 - OS: Linux Mint
 - Browser: Firefox
 - Version: 96.0

lukasbestle:
## Describe the PR

API responses with an active session depend on the user who requested them. For better reliability and also security, intermediary caches must not cache API responses.

`$kirby->session()` already sets a `Cache-Control: no-store` header for dynamic responses, but that was so far not respected by the API (which returns its own custom `Response` object).

### Bug fixes

- API responses now contain the `Cache-Control: no-store` header like other responses with an active session.
- The globally configured headers from `$kirby->response()` are now also respected when a route returns a full `Response` object.

## Breaking changes

None

## Related issues/ideas

- Fixes https://forum.getkirby.com/t/owasp-criteria-auth-rest-api/24690.
- Fixes #4101.

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -10,6 +10,7 @@ use Kirby\Exception\NotFoundException;
 use Kirby\Filesystem\Dir;
 use Kirby\Filesystem\F;
 use Kirby\Http\Request;
+use Kirby\Http\Response;
 use Kirby\Http\Router;
 use Kirby\Http\Server;
 use Kirby\Http\Uri;
@@ -709,14 +710,25 @@ class App
             return $this->io(new NotFoundException());
         }
 
-        // Response Configuration
+        // (Modified) global response configuration, e.g. in routes
         if (is_a($input, 'Kirby\Cms\Responder') === true) {
+            // return the passed object unmodified (without injecting headers
+            // from the global object) to allow a complete response override
+            // https://github.com/getkirby/kirby/pull/4144#issuecomment-1034766726
             return $input->send();
         }
 
         // Responses
         if (is_a($input, 'Kirby\Http\Response') === true) {
-            return $input;
+            $data = $input->toArray();
+
+            // inject headers from the global response configuration
+            // lazily (only if they are not already set);
+            // the case-insensitive nature of headers will be
+            // handled by PHP's `header()` function
+            $data['headers'] = array_merge($response->headers(), $data['headers']);
+
+            return new Response($data);
         }
 
         // Pages
--- a/tests/Cms/App/AppIoTest.php
+++ b/tests/Cms/App/AppIoTest.php
@@ -79,12 +79,30 @@ class AppIoTest extends TestCase
     {
         $input = new Response([
             'code' => 200,
-            'body' => 'Test'
+            'body' => 'Test',
+            'type' => 'text/plain',
+            'headers' => [
+                'X-Foo'  => 'Foobar',
+                'X-Test' => 'Test'
+            ]
         ]);
 
-        $response = $this->app()->io($input);
+        $app = $this->app();
+        $app->response()->header('Cache-Control', 'no-cache');
+        $app->response()->header('X-Foo', 'Bar');
+        $response = $app->io($input);
 
-        $this->assertEquals($input, $response);
+        $this->assertSame([
+            'type' => 'text/plain',
+            'charset' => 'UTF-8',
+            'code' => 200,
+            'headers' => [
+                'Cache-Control' => 'no-cache',
+                'X-Foo'         => 'Foobar',
+                'X-Test'        => 'Test'
+            ],
+            'body' => 'Test',
+        ], $response->toArray());
     }
 
     public function testPage()


afbora:
@lukasbestle I just checked and https://github.com/getkirby/kirby/issues/4101 issue seems to be resolved as well 😅 🎉

lukasbestle:
Looks like Fiber also uses `Response` objects internally.

---

I've been wondering about the handling for `Responder` (`$kirby->response()`) objects:

https://github.com/getkirby/kirby/blob/04127160ed1ab8dc277763cd8ba4ebc5a097bc18/src/Cms/App.php#L700-L703

In this case we also don't respect the global `Responder` object and just use the provided one. Unfortunately I haven't been able to track down where we actually return `Responder` from routes to check if there would be breaking changes.

distantnative:
@lukasbestle but doesn't make this sense (line 702 as is)? When the input is an actual `Kirby\Cms\Responder` object provided, shouldn't we use this as is and not modify it to the global `Kirby\Cms\Responder` configuration? Otherwise one never could provide a `Kirby\Cms\Responder` that differs from the global one.

lukasbestle:
@distantnative That is the question. Have you found where we currently use `Responder`?

distantnative:
@lukasbestle I don't think we do anywhere but it would be a breaking change if we don't support this (passing a `Responder` object that stays unmodified) anymore, wouldn't it?

afbora:
I may not have fully understood the situation. But I just wanted to point it out so I wouldn't miss it. `$kirby->response()` is used in 7 places of the routes. Were you looking for these? 🙈 

Specifically, I tried this and verified that line 702 in the `io()` method is running:
https://github.com/getkirby/kirby/blob/3.6.2/config/routes.php#L49-L60

Others:
https://github.com/getkirby/kirby/blob/3.6.2/config/routes.php#L122-L131
https://github.com/getkirby/kirby/blob/3.6.2/src/Cms/LanguageRoutes.php#L88-L90
https://github.com/getkirby/kirby/blob/3.6.2/src/Cms/LanguageRoutes.php#L93-L95
https://github.com/getkirby/kirby/blob/3.6.2/src/Cms/LanguageRoutes.php#L125-L127
https://github.com/getkirby/kirby/blob/3.6.2/src/Cms/LanguageRoutes.php#L140-L142
https://github.com/getkirby/kirby/blob/3.6.2/src/Cms/LanguageRoutes.php#L145-L147

distantnative:
@afbora I think you found the right places! I was thinking in the wrong direction.

lukasbestle:
Ah, yes. Thanks Ahmet! In these cases we already use the global object as the base, so it's fine.

Maybe we can get away with leaving this case as it is for now. You are right that changing it would be a breaking change.

It's IMO a bit confusing that Response and Responder are treated differently now, but that's just how it is. Maybe we can add a TODO comment to reconsider this for 3.7 with a link to this discussion.

afbora:
How do we plan to move forward in this PR?

distantnative:
@lukasbestle If I understand it correctly, this PR only misses the in-code TODO comment?

lukasbestle:
Two updates:

1. I thought about the `Responder` value and its inconsistency to `Response` again. I think @distantnative is right that it must be possible to pass a response object that's entirely customized without magic header injection. So I added a comment that explains our thoughts for our future selves.
2. I've added a deprecation comment for `Response::charset()`. I believe it's not needed anymore. We should hardcode it to `UTF-8` in the future (maybe 3.8.0).

## Open question

Do we risk the breaking change regarding `Response::charset()`? Thoughts:

- I haven't found a single place in the core where we use it. So it would be only relevant for plugins and custom site routes.
- Frankly I don't think anyone uses it.
- The change means that the charset will be ignored. Kirby will always use `UTF-8`. So *if* anyone uses it, it could break their custom responses unless they are `UTF-8` anyway.

If we don't want to risk it, we need to move this PR to 3.7.0.

distantnative:
@lukasbestle is the charset deprecation related to the other changes? otherwise it'd be easier to move that to a separate PR to decide.

lukasbestle:
@distantnative The `Responder` doesn't support the `charset`, so we cannot carry it through with the current setup.

But I think we can actually modify it to use `Response` instead. Will try it out.

lukasbestle:
@distantnative Your question was really useful. I modified the code so that it no longer introduces a breaking change.

The PR is now ready for review and can be merged into 3.6.5 IMO.

src/Cms/App.php
```diff
@@ -709,14 +710,25 @@ public function io($input)
...
         // Responses
         if (is_a($input, 'Kirby\Http\Response') === true) {
-            return $input;
+            $data = $input->toArray();
+
+            // inject headers from the global response configuration
+            // lazily (only if they are not already set);
+            // the case-insensitive nature of headers will be
+            // handled by PHP's `header()` function
+            $data['headers'] = array_merge($response->headers(), $data['headers']);
```
bastianallgeier, line 729: All of this looks really good, although I would probably prefer an official way to modify a response or just the headers of a response which feels less brute-force. Something like `$response->headers($data['headers'])`

bastianallgeier (approved):

src/Cms/App.php
bastianallgeier: I think we can fix this later. It's still great for the release to have this solution.

lukasbestle: @bastianallgeier The `Response` class is currently immutable. We could change this, but I'm not sure which way is the most future-proof. Our classes are a bit inconsistent about mutability. Some we explicitly made immutable and with some (like `Response`) I'm not sure if it was a deliberate decision.

PR #4150, opened 2022-02-10
Update GitHub actions for CI

afbora:


.github/workflows/ci.yml
```diff
@@ -301,7 +301,7 @@ jobs:
 
       - name: Check for unused Composer dependencies
         if: always() && steps.finishPrepare.outcome == 'success'
-        run: composer unused --no-progress
+        run: composer-unused --no-progress
```
afbora, line 304: No longer works as `composer unused`. Please see the log:

https://github.com/getkirby/kirby/runs/5146628141?check_suite_focus=true

afbora: Also I tried v8 but failed, see logs:
https://github.com/getkirby/kirby/runs/5146326654?check_suite_focus=true

distantnative: But seems to work with your change, right?

afbora: Yes works with current implementation but need to review from @lukasbestle to confirm

lukasbestle (approved):

PR #4151, opened 2022-02-12
Split Github workflows

distantnative:
Based on https://github.com/getkirby/kirby/pull/4150

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -1,9 +1,25 @@
-name: CI
-on: [push, pull_request]
+name: CI - Backend
+on:
+  push:
+    paths:
+      - "!.tx/**"
+      - "!.vscode/**"
+      - "!assets/**"
+      - "!i18n/**"
+      - "!panel/**"
+      - "!scripts/**"
+  pull_request:
+    paths:
+      - "!.tx/**"
+      - "!.vscode/**"
+      - "!assets/**"
+      - "!i18n/**"
+      - "!panel/**"
+      - "!scripts/**"
 
 jobs:
-  backend-tests:
-    name: "Backend: Unit tests - PHP ${{ matrix.php }}"
+  tests:
+    name: "Unit tests - PHP ${{ matrix.php }}"
 
     # if on pull request, only run if from a fork
     # (our own repo is covered by the push event)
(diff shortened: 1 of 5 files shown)


lukasbestle (commented):
Works now. :tada:

Seems like the workflows didn't run because there was no change to any file that was mentioned. I have added the workflows to their own lists so that changes to the workflows always trigger them.

General first thoughts:

- I like the [checks view](https://github.com/getkirby/kirby/pull/4151/checks) with the clear separation of backend, frontend and integration while running. However after the run is complete, the sections are collapsed and need to be opened individually. A bit more cumbersome, but fine IMO.
- The list of checks in the "All checks have passed" view below the PR comments seems to be in completely random order and hard to keep the overview.
- Without required checks we might miss checks that don't run even though they should. This applies both to our CI checks and also to Codecov and other external checks we may have in the future.
- With the previous point in mind, we IMO need to ensure that all checks are run before a release, otherwise we might miss issues. We can still leave all checks `required` for the `main` branch, but how would we run them in the `main` PR? Is there a way to make GitHub ignore the `paths` in PRs against specific branches (here: `main`)? Or would we need to run all checks manually by trigger?

distantnative:
Thanks for your thoughts. Tricky indeed with the required checks.

> we IMO need to ensure that all checks are run before a release, otherwise we might miss issues

Given that we should assume that `main` has a version where all tests passed. Any PR or push to `develop` would be covered by all checks that apply to the changes (depending on what kind of file changed). Which means that even if at one point in time all tests ran together, nothing should've been changed where the respective checks didn't run or pass.

lukasbestle (dismissed):
You are right. As the three CI workflows don't depend on changes to files they don't run on, this is a non-issue.
I've updated the path matching rules again, they should now cover everything they should but nothing they shouldn't.

The only question left is what to do with the required checks. I think it depends on the type of check:

- With the GitHub Actions checks it's quite obvious. I think we can safely assume that GitHub correctly processes the path rules we define. If it didn't, then it would not be any different from the previous situation (we also have to trust that the displayed test result matches the test output for example). So the only issue would be if we forgot to include paths in our config or if we configured them incorrectly (with GitHub not parsing them as we intended). I think the risk is low. So all in all we are pretty safe with CI jobs – if one doesn't run that should run, we will see that because GitHub will display the job as running/stuck.
- With Codecov and other external tools we may use in the future it's different. The checks are only displayed by GitHub after they were added by Codecov. So if Codecov doesn't run for whatever reason, we might never realize that unless we manually verify that each time.

In the end it's a trade-off between safety/reliability and performance. The CI jobs are much more important than Codecov. So I think it can be a valid trade-off to make everything considered.

@distantnative Would be great if you could take a look at the changes again and think about the risk with getting rid of required checks. If you agree with me and everything looks good, I think we can finally merge this.

lukasbestle (approved):

PR #4153, opened 2022-02-12
Update PR template & contributing guidelines

distantnative:
Boiling down out PR template a little

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,12 +1,12 @@
 ## This PR …
 <!--
-A clear and concise description of the PR introduces.
-Use this section for review hints and explanations or discussion points/todos.
+A clear and concise description of the PR.
+Use this section for review hints, explanations or discussion points/todos.
 
-Add the sections you need: Features, Enhancements, Fixes, Deprecated.
+Add releant release notes: Features, Enhancements, Fixes, Deprecated.
 Reference issues from the `kirby` repo or ideas from `feedback.getkirby.com`.
 
-Always mention if PR introduces breaking changes.
+Always mention whether PR introduces breaking changes.
 -->
 
 ### Fixes
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,8 +5,9 @@ Use this section for review hints, explanations or discussion points/todos.
 
 Add releant release notes: Features, Enhancements, Fixes, Deprecated.
 Reference issues from the `kirby` repo or ideas from `feedback.getkirby.com`.
-
 Always mention whether PR introduces breaking changes.
+
+How to contribute: https://contribute.getkirby.com
 -->
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


bastianallgeier (commented):
I think we could boil it down even some more and remove the related issues part. I think issues should be mentioned directly in the description line. 

I.e: 

Fixes: 

- This fixes a b and c #some-issue

distantnative:
Changed as suggested:
Screen Shot 2022-02-15 at 19 49 21

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,51 +1,36 @@
-## Describe the PR
+## This PR …
 <!--
-A clear and concise description of the bug the PR fixes or the feature the PR introduces.
-Use this section for review hints and explanations for easier code understanding if necessary.
--->
-
+A clear and concise description of the PR introduces.
```
nilshoerrmann: Something is wrong with this sentence? Shouldn't this just read "A clear and concise description of the PR."?

distantnative: Good catch.

nilshoerrmann:
As far as I understand this template, it is directed at people contributing for the first time or irregularly so that they know all the requirements. If you want to help outside contributors, it would be great to have some more explainations of the main checklist. This could all be briefly be described on the website with the template linking to it:

> Unit tests for fixed bug/feature

If I'm new to this, I'd need an explainations what unit test are and how the team uses them. There was this great thread on Discord where you explained how to use PHPUnit in very few very simple steps. All contributors who were able to create a pull request should also be able to write a test – they just need to know how (still, many don't).

Are there different requirements for the PHP and JavaScript sides?

I usually find it easier to understand these things by example. And while it will sure be obvious to the advanced user where to find the existing tests, a link to one examplary test in this repo would be really helpful to showcase how to set them up (and how not to). This might sound negligible but this can really make you more confident in what you are doing if this is your first PR.

> In-code documentation (wherever needed)

Documentation is always important but I've seen many different approaches to it. It would be great to explain how you understand "wherever needed":

- Should I always add an explaination why a change was made, maybe linking to the related issue? 
- Or should I only add comments for advanced things like complex regular expressions and such? 
- Should I comment line by line or block by block? 
- Should I add links to external sources?

To speak more generally: what's your documentation policy?

distantnative:
This PR is mainly to cut down the PR template for us as we've realised the current one is a bit cumbersome for our work.

Of course, it would be great to also guide contributors with templates like these and help them. However, I don't think that your suggestions work for everyone. For some, clearly more detailed instructions, policies, standards help and are interesting. But there is also the danger of scaring new contributors away as it becomes overwhelming. Often for these, the message should rather be "try to provide up to the point where you understand and we'll try to take care of the rest".

distantnative:
I have added some info to our contributing guide - I think that's a good middle ground to give some pointers without getting too demanding/overwhelming.

texnixe (dismissed):
See https://github.com/getkirby/kirby/pull/4165

bastianallgeier (approved):

PR #4154, opened 2022-02-13
feat: add enumeration toggle to structure field

pReya:
## Describe the PR
- Add a boolean flag to the Structure field, called `enumeration` which allows to show/hide the prefixed numbers in every table row

Screenshot 2022-02-13 at 18 07 13

## Release notes
- Add a boolean flag to the Structure field, called `enumeration` which allows to show/hide the prefixed numbers in every table row


afbora (commented):
This feature hides the sorting (drag & drop). Is this a intended behavior?

pReya:
@afbora My bad, I did not take into account that the numbering column is also used as the sort handle. This makes it a lot more complicated to hide the numbering, because the sort handle still needs to be displayed somewhere.

Sorry, I thought this was a very simple endeavour – turns out, it's not.

PR #4156, opened 2022-02-14
Fix email sending via PHPMailer in PHP8.1

### LINKED ISSUE #4155: Kirby and PHPMailer differ in their $text initialisation which fails in PHP8.1
## Description

Kirby fails to send emails via PHPMailer when using PHP8.1. This is due to the [initialization of `$text` in Kirby](https://github.com/getkirby/kirby/blob/main/src/Email/Body.php) which differs from [the one in PHPMailer](https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L138).

**Expected behavior**  
Emails can be sent without a PHP error.

## To reproduce

1. Send email via form in Kirby via PHPMailer
2. Returns error `preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated`

## Your setup
**Kirby Version**  
3.6.2, same in 3.6.1.1

**Console output**  
```
Passing null to parameter #2 ($subject) of type string is deprecated" object(Uniform\Exceptions\PerformerException)#339
```

anselmh:
## Describe the PR

This fixes #4155 where a different initialization of $text in the email body in Kirby compared to PHPMailer causes emails not to be sent. It simply adjusts the initialization and since there are no unit tests affected, I didn’t need to change any. This change results in the emails being sent again.

_If there’s more missing, please either request changes or go ahead and change/commit anything you need in addition or even drop this PR if it’s not worth keeping it._

## Release notes

This fixes #4155 where a different initialization of $text in the email body in Kirby compared to PHPMailer causes emails not to be sent.

## Breaking changes

None

## Related issues/ideas

- Fixes #4155 

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Email/Body.php
+++ b/src/Email/Body.php
@@ -25,9 +25,9 @@ class Body
     protected $html;
 
     /**
-     * @var string|null
+     * @var string
      */
-    protected $text;
+    protected $text = '';
 
     /**
      * Email body constructor


src/Email/Body.php
```diff
@@ -25,9 +25,9 @@ class Body
     protected $html;
 
     /**
-     * @var string|null
+     * @var string
      */
-    protected $text;
+    protected $text = '';
```
afbora, line 30: @lukasbestle @distantnative  I think it makes more sense to make the default values of the protected attributes in the Body and Email classes inside the attribute methods, instead of making them empty strings. What do you think?

Like that:

````php
protected $text;

public function text()
{
    return $this->text ?? '';
}
````

distantnative: Maybe the easier fix indeed.

afbora: I mean like that https://github.com/getkirby/kirby/commit/59ad5a6d2c7297e2e6de729818ca48038ca68cab . Which way should we choose?

anselmh: I’ll leave that completely up to you as it’s the choice of how Kirby wants to deal with this. My suggestion is just using PHPMailer’s code method but if you’re using the one @afbora mentioned everywhere else already it may make sense to use this here, too.

afbora: Actually the problem is: the solution in this PR may not work. Because setters can overwrite `null` on property. That's why I looked for a different solution.

anselmh: Thanks for explaining it further. What’d be the case that it wouldn’t work? My current code change solves my error and works in my case but of course I’m not sure if it works in all cases. If your solution is the more robust one though I tend to choose it but leaving ultimately to the Kirby team.

afbora: As you said, it may be working now, but in edge cases, the same problem may occur again.

afbora: Please check out the PR I'm proposing https://github.com/getkirby/kirby/pull/4170

afbora:
Closing in favour of #4170

PR #4158, opened 2022-02-14
Enhance `Uri::current()` method for reverse proxy

### LINKED ISSUE #3947: Installation and login failing in 3.6 (reverse proxy environments)
_Update: the same problem occurs for login and all other panel urls, so in generally kirby 3.6 panel completely doesn't work behind reverse proxy environments any more, at least when using other port numbers internally._ 

**Describe the bug**  
Kirby can not be installed any more behind reverse proxy setups because the installation url is returned wrong.

We are generally using this `url` method in `config.php` to run kirby behind reverse proxy installations (docker environments behind traefik router):

config.php:

```
    // fix to make available behind reverse proxy
    'url' => function () {
        if(isset($_SERVER['HTTP_X_FORWARDED_FOR'])) {
            return $_SERVER['HTTP_X_FORWARDED_PROTO'] . '://' . $_SERVER['HTTP_X_FORWARDED_HOST'];
        }
        return $_SERVER['REQUEST_SCHEME'] . ':'  . '//' . $_SERVER['SERVER_NAME'] . ':' . $_SERVER['SERVER_PORT'];
    }
```
That function helps to output the correct urls using HTTP_X_FORWARDED_HOST so that kirby works behind load balancers or other kinds of reverse proxies like traefik.

The panel first time installation fails in such environments, probably because the `url` method is not used to generate the installation redirect url:

In 3.6 it seems that the configurable `url` method is not used when building the panel redirect url for the webpage. The code in the page usually looks like:

```
// Panel state
   const json = {"$config": ...... }
```

Within the json there is the panel installation like this: 
`"$url":"https:\/\/some.example.host:8080\/panel\/installation"`

The problem is **8080**:
The internal port number is used, instead of the url which is returned from the `url` method above. All other urls in the page are correctly rewritten by that method. But for this single situation (building the redirect for the first installation) the wrong url is returned.

**To Reproduce**  
1. use kirby behind traefik or another proxy
2. use the above `url` method to overwrite the urls to have kirby work in such environments
3. try to load the panel without existing accounts and expect kirby to start installation
4. See error (page stays blank, error in console log because of wrong url)

**Screenshots**  
image

**Kirby Version**  
3.6

afbora:
## Describe the PR

First of all I would like to thank @samzzi for setting up an environment that I can test 🙌 

As a result of my tests, I reached Sam's conclusion on the forum:
https://forum.getkirby.com/t/reverse-proxy-shenanigans/24722

When `Uri::current()` doesn't return the correct url in reverse proxy, the whole system is broken. As a solution, I preferred to use the default ports (80-non ssl and 443-ssl) if there is a reverse proxy. Of course, it can still be overwritten with `forwarded` and `port` parameters.

It works successfully in local, remote server, subfolder, nginx reverse proxy environment and cloudflare. But frankly I am not sure if this change is 100% correct and includes any breaking changes.

### Fixes

- Fixed the panel not working behind reverse proxy setups #3947 

## Breaking changes

None

## Related issues/ideas

- Fixes #3947 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Http/Uri.php
+++ b/src/Http/Uri.php
@@ -254,10 +254,17 @@ class Uri
         );
         $uri = parse_url('http://getkirby.com' . $uri);
 
+        // returns default ports if reverse proxy exists
+        if ($forwarded === false && isset($_SERVER['HTTP_X_FORWARDED_FOR']) === true) {
+            $port = Server::https() === true ? 443 : 80;
+        } else {
+            $port = Server::port($forwarded);
+        }
+
         $url = new static(array_merge([
             'scheme' => Server::https() === true ? 'https' : 'http',
             'host'   => Server::host($forwarded),
-            'port'   => Server::port($forwarded),
+            'port'   => $port,
             'path'   => $uri['path'] ?? null,
             'query'  => $uri['query'] ?? null,
         ], $props));
--- a/tests/Http/UriTest.php
+++ b/tests/Http/UriTest.php
@@ -96,6 +96,34 @@ class UriTest extends TestCase
         $this->assertEquals('a/b/ktest.loc', $uri->path());
     }
 
+    public function testCurrentWithReverseProxyDefault()
+    {
+        $_SERVER['HTTP_HOST'] = 'ktest.loc:8080';
+        $_SERVER['HTTP_X_FORWARDED_FOR'] = '10.1.123.234';
+
+        $uri = Uri::current();
+        $this->assertSame(80, $uri->port());
+    }
+
+    public function testCurrentWithReverseProxyCustomPort()
+    {
+        $_SERVER['HTTP_HOST'] = 'ktest.loc:8080';
+        $_SERVER['HTTP_X_FORWARDED_FOR'] = '10.1.123.234';
+
+        $uri = Uri::current(['port' => 8888]);
+        $this->assertSame(8888, $uri->port());
+    }
+
+    public function testCurrentWithReverseProxyForwarded()
+    {
+        $_SERVER['HTTP_HOST'] = 'ktest.loc:8080';
+        $_SERVER['HTTP_X_FORWARDED_FOR'] = '10.1.123.234';
+        $_SERVER['HTTP_X_FORWARDED_PORT'] = '8888';
+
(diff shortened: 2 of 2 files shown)


rasteiner:
Isn't the problem more that the `$forwarded` variable never gets set? 

`Sever::port()` already has support for proxies, but the `$forwarded` variable gets *never* (I think) passed through as `true` https://github.com/getkirby/kirby/blob/0a1f15d39a4be3e5ab6fae868d2ff607fab3dce9/src/Http/Server.php#L126-L130

I don't know what kind of security risk we could create by "simply" assuming that a request is forwarded when a `X-FORWARDED-` header is present, but if there is a risk this PR seems to ignore it, while if there is no risk maybe it would make sense to deprecate the `$forwarded` variable completely, or change its fallback value to `true` instead of the current `false`

distantnative:
`$forwarded` gets passed through in `Uri::current()` (https://github.com/getkirby/kirby/blob/main/src/Http/Uri.php#L260) which is only used in `Uri::index()` with passing the parameter (https://github.com/getkirby/kirby/blob/main/src/Http/Uri.php#L341)... which leads to `Url::index()`.... but basically nowhere where `$forwarded` is ever set. So it would always be the default `false`.

So I guess that's the main issue :D

samzzi:
> `$forwarded` gets passed through in `Uri::current()` (https://github.com/getkirby/kirby/blob/main/src/Http/Uri.php#L260) which is only used in `Uri::index()` with passing the parameter (https://github.com/getkirby/kirby/blob/main/src/Http/Uri.php#L341)... which leads to `Url::index()`.... but basically nowhere where `$forwarded` is ever set. So it would always be the default `false`.
> 
> So I guess that's the main issue :D

That's also what I found in my latest research (see https://github.com/getkirby/kirby/issues/3947)

afbora:
I've shot another fix https://github.com/getkirby/kirby/pull/4161, Please review it if it's worth a look.

afbora:
I've closed this one for now to avoid conflict. Please see https://github.com/getkirby/kirby/pull/4161

PR #4161, opened 2022-02-16
Remove/deprecate `$forwarded` param from `\Kirby\Http` namespace

### LINKED ISSUE #3947: Installation and login failing in 3.6 (reverse proxy environments)

afbora:
## Describe the PR

Another try for #3947 and see other solution https://github.com/getkirby/kirby/pull/4158

For now I just removed the parameter for POC. If the PR is approved, it will be deprecated.

Since using `HTTP_X_FORWARDED_HOST` directly can create security vulnerability, we can enable it under an option. For ex: `server.forwarded => true` option.

### Fixes

- Fixed the panel not working behind reverse proxy setups #3947 

## Breaking changes

- Removed `$forwarded` param for url methods

## Related issues/ideas

- Fixes #3947 

## Ready?

- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Http/Server.php
+++ b/src/Http/Server.php
@@ -117,16 +117,13 @@ class Server
     /**
      * Returns the correct port number
      *
-     * @param bool $forwarded
      * @return int
      */
-    public static function port(bool $forwarded = false): int
+    public static function port(): int
     {
         // based on forwarded port
-        if ($forwarded === true) {
-            if ($port = static::get('HTTP_X_FORWARDED_PORT')) {
-                return $port;
-            }
+        if ($port = static::get('HTTP_X_FORWARDED_PORT')) {
+            return (int)$port;
         }
 
         // based on HTTP host
@@ -160,12 +157,11 @@ class Server
     /**
      * Returns the correct host
      *
-     * @param bool $forwarded
      * @return string
      */
-    public static function host(bool $forwarded = false): string
+    public static function host(): string
     {
-        $host = $forwarded === true ? static::get('HTTP_X_FORWARDED_HOST') : null;
+        $host = static::get('HTTP_X_FORWARDED_HOST');
 
         if (empty($host) === true) {
             $host = static::get('SERVER_NAME');
--- a/src/Http/Uri.php
+++ b/src/Http/Uri.php
@@ -237,10 +237,9 @@ class Uri
 
     /**
      * @param array $props
-     * @param bool $forwarded
      * @return static
      */
-    public static function current(array $props = [], bool $forwarded = false)
+    public static function current(array $props = [])
     {
         if (static::$current !== null) {
             return static::$current;
@@ -256,8 +255,8 @@ class Uri
 
         $url = new static(array_merge([
             'scheme' => Server::https() === true ? 'https' : 'http',
-            'host'   => Server::host($forwarded),
-            'port'   => Server::port($forwarded),
+            'host'   => Server::host(),
+            'port'   => Server::port(),
             'path'   => $uri['path'] ?? null,
             'query'  => $uri['query'] ?? null,
         ], $props));
@@ -335,10 +334,9 @@ class Uri
      * or any other executed script.
      *
      * @param array $props
-     * @param bool $forwarded
      * @return string
      */
-    public static function index(array $props = [], bool $forwarded = false)
+    public static function index(array $props = [])
     {
         if (Server::cli() === true) {
             $path = null;
@@ -362,7 +360,7 @@ class Uri
             'path'     => $path,
             'query'    => null,
             'fragment' => null,
-        ]), $forwarded);
+        ]));
     }
 
 
--- a/src/Http/Url.php
+++ b/src/Http/Url.php
@@ -100,12 +100,11 @@ class Url
      * Returns the url to the executed script
      *
      * @param array $props
-     * @param bool $forwarded
      * @return string
      */
-    public static function index(array $props = [], bool $forwarded = false): string
+    public static function index(array $props = []): string
     {
-        return Uri::index($props, $forwarded)->toString();
+        return Uri::index($props)->toString();
     }
 
     /**


distantnative:
I think this is a good direction.

Maybe we cannot remove it right away (to don't have breaking changes). We could deprecate all parameters directly and remove in 3.7 (since I would argue that not many users would ever use them).  For `Server::port()` and `Server::host()` we make the default `null` for now and use the `HTTP_X_` values as default - but still support explicit `false` as parameter for now (to not have a breaking change).

I'd leave it to you, @lukasbestle and Bastian whether we need the config option to avoid security risk.

samzzi:
@afbora Tested the hotfix in a few environments and seems to fix the reverse proxy problems 🎉

afbora:
@samzzi It's important that it works in your environment. Thank you for your tests.

I think the PR and issue roadmap will become clear after discussion of whether this change poses a security vulnerability and whether to enable this setting from a global option.

distantnative:
@afbora what do you think about the way not to remove the parameter right away but deprecate it now and remove in 3.7?

afbora:
I removed the param to make the PR easier to understand, but it wasn't my intention to remove it right away. But frankly, I'm starting to think that there isn't anyone using this parameter. 🙂 

> not to remove the parameter right away but deprecate it now and remove in 3.7?

Agree with you. I think this is the best way. It would be nice to keep it in use for 3.6.x 👍

distantnative:
@afbora maybe like https://github.com/getkirby/kirby/commit/5001daf2ed9df4e0ea582b8e8ed82cea493d203d ?

afbora:
@distantnative Yes, We can turn it into PR and continue on yours 👍

afbora:
We also have to decide whether the reverse proxy should be enabled manually or automatically.

bastianallgeier:
Just thinking out loud. The Panel issues seem to be all coming from absolute URLs. Wouldn't it be another option to switch to relative URLs in case of reverse proxy setups?

distantnative:
@bastianallgeier That still wouldn't fix that we have a $forwarded parameter that pre tends to do something but is never used as intended 😄

bastianallgeier:
Absolutely! My comment was more about the security issue with the forwarded host Parameter

bastianallgeier:
Ok, here are some more thoughts.

A given fact is that we cannot trust HTTP_HOST and HTTP_X_FORWARDED_HOST. An option to trust it does not really make sense in my opinion. As a reverse proxy user, I want correct host detection and giving them an option means that they can only get it when they subscribe to an insecure setup. 

When you have to set a config option to enable this anyway, it would be better to just set the correct host instead. 

I think we should split this PR as a first step to simply the process. 

The port detection is definitely better in this PR, so we should fix that one first in the way we already agreed. As a next step we have to take a closer look at secure host detection possibilities.

bastianallgeier:
I also just saw the change in Uri::current. Wouldn't it make sense to put that into Server::port() ?

afbora:
I'm thinking out loud and I'm not sure how logical it is 🤔  

I think it would be safe if we had the `allowed forwarded hosts` option like [servers option](https://getkirby.com/docs/reference/system/options/servers)

````php
return [
    'allowed_forwarded_hosts' => ['static.getkirby.com', 'stage.getkirby.com']
];

bastianallgeier:
Yes, that would definitely be the most secure option. I still wonder if we can keep auto detection somehow.

bastianallgeier:
I also really wonder what changed from 3.5 to 3.6 to break the Panel.

afbora:
> I also really wonder what changed from 3.5 to 3.6 to break the Panel.

As far as I can see, if I'm not looking in the wrong place, the panel addresses are defined as:

In 3.5.8:
https://github.com/getkirby/kirby/blob/3.5.8/src/Cms/Panel.php#L109

In 3.6.2:
https://github.com/getkirby/kirby/blob/3.6.2/src/Panel/View.php#L199

bastianallgeier:
@afbora brilliant. Thanks for tracking that down. I really think we need to revert this.

distantnative:
Closing in favour of https://github.com/getkirby/kirby/pull/4173

PR #4173, opened 2022-02-22
Host detection

bastianallgeier:


src/Cms/App.php
```diff
@@ -1414,7 +1437,11 @@ protected function setSite($site = null)
      */
     public function server()
     {
-        return $this->server = $this->server ?? new Server();
+        if ($this->server !== null) {
+            return $this->server;
+        }
+
+        return $this->server = new Server();
```
distantnative: These whole changes could be reverted now that we don't do anything else in this method.

src/Cms/Environment.php
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * @param \Kirby\Http\Server $server
+     * @param string $root
+     * @param bool|string|array|null $allowed
+     */
+    public function __construct(Server $server, string $root, $allowed = null)
+    {
+        $this->root   = $root;
+        $this->server = $server;
```
distantnative: Why do we need to pass a server object - the server class is static methods only (it's already a bit weird on the App class).

src/Cms/App.php
bastianallgeier: Oh you are right. I guess it should probably even be: 

```
return $this->server ??= new Server();
```

src/Cms/Environment.php
```diff
@@ -0,0 +1,123 @@
...
+
+        // the current URL is detected via possibly insecure HOST headers
```
distantnative:
```suggestion
        // the current URL should be detected via possibly insecure HOST headers
```

src/Cms/Environment.php
```diff
@@ -0,0 +1,123 @@
...
+        if ($allowed === true || $allowed === null) {
+            $this->server->hosts(['*']);
+            $this->uri = Uri::index();
+
+        // the current URL is detected via the server name
```
distantnative:
```suggestion
        // the current URL should be detected via the server name
```

src/Cms/Environment.php
```diff
@@ -0,0 +1,123 @@
...
+        } elseif ($allowed === false) {
+            $this->server->hosts([]);
+            $this->uri = Uri::index();
+
+        // the current URL is predefined and not detected automatically
+        } elseif (is_string($allowed) === true) {
+            $this->uri = new Uri($allowed);
+            $this->server->hosts([$this->uri->host()]);
+
+        // the current URL is auto detected with a host allowlist
```
distantnative:
```suggestion
        // the current URL should be auto detected from a host allowlist
```

src/Cms/App.php
bastianallgeier: ✅

src/Cms/Environment.php
bastianallgeier: You are actually right. We could just use the static methods instead. I wonder if we will ever work with a real instance, but then we need to refactor quite a bit of code anyway. I will simplify it.

src/Cms/Environment.php
```diff
@@ -0,0 +1,116 @@
...
+            foreach ($allowed as $url) {
+                $host = (new Uri($url))->host();
+                $hosts[]    = $host;
+            }
+
+            Server::hosts($hosts);
+            $this->uri = Uri::index();
+        } else {
+            throw new InvalidArgumentException('Invalid allow list setup for base URLs');
+        }
```
afbora, line 93: Shouldn't we throw the exception if it doesn't match any url?

````php
if ($allowed === true || $allowed === null) {
    //
} elseif ($allowed === false) {
    //
} elseif (is_string($allowed) === true) {
    //
} elseif (is_array($allowed) === true) {
    //
}

throw new InvalidArgumentException('Invalid host!');
````

Also when no url is matched, the system works even without a host, but what exactly does allowed urls do? I didn't understand 🤔

Comment (afbora) [reply src/Cms/Environment.php:93]: Then, as far as I understand it's just about prefixing URLs with correct hostnames. Not blocking access to invalid hosts. Do I understand correctly?

Comment (distantnative) [reply src/Cms/Environment.php:93]: You need the else statement since we are not returning anything in that method. In your example, @afbora, every setup would still end up at `throw...`.

Comment (afbora) [reply src/Cms/Environment.php:93]: @distantnative I know, mine was just one such example. I'm trying to describe that will we allow the user to access the site if it doesn't match any url (not type of `url` option as string, array, etc..)? Like your said in notion:

> Q: What happens if an array of URLs is configured but none of them match the current host?
> A: Probably better to throw an exception as clearly a for the developer unintended situation appeared

Comment (bastianallgeier) [reply src/Cms/Environment.php:93]: I was really wondering about this during the implementation as well. We could always fall back to simple relative URLs instead of throwing an error. Do you think this could go wrong or would be worse?

Comment (afbora) [reply src/Cms/Environment.php:93]: I actually like your implementation. As I'm not sure, just wanted to point this out for security reasons. I think @lukasbestle knows these well.

Comment (bastianallgeier) [reply src/Cms/Environment.php:93]: Right now you cannot get an insecure host with this PR unless you actively allow it in your config. Otherwise you will always end up with an empty string. I'm pretty sure we are fine with this, but it would indeed be good to get a second or even third opinion.

Comment (distantnative) [reply src/Cms/Environment.php:93]: My thinking: if someone explicitly configured something and it doesn't match with reality, an error should be thrown. That makes more sense than trying to make things "just work" for a setup that the developer explicitly didn't plan to support.

Comment (bastianallgeier) [reply src/Cms/Environment.php:93]: You are right. We should keep the "just works" version when no url option is set and otherwise throw an error

Comment (distantnative) [reply src/Cms/Environment.php:93]: I guess then it looks good to me

File: src/Cms/Environment.php [hunk 7]
```diff
@@ -0,0 +1,116 @@
...
+
+        // the current URL is predefined and not detected automatically
+        } elseif (is_string($allowed) === true) {
+            $this->uri = new Uri($allowed);
+            Server::hosts([$this->uri->host()]);
+
+        // the current URL should be auto detected from a host allowlist
+        } elseif (is_array($allowed) === true) {
+            foreach ($allowed as $url) {
+                $host = (new Uri($url))->host();
```
distantnative, line 85: Will this preserve `fnmatch` patterns?

src/Cms/Environment.php
afbora: > I guess then it looks good to me

@distantnative Really sorry. Did I misunderstand? Now an exception is thrown when the type (bool, string, array) does not match (not the url) 🤔

distantnative: True we would need to throw another one here https://github.com/getkirby/kirby/pull/4173/files#diff-7c2c2d4cd5cb7754713993a0d4da17c51bbe03fd23fef782e9614c30ac1ed5e8R127 right?

src/Http/Server.php
```diff
@@ -84,59 +93,70 @@ public static function get($key = null, $default = null)
...
+            $host = static::get('HTTP_HOST');
+        }
+
+        if (empty($host) === true) {
+            $host = static::get('SERVER_ADDR');
+        }
+
+        // ignore invalid host names
+        if (static::isAllowedHost($host) === false) {
+            $host = null;
```
afbora, line 127: @distantnative Yes, I think it should throw exception from here like your point as we discussed in https://github.com/getkirby/kirby/pull/4173/files#r812212533

src/Cms/Environment.php
bastianallgeier: To be honest, the patterns don't really work at all in our url config option. How would they be translated back to proper URLs? I think it's nice to have them in the server class, but in this case they don't have any effect.

bastianallgeier:
Argh! I just implemented the exception, but of course this means that hundreds of tests fail, because their host setup is invalid. Back to the drawing board.

distantnative (approved):

PR #4175, opened 2022-02-23
Use standard "application/javascript" MIME type

### LINKED ISSUE #4174: `Mime` class: replace "application/x-javascript" with "application/javascript"
I'm investigating a very odd and randomly occurring issue in the Panel of a Kirby site (not subject of this ticket; we occasionally get a `Field type "text" does not exist` error showing up, which disappears simply with a hard refresh) and one of the many paths I'm looking at is browser caching. Since I noticed the following – and got sucked into the rabbit hole that is MIME type specs – I thought I'd ask:

Looking at the DevTools Network tab, it appears that all Panel JS files are loaded with MIME type `application/javascript`, except for `/media/plugins/index.js` which is loaded with MIME type `application/x-javascript`.

Is there a reason Kirby's `Mime` class uses the non-standard `application/x-javascript` MIME type for JS files?

https://github.com/getkirby/kirby/blob/04127160ed1ab8dc277763cd8ba4ebc5a097bc18/src/Filesystem/Mime.php#L51

If not, may I suggest (secretly hoping this could be behind that odd error I am dealing with :wink:) changing this MIME type to `application/javascript` in compliance with [RFC4329](https://www.rfc-editor.org/rfc/rfc4329.txt), which discourages use of the `x-` prefixed experimental MIME type:

> Use of experimental and unregistered media types, as listed in part above, is discouraged. The media types,
> * application/javascript
> * application/ecmascript
>
> which are also defined in this document, are intended for common use and should be used instead.

sebastiangreger:
## Describe the PR
Replace the non-standard `application/x-javascript` MIME type with `application/javascript`.

### Fixes
Assign MIME type `application/javascript` to JS files

## Breaking changes
None

## Related issues/ideas
- Fixes #4174 

## Ready?
- [ ] CI checks pass

## When merging

### DIFF
--- a/src/Filesystem/Mime.php
+++ b/src/Filesystem/Mime.php
@@ -48,7 +48,7 @@ class Mime
         'html'  => 'text/html',
         'ico'   => 'image/x-icon',
         'ics'   => 'text/calendar',
-        'js'    => 'application/javascript',
+        'js'    => ['application/javascript', 'application/x-javascript'],
         'json'  => ['application/json', 'text/json'],
         'j2k'   => ['image/jp2'],
         'jp2'   => ['image/jp2'],
--- a/src/Filesystem/Mime.php
+++ b/src/Filesystem/Mime.php
@@ -48,7 +48,7 @@ class Mime
         'html'  => 'text/html',
         'ico'   => 'image/x-icon',
         'ics'   => 'text/calendar',
-        'js'    => 'application/x-javascript',
+        'js'    => 'application/javascript',
         'json'  => ['application/json', 'text/json'],
         'j2k'   => ['image/jp2'],
         'jp2'   => ['image/jp2'],


afbora (dismissed):

bastianallgeier (commented):
I think we should convert it to an array and accept both. Otherwise it would be a breaking change.

sebastiangreger:
> I think we should convert it to an array and accept both. Otherwise it would be a breaking change.

:100: My PR was based on rather limited knowledge of what the `Mime` class does in the core. But am I correct in my assumption that this is the piece of code affecting the `/media/plugins/index.js` file being delivered with MIME type `application/x-javascript`. Or did I end up in the wrong place entirely? :roll_eyes:

afbora:
> Otherwise it would be a breaking change.

Yes that is right. But if I didn't misread, it was deprecated in 2006, right?. What kind of break could it be? 🤔 

Of course for now it seems the most reliable way is to support both and remove it in 3.7 (or we may not remove it at all).

bastianallgeier:
If I remember correctly, Firefox often sends those weird x- mime types on uploads. At least it did in the past. I'm not sure if it still happens.

sebastiangreger:
PR updated! I found a moment to test this, and since using an array still resolves the original issue of Kirby delivering JS with this deprecated MIME type keeping it for compatibility should not do any harm, no?

bastianallgeier (approved):

bastianallgeier:
Perfect, thank you! That's exactly why the array is the better solution. It does no harm to leave the old mime type in there. The new one will still work.

PR #4179, opened 2022-02-25
3.6.3 rc.1

bastianallgeier:


afbora:
Upgrade php mailer dependency?

lukasbestle (dismissed):
File changes look good to me.

I'm also not at my desk, so I can't say anything about dependencies. But we could still update them in the next RC.

bastianallgeier:
I've upgraded phpmailer. I agree that we can still upgrade the rest later if there are more updates required.

lukasbestle (approved):

PR #4182, opened 2022-02-28
Makes sure whoops is registered while the configurations is being loaded

### LINKED ISSUE #4180: [3.6.3-rc.1] Debug mode and whoops issue while throwing invalid host exception
## Description

While throwing invalid host exception:

1. It doesn't matter if the debug mode is on or off.
2. PHP native error output instead of Whoops

## Your setup
**Kirby Version**  

3.6.3-rc.1

afbora:
## Describe the PR

Makes sure whoops is registered while the configurations is being loaded

## Release notes

Since this issue comes after the environment class, it doesn't need to be added to the release notes.

## Breaking changes

None

## Related issues/ideas

- Fixes #4180 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -92,13 +92,17 @@ class App
         // register all roots to be able to load stuff afterwards
         $this->bakeRoots($props['roots'] ?? []);
 
-        // stuff from config and additional options
-        $this->optionsFromConfig();
-        $this->optionsFromProps($props['options'] ?? []);
-        $this->optionsFromEnvironment();
-
-        // register the Whoops error handler
-        $this->handleErrors();
+        try {
+            // stuff from config and additional options
+            $this->optionsFromConfig();
+            $this->optionsFromProps($props['options'] ?? []);
+            $this->optionsFromEnvironment();
+        } finally {
+            // register the Whoops error handler
+            // put into try/catch/finally block to to ensure Whoops registered
+            // if there is a problem loading the configurations
+            $this->handleErrors();
+        }
 
         // set the path to make it available for the url bakery
         $this->setPath($props['path'] ?? null);
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -98,9 +98,9 @@ class App
             $this->optionsFromProps($props['options'] ?? []);
             $this->optionsFromEnvironment();
         } finally {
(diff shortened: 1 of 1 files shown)


src/Cms/App.php
```diff
@@ -92,13 +92,17 @@ public function __construct(array $props = [], bool $setInstance = true)
...
-        $this->handleErrors();
+        try {
+            // stuff from config and additional options
+            $this->optionsFromConfig();
+            $this->optionsFromProps($props['options'] ?? []);
+            $this->optionsFromEnvironment();
+        } finally {
+            // register the Whoops error handler
+            // put into try/catch/finally block to to ensure Whoops registered
+            // if there is a problem loading the configurations
```
distantnative: Suggestion:

```
// register the Whoops error handler
// inside of a try-finally block to ensure it's still
// registered even if there is a problem loading the configurations
```

afbora: I'm afk, you can push the commit if you're available.

distantnative: I can do it tonight

```suggestion
            // register the Whoops error handler inside of a
            // try-finally block to ensure it's still registered
            // even if there is a problem loading the configurations
```

lukasbestle (commented):
To be honest I'm not sure if and why this works.

Considering the following test case:

```php
<?php

return [
    'url' => ['https://getkirby.com'],
    'debug' => true
];
```

On `develop` I get Kirby's `fatal.php` view (so Whoops seems to be active). With your PR I get the detailed Whoops error view (as debugging is enabled).

The behavior with this PR makes sense: The exception is thrown and then PHP runs the `finally` code before triggering the fatal error handler.
But why do I get the fatal view anyway even without this change? I'm confused.

afbora:
I'm not sure how you got the fatal.php (you mean ` This page is currently offline due to an unexpected error.` page, right?). I'm getting plain error output from PHP on develop branch:

screen-capture-1364--localhost

lukasbestle (approved):
Very weird. I've tested with a Starterkit again and now get what you also get. With the Sandbox I weirdly still get the fatal view.

But anyway, I can confirm that this change fixes the error and I also get why it does because the fatal error handler is triggered after `finally`. Very elegant solution! 👍

PR #4190, opened 2022-03-06
Query images for image and gallery blocks

### LINKED ISSUE #4189: Can select non-images in image block
## Description
When adding a new image block, I can easily select any file type, even non-image files like PDFs, ZIPs or audio. As the output snippet expects only images, I think it should only be possible to select images in the first place.

When I look into `kirby/config/blocks/image/image.yml` I can see that there is no filtering going on:

```yaml
name: field.blocks.image.name
icon: image
preview: image
fields:
  # ...
  image:
    label: field.blocks.image.name
    type: files
    multiple: false
    image:
      back: black
    uploads:
      template: blocks/image
    when:
      location: kirby
  # ...
```

The image field would need a `query: page.images` to only allow images to be selected.

## To reproduce
1. Have a blueprint with a block editor
2. Login to panel
3. Upload a non Image file
4. Create a new image block
5. Click on "Select file" and check if you can select that non-image file

## Your setup
**Kirby Version**  3.6.2

afbora:
## Describe the PR

Use `query: model.images` for image and gallery blocks to ensure only images selectable.

### Enhancements
- Supports `model.` entry for model queries (as placeholder alias for `page.`, `file.`, `user.` or `site.`), which allows a query syntax that can be shared between different model types.
- Image and gallery blocks picker only shows images for selection

## Breaking changes

None

## Related issues/ideas

- Fixes #4189 

## Ready?

- [x] CI checks pass

## When merging

### DIFF
--- a/config/blocks/gallery/gallery.yml
+++ b/config/blocks/gallery/gallery.yml
@@ -5,7 +5,7 @@ fields:
   images:
     label: field.blocks.gallery.images.label
     type: files
-    query: page.images
+    query: model.images
     multiple: true
     layout: cards
     size: tiny
--- a/config/blocks/gallery/gallery.yml
+++ b/config/blocks/gallery/gallery.yml
@@ -5,6 +5,7 @@ fields:
   images:
     label: field.blocks.gallery.images.label
     type: files
+    query: page.images
     multiple: true
     layout: cards
     size: tiny
--- a/config/blocks/image/image.yml
+++ b/config/blocks/image/image.yml
@@ -13,6 +13,7 @@ fields:
   image:
     label: field.blocks.image.name
     type: files
+    query: page.images
     multiple: false
     image:
       back: black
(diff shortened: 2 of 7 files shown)


distantnative (dismissed):

distantnative:
@afbora Sorry, I just realised using `page.images` might be problematic when we use the blocks in a file or user blueprint.

distantnative:
Might be a good feature to change this https://github.com/getkirby/kirby/blob/main/src/Cms/ModelWithContent.php#L348-L350 to not just support `kirby.`, `site.` and the specific model type (either `page.` or `file.` or `user`) - but to always add also

```
'model' => $this
```

Then we could do `model.images`.

afbora:
Definitely! When I read your first comment, I thought of a solution like `model.*`. Should we apply that to this PR?

distantnative:
Yes, I think it's such a small additive change that we can include it in this PR.

distantnative:
@afbora Added

afbora:
@distantnative Thanks for taking care. I can't approve self PR 🙂 Will we include it in 3.6.3?

distantnative (approved):

bastianallgeier (approved):

Issue #4177, opened 2022-02-23
Kirby\Cms\Blueprint::load fails with multi-language block blueprint


shoogstoel:
## Description
Tried to load a custom block's blueprint using `Kirby\Cms\Blueprint::load('block-name-here')`, resulting in a `TypeError` thrown by `ucfirst`, because Kirby assume the block's name to be a `string`. However, if the name is specified in multiple languages, it is actually an `array`.

**Expected behavior**  
Blueprint to be loaded successfully.

## To reproduce
To reproduce create a custom block with a name in more than 1 language and try to load its blueprint using `Kirby\Cms\Blueprint::load('block-name-here')`.

## Your setup
**Kirby Version**  
3.6.2

**Console output**  
```
{
  "status": "error",
  "message": "ucfirst(): Argument #1 ($string) must be of type string, array given",
  "code": 500,
  "exception": "TypeError",
  "key": null,
  "file": "/kirby/src/Cms/Blueprint.php",
  "line": 343,
  "details": [],
  "route": "page/([a-zA-Z0-9\\.\\-_%= \\+\\@\\(\\)]+)"
}
```

**Your system (please complete the following information)**
 - Device: Apache Docker container running on WSL2
 - OS: Windows 11 22000.527
 - Browser: Microsoft Edge
 - Version:  98.0.1108.56

afbora:
Could you try `::extend()` method like as Kirby blocks use to load?

````php
Kirby\Cms\Blueprint::extend('block-name-here');
````

shoogstoel:
Sorry it tooke me so long, @afbora, can confirm that using `Kirby\Cms\Blueprint::extend` does not produce the error I described.

distantnative:
For sure, the description of `Kirby\Cms\Blueprint::load` isn't correct then

> Loads a blueprint from file or array

But I am also still not sure what your re doing. For one thing, it should probably be e.g. `Kirby\Cms\Blueprint::load('blocks/button')` (when following these docs https://getkirby.com/docs/reference/plugins/extensions/blocks#full-block-plugins__adding-a-default-blueprint-and-snippet)

But why would you specify the blocks name in different languages?

shoogstoel:
I am naming it differently for specific languages in the blueprint.

distantnative:
Can you provide your code?

I would think you don't need multiple blueprints for multiple languages, but rather translate the strings of your block inside the one blueprint.

shoogstoel:
[...]

PR #4192, opened 2022-03-08
Fix loading blueprint with i18n’ed name #4177

distantnative:
### Fixes
- Blueprints with different translations for `name` key now can be loaded
#4177 

### Breaking changes
_None_

## Ready?

### For review team

-  ~~Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)~~

### DIFF
--- a/src/Cms/Blueprint.php
+++ b/src/Cms/Blueprint.php
@@ -76,8 +76,7 @@ class Blueprint
         // normalize the name
         $props['name'] ??= 'default';
 
-        // normalize and translate the name and title
-        $props['name']  = $this->i18n($props['name']);
+        // normalize and translate the title
         $props['title'] = $this->i18n($props['title'] ?? ucfirst($props['name']));
 
         // convert all shortcuts
--- a/src/Cms/Blueprint.php
+++ b/src/Cms/Blueprint.php
@@ -76,7 +76,8 @@ class Blueprint
         // normalize the name
         $props['name'] ??= 'default';
 
-        // normalize and translate the title
+        // normalize and translate the name and title
+        $props['name']  = $this->i18n($props['name']);
         $props['title'] = $this->i18n($props['title'] ?? ucfirst($props['name']));
 
         // convert all shortcuts
@@ -290,9 +291,13 @@ class Blueprint
         // now ensure that we always return the data array
         if (is_string($file) === true && F::exists($file) === true) {
             return static::$loaded[$name] = Data::read($file);
-        } elseif (is_array($file) === true) {
+        }
+
+        if (is_array($file) === true) {
             return static::$loaded[$name] = $file;
-        } elseif (is_callable($file) === true) {
+        }
+
+        if (is_callable($file) === true) {
             return static::$loaded[$name] = $file($kirby);
         }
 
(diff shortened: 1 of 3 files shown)


lukasbestle (commented):
What I find confusing: In "normal" blueprints we use the `title` field for the human-readable blueprint name. Why do we use the `name` field in block blueprints?

I feel like we should rather rename the `name` field to `title` to make it consistent. Then the block blueprints can use the same translation logic in the `Blueprint` class.

distantnative:
@lukasbestle I asked @bastianallgeier the same and he said it's due to the transition from builder and blocks plugins. But we also said, like you suggest, that aliasing/renaming would be good. Similarly to https://github.com/getkirby/kirby/issues/4194.

Nevertheless, I think this would be a good quick fix before we can tackle the consistency updates.

lukasbestle:
OK, I see. Makes sense. I think we should then add TODO comments to ensure that we remove this hack once the consistency is restored.

bastianallgeier:
I wonder if we should actually tackle the inconsistency issue first. Translating the name is weird because it's normally used in a pretty id-ish way, where translations could theoretically be catastrophic. 

So maybe a better solution would be to introduce title as new official option and simply convert name for block fieldsets before it even hits the constructor.

distantnative:
@bastianallgeier I think it all depends on timelines. If the consistency alignment takes us a few month, I'd prefer to do the quick fix like this now. If the consistency alignment can happen e.g. in 3.6.4, then sure.

bastianallgeier:
I don't think we have to make the full alignment already. We'd basically just polyfill the title for now to fix the issue and move on with more consistency fixes later.

bastianallgeier:
I think I would do the following. 

We change this line …

https://github.com/getkirby/kirby/blob/04127160ed1ab8dc277763cd8ba4ebc5a097bc18/src/Cms/Fieldset.php#L57 

… to …

```php
$this->name = $this->createName($params['title'] ?? $params['name'] ?? Str::ucfirst($this->type));
```

This would be a very minimal change for now, which is still 100% compatible with the current syntax. Then @shoogstoel can already use the title option for his blocks that he needs to load with the custom loader. 

The adjusted code for `Blueprint::load` would otherwise only serve this single use-case, which is pretty specific anyway.

distantnative:
Ok, let's do it

distantnative:
Hmm, @bastianallgeier, but with your change, the bug would continue to exist for anyone using translations on `name`, am I right? It would be up to users to change their code to use `title` instead?

bastianallgeier:
Yes, you are right. But only if they load the Block Blueprint manually with Blueprint::load() which is pretty unusual. It should work when you translate the Block name and just use the Blocks field regularly.

distantnative:
Maybe my mind is too twisted, but if the custom block blueprint looks like this:

```
/site/plugins/button/blueprints/blocks/button.yml

name:
  en: Button
  de: Knopf
icon: bolt
fields:
  link:
    type: url
  text:
    type: text
```

This would work?

bastianallgeier:
I thought it would work in our regular setup, yes. Is it failing?

bastianallgeier:
I've tested it and it works. I've already created a replacement for this PR: https://github.com/getkirby/kirby/pull/4239

PR #4196, opened 2022-03-13
Add GitHub wrokflow for releases

distantnative:


.github/workflows/release.yml
```diff
@@ -0,0 +1,19 @@
...
+      - "main"
+      - "release/**"
+  pull_request:
+    branches:
+      - "main"
+      - "release/**"
+
+jobs:
+  ci:
+    name: "Continous integration"
```
lukasbestle:
```suggestion
    name: "Continuous integration"
```

lukasbestle (commented):
I think we can call the workflows directly as jobs with this syntax: https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow
Should be faster as GitHub doesn't need to spin up an Ubuntu VM just to trigger the workflows.

Issue #4191, opened 2022-03-06
[3.6.3-rc.1] `Environment` doesn't check the path in array `url` options


lukasbestle:
## Description

If the `url` option is configured as array, Kirby only checks the hostnames that are contained in these configured allowed URLs. If the configured URLs contain paths, these are neither checked nor used for the system index URL.

**Expected behavior**  

The `Environment` should match the full detected index URL against the configured array, not just the hostname. If the path is invalid, no match will be found and the exception should be thrown.

**Additional context**  

The issue comes from this line:

https://github.com/getkirby/kirby/blob/e891e3e517551e04facd52a90d6e7e474a8f992b/src/Cms/Environment.php#L85

The full `$url` is basically thrown away at this point.

## To reproduce

**`config.php`:**

```php
<?php

return [
    'debug' => true,
    'url' => [
        'https://example.com/my-site/',
        'http://localhost/my-expected-path/',
    ],
];
```

**Actual (detected) index URL:**
`http://localhost/my-actual-path`

**Code in any template:**

```php
echo url('my-page');
```

**Output:**

```
http://localhost/my-actual-path/my-page
```

## Your setup
**Kirby Version**  

3.6.3-rc.1

bastianallgeier:
✅

PR #4199, opened 2022-03-16
Check for valid subfolders when setting an array of allowed base URLs

bastianallgeier:
See: https://github.com/getkirby/kirby/issues/4191

### DIFF
--- a/src/Cms/Environment.php
+++ b/src/Cms/Environment.php
@@ -39,15 +39,9 @@ class Environment
     {
         $this->root = $root;
 
-        // empty hosts for relative URLs may be allowed in
-        // some scenarios. Otherwise an exception is thrown further down.
-        $allowEmptyHost = false;
-
         // This is the default. The current URL should be detected via the server name.
         // Empty hosts are allowed. Should lead to no breaking changes with previous setups.
         if ($allowed === null) {
-            $allowEmptyHost = true;
-
             Server::hosts([]);
             $this->uri = Uri::index();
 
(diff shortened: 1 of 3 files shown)


afbora:
I don't know how important but ending slash allowed url doesn't work:

````php
// doesn't work
return [
    'url' => [
        'http://localhost/test/kirby/develop/',
    ]
];

// works
return [
    'url' => [
        'http://localhost/test/kirby/develop',
    ]
];
````

**Comment by distantnative** [comment]
Given that users are very likely to do it both ways, I think it is rather important to support both

**Comment by distantnative** [comment]
@afbora are you able to create an [EnvironmentTest.php](https://github.com/getkirby/kirby/pull/4199#diff-3ca5a9e062bfedf0e51359b196dd81354fb9c88bb16a2f8e1134c7eac1972285) unit test that shows this fails?

File: src/Cms/Environment.php [hunk 1]
```diff
@@ -82,18 +77,36 @@ public function __construct(string $root, $allowed = null)
             // the current URL should be auto detected from a host allowlist
         } elseif (is_array($allowed) === true) {
             foreach ($allowed as $url) {
```
afbora, line 79:
```suggestion
            $hosts = [];
            foreach ($allowed as $url) {
```

afbora:
@distantnative Yes, I can reproduce the exception with ending slash change in `EnvironmentTest::testAllowFromUrlsWithSubfolders()` test
https://github.com/getkirby/kirby/blob/fix/4191-subfolder-check/tests/Cms/Environment/EnvironmentTest.php#L154-L169

lukasbestle (commented):
We could replace the foreach loop with something like this:

```php
$allowedUris = array_map(fn($url) => new Uri($url), $allowed);
$hosts       = array_map(fn($uri) => $uri->host(), $allowedUris);
```

Then we can do:

```php
$allowedStrings = array_map(fn($uri) => (string)$uri, $allowedUris);

if (in_array((string)$this->uri, $allowedStrings) {
    ...
}
```

lukasbestle:
I suggest we make any further changes to the third and last PR. For review and discussion the separate PRs were really helpful, but if we now make changes here, we will have to rebase against a lot of merge conflicts. We can avoid that by committing to the third branch and then merging just that.

Issue #4185, opened 2022-02-28
[3.6.3-rc.1] New url overwrite config doesn't overwrite panel URL


samzzi:
## Description
When you use the url overwrite option in the config the panel url doesn't get overwritten which stops the panel from working.
Example: https://smooth-meadow-s8t0nkgaql.ploi.link/panel/login

You'll notice in the rendered HTML the `$url` is wrong (although the `$urls` are correct:

```html
const json = { 
... 
"$urls":{"api":"https:\/\/smooth-meadow-s8t0nkgaql.ploi.link\/api","site":"https:\/\/smooth-meadow-s8t0nkgaql.ploi.link\/"},"
 ... 
,"$url":"http:\/\/kirbytest.previewsite.be\/panel\/login", 
...
}};
```

The config setting I used:

```php
return [
    'debug' => true,
    'panel' => ['install' => true],
     'url' => 'https://smooth-meadow-s8t0nkgaql.ploi.link/',
];
```

**Expected behavior**  
The panel to be working? 😊

## To reproduce

1. Use a website setup with multiple URL's where
2. Add the 'url' overwrite in the config to enable the correct url for your site

## Your setup
**Kirby Version**  : 3.6.3-rc.1

**Console output**  

```js
Uncaught SyntaxError: Unexpected token ':'
index.js:1 DOMException: Failed to execute 'pushState' on 'History': A history state object with URL 'http://kirbytest.previewsite.be/panel/login' cannot be created in a document with origin 'https://smooth-meadow-s8t0nkgaql.ploi.link' and URL 'https://smooth-meadow-s8t0nkgaql.ploi.link/panel/login'.
    at Object.onPushState (https://smooth-meadow-s8t0nkgaql.ploi.link/media/panel/911cd65cb7f3f66bc0f75ce7852776e0/js/index.js:1:7778)
    at pt.setState (https://smooth-meadow-s8t0nkgaql.ploi.link/media/panel/911cd65cb7f3f66bc0f75ce7852776e0/js/index.js:1:18165)
    at pt.init (https://smooth-meadow-s8t0nkgaql.ploi.link/media/panel/911cd65cb7f3f66bc0f75ce7852776e0/js/index.js:1:16374)
    at a.created (https://smooth-meadow-s8t0nkgaql.ploi.link/media/panel/911cd65cb7f3f66bc0f75ce7852776e0/js/index.js:1:7473)
[...]

PR #4200, opened 2022-03-16
Overwrite the url correctly from config

bastianallgeier:
**⚠️ First merge https://github.com/getkirby/kirby/pull/4199**

See #4185 

I used the suggestion from @lukasbestle and @afbora which seems to work great. But a real world test would be awesome.


lukasbestle (approved):
I don't have a test setup for this unfortunately and every combination of different headers might behave differently, but the code changes look good in context. 👍

Issue #4198, opened 2022-03-13
[3.6.3-rc.1] `url` option: Constants instead of `true`, `false`, `null`


lukasbestle:
I feel like the `true` and `false` values are too nondescript. What if we replace them with constants like `Server::HOST_ANY_EMPTY` (= null, default), `Server::HOST_ANY` (currently false, new constant value e.g. 1) and `Server::HOST_ANY_INSECURE` (currently true, new constant value e.g. 2)?
Considering that `Server` is already globally aliased, that should be viable even for non-techy users.

**Discussion:**
https://discord.com/channels/525634039965679616/525637437494657024/950339023120003104

bastianallgeier:
✅

PR #4201, opened 2022-03-16
3.6.3-rc.2

bastianallgeier:
- Check for valid subfolders when setting an array of allowed base URLs #4191
- Overwrite the url correctly from config #4185
- Server constants instead of booleans #4198


src/Cms/Environment.php
```diff
@@ -39,62 +39,23 @@ public function __construct(string $root, $allowed = null)
...
-            $this->blockEmptyHost();
-
-            // validate against the list of allowed base URLs
-            if (in_array((string)$this->uri, $allowed) === false) {
-                throw new InvalidArgumentException('The subfolder is not in the allowed base URL list');
-            }
-        } else {
-            throw new InvalidArgumentException('Invalid allow list setup for base URLs');
+        if (is_string($allowed) === true) {
+            return $this->setupFromString($allowed);
```
lukasbestle: One detail: The constructor should not return values, might trigger deprecation warnings in the future.

Either the methods should return `void` or the constructor should use a pattern like this:

```php
$this->setupFromString($allowed);
return;
```

lukasbestle (commented):
I really love the `setup*` methods. Much cleaner. 😍

bastianallgeier:
I've implemented all your comments from the other PRs and closed the other two. 

I kept the foreach loop though instead of three array_maps. Seemed a bit less wasteful to have three loops.

afbora (dismissed):

PR #4208, opened 2022-03-21
Support for IntlDateFormatter

distantnative:
## This PR …

PHP 8.1 deprecates `strftime` which will be dropped completely in PHP 9.
This PR adds support for `IntlDateFormatter` as future alternative. 

### Feature
- New `intl` date handler (e.g. for `date.handler` config option) using the [ICU date/time format syntax](https://unicode-org.github.io/icu/userguide/format_parse/datetime/#datetime-format-syntax)
```php
// different ways to use it
$page->myDateField()->toDate('M/d/yy', 'intl');

$formatter = new IntlDateFormatter(null, IntlDateFormatter::LONG, IntlDateFormatter::SHORT);
$page->myDateField()->toDate(formatter);
```

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -463,11 +463,11 @@ class F
      * Get the file's last modification time.
      *
      * @param string $file
-     * @param string $format
-     * @param string $handler date or strftime
+     * @param string|\IntlDateFormatter|null $format
+     * @param string $handler date, intl or strftime
      * @return mixed
      */
-    public static function modified(string $file, string $format = null, string $handler = 'date')
+    public static function modified(string $file, $format = null, string $handler = 'date')
     {
         if (file_exists($file) !== true) {
             return false;
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -372,11 +372,11 @@ class File
     /**
      * Returns the file's last modification time
      *
-     * @param string $format
-     * @param string|null $handler date or strftime
+     * @param string|\IntlDateFormatter|null $format
+     * @param string|null $handler date, intl or strftime
      * @return mixed
      */
-    public function modified(?string $format = null, ?string $handler = null)
+    public function modified($format = null, ?string $handler = null)
     {
         $kirby = $this->kirby();
 
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -2,7 +2,9 @@
 
 namespace Kirby\Toolkit;
 
+use DateTime;
 use Exception;
+use IntlDateFormatter;
 use Kirby\Exception\InvalidArgumentException;
 
 /**
@@ -264,17 +266,33 @@ class Str
      * according to locale settings
      *
      * @param int|null $time
-     * @param string|null $format
-     * @param string $handler date or strftime
+     * @param string|\IntlDateFormatter|null $format
+     * @param string $handler date, intl or strftime
      * @return string|int
      */
-    public static function date(?int $time = null, ?string $format = null, string $handler = 'date')
+    public static function date(?int $time = null, $format = null, string $handler = 'date')
     {
         if (is_null($format) === true) {
             return $time;
         }
 
-        // separately handle strftime to be able
+        // $format is an IntlDateFormatter instance
+        if (is_a($format, 'IntlDateFormatter') === true) {
+            return $format->format($time ?? time());
+        }
+
+        // `intl` handler
+        if ($handler === 'intl') {
+            $datetime = new DateTime();
+
+            if ($time !== null) {
+                $datetime->setTimestamp($time);
+            }
+
+            return IntlDateFormatter::formatObject($datetime, $format);
+        }
+
+        // handle `strftime` to be able
         // to suppress deprecation warning
         // TODO: remove strftime support for PHP 9.0
         if ($handler === 'strftime') {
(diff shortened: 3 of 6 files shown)


src/Filesystem/F.php
```diff
@@ -463,11 +463,11 @@ public static function mimeToType(string $mime)
      * Get the file's last modification time.
      *
      * @param string $file
-     * @param string $format
-     * @param string $handler date or strftime
+     * @param string|array|null $format
+     * @param string|null $handler date, intl or strftime
```
lukasbestle:
```suggestion
     * @param string $handler date, intl or strftime
```

lukasbestle (commented):
I feel like we should make this as simple as possible. The `IntlDateFormatter` object as `$format` argument makes a lot of sense for complete flexibility, but I think we should avoid that `array` format rabbit hole.

I just saw that there is a [`IntlDateFormatter::formatObject()`](https://www.php.net/manual/en/intldateformatter.formatobject.php) method, which takes a `$format` string: https://unicode-org.github.io/icu/userguide/format_parse/datetime/#datetime-format-syntax

Maybe that in combination with a full `IntlDateFormatter` could already be enough.

distantnative:
I like that. Will try to set it up early next week.

src/Toolkit/Str.php
```diff
@@ -264,17 +266,33 @@ public static function contains(string $string = null, string $needle, bool $cas
...
+    public static function date(?int $time = null, $format = null, string $handler = 'date')
     {
         if (is_null($format) === true) {
             return $time;
         }
 
-        // separately handle strftime to be able
+        // $format is an IntlDateFormatter instance
+        if (is_a($format, 'IntlDateFormatter') === true) {
+            return $format->format($time);
```
lukasbestle: `IntlDateFormatter::format()` doesn't accept `null` as the `$time`. So we need this:

```suggestion
            return $format->format($time ?? time());
```

src/Toolkit/Str.php
```diff
@@ -264,17 +266,33 @@ public static function contains(string $string = null, string $needle, bool $cas
...
+        }
+
+        // if $format is an array, `intl` needs to be (default) handler
+        if (is_array($format) === true) {
+            $handler = 'intl';
+        }
```
lukasbestle: To be honest I feel like we should get rid of the `array` format type / not add it in the first place:

- It's not needed as advanced users can also pass an `IntlDateFormatter` object
- The type hint for the argument becomes confusing (we would need to explain what the array means and that it's only for the `intl` handler)
- We would need checks that prevent `array` for the other two handlers. Otherwise you will get errors if you pass an array for `date()`.

tests/Toolkit/StrTest.php
```diff
@@ -139,6 +140,14 @@ public function testDate()
         $this->assertSame($time, Str::date($time));
         $this->assertSame('29.01.2020', Str::date($time, 'd.m.Y'));
 
+        // `intl` handler
+        $formatter = new IntlDateFormatter(null, IntlDateFormatter::SHORT, IntlDateFormatter::NONE);
+        $this->assertSame($time, Str::date($time, null, 'intl'));
+        $this->assertSame('1/29/20', Str::date($time, 'M/d/yy', 'intl'));
+        $this->assertSame('1/29/20', Str::date($time, $formatter));
+        $this->assertSame('1/29/20', Str::date($time, [IntlDateFormatter::SHORT , IntlDateFormatter::NONE]));
```
lukasbestle: Can you make the tests play with the format a bit more, e.g. also include the time in the format of the `$formatter` object and in another way in the string? This way we have a higher chance of catching regressions down the road.

src/Filesystem/File.php
```diff
@@ -372,11 +372,11 @@ public function mime()
     /**
      * Returns the file's last modification time
      *
-     * @param string $format
-     * @param string|null $handler date or strftime
+     * @param string|array|null $format
```
```suggestion
     * @param string|array|\IntlDateFormatter|null $format
```

Or without `array` if we decide to remove that option.

src/Toolkit/Str.php
```diff
@@ -264,17 +266,33 @@ public static function contains(string $string = null, string $needle, bool $cas
...
+
+        // `intl` handler
+        if ($handler === 'intl') {
+            $datetime = (new DateTime())->setTimestamp($time);
```
```suggestion
            $datetime = new DateTime();

            if ($time) {
                $datetime->setTimestamp($time);
            }

```

bastianallgeier:
I like this direction a lot. @distantnative what do you think about @lukasbestle's suggestions? They look good to me. I think it would be nice to wrap this up. Maybe even already for something like 3.6.x

distantnative:
✅

Tbh I am not sure if I would add these kind of features to 3.6.x releases. We could technically, but at the same time with our rapid releases we want users to update without needing to think much about it. So we are actively working towards a lot less attention of these features if we put them in 3.6.x releases - many users would simply miss that we added this. Sometimes I think it makes sense to rather store them for the next feature release where people will actively check out what new shiny things got added.

lukasbestle (approved):
It's so nice and simple. 🙌

I think it could be merged for either 3.6.5 or 3.7.0. If we merge it for 3.6.5, then users could already adapt to it before we deprecate `strftime` in 3.7 with a deprecation warning. But we could also do it both at the same time.

bastianallgeier:
The feature advertising problem is true. But I also saw quite a few people who already struggled with this. So merging it earlier is really more about already giving them a solution. I think both is fine.

distantnative:
Ok then let's merge it for 3.6.5

Issue #3638, opened 2021-08-16
Structure field with custom column widths breaks on mobile


VIPStephan:
**Describe the bug**  
There seems to be an inconsistency with responsive layout adjustments: I have a table block with a structure field with three columns by default. I want only two columns to show up in the block preview, so I’ve specified two columns with 1/2 width each. This looks fine in the table preview but it also uses the width setting in edit mode where it just shows the first field, and it screws up the width.

**To Reproduce**  
1. Set the viewport width to 1040px or lower.
2. Create a table block, according to https://getkirby.com/docs/reference/panel/blocks/table, with three fields and only two columns specified to show up (see first image below)
3. Add content to the structure field
4. Open edit mode of table block
5. See error.

**Expected behavior**  

If columns are hidden by the system due to CSS media queries, column/width settings should be ignored or adapted.

**Screenshots**  

Blueprint setup (~/site/blueprints/blocks/table.yml):
![table_block_blueprint](https://user-images.githubusercontent.com/13770606/129639854-6cc7d98e-a9bc-456a-9f22-f6ae1ad1cf4e.png)

Default table preview in the main area without column specification:
![default_columns_preview](https://user-images.githubusercontent.com/13770606/129640042-ff21e5ef-6eb0-4877-8cdc-e0b04b068d92.png)

Default table preview in the edit mode overlay:
![default_columns_edit_mode](https://user-images.githubusercontent.com/13770606/129640143-7e2fb60c-31d4-4f59-85dc-ba3c09185533.png)

Table preview in the main area with columns specified:
![2_columns_preview](https://user-images.githubusercontent.com/13770606/129640211-5a46d8a4-a6e8-4e1f-be00-541f937c3a38.png)

Table preview in edit mode overlay with columns specified:
![2_columns_edit_mode](https://user-images.githubusercontent.com/13770606/129640243-bf422ed9-db3f-4b24-bc53-d59ba69011ae.png)

**Kirby Version**  
3.5.7.1

**Desktop (please complete the following information):**  
 - OS: macOS 10.15.7 (Catalina)
 - Browser: Firefox
 - Version: 91

lukasbestle:
@VIPStephan Could you please post the blueprint YAML as text in a Markdown code block:

    ```yaml
    your blueprint
    ```

This makes it a lot easier for us to reproduce and debug this. Thanks!

VIPStephan:
```yaml
name: Tabelle
icon: menu
preview: table
fields:
  rows:
[...]

PR #4215, opened 2022-03-22
New `k-table` for structure field and block preview

distantnative:
## This PR …

Screen Shot 2022-03-22 at 11 13 53

Screen Shot 2022-03-22 at 11 13 48

### Features
- New `k-table` component to allow easily for consistent tables
- Styled preview for tags in tables
- New JS helper `$helpers.object.isEmpty(value)`

### Refactoring
- Structure field uses `k-table`
- New internal components `k-structure-form`
- Table block preview uses `k-table`

### Fixes
- `k-empty` now actually supports its `text` prop
- Improve structure field columns on mobile 
#3638
- Structure field sticky header insider drawers
#4242
- Wrong draggable ghost element for structure field rows
#4242

### Breaking changes
- `k-block-type-table` and `k-structure-field` have been largely refactored. If you (or a plugin) extends these, stuff likely could break.

## Ready?

### For review team


bastianallgeier (commented):
This looks really good! I wonder if we should keep the structure mixin though. Is it possible to use that in a plugin? If yes, it would be a breaking change otherwise.

distantnative:
I don't think plug-ins can. panel/src/config/plugins.js only exposes the section mixin.

Of course, if people would extend the structure field component, their stuff would break. Maybe we need to start taking the Vue side there more serious. I don't think we can avoid many of the braking changes there, as the code base isn't as mature yet. But we probably should start listing them. Wouldn't be an issue here as this is probably more a 3.7.0 PR anyways.

bastianallgeier (approved):

PR #4216, opened 2022-03-22
Add PHP extensions to `composer.json`

distantnative:
### Enhancements
- Adds PHP dependencies to `composer.json`

### Breaking changes
- Could break some setups which install Kirby via composer

## Ready?

### For review team


lukasbestle:
@distantnative I haven't reviewed the changes yet, but I think it makes sense to get this merged given that you already have the code lying around. If you agree, please reopen and I'll take a look, maybe for 3.6.6?

distantnative:
@lukasbestle Reopened - I closed it initially as I kept running into issues with our tests - and then it didn't feel worth it to me to spend too much time on fixing it for the benefit of this PR.

lukasbestle:
Thinking about it again, I think we can only merge this in 3.7.0.

Technically there is no breaking change as Kirby has always required these extensions, but it may break deployment setups that are using Composer. The PHP CLI can sometimes have a different PHP configuration or even version as the server, so this would then break these deployments.

In general I think it's great. Only needs rebasing and `composer normalize`, but that can happen once we are ready for 3.7.

What do you think?

distantnative:
@lukasbestle the CI kept removing some polyfills as unneeded https://github.com/getkirby/kirby/pull/4216/commits/74e7baf505123b0aa6c49917aae9d2b614c66d8e

But wouldn't it be good to have them in there for systems that don't feature these extensions?

lukasbestle:
@distantnative We can add them to this list to ignore them in CI:

https://github.com/getkirby/kirby/blob/3de403cabcd6755bdd95c159bd282ec63ec14658/composer.json#L63-L67

But I'm not sure if we actually need them. Both of these extensions are enabled by default and need to be explicitly disabled at PHP compile time. So I think we can safely make them required. `iconv` is only used in the polyfill for `mbstring` BTW. So I think we can ignore it completely. 😅

lukasbestle:
Since this PR is now pointed at the 3.7 branch, could you please rebase it against that? Otherwise we'll have all the other commits in here as well.

distantnative:
I tried - when I rebase in Tower, nothing changes. But somehow GitHub doesn't recognise it as already rebased.

lukasbestle:
If the commits are already there, a rebase won't remove them since Git cannot know which commits are your newly added ones and which come from somewhere else. You can delete the extra commits from your branch though.

distantnative:
I'm not sure I can follow - what extra commits? This looks like any other rebased branch to me.
Screen Shot 2022-04-09 at 22 58 10

lukasbestle:
Ah, just saw that `develop` and `release/3.7.0` are currently in line.
I've changed the base of this PR and changed it back again. Seems to have fixed it for some reason.

distantnative:
Perfect, thanks!

bastianallgeier (approved):

Issue #4014, opened 2021-12-08
Top-level options not copied when extending a blueprint


yandodov:
## Description
I have a `site/blueprints/pages/article.yml` blueprint:

```yml
title: Article
num: '{{ page.date.toDate("%Y%m%d%H%M") }}'
tabs:
  ...
```

Then, I extend it in `site/blueprint/pages/article-custom.yml`:

```yml
extends: pages/article
tabs:
  ...
```

However, the `num` option is not copied. When I publish the page, the sort number doesn't update. The `title` option is also not copied and even though that wouldn't be very useful, it's expected, since you use extend.

To fix this, I just need to specify the missing options:

```yml
extends: pages/article
title: Article Custom
num: '{{ page.date.toDate("%Y%m%d%H%M") }}'
tabs:
  ...
```

## Your setup
**Kirby Version**  
3.5.7.1

afbora:
✅

PR #4217, opened 2022-03-22
Fix extending some top level props

afbora:
## This PR …

When parent class constructor run, all extended props assigned/set to `$this->props`, but we're still trying to access old `$props` variable after parent constructor called. So I've fixed with calling `$this->props` instead `$props` local variale, exactly what we do in [FileBlueprint::constructor()](https://github.com/getkirby/kirby/blob/3.6.3/src/Cms/FileBlueprint.php#L28-L48)

### Fixes
- Fixed top-level options not copied when extending a blueprint #4014

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/PageBlueprint.php
+++ b/src/Cms/PageBlueprint.php
@@ -25,7 +25,7 @@ class PageBlueprint extends Blueprint
 
         // normalize all available page options
         $this->props['options'] = $this->normalizeOptions(
-            $props['options'] ?? true,
+            $this->props['options'] ?? true,
             // defaults
             [
                 'changeSlug'     => null,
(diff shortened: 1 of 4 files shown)


distantnative:
But now everything we pass as $props parameter to the constructor would be ignored.

afbora:
`$props` params convert into `$this->props` with `parent::__construct($props);` call. Am I missing?

distantnative:
Oh indeed, very good then

distantnative:
Could you add unit tests that would've failed before but now pass with the changes?

afbora:
I've added `PageBlueprintTest::testExtendNum` unit test. This test will fail if you revert the related changes or test on develop branch.

distantnative (approved):
Looks good to me

Issue #4219, opened 2022-03-22
Pagination no longer working in 3.6.3


texnixe:
## Description
Pagination stopped working in 3.6.3.

**Expected behavior**  
Pagination should work as before.

## To reproduce
In a 3.6.3 Starterkit, modify the photography.php template like this:

```
<?php snippet('header') ?>
<?php snippet('intro') ?>

<ul class="grid" style="--gutter: 1.5rem">
  <?php foreach ($page->children()->listed()->paginate(2) as $project): ?>
  <li class="column" style="--columns: 3">
    <a href="<?= $project->url() ?>">
      <figure>
        <span class="img" style="--w:4;--h:5">
          <?= ($cover = $project->cover()) ? $cover->crop(400, 500) : null ?>
        </span>
        <figcaption class="img-caption">
          <?= $project->title()->html() ?>
        </figcaption>
      </figure>
    </a>
  </li>
  <?php endforeach ?>
</ul>
<?php if ($pagination->hasPages()): ?>
  <nav class="project-pagination grid">
    <?php if ($pagination->hasPrevPage()): ?>
    <a class="column" href="<?= $pagination->prevPageURL() ?>" style="<?= $pagination->hasNextPage() ? '--columns: 6' : '--columns: 12' ?>">
    ‹ previous projects page
    </a>
    <?php endif ?>
    <?php if ($pagination->hasNextPage()): ?>
    <a class="column next" href="<?= $pagination->nextPageURL() ?>" style="<?= $pagination->hasPrevPage() ? '--columns: 6' : '--columns: 12' ?>">
      next projects page  ›
    </a>
    <?php endif ?>
  </nav>
<?php endif ?>
<?php snippet('footer') ?>
```
Open page in browser and click on next page link. See how you are redirected to the home page instead of to the next page.

## Your setup
**Kirby Version**  
Kirby 3.6.3

**Your system (please complete the following information)**
 - Device: MacBook Pro (14", 2021)
 - OS: 
 - Browser: Firefox
 - Version: 98.0.1 (64-Bit)

**Additional context**  
This was originally brought up by the Ajax recipe not working anymore, but turned out that even normal pagination doesn't work.

afbora:
Works great for me, am I missing something on setup?

https://user-images.githubusercontent.com/3393422/159572979-c71b86dd-bd90-4189-84f9-b6a98f363d94.mp4

bastianallgeier:
✅

PR #4221, opened 2022-03-23
Fix path detection in Server::requestUri

bastianallgeier:
See: https://github.com/getkirby/kirby/issues/4219

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
   "description": "The Kirby 3 core",
   "license": "proprietary",
   "type": "kirby-cms",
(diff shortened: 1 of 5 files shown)


afbora (approved):
Works great and looks good to me 👍

bastianallgeier:
@afbora thanks for having a look. I think I would feel better if another pair of eyes checked this out before the patch release, just to be sure. @lukasbestle @distantnative do you have time?

lukasbestle (approved):
LGTM 👍

PR #4236, opened 2022-03-27
Add A::filter, A::without

adamkiss:
I often (twice or more every project) need a simple way to filter an array for (or except) given keys, and every time I'm sad there isn't one included in the `A` Toolkit class, so I added them. Both call another added method for simpler filtering of associative (an indexed too) arrays, without needing to remember `ARRAY_FILTER_USE_BOTH` every time!

This time into the proper branch and based on proper branch…

## This PR …

Adds a `::filter()` (filtering using both value and key), `::without()` (everything except given keys) methods to `Kirby\Toolkit\A`

Edit 2: after review 2 `A::except` was renamed to `A::without` to be in line with collections.
Edit 1: after review, `A::where` was renamed to `A::filter`
Original:
Adds a `::where` (filtering using both value and key), `::only` (filtering for only given keys), `::except` (everything except given keys) methods to `Kirby\Toolkit\A`

## Ready?

I think the in-code documentation is sufficient, but am not sure what is the treshold for "Needs code example"

### For review team

### DIFF
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -741,31 +741,13 @@ class A
     }
 
     /**
-     * Filter the array, using only the given key or array of keys
+     * Remove key(s) from an array
      *
      * @param array $array
      * @param int|string|array $keys
      * @return array
      */
-    public static function only(array $array, $keys): array
-    {
-        if (is_int($keys) || is_string($keys)) {
-            $keys = static::wrap($keys);
-        }
-
-        return static::filter($array, function ($value, $key) use ($keys) {
-            return in_array($key, $keys, true);
-        });
-    }
-
-    /**
-     * Filter the array for ever key except the given key or array of keys
-     *
-     * @param array $array
-     * @param int|string|array $keys
-     * @return array
-     */
-    public static function except(array $array, $keys): array
+    public static function without(array $array, $keys): array
     {
         if (is_int($keys) || is_string($keys)) {
             $keys = static::wrap($keys);
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -726,4 +726,53 @@ class A
             return $array;
         }
     }
+
+    /**
+     * Filter the array using the given callback
+     * using both value and key
+     *
+     * @param array $array
+     * @param callable $callback
+     * @return array
+     */
+    public static function where(array $array, callable $callback): array
+    {
+        return array_filter($array, $callback, ARRAY_FILTER_USE_BOTH);
+    }
+
+    /**
+     * Filter the array, using only the given key or array of keys
+     *
+     * @param array $array
+     * @param int|string|array $keys
+     * @return array
+     */
+    public static function only(array $array, $keys): array
+    {
+        if (is_int($keys) || is_string($keys)) {
+            $keys = static::wrap($keys);
+        }
+
+        return static::where($array, function ($value, $key) use ($keys) {
+            return in_array($key, $keys, true);
+        });
+    }
+
+    /**
+     * Filter the array for ever key except the given key or array of keys
+     *
+     * @param array $array
+     * @param int|string|array $keys
+     * @return array
+     */
+    public static function except(array $array, $keys): array
+    {
+        if (is_int($keys) || is_string($keys)) {
+            $keys = static::wrap($keys);
+        }
+
+        return static::where($array, function ($value, $key) use ($keys) {
+            return in_array($key, $keys, true) === false;
+        });
+    }
 }
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -735,7 +735,7 @@ class A
      * @param callable $callback
      * @return array
      */
-    public static function where(array $array, callable $callback): array
+    public static function filter(array $array, callable $callback): array
     {
         return array_filter($array, $callback, ARRAY_FILTER_USE_BOTH);
     }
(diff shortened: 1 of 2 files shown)


src/Toolkit/A.php
```diff
@@ -726,4 +726,53 @@ public static function wrap($array = null): array
...
+
+    /**
+     * Filter the array using the given callback
+     * using both value and key
+     *
+     * @param array $array
+     * @param callable $callback
+     * @return array
+     */
+    public static function where(array $array, callable $callback): array
```
distantnative: I think calling this `filter` would make more sense and be more in line to the PHP name as well as similar to our collection `->filter($callback)` methods.

src/Toolkit/A.php
```diff
@@ -726,4 +726,53 @@ public static function wrap($array = null): array
...
+    }
+
+    /**
+     * Filter the array, using only the given key or array of keys
+     *
+     * @param array $array
+     * @param int|string|array $keys
+     * @return array
+     */
+    public static function only(array $array, $keys): array
```
distantnative: I find these names very vague and not easy to direct understand what they do. Maybe something more like `A::filterKeys($array, $keys)` or `A::withKeys()` and `A::withoutKeys()`?

src/Toolkit/A.php
adamkiss: This one is actually nod to `Arr::where` from Laravel, but they also support other stuff (dot syntax, etc). I'll switch it to filter

src/Toolkit/A.php
adamkiss: I like `A::filterKeys()` and `A::withoutKeys()`, personally. `A::withKeys()` suggest some other key operation, something like going from `[dog => woof]` to `[key=> dog, value=>woof]` or something.

adamkiss: Perhaps `A::onlyKeys()` and `A::withoutKeys()`? Does that sound good?

adamkiss: Damn, the more I think about it, the more does `A::onlyKeys()` also suggest something else. Naming is really hard, that's why I reused names I used for my personal functions previously :D

adamkiss: 🤦‍♂️ I realised I'm stupid, and `A::only()` is already in the toolkit, except called `A::get()` - if used with an array of keys. So to be in line with collections, which also have `->get()` and `->without()`, I renamed this to `A::without()`

src/Toolkit/A.php
```diff
@@ -726,4 +726,35 @@ public static function wrap($array = null): array
...
+    public static function filter(array $array, callable $callback): array
+    {
+        return array_filter($array, $callback, ARRAY_FILTER_USE_BOTH);
+    }
+
+    /**
+     * Remove key(s) from an array
+     *
+     * @param array $array
+     * @param int|string|array $keys
```
lukasbestle, line 747: Super tiny nitpick: I think this argument should be called `$key` for consistency with `A::get()`.

tests/Toolkit/ATest.php
```diff
@@ -698,4 +698,54 @@ public function testWrap()
         $result = A::wrap(null);
         $this->assertSame([], $result);
     }
+
+
+     /**
+     * @covers ::where
+     */
+    public function testWhere()
```
lukasbestle:
```suggestion
     * @covers ::filter
     */
    public function testFilter()
```

tests/Toolkit/ATest.php
```diff
@@ -698,4 +698,54 @@ public function testWrap()
...
+    {
+        $associativeArray = $this->_array();
+        $array = array_keys($associativeArray);
+
+        $array = [...array_keys($associativeArray), ...array_keys($associativeArray)];
+
+        $this->assertSame(['dog' => 'wuff', 'bird' => 'tweet'], A::without($associativeArray, 'cat'));
+        $this->assertSame(['dog' => 'wuff'], A::without($associativeArray, ['cat', 'bird']));
+        $this->assertSame([], A::without($associativeArray, ['cat', 'dog', 'bird']));
+        $this->assertSame($associativeArray, A::without($associativeArray, ['this', 'doesnt', 'exist']));
```
lukasbestle: Another case with a mixture of existing and non-existing methods:

```suggestion
        $this->assertSame(['dog' => 'wuff', 'bird' => 'tweet'], A::without($associativeArray, ['this', 'cat', 'doesnt', 'exist']));
```

lukasbestle (commented):
I love the two methods. Really useful and simple, thanks for your contribution. 💛

src/Toolkit/A.php
adamkiss: @lukasbestle yes, but `$collection->without()` argument is called `$keys` 👀

tests/Toolkit/ATest.php
adamkiss: @lukasbestle Don't quite understand? Updated the tests a bit anyway (according to Github suggestion), but now I realise that might not have been what you meant.

bastianallgeier (dismissed):
LGTM, but @lukasbestle should have another look

tests/Toolkit/ATest.php
lukasbestle: What I meant is that the test should also cover the case where the `$keys` include a key that exists but also keys that don't exist. This way the tests cover a broader range of cases to avoid future regressions.

I've committed my suggestion to the PR branch.

lukasbestle (approved):
I think it's ready to merge now.

@distantnative Should we merge it into the 3.7.0 branch or already into `develop` for 3.6.5?

distantnative:
Let's do 3.6.5

tests/Toolkit/ATest.php
adamkiss: @lukasbestle oh, you meant _keys_, right-o! Thanks for committing it

Issue #4194, opened 2022-03-12
Fix inconsistencies in properties naming 


tobimori:
## Description
With a codebase that has grown over time, elements' properties, especially when writing Kirby blueprints, have generated quite some inconsistencies in naming. The most prominent example would be in naming elements:

**Example**
- Blueprints use `title`
- Fields use `label`
- Blocks use `name`
- Sections use `headline`, while the field section has no such property

**Expected behavior**  
The user would expect to be able to set the name of such elements with the same property. Especially sections using `headline` is often confused with a Fields' `label`, while both even share the same look and feel.

**Additional context**  
While it might not be possible to effectively change all properties' names, as such would cause a huge breaking change for most Kirby blueprints, it'd be possible to alias several properties to the computed (translated) props. An example for the sections' headline mixin would be:

```php
<?php

use Kirby\Toolkit\I18n;

return [
    'props' => [
        /**
         * The headline for the section. This can be a simple string or a template with additional info from the parent page.
         */
        'headline' => function ($headline = null) {
            return I18n::translate($headline, $headline);
        },
        'label' => function ($label = null) {
            return I18n::translate($label, $label);
        }
    ],
    'computed' => [
        'headline' => function () {
            if ($this->headline) {
                return $this->model()->toString($this->headline);
            }

            if ($this->label) {
                return $this->model()->toString($this->label);
            }


            return ucfirst($this->name);
        }
    ]
];
```

There might be more inconsistencies, although, those, like the ones mentioned above for naming elements, are the properties I always confuse while working with blueprints.

> Follow up of discussion [on Discord](https://discord.com/channels/525634039965679616/525641819854471168/950499993209413742)

bastianallgeier:
We've now fixed the inconsistencies by introducing aliases. Here's what we have now: 

- Blueprint: title
- Block: title & label with two individual functions
- Field: label
- Section: label

The old props will keep on working but will be deprecated eventually.

mrflix:
@tobimori thanks for bringing this topic up and thanks @bastianallgeier and team that you adjusted things. Being able to use "label" instead of "headline" for sections was a wonderful change. These days I still often run into accidentally naming blocks using "label". I then realize it when I try to add the new block:
![Screenshot 2024-01-12 at 17 26 31](https://github.com/getkirby/kirby/assets/60777/0d290855-e03d-4a14-a299-bbbe0997bfc7)
[...]

PR #4240, opened 2022-03-28
Support label instead of headline for sections

bastianallgeier:
## Enhancements

- Support `label` as new, better alternative to `headline` for sections https://github.com/getkirby/kirby/issues/4194

## Ready?

### For review team

### DIFF
--- a/config/sections/mixins/headline.php
+++ b/config/sections/mixins/headline.php
@@ -6,15 +6,26 @@ return [
     'props' => [
         /**
          * The headline for the section. This can be a simple string or a template with additional info from the parent page.
+         * @todo deprecate in 3.8
          */
         'headline' => function ($headline = null) {
             return I18n::translate($headline, $headline);
+        },
+        /**
+         * label is the new official replacement for headline
+         */
+        'label' => function ($label = null) {
+            return I18n::translate($label, $label);
         }
     ],
     'computed' => [
         'headline' => function () {
             if ($this->headline) {
                 return $this->model()->toString($this->headline);
+            }
+
+            if ($this->label) {
+                return $this->model()->toString($this->label);
             }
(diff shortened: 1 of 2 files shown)


config/sections/mixins/headline.php
```diff
@@ -6,9 +6,16 @@
     'props' => [
         /**
          * The headline for the section. This can be a simple string or a template with additional info from the parent page.
+         * @todo deprecate in 3.8
```
distantnative: Let's already deprecate it in 3.7 - we can still keep it around for longer and remove it in 3.9 or later. But to already give the signal (just as deprecated code comment and release notes) in 3.7

config/sections/mixins/headline.php
```diff
@@ -6,9 +6,16 @@
...
          */
         'headline' => function ($headline = null) {
             return I18n::translate($headline, $headline);
+        },
+        /**
+         * label is the new official replacement for headline
+         */
+        'label' => function ($label = null) {
```
distantnative, line 17: One question (with no answer myself): if we talk about making things more consistent and fixing `name` with `title`... should it all be `title` eventually?

bastianallgeier: No, I don't think we necessarily have to go for label or title everywhere. 

After this fix we have the following: 

- Blueprint: title
- Block: title & label
- Field: label
- Section: label

Blocks have title and label with two individual functions. The title is shown in the block selector and is THE title for the block. I.e. `Image`. The label is used in the block preview and can be more descriptive with template strings. 

I think this is already pretty consistent.

distantnative: 👍

distantnative (approved):

PR #4250, opened 2022-04-02
Force `Kirby\Toolkit\Collection` iterator for `Query::page` method

### LINKED ISSUE #2738: DB Pagination Bug
**Describe the bug**  
I just ran into an issue where I'm able to paginate DB queries but not able to access the proper pagination object from the returned collection.

**To Reproduce**  
1. Create a DB query and use `$query->page(…)` to paginate
2. Try to fetch the pagination object from the returned collection
3. The pagination object differs from the one created in the `page` method

I think the problem is right about [here](https://github.com/getkirby/kirby/blob/master/src/Database/Query.php#L789). Not sure if this was possible at some point or not. But as far as I can tell it is not possible to set the pagination object with the `$collection->paginate()` right now.

**Expected behavior**  
I should be able to get the proper pagination object from the collection.

**Kirby Version**  
3.3.5

afbora:
### Fixes

- Fixed DB Pagination bug for `$query->page()` method #2738

### Breaking changes

I'm not sure possible breaking changes..

## Ready?

### For review team

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -775,7 +775,11 @@ class Query
 
         // apply it to the dataset and retrieve all rows. make sure to use Collection as the iterator to be able to attach the pagination object
         $iterator   = $this->iterator;
-        $collection = $this->offset($pagination->offset())->limit($pagination->limit())->iterator('Collection')->all();
+        $collection = $this
+            ->offset($pagination->offset())
+            ->limit($pagination->limit())
+            ->iterator('Kirby\Toolkit\Collection')
+            ->all();
 
         $this->iterator($iterator);
 
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -742,7 +742,14 @@ class Collection extends Iterator implements Countable
      */
     public function paginate(...$arguments)
     {
-        $this->pagination = Pagination::for($this, ...$arguments);
+        // set pagination property if passed first argument
+        // already a pagination object, otherwise create new
+        // pagination instance with passed arguments
+        if (is_a($arguments[0], 'Kirby\Toolkit\Pagination') === true) {
+            $this->pagination = $arguments[0]->clone();
+        } else {
+            $this->pagination = Pagination::for($this, ...$arguments);
+        }
 
         // slice and clone the collection according to the pagination
         return $this->slice($this->pagination->offset(), $this->pagination->limit());
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -742,14 +742,7 @@ class Collection extends Iterator implements Countable
      */
     public function paginate(...$arguments)
     {
-        // set pagination property if passed first argument
-        // already a pagination object, otherwise create new
-        // pagination instance with passed arguments
-        if (is_a($arguments[0], 'Kirby\Toolkit\Pagination') === true) {
-            $this->pagination = $arguments[0]->clone();
-        } else {
-            $this->pagination = Pagination::for($this, ...$arguments);
-        }
+        $this->pagination = Pagination::for($this, ...$arguments);
 
         // slice and clone the collection according to the pagination
         return $this->slice($this->pagination->offset(), $this->pagination->limit());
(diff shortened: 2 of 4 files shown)


lukasbestle (commented):
Looks good to me. 👍

Can you add unit test assertions for the pagination values like `$pagination->total()`, `$pagination->page()` and `$pagination->start()`?

afbora:
I think some potential bugs came up when I tried to add the pagination tests. I've solutions but I'm not sure. Need your thoughts:

## Sending/Accepting  pagination object

We send direct pagination object to `->paginate()` method:
https://github.com/getkirby/kirby/blob/fix/2738-db-pagination/src/Database/Query.php#L794

But as far as I see, pagination object doesn't accepted:
https://github.com/getkirby/kirby/blob/fix/2738-db-pagination/src/Toolkit/Pagination.php#L78-L114

I solved this is with accepting direct pagination object. Also we can do this in `Pagination::for()` with returning self.
https://github.com/getkirby/kirby/blob/fix/2738-db-pagination/src/Toolkit/Collection.php#L748-L752

## Assert instance of with inheritance

I think, I couldn't correct unit test for `assertInstanceOf` because of inheritance (`Kirby\Cms\Pagination` always passed for `assertInstanceOf`, but shouldn't). I solved with using `get_class` method but I'm not sure this is right way.

https://github.com/getkirby/kirby/blob/fix/2738-db-pagination/tests/Database/QueryTest.php#L406

distantnative:
### Sending/Accepting pagination object

I would rather fix this in `Pagination::for`. Something like

```
if (is_a($collection, static::class) === true) {
    return $collection;
}
```

### Assert instance of with inheritance

It seems like for all our tests, `assertInstanceOf` should also pass. Right now you have three tests with `assertTrue`.

afbora:
> It seems like for all our tests, assertInstanceOf should also pass. Right now you have three tests with assertTrue.

@distantnative I don't understand what you mean. In following unit test, first assertion passed, but need to failed to compare strict and correct unit tests. That's why I used `get_class` method.

````php
$pagination = new \Kirby\Cms\Pagination();

$this->assertInstanceOf('\Kirby\Toolkit\Pagination', $pagination); // passed, not expected
$this->assertTrue(get_class($pagination) === 'Kirby\Toolkit\Pagination'); // failed, expected
````

lukasbestle:
@afbora A `Kirby\Cms\Pagination` object is also a `Kirby\Toolkit\Pagination` object by inheritance, which is why the "instance of" assertion passes. Could there be a case where the `Cms\Pagination` class misbehaves for our purposes while a `Toolkit\Pagination` works? If that was the case, then `Cms\Pagination` would have a bug as that would break the Liskov Substitution Principle.

afbora:
> A Kirby\Cms\Pagination object is also a Kirby\Toolkit\Pagination object by inheritance, which is why the "instance of" assertion passes.

I know that. I just wanted to make sure it was returning the correct class. Are you saying it doesn't matter?

> Could there be a case where the Cms\Pagination class misbehaves for our purposes while a Toolkit\Pagination works? If that was the case, then Cms\Pagination would have a bug as that would break the Liskov Substitution Principle.

There seems to be no such issue. The `Kirby\Cms\Pagination` class does not override any method (except constructor) of the parent class anyway.

afbora:
To summarize, this PR solves two issues:

1. Now sends correct iterator
2. A possible bug that has not occured yet.  The `Pagination` class wasn't accepting an object of its type.

distantnative:
But it seems to matter which type of Collection class is used. Isn't that the initial problem?

afbora:
@distantnative The initial issue isn't about behavior on same method. The `->page()` and the `->paginate()` method return different pagination objects. 

````php
$query = Db::select('google_countries');
$results = $query->paginate(2);
$pagination = $results->pagination();
var_dump($pagination); // returns \Kirby\Toolkit\Pagination

$query = Db::table('google_countries');
$results = $query->page(3, 2);
$pagination = $results->pagination();
var_dump($pagination); // returns \Kirby\Cms\Pagination
````

We fixed the inconsistency with this PR.

bastianallgeier (approved):

Issue #2831, opened 2020-09-17
DB where Callback Bug


lukaskleinschmidt:
**Describe the bug**  
Using a callback in the where clause takes on everything defined in the query.
```php
$query->where('id', 1);
$query->where(fn ($q) => $q->where('status', 0));
```
Will produce something like this:
```sql
WHERE id = :value_1
AND (
    id = :value_1
    AND status = :value_2
)
```
The problem is that the binding is cloned as well and PDO allows a binding to be used only once.
Following error will be thrown `SQLSTATE[HY093]: Invalid parameter number: parameter was not defined`.

The reason for that is probably the clone [here](https://github.com/getkirby/kirby/blob/7525553b8d9976d6ed08b702f62fc3d368116777/src/Database/Query.php#L970). Not sure if this should simply be a new query instance or if there is actually some stuff that needs be cloned.

**Kirby Version**  
3.3.5

lukasbestle:
✅

PR #4255, opened 2022-04-06
Fix DB where callback with reset props 

afbora:
## This PR …

I'm not sure about solution but since we only need `where` prop in callback, I've fixed the issue with reset `where` prop ~via new `clone` method~. Need your review and feel free to close the PR if not acceptable.

Btw I couldn't write unit test due to can't get query string. But I've tested and query string is correct as expected.

### Fixes

- Fixed DB where callback #2831

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -972,6 +972,11 @@ class Query
                     $this->bindings($sql['bindings']);
                 } elseif (is_callable($args[0]) === true) {
                     $query = clone $this;
+
+                    // since the callback uses its own where condition
+                    // it is necessary to clear/reset the cloned where condition
+                    $query->where = null;
+
                     call_user_func($args[0], $query);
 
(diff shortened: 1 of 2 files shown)


afbora:
If new clone method is not necessary, another solution is simple with clearing cloned query `where` prop:

````php
$query = clone $this;
$query->where = null;
call_user_func($args[0], $query);
````
https://github.com/getkirby/kirby/blob/3.6.4/src/Database/Query.php#L970-L971

distantnative:
Maybe easier to solve without new method? Or at least right now the code coverage check still fails.

afbora:
Finally fixed the unit test issue 😅 Ready to review 🚀

distantnative (approved):

lukasbestle (approved):
Really simple and effective solution. 👍

PR #4260, opened 2022-04-07
Set router callbacks on initialisation 

distantnative:
## This PR …

We added as a quick fix that `App::call()` keeps setting and unsetting the router hooks.
This PR removes the callbacks as static properties and rather as specific properties of the Router instance that get passed on initialisation.

### Refactoring
- Improved internal usage of router callbacks

### Breaking changes
- `Http\Router::$beforeEach` and `Http\Router::$afterEach` aren't static anymore. Pass them as 2nd and 3rd parameter to the constructor instead.

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -329,26 +329,9 @@ class App
      */
     public function call(string $path = null, string $method = null)
     {
-        $router = $this->router();
-
-        /**
-         * @todo Closures should not defined statically but
-         * for each instance to avoid this constant setting and unsetting
-         */
-        $router::$beforeEach = function ($route, $path, $method) {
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):
So much better!

lukasbestle:
I'm thinking we could make it an array like so:

```php
$hooks = [
    'before' => ...,
    'after'  => ...
];

$router = new Router($routes, $hooks);
```

This would allow us to extend the set of hooks in the future without having to add more and more arguments.

distantnative:
No objections from my side, @bastianallgeier ?

bastianallgeier:
Yep, sounds good!

lukasbestle:
I'll create a PR. :)

PR #4262, opened 2022-04-08
Prevent upload same file with different template

### LINKED ISSUE #4076: Duplicate file uploads possible when coming from files section and files field
## Description

Following my [post on Nolt](https://kirby.nolt.io/388): 

When I upload a file from a files field that was already uploaded on that page before, the previously uploaded file gets replaced (or at least the creation date and time are updated), but existing metadata is attached to the new file.

This can lead to unwanted behavior. In my case, I uploaded a file from a files field that is set to use a specific template (using the query and uploads option). But that file was already added to the page using a files section—set to a different template.
The new (same) file was uploaded, but it was assigned the existing field values of the previous file including e. g. the template name, without any info or warning.

It works correctly if the file only has the same name, but not if it’s the exact same file.

**Expected behavior**  

Prevention of duplicate file uploads in that case.

## To reproduce

1. Create a blueprint having a files section and a files field set to use different templates:

```yaml
title: Default Page
sections:
  filesSection:
    type: files
    template: file-a
  fieldsSection:
    type: fields
    fields:
      filesField:
        type: files
        query: page.files.filter('template', 'file-b')
        uploads: file-b
```
2. Upload a file in the files section – it will use `file-a` as template as expected.
3. Upload that same file using the files field.

## Your setup
**Kirby Version**  

3.6.1.1

 - Device: MacBook Pro
 - OS: macOS 12.1
 - Browser: Google Chrome
 - Version: 96.0.4664.110

distantnative:
### Fixes
- When uploading the same file again (same filename, same content) an error is thrown when the assigned template doesn't match with the existing file
https://github.com/getkirby/kirby/issues/4076

## Ready?

### For review team

### DIFF
--- a/src/Cms/FileRules.php
+++ b/src/Cms/FileRules.php
@@ -80,28 +80,30 @@ class FileRules
      */
     public static function create(File $file, BaseFile $upload): bool
     {
-        // first we want to ensure that we are not trying to
-        // create a duplicate file. So if a file
-        // with the same name already exists
+        // We want to ensure that we are not creating duplicate files.
+        // If a file with the same name already exists
         if ($file->exists() === true) {
             // $file will be based on the props of the new file,
-            // so we need to get the props of the existing file
-            $existing = kirby()->file($file->id());
+            // to compare templates, we need to get the props of
+            // the already existing file from meta content file
+            $existing = $file->parent()->file($file->filename());
 
             // if the new upload is the exact same file
-            // and uses the same template, we can continue,
-            // otherwise throw an eroror for duplicate file
+            // and uses the same template, we can continue
             if (
-                $file->sha1() !== $upload->sha1() ||
-                $file->template() !== $existing->template()
+                $file->sha1() === $upload->sha1() &&
+                $file->template() === $existing->template()
             ) {
-                throw new DuplicateException([
-                    'key'  => 'file.duplicate',
-                    'data' => [
-                        'filename' => $file->filename()
-                    ]
-                ]);
+                return true;
             }
+
+            // otherwise throw an error for duplicate file
+            throw new DuplicateException([
+                'key'  => 'file.duplicate',
+                'data' => [
+                    'filename' => $file->filename()
+                ]
+            ]);
         }
 
         if ($file->permissions()->create() !== true) {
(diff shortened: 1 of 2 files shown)


src/Cms/FileRules.php
```diff
@@ -80,8 +80,21 @@ public static function changeSort(File $file, int $sort): bool
...
     public static function create(File $file, BaseFile $upload): bool
     {
+        // first we want to ensure that we are not trying to
+        // create a duplicate file. So if a file
+        // with the same name already exists
         if ($file->exists() === true) {
-            if ($file->sha1() !== $upload->sha1()) {
+            // $file will be based on the props of the new file,
+            // so we need to get the props of the existing file
+            $existing = kirby()->file($file->id());
```
distantnative: If somebody knows a better way for this, would be great

afbora: Accessing the file via the parent is faster (microseconds 😅) than accessing from the Kirby instance. I'm not sure which one more stable.

````php
$existing = $file->parent()->file($file->filename());
````

File: src/Cms/FileRules.php [hunk 2]
```diff
@@ -80,8 +80,21 @@ public static function changeSort(File $file, int $sort): bool
...
+
+            // if the new upload is the exact same file
+            // and uses the same template, we can continue,
+            // otherwise throw an eroror for duplicate file
```
afbora:
```suggestion
            // otherwise throw an error for duplicate file
```

afbora (dismissed):

tests/Cms/Files/FileRulesTest.php
```diff
@@ -113,6 +122,69 @@ public function testCreateExistingFile()
...
+            'parent' => $page,
+            'content' => [
+                'template' => 'cover'
+            ]
+        ]);
+
+        $upload = new BaseFile($testImage);
+        FileRules::create($newFile, $upload);
+
+        // same filename, same template and different file (sha1)
```
distantnative: I think this needs to be in a separate method as the method ends when the first exception is thrown.

afbora: So do we need:

- 3 unit tests? For `same file`, `same file and different template` and `same filename, same template and different file (sha1)`? 

- 2 unit tests? For `true` and `exceptions`.

distantnative: Maybe best to put them in three methods, yes

afbora (approved):

PR #4263, opened 2022-04-10
Cache: Handle `set()` in the core and add `getObject()` method

### LINKED ISSUE #3976: Disable cache on active sessions only
Following up from a long discussion over at https://discord.com/channels/525634039965679616/525641819854471168/914572838143393792

PR #3292 introduced a new rule in 3.5.5 where any call to the `kirby()->session()` method apparently disables the page cache and sends a "absolutely no caching" header to user agents.

As presented in the Discord discussion, this apparently will be triggered by code that merely checks whether a session is present or not – essentially disabling page cache and browser cache for any page that for instance checks for `if(kirby()->user())`, even if there is no active session. This also creates a risk of inadvertently disabling caching by installing a third-party plugin that uses such check somewhere.

Could this rule be "tuned down" by only applying these rules when an active session exists (as the title of the PR indicated "Disable cache on **active** sessions"), rather than triggering it already by the check for an active session? Without deeper knowledge of the considerations that led to this change, it seems that no critical content should be ending up in cache if the rendered page is the "public version" in the first place?

CC @bvdputte

lukasbestle:
### Features

- New `$cache->getObject()` method to get a full `Value` object from the cache, not just the pure value
- New `$cache->store()` method to store a `Value` object (needs to be implemented by cache drivers)

### Deprecated

- Cache drivers should no longer implement the `set()` method starting with Kirby 3.8. Instead, they should implement the new `store()` method that receives the `Value` object instead of the pure cache value and minutes. For compatibility with older Kirby versions, you can still implement `set()` for the time being:

```php
/**
 * Compatibility layer for older Kirby versions before 3.7
 * @todo Remove this method for Kirby 3.8 support
 */
public function set(string $key, $value, int $minutes = 0): bool
{
    $value = \Kirby\Cache\Value::fromArray(compact('value', 'minutes'));

    return $this->store($key, $value);
}
```

## Explanation

I had planned to add support for a `metadata()` array to the `Cache\Value` class to fix #3976. This is now no longer necessary, but I was thinking if we should make the cache drivers independent from the `Cache\Value` implementation anyway. They would just always deal with the `Value` object and its `toJson()` and `fromJson()` methods. But currently the `set()` method doesn't accept the `Value` object, so this PR changes that to make the behavior consistent.

I've also decided to remove the `TODO` comment for the `Value::$minutes` attribute. I feel like the name is totally fine as just `$expiry` would be unclear (is it a timestamp? Or minutes? Or seconds?). The combination of the timestamp and minutes is not super optimal, but any change would bring more breaking changes. So I think we should just keep it named `$minutes`.

## Alternatives

- We can leave it like it is if we decide that we don't need the `Value` flexibility and rather want to avoid the breaking change.
- I've found out about the [PSR-16 caching standard](https://www.php-fig.org/psr/psr-16/). This standard defines a rather simple interface for caching libraries. By switching to this interface, we can make it much easier for users to plug in third party caching libraries without having to implement our own `Cache` class structure.

Before working on this further, I think we need to decide which route we want to go. I currently feel like we should go for either of these alternatives. So either we don't do anything at all or we use the opportunity to switch to PSR-16 rather than making changes to our own class structure.

## Ready?

### For review team

### DIFF
--- a/src/Cache/ApcuCache.php
+++ b/src/Cache/ApcuCache.php
@@ -57,6 +57,7 @@ class ApcuCache extends Cache
      * Internal method to retrieve the raw cache value;
      * needs to return a Value object or null if not found
      *
+     * @internal
      * @param string $key
      * @return \Kirby\Cache\Value|null
      */
@@ -66,21 +67,16 @@ class ApcuCache extends Cache
     }
 
     /**
-     * Writes an item to the cache for a given number of minutes and
+     * Internal method to store the raw cache value;
      * returns whether the operation was successful
      *
-     * <code>
-     *   // put an item in the cache for 15 minutes
-     *   $cache->set('value', 'my value', 15);
-     * </code>
-     *
+     * @internal
      * @param string $key
-     * @param mixed $value
-     * @param int $minutes
+     * @param \Kirby\Cache\Value $value
      * @return bool
      */
-    public function set(string $key, $value, int $minutes = 0): bool
+    public function store(string $key, Value $value): bool
     {
-        return apcu_store($this->key($key), (new Value($value, $minutes))->toJson(), $this->expiration($minutes));
+        return apcu_store($this->key($key), $value->toJson(), $value->expires() ?? 0);
     }
 }
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -100,7 +100,7 @@ abstract class Cache
     abstract public function retrieve(string $key);
 
     /**
-     * Gets an item from the cache
+     * Gets an item from the cache unless it is expired
      *
      * <code>
      *   // get an item from the cache driver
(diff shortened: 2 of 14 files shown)


distantnative:
Two thoughts:

1. In general, I would be a fan to comply to a standard such as PSR-16. But I am afraid how much of a breaking change would it be.
2. Shouldn't we maybe then directly pass the JSON to the cache driver? Then it doesn't need to worry about the `Value` class either.

lukasbestle:
1. I think the only breaking change would be for custom cache drivers. We could ease the transition by allowing both types of cache drivers (auto-detection based on the interface/base class that the driver implements).
2. Good idea, I think that would certainly be easiest. One question we should think about: Are there any cache drivers that don't already use a (JSON) string to store the value, for example by storing the whole object in some way? Those would be the only ones we would need to worry about, for every other driver it would basically be a 1:1 change without any implications on performance etc.

distantnative:
1. Easing it with some auto-detection would make it a lot less painful (even though I also think it's super limited how many would be affected). So I'd be all for standard then.
2. Not really a good insight into if that's a use case. Could see it either way, unsure what's better.

lukasbestle:
Pinging @afbora and @bastianallgeier for general opinions. Does not need to be detailed yet, just your feeling for the general direction.

afbora:
Sorry, I don't have enough knowledge to give direction on this matter 🤷‍♂️

lukasbestle:
Closing in favor of https://github.com/getkirby/backlog/issues/84 (PSR-16 implementation).

PR #4264, opened 2022-04-10
`Http\Request`: Extensions for the auth types; more secure lookup

lukasbestle:
### Features

- `$request->auth()` now also handles Kirby's `Session` auth type automatically
- The auth types for `$request->auth()` can now be extended by plugins:

```php
Kirby\Http\Request::$authTypes['digest'] = 'Superwoman\Digest\Class';
```

### Breaking changes

- Custom auth type classes in the `Kirby\Http\Request\Auth\` namespace now need to be registered in the `Kirby\Http\Request::$authTypes` array. We also recommend to move the classes to your own namespace to avoid interference with core classes.

## Explanation

My main motivation was actually to get rid of the dynamic class lookup and replace it with the fixed list. This avoids weird edge cases where one could somehow instantiate arbitrary objects. Unlikely, but now this risk is gone.

## Ready?

### For review team

### DIFF
--- a/src/Http/Request.php
+++ b/src/Http/Request.php
@@ -24,6 +24,7 @@ class Request
     public static $authTypes = [
         'basic'   => 'Kirby\Http\Request\Auth\BasicAuth',
         'bearer'  => 'Kirby\Http\Request\Auth\BearerAuth',
+        'session' => 'Kirby\Http\Request\Auth\SessionAuth',
     ];
 
     /**
--- a/src/Http/Request.php
+++ b/src/Http/Request.php
@@ -2,8 +2,6 @@
 
 namespace Kirby\Http;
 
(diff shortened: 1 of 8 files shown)


distantnative:
All for it! But probably needs to mention breaking change that a plug-in with a custom one needs now to add itself to the static prop.

lukasbestle:
Oh, right. Didn't think about that. The question is: Should we move this to 3.7 then? I feel like there are no plugins for this at the moment, but of course there could be.

distantnative:
I'd say better be safe, even though I know it's annoying to delay. But gives our users reliability.

lukasbestle:
Added the breaking change to the PR description and moved the milestone.

bastianallgeier (approved):
LGTM!

PR #4266, opened 2022-04-10
Fix database query for non-existing columns 

### LINKED ISSUE #4265: Database generates invalid insert command if the there's an non-existent column data
## Description

If you pass an array with one or more keys that don't exist as columns on a table, kirby generates invalid sql by dropping the column/keys, but keeping the commas and values

**Expected behavior**  

- don't validate the columns at all and let database engine report the failure properly (`invalid column XYZ` vs `error near ','`)
- or clean the data properly, and remove the binding/value as well

## To reproduce

Tested with SQLite:
1. Create a simple SQLite table with fields "id" and "name":
``` sql
CREATE TABLE "kirby_db_test" ("id" integer,"name" TEXT, PRIMARY KEY (id));
```

2. try to insert an array with excess data3. 
``` php
Db::insert('kirby_db_test', [
	'id' => 1,
	'doesnt_exist' => 'oh no',
	'name' => 'test'
]);
```

3. Observe:
- `Db::lastQuery()` - `INSERT INTO "kirby_db_test" ("id", , "name") VALUES (:value_YK5BnEGI, :value_PILOWDiC, :value_X3mUasTS)`
- `Db::lastError()` - `SQLSTATE[HY000]: General error: 1 near ",": syntax error"`

## Your setup
**Kirby Version**  
[3.6.2](https://github.com/getkirby/kirby/releases/tag/3.6.2)

 - PHP v8.0.15
 - Sqlite v3.37.2

afbora:
## This PR …

Ignores non-existing columns.

### Fixes
- Database generates invalid insert command if the there's an non-existent column data #4265

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Database/Sql.php
+++ b/src/Database/Sql.php
@@ -853,8 +853,14 @@ abstract class Sql
         $query    = [];
         $bindings = [];
 
-        foreach ($values as $key => $value) {
-            $fields[] = $this->columnName($table, $key, $enforceQualified);
+        foreach ($values as $column => $value) {
+            $key = $this->columnName($table, $column, $enforceQualified);
+
+            if ($key === null) {
+                continue;
+            }
+
+            $fields[] = $key;
 
             if (in_array($value, static::$literals, true) === true) {
                 $query[] = $value ?: 'null';
@@ -895,6 +901,10 @@ abstract class Sql
 
         foreach ($values as $column => $value) {
             $key = $this->columnName($table, $column, $enforceQualified);
+
+            if ($key === null) {
+                continue;
+            }
 
             if (in_array($value, static::$literals, true) === true) {
                 $query[] = $key . ' = ' . ($value ?: 'null');
--- a/tests/Database/SqlTest.php
+++ b/tests/Database/SqlTest.php
@@ -459,4 +459,38 @@ class SqlTest extends TestCase
             'bindings' => []
         ], $this->sql->having('test < :value'));
     }
+
+    public function testValueSet()
+    {
+        $this->database->createTable('users', [
+            'name' => ['type' => 'varchar']
+        ]);
+
+        $values = $this->sql->values('users', [
+            'name' => 'John Doe',
+            'nonexisting' => 'test',
+        ], ', ', true);
+
+        $this->assertArrayHasKey('query', $values);
+        $this->assertArrayHasKey('bindings', $values);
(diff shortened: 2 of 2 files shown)


src/Database/Sql.php
```diff
@@ -896,6 +902,10 @@ public function valueSet(string $table, $values, string $separator = ',', bool $
         foreach ($values as $column => $value) {
             $key = $this->columnName($table, $column, $enforceQualified);
 
+            if ($key === null) {
+                continue;
+            }
+
             if (in_array($value, static::$literals, true) === true) {
                 $query[] = $key . ' = ' . ($value ?: 'null');
```
afbora, line 910: I have a question independent of this issue. Shouldn't `x IS NULL` instead of `x = null`?

lukasbestle: Could be, yes. I think both should be equivalent, right? The current code is pretty short.

afbora: I had this problem before. As far as I know, `x = null` is not equal with `x IS NULL`. I've googled now:

> “= NULL” is an expression of value. Meaning, if the variable has been set and memory created for the storage of data it has a value. A variable can in fact be set to NULL which means the data value of the objects is unknown. If the value has been set like so:
> 
>  “IS NULL” is a little trickier and is the preferred method for evaluating the condition of a variable being NULL. When you use the “IS NULL” clause, it checks both the address of the variable and the data within the variable as being unknown.
>
> If summary unless you need to check that the value of a variable was set to equal NULL and you have set ANSI_NULLS ON, then always use the “IS NULL” clause  to validate if a variable is NULL. By using = NULL instead you can cause yourself a lot of headaches in trying to troubleshoot issues that may arise from it, now or unexpectedly in the future.

afbora: I can create a separate PR for this bug if you want.

afbora: By the way, I guess this is valid for the `where` condition.

lukasbestle: You are right, `NULL` is a special value that doesn't equal anything, so it cannot be compared to a value.
I think a separate PR with a quick unit test makes sense.

afbora: I will 👍

bastianallgeier (approved):

PR #4268, opened 2022-04-11
`Toolkit\Obj`: support `array` as the `$property` argument for getting multiple properties

adamkiss:
### Features

Added support for getting multiple properties from `Toolkit\Obj` objects and derived objects, like so:

```php
$thing = new Obj(['one' => '👋', 'two' => 'Kirby']);

$properties = $thing->get(['one', 'three'], ['three' => 'fallback']);
// results in ['one' => '👋', 'three' => 'fallback']
```

## Ready?

### For review team

~Because it uses `A::get()` internally, getting non-existent properties in the array returns `['no' => null]`, which I'm not quite sure how to handle - it makes sense from a consistency POV, but might not be expected. I didn't want to pollute the PR with more robust edge case handling.~

### DIFF
--- a/src/Toolkit/Obj.php
+++ b/src/Toolkit/Obj.php
@@ -73,6 +73,10 @@ class Obj extends stdClass
     public function get($property, $fallback = null)
     {
         if (is_array($property)) {
+            if ($fallback === null) {
+                $fallback = [];
+            }
+
             if (!is_array($fallback)) {
                 throw new InvalidArgumentException('The fallback value must be an array when getting multiple properties');
             }
--- a/src/Toolkit/Obj.php
+++ b/src/Toolkit/Obj.php
@@ -64,12 +64,17 @@ class Obj extends stdClass
     /**
      * Property Getter
      *
-     * @param string $property
+     * @param string|array $property
      * @param mixed $fallback
      * @return mixed
      */
-    public function get(string $property, $fallback = null)
+    public function get(string|array $property, $fallback = null)
     {
+        if (is_array($property)) {
+            $filtered = A::get($this->toArray(), $property);
+            return ! empty($filtered) ? $filtered : $fallback;
+        }
+
         return $this->$property ?? $fallback;
     }
 
--- a/src/Toolkit/Obj.php
+++ b/src/Toolkit/Obj.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Toolkit;
 
+use Kirby\Exception\InvalidArgumentException;
 use stdClass;
 
 /**
@@ -71,8 +72,15 @@ class Obj extends stdClass
     public function get(string|array $property, $fallback = null)
     {
         if (is_array($property)) {
-            $filtered = A::get($this->toArray(), $property);
-            return ! empty($filtered) ? $filtered : $fallback;
+            if (!is_array($fallback)) {
+                throw new InvalidArgumentException('The fallback value must be an array when getting multiple properties');
+            }
+
+            $result = [];
+            foreach($property as $key) {
+                $result[$key] = $this->$key ?? $fallback[$key] ?? null;
+            }
+            return $result;
         }
 
         return $this->$property ?? $fallback;
--- a/src/Toolkit/Obj.php
+++ b/src/Toolkit/Obj.php
@@ -63,13 +63,14 @@ class Obj extends stdClass
     }
 
     /**
-     * Property Getter
+     * Gets one or multiple properties of the object
      *
      * @param string|array $property
-     * @param mixed $fallback
+     * @param mixed $fallback If multiple properties are requested:
+     *                        Associative array of fallback values per key
      * @return mixed
      */
-    public function get(string|array $property, $fallback = null)
+    public function get($property, $fallback = null)
     {
         if (is_array($property)) {
             if (!is_array($fallback)) {
@@ -77,7 +78,7 @@ class Obj extends stdClass
             }
 
             $result = [];
-            foreach($property as $key) {
+            foreach ($property as $key) {
                 $result[$key] = $this->$key ?? $fallback[$key] ?? null;
             }
             return $result;
--- a/tests/Toolkit/ObjTest.php
+++ b/tests/Toolkit/ObjTest.php
@@ -19,6 +19,20 @@ class ObjTest extends TestCase
         $this->assertNull($obj->foo);
     }
 
+    public function test__getMultiple()
+    {
+        $obj = new Obj([
+            'one' => 'first',
+            'two' => 'second',
+            'three' => 'third'
+        ]);
+
+        $this->assertEquals('first', $obj->get('one'));
+        $this->assertEquals(['one' => 'first', 'three' => 'third'], $obj->get(['one', 'three']));
+        $this->assertEquals(['one' => 'first', 'three' => 'third', 'eight' => null], $obj->get(['one', 'three', 'eight']));
+        $this->assertEquals($obj->toArray(), $obj->get(['one', 'two', 'three']));
+    }
+
     public function testToArray()
     {
         $obj = new Obj($expected = [
--- a/tests/Toolkit/ObjTest.php
+++ b/tests/Toolkit/ObjTest.php
@@ -29,8 +29,22 @@ class ObjTest extends TestCase
 
         $this->assertEquals('first', $obj->get('one'));
         $this->assertEquals(['one' => 'first', 'three' => 'third'], $obj->get(['one', 'three']));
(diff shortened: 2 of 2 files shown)


distantnative:
Not sure how the functionality should be

```php
$thing = new Obj(['one' => '👋', 'two' => 'Kirby']);
$properties = $thing->get(['foo', 'bar'], 'or-not');

// result: 'or-not'
// result: ['foo' => 'or-not', 'bar' => 'or-not']
```

adamkiss:
@distantnative In my opinion, it should be `result: 'or-not'`, because in less contrived example, the `$fallback` will probably a default array and that makes more sense.

```php
$props = $thing->get(['id', 'name'], ['id'=>null]);
```

I am open to suggestions though

lukasbestle:
But what happens if only one of the multiple properties doesn't exist? I think the method should then return the actual values for the properties that *do* exist and the default only for the others that don't. What about this:

```php
$thing = new Obj(['one' => '👋', 'two' => 'Kirby']);

$thing->get(['one', 'foo', 'bar'], 'or-not'); // ['one' => '👋', 'foo' => 'or-not', 'bar' => 'or-not']

$fallback = ['one' => '1', 'foo' => 'fooo', 'bar' => 'pretty-high'];
$thing->get(['one', 'foo', 'bar'], $fallback); // ['one' => '👋', 'foo' => 'fooo', 'bar' => 'pretty-high']
```

lukasbestle:
Thinking about it, maybe we shouldn't even support the "non-array fallback" case here. Otherwise you could pass an array by accident with the expectation that this value will be used for all non-existing items. If we always handle it like in my second example and fail on non-array fallbacks, the reliability would be improved.

adamkiss:
@lukasbestle In that case, the whole thing would need to be quite a bit more robust, because simple merge in PHP would result in `null` overwriting the fallback

```php
// example
$obj->get(['key'], ['key' => 'fallback']);

// naive way
A::merge($fallback, A::get($this->toArray(), $property);

// result: ['key' => null]
```

It would need to be something completely different to this naive `A::get($this->toArray())` - which I'm fine with writing and testing, but I worry about the impact.

lukasbestle:
Could be something along these lines:

```php
if (is_array($fallback) !== true) {
    throw new InvalidArgumentException('Fallback must be an array when getting multiple properties');
}

$result = [];
foreach ($property as $key) {
    $result[$property] = $this->$key ?? $fallback[$key] ?? null;
}

return $result;
```

adamkiss:
@lukasbestle Updated as per your idea.

I also updated the test, and I also added a test for the invalid fallback argument, but even if it's according to the docs/example from Kirby suite, it still doesn't validate the exception, but fail. I've no idea what's wrong with the test

src/Toolkit/Obj.php
```diff
@@ -64,12 +65,24 @@ public function __get(string $property)
     /**
      * Property Getter
      *
-     * @param string $property
+     * @param string|array $property
      * @param mixed $fallback
      * @return mixed
      */
-    public function get(string $property, $fallback = null)
+    public function get(string|array $property, $fallback = null)
```
lukasbestle: Union type hints were introduced in PHP 8, we are not quite there yet:

```suggestion
    public function get($property, $fallback = null)
```

src/Toolkit/Obj.php
```diff
@@ -64,12 +65,24 @@ public function __get(string $property)
...
     {
+        if (is_array($property)) {
+            if (!is_array($fallback)) {
+                throw new InvalidArgumentException('The fallback value must be an array when getting multiple properties');
+            }
+
+            $result = [];
+            foreach($property as $key) {
```
lukasbestle:
```suggestion
            foreach ($property as $key) {
```

tests/Toolkit/ObjTest.php
```diff
@@ -19,6 +19,34 @@ public function test__get()
         $this->assertNull($obj->foo);
     }
 
+    public function test__getMultiple()
```
lukasbestle: The tests test the `get()` method, so it should be:

```suggestion
    public function testGetMultiple()
```

tests/Toolkit/ObjTest.php
```diff
@@ -19,6 +19,34 @@ public function test__get()
...
+        $this->assertEquals([
+            'one' => 'first',
+            'three' => 'third',
+            'four' => 'fallback',
+            'eight' => null
+        ], $obj->get(['one', 'three', 'four', 'eight'], ['four' => 'fallback']));
+        $this->assertEquals($obj->toArray(), $obj->get(['one', 'two', 'three']));
+    }
+
+    public function test__getMultipleInvalidFallback()
```
lukasbestle:
```suggestion
    public function testGetMultipleInvalidFallback()
```

src/Toolkit/Obj.php
```diff
@@ -64,12 +65,24 @@ public function __get(string $property)
     /**
      * Property Getter
```
lukasbestle:
```suggestion
     * Gets one or multiple properties of the object
```

src/Toolkit/Obj.php
```diff
@@ -64,12 +65,24 @@ public function __get(string $property)
...
      *
-     * @param string $property
+     * @param string|array $property
      * @param mixed $fallback
```
lukasbestle:
```suggestion
     * @param string|array $property
     * @param mixed $fallback If multiple properties are requested:
     *                        Associative array of fallback values per key
```

lukasbestle (commented):
The PHPUnit error is indeed weird...
Can't see anything that could be wrong.

A few minor comments, maybe the resulting rerun of the CI will fix the PHPUnit error as well. ;)
To avoid multiple CI runs, you can add all suggestions you agree with to the batch and commit them at once from the GitHub UI.

tests/Toolkit/ObjTest.php
adamkiss: @lukasbestle oh, right. I’m a dumbass. I totally saw “test__get” and thought it was just a weird notation for test get 🤦🏻‍♂️

src/Toolkit/Obj.php
```diff
@@ -62,14 +63,27 @@ public function __get(string $property)
...
+     * @param string|array $property
+     * @param mixed $fallback If multiple properties are requested:
+     *                        Associative array of fallback values per key
      * @return mixed
      */
-    public function get(string $property, $fallback = null)
+    public function get($property, $fallback = null)
     {
+        if (is_array($property)) {
+            if (!is_array($fallback)) {
```
lukasbestle, line 80: Ah, I now know why the CI doesn't pass. If `$fallback` is `null`, it's not an array. Let's try this:

```suggestion
            if ($fallback === null) {
                $fallback = [];
            }

            if (!is_array($fallback)) {
```

lukasbestle (approved):
LGTM 👍

Leaving the final review to @distantnative.

src/Toolkit/Obj.php
```diff
@@ -62,14 +63,31 @@ public function __get(string $property)
...
+                $fallback = [];
+            }
+
+            if (!is_array($fallback)) {
+                throw new InvalidArgumentException('The fallback value must be an array when getting multiple properties');
+            }
+
+            $result = [];
+            foreach ($property as $key) {
+                $result[$key] = $this->$key ?? $fallback[$key] ?? null;
```
distantnative, line 86: Might be a bit overkill but in general I like when those kind of functions that can be used for multiple call themselves to get the single when looping over the multiple.

```
$result[$key] = $this->get($key, $fallback[$key] ?? null)
```

afbora (approved):
Codes OK 👍 

Just a little thought out loud;
I have a feeling that we should do it in a new/separate method. When I saw this method, I immediately thought of the `A::pluck()` method.

adamkiss:
@afbora That is an option, but both `A::get()` and `Collection::get()` exist and while they are a bit different (collection methods vs single objects), you could also look at it as "collection of properties", and then it's in line with the other two.

Mainly - I've used `Obj` as the Base Class for my "sort of ORM" in two projects this last week, and `$object->get([keys])` fits right in with the other `A`, `$collection` and `$query` methods.

lukasbestle:
@afbora `A::pluck()` plucks a column from every element of a two-level array while `$obj->get()` returns multiple properties of the same single-level object. I agree with Adam that it fits more to `A::get()` than it does to `A::pluck()`.

src/Toolkit/Obj.php
lukasbestle: I agree with you, but for this one-liner I think it's really overkill as it makes the code harder to grasp.

adamkiss: Yeah, I'd prefer to not introduce recursion where it isn't strictly needed.

PR #4270, opened 2022-04-11
Feature: random in `Toolkit\A` and `Toolkit\Collection`

adamkiss:
### Features

New `A::random()` and `$collection->random()` methods to get one or multiple random items from arrays and collections, optionally shuffled. Unless shuffled, the overall order of the returned items is kept.

### Breaking changes

*None*

## Ready?

### For review team
@distantnative said he doesn't hate me yet

### DIFF
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -336,6 +336,29 @@ class A
     }
 
     /**
+     * Returns a number of random elements from an array,
+     * either in original or shuffled order
+     *
+     * @param array $array
+     * @param integer $count
+     * @param boolean $shuffle
+     * @return array
+     */
+    public static function random(array $array, int $count = 1, bool $shuffle = false): array
+    {
+        if ($shuffle) {
+            return array_slice(self::shuffle($array), 0, $count);
+        }
+
+        if ($count === 1) {
+            $key = array_rand($array);
+            return [$key => $array[$key]];
+        }
+
+        return self::get($array, array_rand($array, $count));
+    }
+
+    /**
      * Fills an array up with additional elements to certain amount.
      *
      * <code>
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -340,8 +340,8 @@ class A
      * either in original or shuffled order
      *
      * @param array $array
-     * @param integer $count
-     * @param boolean $shuffle
+     * @param int $count
+     * @param bool $shuffle
      * @return array
      */
     public static function random(array $array, int $count = 1, bool $shuffle = false): array
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -876,7 +876,7 @@ class Collection extends Iterator implements Countable
      *
      * @param int $count
      * @param bool $shuffle
-     * @return $this|static
+     * @return static
      */
     public function random(int $count = 1, bool $shuffle = false)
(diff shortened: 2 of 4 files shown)


src/Toolkit/A.php
```diff
@@ -335,6 +335,29 @@ public static function last(array $array)
         return array_pop($array);
     }
 
+    /**
+     * Returns a number of random elements from an array,
+     * either in original or shuffled order
+     *
+     * @param array $array
+     * @param integer $count
+     * @param boolean $shuffle
```
lukasbestle:
```suggestion
     * @param int $count
     * @param bool $shuffle
```

src/Toolkit/Collection.php
```diff
@@ -870,6 +870,25 @@ public function query(array $arguments = [])
         return $result;
     }
 
+    /**
+     * Returns a new collection consisting of random elements,
+     * from the original collection, shuffled or ordered
+     *
+     * @param int $count
+     * @param bool $shuffle
+     * @return $this|static
```
lukasbestle: Can this ever be `$this`? If I read it correctly, the object is cloned no matter what. In this case:

```suggestion
     * @return static
```

tests/Toolkit/ATest.php
```diff
@@ -265,6 +265,34 @@ public function testLast()
...
+        $this->assertTrue(in_array(array_values($random1)[0], $array));
+        $this->assertTrue(in_array(array_key_first($random1), array_keys($array)));
+
+        // Assert order of keys in non-shuffled random
+        $random2 = A::random($array, 2);
+        $this->assertTrue($arrayKeys[array_key_first($random2)] < $arrayKeys[array_key_last($random2)]);
+
+        // Assert count in completely shuffled array
+        $random3 = A::random($array, 3, true);
+        $this->assertSame(count($array), count($random3));
```
lukasbestle: Shouldn't this always be `3`?

```suggestion
        $this->assertCount(3, $random3);
```

lukasbestle (commented):
💛

src/Toolkit/A.php
adamkiss: This must have been copilot

src/Toolkit/Collection.php
adamkiss: Tbh I copied this from a random other Method 😅

tests/Toolkit/ATest.php
adamkiss: True, but I wanted highlight the fact that the count doesn't change, so when you look at the test a year later, you don't expect t he count to be three, but rather the same as original value.

adamkiss:
@lukasbestle this was quite light on the comments, Lukas 😄 I'll look at the changes later.

tests/Toolkit/ATest.php
lukasbestle: But if the input contains 4 items, won't the test fail then? I mean after all the expected behavior is to get the number of results that were requested, right?

adamkiss:
@lukasbestle you were right, updated. Are we still in the window for 3.6.5?

lukasbestle (approved):
👍

3.6.5 is already nearing release, but we can include this PR in 3.6.6, which is planned to be released in two weeks.

adamkiss:
🤝

Issue #4249, opened 2022-04-01
Thumbs: Consider throwing an error instead of returning the source image when creating the thumb failed


lukasbestle:
See https://github.com/getkirby/kirby/issues/3023#issuecomment-1018032814 and https://github.com/getkirby/kirby/commit/0db632aa76484cb263cab8a9ed545a304d24048a#r33938782.

I think it's a good point. The original image is not only often much too large in file size, it will also not have modifications such as grayscale and cropping. Also the site owner may not want to leak the original image file to the public. An `HTTP 500` error will show clearly that something went wrong with generation of this thumb.

What do you think @getkirby/kirby-staff? What was the reason why we return the source image and is it still relevant today?

iskrisis:
I think it's good idea. I've myself published websites that had thumb issues without me realizing it because i haven't noticed.

If it was setting i would be setting this on every website but setting it as default might bring upset people that relied on this without knowing :D.

lukasbestle:
I wouldn't make this an option to be honest. Options always make Kirby more complex. If we can't think of any reason why someone might still want the current behavior, we might as well get rid of it completely.

iskrisis:
Agree I wouldn't make this an option either. I am just thinking about backwards compatibility? This had to be deliberate decision for some reason. Maybe i just dont realize how it works. Kirby can't know if thumb creation failed right? It only knows the thumb is not there - but that also happens when the thumb haven't finished generation. Maybe this is better solution for first visits than flashing image errors. 🤷

lukasbestle:
The thumb generation is blocking, so Kirby knows the thumb generation is either done or it failed.

Regarding the original motivation: That's the big question. Waiting for insights from the team, especially @bastianallgeier.

andreasotto:
Good point!

Please link to #4145 i ran into trouble recently. Same topic imho.

I guess right now errors in thumbnail generation are quietly skipped?!
If there was error logging, we would have already gained a lot.

iskrisis:
@lukasbestle i think the drivers throw the error correctly https://github.com/getkirby/kirby/blob/6b20fa11843f57cd9a1e611bc9e8e8a91b855156/src/Image/Darkroom/ImageMagick.php#L164 but if i understand correctly PHP error bubbling the erorr is swallowed somewhere here https://github.com/getkirby/kirby/blob/6b20fa11843f57cd9a1e611bc9e8e8a91b855156/src/Cms/Media.php#L131 and the source image is returned instead.

[...]

PR #4275, opened 2022-04-16
Add frame option to ImageMagick Thumb Driver

tobimori:
## This PR …
Adds a frame option to the ImageMagick Thumb/Darkroom Driver, that allows one to select a single frame from an animated or multi-layered image. This should be useful when e.g. generating inline thumbs that shouldn't be animated, with e.g. [kirby-blurry-placeholder](https://github.com/johannschopplich/kirby-blurry-placeholder) - [see also](https://github.com/johannschopplich/kirby-blurry-placeholder/issues/18).

## Ready?

### For review team


tobimori:
Unsure about how detailed the unit test should be, but selecting the first frame should also work on single-layer/single-frame images - so we could extend the current `testProcess` test with the frame option, or add a new one that specifically transforms a GIF.

lukasbestle (commented):
I think it should be a separate unit test with a GIF.

Question: What happens if a subsequent frame is selected for a static image? I.e. if I tried to get the third frame of a JPG? Would be great to have a fallback for that unless ImageMagick already handles this case internally.

distantnative:
@tobimori would you be able to add unit tests for this? honest answer, no grudges if not. only then someone from us would need to pick it up at some point.

tobimori:
I'm moving atm and don't have capacities for that, so I'd appreciate your help :)

lukasbestle:
@tobimori Sorry that we kept this dormant so long. Are you still interested in this feature? Then we can try to finish it together.

tobimori:
I would be, and think it's a general issue that needs to be tackled at some point (especially for use cases like the placeholder plugin), but it's not a priority for me, as the client project I needed this for is done, and I disabled placeholders for all (or they stayed huge, dunno, wasn't a big deal in the end)


If you need any help in finishing this, I'll gladly help, although it's christmas season, and I'm on holidays from the end of next week until new years.

distantnative:
I have to admit I gave unit tests a try once but failed and abandoned it quickly 😅🙈

lukasbestle:
Thanks for your quick reply. I think before we can tackle unit tests, we need to answer this open question:

> What happens if a subsequent frame is selected for a static image? I.e. if I tried to get the third frame of a JPG? Would be great to have a fallback for that unless ImageMagick already handles this case internally.

tobimori:
> Thanks for your quick reply. I think before we can tackle unit tests, we need to answer this open question:
> 
> > What happens if a subsequent frame is selected for a static image? I.e. if I tried to get the third frame of a JPG? Would be great to have a fallback for that unless ImageMagick already handles this case internally.

Still returns the first frame:
CleanShot 2022-12-10 at 22 42 10@2x
-> both same output

lukasbestle:
@tobimori I've rebased the PR against our `develop` branch and made minor updates to the code. I think the change itself is good, we "just" need a good unit test for it that actually tests the behavior. Could you prepare a test case? We can help to integrate it.

tobimori:
CleanShot 2023-01-02 at 13 24 17@2x

Just stumbled upon this while working on a test case. It only happened with a file that has a limited number of frames, not e.g. a .png file that only has one.

The result of this command is two images with the frame attached as suffix.
CleanShot 2023-01-02 at 13 25 45@2x

It's the same as when converting a .gif to a single-level format like .png anyway, this shouldn't be a problem, right?

Edit:
Throws error if out of bounds.
CleanShot 2023-01-02 at 13 46 05@2x

tobimori:
I've added a unit test. It checks a sample GIF with a reference file generated by the same snippet.

tobimori:
Summarizing my previous comments:

Currrent behavior when transforming a .gif with multiple frames to a .png *without selecting a frame*:
CleanShot 2023-01-02 at 14 11 40@2x
Multiple files generated with frame index as suffix.

When transforming a .gif with multiple frames to a .png with *an not existing frame*:
CleanShot 2023-01-02 at 14 13 11@2x
Generates the same files as above, but shows that the command execution failed.

When transforming a .gif with multiple frames to a .png with *an existing frame*:
A .png with the same file name as the .gif is generated as expected with the selected frame.

When transforming a .jpeg/.png without multiple frames to an image with *an not existing frame* selected:
The first frame will always be output.

lukasbestle:
Thanks for adding the tests and for checking the edge cases. I think it would be good to have those covered as well, in the order of importance:

- Is it possible to stop ImageMagick from writing the frames in the "non-existing frame is selected" case? It's good that the error is generated, but an error case shouldn't generate files.
- I think it would be good to have special handling for the case when an animated image is converted to a non-animated image (like GIF to PNG) but no `frame` is passed. This should either trigger an error or use the first frame (whichever is easiest to implement; the main problem probably being that it's hard to detect whether an image is animated). AFAICT, the current behavior leads to broken thumbs anyway because of the automatically inserted suffix that won't be detected by Kirby. So this change would fix a bug and make the behavior consistent.
  - Possible solution from the top of my head: The default for the `frame` prop could be `0` when the target `format` does not have frame support. This would implement the "use the first frame" idea but still with compatibility for non-animated source images. However it's again difficult to detect whether the target format has animation support. We could hard-code `['jpg', 'png']` (with APNG, WebP and AVIF not listed as they support frames), but maybe there's a better option.

Could you please add unit tests for the edge cases as well?

tobimori:
I've looked through the ImageMagick docs and haven't found anything that stops IM from creating files in an error case. :/

tobimori:
I've implemented a new method that checks the frame count using the ImageMagick `identify` CLI. This is unfortunately the only method I found that works, [besides using the `Imagick` extension](https://www.php.net/manual/en/imagick.getnumberimages.php). This requires a new config option, `identifyBin`, which works similar to the current `bin` option when using the ImageMagick driver.

Each image generation now checks using such method whether a frame is in bounds or not, and automatically fallbacks if not converting to a supported format (Currently `avif`, `webp`, `gif` listed as supported). 

Unit tests are added covering the following cases:
- Exception gets thrown when a frame out of bounds is requested
- Output image by fallback is the same as when requesting frame 0 manually
- Frame count method works as expected, returning `2` for a sample GIF and `1` for a PNG.

I think this is a fair trade-off, given this fixes a significant bug that happens when trying to transform a GIF to any other unsupported format, if included in a major release of course. **It'd also open the possibility to add APNG support in the future, as such must be specified with the `APNG:` prefix manually, which we could probably detect using `identify`.**

tobimori:
Any updates?

lukasbestle:
Thanks for your research and sorry for the delay. We haven't yet found the time to look into this again. Especially because the edge cases make it all a lot more complex. To the point where I feel like the current implementation could be a downgrade for 99% of users who don't need to access frames.

tobimori:
> To the point where I feel like the current implementation could be a downgrade for 99% of users who don't need to access frames.

The way I look at this PR now is not an enhancement or feature, **but rather a critical bug fix for everyone working with GIFs as those currently do not get generated/converted properly.** If there is an alternative driver, such as the ImageMagick one, which most important difference to GD is to provide support for animated files, it should support them properly IMHO. (AVIF support is added to SimpleImage as we speak as it's in GDlib by now, so it'll probably be in the next Kirby release too)

If you are completely sure that this is not interesting to you, I'd also be up for moving this in a plugin. Although I heavily think this should be in core.

tobimori:
I moved this into a plugin: https://github.com/tobimori/kirby-magick-extended

lukasbestle:
I'm not sure yet. As I said, we haven't found the time to look at this again in detail due to other projects.

src/Image/Darkroom/ImageMagick.php
```diff
@@ -72,8 +72,47 @@ protected function convert(string $file, array $options): string
 			}
 		}
 
+		// frame option to allow selecting layers for multi-layer or frames for animated images
+		$fileOptions = '';
+		$frame = $options['frame'];
+		$maxFrames = $this->frameCount($file, $options);
```
distantnative, line 78: I think this should happen inside the if clause so we don't run the extra command if no `frame` option has been set

src/Image/Darkroom/ImageMagick.php
```diff
@@ -72,8 +72,47 @@ protected function convert(string $file, array $options): string
...
+		if ($frame !== null) {
+			// check whether frame is in bounds
+			if ($frame < 0) {
+				throw new Exception('Frame option must be a positive integer');
+			}
+
+			if ($frame !== 0 && $frame >= $maxFrames) {
+				throw new Exception('Frame option must be smaller than the number of frames in the image');
+			}
```
distantnative, line 87: Contemplating whether it would be smoother to just clop the bounds to 0/max instead of throwing exceptions.

src/Image/Darkroom/ImageMagick.php
distantnative: Ok I realised we need it for the `elseif` too. But then the name is misleading. This isn't max frames, but just the frame count. So maybe `frames` next to `frame` would be the better name

src/Image/Darkroom/ImageMagick.php
```diff
@@ -72,8 +72,47 @@ protected function convert(string $file, array $options): string
...
+
+			$fileOptions = "[{$frame}]";
+		} elseif ($maxFrames > 1) {
+			// if frame is not set and target format doesn't support multi-layer images, select first frame
+			$targetFormat = $options['format'] ?? F::extension($file);
+			$multiLayerFormats = ['gif', 'avif', 'webp'];
+			if (!in_array($targetFormat, $multiLayerFormats)) {
+				$fileOptions = '[0]';
+			}
```
distantnative, line 96: I think for better understanding of the code this is twisted: We should first on top check if target format isn't multi frame (why layers?) format and for default set it to first frame, unless a specific frame is defined.

src/Image/Darkroom/ImageMagick.php
```diff
@@ -72,8 +72,47 @@ protected function convert(string $file, array $options): string
...
+	/**
+	 * Returns the number of frames in an image
+	 */
+	public function frameCount(string $file, array $options): int
+	{
+		exec($options['identifyBin'] . ' -ping -format "%n\n" ' . escapeshellarg($file), $output, $return);
+
+		// log broken commands
+		if ($return !== 0) {
+			throw new Exception('The ImageMagick frame identification command could not be executed: ' . $file);
```
distantnative, line 112: When can this happen? Are we creating additional server requirements here?

tobimori: Yes. 

While ImageMagick 7+ includes both commands by default as 'identify' and 'convert', they might not be available on the server under the default name, in which case the error would throw.

More: https://imagemagick.org/script/identify.php

src/Image/Darkroom/ImageMagick.php
tobimori: I decided to throw an exception here because I remembered #4249 and thought it was closed already.

src/Image/Darkroom/ImageMagick.php
distantnative: But wondering how widespread that is already (I am not much familiar with it, sorry) - just concerned that suddenly many of our IM users would run into this exception

tobimori: It should be available everywhere where `convert` is installed, as it comes with the default bundle of ImageMagick.

src/Image/Darkroom/ImageMagick.php
distantnative: But with throwing an exception we would even put more fire to #4249 because then Kirby would simply deliver the source image. I feel delivering the first/last frame when requesting a frame out of bounds might still be better than delivering the source image?

tobimori:
Revisiting this after a few months, I think:

- The default behavior should be limiting to frame 0, when the target format does not support frames (e.g. not `avif`, `webp`, `gif`, etc.), otherwise do not pass frame options.
- This would solely fix the bugs described when e.g. converting a GIF to a JPEG and the current unexpected behavior where all frames will be converted separately as JPEG with suffixes. 
- The frame counting behavior and its related options:
  a. should stay in a plugin, and not become part of Kirby core
  b. should always fall back to the first frame, if `identifyBin` is unavailable
  c. should be togglable by either having `identifyBin` undefined by default, or having a separate option to toggle the behavior
- If b or c, this would also allow for APNG support in core.

@distantnative

src/Image/Darkroom/ImageMagick.php
tobimori: I named it layers, probably because different formats handle frames differently. Some are animated (like GIFs), some have them as separate layers for e.g. graphics editing (e.g. TIFFs). Although you're right that in web behavior it's probably only 'multi-frame'.

distantnative:
@tobimori 
Yep, agreed on the first two.
The third, I'd go for (c) - having it undefined by default already adds the option for users to define it and activate those functionalities. Could also be a plugin, but it doesn't seem like much of a burden to have that in the core as well then.

tobimori:
Perfect. I'll get back to this PR in the next few weeks.

tobimori:
Still planned, will wait for v4.

distantnative:
@tobimori if you want, we'd be up to integrate this during the alpha phase already. That way we can also test better if anything breaks.

tobimori:
I don't have the time right now, sorry.

src/Image/Darkroom/ImageMagick.php
```diff
@@ -72,8 +72,47 @@ protected function convert(string $file, array $options): string
...
-		return $command . ' ' . escapeshellarg($file);
+		return $command . ' ' . escapeshellarg($file . $fileOptions);
+	}
+
+	/**
+	 * Returns the number of frames in an image
+	 */
+	public function frameCount(string $file, array $options): int
+	{
+		exec($options['identifyBin'] . ' -ping -format "%n\n" ' . escapeshellarg($file), $output, $return);
```
distantnative, line 108: What's the performance implication of running this command for every thumb additionally? Do we know?

src/Image/Darkroom/ImageMagick.php
distantnative: When trying to fix the conflicts, I ran into this breaking out unit test setup. I am wondering if it would be better to fall back to use the first frame here instead of throwing an exception.

tobimori:
I'm going to close this PR. I haven't had this case since the creation again and would rather be interested in integrating an alternative to ImageMagick (e.g. libvips) in the long term instead. Of course, you can pick up this PR and refactor it, but I don't see any reason to continue working on it.

PR #4277, opened 2022-04-18
Fine-tuned cache handling based on request data

### LINKED ISSUE #3976: Disable cache on active sessions only

### LINKED ISSUE #4299: [3.6.5] Panel constantly reloads and logins frequently fail (unauthenticated / invalid CSRF token)
## Description

After updating to Kirby 3.6.5, we were seeing strange Panel behavior. If you were already logged in and looking at a page, the Panel would just randomly reload the Panel page over and over. Some sections would show errors like "Unauthenticated". Some users were getting popups like "Attempt to use history.pushState() more than 100 times per 30 seconds."

Further, when trying to log in we would sometimes get errors like "Invalid CSRF Token".

Even after rolling back to Kirby 3.6.4 were still getting some wonky behavior, mostly about the login page complaining that session tokens were read-only.

What's somewhat peculiar is that I don't get this erratic behavior on our Dev server or my local setup. So, it's possible that this issue is specific to my Production setup, but I'm unsure where to even look next.

**Expected behavior**  

**Screenshots**  

https://user-images.githubusercontent.com/417182/166346461-43513a59-67c2-4f68-a844-35c886d5148a.mov
2022-05-02 at 1 22 PM

## Your setup
**Kirby Version**  
`3.6.5`
PHP 8.1.4 and 8.1.5

**Console output**  
None

 - Device: Mac
 - OS: macOS
 - Browser: Safari 15.4; Microsoft Edge 101.0.1210.32

**Additional context**  
We use file-based sessions. I tried clearing out our "sessions" folder to see if that would help anything.

After reverting back to Kirby 3.6.4 _some_ Panel calls would still error out and include the following:
```
<b>Fatal error</b>:  Uncaught Kirby\Exception\LogicException: Cannot write to session &quot;1651543253.6a7dee26d03f46c0a8a9&quot;, because it is not locked in /var/www/html/kirby/src/Session/FileSessionStore.php:263
Stack trace:
#0 /var/www/html/kirby/src/Session/Session.php(373): Kirby\Session\FileSessionStore-&gt;set(1651543253, '6a7dee26d03f46c...', '4d7378aed8a740b...')
#1 [internal function]: Kirby\Session\Session-&gt;commit()
#2 {main}
  thrown in <b>/var/www/html/kirby/src/Session/FileSessionStore.php</b> on line <b>263</b><br />
<br />
<b>Fatal error</b>:  Uncaught Kirby\Exception\LogicException: Cannot write to session &quot;1651543253.6a7dee26d03f46c0a8a9&quot;, because it is not locked in /var/www/html/kirby/src/Session/FileSessionStore.php:263
Stack trace:
#0 /var/www/html/kirby/src/Session/Session.php(373): Kirby\Session\FileSessionStore-&gt;set(1651543253, '6a7dee26d03f46c...', '4d7378aed8a740b...')
#1 /var/www/html/kirby/src/Session/Session.php(493): Kirby\Session\Session-&gt;commit()
#2 [internal function]: Kirby\Session\Session-&gt;__destruct()
#3 {main}
  thrown in <b>/var/www/html/kirby/src/Session/FileSessionStore.php</b> on line <b>263</b><br />
```

lukasbestle:
### Features

- New `$kirby->response()->usesAuth()` and `->usesCookie()` method to tell Kirby's automatic caching system about used `Authorization` headers and cookies. These methods are called automatically when the Kirby `Request` and `Cookie` classes are used. #4277
- New `$kirby->request()->hasAuth()` method to check the existence of the `Authorization` header #4277

### Enhancements

- Kirby's response handling now automatically controls page and HTTP caching based on cookies, session and the `Authorization` request header being used by the response. If any of these are used and also contained in the request, the response is considered private. If they are used but not contained in the request, the page is only cached for visitors without these cookies/session/auth. Manual sessions are always considered private. #3976

### Fixes

- Fixed a regression that prevented page and HTTP caching when the session was accessed, even when the session was not active #3976
- The Panel and API now send the `Cache-Control: no-store, private` response header instead of just `Cache-Control: no-store` for better compatibility with caches like CloudFlare #4299

### Breaking changes

*None*

## Explanation

This PR implements the behavior I wrote down in https://github.com/getkirby/kirby/issues/3976#issuecomment-1094297863.

I have done manual tests in the frontend and Panel to confirm the behavior and also added detailed unit tests to avoid regressions in the future.

**But the big question is if the behavior is correct at all.** So I'm happy for your help with real-world testing in actual projects. Does this PR behave like you expect? A few notes and recommendations for testing:

- Make sure that the page cache is enabled. Also ensure the cache is cleared when you start testing.
- Relevant are four cases:
  - Pages that *don't* use the session or cookies.
  - Pages that use the session and/or cookies in various ways with the visitor having the cookie(s) already set when first visiting the page with empty page cache.
  - The same, but after the page was already cached by another visitor that didn't have the cookie(s) set.
  - The same, but with repeated requests by a visitor that doesn't have the cookie(s) set.
- Please note that responses that set cookies or store data in the session are treated like a visitor that already requests with those cookies or session.
- The changes have both an effect on response headers for caching (`Vary` and `Cache-Control`) and also on the page cache (prevention of caching in the first place or prevention of cache retrieval).

Pinging @sebastiangreger @bvdputte @gearsdigital @adamkiss @christophknoth as you were involved in the original issue discussion.

If everything is fine, we plan to release this change in Kirby 3.6.6 (RC on Tuesday 26th).

## Ready?

### For review team

### DIFF
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -1029,11 +1029,19 @@ class Page extends ModelWithContent
 
         // try to get the page from cache
         if (empty($data) === true && $this->isCacheable() === true) {
-            $cache    = $kirby->cache('pages');
-            $cacheId  = $this->cacheId($contentType);
-            $result   = $cache->get($cacheId);
-            $html     = $result['html'] ?? null;
-            $response = $result['response'] ?? [];
+            $cache       = $kirby->cache('pages');
+            $cacheId     = $this->cacheId($contentType);
+            $result      = $cache->get($cacheId);
+            $html        = $result['html'] ?? null;
+            $response    = $result['response'] ?? [];
+            $usesAuth    = $result['usesAuth'] ?? false;
+            $usesCookies = $result['usesCookies'] ?? [];
+
+            // if the request contains dynamic data that the cached response
+            // relied on, don't use the cache to allow dynamic code to run
+            if (Responder::isPrivate($usesAuth, $usesCookies) === true) {
+                $html = null;
+            }
 
             // reconstruct the response configuration
             if (empty($html) === false && empty($response) === false) {
@@ -1060,15 +1068,15 @@ class Page extends ModelWithContent
             // render the page
             $html = $template->render($kirby->data);
 
-            // convert the response configuration to an array
-            $response = $kirby->response()->toArray();
-
             // cache the result
-            if ($cache !== null && $kirby->response()->cache() === true) {
+            $response = $kirby->response();
+            if ($cache !== null && $response->cache() === true) {
                 $cache->set($cacheId, [
-                    'html'     => $html,
-                    'response' => $response
-                ], $kirby->response()->expires() ?? 0);
+                    'html'        => $html,
+                    'response'    => $response->toArray(),
+                    'usesAuth'    => $response->usesAuth(),
+                    'usesCookies' => $response->usesCookies(),
+                ], $response->expires() ?? 0);
             }
         }
 
--- a/src/Cms/Responder.php
+++ b/src/Cms/Responder.php
@@ -43,7 +43,7 @@ class Responder
      * Flag that defines whether the current
      * response can be cached by Kirby's cache
      *
-     * @var string
+     * @var bool
      */
     protected $cache = true;
 
@@ -60,6 +60,23 @@ class Responder
      * @var string
      */
     protected $type = null;
+
+    /**
+     * Flag that defines whether the current
+     * response uses the HTTP `Authorization`
+     * request header
+     *
+     * @var bool
+     */
+    protected $usesAuth = false;
+
+    /**
+     * List of cookie names the response
+     * relies on
+     *
+     * @var array
+     */
+    protected $usesCookies = [];
 
     /**
      * Creates and sends the response
@@ -99,10 +116,68 @@ class Responder
     public function cache(?bool $cache = null)
     {
         if ($cache === null) {
+            // never ever cache private responses
+            if (static::isPrivate($this->usesAuth(), $this->usesCookies()) === true) {
+                return false;
+            }
+
             return $this->cache;
         }
 
         $this->cache = $cache;
+        return $this;
+    }
+
+    /**
+     * Setter and getter for the flag that defines
+     * whether the current response uses the HTTP
+     * `Authorization` request header
+     * @since 3.6.6
+     *
+     * @param bool|null $usesAuth
+     * @return bool|$this
+     */
+    public function usesAuth(?bool $usesAuth = null)
+    {
+        if ($usesAuth === null) {
+            return $this->usesAuth;
+        }
+
+        $this->usesAuth = $usesAuth;
+        return $this;
+    }
+
+    /**
+     * Setter for a cookie name that is
+     * used by the response
+     * @since 3.6.6
+     *
+     * @param string $name
+     * @return void
+     */
+    public function usesCookie(string $name): void
+    {
+        // only add unique names
+        if (in_array($name, $this->usesCookies) === false) {
+            $this->usesCookies[] = $name;
+        }
+    }
+
+    /**
+     * Setter and getter for the list of cookie
+     * names the response relies on
+     * @since 3.6.6
+     *
+     * @param array|null $usesCookies
+     * @return array|$this
+     */
+    public function usesCookies(?array $usesCookies = null)
+    {
+        if ($usesCookies === null) {
+            return $this->usesCookies;
+        }
+
+        $this->usesCookies = $usesCookies;
         return $this;
     }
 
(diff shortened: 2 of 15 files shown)


sebastiangreger:
@lukasbestle Assuming you are specifically looking for "real-life" cross-checks, I ran your outlined experiments on my rather complex website configuration and could not identify any issues.

While my testing was unfortunately not 100% systematic (it's just too messy a setup to qualify as a "clean" test), the mechanism behaved as described. In other words: this fixes the issues I first encountered, leading to #3976, and I did not run into any obvious problems regarding the various "risks" discussed earlier.

This is an impressively wholesome solution, and a great improvement!

lukasbestle:
@sebastiangreger Thank you very much for testing and your feedback. 👍

lukasbestle:
We have decided to move this change to 3.7. This allows further testing and hopefully avoids regressions.

src/Cms/Page.php
```diff
@@ -1029,11 +1029,19 @@ public function render(array $data = [], $contentType = 'html'): string
...
+            $cacheId     = $this->cacheId($contentType);
+            $result      = $cache->get($cacheId);
+            $html        = $result['html'] ?? null;
+            $response    = $result['response'] ?? [];
+            $usesAuth    = $result['usesAuth'] ?? false;
+            $usesCookies = $result['usesCookies'] ?? [];
+
+            // if the request contains dynamic data that the cached response
+            // relied on, don't use the cache to allow dynamic code to run
+            if (Responder::isPrivate($usesAuth, $usesCookies) === true) {
```
distantnative, line 1042: Maybe I am reading this wrong, but to me this looks like we would base the decision whether the current request uses auth or cookies on the cached values, not the current one.

src/Cms/Responder.php
```diff
@@ -99,13 +116,71 @@ public function body(string $body = null)
     public function cache(?bool $cache = null)
     {
         if ($cache === null) {
+            // never ever cache private responses
```
distantnative, line 119: technically, inside this if statement, this would rather be "never ever return cache for private responses". question is if the new if statement shouldn't be at the top of the method to prevent getting as well as setting cache

src/Cms/Responder.php
```diff
@@ -99,13 +116,71 @@ public function body(string $body = null)
...
 
         $this->cache = $cache;
         return $this;
     }
 
+    /**
+     * Setter and getter for the flag that defines
+     * whether the current response uses the HTTP
+     * `Authorization` request header
+     * @since 3.6.6
```
distantnative: We need to update the since tags - not just here

src/Cms/Responder.php
```diff
@@ -310,4 +415,32 @@ public function type(string $type = null)
...
         return $this;
     }
+
+    /**
+     * Checks whether the response needs to be exempted from
+     * all caches due to using dynamic data based on auth
+     * and/or cookies; the request data only matters if it
+     * is actually used/relied on by the response
+     * @internal
+     *
```
distantnative, line 426: Should maybe get a since tag, even if internal?

src/Cms/Page.php
lukasbestle: The `$usesAuth` and `$usesCookies` values come from the cache. But they only store which cookies were accessed by the previously rendered page, no matter if these cookies were already set back then. The current request is checked by `Responder::isPrivate()`. If there is an intersection between the request data and the used data, the cache is ignored.

src/Cms/Responder.php
lukasbestle: `Response::$cache` is the flag that controls whether the current response can be cached, not whether a cached response can be returned.

This `if` only overrides the manually set value if `isPrivate()` is true. Otherwise the manual value "shines through". So I don't think we should limit the setter.

distantnative (approved):

PR #4281, opened 2022-04-20
Fix and support callback option to programmable page and user blueprints

afbora:
## This PR …

Added missing callback support for loading user roles via plugins.

### Enhancements

- Support returning blueprint file path in callback for programmable blueprints

### Fixes

- Fixed callback option for user blueprints load

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Roles.php
+++ b/src/Cms/Roles.php
@@ -106,12 +106,12 @@ class Roles extends Collection
                 continue;
             }
 
-            if (is_array($blueprint) === true) {
-                $role = Role::factory($blueprint, $inject);
-            } elseif (is_callable($blueprint) === true) {
+            // callback option can be return array or blueprint file path
+            if (is_callable($blueprint) === true) {
                 $blueprint = $blueprint($kirby);
-                $blueprint['name'] ??= $blueprintName;
+            }
 
+            if (is_array($blueprint) === true) {
                 $role = Role::factory($blueprint, $inject);
             } else {
                 $role = Role::load($blueprint, $inject);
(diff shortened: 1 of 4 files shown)


src/Cms/Roles.php
```diff
@@ -97,15 +97,21 @@ public static function factory(array $roles, array $inject = [])
...
+        foreach ($kirby->extensions('blueprints') as $blueprintName => $blueprint) {
             if (substr($blueprintName, 0, 6) !== 'users/') {
                 continue;
             }
 
             if (is_array($blueprint) === true) {
+                $role = Role::factory($blueprint, $inject);
+            } elseif (is_callable($blueprint) === true) {
+                $blueprint = $blueprint($kirby);
+                $blueprint['name'] ??= $blueprintName;
```
distantnative: I'd make this a separate if statement before the others. Keeps code a bit more DRY and would allow also returning a string form the callback

afbora: I did exactly as you said at the beginning. But that would be inconsistent as it would be different from page blueprint loads. That's why I did it this way. But if you want, I can develop the callback string to return on page blueprint loads as well.

lukasbestle: As far as I can tell, the only difference would be that the current behavior only allows arrays to be returned from the callback whereas a separate `if` statement before would also allow the `else` case with a string. Would anything speak against supporting that?

afbora: I have no contrary opinion. Yes, it is very easy to implement. I think it would be useful to return a file path, not just an array. If everyone agrees, I can make this development for both parts.

lukasbestle: I agree, if we do it here, we should also support it for page blueprints.

afbora:
Ready to review 🚀

bastianallgeier (approved):
LGTM

PR #4285, opened 2022-04-25
New Environment class

### LINKED ISSUE #4234: Url overwrite config doesn't overwrite protocol
## Description
After upgrading to 3.6(.3.1) I experience a behavior similar to the one described in https://github.com/getkirby/kirby/issues/3645, when I try to access the panel, only for me it’s not the port, but the entire domain:

```
DOMException: Failed to execute 'pushState' on 'History': A history state object with URL 'http://localhost/panel/site' cannot be created in a document with origin 'https://ourdomain.com' and URL 'https://ourdomain.com/panel/site'.
```

The frontend of our website with Kirby works as expected. 

We run Kirby in a Docker setup with two containers: An nginx container that handles the incoming requests and forwards them to a php-fmt container, which runs Kirby. I assume the described behavior happens because within this setup the Kirby server is actually running on localhost (within its container), as it gets its requests directly from the nginx container.

In the frontend we observed a similar behavior, where one would be forwarded to `localhost/de` when loading the page, but there we could fix it simply by setting `'url' => "ourdomain.com"` in `config.php`.

**Expected behavior**  
The user should be forwarded to `https://ourdomain.com/panel/site`. Like the frontend, the Panel should prefer the domain given in`config.php` rather than try to extract the domain from the server’s information (I assume `HTTP_HOST`?)

Edit: While `Server::get('HTTP_HOST')` returns the correct host, `Server::host()` returns `localhost`.

## Your setup
**Kirby Version**  
Everything above 3.6

**Console output**  
```
DOMException: Failed to execute 'pushState' on 'History': A history state object with URL 'http://localhost/panel/site' cannot be created in a document with origin 'https://ourdomain.com' and URL 'https://ourdomain.com/panel/site'.
```

### LINKED ISSUE #4311: kirby->url() adding port and wrong protocol on nginx / reverse proxy
## Description

Upgrade installation on nginx / reverse proxy environment from 3.5 to 3.6.6 breaks urls.

We're running a kirby installation in a nginx / reverse proxy environment (listening to port 85).
'url' => 'HTTPS://www.domain.com' ist hardcoded in custom config file (config.www.domain.com.php).

$kirby->url() and $site->url() return 'HTTP://www.domain.com:85' while $kirby->option('url') returns 'HTTPS://www.domain.com'

So it seems like, the url option is completely irgnored since the update.

bastianallgeier:
### Refactoring
- New object-oriented `Http\Environment` class that replaces the former static `Http\Server` class and takes over all its methods

  New `Environment` class takes care of detecting all necessary parts (host, port, https, etc.) and considers an optional allow list of URLs to validate the current environment:

  ```php
  $env = new Environment([
    'allowed' => [
      'https://dev.getkirby.com',
      'https://test.getkirby.com:9999/subfolder'
   ]
  ]);
  ```

  I.e. if it found to be running on the test.getkirby.com host from above, https is on and the port is 9999 and the subfolder also matches, you can get all those parts now correctly via. 

  ```php
  $env->https();
  $env->host();
  $env->subfolder();
  $env->port();
  ```

- Using the new `Environment` class for all parts of the Kirby app to construct absolute and relative URLs, roots and paths

### Fixes

To be confirmed
- [x] https://github.com/getkirby/kirby/issues/4311
- [x] https://github.com/getkirby/kirby/issues/4234

### Deprecated
- `Cms\App::server()` (aka `$kirby->server()`) will start throwing warnings in 3.8.0 and will be removed in 3.9.0. Use `Cms\App::environment()`  (aka `$kirby->environment()`) instead.
- `Http\Server::` constants throw warnings in 3.7.0 and will be removed in 3.8.0
- `Http\Server` class will be removed in 3.8.0

### Breaking changes

- The `Server::hosts()` method was removed. Please set the allowed URLs via the `url` option.

## Todo
- [x] Move `Cms\System::isLocal()` to `Environment` class?
- [x] `@since` tags

## Ready?

### For review team


lukasbestle:
With `Server` unchanged, does this currently have an effect on the URL generation of the CMS (if so, how)? Or is it just a preparation for the changes in 3.7?

bastianallgeier:
Server is untouched to avoid any breaking changes with Server:: method calls, but the base URL is coming from the Environment class and should be more reliable now. The new logic is as follows: 

A) Try to get a fixed URL from the `url` option. Take https, host, port and subfolder from there. 

… If not …

B) Try to get an array of allowed URLs from the `url` option and detect which one is the correct one by comparing all four parts (https, host, port and subfolder) with the info coming either from the reverse proxy headers or the server config. 

... If not …

C) Try to auto-detect the correct URL parts from the local server config

… unless … 

D) the `url` option is set to `'*'`. In this case it will get the URL parts from insecure forwarded server headers or  HTTP_HOST without extra checks. 

Phew, I hope it gets clear by this pseudo code.

distantnative:
@lukasbestle did you turn this into a draft just to move it towards 3.7? Can we activate it again?

I didn't see any disagreement in the discussion. So since this will now end up in 3.7 directly anyways, can we add the `Server` class changes and then merge it?

lukasbestle:
The code is currently not working. I think Bastian wanted to work on it.

bastianallgeier:
I need to pick it up, but to be honest it's one my "shit I still have to do this" task list. I'm super happy about the Environment class in general, but the failing unit tests gave me massive headaches. Whoever feels like digging into it together with me would be my hero.

distantnative:
I would love to help but I think I'm not really clear on what the plan is:
- Are we getting rid of the `Server` class completely in 3.7? Seems to me that all the methods are also represented in `Environment`.
- If yes, we need to change all mentions of `Server::` to `kirby()->environment()->`
- But what about the usage of her `Server` class in non-`Cms` packages, e.g. in `Http`? Can we rely on `App::instance(null, true)`?

lukasbestle:
I'd like to take another detailed look at this before we merge it. I know we can always update it later, but in this open PR we can attach review comments to the code changes more easily. Since it's a large change and intertwined, I'd say we better not rush it. I'll try to send my review over the weekend.

src/Cms/System.php
```diff
@@ -595,7 +558,7 @@ public function serverSoftware(): ?string
             ];
         }
 
-        $software = $_SERVER['SERVER_SOFTWARE'] ?? '';
+        $software = $this->app->environment()->get('SERVER_SOFTWARE', '');
```
lukasbestle, line 561: I like this change a lot. It makes the whole system a lot easier to test.

There are some other places where we still use `$_SERVER`. Don't know if it's worth the effort with all. Posting this for discussion:

**Easy to update with `$this->environment()`:**

https://github.com/getkirby/kirby/blob/bb7732569a6814bc8279b13c7dec7519b7378f2b/src/Cms/AppErrors.php#L132

**Could be updated with a lazy `App::instance(null, true)` with a fallback to `$_SERVER`:**

https://github.com/getkirby/kirby/blob/bb7732569a6814bc8279b13c7dec7519b7378f2b/src/Api/Api.php#L690
https://github.com/getkirby/kirby/blob/bb7732569a6814bc8279b13c7dec7519b7378f2b/src/Exception/Exception.php#L148-L149
https://github.com/getkirby/kirby/blob/bb7732569a6814bc8279b13c7dec7519b7378f2b/src/Http/Header.php#L133
https://github.com/getkirby/kirby/blob/bb7732569a6814bc8279b13c7dec7519b7378f2b/src/Http/Request.php#L229-L236
https://github.com/getkirby/kirby/blob/bb7732569a6814bc8279b13c7dec7519b7378f2b/src/Http/Request.php#L317
https://github.com/getkirby/kirby/blob/bb7732569a6814bc8279b13c7dec7519b7378f2b/src/Http/Url.php#L189
https://github.com/getkirby/kirby/blob/bb7732569a6814bc8279b13c7dec7519b7378f2b/src/Http/Visitor.php#L58-L61

src/Http/Uri.php
```diff
@@ -234,25 +235,15 @@ public function clone(array $props = [])
...
-        $url = new static(array_merge([
-            'scheme' => Server::https() === true ? 'https' : 'http',
-            'host'   => Server::host(),
-            'port'   => Server::port(),
-            'path'   => $uri['path'],
-            'query'  => $uri['query'],
-        ], $props));
-
-        return $url;
+        return new static(App::instance()->url('current'), $props);
```
lukasbestle: I like the change for the CMS context. Much simpler and more consistent.

For use cases that don't involve the CMS this would be a massive breaking change though. Can we do something like this:

```php
$app = App::instance(null, true);

if ($app) {
    $url = $app->url('current');
} else {
    $url = (new Environment())->url();
}

return new static($url, $props);
```

Same with the `index()` method below.

tests/Cms/Helpers/HelperFunctionsTest.php
```diff
@@ -895,9 +893,9 @@ public function testTwitter()
 
     public function testUrl()
     {
-        $app = $this->kirby->clone([
-            'urls' => [
-                'index' => $url = 'https://getkirby.com'
+        $this->kirby->clone([
+            'options' => [
+                'url' => $url = 'https://getkirby.com'
```
lukasbestle, line 898: Was this just changed from `urls.index` to `options.url` for cleaner code or wouldn't it work otherwise? If the latter, can we avoid this breaking change or is it necessary and we should document it?

src/Http/Environment.php
```diff
@@ -0,0 +1,904 @@
...
+
+        return true;
+    }
+
+    /**
+     * Loads and returns the environment options
+     *
+     * @return array
+     */
+    public function options(): array
```
lukasbestle: I think the option loading part should be handled in the `Cms` namespace, not in `Http`.

What about a `Cms\Environment` class that extends `Http\Environment` and adds the CMS-specific attribute (`$root`) and method (`options()`)?

src/Http/Environment.php
```diff
@@ -0,0 +1,904 @@
...
+
+        if (empty($host) === false) {
+            $parts = $this->detectPortInHost($host);
+
+            $this->portInHost = $parts['port'];
+
+            return $parts['host'];
+        }
+
+        return $this->detectHost(true);
```
lukasbestle: The `detectForwardedHost()` method is only called if `HTTP_X_FORWARDED_HOST` is not empty. Can this line ever be reached?

src/Http/Environment.php
```diff
@@ -0,0 +1,904 @@
...
+        if ($this->detectHttpsOn($this->get('HTTP_X_FORWARDED_SSL')) === true) {
+            return true;
+        }
+
+        if ($this->detectHttpsProtocol($this->get('HTTP_X_FORWARDED_PROTO')) === true) {
+            return true;
+        }
+
+        // fall back to local setup
+        return $this->detectHttps();
```
lukasbestle: Not sure about this. If this line is ever reached, the result will most likely be incorrect as the local HTTPS status doesn't tell anything about the actual protocol that the user sees on the outside. I feel like we should rather throw an Exception if this is reached. Or have I missed anything?

src/Http/Environment.php
```diff
@@ -0,0 +1,904 @@
...
+            return $port;
+        }
+
+        // based on forwarded host
+        if (is_int($this->portInHost) === true) {
+            return $this->portInHost;
+        }
+
+        // fall back to local setup
+        return $this->detectPort();
```
lukasbestle: Same as above

src/Http/Server.php
```diff
@@ -13,343 +13,26 @@
...
-     * $hosts = Server::HOST_FROM_SERVER -> []
-     * $hosts = Server::HOST_FROM_HEADER -> ['*']
-     * $hosts = array                    -> [array of trusted hosts]
-     * $hosts = string                   -> [single trusted host]
-     *
-     * @param string|array|int|null $hosts
-     * @return array
+     * @return \Kirby\Http\Environment
      */
-    public static function hosts($hosts = null): array
```
lukasbestle, line 141: This method was removed in the new implementation as it's no longer needed by the core. But could it be called by users or plugins? Do we need it for compatibility until the `Server` class gets removed in 3.8?

lukasbestle (commented):
Amazing how it all comes together. `Environment` as the new central hub is really great. 👍

Only downside in my view is that it now bundles quite a bit of logic. The separate detection methods help a lot with readability. It's still a bit hard to keep the overview, but that's unavoidable with just the size of the file.

A bit of feedback and a few questions:

src/Http/Environment.php
distantnative: We actually had this at one point - but I think Bastian merged it to reduce complexity. Keep in mind we once said we don't want to artificially uphold the separation too much where it doesn't make sense. Certainly for many Toolkit classes it makes sense, but not sure if anyone really is using the Http package outside of Kirby context.

src/Http/Environment.php
```diff
@@ -0,0 +1,989 @@
...
+            'root'    => null,
+            'allowed' => null
+        ], $options ?? []);
+
+        $this->info          = $this->sanitize($info);
+        $this->cli           = $this->detectCli($options['cli']);
+        $this->ip            = $this->detectIp();
+        $this->host          = null;
+        $this->https         = false;
+        $this->isBehindProxy = $this->detectIfIsBehindProxy();
```
lukasbestle: This line allows visitors/attackers to completely control the `isBehindProxy` flag by setting the forwarded headers, even on setups where the forwarded headers are not used in the end (disabled by the `allowed` config).

I think we should base this value on the fact whether the forwarded host is actually used for the environment detection. For example we could set it in `detectAuto()`.

src/Http/Environment.php
lukasbestle: You are right, just for this one method it doesn't make sense.

I've integrated the `$root` as an argument for the `options()` method. This way, non-CMS users can just ignore the `options()` method entirely.

src/Http/Environment.php
lukasbestle: I've committed a possible fix, please check if you agree with it.

src/Http/Environment.php
```diff
@@ -0,0 +1,981 @@
...
+
+    /**
+     * Class constructor
+     *
+     * @param array|null $options
+     * @param array|null $info Optional override for `$_SERVER`
+     */
+    public function __construct(?array $options = null, ?array $info = null)
+    {
+        $this->detect($options, $info);
```
lukasbestle, line 123: This line (direct detection from the `Environment` constructor) causes a bug in the error handling.

**How to reproduce:**

When you define the `url` option e.g. as:

```php
<?php

return [
    'debug' => true,
    'url'   => ['https://getkirby.com', 'https://trykirby.com']
];
```

and visit from a different domain, the `Environment` class throws the "The environment is not allowed" exception. But the user only sees the fatal view, even if `debug` is enabled.

**Reason for the bug:**

This is because when `App::optionsFromEnvironment()` creates the `Environment` instance, the detection runs before the instance is assigned to `$app->environment`. So by the time the exception is finally thrown, `$app->environment` is still `null`. This means that `AppErrors::handleErrors()` cannot access the CLI status from the environment, which throws another PHP error. So the second fatal error triggers the fatal view.

**Possible fix:**

I think we can only fix this by removing the constructor and only calling `detect()` manually from `App::optionsFromEnvironment()`. On the other hand this has the downside that the `Environment` class can only be used in this way. Calling `detect()` can easily be forgotten.

src/Cms/System.php
bastianallgeier: I started and then thought, it would be better to stop and maybe extract that into another PR. This one is already so big.

src/Http/Uri.php
bastianallgeier: You are right about the breaking change. In general, I feel like we should move away from any form of isolated support though. We complicate so much of our own code by trying to keep parts of it as some sort of independent framework, which we never really use or test this way though.

tests/Cms/Helpers/HelperFunctionsTest.php
bastianallgeier: Just for cleaner testing code. I thought it would be easier to understand. A custom URL setup feels more advanced than just setting the main URL.

src/Http/Environment.php
bastianallgeier: TBH I'm not experienced enough with proxy setups. If the local setup really does not provide any useful reference in such a case, we can really skip it. Wouldn't it be fine to just assume it's not HTTPS in that case instead of throwing an exception?

src/Http/Server.php
bastianallgeier: We introduced it in 3.6.x somewhere. It's super young. I think it would be fine to remove it already again as a breaking change. It's very unlikely that someone already started using it within the last few months. It was hardly documented at all.

src/Http/Environment.php
bastianallgeier: Couldn't we fix it with a similar approach like in the URI class? If the environment is already detected, we use that instance. If it's still null, we use a new "detached" environment instance.

src/Http/Environment.php
bastianallgeier: This makes me wonder if we should introduce a singleton mode for Environment, similar to App::instance.

src/Http/Server.php
distantnative: Agreed, it's a very young internal method, I think this will do no harm removing it now as a breaking change.

src/Cms/System.php
lukasbestle: OK, let's do that once this PR is merged.

src/Http/Uri.php
lukasbestle: I agree. We don't need to explicitly support such use cases, but I think we shouldn't throw rocks in the path if we don't need to. We use Kirby as a framework for the Demo Manager for example. Haven't tested it yet with this PR, but I believe it would break as well.

src/Http/Environment.php
lukasbestle: I think we cannot assume anything really. If the proxy sets the `X-Forwarded-Host` header but none of the headers we can use for port and HTTPS detection, this would IMO either be a misconfiguration in the proxy or a bug in our detection code. So I think we should make it clear that something is wrong.

src/Http/Server.php
lukasbestle: I've added it to the release note list in the first post.

src/Http/Environment.php
lukasbestle: I agree, that makes sense.

src/Http/Environment.php
lukasbestle: I think an `Environment` singleton could already solve this. The `Environment::instance()` method could create the instance, set it to its static singleton property and then trigger the detection.

bastianallgeier: I solved it a bit differently. Rethinking it, another cached singleton would probably lead to more confusion down the road.

src/Http/Environment.php
bastianallgeier: No, it can't. `Environment::detectPortInHost` is even taking care of empty hosts in addition, so the if clause really isn't needed.

src/Http/Environment.php
bastianallgeier: I just checked again and in a non-proxy setup, $_SERVER['HTTPS'] is undefined, if there's no HTTPS setup. IMHO the correct solution would be to simply return false instead of falling back to the local setup. This way, the proxied HTTPS status is considered to be true if it is set correctly, but false if it is undefined.

src/Http/Uri.php
bastianallgeier: Ok, I added a solution to keep things compatible.

src/Http/Environment.php
```diff
@@ -369,8 +369,7 @@ protected function detectForwardedHttps(): bool
             return true;
         }
 
-        // fall back to local setup
-        return $this->detectHttps();
+        return false;
```
lukasbestle, line 372: I'm thinking if we should rather default to `true` for HTTPS. In production it's certainly the most likely case, so it would be safer. Is there a specific reason why the default should be `false`?

lukasbestle: Just saw your comment in the previous discussion above.

> I just checked again and in a non-proxy setup, $_SERVER['HTTPS'] is undefined, if there's no HTTPS setup. IMHO the correct solution would be to simply return false instead of falling back to the local setup.

In the local setup, PHP gets the HTTPS status directly from the web server, so it can 100% rely on that. With a reverse proxy in front, the HTTPS status could really be both and PHP wouldn't know. It's a different situation and I don't think we can compare the two.

I'm still unsure if we should default to one value or if we should consider this case a misconfiguration and throw an exception. If we want to return a fallback default, I think that default should be `true` though.

lukasbestle (approved):

PR #4294, opened 2022-04-29
Improved table previews and usage of table component for plugins

bastianallgeier:
## This PR …

This PR uses the new k-table component to display the plugins in the system view. 

It also introduces redesigned previews for all field types 

## Features

- New field preview in tables for tags, radio, select, multiselect, checkboxes, structures, blocks, layouts, files, pages and users fields.

Screenshot 2022-04-29 at 13 36 43

## Ready?

### For review team


panel/src/components/Forms/Previews/BubblesFieldPreview.vue
```diff
@@ -0,0 +1,56 @@
...
+        if (typeof bubble === "string") {
+          bubble = {
+            value: bubble,
+            text: bubble
+          };
+        }
+
+        for (const option of options) {
+          if (option.value === bubble.value) {
+            bubble.text = bubble.text;
```
lukasbestle: What does this do?

panel/src/components/Forms/Previews/EmailFieldPreview.vue
```diff
@@ -3,10 +3,6 @@ import UrlFieldPreview from "./UrlFieldPreview.vue";
 
 export default {
   extends: UrlFieldPreview,
-  computed: {
-    link() {
-      return this.value?.length > 0 ? "mailto:" + this.value : null;
```
lukasbestle, line 8: With this change, do we even need this component still?

panel/src/components/Forms/Previews/HtmlFieldPreview.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+  <div class="k-html-field-preview" :class="$options.class">
+    {{ column.before }}
+    <!-- eslint-disable-next-line vue/no-v-html -->
+    <div v-html="html" />
```
lukasbestle, line 5: Can’t we just use `value` here and get rid of the computed prop?

panel/src/components/Forms/Previews/TextFieldPreview.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <p class="k-text-field-preview" :class="$options.class">
+    {{ column.before }}
+    <slot>{{ text }}</slot>
```
lukasbestle, line 4: Same here, could be `value`.

panel/src/components/Navigation/Link.vue
```diff
@@ -84,9 +88,16 @@ export default {
         return false;
       }
 
-      // don't route if it's an absolute link
-      if (typeof this.href === "string" && this.href.indexOf("://") > 0) {
-        return false;
+      if (typeof this.href === "string") {
+        // don't route if it's an absolute link
+        if (this.href.includes("://")) {
```
lukasbestle: What about protocol relative URLs? We could catch them here with an additional check “starts with `//`.

panel/src/components/Forms/Previews/BubblesFieldPreview.vue
bastianallgeier: The bubbles field preview is used for select, radio, checkboxes and multiselect fields. This loop translates plain values into proper text if available. I.e. 

```yaml
options: 
  value: a
  text: Option A
```  

The value which the preview receives might be `a`, but it will make sure to show `Option A` instead with this loop.

panel/src/components/Forms/Previews/EmailFieldPreview.vue
bastianallgeier: Good point. Probably only for the class, but we could at least remove the computed link

bastianallgeier: We still need it to keep the correct class name. It could be a breaking change if someone styled it differently.

panel/src/components/Forms/Previews/BubblesFieldPreview.vue
distantnative: @bastianallgeier but shouldn't it then be `bubble.text = option.text`?

panel/src/components/Forms/Previews/HtmlFieldPreview.vue
bastianallgeier: I used the same method here that's also been used in the text field preview. This way you can extend the component and simply overwrite the computed property without having to change the template.

panel/src/components/Navigation/Link.vue
bastianallgeier: ✅

panel/src/components/Forms/Previews/BubblesFieldPreview.vue
bastianallgeier: Oh, ok. I'm obviously blind.

bastianallgeier: ✅

panel/src/components/Forms/Previews/TextFieldPreview.vue
bastianallgeier: See argument above.

afbora:
First review for `ColumnPreviewsA` types

- When list is empty, shows `true` text instead of empty.
- When the list is more than one it looks like below. When this is a very long list, it might not look pretty at all.
screen-capture-1418-Sandbox - Mægazine-localhost

bastianallgeier:
@afbora I wasn't really sure how to handle that to be honest. 

Option A) truncate the list and only keep one line
Option B) align all cells at the top which also isn't ideal

Any other ideas?

afbora:
Vote for `option A` as simple 👍 

Also what would be the option to truncate the contents for all table cell types? 🤔 

````yaml
fields:
  addresses:
    label: Addresses
    type: structure
    excerpt: true
````

May be this is related with https://kirby.nolt.io/52 idea

bastianallgeier:
@afbora I would leave it as it is for now and solve multi-line fields later. The feedback is still there so we can pick it up when we have time. Otherwise this is getting too complicated.

afbora:
OK, what about `When list is empty, shows true text instead of empty.` issue?

distantnative:
I think the origin of this bug is that the mixin sets the default for `value` to `true`. Not sure what that's for, @bastianallgeier - why not allowing empty/no value?

bastianallgeier:
I believe that this was a mistake. I probably thought that this would make sure that all types are allowed.

bastianallgeier:
Ha, found it. TIL that if you use Boolean as the first type in an array of types for a prop, the prop value will be automatically turned into a Boolean, even if Strings or other types are allowed as well. So for Booleans the type must really be the last thing in the array.

afbora (approved):
All fields looks great now! 💛

PR #4295, opened 2022-04-29
System health checks

bastianallgeier:
## Features

- New system health checks to detect unintentionally exposed parts of your installation (git repo, content folder, site folder, kirby folder) 

Screenshot 2022-04-29 at 14 20 11

## Ready?

### For review team

### DIFF
--- a/src/Cms/System.php
+++ b/src/Cms/System.php
@@ -109,6 +109,8 @@ class System
                 return $url . '/composer.json';
             case 'site':
                 return $url . '/templates/default.php';
+            default:
+                return null;
         }
     }
 
--- a/src/Cms/System.php
+++ b/src/Cms/System.php
@@ -94,14 +94,43 @@ class System
      * @param string $folder 'git', 'content', 'site', 'kirby'
      * @return string|null
      */
-    public function folderUrl(string $folder): ?string
+    public function exposedFileUrl(string $folder): ?string
     {
-        if ($folder === 'git') {
-            return url('.git/config');
+        if (!$url = $this->folderUrl($folder)) {
+            return null;
         }
 
-        $root  = $this->app->root($folder);
+        switch ($folder) {
+            case 'content':
+                return $url . '/' . basename($this->app->site()->contentFile());
+            case 'git':
+                return $url . '/config';
+            case 'kirby':
+                return $url . '/composer.json';
+            case 'site':
+                return $url . '/templates/default.php';
+            default:
+                return null;
+        }
+    }
+
+    /**
+     * Returns the URL to a system folder
+     * if the folder is located in the document
+     * root. Otherwise it will return null.
+     *
+     * @param string $folder 'git', 'content', 'site', 'kirby'
+     * @return string|null
+     */
+    public function folderUrl(string $folder): ?string
+    {
         $index = $this->app->root('index');
+
+        if ($folder === 'git') {
+            $root = $index . '/.git';
+        } else {
+            $root = $this->app->root($folder);
+        }
 
         if ($root === null || is_dir($root) === false || is_dir($index) === false) {
             return null;
@@ -123,18 +152,7 @@ class System
         $path = trim(Str::after($root, $index), '/');
 
         // build the absolute URL to the folder
-        $url = url($path);
-
-        switch ($folder) {
-            case 'content':
-                return $url . '/' . basename($this->app->site()->contentFile());
-            case 'kirby':
-                return $url . '/composer.json';
-            case 'site':
-                return $url . '/templates/default.php';
-            default:
-                return null;
-        }
+        return url($path);
     }
 
     /**
--- a/src/Cms/System.php
+++ b/src/Cms/System.php
@@ -108,13 +108,22 @@ class System
             case 'kirby':
                 return $url . '/composer.json';
             case 'site':
-                $files = glob($this->app->root('site') . '/**/*.*');
-                $file  = $files[0] ?? null;
+                $root  = $this->app->root('site');
+                $files = glob($root . '/blueprints/*.yml');
 
-                if (empty($file) === true) {
+                if (empty($files) === true) {
+                    $files = glob($root . '/templates/*.*');
+                }
+
+                if (empty($files) === true) {
+                    $files = glob($root . '/snippets/*.*');
+                }
+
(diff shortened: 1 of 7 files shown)


afbora (commented):
I think it will not work correctly in custom app setup, am I wrong?

bastianallgeier:
@afbora when you use a public folder setup, you are protected anyway. It wouldn't work if you rename the site or content folders and still keep them in the webroot, but that would also be kind of stupid :)

afbora:
For example, I changed the `content` folder to `data`. I didn't do an update from `.htaccess`. Since there is no `content` folder, the system will not be able to access it and will show it as `hidden` so that is safe (actually accessible.). So instead of `window.location.origin` we should use dynamic routes. `this.system.root.content + "/site.txt"` shouldn't it be something like this?

bastianallgeier:
In this case, you are right. But it will be difficult to find the URL for the custom content folder. We don't have a URL setting for it, like for the media folder. 

We could only try to detect for each folder if it is contained in the web root and then detect the URL based on the path. 

Here's a bit of dummy code: 

```php
$content = '/afbora/www/data';
$webroot = '/afbora/www';

if ($content isAChildOf $webroot) {
  $path = Str::after($content, $webroot);
  $url = kirby()->url() . '/' . $path;
}
```

afbora:
Yes you are right, we know root paths but we don't have full url information. I think your sample code for this seems to work well.

A second alternative is to hide this checks data if custom setup is used.

lukasbestle:
Regarding UI: I think we don't need to display the passing checks. Server and PHP are always useful but with the checks it's IMO hard to keep the overview if all of them are displayed.

We could do something like this:

- If everything is fine:
  > Security
  > No suggestions
- If there are one or more issues: Either display each issue or a single box with a list. This way we could also visually highlight that box with the issues more.

src/Cms/System.php
```diff
@@ -86,6 +86,56 @@ public function curl(): bool
         return extension_loaded('curl');
     }
 
+    /**
+     * Returns the URL to any system folder
```
lukasbestle: The method rather returns the URL to a file inside the folder.

src/Cms/System.php
```diff
@@ -86,6 +86,56 @@ public function curl(): bool
...
+
+        $root  = realpath($root);
+        $index = realpath($index);
+
+        // windows
+        $root  = str_replace('\\', '/', $root);
+        $index = str_replace('\\', '/', $index);
+
+        // the folder is not within the document root?
+        if (Str::contains($root, $index) === false) {
```
```suggestion
        if (Str::startsWith($root, $index) === false) {
```

bastianallgeier:
@lukasbestle I've redesigned the list of security issues. It will only be shown if there are any …
Screenshot 2022-05-09 at 12 49 27

src/Cms/System.php
```diff
@@ -86,6 +86,75 @@ public function curl(): bool
...
+
+        switch ($folder) {
+            case 'content':
+                return $url . '/' . basename($this->app->site()->contentFile());
+            case 'git':
+                return $url . '/config';
+            case 'kirby':
+                return $url . '/composer.json';
+            case 'site':
+                return $url . '/templates/default.php';
```
afbora: I guess we will ignore the custom `templates` setup as it would be complex?

bastianallgeier: I was wondering about that as well. The default template is such a great place to check because it is pretty much the only thing that's required in the site folder. But with a custom template engine it would be really difficult. Maybe we can fix that later.

afbora:
I tested it, it works great 👍 
Not critical, but a few minor points:

- Retry button doesn't work on security items related to `.htaccess` changes
- Doesn't use `hidden` and `exposed` translation lines

src/Cms/System.php
lukasbestle: We always say that and never keep track of this stuff. :)

Possible solution:

1. Check if the `default.php` exists in the actual `templates` root. If yes, check if `templates` is inside the `site` root. If yes, use this path.
2. If `default.php` does not exist, use `scandir` to find any file with the basename `default` (with whatever extension).
3. If the `templates` dir is outside of the `site` dir, ?? Maybe fall back to `config`?

bastianallgeier: Wait a second. I think we could actually use scandir to find any file within the site folder and then check access to that file.

bastianallgeier:
@afbora I've fixed your points.

bastianallgeier:
@lukasbestle I've added the check based on the glob result. That seems to work pretty great. It takes whatever file can be found first.

afbora (dismissed):
I predict that it will still not work properly in some custom root usages (because we still think of the `site` root path as 2 parts (`/site/*/*`) and in the root) But I don't see any problem in continuing as it is because it will give correct results for general uses (I think %99 of usages). 

So LGTM and I think it's a very useful feature 👍

src/Cms/System.php
lukasbestle: We need to make sure though that the "any file" does not leak sensitive information via the API. With a template it would be no problem, but apart from that we need to be careful.

bastianallgeier:
They only other option would be to check for each custom root and find out if it is in the web root. That would be super excessive and only useful for a very small percentage of users.

src/Cms/System.php
bastianallgeier: You got a point there. Maybe this entire section should only be accessible for admins anyway.

bastianallgeier: I fixed it by only checking for files within blueprints, templates and snippets folders. In combination with permissions, I think we are good here now

distantnative:
Looks good to me - only the links aren't leading anywhere yet.

distantnative:
My suggestions for the links:

Debug: https://getkirby.com/docs/reference/system/options/debug
HTTPS: https://getkirby.com/docs/guide/security#secure-your-server__use-and-enforce-https
Content/site/kirby/git folder: https://getkirby.com/docs/guide/security#secure-your-server__block-access-to-sensitive-folders-and-files

bastianallgeier:
@distantnative I think we should use shortcuts for those links that we can then use in our site router to redirect them wherever it fits. Otherwise we always have to fix them in the Panel, which is a lot less simple compared to a change in our router.

distantnative:
Ok fine with me

distantnative (approved):

PR #4296, opened 2022-04-29
First steps with code splitting

bastianallgeier:
The title says it all. 

This first step is really simple but still quite effective. 

Screenshot 2022-04-29 at 16 36 40


distantnative:
It seems to still have the same issues of not resolving properly to media URLS including the current hash :/

Screen Shot 2022-05-01 at 21 51 16

panel/src/components/Forms/Writer/Writer.vue
```diff
@@ -149,7 +147,9 @@ export default {
       }
     }
   },
-  mounted() {
+  async mounted() {
+    const Editor = (await import("./Editor.js")).default;
```
johannschopplich: More simple solution:

```js
const { default: Editor } = await import("./Editor.js");
```

bastianallgeier:
@distantnative that's so weird. Works absolutely painless for me. Maybe a different Vite version?

distantnative:
@bastianallgeier no - and that shouldn't matter anyways. We're using the committed dist files - so vite isn't part of the equation at that point anymore.

Howe I can reproduce my error:
1. Check out this PR
2. Delete `sandbox/public/media`
3. Go to http://sandbox.test/panel/pages/notes+exploring-the-universe
4. See console

How the source code looks for me:
Screen Shot 2022-05-06 at 22 32 33

You can see the difference of our fixed asset entries at the bottom and the dynamically inserted ones at the end of the head section.

distantnative:
We might need to wait for this in vite 3 - https://github.com/vitejs/vite/pull/7644 :(

distantnative:
I think I found a way 🥳

@bastianallgeier Please have a close look at https://github.com/getkirby/kirby/pull/4296/commits/35f495d893a4024ed2ef9e67e5d9f5ac7cc4d2b5 - maybe that route could be more elegant

lukasbestle:
@distantnative Couldn't this cause issues with aggressive intermediate or browser caches? Because the URL no longer contains the version hash, these caches wouldn't know that a file was updated.

distantnative:
@lukasbestle could be - but then we need to completely restructure Panel assets. The hash the needs to be put in the file name by vite, not by Kirby's backend in the path

lukasbestle:
To be honest I'd much rather prefer a reliable setup over one with slight performance gains. If we can combine both – great. But without version hashes I fear we will get loads of issue reports that are hard to track down and debug.

distantnative:
I'm still confused that @bastianallgeier said it worked for him before my commits

afbora:
I'm getting following error when add new text block.

> Cannot read properties of null (reading 'destroy')
> index.js:1 TypeError: Cannot read properties of null (reading 'focus')
> index.js:1 TypeError: Cannot read properties of null (reading 'destroy')
> http://localhost/test/kirby/develop/panel/js/vendor.js net::ERR_ABORTED 500 (Internal Server Error)
> http://localhost/test/kirby/develop/panel/js/Editor.js net::ERR_ABORTED 500 (Internal Server Error)

I'm getting following output on console when using `writer` field.

> http://localhost/test/kirby/develop/panel/js/Editor.js net::ERR_ABORTED 500 (Internal Server Error)
> http://localhost/test/kirby/develop/panel/js/vendor.js net::ERR_ABORTED 500 (Internal Server Error)

distantnative:
I can reproduce @afbora's issues - not sure yet why

distantnative:
> I'm getting following error when add new text block.

So the reason seems to be that the `focus` method of the new block is called before the `mounted` hook of the new block has finished running, because it is marked as async (my assumption). Because of that, `this.editor` is still `null`.

Possible fix could be to use `setTimeout` with a sufficient delay here (though we could only estimate how long it takes to load the Editor chunk): https://github.com/getkirby/kirby/blob/develop/panel/src/components/Forms/Blocks/Blocks.vue#L232
instead of `this.$nextTick(() => {`

lukasbestle:
Sounds like a hack to me. If we make the delay too short, it will break on slow connections and users won't know why. If we make it too long, it will feel sluggish.

distantnative:
In that case, splitting Editor won't work at all

bastianallgeier:
Couldn't we simply wait for the editor to load before we call focus? Import is a promise so it shouldn't be too hard.

distantnative:
But how? We call the focus method from outside the component itself (chaining ti back to Blocks.vue) - I'm not aware of a way to call a component's method from outside but waiting until the async mounted/created lifecycle hooks of that component have finished. IMO making the lifecycle hook async also kinda implies that other stuff can already happen while that lifecycle hook (where we do the import) runs separately.

lukasbestle:
Couldn't the lifecycle hook itself use `async`/`await`?

distantnative:
@lukasbestle that's what it does - but from my understanding that's also why the focus method can already be called while the lifecycle hook hasn't even finished (since it's running async)

lukasbestle:
Ah, I just found out that `mounted()` gets called *after* the component has already been mounted (makes sense considering the name 😆). The point of an `async` function is that other functions can wait for it to resolve, but if the component is already mounted, Vue won't have a reason to wait for anything. What we could try is to load the editor in the `created()` hook instead, which is called *before* the component is mounted. If that doesn't work, then `beforeCreate()`.

https://v2.vuejs.org/v2/guide/instance.html#Lifecycle-Diagram

distantnative:
Didn't have success with either of those.

distantnative:
The first issue (path to assets) might be solvable with the new vite 3 alpha https://github.com/vitejs/vite/pull/7644

rasteiner:
> But how? We call the focus method from outside the component itself (chaining ti back to Blocks.vue) - I'm not aware of a way to call a component's method from outside but waiting until the async mounted/created lifecycle hooks of that component have finished. IMO making the lifecycle hook async also kinda implies that other stuff can already happen while that lifecycle hook (where we do the import) runs separately.

I think `this.editor` should be a promise. Methods using `this.editor` should therefore either also be `async` and reference it only with await
```js
async focus() {
  (await this.editor).focus()
}
```

or drop the action when the editor is not ready:
```js
editorSync() {
  let editor = null
  // then is synchronous if the promise has already resolved
  this.editor.then(e => editor = e); 
  return editor;
}
command(command, ...args) {
  const editor = this.editorSync();
  if(!editor) return;

  editor.command(command, ...args);
  
  // could also be written as the one liner:
  // this.editorSync()?.command(command, ...args);
},

getHTML() {
  return this.editorSync()?.getHTML() ?? ''; // or something... 
},
```

distantnative:
I would probably put this on the back burner for 3.7.1 or so where we can tackle and see if we can solve the remaining nasty little issues.

bastianallgeier:
I agree. Should we convert it to a draft then?

distantnative:
Continue here https://github.com/getkirby/kirby/pull/4504

afbora:
I tried dozens of different methods hundreds of times, but I couldn't separate the editor. I don't understand why `Editor.js` is trying to be reloaded together with `vendor.js` while `vuedraggable` can be imported correctly. I'm guessing because `vuedraggable` is a module while `Editor` is just a class. I think it looks very difficult 😞

distantnative:
@afbora why are you using this old branch? This was still vite 2 and has the problems you described.

My new PR is based on vite 3 that solves these issues by default. What's still a problem is that we cannot load Editor async because all kind of methods (inside and outside the component) access `this.editor` and can't handle that the editor might not yet be loaded.

afbora:
No problem 👍 I already used your new PR but since it only includes `vuedraggable` now, I wanted to write it under this PR to avoid confusion.

PR #4298, opened 2022-04-30
New stats section

bastianallgeier:
## Features

The brand new statistics section, can be used to show beautiful stats for your site or shop. Revenue, transactions, Twitter likes, page impressions … it's totally up to you. You can add as many reports to a stats section as needed. Reports can be customized with our query syntax and integrated easily into page models or site methods. 

```yaml
stats:
  type: stats
  reports:
    - label: Revenue
      value: €29,682
      info: +112.5%
      link: https://getkirby.com/shop
      theme: positive
    - label: Transactions
      value: 265
      info: +82.8%
      theme: positive
    - label: Avg. Transaction
      value: €112.01
      info: +16.3%
      theme: positive
    - label: Refunds
      value: €91.75
      info: +49.5%
      theme: positive
```

… with template strings and queries …

```yaml
stats:
  type: stats
  reports:
    - page.revenue
    - label: 
        de: Transactions
        en: Transaktionen
      value: "{{ page.transactions }}"
      info: "{{ page.transactionsIncrease }}"
```

… a custom array of reports coming from a query

```yaml
stats:
  type: stats
  reports: page.reports
```

Screenshot 2022-04-30 at 15 45 45

## Ready?

### For review team

### DIFF
--- a/config/sections/stats.php
+++ b/config/sections/stats.php
@@ -1,0 +1,56 @@
+<?php
+
+use Kirby\Toolkit\I18n;
+
+return [
+    'mixins' => [
+        'headline',
+    ],
+    'props' => [
+        'reports' => function ($reports = null) {
+            if ($reports === null) {
+                return [];
+            }
+
+            if (is_string($reports) === true) {
+                $reports = $this->model()->query($reports);
+            }
+
+            if (is_array($reports) === false) {
+                return [];
+            }
+
+            return $reports;
+        }
+    ],
+    'computed' => [
+        'reports' => function () {
+            $reports = [];
+            $model   = $this->model();
+
+            $value = function ($value) use ($model) {
+                return $value === null ? null : $model->toString($value);
+            };
+
+            foreach ($this->reports as $report) {
+                if (is_string($report) === true) {
+                    $report = $model->query($report);
+                }
+
+                if (is_array($report) === false) {
+                    continue;
+                }
+
+                $reports[] = [
+                    'label' => I18n::translate($report['label'], $report['label']),
+                    'value' => $value($report['value'] ?? null),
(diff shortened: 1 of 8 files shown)


afbora:
I liked it ❤️  I already made a plugin for this but mine looks crappy 🙈 https://github.com/afbora/kirby-widgets

I just have one small suggestion. Can we add a `default`, `before` and `after` options? It would be useful to pull some data from the page data and add a prefix or suffix and show a default data when there is no data.

````yaml
stats:
  type: stats
  reports:
    - label: Revenue
      before: €
      default: 0.00
      value: "{{ page.revenue }}"
      info: +112.5%
      link: https://getkirby.com/shop
      theme: positive
````

**Comment by lukasbestle** [comment]
Those would only apply to the text, not the info. If we want to support both, it could be a child array for both options. But then the option structure gets incredibly complex.

So TBH I feel like the formatting and the default value should be handled by the called methods.

**Comment by bastianallgeier** [comment]
I agree with @lukasbestle. Since the label, value and info can all come from templates, it's pretty easy to add before and after text and the methods can take care of proper defaults.

File: config/sections/stats.php [hunk 1]
```diff
@@ -0,0 +1,56 @@
...
+        }
+    ],
+    'computed' => [
+        'reports' => function () {
+            $reports = [];
+            $model   = $this->model();
+
+            $value = function ($value) use ($model) {
+                return $value === null ? null : $model->toString($value);
+            };
```
distantnative: That would be a great spot for shorthand:

```php
$value = fn ($value) => $value === null ? null : $model->toString($value);
```

distantnative (approved):

Issue #4297, opened 2022-04-29
Preview image from other page not shown when page child of site and status `published`


texnixe:
## Description
When using a pages section with first level pages, `status: published` and the queried preview image is from another page, the preview image is not shown.

**Expected behavior**  
The preview image should always be shown

**Screenshots**  
<img width="1191" alt="Bildschirmfoto 2022-04-29 um 21 52 58" src="https://user-images.githubusercontent.com/5312492/166060403-837aa0f8-8f7d-4bf0-a780-5086bb466f5f.png">

## To reproduce
1. Create several page with template `test` as children of the site object
2. Create a files page with template `files` with at least 1 image
3. Create a site.yml blueprint with the following content:

```
Title: Site

columns:
  left:
    width: 1/2
    sections:
      notes:
        headline: Notes (published)
        type: pages
        parent: kirby.page('notes')
        image: kirby.page('files').image('last-tree-standing.jpg')
        status: published
        layout: cards
        size: tiny
        template: note
  right:
    width: 1/2
    sections:
      published:
        headline: First level test page (Published)
        type: pages
        layout: cards
        size: tiny
        image: kirby.page('files').image('last-tree-standing.jpg')
        status: published
        templates: test
      listed:
        headline: First level test page (Listed)
        type: pages
        status: listed
        layout: cards
        size: tiny
        image: kirby.page('files').image('last-tree-standing.jpg')
        template: test
      drafts:
        headline: First level test page (Draft)
        type: pages
        status: draft
        layout: cards
        size: tiny
        image: kirby.page('files').image('last-tree-standing.jpg')
        template: test
      unlisted:
        headline: First level test page (Unlisted)
        type: pages
        status: unlisted
        layout: cards
        size: tiny
        image: kirby.page('files').image('last-tree-standing.jpg')
        template: test
```

Interestingly, the images also show up in the section with the published pages when adding page with the `files` template:

```
      published:
        headline: First level test page (Published)
        type: pages
        layout: cards
        size: tiny
        image: kirby.page('files').image('last-tree-standing.jpg')
        status: published
        templates:
          - test
          - files # template of the files page
```

Slightly different test setup can be found here: https://github.com/ersoma/starterkit-bughunt

## Your setup
**Kirby Version**  
3.6.5

Additional context:
https://forum.getkirby.com/t/page-image-is-not-displayed-in-pages-section-when-referencing-a-filtered-out-page/25352/5

afbora:
[...]

PR #4308, opened 2022-05-18
Use internal `->filter()` method instead loop that unsetting the original data

afbora:
## This PR …

The issue was that the `unset()` function used during the loop was updating the original data/collection. With @lukasbestle suggestion, we solved the problem by filtering without updating the original data/collection using the internal `->filter()` method.

We also considered the performance issue and we did not see any problem as there was not even a 0.1 second difference in 100k virtual pages.

I haven't added unit testing for now. I can write unit tests if you OK with the PR.

### Fixes

- Preview image from other page was not shown when the page is a child of site and status `published` was queried #4297

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -115,21 +115,24 @@ return [
                     $pages = $this->parent->childrenAndDrafts();
             }
 
-            // loop for the best performance
-            foreach ($pages->data as $id => $page) {
-
+            // filters pages that are protected and not in the templates list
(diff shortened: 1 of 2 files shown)


bastianallgeier (dismissed):

bastianallgeier:
The PR looks good to me. The unit tests would be great to make Codecov and us happy :)

afbora:
I've added unit test and ready to review 🚀

distantnative (approved):

PR #4315, opened 2022-05-25
Use k-stats for system view

distantnative:
### Enhancements
- Uses `k-stats` for system view
Screen Shot 2022-05-27 at 20 01 49

## Ready?

### For review team

### DIFF
--- a/config/sections/stats.php
+++ b/config/sections/stats.php
@@ -21,6 +21,9 @@ return [
             }
 
             return $reports;
+        },
+        'size' => function (string $size = 'large') {
+            return $size;
         }
     ],
     'computed' => [
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -343,6 +343,7 @@
   "license.register.label": "Please enter your license code",
   "license.register.success": "Thank you for supporting Kirby",
   "license.unregistered": "This is an unregistered demo of Kirby",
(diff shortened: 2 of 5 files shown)


afbora (commented):
- The license registration button is both a link and a dialog trigger. When we open in a new tab, it goes to another page. I would expect it to be a `button` if it is not an external link.
- I would prefer the text headers `xl` instead`2xl`.
- Considering the translations in different languages, there is a high probability of gliding/overflow issues on mobile.

screen-capture-1426-System - Mægazine-localhost

distantnative:
1. I will look into why `k-link` isn't rendered only as button in this case
2. That's how `k-stats` is styled. Not sure if we should overwrite styles in this case
3. @bastianallgeier maybe makes sense to let `k-stats` break into single column a bit earlier in general?

lukasbestle:
It looks. So. Great.
Awesome use of `k-stats`. 👍

I agree on the breakpoint for the single column.

distantnative:
Ok I think I solved 2 and 3.

I also realised why 1 is - was mixing up `k-link` and `k-button-link`. Have to see how to solve, just using `k-button` is making some problems.

distantnative:
Ok, I think I solved 1 as well.

afbora (approved):
LGTM, loved 💛

PR #4330, opened 2022-06-02
UI Updates

bastianallgeier:
### Refined UI design

With a refined design for fields, dialogs, dropdowns and more, the Panel got a subtle fresh new paint.

### DIFF
--- a/panel/src/components/Layout/Header.vue
+++ b/panel/src/components/Layout/Header.vue
@@ -1,5 +1,9 @@
 <template>
-  <header :data-editable="editable" class="k-header">
+  <header
+    :data-editable="editable"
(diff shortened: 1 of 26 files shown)


panel/src/components/Layout/Header.vue
```diff
@@ -81,6 +85,9 @@ export default {
   margin-bottom: 2rem;
   border-bottom: 1px solid var(--color-border);
 }
+.k-header[data-tabs] {
```
distantnative: Remember that we should already do `="true"`. Vue 3 will render data attributes with `false` as value still as `="false"` and your current CSS selector would still match that

distantnative (approved):

PR #4331, opened 2022-06-02
New V::empty and V::notEmpty validators

bastianallgeier:
### New `V::empty()` and `V::notEmpty()` validators

Checking for empty values in PHP is less easy than it sounds like. `empty()` is often giving wrong results. Especially if you pass 0 or '0'. The new validators help with this task. They also serve as underlying logic for a better `V::required()` validator. It can now be used in two ways. 

#### With a reference array

```php
V::required('myField', ['myField' => 'Some value']);
```

or just a value

```php
V::required('My value');
```

This makes the method more versatile and it can be used in `V::value()` 

```php
V::value('My value', [
  'required' => true,
  'minlength' => 5,
]);
```

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/V.php
+++ b/src/Toolkit/V.php
@@ -391,6 +391,23 @@ V::$validators = [
     },
 
     /**
+     * Checks for empty values
+     */
+    'empty' => function ($value = null): bool {
+        $empty = ['', null, []];
+
+        if (in_array($value, $empty, true) === true) {
+            return true;
+        }
+
+        if (is_countable($value) === true) {
+            return count($value) === 0;
+        }
+
+        return false;
+    },
+
+    /**
      * Checks if the given string ends with the given value
      */
     'endsWith' => function (string $value, string $end): bool {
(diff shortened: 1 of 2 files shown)


src/Toolkit/V.php
```diff
@@ -390,6 +390,23 @@ public static function __callStatic(string $method, array $arguments): bool
...
 
+    /**
+     * Checks for empty values
+     */
+    'empty' => function ($value = null): bool {
+        $empty = ['', null, []];
+
+        if (in_array($value, $empty, true) === true) {
+            return true;
+        }
```
afbora, line 401: What about empty space, `'    '`? Do you think that we should use `trim` on strings?

distantnative: Maybe trimming white space would make sense, I agree

bastianallgeier: I don't think the validator should do that. It might be the right choice 90% of the time, but a single space could also be interpreted as an actual choice by the user to have that as a intended value.

distantnative: Ok

distantnative (approved):

PR #4334, opened 2022-06-04
Disable failing JS unit tests

distantnative:


lukasbestle (commented):
Have you been able to track down why the tests fail?

To be honest I think we should either fix the tests or remove them entirely. With the commented assertion, the removed snapshot and the renamed test file we won't know in the future why we did it.

distantnative:
No, neither Bastian or I have managed. Though I've also only looked into it briefly. Since our JS unit tests are very basic and only tests so far, I think it's ok to do that and not rely on them.

To be honest, I'd rather keep it that way and wonder later than forgetting about the old code or even having trouble to recover it when we tackle it again.

lukasbestle:
Can we keep the snapshot file then? Would also be good to add links to this PR for future reference.

lukasbestle (approved):

PR #4335, opened 2022-06-04
Make `@todo` changes for 3.7.0

distantnative:
### Deprecation warnings
- API field `page.next` now throws deprecation warning and will be removed in v3.8.0.
- API field `page.prev` now throws deprecation warning and will be removed in v3.8.0.
- `markdown` component: using the deprecated `$inline` parameter will now throw a warning and will be removed in v3.8.0. Use `$options['inline']` instead.
- `Cms\App::kirbytext()`: using the deprecated `$inline` parameter will now throw a warning and will be removed in v3.8.0. Use `$options['markdown']['inline']` instead.
- `Cms\App::markdown()`: using the deprecated `$inline` parameter will now throw a warning and will be removed in v3.8.0. Use `$options['inline']` instead.
- `Cms\File::dragText()` now throws a deprecation warning and will be removed in v3.8.0. Use `$file->panel()->dragText()` instead.
- `Cms\File::panelOptions()` now throws a deprecation warning and will be removed in v3.8.0. Use `$file->panel()->options()` instead.
- `Cms\File::panelPath()` now throws a deprecation warning and will be removed in v3.8.0. Use `$file->panel()->path()` instead.
- `Cms\File::panelPickerData()` now throws a deprecation warning and will be removed in v3.8.0. Use `$file->panel()->pickerData()` instead.
- `Cms\File::panelUrl()` now throws a deprecation warning and will be removed in v3.8.0. Use `$file->panel()->url()` instead.
- `Cms\ModelWithContent::panelIcon()` now throws a deprecation warning and will be removed in v3.8.0. Use `$model->panel()->image()` instead.
- `Cms\ModelWithContent::panelImage()` now throws a deprecation warning and will be removed in v3.8.0. Use `$model->panel()->image()` instead.
- `Cms\ModelWithContent::panelOptions()` now throws a deprecation warning and will be removed in v3.8.0. Use `$model->panel()->options()` instead.
- `Cms\Page::dragText()` now throws a deprecation warning and will be removed in v3.8.0. Use `$page->panel()->dragText()` instead.
- `Cms\Page::panelId()` now throws a deprecation warning and will be removed in v3.8.0. Use `$page->panel()->id()` instead.
- `Cms\Page::panelPath()` now throws a deprecation warning and will be removed in v3.8.0. Use `$page->panel()->path()` instead.
- `Cms\Page::panelPickerData()` now throws a deprecation warning and will be removed in v3.8.0. Use `$page->panel()->pickerData()` instead.
- `Cms\Page::panelUrl()` now throws a deprecation warning and will be removed in v3.8.0. Use `$page->panel()->url()` instead.
- `Cms\Site::panelPath()` now throws a deprecation warning and will be removed in v3.8.0. Use `$site->panel()->path()` instead.
- `Cms\Site:panelUrl()` now throws a deprecation warning and will be removed in v3.8.0. Use `$site->panel()->url()` instead.
- `Cms\User::panelPath()` now throws a deprecation warning and will be removed in v3.8.0. Use `$user->panel()->path()` instead.
- `Cms\User::panelPickerData()` now throws a deprecation warning and will be removed in v3.8.0. Use `$user->panel()->pickerData()` instead.
- `Cms\User:panelUrl()` now throws a deprecation warning and will be removed in v3.8.0. Use `$user->panel()->url()` instead.
- Deprecated `Panel\Document::customCss()` now throws a warning and will be removed in v3.8.0. Use `Panel\Document::customAsset('panel.css')` instead.
- Deprecated `Panel\Document::customJs()` now throws a warning and will be removed in v3.8.0. Use `Panel\Document::customAsset('panel.js')` instead.

### Removed deprecated code

- Deprecated API field `page.panelIcon` has been removed. Use `page.panelImage` instead.
- Deprecated API field `file.panelIcon` has been removed. Use `file.panelImage` instead.
- Deprecated `GET (:all)/lock` API endpoint has been removed.
- Deprecated `GET (:all)/unlock` API endpoint has been removed.
- Deprecated `GET pages/(:any)/children/blueprints` API endpoint has been removed. Use `GET pages/(:any)/blueprints` instead.
- Deprecated `GET site/children/blueprints` API endpoint has been removed. Use `GET site/blueprints` instead.
- `page` helper now always only returns a `Kirby\Cms\Page` object or `null`, never a pages collection. Only allows passing a single id as parameter.
- `pages` helper now always only returns a `Kirby\Cms\Pages` collection or `null`, never single page object.
- Deprecated JS API method `files.rename()` has been removed. Use `files.changeName()` instead.
- Deprecated JS API method `pages.slug()` has been removed. Use `pages.changeSlug()` instead.
- Deprecated JS API method `pages.status()` has been removed. Use `pages.changeStatus()` instead.
- Deprecated JS API method `pages.template()` has been removed. Use `pages.changeTemplate()` instead.
- Deprecated JS API method `pages.title()` has been removed. Use `pages.changeTitle()` instead.
- Deprecated JS API method `site.title()` has been removed. Use `site.changeTitle()` instead.
- Deprecated JS API method `system.info()` has been removed. Use `system.get()` instead.
- Deprecated `Cms\App::setLocale()` has been removed. Use `Toolkit\Locale::set()` instead.
- Deprecated `Cms\Block::_key()` has been removed. Use `Block::type()` instead.
- Deprecated `Cms\Block:: _uid()` has been removed. Use `Block::id()` instead.
- `Http\Server::host()`: removed deprecated `$forwarded` parameter.
- `Http\Server::port()`: removed deprecated `$forwarded` parameter.
- `Http\Uri::current()`: removed deprecated `$forwarded` parameter.
- `Http\Uri::index()`: removed deprecated `$forwarded` parameter.
- `Http\Url::index()`: removed deprecated `$forwarded` parameter.
- Deprecated `Toolkit\I18n::fallback()` has been removed. Use `I18n::fallbacks()` instead.
- `Toolkit\Str::template()`: The `$fallback`, `$start` and `$end` parameters have been removed. Please pass an array to the `$options` parameter instead with `fallback`, `start` or `end` keys.

### Breaking changes
- Creating a `Kirby\Cms\File` object requires the `parent` property now.
- `Toolkit\Str::toBytes()` strictly only accepts a string as parameter now.

### Enhancements

- `Toolkit\Xml::value()` now uses better logic to determine whether a `CDATA` block is needed.

## Ready?

### For review team

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -150,7 +150,7 @@ return [
         // warning for deprecated fourth parameter
         if ($inline === null) {
             $inline = false;
-        } else {
+        } elseif (isset($options['inline']) === false) {
             Helpers::deprecated('markdown component: the $inline parameter is deprecated and will be removed in Kirby 3.8.0. Use $options[\'inline\'] instead.');
         }
 
--- a/config/components.php
+++ b/config/components.php
@@ -4,6 +4,7 @@ use Kirby\Cms\App;
 use Kirby\Cms\Collection;
 use Kirby\Cms\File;
 use Kirby\Cms\FileVersion;
+use Kirby\Cms\Helpers;
 use Kirby\Cms\Template;
 use Kirby\Data\Data;
 use Kirby\Email\PHPMailer as Emailer;
@@ -140,12 +141,18 @@ return [
      * @param array $options Markdown options
      * @param bool $inline Whether to wrap the text in `<p>` tags (deprecated: set via $options['inline'] instead)
      * @return string
-     * @todo add deprecation warning for $inline parameter in 3.7.0
      * @todo remove $inline parameter in in 3.8.0
      */
-    'markdown' => function (App $kirby, string $text = null, array $options = [], bool $inline = false): string {
+    'markdown' => function (App $kirby, string $text = null, array $options = [], bool $inline = null): string {
         static $markdown;
         static $config;
+
+        // warning for deprecated fourth parameter
+        if ($inline === null) {
+            $inline = false;
+        } else {
+            Helpers::deprecated('markdown component: the $inline parameter is deprecated and will be removed in Kirby 3.8.0. Use $options[\'inline\'] instead.');
+        }
 
         // support for the deprecated fourth argument
         $options['inline'] ??= $inline;
--- a/config/components.php
+++ b/config/components.php
@@ -151,7 +151,9 @@ return [
         if ($inline === null) {
             $inline = false;
         } elseif (isset($options['inline']) === false) {
+            // @codeCoverageIgnoreStart
             Helpers::deprecated('markdown component: the $inline parameter is deprecated and will be removed in Kirby 3.8.0. Use $options[\'inline\'] instead.');
+            // @codeCoverageIgnoreEnd
         }
 
         // support for the deprecated fourth argument
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -3,7 +3,6 @@
 use Kirby\Cms\App;
 use Kirby\Cms\Helpers;
 use Kirby\Cms\Html;
-use Kirby\Cms\Pages;
 use Kirby\Cms\Url;
 use Kirby\Filesystem\Asset;
 use Kirby\Filesystem\F;
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -420,7 +420,7 @@ if (Helpers::hasOverride('page') === false) { // @codeCoverageIgnore
      * @param string|null $id
      * @return \Kirby\Cms\Page|null
      */
-    function page($id = null)
+    function page(?string $id = null)
     {
         if (empty($id) === true) {
             return App::instance()->site()->page();
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -3,6 +3,7 @@
 use Kirby\Cms\App;
 use Kirby\Cms\Helpers;
 use Kirby\Cms\Html;
+use Kirby\Cms\Pages;
 use Kirby\Cms\Url;
 use Kirby\Filesystem\Asset;
 use Kirby\Filesystem\F;
@@ -416,23 +417,16 @@ if (Helpers::hasOverride('page') === false) { // @codeCoverageIgnore
      * Fetches a single page or multiple pages by
      * id or the current page when no id is specified
      *
-     * @param string|array ...$id
-     * @return \Kirby\Cms\Page|\Kirby\Cms\Pages|null
-     * @todo reduce to one parameter in 3.7.0 (also change return and return type)
+     * @param string $id
+     * @return \Kirby\Cms\Page|null
      */
-    function page(...$id)
+    function page($id)
     {
         if (empty($id) === true) {
             return App::instance()->site()->page();
         }
 
-        if (count($id) > 1) {
-            // @codeCoverageIgnoreStart
-            Helpers::deprecated('Passing multiple parameters to the `page()` helper has been deprecated. Please use the `pages()` helper instead.');
-            // @codeCoverageIgnoreEnd
-        }
-
-        return App::instance()->site()->find(...$id);
+        return App::instance()->site()->find($id);
     }
 }
 
@@ -441,18 +435,17 @@ if (Helpers::hasOverride('pages') === false) { // @codeCoverageIgnore
      * Helper to build page collections
      *
      * @param string|array ...$id
-     * @return \Kirby\Cms\Page|\Kirby\Cms\Pages|null
-     * @todo return only Pages|null in 3.7.0, wrap in Pages for single passed id
+     * @return \Kirby\Cms\Pages|null
      */
     function pages(...$id)
     {
-        if (count($id) === 1 && is_array($id[0]) === false) {
-            // @codeCoverageIgnoreStart
-            Helpers::deprecated('Passing a single id to the `pages()` helper will return a Kirby\Cms\Pages collection with a single element instead of the single Kirby\Cms\Page object itself - starting in 3.7.0.');
-            // @codeCoverageIgnoreEnd
+        $pages = App::instance()->site()->find(...$id);
+
+        if (is_a($pages, 'Kirby\Cms\Page') === false) {
+            $pages = new Pages([$pages]);
         }
 
-        return App::instance()->site()->find(...$id);
+        return $pages;
     }
 }
 
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -414,13 +414,13 @@ if (Helpers::hasOverride('option') === false) { // @codeCoverageIgnore
 
 if (Helpers::hasOverride('page') === false) { // @codeCoverageIgnore
     /**
-     * Fetches a single page or multiple pages by
-     * id or the current page when no id is specified
+     * Fetches a single page by id or
+     * the current page when no id is specified
      *
-     * @param string $id
+     * @param string|null $id
      * @return \Kirby\Cms\Page|null
      */
-    function page($id)
+    function page($id = null)
     {
         if (empty($id) === true) {
             return App::instance()->site()->page();
@@ -432,7 +432,7 @@ if (Helpers::hasOverride('page') === false) { // @codeCoverageIgnore
 
 if (Helpers::hasOverride('pages') === false) { // @codeCoverageIgnore
     /**
-     * Helper to build page collections
+     * Helper to build pages collection
      *
      * @param string|array ...$id
      * @return \Kirby\Cms\Pages|null
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -439,13 +439,15 @@ if (Helpers::hasOverride('pages') === false) { // @codeCoverageIgnore
      */
     function pages(...$id)
     {
-        $pages = App::instance()->site()->find(...$id);
-
-        if (is_a($pages, 'Kirby\Cms\Page') === false) {
-            $pages = new Pages([$pages]);
+        // ensure that a list of string arguments and an array
+        // as the first argument are treated the same
+        if (count($id) === 1 && is_array($id[0]) === true) {
+            $id = $id[0];
         }
 
-        return $pages;
+        // always passes $id an array; ensures we get a
+        // collection even if only one ID is passed
+        return App::instance()->site()->find($id);
     }
 }
 
--- a/config/sections/mixins/headline.php
+++ b/config/sections/mixins/headline.php
@@ -10,9 +10,10 @@ return [
          * @todo remove in 3.9.0
          */
         'headline' => function ($headline = null) {
-            if ($headline !== null) {
-                Helpers::deprecated('`headline` prop for sections has been deprecated and will be removed in Kirby 3.9.0. Use `label` instead.');
-            }
+            // TODO: add deprecation notive in 3.8.0
+            // if ($headline !== null) {
+            //     Helpers::deprecated('`headline` prop for sections has been deprecated and will be removed in Kirby 3.9.0. Use `label` instead.');
+            // }
 
             return I18n::translate($headline, $headline);
         },
--- a/config/sections/mixins/headline.php
+++ b/config/sections/mixins/headline.php
@@ -1,18 +1,25 @@
 <?php
 
+use Kirby\Cms\Helpers;
 use Kirby\Toolkit\I18n;
 
 return [
     'props' => [
         /**
          * The headline for the section. This can be a simple string or a template with additional info from the parent page.
-         * @todo deprecate in 3.7
+         * @todo remove in 3.9.0
          */
         'headline' => function ($headline = null) {
+            if ($headline !== null) {
+                Helpers::deprecated('`headline` prop for sections has been deprecated and will be removed in Kirby 3.9.0. Use `label` instead.');
+            }
+
             return I18n::translate($headline, $headline);
         },
         /**
-         * label is the new official replacement for headline
+         * The label for the section. This can be a simple string or
+         * a template with additional info from the parent page.
+         * Replaces the `headline` prop.
          */
         'label' => function ($label = null) {
             return I18n::translate($label, $label);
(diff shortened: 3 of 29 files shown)


config/sections/mixins/headline.php
```diff
@@ -1,18 +1,25 @@
...
 return [
     'props' => [
         /**
          * The headline for the section. This can be a simple string or a template with additional info from the parent page.
-         * @todo deprecate in 3.7
+         * @todo remove in 3.9.0
          */
         'headline' => function ($headline = null) {
+            if ($headline !== null) {
+                Helpers::deprecated('`headline` prop for sections has been deprecated and will be removed in Kirby 3.9.0. Use `label` instead.');
```
distantnative: @texnixe Since we just only made the changes to the kits and docs. Maybe wait with the warning until 3.8.0`? What do you think?

src/Cms/App.php
```diff
@@ -893,20 +893,21 @@ public function kirbytags(string $text = null, array $data = []): string
...
-        $options = A::merge([
-            'markdown' => [
-                'inline' => $inline
-            ]
-        ], $data);
+        // warning for deprecated fourth parameter
+        if ($inline === null) {
+            $inline = false;
+        } else {
+            Helpers::deprecated('markdown component: the $inline parameter is deprecated and will be removed in Kirby 3.8.0. Use $options[\'inline\'] instead.');
```
distantnative: Note to myself: needs to say "kirbytext" not "markdown"

lukasbestle:
@distantnative Before we review the changes, could you please take a look at the failing PHP and E2E tests?

config/sections/mixins/headline.php
distantnative: Our sandbox actually is still using it due to the recent additions to the kits only. So really better to wait with the deprecation notice until 3.8.0.

afbora:
Missing changes in the PR:
https://github.com/getkirby/kirby/blob/fix/remove-deprecated/src/Cms/Permissions.php#L162-L166

distantnative:
@afbora I think it's better to extend this one for one more major release, no harm.

distantnative:
@lukasbestle I think everything is good now. The filling code coverage can be ignored as far as I see the detail report.

config/components.php
```diff
@@ -140,13 +141,21 @@
...
         static $config;
 
+        // warning for deprecated fourth parameter
+        if ($inline === null) {
+            $inline = false;
+        } elseif (isset($options['inline']) === false) {
+            // @codeCoverageIgnoreStart
+            Helpers::deprecated('markdown component: the $inline parameter is deprecated and will be removed in Kirby 3.8.0. Use $options[\'inline\'] instead.');
+            // @codeCoverageIgnoreEnd
+        }
```
afbora, line 157: If I understood correctly;

`$options['inline']` is required? I would expect `$inline !== null` check (or simply `else`) for deprecation. If it is not `null`, it means that the 4th parameter is actively being used.

src/Cms/File.php
```diff
@@ -511,22 +506,13 @@ protected function setFilename(string $filename)
     }
 
     /**
-     * Sets the parent model object;
-     * this property is required for `File::create()` and
-     * will be generally required starting with Kirby 3.7.0
+     * Sets the parent model object
      *
      * @param \Kirby\Cms\Model|null $parent
```
afbora:
```suggestion
     * @param \Kirby\Cms\Model $parent
```

src/Toolkit/Str.php
```diff
@@ -1180,36 +1180,19 @@ public static function substr(string $string = null, int $start = 0, int $length
...
-            is_string($fallback) === true ||
-            $start !== '{{' ||
-            $end !== '}}'
-        ) {
-            Helpers::deprecated('Str::template(): The $fallback, $start and $end parameters have been deprecated. Please pass an array to the $options parameter instead with `fallback`, `start` or `end` keys: Str::template($string, $data, $options)');
-        }
-        // @codeCoverageIgnoreEnd
-
-        $options  = $fallback;
         $fallback = is_string($options) === true ? $options : ($options['fallback'] ?? null);
```
afbora: `$options` can't be string.

src/Toolkit/Str.php
```diff
@@ -1180,36 +1180,19 @@ public static function substr(string $string = null, int $start = 0, int $length
...
-     *                                    - callback: to be able to handle each matching result
-     *                                    - start: start placeholder
-     *                                    - end: end placeholder
-     *                                    A simple fallback string is supported for compatibility (but deprecated).
-     * @param string $start Placeholder start characters (deprecated)
-     * @param string $end Placeholder end characters (deprecated)
-     *
-     * @todo Remove `$start` and `$end` parameters, rename `$fallback` to `$options` and only support `array` type for `$options` in 3.7.0
-     *
+     * @param string|array|null $options An options array that contains:
```
afbora: According to `@todo` comment, `$options` supports only `array` type 🤔

config/components.php
distantnative: We need to continue passing $inline as 4th parameter ourselves in App::markdown() -- otherwise we could break custom components already.

In App::markdown() we pass it both as 4th parameter and as array key. That way in our default component we can throw a warning if 4th parameter is passed but not included as array key.

afbora: I got it 👍

config/helpers.php
```diff
@@ -413,46 +414,38 @@ function option(string $key, $default = null)
...
+     * the current page when no id is specified
      *
-     * @param string|array ...$id
-     * @return \Kirby\Cms\Page|\Kirby\Cms\Pages|null
-     * @todo reduce to one parameter in 3.7.0 (also change return and return type)
+     * @param string|null $id
+     * @return \Kirby\Cms\Page|null
      */
-    function page(...$id)
+    function page($id = null)
```
lukasbestle: Since `$site->find()` also only accepts strings:

```suggestion
    function page(?string $id = null)
```

config/helpers.php
```diff
@@ -413,46 +414,38 @@ function option(string $key, $default = null)
...
-     * @todo return only Pages|null in 3.7.0, wrap in Pages for single passed id
+     * @return \Kirby\Cms\Pages|null
      */
     function pages(...$id)
     {
-        if (count($id) === 1 && is_array($id[0]) === false) {
-            // @codeCoverageIgnoreStart
-            Helpers::deprecated('Passing a single id to the `pages()` helper will return a Kirby\Cms\Pages collection with a single element instead of the single Kirby\Cms\Page object itself - starting in 3.7.0.');
-            // @codeCoverageIgnoreEnd
+        $pages = App::instance()->site()->find(...$id);
```
lukasbestle: I think the following should have the same result without the manual creation of the `Pages` instance:

```php
// ensure that a list of string arguments and an array
// as the first argument are treated the same
if (count($id) === 1 && is_array($id[0]) === true) {
    $id = $id[0];
}

// always pass an array to ensure we get a
// collection even if only one ID is passed
return App::instance()->site()->find($id);
```

src/Cms/App.php
```diff
@@ -892,20 +892,23 @@ public function kirbytags(string $text = null, array $data = []): string
...
-            'markdown' => [
-                'inline' => $inline
-            ]
-        ], $data);
+        // warning for deprecated fourth parameter
+        // @codeCoverageIgnoreStart
+        if ($inline === null) {
+            $inline = false;
+        } else {
+            Helpers::deprecated('Cms\App::kribytext(): the $inline parameter is deprecated and will be removed in Kirby 3.8.0. Use $options[\'inline\'] instead.');
```
```suggestion
            Helpers::deprecated('Cms\App::kirbytext(): the $inline parameter is deprecated and will be removed in Kirby 3.8.0. Use $options[\'markdown\'][\'inline\'] instead.');
```

src/Http/Uri.php
```diff
@@ -330,7 +330,7 @@ public function idn()
      * or any other executed script.
      *
      * @param array $props
-     * @return string
+     * @return string|static
```
lukasbestle: It never actually returns string though, does it?

```suggestion
     * @return static
```

src/Toolkit/Xml.php
```diff
@@ -421,8 +421,7 @@ public static function value($value): ?string
             return $value;
         }
 
-        // TODO: in 3.7.0 use ENT_NOQUOTES | ENT_XML1 instead
-        $encoded = htmlentities($value, ENT_COMPAT);
+        $encoded = htmlentities($value, ENT_QUOTES | ENT_SUBSTITUTE);
```
lukasbestle: Why the different flags compared to the TODO comment?

src/Toolkit/Str.php
lukasbestle:
```suggestion
        $fallback = $options['fallback'] ?? null;
```

src/Toolkit/Str.php
```diff
@@ -1180,36 +1180,19 @@ public static function substr(string $string = null, int $start = 0, int $length
...
+     *                                   - fallback: if a token does not have any matches
+     *                                   - callback: to be able to handle each matching result
+     *                                   - start: start placeholder
+     *                                   - end: end placeholder
      * @return string The filled-in string
      */
-    public static function template(string $string = null, array $data = [], $fallback = null, string $start = '{{', string $end = '}}'): string
+    public static function template(string $string = null, array $data = [], array $options = null): string
```
lukasbestle: Don't think it should be able to be `null`:

```suggestion
    public static function template(string $string = null, array $data = [], array $options = []): string
```

src/Toolkit/Xml.php
distantnative: With `ENT_NOQUOTES | ENT_XML1` unit tests failed. Then I googled again and apparently for `htmlentities`  `ENT_QUOTES | ENT_SUBSTITUTE` is the recommended replacement. So I must've slipped up when writing the comment in the first place.

config/helpers.php
distantnative: think you are missing something, like an `A::wrap()` or something else you had in mind? Cause the comment says `always pass an array` - but `$id` could still be just a string which would return a single page, wouldn't it?

lukasbestle: No, since the argument is defined as `...$id` in the function definition, `$id` will always be an array, even if just a single string argument is passed.

distantnative: Ahh, I wasn't aware of that. Then great!

src/Toolkit/Xml.php
distantnative: https://php.watch/versions/8.1/html-entity-default-value-changes

lukasbestle: I traced it back to this discussion: https://github.com/getkirby/kirby/pull/3979#discussion_r763350909

Since the line is used specifically to check if a `CDATA` block is required, we only need to escape characters that are harmful as XML character data outside of CDATA blocks. Quotes are not harmful there (only in attributes), so we don't need to escape them. Whether we use `ENT_SUBSTITUTE` or not doesn't matter as invalid code unit sequences will in any case return a different `$encoded` value. What we do need is `ENT_XML1` as we are encoding specifically for XML output.

So I think our unit tests are wrong and need to be updated. I see two failing ones:

```txt
2) Kirby\Toolkit\XmlTest::testTag
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'<name foo="bar"><![CDATA[Süper Önencœded ßtring]]></name>'
+'<name foo="bar">Süper Önencœded ßtring</name>'

kirby/tests/Toolkit/XmlTest.php:187

3) Kirby\Toolkit\XmlTest::testValue with data set #8 ('töst', '<![CDATA[töst]]>')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'<![CDATA[töst]]>'
+'töst'

kirby/tests/Toolkit/XmlTest.php:205
```

Both these strings don't actually require `CDATA`. If we change the strings to `String with <not> a tag & some text`, the tests will pass again.

distantnative: I'm rather lost with the `XmlTest` test suite - and what can be changed or not without losing an important test case. So I'd rely there on your assessment.

distantnative: This would be one of the failing tests

```
$tag = Xml::tag('name', 'Süper Önencœded ßtring', ['foo' => 'bar']);
$this->assertSame('<name foo="bar"><![CDATA[Süper Önencœded ßtring]]></name>', $tag);
```

Not sure I completely understand your suggestion how to change it when using `ENT_NOQUOTES | ENT_XML1`.

lukasbestle:
```php
$tag = Xml::tag('name', 'String with <not> a tag & some text', ['foo' => 'bar']);
$this->assertSame('<name foo="bar"><![CDATA[String with <not> a tag & some text]]></name>', $tag);
```

lukasbestle: And `XmlTest::valueProvider()` would be:

```php
    public function valueProvider()
    {
        return [
            [true, 'true'],
            [false, 'false'],
            [1, '1'],
            [null, null],
            ['', null],
            ['<![CDATA[test]]>', '<![CDATA[test]]>'],
            ['<![CDATA[String with <not> a tag & some text]]>', '<![CDATA[String with <not> a tag & some text]]>'],
            ['test', 'test'],
            ['String with <not> a tag & some text', '<![CDATA[String with <not> a tag & some text]]>'],
            ['This is a <![CDATA[test]]> with CDATA', '<![CDATA[This is a <![CDATA[test]]]]><![CDATA[> with CDATA]]>'],
            ['te]]>st', '<![CDATA[te]]]]><![CDATA[>st]]>'],
            ['tö]]>st', '<![CDATA[tö]]]]><![CDATA[>st]]>']
        ];
    }
```

distantnative: Implemented, thanks!

config/helpers.php
```diff
@@ -3,6 +3,7 @@
 use Kirby\Cms\App;
 use Kirby\Cms\Helpers;
 use Kirby\Cms\Html;
+use Kirby\Cms\Pages;
```
lukasbestle: Can be removed now I think:

```suggestion
```

distantnative: Ofc, dammit - always the little things :D

lukasbestle (approved):
Wohoo! 🎉

Maybe we convert the release notes to a table in some places like we did in the 3.6 release notes, so it's easier to see what has been deprecated/removed and replaced by what.

PR #4337, opened 2022-06-05
`Str::afterStart()` and `beforeEnd()`; used to make `Environment` code simpler

lukasbestle:
### Features

- New `Str::afterStart()` and `Str::beforeEnd()` methods to remove a substring only from the start or end of a string

## Ready?

### For review team

### DIFF
--- a/src/Http/Environment.php
+++ b/src/Http/Environment.php
@@ -789,10 +789,7 @@ class Environment
         $requestUri = str_replace('//', '/', $requestUri);
 
         // remove the script path only from the beginning of the URI
-        $scriptPath = $this->scriptPath();
-        if ($scriptPath && Str::position($requestUri, $scriptPath) === 0) {
-            $requestUri = Str::substr($requestUri, Str::length($scriptPath));
-        }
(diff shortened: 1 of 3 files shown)


distantnative:
Maybe rather as variations of the existing e.g. `Str::after` method? Via a new parameter

lukasbestle:
They already have three arguments. I feel it could become unwieldy.

Also the new methods need to behave a bit differently in these two cases:

- **Empty needle:** `after` and `before` throw an exception. But with `afterStart` and `beforeEnd` the original string is returned as they don't have to remove anything from the string.
- **Needle that's not found:** `after` and `before` return an empty string. `afterStart` and `beforeEnd` return the original string because they remove the substring only if it's found. The rest of the string needs to stay intact.

Mixing the two behaviors into the same methods doesn't feel right to me.

distantnative:
Ok, make sense. I added the main commit on a separate PR to `develop`: https://github.com/getkirby/kirby/pull/4339
I think it makes sense to first merge this separately from the server/environment PR and not bundle it with it.

lukasbestle:
Good point. I've merged the other PR and will rebase the Environment PR against `develop`.

PR #4338, opened 2022-06-06
Add new public static method `studly` and `camel` for `Str` class

Awilum:
### Features

- New `Str::camel()` and `Str::studly()` methods for case-altering of strings:

```php
echo Str::camel('foo_bar'); //=> fooBar
echo Str::studly('foo_bar'); //=> FooBar
```

## Ready?

### For review team


distantnative:
I think this casing is more commonly known as pascal case? So maybe rather go with hat name?

But love this PR. I think should add `Str::camel` as well (calling your new method plus an extra `lcfirst()`).

Awilum:
@distantnative 

**[StudlyCaps](https://en.wiktionary.org/wiki/StudlyCaps)**, also known as **PascalCase**, implies that the first capital of each subword is capitalized. **camelCase** implies, like a **camel**, that the humps are in the middle, therefore the first letter is not capitalized.

StudlyCaps or PascalCase => **studly**
camelCase => **camel**
snake_case => **snake**
kebab-case => **kebab**

Awilum:
@distantnative if this PR is fine, I will push another PR with new method `camel`

distantnative:
@Awilum my point was not the difference between pascal and camel case, but that I think your new method should be called `Str::pascal()` - from googling this seems to be the more commonly used name instead of studly.

Adding `Str::camel` in addition would be great! :)

lukasbestle:
@distantnative [PSR-1](https://www.php-fig.org/psr/psr-1/) calls them `StudlyCaps`, so I think the naming is fine.

@Awilum Feel free to push another commit for `Str::camel()` to this PR so we can merge both methods together.

src/Toolkit/Str.php
```diff
@@ -1018,6 +1018,18 @@ public static function snake(string $value = null, string $delimiter = '_'): str
         return $value;
     }
 
+    /**
+     * Convert a string to studly caps case.
+     *
+     * @param string $value The string to convert.
+     * 
+     * @return string
```
```suggestion
     * Converts a string to studly caps case
     *
     * @param string $value The string to convert
     * @return string
```

src/Toolkit/Str.php
```diff
@@ -1018,6 +1018,18 @@ public static function snake(string $value = null, string $delimiter = '_'): str
...
+     */
+    public static function studly(string $value): string
```
lukasbestle: For consistency with the other case-altering `Str` methods, the method signature should be:

```suggestion
    public static function studly(string $value = null): string
```

We do this for Postel's law so that an accidental `null` won't cause a PHP error and will be treated like an empty string. So for `null` the return value should be an empty string.
However the documented signature should still only support a string, so the docblock and the `string` type are unchanged.

Awilum:
@distantnative @lukasbestle `studly` and `camel` pushed

tests/Toolkit/StrTest.php
```diff
@@ -892,6 +892,30 @@ public function testSnake()
...
+    /**
+     * @covers ::camel
+     */
+    public function testCamel()
+    {
+        $string = 'foo_bar';
+        $this->assertSame('fooBar', Str::camel($string));
+
+        $string = 'fòô_bàř';
+        $this->assertSame('fòôBàř', Str::camel($string));
```
distantnative: Sorry, one last thing: Can we add tests where the input starts upper case? So we can actually make sure the first character always gets lowercased.

lukasbestle (commented):
Could you please also move the `camel` method and `testCamel` test method so it's in alphabetical order (after `between` and before `contains`)?

Awilum:
@distantnative @lukasbestle Done

lukasbestle (approved):

distantnative:
Thanks a lot @Awilum

PR #4340, opened 2022-06-06
Add new public static method `increment` for `Str` class

Awilum:
### Features
* New `Str::increment()` method for appending a number `-1` to a string or incrementing the ending number to allow `-2`, `-3`, etc.

```php
echo Str::increment('Page-1'); //=> Page-2
echo Str::increment('Page', '_'); //=> Page_1
```

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -540,7 +540,7 @@ class Str
     {
         preg_match('/(.+)' . $separator . '([0-9]+)$/', $string, $match);
 
-        return isset($match[2]) ? $match[1] . $separator . ((int) $match[2] + 1) : $string . $separator . $first;
+        return isset($match[2]) ? $match[1] . $separator . ((int)$match[2] + 1) : $string . $separator . $first;
     }
 
     /**
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -538,7 +538,7 @@ class Str
      */
     public static function increment(string $string, int $first = 1, string $separator = '_'): string
     {
-        preg_match('/(.+)' . $separator . '([0-9]+)$/', $string, $match);
+        preg_match('/(.+)' . preg_quote($separator, '/') . '([0-9]+)$/', $string, $match);
 
         return isset($match[2]) ? $match[1] . $separator . ((int)$match[2] + 1) : $string . $separator . $first;
     }
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -529,7 +529,7 @@ class Str
     }
 
     /**
-     * Add's _1 to a string or increment the ending number to allow _2, _3, etc
+     * Adds `-1` to a string or increments the ending number to allow `-2`, `-3`, etc.
      *
      * @param string $string The string to increment
      * @param string $separator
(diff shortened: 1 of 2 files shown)


src/Toolkit/Str.php
```diff
@@ -528,6 +528,21 @@ public static function from(string $string, string $needle, bool $caseInsensitiv
...
+     * Add's _1 to a string or increment the ending number to allow _2, _3, etc
+     *
+     * @param string $string The string to increment
+     * @param int $first Start with
+     * @param string $separator Separator
+     * @return string
+     */
+    public static function increment(string $string, int $first = 1, string $separator = '_'): string
+    {
+        preg_match('/(.+)' . $separator . '([0-9]+)$/', $string, $match);
```
lukasbestle: To ensure that any separator will work, even if it's a special character in RegEx:

```suggestion
        preg_match('/(.+)' . preg_quote($separator, '/') . '([0-9]+)$/', $string, $match);
```

src/Toolkit/Str.php
```diff
@@ -528,6 +528,21 @@ public static function from(string $string, string $needle, bool $caseInsensitiv
...
+
+        return isset($match[2]) ? $match[1] . $separator . ((int)$match[2] + 1) : $string . $separator . $first;
```
lukasbestle: For better readability:

```suggestion
        if (isset($match[2]) === true) {
            // increment the existing ending number
            return $match[1] . $separator . ((int)$match[2] + 1);
        }

        return $string . $separator . $first;
```

src/Toolkit/Str.php
```diff
@@ -528,6 +528,21 @@ public static function from(string $string, string $needle, bool $caseInsensitiv
...
 
+    /**
+     * Add's _1 to a string or increment the ending number to allow _2, _3, etc
+     *
+     * @param string $string The string to increment
+     * @param int $first Start with
+     * @param string $separator Separator
+     * @return string
+     */
+    public static function increment(string $string, int $first = 1, string $separator = '_'): string
```
lukasbestle: I suggest we swap the arguments as the `$separator` may be changed more often than the starting number. Maybe we should also default to `-` as kebab case for identifiers is more commonly used in Kirby.

```suggestion
     * @param string $separator
     * @param int $first Starting number
     * @return string
     */
    public static function increment(string $string, string $separator = '-', int $first = 1): string
```

tests/Toolkit/StrTest.php
```diff
@@ -425,6 +425,39 @@ public function testFromWithEmptyNeedle()
         Str::from('test', '');
     }
 
+    /**
+     * @covers ::increment
+     */
+    public function testIncrement()
```
lukasbestle, line 431: Could you please also add test cases for the following:

- Input string already contains numbers, but with a different separator; in these cases the method should add the separator and the `$first` number to the end and keep the existing number
- Two cases that show that the `$first` argument is ignored when there's already a number in the input. One case e.g. with an input of `Pöst_5` and `$first = 10` and one with `Pöst_15` and `$first = 10`.

lukasbestle (commented):
Thank you, that's also a very useful method. A few comments:

lukasbestle (approved):

lukasbestle:
@Awilum Thank you very much! 👍

PR #4342, opened 2022-06-08
Add new `hexColor` validator for `V` class

Awilum:
### Features
* New `V::hexColor()` to check for valid hex color

```php
// V::hexColor()
if (V::hexColor("#000000")) {
  // do something...
}
```

## Ready?

### For review team


tests/Toolkit/VTest.php
```diff
@@ -263,6 +263,25 @@ public function testFilename()
...
+        $this->assertTrue(V::hexColor('#ffffff'));
+        $this->assertTrue(V::hexColor('#fff'));
+        $this->assertTrue(V::hexColor('fff'));
+        $this->assertFalse(V::hexColor('42'));
+        $this->assertFalse(V::hexColor(42));
+        $this->assertFalse(V::hexColor([]));
+        $this->assertFalse(V::hexColor(null));
+        $this->assertFalse(V::hexColor(false));
+        $this->assertFalse(V::hexColor(new \stdclass()));
+        $this->assertFalse(V::hexColor(''));
```
afbora, line 282:
```suggestion
        $this->assertFalse(V::hexColor(''));
        $this->assertFalse(V::hexColor('#'));
        $this->assertFalse(V::hexColor('#ff'));
        $this->assertFalse(V::hexColor('##fff'));
        $this->assertFalse(V::hexColor('#???'));
```

src/Toolkit/V.php
```diff
@@ -422,6 +422,17 @@ public static function __callStatic(string $method, array $arguments): bool
...
 
+    /**
+     * Checks for a valid hex color
+     */
+    'hexColor' => function ($value): bool {
+        if (!is_string($value) || $value === '') {
+            return false;
+        }
+
+        return (bool)mb_ereg_match('^#?([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$', $value);
```
afbora:
```suggestion
        return mb_ereg_match('^#?([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$', $value);
```

afbora: I just wonder that should hash `#` be optional?

lukasbestle: When the validator is used for a field, I agree from the user perspective. Users may want to type just the hex digits and make the site normalize the value to a full color value with the prepended hash sign.

From the developer perspective an optional hash sign would complicate the implementation though.

What we can do: Introduce a second argument to the validator like this:

```php
V::hexColor($value, ?bool $hash = true)
```

The behavior would be:

- `true`: Hash is required
- `false`: Hash must *not* be used
- `null`: Hash is optional (with or without doesn't matter)

I agree with @afbora that the default should be to enforce the hash.

texnixe: I'm wondering the same. The main problem I see with this is that the optional `#` wouldn't allow me to enforce one way of entering the hex code.

src/Toolkit/V.php
```diff
@@ -422,6 +422,17 @@ public static function __callStatic(string $method, array $arguments): bool
...
 
+    /**
+     * Checks for a valid hex color
+     */
+    'hexColor' => function ($value): bool {
+        if (!is_string($value) || $value === '') {
+            return false;
+        }
+
+        return mb_ereg_match('^#?([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$', $value);
```
lukasbestle: Two comments on the logic:

- What about 4-digit and 8-digit hex colors with alpha transparency (RGBA)?
- Since the check doesn't involve multibyte characters, I think `preg_match()` should suffice. And since it's only about hex digits, it can be implemented even simpler with:

```php
return ctype_xdigit($value) === true && in_array(strlen($value), [3, 4, 6, 8]) === true;
```

(ignoring the hash sign, which needs to be checked separately depending on the result of our other discussion)

src/Toolkit/V.php
```diff
@@ -422,6 +422,25 @@ public static function __callStatic(string $method, array $arguments): bool
                V::min($value, 2) === true;
     },
 
+    /**
+     * Checks for a valid hex color
+     */
+    'hexColor' => function ($value, ?bool $hash = true): bool {
+        if (!is_string($value) || $value === '') {
+            return false;
+        }
```
afbora, line 431: I'm asking stupid question (not big deal) because I don't know very well. Does it make sense to send the hex code as an `integer`? Like `000` or `22446688` as integer, not string. For example: `22446688` is valid hex code and `ctype_xdigit(22446688)` returns `true`.

PR #4343, opened 2022-06-08
New Toggles field

bastianallgeier:
### New Toggles field

Thanks to @fabianmichael for offering his fantastic multi toggle plugin as foundation for this new field: https://github.com/fabianmichael/kirby-multi-toggle-field/

```yaml
textAlign:
  type: toggles
  label: Text alignment
  options:
    - value: left
      text: Left
      icon: text-left
    - value: center
      text: Center
      icon: text-center
    - value: right
      text: Right
      icon: text-right
    - value: justify
      text: Justify
      icon: text-justify
      
  # additional properties
  # labels: false
  # columns: 2
  # reset: false
  # grow: false
```

## Ready?

### For review team


panel/src/components/Forms/Input/TogglesInput.vue
```diff
@@ -0,0 +1,130 @@
...
+
+<script>
+import { autofocus, disabled, id, required } from "@/mixins/props.js";
+import { required as validateRequired } from "vuelidate/lib/validators";
+
+export const props = {
+  mixins: [autofocus, disabled, id, required],
+  props: {
+    columns: Number,
+    equalize: Boolean,
```
afbora: I think, this prop is redundant? Equalized already.

bastianallgeier: That was a remainder of Fabian's field.

afbora (approved):
LGTM and love that ❤️

lukasbestle:
What is the use case for customizing the `reset` prop? Wouldn't it make sense to link the behavior to the `required` prop directly?

afbora:
@lukasbestle Actually we had the same situation with the select field https://github.com/getkirby/kirby/pull/3373

lukasbestle:
Good point. It's certainly a difficult topic.

But from the top of my head I can't think of any reason why a toggle field should be optional but where it should be impossible to make it empty again. Either the field is required (so an empty value can never be valid) or it is optional (so the initial empty value should be achievable even if an option was selected by accident).

Since the `reset` prop was already supported in Fabian's plugin, we should probably still include it. But maybe we can deprecate it in our implementation and wait for user feedback. If there is any use case we missed, we can still clear the deprecation and keep the prop.

fabianmichael:
@lukasbestle Having a separate reset prop can be useful, when blueprints are extended later on. For example, one of my clients had a two-column text block. At some point we realized that it would make sense to have a setting for vertical alignment of the columns, because sometimes the right column would only contain an image of CTA button. If I would have set up the site from scratch, I would just have marked the `align` field as required with a default value. But since existing blocks won’t be updated automatically and I did not feel like writing a converter script, I just decided to make this field optional, so editors won’t have to deal with an error message when editing an existing page.

Sure, updating the content would have been the much more elegant option. I’ll leave it to you, whether it is necessary to support such a workaround in a field.

lukasbestle:
Thanks for the explanation, that makes sense. So it's basically a "required field that is currently optional".

I'd both be fine with keeping the `reset` prop and deprecating it. What do the others think?

afbora:
I would prefer to continue with `reset` since we have a similar option in the select field (`empty` prop) and Fabian's plugin also has this option.

PR #4344, opened 2022-06-09
Forward extension of accept property to browser

rasteiner:
Previously the `extension` part of the `accept` property in file blueprints was ignored when creating the [HTML `accept` attribute ](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept). 
For file types that are unknown to Kirby this resulted in a `accept="*"` attribute.

For context: https://forum.getkirby.com/t/new-file-blueprint-for-gpx-files/25618

## This PR …
... makes it so that a correct string, combining extensions and mime types in a comma separated list is used instead.
It does this by adding a new `acceptAttribute` method to the `FileBlueprint` class, and replaces the use of `acceptMime` in the files section and files field (currently `upload` mixin)  with that.

## Fixes
 - extension part of accept property is no longer ignored 

### Breaking changes
none

## Ready?

### For review team

### DIFF
--- a/config/fields/mixins/upload.php
+++ b/config/fields/mixins/upload.php
@@ -31,7 +31,7 @@ return [
                     'template' => $template
                 ]);
 
-                $uploads['accept'] = $file->blueprint()->acceptMime();
+                $uploads['accept'] = $file->blueprint()->acceptAttribute();
             } else {
                 $uploads['accept'] = '*';
             }
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -73,7 +73,7 @@ return [
                     'template' => $this->template
                 ]);
 
-                return $file->blueprint()->acceptMime();
+                return $file->blueprint()->acceptAttribute();
             }
 
             return null;
--- a/src/Cms/FileBlueprint.php
+++ b/src/Cms/FileBlueprint.php
@@ -112,6 +112,27 @@ class FileBlueprint extends Blueprint
         // no restrictions, accept everything
         return '*';
     }
+    
+    /**
+     * Returns the list of all accepted file extensions and mime types in a comma separated string.
+     * Compatible with the `accept` attribute of HTML input elements.
+     *
+     * @return string
+     */
+    public function acceptAttribute(): string
+    {
+        $accept = $this->accept();
+
+        $extensionArray = A::wrap($accept['extension'] ?? []);
+        $mimeArray = A::wrap($accept['mime'] ?? []);
+
+        return A::join([
+            ...A::map($extensionArray, function ($extension) {
+                return '.' . $extension;
+            }),
+            ...$mimeArray
+        ]);
+    }
 
     /**
      * @param mixed $accept
--- a/src/Cms/FileBlueprint.php
+++ b/src/Cms/FileBlueprint.php
@@ -117,6 +117,7 @@ class FileBlueprint extends Blueprint
     /**
      * Returns the list of all accepted file extensions and mime types in a comma separated string.
      * Compatible with the `accept` attribute of HTML input elements.
+     * When no restrictions are defined, the string `*` is returned.
      *
      * @return string
      */
@@ -130,7 +131,7 @@ class FileBlueprint extends Blueprint
         return A::join(array_merge(
             A::map($extensionArray, fn ($ext) => ".$ext"),
             $mimeArray
-        ));
+        )) ?: '*';
     }
 
     /**
--- a/src/Cms/FileBlueprint.php
+++ b/src/Cms/FileBlueprint.php
@@ -127,12 +127,10 @@ class FileBlueprint extends Blueprint
         $extensionArray = A::wrap($accept['extension'] ?? []);
         $mimeArray = A::wrap($accept['mime'] ?? []);
 
-        return A::join([
-            ...A::map($extensionArray, function ($extension) {
-                return '.' . $extension;
-            }),
-            ...$mimeArray
-        ]);
+        return A::join(array_merge(
+            A::map($extensionArray, fn ($ext) => ".$ext"),
+            $mimeArray
+        ));
     }
 
     /**
--- a/src/Cms/FileBlueprint.php
+++ b/src/Cms/FileBlueprint.php
@@ -3,6 +3,7 @@
 namespace Kirby\Cms;
 
 use Kirby\Filesystem\F;
+use Kirby\Toolkit\A;
 use Kirby\Toolkit\Str;
 
 /**


lukasbestle:
Thank you for your PR. 👍

The current logic comes from this PR: https://github.com/getkirby/kirby/pull/2973
The expected behavior is IMO that the defined MIME types are used 1:1 for the `accept` attribute. E.g. in the example from the forum, it should be `accept="text/xml, application/gpx, application/gpx+xml"`. It's certainly a bug when `accept="*"` is generated instead.

But to be honest I think we should fix the bug directly in the current implementation instead of replacing it with an entirely different implementation in a separate method. The sole purpose of the `acceptMime()` method is to handle the `accept` attribute generation. So with your PR it would become dead code.

If I remember correctly, we explicitly decided not to include the extensions in the attribute, that's why we have this code comment:

https://github.com/getkirby/kirby/blob/005fc78fcc0e7f99f61e6973efa8ea8ec85741d9/src/Cms/FileBlueprint.php#L77-L80

It's so long ago that I can't remember why we decided that from the top of my head, but before we change the behavior we should track this down.

rasteiner:
> E.g. in the example from the forum, it should be `accept="text/xml, application/gpx, application/gpx+xml"`.

the problem with `text/xml, application/gpx, application/gpx+xml` is that the browser file open dialog will show many xml related files (xml, xbl, xsl, etc...), but **not** .gpx files, because the browser doesn't know that mime code.

The only way of letting the browser select ".gpx" files is to have `accept=".gpx"`

You can test this here: https://codepen.io/rasteiner/pen/BaYGLyw

> The sole purpose of the acceptMime() method is to handle the accept attribute generation. So with your PR it would become dead code.

In that case I'd argue that the function should be deprecated since the accept attribute IMHO really needs the extension part and at that point the function no longer generates just a "mime type". 

About `accept="*"`; currently the mime part is taken 1:1 (that works fine: it does not rewrite an explicit mime type to `*`). It ends up being `*` when the only specified part is an extension that kirby doesn't know. But neither Kirby, nor the browser, can be expected to know every extension, also because my app might have some proprietary file formats, and I should be able to upload `.meow` files if I need to.

rasteiner:
https://github.com/getkirby/kirby/blob/005fc78fcc0e7f99f61e6973efa8ea8ec85741d9/src/Cms/FileBlueprint.php#L77-L80

Doesn't this just mean that it should only fallback to the extension (or type) restriction when no mime type was specified? (I mean I can change the PR to conform to this, though I don't really understand the reasoning). 
Even if this has some forgotten reason, the problem with the current implementation is that it **does not** fallback to the extension when the extension is unknown to Kirby, but falls back to the all allowing `*`. Honestly however a `.xml` restriction is *more restrictive* than `text/xml`, so I kinda find it hard to imagine a case in which `text/xml` is somehow better than `.xml`  other than to conform to the function name...

lukasbestle:
I get your points, but I have the feeling that there might have been a security reason for this like that we wanted to avoid leaking the accepted extensions or something. I will look deeper into this.

lukasbestle:
Quick status update: Haven't gotten to this so far unfortunately, but I haven't forgotten about it.

lukasbestle:
Sorry for the long delay. I finally looked at this again now.

In our [docs](https://getkirby.com/docs/reference/panel/blueprints/file#accept__extended) I found this:

> If you don't set the `mime` option, the valid MIME types for the frontend upload input are auto-detected from the configured `extension` or `type` options. You can however override this behavior by setting the `mime` option as well, for example to define MIME types that are compatible with all browsers while still protecting the uploads based on the extension and/or file type:
> 
> ```yaml
> accept:
>    mime: image/*
>    type: image
> ```
> 
> In this example, Kirby will generate a file upload input with an `accept="image/*"` attribute and will then validate both the MIME type and the detected file type on the server after the file is uploaded.

Our current implementation always converts everything into a MIME type, even if you just pass an extension. This was with the expectation that MIME types are the most universal value for the `accept` attribute. I think there were browser and/or OS compatibility issues when I tested extensions back then. I don't know if that's still the case.

But of course with more obscure file types like GPX, specifying the MIME type in the `accept` attribute isn't very useful. Including the extension is much more useful here (if it works cross-browser).

**However** we cannot just merge the accepted MIME types and extensions into one list. All the accept options in the file blueprint are cumulative. So if you define both a `mime` and an `extension`, the uploaded file will only be allowed if it matches both. If you define a `type`, it must also match that (but a `type` is basically just an alias for a list of extensions).

Unfortunately, there is no way to tell the browser to accept a file only if it matches all rules (cumulative `AND` logic). The browser always reads the list with `OR` logic.

What we could do is the following:

- Instead of converting everything to MIME type (like we do now), we could convert everything to extension.
- If the blueprint includes a MIME wildcard (like `image/*`), we build a list of all extensions (known to Kirby) that match this wildcard.
- Then we can again find the intersection between the generated extension lists for MIMEs, extensions and types and we get a list of extensions that are definitely allowed.
- All of this would be handled by the new `acceptAttribute()` method. We would deprecate the old `acceptMime()` method (and later remove it).

An issue (and probably another reason why I went with MIME types) is that a list of allowed extensions can get very very long.

So we probably need a way for the developer to override all of this logic in the blueprint with a custom `accept` attribute value.

rasteiner:
I think either sending a list of extensions or expanding to one on the client side would produce the most user friendly interaction. 

What I mean by this...

If I as a developer write:
```yaml
accept:
   mime: image/*
   type: image
```
the browser currently receives `image/*` as indication and expands that to **its own** list of extensions. That list is not guaranteed to exactly match the list `type: image` produces in kirby (and not even the one that a mime of `image/*` would produce in kirby). It's therefore possible that the dialog allows me to upload files that are then rejected by Kirby.
This would also be solved with a list of extensions. 

So, yay for that 👍

--- 

About being "vertically `AND` and horizontally `OR`"... Ok, I see it's written like that in the docs. I just wonder if it's really what you want. 
Consider the usefulness of being able to write:
```yaml
accept:
   mime: image/*
   type: image
```
and having the `image/*` be intersected with an "unknown" list of extensions coming from `type: image`...

VS being able to write:
```yaml
accept:
   extension: idml
   type: document
```
and having `idml` be added to what kirby recognizes as "document". 

But I guess this is a breaking change this PR doesn't need to handle 😅

lukasbestle:
Yeah, that's the problem. We have two possible routes:

## 1. Switch to `OR` logic

We change the behavior of the rules to use `OR` logic (meaning that multiple rules widen the set of possible files instead of restricting it). Then we could combine MIME types and extensions in the same `accept` attribute.

**Issue:** Would be a breaking change. It would be quite difficult to migrate the blueprints of existing sites as the previous logic cannot be easily replicated with the `OR` logic.

But I think in the long term the `OR` logic would make sense as it's much easier to understand and also cleaner to implement.

## 2. Keep the `AND` logic

Alternatively we keep the `AND` logic but convert everything to an extension for the `accept` attribute.  

**Issues:**  

- The attribute value can get incredibly long, which could be bad for performance.
- The resulting set of extensions would still not 100% match the blueprint rules. E.g. if you define a `mime` rule, a user could try to upload a file with an extension that is common for that MIME type, but with contents that don't match that MIME type. This file would then be selectable but would be rejected in the backend.

---

All things considered, I think we should go for route 1, but that means that this PR can only come after the breaking change. The next task would be to find a solution to ease the breaking change, so users can migrate easier.

distantnative:
@lukasbestle let's go with option 1 - seems to me as well the more sensible choice. 

As I understand it, our implementation that uses `AND` logic already came up due to shortcomings by browsers. So having a case where those type + mime and now also extension are added and we have a clash with AND/OR might be quite rare.

lukasbestle:
@distantnative Maybe I don't fully understand what you mean, but I think it's not just about browsers. The rules also take effect on the backend during the actual validation, so changing from `AND` to `OR` logic would certainly be a breaking change.

distantnative:
@lukasbestle Not debating that this is a breaking change, we're only talking about the dimensions of the breaking change and if/how to ease it. From your descriptions above, the feature to restrict by multiple things was introduced because browser support for accept extensions wasn't great. But it sounds like it was already supporting more of a niche requirement. Which then breaking it is likely not affecting too many. 

Or maybe let me ask in a different way: what scenarios that do far used AND could not be implemented in a different way with extensions and OR?

lukasbestle:
I can't think of any situations where a current setup cannot be converted to OR logic (by explicitly defining a more strict set of accept rules in the first place).

However the conversion will need to be done manually and in each case the dev needs to really think about the intended logic. So it's going to be difficult.

distantnative:
I'm still trying to understand why it would be so difficult. Yes, it'll be a breaking change by definition, but would it be really a problem for the vast majority?

But if the browser now works by OR - the worst what could happen if that the browser allows selecting a file and then the backend would deny uploading it. Correct? So it wouldn't allow anything that was forbidden before and it wouldn't prevent anything that was possible before.

On top, my understanding was that more elaborate `accept` conditions are also common for cases where a passing an extension to the browser would actually solve this easier. Then that wouldn't be much work either. But maybe I've misunderstood that part.

lukasbestle:
As I wrote, it's not about the browser but about the backend. If you assume that any file can be sent to the server (which an attacker always can), the backend still needs to check if the uploaded file is accepted. Currently it uses `AND` logic for this, so it is quite strict. Switching to `OR` logic there without migrating the blueprints correctly could cause security issues in the worst case or at least unwanted behavior if suddenly file types can be uploaded that weren't expected.

We could of course keep the `AND` logic on the backend and only use `OR` logic for the browser, but TBH I don't think it makes sense. It would be super confusing and there would be no way to configure a matching set of rules. So IMO if we switch to `OR`, we need to do so also on the backend – with the mentioned migration and security implications.

rasteiner:
Sorry to interrupt you, I keep getting notifications about this PR. 😅

I'd still go for a plain list of extensions. Independently of OR and AND. 
It's the only way to *somewhat* guarantee frontend and backend behave the same. You can then later decide if you want OR or AND logic, and the browser will follow. Checking files by extension also is trivial and arguably stricter than mime types, and you can still validate uploaded files against their supposed mime type.  

The two issues mentioned above are:

> - The attribute value can get incredibly long, which could be bad for performance.

I think this needs an evaluation? On my PC the file select dialog starts lagging a bit after 300 extensions; it's still usable though even with 500 (see [playground](https://codepen.io/rasteiner/pen/PoybOzw?editors=0010), warning: 1000 starts to freeze the browser);

Kirby knows [about 70 extensions](https://github.com/getkirby/kirby/blob/ce4aaace87ceca1f019a9cfc0293a69e662f9648/src/Filesystem/F.php#L28-L111)? That doesn't look like a problematic number. And that number is hit only when a dev decides they want *any kind of content*, so it might as well be `*` with no extensions at all.
If you're worried about transmitting the list from kirby to browser, apart from not being reasonably a problem (any kind of thumbnail weighs more than the complete list of extensions), those mime types (or "kirby types") can also be extended to a list of file extensions on the browser (and this is immediate: you might have noticed the time it took to generate the inputs in the playground - or better: **not** have noticed). 

> - The resulting set of extensions would still not 100% match the blueprint rules. E.g. if you define a mime rule, a user could try to upload a file with an extension that is common for that MIME type, but with contents that don't match that MIME type. This file would then be selectable but would be rejected in the backend.

I believe the fact that the behaviour stays the same as it is now (and also what it would be with the alternative) isn't really a valid argument against going this way. 
I mean the same problem applies to mime types, also because afaik the browser just picks up the mime you give it and expands it to its own list of extensions anyway.
The only way around this would be having the panel check the file contents in the browser, which certainly **won't happen anyway** (as it would be pretty much useless from a security pov). 
In other words, that this doesn't solve an impossible problem with no solution can't really be considered an issue. 
And arguably it actually *matches* the blueprint rules 100%. It's the sanitizer that raises errors, not the blueprint; the error isn't "you uploaded the wrong kind of file", it's "your file isn't valid".

lukasbestle:
Hm, I'm confused. Either there is a misunderstanding or any of us miss something.

> I'd still go for a plain list of extensions. Independently of OR and AND.
> It's the only way to somewhat guarantee frontend and backend behave the same. You can then later decide if you want OR or AND logic, and the browser will follow.

AFAICT, the browser always follows OR logic. We can only give it a uni-dimensional list and it will accept any file that will match any value in that list.

So: The backend needs to convert the blueprint config into a list that the browser can understand. Either it concats all possible file types and converts all of them to extensions (= OR logic) or it throws everything into a big venn diagram and only picks the options that match all configured rules (= AND logic).

I think we need to decide on one of those solutions. It's not possible to separately solve this just with frontend or backend changes.

rasteiner:
> AFAICT, the browser always follows OR logic. We can only give it a uni-dimensional list and it will accept any file that will match any value in that list.

My point is really only about mime types vs extensions. 
Whatever you end up doing with the venn diagram, the best way to communicate the result of this conjunction or disjunction to the browser IMO is a plain list of file extensions, not mime types. Because it gives you greater control. 

Lets look at an example:
```
accept:
   mime: image/*
   type: image
```

just for the sake of example, let's imagine that kirby thinks of:
 - `image/*` -> `.jpg, .jp2, .png, .gif` 
 - `type: image` -> `.jpg, .gif, .tga, .avif, .webp`

OR logic would end up in a list of `.jpg, .jp2, .png, .gif, .tga, .avif, .webp`
AND logic would end up in a list of `.jpg, .gif` 

Both can be expressed to the browser as a plain list of extensions, and you don't really care about the "OR logic of the browser", because it's the only logic that makes sense *there* (no file would ever match 2 extensions at the same time anyway). 

****

I wasn't really talking about whether this can be done solely on backend or frontend (I don't think that's very important). 
However it would allow you to **Keep the AND logic** first (no breaking change, but solve this issue), and in a later point in time decide whether you want to switch to an OR logic and have the breaking change, or not.

lukasbestle:
We are entirely on the same page regarding the first part of your message. What you describe in your example completely matches what I meant.

Your last paragraph is indeed a good point and quite convincing. If we keep the AND logic, we can fix this issue without a breaking change.

distantnative:
Also trying to breathe life into this one again:
I've read the long discussion back and forth. What I gathered:

## `AND` vs. `OR`
Currently, the logic is rather exclusive (AND). Going forward to avoid an immediate breaking change 

```yml
accept:
   extension: jpg
   type: image
```

would need to result in only allowing `.jpg`. If that's ensured, we could add a change to 4.x already.

There seems to be an agreement, that `OR` would be the preferred way as it would be way easier to understand. This would be a breaking change we could add in v5. While it could be hard for some edge cases to recreate their exact restrictions, I'd think for the vast majority of cases it won't be any problem.

## Ensure security 

> The resulting set of extensions would still not 100% match the blueprint rules. E.g. if you define a mime rule, a user could try to upload a file with an extension that is common for that MIME type, but with contents that don't match that MIME type. This file would then be selectable but would be rejected in the backend.

I would say if this is the price of moving forward with a better solution and to transition to `OR` it's worth it. It's not ideal that someone can select it in the frontend, but if our backend validation still prevents a malicious attack, that's ok. After all, this affects mostly people that upload files where extension and MIME type don't match. This shouldn't be the common case.

## What to send from the backend to frontend and pass to the browser

> The attribute value can get incredibly long, which could be bad for performance.

This seems to have been proven not a problem. 

I think the best way forward here would really be to generate a list of extensions on the backend and pass those to the frontend/browser. We could already have that in v4.x based on an AND logic and move to OR logic in v5.

rasteiner:
Closing in favor of https://github.com/getkirby/kirby/pull/6263

PR #4346, opened 2022-06-10
Add new public static method `charAt` for `Str` class

Awilum:
### Features
* New `Str::charAt()` method to get the character at specific index.

```php
$string = "Hellö Wörld";
echo Str::charAt($string, 0); //=> H
echo Str::charAt($string, 10); //=> d
```

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -298,6 +298,18 @@ class Str
     }
 
     /**
+     * Returns the character at index.
+     *
+     * @param string $string String to extract the character from
+     * @param int $index Position of the character
+     * @return string
+     */
+    public static function charAt(string $string = null, int $index): string
+    {
+        return static::substr($string, $index, 1);
+    }
+    
+    /**
      * Checks if a str contains another string
      *
      * @param string $string
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -298,7 +298,7 @@ class Str
     }
 
     /**
-     * Returns the character at index.
+     * Returns the character at index
      *
      * @param string $string String to extract the character from
      * @param int $index Position of the character
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -308,6 +308,7 @@ class Str
     {
         return static::substr($string, $index, 1);
     }
+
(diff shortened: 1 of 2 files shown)


src/Toolkit/Str.php
```diff
@@ -297,6 +297,18 @@ public static function camel(string $value = null): string
...
+     *
+     * @param string $string String to extract the character from
+     * @param int $index Position of the character
+     * @return string
+     */
+    public static function charAt(string $string = null, int $index): string
+    {
+        return static::substr($string, $index, 1);
+    }
+
```
```suggestion

```

lukasbestle (approved):

distantnative:
@Awilum @lukasbestle I'm wondering where we reach the point where usefulness and not blowing up the Str class more intersect. There are so many string methods one could implement, but it wouldn't really be that more useful.

Really wondering if

```php
Str::charAt($string, 4)
```

is so much more helpful than doing

```php
Str::substr($string, 4, 1)
```

Awilum:
@distantnative 
> @Awilum @lukasbestle I'm wondering where we reach the point where usefulness and not blowing up the Str class more intersect. There are so many string methods one could implement, but it wouldn't really be that more useful.
> 
> Really wondering if
> 
> ```
> Str::charAt($string, 4)
> ```
> 
> is so much more helpful than doing
> 
> ```
> Str::substr($string, 4, 1)
> ```

So, we should not have `Str` class and use native php function `mb_substr` instead of `Str::substr` ;)

```php
mb_substr($string, 4, 1);
``` 

But as I understand `Str` class and others inside the Toolkit folder is somehow a wrapper for native PHP functions + extended functionality. Toolkit should provides rich functionality to simplify project development (aka framework/stdlib). imho.

Already builtin `Str::substr` is a simple wrapper for `mb_substr` and new `charAt` is a function for a simple job: get single char from a string.

lukasbestle:
@Awilum We discussed your PR in our team and came to the conclusion that we won't merge it because the ratio between the usefulness of the method compared to the reduced overview of the class methods and code that needs to be maintained in the future is too low.

Despite of that, thank you very much for your contributions in the last days. They have all been of very high quality. :)

PR #4348, opened 2022-06-10
Fix `File::previewUrl` on custom `file::url` component 

### LINKED ISSUE #4143: file::url component doesn't work as expected in 3.6.x on File Panel Pages
## Description

The [file::url component](https://getkirby.com/docs/reference/plugins/components/file-urls) doesn't return the correct url to the panel in 3.6.x, downgrading to 3.5.x solves the issue. 

**Expected behavior**  
The open button and image preview links on the panel should use the returned url set on the plugin / component and not the default. See the URLS below on the screenshots bottom left.

**Kirby 3.6.2 Screenshot** 

**Kirby 3.5.8 Screenshot**

## To reproduce

1. Create a custom file::url component plugin
2. Navigate to a file page on the panel
3. Click the open or image preview.
4. See URL returns as default rather than a modified one via the component

afbora:
## This PR …

WIP and POC of @distantnative comment: https://github.com/getkirby/kirby/issues/4143#issuecomment-1075001280 @lukasbestle not agreed about solution but I just created the PR to force fixing the issue soon. Feel free to close any time.

### Fixes
 - `file::url` component doesn't work as expected in 3.6.x on File Panel Pages #4143

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -791,6 +791,18 @@ trait AppPlugins
     }
 
     /**
+     * Checks if a native component was extended
+     * @since 3.7.0
+     *
+     * @param string $component
+     * @return bool
+     */
+    public function isNativeComponent(string $component): bool
+    {
+        return $this->component($component) === $this->nativeComponent($component);
+    }
+
+    /**
      * Returns the native implementation
      * of a core component
      *
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -752,6 +752,11 @@ class File extends ModelWithContent
                     return $this->url();
                 }
 
+                // checks `file::url` component is extended
+                if ($this->kirby()->isNativeComponent('file::url') === false) {
+                    return $this->url();
+                }
+
                 return $url;
             case 'user':
                 return $this->url();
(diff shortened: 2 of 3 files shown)


distantnative:
@lukasbestle I understand that this would not be a 100% solution for custom file url components with CDNs - as then the Panel preview URLs aren't the original ones but the CDN ones again. But maybe still a better fix than status quo?

lukasbestle:
@distantnative Probably, yes. To be honest I find it hard to consider all use cases. May well be that we would have another regression, but maybe we wouldn't. We can certainly try it and 3.7.0 will be the best release to do that.

src/Cms/AppPlugins.php
```diff
@@ -790,6 +790,17 @@ protected function extensionsFromSystem()
         $this->extendTemplates($this->core->templates());
     }
 
+    /**
+     * Checks native component extended
+     *
```
distantnative, line 796: Should add `@since` tag.

lukasbestle: Text suggestion for the docblock:

```txt
Checks if a core component is the unchanged native component
```

bastianallgeier:
I really like this fix because it is so simple. I would go for it.

bastianallgeier (approved):

PR #4349, opened 2022-06-11
Fix multi selecting blocks without click

### LINKED ISSUE #4319: Multi select mode in blocks field is triggered without click
## Description
When I select a block, switch tabs and come back I'm suddenly in multi select mode. I think this happens because the last state the panel sees is me holding the option key for ⌘⌥→. Releasing the option key happens outside of the tab, though.

**Expected behavior**  

Multi select mode should only be triggered on click.

**Screenshots**  

CleanShot 2022-05-27 at 17 22 26 2022-05-27 17_23_46

## To reproduce
1. Select a block
2. Hold the option key
3. Switch the tab
4. Switch back to the tab with the block

## Your setup
**Kirby Version**  
3.6.6

afbora:
## This PR …

The issue fixed with listening `window.addEventListener('blur')` event and reset the multi select mode.

### Fixes
- Multi select mode in blocks field is triggered without click #4319

### Breaking changes

none

## Ready?

- ~[ ] Unit tests for fixed bug/feature~

### For review team

- ~[ ] Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)~

### DIFF
--- a/panel/src/components/Forms/Blocks/Blocks.vue
+++ b/panel/src/components/Forms/Blocks/Blocks.vue
@@ -174,15 +174,7 @@ export default {
 		}
 	},
 	created() {
-		// resets multi selecting on tab change
-		// keep only if there are already multiple selections
-		// triggers `blur` event when tab changed
-		window.addEventListener("blur", () => {
-			if (this.batch.length === 0) {
-				this.isMultiSelectKey = false;
-			}
-		});
-
+		this.$events.$on("blur", this.onBlur);
 		this.$events.$on("copy", this.copy);
 		this.$events.$on("focus", this.onOutsideFocus);
 		this.$events.$on("keydown", this.onKey);
@@ -190,6 +182,7 @@ export default {
 		this.$events.$on("paste", this.onPaste);
 	},
 	destroyed() {
+		this.$events.$off("blur", this.onBlur);
 		this.$events.$off("copy", this.copy);
 		this.$events.$off("focus", this.onOutsideFocus);
 		this.$events.$off("keydown", this.onKey);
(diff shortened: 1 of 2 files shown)


panel/src/components/Forms/Blocks/Blocks.vue
```diff
@@ -595,6 +595,15 @@ export default {
...
     select(block, event = null) {
+      // resets multi selecting on tab change
+      // keep only if there are already multiple selections
+      // triggers `blur` event when tab changed
+      window.addEventListener('blur', () => {
+        if (this.batch.length === 0) {
+          this.isMultiSelectKey = false;
+        }
+      });
+
```
distantnative: Isn't `select()` called on blur anyways already but as `select(null)`? I think we should try to rather work with that.

Actually the following comment already addresses the referenced issue:
```
// checks the event just before selecting the block
// especially since keyup doesn't trigger in with
// `ctrl/alt/cmd + tab` or `ctrl/alt/cmd + click` combinations
// for ex: clicking outside of webpage or another browser tab
if (event && this.isMultiSelectKey) {
```

So to me it seems this comment is actually wrong. Seems like `⌘⌥→` DOES trigger the `keyup` event - or at least on the `keydown` event. We actually listen to both. Maybe that's also the issue.

Or we can inspect the `event` parameter and from that determine which calls of select is actually happening.

Just adding another listener to blur (which could be any blur, not just another tab) - not sure if that feels right.

afbora: > Isn't select() called on blur anyways already but as select(null)? I think we should try to rather work with that.

I know the `select(null)`, already tried. `blur` not triggered for the block when browser tab changed. So I listened the blur event of the window.

> Actually the following comment already addresses the referenced issue:

These lines prevent the meta key from getting stuck when switching from another tab.

bastianallgeier (dismissed):
It works great!

bastianallgeier:
I rebased and moved the event registration to the created method to only register it once. Otherwise it would have been registered multiple times on every select call. 

@distantnative @afbora could you have another look at it?

bastianallgeier (dismissed):

afbora:
Thank you @bastianallgeier for improvement 💛 Still works great! I leave the technical review to @distantnative

distantnative (dismissed):

distantnative (approved):

PR #4352, opened 2022-06-12
Replace `$_SERVER` with new `Environment` class in the core

lukasbestle:
As discussed in https://github.com/getkirby/kirby/pull/4285#discussion_r889576280.

### Enhancements

- Methods in Kirby's core now use the global environment object instead of `$_SERVER` wherever possible
- JSON errors in debug mode now contain the file path relative to the document root for consistency with the API that already handles it like this

## Changes to 3.7 code (not for release notes)

- `Environment::sanitize()` now is a static method (already doesn't depend on the instance)
- New `Environment::getGlobally()` method to access info from the global app object with a fallback to `$_SERVER`

## Ready?

### For review team

### DIFF
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -680,6 +680,12 @@ class Api
      */
     public function responseForException(Throwable $e): array
     {
+        if (isset($this->kirby) === true) {
+            $docRoot = $this->kirby->environment()->get('DOCUMENT_ROOT');
+        } else {
+            $docRoot = $_SERVER['DOCUMENT_ROOT'] ?? null;
+        }
+
         // prepare the result array for all exception types
         $result = [
             'status'    => 'error',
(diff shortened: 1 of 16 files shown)


src/Api/Api.php
```diff
@@ -680,14 +680,20 @@ public function responseFor404(): array
      */
     public function responseForException(Throwable $e): array
     {
+        if (isset($this->kirby) === true) {
+            $docRoot = $this->kirby->environment()->get('DOCUMENT_ROOT');
+        } else {
+            $docRoot = $_SERVER['DOCUMENT_ROOT'] ?? null;
+        }
```
afbora, line 687: Doesn't your new `::getGlobally` method already do what these lines do? Is it returning different value in different instances?

````php
$docRoot = Environment::getGlobally('DOCUMENT_ROOT');
````

lukasbestle: Since we have an actual `$this->kirby` instance here, I think we should use it directly and not rely on the global instance. One could set up another `App` instance and create an `Api` instance from that.

afbora (approved):
LGTM

PR #4357, opened 2022-06-12
`Pages`: Consistent find behavior for non-children

### LINKED ISSUE #4105: `$pages->find()` doesn't work with slug for the default language in a multilang installation
## Description
This code used to work:

```
<?= $site->grandChildren()->find('exploring-the-universe')?>
```

And still does in a non-language installation.

However, it doesn't work in a multilanguage installation **for the default language**, while it works for the non-default languages.

**Expected behavior**  
The code should return the given page in all circumstances.

## Your setup
**Kirby Version**  
3.6.1.1

**Additional context**  
[](https://forum.getkirby.com/t/pages-find-not-working-as-expected/24531)

lukasbestle:
### Fixes

- The behavior of the `$pages->find()`, `$pages->findById()` and `$pages->findByUri()` methods in non-direct-child collections (e.g. `$page->grandChildren()`) is now consistent between single-lang and multi-lang installations #4105

### Enhancements

- `$pages->find()` can now find pages by their translated URI in secondary languages of multi-lang setups, even in collections without a parent (like `$page->grandChildren()`)

### Deprecations

- The `$pages->findById()` and `$pages->findByUri()` methods have been deprecated and will be removed in v3.8.0. If you want to directly get a page by ID, use `$pages->get()`. If you want the previous fuzzy behavior that queries both by ID and URI, use `$pages->find()`.
- The `$pages->findByIdRecursive()` method has been deprecated and will be removed in v3.8.0. Please use `$pages->find()` instead.
- The `$files->findById()` method has been deprecated and will be removed in v3.8.0. Please use `$files->find()` instead.
- The `$pages->findByKey()`, `$files->findByKey()` and `$users->findByKey()` methods have been marked as internal. Please use the `find()` methods instead.

### Breaking changes

- The `$pages->children()`, `$pages->drafts()` and `$pages->index()` methods no longer set the `$pages->parent()` object as the collection items can have multiple different parents. This can change the behavior when finding collection items in secondary languages and when merging collections. #4105

## Open question

What doesn't work is a mixed-language query in the secondary language, something like this:

```php
$site->find('grandma')->grandChildren()->findById('grandma/mother/kind');
```

This hasn't worked before either, only in "root" collections like `$site->children()`. And I don't think we can make it work without a lot of added complexity. I might be wrong though.

For the more common case of the full translated URI I added another case.

@neildaniels Do you think it makes sense? Or do you see a better solution that covers such queries with (partially or fully) translated URIs in the secondary language better?

## Ready?

### For review team

### For Lukas when this PR is merged; see https://github.com/getkirby/kirby/issues/4105#issuecomment-1021543529

- [x] Add the breaking change to 3.6 release notes for future reference for those updating their existing sites to 3.6.
- [x] Document the expected usage of the `find()` and `findBy('slug', ...)` methods in collections better.

### DIFF
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -201,62 +201,38 @@ class Pages extends Collection
      * Finds a page in the collection by id.
      * This works recursively for children and
      * children of children, etc.
+     * @deprecated 3.7.0 Use `$pages->get()` or `$pages->find()` instead
+     * @todo 3.8.0 Remove method
+     * @codeCoverageIgnore
      *
      * @param string|null $id
      * @return mixed
      */
     public function findById(string $id = null)
     {
-        if ($id === null) {
-            return null;
-        }
+        Helpers::deprecated('Cms\Pages::findById() has been deprecated and will be removed in Kirby 3.8.0. Use $pages->get() or $pages->find() instead.');
 
-        // remove trailing or leading slashes
-        $id = trim($id, '/');
-
-        // strip extensions from the id
-        if (strpos($id, '.') !== false) {
-            $info = pathinfo($id);
-
-            if ($info['dirname'] !== '.') {
-                $id = $info['dirname'] . '/' . $info['filename'];
-            } else {
-                $id = $info['filename'];
-            }
-        }
-
-        // try the obvious way
-        if ($page = $this->get($id)) {
-            return $page;
-        }
-
-        $start = is_a($this->parent, 'Kirby\Cms\Page') === true ? $this->parent->id() : '';
-        if ($page = $this->findByIdRecursive($id, $start, App::instance()->multilang())) {
-            return $page;
-        }
-
-        // for secondary languages, try the full translated URI
-        if (
-            App::instance()->multilang() === true &&
-            App::instance()->language()->isDefault() === false &&
-            $page = $this->findBy('uri', $id)
-        ) {
-            return $page;
-        }
-
-        return null;
+        return $this->findByKey($id);
     }
 
     /**
      * Finds a child or child of a child recursively.
+     * @deprecated 3.7.0 Use `$pages->find()` instead
+     * @todo 3.8.0 Integrate code into `findByKey()` and remove this method
      *
      * @param string $id
      * @param string|null $startAt
      * @param bool $multiLang
      * @return mixed
      */
-    public function findByIdRecursive(string $id, string $startAt = null, bool $multiLang = false)
+    public function findByIdRecursive(string $id, string $startAt = null, bool $multiLang = false, bool $silenceWarning = false)
     {
+        // @codeCoverageIgnoreStart
+        if ($silenceWarning !== true) {
+            Helpers::deprecated('Cms\Pages::findByIdRecursive() has been deprecated and will be removed in Kirby 3.8.0. Use $pages->find() instead.');
+        }
+        // @codeCoverageIgnoreEnd
+
         $path       = explode('/', $id);
         $item       = null;
         $query      = $startAt;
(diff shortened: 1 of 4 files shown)


src/Cms/Pages.php
```diff
@@ -230,6 +230,15 @@ public function findById(string $id = null)
...
+        if (
+            App::instance()->multilang() === true &&
+            App::instance()->language()->isDefault() === false &&
+            $page = $this->findBy('uri', $id)
+        ) {
+            return $page;
+        }
+
         $start = is_a($this->parent, 'Kirby\Cms\Page') === true ? $this->parent->id() : '';
         $page  = $this->findByIdRecursive($id, $start, App::instance()->multilang());
```
neildaniels:
```suggestion
        $start = is_a($this->parent, 'Kirby\Cms\Page') === true ? $this->parent->id() : '';
        $page  = $this->findByIdRecursive($id, $start, App::instance()->multilang());

        // for secondary languages, try the full translated URI
        if (
            !$page &&
            App::instance()->multilang() === true &&
            App::instance()->language()->isDefault() === false &&
        ) {
            $page = $this->findBy('uri', $id);
        }
```

I would strongly suggest doing the lookup by `uri` _after_ letting the normal lookup occur by `id`. There is an existing oddity that `findById()` actually _prefers_ untranslated versions. Swapping the order should do a better job at keeping this behavior consistent.

See the end of the `testFindByIdWithSwappedSlugsTranslated()` test case lookups. https://github.com/getkirby/kirby/blob/27ae49e07a91323a05c8b87f950fa6ed27bb64ed/tests/Cms/Pages/PagesTest.php#L551-L653

neildaniels:
@lukasbestle I made one small suggestion on swapping the order of execution, but I otherwise think your solution is the simplest solution that avoids creating even more breaking changes.

However, I would like to call out the fact that `findById()` is now diverging even more from what the name implies. There's currently a `findByUri()` method that would probably be better served by this improvement.

I genuinely think these methods should diverge and start doing different things (with `findById()` strictly operating on `id` and `findByUri()` being as flexible (or more flexible) than current behavior). But, making these diverge would be a significant breaking change on its own.

A minimal change to `findByUri()` could look like this:
```php
    public function findByUri(string $id)
    {
        $page = $this->findById($id);

        if (
            !$page &&
            App::instance()->multilang() === true &&
            App::instance()->language()->isDefault() === false
        ) {
            $page = $this->findBy('uri', $id);
        }

        return $page;
    }
```

src/Cms/Pages.php
lukasbestle: Is there a case where the order of the checks matters, i.e. leads to a different result? The existing unit tests all pass, but maybe there's a case that's currently not tested.

When implementing this, I was thinking that all ways of trying to find the page are equal and that they all contribute equally to a non-`null` return value. I.e. only if all ways fail, the result will be `null`.

So with that in mind, I opted for this order because finding by `uri` is faster than the recursive lookup. So for this simple case of "full localized URI is given", the performance will be higher when the code is in this order.

lukasbestle:
@neildaniels Thank you very much for your quick and detailed reply. 💛

> However, I would like to call out the fact that `findById()` is now diverging even more from what the name implies. There's currently a `findByUri()` method that would probably be better served by this improvement.
> 
> I genuinely think these methods should diverge and start doing different things (with `findById()` strictly operating on `id` and `findByUri()` being as flexible (or more flexible) than current behavior). But, making these diverge would be a significant breaking change on its own.

`findByUri` and `findByKey` are currently both aliases to `findById`.

What we need to keep in mind is that `findByKey` is called by `Toolkit\Collection::find()` (which is inherited by `Pages`). So if you call `$pages->find()`, Kirby will internally also use `findById`, which then supports both querying by ID and by URI. So the breaking change would be even more massive because `find` would need to decide on one of the two alternatives (ID *or* URI).

To be honest I find all these aliases massively confusing. When working on this PR, I was looking through all the code (`Pages`, `Cms\Collection` and `Toolkit\Collection`) to find which method depends on what. Only to find out that they are all exact aliases.

What if we deprecate `findById`, `findByUri` and `findByIdRecursive` entirely and only support `find` in the future? This would mean:

- The code of `findById` is moved to `findByKey`
- `findById` and `findByUri` are made temporary aliases of `findByKey` but deprecated and later removed
- `findByKey` is marked as `@internal` as it's only used by `Toolkit\Collection::find()`
- `findByIdRecursive` is deprecated and later directly integrated into `findByKey`

Another alternative would be to keep `findById` and `findByUri` but deliberately reimplement them to do what their name suggests (like Neil wrote). In this case we would still deprecate `findByIdRecursive` and mark `findByKey` as `@internal`. The `find` method would get the previous behavior that supports both ID and URI querying.

lukasbestle:
Pinging @bastianallgeier and @distantnative for opinions.

src/Cms/Pages.php
neildaniels: Here's an additional test case that could go at the end of `testFindByIdWithSwappedSlugsTranslated()`. It's very similar to the existing last set of tests in that function.

```php
        $pages = new Pages($site->children()->find('aaa', 'aaa/bbb', 'zzz'));
        $this->assertIsPage($pages->findById('aaa'), 'aaa');
        $this->assertIsPage($pages->findById('aaa/bbb'), 'aaa/bbb');
        $this->assertIsPage($pages->findById('zzz'), 'zzz');
        $this->assertIsPage($pages->findById('zzz/yyy'), 'zzz/yyy');
```

In your current proposed logic, under the German test case, the final line will actually return the page that has the `id` of `aaa/bbb`, but if you run the exact same tests under the default language test, it will return `zzz/yyy`.

lukasbestle: You are absolutely right, thanks for spotting! 👍

lukasbestle:
I went through all the `find*` methods of `Pages`, `Files` and `Users`. As we discussed, I deprecated all methods that are duplicates/aliases. While reading though the code of `$pages->findById()`, I realized that if we stripped that from anything URI-related as we planned, it would just be an alias of `$pages->get()`. So I deprecated it as well.

Now we really only have `find()` and `get()` methods in each of the CMS collections. I'm really happy with that result as it's much easier to grasp and less confusing. No functionality is lost as the `find()` methods can do everything that the deprecated methods could, with the addition that you can pass multiple queries and get a filtered collection back.

bastianallgeier (approved):
LGTM

PR #4358, opened 2022-06-13
New icons

bastianallgeier:
## Features

- New icons
  - archive
  - badge
  - brush
  - discord
  - discount
  - folder-structure
  - github
  - money
  - palette
  - pen
  - store
  - ticket

Screenshot 2022-06-13 at 17 01 25

## Enhancements

- Improved file icons

## Ready?

### For review team

- [ ] ~~Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)~~


panel/src/components/Layout/ItemImage.vue
```diff
@@ -18,7 +18,6 @@
     <k-aspect-ratio v-else :ratio="ratio">
       <k-icon
         :color="$helper.color(image.color)"
-        :size="size"
```
distantnative, line 21: Not supporting `size` here doesn't seem right - the component should keep supporting it.

bastianallgeier: You are right. That was obviously just the most lazy way to solve it :)

bastianallgeier: Oh wait, the icon actually never really supported the size attribute. It's just a computed prop that handles the size according to the layout prop.

distantnative: Huh? It does support a `size` prop: https://github.com/getkirby/kirby/blob/main/panel/src/components/Misc/Icon.vue#L7

distantnative: Ah now get it, in `k-item-image` you mean. Ok then this line removal could stay but we should also remove the computed prop.

distantnative: Although this means that also emojis etc. will be sized differently now - not just your new file icons. Not sure if that's what we want.

bastianallgeier: I'll give it another look tomorrow. But I think it's fine if the emojis are also just the regular size in cards. If someone complains we can still turn it back :)

bastianallgeier: Looks good to me …
Screenshot 2022-06-14 at 10 41 14

LeBenLeBen: I’m going to be the one that says it was better before 😬 It helps distinguish pages better I think.

Capture d’écran 2022-07-14 à 14 46 25

Capture d’écran 2022-07-14 à 14 46 16

PR #4360, opened 2022-06-13
GitHub CI workflow updates

afbora:
Would be great if import `uses` from ENV (like `./.github/workflows/repositories.yml`) to manage easy but I couldn't do that.


bastianallgeier (approved):
The test stuff looks good to me. Can't say much about the actions configuration.

Issue #4172, opened 2022-02-22
Topbar and Noticebar (unsaved changes) disappear on mobile Browsers (iOS Safari, iOS Chrome) while editing


teichsta:
## Description
(as discussed here https://forum.getkirby.com/t/topbar-and-noticebar-unsaved-changes-disappear-on-mobile-browsers-ios-safari-ios-chrome/24794?u=teichsta)

When using kirby panel on mobile devices (e.g. iPhone or iPad) the Top- und Noticebar appearance is inconsistent. Sometimes the bars disappear when entering a textarea field (but sometimes not). Once the bars disappeared there is no chance to scroll to them again (as long as the editor is activ). After leaving the editor the bars appear again.

**Expected behavior**  
Both bars shall be visible at any time. Otherwise the editor is not aware of his changes and the need to save these changes to the server.

**Screenshots**  

![image](https://user-images.githubusercontent.com/2108636/155115038-a0c055f7-0509-443c-a4fc-b4fbabccbdfa.png)

It seems that two scrollbars are active at the same time:

![image](https://user-images.githubusercontent.com/2108636/155118375-86ef270d-2055-498e-8c27-efe4afb5fafd.png)

Observed in that context: Sometimes the toolbar of the textarea remains fixed while scrolling (see screenshot, also see the cursor which still is shown while scrolling)

![image](https://user-images.githubusercontent.com/2108636/155115396-0b92aaa9-cdbb-41f9-ba61-ccb0a451bd61.png)

## To reproduce

Add a textarea to your page blueprint and enter the edit mode. The virtual keyboard appears and the bars disappear (or not). Once the bar disappear you will be able (with some hassle and touching on the outer right border) to make a second scrollbar appear which seem to scroll the entire viewport. If you manage to touch this second scrollbar you are able to scroll to either top and noticebar (unsaved changes).

PR #4362, opened 2022-06-14
Make the body scrollable

medienbaecker:
## This PR makes the body scrollable

You positioned `.k-panel-view` absolutely and made it scrollable. Not using the regular body scroll has some drawbacks so I'd suggest using a regular body scroll.

I guess you had a good reason for this decision so please take this PR as a basis for discussion. I'm assuming it has something to do with scrollable drawers and the body scrolling behind them? If that's true there is a modern solution to this: https://ishadeed.com/article/prevent-scroll-chaining-overscroll-behavior/

### Fixes

- [x] Inertia/rubber-band scrolling on macOS and iOS
- [x] ⌘↑ and ⌘↓ shortcuts to go to the page start/end
- [x] Full page panel screenshots
- [x] https://github.com/getkirby/kirby/issues/4172
- [ ] Prevent losing scroll positions across pages/reloads*

\* I expected the scroll position to persist but unfortunately it does not. It probably has something to do with how Vue is populating the DOM. There seems to be a short moment when the page is very short — probably leading to the lost scroll position. This was my main motivation behind the PR because losing the scroll position can be very frustrating.

### Breaking changes

Tbd

### For review team


bastianallgeier:
We tried that a couple times and it always had some form of side effects. But we will maybe try it again for 3.8. It is definitely a cleaner solution.

medienbaecker:
An update regarding the scroll restoration: By adding a temporary `min-height` to `.k-page-view` we can make sure it's not 0px tall while switching views. While this has no effect with the current absolutely positioned `.k-panel-view`, it's working with a regular scrolling body.

distantnative:
@medienbaecker gave it a first try and so far haven't run into any major issues. I think right now we are missing some padding at the page end, that should be fixed - but probably easy to do.

Could you also remove the dist files from the PR, they will always cause merge conflicts. Better to have this without dust files, they will be updated before releases. Thanks!

medienbaecker:
@distantnative I removed the ~~dust~~ dist files and added back the padding I removed for some reason I can't remember.

medienbaecker:
As I mentioned before my main motivation for this PR is to get scroll restoration to work. I have confirmed that while switching views there's a short moment where the body is 0px tall, resulting in lost scroll positions.

In order to fix this we can 1. set the min height to the current height and 2. remove the min height once the new view has loaded.

Now while there is a `onFinish` function it seems to fire before the page is fully loaded, removing the min height too soon. Here's a test with a throttled network connection:

CleanShot 2022-07-06 at 13 07 46 2022-07-06 13_10_16

This is easy enough in `panel/src/fiber/app.js`:
```diff
onStart: () => {
+  document.body.style.minHeight = document.body.clientHeight + "px";
},
```

But this doesn't work:
```diff
onFinish: () => {
  if (this.$api.requests.length === 0) {
    this.$store.dispatch("isLoading", false);
+   document.body.style.minHeight = 0;
  }
},
```

It has probably something to do with my limited Vue knowledge so I'd appreciate any nudge in the right direction.

distantnative:
@medienbaecker It technically isn't wrong: the view has finished loading when the event is fired. But our sections load async, so of course the page height could still change later than that.

medienbaecker:
@distantnative Thanks for the insight! I can see the loading indicator in the breadcrumb is spinning as long as sections are being loaded asynchronously. Would it be possible/a good idea to listen to that?

distantnative:
This would be listening to `this.$store.state.isLoading` - however `isLoading` will be true anytime ay (non-silent) API request is in progress. So not just for loading a view.

medienbaecker:
I could not yet find an elegant way of listening to the non-existing "finished loading everything" event. Do you think it would be feasible to add that to the core?

lukasbestle:
Wouldn't that cause issues when a section updates after the initial load and the size of the section changes? Or e.g. when switching between tabs.

medienbaecker:
@lukasbestle Absolutely not. I outlined my reasoning for listening to the "finished" event in this comment: https://github.com/getkirby/kirby/pull/4362#issuecomment-1176100691

The `min-height` would be removed once the the new view has loaded. We only need this because we don't want the page to be at a 0px height while switching views because this overwrites the saved scroll position to be 0.

If this is not the right way to do it we can of course pursue another approach. For example saving the scroll position somewhere and reapplying this, just like Inertia does: https://inertiajs.com/scroll-management
I think this behavior is a very important part of Inertia and it's currently missing from Kirby's Inertia implementation.

lukasbestle:
I see. @distantnative?

medienbaecker:
If it helps, we can also split the PR: one for the scrollable body and one for scroll preservation. Initially I thought this would be 100% related but it doesn't have to be.

distantnative:
I think it will be hard to implement that as there is no awareness of what everything is. The section all send their own API requests. So the API can say if it's currently running any or. One requests, but if that's all done or just a brief moment in between two sections sending their requests I'm not sure how to implement.

medienbaecker:
Then I'd suggest doing it the Inertia way. If this issue is not as important to you, I can try to understand how the router works and create a separate PR with something similar to Inertia's [scroll regions](https://inertiajs.com/scroll-management#scroll-regions). This could later be adapted to a scrolling body element.

Obviously I'd prefer it if you could work on this. While the code is properly commented I'll certainly need to spend some time figuring out how Kirby's Inertia clone works. But I know there are other issues and features on your roadmap.

distantnative:
@medienbaecker Maybe let's take a step back: what you mainly want is to restore scroll position, right? How would the panel/fiber determine when to restore scroll position?

medienbaecker:
@distantnative Just like the original I guess: https://github.com/inertiajs/inertia/blob/master/packages/inertia/src/router.ts#L420

When the user moves forward/back in history, the scroll position will be restored.

Edit:
I just compared Fiber's behavior to Inertia's. It looks like Kirby doesn't even save anything when navigating through browser history, unlike Inertia. This leads to initially empty views, exactly like when you're navigating regularly.
Something like this would probably be necessary: https://inertiajs.com/remembering-state
What was the reason for building your own Inertia clone again? 😅

distantnative:
@medienbaecker Fiber maybe started as Inertia clone but actually evolved from that with a much better fit for Kirby. Also, less cruft and dead code.

distantnative:
Btw I think I just remembered again why we stuck with making `.k-panel-view` scrollable. If you make body scrollable and have an open drawer that needs to scroll itself, you'll get double scroll bars that keep the layout jumping around on open/close of drawer when you e.g. are on windows (not like safari on Mac that just overlays scrollbars).

medienbaecker:
We can solve this by using [scrollbar-gutter](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter).

igregson:
Every time I try to use my keyboard to scroll in the panel I get frustrated to no end. Please, oh please, restore keyboard scrolling.

medienbaecker:
I don't give up yet. As previously suggested we can split the scroll restoration from the scrollable body with its usability improvements. In my opinion that's a huge improvement already.

Nevertheless, here's a quick recap of my quest to also have scroll restoration in the panel:

1. We can easily save the scroll position of a view [before swapping to another view](https://github.com/getkirby/kirby/blob/main/panel/src/fiber/app.js#L61-L63).
2. When the user navigates back to this view we want to restore the saved scroll position.
3. Since the Panel loads some things asynchronously, we need to listen to the `isLoading` state for the right moment to restore the scroll position.
4. There's already [a check](https://github.com/getkirby/kirby/blob/main/panel/src/fiber/app.js#L43-L47) for the `isLoading` state to stop the loading indicator.

In my tests it looks like the latter is triggered too early. It looks like it's checking for the number of requests before they're initialized. Here's a timeline of events:

Devtools console screenshot

It's hardly visible but the loading indicator also skips some degrees when this happens. It seems like [the comment](https://github.com/getkirby/kirby/blob/main/panel/src/fiber/app.js#L38-L42) is therefore lying. Not great, not terrible you might say. But for scroll restoration this has to be fixed.

You can test this yourself by changing the [onFinish function](https://github.com/getkirby/kirby/blob/main/panel/src/fiber/app.js#L43-L47):

```diff
onFinish: () => {
+	console.log('onFinish', this.$api.requests.length);
	if (this.$api.requests.length === 0) {
		this.$store.dispatch("isLoading", false);
	}
},
```

It will always return `0`.

distantnative:
The issue is that loading one view can trigger multiple `onFinish` when one requests finishes loading before other asynchron requests even have been started.

medienbaecker:
Thanks for your answer, Nico. Triggering multiple `onFinish` would mean I see my `console.log` multiple times. I tested all kinds of different scenarios, e.g. with >10.000 pages and throttled network connection, but I never saw that.

In what scenario could that happen and is it relevant to scroll restoration?

medienbaecker:
Here's how I imagine it to work:

## User navigates from [Page A] to [Page B]:
1. Save height and scrollY of [Page A]
2. Switch the view from [Page A] to [Page B]

## User navigates back from [Page B] to [Page A]:
1. Apply saved height of [Page A] as a min height
2. Switch the view from [Page B] to [Page A]
3. Apply saved scrollY of [Page A]
4. Wait for asynchronous requests
5. Remove the min height

medienbaecker:
Instead of waiting for the asynchronous requests, they could also remember their last size and display a skeleton.

## Without skeleton:
no-skeleton

## With skeleton:
skeleton

distantnative:
I think remembering the height plus skeleton will be quite the hack as we this would easily fail with any custom sections right now.

Personally, I think we need a much cleaner solution for this, which can take more time. But throwing a lot of JavaScript at it I don't think will be a sustainable feature.

medienbaecker:
In my opinion the skeletons would be a big UX improvement but of course you're right about the custom sections.

I don't know if you missed it but I sent you a message on Discord. It would be great to discuss this in a quick call instead of me poking around the code cluelessly.

bastianallgeier:
We will solve this in v4.

PR #4382, opened 2022-06-15
Sections: change `query` to `searchterm`

### LINKED ISSUE #4380: [3.7.0-rc.1] pagedisplay section can't be loaded anymore
## Description

When we add a section with the "page display" plugin, kirby can't load this section.

**Expected behavior**  
Load the page display section and display a list of pages

**Screenshots**  
This is the error message:
Home_Mægazine

## To reproduce
Add a pagedisplay section in a blueprints
tested in the home page blueprint in the Starter-kit

```

options:
  changeStatus: false

sections:
  content:
    type: fields
    fields:
      headline:
        type: text
        width: 1/2
      gap:
        width: 1/2
      subheadline:
        type: text
        width: 1/2

#### Page section here ! #######################################################
  mysection:
    headline: Foo Pages
    type: pagesdisplay
    query: site.index.filterBy('title', '*=', 'Foo')
#########################################################################################

  photography:
    label: Photography
    extends: sections/albums
    status: listed
    size: small
    help: All projects for the homepage are being pulled from the photography page.

```

## Your setup
**Kirby Version**  
3.7.0 -rc-1

Linux, Firefox

distantnative:


config/sections/files.php
```diff
@@ -195,7 +195,6 @@
                 'link'     => $this->link(),
                 'max'      => $this->max,
                 'min'      => $this->min,
-                'query'    => $this->query,
```
distantnative, line 198: @afbora I didn't see the need to pass this back to the frontend - do you see any?

afbora: This was for default search term/query. But we removed this feature while PR in progress but we forgot this line. So OK.

afbora (approved):

PR #4390, opened 2022-06-16
Fix `columns` prop type when empty (array)

distantnative:


afbora (dismissed):
LGTM

panel/src/components/Layout/Collection.vue
```diff
@@ -57,7 +57,7 @@
  */
 export default {
   props: {
-    columns: Object,
+    columns: [Object, Array],
```
afbora:
```suggestion
    columns: {
      type: [Object, Array],
      default() {
        return {};
      }
    },
```

panel/src/components/Layout/Items.vue
```diff
@@ -44,7 +44,7 @@
 export default {
   inheritAttrs: false,
   props: {
-    columns: Object,
+    columns: [Object, Array],
```
afbora:
```suggestion
    columns: {
      type: [Object, Array],
      default() {
        return {};
      }
    },
```

panel/src/components/Layout/Collection.vue
afbora: I think it will be more stable that way.

bastianallgeier (approved):

PR #4392, opened 2022-06-17
Don't pin specific psr/log version

neildaniels:
## This PR …

This unpins a specific version for `psr/log` but simultaneously forces Composer to operate as if it was under PHP 7.4.

`psr/log` was pinned to address an issue where building Kirby on PHP 8 wouldn't provide the necessary interfaces for `filp/whoops`. https://github.com/getkirby/kirby/pull/3895 https://github.com/getkirby/kirby/pull/3912

However, that had the unfortunate side-effect of overly constraining Composer-based Kirby installations, even though Kirby _doesn't_ use `psr/log` and Whoops is fully compatible with newer versions.

I've manually specified a PHP version of `7.4.0` to ensure the pre-built version of Kirby continues to select `psr/log` version `1.1.4` (both `2.x` and `3.x` require PHP 8).

### Refactoring

- Kirby's `composer.json` no longer pins the `psr/log` library to a specific version #4392

### Breaking changes

None

## Ready?

### For review team


lukasbestle (commented):
That's a very nice solution, thank you! 💛
Could you please run `composer update` to get the lockfile in line?

neildaniels:
Yep, done.

afbora:
Correct me if I'm wrong;

If the `platform` setting creates a fake platform for us, don't we need to run an extra `composer check-platform-reqs` command in the ci workflow?

neildaniels:
Good point. I just added that check.

lukasbestle (approved):
`composer analyze` is not actually run by CI. I have added a step to the version-specific CI job so that the platform requirements are checked for each version Kirby supports.

afbora (approved):

PR #4395, opened 2022-06-17
Fix structure field actions 

### LINKED ISSUE #4393: [3.7.0-rc.1] structure field with max value looks off
## description
dont know if you’re aware of this already. but since updating to `3.7.0-rc.1`, my structure fields look kinda strange. the last column either needs more space or the "remove" text needs to be removed (lol). this only happens when having a `max` value set.

**Screenshots**  
structure

## to reproduce
```yml
structure:
    label: Structure
    type: structure
    max: 3
    fields:
        title:
            label: Title
            type: text
        images:
            label: Carousel Images
            type: files
            query: page.images
```

then add 3 entries to the structure field.

---

hope it helps!
🙏🏻 ✨

afbora:
### Fixes

- #4393

### DIFF
--- a/panel/src/components/Navigation/OptionsDropdown.vue
+++ b/panel/src/components/Navigation/OptionsDropdown.vue
@@ -2,7 +2,6 @@
   <!-- Single option = button -->
   <k-button
     v-if="hasSingleOption"
-    v-bind="options[0]"
     :icon="options[0].icon || icon"
     :tooltip="options[0].tooltip || options[0].text"
     class="k-options-dropdown-toggle"
--- a/panel/src/components/Forms/Field/StructureField.vue
+++ b/panel/src/components/Forms/Field/StructureField.vue
@@ -275,8 +275,9 @@ export default {
      */
     options() {
       let options = [];
+      let more = this.duplicate && this.more && this.currentIndex === null;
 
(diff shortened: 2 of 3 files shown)


panel/src/components/Navigation/OptionsDropdown.vue
```diff
@@ -2,7 +2,6 @@
   <!-- Single option = button -->
   <k-button
     v-if="hasSingleOption"
-    v-bind="options[0]"
```
afbora, line 5: I'm not sure what for? This line causing the `e.apply is not a function` error.

bastianallgeier: The idea is that the first item in the options dropdown will be used to define the toggle button if there's just one option in the dropdown. I think we could make it more stable by adding a v-if to the toggle button to check if there are any options at all.

afbora: We're talking about `v-bind`, right? Right now my PR seems to be working great. 

I couldn't foresee where it would error/malfunction. Because we already have `v-if` condition for single and multiple options. Frankly, I have no idea about the solution 🤷‍♂️

bastianallgeier (approved):
I think it's fine like this

PR #4396, opened 2022-06-18
Fix deprecation warning for `$app->markdown()`

lukasbestle:
### Fixes

- Calling `$kirby->markdown()` or `$field->markdown()` no longer triggers a deprecation warning
- Custom `markdown` components no longer break because of `null` being passed to `$inline`

## Ready?

- [x] ~In-code documentation (wherever needed)~

### For review team

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -143,14 +143,12 @@ return [
      * @return string
      * @todo remove $inline parameter in in 3.8.0
      */
-    'markdown' => function (App $kirby, string $text = null, array $options = [], bool $inline = null): string {
+    'markdown' => function (App $kirby, string $text = null, array $options = [], bool $inline = false): string {
         static $markdown;
         static $config;
 
         // warning for deprecated fourth parameter
-        if ($inline === null) {
-            $inline = false;
-        } elseif (isset($options['inline']) === false) {
+        if (func_num_args() === 4 && isset($options['inline']) === false) {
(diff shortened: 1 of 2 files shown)


afbora:
Since we removed the `$inline` parameter, we have a deprecation warning. But it will give an error for users who use `bool $inline = false`. Just like in [unit testing](https://github.com/getkirby/kirby/pull/4396/commits/099c9db25a5b641f92e5d08d48b6b6772ec8f69e). Do we need to make a further explanation in the release notes for this? Because of this change throws fatal error not notice.

afbora (dismissed):

lukasbestle:
Good point. Custom components that don't allow `null` will break. Hm, this is really difficult. We only did the two-step change with the deprecation to avoid a sudden breaking change. But now we suddenly have one. :(

What we could do: Revert to passing `false` by default. And in the core component, instead of checking for `null`, we would check for `func_num_args() < 4`. Not sure if this would break anything else though.

afbora:
I think **func_num_args()** is genius idea 💯

lukasbestle:
The `func_num_args()` check does not cover use cases where the dev calls any of the methods with PHP 8 named arguments and leaves out `$options` completely, but it's unlikely anyone used this feature for these methods. Also it will just lead to a false negative for the deprecation warning, so nothing serious.

bastianallgeier (approved):

PR #4399, opened 2022-06-20
Fix broken params in URLs

### LINKED ISSUE #4367: [3.7.0-rc.1] Panel ignores port configured in config
## Description
When using a non standard port in the `url` option inside `config.php`  the panel does not load.

The console logs:
```
Failed to execute 'pushState' on 'History': A history state object with URL 'http://localhost/panel/installation' cannot be created in a document with origin 'http://localhost:8090'
```

The detection in the new Environment class seems to work since `$kirby->environment()->url()` returns the right path.

### LINKED ISSUE #4394: [3.7.0-rc.1] URL with params redirect to error page
A URL like this: `website.test/view:list` worked as expected in `3.6.6` and directed to the homepage, making the `view` param accessible.

In `3.7.0-rc.1.` the same URL redirects to the error page.

bastianallgeier:
## Fixes

- Params in URLs were ignored after the refactoring with the environment class. They are now properly detected again. https://github.com/getkirby/kirby/issues/4394
- Could be fixing https://github.com/getkirby/kirby/issues/4367 as well, but needs to be tested.

## Breaking changes

`Server::requestUri` now returns an instance of `Kirby\Http\Uri` instead of an array with path and query.

## Ready?

- [x] ~In-code documentation (wherever needed)~

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1222,11 +1222,11 @@ class App
             return $this->path;
         }
 
-        $requestUri  = $this->request()->url()->path()->toString();
-        $scriptPath  = $this->environment()->scriptPath();
-        $requestPath = Str::afterStart($requestUri, $scriptPath);
+        $current = $this->request()->path()->toString();
+        $index   = $this->environment()->uri()->path()->toString();
+        $path    = Str::afterStart($current, $index);
 
-        return $this->setPath($requestPath)->path;
+        return $this->setPath($path)->path;
     }
 
     /**
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -22,6 +22,7 @@ use Kirby\Toolkit\A;
 use Kirby\Toolkit\Config;
 use Kirby\Toolkit\Controller;
 use Kirby\Toolkit\Properties;
+use Kirby\Toolkit\Str;
 use Throwable;
 
 /**
@@ -104,6 +105,9 @@ class App
             $this->handleErrors();
         }
 
+        // a custom request setup must come before defining the path
+        $this->setRequest($props['request'] ?? null);
+
         // set the path to make it available for the url bakery
         $this->setPath($props['path'] ?? null);
 
@@ -114,7 +118,6 @@ class App
         // configurable properties
         $this->setOptionalProperties($props, [
             'languages',
-            'request',
             'roles',
             'site',
             'user',
@@ -1219,7 +1222,11 @@ class App
             return $this->path;
         }
 
-        return $this->path ??= $this->environment()->requestPath();
+        $requestUri  = $this->request()->url()->path()->toString();
+        $scriptPath  = $this->environment()->scriptPath();
+        $requestPath = Str::afterStart($requestUri, $scriptPath);
+
+        return $this->setPath($requestPath)->path;
     }
 
     /**
@@ -1246,9 +1253,11 @@ class App
             return $this->request;
         }
 
+        $env = $this->environment();
+
         return $this->request = new Request([
-            'cli' => $this->environment()->cli(),
-            'url' => $this->url('current')
+            'cli' => $env->cli(),
+            'url' => $env->requestUrl()
         ]);
     }
 
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1223,7 +1223,7 @@ class App
         }
 
         $current = $this->request()->path()->toString();
-        $index   = $this->environment()->uri()->path()->toString();
+        $index   = $this->environment()->baseUri()->path()->toString();
         $path    = Str::afterStart($current, $index);
 
         return $this->setPath($path)->path;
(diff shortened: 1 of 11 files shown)


src/Http/Environment.php
```diff
@@ -1008,12 +995,22 @@ public function toArray(): array
...
             'url'           => $this->url
         ];
     }
 
+    /**
+     * Returns the full base URL object
+     *
+     * @return \Kirby\Http\Uri
+     */
+    public function uri()
```
lukasbestle: I think the name `uri()` is very easy to confuse. It could both be the the base URI or the full request URI. What about `baseUri()`?

Also: For consistency, the `requestUri()` could return a `Http\Uri` object as well. I know that this out of scope of this PR, but we can never make such consistency fixes again after 3.7 without breaking changes.

bastianallgeier: I already refactored the requestUri method to return an URI object as well. 

We now have: 

```php
$environment->url()
$environment->uri()

$environment->requestUrl()
$environment->requestUri()
```

I like the idea to rename uri and url to base uri and base url. Will take care of it.

src/Http/Environment.php
```diff
@@ -180,7 +213,7 @@ public function detect(array $options = null, array $info = null): array
         // @codeCoverageIgnoreStart
         if (is_int($options['allowed']) === true) {
             Helpers::deprecated('
-                Using `Server::` constants for the `url` option has been deprecated and support will be removed in 3.8.0. Use one of the following instead: a single fixed URL, an array of allowed URLs to match dynamically, `*` wildcard to match dynamically even from insecure headers, or `true` to match automtically from safe server variables.
+                Using `Server::` constants for the `baseUrl` option has been deprecated and support will be removed in 3.8.0. Use one of the following instead: a single fixed URL, an array of allowed URLs to match dynamically, `*` wildcard to match dynamically even from insecure headers, or `true` to match automtically from safe server variables.
```
lukasbestle:
```suggestion
                Using `Server::` constants for the `allowed` option has been deprecated and support will be removed in 3.8.0. Use one of the following instead: a single fixed URL, an array of allowed URLs to match dynamically, `*` wildcard to match dynamically even from insecure headers, or `true` to match automtically from safe server variables.
```

src/Cms/App.php
```diff
@@ -1219,7 +1222,11 @@ public function path(): string
             return $this->path;
         }
 
-        return $this->path ??= $this->environment()->requestPath();
+        $current = $this->request()->path()->toString();
+        $index   = $this->environment()->baseUri()->path()->toString();
+        $path    = Str::afterStart($current, $index);
```
lukasbestle, line 1227: This could fail if for some reason `$current` does not start with `$index`. In theory this should never happen, but if it does, this method will return a full URI including scheme and host. Maybe something like this:

```suggestion

        // catch edge cases that shouldn't occur
        // @codeCoverageIgnoreStart
        if (Str::startsWith($current, $index) !== true) {
            throw new LogicException('Could not determine request path');
        }
        // @codeCoverageIgnoreEnd
        
        $path = Str::afterStart($current, $index);
```

lukasbestle (commented):
I've now read through the full code. It's really awesome how everything fits together now and how clean it is. 👍:tada:

Two minor things I noticed, apart from that it's ready to merge IMO.

src/Http/Environment.php
```diff
@@ -317,6 +353,26 @@ protected function detectAuto(bool $insecure = false): void
...
+    protected function detectBaseUri()
+    {
+        $this->baseUri = new Uri([
+            'host'   => $this->host,
+            'path'   => $this->path,
+            'port'   => $this->port,
+            'scheme' => $this->https ? 'https' : 'http',
+        ]);
+
+        $this->baseUrl = $this->baseUri->toString();
```
lukasbestle, line 371: Maybe we could make the `baseUrl` and `requestUrl` props lazy so that they are only initialized when needed.

bastianallgeier: I thought about that as well, but we actually need them immediately in the App class, so it doesn't really make sense.

src/Cms/App.php
bastianallgeier: I wouldn't do that to be honest. This is only about the path. Let's say for some weird reason the base URL has a different path than the request URL, then the request path would still be returned entirely, which sounds right to me.

lukasbestle: Right!

src/Http/Environment.php
lukasbestle: Currently on the go, so can't check. But I thought the App class only uses the Uri objects? If it really uses both strings then I agree with you.

lukasbestle (approved):

src/Http/Environment.php
bastianallgeier: Just double-checked. BaseUrl is used in the Core.php pretty much immediately. I think it doesn't make sense to convert one into a lazy method and keep the other unlazy, so I would keep it as it is.

PR #4401, opened 2022-06-20
Fix nested border radii

### LINKED ISSUE #4363: [3.7.0-rc.1] Same border radius for parent/child elements
## Description

Sorry about the nitpicky issue but I noticed something about the UI update in 3.7.0 — the `border-radius` in particular. When using rounded corners for parent/child elements it looks a lot better to gradually increase/decrease the size than to use the exact same `border-radius`. Otherwise you get ugly edges.

I'd suggest introducing `var(--border-radius-large)` and `var(--border-radius-small)` for dialogs/tags and every other nestable element:

CleanShot 2022-06-14 at 14 54 12@2x
CleanShot 2022-06-14 at 15 19 52@2x

bastianallgeier:
## Fixes

- Fixes nested border radii on tags and dialogs: https://github.com/getkirby/kirby/issues/4363

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Dialogs/Dialog.vue
+++ b/panel/src/components/Dialogs/Dialog.vue
@@ -264,7 +264,7 @@ export default {
   background: var(--color-background);
   width: 100%;
   box-shadow: var(--shadow-lg);
-  border-radius: var(--rounded);
+  border-radius: var(--rounded-md);
   line-height: 1;
   max-height: calc(100vh - 3rem);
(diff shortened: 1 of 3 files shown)


lukasbestle (commented):
Since the issue with the tags only occurs if they are nested in a field, maybe the smaller radius should be specific to tags within fields? Then they can still have the larger radius if the component is used outside of a field.

PR #4404, opened 2022-06-21
Make sure that block content is always an array

bastianallgeier:
## This PR …

We radically removed all old converters for the editor or builder fields to make them compatible with blocks. But those converters at least made sure that $params['content'] is always an array in the Block constructor. This PR checks for valid content arrays in the constructor now and returns an empty array if content is the wrong type. This adds at least a little more compatibility by making the blocks editable again in the Panel. 

## Ready?

### For review team

### DIFF
--- a/src/Cms/Block.php
+++ b/src/Cms/Block.php
@@ -79,11 +79,11 @@ class Block extends Item
 
         // make sure the content is always defined as array to keep
         // at least a bit of backward compatibility with older fields
-        if (is_array($params['content'] ?? []) === false) {
+        if (is_array($params['content'] ?? null) === false) {
             $params['content'] = [];
         }
 
-        $this->content  = $params['content']  ?? [];
+        $this->content  = $params['content'];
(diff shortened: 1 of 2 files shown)


src/Cms/Block.php
```diff
@@ -77,6 +77,12 @@ public function __construct(array $params)
             throw new InvalidArgumentException('The block type is missing');
         }
 
+        // make sure the content is always defined as array to keep
+        // at least a bit of backward compatibility with older fields
+        if (is_array($params['content'] ?? []) === false) {
+            $params['content'] = [];
+        }
+
         $this->content  = $params['content']  ?? [];
```
lukasbestle: To avoid the duplicated fallback code:

```suggestion
        if (is_array($params['content'] ?? null) === false) {
            $params['content'] = [];
        }

        $this->content  = $params['content'];
```

lukasbestle (approved):

PR #4433, opened 2022-06-26
Added boolean type to available columnTypes for createColumn

samzzi:
## This PR …

Adds an extra columnType for creating tables via the SQL and SQLite query builder classes. The reasons for this enhancement is to save up storage in MySQL tables.

* In the SQL query builder it will use the `TINYINT(1)` as boolean field instead of the `bool` alias for maximum compatibility.
* Also added in the SQLite query builder it will use the `INTEGER` as boolean field for columnType consistency with the SQL query builder.

### Features

- New `bool` column type for the `Database` classes

### Breaking changes

No.

### For review team

### DIFF
--- a/src/Database/Sql.php
+++ b/src/Database/Sql.php
@@ -147,7 +147,7 @@ abstract class Sql
             'text'      => '{{ name }} TEXT {{ unique }}',
             'int'       => '{{ name }} INT(11) UNSIGNED {{ null }} {{ default }} {{ unique }}',
             'timestamp' => '{{ name }} TIMESTAMP {{ null }} {{ default }} {{ unique }}',
-            'boolean'   => '{{ name }} TINYINT(1) {{ null }} {{ default }} {{ unique }}'
+            'bool'   => '{{ name }} TINYINT(1) {{ null }} {{ default }} {{ unique }}'
         ];
     }
 
--- a/src/Database/Sql.php
+++ b/src/Database/Sql.php
@@ -146,7 +146,8 @@ abstract class Sql
             'varchar'   => '{{ name }} varchar(255) {{ null }} {{ default }} {{ unique }}',
             'text'      => '{{ name }} TEXT {{ unique }}',
             'int'       => '{{ name }} INT(11) UNSIGNED {{ null }} {{ default }} {{ unique }}',
-            'timestamp' => '{{ name }} TIMESTAMP {{ null }} {{ default }} {{ unique }}'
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
I'm thinking if the name should be `bool` for consistency with PHP types (e.g. we also have an `int` column, not `integer`).

samzzi:
> I'm thinking if the name should be `bool` for consistency with PHP types (e.g. we also have an `int` column, not `integer`).

Makes sense since `int` is used for the integer column type. I've updated the PR.

BTW, fun fact (didn't know this myself), for typecasting in PHP you can use both `(bool)` and `(boolean)` (and `(int)` or `(integer)`) 😉

src/Database/Sql.php
```diff
@@ -146,7 +146,8 @@ public function columnTypes(): array
             'varchar'   => '{{ name }} varchar(255) {{ null }} {{ default }} {{ unique }}',
             'text'      => '{{ name }} TEXT {{ unique }}',
             'int'       => '{{ name }} INT(11) UNSIGNED {{ null }} {{ default }} {{ unique }}',
-            'timestamp' => '{{ name }} TIMESTAMP {{ null }} {{ default }} {{ unique }}'
+            'timestamp' => '{{ name }} TIMESTAMP {{ null }} {{ default }} {{ unique }}',
+            'bool'   => '{{ name }} TINYINT(1) {{ null }} {{ default }} {{ unique }}'
```
lukasbestle:
```suggestion
            'bool'      => '{{ name }} TINYINT(1) {{ null }} {{ default }} {{ unique }}'
```

src/Database/Sql/Sqlite.php
```diff
@@ -44,7 +44,8 @@ public function columnTypes(): array
             'varchar'   => '{{ name }} TEXT {{ null }} {{ default }} {{ unique }}',
             'text'      => '{{ name }} TEXT {{ null }} {{ default }} {{ unique }}',
             'int'       => '{{ name }} INTEGER {{ null }} {{ default }} {{ unique }}',
-            'timestamp' => '{{ name }} INTEGER {{ null }} {{ default }} {{ unique }}'
+            'timestamp' => '{{ name }} INTEGER {{ null }} {{ default }} {{ unique }}',
+            'bool'   => '{{ name }} INTEGER {{ null }} {{ default }} {{ unique }}'
```
lukasbestle:
```suggestion
            'bool'      => '{{ name }} INTEGER {{ null }} {{ default }} {{ unique }}'
```

lukasbestle (approved):

PR #4445, opened 2022-06-29
Support anonymous plugins (without .php file)

### LINKED ISSUE #4419: Initializing plugins which don't require PHP
When building a small UI plugin for the panel I noticed that – although my plugin only features CSS and JavaScript – I had to include an `index.php`. Although this PHP file's requirements are really minimal, I was surprised by the fact that it forced me to pass an empty array of options to work:

```php
Kirby::plugin('hananils/my-plugin', []); // this works
Kirby::plugin('hananils/my-plugin'); // this doesn't
```

This can be difficult to debug when you think you already defined your plugin and don't understand why your CSS and JavaScript files are not included in the panel. 

### Expected behaviour

1. If the plugin could also work without an "empty" `index.php`, this would be preferred. 
2. If this barebone definition is actually required for the system to work, it should be forgiving regarding the empty/missing options array and should accept both.

### Kirby version

Kirby 3.7 RC2 – not tested with prior versions, but my guess is it's the same.

distantnative:
### Enhancements
- Plugins that only provide Panel styles and/or scripts don't need to add an `index.php` anymore, `index.js` and `index.css` will still be included in the Panel by registering an anonymous plugin for them automatically
Fixes #4419

## Ready?

### For review team

### DIFF
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -903,6 +903,8 @@ trait AppPlugins
                 // to be picked up by the Panel\Document class when
                 // rendering the Panel view
                 static::plugin('plugins/' . $dirname, ['root' => $dir]);
+            } else {
+                continue;
             }
 
             $loaded[] = $dir;
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -885,14 +885,25 @@ trait AppPlugins
                 continue;
             }
 
-            $dir   = $root . '/' . $dirname;
-            $entry = $dir . '/index.php';
+            $dir = $root . '/' . $dirname;
 
-            if (is_dir($dir) !== true || is_file($entry) !== true) {
+            if (is_dir($dir) !== true) {
                 continue;
             }
 
-            F::loadOnce($entry);
+            $entry  = $dir . '/index.php';
+            $script = $dir . '/index.js';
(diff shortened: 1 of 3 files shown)


src/Cms/AppPlugins.php
```diff
@@ -885,14 +885,25 @@ protected function pluginsLoader(): array
...
+                F::loadOnce($entry);
+            } elseif (is_file($script) === true || is_file($styles) === true) {
+                // if no PHP file is present but an index.js or index.css,
+                // register as anonymous plugin (without actual extensions)
+                // to be picked up by the Panel\Document class when
+                // rendering the Panel view
+                static::plugin('plugins/' . $dirname, ['root' => $dir]);
+            }
 
             $loaded[] = $dir;
```
lukasbestle, line 910: Not sure about the implications of this line being run even for empty directories now. Previously this line would only be reached if an `index.php` was in the directory, but now even for empty directories.

And with "not sure" I mean really that. Doesn't mean that we need to change it.

distantnative: True, to be safe we could just put that line inside both if/elseif statements

afbora: What do think about adding `continue` within `else` fallback?

distantnative: Would work as well

afbora (approved):

PR #4449, opened 2022-06-30
New plugin static methods: `::register()` and `::get()`

afbora:
## This PR …

- New `Plugin` class alias added
- New `Plugin` static methods (alias of `App::instance()->plugin()`) added: `::get()` and `::register()`

````php
Plugin::get('plugin/test'); // plugin getter
Plugin::register('plugin/test'); // plugin setter without extends
Plugin::register('plugin/test', ['options' => []]); // plugin setter with extends
````

### Fixes
- Initializing plugins which don't require PHP #4419

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Plugin.php
+++ b/src/Cms/Plugin.php
@@ -87,6 +87,17 @@ class Plugin extends Model
 	}
 
 	/**
+	 * Kirby plugin getter
+	 *
+	 * @param string $name
+	 * @return \Kirby\Cms\Plugin|null
+	 */
+	public static function get(string $name)
+	{
+		return App::plugin($name);
+	}
+
+	/**
 	 * Returns the unique id for the plugin
 	 *
 	 * @return string
@@ -171,6 +182,18 @@ class Plugin extends Model
 	public function option(string $key)
 	{
 		return $this->kirby()->option($this->prefix() . '.' . $key);
+	}
+
+	/**
+	 * Registers a plugin in the CMS
+	 *
+	 * @param string $name
+	 * @param array $extends
+	 * @return \Kirby\Cms\Plugin
+	 */
+	public static function register(string $name, array $extends = [])
+	{
+		return App::plugin($name, $extends);
 	}
 
 	/**
--- a/tests/Cms/Plugins/PluginTest.php
+++ b/tests/Cms/Plugins/PluginTest.php
@@ -3,7 +3,7 @@
 namespace Kirby\Cms;
 
 /**
- * @coversDefaultClass Kirby\Cms\Plugin
+ * @coversDefaultClass \Kirby\Cms\Plugin
  */
 class PluginTest extends TestCase
 {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/Cms/Plugin.php [hunk 1]
```diff
@@ -173,6 +184,18 @@ public function option(string $key)
...
+    /**
+     * Kirby plugin register
+     *
+     * @param string $name
+     * @param array $extends
+     * @return \Kirby\Cms\Plugin
+     */
+    public static function register(string $name, array $extends = [])
+    {
+        return App::instance()->plugin($name, $extends);
```
distantnative: If these are static methods, they should also only call static methods (and not on the app instance)

afbora: Could you explain in a little more detail? 🙈

distantnative: You are creating new static methods, e.g. `Plugin::register()` but right now those create a instance/object of Kirby via `App::instance()` - I think this shouldn't be the case, but rather also call `App::plugin()` statically.

afbora: Done 👍

lukasbestle (commented):
Are the methods still needed with the change to automatically register an anonymous plugin?
To be honest I think that the more methods we have, the more the overview is reduced. It will make the method lists in the reference longer without actually adding much benefit.

afbora:
IMO The goal is to make the code clear and easy to use for this PR not just for anonymous plugins. `Kirby::plugin()` that a method that acts as both a getter and a setter is confusing. The fact that each method has only one function clears the confusion. I think `Plugin::get()` and `Plugin::register()` is a consistent change and methods for the future (K4).

It's not a critical or very necessary change. The decision is yours 👍

lukasbestle:
I'd be fine with both. 🙂
@distantnative?

tests/Cms/Plugins/PluginTest.php
```diff
@@ -79,6 +79,18 @@ public function testExtends()
         $this->assertSame($extends, $plugin->extends());
     }
 
+    /**
+     * @covers ::get
+     * @covers ::register
```
lukasbestle: It uses `register`, but it shouldn't cover it as its behavior is not tested in this method:

```suggestion
```

src/Cms/Plugin.php
```diff
@@ -173,6 +184,18 @@ public function option(string $key)
         return $this->kirby()->option($this->prefix() . '.' . $key);
     }
 
+    /**
+     * Kirby plugin register
```
lukasbestle:
```suggestion
     * Registers a plugin in the CMS
```

lukasbestle (approved):

lukasbestle:
Leaving the final review to @distantnative because of the open question whether we need/want the methods.

distantnative:
Idk. Don't want to block this but also not really seeing the extra benefit here. Why not just keep using `Kirby::plugin()`?

lukasbestle:
I agree. OK, then @bastianallgeier should decide whether to merge or close this. :)

bastianallgeier:
Is this just about not needing to add `[]` as second argument to `Kirby::plugin()` ?

afbora:
In particular, yes 🙈 
In general, to clear the confusion by distinguishing between getter and setter.

bastianallgeier:
While I appreciate the effort, I think it's not really needed to be honest.

afbora:
It's okay 👍 Thanks for feedback.

PR #4452, opened 2022-06-30
Use `match` instead of `switch`

distantnative:
## This PR …

### Refactored
- Using PHP `match` statements instead of `switch` in various places

## Ready?

### For review team

### DIFF
--- a/src/Cms/Api.php
+++ b/src/Cms/Api.php
@@ -142,20 +142,14 @@ class Api extends BaseApi
 	{
 		$parent = $parentId === null ? $this->site() : $this->page($parentId);
 
-		switch ($status) {
-			case 'all':
-				return $parent->childrenAndDrafts();
-			case 'draft':
-			case 'drafts':
-				return $parent->drafts();
-			case 'listed':
-				return $parent->children()->listed();
-			case 'unlisted':
-				return $parent->children()->unlisted();
-			case 'published':
-			default:
-				return $parent->children();
-		}
+		return match ($status) {
+			'all'             => $parent->childrenAndDrafts(),
+			'draft', 'drafts' => $parent->drafts(),
+			'listed'          => $parent->children()->listed(),
+			'unlisted'        => $parent->children()->unlisted(),
+			'published'       => $parent->children(),
+			default           => $parent->children()
+		};
 	}
 
 	/**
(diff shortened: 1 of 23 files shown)


src/Cms/Api.php
```diff
@@ -142,20 +142,14 @@ public function pages(string $parentId = null, string $status = null)
...
-            case 'published':
-            default:
-                return $parent->children();
-        }
+        return match ($status) {
+            'all'             => $parent->childrenAndDrafts(),
+            'draft', 'drafts' => $parent->drafts(),
+            'listed'          => $parent->children()->listed(),
+            'unlisted'        => $parent->children()->unlisted(),
+            'published'       => $parent->children(),
```
distantnative: Actually we could remove `'published'` here and just use `default`

src/Database/Query.php
```diff
@@ -522,41 +522,37 @@ public function build(string $type)
...
+				'table'    => $this->table,
+				'values'   => $this->values,
+				'bindings' => $this->bindings
+			]),
+			'delete' => $sql->delete([
+				'table'    => $this->table,
+				'where'    => $this->where,
+				'bindings' => $this->bindings
+			])
+		};
```
afbora: Just for a reminder. FIY, will throw `UnhandledMatchError` when not matched, instead of `null`.  Should we do something?

distantnative: True, so the non-breaking way would probably be to add a last line `default => null` inside `match`, right?

afbora: It seems so. For 3.8, we can add todo to throw exception.

distantnative: Since this will be for 3.8 anyways, maybe let's include it directly here. Just need to make sure to write down the breaking change in the notes.

afbora: Oh, right 😄

src/Filesystem/IsFile.php
```diff
@@ -105,12 +105,10 @@ public function asset($props = null)
...
-		switch ($this->type()) {
-			case 'image':
-				return $this->asset = new Image($props);
-			default:
-				return $this->asset = new File($props);
-		}
+		return match ($this->type()) {
+			'image' => $this->asset = new Image($props),
+			default => $this->asset = new File($props)
+		};
```
lukasbestle:
```suggestion
		return $this->asset = match ($this->type()) {
			'image' => new Image($props),
			default => new File($props)
		};
```

afbora (approved):
LGTM 👍

PR #4454, opened 2022-06-30
Use tabs instead of spaces

distantnative:
## This PR …

I would suggest to update the config settings in this PR, update our local editors... but only update the actual files as we go along (would be a huge PR otherwise).

### Refactored
- Using tabs instead of spaces - reasoning: https://twitter.com/Rich_Harris/status/1541761871585464323

## Ready?

### For review team


lukasbestle:
Probably a good move. Accessibility is one of the most important factors in a decision like this. Most other arguments pro and contra tabs vs spaces are stupid in comparison.

There are only two potential issues we would need to check first:

- It would be a violation of [PSR-12](https://www.php-fig.org/psr/psr-12/#24-indenting). So I believe PHP-CS-Fixer will correct the indentation back every time. We need to check if PHP-CS-Fixer can be configured to use tabs. We would still not conform to the standard, but I guess the standard needs to change then, not us. ;)
- In the past I had issues with lines that were manually indented to match previous or following lines. We need to check if we have any of those in our code. Rough example of what I mean:

```php
    $thisIsAVariable = 'With a value that spans over multiple lines.' .
                       'This would break with tabs because it would require mixed indentation.';
```

Also not sure if we should update the files over time. We would have to disable all linting with Prettier and PHP-CS-Fixer. Also it will litter all PRs with whitespace changes that distract when reviewing the PR.

distantnative:
Apparently there is `->setIndent("\t")` for PHP-CS-Fixer

bastianallgeier (dismissed):
Nice!

lukasbestle (changes requested):
I think the XML files (PHPMD, PHPUnit, Psalm) use twice the amount of tabs as would be necessary. E.g. the first level uses two tabs.

The PHPMD file is also an example for what I meant with the "extra indentation". The attributes for the root tag `<ruleset>` are indented to match the previous lines. I think these should just be one attribute per line without the artificial breaks in the attr values. And then all attributes just indented with one tab.

lukasbestle (approved):

PR #4455, opened 2022-06-30
Use null-safe operator from PHP 8.0

distantnative:
Making use of the null safe operator from PHP 8.0.

### Refactored
- Using PHP's new null safe operator 

## Ready?

### For review team

### DIFF
--- a/config/api/models/System.php
+++ b/config/api/models/System.php
@@ -32,28 +32,23 @@ return [
 		'slugs'        => fn () => Str::$language,
 		'title'        => fn () => $this->site()->title()->value(),
 		'translation' => function () {
-			if ($user = $this->user()) {
-				$translationCode = $user->language();
-			} else {
-				$translationCode = $this->kirby()->panelLanguage();
+			$code   = $this->user()?->language();
+			$code ??= $this->kirby()->panelLanguage();
+
+			if ($translation = $this->kirby()->translation($code)) {
+				return $translation;
 			}
 
-			if ($translation = $this->kirby()->translation($translationCode)) {
-				return $translation;
-			} else {
-				return $this->kirby()->translation('en');
-			}
+			return $this->kirby()->translation('en');
 		},
 		'kirbytext' => fn () => $this->kirby()->option('panel.kirbytext') ?? true,
 		'user' => fn () => $this->user(),
 		'version' => function () {
-			$user = $this->user();
-
-			if ($user && $user->role()->permissions()->for('access', 'system') === true) {
+			if ($this->user()?->role()->permissions()->for('access', 'system') === true) {
 				return $this->kirby()->version();
-			} else {
-				return null;
 			}
+
+			return null;
 		}
 	],
 	'type'   => 'Kirby\Cms\System',
--- a/config/api/models/System.php
+++ b/config/api/models/System.php
@@ -32,8 +32,8 @@ return [
 		'slugs'        => fn () => Str::$language,
 		'title'        => fn () => $this->site()->title()->value(),
 		'translation' => function () {
-			$code   = $this->user()?->language();
-			$code ??= $this->kirby()->panelLanguage();
+			$code = $this->user()?->language() ??
+					$this->kirby()->panelLanguage();
 
 			if ($translation = $this->kirby()->translation($code)) {
 				return $translation;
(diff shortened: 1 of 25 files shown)


config/api/models/System.php
```diff
@@ -32,28 +32,23 @@
 		'slugs'        => fn () => Str::$language,
 		'title'        => fn () => $this->site()->title()->value(),
 		'translation' => function () {
-			if ($user = $this->user()) {
-				$translationCode = $user->language();
-			} else {
-				$translationCode = $this->kirby()->panelLanguage();
-			}
+			$code   = $this->user()?->language();
+			$code ??= $this->kirby()->panelLanguage();
```
lukasbestle: To be honest I think it's a step backwards here. It's much shorter, but if I only looked at the new code, I'd need a lot more time to parse the functionality mentally. Like: "How can `$code` ever be null in the second line? Ah, because of the null-safe operator. But what does this mean for the execution of the second line? Ah, only when `user()` returns `null`." Whereas the old code is really explicit about the two different cases.

config/tags.php
```diff
@@ -117,11 +117,8 @@
 					return $img;
 				}
 
-				if ($link = $tag->file($tag->link)) {
-					$link = $link->url();
-				} else {
-					$link = $tag->link === 'self' ? $tag->src : $tag->link;
-				}
+				$link   = $tag->file($tag->link)?->url();
+				$link ??= $tag->link === 'self' ? $tag->src : $tag->link;
```
lukasbestle, line 121: Same here.

src/Cms/Api.php
```diff
@@ -39,11 +39,9 @@ public function call(string $path = null, string $method = 'GET', array $request
...
 
 		$allowImpersonation = $this->kirby()->option('api.allowImpersonation', false);
-		if ($user = $this->kirby->user(null, $allowImpersonation)) {
-			$translation = $user->language();
-		} else {
-			$translation = $this->kirby->panelLanguage();
-		}
+
+		$translation   = $this->kirby->user(null, $allowImpersonation)?->language();
+		$translation ??= $this->kirby->panelLanguage();
```
lukasbestle, line 44: And here.

lukasbestle (commented):
I love the simplicity the new operator brings for all the "if something then return something->method()" use cases.

Not sure about the more complex cases though. I could get used to it, but at the moment I feel like the slightly shorter code is not worth it for the added mental complexity when reading the code.

config/api/models/System.php
distantnative: Very interesting. For me it's literally exact the opposite: if-else always requires so much more brain capacity for me as my mind needs to keep track of two parallel contexts that keep developing and seeing how to merge them together afterwards. Whereas the usage of `??=` turns it into a one-dimensional order: "if it's still null, then fill it with the following". Reducing for my mind a lot of complexity. But I see your argument, would just go the way the majority of the team feels best with.

bastianallgeier: Hm, I'm a bit torn between the two. I agree with @distantnative that the if clauses are always kind of an obstacle when reading through code. 

But the lines would probably feel better to me as a single expression: 

```php
$code = $this->user()?->language() ?? $this->kirby()->panelLanguage();
```

But then it's quite long. I think I don't any of the three options :)

texnixe: I think all in one long line is still better than the two lines with the `??=` operator if we really want to get rid of the if statements.

lukasbestle: I like the one-liner. If it gets too long, we could also do:

```php
$code = $this->user()?->language() ??
    $this->kirby()->panelLanguage();
```

config/tags.php
distantnative: @bastianallgeier @texnixe @lukasbestle While I can see in the first case that a one-liner makes sense, here that would get really hard to read. I really think the `??=` syntax makes a lot of sense here, maybe just something for you to get used to as `??=` is rather new to see for us?

lukasbestle: > I really think the `??=` syntax makes a lot of sense here, maybe just something for you to get used to as `??=` is rather new to see for us?

I agree. 👍

config/api/models/System.php
afbora: My vote is for one liner too 👍

lukasbestle (approved):

afbora (approved):
All LGTM 👍

PR #4458, opened 2022-06-30
No longer return `$this` on Collection `__set()` method

neildaniels:
## Explanation
This PR changes the behavior of `__set()` methods to no longer return `$this`. Instead there is no return value.

A return type was never formalized in code, but was defined in the DocBlock for `Kirby\Toolkit\Collection`.

Although `Kirby\Toolkit\Collection` returns `$this`, the subclass `Kirby\Cms\Collection` **does not** return anything.

https://github.com/getkirby/kirby/blob/e15a4eacd7190e6088f2087507989892c9948519/src/Cms/Collection.php#L67-L78

The `Kirby\Cms\Pages` class extends `Kirby\Cms\Collection` and does not override `__set()` so it therefor also returns nothing.

`Kirby\Cms\Structure` also extends `Kirby\Cms\Collection`. It attempts to return a value in its reimplementation of `__set()`, but because it's returning `parent::__set()` and the `Kirby\Cms\Collection` implementation does not return anything, it's effectively returning void.

The [formal definition](https://www.php.net/manual/en/language.oop5.overloading.php#object.set) of the magic method `__set()` has a `void` return type.

```
public __set(string $name, mixed $value): void
```

Additionally, the PHP Docs also say that a return a type on `__set()` is always ignored. None of their examples return a value either.

> Note:
The return value of [__set()](https://www.php.net/manual/en/language.oop5.overloading.php#object.set) is ignored because of the way PHP processes the assignment operator. Similarly, [__get()](https://www.php.net/manual/en/language.oop5.overloading.php#object.get) is never called when chaining assignments together like this:
 $a = $obj->b = 8;

### Refactoring

- The `$collection->__set()` methods consistently no longer return the class instance to match PHP's definition for this method.

### Breaking changes
This is technically a breaking change: the return type is ignored, but only for usage such as `$collection->foo = $example;`. Usage by calling the function literally still would return a value `$collection->__set('foo', $example);`.

However, the impact of this breaking change should be minimal, considering that `Kirby\Cms\Collection` never returned a value anyways.

## Ready?

### For review team

### DIFF
--- a/src/Cms/Collection.php
+++ b/src/Cms/Collection.php
@@ -71,8 +71,9 @@ class Collection extends BaseCollection
 	 *
 	 * @param string $id
 	 * @param object $object
+	 * @return void
 	 */
-	public function __set(string $id, $object)
+	public function __set(string $id, $object): void
 	{
 		$this->data[$id] = $object;
 	}
--- a/src/Cms/Structure.php
+++ b/src/Cms/Structure.php
@@ -40,9 +40,11 @@ class Structure extends Collection
 	 *
 	 * @param string $id
 	 * @param array|StructureObject $props
+	 * @return void
+	 *
 	 * @throws \Kirby\Exception\InvalidArgumentException
 	 */
-	public function __set(string $id, $props)
+	public function __set(string $id, $props): void
 	{
 		if (is_a($props, 'Kirby\Cms\StructureObject') === true) {
 			$object = $props;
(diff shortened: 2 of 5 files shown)


distantnative:
@lukasbestle what do you think? maybe this could go into 3.7.1 since it is more like fixing a bug than breaking change?

lukasbestle (approved):

lukasbestle:
I've added `void` type hints for consistency with the formal definition.

distantnative (approved):

PR #4459, opened 2022-06-30
Conditionally define `DS` constant

neildaniels:
## This PR …

This change conditionally defines the `DS` constant only if it hasn't already been defined.

I'm not sure _why_ this constant is still being defined—it seems completely unused. Should this definition just be removed altogether?

If its still important to keep, this change would help me avoid a constant redeclaration error that's coming from a separate library.

### Breaking changes
None.

## Ready?

### For review team

### DIFF
--- a/config/setup.php
+++ b/config/setup.php
@@ -3,7 +3,9 @@
 /**
  * Constants
  */
-define('DS', '/');
+if (!defined('DS')) {
+	define('DS', '/');
+}
 
 /**
  * Class aliases


lukasbestle:
> I'm not sure _why_ this constant is still being defined—it seems completely used. Should this definition just be removed altogether?

I agree. I'd much rather remove the constant completely. Conditionally defining it can cause issues when a user expects the constant to have a specific value, but then it's suddenly defined somewhere else.

In Kirby 2 we used this constant extensively as an alias for `DIRECTORY_SEPARATOR`. Looks like we kept it in Kirby 3 for backwards-compatibility of plugin code.

I think we could remove the constant in 3.8 and document it as a breaking change. Shouldn't affect many users.

neildaniels:
Created an alternative PR that just removes the constant completely: #4463.

distantnative:
I'd vote for removal in 3.8 - @lukasbestle?

PR #4468, opened 2022-07-03
Simplify some if-else statements

distantnative:

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -647,15 +647,7 @@ class App
 		}
 
 		if ($id === '.') {
-			if ($file = $parent->file($filename)) {
-				return $file;
-			}
-
-			if ($file = $this->site()->file($filename)) {
-				return $file;
-			}
-
-			return null;
+			return $parent->file($filename) ?? $this->site()->file($filename);
 		}
 
 		if ($page = $this->page($id, $parent, $drafts)) {
@@ -725,9 +717,9 @@ class App
 		if ($instance === null) {
 			if ($lazy === true) {
 				return static::$instance;
-			} else {
-				return static::$instance ?? new static();
 			}
+
+			return static::$instance ?? new static();
 		}
 
 		return static::$instance = $instance;
@@ -807,10 +799,11 @@ class App
 				return $this->io($e);
 			}
 
-			if ($input->isErrorPage() === true) {
-				if ($response->code() === null) {
-					$response->code(404);
-				}
+			if (
+				$input->isErrorPage() === true &&
+				$response->code() === null
+			) {
+				$response->code(404);
 			}
 
 			return $response->send($html);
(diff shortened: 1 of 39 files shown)


src/Cms/App.php
```diff
@@ -725,9 +717,9 @@ public static function instance(self $instance = null, bool $lazy = false)
...
 			if ($lazy === true) {
 				return static::$instance;
-			} else {
-				return static::$instance ?? new static();
 			}
+
+			return static::$instance ?? new static();
 		}
 
 		return static::$instance = $instance;
```
texnixe: We could get rid of the nested if here by inverting the logic?

```
       if ($instance !== null) {
            return static::$instance = $instance;
        }

        if ($lazy === true) {
            return static::$instance;
        }

        return static::$instance ?? new static();
```

distantnative: Yes!

afbora (approved):
Looks good to me, now really more simple and easy to read the codes 👍

src/Cms/FilePicker.php
```diff
@@ -55,15 +55,14 @@ public function items()
...
-		if (is_a($files, 'Kirby\Cms\Site') === true) {
-			$files = $files->files();
-		} elseif (is_a($files, 'Kirby\Cms\Page') === true) {
-			$files = $files->files();
-		} elseif (is_a($files, 'Kirby\Cms\User') === true) {
-			$files = $files->files();
-		} elseif (is_a($files, 'Kirby\Cms\Files') === false) {
-			throw new InvalidArgumentException('Your query must return a set of files');
-		}
+		$files = match ($files::class) {
```
lukasbestle, line 58: What if an object gets passed that extends these classes, e.g. a custom page model? I think the new check will fail in this case.

src/Cms/PagePicker.php
```diff
@@ -200,13 +200,12 @@ public function itemsForQuery()
...
 		// by converting site and page objects to proper
 		// pages by returning their children
-		if (is_a($items, 'Kirby\Cms\Site') === true) {
-			$items = $items->children();
-		} elseif (is_a($items, 'Kirby\Cms\Page') === true) {
-			$items = $items->children();
-		} elseif (is_a($items, 'Kirby\Cms\Pages') === false) {
-			throw new InvalidArgumentException('Your query must return a set of pages');
-		}
+		$items = match ($items::class) {
```
lukasbestle, line 203: Same here.

src/Cms/FilePicker.php
distantnative: I think you're right! Probably then there is no good way to work with `::class` and `match`

distantnative: From what I've read, this could work

```php
$files = match (true) {
    $files instanceof Site,
    $files instanceof Page,
    $files instanceof User  => $files->files(),
    $files instanceof Files => $files,

    default => throw new InvalidArgumentException('Your query must return a set of files')
};
```

What do you think?

distantnative: https://github.com/getkirby/kirby/pull/4488

PR #4469, opened 2022-07-03
Add dependabot config

distantnative:


.github/dependabot.yml 
```diff
@@ -0,0 +1,21 @@
+version: 1
```
lukasbestle:
```suggestion
version: 2
```

.github/dependabot.yml 
```diff
@@ -0,0 +1,21 @@
...
+updates:
+  - package-ecosystem: "composer"
+    directory: "/"
```
lukasbestle, line 4: To commit the vendor files:

```suggestion
  - package-ecosystem: "composer"
    directory: "/"
    vendor: true
```

lukasbestle (dismissed):

neildaniels:
This is awesome!

I would recommend increasing the open PR limit to something higher than the default of 5—especially if this is only going to run at a weekly cadence.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

distantnative:
@neildaniels good thinking - didn't catch that default, especially since they spell five out :D

afbora:
> error: invalid path '.github/dependabot.yml '

@distantnative `dependabot.yml` filename ending with space and I can't fetch the develop branch. Could you fix that?

https://github.com/getkirby/kirby/blob/develop/.github/dependabot.yml%20

distantnative:
@afbora done 🙏

Issue #4443, opened 2022-06-29
[3.7.0] Panel blank on cloudflare


bogdancondorachi:
## Description
Dating back to this 3.6 issue #4234 I've implemented an '$_SERVER['HTTPS'] = true;' to my production to fix that. Now I've updated to 3.7 and deleted that code to check if it works without. I don't get the same console error as before (actually my console is empty now) but my panel still appears blank for some reason. I've added the correct 'url' in config and also deleted media folder but with no success.

**Console output** 
This error appeared at some point in the console but has recently been empty.

`Mixed Content: The page at 'https://hasty.ai/panel/site' was loaded over HTTPS, but requested an insecure resource 'http://hasty.ai/api/auth'. This request has been blocked; the content must be served over HTTPS.`

lukasbestle:
What is your output of `var_dump($kirby->environment()->toArray())` in any template?

bogdancondorachi:
> 

Here's what I get
```
[...]

PR #4475, opened 2022-07-04
`Environment`: Better reverse proxy URL detection

lukasbestle:
### Enhancements

- Kirby now detects reverse proxy URLs from the standardized `Forwarded` header
- If a reverse proxy is used but does not pass the `X-Forwarded-Host` header or `host` key in the `Forwarded` header, Kirby falls back to the server name; this improves compatibility with CloudFlare #4443

## Notes

I couldn't get the unit tests done today. Do you think it's OK if I add them after tomorrow (after the RC release but before the public release)? Then this enhancement doesn't have to wait for two more weeks.

It's between an enhancement and a regression fix as the old code before 3.7 seems to have worked better in some setups, e.g. on CloudFlare. This PR should restore that behavior.

## Ready?

- [x] Tests and checks all pass (apart from Codecov)

### For review team

### DIFF
--- a/src/Http/Environment.php
+++ b/src/Http/Environment.php
@@ -332,17 +332,25 @@ class Environment
 	protected function detectAuto(bool $insecure = false): void
 	{
 		// proxy server setup
-		if (
-			$insecure === true &&
-			empty($this->info['HTTP_X_FORWARDED_HOST']) === false
-		) {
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):
LGTM. I think it's fine to add the unit tests during the RC phase.

PR #4479, opened 2022-07-05
More `switch` to `match`

afbora:


afbora:
> Error: Unexpected token: }, line: 325, col: 3, file: /home/runner/work/kirby/kirby/src/Cms/Auth.php.

What's wrong with PHPMD? Do you understand what?

src/Cms/Auth.php
```diff
@@ -308,28 +308,21 @@ public function impersonate(string|null $who = null)
...
-					'role'  => 'nobody',
-				]);
-			default:
-				if ($user = $this->kirby->users()->find($who)) {
-					return $this->impersonate = $user;
-				}
-
-				throw new NotFoundException('The user "' . $who . '" cannot be found');
-		}
+		return match ($who) {
```
distantnative: Since all cases assign `$this->impersonate = ` we could pull this to the front of the statement and to `return $this->impersonate = match(`, I think.

distantnative:
Only the one comment, rest looks great!

src/Cms/Auth.php
distantnative: Could also be what phpmd is causing to fail?

afbora: Oh you're right. Could you done this change if you're available please?

distantnative:
@lukasbestle do you know why PHPMD is failing here?

src/Cms/Auth.php
```diff
@@ -308,28 +308,20 @@ public function impersonate(string|null $who = null)
...
+				'id'    => 'kirby',
+				'role'  => 'admin',
+			]),
+			'nobody' => new User([
+				'email' => 'nobody@getkirby.com',
+				'id'    => 'nobody',
+				'role'  => 'nobody',
+			]),
+			default  => $this->kirby->users()->find($who) ?? throw new NotFoundException('The user "' . $who . '" cannot be found')
+		};
```
lukasbestle: PHPMD uses the PDepend library, which comes with its own PHP parser. I assume the false-positive error comes from a spin-off of this issue: https://github.com/pdepend/pdepend/pull/544

We could try this alternatively:

```suggestion
		$this->impersonate = match ($who) {
			null     => null,
			'kirby'  => new User([
				'email' => 'kirby@getkirby.com',
				'id'    => 'kirby',
				'role'  => 'admin',
			]),
			'nobody' => new User([
				'email' => 'nobody@getkirby.com',
				'id'    => 'nobody',
				'role'  => 'nobody',
			]),
			default  => $this->kirby->users()->find($who)
		};
		
		return $this->impersonate ?? throw new NotFoundException('The user "' . $who . '" cannot be found');
```

lukasbestle: Wait, this won't work because of the other cases. Maybe it's already enough to wrap the return value of the default case with `(...)`.

PR #4488, opened 2022-07-11
Fix `match` for model objects

distantnative:


lukasbestle (approved):
Very awesome solution! 👍

PR #4495, opened 2022-07-12
`Toolkit\A`: improve type hinting

distantnative:


bastianallgeier (approved):

src/Toolkit/A.php
```diff
@@ -141,12 +145,14 @@ public static function get($array, $key, $default = null)
...
-	 * @param mixed $separator
+	 * @param array|string $value
+	 * @param array|string $separator
 	 * @return string
 	 */
-	public static function join($value, $separator = ', ')
-	{
+	public static function join(
+		array|string $value,
+		array|string $separator = ', '
```
lukasbestle, line 154: Can `$separator` be an array?

afbora: Lukas right. `$separator` should be string. If `$separator` was the first argument then it could be array but we are using the reverse order of PHP native.

````
# PHP
implode(string $separator, array $array): string

# KIRBY
join($value, $separator): string
````

PR #4497, opened 2022-07-12
UUIDs

distantnative:
WIP, but PR is helpful to see test coverage etc

- [x] Unit tests for cache
- [x] fix broken unit tests in other parts
- [x] full test coverage for `Binding` class
- [x] integrate into lookup methods
- [x] integrate into methods that store references (e.g. picker fields, drag text ...)
- [ ] defaults for `cache.uuid` option
- [ ] more unit tests, e.g. for kirbytags and `->toPages()` etc
- [ ] language support for UUID schema protocol

Requires
- [x] https://github.com/getkirby/kirby/pull/4499
- [ ] https://github.com/getkirby/kirby/pull/4500

### DIFF
--- a/src/Cms/HasFiles.php
+++ b/src/Cms/HasFiles.php
@@ -90,6 +90,11 @@ trait HasFiles
 			return $this->$in()->first();
 		}
 
+		// find by global UUID
+		if (Uuid::is($filename, 'file') === true) {
+			return Uuid::for($filename)->toModel();
+		}
+
 		if (strpos($filename, '/') !== false) {
 			$path     = dirname($filename);
 			$filename = basename($filename);
--- a/config/fields/files.php
+++ b/config/fields/files.php
@@ -128,7 +128,7 @@ return [
 		];
 	},
 	'save' => function ($value = null) {
-		return A::pluck($value, 'uuid');
+		return A::pluck($value, 'id');
 	},
 	'validations' => [
 		'max',
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -638,6 +638,11 @@ class App
 		$id       = dirname($path);
 		$filename = basename($path);
 
+		// find by global UUID
+		if (Uuid::is($path, 'file') === true) {
+			return Uuid::for($path)->toModel();
+		}
+
 		if (is_a($parent, 'Kirby\Cms\User') === true) {
 			return $parent->file($filename);
 		}
@@ -1172,6 +1177,11 @@ class App
 	{
 		if ($id === null) {
 			return null;
+		}
+
+		// find by global UUID
+		if (Uuid::is($id, 'page') === true) {
+			return Uuid::for($id)->toModel();
 		}
 
 		$parent = $parent ?? $this->site();
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -52,12 +52,7 @@ trait FileActions
 				throw new LogicException('The new file exists and cannot be overwritten');
 			}
 
-			// remove the lock of the old file
-			if ($lock = $oldFile->lock()) {
-				$lock->remove();
-			}
-
-			// remove all public versions
+			// remove all public versions, lock and clear UUID cache
 			$oldFile->unpublish();
 
 			// rename the main file
@@ -75,6 +70,7 @@ trait FileActions
 				F::move($oldFile->contentFile(), $newFile->contentFile());
 			}
 
+			// update collections
 			$newFile->parent()->files()->remove($oldFile->id());
 			$newFile->parent()->files()->set($newFile->id(), $newFile);
 
@@ -195,7 +191,7 @@ trait FileActions
 		// run the hook
 		return $file->commit('create', compact('file', 'upload'), function ($file, $upload) {
 
-			// delete all public versions
+			// remove all public versions, lock and clear UUID cache
 			$file->unpublish();
 
 			// overwrite the original
@@ -216,6 +212,9 @@ trait FileActions
 			// add the file to the list of siblings
 			$file->siblings()->append($file->id(), $file);
 
+			// create UUID
+			Uuid::for($file);
+
 			// return a fresh clone
 			return $file->clone();
 		});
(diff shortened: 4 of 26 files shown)


src/Cms/BindingProtocol.php
```diff
@@ -0,0 +1,119 @@
...
+	/**
+	 * supported schemes
+	 */
+	protected static array $schemes = [
+		'site',
+		'page',
+		'file',
+		'user',
+		'block',
+		'structure'
```
bastianallgeier: Maybe stupid, but what about calling it struc://. Then we have about the same length as the rest and it's still easy to read.

src/Cms/HasFiles.php
```diff
@@ -90,6 +92,11 @@ public function file(string $filename = null, string $in = 'files')
 			return $this->$in()->first();
 		}
 
+		// find by global UUID
+		if (Str::contains($filename, 'file://') === true) {
```
bastianallgeier: Shouldn't we use Str::startsWith here? Or maybe even have a global method that simplifies this.

src/Cms/Binding.php
```diff
@@ -0,0 +1,417 @@
...
+	public function id(): string
+	{
+		// if no ID hash (protocol's host) exists yet,
+		// generate a new one and update model as well as protocol
+		// (should only be entered for pages or files)
+		if ($this->protocol->host() === null) {
+			$this->create();
+		}
+
+		return $this->protocol->tostring(false);
```
bastianallgeier:
```suggestion
		return $this->protocol->toString(false);
```

src/Cms/Binding.php
```diff
@@ -0,0 +1,417 @@
...
+	 * Clears UUID from cache
+	 *
+	 * @param bool $recursive also clear subpages from cache
+	 * @return bool
+	 */
+	public function clear(bool $recursive = false): bool
+	{
+		// For pages: if $recursive, also clear UUIDs
+		// from cache for all children recursively
+		if ($this->type() == 'page' && $recursive === true) {
```
bastianallgeier:
```suggestion
		if ($this->type() === 'page' && $recursive === true) {
```

src/Cms/Binding.php
```diff
@@ -0,0 +1,417 @@
...
+	{
+		$id = Str::uuid();
+
+		// TODO: updating block/structure would work different
+
+		// update content file with generated id:
+		// make sure Kirby has the required permissions
+		// for the update action
+		$kirby = App::instance();
+		$user  = $kirby->auth()->currentUserFromImpersonation();
```
bastianallgeier: Can't we just use …

```suggestion
		$user  = $kirby->user();
```

src/Cms/Binding.php
```diff
@@ -0,0 +1,417 @@
...
+		$kirby = App::instance();
+		$site  = $kirby->site();
+		$index = $site->index(true);
+
+		// pages
+		foreach ($index as $page) {
+			static::for($page)->populate();
+		}
+
+		// page files
```
bastianallgeier: I'm not sure if it makes a huge difference, but it feels better to me to do this in a nested foreach loop of the page loop above.

src/Cms/Binding.php
```diff
@@ -0,0 +1,417 @@
...
+	 */
+	public function key(): string
+	{
+		$id = $this->protocol->toString(false);
+
+		// for better performance when using a file-based cache,
+		// turn first two characters of the id into a directory
+		$id = substr($id, 0, 2) . '/' . substr($id, 2);
+
+		// TODO: if we also do blocks and structure fields, maybe good to
```
bastianallgeier: I really like the idea to use the type as first level

src/Cms/Binding.php
```diff
@@ -0,0 +1,417 @@
...
+			return $this->model;
+		}
+
+		return null;
+	}
+
+	/**
+	 * Returns the model type
+	 *
+	 * @return string
```
bastianallgeier: I think we should start removing code comments that are already implied by type hinting. 
There are couple more examples in this file. 

```suggestion
```

src/Cms/Binding.php
```diff
@@ -0,0 +1,417 @@
...
+	{
+		if (
+			$this->type() === 'site' ||
+			$this->type() === 'page' ||
+			$this->type() === 'user'
+		) {
+			return $this->model->id();
+		}
+
+		if ($this->type() === 'file') {
```
bastianallgeier: Shouldn't the $file->id() method also always return the full id including the parent id?

src/Cms/HasFiles.php
distantnative: Yea, probably something like `Binding::isUuid(string|null $type = null)`

src/Cms/Binding.php
distantnative: But we don't want the full id with parent id, we want parent uuid + filename, e.g. `page://page-uuid/filename.jpg` instead of just `page-a/filename.jpg`

src/Cms/BindingProtocol.php
distantnative: If, then I'd be for at least `struct://`

src/Cms/Binding.php
distantnative: Not sure if that's always the same, e.g. when code around already is impersonating anyone. It's exactly what `$kirby->impersonate()` does. In 3.9 we will be bale to streamline this bit of code, once the callback for `$kirby->impersonate()` doesn't bind the app instance anymore.

src/Cms/HasFiles.php
bastianallgeier: Exactly!

src/Cms/Binding.php
bastianallgeier: Ah, you are right. You could be impersonating someone and still have another logged in user. Right. I got you. Then we need to keep it like this until 3.9

src/Cms/Binding.php
bastianallgeier: Forget my comment. I got confused by what value does.

PR #4498, opened 2022-07-12
`App::impersonate()`: don’t bind callback to `App`

distantnative:
## This PR …

No idea why we so far have bound the callback to `App` - to me it makes a lot more sense to leave the callback in its actual `$this` context

### Deprecated
- `$kirby->impersonate($user, $callback)`: the `$callback` will not be bound anymore to the `$kirby` instance in Kirby 3.9.0, `$this` inside the callback will refer to the current context and not `$kirby` instead. Using `$this` as `$kirby` inside the callback will throw a deprecation warning in debug mode.

## Ready?

### For review team

### DIFF
--- a/src/Cms/AppUsers.php
+++ b/src/Cms/AppUsers.php
@@ -60,8 +60,7 @@ trait AppUsers
 		}
 
 		try {
-			// bind the App object to the callback
-			return $callback->call($this, $userAfter);
+			return $callback($userAfter);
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):
I agree! We should get rid of those few remaining mystery $this bindings wherever we can. They are more harmful than they actually solve anything.

PR #4500, opened 2022-07-12
Abstract `Models` class

distantnative:
### Enhancements
- New abstract `Cms\Models` class (extending `Cms\Collection`) as basis for `Cms\Pages`, `Cms\Files` and `Cms\Users`

## Ready?

### For review team

### DIFF
--- a/src/Cms/Files.php
+++ b/src/Cms/Files.php
@@ -19,15 +19,8 @@ use Kirby\Filesystem\F;
  * @copyright Bastian Allgeier
  * @license   https://getkirby.com/license
  */
-class Files extends Collection
+class Files extends Models
 {
-	/**
-	 * All registered files methods
-	 *
-	 * @var array
-	 */
-	public static $methods = [];
-
 	/**
 	 * Adds a single file or
 	 * an entire second collection to the
@@ -44,7 +37,10 @@ class Files extends Collection
 			$this->data = array_merge($this->data, $object->data);
 
 		// add a file by id
-		} elseif (is_string($object) === true && $file = App::instance()->file($object)) {
+		} elseif (
+			is_string($object) === true &&
+			$file = App::instance()->file($object)
+		) {
 			$this->__set($file->id(), $file);
 
 		// add a file object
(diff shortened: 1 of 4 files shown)


lukasbestle:
We also have `Cms\Languages`, which is also a collection of models, but doesn't have support for custom methods. With the single `Model` objects we define the `$methods` property directly in the respective class for this reason. TBH I feel it gets a bit inconsistent.

distantnative:
A more consistent name would probably be `ModelsWithContent` in that way (collection of `ModelWithContent` objects).

lukasbestle:
I like that!

BTW: The custom method-related logic is defined in the `HasMethods` trait. This is also already used by `Cms\Collection`. I believe the `$methods` prop is only set again in the individual models and collections to keep them separate arrays (avoids sharing with the parent object).

So IMO the question is: What do we need the `Models`/`ModelsWithContent` class for in UUIDs? If it's needed to detect if a collection has `$methods`, we can already check for `Cms\Collection`.

distantnative:
To have a place for the `::findByUuid()` method.

https://github.com/getkirby/kirby/pull/4506/files

lukasbestle:
Ah, I see. That makes sense.

What about a `trait HasUuids` instead? Then this can be imported into all model collections that have UUIDs. If we make a whole class, we would reduce our flexibility in the future.

distantnative:
Could be a way. Though I think in general it would make sense to bind pages, files and users collections closer together through a dedicated class.

lukasbestle:
> Though I think in general it would make sense to bind pages, files and users collections closer together through a dedicated class.

It might make sense at some point, but TBH I think it's theoretical at the moment. If we only want to define a single method, I don't see a use case for a whole additional inheritance layer. What happens if we extend that `Models` class with additional methods later but then want to add UUID functionality to collections where these added methods don't fit? That's why I think a trait would be more modular and future-proof.

distantnative:
I'll close this for now - we can revisit it in the context of UUID PR.

PR #4501, opened 2022-07-12
Consistent page path/id methods

### LINKED ISSUE #4489: Consistent page path/id methods
Method | Description | Example | Alternative?
-- | -- | -- | --
`$page→dirname()` | full directory name with num and uid | `2_project-b` | `$page→dirname()`
`$page→diruri()` | full path with sorting num und uid | `1_projects/2_project-b` | `$page→dirpath()`
`$page→id()` | path with uid | `projects/project-b` | `$page→path(false)`
`$page→uri()` | path with translated slugs | `projekte/projekt-b` | `$page→path($lang)`
`$page→uid()` | slug but always untranslated | `project-b` | `$page→slug(false)`
`$page→slug()` | folder name or translated slug | `projekt-b` | `$page→slug($lang)`
`$page→uuid()` | UUID with schema | `page://12345678-90ab-cdef-1234-567890abcdef` |  |

We should try to pave the path in Kirby 3.8, meaning
- all alternative methods should be working as intended in the future
- where needed methods (or parameter usage) should be marked as deprecated in the comments/docblocks
- go through code base and alter references to be consistent with new naming
- docs need to be updated everywhere to only show new methods
- real deprecation warnings should only be thrown after docs have been updated for a significant amount of time (6-12 months)
- removing deprecated methods should follow one major version after deprecation notice

distantnative:
First steps based on https://github.com/getkirby/kirby/issues/4489

### DIFF
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -76,7 +76,7 @@ class Page extends ModelWithContent
 	 *
 	 * @var string
 	 */
-	protected $diruri;
+	protected $dirpath;
 
 	/**
 	 * Draft status flag
@@ -84,13 +84,6 @@ class Page extends ModelWithContent
 	 * @var bool
 	 */
 	protected $isDraft;
-
-	/**
-	 * The Page id
-	 *
-	 * @var string
-	 */
-	protected $id;
 
 	/**
 	 * The template, that should be loaded
@@ -118,6 +111,13 @@ class Page extends ModelWithContent
 	 * @var \Kirby\Cms\Page|null
 	 */
 	protected $parent;
+
+	/**
+	 * The page path
+	 *
+	 * @var string
+	 */
+	protected $path;
 
 	/**
 	 * Absolute path to the page directory
@@ -289,7 +289,7 @@ class Page extends ModelWithContent
 	 */
 	protected function cacheId(string $contentType): string
 	{
-		$cacheId = [$this->id()];
+		$cacheId = [$this->path(false)];
 
 		if ($this->kirby()->multilang() === true) {
 			$cacheId[] = $this->kirby()->language()->code();
@@ -384,7 +384,7 @@ class Page extends ModelWithContent
 	 */
 	public function depth(): int
 	{
-		return $this->depth ??= (substr_count($this->id(), '/') + 1);
+		return $this->depth ??= (substr_count($this->path(false), '/') + 1);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/Cms/Page.php
```diff
@@ -953,13 +958,44 @@ public function parents()
...
 	}
 
+	/**
+	 * Returns the page path,
+	 * e.g. `projects/project-b`
+	 *
+	 * @param string|false|null $lang
+	 * @return string
+	 */
+	public function path(string|false|null $lang = null): string
```
bastianallgeier, line 975: To be honest, I'm not very happy how smart the path method is. When I read through the code and see `$this->path(false)` I instantly need to look up why false is being used before I understand the code. I think I would rather prefer an additional method instead.

src/Cms/Page.php
```diff
@@ -1249,21 +1285,25 @@ protected function setUrl(string $url = null)
...
 
 	/**
-	 * Returns the slug of the page
+	 * Returns the (translated) slug of the page
 	 *
-	 * @param string|null $languageCode
+	 * @param string|false|null $lang
 	 * @return string
 	 */
-	public function slug(string $languageCode = null): string
```
bastianallgeier, line 1257: Same here. I would prefer to reduce those toggle arguments and remove the logic in those methods.

lukasbestle:
The goal was to use as few new methods as possible to avoid blocking too many content fields.

I like the language parameters as you can get the same format of value either for the current language, a specific one or the default language. Combining all three cases in one method for each format improves the overview by a lot.

Instead of `null` and `false` we could use constants. The resulting code would be easier to read, but constants are of course longer and not really common in our code base.

Or we could indeed use separate methods. But then we need good names.

Do you have suggestions?

src/Cms/Page.php
distantnative: Is it actually that smart? I think this is more a perception coming from the current very cluttered and confusing set of different named methods. If I look at the table in https://github.com/getkirby/kirby/issues/4489, to me it makes a lot of sense that `$page→id()` and `$page→uri()` both now are `$page->path()` if we look at the examples what they actually return. I'd think your confusion comes rather from the status quo than the proposed change: the new methods would always return the same kind of thing, only either not translated (`false`), specifically translated (`$lang`) or translated but the current language (`null`).

lukasbestle:
@bastianallgeier Would be great to pick this up again for 3.9.0, for that we need your input.

lukasbestle:
@bastianallgeier Could you explain what the plan for this is? Not do it at all, do it later, do it differently?

lukasbestle:
To be reconsidered later.

Comment from Bastian:

> To be honest, this one feels to me like one of those changes that stir up a hornets nest. I was very hesitant about it back then and I still am.

bastianallgeier:
@distantnative do you want to keep it locally? I would close this right now and maybe come back to it later.

distantnative:
Stored as local backup

Issue #3798, opened 2021-10-12
Stuck in an unsaved changes loop


afbora:
Related https://github.com/getkirby/kirby/pull/3758 PR and still continue unicode characters issue.

You can use following content to reproduce the issue:

````
[...]

PR #4504, opened 2022-07-14
First steps code splitting with Vite 3

distantnative:
### Enhancements
- Reduced size of the Panel's vendor JS file by adding code splitting for the vuedraggable.js library

Before
```
dist/css/style.css   106.29 KiB / gzip: 18.02 KiB
dist/js/index.js     303.70 KiB / gzip: 71.99 KiB
dist/js/vendor.js    375.31 KiB / gzip: 120.69 KiB
``` 

After
```
dist/js/vuedraggable.js   40.74 KiB / gzip: 14.12 KiB 
dist/css/style.css        106.29 KiB / gzip: 18.01 KiB
dist/js/index.js          304.26 KiB / gzip: 72.29 KiB
dist/js/vendor.js         333.99 KiB / gzip: 106.81 KiB
```

## Ready?

### For review team


panel/vite.config.js
```diff
@@ -1,6 +1,7 @@
 /* eslint-env node */
 import fs from "fs";
 import path from "path";
+import md5 from "md5";
```
johannschopplich: ~~@distantnative I'd recommend [ohash](https://github.com/unjs/ohash) instead.~~

Why?
- deps-free (opposed to `md5`, which has 3 deps)
- Well tested.
- Used internally by Nuxt 3 (as for all the other UnJS gems).

Edit: I'm probably mistaken. The Panel URL is defined by Kirby's hashing. I see it now. Sorry for the inconvenience.

distantnative: @johannschopplich yea, we kinda need to match what the backend does here. But should be fine since only relevant for development, not in production

distantnative:
@bastianallgeier 
since the Writer component is such a headache with code splitting, how about we at lest merge already the basic setup that splits at least vuedraggable?

```
vite v3.0.0 building for production...
✓ 560 modules transformed.
dist/js/vuedraggable.js   40.74 KiB / gzip: 14.12 KiB
dist/css/style.css        106.25 KiB / gzip: 18.01 KiB
dist/js/index.js          304.20 KiB / gzip: 72.32 KiB
dist/js/vendor.js         333.99 KiB / gzip: 106.84 KiB
```

vs

```
vite v3.0.0 building for production...
✓ 559 modules transformed.
dist/css/style.css   106.25 KiB / gzip: 18.01 KiB
dist/js/index.js     303.61 KiB / gzip: 71.96 KiB
dist/js/vendor.js    375.31 KiB / gzip: 120.69 KiB
```

distantnative:
@afbora did my previous changes cause the problem you just fixed? and we probably should recompile list files then again.

afbora:
Yes, email mark was throwing error because of email mark has not the editor payload.

panel/src/components/Forms/Writer/Writer.vue
```diff
@@ -174,13 +174,13 @@ export default {
...
 				},
 				update: (payload) => {
-					if (!this.editor) {
+					if (!payload.editor) {
 						return;
 					}
 
 					// compare documents to avoid minor HTML differences
 					// to cause unwanted updates
-					const jsonNew = JSON.stringify(this.editor.getJSON());
```
afbora: Just not sure about these two lines. I don't know if there is a difference in the data that `this` and `payload` contain.

afbora: So the purpose of this change has something to do with code splitting and are we sure it doesn't break anything?

distantnative: In theory we could keep it out of this PR. These are leftovers from my attempt to get Writer split into its own bundle. And there the issue is that relying on `this.editor` being present isn't good, but instead better to work with wha tis passed to the method (as there we can be sure it exists). But maybe indeed easier to just remove the Editor/Writer changes from this PR.

afbora: Ohh. I can confirm that this change reoccurs the https://github.com/getkirby/kirby/issues/3798 issue. So we should revert this two lines.

distantnative: Let's remove all writer changes from this PR

afbora: ✅

afbora (approved):

afbora:
Btw are we sure that want to move this changes to 3.7.3? I would favor of 3.7.4 since we already publish releases in short cycles.

PR #4509, opened 2022-07-17
Fix block `->prev()`, `->next()` and `->siblings()`

### LINKED ISSUE #4480: `$block->prev()` & `$block->next()` return hidden blocks
## Description
I use `$block->prev()` & `$block->next()` to change a block's contents based on the previous or next block's type. When testing, I noticed that blocks that are marked as hidden in the CMS won't be echo'ed, but returned by prev() or next() as they're seemingly still part of the Blocks collection. 

**Expected behavior**  
The next visible block will be returned, whilst hidden blocks will be ignored. As an alternative, `nextVisible()` or `prevVisible()` methods should be provided.

## To reproduce
Run `$block->prev()` or `$block->next()` on any block while the next or previous block is hidden. It will still return the hidden block and not the next shown block.

## Your setup
**Kirby Version**  3.7.0.1

afbora:
## This PR …

Now the previous and next item of the block returns according to own status (visible/hidden). If the block is hidden, the prev() and next() methods return the hidden block. If the block is visible, the prev() and next() methods return the visible blocks.

### Fixes

- `$block->prev()` & `$block->next()` no longer return hidden blocks from a visible block #4480

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Block.php
+++ b/src/Cms/Block.php
@@ -206,6 +206,16 @@ class Block extends Item
 	}
 
 	/**
+	 * Returns the sibling collection that filtered by block status
+	 *
+	 * @return \Kirby\Cms\Collection
+	 */
+	protected function siblingsCollection()
+	{
+		return $this->siblings->filter('isHidden', $this->isHidden());
+	}
+
+	/**
 	 * Returns the block type
 	 *
 	 * @return string
--- a/tests/Cms/Blocks/BlockTest.php
+++ b/tests/Cms/Blocks/BlockTest.php
@@ -260,4 +260,94 @@ class BlockTest extends TestCase
 		$this->assertSame('Lorem ipsum dolor …', $block->excerpt(20));
 		$this->assertSame($expected, (string)$block);
 	}
+
+	public function samplePrevNextBlocks()
+	{
+		return Blocks::factory([
+			[
+				'type' => 'code',
+				'isHidden' => true
+			],
+			[
+				'type' => 'gallery',
+				'isHidden' => true
+			],
+			[
+				'type' => 'heading',
+				'isHidden' => false
+			],
+			[
+				'type' => 'image',
+				'isHidden' => false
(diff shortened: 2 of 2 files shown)


lukasbestle (commented):
I like the solution with the `siblingsCollection`. It's consistent with the `Page` class.

But I wonder if we need `nextHidden` and `prevHidden`. We would also need `nextVisible` and `prevVisible`. And to find siblings no matter if hidden or not: `nextAny` and `prevAny` (in search of better names :D). Maybe we really need all of those, but I'm not sure about the use case yet.

afbora:
1. I didn't add `prevVisible` and `nextVisible` methods. Because there are already visible blocks in the frontend. I thought it might need to access hidden blocks. I can't think of any use for it, but I still think it will be useful 👍 
2. To be honest, I thought of an all-encompassing method but gave up because I couldn't find the right name 😅 The names `prevAny` and `nextAny` are perfect in my opinion 💯 

So if team agreed about Lukas new methods idea, I'll add them: `prevVisible`, `nextVisible`, `prevAny`, `nextAny`

lukasbestle:
I don't have a good overview on the use cases with blocks regarding what's practical and useful. Let's wait for the others' opinions.

distantnative:
I mean, I don't really object the new methods. But also don't really see the use cases. Maybe it's fine not to add even `nextHidden ` etc. but just the modified siblings collection?

afbora:
Actually just new `siblingsCollection()` method fixes the issue. Others optional..

distantnative:
That would be my preference (keeping method count low) - but if there are actual use cases for the new methods, I won't block them.

lukasbestle:
I agree. I think we should only add the siblingsCollection fix for now.

afbora:
Ready to review 🚀

distantnative (approved):

PR #4511, opened 2022-07-17
Fix double slashes in asset URLs with base URL `/`

### LINKED ISSUE #4508: Double-slashes when using assets helper
## Description
After running with panel issues as mentioned in issue #4443 we placed `'url' => '/'` in the config.php to get the panel working again. But that somehow breaks the assets helper, it now puts two slashes before the url:

``

With this in the snippet:

`<?= asset('assets/images/logo.png')->url() ?>`

**Expected behavior** 

A correct asset link, in this example:
`"/assets/images/logo.png"`

## To reproduce

1. Place `'url' => '/'` in config.php
1. Place asset on snippet/template using asset helper
2. It is rendered with double slashes and becomes unresolvable

## Your setup
**Kirby Version**  
3.7.1

We updated to the latest version after learning it fixes panel issues regarding https, but this issue remains somehow.

lukasbestle:
### Fixes

- Asset URLs no longer start with two slashes if the site's base URL was configured as `/` #4508

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/Asset.php
+++ b/src/Filesystem/Asset.php
@@ -38,7 +38,7 @@ class Asset
 		$this->setProperties([
 			'path' => dirname($path),
 			'root' => $this->kirby()->root('index') . '/' . $path,
-			'url'  => $this->kirby()->url('index') . '/' . $path
+			'url'  => $this->kirby()->url('base') . '/' . $path
 		]);
 	}
 
--- a/src/Panel/Document.php
+++ b/src/Panel/Document.php
@@ -276,7 +276,7 @@ class Document
 		try {
 			if (static::link() === true) {
 				usleep(1);
-				Response::go($kirby->url('index') . '/' . $kirby->path());
+				Response::go($kirby->url('base') . '/' . $kirby->path());
 			}
 		} catch (Throwable $e) {
 			die('The Panel assets cannot be installed properly. ' . $e->getMessage());
--- a/tests/Filesystem/AssetTest.php
+++ b/tests/Filesystem/AssetTest.php
@@ -44,6 +44,19 @@ class AssetTest extends TestCase
 		$this->assertSame('/dev/null/' . $file, $asset->id());
 		$this->assertSame('https://getkirby.com/' . $file, $asset->url());
 		$this->assertSame('images', $asset->path());
+
+		$this->app->clone([
+			'urls' => [
+				'index' => '/'
+			]
+		]);
+
+		$asset = $this->_asset($file = 'images/logo.svg');
+
+		$this->assertSame('/dev/null/' . $file, $asset->root());
+		$this->assertSame('/dev/null/' . $file, $asset->id());
+		$this->assertSame('/' . $file, $asset->url());
+		$this->assertSame('images', $asset->path());
 	}
 
 	/**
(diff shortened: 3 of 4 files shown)


src/Filesystem/Asset.php
```diff
@@ -38,7 +38,7 @@ public function __construct(string $path)
 		$this->setProperties([
 			'path' => dirname($path),
 			'root' => $this->kirby()->root('index') . '/' . $path,
-			'url'  => $this->kirby()->url('index') . '/' . $path
+			'url'  => rtrim($this->kirby()->url('index'), '/') . '/' . $path
```
afbora: This solution seems correct. But I think seems still confusing. `$this->kirby()->url('index')` should returns stable/consist value that with or without slash. So we shouldn't need either the `rtrim` method or the extra slash `/` 🤔

lukasbestle: `$this->kirby()->url('index')` by default always returns a value without trailing slash, unless it's `/`. This is handled by `Core::urls()` and `Environment::baseUrl()`.

`rtrim()` was the simplest solution I came up with (in the amount of code) and also prevents regressions in case the `index` URL for some reason ever *does* contain a trailing slash.

Open for suggestions though. :)

afbora: I agree with you and favor the simple solution if a more general but stable solution will cause regressions 👍

afbora: I'm good overview for this parts. But I just wonder that why didn't we do the same for `index` while the `base` url doesn't end with slash. I guess @bastianallgeier have an idea about that.
https://github.com/getkirby/kirby/blob/cfac315bc685193d0fe1b10fe3bf45a96e48fd15/src/Cms/Core.php#L459-L475

lukasbestle: Thinking about it, maybe we should convert `/` to an empty string already in `baseUrl`. Then it would be really consistent. But not sure if this would break other parts of the system.

lukasbestle: Just tested it with an empty `baseUrl`. This breaks immediately because some parts of the core expect the index URL to be non-empty. E.g. `config/routes.php:17` tries to call `Str::after()`, which fails with an empty needle.

So I think for now we need to keep the `baseUrl` as `/` in this special case and handle the trailing slash down the line (like in `Asset` in this case).

afbora: Then it seems like the safest way to go with your simple solution as this PR, do I understand right? 

But I think we need one more change: 
https://github.com/getkirby/kirby/blob/fix/4508-asset-url-double-slash/src/Panel/Document.php#L279

lukasbestle: Hm, good point. I'm a bit hesitant to sprinkle the `rtrim()` fix throughout the codebase through. If more parts are affected, this solution feels like a hack.

I moved this issue to 3.7.3. Let's wait for @bastianallgeier's opinion. Hopefully we can find a cleaner solution that fixes the issue in general.

bastianallgeier: To be honest, the cleanest and technically correct solution would be if all URLs always had a trailing slash. This would be a serious change though that would have serious implications for pretty much any part of the system. I agree that this seems to be the most stable version to keep everything as it is.

lukasbestle: You are right, a trailing slash really makes the most sense.
Do you think it's just risky in general as the whole system needs to be adapted or would it be a breaking change that users and plugin devs would also need to adapt to?

In any case, the switch to a trailing slash is only something we can do in a major release. If you agree, I can verify all uses of the index URL to make sure that they support the special `/` value. Then we can already fix the bug in 3.7.4.

lukasbestle:
Just found out we already have `$kirby->url('base')` that handles the `rtrim()`.

distantnative (approved):

PR #4512, opened 2022-07-17
Fix `query` to `searchterm` for sortable

afbora:
## This PR …

I've fixed the old `query` name to `searchterm` about sortable of sections. 

~But there is still a bug that `$this->searchterm` returns `null` from `sort` mixin. When I try to `App::instance()->request()->get('searchterm')`, value return but `$this->searchterm` not.  The mixins seem to have a problem accessing each other's props/computed values. @distantnative need your help.~

### Fixes
- Fixed sortable issue when search active for sections.

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -55,8 +55,8 @@ return [
 			$files = $files->filter('isReadable', true);
 
 			// search
-			if ($this->search === true && empty($this->searchterm) === false) {
-				$files = $files->search($this->searchterm);
+			if ($this->search === true && empty($this->searchterm()) === false) {
+				$files = $files->search($this->searchterm());
 			}
 
 			// sort
--- a/config/sections/mixins/search.php
+++ b/config/sections/mixins/search.php
@@ -11,7 +11,7 @@ return [
 			return $search;
 		}
 	],
-	'computed' => [
+	'methods' => [
 		'searchterm' => function (): ?string {
 			return App::instance()->request()->get('searchterm');
 		}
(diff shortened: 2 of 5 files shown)


afbora:
I'm also having this issue with the column sort feature in tables (on my local branch). So that feature is waiting for this issue to be resolved.

distantnative:
Maybe deleting the sort mixing and just integrating all these props into the section itself again is the best way forward. Fields/sections are currently a bit of a mess with their array syntax to really rely on when what can be called unfortunately.

afbora:
Sorry to hear that. I'll try to dig a little deeper into the mixins. I hope I find a solution 🤞

distantnative:
@afbora Don't waste too much time. We'll change how fields work soon anyways. I think is fine to just copy the sort mixin things back into the sections themselves. Even if it's a little less DRY

afbora:
Hmm, I've realized that works when convert `searchterm` prop `computed` to `method`, like `->searchterm()`. What do you think? Is it okay to use props like this as methods instead of computed?

distantnative:
Could work. It's really unpredictable how these sections work at the moment, so it's hard to tell if this will cause problems somewhere else.

afbora:
Then I would go with method `searchterm` instead computed until refactoring this part.

afbora:
@distantnative Ready for review.

By the way.  You know, `searchterm` is dynamic data. I don't know the exact property of a `computed` data, but I guess it is not set again after caching? Could it be that somehow the sections were loaded/initializated without the GET/REQUEST data? 🤔

distantnative (approved):

PR #4513, opened 2022-07-18
Fix 3.7 regression : Add missing onInput listener

seb-celinedesign:
## This PR …
Fix a 3.7 regression in StructureField.vue component
The on input listener has been removed when new k-table component was introduced

### Fixes
Just adding the onInput handler back again


afbora (commented):
The PR shouldn't include changes to dist/build files.

seb-celinedesign:
Thx for ur quick response.
I've change the PR. Now it's only containing the StructureField.vue file change.

afbora (approved):

distantnative (approved):

afbora:
Hey @seb-celinedesign This change unfortunately caused a serious regression to 3.7.2: https://github.com/getkirby/kirby/issues/4537. Please can you share why you need `onInput` listener in structure field form? Could you elaborate a little on your use case? What was not working prior to this PR?

seb-celinedesign:
Ok, i understand.
My use case is a Kirby plugin with a template Vue file using the `<k-structure-field>`.
I need to listen to user input on this structure to handle data.
In 3.6, as usual with Vue components, i was happily using the structure onInput event.

Without an onInput event, how do you listen to user structure inputs ? How Kirby is handling that in the panel ?
Thx

distantnative:
@seb-celinedesign And you want to receive the input already immediately when a user is editing one entry in the structure field, not when user is confirming the values for this one entry?

seb-celinedesign:
> @seb-celinedesign And you want to receive the input already immediately when a user is editing one entry in the structure field, not when user is confirming the values for this one entry?

When the user is confirming only (the "Ok" button).

distantnative:
Then listening to `@input` should still work with 3.7.2.1. when "ok" is clicked, we call onFormSubmit, which calls the save method in turn which calls itself onInout.

seb-celinedesign:
My bad, after checking my code, i need to receive the input immediately when a user is editing one field.
In my structure I've a `pages field`, and after a page selection, I need to copy the selected page data into the structure fields.

seb-celinedesign:
Maybe we could create a specific event for structure form inputs ? A event named `onFormInput`, which do not conflict with the standard `onInput` event.
`<k-structure-form @input="onFormInput">`
...
`onFormInput(e) { this.$emit("onFormInput", e); }`
What do you think ?

seb-celinedesign:
I can make a PR for that, if you need

PR #4515, opened 2022-07-18
More deprecated codes

afbora:
### Deprecations

- Removed `dump` component
- Removed locale magic for `Yaml::encode()` method
- Deprecated sections `headline` prop
- Deprecated `$kirby->server()`

## Ready?

### For review team

### DIFF
--- a/src/Data/Yaml.php
+++ b/src/Data/Yaml.php
@@ -24,22 +24,8 @@ class Yaml extends Handler
 	 */
 	public static function encode($data): string
 	{
-		// TODO: The locale magic should no longer be
-		//       necessary when support for PHP 7.x is dropped
-
-		// fetch the current locale setting for numbers
-		$locale = setlocale(LC_NUMERIC, 0);
-
-		// change to english numerics to avoid issues with floats
-		setlocale(LC_NUMERIC, 'C');
-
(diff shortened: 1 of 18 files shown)


afbora:
I've no idea about failing E2E tests 🤷‍♂️I've tested and headlines works great with `label` prop.

afbora:
I've found why E2E test failing. Because cypress using starterkit and it has still use headline prop.

distantnative:
@afbora would you be able to work on a PR for starter kit and demerit to change headline to label?

afbora:
@distantnative E2E tests passed after Bastian updated the sandbox kits.

src/Data/Yaml.php
```diff
@@ -24,22 +24,8 @@ class Yaml extends Handler
 	 */
 	public static function encode($data): string
 	{
-		// TODO: The locale magic should no longer be
```
distantnative, line 27: @lukasbestle i think you added this code originally? If so, could you double check that it's ok that we remove it now with PHP 8?

lukasbestle: We have a unit test for this that still passes. 👍

lukasbestle (approved):

PR #4517, opened 2022-07-18
Fix back button in the login challenge view

### LINKED ISSUE #4397: [panel] Password reset by code: Back button is broken
## Description
On the panel login, after the clicking "forgot your password?" the view is stuck on  the code challenge input.
Clicking 'back' does not switch back to password input.

The route `/logout` gets fetched, responds with a 302 to login and does not have an effect.

**Expected behavior**  
Clicking 'back' should return to the default email & password form.
This should happen even while the code challenge is still active.

## To reproduce

1. Go to /panel
2. Click on 'forgot your password?'
3. Click 'back'
4. Nothing changes, even after the challenge has timed out. The only workaround is to delete the `kirby_session` cookie

## My setup
**Kirby Version**  
3.6.6

**System**
 - OS: macOS, iOS 15.1 
 - Browser: Chrome 102, iOS Safari 15.4 

**Config**  
```php
return [
  'auth' => [
    'methods' => ['password', 'password-reset'],
    'challenge' => [
      'timeout' => 10 * 60, // 10 minutes
      'email' => [
        'from'     => 'me@example.com',
        'fromName' => 'Example',
        'subject'  => 'Login code'
      ]
    ]
]
```

lukasbestle:
### Enhancements

- During logout, the session and Kirby's internal state is now more thoroughly cleaned

### Fixes

- The "Back" button during a password reset or two-step login works again #4397

### Breaking changes

- The Panel view `account.logout` of the `account` area was replaced by the `logout` view of the new `logout` area

## Ready?

### For review team

### DIFF
--- a/config/areas/account/views.php
+++ b/config/areas/account/views.php
@@ -2,7 +2,6 @@
 
 use Kirby\Cms\App;
 use Kirby\Cms\Find;
-use Kirby\Panel\Panel;
 
 return [
 	'account' => [
@@ -17,16 +16,6 @@ return [
 		'action'  => function (string $filename) {
 			return Find::file('account', $filename)->panel()->view();
 		}
-	],
-	'account.logout' => [
-		'pattern' => 'logout',
-		'auth'    => false,
-		'action'  => function () {
-			if ($user = App::instance()->user()) {
-				$user->logout();
-			}
-			Panel::go('login');
-		},
 	],
 	'account.password' => [
 		'pattern' => 'reset-password',
--- a/config/areas/logout.php
+++ b/config/areas/logout.php
@@ -1,0 +1,21 @@
+<?php
+
+use Kirby\Panel\Panel;
+use Kirby\Toolkit\I18n;
+
+return function ($kirby) {
+	return [
+		'icon'  => 'user',
+		'label' => I18n::translate('logout'),
+		'views' => [
+			'logout' => [
+				'pattern' => 'logout',
+				'auth'    => false,
+				'action'  => function () use ($kirby) {
+					$kirby->auth()->logout();
+					Panel::go('login');
+				},
+			]
+		]
+	];
+};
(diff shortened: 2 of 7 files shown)


afbora (dismissed):
LGTM and works great 👍

afbora:
By the way thinking out loud, not important

`account.logout` and `login.logout` have the same function. Would it be possible to combine the two in one place and load globally for whole panel? Do you think it makes sense? Is it easy to implement?

distantnative:
I agree with @afbora - would be neat to not repeat the code. However, here https://github.com/getkirby/kirby/blob/main/src/Panel/Panel.php#L59 we currently make sure that `login` area is only loaded when not logged in and `account` only when logged in. Maybe we need to put the code in its own `logout` area and make sure that one is always available?

lukasbestle:
> Maybe we need to put the code in its own `logout` area and make sure that one is always available?

Could be a way, yes. But it would be overall more code than the duplicated route is right now considering that it's only two lines of actual code. I'd be fine with both solutions. What do you prefer?

afbora:
I would prefer `panel` (`panel.logout`, `panel.*`) area instead `logout` area. So in the future we can add an area that concerns the entire panel here.

lukasbestle:
The Panel itself is already split into areas for site, users, system and plugin areas. We introduced the areas to keep everything modular.

lukasbestle:
Thinking about it again, having a logout route in a login area doesn't make sense semantically. I think we should go for Nico's suggestion. If you agree, I can change it when I come home.

afbora:
As you see fit 👍

lukasbestle:
Ready for review again :)

afbora (approved):
LGTM 👍 

Few notes:
1. Need to update documents (about adding to website docs release checklist) https://getkirby.com/docs/reference/plugins/extensions/panel-areas#core-views
2. I don't know if anyone is using `account.logout` area, but can we call the removal of this area a breaking change?

lukasbestle:
You are right, technically it's a breaking change. Don't think it was overridden by anyone though, so I think 3.7.2 is fine.

distantnative (approved):

PR #4519, opened 2022-07-21
Sorting table via columns

afbora:
## This PR …

Now table data is sortable via new column option that `sortable`.

````yaml
columns:
  width:
    value: '{{ file.width }}'
    sortable: true
  height:
    value: '{{ file.height }}'
    sortable: true
````

chrome-capture-2022-6-21

### Features

- Table layout: change sort via column headers: https://kirby.nolt.io/438

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/sections/mixins/layout.php
+++ b/config/sections/mixins/layout.php
@@ -1,5 +1,6 @@
 <?php
 
+use Kirby\Cms\App;
 use Kirby\Toolkit\I18n;
 use Kirby\Toolkit\Str;
 
@@ -45,16 +46,18 @@ return [
 
 			if ($this->text) {
 				$columns['title'] = [
-					'label'  => I18n::translate('title'),
-					'mobile' => true,
-					'type'   => 'url',
+					'label'    => I18n::translate('title'),
+					'mobile'   => true,
+					'sortable' => true,
+					'type'     => 'url',
 				];
 			}
 
 			if ($this->info) {
 				$columns['info'] = [
-					'label' => I18n::translate('info'),
-					'type'  => 'text',
+					'label'    => I18n::translate('info'),
+					'type'     => 'text',
+					'sortable' => true,
 				];
 			}
 
@@ -91,7 +94,7 @@ return [
 			}
 
 			return $columns;
-		},
+		}
 	],
 	'methods' => [
 		'columnsValues' => function (array $item, $model) {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: config/sections/mixins/layout.php [hunk 1]
```diff
@@ -120,6 +123,39 @@
...
+						return $a[$sortColumn] <=> $b[$sortColumn];
+					}
+				} else {
+					if (is_array($b[$sortColumn])) {
+						return $b[$sortColumn]['text'] <=> $a[$sortColumn]['text'];
+					} else {
+						return $b[$sortColumn] <=> $a[$sortColumn];
+					}
+				}
+			});
```
afbora: I really need your help here. I have to use `usort` function because of data is array (multidimensional). I couldn't solve with Kirby way (especially via collections). What do you have better and stable solutions?

lukasbestle: We could use `array_multisort()` for this, which should be a bit faster than `usort()` (a real world test with a huge list would be good to verify).

Untested example:

```suggestion
			// extract the sort column value from each row
			$columnData = array_map(function ($row) use ($sortColumn) {
				$value = $row[$sortColumn];
				return is_array($value) ? $value['text'] : $value;
			}, $data);

			// sort $data by the sort column values
			array_multisort(
				$columnData,
				$sortDirection === 'desc' ? SORT_DESC : SORT_ASC,
				$data
			);
```

afbora:
@bastianallgeier you said let's postpone this feature because you refactor blueprints. What is the latest status? Shall we close the PR for now?

Btw what I'm really wondering is, what exactly is the connection between the blueprint part and this PR?

config/sections/mixins/layout.php
bastianallgeier: Did you check if A::sort could work here?

bastianallgeier:
@afbora my comment was based on the idea that the blueprint refactoring would already be ready for 3.8. Then the entire solution could be implemented right away without retro-fitting it into the old backend code. But I think it's clear now that it makes sense to implement it already without waiting for the new blueprints.

config/sections/mixins/layout.php
afbora: @bastianallgeier No, `A::sort` doesn't work. Because column value can be array. We need to extract the value like that: `return is_array($value) ? $value['text'] : $value;`

For using `A::sort`, the values in the table layout should be refactored:
https://github.com/getkirby/kirby/blob/develop/config/sections/mixins/layout.php#L98-L103

Current data:
````
Array
(
    [0] => Array
        (
            [title] => Array
                (
                    [text] => Title A
                    [url] => /panel/pages/a
                )
            [info] => Info 1
            [date] => 2020
        )
    [1] => Array
        (
            [title] => Array
                (
                    [text] => Title B
                    [url] => /panel/pages/b
                )
            [info] => Info 2
            [date] => 2021
        )
    [2] => Array
        (
            [title] => Array
                (
                    [text] => Title C
                    [url] => /panel/pages/c
                )
            [info] => Info 3
            [date] => 2022
        )

)
````

We need to normalize values:
````
Array
(
    [0] => Array
        (
            [title] => Title A
            [info] => Info 1
            [date] => 2020
        )
    [1] => Array
        (
            [title] => Title B
            [info] => Info 2
            [date] => 2021
        )
    [2] => Array
        (
            [title] => Title C
            [info] => Info 3
            [date] => 2022
        )

)
````

afbora:
I think we have a problem. If we use `limit` in section, we get paginated data in table layout (Because we are currently sorting at the blueprint level). So it's in the wrong order. It needs to be sorted on fly first and then paginated. 

Do you have an idea about the solution?

afbora:
Normally, the sorting is done through the collection at the blueprint level, while the table layout sorting needs to be done through the array after processing the custom values. This creates a big problem in pagination and requires a more extensive refactoring. Therefore, I'm closing the PR for future reference.

Here's the detailed conversation within the team:
https://discord.com/channels/525634039965679616/1020247250250776636

PR #4520, opened 2022-07-21
Panel redirects from hooks

afbora:
## This PR …

This is WIP and POC pull request. Just implemented for `page.create:after` hook for test. Need your reviews..
Not sure, may be we can catch the redirect exception for all hooks from single/common/global/final point/place.

### Sample codes and use case on implementing custom slug
````php
<?php

use Kirby\Cms\Page;
use Kirby\Panel\Panel;
use Kirby\Toolkit\Str;

return [
    'debug' => true,
     'hooks' => [
        'page.create:after' => function (Page $page) {
            $page = $page->changeSlug(Str::uuid());

            Panel::go($page->panel()->url(true));
        }
    ]
];
````

### Features
- Enable Panel redirects from backend hooks: https://kirby.nolt.io/298

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/areas/site/dialogs.php
+++ b/config/areas/site/dialogs.php
@@ -7,6 +7,7 @@ use Kirby\Exception\InvalidArgumentException;
 use Kirby\Exception\PermissionException;
 use Kirby\Panel\Field;
 use Kirby\Panel\Panel;
+use Kirby\Panel\Redirect;
 use Kirby\Toolkit\I18n;
 use Kirby\Toolkit\Str;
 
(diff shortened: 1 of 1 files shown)


afbora:
Interestingly, this feature is already working 😅

distantnative:
@afbora we should probably document that!

afbora:
You're right. Anyway I just wonder what @bastianallgeier thinks of both documenting and feedback item (Is it same feature with https://kirby.nolt.io/298?)

lukasbestle:
What happens if the hook is called from outside the Panel (e.g. if `$page->update()` is called inside a controller)? Probably the hooks should detect that and only call `Panel::go()` for Panel routes. Maybe we could automate that though and ignore the redirect in this case?

PR #4521, opened 2022-07-21
PHP 8: clean up `Panel` package

distantnative:


src/Panel/Panel.php
```diff
@@ -289,11 +262,8 @@ public static function response($result, array $options = [])
 
 	/**
 	 * Router for the Panel views
-	 *
-	 * @param string $path
-	 * @return \Kirby\Http\Response|false
 	 */
-	public static function router(string $path = null)
+	public static function router(string|null $path = null): Response|null
```
afbora, line 266: Since `Router::execute` returns `mixed`, shouldn't it return `mixed`?

distantnative: Doesn't `Router::execute` return what the callback closure returns? Which would be here `static::response()`.

afbora: Of course, you're right 👍  I missed the closure.

afbora (approved):
LGTM 👍

PR #4522, opened 2022-07-21
PHP 8: clean up `Image` package

distantnative:


src/Image/Darkroom.php
```diff
@@ -40,12 +35,9 @@ public function __construct(array $settings = [])
 	 * Creates a new Darkroom instance for the given
 	 * type/driver
 	 *
-	 * @param string $type
-	 * @param array $settings
-	 * @return mixed
 	 * @throws \Exception
 	 */
-	public static function factory(string $type, array $settings = [])
+	public static function factory(string $type, array $settings = []): static
```
lukasbestle: This will break with plugin classes that don't extend the `Darkroom` class.

src/Image/Exif.php
```diff
@@ -242,10 +198,8 @@ protected function parse()
 
 	/**
 	 * Return the timestamp when the picture has been taken
-	 *
-	 * @return string|int
 	 */
-	protected function parseTimestamp()
+	protected function parseTimestamp(): string|int
```
lukasbestle: The `$timestamp` prop supports `string|null`.

The type hints of this method and the prop taken together, it seems like the `$timestamp` can only ever be `string` in the end.

src/Image/Location.php
```diff
@@ -47,32 +36,24 @@ public function __construct(array $exif)
...
 
 	/**
 	 * Converts the gps coordinates
-	 *
-	 * @param string|array $coord
-	 * @param string $hemi
-	 * @return float
 	 */
-	protected function gps($coord, string $hemi): float
+	protected function gps(string|array $coord, string $hemi): float
```
lukasbestle: `count()` will throw a `TypeError` on string. Not sure how this is supposed to work with the current code.

lukasbestle (approved):

PR #4524, opened 2022-07-21
PHP 8: clean up `Data` package

distantnative:
## Breaking changes

`Kirby\Data\Data::write()` no longer accepts `null` for the `$file` argument

### DIFF
--- a/src/Data/Data.php
+++ b/src/Data/Data.php
@@ -26,10 +26,8 @@ class Data
 {
 	/**
 	 * Handler Type Aliases
-	 *
-	 * @var array
 	 */
-	public static $aliases = [
+	public static array $aliases = [
 		'md'    => 'txt',
 		'mdown' => 'txt',
 		'rss'   => 'xml',
@@ -38,10 +36,8 @@ class Data
 
 	/**
 	 * All registered handlers
-	 *
-	 * @var array
 	 */
-	public static $handlers = [
+	public static array $handlers = [
 		'json' => 'Kirby\Data\Json',
 		'php'  => 'Kirby\Data\PHP',
 		'txt'  => 'Kirby\Data\Txt',
@@ -51,11 +47,8 @@ class Data
 
 	/**
 	 * Handler getter
-	 *
-	 * @param string $type
-	 * @return \Kirby\Data\Handler
 	 */
-	public static function handler(string $type)
+	public static function handler(string $type): Handler
 	{
 		// normalize the type
 		$type = strtolower($type);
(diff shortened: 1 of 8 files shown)


src/Data/Data.php
```diff
@@ -100,10 +85,6 @@ public static function encode($data, string $type): string
 	 * Reads data from a file;
 	 * the data handler is automatically chosen by
 	 * the extension if not specified
-	 *
-	 * @param string $file
-	 * @param string $type
-	 * @return array
 	 */
 	public static function read(string $file, string $type = null): array
```
lukasbestle:
```suggestion
	public static function read(string $file, string|null $type = null): array
```

src/Data/Data.php
```diff
@@ -114,14 +95,12 @@ public static function read(string $file, string $type = null): array
...
-	 *
-	 * @param string $file
-	 * @param mixed $data
-	 * @param string $type
-	 * @return bool
 	 */
-	public static function write(string $file = null, $data = [], string $type = null): bool
-	{
+	public static function write(
+		string|null $file = null,
```
lukasbestle: I think this is worth a breaking change. `fileinfo()` (which is called by `F::extension()`) throws a deprecation warning if it gets passed `null`. Also I really cannot see the use case to write to a file `null`.

```suggestion
		string $file,
```

src/Data/Handler.php
```diff
@@ -54,12 +45,8 @@ public static function read(string $file): array
 
 	/**
 	 * Writes data to a file
-	 *
-	 * @param string $file
-	 * @param mixed $data
-	 * @return bool
 	 */
-	public static function write(string $file = null, $data = []): bool
+	public static function write(string|null $file = null, mixed $data = []): bool
```
lukasbestle: Here we can even change it without a breaking change: `F::write()` only accepts `string`.

```suggestion
	public static function write(string $file, mixed $data = []): bool
```

src/Data/PHP.php
```diff
@@ -19,12 +19,8 @@ class PHP extends Handler
 {
 	/**
 	 * Converts an array to PHP file content
-	 *
-	 * @param mixed $data
-	 * @param string $indent For internal use only
```
lukasbestle, line 24: I think we should keep this param description.

src/Data/PHP.php
```diff
@@ -74,12 +64,8 @@ public static function read(string $file): array
 
 	/**
 	 * Creates a PHP file with the given data
-	 *
-	 * @param string $file
-	 * @param mixed $data
-	 * @return bool
 	 */
-	public static function write(string $file = null, $data = []): bool
+	public static function write(string|null $file = null, mixed $data = []): bool
```
lukasbestle:
```suggestion
	public static function write(string $file, mixed $data = []): bool
```

lukasbestle (approved):

PR #4525, opened 2022-07-21
PHP 8: clean up `Email` package

distantnative:


src/Email/Email.php
```diff
@@ -123,8 +65,6 @@ public function __construct(array $props = [], bool $debug = false)
 
 	/**
 	 * Returns the email attachments
-	 *
-	 * @return array
 	 */
 	public function attachments(): array
```
lukasbestle, line 99: All of these getter methods except `isSent()` can currently also return `null` because the props are all nullable.

Maybe we clean the prop types a bit by updating the setters to replace `null` with an empty array in the `setAttachments()`, `setBcc()`, `setCc()`, `setTo()` and `setTransport()` methods. The getters of the remaining props need to be updated to return nullable values.

distantnative: Can they? `setAttachments` already ensure that an empty array is set if `null` is passed. Calling `resolveEmail()` in the setters also does that in the setters for bcc, cc, too. `::transport()` returns `::defaultTransport()` when it's null.

Type hinting at the top needs to include null as otherwise PHP throws an error, which I think is due to our `Properties` trait magic - not that the properties could actually be null once the constructor and magic has finished.

lukasbestle: You are right. Then I think we should actually add `@var` comments to the props again to mark the actual possible types after the constructor has run. Otherwise we will be confused down the line.

distantnative: Not sure if this would be less confusing (once you forget which one is the right one). I think the actual path forward is getting rid of the Properties trait.

distantnative: Reconsidered, `@var` comments might be good. Adding them.

src/Email/Body.php
```diff
@@ -19,20 +19,11 @@ class Body
...
-	protected $text;
+	protected string|null $html = null;
+	protected string|null $text = null;
 
 	/**
 	 * Email body constructor
-	 *
-	 * @param array $props
 	 */
 	public function __construct(array $props = [])
```
bastianallgeier, line 28: Wouldn't it be a good chance to refactor the constructor and get rid of the Properties trait, setText and setBody in one go? 

```php 
public function __construct(array $props = []) 
{
  $this->html = $props['html'] ?? null,
  $this->text = $props['text'] ?? null,
}
```

bastianallgeier (commented):
I think we could really get rid of all the protected set methods in Body and Email. I just don't know if this would lead to any potential breaking changes. Then it might not be worth it.

distantnative:
I've had a bunch of PRs trying to get rid of the Properties trait in some classes.

Two things are tricky:
- support for old props array as well as promotetes props in constructor (though manageable)
- the clone method

Especially the second one I always saw as essential and too much of a breaking change not to have. But it quickly gets very verbose when the clone method also needs to support props array as well as individual props as parameters.

bastianallgeier:
Very good point. Ok, then let's stick to the changes in this PR. Might also be the more productive way to move forward.

bastianallgeier (approved):

PR #4526, opened 2022-07-21
PHP 8: clean up `Filesystem` package

distantnative:


src/Cms/File.php
```diff
@@ -74,7 +74,7 @@ class File extends ModelWithContent
 	 *
 	 * @var string|null
```
afbora:
```suggestion
```

src/Cms/File.php
```diff
@@ -86,7 +86,7 @@ class File extends ModelWithContent
 	 *
 	 * @var string
```
afbora:
```suggestion
```

src/Cms/File.php
```diff
@@ -74,7 +74,7 @@ class File extends ModelWithContent
 	 *
 	 * @var string|null
 	 */
-	protected $root;
+	protected string|null $root = null;
 
 	/**
 	 * @var string
```
afbora, line 78: I'm asking because I don't know that why doesn't apply typed properties for some?  Like `$blueprint`, `$filename`, `$id` etc..

src/Cms/File.php
afbora:
```suggestion
```

src/Cms/File.php
distantnative: I only made this one change in `Cms\File` because PHP requires it when I make the changes in the `Filesystem` package. I haven't touched `Cms\File` at all otherwise as I want to limit this PR mainly to the `Filesystem` package to make it easier to overlook/review.

lukasbestle: I'm curious: What happens without the explicit `null` default? And why?

distantnative: PHP error that it cannot be accessed before initialized. I think in a lot of places we kinda rely on not set = null. Which worked before. But with type hints, that doesn't fly anymore.

afbora (approved):
LGTM 👍

PR #4527, opened 2022-07-21
PHP 8: clean up some of `Toolkit` package

distantnative:
**⚠️ Requires PHP 8, don't merge before Kirby 3.8**

### DIFF
--- a/src/Toolkit/Component.php
+++ b/src/Toolkit/Component.php
@@ -3,6 +3,7 @@
 namespace Kirby\Toolkit;
 
 use ArgumentCountError;
+use Closure;
 use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Filesystem\F;
@@ -148,7 +149,7 @@ class Component
 	protected function applyProps(array $props): void
 	{
 		foreach ($props as $propName => $propFunction) {
-			if (is_a($propFunction, 'Closure') === true) {
+			if (is_a($propFunction, Closure::class) === true) {
 				if (isset($this->attrs[$propName]) === true) {
 					try {
 						$this->$propName = $this->props[$propName] = $propFunction->call($this, $this->attrs[$propName]);
@@ -175,7 +176,7 @@ class Component
 	protected function applyComputed(array $computed): void
 	{
 		foreach ($computed as $computedName => $computedFunction) {
-			if (is_a($computedFunction, 'Closure') === true) {
+			if (is_a($computedFunction, Closure::class) === true) {
 				$this->$computedName = $this->computed[$computedName] = $computedFunction->call($this);
 			}
 		}
@@ -241,7 +242,7 @@ class Component
 	 */
 	public function toArray(): array
 	{
-		if (is_a($this->options['toArray'] ?? null, 'Closure') === true) {
+		if (is_a($this->options['toArray'] ?? null, Closure::class) === true) {
 			return $this->options['toArray']->call($this);
 		}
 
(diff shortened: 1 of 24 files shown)


src/Toolkit/Component.php
```diff
@@ -21,69 +21,49 @@ class Component
...
+	protected array $methods = [];
 
 	/**
 	 * An array of all component options
 	 * from the component definition
-	 *
-	 * @var array
 	 */
-	protected $options = [];
+	protected array|string $options = [];
```
afbora, line 52: I couldn't find any `string` usage?

src/Toolkit/Date.php
```diff
@@ -264,9 +248,8 @@ public function month(int|null $month = null): int
 	 * Returns the datetime which is nearest to the object's datetime
 	 *
 	 * @param string|int|\DateTimeInterface ...$datetime Datetime strings, UNIX timestamps or objects
-	 * @return string|int|\DateTimeInterface
 	 */
-	public function nearest(...$datetime)
+	public function nearest(...$datetime): string|int|DateTimeInterface
```
afbora: Is there any harm in using it like this?

```suggestion
	public function nearest(string|int|DateTimeInterface ...$datetime): string|int|DateTimeInterface
```

src/Toolkit/I18n.php
```diff
@@ -130,14 +110,12 @@ public static function locale(): string
...
-	 * @param string|null $locale
-	 * @return string|array|null
 	 */
-	public static function translate($key, $fallback = null, string $locale = null)
-	{
+	public static function translate(
+		string|array|null $key,
+		string|array|null $fallback = null,
+		string|null $locale = null
+	): mixed {
```
afbora, line 114: Is it possible to return a value other than `string|array|null` types?

src/Toolkit/Str.php
```diff
@@ -1132,9 +1056,6 @@ public static function startsWith(string $string = null, string $needle, bool $c
 	/**
 	 * Converts a string to studly caps case
 	 * @since 3.7.0
-	 *
-	 * @param string $value The string to convert
-	 * @return string
 	 */
 	public static function studly(string $value = null): string
```
afbora:
```suggestion
	public static function studly(string|null $value = null): string
```

src/Toolkit/I18n.php
distantnative: Yes, e.g. in `I18nTest.php:360` it would be a Closure.

src/Toolkit/Component.php
distantnative: Fields\TagsFieldTest::testOptionsQuery

afbora: Are you sure and could you mixing `Component::$options` with `\Kirby\Form\Options`? Or am I mixing it up. If you're sure, it's fine.

I've run `testOptionsQuery` unit test and `Component::$options` still array.

distantnative: You can see my commute. Had changed it but then our unit tests failed. Then reverted it.

afbora: 🙈 Sorry, missed that!

afbora (dismissed):
LGTM 👍

distantnative:
@afbora added a few more fixes https://github.com/getkirby/kirby/pull/4527/commits/85d61c73f05d91bab914c8551004e7ba7eae346c

afbora:
Looks good me but still it would be nice to have Lukas take a look at the psalm related changes.

distantnative:
Will reopen once merge conflicts etc are solved

PR #4528, opened 2022-07-22
PHP 8: clean up `Http` package

distantnative:


src/Http/Identifiers.php
```diff
@@ -0,0 +1,38 @@
...
+ * A wrapper for URL query and params
+ * @internal
+ *
+ * @package   Kirby Http
+ * @author    Nico Hoffmann <nico@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://opensource.org/licenses/MIT
+ */
+abstract class Identifiers extends Obj
```
distantnative: @bastianallgeier @lukasbestle if you have any idea for a better name, please let me know

lukasbestle: Maybe `KeyValueData`?

distantnative: Mmm that sounds super general. Maybe it's not even worth it to have a common parent class for keeping it DRY just a little bit more. Maybe you creating both classes separately with these few methods existing in both would be the cleaner way after all.

afbora: I am not sure. I'm just thinking out loud. Don't you think it is more appropriate to use `Trait` instead `abstract class`  in this context? 🤔

distantnative: @afbora what would be the benefit?

afbora: Great question 😄 To be honest, I can't show any obvious benefit. It just made more sense to me to use trait in terms of its function.

Does having several methods be similar require inheritance? Even the constructor of `Query` and `Params` is different anyway. `Identifiers::__construct` didn't seem very functional. As a result `Identifiers` class seemed like an unnecessary layer of inheritance to me. Not important. It's just that this extra layer was a bit more tiring when reviewing the code.

distantnative: Yea but that's be the same as a trait. Really leaning much more towards putting the same code just in the two classes. Tiny bit less DRY but so much better to read I think.

lukasbestle: I agree.

distantnative: ✅

src/Http/Cookie.php
```diff
@@ -122,7 +120,7 @@ public static function forever(string $key, string $value, array $options = []):
 	 *                             if the cookie has not been found
 	 * @return mixed The found value
```
afbora:
```suggestion
	 * @return string|array|null The found value
```

src/Http/Cookie.php
```diff
@@ -122,7 +120,7 @@ public static function forever(string $key, string $value, array $options = []):
...
 	 */
-	public static function get(string $key = null, string $default = null)
+	public static function get(string|null $key = null, string|null $default = null): mixed
```
afbora:
```suggestion
	public static function get(string|null $key = null, string|null $default = null): string|array|null
```

src/Http/Url.php
```diff
@@ -219,45 +172,32 @@ public static function short($url = null, int $length = 0, bool $base = false, s
...
 
 	/**
 	 * Smart resolver for internal and external urls
-	 *
-	 * @param string $path
-	 * @param mixed $options
-	 * @return string
 	 */
-	public static function to(string $path = null, $options = null): string
+	public static function to(string|null $path = null, array|null $options = null): string
```
afbora: `Uri::__construct` second parameter (for `$options` argument) not nullable.

src/Http/Visitor.php
```diff
@@ -65,11 +44,8 @@ public function __construct(array $arguments = [])
 	 * Sets the accepted language if
 	 * provided or returns the user's
 	 * accepted language otherwise
-	 *
-	 * @param string|null $acceptedLanguage
-	 * @return \Kirby\Toolkit\Obj|\Kirby\Http\Visitor|null
 	 */
-	public function acceptedLanguage(string $acceptedLanguage = null)
+	public function acceptedLanguage(string|null $acceptedLanguage = null): Obj|Visitor|null
```
afbora:
```suggestion
	public function acceptedLanguage(string|null $acceptedLanguage = null): static|Obj|null
```

afbora (approved):

src/Http/Environment.php
```diff
@@ -233,11 +194,8 @@ public function detect(array $options = null, array $info = null): array
 	/**
 	 * Sets the host name, port, path and protocol from the
 	 * fixed list of allowed URLs
-	 *
-	 * @param array|string $allowed
-	 * @return void
 	 */
-	protected function detectAllowed($allowed): void
+	protected function detectAllowed(array|string|object $allowed): void
```
lukasbestle, line 198: Why `object`? Won't this be wrapped into `[object]` and then tried to be passed to `new Uri()` (which only accepts `array|string`)?

src/Http/Response.php
```diff
@@ -206,13 +175,9 @@ public static function file(string $file, array $props = [])
...
 	 * @since 3.7.0
 	 *
-	 * @param string $url
-	 * @param int $code
-	 * @return void
-	 *
 	 * @codeCoverageIgnore
 	 */
-	public static function go(string $url = '/', int $code = 302)
+	public static function go(string $url = '/', int $code = 302): void
```
lukasbestle, line 180: We could add a `TODO` comment to change the return type to `never` once we drop support for PHP 8.0.

src/Http/Router.php
```diff
@@ -166,13 +149,8 @@ public static function execute(string|null $path = null, string $method = 'GET',
...
 	 * find matches and return all the found
 	 * arguments in the path.
-	 *
-	 * @param string $path
-	 * @param string $method
-	 * @param array $ignore
-	 * @return \Kirby\Http\Route|null
 	 */
-	public function find(string $path, string $method, array $ignore = null)
+	public function find(string $path, string $method, array|null $ignore = null): Route|null
```
lukasbestle, line 153: How can the return type be `null`?

src/Http/Router.php
```diff
@@ -120,23 +109,23 @@ public function call(string $path = null, string $method = 'GET', Closure $callb
...
+			if (is_a($this->beforeEach, Closure::class) === true) {
 				($this->beforeEach)($route, $path, $method);
 			}
 
 			try {
 				if ($callback) {
 					$result = $callback($route);
 				} else {
-					$result = $route->action()->call($route, ...$route->arguments());
+					$result = $route?->action()->call($route, ...$route->arguments());
```
lukasbestle, line 120: Followup to the previous comment: I think it can never be `null`.

PR #4529, opened 2022-07-22
3.7.2-rc.1

bastianallgeier:


lukasbestle:
Looks good except the translations. In some of them, the "Log In" translation got reversed to English. Can we discard those changes in the PR until the translations are fixed on Transifex as well?

bastianallgeier:
@lukasbestle that's the default behaviour by Transifex. If we discard them manually. We have to redo this with every language update. That's going to be pretty shit and error-prone to be honest.

lukasbestle:
I think we can assume that the translations are fixed until the next release. It's just for today because not every translator might have had the time to look into it yet.

bastianallgeier:
Ok, I've reverted the changes on those that fell back to English.

lukasbestle (approved):

PR #4540, opened 2022-07-31
Keep `$page->childrenAndDrafts()` updated

### LINKED ISSUE #2767: Creating new pages doesn't update the $page->childrenAndDrafts() collection
**Describe the bug**  

With the following code:

```php
$collection = site()->find('a-page')->childrenAndDrafts();

$collection->parent()->createChild([
  'slug'     => 'a-new-article',
  'template' => 'article',
  'content' => [
    'title'  => 'A new article',
    'author' => 'Homer Simpson'
  ]
]);
```

the `$collection` object is not updated after the page has been created, which means that the newly created page is not appended to the `$collection`.

The same thing likely happens when a page is modified (`changeSlug`, `delete` etc.).

Originally reported by @andreasotto [in Slack](https://app.slack.com/client/T2URU1Z6X/D5CU7RZM5/thread/C2UTAQAN4-1596969667.198200).

**Expected behavior**  

The existing `$collection` should automatically contain the newly created page.

**Kirby Version**  

3.4.2

**Additional context**  

We [already update the `drafts()` or `children()` collection](https://github.com/getkirby/kirby/blob/3.4.2/src/Cms/PageActions.php#L453-L458), but not the `childrenAndDrafts()` collection. This one is merged on the fly, but merging the collections clones the data array, so because only the individual collections are updated, the combined collection that has already been merged won't be updated automatically.

The only way to fix this would be to cache the `childrenAndDrafts()` collection as a property in the `HasChildren` trait (just like `$children` and `$drafts`). Then we could also update this collection once a new page is created, modified or deleted. However this would increase memory usage.

lukasbestle:
### Fixes

- The `$page->childrenAndDrafts()` collection is now immediately updated when a child page or draft is modified #2767

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/HasChildren.php
+++ b/src/Cms/HasChildren.php
@@ -19,16 +19,24 @@ trait HasChildren
 	/**
 	 * The list of available published children
 	 *
-	 * @var \Kirby\Cms\Pages
+	 * @var \Kirby\Cms\Pages|null
 	 */
 	public $children;
 
 	/**
 	 * The list of available draft children
 	 *
-	 * @var \Kirby\Cms\Pages
+	 * @var \Kirby\Cms\Pages|null
 	 */
 	public $drafts;
+
+	/**
+	 * The combined list of available published
+	 * and draft children
+	 *
+	 * @var \Kirby\Cms\Pages|null
+	 */
+	public $childrenAndDrafts;
 
 	/**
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):
Really nice bit of refactoring!

PR #4541, opened 2022-08-01
Allow panel plugins to load index.dev.mjs instead of index.js (needed by kirbyup)

jonaskuske:
## This PR allows plugins to use a module-based `index.dev.mjs` as entry

> Example repo to try it out: https://github.com/jonaskuske/kirbyup-hmr-playground

I'm currently doing work so Kirby plugins developed with `kirbyup` can use Vite features like HMR during development, which much improves the DX of authoring panel plugins that include Vue components: https://github.com/johannschopplich/kirbyup/issues/17#issuecomment-1200412908

However there are some constraints: Kirby plugins need to load in order (have to be registered on `window.panel.plugins` *before* the panel initializes), and the Vite dev server works with native JS modules, which execute asynchronously – so when panel and plugin are loaded separately, in-order execution can't be ensured.

To work around that, Kirby would either need to support the async registration of plugins (which I expect to be quite the invasive and complex change), *or* include module-based plugins in the panel's module graph via an import to make sure they execute first, which is what this PR implements.

### Implementation details

Instead of loading the panel code directly via `<script type="module" src="path/to/panel.js">`, it is now loaded like this:

```js
<script type="module">
import('path/to/panel.js')
</script>
```

This allows us to inject imports for module-based plugins before the panel code is imported:

```js
<script type="module">
import 'plugin1/index.dev.mjs';
import 'plugin2/index.dev.mjs';

import('path/to/panel.js')
</script>
```

Note that the import of `panel.js` must be a dynamic `import()`: in-order-execution of static imports is only guaranteed for synchronous modules, so if the first import uses top level await, the second import can actually run before the first one.

Loading the panel code through a dynamic import as part of the module body makes sure the panel runs after all plugins are loaded, since a module body executes after all static imports are done.
(in browsers supporting `link[rel=modulepreload]` the panel code is still fetched and compiled ahead of time)

This leaves us with one last problem: a single failing plugin import can prevent the entire panel from loading as there's no way to `catch` errors from a static import. To solve this, plugin modules themselves are loaded through a dynamic import, which can be wrapped in `try/catch`:

```js
<script type="module">
try {
  await Promise.all(['plugin1/index.dev.mjs', 'plugin2/index.dev.mjs'].map(uri => import(uri)))
} catch {
  // some plugins errored but we'll still continue to load path/to/panel.js!
}

import('path/to/panel.js')
</script>
```

As a perf enhancement and to avoid having to copy `index.dev.mjs` to `/media` and then delete it again when the kirbyup development server is stopped and the `index.dev.js` is removed, the plugins are loaded inline, as data URIs:

```js
<script type="module">
try {
  await Promise.all(
    [
    'data:text/javascript;base64,aDfghddplugin1==',
    'data:text/javascript;base64,aDfghddplugin2=='
    ].map(uri => import(uri))
  )
} catch {
  // some plugins errored but we'll still continue to load path/to/panel.js!
}

import('path/to/panel.js')
</script>
```

If a plugin has an `index.dev.mjs` which is loaded through the technique above, its regular `index.js` will be excluded from the `plugins/index.js` bundle. Otherwise, its `index.js` is loaded as part of `plugins/index.js` as it used to.

### Breaking changes

No, the only change is that `panel.js` is now loaded via a dynamic `import()`, which is covered by Kirby's minimum browser requirements. All other changes only have an effect if a `kirbyup` development server is running and created an `index.dev.mjs`.

## Ready?

### For review team

### DIFF
--- a/src/Cache/FileCache.php
+++ b/src/Cache/FileCache.php
@@ -79,19 +79,19 @@ class FileCache extends Cache
 				case '/':
 					break;
 
-				// backslashes get their own marker in the path
-				// to differentiate the cache key from one with forward slashes
+					// backslashes get their own marker in the path
+					// to differentiate the cache key from one with forward slashes
 				case '\\':
 					$keyParts[] = '_backslash';
 					break;
 
-				// empty part means two slashes in a row;
-				// special marker like for backslashes
+					// empty part means two slashes in a row;
+					// special marker like for backslashes
 				case '':
 					$keyParts[] = '_empty';
 					break;
 
-				// an actual path segment
+					// an actual path segment
 				default:
 					// check if the segment only contains safe characters;
 					// underscores are *not* safe to guarantee uniqueness
--- a/src/Cache/FileCache.php
+++ b/src/Cache/FileCache.php
@@ -79,19 +79,19 @@ class FileCache extends Cache
 				case '/':
 					break;
 
-					// backslashes get their own marker in the path
-					// to differentiate the cache key from one with forward slashes
+				// backslashes get their own marker in the path
+				// to differentiate the cache key from one with forward slashes
 				case '\\':
 					$keyParts[] = '_backslash';
 					break;
 
-					// empty part means two slashes in a row;
-					// special marker like for backslashes
+				// empty part means two slashes in a row;
+				// special marker like for backslashes
 				case '':
 					$keyParts[] = '_empty';
 					break;
 
-					// an actual path segment
+				// an actual path segment
 				default:
 					// check if the segment only contains safe characters;
 					// underscores are *not* safe to guarantee uniqueness
--- a/src/Panel/Document.php
+++ b/src/Panel/Document.php
@@ -73,7 +73,7 @@ class Document
 			],
 			'icons' => static::favicon($url),
 			// loader for plugins' index.dev.mjs files – inlined, so we provide the code instead of the asset URL
-			'plugin_imports' => $plugins->read('mjs'),
+			'plugin-imports' => $plugins->read('mjs'),
 			'js' => [
 				'vendor'       => [
 					'nonce' => $nonce,
--- a/src/Panel/Document.php
+++ b/src/Panel/Document.php
@@ -93,17 +93,13 @@ class Document
 					'src'   => static::customAsset('panel.js'),
 					'type'  => 'module'
 				],
-				'plugins-mjs'      => [
-					'nonce' => $nonce,
-					'src'   => $plugins->url('mjs'),
-					'type'  => 'module'
-				],
 				'index'        => [
 					'nonce' => $nonce,
 					'src'   => $url . '/js/index.js',
 					'type'  => 'module'
 				],
-			]
+			],
+			'plugin-imports' => $plugins->read('mjs')
 		];
 
 		// during dev mode, add vite client and adapt
--- a/src/Panel/Document.php
+++ b/src/Panel/Document.php
@@ -72,6 +72,7 @@ class Document
 				'custom'  => static::customAsset('panel.css'),
 			],
 			'icons' => static::favicon($url),
+			'imports' => $plugins->read('mjs'),
 			'js' => [
 				'vendor'       => [
 					'nonce' => $nonce,
@@ -98,8 +99,7 @@ class Document
 					'src'   => $url . '/js/index.js',
 					'type'  => 'module'
 				],
-			],
-			'plugin-imports' => $plugins->read('mjs')
+			]
 		];
 
 		// during dev mode, add vite client and adapt
--- a/src/Panel/Document.php
+++ b/src/Panel/Document.php
@@ -72,7 +72,8 @@ class Document
 				'custom'  => static::customAsset('panel.css'),
 			],
 			'icons' => static::favicon($url),
-			'imports' => $plugins->read('mjs'),
+			// loader for plugins' index.dev.mjs files – inlined, so we provide the code instead of the asset URL
+			'plugin_imports' => $plugins->read('mjs'),
 			'js' => [
 				'vendor'       => [
 					'nonce' => $nonce,
--- a/src/Panel/Document.php
+++ b/src/Panel/Document.php
@@ -95,7 +95,7 @@ class Document
 				],
 				'index'        => [
 					'nonce' => $nonce,
-					'src'   => $url . '/js/index.js',
+					'src'   => $url . '/js/_index.js',
 					'type'  => 'module'
 				],
 			]
--- a/src/Panel/Document.php
+++ b/src/Panel/Document.php
@@ -93,9 +93,14 @@ class Document
 					'src'   => static::customAsset('panel.js'),
 					'type'  => 'module'
 				],
+				'plugins-mjs'      => [
+					'nonce' => $nonce,
+					'src'   => $plugins->url('mjs'),
+					'type'  => 'module'
+				],
 				'index'        => [
 					'nonce' => $nonce,
-					'src'   => $url . '/js/_index.js',
+					'src'   => $url . '/js/index.js',
 					'type'  => 'module'
 				],
 			]
--- a/src/Panel/Plugins.php
+++ b/src/Panel/Plugins.php
@@ -89,12 +89,7 @@ class Plugins
 		if ($type === 'js') {
 			$files = A::filter(
 				$files,
-				function ($file) {
-					$is_js = Str::endsWith($file, 'index.js');
-					$has_dev_mjs = F::exists(preg_replace('/\.js$/', '.dev.mjs', $file));
-
-					return ($is_js && $has_dev_mjs) === false;
-				}
+				fn ($file) => F::exists(preg_replace('/\.js$/', '.dev.mjs', $file))
 			);
 		}
 
--- a/src/Panel/Plugins.php
+++ b/src/Panel/Plugins.php
@@ -83,43 +83,48 @@ class Plugins
 	{
 		$dist = [];
 
-		$files = $this->files();
+		foreach ($this->files() as $file) {
+			// filter out files with a different type
+			if (F::extension($file) !== $type) {
+				continue;
+			}
 
-		// Filter out all index.js files that shouldn't be loaded because an index.dev.mjs exists
-		if ($type === 'js') {
-			$files = A::filter(
-				$files,
-				fn ($file) => F::exists(preg_replace('/\.js$/', '.dev.mjs', $file))
-			);
-		}
+			// filter out empty files and files that don't exist
+			$content = F::read($file);
+			if (!$content) {
+				continue;
+			}
 
-		foreach ($files as $file) {
-			if (F::extension($file) === $type) {
-				if ($content = F::read($file)) {
-					if ($type === 'mjs') {
-						// index.dev.mjs files are turned into data URIs so they can be imported
-						// without having to copy them to /media, then later remove them again
-						$content = F::uri($file);
-					}
+			if ($type === 'mjs') {
+				// index.dev.mjs files are turned into data URIs so they
+				// can be imported without having to copy them to /media
+				// (avoids having to clean the files from /media again)
+				$content = F::uri($file);
+			}
 
-					if ($type === 'js') {
-						$content = trim($content);
+			if ($type === 'js') {
+				// filter out all index.js files that shouldn't be loaded
+				// because an index.dev.mjs exists
+				if (F::exists(preg_replace('/\.js$/', '.dev.mjs', $file)) === true) {
+					continue;
+				}
 
-						// make sure that each plugin is ended correctly
-						if (Str::endsWith($content, ';') === false) {
-							$content .= ';';
-						}
-					}
+				$content = trim($content);
 
-					$dist[] = $content;
+				// make sure that each plugin is ended correctly
+				if (Str::endsWith($content, ';') === false) {
+					$content .= ';';
 				}
 			}
+
+			$dist[] = $content;
 		}
 
 		if ($type === 'mjs') {
-			// If no index.dev.mjs modules exist, we MUST return an empty string instead of loading an empty array.
-			// This is because the module loader code uses top level await, which is not compatible with Kirby's
-			// minimum browser version requirements and therefore mustn't appear in a default setup.
+			// if no index.dev.mjs modules exist, we MUST return an empty string instead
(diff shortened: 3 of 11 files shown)


lukasbestle:
Thank you very much for your contribution. I like the concept as it really feels like something that can just work™ without any need to think about the details as a user. Especially the idea to dynamically generate the `index.mjs` file in kirbyup and remove it again when kirbyup is stopped is really great.

A few comments:

- Currently you dynamically inject the import into the Panel JS in a route. This will break when a CDN is used to serve the `media` folder. Also it's not very efficient to spin up PHP for every Panel JS request. We need a way to inject the import at build time so that the media folder can be served statically.
- Requesting files directly from `site/plugins` will not work with the default and recommended configuration. The `site` directory must not be directly accessible to avoid security breaches.
- I think this should be a dev-only feature. In production it adds additional complexity and is probably not as reliable as a single bundle that can again be served via a CDN if needed. So I'd be a fine with a configuration option that can then be enabled in a domain-specific config file just for the dev domain.

jonaskuske:
> Also it's not very efficient to spin up PHP for every Panel JS request. We need a way to inject the import at build time so that the media folder can be served statically.

Yeah, agree that this isn't optimal! But since the `media` URL can vary, I didn't see how it can be injected at panel build time, which happens during release and before we know anything about a user's configuration.

Another way would be to load both the plugin and panel JS from an inline module:

```html
<script type="module">
import '/media/plugins/index.mjs'

import('/media/panel/assets/js/index.js')
</script>
```

Not sure if the browser's preload scanner detects these imports AOT, if yes it shouldn't make a difference perf-wise, otherwise we'd also need to add a `<link rel="modulepreload">` to the head. Which is a good idea anyway.

> I think this should be a dev-only feature.

Makes sense! I think `.mjs` support in shipped plugins would be cool for "forward-thinking, we support the latest and greatest" reasons, but you're right, not really for practical ones. Dev-only avoids a lot of complexity, and especially...

> Requesting files directly from `site/plugins` will not work with the default and recommended configuration.

...if we know the `index.mjs` is just for development and only ever includes an absolute import that goes to Vite, we can move it to the `media` folder without causing issues for relative imports. (because there are none)  
I think we could even skip loading the actual `.mjs` files all together and embed them inline as `data:` URLs – from my testing, base64-encoded inline modules that include relative (or just not-fully-qualified) URL imports fail silently, but full URL imports (and inline code) work without issues.

Gonna push some changes later today. :) 

Edit:
I also think we can make do without a config option, which would mean *it just works ™️* even better. If plugins have `.mjs` files, include them via top level await, otherwise don't. Maybe change the name to `_index.mjs` to make it clear this is an internal, dev-only feature. Then document the browser version requirements in the docs for `kirbyup serve`.

jonaskuske:
Updated the loading behavior as described in my comment above. Let me know what you think!  

(the `plugins.mjs` is always loaded so we don't have to read its contents before adding the script, but it's empty when there are no `_index.mjs` files – so there's no top level await in the source unless a plugin Vite server is running, the browser requirements stay the same)

src/Panel/Plugins.php
```diff
@@ -42,6 +44,8 @@ public function files(): array
 		foreach (App::instance()->plugins() as $plugin) {
 			$this->files[] = $plugin->root() . '/index.css';
 			$this->files[] = $plugin->root() . '/index.js';
+			// used by kirbyup during plugin development
+			$this->files[] = $plugin->root() . '/_index.mjs';
```
lukasbestle: I think I'd prefer it without the underscore. If we want to make it clear it is only for development, we should maybe use something like `index.dev.mjs`.

src/Panel/Document.php
```diff
@@ -93,6 +93,11 @@ public static function assets(): array
 					'src'   => static::customAsset('panel.js'),
 					'type'  => 'module'
 				],
+				'plugins-mjs'      => [
```
lukasbestle:
```suggestion
				'plugins-mjs'  => [
```

lukasbestle: Can we include this asset conditionally if the file is not going to be empty? This would avoid the extra request in production.

src/Panel/Plugins.php
```diff
@@ -92,6 +112,16 @@ public function read(string $type): string
 			}
 		}
 
+		if ($type === 'mjs') {
+			if (empty($dist)) {
+				return "";
```
lukasbestle:
```suggestion
				return '';
```

src/Panel/Document.php
jonaskuske: Yep, think so. We could probably just inline it, then it'd be one less request in the chain during development (no separate `/plugins.mjs`) and also no unnecessary code in production as we'd just "inline" an empty string there.

src/Panel/Plugins.php
jonaskuske: Done :)

src/Panel/Plugins.php
```diff
@@ -75,9 +79,25 @@ public function read(string $type): string
...
+
+		// filter out index.js files that have an index.dev.mjs counterpart (which takes precedence)
+		if ($type === 'js') {
+			$files = A::filter(
+				$files,
+				fn ($f) => !(
+					Str::endsWith($f, 'index.js') && F::exists(preg_replace('/\.js$/', '.dev.mjs', $f))
+				)
+			);
+		}
```
distantnative: I think this filtering might better done in `Plugins::files()`

views/panel.php
```diff
@@ -38,6 +38,12 @@
   <link nonce="<?= $nonce ?>" rel="<?= $rel ?>" href="<?= Url::to($icon['url']) ?>" type="<?= $icon['type'] ?>">
   <?php endforeach ?>
 
+  <?php foreach ($assets['js'] as $js): ?>
+    <?php if (($js['type'] ?? null) === 'module'): ?>
+    <link rel="modulepreload" href="<?= $js['src'] ?>">
+    <?php endif ?>
+  <?php endforeach ?>
```
distantnative, line 45: Would be good to keep the style consistent: no indent for loops

views/panel.php
```diff
@@ -59,8 +65,12 @@
     window.fiber = json;
   </script>
 
-  <?php foreach ($assets['js'] as $js): ?>
-  <?= Html::tag('script', '', $js) . PHP_EOL ?>
+  <?php foreach ($assets['js'] as $key => $js): ?>
+    <?php if ($key === 'index'): ?>
+    <?= Html::tag('script', [$assets['imports'], "import('{$js['src']}')"], A::without($js, 'src')) . PHP_EOL ?>
```
distantnative: This is super hard to read and understand. I think it would be better to not use `Html::tag` here but really write out the script tag.

distantnative:
I added a few comments. Would be great to have more inline comments as well to understand why we are doing these extra jumps etc. also 2,3 years down the road.

src/Panel/Plugins.php
jonaskuske: Wasn't sure here – the way it was implemented before, `::files()` just listed all *possible* plugin files and then `::read()` did the checks to see if they actually exist. That's why I kept `::files()` pure and added the additional filesystem access inside the `::read()` method instead of in `::files()`.

(but happy to change it if you want!)

lukasbestle: I tend to agree with Jonas. CSS and JS are not filtered in `Plugins::files()` either. On the other hand, the filtering logic would be much simpler in `files()` because we wouldn't need the regex.

src/Cache/FileCache.php
```diff
@@ -79,19 +79,19 @@ protected function file(string $key): string
 				case '/':
 					break;
 
-				// backslashes get their own marker in the path
-				// to differentiate the cache key from one with forward slashes
+					// backslashes get their own marker in the path
+					// to differentiate the cache key from one with forward slashes
```
lukasbestle: Could you please revert the unrelated coding style changes in `Cache\FileCache`, `Cms\Email`, `Database\Sql` and `Toolkit\Html`?

distantnative: @lukasbestle I get the same changes since a while locally from cs fixer. But somehow the CI workflow doesn't want them? Though agreed they shouldn't be in this PR

jonaskuske: Reverted! But yup, that behavior from `php-cs-fixer` is pretty annoying, have to manually run it once for all files I've touched 😅

jonaskuske:
@lukasbestle 
Regarding the failing unit test:
image

I noticed that behavior before: if a JS file contains less than 3 chars, the reported mime type isn't `text/javascript` but `application/octet-stream`. Would you rather have me:

**a)** fix the test (use `ccc` instead of `c` as module contents?)
**b)** fix the code, by adding another entry to the fixing map in `Mime.php`:

```php
'application/octet-stream' => [
  'mjs' => 'text/javascript'
]
```
(option **b)** is currently implemented)

src/Panel/Plugins.php
```diff
@@ -92,6 +112,16 @@ public function read(string $type): string
...
 		}
 
+		if ($type === 'mjs') {
+			if (empty($dist)) {
+				return '';
+			}
+
+			$modules = Json::encode($dist);
+			$modulePromise = "Promise.all($modules.map(url => import(url)))";
+			return "try { await $modulePromise } catch (e) { console.error(e) }" . PHP_EOL;
```
lukasbestle, line 122: Why EOL specifically? Given that we use `\n` throughout the code base, shouldn't we also use it here?

lukasbestle:
I think it's great.

Just three things still to do IMO:

- [x] Could you please update the first post of this PR with the current info?
- [x] Another unit test that covers [this case (no `index.dev.mjs`)](https://github.com/getkirby/kirby/blob/c2dd5e15d2c1d239a3b2856995efe35dfe6ab5fc/src/Panel/Plugins.php#L117) would be great
- [x] Open discussion about where to place the filter code; @distantnative, what do you think?

src/Panel/Plugins.php
distantnative: I see your arguments, I'd be fine to leave it where it is now. Can we just clean up the code Style a bit? We usually don't do this `!` to negate the result as it's harder to red and understand. Would be great to relive this

jonaskuske: Refactored the filter fn a little, what do you think?

jonaskuske:
@lukasbestle All done!


And another thing: just realized that this could actually be implemented without top level await, by using an async IIFE in the main `<script type=module>` and exporting promises from the `index.dev.mjs` files, which the main script can await before importing `panel.js`. This would lower the browser version support requirements, but with this being dev-only and all *current* browsers supporting top level await, I don't think it's worth the change. Opinions?

lukasbestle:
> This would lower the browser version support requirements, but with this being dev-only and all *current* browsers supporting top level await, I don't think it's worth the change.

Fully agree!

src/Panel/Document.php
```diff
@@ -72,6 +72,8 @@ public static function assets(): array
 				'custom'  => static::customAsset('panel.css'),
 			],
 			'icons' => static::favicon($url),
+			// loader for plugins' index.dev.mjs files – inlined, so we provide the code instead of the asset URL
+			'plugin_imports' => $plugins->read('mjs'),
```
lukasbestle:
```suggestion
			'plugin-imports' => $plugins->read('mjs'),
```

src/Panel/Plugins.php
```diff
@@ -75,9 +83,30 @@ public function read(string $type): string
...
+
+		// Filter out all index.js files that shouldn't be loaded because an index.dev.mjs exists
+		if ($type === 'js') {
+			$files = A::filter(
+				$files,
+				function ($file) {
+					$is_js = Str::endsWith($file, 'index.js');
+					$has_dev_mjs = F::exists(preg_replace('/\.js$/', '.dev.mjs', $file));
+
+					return ($is_js && $has_dev_mjs) === false;
```
lukasbestle:
```suggestion
					$isJs = Str::endsWith($file, 'index.js');
					$hasDevMjs = F::exists(preg_replace('/\.js$/', '.dev.mjs', $file));

					return $isJs === false || $hasDevMjs === false;
```

Do we even need the `jsJs` check here though? Down below in the loop we already check the extension. Unless I miss something here, we could replace this with a check just for `.dev.mjs`, which will allow a one-liner arrow function.

tests/Panel/PluginsTest.php
```diff
@@ -28,7 +33,7 @@ public function setUp(): void
 		]);
 	}
 
-	public function createPlugins()
+	public function createPlugins($add_dev_mjs = false)
```
lukasbestle:
```suggestion
	public function createPlugins(bool $addDevMjs = false)
```

tests/Panel/PluginsTest.php
```diff
@@ -38,13 +43,28 @@ public function createPlugins()
...
+
+		F::write($this->tmp . '/site/plugins/c/index.php', '<?php Kirby::plugin("test/c", []);');
+		touch($this->tmp . '/site/plugins/c/index.php', $time);
+		F::write($this->cssC = $this->tmp . '/site/plugins/c/index.css', 'c');
+		touch($this->cssC, $time);
+		F::write($this->jsC = $this->tmp . '/site/plugins/c/index.js', 'c');
+		touch($this->jsC, $time);
+		$this->mjsC = $this->tmp . '/site/plugins/c/index.dev.mjs';
+
+		if ($add_dev_mjs) {
```
lukasbestle:
```suggestion
		if ($addDevMjs === true) {
```

views/panel.php
```diff
@@ -59,8 +65,15 @@
     window.fiber = json;
   </script>
 
-  <?php foreach ($assets['js'] as $js): ?>
+  <?php foreach ($assets['js'] as $key => $js): ?>
+  <?php if ($key === 'index'): ?>
+  <script type="module" nonce="<?= $nonce ?>">
+    <?= $assets['plugin_imports'] ?>
```
lukasbestle:
```suggestion
    <?= $assets['plugin-imports'] ?>
```

lukasbestle (commented):
Mostly coding style stuff, the rest looks really great. Thanks a lot for the extra code comments. 👍

src/Panel/Plugins.php
jonaskuske: You're right, I think it should work. Only ~time~ tests will tell.

lukasbestle (approved):

distantnative (approved):

PR #4543, opened 2022-08-02
Fix double-encoding in table layout

### LINKED ISSUE #4442: Encoding-issue in the new table-layout
I just converted a blog to the new table-layout: https://getkirby.com/docs/reference/panel/sections/pages#layout__table-layout

Using "'" in titles breaks the table-layout encoding. Here is an article:

grafik

Here is the same article in the panel

grafik

In the regular view it works:

grafik

Some kind of encoding issue. The original file is encoded in UTF8 and can be found here:

https://gitlab.com/cniehaus/homepage/-/blob/master/content/1_blogs/20220624_rassismus-geht-uns-alle-an/blogartikel.txt

bastianallgeier:
## This PR …

The column values for the table layout were double encoded. For the Item.vue component, we send safe strings for the text and info. The component uses a span with v-html attribute. That's why the safe string is needed here as default. The field preview components don't do this though. They already rely on Vue's xss prevention – except the HtmlFieldPreview component. That's why all the documented column types actually need unescaped values to avoid double-encoding. 

### Fixes

- Values in the table layout are no longer escaped twice #4442

## Ready?

### For review team

### DIFF
--- a/config/sections/mixins/layout.php
+++ b/config/sections/mixins/layout.php
@@ -96,9 +96,13 @@ return [
 	'methods' => [
 		'columnsValues' => function (array $item, $model) {
 			$item['title'] = [
-				'text' => $item['text'],
+				'text' => $model->toString($this->text),
 				'href' => $model->panel()->url(true)
 			];
+
+			if ($this->info) {
+				$item['info'] = $model->toString($this->info);
+			}
 
 			foreach ($this->columns as $columnName => $column) {
 				// don't overwrite essential columns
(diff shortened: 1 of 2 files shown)


lukasbestle:
Have you found out why it works with explicit `type: text`? Shouldn't this also have called `toSafeString()`?

bastianallgeier:
I can only guess that the if clause didn't work correctly because of the ?? operator without additional parenthesis.

lukasbestle:
> I can only guess that the if clause didn't work correctly because of the ?? operator without additional parenthesis.

I have verified that this is the reason. Good catch! 👍

lukasbestle (approved):
I have not tracked down that all frontend code paths except `html` use Vue's escaping, but there I trust your better overview.

The backend changes look great and my manual tests all worked as well.

Issue #4400, opened 2022-06-20
Text jumps during drag/sorting


manuelmoreale:
## Description
The title of the item inside a pages section jumps left during drag.
Video should be self explanatory. It's probably just a simple CSS issue

**Screenshots**  

https://user-images.githubusercontent.com/1738077/174623027-14acd1a4-6855-4de8-9397-61034c607196.mp4

bastianallgeier:
Is that a new effect or does it really only appear since 3.7.0-rc.1?

afbora:
@bastianallgeier No, already exists in 3.6.6

manuelmoreale:
@bastianallgeier honestly only noticed while starting a new project on 3.7.0 that's why I assumed it was related to the RC. Thanks @afbora for checking

afbora:
### Summary 

We have `dir` attribute on `div` wrapper of panel. But cloned item outside of the wrapper when drag start. So following margin style doesn't effect to item title:

https://github.com/getkirby/kirby/blob/3.6.6/panel/src/components/Layout/Item.vue#L223

medienbaecker:
This is probably related: the `cards` layout items also have some weird style changes while dragging:

<img width="686" alt="CleanShot 2022-06-24 at 13 32 43@2x" src="https://user-images.githubusercontent.com/7975568/175526347-5964f9a7-bae9-4cf9-871a-5abf78a25851.png">

afbora:
✅

PR #4544, opened 2022-08-03
Add dir attribute to body element to fix layout issues in drag ghosts

bastianallgeier:
### Fixes

- Text no longer jumps during drag/sorting #4400 

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Layout/Panel.vue
+++ b/panel/src/components/Layout/Panel.vue
@@ -7,7 +7,7 @@
 		:data-role="role"
 		:data-translation="$translation.code"
 		:data-user="user"
-		:dir="$translation.direction"
+		:dir="dir"
 		class="k-panel"
 	>
 		<slot />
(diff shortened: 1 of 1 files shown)


panel/src/components/Layout/Panel.vue
```diff
@@ -7,7 +7,7 @@
 		:data-role="role"
 		:data-translation="$translation.code"
 		:data-user="user"
-		:dir="$translation.direction"
+		:dir="dir"
```
lukasbestle, line 10: Do we even need to set the attribute here if we already set it on `<body>`?

bastianallgeier: I didn't want to introduce a breaking change here. Those attributes are promoted as ways to handle custom styles in the panel.css. It's unlikely, but if someone does `.k-panel[dir=rtl] { }` this would break now.

lukasbestle (approved):

PR #4555, opened 2022-08-08
Extendable text block field input 

### LINKED ISSUE #3016: Extending blocks type "text" adds p tags to the input
**Description**  
I’m currently extending the text block type:
```
text:
  extends: blocks/text
  fields:
    text:
      type: textarea
```
But when editing the block, there are p tags added to the content. Even while writing in the input field (while the right side modal of the block is opened) there are p tags added to each typed character.

**Expected behavior**  
p tags shouldn't be added to the input field.

**Kirby version**  
3.5

**Desktop**  
 - OS: macOS 10.15.7
 - Browser: Chrome
 - Version: 87

afbora:
## This PR …

Text block input can now be extended more smoothly. It supports all fields.

````yaml
text:
  extends: blocks/text
  fields:
    text:
      type: textarea
      size: large
      spellcheck: false
      buttons: false
````

#### Enhancements

- Now text block field input extendable #3016

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Blocks/Types/Text.vue
+++ b/panel/src/components/Forms/Blocks/Types/Text.vue
@@ -1,10 +1,8 @@
 <template>
-	<k-writer
+	<component
+		:is="component"
 		ref="input"
-		:inline="textField.inline"
-		:marks="textField.marks"
-		:nodes="textField.nodes"
-		:placeholder="textField.placeholder"
+		v-bind="textField"
 		:value="content.text"
 		class="k-block-type-text-input"
 		@input="update({ text: $event })"
@@ -17,10 +15,23 @@
  * @internal
  */
 export default {
-	props: {
-		endpoints: Object
-	},
 	computed: {
+		component() {
+			const type = this.textField.type;
+			if (type === "writer") {
+				return "k-writer";
+			}
+
+			// try load custom component
+			const component = "k-" + type + "-input";
+
+			if (this.$helper.isComponent(component)) {
+				return component;
+			}
+
+			// fallback to writer
+			return "k-writer";
+		},
 		textField() {
 			return this.field("text", {});
 		}
--- a/panel/src/components/Forms/Blocks/Types/Text.vue
+++ b/panel/src/components/Forms/Blocks/Types/Text.vue
@@ -17,13 +17,7 @@
 export default {
 	computed: {
 		component() {
-			const type = this.textField.type;
-			if (type === "writer") {
-				return "k-writer";
-			}
-
-			// try load custom component
-			const component = "k-" + type + "-input";
+			const component = "k-" + this.textField.type + "-input";
 
 			if (this.$helper.isComponent(component)) {
 				return component;

### CODE REVIEW

**Comment by afbora** [comment]
If the PR is appropriate, we can publish it in version `3.7.4`.

**Comment by afbora** [comment]
Maybe we should do as in this PR for all core blocks or find a different general solution. Another extending core block issue: https://github.com/getkirby/kirby/issues/4560

File: panel/src/components/Forms/Blocks/Types/Text.vue [hunk 1]
```diff
@@ -1,10 +1,8 @@
 <template>
-	<k-writer
+	<component
+		:is="component"
 		ref="input"
-		:inline="textField.inline"
-		:marks="textField.marks"
-		:nodes="textField.nodes"
-		:placeholder="textField.placeholder"
+		v-bind="textField || $props"
```
distantnative: What's the fallback to `$props` for? Isn't the only prop `endpoints` and wouldn't fit?

afbora: Yes. `textField` can be empty as `{}`. Thus fallback to default props.

distantnative: But `$props` is just `{ endpoint : ...}` for `Blocks/Types/Text.vue`. Passing that to the component doesn't make sense, does it?

afbora: Yes I know. But if it doesn't make sense to you, we can remove it. I was afraid of breaking things without `endpoints`.

distantnative: But we weren't passing `endpoints` before this PR - do you think it would break?

afbora: I found the source of `endpoints` implementation: https://github.com/getkirby/kirby/commit/81d489d19e5deeab406867bb47eb55cdcd1d112b . I think we doesn't need `endpoints` anymore.

panel/src/components/Forms/Blocks/Types/Text.vue
```diff
@@ -17,10 +15,23 @@
...
 export default {
-	props: {
-		endpoints: Object
-	},
 	computed: {
+		component() {
+			const type = this.textField.type;
+			if (type === "writer") {
+				return "k-writer";
+			}
```
distantnative: We could simplify this further and remove this if statement. Then the code would check for `k-writer-input` which does not exist and still fall back to `k-writer`.

afbora: I thought about it but wanted to avoid any extra checking 👍

distantnative:
@afbora what do you think, is there a general solution or should we adapt each block to be better extendable?

afbora:
The general solution would be: show the default block view (no editable for preview) if it's not the expected type: https://github.com/getkirby/kirby/issues/3016#issuecomment-752271272 But I have no idea about its implementation.

If we do it separately for each block, we will have a richer extending. As in this PR.

bastianallgeier (approved):

PR #4557, opened 2022-08-09
Psalm's security analysis workflow

afbora:
I thought Psalm's security scan would be good. What do you thinking?

https://github.com/psalm/psalm-github-actions
https://github.com/marketplace/actions/psalm-security-scanner-for-php

I saw the information that the workflow report can be hidden (only organization members can access it), but I couldn't set it completely.

**Code scanning alerts**
https://github.com/getkirby/kirby/security/code-scanning?query=is%3Aopen+branch%3Aafbora%2Flab%2Fpsalm-security-analysis

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -242,3 +242,31 @@ jobs:
         with:
           tool_name: PHP-CS-Fixer
           fail_on_error: "true"
+
+  security-analysis:
+    name: "Security Analysis"
+
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+    env:
+      php: "8.0"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # pin@v3
+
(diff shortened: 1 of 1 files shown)


lukasbestle:
We already upload the SARIF file in line 198. Does the action do anything more?

afbora:
@lukasbestle I've removed that step and now seems doesn't working..

lukasbestle:
To be honest I think we are already fully covered regarding Psalm integration in CI.

afbora:
So we already have a security check? I don't remember the workflow making a security notice like below before.

https://github.com/getkirby/kirby/security/code-scanning?query=is%3Aopen+branch%3Aafbora%2Flab%2Fpsalm-security-analysis

lukasbestle:
Yes, we had such a report before. But because we fix them before merging, they don't end up as an active issue in the list.

lukasbestle:
The taint reports are of limited use to us at the moment because Psalm does not understand all code paths yet. That's why we have these disabled at the moment.

afbora:
Got it. So lets close it. Thanks for the review 👍 Maybe we'll take a look here if needed in the future.

PR #4561, opened 2022-08-10
Some Psalm fixes after refactoring

distantnative:


src/Cache/FileCache.php
```diff
@@ -179,7 +179,13 @@ protected function removeEmptyDirectories(string $dir): void
 
 			// checks all directory segments until reaching the root directory
 			while (Str::startsWith($dir, $this->root()) === true && $dir !== $this->root()) {
-				$files = array_diff(scandir($dir) ?? [], ['.', '..']);
+				$files = scandir($dir);
```
distantnative, line 182: `scandir()` return isn't null but false, so fallback to empty array has to be handled differently

src/Data/Data.php
```diff
@@ -54,15 +54,22 @@ public static function handler(string $type): Handler
 		$type = strtolower($type);
 
 		// find a handler or alias
+		$alias   = static::$aliases[$type] ?? null;
 		$handler = static::$handlers[$type] ??
-				   static::$handlers[static::$aliases[$type] ?? null] ??
+				   ($alias ? static::$handlers[$alias] : null) ??
```
distantnative: `static::$handlers[null]` wouldn't be allowed

src/Data/Data.php
```diff
@@ -54,15 +54,22 @@ public static function handler(string $type): Handler
...
-		if ($handler !== null && class_exists($handler)) {
-			return new $handler();
+		if ($handler === null || class_exists($handler) === false) {
+			throw new Exception('Missing handler for type: "' . $type . '"');
 		}
 
-		throw new Exception('Missing handler for type: "' . $type . '"');
+		$handler = new $handler();
+
+		if (is_a($handler, Handler::class) === false) {
```
distantnative, line 67: Double check for handler extending the abstract handler class

src/Image/Image.php
```diff
@@ -111,7 +111,8 @@ public function height(): int
 	 */
 	public function html(array $attr = []): string
 	{
-		return Html::img($this->url(), $attr);
+		$url = $this->url() ?? '';
+		return Html::img($url, $attr);
```
distantnative: `Html::img` only accepts strings

src/Data/Data.php
```diff
@@ -54,15 +54,22 @@ public static function handler(string $type): Handler
 		$type = strtolower($type);
 
 		// find a handler or alias
+		$alias   = static::$aliases[$type] ?? null;
 		$handler = static::$handlers[$type] ??
-				   static::$handlers[static::$aliases[$type] ?? null] ??
+				   ($alias ? static::$handlers[$alias] : null) ??
 				   null;
```
lukasbestle: The `?? null` after the parens doesn't work and will still result in a warning.

```suggestion
				   ($alias ? static::$handlers[$alias] ?? null : null);
```

src/Data/Txt.php
```diff
@@ -40,7 +40,7 @@ public static function encode(mixed $data): string
 	/**
 	 * Helper for converting the value
 	 */
-	protected static function encodeValue(array|string $value): string
+	protected static function encodeValue(array|string|float $value): string
```
lukasbestle, line 43: Where do we pass a `float` at the moment?

src/Image/Image.php
lukasbestle: Does this method make sense when no URL is defined? Shouldn't it rather also return `null` in this case? Would be a breaking change, but I think it makes sense.

src/Text/KirbyTag.php
```diff
@@ -44,8 +45,7 @@ public function __construct(
 			$type = static::$aliases[$type];
 		}
 
-		$kirby    = $data['kirby'] ?? App::instance();
-		$defaults = $kirby->option('kirbytext.' . $type, []);
+		$defaults = $this->kirby()->option('kirbytext.' . $type, []);
```
lukasbestle: The `$data` prop is set below, so it isn't available when `$this->kirby()` is called from here.

src/Data/Txt.php
distantnative: I couldn't track it down but line 49 suggests we kinda support it... I'd rather allow this type without need than miss a usage and break it

src/Image/Image.php
distantnative: Yea was wondering the same but I was afraid it might be too confusing and hard to track down for developers if the whole HTML output disappears/doesn't show up. Clearer to understand that something with the URL must be wrong if the HTML renders with empty `src` attribute.

lukasbestle: Or maybe we should throw an exception that points to the issue?

lukasbestle (approved):

lukasbestle:
All verified. Only the open discussion is left.

src/Image/Image.php
distantnative: I'm fine with throwing an exception instead. I guess that wouldn't be a breaking change, would it? Cause it should throw a PHP error anyways already, right?

lukasbestle: I agree. We already have the `string` type hint in Kirby 3.7.

src/Image/Image.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Kirby\Image;
 
+use Exception;
```
```suggestion
use Kirby\Exception\LogicException;
```

src/Image/Image.php
```diff
@@ -111,7 +112,11 @@ public function height(): int
 	 */
 	public function html(array $attr = []): string
 	{
-		return Html::img($this->url(), $attr);
+		if ($url = $this->url()) {
+			return Html::img($url, $attr);
+		}
+
+		throw new Exception('Calling Image::html() requires that the URL property is not null');
```
```suggestion
		throw new LogicException('Calling Image::html() requires that the URL property is not null');
```

tests/Image/ImageTest.php
```diff
@@ -79,6 +79,17 @@ public function testHtml()
 		$this->assertSame('<img alt="" src="https://foo.bar/cat.jpg">', $file->html());
 	}
 
+	/**
+	 * @covers ::html
+	 */
+	public function testHtmlWithoutUrl()
+	{
+		$this->expectException('\Exception');
```
lukasbestle:
```suggestion
		$this->expectException('Kirby\Exception\LogicException');
```

lukasbestle (approved):

PR #4569, opened 2022-08-12
New `F::unlink()` safest method

### LINKED ISSUE #3039: API cache unlink error
**Describe the bug**  
Strange bug in the api cache, which happens from time to time. I can't figure out what the problem is. My project uses a lot of api caches, which i use inside the template, route or controller. I can just provide screenshots to report the issue, not sure how i can log this? Because the issue is super rare(1 of 100), i can just post what my setup is and the steps to try to reproduce this.
 

**To Reproduce**  
1. Setup a basic api cache, like from the cookbook or take this one. this is located in the home template. 

```php
$named = "y_day_1";
$apiCache = $kirby->cache('api');
$apiDataSave = $apiCache->get($named);
if ($apiDataSave === null) {
    $query = // whatever you wanna store
    $apiDataSave = $query;
    $apiCache->set($named, $apiDataSave, 720);  // 60 = 1h
}
$yoohExists = $apiDataSave;
?>
```

2. On both, first load, and second load the data are loaded/stored properly. In this one i using a 1 day cache time, but this does not matter. I have the same problem with 30min and 1min caches.  Not sure yet this error happens if i deactivate the cache time at all. 

3. You can reload the page as many times you like. As long as the data is served via the cache no issues. The problem is, when the cache is expired and you load the page again, so the api has to rebuild the file and cache.

4. Now, the error happens maybe 1 of 100 times, if step 3 come into play.
5. Error happens -> Error screen
6. Cache file are deleted, and rebuild.
7. Page reload -> Page with new data appears.
8. Everything works except that kirby displays the message with the unlink.

As far as i can tell, the error has nothing todo with the kind of what you caching or the cache time, it's more about how Kirby is handling the check of unlink the file. 

**Additional context**  
Im running all of this inside a vagrant box.

**First guess:**
Maybe it has something todo how vagrant is syncing the files between host and guest system, and when kirby is deleting the new file, vagrant guest is maybe 1ms behind get confused when the unlink is performed? But beside this i have no idea. 
 
 
**Version**  
Kirby 3.5.0
Vagrant debian / PHP 7.4.8
OSX 10.15.7
Safari and Chrome
 

**Screenshots**  
1.
error_1
2.
error_2
3.
error_3

1.-> same error but with another api-cache.
error_other_one

afbora:
### Features

- New `Helpers::handleErrors()` method for custom PHP error handling
- New `F::unlink()` method for idempotent deletion of files and links (native PHP `unlink()` without a warning when the file is already deleted)

### Fixes
- Deleting files throughout the codebase now happens in an idempotent way (if the file is already deleted, no error is thrown) to avoid race conditions, e.g. when cache entries or sessions are cleaned #3039

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -729,11 +729,7 @@ class F
 
 		$file = realpath($file);
 
-		if (file_exists($file) === false) {
-			return true;
-		}
-
-		return unlink($file);
+		return static::unlink($file);
 	}
 
 	/**
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -3,6 +3,7 @@
 namespace Kirby\Filesystem;
 
 use Exception;
+use Kirby\Cms\Helpers;
 use Kirby\Toolkit\I18n;
 use Kirby\Toolkit\Str;
 use Throwable;
@@ -844,6 +845,30 @@ class F
 	public static function typeToExtensions(string $type): ?array
 	{
 		return static::$types[$type] ?? null;
+	}
+
+	/**
+	 * Ensures that a file or link is deleted (with race condition handling)
+	 * @since 3.7.4
+	 */
+	public static function unlink(string $file): bool
+	{
+		return Helpers::handleErrors(
+			fn (): bool => unlink($file),
+			function (&$override, int $errno, string $errstr): bool {
+				// if the file or link was already deleted (race condition),
+				// consider it a success
+				if (Str::endsWith($errstr, 'No such file or directory') === true) {
+					$override = true;
+
+					// drop the warning
+					return true;
+				}
+
+				// handle every other warning normally
+				return false;
+			}
+		);
 	}
 
 	/**
--- a/src/Session/FileSessionStore.php
+++ b/src/Session/FileSessionStore.php
@@ -7,6 +7,7 @@ use Kirby\Exception\Exception;
 use Kirby\Exception\LogicException;
 use Kirby\Exception\NotFoundException;
 use Kirby\Filesystem\Dir;
+use Kirby\Filesystem\F;
 use Kirby\Toolkit\Str;
 
 /**
@@ -322,7 +323,7 @@ class FileSessionStore extends SessionStore
 		}
 
 		// file still exists, delete it
-		if (@unlink($path) !== true) {
+		if (@F::unlink($path) !== true) {
 			// @codeCoverageIgnoreStart
 			throw new Exception([
 				'key'       => 'session.filestore.unexpectedFilesystemError',
--- a/src/Cms/Helpers.php
+++ b/src/Cms/Helpers.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Cms;
 
+use Closure;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Toolkit\Str;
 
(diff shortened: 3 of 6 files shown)


src/Filesystem/F.php
```diff
@@ -846,6 +846,18 @@ public static function typeToExtensions(string $type): ?array
 		return static::$types[$type] ?? null;
 	}
 
+	/**
+	 * Safest way of PHP's native unlink function
```
lukasbestle:
```suggestion
	 * Ensures that a file or link is deleted (with race condition handling)
```

src/Filesystem/F.php
```diff
@@ -846,6 +846,18 @@ public static function typeToExtensions(string $type): ?array
...
+	 */
+	public static function unlink(string $file): bool
+	{
+		try {
+			return unlink($file);
+		} catch (Exception $e) {
+			return false;
```
lukasbestle: Maybe we could check if the file/link is already deleted and treat this case as success. This would catch the race condition error. Also I think we should not swallow the warning by default (to ease debugging e.g. when there are no write permissions or the file system is not available).

```suggestion
			if (file_exists($file) === false) {
				return true;
			}
			
			throw $e;
```

src/Filesystem/F.php
```diff
@@ -846,6 +846,18 @@ public static function typeToExtensions(string $type): ?array
...
 	}
 
+	/**
+	 * Safest way of PHP's native unlink function
+	 */
+	public static function unlink(string $file): bool
+	{
+		try {
+			return unlink($file);
+		} catch (Exception $e) {
```
lukasbestle: Have you verified that catching `Exception` works? AFAIK, a PHP warning is not an exception. Maybe we need to use `set_error_handler()` and `restore_error_handler()` instead (unless they changed the PHP behavior).

src/Session/FileSessionStore.php
```diff
@@ -322,7 +324,7 @@ public function destroy(int $expiryTime, string $id)
 		}
 
 		// file still exists, delete it
-		if (@unlink($path) !== true) {
+		if (F::unlink($path) !== true) {
```
lukasbestle: If we don't swallow the warning in `F::unlink()` (see comment above), we need to do it here and keep the `@` error control operator.

src/Filesystem/F.php
afbora: I started the PR with this change first but the tests failed. I guess `file_exists` fails on link files. Do you have an idea?

lukasbestle: Ah, right. `file_exists()` mentions dirs and files, but not links.

Maybe this: `is_file($file) === false && is_link($file) === false`

afbora: I didn't include it to avoid unnecessary use.  So, do you think there is also a need to use `realpath` before file checks?

lukasbestle: Oh wait. That doesn't work as `is_file()` resolves links (only checks for the file "behind the link") and `is_link()` does not support hardlinks. 🤯

lukasbestle: Maybe we could check for the "No such file or directory" warning that `unlink()` triggers. But then we need to find a way to handle translated warning messages if a translated PHP installation is used.

lukasbestle: Hm, are translated PHP installations still a thing actually? Can't find any trace of this feature but I'm pretty sure I have seen a German PHP error at some weird hosting provider 10 years ago.

src/Filesystem/F.php
afbora: I've tested on my local machine and unit tests. `try catch` seems works great! 🥳 🤷‍♂️ 

- file exists case ✅ 
- no file case ✅ 
- file exists with no permission case ✅

afbora: When I removed the `try catch` block, throws exception (debug enabled). Offline page shows on debug mode disabled.

afbora: I have an idea about that I think `whoops` handle this. Is it possible? Because `Whoops\Exception\ErrorException` throws when `unlink` fails.

src/Filesystem/F.php
afbora: I don't think we really need translation of PHP exceptions. It could happen if there were constants defined like in `UPLOAD_ERR_*`: https://github.com/getkirby/kirby/blob/3.7.3/src/Api/Api.php#L744-L752

src/Filesystem/F.php
afbora: The PR seems ready when we are sure about try catch in unlink error handler.

lukasbestle: Yes, I agree that this must be because of Whoops. If I set the `whoops` option to `false`, it doesn't work. So I think we should really use an error handler instead for maximum compatibility.

src/Filesystem/F.php
lukasbestle: What I meant was to check the message of the warning PHP triggers. If the warning message is "No such file or directory", then we know it's fine and we can `return true`. Otherwise, we would throw/trigger that warning again and return `false`.

But such a check for the warning message would break if that message is already translated by PHP.

afbora:
@lukasbestle I'm out of city this week so I won't be able to complete the PR for 3.7.4. We can move it to 3.7.5 if you can't take care or you're not available.

lukasbestle:
I have implemented it with an error handler as we discussed. Unit tests are still missing. I'll add those if you agree on the solution.

afbora:
TBH I didn't use set_error_handler before and I can't test the PR now. I trust your solution 🙂

distantnative:
Not sure either how to test this 😅

lukasbestle:
Don't worry, we have unit tests 🙂

distantnative (approved):

PR #4575, opened 2022-08-14
`Http\Uri`: definable schemes

distantnative:
### Refactoring
- Allows modifying allowed schemes for `Http\Uri` when extending via `Uri::$schemes` property

## Ready?

### For review team

### DIFF
--- a/src/Http/Uri.php
+++ b/src/Http/Uri.php
@@ -91,6 +91,13 @@ class Uri
 	protected $scheme = 'http';
 
 	/**
+	 * supported schemes
+	 *
+	 * @var array
+	 */
+	protected static $schemes = ['http', 'https', 'ftp'];
+
+	/**
 	 * @var bool
 	 */
 	protected $slash = false;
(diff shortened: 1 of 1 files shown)


afbora:
For users or an internal extending? I just wonder because of `protected` property.

distantnative:
For my use case (UUIDs) it would work with protected (as I am extending `Uri` and then the child class can redefine the allowed schemes. But if you think this would be useful as public, we can do that too.

Maybe @lukasbestle has a thought if that would be a security risk potentially?

lukasbestle (commented):
I think it's sensible to keep this protected just for internal use. Different URI types should have a separate class to avoid confusion.

We could even make it a class constant to make this internal nature even more clear.

distantnative:
@lukasbestle I would leave it a property as for the UUID uri we might want to make it a public property to allow developers extending the supported UUID schemes... a bit far fetched, but I think leaving it as property would be fine then.

lukasbestle (approved):

PR #4577, opened 2022-08-14
Throw user back to login page on challenge expiry

### LINKED ISSUE #4087: Login with timed-out code should throw user out
## Description
When trying to log in via code and the code has timed out, there is no way back to the login form (back button doesn't do anything) and no way to resend the code. The only way out for the user seems to be to kill the session manually.

**Expected behavior**  
There should either be a "Resend code" button or the back button should bring the user back to the login form or  the user would be redirected to the login form.

## To reproduce

1. Enable login with code in config
2. Go the `/panel/login` and request a code
3. Wait long enough for the code to time out
4. Try to log in with the timed-out code
5. See error message
6. Try the back button => nothing
7. Refresh page => you still get the form that requests the login code

## Your setup
**Kirby 3.6.1.1**  

 - Device: MacBook Pro (14", 2021)
 - OS: macOS 12.1
 - Browser: Firefox
 - Version: 95.0.2

lukasbestle:
### Enhancements

- When a login challenge has expired, the user is redirected to the login page #4087

### Refactoring

- Custom login views no longer need to display errors themselves with `k-login-alert`, instead they can `this.$emit("error", error)`

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Views/LoginView.vue
+++ b/panel/src/components/Views/LoginView.vue
@@ -61,6 +61,13 @@ export default {
 				return;
 			}
 
+			if (error.details.challengeDestroyed === true) {
+				// reset from the LoginCode component back to Login
+				await this.$reload({
+					globals: ["$system"]
+				});
+			}
+
 			this.issue = error.message;
 		}
 	}
--- a/panel/src/components/Views/LoginView.vue
+++ b/panel/src/components/Views/LoginView.vue
@@ -1,14 +1,20 @@
 <template>
 	<k-panel>
-		<k-view v-if="form === 'login'" align="center" class="k-login-view">
-			<k-login-plugin :methods="methods" />
-		</k-view>
-		<k-view
-			v-else-if="form === 'code'"
-			align="center"
-			class="k-login-code-view"
-		>
-			<k-login-code v-bind="$props" />
+		<k-view align="center" :class="viewClass">
+			<!-- <div> as a wrapper so that <k-view>
+			     has a single child for Flexbox layout -->
+			<div>
+				<h1 class="sr-only">
+					{{ $t("login") }}
+				</h1>
+
+				<k-login-alert v-if="issue" @click="issue = null">
+					{{ issue }}
+				</k-login-alert>
+
+				<k-login-code v-if="form === 'code'" v-bind="$props" @error="onError" />
+				<k-login-plugin v-else :methods="methods" @error="onError" />
+			</div>
 		</k-view>
 	</k-panel>
 </template>
@@ -24,21 +30,39 @@ export default {
 		methods: Array,
 		pending: Object
 	},
+	data() {
+		return {
+			issue: ""
+		};
+	},
 	computed: {
 		form() {
 			if (this.pending.email) {
 				return "code";
 			}
 
-			if (!this.$user) {
-				return "login";
+			return "login";
+		},
+		viewClass() {
+			if (this.form === "code") {
+				return "k-login-code-view";
 			}
 
-			return null;
+			return "k-login-view";
 		}
 	},
 	created() {
 		this.$store.dispatch("content/clear");
+	},
+	methods: {
+		async onError(error) {
+			if (error === null) {
+				this.issue = null;
+				return;
+			}
+
+			this.issue = error.message;
+		}
 	}
 };
 </script>
--- a/panel/src/components/Forms/Login.vue
+++ b/panel/src/components/Forms/Login.vue
@@ -1,13 +1,5 @@
 <template>
 	<form class="k-login-form" @submit.prevent="login">
-		<h1 class="sr-only">
-			{{ $t("login") }}
-		</h1>
-
-		<k-login-alert v-if="issue" @click="issue = null">
-			{{ issue }}
-		</k-login-alert>
-
 		<div class="k-login-fields">
 			<button
 				v-if="canToggle === true"
@@ -51,7 +43,6 @@ export default {
 		return {
 			currentForm: null,
 			isLoading: false,
-			issue: "",
 			user: {
 				email: "",
 				password: "",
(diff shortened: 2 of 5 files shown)


panel/src/components/Views/LoginView.vue
```diff
@@ -24,21 +30,46 @@ export default {
...
 		form() {
 			if (this.pending.email) {
 				return "code";
 			}
 
-			if (!this.$user) {
-				return "login";
+			return "login";
+		},
+		viewClass() {
```
distantnative, line 46: Is there a reason why we need these classes on `k-view` and cannot simply stick them on their own components?

lukasbestle: They might be necessary for custom login form components. E.g. if you need an entirely different layout that isn't centered. Also changing it now would be a breaking change.

distantnative: But those wouldn't have control over which classes to add anyways, it's always just those two from us. So not entering would and could be done via CSS anyways. It's a bit of a breaking change - but not more than many of the other changes in this PR. So I think that would be fine.

lukasbestle: Yes, the idea is: A user defines a custom login form component and then also needs CSS changes to the view. Without the class on the view, there's no way to style it.

lukasbestle: > It's a bit of a breaking change - but not more than many of the other changes in this PR

Are there any other breaking changes in the PR?

distantnative: Ok, then we should change this in 3.8 only.

Since people extend all kind of things in the Panel, almost all changes are kind of breaking changes - e.g. removing `this.issue` and corresponding functionality from the child components. I'm totally on your side that we need to make these and that the risk is low - but technically the Panel is very undefined in terms of what developers can rely on, what's off limits etc - and so almost any change becomes a breaking change in some way.

lukasbestle: I first also thought that removing `this.issue` from core child components was a breaking change, but it isn't. If the child component displays its own `LoginAlert`, that's completely fine. Only the `LoginCode` component must not do that, but for that we don't have a plugin API.

distantnative: But that's what I mean, the Panel isn't really defined in terms of plugin APIs - we've encouraged users to extend and overwrite all kind of components. Hence, if someone extends LoginCode it might work unexpectedly different for them now.

lukasbestle: We do have a plugin API for Login, but none for LoginCode.
I'd still be fine to move the fix to 3.8.0.

distantnative: We do have `plugins.components` which we have encouraged a lot to do, so also `LoginCode` can be extended/replaced via that API.

distantnative:
Seems to work for me in a way, except that I get thrown back to the initial login already after about 5 minutes and not 10 minutes like the email says:
Screenshot 2022-08-27 at 18 27 49

lukasbestle:
@distantnative I think that's the [ping to the API that happens every five minutes](https://github.com/getkirby/kirby/blob/44158e8a5856384f35b07263240abc96db16ae7e/panel/src/config/api.js#L61). But that itself shouldn't actually throw you back to the login page (because that error response doesn't contain the `challengeDestroyed` value in `details`). However for some reason it does (I can reproduce the issue).

Some other part of the Panel seems to reload the view whenever this ping request fails.

distantnative:
@lukasbestle could it be this? https://github.com/getkirby/kirby/blob/main/panel/src/config/api.js#L23-L29

distantnative:
And this would be the 5 minute interval: https://github.com/getkirby/kirby/blob/main/panel/src/config/api.js#L54

Maybe aligning it to 10 minutes could make sense. Or only setting that interval after logging in (and login code view having their own check interval)?

lukasbestle:
Yep, confirmed by commenting these lines out.

For the login view (= Panel state where no user is currently logged in) we either need to disable the periodic ping entirely or we need to prevent the redirection to logout.

But that's a fix unrelated to this PR. If you want to test just this PR, you can set the challenge timeout e.g. to one minute with this option: https://getkirby.test/docs/reference/system/options/auth#authentication-challenges__code-timeout

lukasbestle:
Moved that bug to https://github.com/getkirby/kirby/issues/4613.

distantnative:
It's actually not unrelated. If I set the challenge timeout to 2 minutes, nothing happens after 2 minutes. Only after 5 minutes and that is then due to the regular authentication ping, not specifically related to the login code calling expiry.

lukasbestle:
Nothing is supposed to happen after two minutes. But if you then enter a code, you get redirected to the login page. Instead of the previous behavior that kept you on the code view forever. The code view does not ping the backend regularly, it can only know about the expired challenge if the backend tells it about it.

distantnative:
Ok, I would rephrase the release notes then, cause "when expires" and "automatically redirected" let me assume that this happens without any further interaction.

However, I set the timeout to 1 minute. Now when I enter the code after two minutes, this error message appears (if that's intended, message should be easier to understand). I still remain on the code input - I'm not thrown back to the main login form. Only when I click on the back button myself.

Screenshot 2022-08-27 at 21 12 34

lukasbestle:
Good point, I updated the release notes.

The error message is only displayed during debugging. Otherwise you just get "Invalid code".

I cannot reproduce the issue that you stay on the code view. Does the error in console include the `details.challengeDestroyed = true` flag? If so, please verify your frontend dist files are up to date.

lukasbestle:
This is what I get with a timeout of 10 seconds:

https://user-images.githubusercontent.com/1595007/187045185-0f48c241-f37d-4517-b806-210dfaab80ea.mov

distantnative (approved):
Ok now it worked for me as well. Might've been an issue between running in `rpm dev` and with list files.

PR #4578, opened 2022-08-14
Re-enable frontend coding style checks in CI

lukasbestle:
### Enhancements
- `k-text` has a new `html` prop to pass content that should be rendered as HTML (instead of providing text via the default slot). Use carefully as rendering HTML can be a gateway for XSS attacks.

### Refactoring

- Our CI now checks frontend coding style again

### Breaking changes

None

### For review team

### DIFF
--- a/panel/src/components/Layout/Overlay.vue
+++ b/panel/src/components/Layout/Overlay.vue
@@ -12,7 +12,7 @@
 			@mousedown="close"
 		>
 			<k-loader v-if="loading" class="k-overlay-loader" />
-			<slot v-else :close="close" :isOpen="isOpen" />
+			<slot v-else :close="close" :is-open="isOpen" />
 		</div>
 	</portal>
 </template>
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -80,6 +80,7 @@ jobs:
           cache-dependency-path: panel/package-lock.json
 
       - name: Install npm dependencies
+        id: finishPrepare
         run: npm ci
         working-directory: panel
 
--- a/panel/src/components/Dialogs/FormDialog.vue
+++ b/panel/src/components/Dialogs/FormDialog.vue
@@ -8,8 +8,7 @@
 		@submit="$refs.form.submit()"
 	>
 		<template v-if="text">
-			<!-- eslint-disable-next-line vue/no-v-html -->
-			<k-text v-html="text" />
+			<k-text :html="text" />
 		</template>
 		<k-form
 			v-if="hasFields"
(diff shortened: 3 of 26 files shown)


panel/src/components/Layout/Overlay.vue
```diff
@@ -12,7 +12,7 @@
 			@mousedown="close"
 		>
 			<k-loader v-if="loading" class="k-overlay-loader" />
-			<slot v-else :close="close" :isOpen="isOpen" />
+			<slot v-else :close="close" :is-open="isOpen" />
```
lukasbestle, line 15: Not sure what this change does TBH

lukasbestle:
There are two sets of errors still left:

- **Using v-html on component may break component's content** (https://github.com/vuejs/eslint-plugin-vue/issues/1724)
  These `v-html` uses may actually be an issue and we need to verify each one
- **Prop "value" should define at least its type** for `TableCell.vue`
  This prop is explicitly mixed; is there a way to define a mixed type in Vue or do we need to silence that error?

panel/src/components/Layout/Overlay.vue
johannschopplich: [vue/attribute-hyphenation](https://eslint.vuejs.org/rules/attribute-hyphenation.html) requires the props to be hyphenated.

distantnative:
Regarding `vue/no-v-text-v-html-on-component`, all usages are on `k-text` which is fine as it is a single div-container anyways filled by the slot, so filling it with HTML doesn't break anything. We could add an `html` prop to `k-text` though and then use `v-html` in `Text.vue`, minimising it to one mention in our code.

Regarding `vue/require-prop-types`, we might just define all possible types then: String, Number
Boolean, Array, Object, Date, Function, Symbol. However, this could lead to issues maybe if a certain value is casted to a specific type because the order we mention them (not sure if that's really a case). Could be an argument for just silencing this one.

distantnative:
Like this https://github.com/getkirby/kirby/commit/ad4c97455b8c16286486c76c8b093aeb8296b44b

lukasbestle:
Both sounds great! 👍

> Could be an argument for just silencing this one.

I agree.

panel/src/components/Layout/Overlay.vue
lukasbestle: Yes, I understand that part. But I was wondering if the different syntax could break behavior.

johannschopplich: It shouldn't. As I understand it, the Vue compiler will handle pascal case from props and attribute case in templates. The compiler is case-tolerant, that's why it's working currently as well.

distantnative (approved):

PR #4582, opened 2022-08-18
Ease deprecation of `headline` option

distantnative:
We decided that this is a bit too harsh and we will delay error messages for now.

### DIFF
--- a/config/presets/files.php
+++ b/config/presets/files.php
@@ -5,7 +5,7 @@ use Kirby\Toolkit\I18n;
 return function (array $props) {
 	$props['sections'] = [
 		'files' => [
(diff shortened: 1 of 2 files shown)


config/presets/files.php
```diff
@@ -5,7 +5,6 @@
 return function (array $props) {
 	$props['sections'] = [
 		'files' => [
-			// TODO: Remove `headline` check in 3.9.0
 			'label'    => $props['headline'] ?? $props['label'] ?? I18n::translate('files'),
```
afbora: Lets change the `label` and `headline` index order.

afbora (approved):

PR #4593, opened 2022-08-21
Fix `F::remove()` with non-existing files

lukasbestle:


src/Filesystem/F.php
```diff
@@ -728,6 +728,9 @@ public static function remove(string $file): bool
 		}
 
 		$file = realpath($file);
+		if (is_string($file) === false) {
```
afbora, line 731: Not important. I just wonder why you prefer `is_string` check instead simply `$file ===  false`? About easy code reading/reviewing?

afbora (approved):

src/Filesystem/F.php
lukasbestle: At the moment it's the same. My thought was that if `realpath()` ever returns something else than `string|false` for whatever reason, it's safer. Because `F::unlink()` only supports strings.

afbora: I know it's the same. I was just curious because `realpath()` doesn'treturn any value other than `string|false`.

PR #4596, opened 2022-08-23
Add `playsinline` attribute for videos if `autoplay` is set

### LINKED ISSUE #4595: Video Kirbytag should render `playsinline` attribute
## Description
I’m not 100% sure if this should be considered a bug or a feature request. When setting `autoplay: true`, the video Kirbytag renders `autoplay` and `muted` attributes. On iOS, a third attribute, `playsinline` is required:

> Video elements that include `<video autoplay>` play automatically when the video loads in Safari on macOS and iOS, only if those elements also include the playsinline attribute. – [Source](https://developer.apple.com/documentation/webkit/delivering_video_content_for_safari#3030251)

That means that HTML videos rendered by the video Kirbytag won’t start playback automatically on iOS.

**Expected behavior**  
* The video Kirbytag should have a `playsinline` attribute.
* If the `playsinline` attribute isn’t explicitly set, it should default to `true` if `autoplay` is set, and `false` otherwise. (This is behavior is analogous to the `mute` attribute.)

## To reproduce

1. Render the following Kirbytext:
   ```
   (video: https://storage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4 autoplay: true)
   ```
2. Open the page in iOS Safari.
3. The playback should start automatically, but it doesn’t.

## Your setup
**Kirby Version**  
3.7.1

 - Device: iPHone SE 2 (Xcode simulator)
 - OS: iOS 15.5
 - Browser: Safari
 - Version: 15.5

**Additional context**
I have a branch with a fix ready and can open a PR if you consider this to be a sensible change.

tillprochaska:
### Enhancements

- The `video` KirbyTag will automatically add the `playsinline` attribute whenever the `autoplay` option is enabled #4595

### Breaking changes

None relevant for release notes

## More details

The `playsinline` attribute is required in order to start video playback automatically on iOS:

> Video elements that include `<video autoplay>` play automatically when the
> video loads in Safari on macOS and iOS, only if those elements also include the
> `playsinline` attribute.

Source: https://developer.apple.com/documentation/webkit/delivering_video_content_for_safari#3030251

* This PR adds a `playsinline` attribute to the video Kirbytag.
* The attribute defaults to `true`, if `autoplay: true` is set, and `false` otherwise.

Closes #4595.

### Breaking changes
This changes how `(video: my-video.mp4 autoplay: true)` is rendered, so technically, this is a breaking change. That means that a video will be played inline (and not in full screen mode) on mobile user agents such as iOS and Android. 

```diff
- <video autoplay controls muted>...</video>
+ <video autoplay controls muted playsinline>...</video>
```
However, this won’t break a websites layout or cause other errors. I do think this is actually closer to end users expectations when using the `autoplay` attribute on the video Kirbytag.

## Ready?

### For review team

### DIFF
--- a/config/tags.php
+++ b/config/tags.php
@@ -289,13 +289,13 @@ return [
 				// booleans: autoplay, controls, loop, muted
 				// strings : poster, preload
 				// for ex  : `autoplay` will not work if `false` is a `string` instead of a `boolean`
-				$attrs['autoplay'] = $autoplay = Str::toType($tag->autoplay, 'bool');
-				$attrs['controls'] = Str::toType($tag->controls ?? true, 'bool');
-				$attrs['loop']     = Str::toType($tag->loop, 'bool');
-				$attrs['muted']    = Str::toType($tag->muted ?? $autoplay, 'bool');
+				$attrs['autoplay']    = $autoplay = Str::toType($tag->autoplay, 'bool');
+				$attrs['controls']    = Str::toType($tag->controls ?? true, 'bool');
+				$attrs['loop']        = Str::toType($tag->loop, 'bool');
+				$attrs['muted']       = Str::toType($tag->muted ?? $autoplay, 'bool');
 				$attrs['playsinline'] = Str::toType($tag->playsinline ?? $autoplay, 'bool');
-				$attrs['poster']   = $tag->poster;
-				$attrs['preload']  = $tag->preload;
+				$attrs['poster']      = $tag->poster;
+				$attrs['preload']     = $tag->preload;
 			}
 
 			// handles local and remote video file
--- a/config/tags.php
+++ b/config/tags.php
@@ -246,6 +246,7 @@ return [
 			'height',
 			'loop',
 			'muted',
+			'playsinline',
 			'poster',
 			'preload',
 			'style',
@@ -292,6 +293,7 @@ return [
 				$attrs['controls'] = Str::toType($tag->controls ?? true, 'bool');
 				$attrs['loop']     = Str::toType($tag->loop, 'bool');
 				$attrs['muted']    = Str::toType($tag->muted ?? $autoplay, 'bool');
+				$attrs['playsinline'] = Str::toType($tag->playsinline ?? $autoplay, 'bool');
 				$attrs['poster']   = $tag->poster;
 				$attrs['preload']  = $tag->preload;
 			}
--- a/tests/Text/KirbyTagsTest.php
+++ b/tests/Text/KirbyTagsTest.php
@@ -535,18 +535,18 @@ class KirbyTagsTest extends TestCase
 			'options' => [
 				'kirbytext' => [
 					'video' => [
-						'autoplay' => true,
-						'caption'  => 'Lorem ipsum',
-						'controls' => false,
-						'class'    => 'video-class',
-						'height'   => 350,
-						'loop'     => true,
-						'muted'    => true,
+						'autoplay'    => true,
+						'caption'     => 'Lorem ipsum',
+						'controls'    => false,
+						'class'       => 'video-class',
+						'height'      => 350,
+						'loop'        => true,
+						'muted'       => true,
 						'playsinline' => true,
-						'poster'   => 'sample.jpg',
-						'preload'  => 'auto',
-						'style'    => 'border: none',
-						'width'    => 500
+						'poster'      => 'sample.jpg',
+						'preload'     => 'auto',
+						'style'       => 'border: none',
+						'width'       => 500
 					]
 				]
 			],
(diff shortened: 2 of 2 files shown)


config/tags.php
```diff
@@ -292,6 +293,7 @@
 				$attrs['controls'] = Str::toType($tag->controls ?? true, 'bool');
 				$attrs['loop']     = Str::toType($tag->loop, 'bool');
 				$attrs['muted']    = Str::toType($tag->muted ?? $autoplay, 'bool');
+				$attrs['playsinline'] = Str::toType($tag->playsinline ?? $autoplay, 'bool');
```
afbora, line 296: > Breaking changes
> This changes how (video: my-video.mp4 autoplay: true) is rendered, so technically, this is a breaking change. That means that a video will be played inline (and not in full screen mode) on mobile user agents such as iOS and Android.

So when `autoplay` is enabled, should the `playsinline` be enabled too?

If we don't bind the `playsinline` attribute to `autoplay`, I guess there will be no breaking change if `playsinline` manual is enabled, right?

tillprochaska: > So when autoplay is enabled, should the playsinline be enabled too?

From the perspective of a rather non-technical user, the expectation is that adding `(video: my-video.mp4 autoplay: true)` will actually render a video that starts playback automatically. However, that is not the case (at least for iOS users), as in addition to the `autoplay` attribute  `playsinline` and `muted` are required in order to actually enable automatic playback on iOS.

> If we don't bind the playsinline attribute to autoplay, I guess there will be no breaking change if playsinline manual is enabled, right?

Yes, that’s true! I do think that defaulting to the value of the `autoplay` is sensible though for the UX reason outlined above. Also, the same logic is already implemented for the `muted` attribute, and I think it would be nice for both attributes to behave consistently.

distantnative: I'm not yet familiar with this new attribute. Does it still allow users to open full-screen mode when they want to really watch the video and listen to the audio? Or does it suppress that completely?

tillprochaska: Hi @distantnative, as long as the `controls` attribute is present (the video Kirbytag renders it by default), users have access to all the default playback controls, including volume and fullscreen toggles.

The video below shows the behavior on iOS for this HTML snippet:

```html
<video autoplay muted playsinline controls width="100%">
  <source type=video/mp4 src=https://storage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4 />
</video>
```

https://user-images.githubusercontent.com/1512805/186371452-1d382560-1c1c-499d-92a4-9034f50ecf2d.mov

distantnative: Ok then I think it is a small breaking change but one that makes a lot of sense to me. I'd suggest to include it into 3.8

tillprochaska: Forgot to mention one more thing: On desktop browsers, `playsinline` doesn’t have any effect at all, i.e. it does not break existing behavior, as all major desktop browser play videos inline by default:

> The absence of the `playsinline` attribute does not imply that the video will display fullscreen by default. Indeed, most user agents have chosen to play all videos inline by default, and in such user agents the `playsinline` attribute has no effect.

https://html.spec.whatwg.org/multipage/media.html#attr-video-playsinline

To summarize:

* On desktop browsers, videos are dispalyed inline by default, there’s no breaking change.
* I’ve also tested this in Chrome and Firefox on Android 11. Videos are displayed inline by default as well.
* On iOS, `(video: my-video.mp4 autoplay: true)` previously had no effect because of the missing `playsinline` attribute.

So the breaking changes will be limited to very few browsers, if any.

distantnative (dismissed):

lukasbestle (commented):
Could you please also add a unit test for `(video: sample.mp4 autoplay: true)` that tests the automatically added `muted` and `playsinline` attributes?

distantnative:
Can't really add to this PR from a fork, but would be like this

```php
public function testVideoAutoplayRelatedAttrs()
{
	$kirby = new App([
		'roots' => [
			'index' => '/dev/null',
		],
		'site' => [
			'children' => [
				[
					'slug' => 'test',
					'content' => [
						'text' => '(video: sample.mp4 autoplay: true)'
					],
					'files' => [
						['filename' => 'sample.mp4']
					]
				]
			]
		]
	]);

	$page  = $kirby->page('test');
	$video = $page->file('sample.mp4');

$expected = '<figure class="video"><video autoplay controls muted playsinline><source src="' . $video->url() . '" type="video/mp4"></video></figure>';
	$this->assertSame($expected, $page->text()->kt()->value());
}
```

lukasbestle (approved):
Added the unit test with our maintainer write permissions

tillprochaska:
@distantnative @lukasbestle Thanks for taking care of the changes! 🙂

lukasbestle:
@tillprochaska Thanks for your PR 👍

PR #4606, opened 2022-08-26
Disable CI Codecov reporting in fork runs

lukasbestle:
### Fixes

- In GitHub forks of the `kirby` repository, the CI no longer fails at the Codecov reporting step

### For review team

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -111,6 +111,10 @@ jobs:
         run: psalm --output-format=github --php-version=${{ matrix.php }} --report=sarif/psalm.sarif --report-show-info=false
 
       - name: Upload coverage results to Codecov
+        env:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          PHP: ${{ matrix.php }}
+        if: env.token != ''
         uses: codecov/codecov-action@66b3de25f6f91f65eb92c514d31d6b6f13d5ab18 # pin@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/backend.yml
```diff
@@ -111,15 +111,17 @@ jobs:
         run: psalm --output-format=github --php-version=${{ matrix.php }} --report=sarif/psalm.sarif --report-show-info=false
 
       - name: Upload coverage results to Codecov
+        env:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          PHP: ${{ matrix.php }}
+        if: env.token != ''
```
afbora, line 117: I know it's extreme but what if the forked repo has the secret value `CODECOV_TOKEN`?

If the above situation is possible, the repository name can be checked. 

````yaml
if: ${{ github.repository == 'getkirby/kirby' }}
````

TBH I'm not sure. I'm just thinking out loud.

lukasbestle: I tried that as well, but that would break for the PR run. In this case we *do* want the Codecov reporting. My if logic to fix this required three lines, so this is much simpler.

Also I think it's not an issue. If the secret value is set, then Codecov will work as expected.

afbora: That's right, we want it to work successfully, not throw an error and fail after all 👍

afbora (approved):

PR #4608, opened 2022-08-27
Upgrade dependencies

distantnative:
## This PR …

I checked composer and npm upgrades release notes individually.
You probably need to update your local setups to match the cs fixer version and similar.

### Enhancements
- Upgraded dependencies

## Ready?

### For review team


distantnative:
@lukasbestle Do you know why the composer stuff in the CI does not install now?

.github/workflows/backend.yml
```diff
@@ -174,7 +174,7 @@ jobs:
           extensions: ${{ env.extensions }}
           coverage: none
           tools: |
-            composer:2.3.7, composer-normalize:2.28.0, composer-require-checker:4.0.0,
+            composer:2.4.1, composer-normalize:2.28.3, composer-require-checker:4.1.0,
```
afbora: I've done [few tests](https://github.com/getkirby/kirby/tree/afbora/test-dependencies). `composer-normalize` doesn't support composer versions newer than `2.3.8`.

Actually it seems to support 2.3.9 https://github.com/ergebnis/composer-normalize/pull/942 but I guess there is another problem

distantnative: thanks! downgraded for now

src/Cache/FileCache.php
```diff
@@ -79,19 +79,19 @@ protected function file(string $key): string
 				case '/':
 					break;
 
-				// backslashes get their own marker in the path
-				// to differentiate the cache key from one with forward slashes
+					// backslashes get their own marker in the path
+					// to differentiate the cache key from one with forward slashes
```
lukasbestle, line 84: We need to move these comments below the `case` statements. With the new indentation they are quite confusing.

src/Cms/Email.php
```diff
@@ -123,7 +122,7 @@ protected function template(): void
 					$this->props['body']['text'] = $text->render($data);
 				}
 
-				// fallback to single email text template
+			// fallback to single email text template
```
lukasbestle, line 125: Same here

src/Database/Sql.php
```diff
@@ -708,11 +707,11 @@ public function splitIdentifier($table, $identifier): array
 			case 1:
 				return [$table, $this->unquoteIdentifier($parts[0])];
 
-			// qualified identifier
+				// qualified identifier
```
lukasbestle: And here

src/Toolkit/Html.php
```diff
@@ -594,7 +594,7 @@ public static function youtube(string $url, array $options = [], array $attr = [
 
 				break;
 
-			// regular video URLs
+				// regular video URLs
```
lukasbestle: And here

lukasbestle (commented):
Wondering why the frontend and integration workflows are not executed on the `push` event. Because the `package.json` was changed, they should have.

src/Cache/FileCache.php
afbora: To be honest, these indentations are getting on my nerves 😩

lukasbestle: We only need to fix them once by moving the comments, then we should be fine in the long term.

src/Cms/Email.php
distantnative: I think this one makes actually sense as is

distantnative:
@lukasbestle Interestingly, from the frontend workflow file the code style runs, the unit tests get skipped. SO it's probably not related to the `on:` config of the frontend file. Maybe it's this part:

```
    if: >
      github.event_name != 'pull_request' ||
      github.event.pull_request.head.repo.full_name != github.repository
```

? We have the same in the E2E tests, so if that if statement is somehow flawed, could be that.

lukasbestle:
We also have the same in the `backend` workflow file. It shouldn't prevent the action from being run on `push` though as on `push` the event name is `!= 'pull_request'`.

src/Cms/Email.php
lukasbestle: Ah, right. Missed that this one was changed in the opposite direction. 👍

distantnative:
Though if I read it correctly, all pull_request ones are skipped for backend as well, only the push ones did run for backend.

lukasbestle:
Yes, that's what that `if:` line does. For backend it works, but for frontend it doesn't. Frontend doesn't even run on `push`.

lukasbestle (approved):
I think we can still merge this PR already. I think the frontend workflow issue is independent from the changes in this PR. It only surfaced here.

PR #4609, opened 2022-08-27
Replace `is_a()` with `instanceof`

distantnative:
## This PR …

To improve static analysis.
Also more performant.

### Refactoring
- Use PHP's `instanceof` instead of `is_a()`

## Ready?

### For review team

### DIFF
--- a/src/Form/Options.php
+++ b/src/Form/Options.php
@@ -87,7 +87,7 @@ class Options
 
 		// add the model by the proper alias
 		foreach (static::aliases() as $className => $alias) {
-			if (is_a($model, $className) === true) {
+			if ($model instanceof $className) {
 				$data[$alias] = $model;
 			}
 		}
--- a/config/components.php
+++ b/config/components.php
@@ -4,7 +4,9 @@ use Kirby\Cms\App;
 use Kirby\Cms\Collection;
 use Kirby\Cms\File;
 use Kirby\Cms\FileVersion;
+use Kirby\Cms\Page;
 use Kirby\Cms\Template;
+use Kirby\Cms\User;
 use Kirby\Data\Data;
 use Kirby\Email\PHPMailer as Emailer;
 use Kirby\Filesystem\F;
@@ -177,11 +179,11 @@ return [
 			$keys = array_keys($data);
 			$keys[] = 'id';
 
-			if (is_a($item, 'Kirby\Cms\User') === true) {
+			if ($item instanceof User) {
 				$keys[] = 'name';
 				$keys[] = 'email';
 				$keys[] = 'role';
-			} elseif (is_a($item, 'Kirby\Cms\Page') === true) {
+			} elseif ($item instanceof Page) {
 				// apply the default score for pages
 				$options['score'] = array_merge([
 					'id'    => 64,
--- a/config/fields/mixins/upload.php
+++ b/config/fields/mixins/upload.php
@@ -51,7 +51,7 @@ return [
 				$parent = $this->model();
 			}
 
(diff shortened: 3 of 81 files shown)


afbora:
You're right. `instanceof` seems more performant. Here is performance results of [is_a()](https://php.net/is_a) and [instanceof](https://php.net/instanceof):

````

Test name       Repeats         Result          Performance     
instanceof      10000           0.028343 sec    +0.00%
is_a()          10000           0.043927 sec    -54.98%
````
Test source is [here](https://github.com/nafigator/php-bench/blob/master/Tests/IsAVsInstanceOf.php).

**Comment by distantnative** [comment]
@afbora is that still the case for PHP 7/8? I had seen a similar (the same?) stats but then also read that this might only be true for e.g. PHP 5

**Comment by afbora** [comment]
I've done local test, here the results 🎉 

### PHP 8.0.3 (~%80-110)
````bash
PHP 8.0.3 (cli) (built: Mar  2 2021 23:34:05) ( ZTS Visual C++ 2019 x64 )
Copyright (c) The PHP Group
Zend Engine v4.0.3, Copyright (c) Zend Technologies

[============================================================>] 10000 u | 1162147 u/s | Est: 0.0 s | Mem: 569.93 KB | Max: 608.80 KB
[============================================================>] 10000 u | 1666787 u/s | Est: 0.0 s | Mem: 573.91 KB | Max: 612.27 KB
Test name       Repeats         Result          Performance
instanceof      10000           0.001720 sec    +0.00%
is_a()          10000           0.003462 sec    -101.28%
````
### PHP 8.1.1 (~%80-115)
````bash
PHP 8.1.0 (cli) (built: Nov 23 2021 21:43:17) (NTS Visual C++ 2019 x86)
Copyright (c) The PHP Group
Zend Engine v4.1.0, Copyright (c) Zend Technologies

[============================================================>] 10000 u | 998905 u/s | Est: 0.0 s | Mem: 530.28 KB | Max: 570.51 KB
[============================================================>] 10000 u | 1420257 u/s | Est: 0.0 s | Mem: 533.62 KB | Max: 570.81 KB
Test name       Repeats         Result          Performance
instanceof      10000           0.002144 sec    +0.00%
is_a()          10000           0.004566 sec    -112.97%
````

**Comment by afbora** [comment]
@distantnative Results vary from moment to moment, but overall it's clear that there is a performance difference of over 75%. `instanceof` is way ahead.

**Review by afbora — DISMISSED** [review]
LGTM 👍

**Review by lukasbestle — COMMENTED** [review]
LGTM

Can we add a rule to our CS fixer config that will automatically convert to `instanceof` in the future so we don't use `is_a()` by accident?

File: src/Form/Options.php [hunk 1]
```diff
@@ -87,7 +87,7 @@ protected static function data($model): array
 
 		// add the model by the proper alias
 		foreach (static::aliases() as $className => $alias) {
-			if (is_a($model, $className) === true) {
+			if ($model instanceof $className) {
```
afbora, line 90: I just really wonder that does `instanceof` also work correctly when  the comparator (`$className`) is string? 🤯

distantnative: I thought first it would not, which is why I left it out. But it seems to actually work. At least as long as the right side is a variable with the class name string. What doesn't work is `$model instance 'className'`.

afbora: Thanks! Now I tried different combinations. It does indeed seem to be working correctly 👍

lukasbestle: Works since PHP 8.0.

lukasbestle (approved):

PR #4610, opened 2022-08-27
Some small psalm fixes

distantnative:


config/helpers.php
```diff
@@ -60,6 +60,7 @@ function collection(string $name)
 	 *
 	 * @param string|null $check Pass a token here to compare it to the one in the session
 	 * @return string|bool Either the token or a boolean check result
+	 * @psalm-return ($check is string ? bool : string)
```
lukasbestle: For `$check === null`, the return value will still be `bool`. Only if the argument is not passed, this will return `string`. Question is how to write that in Psalm syntax.

src/Cms/App.php
```diff
@@ -520,6 +520,7 @@ public function core()
 	 *
 	 * @param string|null $check Pass a token here to compare it to the one in the session
 	 * @return string|bool Either the token or a boolean check result
+	 * @psalm-return ($check is string ? bool : string)
```
lukasbestle: Same here

config/helpers.php
distantnative: Hmm I see, yea no idea. I'll remove it.

lukasbestle (approved):

PR #4611, opened 2022-08-27
Fix error messages for field methods

distantnative:
DRY-ed the error message and added unit tests for `$field->toBlocks()` and `$field->toStructure()`.

Would be great to have this merged for 3.8 as this would improve code coverage for UUID PR.

### DIFF
--- a/config/methods.php
+++ b/config/methods.php
@@ -62,9 +62,22 @@ return function (App $app) {
 		 * @return \Kirby\Cms\Blocks
 		 */
 		'toBlocks' => function (Field $field) {
-			$blocks = Blocks::parse($field->value());
-			$blocks = Blocks::factory($blocks, ['parent' => $field->parent()]);
-			return $blocks->filter('isHidden', false);
+			try {
+				$blocks = Blocks::parse($field->value());
+				$blocks = Blocks::factory($blocks, [
+					'parent' => $field->parent()
+				]);
+				return $blocks->filter('isHidden', false);
+
+			} catch (Throwable) {
+				$message = 'Invalid blocks data for "' . $field->key() . '" field';
+
+				if ($field->parent() !== null) {
+					$message .= ' on parent "' . $field->parent()->title() . '"';
+				}
+
+				throw new InvalidArgumentException($message);
+			}
 		},
 
 		/**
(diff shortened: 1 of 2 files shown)


lukasbestle:
> I looked through the code but couldn't find where an exception would be thrown.

Custom blocks model maybe?

distantnative:
@lukasbestle I don't think so, those would need to throw errors in their constructor for regular data input. Even if, the old error message would not be fitting. I am pretty sure that it just got copied over from `$field->toStructure()` without realising that `Blocks::parse()` wouldn't throw any error.

lukasbestle:
In theory the constructor could throw an exception if the data is somehow malformed.

distantnative:
In what form would data need to be malformed that it remains malformed after `Blocks::parse()` but raises an exception in a custom block model's constructor while being valid for our regular block constructor?

If you have a case, could you please add it as a unit test? Cause that's what I wanted to do in the first place but couldn't come up with one. Without any unit test covering it, I'd rather remove it as suggested.

lukasbestle:
`Blocks::parse()` only decodes JSON, but doesn't check the fields of the block. The block's constructor could validate the fields.

lukasbestle:
I found one case where an exception is thrown:

https://github.com/getkirby/kirby/blob/44158e8a5856384f35b07263240abc96db16ae7e/src/Cms/Block.php#L77

distantnative:
@lukasbestle but the field method creates new block instances only via their `::factory()` method which already ensures that this isn't the case.

lukasbestle:
For me it doesn't. `Block::factory([])` throws that exception for me.

config/methods.php
```diff
@@ -63,16 +63,16 @@
...
 				return $blocks->filter('isHidden', false);
 			} catch (Throwable) {
-				if ($field->parent() === null) {
-					$message = 'Invalid blocks data for "' . $field->key() . '" field';
-				} else {
-					$message = 'Invalid blocks data for "' . $field->key() . '" field on parent "' . $field->parent()->title() . '"';
+				$message = 'Invalid blocks data for "' . $field->key() . '" field';
+
+				if ($field->parent() !== null) {
+					$message .= ' on parent "' . $field->parent()->title() . '"';
```
lukasbestle: For consistency:

```suggestion
				if ($parent = $field->parent()) {
					$message .= ' on parent "' . $parent->title() . '"';
```

distantnative:
Sorry that this one was so complicated for me 🙈🙈🙈

lukasbestle (approved):

PR #4612, opened 2022-08-27
Unique IDs

distantnative:
### Feature
- Unique IDs (UUIDs) for pages, files, users and site https://kirby.nolt.io/16 (blocks and structures to follow in a future release, needs some other foundation work on those field objects)
  - Format: `page://OPJmmoAT1gGocMW`
  - All methods that look up an id/path now also support looking up UUIDs
  - UUIDs are stored as `uuid` field in content files
  - UUIDs are lazily cached in `cache.uuid` (active by default) for more performant lookup
  - Replace ID generator by setting `Uuid\Id::$generator = fn ($length) => ...`
- Picker fields have a new `store` blueprint option (`uuid` or `id` as value, `uuid` default) to configure how to store the value in the content file
- Permanent URLs for pages and files with UUIDs: `site.com/@/page/OPJmmoAT1gGocMW`
  - New `$page->permalink()` and `$file->permalink()` methods

### Breaking changes
- `uuid` cannot be used as field name or custom method for any page, file, user or site anymore
- `permalink` cannot be for custom method names for any page or file and content fields with the same name can only be accessed via `$model->content()->get('permalink')`
- First-level URL path `@` now blocked 
- Panel drag texts now use UUIDs (Kirbytext) or permalinks (Markdown) for absolute references

### Supported models

Model | ID | Cache
-- | -- | --
Site | site:// | not cached (there is only one)
Page | page:// | `$page->id()`
File | file:// | `['parent' => '$scheme://$parent->uuid()', 'filename' => $file→filename()]`
User | user:// | not cached (users already have unique IDs)
Block (future extension) | block:// | `['parent' => '$scheme://$parent→uuid()', 'field' => $fieldName, 'id' => $block→id()]`
Structure Item (future extension) | struct:// | `['parent' => '$scheme://$parent→uuid()', 'field' => $fieldName, 'id' => $structureItem→id()]`

### UUID Uri pattern

```bash
site://
user://user-id
page://HhX1YtRR2ImG6h4
file://HhX1YtRR2ImG6h4
block://HhX1YtRR2ImG6h4
struct://HhX1YtRR2ImG6h4
```

### Persistent storage

UUID’s id string is stored directly in the content files in the `uuid` field so that it becomes part of the content e.g. for version control in a Git repo. The content files are the source of truth.

```
Uuid: HhX1YtRR2ImG6h4

----

Title: ...
```

For blocks and structure entries, a `uuid` field is added to their content as well. For users, the id is in the account folder itself, site does not need an id - it’s already singular.

### Caching
The target models of the IDs are cached in a simple Kirby `uuid` default cache.

```
/cache
  /uuid
    /page    <-- Model type from the scheme
      /Hh
        X1YtRR2ImG6h4.txt
```

The structure with subdirectories for the type and the first two characters of the ID has the following advantages:

- Fast lookup
- Still browsable manually if necessary.
- We won’t end up with a huge file or a super long list of files
- The license server is a good real life example for the stability and performance of thousands of entries

The use of a simple Kirby cache has the advantage that users can use any existing cache driver (e.g. Memcached or Redis). This allows for even better performance. Alternative cache handlers from `Kirby\Cache` can be configured to be used instead. This way we can make the feature itself a thin abstraction layer that doesn’t depend on the actual technical implementation.

### Key methods
- `$model->uuid()` retrieves the $uuid object for a model
- `Uuid::for($model)` does the same
- `Uuid::for('page://HhX1YtRR2ImG6h4')` retrieves the $uuid object for a UUID string
- `$uuid→toString()` returns the UUID string, e.g. `page://HhX1YtRR2ImG6h4`
- `$uuid→model($lazy)` returns the corresponding model object (if lazy only from cache, otherwise also by going through the global index when not found in cache)
- `$uuid→clear()`, `$uuid→populate()`, `$uuid→isCached()` do what you think in regard to caching
- `Uuid::is(string, $type)` checks if a string is a UUID - even if of a specific type
- `Uuid::generate()` returns a new ID string
- `Uuids::populate($type)` allows pre-filling the cache for all UUIDs (or just of a specific type)

### New namespace `Uuid`
- **Uuid (*abstract*)**
  Holds most of the methods for a single UUID, all other UUID classes build on top of this
    - **ModelUuid (*abstract*)**
      Abstraction layer to DRY logic for UUIDs for ModelWithContent objects that store the id string in their content file
        - **PageUuid**
        - **FileUuid**
    - **FieldUuid (*abstract*)**
      Abstraction layer to DRY logic for UUIDs for objects that are generated from a field’s content
        - **BlockUuid**
        - **StructureUuid**
    - **UserUuid**
    - **SiteUuid**
- **HasUuids**
  Trait to add a findByUuid method to all model collections (e.g. pages)
- **Identifiable**
  Interface that only requires a uuid method to be present. Used for typing the various objects that can have a UUID.
- **Uri**
  Extended Http\Uri that handles/parses the UUID  uri scheme/protocol
- **Uuids**
  Holds some static methods that don’t apply to a single UUID

## Ready?

### For review team


config/fields/mixins/picker.php
```diff
@@ -67,12 +67,19 @@
 			return $search;
 		},
 
+		/**
+		 * Whethter to store UUID or path in
+		 * content file
```
lukasbestle:
```suggestion
		 * Whether to store UUID or path in
		 * the content file
		 *
		 * @param string $store 'uuid'|'id'
```

config/routes.php
```diff
@@ -101,6 +102,19 @@
...
+		[
+			'pattern' => '@/(page|file)/(:all)',
+			'method'  => 'ALL',
+			'env'     => 'site',
+			'action'  => function (string $type, string $id) use ($kirby) {
+				// try to resolve to model, but only from UUID cache
+				$model = Uuid::for($type . '://' . $id)->resolve(true);
+				return $kirby
+					->response()
+					->redirect($model?->url() ?? 'error');
```
lukasbestle: I think it would be more consistent with the media routes if we render the error page directly. Untested, but I think this should work:

```suggestion

				if ($model !== null) {
					return $kirby
						->response()
						->redirect($model->url());
				}

				// render the error page
				return false;
```

lukasbestle: Also I noticed we sometimes use `$kirby->response()` in these routes, but sometimes we return a `Response` object directly (e.g. in the media routes). At the moment I don't know which of those is more correct.

config/routes.php
```diff
@@ -101,6 +102,19 @@
 				return Panel::router($path);
 			}
 		],
+		// permalinks for page/file UUIDs
+		[
+			'pattern' => '@/(page|file)/(:all)',
+			'method'  => 'ALL',
```
lukasbestle, line 108: What would be the use case for a non-`GET` request to this route?

config/tags.php
```diff
@@ -170,6 +171,15 @@
 				$tag->value = Url::to($tag->value, $tag->lang);
```
lukasbestle, line 171: Can this call work with UUIDs? Because this is called before the transformation from a UUID URI to a standard URL. I have the feeling we need to move the UUID transformation above this. Instead of `resolve()->url()` we can then use `resolve()->path()` (or `uri()` before the method refactoring). The `Html` class will then take care to convert the path to a full URL even if no language was passed.

src/Cms/Core.php
```diff
@@ -242,6 +244,7 @@ public function fields(): array
 			'toggles'     => $this->root . '/fields/toggles.php',
 			'url'         => $this->root . '/fields/url.php',
 			'users'       => $this->root . '/fields/users.php',
+			'uuid'        => $this->root . '/fields/uuid.php',
```
lukasbestle: I think this might be a leftover from something because the file does not exist.

```suggestion
```

src/Uuid/Uuids.php
```diff
@@ -0,0 +1,76 @@
...
+	public static function cache(): Cache
+	{
+		return App::instance()->cache('uuid');
+	}
+
+	/**
+	 * Populates cache with Uuids for all identifiable models
+	 * that need to be cached (not site and users)
+	 */
+	public static function populate(string $type = 'all'): void
```
lukasbestle, line 33: I think it would be great to name the possible values for the `$type` property like the UUID schemes (so `all`, `page`, `file`, `block`, `struct`). One reason is consistency, but it also makes it easier to connect different custom logic without having to employ a lookup table between scheme and `$type` value.

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,318 @@
...
+use Kirby\Cms\Collection;
+use Kirby\Cms\File;
+use Kirby\Cms\Page;
+use Kirby\Cms\Site;
+use Kirby\Cms\User;
+use Kirby\Toolkit\Str;
+
+/**
+ * The `Uuid` classes provide an interface to connect
+ * indetifiable models (page, file, site, user, blocks,
```
lukasbestle:
```suggestion
 * identifiable models (page, file, site, user, blocks,
```

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,318 @@
...
+{
+	protected const TYPE = 'uuid';
+
+	/**
+	 * Customisable callback function for generating new ID strings
+	 * instead of using Nanoid. Receives length of string as parameter.
+	 */
+	public static Closure|null $generator = null;
+
+	public Collection|null $context;
```
lukasbestle, line 55: Suggestion for a docblock to make clearer what the context is for and what it does:

```suggestion
	/**
	 * Collection that is likely to contain the model and
	 * that will be checked first to speed up the lookup
	 */
	public Collection|null $context;

```

I hope I understood it correctly.

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,318 @@
...
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ */
+class Uuid
+{
+	protected const TYPE = 'uuid';
+
+	/**
+	 * Customisable callback function for generating new ID strings
+	 * instead of using Nanoid. Receives length of string as parameter.
```
lukasbestle:
```suggestion
	 * instead of using Nano ID. Receives length of string as parameter.
```

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,318 @@
...
+		$this->model   = $model;
+
+		if ($model) {
+			$this->uri = new Uri([
+				'scheme' => static::TYPE,
+				'host'   => static::retrieveId($model)
+			]);
+		} elseif ($uuid) {
+			$this->uri = new Uri($uuid);
+		}
```
lukasbestle: In theory both `$uuid` and `$model->uuid()` should be the same and also in theory one should never construct a `Uuid` object with both a UUID string and a model object, but what happens if both are passed and there is a mismatch? Maybe we should throw an exception if both are passed and don't match.

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	/**
+	 * Removes Uuid from cache,
+	 * recursively if needed
+	 */
+	public function clear(bool $recursive = false): bool
+	{
+		// For all models with children: if $recursive,
+		// also clear UUIDs rom cache for all children
```
lukasbestle:
```suggestion
		// also clear UUIDs from cache for all children
```

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,318 @@
...
+	/**
+	 * Look up Uuid in cache and resolve
+	 * to identifiable model object.
+	 * Implemented on child classes.
+	 *
+	 * @codeCoverageIgnore
+	 */
+	protected function findByCache(): Identifiable|null
+	{
+		return null;
```
lukasbestle: I think we should throw a `LogicException` in these methods that are supposed to be overridden. This ensures that we can never miss to implement them. Abstract methods would of course be even better, but I assume there is a reason why we can't use them here.

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,318 @@
...
+		if (is_string($seed) === true) {
+			return match (Str::before($seed, '://')) {
+				'page'   => new PageUuid(uuid: $seed, context: $context),
+				'file'   => new FileUuid(uuid: $seed, context: $context),
+				'site'   => new SiteUuid(uuid: $seed, context: $context),
+				'user'   => new UserUuid(uuid: $seed, context: $context),
+				// TODO: activate for uuid-block-structure-support
+				// 'block'  => new BlockUuid(uuid: $seed, context: $context),
+				// 'struct' => new StructureUuid(uuid: $seed, context: $context),
+				default  => throw new InvalidArgumentException('Invalid uuid uri:' . $seed)
```
lukasbestle:
```suggestion
				default  => throw new InvalidArgumentException('Invalid UUID URI: ' . $seed)
```

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,318 @@
...
+				=> new SiteUuid(model: $seed, context: $context),
+			$seed instanceof User
+				=> new UserUuid(model: $seed, context: $context),
+			// TODO: activate for uuid-block-structure-support
+			// $seed instanceof Block
+			// 	=> new BlockUuid(model: $seed, context: $context),
+			// $seed instanceof StructureObject
+			// 	=> new StructureUuid(model: $seed, context: $context),
+			default
+			=> throw new InvalidArgumentException('Uuid not supported for:' . get_class($seed))
```
lukasbestle:
```suggestion
			=> throw new InvalidArgumentException('UUID not supported for: ' . get_class($seed))
```

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,318 @@
...
+	/**
+	 * Returns key for cache entry
+	 */
+	public function key(): string
+	{
+		$id = $this->id();
+
+		// for better performance when using a file-based cache,
+		// turn first two characters of the id into a directory
+		$id = substr($id, 0, 2) . '/' . substr($id, 2);
```
lukasbestle: Should use `Str` methods to be multi-byte safe in case a user uses custom IDs with Unicode characters.

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,318 @@
...
+	 *
+	 * @return bool
+	 */
+	public function populate(): bool
+	{
+		return Uuids::cache()->set($this->key(), $this->value());
+	}
+
+	/**
+	 * Returns the full UUID string incl. schema
```
lukasbestle:
```suggestion
	 * Returns the full UUID string incl. scheme
```

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,318 @@
...
+		// make sure id is generated if
+		// it doesn't exist yet
+		$this->id();
+
+		return $this->uri->toString();
+	}
+
+	/**
+	 * Tries to find the idetifiable model in cache
+	 * or index and return the object
```
lukasbestle:
```suggestion
	 * Tries to find the identifiable model in cache
	 * or index and returns the object
	 *
	 * @param bool $lazy If `true`, only lookup from cache
```

lukasbestle (commented):
First set of comments

config/routes.php
distantnative: Couldn't there be various use cases to send a POST request to a page (form submission etc)? Admittedly, less likely that the permalink would be used for it but should we prevent it?

config/routes.php
distantnative: Me neither.

config/tags.php
distantnative: You are right, this won't work. If we redo it as you suggest however, we don't end up with URLs pointing to the file in media 


```
-'<a href="https://getkirby.com/media/pages/a/72f9c7035c-0/foo.jpg">file</a>'
+'<a href="https://getkirby.com/a/foo.jpg">file</a>'
```

src/Uuid/Uuid.php
distantnative: The issue with abstract methods is that all child classes need to implement them - while e.g. `UserUuid` and `SiteUuid` have no use for it. LogicException sounds like a good way then, will try it.

config/routes.php
distantnative: I think using `->response()->redirect()` here makes sense. We can revisit all routes in general at another point if we want to make it more uniform.

config/tags.php
distantnative: Was actually looking at the wrong unit test.

distantnative: It does seem to work as is. Added a unit test: https://github.com/getkirby/kirby/pull/4612/commits/daadb0a9308c1818e24c71285f9f71cd2dcd6b0b

src/Uuid/ModelUuid.php
```diff
@@ -0,0 +1,104 @@
...
+
+		// generate ID and write to content file
+		$id = static::generate();
+
+		// make sure Kirby has the required permissions
+		// for the update action
+		$kirby = App::instance();
+		$user  = $kirby->auth()->currentUserFromImpersonation();
+		$kirby->impersonate('kirby');
+		$this->model = $this->model->update(['uuid' => $id]);
```
bastianallgeier: This line is causing an endless loop and I'm not entirely sure why yet. It is solved by just using the underlying save method instead but we need to double-check if that has any side effects.

```suggestion
		$this->model = $this->model->save(['uuid' => $id]);
```

bastianallgeier (dismissed):
After the last few glitches I found, it seems super solid to me. I think we need to test it with bigger installations in the RC phase, but I would merge it.

lukasbestle (changes requested):
Currently in the process of reviewing the remaining code changes. My plan is to send the review today.

src/Cache/FileCache.php
```diff
@@ -90,7 +90,7 @@ protected function file(string $key): string
 					// check if the segment only contains safe characters;
 					// underscores are *not* safe to guarantee uniqueness
```
distantnative: @bastianallgeier seems like your change contradicts this comment

bastianallgeier: Shit. That happens when you don't read the comments. @lukasbestle do you remember why they are not safe? I don't really understand the comment to be honest.

lukasbestle: It's literally in the comment: 😅

> as they are used in the special cases

If the cache ID ever contains `/_empty` for example, the cache will be broken.

If we want to ensure that the UUID is compatible with the cache without having to use hashes, we can either use a custom Nano ID alphabet without the underscore or we can rewrite the cache implementation to use a different special character like a `*`. I'd prefer the first solution as it avoids the breaking change.

bastianallgeier: Ah, that makes sense. I didn't understand what's meant with "special cases". Do we need to modify the Nano ID implementation to support custom alphabets? Otherwise that's indeed the nicer solution.

lukasbestle: What I don't really get is what the difference between `Hidehalo\Nanoid\Client::MODE_NORMAL` and `MODE_DYNAMIC` is supposed to be. The default `MODE_NORMAL` skips basically all the generator code and just uses `mt_*` functions in `Client::normalRandom()`. The `mt_*` functions are not as random as `random_bytes()` and also at this point the whole library is useless as we could just use a similar logic to `Client::normalRandom()` directly in our own code.

If we find out that we want to use `MODE_DYNAMIC`, then we can use something like this:

```php
$core      = new Hidehalo\Nanoid\Core();
$generator = new Hidehalo\Nanoid\Generator();
$alphabet  = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';

$uuid = $core->random($generator, $length, $alphabet);
```

But at the moment I'm a bit confused about the Nanoid library.

lukasbestle: I'm wondering if we shouldn't just use `Str::random($length, 'alphaNum')` instead of the Nanoid library. Then we can avoid the extra library. And the two fewer characters (`-` and `_`) aren't going to the significantly increase the collision probability.

distantnative: I think we wanted to stick to a standard format here. I think therefore, it makes more sense to use the "official" (at least listed) PHP library for Nanoid. Whether to use normal or dynamic I also don't fully understand, especially as there eis no equivalent in the JS version. But I would stick to the documented methods, so 

```php
$client->formattedId('0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ', $length);
```

distantnative: `Str::random()` could be an option too... I mean in the beginning we wanted to go with UUIDv4 as standard. But then diverted to Nanoid because it would be less of a monster as string. But in the end, `Str::random()` could do the job as well.

distantnative: And for people that really want a specific format, `Uuid` already provides a way to customise the generator.

lukasbestle: Well, I feel like Nanoid is really just a fancy name for "unique case-sensitive ID without a fixed structure". The actual generator code doesn't do any magic, it is probably just a bit more efficient than our current `Str::random()` implementation. But we can also optimize our own implementation without having to use a third-party dependency.

distantnative: I'd be fine with this 👍

distantnative: Removed the nano library and reverted Basti's change to FileCache regex.

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,334 @@
+<?php
+
+namespace Kirby\Uuid;
+
+use Closure;
+use Generator;
+use InvalidArgumentException;
```
lukasbestle: Why the PHP exception in particular and not `Kirby\Exception\InvalidArgumentException`?

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,334 @@
...
+	 * Removes Uuid from cache,
+	 * recursively if needed
+	 */
+	public function clear(bool $recursive = false): bool
+	{
+		// For all models with children: if $recursive,
+		// also clear UUIDs from cache for all children
+		if (
+			$recursive === true &&
+			$this->model &&
```
lukasbestle: I have the feeling this method should call `resolve()` if `$recursive` is `true` to ensure consistent behavior no matter if the model has already been resolved or not.

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,334 @@
...
+			// 	=> new StructureUuid(model: $seed, context: $context),
+			default
+			=> throw new InvalidArgumentException('UUID not supported for: ' . get_class($seed))
+		};
+	}
+
+	/**
+	 * Generate a new ID string
+	 */
+	final public static function generate(int $length = 15): string
```
lukasbestle: I suggest to use a default length of 16 because 128 bits are a common field length for ID fields (could be relevant for custom cache drivers).

```suggestion
	final public static function generate(int $length = 16): string
```

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,334 @@
...
+
+		return null;
+	}
+
+	/**
+	 * Retrieves the existing ID string for the model
+	 */
+	public static function retrieveId(Identifiable $model): string|null
+	{
+		return $model->id();
```
lukasbestle, line 310: I know this is only a fallback, but wouldn't it be incorrect? As `$model->id()` basically returns the value instead of the "UUID without scheme".

src/Uuid/ModelUuid.php
```diff
@@ -0,0 +1,104 @@
...
+
+		// update the Uri object
+		$this->uri->host($id);
+
+		return $id;
+	}
+
+	/**
+	 * Retrieves the existing ID string for the model
+	 * or generates a new one, if required
```
lukasbestle: I don't think UUID generation happens when this method is called, does it?

src/Uuid/FileUuid.php
```diff
@@ -0,0 +1,89 @@
...
+			return null;
+		}
+
+		// value is itself another Uuid\Uri string
+		// e.g. page://page-uuid/filename.jpg
+		$uuid = new Uri($value);
+
+		// we need to resolve the parent UUID to its model
+		// and then query for the file by filename
+		$parent   = Uuid::for($uuid->base())->resolve();
```
lukasbestle: Just realized that the `page://page-uuid/filename.jpg` format for file UUIDs means that UUIDs must never contain slashes, which is relevant for manual UUIDs and custom generators. We need to document this.

Alternatively we could change the cache storage to use e.g. a JSON string of `['parentUuid' => '...', 'filename' => '...']`. Would that be viable?

(Same with the field UUIDs.)

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,334 @@
...
+		return $this->uri->toString();
+	}
+
+	/**
+	 * Tries to find the identifiable model in cache
+	 * or index and returns the object
+	 *
+	 * @param bool $lazy If `true`, only lookup from cache
+	 */
+	public function resolve(bool $lazy = false): Identifiable|null
```
lukasbestle: Can we rename this method `model()`? The `Uuid` class has many methods and while reviewing the code I often had to look up what they do. `model()` would IMO be clearer.

src/Uuid/HasUuids.php
```diff
@@ -0,0 +1,31 @@
...
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ */
+trait HasUuids
+{
+	/**
+	 * Find a single element by global UUID
+	 */
+	public function findByUuid(
```
lukasbestle: I feel like this method should be `protected` given that we just deprecated the various `findBy*` methods in the CMS collections.

(Note: `findByKey()` is only `public @internal` because it is inherited from `Toolkit\Collection`.)

src/Cms/FileActions.php
```diff
@@ -305,7 +311,15 @@ public function replace(string $source)
 	 */
 	public function unpublish()
```
lukasbestle: I think in the `$file->replace()` case this extended functionality will lead to incorrect results. In this case the lock doesn't need to be deleted and the UUID cache doesn't need to be cleared.

(Note: If we decide that we do want to keep these actions in `unpublish()`, we can remove the extra lock file deletion in the `delete()` method.)

src/Panel/File.php
```diff
@@ -299,8 +302,7 @@ public function pickerData(array $params = []): array
 			'filename' => $name,
 			'dragText' => $this->dragText('auto', $absolute ?? false),
 			'type'     => $this->model->type(),
-			'url'      => $this->model->url(),
-			'uuid'     => $uuid ?? $id,
```
lukasbestle, line 303: Can you explain what this did before and why it can be removed? Don't fully understand the logic.

src/Uuid/Uuid.php
distantnative: A lot of the upcoming blueprint refactoring by @bastianallgeier is structured along resolve/render which is why I thought it could be nice to stick to that terminology here as well.

src/Panel/File.php
distantnative: Me neither - @bastianallgeier needs to answer this, I couldn't find out any meaning from reading the code and it worked without it during my testing. That's why I considered it dead code.

src/Uuid/FileUuid.php
distantnative: I'd be fine with either way - if the first we should document it as well as check for in the generate method and throw an exception if it contains slashes.

lukasbestle: I think if nothing speaks against using JSON, I'd prefer that. It would be much more future-proof and robust. An exception can only catch literally random cases that happen to contain a slash.

src/Uuid/Uuid.php
lukasbestle: OK, I see. Leaving this discussion open for a second opinion from the team. Anyone can feel free to make it as resolved.

src/Uuid/ModelUuid.php
```diff
@@ -0,0 +1,104 @@
...
+	 */
+	public static function retrieveId(Identifiable $model): string|null
+	{
+		return $model->content()->get('uuid')->value();
+	}
+
+	/**
+	 * Returns permalink url
+	 */
+	public function url(): string
```
lukasbestle, line 99: This method should populate the cache to ensure that the resulting permalink can be reached.

Open question: What would the performance impact be to run populate on every request that needs the permalink? Can we maybe optimize it somehow to check if the cache is already populated?

src/Uuid/Uuid.php
distantnative: `retrieveId()` is used for the uri host - for all models that don't store their UUID ID string in a content file this is likely to be `->id()`, e.g. for users. So for me it makes sense to define this as the baseline and then `ModelUuid` redefined `retrieveId()` to look it up in the content file instead.

src/Uuid/ModelUuid.php
distantnative: So far I didn't check if the cache is already populated when populating as I thought it would be much less performant to always first read/check the cache instead of just writing to it.

src/Uuid/FileUuid.php
distantnative: Should we generate the JSON directly or rathe pass an array and let the cache river handle how to store it, e.g. as JSON?

distantnative: I've implemented it as array for now

src/Cms/FileActions.php
distantnative: I've solved it by adding an `$onlyMedia` flag to `unpublish()`.

src/Uuid/ModelUuid.php
lukasbestle: Makes sense. For this case I feel like we should still read the cache first. Writing on every request is going to result in a lot of wear for SSDs.

src/Panel/File.php
bastianallgeier: There's a dark memory that this was only added for some potential plugins that would need the URL. Don't remember at all why we added this weird uuid thing.

src/Uuid/Uuid.php
bastianallgeier: Both names work for me tbh.

lukasbestle: What I like about `model()` is that it matches the prop name, like we do it in other places.

src/Panel/File.php
distantnative: We still have a `uuid` entry (now defined in Panel\Model and now containing the new UUID. But maybe that's sufficient for potentially existing plugins?

src/Uuid/Uuid.php
distantnative: Add model() as alias?
I don't think render would make much sense on its own.

lukasbestle: An alias would make it even more confusing TBH. What if we rename render to toString?

distantnative: Not sure, would be a good chance to use the same terminology here that we will use so much in the future. Probably @bastianallgeier needs to make a decision even if he likes both 😄

src/Panel/File.php
bastianallgeier: I would remove it. I'm really not sure if it has ever been used. We have new ideas for the pickers anyway and if a plugin has a problem with the removal, we can fix it easily.

src/Uuid/Uuid.php
bastianallgeier: I think the new stuff is still very much under construction. There are a few things that already seem very useful, but I'm not 100% sure about how "final" render and resolve really are. Maybe that's already a good indicator to actually stick with model instead.

src/Uuid/ModelUuid.php
distantnative: Added it

src/Uuid/Uuid.php
distantnative: Giving in to the majority - made the changes.

lukasbestle (approved):
Awesome, looks great! 🎉

I've added a few minor fixes. @distantnative Could you please double-check if they are correct?
Also updated the first post with the changes to method names and stored cache values.

distantnative:
Looks good to me - thanks!

Issue #4613, opened 2022-08-27
Periodic API ping breaks login view


lukasbestle:
## Description

The Panel periodically pings the API to keep the session fresh:

https://github.com/getkirby/kirby/blob/44158e8a5856384f35b07263240abc96db16ae7e/panel/src/config/api.js#L61

If no user is logged in, this request fails and triggers a logout:

https://github.com/getkirby/kirby/blob/44158e8a5856384f35b07263240abc96db16ae7e/panel/src/config/api.js#L22-L29

This causes a UI reset if the login page is left open for 5 minutes. This is bad e.g. during a code challenge (normal timeout for that is 10 minutes) and also when the user leaves the login page open to lookup a password etc.

**Expected behavior**  

The redirection to logout shouldn't happen.

## To reproduce

1. Open login page
2. Enter some text but don't submit (or request a code challenge and leave that open)
3. Wait for 5 minutes

## Your setup
**Kirby Version**  

`develop`

**Additional context**  

We need to check if the session actually needs refreshing before login. Might be that we only set the session inactivity timeout on login.

If the session *does* need refreshing, the ping should occur but the Panel should not redirect to logout when no user is logged in.
If it *doesn't* need refreshing, the ping should be disabled completely during login.

lukasbestle:
OK, so it seems like our backend code doesn't differentiate between sessions with a logged in user and sessions without user. Both have a timeout of half an hour by default (can be overridden or disabled with the `session.timeout` option).

So we do need the regular ping, but the `$go('logout')` shouldn't be called if there isn't an active user in the frontend store. This is the easy part.

More difficult: The ping response also triggers a dialog:

<img width="516" alt="Bildschirmfoto 2022-08-27 um 21 10 53" src="https://user-images.githubusercontent.com/1595007/187044775-1d5f87d2-518c-4ea5-a686-bdef9f34f0a1.png">

Where is this triggered and how can we suppress it?

distantnative:
Where do you get the dialog? Is that only when you remove `$go('logout')`?

lukasbestle:
Yes. With the current code that dialog flashes for a tiny moment before the view is reloaded. If I disable `$go('logout')`, the dialog stays.

distantnative:
Ok so this happens because our API library while passing errors to the onError callback (where our app puts its logic then) still throws all errors: https://github.com/getkirby/kirby/blob/main/panel/src/api/request.js#L84-L85

lukasbestle:
[...]

PR #4614, opened 2022-08-27
Fix API ping

distantnative:
## This PR …

- `api.onError()`: allows returning `false` to suppress throwing error
- suppresses throwing unauthenticated error when redirecting to logout
- new `auth/api` endpoint
- used for regular pinging
- just one continuous interval used for pinging, no need for resetting on regular API calls anymore

### Fixes
- #4613

## Ready?

### For review team

### DIFF
--- a/config/api/routes/auth.php
+++ b/config/api/routes/auth.php
@@ -110,7 +110,8 @@ return [
 		'method'  => 'POST',
 		'auth'    => false,
 		'action'  => function () {
-			$this->kirby()->auth()->user();
+			// refresh the session timeout
+			$this->kirby()->session();
 			return true;
 		}
 	],
--- a/config/api/routes/auth.php
+++ b/config/api/routes/auth.php
@@ -105,4 +105,13 @@ return [
 			return true;
 		}
 	],
+	[
+		'pattern' => 'auth/ping',
+		'method'  => 'POST',
+		'auth'    => false,
+		'action'  => function () {
+			$this->kirby()->auth()->user();
+			return true;
+		}
+	],
 ];
--- a/panel/src/api/request.js
+++ b/panel/src/api/request.js
@@ -83,7 +83,7 @@ export default (config) => {
 				config.onComplete(id);
 
 				// pass to error callback and throw error
-				// except when callback explcitly suppresses it
+				// except when callback explicitly suppresses it
 				if (config.onError(e) !== false) {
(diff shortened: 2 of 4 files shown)


config/api/routes/auth.php
```diff
@@ -105,4 +105,13 @@
 			return true;
 		}
 	],
+	[
+		'pattern' => 'auth/ping',
+		'method'  => 'POST',
+		'auth'    => false,
+		'action'  => function () {
+			$this->kirby()->auth()->user();
```
lukasbestle:
```suggestion
			// refresh the session timeout
			$this->kirby()->session();
```

panel/src/api/request.js
```diff
@@ -81,8 +81,12 @@ export default (config) => {
 			} catch (e) {
 				this.running--;
 				config.onComplete(id);
-				config.onError(e);
-				throw e;
+
+				// pass to error callback and throw error
+				// except when callback explcitly suppresses it
```
lukasbestle:
```suggestion
				// except when callback explicitly suppresses it
```

panel/src/config/api.js
```diff
@@ -48,16 +49,13 @@ export default {
 					}
 
 					Vue.$api.requests.push(requestId);
-				},
-				onSuccess: () => {
-					clearInterval(Vue.$api.ping);
-					Vue.$api.ping = setInterval(Vue.$api.auth.user, 5 * 60 * 1000);
```
lukasbestle: We had this code so that the ping can be skipped while the user is active. So the ping would only be sent if there was no activity in the last five minutes. I think this still makes sense.

panel/src/config/api.js
```diff
@@ -48,16 +49,13 @@ export default {
...
 				}
 			},
-			ping: null,
 			requests: []
 		});
 
-		Vue.$api.ping = setInterval(Vue.$api.auth.user, 5 * 60 * 1000);
+		// regularly ping API to keep session alive
+		// (if any use is authenticated)
```
lukasbestle:
```suggestion
		// (if any user is authenticated)
```

lukasbestle: Actually I think we can remove this line entirely. The session needs to be kept alive even if no user is authenticated.

lukasbestle (approved):

PR #4615, opened 2022-08-29
Switches to GitHub stale action

afbora:
Actually Probot Stale issue was [deprecated](https://github.com/probot/probot.github.io/pull/376) but I didn't aware of this when I adding the probot stale bot. Related source of issue: https://github.com/probot/stale/issues/384

Now we have to switch to GitHub action. But I need help you to check out action configuration (especially about permissions and other missing options). Here the action and all options: https://github.com/actions/stale


lukasbestle (approved):

lukasbestle:
I've uninstalled the old app now to avoid that the two integrations start fighting. ;)

PR #4622, opened 2022-09-03
Parent model for filesystem

### LINKED ISSUE #4274: file::url component throws error in 3.6.4
## Description
When adding the file::url component as described in the ["Virtual pages from images" recipe](https://getkirby.com/docs/cookbook/templating/virtual-pages-image-gallery), the Panel throws a "The JSON response could not be parsed" error .
Additionally, I get a 502 Bad Gateway error.

The example works perfectly in 3.5.8.

## To reproduce
Add the plugin as described in the recipe to a fresh 3.6.4 Starterkit (probably the same in all 3.6.x versions, not tested yet).

## Your setup
**Kirby Version**  
3.6.4
**Console output**  

MacOs 12.1 with Firefox 99
Laravel Valet as dev environment

Probably related to the same changes that lead to this issue: https://github.com/getkirby/kirby/issues/4143

afbora:
## This PR …

When a method of the filesystem object was called in the `file::url` component, there was an infinite loop and the system was throwing an error. Now, instead of sending the `url` and `root` attributes when instancing the filesystem object, the issue is solved by sending its parent file object.

### Fixes
- Calling a `$file` method inside a custom `file::url` component no longer causes an infinite loop #4274

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -333,6 +333,14 @@ class File
 	}
 
 	/**
+	 * Returns the parent file model
+	 */
+	public function model(): object|null
+	{
+		return $this->model;
+	}
+
+	/**
 	 * Returns the file's last modification time
 	 *
 	 * @param string|null $handler date, intl or strftime
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -31,7 +31,7 @@ class File
 
 	/**
 	 * Parent file model
-	 * The model object must use the "Kirby\Filesystem\IsFile" trait
+	 * The model object must use the `\Kirby\Filesystem\IsFile` trait
 	 */
 	protected object|null $model = null;
 
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -31,8 +31,9 @@ class File
 
 	/**
 	 * Parent file model
+	 * The model object must use the "Kirby\Filesystem\IsFile" trait
 	 *
-	 * @var \Kirby\Filesystem\IsFile|null
+	 * @var object|null
 	 */
 	protected $model;
 
@@ -431,7 +432,7 @@ class File
 	/**
 	 * Setter for the parent file model
 	 *
-	 * @param \Kirby\Filesystem\IsFile|null $model
+	 * @param object|null $model
 	 * @return $this
 	 *
 	 * @throws \Kirby\Exception\InvalidArgumentException
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -426,11 +426,7 @@ class File
 	 */
 	public function root(): string|null
 	{
-		if ($this->root !== null) {
-			return $this->root;
-		}
-
-		return $this->root = $this->model?->root();
+		return $this->root ??= $this->model?->root();
 	}
 
 	/**
@@ -478,18 +474,10 @@ class File
 	 */
 	public function url(): string|null
 	{
-		if ($this->url !== null) {
-			return $this->url;
-		}
-
 		// lazily determine the URL from the model object
 		// only if it's needed to avoid breaking custom file::url
 		// components that rely on `$cmsFile->asset()` methods
-		if ($this->model !== null) {
-			return $this->url = $this->model->url();
-		}
-
-		return null;
+		return $this->url ??= $this->model?->url();
 	}
 
 	/**
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -4,6 +4,7 @@ namespace Kirby\Filesystem;
 
 use IntlDateFormatter;
 use Kirby\Cms\App;
+use Kirby\Cms\File as CmsFile;
 use Kirby\Exception\Exception;
 use Kirby\Http\Response;
 use Kirby\Sane\Sane;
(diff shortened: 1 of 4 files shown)


src/Filesystem/File.php
```diff
@@ -447,7 +474,18 @@ protected function setUrl(string|null $url = null): static
...
+		}
+
+		// lazily determine the URL from the model object
+		// only if it's needed to avoid breaking custom file::url
+		// components that rely on `$cmsFile->asset()` methods
+		if ($this->model !== null) {
+			return $this->url = $this->model->url();
+		}
+
+		return null;
```
distantnative: Lines 481-488 could be simply

```php
return $this->url = $this->model?->url();
```

If model is null the whole return statement will still be null.

distantnative:
Looking at this PR, wondering if `Filesystem\File` should treat `root` the same way as we now handle `url`?

afbora:
Yes, that makes sense. We can do it for `root` as we did for the `url`. It would be something like this. I got it right, right?

````php
/**
 * Returns the given file path
 */
public function root(): string|null
{
  if ($this->root !== null) {
    return $this->root;
  }

  return $this->root = $this->model?->root();
}
````

**Comment by distantnative** [comment]
I think so yes!

**Comment by afbora** [comment]
`root` handled and simplified ✅

**Comment by distantnative** [comment]
Sadly we cannot see the codecov diff right now. I think it's because the last commit on `develop` didn't run code coverage check. Would be good to add a unit test for the initial issue why we tackled this.

File: src/Filesystem/IsFile.php [hunk 2]
```diff
@@ -84,10 +84,13 @@ public function asset(array|string|null $props = null): File
...
 
-		$props = $props ?? [
-			'root' => $this->root(),
-			'url'  => $this->url()
-		];
+		if (is_array($props) === false) {
+			$props = [
+				'root' => $props ?? $this->root()
+			];
+		}
```
distantnative, line 91: If we already add the model, maybe this all should rather read now

```php
$props ??= ['model' => $this];
```

afbora: But if `$props` is not `null`, the `model` won't added. The `model` should always be added.

afbora: Ahh, I get it now what you mean 👍

afbora: I couldn't simplify any further. The `string` $props (to use as root) parameter is now useless. Should we deprecate?

````php
public function asset(array|string|null $props = null): File
{
	if ($this->asset !== null) {
		return $this->asset;
	}

	$props ??= [];

	if (is_string($props) === true) {
		$props = ['root' => $props];
	}

	$props['model'] ??= $this;

	return $this->asset = match ($this->type()) {
		'image' => new Image($props),
		default => new File($props)
	};
}
````

Comment (distantnative) [reply src/Filesystem/IsFile.php:91]: > The string $props (to use as root) parameter is now useless. 

Why? You still have `is_string($props) === true`. I think this looks good.

**Comment by distantnative** [comment]
Looks good! Just needs testing

**Comment by afbora** [comment]
I'm pretty sure that all changed lines covered (checked with debug on IDE) but codecov says  95.27% 🤷‍♂️

File: tests/Filesystem/FileTest.php [hunk 3]
```diff
@@ -396,6 +405,41 @@ public function testMime()
...
 	}
 
+	/**
+	 * @covers ::model
+	 * @covers ::setModel
+	 */
+	public function testModel()
+	{
+		$parent = \Kirby\Cms\Page::factory(['slug' => 'test']);
+		$model = \Kirby\Cms\File::factory([
```
lukasbestle: Could you move these imports for `Page` and `File` to `use` imports at the top of the file?

```php
use Kirby\Cms\File as CmsFile;
use Kirby\Cms\Page;
```

Same for the `testInvalidModel()` method.

src/Filesystem/File.php
```diff
@@ -28,6 +29,12 @@ class File
 {
 	use Properties;
 
+	/**
+	 * Parent file model
+	 * The model object must use the "Kirby\Filesystem\IsFile" trait
```
```suggestion
	 * The model object must use the `Kirby\Filesystem\IsFile` trait
```

lukasbestle (commented):
Could you also add tests for the automatic `url` and `root` from the model object?

afbora:
All done, ready to review 🚀

src/Filesystem/File.php
```diff
@@ -322,7 +329,15 @@ public function match(array $rules): bool
 	 */
 	public function mime(): string|null
 	{
-		return Mime::type($this->root);
+		return Mime::type($this->root());
+	}
+
+	/**
+	 * Returns the parent file model
```
distantnative:
```suggestion
	 * Returns the parent file model, which uses this instance as proxied file asset
```

src/Filesystem/File.php
```diff
@@ -417,7 +432,24 @@ public function rename(string $newName, bool $overwrite = false): static
 	 */
 	public function root(): string|null
 	{
-		return $this->root;
+		return $this->root ??= $this->model?->root();
+	}
+
+	/**
+	 * Setter for the parent file model
```
distantnative:
```suggestion
	 * Setter for the parent file model, which uses this instance as proxied file asset
```

src/Filesystem/File.php
```diff
@@ -417,7 +432,24 @@ public function rename(string $newName, bool $overwrite = false): static
...
+	 *
+	 * @return $this
+	 *
+	 * @throws \Kirby\Exception\InvalidArgumentException
```
distantnative:
```suggestion
	 * @throws \Kirby\Exception\InvalidArgumentException When the model does not use the `Kirby\Filesystem\IsFile` trait
```

lukasbestle (approved):

PR #4627, opened 2022-09-04
New `Cache::getOrSet()` method

distantnative:
### Feature
- New `Cache::getOrSet($key, $callback, $minutes)` method: Retrieves the value from cache if possible, if not via callback function and adds it to the cache

Makes for much compacter code:
```php
$value = $cache->getOrSet('my.key', function () { ... });
```
instead of
```php
$value = $cache->get('my.key');

if ($value !== null) {
    return $value;
}

// logic to retrieve value
$value = ....

$cache->set('my.key', $value);
return $value;
```

## Ready?

### For review team

### DIFF
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -156,16 +156,17 @@ abstract class Cache
 	 */
 	public function getOrSet(
 		string $key,
-		Closure $getter,
+		Closure $result,
 		int $minutes = 0
 	): mixed {
-		if ($value = $this->get($key)) {
-			return $value;
+		$value  = $this->get($key);
+		$result = $value ?? $result();
+
+		if ($value === null) {
+			$this->set($key, $result, $minutes);
 		}
 
-		$value = $getter();
-		$this->set($key, $value, $minutes);
-		return $value;
+		return $result;
 	}
 
 	/**
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -152,6 +152,7 @@ abstract class Cache
 	 * Returns a value by either getting it from the cache
 	 * or via the callback function which then it stored in
 	 * the cache for future retrieval
+	 * @since 3.8.0
 	 */
 	public function getOrSet(
 		string $key,
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -150,7 +150,7 @@ abstract class Cache
 
 	/**
 	 * Returns a value by either getting it from the cache
-	 * or via the callback function which then it stored in
+	 * or via the callback function which then is stored in
 	 * the cache for future retrieval. This method cannot be
 	 * used for `null` as value to be cached.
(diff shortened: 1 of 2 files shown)


src/Cache/Cache.php
```diff
@@ -143,22 +111,72 @@ public function expired(string $key): bool
...
+	 * the cache for future retrieval
+	 */
+	public function getOrSet(
+		string $key,
+		Closure $getter,
+		int $minutes = 0
+	): mixed {
+		if ($value = $this->get($key)) {
+			return $value;
+		}
```
afbora: Can the cache value be `false`? Technically it seems possible, but if there is a real use, we need to change this line.

src/Cache/Cache.php
```diff
@@ -143,22 +111,72 @@ public function expired(string $key): bool
...
+
+		// return the pure value
+		return $value->value();
+	}
+
+	/**
+	 * Returns a value by either getting it from the cache
+	 * or via the callback function which then it stored in
+	 * the cache for future retrieval
```
afbora:
```suggestion
	 * the cache for future retrieval
	 * @since 3.8.0
```

src/Cache/Cache.php
```diff
@@ -143,22 +111,74 @@ public function expired(string $key): bool
...
+	public function getOrSet(
+		string $key,
+		Closure $result,
+		int $minutes = 0
+	): mixed {
+		$value  = $this->get($key);
+		$result = $value ?? $result();
+
+		if ($value === null) {
+			$this->set($key, $result, $minutes);
```
lukasbestle, line 167: If a user explicitly stores `null` in the cache (for whatever reason), this code will treat the value like it wasn't stored. For consistency and reliability, I think we should differentiate between these cases.

What we could do is to move the expiry check from `get()` into a new protected method like `retrieveValid(string $key): Value|null`. The `get()` method can then just do:

```php
$value = $this->retrieveValid($key);

if ($value) {
	return $value->value();
}

return $default;
```

(Note: something like `$this->retrieveValid($key)?->value() ?? $default` will again *not* work for the same reason.)

The logic in `getOrSet()` can then be changed to:

```php
$value = $this->retrieveValid($key);

if ($value) {
	return $value->value();
}

$result = $result();
$this->set($key, $result, $minutes);

return $result;
```

distantnative: I feel like `retrieveValid` would over engineer it a bit - and I'd find it hard to understand the difference as user. We use the pattern of `getOrSet` in a few places already - in situations where the cached value cannot be `null`. I think it would be a lot easier to just add a comment to the docblock that this cannot be used on a `null` cache value - which would be an extreme edge case anyways.

lukasbestle: Fine for me. :)

lukasbestle (approved):

PR #4631, opened 2022-09-05
Fix table layout heading alignment

### LINKED ISSUE #4630: table layout: alignment does not affect column heading
When aligning right in the pagetable extension, the column heading also aligns right, but in the new table layout the column heading is unaffected.

I feel this is the wrong behavior.

afbora:
## This PR …

Fixes table layout heading alignment and some RTL issues.

### Fixes
- table layout heading alignment working properly #4630 

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Layout/Table.vue
+++ b/panel/src/components/Layout/Table.vue
@@ -14,6 +14,7 @@
 				<th
 					v-for="(column, columnIndex) in columns"
 					:key="columnIndex + '-header'"
+					:data-align="column.align"
 					:data-mobile="column.mobile"
 					:style="'width:' + width(column.width)"
 					class="k-table-column"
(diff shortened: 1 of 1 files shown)


panel/src/components/Layout/Table.vue
```diff
@@ -418,6 +399,15 @@ export default {
 	height: var(--table-row-height);
 }
 
+/* Text aligment */
+.k-table-column[data-align] {
+	text-align: var(--align) !important;
```
distantnative, line 404: Would be great if we can avoid this `!important`. Is it needed if we would remove the `text-align: start;` from above?

afbora: I already tried but unfortunately when remove the `text-align: start;`, table header centered (seems default behaviour, not via custom css)

afbora: Like that when your suggestion applied

captures_chrome-capture-2022-8-6 (1)

distantnative (approved):

PR #4633, opened 2022-09-06
`Disk quota exceeded` error handling

afbora:
## This PR …

When the disk quota was exceeded on the server, the data was lost by overwriting the contents empty. The `F::write` method now tries to overwrite a temporary file (`default.txt~`). If successful, the temporary file will be overwritten(rename) with the original file. So the original file is always safe.

### Enhancements
- Data/content is now safe when server disk quota exceeded

### Breaking changes

- Contains a breaking change for those using `~` at the end of the content filenames: `/content/site.txt~`

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -889,10 +889,15 @@ class F
 			}
 		);
 
-		// if everything is ok rename temporary file with original
-		if ($result === true) {
-			rename($temp, $file);
-			return true;
+		try {
+			// if everything is ok, rename temporary file with original
+			if ($result === true) {
+				rename($temp, $file);
+
+				return true;
+			}
+		} catch (Throwable) {
+			$result = false;
 		}
 
 		// removes the temporary file if the result is failed (on warnings)
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -4,6 +4,7 @@ namespace Kirby\Filesystem;
 
 use Exception;
 use IntlDateFormatter;
+use Kirby\Cms\App;
 use Kirby\Cms\Helpers;
 use Kirby\Toolkit\I18n;
 use Kirby\Toolkit\Str;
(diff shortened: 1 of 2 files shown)


src/Filesystem/F.php
```diff
@@ -860,6 +862,49 @@ public static function write(string $file, mixed $content, bool $append = false)
...
+		// @codeCoverageIgnoreEnd
+
+		try {
+			// if everything is ok, rename temporary file with original
+			if ($result === true) {
+				rename($temp, $file);
+
+				return true;
+			}
+		} catch (Throwable) {
```
lukasbestle, line 901: When will `rename()` throw an exception? I believe the exception again comes from Whoops only.

lukasbestle (commented):
We need some kind of race condition handling to cover cases like this:

```
Process 1     Process 2

Write Temp
Content 1

              Write Temp
              Content 2

Move File
Content 2
(incomplete)

              Move File
              Error because temp file is gone!
```

A possible solution could be a write lock on the temp file so that only one process can access the file at the same time. But even that may not be entirely safe.

src/Filesystem/F.php
afbora: I'm getting error on `testWriteHandleWarning` unit test without try catch block. Do you?

afbora:
````
rename(\kirby\tests\Filesystem/tmp/test~,\kirby\tests\Filesystem/tmp/test): Access Denied (code: 5)
 \kirby\src\Filesystem\F.php:896
 \kirby\tests\Filesystem\FTest.php:847
 ````

afbora:
> We need some kind of race condition handling to cover cases like this:

I have no idea how to test this.

> A possible solution could be a write lock on the temp file so that only one process can access the file at the same time.

We are already using the `LOCK_EX` mode. Doesn't this work?

src/Filesystem/F.php
lukasbestle: That's likely a warning and not an exception though. For PHP warnings and errors we need the error handler. We cannot catch these with `try catch` if Whoops is disabled.

lukasbestle:
> We are already using the `LOCK_EX` mode. Doesn't this work?

You are right, this should catch most cases (including the one from my example). There could still be a race condition when another process starts writing just before the file is renamed in the current process, because the exclusive lock is lifted just before the renaming operation. If this race condition ever occurs, it would be a regression and a hard to find one. But I'm not sure if there's any way to keep the whole write and rename process locked.

lukasbestle:
Another potential issue that came to my mind: Does appending work with the current setup? Because the temp file does not exist before the operation, we are probably currently appending to an empty temp file and then moving this almost empty temp file over the original file.

bastianallgeier:
Have you checked performance implications?

src/Filesystem/F.php
afbora: Yes, this is warning.

afbora:
@lukasbestle To be honest, I'm starting to get a little lost 🙈 What should we do or how should we proceed?

It would be a very useful development, but if there are too many edge-cases or if it will cause regression, let's put it aside for now. Because I can't comment on the edge-cases you mentioned.

lukasbestle:
Yeah, I think at the moment this is too risky.

afbora:
I'm closing it for now to review again in the coming days/versions.

PR #4640, opened 2022-09-14
New object field

bastianallgeier:
### New object field

The new object field allows to create nested yaml objects. This is super handy for more complex settings or nested structures. The object field definition is very similar to a structure field. You can define any set of fields for the object with the fields option. 

```yaml
person:
  label: Person
  type: object
  fields:
    name:
      type: text
    phone:
      type: tel
    url:
      type: url
    tags: 
      type: tags
```

Screenshot 2022-09-14 at 15 32 39

The result is stored as yaml in the content file and can be used in your templates with the new `$field->toObject()` method.

```php
<?php $person = $page->person()->toObject(); ?>
<dl>
  <dt>Name</dt>
  <dd><?= $person->name() ?></dd>
  <dt>Phone</dt>
  <dd><?= $person->phone() ?></dd>
  <dt>URL</dt>
  <dd><?= $person->url() ?></dd>
  <dt>Tags</dt>
  <dd><?= $person->tags() ?></dd>
</dl>
<?php endif ?>
```

The object is a content object with regular Kirby field methods. This means that you can use all the typical methods for object fields, that you would normally use for page fields. You can even add files and related pages with files and pages fields to an object. Pretty much like you would in a structure field. 

## Ready?

### For review team


lukasbestle:
So it's basically a structure field with exactly one item, right? How and in which cases would you nest them as you describe?

bastianallgeier:
@lukasbestle yes, exactly. It adds the possibility to build quite complex data structures. 

Let's stick with the person example from above.

```yaml
person:
  label: Person
  type: object
  fields:
    name:
      type: text
    homeAddress:
      type: object
      label: Home Address
      fields:
        street:
          type: text
        zip:
          type: text
        city:
          type: text
    workAddress:
      type: object
      label: Work Address
      fields:
        street:
          type: text
        zip:
          type: text
        city:
          type: text
    phone:
      type: tel
    url:
      type: url
    tags:
      type: tags
```

https://user-images.githubusercontent.com/24532/190186659-7cd0991e-e87f-4804-9e90-d58bf5bf46ee.mp4

lukasbestle:
Ah, that's awesome! Is there an easy way we could add a `preview` prop with a query template that would be displayed for such nested objects instead of the curly braces object placeholder?

bastianallgeier:
@lukasbestle yes, definitely. It was just beyond the scope for the first version of the field, but there's so much potential for all of this. The previews are all based on the existing preview components that we use for the structure field table. The better those preview components get, the better the structure field and the object field get at the same time.

panel/src/components/Layout/Table.vue
```diff
@@ -211,7 +211,7 @@ export default {
 		 */
 		hasOptions() {
 			return (
-				this.options ||
+				(this.options && this.options.length > 0) ||
```
distantnative: `this.options?.length`

afbora:
At first, I really like the field 💛 I've few points:

1. **Empty behavior for person one in screenshot**
When the field is empty but field is not required, it looks like data needs to be entered but it is blank. Would it make more sense to open a button like "enter data" after clicking it? I'm not sure, just an idea.

2. **Responsive for person two in screenshot**
If the field value is too long, it shifts on mobile. It would be nice if something like `text-overflow: ellipsis` was set.

captures_chrome-capture-2022-8-15

config/fields/object.php
```diff
@@ -0,0 +1,80 @@
...
+
+			if (empty($errors) === false) {
+				// use the first error for details
+				$error = A::first($errors);
+
+				throw new InvalidArgumentException([
+					'key'  => 'object.validation',
+					'data' => [
+						'label' => $error['label'],
+						'message' => A::first($error['message'])
```
distantnative: Why are we here only using the first error and the first message?

panel/src/components/Drawers/FormDrawer.vue
```diff
@@ -80,10 +80,14 @@ export default {
 
 			if (focus !== false) {
 				setTimeout(() => {
-					let autofocus =
-						Object.values(this.fields).filter(
+					let autofocus = focus;
+
+					if (autofocus === true) {
```
distantnative: This looks a bit weird, we first check `focus` to not be `false` to then immediately check `autofocus` (which is just `focus`) to be `true`?

panel/src/components/Forms/Field/ObjectField.vue
```diff
@@ -0,0 +1,87 @@
...
+
+export default {
+	mixins: [Field, Input],
+	props: {
+		fields: Object,
+		value: Object
+	},
+	data() {
+		return {
+			autofocus: null,
```
distantnative: `autofocus` is already a prop inherited from `Input.vue`, which is why Vue throws a console error

```
The data property "autofocus" is already declared as a prop. Use prop default value instead.
```

distantnative:
I would agree with @afbora that currently required/non-required fields don't get any differentiation in the preview, which makes it hard to see without opening the drawer if something still needs to be filled in or not.

config/fields/object.php
afbora: I agree. Should be validate all fields and shows all errors, not first.

This is also related with https://github.com/getkirby/kirby/pull/4629 enhancement since there are multiple field errors. So error should be thrown like following:

````php
throw new InvalidArgumentException(implode("\n", ['Error one', 'Error two', 'Error three']));
````

**Comment by distantnative** [comment]
One more question: Are conditional fields via `when` supported?

**Comment by afbora** [comment]
I've tested `when` prop. Works generally. But there are few issues:

- validation doesn't work properly
- a field with the `when` prop is still visible in the table preview.

**Comment by bastianallgeier** [comment]
I've solved the JS and CSS issues and added an empty state for objects. The validation part is still WIP. 

https://user-images.githubusercontent.com/24532/190629785-dc1ca749-488a-4c8b-8aab-50b02f5f59e1.mp4

**Comment by bastianallgeier** [comment]
Unit tests are now done and validation is slightly improved. Unfortunately, we don't really have a good way to create nice validation dialog messages for nested fields. That's why I stopped on the first field. I think we need to improve this with the new validation architecture.

**Comment by distantnative** [comment]
Ideally, the field should also be validated when closing the drawer - and show a red outline when invalid. But I think that will be easier with the blueprint factoring. So maybe it's fine for now.

**Review by distantnative — DISMISSED** [review]

**Comment by bastianallgeier** [comment]
@distantnative validation really is a tricky problem with nested fields. In our vision of a field-based async validation on blur or input, this problem will be handled completely differently. I think there's no chance to get this right at the moment.

**Comment by distantnative** [comment]
@bastianallgeier agreed - are you fine with the simple validation checks I added? They don't fully do the trick, but at least for some basics like required fields.

**Comment by afbora** [comment]
Getting `Cannot read properties of null (reading 'name')`

Try to use following blueprint with select `design` and no object value.

````yaml
  category:
    label: Category
    type: select
    options:
      design: Design
      architecture: Architecture
      photography: Photography
      3d: 3D
      web: Web
      
  person:
    label: Person
    type: object
    when:
      category: design
    fields:
        ...
        ...

**Comment by afbora** [comment]
Another issue is about validation. 

**Follow the steps:**

1. Select `design`
2. Add empty person object (do not fill required inputs)
3. Select another category
4. Try to save (not for drafts) or save + publish (for drafts)
5. See issue (expected to save)

````yaml
fields:
  category:
    label: Category
    type: select
    options:
      design: Design
      architecture: Architecture
      photography: Photography
      3d: 3D
      web: Web
      
  person:
    label: Person One
    type: object
    when:
      category: design
    fields:
      category:
        label: Category
        type: select
        options:
          design: Design
          architecture: Architecture
          photography: Photography
          3d: 3D
          web: Web
      name:
        type: text
        required: true
      phone:
        type: tel
        required: true
      url:
        type: url
        required: true
        when:
          category: design
      tags:
        type: tags
        required: true
````

**Comment by bastianallgeier** [comment]
@afbora weird. I tried to reproduce your issues, but it's working all fine for me.

**Comment by afbora** [comment]
@bastianallgeier I've updated my local repo and dist files. Both issues still continue.

**Comment by afbora** [comment]
Also third issue is showing like invalid even not required

captures_chrome-capture-2022-8-20

**Comment by distantnative** [comment]
@afbora Oh that's a conceptual question: I made the field invalid when a du field is required but empty. But if the field itself isn't required, that's stupid. Maybe the condition needs to be: invalid if fields are required but empty while at least one subfield is not empty.

**Comment by bastianallgeier** [comment]
@distantnative to be honest, I would remove the entire dynamic invalidation code. None of that will ever be truly stable as long as the fields don't properly validate themselves in the frontend. I think it's more confusing than helpful if the object seems to be invalid in cases when it is actually fine or vice versa.

**Comment by afbora** [comment]
@distantnative If we think of it as a single field, it may make sense, but if more than one field changes according to the value of a field, it may not make sense to have all of them required.

File: panel/src/components/Forms/Field/ObjectField.vue [hunk 5]
```diff
@@ -0,0 +1,176 @@
...
+			// if field itself is required and empty
+			if (this.required === true && this.isEmpty) {
+				return true;
+			}
+
+			// if subfields are required but empty
+			for (const field in this.fields) {
+				if (
+					this.fields[field].required === true &&
+					this.$helper.object.isEmpty(this.value[field]) === true
```
afbora:
```suggestion
					this.$helper.object.isEmpty(this.value?.[field]) === true
```

This fixes the https://github.com/getkirby/kirby/pull/4640#issuecomment-1252015259 issue

distantnative:
True, conditions aren't even a part of this so far. Then I'm with @bastianallgeier - best to throw it out again.

afbora:
Yes it makes sense to remove it. For example; you want to allow the user to add a person/contact. But not required (think of it as appearing on the frontend only if it adds). But if the user adds a person, you definitely want to make the name mandatory.

bastianallgeier:
I think backend validation is already taking care of validation objects correctly. That's the most important part. The UI stuff is nice to have but not possible in a good way with our current setup. I'll remove it again from the component.

panel/src/components/Forms/Field/ObjectField.vue
```diff
@@ -0,0 +1,156 @@
...
+		},
+		onDrawerInput(value) {
+			this.object = value;
+			this.$emit("input", this.object);
+		},
+		onErrors(fields) {
+			this.errors = Object.keys(fields).filter(
+				(field) => fields[field].$invalid
+			);
+		},
```
distantnative: @bastianallgeier you want to keep these parts of validation?

bastianallgeier: Damn, I missed that.

afbora:
Looks great 💛 

But I have one more small question. sorry for that @bastianallgeier 
None of our fields save data as `null` instead empty string. But it didn't seem consistent to me as the object field saves `null` when there is no data. For example; It will not work for `$field->exists()` .

I know it's because of the empty state but still maybe you have a solution in mind. What do you think?

I don't want to complicate PR any more 🙈 It would be useful to say again that it is great as it is.

afbora (dismissed):
❤️

distantnative (commented):
I think that's a valid point though: wouldn't it still work if we save an empty string? As we check for empty on value, that would still return true.

afbora:
I thought it would break whether the form is open or closed when the field is empty, but yes it still seems to work great with the empty string.

config/fields/object.php
```diff
@@ -0,0 +1,104 @@
...
+			return new Form([
+				'fields' => $this->attrs['fields'],
+				'values' => $values,
+				'model'  => $this->model
+			]);
+		},
+	],
+	'save' => function ($value) {
+		if (empty($value) === true) {
+			return null;
```
afbora:
```suggestion
			return [];
```

config/fields/object.php
```diff
@@ -0,0 +1,104 @@
...
+				throw new Exception('Please provide some fields for the object');
+			}
+
+			return $this->form()->fields()->toArray();
+		},
+		'value' => function () {
+			$data = Data::decode($this->value, 'yaml');
+
+			if (empty($data) === true) {
+				return null;
```
afbora:
```suggestion
				return [];
```

lukasbestle:
In the long term I think `null` makes a lot of sense. The casting to an empty string should happen in the storage layer (Data class for example).

But until the form refactoring it's probably best to keep it consistent.

bastianallgeier:
You are right. I had to fix in a couple more places but now it should work with empty strings.

afbora (approved):

PR #4646, opened 2022-09-15
Fix disabled cache on empty auth header

### LINKED ISSUE #4634: Pages cache not working (neither stored not used) when checking for logged in user
## Description
Having read @lukasbestle's explanation in #4423, I figured the following would work. [This thread](https://forum.getkirby.com/t/pages-with-protected-content-cannot-be-cached/26387/2) in the forum suggests this should work too so I'm mentioning this as an issue. Starting with a fresh starterkit, I activate the cache in `config.php`:

```php
'cache' => [
    'pages' => [
        'active' => true
    ]
]
```

Pages are stored. Let's say I want to add the ability for my logged in users to have an edit button available on any front-end page which redirects to this very same page in the panel. I clear the cache and I add in my `header.php`:

```php
<?php if(kirby()->user()): ?>
    <a class="edit" href="<?php echo $page->panel()->url() ?>">Edit content</a>
<?php endif; ?>
```

As soon as I do, the cache stops storing pages, the `Cache-Control: no-store, private` header is always set despite me not being logged in. Here's a video of this:

https://user-images.githubusercontent.com/14079751/189099427-fdbd6cd6-ac20-4d42-b355-d5e5d695af95.mp4

## Your setup
Kirby 3.7.5

lukasbestle:
### Fixes

- Caching is no longer disabled when the request passes an empty `Authorization` header. #4634
- Calling `$app->user()` no longer sets the response flag for "uses auth header" unless the `api.basicAuth` option is enabled.

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Auth.php
+++ b/src/Cms/Auth.php
@@ -716,6 +716,10 @@ class Auth
 
 		if (
 			$basicAuth === true &&
+
+			// only get the auth object if the option is enabled
+			// to avoid triggering `$responder->usesAuth()` if
+			// the option is disabled
 			$request->auth() &&
 			$request->auth()->type() === 'basic'
 		) {
--- a/src/Cms/Auth.php
+++ b/src/Cms/Auth.php
@@ -712,9 +712,13 @@ class Auth
 	public function type(bool $allowImpersonation = true): string
 	{
 		$basicAuth = $this->kirby->option('api.basicAuth', false);
-		$auth      = $this->kirby->request()->auth();
+		$request   = $this->kirby->request();
 
-		if ($basicAuth === true && $auth && $auth->type() === 'basic') {
+		if (
+			$basicAuth === true &&
+			$request->auth() &&
+			$request->auth()->type() === 'basic'
+		) {
 			return 'basic';
 		}
 
--- a/src/Http/Request.php
+++ b/src/Http/Request.php
@@ -151,7 +151,7 @@ class Request
 			$kirby->response()->usesAuth(true);
 		}
 
-		if ($auth = $this->options['auth'] ?? $this->header('authorization')) {
+		if ($auth = $this->authString()) {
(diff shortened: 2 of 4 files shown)


src/Cms/Auth.php
```diff
@@ -712,9 +712,13 @@ public function track(string|null $email, bool $triggerHook = true): bool
...
 	{
 		$basicAuth = $this->kirby->option('api.basicAuth', false);
-		$auth      = $this->kirby->request()->auth();
+		$request   = $this->kirby->request();
 
-		if ($basicAuth === true && $auth && $auth->type() === 'basic') {
+		if (
+			$basicAuth === true &&
+			$request->auth() &&
+			$request->auth()->type() === 'basic'
```
distantnative, line 724: Could be shortened to `$request->auth()?->type() === 'basic'`

lukasbestle: No, because `$request->auth()` can return `false`, which would lead to "call to type() on bool".

distantnative: 👍

distantnative: But can I ask what this change is for? It seems to have the same functionality as before, just now spelling out `$request->auth()` twice instead keeping it in a variable

lukasbestle: See the unit tests. If we call `auth()` beforehand, it will always trigger `$app->response()->usesAuth()`, even if the auth is never actually used.

distantnative: Maybe we can add a comment for us others to later still understand that basically `$basicAuth === true` is the condition that really needs to be true before we even call `->auth()`. I'm afraid we'll read it in the future and this won\t be so clear to us anymore.

distantnative (approved):

Issue #4518, opened 2022-07-21
DB query where clause breaks when parameter is 0


toebu:
When you have a db query, and you try to set `0` as the parameter (`$query->where('field', '=', 0);`), the created query is invalid.
It looks like this is because in [`Kirby\Database\Query::filterQuery`](https://github.com/getkirby/kirby/blob/main/src/Database/Query.php#L944) you use `in_array()` without strict mode. The last argument will then always be removed and treated as the mode instead of the parameter.

This would be fixed by using strict mode `in_array($mode, ['AND', 'OR'], true)`. Probably, this would also have to be changed in other places, such as in [andWhere](https://github.com/getkirby/kirby/blob/main/src/Database/Query.php#L944) and [orWhere](https://github.com/getkirby/kirby/blob/main/src/Database/Query.php#L410).

I'm using the workaround to for now always explicitly state the mode `$query->where('field', '=', 0, "AND');`.

A related problem that comes up here: if you're parameter happens to be "AND" or "OR", this will also not work as intended... But that's a bit more tough to fix.

lukasbestle:
> This would be fixed by using strict mode `in_array($mode, ['AND', 'OR'], true)`

To be honest I don't see how the strict mode will change the behavior. `in_array(0, ['AND', 'OR'])` without the strict mode and `in_array(0, ['AND', 'OR'], true)` both return `false`.

> A related problem that comes up here: if you're parameter happens to be "AND" or "OR", this will also not work as intended

`AND` and `OR` (unquoted!) are not valid values in SQL. Is there a use case I missed?

toebu:
> To be honest I don't see how the strict mode will change the behavior. `in_array(0, ['AND', 'OR'])` without the strict mode and `in_array(0, ['AND', 'OR'], true)` both return `false`.

I searched around a bit more, and found out this is depends on the PHP version. I'm using PHP 7.4, so for me the two expressions return something else. In PHP 8 this was fixed, that's why you see another behaviour. See for example https://wiki.php.net/rfc/string_to_number_comparison.

Kirby says it supports PHP 7.4, so one should fix it for this PHP version. There, you have to explicitly tell in_array to use strict mode.

> > A related problem that comes up here: if you're parameter happens to be "AND" or "OR", this will also not work as intended
> 
> `AND` and `OR` (unquoted!) are not valid values in SQL. Is there a use case I missed?

The 'AND' will be passed with sql parameter binding, so quotation happens after the code that is problematic.

The use case is this:
[...]

PR #4650, opened 2022-09-17
`Database\Query`: Separate `$mode` for `filterQuery()`

### LINKED ISSUE #4564: `Database\Query`: Separate `$mode` for `filterQuery()`
https://github.com/getkirby/kirby/issues/4518#issuecomment-1212402524

distantnative:
### Refactoring
- `Database\Query`: Added new third parameter `$mode` for `filterQuery()`
#4564

## Ready?

### For review team

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -404,18 +404,7 @@ class Query
 	 */
 	public function orWhere(...$args)
 	{
-		$mode = A::last($args);
-
-		// if there's a where clause mode attribute attached…
-		if (in_array($mode, ['AND', 'OR'], true) === true) {
-			// remove that from the list of arguments
-			array_pop($args);
-		}
-
-		// make sure to always attach the OR mode indicator
-		$args[] = 'OR';
-
-		$this->where(...$args);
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):

lukasbestle:
Does this fix the bug? Given that the last argument is still used, I think you can still not pass a literal `AND` or `OR` without it getting treated as mode.

distantnative:
Then I didn't understand the bug, I thought I followed the description from https://github.com/getkirby/kirby/issues/4518#issuecomment-1212402524 1:1.

distantnative:
Do I get you right then, that `filterQuery` shouldn't at all test for this anymore?

```php
$last   = A::last($args);

// if there's a where clause mode attribute attached…
if (in_array($last, ['AND', 'OR'], true) === true) {
```

?

lukasbestle:
Yes, I think so. This would be a breaking change, but with the `->orWhere()` and `->andWhere()` methods there are more clean alternatives.

If we don't want a breaking change, we could add a deprecation warning instead.

PR #4653, opened 2022-09-17
Table: fix mobile width

### LINKED ISSUE #4415: table layout: strange results when combining mobile & width
## description
i am working with the new `layout: table` option and so far i really love it! 🥰 ✨ i wanted to show more content on mobile, so i played around with the new `mobile: true` option for columns. it works, but i feel that when combined with smaller width options, it can produce strange / unexpected results.

in my example i have a "total" column, where i set the option `width: 1/12`. looks & works great on desktop. but now when viewing the column on mobile, it still uses the `1/12` width, cutting off nearly the entire content.

desktop

width

**expected behavior**
i was expecting that when on a mobile breakpoint, any set width options would be ignored and column sizes were set to `auto`. this only made sense to me, because it is possible to hide columns on mobile that are visible on desktop. so already set widths might lead to new, unexpected grid layouts on mobile.

no-width

another solution i can think of would be adding custom widths for mobile, maybe like so:

```yml
columns:
    total:
        label: Total
        width: 1/12
        mobile:
            enabled: true
            width: 1/2
```

## to reproduce
add a pages section with the following settings:

```yml
headline: Headline
type: pages
layout: table
columns:
    total:
        label: Total
        width: 1/12
        mobile: true
```

---

thank you so much for looking into this!
such a great release!
🌻

distantnative:
### Fixes
- Table layout: on narrow viewports all columns are now assigned an auto width instead of the defined columns size
#4415

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Layout/Table.vue
+++ b/panel/src/components/Layout/Table.vue
@@ -16,7 +16,7 @@
 					:key="columnIndex + '-header'"
 					:data-align="column.align"
 					:data-mobile="column.mobile"
-					:style="'width:' + width(column.width)"
+					:style="'--width:' + width(column.width)"
 					class="k-table-column"
 					@click="
 						onHeader({
@@ -90,7 +90,7 @@
 					:row="row"
 					:mobile="column.mobile"
 					:value="row[columnIndex]"
-					:style="'width:' + width(column.width)"
+					:style="'--width:' + width(column.width)"
 					class="k-table-column"
 					@click.native="
 						onCell({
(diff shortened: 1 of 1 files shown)


afbora:
As [I said before](https://github.com/getkirby/kirby/issues/4415#issuecomment-1174130856), this change doesn't quite fix the issue:

captures_chrome-capture-2022-8-19

distantnative:
Mmm liked for me fine with structure field table, guess sections might behave differently

distantnative:
Closed as not working

Issue #4413, opened 2022-06-22
Dialogs: disable buttons while processing


teichsta:
## Description

our users sometimes report a TypeError which i was not able to reproduce until today.

The users are mainly Windows-User and hence used to doubleclick on things – even on the web. If you try to rename a page and commit the dialog with a (quick) doubleclick on change you can safely reproduce this issue:

`TypeError: Cannot read properties of undefined (reading 'error') at a.onSubmit (index.js:formatted:2076:35)`

The doubleclick also seems to be the reason for error reports when duplicating and deleting pages where the dialog says “page exists already” or “page is already deleted” which is correct from kirbys point of view as the first click of the doubleclick leads to intended change already.

**Expected behavior**  

No error should occur and ideally doubclicks should be prevented. A Forum-User (Manuel) suggested here https://forum.getkirby.com/t/dialogs-typeerror-on-double-click/25651/4?u=teichsta to disable buttons after click for some time useing the `setTimeout()` method.

## To reproduce

1. Go to panel
2. Click on "rename" or "delete" page
3. Doubleclick and "OK", "Rename"

## Your setup
**Kirby Version**  
3.6.2

(see forum post: https://forum.getkirby.com/t/dialogs-typeerror-on-double-click/25651)

afbora:
I can't reproduce the issue with fast double click 🙂 but can with running following JS code in browser console:

````js
document.getElementById("unique-id").click();
document.getElementById("unique-id").click();
````

teichsta:
it turns out, that i wasn't able to reproduce this `TypeError` either BUT there is still an issue with double clicks. With a new starterkit i could reproduce this error:

1. click "add" new page
2. type title for new page
3. doubleclick "create draft"

It was also possible with "rename page" and "delete page" after some tries (sometimes it didn't occured on first try)

<img width="1502" alt="image" src="https://user-images.githubusercontent.com/2108636/175261990-4b47dd53-f727-4a40-a872-619f93c5cdae.png">

or kirby shows this dialog:

<img width="650" alt="image" src="https://user-images.githubusercontent.com/2108636/175263791-8b9ec790-e34a-4226-aeb2-3a5c3612ff42.png">

sometimes the dialog shortly appears in the background and get's closed automatically.

bastianallgeier:
✅

afbora:
@distantnative I can still reproduce the issue with running lines from console. This means that the issue will still occur if clicked too fast.

````js
document.getElementById("unique-id").click();
[...]

PR #4654, opened 2022-09-17
FiberDialog: disabled submit btn while processing

distantnative:
### Enhancement
- Fiber dialogs: the submit button is now disabled while processing a submit request to prevent sending multiple requests at once accidentally
#4413

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Dialogs/FiberDialog.vue
+++ b/panel/src/components/Dialogs/FiberDialog.vue
@@ -2,8 +2,9 @@
 	<component
 		:is="component"
 		ref="dialog"
-		:visible="true"
 		v-bind="props"
+		:disabled="isProcessing"
+		:visible="true"
 		@cancel="onCancel"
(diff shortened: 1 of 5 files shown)


bastianallgeier (approved):

bastianallgeier:
Love it!

panel/src/components/Dialogs/FiberDialog.vue
```diff
@@ -29,6 +35,7 @@ export default {
 		},
 		async onSubmit(value) {
 			let dialog = null;
```
afbora, line 37: @distantnative I've an idea about that. Actually, while the process is in progress, how about we disable the button while preventing a new action from being made at the same time?

````js
  // do not handle a new request while a request is in progress
  if (this.isProcessing === true) {
    return false;
  }

  let dialog = null;
  this.isProcessing = true;
````

PR #4656, opened 2022-09-17
`F::load()`: guard from output

### LINKED ISSUE #4391: Corrupted media on first serve because of stray whitespace
## Description
Before media files are moved to the media folder - when they are server through Kirby/PHP. If you accidentally output something you will corrupt the file thats about to be served. 

This is as easy as starting `<?php` on second line in a plugin instead of first line.

## Expected behavior 
It would be great if there was some safety. One would either get error or the output was thrown away. It is really hard to debug. I spend multiple hours trying to debug something that turned out to be empty line in beginning of file in some plugin. It's hard to notice because all other stuff - json/html is fine plus after the first page load its also fine for most users.

## To reproduce
Output whitespace anywhere - in entry `index.php`, `config/config.php`, `plugins/index.php` like
```php
//have

//empty line here
<?php
 mycode();
```

Tested on Kirby 3.6.6. but i assume it happens on all versions.

distantnative:
### Enhancement
- `F::load()` and `F::loadOnce()`: guarded from unintended output 
#4391

## Ready?

### For review team

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -187,7 +187,7 @@ class Response
 	public static function guardAgainstOutput(Closure $callback, ...$args): mixed
 	{
 		$before = headers_sent();
-		$result = $callback($args);
+		$result = $callback(...$args);
 		$after  = headers_sent($file, $line);
 
 		if ($before === false && $after === true) {
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -184,6 +184,10 @@ class Response
 		die(static::redirect($url, $code));
 	}
 
+	/**
+	 * Ensures that the callback does not produce the first body output
+	 * (used to show when loading a file creates side effects)
+	 */
 	public static function guardAgainstOutput(Closure $callback, ...$args): mixed
 	{
 		$before = headers_sent();
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -2,7 +2,9 @@
 
 namespace Kirby\Http;
 
+use Closure;
 use Exception;
+use Kirby\Exception\LogicException;
 use Kirby\Filesystem\F;
 use Throwable;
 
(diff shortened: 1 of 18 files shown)


src/Http/Response.php
```diff
@@ -182,6 +184,19 @@ public static function go(string $url = '/', int $code = 302): void
...
 	}
 
+	public static function guardAgainstOutput(Closure $callback, ...$args): mixed
+	{
+		$before = headers_sent();
+		$result = $callback($args);
+		$after  = headers_sent();
+
+		if ($before === false && $after === true) {
+			throw new LogicException('Disallowed output from file $file:$line, possible accidental whitespace?');
```
lukasbestle:
```suggestion
		$after  = headers_sent($file, $line);

		if ($before === false && $after === true) {
			throw new LogicException("Disallowed output from file $file:$line, possible accidental whitespace?");
```

distantnative: It's that easy?? 😅👍

distantnative:
@lukasbestle not sure if this test is actually working or I am testing it wrong. A unit test like this does not fail (while it probably should):

```php
/**
 * @covers ::load
 */
public function testLoadWithInvalidOutput()
{
	F::load(__DIR__ . '/fixtures/f/load/FileWithOutput.php');
}
```

With a file

```php
// empty line

// empty line
<?php
return [];
```

lukasbestle:
I think `headers_sent()` doesn't work on the CLI, so it might not be possible to write unit tests for it.

distantnative:
Hmm I see. 
@iskrisis could you please check if this PR solves your issue?

iskrisis:
@distantnative it does seem to work for the problem i eccountered

lukasbestle (approved):
I have found a way to test this on the CLI by mocking the `headers_sent()` function. It even showed a minor bug that I missed during visual review.

@distantnative Could you take a final look at the changes and tests?

distantnative:
Looks good to me!

PR #4659, opened 2022-09-17
System view PR tweaks

distantnative:
Tweaked a little the `k-stats` setup to get rid of the `data-click` styling hack.
And rephrased system view a bit to make it even easier to follow what's happening.

### DIFF
--- a/panel/src/components/Layout/Stats.vue
+++ b/panel/src/components/Layout/Stats.vue
@@ -1,11 +1,11 @@
 <template>
 	<dl class="k-stats" :data-size="size">
 		<component
-			:is="isLink(report)"
+			:is="component(report)"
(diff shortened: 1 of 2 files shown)


panel/src/components/Layout/Stats.vue
```diff
@@ -27,14 +25,27 @@ export default {
 		}
 	},
 	methods: {
```
lukasbestle, line 27: I think the method naming of `isLink` (suggests a boolean) and `onClick` (suggest a handler) now doesn't really match the purpose of these methods. What about `component` and `target`?

distantnative: I like 👍

PR #4660, opened 2022-09-17
Blocks: Delete empty text and list block w/ backspace

distantnative:
### Enhancements
- Empty text and list blocks get deleted when hitting backspace and block above gets focused
- `Ctrl + Enter` brings up block selector to append new block after current one

## Wanted - but still struggling to achieve
- [ ] Support list input rules in text block and when matched, split text block at current position/selection into two text blocks and insert a new list block in between

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Blocks/Block.vue
+++ b/panel/src/components/Forms/Blocks/Block.vue
@@ -11,10 +11,12 @@
 		:data-translate="fieldset.translate"
 		class="k-block-container"
 		tabindex="0"
-		@keydown.ctrl.shift.down.prevent="$emit('sortDown')"
-		@keydown.ctrl.shift.up.prevent="$emit('sortUp')"
 		@focus="$emit('focus')"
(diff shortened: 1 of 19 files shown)


distantnative:
Needs to be revisited once I have a better idea how to handle splitting blocks up - only then this really makes sense.

doup:
@distantnative was the idea to also be able to split a text block at an arbitrary position?

Imagine you have a long block of text, and you want to add an image between two paragraphs… right now it's not user friendly at all. It would nice if there was some shortcut AND maybe some scissors button that floats at the same height as the cursor.

The logical follow up for me would be that if two text blocks are in succession, then with backspace and/or via floating button be able to merge both of them to a single text block, but respecting the paragraphs.

Should I create an issue? Or is this managed in Nolt? Is it planned? 😇 
This is one of those papercuts fix that would make editors lifes easier.

PR #4661, opened 2022-09-18
New `$cache->enabled()` method

lukasbestle:
**⚠️ Requirement for the update check feature**

### Features

- New `$cache->enabled()` method to check if the cache is ready to store values beyond the current request

### Breaking changes

None

## Migration guide (plugin devs)

Custom cache drivers should implement the new `enabled()` method. The method will be required in a future Kirby release. Until you implement this method, Kirby will assume that your cache driver is always enabled.

## Ready?

### For review team

### DIFF
--- a/src/Cache/MemoryCache.php
+++ b/src/Cache/MemoryCache.php
@@ -19,6 +19,15 @@ class MemoryCache extends Cache
 	protected array $store = [];
 
 	/**
+	 * Returns whether the cache is ready to
+	 * store values
+	 */
+	public function enabled(): bool
+	{
+		return true;
+	}
+
+	/**
 	 * Writes an item to the cache for a given number of minutes and
 	 * returns whether the operation was successful
 	 *
--- a/src/Cache/ApcuCache.php
+++ b/src/Cache/ApcuCache.php
@@ -16,6 +16,15 @@ use APCUIterator;
 class ApcuCache extends Cache
 {
 	/**
+	 * Returns whether the cache is ready to
+	 * store values
+	 */
+	public function enabled(): bool
+	{
+		return apcu_enabled();
+	}
+
+	/**
 	 * Determines if an item exists in the cache
 	 */
 	public function exists(string $key): bool
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -51,6 +51,18 @@ abstract class Cache
 	}
 
 	/**
+	 * Returns whether the cache is ready to
+	 * store values
+	 */
+	public function enabled(): bool
+	{
+		// TODO: Make this method abstract in a future
+		// release to ensure that cache drivers override it;
+		// until then, we assume that the cache is enabled
+		return true;
+	}
+
+	/**
 	 * Determines if an item exists in the cache
 	 */
 	public function exists(string $key): bool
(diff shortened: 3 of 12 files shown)


src/Cache/MemoryCache.php
```diff
@@ -18,6 +18,15 @@ class MemoryCache extends Cache
 	 */
 	protected array $store = [];
 
+	/**
+	 * Returns whether the cache is ready to
+	 * store values beyond the current request
+	 */
+	public function enabled(): bool
+	{
+		return false;
```
distantnative: Initially I thought this new method is about right at this moment the cache is enabled... but this comment makes me think it's more about whether the cache will persist after the current request. In that case, I think the name is rather misleading. Maybe rather something like `isPersistent`.

distantnative (changes requested):

src/Cache/MemoryCache.php
lukasbestle: Good point.

The method currently only returns `true` if two conditions are fulfilled:

1. The cache is available at all.
2. The cache persists beyond the current request.

Maybe this combination is too confusing. What if we only make it check the first condition. The persistence part can then be checked in the calling methods by simply checking `instanceof NullCache` and `instanceof MemoryCache`. I don't think there are any plugin caches that are not persistent.

lukasbestle: I have now implemented it so that the persistence no longer counts. This means that `MemoryCache` now returns `true`. The only cache that always returns false is `NullCache` as it's really never an active cache.

distantnative: I feel like `NullCache` should be `true` as well - it is designed to be used like an actual cache (just without doing anything). Might contradict situations where a cache type is unknown where checking for enabled would now fail, while so far the NullCache behaviour was the preferred way.

lukasbestle: Thinking about it again, I think returning `false` for `NullCache` was correct:

The only reason why we have this class is to return it as a placeholder when a cache is disabled via config. So any `NullCache` is by definition a disabled cache.

bastianallgeier: I agree that NullCache should return false. It really is just a placeholder instead of not returning any cache instance on an inactive cache.

bastianallgeier (approved):

src/Cache/MemoryCache.php
bastianallgeier: @distantnative could you resolve the conversation if you agree?

distantnative: Agree

bastianallgeier (approved):

bastianallgeier:
@distantnative I think you also need to review it again

distantnative (approved):
@bastianallgeier I think an approval form you would've overruled mine as well

bastianallgeier:
@distantnative nope, that didn't work for some reason.

lukasbestle:
Now NullCache returns true though, doesn't it?

lukasbestle:
BTW: Nico had to approve because he had selected "request changes". This can be dismissed in the merge box of the PR, but otherwise it blocks merging. If you select "comment" then it doesn't block.

PR #4662, opened 2022-09-18
PHP 8: Clean up `Cms\Plugin` class

lukasbestle:


src/Cms/Plugin.php
```diff
@@ -79,25 +73,24 @@ public function authorsNames(): string
 	}
 
 	/**
-	 * @return array
+	 * Returns the associative array of plugin extensions
```
distantnative:
```suggestion
	 * Returns the associative array of extensions the plugin bundles
```

Might make the distinction between plugin vs. extension even clearer

distantnative (dismissed):
One little  suggestion, but looks good to me

distantnative (approved):

PR #4667, opened 2022-09-19
Fix editing escaped tags text

### LINKED ISSUE #4664: [develop] tags field escaping issue when editing the tag
simple tags field escaping when editing

````yaml
tags:
  label: Tags
  type: tags
````

tags-field-issue

_Originally posted by @afbora in https://github.com/getkirby/kirby/issues/4624#issuecomment-1250802381_

distantnative:
Fixes #4664

@afbora can you confirm?

### DIFF
--- a/panel/src/components/Forms/Input/TagsInput.vue
+++ b/panel/src/components/Forms/Input/TagsInput.vue
@@ -210,7 +210,7 @@ export default {
 			this.newTag = null;
 		},
 		edit(tag) {
-			this.newTag = tag.text;
+			this.newTag = this.$helper.string.unescapeHTML(tag.text);
 			this.$refs.input.select();
 			this.remove(tag);
 		},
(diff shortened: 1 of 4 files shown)


afbora:
This PR partially solved the issue, could you try following tags please:

- `Foo "Bar"`
- `Foo 'Bar'`
- Foo `Bar` (with "`")
- Any HTML markup

distantnative:
I think `replaceAll` should solve these

afbora:
All fixed except one that I'm not sure because I'm tried to payload XSS (``) but behavior not consistent. Others works correct but following example not. 

````yaml
tags:
  label: Tags
  type: tags
            
tags_from_other_field:
  label: Tags (Options from other fields)
  type: tags
  options: query
  query: page.tags.split
````

Input: ``
Output: ``

I'm still fine, because other string samples works great!. I think this inconsistent is not important. But would be good if fixed anyway. You know better.

distantnative:
Found the issue

afbora:
Works great for `Foo+Bar=Baz?Low*Kit#Tri$Get#Pot"You'RingYea^Alt(Tar)For[Tea]Sai!Fre{Erf}Def\Top/Lan<Hif>Req-Das‒New.Vue:Und;Sor|Ort~Lui` string 😅

afbora (approved):

PR #4670, opened 2022-09-19
Use CSS’s `aspect-ratio`

distantnative:
## This PR …

Not that it would pay off to raise browser requirements just for this, but I think we might want to raise them slowly anyways. Versions would still apply to 90.11% of global traffic, based on https://caniuse.com/mdn-css_properties_aspect-ratio

Tagging this with 3.8.0 just since we should make a decision there if to include in 3.8 or if it has to wait until 3.9 (due to breaking change).

### Refactoring
- Uses `aspect-ratio` CSS property

### Breaking changes
- Higher browser requirements
  - Chrome 88
  - Edge 88
  - Safari 15
  - Firefox 89
  - Opera 74
  - Chrome for Android 105
  - Safari on iOS 15
  - Opera Mobile 64 

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Layout/AspectRatio.vue
+++ b/panel/src/components/Layout/AspectRatio.vue
@@ -1,9 +1,5 @@
 <template>
-	<span
-		:data-cover="cover"
-		:style="{ 'padding-bottom': ratioPadding }"
-		class="k-aspect-ratio"
-	>
+	<span :data-cover="cover" :style="'--ratio:' + ratio" class="k-aspect-ratio">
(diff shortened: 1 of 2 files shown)


lukasbestle:
To be honest I feel like a refactoring alone does not make higher browser requirements worth it. Or does the change also come with improvements for the users?

distantnative:
You're right. I was thinking it along the lines that we might want to push for modern browsers anyways. Also for getting at some point to container query support. But of course only refactoring isn't a great reason for it.

PR #4675, opened 2022-09-20
`A::merge`: support multiple arrays

distantnative:
### Enhancement
- `A::merge($array1, $array2, $array3, A::MERGE_APPEND)` now supports to merge more then 2 arrays

## Ready?

### For review team

- [X] ~~Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)~~

### DIFF
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -156,47 +156,65 @@ class A
 	/**
 	 * Merges arrays recursively
 	 *
-	 * @param int $mode Behavior for elements with numeric keys;
-	 *                  A::MERGE_APPEND:    elements are appended, keys are reset;
-	 *                  A::MERGE_OVERWRITE: elements are overwritten, keys are preserved
-	 *                  A::MERGE_REPLACE:   non-associative arrays are completely replaced
+	 * If last argument is an integer, it defines the
+	 * behavior for elements with numeric keys;
+	 * - A::MERGE_OVERWRITE:  elements are overwritten, keys are preserved
+	 * - A::MERGE_APPEND:     elements are appended, keys are reset;
+	 * - A::MERGE_REPLACE:    non-associative arrays are completely replaced
 	 */
-	public static function merge(array $array1, array $array2, int $mode = A::MERGE_APPEND): array
+	public static function merge(array|int ...$arrays): array
 	{
-		$merged = $array1;
+		// get mode from parameters
+		$last = A::last($arrays);
+		$mode = is_int($last) ? array_pop($arrays) : A::MERGE_APPEND;
+
+		// get the first two arrays that should be merged
+		$merged = array_shift($arrays);
+		$join   = array_shift($arrays);
 
 		if (
-			static::isAssociative($array1) === false &&
+			static::isAssociative($merged) === false &&
 			$mode === static::MERGE_REPLACE
 		) {
-			return $array2;
-		}
+			$merged = $join;
+		} else {
+			foreach ($join as $key => $value) {
+				// append to the merged array, don't overwrite numeric keys
+				if (
+					is_int($key) === true &&
+					$mode === static::MERGE_APPEND
+				) {
+					$merged[] = $value;
 
-		foreach ($array2 as $key => $value) {
-			// append to the merged array, don't overwrite numeric keys
-			if (is_int($key) === true && $mode === static::MERGE_APPEND) {
-				$merged[] = $value;
+				// recursively merge the two array values
+				} elseif (
+					is_array($value) === true &&
+					isset($merged[$key]) === true &&
+					is_array($merged[$key]) === true
+				) {
+					$merged[$key] = static::merge($merged[$key], $value, $mode);
 
-			// recursively merge the two array values
-			} elseif (
-				is_array($value) === true &&
-				isset($merged[$key]) === true &&
-				is_array($merged[$key]) === true
-			) {
-				$merged[$key] = static::merge($merged[$key], $value, $mode);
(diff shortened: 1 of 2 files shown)


src/Toolkit/A.php
```diff
@@ -156,47 +156,65 @@ public static function join(array|string $value, string $separator = ', '): stri
...
+	 * - A::MERGE_REPLACE:    non-associative arrays are completely replaced
 	 */
-	public static function merge(array $array1, array $array2, int $mode = A::MERGE_APPEND): array
+	public static function merge(array|int ...$arrays): array
 	{
-		$merged = $array1;
+		// get mode from parameters
+		$last = A::last($arrays);
+		$mode = is_int($last) ? array_pop($arrays) : A::MERGE_APPEND;
+
```
lukasbestle, line 170: Because the method now accepts `array|int`, IMO we should ensure that `$arrays` only contains arrays at this point, otherwise the following code will break.

lukasbestle (commented):
Have you tested the performance with large arrays? `array_shift()` and `array_unshift()` are both quite expensive operations because they have to move the whole data in memory back and forth.

src/Toolkit/A.php
distantnative: Won't this be rather expensive for all the regular use cases just to throw an exception for the case when the code would break anyways?

lukasbestle: If we don't check, the method will break in unexpected ways and the resulting PHP warnings will be rather opaque.

lukasbestle:
@distantnative What about this question?

> Have you tested the performance with large arrays? `array_shift()` and `array_unshift()` are both quite expensive operations because they have to move the whole data in memory back and forth.

distantnative:
We don't have any functional performance testing set up, so no.

src/Toolkit/A.php
distantnative: Still, I don't think such a performance-heavy check is warranted for the rare cases where someone passes two integers to an array method. I'd accept it to break in rather opaque ways in those rare cases instead of slowing down 99.9% of the cases.

lukasbestle:
To be honest I'm much more worried about the 99.99% case of just passing two arrays to this method if we talk about performance. `array_splice()` and `array_unsplice()` could really be slow, but we'd have to test this. We could still test this right now with a simple loop and `microtime(true)`. What's important is that the input arrays need to be rather large and nested.

distantnative:
I'm not good at this - would this be reasonable code for testing?

```php
protected function getArray(): array
{
	$array = [];

	foreach (range(0, rand(10, 100)) as $one) {
		$subarray = [];

		foreach (range(0, rand(25, 75)) as $two) {
			$subsubarray = [];

			foreach (range(0, rand(25, 50)) as $three) {
				$subsubarray[] = range(50, 100);
			}

			$subarray[] = $subsubarray;
		}

		$array[] = $subarray;
	}

	return $array;
}

public function testAPerformance()
{
		$array1 = $this->getArray();
		$array2 = $this->getArray();

		$start = microtime(true);
		A::merge($array1, $array2);
		$end = microtime(true);
		var_dump($end - $start);
}
```

lukasbestle:
Looks good in general I'd say. But I'd also add some associative arrays and longer string values into the mix.

distantnative:
My modified array generator:

```php
protected function getArray(): array
{
	$array = [];

	foreach (range(0, rand(10, 100)) as $one) {
		$subarray = [];

		foreach (range(0, rand(25, 75)) as $two) {
			$subsubarray = [];

			foreach (range(0, rand(25, 50)) as $three) {
				$subsubarray[Str::random(10)] = Str::random(25);
			}

			$subarray[Str::random(15)] = $subsubarray;
		}

		$array[] = $subarray;
	}

	return $array;
}
```

## Results
### `develop`
0.00037407875061035156
0.0003819465637207031
0.0004010200500488281
0.0005331039428710938
0.00042510032653808594
0.0004038810729980469
0.0004420280456542969
0.0004417896270751953
0.00039505958557128906
0.0004391670227050781

### This PR
0.00036978721618652344
0.00048089027404785156
0.00042700767517089844
0.0004279613494873047
0.00045108795166015625
0.00038504600524902344
0.0004589557647705078
0.00039505958557128906
0.0005130767822265625
0.00040912628173828125


I don't think based on this test that there is any indication for worse performance through this PR.

tests/Toolkit/ATest.php
```diff
@@ -209,36 +209,140 @@ public function testMapWithClassMethod()
...
+		$b        = [1 => 'c', 3 => 'd', 5 => 'a'];
+
+		// A::MERGE_APPEND
+		$expected = ['a', 'b', 'c', 'd', 'a'];
+		$result   = A::merge($a, $b);
+		$this->assertSame($expected, $result);
+		$result   = A::merge($a, $b);
+		$this->assertSame($expected, $result, true);
+		$result   = A::merge($a, $b);
+		$this->assertSame($expected, $result, A::MERGE_APPEND);
```
lukasbestle: I think the last argument was supposed to be passed to `A::merge()`, not the assertion, right?

lukasbestle (commented):
Looks great! 💛
I think with these numbers we can merge the change with confidence. Looks like the PHP devs came a long way with performance optimization of the array functions.

tests/Toolkit/ATest.php
bastianallgeier: I just fixed it.

bastianallgeier (approved):

PR #4676, opened 2022-09-20
Update check for Kirby and plugins

lukasbestle:
## Update check for Kirby and plugins

Screenshot

Our new update check in the enhanced system view makes it easy to stay informed about the version and security status of Kirby and your installed plugins. While feature updates are not always needed for finished sites, keeping an eye on security issues and important security messages is really important to keep your sites secure and healthy. We already publish all relevant information on our [security page](https://getkirby.com/security). The new update check brings this information and more right into the Panel so you can get a quick overview of the status of your particular site.

We designed the feature to work in your interest: It always prefers the minimum possible security update over feature updates to make security updates as easy as possible. It never performs an auto-update but just displays the relevant information. It never transmits any personal or site data to our server but processes the necessary data directly on your own server. And if you prefer to keep up to date with updates in another way, you can fine-tune the behavior or disable the feature all-together:

```php
<?php

// global configuration
return [
  'updates' => true | 'security' | false
];

// separate configuration for Kirby and plugins
return [
  'updates.kirby'   => true | 'security' | false,
  'updates.plugins' => true | false,
];

// or even configuration by plugin
return [
  'updates.plugins' => [
    '*'                      => true | false,
    'superwoman/*'           => true | false,
    'superwoman/superplugin' => true | false
  ]
];
```

---

## Related links

https://kirby.nolt.io/111

## Testing information

To test different scenarios, please modify the version number in `kirby/composer.json`. Changes to the version number will be picked up immediately. After a reload of the system view you can verify if the output matches what you expect.

Automated tests are still missing for parts of the internal update check logic. I will add complete unit and integration tests for all the logic until the 3.8 release. My goal is to write detailed integration tests with lots of different raw inputs and their expected outputs, including edge cases.

## Open questions

- [x] I'm not 100 % happy with the `TableUpdateStatusCell.vue` component yet. Its logic works well, but it could use better styling. I'm mostly not sure about the link styling: The underline is consistent with `UrlFieldPreview`, but the visual semantics differ from the Kirby update status in the `k-stats`. In both cases the link leads to the release notes of the update (if an update is available) or of the current version (if no update is available), but the stat (Kirby) & table cell (plugin) always display the current version. So I'm wondering if we should make the whole table cell clickable. Then it's more clear that the link is about the version and its update, not just the current version.
- [ ] The plan is to use `https://assets.getkirby.com` (KeyCDN) to serve the update files. We need to verify that we can indeed use KeyCDN for this (bandwidth, cost, caching behavior...).

## Ready?

### For review team

  - `update` option
  - Privacy policy
  - Privacy guide


config/areas/system/views.php
```diff
@@ -21,7 +22,13 @@
 				[
 					'label' => I18n::translate('version'),
 					'value' => $kirby->version(),
-					'link'  => 'https://github.com/getkirby/kirby/releases/tag/' . $kirby->version()
+					'info'  => $updateStatus?->label(),
+					'link'  => (
+						$updateStatus ?
+						$updateStatus->url() :
+						'https://github.com/getkirby/kirby/releases/tag/' . $kirby->version()
```
distantnative, line 28:
```php
$updateStatus?->url() ?? 'https://github.com/getkirby/kirby/releases/tag/' . $kirby->version()
```

config/areas/system/views.php
```diff
@@ -35,18 +42,24 @@
 
 			$plugins = $system->plugins()->values(function ($plugin) {
 				$authors = $plugin->authorsNames();
+				$version = (
+					$plugin->version() ?
+					$plugin->updateStatus()?->toArray() :
+					null
+				) ?? $plugin->version() ?? '–';
```
distantnative: Maybe it's too late today for me to understand this: But aren't we already checking in the parentheses for `$plugin->version()`? But if that exists we go for `$plugin->updateStatus()`. But if `$plugin->version()` return null, we jump to `null` in the parentheses just to fallback to `$plugin->version()` outside again?

i18n/translations/en.json
```diff
@@ -487,11 +487,23 @@
 
   "stats.empty": "No reports",
   "system.issues.content": "The content folder seems to be exposed",
+  "system.issues.eol.kirby": "Your installed Kirby version is end of life and will not receive further security updates",
```
distantnative:
```suggestion
  "system.issues.eol.kirby": "Your installed Kirby version has reached end-of-life and will not receive further security updates",
```

i18n/translations/en.json
```diff
@@ -487,11 +487,23 @@
...
+  "system.issues.eol.plugin": "Your installed version of the { plugin } plugin is end of life and will not receive further security updates",
```
distantnative:
```suggestion
  "system.issues.eol.plugin": "Your installed version of the { plugin } plugin is has reached end-of-life and will not receive further security updates",
```

panel/src/components/Layout/TableUpdateStatusCell.vue
```diff
@@ -0,0 +1,41 @@
...
+	computed: {
+		hasInfo() {
+			return typeof this.value !== "string";
+		},
+		url() {
+			return this.hasInfo ? this.value.url : null;
+		},
+		version() {
+			return this.hasInfo ? this.value.version : this.value;
+		}
```
distantnative: I think we can spare `hasInfo`. On the `v-if` we can directly check for `value.label`. And in the computed props we can directly return `this.value.url ?? null` and `this.value.version ?? this.value`.

panel/src/config/components.js
```diff
@@ -108,6 +108,7 @@ import Panel from "@/components/Layout/Panel.vue";
 import Stats from "@/components/Layout/Stats.vue";
 import Table from "@/components/Layout/Table.vue";
 import TableCell from "@/components/Layout/TableCell.vue";
+import TableUpdateStatusCell from "@/components/Layout/TableUpdateStatusCell.vue";
```
distantnative, line 111: This should probably rather go into `Forms/Previews` like other table cell previews.

src/Cms/Plugin.php
```diff
@@ -212,6 +214,53 @@ public function toArray(): array
 		];
 	}
 
+	/**
+	 * Returns the update status object unless the
+	 * update check has been disabled for the plugin
+	 *
+	 * @param array|null $data Custom override for the getkirby.com update data
+	 */
+	public function updateStatus(array|null $data = null): UpdateStatus|null
```
distantnative, line 225: Wondering if `updates()` would be a more Kirby-like name and still fitting.

src/Cms/Plugin.php
```diff
@@ -212,6 +214,53 @@ public function toArray(): array
...
+		$option = $kirby->option('update.plugins');
+
+		// specific configuration per plugin
+		if (is_array($option) === true) {
+			// filter all option values by glob match
+			$option = array_filter(
+				$option,
+				fn ($value, $key) => fnmatch($key, $this->name()) === true,
+				ARRAY_FILTER_USE_BOTH
+			);
```
distantnative, line 240: We could use `A::filter` here

src/Cms/System/UpdateStatus.php
```diff
@@ -0,0 +1,680 @@
...
+	 * Returns all vulnerabilities the current version
+	 * is affected by unless no data is available
+	 */
+	public function vulnerabilities(): array|null
+	{
+		if (isset($this->vulnerabilities) === true) {
+			return $this->vulnerabilities;
+		}
+
+		if (!$this->data || !$this->currentVersion) {
```
distantnative:
```suggestion
		if ($this->data === null || $this->currentVersion === null) {
```

Let's be explicit in the if statements, the `!` is always hard to understand what it really is supposed to test for.

distantnative (commented):
Some first comments - I have given up to review more tonight, realising how complex the new class is. Will try to have a closer look with a fresh mind.

On the config options, I think if the set is 
```
'display' | 'display-security' | false,
```
we might want to simplify it to align it with more of our other options and do
```
true | 'security' | false
```

I do agree on the styling of that cell, but that's usually best left with @bastianallgeier.

src/Cms/System/UpdateStatus.php
lukasbestle: To be honest my goal with these checks was to be as lax as possible explicitly. If the value ends up to be falsy but not null for some reason, we still need to return early. Checking strictly for null would be a downgrade from that safety. We do have prop and return types everywhere, but this provides an extra level of safety.

I can still change it, but I think it's another one of these general decisions we need to make as a team.

config/areas/system/views.php
lukasbestle: This won't work because `$updateStatus->url()` can also return `null` (for "no URL"). We only need the fallback if the whole `$updateStatus` object is `null`.

config/areas/system/views.php
lukasbestle: You are right, that makes no sense. I'll double-check what I intended to do.

panel/src/components/Layout/TableUpdateStatusCell.vue
lukasbestle: Certainly an option. I'm just wondering if it is too clever. The `hasInfo` value makes it easy to see that the cell can have two modes (strings and a whole object).

panel/src/config/components.js
lukasbestle: It's not a field preview though, is it? `TableCell.vue` supports either table cell components or field preview components. I used the former here.

src/Cms/Plugin.php
lukasbestle: Thought about it as well. I find `updates()` a bit weird because there can only be a single update. But a singular `update()` sounds like a verb. So I decided on `updateStatus()` even though it's also not perfect.

lukasbestle:
Regarding the config options: My thought was to make the options extensible in the future, e.g. with `notify` to send an update notification to the admins or `autoupdate` (in case this is ever needed/requested). But maybe I'm planning ahead too far. I'd be fine with your suggestion as it's indeed a lot simpler.

src/Cms/System/UpdateStatus.php
distantnative: Agreed, should be a common decision we all stick too - I thought we had made this one with v3, but happy to revisit and renew or change it. 

My argument for making it explicit would be readability and by that making our code easier to understand for ourselves. For you, this might be a given as you wrote that code. But for us others on the team things like this make it harder to understand as it opens up a lot of options. Given that `data` can only be an array or null, and `currentVersion` only a string or null, I think the openness doesn't add a lot more functionality, just decreases readability. For me to understand this, I had to jump to the very top of the file to read the property types, just to jump back down again. Would be much easier if I could understand that code in place.

lukasbestle: IMO the possible types don't really matter here. It's just "if this value is not set, then we cannot continue and need to return early".

bastianallgeier:
A few quick first thoughts from me as well. 

- It already works really well. Nice job! 
- I agree with @distantnative's thought on the available config flags (true, false, security) I think this could still be extended later. 
- My first thought on the config option naming was, that it should be called `updates` instead of `update`
- The different text colors in the system view show one thing clearly: our colors our fucking ugly for text. Especially the blue and green. I think we need a different way to style this here. 
- I'm not very happy with the info second line in our Stats component. I wonder if this could be done more elegantly, but that's definitely my job.
- When security updates are activated, there's the text "No known vulnerabilities" next to a plugin. I think we should just show a green / ok status here. It feels weird in combination with plugins without a version number. What about them? Do they have any vulnerabilities? I think sometimes too much positive information on first sight is worse than having none. It could still appear in a tooltip for example.

lukasbestle:
I agree on the config flags and config option naming. I will change that.

Regarding styling and text colors: I'll leave this to you as I don't want to butcher the Panel design. :)

> When security updates are activated, there's the text "No known vulnerabilities" next to a plugin. I think we should just show a green / ok status here.

I think this is as expected. If you only want security updates to be displayed, then this is what you get. "No known vulnerabilities" means that an update is available but that it isn't necessary to fix vulnerabilities.

With the current getkirby.com setup we can't publish vulnerabilities for plugins, but we could easily add this later when we extend our hub for plugin devs. That's why the core already treats plugins the same as Kirby regarding the update logic.

> It feels weird in combination with plugins without a version number. What about them? Do they have any vulnerabilities?

Without a version number we cannot do any update checks. This is why the update check is only enabled if the plugin has a version number.

> It could still appear in a tooltip for example.

Just the "No known vulnerabilities" text or every status value? TBH I wouldn't differentiate between the different status values. It wouldn't be consistent with the Kirby update status in the stats widget.

distantnative:
"No known vulnerabilities" just feels like a lot of positive endorsement for plugins where we have no clue about their vulnerability status.

lukasbestle:
If vulnerabilities were known, we would add them to the plugin.json on getkirby.com. If nothing is in there, then the core does know for sure that there are no known vulnerabilities.

lukasbestle:
Idea for plugins without version number: What if we display the "Could not check for updates" text in this case? Then the only plugins without a status text would be those where the config option was explicitly set to `false`.

distantnative:
I know that your are technically right. Nevertheless, we also know how users perceive "No known vulnerabilities". And it's not the technical understanding - "the Kirby team does not know about any vulnerabilities with this plugin" - but more a general one: there are no problems, this plugin is great. And given that we have not even a good setup and mechanism with plugin developers to handle security issues, I don't think it's a good UX, maybe even dangerous - even though the message is technically correct.

lukasbestle:
I see what you mean.

Idea: We could give the plugin.json a new field `securitySupport: true | false`. The default for now would be `false`. Only if the dev explicitly opts in (once we have the hub), we would set it to `true`.

In mode `false`, the core would display a message like "Security status unknown".

distantnative:
That would be an option. But I feel we are getting way ahead of ourselves. For the moment, I think the best experience for plugins (core is different) would be to only show a message when we know of an issue. If we don't know a thing (which is the case for 99% right now), no statement.

lukasbestle:
Sounds good. So basically we hardcode the "Security status unknown" message for plugins while the core stays "No known vulnerabilities", right?

lukasbestle:
If I think about it, maybe we should just not support the security-only mode for plugins at all until we have the hub. That would make it all a lot simpler. So plugin updates can only be configured true or false.

distantnative:
Tbh I think for me personally the best experience would be if nothing is shown when no action is suggested - so up-to-date and no-vuln would just not show any message. And only if one wants to inform about an update available or a security issue to patch, then show a message.

lukasbestle:
I feel like it could be confusing because the user won't know if the update check is enabled and working. It makes the status quite uncertain to the user.

config/areas/system/views.php
lukasbestle: @bastianallgeier I know your rewrite is not done yet, just before I forget about it: The general fallback to `$plugin->version()` is needed when the update check is disabled for one plugin or all plugins. In this case the `$plugin->updateStatus()` is `null`, so the Vue component can only get the version number as string. In your current version, support for the fallback is completely removed.

lukasbestle:
@bastianallgeier Quick comment on the version dropdown: The `url` is intended as a release notes URL, not specifically a download URL. Release notes often lead to the download, but I feel like a text like "Version information" and the icon `open` would better there.

Also the `targetVersion` is not necessarily the latest version. If there is a paid upgrade or a security issue, the `targetVersion` will be older.

config/areas/system/views.php
lukasbestle: @distantnative Reading my original code again, it made sense, but is confusing. If the version is there but `$plugin->updateStatus()` returns `null` (= update check disabled), it used `$plugin->version()` as fallback. This is the part that's missing now with the new code.

distantnative: @lukasbestle you know I'm not an `else` fan :P But in this case I think an if-else statement is in order. trying to make this a one liner really hinders understanding it.

lukasbestle: On the upside, the current code has the advantage that it displays the latest version even if there is no current version.

So I think the ideal way would be this:

```php
$version = $plugin->updateStatus()?->toArray() ?? $plugin->version() ?? '?';
```

And then of course the frontend component needs to deal with the string and only display a bubble without dropdown or something along those lines.

lukasbestle:
I've implemented what we discussed today. :)

src/Cms/System/UpdateStatus.php
```diff
@@ -0,0 +1,708 @@
...
+	public function toArray(): array
+	{
+		return [
+			'icon'          => $this->icon(),
+			'label'         => $this->label(),
+			'name'          => $this->pluginName,
+			'theme'         => $this->theme(),
+			'url'           => $this->url(),
+			'version'       => $this->currentVersion ?? '?',
+			'targetVersion' => $this->targetVersion() ?? $this->currentVersion ?? '?'
```
lukasbestle: @bastianallgeier I guess you added this fallback to the current version to display the latest version even if the version is already up to date, right?

There is also the case of an unreleased version (current version is higher than the latest version), in this case the fallback will break. If we want to display the latest version, we can get it with `$this->data['latest'] ?? null` more reliably.

Suggestion: What if the dropdown displays three versions like this (now that we have the room):

Screenshot

The data can come from here:

```suggestion
			'currentVersion' => $this->currentVersion ?? '?',
			'latestVersion'  => $this->data['latest'] ?? '?',
			'targetVersion'  => $this->targetVersion()
```

`targetVersion` explicitly has *no* fallback. If it's null, the dropdown would hide that "Suggested version" line entirely and only display current and latest version.

lukasbestle: Thinking about it, the suggested version is never needed as it's already contained in the status label. So I think current version and latest version are enough:

```suggestion
			'currentVersion' => $this->currentVersion ?? '?',
			'latestVersion'  => $this->data['latest'] ?? '?',
```

lukasbestle:
I think the `TableUpdateStatusCell` is now at a point where its logic covers all possible cases the backend could throw at it. I think it would be best if @bastianallgeier takes over from here again for styling finalization and cleanup.

src/Cms/Plugin.php
distantnative: I was thinking again about it, but is it really a single update when I'm on version 3.7.1 and the system gives me info on updating to 3.7.5 - or is that four updates?

lukasbestle: The class only ever suggests a single update. It is the user's responsibility to decide whether intermediate steps are useful and/or necessary.

distantnative: It suggests a version number, right? My point was that this would/could suggest multiple updates - which would make the name not that strange at all.

lukasbestle: Suggesting multiple updates would require a rewrite because there can only be one return value for each of the methods. Also I feel like multiple updates would be too overwhelming (e.g. if there is a security update, the user shouldn't need to care that there is also a feature update and an upgrade available – until that security update is done).

distantnative: Again, I'm not talking about changing anything but the name. Returning an info about 3.7.5 when you're on 3.7.1 is info about updates, not just one. And since I think the current name is a bit cumbersome, my only argument here is to use `Updates` and that it does make sense. Not to change anything how it works or what is returned.

distantnative: But if you like the current name that much, fine, we can stick with it.

lukasbestle: I don't like either name 100 %. :)

lukasbestle:
The unit tests currently fail. I'll fix them and add more unit tests when I come home.

bastianallgeier:
I've added my design changes. Anything else we need to have a look at for tomorrow?

lukasbestle:
Everything should be ready now. PHPUnit tests (unit and integration tests) for `Plugin::updateStatus()` and the `UpdateStatus` class are missing. I will add those in the coming days in a separate PR.

src/Cms/Plugin.php
bastianallgeier: Let's stay with updateStatus for now.

bastianallgeier (dismissed):

bastianallgeier:
@distantnative @afbora do you want to have another look at it before we merge?

afbora:
Let me try to few test please

distantnative:
I won't get to it today, so please just wait for @afbora and I will test more during RC2

afbora:
I loved PR ❤️ It's nice to see Kirby become more systematic.
I've few points about PR. I think these are not critical. I'll more test while RC process.

1. **Private repository behavior**

screen-capture-1457-System - Mægazine-localhost

**Expectation**

- Latest version: Unknown (or something like that)
- Update status: Could not check for updates

2. **HTTP error 404 issue**

I'm getting following error with following setup

> Could not load update data for plugin jg/fields-block: HTTP error 404
> Could not load update data for plugin test/test: HTTP error 404

**Tested plugins**

screen-capture-1458-System - Mægazine-localhost

````php
// test plugin
Kirby::plugin('test/test', []);
````

lukasbestle:
Thanks for testing! 💛

Number 1 seems to be a bug in the plugin model of getkirby.com. If the latest version is not known, it should set the value to `null` instead of `false`. Then the Panel UI should work as expected.

Number 2 is as expected IMO. The errors only occur in debug mode. Otherwise the Panel will just display "Could not check for updates" without throwing an exception. The first plugin probably needs to be published to getkirby.com or has a different name there. For the second example (custom, unpublished plugin), one should disable the update check via the Kirby config.

lukasbestle:
I've created a PR for `jongacnik/fields-block`: https://github.com/jongacnik/kirby-fields-block/pull/11

lukasbestle:
I'll fix getkirby.com tonight for the private plugin repo.

afbora:
> Number 2 is as expected IMO. The errors only occur in debug mode. Otherwise the Panel will just display "Could not check for updates" without throwing an exception. The first plugin probably needs to be published to getkirby.com or has a different name there. For the second example (custom, unpublished plugin), one should disable the update check via the Kirby config.

@lukasbestle I don't fully agree. Because it is not a critical and useful error. For example, it throws an exception when it is not compatible with Kirby version and this is correct behavior. But I think it is not the right behavior to show an error screen because it just cannot find the version information. Because the lack of version information is of no use to the developer or user. It is sufficient to have an error log in the console.

screen-capture-1459-Error - Mægazine-localhost

lukasbestle:
Console is a great idea! We could solve this by catching the exception and converting it to a prop for the system view. This makes the code of the system view a bit more complex, but it's certainly useful. I'll see what I can do.

bastianallgeier:
@lukasbestle @afbora do you think this is something that needs to be ready for rc.2? Could be a good task for rc.3. Or would it lead to too many complaints?

lukasbestle:
The error should only be displayed once. After another reload it's gone for three days. I think we can already ship it with RC.2 if you want to publish it today.

afbora:
> The error should only be displayed once. After another reload it's gone for three days. 

Yes, exactly.

> do you think this is something that needs to be ready for rc.2?

Lukas knows better if everything is ready but I would add it to RC2.

bastianallgeier:
I think there's no need to rush it today. But I think it would be nice to get it ready tomorrow. If you manage to already come up with something, great! If not, that's fine too.

lukasbestle:
I've changed the `UpdateStatus` class so that it no longer throws exceptions but instead collects them in an instance prop. This instance prop is then passed to the frontend as a view prop and printed to console there.

**Important:** When testing, please make sure to clear the cache before as the cache format on error has changed from the previous implementation (string instead of `false`).

lukasbestle:
Now working on the `plugin.json` output on getkirby.com for private repos.

lukasbestle:
@afbora Private repos now look like this (tested with an empty fake plugin):

Bildschirmfoto 2022-09-27 um 22 06 58

Would be great if you could test it with the real one.

afbora:
@lukasbestle I saw the console errors but private repo plugin still same

screen-capture-1461-System - Mægazine-localhost

afbora:
Not important a little suggestion 💡  
I suggest using `N/A` instead of `?`. It looks better semantically and visually. Of course you know better. What do you think?

afbora:
> I suggest using `N/A` instead of `?`. It looks better semantically and visually. Of course you know better. What do you think?

Nevermind. This will create translation problem.

lukasbestle:
> @lukasbestle I saw the console errors but private repo plugin still same

Have you cleared the cache? Looks like your installation still uses the old data from getkirby.com.

afbora:
Yes, all of caches.

lukasbestle:
If you open the cache file for the plugin, does it contain the value `"latest": false`? Maybe something is stuck in the CDN cache in your CDN region. Depending on when you tested, this may be normal (CDN caches for half an hour).

afbora:
> If you open the cache file for the plugin, does it contain the value `"latest": false`?

Yes.

> Maybe something is stuck in the CDN cache in your CDN region.

Yes, maybe. Because the download link is still old (I updated the download links yesterday)

lukasbestle:
If I request https://assets.getkirby.com/plugins/owebstudio/clear-cache.json, I don't get the update check data (including the `latest` key) as expected, just the general plugin data. What do you get?

afbora:
### getkirby plugin data

https://assets.getkirby.com/plugins/owebstudio/clear-cache.json

````json
{
    "title": "Clear Cache",
    "url": "https://getkirby.com/plugins/owebstudio/clear-cache",
    "author": {
        "name": "oweb studio",
        "url": ""
    },
    "repository": null,
    "download": "https://owebstudio.com/kirby/plugins/clear-cache",
    "category": "Performance",
    "description": "Clean temporary files and folders such as cache, media, .lock and .logins on your website.",
    "screenshot": "https://assets.getkirby.com/media/pages/plugins/owebstudio/clear-cache/b09a61a342-1663236193/screenshot.png"
}
````


### Local cache data
````json
{
  "created": 1664347957,
  "minutes": 4320,
  "value": {
    "title": "Clear Cache",
    "url": "https:\/\/getkirby.com\/plugins\/owebstudio\/clear-cache",
    "author": {
      "name": "oweb studio",
      "url": ""
    },
    "repository": null,
    "download": "https:\/\/owebstudio.lemonsqueezy.com\/checkout\/buy\/aa39efcb-c69b-430e-a3e7-a08900a81006",
    "category": "Performance",
    "description": "Clean temporary files and folders such as cache, media, .lock and .logins on your website.",
    "screenshot": "https:\/\/assets.getkirby.com\/media\/pages\/plugins\/owebstudio\/clear-cache\/b09a61a342-1663236193\/screenshot.png",
    "latest": false,
    "versions": {
      "0": {
        "description": "Latest release",
        "status": "latest"
      },
      "*": {
        "description": "Actively supported",
        "latest": false,
        "status": "active-support"
      }
    },
    "urls": {
      "*": {
        "changes": null,
        "download": "https:\/\/owebstudio.lemonsqueezy.com\/checkout\/buy\/aa39efcb-c69b-430e-a3e7-a08900a81006",
        "upgrade": null
      }
    },
    "incidents": [
      
    ],
    "messages": [
      
    ],
    "_version": "1.0.2"
  }
}
````

afbora (approved):
Looks amazing ❤️

bastianallgeier:
Yay, I'm so happy that this is finally approved and merged

PR #4685, opened 2022-09-22
Fixes object field related console errors

### LINKED ISSUE #4681: [3.8.0-rc.1] new object field some console errors
Sorry for overlooking the errors in the console before the PR merged.

### Few issues 

1. When first click add and open the drawer

> Uncaught TypeError: Cannot read properties of undefined (reading 'isOpen')
> Uncaught TypeError: Cannot read properties of undefined (reading 'open')

2. When click `tags` field

> Uncaught TypeError: Cannot read properties of undefined (reading 'focus')

### Sample blueprint

````yaml
fields:      
  person:
    label: Person
    type: object
    fields:
      category:
        label: Category
        type: multiselect
        options:
          design: Design
          architecture: Architecture
          photography: Photography
          3d: 3D
          web: Web
      name:
        type: text
      pages:
        type: pages
      files:
        type: files
      tags:
        type: tags
````

afbora:
Fixes #4681 

I'm not 100% sure this is the solution. Because it contains specific changes to solve only related problem.

### DIFF
--- a/panel/src/components/Drawers/FormDrawer.vue
+++ b/panel/src/components/Drawers/FormDrawer.vue
@@ -85,11 +85,9 @@ export default {
 				)?.name;
 			}
 
-			this.$nextTick(() => {
-				this.$nextTick(() => {
-					this.focus(focus);
-				});
-			});
+			setTimeout(() => {
+				this.focus(focus);
+			}, 10);
 		}
 	}
 };
--- a/panel/src/components/Drawers/FormDrawer.vue
+++ b/panel/src/components/Drawers/FormDrawer.vue
@@ -80,7 +80,7 @@ export default {
 			this.tab = tab || this.firstTab.name;
 
 			if (focus !== false) {
-				setTimeout(() => {
+				this.$nextTick(() => {
 					if (focus === true) {
 						focus = Object.values(this.fields).find(
 							(field) => field.autofocus === true
@@ -88,7 +88,7 @@ export default {
 					}
 
 					this.focus(focus);
-				}, 1);
+				});
 			}
 		}
 	}
--- a/panel/src/components/Drawers/FormDrawer.vue
+++ b/panel/src/components/Drawers/FormDrawer.vue
@@ -79,17 +79,17 @@ export default {
 			this.$refs.drawer.open();
 			this.tab = tab || this.firstTab.name;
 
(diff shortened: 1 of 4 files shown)


bastianallgeier:
This PR breaks the autofocus feature in the drawer.

bastianallgeier:
@afbora I fixed it a bit differently. I think the setTimeout is the root cause of the issue, but one nextTick isn't enough. I reverted your changes to the MultiSelect Input because they aren't needed if the field is created correctly. Could you check if it works for you with this fix?

afbora:
@bastianallgeier Same..

screen-capture-1456-Sandbox - Kirby Demo-localhost

Also console errors

> TypeError: Cannot read properties of undefined (reading 'open')
> Uncaught TypeError: Cannot read properties of undefined (reading 'querySelector')

bastianallgeier:
I don't get it. How can this be so different between Win and Mac? Sorry for force-pushing over your PR? Do you still have the local branch? Could you force-push again to overwrite my stuff?

afbora:
Unfortunately not. Because I updated my local to test it 😅

bastianallgeier:
Yikes, so sorry! I will try to reconstruct it.

bastianallgeier:
I've reimplemented the changes. Could you please have a look at them if I missed anything. I also simplified the focus code again and increased the timeout. This should make the focus method more reliable when it relies on elements being there.

afbora:
Seems only `tags` field focus issue left (when click tags preview on object field)

> Uncaught TypeError: Cannot read properties of undefined (reading 'focus')

afbora:
While there is no such issue in the structure or blocks field, I still could not understand why we had this error in the object field.

bastianallgeier:
@afbora could you take care of the tags focus issue. I'm not entirely sure what you mean? I also have absolutely no idea why it happens here.

afbora:
@bastianallgeier I've pushed a commit that fix the tags field issue but I'm still not sure about the solution. I don't understand why it gives this error only in object field 🤷‍♂️

bastianallgeier (approved):

bastianallgeier:
@afbora I think we will meet this again when we move the structure field to the drawer. The changes right now seem pretty solid and I would go for it and wait what else turns up later.

PR #4691, opened 2022-09-23
Fix panel favicon for dark mode

afbora:
## This PR …

The SVG had to be loaded last (I think) for the panel to show the correct favicon in dark mode.

### Enhancements
- Now the correct favicon for the panel is shown in dark mode

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Panel/Document.php
+++ b/src/Panel/Document.php
@@ -149,11 +149,12 @@ class Document
 	}
 
 	/**
-	 * Returns array of favion icons
+	 * Returns array of favicon icons
 	 * based on config option
 	 * @since 3.7.0
 	 *
 	 * @param string $url URL prefix for default icons
+	 * @throws \Kirby\Exception\InvalidArgumentException
 	 */
 	public static function favicon(string $url = ''): array
 	{
@@ -163,13 +164,13 @@ class Document
 				'type' => 'image/png',
 				'url'  => $url . '/apple-touch-icon.png',
 			],
-			'shortcut icon' => [
-				'type' => 'image/svg+xml',
-				'url'  => $url . '/favicon.svg',
-			],
 			'alternate icon' => [
 				'type' => 'image/png',
 				'url'  => $url . '/favicon.png',
+			],
+			'shortcut icon' => [
+				'type' => 'image/svg+xml',
+				'url'  => $url . '/favicon.svg',
 			]
 		]);
 


bastianallgeier:
Is this actually broken somewhere? It already worked in dark mode so far in Firefox and Chrome for me.

afbora:
Ohh really? It works for me only when I change the order of SVG and PNG. I'll track the issue a little more.

afbora:
Also works with `'sizes' => 'any',` for `ico` and `png` but not for `svg`. I'm still not sure what exactly the solution. I'm sure that it hasn't worked for me for a long time.

https://css-tricks.com/favicons-how-to-make-sure-browsers-only-download-the-svg-version/

````php
$icons = $kirby->option('panel.favicon', [
    'apple-touch-icon' => [
        'type'  => 'image/png',
        'url'   => $url . '/apple-touch-icon.png',
        'sizes' => 'any',
    ],
    'shortcut icon' => [
        'type' => 'image/svg+xml',
        'url'  => $url . '/favicon.svg',
    ],
    'alternate icon' => [
        'type'  => 'image/png',
        'url'   => $url . '/favicon.png',
        'sizes' => 'any',
    ]
]);
````

**Comment by afbora** [comment]
@bastianallgeier Please could you check out few articles about that?

https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types#icon

> If there are multiple <link rel="icon">s, the browser uses their [media](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-media), [type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-type), and [sizes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-sizes) attributes to select the most appropriate icon. If several icons are equally appropriate, the last one is used. If the most appropriate icon is later found to be inappropriate, for example because it uses an unsupported format, the browser proceeds to the next-most appropriate, and so on.

An article about that 
https://medium.com/web-dev-survey-from-kyoto/favicon-nightmare-how-to-maintain-sanity-7628bfc39918

**Comment by bastianallgeier** [comment]
If this is expected behaviour and it fixes your problem, I'm fine with it!

**Review by bastianallgeier — APPROVED** [review]

**Comment by afbora** [comment]
Thanks but still it would be nice to get @lukasbestle and @distantnative's thoughts too so that there is no regression.

**Comment by lukasbestle** [comment]
I haven't noticed this issue so far (using light mode on my Mac), but your sources seem plausible.

**Comment by jakobgrommas** [comment]
Since SVG icons are [not supported](https://caniuse.com/link-icon-svg) on Safari, it could be a nice feature to add dark mode option for ICO or PNG favicons like:
````php
    'favicon' => [
      'apple-touch-icon' => [
        'type' => 'image/png',
        'url'  =>  'assets/images/light/apple-touch-icon.png',
        'sizes' => 'any'
      ],
      'shortcut icon' => [
        'type' => 'image/svg+xml',
        'url'  => 'assets/images/color/favicon.svg'
      ],
      'alternate icon' => [
        'type' => 'image/png',
        'url'  => 'assets/images/light/favicon.png',
        'sizes' => 'any'
      ],
      'apple-touch-icon' => [
        'type' => 'image/png',
        'url'  =>  'assets/images/dark/apple-touch-icon.png',
        'sizes' => 'any',
        'prefers-color-scheme' => 'dark'
      ],
      'alternate icon' => [
        'type' => 'image/png',
        'url'  => 'assets/images/dark/favicon.png',
        'sizes' => 'any',
        'prefers-color-scheme' => 'dark'
      ]
    ]
  ````

**Comment by afbora** [comment]
We can't use same `rel` multiple times with this syntax. We need new `rel` array key for this:

````php
[
  [
    'rel'  => 'apple-touch-icon',
    'type' => 'image/png',
    'url'  =>  'apple-touch-icon.png',
  ]
]
````

Of course, it would be good to hear your team's thoughts on the issue first @distantnative @lukasbestle

**Comment by jakobgrommas** [comment]
Makes totally sense, thanks for your improved suggestion!

**Comment by lukasbestle** [comment]
I've not seen the `prefers-color-scheme` attribute used with the `link` tag so far. Does this really work?

**Comment by jakobgrommas** [comment]
AFAIK it works, as you can see in this example:

Bildschirmfoto 2023-08-22 um 13 05 20

Frontend on left tab with, backend on right tab without `prefers-color-scheme`, tested in Safari Version 16.6 (18615.3.12.11.2).

**Comment by lukasbestle** [comment]
Ah, the `media` attribute with `prefers-color-scheme` as query. Yeah, that will work.

@afbora If you have time, could you prepare a v4 PR with an implementation like you suggested? I think it would make sense to use the attributes 1:1 to avoid confusion, so e.g.:

```php
[
  [
    'rel'  => 'apple-touch-icon',
    'type' => 'image/png',
    'url'  =>  'apple-touch-icon-dark.png',
    'media' => '(prefers-color-scheme: dark)',
  ]
]
```

afbora:
@lukasbestle I've tried in https://github.com/getkirby/kirby/pull/5657, I hope it's acceptable.

Issue #4703, opened 2022-09-26
[3.8.0-rc.1] Picker issue for those which don't have uuid yet


afbora:
If you select a page that does not yet have a uuid with the pages field, it disappears when the page is refreshed. This is a possible bug for those who upgraded to 3.8.

> @lukasbestle  The UUID should have been generated once the data for the pages field was generated. If this does not happen, then there's a bug.

bastianallgeier:
✅

PR #4706, opened 2022-09-27
Populate the UUID cache when a UUID is printed

bastianallgeier:
## This PR …

When a UUID is converted to string, the UUID is generated if it is missing. But so far, the UUID wasn't added to the cache. This means that it couldn't be found afterwards, which broke the pages picker. This PR would fix that by also populating the cache in the toString method. @distantnative needs to check if that would have any unwanted side effects. 

### Fixes

- #4703

### Breaking changes

- None

## Ready?

### For review team

### DIFF
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -303,6 +303,25 @@ trait FileActions
 	}
 
 	/**
+	 * Stores the content on disk
+	 *
+	 * @internal
+	 * @param array|null $data
+	 * @param string|null $languageCode
+	 * @param bool $overwrite
+	 * @return static
+	 */
+	public function save(array $data = null, string $languageCode = null, bool $overwrite = false)
+	{
+		$file = parent::save($data, $languageCode, $overwrite);
+
+		// update model in siblings collection
+		$file->parent()->files()->set($file->id(), $file);
+
+		return $file;
+	}
+
+	/**
 	 * Remove all public versions of this file
 	 *
 	 * @return $this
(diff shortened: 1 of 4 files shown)


distantnative:
Hmmm one can argue when to populate the cache but a UUID should still be found even if not in the cache. If that doesn't happen, something else is broken and we shouldn't it just cover up by populating the cache earlier.

distantnative:
Not saying that we shouldn't do this PR, but I think it's more a cover-up than a fix of the problem.

My guess is that the updated page (now with UUID) isn't updated in the relevant collections right away, so that's why it can't be found.

distantnative:
I think that's the downside because we aren't using ->update()

bastianallgeier:
Ahh, that makes sense. But we already moved the collection updates to the save method for that reason. I wonder why it still cannot be found.

distantnative:
I think this PR still makes sense. But I'm afraid the issue pops up elsewhere maybe.

distantnative:
@bastianallgeier looking at `::save` and `::updateParentCollections` the logic looks very different, much more basic in `::save`. Could it be that? Why are we trying to implement the same in two places? Shouldn't we use the common `::updateParentCollections`?

bastianallgeier:
@distantnative you are completely right. The updateParentCollections should be used. I refactored it and unfortunately it doesn't solve the problem by itself. The refactoring is still useful here and I think the combination of both solutions is stable enough to consider it done.

distantnative (approved):

PR #4710, opened 2022-09-27
Fix `Uuid::is()`

### LINKED ISSUE #4709: [3.8.0-rc.1] False positives in UUID validator
## Description
The following all return `true` despite the string not being a valid Kirby UUID:
```
`Uuid::is('page://')`
`Uuid::is('page://', 'page')`
`Uuid::is('file://')`
`Uuid::is('file://', 'file')`
// etc.
```

**Expected behavior**  

The validation should only return `true` if the string resembles the UUID format both before and (where applicable, i.e. not for `site`) after the slashes

**Additional context**  

Two apparent reasons:
- when a `$type` variable is present, the part after the slashes is not being checked for in `return Str::startsWith($string, $type . '://')`
- when only a string is given, the regex does not seem to check for the presence of at least one character (not very practical, but that would probably manifest the minimum requirement for a UUID with a custom generator?)

distantnative:
Fixes #4709

@sebastiangreger would be great if you could give this a try 🙏

### DIFF
--- a/src/Uuid/Uuid.php
+++ b/src/Uuid/Uuid.php
@@ -228,13 +228,18 @@ class Uuid
 		string $string,
 		string|null $type = null
 	): bool {
-		if ($type !== null) {
-			return Str::startsWith($string, $type . '://');
+		$type  ??= implode('|', Uri::$schemes);
+		$pattern = sprintf('!^(%s)://([a-zA-Z0-9]*)!', $type);
+
+		if (preg_match($pattern, $string, $matches) !== 1) {
+			return false;
 		}
 
-		// try to match any of the supported schemes
-		$pattern = sprintf('!^(%s)://!', implode('|', Uri::$schemes));
-		return preg_match($pattern, $string) === 1;
+		if ($matches[1] === 'site') {
+			return strlen($matches[2]) === 0;
+		}
+
+		return strlen($matches[2]) > 0;
 	}
 
 	/**
--- a/src/Uuid/Uuid.php
+++ b/src/Uuid/Uuid.php
@@ -229,7 +229,7 @@ class Uuid
 		string|null $type = null
 	): bool {
 		$type  ??= implode('|', Uri::$schemes);
-		$pattern = sprintf('!^(%s)://([a-zA-Z0-9]*)!', $type);
+		$pattern = sprintf('!^(%s)://(.*)!', $type);
 
 		if (preg_match($pattern, $string, $matches) !== 1) {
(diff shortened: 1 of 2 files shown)


src/Uuid/Uuid.php
```diff
@@ -228,13 +228,18 @@ final public static function is(
 		string $string,
 		string|null $type = null
 	): bool {
-		if ($type !== null) {
-			return Str::startsWith($string, $type . '://');
+		$type  ??= implode('|', Uri::$schemes);
+		$pattern = sprintf('!^(%s)://([a-zA-Z0-9]*)!', $type);
```
lukasbestle: Custom generators could use other character sets:

```suggestion
		$pattern = sprintf('!^(%s)://(.*)!', $type);
```

distantnative: But wouldn't those clash with our caches - so we should enforce this set?

lukasbestle: The file cache will still work, it will just not be as pretty with the hash. Other caches should be completely unaffected.

distantnative: Didn't we just had the talk that the file cache especially doesn't work with e.g. underscores https://github.com/getkirby/kirby/blob/main/src/Cache/FileCache.php#L95-L98?

distantnative: Ah I see, what you mean with the hash.... but why then did we threw out underscores from our UUID generator?

lukasbestle: Because we didn't want that hash by default for the 99% use case.

sebastiangreger: Tested it @distantnative – the original edit indeed resolves the issue when sticking to the core generator.

As for the 1% use case, it might be worth considering at least all possible ID formats and their character sets that were offered by AutoID out of the box, as I imagine those to be the most likely cases to keep their old ID format and implement a custom generator when migrating?
- IncrementingGenerator `[0-9]`
- NanoGenerator `[A-Za-z0-9_-]{21}`
- TokenGenerator `[a-zA-Z0-9]{8}`
- UUIDGenerator `[a-f0-9-]{32}`

In other words: in favour of the suggested edit. The only way to truly solve this would be to make the regex configurable when implementing a custom generator, but that may be overkill and add unnecessary complexity. The baseline definition of a Kirby UUID is `{type}://{something}`. I keep using my own custom format, and will simply implement my own verification method alongside my generator.

lukasbestle (approved):

PR #4711, opened 2022-09-27
Fix `->findBy()` for UUID object

### LINKED ISSUE #4704: [3.8.0-rc.1] Reverse relation with UUID via `findBy` not working
## Description

- Given I have two content types: `event` and `post`
- Where `post` has a field `relatedevent`, which stores an `event` UUID
- When getting `post => event` relation via `find('page://event-id')` works OK
- But, the other way around `event => post` via `findBy` doesn't work when just using `uuid()`
  - This fails: `$this->kirby()->collection('blog/posts')->findBy('relatedevent', $this->uuid())`
  - This works: `$this->kirby()->collection('blog/posts')->findBy('relatedevent', $this->uuid()->toString())`

**Expected behavior**  

- It should work with just `$this->uuid()`, without the need to cast to string.

## Your setup

**Kirby Version** : `3.8.0-rc.1`

distantnative:
Fixes #4704 (see for example)

### DIFF
--- a/src/Cms/Collection.php
+++ b/src/Cms/Collection.php
@@ -133,7 +133,8 @@ class Collection extends BaseCollection
 	 */
 	public function findBy(string $attribute, $value)
 	{
-		// $value: cast UUID object to string
+		// $value: cast UUID object to string to allow uses
+		// like `$pages->findBy('related', $page->uuid())`
 		if ($value instanceof Uuid) {
 			$value = $value->toString();
 		}
--- a/src/Cms/Collection.php
+++ b/src/Cms/Collection.php
@@ -6,6 +6,7 @@ use Closure;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Toolkit\Collection as BaseCollection;
 use Kirby\Toolkit\Str;
+use Kirby\Uuid\Uuid;
 
 /**
  * The Collection class serves as foundation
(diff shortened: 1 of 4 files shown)


src/Toolkit/Collection.php
```diff
@@ -412,6 +412,11 @@ public function find(...$keys)
 	 */
 	public function findBy(string $attribute, $value)
 	{
+		// $value: cast object to string, e.g. UUID object
+		if (is_object($value) === true) {
```
bastianallgeier: As Lukas posted in Discord, this one would be less prone to unexpected regressions. 

```suggestion
		if (is_a($value, 'Kirby\Uuid\Uuid') === true) {
```

distantnative: But then we should do that in Cms\Collection. And we need to get used to `instanceof`, not `is_a`.

src/Cms/Collection.php
```diff
@@ -123,6 +124,23 @@ public function append(...$args)
...
+	/**
+	 * Find a single element by an attribute and its value
+	 *
+	 * @param string $attribute
+	 * @param mixed $value
+	 * @return mixed|null
+	 */
+	public function findBy(string $attribute, $value)
+	{
+		// $value: cast UUID object to string
```
lukasbestle:
```suggestion
		// $value: cast UUID object to string to allow uses
		// like `$pages->findBy('related', $page->uuid())`
```

lukasbestle (commented):
IMO there should be an additional unit test for this in `Cms\CollectionTest`. The test in `HasUuidsTest` is more an integration test.

distantnative:
👍 can't approve myself

bastianallgeier (approved):

PR #4712, opened 2022-09-27
UUID: fix changing page slug

### LINKED ISSUE #4696: [3.8.0-rc.1] Changing URL leads to duplicate folders because of UUID
> in a starterkit changing the URL for  notes to blog. The change actually works (renames folder to blog, new blog folder includes all old subpages), but indeed notes remains with default blueprint. Looks like because a UUID is created in the process

Screenshot 2022-09-24 at 14 15 56

distantnative:
Fixes #4696

### DIFF
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -107,6 +107,9 @@ trait PageActions
 				'root'    => null
 			]);
 
+			// clear UUID cache recursively (for children and files as well)
+			$oldPage->uuid()->clear(true);
+
 			if ($oldPage->exists() === true) {
 				// remove the lock of the old page
 				$oldPage->lock()?->remove();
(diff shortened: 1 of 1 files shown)


lukasbestle (commented):
There could be race condition if another request caches a UUID while the slug is changed/the page is deleted (i.e. between clearing the cache and actually modifying the content directory). It's pretty unlikely, but in a million requests it may happen.

I think it would still be safest to do the cache clearing after the change and only get the UUID object before. Of course only if it works reliably this way (including the recursive changes). Otherwise I'd be fine with these changes.

distantnative:
@lukasbestle could you test if your suggested way works?

lukasbestle (approved):
It doesn't work without more major changes unfortunately. If there are subpages that are moved/deleted as well, we'll have the same original issue for the subpages.

So I think this is a good fix for now. The action logic is probably not completely safe against race conditions right now and this one is not too critical (can be fixed with clearing the cache if it ever happens).

PR #4717, opened 2022-09-28
Bring back `DS`

distantnative:
- [x] Remove from breaking change list

### DIFF
--- a/config/setup.php
+++ b/config/setup.php
@@ -1,6 +1,12 @@
 <?php
 
 /**
+ * Constants
+ * @deprecated 3.8.0 Use `/` instead
+ */
+define('DS', '/');
+
+/**
  * Class aliases
  */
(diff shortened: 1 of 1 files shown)


lukasbestle (approved):

psterker:
Sadly, this will clash with other tools like PHPStan or Laravel, resulting in the following warning:

`Warning: Constant DS already defined in /my-project-path/kirby/config/setup.php on line 7` 

A simple `if (!defined('DS'))` check would do the trick.

lukasbestle:
Making the definition lazy (skipping it if it already exists) prevents the code from behaving deterministically.

However we could remove the constant in v4 entirely (given it was already deprecated in 3.8). What do you think @distantnative?

distantnative:
@lukasbestle yes, no objection from my side

lukasbestle:
For reference: #5590

PR #4718, opened 2022-09-28
Update check: hide button if no url

### LINKED ISSUE #4716: [3.8.0-rc.2]: Update check - don't show "version info" button when no info
Currently, the button gets only disabled. Which is rather confusing as one has only one dropdown open and can't quite realise the difference between active/disabled right away. Button should simply be hidden when not available.

Screenshot 2022-09-28 at 19 34 25

distantnative:
### Fixes
- https://github.com/getkirby/kirby/issues/4716

Screenshot 2022-09-28 at 19 53 45

Screenshot 2022-09-28 at 19 53 41

### DIFF
--- a/panel/src/components/Layout/TableUpdateStatusCell.vue
+++ b/panel/src/components/Layout/TableUpdateStatusCell.vue
@@ -32,11 +32,7 @@
 					</div>
 				</dl>
 
-				<k-dropdown-item
-					:disabled="value.url === null"
-					icon="open"
-					:link="value.url"
-				>
+				<k-dropdown-item v-if="value.url" icon="open" :link="value.url">
 					{{ $t("versionInformation") }}
 				</k-dropdown-item>
 			</k-dropdown-content>
@@ -85,8 +81,6 @@ export default {
 
 .k-plugin-info {
 	padding: 1rem;
-	border-bottom: 1px solid var(--color-gray-700);
-	margin-bottom: 0.5rem;
 	width: 20rem;
 }
 .k-plugin-info div {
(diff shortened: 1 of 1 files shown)


lukasbestle:
Not sure TBH. We disable buttons throughout the UI for any feature that is possible but currently not available.

A missing button is IMO more confusing than helpful. A user may think "wait a second, wasn't there a button recently? Why is it gone?" without realizing that other plugins may have that button. Hiding buttons is certainly a possibility, but it's confusing because we don't do it anywhere else.

distantnative:
I literally tried to click a disabled button for 15 seconds and thought Vue was broken, because I didn't at all realise its state could be disabled. A user does not have the background knowledge that these buttons can be presented enabled/disabled. Due to the nature of the dropdown a user also only sees one button at a time. Our disable state is not recognisable as disabled on top of a black background if you only see one disabled button with no comparison to an enabled styling.

We might want to solve this differently. But the current UI is a very broken UX.

lukasbestle:
Yeah, that makes sense. I'm not against the change, let's see what the others think.

afbora:
> I literally tried to click a disabled button for 15 seconds and thought Vue was broken, because I didn't at all realise its state could be disabled.

Same 🙈 I already thought of suggesting this but then gave up. Now I totally agree with @distantnative 👍

bastianallgeier:
I think it makes a lot of sense here to remove the button entirely. Related question but for a different PR: why don't we link to the 1.0.0 version in the example above?

bastianallgeier (approved):

lukasbestle:
> Related question but for a different PR: why don't we link to the 1.0.0 version in the example above?

@bastianallgeier Because we don't know the repo URL for that (no data on getkirby.com).

bastianallgeier:
Ah, that makes sense.

PR #4722, opened 2022-09-28
System view: improve console messages

### LINKED ISSUE #4721: [3.8.0-rc.2] System view: unclear console errors
The new console errors from the update check are very unclear currently:
Screenshot 2022-09-28 at 20 21 47

Knowing how our users have reacted to console errors, this will raise a lot of questions as the console output does not explain why these errors are happening or what the user can do to solve them. We probably should add an explanation as we did for the folder checks. And maybe not throw them as console errors but notices or info.

distantnative:
Fixes https://github.com/getkirby/kirby/issues/4721

Screenshot 2022-09-28 at 20 35 31


panel/src/components/Views/SystemView.vue
```diff
@@ -106,7 +103,18 @@ export default {
...
+		console.info(
+			`System health checks ended. ${
+				promises.length - this.accessible.length
+			} issues found.`
+		);
+
+		// print exceptions from the update check to console for debugging
+		if (this.exceptions.length > 0) {
+			console.info("----");
+			console.info("Running plugin version checks for the Panel system view:");
```
lukasbestle: Not just plugins, could also be Kirby itself:

```suggestion
			console.info("Running the update check for Kirby and plugins.");
```

lukasbestle:
I deliberately placed the update check logs at the top of the method because
1. the update check is already done when this code is run (so it is run before the folder checks in the execution order)
2. placing the logs after the `await` may delay them so much that they are not noticed during debugging. In most setups, the folder checks will return a HTTP response immediately, but they may also time out (which takes a while).

panel/src/components/Views/SystemView.vue
```diff
@@ -106,7 +103,18 @@ export default {
...
+		console.info(
+			`System health checks ended. ${
+				promises.length - this.accessible.length
+			} issues found.`
+		);
+
+		// print exceptions from the update check to console for debugging
+		if (this.exceptions.length > 0) {
+			console.info("----");
+			console.info("Running the update check for Kirby and plugins.");
```
lukasbestle: Given that the output is only logged when there are exceptions, maybe we should even reword the message to something like "The following errors occurred during the update check of Kirby and/or plugins:". Or alternatively the message should be printed always and then there should also be a final "Update check ended." message.

lukasbestle:
@distantnative I've pushed a suggestion:

Bildschirmfoto 2022-10-03 um 13 39 10

I decided for the output only on error because the Panel doesn't know from the list of exceptions if the update check is enabled at all. If the user disables the update check, a message like "Running update check" would be weird.

What do you think? We could make the "Running update check" message work with a few backend changes if you prefer it.

distantnative:
I'm fine with this.

lukasbestle (approved):

PR #4741, opened 2022-10-03
More efficient lock check

### LINKED ISSUE #4740: Query in tag field polled every 10s while page is open in panel
## Description
Tag fields (potentially other field types as well; untested) with queries execute that query every 10 seconds while the page in question is open in the panel. With slightly more complex queries (e.g. when using `->index()`) this can lead to a measurable increase in server load.

**Expected behavior**  
The query should not be executed repeatedly.

## To reproduce
1. Install a fresh Starterkit (3.7 is fine, as this is not a new issue)
2. Add this little plugin

```php
// site/plugins/my-plugin/index.php
<?php
Kirby::plugin('my/plugin', [
  'pageMethods' => [
    'alltags' => function () {
      F::write(kirby()->root('logs')  . '/alltags.log', date('H:i:s'). "\r\n", true);
      return ['tag1', 'tag2', 'tag3'];
    }
  ]
]);
```

3. Change the `tag` field in the `note` blueprint as follows:

```yaml
# site/blueprints/pages/note.yml
tags:
  options: query
  query: page.alltags

```

4. Open an existing note page in the panel
5. Wait a minute and open file `site/logs/alltags.log`
6. See how the field query got executed every 10 seconds:
```
07:54:08
07:54:18
07:54:28
07:54:38
07:54:48
07:54:58
```

## Your setup
Fresh Starterkit 3.7.5 (also present after updating to 3.8.0-rc.1)
 - OS: Linux Mint
 - Browser: Firefox
 - Version: 104.0.2

bastianallgeier:
## This PR …

When we switched to Fiber, we removed the GET route for locks and instead reloaded the view every 10 seconds and passed an only rule, to only return `$view.props.lock`. The only problem: To build the `$view` array, the backend would process a lot more than necessary. Even with the only rule, tabs, content, permissions and more would be reevaluated on every call every 10 seconds. This is not noticable if the view is fast, but when there are complex queries involved it can get really messy and the UUIDs don't make this problem better. 

That's why I brought back the GET request for the lock status and instead of using the Fiber call, the checks are now done through the API again. I've unified the code, which is used in the Fiber view and in the lock check. They now return identical objects or false if locking is disabled. This means that we can still get the first state from Fiber and every next update is done through the API. Ideally, we would have additional Fiber routes that could do the same thing without relying on the API at some point. But this seems like a rather simple fix for a pretty big issue. 

@distantnative I think you really should have a look at this. You are most familiar with locking. I hope I didn't miss anything.

### Fixes

- #4740

## Ready?

### For review team

### DIFF
--- a/src/Panel/Model.php
+++ b/src/Panel/Model.php
@@ -235,18 +235,7 @@ abstract class Model
 	public function lock(): array|false
 	{
 		if ($lock = $this->model->lock()) {
-			if ($lock->isUnlocked() === true) {
-				return ['state' => 'unlock'];
-			}
-
-			if ($lock->isLocked() === true) {
-				return [
-					'state' => 'lock',
-					'data'  => $lock->get()
-				];
-			}
-
-			return ['state' => null];
+			return $lock->toArray();
 		}
 
 		return false;
--- a/config/api/routes/lock.php
+++ b/config/api/routes/lock.php
@@ -5,21 +5,15 @@
  * Content Lock Routes
  */
 return [
-    [
-        'pattern' => '(:all)/lock',
-        'method'  => 'GET',
-        'action'  => function (string $path) {
-			if ($lock = $this->parent($path)->lock()) {
-                return [
-					'lock' => $lock->toArray()
-				];
-			}
-
-            return [
-                'lock' => false
-            ];
-        }
-    ],
+	[
+		'pattern' => '(:all)/lock',
+		'method'  => 'GET',
+		'action'  => function (string $path) {
+			return [
+				'lock' => $this->parent($path)->lock()?->toArray() ?? false
+			];
+		}
+	],
 	[
 		'pattern' => '(:all)/lock',
 		'method'  => 'PATCH',
--- a/config/api/routes/lock.php
+++ b/config/api/routes/lock.php
@@ -5,6 +5,21 @@
  * Content Lock Routes
  */
 return [
+    [
+        'pattern' => '(:all)/lock',
+        'method'  => 'GET',
+        'action'  => function (string $path) {
+			if ($lock = $this->parent($path)->lock()) {
+                return [
+					'lock' => $lock->toArray()
+				];
+			}
+
+            return [
+                'lock' => false
+            ];
+        }
+    ],
 	[
 		'pattern' => '(:all)/lock',
 		'method'  => 'PATCH',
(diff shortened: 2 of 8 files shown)


src/Panel/Model.php
```diff
@@ -235,18 +235,7 @@ public function isDisabledDropdownOption(string $action, array $options, array $
...
-					'state' => 'lock',
-					'data'  => $lock->get()
-				];
-			}
-
-			return ['state' => null];
+			return $lock->toArray();
 		}
 
 		return false;
```
distantnative, line 241: We could simplify this then even further

```php
return $this->model->lock()?->toArray() ?? false;
```

distantnative:
This would be a way and looks fine so far (would like to do some actual testing still though).
But I wonder whether the same thing (Fiber not respecting the `only`-nature of a reload request) would bite us in other parts too and we should rather fix that? 

To my understanding we would need to wrap a lot more into Closures in `Panel\Model::props()` etc.

bastianallgeier:
The general problem is indeed that we are not clean enough with the way we send data to the views. Other "only" requests would indeed have the same problem. But when I looked at it, it seems like the far more risky fix so shortly before the release.

distantnative:
Keep seeing the Fiber requests still
Screenshot 2022-10-03 at 12 56 55

bastianallgeier:
Did you run npm run dev to test? Haven't updated the dist files.

distantnative:
Yes, always just running on `npm run dev`. But trying to narrow it down, this seems weird.

distantnative:
Ok somehow vite didn't pick up the changes in dev mode. Now working.

bastianallgeier:
Now Codecov discovered that there are lock routes for the first time and we have to write tests for them 🙈

distantnative:
Codecov knew already about them, but our setup is that we are not allowed to leave any touched line uncovered even if it was not covered before. Very strict @lukasbestle

lukasbestle:
Codecov is only complaining about the single line in the new `GET` route. The others were already uncovered, but because they were not changed, there is no additional risk from this PR. As we discussed before, we can replace the tests in such cases with manual tests and extra carefulness if automated tests are difficult or not viable with the current code base.

bastianallgeier:
I've created a simple integration test for the get route.

distantnative (approved):

PR #4742, opened 2022-10-03
`OptionsQuery`: fix for array

### LINKED ISSUE #4739: [3.8.0-rc.2] Missing polyfill for tags field with query for plain array values
Regretfully, I don't have the resources right now to create a complete ticket with reproducible step-by-step code etc., but want to mention this nevertheless as this was a really hard nut to crack:

In 3.7 I was using this blueprint field definition, pulling in a plain array of tags using a site method:

```
tags:
  type: tags
  label: Tags
  options: query
  query:
    fetch: site.fieldcache('tags')
    text: "{{ arrayItem.value }}"
    value: "{{ arrayItem.value.slug }}"
```

The data returned by `$site->fieldcache('tags')` is of format:

```
[
  0 => 'tag1',
  1 => 'tag2',
  2 => 'tag3',
  // etc.
]
```

Related to: #4624 this no longer works in 3.8.0-rc.2 and actually leads to a very long loading time for the panel pages baring such field (is the server side of the Fiber exchange getting choked up on this in the background?). I had to change it to this in order to get my tag suggestions back and get rid of the long loading times:

```
tags:
  type: tags
  label: Tags
  options: query
  query: site.fieldcache('tags')
```

distantnative:
Fixes #4739

### DIFF
--- a/src/Option/OptionsQuery.php
+++ b/src/Option/OptionsQuery.php
@@ -3,9 +3,9 @@
 namespace Kirby\Option;
 
 use Kirby\Cms\Block;
+use Kirby\Cms\Field;
 use Kirby\Cms\File;
 use Kirby\Cms\ModelWithContent;
-use Kirby\Cms\Nest;
 use Kirby\Cms\Page;
 use Kirby\Cms\StructureObject;
 use Kirby\Cms\User;
@@ -34,6 +34,20 @@ class OptionsQuery extends OptionsProvider
 	) {
 	}
 
+	protected function collection(array $array): Collection
+	{
+		foreach ($array as $key => $value) {
+			if (is_scalar($value) === true) {
+				$array[$key] = new Obj([
+					'key'   => new Field(null, 'key', $key),
+					'value' => new Field(null, 'value', $value),
+				]);
+			}
+		}
+
+		return new Collection($array);
+	}
+
 	public static function factory(string|array $props): static
 	{
 		if (is_string($props) === true) {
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):

src/Option/OptionsQuery.php
```diff
@@ -51,19 +65,20 @@ public static function factory(string|array $props): static
 	 * Returns defaults for the following based on item type:
 	 * [query entry alias, default text query, default value query]
 	 */
-	protected function itemToDefaults(object $item): array
+	protected function itemToDefaults(array|object $item): array
 	{
 		return match (true) {
+			is_array($item),
 			$item instanceof Obj => [
 				'arrayItem',
```
lukasbestle, line 73: @distantnative I don't fully understand the logic, but it looks like with this change there's a mismatch between the query entry alias (still `arrayItem`) and the default queries. Is this intended?

distantnative: It is intended. We are shifting to using `item.` in the queries. You can see in line 165 that we always pass the item as `item` data to the query. However, people have used `arrayItem` and `structureItem` in their blueprints and will continue to do so. So line 73 makes sure that this is passed as alias, retrieved in line 164 and also added as data entry in line 165 to access the item in the query.

lukasbestle: Makes sense. I was just confused why lines 74-75 and 80-81 were changed but the lines for other item types weren't. Couldn't we change the others to `item` as well? Also a comment in line 163 about the alias (something like "the alias is supported for backwards-compatibility") would be great.

distantnative: I think for the other types we should stick with their naming. I don't think there is any benefit in calling `page` now `item` even yes technically that can be done.

lukasbestle: You are right, that makes sense. Now I understand the change. 😄

PR #4743, opened 2022-10-03
Fix creating page with UUID in blueprint query

### LINKED ISSUE #4736: [3.8.0-rc.2] $page->createChild throws "duplicate error"
_Note: some obsolete descriptions erased; see thread below for more detailed investigation on when this error occurs_

## Description
Trying to add a new note using
```
page('notes')>createChild([
    'slug' => 'test',
    'template' => 'post',
    'draft' => true,
]);
```
I receive an error `Post could not be created: A page draft with the URL appendix "test" already exists.` The error does not occur with 3.7.x but occurs starting 3.8.0-rc.1 ~, but **only if the content has already assigned UUIDs** (it can only be replicated with a 3.7 starterkit after running `Kirby\Uuid\Uuids::populate()`; before populating the starterkit content with UUIDs it works without throwing that error).~

~## To reproduce~
~1. Install a fresh 3.7 starterkit and upgrade the `kirby` folder to 3.8.0-rc.2~
~2. Create site/template/sandbox.php~
~3. Enter above code into the template and call the sandbox page – all works fine, new draft created~
~4. Delete the newly created draft to reset the scenario~
~5. Run `Kirby\Uuid\Uuids::populate()` once using that sandbox~
~6. Enter above code back into the template again and call the sandbox page~
~7. See error~

## Your setup
3.8.0-rc.1 + 3.8.0-rc.2

## Additional note
Typo in the release notes for 3.8-rc.1: `Kirby\Uuids\Uuids::populate()` should be `Kirby\Uuid\Uuids::populate()`

distantnative:
Fixes #4736

### Explanation
- `Form` class calls blueprint which (sadly still) runs all queries
- When uuid gets called from one of those queries, it creates the page already, page rules for duplicate detection throws exception
- `Form::for($page` - `$page` is the one that gets passed to blueprint queries
- Thus `$page` passed to `Form::for()` already needs to include generated UUID

## TODO
- [x] @sebastiangreger @doup can you confirm this fixes your issues?
- [x] Review by team

### DIFF
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -467,7 +467,7 @@ trait PageActions
 		// make sure that a UUID gets generated and
 		// added to content right away
 		$props['content'] ??= [];
-		$props['content']['uuid'] = Uuid::generate();
+		$props['content']['uuid'] ??= Uuid::generate();
 
 		// create a temporary page object
 		$page = Page::factory($props);
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -464,21 +464,19 @@ trait PageActions
 		$props['template'] = $props['model'] = strtolower($props['template'] ?? 'default');
 		$props['isDraft']  = ($props['draft'] ?? true);
 
+		// make sure that a UUID gets generated and
+		// added to content right away
+		$props['content'] ??= [];
+		$props['content']['uuid'] = Uuid::generate();
+
 		// create a temporary page object
 		$page = Page::factory($props);
 
-		// make sure that a UUID gets generated and
-		// added to the page object right away
-		$page = $page->clone(['content' => ['uuid' => Uuid::generate()]]);
-
-		// gather content
-		$content = $props['content'] ?? [];
-
 		// create a form for the page
-		$form = Form::for($page, ['values' => $content]);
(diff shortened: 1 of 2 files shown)


doup:
I've tried with the updated `src/Cms/PageActions.php` and now it's working. 👍 
Thanks!

src/Cms/PageActions.php
```diff
@@ -464,18 +464,16 @@ public static function create(array $props)
...
-		$page = Page::factory($props);
-
-		// gather content
-		$content = $props['content'] ?? [];
-
 		// make sure that a UUID gets generated and
 		// added to content right away
-		$content['uuid'] = Uuid::generate();
+		$props['content'] ??= [];
+		$props['content']['uuid'] = Uuid::generate();
```
distantnative: @Lukasbestle should we support passing a manual UUID here? So rather `??=`?

lukasbestle: Sounds good! 👍

sebastiangreger:
I, too, can confirm: with this branch the duplicate error is gone and the page is created correctly :partying_face:

bastianallgeier (approved):
I love how simple and clear this is.

PR #4746, opened 2022-10-03
`Dir::make()`: prevent race condition

### LINKED ISSUE #4745: [3.8.0-rc.2] UUID `mkdir(): File exists`
Screenshot 2022-10-03 at 19 44 46

Race condition when multiple sections etc try to populate UUID to cache.
File cache simultaneously tries to create the directories - checks at top of `Dir::make()` don't help due to the concurrency of the requests.

distantnative:
Fixes #4745

### Enhancements
- Prevent race condition errors for `Dir::make()`

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/Dir.php
+++ b/src/Filesystem/Dir.php
@@ -4,6 +4,7 @@ namespace Kirby\Filesystem;
 
 use Exception;
 use Kirby\Cms\App;
+use Kirby\Cms\Helpers;
 use Kirby\Cms\Page;
 use Kirby\Toolkit\Str;
 use Throwable;
@@ -391,7 +392,20 @@ class Dir
 			throw new Exception(sprintf('The directory "%s" cannot be created', $dir));
 		}
 
-		return mkdir($dir);
+		return Helpers::handleErrors(
+			fn (): bool => mkdir($dir),
+			function (&$override, int $errno, string $errstr): bool {
+				// if the dir was already created (race condition),
+				// consider it a success
+				if (Str::endsWith($errstr, 'File exists') === true) {
+					// drop the warning
+					return $override = true;
+				}
+
+				// handle every other warning normally
+				return false;
+			}
+		);
 	}
 
 	/**


distantnative:
@lukasbestle Am I overdoing it if I'd reduce this to

```php
return Helpers::handleErrors(
    fn (): bool => mkdir($dir),
    fn (&$override, int $errno, string $errstr): bool => $override = Str::endsWith($errstr, 'File exists')
);
```

lukasbestle:
Don't think that's still readable. TBH I think your change to combine the override and return value already goes quite far.

distantnative:
I get what you mean but to be honest I don't think the issue here is caused by the condensed code but the method syntax in general. That override variable and what one returns are very unclear when you don't know the handleError method itself. I think we need to refactor the method instead then. Could be one parameter for the condition when to suppress an error, second parameter for an altered return value if needed anything else than `true` when suppressed.

lukasbestle:
I like that idea. It would duplicate the logic in a lot of cases (as both callbacks would often use the same `if` checks), but it's certainly easier to read and understand than a variable passed by reference.

If we want to do it, 3.8 would be a good option because of the breaking change.

I'd also be fine with the current version though.

distantnative:
I'll try something this evening, I think this could be good for 3.8. And I think it won't have duplicated if-codes as those will be only necessary if there would be different conditions with different override/fallback values. Id there is just one, it should be sufficient to do 

```php
Helpers::handleErrors(
    fn () => ...,
    fn ($code, $message) => any condition when to suppress,
    value when suppressed
)
```

And `true` would be the default fallback for when the error is suppressed. Which would turn our boat existing use cases into just

```php
Helpers::handleErrors(
    fn () => ...,
    fn ($code, $message) => condition when to suppress
)
```

distantnative:
@bastianallgeier if you want to merge this already for RC3, go ahead. Will then change it on the refactoring PR for tomorrow.

afbora (approved):
I can confirm that the PR fix the issue. I can reproduce the issue with develop branch and no more `mkdir(): File exists` error seen with this PR 👍

PR #4750, opened 2022-10-04
Switch from `class_uses()` to duck test for `File::setModel()` 

### LINKED ISSUE #4748: `File::setModel()`: Switch from `class_uses()` to duck test
The following lines cause a regression when a model class is extended in a plugin (because `class_uses()` does not check parents):

https://github.com/getkirby/kirby/blob/cfdba1b81fc68d5bdf6561fc47858f6c966100d3/src/Filesystem/File.php#L447-L449

Instead we want to check for the existence of the `root()` and `url()` methods.

See https://github.com/mullema/k3-image-clip/issues/25 for the original discussion.

afbora:
### Fixes
- File class extending works as before #4748

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -444,7 +444,7 @@ class File
 	 */
 	protected function setModel(object|null $model = null): static
 	{
-		if ($model !== null && in_array(IsFile::class, class_uses($model)) !== true) {
+		if ($model !== null && method_exists($model, 'hasIsFileTrait') !== true) {
 			throw new InvalidArgumentException('The model object must use the "Kirby\Filesystem\IsFile" trait');
 		}
 
--- a/src/Filesystem/IsFile.php
+++ b/src/Filesystem/IsFile.php
@@ -109,6 +109,16 @@ trait IsFile
 		return file_exists($this->root()) === true;
 	}
 
+	/**
+	 * To check the existence of the IsFile trait
+	 *
+	 * @todo Switch to class constant in traits when min PHP version 8.2 required
+	 * @codeCoverageIgnore
+	 */
+	protected function hasIsFileTrait(): bool
+	{
+		return true;
+	}
 
 	/**
 	 * Returns the app instance
--- a/tests/Filesystem/FileTest.php
+++ b/tests/Filesystem/FileTest.php
@@ -8,6 +8,11 @@ use PHPUnit\Framework\TestCase as TestCase;
 
 require_once __DIR__ . '/mocks.php';
 
+class InvalidFileModel
+{
+	public string $foo = 'bar';
+}
+
 /**
  * @coversDefaultClass \Kirby\Filesystem\File
  */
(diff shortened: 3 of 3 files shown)


afbora:
@lukasbestle Another potential issue came up where we were checking for `root` and `url`. If a model has `url` and `root` method (like `Page` class), it is accepted. You can check out `Kirby\Filesystem\FileTest::testInvalidModel` unit test. Do you have an idea?

src/Filesystem/File.php
```diff
@@ -441,11 +442,17 @@ public function root(): string|null
...
 	 */
 	protected function setModel(object|null $model = null): static
 	{
-		if ($model !== null && in_array(IsFile::class, class_uses($model)) !== true) {
-			throw new InvalidArgumentException('The model object must use the "Kirby\Filesystem\IsFile" trait');
+		// TODO: create an interface instead of checking methods
+		if (
+			$model !== null &&
+			method_exists($model, 'root') !== true &&
+			method_exists($model, 'url') !== true
```
lukasbestle: This should be an or check because both methods must exist:

```suggestion
			(
				method_exists($model, 'root') !== true ||
				method_exists($model, 'url') !== true
			)
```

For cleaner code, we could move the `null` case to the top:

```php
if ($model === null) {
	$this->model = null;
	return $this;
}

if (
	method_exists($model, 'root') !== true ||
	method_exists($model, 'url') !== true
) {
	throw new InvalidArgumentException('The model object must use the "Kirby\Filesystem\IsFile" trait or have "root" and "url" methods');
}

$this->model = $model;
return $this;
```

lukasbestle (commented):
It's not ideal, but I feel like it's fine that other models are also accepted. We only need the model object for the root and URL and once those are there, everything will work as expected. It is unlikely that someone will pass a page as the model.

src/Filesystem/File.php
afbora: Thanks for suggestion 👍

lukasbestle (dismissed):

lukasbestle:
@distantnative Could you take a look at this as well? The solution is not 100% optimal, maybe you have a better idea. Otherwise I think we can still merge it because it fixes the regression.

distantnative:
I think this would be fine.
But had another idea, just want to share: what if the `IsFile` trait would add a class constant that our method would check for? This one would be inherited (whereas the has trait check doesn't work for parents).

lukasbestle:
OMG, that's an awesome hack! It's a hack, but I think it will work really well.

distantnative:
Except that I just read constants in traits are only a PHP 8.2 thing 😆
But I think a method `protected ::hasIsFileTrait(): bool` or so could just serve the same functionality

afbora:
Thank you @distantnative for great idea! I've implemented it. Ready to review 🚀

bastianallgeier (approved):

bastianallgeier:
Nice one!

PR #4751, opened 2022-10-04
Retrieve plugin versions from composer data 

### LINKED ISSUE #4733: Retrieve plugin versions from `vendor/composer/installed.php` if not provided in plugin's `composer.json`
https://github.com/mzur/kirby-uniform/pull/227
https://github.com/mzur/kirby-uniform/pull/229

afbora:
## This PR …

Do you think this development can be unit tested?

### Enhancements

- Improved version information of plugins installed with composer #4733

### Breaking changes

None

## Ready?

### For review team


lukasbestle (commented):
The code looks great!

Regarding unit tests: We could create a mock directory with just a `composer/installed.php` in it and register it with `(new Composer\Autoload\ClassLoader($path))->register()`. Then the `InstalledVersions` class should pick it up automatically. We need to make sure to call `unregister()` when tearing down the test class though to avoid breaking other tests.

So basically: `setUpBeforeClass()` registers the directory and stores it in a static prop of the test class. `tearDownAfterClass()` unregisters it again.

afbora:
Thank you so much @lukasbestle for detailed  unit test steps 🔥 I've never written such a test before. I applied what you wrote. I hope it's true 😅 Feel free to push your suggestions..

lukasbestle (commented):
Already really great! 👍

Two suggestions:

- Please move the test into the `PluginTest` class. A class `PluginComposerTest` would refer to a tested class `PluginComposer`, but we are still testing the `Plugin` class. You can add the set up and tear down code in the existing methods in `PluginTest`.
- Could you add a test for the override by `composer.json`? So another plugin that is contained in `composer/installed.php` but defines a different version in `composer.json`? `composer.json` should always be used then.

afbora:
@lukasbestle Done 👍

lukasbestle (dismissed):
I've tested it with the Uniform plugin (which currently does not define the `version` key) and found out that we need to check for the Composer package name, not the Kirby plugin name (they might differ). Now everything seems to be working. I have also tweaked the tests a bit for reliability and to avoid code duplication.

@afbora Could you take a final look at my changes?

afbora:
Thank you @lukasbestle Nice catch! 👍  
I've tested few more plugins like you and I've found an issue: ` 1.0.0+no-version-set`. 
Could you test following plugins:

- https://github.com/thathoff/kirby-git-content
- https://github.com/OblikStudio/kirby-memsource

````php
'root' => array(
    'name' => 'oblik/kirby-memsource',
    'pretty_version' => '1.0.0+no-version-set',
    'version' => '1.0.0.0',
    'reference' => NULL,
    'type' => 'kirby-plugin',
    'install_path' => __DIR__ . '/../../',
    'aliases' => array(),
    'dev' => true,
),
````

**Comment by afbora** [comment]
Also [uniform plugin](https://github.com/mzur/kirby-uniform) still doesn't work for me. Work for you?

screen-capture-1474-System - Mægazine-localhost

**Comment by lukasbestle** [comment]
I tested like this:

```sh
composer create-project getkirby/starterkit test
cd test
composer require thathoff/kirby-git-content oblik/kirby-memsource mzur/kirby-uniform

# manually update Kirby to this branch
# installing Kirby via Composer does not work because many plugins require
# `getkirby/cms: ^3.x`, which does not match `dev-*` branch versions
# (just a temporary issue, after the release, the following should not be necessary)
composer require composer/semver
rm -R kirby
git clone --branch fix/4733-plugin-version-from-composer https://github.com/getkirby/kirby.git
rm -R kirby/vendor
```

Which gives me this:

Bildschirmfoto 2022-10-05 um 22 13 25

afbora:
`Git Content` plugin doesn't includes `vendor` and I've to run composer install as manually.

### My steps

- Download the plugin as ZIP: https://github.com/thathoff/kirby-git-content/archive/refs/tags/v5.0.1.zip
- Unzip to `/site/plugins` directory
- `cd` into plugin directory
- run `composer install`
- check system view from panel
- see version: `1.0.0+no-version-set`

lukasbestle:
That's as expected. If you install from ZIP, Composer won't know where you downloaded it from. Because there is no `version` key in `composer.json`, it needs to set the version to unknown.

Getting the version from Composer only works if the plugin was installed with Composer. That's the reason why we recommend devs to include a `version` field for ZIP and Git submodule installation.

However the question is what our code can do with that unknown version. Expected behavior would be to return `null` from `$plugin->version()`. Maybe we need to hardcode that special string in an `if` clause and return `null`.

lukasbestle:
I feel like this deserves more testing. Should we move it to 3.8.1?

afbora:
As it stands, I'm still fine with `+no-version-set` version. I think, this is not important thing. If we can still improve it in RC or 3.8.1. But still, you know better 👍

bastianallgeier:
Let's move it to 3.8.1 if you are not 100% done with testing yet.

lukasbestle (approved):
Added a check for `+no-version-set`.

@afbora Could you test it again and check if we got all cases?

afbora:
@lukasbestle Yes, seems all cases we know works fine 👍

PR #4753, opened 2022-10-04
Fix broken label/headline in info section

bastianallgeier:
## This PR …

After moving to label instead of headline for the info section, we introduced a regression which actually broke both variants.

### Fixes

- The info section label can be set again with the deprecated headline option or the new label option. 

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/config/sections/info.php
+++ b/config/sections/info.php
@@ -25,9 +25,9 @@ return [
 	],
 	'toArray' => function () {
 		return [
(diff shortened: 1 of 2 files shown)


bastianallgeier:
Just realized that I fixed it in the wrong place. It makes a lot more sense to already use label in the response and fix the section vue component.

distantnative (approved):

PR #4755, opened 2022-10-04
Disable UUIDs via config option

distantnative:
### Enhancements
- UUIDs can be deactivated locally via `content.uuid` option

## Ready?

### For review team

### DIFF
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -451,9 +451,9 @@ class File extends ModelWithContent
 	 * Return the permanent URL to the file using its UUID
 	 * @since 3.8.0
 	 */
-	public function permalink(): string
+	public function permalink(): string|null
 	{
-		return $this->uuid()->url();
+		return $this->uuid()?->url();
 	}
 
 	/**
--- a/src/Panel/Page.php
+++ b/src/Panel/Page.php
@@ -49,18 +49,15 @@ class Page extends Model
 
 		$title = $this->model->title();
 
-		// get permalink/UUID as URL
-		$url = match ($type) {
-			'markdown' => $this->model->permalink(),
-			default    => $this->model->uuid()
-		};
-		// if UUIDs are disabled, fall back to actual URL
-		$url ??= $this->model->url();
+		// type: markdown
+		if ($type === 'markdown') {
+			$url = $this->model->permalink() ?? $this->model->url();
+			return '[' . $title . '](' . $url . ')';
+		}
 
-		return match ($type) {
-			'markdown' => '[' . $title . '](' . $url . ')',
-			default    => '(link: ' . $url . ' text: ' . $title . ')'
-		};
+		// type: kirbytext
+		$link = $this->model->uuid() ?? $this->model->uri();
+		return '(link: ' . $link . ' text: ' . $title . ')';
 	}
 
 	/**
--- a/src/Panel/Page.php
+++ b/src/Panel/Page.php
@@ -49,9 +49,17 @@ class Page extends Model
 
 		$title = $this->model->title();
 
+		// get permalink/UUID as URL
+		$url = match ($type) {
+			'markdown' => $this->model->permalink(),
+			default    => $this->model->uuid()
+		};
+		// if UUIDs are disabled, fall back to actual URL
+		$url ??= $this->model->url();
+
 		return match ($type) {
-			'markdown' => '[' . $title . '](' . $this->model->permalink() . ')',
-			default    => '(link: ' . $this->model->uuid() . ' text: ' . $title . ')'
+			'markdown' => '[' . $title . '](' . $url . ')',
+			default    => '(link: ' . $url . ' text: ' . $title . ')'
 		};
 	}
 
--- a/src/Uuid/Uuids.php
+++ b/src/Uuid/Uuids.php
@@ -92,7 +92,7 @@ class Uuids
 	public static function generate(string $type = 'all'): void
 	{
 		static::each(
-			fn (Identifiable $model) => Uuid::for($model)->id(),
+			fn (Identifiable $model) => Uuid::for($model)?->id(),
 			$type
 		);
 	}
@@ -106,7 +106,7 @@ class Uuids
 	public static function populate(string $type = 'all'): void
 	{
 		static::each(
-			fn (Identifiable $model) => Uuid::for($model)->populate(),
+			fn (Identifiable $model) => Uuid::for($model)?->populate(),
 			$type
 		);
 	}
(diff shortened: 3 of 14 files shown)


src/Uuid/Uuids.php
```diff
@@ -79,6 +79,11 @@ public static function each(Closure $callback, string $type = 'all'): void
 		// }
 	}
 
+	public static function enabled(): bool
+	{
+		return App::instance()->option('content.uuid') !== false;
```
lukasbestle, line 85: For strict checking of the option value:

```suggestion
		return App::instance()->option('content.uuid', true) === true;
```

tests/Uuid/UuidTest.php
```diff
@@ -246,6 +266,19 @@ public function testIs()
 		$this->assertFalse(Uuid::is('not a page://something'));
 	}
 
+	/**
+	 * @covers ::is
+	 */
+	public function testSsConfigDisabled()
```
lukasbestle:
```suggestion
	public function testIsConfigDisabled()
```

lukasbestle (commented):
Two places that IMO also need a check:

- `Uuids::each()` needs a check for `enabled()` (that just returns early) because `generate()` and `populate()` depend on `Uuid::for()`.
- `Panel\File::dragText()` and `Panel\Page::dragText()` need a fallback to the model URL instead of permalink/UUID.

src/Uuid/Uuids.php
distantnative: Since the default is activated, I think it makes more sense to check explicitly for false. And all other values still don't deactivate UUIDs, only false does.

distantnative:
Solved it rather in `Uuids::generate()` and `Uuids::populate()` directly as that's where the breaking call would occur

src/Cms/File.php
```diff
@@ -451,9 +451,9 @@ public function parents()
 	 * Return the permanent URL to the file using its UUID
 	 * @since 3.8.0
 	 */
-	public function permalink(): string
+	public function permalink(): string|null
```
bastianallgeier, line 454: Wouldn't it make sense to return the regular URL when UUIDs are disabled?

distantnative: Wasn't sure here. When somebody expects to get a permalink, is it better they fail when UUIDs are disabled (`null`) or better they receive something else, which is not a permalink though?

lukasbestle: I agree with Nico. Returning `null` is much safer. If the dev explicitly disables UUIDs, permalinks just won't be available.

src/Panel/Page.php
```diff
@@ -49,9 +49,17 @@ public function dragText(string|null $type = null): string
 
 		$title = $this->model->title();
 
+		// get permalink/UUID as URL
+		$url = match ($type) {
+			'markdown' => $this->model->permalink(),
+			default    => $this->model->uuid()
+		};
+		// if UUIDs are disabled, fall back to actual URL
+		$url ??= $this->model->url();
```
lukasbestle: I think in the `default` case (KirbyText) this could fall back to `$this->model->uri()` to be independent from the host.

At this point I think it could make sense have a top-level `if ($type === 'markdown')`, otherwise we'll have three `match ()` structures for the same options. 😅

src/Uuid/Uuids.php
```diff
@@ -87,7 +92,7 @@ public static function each(Closure $callback, string $type = 'all'): void
 	public static function generate(string $type = 'all'): void
 	{
 		static::each(
-			fn (Identifiable $model) => Uuid::for($model)->id(),
+			fn (Identifiable $model) => Uuid::for($model)?->id(),
```
lukasbestle: This certainly works, but what's the point of looping through every single model if we skip the operation for every one?

It's unlikely that these methods are called at all with disabled UUIDs, but if they are, I think they should return as quickly as possible (which is why I suggested to change the `each()` method). Alternatively we could also return early or (even better) throw an exception at the top of `generate()` and `populate()`.

distantnative: What's the point of calling `Uuids::generate()` when you have disabled UUIDs? I think we just need to make sure that the code does not break, which this does, and can focus on keeping the code concise here.

src/Cms/File.php
bastianallgeier: Ok, convinced.

src/Uuid/Uuids.php
lukasbestle: I feel like every method should do the best it can to handle the disabled UUIDs.

One case where these methods could be called is from plugins like Janitor. They could check for `Uuids::enabled()`, but we cannot expect that everyone optimizes their code perfectly. We just cannot know when these methods are called for which reasons from which code.

We have the opportunity to do it properly, and TBH I strongly think that making code concise is really far down the priority list here. It's a good goal, but not the most important one compared to performance and energy use.

src/Panel/Page.php
```diff
@@ -49,10 +49,15 @@ public function dragText(string|null $type = null): string
...
-			default    => '(link: ' . $this->model->uuid() . ' text: ' . $title . ')'
-		};
+		// type: markdown
+		if ($type === 'markdown') {
+			$url = $this->model->permalink() ?? $this->model->url();
+			return '[' . $title . '](' . $url . ')';
+		}
+
+		// type: kirbytext
+		$link = $this->model->uuid() ?? $this->model->url();
```
lukasbestle: Already much better, but what about using `uri()` here as I suggested?

src/Uuid/Uuids.php
distantnative: It's not like the current solution wouldn't work 100% fine... let's get a third opinion here.

src/Panel/Page.php
distantnative: 🙈

src/Uuid/Uuids.php
lukasbestle: I think we also need to consider the "debuggability" of the code. If you call the methods and they do absolutely nothing, it could take ages to find out that the reason is the disabled option. So I'd much prefer exceptions.

bastianallgeier: I agree with @lukasbestle here. I generally would always avoid any form of code execution as early as possible if it is not needed.

lukasbestle: @bastianallgeier What about throwing an exception instead of just returning? I think it would be much more explicit and easier to debug.

bastianallgeier: Yep, that's much better and it's also easier to test.

bastianallgeier (approved):

PR #4760, opened 2022-10-04
Refactor `Helpers::handleErrors()`

distantnative:
### Refactored
- New syntax for `Helpers::handleErrors()`:
```php
Helpers::handleErrors(
    fn () => // error to suppress,
    fn (int $errno, string $errstr): bool => // condition when to suppress error
    $value // return value for when the error is suppressed
);
```

This breaks the method quite a bit, but given that it is very, very niche and only was introduced in 3.7.4, I think it is ok to do this in 3.8.0 without any deprecation path prior to this.

### Breaking changes
- `Helpers::handleErrors()`'s 2nd parameter now only determines whether the error is suppressed, its 3rd parameter defines what is returned when an error is suppressed

## Ready?

### For review team

### DIFF
--- a/src/Cms/Helpers.php
+++ b/src/Cms/Helpers.php
@@ -64,27 +64,36 @@ class Helpers
 	 * @since 3.7.4
 	 *
 	 * @param \Closure $action Any action that may cause an error or warning
-	 * @param \Closure $handler Custom callback like for `set_error_handler()`;
-	 *                          the first argument is a return value override passed
-	 *                          by reference, the additional arguments come from
-	 *                          `set_error_handler()`; returning `false` activates
-	 *                          error handling by Whoops and/or PHP
-	 * @return mixed Return value of the `$action` closure, possibly overridden by `$handler`
+	 * @param \Closure $condition Closure that returns bool to determine if to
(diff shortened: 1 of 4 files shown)


src/Cms/Helpers.php
```diff
@@ -64,29 +64,38 @@ public static function dump($variable, bool $echo = true): string
...
-		$oldHandler = set_error_handler(function () use (&$override, &$oldHandler, $handler) {
-			$handlerResult = $handler($override, ...func_get_args());
+		$override = null;
 
-			if ($handlerResult === false) {
+		$handler = set_error_handler(function () use (&$override, &$handler, $condition, $fallback) {
+			// check if suppress condition is met
+			$suppress = $condition(...func_get_args());
+
+			if ($suppress === false) {
```
lukasbestle: I feel like with the new fallback handling, we should rather check for an explicit `true` value. So if the condition returns `null`/`void`, the error won't be suppressed.

```suggestion
			if ($suppress !== true) {
```

lukasbestle (commented):
Absolutely love how simple it is this way. 💛

Only a small suggestion:

bastianallgeier (approved):

PR #4763, opened 2022-10-04
Also use tabular nums with disabled update check

lukasbestle:


bastianallgeier (approved):
Ah, I totally missed that.

PR #4771, opened 2022-10-07
Support UUIDs in API page routes

### LINKED ISSUE #4769: JS api: support for UUIDs

distantnative:
### Fixes
- #4769 

## Ready?

### For review team

### DIFF
--- a/config/api/routes/files.php
+++ b/config/api/routes/files.php
@@ -7,7 +7,13 @@ $pattern = '(account|pages/[^/]+|site|users/[^/]+)';
  * Files Routes
  */
 return [
-
+	[
+		'pattern' => 'files/@/(:any)',
+		'method'  => 'GET',
+		'action'  => function (string $id) {
+			return $this->file('file://' . $id, false);
+		}
+	],
 	[
 		'pattern' => $pattern . '/files/(:any)/sections/(:any)',
 		'method'  => 'GET',
--- a/panel/src/api/files.js
+++ b/panel/src/api/files.js
@@ -9,7 +9,22 @@ export default (api) => {
 			return api.delete(parent + "/files/" + filename);
 		},
 		async get(parent, filename, query) {
-			let file = await api.get(parent + "/files/" + filename, query);
+			let file;
+
+			if (parent.includes("://") === true) {
+				// UUID: $api.files.get("file://hbjkh324k234", { query })
+				if (!query && filename) {
+					query = filename;
+				}
+
+				file = await api.get(
+					"files/@/" + parent.substring(parent.indexOf("://") + 3),
+					query
+				);
+			} else {
+				// standard model ID + filename
+				file = await api.get(parent + "/files/" + filename, query);
+			}
 
 			if (Array.isArray(file.content) === true) {
 				file.content = {};
--- a/panel/src/api/pages.js
+++ b/panel/src/api/pages.js
@@ -55,6 +55,11 @@ export default (api) => {
 			return page;
 		},
 		id(id) {
+			// UUID
+			if (id.includes("://") === true) {
+				return "@/" + id.substring(id.indexOf("://") + 3);
+			}
+
 			return id.replace(/\//g, "+");
 		},
 		async files(id, query) {
--- a/config/api/routes/pages.php
+++ b/config/api/routes/pages.php
@@ -5,7 +5,15 @@
  * Page Routes
  */
 return [
-
+	[
+		'pattern' => 'pages/@/(:any)',
+		'method'  => 'GET',
+		'action'  => function (string $id) {
+			// make sure not to decode ID again (replace + with /)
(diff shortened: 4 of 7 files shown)


lukasbestle (dismissed):

distantnative:
@bastianallgeier would be good to get your perspective on this solution

bastianallgeier:
Can you give an example of a query with the @?

distantnative:
Some people were trying to do `this.$api.pages.get("page://fgosyuh38432j")` - which currently fails due to our conversion of `/`. So my idea was to convert `://` first with something else.

bastianallgeier:
Ahhh, now I get it. The @ is just internally. The user still passes a regular UUID. Sounds good in general. I just wonder what happens to files with something like `filename@2x.png` in the path

distantnative:
Right now the implementation is just for pages... :/ maybe you have a better idea

bastianallgeier:
I would suggest to move this either to 3.8.2 or solve it until the release if we have a better idea. But right now it feels like a partial solution and I don't really have a better idea for a good replacement yet.

distantnative:
Yea, true.

Also adding another problem:
One cannot select uuid yet as it is not supported in the API models.

panel/src/api/files.js
```diff
@@ -9,7 +9,22 @@ export default (api) => {
...
+			let file;
+
+			if (parent.includes("://") === true) {
+				// UUID: $api.files.get("file://hbjkh324k234", { query })
+				if (!query && filename) {
+					query = filename;
+				}
+
+				file = await api.get(
+					"files/@/" + parent.substring(parent.indexOf("://") + 3),
```
lukasbestle, line 21: I feel like the method should validate that the scheme is `file://` and then use the fixed offset of `7` for the `substring`.

This catches invalid uses such as `$api.files.get("page://hbjkh324k234")`. A call like this could happen by accident and it would be much easier to debug if there was a clear error in this case, even before the request is sent.

panel/src/api/pages.js
```diff
@@ -55,6 +55,11 @@ export default (api) => {
 			return page;
 		},
 		id(id) {
+			// UUID
+			if (id.includes("://") === true) {
+				return "@/" + id.substring(id.indexOf("://") + 3);
```
lukasbestle, line 60: Same here:

```suggestion
			if (id.startsWith("page://") === true) {
				return "@/" + id.substring(7);
```

config/api/routes/files.php
```diff
@@ -7,7 +7,13 @@
  * Files Routes
  */
 return [
-
+	[
+		'pattern' => 'files/@/(:any)',
+		'method'  => 'GET',
+		'action'  => function (string $id) {
+			return $this->file('file://' . $id, false);
```
lukasbestle, line 14: Doesn't this call `Cms\Api::file()`? This method expects a string as second argument. I think it needs to be:

```suggestion
			return $this->file('file://' . $id);
```

And then `Api::file()` needs to support the `null` value.

bastianallgeier:
TBH I feel like we are creating an additional mess here. API::file and API::page with additional arguments for decode or the magic second null argument for finding files with UUIDs, scream for a cleaner option. I think we also still don't really solve the general architectural problem. What about all the related routes for a page? It's still impossible to use UUIDs to delete a page, update a title etc.

distantnative:
True. But then I really have no idea how to support UUIDs

lukasbestle:
I still think the URL patterns `/pages/@/...` and `/files/@/...` are the right base for this feature as the `@` sign is already reserved, so it makes a lot of sense.

As far as I can see, the only parts we need to solve are:

- the call signatures for the JS methods (cleaner arguments) and
- additional routes for operations on files/pages (like `DELETE`, `PATCH` etc.).

distantnative:
@bastianallgeier @lukasbestle  we really need a decision here. I want to either finish it or delete this PR, but we cannot keep pushing it aside. I'm leaning to deleting, this doesn't;t seem right yet as solution. But we also need to acknowledge how broken our API is for UUID usage.

lukasbestle:
I agree it's not there yet in its current form. But I think we can't just close it without a replacement PR. In the end we need to solve API access to UUIDs, don't we?

distantnative:
It just doesn't help much to have this PR around while we don't engage. But I also think we aren't getting much further here without Bastian.

bastianallgeier:
After all the work on v4 and on solving issues more pragmatically, my perspective shifted quite a bit. The mess on the server can be cleaned up later if even necessary. As long as it is stable I'm happy. Looking at it again, I think the route pattern is indeed good and that's what counts the most for our users. I even think that we can start with all the get routes first and then add more actions later to move forward.

distantnative:
@bastianallgeier you solved this separately for v4, right? So this PR can be closed?

PR #4773, opened 2022-10-08
Remove unnecessary `else` statements

distantnative:
### Refactoring
- Removed unnecessary else statement where it helps to improve readability of the code

## Ready?

### For review team

### DIFF
--- a/src/Cms/Content.php
+++ b/src/Cms/Content.php
@@ -115,7 +115,7 @@ class Content
 			$oldField = $oldFields->get($name);
 
 			// field name and type matches with old template
-			if ($oldField && $oldField->type() === $newField->type()) {
+			if ($oldField?->type() === $newField->type()) {
 				$data[$name] = $content->get($name)->value();
 			} else {
 				$data[$name] = $newField->default();
--- a/src/Cms/System.php
+++ b/src/Cms/System.php
@@ -322,12 +322,11 @@ class System
 
 		// only return the actual license key if the
 		// current user has appropriate permissions
-		$user = $this->app->user();
-		if ($user && $user->isAdmin() === true) {
+		if ($this->app->user()?->isAdmin() === true) {
 			return $license['license'];
-		} else {
-			return true;
 		}
+
+		return true;
 	}
 
 	/**
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -899,8 +899,6 @@ class Query
 			$column = Str::lower($match[1]);
 			return $this->findBy($column, $arguments[0]);
 		}
-
-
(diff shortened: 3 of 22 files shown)


afbora (dismissed):

src/Cms/Content.php
```diff
@@ -115,7 +115,7 @@ public function convertTo(string $to): array
 			$oldField = $oldFields->get($name);
 
 			// field name and type matches with old template
-			if ($oldField && $oldField->type() === $newField->type()) {
+			if ($oldField?->type() === $newField->type()) {
```
lukasbestle, line 118: TBH I think this is less clear now. Previously you could see on the first glance that this `if` only applies if there is an `$oldField`. Now you need to spot that question mark.

src/Cms/System.php
```diff
@@ -322,12 +322,11 @@ public function license()
 
 		// only return the actual license key if the
 		// current user has appropriate permissions
-		$user = $this->app->user();
-		if ($user && $user->isAdmin() === true) {
+		if ($this->app->user()?->isAdmin() === true) {
```
lukasbestle, line 325: Here I like it on the other hand because it really makes the code a lot shorter and increases the overview.

src/Database/Query.php
```diff
@@ -898,9 +898,10 @@ public function __call(string $method, array $arguments = [])
 		if (preg_match('!^findBy([a-z]+)!i', $method, $match)) {
 			$column = Str::lower($match[1]);
 			return $this->findBy($column, $arguments[0]);
-		} else {
-			throw new InvalidArgumentException('Invalid query method: ' . $method, static::ERROR_INVALID_QUERY_METHOD);
 		}
+
+
```
lukasbestle: Double newline:

```suggestion

```

src/Cms/Content.php
distantnative: The check s there to see if old and new field have the same type. I think it's rather clear that if oldField does not exist the type wouldn't be the same even if we don't first check explicitly for oldField on its own.

distantnative:
@lukasbestle is this ok to be merged then? Needs an approval then, can't merge without. thanks!

bastianallgeier:
@distantnative the code cov issue is really just about rediscovering lines that are as uncovered as they were before, right? Otherwise it looks good to me. Feels nice and clean.

bastianallgeier (approved):

distantnative:
@bastianallgeier yes that's my understanding. So I think it's one of the cases where manual review can replace that check.

lukasbestle:
Yep.

PR #4775, opened 2022-10-09
Whoops: exceptions to PHP error log

### LINKED ISSUE #4764: Fatal errors not logged
## Description
Errors are never logged in the default PHP log file. On Linux systems, that's likely `/var/log/php-fpm/error.log`. This is an issue when you've disabled debug mode on your live environment and there are HTTP 500 errors. Locally, errors aren't logged even when debug mode is enabled.

**Expected behavior**  
Errors should always get logged in the error log file.

My guess is that this happens because the Whoops handler intercepts the default PHP error handling behavior. If I comment out this line, I correctly see errors logged:

https://github.com/getkirby/kirby/blob/3.5.7.1/src/Cms/AppErrors.php#L96

If I set the `whoops` option to `false` in config.php, it also seems to fix the problem. But I don't want that on production, because now the default PHP error behavior kicks in and spits out file paths, line numbers, and other sensitive information.

In summation, I need Kirby to obscure the error so bad actors can't take advantage of it, while logging it so I can debug it later.

## To reproduce
1. Call an unexisting function in a template, e.g. `lorem_ipsum()` to trigger an error
2. Open the error log file and you should see that there's no error logged

## Your setup
**Kirby Version**  
3.5.7.1

- OS: Windows 10

distantnative:
### Fixes
- Fatal errors are written to PHP error log 
#4764

## Ready?

### For review team

### DIFF
--- a/src/Cms/AppErrors.php
+++ b/src/Cms/AppErrors.php
@@ -164,6 +164,7 @@ trait AppErrors
 		$whoops->clearHandlers();
 		$whoops->pushHandler($handler);
 		$whoops->pushHandler($this->getExceptionHookWhoopsHandler());
+		$whoops->pushHandler($this->getFatalErrorLogWhoopsHandler());
 		$whoops->register(); // will only do something if not already registered
 	}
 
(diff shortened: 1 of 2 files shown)


src/Cms/AppErrors.php
```diff
@@ -180,6 +181,19 @@ protected function getExceptionHookWhoopsHandler(): CallbackHandler
...
 
+	/**
+	 * Initializes a callback handler for logging all fatal errors
+	 *
+	 * @return \Whoops\Handler\CallbackHandler
+	 */
+	protected function getFatalErrorLogWhoopsHandler(): CallbackHandler
+	{
+		return new CallbackHandler(function ($exception, $inspector, $run) {
+			error_log($exception->getMessage());
```
lukasbestle: I'd use just `error_log($exception)` so that the stack trace is printed as well.

lukasbestle (commented):
Maybe we could even merge this into the current `getExceptionHookWhoopsHandler()` method (which would then be renamed of course). I don't think there's a need for two separate handlers.

lukasbestle (approved):

Issue #4762, opened 2022-10-04
`$uuid->isCached()`: Return `true` if the UUID object was already loaded from cache


lukasbestle:
Avoids another round trip to the cache.

Details: https://github.com/getkirby/kirby/issues/4757#issuecomment-1267529783

lukasbestle:
This enhancement is not useful as discussed in https://github.com/getkirby/kirby/pull/4778#issuecomment-1297885699.

PR #4778, opened 2022-10-10
`Uuid`: don’t populate when read from cache

distantnative:
@lukasbestle Seems like I am too blind to realise why suddenly unit tests keep failing.

### DIFF
--- a/src/Uuid/Uuid.php
+++ b/src/Uuid/Uuid.php
@@ -55,6 +55,12 @@ class Uuid
 	 */
 	public Collection|null $context;
 
+	/**
+	 * Flag whether the model instance was
+	 * looked up from the cache
+	 */
+	protected bool $fromCache = false;
+
 	public Identifiable|null $model;
 	public Uri $uri;
 
@@ -103,6 +109,9 @@ class Uuid
 				}
 			}
 		}
+
+		// reset cache flag as UUID won't be in cache anymore
+		$this->fromCache = false;
 
 		return Uuids::cache()->remove($this->key());
 	}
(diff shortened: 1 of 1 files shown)


lukasbestle:
From reading through the code I also can't find the reason. Might need step debugging with Xdebug to find which methods are called in which order.

distantnative:
Found a way to implement https://github.com/getkirby/kirby/issues/4762 properly without breaking unit tests.

However, for that to work `Uuid::populate()` must first run `Uuid::key()` and `Uuid::value()` before checking and escaping out if already cached. Not sure if that's really the better performance trade off just to save some additional calls of `Uuid::isCached()`.

distantnative:
The more I think about this, the more I think it's a bad idea and we should just stick with the current checking the cache itself.

lukasbestle:
> Found a way to implement #4762 properly without breaking unit tests.
> 
> However, for that to work `Uuid::populate()` must first run `Uuid::key()` and `Uuid::value()` before checking and escaping out if already cached.

Could you please explain why this is necessary and why the tests broke without this?

The idea of #4762 was to skip the populate step when we actively know that we got the ID from cache (thus making the process faster in the 99% case of a repeated page load). If we call `value()`, we don't gain anything as it calls `model()`, which queries the value from cache (exactly what would happen anyway just below).

distantnative:
Cause we only know if it got retrieved from cache after calling `->model()` once. Neither way of initializing (via passing a model object or an ID string) says whether it's already cached or not. But that's how you would call the populate method, either `Uuid::for($page)->populate()` or `UUID::for('4688ghftj57gjb')->populate()`. 

And you are right, we wouldn't gain anything, which is why I suggest to close this. I don't think the idea actually works.

bastianallgeier:
Ok, I'm closing it.

PR #4779, opened 2022-10-10
feat: externalize vue, make accessible to plugins

### LINKED ISSUE #4735: Plugins break if they use Vue 2.7 functionality as the Panel doesn't expose the required functions
## Description

Panel plugins are built with Vue, but don't ship their own runtime. Instead, they're loaded into the Vue instance the Panel app itself is using. Because of this, Vue should be externalized in the build config of `kirbyup` or a respective manual plugin setup, so your component is then compiled to code similar to this:

```js
(function(_vue) {
  return {
    name: 'MyKirbyPanelComponent',
    data: () => ({ level: 50 }),
    created() {
      console.log('Panel component ready!')
    },
    methods: {
      inc() { this.level += 10 }
    }
    template: `
    <div>
	  <p>Kirby is this awesome: <b>{{ level }}/100</b></p>
      <button @click="inc">Increase awesomeness</button>
    </div>`
  }
})(Vue) /* pass in the global window.Vue instead of shipping your own instance of "vue" */
```

Unfortunately, while Kirby *does* provide access to the global Vue instance via the `use` option, it is not exposed globally as `window.Vue`, `window.panel.Vue2` or similar, so in the example above, `Vue` at the bottom is actually undefined and breaks the plugin if used. This has caused some issues in the past, e.g. https://github.com/johannschopplich/kirbyup/issues/8 or https://github.com/johannschopplich/kirbyup/issues/24 (where workarounds could be implemented, e.g. by using a separate library instead of Vue for the event bus)

If some compiled components need access, Kirby plugins can work around that through the `use` option and expose it themselves, using plugin code like this:

```js
panel.plugin('my/example', {
  use: {
    plugin(Vue) {
     window._Vue = Vue
    }
  }
})
```

Now Vue 2.7 introduced new functionality for which access to the `Vue` package is required, but due to module shenanigans the plugin workaround from above is not sufficient anymore:
The new functionality is implemented using functions that are independent from the main `Vue` constructor, like `useCssVars()` or `onMounted()`. In the universal UMD build of Vue, those functions are attached to `Vue`, so can be accessed through e.g., `Vue.useCssVars`. However, in the ES module build, those functions are (only) exported separately:

```js
export default Vue

export function useCssVars() {}
```

In that case, the new 2.7 functions can not be accessed via `Vue.useCssVars`. Since Kirby bundles this ES module version, the `Vue` constructor it exposes via the `use` option does not have the new functions attached and thus 2.7 functionality is not usable at all, even if a plugin opts-in by using the `use` option to give its components access to `Vue`.

Here's an overview of Vue 2.7 features and changes that all rely on `Vue` to be there and break otherwise:

<details>
<summary><b>Most prominent, support for the Composition API</b></summary>

With the Composition API, components can now use `setup` and `<script setup>`. The (compiled) component from above can now look like this:

```js
(function(_vue) {
  return {
    name: 'MyKirbyPanelComponent',
    setup() {
      const level = _vue.ref(50)
      const inc = () => count.value++
      _vue.onCreated(() => console.log('Panel component ready!'))
      return { level, inc }
    },
    template: `
    <div>
	  <p>Kirby is this awesome: <b>{{ level }}/100</b></p>
	  <button @click="inc">Increase awesomeness</button>
    </div>`
  }
})(Vue) /* pass in the global window.Vue instead of shipping your own instance of "vue" */
```

Notice how now the code *does* try to access properties of Vue itself, like `Vue.ref` or `Vue.onMounted`!

</details>

<details>
<summary><b>Dynamic CSS using <code>v-bind()</code> in CSS</b></summary>

You can now write styles like:

```html
<script>
//...
data: () => ({ headlineColor: 'rebeccapurple' }),
methods: {
  changeHeadlineColor(color) {
    this.headlineColor = color
  }
}
</script>

<style>
h1 {
  color: v-bind(headlineColor);
}
</style>
```

The compiled code will try to access `Vue.useCssVars`.

</details>

<details>
<summary><b>JSX</b></summary>

Render functions compiled from JSX now use the global `h` export instead of a parameter:

```js
//    2.7
render() { return _vue.h('h1', 'Hello World') }
// <= 2.6
render(_h) { return _h('h1', 'Hello World') }
```

</details>

So like before, by default **Panel plugins break if the plugins tries to use Vue or any of its methods**. Now with version 2.7 and its support for the Composition API (`setup`/`<script setup>`), that is both much more likely to happen *and* can't be worked around anymore through `use`. So instead of your component you'll see:

```
Uncaught ReferenceError: Vue is not defined
    at index.js?1664631505:3:1174

// or, if Vue is exposed:

Uncaught TypeError: Vue.useCssVars is not a function
    at index.js?1664631505:3:1174
```

😢

**Expected behavior**  

Either:

- Panel plugins don't break if they try to access Vue and its methods

or:

- the limitation that plugin code must not access Vue and therefore must not take advantage of the `Composition API`, `Dynamic CSS` and other functionality that relies on Vue to be accessible, is clearly documented

## To reproduce

https://github.com/jonaskuske/kirby-vue-2-7-playground

## Your setup
**Kirby Version**  

3.7

**Console output**  

```
Uncaught ReferenceError: Vue is not defined
    at index.js?1664631505:3:1174
```

 - Device: PC 
 - OS: Win 10 
 - Browser: Edge 
 - Version: 105 

**Additional context**  

- https://blog.vuejs.org/posts/vue-2-7-naruto.html

**Suggestions for fixing this (happy to submit a PR!):**

- provide the full UMD build `vue.min.js` instead of the ESM version with separated exports
- optional but would make for a much better experience:
	add `externals: ['vue']` to the Panel's Vite config so it externalizes `vue` and plugins can access `window.Vue` the same way the panel does

Or:
- accept the limitations and document them well

jonaskuske:
## This PR …

...externalizes Vue so that plugins can access its functions and methods without jumping through hoops.

There's been no definitive call yet on which way to take https://github.com/getkirby/kirby/issues/4735 – documenting the limitation or fixing the errors. This PR is an example of what a fix could look like.

TL;DR on what this is about:
- Kirby currently doesn't expose Vue to plugins through a global variable
- this is ok for most 2.6 use cases, but breaks Vue 2.7 functionality
- it can be fixed by exposing Vue somewhere on window and letting the Panel also access it from there ("externalizing Vue"), which is what this PR is doing
- Vue 3 will require a setup like this, unless support for Single File Components is dropped and features like CSS vars and composables remain unsupported (string/DOM templates + options API should work without access to `Vue`)

The main reservation voiced in the issue thread is potential future breakage, as plugins can access `Vue` and thus register global directives, components etc. that affect the entire app. This capability is not new though, plugins can already access `Vue` and do global modifications through the `use` option of Kirby plugins. However, the `use` way is more explicit and cumbersome and thus less likely to be used, so the global variable might indeed be a bit "riskier" in that regard.

As for the implementation, I've exposed Vue on `window.Vue` for now as that's the default and made the implementation easier. It allows us to use the dist files provided by Vue instead of having to bundle our own UMD files that set some custom global variable.

### Fixes

Fixes https://github.com/getkirby/kirby/issues/4735

### Breaking changes

No

## Ready?

### For review team


distantnative:
> However, the use way is more explicit and cumbersome and thus less likely to be used, so the global variable might indeed be a bit "riskier" in that regard.

I'll try to explain my concern here again, but maybe also my understanding is flawed here. While Vue 2 uses a lot the imported `Vue` for all kind of things, Vue 3 moves almost completely away from that and ties all to the actual `app` instance. When we offer an actual `use` intention, it makes it quite easy for us when moving from Vue 2 to 3 not to tie those extensions to `Vue.use()` anymore but `app.use()`. If we now expose the global `Vue` and plugins start using it, we will just have to remove that and completely break those plugins when moving to Vue 3. Which is why I'd perceive exposing `Vue` like this as an additional hurdle to move to Vue 3 at any point in time.

jonaskuske:
> I'll try to explain my concern here again

Thanks! That was helpful :)

> When we offer an actual use intention, it makes it quite easy for us when moving from Vue 2 to 3 not to tie those extensions to `Vue.use()` anymore but `app.use()`.

That is true! I just don't think it plays more than a teeny-tiny role in the grand scheme of things, because it's a small change affecting only a handful of plugins (if any at all), has a quick and easy fix and bluntly spoken, an upgrade to Vue 3 will completely break virtually all plugins anyway, unfortunately.

`Vue.{use,filter,..}` being moved to the app instance is just the first on a [sizable list of breaking changes](https://v3-migration.vuejs.org/breaking-changes/), so the messaging from Kirby's side would just be "yeah, number 1 out of 35 on the list of breaking changes is applicable, just like the other ones". But those breaking changes aren't even the main problem with the Vue 3 migration! From a quick browse through the plugin gallery, basically all Panel plugins use `.vue` Single File Components instead of string templates. This means that **none** of these Panel plugins is going to work in Vue 3, all of them will need a re-compilation at the least.

We can ask the question the other way around:
#### What advantage does not exposing `Vue` have for the Vue 3 migration?

The only case where the change would cause a plugin to break that otherwise wouldn't is a hypothetical plugin that's
- released some time from now
- only uses string-based templates
- only uses template semantics and features available in both Vue 2 and 3
- explicitly uses `window.Vue.component()` or similar methods to globally register a component
	(instead of using the component option of the Vue component or the Kirby plugin definition)

I don't expect this to apply even to a single plugin tbh.

Any other component that might use `Vue.{use,filter,..}` will have to do a manual migration anyway, and during this migration run one additional search-and-replace. And there too, I expect this to affect a handful of plugins at most, as it can only happen for new/updated plugins and globally registering stuff isn't that common to begin with.

So unless I'm missing something here, I don't think either of this has enough of an impact to justify not supporting 2.7 (and getting the occasional "why does my code break when I or some library imports Vue?" issues).

jonaskuske:
And to clear this up:

> While Vue 2 uses a lot the imported Vue for all kind of things, Vue 3 moves almost completely away from that and ties all to the actual app instance

Vue 2 uses the imported Vue for six things that Vue 3 doesn't: the `config` property and the methods for globally registering a `component`, `filter`, `directive`, `mixin` or a plugin with `use`. Since with Vue 2 those are found on `Vue` itself, you can have multiple apps on one page, but if you register a global component for use in one of them, it will be available in all other apps too:

<p align=center></p>


In Vue 3, those "global" methods are found on the app instance returned from `createApp`, so you can have multiple apps with different global components, directives etc. each:

<p align=center></p>

But everything else that's required to render and run a Vue app, like `createElement` (Vue 3), `useCssVars` or `onMounted` remains exposed on `Vue` itself, from where it's imported – the black portions in the graphic above. In Vue 2, those imports are only necessary for 2.7 functionality, which is why in <=2.6 you get away with not exposing Vue at all, which is what Kirby is doing. That's possible because a compiled component in Vue 2 looks similar to this:

```js
render(h, ctx) {
  return ctx.createElement('p', {}, [ctx.createTextElement('Hello world!')])
}
```

Essential functions like `createElement` are found on the context that Vue passes into the component, so the component doesn't have to access Vue directly. In Vue 3 however, that context is not passed in and components access `Vue` directly:

```js
render() {
  return Vue.createElement('p', {}, [Vue.createTextElement('Hello world!')])
}
```

So without imposing massive restrictions (no Single File Components, no 3rd party libraries, no composables, much more), Vue 3 is going to require this externalized and shared setup I'm proposing here. Which of course is no reason to merge this *now*, just to clear up why I disagreed in the issue thread and said that it will be required in the future :)

lukasbestle:
If I understood it correctly, this means that plugin devs who use `window.Vue` while Kirby still uses Vue 2 can continue to use `window.Vue` with Vue 3, but they'll need to migrate their code to be compatible with Vue 3 in any case (no matter if we externalize `Vue` now or don't).

Does this mean that if we externalize `Vue` now and keep it this way when we move to Vue 3, simpler plugins that aren't affected by Vue 3 breaking changes will continue working, while every single plugin that uses SFC will break if we don't?

jonaskuske:
> plugin devs who use `window.Vue` while Kirby still uses Vue 2 can continue to use `window.Vue` with Vue 3

Yep! This component will work in both Vue 2 and Vue 3 for example:

```js
const Component = {
  setup() {
    const count = window.Vue.ref(0)
    const increase = () => count.value++

    window.Vue.watch(count, () => log('Count increased!'))

    return { count, increase }
  },
  template: '<button @click="increase">{{ count }}</button>'
}
```

> but they'll need to migrate their code to be compatible with Vue 3 in any case (no matter if we externalize Vue now or don't)

If they're using SFCs or are affected by any other breaking changes then yes, they will.
If usage of `window.Vue.{component,use,mixin,directive}` is the sole reason a plugin needs to migrate, then not externalizing would have prevented the migration, as the plugins wouldn't have access to these methods in the first place. Whether that outweighs support for 2.7 is the gretchenfrage here.

> Does this mean that if we externalize Vue now and keep it this way when we move to Vue 3, simpler plugins that aren't affected by Vue 3 breaking changes will continue working

If simple includes "not using SFCs" then yes.

> while every single plugin that uses SFC will break if we don't?

Every SFC will break if we don't externalize *as part of the Vue 3 migration*. (or more correct: it'll break anyway, but without externalizing it can't be fixed by plugin devs)
Whether we externalize *now* doesn't change the migration story, it just enables new features and prevents some errors at the cost of providing access to a few methods that are affected by a breaking change and thus require a migration if used.

distantnative:
I still don't quite get why SFCs would automatically break with Vue 3 - but that doesn't matter on the decision here.

jonaskuske:
> I still don't quite get why SFCs would automatically break with Vue 3 - but that doesn't matter on the decision here.

Because the `/index.js` that kirbyup (or similar) compiled from it is Vue 2-specific.

```js
// src/Demo.vue
<template><p>Hello!</p></template>

// src/index.js
import Demo from './Demo.vue'
panel.plugin('name', { components: { Demo } })
```

↓

`kirbyup build src/index.js`

↓

```js
// /index.js (simplified)
var Demo = function render() {
  var _vm = this, _c = _vm._self._c;
  return _c("p", [_vm._v("Hello!")]);
};
panel.plugin("name", { components: { Demo } });
```
This code works in Vue 2, but if you load the `Demo` component into a Vue 3 app, you'll just see
```
[Vue warn]: Property "_self" was accessed during render but is not defined on instance. 
  at <App>
        
[Vue warn]: Unhandled error during execution of render function 
  at <App>
        
Uncaught TypeError: Cannot read properties of undefined (reading '_c')
  at Proxy.render (index.js:59:36)
```
because `_self`, `_c`, `_v` and a lot of other properties used by Vue 2 components don't exist anymore in Vue 3.

To fix this, a plugin dev can run `kirbyup` again, with a future version that uses version 3 of `vue/compiler-sfc` instead of version 2.

↓

`kirbyup build src/index.js`

↓

```js
// /index.js (simplified)
(function(vue) {
  "use strict";
  var Demo = function _sfc_render(_ctx, _cache) {
    return vue.openBlock(), vue.createElementBlock("p", null, "Hello!");
  }
  panel.plugin("name", { components: { Demo } });
})(Vue);
```

Now that it's compiled for Vue 3, the plugin and its `Demo` component will work again, as long as `window.Vue` exists.
(and points to Vue 3, it'll not work in Vue 2 anymore)

distantnative:
@jonaskuske sorry that it's taken such a long time. I am slowly arriving at the idea and gave it a spin. The implementation looks quite sensible to me however I've been running in a few problems:

1. It looks like Vue gets loaded twice at least in development mode:
Screenshot 2023-04-16 at 18 51 56

2. I am getting errors when trying to open any dialog/overlag. Which lets me think that code might in general not function 1:1?
Screenshot 2023-04-16 at 18 52 16

3. in our v4 develop setup I'm seeing even more errors:
Screenshot 2023-04-16 at 18 55 33

jonaskuske:
Huh interesting, that shouldn't happen. Will investigate some time next week! :)

jonaskuske:
I guess the v4 setup isn't public yet, didn't find anything, but luckily I saw the `$attrs is readonly` errors in v3, too. Both those and the overlay errors disappear once Vue isn't loaded twice anymore.


Vite was pre-bundling `vue`, so dependencies like `@linusborg/vue-simple-portal` that directly import Vue themselves would be rewritten by Vite to something like:

```js
import { Vue } from '.vite/deps/aae7u234gskfug82fs.js'
``` 

Because of this, the import was not treated as a `vue` import that should be externalized but left as-is, so the bundled Vue was loaded in addition to the external one. After preventing `vue` from being pre-bundled through `optimizeDeps.exclude`, everything works :)

distantnative:
Seems to work great now! 💛
Also for v4 (which is indeed not public yet until the beta)

distantnative (approved):

jonaskuske:
Yay! 😊

Btw, should v4 move forward with Vue 3, the implementation will need one more addition. Instead of [L122](https://github.com/getkirby/kirby/pull/4779/files#diff-e2db9f4254a51c2f39fb8cbdf5efc5eeef15f924c2c31a0578c202802ae4143aR122) where the dev build of Vue is loaded as part of the `$isDev === true` instructions, this:

```php
if ($isDev === true || $assets['plugin-imports'] !== '') {
	$assets['js']['vue']['src'] = $url . '/node_modules/vue/dist/vue.js';
}
```

In Vue 2, plugins provided the HMR runtime themselves, in Vue 3 it's part of the development build. So for `kiryup` plugin HMR to work, the Panel needs to load the dev version of Vue not only when itself is in dev mode, but also when a plugin is in dev mode (`plugin-imports !== ''`)

distantnative:
That's a good heads-up. I still doubt though that there is a path for Vue 3 in the first release of Kirby 4 though. There are some path dependencies left that we first need to get out of the way (e.g. vuelidate)

bastianallgeier:
@distantnative do you think we can really put it in 3.9.4? I didn't have time to check out the details, but I would feel better to put it in 3.9.5 with a proper rc phase.

distantnative:
Ok

distantnative:
Manually rebased https://github.com/getkirby/kirby/pull/5158

Issue #4780, opened 2022-10-11
Picker fields: `store` description wrong


distantnative:
Should be `uuid` or `id` - or we need to alias `path`.

bastianallgeier:
✅

PR #4783, opened 2022-10-13
Fix `store` description of picker fields

afbora:
### Fixes
- Picker fields `store` prop description corrected #4780 

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/fields/mixins/picker.php
+++ b/config/fields/mixins/picker.php
@@ -69,7 +69,7 @@ return [
 		},
 
 		/**
-		 * Whether to store UUID or ID in the 
+		 * Whether to store UUID or ID in the
 		 * content file of the model
 		 *
 		 * @param string $store 'uuid'|'id'
--- a/config/fields/mixins/picker.php
+++ b/config/fields/mixins/picker.php
@@ -69,8 +69,8 @@ return [
 		},
 
 		/**
-		 * Whether to store UUID in the content file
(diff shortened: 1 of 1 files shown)


config/fields/mixins/picker.php
```diff
@@ -68,13 +69,13 @@
 		},
 
 		/**
-		 * Whether to store UUID or path in
-		 * the content file
+		 * Whether to store UUID in the content file
+		 * or ID of the model
```
bastianallgeier:
```suggestion
		 * Whether to store UUID or ID in the 
		 * content file of the model
```

bastianallgeier (dismissed):

config/fields/mixins/picker.php
afbora: Does the content file contain the ID data?

bastianallgeier (approved):

config/fields/mixins/picker.php
bastianallgeier: We could probably just remove the content file part, but what it means is that the UUID or ID is stored in the field in the content file. 

I.e. 

```
related: 

- blog/article-a
- blog/article-b
```

vs. 

```
related: 

- page://blabla
- page://blablabla
```

afbora: Thanks for reply. I got it what you mean 👍

PR #4790, opened 2022-10-17
Fixes generating `uuid` value when `content.uuid` option disabled

### LINKED ISSUE #4787: Draft pages include uuid when content.uuid option is false
## Description

I currently have `content.uuid` set to `false` in the Kirby config file. This works as promised per PR #4755 and does not add `Uuid:` fields to my existing content. (Thank you!)

However, if I create a new draft page in the panel, a `Uuid:` field is added to the content, even though the config is set to false.

**Expected behavior**  

When `'content.uuid' => false` do not add the `Uuid:` field to new draft pages.

## To reproduce

1. Go to `website/panel/site`
2. Click on ‘Add’
3. Fill in the required fields and click ‘Create draft”
4. Open the content file: `content/_drafts/title-of-draft-page/page.txt`
5. Observe the unwanted `Uuid:` field

## Your setup
**Kirby Version**  
3.8.1-rc.1

**Console output**  
n/a

**System**
- Device: MacBook Pro
- OS: macOS 10.13.6 High Sierra
- Browser: Vivaldi
- Version: 5.5.2805.35 (Chrome/106.0.5249.113)

afbora:
### Fixes
- Fixes generating `uuid` value for models when `content.uuid` option disabled #4787

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -155,7 +155,9 @@ trait FileActions
 		$copy = $page->clone()->file($this->filename());
 
 		// overwrite with new UUID (remove old, add new)
-		$copy = $copy->save(['uuid' => Uuid::generate()]);
+		if ($uuid = static::uuidGenerate()) {
+			$copy = $copy->save(['uuid' => $uuid]);
+		}
 
 		return $copy;
 	}
@@ -191,7 +193,7 @@ trait FileActions
 
 		// make sure that a UUID gets generated and
 		// added to content right away
-		$content['uuid'] = Uuid::generate();
+		$content['uuid'] ??= static::uuidGenerate();
 
 		// create a form for the file
 		$form = Form::for($file, ['values' => $content]);
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -8,6 +8,7 @@ use Kirby\Exception\LogicException;
 use Kirby\Filesystem\F;
 use Kirby\Form\Form;
 use Kirby\Uuid\Uuid;
+use Kirby\Uuid\Uuids;
 
 /**
  * FileActions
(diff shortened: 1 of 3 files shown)


src/Cms/ModelWithContent.php
```diff
@@ -637,6 +638,19 @@ public function uuid(): Uuid|null
 		return Uuid::for($this);
 	}
 
+	/**
+	 * Returns a new ID string if `content.uuid` option not disabled
+	 * @since 3.8.1
+	 */
+	public static function uuidGenerate(): string|null
```
bastianallgeier: The method name isn't ideal in my opinion. I would go for something like Uuid::generateIfEnabled to make it clearer what's happening here. I was thinking first if it should actually be implemented directly in Uuid::generate instead, but I guess we could still need this even if uuids are disabled, right?

afbora: > The method name isn't ideal in my opinion.

I was waiting for this comment 😅 I couldn't find a better name. I thought of a similar name `generateIfEnabled` but it felt like overkill 🙂 I'm open to all suggestions.

> I was thinking first if it should actually be implemented directly in Uuid::generate instead, but I guess we could still need this even if uuids are disabled, right?

Yes, this was discussed here https://github.com/getkirby/kirby/issues/4787#issuecomment-1279447162

distantnative:
Why are we complicating this? We can simply use `Uuid::enabled()` in the if statement. It's not really more code.

src/Cms/ModelWithContent.php
distantnative: Yes. I wrote in the issue why we should not disable the generate method itself.

afbora:
I've changed from separate `uuidGenerate` method to inline  `Uuids::enabled()` usage. I didn't squash the commits to you see diffs.

src/Cms/FileActions.php
```diff
@@ -191,7 +194,9 @@ public static function create(array $props)
 
 		// make sure that a UUID gets generated and
 		// added to content right away
-		$content['uuid'] = Uuid::generate();
+		if (Uuids::enabled() === true) {
+			$content['uuid'] ??= Uuid::generate();
+		}
```
lukasbestle, line 199: @distantnative Was there a particular reason why you used `??=` in `PageActions`, but `=` here? I think Ahmet's change to always use `??=` makes sense, just want to be sure.

distantnative: No, I think @afbora is right

bastianallgeier (approved):

PR #4800, opened 2022-10-21
Fix potential content loss during UUID generation

bastianallgeier:
## This PR …

This PR fixes various smaller problems in the error handling while files are being read or written. It also fixes an issue in model section components that lead to duplicate requests for pages and files sections. In addition to that, it handles auto-generation of UUIDs more carefully and already overwrites the model content in memory with the new UUID. 

### Fixes
- Fixed potential content loss when auto-generating UUIDs.

### Breaking changes

- None

## Ready?

### For review team


src/Cms/ModelWithContent.php
```diff
@@ -375,11 +375,17 @@ public function query(string $query = null, string $expect = null)
...
 	{
-		try {
-			return Data::read($this->contentFile($languageCode));
-		} catch (Throwable) {
+		$file = $this->contentFile($languageCode);
+
+		// only if the content file really does not exist, it's ok
+		// to return empty content. Otherwise this could lead to
+		// content loss in case of file reading exceptions
+		// (i.e. if the file is locked)
```
lukasbestle: A locked file does not throw an exception. `flock()` does "advisory locking", which means that access isn't prevented. Advisory locking only works if all processes use `flock()` and check for the lock.

src/Cms/ModelWithContent.php
```diff
@@ -375,11 +375,17 @@ public function query(string $query = null, string $expect = null)
...
+		if (file_exists($file) === false) {
 			return [];
 		}
+
+		return Data::read($file);
```
lukasbestle, line 387: The order of the method calls introduces another race condition. If the file is deleted right between `file_exists()` and `Data::read()`, the method will fail.

src/Cms/ModelWithContent.php
bastianallgeier: That's a shame. Then I need to change the doc block, but the fix is still super important. Even if there's just a permission problem or any kind of reading exception, the content would have been an empty array. I really wonder why this didn't hit us more often so far.

src/Cms/ModelWithContent.php
bastianallgeier: Ok, that's true and another potential problem, but not the cause of the UUID issue. I don't think we should fix more theoretical issues at this point. The problems that we have with this are already big enough on their own :)

lukasbestle: The fix for this is pretty simple. I'll push a fix over the weekend.

lukasbestle: Given how we set up `Handler::read()` and `F::read()`, it's not as simple as expected because there can be both PHP warnings and exceptions to take care of.

So I think we can ship it like this for now. It may generate more false-negative error messages, but at least there won't be undetected read errors.

lukasbestle (approved):
I've changed the following:

- Added a fix for existing but empty `Uuid` fields
- Updated the code comments
- Prepared the release commits (dist files, version number)
- Rebased against `main` so that the commits from `develop` will only end up in 3.8.2, not 3.8.1.1

During testing, I've discovered this issue: https://github.com/getkirby/kirby/issues/4802
Also pretty fatal unfortunately, but not a regression from this PR (same behavior in 3.8.1).

afbora (approved):
LGTM 👍

PR #4803, opened 2022-10-23
Fix UUID generation on multilang setup 

### LINKED ISSUE #4802: UUIDs get copied to and generated on secondary languages
## Description

If I create a new page in a multilang setup, I get a UUID in the primary language as expected. But once a translation is created by saving a secondary language, this UUID is then copied to the content file of the secondary language.

Also if I have a migrated page (without UUID) and access `$page->uuid()` when a secondary language is active, the UUID is generated and stored for the secondary language. The `Uuid` field of the primary language stays empty.

**Why this is fatal:** If a UUID gets generated in a secondary language, a *different* UUID may later be generated in the primary language. At this point the same model will have more than one UUID.

**Expected behavior**  

- UUIDs should only be generated for the primary language.
- All content files of secondary languages must not get the `Uuid` field (basically the behavior should be that of `translate: false`).
- When querying the content for the `Uuid` field, Kirby should always look at the primary language (even if a different value is stored in the secondary language).

### Copy UUID to secondary language

**Prerequisites:**

1. Enable multilang (here with `en` as default and `de` as secondary)
2. Create a new page (gets `.en.txt` content file, which contains a `Uuid` field)

**Reproduction:**

3. Switch to the secondary language in the Panel, change something and save the content
4. See that the `Uuid` field gets copied to the secondary language.

### Generate UUID on secondary language

**Prerequisites:**

1. Enable multilang (here with `en` as default and `de` as secondary)
2. Manually create a new page without the `Uuid` field
3. Put `$page->uuid()` into the template somewhere

**Reproduction:**

4. Access the page in the secondary language
5. See that the UUID is saved to the secondary language.

## Your setup
**Kirby Version**  

3.8.1

afbora:
## This PR …

Fixes UUID generation on non-default languages for:

✅ Auto generating
✅ Creating models as manually 
✅ Creating models from Panel UI

### Fixes

- Fixes UUID generation on multilang setup #4802 

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -925,14 +925,14 @@ class App
 		}
 
 		if ($code === 'default') {
-			return $this->languages()->default();
+			return $this->defaultLanguage();
 		}
 
 		if ($code !== null) {
 			return $this->languages()->find($code);
 		}
 
-		return $this->language = $this->language ?? $this->languages()->default();
+		return $this->language = $this->language ?? $this->defaultLanguage();
 	}
 
 	/**
--- a/src/Uuid/ModelUuid.php
+++ b/src/Uuid/ModelUuid.php
@@ -92,14 +92,8 @@ abstract class ModelUuid extends Uuid
 	 */
 	public static function storeId(Identifiable $model, string $id): Identifiable
 	{
-		// make sure Kirby has the required permissions
-		// for the update action
-		$kirby = $model->kirby();
-		$user  = $kirby->auth()->currentUserFromImpersonation();
-		$kirby->impersonate('kirby');
-
 		// get the content array from the page
-		$data = $model->readContent('default');
+		$data = $model->content('default')->toArray();
 
 		// check for an empty content array
 		// and read content from file again,
@@ -114,24 +108,16 @@ abstract class ModelUuid extends Uuid
 			$data['uuid'] = $id;
 		}
 
-		// overwrite the content in memory and in the content file;
-		// use the most basic write method to avoid object cloning
+		// overwrite the content in memory and in the content file
 		$model->content('default')->update($data);
+
+		// use the most basic write method to avoid object cloning
 		$model->writeContent($data, 'default');
 
 		// update the default translation
-		if ($kirby->multilang() === true) {
+		if ($model->kirby()->multilang() === true) {
 			$model->translation('default')->update($data);
 		}
-
-		// switch back to the original user
-		$kirby->impersonate($user);
-
-		// TODO: replace the above in 3.9.0 with
-		// App::instance()->impersonate(
-		// 	'kirby',
-		// 	fn () => $model->writeContent($data, $languageCode)
-		// );
 
 		return $model;
 	}
--- a/src/Uuid/ModelUuid.php
+++ b/src/Uuid/ModelUuid.php
@@ -63,50 +63,8 @@ abstract class ModelUuid extends Uuid
 		// generate a new ID (to be saved in the content file)
 		$id = static::generate();
 
-		// make sure Kirby has the required permissions
-		// for the update action
-		$kirby = $this->model->kirby();
-		$user  = $kirby->auth()->currentUserFromImpersonation();
-		$kirby->impersonate('kirby');
-
-		// if multilang enabled and current language is not default
-		// we store the UUID to the default language
-		$languageCode = null;
-		if (
-			$kirby->multilang() === true &&
-			$kirby->languageCode() !== $kirby->defaultLanguage()->code()
-		) {
-			$languageCode = $kirby->defaultLanguage()->code();
-		}
-
-		// get the content array from the page
-		$data = $this->model->content($languageCode)->toArray();
-
-		// check for an empty content array
-		// and read content from file again,
-		// just to be sure we don't lose content
-		if (empty($data) === true) {
-			usleep(1000);
-			$data = $this->model->readContent($languageCode);
-		}
-
-		// add the UUID to the content array
-		if (empty($data['uuid']) === true) {
-			$data['uuid'] = $id;
-		}
-
-		// overwrite the content in memory and in the content file;
-		// use the most basic write method to avoid object cloning
-		$this->model->content($languageCode)->update($data);
-		$this->model->writeContent($data, $languageCode);
-
-		$kirby->impersonate($user);
-
-		// TODO: replace the above in 3.9.0 with
-		// App::instance()->impersonate(
-		// 	'kirby',
-		// 	fn () => $this->model = $this->model->writeContent($data, $languageCode)
-		// );
+		// store the new UUID
+		$this->model = static::storeId($this->model, $id);
 
 		// update the Uri object
 		$this->uri->host($id);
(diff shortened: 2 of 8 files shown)


src/Uuid/ModelUuid.php
```diff
@@ -69,15 +69,31 @@ public function id(): string
...
 		// just to be sure we don't lose content
 		if (empty($data) === true) {
 			usleep(1000);
-			$data = $this->model->readContent();
+			$data = $this->model->readContent($languageCode);
+		}
+
+		// return UUID from the default language content if available
+		if ($isMultilang === true && empty($data['uuid']) === false) {
+			return $data['uuid'];
```
lukasbestle: Shouldn't this already be covered by `Uuid::__construct()`/`ModelUuid::retrieveId()`? Kirby has an automatic fallback to the default language content already.

lukasbestle (commented):
I wasn't able to test the PR yet, but I'd say the changes look pretty good already. They cover two of the three "expected behavior" items.

To cover the third, I think we should modify `ModelUuid::retrieveId()` to always access the default language content.

src/Uuid/ModelUuid.php
afbora: @lukasbestle Ahh, nice catch! If the secondary language is requested, calling `::retrieveId()` will already run `->id()` for the primary and the UUID will be generated and saved for primary. Did I understand the process correctly?

afbora: > To cover the third, I think we should modify ModelUuid::retrieveId() to always access the default language content.

I've tested and seems works great. Do you think still need this one?

lukasbestle: > If the secondary language is requested, calling `::retrieveId()` will already run `->id()` for the primary and the UUID will be generated and saved for primary. Did I understand the process correctly?

Yep. But that only works if there is no different UUID stored in the secondary language content file...

> Do you think still need this one?

...so I think we need the change in `retrieveId()`. The UUID cannot end up in secondary language content files programmatically with your change in `saveTranslation()`, but it could still end up there manually.

lukasbestle: PS: The call to `->id()` only happens in `Uuid::retrieveId()`, but this method is overridden in `ModelUuid::retrieveId()` with a retrieval from content. I think we only need to pass the default language code there.

afbora: Could you check out the commit https://github.com/getkirby/kirby/pull/4803/commits/09b9d314676318c0822a9783c6b760b14b70a27a

afbora: I couldn't figure out the CI issue about kirbytags: https://github.com/getkirby/kirby/actions/runs/3311545395/jobs/5467126691

lukasbestle: The change looks good to me.

I've fixed the unit test. The failure was caused by the UUID being stored in the translated content.

lukasbestle:
The changes look good to me. I don't have more time to test this manually and add unit tests unfortunately.

@bastianallgeier Could you please test the PR manually and confirm that the issues in #4802 can no longer be reproduced?
@afbora Could you please add three unit tests (one for `ModelWithContent` that tests that UUIDs are not stored for secondary languages and two for `ModelUuid` that tests the generation and retrieval of UUIDs from the primary language)?

bastianallgeier (approved):

bastianallgeier:
@lukasbestle could you have one last look just to make sure we don't miss anything?

lukasbestle:
I'd say manual tests are more important here. So if everything works, feel free to already merge and release this.

I briefly looked at the tests: I think two things are missing (but could be added for 3.8.2):

- Assertions in the `PageUuid` and `FileUuid` that query the raw content data for both en and de and verify that the UUID is only stored in en.
- The test for `ModelWithContent` (no UUID in translated content)

src/Cms/App.php
```diff
@@ -918,7 +918,7 @@ public function kirbytext(string $text = null, array $options = []): string
 	 * @param string|null $code
 	 * @return \Kirby\Cms\Language|null
 	 */
-	public function language(string $code = null)
+	public function language(string $code = null): ?Language
```
distantnative: Should be `Language|null` for consistency

distantnative:
I think we need at least one more unit test for something we fixed:

- [x] Calling `ModelWithContent::content()` by passing the default language code, updating that content, trying to retrieve the updated content by calling `ModelWithContent::content()` without a parameter

bastianallgeier:
@distantnative I added unit tests to check if updating the current language works as expected. After double-checking the problems @afbora discovered I realised that our caching logic for content in multi-language sites is actually different than what I thought. The current language's content object is cached. Not the default language. This makes a lot of sense, but was very difficult to understand from our code. 

This means that the content object for the default language and any other translation is always fetched from ModelWithContent::translation without being cached. Therefor, we cannot update content objects for inactive languages at all. We need to update the translation object. This is happening in the UUID::storeId method and works great. 

We have to refactor this entire code part at some point. I think caching on the model level is just plain-wrong and leads to so many issues. The caching has to happen in the content and translation objects. But this is another chapter.

src/Uuid/ModelUuid.php
```diff
@@ -63,21 +63,50 @@ public function id(): string
...
+	}
+
+	/**
+	 * Stores the UUID for the model and makes sure
+	 * to update the content file and content object cache
+	 *
+	 * @param \Kirby\Cms\ModelWithContent $model
+	 * @param string $id
+	 */
+	public static function storeId(Identifiable $model, string $id): Identifiable
```
lukasbestle: I feel like this method should be `protected`. Users shouldn't be able to store a new UUID because it would break references. Also I think it can be an instance method, then we won't need to pass `$model`.

src/Uuid/ModelUuid.php
```diff
@@ -63,21 +63,50 @@ public function id(): string
...
 		// make sure Kirby has the required permissions
 		// for the update action
-		$kirby = $this->model->kirby();
+		$kirby = $model->kirby();
 		$user  = $kirby->auth()->currentUserFromImpersonation();
 		$kirby->impersonate('kirby');
 
 		// get the content array from the page
-		$data = $this->model->content()->toArray();
+		$data = $model->readContent('default');
```
lukasbestle: Is it necessary to read the content from file here in the "first try"? Wouldn't this work as well:

```suggestion
		$data = $model->content('default')->toArray();
```

src/Uuid/ModelUuid.php
```diff
@@ -87,32 +116,24 @@ public function id(): string
...
+		}
 
+		// switch back to the original user
 		$kirby->impersonate($user);
 
 		// TODO: replace the above in 3.9.0 with
 		// App::instance()->impersonate(
 		// 	'kirby',
-		// 	fn () => $this->model = $this->model->writeContent($data)
+		// 	fn () => $model->writeContent($data, $languageCode)
```
lukasbestle:
```suggestion
		// 	fn () => $model->writeContent($data, 'default')
```

However I'm wondering if we even need impersonation at all (now that we use `writeContent()`, which doesn't check permissions). Removing all of the impersonation stuff would make the code much simpler.

src/Uuid/ModelUuid.php
```diff
@@ -87,32 +116,24 @@ public function id(): string
...
-	/**
-	 * Retrieves the ID string (UUID without scheme) for the model
-	 * from the content file, if it is already stored there
-	 *
-	 * @param \Kirby\Cms\ModelWithContent $model
-	 */
-	public static function retrieveId(Identifiable $model): string|null
-	{
-		return $model->content()->get('uuid')->value();
+		return $model;
```
lukasbestle: Do we need to return the `$model`? Since it is no longer cloned during the update, I feel like we can get away with a `void` return type.

lukasbestle (commented):
It's so much cleaner and more solid already. 💛
I can't test the changes manually right now, so I can only contribute a few thoughts on the code itself:

src/Uuid/ModelUuid.php
bastianallgeier: I adjusted the signature of the method to retrieveId. It could indeed be an instance method. I'm fine with both ways. But I think a public method is actually quite useful. Not necessarily for other users, but for us.

src/Uuid/ModelUuid.php
bastianallgeier: Uhhh, I haven't thought about that. You are totally right. I'll give it a try.

src/Uuid/ModelUuid.php
bastianallgeier: I think something's wrong in the diff here. It's not returning just the model. It's returning the uuid vale.

src/Uuid/ModelUuid.php
bastianallgeier: I had another look and I think you are right. By converting it to a protected method, we have more control and can rethink this method later if needed. If we make it public it immediately becomes a "feature" and if we want to change it it would be a breaking change. I also just converted it to non-static.

lukasbestle: We could also make it `public` with `@internal` later if we ever need it somewhere else.

src/Uuid/ModelUuid.php
lukasbestle: The `$model` is returned from `storeId()`, the `retrieveId()` method only appears in the diff because it was moved upwards.

src/Uuid/ModelUuid.php
```diff
@@ -63,56 +63,58 @@ public function id(): string
...
-
-		$kirby->impersonate($user);
-
-		// TODO: replace the above in 3.9.0 with
-		// App::instance()->impersonate(
-		// 	'kirby',
-		// 	fn () => $this->model = $this->model->writeContent($data)
-		// );
+		// overwrite the content in memory and in the content file
+		$this->model->content('default')->update($data);
```
lukasbestle: Isn't this the `else` case when `multilang` is disabled? Or is it needed also in `multilang`?

bastianallgeier: For some reason, the last comments always seem to point to outdated code examples. Could you tell me which line you mean?

bastianallgeier: Ah, now I get it. Let me check.

lukasbestle (approved):
Code looks good to me. As I wrote, cannot test it manually right now.

PR #4811, opened 2022-10-26
More unnecessary else statements

distantnative:


src/Http/Request/Body.php
```diff
@@ -46,15 +46,11 @@ public function __construct(array|string|null $contents = null)
...
-				$this->contents = $_POST;
-			} else {
-				$this->contents = file_get_contents('php://input');
-			}
+		if (empty($_POST) === false) {
+			return $this->contents ??= $_POST;
 		}
 
-		return $this->contents;
+		return $this->contents ??= file_get_contents('php://input');
```
afbora: IMHO, although the changes are short and correct, I think the readability of the code here is very low. I think we could simply write it like following. Maybe longer, but it's easier to understand that it has to look in the cache first.

````php
public function contents(): string|array
{
    if ($this->contents !== null) {
        return $this->contents;
    }

    if (empty($_POST) === false) {
        return $this->contents = $_POST;
    }

    return $this->contents = file_get_contents('php://input');
}
````

Comment (afbora) [reply src/Http/Request/Body.php]: Or seeing `??=` as one and on a single line also looks good for readability.

````php
public function contents(): string|array
{
    return $this->contents ??= empty($_POST) === false ? $_POST : file_get_contents('php://input');
}

Comment (distantnative) [reply src/Http/Request/Body.php]: Personally I find the first alternative a bit redundant, the one liner hard to read. I'd still stick to my suggestion. But let's get more opinions and settle for the most favorite one. These things are often a matter of taste and there we can go for the broadest one.

**Review by afbora — DISMISSED** [review]
LGTM 👍 except small improvements. You can merge as current if you want, not important.

File: src/Http/Request/Body.php (hunk 1 above)
Comment (lukasbestle) [reply src/Http/Request/Body.php]: I entirely agree with Ahmet and really like his first suggestion.

It's so much easier to read as the different logic alternatives have their own block, so mentally you can think about each alternative separately and don't need to consider all the ways how the logic could intertwine.

Also it makes testing much easier as we immediately know there are three cases to test (which can be confirmed with coverage).

Even though the PHP 8 syntax is really nice and useful in many places, I think we need to be super careful not to use it in places where it may even be detrimental and could cause hard to debug bugs in the future – just for a few saved lines that don't affect our users at all.

File: src/Cms/FilePicker.php [hunk 2]
```diff
@@ -41,13 +41,14 @@ public function items()
...
-			$query = $model::CLASS_ALIAS . '.files';
-		}
+		$query = match (true) {
+			empty($this->options['query']) === false
+				=> $this->options['query'],
+			$model instanceof File
+				=> 'file.siblings',
+			default
+			=> $model::CLASS_ALIAS . '.files'
+		};
```
lukasbestle, line 51: Does this instance of `match` really help? It's even one extra line of code and TBH I find it harder to read than `if elseif else` (considering that each `if` uses a completely different condition).

src/Cms/Media.php
```diff
@@ -97,16 +97,17 @@ public static function thumb($model, string $hash, string $filename)
...
+			// assets
+			is_string($model)
+				=> $kirby->root('media') . '/assets/' . $model . '/' . $hash,
+			// parent files for file model that already included hash
+			$model instanceof File
+				=> dirname($model->mediaRoot()),
+			// model files
+			default
+			=> $model->mediaRoot() . '/' . $hash
+		};
```
lukasbestle, line 110: Same here.

src/Cms/Media.php
```diff
@@ -117,11 +118,12 @@ public static function thumb($model, string $hash, string $filename)
...
-				$source = $kirby->root('index') . '/' . $model . '/' . $options['filename'];
-			} else {
-				$source = $model->file($options['filename'])->root();
-			}
+			$source = match (true) {
+				is_string($model) === true
+					=> $kirby->root('index') . '/' . $model . '/' . $options['filename'],
+				default
+				=> $model->file($options['filename'])->root()
+			};
```
lukasbestle, line 126: And here.

lukasbestle (commented):
All the other `match` uses in this PR are really awesome, but there are three where I find that `match` doesn't add value:

src/Http/Request/Body.php
distantnative: You are right 👍

src/Cms/FilePicker.php
distantnative: I think it does as it makes clear from the start that it's all about assigning different values to `$query`. In the `if else if else` statement, I need to look into each statement to see whether it also assigns `$query` or something else is done in there.

src/Cms/Media.php
distantnative: Same as above

src/Cms/Media.php
distantnative: Same as above

src/Cms/FilePicker.php
bastianallgeier: I agree with @distantnative. I think the nicest part for me about match is that I can quickly read it like a return switch once I understood it. It removes all those weird if else statements where you need to define a variable first just to replace the content in the if/else statement afterwards.

src/Cms/Media.php
bastianallgeier: I guess it's really a matter of taste, but I'm also on @distantnative's side here.

src/Cms/Media.php
bastianallgeier: Yep, same. I guess it comes down to a general team decision unless you are not super passionate about it @lukasbestle :)

distantnative:
@lukasbestle if you aren't super passionate about the three open comments, we could merge this I guess

bastianallgeier (approved):

bastianallgeier:
@lukasbestle ping

lukasbestle (approved):

PR #4819, opened 2022-10-27
Order-sensitive `v-bind` usage

distantnative:
Prepare for https://v3-migration.vuejs.org/breaking-changes/v-bind.html


bastianallgeier:
I would love this so much, but as far as I understand it, this would break stuff right now because the order is not right yet. Or am I missing something?

distantnative:
As far as I understand it, right now `:foo` takes always priority over `v-bind="{foo}"`. In Vue 3, it depends on the order. So we need to make sure that `v-bind` needs to go before everything else if we want to have the same result as currently. And as far as I understand it this shouldn't change any result currently.

bastianallgeier (approved):

PR #4820, opened 2022-10-27
Vue: restructure app setup

distantnative:
In anticipation of Vue 3:
- remove importing Vue globally as much as possible
- register e.g. component instead via Vue plugin that receives app instance in Vue 3 (in Vue 2.7 still global Vue)


bastianallgeier:
@distantnative could you have another look at the index.js and see if I fixed the conflict correctly?

bastianallgeier:
Shit. Looks like i made a mistake, but I'm not sure what's the correct way to do it.

distantnative:
@bastianallgeier fixed it

bastianallgeier (approved):

PR #4822, opened 2022-10-27
Remove `v-model`

distantnative:
### Refactoring
- Replaced all `v-model` occurrences with `:value="value" @input="value = $event"`, due to https://v3-migration.vuejs.org/breaking-changes/v-model.html

### Deprecated
- Usage of `v-model` on all Kirby components (and in general) has been deprecated. Use `:value="" @input=""` instead, e.g.
```vue
// before
v-model="myValue"

// now
:value="myValue"
@input="myValue = $event"
```
- Start migrating plugins, behaviour will be officially unsupported in Kirby 3.9
- During that time we should actively look through popular plugins and send PRs to replace `v-model`
- When switching to Vue 3.0 (ETA not known yet, but if then after Kirby 3.9) usage of `v-model` will break

## Ready?

### For review team


panel/src/components/Forms/Fieldset.vue
```diff
@@ -16,13 +16,16 @@
...
-							v-model="value[fieldName]"
+							:disabled="disabled || field.disabled"
 							:form-data="value"
 							:name="fieldName"
 							:novalidate="novalidate"
-							:disabled="disabled || field.disabled"
+							:value="value[fieldName]"
 							v-bind="field"
-							@input="$emit('input', value, field, fieldName)"
+							@input="
```
bastianallgeier: this looks to me like we should put it into an onInput method instead.

bastianallgeier (commented):
I like that it actually is easier to understand than the magic of v-model

panel/src/components/Forms/Fieldset.vue
distantnative: 👍

bastianallgeier (dismissed):

bastianallgeier (approved):

PR #4832, opened 2022-11-02
Fix duplicating UUIDs on duplicating pages 

### LINKED ISSUE #4831: Duplicating pages also duplicates file UUIDs
## Description
**Problem A:** 
If you duplicate a page, and activate 'copy files?', the files do not get assigned a new, unique ID.
Instead the file UUID gets copied and is no longer unique.
 This is especially bad when you replace a photo on the new page, and it affects the original page as well.

**Problem B:**
Also, not activating 'copy files?' used to generate a new page with only text content and no images. Now, it still copies the file-UUIDs from the original content file, and keeps the connection to the images from the original page.
This can be confusing for the user, because it is not obvious where those images are coming from. 

### Expected behavior

Problem A:
 Duplicating a page with files should generate new UUIDs for all files.

Problem B:
Duplicating a page without files should remove links to file UUIDs not belonging to the new page.

### To reproduce
1. Create a page with images
2. duplicate that page 
    A) with files and
    B) without files

## Your setup
**Kirby Version**  
3.8.1.1

afbora:
## This PR …

Fixes the following issues;

✅ Duplicating files uuid for duplicating pages
✅ Duplicating pages uuid for duplicating pages *

_* I've found the issue while testing the issue. Reproducible with duplicating a page while secondary lang enabled_

### Fixes

- Fixed duplicating UUIDs for pages and files on duplicating pages #4831

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -434,18 +434,38 @@ trait PageActions
 
 		$copy = $parentModel->clone()->findPageOrDraft($slug);
 
-		// remove all translated slugs
 		if ($this->kirby()->multilang() === true) {
+			// remove all translated slugs
 			foreach ($this->kirby()->languages() as $language) {
-				if ($language->isDefault() === false && $copy->translation($language)->exists() === true) {
+				if (
+					Uuids::enabled() === true &&
+					$language->isDefault() === true
+				) {
+					// overwrite with new UUID for the page and files
+					// for default language (remove old, add new)
+					$copy = $copy->save(['uuid' => Uuid::generate()], $language->code());
+
+					if ($files !== false) {
+						foreach ($copy->files() as $file) {
+							$file->save(['uuid' => Uuid::generate()], $language->code());
+						}
+					}
+				} elseif (
+					$language->isDefault() === false &&
+					$copy->translation($language)->exists() === true
+				) {
 					$copy = $copy->save(['slug' => null], $language->code());
 				}
 			}
-		}
-
-		// overwrite with new UUID (remove old, add new)
-		if (Uuids::enabled() === true) {
+		} elseif (Uuids::enabled() === true) {
+			// overwrite with new UUID for the page and files (remove old, add new)
 			$copy = $copy->save(['uuid' => Uuid::generate()]);
+
+			if ($files !== false) {
+				foreach ($copy->files() as $file) {
+					$file->save(['uuid' => Uuid::generate()]);
+				}
+			}
 		}
 
 		// add copy to siblings
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -28,6 +28,55 @@ use Kirby\Uuid\Uuids;
 trait PageActions
 {
 	/**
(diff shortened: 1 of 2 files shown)


afbora:
I may have written a little too much if-else. I may need your help for simplifying.

src/Cms/PageActions.php
```diff
@@ -434,18 +434,38 @@ public function copy(array $options = [])
...
+					Uuids::enabled() === true &&
+					$language->isDefault() === true
+				) {
+					// overwrite with new UUID for the page and files
+					// for default language (remove old, add new)
+					$copy = $copy->save(['uuid' => Uuid::generate()], $language->code());
+
+					if ($files !== false) {
+						foreach ($copy->files() as $file) {
+							$file->save(['uuid' => Uuid::generate()], $language->code());
```
afbora: I'm not sure that is there anything we need to do in the cached data/objects for updated files?

lukasbestle (commented):
> I may have written a little too much if-else. I may need your help for simplifying.

You could clean this up by moving the modification logic to a method `protected adaptCopy($copy)` or something along those lines. This method can then have separate `if` blocks for the different cases and immediately return once a case has been found and the changes were performed, so we won't need the `else`/`elseif` blocks.

I think the logic itself is correct.

src/Cms/PageActions.php
lukasbestle: Which caches do you mean in particular? I could well be wrong, but I think the object won't yet be cached anywhere because it was newly created (it's a new copy after all). If we *do* have a caching issue here, it will likely also affect the copied page object.

afbora: You're right. A fresh copy and the `::updateParentCollections()` method is working last. Then there is no problem 👍

afbora: Sometimes I get confused because of this immutable. I've one more question about that; 

````php
foreach ($copy->files() as $file) {
  $file->save(['uuid' => Uuid::generate()], $language->code());
}
````

When the `$file->save()` method is called, the file objects in `$copy->files()` are also updated, right?
So the `$copy` object now has file objects with new UUIDs, right?

bastianallgeier (dismissed):
Looks good to me! Do you want to provide the missing unit tests, or should I look into them?

src/Cms/PageActions.php
bastianallgeier: I really hope so. At least it should be like that if our code is correct. But to be honest, I would need to test it myself.

bastianallgeier:
@afbora do you want to have another look at the hopefully updated files collection or is it fine like this?

afbora:
Yes, I'd better test it manually. But if you or Lukas are sure about it, it's fine. I'm not very familiar with Kirby immutable stuff, I'm just not sure 🙂

bastianallgeier:
Let me try adding a unit test for it.

bastianallgeier:
Ok, that was easier than I thought. It's all looking good.

bastianallgeier (approved):

PR #4837, opened 2022-11-09
Kirby Template Components

bastianallgeier:
## This PR …

This PR introduces the new component engine for Kirby templates. We already had the layout plugin with a very similar approach, but never managed to get nesting right. This version is now able to support indefinitely nested components and also handles slots a lot more predictably. It feels ready for prime-time. 

### Features

Using components in your templates is pretty straight forward. 

```php
<?php component('article') ?>
  <?php slot('header') ?>
    <h1>This is the title</h1>
  <?php endslot() ?>

  <?php slot('body') ?>
    <p>This is some body text</p>
  <?php endslot() ?>

  <?php slot('footer') ?>
    This is the footer
  <?php endslot() ?>
<?php endcomponent() ?>
```

Right now, this would look for an article snippet: 

```php
<article>
  <header>
    <?= $slots->header() ?>
  </header>
  <?= $slots->body() ?>
  <footer>
    <?= $slots->footer() ?>
  </footer>
</article>
```

Component files are stored in `site/components`

### Refactoring

- The `Kirby\Cms\Template` class moved to `Kirby\Template\Template`

### Breaking changes

None

## Ready?

- [x] Unit tests for fixed bug/feature
- [x] In-code documentation (wherever needed)
- [x] Tests and checks all pass

### For review team

- [ ] Add changes to release notes draft in Notion
- [ ] Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)

PR #4839, opened 2022-11-11
Better attr() helper (and new friends) for templates and snippets

fabianmichael:
## Better attribute handling for snippets and templates

Whenever a snippet or component (see #4837) is included into a template, one often wants to pass additional variables and other parameters. Vue.js and Laravel blade templates provide means for merging the HTML attributes of components with ones fromt he outside. 

### Example

A button component exists as a snippet in `site/snippets/button.php`:

```php
<button class="button"><?= html($text ?? 'Button text') ?></button>
```

A common situation would be the requirement to add attributes when calling the `snippet('button')` helper class, e.g. `class`, `data-*`, `title`, `aria-*` etc. Developers cannot handly every possible attribute for each component. The previous `attr()` helper could help here:

```php
<button <?= attr($attr ?? []) ?> class="button"><?= html($text ?? 'Button text') ?></button>
```

This works better, but we still cannot extend the `class` attribute easily. Enter the new `attr()` helpers:

```php
<button <?= attr([
    'class' => 'button',
])->merge($attr ?? []) ?>><?= html($text ?? 'Button text') ?></button>
```

Even shorter:

```php
<button <?= classes('button')->merge($attr ?? []) ?>><?= html($text ?? 'Button text') ?></button>
```

This becomes even cooler, because the classes can be assigned conditionally as an array:

```php
<?php

$text = $text ?? 'Button text';
$size = $size ?? 'normal';
$theme = $theme ?? null;
$attr = $attr ?? [];

?>
<button <?= attr([
    'role' => 'button',
])->class([
    'button',
    "button--{$size}",
    "button--{$theme}" => $theme, // will only be merged, if $theme is trueish
])->merge($attr) ?>><?= html($text) ?></button>
```

This is already cool and makes working with attributes for snippets much easier, e.g. is we use the button in `site/snippets/menu.php`:

```php
<nav class="menu">
    […]
    <?php snippet('button', [
        'text' => 'Toggle Menu',
        'attr' => [
            'class' => 'menu__button',
            'aria-controls' => 'menu-popup',
            'aria-expanded' => false,
            'role' => 'teapot', // overrrides the default attribute
        ],
    ]) ?>
</nav>
```

We can even have additional attributes, that will just be merged instead of overridden, acting like the `class` attribute by wrapping their default value with the `prepends()` helper function:

```php
<?php

$text = $text ?? 'Button text';
$size = $size ?? 'normal';
$theme = $theme ?? null;
$attr = $attr ?? [];

?>
<button <?= attr([
    'role' => 'button',
    'style' => prepends('--data-text-length: ' . mb_strlen($text) . ';'), // don’t forget the trailing semicolon, values are always joined by a single space.
])->class([
    'button',
    "button--{$size}",
    "button--{$theme}" => $theme,
])->merge($attr) ?>><?= html($text) ?></button>
```

Now we can also extend the style attribute without overriding its complete value:

```php
<nav class="menu">
    […]
    <?php snippet('button', [
        'text' => 'Toggle Menu',
        'attr' => [
            'class' => 'menu__button',
            'aria-controls' => 'menu-popup',
            'aria-expanded' => false,
            'role' => 'teapot', // overrrides the default attribute
            'style' => '--menu-order: 5;', // this will be appended to the style attribute’s value
        ],
    ]) ?>
</nav>
```

If our snippet does not have any attributes by default, we can also add conditional spacing like in `Html::attr()` by calling the `before()` and `after()` methods:

```php
<button<?= attr($attr ?? [])->before(' ')><?= html($text ?? null) ?>></button>
```

### Fixes

This allows for much more flexibility attempting to work with components. The basic API provided by this plugin would possibly be deeper integrated into the core. For example, the `snippet()` helper could take an optional `$attr` parameter, that would default to an empty array `[]`, so `$attr ?? []` would not have to be called within each snippet manually.

I am also not sure, whether my placement and naming of the classes is optimal, but you are much better in structuring code than I am. If I can assist you any further, please let me know.

### Breaking changes

The `attr()` helper will now return a `Kirby\Toolkit\Attributes` object instead of a string. But casting the object to a string will return the same result, as the previous `attr()` helper.

## Ready?

### For review team

### DIFF
--- a/config/helpers.php
+++ b/config/helpers.php
@@ -33,11 +33,26 @@ if (Helpers::hasOverride('attr') === false) { // @codeCoverageIgnore
 	 * @param array|null $attr A list of attributes as key/value array
 	 * @param string|null $before An optional string that will be prepended if the result is not empty
 	 * @param string|null $after An optional string that will be appended if the result is not empty
-	 * @return string|null
+	 * @return \Kirby\Toolkit\Attributes
 	 */
-	function attr(array|null $attr = null, string|null $before = null, string|null $after = null): string|null
+	function attr(?array $attr = null, ?string $before = null, ?string $after = null): Attributes
 	{
-		return Html::attr($attr, null, $before, $after);
+		return Attributes::from($attr)
+			->before($before)
+			->after($after);
+	}
+}
+
+if (Helpers::hasOverride('classes') === false) { // @codeCoverageIgnore
+	/**
+	 * Returns the result of a collection by name
+	 *
+	 * @param string $classes
+	 * @return \Kirby\Toolkit\Attributes
+	 */
+	function classes(array|string $classes = []): Attributes
+	{
+		return Attributes::from()->class($classes);
 	}
 }
 
@@ -476,6 +491,20 @@ if (Helpers::hasOverride('params') === false) { // @codeCoverageIgnore
 	function params(): array
 	{
 		return App::instance()->request()->url()->params()->toArray();
+	}
+}
+
+if (Helpers::hasOverride('prepends') === false) { // @codeCoverageIgnore
+	/**
+	 * Generates a list of HTML attribute whose value will be merged
+	 * instead of replaced, when merging its value.
+	 *
+	 * @param mixed $value A list of attributes as key/value array
+	 * @return \Kirby\Toolkit\Attribute
+	 */
+	function prepends(mixed $value): Attribute
+	{
+		return Attribute::prepends($value);
 	}
 }
 
--- a/src/Toolkit/Attribute.php
+++ b/src/Toolkit/Attribute.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace Kirby\Toolkit;
+
+/**
+ * The `Attribute` class represents the value of an HTML attributes
+ * and provides various utilities for merging and joinging attribute
+ * values.
+ *
+ * @package   Kirby Toolkit
+ * @author    Fabian Michael <hallo@fabianmichael.de>
+ * @link      https://fabianmichael.de
+ * @copyright Bastian Allgeier
+ * @license   https://opensource.org/licenses/MIT
+ */
+class Attribute
+{
+	protected mixed $value;
+	protected bool $prepends = false;
+
+	protected function __construct(mixed $value, bool|null $prepends = null)
+	{
+		if (is_a($value, static::class)) {
+			// If an instance of this class is passed as value,
+			// extract all informartion and transfer to this instance.
+			$this->value = $value->value();
+			$prepends = $value->prepends;
+		} else {
+			$this->value = $value;
+		}
+
+		$this->prepends = $prepends ?? false;
+	}
+
+	/**
+	 * Creates a new instance from an arbitrary value. Passing an
+	 * instance of this class to this method will return that
+	 * instance instead.
+	 */
+	public static function from(mixed $value = null): static
+	{
+		return new static($value, false);
+	}
+
+	/**
+	 * Creates an instance of this class, whose value will be merged
+	 * instead of repleat
+	 */
+	public static function prepends(mixed $value = null): static
+	{
+		return new static($value, true);
+	}
+
+	/**
+	 * If an attribute is merged with a new value, this will either
+	 * return a new instance with the new value
+	 */
+	public function merge(mixed $attribute): static
+	{
+		if ($this->prepends) {
+			$append = (string) $attribute;
+
+			if (!empty($append) && strlen($append) > 0) {
+				return static::prepends($this->value . r(!empty($this->value), ' ') . $append);
+			}
+		}
+
+		return new static($attribute);
+	}
+
+	/**
+	 * Returns the raw value.
+	 */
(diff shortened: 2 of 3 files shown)


distantnative:
I haven't had the chance to look into it in detail but some first thoughts that came up:

1. If, this should go into the new upcoming Template namespace. For Toolkit, I feel like it's too specific to a type of attribute (HTML) that the name can be confusing.
2. Not a fan of an additional helper such as `prepend` with such a generic name that then only relates to attributes.
3. Is this quite Armbandes but rather hidden functionality that goes way beyond what should be in the core? Here I don't really have an opinion yet. But it's a thought whether this needs to be in the core or better a plug-in just used by a few users.

fabianmichael:
hi @distantnative, thanks for your feedback.


> If, this should go into the new upcoming Template namespace. For Toolkit, I feel like it's too specific to a type of attribute (HTML) that the name can be confusing.

You are completely right, that makes a lot of sense. I just did not know where to put it, since I pulled it out the codebase one of my projects.

> Not a fan of an additional helper such as prepend with such a generic name that then only relates to attributes.

I Algo greem, I basically see the following options for dealing with that:

1. In my project, I named the `attr()` helper `attributes()`, so I don’t have to deal with manually setting constants. The `prepends()` helper is named `attribute_prepends()` there, which does not fit in with the other helpers, but would save us from having a function with such a misleading name.
2. Register an `attr` class in the global namespacing and define static methods like `attr::prepends()`. Also not ideal, since we would not have a class that does not fit in with our namespaces (reminds me a bit of the the Kirby 2 days back then …)
3. The prepending behavior is only ever used by the "class" attribute by default and opt-in for all other attributes, so we could expect devs to call `\Kirby\Template\Attribute::prepends()` if they ever need to use it. We could extend the behavior to all HTML attributes where we think this would make sense as well (probably only `style`, but with `;` as separator), so most people will never have to touch it at all.
4. Provide a method on the `Attributes` class. If this would be baked directly into the `snippets` and `components` core components (no pun intended), we could also consider to provide an instance of the class to every snippet and component, e.g. `$attributes` or `$attr` (that’s what Blade basically does):

```php
# site/components/button.php

<button <?= $attr->merge([
  'class' => 'button',
  'style' => $attr->prepends('--some-property: 123;'),
]) ?>>[…]</button>
```

By using PHP’s [`__invoke()`](https://www.php.net/manual/de/language.oop5.magic.php#object.invoke) magic method, the Syntax could even be shortened


```php
# site/components/button.php

<button <?= $attr([
  'class' => 'button',
  'style' => $attr->prepends('--some-property: 123;'),
]) ?>>[…]</button>
```

But this would rather deeply intervene with Kirby’s logic, so it should be considered well.

> Is this quite Armbandes but rather hidden functionality that goes way beyond what should be in the core? Here I don't really have an opinion yet. But it's a thought whether this needs to be in the core or better a plug-in just used by a few users.

For me personally, this solves some of the biggest headaches, that I alwas had when working with components in Kirby and is based upon my experience when working with Vue.js, Laravel Blade and Custom Elements. Everytime I am finding myself in a situation where I have to split my design/code into components, I am facing the same challenges. Here is an excerpt of the code for a text input snippet:

```php
<?php

$type = $type ?? 'text';
$error = $error ?? false;
$class = $class ?? '';
$value = $value ?? '';

?>
<input class="input<?= r($type === 'password', ' js-form-password') . r($error, ' is-error') ?><?= !empty($class) ? ' ' . html($class) : '' ?>" value="<?= html($value) ?>">
```

This could also be written as:


```php
<input <?= attr([
  'value' => $value ?? '',
  'type' => ($type = $type ?? 'text'),
])->class([
  'input',
  'is-error' => $error ?? false,
  'js-form-password' => ($type === 'password')
])->merge($attr ?? '') ?>>
```

Of course, the code above could probably be written nicer without any additional features in core, but I think you get the idea. As we are moving towards component-based design and development, I really think we need a way of addressing this within Kirby.

However, if you decide against following this idea any further for now, I am also happy to release it as a plugin so it can mature for now.

lukasbestle:
I agree that more powerful helpers to work with attributes can often be useful. However I also agree with Nico that the helpers in their proposed form add too much complexity to the core:

- The current names of the two new helpers are too generic IMO and prefixing the names with `attr` or similar is not very clean either. Naming things is hard, but I feel like additional helpers will always add confusion (no matter how they are named).
  - I'd say the `prepends()` helper feels too much like magic. Maybe you can find a way to integrate the functionality directly into the `$attr` object structure.
  - The `classes()` helper feels redundant to me, like just a shortcut.
- Which brings me to another point: I don't fully understand the entirety of features yet, but from your description it looks like your implementation provides several different ways to achieve the same result. This adds a lot of mental complexity and also makes it hard to document and support the features properly. We generally prefer to have one way that is as clear and clean as possible.

The issue with merging this into the core is that we would need to keep backwards compatibility for any future changes (unless we deprecate old behavior, which is often hard to do). IMO this is not yet ready for prime time. I agree with you and Nico that you should first let the concept mature in a plugin. We can reconsider this for the core at a later time.

PR #4855, opened 2022-11-19
Structure field: delete all button

afbora:
## This PR …

screen-capture-1480-Test - Mægazine-localhost

### 🎉  Features

- Structure field: delete all button https://kirby.nolt.io/458

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -290,6 +290,7 @@
 	"field.pages.empty": "No pages selected yet",
 
 	"field.structure.delete.confirm": "Do you really want to delete this row?",
+	"field.structure.delete.confirm.all": "Do you really want to delete all entries?",
 	"field.structure.empty": "No entries yet",
 
 	"field.users.empty": "No users selected yet",
--- a/panel/src/components/Forms/Field/StructureField.vue
+++ b/panel/src/components/Forms/Field/StructureField.vue
@@ -1,27 +1,21 @@
 <template>
 	<k-field v-bind="$props" class="k-structure-field" @click.native.stop>
-		<!-- Add button -->
 		<template #options>
-			<k-button-group>
-				<k-button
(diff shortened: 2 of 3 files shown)


bastianallgeier:
I think I would go for the same dropdown as the blocks field here.

afbora:
@bastianallgeier I thought about it but gave up. Because after adding a block we don't have to use dropdown to add the next blocks. But for the structure field the developer will always have to click the dropdown two times.

bastianallgeier:
I think we could do the same for the structure field: https://github.com/getkirby/kirby/pull/4858

- The plus button is visible below the last row as soon as the field is not empty and when more items can be added
- We replace the add button at the top with a dropdown and include the delete all option there. 

For future releases, we should also add "Insert before" and "Insert after" options to the row dropdowns. This would be aligning the features with blocks and layouts and could be super useful for the structure fields as well.

afbora:
@bastianallgeier I've updated the PR

screen-capture-1480-Test - Mægazine-localhost

bastianallgeier:
This is cool. I just realised that we have no proper solution as soon as pagination appears. I'm afraid we have to move it forward to the next milestone

afbora:
Where do you think we should put the add button? Now currently like following with this PR:

captures_chrome-capture-2022-11-8 (1)

afbora:
An idea 💡 Did you like it?

captures_chrome-capture-2022-11-8 (2)

afbora:
Could you give me an idea please? @getkirby/kirby-staff 😅

afbora:
Otherwise, I suggest to removing `Add` feature for the PR. We can implement just  the `Delete all` feature 🙂

bastianallgeier:
Your design idea was brilliant. I just moved it around …
Screenshot 2022-12-13 at 14 22 33

bastianallgeier (approved):

PR #4857, opened 2022-11-19
New hooks

afbora:
### 🎉 Features
- New hooks: https://kirby.nolt.io/419
  - `page.render:before|after`: The page template is triggered during rendering and the output/html can be modified.
  - `page.visit:before|after`: It is triggered when the page is visited.
  - `language.create:before|after`: It is triggered when a new language is created.
  - `language.update:before|after`: It is triggered when the language is updated.
  - `language.delete:before|after`: It is triggered when the language is deleted.
  - `cache.flush:after`: It is triggered when any type (pages, etc..) of cache is flushed.

### Breaking changes
None

## Ready?

### For review team


afbora:
A pre-review would be great before a detailed review please 🙏

bvdputte:
Oh man, a cache clear hook 😍🤤
Been longing for this for years: https://forum.getkirby.com/t/cache-clear-hook/15999/7

Would the `page.visit` hooks work when cache is enabled? 🤔

afbora:
> Oh man, a cache clear hook 😍🤤

It may be a little early to celebrate 😅 Because there may be issues that I cannot foresee. We have to wait for the team's review 🤞 Sometimes things can be more complicated than we expected. This may mean delaying the features for a while.

> Would the page.visit hooks work when cache is enabled? 🤔

Nice question! Answer is `no` for now. Because `$page->visit()` method only run after the processes: `$page->render() > $page->controller() > $page->visit()` and cache control in the `$page->render()` method. I'm not sure but maybe it is necessary to move the `page.visit` hook into routes (or more correct place) instead of `site::visit()` method.

src/Cms/Site.php
```diff
@@ -640,9 +640,15 @@ public function visit($page, string|null $languageCode = null)
...
 
+		// trigger before hook
+		$this->kirby()->trigger('page.visit:before', compact('page', 'languageCode'));
+
 		// set the current active page
 		$this->setPage($page);
 
+		// trigger after hook
+		$this->kirby()->trigger('page.visit:after', compact('page', 'languageCode'));
+
```
afbora, line 651: Maybe we should move the `page.visit` hook to the beginning and end of the `page::render()` method instead of `site::visit()`

afbora: It will not work for cached page when we trigger it in `site::visit()`.

distantnative: Depending on the cache type (e.g. staticache), this won't work either way. So maybe it's better not expect this hook to work for cached pages.

afbora: Understood. What if we put it on the routes? Do you think it makes sense?

bastianallgeier (commented):
I think we should split this PR. 

- The language hooks are a no-brainer and I would instantly add them.
- The cache hooks are fantastic but it feels wrong to embed them so deep in the cache drivers. 
- The page.render hooks are great
- The page.visit hooks seem tricky. I think it could work to add it to the route, but I'm not 100% sure.

afbora:
Fantastic! Let's separate what we agree on. When the tasks in this PR are completed, we delete this PR.

afbora:
Separate PRs

  - [x] `page.render:before|after`: https://github.com/getkirby/kirby/pull/4862
  - [x] `language.*.before|after` https://github.com/getkirby/kirby/pull/4861
  - [x] ~`cache.flush:after`~: this hook is canceled/delayed https://github.com/getkirby/kirby/pull/4857#issuecomment-1328110367
  - [ ] `page.visit:before|after`

afbora:
> The cache hooks are fantastic but it feels wrong to embed them so deep in the cache drivers.

@bastianallgeier Yes, I agree. But to flushing, the `->flush()` method of each driver is called. We need a method/function (like finally in try/catch/finally) that always runs after the `->flush()` method is called, but I don't know if there is such a feature in PHP. Do you have an idea? @lukasbestle @distantnative

afbora:
> The page.visit hooks seem tricky. I think it could work to add it to the route, but I'm not 100% sure.

Where is the right place? How about we put it in `$kirby->resolve()` method? Because routes call this method.

src/Cms/Site.php
lukasbestle: I agree with Nico that the hook shouldn't be relied on for cached pages. After all the purpose of the cache is to speed up processing as much as possible. Also, they still won't be called if Staticache is used (no matter if we put them in the routes or keep them here).

Which use case do you see for the `page.visit` hook?

lukasbestle:
> But to flushing, the `->flush()` method of each driver is called. We need a method/function (like finally in try/catch/finally) that always runs after the `->flush()` method is called

I think in the long run we should abstract away the cache drivers as much as possible (e.g. by switching to the [PSR-16](https://www.php-fig.org/psr/psr-16/) interface). This would make it a lot easier to extend the caching system. Also cache drivers wouldn't have to implement all shared custom logic that we need for Kirby. Such logic would end up in a "caching middleware" in Kirby. All code in Kirby would no longer call methods in the driver directly, but instead in the middleware.

At the moment it's pretty much impossible to implement the `cache.flush` hook because it would be a breaking change for cache driver plugins (all plugins would have to implement it as well).

src/Cms/Site.php
afbora: > Which use case do you see for the page.visit hook?

For example; For updating or processing a data via the plugin when the page is visited.

lukasbestle: Can that work at all when the cache is active?

afbora: No, but it will work if we put it in routes. Also it will not work when there is a static cache. But when there is static cache, nothing works anyway. Controllers, models, routes, etc. So it's not just `page.visit`.

lukasbestle: That's right. 👍

afbora:
Understood, thanks for the explanation 👍  

Yes, my logic is pretty rough. The `middleware` logic makes a lot of sense. I think it would be great to refactor not just the cache but maybe a lot of parts in Kirby with the middleware pattern (may be K4 or K5 :))

src/Cms/Site.php
afbora: So where do you think the `page.visit` hook should be triggered? What do you think about having it in `App::resolve()`?

distantnative:
@afbora shall we put the `page.visit` hook in its own issue to discuss a potential implementation? then we could wrap-up/close this PR

afbora:
Closing favor of https://github.com/getkirby/backlog/issues/83

PR #4858, opened 2022-11-21
Feature: add plus button to blocks field

wiegertschouten:
I already posted https://kirby.nolt.io/487 last week, but then I decided to go ahead and give it a try myself. The reasoning behind it (as posted on Nolt) is the following:

> The blocks field currently has no directly clickable button to add blocks. You have to either click the three dots at the top, or select a block and then click the plus sign. This is not always intuitive. It’s also not consistent with the layout field, which does have a little plus button at the bottom.

I basically copied the add button from the layout field to the blocks field. I did not change any CSS. This way, it's nicely consistent with the layout field. It looks like this:

image

### Enhancements
Adds a plus button to the blocks field. When there are no blocks, or when the maximum amount of blocks is reached, the button is not visible.

### Breaking changes
None as far as I am aware.

## Ready?
Since I do not know a lot about automated testing on the front end, I have not created any tests. I have run the `format` and `lint` commands.

### For review team


afbora:
The PRs can't be includes build files. Could you remove the `dist` files (panel/dist/js/index.js) from changes, please?

wiegertschouten:
Done. Not sure how this works though, the dist files are not in the `.gitignore`. How should I do this in the future?

bastianallgeier (approved):
This is great. I moved the styles to the field component. But the rest is good to go. Thanks for the contribution. Next time just don't build the dist files at all. We do that during CI and for the final build. `npm run dev` is enough while working on a feature.

wiegertschouten:
Great, thanks for considering (and merging) the change, I feel a bit proud right now ☺️

Is there documentation on how to build and debug panel assets? I found the `npm run dev` script in the `/kirby/panel` folder, but running it and then visiting `localhost:3000` as indicated from the terminal returns a 404. So I ran `npm run build` every time, and then deleted the cached assets after every build 🙈 Not a very efficient workflow, but I couldn't find out how to do it otherwise. Also explains the presence of the `dist` files. Any directions?

Issue #4838, opened 2022-11-10
Select field has no initial value in structure/object fields


medienbaecker:
## Description

When using a [select](https://getkirby.com/docs/reference/panel/fields/select) field inside of a [structure](https://getkirby.com/docs/reference/panel/fields/structure) or [object](https://getkirby.com/docs/reference/panel/fields/object), the field doesn't initially have a value — compared to using it directly in the page blueprint.

### Why is this bad?

Let's say you're using a select field and other fields react to it via field conditions:

```yml
select:
  options:
    - Option A
    - Option B
    - Option C
text:
  when:
    select: ""
```

The text field is not shown initially. Only when changing the select field to some option and back to nothing it's shown. Interestingly enough it also works by saving the empty value and opening the structure/object again. I tried to set `default: ""` but it didn't work either.

![select-when](https://user-images.githubusercontent.com/7975568/201051459-77bd2e3d-acd7-401d-8908-56e15bd3d932.gif)

**Expected behavior**  

I'd expect the select field to have an initial value of `""` just like it does outside of structure or object fields.

## To reproduce

1. Add a structure or object field to your blueprint
2. Add a select field to the structure/object field
3. Add a field with a when condition for the select field to the structure/object

```yml
structure:
  fields:
    select:
      options:
        - Option A
        - Option B
        - Option C
    text:
      when:
        select: ""
# or:
object:
  fields:
    select:
      options:
        - Option A
        - Option B
        - Option C
    text:
      when:
        select: ""
```

## Your setup
**Kirby Version**  
3.8.1.1

afbora:
### Summary

Because when the page is created, the value of the empty field is empty string `" "`, while the value of a field is `null` when creating an entry in the structure or object field.

afbora:
I'm not sure that this idea could be break somethings but I've a solution. We can set empty string as fallback value like that

````js
const value = values[key.toLowerCase()] ?? "";
````

Related line:

https://github.com/getkirby/kirby/blob/3e5899b628a2b59c1833a8d79eda292a7fe06843/panel/src/helpers/field.js#L21

lukasbestle:
@afbora Does the issue only affect the structure field or do other fields behave the same inside a structure or object field?

afbora:
@lukasbestle for structure, object field and for each field containing the form (not sure).

The source of the issue: the field values are empty string when the page is created. But the field values in the form (with in structure and object field) created by the fields are assigned as `null`.

[...]

PR #4859, opened 2022-11-21
Improve the `isVisible()` helper

afbora:
## This PR …

Since the value `field.when` conditions can never be `null`, I don't mind converting it to an empty string always.

### Fixes
- Select field has no initial value in structure/object fields #4838

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/helpers/field.js
+++ b/panel/src/helpers/field.js
@@ -18,7 +18,7 @@ export function isVisible(field, values) {
 	let result = true;
 
 	Object.keys(field.when).forEach((key) => {
-		const value = values[key.toLowerCase()];
+		const value = values[key.toLowerCase()] ?? "";
 		const condition = field.when[key];
 
 		if (value !== condition) {


bastianallgeier:
Are you sure that it can never be null?

afbora:
How can it possible? Does blueprint supports `null` like following?

````yaml
text:
  type: text
  when:
    category: null
````

**Comment by bastianallgeier** [comment]
I think yaml actually does. But I would need to check that.

**Comment by lukasbestle** [comment]
This example:

```php
<?php

$yaml = <<<YAML
text:
  type: text
  when:
    category: null
YAML;

var_dump(Data::decode($yaml, 'yaml'));
```

gives me:

```text
test.php:10:
array (size=1)
  'text' => 
    array (size=2)
      'type' => string 'text' (length=4)
      'when' => 
        array (size=1)
          'category' => null
```

distantnative:
Based on this, the change of the PR is probably not the right solution for the linked issue (or rather for the linked issue it would work, but could create more problems elsewhere for us).

afbora:
I'm closing the PR and I'll re-work on this issue 👍

PR #4861, opened 2022-11-23
New hooks: `language.*:before|after`

afbora:
### 🎉 Features

- New language hooks: https://kirby.nolt.io/419
  - `language.create:before|after`: It is triggered when a new language is created.
  - `language.update:before|after`: It is triggered when the language is updated.
  - `language.delete:before|after`: It is triggered when the language is deleted.

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -211,9 +211,9 @@ class Language extends Model
 		}
 
 		// trigger before hook
-		$kirby->trigger('language.create:before', compact('props'));
-
 		$language = new static($props);
+
+		$kirby->trigger('language.create:before', ['language' => $language, 'input' => $props]);
 
 		// validate the new language
 		LanguageRules::create($language);
@@ -225,10 +225,10 @@ class Language extends Model
 		}
 
 		// update the main languages collection in the app instance
-		App::instance()->languages(false)->append($language->code(), $language);
+		$kirby->languages(false)->append($language->code(), $language);
 
 		// trigger after hook
-		$kirby->trigger('language.create:after', compact('language', 'props'));
+		$kirby->trigger('language.create:after', ['language' => $language, 'input' => $props]));
 
 		return $language;
 	}
@@ -679,7 +679,7 @@ class Language extends Model
 		// trigger before hook
 		$kirby->trigger('language.update:before', [
 			'language' => $this,
-			'props' => $props
+			'input' => $props
 		]);
 
 		// convert the current default to a non-default language
@@ -713,7 +713,7 @@ class Language extends Model
 		$kirby->trigger('language.update:after', [
 			'newLanguage' => $language,
 			'oldLanguage' => $this,
-			'props' => $props
+			'input' => $props
 		]);
 
 		return $language;
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -210,6 +210,9 @@ class Language extends Model
 			$props['default'] = true;
 		}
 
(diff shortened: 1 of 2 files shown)


src/Cms/Language.php
```diff
@@ -210,6 +210,9 @@ public static function create(array $props)
 			$props['default'] = true;
 		}
 
+		// trigger before hook
+		$kirby->trigger('language.create:before', compact('props'));
+
 		$language = new static($props);
```
bastianallgeier: I think we should use the same parameters as we do in Page::create. 

```suggestion
		$language = new static($props);

		$kirby->trigger('language.create:before', ['language' => $language, 'input' => $props]);
```

src/Cms/Language.php
```diff
@@ -224,6 +227,9 @@ public static function create(array $props)
 		// update the main languages collection in the app instance
 		App::instance()->languages(false)->append($language->code(), $language);
```
bastianallgeier: We already have `$kirby` and we could clean this up. 

```suggestion
		$kirby->languages(false)->append($language->code(), $language);
```

src/Cms/Language.php
```diff
@@ -224,6 +227,9 @@ public static function create(array $props)
...
 
+		// trigger after hook
+		$kirby->trigger('language.create:after', compact('language', 'props'));
```
bastianallgeier: Same as above. It should get the language and the input – not props.

```suggestion
		$kirby->trigger('language.create:after', ['language' => $language, 'input' => $props]));
```

src/Cms/Language.php
```diff
@@ -660,6 +676,12 @@ public function update(array $props = null)
 		// validate the updated language
 		LanguageRules::update($updated);
 
+		// trigger before hook
+		$kirby->trigger('language.update:before', [
+			'language' => $this,
+			'props' => $props
```
bastianallgeier:
```suggestion
			'input' => $props
```

src/Cms/Language.php
```diff
@@ -687,6 +709,13 @@ public function update(array $props = null)
 		// make sure the language is also updated in the Kirby language collection
 		App::instance()->languages(false)->set($language->code(), $language);
 
+		// trigger after hook
+		$kirby->trigger('language.update:after', [
+			'newLanguage' => $language,
+			'oldLanguage' => $this,
+			'props' => $props
```
bastianallgeier:
```suggestion
			'input' => $props
```

src/Cms/Language.php
```diff
@@ -225,10 +225,10 @@ public static function create(array $props)
 		}
 
 		// update the main languages collection in the app instance
-		App::instance()->languages(false)->append($language->code(), $language);
+		$kirby->languages(false)->append($language->code(), $language);
 
 		// trigger after hook
-		$kirby->trigger('language.create:after', compact('language', 'props'));
+		$kirby->trigger('language.create:after', ['language' => $language, 'input' => $props]));
```
bastianallgeier: Yikes, this happens when I use the code suggestion feature here. There are now additional parenthesis. I guess it's easier to fix this locally.

afbora: Actually I already fixed. This line should be outdated. Where do you found this?

bastianallgeier: Oh great. Then it was just Github showing the outdated version.

bastianallgeier (approved):

PR #4871, opened 2022-11-28
Fixes duplicating UUIDs for children and its files

### LINKED ISSUE #4864: When duplicating a page with subpages in the panel, the newly created subpages don't have new UUIDs
## Description
When using the panel to duplicate a page with subpages, the new subpages keep the UUIDs of the pages they were copied from, instead of having new ones generated

**Expected behavior**  
I would expect all the new subpages to have new UUIDs generated

## To reproduce

1. Create a new page, and create some subpages, in the panel
2. Click on the 3 dot menu for the new parent page and select "Duplicate"
3. Check the UUIDs of the subpages in the newly duplicated page, they will be the same as the originals

## Your setup
**Kirby Version**  
3.8.2

 - Device: Mac mini 2018
 - OS: 13.0.1
 - Browser: Firefox
 - Version: 107.0

afbora:
### Fixes

- Fixes duplicating UUIDs for children and its files #4864

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -53,6 +53,7 @@ trait PageActions
 					// regenerate UUIDs of all page children
 					if ($children !== false) {
 						foreach ($copy->index(true) as $child) {
+							// no $children because we already operate on the index
 							$this->adaptCopy($child, $files);
 						}
 					}
@@ -84,6 +85,7 @@ trait PageActions
 			// regenerate UUIDs of all page children
 			if ($children !== false) {
 				foreach ($copy->index(true) as $child) {
+					// no $children because we already operate on the index
 					$this->adaptCopy($child, $files);
 				}
 			}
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
Looks really good. 👍

However I think we'll have a performance issue when the children are nested more than one layer deep: We call `$page->index()` and on each result we recursively call `adaptCopy()` again, which in turn calls `$page->index()`. So grandchildren and grand-grandchildren are adapted multiple times.

We can either replace the `index()` with `childrenAndDrafts()` or pass `false` as the `$children` argument for the recursive `adaptCopy()`. Would be interesting to see if one variant is faster than the other. I'd expect the second option to be faster because it doesn't need to keep the whole index in memory, but you never know. ;)

afbora:
Oh, nice catch, thanks! 👍 
I've updated the PR with your second suggestion.

lukasbestle:
Could you please add a short code comment that explains why `$children` is not passed? Something like `no $children because we already operate on the index`.

lukasbestle:
I think the code looks good, but I couldn't test it so far.

bastianallgeier (approved):
Let's get this in the RC and test it there.

PR #4872, opened 2022-11-28
`Dir::copy()`: Support skipping the global ignore list

lukasbestle:
### Enhancements

- The `Dir::copy()` method now supports `false` as the argument for `$ignore`. In this mode, all files are copied, including those listed in the global `Dir::$ignore` list.

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/Dir.php
+++ b/src/Filesystem/Dir.php
@@ -48,12 +48,16 @@ class Dir
 
 	/**
 	 * Copy the directory to a new destination
+	 *
+	 * @param array|false $ignore List of full paths to skip during copying
+	 *                            or `false` to copy all files, including
+	 *                            those listed in `Dir::$ignore`
 	 */
 	public static function copy(
 		string $dir,
 		string $target,
 		bool $recursive = true,
-		array $ignore = []
+		array|bool $ignore = []
 	): bool {
 		if (is_dir($dir) === false) {
 			throw new Exception('The directory "' . $dir . '" does not exist');
@@ -67,10 +71,13 @@ class Dir
 			throw new Exception('The target directory "' . $target . '" could not be created');
 		}
 
-		foreach (static::read($dir) as $name) {
+		foreach (static::read($dir, $ignore === false ? [] : null) as $name) {
 			$root = $dir . '/' . $name;
 
-			if (in_array($root, $ignore) === true) {
+			if (
+				is_array($ignore) === true &&
+				in_array($root, $ignore) === true
+			) {
 				continue;
 			}
 
--- a/tests/Filesystem/DirTest.php
+++ b/tests/Filesystem/DirTest.php
@@ -55,6 +55,7 @@ class DirTest extends TestCase
 
 		$this->assertTrue(file_exists($target . '/a.txt'));
 		$this->assertTrue(file_exists($target . '/subfolder/b.txt'));
+		$this->assertFalse(file_exists($target . '/subfolder/.gitignore'));
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


bastianallgeier:
The both lists in combination don't seem very intuitive at first sight. At least I'm not fully sure I understand the concept :)

lukasbestle:
The reason why I picked this up was the Sandbox. There we use `Dir::copy()` to install and save the environments, but files like `.gitignore` weren't copied because `Dir::read()` (used internally by `Dir::copy()`) ignores them in the static `Dir::$ignore` ignore list. So if we want `Dir::copy()` to copy the entire directory, we need a way to tell it not to use the static ignore list. We need this for submodule support in the Sandbox, but I think it could also be useful in other cases.

My first solution was to simply override the static ignore list so that it is never used by `Dir::copy()`, with the thought that any ignores could still be passed with the `$ignore` argument. However it would be a huge breaking change.

So I ended up with another argument that defaults to the current behavior (`null` uses the static ignore list). But it can also be set to an array for a custom ignore list that applies to every directory level. So e.g. you could say `['.htaccess']` to just ignore htaccess files, no matter on which directory level they are. This was so far not possible with `Dir::copy()`.

I think the separate `$globalIgnore` argument is a simple and flexible implementation without breaking change. I'm open for naming alternatives if it helps.

distantnative:
I do understand the use case but have to agree that having too ignore parameter is quite confusing. And not sure this can be fixed through different naming.

My understanding is that the existing parameter supports ignoring of specific paths while your new parameter supports file names no matter at which path? Couldn't we rather support wildcards/glob or similar for the existing parameter then?

distantnative:
And for our specific use case in the sandbox it might be indeed best to unser and reset the global list. Why would that be a breaking change if we just do that in our sandbox?

lukasbestle:
The issue with wildcards/globs is that the static ignore list would still apply. Without a separate argument, there wouldn't be a way to disable the static ignore list without a breaking change.

I'm closing this then. Unsetting and resetting the static list from the Sandbox feels a bit like a hack, but I think it's still OK and better than a core implementation we are not happy with.

distantnative:
And passing `false` to the existing parameter to disable the global prop?

lukasbestle:
Oh, that could indeed work to disable ignores completely. 👍
Do you think it's useful on its own or would we also need a way to say "don't use the global ignores, but use this custom list instead"? For that I don't have an idea yet.

distantnative:
Ok, I had a bit closer look and I think the main problem is that so far `::copy` uses its parameter as own ignore list, not interacting with how `::read` handles ignoring (falling back then to the global list). I feel like both methods should better work together here.

If we would pass `$ignore` to `::read()` it is a bit of a breaking change because so far the parameter and global list would both be applied. When passing the local list over to read, the global list would be ignored. This is a breaking change, I am still wondering if this wouldn't be the correct way to do it, also to be consistent with `::read()` on its own.

Would it be too cumbersome to require people who want both global and local list to pass in the future `::copy($dir, 4target, true, array_merge(Dir::$ignore, [$a, $b, $c]))`? Maybe yes. Don't have the best idea in mind yet - but maybe we get the discussion going in that direction: How could we improve `::copy` by actually passing its `$ignore` to `::read` instead of doing its own check in the loop?

lukasbestle:
An additional issue is that the `::copy()` ignore list are full paths while the `::read()` ignore list just applies to a single directory level. I think it's confusing to combine both into one list, otherwise it would be impossible to express "just ignore files named XYZ at the root of the copied directory". For that we'd need to replace everything with wildcard patterns, support a leading slash to match on the root etc. It all gets very complex that way.

lukasbestle:
@distantnative I'm currently leaning towards just supporting `false` for "don't use the global ignore list", without any more magic. What do you think?

distantnative:
Maybe that is the best actionable solution indeed. I was trying to find a complete solution but didn't manage.

distantnative (approved):

PR #4886, opened 2022-12-03
Upgrade dependencies

distantnative:
### Refactored
- Upgraded composer and npm dependencies
- Upgraded CI dependencies

### ⚠️ Update local setups as well:
- phpunit:9.5.26
- psalm:5.1.0
- php-cs-fixer:3.13.0
- composer:2.4.4
- phpmd:2.13.0

## Ready?

### For review team


lukasbestle (commented):
The comments after the action version hashes are from [pin-github-actions](https://github.com/mheap/pin-github-action). If we add the minor versions to the comments, we can no longer auto-update the actions with the tool as they will be strictly pinned.

distantnative:
It's the first time I hear of that tool 😅 can we add explaining comments to the workflow files? If I knew, I wouldn't have needed to go through each dependency manually and thinking "we should be more specific with the version comment to make this easier for next time" 🙈 🙈 🙈

distantnative:
Although, isn't manually updating actually better here? Don't think the tool offers a good way to see what changed, if there is a new major version etc. Or does it?

lukasbestle:
Sorry for the manual work you had to go through. We discussed this tool way back when we had the issues of CI randomly breaking all the time, but it's long ago and since then only Ahmet and I have been updating CI.

I'm 💯 for adding a link to each workflow file.

lukasbestle:
> Although, isn't manually updating actually better here? Don't think the tool offers a good way to see what changed, if there is a new major version etc. Or does it?

It does neither of these things. For us, the only purpose of pinning the hashes is CI reliability (at least at the moment). Before we used it, CI always used the latest versions and they would often break our CI runs because of some breaking changes or regressions. Pinning also comes with security advantages, but we haven't used it for that so far.

So we didn't really review the changes in the actions on each update. I read through the release pages of each action that the tool *did* update, but didn't open every single one to check for major updates.

I'm fine with either approach: Full manual or partly automated with the tool.

distantnative:
@lukasbestle would it work if I leave my manual commit signature updates but just revert the `pin@` comment changes?

lukasbestle:
> @lukasbestle would it work if I leave my manual commit signature updates but just revert the `pin@` comment changes?

Yeah, sure.

src/Toolkit/Iterator.php
```diff
@@ -15,6 +15,8 @@
  * @link      https://getkirby.com
  * @copyright Bastian Allgeier
  * @license   https://opensource.org/licenses/MIT
+ *
+ * @psalm-suppress MissingTemplateParam
```
lukasbestle: What does this do? Or rather, why is this needed? I don't really understand it from the Psalm docs.

distantnative: Psalm started throwing this type of error for the `include $autoloader` calls - disregarding that we just check before for `is_file($autoloader)`. I took it as false positive from their new 5.x version. Thought it's the easiest to suppress it here for now.

lukasbestle: I think you are referring to the other Psalm-related change. I agree with you on that one.

But this one I don't understand.

distantnative: My mind only saw "missing" in the GitHub preview and jumped to the wrong one.

I don't either honestly. It might be new with psalm 5: https://psalm.dev/running_psalm/issues/MissingTemplateParam/ seems like psalm would like us to specify how we exactly implement IteratorAggregate. I wouldn't quite know how that specification should look like. I don't think it gives us much benefits. It has worked without so far. That's why I suppressed it.

lukasbestle: I looked into it more and fixed it with Psalm annotations.

lukasbestle (dismissed):

src/Toolkit/Iterator.php
distantnative: I think we would have to add these to every class that extends Iterator. Not sure that is worth it.

lukasbestle: You are right, I didn't realize this was the case. I've reverted it and added a comment for future reference.

lukasbestle (approved):

PR #4887, opened 2022-12-03
PHP 8.2 prep: Parsedown - fix `mb_convert_encoding` deprecation

distantnative:
https://php.watch/versions/8.2/mbstring-qprint-base64-uuencode-html-entities-deprecated

### DIFF
--- a/dependencies/parsedown-extra/ParsedownExtra.php
+++ b/dependencies/parsedown-extra/ParsedownExtra.php
@@ -571,7 +571,8 @@ class ParsedownExtra extends Parsedown
         $DOMDocument = new DOMDocument();
 
         # http://stackoverflow.com/q/11309194/200145
-        $elementMarkup = mb_convert_encoding($elementMarkup, 'HTML-ENTITIES', 'UTF-8');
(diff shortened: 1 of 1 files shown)


bastianallgeier (commented):
The example from your link is also encoding the string to UTF-8. Do you think we can skip this?

distantnative:
@bastianallgeier My understanding in the example is that their original string is in `ISO-8859-1` and that's why they need to ensure now to still convert it to `UTF-8`:

```diff
  $fromEncoding = 'ISO-8859-1';
- $encodedStr = mb_convert_encoding($str, 'HTML-Entities', $fromEncoding);
+ $encodedStr = mb_convert_encoding($str, 'UTF-8', $fromEncoding);
```

However, our Parsedown string already is UTF-8 based on the old code line

```php
$elementMarkup = mb_convert_encoding($elementMarkup, 'HTML-ENTITIES', 'UTF-8');
```

The third parameter is the format from which should be encoded.

bastianallgeier:
Ah, thanks for the explanation. That makes sense.

bastianallgeier (approved):

PR #4888, opened 2022-12-03
PHP 8.2 prep: Fix dynamic properties in search component

distantnative:
### Refactored 
- Search core component does not impose dynamic properties (`searchScore` and `searchHits`) on search items

## Ready?

### For review team

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -142,6 +142,7 @@ return [
 	 * @return \Kirby\Cms\Collection|bool
 	 */
 	'search' => function (App $kirby, Collection $collection, string $query = null, $params = []) {
+		// empty search query
 		if (empty(trim($query ?? '')) === true) {
 			return $collection->limit(0);
 		}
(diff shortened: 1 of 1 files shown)


bastianallgeier (commented):
This would be a breaking change. I'm not sure if we document it properly somewhere. But the reason for the dynamic properties is to be able to access them after running the search to get more info about the search hits and score for each item in the collection.

distantnative:
Ok but then we need the breaking change or we have to add the properties to each and any class that could be an item here.

PHP 8.2 just starts throwing warnings everywhere because of these.

distantnative:
And was this actually ever documented anywhere? Or have you been the only one knowing about this possibility? 😄

distantnative:
I had a look at the code again

> add the properties to each and any class that could be an item here

This won't work as the search core component is user by `Cms\Collection::search()`. So all kinds of objects could be in a collection. I only see the way to make this a breaking change - if we want to consider it a breaking change, still think this has been internal code not exposed to anybody as functionality.

bastianallgeier:
I think you are right. I checked again and I couldn't find it anywhere in our docs. PHP 8.2 support also seems like a much more important step than keeping those.

bastianallgeier (approved):

PR #4903, opened 2022-12-08
i18n/en - feburary fix

adamkiss:
### Fixes
i18 - english translation had a February name set to `Feburary`

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -393,7 +393,7 @@
 	"months.april": "April",
 	"months.august": "August",
 	"months.december": "December",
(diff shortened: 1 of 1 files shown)


distantnative:
@bastianallgeier can we fix this by merging or do we need to fix it inside Transifix?

bastianallgeier:
We need to fix it in Transifex. The big problem: all translations of February will afterwards be reverted to English. That's how Transifex handles a change in the source string.

lukasbestle:
But we still need to merge this to prevent the change from being reverted on the next import, right?

PR #4905, opened 2022-12-08
Fix `->words()` field method

### LINKED ISSUE #4904: PHP 8.1: ->words() method breaks in loop
## Description
After upgrading to php 8.1, when wanting get total word count over a set of blocks, it returns an error: 
`strip_tags(): Passing null to parameter #1 ($string) of type string is deprecated`

(Use case: to tally word count of an article consisting of different block types, then calculate reading time)

## To reproduce
Working off the plainkit, I simply added a blocks field:
```yaml
columns:
  main:
    width: 2/3
    sections:
      fields:
        type: fields
        fields:
          text:
            type: textarea
            size: huge
          blocks:
            type: blocks
```
And in default.php, I tried to tally up all words:
```php
Word count in textarea: <?= $page->text()->words(); // this works ?>
<br>
<?php
  $counter = 0;
  foreach($page->blocks()->toBlocks() as $block) {
    $counter += $block->text()->words(); // error
  };
  echo 'Word count in blocks: ' . $counter;
?>

```

This running php 8.1.13.
Switching back to php@8.0, word count returns as expected.

Playing around with the [`words()` method](https://github.com/getkirby/kirby/blob/3.8.3/config/methods.php#L328), it works if value is not passed as argument:
```php
'words' => function (Field $field) {
  return str_word_count(strip_tags($field));
},
```
But, I'm not so versed in PHP to suggest this as the solution 😬

## Your setup
**Kirby Version**  
Kirby 3.8.3

PHP 8.1.13 on MacOS Ventura

afbora:
## This PR …

~Actually for solution, we can use simply `strip_tags($field->value ?? '')` but I preferred the `Str::unhtml()` method.~

### Fixes
- `->words()` field method works as expected on `null` values #4904 

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/methods.php
+++ b/config/methods.php
@@ -326,7 +326,7 @@ return function (App $app) {
 		 * Returns the number of words in the text
 		 */
 		'words' => function (Field $field) {
-			return str_word_count(strip_tags($field->value));
+			return str_word_count(strip_tags($field->value ?? ''));
 		},
 
 		// manipulators
--- a/tests/Cms/Fields/FieldMethodsTest.php
+++ b/tests/Cms/Fields/FieldMethodsTest.php
@@ -722,6 +722,7 @@ class FieldMethodsTest extends TestCase
 	{
 		$text = 'this is an example text';
 		$this->assertSame(5, $this->field($text)->words());
+		$this->assertSame(0, $this->field(null)->words());
 	}
 
 	public function testXml()


distantnative (dismissed):

lukasbestle (commented):
`Str::unhtml()` also decodes HTML entities, which could lead to double decoding depending on how the field method is used. I think `?? ''` is safer.

afbora:
Does this affect word count?

lukasbestle:
As a "word" is defined locale dependent, it could in theory. What I like about `?? ''` is that it's explicit in what it does, so even if we later change any of the underlying methods, we won't have a regression.

lukasbestle (approved):

PR #4910, opened 2022-12-08
Snippets with slots

distantnative:
_Alternative to https://github.com/getkirby/kirby/pull/4837_

## This PR …

This PR supercharges our snippets with the ability to host slots. We already had the layout plugin with a very similar approach, but never managed to get nesting right. This version is now able to support indefinitely nested snippets and also handles slots a lot more predictably. It feels ready for prime-time.

### Features

Using regular snippets still works as you are used to
```php
<?php snippet('copyright') ?>
``` 

However, snippets can now also used with slots. And it's quite straightforward:

```php
<?php snippet('article', slots: true) ?>
  <?php slot('header') ?>
    <h1>This is the title</h1>
  <?php endslot() ?>

  <?php slot('body') ?>
    <p>This is some body text</p>
  <?php endslot() ?>
<?php endsnippet() ?>
```

And this is how your article snippet (still `site/snippets/article.php`) would look:

```php
<article>
  <?php if ($header = $slots->header()): ?>
  <header>
    <?= $header ?>
  </header>
  <?php endif ?>
  
  <?= $slots->body() ?>
</article>
```

But it also supports the beloved syntax from the layout plugin:

 ```php
<?php snippet('layout', slots: true) ?>

This is my content
```

 ```php
<html>
  <head></head>
  <body>
  <?= $slot // same as $slots->default() ?>
  </body>
</html>
```

### Refactoring
The `Kirby\Cms\Template` class moved to `Kirby\Template\Template`. An alias has been added.

### Breaking changes
- `snippet` core component receive an additional parameter `bool $slots` and returns `\Kirby\Template\Snippet|string`
- `$kirby->snippet()` and the `snippet()` helper return `\Kirby\Template\Snippet|string|null`

## Ready?

### For review team


bastianallgeier:
Do you really think we have to wait for 3.9 to introduce this? The breaking changes seem to be pretty non-breaking to me unless you start using the slots param.

distantnative:
I think this is such a major feature that it deserves a major release even just for the marketing/attention aspect.

distantnative:
And the breaking change is not about whether to use slots or not. Any custom snippet core component will break because PHP will enforce the parameter signature(we'd pass one parameter more than the old custom snippet components would accept).

bastianallgeier:
You are right. I didn't think about the plugins.

bastianallgeier (approved):

Issue #4906, opened 2022-12-08
Dom unit tests


afbora:
The annoying dom unit tests need to be fixed:
https://github.com/getkirby/kirby/actions/runs/3630541371/jobs/6124007281

distantnative:
@afbora do you see these errors in your local setup as well? Bastian and I do, have been for a while. But e.g. I think for Lukas the tests pass as they are right now locally.

afbora:
@distantnative Yes, my tests fails as well:

![image (5)](https://user-images.githubusercontent.com/3393422/206414963-007eb618-ea5d-411b-80f8-f0b59d29c303.png)

distantnative:
@lukasbestle what about you? Or do we just have to adapt the unit tests?

afbora:
I'm not sure but could it be depending on PHP, DOM/XML version or local setup settings?

PR #4914, opened 2022-12-09
Temporary fix for the broken unit tests

bastianallgeier:
## This PR …

I could not find out why the line break between html and body tags gets removed. It seems like this has nothing to do with our code but the version of the Dom extension. 

## Fixes

- https://github.com/getkirby/kirby/issues/4906


bastianallgeier:
Ha, interesting. It's a different set of issues on PHP 8.1 🙈

PR #4915, opened 2022-12-09
Automatic `alt` text for assets, file versions

distantnative:
## This PR …

We already use the `alt` content field as automatic fallback when using `Cms\File::html()`. However, this does not get called for thumbs or other instances based on `FileVersion` or the `Image` class...

### Enhancement
- `Image\Image` uses its model's `alt` content field as fallback for rendering an `alt` tag to better provide accessible defaults

## Ready?

### For review team

### DIFF
--- a/src/Image/Image.php
+++ b/src/Image/Image.php
@@ -112,6 +112,12 @@ class Image extends File
 	 */
 	public function html(array $attr = []): string
 	{
+		// if no alt text explicitly provided,
+		// try to infer from model content file
+		if ($alt = $this->model?->alt()) {
+			$attr['alt'] ??= $alt;
+		}
+
 		if ($url = $this->url()) {
 			return Html::img($url, $attr);
 		}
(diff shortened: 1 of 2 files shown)


lukasbestle (approved):
Really simple and useful!

distantnative:
@texnixe Do you know if we mention somewhere in the docs that for this reason, people should best use `alt` as filename on their images (and not e.g. caption...)?

lukasbestle:
@distantnative Do you mean field name instead of filename?

distantnative:
Yes - I really need to do something about my autocorrect. It's so bad and I don't realise it so often 🙈

lukasbestle:
I'd say there are use cases for both `alt` and `caption` fields. Alt texts and captions are completely different in their purpose, form and content, so you may well have a presentational image without alt text but with caption, while in other situations you may just want to have an alt text but no caption, or even both. So I think what the docs should explain is that the `alt` field is a "special" field that is automatically used by Kirby for alt texts.

PR #4919, opened 2022-12-09
DELETE `lock`: fail gracefully if already deleted

### LINKED ISSUE #4897: `The page "*" cannot be found` error when page deleted
## To reproduce

1. Create a page from panel
2. Edit the page without save
3. Try to delete
4. See error console (a dialog appears very quickly, closes, and sometimes remains visible)

## Your setup
**Kirby Version**  
3.8.3-rc.1 & 3.8.x (and probably 3.7.x, not tested yet)

**Console output**  
````json
{
    "status": "error",
    "message": "The page \"blog/test\" cannot be found",
    "code": 404,
    "exception": "Kirby\\Exception\\NotFoundException",
    "key": "error.page.notFound",
    "file": "/test/kirby/starterkit-3.8.3-rc.1/kirby/src/Cms/Find.php",
    "line": 85,
    "details": [],
    "route": "(.*)/lock"
}
````

distantnative:
Fixes #4897

### Fixes
- Fixed console errors for failed `lock` request when deleting page with unsaved changes
#4897

## Ready?

### For review team

### DIFF
--- a/config/api/routes/lock.php
+++ b/config/api/routes/lock.php
@@ -4,6 +4,9 @@
 /**
  * Content Lock Routes
  */
+
+use Kirby\Exception\NotFoundException;
+
 return [
 	[
 		'pattern' => '(:all)/lock',
@@ -25,7 +28,11 @@ return [
 		'pattern' => '(:all)/lock',
 		'method'  => 'DELETE',
 		'action'  => function (string $path) {
-			return $this->parent($path)->lock()?->remove();
+			try {
+				return $this->parent($path)->lock()?->remove();
+			} catch (NotFoundException) {
+				return true;
+			}
 		}
 	],
 	[
@@ -39,7 +46,11 @@ return [
 		'pattern' => '(:all)/unlock',
 		'method'  => 'DELETE',
 		'action'  => function (string $path) {
-			return  $this->parent($path)->lock()?->resolve();
+			try {
+				return $this->parent($path)->lock()?->resolve();
+			} catch (NotFoundException) {
+				return true;
+			}
 		}
 	],
 ];


afbora:
This PR fixes the issue 👍 But we can think make it for also should apply for `PATCH` routes. Because If a lock request comes at a time when there is no page, it will still give an error. Rare but possible :)

afbora (dismissed):

distantnative:
Can you produce such a situation? I'm. It sure what the right return value in the case for PATCH would be. If we have an actual situation, maybe we know.

afbora:
I reduced the `/lock` request seconds to less than 10 seconds (because it was unlikely that I would make a request at exactly the 10th second), I could see the error dialog.

> The page "notes/test" cannot be found

Like I said, it's unlikely, but it could happen.

distantnative:
Hmm, this is difficult. I think if we return `false`, we will get an error as well. But passing `true` seems wrong - if the page cannot be found the lock info wasn't updated at all.

lukasbestle:
I feel this behavior is caused by a race condition in the Panel frontend rather than by the API. I think the proper solution would need to be that the Panel no longer requests locks after the delete request for the page was sent.

distantnative:
> I feel this behavior is caused by a race condition in the Panel frontend rather than by the API. I think the proper solution would need to be that the Panel no longer requests locks after the delete request for the page was sent.

I tried to find that but I couldn't. Can't stop locking quickly enough after deletion to not have the lock request already sent. So this is the only solution I could come up, I think it's better than none (and for the DELETE request I think the current implementation also makes sense).

lukasbestle (dismissed):
I agree. If we find the source of the race condition for `PATCH` later, we can still fix that independently.

lukasbestle (approved):

PR #4920, opened 2022-12-09
Fix resolving paths with trailing dot

### LINKED ISSUE #4843: Trailing dot on page route should trigger 404
## Description
The path `/account.` currently resolves to the page at `/account`.

Pages with nonexistant content representations (e.g. `/account.json`) currently return 404. For consistency, `/account.` should be treated the same (either as a "nonexistant content representation" or as invalid syntax).

I discovered this problem when trying to use a custom route to block access to my Account page:

```php
<?php
return [
    'routes' => [
        [
            'pattern' => ["account", "account/(:all)"],
            'action' => function () {
                if (!kirby()->user()) {
                    return go("/login");
                }
                return $this->next();
            }
        ]
    ]
];
```

The path `/account.` did not match my custom route, but Kirby resolved it to my account page, skipping the user login check entirely.

**Expected behavior**  

`/account.` should be treated as either a nonexistant content representation or as invalid syntax. This will make it possible to consistently overwrite the default path resolution behavior with a simple custom route.

## To reproduce

1. Publish a page with a given slug e.g. `/example`
1. Access the page with the path `/example.`
1. (Optionally) Compare to accessing the page with a nonexistant content representation `/example.json`

## Your setup
**Kirby Version**  
3.7.5

 - Device: M1 MacBook Air
 - OS: macOS 12.4
 - Browser: Chrome
 - Version: 106

distantnative:
### Fixes
- Treating paths with trailing dot as non-existing content representation
#4843

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1307,7 +1307,7 @@ class App
 		$extension = F::extension($path);
 
 		// no content representation? then return the page
-		if (empty($extension) === true) {
+		if (empty($extension) === true && Str::endsWith($path, '.') === false) {
 			return $page;
 		}
 
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1272,6 +1272,11 @@ class App
 	 */
 	public function resolve(string $path = null, string $language = null)
 	{
+		// directly prevent path with incomplete content representation
+		if (Str::endsWith($path, '.') === true) {
+			return null;
+		}
+
 		// set the current translation
 		$this->setCurrentTranslation($language);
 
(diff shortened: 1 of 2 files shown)


src/Cms/App.php
```diff
@@ -1307,7 +1307,7 @@ public function resolve(string $path = null, string $language = null)
 		$extension = F::extension($path);
 
 		// no content representation? then return the page
-		if (empty($extension) === true) {
+		if (empty($extension) === true && Str::endsWith($path, '.') === false) {
```
lukasbestle: With this change, the representation lookup logic will continue for such paths with a trailing dot. So Kirby would try to find a template like `default..php`.

Do we explicitly want to support this? If not, I think it would be better if we instead used `if (Str::endsWith($path, '.') === true) { return null; }` already pretty much at the top of the method for optimal performance.

distantnative: I think you are right, I don't think it makes much sense to support basically an empty representation type with `default..php`

lukasbestle (approved):

PR #4921, opened 2022-12-09
`Dimensions::forSvg`: support for percentages

### LINKED ISSUE #4793: Svg: wrong width and height dimensions detected
I'm working with this svg file and got confused because `$file->width()` returned `100` while it's actually `331`:

bistum-hildesheim-schwarz

Turns out the width and height are set to `100%` and the real dimensions are in the the `viewBox`. I extracted the logo out of a pdf file and exported it using Affinity Photo. While I would say that the svg file is wrong it would still be cool if Kirby would master it. 

Currently (Kirby 3.8.1) `Dimensions::forSvg` reads the `width` and `height` attributes and falls back to the values in `viewBox` when either `width` or `height` are zero:

```php
$attr   = $xml->attributes();
$width  = (int)($attr->width);
$height = (int)($attr->height);
if (($width === 0 || $height === 0) && empty($attr->viewBox) === false) {
	$box    = explode(' ', $attr->viewBox);
	$width  = (int)($box[2] ?? 0);
	$height = (int)($box[3] ?? 0);
}
```

What do you think about checking the uncasted values for a percentage sign and resort in that case to `viewBox` too to get the correct dimensions?

distantnative:
### Enhancements
- `Dimensions::forSvg`: support for percentages
#4793

## Ready?

### For review team

### DIFF
--- a/src/Image/Dimensions.php
+++ b/src/Image/Dimensions.php
@@ -2,6 +2,8 @@
 
 namespace Kirby\Image;
 
+use Kirby\Toolkit\Str;
+
 /**
  * The Dimension class is used to provide additional
  * methods for images and possibly other objects with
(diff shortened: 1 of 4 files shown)


lukasbestle (approved):
I think the `circle-offset` example shows that this solution of falling back to the `viewBox` is not perfect in edge cases, but it's still a great improvement for the 99% case (pun not intended 😄).

PR #4922, opened 2022-12-10
`Toolkit\Silo`: code + tests improvements

distantnative:
### Refactoring
- Improvements to `Toolkit\Silo` code quality

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/Silo.php
+++ b/src/Toolkit/Silo.php
@@ -15,19 +15,12 @@ namespace Kirby\Toolkit;
  */
 class Silo
 {
-	/**
-	 * @var array
-	 */
 	public static $data = [];
 
 	/**
-	 * Setter for new data.
-	 *
-	 * @param string|array $key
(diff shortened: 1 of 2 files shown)


src/Toolkit/Silo.php
```diff
@@ -15,19 +15,12 @@
  */
 class Silo
 {
-	/**
-	 * @var array
-	 */
 	public static $data = [];
```
bastianallgeier, line 18:
```suggestion
	public static array $data = [];
```

distantnative: I left it out again, because then PHP fails unless we also add the type to `Config::$data`. Which wouldn't be a problem for us, but in theory this could be a breaking change for others extending `Silo`. So I started to be a bit more cautious with adding typing in minor versions. What do you think?

bastianallgeier (approved):

PR #4930, opened 2022-12-10
Register dialog: info field for domain, warning for local domain

### LINKED ISSUE #4795: Don't allow accidental Kirby license registrations to non-existing TLDs
## Description
While deploying a new site, I registered the new site and then promptly lost the registration again by re-uploading my local site folder. Then I thought, well let me instead register it locally, so it's registered and then push it online.

Well, as you can guess (and I couldn't) that made it registered to my local testing domain `foo.test`. And now I cannot change this anymore. Not sure why it chooses to stick with the second registration that I tried and not with the first one, but ...

It would be great if the registration system would be smart enough to realize that `foo.test` is *obviously* a local testing domain and should not be the domain to marry your license with permanently.

I guess deployment issues like this can happen, so it would be good if domains like `foo.test` are just ignored and the system waits for the actual domain to register with.

**Expected behavior**  
Don't allow licenses to register to domains like `foo.test`.

distantnative:
### Enhancements
- The register dialog shows now an info field for which domain the license will be registered, including a warning when restoring a local domain
#4795

Screenshot 2022-12-10 at 18 11 12
Screenshot 2022-12-10 at 18 12 01

### Refactoring
- Cleaned up `Cms\System` class

## Ready?

### For review team


i18n/translations/en.json
```diff
@@ -349,6 +349,8 @@
 	"license.manage": "Manage your licenses",
 	"license.register.help": "You received your license code after the purchase via email. Please copy and paste it to register.",
 	"license.register.label": "Please enter your license code",
+	"license.register.domain": "Register your license for <strong>{host}</strong>",
```
lukasbestle:
```suggestion
	"license.register.domain": "Your license will be registered to the domain <strong>{host}</strong>.",
```

i18n/translations/en.json
```diff
@@ -349,6 +349,8 @@
...
+	"license.register.local": "You are about to register your license for your local domain <strong>{host}</strong>. If this site will be deployed to a public domain, please register it there instead. If <strong>{host}</strong> is the domain you want to license Kirby to, please continue.",
```
lukasbestle: I think it's enough to bold the domain once:

```suggestion
	"license.register.local": "You are about to register your license for your local domain <strong>{host}</strong>. If this site will be deployed to a public domain, please register it there instead. If {host} is the domain you want to license Kirby to, please continue.",
```

src/Cms/System.php
```diff
@@ -231,18 +205,15 @@ public function init()
...
 	 * option must be explicitly set to true
 	 * to get the installer up and running.
-	 *
-	 * @return bool
 	 */
 	public function isInstallable(): bool
 	{
-		return $this->isLocal() === true || $this->app->option('panel.install', false) === true;
+		return $this->isLocal() === true ||
+			   $this->app->option('panel.install', false) === true;
```
lukasbestle: Line 212 has mixed tabs and spaces. We could fix this by using spaces after the indentation, but that gets hard to manage. What about this:

```suggestion
		return (
			$this->isLocal() === true ||
			$this->app->option('panel.install', false) === true
		);
```

src/Cms/System.php
```diff
@@ -633,7 +571,8 @@ public function updateStatus(array|null $data = null): UpdateStatus|null
 		}
 
 		$kirby  = $this->app;
-		$option = $kirby->option('updates.kirby') ?? $kirby->option('updates') ?? true;
+		$option = $kirby->option('updates.kirby') ??
+				  $kirby->option('updates', true);
```
```suggestion
		$option = (
			$kirby->option('updates.kirby') ??
			$kirby->option('updates', true)
		);
```

i18n/translations/en.json
distantnative: Tried to keep this extra short to have no line break

src/Cms/System.php
distantnative: Don't we have that in quite a few places to align items if e.g. half a tab is needed?

i18n/translations/en.json
lukasbestle: I thought so, it's definitely a good point. On the other hand, I find it weird that it's not a full sentence, IMO it looks misplaced in an info box.

We could leave out "the domain", so it's just:

> Your license will be registered to sandbox.test.

src/Cms/System.php
lukasbestle: What is "half a tab"? That would require everyone to use the same tab width. You can see it in the GitHub diff preview above that it breaks once a different tab width is used (here: 8 spaces). We all use 4 spaces, but the beauty of tab indentation is that everyone can use whatever width they prefer. Mixing spaces and tabs at the start of the line is really weird IMO.

lukasbestle: To answer your question: I don't know if we use it anywhere else already.

distantnative: We have it a lot: if you do a global search in PHP files for one tab followed by two spaces, you'll see many of the places where we do this. 

> That would require everyone to use the same tab width.

That's true. Then again we have `.editorconfig` defining this for everybody (just GitHub ignoring it, which many consider a bug).

distantnative: I've removed them in all the places and used a pattern that we already have in other places:

```php
return
	$this->isLocal() === true ||
	$this->app->option('panel.install', false) === true;
```

lukasbestle (approved):
I like that syntax, it's still quite compact. At least in PHP with the mandatory semicolons. In JS it wouldn't work, but there we have Prettier anyway.

PR #4931, opened 2022-12-10
`thumb` component: process in tmp location

### LINKED ISSUE #4632: Broken thumb handling when jobs file still exists
## Description
We ran into the situation where thumb generation for a large image failed (converting from JPEG source to WebP thumb): the generated thumb was only half rendered, the lower part missing, so I guess there was some kind of timeout while writing to disk. Kirby did not delete the corresponding jobs file so the media folder was in a conflicting state: we had both, a generated (but broken) thumb and a jobs file. 

When requesting the image, Kirby served the broken thumb.

**Expected behavior**  
When Kirby detects an existing jobs file, it should always generate a new thumb because most certainly something failed when creating any existing thumb.

## Your setup
**Kirby Version**  
3.7.5

distantnative:
### Fixes
- `thumb` core component: processes the image in a  temporary location first before moving it into the final location to prevent broken media files 
 #4632

## Ready?

### For review team

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -314,8 +314,18 @@ return [
 		$options  = $darkroom->preprocess($src, $options);
 		$root     = (new Filename($src, $dst, $options))->toString();
 
-		F::copy($src, $root, true);
-		$darkroom->process($root, $options);
+		// Generate `$tmp`/`$path` location where file first is moved
+		// and processed before moved to the final location.
+		// When there are problems during the generation, this
+		// ensures not to leave a broken media file in the final
+		// location but rather none, which will trigger a new generation
+		// process during the next request (as the job file will still remain)
+		$tmp  = tmpfile();
+		$path = stream_get_meta_data($tmp)['uri'];
+
+		F::copy($src, $path, true);
+		$darkroom->process($path, $options);
+		F::move($path, $root, true);
 
 		return $root;
 	},
--- a/src/Cms/Media.php
+++ b/src/Cms/Media.php
@@ -119,7 +119,7 @@ class Media
 			}
 
 			$source = match (true) {
-				is_string($model) === true
+				is_string($model)
 					=> $kirby->root('index') . '/' . $model . '/' . $options['filename'],
 				default
 				=> $model->file($options['filename'])->root()
--- a/tests/Cms/Media/MediaTest.php
+++ b/tests/Cms/Media/MediaTest.php
@@ -9,27 +9,28 @@ use PHPUnit\Framework\TestCase;
 class MediaTest extends TestCase
 {
 	protected $app;
-	protected $fixtures;
+	protected $fixtures = __DIR__ . '/fixtures';
+	protected $tmp;
 
 	public function setUp(): void
 	{
 		$this->app = new App([
 			'roots' => [
-				'index' => $this->fixtures = __DIR__ . '/fixtures/MediaTest',
+				'index' => $this->tmp = __DIR__ . '/tmp',
 			],
 		]);
 
-		Dir::make($this->fixtures);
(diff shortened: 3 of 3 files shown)


distantnative:
There seem to be some issues with this, @lukasbestle. When I let the exceptions bubble up in `Media::thumb()`, I get

```
1) Kirby\Cms\MediaTest::testThumb
Kirby\Exception\Exception: The file "/Users/nhoffmann/Sites/kirby/sandbox/kirby/tests/Cms/Media/tmp/media/site/85688e905c-1670757882/.jobs/test.jpg.json" does not exist or cannot be read

/Users/nhoffmann/Sites/kirby/sandbox/kirby/src/Data/Handler.php:41
/Users/nhoffmann/Sites/kirby/sandbox/kirby/src/Data/Data.php:97
/Users/nhoffmann/Sites/kirby/sandbox/kirby/src/Cms/Media.php:115
/Users/nhoffmann/Sites/kirby/sandbox/kirby/tests/Cms/Media/MediaTest.php:228

2) Kirby\Cms\MediaTest::testThumbStringModel
Whoops\Exception\ErrorException: imagecreatefromstring(): Data is not in a recognized format

/Users/nhoffmann/Sites/kirby/sandbox/kirby/vendor/claviska/simpleimage/src/claviska/SimpleImage.php:185
/Users/nhoffmann/Sites/kirby/sandbox/kirby/src/Image/Darkroom/GdLib.php:29
/Users/nhoffmann/Sites/kirby/sandbox/kirby/config/components.php:329
/Users/nhoffmann/Sites/kirby/sandbox/kirby/src/Cms/App.php:1689
/Users/nhoffmann/Sites/kirby/sandbox/kirby/src/Cms/Media.php:129
/Users/nhoffmann/Sites/kirby/sandbox/kirby/tests/Cms/Media/MediaTest.php:281
```

lukasbestle:
For the first one, the test comment says "// invalid with no job file", so I guess it is expected that the exception should get swallowed.

The second error could mean that the `file_get_contents()` that happens inside `SimpleImage` doesn't return the full image contents. Maybe the copy to the temp file fails. Could you check what `$file` contains after `SimpleImage.php:179`?

distantnative:
But if the first one would be fully expected, the unit tests shouldn't actually fail weirdly, but it does (mime type change): https://github.com/getkirby/kirby/actions/runs/3668731057/jobs/6202058772

distantnative:
Reverted the clean up to the tests to make sure it's not a bug in there, but errors remain the same.

It seems like the thumb component already receives not a proper source string.

lukasbestle:
It looks like the issue is a deeper bug that was there before but is now uncovered.

`Media::thumb()` accepts either a string for the parent path or a model object (which can apparently be either a parent object like `Page` or `Site` or a `File` object itself). But a `File` object is not fully supported:

Here it is used to determine the target media root:

https://github.com/getkirby/kirby/blob/e24b2c43922e1c485ad9952663cfdf43aec593a3/src/Cms/Media.php#L100-L110

But here when determining the source path we always call `$model->file($options['filename'])`, which when called on a `File` object returns a `Field` object, not the `File` object that's expected for the following `->root()` call:

https://github.com/getkirby/kirby/blob/e24b2c43922e1c485ad9952663cfdf43aec593a3/src/Cms/Media.php#L121-L126

The result is that the `thumb` component gets an empty string for its `$src` path argument. If you replace `$file` with `$file->parent()` in line 230 of the `MediaTest.php`, at least this test succeeds. But I'm not sure what the intended behavior is and how to fix it.

distantnative:
Closing this for now. PR is referenced in corresponding issue to revisit the comments etc. when tackling this again

PR #4933, opened 2022-12-10
Add `Content-Type: application/json` for fiber requests

afbora:
### Fixes

- Fixes blocking requests by ModSecurity (owasp rules)

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/api/request.js
+++ b/panel/src/api/request.js
@@ -12,6 +12,8 @@ export async function toJson(response) {
 	return data;
 }
 
+import { toLowerKeys } from "../helpers/object.js";
+
 export default (config) => {
 	return {
 		running: 0,
@@ -23,7 +25,7 @@ export default (config) => {
 				headers: {
 					"x-requested-with": "xmlhttprequest",
 					"content-type": "application/json",
-					...options.headers
+					...toLowerKeys(options.headers ?? {})
 				}
 			});
 
--- a/panel/src/api/request.js
+++ b/panel/src/api/request.js
@@ -23,7 +23,6 @@ export default (config) => {
 				credentials: "same-origin",
 				cache: "no-store",
 				headers: {
-					"x-requested-with": "xmlhttprequest",
 					"content-type": "application/json",
 					...toLowerKeys(options.headers ?? {})
 				}
--- a/panel/src/fiber/index.js
+++ b/panel/src/fiber/index.js
@@ -235,6 +235,7 @@ export default class Fiber {
 				cache: "no-store",
 				headers: {
 					...this.options.headers(),
+					"Content-Type": "application/json",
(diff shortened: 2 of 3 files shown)


lukasbestle (dismissed):

panel/src/fiber/index.js
```diff
@@ -235,6 +235,7 @@ export default class Fiber {
 				cache: "no-store",
 				headers: {
 					...this.options.headers(),
+					"Content-Type": "application/json",
```
afbora: @lukasbestle Does it matter if it is lowercase or not?

afbora: I just wonder because it's lowercased in `api::request()`.

lukasbestle: [MDN says](https://developer.mozilla.org/en-US/docs/Web/API/Headers) that `fetch` handles headers case-insensitively, so I guess it doesn't matter. The same applies to the HTTP spec itself.

afbora: ActualIy I mean for overwriting content type.

lukasbestle: I just tested it to make sure:

```js
fetch('/', {
  headers: {
    "Content-Type": "application/json",
    "content-type": "text/something"
  }
});
```

In Safari 15.6.1, `fetch()` merges the two unfortunately, so the request header becomes `application/json, text/something`.

That's indeed not great as it means that the headers are treated *partly* case-sensitively.

afbora: What do you suggest? 

- Always use lowercase?
- Is it always to lowercase the keys before merging them? 

Do you have another idea?

lukasbestle: > Is it always to lowercase the keys before merging them?

I think that's a great idea!

distantnative: We should do it then in `request.js` as well

afbora: Done ✅

panel/src/api/request.js
```diff
@@ -23,7 +25,7 @@ export default (config) => {
 				headers: {
 					"x-requested-with": "xmlhttprequest",
```
afbora: @distantnative Do we need `x-requested-with` for `fiber.js` too?

bastianallgeier: We used this initially to detect fiber requests, but as far as I know, it's no longer needed.

afbora: 👍

bastianallgeier (dismissed):
Looks good to me. Do you want to remove the x-requested-with header or do we keep it?

afbora:
I thought it wasn't needed in just fiber.js. I've removed ✅

bastianallgeier (approved):

PR #4934, opened 2022-12-10
`Xml::attr()`: deprecate empty string as value

### LINKED ISSUE #4752: Html::attr() and attr() skip attributes without value
## Description
If I try to use these helper methods for empty attributes like

```php
Html::attr(['autoplay' => '', 'controls' => ''])
```
the attributes will not be inserted at all.

**Expected behavior**  
The attributes should be inserted.
Ideally, even a syntax like
```php
Html::attr(['autoplay', 'controls'])
```
should be possible, but I imagine this could get complicated when non-value and value attributes are mixed. Either way, attributes with an empty value should *not* be skipped.

## Your setup
**Kirby 3.7.5**

distantnative:
### Deprecated
- Passing an empty string as value to `Xml::attr()` has been deprecated and will throw a warning. In a future version, passing an empty string won't omit the attribute anymore but render it with an empty value. To omit the attribute, please pass `null`

Closes #4752.

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/Xml.php
+++ b/src/Toolkit/Xml.php
@@ -94,9 +94,11 @@ class Xml
 		// TODO: In 3.10, treat $value === '' to render as name=""
 		if ($value === null || $value === '' || $value === []) {
 			// TODO: Remove in 3.10
+			// @codeCoverageIgnoreStart
 			if ($value === '') {
 				Helpers::deprecated('Passing an empty string as value to `Xml::attr()` has been deprecated. In a future version, passing an empty string won\'t omit the attribute anymore but render it with an empty value. To omit the attribute, please pass `null`.');
 			}
+			// @codeCoverageIgnoreEnd
 
 			return null;
 		}
--- a/src/Toolkit/Xml.php
+++ b/src/Toolkit/Xml.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Toolkit;
 
+use Kirby\Cms\Helpers;
 use SimpleXMLElement;
 
 /**
@@ -90,10 +91,18 @@ class Xml
 			return implode(' ', $attributes);
 		}
 
+		// TODO: In 3.10, treat $value === '' to render as name=""
 		if ($value === null || $value === '' || $value === []) {
+			// TODO: Remove in 3.10
+			if ($value === '') {
+				Helpers::deprecated('Passing an empty string as value to `Xml::attr()` has been deprecated. In a future version, passing an empty string won\'t omit the attribute anymore but render it with an empty value. To omit the attribute, please pass `null`.');
+			}
+
 			return null;
 		}
 
+		// TODO: In 3.10, add deprecation message for space = empty attribute
+		// TODO: In 3.11, render space as space
 		if ($value === ' ') {
 			return $name . '=""';
 		}


lukasbestle (approved):

distantnative:
Thanks for the review and the notice on top. Switched back to draft as this will probably safer to make sure it won't be merged before 3.9

afbora:
````php
<?= Html::email('john@doe.com') ?>
````

@distantnative Above simple sample code throws deprecation warning. Is this intended? Do you want to me create new issue?

> Passing an empty string as value to `Xml::attr()` has been deprecated. In a future version, passing an empty string won't omit the attribute anymore but render it with an empty value. To omit the attribute, please pass `null`.

**Comment by lukasbestle** [comment]
Good catch @afbora! I've created a PR: https://github.com/getkirby/kirby/pull/4948

**Comment by MoritzLost** [comment]
I just stumbled upon this – this deprecation notice is really inconvenient. We're using the `attr` function all over the place, often with default arguments that may or may not be empty. For example, a lot of components support optional classes that default to an empty string. It doesn't matter at all if the `class` attribute is left out or rendered with an empty value. What _does_ matter is that I'm now getting error like this from all over the place: 

error-deprecation-notice

I didn't even know that empty attribute values resulted in the attribute being left out instead, I just thought they were included with an empty value. This is what will happen in 3.10, but the current behaviour also works in 3.9. However, because of the deprecation notice, I now have to add explicit empty checks to a bunch of templates.

In the future, I would suggest to more closely examine if a deprecation notice is actually necessary. In this case, the behaviour could've just been brought in line with the intended behaviour with no problems. Who is this deprecation notice for? The only case where this would make a difference is if you were explicitly using `['data-my-attribute' => '']` with an empty string to leave out the attribute, which … why? The normal way to do this would be to set the attribute to `false` or leave it out entirely.

Having to do all this work just to get rid of a deprecation notice is really annoying, especially since we're already using the function as intended and this work will be obsolete in 3.10.

**Comment by lukasbestle** [comment]
@MoritzLost Sorry for the inconvenience.

To explain why we made the change and need the deprecation warning:

The original syntax for the `attr()` method was a complete mess. To produce an attribute with an empty value, you needed to pass a single space as the value, which is super confusing and inconsistent. The goal was to find a new syntax that was consistent and logical (https://github.com/getkirby/kirby/issues/4752). Of course it makes a lot of sense that an empty string produces an empty attribute string. But the empty string was already in use for "omit the attribute entirely". This is why we need the deprecation warning for the [multi-step migration process](https://github.com/getkirby/kirby/issues/4752#issuecomment-1267969344) to avoid an immediate breaking change:

1. First deprecate passing an empty string, instead suggest `null` (which is what we did in 3.9).
2. Then switch over to an empty string meaning an empty attribute. At the same time deprecate passing a space, instead suggest an empty string.
3. Finally switch over a space meaning a space.

So **TL;DR:** In 3.10, there will no longer be a deprecation warning for an empty string, instead the behavior will change.

We keep breaking changes to a minimum as much as possible, but in order to modernize the code base, they are sometimes necessary. The deprecation warning allows you to make the necessary changes step by step. In production, you can disable error reporting for deprecations (`E_USER_DEPRECATED`) and your site will work as before. We feel this is a much smoother migration process than an immediate breaking change – or at worst, three of them if we had combined all changes into one release.

**Comment by lukasbestle** [comment]
@MoritzLost Sorry, completely misread your message. I understand what you mean. We need to reconsider this for 3.9.1.

**Comment by neildaniels** [comment]
Just tried upgrading my installation to 3.9.0 and ran into this deprecation message. I agree that this particular deprecation notice seems unnecessary.

As an aside: I wish there was a more granular way to silence deprecation notices in these types of situations (_especially_ when I prefer the "future" behavior). Turning off all deprecation messages is unfortunate because then I can't really go through and find other types of deprecation notices.

**Comment by MoritzLost** [comment]
@lukasbestle Thanks for the explanation. Yes, the deprecation notice is nice if someone was using the previous syntax `['data-my-attribute' => '']` to let them know about the change in behaviour. However, this deprecation notice now prevents me from using this syntax in the intended way. So the deprecation / change in behaviour isn't the problem, but the notice itself is. I can turn off the notice in production, but I don't want to do that in my development environment – I want to see deprecation notices to be able to fix them, after all. And since those deprecation notices trigger errors (I think this is because [Whoops](https://github.com/filp/whoops) converts those notices to errors?), I can't just ignore them either. So I have to add some unnecessary checks to my code just to get rid of the deprecation notice that will be gonewhen 3.10 is released, anyway.

In any case, since 3.9.0.0 was a major release, the behaviour could've just been changed and mentioned in the upgrade docs as a breaking change. After all, deprecations are usually used for minor updates, so developers can prepare for the next major version, where it's understood and clearly documented that breaking changes will occur.

Would be nice to get a solution for this in 3.9.1. `['class' => '']` should just work as intended and not trigger deprecation notices.

**Comment by lukasbestle** [comment]
@MoritzLost @neildaniels Thanks for your feedback and ideas!

As explained, we want to ensure a smooth migration process for any breaking changes, so we introduce deprecations one major version ahead of the actual breaking change. At least that's how we did it in the most recent major releases.

We agree that this particular deprecation notice is not useful for many use cases. So we have built a new way to [disable particular deprecation warnings](https://github.com/getkirby/kirby/pull/5014). This change will be in 3.9.1 and we will use it to disable the `Xml::attr()` deprecation warning by default.

We can also use that same feature in the future to introduce deprecations in minor releases (without warning) and then enable the warning in the next major release.

**Comment by nilshoerrmann** [comment]
So from what I understand, the deprecation warning is now silent, but the deprecation of the former behaviour is still in place.

We are using `Html::attr()` a lot to set attributes dynamically and have so far relied on either empty or `null` values to not render attributes. The old behaviour of allowing empty string to not render an attribute was especially helpful when the attribute was based on a field value: if the field was not filled (= empty string), the attribute was omitted.

For us, this change requires many changes across projects to cast empty strings to `null` when setting attributes dynamically.

Wouldn't it be a better option to use `true` to set empty attributes? `true` is not a value an html attribute accepts directly so it always needs to be casted to get rendered properly (either converting it to the string `"true"` or the integer `1`) and `true` correctly states that the attribute should truly been rendered (aside from XHTML, the attribute is not actually empty, it's boolean set or not set). Also, this would allow to use Kirby's toggle field to set empty attribute from the panel, e. g. having a field `autoplay` set to `true` and outputting it like so:

```php
Html::attr([
  'autoplay' => $page->autoplay()->toBool()
]);
```

lukasbestle:
@nilshoerrmann Thank you for your feedback. It's certainly not too late yet to reconsider this again.

A boolean is already supported for `Html::attr()`. `true` means "attribute without value" and `false` means "no attribute at all".

The issue however is in `Xml::attr()`. In many XML-based languages, boolean attributes are represented as `attribute="attribute"` with the value being equal to the name. So that's what `Xml::attr()` currently does when you pass it `true`. If we would treat an empty string as a missing attribute, there would no longer be a way to generate an empty attribute value with `Xml::attr()`.

However you are completely right that the current status quo is not optimal either. The use case of passing an empty string from a field makes a lot of sense.

To be honest I feel like we are bit caught between two stools. We deprecated the empty string because of the request of a user who needed the exact opposite behavior and expected that an empty string renders an empty attribute.

neildaniels:
I'd vote for just reverting back to the original behavior and declaring that `Xml::attr()` does not output something like `autoplay controls`. The more verbose format of `autoplay="autoplay" controls="controls"` is valid.

If you want the extremely concise version, just put that directly in your code? Or use `implode(' ', ['autoplay', 'controls'])`?

```
<video <?= Html::attr(['autoplay' => '', 'controls' => '']) ?>></video>
```
 seems like a really roundabout way to just write
```
<video autoplay controls></video>
```

nilshoerrmann:
@lukasbestle I understand the need to allow empty values and boolean attributes for different use cases. The question to me is: which is the most common one.

The issue that initialised this change asked for boolean attributes like `controls` and such. It‘s important to note that and empty value doesn‘t equal a boolean state (neither in HTML nor in XML). For boolean states `attribute="value"` equals `attribute=""` as both is boolean true. You need to remove the attribute for boolean false.

The open question with the old implementation is: how do we create empty attribute as they may well be needed in some places? But the question is also: which is more common: empty attributes or boolean attributes? I‘d say boolean attributes are more common and should thus require less code to add them. But the new implementation in 3.9 gives less options to set boolean attributes.

As I say – from my work experience – the need for plain empty attributes is way lower than actual boolean attributes. Would it be an idea to offer a specific method for empty string values? Like:

```php
Html::attr([
  'data-empty-string' => Html::empty()
]);
```

MoritzLost:
@lukasbestle @nilshoerrmann @neildaniels None of the suggestions will be completely backwards compatible anyway, I've already updated my code for 3.9, and others will probably have to as well. Since there are different conflicting preferences, we should talk about what makes sense going forward. If we're talking merely about what makes the most sense, the behaviour should be _intuitive_. Attributes without values in HTML are called `boolean`. If I want a `boolean` value, it makes sense and is intuitive to pass a `boolean` as the value. If I want an attribute that takes a string value, I will pass it a string. If I pass an empty string, it makes sense to get an attribute with an empty string as its value. So the behaviour should be as follows:

```php
Html::attr([
    // class="my-class"
    'class' => 'my-class',

    // class=""
    'class' => '',

    // download
    'download' => true,

    // [no output]
    'download' => false,
]);
```

Not sure about `null`, I don't think null should be part of this interface at all, it doesn't have a natural counterpart in HTML attributes. But it's probably fine to treat it the same way as `false`, since that's the current behaviour.

This is also how Yii2's [`Html::renderTagAttributes()` treats its parameters](https://www.yiiframework.com/doc/api/2.0/yii-helpers-basehtml#renderTagAttributes()-detail).

Sidenote: I don't care about XML, alongside probably 99 % of users of this method. XML quirks shouldn't result in a worse interface for everyone using `Html::attr`. If XML really requires some different edge-case handling, put that in `Xml::attr`, but leave it out of `Html::attr`.

MoritzLost:
@lukasbestle Another related point: Regardless of syntax, it needs to be possible to explicitly create an attribute with an empty string value. For most attributes (like `class`), it doesn't matter if they're left out or included with an empty string value. But a notable exception is the `alt` attribute. An image with `alt=""` (empty string value) is used to communicate that an image is purely decorative and can be passed over by screenreaders. Leaving out the `alt` attribute completely in this case is a mistake. So an empty string should _definitely_ result in an attribute with an empty string value, not in the attribute being left out completely.

`['alt' => '']` => `alt=""`
`['alt' => false]` => `[no output]`

lukasbestle:
TBH I agree with @MoritzLost:

> Since there are different conflicting preferences, we should talk about what makes sense going forward. If we're talking merely about what makes the most sense, the behaviour should be _intuitive_. Attributes without values in HTML are called `boolean`. If I want a `boolean` value, it makes sense and is intuitive to pass a `boolean` as the value. If I want an attribute that takes a string value, I will pass it a string. If I pass an empty string, it makes sense to get an attribute with an empty string as its value.

That was the goal with the new implementation in 3.9 and the reason why we changed the behavior of the empty string from "no output" to "attribute with an empty value".

I think the `alt=""` example pretty much sums it up quite well. Before 3.9, the only way to achieve that was to pass a string with a single space (`['alt' => ' ']`). That was super weird and not intuitive at all. And then again it opens the question how a user would create an attribute where the value is actually a single space. There are certainly examples where this may be necessary (even if they are not very common).

I feel like it won't help us to decide based on which type of value is most common. It needs to be possible to create every output that's possible in XML and HTML. Moritz' comments show the reasons for that very well.

Now the question is if we can find a solution to ease the transition for cases where a possibly-empty field value is passed as an attribute value.

@nilshoerrmann's suggestion could work, maybe adapted to use a constant like `Html::EMPTY`. But it would mean that passing an actual empty string will lead to a result that is not always what's expected (omitting the attribute entirely). So I think it would be even better to have a solution that works the other way around: Empty string meaning an empty value by default and some other syntax to tell the method to treat an empty string as missing attribute instead.

nilshoerrmann:
I think @MoritzLost‘s arguments are convincing, too, especially with the `alt` attribute example.

It would be nice to either have an additional parameter on the `attr` method to remove empty attributes or to have field methods that conditionally return values, e. g. `valueOrFalse()` (for text-based fields) or `emptyOrTrue()` and `emptyOrFalse()` (for toggle fields). But the latter might be too specific as a core feature.

lukasbestle:
I like the idea to solve this with field methods as it makes the dev's intention most clear. Also we wouldn't need to add new arguments or any code at all really.

For toggle fields there is already `$field->toBool($default)`. If the field is empty, the provided default will be used.
For other kinds of fields, you can use `$field->or(null)->value()` for the same effect.

nilshoerrmann:
> For other kinds of fields, you can use $field->or(null)->value() for the same effect.

Oh, wow, I totally wasn't aware that the `or` method allowed for something like this. It would be great to have some more examples on the methods and on the `attr` helper docs pages!

lukasbestle:
@nilshoerrmann I've added examples to the reference of the `or()` and `toBool()` field methods and the `Xml::attr()` and `Html::attr()` methods. Do these solutions resolve the issue for you?

nilshoerrmann:
Well, this won‘t help us upgrading much but I think it resolves the issue for Kirby and improves the situation for other developers – so yes, I‘m fine with it 👍

Issue #4909, opened 2022-12-08
Structure-field : Use props.fields instead of attrs.fields for building the Form.


Daandelange:
Hello, I'm reopening a ticket for #4107 as it got lost in stale issues.

## Description
The structure field currently uses `$this->attrs['fields']` to build the form. I think it should use `$this->props['fields']`.
https://github.com/getkirby/kirby/blob/3e5899b628a2b59c1833a8d79eda292a7fe06843/config/fields/structure.php#L176

The purpose of my request is to make the `structure-field` better re-useable. Unless you really need to use the user-provided attributes from the blueprint, I think it makes more sense to use `$this->props['fields']` (or `$this->fields` to accept computed fields too) instead of `$this->attrs['fields']` so that the StructureField is easier to extend, with less redondant code. (see snippet below)

If other people are running into this issue, the easiest way I found is to add a `form` to your extended-structure-field's `methods`. This will also ensure compatibility with current Kirby version (3.8.3) and before.

```php
// See comments for different workarounds.
Kirby::plugin('kirbybugs/staticstructure', [
  'extends' => 'structure',
  'props' => [
    // Make fields static or whatever
    'fields' => function($fields=[]){
      //$this->attrs['fields']=[...]; // Solution 2 : Here you could eventually force the prop into the attrs so that the native methods.form picks the right one.
      return [ 'test'=>['label'=>'TestField','name'=>'test', 'type'=>'text'] ];
    },
  ],
  'methods' => [ // Solution 1
    // Overrides structure-field::methods::form() which grabs the form from $this->attrs['fields'] instead of the props.
    // Until https://github.com/getkirby/kirby/issues/4107 is merged, and after, this will still be needed this for compatibility with older kirby versions.
    'form' => function (array $values = []) { 
      return new Form([
        'fields' => $this->fields, // <-- only line changed compared to native function. Fix proposal.
        'values' => $values,
        'model'  => $this->model
      ]);
    },
  ],
]);
```
**Expected behavior**  
I assume/expect that the panel Form is build with my modified props when extending the structure-field.

## Proposed solution
Changing `$this->attrs['fields']` to `$this->fields` solves the issue in `kirby/config/fields/structure.php`.

## To reproduce
Use the snippet above without the `'methods'` part, put the field in a blueprint and go to the corresponding panel page. See the structure form fields corresponding to the page blueprint fields (user-fields); not the static fields injected in `props['fields']. The template code works as expected however.

## Your setup
[...]

PR #4936, opened 2022-12-11
Structure field: use `$this->fields`

distantnative:
## This PR …

This PR is one of these that I am a bit concerned of because our field setup is a mess (= hard to understand) and I can't trust unit tests to have our back as much here. From my understanding the change should be ok though: While I first thought it would result in a loop, as the computed `fields` also calls the `form()` method, this does not seem to be the case. When the computed prop function is called, `$this->field` contains the result of the normal `props` function which afterwards gets overwritten by the result from the computed function.

### Refactoring
- Structure field uses `$this->fields` instead of `$this->attrs['fields']`
#4909

## Ready?

- [ ] ~~Unit tests for fixed bug/feature~~

### For review team

### DIFF
--- a/config/fields/structure.php
+++ b/config/fields/structure.php
@@ -173,7 +173,7 @@ return [
 		},
 		'form' => function (array $values = []) {
 			return new Form([
-				'fields' => $this->attrs['fields'],
+				'fields' => $this->fields,
 				'values' => $values,
 				'model'  => $this->model
 			]);


bastianallgeier:
Uh, I'm concerned about this one as well. I think I remember that there was a reason to use attrs, but I cannot figure out why at the moment. I'm quite confident though that there's a high risk of a regression.

distantnative:
Would you think `$this->props['fields']` is better?

bastianallgeier:
No, I think it's exactly the same problem.

PR #4941, opened 2022-12-11
Disabled button: improve accessibility

distantnative:
So far, disabled buttons don't exists for screenreader. Which is bad when we communicated things like status etc. via disabled buttons. So we need to improve this.

## TODO
- [ ] @bastianallgeier focus state needs some styling I think. How to make it work no matter what context the disabled button is in?

### DIFF
--- a/panel/src/components/Navigation/ButtonDisabled.vue
+++ b/panel/src/components/Navigation/ButtonDisabled.vue
@@ -1,15 +1,16 @@
 <template>
-	<span
+	<button
 		:id="id"
 		:data-disabled="true"
 		:data-responsive="responsive"
 		:data-theme="theme"
 		:title="tooltip"
 		class="k-button"
+		aria-disabled="true"
 	>
 		<k-icon v-if="icon" :type="icon" :alt="tooltip" class="k-button-icon" />
 		<span v-if="$slots.default" class="k-button-text"><slot /></span>
-	</span>
+	</button>
 </template>
 
 <script>


panel/src/components/Navigation/ButtonDisabled.vue
```diff
@@ -1,15 +1,16 @@
 <template>
-	<span
+	<button
 		:id="id"
 		:data-disabled="true"
 		:data-responsive="responsive"
 		:data-theme="theme"
 		:title="tooltip"
 		class="k-button"
+		aria-disabled="true"
```
bastianallgeier, line 9: Shouldn't this just be disabled instead?

distantnative: No, screenreaders will ignore it then. As I read it, it'd be a bad practice in terms of accessibility.

bastianallgeier: Hm, so we basically need two types of disabled buttons, right? Those who should not be read, because it doesn't make sense (i.e. deactivated in dropdowns) and those who still need to convey a message (i.e. status)

distantnative: But why should they be visible but disabled for some people but completely non existent for screenreaders? Shouldn't they just as much get the option to focus it and get read out that there is this button/entry but disabled?

bastianallgeier: Difficult question. There are quite a few folks who hate that we show disabled buttons. I've always belonged to the group who's advocating to show all available options, even if they are disabled. But I don't know how comfortable this is as a blind user. If I can quickly scan the list, I can ignore the disabled ones quite easily. But if I need to navigate through them one by one, it seems quite annoying.

distantnative: Not sure if more annoying or better than not being aware of its existence at all. Too much to assume for me not having any experience as screenreader user.

But the current situation needs to change. Don't think just about drop-down menu, but the status flag is the perfect example that even when disables still is there to communicate information. It must not continue that screenreaders are totally kept away from that information.

distantnative: Or put different: why are we treating our users differently? For those without screenreader we decided it's important that they still see the buttons, dimmed. But for those with screenreaders we decide that shouldn't be the case? I could see arguments for both ways but then we should do it the same way for all users.

bastianallgeier: Your last comment really convinced me. Then we should get this solution done for the 3.9 launch.

bastianallgeier:
Ürks, this is turning into a rabbit hole. 

- When disabled buttons are no longer spans, the switch between ButtonNative and ButtonDisabled is completely useless
- Disabled state could be handled in ButtonNative and ButtonLink which would simplify the setup
- But this means that listeners must not be applied to such disabled elements
- Focus style is currently handled with the data-tabbed implementation, but it has some quirks in the CSS. We should solve this with focus-visible instead, which is now widely supported. 
- The status-icon button component has its own styling rules, which don't place nicely with just improving the focus state for the button component. 

This is one of those places, where we can either fix it quickly for 3.9 and leave the focus styles pretty much as they are. Or we get it right but also refactor our Button component.

distantnative:
I think I'd rather get it right in 3.9.1 - it has been like this for years, probably makes not much difference if now a couple weeks more. But then much better.

lukasbestle:
If we refactor the button component (removing `ButtonDisabled` and merging the behavior into `ButtonNative` and `ButtonLink`), wouldn't this still be something for a major release like 3.9 because of the breaking change?

distantnative:
@lukasbestle I think you are right that this would need to be considered - whether it works for 3.9.1 or 3.10.0. But I wouldn't rush it for 3.9.0

distantnative:
Closing this in favour of https://github.com/getkirby/kirby/pull/5003

PR #4942, opened 2022-12-11
Improvements for `Toolkit\A`

distantnative:
Removing some loops in favour of native functions

### DIFF
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -114,9 +114,16 @@ class A
 
 					// the element needs to exist and also needs to be an array; otherwise
 					// we cannot find the remaining keys within it (invalid array structure)
-					if (isset($array[$currentKey]) === true && is_array($array[$currentKey]) === true) {
+					if (
+						isset($array[$currentKey]) === true &&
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):

src/Toolkit/A.php
```diff
@@ -467,13 +480,7 @@ public static function move(array $array, int $from, int $to): array
...
 	public static function missing(array $array, array $required = []): array
 	{
-		$missing = [];
-		foreach ($required as $r) {
-			if (isset($array[$r]) === false) {
-				$missing[] = $r;
-			}
-		}
-		return $missing;
+		return array_values(array_diff($required, array_keys($array)));
```
lukasbestle, line 483: I think this is a breaking change. A value like `['cat' => null]` was treated as missing before, but isn't anymore with `array_keys()`. I guess the expected behavior depends on the use case, but we should document this.

lukasbestle: I've added it as a breaking change to the 3.9.0-rc.1 release notes.

PR #4959, opened 2022-12-16
Improve `Toolkit\Xml`

distantnative:
## Breaking changes

- `Html::tag()` and `Xml::tag()` no longer accept a `null` value for the `$attr` argument. The default value has been changed to an empty array. If you explicitly pass `null` for this argument, please change it to an empty array as well.

### DIFF
--- a/src/Toolkit/Xml.php
+++ b/src/Toolkit/Xml.php
@@ -18,10 +18,8 @@ class Xml
 {
 	/**
 	 * HTML to XML conversion table for entities
-	 *
-	 * @var array
 	 */
-	public static $entities = [
+	public static array|null $entities = [
 		'&nbsp;' => '&#160;', '&iexcl;' => '&#161;', '&cent;' => '&#162;', '&pound;' => '&#163;', '&curren;' => '&#164;', '&yen;' => '&#165;', '&brvbar;' => '&#166;', '&sect;' => '&#167;',
 		'&uml;' => '&#168;', '&copy;' => '&#169;', '&ordf;' => '&#170;', '&laquo;' => '&#171;', '&not;' => '&#172;', '&shy;' => '&#173;', '&reg;' => '&#174;', '&macr;' => '&#175;',
 		'&deg;' => '&#176;', '&plusmn;' => '&#177;', '&sup2;' => '&#178;', '&sup3;' => '&#179;', '&acute;' => '&#180;', '&micro;' => '&#181;', '&para;' => '&#182;', '&middot;' => '&#183;',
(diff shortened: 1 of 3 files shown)


src/Toolkit/Xml.php
```diff
@@ -66,13 +64,13 @@ class Xml
...
-	 * @param string|array $name String: A single attribute with that name will be generated.
-	 *                           Key-value array: A list of attributes will be generated. Don't pass a second argument in that case.
+	 * @param string|array|null $name String: A single attribute with that name will be generated.
+	 *                                Key-value array: A list of attributes will be generated. Don't pass a second argument in that case.
 	 * @param mixed $value If used with a `$name` string, pass the value of the attribute here.
 	 *                     If used with a `$name` array, this can be set to `false` to disable attribute sorting.
 	 * @return string|null The generated XML attributes string
 	 */
-	public static function attr($name, $value = null): string|null
+	public static function attr(string|array|null $name, $value = null): string|null
```
afbora:
```suggestion
	public static function attr(string|array|null $name, mixed $value = null): string|null
```

src/Toolkit/Xml.php
```diff
@@ -408,9 +397,6 @@ public static function tag(string $name, $content = '', array $attr = null, stri
 
 	/**
 	 * Properly encodes tag contents
-	 *
-	 * @param mixed $value
-	 * @return string|null
 	 */
 	public static function value($value): string|null
```
afbora, line 405:
```suggestion
	public static function value(mixed $value): string|null
```

src/Toolkit/Xml.php
distantnative: We decided not to add the mixed type hint anymore as it doesn't add any real benefit while still making PHP complain e.g. when a child class doesn't declare it.

afbora: Got it! Then if there is no type it means mixed type.

afbora: What if we add mixed to subclasses? Especially useful in IDE.

distantnative: We don't always have control of subclasses (plugins etc). And what would be the benefit for IDEs? Mixed means as much as "we don't know which type"

afbora: When calling methods in the IDE, it's very helpful to see what type of argument they want and their return type. I don't fully understand why we don't want to use the `mixed` type that came officially in PHP 8.0. It's not a big deal though 👍

distantnative: But `mixed` isn't any type - so the IDE doesn't show you any useful info. `mixed` says "we don't know".

And PHP doesn't like it a parent class adds `Mixed` but a subclass doesn't. We accept that risk for when type hinting adds benefit - but in the case of mixed it doesn't add any benefit, so we said we'd rather leave it.

afbora: Understand, it's "pseudo" type.

afbora (dismissed):

src/Toolkit/Xml.php
lukasbestle: How can `$name` be `null`? Won't this break on string concatenation? Also, what would the use case be to pass `null`?

distantnative: I remember again - our unit tests start failing a lot when not allowing for `null`: https://github.com/getkirby/kirby/actions/runs/3786702322/jobs/6437840246

distantnative: I assume those currently work, because they also get caught by `$value === null`

lukasbestle: All of these erroneous calls come from `Toolkit/Xml.php:373` (`Xml::tag()`, which accepts `null` for the `$attr` argument).

I'm pretty sure there is no reason to support `null` for this argument either. So I think we should change the default to this:

```diff
- public static function tag(string $name, $content = '', array $attr = null
+ public static function tag(string $name, $content = '', array $attr = []
```

Same for the `Html` class. This is a breaking change, but I think it makes sense. Means that we should merge this for 3.9.0 though.

distantnative: There needed to be one additional line changed which calls `::tag()` with all arguments. But I think then it works. Can you have a look? And then we should add it as breaking change.

lukasbestle (approved):
Looks good to me. 👍

PR #4961, opened 2022-12-17
`Toolkit\Str`: clean up

distantnative:


src/Toolkit/Str.php
```diff
@@ -323,12 +281,9 @@ public static function contains(string $string = null, string $needle, bool $cas
 	 * Convert timestamp to date string
 	 * according to locale settings
 	 *
-	 * @param int|null $time
-	 * @param string|\IntlDateFormatter|null $format
 	 * @param string $handler date, intl or strftime
-	 * @return string|int
 	 */
-	public static function date(int|null $time = null, $format = null, string $handler = 'date')
+	public static function date(int|null $time = null, string|IntlDateFormatter $format = null, string $handler = 'date')
```
afbora:
```suggestion
	public static function date(int|null $time = null, string|IntlDateFormatter $format = null, string $handler = 'date'): string|int|false
```

afbora (dismissed):

src/Toolkit/Str.php
```diff
@@ -499,9 +442,6 @@ public static function excerpt($string, $chars = 140, $strip = true, $rep = ' 
 	/**
 	 * Convert the value to a float with a decimal
 	 * point, no matter what the locale setting is
-	 *
-	 * @param string|int|float $value
-	 * @return string
 	 */
 	public static function float($value): string
```
```suggestion
	public static function float(string|int|float $value): string
```

distantnative: This is true based on the removed docblock but not quite true to reality, e.g. at least `null` has been also supported.

lukasbestle: Good point, `null` as well. But that should be it then, all other types will break the code I think.

lukasbestle (approved):

PR #4963, opened 2022-12-17
Snippets with slots: Enhancements and fixes

lukasbestle:
### Deprecated

- Manually passing `$slot` or `$slots` as data variables to snippets is deprecated. In a future Kirby version, those variables will be overridden with the `Slot` and `Slots` objects.

### Enhancements (only for RC.2 changelog)

- The variable scope inside a snippet is the same no matter if the snippet was called with or without slots. This means that the `$slot` and `$slots` objects are also present when no slots were passed.

### Fixes (only for RC.2 changelog)

- Calling a snippet that does not exist in `slots` mode no longer throws a PHP error but results in an empty snippet output like with unslotted snippets

### Refactoring (only for RC.2 changelog)

- Improved code quality and tests of the `Template` package

## Ready?

### For review team


lukasbestle:
@distantnative @bastianallgeier I'm wondering if we can ease the breaking change of the `$data`, `$slot` and `$slots` variables. This breaking change was already in the slots implementation, but now it takes effect for ordinary snippets as well.

Suggestion: We could let users overwrite these variables in the data array but trigger a deprecation warning. In 3.10 we could then switch the merge order so that these variables are always set by Kirby. What do you think?

distantnative:
For my understanding, what's the main benefit of passing $slot and $slots now to regular snippets as well? Does it outweighs the breaking change that we're overwriting existing data values?

lukasbestle:
> For my understanding, what's the main benefit of passing $slot and $slots now to regular snippets as well?

The idea originated in [this discussion on Discord](https://discord.com/channels/525634039965679616/1052241563528462448/1052951708029956096), continued [here](https://discord.com/channels/525634039965679616/1052241563528462448/1052956115794677790).

TL;DR: If you want to call a slot snippet without slots, the snippet will break if it tries to access the `$slots` object. This can only be circumvented with inelegant syntax like `if (isset($slots) && $header = $slots->header())`. So e.g. snippets defined in plugins that don't know how they will be called would need to adapt to this special case.

If we always pass the same variables to the snippet, slots and data can be used in a flexible way with the same snippets. The creator of the snippet doesn't need to think about the different behavior of slot snippets. Kirby would behave consistently no matter how the snippet is called.

Even if we ignore the `$slot` and `$slots` vars, there is still the `$data` variable that is also only available in snippets with slots. This is an inconsistency that feels like a bug to me.

I think the solution to deprecate those variables in 3.9 is really manageable and will lead to a clean setup in the long term compared to the status quo with the inconsistency.

distantnative:
Ok got it. Thanks for the summary 👍🙏

My concern would be that it makes the whole system likely harder to debug. I don't think that many snippets that were designed to be used with slots will make much sense to be called without any. Right now, the error messages would point people towards that something isn't set up properly. If we just fill the non-defined slots with null, this will make it a lot harder to realize that something isn't right.

distantnative:
But if you all think that using the same snippet with and without slots is an actual thing, I'm ok with it.

lukasbestle:
The situation that a slot is missing can occur already: If your snippet expects a default slot but the caller defines a named slot (or the other way around).

lukasbestle:
What was the reason why we added the change to pass a fixed `$data` variable to snippets with slots? Is it necessary for the slots use case?

I'm asking because `$data` is a really common name and we even use it in two of our cookbook recipes. `$slot` and `$slots` are pretty specific on the other hand and it is unlikely that users will run into the breaking change for these two.

I'm considering to drop the `$data` var entirely and only pass `$slot` and `$slots`.

distantnative:
I'd be with you there. @bastianallgeier was there any specific reason?

lukasbestle:
I've removed the `$data` variable in the last commit. We can revert that commit if we decide that we want to keep the `$data` variable.

lukasbestle:
## Summary

Apart from refactoring and a small bug fix to the slots implementation, this PR also implements a new feature based on [this discussion on Discord](https://discord.com/channels/525634039965679616/1052241563528462448/1052951708029956096), continued [here](https://discord.com/channels/525634039965679616/1052241563528462448/1052956115794677790): All snippets now receive the `$slot` and `$slots` variables, even if the snippet was called without slots. This enables devs to add support for optional slots to their snippets, increasing compatibility and increasing behavior consistency.

I implemented the feature by using the `$snippet->scope()` method for all snippet calls, no matter if the snippet was called with or without slots.

There is one issue though: Using the `scope()` method 1:1 would be a breaking change for existing snippets as it would no longer be possible to pass the `$data`, `$slot` and `$slots` variables as data variables (they would get overridden by the `scope()` method during merging).

I solved this in two steps:

- For the `$slot` and `$slots` variables the merging order is flipped for 3.9. This means that these variables are in general passed to the snippet but can be overridden in the data array. But passing them is deprecated and the merge order will be flipped back in 3.10.
- I removed the `$data` variable completely in the `scope()` method. The `$data` variable is still available inside the snippet as it's injected by `F::load()`. But it can be used as a variable name in the data array (which is commonly done even in our docs and recipes). The name just feels too general for a reserved variable. If the dev really needs all variables in the snippet, they can use the `get_defined_vars()` function. It contains the variables `$slot`, `$slots` and some other internal variables as well, but also all custom data vars.

Everything described above is currently implemented in this PR.
What we could do is to add a new `Snippet::$data` variable that is always set to the data of the current snippet and could be used inside the snippet. But I wonder if it's necessary. I don't fully understand the use case why we added the `$data` variable to the `scope()` method in the first place, so I am waiting for feedback.

---

@bastianallgeier What do you think? In case something needs to be changed, I have more time this week, so I'd appreciate your feedback.

bastianallgeier (dismissed):
I cannot trace back when or where we used `$data` actively. I think I would prefer to deprecate it in 3.9 first before we remove it in 3.10. This seems more user-friendly for those who might actually use it. 

One thing I noticed in the PR is the order of the class properties. I think we should agree on something here. I normally try to put them in alphabetical order. I seem to have missed this here originally and now it's still not ordered. But we never really agreed on this anyway. We do it for methods though and I really like it there. I think it would be good to have a standard.

bastianallgeier:
Well, I actually didn't mean to approve immediately 🙈. Hit the wrong button. But I think it's fine, except of my suggestion to deprecate data later.

lukasbestle:
I can reorder the class properties. I think it makes sense to have all public ones first and then all protected ones. Then we can sort alphabetically inside the respective visibility category. There's a PHP-CS-Fixer rule for that, but we haven't enabled it yet because it would create massive, hard to review diffs with our current code base. I fully agree we should already respect the order manually.

lukasbestle:
> I cannot trace back when or where we used `$data` actively. I think I would prefer to deprecate it in 3.9 first before we remove it in 3.10. This seems more user-friendly for those who might actually use it.

@bastianallgeier I tracked it down now: You added the override in commit 08b868142346f5700a810dd6a8cc8064d2ec09e4. Back then the class was called `Template\Container` and the variable was called `$props`. Nico then refactored the class to be called `Snippet` and the variable to be called `$data` once we decided that we want "snippets with slots" (commit 62a7737e712853c1e18718f9359c1c7f80f86747).

I can't find any place in traditional snippets that has overridden the `$props` or `$data` variables, so it seems to have been a new feature just for the `Container` class (and therefore just for the slots use case).

So if I didn't miss anything, there is nothing to deprecate here as the feature was never in a stable release.

src/Template/Slot.php
```diff
@@ -21,14 +21,13 @@ class Slot
...
+	protected bool $open = false;
 
 	/**
 	 * Creates a new slot for the given snippet
 	 */
 	public function __construct(
-		public Snippet $snippet,
-		public string $name,
+		protected string $name,
 		public string|null $content = null
```
distantnative: Why can't both be public? Doesn't make most sense to me to have one protected, one public.

distantnative (commented):
I see it like Lukas, I don't think there is any breaking change/deprecation, this is all new code.

I had another look though and am wondering if all the protected props are that useful. Especially when we have sometimes one prop public, one protected without really an obvious reason. Can't they just be public, where is the real benefit making them protected?

src/Template/Slot.php
lukasbestle: `$content` needs to be "internally public" because it is set from the `Snippet` class. `$name` on the other hand is supposed to be readonly.

The reason I made the props `protected` is because we use protected props basically all around the core. Only very recent PRs use `public` props just to make testing easier (and I'm really not a fan of that as we discussed before). But public props at the same time reduce the real-world reliability as it's very easy to override them by accident. Also they would all end up in auto-generated docs (e.g. IDE code suggestions), which would make it seem like we actually *intend* the props to be set from the outside. Access from tests should IMO never be more important than the intended semantics.

distantnative: Ok, I agree with that.

Just wondering about property order then. In this specific case, it would seem a bit odd that content comes before name, to me at least.

bastianallgeier: This is a thing I was wondering about in other places with the new PHP 8 style. Using the property shortcuts seems super useful at first, but we also lose a bit of control over here. But I think this is something we have to find out bit by bit with more use cases. The property order in the constructor should first be determined by logic/human aspects. If it does not fit the desired property order in the class, we should maybe abandon the shortcuts in such cases.

bastianallgeier:
@lukasbestle yay, thanks for investigating. I really thought that the $data key has been there before and is just one of our old forgotten artifacts of ancient snippet design. Then I'm totally fine to throw it out as well.

src/Template/Slot.php
lukasbestle: I agree. I'd move the properties to "real" properties in this case. This also allows to easily add an `@internal` hint without blowing up the constructor arguments.

lukasbestle:
I have moved the constructor props to real props to increase the overview. I think constructor props are only really useful if they are the only props of a class. Once we have other props, it gets confusing because you can easily miss that there are more props that are defined by the constructor.

I've also sorted the props according to the rules "first public, then protected" and "alphabetical sorting within the visibility categories".

The PR is now ready for review again.

bastianallgeier:
I love this separation. We can use the shortcuts when there are not other props. Otherwise we stick with the traditional style.

bastianallgeier (approved):
I like it a lot now.

PR #4969, opened 2022-12-22
Fix date field year change issue

### LINKED ISSUE #4968: [Kirby 3.8.3] Date field `Maximum call stack size exceeded` error
Reported by `@vikwinchester` from Discord.

afbora:
## This PR …
This issue was related about casting.  I'm not sure what kind of casting problem the removal of the `v-model` caused, but when the year changed, the data returns as a `string` and I fixed it by casting with the `Number()` method.

### Fixes
- Date field `Maximum call stack size exceeded` error #4968

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Calendar.vue
+++ b/panel/src/components/Forms/Calendar.vue
@@ -9,14 +9,14 @@
 					:disabled="disabled"
 					:required="true"
 					:value="current.month"
-					@input="current.month = $event"
+					@input="current.month = Number($event)"
 				/>
 				<k-select-input
 					:options="years"
 					:disabled="disabled"
 					:required="true"
 					:value="current.year"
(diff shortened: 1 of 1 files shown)


distantnative:
We need it for current.month then probably as well

afbora:
I guess, might be not needed. We can cast it to be sure anyway. Because we have calculation for the year but not for the month value:
https://github.com/getkirby/kirby/blob/fix/4968-date-field-error/panel/src/components/Forms/Calendar.vue#L181-L182

afbora:
I've also implement for the month anyway to be sure 👍

distantnative:
I think that's right as both used `v-model.number` before

distantnative (approved):

PR #4973, opened 2022-12-27
PHPUnit: Prevent creation of stray site directory

lukasbestle:
### Refactoring

- More of our PHPUnit tests use a temporary directory now to prevent tests from creating a stray `site` directory in the top-level of the project

## Ready?

### For review team

### DIFF
--- a/tests/Cms/Files/FilesTest.php
+++ b/tests/Cms/Files/FilesTest.php
@@ -122,7 +122,7 @@ class FilesTest extends TestCase
 	 */
 	public function testFindByUuid()
 	{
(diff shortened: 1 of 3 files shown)


afbora:
I see a lot of uses for `new App()`. Do we need to check other tests as well? What should we pay attention to?

lukasbestle:
With these changes there is no stray `site` directory anymore, which means that all other uses of `new App()` don't try to create directories.

Ideally we would migrate all tests to use temporary directories, but that can happen slowly.

distantnative (approved):

PR #4983, opened 2023-01-02
Fix hook parameter for `page.changeSlug`

### LINKED ISSUE #4981: page.changeSlug:before ->languageCode parameter always return null
## Description

Following https://getkirby.com/docs/reference/plugins/hooks/page-changeslug-before, `languageCode` should return the language code as string, but it's always `null`.

**Expected behavior**  

languageCode returns the correct languageCode.

## To reproduce

Setup a multilanguage website in Kirby,
add this hook, and check for the languageCode parameter.

```php
'page.changeSlug:before' => function ($page, $slug, $languageCode) {
    ray($page);
    ray($slug);
    ray($languageCode);
},
```

## Your setup
**Kirby Version**  
3.8.3

 - Device: Mac
 - OS: macOS 12.6.1

**Additional context**  

After a little bit of digging I found out languageCode is never set here: https://github.com/getkirby/kirby/blob/7dcf359e3da2fa09923a3e0d851f2a74ae668e78/src/Cms/PageActions.php#L214

Changing `$languageCode` to `$language->code()` here: https://github.com/getkirby/kirby/blob/7dcf359e3da2fa09923a3e0d851f2a74ae668e78/src/Cms/PageActions.php#L226 seems to fix it. But I don't really "understand" the flow, and assume it should come in as a parameter somehow (which it doesn't, it's always `null`).

distantnative:
### Fixes
- The `page.changeSlug` hook now receives the correct `$languageCode`
#4981

## Ready?

### For review team

### DIFF
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -223,7 +223,7 @@ trait PageActions
 			throw new InvalidArgumentException('Use the changeSlug method to change the slug for the default language');
 		}
 
-		$arguments = ['page' => $this, 'slug' => $slug, 'languageCode' => $languageCode];
+		$arguments = ['page' => $this, 'slug' => $slug, 'languageCode' => $language->code()];
 		return $this->commit('changeSlug', $arguments, function ($page, $slug, $languageCode) {
 			// remove the slug if it's the same as the folder name
 			if ($slug === $page->uid()) {


lukasbestle (approved):
I think we could merge this into `develop` for 3.9.0.

PR #4986, opened 2023-01-04
New ghost page class

bastianallgeier:


afbora:
This is a temporary solution that could cause potential issues. It requires a more comprehensive and stable solution. So needs to be closed.

PR #4989, opened 2023-01-05
Add asset methods for plugins

tobimori:
## This PR …
Adds asset methods that work similar to file methods or page methods, just for use with the `Kirby/Filesystem/Asset` class that's e.g. used by the `asset()` helper.

### Why is this important?
There are many plugins out there that can e.g. apply image transforms such as [johannschopplich/kirby-blurry-placeholder](https://github.com/johannschopplich/kirby-blurry-placeholder) or my own [kirby-blurhash](https://github.com/tobimori/kirby-blurhash) that one might want to use on a static image, but they only work conveniently with Kirby files itself due to file methods not being supported for simple assets.

See also: https://github.com/johannschopplich/kirby-blurry-placeholder/issues/15

This should also be in favor of Kirby, as the docs state: "Anything in your public path can be converted to an Asset object to use the same handy file methods as for any other Kirby files."
They obviously can't be allowed access to regular custom file methods, as those might access properties only available on the `File` object.

### What else to consider?

An alternative to this could be to allow `imageMethods` which extend the `Kirby/Image/Image` class. They would automatically inherit to assets and image files, but thus also allow less customization due to the lack of access to e.g.`media` path info.

### Breaking changes
None.

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/Asset.php
+++ b/src/Filesystem/Asset.php
@@ -3,6 +3,8 @@
 namespace Kirby\Filesystem;
 
 use Kirby\Cms\FileModifications;
+use Kirby\Cms\HasMethods;
+use Kirby\Exception\BadMethodCallException;
 
 /**
  * Anything in your public path can be converted
@@ -20,6 +22,7 @@ class Asset
 {
 	use IsFile;
 	use FileModifications;
+	use HasMethods;
 
 	/**
 	 * Relative file path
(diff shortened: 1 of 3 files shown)


tobimori:
hey, any update on this PR? is this something you generally don't want, or have you just not found the time to look at it yet?

src/Filesystem/Asset.php
```diff
@@ -38,6 +41,31 @@ public function __construct(string $path)
...
+
+		// asset methods
+		if ($this->hasMethod($method)) {
+			return $this->callMethod($method, $arguments);
+		}
+
+		// asset method proxy
+		if (method_exists($this->asset(), $method)) {
+			return $this->asset()->$method(...$arguments);
+		}
```
lukasbestle: Tiny issue: The order of these two cases should be swapped for consistency with `File::__call()`.

lukasbestle (commented):
Sorry for the delay, we were indeed quite busy with 3.9 and 3.9.1. We haven't forgot about your other PR either, but it's going to take a bit of time to review.

I really like the idea of asset methods. I'd say they fit in quite well conceptually. Image methods would have an overlap with the existing file methods, but asset methods are really separate and useful on their own.

src/Filesystem/Asset.php
tobimori: Updated :)

lukasbestle (dismissed):

lukasbestle (approved):

afbora (approved):
Loved new methods 💛

distantnative (approved):

PR #5008, opened 2023-01-17
Fix ParsedownExtra for emojis

distantnative:
### Fixes
- Emojis are now correctly treated again when using Parsedown Extra

## Ready?

### For review team

### DIFF
--- a/dependencies/parsedown-extra/ParsedownExtra.php
+++ b/dependencies/parsedown-extra/ParsedownExtra.php
@@ -573,6 +573,17 @@ class ParsedownExtra extends Parsedown
         # http://stackoverflow.com/q/11309194/200145
         $elementMarkup = htmlentities($elementMarkup);
 		$elementMarkup = htmlspecialchars_decode($elementMarkup);
+		// Migrating away from `mb_convert_encoding($elementMarkup,
+		// 'HTML-ENTITIES', 'UTF-8');` has cause emojis not to be
+		// converted into HTML codes. we're fixing this by
+		// treating the input string as UTF-8 and convert
+		// every Unicode character with 8 or more bits
+		// (= character code 128 or 0x80 up to the Unicode
+		// limit of 0x10ffff) to an entity; the third and
+		// fourth arguments are not needed for our use case
+		// and are set to the default values (no offset and
+		// a full mask)
+		$elementMarkup = mb_encode_numericentity($elementMarkup, [0x80, 0x10ffff, 0, 0xffffff], 'UTF-8');
 
         # Ensure that saveHTML() is not remove new line characters. New lines will be split by this character.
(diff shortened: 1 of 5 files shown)


dependencies/parsedown-extra/ParsedownExtra.php
```diff
@@ -573,6 +573,17 @@ protected function processTag($elementMarkup) # recursive
         # http://stackoverflow.com/q/11309194/200145
         $elementMarkup = htmlentities($elementMarkup);
 		$elementMarkup = htmlspecialchars_decode($elementMarkup);
```
lukasbestle: Shouldn't these two lines be removed then?

dependencies/parsedown-extra/ParsedownExtra.php
```diff
@@ -573,6 +573,17 @@ protected function processTag($elementMarkup) # recursive
...
+		// Migrating away from `mb_convert_encoding($elementMarkup,
+		// 'HTML-ENTITIES', 'UTF-8');` has cause emojis not to be
+		// converted into HTML codes. we're fixing this by
+		// treating the input string as UTF-8 and convert
+		// every Unicode character with 8 or more bits
+		// (= character code 128 or 0x80 up to the Unicode
+		// limit of 0x10ffff) to an entity; the third and
+		// fourth arguments are not needed for our use case
+		// and are set to the default values (no offset and
+		// a full mask)
```
lukasbestle: Fixing typos and rewording for an easier understanding:

```suggestion
		// Migrating away from `mb_convert_encoding($elementMarkup, 'HTML-ENTITIES',
		// 'UTF-8');` has caused multibyte characters like emojis not to be converted into entities,
		// which is needed so that the `DOM` extension can properly parse the markup.
		// The following line works like this: It treats the input string as UTF-8 and converts
		// every Unicode character with 8 or more bits (= character code starting at 128 or 0x80
		// up to the Unicode limit of 0x10ffff) to an entity; the third and fourth arguments for
		// the map are not needed for our use case and are set to the default values
		// (no offset and a full mask)
```

dependencies/parsedown-extra/ParsedownExtra.php
distantnative: I thought the snippet you gave is only for emoji. It replaces those two lines too?

lukasbestle: If I read the StackOverflow thread linked in the comment above, it only points to the fact that `DOMDocument` does not support UTF-8 characters, which is why everything non-ASCII was converted to an entity. This is what my new code does (converting everything with 8 bits or more to an entity, leaving ASCII alone).

lukasbestle (approved):

PR #5011, opened 2023-01-18
give `when` property of field access to parent fieldsets

rasteiner:
## This PR …
The `when` property of fields currently only has access to the fields of the fieldset to which it belongs to. 
This PR gives the query also access to fields in "parent" fieldsets (think nested structure fields, for example).

This happens by providing / injecting a reactive `context` object from fieldset to fieldset. Each of them merging the injected context with its own value. 

This behaviour mirrors the concepts of scoping and shadowing present in many programming languages. 

This is a draft because the current implementation lacks the serverside counterpart (about handling required hidden fields). But before proceeding I wanted to get some feedback on whether this could ever be merged or not. 

Thank you for your consideration :)

### Example
```yml
title: Default Page

fields:
  outer:
    type: structure
    fields:
      toggle:
        type: toggle
      
      inner:
        type: structure
        fields:
          info2:
            type: info
            text: >
              A nested info field that is always visible.
              There is also a second info field that shows when the outer toggle is yes
          info3:
            type: info
            text: A nested info field that is only visible when outer toggle is yes
            when:
              toggle: true
```

### Fixes
https://forum.getkirby.com/t/conditional-field-in-structure-not-working-what-am-i-doing-wrong/27552

### Breaking changes
Not sure. I think not. 

## Ready?
- [ ] Server side code

### For review team

### DIFF
--- a/panel/src/components/Forms/Fieldset.vue
+++ b/panel/src/components/Forms/Fieldset.vue
@@ -3,7 +3,7 @@
 		<k-grid>
 			<template v-for="(field, fieldName) in fields">
 				<k-column
-					v-if="$helper.field.isVisible(field, value)"
+					v-if="$helper.field.isVisible(field, context)"
 					:key="field.signature"
 					:width="field.width"
 				>
@@ -47,10 +47,23 @@
 </template>
 
 <script>
+import { computed } from 'vue';
+
 /**
  * The Fieldset component is a wrapper around manual field component creation. You simply pass it an fields object and all field components will automatically be created including a nice field grid. This is the ideal starting point if you want an easy way to create fields without having to deal with a full form element.
  */
 export default {
+	provide() {
+		return {
+			context: computed(() => this.context)
+		};
+	},
+	inject: {
+		injectedContext: {
+			from: "context",
+			default: {}
+		}
+	},
 	props: {
 		config: Object,
 		disabled: Boolean,
@@ -78,6 +91,14 @@ export default {
 		return {
 			errors: {}
 		};
+	},
+	computed: {
+		context() {
+			return {
+				...this.injectedContext,
+				...this.value
+			}
+		}
 	},
 	methods: {
 		/**
--- a/panel/src/components/Forms/Fieldset.vue
+++ b/panel/src/components/Forms/Fieldset.vue
@@ -47,7 +47,7 @@
 </template>
 
 <script>
-import { computed } from 'vue';
+import { computed } from "vue";
 
 /**
  * The Fieldset component is a wrapper around manual field component creation. You simply pass it an fields object and all field components will automatically be created including a nice field grid. This is the ideal starting point if you want an easy way to create fields without having to deal with a full form element.
@@ -97,7 +97,7 @@ export default {
 			return {
 				...this.injectedContext,
 				...this.value
-			}
+			};
 		}
 	},
 	methods: {


panel/src/components/Forms/Fieldset.vue
```diff
@@ -47,10 +47,23 @@
...
  * The Fieldset component is a wrapper around manual field component creation. You simply pass it an fields object and all field components will automatically be created including a nice field grid. This is the ideal starting point if you want an easy way to create fields without having to deal with a full form element.
  */
 export default {
+	provide() {
+		return {
+			context: computed(() => this.context)
+		};
+	},
+	inject: {
+		injectedContext: {
```
distantnative, line 62: I think you need to explain this to me, I don't quite understand what's happening here

panel/src/components/Forms/Fieldset.vue
```diff
@@ -3,7 +3,7 @@
 		<k-grid>
 			<template v-for="(field, fieldName) in fields">
 				<k-column
-					v-if="$helper.field.isVisible(field, value)"
+					v-if="$helper.field.isVisible(field, context)"
```
distantnative, line 6: Is `context` still a single value? It sounds like ti would be now multiple fieldsets/many values

rasteiner: `value` was the "value of the fieldset", in other words this always contained "many values". 
`context` however now combines the values of many fieldsets (it includes all parent fieldsets, "flattened" to a single object). This is how we give the `isVisible` helper access to fields in parent fieldsets. 

The `isVisible` interface doesn't change.

distantnative: What does happen when layers contain fields with the same name?

panel/src/components/Forms/Fieldset.vue
rasteiner: each fieldset provides a "context" (which combines the fields it received with its own fields) to child fieldsets. 
This is done here:
```js
provide() {
	return {
                // wrap in `computed` to make it a reactive variable
		context: computed(() => this.context)
	};
},
```
`this.context` refers to the computed value:
```js
computed: {
	context() {
		return {
			...this.injectedContext,
			...this.value
		};
	}
},
```

---- 

While injecting the context from parent fieldsets, we **rename** that provided property to `injectedContext`. It could also be called `parentContext`, if that makes it clearer. 
It's just to distinguish the injected (parent) context from the locally computed context (otherwise the computed method would infinitely call itself).

panel/src/components/Forms/Fieldset.vue
rasteiner:
```js
return {
	...this.injectedContext,
	...this.value
};
```

`this.value` shadows variables that might have the same name. This is how it's normally handled in programming languages which support scoping. Example in javascript:

```js
const variable = 'outer';
{
  const variable = 'inner';
  console.log(variable); // -> 'inner'
}
console.log(variable); // -> 'outer'
```

distantnative:
Thanks for explaining, I think I have a bit of a clue now how it works. But I remain quite sceptical that this is a sensible direction to head. The implementation seems to solve a problem for the few users that really understand how it works and can utilise it. For others, though it will be very confusing why in their `when` query they can assess a value of a field they might not even realize where it's coming from, instead of using something like `parent.toggle` or so.

rasteiner:
> The implementation seems to solve a problem for the few users that really understand how it works and can utilise it. For others, though it will be very confusing why in their when query they can assess a value of a field they might not even realize where it's coming from, instead of using something like parent.toggle or so.

I respectfully disagree. 
In my opinion, you don't have to understand how something works to successfully use it. I personally would find `parent.fieldname` much more confusing, since it's very different to what I, as web developer, am used to: simply having access to my variables that are in scope. 

`parent` reminds me personally of something like `parent::` in PHP (i.e. a base class). Or maybe even the "parent page" in Kirby. Definitely not "the closest outer scope". 

It's not like values just pop up from nowhere, users have to explicitly ask for them. They'd have to be confused about why their `when` property, which they have actively written like that, actually works? I find that hard to believe. 

I think [the linked thread](https://forum.getkirby.com/t/conditional-field-in-structure-not-working-what-am-i-doing-wrong/27552) shows what users are expecting (at least 1 of them). 

But hey; opinions... 

----

Patch for "only with `parent.` variant" to put into the computed method (untested):
```js
const renamed = Object.fromEntries(
    Object.entries(this.injectedContext).map(([k,v]) => ["parent." + k, v])
);
return {
    ...renamed,
    ...this.value
};
```

distantnative:
Apologies for the long wait. We discussed this today in the team again. In the end, we decided not to pursue this idea. It's a little too magical and bears the risk of really tricky side effects and edge cases. It also could be a barrier to move to a more profound solution in regards to storing content and more complex conditionals. We do agree that the scope of `when` is currently quite limited.

PR #5021, opened 2023-01-24
Fix snippets with `null` argument 

### LINKED ISSUE #5004: If null is passed to snippet helper, `$file` in snippet component is undefined
## Description
When calling `snippet()` with a dynamic method that returns a `null` value

```
snippet($page->field()->value());
```
`$file` in the snippet component is undefined. This error could be prevented by initializing `$file`.

## Your setup
**Kirby Version 3.8.4**

afbora:
## This PR …

I've implement @lukasbestle idea in https://github.com/getkirby/kirby/issues/5004#issuecomment-1399274649. But IMHO this is not bug, an enhancement.

### Enhancements
- Passing the snippet name argument as `null` is now handled gracefully #5004

### Breaking changes
None so far.

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1612,12 +1612,11 @@ class App
 	 * Uses the snippet component to create
 	 * and return a template snippet
 	 *
-	 * @param mixed $name
 	 * @param array|object $data Variables or an object that becomes `$item`
 	 * @param bool $return On `false`, directly echo the snippet
 	 * @psalm-return ($return is true ? string : null)
 	 */
-	public function snippet($name, $data = [], bool $return = true, bool $slots = false): Snippet|string|null
+	public function snippet(string|array|null $name, $data = [], bool $return = true, bool $slots = false): Snippet|string|null
 	{
 		if (is_object($data) === true) {
 			$data = ['item' => $data];
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1615,7 +1615,6 @@ class App
 	 * @param mixed $name
 	 * @param array|object $data Variables or an object that becomes `$item`
 	 * @param bool $return On `false`, directly echo the snippet
-	 * @return string|null
 	 * @psalm-return ($return is true ? string : null)
 	 */
 	public function snippet($name, $data = [], bool $return = true, bool $slots = false): Snippet|string|null
--- a/src/Template/Snippet.php
+++ b/src/Template/Snippet.php
@@ -157,11 +157,10 @@ class Snippet extends Tpl
 		array $data = [],
 		bool $slots = false
 	): static|string {
-		if ($name === null) {
-			return '';
-		}
-
-		$file = static::file($name);
+		// instead of returning empty string when `$name` is null
+		// allow rest of code to run, otherwise the wrong snippet would be closed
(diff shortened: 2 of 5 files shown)


src/Cms/App.php
```diff
@@ -1615,7 +1615,6 @@ public function smartypants(string $text = null): string
 	 * @param mixed $name
```
```suggestion
```

src/Cms/App.php
```diff
@@ -1615,7 +1615,6 @@ public function smartypants(string $text = null): string
...
 	 * @param array|object $data Variables or an object that becomes `$item`
 	 * @param bool $return On `false`, directly echo the snippet
-	 * @return string|null
 	 * @psalm-return ($return is true ? string : null)
 	 */
 	public function snippet($name, $data = [], bool $return = true, bool $slots = false): Snippet|string|null
```
lukasbestle: Any other value for `$name` will break in the component anyway, so it can already be blocked here:

```suggestion
	public function snippet(string|array|null $name, $data = [], bool $return = true, bool $slots = false): Snippet|string|null
```

src/Template/Snippet.php
```diff
@@ -153,10 +153,14 @@ public function endslot(): void
 	 * or the template string for self-enclosed snippets
 	 */
 	public static function factory(
-		string|array $name,
+		string|array|null $name,
 		array $data = [],
 		bool $slots = false
 	): static|string {
+		if ($name === null) {
+			return '';
```
lukasbestle: Won't this break when using this:

```php
snippet(null, slots: true);
echo 'Some text';
endsnippet();
```

I think we need to let the rest of the code run and rather change the following line:

```php
$file = $name !== null ? static::file($name) : null;
```

Could you please also add a unit test where a single snippet name is passed as `null` but in slots mode?

lukasbestle:
PS: Changed it to a regression as I think it passing `null` worked with the old implementation before 3.9.

afbora:
> PS: Changed it to a regression as I think it passing `null` worked with the old implementation before 3.9.

@lukasbestle I've tested `snippet(null)` on 3.7.x and 3.8.x and still getting error: `Undefined variable $file`

src/Template/Snippet.php
afbora: Yes, I didn't do it on purpose. Why are we running `slots` for a non-existent snippet? Is such a use correct or reasonable?

````php
snippet(null, slots: true);
echo 'Some text';
endsnippet();
````

What should the output of this be? Just `Some text`?

lukasbestle: I'd say the slot should be swallowed in this case as it's not intended to be printed directly. It's the same as if the slot is passed to an existing snippet but not used there (e.g. if the snippet is completely empty). So if we would print the slot for non-existing snippets, the behavior would be inconsistent. And potentially we would have issues with nested snippets (the wrong snippet would be closed).

afbora: > the wrong snippet would be closed

You're right. I missed that. I've pushed the changes https://github.com/getkirby/kirby/commit/6b5239ad4663170492e8d877870f3fa7c44b56ad

bastianallgeier (approved):
Really nice!

PR #5032, opened 2023-02-02
Some missing A::functions

adamkiss:
## This PR …
Adds some additional `A::` methods; These are basically just wrappers around matching `array_*` functions, but I often reach for them only to realise they are not there, and have to mix `array_thing(A::other($array))`.

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -161,7 +161,7 @@ class A
 	}
 
 	/**
-	 * Checks in array has a value
+	 * Checks if array has a value
 	 *
 	 * @param array $array
 	 * @param mixed $value
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -53,7 +53,7 @@ class A
 	 * @param array $array
 	 * @return int
 	 */
-	public static function count(array $array = []): int
+	public static function count(array $array): int
 	{
 		return count($array);
 	}
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -48,6 +48,21 @@ class A
 	}
 
 	/**
+	 * Counts the number of elements in an array
+	 *
+	 * @param array $array
+	 * @return int
+	 */
+	public static function count(array $array = []): int
+	{
+		if (!is_array($array)) {
+			throw new Exception('Invalid argument type: ' . gettype($array));
+		}
+
+		return count($array);
+	}
+
+	/**
 	 * Gets an element of an array by key
 	 *
 	 * <code>
@@ -198,7 +213,7 @@ class A
 				) {
 					$merged[] = $value;
 
-				// recursively merge the two array values
+					// recursively merge the two array values
 				} elseif (
 					is_array($value) === true &&
 					isset($merged[$key]) === true &&
@@ -206,7 +221,7 @@ class A
 				) {
 					$merged[$key] = static::merge($merged[$key], $value, $mode);
 
-				// simply overwrite with the value from the second array
+					// simply overwrite with the value from the second array
 				} else {
 					$merged[$key] = $value;
 				}
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -165,6 +165,28 @@ class A
 	}
 
 	/**
+	 * Checks in array has a value
+	 *
+	 * @param array $array
+	 * @return bool
+	 */
+	public static function has(array $array, $value): bool
+	{
+		return in_array($value, $array);
+	}
+
+	/**
+	 * Checks in array includes a value (alias for A::has)
+	 *
+	 * @param array $array
+	 * @return bool
+	 */
+	public static function includes(array $array, mixed $value): bool
+	{
+		return A::has($array, $value);
+	}
+
+	/**
 	 * Joins the elements of an array to a string
 	 */
 	public static function join(array|string $value, string $separator = ', '): string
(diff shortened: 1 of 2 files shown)


src/Toolkit/A.php
```diff
@@ -47,6 +47,21 @@ public static function apply(array $array, ...$args): array
...
+	 * @return int
+	 */
+	public static function count(array $array = []): int
+	{
+		if (!is_array($array)) {
+			throw new Exception('Invalid argument type: ' . gettype($array));
+		}
+
+		return count($array);
+	}
```
afbora: Argument is `array $array = []`  and control is `!is_array($array)` 😅

src/Toolkit/A.php
```diff
@@ -149,6 +164,28 @@ public static function get(
...
+	/**
+	 * Checks in array includes a value (alias for A::has)
+	 *
+	 * @param array $array
+	 * @return bool
+	 */
+	public static function includes(array $array, mixed $value): bool
+	{
+		return A::has($array, $value);
+	}
```
afbora: I don't think needed an alias method.

src/Toolkit/A.php
```diff
@@ -149,6 +164,28 @@ public static function get(
...
+	/**
+	 * Checks in array has a value
+	 *
+	 * @param array $array
+	 * @return bool
+	 */
+	public static function has(array $array, $value): bool
+	{
+		return in_array($value, $array);
+	}
```
afbora:
```suggestion
	public static function has(array $array, $value, bool $strict = false): bool
	{
		return in_array($value, $array, strict);
	}
```

src/Toolkit/A.php
adamkiss: Leftover from when I haven't realized php type checking will catch the error. Will remove

src/Toolkit/A.php
adamkiss: Convinced myself it's a good idea for people coming from js land. Will remove.

src/Toolkit/A.php
adamkiss: Will add with a matching test

distantnative:
Usually, I am against in adding to many core functions just as aliases, it can bloat up the toolkit a lot for little added value. However, I find these picked methods a good choice and wouldn't mind having them included :) thanks!

src/Toolkit/A.php
```diff
@@ -149,6 +160,19 @@ public static function get(
 		return $default;
 	}
 
+	/**
+	 * Checks in array has a value
```
lukasbestle:
```suggestion
	 * Checks if array has a value
```

tests/Toolkit/ATest.php
```diff
@@ -376,6 +401,56 @@ public function testShuffle()
...
+
+		$reduced = A::reduce($array, function ($carry, $item) {
+			return $carry . $item;
+		}, '');
+		$this->assertSame('miaowufftweet', $reduced);
+
+		$reduced = A::reduce([1, 2, 3], function ($carry, $item) {
+			return $carry + $item;
+		}, 0);
+		$this->assertSame(6, $reduced);
```
lukasbestle: To test that the third param is not ignored:

```suggestion
		$reduced = A::reduce([1, 2, 3], function ($carry, $item) {
			return $carry + $item;
		}, 42);
		$this->assertSame(48, $reduced);
```

adamkiss: That is a great idea

adamkiss (commented):
Applied Lukas' changes

src/Toolkit/A.php
```diff
@@ -47,6 +47,17 @@ public static function apply(array $array, ...$args): array
 		return $array;
 	}
 
+	/**
+	 * Counts the number of elements in an array
+	 *
+	 * @param array $array
+	 * @return int
+	 */
+	public static function count(array $array = []): int
```
lukasbestle: Just saw that this argument is optional. Why is that?

lukasbestle (commented):
Apart from one question this gets my approval.

src/Toolkit/A.php
adamkiss: You know, I don't think there is a great need for that - I think it's just automatic defensive default value on my end.

lukasbestle: I'd be totally fine with it. I was just confused because we don't do it for any other `A` method. We do it in the `Str` class in a few places and I always found that weird. After all the caller wants the method to do something, so why call it without arguments?

distantnative: I think we have it in some places for when you pass a field or method call directly to these method that could also yield `null` - that way the method doesn't crash into an error but gracefully still gets you, e.g. `0` in this case. Which is probably a valid result in most cases when you wanted to count something that got you `null`.

distantnative (dismissed):

src/Toolkit/A.php
lukasbestle: That's not what this does though. You cannot pass `null`, the default just allows you to call the method without any argument at all.

distantnative: True I think what I described we do in the Str class. Either it would need to be `= null` here or not optional.

lukasbestle: I think it should be not optional for consistency with all the other `A` methods.

distantnative: 👍

lukasbestle (approved):

distantnative (approved):

PR #5034, opened 2023-02-02
Regular expressions for `Str`

adamkiss:
## This PR …
introduces three new functions to `Str` Toolkit class - match and matchAll.

These are wrappers around default PHP with:
- default Str method order (string to with on first, arguments later)
- simplified working with result

with default `preg_match` (and all), you have to capture the matches (which is what I usually want/need) in a temporary value, while often just throwing away the return value of the original function.

This also includes `Str::matches($string, $pattern)` which is a simple syntactic sugar for simple matching string against a regular expression

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -522,6 +522,20 @@ class Str
 		return preg_replace('!^(' . preg_quote($trim) . ')+!', '', $string);
 	}
 
+	/**
+	 * Match string against a regular expression and return matches
+	 *
+	 * @param string $string The string to match
+	 * @param string $pattern The regular expression
+	 * @param int $flags Optional flags for the match
+	 * @param int $offset Optional offset for the match
+	 * @return array|null The matches or null if no match was found
+	 */
+	public static function match(string $string, string $pattern, int $flags = 0, int $offset = 0): ?array
+	{
+		$result = preg_match($pattern, $string, $matches, $flags, $offset);
+		return ($result > 0) ? $matches : null;
+	}
 
 	/**
 	 * Get a character pool with various possible combinations
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -527,23 +527,23 @@ class Str
 	 *
 	 * @param string $string The string to match
 	 * @param string $pattern The regular expression
-	 * @param int $flags Optional flags for the match
-	 * @param int $offset Optional offset for the match
+	 * @param int $flags Optional flags for PHP `preg_match()`
+	 * @param int $offset Positional offset in the string to start the search
 	 * @return array|null The matches or null if no match was found
 	 */
 	public static function match(string $string, string $pattern, int $flags = 0, int $offset = 0): ?array
 	{
 		$result = preg_match($pattern, $string, $matches, $flags, $offset);
-		return ($result > 0) ? $matches : null;
+		return ($result === 1) ? $matches : null;
 	}
 
 	/**
-	 * Syntax sugar for boolean string matching
+	 * Check whether a string matches a regular expression
 	 *
 	 * @param string $string The string to match
 	 * @param string $pattern The regular expression
-	 * @param int $flags Optional flags for the match
-	 * @param int $offset Optional offset for the match
+	 * @param int $flags Optional flags for PHP `preg_match()`
+	 * @param int $offset Positional offset in the string to start the search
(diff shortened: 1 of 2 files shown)


src/Toolkit/Str.php
```diff
@@ -522,6 +522,49 @@ public static function ltrim(string $string, string $trim = ' '): string
...
+	 * @param string $string The string to match
+	 * @param string $pattern The regular expression
+	 * @param int $flags Optional flags for the match
+	 * @param int $offset Optional offset for the match
+	 * @return array|null The matches or null if no match was found
+	 */
+	public static function match(string $string, string $pattern, int $flags = 0, int $offset = 0): ?array
+	{
+		$result = preg_match($pattern, $string, $matches, $flags, $offset);
+		return ($result > 0) ? $matches : null;
```
lukasbestle: I think this should be:

```suggestion
		return ($result === 1) ? $matches : null;
```

src/Toolkit/Str.php
```diff
@@ -522,6 +522,49 @@ public static function ltrim(string $string, string $trim = ' '): string
 		return preg_replace('!^(' . preg_quote($trim) . ')+!', '', $string);
 	}
 
+	/**
+	 * Match string against a regular expression and return matches
+	 *
+	 * @param string $string The string to match
+	 * @param string $pattern The regular expression
+	 * @param int $flags Optional flags for the match
```
```suggestion
	 * @param int $flags Optional flags for PHP `preg_match()`
```

src/Toolkit/Str.php
```diff
@@ -522,6 +522,49 @@ public static function ltrim(string $string, string $trim = ' '): string
...
+	 * @param int $offset Optional offset for the match
```
```suggestion
	 * @param int $offset Positional offset in the string to start the search
```

src/Toolkit/Str.php
```diff
@@ -522,6 +522,49 @@ public static function ltrim(string $string, string $trim = ' '): string
...
+		return ($result > 0) ? $matches : null;
+	}
+
+	/**
+	 * Syntax sugar for boolean string matching
+	 *
+	 * @param string $string The string to match
+	 * @param string $pattern The regular expression
+	 * @param int $flags Optional flags for the match
+	 * @param int $offset Optional offset for the match
```
```suggestion
	 * @param int $flags Optional flags for PHP `preg_match()`
	 * @param int $offset Positional offset in the string to start the search
```

src/Toolkit/Str.php
```diff
@@ -522,6 +522,49 @@ public static function ltrim(string $string, string $trim = ' '): string
...
+	 * @return array|null The matches or null if no match was found
+	 */
+	public static function match(string $string, string $pattern, int $flags = 0, int $offset = 0): ?array
+	{
+		$result = preg_match($pattern, $string, $matches, $flags, $offset);
+		return ($result > 0) ? $matches : null;
+	}
+
+	/**
+	 * Syntax sugar for boolean string matching
```
```suggestion
	 * Check whether a string matches a regular expression
```

src/Toolkit/Str.php
```diff
@@ -522,6 +522,49 @@ public static function ltrim(string $string, string $trim = ' '): string
...
+		return static::match($string, $pattern, $flags, $offset) !== null;
+	}
+
+	/**
+	 * Match string against a regular expression and return all matches
+	 *
+	 * @param string $string The string to match
+	 * @param string $pattern The regular expression
+	 * @param int $flags Optional flags for the match
+	 * @param int $offset Optional offset for the match
```
```suggestion
	 * @param int $flags Optional flags for PHP `preg_match_all()`
	 * @param int $offset Positional offset in the string to start the search
```

src/Toolkit/Str.php
```diff
@@ -522,6 +522,49 @@ public static function ltrim(string $string, string $trim = ' '): string
...
+	 * @return array|null The matches or null if no match was found
+	 */
+	public static function matchAll(string $string, string $pattern, int $flags = 0, int $offset = 0): ?array
+	{
+		$result = preg_match_all($pattern, $string, $matches, $flags, $offset);
+		return ($result > 0) ? $matches : null;
```
lukasbestle, line 566:
```suggestion
		return ($result === 1) ? $matches : null;
```

lukasbestle (commented):
Really useful, thanks! Also great unit tests.

A few minor comments that apply to all methods, so the number of comments looks higher than it is.

lukasbestle (approved):

PR #5036, opened 2023-02-02
`A::fill()` accept callable as fill value

adamkiss:
## This PR …
Sometimes, when doing parametric stuff, it's nice to create an array with some default values - which are not the same.

Small change to `A::fill` makes it accept a closure/callable, so it can be used to create new arrays with useful values:

```php
$new = A::fill([], 10, fn($i) => $i); // [0, 1, 2, 3, …]
$new2 = A::fill($new, 15, fn($i) => $i); // correctly adds [10, 11, …]
```

## Ready?

### For review team

### DIFF
--- a/tests/Toolkit/ATest.php
+++ b/tests/Toolkit/ATest.php
@@ -2,7 +2,6 @@
 
 namespace Kirby\Toolkit;
 
-use Kirby\Toolkit\V;
 use PHPUnit\Framework\TestCase;
 
 /**
@@ -457,10 +456,10 @@ class ATest extends TestCase
 			'elephant',
 			'elephant',
 			'elephant'
-		], A::fill($array, 6, fn() => 'elephant'));
+		], A::fill($array, 6, fn () => 'elephant'));
 
 		// Callable with Closure
-		$this->assertSame([1, 2, 3], A::fill([], 3, fn(int $i) => $i + 1));
+		$this->assertSame([1, 2, 3], A::fill([], 3, fn (int $i) => $i + 1));
 
 		// callable with callable
 		$this->assertSame([false, true, false], A::fill([], 3, [V::class, 'accepted']));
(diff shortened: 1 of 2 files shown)


adamkiss:
Current error is something with CI, not in my code AFAICT

distantnative:
I think we can make this a bit cleaner by letting the loop start at the array count and go up to the limit

tests/Toolkit/ATest.php
```diff
@@ -447,6 +447,26 @@ public function testFill()
...
+		$this->assertSame([
+			'miao',
+			'wuff',
+			'tweet',
+			'elephant',
+			'elephant',
+			'elephant'
+		], A::fill($array, 6, function () {
+			return 'elephant';
+		}));
```
distantnative: Let's use arrow functions here, e.g.
```php
A::fill($array, 6, fn () => return 'elephant');
```


But also below.

tests/Toolkit/ATest.php
```diff
@@ -447,6 +447,26 @@ public function testFill()
...
+
+		// Callable with Closure
+		$this->assertSame([1, 2, 3], A::fill([], 3, function (int $i) {
+			return $i + 1;
+		}));
+
+		// callable with callable
+		$this->assertSame([false, true, false], A::fill([], 3, [\Kirby\Toolkit\V::class, 'accepted']));
```
distantnative: Please put rather a `use` statement on top of the file, not the full namespace classname inline.

tests/Toolkit/ATest.php
adamkiss: Yeah, i am a bit misty about what php version introduced some modern features i use in my code, so i didn’t want to be overzealous in far reaching kirby code, even though this is a test. Will update

lukasbestle (approved):

PR #5037, opened 2023-02-02
Add Macroable trait to Str and A classes

adamkiss:
## This PR …
adds a `Macroable` trait to add custom macros to static toolkit classes, currently added to `Str` and `A` toolkits.

You can add methods to any kind of `\Kirby\Cms\*` object and collection, but you can't with toolkit classes. Well now you can!

```php
// Define a macro
Str::_addMacro('bold', fn($str) => "<strong>{$str}</strong>"));

// In a template
echo Str::bold('favoured by fortune');
```

When I talked about this with @lukasbestle privately, he suggested I leave the `_addMacro` off and just let a programmer write to public `Class::$_macros`. I decided against to provide a little bit of sanity check:

- if you add a macro named after an existing method and are surprised with results
- if you try to add a macro and forget you already added it elsewhere (or try to overwrite macro added by a plugin)

I also decided for a really ugly `_underscoreMethods`, to emphasize the meta programming aspect of this, and in case anyone really needs a custom `Str::hasMacro` method.

I am open to suggestions.

### Fixes
That sadness when you really want a `A::something` method and it doesn't exist.

### Breaking changes
None.

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -4,7 +4,6 @@ namespace Kirby\Toolkit;
 
 use Closure;
 use Exception;
-use Kirby\Toolkit\Macroable;
 
 /**
  * The `A` class provides a set of handy methods
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -20,7 +20,7 @@ use Exception;
  */
 class A
 {
-	use Macroable;
+	use HasMethods;
 
 	/**
 	 * Appends the given array
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -4,6 +4,7 @@ namespace Kirby\Toolkit;
 
 use Closure;
 use Exception;
+use Kirby\Toolkit\Macroable;
 
 /**
  * The `A` class provides a set of handy methods
@@ -20,6 +21,8 @@ use Exception;
  */
 class A
 {
+	use Macroable;
+
 	/**
 	 * Appends the given array
 	 */
--- a/src/Toolkit/HasMethods.php
+++ b/src/Toolkit/HasMethods.php
@@ -16,7 +16,7 @@ use Exception;
  * @copyright Adam Kiss
  * @license   https://opensource.org/licenses/MIT
  */
-trait Macroable
+trait HasMethods
 {
 	public static $methods = [];
 
--- a/src/Toolkit/Macroable.php
+++ b/src/Toolkit/Macroable.php
@@ -15,7 +15,8 @@ use Exception;
  * @copyright Adam Kiss
  * @license   https://opensource.org/licenses/MIT
  */
-trait Macroable {
+trait Macroable
+{
 	private static $_macros = [];
 
 	/**
@@ -25,7 +26,8 @@ trait Macroable {
 	 * @param callable $macro
 	 * @return void
 	 */
-	public static function _addMacro(string $name, callable $macro) {
+	public static function _addMacro(string $name, callable $macro)
+	{
 		if (method_exists(static::class, $name)) {
 			throw new Exception('Class "' . static::class . "\" already contains static method \"{$name}\"");
 		}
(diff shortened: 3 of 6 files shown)


distantnative:
My few initial thoughts:
- `macros` is not a good name here, it brings along certain associations that aren't meant here. When we call them `methods` in other classes, why not here?
- No `addMacro` method. Also no weird underscore methods. That's very uncommon for Kirby and such inconsistencies rather lead to a lot of confusion. Since from my view this is a rather rare use case anyways `Str::$methods['something'] = …` should be sufficient. It's also very consistent with page methods etc. in how it would behave if you register a method that already exists.
- Do we really need this? Not sure this isn't more bloat than solving a lot of needs for adding methods to our toolkit classes

Happy to discuss these :)

adamkiss:
@distantnative 

1. This is vernacular used in the wider php world, but i have no problem being more in line with vernacular used in kirby
2. Yes, I might have been overzealous in “keeping it safe”
3. Yes, for the same reason collection* methods are such a nice feature. It cleans the code a lot, and if you’re working with pure data, that is always a win.

\* - Collection, Page, Site, etc.

happy to simplify the code.

lukasbestle:
Thanks for making the changes, I agree with Nico on those.

I think the name of the trait should be `HasMethods` for consistency with `Cms\HasMethods` (which uses instance methods, so it cannot be reused for the toolkit).

adamkiss:
@lukasbestle @distantnative I'll upldate the code shortly

One question: Pure Toolkit Collection is the last one of Toolkit and Derived classes that doesn't have a way to add custom methods, but is too different from Str&Array to use the UseMethods macro. Should I:

1.) Copy over the funcionality from HasMethods that'll work with collectin to Collection PHP
	a.) This PR
	b.) New PR
2.) Create "HasStaticMethods" (this) & "HasMethods" (specifically for Collections)?
3.) Ignore this fact

I don't think it's possible to do a trait that works for both static classess `Str`, `A`, and a proxy class like `Collection`.

distantnative:
Tbh I'm still a bit hesitant as this is adding more to the code base being a lot of patchwork with Cms\HasMethods, Toolkit\HasMethods, maybe something new for collections too... maybe we need to take a step back and see how to get this more consistent.

distantnative:
I take my last comment back.
I think ti should be one `Toolkit\HasMethods` class and maybe that one needs to props, `staticMethods` and `methods`. Then it could be used for all 3 classes.

lukasbestle:
That could indeed be a way. 👍

One question we will need to answer though is what would happen with the methods in the trait. Those are currently all instance methods and we would need the same ones as static methods. Another instance (pun intended) where I'm sad that PHP does not support overloading. Would be so great to be able to define the same methods both as instance and static methods and handle the calls differently depending on that.

adamkiss:
re @lukasbestle - I'm confused, as netiher A nor Str currently use any traits, so I'm not sure what methods in what trait are you talking about.

re @distantnative - I'm personally unsure. I think I'd prefer two different traits - `HasInstanceMethods`, `HasStaticMethods`, both to set `$methods`. I think consistency in assigning custom methods to all classes via `Class::$methods` is more important than saving one file / having less traits.

lukasbestle:
I was referring to the `HasMethods` trait.

lukasbestle:
> I think I'd prefer two different traits - `HasInstanceMethods`, `HasStaticMethods`, both to set `$methods`

This would mean that the two traits wouldn't be useable together. Could be fine, but maybe we'll have a case in the future where we need both together. So the question is whether it makes sense to plan ahead for that potential future case.

distantnative:
I don't quite see a healthy way forward right now to be honest. I would suggest to firs try the concept out in a plugin, extending the `Toolkit\A` and `Toolkit\Str` classes. And once the concept has matured and proven itself, it might still be an option to include it in the core. But right now it doesn't seem at a point where adding this would be future safe.

PR #5038, opened 2023-02-03
Fluency for Kirby Toolkit

adamkiss:
## This PR …
Adds a fluent usage to `Str` and `A` classes via two new classes `FluentString` and `FluentArray`, both based on `Fluent`, which is the basis of this addition. Example:

```php
$x = Fluent::string('1, 2, 3')
	->split(',')
	->map(fn($s) => trim($s))
    ->join(','); // returns '1,2,3'
```

Not the most exciting example, but shows what's possible in more complicated scenarios, and even show `splitMapJoin` has a potential to clean up some workflows.

There is, however a number of questions, so I wanted to open a discussion first.

1. Originally, when I proposed this idea to Lukas, the idea was to add a trait to both `A` and `Str`, and use that trait to `__call` the static methods and return re-wrapped fluent value. Unfortunately, `$instance->staticMethod` still calls the static method, not `__call('staticMethod')`, and optional throws a `E_STRICT` to go along with it.
2. The main thing happens in `Fluent` (wrapping the value, calling correct static class and returning rewrapped value), but I also extended it to two separate classes for simplified IDE usage: rewriting static methods via the `@method` phpdoc makes the IDE aware of just what is available at the object. If there's a more elegant solution, I couldn't think of it.
3. there's no connect with `Collections`, just yet. I'd make that possible via `A::toCollection()`, and `$collection->toFluentArray()` to further allow chaining (for instance `Db::select()->toFluentArray()->reduce()`, or other possible uses)
4. I'd also propose three new helper facades - `a()`, `str()`, and `collect()` to create a fluid representations of these objects.

Once these points are talked through, I'll either finish the `@method` documentation, additional methods and helpers, or whatever else we agree on.

### Breaking changes
None yet, AFAIK.

## Ready?

- [ ] Discussion closed

### For review team

### DIFF
--- a/src/Toolkit/Fluent.php
+++ b/src/Toolkit/Fluent.php
@@ -1,0 +1,79 @@
+<?php
+
+namespace Kirby\Toolkit;
+
+use Error;
+
+/**
+ * Base for Fluent classes, actually containing the logic
+ * of the fluency
+ *
+ * @package   Kirby Toolkit
+ * @author    Adam Kiss <iam@adamkiss.com>
+ * @link      https://getkirby.com
+ * @copyright Adam Kiss
+ * @license   https://opensource.org/licenses/MIT
+ */
+class Fluent
+{
+	protected $value = null;
+
+	public function __construct($value = null){
+		$this->value = $value;
+	}
+
+	/**
+	 * Static wrapper for the FluentString constructor
+	 *
+	 * @param string $value
+	 * @return FluentString
+	 */
+	public static function string(string $value = ''): FluentString
+	{
+		return new FluentString($value);
+	}
+
+	/**
+	 * Static wrapper for the FluentArray constructor
+	 *
+	 * @param array $value
+	 * @return FluentArray
+	 */
+	public static function array(array $value = []): FluentArray
+	{
+		return new FluentArray($value);
+	}
+
+	/**
+	 * Proxy for all instance calls
+	 *
+	 * @param string $name
+	 * @param array $arguments
+	 * @return mixed
+	 */
+	public function __call(string $name, array $arguments): mixed
+	{
+		$return = match (true) {
+			is_array($this->value)
+			|| $this::class === FluentArray::class => A::{$name}($this->value, ...$arguments),
+			is_string($this->value)
+			|| $this::class === FluentString::class => Str::{$name}($this->value, ...$arguments),
+			default => throw new Error('Fluent can only be used with arrays or strings'),
+		};
+
+		return match (true) {
+			is_array($return) => new FluentArray($return),
+			is_string($return) => new FluentString($return),
+			default => $return,
+		};
+	}
+
+	/**
+	 * Gives access to self and the returns self without change
+	 */
+	public function tap(callable $callback): self
+	{
+		$callback($this);
+		return $this;
+	}
+}
--- a/src/Toolkit/FluentArray.php
+++ b/src/Toolkit/FluentArray.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Kirby\Toolkit;
+
+/**
+ * FluentString - Fluent extension for proper intelisense in IDEs
+ *
+ * @package   Kirby Toolkit
+ * @author    Adam Kiss <iam@adamkiss.com>
+ * @link      https://getkirby.com
+ * @copyright Adam Kiss
+ * @license   https://opensource.org/licenses/MIT
+ *
+ * TODO: Add @method calls for IDE understanding of the code
+ */
+class FluentArray extends Fluent
+{
+	public function value(): array
+	{
+		return $this->value;
+	}
+}
--- a/src/Toolkit/FluentString.php
+++ b/src/Toolkit/FluentString.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Kirby\Toolkit;
+
+/**
+ * FluentString - Fluent extension for proper intelisense in IDEs
+ *
+ * @package   Kirby Toolkit
+ * @author    Adam Kiss <iam@adamkiss.com>
+ * @link      https://getkirby.com
+ * @copyright Adam Kiss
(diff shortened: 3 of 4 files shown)


distantnative:
I'm having problems with the whole `fluent` term in the context of Kirby, maybe it's just me, but it feels very foreign. Could it be just `String`, `Array` or `StringObj`, `ArrayObj`? That feels more within the Kirby terminology. Then it could also be `::toObj` which would feel very natural.

adamkiss:
Again, Fluency is a vernacular, AFAIK, so I started with that. Out of the options you've presented, I like probably `*Obj` options more, because I personally always find `array` vs `Array` distinction (one runtime type, one library object) iffy in other languages/libraries, and the possible confusion (resulting in "I'm tired, so I debugged this one Title cased word for two hours" :D ) a bigger problem than "this doesn't feel like Kirby".

There's already a bit of confusion going around `Db` & `Database` in my experience.

`*Obj` sidestep that quite nicely IMO. I'll update the code.

Edit: One thing, @distantnative ; How would we name the `Fluent` class? `Obj` is obviously something else, so maybe `TypeObj`? `SelfObj`?

lukasbestle:
@distantnative How would the `::toObj()` method be used? Won't it be confused with the existing `Obj` class that does completely different things?

distantnative:
Could be, I was thinking out loud to start discussing better ideas.

I just think we're entering not great territory when introducing methods like `->toFluentArray()`. To me that sounds a lot like naming from all kinds of libraries that I dread. I think the benefit of Kirby always has been great PHP api naming, one that feels lightweight and intuitive. With this PR so far I feel like it's all about achieving one thing (chainable toolkit methods) but without finding a way that feels good, intuitive, like home for Kirby. That's why I wanted to open up the discussion - don't think we could merge this with the current naming.

(Not bashing on the work, Adam. It's a good PR and I understand the wish for the feature. But sometimes we have to take a few rounds especially with naming etc to get it to a great quality for Kirby)

lukasbestle:
I fully agree that I'm not 100% happy with the naming yet either. But TBH I find `$collection->toObj()` even more confusing than `$collection->toFluentArray()` as it doesn't tell that one can call array methods on the return value. But we could use `->toArrayObj()` IMO.

Another thing I'm not sure about is the `@method` type hints inside `FluentArray` and `FluentString`. Unfortunately it doesn't seem to be possible to add an annotation like "any static method of class X can be called non-statically here". So we would need to keep these lists updated continuously. As a manual process, this will probably lead to outdated in-code docs at some point. But getting rid of the annotations entirely is also not great as IDEs will have no idea which methods these objects have.

lukasbestle:
@adamkiss A possible name for the `Fluent` class could be `Proxy`.

adamkiss:
@lukasbestle @distantnative 

Naming - out of your suggestions I like:
```
- ProxyObj
	- StringObj
	- ArrayObj
```

As for @method type hints, I agree that manually keeping two files in line is less then ideal, but it's the best (the only)? way to have the type hints in IDEs, and I think the value outweighs the price. If need be, as long as I use Kirby, I'll maintain the type hints with new Kirby versions, unless I'm unavailable at the time. The Proxy Object is just too… useful to me to pass on it, in many cases.

Edit: I'll hold off any further changes, until we agree on naming, but I'd really like this to get in :)

distantnative:
It still feels very foreign and odd, like an alien inside Kirby. Maybe @bastianallgeier has a good idea if/how this could fit inside Kirby.

adamkiss:
@distantnative Does it? I'll of course wait for Bastian to weigh in, but in my opinion, it "feels" like a part of Kirby - if you think upon Collections (which are a type of this, both root toolkit object and the derived ones - pages, users, site even), Fields at cetera, chainging methods over proxied values is one of the most valuable and intuitive parts of the Kirby.

lukasbestle:
On `@method` type hints: We discussed this recently in https://github.com/getkirby/kirby/issues/5039. I think they would rather be something for a separate tool or stub library.

On the general feature: I agree with Adam. Fluency is what makes Kirby really great. So far we don't have it for Toolkit classes, but maybe just because we didn't realize we could use the same pattern here as well. My current feeling is that the feature is great, it just needs to be reduced in complexity implementation-wise.

distantnative:
I didn't mean fluency isn't Kirby-like. But the implementation, the very technical names, the connection between those classes... it makes it feel like something stuck onto Kirby with duck tape and not like a natural part of it. Which is really hard to do. I don't want to put down the work at all. But we see it in so many of our own PRs where it often really takes time and many new approaches until finding something that feels right. To me, this doesn't yet. But if the rest of the team all think it does, it'll be.

lukasbestle:
I fully agree with you on that. So to summarize: The feature is great, the implementation is not quite there yet.

bastianallgeier:
Sorry for being late to the party. I love the general concept, but I have to agree with the others so far. But to be honest, our biggest task right now is to clean up and throw out things where we went too far. That's why I really wonder if this should be in the core at all. This is not meant to diminish the usefulness of it at all. But I'm very hesitant to add code that we would not immediately need ourselves at the moment. Do you think it could be a plugin instead?

distantnative:
I thought this morning about an idea - still maybe more suited as plugin: 

`Kirby\Fluent` namespace with `Kirby\Fluent\Str`, `Kirby\Fluent\A` classes plus maybe a helper function `fluent()`, so that you could make it like


```php
fluent('hello')->upper()->after('ll')
```

I think that could also serve well enough as connector e.g. to Collections:

```php
fluent(Db::select())->reduce()
```

bastianallgeier:
@distantnative I like that.

distantnative:
If there is still interest, I would suggest to create it as a plugin in the way I described (Kirby\Fluent and fluent helper). That would help us to mature functionality and see interest in the community. Based on that it’ll be easier to decide if/when to include this in the core.

PR #5043, opened 2023-02-05
Allow individual isDraft in props to work

### LINKED ISSUE #5041: Pages::factory() ignores individual isDraft statuses
## Description

The [documentation](https://getkirby.com/docs/reference/objects/cms/pages/factory) for `Pages::factory()` indicates that the props array for each page can use `'isDraft'` to set the draft flag for each page individually. However, [Pages.php:178](https://github.com/getkirby/kirby/blob/main/src/Cms/Pages.php#L178) unconditionally sets the 'isDraft' property for each page array to the value of the `$draft` parameter, which defaults to `false`.

This means that all pages created with a single call to the `Pages::factory()` method will always have the same draft status - `false` by default.

**Expected behavior**  
Calling `Pages::factory()` like this:

```
$pages = Pages::factory([
	[
		'template' => 'default',
		'slug' => 'page-1',
		'content' => ['title' => 'Page 1'],
		'isDraft' => false,
	], [
		'template' => 'default',
		'slug' => 'page-2',
		'content' => ['title' => 'Page 2'],
		'isDraft' => true,
	],
]);
```

Should return a collection with two pages, one being a draft and the other not. Instead both pages are not drafts.

## To reproduce

```
<?php

$pages = Pages::factory([
	[
		'template' => 'default',
		'slug' => 'page-1',
		'content' => ['title' => 'Page 1'],
		'isDraft' => false,
	], [
		'template' => 'default',
		'slug' => 'page-2',
		'content' => ['title' => 'Page 2'],
		'isDraft' => true,
	],
]);

var_dump($pages->nth(0)->isDraft());
var_dump($pages->nth(1)->isDraft());

?>
```

## Your setup
**Kirby Version**
3.8.1.1 & 3.9.1

SeriousKen:
Fix to allow individual isDraft to override the factory default if set in the array of props (issue https://github.com/getkirby/kirby/issues/5041)

### Fixes
Issue https://github.com/getkirby/kirby/issues/5041

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -175,7 +175,7 @@ class Pages extends Collection
 			$props['kirby']   = $kirby;
 			$props['parent']  = $parent;
 			$props['site']    = $site;
-			$props['isDraft'] = $draft ?? $props['isDraft'] ?? false;
+			$props['isDraft'] = $draft ?? $props['isDraft'] ?? $props['draft'] ?? false;
 
 			$page = Page::factory($props);
 
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -175,7 +175,7 @@ class Pages extends Collection
 			$props['kirby']   = $kirby;
 			$props['parent']  = $parent;
 			$props['site']    = $site;
-			$props['isDraft'] = $draft;
+			$props['isDraft'] ??= $draft;
 
 			$page = Page::factory($props);
 
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -154,7 +154,7 @@ class Pages extends Collection
 	 *
 	 * @param array $pages
 	 * @param \Kirby\Cms\Model|null $model
-	 * @param bool $draft
+	 * @param bool|null $draft
 	 * @return static
 	 */
 	public static function factory(array $pages, Model $model = null, bool $draft = null)
@@ -175,7 +175,7 @@ class Pages extends Collection
 			$props['kirby']   = $kirby;
 			$props['parent']  = $parent;
 			$props['site']    = $site;
-			$props['isDraft'] = $draft ?? ($props['isDraft'] ?? false);
+			$props['isDraft'] = $draft ?? $props['isDraft'] ?? false;
 
 			$page = Page::factory($props);
 
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -157,7 +157,7 @@ class Pages extends Collection
 	 * @param bool $draft
 	 * @return static
 	 */
-	public static function factory(array $pages, Model $model = null, bool $draft = false)
+	public static function factory(array $pages, Model $model = null, bool $draft = null)
 	{
 		$model  ??= App::instance()->site();
 		$children = new static([], $model);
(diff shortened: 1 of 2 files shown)


distantnative:
@SeriousKen Thanks for the PR!
Though I think this fix is not fully correct. I think right now the issue is that if the `$draft` parameter is set to `false` as default it still overwrites any custom `isDraft` - that's a bug indeed. 

However, when somebody is explicitly calling the method with a `$draft` parameter, I think this should overwrite it, which your PR would prevent.

So I think we should change the parameter default to `null` and do nothing when it's `null` but ensure the overwrite when an actual value is passed.

SeriousKen:
I've updated the PR to reflect the changes in priority of the `$draft` parameter vs individual `isDraft` prop.

distantnative:
@SeriousKen same here, if you could add unit tests that would be splendid. if that is out of your comfort zone, let us know :)

SeriousKen:
I've added a test case that should cover this PR.

src/Cms/Pages.php
```diff
@@ -175,7 +175,7 @@ public static function factory(array $pages, Model $model = null, bool $draft =
 			$props['kirby']   = $kirby;
 			$props['parent']  = $parent;
 			$props['site']    = $site;
-			$props['isDraft'] = $draft;
+			$props['isDraft'] = $draft ?? $props['isDraft'] ?? false;
```
afbora: @distantnative @lukasbestle We allow `draft` argument for `Page::create`. Do you think we should implement it here as well for consistency?

https://github.com/getkirby/kirby/pull/5044/files#diff-8e8501f5cf24c63dea196abff40bfb44322604ba3a244561300db2c053d2242aR526

````php
$props['isDraft'] = $draft ?? $props['isDraft'] ?? $props['draft'] ?? false;
````

Or will we remove one of both (`isDraft` or `draft`) in future versions?

**Review by afbora — DISMISSED** [review]

File: src/Cms/Pages.php (hunk 1 above)
Comment (lukasbestle) [reply src/Cms/Pages.php]: Good question. I was confused by this before and TBH not fully understood why we had this inconsistency (`draft` vs. `isDraft`) in the first place.

My first response would be to deprecate `draft` as the actual prop is called `$isDraft`. But as I said, I don't know the history of this.

Comment (distantnative) [reply src/Cms/Pages.php]: Me neither tbh. Probably just historically grown into this.

Comment (afbora) [reply src/Cms/Pages.php]: Then let's decide on one and add the information to deprecate the other as a comment to the codes.

Comment (lukasbestle) [reply src/Cms/Pages.php]: I think before we can deprecate any prop, we need to understand why we had both. So I'm leaning towards adding support for `draft` here for the time being to ensure consistency. We can still deprecate one of them later.

Comment (distantnative) [reply src/Cms/Pages.php]: 👍

Comment (afbora) [reply src/Cms/Pages.php]: Then it means that we can change this line as below and merge this and #5044 PR, right?

```suggestion
			$props['isDraft'] = $draft ?? $props['isDraft'] ?? $props['draft'] ?? false;
```

distantnative (approved):

PR #4482, opened 2022-07-06
Changed "Login" to "Log in"

purplespider:
“Login” is a noun, not a verb, so the button text on the login form (which is where this text is used) should really be “Log in”. 😊 http://loginisnotaverb.com

### Fixes
https://kirby.nolt.io/444

### Breaking changes
None I can see.

## Ready?

- [x] ~Unit tests for fixed bug/feature~
- [x] ~In-code documentation (wherever needed)~
- [x] Tests and checks all pass

### For review team

- [ ] Add changes to release notes draft in Notion
- [x] ~Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)~

PR #5047, opened 2023-02-08
Correct "Logout" menu label to "Log out"

purplespider:
Corrected the menu label from "Logout" (noun) to "Log out" (verb).

Screenshot 2023-02-08 at 10 45 55@2x

(I should have picked this up in #4482 !)

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -388,7 +388,7 @@
 	"login.toggleText.password-reset.email": "Forgot your password?",
 	"login.toggleText.password-reset.email-password": "← Back to login",
 
-	"logout": "Logout",
+	"logout": "Log out",
 
 	"menu": "Menu",
 	"meridiem": "AM/PM",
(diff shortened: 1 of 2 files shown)


afbora (approved):

lukasbestle (approved):
Thanks for the fix!

@afbora When merging it we need to be careful to update the translations in Transifex like last time.

afbora:
@lukasbestle Yes, I or Bastian can take care it.

afbora:
@bastianallgeier Did you push the change to transifex? (in order not to overlook)

bastianallgeier:
I just did. It looks like Transifex was smart enough to use the old version. At least for German. Can you check if it also worked for Turkish? I think we need to double-check again before the release.

afbora:
Yes, auto translated for Turkish language 👍 I'll check all languages now.

afbora:
✅

PR #5048, opened 2023-02-08
Whoops: Add support for `whoops_blacklist` config option

HYR:
## This PR …
Adds support for `whoops_blacklist` config option to mask variables that are displayed when showing errors with Whoops.

```php
return [
    // mask everything
    'whoops_blacklist' => [
        '_COOKIE' => array_keys($_COOKIE),
        '_SERVER' => array_keys($_SERVER),
        '_ENV' => array_keys($_ENV),  
    ]
];
```

```php
return [
    // mask specific things
    'whoops_blacklist' => [
        '_SERVER' => [
            'AWS_ACCESS_KEY_ID',
            'AWS_SECRET_ACCESS_KEY',  
        ],
    ]
];
```

See this Stack Overflow thread for more details:
https://stackoverflow.com/questions/46407009/how-to-hide-env-passwords-in-laravel-whoops-output/47665048

Used `whoops_blacklist` config option  instead of `debug_blacklist` to be more specific about how the blacklist applies. And matches nicely with already existing `'whoops' => false` config option.

## Ready?

### For review team


distantnative:
I think it should be `whoops.blacklist` then to be consistent in our option naming. But then we also need to make sure that this doesn't clash with our existing checks for the `whoops` option itself.

distantnative:
@HYR will you adapt the PR to use dot notation? In that case, here https://github.com/getkirby/kirby/blob/main/src/Cms/AppErrors.php#L76 we should probably rather do
```php
if ($this->option('whoops', true) !== false) {
```

And wondering if we could add a unit test, but that might be a bit tricky.

distantnative:
@HYR would you adapt this PR as suggested or do you think by now it's not worth pursuing?

HYR:
@distantnative Sorry for letting this go stale. What's the best way to resurrect this?

HYR:
New PR with requested change: https://github.com/getkirby/kirby/pull/5403

PR #5053, opened 2023-02-12
Provide `$field` prop to block/layout objects

lukasbestle:
### Features

- Block and layout objects for the blocks and layout fields now allow access to the parent field object with `$block->field()`, `$layout->field()` and `$layoutColumn->field()`.

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Item.php
+++ b/src/Cms/Item.php
@@ -27,6 +27,11 @@ class Item
 	public const ITEMS_CLASS = Items::class;
 
 	/**
+	 * @var \Kirby\Cms\Field|null
+	 */
+	protected $field;
+
+	/**
 	 * @var string
 	 */
 	protected $id;
@@ -57,6 +62,7 @@ class Item
 
 		$this->id       = $params['id']       ?? Str::uuid();
 		$this->params   = $params;
+		$this->field    = $params['field']    ?? null;
 		$this->parent   = $params['parent']   ?? App::instance()->site();
 		$this->siblings = $params['siblings'] ?? new $siblingsClass();
 	}
@@ -70,6 +76,14 @@ class Item
 	public static function factory(array $params)
 	{
 		return new static($params);
+	}
+
+	/**
+	 * Returns the parent field if known
+	 */
+	public function field(): Field|null
+	{
+		return $this->field;
 	}
 
 	/**
--- a/src/Cms/Item.php
+++ b/src/Cms/Item.php
@@ -26,10 +26,7 @@ class Item
 
 	public const ITEMS_CLASS = Items::class;
 
-	/**
-	 * @var \Kirby\Cms\Field|null
-	 */
-	protected $field;
+	protected Field|null $field;
 
 	/**
 	 * @var string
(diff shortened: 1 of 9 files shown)


src/Cms/Item.php
```diff
@@ -26,6 +26,11 @@ class Item
 
 	public const ITEMS_CLASS = Items::class;
 
+	/**
+	 * @var \Kirby\Cms\Field|null
+	 */
+	protected $field;
```
distantnative: Could we start typing new properties? I know we're running into problems with our existing ones - as our docs and code base aren't as consistent as they would need to be. But maybe we could start with new ones?

distantnative (commented):
Looks good to me, one small comment

distantnative (dismissed):

afbora:
@lukasbestle @distantnative I think we need also implement the `field` prop to following lines?

- https://github.com/getkirby/kirby/blob/feature/items-field-prop/src/Cms/Layouts.php#L101
- https://github.com/getkirby/kirby/blob/feature/items-field-prop/src/Form/Field/LayoutField.php#L32
- https://github.com/getkirby/kirby/blob/feature/items-field-prop/src/Form/Field/LayoutField.php#L149

lukasbestle:
@afbora Those are part of the reason why the `$field` prop is optional (the other reason being that it avoids the breaking change). I don't think we can get to the field object easily from these places.

**Edit:** Forget what I wrote, I think it should actually be doable. Thanks for spotting! Do you have time to update the code? Otherwise let's move this to 3.9.3.

afbora:
> Do you have time to update the code?

Actually this is easy for first one but other two ones are a little complicated since that ones inherit `Form\FieldClass` not `Cms\Field`. 

Possible solutions in my mind:
1. Support `FieldClass` class: `protected Field|FieldClass|null $field;`
2. Add new `field` prop/method that returns new `Field` object to `FieldClass` class.

Which way do we should follow?

lukasbestle:
The goal of the new field prop is to make UUIDs for blocks possible in a future version. So I think we can't make it a form field, it needs to be a content field.

afbora (dismissed):
I've added only for first one (content field) 👍

afbora:
I've created a PR about failing CI workflow https://github.com/getkirby/v4/pull/24

afbora:
I've created a PR that fixes CI workflow issue https://github.com/getkirby/kirby/pull/5055

afbora (approved):

lukasbestle:
@afbora Thanks for updating the `Layouts` code and force-pushing the branch again.

I checked the other two uses in `LayoutField` again: I don't think we'll need to update those anyway as these calls won't be relevant for block UUIDs. Or if they are, we could still update them when we get to it. Maybe we'll have refactored the field setup anyway by then.

lukasbestle:
So looks good to me. Maybe @distantnative can have a final look before we merge this.

distantnative (approved):

PR #5063, opened 2023-02-21
Fix `->inline()`, `->nl2br()` and `->toTimestamp()` field methods when field is empty or not exists

### LINKED ISSUE #5061: `inline` field method throws when empty field value is passed
## Description
strip_tags(): Passing null to parameter #1 ($string) of type string is deprecated

**Expected behavior**  
Fall back to empty string when field value is null

## To reproduce
```php
<?= $page->nonExistingFieldOrEmpty()->inline();
```

## Your setup
**Kirby Version**  
3.9.1

afbora:
### Fixes
- Fixes `->inline()`, `->nl2br()` and `->toTimestamp()` field methods when field is empty or not exists

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/methods.php
+++ b/config/methods.php
@@ -278,10 +278,10 @@ return function (App $app) {
 		 * Converts the field value to a Unix timestamp
 		 *
 		 * @param \Kirby\Cms\Field $field
-		 * @return int
+		 * @return int|false
 		 */
-		'toTimestamp' => function (Field $field): int {
-			return strtotime($field->value);
+		'toTimestamp' => function (Field $field): int|false {
+			return strtotime($field->value ?? '');
 		},
 
 		/**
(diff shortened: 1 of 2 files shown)


afbora:
Actually there is same issue for the `->toTimestamp()` as well but I'm not sure about that. Do you think we should also implement for the `->toTimestamp()`? @distantnative

distantnative (dismissed):

lukasbestle:
@afbora You are right, `toTimestamp` has the same issue. We can solve it the same way by falling back to an empty string. But we need to update the return type hint to `int|false` (not a breaking change as that could already have happened for malformed field values).

lukasbestle (approved):

PR #5067, opened 2023-02-25
Fix type/id in fieldsets

plain-solutions-gmbh:
## This PR …

I think, you're mixing type and id of the fieldsets in some way. In this branch, i detached them. Now you will be able to use a block multiple times with other settings like this:

```
  headings:
    label: Heading types
    type: blocks
    fieldsets:
      big_headlines: h1
      small_headliness: h2h3h4 
```

/blueprints/blocks/h1.yml
```
title: Big Headlines
icon: headline
type: heading
fields:
  level:
    type: hidden
    default: "h1"
```

/blueprints/blocks/h2h3h4.yml
```
title: Small Headlines
icon: title
type: heading
fields:
  level:
    label: field.blocks.heading.level
    type: select
    empty: false
    default: "h2"
    options:
      - h2
      - h3
      - h4
```

In this PR i adapted the choose method of the blocks components. But still there is some more to change.
And also i didn't test just shortly.

## Ready?

### For review team


distantnative:
@youngcut not sure what you are trying to solve here by splitting up what is the same. In your example, now you have to list id and type, but the result would be just the same as this, wouldn't it?

```yaml
fields:
  text:
    type: blocks
    fieldsets:
      - h1
      - h2h3h4
```

plain-solutions-gmbh:
I think no. It gets equal by this operation:

`$this->type = $params['id'] = $params['type'];` [View line](https://github.com/getkirby/kirby/blob/617a07aa0151aa4b1a3735b3b5e1a41ccc244a43/src/Cms/Fieldset.php#L49)

**id:** is used to get the fieldset (e.g. in a api request: `http://my-site.com/api/pages/home/fields/my_field/fieldsets/heading_big) and should be **unique**
**type** defines the kind of component that should be loadet.

The way it is now, you can only use one kind of block type per block field. And inhered Blocks with the same type will be overwritten. If i could use the same block type more than once, i could predefined fields in different ways. Like in this example, you could show the title with a hint, that is only one h1 allowed per page. But this is only one example.

texnixe:
Maybe I'm missing something, but I don't understand the example, been using

```
fields:
  text:
    type: blocks
    fieldsets:
      - h1
      - h2
      - h3
```
all the time without issues.

afbora:
Same🧐🤔

plain-solutions-gmbh:
Maybe, it's more a feature than a bug (I always wanted to write this sentence 😝).

I like to use one block (component) multiple times. Here's the schema of my intension:
columns

The section has this fieldset:
```
name: Section
icon: col
wysiwyg: true 
fields:
  columns:
    type: columns
    fieldsets:
      - col_1_1
      - col_1_2
      - col_1_3
```
The component type **column** used multiple time. With a different fieldset.
This way, you can have some preset values for the editors.

If you do this. You will see only col_1_3 in the selection. Cause the type (instead if the id) is a key in a array. -> swoosh.

Maybe i'm the only one with this need. 🤔 I can understand, that you prefer to keep your hands off it rather than let it get you into trouble somewhere else.

plain-solutions-gmbh:
But I thought that this could be done with a few lines in the Fieldsets class. But this stuff drags on until the component. And to create a suitable solution that is adapted to all use cases afterwards will be difficult.

texnixe:
@youngcut What's the purpose of the `type: column` setting in the col_x_x fieldsets? If you remove that, your blocks should work as expected without any changes to the source code.

plain-solutions-gmbh:
if i remove `type: column`, then Kirby doesn't know which block for the col_x_x should be used.

distantnative:
@youngcut when you say 

> Kirby doesn't know which block for the col_x_x should be used

do you mean the HTML/PHP snippet? Basically that you would like to have different blocks to choose from in the Panel (with different blueprint setup, e.g. fields) but they should all use the same snippets - frontend as well as Vue in the Panel?

plain-solutions-gmbh:
Hi Nico. It's more about the Vue Component. But also the snippet `snippet/block/column.php`. The id should be more for preset and saving stuff. It's pretty complex to explain. (Im not good in it 🥺)

But anyway. If nobody else is missing this thing... I've find another solution for my case. See the [Section Block](https://github.com/youngcut/kirby-section-block). I solve this with dynamic blueprints that i slipped in between. The workaround works perfect. And it gives the required possibilities.
(But out of this context, the plugin stills has a bug inside: 'lost focus on input in column/content blocks' - If someone take care of it... - cause i think a lot of people can take a use for that plugin.)

PR #5069, opened 2023-02-25
`Dir::index()`: fix `$ignore` parameter

### LINKED ISSUE #5010: `Dir::index` - `ignore` argument doesn't do anything
## Description
One of the parameters of `Dir::index` doesn't do anything other than being pass along to a recursive, call where it doesn't do anything again.

**Expected behavior**  
Makes `Dir::index` ignore file, either by relative name (`file.txt`), or absolute name.

## To reproduce
Just look at the source: https://github.com/getkirby/kirby/blob/066b0506a81fdd331edc54add601d907a61ebbb4/src/Filesystem/Dir.php#L143-L165

`$ignore` is used only once, and that's when the recursion is called.

distantnative:
### Fixes
- `Dir::index()`: `$ignore` parameter works now 
 #5010

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/Dir.php
+++ b/src/Filesystem/Dir.php
@@ -143,7 +143,7 @@ class Dir
 	public static function index(
 		string $dir,
 		bool $recursive = false,
-		array|null $ignore = null,
+		array|null $ignore = [],
 		string $path = null
 	): array {
 		$result = [];
@@ -151,6 +151,10 @@ class Dir
 		$items  = static::read($dir);
 
 		foreach ($items as $item) {
+			if (in_array($item, $ignore) === true) {
+				continue;
+			}
+
 			$root     = $dir . '/' . $item;
 			$entry    = $path !== null ? $path . '/' . $item : $item;
 			$result[] = $entry;
--- a/tests/Filesystem/DirTest.php
+++ b/tests/Filesystem/DirTest.php
@@ -169,10 +169,10 @@ class DirTest extends TestCase
 	public function testIndex()
 	{
 		Dir::make($dir = $this->tmp);
-		Dir::make($sub = $this->tmp . '/sub');
+		Dir::make($this->tmp . '/sub');
 
-		F::write($a = $this->tmp . '/a.txt', 'test');
-		F::write($b = $this->tmp . '/b.txt', 'test');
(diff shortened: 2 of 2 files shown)


src/Filesystem/Dir.php
```diff
@@ -143,14 +143,18 @@ public static function files(
...
-		array|null $ignore = null,
+		array|null $ignore = [],
 		string $path = null
 	): array {
 		$result = [];
 		$dir    = realpath($dir);
 		$items  = static::read($dir);
 
 		foreach ($items as $item) {
+			if (in_array($item, $ignore) === true) {
```
lukasbestle, line 154: Needs to check if `$ignore` is an array first (could be `null`).

lukasbestle (commented):
I think the ignore implementation should be consistent with the one of the `Dir::copy()` method as both have a recursive mode. `Dir::copy()` uses the full path to compare against the ignore list, not just the path segment.

Also it should support the same `$ignore = false` mode like `Dir::copy()` does (which disables the global ignore list).

distantnative:
@lukasbestle but wouldn't it be then inconsistent to `Dir::read()`? Which I thought is much closer related to the index method.

src/Filesystem/Dir.php
distantnative: Probably better not allow passing `null` - or what use case would passing `null` have?

lukasbestle:
`Dir::read()` is non-recursive though. I see what you mean and I feel like there is no right or wrong here.

src/Filesystem/Dir.php
lukasbestle: Not adding a breaking change. It could be that some users explicitly copied the default values into their method calls (for whatever reason). This will then break.

I'm in favor of removing the `null` value (or replacing it with `array|bool` like with `Dir::copy()` like I wrote above). But then it should happen in a major release.

bastianallgeier:
I think I would keep null for 3.9.2 and remove it in the next major version.

src/Filesystem/Dir.php
bastianallgeier: Let's change the type definition in the next major.

bastianallgeier (approved):

Issue #5007, opened 2023-01-17
Panel: Language selector out of sync when using browser back and forward


chaeringer:
(First of all thanks a lot for your amazing work with Kirby over the years! 👏)

## Description

**Expected behavior**  

Language selector and `language` URL parameter should always be in sync and content changes should only affect the current active language. 

**Screenshots**  
![Screenshot 2023-01-17 at 19 59 47](https://user-images.githubusercontent.com/1009778/212989329-54a3ef1b-e772-4066-b213-5a2e7eddd5f9.png)

https://user-images.githubusercontent.com/1009778/212992306-d4ddc60e-0541-474a-835c-6efb8db1279a.mov

## To reproduce

0. Kirby page with multiple languages
1. Go to any page and change the language with the language selector
2. Use the browser back or forward button
3. Language selector and `language` URL parameter are out of sync
4. Change content in one language. Save the changes. Content in other language is overwritten 
5. In my example above english content is overwritten with german content.  This issue "works" in both directions.
6. 🚨 Content is lost

https://github.com/chaeringer/kirby-starterkit-testcase (newest starterkit with multiple languages)

## Your setup
**Kirby Version**  
Kirby 3.9.0 
I was able to reproduce the issue also with 3.8.1.1 

**Your system**
- Chrome Version 108.0.5359.124
and
- Firefox 109.0
- macOS BigSur Version 11.6.2

afbora:
@distantnative Could we solve with listening `onpopstate`?

````js
window.onpopstate = function () {
  // reload page with fiber or native
};
````

stananiev:
Yep, I have the same issue!

distantnative:
@afbora definitely is on the right track that Fiber hasn't handled popping the state properly. I've a first PR that could solve the issue, have to test for side effects etc.

bastianallgeier:
I think I have a fix (see PR above)

PR #5070, opened 2023-02-25
Fiber: handle popping state

distantnative:
### Fixes
- Fiber handles popping the window state now, fixes issues with e.g. language selector on back navigation
#5007

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/fiber/app.js
+++ b/panel/src/fiber/app.js
@@ -114,9 +114,6 @@ export default {
 				};
 			}
 		});
-
-		// back button event
-		window.addEventListener("popstate", this.$reload);
 	},
 	methods: {
 		/**
--- a/panel/src/fiber/index.js
+++ b/panel/src/fiber/index.js
@@ -66,6 +66,9 @@ export default class Fiber {
 
 		// set initial state
 		this.setState(state);
+
+		// set up event handlers
+		window.addEventListener("popstate", this.popState.bind(this));
 	}
 
 	/**
@@ -294,6 +297,12 @@ export default class Fiber {
 			return data;
 		} finally {
 			this.options.onFinish(options);
+		}
+	}
+
+	async popState(event) {
+		if (event.state?.$url) {
+			window.location = event.state.$url;
 		}
 	}
 


distantnative:
@afbora your testing would be great on this one

afbora:
@distantnative Language drop down seems fixed but field contents still old:

Steps:
- Open a page (main language), edit and save
- Switch secondary language and use browser back (same for using forward)
- See old content value

distantnative:
Ugh, so the popped state is old old... :(

distantnative:
Trying it with a hard reload on pop event now...

distantnative:
Found the next problem with that myself:
- switch to non-default language
- change something but don't save
- hit back button
- unsaved change now gets applied to default language

afbora:
Nice working 🎉 except a little dialog and console error:

> TypeError: Cannot read properties of null (reading '$url')

Steps:

- Open fresh browser tab
- Enter panel login address and log in
- Use browser back
- See dialog and console error

afbora:
> Found the next problem with that myself:

I can't reproduce that issue 🤔

distantnative:
@afbora realised it's messed up: happens for me with the time field when I changed the value but am still in the field and then hit the back button. I think what must happen is that on field blur the input gets sent to the content store at the same time that the pop event takes place.

afbora:
I don't know what issue you're having, but it seems to work great for me now 🎉

afbora:
> Found the next problem with that myself:

I can reproduce now with following steps:

- Open `Notes` page in starterkit panel
- Edit `Exploring the universe` page in notes
- Edit a block without saving
- Use browser back
- See unsaved changes for `Notes` page

I have no idea about that 🤷‍♂️

distantnative:
@afbora tried a more hardcore version with `window.location` which I think fixes most problems, but prevents us to go forward after going backwards once :/

distantnative:
I keep thinking that something like

```js
async popState(event) {
	this.setState(event.state, { replace: true });
}
```

should be the right way.

But that has the same content store problem you described, @afbora .

afbora:
> @afbora tried a more hardcore version with `window.location` which I think fixes most problems, but prevents us to go forward after going backwards once :/

@distantnative It seems to be working for me. Both backwards and forwards work without any problems. Am I missing something?

PR #5071, opened 2023-02-25
Image orientation Exif

### LINKED ISSUE #2695: Exif orientation tag related issues and how to solve them
**Describe the bug**  

If an image includes [EXIF orientation code](https://www.impulseadventure.com/photo/exif-orientation.html), both GD and ImageMagick behave wildly inconsistently: It might be anything from resizing with dimensions switched, but correct orientation (i.e. `thumb(400, 300)` creates a 300x400 image), through stretched images, etc.

**To Reproduce**  
1. Go to any image field
2. Test any of the images available here: https://github.com/recurser/exif-orientation-examples
3. See the bug

**Screenshots**  

These are screenshots from one of the many seemingly unrelated problems, in this case it's from https://github.com/flokosiol/kirby-focus/issues/50

79700605-29bac500-8297-11ea-8a36-381342cafa9a
79700610-36d7b400-8297-11ea-9724-a3888179f734

**Additional context**  

Few of the solutions found on the internet deal with this by basically "burning the orientation in" and removing the exif orientation tag on upload.

Some examples are in [this forum thread](https://forum.getkirby.com/t/problem-with-image-orientation-on-upload/13794/9), my own looks like this:

``` PHP
<?php

use Kirby\Cms\App;
use Kirby\Toolkit\Str;

App::plugin('adamkiss/fix-orientation', [
	'hooks' => [
		'file.create:after' => function ($file) {
			if (in_array(Str::lower($file->extension()), ['jpg', 'jpeg'])) {
				$filename = $file->root();
				$image = imagecreatefromjpeg($filename);
				$exif = exif_read_data($filename);

				if (!empty($exif['Orientation'])) {
					switch ($exif['Orientation']) {
						case 3:
							$image = imagerotate($image, 180, 0);
							break;

						case 6:
							$image = imagerotate($image, -90, 0);
							break;

						case 8:
							$image = imagerotate($image, 90, 0);
							break;
					}
				}
				imagejpeg($image, $filename);
			}
		}
	]
]);
``` 

One problem with this is that the "fix" changes picture files, which might not be always applicable, so maybe configurable option "burn in exif orientation", with default state either backwards compatible (off) or forward problem solving (on) :)

distantnative:


rasteiner:
What exactly do you want to test? 
The "auto orient" feature of SimpleImage and ImageMagick seemed to actually work (the images Adam posted actually were all facing the right direction), the problem was just that some of them were stretched because `Dimensions` got the geometry wrong. 
ImageMagick (without auto-orient option) then actually kept the exif tag in the output file, therefore browsers might actually have interpreted those and showed the images in the right orientation and aspect ratio (unless users put `width` and `height` onto the `` tag, when taking those from Kirby), while GD always stripped the tags... That might explain some differences.

In the end it seems that all that needs to be tested, unless you also want to the test the auto orient feature of SimpleImage and ImageMagick as a kind of "integration test", is that Dimensions always gives you "50x100" for all of my images.

Something like:

```php

function testForImage() {
	for($i = 1; $i <= 8; $i++) {
		$dimensions = Dimensions::forImage(static::FIXTURES . "/image/orientation_$i.jpg");
		$this->assertSame(50, $dimensions->width());
		$this->assertSame(100, $dimensions->height());
	}
}

```

If you also want to test the auto-orient feature, one approach could be something like here: https://github.com/rasteiner/exif-orientation-test-images/blob/main/src/Tester.php (measuring pixel colors)

distantnative:
When I only changed `Dimensions::forImage()` the images got actually deformed during the upload. That's why both changes were added. And I would like to have tests that cover all variations and aspects of this for both drivers, so we can be sure in the future that any change won't revert it for any driver. This has kept popping up and I'm tired of this exif weirdness, so when we merge this, it must include unit tests that keep us safe going forward.

rasteiner:
It's absolutely right to have both changes :).
Since Kirby is now reading the orientation, it's right that the drivers now orient the images **before** applying the width / height that Dimensions is giving them. 
I guess the problem, even after this, will pop up *again* for people that disable "auto-orient":

 - user uploads a portrait of his face: the data is saved as landscape 1000x500, but the exif tag rotates it on its side (ending up as 500x1000)
 - user disables auto orient via config thinking "browsers are intelligent nowadays"
 - Dimensions says that it is 500x1000, and generates a 50% thumbnail like ``
 - GD strips away the tag
 - browser ends up with a face that is rotated 90 degrees and is also stretched

Maybe Dimensions should ignore Exif for people that disable "auto orient", the image will still end up rotated, but at least not stretched...

lukasbestle:
I agree. Either we auto-orient or we don't. Doing it partly is weird.

distantnative:
I've been a bit hesitant here because I think your suggestion mixes up two related functionalities but they are not the same. We need to think of both:

- What does `Image\Dimensions` return for all these images with different exif info? This should already be always correct just on its own.
- How to our thumb drivers generate resized versions from this? There we should differentiate between auto-orient active or not.

I don't think it's right if we mix in whether auto orient is active in the first one. The Dimensions class should give us values we consider correct no matter of any resizing options.

lukasbestle:
If we only use the EXIF data for the thumb but not for the image dimensions, we will have the issue that Roman has explained with his example above.

I'm thinking whether we should deprecate `autoOrient` completely. In which use case would you want to disable it? If we always use EXIF rotation data both for image dimensions and the thumb, the result will always be correct, no?

distantnative:
Closing this for now, PR is referenced in issue to revisit the comments in here later.

Issue #4869, opened 2022-11-28
Language snippet is not translated in user blueprint


chrfickinger:
## Description
A blueprint of a page allows to specify a language snippet for title, e.g.
```yml
title: my.translated.snippet
```

This behavior does not work for User blueprints. I have to put the translated stuff right away in the blueprint, like:
```
title:
    de: German string
    en: English string
```

afbora:
I can't reproduce the issue. How do you define the `my.translated.snippet` string?

chrfickinger:
Hi Ahment.
I could help myself.
Strings are defined correctly (e.g. `site/languages/en.php`), but it seems to be necessary to fill in both language files for it to work.

Example: If I have set `english` as panel language, `german` must be present as well, otherwise the snippet will not be translated.

chrfickinger:
i forgot to mention, english is the default language in my project and it is not set.
So, the solution is: Every snippet must be available at least in default language to work.

afbora:
- Default language is EN
- Secondary language is DE
- Related translation is defined in DE but not in EN

Is this correct?

chrfickinger:
Yes, correct.

But there is another strange behavior: I added the missing language snippet, but it's not translated into secondary language. Only default language is used, no matter which language is selected.

afbora:
Weird! Could you post your translation snippet, please? And which Kirby version do you use?

chrfickinger:
Kirby 3.8.2

`site/language/en.php`
```
return [
    'code' => 'en',
    'default' => true,
    'direction' => 'ltr',
    'locale' => [
        'LC_ALL' => 'en_GB'
    ],
    'name' => 'English',
    'translations' => [
        'panel.user.admin.title' => 'Admin',
        'panel.user.admin.description' => 'All permissions',
        'panel.user.editor.title' => 'Editor',
        'panel.user.editor.description' => 'Restricted permissions'
    ],
    'url' => '/'
];
```
`site/language/de.php`
```
return [
    'code' => 'de',
    'default' => false,
    'direction' => 'ltr',
    'locale' => [
        'LC_ALL' => 'de_DE'
    ],
    'name' => 'Deutsch',
    'translations' => [
        'panel.user.admin.title' => 'Administrator',
        'panel.user.admin.description' => 'Alle Berechtigungen',
        'panel.user.editor.title' => 'Redakteur',
        'panel.user.editor.description' => 'Eingeschränkte Berechtigungen'
    ],
    'url' => 'de'
[...]

PR #5072, opened 2023-02-25
Fix translated blueprint title

distantnative:
## This PR …

The main change is here: https://github.com/getkirby/kirby/pull/5072/files#diff-b52d06f8dd9e8c5cfc7e124ef31d4654d293561b8ab1a75b65a4efad5b24dd28R322
Moving `?? $value` outside the method call.
The rest is cleaning up, so I could even understand our code :D

### Fixes
- Blueprints: when looking up i18n strings, fallback locales now take precedences over the i18n key as fallback
#4869 

### Breaking changes
None known

## Ready?

### For review team

### DIFF
--- a/src/Cms/Blueprint.php
+++ b/src/Cms/Blueprint.php
@@ -77,7 +77,8 @@ class Blueprint
 		$props['name'] ??= 'default';
 
 		// normalize and translate the title
-		$props['title'] = $this->i18n($props['title'] ?? ucfirst($props['name']));
+		$props['title'] ??= ucfirst($props['name']);
+		$props['title']   = $this->i18n($props['title']);
 
 		// convert all shortcuts
 		$props = $this->convertFieldsToSections('main', $props);
@@ -311,14 +312,14 @@ class Blueprint
 
 	/**
 	 * Used to translate any label, heading, etc.
-	 *
-	 * @param mixed $value
-	 * @param mixed $fallback
-	 * @return mixed
 	 */
-	protected function i18n($value, $fallback = null)
-	{
-		return I18n::translate($value, $fallback ?? $value);
+	protected function i18n(
+		string|array|null $value,
+		string|array $fallback = null
+	): string|null {
+		// use $value as final fallback to show i18n strings
+		// (e.g. my.editor.title) when no actual translation is found
+		return I18n::translate($value, $fallback) ?? $value;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


afbora:
@distantnative I am curious about something:

- Default language EN, secondary language DE.
- Translation `panel.user.editor.title` is defined as `Editor` for EN but not for DE.
- Is it normal to still show `panel.user.editor.title` string when DE language is enabled, instead `EN` translation should appear as fallback `Editor`?

distantnative:
@afbora I'd that still the case? I thought this would fix it. The new test I added is specifically that use case.

afbora:
Unfortunately, doesn't work for me:

````
<?php
// /site/languages/en.php
return [
    'code' => 'en',
    'default' => true,
    'direction' => 'ltr',
    'locale' => [
        'LC_ALL' => 'en_US'
    ],
    'name' => 'EN',
    'translations' => [
        'role.editor' => 'Editor',
    ],
    'url' => NULL
];

<?php
// /site/languages/de.php
return [
    'code' => 'de',
    'default' => false,
    'direction' => 'ltr',
    'locale' => [
        'LC_ALL' => 'de_DE'
    ],
    'name' => 'DE',
    'translations' => [],
    'url' => NULL
];

// /site/blueprints/users/editor.yml
title: role.editor
````

- Apply above setup
- Change your account language to DE
- Go to `Users` page and try to create new account
- See editor title issue

**Comment by afbora** [comment]
@distantnative I've added unit test https://github.com/getkirby/kirby/pull/5072/commits/5e878f664fd59eee8a5e3cf369671c0f93a30f00 that fails for you can reproduce the issue. You can delete or improve the unit test.

**Comment by afbora** [comment]
@distantnative Works for user blueprints 🎉 Btw, page blueprints `title` prop also support translation keys, right?

**Review by bastianallgeier — APPROVED** [review]

File: src/Cms/Blueprint.php [hunk 1]
```diff
@@ -341,20 +342,10 @@ public static function load(string $name): array
...
-			// inject the filename as name if no name is set
-			$props['name'] ??= $name;
-
-			// normalize the title
-			$title = $props['title'] ?? ucfirst($props['name']);
+		// inject the filename as name if no name is set
+		$props['name'] ??= $name;
 
-			// translate the title
-			$props['title'] = I18n::translate($title, $title);
```
afbora, line 352: @distantnative What was the reason for removing these lines? This is the cause of the #5086 issue.

distantnative: I made this change in another PR as well cause it looked like it just does the same again in the constructor, but maybe not

PR #5073, opened 2023-02-25
UUID: more conservative ID generation

distantnative:
## Please review but do not merge

### DIFF
--- a/src/Uuid/Uuid.php
+++ b/src/Uuid/Uuid.php
@@ -40,7 +40,7 @@ use Kirby\Toolkit\Str;
  * @copyright Bastian Allgeier
  * @license   https://getkirby.com/license
  */
-class Uuid
+abstract class Uuid
 {
 	protected const TYPE = 'uuid';
 
@@ -213,12 +213,10 @@ class Uuid
 	/**
 	 * Returns the UUID's id string (UUID without scheme);
 	 * in child classes, this method must ensure that the
-	 * model has an ID
+	 * model has an ID (or generate a new one if the model
+	 * does not have one yet)
 	 */
-	public function id(): string
-	{
-		return $this->uri->host();
-	}
+	abstract public function id(): string;
 
 	/**
 	 * Generator function that creates an index of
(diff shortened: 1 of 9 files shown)


src/Uuid/Uuid.php
```diff
@@ -271,21 +275,30 @@ final public static function is(
...
 	 */
-	public function key(): string
+	public function key(bool $generate = false): string|null
 	{
-		$id = $this->id();
-
-		// for better performance when using a file-based cache,
-		// turn first two characters of the id into a directory
-		$id = Str::substr($id, 0, 2) . '/' . Str::substr($id, 2);
+		$id = $generate ? $this->id() : $this->uri->host();
```
lukasbestle: For clarity:

```suggestion
		// the generation happens in the child class
		// that overrides the `id()` method
		$id = $generate === true ? $this->id() : $this->uri->host();
```

lukasbestle (commented):
Reading through the diff I like this implementation a lot and feel like it is almost there. Two thoughts:

- Wouldn't `Uuid::id()` need to return `string|null`, already before these changes? `ModelUuid::id()` will always return `string`, but the basic implementation doesn't ensure that the ID is generated.
  - So I think we need to change the type hint of `Uuid::id()`.
  - We also need to check whether there are any uses of the `id()` and `key()` methods in our code that always expect a `string`.
  - The change to `key()` is a breaking change. Which is fine, but we need to add it to the release notes.
- Do you feel like we have sufficient automated tests for the UUID business logic already? Because I think just from reading the code changes, we will easily miss issues. But if the tests already cover many edge cases, we can merge this more confidently.

distantnative:
@lukasbestle 
- I turned `Uuid::id()` into an abstract method that way we can keep only `string` as return type. Cause you were right that in theory the code would allow `null`, but none of the existing Uuid classes could actually result in null: page and file are covered by `ModelUuid` which would generate the ID, site always can return an empty string as ID, users are guaranteed to have an ID set on initiation
- Also reworked the `findByCache` methods to support `->key()` returning `null`
- Unit tests: it's super hard for me to say. I think test coverage isn't bad, both measured but also what cases I could think of, but of course that's super limited by my perspective and oversights

lukasbestle:
Sorry for the delay, finally read through your changes and the whole diff again.

I really like it and I think we can merge it like this. **However** I'd merge it into v4 because of the breaking changes. If I didn't miss anything, we have the following:

- New behavior of `Uuid::key()` (does not generate a UUID by default)
- `Uuid` is now an abstract class, the `id()` method needs to be defined in a child class and is responsible to generate a UUID if it wasn't already set.

distantnative:
I agree with v4, that's why I marked it o not be merged here. Just wanted to keep it here for review with the previous discussion.

Would you be ok that I merge it directly into v4-develop?

PR #5074, opened 2023-02-25
Fix copy&paste for nested blocks

### LINKED ISSUE #4282: Copy&Paste not working on nested blocks
## Description
Copy&Paste not working on nested blocks

When copying blocks with keystroke or with options menu, pasting them does result in the paste dialog being shown, but cmd+v makes the dialog disappear and then nothing else happens.

On the parent block everything's working as expected.

## To reproduce

create a nested blocks setup on a page:

```yaml
fields:
  blocks:
    type: blocks
    fieldsets:
      nested:
        name: Nested
        label: Nested
        fields:
          blocks:
            type: blocks
            fieldsets:
              item:
                name: Item
                label: Item
                fields:
                  title:
                    type: text
```

## 

## Your setup
**Kirby Version**  

3.6.5

**Console output**  

no errors or warnigns in the console

 - OS: macOS
 - Browser: Latest stable Chrome, latest stable Firefox

distantnative:
## This PR …

- Blocks `isEditing` state has been global based on any open dialog/drawer
- PR localises this state and tracks opening/closing the drawer instead

### Fixes
- Nested blocks can be copied/pasted 
#4282

## Ready?

### For review team

- [X] ~~Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)~~

### DIFF
--- a/panel/src/components/Forms/Blocks/Block.vue
+++ b/panel/src/components/Forms/Blocks/Block.vue
@@ -43,8 +43,12 @@
 			:title="fieldset.name"
 			:value="content"
 			class="k-block-drawer"
-			@close="focus()"
+			@close="
+				focus();
+				$emit('close');
+			"
 			@input="$emit('update', $event)"
+			@open="$emit('open')"
 		>
 			<template #options>
 				<k-button
--- a/panel/src/components/Forms/Blocks/Blocks.vue
+++ b/panel/src/components/Forms/Blocks/Blocks.vue
@@ -174,7 +174,7 @@ export default {
 	},
 	created() {
 		this.$events.$on("blur", this.onBlur);
-		this.$events.$on("copy", this.copy);
+		this.$events.$on("copy", this.onCopy);
 		this.$events.$on("focus", this.onOutsideFocus);
 		this.$events.$on("keydown", this.onKey);
 		this.$events.$on("keyup", this.onKey);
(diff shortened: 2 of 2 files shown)


afbora:
I don't understand the codes but it seems to work great 🙈 🎉 

---

I just wondering the codes, not an issue. When the blocks field is opened, `isEditing = true` is set, but the `copy()` method should returns `false` when `isEditing = true`. I added `console.log(this.isEditing);` to the beginning of the `copy()` method. When I tried to copy with `CTRL + C` I noticed that `copy()` method works two times and the first one returns `true` and the second one returns `false`. But I don't understand how.

distantnative:
@afbora I think I know what's happening. Our `Blocks.vue` is listening for just copy/paste commands globally. So the 2 console entries you (and I see) aren't from the one blocks field but - at least in my case - from the two blocks field I have on the same page. Baffled that it's been like that for all the time - but likely the `this.selectedOrBatched.length` check always saved us from worse.

afbora (approved):
Working great for me 💛

bastianallgeier:
Do we want to add this to 3.9.2?

afbora:
3.9.2 would be great if you, as the author of the blocks field, think the PR is ready. Otherwise let's move it to the next version.

PR #5076, opened 2023-02-26
Panel: Fix `field.isVisible` helper 

### LINKED ISSUE #4838: Select field has no initial value in structure/object fields

distantnative:
## This PR …

- We cannot simply add an empty string as fallback here https://github.com/getkirby/kirby/pull/5076/files#diff-04e8dc6763a997696e500ceeb03a7f770ea1c14df821298111e13dbecaaeef5eR12 - this will screw with all fields that don't expect a string as value prop (e.g. blocks). They rely on `undefined` to have their prop default kick in.
- Instead we check in the `inVisible` helper if the `when` condition is set to an empty string that `undefined` will be just as fine

### Fixes
- Fix displaying conditional fields for empty `when` condition
#4838

### Enhancements
- New `$helper.field.form(fields)` function to set up values for an empty form

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Field/ObjectField.vue
+++ b/panel/src/components/Forms/Field/ObjectField.vue
@@ -93,16 +93,7 @@ export default {
 	},
 	methods: {
 		onAdd() {
-			this.object = {};
-
-			for (const fieldName in this.fields) {
-				const field = this.fields[fieldName];
-
-				if (field.default) {
-					this.object[fieldName] = this.$helper.clone(field.default);
-				}
-			}
-
+			this.object = this.$helper.field.form(this.fields);
 			this.$emit("input", this.object);
 			this.open();
 		},
--- a/panel/src/components/Forms/Field/StructureField.vue
+++ b/panel/src/components/Forms/Field/StructureField.vue
@@ -369,15 +369,8 @@ export default {
 				return false;
 			}
 
-			// create entry data from field defaults
-			let data = {};
-
-			for (const fieldName in this.fields) {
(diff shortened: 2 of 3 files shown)


afbora (approved):
Works great and LGTM🎉

PR #5077, opened 2023-02-26
Panel: more shortcuts for search

distantnative:
### Enhancements
- Panel search now also opens on `Cmd+K` and `Cmd+/` in addition to `Cmd+Shift+F` (`Ctrl` on Windows)
https://kirby.nolt.io/511

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Navigation/Search.vue
+++ b/panel/src/components/Navigation/Search.vue
@@ -163,7 +163,8 @@ export default {
 				this.select(this.selected - 1);
 			}
 		},
-		open() {
+		open(e) {
+			e?.preventDefault();
 			this.$refs.overlay.open();
 		},
 		async search(query) {
--- a/panel/src/components/Navigation/Search.vue
+++ b/panel/src/components/Navigation/Search.vue
@@ -104,9 +104,13 @@ export default {
 	created() {
 		this.search = debounce(this.search, 250);
 		this.$events.$on("keydown.cmd.shift.f", this.open);
+		this.$events.$on("keydown.cmd.k", this.open);
+		this.$events.$on("keydown.cmd./", this.open);
 	},
 	destroyed() {
 		this.$events.$off("keydown.cmd.shift.f", this.open);
+		this.$events.$off("keydown.cmd.k", this.open);
+		this.$events.$off("keydown.cmd./", this.open);
 	},
 	methods: {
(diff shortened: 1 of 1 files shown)


afbora:
- ✅ CTRL + /  
- ✅ CTRL + Shift + F  
- ✅ CTRL + K  

Working great on windows but a little issue happening. When using `CTRL + K`, also browser search enabling and conflicts, here screencast:

https://user-images.githubusercontent.com/3393422/221522681-05a4cf2a-9390-4b5d-b982-8ce87c1e2ada.mp4

bastianallgeier:
@afbora I think we should leave it like that. It wouldn't feel right to me to block the default behaviour. What do you think? /cc @distantnative

afbora:
Yes, I agree with Bastian. Our use of browsers default shortcuts doesn't sound very functional and correct. If no browser is using `CTRL/CMD + /` can stay.

distantnative:
The thing is Cmd+k was the one actually requested. So we would need to split it up, support only that one on Mac while not adding ctrl+k on windows?

afbora:
It doesn't make sense not to apply it to one and apply it to the other. There are two options:
- Assign the same key in Windows and MacOS
- Assigning different keys in Windows and MacOS

If possible, I would be in favor of the first option.

distantnative:
I would also say that we should have the same between both OS.... but then we cannot do `Cmd+K` at all :/

afbora:
@distantnative I've tested `CTRL + K` on notion web app, works great without conflicting. I think, notion overwrite/prevent browser shortcut. You should take a look.

afbora:
I've tested following changes and seems works great, what do you think?

```diff
- open() {
+ open(e) {
+  e.preventDefault();
   this.$refs.overlay.open();
},
````
https://github.com/getkirby/kirby/blob/fix/panel-search-keyshortcut/panel/src/components/Navigation/Search.vue#L166-L168

distantnative:
@afbora does this work for you?

afbora (approved):

distantnative:
I think this one would be safe enough to still include in 3.9.2

lukasbestle (commented):
I haven't been able to test it myself, but if it works I think it's a great addition.
👍 on 3.9.2.

PR #5078, opened 2023-02-26
Clean up `Toolkit` package

distantnative:


src/Toolkit/A.php
```diff
@@ -39,11 +39,10 @@ public static function append(array $array, array $append): array
...
 	public static function apply(array $array, ...$args): array
 	{
-		array_walk_recursive($array, function (&$item) use ($args) {
-			if ($item instanceof Closure) {
-				$item = $item(...$args);
-			}
-		});
+		array_walk_recursive(
+			$array,
+			fn (&$item) => $item = $item instanceof Closure ? $item(...$args) : $item
```
lukasbestle: Is this really so much easier to read? Also it will return `$item`. Have you checked if `array_walk_recursive()` can deal with that?

src/Toolkit/Date.php
```diff
@@ -70,20 +69,17 @@ public function ceil(string $unit)
 	/**
 	 * Returns the interval between the provided and the object's datetime
 	 *
-	 * @param string|int|\DateTimeInterface $datetime
-	 * @param \DateTimeZone|null $timezone Optional default timezone if `$datetime` is string
-	 * @return \DateInterval
+	 * @param \DateTimeZone $timezone Optional default timezone if `$datetime` is string
```
lukasbestle: This param is still optional (same a lot of times below). Why have you removed the `null` option?

```suggestion
	 * @param \DateTimeZone|null $timezone Optional default timezone if `$datetime` is string
```

src/Toolkit/Date.php
```diff
@@ -70,20 +69,17 @@ public function ceil(string $unit)
...
 	 */
-	public function compare($datetime = 'now', ?DateTimeZone $timezone = null)
-	{
+	public function compare(
+		string|int|DateTimeInterface $datetime = 'now',
+		DateTimeZone $timezone = null
+	): DateInterval {
```
afbora: Should we consider only this PR or the whole Toolkit package? For example, this change is not applied to `Date::__construct()` (due to run `new static($datetime, $timezone)` in following codes). Do you want me to complete the missing ones for the whole Toolkit package with a new PR?

lukasbestle: Let's start with these classes first and open the next PR afterwards. Otherwise it's going to be too much at once. :)

lukasbestle: Let's start with these classes first and open the next PR afterwards. Otherwise it's going to be too much at once. :)

afbora: OK, you're right 👍

src/Toolkit/A.php
afbora: I must admit that it took me a while to understand this code 🙈 and I was undecided whether to comment or not.

src/Toolkit/Date.php
afbora: A similar situation happened here: https://github.com/getkirby/kirby/pull/5079#discussion_r1119076118. I think Nico removed it specifically for use cases where the `null` parameter is not sent.

src/Toolkit/Date.php
```diff
@@ -472,21 +441,16 @@ public function timestamp(): int
 
 	/**
 	 * Returns the timezone object
-	 *
-	 * @return \DateTimeZone
 	 */
-	public function timezone()
+	public function timezone(): DateTimeZone
```
afbora:
```suggestion
	public function timezone(): DateTimeZone|false
```

src/Toolkit/Str.php
```diff
@@ -249,8 +261,11 @@ public static function beforeEnd(string $string, string $needle, bool $caseInsen
 	/**
 	 * Returns everything between two strings from the first occurrence of a given string
 	 */
-	public static function between(string $string = null, string $start, string $end): string
-	{
+	public static function between(
+		string $string = null,
```
afbora, line 265: Why `$string` argument defaults to `null`. It doesn't make sense to me to send other parameters without sending this parameter. Shouldn't it be `string $string` or  `string|null $string = null`? The same situation exists in a few other methods in this class.

afbora (commented):
Also I think there is a serious inconsistency between the `$string` arguments of the methods in the `Str` class. Some methods accept `null`, some don't accept `null` at all, some default to `null`. Maybe this is not the topic of this PR, but I wanted to point it out.

src/Toolkit/Date.php
distantnative: Don't we have that in many places where we mark parameters as optional via `= null` but don't add `null` again as type hint before the variable name? As this mainly should read "$timezone needs to be of DateTimeZone, but is optional", not that `null` is an actual type for `$timezone` that had another function/meaning.

Given, we also have a lot of places with still `|null` - but I think that's actually not right in most of these.

src/Toolkit/Str.php
distantnative: I think it's a historical reason - my assumption I explained here too https://github.com/getkirby/kirby/pull/5032#discussion_r1128538005 - that the `Str` methods should be forgiving when applied, e.g. on a field that actually in the end returns `null`. So not to get a PHP error right away.

distantnative:
@afbora Yea, I don't really want to tackle that with this PR. It's true but then this would become quite some refactoring with breaking changes

src/Toolkit/Date.php
distantnative: @lukasbestle @afbora thoughts?

lukasbestle: But then it needs a `?` before the type, doesn't it? The docblock type overrides the type-hint, so for machines it's no longer visible that the argument is optional.

And given that we got rid of the `?` in a lot of places (or even everywhere?), I think that just leaves `|null`.

distantnative: Doesn't `= null` marks it for machine as well as optional? Isn't that the implication?

distantnative: Ok, now I understood it: you're fine with just `= null` in the actual PHP code, but in the DocBlock you think it's still required as this will take priority for the docs?

lukasbestle: Ah, I see what you mean. If you use `type $attr = null`, PHP automatically treats this as `type|null $attr = null` internally. But IMO this is a not an actual PHP feature but just a compatibility hack. They probably went with this auto-conversion to avoid that a lot of code bases break.

IMO we shouldn't rely on this compatibility hack and rather type-hint explicitly. An alternative could be the `?type` syntax (which to me looks like a good fit for the "optional argument" use case), but we already decided against that because it's not compatible with multiple types.

So I think that only leaves us `|null`.

distantnative: I added it. Also rebased the PR but only the last commit is new/has the new changes.

src/Toolkit/Date.php
```diff
@@ -472,21 +445,16 @@ public function timestamp(): int
...
 	}
 
 	/**
 	 * Returns an instance of the beginning of the current day
-	 *
-	 * @param \DateTimeZone|null $timezone
-	 * @return static
 	 */
-	public static function today(?DateTimeZone $timezone = null)
+	public static function today(DateTimeZone $timezone = null): static
```
afbora:
```diff
- public static function today(DateTimeZone $timezone = null): static
+ public static function today(DateTimeZone|null $timezone = null): static
```

I apologize for taking this so long 🙈 Shouldn't it be like this?

afbora (approved):
LGTM 👍

bastianallgeier:
Would merge this after the 3.9.4 release.

PR #5079, opened 2023-02-26
Clean up `Toolkit\Dom`

distantnative:


src/Toolkit/Dom.php
```diff
@@ -502,12 +495,12 @@ public static function remove(DOMNode $node): void
...
 	 *
-	 * @param string $query
 	 * @param \DOMNode|null $node Optional context node for relative queries
-	 * @return \DOMNodeList|false
 	 */
-	public function query(string $query, ?DOMNode $node = null)
-	{
+	public function query(
+		string $query,
+		DOMNode $node = null
```
afbora:
```suggestion
		DOMNode|null $node = null
```

afbora (approved):

src/Toolkit/Dom.php
distantnative: @bastianallgeier @afbora was this really needed? From my view it wouldn't make much sense to support actively passing `null` here, or does it? Only to leave it optional (so `= null` but when passing something it shouldn't be null).

afbora: You're right, maybe we don't have such a use case, but I wanted to implement it to be consistent with `DOMXPath::query()`. It's not very important.

image (17)

PR #5081, opened 2023-02-27
Feature: Add `options` to `collection()` helper

ceesvanegmond:
## This PR adds extra functionality to the existing `collection()` helper.

I love to use this helper but I miss the functionality to pass in extra arguments.
This PR fixes this and add an option to send an array of extra arguments to the `collection()` helper.

For example; take a look at the following code

```php
<?php
// news.php
$articles = collection('news');

// collection/news.php
use Kirby\Cms\Site;

return function (Site $site) {
    return $site->find('news')->children()->template('article');
};
```

Maybe sometimes you want to receive the news articles based asc on publish date; sometimes you want to it reversed.
It would be a bit of a shame to make two collections for it. Passing an extra argument is much easier!

```php
<?php
// news.php
$articles = collection('news', ['order' => 'asc');

// collection/news.php
use Kirby\Cms\Site;

return function (Site $site, string $order) {
    return $site->find('news')->children()->template('article')->orderBy($order);
};
```
You can even build complete complex filter systems with this extra argument.

### Breaking changes
- None; everything is backwards compatible

## Ready?

### For review team


afbora:
IMHO, filtering collections by sending parameters seems useful. I like the idea 👍

sietseveenman (dismissed):

distantnative:
Can't it just be called `$options`? Why `extra`?

ceesvanegmond:
@distantnative I think you're right; `$options` is a better name! Just changed in my latest commit.

src/Cms/App.php
```diff
@@ -365,16 +365,17 @@ public function clone(array $props = [], bool $setInstance = true)
...
+	public function collection(string $name, array $options = [])
 	{
-		return $this->collections()->get($name, [
+		return $this->collections()->get($name, array_merge([
 			'kirby' => $this,
 			'site'  => $this->site(),
 			'pages' => $this->site()->children(),
 			'users' => $this->users()
-		]);
+		]), $options);
```
lukasbestle: I think the merge order should be the other way around so that the core objects always override the passed options. This ensures that the core objects can always be relied on.

tests/Cms/Collections/CollectionTestCase.php
```diff
@@ -7,11 +7,12 @@
 class CollectionTestCase extends TestCase
 {
 	public $collection = null;
+	public $collectionOptions = [];
```
lukasbestle, line 10: Could you please add unit tests for the new behavior?

- Passing an option to a collection and using it inside the collection
- Trying to access an option inside a collection that was not passed (should be provided to the collection function as `null`)
- Trying to override a core variable like `$site` with an option (which should be ignored)

lukasbestle (commented):
I like it a lot too! 👍
Just a few comments:

ceesvanegmond:
@lukasbestle I added the tests and swapped the arguments. What I do see is that there are several assert methods that doesn't do anything right now. For example:

```php
// tests/Cms/Collections/CollectionTestCase.php

public function collectionRoot() {} 
public function collectionPagination() {} 
public function assertCollectionCount(int $count) {} 
public function assertCollectionHasPagination() {} 
public function assertCollectionHasNoPagination() {} 
```

and the ```public function testCollectionType()``` is beiing skipped.
Is there an reason why these methods are here? Could be removed no?

src/Cms/App.php
```diff
@@ -365,16 +365,17 @@ public function clone(array $props = [], bool $setInstance = true)
...
+	public function collection(string $name, array $options = [])
 	{
-		return $this->collections()->get($name, [
+		return $this->collections()->get($name, array_merge($options, [
 			'kirby' => $this,
 			'site'  => $this->site(),
 			'pages' => $this->site()->children(),
 			'users' => $this->users()
-		]);
+		]));
```
distantnative, line 378: Is it intended that `$options` cannot overwrite these defaults?

tests/Cms/Collections/fixtures/collections/options_with_default.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+use Kirby\Cms\App;
+
+return function (int $a, int $b, int $c, App $kirby) {
```
distantnative, line 5: If I understand the code correctly, like in controllers, we don't need to rely on the order of arguments. Would be good to have one test where the Closure has the arguments in a different order than they are passed in the test to the callback.

src/Cms/App.php
lukasbestle: It was the other way around originally, but I suggested to implement it like this: https://github.com/getkirby/kirby/pull/5081#discussion_r1119828638

Do you see a use case to override the core objects?

lukasbestle:
> What I do see is that there are several assert methods that doesn't do anything right now. For example:

Oh, I think there is a misunderstanding. The file you are referring to is the `CollectionTestCase` class. It is not intended for tests, but as a base class for actual test classes like `CollectionTest`. But as far as I can tell, the `CollectionTestCase` is currently dead code. It was there since 3.0.0 but at least at the moment, no actual test class extends it. So I think we can remove that file completely.

Tests for the new behavior rather belong in the `tests/Cms/App/AppTest.php` and `tests/Cms/Helpers/HelperFunctionsTest.php` files next to the existing tests.

src/Cms/App.php
distantnative: I just found it a bit odd as we usually allow defaults to be overwritten by specifically provided options. But have no strong feelings for this particular use case.

distantnative:
Moved this here https://github.com/getkirby/kirby/pull/5333 to work on the last tests etc.

PR #5082, opened 2023-02-27
Clean up helpers

afbora:


config/helpers.php
```diff
@@ -173,7 +169,6 @@ function esc(string $string, string $context = 'html'): string
 	 *
 	 * @param mixed $key The key to look for. Pass false or null to return the entire request array.
 	 * @param mixed $default Optional default value, which should be returned if no element has been found
-	 * @return mixed
 	 */
 	function get($key = null, $default = null)
```
lukasbestle: This could be type-hinted with `: mixed` as well for clarity.

config/helpers.php
```diff
@@ -202,12 +193,8 @@ function gist(string $url, string|null $file = null): string
 	/**
 	 * Redirects to the given Urls
 	 * Urls can be relative or absolute.
-	 *
-	 * @param string $url
-	 * @param int $code
-	 * @return void
 	 */
-	function go(string $url = '/', int $code = 302)
+	function go(string $url = '/', int $code = 302): void
```
lukasbestle, line 199: Could use a `@todo` comment to change to `: never` once support for PHP 8.0 is dropped.

config/helpers.php
```diff
@@ -424,10 +368,6 @@ function markdown(string|null $text = null, array $options = []): string
 if (Helpers::hasOverride('option') === false) { // @codeCoverageIgnore
 	/**
 	 * Shortcut for `$kirby->option($key, $default)`
-	 *
-	 * @param string $key
-	 * @param mixed $default
-	 * @return mixed
 	 */
 	function option(string $key, $default = null)
```
lukasbestle: Also add `: mixed` here for clarity

config/helpers.php
```diff
@@ -505,10 +434,8 @@ function params(): array
 	/**
 	 * Smart version of return with an if condition as first argument
 	 *
-	 * @param mixed $condition
 	 * @param mixed $value The string to be returned if the condition is true
 	 * @param mixed $alternative An alternative string which should be returned when the condition is false
-	 * @return mixed
 	 */
 	function r($condition, $value, $alternative = null)
```
lukasbestle: Also add `: mixed` here for clarity

config/helpers.php
```diff
@@ -521,26 +448,22 @@ function r($condition, $value, $alternative = null)
...
-	 * @return mixed
 	 */
-	function router(string|null $path = null, string $method = 'GET', array $routes = [], Closure|null $callback = null)
-	{
+	function router(
+		string|null $path = null,
+		string $method = 'GET',
+		array $routes = [],
+		Closure|null $callback = null
+	) {
```
lukasbestle:
```suggestion
	): mixed {
```

config/helpers.php
```diff
@@ -549,10 +472,6 @@ function site()
 if (Helpers::hasOverride('size') === false) { // @codeCoverageIgnore
 	/**
 	 * Determines the size/length of numbers, strings, arrays and countable objects
-	 *
-	 * @param mixed $value
-	 * @return int
-	 * @throws \Kirby\Exception\InvalidArgumentException
```
lukasbestle, line 555: Why is this line removed?

config/helpers.php
```diff
@@ -615,14 +528,12 @@ function svg($file)
...
-	 * @param string|null $locale
-	 * @return array|string|null
 	 */
-	function t($key, string $fallback = null, string $locale = null)
-	{
+	function t(
+		string|array $key,
+		string|null $fallback = null,
+		string|null $locale = null
+	): array|string|null {
```
lukasbestle: The type hints don't fully match `I18n::translate()`

config/helpers.php
afbora: I know but https://github.com/getkirby/kirby/pull/4959#discussion_r1056177720 🙈

lukasbestle: Ah, right. I forgot about that conversation.

TBH I feel like `mixed` *does* add value because the user knows that the method may actually return any type (in contrast to a method where we forgot to document the return type or still need to clean the type hints).

Pinging @distantnative for thoughts.

afbora: There's no harm in not adding it but I'm definitely in favor of adding it. I think specifying the mixed type rather than nothing is very useful both for reading the code and for the IDE.

config/helpers.php
afbora: We discussed long ago that we only use the `@throws` tag for exceptions thrown within the scope.

config/helpers.php
afbora: ✅

config/helpers.php
afbora: ✅

config/helpers.php
lukasbestle: That makes sense. At least it's consistent then. For users it's a lot less useful, but without tools it will be incredibly hard to track all exceptions to every public method where they could be thrown.

config/helpers.php
distantnative: I think for helpers it's ok, but what you need to keep in mind is that adding a return type hint can be quite the breaking change. If we add it on a class method and someone extended the class but their method now doesn't have the same signature, it'll break completely.

So we said we would only want to do this for places where adding a type hint really has a benefit, more than just a reminder for us. I still think not adding mixed has more benefits than adding it.

lukasbestle: If we really don't want to do this for methods, I think we shouldn't do it for helpers either.

I guess we just disagree on whether we want to do this in general. We also add type hints for other return types all the time in the cleanup PRs. Is `mixed` different here?

distantnative: Yes, other type hints actually help us to make methods safer (as they check that only those types are returned), better UX in IDEs etc.

Adding `mixed` doesn't add anything functionally, just visually.

distantnative: We also arrived at this decision because I had added `mixed` in various places and got feedback during an RC that now many extended classes of developers broke. For too little benefit, in
my view.

afbora: An article about that 😅 
https://devalmonte.com/blog/2018-10-10/why-we-need-mixed-types-in-php/

distantnative: Even after reading the article I don't see how this little added explicitness compensates the breaking character of such changes.

lukasbestle: The `mixed` type *does* show up in IDEs and it also makes it clearer for us that the type really is mixed. Alternatively we could also keep the docblock type hints for that, but TBH I don't think that's the right way considering that PHP natively supports it.

We had those breaking changes in that one PR that we reverted, but that was a special case IMO. 95% of classes are not extended from plugins. And even if they are, `mixed` hints are not going to be the only problems. As I wrote we are already in the process of adding all the other types anyway.

distantnative: Yes, my point was that the same shows up in IDEs also when you don't add it. Not specified = mixed, if anything that isn't mixed is specified.

But if you all don't mind more potential for breaking changes, let's do it. I am not against `mixed` - I just was afraid to break too much for much less benefit than e.g. type hinting a string.

afbora: If `mixed` was added to all the clean up PRs we are doing now, what kind of breaks would happen? There is no difference between adding `string` or `mixed` to the return type of a method in a non-extended class. On the contrary, if it is an extended class, whatever you add to the return type will be broken. I don't think this is a situation related to `mixed`.

distantnative: Yes, but we are turning in circles. I keep saying that `string` etc. also are breaking changes but they add a significant benefit of type checking, while adding `mixed` functionally is the same as adding no hint (also IDEs would already assume mixed if nothing is added), it's just a little visual help. My take was, the better type checking for string etc. is worth it, just the visual hint isn't. But since more of you see it differently, let's go with your take.

afbora: I completely agree. + But I think it would be useful to use this visual clue that PHP natively supports for places where there is no breakage change. 

I also find this discussion useful for improving the code. As I said, it's not critical for me, I can live without `mixed` 😅  I leave the final decision to you 🤝

lukasbestle: I'd say we can certainly do it for the helpers as it's not a breaking change here (functions cannot be extended).

In the future, we should always review if type hints are a breaking change (no matter if `string`, `mixed` or anything else). And that depends on whether the function/method can be extended by users or plugins. With the cache drivers it was the worst case, most of the time it's pretty much unlikely.

distantnative:
@afbora could you check my last commit as well? Thanks!

afbora:
Thank you Nico. Now I think ready to review and merge.

distantnative (approved):

PR #5083, opened 2023-02-28
3.9.2-rc.1

bastianallgeier:


lukasbestle (dismissed):
LGTM 👍

Regarding the translation file encoding: Looks like Transifex changed/updated their JSON library so it no longer encodes Unicode characters with escape sequences. But I'd say the translation strings look fine.

@bastianallgeier @distantnative Something seems to differ between your Composer versions because the autoloader files keep changing back and forth. Could you both please update to the latest version with `composer self-update`?

bastianallgeier:
Just updated my local composer version. I also agree that the encoding seems to not be a problem. I actually prefer that they no longer encode it.

lukasbestle (approved):

Issue #4225, opened 2022-03-24
Logical inconsistency while navigating through panel users (prev/next buttons disappear on own account)


grommas:
## Description
When logged in as user and navigating through the panel users, you can use the arrow buttons in the upper right corner of the header. When landing on your own user account, Kirby automatically redirects from `panel/users` to `panel/account` and the buttons disappear.

**Expected behavior**  
Since your own account can be between other users, I expect to stay inside `panel/users` to navigate through all users back and forward.

## To reproduce

1. Go to accounts
2. Click on random user (not yourself)
3. Navigate to your user account through arrow buttons
4. Page redirects, arrow buttons disappear

## Your setup
**Kirby Version** 
3.6.3.1

distantnative:
I think the only viable way here would be to separate the current user/account completely from other users, e.g. have two sections on the users view - first listing just the current account, second listing all other users. Between those other users one could then move prev/next without problems.

lukasbestle:
Would it be viable to use the User route also for the current account when it is accessed through the Users view? We already have access to the Account view from the menu, so I think the Users view shouldn't differentiate at all and use the User view for everyone.

distantnative:
If I recall correctly we had this before the fiber introduction and it led to quite a number of problems (I think it were situations like changing the language of the current user via the user not account view and similar actions that would have much more far reaching implications). But maybe @bastianallgeier remembers the details why we split this up fundamentally.

grommas:
I don't have expertise and insights like you have, but one idea could be to generally handle users in an own section including the current user. For the dropdowns current user option we could then filter this section to only show the current user. But that might not solve the complications you are talking about. Thanks for having a look!

bastianallgeier:
For now, I can only add a quick summary because it's still not easy to solve: 

- `$user->panel()->url()` uses the `/panel/accounts` URL for logged in users instead of linking to `/panel/users/$userid` I would say that this is still correct. Linking to the accounts page is a more resilient way of handling it because a user might not have access to the entire users area, but still should be able to edit their own data. 
- The users list and profiles in the users area use `$user->panel()->url()` to get to the right user view. This is why you end up in the accounts view if you click on your own user row. The prev/next buttons use this as well. This creates the inconsistency. 

Possible solutions: 

- we could either make the `$user->panel()->url()` method smarter to skip accounts in the users area
- we could also manually link to the user view and also change the prev/next buttons manually to avoid links to accounts.

PR #5088, opened 2023-03-02
Fix prev/next navigation for users view 

afbora:
## This PR …

I came up with a simple solution. By checking the last part of the URL, I can tell if the user is on their own page or not, so I can dynamically send different dynamic props to the fiber.

It might be a very stupid idea 🙈, so I created the PR as a draft. If it makes sense to you, I'll keep working on it.

### Fixes
- Fix prev/next navigation for users view #4225

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Panel/User.php
+++ b/src/Panel/User.php
@@ -143,12 +143,28 @@ class User extends Model
 	}
 
 	/**
+	 * Checks whether the user is on their own account page
+	 * by checking the session and the last part of the URL
+	 *
+	 * For examples:
+	 * /panel/users/*** > FALSE
+	 * /custom-panel-address/account > TRUE
+	 *
+	 * @internal
+	 */
+	protected function isAccountView(): bool
+	{
+		return $this->model->isLoggedIn() === true
+			&& $this->model->kirby()->request()->path()->last() === 'account';
+	}
+
+	/**
 	 * Returns the full path without leading slash
 	 */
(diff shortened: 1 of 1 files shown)


afbora:
Failed unit tests are related to urls. Do you mind sending requests to `/users/***` instead of `/account`? Requests are sent to `/account` only when they are on their own account page, otherwise they are made to `/users/***`.

src/Panel/User.php
```diff
@@ -142,13 +142,29 @@ protected function imageSource(
...
+	}
+
 	/**
 	 * Returns the full path without leading slash
 	 */
 	public function path(): string
 	{
 		// path to your own account
-		if ($this->model->isLoggedIn() === true) {
+		if ($this->isAccountView() === true) {
```
distantnative, line 167: Shouldn't this method always return the path to account if a user is logged in and not only when the current request is coming already from/for the account view?

afbora: I understand what you mean. If the path of the logged in user is `/users/***` instead of `/account`, what changes, what breaks? After all `/account` is just a static pattern that we manually set in the routes, and it is just a route and a view that is only used to edit the current user's profile.

afbora:
I don't think this PR will be very functional after v4. It didn't exactly solve the problem anyway.

PR #5090, opened 2023-03-04
Refactor `I18n::translate()`

### LINKED ISSUE #4869: Language snippet is not translated in user blueprint

### LINKED ISSUE #5086: Translation doesn't working in page blueprints `title` prop
This issue is regression of #5072 PR

```yaml
title: page.test
```

```php
<?php
// /site/languages/en.php
return [
    'code' => 'en',
    'default' => true,
    'direction' => 'ltr',
    'locale' => [
        'LC_ALL' => 'en_US'
    ],
    'name' => 'EN',
    'translations' => [
        'page.test' => 'Simple Page'
    ],
    'url' => NULL
];
````
```php
<?php
// /site/languages/de.php
return [
    'code' => 'de',
    'default' => false,
    'direction' => 'ltr',
    'locale' => [
        'LC_ALL' => 'de_DE'
    ],
    'name' => 'DE',
    'translations' => [
        'page.test' => 'Einfache Seite'
    ],
    'url' => NULL
];
```

distantnative:
- [x] Revert https://github.com/getkirby/kirby/pull/5072 which will solve #5086
- [x] Rebase PR
- [x] Unit tests all the way for each and every branch of `I18n::translate()`

This PR then should fix properly https://github.com/getkirby/kirby/issues/4869

### Breaking change
- If $fallback is an array and neither array $key nor array $fallback have an entry for the locale, the first $key array element will now be returned not the one from $fallback. If $fallback is a string, it will be considered with priority over both of these.

### DIFF
--- a/src/Cms/Blueprint.php
+++ b/src/Cms/Blueprint.php
@@ -77,7 +77,8 @@ class Blueprint
 		$props['name'] ??= 'default';
 
 		// normalize and translate the title
-		$props['title'] = $this->i18n($props['title'] ?? ucfirst($props['name']));
+		$props['title'] ??= ucfirst($props['name']);
+		$props['title']   = $this->i18n($props['title']);
 
 		// convert all shortcuts
 		$props = $this->convertFieldsToSections('main', $props);
@@ -221,14 +222,12 @@ class Blueprint
 
 	/**
 	 * Create a new blueprint for a model
-	 *
-	 * @param string $name
-	 * @param string|null $fallback
-	 * @param \Kirby\Cms\Model $model
-	 * @return static|null
 	 */
-	public static function factory(string $name, string $fallback = null, Model $model)
-	{
+	public static function factory(
+		string $name,
+		string $fallback = null,
+		Model $model
+	): static|null {
 		try {
 			$props = static::load($name);
 		} catch (Exception) {
@@ -318,7 +317,7 @@ class Blueprint
 	 */
 	protected function i18n($value, $fallback = null)
 	{
-		return I18n::translate($value, $fallback ?? $value);
+		return I18n::translate($value, $fallback) ?? $value;
 	}
 
 	/**
@@ -333,28 +332,21 @@ class Blueprint
 
 	/**
 	 * Loads a blueprint from file or array
-	 *
-	 * @param string $name
-	 * @return array
 	 */
 	public static function load(string $name): array
 	{
 		$props = static::find($name);
 
-		$normalize = function ($props) use ($name) {
-			// inject the filename as name if no name is set
-			$props['name'] ??= $name;
+		// inject the filename as name if no name is set
+		$props['name'] ??= $name;
 
-			// normalize the title
-			$title = $props['title'] ?? ucfirst($props['name']);
+		// normalize the title
+		$title = $props['title'] ?? ucfirst($props['name']);
 
-			// translate the title
-			$props['title'] = I18n::translate($title, $title);
+		// translate the title
+		$props['title'] = I18n::translate($title) ?? $title;
 
-			return $props;
-		};
-
-		return $normalize($props);
+		return $props;
 	}
 
 	/**
--- a/src/Toolkit/I18n.php
+++ b/src/Toolkit/I18n.php
@@ -25,55 +25,46 @@ class I18n
 
 	/**
 	 * Current locale
-	 *
-	 * @var string|\Closure
 	 */
-	public static $locale = 'en';
+	public static string|Closure|null $locale = 'en';
 
 	/**
 	 * All registered translations
-	 *
-	 * @var array
 	 */
-	public static $translations = [];
+	public static array $translations = [];
 
 	/**
-	 * The fallback locale or a
-	 * list of fallback locales
-	 *
-	 * @var string|array|\Closure|null
+	 * The fallback locale or a list of fallback locales
 	 */
-	public static $fallback = ['en'];
+	public static string|array|Closure|null $fallback = ['en'];
 
 	/**
 	 * Cache of `NumberFormatter` objects by locale
-	 *
-	 * @var array
 	 */
-	protected static $decimalsFormatters = [];
+	protected static array $decimalsFormatters = [];
 
 	/**
 	 * Returns the list of fallback locales
 	 */
 	public static function fallbacks(): array
 	{
-		if (
-			is_array(static::$fallback) === true ||
-			is_string(static::$fallback) === true
-		) {
-			return A::wrap(static::$fallback);
+		if (is_callable(static::$fallback) === true) {
+			static::$fallback = (static::$fallback)();
 		}
 
-		if (is_callable(static::$fallback) === true) {
-			return static::$fallback = A::wrap((static::$fallback)());
+		if (is_array(static::$fallback) === true) {
+			return static::$fallback;
+		}
+
+		if (is_string(static::$fallback) === true) {
+			return A::wrap(static::$fallback);
 		}
 
 		return static::$fallback = ['en'];
 	}
 
 	/**
-	 * Returns singular or plural
-	 * depending on the given number
+	 * Returns singular or plural depending on the given number
 	 *
 	 * @param bool $none If true, 'none' will be returned if the count is 0
 	 */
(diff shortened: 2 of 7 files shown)


afbora (commented):
@distantnative Unfortunately, it's still not fully works for fallback translation. Here steps:

- Add `page.test' translation for EN but not for DE
- Set DE language for logged user
- Try to create a new page
- See issue

afbora:
The unit test works, but a different scenario seems to be running when creating a page.

distantnative:
🤞 maybe this time

afbora (dismissed):
Works great 🎉

lukasbestle (commented):
Could you also add a test where the title is defined as an array like:

```yaml
title:
  en: Simple Page
  de: Einfache Seite
```

Apart from that the code LGTM.

lukasbestle:
Hm, isn't it actually weird that this works? I would have expected it to behave like in other places: Kirby should just take the translated string, not try to look it up in the global translations another time.

distantnative:
Our i18n system is weird. Ideally we would disallow any direct i18n keys but rather only support
```
title:
  *: my.i18n.key
```

But this would be a breaking change, so I think we have to go with supporting some quirky cases for now. What would be your solution?

distantnative:
It works because https://github.com/getkirby/kirby/blob/main/src/Cms/Blueprint.php#L352 runs first and then https://github.com/getkirby/kirby/blob/main/src/Cms/Blueprint.php#L80 again. But I think if we would change either, a lot breaks.

lukasbestle:
OK, that explains why it happens and I agree we shouldn't change that behavior. Too much of a breaking change.

However I would vote that we don't have a unit test for this as the use case is not realistic (why would I have a different translation key per language?). If we have the unit test, that will fail when we eventually fix the weird behavior.

What I rather suggested to test was a simple case of translated titles directly in the blueprint (see my example above).

lukasbestle (dismissed):

afbora:
Wait! I've found an issue 🐛

afbora:
> Der Bereich "pages" konnte nicht geladen werden: Kirby\Toolkit\I18n::translateFromTranslations(): Return value must be of type string, null returned

````yaml
title: 
    en: Simple Page
    tr: Basit Sayfa
````
### Reproduce steps:

1. Use above blueprint for `pages/default.yml` title
2. Change logged user language to DE (or a language that is not in the title array)
3. Go to any section that available for default page

**Comment by distantnative** [comment]
This is a pure nightmare. I think our I18n is so broken, trying to flexibly support a dozen of different use cases that I don't know if we'll ever find one solution for all problems. I doubt it by now - without any big refactoring that adds more explicitness instead of the all-giving flexibility 😣

**Comment by distantnative** [comment]
@afbora could you try in `I18n::translateFromTranslations()` to change the last return to something like


```php
return
	$fallback[$locale] ??
	$fallback[Str::before($locale, '_')] ??
	$translations['en'] ??
	$fallback['en'] ??
	reset($translations);
```

And we need to adapt the return types of that method. `reset()` could return `false`, the `::translate()` call even more types of what would then get returned.

src/Cms/Blueprint.php
```diff
@@ -345,6 +340,12 @@ public static function load(string $name): array
 		// inject the filename as name if no name is set
 		$props['name'] ??= $name;
 
+		// normalize the title
+		$title = $props['title'] ?? ucfirst($props['name']);
+
+		// translate the title
+		$props['title'] = I18n::translate($title) ?? $title;
```
lukasbestle, line 347: In other places we use `I18n::translate($title, $title)` to fix exactly that issue. Because then the `translate` method handles it by taking the first value from the array. As far as I remember we had that before here and you switched to `??`. What was the reason for that again?

distantnative: Haha yes that's the beginning of the whole journey. When we pass the same array as fallback, we ended up with the problem that i18n keys would not use the default's language value from the global translations strings as fallback.

distantnative:
The most productive probably would be to clear our mind and start fresh with a unit test suite for I18n that really tries to cover all our weird cases. And then go back and look at our code how to fix/make it happen. Even if that delays 3.9.2 quite a bit potentially.

lukasbestle:
Or we revert https://github.com/getkirby/kirby/pull/5072 for now and move the fix for the original issue to one of the next releases.

afbora:
> @afbora could you try in `I18n::translateFromTranslations()` to change the last return to something like

@distantnative Same :(

distantnative:
> Or we revert https://github.com/getkirby/kirby/pull/5072 for now and move the fix for the original issue to one of the next releases.

@lukasbestle at this point we are trading bug for bug, not sure which one is more critical

EDIT: not saying I'm against it, just resigning quite a bit right now.

distantnative:
While I like how sleek `??=` makes the method signature, it could be a little ditch in performance but even more less helpful for code coverage (one test will show as covering all). Probably will need to turn it back into a lot of `if ($result = …) return $result` statements.

distantnative:
@lukasbestle @afbora 
I think the last part of `I18n::translate()` should actually be:

```php
// at the end, if a string was provided as fallback
if (is_string($fallback) === true) {
	return $fallback;
}

// otherwise the first array element of the input
if (is_array($key) === true) {
	return reset($key);
}
// or the first array element of the fallback
if (is_array($fallback) === true) {
	return reset($fallback);
}

return null;
```

However, this breaks an existing test:

```php
I18n::translate(['de' => 'Speichern'], ['first' => 'Fallback'])
```
Before the result would be `Fallback`, with the suggested it'd be `Speichern`. So this would technically be a breaking change, but in my mind this makes more sense then (new unit tests):


```php
// use the fallback value if there's no english translation
// and the fallback is a string
$this->assertSame(
	'Fallback',
	I18n::translate(['de' => 'Speichern'], 'Fallback')
);

// use the first value if there's no english translation
$this->assertSame(
	'Speichern',
	I18n::translate(['de' => 'Speichern'], ['first' => 'Fallback'])
);
```

What do you think?

lukasbestle:
That would be the same as passing `($title, $title)` as arguments though, right?

distantnative:
@lukasbestle Only if $title isn't an array. Also see my last bigger commit, I rewrote the translate method as the old one had some flaws that were hard to spot. Will push some more inline comments in a few minutes to make it much clearer.

Would still advise to revert the old change for 3.9.2 and then include this better refactoring/fixing in 3.9.3.

distantnative:
@afbora case for you detective: why is this unit tests https://github.com/getkirby/kirby/pull/5090/files#diff-9ca61ecf734bfac6fc1c7e48876140e7510a8d55977af0ffb23c15bb99048e0dR160-R167 not considered to cover https://app.codecov.io/gh/getkirby/kirby/commit/313f1fdfd2dea3e6b1a356292124e2854b535c1b/blob/src/Toolkit/I18n.php#L161 ?

afbora:
@distantnative Because the `static::translation()` method already checks the short locale. 
https://github.com/getkirby/kirby/blob/fix/blueprint-title/src/Toolkit/I18n.php#L267-L270

Therefore, since it is checked in the previous line, it never passes to the line you mentioned.
https://github.com/getkirby/kirby/blob/fix/blueprint-title/src/Toolkit/I18n.php#L158

distantnative:
🤦‍♂️ thank you!

lukasbestle:
Yeah, to be honest the setup is too complex for me to fully understand it at the moment. Fully agree we should revert the old change and take a step back.

afbora (commented):
I've tested all scenarios on page and user blueprint titles and seems works great 🎉 Need more eyes to approve the PR 👀

src/Toolkit/I18n.php
```diff
@@ -98,107 +89,158 @@ public static function formatNumber(int|float $number, string $locale = null): s
 	}
 
 	/**
-	 * Returns the locale code
+	 * Returns thecurrent locale code
```
afbora:
```suggestion
	 * Returns the current locale code
```

src/Toolkit/I18n.php
```diff
@@ -98,107 +89,158 @@ public static function formatNumber(int|float $number, string $locale = null): s
...
+
+		// if the $fallback parameter is an array, we can assume
+		// that it's also an array with entries per locale:
+		// check with long and short locale if we find a matching entry
+		if ($result = $fallback[$locale] ?? null) {
+			return $result;
+		}
+		if ($result = $fallback[$shortLocale] ?? null) {
+			return $result;
+		}
```
afbora, line 214:
```suggestion
		if ($result = $fallback[$locale] ?? $fallback[$shortLocale] ?? null) {
			return $result;
		}
```

afbora (commented):
The above changes are not critical, just optional. Also I didn't see a `title` test for `UserBlueprint`, it would be nice to add unit test for that if I'm not mistaken. I can take care if you wish.

src/Toolkit/I18n.php
distantnative: This one I don't wanna do. I kept every possibility now as its own if Statement because that way test coverage really can show us if we indeed have each and every outcome covered by tests.

I really love chaining things with ?? But for that reason I removed it a lot in this method.

distantnative:
@afbora That would be so helpful!

src/Toolkit/I18n.php
afbora: You're right about unit test coverage 💯

lukasbestle (commented):
I still can't fully wrap my head around all edge cases, but I think the new logic looks sensible. Good job on the code comments, they really help to understand the order of the new logic. 👍

I know that the `I18n::translate($value, $value)` pattern for fallbacks is used quite a lot, both in our code and in plugins. So it could be that the breaking change affects some of those places directly. Do you think it's still safe to merge this in a minor release?

distantnative:
Will merge manually in v4 develop branch

PR #5091, opened 2023-03-05
Blueprints: fix name for system default files

distantnative:
### Fixes
- Default blueprints for pages and files that are bundled with Kirby now correctly provide their names as `default`

## Ready?

### For review team

### DIFF
--- a/config/blueprints/blocks/code.yml
+++ b/config/blueprints/blocks/code.yml
@@ -1,4 +1,4 @@
-name: Code
+title: Code
 icon: code
 fields:
   code:
--- a/config/blueprints/blocks/heading.yml
+++ b/config/blueprints/blocks/heading.yml
@@ -1,3 +1,4 @@
+title: Heading
 icon: title
 fields:
   text:
--- a/config/blueprints/blocks/image.yml
+++ b/config/blueprints/blocks/image.yml
@@ -1,4 +1,4 @@
-name: Image
+title: Image
 icon: image
 fields:
   image:
--- a/config/blueprints/blocks/quote.yml
+++ b/config/blueprints/blocks/quote.yml
@@ -1,4 +1,4 @@
-name: Quote
+title: Quote
 icon: quote
 fields:
   text:
--- a/config/blueprints/blocks/table.yml
+++ b/config/blueprints/blocks/table.yml
@@ -1,4 +1,4 @@
-name: Table
+title: Table
 icon: menu
 fields:
   rows:
--- a/config/blueprints/blocks/text.yml
+++ b/config/blueprints/blocks/text.yml
@@ -1,4 +1,4 @@
-name: Text
+title: Text
 icon: text
 fields:
   text:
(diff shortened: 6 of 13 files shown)


lukasbestle (commented):
Two questions:

- Where does Kirby try to load the files from the `config/blueprints` directory? Do we need to remove that logic now that the files no longer exist?
- Do we need localization for the fallback titles?

distantnative:
Here: https://github.com/getkirby/kirby/blob/main/src/Cms/Core.php#L109-L117
Which made me spot that there is also a default site blueprint: https://github.com/getkirby/kirby/blob/main/config/blueprints/site.yml

So maybe we even leave them in the files - in the end the default block blueprints aren't much easier to spot either. But add comments in the blueprint method that makes it easy to understand that `extensions` can also mean from our system folder in this case.

On i18n, I wouldn't want to tackle that here. It complicates this topic further. And we aren't removing any.

lukasbestle:
> So maybe we even leave them in the files - in the end the default block blueprints aren't much easier to spot either. But add comments in the blueprint method that makes it easy to understand that `extensions` can also mean from our system folder in this case.

I agree.

> On i18n, I wouldn't want to tackle that here. It complicates this topic further. And we aren't removing any.

Adding translation strings for `page => Page` and `file => File` seems like such a low-hanging fruit to me now that we are touching this.

lukasbestle (approved):
I think the changes make sense. Overriding the `name` is weird and I don't see a reason why we needed to do that.

Regarding I18n: I'm fine without it. We can still add strings later.

afbora:
Maybe I'm tired at the moment, or maybe I missed a conversation, but can someone tell me the difference between the blueprints of the blocks in these two different directories? `config/blocks/***/***.yml` vs `config/blueprints/blocks/***.yml`

- https://github.com/getkirby/kirby/tree/fix/blueprint-defaults/config/blocks
- https://github.com/getkirby/kirby/tree/fix/blueprint-defaults/config/blueprints/blocks

bastianallgeier:
> Maybe I'm tired at the moment, or maybe I missed a conversation, but can someone tell me the difference between the blueprints of the blocks in these two different directories? `config/blocks/***/***.yml` vs `config/blueprints/blocks/***.yml`
> 
> * https://github.com/getkirby/kirby/tree/fix/blueprint-defaults/config/blocks
> * https://github.com/getkirby/kirby/tree/fix/blueprint-defaults/config/blueprints/blocks

Oh my, I actually think that the ones in config/blueprints/blocks are just left-overs. 🙈

afbora:
And this PR have changes on `config/blueprints/blocks/***.yml` 🙈

bastianallgeier:
I think it would be best to remove them entirely. I just checked and the Core class loads block blueprints from config/blocks. There's no reference to the other blueprint files.

distantnative:
@bastianallgeier Agreed to then remove `config/blueprints/blocks` completely. We still should add better inline comments to point out that some blueprints could come not just from`site/blueprints` or plugins but `config/blocks` too.

What do we do with the default blueprints for page, file and site? Move them to hardcoded PHP arrays in `Blueprint::find()` instead?

distantnative:
Will open a new PR where I remove that folder

Issue #5085, opened 2023-02-28
Finalize "`Dir::index()`: fix `$ignore` parameter"


lukasbestle:
The PR https://github.com/getkirby/kirby/pull/5069 still has open questions. I think we should resolve those until the final release of 3.9.2 because any future change would be a breaking change:

**No-brainers:**

- [x] Check if `$ignore` is an array in line 154.
- [x] Deprecate the `null` value for `$ignore`.
- [x] Add a `@todo` comment to remove support for `$ignore = null` in a major release.

**Still to be discussed:**

- [x] Should the `$ignore` param be consistent with `Dir::read()` (ignore on each path segment level) or with `Dir::copy()` (full paths)?
  - I still feel like it should ignore on the full path because of the recursive mode that `Dir::read()` does not have but that `Dir::copy()` does have.
- [x] Do we want to support `$ignore = false` for consistency with `Dir::copy()` (ignores the global ignore list)?

lukasbestle:
@distantnative How do we proceed with this? TBH I feel like we should also revert #5069 and include it in a later release. If we release it as-is, we will not have the opportunity to get the feature right later without a breaking change.

distantnative:
Tbh I don't quite understand what happened with the discussion etc when it got merged. Probably something rather for @bastianallgeier.

lukasbestle:
I think it just got approved and merged without having the discussion at all.

bastianallgeier:
I misunderstood the discussion. I thought that all the open questions could/should be retrofitted other repo in because they would already introduce breaking changes. I will take care of the no-brainers today and also check the remaining discussion points again. Sorry for merging it too quickly.

bastianallgeier:
Ok, I had another look at it and I agree with @lukasbestle. The `Dir::index()` method is way closer to `Dir::copy()`. I wonder why `$recursive` is false by default. I think it should actually be true. But well, that's another breaking change. 

I aligned the two methods now and I think it's the right way forward. What do you think? 

Ideally it should be possible to pass wildcards at some point, but this is another future project.

bastianallgeier:
✅

PR #5094, opened 2023-03-07
Align $ignore param in Dir::index with Dir::copy

bastianallgeier:
See: https://github.com/getkirby/kirby/issues/5085

### DIFF
--- a/src/Filesystem/Dir.php
+++ b/src/Filesystem/Dir.php
@@ -57,7 +57,7 @@ class Dir
 		string $dir,
 		string $target,
 		bool $recursive = true,
-		array|bool $ignore = []
+		array|false $ignore = []
 	): bool {
 		if (is_dir($dir) === false) {
 			throw new Exception('The directory "' . $dir . '" does not exist');
@@ -141,18 +141,19 @@ class Dir
 	 * Read the directory and all subdirectories
 	 *
 	 * @todo Remove support for `$ignore = null` in a major release
-	 * @param array|null $ignore Passing null is deprecated.
+	 * @param array|false|null $ignore Array of absolut file paths;
+	 *                                 `false` to disable `Dir::$ignore` list
+	 *                                 (passing null is deprecated)
 	 */
 	public static function index(
 		string $dir,
 		bool $recursive = false,
-		array|bool|null $ignore = [],
+		array|false|null $ignore = [],
 		string $path = null
 	): array {
 		$result = [];
 		$dir    = realpath($dir);
 		$items  = static::read($dir, $ignore === false ? [] : null);
-		$ignore = (array)$ignore;
 
 		foreach ($items as $item) {
 			$root = $dir . '/' . $item;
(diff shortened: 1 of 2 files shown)


src/Filesystem/Dir.php
```diff
@@ -139,23 +139,31 @@ public static function files(
...
 		bool $recursive = false,
-		array|null $ignore = [],
+		array|bool|null $ignore = [],
 		string $path = null
 	): array {
 		$result = [];
 		$dir    = realpath($dir);
-		$items  = static::read($dir);
+		$items  = static::read($dir, $ignore === false ? [] : null);
+		$ignore = (array)$ignore;
```
distantnative: If `$ignore` is `false` this will create `[false]`

tests/Filesystem/DirTest.php
```diff
@@ -231,7 +231,11 @@ public function testIndexIgnore()
 			'sub/sub/c.txt'
 		];
 
-		$this->assertSame($expected, Dir::index($dir, true, ['a.txt']));
+		$this->assertSame($expected, Dir::index($dir, true, [
+			$this->tmp . '/a.txt',
+			$this->tmp . '/sub/a.txt',
+			$this->tmp . '/sub/sub/a.txt'
+		]));
```
distantnative, line 265: Can we also add a unit test for `$ignore = false`?

src/Filesystem/Dir.php
```diff
@@ -139,23 +139,31 @@ public static function files(
...
 	 * Read the directory and all subdirectories
+	 *
+	 * @todo Remove support for `$ignore = null` in a major release
+	 * @param array|null $ignore Passing null is deprecated.
 	 */
 	public static function index(
 		string $dir,
 		bool $recursive = false,
-		array|null $ignore = [],
+		array|bool|null $ignore = [],
```
distantnative: I think PHP actually allows to type hint for `array|false|null` - otherwise we'd also need to handle `true`.

src/Filesystem/Dir.php
lukasbestle: I think we can remove this line completely as we check for `is_array()` below anyway.

src/Filesystem/Dir.php
```diff
@@ -139,23 +139,31 @@ public static function files(
 
 	/**
 	 * Read the directory and all subdirectories
+	 *
+	 * @todo Remove support for `$ignore = null` in a major release
+	 * @param array|null $ignore Passing null is deprecated.
```
lukasbestle:
```suggestion
	 * @param array|false|null $ignore Passing null is deprecated.
```

lukasbestle (commented):
I agree with Nico's review comments. Apart from those the changes look great! :+1:

src/Filesystem/Dir.php
distantnative: We should probably change this in `Dir::copy()` right away as well

bastianallgeier:
Thanks for the fixes. I think you still need to approve it, because I can't :)

distantnative (approved):

PR #5099, opened 2023-03-12
Never escape Unicode characters in blocks/layout fields

lukasbestle:
### Enhancements

- Unicode characters and slashes are no longer escaped in blocks and layout fields, even if the `pretty: true` mode is not used https://kirby.nolt.io/518

### Breaking changes

None

## Ready?

- [x] Unit tests for fixed bug/feature (will add if accepted)

### For review team


distantnative:
Looks good to me but should we add a test that checks for no escaping even when without `pretty`?

distantnative (approved):

PR #5105, opened 2023-03-15
Fix wrong sort number when using limit in files section

### LINKED ISSUE #5101: Sort order is random when uploading files in panel
## Description

I have a page with a files section (at the moment about 1000 images). When I upload new files, they get a somewhat random sort number assigned—one that already exists (!). It seems to have something to do with the pagination / `limit` setting in the blueprint, since it’s often the number set as `limit` + 1. But it also depends on the current pagination page of the files section, so e. g. when I’m on the last page the file(s) get a different sort number (a lower one, to be precise).

**Expected behavior**  

The files being added to the beginning or end of the files section.

## To reproduce

1. Create a blueprint with a files section that is `sortable`
2. (Add some / 1000 files)
3. Upload a new file

## Your setup
**Kirby Version**  

3.9.2

 - Device:  MacBook Pro 16" 
 - OS:  macOS 13.1
 - Browser:  Safari
 - Version:  16.2

afbora:
## This PR …

This issue is fixed with using `total` computed data instead counting paginated data.

### Fixes
- Fix wrong sort number when using limit in files section #5101

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -178,6 +178,9 @@ return [
 				'max'        => $max,
 				'api'        => $this->parent->apiUrl(true) . '/files',
 				'attributes' => array_filter([
+					// TODO: an edge issue that needs to be solved:
+					//		 if multiple users load the same section at the same time
+					// 		 and upload a file, uploaded files have the same sort number
 					'sort'     => $this->sortable === true ? $this->total + 1 : null,
 					'template' => $template
 				])
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -162,10 +162,9 @@ return [
 			}
 
 			// count all uploaded files
-			$total = count($this->data);
-			$max   = $this->max ? $this->max - $total : null;
+			$max = $this->max ? $this->max - $this->total : null;
 
-			if ($this->max && $total === $this->max - 1) {
+			if ($this->max && $this->total === $this->max - 1) {
 				$multiple = false;
 			} else {
 				$multiple = true;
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
Ah, that's a really simple bug, but hard to spot. 👍
Could you add a test for this to `FilesSectionTest`?

Reading the code I'm also curious: What happens if another user uploads a file after the Panel view with the section was loaded? Won't the sort number be outdated then?

afbora:
> Could you add a test for this to FilesSectionTest?

AFAIK Unfortunately, we are unable to perform file upload tests since `move_uploaded_file()` not working with unit test.

> What happens if another user uploads a file after the Panel view with the section was loaded? Won't the sort number be outdated then?

Yes, both user will have same `sort` number. But I think this is a different issue and it seems that the files need to be recounted on the backend side to fix it.

lukasbestle:
> AFAIK Unfortunately, we are unable to perform file upload tests since `move_uploaded_file()` not working with unit test.

I don't think we need to test actually uploading a file. Just the value of the `upload` prop of the section. We already call it in the `testMax()` method, but there don't seem to be tests specific to the `upload` prop.

> > What happens if another user uploads a file after the Panel view with the section was loaded? Won't the sort number be outdated then?
> 
> Yes, both user will have same `sort` number. But I think this is a different issue and it seems that the files need to be recounted on the backend side to fix it.

I agree, it's nothing we can solve immediately. Maybe we could add a `TODO` comment to move the generation of the sorting ID to the backend in a later version. Until then, your fix already improves the behavior a lot.

lukasbestle (approved):
👍

PR #5107, opened 2023-03-16
Proposal to remove lowercase custom field methods

lukaskleinschmidt:
While trying out some stuff for automatically creating type hints for Kirby classes I ran into the problem that currently custom field methods are stored with lowercase keys. This makes it impossible to create the proper type hints for those as I cant reverse the keys back to the initial value `topages > toPages`. Although I could get the intended names directly from `config/methods.php` this won't work for plugins.

### Proposal
1. Store custom field methods with their intended names
2. For backwards compatibility add any required aliases 9417c9971ac83ad72368560d6bb82b135929884f

### Breaking changes
None if the aliases get created

## Ready?

### For review team


lukaskleinschmidt:
Ok I was proven wrong 😄
Right now there might be some breaking changes.

Edit: Should be fixable but perhaps we should discuss the basic idea first…

lukasbestle:
Right now Kirby treats field methods completely case-insensitively. So you could not just call them as `toPages` or `topages` but also as `tOpAgEs`.

The question is why we implemented it like this in the first place. Other methods in Kirby are not case-insensitive, so why are field methods?

If it turns out to be a leftover that is no longer relevant, I'd be in favor of removing the case-insensitive handling entirely. This would be a breaking-change, so only possible in a major version.

bastianallgeier:
That's mostly because any PHP method is case insensitive. That's at least the reason I remember

lukaskleinschmidt:
Thats true. Also for me I found another way to get the intended method names.

Just a few thoughts
- The only difference to classes that use the `HasMethods` trait is the use of aliases
- The `HasMethods` trait does the method calls case sensitive
- Field methods are documented case sensitive
- I no longer depend on the stored method names for my use case 😅

So not sure if it's worth streamlining or not.

lukasbestle:
I had always thought PHP method names are case-sensitive, at least in recent versions. 🤯
But of course you are right.

I think there are arguments for both (keeping the case-insensitive behavior or making field methods case-sensitive). If we built field methods as a new feature, I feel like we might end up with a case-sensitive implementation. However I'm worried about the breaking change when we change the existing behavior.

So I think given that there is currently no other feature, bug fix or use case that depends on a case-sensitive behavior, it isn't worth the hassle for our users to change this behavior now. We will pick this up in the future if it becomes necessary.

PR #5117, opened 2023-03-17
Stale bot only for when actions needed

distantnative:
Apply stale bot on issues and PRs only when certain input/actions are needed from the user.
If they don't happen for 60 days, gets marked as stale and after 15 additional days closed.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -26,7 +26,7 @@ jobs:
           stale-issue-message: "This issue has been marked as stale because it requires further input but has not seen activity in the past months. This is for us to prioritize issues that are still relevant and actionable. It will be closed if no further activity occurs within the next 15 days. If this issue is still relevant to you, please help us in gathering the necessary input."
 
           # Comment on the staled PRs
-          stale-pr-message: "This PR has been marked as stale because it requires further changes but has not seen activity in the past months. This is for us to prioritize PRs that can be reviewed and merged. It will be closed if no further activity occurs within the next 15 days. If you still have interest in this PR, please help us finalizing it."
+          stale-pr-message: "This PR has been marked as stale because it requires further changes but has not seen activity in the past months. This is for us to prioritize PRs that can be reviewed and merged. It will be closed if no further activity occurs within the next 15 days. If you still have interest in this PR, please help us finalizing it. Please let us know in case you are stuck on the required changes."
 
           # Label to apply on staled issues
           stale-issue-label: "type: stale 💤"
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -14,22 +14,31 @@ jobs:
       - uses: actions/stale@6f05e4244c9a0b2ed3401882b05d701dd0a7289b # pin@v7
         with:
           # Idle number of days before marking issues stale, set `-1` to disable
-          days-before-issue-stale: 180
+          days-before-issue-stale: 75
 
           # Idle number of days before marking issues stale, set `-1` to disable
-          days-before-pr-stale: -1
+          days-before-pr-stale: 75
 
           # Idle number of days before closing stale issues/PRs
-          days-before-close: 14
+          days-before-close: 21
 
           # Comment on the staled issues
-          stale-issue-message: "This issue has been automatically marked as stale because it has not had recent activity. This is for us to prioritize issues that are still relevant to our community. It will be closed if no further activity occurs within the next 14 days. If this issue is still relevant to you, please leave a comment."
+          stale-issue-message: "This issue has been marked as stale because it requires further input but has not seen activity in the past months. This is for us to prioritize issues that are still relevant and actionable. It will be closed if no further activity occurs within the next 21 days. If this issue is still relevant to you, please help us in gathering the necessary input."
+
+          # Comment on the staled PRs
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -14,22 +14,31 @@ jobs:
...
           stale-issue-label: "type: stale 💤"
 
+          # Label to apply on staled PRs
+          stale-pr-label: "type: stale 💤"
+
           # Reason to use when closing issues
           close-issue-reason: not_planned
 
+          # Labels to check for stale issues/PRs
+          any-of-labels: "needs: changes 🔁,"
```
afbora:
```suggestion
          any-of-pr-labels: "needs: changes 🔁"
          any-of-issue-labels: "needs: discussion 🗣,needs: decision 🗳,needs: information ❓,needs: replication 🔬"
```

afbora: Maybe just enough `needs: information ❓,needs: replication 🔬` for the `any-of-issue-labels`

afbora (commented):
Also we can use `exempt-milestones`, `remove-issue-stale-when-updated` labels.

.github/workflows/stale.yml
distantnative: Adapted some of it, wouldn't include discussion/decision as that's an action we as team need to do, not the user.

distantnative:
`remove-stale-when-updated` is already true by default

.github/workflows/stale.yml
```diff
@@ -38,7 +38,9 @@ jobs:
           close-issue-reason: not_planned
 
           # Labels to check for stale issues/PRs
-          any-of-labels: "needs: changes 🔁,"
+          any-of-pr-labels: "needs: changes 🔁,needs: docs 📝,needs: tests  🧪"
```
afbora: - `needs: docs`: PR is almost ready but needs to be documented 
- `needs: tests`: PR is almost ready but unit tests are needed

So why are we close PR that is almost done? It would be a better decision to close PRs that have not been decided and discussed for a long time than those with missing unit testing.

distantnative: But are they really almost ready if months pass without them getting finalised?

distantnative: >  It would be a better decision to close PRs that have not been decided and discussed for a long time than those with missing unit testing.

I think when the required action is by us the team (discussion, decision), it shouldn't ever be a bot closing the PR but only us ourselves. We shouldn't shove that responsibility to a bot.

afbora: You are right about responsibility 👍 Then I'm in favor of removing `needs: docs` and `needs: tests` labels as well from PR labels. I think these are almost complete PRs.

distantnative: Not sure...if they are around for two months without getting finalised and then after the warning by stalest nobody finalises them in the next 15 days, I think they're far from almost complete or even easy to complete. Otherwise one would manage to complete them within 60 + 15 days, right?

afbora: If it has not been completed for a long time (for needs docs and test), it would be better for the team to decide this rather than throwing whole the PR in the trash with a bot.

afbora: I don't understand; If we don't put the responsibility for undecided or undiscussed uncertain PRs on the bot, why do we leave the automatic close of almost completed PRs to the bot? 🤔

distantnative: You're right

afbora (dismissed):

.github/workflows/stale.yml
```diff
@@ -14,22 +14,33 @@ jobs:
...
+          days-before-close: 15
 
           # Comment on the staled issues
-          stale-issue-message: "This issue has been automatically marked as stale because it has not had recent activity. This is for us to prioritize issues that are still relevant to our community. It will be closed if no further activity occurs within the next 14 days. If this issue is still relevant to you, please leave a comment."
+          stale-issue-message: "This issue has been marked as stale because it requires further input but has not seen activity in the past months. This is for us to prioritize issues that are still relevant and actionable. It will be closed if no further activity occurs within the next 15 days. If this issue is still relevant to you, please help us in gathering the necessary input."
+
+          # Comment on the staled PRs
+          stale-pr-message: "This PR has been marked as stale because it requires further changes but has not seen activity in the past months. This is for us to prioritize PRs that can be reviewed and merged. It will be closed if no further activity occurs within the next 15 days. If you still have interest in this PR, please help us finalizing it."
```
lukasbestle:
```suggestion
          stale-pr-message: "This PR has been marked as stale because it requires further changes but has not seen activity in the past months. This is for us to prioritize PRs that can be reviewed and merged. It will be closed if no further activity occurs within the next 15 days. If you still have interest in this PR, please help us finalizing it. Please let us know in case you are stuck on the required changes."
```

lukasbestle (commented):
I like the changes a lot. For PRs with missing tests/docs we could still manually explain that we cannot merge the PR without them and attach the `stale` label manually so that the bot can close the issue if there is no response within 15 days. So basically the best of both worlds: Human decision without the need to keep the PR on the radar to close it.

One suggestion to make the bot a bit more friendly:

afbora (approved):

lukasbestle:
Removed the stale label from every issue that currently had it to prevent the bot from closing these issues.

PR #5119, opened 2023-03-20
Fix date changes with adding new interpret formats

### LINKED ISSUE #5118: Date field with display property changes value while editing
## Description
Date fields with a display property that does not include a day e.g `display: MM-YYYY` change value while editing. In my case it always changes to 08-2024.

## To reproduce

1. Add the following field to a blueprint
```php
date:
    type: date
    display: MM YYYY
```
2. Select a date using the date picker
3. Focus the date field text input
4. Blur the date field text input

## Setup
**Kirby 3.9.2**  

**System**
 - OS: macOS
 - Browser: Chrome
 - Version: 109

afbora:
## This PR …

@distantnative I've fixed the date changes with adding new interpret formats. But I'm not sure about the solution 🙈 Feel free to close the PR .

### Fixes
- Date field with display property changes value while editing #5118

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/libraries/dayjs-interpret.js
+++ b/panel/src/libraries/dayjs-interpret.js
@@ -24,8 +24,11 @@ export default (option, Dayjs, dayjs) => {
 				"MMM D YYYY": false,
 				"MMM DD YY": false,
 				"MMM D YY": false,
+				"MMM YYYY": true,
 				"MMM DD": false,
 				"MMM D": false,
+				"MM YYYY": true,
+				"M YYYY": true,
 
 				"DD MMMM YYYY": false,
 				"DD MMMM YY": false,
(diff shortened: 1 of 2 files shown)


distantnative:
We should for sure add unit tests before attempting to merge this

afbora:
@distantnative was this PR risky for 3.9.3?

distantnative:
@afbora I think it was rather missed because the status was still draft

afbora:
Let's make it ready if you think the PR has correct solution.

lukasbestle (approved):
I think the consideration here needs to be whether the formats are "obvious" and unambiguous enough.

For `MMM YYYY` that's definitely the case.

For `MM YYYY` and `M YYYY`, the month number and year is certainly the most likely interpretation if one enters `02 2022` for example. It could also be the week of the year instead of the month though (so in that case the value would resolve to `2022-01-10`, the first day of the second week of 2022 – at least in Europe with Monday as the first day of the week 😅).

However we currently only support year, month and day as date tokens, right? If that's the case and will stay like that, I think we can safely merge this PR.

PR #5124, opened 2023-03-25
`F::move()`: Support non-atomic moves

### LINKED ISSUE #5024: Error message on upload "Operation not permitted"
## Description
In some environments, after uploading files through the panel the following error message is shown:

```
myfilename.jpg

rename(/tmp/63d1538df3338.20221110myfilename.jpg,/home/path/to/web/root/content/home/myfilename.jpg) [<a href='https://secure.php.net/manual/en/function.rename.php'>function.rename.php</a>]: Operation not permitted
```
The file is however correctly uploaded, implying the fact that this is not a permissions issue. The error message and `400` status of the request also prevent a `files` field to automatically select the *just uploaded* file. 

**Expected behavior**  
No error message should be shown if the upload has succeeded. 

**Screenshots**  
image

## To reproduce

This issue is only visible in certain environments. See *Additional Context*

1. Login to the panel
2. Upload or replace a file

## Your setup
**Kirby Version**  
Kirby 3.9.0

**Console output**  
no stack trace is included, even with `'debug' => true`

**Server system**
FreeBSD 12.4-RELEASE releng/12.4-n235809-f2877ac4fb03 HOSTPOINT  amd64

**Additional context**  
Searching the internets, this corresponds:
https://stackoverflow.com/questions/45692433/php-rename-from-tmp-throws-warning-although-actually-works
The answer there mentions that `rename` throws a warning when moving files across file systems. It suggests using `copy` + `unlink` instead.
Since 3.9.0-rc-2 introduced ["Support moving source files on create and replace"](https://github.com/getkirby/kirby/commit/14f2e69bd68f8970fa504f7ac841d02b42503601) this might be something. Before 3.9 files were simply copied. 
This also explains why the error only shows up in certain environments (`/tmp` and `/content` must be on different file systems / mounts).

lukasbestle:
### Fixes

- Uploading files no longer fails with the error "Operation not permitted" when the temporary directory is on another filesystem #5024
- `F::move()` now detects if the source and target are on different filesystems; it then copies the file and deletes the source on success instead of atomically moving the file #5124

## Ready?

- [x] Unit tests for fixed bug/feature (test added, but skipped in CI because tests are run inside a container; this is why code coverage only displays 50%)

### For review team

### DIFF
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -521,12 +521,26 @@ class F
 			Dir::make($directory, true);
 		}
 
-		// actually move the file
-		if (rename($oldRoot, $newRoot) !== true) {
-			return false;
+		// atomically moving the file will only work if
+		// source and target are on the same filesystem
+		if (stat($oldRoot)['dev'] === stat($directory)['dev']) {
+			// same filesystem, we can move the file
+			return rename($oldRoot, $newRoot) === true;
 		}
 
-		return true;
+		// not the same filesystem; we need to copy
+		// the file and unlink the source afterwards
+		if (copy($oldRoot, $newRoot) === true) {
+			return unlink($oldRoot) === true;
+		}
+
+		// copying failed, ensure the new root isn't there
+		// (e.g. if the file could be created but there's no
+		// more remaining disk space to write its contents)
+		// @codeCoverageIgnoreStart
+		static::remove($newRoot);
+		return false;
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
I haven't tested it, but the code comments and tests look good to me. You can merge it unless you want another look from @distantnative or @afbora

afbora (approved):
Works great in my tests on local and live server 👍

PR #5127, opened 2023-03-28
Block label: strip HTML

distantnative:
### Fixes
- HTML gets now stripped from Block labels https://kirby.nolt.io/421

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Blocks/BlockTitle.vue
+++ b/panel/src/components/Forms/Blocks/BlockTitle.vue
@@ -37,7 +37,8 @@ export default {
 				this.fieldset.label,
 				this.content
 			);
-			return label === "…" ? false : label;
+
+			return label === "…" ? false : this.$helper.string.stripHTML(label);
 		},
 		name() {
 			return this.fieldset.name;


lukasbestle (approved):
I was thinking whether there would be use cases where stripping the tags would result in an unexpected or even broken result. But given that `stripHTML()` only removes actual tags that are both opened and closed, the only exception I can think of would be Markdown links like `<https://getkirby.com>`.

So I think this is already much better than the status quo.

distantnative:
> But given that stripHTML() only removes actual tags that are both opened and closed

Not sure I can follow you here. `stripHTML` also would strip a single opening tag. But since our output for the label is not as HTML but escaped by Vue via `{{ label }}` I think any result of this will be still better than currently.

lukasbestle:
> Not sure I can follow you here. `stripHTML` also would strip a single opening tag.

I didn't mean a closing HTML tag but a tag where there must be both a `<`, one or more other character and a `>`.

Not talking about security, just about potential edge cases where we might strip too much. Markdown links are one such edge case, but as I wrote I think stripping is still better than not stripping.

Issue #5120, opened 2023-03-20
Self overwriting custom methods


lukaskleinschmidt:
## Description
I came across an issue for some classes which extend a class using hte `HasMethods` trait.
The `Kirby\Cms\Layouts` or `Kirby\Cms\Blocks` classes for example extend the `Kirby\Cms\Collection` class (at least at the end).

This makes all declared `Collection` methods also available in the extended classes. However you could run into issues (or at least unexpected behaviour) when you extend those classes and use the same name for the methods.

```php
App::plugin('my/plugin', [
    'blocksMethods' => [
        'test' => function () {
            return 'blocks method';
        },
    ],
    'layoutsMethods' => [
        'test' => function () {
            return 'layouts method';
        },
    ],
]);
```
```php
$collection = new Collection([]);
var_dump($collection->test()); // layouts method

$collection = new Layouts([]);
var_dump($collection->test()); // layouts method

$collection = new Blocks([]);
var_dump($collection->test()); // layouts method
```
If you would define `public static $methods = [];` in the `Blocks` or `Layouts` classess this would fix the "bleeding" but you would loose the inherited custom methods from the `Collection` class itself.

Fortunately the `$methods` are already static so you could actually kinda extend them like this without having the problem of possible method overwrites.
```php
trait HasMethods
{
    public static array $methods = [];

    public static function getMethods(): array
    {
        $methods = static::$methods;

        if ($parent = get_parent_class(static::class)) {
            $methods = array_merge($parent::getMethods(), $methods);
        }

        return $methods;
    }
}
```
Or maybe prefix the methods for the actually targeted class?
Let me know what you think and I might be able to help prepare a PR for this.

Those two classes where just an example I was currently working with.
So there could be more classes with the same problem.

## Your setup
**Kirby Version**  
3.9.2

lukasbestle:
I like the idea to query the parent class or even all parent classes to allow inheritance. That would definitely allow a much more deterministic behavior. Because the inheritance chains are typically not that long, I don't think there would be a huge performance impact either.

PR #5128, opened 2023-03-29
Custom method inheritance

lukaskleinschmidt:
## This PR fixes the problems outlined here #5120 

I had another look and the trait already does something like this in the [getMethod()](https://github.com/getkirby/kirby/blob/main/src/Cms/HasMethods.php#L66) method.
Nevertheless the method "bleeding" is still a problem for some classes.

I added two commits to this PR.
1. Addresses the classes that can/should be extended using a plugin.
2. Adds the `public static $methods = [];` to classes that are also affected by this.
  These classes cannot be extended by a plugin, but a developer might want to extend these for one reason or another.

## Ready?

### For review team

### DIFF
--- a/src/Cms/Blocks.php
+++ b/src/Cms/Blocks.php
@@ -26,6 +26,13 @@ class Blocks extends Items
 	public const ITEM_CLASS = Block::class;
 
 	/**
+	 * All registered blocks methods
+	 *
+	 * @var array
+	 */
+	public static $methods = [];
+
+	/**
 	 * Return HTML when the collection is
 	 * converted to a string
 	 *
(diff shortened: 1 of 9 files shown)


bastianallgeier:
TBH, I wasn't aware of the method bleeding. Your PR looks good to me, but @lukasbestle seems to be a bit more involved in this already so I would like to wait for his comment.

lukasbestle (approved):
Thank you. I forgot that we already look for the methods in the parent class, we just missed to define the `$methods` property in the new classes for `Blocks` and `Layouts`.

I think it's good to also add the prop to the other classes. I looked through the code base and it very much looks like you caught all classes that are children of classes with `HasMethods`. 👍

PR #5133, opened 2023-04-04
fix image alt adoption in default image block template

### LINKED ISSUE #5132: Alt text from image source does not get adopted
https://github.com/getkirby/kirby/blob/ce4aaace87ceca1f019a9cfc0293a69e662f9648/config/blocks/image/image.php#L14
In this line the `$alt` is not null so the field value of the image source never gets adopted even there is no text entered in the image block.
I could provide you with a PR with the fix i created for my project, but i wanted to ask if this intended behaviour before submitting it.

holmey:
$alt does not evaluate as null so the field value of the image source never gets adopted

Fixes #5132


afbora (dismissed):

distantnative:
We should add a unit test just to catch any regression in the future. If you @holmey can't do it, we can take over :)

holmey:
I have not found any other tests for the templates in the config/block folder on the first sight. So maybe you should continue from here or give me a hint how you would like to approach this scenario.

afbora (approved):

bastianallgeier:
This is great! Thank you for the PR 🙌

PR #5142, opened 2023-04-10
Fix detecting language with locale

### LINKED ISSUE #5135: locale matching does not work
## Description
Locale matching with language detection does not work with multiple locales for the same language.
`en-US` always just matches `en`when present. 

In https://github.com/getkirby/kirby/blob/main/src/Cms/App.php#L588 the language never matches for the locale. It's always the code that matches in the end. 
The reason lies in https://github.com/getkirby/kirby/blob/main/src/Toolkit/Collection.php#L413 because that compares an array (the locale configuration of a language) with a string (the locale string from the users browser language).

and because `en_US` is really not `array (6 => 'en_US')`
that part will never match anything.

**Expected behavior**  
I would expect the that the locale array is resolved to just `LC_ALL` so a match can happen and the user is redirected to the correct page

## To reproduce
In a fresh plainkit I create three languages:
de.php
```php
<?php
return [
    'code' => 'de',
    'default' => true,
    'direction' => 'ltr',
    'locale' => 'de_DE',
    'name' => 'Deutsch',
    'url' => '/de/'
];
```

en.php
```php
<?php
return [
    'code' => 'en',
    'default' => false,
    'direction' => 'ltr',
    'locale' => 'en_GB',
    'name' => 'English (EU)',
    'url' => '/en/'
];
```

us.php
```php
<?php
return [
    'code' => 'us',
    'default' => false,
    'direction' => 'ltr',
    'locale' => 'en_US',
    'name' => 'English (US)',
    'url' => '/us/'
];
```

and in the config set languages active and language detection to active
```php
return [
    'languages' => true,
    'languages.detect' => true,
];
```
visit homepage with browser language set to `en-US`

## Your setup
**Kirby Version**  
Current Plainkit 3.9.3

 - Device: MBP 16" 2021
 - OS: mac os 13.2.1
 - Browser: firefox, chrome. Set language to en-US
 - Version: doesn't matter

**Additional context**  
kirby running in a ddev managed docker container

afbora:
## This PR …

Details about issue and solution are here: https://github.com/getkirby/kirby/issues/5135#issuecomment-1496578886 and https://github.com/getkirby/kirby/issues/5135#issuecomment-1500974118

### Fixes
- Fix detecting language with locale #5135

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/tests/Cms/App/AppLanguagesTest.php
+++ b/tests/Cms/App/AppLanguagesTest.php
@@ -57,7 +57,8 @@ class AppLanguagesTest extends TestCase
 			['en-US', 'us'],
 			['de', 'de'],
 			['fr', 'en'],
-			['en-US,en;q=0.5', 'us'],
+			['en-US, en;q=0.5', 'us'],
+			['en-US;q=0.5, de;q=0.8, fr;q=0.9', 'de'],
 		];
 	}
 
--- a/tests/Cms/App/AppLanguagesTest.php
+++ b/tests/Cms/App/AppLanguagesTest.php
@@ -64,6 +64,7 @@ class AppLanguagesTest extends TestCase
 
 	/**
 	 * @dataProvider detectedLanguageProvider
+	 * @backupGlobals enabled
 	 */
 	public function testDetectedLanguage($accept, $expected)
 	{
@@ -93,6 +94,6 @@ class AppLanguagesTest extends TestCase
 			]
 		]);
 
-		$this->assertSame($app->detectedLanguage()->code(), $expected);
+		$this->assertSame($expected, $app->detectedLanguage()->code());
 	}
 }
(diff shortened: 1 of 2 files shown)


tests/Cms/App/AppLanguagesTest.php
```diff
@@ -48,4 +48,50 @@ public function testLanguageCode()
...
+
+	public function detectedLanguageProvider(): array
+	{
+		return [
+			['en', 'en'],
+			['en-GB', 'en'],
+			['en-US', 'us'],
+			['de', 'de'],
+			['fr', 'en'],
+			['en-US,en;q=0.5', 'us'],
```
afbora: Please let me know if you have different accepted settings/cases in mind.

lukasbestle (approved):

PR #5097, opened 2023-03-09
Inherit 'when' for fieldgroups

plain-solutions-gmbh:
## This PR gives the ability to set `when` props to a group of fields.

It will be inherit to his child fields. Use it like this:
```yml
fields:
  showit:
    type: toggle
  my_group:
    type: group
    when:
      showit: true
    fields:
      my_field1:
        type: text
      my_field2:
        type: text

```

## Ready?

- [ ] Unit tests for fixed bug/feature
- [ ] In-code documentation (wherever needed)
- [ ] Tests and checks all pass

### For review team

- [ ] Add changes to release notes draft in Notion
- [ ] Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)

PR #5143, opened 2023-04-10
Inherit `when` for field groups

afbora:
## This PR …

Implements https://github.com/getkirby/kirby/pull/5097 PR

This PR gives the ability to set `when` props to a group of fields.

It will be inherit to his child fields. Use it like this:
```yml
fields:
  showit:
    type: toggle

  my_group:
    type: group
    when:
      showit: true
    fields:
      my_field1:
        type: text
      my_field2:
        type: text

```

### Enhancements

- Inherit `when` for field groups

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Blueprint.php
+++ b/src/Cms/Blueprint.php
@@ -463,7 +463,7 @@ class Blueprint
 
 			if (isset($props['when']) === true) {
 				$fields = array_map(
-					fn ($field) => array_merge_recursive(['when' => $props['when']], $field),
+					fn ($field) => array_replace_recursive(['when' => $props['when']], $field),
 					$fields
 				);
 			}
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
Code-wise it looks great. 👍

But it would be good to have another review by @distantnative or @bastianallgeier: Do you agree that this feature is useful? Does it fit with the plans for the blueprint refactoring?

lukasbestle:
Oh, looks like we need `array_replace_recursive()` instead to avoid that multiple values get merged into an array. Or have I missed something @afbora?

lukasbestle (approved):
Approving the code changes, but see my comment above:

> Code-wise it looks great. 👍
>
> But it would be good to have another review by @distantnative or @bastianallgeier: Do you agree that this feature is useful? Does it fit with the plans for the blueprint refactoring?

bastianallgeier:
I like it and the implementation looks pretty straight forward.

PR #5145, opened 2023-04-11
Symfony YAML library

distantnative:
### Enhancements
- Introduced new YAML handler that can be activated by setting the config option `yaml` to `'symfony'`. We plan to switch over to the Symfony handler in the next major releases. For now, `'spyc'` remains the default.

### Breaking changes
None

## Ready?

### For review team


bastianallgeier:
@distantnative do you see a chance of fixing the code style issues? Are they issues at all or did the process just fail?

distantnative:
@bastianallgeier the weird thing is that my local Cs fixer doesn't pick it up at all

src/Data/Yaml.php
```diff
@@ -42,16 +44,15 @@ public static function decode($string): array
...
+			default   => YamlSpyc::decode($string)
+		};
+	}
 
-		if (is_array($result) === true) {
-			return $result;
-		}
+	public static function handler(): string
+	{
+		return App::instance(null, true)?->option('yaml') ?? 'spyc';
```
afbora:
```suggestion
		return App::instance(null, true)?->option('yaml.handler') ?? 'spyc';
```

tests/Data/YamlSymfonyTest.php
```diff
@@ -0,0 +1,136 @@
...
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @coversDefaultClass \Kirby\Data\Yaml
+ */
+class YamlSymfonyTest extends TestCase
+{
+	public function setUp(): void
+	{
+		new App(['options' => ['yaml' => 'symfony']]);
```
afbora:
```suggestion
		new App(['options' => ['yaml.handler' => 'symfony']]);
```

composer.json
```diff
@@ -44,7 +44,8 @@
 		"michelf/php-smartypants": "1.8.1",
 		"phpmailer/phpmailer": "6.8.0",
 		"symfony/polyfill-intl-idn": "1.27.0",
-		"symfony/polyfill-mbstring": "1.27.0"
+		"symfony/polyfill-mbstring": "1.27.0",
+		"symfony/yaml": "^5.4"
```
afbora:
```suggestion
		"symfony/yaml": "5.4.21"
```

src/Data/Yaml.php
```diff
@@ -42,16 +44,15 @@ public static function decode($string): array
...
+		return match (static::handler()) {
+			'symfony' => YamlSymfony::decode($string),
+			default   => YamlSpyc::decode($string)
+		};
+	}
 
-		if (is_array($result) === true) {
-			return $result;
-		}
+	public static function handler(): string
```
lukasbestle, line 58: I think it would be good to also have a doc block here. It's not immediately obvious that the return value will be just the handler type without looking at the code. Of course the return type already tells us that it's not an object like with `Data::handler()`, but a string could still be anything.

Also I'd mark it `@internal` as we will remove this method again once we remove `spyc` support.

```suggestion
	/**
	 * Returns the configured YAML library (`'symfony'` or `'spyc'`)
	 * @internal
	 */ 
	public static function handler(): string
```

src/Data/YamlSymfony.php
```diff
@@ -0,0 +1,42 @@
...
+class YamlSymfony
+{
+	/**
+	 * Converts an array to an encoded YAML string
+	 */
+	public static function encode($data): string
+	{
+		return Symfony::dump(
+			$data,
+			2,
```
lukasbestle: This second argument "customizes the level at which the output switches from the expanded representation to the inline one" according to the docs.

Have you tested what happens if we throw a deeply nested array at it? IMO it must not start to use the inline syntax at all considering our deeply nested blueprint structures.

src/Data/Yaml.php
lukasbestle: Good idea IMO. For now we don't have any other options in this namespace, but in the future we might have (like in the `markdown` namespace). If we already use a sub-option, we don't need to deal with naming conflicts here down the road.

src/Data/YamlSymfony.php
afbora: Output with nested structure field:

### Before
````
----

Addresses:

- 
  street: A
  zip: B
  city: C
  addresses:
    - 
      street: "1"
      zip: "2"
      city: "3"
    - 
      street: "4"
      zip: "5"
      city: "6"
    - 
      street: "7"
      zip: "8"
      city: "9"

----
````

### After
````
----

Addresses:

-
  street: A
  zip: B
  city: C
  addresses: [{ street: '1', zip: '2', city: '3' }, { street: '4', zip: '5', city: '6' }]

----
````

Comment (lukasbestle) [reply src/Data/YamlSymfony.php]: Thanks for testing! I feel like the nested structure still makes more sense for us, considering version control with Git and manual editing by humans.

Is there a way to disable the inline feature completely? Or do we need to pass a higher value like `9999`?

Comment (afbora) [reply src/Data/YamlSymfony.php]: I tested it, I couldn't find a better solution than 9999. 

Also thinking out loud, just a idea: do you think it would be a good idea to set custom parameters for yaml handler from config.php?

````php
'yaml' => [
    'handler' => 'symfony',
    'params' => [
        'inline' => 0,
        'indent' => 4
    ],
]
````

Comment (lukasbestle) [reply src/Data/YamlSymfony.php]: I'd name it `options`, apart from that 👍👍.

**Edit:** No, `params` is actually great IMO.

File: src/Data/Yaml.php [hunk 6]
```diff
@@ -50,8 +50,13 @@ public static function decode($string): array
...
 
+	/**
+	 * Returns which YAML parser (`spyc` or `symfony`)
+	 * is configured to be used
+	 * @internal
+	 */
 	public static function handler(): string
 	{
-		return App::instance(null, true)?->option('yaml') ?? 'spyc';
+		return App::instance(null, true)?->option('yaml.handler') ?? 'spyc';
```
afbora, line 60:
```suggestion
		return App::instance(null, true)?->option('yaml.handler', 'spyc');
```

src/Data/YamlSymfony.php
```diff
@@ -21,10 +22,12 @@ class YamlSymfony
...
 	public static function encode($data): string
 	{
+		$kirby  = App::instance(null, true);
+
 		return Symfony::dump(
 			$data,
-			2,
-			2,
+			$kirby?->option('yaml.params.inline') ?? 9999,
+			$kirby?->option('yaml.params.indent') ?? 2,
```
afbora, line 30:
```suggestion
			$kirby?->option('yaml.params.inline', 9999),
			$kirby?->option('yaml.params.indent', 2),
```

src/Data/Yaml.php
distantnative: This won't work if the lazy App::instance returns null

src/Data/YamlSymfony.php
distantnative: Same here

src/Data/Yaml.php
afbora: Oh sorry, I missed that!

afbora (approved):
LGTM 👍

PR #5153, opened 2023-04-19
Response: support range requests for files

distantnative:
It's a start... but I think I am at the point where I need input from others

### DIFF
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -784,6 +784,39 @@ class F
 	}
 
 	/**
+	 * Continously outputs the file between the provided range
+	 */
+	public static function stream(
+		string $file,
+		int $start = 0,
+		int $end = null
+	): void {
+		$size   = static::size($file);
+		$end  ??= $size;
+
+		if ($end === $size) {
+			echo static::read($file);
+			return;
+		}
+
+		$handle = fopen($file, 'rb');
+		fseek($handle, $start);
+
+		if (!feof($handle)) {
+			throw new Exception('Invalid file handle');
+		}
+
+		while ($start < $end) {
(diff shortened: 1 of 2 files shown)


src/Filesystem/F.php
```diff
@@ -783,6 +783,39 @@ public static function size(string|array $file): int
...
+	 */
+	public static function stream(
+		string $file,
+		int $start = 0,
+		int $end = null
+	): void {
+		$size   = static::size($file);
+		$end  ??= $size;
+
+		if ($end === $size) {
```
lukasbestle, line 797: Shouldn't this also check for start being 0? Otherwise the full file is returned even when the beginning wasn't requested.

src/Filesystem/F.php
```diff
@@ -783,6 +783,39 @@ public static function size(string|array $file): int
...
+	 * Continously outputs the file between the provided range
+	 */
+	public static function stream(
+		string $file,
+		int $start = 0,
+		int $end = null
+	): void {
+		$size   = static::size($file);
+		$end  ??= $size;
+
```
lukasbestle, line 796: I think there should be checks for `start < 0`, `start >= end` and `end > size` here as well, throwing an exception to provide extra safety and for those who use this method directly.

lukasbestle (commented):
I haven't yet tested it or checked for edge cases, but in general (code structure) it looks really good already.

bastianallgeier:
What's the state of this? Should we keep this in the lab and remove the PR?

distantnative:
It's out here because  I need help from the rest of the team.  I won't get much further on my own. If this happens now or when, to be decided.

distantnative:
Closing this for now. PR is referenced in corresponding issue and can be revisited when tackling this again.

Issue #5171, opened 2023-05-09
get(#) retrieves parameters by index rather than by key


Jayshua:
## Description

When submitting a form using numeric field names like `<input name="1"><input name="0">` the browser submits the form data in url encoded format like `1=first+value&0=second+value`. The `get` helper, when called like `get(1)` or with a string like `get('1')`, will retrieve the value at *index* 1 in the encoded string ('second+value' in this example) rather than with *key* 1 ('first+value' in this example).

**Expected behavior**
I expected `get` to retrieve by key rather than by index, matching PHP's behavior when using `$_POST`.

**My Diagnosis**
`\Kirby\Http\Request::data()` calls `array_merge` to combine the body and query param data into a single data array. `array_merge` is documented to renumber numeric keys. I believe `data` should call `array_replace` instead, which combines the arrays without renumbering them.

## To reproduce

1. Replace index.php with:
   ```php
   <?php
   require 'kirby/bootstrap.php';

   new Kirby();
   var_dump(get(1));
   var_dump(get('1'));
   var_dump($_POST[1]);
   var_dump($_POST['1']);
   ```

2. Submit a POST request in the standard `application/x-www-form-urlencoded` format. Here are some examples:
   `1=first+value&0=second+value`
   `0=second+value&1=first+value`

## Your setup
**Kirby Version**: 3.9.2

lukasbestle:
Thanks, looks like we hadn't expected/thought of numeric query params when implementing the logic.

afbora:
@lukasbestle I've looked into issue. I've found the source of issue. This issue is coming from following line:

https://github.com/getkirby/kirby/blob/282b1e25faa0bc24c903237a7484dc45fa475c36/src/Http/Request.php#L196-L199

As as solution, I think we need to use `array_replace` instead `array_merge`. I've tested and seems works great. But I'm not sure if this change will break anything. What do you think?

> array_merge documentation: values in the input arrays with numeric keys will be renumbered with incrementing keys starting from zero in the result array.

lukasbestle:
The only difference between `array_merge()` and `array_replace()` that I'm aware of is the handling of numeric keys. Since we want PHP to treat keys as strings, not as numbers, `array_replace()` is indeed the correct choice here IMO.

PR #5173, opened 2023-05-10
Fix retrieving numeric parameters

afbora:
### Fixes
- Correctly process query params and body fields with numeric keys #5171

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/tests/Http/RequestTest.php
+++ b/tests/Http/RequestTest.php
@@ -70,6 +70,25 @@ class RequestTest extends TestCase
 		$this->assertNull($request->get('c'));
 	}
 
+	public function testDataNumeric()
+	{
+		$request = new Request([
+			'body'   => [1 => 'a'],
+			'query'  => [
+				'0' => 'b',
+				2   => 'c'
+			]
+		]);
+
+		$this->assertSame([1 => 'a', 0 => 'b', 2 => 'c'], $request->data());
+		$this->assertSame('b', $request->get(0));
+		$this->assertSame('b', $request->get('0'));
+		$this->assertSame('a', $request->get(1));
+		$this->assertSame('a', $request->get('1'));
+		$this->assertSame('c', $request->get(2));
+		$this->assertSame('c', $request->get('2'));
+	}
+
 	public function test__debuginfo()
 	{
 		$request = new Request();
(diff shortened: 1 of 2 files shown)


tests/Http/RequestTest.php
```diff
@@ -70,6 +70,25 @@ public function testData()
...
+	{
+		$request = new Request([
+			'body'   => [1 => 'a'],
+			'query'  => [
+				'0' => 'b',
+				2   => 'c'
+			]
+		]);
+
+		$this->assertSame([1 => 'a', 0 => 'b', 2 => 'c'], $request->data());
```
lukasbestle, line 83: I think in the real world, all keys in the `body` and `query` input arrays should be numeric strings instead of integers (but would be good to verify that).

However with numeric strings, the call to `$request->data()` should also return numeric strings as keys and not convert them to ints. In this test `'0'` starts out as a numeric string but suddenly seems to get converted to an integer, which is IMO wrong. Why is that happening?

I'd expect the following result:

```suggestion
		$this->assertSame(['1' => 'a', '0' => 'b', '2' => 'c'], $request->data());
```

afbora: Dump for `?1=b&0=a&2=c&foo=bar`

### Output
````php
// var_dump($_GET);
array(4) {
  [1]=>
  string(1) "b"
  [0]=>
  string(1) "a"
  [2]=>
  string(1) "c"
  ["foo"]=>
  string(3) "bar"
}

// var_dump(get());
array(4) {
  [1]=>
  string(1) "b"
  [0]=>
  string(1) "a"
  [2]=>
  string(1) "c"
  ["foo"]=>
  string(3) "bar"
}
````

So `get()` integer array index is same with `$_GET`

afbora: @lukasbestle You will also be surprised! 😅 

````php
var_dump(['0' => 'Wow!']);

array(1) {
  [0]=>
  string(4) "Wow!"
}
````

lukasbestle: I'm indeed surprised by yet another PHP quirk I wasn't aware of until now. 🤯
PHP actually converts numeric keys to integers internally. Not just in the `var_dump()` output, but even when calling `is_int($key))` within a loop for example.

lukasbestle (approved):
💛

Issue #5039, opened 2023-02-03
`$page->next()` not returning correct class type


tobimori:
## Description
When using `$page->next()->url()`, my IDE returns that `url()` is an undefined function.
This was first posted by me in a discussion on [Discord](https://discord.com/channels/525634039965679616/525641819854471168/1070854232221155418).
![image](https://user-images.githubusercontent.com/29142128/216572923-a35bc173-8dcd-4edc-afa4-d6e5240227e7.png)

This most likely happens because the `next()` (and probably also `prev()`) methods that are defined in https://github.com/getkirby/kirby/blob/main/src/Cms/HasSiblings.php were typed to always return a `Kirby\Cms\Model` object, which obviously do not contain page-related methods.

**Expected behavior**  
Ideally, my IDE would detect that the return value of `$page->next()` is a `Page` class, and thus I can use `url()` or other page-specific methods.
<img width="671" alt="CleanShot 2023-02-03 at 11 11 53@2x" src="https://user-images.githubusercontent.com/29142128/216573596-7be8f71b-8ecd-4862-ab1b-b506c9d16873.png">

## Workaround
@lukaskleinschmidt posted the following workaround on Discord, which (option B) works quite well for me.
```php
<?php
// You can put a file called autocomplete.php in the root folder for example.

/**
 * A helper file for Kirby, to provide autocomplete information to your IDE
 */

namespace Kirby\Cms {

    /**
     * Option A.
     * 
     * @method \Kirby\Cms\Page|null next($collection = null)
     */
    class Page {}

    /**
     * Option B will also work for custom page models.
     * 
     * @method static|null next($collection = null)
     */
    class Page {}

}
```

Pretty sure that this can be adapted to be included 'by default'.

## Your setup
**Kirby Version**  
3.9.0

lukasbestle:
I wonder if there's anything that speaks against type-hinting `HasSiblings::next()` and `HasSiblings::prev()` with a return value of `static` in the first place. The CMS collections like `Cms\Pages` validate that each element is of the same type, so these methods should always return an object of the same type as the current one.

distantnative:
I was thinking about the same thing, I don't think we've ever really mixed collections.

But what about page models? Static would be fine or will it take on the more specific class?

lukasbestle:
Right, that's a problem. I think we need to add `@method` type hints to each model class with the fixed class name then.

lukaskleinschmidt:
[...]

PR #5175, opened 2023-05-12
Fix type hints for `HasSiblings`; fix behavior of `$collection->indexOf()`

lukasbestle:
### Fixes

- `$collection->indexOf()` returns `false` instead of the index `0` when the passed element was not found in the collection #5175
- The collection `next()` and `prev()` methods are now type-hinted to return `static` for improved IDE support #5039

## Ready?

### For review team

### DIFF
--- a/src/Cms/Collection.php
+++ b/src/Cms/Collection.php
@@ -209,9 +209,9 @@ class Collection extends BaseCollection
 	 * or ids and then search accordingly.
 	 *
 	 * @param string|object $needle
-	 * @return int
+	 * @return int|false
 	 */
-	public function indexOf($needle): int
+	public function indexOf($needle): int|false
 	{
 		if (is_string($needle) === true) {
 			return array_search($needle, $this->keys());
--- a/src/Cms/HasSiblings.php
+++ b/src/Cms/HasSiblings.php
@@ -19,9 +19,9 @@ trait HasSiblings
 	 *
 	 * @param \Kirby\Cms\Collection|null $collection
 	 *
-	 * @return int
+	 * @return int|false
 	 */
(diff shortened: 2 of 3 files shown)


lukasbestle:
The unit test failures seem to be caused specifically by the added return type hints for `HasSiblings::next()` and `::prev()`. Once I remove the return type hints, everything passes.

@afbora Could you please look into the root cause for this?

afbora:
I can't reproduce the unit test errors in GitHub CI on my local 🤷‍♂️ So I haven't caught exactly why, but I have an idea.

I see that `\Kirby\Form\FieldClass` is the only one that uses the `HasSiblings` trait differently. The others seem to be Collection items while `FieldClass` is not. I think this is the cause of the error.

https://github.com/getkirby/kirby/blob/282b1e25faa0bc24c903237a7484dc45fa475c36/src/Form/FieldClass.php#L28

lukasbestle:
Thanks, indeed there seems to be something weird going on in the `Form` package.

Since `Block::toHtml()` swallows the error, the test output doesn't show the cause. I've commented the `try catch` block out and got this:

```
TypeError: Kirby\Cms\Item::next(): Return value must be of type ?Kirby\Cms\HeadingBlock, Kirby\Cms\DefaultBlock returned
```

So seems to be a deeper problem. I've reduced this PR to the minimum viable fix to avoid regressions.

afbora (dismissed):
LGTM 👍 

@lukasbestle But also I think an description or todo about the failure of unit tests into doc-block would be good for the future.

bastianallgeier (dismissed):

afbora (approved):

bastianallgeier (approved):

PR #5190, opened 2023-05-25
[v4] `$model->file()`: adhere to sort order

distantnative:
## This PR …

Especially in the Panel, it currently seems odd that the default preview image is not the first image as shown not he Panel page

### Enhancement
- `$model->file()`, `$model->image()` etc. return the first file/image considering the `sort` field (and not just the filename).

### Breaking changes
- `$model->file()`, `$model->image()` etc. return a different file now, if the lowest `sort` field value is not also the first alphabetical filename

## Ready?

### For review team

### DIFF
--- a/src/Cms/HasFiles.php
+++ b/src/Cms/HasFiles.php
@@ -90,7 +90,10 @@ trait HasFiles
 	public function file(string $filename = null, string $in = 'files')
 	{
 		if ($filename === null) {
-			return $this->$in()->first();
+			return $this
+				->$in()
+				->sortBy('sort', 'asc', 'filename', 'asc')
+				->first();
 		}
 
 		// find by global UUID


lukasbestle (approved):
I think it makes sense. But more opinions would be good.

distantnative:
What bothers me is that now the system needs to read every content file just to return one file

lukasbestle:
Hm, that's true. Maybe we could use a cache for the sorting order and file templates. But that adds complexity again.

distantnative:
Yea, there my gut tells me that it's out of proportion. Trying to fix something nobody has much complained so far. Maybe something to take to #concepts to get more input on how benefit vs. costs would play out for users.

bastianallgeier:
While I agree that it makes more sense, I also agree that it feels too massive for such a simple comfort feature

distantnative:
Let's close it?

PR #5191, opened 2023-05-25
[v4] Panel: add pagination to search results

distantnative:
# Breaking changes
- Removed `this.$config.search` from Panel
- Area `search` plugins receive two additional arguments for their `query` callback: `$limit` and `$page` to be used to paginate the results. They should then return an array with entries `results` and `pagination`.

### DIFF
--- a/src/Panel/Panel.php
+++ b/src/Panel/Panel.php
@@ -449,8 +449,7 @@ class Panel
 					$kirby   = App::instance();
 					$request = $kirby->request();
 					$query   = $request->get('query');
-					$limit   = (int)$request->get('limit', 10);
-					$limit ??= $kirby->option('panel.search.limit', 10);
+					$limit   = (int)$request->get('limit', $kirby->option('panel.search.limit', 10));
 					$page    = (int)$request->get('page', 1);
 
 					return $params['query']($query, $limit, $page);
--- a/src/Panel/Panel.php
+++ b/src/Panel/Panel.php
@@ -198,7 +198,7 @@ class Panel
 		$request = App::instance()->request();
 
 		return Response::json($data, $code, $request->get('_pretty'), [
-			'X-Fiber' => 'true',
+			'X-Fiber'       => 'true',
 			'Cache-Control' => 'no-store, private'
 		]);
 	}
@@ -446,9 +446,14 @@ class Panel
 				'type'    => 'search',
 				'area'    => $areaId,
 				'action'  => function () use ($params) {
-					$request = App::instance()->request();
+					$kirby   = App::instance();
+					$request = $kirby->request();
+					$query   = $request->get('query');
+					$limit   = (int)$request->get('limit', 10);
+					$limit ??= $kirby->option('panel.search.limit', 10);
+					$page    = (int)$request->get('page', 1);
 
-					return $params['query']($request->get('query'));
+					return $params['query']($query, $limit, $page);
 				}
 			];
 		}
--- a/src/Panel/Search.php
+++ b/src/Panel/Search.php
@@ -19,15 +19,4 @@ use Kirby\Http\Response;
 class Search extends Json
 {
 	protected static string $key = '$search';
-
-	public static function response($data, array $options = []): Response
-	{
-		if (is_array($data) === true) {
-			$data = [
-				'results' => $data
-			];
-		}
-
-		return parent::response($data, $options);
-	}
 }
--- a/src/Panel/Search.php
+++ b/src/Panel/Search.php
@@ -19,4 +19,23 @@ use Kirby\Http\Response;
 class Search extends Json
 {
 	protected static string $key = '$search';
+
+	public static function response($data, array $options = []): Response
+	{
+		if (
+			is_array($data) === true &&
+			array_key_exists('results', $data) === false
+		) {
+			$data = [
+				'results'    => $data,
+				'pagination' => [
+					'page'      => 1,
+					'limit'     => $total = count($data),
+					'total'     => $total
+				]
+			];
+		}
+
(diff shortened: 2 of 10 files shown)


bastianallgeier:
TBH, I think we need a different concept for this. In hindsight it was a mistake to bind search plugins so closely to the look and feel of results and such stuff as pagination. If we introduce this, all search plugins need to be changed. I wonder if it makes more sense to just let search plugins focus on the results and we handle how they look like. Or we have another layer for the result previews. I'm not sure.

lukasbestle:
I think we do need pagination right in the search logic though. If we first let a search plugin return all results and paginate it afterwards in the core, the performance will be pretty bad for "proper" search APIs.

distantnative:
I think Basti's idea is let the plugins return right before the paginate call and rather do the paginate call on any of them in our core and then depending on what kind of Collection it is transform them into previews.

The problem here is that search plugins are mainly used for types that aren't our core types. So we couldn't anticipate how to generate the preview data for them.

lukasbestle:
Well yeah, but wouldn't this mean then that search plugins would first need to return a collection with 80 results, 60 of which will then be dropped by the core during pagination? If the search plugin handles pagination, it can return just the needed 20 results in the first place.

distantnative:
But isn't it just the same? Whether the plugin adds the `->paginate()` call or the core right after the plugin doesn't really matter in the way we chain our collection calls, does it? Both first need to do the search filter call with 80 results and then paginate...

distantnative:
My main problem is that only the plugins would know how to generate a preview for the specific item.

lukasbestle:
> Both first need to do the search filter call with 80 results and then paginate...

Not necessarily. With the current implementation of our collection classes yes, but a proper pagination implementation would only fetch the requested pagination subset from the database/search server in the first place and then only add markers for the total available number of results and the current page, offset and limit.

distantnative:
You're right. I was always only thinking in terms of our Collections even for custom types. But they could come from many sources.

So that's 2 reasons then even to keep it all in the search plugins/callbacks. Even if it's a bit redundant for our 3 types.

distantnative:
In that case, this could be already reviewed and merged (depends on review result), @bastianallgeier

src/Panel/Search.php
```diff
@@ -19,15 +19,4 @@
 class Search extends Json
 {
 	protected static string $key = '$search';
-
-	public static function response($data, array $options = []): Response
-	{
-		if (is_array($data) === true) {
-			$data = [
-				'results' => $data
```
bastianallgeier, line 27: Oh, couldn't we just use this place to polyfill old search plugins that don't define their own results and pagination keys?

distantnative: As in `if is array but not has array key "results"`?

distantnative: But how would we get the pagination for them when they're already transformed to preview items?

bastianallgeier: I think we need to switch off pagination for those custom searches that don't define it. Then it's an opt-in feature.

distantnative: Or we just treat it as `page: 1, limit: array.length, total: array.length`

bastianallgeier: Great idea

bastianallgeier (approved):

distantnative:
@bastianallgeier since search plugins need to implement the pagination on their side, I was wondering if we could somehow simplify for them getting page and limit, instead of

```php
(int)$kirby->request()->get('limit', 10),
(int)$kirby->request()->get('page', 1)
```

Maybe to hand them in addition to the `$query` to the `query` callback?

src/Panel/Panel.php
```diff
@@ -446,9 +446,14 @@ public static function routesForSearches(string $areaId, array $area): array
...
 				'area'    => $areaId,
 				'action'  => function () use ($params) {
-					$request = App::instance()->request();
-
-					return $params['query']($request->get('query'));
+					$kirby   = App::instance();
+					$request = $kirby->request();
+					$query   = $request->get('query');
+					$limit   = (int)$request->get('limit', 10);
+					$limit ??= $kirby->option('panel.search.limit', 10);
```
afbora: I didn't understand these lines. First line has casting integer with fallback set, second line uses `??=` 🤔 

Also I couldn't get the new `panel.search.limit` setting to work.

afbora:
Also needs to add to PR description about new `panel.search.limit` option.

src/Panel/Panel.php
distantnative: Oh I think the `, 10` shouldn't be there on line 452.

distantnative:
@afbora the option isn't new, but undocumented so far

src/Panel/Panel.php
afbora: No, also casting prevents to running next line. I think `(int)null` means `0`.

distantnative: True. Ok I tried to format it nicely, but then it's better to make it a one liner:
```
$limit  = (int)$request->get('limit', $kirby->option('panel.search.limit', 10));
```

afbora: Also one more thing; `$request->get('limit')` return always `15`. This prevents running fallback option. Do you know why?

```
/panel/search?type=pages&query=test
```

distantnative: For the search view that's correct. The option is only meant for the dialog (as it has been in the past)

afbora: OK

afbora:
```suggestion
					$limit   = (int)$request->get('limit', $kirby->option('panel.search.limit', 10));
```

PR #5207, opened 2023-05-29
`Query` package: remove strict types again

### LINKED ISSUE #5185: [v4] Translate function doesn't work with field values anymore
## Description
I use the following code to get a translated label for a value from a site field, which worked fine in Kirby 3:

https://github.com/tobimori/kirby-seo/blob/main/blueprints/fields/og-group.yml#L35
```yaml
    placeholder: "{{ t('default-select') }} {{ t(site.twitterCardType) }}"
```

Yet, with Kirby 4, I'm missing the output for the translated card type:
CleanShot 2023-05-25 at 12 06 11@2x

When removing the translation helpers, I can see that the data can be accessed correctly.
CleanShot 2023-05-25 at 12 06 38@2x

It works correctly when appending `.value` to `site.twitterCardType`:
```yaml
    placeholder: "{{ t('default-select') }} {{ t(site.twitterCardType.value) }}"
```

CleanShot 2023-05-25 at 12 07 53@2x

## To reproduce
Install my SEO plugin with Kirby 4.0, or test using the code snippets above.

## Your setup
**Kirby Version**  4.0-alpha.1

distantnative:
## This PR …

Strict types are causing issues when we want to support passing a `Field` object which gets casted to a string (the field value). I currently see the benefits of stricter types not outweighing the issues we would have to add checks for `Field` objects all over. That's why I suggest that we just revert the strict types for now again.

### Fixes
- #5185

### DIFF
--- a/src/Query/Argument.php
+++ b/src/Query/Argument.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types = 1);
-
 namespace Kirby\Query;
 
 use Closure;
--- a/src/Query/Arguments.php
+++ b/src/Query/Arguments.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types = 1);
-
 namespace Kirby\Query;
 
 use Kirby\Toolkit\A;
--- a/src/Query/Expression.php
+++ b/src/Query/Expression.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types = 1);
-
 namespace Kirby\Query;
 
 use Kirby\Exception\LogicException;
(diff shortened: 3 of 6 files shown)


lukasbestle (approved):
I agree this is a good first solution.

How would we fix this issue in the long run when we want/need to have strict types at some point?

distantnative:
It would be adding a lot of checks for `Field` and resolving explicitly to `->value()`. It's one of these areas where strictness and the relaxed-/comfort-handling of Kirby clash.

lukasbestle:
I see. So right now we are relying on PHP auto-converting to string, right?

We could have a static helper method like `Field::toString(Field|string $field): string` that handles this. But it's not pretty either.

PR #5211, opened 2023-05-29
Fix page move dialog + tree recursion

### LINKED ISSUE #5189: [v4] Moving pages into site not possible
## Description

It is not possible to move a page directly to the site. An error always appears there.

**Screenshots**  
Bildschirmfoto 2023-05-25 um 14 12 01

## To reproduce

Just try to push a page that has a parent page into the site. With me there always appears an error

distantnative:
## This PR …

- `k-page-tree` fully extends `k-tree`
- Uses actual recursion of `k-tree` (or its extensions/versions like `k-page-tree`)

### Fixes
- #5189

### DIFF
--- a/src/Cms/Find.php
+++ b/src/Cms/Find.php
@@ -77,9 +77,16 @@ class Find
 	public static function page(string $id)
 	{
 		// decode API ID encoding
-		$id = str_replace(['+', ' '], '/', $id);
+		$id    = str_replace(['+', ' '], '/', $id);
+		$kirby = App::instance();
 
-		$page = App::instance()->page($id);
+		// support looking up the site itself,
+		// as this will fail when using $kirby->page()
+		if ($id === '/' || $id === 'site://') {
+			return $kirby->site();
+		}
+
+		$page = $kirby->page($id, null, true);
 
 		if ($page?->isAccessible() === true) {
 			return $page;
--- a/src/Cms/Find.php
+++ b/src/Cms/Find.php
@@ -79,14 +79,7 @@ class Find
 		// decode API ID encoding
 		$id    = str_replace(['+', ' '], '/', $id);
 		$kirby = App::instance();
-
-		// support looking up the site itself,
-		// as this will fail when using $kirby->page()
-		if ($id === '/' || $id === 'site://') {
-			return $kirby->site();
-		}
-
-		$page = $kirby->page($id, null, true);
+		$page  = $kirby->page($id, null, true);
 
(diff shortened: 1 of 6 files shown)


src/Cms/Find.php
```diff
@@ -77,9 +77,16 @@ public static function language(string $code)
 	public static function page(string $id)
 	{
 		// decode API ID encoding
-		$id = str_replace(['+', ' '], '/', $id);
+		$id    = str_replace(['+', ' '], '/', $id);
+		$kirby = App::instance();
+
+		// support looking up the site itself,
+		// as this will fail when using $kirby->page()
+		if ($id === '/' || $id === 'site' || $id === 'site://') {
```
lukasbestle: What happens with a page with slug `site`?

distantnative: 👍

afbora:
@distantnative I can move a note or album page under to site/root on starterkit. Is this expected?

````yaml
columns:
  - width: 1/2
    sections:
      albums: sections/albums

  - width: 1/2
    sections:
      notes: 
        extends: sections/notes
        parent: kirby.page("notes")

      pages:
        type: pages
        create: default
        templates:
          - about
          - home
          - default
````

**Comment by distantnative** [comment]
I guess so - since there are two sections on the site.yaml that do allow these templates.

**Comment by afbora** [comment]
But its parents (photography and notes) allows it (album and note), not the site.

**Comment by distantnative** [comment]
You are right. But I think this is a separate problem, that we should deal with in a separate issue and PR. `PageRules::move()` just collects the blueprints but doesn't first check if the section refers to the current page

File: src/Cms/Find.php [hunk 2]
```diff
@@ -77,9 +77,16 @@ public static function language(string $code)
 	public static function page(string $id)
 	{
 		// decode API ID encoding
-		$id = str_replace(['+', ' '], '/', $id);
+		$id    = str_replace(['+', ' '], '/', $id);
+		$kirby = App::instance();
+
+		// support looking up the site itself,
+		// as this will fail when using $kirby->page()
+		if ($id === '/' || $id === 'site://') {
```
bastianallgeier: This feels wrong to me. Why would the page method also be able to find the site? Is this just for convenience?

distantnative: Would you rather do it in the page move dialog submit handler?

bastianallgeier: Yes, I think so.

bastianallgeier:
LGTM now.

PR #5212, opened 2023-05-29
Rename Panel API

distantnative:
- Renamed panel `Module` to `State`
- Renamed panel `Island` to `Modal`


panel/src/panel/drawer.test.js
```diff
@@ -11,7 +11,7 @@ describe.concurrent("panel.drawer", () => {
 			component: null,
 			isLoading: false,
 			isOpen: false,
-			island: true,
```
bastianallgeier, line 14: While I like the new naming everywhere, I wonder if we should use something like legacy here instead.

distantnative: True that would be more explanatory what the flag is used for

bastianallgeier: I think it also makes more sense to turn this around. legacy: false is then the new default and this is how it should be, since we are phasing out the others

distantnative: 👍

distantnative: I can change it tonight or if you want to go ahead, feel free to make the change

distantnative: @bastianallgeier changed it

bastianallgeier (approved):

PR #5213, opened 2023-05-29
Panel: more error log in debug mode

distantnative:
- New `panel.error()` method that handles logging error in debug mode and opening notification if needed


panel/src/panel/upload.js
```diff
@@ -244,6 +244,10 @@ export default (panel) => {
 					filename: file.name + "." + file.extension,
 					url: this.url,
 					error: (xhr, src, response) => {
+						if (panel.debug) {
```
bastianallgeier: Why don't we use panel.error here as well?

distantnative: Cause I'm stupid 😆

bastianallgeier: That's what you said 🤣

bastianallgeier (approved):

PR #5216, opened 2023-05-29
Fix field API

### LINKED ISSUE #5184: [v4] New File Picker not working inside Block
## Description
With Kirby 4.0, I now get an error when I try to open the file picker inside a Block:
CleanShot 2023-05-25 at 11 54 03@2x
The same blueprint works fine when placed outside a block.

```json
{
    "status": "error",
    "message": "Call to a member function filepicker() on array",
    "code": 500,
    "exception": "Error",
    "key": null,
    "file": "../kirby/config/fields/files.php",
    "line": 95,
    "details": [],
    "route": "pages/([a-zA-Z0-9\\.\\-_%= \\+\\@\\(\\)]+)/fields/([a-zA-Z0-9\\.\\-_%= \\+\\@\\(\\)]+)(?:/(.*))?"
}
```

**Expected behavior**  
The File picker should open normally.

## To reproduce
Use my example blueprint for a field inside a block:
```yaml
      images:
        label:
          de: Hintergrundbilder
          en: Background images
        type: files
        uploads:
          parent: site.find('assets')
          template: image
        image:
          cover: false
        text: '{{ file.alt }} ({{ file.filename }})'
        info: '{{ file.tags.join(", ") }}'
        query: site.find('assets').images.filterBy('template', 'image')
        multiple: true
        layout: cards
        size: medium
```
I tested it with, and without a query, and it seemingly doesn't matter.
The same blueprint works fine on page level.

## Your setup
**Kirby Version**  4.0-alpha.1

distantnative:
### Fixes
- #5184

### DIFF
--- a/config/fields/mixins/upload.php
+++ b/config/fields/mixins/upload.php
@@ -22,9 +22,9 @@ return [
 				$uploads = [];
 			}
 
-			$template = $uploads['template'] ?? null;
+			$uploads['accept'] = '*';
 
-			if ($template) {
+			if ($template = $uploads['template'] ?? null) {
 				$file = new File([
 					'filename' => 'tmp',
 					'parent'   => $this->model(),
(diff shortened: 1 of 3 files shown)


distantnative:
@bastianallgeier I think we can merge this nevertheless of the coverage check

PR #5222, opened 2023-05-30
Adding Variadic Controller Arguments

lukaskleinschmidt:
## This PR …
Currently working on an update for a plugin where I previously used the `Kirby\Toolkit\Controller` to pass in arguments.
Thought this might be something for the core as well, as it does not break anything and allows for some advanced implementations without adding (🤞) any extra complexity. 

Edit:
On second thought this actually fixes some edge cases:

```php
$controller = new Controller(function ($c, ...$b) {
	// Current implementation
	// $c = 'C'
	// $b = ['B']

	// Should be
	// $c = 'C'
	// $b = ['A', 'B']
});

$controller->call(null, [
	'a' => 'A',
	'b' => 'B',
	'c' => 'C'
]);
```

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/Controller.php
+++ b/src/Toolkit/Controller.php
@@ -31,7 +31,7 @@ class Controller
 		foreach ($info->getParameters() as $param) {
 			$name = $param->getName();
 
-			if ($param->isVariadic()) {
+			if ($param->isVariadic() === true) {
 				$args += $data;
 			} else {
 				$args[$name] = $data[$name] ?? null;
--- a/src/Toolkit/Controller.php
+++ b/src/Toolkit/Controller.php
@@ -26,11 +26,19 @@ class Controller
 	public function arguments(array $data = []): array
 	{
 		$info = new ReflectionFunction($this->function);
+		$args = [];
 
-		return A::map(
-			$info->getParameters(),
-			fn ($parameter) => $data[$parameter->getName()] ?? null
-		);
+		foreach ($info->getParameters() as $param) {
+			$name = $param->getName();
+
+			if ($param->isVariadic()) {
+				$args += $data;
+			} else {
+				$args[$name] = $data[$name] ?? null;
+			}
+		}
+
+		return $args;
 	}
 
 	public function call($bind = null, $data = [])
--- a/tests/Toolkit/ControllerTest.php
+++ b/tests/Toolkit/ControllerTest.php
@@ -32,6 +32,19 @@ class ControllerTest extends TestCase
 	/**
 	 * @covers ::arguments
 	 */
+	public function testArgumentsOrder()
+	{
+		$controller = new Controller(fn ($b, $a) => $b . $a);
+
+		$this->assertSame('BA', $controller->call(null, [
+			'a' => 'A',
+			'b' => 'B'
+		]));
+	}
+
+	/**
+	 * @covers ::arguments
+	 */
 	public function testVariadicArguments()
 	{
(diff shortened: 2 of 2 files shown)


src/Toolkit/Controller.php
```diff
@@ -26,11 +26,19 @@ public function __construct(protected Closure $function)
...
+		$args = [];
 
-		return A::map(
-			$info->getParameters(),
-			fn ($parameter) => $data[$parameter->getName()] ?? null
-		);
+		foreach ($info->getParameters() as $param) {
+			$name = $param->getName();
+
+			if ($param->isVariadic()) {
```
lukasbestle:
```suggestion
			if ($param->isVariadic() === true) {
```

tests/Toolkit/ControllerTest.php
```diff
@@ -29,6 +29,20 @@ public function testArguments()
 		]));
 	}
 
+	/**
+	 * @covers ::arguments
+	 */
+	public function testVariadicArguments()
+	{
+		$controller = new Controller(fn ($c, ...$args) => $c . implode('', $args));
```
lukasbestle: To ensure that `$c` still receives its value correctly:

```suggestion
		$controller = new Controller(fn ($c, ...$args) => $c . '/' . implode('', $args));
```

tests/Toolkit/ControllerTest.php
```diff
@@ -29,6 +29,20 @@ public function testArguments()
...
+
+		$this->assertSame('CAB', $controller->call(null, [
```
lukasbestle:
```suggestion
		$this->assertSame('C/AB', $controller->call(null, [
```

lukasbestle (commented):
I like the idea and also implementation.

Only thought I had is that the name of the variadic argument no longer gets a meaning unlike with normal arguments. But it's probably a non-issue as it changes nothing about the current behavior.

lukaskleinschmidt:
The ues case is probably limited. For reference the origin of the idea:
https://github.com/lukaskleinschmidt/kirby-snippet-controller/issues/1#issuecomment-1567858154

lukasbestle (dismissed):
I like it. But would be good to have a second team opinion on the general concept.

distantnative:
Doesn't seem to impact the existing usage and I'd be fine to have this addition. My head doesn't quite parse how `$args += $data` just adds part of the array, e.g. `$c, ...$args` not also collecting `$c` in `$args`

lukaskleinschmidt:
The `$args += $data` only adds values from `$data` that are not already present in the `$args` array (by key).
It is similar to using `$args = array_merge($args, $data)` in this case as we work with string keys here.

Apparently it is called a union operation.
https://www.php.net/manual/en/language.operators.array.php

distantnative:
I think we are missing a test that checks whether we can use parameters no matter of their order, or am I just missing it? Would be good to ensure that the change doesn't mess with that.

lukaskleinschmidt:
An independend one from the two that test the arguments?

distantnative:
Well, `testArguments` just uses them in the order as they are provided. Would probably be interesting to test `fn( $b, $a)`

lukaskleinschmidt:
I could add another `testArgumentsOrder` test to the PR

distantnative (approved):

lukasbestle (approved):

bastianallgeier:
⏸ Merge after 3.9.5 has been released.

PR #5232, opened 2023-06-01
Query language support for pages section `create` and `templates` and files section `template` props

afbora:
## This PR …

Implements https://kirby.nolt.io/130 (59 upvotes)

By providing query language support, it was made to cover the this nolt idea and to make it a more flexible structure. 

````yaml
sections:
  pages:
    type: pages
    create: page.createTemplates()
    templates: site.availableTemplates()
  files:
    type: files
    template: site.availableTemplate()
````

### Features

- Query language support for pages section `create` and `templates` and files section `template` props https://kirby.nolt.io/130

### Breaking changes

- Pages and files sections won't allow anymore templates named `page`, `site`, `kirby`, `model` as value for the `create` and `template`/`templates` options

## Ready?

### For review team

### DIFF
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -2,6 +2,7 @@
 
 use Kirby\Cms\File;
 use Kirby\Cms\Files;
+use Kirby\Exception\InvalidArgumentException;
 use Kirby\Toolkit\I18n;
 
 return [
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -39,6 +39,23 @@ return [
 		}
 	],
 	'computed' => [
+		/**
+		 * Do not change the order of this alphabetically.
+		 * It must be processed before other computed calls to work correctly.
+		 */
+		'template' => function () {
+			$template = $this->template;
+
+			if ($query = $this->model->query($template)) {
+				if (is_string($query) === false) {
+					throw new InvalidArgumentException('Invalid `template` prop: query must return a string.');
+				}
+
+				$template = $query;
+			}
+
+			return $template;
+		},
 		'accept' => function () {
 			if ($this->template) {
 				$file = new File([
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -51,13 +51,59 @@ return [
 			return $status;
 		},
 		/**
+		 * Filters the list by single template.
+		 */
+		'template' => function (string $template = null) {
+			return $template;
+		},
+		/**
 		 * Filters the list by templates and sets template options when adding new pages to the section.
 		 */
 		'templates' => function ($templates = null) {
-			return A::wrap($templates ?? $this->template);
+			return $templates;
 		}
 	],
 	'computed' => [
+		'create' => function () {
+			$create = $this->create;
+
+			if (is_string($create) === true) {
+				if ($query = $this->model->query($create)) {
+					if (
+						is_string($query) === false &&
+						is_array($query) === false
+					) {
+						throw new InvalidArgumentException('Invalid `create` prop: query must return string or array.');
+					}
+
+					$create = $query;
+				}
+			}
+
+			return $create;
+		},
+		/**
+		 * Do not change the order of this alphabetically.
+		 * It must be processed before other computed calls to work correctly.
+		 */
+		'templates' => function () {
+			$templates = $this->templates ?? $this->template;
+
+			if (is_string($templates) === true) {
+				if ($query = $this->model->query($templates)) {
+					if (
+						is_string($query) === false &&
+						is_array($query) === false
+					) {
+						throw new InvalidArgumentException('Invalid `templates` prop: query must return string or array.');
+					}
+
+					$templates = $query;
+				}
+			}
+
+			return A::wrap($templates);
+		},
 		'parent' => function () {
 			$parent = $this->parentModel();
 
--- a/tests/Cms/Sections/FilesSectionTest.php
+++ b/tests/Cms/Sections/FilesSectionTest.php
@@ -738,4 +738,38 @@ class FilesSectionTest extends TestCase
 
 		$this->assertSame(4, $section->upload()['attributes']['sort']);
 	}
+
+	public function testQueryTemplate()
+	{
+		$app = $this->app->clone([
+			'options' => [
+				'template' => 'foo'
+			]
+		]);
+
+		$app->impersonate('kirby');
+
+		$model = new Page([
+			'slug'  => 'test',
+			'files' => [
+				['filename' => 'a.jpg'],
+				['filename' => 'b.jpg'],
+				[
+					'filename' => 'c.jpg',
+					'content' => [
+						'template' => 'foo'
+					]
+				],
+			]
+		]);
+
+		$section = new Section('files', [
+			'name'     => 'test',
+			'model'    => $model,
+			'template' => 'kirby.option("template")'
+		]);
+
+		$this->assertSame('foo', $section->template());
+		$this->assertCount(1, $section->files());
+	}
 }
--- a/tests/Cms/Sections/FilesSectionTest.php
+++ b/tests/Cms/Sections/FilesSectionTest.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Cms;
 
+use Kirby\Exception\InvalidArgumentException;
 use Kirby\Filesystem\Dir;
 use PHPUnit\Framework\TestCase;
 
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by distantnative** [comment]
Misses the breaking change:
- Pages and files sections won't allow anymore templates named `page`, `site`, `kirby`, `model` as value for the `create` and `template`/`templates` options

**Comment by afbora** [comment]
That's true. But it can be in a array. It needs to be expressed correctly.

````yaml
# Correct
templates:
  - page
  
# Invalid, Breaking change
templates: page

# Invalid, Breaking change
template: page
````

**Comment by distantnative** [comment]
@afbora maybe then write "as **string** value for the"

**Comment by afbora** [comment]
@distantnative To be honest, I didn't expect this breaking change. How can we prevent it? `page` by itself is not considered a query for this case. How about like that?

````php
// detect prop is query language
if (Str::contains($template, '.') === true) {
  if ($query = $this->model->query($template)) {
    // todo
  }
}

**Comment by distantnative** [comment]
@afbora would that still work if we have an actual template name string that has a dot? Cause in theory a template name could have a dot in it

**Comment by afbora** [comment]
It will not work :( Is there any other way to tell if a string is a query or not?

I'm thinking out loud. What if we apply a syntax like in `select` field?

````yaml
template:
  type: query
  query: site.availableTemplate

template:
  type: api
  url: "{{ site.url }}/my-api/templates.json"
````

**Comment by distantnative** [comment]
@afbora Maybe it would work if we don't throw an error if the query result isn't a string. Then the code would try to parse it as a query but likely produce nothing or not a string. And in that case instead of throwing an error, we could just fall back at using the original string as template name.

**Comment by lukasbestle** [comment]
That could come with quite a performance hit when no query is used, no?

**Comment by distantnative** [comment]
@lukasbestle Certainly a few calls made, but not sure anything significant. It's a trade off between supporting queries for these options or not, I guess.

**Comment by lukasbestle** [comment]
I quite like the idea to make it:

```yml
template:
  query: ...
```

Then we can reliably detect if it's a query or not. We don't even need the `type` here, at least for now. So we just need to check if it's an array that has the `query` key.

distantnative:
I am afraid we are diving into territories where users don't understand the complex blueprint options anymore with every level we add. Many are already struggling. So why in one case I do directly use a query as string, in another I had to use `" {{ }}"` and now then in even other cases I have to include a `query` suboption?

distantnative:
Would it be an option to require `"{{ }}"` here when using a query and qe work on `->query()`/`->toString()` or a new method to support parsing these templates but allowing not just a string as result but also e.g. an array (if these is nothing else in the string template but the `{{ query }}`)

afbora:
I leave it to you to decide whether to close this PR or how to proceed.

lukasbestle:
@afbora What do you think about Nico's last suggestion? I'd say the `{{ }}` syntax would indeed be clearer. Then one could do:

```yaml
# just one fixed template
templates: article

# fixed list
templates:
  - article
  - recipe

# query (can return a string or array)
templates: "{{ site.availableTemplates }}"
```

afbora:
I'm not sure, I find the idea a bit confusing, and it doesn't reflect the Kirby style. It seems more like a temporary solution or hack to me. Because the developer might think that he can combine multiple queries. How do we tell them to use the `{{ }}` syntax here? Even though we are using the `{{ query }}` syntax, a developer using Kirby would know that in this case they should use `template: query` in this case, and a developer using Kirby would think and know to use `template: "{{ query }}"` might be unfamiliar. 

Since you know the big picture better, I leave the decision to you.

lukasbestle:
I'd say nothing speaks against using multiple queries in such a string. Each query will be evaluated as normal and their results will be concatenated to a single string. Of course this will result in an error if one of the queries returns anything that's not a string.

So the logic of that new "query method with support for non-strings" would be:

- if the query only contains a single query, starts with `{{` and ends with `}}`, it will return exactly what the query returns
- otherwise, it needs to check if each query result is a string. If yes: behave exactly like `$model->toString()`. If no: exception.

For the use cases in this PR (`create`/`templates` props) it of course doesn't make sense to use multiple queries in a single string, but at least we'll have consistent syntax here. And with the logic above, we will be prepared for other use cases in the future that will then also be consistent.

afbora:
Thank you @lukasbestle for the explain 👍 

Okay, critical question;

If the developer does not have a reserved template name such a `kirby`, `site`, `page`, etc; will they be able to use the `templates: query` syntax? Or does it always have to use `templates: "{{ query }}"`?

A new idea, I'm thinking out loud, it might be stupid.

For the future, do you think it would be silly or unnecessary if we had a prefix like `query://kirby.option('debug')` for all queries?

lukasbestle:
I'd say a query would only be executed when the `{{ }}` syntax is used. A string without curly braces will run through the query method, but it will be returned 1:1 because no query to execute will be found.

A prefix could be a possible way as well. It would also fit better to YAML syntax as it would not be necessary to put the query in quotes. But TBH I don't really like the URI scheme as prefix, maybe we can find something else.

afbora:
@distantnative What do you think about that?

bastianallgeier:
I'm only watching fromt the sidelines, but I find it really difficult to make a good decision here. I agree with @afbora that it feels weird to use the curly brackets, but then don't offer full templates. But I also don't feel comfortable about the prefix. 

What if we think in terms of a blueprint builder UI? How would the configuration for the section look there? We would probably need some kind of toggle that switches between two different input types 

A/B Toggle

A) input for a list of templates
B) input for a query

Could it maybe simplfiy the entire problem to use a completely different option name? 

```yaml
templates: 
- a
- b
- c
```

```yaml
templatesQuery: page.myTemplates
```

It's more verbose and adding such a new option would mean that it overwrites the templates option. But this is like in Vue components when you define a render method that automatically overwrites templates.

afbora:
Hmm, personally I still can't say "this is the right solution", considering that the main point of feedback is "Exclude templates": https://kirby.nolt.io/130. Maybe we can find a different and more accurate solution.

bastianallgeier:
When all of this turns out to be super complicated, maybe we need to get back to the drawing board. I like our queries, but I'm not sure if it's a good decision to automatically throw them at everything. 

From a user / developer perspective, it would probably be a lot easier to have a simple exclude syntax. 

```yaml
templates: 
  - !home
  - !error
```

afbora:
The `!home` syntax is easy for the user, but not so easy to implement. The `templates` prop is for selecting anyway. It's confusing to both include and exclude. For example, what does the following mean?

````yaml
templates:
  - default
  - contact
  - !home
  - !error
````

As you can see, this is an unreasonable select. So how do we explain its use/syntax (only exclude should be used) to developers?

**Comment by bastianallgeier** [comment]
Yep, that's true. It would be either a list with or a list without exclamation marks. Both does not work. That would probably speak for something like: 

```yaml
templatesIgnore: 
  - home
  - error
```

afbora:
Then this will be the same as the `templatesQuery` prop we talked about in PR. Which one do you think we should prefer? `templatesQuery` or `templatesIgnore` or both?

lukasbestle:
I'd say both are useful on their own, especially in a blueprint editor context where one could toggle between the modes.

However with all of them being present in the blueprint syntax, it could be hard to document without huge complexity.

bastianallgeier:
I also think that both would be super confusing. I tend to lean towards the simpler one. The query option is really just usable for those who are willing to build models or page methods and I think that's already quite a high obstacle.

afbora:
Please vote with emojis for the options @getkirby/kirby-staff 

- 👍 Only query feature `templatesQuery`
- ❤️ Only ignore feature `templatesIgnore`
- 🚀 Both features

afbora:
I've got another idea. How about combining `query` and `ignore/exclude`? @bastianallgeier 

### Array option
````yaml
templatesIgnore: 
  - home
  - error
````

### String option as query
````yaml
templatesIgnore: page.myTemplates
````

### Invalid uses
````yaml
templatesIgnore: page

templatesIgnore: 
  - page.myTemplates
  - site.myTemplates
````

bastianallgeier:
@afbora but then the template option would be expected to work in the same way, wouldn't it?

afbora:
Oh, I'm sorry. There's a point I'm missing. It would not be nice for developers to have a query option to ignore but not to select.

Can you vote for that? https://github.com/getkirby/kirby/pull/5232#issuecomment-1597010730

afbora:
I'm closing this PR favor of implementation of `templatesIgnore` instead query language support.

PR #5234, opened 2023-06-01
Fix type hints for `search` component

### LINKED ISSUE #5233: [v4-alpha.2] Searching broken on collections
## Description

Getting following error while searching on collection:

````php
collection('products')->search('search text', 'title|summary|text');
````

> Kirby\Cms\Search::collection(): Argument #3 ($params) must be of type array, string given, called in \kirby\src\Cms\Collection.php on line 354

Related commit: https://github.com/getkirby/kirby/commit/99782e74cb032e53d7f877331e59e75c0f025c11#diff-1e3b7324cb1941470f3429a3a5fc4ed5f94cf4eae6fcbf6af94dcbca6d50008a

### Possible solution

3rd argument type should be `mixed` or `string|array` on following line:
https://github.com/getkirby/kirby/blob/4.0.0-alpha.2/src/Cms/Search.php#L32

### PR DESCRIPTION
**Pull request by distantnative**
Fixes #5233

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -138,11 +138,13 @@ return [
 	 *
 	 * @param \Kirby\Cms\App $kirby Kirby instance
 	 * @param \Kirby\Cms\Collection $collection Collection of searchable models
-	 * @param string $query
-	 * @param mixed $params
-	 * @return \Kirby\Cms\Collection|bool
 	 */
-	'search' => function (App $kirby, Collection $collection, string $query = null, $params = []) {
+	'search' => function (
+		App $kirby,
+		Collection $collection,
+		string $query = null,
+		string|array $params = []
+	): Collection {
 		// empty search query
 		if (empty(trim($query ?? '')) === true) {
 			return $collection->limit(0);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by afbora — COMMENTED** [review]
What about following line?

https://github.com/getkirby/kirby/blob/v4/fix/5233-search-collection/src/Cms/Collection.php#L352

File: config/components.php [hunk 1]
```diff
@@ -138,11 +138,13 @@
...
-	 * @param string $query
-	 * @param mixed $params
-	 * @return \Kirby\Cms\Collection|bool
 	 */
-	'search' => function (App $kirby, Collection $collection, string $query = null, $params = []) {
+	'search' => function (
+		App $kirby,
+		Collection $collection,
+		string $query = null,
+		string|array $params = []
```
afbora, line 146: Just to be sure, this change will not break existing search component installations, right?

distantnative: Changing the type hinting on the default component shouldn't change anything for custom components

afbora (approved):

PR #5239, opened 2023-06-03
Page tree: use blueprint icons

distantnative:
### Enhancement
- Page tree now uses the icon defined in each page's blueprint
- 
Screenshot 2023-06-03 at 13 07 34

### DIFF
--- a/panel/src/components/Navigation/PageTree.vue
+++ b/panel/src/components/Navigation/PageTree.vue
@@ -52,7 +52,7 @@ export default {
 	methods: {
 		async load(path) {
 			const { data } = await this.$api.get(path + "/children", {
-				select: "hasChildren,id,title,uuid",
+				select: "hasChildren,id,panelImage,title,uuid",
 				status: "all"
 			});
 
@@ -63,6 +63,7 @@ export default {
 
 				pages[id] = {
 					id,
+					icon: page.panelImage.icon,
 					label: page.title,
 					hasChildren: page.hasChildren,
 					children: "/pages/" + this.$api.pages.id(page.id),


afbora (approved):
Looks amazing 💛

bastianallgeier:
The emojis seem to be cut off for me. I also wonder if we should replace our native icons with emojis for a more consistent style. 
Screenshot 2023-06-03 at 15 32 31

distantnative:
I don't think we should replace them - I think there is quite did. Free rent fractions in our user base preceding icons vs. emojis.

distantnative:
@bastianallgeier might your issue be related to `fix emoji alignment on high-res screens` in `k-icon`? Looks totally fine for me, I don't think this should block merging this PR.

Screenshot 2023-06-03 at 15 42 54

bastianallgeier:
Oh yes, that could totally be the issue. I'm still kind of not super happy about the alignment and the mixture of icons to be honest. But on the other hand the emojis really improve visual navigation

distantnative:
I mean the mixture totally just depends on how the user configures their blueprints

bastianallgeier:
Ok, the emojis are indeed fine on the retina display. That's so weird again. I just checked out Notion again and their mixture of emojis and their icons isn't any better either. So I'm fine with it, but we need to fix the alignment first. 

Screenshot 2023-06-03 at 15 47 38

distantnative:
@bastianallgeier But I would think this has nothing to do with this PR but is an issue of `k-icon` in general? Maybe exactly because of the quoted line, increasing the font size of emojis?

bastianallgeier:
You could be right about that. k-icon should give a consistent box for emojis and icons.

bastianallgeier:
It's fixed if I set the width and height of the k-icon span

distantnative:
@bastianallgeier it would/will become a non-issue on that one other branch we have

bastianallgeier:
Do you think it's a problem to already fix it in this PR? The fix is pretty minimal

distantnative:
@bastianallgeier feel free :)

bastianallgeier:
I felt free, but then realized that it's the same weird center alignment issue on different resolutions. This emoji stuff is driving me insane.

PR #5241, opened 2023-06-03
Block: use `panel.drawer`

distantnative:
Using the nested block setup from https://staging.getkirby.com/releases/4.0/block-field-improvements#field-preview

## Known issues
- [ ] Adding a `code` block to the `Blocky` field in the fields block results in error because the `endpoint.field` is undefined (bug not present on `develop` branch; also working when doing the same inside the drawer)
- [ ] Navigating into the 2nd level of nested block drawer works, but when confirming/submitting the 2nd level drawer it throws a `Cannot read properties of undefined (reading 'focus')` error
- [ ] Reactivity doesn't seem to work through multiple layers: changes in the parent drawer don't show up in the fields block preview. Nor do they sometimes trigger the FormButtons changes bar.

### DIFF
--- a/panel/src/components/Forms/Blocks/Block.vue
+++ b/panel/src/components/Forms/Blocks/Block.vue
@@ -53,47 +53,6 @@
 			}"
 		/>
 
-		<k-form-drawer
-			v-if="isEditable && !isBatched"
-			:id="id"
-			ref="drawer"
-			:icon="fieldset.icon || 'box'"
-			:tabs="tabs"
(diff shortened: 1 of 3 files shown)


distantnative:
Narrowed down a first issue:
https://github.com/getkirby/kirby/blob/v4/fix/block-drawer/panel/src/panel/drawer.js#L51-L54
`this.path` is always `null` for the block drawer. Not 100% sure what it should be. (But that's why the parent isn't set)

PR #5243, opened 2023-06-04
PageCreateDialog: fix supported field types

distantnative:
Excluding `textarea` and `writer` for now until we have a better idea how to handle them.

### DIFF
--- a/src/Panel/PageCreateDialog.php
+++ b/src/Panel/PageCreateDialog.php
@@ -30,19 +30,21 @@ class PageCreateDialog
 		'date',
 		'email',
 		'info',
-		'multiselect',
-		'number',
+		'line',
+		'link',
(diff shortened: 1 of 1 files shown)


afbora:
### Panel UI

screen-capture-1593-Site - Mægazine-localhost

### Console

> Field type "writer" not supported in create dialog

### Expected

The error should be visible for users.

distantnative:
@bastianallgeier do you know why the exception doesn't end up in the dialog error notification?

bastianallgeier:
I think there's still an API response case that is not correctly handled in our error handling / notification logic. I came across that myself recently.

distantnative:
Should we merge this unrelated?

afbora:
Should I create new issue then?

PR #5246, opened 2023-06-04
`Toolkit\Str`: add type hints for replace methods

distantnative:


src/Toolkit/Str.php
```diff
@@ -712,14 +727,12 @@ public static function random(
...
-	 * @todo the types aren't correct, refactor to apply native type hinting
 	 */
 	public static function replace(
-		$string,
-		$search,
-		$replace,
-		$limit = -1
+		string|array|Collection $string,
+		string|array|Collection $search,
+		string|array|Collection $replace,
```
afbora, line 733: Also needs to update doc-block for Collection type.

afbora (approved):
LGTM 👍

PR #5259, opened 2023-06-07
`->toUrl()` to return `null` on empty fields

### LINKED ISSUE #5258: [v4] `toUrl()` of empty link field returns page url
## Description
When using `toUrl()` on an empty link field, the page url is returned. This is bad because you might want to check if the link is valid by using something like this:

```php
<?php if($url = $block->link_field()->toUrl()): ?>
  <a href="<?= $url ?>">My link</a>
<?php endif ?>
```

**Expected behavior**  
I'd expect the `toUrl()` method to return `false` when the field is empty. This way the code example above would not result in an a tag at all.

## To reproduce

1. Add a link field and leave it empty
2. Return the field with the `toUrl()` method

## Your setup
**Kirby Version**  
4.0.0-alpha.2

distantnative:
- `->toUrl()` to return `null` on empty fields
#5258

### DIFF
--- a/config/methods.php
+++ b/config/methods.php
@@ -286,12 +286,9 @@ return function (App $app) {
 
 		/**
 		 * Turns the field value into an absolute Url
-		 *
-		 * @param \Kirby\Cms\Field $field
-		 * @return string
 		 */
-		'toUrl' => function (Field $field): string {
-			return Url::to($field->value);
+		'toUrl' => function (Field $field): string|null {
+			return $field->value ? Url::to($field->value) : null;
 		},
 
 		/**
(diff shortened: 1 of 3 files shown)


config/methods.php
```diff
@@ -286,12 +286,9 @@
...
 		/**
 		 * Turns the field value into an absolute Url
-		 *
-		 * @param \Kirby\Cms\Field $field
-		 * @return string
 		 */
-		'toUrl' => function (Field $field): string {
-			return Url::to($field->value);
+		'toUrl' => function (Field $field): string|null {
+			return $field->value ? Url::to($field->value) : null;
```
bastianallgeier: Maybe we should do an empty check here.

bastianallgeier (approved):

PR #5260, opened 2023-06-07
Color field: fix flex spacing

### LINKED ISSUE #5253: [v4] Color field: odd placement of 6+ options
## Description
I love the new color field! Today I noticed when adding more than 5 options to a color field, the options in the second line are oddly placed.

**Expected behavior**  
I'd expect the options to be left aligned so they're directly under the options of the first line. Or to put it differently, I'd expect `justify-content` it to be `start` instead of `space-around`. Personally I also find a single line of e.g. 3 boxes better left aligned. But you probably had a reason to use `space-around`?

**Screenshots**  

Color field with 7 options

## To reproduce

1. Add a color field to your blueprint
2. Add more e.g. 7 options to it
3. Have a look at the Panel

## Your setup
**Kirby Version**  
Kirby 4.0.0-alpha.2

distantnative:
Fixes #5253

### DIFF
--- a/panel/src/components/Forms/Element/Color/color.css
+++ b/panel/src/components/Forms/Element/Color/color.css
@@ -1,3 +1,9 @@
+:root {
+	--color-preview-rounded: var(--rounded-sm);
+	--color-preview-size: 1.5rem;
+	--color-preview-darkness: 0%;
+}
+
 .k-color {
 	--h: 0;
 	--s: 0%;
@@ -22,8 +28,7 @@
 
 .k-color .k-coords {
 	aspect-ratio: 1/1;
-	background:
-		linear-gradient(to bottom, transparent, #000),
+	background: linear-gradient(to bottom, transparent, #000),
 		linear-gradient(to right, #fff, hsl(var(--h), 100%, 50%));
 }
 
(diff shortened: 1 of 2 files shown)


bastianallgeier:
Wouldn't this be a better fit for grid without a magic number to determine the number of items?

distantnative:
Wouldn't we still need to do the math to make the grid the right size and not too narrow or too wide? I mean, it's not really magic, just math 😂

bastianallgeier:
What I mean with magic is that you have to add a comment to the code to understand why it is 0.6rem. It is not self-explainatory until you do the math. If you use grid, we will always head back into this later and see that we want to have 6 items per row.

distantnative:
Tbh I really don't see the benefit over a grid saying it's 6 vs. a comment saying it's 6.

But feel free to change it.

bastianallgeier:
I changed the layout to grid and also made the preview sizing a bit more flexible. 

Then I used this chance to add a faded outline over the color swatch, which blends with the color. This looks a lot better than mixing colors with a gray border. 

Screenshot 2023-06-08 at 13 09 25

bastianallgeier (approved):

PR #5264, opened 2023-06-10
No redirect to logout when already logged out

### LINKED ISSUE #5263: [v4] Infinite reload with Custom Block Selector Plugin installed
## Description
With the [Kirby Visual Block Selector Plugin](https://github.com/junohamburg/kirby-visual-block-selector) installed, the panel reloads indefinitely.

## To reproduce
Install Kirby 4 Beta & the linked plugin above.

## Your setup
**Kirby Version**  v4.0.0-alpha.2

distantnative:
Avoid redirect loops

@bastianallgeier not sure if the second part is a correct fix to our blank error dialogs issue

### DIFF
--- a/panel/src/panel/notification.js
+++ b/panel/src/panel/notification.js
@@ -56,7 +56,12 @@ export default (panel = {}) => {
 		 */
 		error(error) {
 			if (error instanceof AuthError) {
-				return panel.redirect("logout");
+				// only redirect to logout if panel state actually
+				// assumes the user is authenticated, otherwise
+				// continue to show a normal error notification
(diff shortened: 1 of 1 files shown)


bastianallgeier:
@distantnative I think we could potentially lose stuff in the error object this way. I.e. the code.

distantnative:
But without it our panel shows empty dialogs etc. it doesn’t seem able to handle actual error instances

bastianallgeier (approved):
I've looked at it again, and we might indeed find out that we miss keys from the object later. But it's no problem to add them here.

PR #5283, opened 2023-06-13
Content storage (backend PR 1)

lukasbestle:
### Breaking changes

- Extending the internal `$model->contentFile()`, `$model->contentFiles()`, `$model->contentFileDirectory()`, `$model->contentFileExtension()`, `$model->contentFileName()`, `$model->readContent()` and `$model->writeContent()` methods in a page model will no longer have an effect as these methods are no longer called by the core. Please extend the new `Kirby\Content\PlainTextContentStorage` class instead and return an instance of your custom class from `$model->storage()`. However please note that the interface of `PlainTextContentStorage` is internal and may change in the future.
- The `$translations->start()` and `$translations->stop()` methods were no longer in use and have been removed. Content files are automatically converted by the `Language` class.

### Deprecated

- The internal `$model->contentFile()`, `$model->contentFiles()`, `$model->contentFileDirectory()`, `$model->contentFileExtension()` and `$model->contentFileName()` methods have been deprecated and will be removed in Kirby 5.

### Refactoring

- New `Content` namespace with the `Content`, `ContentTranslation` and `Field` classes (compatibility aliases are provided for now)
- Improve code quality and tests in the `Content` namespace
- New internal `ContentStorage` classes to bundle all low-level content logic in one place

## Ready?

### For review team


distantnative:
This is great work! 👏 Some first observations:

- First 3 commits look just fine 👍
- Probably controversial: `VersionIdentifier` and `VersionTemplate` I am not a fan tbh. I don't fully understand their purpose, but they generate a lot of complexity in the code/make it hard to read and understand. And I don't quite think that's necessary. I am afraid, we're trying to take too many steps at once here. For what we want to achieve with v4, I think we should be using simple string values for this first big step this summer. Even knowing the whole concept, it's so hard to understand what they're meant for. I think we will be better off next year when native Enums would also be available for us. And when the changes backend code has had time to mature - we'll also have made some learnings that will inform the best way then.
- Then again, I am a big fan of the `ContentStorage` class and surroundings. Or actually the interface - but there's my question whether this should be rather an abstract class? Which would be more common for Kirby.
- `PlainTextContentStorage:: contentFile()` hardcodes the behavior for the different model types. Which would make it very difficult to extend this for developers. Not sure if an issue, but made me stop when reading.
- We shouldn't just deprecate the old methods, but ideally also change those methods to directly use the new code, so to be just pricy calls to the new class methods, e.g. `File:: contentFileDirectory()` still keeps the old code

I think it makes sense to first talk about these broader thoughts. Before diving into more detail comments, e.g. especially naming I find sometimes hard to understand. I think there we can help as a team to find terms that make sense for all/most of us.

lukasbestle:
Thanks for your review.

I've written you directly on Discord about the VersionIdentifier/Template classes.

About making an abstract class: I've decided against it as the only method we could predefine would be the constructor. Writing unit tests that cover that one line would be a pain as we would need to define a mock class with all the other methods. There wouldn't be a real benefit in the real world that outweighs this, so an interface seemed more fitting. I see it a bit the other way around: Maybe we should use interfaces more in the future where they make sense.

Hardcoded model types: I had discussed this with Basti a while back in Notion and we came to the conclusion that it's a necessary change to avoid massive complexity. There are currently no other model types with storage and if a dev really wants to override the behavior, they can extend the storage handler class and return their custom instance from the model class.

Change old methods: That's what I've done in the places where it makes sense, e.g. `ModelWithContent::contentFile()`. With methods like `contentFileDirectory()` it's a bit different as those have always been `@internal` implementation details and called from nowhere else, so I don't see the benefit of actively proxying them. Especially because what they do is so simple and won't change until those methods are finally removed in v5.

distantnative:
Thanks for the explanations. They clear my doubts/concerns for these.

Shall we bring the versionidentifier/template discussion here? I don’t think that’s something bilateral to discuss but concerns the whole team. I think your main point was type safety there. I’m trying to be open to this, so let’s say we don’t go with simple strings. But…

- While VersionIdentifier I can vaguely guess from the name what it does, VersionTemplate just leaves question marks above my head. I think here we should work on terminology at least.
- We have those specific classes but no Content\Version class, which seems odd when trying to understand how everything relates. Wouldn’t it make sense to rather combine these other classes into a single class that defines a Version?

I think this debate is important also looking ahead to PR 2 where currently classes named ModelVersion and ContentVersions shall be introduced. I think we really run danger of a complexity that we ourselves won’t understand a few months down the road because the names aren’t really guiding understanding.

lukasbestle:
Yes, I think naming is something we can indeed work on. Maybe let's have a short phone meeting to work on this? Probably quicker than comment ping pong.

Combining those classes into one wouldn't work, again for type safety. We need the difference between "existing version" (identifier) and "new version" (template), so they need to be independent classes.

src/Cms/File.php
```diff
@@ -452,7 +459,10 @@ public function modified($format = null, string $handler = null, string $languag
 	 */
 	protected function modifiedContent(string $languageCode = null): int
```
distantnative, line 459: Shouldn't we be able to specify for which content (version identifier) we want the timestamp?

src/Cms/Language.php
```diff
@@ -603,19 +642,36 @@ public function update(array $props = null): static
...
-				touch($page->contentFile($code));
+				if ($page->storage()->exists(VersionIdentifier::changes(), $this) === true) {
+					$page->storage()->touch(VersionIdentifier::changes(), $this);
+				}
+				if (
+					$page->isDraft() === false &&
+					$page->storage()->exists(VersionIdentifier::published(), $this) === true
+				) {
+					$page->storage()->touch(VersionIdentifier::published(), $this);
+				}
```
distantnative, line 673: Based on these calls, I wonder if we couldn't integrate that check into `Storage::touch` directly. Can only be touched if it actually exists, if it doesn't just don't do anything inside `Storage::touch`

src/Cms/ModelWithContent.php
```diff
@@ -42,6 +45,7 @@ abstract class ModelWithContent implements Identifiable
 	public Content|null $content;
 	public static App $kirby;
 	protected Site|null $site;
+	protected ContentStorage|null $storage;
```
distantnative: Should we rather start working without `null` and then checking for `isset()` on properties?

src/Cms/ModelWithContent.php
```diff
@@ -169,59 +173,45 @@ public function content(string $languageCode = null)
...
-		if ($this->kirby()->multilang() === true) {
-			if ($language = $this->kirby()->languageCode($languageCode)) {
-				return $directory . '/' . $filename . '.' . $language . '.' . $extension;
-			}
+		Helpers::deprecated('The internal $model->contentFile() method has been deprecated', 'model-content-file');
 
-			throw new InvalidArgumentException('Invalid language: ' . $languageCode);
-		}
+		$identifier = $this::CLASS_ALIAS === 'page' && $this->isDraft() === true ?
+			VersionIdentifier::changes() :
```
distantnative: Do I understand this correctly that going forward drafts won't stay in `_drafts` anymore, but just treated like `_changes`? Or maybe that's handled in the classes (I have trouble keeping an overview), but then shouldn't we better have `VersionIdentifier::draft()` to help ourselves when reading code?

src/Cms/ModelWithContent.php
```diff
@@ -169,59 +173,45 @@ public function content(string $languageCode = null)
...
 	public function contentFiles(): array
 	{
-		if ($this->kirby()->multilang() === true) {
-			$files = [];
-			foreach ($this->kirby()->languages()->codes() as $code) {
-				$files[] = $this->contentFile($code);
-			}
-			return $files;
-		}
+		Helpers::deprecated('The internal $model->contentFiles() method has been deprecated', 'model-content-file');
```
distantnative: Can you please add to all deprecation messages what to use instead? That would also be helpful for me to understand what replaces what now.

src/Cms/ModelWithContent.php
```diff
@@ -443,7 +444,10 @@ public function kirby()
 	 */
 	public function lock()
 	{
-		$dir = $this->contentFileDirectory();
+		$dir = $this->root();
+		if ($this::CLASS_ALIAS === 'file') {
+			$dir = dirname($dir);
+		}
```
distantnative, line 448: Shouldn't this be then `$this->storage()->contentFileDirectory()` or so? We shouldn't duplicate logic, even if it's just a few lines.

src/Cms/ModelWithContent.php
```diff
@@ -524,16 +528,21 @@ public function query(string $query = null, string $expect = null)
...
+		try {
+			return $this->storage()->read(
+				$identifier,
+				$this->storage()->languageCodeToObject($languageCode)
+			);
+		} catch (NotFoundException $e) {
+			// only if the content file really does not exist, it's ok
+			// to return empty content. Otherwise this could lead to
+			// content loss in case of file reading issues
 			return [];
```
distantnative, line 542: Could we move this try-error inside the `Storage::read` method? Would probably be beneficial for any call to it, wouldn't it?

src/Cms/ModelWithContent.php
```diff
@@ -842,9 +860,24 @@ public function uuid(): Uuid|null
...
+
+		$identifier = $this::CLASS_ALIAS === 'page' && $this->isDraft() === true ?
+			VersionIdentifier::changes() :
+			VersionIdentifier::published();
+
+		// we can only update if the version already exists
+		if ($this->storage()->exists($identifier, $language) === true) {
+			$this->storage()->update($identifier, $language, $data);
+			return true;
+		}
```
distantnative, line 874: I feel like `Storage::update()` should take care of this check itself, really ensuring that it never updates non-existing versions

src/Content/ContentStorage.php
```diff
@@ -0,0 +1,85 @@
...
+	 */
+	public function touch(VersionIdentifier $version, Language $lang): void;
+
+	/**
+	 * Updates the content fields of an existing version
+	 *
+	 * @param \Kirby\Cms\Language $lang Language with code `'default'` in a single-lang installation
+	 * @param array<string, string> $fields Content fields
+	 *
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
```
distantnative, line 81: I think we could reduce complexity if these methods all handle the `non existing` case themselves instead of throwing exceptions.

distantnative: I also don't think `PlainTextContentStorage` is throwing them all currently.

src/Content/PlainTextContentStorage.php
```diff
@@ -0,0 +1,289 @@
...
+	/**
+	 * Shared helper method as adapter between core methods that currently
+	 * still take the language code and this content storage class that
+	 * takes a language object even in single-lang
+	 * @internal
+	 * @todo not needed anymore when the core is always-multi-lang
+	 *
+	 * @param bool $force If set to `true`, the language code is not validated
+	 */
+	public function languageCodeToObject(string|null $languageCode = null, bool $force = false): Language
```
distantnative:
```suggestion
	public function language(string|null $languageCode = null, bool $force = false): Language
```

src/Content/ContentStorage.php
```diff
@@ -0,0 +1,85 @@
...
+ * @package   Kirby Content
+ * @author    Lukas Bestle <lukas@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ */
+interface ContentStorage
+{
+	public function __construct(ModelWithContent $model);
+
```
distantnative, line 24: Should the interface also require `::language()`?

src/Content/PlainTextContentStorage.php
```diff
@@ -0,0 +1,289 @@
...
+				break;
+			case 'site':
+			case 'user':
+				$filename = $this->model::CLASS_ALIAS;
+				break;
+			default:
+				// @codeCoverageIgnoreStart
+				throw new LogicException('Cannot determine content filename for model type "' . $this->model::CLASS_ALIAS . '"');
+				// @codeCoverageIgnoreEnd
+		}
```
distantnative:
```php
		$directory = match ($this->model::CLASS_ALIAS) {
			'file'  => dirname($this->model->root()),
			default => $this->model->root()
		};

		$filename = match ($this->model::CLASS_ALIAS) {
			'file'  => $this->model->filename(),
			'page'  => $this->model->intendedTemplate()->name(),
			'site',
			'user'  => $this->model::CLASS_ALIAS,
			// @codeCoverageIgnoreStart
			default => throw new LogicException('Cannot determine content filename for model type "' . $this->model::CLASS_ALIAS . '"')
			// @codeCoverageIgnoreEnd
		};
```

For me this would read more easily, directly seeing which variable ends with what content.

src/Content/PlainTextContentStorage.php
```diff
@@ -0,0 +1,289 @@
...
+	 * @internal To be made `protected` when the CMS core no longer relies on it
+	 */
+	public function contentFiles(VersionIdentifier $version): array
+	{
+		if ($this->model->kirby()->multilang() === true) {
+			$files = [];
+			foreach ($this->model->kirby()->languages() as $lang) {
+				$files[] = $this->contentFile($version, $lang);
+			}
+			return $files;
```
distantnative: I think this can be reduced to

```php
return $this->model->kirby()->languages()->values(
	fn ($lang) => $this->contentFile($version, $lang)
);
```

src/Content/PlainTextContentStorage.php
distantnative: I think this would be leaner, self-explanatory and more in line with naming in other Kirby classes.

distantnative (commented):
First batch of thoughts and comments.

I thought through the breaking changes etc. and agree that this is probably the best way to go forward with it.

Some other general thoughts:
- Still not really convinced about `VersionIdentified` and `VersionTemplate`. It adds a new paradigm to Kirby that will be pretty confusing not just for developers but ourselves going forwards. Right now they really only represent two string values, which we usually just handle as normal strings (eg. page status). I would really appreciate if we could skip at least VersionTemplate for now and add it in a future version when we tackle revisions if it is absolutely necessary for these. For now it just bloats up the complexity.
- Many (all?) storage units receive an identifier and/or language object as parameters.  Wouldn't it make the code easier to read and understand when we can do something like `$this->storage($identifier, $language)->read()` as it points our that the storage depends on those two (there is a specific storage tied to these two parameters) and less the individual methods?
- I miss that we treat the default language as just another ContentTranslation. That might be part of PR 2, though I am not sure based on the Notion concept. I added a comment on Notion.

Sorry if some/many comments lack understanding, it's quite hard to review this PR tdue to its size but also its isolation. I tried my best, but probably overlooking some things when I make suggestions.

distantnative:
Can I make one more suggestion: I think it would be good if we don’t directly merge this into `v4/develop` but e.g. `v4/changes` first. I think this kind of separation was really good with other big refactoring parts we did early on for v4

src/Cms/File.php
lukasbestle: Ideally yes. I had thought about adding a fourth argument to `$file->modified()` to control this. We have the same issue with `$page->modified()`, `$user->modified()`, `$model->readContent()` and `$model->writeContent()`.

But in the end it didn't feel right to me to do add more arguments, especially to public methods when the content storage stuff is all still internal and we haven't had the chance to try out the interface in practice.

src/Cms/Language.php
lukasbestle: So far one of my design goals of the `ContentStorage` interface was to make it as strict as possible to make it as easy as possible to catch bugs from conflicting logic. I suggest we keep it like this for now. Once everything has settled, we can reduce the strictness again where it makes sense.

But TBH I'm not sure if it's a good idea to silently "not do anything" in an error case. Dropping too many exceptions or silently correcting method input has bitten us quite often in the past.

src/Cms/ModelWithContent.php
lukasbestle: Good point. 👍

Looks like the `??=` operator also works with non-set props, so I just removed the `|null` type.

src/Cms/ModelWithContent.php
lukasbestle: Not quite. The folder structure doesn't change. Drafts are still stored in the `_drafts` dir of the parent model.

But internally, a draft is always treated as a "changes" version. This is to ensure consistency in the implementation of the changes lifecycle. A draft should work exactly the same UX-wise as the changes of a published page, so both are a changes version.

Meaning: A draft only ever has a changes version, a published page can have both a published and a changes version (a published version always, a changes version only if changes exist).

src/Cms/ModelWithContent.php
lukasbestle: There are no 1:1 replacements for most of them, they will be removed completely in v5. As all methods were previously internal, I don't expect anyone to have used them. The deprecation warnings were added out of "extra caution" to catch potential issues with sites/plugins easier.

The only methods that have replacements are `contentFile()` and `contentFiles()`. They are now `storage()->contentFile()` and `storage()->contentFiles()`. But those are also internal and may be removed in a future major version, so I'd like to avoid pointing site and plugin devs to them in the deprecation warning.

src/Cms/ModelWithContent.php
lukasbestle: This whole method will be gone in v4 (or at least deprecated) because the old locking system will no longer be needed. So it's only a temporary change to keep the system working after merging this PR.

src/Cms/ModelWithContent.php
lukasbestle: TBH I disagree. From my comment above:

> I'm not sure if it's a good idea to silently "not do anything" in an error case. Dropping too many exceptions or silently correcting method input has bitten us quite often in the past.

Here in this case we return an empty array for backwards-compatibility, but that may not be wanted in every case. Or actually it may not be wanted at all, we just have to do it here to avoid breakage.

Keep in mind that `readContent()` is basically deprecated because it doesn't support the changes version, but we can't formally add a deprecation comment or warning yet because there is no user-facing replacement until the content storage layer is stable and no longer internal.

src/Cms/ModelWithContent.php
lukasbestle: It already does, but then it throws an exception (see my comments above for the reasoning). Having an uncaught exception here would break this method, so we need the `exists()` check.

Actually it would be more robust to replace the `exists()` check with a `try catch` block for `NotFoundException`, but I'm not sure if that's too much.

src/Content/ContentStorage.php
lukasbestle: It is, see the `ensureExistingVersion` method that's called from every method that has the `@throws` annotation.

> I think we could reduce complexity if these methods all handle the non existing case themselves instead of throwing exceptions.

What solution do you think of? Could you maybe write a short sentence for each method with what it should do in the "non existing" case?

src/Content/PlainTextContentStorage.php
lukasbestle: 👍

src/Content/ContentStorage.php
lukasbestle: Not sure. Thinking about it again, I feel like `language()` as a helper method doesn't belong in this class at all, especially because the implementation won't differ between different storage handlers. Maybe we can find a better place for it.

src/Content/PlainTextContentStorage.php
lukasbestle: 👍

src/Content/PlainTextContentStorage.php
lukasbestle: 👍

lukasbestle:
> Can I make one more suggestion: I think it would be good if we don’t directly merge this into `v4/develop` but e.g. `v4/changes` first. I think this kind of separation was really good with other big refactoring parts we did early on for v4

I'd still like to put this into the next alpha to get it out there for testing. Do you think a separate branch makes sense in addition to that or do you rather suggest to not put this into the next alpha at all?

distantnative:
I think regardless of whether we put it in the next alpha or not, we should keep the whole changes refactoring separate from `v4/develop` for as long as it has proven unproblematic (similarly how we did with `v4-lab`, `v4-vue`). This fail-safe has given us some peace of mind that is needed here to in my opinion.

src/Content/ContentStorage.php
distantnative: Ok, maybe phrased differently: can there be an implementation of `ContentStorage` that doesn't need a language method?

src/Content/ContentStorage.php
distantnative: You're right. In these cases it wouldn't be good e.g. if update just doesn't do anything and pretends it succeeded.

src/Cms/Language.php
distantnative: While I agreed with you on the `ContentStorage::touch` probably throwing an exception, I just can't look passed how verbose this code is. It also makes the code hard to read as the nuances are easily missed between the cruft. Wondering if there is an easier way, that's still not unsafe in this specific case (where we don't care what happens if the file doesn't exist) that doesn't create such bloat.

src/Cms/ModelWithContent.php
distantnative: Not anyone? Is it that safe - I would think someone that tried to implement some virtual page storage that's not our text files would have needed to use those, wouldn't they? For them, I think the deprecation message needs to point them into any direction they can go forward with. Just "tough luck" doesn't feel good.

src/Cms/ModelWithContent.php
distantnative: Ok

src/Cms/ModelWithContent.php
distantnative: Try-catch would match `readContent`, so maybe that'd be more consistent then

src/Cms/Language.php
distantnative: But if we need to merge before finding a solution because we're running out of time, can't hold this back. I'm quite unhappy that now time doesn't leave us with a chance to get these things right and I am afraid we won't get around to fix them later on.

src/Content/ContentStorage.php
lukasbestle: No, since passing the language as object is already a requirement in the `ContentStorage` interface.

The language method is only there because the core passes around language codes as strings at the moment. With our planned "always multilang" refactoring I think that we will be able to update other places in the core to use the same logic, so the language method is not specific to the content storage at all, just used here for the first time.

src/Cms/ModelWithContent.php
lukasbestle: We'll see what our users actually need once they test alpha 4 with their sites and plugins. 🙂

src/Cms/Language.php
lukasbestle: I agree the code is quite verbose at the moment because it directly calls the low-level storage methods. Maybe we can build mid-level methods to make this more DRY.

I've added a new section "clean-up tasks for later" to the concept and added a task there. If we collect such issues there, we will not forget about them.

src/Content/ContentStorage.php
distantnative: I think that’s actually an issue. We’re diverting here quite a lot from the usual Kirby way which is very focused on ease of use of its interfaces/methods/API. Which is why we allow passing a language as language code string almost everywhere. It’s not as great for type hinting as a language object, but it raises the UX (or developer X) a Lot.

With requiring a language object, the code of the language class gets leaner and easier. But it’s actually just externalizing the costs and blows up the code all around the system.

I think we should keep this quality of ease of use and allow passing a language code string and let the class handle getting the object internally. This would be more consistent with Kirby.

(It’s probably also the biggest edge I have with VersionIdentifier etc - using them internally can make a lot of sense, but externalizing the work to pass them as object instead of a simple string doesn’t feel right and match how Kirby usually deals with this and offers an easy-to-use interface, e.g. page status as string)

src/Cms/Language.php
distantnative: I’m doubtful we’ll get around to it later (resource wise) if we don’t do it right from the start. But what can I do with that time pressure…

src/Cms/ModelWithContent.php
distantnative: I don’t quite get what’s the issue of pointing them directly in a good direction, even avoid frustrations during the alpha.

src/Content/ContentStorage.php
lukasbestle: On passing language codes as strings:

The language method took over its logic 1:1 from the old `$model->contentFile()` method, including the `$force` behavior that is needed for the conversion code in the `Language` class.

Where do we put all of that logic when we switch to strings for the language code arguments? I see two alternatives:

- We handle all of the complexity inside the content storage handler. This has two gotchas: a) each handler would need to implement the same logic and b) every handler method would need an additional `$force` argument, which I think would be a huge downgrade in code quality. So IMO this is not really an option.
- We handle the complexity in the calling code and only pass the "final" language code to the storage handler that can be used 1:1 for the content filename. To avoid code duplication, I'd still put the logic for the language code conversion in a helper method and we'd need to find a place for that.

distantnative: I would handle it inside. We could ourselves also handle it outside reasonably but then every developer calling these methods would need to have the same awareness etc. - they’re much more used to pass the language code string as everywhere else in the system. So I think this would offer a much better developer experience if we handle it inside. 

I’ll look closer at it tonight if there aren’t good ways to handle this inside once and not on every method. Or maybe it isn’t the biggest deal to do it on the methods how we’ve done so far too. I’ll need to do a closer take on that.

lukasbestle: I'd much prefer making the storage handlers as dumb as possible regarding this. They shouldn't have to think about language edge cases, just receive either a language code or `null` in single-lang.

distantnative: I know, but just externalising these complexities into the code that developers have to write can't be a good answer either. We need to find some middle ground.

distantnative: Looking at the `$force` parameter, that seems to really be the culprit. I am not Sure if it actually should exist. Or if we are just using it to not face the actual correct behavior in other parts.

So if we look at where this is used:
- `Cms\Language`: There's code in those methods that actually doesn't belong in the language class. The language class should not be concerned with moving the files. In the Language class there should probably just be one call per object, something like `$model->storage()->convertVersionsForLanguage(string $from, string $to)` and `$model->storage()->deleteVersionsForLanguage(string $code)`. And then each storage handler needs to know what to do. In our case PlainStorage to do the `F::move` calls. I could imagine that in this context the need for the $force parameter disappears.
- `Cms\ContentTranslation:: contentFile`: seems like an odd case to force this, there shouldn't be a content translation for any invalid language, that'd be a problem itself
- `Cms\Translations`: I think those two methods your PR removed already anyways

lukasbestle: As Nico and I discussed separately, this is something we want to improve, but outside of this PR. Nico will work on a solution. I'll leave this discussion open for reference.

PR #5297, opened 2023-06-17
Clean up `search` component

### LINKED ISSUE #5294: `search` component: no results for short query
Not 100% sure yet, but wanted to note this down:

It seems like our core `search` component returns an empty collection for any search term that is shorter than `minLength`. I would think however, that it should rather contain the full collection.

distantnative:


distantnative:
Based on the failing SitePagesTest, the previous behavior could have been intended.
But I am not sure if that's correct: if I search for `"tr"` but the `minLength` is `3`, should I get an empty collection or the full collection (as currently returned when the query is empty.

I would think empty/too short query should be treated the same.

tests/Cms/Collections/SearchTest.php
```diff
@@ -33,6 +33,10 @@ public function testCollection()
 
 		$search = Search::collection($collection, 'm', ['minlength' => 1]);
 		$this->assertCount(3, $search);
+
+		// minLength of 2 is not reached, should return foll collection
```
bastianallgeier:
```suggestion
		// minLength of 2 is not reached, should return full collection
```

bastianallgeier:
I agree that the behaviour should be the same. But I honestly wonder if returning the full collection is what is expected at all. On the other hand, I'm not sure if it's worth to add a breaking change here. That's why I would probably vote to keep on returning the full collection.

lukasbestle:
I think there are arguments for either solution:

- Returning all results makes sense semantically, otherwise there's no point to have `minlength` at all.
- Returning no results is clearer to the user because they can clearly see that their query was invalid.

distantnative:
I have changed the behavior back:
Both for empty as well as too short queries, an empty collection is returned.

This PR now is mainly a code cleanup of the search component that avoids some regex calls if the query is too short anyways.

bastianallgeier (approved):

PR #5298, opened 2023-06-18
`Structure` to extend `Item` classes

distantnative:
## This PR …

- Align the structure classes to be built on top of `Cms\Item`, same as blocks, layouts etc.
- I think this will help to have consistent support for UUIDs for blocks and structure objects

### Feature
- Custom methods for structure and structure object

### Refactored
- `Kirby\Cms\Structure` extends `Kirby\Cms\Items` and `Kirby\Cms\StructureObject` extends `Kirby\Cms\Item`

### Breaking changes
- `new Structure()`/`new StructureObject()` don't work anymore as before. Use `Structure::factory()`/`StructureObject::factory()` instead
- Structure object IDs aren't simply their collection index numbers anymore but receive a `Str::uuid()` as blocks do
- `Items::factory()` and all inheriting classes throw an exception now if malformed data is passed

## Ready?

### For review team

### DIFF
--- a/config/methods.php
+++ b/config/methods.php
@@ -262,12 +262,15 @@ return function (App $app) {
 		 */
 		'toStructure' => function (Field $field) {
 			try {
-				return new Structure(Data::decode($field->value, 'yaml'), $field->parent());
+				return Structure::factory(
+					Data::decode($field->value, 'yaml'),
+					['parent' => $field->parent()]
+				);
 			} catch (Exception) {
 				$message = 'Invalid structure data for "' . $field->key() . '" field';
 
 				if ($parent = $field->parent()) {
-					$message .= ' on parent "' . $parent->title() . '"';
+					$message .= ' on parent "' . $parent->id() . '"';
 				}
 
 				throw new InvalidArgumentException($message);
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -73,6 +73,8 @@ trait AppPlugins
 		'sections' => [],
 		'siteMethods' => [],
 		'snippets' => [],
+		'structureMethods' => [],
+		'structureObjectMethods' => [],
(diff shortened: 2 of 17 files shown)


bastianallgeier (approved):
I like this a lot. Do we want to put this in alpha 4?

iskrisis:
Not sure this is intended or in development. But @distantnative it seems that the ids are not persisted.

nilshoerrmann:
> Kirby\Cms\Structure extends Kirby\Cms\Items and Kirby\Cms\StructureObject extends Kirby\Cms\Item

Would it make sense to rename `StructureObject` to `StructureItem`?

distantnative:
@nilshoerrmann we were trying to stay away from those kind of breaking changes for now

nilshoerrmann:
Totally understandable. I just wanted to ask because `StructureObject` always sounded strange in context of other names within Kirby (page/pages, file/files, user/users, but not structures/structure). But never mind :)

iskrisis:
Sorry to interrupt the conversation but coming back to the initial question... should i file issue about structures not persisting ids like blocks do? or is it something you all know about.

distantnative:
I'm rather undecided. It's already in the breaking changes section and structure objects also didn't have a persistent id before, did they?

iskrisis:
In v3 the ids are indexOf and thus kinda persistent. I found out about this by helping user on Discord who was updating his site to v4. 
I have not tested this but i think he was using structureObject->id() in v3 as value in select fields. Since id() was index before it was working but now it was confusing for him because the id() was changing all the time.

I understand this is something thats moving so maybe its OK.

distantnative:
Hmmm I see. Could you maybe open an issue (I'm on my phone) so we don't forget? Wasn't a persistent issue either when being moved but I can see the use case where it's at least persistent as long as touched.

PR #5299, opened 2023-06-18
Block default model: change key to `default`

distantnative:
## This PR …

I've been wondering why we ask people to register the default block model with `Kirby\Cms\Block` and not with a simple `default` as key like we do it in most other cases: https://getkirby.com/docs/reference/plugins/extensions/block-models#overwriting-the-default-block-class

It's an added breaking change, but I think it would be good to remove the tech-term for the more common one.

### Breaking changes
- Registering a default block model now needs to be done with the key `default`, not `Kirby\Cms\Block`

## Ready?

### For review team

### DIFF
--- a/src/Cms/Block.php
+++ b/src/Cms/Block.php
@@ -170,7 +170,7 @@ class Block extends Item
 		}
 
 		// default model for blocks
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
You are totally right. This is much better.

PR #5300, opened 2023-06-18
Fix block collapsing inside layouts

### LINKED ISSUE #5289: [v4] Layout field layout should also collapse when dragging for consistency with Blocks field
## Description
When dragging layouts in a Layouts field, they will not be collapsed and all blocks inside stay full size. I expect them to be collapsed similar to when dragging single blocks.

## Your setup
**Kirby Version**  v4.0.0-alpha.2

distantnative:
## This PR …

@bastianallgeier did these CSS lines have any other purpose than blowing up the last block in a layout to fill the full layout? And to make text blocks demand the most space?

If there aren't any other side effects, I think the benefit of collapsible blocks is a better choice.

### Fixes
- #5289

### DIFF
--- a/panel/src/components/Forms/Blocks/Types/Text.vue
+++ b/panel/src/components/Forms/Blocks/Types/Text.vue
@@ -88,9 +88,4 @@ export default {
 	line-height: 1.5em;
 	height: 100%;
 }
-.k-block-type-text,
-.k-block-container-type-text,
-.k-block-type-text .k-writer .ProseMirror {
-	height: 100%;
-}
 </style>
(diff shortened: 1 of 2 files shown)


bastianallgeier:
I think it was about being able to focus that block when you click inside the layout. But I don't really remember anymore TBH.

distantnative:
@bastianallgeier Can you try? Focusing seems to still work for me when clicking on the background even if it’s not the block itself anymore

bastianallgeier:
It's as I guessed. The pink area is not clickable. 

Screenshot 2023-06-19 at 17 16 22

distantnative:
Hmm weird worked for me yesterday. Will try again in different browsers.

Still think the collapsing could be more beneficial

bastianallgeier:
Wouldn't it be possible to reset the height when the block is collapsed?

distantnative:
@bastianallgeier I think I found a way that allows for both

bastianallgeier:
The last one is still not expanded until the end of the layout column, but I think it's fine. The collapsing feature is indeed more important.

PR #5301, opened 2023-06-18
Fix users field default

### LINKED ISSUE #5284: Users field: current user stored unless default set to false
## Description
With a users field like this

```
Title: Test

fields:
  mentor:
    type: users
    query: kirby.users.filterBy('role', 'in', ['student', 'teacher'])
```

When you create a new page via the frontend, not matter what you pass to `mentor` or don't set the field, the current users is always stored. The same happens when creating a new Page via the Panel.

Example frontend code:

```
kirby()->impersonate('kirby');
$page = Page::create([
  'slug'     => 'new-page',
  'template' => 'test',
  'content' => [
    'title'  => 'New Page',
  ]
]);
```
It needs:
```
  mentor:
    type: users
    query: kirby.users.filterBy('role', 'in', ['student', 'teacher'])
    default: false
```
to prevent that behavior.

**Expected behavior**  
The field should stay empty when no default user is explicitly set.

## Your setup

Tested with Kirby 3.9.5

distantnative:
## This PR …

@texnixe since query support pose the question again what happens if `username === query entry` (e.g. page, site...), I didn't pursue that option. Instead, now it needs to be `$default === true` and not `$default === null` to automatically set the current user. Do you think that works?

### Fixes
- Users field does not anymore use a default if none set #5284

### Enhancement
- Users field: set `default: true` to always use the currently logged in user as default

### Breaking changes
- Users field doesn't automatically uses the current user as default, add `default: true` to keep this functionality

## Ready?

### For review team

### DIFF
--- a/config/fields/users.php
+++ b/config/fields/users.php
@@ -24,18 +24,8 @@ return [
 		/**
 		 * Default selected user(s) when a new page/file/user is created
 		 */
-		'default' => function ($default = null) {
-			if ($default === false) {
-				return [];
-			}
-
-			if ($default === null && $user = $this->kirby()->user()) {
-				return [
-					$this->userResponse($user)
-				];
-			}
-
-			return $this->toUsers($default);
+		'default' => function (string|array|bool $default = null) {
+			return $default;
 		},
 
 		'value' => function ($value = null) {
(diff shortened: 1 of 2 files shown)


bastianallgeier:
If there's a breaking change, we can't really add it to 3.9.6, right?

distantnative:
@bastianallgeier Question is if it’s a breaking change or a bug how it has been so far

afbora:
It's a bug but I'd also say it's a breaking change since it's been used that way for a long time. I think v4 is a great way to go for this kind of risky cases.

distantnative:
Ok moved it to v4.

config/fields/users.php
```diff
@@ -24,29 +24,31 @@
 		/**
 		 * Default selected user(s) when a new page/file/user is created
 		 */
-		'default' => function ($default = null) {
-			if ($default === false) {
+		'default' => function (string|array|bool $default = null) {
```
bastianallgeier: Shouldn't the type hints be: 

```suggestion
		'default' => function (string|array|bool|null $default = null) {
```

bastianallgeier (approved):

Issue #5204, opened 2023-05-29
[v4] Link field: Browsers don't collapse on blur 


distantnative:
Only when clicking on the upper input part but not when focusing another field

PR #5306, opened 2023-06-18
Link field improvements

### LINKED ISSUE #5244: [v4] Link field: option to limit what types to support
E.g. not every link field makes much sense to have the `tel` option.

### LINKED ISSUE #5278: [v4] Writer field: Allow any URI scheme as Urls in Link field
## Description
I want to add a link with a less common [URI scheme](https://en.wikipedia.org/wiki/List_of_URI_schemes) in a writer field, for example:
```
facetime://+123456789
smb://server
maps:q=Berlin
```
It is possible to add these Urls in the Link dialog of the writer field. But after saving the changes and reloading the panel, the links in the writer field are gone. 

## Your setup
**Kirby Version**  
4.0.0-alpha.3

distantnative:
## To do
- [x] Improve naming of changes
- [x] Frontend detection to only check for active types
- [x] Backend validation to only validate active types
- [x] Backend validation to include other

### Features

- New anchor link type
Screenshot 2023-06-23 at 13 30 18

- New custom link type
Screenshot 2023-06-23 at 13 30 29

- New anchor icon

### Enhancements
- Link field
  - Allows selecting which types (url, email, tel...) to include
  https://github.com/getkirby/kirby/issues/5278
  - New `Other` type that allows any other input
  https://github.com/getkirby/kirby/issues/5244

### Fixes

- An outside click will now collapse the link field https://github.com/getkirby/kirby/issues/5204


tobimori:
just fyi: The plugin uses `linkTypes`.

`options` seems more in line with the rest of the fields to me, but it might be useful to reserve it for other options (like choosing queries for files/pages)

distantnative:
Let's say it'd be `options`. The first iteration would probably to allow

```yaml
options:
  - url
  - email
```

which then could though develop into

```yaml
options:
  - url
  page:
    query: site.page('album').children
```

bastianallgeier:
I've changed the name to options and rebased it on my other link field improvement branch.

bastianallgeier (approved):

PR #5309, opened 2023-06-18
Fix block keydown propagation

### LINKED ISSUE #5272: [v4] Keyboard shortcuts for Nested Blocks field get applied to parent
## Description
When using nested blocks inside a field, hotkeys will be applied both to the selected blocks field and the parent blocks field.

https://streamable.com/idlffi

**Expected behavior**  
Keyboard shortcuts would not apply to parent if selecting a nested blocks field

## Your setup
**Kirby Version**  4.0.0-alpha.2

distantnative:
### Fixes
- #5272


bastianallgeier:
TBH, I think we can ignore this. With the latest drawer changes, we no longer render two drawers at once.

distantnative:
@bastianallgeier No of course we do and continue to do so with the fields previews block

bastianallgeier:
Oh right. It makes sense that we need it for the fields preview. But how do we render two drawers at once?

distantnative:
Why drawers? There it’s just nested blocks. As you said, one drawer. But in the preview you have a blocks inside a block and so we need to stop propagation

bastianallgeier:
🙈 Totally stood on se Leitung

bastianallgeier (approved):

Issue #5247, opened 2023-06-05
Allow plugins to explicitly define their assets


rasteiner:
## Description
Only some hardcoded file extensions can be served via the core plugin assets media route. 

**Expected behavior**
The plugin assets media route should behave like the other media routes. 

**Screenshots**  
https://github.com/getkirby/kirby/blob/282b1e25faa0bc24c903237a7484dc45fa475c36/config/routes.php#L62-L68

## Your setup
**Kirby Version** 
Since 3.0

**Additional context**  
Avowedly this is done to protect the client side from dangerous files, but in my opinion this thought ignores the fact that the most dangerous files in the context of browsers (js files) are already allowed. 

From the outside it looks like "plugin code" is considered inherently unsafe, this would make it deserve *special rules to protect the user*... While being questionable (imo it's just "user code"), this in all cases ignores the fact that plugin code is anyway executed with the same privileges as Kirby and the rest of the user code - there just doesn't exist a "protect the user from plugins", be it server or client side. 

Forcing plugin devs to write their own route to serve "unknown assets" (because plugin devs don't have the choice to "just not serve them") in my opinion just leads to more, potentially unmaintained, code duplication.

distantnative:
I think I remembered why: This is maybe less about the user but rather the developer. Plugin folders could contain a number of files that you might not want publicly accessible - composer/packagejson files that reveal what versions of the plugins you use, READMEs, maybe even license files.

Not sure if it's a real problem, but I think that was the thought to define a list of extension types that are very likely intended to be published/public.

rasteiner:
In my humble opinion it would be safe to assume that files that are put into the "assets" folder are meant to be public. 

I mean the route doesn't publish the whole plugin folder, only the "assets" folder. Package and composer files normally stay in the plugin root (and if put under assets, the current route would already serve them). Finally, if a dev puts his secret files into a folder called "assets", I think that could be smoothly considered user error.

distantnative:
You're right. I missed that part.

bastianallgeier:
While I agree with the notion that this shouldn't be too opinionated. I don't agree that this is just a user error. A plugin can easily do quite some damage here by using a PHP file as asset for example. Intentionally or not. There are other vectors that this could introduce. I think it should be an allow list that can be extended. But it needs to happen deliberately in my opinion.

rasteiner:
[...]

PR #5312, opened 2023-06-18
Plugin assets: allow any file type for media route

distantnative:
Fixes 

### Enhancements
- Plugin files from the plugin's `assets` folder can now be accessed via its media url no matter what file extension they have
#5247

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -60,10 +60,18 @@ return function ($kirby) {
 			}
 		],
 		[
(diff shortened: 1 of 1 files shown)


bastianallgeier:
I think this could turn into a security risk tbh. You could easily put a PHP file in there that runs some nasty commands via URL. There are more possible ideas. I think this needs to be an allow list that can be configured.

distantnative:
Closing this based on discussion in https://github.com/getkirby/kirby/issues/5247

PR #5329, opened 2023-06-27
File blueprint `focus` option

### LINKED ISSUE #5316: [v4] Any file with Image Preview allows setting the Focus Point, not only images
## Description
I have a file template, that allows the client to upload documents (pdf) which has a Files field for selecting a poster/thumbnail image that's shown in the frontend.

Using the following syntax, I make the thumbnail available as preview in the panel too:
```yaml
image:
  icon: file
  back: "#f3f3f3"
  cover: false
  query: file.content.thumbnail.toFile
```

As soon as such an image is set, the focus point option is available - even if the file isn't an image:
CleanShot 2023-06-19 at 19 58 33@2x

**Expected behavior**  
This might be the expected behavior, but I'd assume it will confuse my clients in some cases. (e.g. Why can I set two focus points for such a document, one for the thumbnail, and one for the document itself?)

My recommendation would be the following:
- Automatically have it active when the file type is image, not when the image preview is set.
- Allow manual activation for other file types, and also disable for image file type, in file blueprint.

## Your setup
**Kirby Version**  v4.0-alpha.3

distantnative:
### Enhancement
- New `focus` option for file blueprints to enable/disable setting focus point in the Panel (defaults to true for all viewable images, false for all other files)
#5316

## Ready?

### For review team

### DIFF
--- a/src/Panel/File.php
+++ b/src/Panel/File.php
@@ -267,6 +267,40 @@ class File extends Model
 	}
 
 	/**
+	 * Whether focus can be added in Panel view
+	 */
+	public function isFocusable(): bool
+	{
+		// blueprint option
+		$option     = $this->model->blueprint()->focus();
+		// fallback to whether the file is viewable
+		// (images should be focusable by default, others not)
+		$option   ??= $this->model->isViewable();
+
+		if ($option === false) {
+			return false;
+		}
+
+		// ensure that user can update content file
+		if ($this->options()['update'] === false) {
+			return false;
+		}
+
+		$kirby = $this->model->kirby();
+
+		// ensure focus is only added when editing primary/only language
+		if (
+			$kirby->multilang() === false ||
+			$kirby->languages()->count() === 0 ||
+			$kirby->language()->isDefault() === true
+		) {
+			return true;
+		}
+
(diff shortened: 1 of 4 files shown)


src/Panel/File.php
```diff
@@ -266,6 +266,40 @@ protected function imageSource(
...
+
+		// ensure that user can update content file
+		if ($this->options()['update'] === false) {
+			return false;
+		}
+
+		$kirby = $this->model->kirby();
+
+		// ensure focus is only added when editing primary/only language
+		if (
```
bastianallgeier, line 292: I think this is the wrong place. A file is still focusable even if a non-default language is selected. I think we need to make sure that the focus value is handled differently in writes and only added to the default language there instead.

distantnative: But `isFocusable` means whether a focus can be set - and a focus should only be set in the primary language. It can be read in any. Keep in mind that this is the File-Panel class not the main File class.

bastianallgeier: Oh, I missed that it's not in the File class. Then it totally makes sense.

bastianallgeier (approved):

PR #5339, opened 2023-06-30
`FileActions`: consistent multi-lang handling

distantnative:
## This PR …

Isolated improvement that will help us in the changes refactoring.

Makes the multi-lang handling consistent between the FileActions methods and prepares the code to easier adapt to a setup where single-lang is just multilane with lang count = 1.

### Breaking changes
- `$languages->codes()` now returns`['default']` for single language installations

## Ready?

### For review team

### DIFF
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -230,14 +230,12 @@ trait FileActions
 				throw new LogicException('The file could not be created');
 			}
 
-			// always create files in the default language
-			$languageCode =
-				$file->kirby()->multilang() ?
-				$file->kirby()->defaultLanguage()->code() :
-				null;
-
 			// store the content if necessary
-			$file->save($file->content()->toArray(), $languageCode);
+			// (always create files in the default language)
+			$file->save(
+				$file->content()->toArray(),
+				$file->kirby()->defaultLanguage()?->code()
+			);
 
 			// add the file to the list of siblings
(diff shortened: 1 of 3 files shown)


src/Cms/FileActions.php
```diff
@@ -57,16 +57,19 @@ public function changeName(string $name, bool $sanitize = true): static
...
-					// rename the content file
-					F::move($oldFile->contentFile($translationCode), $newFile->contentFile($translationCode));
-				}
-			} else {
+			$translations =
+				$newFile->kirby()->multilang() ?
+				$newFile->translations()->values(
+					fn ($translation) => $translation->code()
+				) :
+				['default'];
```
lukasbestle: Can we make this a method in the `Languages` collection class to make the logic DRY?

Maybe we could extend the existing `$kirby->languages()->codes()` method with a boolean flag that takes care of the single-lang mode.

distantnative: Just that some to `$kirby->languages()` other parts do `$model->translations()`

lukasbestle: We just need the codes and those are always the same no matter if we access them via languages or translations:

https://github.com/getkirby/kirby/blob/6b5dda600472b59a3a779f7e6b8e9e130cc414a1/src/Cms/ModelWithContent.php#L600-L611

lukasbestle (approved):

PR #5340, opened 2023-07-01
More `Cms` namespace type hinting/clean up

distantnative:


src/Cms/LanguageRouter.php
```diff
@@ -106,26 +91,32 @@ public function routes(): array
 	 * Wrapper around the Router::call method
 	 * that injects the Language instance and
 	 * if needed also the Page as arguments.
-	 *
-	 * @param string|null $path
-	 * @return mixed
 	 */
-	public function call(string $path = null)
+	public function call(string $path = null): mixed
```
afbora:
```suggestion
	public function call(string|null $path = null): mixed
```

src/Cms/LanguageRouter.php
```diff
@@ -106,26 +91,32 @@ public function routes(): array
...
 	{
-		$language = $this->language;
-		$kirby    = $language->kirby();
-		$router   = new Router($this->routes());
+		$language       = $this->language;
+		$kirby          = $language->kirby();
+		$this->router ??= new Router($this->routes());
```
afbora, line 99: Caching router instance is intended?

src/Cms/LanguageRoutes.php
```diff
@@ -73,23 +67,28 @@ public static function fallback(App $kirby): array
...
+
 						if (
 							$kirby->option('languages.detect') === true &&
-							$page->translation($kirby->detectedLanguage()->code())->exists() === true
+							$page->translation($lang)->exists() === true
 						) {
 							return $kirby
 								->response()
-								->redirect($page->url($kirby->detectedLanguage()->code()));
+								->redirect($lang);
```
afbora:
```suggestion
								->redirect($page->url($lang));
```

src/Cms/Languages.php
```diff
@@ -24,12 +24,13 @@ class Languages extends Collection
...
 	 *
-	 * @param array $objects `Kirby\Cms\Language` objects
 	 * @param null $parent
 	 * @throws \Kirby\Exception\DuplicateException
 	 */
-	public function __construct($objects = [], $parent = null)
-	{
+	public function __construct(
+		array $objects = [],
+		$parent = null
```
afbora, line 32:
```suggestion
		object|null $parent = null
```

afbora (commented):
I can only find above minor things from mobile 😅

src/Cms/LanguageRouter.php
distantnative: We always had the property in the class but it was never used, thought this probably was supposed to always be cached

src/Cms/Languages.php
distantnative: According to the doc block, we actually don't want this parameter to be set to anything

afbora (approved):

PR #5344, opened 2023-07-03
Writer: `quote` node

distantnative:
### Feature
- Writer: blockquote `quote` node (by default not included in nodes dropdown)

Screenshot 2023-07-04 at 00 43 26


luxuryluke:
Thank you!

panel/src/components/Forms/Writer/Writer.vue
```diff
@@ -450,6 +452,14 @@ export default {
 	font-family: var(--font-mono);
 }
 
+.k-writer blockquote {
+	font-size: var(--text-xl);
```
afbora: The xl font size looks too big because fonts are shrunk in the new UI. `lg` is better.

```suggestion
	font-size: var(--text-lg);
```

afbora (approved):

Issue #5349, opened 2023-07-04
[v4] Error while deleting field in duplicated layout


JerikoOne:
## Description
When a block inside a duplicated layout is deleted an error message "Cannot read properties of undefined (reading '0')" pops up.

**Expected behavior**  
The block gets deleted without an error message.

**Screenshots**  
<img width="878" alt="image" src="https://github.com/getkirby/kirby/assets/3002580/c79bf2ac-10ac-44c8-b132-3b9d05b37e6e">

## To reproduce
1. Create a layout. Which one doesn't seem to matter
2. Create a block, e.g. "Text" with some text in it
3. Duplicate the layout
4. Inside the duplicated layout, delete the block.
5. See error

## Your setup
**Kirby Version**  
4.0.0-alpha.4

**Console output**  
No console output

**Your system (please complete the following information)**
 - Device: Macbook Pro
 - OS: macOS Ventura 13.3.1
 - Browser: Brave 
 - Version: Version 1.52.129 Chromium: 114.0.5735.198

PR #5350, opened 2023-07-04
Fix duplicated layout issues

afbora:
### Fixes
- Fix duplicated layout issues #5349

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Layouts/Layouts.vue
+++ b/panel/src/components/Forms/Layouts/Layouts.vue
@@ -108,15 +108,13 @@ export default {
 			this.$refs.selector.open({ rowIndex, layoutIndex, layout });
 		},
 		duplicate(index, layout) {
-			let copy = {
-				...this.$helper.clone(layout),
-				id: this.$helper.uuid()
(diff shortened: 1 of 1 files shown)


panel/src/components/Forms/Layouts/Layouts.vue
```diff
@@ -116,7 +116,7 @@ export default {
 			// replace all unique IDs for columns and blocks
 			copy = this.updateIds(copy);
```
distantnative: Could you add a comment above here that updateIds always returns an array? Do that our copy object now is an array. Maybe we should even rather use `const copies` for this line? Cause then your change makes a lot more sense when reading it. First I was very irritated why that change.

afbora: I've improved the method and add a description https://github.com/getkirby/kirby/pull/5350/commits/12045057b6e92bf0788c3c00692b268592b62ad4

bastianallgeier (approved):
LGTM

Issue #5351, opened 2023-07-05
[v4] Additional divider at end of Node list when disabling nodes


tobimori:
## Description
![CleanShot 2023-07-06 at 00 14 39@2x](https://github.com/getkirby/kirby/assets/29142128/51cd26f3-a234-49d8-a736-e322d0bdc18d)

Example config:
```yaml
writer:
	type: writer
	label: Text
	marks:
	  - bold
	  - link
	  - email
	nodes:
	  - heading
	  - paragraph
	headings:
	  - 2
```

## Your setup
**Kirby Version**  4.0.0-alpha.4

PR #5352, opened 2023-07-06
Fix additional divider on writer field 

afbora:
### Fixes
- Fix additional divider on writer field #5351

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Writer/Toolbar.vue
+++ b/panel/src/components/Forms/Writer/Toolbar.vue
@@ -12,7 +12,7 @@
 				@click="$refs.nodes.toggle()"
 			/>
 			<k-dropdown-content ref="nodes">
-				<template v-for="(node, nodeType) in nodeButtons">
+				<template v-for="(node, nodeType, nodeIndex) in nodeButtons">
 					<k-dropdown-item
 						:key="nodeType"
 						:current="activeNodeButton?.id === node.id"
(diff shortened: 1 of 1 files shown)


panel/src/components/Forms/Writer/Toolbar.vue
```diff
@@ -22,7 +22,7 @@
 					>
 						{{ node.label }}
 					</k-dropdown-item>
-					<hr v-if="node.separator === true" :key="nodeType + '-divider'" />
+					<hr v-if="node.separator === true && nodeIndex !== Object.keys(nodeButtons).length - 1" :key="nodeType + '-divider'" />
```
distantnative, line 25: I think this would only fix the additional divider when at the end, wouldn't it?
But maybe there are other scenarios?

Wondering if hiding it via CSS could be the more reliable solution. In the direction of `hr + hr { display: none }`

afbora: That's exactly the issue. 

> Additional divider at end of Node list when disabling nodes

I think if it is the divider and the last index, it's clearly an unnecessary divider. It adapts to all cases.

afbora: Also I don't think `hr + hr` correctly works. This requires two consecutive hr. That's not the issue.

````
- Menu item
____________
- Menu item
- Menu item
____________ <-- the issue
````

distantnative: True

distantnative (approved):

PR #5362, opened 2023-07-09
Fix drag’n’drop between blocks fields

### LINKED ISSUE #5290: Drag'n'drop between block fields broken
## Description
create a block and drag it to another block field. keep holding the mouse button and drag. it disappears in the initial and appears in the second. but on releasing the mouse button it disappears.

**Expected behavior**  
preferable allow dragndrop or make sure the UI does not act like it would be possible.

**Screenshots**  
CleanShot 2023-06-15 at 13 35 31

distantnative:
## This PR …

Debouncing the input handler is a bad idea for situations where two fields' value change at once (e.g. dragging one block from one field to another, both blocks simultaneously will emit an input event).

If we think certain fields would be emitting their input way too often, we should debounce this on the emitter, not on the receiving fields section.

### Fixes
- Drag'n'drop between blocks fields works agaiN
#5290

### For review team

### DIFF
--- a/panel/src/components/Sections/FieldsSection.vue
+++ b/panel/src/components/Sections/FieldsSection.vue
@@ -17,7 +17,6 @@
 
 <script>
 import SectionMixin from "@/mixins/section.js";
-import debounce from "@/helpers/debounce.js";
 
 export default {
 	mixins: [SectionMixin],
@@ -42,7 +41,6 @@ export default {
 		}
 	},
 	created() {
-		this.input = debounce(this.input, 50);
 		this.fetch();
 	},
 	methods: {


bastianallgeier:
We introduced this to fix a pretty big performance issue with larger forms. I think if we remove it, we need to go through each field first and find the culprits.

distantnative:
@bastianallgeier is this true? got introduced here https://github.com/getkirby/kirby/commit/2b6c0b5f3c2639d0abdcaaefdac52cb0d2661597#diff-864fe7d5e19566c8e25118990298f723f8bd9b5b335d183f48fca186bda0aff4 which looks like it was just added in the mix when touching other debounce calls

bastianallgeier:
Yes, but I actually really remember how it was about a pretty big performance issue back then.

bastianallgeier:
I don't think we can add this to the RC today. I'm really worried that we will break the things we fixed with the debounce.

distantnative:
Can you remember any more details? It's really hard to work with this that it was just some performance issue back then.

And currently this keeps a pretty major bug in the core (two inputs cannot change their value at the same time).

bastianallgeier:
If I remember correctly, the performance issues started to happen when the blueprints got a little more complex. The main fault here seems to be the content changes checks.

distantnative:
Closing this for now even though it feels a bit unsatisfying - we do need a clearer "this PR does not work, because when you do X, Y performance is horrible". The vague descriptions of something that might could happen aren't so helpful when they block fixing such a drastic bug.

Issue #5359, opened 2023-07-09
User blueprint `home` does not redirect on Panel login


olivertedwards:
## Description

Previously, we could specifiy in a user blueprint where a user should be redirected upon login via the panel login form. I set up two blueprints (admin.yml and user.yml) where admin has panel access and user does not, and specified that the user should be redirected to the frontend home page after login.

However, when trying to log in as a user (associated with user.yml) I get a Load failed error, and when trying to reload the panel, the page fails completely with a Too many page redirects error. This results in a total panel failure, and the only way to make it work again, is to delete all session files in `site/sessions`.

Here is the user.yml setup, that works in the latest Kirby 3.9.5 versions, but not in the latest 4.0.0-alpha.4 version (have double checked and tested in both versions):

```yml
title: User
home: /
permissions:
  access:
    panel: false
```

**Expected behaviour**
As in the stable K3 release version, I was expecting the user to be redirected to the home page as instructed through `home: /`.

**Screenshots**
Before login
<img width="946" alt="Screenshot 2023-07-09 at 15 47 00" src="https://github.com/getkirby/kirby/assets/2511797/47be1973-26a9-41ab-9544-3d3ea0529d79">
After entering user credentials and clicking login
<img width="1074" alt="Screenshot 2023-07-09 at 15 57 57" src="https://github.com/getkirby/kirby/assets/2511797/ff680aa8-64e6-4181-933d-dc176d26481b">
After trying to reload the panel login page
<img width="1080" alt="Screenshot 2023-07-09 at 15 46 32" src="https://github.com/getkirby/kirby/assets/2511797/371ec247-93fd-4f72-babc-36728bebfa11">

## Setup
- Kirby 4.0.0-alpha.4

**Console output**  
- TypeError: Load failed (error index.js:1:368579) — see screenshot above

**System**
 - Device: MacBook Pro
 - OS: macOS
 - Browser: Safari
 - Version: 16.5

distantnative:
I can track it back to this change: https://github.com/getkirby/kirby/pull/5313/files

Even though, even when reverting the change, a console error flashes before the redirect happens:

```
panel.js:163 TypeError: Cannot read properties of undefined (reading 'json')
    at Object.get (:3000/src/panel/panel.js:194:19)
    at async Object.open (:3000/src/panel/panel.js:215:19)
    at async Object.load (:3000/src/panel/feature.js:86:4)
```

distantnative:
[...]

Issue #5266, opened 2023-06-11
Panel url redirect issue when running on subfolder


philipptrenz:
## Description

When running Kirby from a subfolder, e.g. `example.com/sub` by configuring the `url` parameter in *config.php*, there occurs a redirect error when logging into the panel at `/sub/panel` or when opening `/sub/panel/site` (cf. screenshots). 

**Expected behavior**  

Not being redirected to `/ndc-sdg/panel/ndc-sdg/panel/site`, which leads to an error, but to `/ndc-sdg/panel/site`, which works as expected when the url is entered manually or is selected from the top menu bar.

**Screenshots**  

<img width="501" alt="Bildschirmfoto 2023-06-11 um 12 50 29" src="https://github.com/getkirby/kirby/assets/11627598/f83d15df-2307-4afc-8cdb-79cdeac5db67">
<img width="1459" alt="Bildschirmfoto 2023-06-11 um 12 45 18" src="https://github.com/getkirby/kirby/assets/11627598/ce1b589a-7ab4-4e35-9512-1cbb1fdc286f">

### Configure Kirby to run under a subdirectory
(e.g. `example.com/ndc-sdg`)
1. Configure `'url' => '/ndc-sdg'` in *config.php*
2. Use the following *RewriteRules* in *.htaccess*:
```
RewriteRule ^ndc-sdg/assets/?(.*)$ /assets/$1 [L]
RewriteRule ^ndc-sdg/media/?(.*)$ /media/$1 [L]
```
### Open the panel

1. Open `example.com/ndc-sdg/panel`

## Your setup
**Kirby Version**  
3.9.5

**PHP Version**
8.2.6  

**Console output**  
```
site:1     GET http://127.0.0.1:8989/ndc-sdg/panel/ndc-sdg/panel/site 500 (Internal Server Error)
```

**Your system (please complete the following information)**
 - Device: not relevant
 - OS: not relevant
 - Browser: Chrome 
 - Version: 114.0.5735.106

lukasbestle:
@afbora Can you replicate this in your local subfolder setup?

afbora:
@lukasbestle Yes, I can. Works when enter manually `/sub/panel/site` directly. But doesn't work when enter `/sub/panel`. Redirects to `/sub/panel/sub/panel/site`. 

Is there a part you want me to test?

afbora:
[...]

PR #5366, opened 2023-07-11
Fix user blueprint home option and running on subfolder

afbora:
## This PR …

After much debugging and testing I finally found a clue to the solution. I hope it's acceptable.

Since there are too many cases, we broke one while fixing the other. Now I think I have fixed both cases, but the HTTP class is too complicated for me. @bastianallgeier will decide how good and stable the fix is 😅 

The critical part of this fix is that if the relative path contains the root directory of the application, it removes it to avoid duplication.

### Fixes
- Fix user blueprint home option #5359 (regression)
- Fix panel url redirect issue when running on subfolder #5266

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Panel/Home.php
+++ b/src/Panel/Home.php
@@ -153,15 +153,6 @@ class Home
 	}
 
 	/**
-	 * Returns the path after /panel/ for the home url
-	 */
-	public static function path(): string|null
-	{
-		$url = Home::url();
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):
This makes a lot of sense to me. Thanks for looking into it.

PR #5368, opened 2023-07-13
Fix `->toUrl()` method for invalid uuids 

### LINKED ISSUE #5367: [v4] `toUrl()` on Link field with Page that got deleted throws error
## Description
Kirby's Url component will throw an error when trying to create the Url for a page that doesn't exist anymore.

**Expected behavior**  
`null` return, no error; similar to when the field would be empty (it's shown empty in the panel as well).

## To reproduce
1. Create a Link Field
2. Set a page in such Link Field
3. Access the URL via `$field->toUrl()`
4. Delete the Page
5. Kirby will throw an error

## Your setup
**Kirby Version**  4.0.0.alpha-4

afbora:
### Fixes
- Fix `->toUrl()` method for invalid uuids #5367

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -404,7 +404,7 @@ return [
 				Uuid::is($path, 'file') === true
 			)
 		) {
-			$path = Uuid::for($path)->model()->url();
+			$path = Uuid::for($path)->model()?->url();
 		}
 
 		$url = Url::makeAbsolute($path, $kirby->url());
--- a/config/components.php
+++ b/config/components.php
@@ -8,6 +8,7 @@ use Kirby\Cms\Page;
 use Kirby\Cms\User;
 use Kirby\Data\Data;
 use Kirby\Email\PHPMailer as Emailer;
+use Kirby\Exception\NotFoundException;
 use Kirby\Filesystem\F;
 use Kirby\Filesystem\Filename;
 use Kirby\Http\Uri;
@@ -349,6 +350,7 @@ return [
 	 *
 	 * @param string|null $path URL path
 	 * @param array|string|null $options Array of options for the Uri class
+	 * @throws \Kirby\Exception\NotFoundException
 	 */
 	'url' => function (
 		App $kirby,
@@ -404,7 +406,13 @@ return [
 				Uuid::is($path, 'file') === true
 			)
 		) {
-			$path = Uuid::for($path)->model()?->url();
+			$model = Uuid::for($path)->model();
+
+			if ($model === null) {
+				throw new NotFoundException('The model could not be found for "' . $path . '" uuid');
+			}
+
+			$path = $model->url();
 		}
 
 		$url = Url::makeAbsolute($path, $kirby->url());
--- a/config/components.php
+++ b/config/components.php
@@ -350,7 +350,7 @@ return [
 	 *
 	 * @param string|null $path URL path
 	 * @param array|string|null $options Array of options for the Uri class
-	 * @throws \Kirby\Exception\NotFoundException
+	 * @throws \Kirby\Exception\NotFoundException If an invalid UUID was passed
 	 */
 	'url' => function (
 		App $kirby,
--- a/config/methods.php
+++ b/config/methods.php
@@ -270,7 +270,11 @@ return function (App $app) {
 		 * Turns the field value into an absolute Url
 		 */
 		'toUrl' => function (Field $field): string|null {
-			return $field->isNotEmpty() ? Url::to($field->value) : null;
+			try {
+				return $field->isNotEmpty() ? Url::to($field->value) : null;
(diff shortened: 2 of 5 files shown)


tests/Cms/Fields/FieldMethodsTest.php
```diff
@@ -446,6 +446,53 @@ public function testToUrlEmpty()
...
+		$file = $app->file('file://my-file-uuid');
+		$field = $this->field('file://my-file-uuid');
+		$this->assertSame('/media/pages/test/' . $file->mediaHash() . '/test.jpg', $field->toUrl());
+		$this->assertSame($file->url(), $field->toUrl());
+	}
+
+	public function testToInvalidUuidUrl()
+	{
+		$field = $this->field('page://invalid');
+		$this->assertSame('/', $field->toUrl());
```
lukasbestle: TBH I feel like this is unexpected. I'd expect `null` here.

afbora: Hmm, the `url` component always returns a string and it is revealed in this component whether the `uuid` is valid or invalid:
https://github.com/getkirby/kirby/blob/v4/fix/5367-toUrl-invalid-uuid/config/components.php#L400-L408

afbora: Alternative solutions:

1. The easiest way is to return `null` for `url` component. This seems to be a big breaking change.
2. We need to send the path to the `Url::to()` method after checking `uuid`. But this time `Url::to()` will no longer have `uuid` support.

lukasbestle: You are right, this is a limitation of `Url::to()` and the `url` component.

Thinking about different options, I feel like there is no ideal solution that is both clean and avoids breaking changes.

Returning `null` seems useful, however there would indeed be quite a breaking change in such a central code location.

A few more options I thought about for the "UUID was passed but invalid" case:

3. Throw a `NotFoundException` (more or less backwards-compatible as there was also a PHP error in this case until now)
4. Return the passed UUID string unchanged (keeps the "always a string" return value, but the UUID may not be a useful value in real use)
5. Return `null` from the `url` component but convert that to one of options 1, 3 or 4 in `Url::to()`. Then `Url::to()` could get a new param to control the behavior. But probably needlessly complex.

So far I favor option 3 TBH. It's the least amount of breaking change, but we could still catch the exception in the `toUrl()` field method and return `null` from there if it gets thrown.

What do you think?

afbora: Option 3 seems ideal. Already throwing an uncontrolled exception. But I have one more idea.

 The `url` component always returns `string`, right?. Except for one case: invalid uuid. If it returns `null` only in the invalid uuid case, we can say that there is almost no breaking change 😇. Anyway, the case that throws an exception now returns `null`. I think it looks like a good bug-fix from the outside 🙂

lukasbestle: The difference between `NotFoundException` and returning `null` is that the exception is clearly visible and the `null` value can be missed. So I feel like `NotFoundException` would be safer as it's easier to adapt other parts of the code (both by us and by site and plugin devs).

afbora: Here the changes https://github.com/getkirby/kirby/pull/5368/commits/abc8965a7e6cccb54c862ee60f6d8128fbbee8dc

afbora: Should we declare breaking change?

bastianallgeier: @afbora I don't think we need to declare it a breaking change. UUID support didn't exist in 3.9.x yet.

config/methods.php
```diff
@@ -270,7 +270,11 @@
 		 * Turns the field value into an absolute Url
 		 */
 		'toUrl' => function (Field $field): string|null {
-			return $field->isNotEmpty() ? Url::to($field->value) : null;
+			try {
+				return $field->isNotEmpty() ? Url::to($field->value) : null;
+			} catch (Throwable) {
```
lukasbestle: To avoid swallowing other kinds of errors:

```suggestion
			} catch (NotFoundException) {
```

Might need `use` at the top of the file.

config/components.php
```diff
@@ -349,6 +350,7 @@
 	 *
 	 * @param string|null $path URL path
 	 * @param array|string|null $options Array of options for the Uri class
+	 * @throws \Kirby\Exception\NotFoundException
```
lukasbestle:
```suggestion
	 * @throws \Kirby\Exception\NotFoundException If an invalid UUID was passed
```

lukasbestle (approved):

Issue #5332, opened 2023-06-28
[v4] Language editor doesn't support keys with spaces


mrflix:
I just looked at the language editor for the first time. Beautiful work!
I created some variables in `languages/de.php` and then explored them there. Turn's out that the UI breaks when the key contains a space. This is the error I see when clicking on **Jetzt anmelden!**:

![Screenshot 2023-06-28 at 15 33 31](https://github.com/getkirby/kirby/assets/60777/98ca6d60-723a-408a-98bd-670f3d81f7c4)

Capitalized keys also "wrongly" get lowercased:

![Screenshot 2023-06-28 at 15 38 34](https://github.com/getkirby/kirby/assets/60777/5a684fbf-3ae9-44a0-8e55-8a6ff38ea63a)

I didn't run in problems yet when using the primary-language value as the key in translations.
I think it's nice that they're very verbose opposed to generic terms like `signup.button`. Those make a lot of sense in big projects but I don't see their need in small projects.

Version: alpha.3

distantnative:
Not quite a bug, but intended to be that way. One can probably argue if it should be different. But supporting more characters isn't 100% trivial as the key itself is used in the request endpoint string.

lukasbestle:
Could we percent-encode it in the URL?

afbora:
I think we can allow all characters `0-9A-Za-z\-\_\.` that can be used in the key. For ex: `error.loginForm-success_text`

This could be all the characters that [PHP array()](https://www.php.net/manual/en/language.types.array.php) allows.

> The key can either be an int or a string. The value can be of any type.
> 
> Additionally the following key casts will occur:
> 
> - Strings containing valid decimal ints, unless the number is preceded by a + sign, will be cast to the int type. E.g. the key "8" will actually be stored under 8. On the other hand "08" will not be cast, as it isn't a valid decimal integer.
> - Floats are also cast to ints, which means that the fractional part will be truncated. E.g. the key 8.7 will actually be stored under 8.
> - Bools are cast to ints, too, i.e. the key true will actually be stored under 1 and the key false under 0.
> - Null will be cast to the empty string, i.e. the key null will actually be stored under "".
> - Arrays and objects can not be used as keys. Doing so will result in a warning: Illegal offset type.

lukasbestle:
@afbora TBH I don't quite understand how arrays play into this. A space would also be a valid character in an array key.

afbora:
[...]

PR #5370, opened 2023-07-13
Improvements language translations editor

afbora:
### Features
- New `languages.variables` option to disable managing translations in the Panel

### Enhancements
- Support any chars in language variable keys

### Fixes
- Language editor doesn't support keys with spaces #5332

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/areas/languages/dialogs.php
+++ b/config/areas/languages/dialogs.php
@@ -46,16 +46,16 @@ $translationDialogFields = [
 	'key' => [
 		'counter' => false,
 		'icon'    => null,
-		'label'   => 'Key',
-		'type'    => 'slug',
-		'allow'   => 'a-z0-9_-',
-		'width'   => '1/3',
+		'label'   => I18n::translate('language.variable.key'),
+		'type'    => 'text',
+		'width'   => '1/1',
 	],
 	'value' => [
 		'counter' => false,
-		'label'   => 'Value',
-		'type'    => 'text',
-		'width'   => '2/3',
+		'buttons' => false,
+		'label'   => I18n::translate('language.variable.value'),
+		'type'    => 'textarea',
+		'width'   => '1/1',
 	]
 ];
 
@@ -208,7 +208,7 @@ return [
 	'language.translation.delete' => [
 		'pattern' => 'languages/(:any)/translations/(:any)/delete',
 		'load'    => function (string $languageCode, string $translationKey) {
-			$variable = Find::language($languageCode)->variable($translationKey);
+			$variable = Find::language($languageCode)->variable($translationKey, true);
 
 			if ($variable->exists() === false) {
 				throw new NotFoundException('The variable could not be found');
@@ -217,18 +217,18 @@ return [
 			return [
 				'component' => 'k-remove-dialog',
 				'props' => [
-					'text' => 'Do you really want to delete the variable for "' . $translationKey . '"?'
+					'text' => 'Do you really want to delete the variable for "' . Escape::html($variable->key()) . '"?'
 				],
 			];
 		},
 		'submit' => function (string $languageCode, string $translationKey) {
-			return Find::language($languageCode)->variable($translationKey)->delete();
+			return Find::language($languageCode)->variable($translationKey, true)->delete();
 		}
 	],
 	'language.translation.update' => [
 		'pattern' => 'languages/(:any)/translations/(:any)/update',
 		'load'    => function (string $languageCode, string $translationKey) use ($translationDialogFields) {
-			$variable = Find::language($languageCode)->variable($translationKey);
+			$variable = Find::language($languageCode)->variable($translationKey, true);
 
 			if ($variable->exists() === false) {
 				throw new NotFoundException('The variable could not be found');
@@ -251,7 +251,7 @@ return [
 			];
 		},
 		'submit' => function (string $languageCode, string $translationKey) {
-			Find::language($languageCode)->variable($translationKey)->update(
+			Find::language($languageCode)->variable($translationKey, true)->update(
 				App::instance()->request()->get('value')
 			);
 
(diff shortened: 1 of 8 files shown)


panel/src/components/Views/LanguageView.vue
```diff
@@ -87,7 +87,7 @@ export default {
 		},
 		option(option, row) {
 			this.$panel.dialog.open(
-				`dialogs/languages/${this.id}/translations/${row.key}/${option}`
+				`dialogs/languages/${this.id}/translations/${encodeURIComponent(window.btoa(row.key))}/${option}`
```
afbora: Just `encodeURIComponent` was not enough. To support HTML tags in addition to normal strings, I developed a solution by first encoding the keys with base64 and then decoding it on the PHP side.

config/areas/languages/dialogs.php
```diff
@@ -217,18 +217,18 @@
 			return [
 				'component' => 'k-remove-dialog',
 				'props' => [
-					'text' => 'Do you really want to delete the variable for "' . $translationKey . '"?'
+					'text' => 'Do you really want to delete the variable for "' . Escape::html($variable->key()) . '"?'
```
distantnative: Do you want to turn these into i18n strings as well if you’re already touching this code?

afbora: Since we have removed any character restrictions, the keys can be used as translations.

The reason for the change in this line is that it is base64 encoded and can have HTML tags.

distantnative: Yes but I mean that this text shouldn’t be hard coded English, an oversight when this was added initially. So when your PR fixes the functionality, could you also switch out the hardcoded English?

afbora: Oh sorry, sometimes I think too complex 😅 I'll update non-translated strings 👍

config/areas/languages/dialogs.php
```diff
@@ -46,16 +46,16 @@
 	'key' => [
 		'counter' => false,
 		'icon'    => null,
-		'label'   => 'Key',
-		'type'    => 'slug',
-		'allow'   => 'a-z0-9_-',
-		'width'   => '1/3',
+		'label'   => I18n::translate('language.variable.key'),
+		'type'    => 'text',
+		'width'   => '1/1',
```
lukasbestle: I think this line can be removed now as full width is the default anyway:

```suggestion
```

config/areas/languages/dialogs.php
```diff
@@ -46,16 +46,16 @@
...
 	],
 	'value' => [
 		'counter' => false,
-		'label'   => 'Value',
-		'type'    => 'text',
-		'width'   => '2/3',
+		'buttons' => false,
+		'label'   => I18n::translate('language.variable.value'),
+		'type'    => 'textarea',
+		'width'   => '1/1',
```
lukasbestle:
```suggestion
```

src/Cms/Language.php
```diff
@@ -566,8 +566,13 @@ public function update(array $props = null): static
 	 * Returns a language variable object
 	 * for the key in the translations array
 	 */
-	public function variable(string $key): LanguageVariable
+	public function variable(string $key, bool $decode = false): LanguageVariable
 	{
+		// decode for encoded strings
+		if ($decode === true) {
+			$key = base64_decode(urldecode($key));
```
lukasbestle: `encodeURIComponent` percent-escapes any special character. Its mirror on the PHP side is `rawurldecode()`. `urldecode()` converts a `+` to a space, `rawurldecode()` doesn't. In this direction (encode in JS, decode in PHP) it's not a problem, I think we should still use `rawurldecode()` for clarity though.

panel/src/components/Views/LanguageView.vue
lukasbestle: I'd be fine with base64, but I'm curious in which cases plain URL encoding would break. I have tried `<script>alert("XSS with spaces")</alert>` (encoded as `%3Cscript%3Ealert(%22XSS%20with%20spaces%22)%3C%2Falert%3E`) and that seemed to work.

Do you have an example for a key that would break without base64 encoding?

lukasbestle:
General comment: I really like the option to disable the variable views. Maybe we could later add a permission for this (like `languages.updateVariables`), but even with a permission the option would IMO still be useful (when the site doesn't use variables at all).

panel/src/components/Views/LanguageView.vue
afbora: image (19)

`<script>alert('XSS');</script>` doesn't for me on windows local.

> Not Found
> The requested URL was not found on this server.
> Apache/2.4.46 (Win64) OpenSSL/1.1.1h PHP/8.1.10 Server at localhost Port 80

afbora:
I like the permissions idea @lukasbestle In addition to completely disabling it, can allow some users and not others. I think more permission would be great:

````yaml
permissions:
    language:
        # enables-disables like new `languages.variables` option
        variables: true
        createVariables: false
        updateVariables: false
        deleteVariables: false
````

For example, for the editor role, the following permission could be used for the editor role to only update variables:

````yaml
title: Editor
permissions:
    language:
        variables: true
        createVariables: false
        updateVariables: true
        deleteVariables: false
````

panel/src/components/Views/LanguageView.vue
lukasbestle: That's weird. There must be something wrong in our router then. On macOS with Caddy and PHP 8.1 it works fine...

Could you add TODO comments to the lines where the Base64 encoding and decoding happens with a short explanation that this is necessary only for compatibility?

afbora: Here the changes https://github.com/getkirby/kirby/pull/5370/commits/1bf1d68fb47933739ccf005bb8df2b0345ebe632

But my explanations may not be very clear. Sorry for that 🙈

afbora:
Closing favor of https://github.com/getkirby/kirby/pull/5387

PR #5374, opened 2023-07-16
[v4] Fix selected link in link dialog

### LINKED ISSUE #5198: Selected link issue for textarea field
## Description

screen-capture-1585-Test - Kirby Demo-localhost

afbora:
## This PR …
With your better knowledge of JS, the code and regex can be improved.

### Fixes
- Fix selected link in link dialog #5198

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Input/TextareaInput.vue
+++ b/panel/src/components/Forms/Input/TextareaInput.vue
@@ -143,7 +143,7 @@ export default {
 			this.$panel.dialog.open({
 				component: "k-toolbar-" + dialog + "-dialog",
 				props: {
-					value: { title: this.selection() }
+					value: this.parseSelection()
 				},
 				on: {
 					cancel: restoreSelection,
@@ -241,6 +241,36 @@ export default {
 		},
 		onSubmit($event) {
 			return this.$emit("submit", $event);
+		},
+		parseSelection() {
+			const selection = this.selection();
+
+			if (selection?.length === 0) {
+				return {
+					href: null,
+					title: null
+				}
+			}
+
+			let regex;
+			if (this.$panel.config.kirbytext) {
+				regex = /^\(link:\s*(?<url>.*?)(?:\s*text:\s*(?<text>.*?))?\)$/is;
+			} else {
+				regex = /^(\[(?<text>.*?)\]\((?<url>.*?)\))|(<(?<link>.*?)>)$/is;
+			}
+
+			const matches = regex.exec(selection);
+			if (matches !== null) {
+				return {
+					href: matches.groups.url ?? matches.groups.link,
+					title: matches.groups.text ?? null
+				}
+			} else {
+				return {
+					href: null,
+					title: selection
+				}
+			}
 		},
 		prepend(prepend) {
 			this.insert(prepend + " " + this.selection());


panel/src/components/Forms/Toolbar/LinkDialog.vue
```diff
@@ -65,6 +61,27 @@ export default {
...
 		},
+		open(input, selection) {
+			this.parseDOM(selection);
+			this.$refs.dialog.open();
+		},
+		parseDOM(selection) {
+			if (selection?.length === 0) {
+				this.value.text = null;
+				this.value.url  = null;
+			}
```
distantnative: Just a quick thought while reading this on my phone: should probably return here to not run the regex still with no selection.

afbora: You're right!

panel/src/components/Forms/Toolbar/LinkDialog.vue
```diff
@@ -65,6 +61,28 @@ export default {
...
+			this.$refs.dialog.open();
+		},
+		parseDOM(selection) {
+			if (selection?.length === 0) {
+				this.value.text = null;
+				this.value.url  = null;
+				return;
+			}
+
+			const regexp = new RegExp(`^\\(link:\\s+?(.*?)\\s+?(?:text:\\s+?(.*?))?\\)$`, "gis");
```
lukasbestle: I think this can be replaced with this to avoid some escaping:

```suggestion
			const regexp = /^\(link:\s+?(.*?)\s+?(?:text:\s+?(.*?))?\)$/gis;
```

However I'd make the following changes:

```suggestion
			const regexp = /^\(link:\s*(.*?)(?:\s*text:\s*(.*?))?\)$/is;
```

Because:

- the spaces after tag names are not required
- greedy matches for the spaces mean that multiple spaces are trimmed away instead of being matched by the following capture group
- global flag is not needed as we match the whole string

panel/src/components/Forms/Toolbar/LinkDialog.vue
```diff
@@ -65,6 +61,28 @@ export default {
...
+			const matches = [...selection.matchAll(regexp)];
+
+			if (matches.length) {
+				this.value.url  = matches[0][1];
+				this.value.text = matches[0]?.[2] ?? null;
```
lukasbestle: A bit simpler:

```suggestion
			const matches = regexp.exec(selection);

			if (matches !== null) {
				this.value.url  = matches[1];
				this.value.text = matches[2] ?? null;
```

afbora:
Thank you @lukasbestle It's much better now. I was sure it was a more readable code 😅

panel/src/components/Forms/Toolbar/LinkDialog.vue
```diff
@@ -65,6 +61,28 @@ export default {
 				return `<${this.value.url}>`;
 			}
 		},
+		open(input, selection) {
+			this.parseDOM(selection);
+			this.$refs.dialog.open();
+		},
+		parseDOM(selection) {
```
distantnative: One last comment I have is the name: is this really DOM-related? Would think more like “parseSelection”

afbora: Actually I inspired from writer nodes 🙈 and that's really a node in there. But this is just string selection as your said. You're right 👍

afbora:
One more question. This fix is only for kirbytext syntax. But link dialog also generates link syntax for markdown. Do we need to do the same fix for markdown too?

lukasbestle:
We don't need to, but you are right that it would be a useful addition while we are at it.

afbora:
@lukasbestle I've implemented for the markdown syntax: https://github.com/getkirby/kirby/pull/5374/commits/3f8e511dd68abf71f7045ff723b49dcd76798ba1

However, I had to use named groups because the order of `text` and `url` match is different in markdown syntax with text and without text (See following possible issue example). I am sure there is a better regex solution.

### Possible issue without named groups
````
<http://wikipedia.org>             -> matches[1] = url
[Wikipedia](http://wikipedia.org)  -> matches[1] = text and matches[2] = url
````

distantnative:
The create methods seem unused.

lukasbestle (approved):
Looks very clean now IMO 😍

PR #5384, opened 2023-07-19
[v4] Expandable drawers

afbora:
## This PR …

https://github.com/getkirby/kirby/assets/3393422/3ff11393-a4c6-45d1-a95e-e84d36c71a14

### Enhancement
- Expandable drawers

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Drawers/BlockDrawer.vue
+++ b/panel/src/components/Drawers/BlockDrawer.vue
@@ -1,7 +1,6 @@
 <template>
 	<k-form-drawer
 		ref="drawer"
-		:expand="isExpanded"
 		class="k-block-drawer"
 		v-bind="$props"
 		@cancel="$emit('cancel', $event)"
@@ -11,11 +10,6 @@
 		@tab="$emit('tab', $event)"
 	>
 		<template #options>
-			<k-button
-				class="k-drawer-option k-drawer-option-expand"
-				:icon="isExpanded ? 'collapse' : 'expand'"
-				@click.prevent.stop="isExpanded = !isExpanded"
-			/>
 			<k-button
 				v-if="hidden"
 				class="k-drawer-option"
@@ -45,9 +39,6 @@ import { props as Fields } from "./Elements/Fields.vue";
 
 export const props = {
 	props: {
-		isExpanded: {
-			type: Boolean
-		},
 		hidden: {
 			type: Boolean
 		},
@@ -75,11 +66,3 @@ export default {
 	]
 };
 </script>
-
-<style>
-@media screen and (max-width: 50rem) {
-	.k-button.k-drawer-option-expand {
-		display: none;
-	}
-}
-</style>
--- a/panel/src/components/Drawers/BlockDrawer.vue
+++ b/panel/src/components/Drawers/BlockDrawer.vue
@@ -2,6 +2,7 @@
 	<k-form-drawer
 		ref="drawer"
 		class="k-block-drawer"
+		:expandable="true"
 		v-bind="$props"
 		@cancel="$emit('cancel', $event)"
 		@crumb="$emit('crumb', $event)"
--- a/panel/src/components/Drawers/BlockDrawer.vue
+++ b/panel/src/components/Drawers/BlockDrawer.vue
@@ -1,6 +1,7 @@
 <template>
 	<k-form-drawer
 		ref="drawer"
+		:expand="isExpanded"
 		class="k-block-drawer"
 		v-bind="$props"
 		@cancel="$emit('cancel', $event)"
@@ -10,6 +11,11 @@
 		@tab="$emit('tab', $event)"
 	>
 		<template #options>
+			<k-button
+				class="k-drawer-option k-drawer-option-expand"
+				:icon="isExpanded ? 'collapse' : 'expand'"
+				@click.prevent.stop="isExpanded = !isExpanded"
+			/>
 			<k-button
 				v-if="hidden"
 				class="k-drawer-option"
@@ -39,6 +45,9 @@ import { props as Fields } from "./Elements/Fields.vue";
 
 export const props = {
 	props: {
+		isExpanded: {
+			type: Boolean
+		},
 		hidden: {
 			type: Boolean
 		},
@@ -66,3 +75,11 @@ export default {
 	]
 };
 </script>
+
+<style>
+@media screen and (max-width: 50rem) {
+	.k-button.k-drawer-option-expand {
+		display: none;
+	}
+}
+</style>
--- a/panel/src/components/Drawers/BlockDrawer.vue
+++ b/panel/src/components/Drawers/BlockDrawer.vue
@@ -2,7 +2,6 @@
 	<k-form-drawer
 		ref="drawer"
 		class="k-block-drawer"
-		:expandable="true"
 		v-bind="$props"
 		@cancel="$emit('cancel', $event)"
 		@crumb="$emit('crumb', $event)"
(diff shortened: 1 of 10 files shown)


afbora:
~I think I should also apply this enhancement to the structure drawer.~ ✅

panel/src/mixins/drawer.js
```diff
@@ -14,6 +14,10 @@ export default {
 			default: false,
 			type: Boolean
 		},
+		expand: {
```
distantnative: I'd suggest `isWide` as prop name here.

panel/src/components/Drawers/BlockDrawer.vue
```diff
@@ -39,6 +45,9 @@ import { props as Fields } from "./Elements/Fields.vue";
 
 export const props = {
 	props: {
+		isExpanded: {
```
distantnative: Needs to be a data entry not a prop here as you want to manipulate it directly in the code above (and Vue does not allow altering props directly).

distantnative (commented):
If you want to support it in more drawers, I think the whole functionality should be built into the low-level drawer. And then a prop that just activates/deactivates the toggle button.

panel/src/components/Drawers/Drawer.vue
```diff
@@ -3,6 +3,7 @@
 		<form
 			:aria-disabled="disabled"
 			:class="$vnode.data.staticClass"
+			:data-expand="isExpanded"
```
distantnative:
```suggestion
			:data-expanded="isExpanded"
```

panel/src/components/Drawers/Drawer.vue
```diff
@@ -80,10 +88,19 @@ export default {
...
 	flex-basis: var(--drawer-width);
 	position: relative;
-	display: flex;
 	flex-direction: column;
 	background: var(--drawer-color-back);
 	box-shadow: var(--drawer-shadow);
 	container-type: inline-size;
 }
+
+.k-drawer[data-expand="true"] {
```
distantnative:
```suggestion
.k-drawer[data-expanded="true"] {
```

panel/src/components/Forms/Blocks/Block.vue
```diff
@@ -264,6 +264,7 @@ export default {
 					submit: this.submit
 				},
 				props: {
+					expandable: true,
```
distantnative: Since this isn't a dynamic value but always the same, it would be better to not pass this in the call but rather declare in `BlockDrawer` the `expandable` prop's default as `true`.

panel/src/components/Forms/Field/StructureField.vue
```diff
@@ -370,6 +370,7 @@ export default {
 			this.$panel.drawer.open({
 				component: "k-structure-drawer",
 				props: {
+					expandable: true,
```
distantnative: Same for `StructureDrawer`

afbora:
Thanks @distantnative Now changes are right?

distantnative:
@afbora not quite what I meant: your commit hardcoded the prop value, I just pushed one that only sets a different default for those drawers.

What do you think about turning the toggle button into a similar version of the menu expand button?
Screenshot 2023-07-19 at 20 56 23

afbora:
It would be great to be compatible with the sidebar 👍 

But a side note; I think this button (sidebar expand/collapse handle) usually means to expand a hidden place (like the sidebar), would that be confusing?

lukasbestle:
I think it would fit quite well. Both buttons "make the box wider" and in the sidebar, that has the result that more information is visible. In the drawer that could also be the case with small columns or custom Panel CSS. So the UX IMO still makes sense with the sidebar-style expand button.

distantnative:
I'll try to push a draft tonight. Already have something local, but I'd like to DRY it a bit between menu and drawer.

distantnative:
Added a draft. But also got worried quite a bit while working on it if we are heading into a good direction here:

Does this make sense like this?
Drawer constantly resets to non-expanded state. Wouldn't we need to store the expanded state? What about nested drawers then? Etc.

I am not sure that without these it's really a good feature. Then again diving into these complexity blows this up so much that I am not sure the UX benefit really justifies the bloat of adding it.

afbora:
Expand does not need to be used in every drawer. It is only used when you need to see the content better or edit it more easily. There is no problem with a single level, but for example when nested field is used, it would be more correct to keep the expand state as you said. 

I didn't expect this issue 😞 the solution I can think of is to keep single and global data in the expand state information. If the drawer is expanded, all expandable drawers are opened wide. I'm not sure how ridiculous this idea is 🙈

lukasbestle:
@afbora I like the idea to keep the expand state globally. It would also avoid that the drawer jumps around when opening a nested drawer. Could we implement it so the expand state is reset to `false` when the last drawer is closed? So as long as drawers are open, the expand state is shared but if a fresh drawer is opened, it starts out in the compact state.

afbora:
I had a new idea:

If we can allow to use a wide drawer according to the data type. Like a new `selector`  option. This is both useful and does not require any state. 

````yaml
fields:
  layout:
    type: layout
    drawer: large
      
  structure:
    type: structure
    drawer: large
      
  blocks:
    type: blocks
    # may be like that to support new option for future
    drawer:
      size: large
````

What do you think? I know you don't want to add new options, but I wanted to express my opinion anyway.

lukasbestle:
I'd say there are arguments for and against making this a blueprint prop.

Pro: Depending on how the drawer contents are layed out (e.g. use of columns), it can sometimes make sense to open the drawer wider and that certainly depends on the individual field (= good fit for a blueprint prop).
Contra: It could be surprising to users that some drawers open on the side and some in a much wider view. Having a global behavior with the expand button would be more consistent. Also it would avoid adding more blueprint props to the already large list.

I'd be fine with both. A prop would definitely be easier to implement as we don't need to globally handle the expanded/contracted states in the frontend. But I'm not sure if it outweighs the cons.

afbora:
Thank @lukasbestle for the valuable feedback. Then the best way to keep it global without a session is to store it in the top level DOM like we do as in overlay `data-overlay="true"`

`<html data-expanded="true">`

Even if it switches between pages, the expand property will keep its last state because the HTML attribute will not change (unless the page is hard refreshed).

I don't want to spend extra time on this in the v4 roadmap. We can implement these ideas for 4.1 or 4.2 if the proposed solutions are suitable.

lukasbestle:
I've reopened the PR as a TODO for later discussion.

distantnative:
@lukasbestle I'd like to close this. Our PRs are currently crowding up with these non-actionable drafts etc. - mostly also my fault. I'd like to get them down again to active developing drafts or ready for review.

lukasbestle:
I don't like to just close things and forget about them. If we do close it (which I'm fine with), we need a new place where discussion can continue. Otherwise all the prior work is lost.

distantnative:
Could be an issue or could be an Nolt. I just know we had it time with too many stalled PRs and recently we managed to stay away from it. I started to put too many not-ready ones up myself and will work on taking them down or getting them active again this weekend.

distantnative:
@afbora would you mind opening an issue or Nolt ticket (either works for me, whichever you prefer) with your latest suggestion and thoughts and backlink to this PR? Might make most sense if you're the owner of it, since you've been the most involved/original author. I don't want this to get lost, but I think it'll also need some time to find a good solution. I'd suggest we pick that up after 4.0.

afbora:
OK, I moved to the Nolt https://kirby.nolt.io/581

PR #5398, opened 2023-07-20
[v4] Page create dialog: allow custom status for anyone

### LINKED ISSUE #5365: [v4] Create listed pages bypassing permissions
## Description
Utilizing the new [page creation dialogs](https://getkirby.com/releases/4.0/page-creation-dialogs) with the [custom page creation status](https://getkirby.com/releases/4.0/page-creation-dialogs#custom-page-creation-status) currently has some limitations.

```yml
create:
  status: listed

options:
  changeStatus: false
```
Assuming the bluprint specified above. When you create a new page you currently get the following error:
```
The status for this page cannot be changed
```

Even though the `changeStatus` option is set to `false` the error was unexpected. It may be beneficial to be able to bypass permissions for that particular occasion. There should be no harm in that as the intention to create a page as listed is still coming from the blueprint itself.

Edit:
The current implementation will still `publish` a page so it is at least `unlisted` and not a `draft` anymore. In some cases, this may be the preferred behavior. But in my opinion both ways should be possible. Perhaps having an additional option to `force` the status might work. In general, such a method or flag in the page method would give developers a bit more flexibility for some edge cases.

```yml
create:
  forceStatus: listed
```

For reference:
https://discord.com/channels/525634039965679616/1111220342522904706/1127913109311459470

distantnative:
- Fixes #5365

### DIFF
--- a/src/Panel/PageCreateDialog.php
+++ b/src/Panel/PageCreateDialog.php
@@ -246,7 +246,12 @@ class PageCreateDialog
 		$status = $this->blueprint()->create()['status'] ?? 'draft';
 
 		if ($status !== 'draft') {
-			$page->changeStatus($status);
+			// grant all permissions as the status is set in the blueprint and
+			// should not be treated as if the user would try to change it
+			$page->kirby()->impersonate(
+				'kirby',
+				fn () => $page->changeStatus($status)
+			);
 		}
 
 		$payload = [


lukasbestle (approved):
I think that's not only the simplest but also the correct solution. If the blueprint defines the target status, it is intended behavior that permissions don't play a role.

Issue #5391, opened 2023-07-19
[v4] Cardlets layout broken in Safari


tobimori:
## Description
![CleanShot 2023-07-19 at 19 34 43@2x](https://github.com/getkirby/kirby/assets/29142128/3d64b368-a631-4214-a180-f99e78409629)

## Your setup
**Kirby Version**  4.0-alpha.5

PR #5412, opened 2023-07-25
Fix cardlets layout in Safari

distantnative:
### Fixes
- #5391

### DIFF
--- a/panel/src/components/Collection/Items.vue
+++ b/panel/src/components/Collection/Items.vue
@@ -151,12 +151,10 @@ export default {
 
 /** Cardlets */
 .k-items[data-layout="cardlets"] {
+	--items-size: 1fr;
 	display: grid;
 	gap: 0.75rem;
-	grid-template-columns: repeat(
-		auto-fill,
-		minmax(var(--items-size, 10rem), 1fr)
-	);
+	grid-template-columns: repeat(auto-fill, minmax(var(--items-size), 1fr));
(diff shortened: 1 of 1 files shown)


bastianallgeier:
This changes the cardlets pretty drastically and they no longer extend to the full width. Is this intended?

distantnative:
Really? No they seemed to work fine for me - then something went wrong

bastianallgeier:
auto-fit and auto-fill do very different things. But after trying around with it in Safari, it seems like it really hates auto-fit and I can't find a way so far to use it.

distantnative:
Exactly what I encountered

distantnative:
@bastianallgeier can you exactly describe where `auto-fit` and `auto-fill` differ in your use cases? Cause I am trying both with Chrome and get 100% the same results for my tests.

bastianallgeier:
It's especially noticeable on bigger screens. 

auto-fit:

Screenshot 2023-08-02 at 10 05 14

auto-fill:

Screenshot 2023-08-02 at 10 05 03

distantnative:
Don't you think in this case auto-fill looks actually even better?

bastianallgeier:
I think it feels better on a large screen if the cardlets expand, but there's not really a good way to fix this right now, so auto-fill is obviously the more straight forward solution. I went for a slightly bigger cardlet size once the space is available.

PR #5424, opened 2023-07-28
[v4] Search permissions

afbora:
### Usage

Partially disable for a role
````yaml
title: Editor

permissions:
  search:
    files: false
    pages: true
    users: false
````

Fully disable for a role
````yaml
title: Editor

permissions:
  search: false
````

Note that if the search is disabled, the menu comes to the top. Is this a problem?

screen-capture-1645-Users - Mægazine-localhost

### Feature
Implements https://kirby.nolt.io/433

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/areas/search.php
+++ b/config/areas/search.php
@@ -2,10 +2,14 @@
 
 use Kirby\Toolkit\I18n;
 
-return function () {
-	return [
-		'icon'    => 'search',
-		'label'   => I18n::translate('search'),
-		'views'   => require __DIR__ . '/search/views.php'
-	];
+return function ($kirby) {
+	if ($kirby->user()?->role()->permissions()->for('search') === true) {
+		return [
+			'icon'    => 'search',
+			'label'   => I18n::translate('search'),
+			'views'   => require __DIR__ . '/search/views.php'
+		];
+	}
+
+	return [];
 };
--- a/config/areas/site/searches.php
+++ b/config/areas/site/searches.php
@@ -4,8 +4,11 @@ use Kirby\Cms\App;
 use Kirby\Toolkit\Escape;
 use Kirby\Toolkit\I18n;
 
-return [
-	'pages' => [
+$kirby = App::instance();
+$searches = [];
+
+if ($kirby->user()?->role()->permissions()->for('search', 'pages') === true) {
+	$searches['pages'] = [
 		'label' => I18n::translate('pages'),
 		'icon'  => 'page',
 		'query' => function (string $query = null, int $limit, int $page) {
@@ -27,8 +30,11 @@ return [
 				'pagination' => $pages->pagination()->toArray()
 			];
 		}
-	],
-	'files' => [
+	];
+}
+
+if ($kirby->user()?->role()->permissions()->for('search', 'files') === true) {
+	$searches['files'] = [
 		'label' => I18n::translate('files'),
 		'icon'  => 'image',
 		'query' => function (string $query = null, int $limit, int $page) {
@@ -52,5 +58,7 @@ return [
 				'pagination' => $files->pagination()->toArray()
 			];
 		}
-	]
-];
+	];
+}
+
+return $searches;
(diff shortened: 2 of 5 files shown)


distantnative:
Why would we have search permissions? Isn't search already regulated through what objects can be listed/access in the Panel for the user and thus also found through search?

afbora:
access/list solves part of it. But it doesn't make sense to show non-functional parts for limited users. For examples;

- I don't want someone without access to users to see the user search
- I don't want a role to be able to search through all files
- I don't want a role that has access to a parent and can add `default` under it to search through all `default` pages.
- I don't want a role with access to only a few pages to be able to search.

I can tell you many more cases of this.

distantnative:
I don't agree with these. I think it's bad UX to prevent people to search form what they've access to.

The only one I agree is the users one. But that shouldn't be an additional permissions setting but we should hide the user search from the dropdown if someone has no access to the users area. Which I actually thought would already be the case through the area logic. But if it still is shown for someone without access to the users area, we should change that.

afbora:
OK. So let me ask you, would it make sense for you to be able to enable/disable search (without type-based permissions)? Like that;

````yaml
title: Editor

permissions:
  access:
      search: false
````

lukasbestle:
I agree with Nico. The search view and permission should be directly tied to the access permissions. If I can list and read content, I should be able to search through it as well.

afbora:
OK, So search types should be connected access permissions. What do you think enable/disable search feature via permissions?

lukasbestle:
@HYR @vauvarin Since you suggested disabling search by role: Could you maybe share your use case for this?

vauvarin:
Hi, 

I often need to create specific panels for several roles. For example an admin and a candidate. The candidate can create application files (Kirby pages) and when he connects to his panel, I filter the pages so that he only sees his application files. In this case, I don't want him to be able to use the general search engine because I don't want him to be able to see the titles of other candidates' files. If I want it to be able to search in its pages only, I use the "search" option of the pages section https://getkirby.com/docs/reference/panel/sections/pages#search but I would like to disable the general search for this role.

HYR:
So @vauvarin's use case is similar to mine. It's been a while since the suggestion but if I remember right my use case was the user only had access to a page or so of content. They had no need for the search capability as the user was highly restricted in what they could do. The search was showing information they had no need to see and would actually confuse the situation. I think the majority of the time the search is useful but in some cases it only serves to confuse the user, perhaps making theme believe they have access to more then they actually do. In those cases it's nice to disable/hide the search for that role.

distantnative:
So my concern here is that if users would see things in search that they shouldn't (and thus would confuse them), they actually would've access to these things. Which is the actual problem then. One should be able to configure things so that they only have access to what they should have and that should only be what appears in search.

Just hiding search sounds to me like we'd incentivize false security.

kiahreading:
Hello,

My use case is very similar to @vauvarin's. When providing a limited version of the panel to a subset of public users outside of the organisation, it would be great to be able to fully disable search for this user role. This would simplify their experience within the panel and control expectations of access for this user.

@distantnative, I agree that if only implementing this and not taking adequate care with individual page permissions this could incentivize false security. However if **both**, were utilised, I think the outcome would be a better experience overall.

lukasbestle:
Thanks for your input everyone.

I agree with Nico: Search results should only contain items that the user has access to. That's top priority.

On top of that, I get the idea to disable search for specific roles. But TBH it doesn't feel like a permission to me, more like a "UX feature hiding measure". If search can only find items the user has access to, disabling the search is not a security measure that fits with the term "permission". So I'm not sure if the semantics make sense.

afbora:
Then the search part/view should be fixed according to the permissions.

What else besides permissions can be used to completely disable search by role?

lukasbestle:
Hm, that's a good question. Maybe it is a permission after all, I'm just not sure.

PR #5425, opened 2023-07-28
[v4] Load area views dynamically

afbora:
## This PR …

I think it will add convenience and flexibility in terms of code readability. If you wish, we can also support closure instead of just bool data.

Feel free to close the PR if not acceptable 👍 

### Refactoring
- Load area `views` dynamically with new `when` prop

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Panel/Panel.php
+++ b/src/Panel/Panel.php
@@ -474,6 +474,7 @@ class Panel
 			$view['type'] = 'view';
 
 			$when = $view['when'] ?? null;
+			unset($view['when']);
 
 			// checks if there is a load view condition
 			if ($when instanceof Closure) {
@@ -483,8 +484,6 @@ class Panel
 			} else {
 				$routes[] = $view;
 			}
-
-			unset($view['when']);
 		}
 
 		return $routes;
--- a/src/Panel/Panel.php
+++ b/src/Panel/Panel.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Panel;
 
+use Closure;
 use Kirby\Cms\App;
 use Kirby\Cms\Url as CmsUrl;
 use Kirby\Cms\User;
@@ -472,8 +473,14 @@ class Panel
 			$view['area'] = $areaId;
 			$view['type'] = 'view';
 
-			// loads the view only if it meets the condition
-			if (($view['when'] ?? true) === true) {
+			$when = $view['when'] ?? null;
+
+			// checks if there is a load view condition
+			if ($when instanceof Closure) {
+				if ($when($view, $area) === true) {
+					$routes[] = $view;
+				}
+			} else {
 				$routes[] = $view;
 			}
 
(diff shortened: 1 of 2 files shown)


bastianallgeier:
I would wait for comments from the others as well, but I like the idea. Especially since we could use this for our own areas as well.

distantnative:
I think right now the name would be a bit confusing as people will expect to put in a Kirby query. Or at least a closure.

afbora:
This is possible. You also want it to support `closure` function, right?
What are your name suggestions? My other name idea is `if`, for ex:

````php
return [
  'language' => [
    'pattern' => 'languages/(:any)',
    
    'if'      => App::instance()->option('languages.variables', true) !== false,
        
    'if'      => function () {
      return App::instance()->option('languages.variables', true) !== false;
    },
        
    'action'  => function (string $code) {
      //
    }
  ]
];
````

**Comment by distantnative** [comment]
I think as a closure `when` might work. I'm thinking we should maybe even only support closures, not direct boolean values. With direct values, option calls can be very common and I recall that we have had nasty bugs with this where options might not always be ready when those are already called. So enforcing Closures could be the cautious way forward here.

File: src/Panel/Panel.php [hunk 1]
```diff
@@ -471,7 +472,19 @@ public static function routesForViews(string $areaId, array $area): array
...
+			// checks if there is a load view condition
+			if ($when instanceof Closure) {
+				if ($when($view, $area) === true) {
+					$routes[] = $view;
+				}
+			} else {
+				$routes[] = $view;
+			}
+
+			unset($view['when']);
```
lukasbestle: This won't take effect on the `$routes` array as the `$view` array is copied to `$routes` by value, not by reference.

We could solve this by adding `$key => $view` in line 472, using that `$key` in lines 481 and 484 and then doing this here:

```suggestion
			unset($routes[$key]['when']);
```

lukasbestle (commented):
Really clean, I like it. One comment:

src/Panel/Panel.php
afbora: Correct. But I've fixed with another simple way 🙂

lukasbestle (approved):
That's even cleaner. 👍

distantnative:
Could this be

```php
if ($when instanceof Closure === false || $when(...) === true)
```

?

lukasbestle:
It could be, but I feel like this would become too cryptic. When reading this later, we would first need to understand which cases there can be. With an explanatory comment I'd be fine with it though. Maybe something like:

```php
// enable the route by default, but if there is a
// when condition closure, it must return `true`
if (
	$when instanceof Closure === false ||
	$when($view, $area) === true
) {
	$routes[] = $view;
}
```

afbora:
Yes, more simple. But if `when` is not a closure an exception will be thrown. I guess this is developer responsibility.

lukasbestle:
Why that? If it's not a closure, the second part of the expression after `||` will not be evaluated.

afbora:
You can push the changes to the branch. I'm not able to do right now.

lukasbestle (approved):

distantnative:
Ah merged it and only then realized that this doesn't have any unit test. Will try to add one manually.

PR #5429, opened 2023-07-29
Fix nullable search query

### LINKED ISSUE #5428: Search component broken when query is `null`
## Description

> Kirby\Cms\Core::{closure}(): Argument #3 ($query) must be of type string, null given, called in /kirby/src/Cms/Search.php on line 40

Regression PR: [Clean up search component](https://github.com/getkirby/kirby/pull/5297)

## To reproduce
````php
$results = $site->search(get('not-exists-param'), 'title|text');
````

## Your setup
Kirby 3.9.6

### PR DESCRIPTION
**Pull request by afbora**
### Fixes
- Fix nullable search query #5428

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -154,7 +154,6 @@ return [
 			'words'     => false,
 		];
 
-		$collection = clone $collection;
 		$options    = array_merge($defaults, $params);
 		$query      = trim($query ?? '');
 
--- a/config/components.php
+++ b/config/components.php
@@ -140,16 +140,9 @@ return [
 	'search' => function (
 		App $kirby,
 		Collection $collection,
-		string $query = null,
+		string|null $query = '',
 		array|string $params = []
 	): Collection|bool {
-		$query = trim($query ?? '');
-
-		// empty search query
-		if (empty($query) === true) {
-			return $collection->limit(0);
-		}
-
 		if (is_string($params) === true) {
 			$params = ['fields' => Str::split($params, '|')];
 		}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: config/components.php [hunk 1]
```diff
@@ -140,9 +140,16 @@
...
 		Collection $collection,
-		string $query = '',
-		$params = []
+		string $query = null,
+		array|string $params = []
 	): Collection|bool {
+		$query = trim($query ?? '');
+
+		// empty search query
+		if (empty($query) === true) {
```
lukasbestle: Could we also cover the "shorter than minlength" check from line 168 here? Or rather delete this additional check and move the `clone $collection` in line 164 after the check in line 168? This avoids duplicate code.

config/components.php
```diff
@@ -140,9 +140,16 @@
 	'search' => function (
 		App $kirby,
 		Collection $collection,
-		string $query = '',
-		$params = []
+		string $query = null,
```
lukasbestle: This should still primarily be a string IMO, but a nullable one:

```suggestion
		string|null $query = '',
```

config/components.php
```diff
@@ -154,9 +154,9 @@
 			'words'     => false,
 		];
 
-		$collection  = clone $collection;
-		$options     = array_merge($defaults, $params);
-		$query       = trim($query);
+		$collection = clone $collection;
```
lukasbestle: I think this clone operation can be removed completely as both `$collection->limit()` (via `$collection->slice()`) and `$collection->filter()` also clone the collection before any operation. Removing this line would increase performance for large collections.

```suggestion
```

bastianallgeier (approved):

bastianallgeier (approved):

bastianallgeier (approved):

PR #5431, opened 2023-07-29
New `panel.frameAncestors` option

lukasbestle:
According to the [OWASP clickjacking cheat sheet](https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html), the `'self'` value should only be used if necessary, `'none'` should be the default. So I've added an option like proposed by @afbora that allows to override the default behavior.

### Features

- The `Content-Security-Policy: frame-ancestors` header sent by the Panel (introduced in 3.9.6) can now be customized with an option if needed:
  ```php
  <?php
  
  return [
    'panel' => [
      // allow frame embedding from the same domain
      'frameAncestors' => true,
      
      // allow frame embedding from the same *and* from the specified domains
      'frameAncestors' => ['*.example.com', 'https://example.com'],
      
      // allow frame embedding on any domain (not recommended)
      'frameAncestors' => '*',
    ]
  ];
  ```

## Ready?

### For review team

### DIFF
--- a/src/Panel/Document.php
+++ b/src/Panel/Document.php
@@ -278,8 +278,16 @@ class Document
 			'panelUrl' => $uri->path()->toString(true) . '/',
 		]);
 
+		$frameAncestors = $kirby->option('panel.frameAncestors');
+		$frameAncestors = match (true) {
+			$frameAncestors === true   => "'self'",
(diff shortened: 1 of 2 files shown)


afbora (approved):
I've updated the description with adding an example about allowing frame for any domains. 

Works great and LGTM 👍

PR #5432, opened 2023-07-30
Re-populate uuid when invalid model id cached

### LINKED ISSUE #5430: Uuid not corrected when cached
## Description
If a cached UUID points to a wrong id (pseudo folder path), it is not corrected when retrieved. 
However, when a UUID is not cached, it is created upon retrieval. 

**Expected behavior**  
The cached value should be updated when it is found to be incorrect. Just like when it is not found. 
The wrong cache could at least be removed, so that it's corrected one page load later. 

## To reproduce
1. have a page
2. make sure a UUID is generated
3. change the page's folder path manually in your explorer / finder
4. search for the page with `page('page://myuuid')`, a page will be found anyway (although by doing a full index scan)
5. search for the page with `page('page://myuuid')` again, a full index scan is done again

## Your setup
**Kirby Version**  
I've noticed this in v4, but I guess this is the same in earlier versions

**Additional context**
Deleting the cache resolves the problem.

afbora:
## This PR …

If a cached model cannot be found in the caches, it will force it to update its new value. 

### Fixes
- Uuid not corrected when cached #5430

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Uuid/Uuid.php
+++ b/src/Uuid/Uuid.php
@@ -309,7 +309,8 @@ class Uuid
 				// lazily fill cache by writing to cache
 				// whenever looked up from index to speed
 				// up future lookups of the same UUID
-				$this->populate();
+				// also force to update value again if it is already cached
+				$this->populate($this->isCached());
 
 				return $this->model;
 			}
@@ -320,12 +321,10 @@ class Uuid
 
 	/**
 	 * Feeds the UUID into the cache
-	 *
-	 * @return bool
 	 */
-	public function populate(): bool
+	public function populate(bool $force = false): bool
 	{
-		if ($this->isCached() === true) {
+		if ($force === false && $this->isCached() === true) {
 			return true;
 		}
 
--- a/src/Uuid/SiteUuid.php
+++ b/src/Uuid/SiteUuid.php
@@ -46,7 +46,7 @@ class SiteUuid extends Uuid
 	/**
 	 * Pretends to fill cache - we don't need it in cache
 	 */
-	public function populate(): bool
+	public function populate(bool $force = false): bool
 	{
 		return true;
 	}
--- a/src/Uuid/UserUuid.php
+++ b/src/Uuid/UserUuid.php
@@ -46,7 +46,7 @@ class UserUuid extends Uuid
 	/**
 	 * Pretends to fill cache - we don't need it in cache
 	 */
-	public function populate(): bool
+	public function populate(bool $force = false): bool
 	{
 		return true;
 	}
--- a/tests/Uuid/UuidTest.php
+++ b/tests/Uuid/UuidTest.php
@@ -3,7 +3,6 @@
 namespace Kirby\Uuid;
 
 use Generator;
-use Kirby\Cms\Page;
 use Kirby\Exception\LogicException;
 use Kirby\Toolkit\Str;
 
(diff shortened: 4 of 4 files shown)


distantnative:
@afbora we should then use something like `Dir::move()` for a test. This change shouldn't go in without unit tests.

afbora:
I already tried with `Dir::move()` but I couldn't re-instance the app after the page moved.

distantnative:
I guess you could also test it by just writing manually a wrong cache entry

afbora:
@distantnative I've found a way. Could you check out please? https://github.com/getkirby/kirby/pull/5432/commits/e9f241c5d95634d23a53aa582bdcb40be32d12a0 I hope it correct.

distantnative:
@afbora I'm not sure about the second half of the test, cause it's more about testing what's in the cache but not testing what happens when an invalid cache value exists.

I think I'd go the route I mentioned before: alter manually the cached value and then test if Kirby behaves correctly when trying to retrieve a page form this invalidly cached uuid

afbora:
I'm sure that the unit test checks the correct part. Because the unit test fails when I revert the relevant changes (https://github.com/getkirby/kirby/pull/5432/commits/fc50638c892d3bd1e659f5f09d750e3e184ad5c8). Doesn't this show that the unit test is correct?

distantnative:
The problem is that it more tests our specific implementation (what's in the cache) than the functionality (what happens if a cache value is invalid).

We have that in a lot of places, where we test too much just for our exact implementations and if we ever change them all those tests fail.

afbora:
@distantnative Like that https://github.com/getkirby/kirby/pull/5432/commits/05a99559a6b07fd45888756563792422b5aaa2ee?

distantnative:
@afbora actually not quite. This still doesn't test the behaviour of `Uuid::model()` and you can set the cache easier. I thought of something like

```php
public function testCacheInvalidModelId()
{
	$page = $this->app->page('page-a');
	$key  = 'page/my/-page';
	$id   = 'page://my-page';
	$uuid = Uuid::for($id);

	$this->assertFalse($uuid->isCached());
	$this->assertSame($key, $uuid->key());
	$this->assertSame($uuid->toString(), $id);
	$this->assertTrue($uuid->isCached());
	$this->assertSame(Uuids::cache()->get($key), 'page-a');
	$this->assertSame($page, $this->app->page($id));

	// modify cache data manually to something invalid
	Uuids::cache()->set($key, 'invalid-id');

	$uuid = Uuid::for($id);
	$this->assertSame(Uuids::cache()->get($key), 'invalid-id');
	$this->assertNull($uuid->model(true));
	$this->assertSame(Uuids::cache()->get($key), 'invalid-id');
	$this->assertSame($page, $uuid->model());
	$this->assertSame(Uuids::cache()->get($key), 'page-a');
}
```

distantnative (approved):
I think this looks good now. But given that UUID still is a rather sensitive feature, would be great to have another pair of eyes on this from @lukasbestle

src/Uuid/Uuid.php
```diff
@@ -309,7 +309,8 @@ public function model(bool $lazy = false): Identifiable|null
 				// lazily fill cache by writing to cache
 				// whenever looked up from index to speed
 				// up future lookups of the same UUID
-				$this->populate();
+				// also force to update value again if it is already cached
+				$this->populate($this->isCached());
```
lukasbestle, line 313: We could always pass `true` here. If the UUID is not already cached, the `$force` param in `populate()` doesn't change behavior anyway.

Another thought I had: If the value in the cache was already the same but incorrect, this would write to the cache over and over on every request (resulting in wear on the drive, especially SSDs). This edge case could only happen if there was an incorrect implementation elsewhere (e.g. if `findByCache()` was broken so it would never return a model). If we want to protect against this potential bug, we would need to do this:

- Check if the UUID is already cached.
- If yes, get the value from cache and compare it to the current value.
- If they are the same, throw a `LogicException('Could not find UUID by cache, but the current model ID was already cached before')`.
- Otherwise call `$this->populate(true)` to update the cache with the new value.

Not sure if it's worth it, but it would add extra safety to the implementation.

distantnative: Tbh, I think if we're afraid that `findByCache()` or `findByIndex()` could be incorrectly implemented, we should add more unit tests for these. Not try to mitigate it in other parts if the code.  Those should rely upon `findByCache()` returning the model if its correctly cached as well as `findByIndex()` returning the correct model that holds the uuid.

lukasbestle: I feel like this is a kind of bug that is really hard to catch in unit tests, but rather easy to catch at runtime.

However I just realized that my proposed fix would be affected by a race condition if the cache is updated by another thread. So it's not going to work.

I'll see if we can build a unit test to verify this edge case.

Issue #5427, opened 2023-07-28
[v4] You can't use an object in a structure field


samzzi:
## Description
If you add a field of type object in a structure the panel doesn't seem to be able to save its values.

**Expected behavior**  
To be able to save the values of the object. It does work in 3.x

## To reproduce

Add following blueprint:

```yaml
fields:
  mystructure:
    type: structure
    fields:
      myfield:
        type: object
        fields:
          title:
            type: text
          label:
            type: text
```

And try to fill the object via the panel.

## Your setup
**Kirby Version**  
4.0.0-alpha.5

**Your system (please complete the following information)**
 - Device: Macbook pro
 - OS: macOS Ventura 13.5
 - Browser: Chrome
 - Version: 115

samzzi:
I believe this might be related to https://github.com/getkirby/kirby/issues/5385

Issue #5411, opened 2023-07-25
[v4] Props passed to Drawer components are not reactive


ovenum:
## Description

When Using the new `$panel.drawer.open()` method from within a Field, the properties of the props object loose their reactivity.

**Expected behavior**  

The props to behave just like Vue component props or to rename the props Drawer config option to avoid confusion. 

**Screenshots**  

## To reproduce

Create a Field Plugin 
```php
<?php

use Kirby\Cms\App as Kirby;

Kirby::plugin('namespace/plugin', [
  'fields' => [
    'customobject'  => [
      'extends' => 'object'
      ]
    ]
  ]);

```

```js
panel.plugin("namespace/plugin", {
  fields: {
    customobject: {
      extends: "k-object-field",

      computed: {
        customFieldProperty() {
          return this.object;
        },
      },

      methods: {
        open(field) {
          console.log("Open Form Drawer");
          if (this.disabled) {
            return false;
          }

          this.$panel.drawer.open({
            component: "k-custom-form-drawer",
            props: {
              // Passing down the property to the drawer
              customProperty: this.customFieldProperty,

              breadcrumb: [],
              icon: "box",
              tab: "object",
              tabs: {
                object: {
                  fields: this.$helper.field.subfields(this, this.fields),
                },
              },
              title: this.label,
              value: this.object,

            },
            on: {
              input: this.onDrawerInput.bind(this),
            },
          });
        },
      },
    },
  },

  components: {
    "k-custom-form-drawer": {
      extends: "k-form-drawer",
      props: {
        customProperty: Boolean,
      },

      watch: {
       // Should log the ObjectField object property to the console when a field value changes.
        customProperty(value) {
          console.log("customProperty", value);
        },
      },
    },
  },
});

```

And a page blueprint using the the created custom field
```yaml
fields:
  drawerPropsTestField:
    type: customobject
    fields:
      text:
        type: writer
```

When editing the a field of the custom field, the expected drawer behaviour should be a console.log message. But the drawer does not pick up on the change.  Watching `value`  (also passed as prop) from the drawer will work.

## Your setup
**Kirby Version**  
Kirby 4.0.0-alpha.5

## Further Notes
Just for reference the [Discord Thread](https://discord.com/channels/525634039965679616/1133324759975870525) where the issue was found. Please use the code from the here and not the Discord thread to reproduce the issue.

ovenum:
[...]

Issue #5385, opened 2023-07-19
[v4] Nested structure loses content


medienbaecker:
## Description
When using a structure field inside of a structure field, the content of the entries in the nested structure field seems to get lost.

![nested-structure](https://github.com/getkirby/kirby/assets/7975568/21c3f384-e79a-49c2-a4be-822237b26c91)

## To reproduce

1. Create a nested structure field:
```yml
structure:
    fields:
      structure:
        label: Nested Structure
        type: structure
        fields:
          text: true
```
2. Add a structure entry to the top-level structure field
3. Add an entry to the nested structure field with some content
4. Close the drawer, click the little checkmark button or hit ⌘S 
5. The content is gone

## Your setup
**Kirby Version**  
4.0.0-alpha.5

**Your system (please complete the following information)**
 - OS: macOS Ventura 13.4.1
 - Browser: Chrome
 - Version: 115

bastianallgeier:
I've just set up a test setup to reproduce it and unfortunately it happens for me as well. Just to confirm, do you already notice that the content will stay as soon as you try again?

medienbaecker:
> Just to confirm, do you already notice that the content will stay as soon as you try again?

@bastianallgeier Sorry, I'm afraid I don't understand your question. It looks like the content is just cosmetic. When editing the nested structure field I don't even get the "Revert/Save" prompt.

afbora:
As far as I have tested, the last added row is not emitted. The input is triggered but not saved to localStorage. But I could not find the cause of the issue.

MathiasGmeiner:
For me it happens for nested blocks as well.

bastianallgeier:
@MathiasGmeiner nested blocks in a structure field?

MathiasGmeiner:
@bastianallgeier no, sorry for being so vague. I'm dealing with nested blocks in a blocks field. Could this be a different issue altogether? My first try was a structure field within a block field, and then I tried a block inside a block field.

bastianallgeier:
@MathiasGmeiner I'll try to reproduce this, but in our current nested blocks test setup I didn't notice this so far.

MathiasGmeiner:
@bastianallgeier ok, let me know if I can help.

vincentrouleau:
I have the same issue in nested structures. I can send you my BPs if needed.

MathiasGmeiner:
I'm not sure if I'm doing something wrong with the nested block setup. 
This is my setup:

test.yml
```yml
[...]

PR #5437, opened 2023-08-01
Fix for nested drawers

bastianallgeier:
This is a complicated beast. Let me try to break it down: 

- Nested drawers all get rendered but layered on top of each other. Only the top-most drawer is visible.
- The top-most drawer gets all the event listeners that would submit, close, update or cancel the drawer
- So far none of the underlying drawers did get any event listeners. But this would mean that the top most drawer could not send its input event to the parent drawer above and thus the input event does not propagate. This leads to incorrect values. 

My first reflex was to add the event listeners to all drawers, but this leads to infinite loops. When only the custom event listeners are added, the fields can still receive input calls, but those don't interfere with other drawer logic. 

It feels pretty clean to me now, but the nested drawers remain a really tricky case. 

I fixed and refactored a few more things along the way to get object and structure fields as similar as possible. 

## Fixes

- https://github.com/getkirby/kirby/issues/5385
- https://github.com/getkirby/kirby/issues/5427
- https://github.com/getkirby/kirby/issues/5411

### DIFF
--- a/panel/src/components/Drawers/FiberDrawer.vue
+++ b/panel/src/components/Drawers/FiberDrawer.vue
@@ -8,7 +8,7 @@
 			:disabled="isCurrent(drawer.id) === false"
 			:visible="true"
 			v-bind="drawer.props"
-			v-on="isCurrent(drawer.id) ? $panel.drawer.listeners() : {}"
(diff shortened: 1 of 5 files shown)


afbora:
I've tested nested structure, object, blocks field and all works great for me 🎉

PR #5439, opened 2023-08-01
Fix slug input

### LINKED ISSUE #5423: [v4] 'slugs' => 'de' broken
## Description
It seems that the config option `'slugs' => 'de'` is broken in 4.0.0-alpha.5. When I activate it and create a new page with the title "Über uns", the auto-created slug is "uber-uns".

**Expected behavior**  
I'd expect the slug to be "ueber-uns".

**Screenshots**  
Screenshot 2023-07-28 at 16 27 56

**Kirby Version**   
4.0.0-alpha.5

distantnative:
## This PR …

Since `this.$language`/`this.$panel.language` now always is an object, we need to check for `this.$panel.language.code` to check if an actual language object was loaded.

### Fixes regression
- https://github.com/getkirby/kirby/issues/5423

### DIFF
--- a/panel/src/components/Forms/Input/SlugInput.vue
+++ b/panel/src/components/Forms/Input/SlugInput.vue
@@ -51,7 +51,9 @@ export default {
 	data() {
 		return {
 			slug: this.sluggify(this.value),
-			slugs: this.$language ? this.$language.rules : this.$panel.system.slugs,
+			slugs: this.$panel.language.code
+				? this.$panel.language.rules
+				: this.$panel.system.slugs,
 			syncValue: null
 		};
 	},
--- a/panel/src/components/Forms/Input/SlugInput.vue
+++ b/panel/src/components/Forms/Input/SlugInput.vue
@@ -51,7 +51,7 @@ export default {
 	data() {
 		return {
 			slug: this.sluggify(this.value),
-			slugs: this.$panel.language.code
+			slugs: this.$panel.multilang
 				? this.$panel.language.rules
 				: this.$panel.system.slugs,
 			syncValue: null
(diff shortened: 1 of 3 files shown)


afbora:
I was definitely going to implement this solution but I didn't create a PR because I thought it would be a temporary solution. I don't think checking with `this.$panel.language.code` is a logical solution. If the `language` object doesn't return `null`, I think it is better to proceed with a new method like `this.$panel.multilang`.

distantnative:
Is the check really about multilang yes or no? I was thinking about this too but was unsure I would overlook maybe another use case.

afbora:
`App::multilang()` means there is at least one language. If there is at least one language, doesn't that mean that `this.$panel.language` object is filled? If no, then you're right.

distantnative:
Cause `this.$panel.multilang` does already exist, so this would be easy to use.

afbora:
Unfortunately `this.$panel.multilang` doesn't work. because this prop value coming from `Panel::multilang()` not `App::multilang()`. `Panel::multilang()` checks also `languages` option. 

So if the `languages` option is enabled and there are no added languages, the `this.$panel.multilang` prop will return `true`.  As far as I understand, the purpose of the `Panel::multilang()` method is to enable/disable the panel languages view.

I'm not sure which way is the right way forward.

distantnative:
Ok then I'd go back to `panel.language.code`

bastianallgeier:
What about panel.languages.length?

bastianallgeier:
I think the default value for language.rules was wrong. I used an array for that, but the rules are actually an object. Looking at it again, I think it makes most sense to set the rules to null if they don't exist. Then the code for the slug input is super simple.

distantnative:
👍

Issue #5393, opened 2023-07-19
[v4] Drawer buttons bar right padding issue


afbora:
There is a unnecessary padding on right of drawer buttons bar

![image (23)](https://github.com/getkirby/kirby/assets/3393422/6c0db1a2-0b6a-4b6e-b439-b8b17c4e4bd8)

distantnative:
@afbora is this due to `scrollbar-gutter: stable` on `html`?

afbora:
@distantnative Yes. I have no idea about `scrollbar-gutter: stable` but fixes when I removed.

bastianallgeier:
The scrollbar-gutter: stable rule makes sure that the main layout doesn't shift when you open a dialog and overflow on the body is hidden.

distantnative:
@bastianallgeier What does this mean then? I don't think we can remove `crollbar-gutter: stable` in that case. Is there a way to still style it? Or are we fine with everyone with actual scrollbars (no floating Mac ones) to have that gap?

distantnative:
Maybe we could just add a pseudo-element `::after` the header that would have the white bar fill the gap.

afbora:
Note that `scrollbar-gutter: stable` removes drawer scroll bar.

afbora:
Some browsers still doesn't support:
https://caniuse.com/mdn-css_properties_scrollbar-gutter

PR #5443, opened 2023-08-02
Fix scroll bar issues in drawers and dialogs

bastianallgeier:
I think I have a solid solution for scroll bars on the body and in the drawer. This no longer relies on the new scrollbar-gutter rule, but will always force a scrollbar on the document and still hide it properly once an overlay is opened. 

Since we cannot use :has so far, the modal.js module will now add the data-overlay attribute to the html element in JS. This needs a bit more work and I had to introduce a new `panel.overlays()` method to count the number of open overlays. Otherwise it could happen that a drawer and a dialog is open, but once the dialog is closed the scroll lock is incorrectly disabled already. 

I've tested it all in Firefox, Safari and Chrome and with enabled and disabled scroll bars on macOS and it feels very nice to me. @afbora it would be awesome if you could give it another look on Windows. 

## Fixes

https://github.com/getkirby/kirby/issues/5393

### DIFF
--- a/panel/src/components/Layout/Overlay.vue
+++ b/panel/src/components/Layout/Overlay.vue
@@ -176,7 +176,10 @@ export default {
 }
 
 /* Scroll lock */
-html:has(.k-overlay[open]) {
+html[data-overlay] {
 	overflow: hidden;
+}
+html[data-overlay] body {
+	overflow: scroll;
 }
 </style>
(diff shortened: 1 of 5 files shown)


afbora (approved):
I can't even see the slightest problem. It looks great for me 🎉

PR #5461, opened 2023-08-05
Upload preview for SVGs

### LINKED ISSUE #5436: [v4] svg preview in file upload & file replace not working
## Description
When uploading and replacing files, no preview gets loaded for svg files.

**Expected behavior**  
I'd expect to be able to preview the svg file.

**Screenshots**  
Screenshot 2023-08-01 at 11 04 16

Screenshot 2023-08-01 at 11 00 34

**Kirby Version**  
4.0.0-alpha.5

distantnative:
### Fixes
- Upload dialog: preview for other file types than images

### Enhancements
- Upload dialog: preview SVG files
#5436

### DIFF
--- a/panel/src/components/Dialogs/UploadDialog.vue
+++ b/panel/src/components/Dialogs/UploadDialog.vue
@@ -23,13 +23,8 @@
 					>
 						<a :href="file.url" class="k-upload-item-preview" target="_blank">
 							<k-image-frame
-								v-if="file.type.match('(jpg|jpeg|gif|png|webp|avif)')"
+								v-if="isPreviewable(file.type)"
 								:cover="true"
-								:src="file.url"
-								back="pattern"
-							/>
-							<k-image-frame
-								v-else-if="file.type.match('svg')"
 								:src="file.url"
 								back="pattern"
 							/>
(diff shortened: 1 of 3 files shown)


panel/src/components/Dialogs/UploadDialog.vue
```diff
@@ -28,6 +28,11 @@
 								:src="file.url"
 								back="pattern"
 							/>
+							<k-image-frame
+								v-else-if="file.type.match('svg')"
```
lukasbestle: Would this also match a file whose extension only *contains* `svg` (e.g. `svgz` or `abcdesvgabcde`)? Same for the pixel formats above.

distantnative: Good point. I think we need to differentiate between upload and upload replace. The latter matches on the url. There it is on the extension. This one is on the mime type. I think both could have problems as you point out.

distantnative: Probably best to switch to using mime types for both dialogs. And then look for exact matches on the full mime type strings.

distantnative:
```html
<k-image-frame
	v-if="
		[
			'image/jpeg',
			'image/jpg',
			'image/gif',
			'image/png',
			'image/webp',
			'image/avif'
		].includes(file.type)
	"
	:cover="true"
	:src="file.url"
	back="pattern"
/>
<k-image-frame
	v-else-if="file.type === 'image/svg+xml'"
	:src="file.url"
	back="pattern"
/>
```

distantnative: https://github.com/getkirby/kirby/pull/5461/commits/8760ce4236c68595d3f418a96c03a00d8c09974e

afbora (approved):
LGTM

PR #5479, opened 2023-08-13
[v4] More user-friendly interface for uploading multiple files to the Panel

michnhokn:
## This PR provides a more user-friendly interface for uploading files to the Panel.
Especially if you have a long list of files you want to upload at once, the current dialogue in the panel becomes too long to interact with nicely. To optimise this behaviour, the following changes have been made:
* The file list in the dialogue has a maximum height and becomes scrollable when the list becomes too long.
* The upload helper is now async to help you upload 5 files at a time.
* The upload button shows the current progress of the total number of files uploaded. <= to be discussed

new-file-upload-preview

### Breaking changes
- `this.$helper.upload` now returns a promise. 

## Ready?

### For review team

https://getkirby.com/docs/reference/plugins/internals/helpers#reference__this-helper-upload


distantnative:
@michnhokn Thanks for the PR! 💛  While I think some things (e.g. max-height and scrolling list) are a great change, others seem to be quite a step back to me:

1. Why only allow one file uploading at a time and thus making the whole upload take significantly longer?
2. Why not closing the dialog when the whole user-intended action has succeeded? Seems to me like back UX to force the user to do another click.
3. Your PR is also adding a bunch of code style changes that we would not like to see merged (e.g. single quotes, trailing commas, removal of spaces in objects). This also makes it really hard to review as it obscures the actual changes.
4. While I like the loading icon on the submit button - I feel like that's something we should rather try to implement overall for the dialogs, would make sense to have it for any dialog (but can't yet say how feasible this is with different parts that can be called on submit)
5. Please don't include compiled assets. They cause a bunch of merge conflicts along the way. Would be great if you could exclude those again.

michnhokn:
Hi @distantnative ,

Thank you for the great feedback. Since this is my first PR of course I did some things wrong 😅.
Let me get to the points:

1. True, but only until a certain amount of files is reached. I have cases of 700+ files being uploaded and getting the browser to its request limit is not very helpful for example with a lock request still happening. Maybe uploading in bulk (maybe configurable) would be a good compromise?

2. I get the point. Maybe this could be converted into a success message in the panel instead.

3. I will fix the code styles 

4. Okay I see. Might be good to have this panel wide. I did not felt good, implementing it tbh.
Although I would like the user not to be able to trigger a upload multiple times. Especially if multiple files at once are uploaded in the background. Let me know if I somehow can contribute to brainstorming here!

5. Also my bad! Still I would recommend we ignore build assets from git then. As you can see I pushed them with a second commit because I thought it has to be in it. Maybe you could adjust the contribution guidelines. I think it is not in there yet.

🙏

michnhokn:
@distantnative I just did some changes and formatted the code accordingly.
I added bulk uploading with currently 5 files at a time. I think it's the best compromise. What do you think?

I also thought about what you said about the upload button. I think it is fair to have a different logic for the upload dialogue, because unlike the rest of the dialogs it is a process that does things in the background.
Still, it might be an idea to implement some sort of icon-changing behaviour in the future.

lukasbestle:
> Still I would recommend we ignore build assets from git then. As you can see I pushed them with a second commit because I thought it has to be in it. Maybe you could adjust the contribution guidelines. I think it is not in there yet.

Thanks for pointing that out, we indeed don't have that in the contributing guide at the moment. Created a PR: #5480
Unfortunately we can't ignore the dist files in the repo, otherwise Kirby couldn't be installed from a direct ZIP download or as Git submodule without manually building the dist files.

michnhokn:
@lukasbestle Ah I see that makes sense!
Love the PR 👍

distantnative:
@michnhokn 
Thanks for the changes - and apologies already that this PR will likely drag out quite a bit before merging, but I think for us it's just good to not simply merge it but direct the parts and see what fits where and how the best:

- I still think that the loading icon part in the submit button should be a general matter taken care by the `k-dialog-buttons` component. While we have a lot of JS-side stuff going on for uploads, we also have a bunch of other dialogs that all trigger actions (e.g. moving page, deleting...) where a delay might occur. The submit button should always be just triggered once and show a disabled and loading state. I'll try to look into this.
- A success message in the Panel makes a lot of sense to me. Something like `x files uploaded`
- I get the point of 700 files going at ones is an issue. Letting them run in chunks seems like a good idea to me.

panel/src/components/Dialogs/UploadDialog.vue
```diff
@@ -125,6 +154,9 @@ export default {
 .k-upload-items {
 	display: grid;
 	gap: 0.25rem;
+	max-height: 64dvh;
```
distantnative, line 157: `dvh` isn't supported by all browsers yet, we probably need a `vh` fallback above.
Where does 64 come from? Any specific reason?

panel/src/panel/upload.js
```diff
@@ -234,6 +236,12 @@ export default (panel) => {
 				return this.done();
 			}
 
+			this.running = true;
+
+			const bulkFiles = [];
+			let page = 0;
+			let counter = 0;
```
distantnative, line 243: I think we could get this part a bit cleaner.

Chunking the files with something like (untested, but as general idea):

```js
const chunks = []:
while (chunks.length * 5 < this.files.length) {
  chunks.push(this.files.slice(this.chunks.length * 5, 5));
}
```

Also was wondering if https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of could help, but maybe not.

panel/src/components/Dialogs/UploadDialog.vue
michnhokn: Nope, not really. Maybe we can calculate something based on the item height so we make sure the last visible item in the scroll container is cut off so the user knows there is more to come?

panel/src/panel/upload.js
michnhokn: I will look into this. But to be honest JS is not my strongest language :) So I'm thankful getting some feedback on this.

distantnative: Yes, happy to help. Think we need to build this bit by bit.
A first part https://github.com/getkirby/kirby/pull/5486

distantnative:
I've started to split the various parts of into single PRs. Makes it a lot easier to review and consider the changes:

- https://github.com/getkirby/kirby/commit/bc74c2b116d9b154fee66bb6e00acc067f39ef9e
- https://github.com/getkirby/kirby/pull/5486
- https://github.com/getkirby/kirby/pull/5487

If you have the time and willingness to have a look at these too, would be appreciated :)

distantnative:
I tried an implementation with the 5-files chunks. And while I got it to work, it didn't feel right. I don't think chunks is the right way to got. Instead I think it needs to be:

Start with running X files (e.g. 5).
Whenever a files finishes uploading, start the next one.

This will be more performant than having each chunk to wait for it's slowest file until to start any new files. It still would ensure not having more then, e.g. 5, uploads running at the same time.

distantnative:
@bastianallgeier what do you think? I tried the scrollable file list, but it's hard to find a way where this is really obvious and not too hidden through a "perfect" cut off.

bastianallgeier:
@distantnative I'm undecided. I kind of like the new dialog styles without the scrolling dialog body tbh.

distantnative:
@bastianallgeier I actually agree at the moment, I would leave the current behavior.

However, we've added most of the other parts suggested in this PR separately. With https://github.com/getkirby/kirby/pull/5555 adding a few more bits.

I would suggest to close this then. @michnhokn Thanks for the pointers into right directions.

PR #5485, opened 2023-08-16
Structure field: i18n for column label

### LINKED ISSUE #5483: [v4] Structure field: Titles in structure tables in multiple languages not working properly
## Description
When working with columns to show the panel user the necessary date, you can use the `columns` to determine which structure items are initially visible. That's a great option, but the titles of the columns don't seem to accept multiple languages **when you're trying to display a structure table inside a structure table**; the lines below don't give the expected outcome:

```
label:
    nl: Menuitems
    en: Menu items

type: structure

columns:
    dish:
        width: 5/6
        label:
            nl: Gerecht
            en: Dish
    price:
        width: 1/6
        label:
            nl: Prijs
            en: Price
        before: €
        align: right
```

... whereas this works, but doesn't give me the option to write in multiple languages:

**Expected behavior**  
To see the right label, not what's visible on the screenshot below.

```
label:
    nl: Menuitems
    en: Menu items

type: structure

columns:
    dish:
        width: 5/6
        label: Dish
    price:
        width: 1/6
        label: Price
        before: €
        align: right
```

**Screenshots**  
image

## Setup
4.0.0.alpha-6

distantnative:
## This PR …

Needed to refactor the general code a bit to add a fix in just one place, not multiple. I think I found also a cleaner and compact solution instead of the if/else-two loops previous one. Would be great to have close eyes whether the refactoring is messing anything else up.

### Fixes
- Structure field: translate column label correctly
#5483

## Ready?

### For review team

### DIFF
--- a/config/fields/structure.php
+++ b/config/fields/structure.php
@@ -108,55 +108,44 @@ return [
 			return $this->form()->fields()->toArray();
 		},
 		'columns' => function () {
-			$columns = [];
-			$mobile  = 0;
+			$columns   = [];
+			$blueprint = $this->columns;
 
-			if (empty($this->columns) === true) {
-				foreach ($this->fields as $field) {
-					// Skip hidden and unsaveable fields
-					// They should never be included as column
-					if (
-						$field['type'] === 'hidden' ||
-						$field['hidden'] === true ||
-						$field['saveable'] === false
-					) {
-						continue;
-					}
+			// if no custom columns have been defined,
+			// gather all fields as columns
+			if (empty($blueprint) === true) {
+				$fields    = array_column($this->fields, 'name');
+				$blueprint = array_fill_keys($fields, true);
+			}
 
-					$columns[$field['name']] = [
-						'type'  => $field['type'],
-						'label' => $field['label'] ?? $field['name']
-					];
+			foreach ($blueprint as $name => $column) {
+				$field = $this->fields[$name] ?? null;
+
+				// Skip empty, hidden and unsaveable fields
+				// They should never be included as column
+				if (
+					empty($field) === true ||
+					$field['type'] === 'hidden' ||
+					$field['hidden'] === true ||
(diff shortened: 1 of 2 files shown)


config/fields/structure.php
```diff
@@ -108,55 +108,44 @@
...
-					];
+			foreach ($blueprint as $name => $column) {
+				$field = $this->fields[$name] ?? null;
+
+				// Skip empty, hidden and unsaveable fields
+				// They should never be included as column
+				if (
+					empty($field) === true ||
+					$field['type'] === 'hidden' ||
+					$field['hidden'] === true ||
```
lukasbestle: Could there be a use case where a hidden field was explicitly defined as a column? Previously this worked, now the column will be hidden.

lukasbestle (commented):
Really clean and a good refactoring. I noticed one breaking change that we may either want to avoid or at least document in the release notes.

config/fields/structure.php
distantnative: Good catch, but thinking about it I have a hard time to imagine a use case where the field inside the drawer is supposed to be hidden but then shows up in the preview table. 

But in general open to either (allowing it or mentioning as breaking change).

lukasbestle: Maybe it could be used for auto-generated information that should be visible in the table but does not need to be edited and is therefore hidden from the form.

If the code could be easily modified to allow this, I'd prefer that. Otherwise I'm also fine with mentioning it as a breaking change. If users are affected by this, they will surely tell us so.

distantnative: I'd think that even for that a disabled field would make more sense in the drawer than hiding it.

But I added a commit that supports it, I think. Just filters hidden fields from the default columns, but allows them when columns have been specified.

lukasbestle (approved):
Looks correct to me. 👍

PR #5486, opened 2023-08-16
`array.chunks` JS helper

distantnative:
## Enhancements
- New `$helper.array.chunks()` JS function
```js
const input = [1, 2, 3, 4, 5];
const result = chunks(input, 2);
// [[1, 2], [3, 4], [5]]
```


distantnative:
Won't probably need it, closing.

PR #5493, opened 2023-08-20
Pages field preview: fix escaping

### LINKED ISSUE #4041: Pages field preview: ampersand shows as `&amp;`
## Description
It seems, like HTML special characters are escaped twice in the field previews used by structure fields.

**Expected behavior**  
The page title should appear as entered by the user

**Screenshots**  
Bildschirmfoto 2021-12-13 um 18 53 54

## To reproduce
Select any page with an ampersand in its title in a structure field that has a `pages` field.

## Your setup
Kirby Version: 3.6.1.1

**Your system**
 - Device: MacBook Pro 14"
 - OS: macOS 12.0.1
 - Browser: Firefox
 - Version: 95.0

distantnative:
### Enhancements
- New `html` prop for `k-bubble`, `k-bubbles` and `k-bubbles-field-preview`

### Fixes
- Page preview field: fixed escaping
 #4041

### DIFF
--- a/panel/src/components/Forms/Previews/BubblesFieldPreview.vue
+++ b/panel/src/components/Forms/Previews/BubblesFieldPreview.vue
@@ -1,14 +1,15 @@
 <template>
 	<div class="k-bubbles-field-preview" :class="$options.class">
-		<k-bubbles :bubbles="bubbles" />
+		<k-bubbles :bubbles="bubbles" :html="html" />
 	</div>
 </template>
 
 <script>
 import FieldPreview from "@/mixins/forms/fieldPreview.js";
+import { props as Bubbles } from "@/components/Layout/Bubbles.vue";
 
 export default {
-	mixins: [FieldPreview],
+	mixins: [FieldPreview, Bubbles],
 	inheritAttrs: false,
 	props: {
 		value: [Array, String]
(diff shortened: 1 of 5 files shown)


lukasbestle (commented):
TBH I find it risky because the output happens two components down. We still don't have a way to statically check the "escaping chain" in CI (i.e. verify that a use of `v-html` only ever receives a safe value). So if we were ever to introduce HTML output into `k-bubbles` or `k-bubble` for whatever reason, we might miss this source of risky HTML.

It would be different if the output was in the same file as it's less easily missed.

distantnative:
What's the alternative? I don't think we can leave this unfixed for almost 2 years 😬

lukasbestle:
Could we handle the HTML output (either unescape or use `v-html`) directly in `k-bubble`? For example with a `html: true|false` prop that gets passed to `k-bubbles` and down to `k-bubble`?

bastianallgeier:
I like the idea to introduce a new html prop. I think it makes the entire flow more transparent for us in the future.

distantnative:
Implemented it as suggested

bastianallgeier (approved):
I like it. @lukasbestle what do you think?

bastianallgeier (approved):

bastianallgeier:
I already merged it because I think it's far better than what we currently have. We could still make adjustments if needed later.

PR #5494, opened 2023-08-20
Reduce JS `forEach` usage

distantnative:


ovenum:
Just out of pure curiosity: is this for performance optimisation and is `.forEach` really that bad in the context of the kirby panel? Or what is the motivation behind not using forEach functions?

distantnative:
@ovenum More about consistent coding style. I personally think the for loops are also easier to read and overlook

ovenum:
Thanks for the explanation @distantnative! Relieved that forEach has not some hidden problem i’ve never heard of : )

bastianallgeier (approved):

PR #5495, opened 2023-08-20
Fixes for PHP types

distantnative:
Some type-related fixes that came up when I tried strict typing.

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -493,7 +493,7 @@ class Str
 		$decimal = strrchr($value, '.');
 		$decimal = match ($decimal) {
 			false   => 0,
-			default => strlen(substr($decimal, 1))
+			default => strlen($decimal) - 1
 		};
 
 		return number_format((float)$value, $decimal, '.', '');
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -481,13 +481,21 @@ class Str
 		// make sure $value is not null
 		$value ??= '';
 
+		// turn the value into a string
+		$value = (string)$value;
+
 		// Convert exponential to decimal, 1e-8 as 0.00000001
 		if (strpos(strtolower($value), 'e') !== false) {
 			$value = rtrim(sprintf('%.16f', (float)$value), '0');
 		}
 
 		$value   = str_replace(',', '.', $value);
-		$decimal = strlen(substr(strrchr($value, '.'), 1));
+		$decimal = strrchr($value, '.');
+		$decimal = match ($decimal) {
+			false   => 0,
+			default => strlen(substr($decimal, 1))
+		};
+
 		return number_format((float)$value, $decimal, '.', '');
 	}
 
@@ -972,7 +980,7 @@ class Str
 	public static function short(
 		string $string = null,
 		int $length = 0,
-		string $appendix = '…'
+		string|false $appendix = '…'
 	): string {
 		if ($string === null) {
 			return '';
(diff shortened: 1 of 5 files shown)


src/Toolkit/V.php
```diff
@@ -449,7 +449,7 @@ public static function __callStatic(string $method, array $arguments): bool
 	 * Checks if the value matches the given regular expression
 	 */
 	'match' => function ($value, string $pattern): bool {
-		return preg_match($pattern, $value) !== 0;
+		return preg_match($pattern, (string)$value) !== 0;
```
lukasbestle: Shouldn't this actually be:

```suggestion
		return preg_match($pattern, (string)$value) === 1;
```

Otherwise the validator will return `true` if `preg_match()` returns `false` (= if an error occurred).

src/Toolkit/Str.php
```diff
@@ -972,7 +980,7 @@ public static function safeTemplate(
 	public static function short(
 		string $string = null,
 		int $length = 0,
-		string $appendix = '…'
+		string|false $appendix = '…'
```
lukasbestle, line 983: Was `false` as the argument already supported or used before?

src/Toolkit/Str.php
```diff
@@ -481,13 +481,21 @@ public static function float(
...
 		if (strpos(strtolower($value), 'e') !== false) {
 			$value = rtrim(sprintf('%.16f', (float)$value), '0');
 		}
 
 		$value   = str_replace(',', '.', $value);
-		$decimal = strlen(substr(strrchr($value, '.'), 1));
+		$decimal = strrchr($value, '.');
+		$decimal = match ($decimal) {
+			false   => 0,
+			default => strlen(substr($decimal, 1))
```
lukasbestle: This can be shortened to:

```suggestion
			default => strlen($decimal) - 1
```

src/Toolkit/V.php
distantnative: I think you're right. Didn't question it since I hadn't touched that part, but especially with strict comparisons `false` would be an issue now.

src/Toolkit/Str.php
distantnative: Yes, not in tests but in `Str::slug()`: https://github.com/getkirby/kirby/blob/v4/develop/src/Toolkit/Str.php#L1114

distantnative:
@lukasbestle interestingly https://github.com/getkirby/kirby/pull/5495/commits/581d10d1baf34dd195252c0c3fbcd6855450bd98 seems to let the unit tests crash under PHP 8

bastianallgeier:
@lukasbestle does anything speak against reverting to substring?

lukasbestle:
I don't think reverting helps us here. We are scraping so close under the memory limit that basically any change will make us go over it. See our memory leak thread in Discord.

distantnative:
Can we then just merge this PR as we need to fix the test/memory issue separately?

bastianallgeier (approved):

src/Toolkit/Str.php
lukasbestle: @distantnative Should we really officially support this though by adding the type hint for `false`? If `false` is passed to an argument that is hinted for `string`, the `false` value is converted to an empty string. This can be made explicit in the calling code (here: `Str::slug()`) by passing an empty string instead of `false` in the first place.

I feel like supporting `false` is unnecessary because it doesn't add any benefit over an empty string, but makes the method signature and code more complex.

lukasbestle: @distantnative Ping

distantnative: Idk, I'm indifferent, don't have any issue with supporting `false` either

PR #5511, opened 2023-08-27
Fix permissions for almighty `kirby`

### LINKED ISSUE #5365: [v4] Create listed pages bypassing permissions

distantnative:
## This PR …

Currently, if a blueprint option disables a permission also for admins, event he almighty `kirby` user with fail at the permission because it is treated as just an admin.

This PR elevates its position as now the almighty `kirby` user will really be almighty and supersede any permissions settings etc.

This is needed, e.g. so that developers can disable changing the page status via the UI even for admins, but still allow the page create dialog to create a page directly as listed (which involves changing the status, which is done by impersonating `kirby`).

- [x] Merge first https://github.com/getkirby/kirby/pull/5514

### Fixes
-  #5365

### Breaking change
- When impersonating the almighty `kirby` user, any permission check will succeed even if permission has been disabled for regular admins

## Ready?

### For review team

### DIFF
--- a/src/Cms/Auth.php
+++ b/src/Cms/Auth.php
@@ -390,6 +390,13 @@ class Auth
 
 		// validate the user and log in to the session
 		$user = $this->validatePassword($email, $password);
+
+		// @codeCoverageIgnoreStart
+		if ($user->id() === 'kirby') {
+			throw new PermissionException('The almighty user "kirby" cannot be used for login, only for raising permissions in code via `$kirby->impersonate()`');
+		}
+		// @codeCoverageIgnoreEnd
+
 		$user->loginPasswordless($options);
 
 		// clear the status cache
--- a/src/Cms/Auth.php
+++ b/src/Cms/Auth.php
@@ -390,13 +390,6 @@ class Auth
 
 		// validate the user and log in to the session
 		$user = $this->validatePassword($email, $password);
-
-		// @codeCoverageIgnoreStart
-		if ($user->id() === 'kirby') {
-			throw new PermissionException('The almighty user "kirby" cannot be used for login, only for raising permissions in code via `$kirby->impersonate()`');
-		}
-		// @codeCoverageIgnoreEnd
-
 		$user->loginPasswordless($options);
 
 		// clear the status cache
--- a/src/Cms/ModelPermissions.php
+++ b/src/Cms/ModelPermissions.php
@@ -45,8 +45,16 @@ abstract class ModelPermissions
 
 	public function can(string $action): bool
 	{
+		$user = $this->user->id();
 		$role = $this->user->role()->id();
 
+		// the almighty `kirby` user can do anything
+		if ($user === 'kirby' && $role === 'admin') {
+			return true;
+		}
+
+		// users with the `nobody` role can do nothing
+		// that needs a permission check
 		if ($role === 'nobody') {
 			return false;
 		}
--- a/src/Cms/ModelPermissions.php
+++ b/src/Cms/ModelPermissions.php
@@ -48,23 +48,25 @@ abstract class ModelPermissions
 		$user = $this->user->id();
 		$role = $this->user->role()->id();
 
-		// the almighty `kirby` user can do anything
-		if ($user === 'kirby' && $role === 'admin') {
-			return true;
-		}
-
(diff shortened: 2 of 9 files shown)


distantnative:
@lukasbestle I feel like this change makes sense, if we call it "almighty" - but it is a big breaking change, also seen in our unit tests where we expected even with `$kirby->impersonate('kirby')` many permission checks to fail. Which I think is a bit weird, given the almighty title, but has been the states quo.

Or do you have another idea how https://github.com/getkirby/kirby/blob/v4/develop/src/Panel/PageCreateDialog.php#L251-L255 could not fail at the permissions checks from https://github.com/getkirby/kirby/blob/v4/develop/src/Cms/ModelPermissions.php#L46-L83 where if a blueprint option disables changing status also for admins the impersonated `kirby` user will fail as well?

lukasbestle:
I can't think of an alternative to an "almighty" user. I think it's the correct approach.

However I feel like we should try to avoid the breaking change. It could be critical if some part of Kirby or site/plugin code has relied on the fact that the `kirby` user is an admin. So it could introduce vulnerabilities.

Maybe we could use the opportunity to introduce new "internal" users with a more robust foundation, i.e. all based on special user IDs that are blocked by `UserRules::validId()`, so they are guaranteed to never occur in actual users. So something like:

- `_nobody`
- `_admin`
- `_almighty`

I think that would be clearer than the `kirby` name we currently use. With the underscore, those names cannot have occurred naturally before (because the auto-generated user IDs are alphanumerical).

distantnative:
As discussed separately, I think the underscores are a bit odd in the context of Kirby.

I think there are two ways forward:
- making `kirby` really almighty (this PR), risk if existing setups rely of it being limited by admin permissions
- creating another new pseudo-user/role that is truly almighty

lukasbestle:
As we discussed, it is pretty unlikely that anyone who used the `kirby` user actually intended it to be limited by admin permissions. It was always meant as an almighty user and to be used with the intention "perform the following actions without checking for permissions".

Option 2 would be extra careful, but it comes with major complexity caveats that may harm security more than help. So I now think that this wouldn't be worth it.

I think with the documented breaking change, we are fine.

bastianallgeier:
I agree with Lukas' thoughts. It's also the perfect timing to make this changes with v4.

src/Cms/Auth.php
```diff
@@ -390,6 +390,13 @@ public function login(
 
 		// validate the user and log in to the session
 		$user = $this->validatePassword($email, $password);
+
+		// @codeCoverageIgnoreStart
+		if ($user->id() === 'kirby') {
+			throw new PermissionException('The almighty user "kirby" cannot be used for login, only for raising permissions in code via `$kirby->impersonate()`');
+		}
+		// @codeCoverageIgnoreEnd
```
lukasbestle: What do you think about moving this check to the `$user->loginPasswordless()` method? There it would fully prevent any kind of login and also we would be able to test this in CI.

lukasbestle:
I've added unit and integration tests for `$auth->login()` as well to reduce the risk of future regressions in this important area of system security.

lukasbestle (approved):

PR #5515, opened 2023-08-27
Additional PHP unit tests

distantnative:


src/Toolkit/View.php
```diff
@@ -52,26 +51,17 @@ public function file(): string
 		return $this->file;
 	}
 
-	/**
-	 * Creates an error message for the missing view exception
-	 */
-	protected function missingViewMessage(): string
```
bastianallgeier: I would not remove this here. I vaguely remember that there was a reason for this. I think it was about making extending the class easier.

distantnative: Ok

bastianallgeier (approved):
LGTM

PR #5520, opened 2023-08-29
Correct styling for a disabled writer with toolbar

bastianallgeier:
Screenshot 2023-08-29 at 11 48 20

### DIFF
--- a/panel/src/components/Forms/Writer/Writer.vue
+++ b/panel/src/components/Forms/Writer/Writer.vue
@@ -10,7 +10,7 @@
 		class="k-writer"
 	>
 		<k-writer-toolbar
-			v-if="editor"
+			v-if="editor && !disabled"
 			ref="toolbar"
 			v-bind="toolbar"
 			:editor="editor"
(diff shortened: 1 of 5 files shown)


distantnative:
I would actually rather style it the same way as the toolbar in the textarea field - which just isn't displayed when the field is disabled.

bastianallgeier:
It's a bit more complicated than in a textarea, but I will give it a try.

bastianallgeier:
Ok, I think I got a good solution and also fixed a few more display glitches between textarea and writer. You can check out the latest lab version for more comparison test setups between the two fields.

panel/src/components/Forms/Writer/Writer.vue
```diff
@@ -2,17 +2,19 @@
...
 		:spellcheck="spellcheck"
 		class="k-writer"
 	>
 		<k-writer-toolbar
-			v-if="editor"
+			v-if="editor && !disabled"
 			ref="toolbar"
 			v-bind="toolbar"
 			:editor="editor"
+			:disabled="disabled"
```
distantnative: This seems contradictory, if we don't render `k-writer-toolbar` when disabled, we don't need to pass it as prop.

distantnative (commented):
Works fine but had one little comment: if we don't load the writer toolbar when writer is disabled we don't need to pass the prop as well as add the prop and styling in the component.

bastianallgeier:
I fixed it and removed the redundant prop

PR #5528, opened 2023-08-30
Styling fix for the toolbar in the text block

### LINKED ISSUE #5502: [v4] New sticky Writer Toolbar in Block Text preview inconsistent with Writer Field
## Description
The new _sticky_ Writer Toolbar behaves differently when rendered within the WriterField component or within a Block Text preview component. 

**Expected behavior**  
The TextPreview Toolbar position, size and spacing to be more consistent with the Writer Field toolbar.
- More spacing between the toolbar and field content.
- The Toolbar to extent across the available horizontal space. Like it does in the Field. 

**Screenshots**  
image
WriterField Toolbar with good spacing to the fields text content and stretching across horizontally.

image

Toolbar rendered inside TextPreview component has not enough spacing to text content and does not fill the available horizontal space, which looks odd.

## To reproduce
Create a blocks field with the blueprint below and add a new Text Block with some content.
```yaml
type: blocks
fieldsets:
  text:
    name: field.blocks.text.name
    icon: text
    wysiwyg: true
    preview: text
    fields:
      text:
        type: writer
        label: Text
        headings:
          - 2
          - 3
        toolbar:
          inline: false 
```

## Your setup
**Kirby Version**  
4.0.0-alpha.7

bastianallgeier:
## Fixes

- Styling fix for the toolbar in the text block #5502 

Screenshot 2023-08-30 at 12 44 31

### DIFF
--- a/panel/src/components/Forms/Blocks/Types/Text.vue
+++ b/panel/src/components/Forms/Blocks/Types/Text.vue
@@ -84,7 +84,15 @@ export default {
 
 <style>
 .k-block-type-text-input {
-	line-height: 1.5em;
+	line-height: 1.5;
 	height: 100%;
+}
+.k-block-type-text-input.k-writer:not([data-toolbar-inline="true"]) {
+	gap: var(--spacing-2);
+}
+.k-block-type-text-input.k-writer:not([data-toolbar-inline="true"])
+	.k-writer-toolbar {
+	border: 1px solid var(--toolbar-border);
+	border-radius: var(--rounded);
 }
 </style>
--- a/panel/src/components/Forms/Blocks/Types/Text.vue
+++ b/panel/src/components/Forms/Blocks/Types/Text.vue
@@ -87,12 +87,14 @@ export default {
 	line-height: 1.5;
 	height: 100%;
 }
-.k-block-type-text-input.k-writer:not([data-toolbar-inline="true"]) {
-	gap: var(--spacing-2);
+.k-block-container.k-block-container-type-text {
+	padding: 0;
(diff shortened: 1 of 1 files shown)


bastianallgeier:
It's not the most beautiful solution in the world with the two toolbars, but there are not too many other alternatives to do this.

ovenum:
Would prefer the less floating toolbar as used in the Writer field, it feels more attached to field it belongs to. 
But also can’t see a better solution for the block toolbar overlapping the writer toolbar. So this looks like the best solution.

distantnative:
Why does it need to stretch full width if it doesn't need the space? Looks a little odd

bastianallgeier:
It really doesn't. That's just closer to what's happening in the textarea or the write field.

distantnative:
Ok only now realized that this if for `inline: false `. But honestly it feels broken like this. Is there no way to not have the padding wrap around?

bastianallgeier:
We could totally solve this with :has 🙈

distantnative:
@bastianallgeier Could we circumvent this with adding lines in this direction?


```css
.k-block-container.k-block-container-type-text {
	padding: 0;
}
.k-block-type-text-input.k-writer[data-toolbar-inline="true"] {
	padding: 0.75rem;
}
.k-block-type-text-input.k-writer:not([data-toolbar-inline="true"])
	> .ProseMirror {
	padding: 0.75rem;
}
```

bastianallgeier:
@distantnative uhh, that's actually pretty nice and simple. Just changed it.

PR #5537, opened 2023-08-31
More space for selector options

### LINKED ISSUE #5533: [v4] Multiselect field always offers search & create features / cuts off long options
## Description

1. [x] The multiselect field always shows a **search/filter**-field _(ignoring the blueprint-setting `search: false`)_
2. [x] The multiselect field always comes up with a misleading **create**-option whenever a search for a non-existing option is performed. _(This happens with both static and dynamic options, using the feature doesn't do anything at all.)_
3. [x] The multiselect field cuts of **options with long titles** due to its limited width.

**Expected behavior**  

1. The multiselect field should not contain a **create**-option by default. _(if at all, this would have to be extensively configurable)_
2. The multiselect field should not contain a **search/filter**-field if the blueprint is set to `search: false`.
3. The multiselect field should not **cut off options** as long as there is enough space left to show them properly.

**Screenshots**

kirby-multiselect-bug-v4

## To reproduce

1. Take the first example from the official **[multiselect-field-reference](https://getkirby.com/docs/reference/panel/fields/multiselect)** and try it out in the [latest alpha-release](https://github.com/getkirby/kirby/releases/tag/4.0.0-alpha.7).
2. Add/Remove the `search: false`-setting to the blueprint.
3. Try to use the **create**-option.
4. Try out some options with really long names

## Your setup

**Kirby Version:**   [4.0.0-alpha.7](https://github.com/getkirby/kirby/releases/tag/4.0.0-alpha.7).

bastianallgeier:
## Fixes 

- Offer more space for options in the selector dropdown https://github.com/getkirby/kirby/issues/5533

### DIFF
--- a/panel/src/components/Dropdowns/SelectorDropdown.vue
+++ b/panel/src/components/Dropdowns/SelectorDropdown.vue
@@ -59,7 +59,8 @@ export default {
 <style>
 .k-selector-dropdown-content {
 	--color-text-dimmed: var(--color-gray-400);
-	width: 15rem;
+	min-width: 15rem;
+	max-width: 30rem;
 	padding: 0;
 }
 .k-selector-dropdown .k-selector-header {


distantnative (approved):
Guess this is a bit better, still realizing that I'm not so happy with multi select being almost the same like tags (not full width, not staying open to select multiple, removing selected options from the dropdown instead of showing them as selected in the dropdown).

bastianallgeier:
I agree. It's more like an improvement for the current state to make it more usable. It's not perfect yet.

PR #5538, opened 2023-08-31
Only apply block delete shortcut when no input is focused

### LINKED ISSUE #5534: [v4] Shortcut for deleting a block overwrites Windows shortcut for deleting a word
## Description
CTRL+BACKSPACE is an OS-wide shortcut for deleting the last word in a text field. It works across all Windows-application.

**Expected behavior**  
When I edit text in a Kirby text block, I expect the exact same behavior described above.

**Actual Behavior**
Kirby will ask me, if I really would like to delete this text block. 

## To reproduce

1. Add a text block to any page
2. Start typing text in this text block
3. Press CTRL+BACKSSPACE on a Windows machine

## Your setup
**Kirby Version**  
Kirby 4 Alpha 7

 - OS: Windows 11 Pro

bastianallgeier:
## Fixes

- The keyboard shortcut to delete blocks is no longer preventing the default keyboard shortcut in text inputs and writer fields. https://github.com/getkirby/kirby/issues/5534

### DIFF
--- a/panel/src/components/Forms/Blocks/Block.vue
+++ b/panel/src/components/Forms/Blocks/Block.vue
@@ -19,7 +19,7 @@
 		@keydown.ctrl.alt.up.prevent.stop="$emit('selectUp')"
 		@keydown.ctrl.shift.down.prevent.stop="$emit('sortDown')"
 		@keydown.ctrl.shift.up.prevent.stop="$emit('sortUp')"
-		@keydown.ctrl.backspace.prevent.stop="remove"
+		@keydown.ctrl.backspace.stop="backspace"
 		@focus.stop="$emit('focus')"
 		@focusin.stop="onFocusIn"
 	>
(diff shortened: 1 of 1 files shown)


panel/src/components/Forms/Blocks/Block.vue
```diff
@@ -202,6 +202,17 @@ export default {
...
 	},
 	methods: {
+		backspace(e) {
+			// ignore the shortcut when an input is focused
+			if (e.target.matches("[contenteditable], input, textarea")) {
+				return false;
+			}
+
+			e.preventDefault();
+			e.stopPropagation();
```
distantnative: Isn't this redundant since the line above still has `.stop`?

bastianallgeier: You are right. It's redundant.

PR #5544, opened 2023-08-31
Page move dialog: better Site handling

distantnative:
Also gets us a disabled site if the page cannot be moved there


bastianallgeier:
That's nice. I think I would still return an array for the root level with just one item for the site and not wrap it afterwards in JS. It feels more correct from a data structure perspective, since the tree always expects an array of items on any level.

distantnative:
Done :)

bastianallgeier (approved):

PR #5545, opened 2023-08-31
Filter icons.svg comment from output

distantnative:


src/Panel/Assets.php
```diff
@@ -151,7 +151,9 @@ public function icons(): string
 	{
 		$dir  = $this->kirby->root('panel') . '/';
 		$dir .= $this->dev ? 'public' : 'dist';
-		return F::read($dir . '/img/icons.svg');
+		$icons = F::read($dir . '/img/icons.svg');
+		$icons = preg_replace('/<!--(.|\s)*?-->/', '', $icons);
```
lukasbestle, line 155: Would this work as well (better performance)?

```suggestion
		$icons = Str::after($icons, '-->');
```

distantnative: Isn't this making too many assumptions? What if we ever add something before the comment in the SVG file but don't remember this code part?

lukasbestle: TBH I think this is rather theoretical. The purpose of this PR is to remove the license comment and I don't see a reason why it needs to be anywhere else than at the beginning of the file.

If we do ever change that for whatever reason in the future, we can change this line as well. We will see when it breaks and can track down that this line is responsible.

bastianallgeier: If preg_replace does not have any noticeable performance hit with the large icon file, I would use the more robust solution.

bastianallgeier:
Let's merge it like this. It's easy enough to optimize later if needed.

Issue #5489, opened 2023-08-17
[v4] Link Field: Pages are not visible when UUID's are turned off


modufolio:
## Description
Pages are not visible when UUID's are turned off

Only one is displayed

## To reproduce
Add this to the config file:
`    'content' => [
        'uuid' => false
    ],
`

One will be displayed,
when this is removed, the error page is shown too.

## Your setup
**Kirby Version**  
4.0.0-alpha.7

afbora:
Seems link field only works with UUID.

And when clicked to first one:

> Cannot read properties of null (reading 'trim')

afbora:
Important question is; now link field detects based on `file://` and `page://` strings. How can we detect pages and files when UUID disabled?

lukasbestle:
We could use URIs with the page ID like `page://projects/project-a`. But then we need to make sure that those URIs also work when UUID support is enabled (otherwise sites will break when the UUID option is changed with existing content). So not simple.

distantnative:
I don't think the link field can meaningfully support the page or file option without UUIDs. IMO when UUIDs are disabled, the link field should not allow these options.

Or it allows them for picking, but then absolute URLs get inserted and on editing, the url option is used.

lukasbestle:
> Or it allows them for picking, but then absolute URLs get inserted and on editing, the url option is used.

That sounds like a manageable solution to me. The UX really isn't great, but everything else is a huge technical challenge just for those who need to disable UUIDs.

afbora:
I think @distantnative solution makes sense 👍

bastianallgeier:
I think that this will become a lot easier after this PR https://github.com/getkirby/kirby/pull/5531

distantnative:
@bastianallgeier do you think with this PR there is a different solution than

> I don't think the link field can meaningfully support the page or file option without UUIDs. IMO when UUIDs are disabled, the link field should not allow these options.
> 
> Or it allows them for picking, but then absolute URLs get inserted and on editing, the url option is used.

bastianallgeier:
@distantnative no, I just think it will be easier to implement now.

PR #5547, opened 2023-09-01
Fixed link field when UUIDs are switched off

bastianallgeier:
## Fixes 

- The link field is now usable when UUIDs are switched off. The page and file browsers will insert the absolute URL instead of the UUID https://github.com/getkirby/kirby/issues/5489

### DIFF
--- a/panel/src/components/Forms/Field/LinkField.vue
+++ b/panel/src/components/Forms/Field/LinkField.vue
@@ -77,7 +77,7 @@
 					<k-page-tree
 						:current="getPageUUID(value)"
 						:root="false"
-						@select="selectPage($event)"
+						@select="selectModel($event)"
 					/>
 				</div>
 			</div>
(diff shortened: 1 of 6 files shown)


distantnative:
aria-select doesn't work in PageTree anymore

panel/src/components/Forms/Field/LinkField.vue
```diff
@@ -345,6 +349,24 @@ export default {
...
+		},
+		selectFile(file) {
+			if (file.uuid) {
+				this.onInput(file.uuid);
+				return;
+			}
+
+			this.switchType("url");
+			this.onInput(file.url);
+		},
```
afbora: This code block can be simplify like that

```suggestion
		selectModel(model) {
			if (model.uuid) {
				this.onInput(model.uuid);
				return;
			}

			this.switchType("url");
			this.onInput(model.url);
		},
```

afbora:
Minor issue on linking to `site`:

https://github.com/getkirby/kirby/assets/3393422/9efc0c30-d290-4cb7-8741-5e8cb55afab0

bastianallgeier:
@getkirby/frontend 

- fixed aria-selected in the tree
- fixed site link

afbora (approved):
Looks great to me 🎉

PR #5553, opened 2023-09-01
Input CSS refactoring

bastianallgeier:
## Enhancements

- Simplified and cleaned up input CSS 
- Better configuration options for inputs through CSS variables
- Increased font size to 16px on mobile

### DIFF
--- a/panel/src/components/Dialogs/SearchDialog.vue
+++ b/panel/src/components/Dialogs/SearchDialog.vue
@@ -190,7 +190,6 @@ export default {
 }
 
 .k-search-dialog-input {
-	--input-height: var(--height-lg);
 	--button-height: var(--input-height);
 	display: flex;
 	align-items: center;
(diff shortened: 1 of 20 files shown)


distantnative:
@bastianallgeier should we remove `data-theme` then completely from `k-input`?

bastianallgeier:
It's already a pretty aggressive refactoring, but I think it's urgently needed to get to a cleaner state. Especially for the font size changes on smaller screens, but also in general for more flexible styling. 

I'm not sure if we still need the theme somehow. I'd say no.

distantnative:
I just saw that you removed the CSS rules for it, which is fine to me I think (hope we are not using an input somewhere that doesn't need that styling).

But so I thought, the data attribute probably has no use anymore either.

bastianallgeier:
The concept now is that the input variables offer a far more fine grained control. Even if there's an input that is not used within the field context, it's now a lot easier to style it. The theme context was always confusing. Where do we draw the line? Which rules are really just for the field and which are needed for any use of the input. I think this is now a lot better. I'd say we remove it the more I think about it.

distantnative:
I'm fine with that 👍 will you do the change?

distantnative (approved):

PR #5554, opened 2023-09-01
CI: Only enable Whoops when it is tested

lukasbestle:
This first step already helps us to make CI succeed on PHP 8.0. I'll still work on the other quick fix as we discussed, but that may take a little longer.

### Refactoring

- Whoops is now generally disabled during PHPUnit test runs to reduce memory usage during tests

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -5,6 +5,7 @@
 	xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd"
 
 	bootstrap="tests/bootstrap.php"
+	convertDeprecationsToExceptions="true"
 	colors="true"
 	verbose="true"
 	stderr="true"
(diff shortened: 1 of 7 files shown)


distantnative (approved):
Works for me locally and I implementation looks quite clean.

Didn't quite understand the Assert/Deprecated PHPUnit stuff, but guess that's intended/fine.

lukasbestle:
The `Error\Deprecated` and `Error\Warning` classes are exceptions thrown by PHPUnit when a PHP (deprecation) warning was triggered. So by catching those, we can test for the warnings.

I first used `$this->expectException(Deprecated::class)`, however that lead to a PHPUnit deprecation warning by itself because PHPUnit 10 will no longer convert warnings to exceptions. That's why the `try/catch` is necessary for now. We'll need yet another solution once we switch to PHPUnit 10: https://github.com/sebastianbergmann/phpunit/issues/5062

`Assert::fail()` throws an exception that makes the test fail. That's needed to mark the test as failed if the `catch` didn't run as expected.

PR #5555, opened 2023-09-01
Upload: better async upload queue

distantnative:
Hi @afbora, would love to get your help on this, testing all kinds of upload scenarios and if they are still working. Thanks!

### DIFF
--- a/panel/src/panel/upload.js
+++ b/panel/src/panel/upload.js
@@ -1,6 +1,7 @@
 import { uuid } from "@/helpers/string";
 import State from "./state.js";
 import listeners from "./listeners.js";
+import queue from "@/helpers/queue.js";
 import upload from "@/helpers/upload.js";
 import { extension, name, niceSize } from "@/helpers/file.js";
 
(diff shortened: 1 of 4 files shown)


afbora:
Uploading avatar throws following error:

> Uncaught TypeError: this.start is not a function at HTMLInputElement.<anonymous> (index.min.js:1:379743)

https://github.com/getkirby/kirby/blob/v4/enhancement/upload-queue/panel/src/panel/upload.js#L179

afbora:
~I don't know following issue is related with this PR: Creating thumbnail issue with replace while in lock state.~

Edit: Moved to https://github.com/getkirby/kirby/issues/5566

panel/src/panel/upload.js
```diff
@@ -164,7 +176,7 @@ export default (panel) => {
 					// if upload should start immediately
 					this.set(options);
 					this.select(event.target.files);
-					this.start();
+					this.submit();
```
afbora, line 179:
```js
await this.submit();
```

It doesn't have to be like this, does it?

distantnative: I don't think so, because we want `this.input.remove()` to happen directly even while the upload is running.

distantnative:
@afbora Not sure the thumbnail issue is related. Could be a general v4 thing or even broken in v3?

afbora:
@distantnative You're right. That issue exists already 3.9.6. I moved to https://github.com/getkirby/kirby/issues/5566

afbora (approved):
LGTM. All works great for me 👍

PR #5557, opened 2023-09-02
Plugin `assets` extension

### LINKED ISSUE #5247: Allow plugins to explicitly define their assets

distantnative:
### Features
- New `assets` extension that allows plugins to specify assets from custom paths and with a wider range of extensions than previously supported. https://github.com/getkirby/kirby/issues/5247

### Breaking changes
- Files in a plugin's `assets` directory are now always assumed to be public, independent of their file extension. If your plugin needs to store other files in the assets directory, please use the new `assets` extension to explicitly define the public assets.

### DIFF
--- a/src/Cms/Plugin.php
+++ b/src/Cms/Plugin.php
@@ -2,11 +2,13 @@
 
 namespace Kirby\Cms;
 
+use Closure;
 use Composer\InstalledVersions;
 use Exception;
 use Kirby\Cms\System\UpdateStatus;
 use Kirby\Data\Data;
 use Kirby\Exception\InvalidArgumentException;
+use Kirby\Filesystem\Dir;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\Str;
 use Kirby\Toolkit\V;
@@ -57,6 +59,24 @@ class Plugin
 	public function __call(string $key, array $arguments = null): mixed
 	{
 		return $this->info()[$key] ?? null;
+	}
+
+	public function assets(): array
+	{
+		if ($assets = $this->extends['assets'] ?? null) {
+			if ($assets instanceof Closure) {
+				$assets = $assets();
+			}
+
+			return $assets;
+		}
+
+		$assets = $this->root() . '/assets';
+
+		return A::map(
+			Dir::index($this->root() . '/assets', true),
+			fn ($filename) => 'assets/' . $filename
+		);
 	}
 
 	/**
--- a/src/Cms/Plugin.php
+++ b/src/Cms/Plugin.php
@@ -63,10 +63,7 @@ class Plugin
 	}
 
 	/**
-	 * Returns a specific
-	 *
-	 * @param string $name
-	 * @return string|null
+	 * Returns the absolute path to a specific asset
 	 */
 	public function asset(string $path): string|null
 	{
@@ -92,13 +89,11 @@ class Plugin
 			// normalize array: use relative path as
 			// key when no key is defined
 			foreach ($assets as $key => $asset) {
-				unset($assets[$key]);
-
 				if (is_int($key) === true) {
+					unset($assets[$key]);
 					$key = Str::after($asset, $this->root() . '/');
+					$assets[$key] = $asset;
 				}
-
-				$assets[$key] = $asset;
 			}
 		}
 
@@ -109,7 +104,10 @@ class Plugin
 			$root   = $this->root() . '/assets';
 
 			foreach (Dir::index($root, true) as $asset) {
-				$assets['assets/' . $asset] = $root . '/' . $asset;
+				$path = $root . '/' . $asset;
+				if (is_file($path) === true) {
+					$assets['assets/' . $asset] = $path;
+				}
 			}
 		}
 
--- a/src/Cms/Plugin.php
+++ b/src/Cms/Plugin.php
@@ -62,23 +62,48 @@ class Plugin
 		return $this->info()[$key] ?? null;
 	}
 
-	public function asset(string $name): string|null
+	/**
+	 * Returns a specific
+	 *
+	 * @param string $name
+	 * @return string|null
+	 */
+	public function asset(string $path): string|null
 	{
-		return $this->assets()[$name] ?? null;
+		return $this->assets()[$path] ?? null;
 	}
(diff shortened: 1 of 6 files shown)


distantnative:
@lukasbestle is it problematic that now a wider range of files can be accessed from the plugin's `assets` folder if the extension is not defined (so from the automatically generated list)?

lukasbestle:
Probably not in practice. An edge case could be that an existing plugin contains a PHP file in its assets folder that is not meant to be accessed directly. Once this file is in the `media` folder, the web server will execute it on the next request.

I haven't looked at your code yet, but as far as I remember, we planned to deprecate the non-explicit access, right? In this case I think it would be good to keep the current behavior regarding file types and not widen the possible list for the deprecated way.

distantnative:
Not sure I would deprecate the non-explicit/automatic way. Seems like forcing extra work on plugin developers that in many cases might not be beneficial. How about simply restricting the automatic list to the same extensions as currently restricted by the route?

lukasbestle:
Hm, not sure. If we were to deprecate the automatic way, I'd agree with you to keep the list restricted.

But you have a point that keeping the automatic way might make sense. And in that case, limiting the extensions is weird. Let's say you have ten assets (4 CSS files, 5 JS files and one file of a type that we don't support in our allowlist). Just because of that one weird file you would now have to explicitly define all ten assets to circumvent the extension allowlist.

Do you think it would be an option to mark this as a breaking change or would this be too weird/ineffective:

> Files in the `assets` directory of a plugin are now always assumed to be public, independent of their file extension. If your plugin needs to store other files in the `assets` directory, please use the new `assets` extension to explicitly define the public assets.

distantnative:
@lukasbestle I am fine with it being a breaking change. Have added it to the initial post. I think this is then ready for review.

src/Cms/Plugin.php
```diff
@@ -59,6 +62,60 @@ public function __call(string $key, array $arguments = null): mixed
...
+			// normalize array: use relative path as
+			// key when no key is defined
+			foreach ($assets as $key => $asset) {
+				unset($assets[$key]);
+
+				if (is_int($key) === true) {
+					$key = Str::after($asset, $this->root() . '/');
+				}
+
+				$assets[$key] = $asset;
```
lukasbestle: To increase performance:

```suggestion
				if (is_int($key) === true) {
					unset($assets[$key]);
					$key = Str::after($asset, $this->root() . '/');
					$assets[$key] = $asset;
				}
```

src/Cms/Plugin.php
```diff
@@ -59,6 +62,60 @@ public function __call(string $key, array $arguments = null): mixed
...
+		}
+
+		// fallback: if no assets are defined in the plugin extension,
+		// use all files in the plugin's `assets` directory
+		if ($assets === null) {
+			$assets = [];
+			$root   = $this->root() . '/assets';
+
+			foreach (Dir::index($root, true) as $asset) {
+				$assets['assets/' . $asset] = $root . '/' . $asset;
```
lukasbestle: I think `Dir::index()` also returns directories, but directories cannot be served as assets:

```suggestion
				$path = $root . '/' . $asset;
				if (is_file($path) === true) {
					$assets['assets/' . $asset] = $path;
				}
```

src/Cms/PluginAssets.php
```diff
@@ -25,7 +25,7 @@ class PluginAssets
 	public static function clean(string $pluginName): void
```
lukasbestle, line 25: With the `assets` extension, there can be assets where the absolute path does not match the relative path (e.g. `'some-dependency.css' => __DIR__ . '/vendor/some/dependency/styles.css'`). I think we need to tap into `$plugin->assets()` to look up the whether an asset exists.

src/Cms/PluginAssets.php
```diff
@@ -51,22 +51,25 @@ public static function clean(string $pluginName): void
...
-		string $filename
+		string $path
 	): Response|null {
 		if ($plugin = App::instance()->plugin($pluginName)) {
-			$source = $plugin->root() . '/assets/' . $filename;
+			$assets = $plugin->assets();
 
-			if (F::exists($source, $plugin->root()) === true) {
+			if (
+				($asset = $assets[$path] ?? null) &&
```
lukasbestle: Could we use `$asset = $plugin->asset($path)` instead? Then we don't need the `$assets` variable declaration above.

src/Cms/Plugin.php
```diff
@@ -59,6 +62,60 @@ public function __call(string $key, array $arguments = null): mixed
 		return $this->info()[$key] ?? null;
 	}
 
+	/**
+	 * Returns a specific
+	 *
+	 * @param string $name
+	 * @return string|null
```
lukasbestle:
```suggestion
	 * Returns the absolute path to a specific asset
```

src/Cms/PluginAssets.php
distantnative: @lukasbestle I pushed a suggestion

lukasbestle: I like it. Very clean!

lukasbestle (approved):

PR #5562, opened 2023-09-02
Items: expose options slot for table layout

### LINKED ISSUE #5561: `k-collection`: table layout doesn't expose options slot

distantnative:
### Fixes
- `k-collection` and `k-items`: `options` slot gets properly exposed also for table layout

### DIFF
--- a/panel/src/components/Collection/Items.vue
+++ b/panel/src/components/Collection/Items.vue
@@ -6,7 +6,11 @@
 		@change="$emit('change', $event)"
 		@sort="$emit('sort', $event)"
 		@option="onOption"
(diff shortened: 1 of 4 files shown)


panel/src/components/Collection/Items.vue
```diff
@@ -38,7 +42,7 @@
 					@option="onOption($event, item, itemIndex)"
```
bastianallgeier, line 42:
```suggestion
					@option="onOption($event, item, itemIndex: index)"
```

bastianallgeier (approved):

PR #5567, opened 2023-09-03
Fix `array.sortBy`

distantnative:
I am a bit confused cause this helper has been around for ages. But it didn't work for me. The imported `sort` function returns a function first itself. So how could we directly use it so far?

The structure field works with this - did that ever really work?

### DIFF
--- a/config/fields/structure.php
+++ b/config/fields/structure.php
@@ -183,7 +183,12 @@ return [
 		$data = [];
 
 		foreach ($value as $row) {
-			$data[] = $this->form($row)->content();
+			$row = $this->form($row)->content();
+
+			// remove frontend helper id
+			unset($row['_id']);
+
+			$data[] = $row;
 		}
 
 		return $data;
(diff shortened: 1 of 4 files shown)


bastianallgeier:
Wow, this PR started quite a few things :)

You are right and sorting did in fact not really work

Once I checked your PR, I realized that your fix lead to content loss in the structure field when the sortBy option is used. The drawers are opened by array index, but when sorting is applied during editing, the array index might change. This will completely break entries, because a totally different entry is suddenly edited. It was a mess. 

To fix this, I started adding a private _id to every row in the structure field. This also improves the rest of the code quite a bit, but it needed a lot of refactoring. The great side-effect of this is that we no longer have those hacks to work around changing indexes in paginated structure fields. This makes the field a lot more stable in general and we could also use this a bit later to finally get UUIDs working for structure items. 

As another quick fix, I also changed the outline for sortable rows. 

I feel like it's all pretty clean now and the tests in the lab work fine, but we should definitely have a closer look at all of this together. 

I also added more sorting test setups to the lab.

distantnative:
Looks good to me, but if @afbora could give it some more testing I'd feel a bit safer.

afbora:
@distantnative Which parts should I test, just the structure field?

distantnative:
@afbora yes, structure field with `sortBy` - does it sort correctly, is it correctly sorted after editing the value to be sorted etc.

afbora (approved):

PR #5569, opened 2023-09-05
Unify k-text and k-writer

bastianallgeier:
When I went through a couple more display issues, I realized that we are basically use redundant styles for the writer and the text component. This leaves a lot of room for issues and already created some. I've moved all the text styles into the k-text component and applied k-text to the ProseMirror container. This way we can get rid of all those redundant styles. I fixed a few glitches along the way. 

I've added a lot of new test setups for the text component to the sandbox to try all this.


distantnative:
@bastianallgeier It looks like you have now link and list both in k-text but also individual files

distantnative:
And what happened to kbd?

bastianallgeier:
My bad. I didn't clean up those files after moving it back Text.vue. 

I'm still not 100% decided on the best way to organize the styles. Should they all live in separate files or in a single file? 

I moved the kbd styles to reset.css because they apply everywhere. I think that's in general a good rule of thumb.

PR #5578, opened 2023-09-07
Clean up `k-pagination`

distantnative:
### Enhancement
- `k-pagination` always can be navigated by keys (no extra prop needed anymore)

### Refactored
- `k-pagination`: removed unused `align` and `dropdown` props

### Breaking changes
- `k-pagination` doesn't support setting custom labels/titles via `nextLabel`, `prevLabel` or `pageLabel`

### DIFF
--- a/panel/src/components/Navigation/Pagination.vue
+++ b/panel/src/components/Navigation/Pagination.vue
@@ -1,23 +1,33 @@
 <template>
-	<k-button-group
-		v-if="isVisible"
-		:data-align="align"
-		layout="collapsed"
-		class="k-pagination"
-	>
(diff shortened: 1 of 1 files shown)


bastianallgeier:
I don't think we can handle keyboard navigation like that. If we make it global and there are multiple paginated sections in a view, they will all be paginated at the same time. I guess we could do it as soon as one of the buttons is focused.

distantnative:
@bastianallgeier I really just removed the need to set the prop, didn't realize that this was broken before already

distantnative:
@bastianallgeier added an alternative

bastianallgeier:
Oh, that's really nice!

bastianallgeier (approved):

PR #5581, opened 2023-09-08
Add CLI specific config file 

lukaskleinschmidt:
## This PR …
Adds the option to define a CLI specific config file `config.cli.php`. You can already do something like this by using the `ready` callback and define your CLI specific options in there. Probably only for edge cases but when you need CLI specific options I think this would fit in quite nicely with the current system of environment (host name or ip address) specific options.

```php
// config.php
return [
    'ready' => function (App $kirby) {
        $options = [
            //
        ];

        if ($kirby->environment()->cli()) {
            $options = array_merge($options, [
                'option.one' => 1,
                'option.two' => 2,
            ]);
        }

        return $options;
    }
];

// config.cli.php
return [
    'option.one' => 1,
    'option.two' => 2,
];
```

Related: https://discord.com/channels/525634039965679616/525641819854471168/1149657559746224188

### Breaking changes
Hopefully none.

## Ready?

### For review team

### DIFF
--- a/src/Http/Environment.php
+++ b/src/Http/Environment.php
@@ -772,7 +772,7 @@ class Environment
 
 	/**
 	 * Loads and returns options from environment-specific
-	 * PHP files (by cli, host name and server IP address)
+	 * PHP files (by host name and server IP address or CLI)
 	 *
 	 * @param string $root Root directory to load configs from
 	 */
--- a/src/Http/Environment.php
+++ b/src/Http/Environment.php
@@ -772,12 +772,21 @@ class Environment
 
 	/**
 	 * Loads and returns options from environment-specific
-	 * PHP files (by host name and server IP address)
+	 * PHP files (by cli, host name and server IP address)
 	 *
 	 * @param string $root Root directory to load configs from
 	 */
 	public function options(string $root): array
 	{
+		// load the config for the cli
+		if ($this->cli() === true) {
+			return F::load(
+				file: $root . '/config.cli.php',
+				fallback: [],
+				allowOutput: false
+			);
+		}
+
 		$configHost = [];
 		$configAddr = [];
 
--- a/src/Http/Environment.php
+++ b/src/Http/Environment.php
@@ -778,20 +778,21 @@ class Environment
 	 */
 	public function options(string $root): array
 	{
-		// load the config for the cli
-		if ($this->cli() === true) {
-			return F::load(
-				file: $root . '/config.cli.php',
-				fallback: [],
-				allowOutput: false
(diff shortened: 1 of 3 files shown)


src/Http/Environment.php
```diff
@@ -772,18 +772,28 @@ public function isLocal(): bool
 
 	/**
 	 * Loads and returns options from environment-specific
-	 * PHP files (by host name and server IP address)
+	 * PHP files (by cli, host name and server IP address)
```
distantnative:
```suggestion
	 * PHP files (by host name and server IP address or CLI)
```

distantnative (commented):
Overall, I like the idea.

Any thoughts if this could cause problems elsewhere, @bastianallgeier @lukasbestle?

lukasbestle:
In theory it could clash with a host named `cli`, but that's very unlikely in practice and will only take effect if the host `cli` was explicitly allowlisted in the main `config.php`.

So I'd say it should be safe.

distantnative:
@lukaskleinschmidt Would you mind adding unit tests for this? Once with cli active and extra config loading, once with inactive and ignoring the cli config file? You can force the cli state for testing as done here https://github.com/getkirby/kirby/blob/develop/tests/Http/EnvironmentTest.php#L223-L231

lukaskleinschmidt:
I will try to get to it. But currently too busy

bastianallgeier:
I really like it as well!

lukaskleinschmidt:
I had the same issue locally regarding the failed test https://github.com/getkirby/kirby/actions/runs/6429036429/job/17457332224?pr=5581 but wanted to make sure this is not related to my setup.

Changing the test case for the [failing test](https://github.com/getkirby/kirby/pull/5581/commits/f2dc0cbd62500233dbfd6d6d6cf0662c7320d662#diff-47b3e9799786d5c74b7b868ceeba121be64d5dbec5da374c35873d06a987e5f8R1092) would fix the issue. But not sure if this is the proper way to do this here.
```php
public function testOptionsFromInvalidHost()
{
	$env = new Environment([
		'cli' => false, // <- this one here
		'allowed' => [
			'http://example.de'
		]
	], [
		'SERVER_NAME' => 'example.com'
	]);

	$this->assertSame([], $env->options($this->config));
}
```

lukasbestle:
@lukaskleinschmidt I think that would be the correct solution. The test assumes an HTTP request with the `SERVER_NAME` environment variable, so we need to fake CLI mode off for the test to make sense.

lukasbestle (approved):

PR #5582, opened 2023-09-08
Input fixes

bastianallgeier:
While working on a clean up of our inputs, I discovered a lot of smaller issues and added quick fixes to this PR. 

I hope the commit messages are clear enough for each fix.

### DIFF
--- a/panel/src/components/Forms/Input/CheckboxesInput.vue
+++ b/panel/src/components/Forms/Input/CheckboxesInput.vue
@@ -1,6 +1,5 @@
 <template>
 	<ul
-		v-if="options.length"
 		:style="'--columns:' + columns"
 		class="k-checkboxes-input k-grid"
 		data-variant="choices"
@@ -9,7 +8,6 @@
 			<k-choice-input v-bind="choice" @input="input(choice.value, $event)" />
 		</li>
 	</ul>
-	<k-empty v-else icon="info">{{ $t("options.none") }}</k-empty>
 </template>
 
 <script>
@@ -24,7 +22,10 @@ import {
 export const props = {
 	mixins: [autofocus, disabled, id, name, required],
 	props: {
-		columns: Number,
+		columns: {
+			default: 1,
+			type: Number
+		},
 		max: Number,
 		min: Number,
 		options: {
@@ -59,7 +60,7 @@ export default {
 					disabled: this.disabled,
 					info: option.info,
 					label: option.text,
-					name: this.name,
+					name: this.name ?? this.id,
 					theme: this.theme,
 					type: "checkbox",
 					value: option.value
--- a/panel/src/components/Forms/Input/ListInput.vue
+++ b/panel/src/components/Forms/Input/ListInput.vue
@@ -2,7 +2,7 @@
 	<k-writer
 		ref="input"
 		v-bind="$props"
-		:extensions="extensions"
+		:extensions="listExtensions"
 		:value="list"
 		class="k-list-input"
 		@input="onInput"
@@ -10,25 +10,17 @@
 </template>
 
 <script>
-import { disabled } from "@/mixins/props.js";
+import { props as WriterInputProps } from "@/components/Forms/Input/WriterInput.vue";
 import ListDoc from "@/components/Forms/Writer/Nodes/ListDoc.js";
 
 export const props = {
-	mixins: [disabled],
+	mixins: [WriterInputProps],
 	inheritAttrs: false,
 	props: {
-		autofocus: Boolean,
-		keys: Object,
 		nodes: {
 			type: Array,
 			default: () => ["bulletList", "orderedList"]
-		},
-		marks: {
-			type: [Array, Boolean],
-			default: true
-		},
-		spellcheck: Boolean,
-		value: String
+		}
 	}
 };
 
@@ -42,7 +34,7 @@ export default {
 		};
 	},
 	computed: {
-		extensions() {
+		listExtensions() {
 			return [
 				new ListDoc({
 					inline: true,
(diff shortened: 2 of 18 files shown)


panel/src/components/Forms/Input/ListInput.vue
```diff
@@ -42,7 +34,7 @@ export default {
 		};
 	},
 	computed: {
-		extensions() {
+		listExtensions() {
```
bastianallgeier, line 37: This will clash with the extensions prop from the writer.

panel/src/components/Forms/Input/CheckboxesInput.vue
```diff
@@ -9,7 +8,6 @@
 			<k-choice-input v-bind="choice" @input="input(choice.value, $event)" />
 		</li>
 	</ul>
-	<k-empty v-else icon="info">{{ $t("options.none") }}</k-empty>
```
bastianallgeier, line 12: The inputs shouldn't be responsible for this. We can handle that on a higher level in inputbox components.

panel/src/components/Forms/Input/RadioInput.vue
```diff
@@ -9,7 +8,6 @@
 			<k-choice-input v-bind="choice" @input="$emit('input', choice.value)" />
 		</li>
 	</ul>
-	<k-empty v-else icon="info" theme="info">{{ $t("options.none") }}</k-empty>
```
bastianallgeier, line 12: See my comment above in the checkboxes component

panel/src/components/Forms/Input/RadioInput.vue
```diff
@@ -41,7 +39,7 @@ export default {
 					disabled: this.disabled,
 					info: option.info,
 					label: option.text,
-					name: this.name,
+					name: this.name ?? this.id,
```
bastianallgeier, line 42: The radios and checkboxes won't work properly without a name. That's why it's better to fall back to the id.

panel/src/components/Forms/Input/SelectInput.vue
```diff
@@ -157,6 +163,10 @@ export default {
 	padding: var(--input-padding);
 	border-radius: var(--input-rounded);
 }
+.k-select-input[data-empty="true"] {
```
bastianallgeier, line 166: I'm using the same gray color here as the other text inputs use for their placeholder.

panel/src/components/Forms/Input/TextareaInput.vue
```diff
@@ -104,6 +104,9 @@ export default {
 	watch: {
 		value() {
 			this.onInvalid();
+			this.$nextTick(() => {
+				this.$refs.input.autosize();
```
bastianallgeier, line 108: I'm not sure if this also fixes the jumping issue, but it definitely will fix textareas that might receive updates from somewhere else through reactive changes.

panel/src/components/Forms/Input/UrlInput.vue
```diff
@@ -9,6 +9,10 @@ export const props = {
 			type: String,
 			default: "url"
 		},
+		placeholder: {
+			type: String,
+			default: () => window.panel.$t("url.placeholder")
```
bastianallgeier, line 14: We do this in the email and tel inputs and we should really do it here as well.

panel/src/components/Forms/Writer/Toolbar.vue
```diff
@@ -69,7 +69,18 @@ export default {
 			type: Boolean
 		},
 		marks: {
-			default: true,
+			default: () => [
```
bastianallgeier, line 72: Without a predefined toolbar layout, the buttons are all over the place when you just pass true or null.

panel/src/components/Forms/Writer/Writer.vue
```diff
@@ -65,7 +65,10 @@ export const props = {
 			})
 		},
 		extensions: Array,
-		headings: [Array, Boolean],
+		headings: {
+			default: () => [1, 2, 3, 4, 5, 6],
```
bastianallgeier, line 69: Without a default for the headings levels, the writer won't work at all if not used in a field.

panel/src/components/Forms/Input/CheckboxesInput.vue
distantnative: But should we do that change then only when the input box component is added?

panel/src/components/Forms/Input/ListInput.vue
```diff
@@ -2,33 +2,25 @@
...
 </template>
 
 <script>
-import { disabled } from "@/mixins/props.js";
+import { props as WriterInputProps } from "@/components/Forms/Input/WriterInput.vue";
 import ListDoc from "@/components/Forms/Writer/Nodes/ListDoc.js";
 
 export const props = {
-	mixins: [disabled],
+	mixins: [WriterInputProps],
```
distantnative, line 17: Naming is a bit odd, I think in other cases we would just name this `WriterInput`

panel/src/components/Forms/Input/CheckboxesInput.vue
bastianallgeier: This has been added somewhere in the alphas. The inputs never dealt with empty options in v3. But I also don't mind to keep it in there until the inputboxes are ready

distantnative (commented):
I'd change the name of the props imports, see comment, but other than that looks good

panel/src/components/Forms/Input/ListInput.vue
distantnative: E.g. https://github.com/getkirby/kirby/blob/v4/lab/inputboxes/panel/src/components/Dialogs/FormDialog.vue#L23 it's just `Fields` not `FieldsProps`

bastianallgeier: I agree that it should be consistent. But we are actually really just importing the props as mixin here. I really think that the "WriterInput" version is more confusing to me than the more verbose variant.

PR #5585, opened 2023-09-08
Deprecate `k-dropdown`

distantnative:
### Deprecated
- `<k-dropdown>` was deprecated. Use `<k-dropdown-content>` as standalone instead.

### DIFF
--- a/panel/src/components/Views/Files/FilePreview.vue
+++ b/panel/src/components/Views/Files/FilePreview.vue
@@ -162,7 +162,7 @@ export default {
 	max-width: 100cqw;
 	max-height: 100cqh;
 }
-.k-file-preview-thumb .k-dropdown {
+.k-file-preview-thumb > .k-button {
 	position: absolute;
 	top: var(--spacing-2);
 	inset-inline-start: var(--spacing-2);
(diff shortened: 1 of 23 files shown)


panel/src/components/Views/Files/FilePreview.vue
```diff
@@ -13,15 +13,13 @@
...
-				<k-dropdown>
-					<k-button
-						icon="dots"
-						size="xs"
-						style="color: var(--color-gray-500)"
-						@click="$refs.dropdown.toggle()"
-					/>
-					<k-dropdown-content ref="dropdown" :options="options" theme="light" />
-				</k-dropdown>
+				<k-button
```
bastianallgeier, line 15: This is breaking the layout in the files preview

panel/src/components/Views/Users/UserAvatar.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<k-dropdown>
+	<div>
```
bastianallgeier, line 2: Is the div here really needed?

distantnative: Since Vue SFC can only have a single element as root, yes.

panel/src/components/Views/Files/FilePreview.vue
distantnative: Fixed it

panel/src/components/Views/Users/UserAvatar.vue
bastianallgeier: 🙈 of course.

bastianallgeier (approved):

PR #5593, opened 2023-09-09
Update filename when changing format on upload 

### LINKED ISSUE #5530: Uploader creates images with wrong extension if format is converted
I'm using the new [Uploader](https://getkirby.com/releases/4.0/uploader#optimize-files-on-upload). When I tell it to convert images to another format, the resulting file extension is wrong.

This is my file blueprint:

```
accept:
  type: image

create:
  width: 500
  height: 500
  format: "webp"
```

Now if I upload an `image.jpg` it is converted to the **webp** format but the filename in the content folder stays `image.jpg`. 

I think it should be `image.webp` instead.

## Your setup
**Kirby Version**  
4.0.0-alpha.7

distantnative:
### Fixes
- #5530

## Ready?

### For review team

### DIFF
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -22,25 +22,34 @@ use Kirby\Uuid\Uuids;
 trait FileActions
 {
 	/**
-	 * Renames the file without touching the extension
+	 * Renames the file (optionally also the extension).
 	 * The store is used to actually execute this.
 	 *
 	 * @throws \Kirby\Exception\LogicException
 	 */
-	public function changeName(string $name, bool $sanitize = true): static
-	{
+	public function changeName(
+		string $name,
+		bool $sanitize = true,
+		string|null $extension = null
+	): static {
 		if ($sanitize === true) {
 			$name = F::safeName($name);
 		}
 
+		// if no extension is passed, make sure to maintain current one
+		$extension ??= $this->extension();
+
 		// don't rename if not necessary
-		if ($name === $this->name()) {
+		if (
+			$name === $this->name() &&
+			$extension === $this->extension()
+		) {
 			return $this;
 		}
 
-		return $this->commit('changeName', ['file' => $this, 'name' => $name], function ($oldFile, $name) {
+		return $this->commit('changeName', ['file' => $this, 'name' => $name, 'extension' => $extension], function ($oldFile, $name, $extension) {
 			$newFile = $oldFile->clone([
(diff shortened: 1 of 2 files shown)


marcojakob:
I gave this a spin - loving it!

I still found a small bug: **If a file with the same name already exists, the original and untransformed file ends up in the folder.**

Here is how to reproduce it:

1. Use the new files upload blueprint to transform an image to `webp` (or anything else)

```
create:
  format: "webp"
```

2. Upload a file with the name `test.png` --> file gets converted to `test.webp`
3. Then upload another file with the same name `test.png`

Now an error message is shown: 

image

The error message is correct. But the original file `test.png` ends up in the folder despite the error. Depending on the panel setup it might not even be deletable through the panel any more.

marcojakob:
Another related UI behavior that seems a little weird:

- A file is uploaded or transformed to a file `test.webp`
- Now if the user chooses to **Replace** the image, the `accept` setting of the file blueprint is ignored and only files with `webp` are selectable.

image

image

I'm on Windows just now, haven't tested on my Mac yet.

distantnative:
The latest post is expected. Files always can only be replaced by a file with the same extension. That also applies here.

marcojakob:
> The latest post is expected. Files always can only be replaced by a file with the same extension. That also applies here.

Thanks @distantnative for clarifying.

Still, I don't understand the reasoning behind that and it doesn't seem to make sense for the user to suddenly be restricted in the file extension. Maybe I don't see something or it has historical reasons.

Should I create a new issue where this could be discussed? I could also live with it since there is a workaround: We can set `replace: false` and force users to first delete the old and then add the new image.

bastianallgeier:
@marcojakob the main reason for this is the placement of files in templates. Before UUIDs this was only possible via filename and it is still the more readable and effortless option. But then there's a risk that a reference breaks when the file extension changes. But when the chose format matches with the file extension, I think we could totally accept it anyway because the integrity would still be given.

distantnative:
@marcojakob @bastianallgeier let's move that to a separate issue - looking at the beta stage we should try to wrap up PRs in their scope.

marcojakob:
@distantnative You mean both topics? Also https://github.com/getkirby/kirby/pull/5593#issuecomment-1714609575

or just the **replace** function?

distantnative:
Tried to implement it now in a way that the file rules already observe the converted/renamed file.

bastianallgeier (approved):
Looks good!

PR #5602, opened 2023-09-09
Support multiple Panel assets and absolute URLs

### LINKED ISSUE #5180: Allow URLs and multiple custom panel styles/scripts
## Description
While working on a solution for [this issue](https://github.com/lukaskleinschmidt/kirby-laravel-vite/issues/1) of my vite plugin I came across some shortcomings of the current custom panel styles/scripts implementation.

The shortcomings:
1. You can only have a path to a file
2. You can only define one file respectively

What would be nice to have:
1. URL support for custom panel assets
2. The possibility to add multiple assets

**Reasons**
In this case vite specific but most likely not limited to that.
To be able to make use of vites dev server or hot reloading you need to be able to define a full URL to work when developing. In addition to that you need to include the `@vite/client` to make hot reloading work. This requires you to include multiple script assets when using a custom `panel.js` as well.

Right now I use the `panel.route:after` hook to inject the tags with a [regex](https://github.com/lukaskleinschmidt/kirby-laravel-vite/commit/4b46c493f48332592862dc46a0e0440c10264d90#diff-7413d6453f901e939bbd840c8f0d1c7b20c2ca0e7f71741e4e07c6cf036f16c0R38) to make it work. But in general I think this might be useful outside of vite as well.

distantnative:
## This PR …

@lukaskleinschmidt Could you please test if this is what you had in mind with your issue?

### Enhancement
- `panel.css` and `panel.js` config options now also support arrays with multiple entries as well as absolute URLs

### Breaking changes
- `Pane\Assets::custom()` now returns an array

## Ready?

### For review team

### DIFF
--- a/src/Panel/Assets.php
+++ b/src/Panel/Assets.php
@@ -3,11 +3,13 @@
 namespace Kirby\Panel;
 
 use Kirby\Cms\App;
+use Kirby\Cms\Url;
 use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Filesystem\Asset;
 use Kirby\Filesystem\Dir;
 use Kirby\Filesystem\F;
+use Kirby\Toolkit\A;
 
 /**
  * The Assets class collects all js, css, icons and other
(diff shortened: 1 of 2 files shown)


lukaskleinschmidt:
Works like a charm

bastianallgeier (approved):

PR #5608, opened 2023-09-09
`$app->collection()` & template controllers: Lazy value resolver

### LINKED ISSUE #5607: [v4] Remove users from collection list for performance.
## Description

When using the collection helper, the user collection is loaded too.
```
	public function collection(string $name, array $options = [])
	{
		return $this->collections()->get($name, array_merge($options, [
			'kirby' => $this,
			'site'  => $site = $this->site(),
			'pages' => $site->children(),
			'users' => $this->users()
		]));
	}
```

$notes = collection('notes');

When you have 1000 users in the user collection, you want to avoid using this helper at the moment or use a custom app class.

## Fix

```
	public function collection(string $name, array $options = [])
	{
		return $this->collections()->get($name, array_merge($options, [
			'kirby' => $this,
			'site'  => $site = $this->site(),
			'pages' => $site->children(),
		]));
	}
```

## Your setup
**Kirby Version**  
4.0.0-beta.1

distantnative:
### Features

- New `LazyValue` class that can be used as controller argument to lazily resolve a value only when the controller function requests it

### Enhancements

- Collections and controllers now resolve many of Kirby's objects lazily only when the collection/controller requests them

## Ready?

### For review team

### DIFF
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -13,6 +13,7 @@ use Kirby\Http\Uri;
 use Kirby\Panel\Page as Panel;
 use Kirby\Template\Template;
 use Kirby\Toolkit\A;
+use Kirby\Toolkit\LazyValue;
 use Kirby\Toolkit\Str;
 use Throwable;
 
@@ -308,14 +309,20 @@ class Page extends ModelWithContent
 		$data = array_merge($data, [
 			'kirby' => $kirby = $this->kirby(),
 			'site'  => $site  = $this->site(),
-			'pages' => $site->children(),
-			'page'  => $site->visit($this)
+			'pages' => new LazyValue(fn () => $site->children()),
+			'page'  => new LazyValue(fn () => $site->visit($this))
 		]);
 
 		// call the template controller if there's one.
-		$controllerData = $kirby->controller($this->template()->name(), $data, $contentType);
+		$controllerData = $kirby->controller(
+			$this->template()->name(),
+			$data,
+			$contentType
+		);
 
 		// merge controller data with original data safely
+		// to provide original data to template even if
+		// it wasn't returned by the controller explicitly
 		if (empty($controllerData) === false) {
 			$classes = [
 				'kirby' => App::class,
@@ -325,17 +332,19 @@ class Page extends ModelWithContent
 			];
 
 			foreach ($controllerData as $key => $value) {
-				if (array_key_exists($key, $classes) === true) {
-					if ($value instanceof $classes[$key]) {
-						$data[$key] = $value;
-					} else {
-						throw new InvalidArgumentException('The returned variable "' . $key . '" from the controller "' . $this->template()->name() . '" is not of the required type "' . $classes[$key] . '"');
-					}
-				} else {
-					$data[$key] = $value;
-				}
+				$data[$key] = match (true) {
+					// original data wasn't overwritten
+					array_key_exists($key, $classes) === false => $value,
+					// original data was overwritten, but matches expected type
+					$value instanceof $classes[$key] => $value,
+					// throw error if data was overwritten with wrong type
+					default => throw new InvalidArgumentException('The returned variable "' . $key . '" from the controller "' . $this->template()->name() . '" is not of the required type "' . $classes[$key] . '"')
+				};
 			}
 		}
+
+		// unwrap remaining lazy values in data
+		$data = LazyValue::unwrap($data);
 
 		return $data;
 	}
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -344,6 +344,7 @@ class Page extends ModelWithContent
 		}
 
 		// unwrap remaining lazy values in data
+		// (happens if the controller didn't override an original lazy Kirby object)
 		$data = LazyValue::unwrap($data);
 
 		return $data;
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -29,6 +29,7 @@ use Kirby\Text\KirbyTags;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\Config;
 use Kirby\Toolkit\Controller;
+use Kirby\Toolkit\LazyValue;
 use Kirby\Toolkit\Str;
 use Kirby\Uuid\Uuid;
 use Throwable;
@@ -364,7 +365,10 @@ class App
 	{
 		return $this->collections()->get($name, array_merge($options, [
 			'kirby' => $this,
-			'site'  => $this->site(),
+			'site'  => $site = $this->site(),
+			'pages' => new LazyValue(fn () => $site->children()),
+			'users' => new LazyValue(fn () => $this->users())
+
 		]));
 	}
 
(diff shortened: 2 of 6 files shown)


afbora (dismissed):
I liked the idea 👍 Simple and effective.

lukasbestle (commented):
Does this really increase performance much? The collection closure is only called once per request and the objects already exist in memory and are only referenced (not cloned).

Reading the docblock ("all relevant dependencies"), I feel like the `$pages` and `$users` variables are probably quite common in collections as they are the natural starting point for any custom filtering etc. So the breaking change would be significant. That's IMO only worth it if there is a real and significant performance improvement tied to it.

distantnative:
Well if you have some large users collection etc, I think this could be quite the pain point if you actually just want to use pages in your collection

lukasbestle:
As I wrote, AFAICT the variable does not get cloned, only referenced. An object reference is quite small and is the same no matter if the object itself is big or small.

distantnative:
But if the object wasn't created before?

bastianallgeier:
This is a pretty massive breaking change and it feels like a little too late to introduce this to be honest. If we still want to do this, we could try to use a closure instead and resolve those in the controller logic.

distantnative:
@bastianallgeier I kinda like the idea of wrapping them in closures. However, that would mean that controller arguments can't be closures anymore as the Controller class would then unwrap any closure as parameter.

lukasbestle:
> But if the object wasn't created before?

You are right, that's a good point.

> I kinda like the idea of wrapping them in closures. However, that would mean that controller arguments can't be closures anymore as the Controller class would then unwrap any closure as parameter.

What if we allow to pass a single closure to `$controller->call()` that then receives the names of the used parameters and can lazily provide them?

So something like:

```php
return $this->collections()->get($name, function (array $args) use ($options) {
	$options['kirby'] = $this;

	if (in_array('site', $args) === true) {
		$options['site'] = $this->site();
	}

	if (in_array('pages', $args) === true) {
		$options['pages'] = $this->site()->children();
	}

	if (in_array('users', $args) === true) {
		$options['users'] = $this->users();
	}

	return $options;
});
```

The `Controller` class would then first get just a list of argument names from the reflection, then call the closure and finally pass the return value of the closure through `$controller->arguments()`.

It's more verbose, but avoids the breaking change of not being able to provide closures as arguments. The current way to just pass an array would still work and would be preferred if dynamic behavior is not needed.

distantnative:
@lukasbestle Would that work with the variadic parameter we currently support?

lukasbestle:
Hm, you are right. Not really, because then we would need to tell the closure to return everything it can, e.g. with a separate `bool $variadic` closure argument. But then it gets too complex IMO.

lukasbestle:
Another alternative:

A new `bool $lazy = false` param for `Controller::call()`.

By default it would be false for backwards-compatibility (= if a closure is passed in the data array, it is passed 1:1 to the controller closure).

But if set to true, closures will be called if their value if needed. In this mode, you could still pass a closure to the controller by returning a closure from the data closure.

The migration could look like this:

- In v4, we introduce the new lazy mode with a default of false.
- If the data contains a closure but lazy is false, a deprecation warning is thrown.
- In v5 we remove the lazy param and make the method always use lazy mode.

Yo Dawg, heard you like closures, So I returned a closure from a closure so I can pass a closure to your closure

bastianallgeier:
Quick thought: What about a custom LazyClosure that can be detected reliably by the controller to not accidentally unwrap a wanted Closure?

afbora:
The first thing that came to my mind was to use `$controller->arguments()` to extract parameters and send objects according to the requested parameters as Lukas proposed. If the parameter is variadic, all objects should be sent as it is now.


Example spaghetti code snippet that explaining how it works 🙈 

````php
$data = $controller->extractDataFromArguments();
$controller->call(null, $data);

public function extractDataFromArguments(): array
{
    $kirby = App::instance();
    $info  = new ReflectionFunction($this->function);
    $args  = [];
    
    foreach ($info->getParameters() as $param) {
        $name = $param->getName();

        if ($param->isVariadic() === false) {
            $args[] = $name;        
        }
    }
    
    $data = [];
    foreach ($args as $key) {
        switch ($key) {
            case 'kirby':
                $data['kirby'] = $kirby;
                break;
            case 'site':
                $data['site'] = $kirby->site();
                break;
            case 'pages':
                $data['pages'] = $kirby->site()->children();
                break;
            case 'users':
                $data['users'] = $kirby->users();
                break;
            default:
                return [
                    'kirby' => $kirby,
                    'site'  => kirby->site(),
                    'pages' => kirby->site()->children(),
                    'users' => $kirby->users()
                ];
        }
    }

    return $data;
}
````

**Comment by distantnative** [comment]
I guess that would work @afbora but I don't feel super good about hardcoding these few arbitrary data points. It would also be a breaking change as nobody would be able to pass other data for those keys anymore.

**Comment by afbora** [comment]
Yeah, it doesn't look very good :) I thought maybe it could be written in a better way using this logic.

This piece of code may contain a breaking change, this was just an example. Yes, custom data parameter should be sent either in this logic.

**Comment by lukasbestle** [comment]
> Quick thought: What about a custom LazyClosure that can be detected reliably by the controller to not accidentally unwrap a wanted Closure?

I thought about that as well but got stuck in a dead end because `Closure` is a final class. But now I realized we could just wrap the closure in another class instance:

```php
return $this->collections()->get($name, [
    'pages' => new LazyClosure(fn () => $this->site()->children())
]);
```

I'd say that doesn't look too bad. It also completely avoids any breaking change and is easy to understand when reading the code.

distantnative:
I like it. Just contemplating the name... cause those are actually the closures that should be unwrapped vs.  normal closures being left untouched. But maybe it's still fine.

lukasbestle:
Maybe `LazyValue`?

bastianallgeier:
I'm happy with LazyValue. We could even reuse this in other places later. Maybe it should be in the Toolkit namespace then.

distantnative:
Added a first version, what do you think?

afbora:
I liked a lot new `LazyValue` class 👍 

I'm thinking out loud. Not sure about the idea. If we do the same for the page controller, will there be a performance increase? Because I think page controller is used more than collections controller.

https://github.com/getkirby/kirby/blob/v4/enhancement/collections-remove-users/src/Cms/Page.php#L308-L313

src/Cms/Page.php
```diff
@@ -325,18 +332,20 @@ public function controller(
...
+					array_key_exists($key, $classes) === false => $value,
+					// original data was overwritten, but matches expected type
+					$value instanceof $classes[$key] => $value,
+					// throw error if data was overwritten with wrong type
+					default => throw new InvalidArgumentException('The returned variable "' . $key . '" from the controller "' . $this->template()->name() . '" is not of the required type "' . $classes[$key] . '"')
+				};
 			}
 		}
 
+		// unwrap remaining lazy values in data
```
lukasbestle, line 346:
```suggestion
		// unwrap remaining lazy values in data
		// (happens if the controller didn't override an original lazy Kirby object)
```

lukasbestle (commented):
I like it a lot, very clean. 👍
Just needs tests.

lukasbestle (approved):

afbora:
How reasonable is it to bring it also to 3.9? I think it can be a good performance enhancement for those who will not switch to v4 or who are thinking of switching to v4 in the future.

PR #5609, opened 2023-09-09
`Panel\Menu::link()` helper for custom menu links

distantnative:
## This PR …

Credits to @lukaskleinschmidt 

- [ ] @bastianallgeier do you think this makes sense to include in the core?
- [ ] Unit tests

### Enhancement
- New `Panel\Menu::link()` helper method to generate custom Panel menu links:

```php
'panel' => [
	'menu' => [
		'site',
		'-',
		'fields' => Menu::link('pages/fields', 'Fields', 'pen'),
		'sections' => Menu::link('pages/sections', 'Sections', 'grid'),
	]
]
```

Registering those links with the helper will also ensure that the `site` menu entry is not shown as current when one of the custom links is active/current (even though the linked target likely will lie within the site area).

### DIFF
--- a/config/areas/site.php
+++ b/config/areas/site.php
@@ -1,6 +1,9 @@
 <?php
 
+use Kirby\Cms\App;
+use Kirby\Panel\Menu;
 use Kirby\Toolkit\I18n;
+use Kirby\Toolkit\Str;
 
 return function ($kirby) {
 	return [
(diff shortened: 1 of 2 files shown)


bastianallgeier:
I think the main question is for us how we handle PHP classes for UI features in the future. This is not just about menu links, but also about dialogs and other stuff. It's still a bit chaotic at the moment.

distantnative:
@bastianallgeier I think you need to elaborate a bit cause I don't understand your thoughts yet

lukaskleinschmidt:
I think he is talking about a (most likely) class based representation of a vue component. Like the stuff that should end up as props in the `Menu.vue` component or any other component that is kinda configurable?

bastianallgeier:
What I mean is that we start to have more and more of those classes that somehow represent frontend features (i.e. PageCreateDialog). Those classes are responsible to create the right props for frontend components. That's the power of Fiber and I think we should follow this direction more. But that would speak for something like

```php
$menu = new Kirby\Panel\Ui\Menu();
$dialog = new Kirby\Panel\Ui\Dialog();
```

etc. 

Right now, it's somehow stuffed into the Panel namespace without a proper concept. If we introduce this helper though, it will become a real tool for plugin devs and that's why I would love to already find a good place for it.

distantnative:
Ok, er can also pause the helper for now until we find a proper solution without having to rush the solution just for this PR

PR #5610, opened 2023-09-09
Text field: `font` option for monospace

distantnative:
### Enhancement
- Text fields: new `font: monospace` option https://kirby.nolt.io/558

### DIFF
--- a/config/fields/text.php
+++ b/config/fields/text.php
@@ -28,6 +28,13 @@ return [
 		},
 
 		/**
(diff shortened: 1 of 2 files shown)


afbora:
Wouldn't it be better to allow any font that added with a plugin or custom panel css, not just `monospace`?

distantnative:
@afbora this is the same as the `font` option of the textarea field

afbora (approved):

PR #5613, opened 2023-09-09
Fix search type permissions

afbora:
## This PR …

Now search parts/types respect to user permissions.

````yaml
permissions:
    access:
        site: false # disable pages and files type
        users: false # disable user type
````

### Fixes
- Fix search type permissions

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/areas/search/views.php
+++ b/config/areas/search/views.php
@@ -5,6 +5,20 @@ use Kirby\Cms\App;
 return [
 	'search' => [
 		'pattern' => 'search',
+		'when'    => function (): bool {
+			$kirby = App::instance();
+
+			return (
+				$kirby->user()
+					?->role()
+					->permissions()
+					->for('access', 'site') === true ||
+				$kirby->user()
+					?->role()
+					->permissions()
+					->for('access', 'users') === true
+			);
+		},
 		'action'  => function () {
 			return [
 				'component' => 'k-search-view',
--- a/config/areas/search/views.php
+++ b/config/areas/search/views.php
@@ -1,19 +1,18 @@
 <?php
 
 use Kirby\Cms\App;
+use Kirby\Panel\Panel;
+use Kirby\Panel\View;
 
 return [
 	'search' => [
 		'pattern' => 'search',
 		'when'    => function (): bool {
-			if ($user = App::instance()->user()) {
-				$permissions = $user->role()->permissions();
+			// checks core and custom search areas
+			$permissions = App::instance()->user()?->role()->permissions()->toArray() ?? [];
+			$searches    = View::searches(Panel::areas(), $permissions);
 
-				return $permissions->for('access', 'site') === true ||
-					$permissions->for('access', 'users') === true;
-			}
-
-			return false;
+			return empty($searches) === false;
 		},
 		'action'  => function () {
 			return [
--- a/config/areas/search/views.php
+++ b/config/areas/search/views.php
@@ -6,18 +6,14 @@ return [
 	'search' => [
 		'pattern' => 'search',
 		'when'    => function (): bool {
-			$kirby = App::instance();
+			if ($user = App::instance()->user()) {
+				$permissions = $user->role()->permissions();
 
-			return (
-				$kirby->user()
-					?->role()
-					->permissions()
-					->for('access', 'site') === true ||
-				$kirby->user()
-					?->role()
-					->permissions()
-					->for('access', 'users') === true
-			);
+				return $permissions->for('access', 'site') === true ||
+					$permissions->for('access', 'users') === true;
+			}
+
+			return false;
 		},
 		'action'  => function () {
 			return [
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: config/areas/search/views.php [hunk 1]
```diff
@@ -5,6 +5,20 @@
...
+				$kirby->user()
+					?->role()
+					->permissions()
+					->for('access', 'site') === true ||
+				$kirby->user()
+					?->role()
+					->permissions()
+					->for('access', 'users') === true
+			);
+		},
```
distantnative, line 16: What if a plugin area has registered a search extension?

afbora: I'm not sure, but wouldn't that be the responsibility of the plugin?

afbora: Can you share me a sample search extension? Let's see what happens.

distantnative: The plugin takes care of the search type - but with that change you'd take away the search view even if the plugin's search type is still allowed for the user

distantnative: https://getkirby.com/docs/reference/plugins/extensions/panel-search

config/areas/site/searches.php
```diff
@@ -4,53 +4,63 @@
...
-				->search($query)
-				->filter('isListable', true)
-				->paginate($limit, $page);
+if (
+	App::instance()
+		->user()
+		?->role()
+		->permissions()
+		->for('access', 'site') === true
+) {
```
distantnative, line 13: Do avoid wrapping everything in an if statement, could we flip it around?

```php
if (... === false) {
  return [];
}
```

panel/src/components/View/Menu.vue
```diff
@@ -53,6 +54,12 @@ export default {
 	computed: {
 		menus() {
 			return this.$panel.menu.entries.split("-");
+		},
+		canSearch() {
+			return (
+				this.$panel.permissions.access['site'] === true ||
+				this.$panel.permissions.access['users']
```
distantnative: Same as with the PHP search view, this cannot be tied to just the site and users. Rather should be something like checking the length of `panel.searches`

panel/src/components/Dialogs/SearchDialog.vue
```diff
@@ -105,6 +104,14 @@ export default {
...
 				Object.values(this.$panel.searches)[0]
 			);
+		},
+		type() {
+			const type = this.$panel.view.search;
+
+			return (
+				this.$panel.searches[type] ?
+					type : Object.keys(this.$panel.searches)[0]
+			);
```
distantnative: Won't this break the types dropdown if you move `type` from data to be a computed prop? cause now the type cannot be reassigned.

afbora: I don't know the Vue process very well. Data and computed do the same thing in my knowlegde/view 🙈 One is predefined and the other is computed 😅

distantnative: Yes, but if it is a computed prop, you cannot do `type = 'users'` anymore. But that's what's done via the type dropdown.

afbora:
@distantnative I've improved the codes. How about now?

distantnative:
I am not sure if it's worth the extra performance load to not add the search view.
In my opinion, hiding the button could be sufficient. Or we add in the Vue SearchView component a redirect when search type length is 0.

Another thought I had:
Should the permission logic really be added to the individual files? Or do we rather need a central logic. Cause I would imagine that if a user has no access to an area, it should never has access to the searches from that area, should it?

afbora:
> In my opinion, hiding the button could be sufficient.

OK, if it does not create a security problem, why not? I'm just not sure.

> Cause I would imagine that if a user has no access to an area, it should never has access to the searches from that area

Can you still access it with this PR right now? Can you give me an example?

distantnative:
What's the security problem with a search view without types?

And e.g. I add a custom area for todos and a todos search type. I would have to manually implement the check for my todos search type as well to only add it when the user has access to my todos area. But I think we could directly implement this in `View::searches`:

`if no accesss permissions for area, skip area in loop`

distantnative:
@afbora was thinking in this direction https://github.com/getkirby/kirby/commit/a37869f145797686ecac2849dcc074bb63c48cc7

afbora:
@distantnative OK, I like the your idea but custom search areas doesn't work for your branch.

distantnative:
@afbora force-pushed a fix to my branch, hope it works now

afbora:
@distantnative Works great now for that issue.

But I found an issue. Default `pages` types selected when only one type available.

````yaml
permissions:
    access:
        site: false
        users: true
````

Steps:
1.  Use above blueprint for a role (for ex: editor)
2. Log-in as editor
3. Try to search from `/panel/search` view (not from dialog)
4. You can't find anything about users
5. You'll see the url is: `/panel/search?type=pages&query=foo`

Another steps:
When you enter `/panel/search`, you'll be redirected to `/panel/search?type=pages` even if a role have not `site` permission.

I think you should create a PR and lets discuss about issues in there.

distantnative:
@afbora created the PR with a fix here: https://github.com/getkirby/kirby/pull/5757

PR #5614, opened 2023-09-10
Nested dialogs

distantnative:


distantnative:
@bastianallgeier on second thoughts, should we move this to 4.1 or 4.2?

bastianallgeier:
@distantnative it seems like a pretty big change for the beta. Do you think we can still offer this without a breaking change later though?

distantnative:
@bastianallgeier looking at the code, I don't see a breaking change. So 4.1, 4.2 would be fine IMO

distantnative:
Will open this again when the time is right

PR #5631, opened 2023-09-14
Fix `$collection->group()` for case-sensitive

### LINKED ISSUE #5628: Collection->group() throws error with $i set to false
## Description
When grouping a collection, the key is by default lower-cases. This is not always useful. Theoretically, I should be able to pass a $i = false as second argument  to prevent lowercasing, but then the method breaks, because then $value is not a string. So 

```
// ignore upper/lowercase for group names

$value = $i ? Str::lower($value) : $value->value();
```
would fix this.

**Expected behavior**  
Method should work with $i set to false

## To reproduce
Try to group a collection by a key, passing `false` as a second argument:

```
$grouped = $page->children()->groupBy('category', false);
```

Error: Illegal offset type in isset or empty in Kirby\Cms\Collection #174

## Your setup
**Kirby Version**  
Kirby 3.9.6.1

https://forum.getkirby.com/t/groupby-kills-the-spelling-of-the-grouped-by-field/29439/2?u=texnixe

distantnative:
### Fixes
- Fix `$collection->group()` for case-sensitive
#5628

## Ready?

### For review team

### DIFF
--- a/src/Cms/Collection.php
+++ b/src/Cms/Collection.php
@@ -168,9 +168,7 @@ class Collection extends BaseCollection
 				}
 
 				// ignore upper/lowercase for group names
-				if ($i) {
-					$value = Str::lower($value);
-				}
+				$value = $i === true ? Str::lower($value) : (string)$value;
 
 				if (isset($groups->data[$value]) === false) {
 					// create a new entry for the group if it does not exist yet
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -531,7 +531,7 @@ class Collection extends Iterator implements Countable
 				$value = $this->getAttribute($item, $field);
 
 				// ignore upper/lowercase for group names
-				return $i === true ? Str::lower($value) : $value;
+				return $i === true ? Str::lower($value) : (string)$value;
 			});
 		}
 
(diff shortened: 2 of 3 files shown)


afbora:
Why is the parameter name `$i`?. I spent a while trying to figure out what `$i` is for. How about a better variable name?

lukasbestle:
What about `$ignoreCase`? It would be a breaking change with PHP named params, but I agree it would be very worth it.

afbora:
Or may be `bool $caseSensitive = false`

distantnative:
@afbora But then it wouldn't be just a breaking change of the parameter name but all existing calls would do exactly the opposite of what they did before

distantnative:
It gets even messier. `Collection` has a `$caseSensitive` class prop.  Though this seems to be only used in one test. I am wondering if we should deprecate that to get a bit more clarity.

On the parameter name, I'd go with `$caseInsensitive` as that's used throughout the `Str` class.

distantnative:
Ok, revert... if we rename this parameter, this fix can't be in 3.9.7.

I suggest we merge the fix without the renamed parameter and rename the parameter for v4: https://github.com/getkirby/kirby/pull/5634

lukasbestle (approved):

lukasbestle:
> `Collection` has a `$caseSensitive` class prop. Though this seems to be only used in one test. I am wondering if we should deprecate that to get a bit more clarity.

We introduced that not too long ago, I use that in the Versions plugin.

PR #5635, opened 2023-09-16
Deprecate `Collection::$caseSensitive`

distantnative:
### Deprecated
- `Collection::$caseSensitive` has been deprecated and will be removed in a future version.

### DIFF
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -5,6 +5,7 @@ namespace Kirby\Toolkit;
 use Closure;
 use Countable;
 use Exception;
+use Kirby\Cms\Helpers;
 
 /**
  * The collection class provides a nicer
@@ -29,6 +30,7 @@ class Collection extends Iterator implements Countable
 	/**
 	 * Whether the collection keys should be
 	 * treated as case-sensitive
+	 * @deprecated 4.0.0
 	 *
 	 * @var bool
 	 */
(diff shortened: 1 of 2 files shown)


lukasbestle:
We introduced this not too long ago in https://github.com/getkirby/kirby/pull/2613. I use it in the Versions plugin. As Bastian wrote in that PR, it may actually make sense to use the case-sensitive mode as the default in the future.

I agree that the name is not great though when compared to the `$caseInsensitive` variables we use in other places.

Maybe we could introduce a new `$caseInsensitive` prop and named constructor argument as a replacement.

distantnative:
For me the part I really don't like is having such a prop for the class, but then having other methods that ignore this one but use a parameter. That just feels very inconsistent.

lukasbestle:
There is a connection, but I feel like one way cannot replace the other. The class property handles the collection itself and the method argument for `group()` handles the derived grouped collection. Depending on which field you group by, the derived collection is not necessarily going to need the same behavior as the collection itself. E.g. the collection itself uses case-sensitive UUIDs as keys but the derived collection is grouped by a case-insensitive field value.

distantnative:
I see your point.

But I don't think the current state is fine. It feels like a real hack for a specific use case, not like an actual class design. One could make it the default mode to be case-sensitive. v4 would be a good moment then.

But I am wondering if that makes sense in the context of Kirby where so much else is always case-insensitive. Feels like a bit of an alien.

lukasbestle:
I also get your point. :)

The huge advantage of the case-sensitive mode is the increased performance as it avoids having to `strtolower()` every key, both when setting and when getting. So case-sensitive mode should really be the default for all use cases that don't explicitly need the insensitive behavior.

From the top of my head I don't have a good overview which parts of Kirby are case-sensitive and which aren't. My guess is that we are case-insensitive everywhere PHP is (e.g. fields, because PHP method calls are case-insensitive anyway). But e.g. with arrays, PHP is already case-sensitive. The question is: Would parts break (and if so, which) if we switch to case-sensitive collections? TBH I feel like this is too much of a gamble so late in the process.

I'm just not a huge fan of deprecating the prop without providing an alternative. What if we add a todo comment with a link to this discussion so we can re-evaluate this early in the v5 process?

PR #5638, opened 2023-09-16
Upload dialog: fix error overflow

### LINKED ISSUE #5622: [v4] New Upload Dialog: Error message breaks layout
## Description
Trying to upload a file that already exists, the remove button is pushed outside the container by the error message:

Screenshot-1
Screenshot 2

## Your setup
**Kirby Version**  
v4-beta.1

distantnative:
@bastianallgeier I feel like it's the right direction to merge the two grid rows - but if the error text gets really long, it'll still break the layout.


bastianallgeier:
I've added a reproduction test to the lab and fixed a few more things. I think the result is pretty nice now …

Screenshot 2023-09-18 at 13 45 40

bastianallgeier (approved):

Issue #5615, opened 2023-09-11
Database: combination of unsuccessful `first()` and custom fetch Closure throw an exception


adamkiss:
## Description

```php
// throws exception
db()->table('users')
  ->where('username', 'doesnt-exist-at-all')
  ->fetch(fn($row) => some_custom_method($row))
  ->first();

// works
db()->table('users')
  ->where('username', 'doesnt-exist-at-all')
  ->fetch(fn($row) => some_custom_method($row))
  ->limit(1)
  ->all();
```

A combination of database `fetch` Closure and unsuccessful `first` query results in error in `src/Database/Database.php:383`.

The difference between the `first()` call and the `->limit(1)->all()` call is that one calls `PDOStatement::fetch` and the other calls `PDOStatement::fetchAll` - and Kirby, expecting empty array in case of query with 0 returned rows, just tries to foreach the results.

I think the fix is to simply add guard against `$results !== false` (or maybe `is_array($results`), but you might also want to force `fetchAll` method if fetch is closure… there are many ways how to go about this.

**Expected behavior**  

I expect both calls to succeed (or return falsy result).

## To reproduce

1. Create a database
2. Do an unsuccessful call
3. Observe error

## Your setup
**Kirby Version**  
3.9.6.1

PR #5639, opened 2023-09-16
Fix failed database query with fetch closure

distantnative:
### Fixes
- #5615

## Ready?

### For review team

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -3,9 +3,10 @@
 namespace Kirby\Database;
 
 use Closure;
-use Exception;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Toolkit\A;
+use Kirby\Toolkit\Collection;
+use Kirby\Toolkit\Obj;
 use Kirby\Toolkit\Str;
 use PDO;
 use PDOStatement;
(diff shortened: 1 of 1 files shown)


distantnative:
@adamkiss does it work for your case when you add this check https://github.com/getkirby/kirby/pull/5639/files#diff-269fc35a258fbfc31c559ec9e4bbda254659f9007d4af68474f5663bbd9eda11R389 ?

lukasbestle:
Doesn't this change completely disable the fetch closure if there is just one result? I think we either need to manually call the closure with the single result (if there is one) or force the `fetchAll()` mode if a fetch closure was passed as Adam suggested.

adamkiss:
Let me test it, Nico

adamkiss:
No, it doesn't, and it breaks in different ways (expecting array of rows, array of row data from `first()` iterates over the columns and sends wrong data to the closure)

adamkiss:
@distantnative I thought of something more along these lines:

```php
// apply the fetch closure to all results if given
if ($options['fetch'] instanceof Closure) {
	// ↓
	// wrap a singular result in an array
	if ($options['method'] === 'fetch') {
		$results = ($results === false) ? [] : [$results];
	}
	// ↑

	foreach ($results as $key => $result) {
		$results[$key] = $options['fetch']($result, $key);
	}
}
``` 

But that doesn't respect `iterator` and always returns an array. I can't think of a more elegant code right now.

lukasbestle:
What about something like:

```php
if ($options['fetch'] instanceof Closure) {
	if (is_iterable($results) === true) {
		foreach ($results as $key => $result) {
			$results[$key] = $options['fetch']($result, $key);
		}
	} else {
		$results = $options['fetch']($results, null);
	}
}
```

distantnative:
Closing in favor of https://github.com/getkirby/kirby/pull/5640

PR #5640, opened 2023-09-16
Fix failing first/one/row queries when using fetch closure

### LINKED ISSUE #5615: Database: combination of unsuccessful `first()` and custom fetch Closure throw an exception

adamkiss:
### Fixes

- #5615 

This PR replaces #5639 so I can work on it instead Nico. I also added his reformatting from that PR, and added a unit test. 

Note: I had to change function descriptions on first/one/row, because since `->fetch()->first()` now works, it can return absolutely anything (in tests, I use the fetch function to return string)

cc @lukasbestle @distantnative 

## Ready?

### For review team

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -391,7 +391,7 @@ class Database
 				foreach ($results as $key => $result) {
 					$results[$key] = $options['fetch']($result, $key);
 				}
-			} else if ($options['method'] === 'fetch' && $results !== false) {
+			} elseif ($options['method'] === 'fetch' && $results !== false) {
 				// fetching a single record
 				$results = $options['fetch']($results);
 			}
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -3,9 +3,10 @@
 namespace Kirby\Database;
 
 use Closure;
-use Exception;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Toolkit\A;
+use Kirby\Toolkit\Collection;
+use Kirby\Toolkit\Obj;
 use Kirby\Toolkit\Str;
 use PDO;
 use PDOStatement;
@@ -337,13 +338,16 @@ class Database
 	/**
 	 * Executes a sql query, which is expected to return a set of results
 	 */
-	public function query(string $query, array $bindings = [], array $params = [])
-	{
+	public function query(
+		string $query,
+		array $bindings = [],
+		array $params = []
+	) {
 		$defaults = [
 			'flag'     => null,
 			'method'   => 'fetchAll',
-			'fetch'    => 'Kirby\Toolkit\Obj',
-			'iterator' => 'Kirby\Toolkit\Collection',
+			'fetch'    => Obj::class,
+			'iterator' => Collection::class,
 		];
 
 		$options = array_merge($defaults, $params);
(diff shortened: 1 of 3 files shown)


src/Database/Query.php
```diff
@@ -623,7 +623,7 @@ protected function execute(string|array $sql, array $params = [])
 	/**
 	 * Selects only one row from a table
 	 */
-	public function first(): object|array|false
+	public function first(): mixed
```
distantnative, line 626: @adamkiss what else do we need to support here? with what use case do the more specific typings fail?

adamkiss: @distantnative anything that your `fetch(Closure)` returns. If it returns a string… `first()` returns a string as well. See the added unit test - that's how I actually found out that `->fetch()->first()` will do that.

src/Database/Database.php
```diff
@@ -379,8 +386,14 @@ public function query(string $query, array $bindings = [], array $params = [])
...
-			foreach ($results as $key => $result) {
-				$results[$key] = $options['fetch']($result, $key);
+			if ($options['method'] === 'fetchAll') {
+				// fetching multiple records
+				foreach ($results as $key => $result) {
+					$results[$key] = $options['fetch']($result, $key);
+				}
+			} elseif ($options['method'] === 'fetch' && $results !== false) {
+				// fetching a single record
+				$results = $options['fetch']($results);
```
lukasbestle: I think we need to explicitly pass `null` as the second argument, otherwise we will get "too few arguments" errors when using a closure or callable that expects two arguments.

distantnative: Not sure if I would tie this as closely to the exact methods. How about
`is_iterable($results)` for the first if and just `$results !== false` for the elseif?

lukasbestle (commented):
I haven't tested it myself, but the changes look good to me. A small suggestion:

src/Database/Database.php
adamkiss: Can't do - the method is the differentiator. the fetch (`->first()` et al) returns:

```
[
	'column1' => 'value',
	'column2' => 'value'
]
```

which passes `is_iterable()` as well, except now we're trying to send 'value' to closure expecting array.

lukasbestle: Ah, right, of course. Each row is an array.

lukasbestle (approved):

PR #5641, opened 2023-09-16
Proper `PluginAsset` and `PluginAssets` classes

### LINKED ISSUE #5164: Plugin asset CSS files miss timestamp in URL
## Description
According to [Plugin Basics > Timestamps](https://getkirby.com/docs/guide/plugins/plugin-basics#plugin-assets__timestamps) any asset included inside a plugin folder should automatically receive a timestamp when cached in the media folder. This is not happening, and it seems this has not worked for several years:

- Case 1: https://forum.getkirby.com/t/my-plugin-assets-are-missing-the-timestamp-in-the-file-name/28477
- Case 2: https://forum.getkirby.com/t/plugin-asset-url-helper/18209

**Expected behavior**  
Referring to an asset included inside a plugin should return a URL that points to a cached file which has a URL such as this:
`/media/plugins/superwoman/superplugin/myCss.1520265293.css`

**Actual Behavior**
The actual result of referring to an asset inside  a plugin folder is:
`/media/plugins/superwoman/superplugin/myCss.css`

**Screenshots**  
CleanShot 2023-04-30 at 11 35 23@2x

CleanShot 2023-04-30 at 11 34 45@2x

## To reproduce
1. Go to create a plugin and add a CSS file inside the asset folder of your plugin
2. Include the CSS file in a page template using `css('PATH')`
3. Open a page using this template
4. Look for the `<link>`-tag in the HTML

## Your setup
Kirby v3.9.1

**Console output**  
-

- PHP: 8.1.17

**Additional context**  
-

distantnative:
### Features
- Plugin assets get easily exposed via the PHP API
  - New `$plugin->assets()` collection
  - New `$plugin->asset('styles.css')` method
  - New `PluginAsset` object with many methods, e.g. `$plugin->asset('styles.css')->url()`
  - Plugin asset's media url contains a modification timestamp to easily cachebust (e.g. `https://getkirby.com/media/plugins/getkirby/test-plugin/2375797551-472389240/styles.css`)
  - `css()` and `js()` helpers support passing plugin and plugin assets objects to include all assets of the plugin:

```php
css([
  'assets/css/index.css',
  $kirby->plugin('foo/bar')
]);

css([
  'assets/css/index.css',
  $kirby->plugin('foo/bar')->assets(),
]);

css([
  'assets/css/index.css',
  $kirby->plugin('foo/bar')->asset('styles.css'),
]);
```

### Fixes
- #5164
- https://github.com/getkirby/getkirby.com/issues/148

### Deprecated
- Plugin asset urls now include a media hash. If you have hardcoded URLs to the plugin's media folder anywhere, you will need to update them to work in future versions. Best use `$kirby->plugin('plugin/name')->asset('yourAsset.css)->url()`

### Breaking changes
- `Plugin::assets()` doesn't return an array anymore, but a `PluginAssets` collection

## Ready?

### For review team


src/Cms/PluginAssets.php
```diff
@@ -42,6 +49,73 @@ public static function clean(string $pluginName): void
...
+				if (is_file($root . '/' . $path) === true) {
+					$assets[$path] = $root . '/' . $path;
+				}
+			}
+		}
+
+		$collection = new static([], $plugin);
+
+		foreach ($assets as $path => $root) {
+			$collection->$path = new PluginAsset($path, $root, $plugin);
```
lukasbestle: To avoid the magic setter:

```suggestion
			$collection->data[$path] = new PluginAsset($path, $root, $plugin);
```

src/Cms/PluginAsset.php
```diff
@@ -0,0 +1,112 @@
...
+	 */
+	public function mediaUrl(): string
+	{
+		return $this->plugin()->mediaUrl() . '/' . $this->path() . '?m=' . $this->mediaHash();
+	}
+
+	/**
+	 * Timestamp when asset file was last modified
+	 *
+	 * @codeCoverageIgnore
```
lukasbestle: We can test this with `touch($file, 1337000000)` in the test setup to get a known modification time.

src/Cms/PluginAsset.php
```diff
@@ -0,0 +1,112 @@
...
+	public function plugin(): Plugin
+	{
+		return $this->plugin;
+	}
+
+	/**
+	 * Publishes the asset file to the plugin's media folder
+	 * by creating a symlink
+	 *
+	 * @codeCoverageIgnore
```
lukasbestle: Is there something blocking us from testing this?

lukasbestle (commented):
I think it would be great if we had a way to avoid the timestamp in the query string. Query strings [prevent any kind of caching by most proxy caches](https://www.stevesouders.com/blog/2008/08/23/revving-filenames-dont-use-querystring/), while timestamps in the path don't.

What about inserting the timestamp directly into the `mediaRoot` like:

```php
return $this->plugin()->mediaRoot() . '/' . $this->mediaHash() . '/' . $this->path();
```

During the cleanup, we would still loop over the files in the media folder and whenever a file exists in the plugin, calculate the "correct" media hash. All media hash folders that no longer exist will be deleted. Would also simplify the logic because with the path as I suggested above, every plugin asset gets its own media hash folder, so each top-level folder matches 1:1 with a plugin asset (existing or not).

distantnative:
@lukasbestle Understood why we need to put the mediaHash in the path not a query parameter.

However, then it should go into mediaUrl, not mediaRoot, right? It doesn't matter whether the file is stored in the media folder in such a hash directory or not, the important part is what's in the Url.

But then we would need to modify the route to grab that mediaHash - and that would be a big breaking change as existing media urls wouldn't work anymore.

How about inserting it in the filename? `styles.29231278-123792310.css`?

lukasbestle:
Isn't the URL dependent on the root? If the web server serves the file directly, it will do so based on the root where it is stored.

I'd be fine with inserting it in the filename.

About the breaking change: What if we published both to the versioned and unversioned paths?

distantnative:
mediaRoot = absolute path inside to the location inside the media folder
mediaUrl = pattern that needs to be matched by our plugin assets route `'media/plugins/$provider/$pluginName/$path'`

So technically, the mediaHash doesn't need to be part of where we store the actual file, as long as it's in the url and our route + PluginAssets::resolve() can connect it to the right file.

lukasbestle:
But that would mean that every request to a plugin asset would have to pass through our router, right? With a root that matches the URL path, the web server could serve already existing files for us like it does with other files in the `media` folder.

distantnative:
@lukasbestle Updated it with adding the hash to the path. I don't think we can avoid the breaking change as I can't think of a way to have two routes (one matching with hash, other without hash) given that the asset will not just have a filename, but a whole path.

lukasbestle:
I have an idea, but it's not fully thought through yet:

1. Replace the route pattern with this, making the hash optional:

```diff
- $media . '/plugins/(:any)/(:any)/(:any)/(:all)'
+ $media . '/plugins/(:any)/(:any)/(?:(:any)/)?(:all)'
```

2. In `PluginAssets::resolve()`, check if `$hash` is set. If it is, we have a path like `/plugins/getkirby/test-plugin/some/asset.js`.  
    1. Try to find the asset with `$plugin->asset($hash . '/' . $path)`. If that succeeds, the route was accessed without hash, so we publish the asset to the hash-less path and return it.
    2. Otherwise we try `$plugin->asset($path)` and validate its hash as in the current implementation.
4. If `$hash` is not set, we have a path like `/plugins/getkirby/test-plugin/some-asset.js`.  
    1. Try to find the asset with `$plugin->asset($path)`. If that succeeds, skip hash validation, publish to the hash-less path and return it.

This should cover all four cases:
1. With hash and subfolder: `/plugins/getkirby/test-plugin/110971429-1337000000/some/asset.js`
3. With hash, without subfolder: `/plugins/getkirby/test-plugin/110971429-1337000000/some-asset.js`
4. Without hash, but with subfolder: `/plugins/getkirby/test-plugin/some/asset.js`
6. Without hash or subfolder: `/plugins/getkirby/test-plugin/some-asset.js`

distantnative:
@lukasbestle Currently thinking that the breaking change would maybe be the less evil than having support both. What do you think?

lukasbestle:
It would be quite a breaking change though. So far, plugin devs could only hardcode their asset paths. So every single plugin that uses assets will break.

distantnative:
@lukasbestle I keep reading this https://github.com/getkirby/kirby/pull/5641#issuecomment-1722469990 but it breaks my head a little bit each time as we'd need different versions of `$asset->publish()` I believe etc.

Really worried we are creating code we will have a super hard time to understand down the road

lukasbestle:
It would certainly make the code more complex. However I only see it as a migration path. We don't need to code in this logic forever, only until v5. This would allow us to deprecate the old URL structure in v4 without immediately breaking *every single* plugin asset URL.

I'm really not comfortable to add in this massive breaking change so late in the v4 process. A lot of plugins were already updated for v4 and now they would break (again), which can create a lot of versioning confusion when a plugin version that is marked to be compatible with v4 actually isn't.

So I suggest an approach like:

1. Add the backwards-compatible logic in a compact, internal and totally separate way that makes it easy to rip it out in v5. It does not need to be beautiful.
2. Clearly communicate in the v4 release notes that hardcoded URLs are deprecated, will be removed in v5 and that plugin devs should only use the asset URL method going forward.
3. Rip out the backwards-compatible code in v5.

bastianallgeier:
From what I can tell in the PR description and the comments, this looks super cool to me. Do you need me for the last steps here? Should I also have another look at the code?

lukasbestle:
Would be great to have some more eyes on this. I'll also take another look at the code, but your review is well appreciated.

distantnative:
Yes please review 👍

bastianallgeier:
The code looks good to me, but I did not have the time yet to create a test setup for it.

bastianallgeier (approved):

PR #5644, opened 2023-09-16
[v4] PHP 8.3 support

distantnative:
## This PR …

- [ ] First finish https://github.com/getkirby/kirby/pull/5648

### Features
- PHP 8.3 support


src/Content/ContentStorage.php
```diff
@@ -259,7 +259,7 @@ public function touchLanguage(string|null $lang): void
 	public function update(
 		string $version,
 		string|null $lang = null,
-		array $fields
+		array $fields = []
```
lukasbestle, line 262: Why is `$lang` optional? Couldn't it be:

```php
string|null $lang,
array $fields
```

lukasbestle (commented):
In 3.9.6, the `Find::file()` method didn't yet have the `empty($path)` check, I believe that was added in v4. So in 3.9.6 a call without path would have already been a fatal error, making these changes not breaking I think.

distantnative:
Will open again when ready

PR #5645, opened 2023-09-16
Tweak Vite setup

distantnative:
Mainly moves the Kirby dev mode flag creation into a Vite plugin


lukasbestle (commented):
Works great. Do we need the separate file or could we also define the plugin function inside `vite.config.js`?

distantnative:
@lukasbestle both works, I kinda liked the extra file just so that the config file gets a little leaner

lukasbestle (approved):
Makes sense, the config is already long enough. 👍

PR #5652, opened 2023-09-17
Prominent register menu btn

distantnative:
Screenshot 2023-09-17 at 17 30 35

### DIFF
--- a/src/Panel/Menu.php
+++ b/src/Panel/Menu.php
@@ -217,6 +217,8 @@ class Menu
 				'icon'     => 'key',
 				'dialog'   => 'registration',
 				'text'     => I18n::translate('license.register'),
+				'variant'  => 'filled',
+				'theme'    => 'notice',
 			]);
 		}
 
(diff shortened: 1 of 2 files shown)


bastianallgeier:
I think it actually needs to be even more prominent in the end. But I think this is a good first step.

bastianallgeier (approved):

nilshoerrmann:
Not sure if this is the right place for feedback but looking at the current button style, I'd think an error occured. But this should be a positive action in my eyes (you want people to buy a license, and they want to register their installs). This could be more like a blue post-it saying "You are using an unregistered version of Kirby [ Register now ]".

bastianallgeier:
@nilshoerrmann I like the idea to turn it into something positive.

nilshoerrmann:
> This could be more like a blue post-it saying "You are using an unregistered version of Kirby [ Register now ]".

In the style of "Try › ❤️ › Buy", you could also write something like "Ready to launch? [ Register now ]" making clear that trying things out first is fine but you are required to pay for public sites.

nilshoerrmann:
Just to get this out of my head, could be something like this:

kirby_register-now

PR #5654, opened 2023-09-17
TOTP challenge

distantnative:
### TODO

- [x] @lukasbestle Additional TOTP class enhancements
- [x] @lukasbestle TOTP: Include issuer in label string
- [x] https://github.com/getkirby/kirby/pull/5666
- [x] @distantnative Allow admins to disable the second factor of another user
- [x] @distantnative Secret storage in `.htpasswd`
- [x] @lukasbestle Confirm that plain-text secret storage is unavoidable
- [x] @distantnative Activate dialog: add sentence how important it is to store the secret, otherwise login won't work
- [x] https://github.com/getkirby/kirby/pull/5668
- [ ] Point link in setup dialog to auth guide
- [x] @distantnative  Tests

### Setup
Config

```php
'auth' => [
  'methods' => [
    'password' => ['2fa' => true]
  ]
]
```

TOTP dialog

## Features

- TOTP (time-based one-time codes) are now supported for two-factor authentication
- New `Kirby\Toolkit\Totp` class

## Fixes
- Exceptions don't prefix i18n keys with error prefix if already prefixed

## Deprecated

- The internal method `$user->passwordFile()` was renamed `$user->secretsFile()`. The old method is kept as an alias for now but will be removed in a future version.

## Breaking changes

- The file format of the user secrets file (`.htpasswd`) was changed. The old format with just the password hash is still supported, however v4 can now store additional secrets in a JSON object in the second line of the file. This makes secrets files with a TOTP secret or other secrets incompatible with Kirby 3 sites.


lukasbestle:
I've pushed my TOTP implementation.

One bug I noticed while testing:

If I leave the confirmation field blank, the dialog closes, but the secret is not stored with the account. So as a user I would think that TOTP has been activated when it has in fact not been. I think it would be good to make the confirmation field required and enforce the validation in the submit action on the backend.

distantnative:
@lukasbestle You're right. We need to separate activation and disabling into their own dialogs - then it won't be a problem. And cleaner overall.

lukasbestle:
About secret storage: I was thinking we could store it in the second line of `.htpasswd`. We chose that filename back then because Apache blocks access to it by default, so it's another layer of security.

Using the file for both would be kind of a breaking change (accounts that enable TOTP in v4 cannot be used in v3), but I think it would be worth it.

We could extend the same format later by storing the FIDO2 public key in the third line etc.

distantnative:
@lukasbestle I think that's ok - we have had other changes in the past where migration backwards isn't possible.

lukasbestle:
> Should TOTP be available for reset password, code? or just password with 2fa?

For code login I think it can make sense. For password reset it's not useful, but also not harmful IMO.

distantnative:
@lukasbestle I added it as todo, but maybe `['2fa' => 'optional'] mode` should be a separate PR as that change alone we need to review very closely. And TOTP support can be also added without that enhancement.

distantnative:
Switched it to Ready for review so code coverage tests run

config/areas/users/dialogs.php
```diff
@@ -336,4 +336,69 @@
...
+					$user->validatePassword($password);
+				}
+
+				// Remove the TOTP secret from the account
+				$user->changeTotp(null);
+
+				return [
+					'message' => I18n::translate('login.totp.disable.success')
+				];
+			} catch (Exception $e) {
```
lukasbestle: I think we should only catch the specific exception we need to rethrow. All other exceptions can be passed upwards 1:1.

config/areas/users/dialogs.php
```diff
@@ -336,4 +336,69 @@
...
+				// Catch and re-throw exceptions so that any
+				// Unauthenticated exception for incorrect passwords
+				// does not trigger a logout
+				throw new InvalidArgumentException($e->getMessage());
```
lukasbestle: Would be good to pass the `$e` as `$previous` to the new exception to ease debugging the exception chain.

src/Panel/User.php
```diff
@@ -73,19 +73,40 @@ public function dropdown(array $options = []): array
...
 		];
 
-		$result[] = [
-			'dialog'   => $url . '/changeLanguage',
-			'icon'     => 'translate',
-			'text'     => I18n::translate('user.changeLanguage'),
-			'disabled' => $this->isDisabledDropdownOption('changeLanguage', $options, $permissions)
-		];
+		if ($this->model->kirby()->system()->is2FAWithTOTP() === true) {
+			if ($account || $this->model->kirby()->user()->isAdmin()) {
```
lukasbestle, line 93: I think we can combine these two `if`s into one.

src/Panel/User.php
```diff
@@ -73,19 +73,40 @@ public function dropdown(array $options = []): array
...
+						'dialog'   => $url . '/totp/disable',
+						'icon'     => 'qr-code',
+						'text'     => I18n::translate('login.totp.disable.option'),
+					];
+				} elseif ($account) {
+					$result[] = [
+						'dialog'   => $url . '/totp/activate',
+						'icon'     => 'qr-code',
+						'text'     => I18n::translate('login.totp.activate.option'),
+						'disabled' => !$account
```
lukasbestle: Can this line ever take effect with the `elseif ($account)` above? Looks like a leftover.

config/areas/users/dialogs.php
```diff
@@ -336,4 +336,69 @@
...
+			$password = $kirby->request()->get('password');
+
+			try {
+				// Admins can disable TOTP without password
+				if (
+					$kirby->user()->isAdmin() === false ||
+					$kirby->user()->is($user) === true
+				) {
+					$user->validatePassword($password);
+				}
```
lukasbestle: I think it would be good to immediately block other non-admin users from calling this dialog submit route at all. With the current code there could be the case of another user (non-admin) who knows the password of another user but not the TOTP code – they could then call the route manually and provide the password.

I know there's another protection in `UserRules`, but another layer would IMO be good in such a critical part of the system.

So there should be three cases:

- Current user: Validate password (doesn't matter if admin or not)
- Admin: Don't validate password
- All other cases: Error

src/Panel/User.php
distantnative: Tbh, I think that just harms legibility in this case. I think it's good to read them as two distinct steps/checks.

lukasbestle: I think the following is just as readable but cleaner:

```php
if (
    $this->model->kirby()->system()->is2FAWithTOTP() === true &&
    ($account || $this->model->kirby()->user()->isAdmin())
) {
```

If you disagree, I'm fine with keeping it as it is.

config/areas/users/dialogs.php
distantnative: You mean just catching `InvalidArgumentException`? This will still catch several exceptions as only one is specifically sending a 401 code that we'll need to prevent

config/areas/users/dialogs.php
distantnative: @lukasbestle I don't quite understand our exception classes - how can I pass a message as well as previous?

config/areas/users/dialogs.php
distantnative: Does this solve what you mean?

```php
if ($kirby->user()->is($user) === true) {
	$user->validatePassword($password);
} else if ($kirby->user()->isAdmin() === false) {
	throw new PermissionException('You are not allowed to disable TOTP for other users');
}
```

config/areas/users/dialogs.php
lukasbestle: Yes, but IMO still better than catching every exception type. Especially because we re-throw a `InvalidArgumentException`, so if another exception type was to be thrown, the exception class would be converted incorrectly.

config/areas/users/dialogs.php
lukasbestle: This should do it:

```php
throw new InvalidArgumentException([
	'fallback' => $e->getMessage(),
	'previous' => $e
]);

config/areas/users/dialogs.php
lukasbestle: Yes, that looks great!

distantnative:
@lukasbestle I'd suggest to merge it and add the link to the docs later once they've been created

lukasbestle (approved):
I agree. Apart from that it's already done and can be tested during the beta. 🎉

I've made some changes during review and testing, hope that's OK. Could you please check them before you merge?

PR #5667, opened 2023-09-19
Query: `qr` entry function

distantnative:
### Enhancement
- Kirby queries now  support `qr()` entry function

## Ready?

### For review team

### DIFF
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -9,6 +9,7 @@ use Kirby\Cms\File;
 use Kirby\Cms\Page;
 use Kirby\Cms\Site;
 use Kirby\Cms\User;
+use Kirby\Image\QrCode;
 use Kirby\Toolkit\I18n;
 
 /**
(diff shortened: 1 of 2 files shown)


afbora:
LGTM 👍 But breaking changes didn't seem to me to be required. IMHO it would be better not to add breaking changes during the beta process unless it is required.

lukasbestle (commented):
Hm, couldn't this be quite dangerous? Because there is no allowlist, one could call any built-in PHP function.

distantnative:
@lukasbestle would it be better than to manually implement most/all Kirby helper functions as query entry functions?

distantnative:
As a side note: Kirby queries already allow calling any kind of method they can reach from their entry points.

distantnative:
Ok, I changed and limited it to just adding a `qr` entry function

lukasbestle (approved):

PR #5682, opened 2023-09-23
Fix sticky columns

### LINKED ISSUE #5664: [v4] Sticky column doesn't stop at end of column
## Description
CleanShot 2023-09-19 at 12 17 26@2x

```
    columns:
      - width: 1/4
        sticky: true
        fields:
          tagsHeadline:
            label: blabla
            type: headline
      - width: 3/4
        fields:
			[...]
      - width: 1 # here the column should stop being sticky
        fields:
          line1:
            type: line
      - width: 1/4
        fields:
          contentHeadline:
            label: blabla2
            type: headline
      - width: 3/4
        fields:
			[...]
```

## Your setup
**Kirby Version**  v4.0-beta.1

distantnative:
## This PR …

Wrapping child div needed so the sticky-ness can be contained

### Fixes
- #5664

### DIFF
--- a/panel/src/components/Layout/Column.vue
+++ b/panel/src/components/Layout/Column.vue
@@ -1,7 +1,10 @@
 <template>
 	<div :style="'--width:' + width" :data-sticky="sticky" class="k-column">
-		<!-- @slot Column content -->
-		<slot />
+		<!-- additional <div> needed to ensure sticky columns behave correctly -->
+		<div>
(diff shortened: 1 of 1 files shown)


bastianallgeier:
I kind of feel like we should only have that additional div if the column is actually sticky.

bastianallgeier (approved):
I've added an if clause and a sandbox example for a sticky setup

PR #5686, opened 2023-09-24
Streamline frame and tag usage

distantnative:
- `k-tag` supports an image/icon frame
- New `k-color-frame`
- Link field uses native `k-tag` image for preview
- Color field preview uses `k-tag` with `k-color-frame` in image slot


bastianallgeier:
Not a big fan of the image-icon frame name to be honest. What if we call it k-auto-frame. Then it could toggle between more preview types later if necessary.

bastianallgeier:
I would also handle more stuff like deprecating the bubbles in a different PR and focus on the color frame in this one.

distantnative:
Then it would need to support the color frame directly too. I'm afraid it would become too magical eventually. Image + icon is a use case we're using in many places, all other not really.

distantnative:
@bastianallgeier I've removed `k-image-icon-frame` again to not block this.

bastianallgeier (approved):

PR #5692, opened 2023-09-25
Various input fixes

bastianallgeier:
This PR is coming from the inputbox refactoring and moves over a couple of the fixes

## Fixes
- Fixed translation string for the blocks field
- The link field shows up correctly if no options are defined

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -277,7 +277,7 @@
 	"field.blocks.delete.confirm.all": "Do you really want to delete all blocks?",
 	"field.blocks.delete.confirm.selected": "Do you really want to delete the selected blocks?",
 	"field.blocks.empty": "No blocks yet",
-	"field.blocks.fieldsets.empy": "No fieldsets yet",
+	"field.blocks.fieldsets.empty": "No fieldsets yet",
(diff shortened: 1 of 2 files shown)


panel/src/components/Forms/Field/BlocksField.vue
```diff
@@ -108,14 +108,14 @@ export default {
 			return this.$helper.object.length(this.fieldsets) > 0;
 		},
 		isEmpty() {
-			return this.value.length === 0;
+			return this.value?.length === 0;
```
distantnative: Why is this needed? Like just curious to understand. Value has the array type and a default set. Why does it result in being null?

bastianallgeier: Thinking some more about it, I probably caused this unintentionally in the UI lab. The value is null there by default for all examples to achieve proper reactivity. But this is not really representative of real-world usage. It just simplifies the example setup. I will check this again.

bastianallgeier: Ok, it's a problem how the default value is set in the examples. I've reverted the changes and fixed the examples instead.

distantnative (approved):

PR #5696, opened 2023-09-25
k-coloroptions-input

bastianallgeier:
## Features

- New `k-coloroptions-input`

Screenshot 2023-09-25 at 16 46 43

The new input is used as foundation for the color field and the k-colorpicker-input, but can now also be used on its own. 

```html
<k-coloroptions-input :options="options" :value="value=" @input="value = $event" />
```

### DIFF
--- a/panel/src/components/Forms/Field/ColorField.vue
+++ b/panel/src/components/Forms/Field/ColorField.vue
@@ -1,22 +1,13 @@
 <template>
 	<k-field v-bind="$props" :input="_uid" class="k-color-field">
 		<!-- Mode: options -->
-		<div
+		<k-coloroptions-input
 			v-if="mode === 'options'"
-			style="--preview-width: var(--field-input-height)"
+			:options="convertedOptions"
+			:value="value"
 			class="k-color-field-options"
-		>
-			<button
-				v-for="color in convertedOptions"
-				:key="color.value"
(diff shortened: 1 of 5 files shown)


distantnative:
@bastianallgeier keyboard navigation seems broken, I can only reach the first or last option (depending if focus tabbing forward or backwards).

bastianallgeier:
It's a radio input. Once you set the focus, choice is made with arrow keys.

distantnative:
Ahh ok that wasn't quite that obvious. Though e.g. on https://sandbox.test/panel/pages/fields+color?tab=options navigating with arrow keys will jump between different fields.

distantnative:
It's due to the missing name. 
I think we once need to pass a name where possible but also fall back to a uuid if no name is passed.

bastianallgeier:
I've rebased it on v4/develop and added the _uid if there's no name

distantnative:
@bastianallgeier Should we pass the name in the colorfield context?

bastianallgeier:
@distantnative yes, definitely makes sense!

distantnative:
Shouldn't there be some visual indication when the input is focussed? That's currently missing (which then feels like the focus just disappeared).

bastianallgeier:
@distantnative ☑️

distantnative (approved):

PR #5697, opened 2023-09-26
Field preview refactoring

bastianallgeier:
I've created example setups for all field previews in the sandbox and found quite a few inconsistencies and smaller issues. 

## Enhancements

- Improved focus styles for links and the flag preview
- Improved text overflow behavior for links

## Refactoring

- `k-time-field-preview` now extends `k-date-field-preview` and improves time parsing and the default formatting
- The fieldpreview mixin defines proper defaults for column and field
- `k-toggle-field-preview` uses the low level `k-toggle-input` instead of `k-input` to avoid unnecessary markup

## Fixes

- `k-bubbles-field-preview` and all other previews that extend it now correctly display when there are no bubbles
- `k-color-field-preview` correctly displays the pattern when no color is set

### DIFF
--- a/panel/src/components/Forms/Previews/FlagFieldPreview.vue
+++ b/panel/src/components/Forms/Previews/FlagFieldPreview.vue
@@ -1,6 +1,6 @@
 <template>
-	<div class="k-flag-field-preview">
-		<k-button v-if="value" v-bind="status" size="md" />
+	<div v-if="value" class="k-flag-field-preview">
+		<k-button v-bind="status" size="md" />
 	</div>
 </template>
 
--- a/panel/src/components/Forms/Previews/FlagFieldPreview.vue
+++ b/panel/src/components/Forms/Previews/FlagFieldPreview.vue
@@ -1,5 +1,7 @@
 <template>
-	<k-button v-bind="status" class="k-flag-field-preview" />
+	<div class="k-flag-field-preview">
+		<k-button v-if="value" v-bind="status" size="md" />
+	</div>
 </template>
 
 <script>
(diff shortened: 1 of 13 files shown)


panel/src/components/Forms/Previews/FlagFieldPreview.vue
```diff
@@ -1,20 +1,21 @@
 <template>
-	<k-button v-bind="status" class="k-flag-field-preview" />
+	<div class="k-flag-field-preview">
+		<k-button v-if="value" v-bind="status" size="md" />
```
distantnative: Should the wrapping div remain if no value is set?

panel/src/components/Forms/Previews/ToggleFieldPreview.vue
```diff
@@ -1,34 +1,28 @@
...
-	props: {
-		field: Object,
-		value: Boolean,
-		column: Object
-	},
+	mixins: [FieldPreview],
 	computed: {
 		text() {
-			return this.column.text !== false ? this.field.text : null;
+			return this.column?.text !== false ? this.field?.text : null;
```
distantnative: Since you remove the `?` in other previews, why adding them here?

bastianallgeier: Makes sense to remove them here as well.

panel/src/components/Forms/Previews/FlagFieldPreview.vue
bastianallgeier: Much better to also remove the div.

distantnative (approved):

PR #5698, opened 2023-09-26
New k-timeoptions-input

bastianallgeier:
## Enhancements

- New `selected` prop for `k-button` to set the `aria-selected` attribute.

## Refactoring

- `k-timeoptions-input` replaces `k-times`. `k-times` is still available as deprecated alias.
- Date and Time fields use the new input

### DIFF
--- a/panel/src/components/Forms/Field/DateField.vue
+++ b/panel/src/components/Forms/Field/DateField.vue
@@ -65,7 +65,7 @@
 						@click="$refs.times.toggle()"
 					/>
 					<k-dropdown-content ref="times" align-x="end">
-						<k-times
+						<k-timeoptions-input
 							:display="time.display"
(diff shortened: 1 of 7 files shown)


distantnative:
@bastianallgeier the focus styling still seems to miss when focus is set programmatically via the method

bastianallgeier:
Hm, seems to be another Safari thing. The focus style works fine here – even when triggered via button

distantnative:
But I was using Chrome.
When tabbing it showed. When using the focus button in the sandbox not.

bastianallgeier:
Then it only works in Firefox. Maybe we can tackle this after beta 2?

distantnative:
If you want to merge it before, ok with me. it's "only" focus

PR #5699, opened 2023-09-26
New k-colorname-input

bastianallgeier:
## Features

- New `k-colorname-input`

The color name input enhances a regular text field to parse color values and return formatted color strings as value. 

## Enhancements

- The color field now also translates valid color names 

## Refactoring

- `k-color-field` now uses the new `k-colorname-input`


distantnative:
Color field is broken - black background and native color input dialog 

Screenshot 2023-09-26 at 19 42 28

Either StringInput must not bind type of the input or we need to explicitly pass type text from colornameinput.

bastianallgeier:
@distantnative just fixed it.

PR #5706, opened 2023-09-27
Fixes for various block issues

bastianallgeier:
While setting up the block examples, I found a couple issues in block components. 

## Fixes

- Added default values for object props to avoid breaks
- Fixed various inconsistencies in `k-block-title` styles
- Better defaults and removed outdated props in `k-block-figure`
- Fixed padding in the block header of the field block type component.

### DIFF
--- a/panel/src/components/Forms/Blocks/Types/Image.vue
+++ b/panel/src/components/Forms/Blocks/Types/Image.vue
@@ -44,7 +44,7 @@ export default {
 				return this.content.src;
 			}
 
-			if (this.content.image[0]?.url) {
+			if (this.content.image && this.content.image[0]?.url) {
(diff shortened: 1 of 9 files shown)


panel/src/components/Forms/Blocks/Types/Image.vue
```diff
@@ -44,7 +44,7 @@ export default {
 				return this.content.src;
 			}
 
-			if (this.content.image[0]?.url) {
+			if (this.content.image && this.content.image[0]?.url) {
```
distantnative:
```suggestion
			if (this.content.image?.[0]?.url) {
```

distantnative (approved):

PR #5722, opened 2023-10-03
Activation

### LINKED ISSUE #5716: [v4] Friendlier activation reminder

distantnative:
When menu is collapsed, show fixed and allow hiding for one sessionStorage length
Screenshot 2023-10-07 at 14 30 49

When menu is expanded, always show it
Screenshot 2023-10-07 at 14 31 11

### DIFF
--- a/panel/src/components/View/Activation.vue
+++ b/panel/src/components/View/Activation.vue
@@ -1,0 +1,75 @@
+<template>
+	<aside v-if="!$panel.license" :data-closed="!isOpen" class="k-activation">
+		<span class="k-text">Ready to launch?</span>
+
+		<k-button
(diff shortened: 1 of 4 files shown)


lukasbestle:
I like your new solution for the collapsed menu mode. 👍

An idea, but don't know if it's viable or useful: If the menu is collapsed and the activation box was closed, we could display a button in the menu (key icon, green background). This way, the activation would still be accessible even if the box was closed in the same session. It would also be a constant reminder for the unactivated state in the UI.

distantnative:
Better alternative IMO: https://github.com/getkirby/kirby/tree/v4/enhancement/activation-floating-bar

PR #5724, opened 2023-10-03
Implement A::every(), A::some(), A::find()

rasteiner:
### Features

- New `A::every()`, `A::find()` and `A::some()` methods that implement the functionality of the JavaScript functions with the same names

### Breaking changes
none

## Discussion
The functions mirror the JS ones:
1. in nomenclature, other languages use `any` and `all` instead of `some` and `every`. Idk what you prefer. 
2. in behaviour and function signature: falsy and truthy values are considered `false` and `true`, the callbacks all get `$value, $key, $array` as arguments. 

## Ready?
I've added `@since 3.9.8` tags to the functions, that needs to be corrected of course should the changes be merged later. 

### For review team

### DIFF
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -60,6 +60,76 @@ class A
 	}
 
 	/**
+	 * Checks if every element in the array passes the test
+	 *
+	 * <code>
+	 * $array = [1, 30, 39, 29, 10, 13];
+	 *
+	 * $isBelowThreshold = fn($value) => $value < 40;
+	 * echo A::every($array, $isBelowThreshold) ? 'true' : 'false';
+	 * // output: 'true'
+	 *
+	 * $isIntegerKey = fn($value, $key) => is_int($key);
+	 * echo A::every($array, $isStringKey) ? 'true' : 'false';
+	 * // output: 'true'
+	 * </code>
+	 *
+	 * @param array $array
+	 * @param callable(mixed $value, int|string $key, array $array):bool $test
+	 * @return bool
+	 */
+	public static function every(array $array, callable $test): bool
+	{
+		foreach ($array as $key => $value) {
+			if (!$test($value, $key, $array)) {
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	/**
+	 * Finds the first element matching the given callback
+	 *
+	 * <code>
+	 * $array = [1, 30, 39, 29, 10, 13];
+	 *
+	 * $isAboveThreshold = fn($value) => $value > 30;
+	 * echo A::find($array, $isAboveThreshold);
+	 * // output: '39'
+	 *
+	 * $array = [
+	 *   'cat' => 'miao',
+	 *   'cow' => 'moo',
+	 *   'colibri' => 'humm',
+	 *   'dog' => 'wuff',
+	 *   'chicken' => 'cluck',
+	 *   'bird' => 'tweet'
+	 * ];
+	 *
+	 * $keyNotStartingWithC = fn($value, $key) => $key[0] !== 'c';
+	 * echo A::find($array, $keyNotStartingWithC);
+	 * // output: 'wuff'
+	 * </code>
+	 *
+	 * @since 3.9.8
+	 * @param array $array
+	 * @param callable(mixed $value, int|string $key, array $array):bool $callback
+	 * @return mixed
+	 */
+	public static function find(array $array, callable $callback): mixed
+	{
+		foreach ($array as $key => $value) {
+			if ($callback($value, $key, $array)) {
+				return $value;
+			}
+		}
+
+		return null;
+	}
+
+	/**
 	 * Gets an element of an array by key
 	 *
 	 * <code>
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
Thanks for your contribution and your super well-documented and well-tested code.

The methods look good to me. I think it makes sense to use the JS naming as it fits quite well to use another web language/standard library as a template. The callback signature is fine for me.

What I'm not 100% sure about is if we should add it to the core as the methods reduce the overview and add (even if little) complexity. Can you share what you are using them for?

rasteiner:
Really anywhere I'd like to "filter with early exit", especially in places where an expression is expected, without wanting the structural overhead of "for each...if...assign to variable...break". I feel like it's the same kind of terseness the Php 8 "match" expression gives you. 

Imo those, together with map, filter, reduce, are really the basic building blocks for arrays. And I have been reaching for them in different occasions, when not finding them in the docs, I instinctively then expect "::has" to accept a callback (would be "some", or the inverse of "every"), or "::get()" to accept a callback (would be "find").

The last time I had the instinct to use something like "some" or "every" was this morning when refactoring my "panel.menu" config to something like this:

```php
// ...
'site' => [
  // ...
  'current' => fn($id) => 
    $id === 'site' && 
    !A::some($pageAreas, fn($url) => Str::startsWith($kirby->request()->url(), $url))
]
```
Without A::every or A::some, this looks like:

```php
// ...
'site' => [
  // ...
  'current' => function($id) use($pageAreas, $kirby) {
    if($id !== 'site') return false;

    foreach($pageAreas as $url) {
      if(Str::startsWith($kirby->request()->url(), $url)) {
        return false;
      }
    }

    return true;
  }
]
```

(sorry for the probable errors in code, I'm writing this from memory on my phone). 

I know that's only an opinion, but I find the phrasing
> "Is current when none of page area urls are the start of the request url" 

to express the intent more explicitly than the imperative approach of

> "Is current when for each page area if its url is the start of the request url then return false, then return true"

If you find it easier, "every" then also gives you the option to express the idea as

> is current when every page area url is not the start of the request url

... not sure I'm getting the point across, it's just much more expressive imo

rasteiner:
As well as.. you know, that thing about higher order functions helping with composability.  
It's an advantage if I can define an idea as a function and then get different behaviors out of it by passing that same function to either "filter", "some", "every" or "find".

```php
function isAnimal($name) {
  // implementation 
}

$firstAnimal = A::find($array, isAnimal(...));
$hasAnimals = A::some($array, isAnimal(...));
$hasOnlyAnimals = A::every($array, isAnimal(...));
$allAnimals = A::filter($array, isAnimal(...));
```

nilshoerrmann:
> What I'm not 100% sure about is if we should add it to the core as the methods reduce the overview and add (even if little) complexity.

From a user's point of view, these base method would be very welcome :+1:
`some` and `every`could be very helpful on collections as well, not just arrays …

bastianallgeier:
I like them as well and I think they totally deserve a place in the core.

rasteiner:
> `some` and `every` could be very helpful on collections as well, not just arrays …

I agree, at least for collections we get "collection methods" as a plugin option though, there's nothing for `A`, and a plugin would need to have its own class (like `Arr`, and users then need to remember when to use `A` and when `Arr`). 

**Unrelated:**
A future step, or if you prefer I can include this here, would be changing `A::filter` so that its callback also gets the array as third argument, for consistency and composability. Sending the array to the callback allows you to extract logic even more as you no longer need to declare the predicate in the same place you use it (helps with the DRY and unit tests).

Stupid example:

```php
$array = [
  [ 'type' => 'dog',  'name' => 'Rex' ],
  [ 'type' => 'cat',  'name' => 'Mittens' ],
  [ 'type' => 'dog',  'name' => 'Fido' ],
  [ 'type' => 'cat',  'name' => 'Whiskers' ],
  [ 'type' => 'dino', 'name' => 'Rex' ],
];

function by($key, $value) {
  return fn($v) => $v[$key] === $value;
}

function isFirstOf($prop) {
  return fn($value, $key, $array) => A::find($array, by($prop, $value[$prop])) === $value;
}

// keep only cats
$allCats = A::filter($array, by('type', 'cat'));

// keep only the first of each type
$uniqueByType = A::filter($array, isFirstOf('type'));

// keep only the first of each name
$uniqueByName = A::filter($array, isFirstOf('name'));

```

lukasbestle (approved):

lukasbestle:
> A future step, or if you prefer I can include this here, would be changing `A::filter` so that its callback also gets the array as third argument, for consistency and composability.

I very much like the idea. Wouldn't it mean though that we couldn't use the native PHP `array_filter()` function anymore? As it's native code, it will likely have a better performance than user-space code.

In any case, I think we should tackle that in a separate PR.

rasteiner:
Yes, it would probably be a bit slower. One would need to test that though: I'd guess that the most time with array_filter is spent pushing memory around to allocate the stacks for the function calls and running the php code of the callbacks, so array_filter never had good performance when compared to an inline foreach. 

I won't be seeing computers today, but I would guess that a benchmark could look like this:

```
Inline foreach: ---------- 100%
array_filter:   -------------------- 200%
new a::filter:  --------------------- 205%
```

(With the relative difference decreasing with larger callback functions).

So, in the larger picture I don't think it would represent a problem. Performance critical code shouldn't use the functional style anyway.

A special case could probably be array_filter when called without callback (removes empty elements), but we could check for that and eventually redirect to native array_filter in that case.

PR #5733, opened 2023-10-06
Fix custom writer marks and nodes name

### LINKED ISSUE #5731: [v4] Selecting spesific custom `writerMarks` doesn't work
I'm not sure this is a bug or I'm missing some setup. Selecting spesific custom `writerMarks` doesn't work for writer field.

### Plugin

````js
panel.plugin("my/writer-mark", {
  writerMarks: {
    smile: {
      get button() {
        return {
          icon: "smile",
          label: "Smile"
        }
      }
    }
  }
});
````

### Blueprint
````yaml
fields:
  writer:
    type: writer
    marks:
      - bold
      - smile
    toolbar:
      inline: false
      marks:
        - bold
        - smile
````

### Result
screen-capture-1691-Test - Mægazine-localhost

## Additional Context

I can see the button when `marks: true`

````yaml
fields:
  writer:
    type: writer
    marks: true
    toolbar:
      inline: false
````

screen-capture-1692-Test - Mægazine-localhost

Kirby 4.0.0-beta.2

### PR DESCRIPTION
**Pull request by afbora**
### Fixes
- Fix custom writer marks and nodes name #5731

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Writer/Writer.vue
+++ b/panel/src/components/Forms/Writer/Writer.vue
@@ -252,12 +252,12 @@ export default {
 			const plugins = window.panel.plugins.writerMarks ?? {};
 			const marks = {};
 
-			// take each extenstion object and turn
+			// take each extension object and turn
 			// it into an instance that extends the Mark class
 			for (const name in plugins) {
 				marks[name] = Object.create(
 					Mark.prototype,
-					Object.getOwnPropertyDescriptors(plugins[name])
+					Object.getOwnPropertyDescriptors({name, ...plugins[name]})
 				);
 			}
 
@@ -305,12 +305,12 @@ export default {
 			const plugins = window.panel.plugins.writerNodes ?? {};
 			const nodes = {};
 
-			// take each extenstion object and turn
+			// take each extension object and turn
 			// it into an instance that extends the Node class
 			for (const name in plugins) {
 				nodes[name] = Object.create(
 					Node.prototype,
-					Object.getOwnPropertyDescriptors(plugins[name])
+					Object.getOwnPropertyDescriptors({name, ...plugins[name]})
 				);
 			}
 

### CODE REVIEW

File: panel/.eslintrc.js [hunk 1]
```diff
@@ -1,5 +1,8 @@
 module.exports = {
 	extends: ["eslint:recommended", "plugin:vue/recommended", "prettier"],
+	parserOptions: {
+		ecmaVersion: 2021
+	},
```
afbora: Not sure about this change. I have to use version 2021 (currently 2020) to use null coalescing assignment operator `??=` in js.

bastianallgeier: If this does not break anything else, I'm happy about it. We could definitely use this in more places.

distantnative: Is this automatically compiling backwards on build - or are we raising browser requirements with this?

panel/src/components/Forms/Writer/Writer.vue
```diff
@@ -252,12 +252,15 @@ export default {
...
+			// take each extension object and turn
 			// it into an instance that extends the Mark class
 			for (const name in plugins) {
+				const plugin = plugins[name];
+				plugin.name ??= name;
+
 				marks[name] = Object.create(
 					Mark.prototype,
-					Object.getOwnPropertyDescriptors(plugins[name])
+					Object.getOwnPropertyDescriptors(plugin)
```
distantnative: This is actually exactly what I had in mind, just didn't realize this is already the right spot when I wrote from my phone.

Might have written is just as 

```suggestion
					Object.getOwnPropertyDescriptors({ name, ...plugins[name] })
```

to make it a one-liner change

afbora: That's what I need 💯

distantnative (approved):

PR #5742, opened 2023-10-07
Improve tags selector

### LINKED ISSUE #5723: Tags field should hide "Add" button when disabled
## Description
The tags field has the ability to query tags from other fields or pages and allow the user to select from those values. When adding tags on the fly is disabled, the Add button doesn't disappear which is confusing.

**Expected behavior**  
When using the tags field with `accept: options`, the "Add" button and also the "Create" button should disappear.

**Screenshots**  
Screenshot 2023-10-03 at 11 06 15
Screenshot 2023-10-03 at 11 15 48

Add a tag field using this code:

````yaml
years:
  type: tags
  accept: options
  options: query
    type: query
    query: page('people').yeartags.split # replace for a working query
````

## Your setup
Kirby 4 Beta 1

distantnative:
I think the tags selector is still a quite messy UX. Trying to improve it step by step. Two first improvements

- Remove the label on top of the selector. IMO it's more confusing than providing guidance for users. Removing it does improve the understanding and UX for me.
- Don't show "no options" when query doesn't show any matches and creating a new option is allowed. The create button already provides enough context of what action is available. For `accept: options` keep the empty text to give context what's happening.

### Fixes
- https://github.com/getkirby/kirby/issues/5723

### DIFF
--- a/panel/src/components/Forms/Selector.vue
+++ b/panel/src/components/Forms/Selector.vue
@@ -5,9 +5,6 @@
 		:data-has-current="filtered?.includes(selected)"
 	>
 		<header class="k-selector-header">
-			<h2 v-if="label" class="k-selector-label">
-				{{ label }}
-			</h2>
 			<div v-if="showSearch" class="k-selector-search">
 				<input
 					ref="input"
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
I think you are right. I really like the changes.

Issue #5319, opened 2023-06-21
Old session token keys should (temporarily) be able to read and write to new session data


s3ththompson:
## Description
I have a Kirby site that lazily loads two session-based pieces of content after initial page load: error messages (flashed to the session during the previous request) and a shopping cart (initialized with session data). For reasons that will become important later, both pieces of content need to read AND write session data to load.

Lazy loading this information also allows me to ensure that the rest of my Kirby pages don't use auth or session data, so they can be cached for all users.

In the wild, one or the other of these requests often return a 500 error: 

> Session "1687782686.0fba42e0ea550758ac47" is currently read-only because it was accessed via an old session

In [this very helpful comment](https://github.com/getkirby/kirby/issues/2505#issuecomment-596743043) on #2505, @lukasbestle explains why:

> Sessions are regularly renewed for security. Each time a session is renewed, the session token (= the session cookie) changes and is retransmitted to the browser. So in normal operation the browser will have the new session cookie already for the next requests and will use that – which will work without issues. If the API still receives requests with the old token however, it will internally use the new session data for a limited time of 30 seconds, but in a read-only mode. If the API request tries to modify any data (like when trying to log in), the error is thrown.

Here is a timeline of what happens:

<img width="702" alt="waterfall" src="https://github.com/getkirby/kirby/assets/970121/301fcd56-26ca-4bce-9b26-a866857ce22a">

0. [Before page load] Current session times out (but has not yet expired)
1. Initial page load (`/events`) (from pages cache, does not use auth or session data)
2. Two parallel requests (`/cart` and `/flash`) are sent for lazy-loaded content (both attempt to read/write session data using old token key)
3. `/flash` successfully returns response (after attempting to write to the timed out session with an old token key, Kirby successfully renewed session, returning new token key with this response)
4. `/cart` returns above 500 error (after attempting to write to the new session using old token key) 

**Expected behavior**  

Both parallel requests (`/cart` and `/flash`) succeed, because Kirby's 30 second grace period (see [above comment](https://github.com/getkirby/kirby/issues/2505#issuecomment-596743043)) for accessing sessions with old token keys should be extended to support reading *and* writing data to the new session.

As far as I can tell from @lukasbestle's comment, the main reason that the existing feature only supports read-only session access during the 30 second grace period is simply because most Panel access is read-only after login.

## Your setup
**Kirby Version**  
3.9.5

**Your system (please complete the following information)**
 - Device: Server
 - OS: Linux 5.15.0
[...]

PR #5750, opened 2023-10-07
Session: Allow write access to moved sessions

lukasbestle:
To explain why it we had the exception before and why encryption is necessary to avoid the exception:

- Our session implementation is designed so that attackers with read access to the session files (e.g. when HTTP access to `site/sessions` isn't blocked and directory indexes are enabled) cannot impersonate a session and therefore a logged in user.
- To make this work, each session has a token key that is included in the session cookie but never stored anywhere. Only requests that contain the valid token key can use the session.
- If a session token is regenerated, the old session stays valid temporarily (for 30 seconds). This is done by replacing the old session with a reference to the new one. But we cannot store the token key of the new session in the old session, otherwise the security requirement of the first point is broken.
- Before this PR, the session implementation circumvented this problem by allowing to access the new session without key when (and only when) it was accessed through the old session with a valid token key. However a write access to the new session was not possible without knowing the new key because writing to a session requires the key to recalculate the HMAC that is stored to verify the key later.
- This PR fixes this issue by storing the encrypted token key of the new session in the old session. Only the actual user (who knows the old token key) can decrypt it, so the attack scenario of the first point is still taken care of.
- Encryption is done with the PHP `sodium` extension, which provides secure defaults and is therefore a lot more solid than OpenSSL. To avoid having to add `sodium` to Kirby's list of required extensions, the key storage is optional (only if `sodium` is available). Otherwise the session class behaves exactly like before.

---

### Features

- New `Kirby\Toolkit\SymmetricCrypto` class (user-friendly and safe abstraction for symmetrical authenticated encryption using the PHP `sodium` extension)
  ```php
  // encryption/decryption with a password
  $crypto     = new SymmetricCrypto(password: 'super secure');
  $ciphertext = $crypto->encrypt('a very confidential string');
  $plaintext  = $crypto->decrypt($ciphertext);

  // encryption with a random key
  $crypto     = new SymmetricCrypto();
  $ciphertext = $crypto->encrypt('a very confidential string');
  $secretKey  = $crypto->secretKey();

  // encryption/decryption with a previously generated key
  $crypto     = new SymmetricCrypto(secretKey: $secretKey);
  $ciphertext = $crypto->encrypt('a very confidential string');
  $plaintext  = $crypto->decrypt($ciphertext);
  ```

### Enhancements

- The "Session ... is currently read-only because it was accessed via an old session" error is circumvented when the PHP `sodium` extension is available #5319

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,6 +57,7 @@
 		"ext-fileinfo": "Improved mime type detection for files",
 		"ext-intl": "Improved i18n number formatting",
 		"ext-memcached": "Support for the Memcached cache driver",
+		"ext-sodium": "Support for the crypto class and more robust session handling",
 		"ext-zip": "Support for ZIP archive file functions",
 		"ext-zlib": "Sanitization and validation for svgz files"
 	},
--- a/src/Session/Session.php
+++ b/src/Session/Session.php
@@ -578,7 +578,7 @@ class Session
 			$this->tokenKey === null ||
 			SymmetricCrypto::isAvailable() === false
 		) {
-			return null;
+			return null; // @codeCoverageIgnore
 		}
 
 		return new SymmetricCrypto(secretKey: hex2bin($this->tokenKey));
--- a/src/Session/Session.php
+++ b/src/Session/Session.php
@@ -10,6 +10,7 @@ use Kirby\Exception\NotFoundException;
 use Kirby\Http\Cookie;
 use Kirby\Http\Url;
 use Kirby\Toolkit\Str;
+use Kirby\Toolkit\SymmetricCrypto;
 use Throwable;
 
 /**
@@ -38,7 +39,7 @@ class Session
 	protected $lastActivity;
 	protected $renewable;
 	protected $data;
-	protected $newSession;
+	protected array|null $newSession;
 
 	// temporary state flags
 	protected $updatingLastActivity = false;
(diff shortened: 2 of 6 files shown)


distantnative:
@lukasbestle Do you have any insight in how commonly the "sodium" extension is pre-installed? Otherwise maybe this is an elaborate fix, but that might not solve the issue for that many people?

lukasbestle:
PHP ships with `sodium` since 7.2. PHP can be compiled without it, but I'd expect most hosting providers to include it. So it's certainly not one of those extensions one would have to explicitly install via PECL.

distantnative:
Ok great 👍

Can I suggest to move some of the code in separate methods? I find the session classes quite hard to understand because a lot is happening in some of the methods which my head needs to keep track of. I'm happy to suggest some changes for the existing classes to make them a bit easier to understand for those of us that have not designed them. But would be happy if this PR could maybe not add to it.

lukasbestle:
I agree it makes sense to refactor the class, however if we do it in this PR, we will have pretty drastic merge conflicts when we merge back into v4 because `v4/develop` already contains refactorings. So I suggest we refactor separately in v4.

distantnative:
@lukasbestle no not refactoring the classes overall on this PR - but don't you think those new code blocks would make sense in separate methods instead of adding them to existing already-long methods?

lukasbestle:
I have another idea, wait a moment.

bastianallgeier:
It looks good to me. @distantnative do you want to have a last look?

distantnative:
To be honest, the real logic goes beyond my understanding, but I do like the separation of code now a lot :)

bastianallgeier (approved):

s3ththompson:
Thank you for the extensive fix! I'm very grateful to see a fix to #5319.

Issue #5629, opened 2023-09-14
[v4] multiselect field overflow issue on mobile


afbora:
![image](https://github.com/getkirby/kirby/assets/3393422/3f86b4bd-a866-4089-8f86-c59acfe338b1)

## Your setup
4.0.0-beta.1
Chrome 116

Issue #5627, opened 2023-09-14
[v4] Multiselect field should indicate that the user can scroll


tobimori:
## Description
This is a multi select field that has so many entries that you can scroll.
![CleanShot 2023-09-14 at 12 29 18@2x](https://github.com/getkirby/kirby/assets/29142128/4b6612ab-bd80-448c-8948-3835323fc4c3)

However, on first glance, it doesn't look like you can scroll.

A better solution would be to cut the 11th item like so:
![CleanShot 2023-09-14 at 12 31 33@2x](https://github.com/getkirby/kirby/assets/29142128/22152f68-28a1-4d9e-b6cb-f5b72f9b5797)

Using v4 beta.1

afbora:
@tobimori Which browser do you use?

tobimori:
Chrome 116

afbora:
For me Chrome 116 🤔 

![image (27)](https://github.com/getkirby/kirby/assets/3393422/232379e5-22fc-4baa-ac22-69367edf83ff)

tobimori:
yes, that's the current behaviour. the second screenshot is imo the behaviour how it should be so the user knows they can scroll. there is no scroll bar on macOS.

afbora:
My personal humble opinion; I don't know how logical it is to cut the last item when there already seems to be a scroll bar.

tobimori:
The scrollbar is only visible on Windows devices (and maybe Linux) phones usually don't have a scrollbar as well as macOS.

lukasbestle:
I think making the list box slightly larger to cut off the 11th item is a good cross-browser and cross-OS solution. It doesn't break anything on Windows while it really helps on macOS.

luxuryluke:
Love this solution.

PR #5752, opened 2023-10-07
Multiselect input: UX improvements

### LINKED ISSUE #5744: [v4] Multiselect field: don't hide selected values, but highlight them
The UX of the multi select field got worse with v4 - as it is currently simply a tags field in a way.
However, a multiselect field should always show all options in its dropdown, never filter them. And instead should show those options that are selected with a highlighted styling.

Screenshot 2023-10-07 at 16 19 20

distantnative:
@bastianallgeier @afbora 
The code is still shit - just a rough implementation. But would be great to get your feedback on the UX. I think we should first get that right, agree on something and then I can work on implementing it in a clean way in the components.

Screenshot 2023-10-16 at 20 31 25

- Shows selected state in the dropdown
- Clicking on already selected options will remove them (unselect them)
- Dropdown fits to width of input


bastianallgeier:
Just a quick thought before I have a chance to try it. What if we use the checkboxes and radio inputs in the dropdown? This could even introduce the info line and we would be able to offload a lot of code to those components.

afbora:
@distantnative Maybe you want to take a look these issues before refactoring the field:
- https://github.com/getkirby/kirby/issues/5629
- https://github.com/getkirby/kirby/issues/5627

distantnative:
@bastianallgeier would that work for the tags input context as well?
@afbora I think we need to tackle those at the end, first find a UX that works

bastianallgeier:
@distantnative yes, I think so and it would semantically also be correct.

distantnative:
Closing in favor of https://github.com/getkirby/kirby/pull/5822

PR #5753, opened 2023-10-07
`k-navigate` as Vue component

distantnative:


bastianallgeier (commented):
I wonder what would happen if we turn this into a renderless component.

distantnative:
@bastianallgeier didn't the that to work as due wasnt  happy to have multiple elements out of the render  function, so needed to wrap with a div either way But maybe you know a way.

bastianallgeier:
That makes sense!

bastianallgeier (approved):

PR #5757, opened 2023-10-08
Fix search type permissions

distantnative:
### Fixes
- Search now takes access permissions into account what types can be shown

## Ready?

### For review team

### DIFF
--- a/config/areas/search/views.php
+++ b/config/areas/search/views.php
@@ -9,7 +9,7 @@ return [
 			return [
 				'component' => 'k-search-view',
 				'props' => [
-					'type' => App::instance()->request()->get('type', 'pages'),
+					'type' => App::instance()->request()->get('type'),
 				]
 			];
 		}
--- a/config/areas/search/views.php
+++ b/config/areas/search/views.php
@@ -9,7 +9,7 @@ return [
 			return [
 				'component' => 'k-search-view',
 				'props' => [
-					'type' => App::instance()->request()->get('type') ?? 'pages',
+					'type' => App::instance()->request()->get('type', 'pages'),
 				]
 			];
 		}
--- a/config/areas/search/views.php
+++ b/config/areas/search/views.php
@@ -1,6 +1,8 @@
 <?php
 
 use Kirby\Cms\App;
+use Kirby\Panel\Panel;
+use Kirby\Panel\View;
 
 return [
 	'search' => [
--- a/panel/src/components/View/Menu.vue
+++ b/panel/src/components/View/Menu.vue
@@ -9,6 +9,7 @@
 		<div class="k-panel-menu-body">
 			<!-- Search button -->
 			<k-button
+				:disabled="!hasSearch"
 				:text="$t('search')"
 				icon="search"
 				class="k-panel-menu-search k-panel-menu-button"
@@ -55,6 +56,9 @@ export default {
 		};
 	},
 	computed: {
+		hasSearch() {
+			return this.$helper.object.length(this.$panel.searches) > 0;
+		},
 		menus() {
 			return this.$panel.menu.entries.split("-");
 		}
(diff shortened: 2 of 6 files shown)


afbora:
I can still see the search icon in sidebar when I disabled for all access. Also I see when I clicked to search icon:

````yaml
permissions:
  access:
    site: false
    users: false
    todos: false
````

screen-capture-1693-System - Mægazine-localhost


### Console 
> TypeError: Cannot read properties of undefined (reading 'icon')

**Comment by afbora** [comment]
Now working correct 👍  

One more last minor thing. Disable all types and visit `/panel/search`. You'll see empty page with following console error:

> TypeError: Cannot read properties of undefined (reading 'id')

I know that user can not access to `search` view. But I feel like we should disable search view something like that https://github.com/getkirby/kirby/pull/5613/files#diff-4eafe8f0737452cffaace81565ddb4cdcf8cd84dfbebf63b39e19346a850b971R10-R16

### Disabling search view
screen-capture-1694-Error - Mægazine-localhost

What do you think?

**Comment by bastianallgeier** [comment]
I think it's a good idea to disable the search view entirely in that case. @afbora do you want to give it a shot? Or should we wait for @distantnative

**Comment by afbora** [comment]
@bastianallgeier I've implemented the disabling entirely search view. I hope it's acceptable for you and Nico.

File: config/areas/search/views.php [hunk 1]
```diff
@@ -1,10 +1,19 @@
...
 
 return [
 	'search' => [
 		'pattern' => 'search',
+		'when'    => function (): bool {
+			// checks core and custom search areas
+			$permissions = App::instance()->user()?->role()->permissions()->toArray() ?? [];
+			$searches    = View::searches(Panel::areas(), $permissions);
+
+			return empty($searches) === false;
```
distantnative: @bastianallgeier @afbora I am just a bit concerned how this performance-wise adds to each load. Having another place where all the Panel areas, searches and permissions need to be gathered.

bastianallgeier: Damn, you are right. I didn't even think about that to be honest.

afbora: I ask because I don't know: Aren't searches, views, areas and permissions already handles/processed and cached for a user accessing the panel?

bastianallgeier: No, they aren't. They are loaded with every view reload.

afbora: No, I didn't mean that. For example, before the lines in this PR run, aren't the permissions already processed and cached? If the methods called in this PR lines are not called before, you are right.

bastianallgeier: Yes, permissions should be cached. But searches and areas aren't

afbora: OK, so what's your suggestion to disable entirely search view? Or go without disabling like preview state of the PR?

bastianallgeier: For now, I would not work with the when condition here. As a long-term project, we should refactor our Panel classes though to not use static methods and keep all of this in a more reusable way.

afbora: OK then the PR is ready to merge after revert this lines 👍

afbora (approved):

panel/src/components/View/Menu.vue
```diff
@@ -9,6 +9,7 @@
 		<div class="k-panel-menu-body">
 			<!-- Search button -->
 			<k-button
+				:disabled="!hasSearch"
```
bastianallgeier: Wouldn't it be better to completely remove the button if there's no search?

afbora: When completely removed, the main menu comes to top. I think that's why Nico left it disable instead of remove.

bastianallgeier: @distantnative does that lead to a weird ui issue? Otherwise it makes a lot of sense if it moves up

distantnative: I found it looked a bit odd so close to the top and not aligned with the header headline. But won't object if you think differently.

bastianallgeier: I checked again and I think it's better if the menu moves up. I've changed it.

PR #5758, opened 2023-10-08
Parsley inline: fix trimming spaces-only elements

### LINKED ISSUE #4702: Blocks: pasting formatting lists removes spacing
## Description
When you try to paste/import some text from your clipboard that has lists that are formatted, the text of the list is rendered without the spaces around the formatted text.

**Expected behavior**  
The text of the list should be kept exactly as copied

**Screenshots**  

https://user-images.githubusercontent.com/39380741/192341077-1ecec363-23d4-431b-8b49-0cf7001d9f50.mp4

## To reproduce

1. Go on panel to a page that has blocks
2. Add some test list with list block
3. Add formatting to some words like bold, link
4. Save and open page
5. Copy the frontend rendered list
6. ctrl+v or paste on the blocks again
7. You should see that the same text appear without spaces around the words you've formatted

## Your setup
**Kirby Version**  
3.7.5

**Console output**  
nothing on console

 - Device: Laptop
 - OS: Windows 11
 - Browser: Edge
 - Version: 105.0.1343.50

**Additional context**  
It seems that if I add a list block then paste the text inside it, it will render as it should. Only when trying the ctrl+v function seems to not work.

distantnative:
### Fixes
- Blocks field: pasting HTML does not remove crucial spaces anymore in inline contexts
#4702

## Ready?

### For review team

### DIFF
--- a/src/Parsley/Inline.php
+++ b/src/Parsley/Inline.php
@@ -28,7 +28,15 @@ class Inline
 	public function __construct(DOMNode $node, array $marks = [])
 	{
 		$this->createMarkRules($marks);
-		$this->html = trim(static::parseNode($node, $this->marks) ?? '');
+
+		$html = static::parseNode($node, $this->marks) ?? '';
+
+		// only trim HTML if it doesn't consist of only spaces
+		if (strlen($html) !== substr_count($html, ' ')) {
+			$html = trim($html);
+		}
+
+		$this->html = $html;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/Parsley/Inline.php
```diff
@@ -28,7 +28,15 @@ class Inline
 	public function __construct(DOMNode $node, array $marks = [])
 	{
 		$this->createMarkRules($marks);
-		$this->html = trim(static::parseNode($node, $this->marks) ?? '');
+
+		$html = static::parseNode($node, $this->marks) ?? '';
+
+		// only trim HTML if it doesn't consist of only spaces
+		if (strlen($html) !== substr_count($html, ' ')) {
```
lukasbestle: I think this would be safer as it also works with other kinds of whitespace:

```suggestion
		if (trim($html) !== '') {
```

bastianallgeier (approved):

PR #5759, opened 2023-10-08
Don’t override permissions via any roles’ options

### LINKED ISSUE #4966: Setting permissions for a certain role via the page option removes permissions for all other roles
## Description
I have the following permission setup:

user blueprint for admin:
```
permissions:
  pages:
    update: true
```

user blueprint for editor:
```
permissions:
  pages:
    update: false
```

page blueprint:
```
options:
  update:
    editor: true
```

This results in the admin role (and all others roles) to have no update permission. When all I wanted to do was give the editor something extra for this page.

**Expected behavior**  
I would expect the admin role to still have the update permissions set in the roles blueprint, because I did not change the admin permission in the page blueprint.

If I WANTED to block all other permissions I would use this page blueprint:
```
options:
  update:
    *: false
    editor: true
```

## Your setup
3.8.3

distantnative:
### Fixes
- Options, e.g. page `options`, won't override other roles' permissions anymore 
 #4966

e.g.

```yml
options:
  update:
    editor: true
```

Would set `update` to false for any other roles, even if their permissions would allow it.

Before:
- check entry for role's name
- check `*` entry
- deny, no matter if permissions generally allow

With this PR
- check entry for role's name
- check `*` entry
- whether permissions generally allow

## Ready?

### For review team

### DIFF
--- a/src/Cms/ModelPermissions.php
+++ b/src/Cms/ModelPermissions.php
@@ -85,7 +85,13 @@ abstract class ModelPermissions
 				is_array($options) === true &&
 				A::isAssociative($options) === true
 			) {
-				return $options[$role] ?? $options['*'] ?? false;
+				if (isset($options[$role]) === true) {
+					return $options[$role];
+				}
+
+				if (isset($options['*']) === true) {
+					return $options['*'];
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


lukasbestle (approved):
Makes sense to me. I think the fallback to `false` came from a safety standpoint (rather prevent an action if it isn't explicitly allowed). But this only makes sense for permissions, not their override in options.

I think we should mention this as a breaking change though.

PR #5760, opened 2023-10-08
Consistent file sanitize methods

distantnative:


bastianallgeier (approved):
I like the separation here.

PR #5761, opened 2023-10-08
Fields w/ options: fix too strict sanitizing

### LINKED ISSUE #5013: select field with structure query does not save passed value
## description
i have a `select` field which queries a `structure` field from another page. this works fine. because we don’t have structure object uuids yet, it only saves the regular id _(= index)_ of the structure object in the content txt file. but that’s fine for me.

now i want to manually update the select field value, without using the panel, but through code. but no matter what value i pass, it just doesn’t get saved.

**the select field:**
```yml
select:
    type: select
    options:
        type: query
        query: site.structure.toStructure
        text: '{{ item.name }}'
```

**the update code:**
```php
$page->update([ 'select' => 0 ]);
```

if i set the field type to `multiselect`, `text` or anything else, the value gets saved in the content txt file perfectly fine. but as soon as i use `select`, it stops working. 

if i switch the select field to the regular options with hard-coded values and then pass one of them, it does get saved as well. so i bet there is some internal validation which checks if the select option actually exists. but with dynamic queries it doesn’t work.

i don’t exactly know what the [$validate](https://getkirby.com/docs/reference/objects/cms/page/update) attribute is for, but it is set to `false` by default. i thought having this off, would mean no validation is done on kirby’s side?

## expected behavior
the passed value will be saved in the content.txt file. simple as that.

## your setup
kirby `3.9`

🙏🏻 ✨

distantnative:
### Fixes
- Select options with integer values work properly now
#5013

## Ready?

### For review team

### DIFF
--- a/config/fields/mixins/options.php
+++ b/config/fields/mixins/options.php
@@ -36,7 +36,7 @@ return [
 		},
 		'sanitizeOption' => function ($value) {
 			$options = array_column($this->options(), 'value');
-			return in_array($value, $options, true) === true ? $value : null;
+			return in_array($value, $options) === true ? $value : null;
 		},
 		'sanitizeOptions' => function ($values) {
 			$options = array_column($this->options(), 'value');
--- a/tests/Form/Fields/SelectFieldTest.php
+++ b/tests/Form/Fields/SelectFieldTest.php
@@ -118,7 +118,10 @@ class SelectFieldTest extends TestCase
 			['a', 'a'],
 			['b', 'b'],
 			['c', 'c'],
-			['d', '']
+			['d', ''],
+			['1', '1'],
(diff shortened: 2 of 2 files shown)


lukasbestle (commented):
This is indeed a fix for the issue as we previously discussed. Have you checked if it's viable to cast to string earlier to fix such issues in general, independent of the field type?

distantnative:
Not sure where you are thinking of doing such a cast - but do the answer is: no I haven't. I'd also be afraid of broader regressions doing a broad change. The fields code isn't the most stable one.

bastianallgeier (approved):
I would merge it like this and keep the risk down.

afbora:
Why didn't we include it in 3.9.8?

PR #5770, opened 2023-10-09
Language label for the language switcher dropdown

distantnative:
## This PR …

- [x] Support to edit in language dialog

### Enhancement
- Language: customize the `label` of the Panel's languages dropdown for the currently selected language

### DIFF
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -428,6 +428,7 @@ class Language
 			'code'         => $this->code(),
 			'default'      => $this->isDefault(),
 			'direction'    => $this->direction(),
+			'label'        => $this->label(),
 			'locale'       => Locale::export($this->locale()),
 			'name'         => $this->name(),
 			'translations' => $this->translations(),
@@ -570,8 +571,10 @@ class Language
 	 * Returns a language variable object
 	 * for the key in the translations array
 	 */
-	public function variable(string $key, bool $decode = false): LanguageVariable
-	{
+	public function variable(
+		string $key,
+		bool $decode = false
+	): LanguageVariable {
 		// allows decoding if base64-url encoded url is sent
 		// for compatibility of different environments
 		if ($decode === true) {
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -39,6 +39,7 @@ class Language
 	protected string $code;
 	protected bool $default;
 	protected string $direction;
+	protected string $label;
 	protected array $locale;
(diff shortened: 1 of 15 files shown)


lukasbestle (commented):
TBH I'd keep the "translate" icon. One could still hide it via CSS if it doesn't fit.

afbora:
What do you think about also new optional `label` field for the language dialogs? Maybe it should be sync with code field as default.

image (4)

distantnative:
Added

Screenshot 2023-10-19 at 18 54 58

tests/Cms/Languages/LanguageTest.php
```diff
@@ -281,6 +281,25 @@ public function testIsDefault()
 		$this->assertFalse($language->isDefault());
 	}
 
+	/**
+	 * @covers ::label
+	 */
+	public function tesLabel()
```
lukasbestle:
```suggestion
	public function testLabel()
```

lukasbestle (commented):
If the label is left blank, the fallback to the code does not work.

Two possible solutions I can think of:

- Change `Language::update()` to replace an empty string for the label with `null`. This way, a dev could still manually set an empty string in the language file. But if changed via the UI, leaving the field blank will make it fall back to the code.
- In the frontend, use a similar logic like with the code and locale fields (when changing the name or code fields, automatically update the label fields with the uppercase code). This one would probably be the most flexible as then the label could also be manually set to an empty string from the UI.

src/Cms/Language.php
```diff
@@ -59,6 +60,7 @@ public function __construct(array $props)
 		$this->code         = trim($props['code']);
 		$this->default      = ($props['default'] ?? false) === true;
 		$this->direction    = ($props['direction'] ?? null) === 'rtl' ? 'rtl' : 'ltr';
+		$this->label        = trim($props['label'] ?? Str::upper($this->code));
```
afbora, line 63: This line is not working correctly. The `label` prop from the frontend is always sent as empty string. So if user doesn't enter label, language dropdown shows empty string.

Should be empty check for `$props['label']` instead null check or `label` argument should be sent as `null` if not entered.

lukasbestle: Yep, also ran into that. See the possible solutions in the comment above.

afbora: Oh, I missed that.

distantnative:
Closing this for now until it's ready

PR #5771, opened 2023-10-09
`Toolkit\Date`: Allow formatting with the globally configured handler

lukasbestle:
Taken from https://discord.com/channels/525634039965679616/1111220342522904706/1149433392543313921. I had first planned to override the native `format()` method, however that would break use cases that expect to use `date()`-style formatting. So I made it a separate method.

### Features

- New `$date->formatWithHandler()` method for Kirby date objects that allows to use different date handlers or even the globally configured one (default).

### Enhancements

- `Str::date()` and its dependents (e.g. `F::modified()`, `File::modified()`, `Dir::modified()`) now respect the globally configured date handler

### Breaking changes

- The `Str::date()` method and its dependents (e.g. `F::modified()`, `File::modified()`, `Dir::modified()`) no longer use the `date` handler if a different handler was configured globally with the `date.handler` option

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -319,7 +319,7 @@ class Str
 
 		// automatically determine the handler from global configuration
 		// if an app instance is already running; otherwise fall back to
-	 	// `date` for backwards-compatibility
+		// `date` for backwards-compatibility
 		if ($handler === null) {
 			$handler = App::instance(null, true)?->option('date.handler') ?? 'date';
 		}
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -318,16 +318,10 @@ class Str
 		}
 
 		// automatically determine the handler from global configuration
-		// if an app instance is already running
+		// if an app instance is already running; otherwise fall back to
+	 	// `date` for backwards-compatibility
 		if ($handler === null) {
-			if ($app = App::instance(null, true)) {
-				$handler = $app->option('date.handler', 'date');
-			}
-
-			// fall back to `date` for backwards-compatibility
-			// when the method was called without app and without
-			// overriding the default handler
-			$handler ??= 'date';
+			$handler = App::instance(null, true)?->option('date.handler') ?? 'date';
 		}
 
 		// `intl` handler
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -6,6 +6,7 @@ use Closure;
 use DateTime;
 use Exception;
 use IntlDateFormatter;
+use Kirby\Cms\App;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Query\Query;
 use Throwable;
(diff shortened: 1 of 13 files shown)


bastianallgeier (approved):
I like this a lot. It cleans up the date code across the board very nicely. @lukasbestle could you have a look at the failing unit tests though. I don't really understand how the email input is affected here.

src/Toolkit/Str.php
```diff
@@ -315,6 +317,19 @@ public static function date(
...
+		// if an app instance is already running
+		if ($handler === null) {
+			if ($app = App::instance(null, true)) {
+				$handler = $app->option('date.handler', 'date');
+			}
+
+			// fall back to `date` for backwards-compatibility
+			// when the method was called without app and without
+			// overriding the default handler
+			$handler ??= 'date';
```
afbora: If you didn't write a long version on purpose, I would write a short one like this:

```suggestion
			$handler = App::instance(null, true)?->option('date.handler') ?? 'date';
```

lukasbestle:
I think this is because the `VTest` class doesn't use its own app instance, so it worked before only by accident.

Also another test class before seems to set another locale, breaking the `DateTest` when run together.

I'll clean it up when I come home.

src/Toolkit/Str.php
```diff
@@ -315,6 +317,13 @@ public static function date(
 			return $format->format($time ?? time());
 		}
 
+		// automatically determine the handler from global configuration
+		// if an app instance is already running; otherwise fall back to
+	 	// `date` for backwards-compatibility
+		if ($handler === null) {
+			$handler = App::instance(null, true)?->option('date.handler') ?? 'date';
+		}
```
afbora, line 325: Sorry for the multiple suggestions

```suggestion
		$handler ??= App::instance(null, true)?->option('date.handler') ?? 'date';
```

lukasbestle: It's indeed shorter and does the same, but here I'd say the current version is more readable as it visually fits quite well with the `if null`, `if intl`, `if strftime` blocks. But I'd be fine with the change.

afbora: I suggested it because we have been using the `??=` syntax a lot lately. I'm OK with both. Feel free apply or discard 👍

lukasbestle: I'm sure @distantnative will agree with you. I'm fine with it either way, but I prefer the current version for the mentioned reason. :)

afbora: I'm sure Nico will even update this line with `??=` when he refactor some parts 😄 You might need a code comment like that:

> // Nico, please don't refactor this line with the `??=`

afbora (approved):

src/Toolkit/Str.php
bastianallgeier: I agree that the readability is better like this. But I also really don't mind too much. Both ways work for me.

PR #5776, opened 2023-10-10
Use array unpacking to replace `array_merge`

distantnative:
## This PR …

I know you will laugh at me and slightly hate me... but I had this already ready for v5 😆 Since it looks like we're dropping PHP 8.0 with v4 already, this is potentially ripe for review. 

But given the vast of changes, even if little, I'd also be 100% fine if we put this off for 4.1 or so.

### DIFF
--- a/src/Panel/Field.php
+++ b/src/Panel/Field.php
@@ -41,7 +41,7 @@ class Field
 		foreach ($field->dialogs() as $dialogId => $dialog) {
 			$routes = [
 				...$routes,
-				Dialog::routes(
+				...Dialog::routes(
 					id: $dialogId,
 					areaId: 'site',
 					options: $dialog
--- a/src/Panel/Field.php
+++ b/src/Panel/Field.php
@@ -39,11 +39,14 @@ class Field
 		$routes = [];
 
 		foreach ($field->dialogs() as $dialogId => $dialog) {
-			$routes = array_merge($routes, Dialog::routes(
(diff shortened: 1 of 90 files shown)


bastianallgeier:
Yep, looking at it again, it feels like we shouldn't do it at this point. I think it would be good to keep it on a lab branch, but close the PR. We are already suffering from too many unclosed PRs again.

afbora:
I've reviewed half of the PR. I'll finish the rest today. The changes are simple even though there are a lot of changes. `array_merge to [...]`. I'm not against including it in the beta process. I also loved the new unpacking feature 💛

src/Panel/Field.php
```diff
@@ -39,11 +39,14 @@ public static function dialog(
...
 
 		foreach ($field->dialogs() as $dialogId => $dialog) {
-			$routes = array_merge($routes, Dialog::routes(
-				id: $dialogId,
-				areaId: 'site',
-				options: $dialog
-			));
+			$routes = [
+				...$routes,
+				Dialog::routes(
```
afbora:
```suggestion
				...Dialog::routes(
```

distantnative:
@afbora thank you - I suggest you add your review comments, I fix them and we then close the PR and I reopen it for 4.x

afbora:
I found only one problem while reviewing. I have no other comments.

distantnative:
Will open this again for 4.x

Issue #5730, opened 2023-10-05
Upload of the identical file should not change the UUID


thin-k-design:
## Description
So normally, if you try to upload a file with the same name, you get an error like "The file with the name 'myfile.txt' already exists". Which is the behaviour you would expect to avoid replacing any file by mistake. 
If you try to upload exactly the same file, it is allowed. Which should not be really a problem as it is the same file. But the uuid of the file is also changed and therefore any field linking to this file become empty.

**Expected behavior**  
Either the same file should get the same error, that it already exists, or the uuid should not be changed at all.

## To reproduce
1. Go to any files field of files section
2. Upload a file
3. Note down the uuid in the content folder
4. Upload the same file
5. Notice that the uuid in the content folder has changed

## Your setup
I tried Kirby v3.6, v3.9 and v4.0-beta-2

**Your system (please complete the following information)**
 - OS: Windows
 - Browser: Chrome
 - Version: latest

distantnative:
I'd say the UUID shouldn't be changed in this case.

bastianallgeier:
I agree. The concept behind it was that this is basically treated like unchanged content. Just like when you submit the filename dialog without changing the name.

afbora:
I've implemented an idea in https://github.com/getkirby/kirby/pull/5781. Needs your reviews, please.

PR #5781, opened 2023-10-11
Don't change uuid if exact same file

afbora:
## This PR …

PR is ready to review. Just drafted because not sure about the solution.

### Fixes
- Upload of the identical file should not change the UUID #5730

### Breaking changes
None

## Ready?

### For review team

~~- [ ] Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)~~

### DIFF
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -229,9 +229,25 @@ trait FileActions
 		// gather content
 		$content = $props['content'] ?? [];
 
-		// make sure that a UUID gets generated and
-		// added to content right away
-		if (Uuids::enabled() === true) {
+		// make sure that a UUID gets generated
+		// and added to content right away
+		if (
+			Uuids::enabled() === true &&
+			empty($content['uuid']) === true
+		) {
+			// sets the current uuid if it is the exact same file
+			if ($file->exists() === true) {
+				$existing = $file->parent()->file($file->filename());
+
+				if (
+					$file->sha1() === $upload->sha1() &&
+					$file->template() === $existing->template()
+				) {
+					// use existing content data if it is the exact same file
+					$content = $existing->content()->toArray();
+				}
+			}
+
 			$content['uuid'] ??= Uuid::generate();
 		}
 
--- a/tests/Cms/Files/FileActionsTest.php
+++ b/tests/Cms/Files/FileActionsTest.php
@@ -545,6 +545,33 @@ class FileActionsTest extends TestCase
 	/**
 	 * @dataProvider parentProvider
 	 */
+	public function testCreateDuplicate($parent)
+	{
+		$source = static::TMP . '/source.md';
+
+		// create the dummy source
+		F::write($source, '# Test');
+
+		$result = File::create([
+			'filename' => 'test.md',
+			'source'   => $source,
+			'parent'   => $parent
+		]);
+
+		$uuid = $result->content()->get('uuid')->value();
+
+		$duplicate = File::create([
+			'filename' => 'test.md',
+			'source'   => $source,
+			'parent'   => $parent
+		]);
+
+		$this->assertSame($uuid, $duplicate->content()->get('uuid')->value());
+	}
+
+	/**
+	 * @dataProvider parentProvider
+	 */
 	public function testCreateMove($parent)
 	{
 		$source = static::TMP . '/source.md';


src/Cms/FileActions.php
```diff
@@ -229,10 +229,25 @@ public static function create(array $props, bool $move = false): File
...
+		) {
+			// sets the current uuid if it is the exact same file
+			// otherwise generates new uuid
+			if ($file->exists() === true) {
+				$existing = $file->parent()->file($file->filename());
+				if (
+					$file->sha1() === $upload->sha1() &&
+					$file->template() === $existing->template()
+				) {
+					$content['uuid'] = $existing->content()->get('uuid')->value();
```
afbora: I tried unset and set null solutions for this line but both no luck:

````php
unset($content['uuid']);

$content['uuid'] = null;
````

File: src/Cms/FileActions.php [hunk 2]
```diff
@@ -229,10 +229,25 @@ public static function create(array $props, bool $move = false): File
...
+			Uuids::enabled() === true &&
+			empty($content['uuid']) === true
+		) {
+			// sets the current uuid if it is the exact same file
+			// otherwise generates new uuid
+			if ($file->exists() === true) {
+				$existing = $file->parent()->file($file->filename());
+				if (
+					$file->sha1() === $upload->sha1() &&
+					$file->template() === $existing->template()
```
lukasbestle, line 244: If we cannot avoid calling `create()` at all: Can we move this logic to a separate method to avoid duplicating it between `FileActions` and `FileRules`?

lukasbestle (commented):
I was thinking if there is a way to skip the `create()` action entirely and exit early (successfully though). So basically everything would look as if the file `create()` action was run, but nothing would actually be done.

The question however: Would we want to run the hooks still or skip them as well?

afbora:
@lukasbestle I like the idea and I've tried like that 145a1500c3a965f9d38bd572ac5d33213fd4a633? What do you think?

lukasbestle:
To be honest I don't know what would be the best place for this logic. As I wrote, I'm not sure whether one would want the hooks to fire or not. If we need the hooks, it looks like we would need to build this "early skip" logic into `FileActions::create()` directly. It would also allow us to have the logic in just one place and avoid the duplication with `FileRules`. But there are too many question marks in my head at the moment to think of a good solution.

A minor issue: Passing a different `sort` value would be ignored with the current implementation. But before we fix that, we first need to know the general direction.

afbora:
If it's exactly the same file and no metadata is changed, why re-upload the file? If it's already the same, it sounds like the user is doing it by mistake or without realizing it.

I'm not sure what the correct behavior should be.

lukasbestle:
> If it's already the same, it sounds like the user is doing it by mistake or without realizing it.

Yes, I think that was the idea behind it. This would IMO speak for skipping the hooks. What do you think @distantnative @bastianallgeier?

Maybe we could still move the logic to the `create()` action though to avoid duplication with the `FileRules::create()` logic.

bastianallgeier:
I also think that we could skip the hooks.

bastianallgeier:
Just looking at this again. 

1. When the meta data (sort or template) is different, I feel like we need to throw an error early on without actually uploading the file

2. The logic to match the source could be in the Filesystem\File class if you ask me. Something like `File::isIdentical()` or `File::isSame()`. I think it's very useful for other scenarios.

lukasbestle:
Fully agree with both points. Now that I think about it again, it's obvious. Creating a file a second time can only succeed if the metadata is 100% identical, otherwise the update rules are circumvented.

afbora:
I got confused at one point. In `File::create()`, the uuid is generated and set before than rules. `FileRules` runs just before the file is actually uploaded. So when we compare the the metadata, the uuid is already fresh generated and different than existing file uuid. So in `File::isIdentical()`, we need to remove the `uuid` data when comparing the metadata. Is that correct?

bastianallgeier:
@afbora sorry for the delay. I think File::isIdentical should check the sha1 content. The rest can be ignored and sort and template checks can be separate in my opinion.

afbora:
I couldn't implemented `File::isIdentical()` structure, sorry.  When I tried to move the related code under `File::isIdentical()`, it didn't seem useful or logical to me. There would still be too much repetitive code. Therefore, I preferred to write it plain. We can still move to the new method after building correct structure.

bastianallgeier:
I'm looking at this again after way too much time and I have to say that I'm a bit confused. How can the same file be uploaded twice? Our file action checks already prevent this by checking for the same filename.

afbora:
@bastianallgeier Let me give you an clue.

https://github.com/getkirby/kirby/blob/develop-minor/src/Cms/FileRules.php#L120-L125

Currently if new uploaded file strictly same with a file that in server, will upload again with new generated uuid.

bastianallgeier:
I understand that part, but I still don't understand why the uploader already hits a duplicate error earlier and how we could ever reach those lines.

afbora:
No. 

1. Only if both (new  and server file) are identical, the upload process will continue (so don't throw exception). 

2. If it is a different file with the same filename, the uploader will give an error. (this is expected and correct)

Unexpected behavior is (first option) that the UUID is regenerated when the same file with the same name is uploaded.

bastianallgeier:
I understand that 1. is the intention. But I really can't get it to work. When I use the exact same file, the uploader still throws a duplicate warning. Is it working as expected for you? Maybe I'm doing something wrong.

afbora:
@bastianallgeier This is easy reproduce for me

https://github.com/getkirby/kirby/assets/3393422/1cf4aa41-c0ea-48db-85df-a1435f3190fd

bastianallgeier:
Ok I have absolutely no idea what I did wrong the first time. Now it's working fine 🙈

bastianallgeier (dismissed):

bastianallgeier (approved):

Issue #5787, opened 2023-10-12
Duplicating a duplicated page


nilshoerrmann:
## Description
Duplicating an already duplicated page creates an error because Kirby seems not to append a new suffix to the page name because it already exists.

**Expected behavior**  
If a page slug already ends on `-copy`, Kirby should start adding and raising an index number to find a unique slug. This should not fail after the first attempt, trying `-copy1`, `-copy2`, `-copy3`. There might well be a limit of tries but in most cases it should be possible to find an available name within a few tries.

**Screenshots**  
<img width="499" alt="grafik" src="https://github.com/getkirby/kirby/assets/25466/b4af7174-7357-44ad-a6a0-33d6a393f03f">

## To reproduce
1. Duplicate a page.
2. Duplicate that page again.

## Your setup
**Kirby Version**  
Kirby 3.9.6.1

afbora:
![screen-capture-1697-Exploring the universe Copy - Mægazine-localhost](https://github.com/getkirby/kirby/assets/3393422/ead92036-064c-4d1e-93f6-eb20d718a36d)

I can't reproduce the issue. Do you have custom model or plugin?

nilshoerrmann:
We do have a custom model but only to generate custom info texts for the panel section display.

afbora:
Do you have the issue on starter or plainkit?

nilshoerrmann:
I just tested again and it does not happen on the first copy. Take this:

<img width="740" alt="grafik" src="https://github.com/getkirby/kirby/assets/25466/27ea6bb1-0801-4dc5-a22a-9abbff9ade1e">

It works, if I try to duplicate the third page.
It fails, if I do the same on the first or second page.

nilshoerrmann:
> Do you have the issue on starter or plainkit?

I'll have to check that later because I have none installed on my machine right now.

lukasbestle:
> It works, if I try to duplicate the third page.
It fails, if I do the same on the first or second page.

Ah, yes. That makes sense.

We only append Copy but don't check if the resulting UID is already taken. I agree with your suggestion to append numbers. It could work like this:

1. If the UID does not already end with `-copy`, append `-copy`
2. Check if the resulting page exists (`while` loop). In the loop, append a number that is counted up.
3. Finally append the same number to the suggested title of the duplicate.

afbora:
Now I get it, it's already a page with the same `-copy` suffix. I honestly didn't think it was a bug 🙂

[...]

PR #5788, opened 2023-10-12
Append copy suffix of duplicated pages

afbora:
### Enhancement
- Append copy suffix of duplicated pages #5787

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/areas/site/dialogs.php
+++ b/config/areas/site/dialogs.php
@@ -479,6 +479,26 @@ return [
 				];
 			}
 
+			$slugAppendix  = Str::slug(I18n::translate('page.duplicate.appendix'));
+			$titleAppendix = I18n::translate('page.duplicate.appendix');
+
+			// if the item to be duplicated already exists
+			// add a suffix at the end of slug and title
+			$duplicateSlug = $page->slug() . '-' . $slugAppendix;
+			$siblingKeys   = $page->parentModel()->childrenAndDrafts()->pluck('uid');
+
+			if (in_array($duplicateSlug, $siblingKeys) === true) {
+				$suffixCounter = 2;
+				$newSlug       = $duplicateSlug . $suffixCounter;
+
+				while (in_array($newSlug, $siblingKeys) === true) {
+					$newSlug = $duplicateSlug . ++$suffixCounter;
+				}
+
+				$slugAppendix  .= $suffixCounter;
+				$titleAppendix .= ' ' . $suffixCounter;
+			}
+
 			return [
 				'component' => 'k-form-dialog',
 				'props' => [
@@ -487,8 +507,8 @@ return [
 					'value' => [
 						'children' => false,
 						'files'    => false,
-						'slug'     => $page->slug() . '-' . Str::slug(I18n::translate('page.duplicate.appendix')),
-						'title'    => $page->title() . ' ' . I18n::translate('page.duplicate.appendix')
+						'slug'     => $page->slug() . '-' . $slugAppendix,
+						'title'    => $page->title() . ' ' . $titleAppendix
 					]
 				]
 			];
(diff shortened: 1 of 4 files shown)


bastianallgeier:
I wonder if we could somehow press that in a reusable Str method.

afbora:
I tried it but it didn't seem very useful to me.

afbora:
We update the title like `Hello World Copy` to `Hello World Copy 1` but what if user already updated only title like `New Hello World` in duplicate dialog. So new title would be `New Hello World 1`. This doesn't make sense. May be we need to check title ends with `Copy` string, not sure.

src/Cms/PageActions.php
```diff
@@ -726,6 +726,24 @@ public function duplicate(string $slug = null, array $options = [])
 		// create the slug for the duplicate
 		$slug = Str::slug($slug ?? $this->slug() . '-' . Str::slug(I18n::translate('page.duplicate.appendix')));
 
+		// if the item to be duplicated already exist
+		// add a suffix at the end of slug and title
+		$siblingKeys = $this->parentModel()->childrenAndDrafts()->keys();
+		if (in_array($slug, $siblingKeys) === true) {
+			$suffixCounter = 1;
```
lukasbestle: I am thinking to start with `2` instead. Then the sequence would be:

```
Some Page
Some Page Copy
Some Page Copy 2
```

I think that fits better than `Copy` to `Copy 1`.

lukasbestle (commented):
> We update the title like Hello World Copy to Hello World Copy 1 but what if user already updated only title like New Hello World in duplicate dialog. So new title would be New Hello World 1. This doesn't make sense. May be we need to check title ends with Copy string, not sure.

The "Copy" suffix is already suggested to the user in the duplicate dialog, right? And the code changed by this PR only runs after the dialog has been confirmed if I understood it correctly.

If we already suggest the "Copy 2" suffix in the dialog, I think that would solve the issue you describe. If the user changes the suggested title and slug, they are responsible that the chosen slug does not already exist. But the suggested slug should be valid.

afbora:
I fixed the issue in backend side in the PR. As far as I understand you mean that we should suggest (move backend to frontend) in duplicate dialog, right?

**Existing page**
> Title: Test Copy
> Slug: test-copy

**`test-copy` duplicate dialog**
> Title: Test Copy 2
> Slug: test-copy-2

lukasbestle:
The frontend does not have a way to check if the page already exists. But the Panel backend that generates the props for the duplicate dialog does.

distantnative:
Isn't the duplicate dialog a dedicated dialog? And then the load closure for that dialogue could check what name would actually be available and pass it to the frontend. If the user changes it to something unavailable they get an error, would be just like the create dialog then.

lukasbestle:
Exactly.

afbora:
You're absolutely right! Now more make sense. Thanks for feedback 👍 Now ready to review (I'll rebase if it's acceptable)

config/areas/site/dialogs.php
```diff
@@ -479,6 +479,26 @@
...
+				$suffixCounter = 2;
+				$newSlug       = $duplicateSlug . $suffixCounter;
+
+				while (in_array($newSlug, $siblingKeys) === true) {
+					$newSlug = $duplicateSlug . ++$suffixCounter;
+				}
+
+				$slugAppendix  .= $suffixCounter;
+				$titleAppendix .= ' ' . $suffixCounter;
+			}
```
distantnative, line 500: I think it would be great to add this code as a universal method to  the `Str` class. Not sure about the name but maybe

```php
$siblings = $page->parentModel()->childrenAndDrafts()->pluck('uid');
$slug = Str::unique($page->slug(), fn ($slug) => in_array($slug, $siblings) === false, $appendix);
```

afbora: I tried it in a very similar way to yours, but it seemed to me to be a very special method for this case.

Also `Str::unique()` should return the new appendix, not the updated version of the first parameter. Because we need to use appendix for both slug and title.

As a name, it can be `Str::appendix()`.

bastianallgeier (approved):
I thought some more about it and we try really hard to kind of put the logic somewhere else. But when it's not really obvious how to make a useful method out of it, it's maybe not a very helpful approach to keep this PR waiting. If we find a good idea for it later, we could still refactor it.

PR #5792, opened 2023-10-13
Item docs and table option slot fix

bastianallgeier:
## Fixes

- Fixed option slot in the items table. The options column was always visible, no matter if the slot was set or not.

## Enhancements

- Additional docs for `k-items` and `k-item`


bastianallgeier:
I have no idea how to fix the eslint problem here. It looks like it is around for a while already.

panel/src/components/Collection/Items.vue
```diff
@@ -7,7 +7,10 @@
 		@sort="$emit('sort', $event)"
 		@option="onOption"
 	>
-		<template #options="{ row: item, rowIndex: index }">
+		<template
+			v-if="$scopedSlots.options"
+			#options="{ row: item, rowIndex: index }"
+		>
```
distantnative, line 13: Can you give an example when this is needed? Isn't the slot only filled anyways if provided?

PR #5802, opened 2023-10-14
Refactor and share collection/items props

distantnative:
### Enhancement
- New `layout` prop mixin


bastianallgeier (approved):
This is great. I wanted to do exactly the same after setting up the items lab examples :)

PR #5806, opened 2023-10-15
Refactor `k-toolbar`

distantnative:
## This PR …

Splitting the toolbar components: a dumb one that is there to ensure consistent markup between the different toolbar variants (also great reusable for developers). And then a textarea and a writer one that mainly do the logic of what buttons should be displayed and what happens when they are clicked.

I hope with this basis we can then tackle the various Safari toolbar bugs. I would hope that those can be reproduced in isolation with the new dumb `k-toolbar`. Then it would be much easier to work on debugging if we can disregard all the complex logic from the writer toolbar.

Best to check out the `toolbar` lab examples

### Enhancements
- New dumb `k-toolbar`

### Refactored
- Refactored toolbars for writer and textarea input with new `k-toolbar`
- Implemented `k-block-options` with `k-toolbar`

### Breaking changes
- Textarea input now uses `k-textarea-toolbar` not `k-toolbar`


bastianallgeier (approved):
Super happy about this!

PR #5808, opened 2023-10-15
Generate UI docs in `panel/tmp` during `dev`

distantnative:
Still the question remains what happens with leftover UI docs in `panel/tmp` when switching branches and maybe now an untrue docs JSON stays in `panel/tmp`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 # build files
 /dist.zip
 /kirby.zip
+/panel/tmp
 
 # cs fixer
 .php-cs-fixer.cache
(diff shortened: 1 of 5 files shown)


distantnative:
One idea could be to delete `panel/tmp` when stopping `npm run dev`

lukasbestle:
What is the plan for releases? Don't we need to build and ship the UI docs to make the lab work even with production versions? That would mean that we would have to manually build the docs. If we just run `npm run build`, we could easily forget that.

distantnative:
Vite plugin takes care of generating the docs on `npm run build`

bastianallgeier:
I think this looks like a good solution. I like the idea to automatically delete the folder when dev is stopped. I'd suggest to merge this already and then keep on improving it as we go.

bastianallgeier (approved):

distantnative:
go for it :)

PR #5812, opened 2023-10-16
Toolbar: fix closing dropdown on click

### LINKED ISSUE #5810: [v4] Blocks fields dropdown doesn't close after options click
## To reproduce

1. Go to any blocks field
2. Click "Paste" from options dropdown (or hide, duplicate, copy, paste after option)
3. You'll see options dropdown still open

## Your setup
Kirby beta.3(v4/develop)

distantnative:


afbora:
Fixes for a single block options dropdown but not for blocks field options dropdown:

image (6)

bastianallgeier:
I think this is an issue in DropdownContent and not in the toolbars. 

The onClick method in k-dropdown-content checks for the dialog element to be clicked and otherwise won't close the dropdown. @distantnative do you know why we introduced it here as well?

distantnative:
No I don't  recall.
Speculation: you might want to have drop-downs that you can interact with without them closing on every click

bastianallgeier:
Ok, I've checked the history and this goes way back. It is solved already when we use the options prop instead of defining custom dropdown items and maybe that's the better fix here as well.

bastianallgeier:
I've used the :options prop instead of rebuilding the dropdown items in the toolbar. I wonder if we should use text instead of label here as well though.

afbora:
I've also found following that seems related this issue:

https://github.com/getkirby/kirby/assets/3393422/eda7e18c-9fbe-4dfd-b9d1-0848c1b5e9c8

bastianallgeier:
I've refactored all dropdowns to use the options prop. I think this is definitely better than before, but we still need to decide if we want to always close the dropdown on click as it was in 3.9 or if we want to keep this breaking change.

distantnative:
I don't yet understand how using the `options` prop solves this issue

bastianallgeier:
The options click event already closes the dropdown

distantnative:
Ok but then we are just avoiding the issue ourselves but it's still there for anybody manually using dropdown items or so?

bastianallgeier:
That's what I mean. It seems like we are carrying this breaking change around for a while already. I couldn't fully trace it back, but it's there in beta.2 already. We could simply remove the check in the onClick method and this would also fix the problem again for every other dropdown. But then you would need to use .stop in click events that should not propagate and close the dropdown automatically.

distantnative:
I think
- make sure it works as in v3 to not have a breaking change for developers
- see if we can find the use case again why we added the change
- solve that in a different way

bastianallgeier:
Ok, I checked again and we actually really had click events that automatically closed the dropdowns in v3 and "forgot" about them somehow in v4. I've fixed it.

distantnative:
Can't approve this as this is technically my PR - so here: 👍

PR #5840, opened 2023-10-21
Fix unfocus blocks

### LINKED ISSUE #5621: [v4] Blocks field: Remove focus on click outside field
## Description
When I select a block, the blue outline and the block options appear. After that, I click outside the blocks field and ...
... in Kirby 3, the outline and the block options disappear.
... in Kirby 4, the outline and the block options do not disappear.

Screenshot

## Expected behaviour
All other fields lose focus when clicking outside the field, so the blocks field should probably behave the same.

## Your setup
**Kirby Version**  
v4-beta.1

distantnative:
### Fixes
- #5621

### DIFF
--- a/panel/src/components/Forms/Blocks/Blocks.vue
+++ b/panel/src/components/Forms/Blocks/Blocks.vue
@@ -93,10 +93,9 @@ export default {
 	},
 	data() {
 		return {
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):
Works great

PR #5862, opened 2023-10-23
Strip HTML from item title

### LINKED ISSUE #5856: [v4] Pages Section renders html tags in h3 title attribute
## Description

Passing a HTML string to a pages sections blueprint `text` property will render the HTML Tags inside the `<k-item> h3` title attribute.  Please see the screenshot for a better description.

**Expected behavior**  

Strip HTML tags, leading & trailing whitespace and line breaks from the `text` property value  before it is used for the title attribute

**Screenshots**  

image

image

## To reproduce
```php
// Create a Page Model with a panelText method that return a HTML string.
class TestPage extends Page {
  public function panelText(): string {
    return "<em>Some Text</em> " . $this->title()->value();
  }
}
```

```yaml
testSection: 
  type: pages
  text: "{< page.panelText >}"
```

## Your setup
**Kirby Version**  

4.0.0-beta.2

distantnative:
### Fixes
- #5856

### DIFF
--- a/panel/src/components/Collection/Item.vue
+++ b/panel/src/components/Collection/Item.vue
@@ -23,7 +23,7 @@
 
 		<!-- Content -->
 		<div class="k-item-content">
-			<h3 class="k-item-title" :title="$helper.string.unescapeHTML(text)">
+			<h3 class="k-item-title" :title="title">
 				<k-link v-if="link !== false" :target="target" :to="link">
 					<!-- eslint-disable-next-line vue/no-v-html -->
 					<span v-html="text ?? '–'" />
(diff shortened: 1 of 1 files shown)


bastianallgeier:
If we actually strip HTML here, wouldn't it make sense to do this on the server instead?

distantnative:
Im indifferent there. Less stuff to do in JS vs. smaller payload per item

bastianallgeier:
You are right. The additional payload seems worse.

bastianallgeier (approved):

lukasbestle:
This works if the `text` prop has been set to an HTML string (and that HTML is rendered in the Panel), but not if the page title includes HTML special characters for other reasons (e.g. if the title field contains those characters without any other configuration).

**To reproduce:**

- Set the title of a page `This is a page with < HTML characters >`
- The title will be displayed 1:1 in the page headline and in the pages field (which is correct to prevent XSS), but the title will be `This is a page with `.

bastianallgeier:
You are right :(

PR #5863, opened 2023-10-23
Fix `Cms\Collection::pagination()` return type

### LINKED ISSUE #5854: [v4] PHPDoc for `$collection->pagination()` results in wrong IDE hints
## Description
`$pages->pagination()` returns a different Pagination class than `$collection->pagination()`, yet the PHPDoc says that it's always `Toolkit/Pagination` that's returned which results in the folllowing error with [this example from the docs](https://getkirby.com/docs/cookbook/templating/pagination).
CleanShot 2023-10-23 at 01 12 44@2x

Not sure if this is new with v4, or how you want to tackle these types of issues at all.

## Your setup
**Kirby Version**  4.0-beta.2

distantnative:
### Fixes
- #5854

### DIFF
--- a/src/Cms/Collection.php
+++ b/src/Cms/Collection.php
@@ -260,7 +260,7 @@ class Collection extends BaseCollection
 	 * Add pagination and return a sliced set of data.
 	 *
 	 * @param mixed ...$arguments
-	 * @return \Kirby\Cms\Collection
+	 * @return $this|static
 	 */
 	public function paginate(...$arguments)
 	{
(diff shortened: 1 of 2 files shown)


src/Cms/Collection.php
```diff
@@ -273,6 +273,16 @@ public function paginate(...$arguments)
 		);
 	}
 
+	/**
+	 * Get the pagination object
+	 *
+	 * @return \Kirby\Cms\Pagination|null
+	 */
+	public function pagination()
```
bastianallgeier, line 281: Wouldn't it make sense to add return type hinting here? Or would that clash with the parent class?

distantnative: Yes I avoided it cause then we need to update all collection classes at once and maybe that's better for a separate step.

bastianallgeier (approved):

Issue #5877, opened 2023-10-25
[v4] Content kept in file when clearing an object field.


vincentrouleau:
## Description

When clearing a block field, content is not actually deleted from the page file.
Editing the fields works as expected but clearing the whole object tend to keep some if not all values.

This was tested in a page blueprint. Not tested in a block.

https://github.com/getkirby/kirby/assets/11572524/7992066a-2523-4313-9d7e-3d23982ab519

PR #5878, opened 2023-10-25
Fix object field empty state

afbora:
### Fixes
- Empty content state doesn't working
- Content kept in file when clearing an object field. #5877

### Breaking changes
None

## Ready?

- [ ] ~Unit tests for fixed bug/feature~

### For review team

### DIFF
--- a/panel/src/helpers/field.js
+++ b/panel/src/helpers/field.js
@@ -9,7 +9,7 @@ export function form(fields) {
 	const form = {};
 
 	for (const fieldName in fields) {
-		form[fieldName] = clone(fields[fieldName].default);
+		form[fieldName] = clone(fields[fieldName].default ?? null);
 	}
 
 	return form;
--- a/panel/src/components/Forms/Field/ObjectField.vue
+++ b/panel/src/components/Forms/Field/ObjectField.vue
@@ -69,7 +69,7 @@ export default {
 	},
 	data() {
 		return {
-			object: {}
+			object: null
 		};
 	},
 	computed: {
@@ -77,11 +77,7 @@ export default {
 			return this.$helper.object.length(this.fields) > 0;
 		},
 		isEmpty() {
-			if (!this.object) {
-				return true;
-			}
-
-			return this.$helper.object.length(this.object) === 0;
+			return !this.object;
 		},
 		isInvalid() {
 			return this.required === true && this.isEmpty;
@@ -122,7 +118,7 @@ export default {
 			return fields;
 		},
 		remove() {
-			this.object = {};
+			this.object = null;
 			this.save();
 		},
 		// TODO: field is not yet used to pre-focus correct field
@@ -161,10 +157,10 @@ export default {
 			});
 		},
 		save() {
-			this.$emit("input", this.object);
+			this.$emit("input", this.object ?? "");
 		},
 		valueToObject(value) {
-			return typeof value !== "object" ? {} : value;
+			return typeof value !== "object" ? null : value;
 		}
 	}
 };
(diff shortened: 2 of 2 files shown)


distantnative:
@afbora was this already an issue in v3? If not, why in v3 does it work with `null` but not in v4?

afbora:
No. Working in v3 since we use `{}`:

https://github.com/getkirby/kirby/blob/3.9.7/panel/src/components/Forms/Field/ObjectField.vue#L109

distantnative:
@afbora but this PR changes it in other places too where in v3 we use `null`, e.g. https://github.com/getkirby/kirby/blob/3.9.7/panel/src/components/Forms/Field/ObjectField.vue#L120

afbora:
I know. Actually I think doesn't matter since `isEmpty()` method covers `null` and `{}`. Am I wrong?

https://github.com/getkirby/kirby/blob/3.9.7/panel/src/components/Forms/Field/ObjectField.vue#L75-L81

distantnative:
That's probably right. In the end, I like it better if we can stick to one type - so your PR is probably already on the right side. The main regression I'd be worry about is that saving cleared/no content would not work anymore. But that's already exactly the case your PR seems to fix.

distantnative:
After this PR, I cannot add an empty object field (e.g. in our lab environment going to https://sandbox.test/panel/pages/fields+object and clicking the plus next to the "Simple" label).

On the develop branch, if I do this an empty object gets added (which shows me the empty preview). With the PR nothing is shown.

I think the field needs to have distinct states between: no content vs. empty content.

afbora:
Of course, it should be like that. I was tested no content and empty content scenario with `files` field, was working. But not working with a simple text field. So I can reproduce either.

distantnative:
I think that's broken: If I click the + button, an object field should be created - even if all its fields are empty. It should make no difference if the subfield is a text field or a files field. 

But yes, I saw this is also broken in v3 already in some way.

afbora:
There's something weird. I don't think it has anything to do with `null`, something is clearing the data of `object` but I haven't found it yet.

afbora:
Oh man. You're right. I've always test [example blueprint](https://getkirby.com/docs/reference/panel/fields/object#example) and was working. I can able to reproduce with single field. So yes this issue already exist in v3.

distantnative:
I think the state before this PR is actually more correct: `null` = empty, field itself has no content. `{}` = field is not empty, just maybe the subfields are empty.

afbora:
@distantnative https://github.com/getkirby/kirby/pull/5878/commits/2840081948c14e70ee79bf01c763788a4629e832 commit works for me but not sure about correct implementation. What do you think?

afbora:
Sorry. I've found a bug for latest commit.

afbora:
@distantnative Btw there are two issues:

- Empty content issue (v3 and v4)
- Content kept in file when clearing an object field (v4)

I think I've found a solution for both issues in https://github.com/getkirby/kirby/pull/5878/commits/295b31629bc21592194f6da87b5753a0c01d8440


````js
//  no content
{}

// empty content
{
  name: null
}

// filled content
{
  name: "John Doe"
}
````

File: panel/src/helpers/field.js [hunk 1]
```diff
@@ -9,7 +9,7 @@ export function form(fields) {
 	const form = {};
 
 	for (const fieldName in fields) {
-		form[fieldName] = clone(fields[fieldName].default);
+		form[fieldName] = clone(fields[fieldName].default ?? null);
```
afbora, line 12: ⚠️ This is a not solution idea!

I changed it `null` fallback (instead `undefined`) to give you an idea about the issue. Because `{name: undefined}` object stored as `{}`. I couldn't understand it.

afbora:
Closed favor of https://github.com/getkirby/kirby/pull/5900

PR #5885, opened 2023-10-29
Writer toolbar: fix node dropdown and empty marks

### LINKED ISSUE #5883: [v4] Writer nodes dropdown shown even if only 1 node dropdown entry
## Description
CleanShot 2023-10-29 at 12 30 46@2x
```yaml
              benefits:
                label:
                  de: Vorteile
                  en: Benefits
                type: list
                marks: false
                nodes:
                  - bulletList
```

With the following blueprint, the paragraph entry is shown in a node list even though it is not selectable and just used as a "hidden" node.

## Your setup
**Kirby Version**  v4-beta.2

distantnative:
### Fixes
- #5883
- Toolbar will ignore default mark/nodes list if no marks/nodes are allowed in writer instance (to not show just separators next to each other, which would be the only default entries that can be rendered)

### DIFF
--- a/panel/src/components/Forms/Writer/Toolbar.vue
+++ b/panel/src/components/Forms/Writer/Toolbar.vue
@@ -155,10 +155,7 @@ export default {
 		 * Whether there are any nodes to show in the toolbar
 		 */
 		hasNodes() {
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):
LGTM

PR #5886, opened 2023-10-29
Writer: support switch from list to paragraph node

distantnative:
### Enhancement
- Writer supports directly switching from a list to paragraph


bastianallgeier (approved):
This is so much better!

ovenum:
Will this also work for the Quote node? Could not find it referenced in the commits for this PR. 
Just assuming it’s also a node that can contain other nodes and should be handled the same as `bulletList` and `orderedList` for a complete fix (?)

PR #5887, opened 2023-10-29
Toolbar: fix btn not clickable in Safari

### LINKED ISSUE #5399: [v4] Blocks options not clickable when multiple selected in Safari
## Description
While testing the v4 I noticed in Safari the block merge feature doesn't seem to work. Block splitting does work but merging doesn't. Tested Brave as well and it work as expected there.

**Expected behavior**  
The expectation was that it would join two text blocks shown in the video.

**Video**  
https://share.cleanshot.com/GLP1LCxw7V4Bnctphff1

 - Device: MacBook Pro
 - OS: MacOS 13.3.1
 - Browser: Safari
 - Version: 16.4

### LINKED ISSUE #5442: [v4] nodes in writer field not accessible in Safari
## Description
The writer fields inline toolbar disappears when clicking on the node dropdown.

**Expected behavior**  
The node dropdown options should open instead of closing the toolbar.

**Screenshots**  
Bildschirmaufnahme-2023-08-02-um-10 19 31

## To reproduce
1. Open Safari
2. Open a page containing a writer field with inline toolbar in panel
3. Select a text
4. Try to change the node type

## Setup
 - Kirby: 4.0.0-alpha.5 and 4.0.0-alpha.6
 - Device: MacBook Pro (13-inch, 2017)
 - OS: macOS 13.5 (22G74)
 - Browser: Safari
 - Version: 16.6 (18615.3.12.11.2)

distantnative:
### Fixes
- https://github.com/getkirby/kirby/issues/5442
- https://github.com/getkirby/kirby/issues/5399

### DIFF
--- a/panel/src/components/Forms/Blocks/BlockOptions.vue
+++ b/panel/src/components/Forms/Blocks/BlockOptions.vue
@@ -1,5 +1,9 @@
 <template>
-	<k-toolbar :buttons="buttons" class="k-block-options" />
+	<k-toolbar
+		:buttons="buttons"
+		class="k-block-options"
+		@mousedown.native.prevent
+	/>
 </template>
 
 <script>
--- a/panel/src/components/Forms/Toolbar/Toolbar.vue
+++ b/panel/src/components/Forms/Toolbar/Toolbar.vue
@@ -11,6 +11,7 @@
 				:icon="button.icon"
 				:title="button.label"
 				:class="['k-toolbar-button', button.class]"
+				tabindex="0"
 				@keydown.native="button.key?.($event)"
 				@click="
 					button.dropdown?.length
--- a/panel/src/components/Forms/Writer/Extensions/Toolbar.js
+++ b/panel/src/components/Forms/Writer/Extensions/Toolbar.js
@@ -17,15 +17,6 @@ export default class Toolbar extends Extension {
 	}
 
 	init() {
-		this.editor.on("blur", ({ event }) => {
-			// don't close the toolbar if the blur event happened inside the writer
-			if (this.writer.$el.contains(event.target) === true) {
-				return;
-			}
-
-			this.component?.close(event);
-		});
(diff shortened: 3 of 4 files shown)


bastianallgeier:
Unfortunately, this does not solve the problems for me in Safari 16.5.2

distantnative:
In Safari 17.2 it does so might be still worth not discarding this fix as it partially then help at least. I'd have to hand over fixing for Safari 16.x to you as I can't test it.

bastianallgeier:
This stuff is so weird. If I add a mousedown.stop to the toolbar buttons, it works for me partially. It seems like the event order is somehow messed up on Safari and the mousedown does already trigger something further up to close the toolbar again. But that still does not solve the dropdowns then. I wonder if we should fix this at all for older Safaris or just rely on updates. It's obviously a rather nasty Safari issue after all.

distantnative:
I mean the problem is a bit that there's probably still quite a bunch of Safari 16 out there and it really affects the usability for them :/

bastianallgeier:
Yes, definitely. I also cannot upgrade at the moment because I'm still on an older OS version.

bastianallgeier:
Ok, I think I solved it in a way that should work for all browsers. But it would be good to have a few more tests.

distantnative:
@bastianallgeier seeing that your changes only concern the Writer field, does https://github.com/getkirby/kirby/pull/5887/files#diff-9cc2332bf188c695fcdd8b8e50633fe6e232322b020babd71492536e2038d662R14 fix the toolbar issue for you in Safari for e.g. block options dropdown already?

bastianallgeier:
No it doesn't. I need to look at that one as well.

bastianallgeier:
Ok, preventing the mousedown event in the block option toolbar does the trick.

bastianallgeier (approved):

PR #5889, opened 2023-10-29
Writer: rely on Vue defaults, more consistent

### LINKED ISSUE #5851: [v4] Image Blocks preview shows line in caption Writer toolbar even without nodes
## Description
In the Image Blocks preview:
CleanShot 2023-10-22 at 14 43 22@2x

In the sidebar:
CleanShot 2023-10-22 at 14 43 50@2x

Blueprint:
```yaml
name: field.blocks.image.name
icon: image
preview: image
wysiwyg: true

fields:
  location:
    type: hidden
    default: 'kirby'
  image:
    extends: fields/image
    layout: cards
    size: xl
  caption:
    label: field.blocks.image.caption
    extends: fields/writer
```

## Your setup
**Kirby Version**  4.0-beta.2

distantnative:
## This PR …

Instead of transforming defaults in the PHP just for a writer field, we should have this default fallback logic baked into the Vue component so that it's present for all use cases (e.g. block previews).

### Fixes
- #5851

### DIFF
--- a/config/fields/writer.php
+++ b/config/fields/writer.php
@@ -61,23 +61,6 @@ return [
 		}
 	],
 	'computed' => [
-		'toolbar' => function () {
-			return array_merge([
-				'inline' => true,
-				'marks'  => $this->marks ?? [
-					'bold',
-					'italic',
-					'underline',
-					'strike',
-					'code',
-					'|',
-					'link',
-					'email',
-					'|',
-					'clear'
-				]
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
This is a great idea! We should have a more consistent way to setup defaults in general. Sometimes they are in the backend, sometimes in the frontend.

PR #5891, opened 2023-10-29
Fix `PluginAssets::clean()`

### LINKED ISSUE #5836: [v4] Plugin assets resolves as JSON (500 Internal Server Error)
## Description

Sometimes plugin assets resolves as JSON. This issue does not always occur. I don't know exactly when it occurs.

> Request URL: /media/plugins/afbora/test/2538303991-1697822431/css/style.css
> Request Method: GET
> Status Code: 500 Internal Server Error
> Content-type: application/json; charset=UTF-8
>
> Refused to apply style from '/media/plugins/afbora/test/2538303991-1697822431/css/style.css' because its MIME type ('application/json') is not a supported stylesheet MIME type, and strict MIME checking is enabled.

Use following code to reproduce:
````php
<?= css(kirby()->plugin('afbora/test')->asset('css/style.css')) ?>
````

## To reproduce

1. Download plainkit with v4 here: [plugin-assets-resolve.zip](https://github.com/getkirby/kirby/files/13056794/plugin-assets-resolve.zip)
2. Go to homepage and open browser console
3. Try to refresh the page multiple times
4. Check browser console after each refresh
5. You'll see 500 internal server issue for a asset

## Your setup
Kirby 4.0.0-beta.3/v4-develop

## Additional context
When I checked `/media/plugins/afbora/test` directory, It's empty. When I visit `/media/plugins/afbora/test/2538303991-1697822431/css/style.css` url from browser, opens correctly in browser and I can see the `style.css` in `/media/plugins/afbora/test`. But I refresh the homepage, `/media/plugins/afbora/test` directory cleared again.

### PR DESCRIPTION
**Pull request by distantnative**
## This PR …

Method now ensures that also all parent folders of plugin assets are considered when filtering those to arrive with a list of stale files. Previously, those weren't included while the index from the media folder did list folders and would thus delete those.

### Fixes
- #5836

## Ready?

### For review team

### DIFF
--- a/src/Cms/PluginAssets.php
+++ b/src/Cms/PluginAssets.php
@@ -31,21 +31,34 @@ class PluginAssets extends Collection
 			$media  = $plugin->mediaRoot();
 			$assets = $plugin->assets();
 
-			// get outdated media files by comparing all
-			// files in the media folder against the current set
-			// of asset paths
-			$files  = Dir::index($media, true);
+			// get all media files
+			$files = Dir::index($media, true);
+
+			// get all active assets' paths from the plugin
 			$active = $assets->values(
-				fn ($asset) => $asset->mediaHash() . '/' . $asset->path()
+				function ($asset) {
+					$path  = $asset->mediaHash() . '/' . $asset->path();
+					$paths = [];
+					$parts = explode('/', $path);
+
+					// collect all path segments
+					// (e.g. foo/, foo/bar/, foo/bar/baz.css) for the asset
+					for ($i=0; $i < count($parts); $i++) {
+						$paths[] = implode('/', array_slice($parts, 0, $i + 1));
+
+						// TODO: remove when media hash is enforced as mandatory
+						$paths[] = implode('/', array_slice($parts, 1, $i + 1));
+					}
+
+					return $paths;
+				}
 			);
 
-			// TODO: remove when media hash is enforced as mandatory
-			// @codeCoverageIgnoreStart
-			$active = array_merge($active, $assets->values(
-				fn ($asset) => $asset->path()
-			));
-			// @codeCoverageIgnoreEnd
+			// flatten the array and remove duplicates
+			$active = array_unique(array_merge(...$active));
 
+			// get outdated media files by comparing all
+			// files in the media folder against the set of asset paths
 			$stale  = array_diff($files, $active);
 
 			foreach ($stale as $file) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Cms/PluginAssets.php [hunk 1]
```diff
@@ -31,21 +31,34 @@ public static function clean(string $pluginName): void
...
+					$paths = [];
+					$parts = explode('/', $path);
+
+					// collect all path segments
+					// (e.g. foo/, foo/bar/, foo/bar/baz.css) for the asset
+					for ($i=1, $max = count($parts); $i <= $max; $i++) {
+						$paths[] = implode('/', array_slice($parts, 0, $i));
+
+						// TODO: remove when media hash is enforced as mandatory
+						$paths[] = implode('/', array_slice($parts, 1, $i));
```
afbora, line 50: I didn't understand what exactly this line does.

**With this line:**
````php
  array(6) {
    [0]=>
    string(21) "1362650116-1646303478"
    [1]=>
    string(2) "js"
    [2]=>
    string(24) "1362650116-1646303478/js"
    [3]=>
    string(16) "js/jquery.min.js"
    [4]=>
    string(38) "1362650116-1646303478/js/jquery.min.js"
    [5]=>
    string(16) "js/jquery.min.js"
  }
````

**Without this line:**
Following seems more correctly than first one. Not sure.

````php
  array(3) {
    [0]=>
    string(21) "1362650116-1646303478"
    [1]=>
    string(24) "1362650116-1646303478/js"
    [2]=>
    string(38) "1362650116-1646303478/js/jquery.min.js"
  }
  ````

afbora: I guess I understand. There is a case where it can be used without hash, so we need hashless paths as well. Right?

distantnative: For backwards compatibility we also support media paths that don't include the media hash. So if we have the asset path `js/jquery.min.js` with media hash `1362650116-1646303478` there are actually two locations in the media folder where it could end up:
- `1362650116-1646303478/js/jquery.min.js`
- `js/jquery.min.js`

This is why we're collecting path segments for both options:
- `1362650116-1646303478/`, `1362650116-1646303478/js/` and `1362650116-1646303478/js/jquery.min.js`
- `js/` and `js/jquery.min.js`

Once we remove this deprecated option of having media URLs without hash, we can remove that line.

afbora: Thank you! 🙌

afbora (approved):
Works great and LGTM 👍

Issue #5712, opened 2023-09-29
Writer field: toolbar causes selection to jump


rasteiner:
## Description

When selecting text in a writer field, the toolbar appears. Sometimes, while dragging the mouse to select the text, your mouse cursor happens to go over the toolbar, in that case the "head" of the text selection (the "moving side") jumps to a new place. It seems to move almost exactly the same distance as the distance between the edge of the toolbar and the mouse position before entering it (see video and then screenshot). 

**Expected behavior**  
Toolbar should have no influence over what text is selected

**Screenshots**  

https://github.com/getkirby/kirby/assets/6684137/17865164-8c79-48b9-aaad-5fc39b92a793

![image](https://github.com/getkirby/kirby/assets/6684137/27ef68d7-6f29-4ab6-b04a-dfac27d60a0c)
(the 2 red bars have the same length)

## To reproduce

1. Open Google Chrome
2. Go to a writer field, enter some text
3. Click and drag your mouse to select text, while doing so move the mouse cursor over the toolbar (easier at the top of the writer field, or between paragraphs). 
4. See selection unexpectedly change 

## Your setup
**Kirby Version** 
Tested in 3.9.6.1, 4.0.0-beta.1 and 4.0.0-beta.2

**Console output**  

**Your system (please complete the following information)**
 - Device: PC
 - OS: Windows 11
 - Browser: Chrome, Edge
 - Version: 117

**Additional context**  

I have never noticed this before, so I checked if this could be a regression of beta.2... 
But beta.1 and actually also 3.9.6.1 do the same. I kind of suspect that this is somewhat new to Chrome (and Edge), as Firefox and Safari don't show this behaviour. 
On Chrome on Mac, I can reproduce the issue, but only when moving the mouse on a 3px tall row of pixels at the bottom of the toolbar. While on Windows the effect is jarring even inside of the writer field when I "select upwards" and the toolbar lags behind a bit and the cursors happens to go over it.

rasteiner:
So, I can reproduce a very similar issue with the [TipTap bubble menu demo](https://tiptap.dev/api/extensions/bubble-menu), so I guess it's either ProseMirror or Chrome. That's a bit frustrating...

Issue #5449, opened 2023-08-03
[v4] Layout Toolbar Overflow


lukaskleinschmidt:
## Description
There seems to be an overflow issue in narrow layouts.

![grafik](https://github.com/getkirby/kirby/assets/10027897/22e1392e-e8e4-4f28-ad87-b0b0d52ca508)

## Your setup
4.0.0-alpha.6

**Your system (please complete the following information)**
 - Browser: Firefox
 - Version: 117

distantnative:
I'm afraid this might need a JS solution:
1. measure available width (max. width of field)
2. measure space nodes and marks would take
3. move marks into a "..." dropdown as long as (2) is larger than (1)

bastianallgeier:
One part is already fixed. The layout box is no longer cutting off the toolbar. But the toolbar is still not at full width and will have a scrollbar even if there's enough space. I think this is related to the new positioning logic.

distantnative:
@bastianallgeier Three things:
1. `.k-writer-toolbar[data-inline="true"]` should have a `max-width: none;` added to be full-width without scrollbar
2. The positioning logic still fails: it first tries to correct overflow to the left, then to the right. If the toolbar is larger than the container, it first shifts it to the right, then back to the left and will have an overflow. This is because we can only assess the overflow/offset in relation to the container - not the full window (where it might have more space to either side).
3. With `position: absolute` it'll be very hard to break out of the editor itself. If we want to use the space around the editor itself for the toolbar to overlay, we'll need to move to `position: fixed`

PR #5892, opened 2023-10-29
Fix Writer toolbar positioning

distantnative:
### TBD
- [ ] This can clash with container queries when there is too little space and it will reach into another section that is defined by a container query, that section will overlap the toolbar. I don't think we can avoid this if we don't want to turn the toolbar into a dialog element.

### Fixes
- Writer toolbar now isn't just contained to its editor but will extend to space around it if necessary (= toolbar larger than editor), but will ensure not to reach outside viewport
https://github.com/getkirby/kirby/issues/5449
- https://github.com/getkirby/kirby/issues/5712

### DIFF
--- a/panel/src/components/Forms/Writer/Toolbar.vue
+++ b/panel/src/components/Forms/Writer/Toolbar.vue
@@ -296,16 +296,25 @@ export default {
 			let x = selection.x - editor.x + selection.width / 2 - toolbar.width / 2;
 			let y = selection.y - editor.y - toolbar.height - 5;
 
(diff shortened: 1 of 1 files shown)


bastianallgeier:
I could not find any issues in my tests. Do you want to continue even with the TBD task?

distantnative:
@bastianallgeier if you're ok either way the issue under TBD

bastianallgeier:
I'm fine with the current state. We can still try to tackle the remaining issue later.

PR #5898, opened 2023-10-31
Fix quote node

### LINKED ISSUE #5894: [v4] Quote node not shown as active node
Instead the paragraph node is shown as active

distantnative:
## This PR …

I know this isn't the best solution - but it"s what we use for lists so far, so I think best quick fix is to do the same for the quote node.

### Fixes
- https://github.com/getkirby/kirby/issues/5894
- Allows to switch directly to paragraph from quote node


bastianallgeier (approved):
I agree that it could be less hard-wired, but it does the job and is a big UX improvement for quotes.

PR #5900, opened 2023-11-01
Better way to handle default values for objects

bastianallgeier:
I think this is a more stable option to create default objects for a set of fields. It is a bit more complex, but I think we can handle it. 

@afbora @distantnative what do you think about it? 

## Fixes

- Improved empty content state for object fields
- Content is no longer kept when clearing an object field #5877

## Enhancements

- New `$helper.field.defaultValue(field)` method
- Better default value creation in `$helper.field.form(fields)`

### DIFF
--- a/panel/src/helpers/field.js
+++ b/panel/src/helpers/field.js
@@ -13,13 +13,14 @@ export function defaultValue(field) {
 	const component =
 		window.panel.app.$options.components[`k-${field.type}-field`];
 	const valueProp = component?.options.props.value;
-	const valuePropDefault = valueProp?.default;
 
 	// if the field has no value prop,
 	// it will be completely skipped
 	if (valueProp === undefined) {
 		return undefined;
 	}
+
+	const valuePropDefault = valueProp?.default;
 
 	// resolve default prop functions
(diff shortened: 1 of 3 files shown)


panel/src/helpers/field.js
```diff
@@ -1,5 +1,38 @@
...
+ */
+export function defaultValue(field) {
+	if (field.default !== undefined) {
+		return clone(field.default);
+	}
+
+	const component =
+		window.panel.app.$options.components[`k-${field.type}-field`];
+	const valueProp = component?.options.props.value;
+	const valuePropDefault = valueProp?.default;
```
distantnative: We could move this below line 22. I know, total nitpick - ignore it if too much

distantnative:
Can we move this to beta.4/RC? Won't be able to test this sufficiently before later tomorrow unfortunately.

afbora (approved):
I can't comment about new `defaultValue` helper but works great for me 👍

afbora:
@distantnative IMHO would be great for RC3 since object field values can't deletable.

bastianallgeier:
Ok, after @afbora's tests, let's merge this one for the beta.3.

Issue #5251, opened 2023-06-06
[v4] Textarea scrolls to start when inserting newlines


mrflix:
Using `4.0.0-alpha.2` I experience the following behavior when editing long text in the textarea: the page scrolls up when I add newlines.

https://github.com/getkirby/kirby/assets/60777/b0dc2244-59c0-48b7-8dde-318bb9b210ab

distantnative:
@mrflix by editing you mean simply typing a character or something special like pasting?

mrflix:
Pressing Enter makes the page scroll to the top. Inserting text also makes the scroll jump when I move the cursor and type.

https://github.com/getkirby/kirby/assets/60777/2513743f-c5b6-4831-b17d-abcdc8811428

Seems like a `scrollIntoView` party. Browser is latest Chrome on macOS Catalina.

distantnative:
@mrflix my hunch is rather a focus party

distantnative:
@bastianallgeier I think this is actually caused by the new `Autoresize` custom element. Do you remember why on every resize you first set the height to `0`?

bastianallgeier:
That's doing the trick to calculate the overscrolling height as far as I remember.

distantnative:
I think that's what makes jump the viewport to somewhere completely else on very long textareas.

Also saw https://css-tricks.com/the-cleanest-trick-for-autogrowing-textareas/ which is based on mainly CSS. Maybe an alternative.

bastianallgeier:
Uh, that looks neat indeed. We should give it a try.

distantnative:
For that implementation, it would be much easier to directly integrate it into `k-textarea-input` instead of trying got offer a standalone `k-autosize` wrapper. I don't think we're using it at all in any other way than for the texture input. So maybe we shouldn't burden ourselves with trying to offer a standalone solution?

bastianallgeier:
Sounds good to me.

distantnative:
I did replace it and it's nice, but it doesn't fix the mentioned issue. Enter seems to cause the viewport to jump, putting the new line on top (below the toolbar). Might be a common browser behaviour, which is why the autosize library caches and restores scroll position constantly.

bastianallgeier:
@distantnative Hm, maybe we need to get back to the old library then? It seems like a pretty complicated part that could actually not be worth the effort right now.

distantnative:
[...]

PR #5907, opened 2023-11-02
Revert autoresize scroll restore

### LINKED ISSUE #5868: [v4] Textarea doesn't scroll to the new line
## Description

Related PR: https://github.com/getkirby/kirby/pull/5846

## To reproduce

1. Go to textarea field
2. Enter long long text
3. Enter multiple new lines
4. See issue that you won't see last line (browser should scroll to end of text)

## Your setup
**Kirby Version**  
Beta.3 (v4/develop)

### LINKED ISSUE #5871: [v4] Textarea: doesn't restore size after revert changes
## To reproduce

1. Go to textarea field
2. Enter long long text
3. Revert page changes 
4. See issue that textarea size still long (should be restore initial size)

## Your setup
**Kirby Version**  
Beta.3 (v4/develop)

### LINKED ISSUE #5906: [v4] Textarea toolbar button click makes scroll position jump
## Description
When editing content in a textarea field, the toolbar creates unexpected Markdown for block elements. Think of this existing content, the pipe being the cursor:

```md
# An example |headline
```

If I select headline 3 in the toolbar, the following will be created:

```md
# An example ### |headline
```

Same example as before, if I click headline 1 again to unselect it and make the current line a paragraph again, the following will be created:

```md
# An example # |headline
```
 
**Expected behavior**  
The block level elements should alter the Markdown blocks and not thus change the marks at the very beginning of the line.

## To reproduce

```yml
title: Markdown Test
fields:
  text:
    label: Text
    type: textarea
```

## Your setup
**Kirby Version**  
Kirby 4 Beta 3 on Firefox 119 (macOS Sonoma)

distantnative:
### Fixes
- https://github.com/getkirby/kirby/issues/5906
- https://github.com/getkirby/kirby/issues/5871
- https://github.com/getkirby/kirby/issues/5868

### Known issue
- Need to reopen #5251

### DIFF
--- a/panel/src/components/Forms/Autosize.js
+++ b/panel/src/components/Forms/Autosize.js
@@ -17,12 +17,10 @@ export default class Autosize extends HTMLElement {
 			textarea.autosize = () => {
 				textarea.style.height = "auto";
 				textarea.style.height = textarea.scrollHeight + "px";
-				this.restoreScroll();
 			};
 
 			// trigger resize on input
 			textarea.addEventListener("input", () => textarea.autosize());
-			textarea.addEventListener("beforeinput", () => this.storeScroll());
 		}
 
 		// resize all textareas when the container size changes
@@ -37,16 +35,5 @@ export default class Autosize extends HTMLElement {
 
 	disconnectedCallback() {
 		this.resizer.unobserve(this);
-	}
-
-	restoreScroll() {
-		if (this.scrollY !== undefined) {
-			window.scroll(0, this.scrollY);
-			this.scroll = undefined;
-		}
-	}
-
-	storeScroll() {
-		this.scrollY = window.scrollY;
 	}
 }
--- a/panel/src/components/Forms/Input/TextareaInput.vue
+++ b/panel/src/components/Forms/Input/TextareaInput.vue
@@ -112,8 +112,10 @@ export default {
 		}
 	},
 	watch: {
-		value() {
+		async value() {
 			this.onInvalid();
+			await this.$nextTick();
+			this.$refs.input.autosize();
 		}
 	},
 	mounted() {


afbora:
Before the reviewing the PR; Sorry for the curiosity. I just wonder that why [we removed](https://github.com/getkirby/v4/pull/33/files) the `autosize` dependency? What was the reason? Maybe we can avoid potential issues if we switch back to this dependency.

distantnative:
@afbora the discussion in https://github.com/getkirby/kirby/issues/5251 says it already exists in v3.x where we still used that library. That's why so far I have assumed that the library has the same issue.

afbora:
Hmm, actually I can't reproduce the scroll jump issue in 3.9.5 or 3.9.7 🤷‍♂️ This issue looks like only valid in v4 for me.

distantnative:
I'll set up a v4 branch with the old library for testing

distantnative:
Here is the branch https://github.com/getkirby/kirby/tree/v4/fix/autosize

However, I can reproduce the issue both, on this branch as well as on 3.9.5.

afbora:
@distantnative I think It's not exactly the same issue. In 3.9.7 the scroll jumps up a bit and I can still see the paragraph, but in beta.2 the scroll goes to the top.

### 3.9.7

https://github.com/getkirby/kirby/assets/3393422/eae47dd5-a62c-44f2-9122-19957da175de


### Beta.2

https://github.com/getkirby/kirby/assets/3393422/dae2a96f-ed51-4b73-8e20-f2dc879b20b2

afbora:
Also I think this is intended behavior that scrolling to start of new paragraph (not top of the field) in 3.9.x. Also you can test in https://www.jacklmoore.com/autosize (you'll see the same behavior)

The only problem is that our sticky toolbar closes/covers the first paragraph (when scroll to the start of new paragraph). If we fix it, there is no problem.

distantnative:
Ok then maybe using  https://github.com/getkirby/kirby/tree/v4/fix/autosize would be better. But I don't have a good idea how to fix that the scroll is set to behind our toolbar.

afbora:
OK. I've tested your new branch but doesn't behave exactly like 3.9.x. The beginning of the paragraph goes up a little more than usual. Because we have new sticky page header. If you test it after deleting page header (`<header class="k-header k-page-view-header"></header>`) from browser console. You'll understand the issue.

Also I think we should remove old `<k-autosize>` tag from new branch.

distantnative:
What do you mean with that last comment? We do remove the k-autosize: https://github.com/getkirby/kirby/commit/2c3850c335742bc54d9bfe61b8f3dea37f7eaab9#diff-08c8cb2c84dd92020daa4b0d4349c224e528cb5f3b8912c7ddbcf17c6d2f3fdfL33

And yes, v4 has more "overlapping" due to the header. But how can we solve this as the scroll restoration by the autosize library is magic/obscure?

afbora:
@distantnative I mean this line https://github.com/getkirby/kirby/blob/v4/fix/autosize/panel/src/components/Forms/Input/TextareaInput.vue#L13

distantnative:
Ohhh right!

afbora:
@distantnative If you create new PR for the branch, lets discuss about the solution in it.

distantnative:
👉 https://github.com/getkirby/kirby/pull/5914

Issue #5911, opened 2023-11-03
[v4] TOTP Setup dialog overflows


therephil:
## Description
the TOTP Setup dialog overflows because the `fields` variant on the grid adds 12 columns that are empty but still have a gap.
`gap: var(--spacing-8);` 

**Screenshots**  
<img width="681" alt="screenshot" src="https://github.com/getkirby/kirby/assets/4098146/f2025aa4-cfef-4a01-8ecf-35d167992b45">
<img width="694" alt="screenshot-2" src="https://github.com/getkirby/kirby/assets/4098146/88128e88-cd7b-443a-945a-09132a099b44">

## To reproduce

1. Go to a user account
2. Click on 'set up one-time codes'
3. See error

## Your setup

- Version: Beta.3
- Browser: FF Dev 120.0b4 and Chrome 119.0.6045.105 
- OS: Mac OS Ventura 13.5.1

PR #5912, opened 2023-11-03
Fix grid and border radius

bastianallgeier:
## Fixes

- https://github.com/getkirby/kirby/issues/5911


distantnative:
@bastianallgeier wouldn't this make sense to generally fix this for `k-dialog-fields`? I'd think any use could have the same issue?

bastianallgeier:
I've moved it to the dialogs fields component

distantnative (approved):

Issue #5910, opened 2023-11-03
[v4] Layout field: drag-drop a block into empty layout doesn't working


afbora:
## To reproduce

1. Go to layout field
2. Add a new layout and add into few blocks
3. Add a new layout and try to drag-drop a block into new empty layout
4. See error that you can't drop

## Your setup
**Kirby Version**  

Beta.3

distantnative:
Do you think this is connected to https://github.com/getkirby/kirby/issues/5290 or unique issue?

bastianallgeier:
I think it is connected to the flex thing that we had a while ago which we changed to improve the sizing on drag & drop. This means that the blocks are not full height when empty and thus you cannot drag something into an empty layout.

afbora:
I think it's not connected since works as expected in 3.9.7 and beta.2.

distantnative:
But if it worked in beta.2 ... I think @bastianallgeier the change you refer to was not just in beta.3 but earlier, or?

bastianallgeier:
That's true. Need to dig deeper then :(

distantnative:
Not too deep, stupid mistake when trying to clean up: 
The fault is this line https://github.com/getkirby/kirby/blob/v4/develop/panel/src/components/Forms/Blocks/Blocks.vue#L9 - when no blocks exist, k-draggable right now isn't in the DOM and cannot accept blocks dragged over it

PR #5915, opened 2023-11-03
Fix drag to empty blocks list

distantnative:
### Fixes
- #5910


bastianallgeier:
It works well for layouts. I tested it first in our blocks field lab setup and it's impossible there to drag a block from one field to another. But I'm not sure if this is even related.

distantnative:
@bastianallgeier thats unrelated, that's still the issue we never resolved here: https://github.com/getkirby/kirby/pull/5362

afbora (approved):

PR #5916, opened 2023-11-03
Tweak style of Toggle input

distantnative:
I think mainly adding the transition makes it feel a lot more like in v3


bastianallgeier:
I sped up the transition a bit. It felt too sluggish to me. I also did not animate the checkmark. It looks weird when it suddenly appears in black to just transform to white quickly afterwards.

bastianallgeier (approved):

PR #5917, opened 2023-11-03
Fix CalendarInput

distantnative:
## This PR …

Guess I found where we didn't want the dropdown to close on click...

TBD if the dropdown should stay open also on input. Right now, it feels better to me if it doesn't close immediately on select

### Fixes
- Calendar input dropdown doesn't close without effect on click
- Calendar input dropdown stays open when selecting a date
- Calendar input dropdown: when clicking today button, also update the dropdown to show that date


luxuryluke:
Yay!

bastianallgeier (approved):

PR #5921, opened 2023-11-04
Activation bubble

bastianallgeier:
## Enhancements

- New activation tooltip and better design for the activation button in the menu.

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -412,9 +412,11 @@
 	"languages.secondary.empty": "There are no secondary languages yet",
 
 	"license": "License",
+	"license.activate": "Activate it now",
 	"license.buy": "Buy a license",
 	"license.register": "Register",
 	"license.manage": "Manage your licenses",
+	"license.ready": "Ready to launch your site?",
 	"license.register.help": "You received your license code after the purchase via email. Please copy and paste it to register.",
 	"license.register.label": "Please enter your license code",
 	"license.register.domain": "Your license will be registered to <strong>{host}</strong>.",
--- a/panel/src/components/View/Menu.vue
+++ b/panel/src/components/View/Menu.vue
@@ -31,27 +31,15 @@
 					class="k-panel-menu-button"
 				/>
 			</menu>
-			<menu v-if="$panel.license === false" class="k-registration">
+			<menu v-if="$panel.license === false">
 				<k-button
+					:text="'Activate'"
+					class="k-activation-button k-panel-menu-button"
 					icon="key"
 					variant="filled"
-					class="k-registration-button k-panel-menu-button"
 					@click="$dialog('registration')"
-				>
-					Activate
-				</k-button>
-
-				<div class="k-registration-bubble">
-					<p>
-						<strong>Ready to launch your site?</strong>
-						<a href="https://getkirby.com/buy" target="_blank">Buy a license</a>
-						and
-						<button type="button" @click="$dialog('registration')">
-							activate it now
-						</button>
-					</p>
-					<k-button class="k-registration-toggle" icon="cancel-small" />
-				</div>
+				/>
+				<k-activation />
 			</menu>
 		</div>
 
@@ -223,6 +211,20 @@ export default {
 	border-end-end-radius: var(--button-rounded);
 }
 
+/* Activation */
+.k-panel-menu .k-activation-button {
+	--button-color-back: var(--color-pink-300);
+	--button-color-text: var(--color-pink-800);
+	border: 1px solid var(--color-pink-400);
+}
+.k-panel-menu .k-activation {
+	display: none;
+	position: absolute;
+	bottom: var(--menu-padding);
+	inset-inline-start: 100%;
+	margin-left: var(--menu-padding);
+}
+
 /* Desktop size */
 @media (min-width: 60rem) {
 	/* The menu is always visible on desktop sizes */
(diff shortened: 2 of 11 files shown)


afbora:
~Why I see double? 👀 😅~

bastianallgeier:
@afbora try to reload the browser. I think your state is still messed up.

i18n/translations/en.json
```diff
@@ -412,9 +412,11 @@
 	"languages.secondary.empty": "There are no secondary languages yet",
 
 	"license": "License",
+	"license.activate": "Activate it now",
 	"license.buy": "Buy a license",
 	"license.register": "Register",
```
lukasbestle, line 417: We still have a lot of "Register" wording in the Panel, especially in the registration dialog but also in the system view (license status). Do we want to bring them in line or is it intentional?

panel/src/components/View/Menu.vue
```diff
@@ -31,6 +31,16 @@
 					class="k-panel-menu-button"
 				/>
 			</menu>
+			<menu v-if="$panel.license === false">
```
lukasbestle, line 34: Probably a bit of an edge case, but in Safari, the optional status bar (disabled by default and if enabled, only displayed when hovering a link for example) is displayed right where the activation menu and bubble is. Don't know if and how we would avoid that as we would need a different/higher position.

Bildschirmfoto 2023-11-04 um 16 10 15

i18n/translations/en.json
bastianallgeier: I think we need to clean this up somehow, but it's maybe too much for this PR?

panel/src/components/View/Menu.vue
bastianallgeier: The only solution for this would be a bottom margin for the last item in the menu, but I think it would look pretty weird.

panel/src/components/View/Menu.vue
```diff
@@ -201,6 +211,20 @@ export default {
...
 }
 
+/* Activation */
+.k-panel-menu .k-activation-button {
+	--button-color-back: var(--color-pink-300);
+	--button-color-text: var(--color-pink-800);
+	border: 1px solid var(--color-pink-400);
+}
+.k-panel-menu .k-activation {
+	display: none;
```
lukasbestle, line 221: On browser sizes smaller than 60rem, the activation bubble is completely hidden. With the hamburger menu hidden, this makes sense, but is there a way to display the bubble when the hamburger menu is active? Or if that won't work layout-wise, maybe fall back to a card?

i18n/translations/en.json
lukasbestle: #5922

panel/src/components/View/Menu.vue
lukasbestle: Yeah, it would need 1.5rem of margin. Or padding for the menu body. I had never really noticed that before, but with the colorful menu button and bubble as the last item, it is much more noticeable than with a text-and-icon-only menu button and feels very weird.

Could we maybe reorder the menu again so it's not last?

bastianallgeier: The beauty of having it as the last item, is the alignment with the notification on the right. This cleans up the bottom part nicely. If we move it up again, it will stick out a lot more annoyingly than it does now and I don't like the idea of that. I would keep it like this tbh.

panel/src/components/View/Menu.vue
bastianallgeier: This is by far the most practical approach to it. We can work on a better responsive state later, but considering the time pressure I don't think this is a detail we need to focus on. The mobile version already has the full-width activation button in the menu

panel/src/components/View/Menu.vue
distantnative: I think it's ok, it's only for those with that floating status bar and only when hovering a link partially overlaid

lukasbestle: It's OK for me, if just slightly annoying. But I guess I and others can get used to it. Please test it though, in the real world it is more weird than in the screenshot because of the visual flickering that it causes.

panel/src/components/View/Menu.vue
distantnative: I agree that we can leave it like this for now. I could see us moving the bubble inside the menu on narrow screens:

Screenshot 2023-11-04 at 17 13 21

panel/src/components/View/Menu.vue
distantnative: https://github.com/getkirby/kirby/assets/3788865/c1215ebb-a916-432e-8a2c-599838fff535

What exactly do you mean by flicker?

Actually, I found it in real even less a problem than from the screenshot. It's the same in Chrome btw. Only that the status bar is black there, not the glass effect. (edit: or I guess white when not using dark mode)

distantnative:
@bastianallgeier I think the markup in the bubble is not quite right as now there is multiple spaces before/after link and button.

panel/src/components/View/Menu.vue
lukasbestle: Ah yes, that's identical as in Chrome (indeed the color is just dark mode).

What I mean with flickering is that the status bar is not there all the time, but keeps appearing and disappearing if you hover and no longer hover something. So in the corner of the eye, it's like a blinking pink box. If it was constantly covered, it would look worse from a design point of view, but it would at least be better from a UX point of view.

distantnative: But is that experience so much different without the activation bubble? What you describe feels to me like the general issue (for some people, for others not) with that status bar not having a fixed place but coming in and out, in and out. To me visually it's not much differently to when it comes in over the white background within that bubble.

lukasbestle: The black status bar over the grey menu, sure. But for me it was new because the light-mode status bar is so subtle that I never noticed that it covered the last menu item. With the pink button and also the bubble with its text that's also cut off, it's a different story.

lukasbestle: So yes, I think my brain just needs to adapt to this and then I'll not notice it. It's going to be the same for our users though, so some kind of solution would help a lot. But if we can't find one that doesn't create other design problems, then it's fine.

distantnative (approved):

Issue #5573, opened 2023-09-06
[v4] Missing selected item indicator in move page dialog


afbora:
## Description

Actually `Blog` item selected but there is no visible indicator

![image (26)](https://github.com/getkirby/kirby/assets/3393422/923c6179-a62f-4101-82bd-444027b92ac5)

## Your setup
4.0.0-beta.1

distantnative:
@bastianallgeier what was the motivation to complete remove the `identifier` prop from PageTree?
Cause I think that broke this - as the move dialog always wants to use the id, even when uuid support is present.

bastianallgeier:
I completely missed this issue. I think the idea was to decide this on the backend instead. But I need to have a closer look.

PR #5927, opened 2023-11-05
Page move dialog: fix currently selected

distantnative:
## This PR …

Since `k-page-tree` only uses UUID now, need to switch over

### Fixes
- #5573

### DIFF
--- a/config/areas/site/dialogs.php
+++ b/config/areas/site/dialogs.php
@@ -523,7 +523,7 @@ return [
 				'props' => [
 					'value' => [
 						'move'   => $page->panel()->url(true),
-						'parent' => $page->parent()?->panel()->url(true) ?? '/site'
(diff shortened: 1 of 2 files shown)


bastianallgeier (commented):
But this would mean that the parent won't work when UUIDs are deactivated.

distantnative:
@bastianallgeier True - but then I think we have to dial back you PageTree changes and add a prop again that decides whether to use id or uuid

bastianallgeier:
I think I found a fix. The item.value is intended to store whatever is being submitted and we can control this pretty easily from the backend: https://github.com/getkirby/kirby/pull/5948

distantnative:
Closing in favor of https://github.com/getkirby/kirby/pull/5948

PR #5937, opened 2023-11-07
New license registration code

bastianallgeier:


src/Cms/System.php
```diff
@@ -286,21 +286,24 @@ public function license(): string|bool
 			$license['date'],
 			$license['email'],
 			$license['domain'],
-			$license['signature']
+			$license['signature'],
+			$license['type'],
```
lukasbestle: I think it would be good to also include the revenue limit in the license data (and JSON for the signature) so we can display it in the license dialog along with an explanation about the license type.

panel/src/components/Dialogs/LicenseDialog.vue
```diff
@@ -0,0 +1,67 @@
...
+						<td class="k-text">
+							<code>{{ license.license }}</code>
+						</td>
+					</tr>
+					<tr>
+						<th>Purchased</th>
+						<td>{{ license.date }}</td>
+					</tr>
+					<tr>
+						<th>Activated</th>
```
lukasbestle:
```suggestion
						<th>First Activated</th>
```

src/Cms/System.php
distantnative: Isn't this implied via the type?

lukasbestle: Then we would need to hardcode the limit in the core though.

bastianallgeier: I think both ways have their problems. If we hardcode it in the license file, we cannot retroactively change it for existing licenses anymore without having wrong info in the dialog. Maybe it makes sense to not show this here, but in the hub?

panel/src/components/Dialogs/LicenseDialog.vue
bastianallgeier: Is "First" really needed here?

src/Cms/System.php
distantnative: Since we don't offer a way (neither require) to switch during the 3 years the license if the conditions change, I think it actually doesn't make too much sense to show them during this time, does it? But only when buying and upgrading.

src/Cms/License.php
```diff
@@ -0,0 +1,431 @@
...
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ */
+class License
+{
+	public const ACTIVE   = 'active';
+	public const EXPIRED  = 'expired';
+	public const INVALID  = 'invalid';
+	public const OUTDATED = 'outdated';
```
distantnative: Maybe you want to use an Enum here?

src/Cms/System.php
lukasbestle: We can never retroactively change the limit for existing licenses. Our customers have a contract with us at the time of their purchase.

The reason why I think it should be in the dialog: So that users know if their license is at all the right one. If a very large business has a Basic license, everyone with access to the system view will be able to see that it's not right, creating pressure to buy the right license in the first place.

panel/src/components/Dialogs/LicenseDialog.vue
lukasbestle: Yes, because otherwise you could be confused that a new activation didn't update the date. For licensing, only the first activation is relevant.

distantnative: Maybe that gets to the root of it: Would people actually activate again when upgrading to additional years? Wouldn't the license code etc. all remain?

src/Cms/License.php
distantnative: https://github.com/getkirby/kirby/pull/5952

panel/src/components/Dialogs/LicenseDialog.vue
lukasbestle: At the moment, an upgrade generates a new license with a new license code. But that is not a necessity.

However in any case, you would need to activate your license again, otherwise the new Included Updates Period won't be in the license file and so Kirby would not be able to know that you are now allowed to use newer Kirby releases.

distantnative: Makes sense, otherwise we would need to communicate with the Hub constantly and we don't want that.

bastianallgeier: I've completely removed it in the latest version. I think the license dialog should be as simple as possible and the hub can still have the info that we really want to show.

i18n/translations/en.json
```diff
@@ -132,8 +133,9 @@
 	"error.layout.validation.block": "There's an error on the \"{field}\" field in block {blockIndex} using the \"{fieldset}\" block type in layout {layoutIndex}",
 	"error.layout.validation.settings": "There's an error in layout {index} settings",
 
-	"error.license.format": "Please enter a valid license key",
+	"error.license.domain": "The domain for the license is missing",
 	"error.license.email": "Please enter a valid email address",
+	"error.license.format": "Please enter a valid license key",
```
lukasbestle: For consistency:

```suggestion
	"error.license.format": "Please enter a valid license code",
```

i18n/translations/en.json
```diff
@@ -416,13 +418,26 @@
...
+	"license.status.legacy.label": "Please renew your license",
+	"license.status.missing.bubble": "Ready to launch your site?",
+	"license.status.missing.info": "No valid license",
+	"license.status.missing.label": "Please activate your license",
 	"license.manage": "Manage your licenses",
-	"license.ready": "Ready to launch your site?",
+	"license.purchased": "Purchased",
 	"license.success": "Thank you for supporting Kirby",
 	"license.unregistered.label": "Unregistered",
+	"license.updates": "Includes free updates until",
```
lukasbestle: I think this was left behind in a refactoring as I cannot find any use of this translation string.

```suggestion
```

src/Cms/License.php
```diff
@@ -0,0 +1,512 @@
...
+	public function __construct(
+		protected string|null $activation = null,
+		protected string|null $code = null,
+		protected string|null $domain = null,
+		protected string|null $email = null,
+		protected string|null $order = null,
+		protected string|null $date = null,
+		protected string|null $signature = null,
+	) {
+		// sanitize the email address
```
lukasbestle:
```suggestion
		// normalize the email address
```

src/Cms/License.php
```diff
@@ -0,0 +1,512 @@
...
+		if (
+			$this->domain !== null &&
+			$this->order !== null &&
+			$this->date !== null &&
+			$this->signature !== null &&
+			$this->hasValidEmailAddress() === true &&
+			$this->type() !== LicenseType::Invalid
+		) {
+			return true;
+		}
```
lukasbestle: Isn't the code also a required component?

src/Cms/License.php
```diff
@@ -0,0 +1,512 @@
...
+
+		return $response->json();
+		// @codeCoverageIgnoreEnd
+	}
+
+	/**
+	 * Prepares the domain to be comparable
+	 */
+	protected function sanitizeDomain(string $domain): string
```
lukasbestle:
```suggestion
	protected function normalizeDomain(string $domain): string
```

src/Cms/License.php
```diff
@@ -0,0 +1,512 @@
...
+	 * Checks if the license is on the correct domain
+	 */
+	public function isOnCorrectDomain(): bool
+	{
+		if ($this->domain === null) {
+			return false;
+		}
+
+		// compare domains
+		if ($this->sanitizeDomain(App::instance()->system()->indexUrl()) !== $this->sanitizeDomain($this->domain)) {
```
lukasbestle:
```suggestion
		if ($this->normalizeDomain(App::instance()->system()->indexUrl()) !== $this->normalizeDomain($this->domain)) {
```

src/Cms/License.php
```diff
@@ -0,0 +1,512 @@
...
+
+		// verify the license signature
+		$data      = json_encode($this->signatureData());
+		$signature = hex2bin($this->signature);
+
+		if (openssl_verify($data, $signature, $pubKey, 'RSA-SHA256') !== 1) {
+			return false;
+		}
+
+		return true;
```
lukasbestle:
```suggestion
		return openssl_verify($data, $signature, $pubKey, 'RSA-SHA256') === 1;
```

src/Cms/License.php
```diff
@@ -0,0 +1,512 @@
...
+	/**
+	 * Returns the renewal date
+	 */
+	public function renewal(string|IntlDateFormatter|null $format = null): int|string|null
+	{
+		if ($this->activation === null) {
+			return null;
+		}
+
+		$time = strtotime('+3 years', $this->activation());
```
lukasbestle, line 364: Maybe it would make sense to add the renewal date to the license data. Otherwise we could never change the duration in the future because old versions would still have the hardcoded three years.

src/Cms/LicenseStatus.php
```diff
@@ -0,0 +1,133 @@
...
+	 */
+	case Active = 'active';
+
+	/**
+	 * Only used for the demo instance
+	 */
+	case Demo = 'demo';
+
+	/**
+	 * The free feature period of
```
lukasbestle:
```suggestion
	 * The included updates period of
```

src/Cms/LicenseStatus.php
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	/**
+	 * The info text is shown in the license dialog
+     * in the status row.
+	 */
+	public function info(string|null $end = null): string
+	{
+		return match ($this) {
+			static::Demo => 'This is a demo installation',
```
lukasbestle: Couldn't we also use translation keys for the demo mode strings? If a user of our demo changes the user language, this will then update as well.

src/Cms/License.php
bastianallgeier: If the code isn't there, the license type will be invalid. But I think we could already stop earlier here by actually checking for an empty code.

lukasbestle: Ah, alright. It confused me because it was not immediately obvious. The `hasValidEmailAddress` check was easier to spot. ;)

src/Cms/License.php
bastianallgeier: I generally agree, but looking at how brittle it is to get the verification right on both ends, every bit of information that does not have to be in the license file is better for us. I think we can easily solve this in later releases.

lukasbestle: You are right. I also went through the possible cases in my head and actually there is no case where we the hardcoded period in an older Kirby version could break the licensing. We can only change the duration for newly purchased licenses and the renewal date will always be after the purchase. So A Kirby version released before any change to the duration of the period won't care about the change.

And in the worst case, we could change the license code prefix, which will prevent activation in older releases that don't yet have the required change.

Issue #5941, opened 2023-11-09
[v4] Structure field: pagination position broken


afbora:
## Description

![screen-capture-1732-Test - Mægazine-localhost](https://github.com/getkirby/kirby/assets/3393422/d41a17fd-1106-421c-aecf-fa22700c604c)

## Your setup
**Kirby Version**  
Beta.3

bastianallgeier:
@afbora could you try to fix it? I think it would be better to put the pagination below the table and not try to have it in the tfoot

afbora:
Already not in `<table>`.

````html
<div class="k-table">
  <table></table>
  <div class="pagination"></div>
</div>

<footer>
  +
</footer>
````

 I've no idea about correct solution. I think pagination should be in left or right of plus icon. Something like following. But table and footer is in separated components.

![screen-capture-1733-Test - Mægazine-localhost](https://github.com/getkirby/kirby/assets/3393422/7334d2b0-b8f2-48c6-bb9f-1ec41834b88c)

PR #5951, opened 2023-11-10
Fix structure field footer

afbora:
## This PR …
@bastianallgeier That's all I could do. I tried to use slot in `k-table` component and css grid for the UI. Feel free to close the PR if not acceptable for you 👍 

screen-capture-1735-Test - Mægazine-localhost

But still there is a minor issue that I couldn't fix it 🙈 If only plus icon available, it should be centered but not for this PR 🙁 Would be nice if you help about that.

screen-capture-1736-Test - Mægazine-localhost

### Fixes
- Structure field: pagination position broken #5941

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/lab/components/tables/1_horizontal/index.vue
+++ b/panel/lab/components/tables/1_horizontal/index.vue
@@ -125,7 +125,7 @@
 						:handle="true"
 						:options="{
 							fallbackClass: 'k-table-row-fallback',
-							ghostClass: 'k-table-row-ghost',
+							ghostClass: 'k-table-row-ghost'
 						}"
 						element="tbody"
 					>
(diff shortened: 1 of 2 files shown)


distantnative:
@bastianallgeier @afbora are we sure we want to move the pagination outside the table (instead of being its footer)? I think we did the originally exactly because it starts looking quite weird to be on one side of a centre add button etc.

bastianallgeier:
If we put it inside the table we need to adjust the styling pretty heavily. I would put the navigation on the right and the add button on the left. At least for 4.0

bastianallgeier:
Damn, I missed that the pagination is baked into the table component. This is why this is difficult and also actually easier to really overwrite the styles of the pagination instead. I'm sorry for missing that. I've added the new styles and it works well now without too many additional overwrites. 

Screenshot 2023-11-13 at 16 40 47

bastianallgeier (approved):

PR #5958, opened 2023-11-14
3.9.8

bastianallgeier:


bastianallgeier:
It's super weird. When I go to the details of the checks, they all pass, but here the E2E tests are still blocked. I would ignore it.

distantnative (approved):

lukasbestle:
@bastianallgeier The E2E check was still marked as required in the branch protection rule. I've removed it.

To explain: The required checks are just what we defined GitHub to expect as successful CI runs before a PR can be merged into the protected branch. Since we removed the E2E check from the workflow, this check can never run, but GitHub still expected and required it because we told it to do so.

PR #5971, opened 2023-11-18
Add Vue target version to jsconfig

lukasbestle:
The Vue Language Features extension for VSCode has complained that the installed Vue version does not match the configured one. This change fixes it, but I have not done any research if there are any other implications. Please do before merging.


distantnative (approved):
Haven't had that issue - but also don't see any problem with adding this config.

PR #5983, opened 2023-11-21
ImageMagick: fix focus-cropping bug

### LINKED ISSUE #5982: [v4] Error when focus-cropping an image that's already in the correct ratio in ImageMagick
## Description
I replaced some images and then got this error:
Screenshot 2023-11-21 at 13 58 51

The images are cropped like this:
```php
$image->crop(400, 400, ['crop' => $image->focus()]);
```

The previous images were of arbitrary size and the replaced images were already prepared in the final aspect ratio.
I suspect that this is the culprit:

Screenshot 2023-11-21 at 14 00 24

## To reproduce

You have to call crop with the focus option on an image in the same aspect ratio the image already has. `Focus::coords()` will return `null` and this will break the ImageMagick.php.

## The fix

Moving `$focus = Focus::coords()` into an `if` construct like you already do it in GdLib.php fixes it:

Screenshot 2023-11-21 at 14 07 29

Almost:

Screenshot 2023-11-21 at 14 16 23

We need to fall back to a supported crop gravity when not using the focus coords:

Screenshot 2023-11-21 at 14 18 07

mrflix:
Focus::coords returns null when the source ratio is the same as the crop ratio. Adding a check for that case.

fixes #5982 

## Ready?

### For review team


distantnative:
Thanks for the PR!

To spare ourselves the else part, how about we transform the whole section below into

```php
// translate the gravity option into something imagemagick understands
$gravity = match ($options['crop'] ?? null) {
	'top left'     => 'NorthWest',
	'top'          => 'North',
	'top right'    => 'NorthEast',
	'left'         => 'West',
	'right'        => 'East',
	'bottom left'  => 'SouthWest',
	'bottom'       => 'South',
	'bottom right' => 'SouthEast',
	default        => 'Center'
};
```

mrflix:
TIL about [match](https://www.php.net/manual/en/control-structures.match.php). I adjusted my commit. By force. Oops.  Here is a new PR: https://github.com/getkirby/kirby/pull/5985

distantnative:
@mrflix thanks! And you're in for a treat. Match is my favorite PHP 8 feature maybe.

PR #5984, opened 2023-11-21
Upgrade to Vite 5

distantnative:
- Upgrades to vite 5
- Manually includes tailored prismjs
- Remove lightning's again (much larger bundle size with it)


distantnative:
Ok the bundle sizes are very different, so I think it somehow slimmed prismjs down.
Unfortunately the plugin isn't maintained anymore and clashes with Vite 5 moving to ESM for the config file.

bastianallgeier:
Is it worth to upgrade at all before the launch?

distantnative:
No. Let's do it for 4.1 - issue/questions remain though

bastianallgeier:
When I installed the plugin, it was part of the installation process to put it in the vite config. I honestly never thought about what it does exactly.

bastianallgeier (approved):

PR #5991, opened 2023-11-23
Slug field: respect custom slug allow from `Str`

### LINKED ISSUE #5929: [v4] Panel unaware of custom slug rules
## Description
Kirby allows for the creation of custom slugs rules by setting the static property of the `Str` class:

```php
Str::$defaults['slug']['allowed'] = 'my rules here';
```

These settings are not passed to the `create` and `changeTitle` dialogs.

**Expected behavior**  
The panel should respect the general slug settings.

## To reproduce
Set custom rules in a page model or the config, create a page or update a title. The slug will always be created based on the default settings.

**Additional context**  
Passing `'allow' => Str::$defaults['slug']['allowed']` to the slug field will fix this issue, see here:

- https://github.com/getkirby/kirby/blob/v4/develop/src/Panel/PageCreateDialog.php#L116-L120
- https://github.com/getkirby/kirby/blob/v4/develop/config/areas/site/dialogs.php#L212-L221

distantnative:
### Fixes
- Slug field respects custom allowed slug characters from `Str::$defaults['slug']['allowed']`

### Breaking changes
- `$helper.string.slug`: the `allow` parameter now defines the whole set of allowed characters, not just the characters in addition to `a-z0-9`

### DIFF
--- a/panel/src/helpers/string.js
+++ b/panel/src/helpers/string.js
@@ -161,16 +161,15 @@ export function rtrim(string = "", replace = "") {
  * Convert string to ASCII slug
  * @param {string} string string to be converted
  * @param {array} rules ruleset to convert non-ASCII characters
- * @param {array} allowed list of allowed non-ASCII characters
- * @param {string} separator character used to replace e.g. spaces
+ * @param {array} allowed list of allowed characters (default: a-z0-9)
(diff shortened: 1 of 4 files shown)


bastianallgeier (approved):

nilshoerrmann:
Cool – thanks!

Issue #5993, opened 2023-11-23
Panel: can't read ellipsed labels


rasteiner:
## Description
When the label of a field or section is cut off by an ellipsis, there's no way to see the complete label (except maybe making the window smaller until the layout wraps to a single column). 

**Expected behavior**  
I expected something like a tooltip to show up un mouse hover. Maybe via `title` attribute.

**Screenshots**  
![image](https://github.com/getkirby/kirby/assets/6684137/13054a39-8eb6-4c42-acbc-97e381d895bd)

## Your setup
**Kirby Version**  
4.0.0 rc 2

**Additional context**  
According to accessibility people `title` shouldn't contain anything of vital information, as it normally requires a mouse cursor to be seen. On the other hand, according to accessibility people, `text-overflow: ellipsis` shouldn't even exist. So I guess `title`, while not perfect, would make it accessible at least to some people?  (a "better than nothing" approach)

Of course the most accessible thing to do would be wrapping the text, no ellipse... Maybe once `display: subgrid` has acceptable support we could keep stuff aligned even with multi-line labels. 🤔

PR #5994, opened 2023-11-23
Fix header wrapping for fields and sections

### LINKED ISSUE #5866: [v4] Field label overflows and icons stack vertically
## Description
Screenshot should be self explainatory. The field's label doesn't wrap or gets trimmed and as a result first it spills outside of the field container and second it forces the icons to stack vertically.

**Expected behavior**  
The text should either wrap on a new line or have some text overflow in place and the two icons for the file upload/select should stay horizontal.

**Screenshots**  
Screenshot 2023-10-24 at 10 04 29 AM

Screenshot is from latest firefox. Latest Safari does have text overflow in place with ellipses but still suffers from the vertical stacking of the icons.

## To reproduce
Create a file field with a very, very long title.

## Your setup
**Kirby Version**  
Kirby 4 - Beta 2

distantnative:
### Fixes
- #5866
- #5993

### DIFF
--- a/panel/lab/components/field/index.vue
+++ b/panel/lab/components/field/index.vue
@@ -27,6 +27,17 @@
 				<k-input :value="value" id="d" type="text" @input="value = $event" />
 			</k-field>
 		</k-lab-example>
+		<k-lab-example label="Options Slot & excessive label">
+			<k-field
+				input="d"
+				label="This is a very very long label that could push the options aside and break stuff"
+			>
+				<k-button icon="add" size="xs" slot="options" variant="filled">
+					Add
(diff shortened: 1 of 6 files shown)


bastianallgeier:
I had to set a min-width on the label to really get truncation working, but now it's fine. 

I've set up a few more lab examples for the field and new ones for k-section to test this more easily. That's when I discovered that section and field components have quite some inconsistencies. So I fixed those in k-section which aren't too massive for rc.3. @distantnative could you still have another look please?

bastianallgeier (approved):

PR #5996, opened 2023-11-24
Models fields: add disable style

### LINKED ISSUE #5959: Pages Field with `translate: false` is not dimming appearance
## Description
A `pages-field` with `translate:false` does not change it's appearance, if i switch to a non-default-language,
like other fields does. it only disables the select-button, but appears translatable.

Similar to #5957 

Expected behavior
The `Pages-Field` should dim like other fields does, wenn `translate: false` and you select a non-default-language

**Screenshots**  
image

## To reproduce
add a `pages-field` and set `translate: false` and switch to a non-default language

## Your Setup
Kirby Version
k4-rc.1

Your system (please complete the following information)

Device: macbook
OS: macOS, 13.4.1
Browser: Chrome, Safari, FF

distantnative:
### Fixes
- #5959

Screenshot 2023-11-24 at 10 02 59

### DIFF
--- a/panel/lab/components/item/1_list/index.vue
+++ b/panel/lab/components/item/1_list/index.vue
@@ -69,5 +69,13 @@
 				width="1/2"
 			/>
 		</k-lab-example>
+		<k-lab-example label="Theme: disabled">
+			<k-item
+				:image="{ src: 'https://picsum.photos/800/600' }"
+				info="With some info"
+				text="This is a nice item"
+				theme="disabled"
+			/>
+		</k-lab-example>
 	</k-lab-examples>
 </template>
--- a/panel/lab/components/item/2_cards/index.vue
+++ b/panel/lab/components/item/2_cards/index.vue
@@ -73,5 +73,14 @@
 				width="1/2"
 			/>
 		</k-lab-example>
+		<k-lab-example label="Theme: disabled">
+			<k-item
+				:image="{ src: 'https://picsum.photos/800/600' }"
+				info="With some info"
+				layout="cards"
+				text="This is a nice item"
+				theme="disabled"
+			/>
+		</k-lab-example>
 	</k-lab-examples>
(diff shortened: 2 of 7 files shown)


bastianallgeier:
I'm not sure about this one. We also don't dim page and files sections because you can still click through them to arrive on the page or file.

distantnative:
That's true - but these are fields and I have to agree with the issue author that it's quite weird that all feels get a dimmed style when they are disabled, just these stroll along as if they aren't disabled.

It also makes it harder to understand the differentiation fields vs. sections when those fields behave like the odd ones among the fields.

bastianallgeier:
Yeah, it's tricky. This is also about the fields vs. section dilemma that we often face. I think it makes more sense for the field but not for the section. Let's merge this.

bastianallgeier (approved):

bastianallgeier:
But before we merge: I think it would be good to introduce this as a proper state for items instead. This could be useful in other places as well.

distantnative:
@bastianallgeier will make it a state of k-item

And agreed, sections should stay the same because sections don't have this relation to disabled (e.g. because of translation, update permission etc)

distantnative:
@bastianallgeier would you call it theme then? Cause I think disabled in the context of k-item is wrong. The item isn't disabled, this would imply we should also disable link etc. - it's just shown visually different in a disabled field.

bastianallgeier:
Agreed, I think theme makes sense.

distantnative:
@bastianallgeier: added a version

bastianallgeier (approved):
Lovely

dennisbaum:
Thank you guys! In our opinion, it is exactly like nico said - and we are aware of the behavior, that you can still click the page or file. but that way it looks now, editors can understand, that that fields aren't translatable in a non default language. Tnx again!

PR #6000, opened 2023-11-24
`Xml::attr()`: Support empty string as empty value

lukasbestle:
### Enhancements

- `Xml::attr()`: Support passing an empty string as value to generate an attribute with an empty value

### Deprecated

- Passing a single space as value to `Xml::attr()` (with the intention to generate an attribute with an empty value) has been deprecated in favor of passing an empty string.

### Breaking changes

- Passing an empty string as value to `Xml::attr()` no longer omits the attribute but generates an attribute with an empty value.

## Ready?

### For review team

### DIFF
--- a/config/tags.php
+++ b/config/tags.php
@@ -113,7 +113,7 @@ return [
 		'html' => function (KirbyTag $tag): string {
 			if ($tag->file = $tag->file($tag->value)) {
 				$tag->src       = $tag->file->url();
-				$tag->alt     ??= $tag->file->alt()->or(' ')->value();
+				$tag->alt     ??= $tag->file->alt()->or('')->value();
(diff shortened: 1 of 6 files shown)


distantnative:
@lukasbestle Would it be an option instead of deprecating the single space to use trim? Then it doesn't matter whether empty string, single space...

lukasbestle:
Hm, I don't understand how trim could help here TBH.

The single space was a hack because the empty string was already taken for "no attribute". So there will still be uses with a single space where the dev actually intended to generate an empty attribute but had to use the hack.

Do you have an example how it could work with trim?

distantnative:
Ok yea forget it

distantnative (approved):

PR #6007, opened 2023-11-27
Slug field: respect custom slug allow from `Str`

distantnative:
Original PR https://github.com/getkirby/kirby/pull/5991

This time, instead of adding the default in the parameter definition, we check also for e.g. empty string, arrays and set the default.

### DIFF
--- a/panel/src/helpers/string.js
+++ b/panel/src/helpers/string.js
@@ -161,8 +161,8 @@ export function rtrim(string = "", replace = "") {
  * Convert string to ASCII slug
  * @param {string} string string to be converted
  * @param {array} rules ruleset to convert non-ASCII characters
- * @param {array} allowed list of allowed non-ASCII characters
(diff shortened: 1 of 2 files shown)


bastianallgeier:
My tests all worked fine so far.

bastianallgeier (approved):

nilshoerrmann:
I'm trying this out by adding `Str::$defaults['slug']['allowed'] = 'a-z0-9.';` to a page model and it doesn't work. Is it possible that you forgot to add the `allow` prop to the slug field so it always falls back to the default? In the previous PR you added this which is missing in the released version: https://github.com/getkirby/kirby/pull/5991/files#diff-96fa15f1e6b351775f11b5fb887a7c98b34e83d1ca3167ff234d186489f5f1a9R226

nilshoerrmann:
Okay, I can confirm that adding that missing line fixes the issue. For clarity, I created a new issue #6048.

PR #6029, opened 2023-11-30
`Panel\Model::image()`: fix `false` from blueprint

### LINKED ISSUE #6020: Broken page list in "move page" and "select page" widgets (server error)
## Description

I’m testing an update of a 3.9 site to 4.0 and it looks like the page list widget used in a few places, namely in the "Move page" dialog and in the page selection widget of the [Retour for Kirby plugin](https://github.com/distantnative/retour-for-kirby), breaks with a server error:

```json
{
    "code": 500,
    "error": "array_merge(): Argument #3 must be of type array, bool given"
}
```

Some URLs for GET requests from the Panel that show this exact error:

- `http://localhost:8000/panel/site/tree?parent=%2Fsite`
- `http://localhost:8000/panel/site/tree?move=%2Fpages%2Fcontact&parent=%2Fsite`

**Expected behavior**  

The request to `/panel/site/tree` doesn't fail.

**Screenshots**  

Screenshot 2023-11-29 at 11 12 13

## To reproduce

I don't have steps to reproduce or a minimal reproduction yet.

I understand this is very little information to go with, and I suspect it's something in my setup (otherwise there would be issues for this filed already). I’m not confident I can make a simple repro through trial and error.

Any pointers on how I could investigate this?

I have `'debug' => true, 'whoops' => true` in my config, but those Panel API requests don't show more information about where the issue occurred.

## Your setup

**Kirby Version**  

4.0.0

**Console output**  
See above for the failing API request. There's another JS error after that, which I suspect comes from not handling the error response from the API when rendering the UI.

 - Device: M1 Mac
 - OS: macOS
 - Browser: Chrome
 - Version: 119

distantnative:
## This PR …

For next minor release as side effects aren't as easy to oversee as that it would qualify for patch release.

### Fixes
- Fix handling `image: false` in blueprint for Panel preview images  #6020

## Ready?

### For review team

- [x] ˜˜Add to [website docs release checklist](https://github.com/getkirby/getkirby.com/pulls) (if needed)˜˜

### DIFF
--- a/src/Panel/Model.php
+++ b/src/Panel/Model.php
@@ -103,6 +103,18 @@ abstract class Model
 			return null;
 		}
 
+		// switched off from blueprint,
+		// only if not overwritten by $settings
+		$blueprint = $this->model->blueprint()->image();
+
+		if ($blueprint === false) {
+			if (empty($settings) === true) {
+				return null;
+			}
+
+			$blueprint = null;
+		}
+
 		// skip image thumbnail if option
 		// is explicitly set to show the icon
 		if ($settings === 'icon') {
@@ -116,7 +128,7 @@ abstract class Model
 		$settings = array_merge(
 			$this->imageDefaults(),
 			$settings ?? [],
-			$this->model->blueprint()->image() ?? [],
+			$blueprint ?? [],
 		);
 
 		if ($image = $this->imageSource($settings['query'] ?? null)) {
--- a/tests/Panel/ModelTest.php
+++ b/tests/Panel/ModelTest.php
@@ -386,6 +386,38 @@ class ModelTest extends TestCase
 	/**
 	 * @covers ::image
 	 */
+	public function testImageWithBlueprintFalse()
+	{
+		$app  = $this->app->clone([
+			'blueprints' => [
+				'pages/foo' => [
+					'image' => false
+				]
+			],
+			'site' => [
+				'children' => [
+					[
+						'slug' => 'test',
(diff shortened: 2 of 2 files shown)


src/Panel/Model.php
```diff
@@ -116,7 +128,7 @@ public function image(
 		$settings = array_merge(
 			$this->imageDefaults(),
 			$settings ?? [],
-			$this->model->blueprint()->image() ?? [],
+			$blueprint ?? [],
```
fvsch, line 131: A possible shorter fix would be to discard all falsy values with:

```php
$this->model->blueprint()->image() ?: [],
```

distantnative: But then `image: false` wouldn't have any effect at all, would it? I think the intended logic would be that if no settings are set and blueprint says false, return null (no image). And not to use the defaults.

fvsch: I think it depends if you want to consider that `image: false` is incorrect input that should be ignored, or a new feature you want to add.

(Personally I don't think I have a use case for `image: false` disabling the default image, at least not currently. I did have `image: false` in one of my blueprints, but I think that was just leftover cruft from some trial-and-error when originally writing those blueprints with Kirby 3.9.)

distantnative: I think it would be more consistent in the handling, where you can define a general default in the model blueprint in general but this can be overwritten by the option in the section definition.

bastianallgeier (approved):

PR #6031, opened 2023-11-30
Fixes the bottom margin of the pagination in a models dialog

### LINKED ISSUE #6010: [v4] Pagination UI issue in picker dialogs
image (1)

Kirby 4.0.0-rc.4
PHP 8.1.23

bastianallgeier:
### Fixes

- Fixes the bottom margin of the pagination in a models dialog #6010

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Dialogs/ModelsDialog.vue
+++ b/panel/src/components/Dialogs/ModelsDialog.vue
@@ -165,7 +165,7 @@ export default {
 .k-models-dialog .k-list-item {
 	cursor: pointer;
 }
-.k-models-dialog .k-collection-footer {
-	margin-bottom: var(--spacing-6);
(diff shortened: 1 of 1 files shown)


distantnative:
Shouldn't we rather remove https://github.com/getkirby/kirby/blob/main/panel/src/components/Dialogs/Dialog.vue ? Or can we add there a comment what the negative margin is needed for? Looking at it and I can't relate it to anything.

bastianallgeier:
I was thinking about removing the block as well. I'm pretty sure it's old and no longer needed. But that is probably not what we should do in a patch, right? I changed the fix for the margin a bit though.

distantnative:
True for a patch this is safer.

distantnative (approved):

PR #6042, opened 2023-11-30
Update fr.json

LaurentTacco:
- Add missing word
- Replace remaining prime symbols (') by apostrophes (’)


lukasbestle:
As Bastian wrote on Discord, we unfortunately need to close this because any changes in the repo would be overwritten by Transifex.

PR #6056, opened 2023-12-09
Color field: options query support

### LINKED ISSUE #6053: Color field missing ability to use queried options
## Description
Received permission from Bastian to add this as an issue.

Trying to move from a 3rd-party color field to the new native field. Currently, we pull the list of default colors available from the config for each site. The problem is that the new Color field doesn't support query language for pulling in options.

### Expected behavior
Expected the ability to pull in a list of colors from the config to my reusable field.

#### Config
```php
    return [
        'customizable' => [
            'palette-options' => [
                'Sunny rays'       => '#F8B195',
                'First-love blush' => '#F67280',
                'Cherry blossom'   => '#C06C84',
                'Morning gloom'    => '#6C5B7B',
                'Midnight rain'    => '#355C7D'
            ]
        ]
    ]
```

#### Field (color-palette.yml)
```yaml
label: Color
type: color
options:
  type: query
  query: kirby.option('customizable.palette-options')
  key: "{{ arrayItem.key }}"
  value: "{{ arrayItem.value }}"
width: 1/2
```

## Your setup
**Kirby Version**  
Kirby 4.0.1

distantnative:
### Features
- Color field: query and API support for `options`

```yml
myColorField:
  type: color
  options:
    type: query
    query: kirby.option('my.colors')
```

When passing an options array, keep in mind that keys will be used as the color values and the corresponding array value as preview text (for associative arrays, simple arrays are all color values only).

## Ready?

### For review team

### DIFF
--- a/config/fields/color.php
+++ b/config/fields/color.php
@@ -1,6 +1,7 @@
 <?php
 
 use Kirby\Exception\InvalidArgumentException;
+use Kirby\Field\FieldOptions;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\Str;
 
@@ -52,9 +53,18 @@ return [
 			return Str::lower($this->default);
 		},
 		'options' => function (): array {
-			return A::map(array_keys($this->options), fn ($key) => [
-				'value' => $this->options[$key],
-				'text'  => is_string($key) ? $key : null
+			// resolve options to support manual arrays
+			// alongside api and query options
+			$props   = FieldOptions::polyfill($this->props);
+			$options = FieldOptions::factory($props['options']);
+			$options = $options->render($this->model());
+
+			// flip value and text as the notation of the
+			// color options is `text: value`;
+			// only add text if it is different from value
+			return A::map($options, fn ($option) => [
+				'value' => $option['text'],
+				'text'  => $option['value'] !== $option['text'] ? $option['value'] : null
 			]);
 		}
 	],


lukasbestle:
@distantnative Could you please add examples to the XSS lab in the sandbox (basically copying the examples for the other options fields) to ensure that we never accidentally introduce XSS? Options fields are always a bit tricky here.

bastianallgeier:
Something is weird. I cannot get a single setup in the lab running for some reason. @distantnative Could you post your test blueprint setup here?

distantnative:
@bastianallgeier that's already included in the first post here. use it together with e.g. this config option

```php
	'my' => [
		'colors' => [
			'#3e3e3e',
			'#aaa',
			'#ddd',
		]
	]
```

bastianallgeier:
So weird. I must have had a typo somewhere. I tried the blueprint and added my own config options for it, but it didn't work the first time. Now it's all fine.

bastianallgeier (approved):

lukasbestle:
@distantnative I have added the various "options" configurations for the color field to our XSS sandbox and discovered some bugs:

## Actual bug

- I couldn't get the field to use an associative array from a query. So if I modify [your example here](https://github.com/getkirby/kirby/pull/6056#issuecomment-1858786851) to use keys with labels, those labels aren't used and just the color value is taken. If I do it like you write in the changelog snippet ("When passing an options array, keep in mind that keys will be used as the color values and the corresponding array value as preview text"), the field doesn't render at all.
  **To reproduce:** "Color (query, defaults)" field in the XSS sandbox

## Consistency issues compared to other options fields

- String templates are not parsed in hardcoded options. They weren't before this PR and it doesn't change anything about hardcoded options, but I think it's confusing that color is now an options field but without the same options support in the blueprint.
  **To reproduce:** "Color (hardcoded 1)" to "Color (hardcoded 3)" fields in the XSS sandbox
- Text of hardcoded options is not rendered as HTML; all other options fields except select (which cannot render HTML) do this
  **To reproduce:** "Color (hardcoded 1)" field in the XSS sandbox
- The `{< >}` placeholders are not supported for options from query and API, instead they render a blank label; again a consistency issue with the other options fields
  **To reproduce:** "Color (API, with HTML)" and "Color (query, custom HTML query)" fields in the XSS sandbox

---

I think it would also be good to add PHPUnit tests for the new options features to the `ColorFieldTest`.

distantnative:
I'll try to take a look at the examples to better understand your comments.

I think the first one is that if your return direct PHP for an option this has to be already in the correct format (as with other parts in Kirby, e. g. programmable blueprints). So no magic rewriting form associative array. But I'll need to have a closer look on that.

The second part I don't quite understand. Where and why should the color field now render HTML?

lukasbestle:
> I think the first one is that if your return direct PHP for an option this has to be already in the correct format (as with other parts in Kirby, e. g. programmable blueprints). So no magic rewriting form associative array. But I'll need to have a closer look on that.

Hm, then maybe I misunderstood the release notes snippet from the first post. I've also tried returning `['text' => '#aaa', 'value' => 'Some label']` as an option, but the result was the same (color is used, but label is ignored). Am I doing it incorrectly?

> The second part I don't quite understand. Where and why should the color field now render HTML?

The labels for all options fields except `select` support HTML rendering. Given that the color field is now essentially an options field, I think it would make sense to also support it here for consistency.

distantnative:
I think we still need to look at the context of the field and still dont see why and how HTML would make sense for the color field here, similarly to how it's not for the select field.

lukasbestle:
The only reason why the select field does not support HTML is that the `<select>` input cannot render HTML. But the label in the color input can, just like the other options fields.

lukasbestle:
Another thing I thought about is the flipped text and value props. With an associative array those allow to use the same key/value structure as with the hardcoded options, but with more complex data or customized queries it's IMO really confusing that the color code is the `text` and the label the `value`.

Maybe it was a mistake to make the labels the keys in the hardcoded options. Was there a technical reason why we did it that way and not like `"#aaa": Label`?

distantnative:
I also wondered if that was a mistake, but I don't think we could easily change this now, or do you have an idea?

Let's continue here: https://github.com/getkirby/kirby/issues/6101 
I easily overlook discussions in merged PRs/closed issues

PR #6059, opened 2023-12-09
PHPUnit 10

distantnative:
Some of the changes you see are because with PHPUnit 9 PHP warnings would be converted to Exceptions. So falsely, we put some throw-catch logic in places to deal with though. But in reality those were just warnings that would not be matched (and it showed now with PHPUnit 10), so we need to refactor some smaller bits here and there.

### Features
- New `A::implode()` method that also handles nested arrays

### Fixed
- Fixed PHP warning for structure fields without subfields
- Fixed PHP warning for `F::size()` with non-existing file
- Fixed PHP warning for unserializing `Session` data
- Fixed PHP warning for `V::message()` with nested arrays

### Housekeeping
- Upgraded unit testing to PHPUnit 10


src/Toolkit/A.php
```diff
@@ -351,6 +351,31 @@ public static function has(
...
+	): string {
+		$result = '';
+
+		foreach ($array as $value) {
+			if (empty($result) === false) {
+				$result .= $separator;
+			}
+
+			if (is_array($value) === true) {
+				$result .= '[' . static::implode($value, $separator) . ']';
```
lukasbestle: I'm not sure about the bracket syntax TBH. It shifts the method from a pure implode method somewhat to a serialization method. From the name I'd expect to get a linear list of values. That would be achieved by:

```suggestion
				$result .= static::implode($value, $separator);
```

Would that be a problem for the validation use case where you use the new implode method?

src/Toolkit/A.php
```diff
@@ -351,6 +351,31 @@ public static function has(
 		return in_array($value, $array, $strict);
 	}
 
+	/**
+	 * Join array elements as a string,
+	 * also supporting nested arrays
+	 */
+	public static function implode(
+		array $array,
+		string $separator = ', '
```
lukasbestle: For consistency with the core PHP `implode()` function:

```suggestion
		string $separator = ''
```

lukasbestle (commented):
Awesome, nice work. 👍

Love that it uncovered core issues. Could you please add those to the release note snippet as well?

src/Toolkit/A.php
distantnative: It could be a bit confusing, but it's not a real-world problem anyway. So I'm fine either way.

lukasbestle (dismissed):
I've updated the PHPUnit config with some improvements from the defaults. @distantnative Could you please take a look if the changes make sense?

PR #6072, opened 2023-12-15
Don’t pass kirby as property to models

### LINKED ISSUE #6061: [K4] Session: need to serialize data. Regression?
## Description
When I set a value in Session which is not serialize, I get an error:

`Uncaught Exception: Serialization of ‘Closure’ is not allowed in /home/…/kirby/src/Session/Session.php:372`

I didn't get this error with Kirby 3.*

## To reproduce
Run this code in a controllers or templates:

`$session = $kirby->session(); `
`$session->set( 'title', $page->title() );`
`$session->set( 'pages', $site->children()->unlisted() );`

## Your setup
Kirby 4.0.1

bastianallgeier:
## This PR …

With the refactoring of property setters in models in v4, Kirby got passed as an instance to `propertyData['kirby']` This broke serializing and could have other side effects. I wonder if some whoops issues are related to it as well. 

I've tracked down all the parts where we pass Kirby to a model. This should have no side-effects in theory. The models use static::$kirby as property and don't fetch the Kirby instance from the propertyData array anyway. But it remains a bit risk. 

### Fixes

- Objects are serializable again https://github.com/getkirby/kirby/issues/6061

### Breaking changes

- Hopefully none

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1482,9 +1482,7 @@ class App
 	protected function setSite(Site|array $site = null): static
 	{
 		if (is_array($site) === true) {
-			$site = new Site($site + [
-				'kirby' => $this
-			]);
+			$site = new Site($site);
 		}
 
 		$this->site = $site;
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1499,7 +1499,6 @@ class App
 		return $this->site ??= new Site([
 			'errorPageId' => $this->options['error'] ?? 'error',
 			'homePageId'  => $this->options['home']  ?? 'home',
-			'kirby'       => $this,
 			'url'         => $this->url('index'),
 		]);
 	}
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1466,9 +1466,7 @@ class App
 	protected function setRoles(array $roles = null): static
 	{
 		if ($roles !== null) {
-			$this->roles = Roles::factory($roles, [
-				'kirby' => $this
-			]);
+			$this->roles = Roles::factory($roles);
 		}
 
 		return $this;
--- a/src/Cms/AppUsers.php
+++ b/src/Cms/AppUsers.php
@@ -86,9 +86,7 @@ trait AppUsers
 	protected function setUsers(array $users = null): static
 	{
 		if ($users !== null) {
-			$this->users = Users::factory($users, [
-				'kirby' => $this
-			]);
+			$this->users = Users::factory($users);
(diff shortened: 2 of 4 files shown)


lukasbestle (dismissed):
Passing the app object as property indeed seems superfluous as it isn't used anywhere at the moment. I can't find any reason why this would break anything.

The question is what our future plans with object coupling are. I think in general it would be desirable to avoid using global objects wherever possible, e.g. when using Kirby on the CLI with multiple independent app objects. But that's definitely a question for the future, we can still add the app objects as props back in later.

What we could also consider in the future is to implement custom serialization logic with the `Serializable` interface. It really doesn't make much sense to serialize all property data with every object.

afbora:
Not sure but I think also you need to check out following lines: (check `serialize($kirby->user())` to test it)
- https://github.com/getkirby/kirby/blob/fix/6061-serializable-objects/src/Cms/AppUsers.php#L90
- https://github.com/getkirby/kirby/blob/fix/6061-serializable-objects/src/Cms/AppUsers.php#L131

src/Cms/App.php
```diff
@@ -1499,7 +1499,6 @@ public function site(): Site
 		return $this->site ??= new Site([
 			'errorPageId' => $this->options['error'] ?? 'error',
 			'homePageId'  => $this->options['home']  ?? 'home',
-			'kirby'       => $this,
```
afbora, line 1502: What's the difference with the following line?
https://github.com/getkirby/kirby/blob/3.9.8/src/Cms/App.php#L1554

bastianallgeier:
@afbora I found two more places where we pass kirby to the user models.

src/Cms/App.php
bastianallgeier: No difference. I just removed it from there as well.

afbora:
@bastianallgeier Sorry, I just pointing what I found similar ones. Is this possible one?
https://github.com/getkirby/kirby/blob/fix/6061-serializable-objects/src/Cms/App.php#L1470

bastianallgeier:
@afbora yep, we don't even use it in the role class

lukasbestle (approved):
Those changes all look safe to me as I can't find any place in our code that tries to access those `kirby` props. Also can't find more places where we pass but don't use the `kirby` prop. A few classes (`Cms\Api` and `Cms\Auth\Status`) actually rely on it, but those are commonly not serialized.

distantnative (approved):

PR #6085, opened 2023-12-22
Improve link kirbytag behavior when uuid point to non-existing page

### LINKED ISSUE #6083: kirbytext() fails, if text contains a link to a non-existing page
## Description
With the new link-field in Kirby 4 it is possible to add links to intern pages in textarea fields. This looks like this:

```(link: page://sk0iiTNRft7shn6R text: Link)```

If the page is deleted afterwards or if the ID is manually changed by the user, the link-conversion by ```kirbytext()``` causes an php fatal error.

**Expected behavior**  
Link should be removed because it's not valid.

## To reproduce

1. Add a page A
2. Link to page A in a textarea field on page B
3. delete page A
4. use ```<?= $page->textareafield()->kirbytext() ?>``` in a template

## Your setup
**Kirby Version**  
Kirby 4

**Console output**  
Call to a member function url() on null in /kirby/config/tags.php line 200

### PR DESCRIPTION
**Pull request by afbora**
### Fixes
- kirbytext() fails, if text contains a link to a non-existing page #6083

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/tags.php
+++ b/config/tags.php
@@ -210,7 +210,7 @@ return [
 						throw new NotFoundException('The linked page cannot be found');
 					}
 				} else {
-					$tag->value = $tag->kirby()->site()->errorPageId();
+					$tag->value = Url::to($tag->kirby()->site()->errorPageId());
 				}
 			}
 
--- a/config/tags.php
+++ b/config/tags.php
@@ -2,6 +2,7 @@
 
 use Kirby\Cms\Html;
 use Kirby\Cms\Url;
+use Kirby\Exception\NotFoundException;
 use Kirby\Text\KirbyTag;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\Str;
@@ -197,7 +198,20 @@ return [
 				Uuid::is($tag->value, 'page') === true ||
 				Uuid::is($tag->value, 'file') === true
 			) {
-				$tag->value = Uuid::for($tag->value)->model()?->url() ?? '';
+				$tag->value = Uuid::for($tag->value)->model()?->url();
+			}
+
+			// if url is empty, throw exception or link to the error page
+			if ($tag->value === null) {
+				if ($tag->kirby()->option('debug', false) === true) {
+					if (empty($tag->text) === false) {
+						throw new NotFoundException('The linked page is not found for the link text "' . $tag->text . '"');
+					} else {
+						throw new NotFoundException('The linked page is not found');
+					}
+				} else {
+					$tag->value = $tag->kirby()->site()->errorPageId();
+				}
 			}
 
 			return Html::a($tag->value, $tag->text, [
--- a/config/tags.php
+++ b/config/tags.php
@@ -197,7 +197,7 @@ return [
 				Uuid::is($tag->value, 'page') === true ||
 				Uuid::is($tag->value, 'file') === true
 			) {
-				$tag->value = Uuid::for($tag->value)->model()->url();
+				$tag->value = Uuid::for($tag->value)->model()?->url() ?? '';
 			}
 
 			return Html::a($tag->value, $tag->text, [
--- a/config/tags.php
+++ b/config/tags.php
@@ -205,9 +205,9 @@ return [
 			if ($tag->value === null) {
 				if ($tag->kirby()->option('debug', false) === true) {
 					if (empty($tag->text) === false) {
-						throw new NotFoundException('The linked page is not found for the link text "' . $tag->text . '"');
+						throw new NotFoundException('The linked page cannot be found for the link text "' . $tag->text . '"');
 					} else {
-						throw new NotFoundException('The linked page is not found');
+						throw new NotFoundException('The linked page cannot be found');
 					}
 				} else {
 					$tag->value = $tag->kirby()->site()->errorPageId();
--- a/tests/Text/KirbyTagsTest.php
+++ b/tests/Text/KirbyTagsTest.php
@@ -5,6 +5,7 @@ namespace Kirby\Text;
 use Exception;
 use Kirby\Cms\App;
 use Kirby\Exception\InvalidArgumentException;
+use Kirby\Exception\NotFoundException;
 use Kirby\Filesystem\Dir;
 use Kirby\Filesystem\F;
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by lukasbestle — COMMENTED** [review]
TBH I'm not sure about this. It could be pretty unexpected to get a link to the homepage instead of the intended deeplink.

An alternative could be to return an empty string instead (meaning the KirbyTag would produce an empty output, i.e. no link at all).

However both solutions have in common that they are very hard to debug. The exception on the other hand breaks the whole page. Difficult...

**Comment by afbora** [comment]
What about link to error page instead homepage?

**Comment by lukasbestle** [comment]
I like that. Then it's just a "broken link" as it would be with a hardcoded URL or path.

**Comment by distantnative** [comment]
I think other CMS in similar cases would output the text without any link in this case.

**Comment by afbora** [comment]
Good comments. I would merge ideas :)

- Link to the `error` page or throw exception if `debug` enabled
- Remove link from text if `debug` disabled

**Comment by afbora** [comment]
## when text attr set

Will removed link. I got this case, OK.

### Input
````
Lorem ipsum (link: page://not-exist text: Click Here) dolor sit amet
````

### Output
````
Lorem ipsum Click Here dolor sit amet
````


## when no text attr set
What would expected when no link text set? 

### Input
````
Lorem ipsum (link: page://not-exist) dolor sit amet
````

### Output

Remove whole kirbytag like that?
````
Lorem ipsum dolor sit amet
````

**Comment by distantnative** [comment]
Given the amount of discussion about this, I think this should not go into a patch release.

On the discussion points
- What happens with a permalink when the UUID does not exist? Is it an exception or redirect to error? I would say the kirbytag should behave the same when debug is active
- With debug false, I think just rendering the text, not the link is ok.
- When no text is set, url should be rendered as pure text.

**Comment by afbora** [comment]
> When no text is set, url should be rendered as pure text.

@distantnative So what would happen when no text is set, uuid is not exists and debug active?

**Comment by distantnative** [comment]
@afbora depends what our answer is to my first bullet point. Either exception or permalink as pure text.

**Comment by GiantCrocodile** [comment]
Another thing to consider: The expected behavior might depend on the importancy of the link. If it's a link to a legal document or important document, it should always fail to prevent failing legal or other processes afterwards. So maybe it might be important to have a way that the editor/admin can define the expected behavior by the importancy of the link.

**Comment by lukasbestle** [comment]
Considering everything, I think the following would be the best compromise to fulfill the most requirements:

- If debug mode is enabled, throw an exception to allow to pinpoint the source of the error. IMO that's what debug mode is for.
- Otherwise, render a link to the error page. It keeps the intended content structure, avoids printing the UUID to visitors who won't know what it means and also makes it visible to visitors that there should be a link but it isn't working (thus prompting them to contact the editors to correct the issue on the page).

File: config/tags.php [hunk 1]
```diff
@@ -197,7 +198,20 @@
...
-				$tag->value = Uuid::for($tag->value)->model()->url();
+				$tag->value = Uuid::for($tag->value)->model()?->url();
+			}
+
+			// if url is empty, throw exception or link to the error page
+			if ($tag->value === null) {
+				if ($tag->kirby()->option('debug', false) === true) {
+					if (empty($tag->text) === false) {
+						throw new NotFoundException('The linked page is not found for the link text "' . $tag->text . '"');
+					} else {
```
afbora: Feel free remove this line. This is only to easy track down for the editor.

config/tags.php
```diff
@@ -197,7 +198,20 @@
...
 			) {
-				$tag->value = Uuid::for($tag->value)->model()->url();
+				$tag->value = Uuid::for($tag->value)->model()?->url();
+			}
+
+			// if url is empty, throw exception or link to the error page
+			if ($tag->value === null) {
+				if ($tag->kirby()->option('debug', false) === true) {
+					if (empty($tag->text) === false) {
+						throw new NotFoundException('The linked page is not found for the link text "' . $tag->text . '"');
```
bastianallgeier:
```suggestion
						throw new NotFoundException('The linked page cannot be found for the link text "' . $tag->text . '"');
```

bastianallgeier (commented):
I would change the wording for the exception (see suggestion). But I'm happy with the rest. @lukasbestle @distantnative do you want to have a last look at it? I would also move it to 4.1 though.

config/tags.php
```diff
@@ -197,7 +198,20 @@
...
+			// if url is empty, throw exception or link to the error page
+			if ($tag->value === null) {
+				if ($tag->kirby()->option('debug', false) === true) {
+					if (empty($tag->text) === false) {
+						throw new NotFoundException('The linked page cannot be found for the link text "' . $tag->text . '"');
+					} else {
+						throw new NotFoundException('The linked page cannot be found');
+					}
+				} else {
+					$tag->value = $tag->kirby()->site()->errorPageId();
```
bastianallgeier: Ok, one last change. I think I would do this here.

```suggestion
					$tag->value = $tag->kirby()->site()->errorPage()->url();
```

bastianallgeier: The rest looks really good now.

lukasbestle (commented):
Awesome! 👍
I agree with Bastian on the suggested change, apart from that it's great.

bastianallgeier (dismissed):

config/tags.php
```diff
@@ -197,7 +198,20 @@
...
+			// if url is empty, throw exception or link to the error page
+			if ($tag->value === null) {
+				if ($tag->kirby()->option('debug', false) === true) {
+					if (empty($tag->text) === false) {
+						throw new NotFoundException('The linked page cannot be found for the link text "' . $tag->text . '"');
+					} else {
+						throw new NotFoundException('The linked page cannot be found');
+					}
+				} else {
+					$tag->value = $tag->kirby()->site()->errorPage()->url();
```
bastianallgeier: It only happens if we don't have an error page, but I think this is why the tests fail. 

```suggestion
					$tag->value = $tag->kirby()->site()->errorPage()?->url();
```

afbora: What about `Url::to($tag->kirby()->site()->errorPageId())`?

bastianallgeier: Good idea.

bastianallgeier (approved):

PR #6090, opened 2023-12-23
Allow activation of legacy and inactive licenses

### LINKED ISSUE #6089: Activation of legacy licenses fails
When activating a Kirby 4 site with a Kirby 3 license, the activation fails with "The license could not be verified". However the hub still activates the license to the domain.

lukasbestle:
### Fixes

- Kirby 3 licenses can now activated from the Kirby 4 Panel to allow direct renewal #6089

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/LicenseStatus.php
+++ b/src/Cms/LicenseStatus.php
@@ -44,6 +44,22 @@ enum LicenseStatus: string
 	case Missing = 'missing';
 
 	/**
+	 * Checks if the license can be saved when it
+	 * was entered in the activation dialog;
+	 * renewable licenses are accepted as well
+	 * to allow renewal from the Panel
+	 */
+	public function activatable(): bool
+	{
+		return match ($this) {
+			static::Active   => true,
+			static::Inactive => true,
+			static::Legacy   => true,
+			default          => false
+		};
+	}
+
+	/**
 	 * Returns the dialog according to the status
 	 */
 	public function dialog(): string
--- a/src/Cms/LicenseStatus.php
+++ b/src/Cms/LicenseStatus.php
@@ -52,10 +52,10 @@ enum LicenseStatus: string
 	public function activatable(): bool
 	{
 		return match ($this) {
-			static::Active   => true,
-			static::Inactive => true,
-			static::Legacy   => true,
-			default          => false
+			static::Active,
+			static::Inactive,
+			static::Legacy    => true,
+			default           => false
 		};
 	}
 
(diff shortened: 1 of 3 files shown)


src/Cms/LicenseStatus.php
```diff
@@ -43,6 +43,22 @@ enum LicenseStatus: string
...
+	 */
+	public function activatable(): bool
+	{
+		return match ($this) {
+			static::Active   => true,
+			static::Inactive => true,
+			static::Legacy   => true,
+			default          => false
+		};
+	}
```
distantnative, line 60: Just two small suggestions, not essential for me
- `canBeActivated` could be an easier to parse name
- I like to use match statements that produce the same result in union, so

```php
			static::Active,
			static::Inactive,
			static::Legacy    => true,
```

lukasbestle: I agree on the name, but it would be inconsistent with the existing `renewable` name. Can `LicenseStatus` be considered internal so that we can change both?

Also agree on the `match` statement. We can do the same in `renewable`/`canBeRenewed`.

distantnative: I think we can/should consider it as internal, maybe needs the tag added if not yet. I'm fine with changing both or leaving activatable... you are right that it should be consistent.

bastianallgeier: How strict do we want to be with semver here? I'm also fine with both ways.

distantnative: I think in general these classes/enums should be off limit for developers and we should mark them as such. Nobody should fiddle with our license code.

bastianallgeier: But this means that we should at least make this change in 4.1 and also introduce the internal tags with the PR, right?

distantnative: True. I think timing wise we could release 4.1 then very soon in the new year.

lukasbestle: > Nobody should fiddle with our license code.

Not even sure TBH. Especially the status could be used in agency-internal project management plugins or monitoring tools like Kirby Fleet Manager.

For me the `@internal` tag is more for code that we don't consider stable or don't see a reason to document.

IMO I'd keep the `LicenseStatus` enum public and also keep the method names to avoid a breaking change. If we are all fine with both names, I don't think we should break SemVer for no real reason.

distantnative: Fair enough

bastianallgeier: Works for me as well.

bastianallgeier (approved):

PR #6093, opened 2023-12-27
Refactor deprecation warning tests

distantnative:
PHPUnit 10 won't throw errors anymore, which is why we need to switch to a different test setup for our deprecation errors

### DIFF
--- a/tests/Cms/HelpersTestCase.php
+++ b/tests/Cms/HelpersTestCase.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Kirby\Cms;
+
+use Closure;
+
+class HelpersTestCase extends TestCase
+{
+	protected $hasErrorHandler = false;
+
+	public function tearDown(): void
+	{
+		parent::tearDown();
+
+		if ($this->hasErrorHandler === true) {
+			restore_error_handler();
+			$this->hasErrorHandler = false;
+		}
+	}
+
+	public function assertError(
+		int $expectedErrorType,
+		string $exptectedErrorMessage,
+		Closure $callback,
+		bool $expectedFailure = true
+	) {
+		$this->hasErrorHandler = true;
+
+		$called = false;
+
+		set_error_handler(
+			function (int $errno, string $errstr) use ($expectedErrorType, $exptectedErrorMessage, &$called) {
+				$called = true;
+				$this->assertSame($expectedErrorType, $errno);
+				$this->assertSame($exptectedErrorMessage, $errstr);
+			}
+		);
+
+		$result = $callback();
+
+		if ($expectedFailure === false) {
+			$this->assertFalse($called);
+			return $result;
+		}
+
+		$this->assertTrue($called);
+	}
+}
--- a/tests/Cms/HelpersTestCase.php
+++ b/tests/Cms/HelpersTestCase.php
@@ -6,15 +6,15 @@ use Closure;
 
 class HelpersTestCase extends TestCase
 {
-	protected $hasErrorHandler = false;
+	protected int $activeErrorHandlers = 0;
 
 	public function tearDown(): void
 	{
 		parent::tearDown();
 
-		if ($this->hasErrorHandler === true) {
+		while ($this->activeErrorHandlers > 0) {
 			restore_error_handler();
-			$this->hasErrorHandler = false;
+			$this->activeErrorHandlers--;
 		}
 	}
 
@@ -24,7 +24,7 @@ class HelpersTestCase extends TestCase
 		Closure $callback,
 		bool $expectedFailure = true
 	) {
-		$this->hasErrorHandler = true;
+		$this->activeErrorHandlers++;
 
 		$called = false;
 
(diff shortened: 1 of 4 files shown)


tests/Cms/HelpersTestCase.php
```diff
@@ -0,0 +1,48 @@
...
+		int $expectedErrorType,
+		string $exptectedErrorMessage,
+		Closure $callback,
+		bool $expectedFailure = true
+	) {
+		$this->hasErrorHandler = true;
+
+		$called = false;
+
+		set_error_handler(
```
bastianallgeier, line 31: Do we need to unset that error handler somehow to avoid side effects?

distantnative: Gets unset in the tearDown

lukasbestle: I think it would be safer to restore it directly in the assertion. Otherwise it will break if multiple assertions are used in one test (as multiple error handlers will be registered but only one will be restored). That would be quite hard to debug.

distantnative: I just want to keep this PR as low key as possible and we have used tearDown already so far for this

tests/Cms/HelpersTestCase.php
```diff
@@ -0,0 +1,48 @@
...
+			restore_error_handler();
+			$this->hasErrorHandler = false;
+		}
+	}
+
+	public function assertError(
+		int $expectedErrorType,
+		string $exptectedErrorMessage,
+		Closure $callback,
+		bool $expectedFailure = true
```
lukasbestle: PHPUnit assertions always follow the argument convention `assertSomething($expected, $actual)`. For consistency and to reduce the number of arguments I'd suggest:

```suggestion
		array|false $expected,
		Closure $callback
```

If `$expected` is an array, a failure is expected with `$errorType = $expected[0]` and `$errorMessage = $expected[1]`.

tests/Cms/HelpersTestCase.php
lukasbestle: We still need the `hasErrorHandler` check, but IMO that one belongs back into the `HelpersTest` class where the `hasErrorHandler` prop is set by tests.

distantnative: Why? It's needed for `HelpersTestCase::assertError` too. Although I saw that you just removed it, but I don't understand why.

tests/Cms/HelpersTestCase.php
distantnative: There are also standard assertion methods in PHPUnit with only one or more than two arguments. I would rather keep it instead of making it more cryptic.

tests/Cms/HelpersTestCase.php
lukasbestle: My commit was for an unrelated method where I spotted it being used superfluously. Doesn't relate to the `assertError()` method.

Until now, the `$hasErrorHandler` prop was only needed for tests that set a custom error handler. If those tests fail, their code stops executing immediately and they cannot reset the handler again, so it has to be done in the `tearDown()` method.

Thinking about it again, the same can happen with the assertion method (didn't realize that before): If the assertion fails, execution stops immediately here as well and we cannot reset the error handler. So we do need the `tearDown()` method and I'm fine with leaving it in the `HelpersTestCase` as it needs to be used there.

However we still have the issue of not being able to call the `assertError()` method multiple times in a single test without accumulating error handlers. We could fix that by turning the `bool $hasErrorHandler = false` prop into an `int $activeErrorHandlers = 0` prop. We can increase that number by one for each error handler we set. In `tearDown()`, we can then reset all handlers in a loop until the count is at 0 again.

distantnative: Do we need to use it more than once per test method?

tests/Cms/HelpersTestCase.php
lukasbestle: You are right that it would become more cryptic. I don't really like either solution, but I can't find one that ticks both the "not cryptic" and "consistent with PHPUnit" boxes. So it's fine for me.

However there's a typo in the second argument name. Could you please fix that?

```diff
- $exptectedErrorMessage
+ $expectedErrorMessage
```

lukasbestle:
Looking into the CI failure. Something seems to be off as my commit worked locally before but now works neither in CI nor locally.

tests/Cms/HelpersTestCase.php
lukasbestle: Implemented that locally, will commit it with the CI fix.

lukasbestle: > Do we need to use it more than once per test method?

No, but if we ever do, we would cause a hard to debug issue in CI.

lukasbestle:
@distantnative Reverted my earlier commit as it broke subsequent tests, but I couldn't find out why.

From reading the code, this test should not need to restore the error handler as the one set inside `Helpers::handleErrors()` is already reset within the method and no other error handler is set:

https://github.com/getkirby/kirby/blob/ba8e9e511d797c0c2bc9278d5484020c4aa8cfce/tests/Cms/Helpers/HelpersTest.php#L198-L207

Can you remember why you added the `hasErrorHandler` line in this test?

lukasbestle (approved):

PR #6094, opened 2023-12-27
Structure field: clone row on duplicate

### LINKED ISSUE #6073: Clone row on duplicate In structure field
## Description
I have a very strange issue in a custom field implementation: I moved the actual editing of the blocks field into a drawer. This works fine. However, if have an issue with structure fields inside of a block, when using my custom field:
- I'm duplicating an existing structure row with content
- I edit the new row
- Changes I do in the duplicated row are also made in the original row. This is the case for all fields with arrays or objects as data structures, e.g. blocks, files, …. But not for simple text fields.
- Once I reload the Panel, everything works fine and my changes are only happening in the row I'm currently working on. 

With the original Kirby implementation everything works fine. But for some reason it occurs with my custom field.

**Expected behavior**  
The content of the duplicated structure should only be changed in the duplicated row and not also in the original row.

**Sugessted fix**
After a lot of digging I didn't find the issue witting my field implementation. However I found a way to fix it within Kirby. In the Structure field component (https://github.com/getkirby/kirby/blob/main/panel/src/components/Forms/Field/StructureField.vue#L458) instead of doing:

```js
case "duplicate":
	this.add({
		...row,
		_id: this.$helper.uuid()
	});
	break;
```

doing this:
```js
case "duplicate":
	this.add({
		...this.$helper.clone(row),
		_id: this.$helper.uuid()
	});
	break;
```
is fixing the issue with row duplication/editing in my custom field. I don't know what the real root cause for the issue in my field is, but this seems to be a plausible fix for me. The same is already done for block duplication, so would it be possible, to do the cloning for structures as well?

## Your setup
**Kirby Version**  4.0.0

distantnative:
### Fixes
- #6073

### For review team

### DIFF
--- a/panel/src/components/Forms/Field/StructureField.vue
+++ b/panel/src/components/Forms/Field/StructureField.vue
@@ -457,7 +457,7 @@ export default {
 
 				case "duplicate":
 					this.add({
-						...row,
+						...structuredClone(row),
 						_id: this.$helper.uuid()
 					});
 					break;


bastianallgeier:
I wonder if we should switch our clone helper to structuredClone

distantnative:
@bastianallgeier you mean as in https://github.com/getkirby/kirby/pull/6086? 😜

bastianallgeier:
Yep, looks like I need to get up to speed with the PRs first 🙈

bastianallgeier (approved):

PR #6097, opened 2023-12-29
Replace `this._uid`

distantnative:
Vue 3 doesn't like `this._uid` so we should already replace it in our usage.

### Refactored
- Removed `this._uuid` usage

### DIFF
--- a/panel/src/components/Dropdowns/DropdownContent.vue
+++ b/panel/src/components/Dropdowns/DropdownContent.vue
@@ -17,10 +17,10 @@
 			<!-- @slot Content of the dropdown which overrides passed `options` prop -->
 			<slot>
 				<template v-for="(option, index) in items">
-					<hr v-if="option === '-'" :key="_uid + '-item-' + index" />
+					<hr v-if="option === '-'" :key="id + '-item-' + index" />
 					<k-dropdown-item
 						v-else-if="option.when ?? true"
-						:key="_uid + '-item-' + index"
+						:key="id + '-item-' + index"
 						v-bind="option"
 						@click="onOptionClick(option)"
 					>
@@ -116,6 +116,11 @@ export default {
 			items: [],
 			opener: null
 		};
+	},
+	computed: {
+		id() {
+			return useUid();
+		}
 	},
 	created() {
 		if (this.align) {
--- a/panel/src/components/Dropdowns/DropdownContent.vue
+++ b/panel/src/components/Dropdowns/DropdownContent.vue
@@ -17,10 +17,10 @@
 			<!-- @slot Content of the dropdown which overrides passed `options` prop -->
 			<slot>
 				<template v-for="(option, index) in items">
-					<hr v-if="option === '-'" :key="id + '-item-' + index" />
+					<hr v-if="option === '-'" :key="'separator-' + index" />
 					<k-dropdown-item
 						v-else-if="option.when ?? true"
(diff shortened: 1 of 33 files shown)


panel/src/components/Dropdowns/DropdownContent.vue
```diff
@@ -17,10 +17,10 @@
 			<!-- @slot Content of the dropdown which overrides passed `options` prop -->
 			<slot>
 				<template v-for="(option, index) in items">
-					<hr v-if="option === '-'" :key="_uid + '-item-' + index" />
+					<hr v-if="option === '-'" :key="id + '-item-' + index" />
```
bastianallgeier: Is this really needed here at all? Do keys need to be globally unique or just within a component? Because otherwise `'item-' + index'` seems totally fine to me.

distantnative: Hmmm actually I'd think just unique within the loop

bastianallgeier: I think so too. I honestly believe that this is from a time when I still thought they need to be globally unique. But I never ran into issues with keys that are just unique within the loop ever since.

distantnative: I'll change it

bastianallgeier (commented):
Just a thought. Wouldn't it be more accessible through $helper.guid or something like that?

distantnative:
@bastianallgeier we'll need more testing but I think we could even get rid of the uid completely

panel/src/components/Forms/Field/TimeField.vue
```diff
@@ -1,8 +1,7 @@
 <template>
-	<k-field v-bind="$props" :input="_uid" class="k-time-field">
+	<k-field v-bind="$props" class="k-time-field">
```
bastianallgeier:
```suggestion
	<k-field v-bind="$props" :input="id" class="k-time-field">
```

panel/src/components/Forms/Input/CheckboxesInput.vue
```diff
@@ -54,6 +54,7 @@ export default {
 					autofocus: this.autofocus && index === 0,
 					checked: this.selected.includes(option.value),
 					disabled: this.disabled || option.disabled,
+					id: this.id,
```
bastianallgeier: I don't think we can do this. This would assign the same id to multiple checkboxes, which is not allowed afaik

panel/src/components/Forms/Input/RadioInput.vue
```diff
@@ -41,6 +41,7 @@ export default {
 					autofocus: this.autofocus && index === 0,
 					checked: this.value === option.value,
 					disabled: this.disabled || option.disabled,
+					id: this.id,
```
bastianallgeier: Same here

panel/src/components/Forms/Input/CheckboxesInput.vue
distantnative: What do you think about this solution https://github.com/getkirby/kirby/pull/6097/commits/6d8b2227c277f64014c30eb30a46e022cec858b6 ?

bastianallgeier: I like it!

bastianallgeier (approved):

PR #6099, opened 2023-12-29
Goodbye `vuelidate`

distantnative:
### Enhancements
- Relying front-end validation on native HTML invalid states

### Refactoring
- Removed `vuelidate` library

### Breaking changes
- `k-form` and `k-fieldset` as well as many fields and inputs don't emit an `invalid` event anymore. Use native HTML invalid state of elements instead.
- `novalidate` prop has been removed from all elements but `k-form`

## Review
- [x] Add to Notion changelog


distantnative:
@bastianallgeier Could I pick your brain

> Consider if to add alternative to ArrayInput e.g. for ObjectField (works with ArrayInput, but naming wise a bit confusing)

How do you feel about the `ArrayInput` after a while now? Still good for checkboxes etc.? And what about e.g. ObjectField?

bastianallgeier:
@distantnative I was thinking about this. How would we handle this if it was plain html & js and what is the most accessible solution? Maybe we are already trying to make it too smart again?

distantnative:
@bastianallgeier could we create a similar thing to ArrayInput with a mixin that provides most of the validate methods etc. and relies on a computed prop `input()` e.g. that points to the HTMLElement of the specific component to which the validity etc. should be applied to?

bastianallgeier:
TBH whenever I come back to this, I think that we need to break it down like we did with a couple monster tasks during v4. It is just too overwhelming in a single PR.

distantnative:
@bastianallgeier That's what I've already been doing with many recent PRs. They all came from this. For me this PR is a good progress bar to see what we still can isolate etc.

It will always be a big PR still in the end. We have so many inputs and fields and some parts can only be done at once.

The biggest showstopper here IMO is that we don't know how to pursue with ArrayInput.

The rest then is mainly just double checking if all validations really exist and add lab examples where missing.

bastianallgeier:
Ok, that makes sense. I agree that the array input is super weird. Maybe we need to tackle that separately with a concept first and work on it together.

distantnative:
@bastianallgeier I think the crucial part here is to decide if the use of `novalidate` here is correct

bastianallgeier (approved):
I love how much code this removes and how much simpler the inputs are afterwards. I think the form handling is great like that. We should keep an eye on the dialogs and other forms, but from what I could see it all works as I would expect. It actually works even better in some parts because the field validation is a lot more consistent now and the browser behavior on submit is more reliable.

PR #6100, opened 2023-12-30
Replace `instanceof Vue`

distantnative:
Vue 3 hates `instanceof Vue`. So already trying to find alternatives.

What I mean by "hates":

https://github.com/getkirby/kirby/assets/3788865/bdcf76f0-db62-47ad-a6f7-20abfb737690

### DIFF
--- a/panel/src/helpers/isComponent.js
+++ b/panel/src/helpers/isComponent.js
@@ -6,12 +6,3 @@
 export default (name) => {
 	return typeof window.Vue.options.components[name] === "function";
 };
-
-/**
- * Checks if the coponent is an instance of Vue
- * @param {object} component
- * @returns {bool}
- */
-export function isVueComponent(component) {
-	return typeof component?.render === "function";
-}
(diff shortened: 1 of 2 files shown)


panel/src/helpers/isComponent.js
```diff
@@ -6,3 +6,12 @@
...
 	return typeof window.Vue.options.components[name] === "function";
 };
+
+/**
+ * Checks if the coponent is an instance of Vue
+ * @param {object} component
+ * @returns {bool}
+ */
+export function isVueComponent(component) {
+	return typeof component?.render === "function";
```
bastianallgeier: Does this work when components set a template instead of defining a render function?

distantnative: Hmm no and this doesn't seem to work as I thought (even though it's the suggested way when googling).

I think there are two ways:
- check for `component._isVue`
- check for `component.$el`

I would lean towards the second but not as `isVueComponent` helper but directly just replacing the line in `DropdownContent`

bastianallgeier (approved):
Ha, that's perfect!

PR #6102, opened 2023-12-30
Remove input theme field mentions

distantnative:
As far as I see it these are leftovers and don't serve any purpose anymore


bastianallgeier (approved):
LGTM

PR #6103, opened 2023-12-30
Use `StringInput` for all text inputs

distantnative:
### Refactored
- use `k-string-input` for all text inputs

### DIFF
--- a/panel/src/components/Forms/Input/PasswordInput.vue
+++ b/panel/src/components/Forms/Input/PasswordInput.vue
@@ -25,6 +25,6 @@ export const props = {
  * @example <k-input :value="password" @input="password = $event" name="password" type="password" />
  */
 export default {
-	mixins: [StringInput]
+	mixins: [StringInput, props]
 };
 </script>
--- a/panel/src/components/Forms/Input/PasswordInput.vue
+++ b/panel/src/components/Forms/Input/PasswordInput.vue
@@ -1,16 +1,22 @@
+<template>
+	<k-string-input
+		v-bind="$props"
+		type="password"
+		class="k-password-input"
+		@input="$emit('input', $event)"
+		@invalid="($invalid, $v) => $emit('invalid', $invalid, $v)"
+	/>
+</template>
+
 <script>
-import TextInput, { props as TextInputProps } from "./TextInput.vue";
+import StringInput, { props as StringInputProps } from "./StringInput.vue";
 
 export const props = {
-	mixins: [TextInputProps],
+	mixins: [StringInputProps],
 	props: {
 		autocomplete: {
 			type: String,
 			default: "new-password"
-		},
-		type: {
-			type: String,
-			default: "password"
 		}
 	}
 };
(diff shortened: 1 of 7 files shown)


panel/src/components/Forms/Input/PasswordInput.vue
```diff
@@ -19,7 +25,6 @@ export const props = {
  * @example <k-input :value="password" @input="password = $event" name="password" type="password" />
  */
 export default {
-	extends: TextInput,
-	mixins: [props]
```
bastianallgeier, line 23: Why are the props removed here?

panel/src/components/Forms/Input/SlugInput.vue
```diff
@@ -17,7 +17,7 @@
 		spellcheck="false"
 		type="text"
 		class="k-text-input"
-		v-on="listeners"
+		@input="$emit('input', $event)"
```
bastianallgeier, line 20: Not sure about this one. Wouldn't this remove the validate listener from the text input, which could be a breaking change?

panel/src/components/Forms/Input/TelInput.vue
```diff
@@ -20,7 +27,6 @@ export const props = {
  * @example <k-input :value="tel" @input="tel = $event" name="tel" type="tel" />
  */
 export default {
-	extends: TextInput,
-	mixins: [props]
```
bastianallgeier, line 24: See above.

panel/src/components/Forms/Input/UrlInput.vue
```diff
@@ -23,7 +29,6 @@ export const props = {
  * @example <k-input :value="url" @input="url = $event" name="url" type="url" />
  */
 export default {
-	extends: TextInput,
-	mixins: [props]
```
bastianallgeier, line 27: One more

bastianallgeier (commented):
I think this clean-up is really good. But I wonder if we are again running into breaking change territory with the CSS classes. So far, all the refactored fields hat a k-text-input class which is now replaced by k-string-input. This could lead to problems.

panel/src/components/Forms/Input/SlugInput.vue
distantnative: True probably needs `@invalid="($invalid, $v) => $emit('invalid', $invalid, $v)"` as the other inputs as well

distantnative: Oh not, cause it's a native `input` element, I don't think the previous invalid event listener had any real function on this native element

distantnative:
@bastianallgeier that's true. maybe rather `v5/develop` then?

bastianallgeier (approved):

bastianallgeier:
One last thought before we move on with this. Can we still properly work on our inputs during v4 when we merge this and also maybe the vuelidate stuff or do we need to do more in v4 as preparation? I just don't want us to get in the way with ourselves.

distantnative:
I think interior be good to have this before the vuelidate work (it's basically already in the vuelidate PR but thought it should be separate).

Of course we'll need to be careful about merge conflicts if we make changes to those specific inputs in v4.x

bastianallgeier:
Ok, lets just go for it. It served us well during the v4 alphas to move ahead with such tasks and then just keep an eye on merging changes back in quickly.

PR #6108, opened 2024-01-01
Fix `counter.js` for stricter Vue 3 checks

distantnative:
### For review team

### DIFF
--- a/panel/src/components/Forms/Counter.vue
+++ b/panel/src/components/Forms/Counter.vue
@@ -1,9 +1,12 @@
 <template>
 	<span :data-invalid="!valid" class="k-counter">
 		<span>{{ count }}</span>
-		<span v-if="min && max" class="k-counter-rules">({{ min }}–{{ max }})</span>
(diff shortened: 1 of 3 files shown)


panel/src/components/Forms/Counter.vue
```diff
@@ -1,9 +1,12 @@
 <template>
 	<span :data-invalid="!valid" class="k-counter">
 		<span>{{ count }}</span>
-		<span v-if="min && max" class="k-counter-rules">({{ min }}–{{ max }})</span>
-		<span v-else-if="min" class="k-counter-rules">≥ {{ min }}</span>
-		<span v-else-if="max" class="k-counter-rules">≤ {{ max }}</span>
+
+		<span class="k-counter-rules">
```
bastianallgeier: Shouldn't the span still be removed if min and max are not set?

distantnative: True

bastianallgeier (approved):

PR #6109, opened 2024-01-01
Fix CSS selectors for boolean attributes

distantnative:
### Fixes
- Use strict CSS selectors for boolean (data) attributes

### Breaking changes
- CSS attribute selectors must be written fully qualified (e.g. `[data-hidden="true"]` as only `[data-hidden]` can start matching also elements where that attribute is `false`)

### For review team

### DIFF
--- a/panel/src/components/Layout/Table.vue
+++ b/panel/src/components/Layout/Table.vue
@@ -399,7 +399,7 @@ export default {
 	color: var(--table-color-th-text);
 	background: var(--table-color-th-back);
 }
-.k-table th[data-has-button] {
+.k-table th[data-has-button="true"] {
 	padding: 0;
 }
 .k-table th button {
--- a/panel/lab/components/tables/1_horizontal/index.vue
+++ b/panel/lab/components/tables/1_horizontal/index.vue
@@ -47,8 +47,8 @@
 				<table>
 					<thead>
 						<tr>
-							<th data-has-button><button>Name</button></th>
-							<th data-has-button><button>Date</button></th>
+							<th data-has-button="true"><button>Name</button></th>
+							<th data-has-button="true"><button>Date</button></th>
 						</tr>
 					</thead>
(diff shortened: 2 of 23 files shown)


panel/src/components/Layout/Table.vue
```diff
@@ -398,7 +398,7 @@ export default {
 	color: var(--table-color-th-text);
 	background: var(--table-color-th-back);
 }
-.k-table th[data-has-button] {
```
bastianallgeier, line 402: While going through all the examples, I ran into a breaking issue in the lab. If you ever use one of the attributes manually and you don't make it strict, this is basically a breaking change. I wonder how we deal with this.

distantnative: What do you mean with using them manually?

bastianallgeier: In the lab, we use `<th data-has-button>` in our example markup. That used to work so far, but does no longer work now.

distantnative: Hmmm we could for the moment to add selectors for both. But those use cases will break sooner or later anyway when Vue renders `data-attribute="false"` ad those selectors just say "should exist at all" not whether they are true/positive.

bastianallgeier: That's true. I just really wonder if we need to consider this a breaking change or not. Feels still a bit new to get used to such cases.

distantnative: We could also add it to `v5/develop` instead and list it as breaking change.

bastianallgeier: I think if we want to be truly carefuly, we should do it like that. It feels at least weird to do that already in 4.1, that's why I'm leaning towards that suggestion tbh.

bastianallgeier (approved):

PR #6112, opened 2024-01-01
Improve file sorting performance

### LINKED ISSUE #6063: File sorting performance
## Description
Sorting files in a pages section is unnecessarily slow: Flipping 2 images in a pages section showing 60 images takes over 20 seconds on my test server. During this time the panel is unresponsive. 

**Expected behavior**  
Sorting files in the panel should be an interactive experience. 

**Screenshots**  
image

## To reproduce

1. Install plainkit
2. Add a files section to a page blueprint:
```yml
sections:
  files:
    layout: cards
    limit: 100
```
3. Add about 60 images, their size doesn't matter
4. Flip any 2 images and wait

## Your setup
**Kirby Version**  
4.0.1

**Additional context**  
The waiting time is spent in two places which can be optimized. 
1. Even when switching the place of 2 adjacent images, the "changeSort" commit action is applied to every image in the collection, while it would be enough to only change the two affected files. Adding a check like
    ```php
    		if($this->sort()->isNotEmpty() && $this->sort()->toInt() === $sort) {
    			return $this;
    		}
    ```
    to [`changeSort()` in FileActions.php](https://github.com/getkirby/kirby/blob/4.0.1/src/Cms/FileActions.php#L101-L108) in the best case reduces the 20 seconds to 10.2 seconds (of course putting, for example, the last image to the start still affects all of the images). 
    The slow part in `commit()` is the creation of [`$old = $this->hardcopy()`](https://github.com/getkirby/kirby/blob/4.0.1/src/Cms/FileActions.php#L158), in there it's the [applying](https://github.com/getkirby/kirby/blob/4.0.1/src/Cms/ModelWithContent.php#L353) of the `blueprints` method that takes almost all time (e.g. 186ms out of 187ms). 
2. The other half is spent in producing a response for the API call which is then [promptly discarded by the frontend](https://github.com/getkirby/kirby/blob/4.0.1/panel/src/components/Sections/FilesSection.vue#L92-L95). The easiest would be to just return a `true` from the API route, this reduces the 10.2s to 0.2s. However, I've dug a bit deeper and the problem in the File API model seems to be the [`options` property](https://github.com/getkirby/kirby/blob/4.0.1/config/api/models/File.php#L31), which makes up 99% of the time spent: in particular it's the [`canChangeTemplate` permission](https://github.com/getkirby/kirby/blob/4.0.1/src/Cms/FilePermissions.php#L18-L25) which is slow. 

Besides the workarounds listed here above, both issues seem to point to the [`blueprints` method](https://github.com/getkirby/kirby/blob/4.0.1/src/Cms/File.php#L158) (one via `hardcopy`, the other via `canChangeTemplate`). That also explains why the time is equally split between the two paths. 

I don't really understand everything that happens in `File::blueprints()`, but I wonder if the result couldn't be cached "by parent", rather than as instance variable for a single File. This expects a parent to always have the same blueprint during a request, and therefore also the same accepted file blueprints. I think this is a safe assumption? 

Here's an example implementation, just for the tests (I know the code isn't very Kirby like, but it shows the speedup from 20 seconds to 0.2 seconds): (my additions are marked with a "ADDED THIS" comment)
```php
	public function blueprints(string $inSection = null): array
	{
		if ($inSection === null && $this->blueprints !== null) {
			return $this->blueprints; // @codeCoverageIgnore
		}

		// always include the current template as option
		$template  = $this->template() ?? 'default';
		$templates = [$template];
		$parent    = $this->parent();

		// ADDED THIS:
		static $cache = new \WeakMap();
		if($inSection === null && $cache->offsetExists($parent)) {
			return $cache->offsetGet($parent);
		}

		// what file templates/blueprints should be considered is
		// defined bythe parent's blueprint: which templates it allows
		// in files sections as well as files fields
		$blueprint = $parent->blueprint();

		$fromFields = function ($fields) use (&$fromFields, $parent) {
			$templates = [];

			foreach ($fields as $field) {
				// files or textare field
				if (
					$field['type'] === 'files' ||
					$field['type'] === 'textarea'
				) {
					$uploads = $field['uploads'] ?? null;

					// only if the `uploads` parent is the actual parent
					if ($target = $uploads['parent'] ?? null) {
						if ($parent->id() !== $target) {
							continue;
						}
					}

					$templates[] = $uploads['template'] ?? 'default';
					continue;
				}

				// structure field
				if ($field['type'] === 'structure') {
					$fields    = $fromFields($field['fields']);
					$templates = array_merge($templates, $fields);
					continue;
				}
			}

			return $templates;
		};

		// collect all allowed templates…
		foreach ($blueprint->sections() as $section) {
			// if collecting for a specific section, skip all others
			if ($inSection !== null && $section->name() !== $inSection) {
				continue;
			}

			//  …from files sections
			if ($section->type() === 'files') {
				$templates[] = $section->template() ?? 'default';
				continue;
			}

			//  …from fields
			if ($section->type() === 'fields') {
				$fields    = $fromFields($section->fields());
				$templates = array_merge($templates, $fields);
			}
		}

		// make sure every template is only included once
		$templates = array_unique(array_filter($templates));

		// load the blueprint details for each collected template name
		$blueprints = [];

		foreach ($templates as $template) {
			// default template doesn't need to exist as file
			// to be included in the list
			if ($template === 'default') {
				$blueprints[$template] = [
					'name'  => 'default',
					'title' => '– (default)',
				];
				continue;
			}

			if ($blueprint = FileBlueprint::factory('files/' . $template, null, $this)) {
				try {
					// ensure that file matches `accept` option,
					// if not remove template from available list
					$this->match($blueprint->accept());

					$blueprints[$template] = [
						'name'  => $name = Str::after($blueprint->name(), '/'),
						'title' => $blueprint->title() . ' (' . $name . ')',
					];
				} catch (Exception) {
					// skip when `accept` doesn't match
				}
			}
		}

		$blueprints = array_values($blueprints);

		// sort blueprints alphabetically while
		// making sure the default blueprint is on top of list
		usort($blueprints, fn ($a, $b) => match (true) {
			$a['name'] === 'default' => -1,
			$b['name'] === 'default' => 1,
			default => strnatcmp($a['title'], $b['title'])
		});

		// no caching for when collecting for specific section
		if ($inSection !== null) {
			return $blueprints; // @codeCoverageIgnore
		}

		// ADDED THIS:
		$cache->offsetSet($parent, $blueprints);

		return $this->blueprints = $blueprints;
	}
```

distantnative:
## This PR …

- [x] Add unit tests for `HasFiles::acceptedFileTemplates()`
- [x] Restore API endpoint, but create own request Panel in site area to use without return 

 
Best to review each commit separately as each contains one contained improvement 

### Enhancements
- Significantly improved file sorting performance (thanks @rasteiner)

## Ready?

### For review team


bastianallgeier:
Maybe we could kill two birds with one stone here. We want the Panel to be independent from the API anyway. We could keep the return value for the API call to avoid breaking changes. But have our own sorting request in the site area.

src/Cms/File.php
```diff
@@ -157,14 +158,26 @@ public function blueprint(): FileBlueprint
...
 		if ($inSection === null && $this->blueprints !== null) {
 			return $this->blueprints; // @codeCoverageIgnore
 		}
 
+		$parent = $this->parent();
+
+		// get cached results from other models that share the same parent
+		// and thus the same available blueprints
+		// (except when collecting for a specific section)
+		static $cache = new WeakMap();
```
rasteiner: This is what I meant with "not very Kirby like", I did it this way to keep the example implementation code to the body of a single function, but I think it could be more useful to cache the "accepted file blueprints" in the parent model, rather than statically between the children. 
This could probably be done by adding something like a dedicated "acceptedFileBlueprints" method to the `HasFiles` trait, and cache the result there.

distantnative: @lukasbestle maybe also good to get your perspective on this: should we move all the logics to a `HasFiles::acceptedFileBlueprints()` and `File::blueprints()` would just call `$this->parent()->acceptedFileBlueprints()`?

distantnative: I think it might actually be harder to do it on the parent model: these lines https://github.com/getkirby/kirby/pull/6112/files#diff-3b700032469a1feb9c9a9eab21e85c3b799a16b107327bc5173c1e375e962f06R179-R180 won't be able to do in a method of the parent... so we would have to run similar code twice.

rasteiner: Oh, I see. 
It was just that `static` keyword inside of the function that smelled a bit for me; not sure why... Probably because it means only the `File::blueprints` function benefits from the cache.

I've gone a bit deeper, inside of the `blueprints()` function the problematic path is actually:
 - [`$blueprint->sections()`](https://github.com/getkirby/kirby/blob/4.0.3/src/Cms/File.php#L208C12-L208C34) 
 - [`$this->section($section['name'])`](https://github.com/getkirby/kirby/blob/main/src/Cms/Blueprint.php#L716C21-L716C53)
 - [`new Section($props['type'], $props)`](https://github.com/getkirby/kirby/blob/main/src/Cms/Blueprint.php#L706C10-L706C45)
 - [`parent::__construct($type, $attrs);`](https://github.com/getkirby/kirby/blob/main/src/Cms/Section.php#L46C3-L46C38)
 - [`$this->applyComputed($options['computed']);`](https://github.com/getkirby/kirby/blob/main/src/Toolkit/Component.php#L91C4-L91C46)
 - In there it's the [`data` computed prop of the `files` section](https://github.com/getkirby/kirby/blob/main/config/sections/files.php#L102-L138) 
 - In there it's split between `$panel->image(...)` (60%), `$file->mime(...)` (25%) and `$file->toSafeString(...)` (5%). The remaining 10% is more or less equally split between the rest. 

I guess the main problem is that for every file in the page, the `applyComputed` method is called for every file, twice: that's `O(2n²)`. So for 5 files the `data` computed prop of the files section is called 50 times (2*5²), for 100 files it's called 20'000 times. 

Options I see:
1. just go with the (perhaps smelly) local static WeakMap
2. split `File::blueprints` into a new `HasFiles::acceptedFileTemplates()` function call (which caches the results in the parent), then add the current file template and continue by starting from the line [`$templates = array_unique(array_filter($templates));`](https://github.com/getkirby/kirby/blob/4.0.3/src/Cms/File.php#L228).
3. have a cache for the `Section` **instances** in [`Blueprint::section(...)`](https://github.com/getkirby/kirby/blob/4.0.3/src/Cms/Blueprint.php#L693-L707), currently only the props are cached.

distantnative: 3. -> tried something, haven't encountered side regressions so far: https://github.com/getkirby/kirby/pull/6175

distantnative: @rasteiner I tried no. 2 in this PR and also no 3 https://github.com/getkirby/kirby/pull/6175

distantnative:
@bastianallgeier sorting request in the site area or would it make more sense to add API route support to sections (like we have for fields)?

bastianallgeier:
@distantnative you are right. Makes more sense!

src/Cms/File.php
lukasbestle: Haven't reviewed the full code of this PR so far, but at the first glance I very much like splitting up the method into multiple smaller ones. The previous/current implementation is *a lot* of logic in a single method.

distantnative: @lukasbestle That's already done

bastianallgeier:
Looking at this again, I was wondering if we should move the acceptedFileTemplates method to the Blueprint classes instead.

rasteiner:
> Looking at this again, I was wondering if we should move the acceptedFileTemplates method to the Blueprint classes instead.

Currently there's a one to one mapping between model instances and their blueprint (you create a `new` blueprint for every model instance, even if it's the "same" blueprint as that of other models). Therefore every blueprint instance actually refers to a single instance of a model. The inherent idea of "blueprints" would however suggest that the same blueprint would be used by multiple models. 

As long as there is this 1 to 1 mapping, I don't think it really makes a difference (other than semantics): it's the same code with the same outcome, simply executed in another place. 
If you intend to break this mapping in the future (for example to improve efficiency) and have the same blueprint instance for many models, placing the function in the blueprint rather than the model would make a difference if, for example, the accepted file templates depend on a specific model (e.g. a model accepting only a certain number of a file type).

distantnative:
@rasteiner but I think @bastianallgeier is right, cause even if blueprint object were unique per blueprint (and not 1:1 for each model), the logic in `acceptedFileTemplates` doesn't depend on the model but the blueprint.

distantnative:
@bastianallgeier I think I could use help with creating the API endpoint for the section (aka also first adding API endpoints feature for sections as we have for fields)

bastianallgeier:
Ok, for some reason I totally thought that we already had the section API, which is obvisouly not true 🙈 I think it's a good idea to add it to 4.2 as well and then get this part of the task done. It's not super complicated. But it's definitely beyond the scope of this PR. Let's wrap this one up first by adding the unit tests and reverting the breaking change in the API route. I will start a new branch with the section API. I've already started working on it.

bastianallgeier:
@distantnative I've added unit tests and refactored the detection for templates in fields a bit more. I also realised by doing so that we completely ignore fields in fieldsets so far. I would suggest to handle this in a separate PR though. It's going to be quite a challenge.

rasteiner:
> @rasteiner but I think @bastianallgeier is right, cause even if blueprint object were unique per blueprint (and not 1:1 for each model), the logic in `acceptedFileTemplates` doesn't depend on the model but the blueprint.

Sure, I wasn't saying that one approach is right or not. Just saying that one might have more future consequences than the other. 
Technically (even though this isn't implemented yet), having a page blueprint like:
```yml
sections:
  files:
    template: foo
    max: 1
```
could mean that some models would accept a new "foo" file, while others (which already have a "foo") might not.

I guess it's therefore debatable if "acceptedFileTemplates" is a property of the model or the blueprint: the blueprint specifies the same rules for many models, but those same rules might give different results depending on the model. So, finally, is it the blueprint which accepts a file template, or is it the model? (honest question)

Should you decide to decouple model and blueprint, having this functionality in the blueprint, when results depend on the model like in the example above, would bring back the need for a `static WeakMap<HasFiles, array>` to cache results in the blueprint. While having it in the model instead, the cache would stay a "simple" array. 

In the end, however, I guess it doesn't really matter and this is bikeshedding. If it were up to me to design this from ground up, I'd probably go for a façade function in the model which caches the results of a call to a blueprint function which accepts a model as parameter. (pseudocode: `return $this->cache = $this->blueprint()->acceptedFileTemplates($this);`).

distantnative:
But in your example, `foo` would always be an accepted file template from the model/blueprint. The specific section is defining a limit which shouldn't limit what this method returns. I know that this might open up possibilities where changing a file template somewhere ends up in this section having 2 entries. This however is really this area where we hit limits with the section based rules and we have to accept some shortcomings. We've seen it with other occasions, such things would only be solved properly with page-wide definitions, not sections.

distantnative:
@bastianallgeier I am fine with first merging this, but we should add an issue for the lack of checking field sets

src/Cms/Blueprint.php
```diff
@@ -157,6 +157,29 @@ protected function acceptedFileTemplatesFromFields(array $fields): array
 				$templates = array_merge($templates, $fields);
 				continue;
 			}
+
+			// layout and blocks fields
+			if (isset($field['fieldsets']) === true && is_array($field['fieldsets']) === true) {
+				$fieldsets = $this->acceptedFileTemplatesFromFieldsets($field['fieldsets']);
+				$templates = array_merge($templates, $fieldsets);
```
distantnative: Missing a chance to do 

```php
$templates = [
    ...$templates,
    ...$this->acceptedFileTemplatesFromFieldsets($field['fieldsets'])
];
```

:D

bastianallgeier: Much nicer!

bastianallgeier (approved):

bastianallgeier:
It felt better to solve this directly instead of moving it to yet another issue.

distantnative:
Looks good to me. Tests also only seem to fail because Codecov upload has a hiccup once again

PR #6113, opened 2024-01-01
Video block: support for local video files

### LINKED ISSUE #6062: Improve video block with support for local videos
## Description

While the current video block can include local videos by entering their direct URL it is quite tedious to do so. For a project of mine, I've added support for this locally - maybe you could re-use those parts?

First, I've extended the video blueprint similar to what the image block does:

```
name: field.blocks.video.name
icon: video
preview: video
fields:
  location:
    label: field.blocks.image.location
    type: radio
    columns: 2
    default: "web"
    options:
      kirby: "{{ t('field.blocks.image.location.internal') }}"
      web: "{{ t('field.blocks.image.location.external') }}"
  url:
    label: field.blocks.video.url.label
    type: url
    placeholder: field.blocks.video.url.placeholder
    when:
      location: web
  video:
    type: files
    query: model.videos
    multiple: false
    # you might want to add a template for videos here
    when:
      location: kirby
  poster:
    type: files
    query: model.images
    multiple: false
    image:
      back: black
    uploads:
      template: blocks/image
    when:
      location: kirby
  caption:
    label: field.blocks.video.caption
    type: writer
    inline: true
```

Next I've created a small plugin at `site/plugins/videopreview/index.php` in order to overwrite the video block preview:

```
<?php

Kirby::plugin('bii/videopreview', []);
```

Next, I've extended https://github.com/getkirby/kirby/blob/main/panel/src/components/Forms/Blocks/Types/Video.vue at `site/plugins/videopreview/index.js` in order to show the poster image and switch between local and remote videos:

```
panel.plugin("bii/videopreview", {
  blocks: {
    video: {
      computed: {
        captionMarks() {
          return this.field("caption", { marks: true }).marks;
        },
        video() {
          if (this.content.location === "web") {
            return this.$helper.embed.video(this.content.url ?? "", true);
          }

          if (this.content.video?.[0]?.url) {
            return this.content.video[0].url;
          }

          return false;
        },
        location() {
          return this.content.location;
        },
        poster() {
          if (this.content.poster?.[0]?.url) {
            return this.content.poster[0].url;
          }
        }
      },
      template: `
        <k-block-figure
          :caption="content.caption"
          :caption-marks="captionMarks"
          :empty-text="$t('field.blocks.video.placeholder') + ' …'"
          :is-empty="!video"
          empty-icon="video"
          @open="open"
          @update="update"
        >
          <k-frame ratio="16/9" v-if="location == 'web'">
            <iframe
              v-if="video"
              :src="video"
              referrerpolicy="strict-origin-when-cross-origin"
            />
          </k-frame>
          <k-frame ratio="16/9" v-if="location == 'kirby'">
            <video
              v-if="video"
              controls
              :src="video"
              :poster="poster"
            />
          </k-frame>
        </k-block-figure>
      `
    }
  }
});
```

Lastly, I prevent clicks on the video itself by adding the following to `site/plugins/videopreview/index.css`:

```
.k-block-figure video {
  pointer-events: none;
}
```

Everything works just fine for us like this, but it would be awesome if this could be integrated in Kirby as well.

## Your setup
**Kirby Version**  
4.0.1

distantnative:
## TBD
Blocks created before this change will not have any value selected for `location` in the drawer at first

### Enhancements
- Video block now also supports local videos by default

### For review team


afbora (commented):
- `Block error: "Call to a member function url() on null" in block type: "video"`

Poster is not required (agreed that poster shouldn't be required). But If I don't select a poster, getting error as expected. 

- I think also some attributes can be manageable with toggle fields such a like: autoplay, muted, controls, loop, preload (via select field).

bastianallgeier:
This is great, but it also really makes me wish that we had block controllers even harder.

distantnative:
@bastianallgeier I think that's its own topic and shouldn't be tied to this PR

@afbora will look into it, already think I know what to fix. And indeed maybe we should add some more fields for those attributes.

distantnative:
@afbora I think I fixed the error and also added fields for the attributes

afbora:
Works perfect! One more thing, now need translation strings for labels of new video attributes/fields.

afbora:
@distantnative Why you moved to 4.2.0? I think the PR is ready to merge for the 4.1.0

afbora (dismissed):

bastianallgeier:
@afbora I think the TBD part is still tricky. For existing blocks, the drawer shows a pretty weird state so far. I haven't found a good solution for this either.

distantnative:
Exactly that. And IMO it's not crucial to be in 4.1 - so instead of blocking 4.1-rc.1 I'd rather move it to 4.2

afbora:
Oh, I missed that part. You're right. Existing video blocks doesn't work with default `web` value 🤔

distantnative:
Not sure how to move this forward. I guess there is no way to make `web` a default for when no option is set?

afbora:
Not sure but can we apply a simple enhancement to the `when` prop? For ex: empty string or `null` check.

```yaml
when:
  location: ""
  
when:
  location: null
```

distantnative:
@afbora if we could, I think people would like this in general. Maybe really something to look into, if there is a way that's not too complicated/hacky

afbora:
@distantnative Sorry. That's not gonna work. We will have to do both empty check and value check.

```yaml
when:
  location: 
   - null
   - web
```

This is not an easy solution (https://kirby.nolt.io/13). We have to look for another way.

distantnative:
@bastianallgeier I have no good idea for the remaining TBD. Wondering if it's ok since the snippet will continue to work, just the Panel interface will not fully show first.

lukasbestle:
About the backwards-compatibility: Also not something for this PR, but could we make the form merge the existing block content with the defaults in the blueprint? Meaning that all fields that don't yet exist at all in the existing content are populated with the default. Whereas fields that do exist but are empty will stay empty.

distantnative:
It's even more difficult, because it's not like `location` would be empty/null if not set, but an empty string. And it'll be hard to differentiate between actual empty field (intentional empty string) and no value (still empty string, but only because this is how it gets passed around in our frontend).

afbora:
I've thought many times about how to solve this in a simple way. The only one I can think of is to support the Expression Language feature.

We can't use Javascript expression language. Because we need to use the same language to validate consistently in the backend and frontend when the form is submitted. If we use Symfony expression language we can send requests to the backend and process them.

https://symfony.com/doc/current/components/expression_language.html

````yaml
when: color in ['black', 'white']
````

distantnative:
I really thought about using Expression Language also in general for Kirby queries - to not have two similar systems side by side, but one consistent system. Unfortunately, Expression Language doesn't support a few things our queries are capable of (e.g. dropping empty `()` for method calls, inline closures, intercepting expression chains, easy top level functions).

It's a pity cause the other parts of expression language, especially the parsing are way superior to our regex classes.

afbora:
> Expression Language doesn't support a few things our queries are capable

OK. So what about [extending the ExpressionLanguage](https://symfony.com/doc/current/components/expression_language.html#extending-the-expressionlanguage)?

distantnative:
@afbora i read this but as far as I understand it, it only allows to add function calls that are available in the expressions. But not to change the things I mentioned (e.g. making `()` optional).

afbora:
I see. Actually even if we don't use this EL library, I think it would make sense to require `()`  when there are no arguments. This makes the difference between method and property and makes it easier to read the queries. I'm not sure 🤷‍♂️ Thinking loud.

distantnative:
I agree if we could start over fresh. But we have queries without `()` all over our existing Kirby sites. So requiring them would be a huge breaking change (and one that would maybe annoy people as they don't get any extra value out of it).

lukasbestle:
> It's even more difficult, because it's not like `location` would be empty/null if not set, but an empty string. And it'll be hard to differentiate between actual empty field (intentional empty string) and no value (still empty string, but only because this is how it gets passed around in our frontend).

But the stored data in the content file does not have the field set in the JSON at all, no? So only at some point the content gets normalized to the current blueprint – and at exactly that point, we'd need to substitute the default.

distantnative:
Yes but already in the backend. And I think changing that behavior of our forms will be a rather big breaking change with all kind of potential side effects. But yea I agree that it would be good, but brings us back to our form and field classes.

lukasbestle:
Yeah, that's what I meant with "not for this PR". ;)

I wonder if we really need to solve that before we merge this feature. Since the location field is not required, existing blocks without it can still be saved and will work as before, no? I assume it's just about the editing experience where all other fields are first hidden until a selection is made.

distantnative:
Right now the block preview also doesn't show but I think I can fix that.

Edit: ✅

distantnative:
I would think this is ready to be merged then.

bastianallgeier (approved):

PR #6116, opened 2024-01-03
add settings and method to sharpen images

dgsiegel:
## This PR …

As both SimpleImage and ImageMagick have capabilities to sharpen thumbnails, I'd like to make use of them. This PR adds a thumb config option `sharpen` which can be used to sharpen an image.

Will add tests, if this is something you'd like to include. Also, instead of giving a number between 0-100, we could also just use the default of 50 and make `sharpen` a boolean config option.

## Ready?

### For review team

### DIFF
--- a/src/Image/Darkroom.php
+++ b/src/Image/Darkroom.php
@@ -60,6 +60,7 @@ class Darkroom
 			'quality'     => 90,
 			'scaleHeight' => null,
 			'scaleWidth'  => null,
+			'sharpen'     => null,
 			'width'       => null,
 		];
 	}
@@ -91,6 +92,11 @@ class Darkroom
 		if (isset($options['bw']) === true) {
 			$options['grayscale'] = $options['bw'];
 			unset($options['bw']);
+		}
+
+		// normalize the sharpen option
+		if ($options['sharpen'] === true) {
+			$options['sharpen'] = 50;
 		}
 
 		$options['quality'] ??= $this->settings['quality'];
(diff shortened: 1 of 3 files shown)


src/Image/Darkroom.php
```diff
@@ -93,6 +94,11 @@ protected function options(array $options = []): array
 			unset($options['bw']);
 		}
 
+		// normalize the sharpen option
+		if ($options['sharpen'] === true) {
+			$options['sharpen'] = 50;
```
bastianallgeier, line 99: TBH I have no experience with the sharpen filter. Would you say that 50 is the most universal default?

dgsiegel: @bastianallgeier 50 is the default SimpleImage uses, see here: https://github.com/claviska/SimpleImage/blob/master/src/claviska/SimpleImage.php#L2059

The default value of 50 will be normalized to `0.5` with this line `$amount = max(1, min(100, $amount)) / 100;`. The resulting matrix the becomes (https://github.com/claviska/SimpleImage/blob/master/src/claviska/SimpleImage.php#L2064-L2067):

```
[-1, -1, -1],
[-1, 16, -1],
[-1, -1, -1],
```

Imagine, the Craft CMS image library, hardcodes the matrix with the same values: https://github.com/pixelandtonic/Imagine/blob/4d9bb596ff60504e37ccf9103c0bb705dba7fec6/src/Gd/Effects.php#L120

I guess the main question here would be if this value should be configurable or just simply a boolean.

bastianallgeier: Sounds good to me!

bastianallgeier:
Since there's no good method to contribute to your PR directly, I've created a new one to add sharpen to the media URLs and add unit tests: https://github.com/getkirby/kirby/pull/6164

Thank you very much for your help! We will make sure to mention you in the release notes for 4.1

dgsiegel:
> Thank you very much for your help! We will make sure to mention you in the release notes for 4.1

Thank you! Looking forward to the next release :)

PR #6118, opened 2024-01-04
PHPUnit reliability refactoring (part 1)

lukasbestle:
## Info

First of two PRs. The second one is already done and contains all the temp folder refactoring throughout all tests. To make reviewing easier, I've split the other changes out into this PR. Will push the second PR once this one is merged.

Would be good to have both reviewed and merged soon to avoid merge conflicts. 🙏

---

### Fixes

- Fix fallbacks in the `Kirby/Api/Api` and `Kirby/Form/Form` classes when no app object is loaded
- The CLI detection also works for processes not attached to a terminal (e.g. externally called from another script)

### Housekeeping

- Our PHPUnit tests can now be run individually with process isolation
- New internal PHPUnit extension to manage constants and the temporary folder
- Running the PHPUnit tests no longer spills files outside the repo path
- Replace remote testing domain with example.com
- Delete temporary testing files from the repo

## Ready?

### For review team

### DIFF
--- a/src/Cms/Core.php
+++ b/src/Cms/Core.php
@@ -310,13 +310,22 @@ class Core
 	 */
 	public function roots(): array
 	{
+		$indexClosure = function (array $roots) {
+			// prevent PHPUnit tests from accessing files outside the repo
+			if (defined('KIRBY_TESTING') === true && KIRBY_TESTING === true) {
+				return '/dev/null';
+			}
+
+			return dirname(__DIR__, 3); // @codeCoverageIgnore
+		};
+
 		return $this->cache['roots'] ??= [
 			'kirby'       => fn (array $roots) => dirname(__DIR__, 2),
 			'i18n'        => fn (array $roots) => $roots['kirby'] . '/i18n',
 			'i18n:translations' => fn (array $roots) => $roots['i18n'] . '/translations',
 			'i18n:rules'  => fn (array $roots) => $roots['i18n'] . '/rules',
 
-			'index'       => fn (array $roots) => dirname(__DIR__, 3),
+			'index'       => $indexClosure,
 			'assets'      => fn (array $roots) => $roots['index'] . '/assets',
 			'content'     => fn (array $roots) => $roots['index'] . '/content',
 			'media'       => fn (array $roots) => $roots['index'] . '/media',
--- a/src/Cms/Core.php
+++ b/src/Cms/Core.php
@@ -30,6 +30,11 @@ use Kirby\Form\Field\LayoutField;
  */
 class Core
 {
+	/**
+	 * Optional override for the auto-detected index root
+	 */
+	public static string|null $indexRoot = null;
+
 	protected array $cache = [];
 	protected string $root;
 
@@ -310,22 +315,13 @@ class Core
 	 */
 	public function roots(): array
 	{
-		$indexClosure = function (array $roots) {
-			// prevent PHPUnit tests from accessing files outside the repo
-			if (defined('KIRBY_TESTING') === true && KIRBY_TESTING === true) {
-				return '/dev/null';
-			}
-
-			return dirname(__DIR__, 3); // @codeCoverageIgnore
-		};
-
 		return $this->cache['roots'] ??= [
 			'kirby'       => fn (array $roots) => dirname(__DIR__, 2),
 			'i18n'        => fn (array $roots) => $roots['kirby'] . '/i18n',
 			'i18n:translations' => fn (array $roots) => $roots['i18n'] . '/translations',
 			'i18n:rules'  => fn (array $roots) => $roots['i18n'] . '/rules',
 
-			'index'       => $indexClosure,
+			'index'       => fn (array $roots) => static::$indexRoot ?? dirname(__DIR__, 3),
 			'assets'      => fn (array $roots) => $roots['index'] . '/assets',
 			'content'     => fn (array $roots) => $roots['index'] . '/content',
 			'media'       => fn (array $roots) => $roots['index'] . '/media',
--- a/src/Http/Environment.php
+++ b/src/Http/Environment.php
@@ -322,14 +322,15 @@ class Environment
 		}
 
 		// @codeCoverageIgnoreStart
-		if (php_sapi_name() === 'cli') {
+		$sapi = php_sapi_name();
+		if ($sapi === 'cli') {
 			return true;
 		}
 
 		$term = getenv('TERM');
 
 		if (
-			substr(php_sapi_name(), 0, 3) === 'cgi' &&
+			substr($sapi, 0, 3) === 'cgi' &&
 			$term &&
 			$term !== 'unknown'
 		) {
--- a/src/Http/Environment.php
+++ b/src/Http/Environment.php
@@ -322,10 +322,14 @@ class Environment
 		}
 
 		// @codeCoverageIgnoreStart
+		if (php_sapi_name() === 'cli') {
+			return true;
+		}
+
 		$term = getenv('TERM');
 
 		if (
-			substr(PHP_SAPI, 0, 3) === 'cgi' &&
+			substr(php_sapi_name(), 0, 3) === 'cgi' &&
 			$term &&
 			$term !== 'unknown'
 		) {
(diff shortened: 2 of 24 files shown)


src/Cms/Core.php
```diff
@@ -310,13 +310,22 @@ public function load(): Loader
...
 	public function roots(): array
 	{
+		$indexClosure = function (array $roots) {
+			// prevent PHPUnit tests from accessing files outside the repo
+			if (defined('KIRBY_TESTING') === true && KIRBY_TESTING === true) {
+				return '/dev/null';
+			}
+
+			return dirname(__DIR__, 3); // @codeCoverageIgnore
+		};
```
distantnative: You see any chance how we could keep this test-specific code out of our core classes?

src/Http/Environment.php
```diff
@@ -322,10 +322,14 @@ protected function detectCli(bool|null $override = null): bool
...
 		// @codeCoverageIgnoreStart
+		if (php_sapi_name() === 'cli') {
+			return true;
+		}
+
 		$term = getenv('TERM');
 
 		if (
-			substr(PHP_SAPI, 0, 3) === 'cgi' &&
+			substr(php_sapi_name(), 0, 3) === 'cgi' &&
```
distantnative: To avoid calling the function twice, we could store `$sapi = php_sapi_name()` first

tests/Cms/Site/SiteActionsTest.php
```diff
@@ -88,10 +88,10 @@ public function testSave()
 	public function testUpdate()
 	{
 		$site = $this->site()->update([
-			'copyright' => 2018
+			'copyright' => '2018'
 		]);
 
-		$this->assertSame(2018, $site->copyright()->value());
+		$this->assertSame('2018', $site->copyright()->value());
```
distantnative, line 94: Why this has to change? Just curious as it is outside the line of the other changes.

tests/Cms/System/LicenseTest.php
```diff
@@ -11,6 +11,8 @@
  */
 class LicenseTest extends TestCase
 {
+	public const TMP = KIRBY_TMP_DIR . '/Cms.License';
```
distantnative, line 14: Are we not doing `/Cms/License` as this would have to create nested folders? Otherwise, I think that would be preferable over the dot notation.

tests/Filesystem/FileTest.php
```diff
@@ -382,6 +387,9 @@ public function testMatchMimeException()
 		$this->expectException(Exception::class);
 		$this->expectExceptionMessage('Invalid mime type: text/plain');
 
+		// load translations
```
distantnative: Could you please rephrase this a little? From the code that follows it doesn't get clear why we would need the translations.

tests/Http/CookieTest.php
```diff
@@ -7,6 +7,11 @@
 
 class CookieTest extends TestCase
 {
+	public function tearDown(): void
+	{
+		Cookie::$key = 'KirbyHttpCookieKey';
```
distantnative: Would be safer for future changes if a `setUp` method stores the original key and `tearDown` restores that (whatever it initially was).

tests/PhpUnitExtension.php
```diff
@@ -0,0 +1,70 @@
...
+	{
+		// determine a unique path to a temporary directory
+		$tempDir = __DIR__ . '/tmp';
+
+		// when running via ParaTest, use a separate directory for each process
+		if (getenv('UNIQUE_TEST_TOKEN') !== false) {
+			$tempDir .= '/' . getenv('UNIQUE_TEST_TOKEN');
+		}
+
+		define('KIRBY_TMP_DIR', $tempDir);
```
distantnative: Feel free to dismiss this immediately as this is a super nitpick observation, but most times we abbreviate it `tmp`, but here in the variable `temp`

tests/Query/QueryDefaultFunctionsTest.php
```diff
@@ -16,6 +17,13 @@
  */
 class QueryDefaultFunctionsTest extends \PHPUnit\Framework\TestCase
 {
+	public const TMP = KIRBY_TMP_DIR . '/Query.QueryDefaultFunctions';
+
+	public function tearDown(): void
+	{
+		Dir::remove(static::TMP);
+	}
```
distantnative, line 25: Question for understanding: why do the individual test classes have to take care of removing the tmp folder, but not creating it?

distantnative:
Looks pretty good, left a few comments and questions.

Also

> Our PHPUnit tests can now be run individually with process isolation

I don't think this is in this PR yet, is it?

And as I understand this PR will fix the weird `site`  files plopping up in the sandbox? That alone would be gold.

src/Cms/Core.php
distantnative: Alternatively if this has to be here, we could move the check to a new `App::isRunningUnitTests(): bool` and then do

```php
'index'       => fn () => match ($this->isRunningUnitTests()) {
    true  => '/dev/null', // @codeCoverageIgnore
    false => dirname(__DIR__, 3); // @codeCoverageIgnore
}
```

lukasbestle: I pushed another suggestion that at least keeps the "testing mode" logic out of the core.

tests/Cms/Site/SiteActionsTest.php
lukasbestle: See the commit description:

> With process isolation, the behavior of the form classes is different and the value was converted to string. By already passing a string, the test succeeds consistently both with and without process isolation.

tests/Cms/System/LicenseTest.php
lukasbestle: Yes, with nested folders we can't clean them up as easily. The entire `tmp` directory is removed anyway after the test runs and I even considered just using `KIRBY_TMP_DIR` everywhere, but we have some tests that don't fully clean up after themselves, so separating each test class is safest – and that requires using one directory level, so no slashes.

tests/Filesystem/FileTest.php
lukasbestle: It's for the exception message that gets checked above. I've added that to the comments.

tests/PhpUnitExtension.php
lukasbestle: Good catch. Originally I had called the constant `KIRBY_TEMP_DIR`, but renamed it for exactly that reason. The variable was a leftover.

tests/Query/QueryDefaultFunctionsTest.php
lukasbestle: It depends. Some test classes do create the tmp folder because the tests only write files to it. But other classes like this one use the tmp folder as the index root of an app instance. The app automatically creates the folder if it doesn't exist. That's why creating is not needed here. But it still needs to be removed for the following tests in the same class and for general cleanup.

lukasbestle:
Thanks for your super quick and helpful review. 💛

> > Our PHPUnit tests can now be run individually with process isolation
> 
> I don't think this is in this PR yet, is it?

Partly. You are right that the full fix for process isolation will only be in the second PR, but I've already added it here because most fixes are contained in this PR and I wanted to avoid forgetting about this release note entry.

Process isolation is not fully equal to test parallelization though. Process isolation is `phpunit --process-isolation`, which runs all tests sequentially but every single one in a separate process. It's super super slow (~ 10 *minutes* on my machine), but helpful for debugging. Parallelization requires that all tests work with process isolation, but parallelization also has additional requirements that will be mostly met by the second PR. There still seems to be a bug in ParaTest that prevents us from using it already, but the refactorings are a good preparation in any case.

> And as I understand this PR will fix the weird `site` files plopping up in the sandbox? That alone would be gold.

Yes! 🎉

tests/Cms/Site/SiteActionsTest.php
lukasbestle: Don't ask me why the Form classes do that BTW. Wasn't able to track that down, but with the old form classes I also don't think it's really worth investigating. ;)

distantnative (approved):

PR #6124, opened 2024-01-05
DRY tests with `Kirby\TestCase`

distantnative:


tests/TestCase.php
```diff
@@ -0,0 +1,51 @@
...
+		}
+	}
+
+	/**
+	 * Checks if the test class extending this test case class
+	 * has defined a temporary directory
+	 */
+	protected function hasTmp(): bool
+	{
+		return defined(get_class($this) . '::TMP');
```
lukasbestle: Reading this again myself, I think this can be:

```suggestion
		return defined(static::class . '::TMP');
```

lukasbestle (commented):
Good idea. Can we replace all other uses of `PHPUnit\Framework\TestCase` with this as well to have the methods available throughout the system?

distantnative:
@lukasbestle change it so that all test classes use `Kirby\TestCase`. Added `assertIsPage` in a few places, but I think it's too big to change all tests to use these methods all at once.

bastianallgeier (dismissed):
This is great. I'm really happy that we are finally start cleaning up the tests.

bastianallgeier (approved):

PR #6125, opened 2024-01-06
Use more PHP arrow functions

distantnative:
### Housekeeping
- Using more PHP arrow functions in our code

### DIFF
--- a/src/Cms/ModelWithContent.php
+++ b/src/Cms/ModelWithContent.php
@@ -784,10 +784,17 @@ abstract class ModelWithContent implements Identifiable
 			]);
 		}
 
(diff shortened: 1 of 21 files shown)


bastianallgeier (commented):
TBH, I sometimes really find them harder to read. Do you think there's an actual benefit to convert them all?

distantnative:
Oh really? Didn't think that would be the case. For me they are easier to parse, because they are more compact and also less clutter (e.g. `use ()`).

bastianallgeier:
Getting rid of `use()` is indeed a good argument. I think after all it's about getting used to it – no pun intended :)

bastianallgeier (approved):

PR #6127, opened 2024-01-08
Update de.json: Voralge -> Vorlage

FynnZW:
Just a quick typo fix I noticed


distantnative:
We need to do those changes on our Translation tool Transfix. Just made the fix there - will be included with the next update. Thank you!

PR #6128, opened 2024-01-08
Fix overflow in narrow LayoutColumn with fields

distantnative:
### Fixes
- Fixes overflow in narrow layout field columns when using a block with `preview: fields`


bastianallgeier:
I've moved the media query to the form because it would otherwise affect the blocks focus outline with an overflow issue in the layouts field.

bastianallgeier (approved):

Issue #6117, opened 2024-01-04
Language argument doesn't passed to hooks from dialogs/areas


afbora:
## Description

When you try to update the model title from dialogs in multi language setup, `$languageCode` returns `null` always. I don't remember when it's broken but I think this is a regression.

````php
return [    
    'hooks' => [
        'site.changeTitle:before' => function (Kirby\Cms\Site $site, string $title, ?string $languageCode = null) {
             // language code returns null
        },
        'page.changeTitle:before' => function (Kirby\Cms\Page $page, string $title, ?string $languageCode = null) {
            // language code returns null
        }
    ]
];
````

## Your setup
**Kirby Version**  
3.9.10 & 4.0.2

afbora:
I'll implement but I've a question. In `page.changeSlug:before` hook, `$languageCode` returns as

PR #6130, opened 2024-01-09
Pass `languageCode` argument for some missing hooks

afbora:
## This PR …

Fixes missing `$languageCode` argument for `site.changeTitle:before` and `page.changeTitle:before` hooks.

### Fixes
- Language argument doesn't passed to hooks from dialogs/areas #6117

### Breaking changes

May be `$languageCode` in `Site::changeTitle()` type hint could be a breaking change, not sure.

## Ready?

### For review team

### DIFF
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -366,7 +366,19 @@ trait PageActions
 		string $title,
 		string|null $languageCode = null
 	): static {
+		// if the `$languageCode` argument is not set and is not the default language
+		// the `$languageCode` argument is sent as the current language
+		if (
+			$languageCode === null &&
+			$language = $this->kirby()->language()
+		) {
+			if ($language->isDefault() === false) {
+				$languageCode = $language->code();
+			}
+		}
+
 		$arguments = ['page' => $this, 'title' => $title, 'languageCode' => $languageCode];
+
 		return $this->commit('changeTitle', $arguments, function ($page, $title, $languageCode) {
 			$page = $page->save(['title' => $title], $languageCode);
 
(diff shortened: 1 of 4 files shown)


src/Cms/PageActions.php
```diff
@@ -366,7 +366,19 @@ public function changeTitle(
...
+		// if the `$languageCode` argument is not set and is not the default language
+		// the `$languageCode` argument is sent as the current language
+		if (
+			$languageCode === null &&
+			$language = $this->kirby()->language()
+		) {
+			if ($language->isDefault() === false) {
+				$languageCode = $language->code();
+			}
+		}
```
afbora, line 378: One liner if you want:

```php
$languageCode ??= $this->kirby()->language()?->isDefault() === false ? $this->kirby()->language()->code() : null;
````

bastianallgeier: Is it needed here to check for the default language? Couldn't we just always send the code in multi-language mode?

afbora: I already thought like that but we had implemented like in the pr: https://github.com/getkirby/kirby/blob/4.0.3/src/Cms/PageActions.php#L161

afbora: I already thought like that but we had implemented like in the pr: https://github.com/getkirby/kirby/blob/4.0.3/src/Cms/PageActions.php#L161

bastianallgeier: I'm not sure I understand what you mean

afbora: Language code passed as null only when not multi-lang or default language for `changeSlug` action.

bastianallgeier (approved):

PR #6147, opened 2024-01-15
Pages section table: add field types for columns

### LINKED ISSUE #6120: v4 color field previews
## description
there is a difference when previewing `color` fields from a `structure` field vs. a `pages` section. for `structure` fields, the **color itself** is shown in the table preview. for `pages` sections, only the **color value** is shown in the table preview.

**expected behavior**  
i was expecting the table preview for `color` fields to be the same for both `structure` fields and `pages` sections. i don’t know if this is how you want it to be or not? i found it a little confusing though.

**screenshots** 
structure field
structure

pages section
pages

## to reproduce
**structure field**
```yml
# categories.yml
sections:
    categories:
        type: fields
        fields:
            categories:
                type: structure
                fields:
                    title:
                        type: text
                    color:
                        type: color
```

**pages section**
```yml
# categories.yml
sections:
    categories:
        type: pages
        layout: table

# category.yml
sections:
    category:
        type: fields
        fields:
            color:
                type: color
```

## to fix
i already found an easy fix, which however does not seem to appear in the current [docs](https://getkirby.com/docs/reference/panel/sections/pages#layout__table-layout). maybe it would be nice to let everyone know that setting `type: color` for the column is allowed and works fine?

**pages section**
```yml
# categories.yml
sections:
    categories:
        type: pages
        layout: table
        columns:
            color:
                type: color
```

fix

distantnative:
### Enhancement
- Page and files section with table layout will now try to infer the column type from the blueprint of the first model when no type has been provided explicitly

### Breaking changes
I think none as `columnsWithTypes` is a new function that our pages/files section uses - but any plugin extending those sections can still use `columns` like before. 

### For review team

### DIFF
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -55,7 +55,7 @@ return [
 		'parent' => function () {
 			return $this->parentModel();
 		},
-		'files' => function () {
+		'models' => function () {
 			if ($this->query !== null) {
 				$files = $this->parent->query($this->query, Files::class) ?? new Files([]);
 			} else {
@@ -99,6 +99,9 @@ return [
 
 			return $files;
 		},
+		'files' => function () {
+			return $this->models;
+		},
 		'data' => function () {
 			$data = [];
 
@@ -106,7 +109,7 @@ return [
 			// a different parent model
 			$dragTextAbsolute = $this->model->is($this->parent) === false;
 
-			foreach ($this->files as $file) {
+			foreach ($this->models as $file) {
 				$panel = $file->panel();
 
 				$item = [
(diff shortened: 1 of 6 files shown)


bastianallgeier:
What happens when one of the pages has a different blueprint with different field types?

distantnative:
@bastianallgeier Would clash. But no idea how to resolve that.

bastianallgeier:
I wonder if the table layout should actually be limited to pages sections that only define a single template. That would make a lot of sense, not only if we implement this.

bastianallgeier (approved):

PR #6148, opened 2024-01-15
PageCreateDialog: use field defaults

### LINKED ISSUE #6119: Publishing a page using `create.status` option doesn't work if page has required fields & defaults for them
## Description
I have a page with fields a & b. In the blueprint, I define

```yaml
create:
  status: listed
```

to publish the page directly after creating it.
Field a is a required field and therefore creation of the page fails, because it's not filled.

I want to mitigate this issue by adding a default for field a.
Because Kirby (afaik) handles default data from blueprints in the panel as well, it won't be respected by Page::create() and it still doesn't work. There is no issue when setting the status from the panel, as Kirby saves the default data from the blueprint correctly.

## Your setup
**Kirby Version**  4.0.2

distantnative:
### Fixes
- Page create dialog: custom fields in dialogs are prefilled with default values
- Page create dialog: no error is thrown when directly publishing page with required fields that have a default value 
#6119

## Ready?

### For review team

### DIFF
--- a/src/Panel/PageCreateDialog.php
+++ b/src/Panel/PageCreateDialog.php
@@ -236,8 +236,12 @@ class PageCreateDialog
 			$content[$name] = $input[$name] ?? null;
 		}
 
+		// create temporary form to sanitize the input
+		// and add default values
+		$form = Form::for($this->model(), ['values' => $content]);
+
 		return [
-			'content'  => $content,
+			'content'  => $form->strings(true),
 			'slug'     => $input['slug'],
 			'template' => $this->template,
 		];
--- a/src/Panel/PageCreateDialog.php
+++ b/src/Panel/PageCreateDialog.php
@@ -305,7 +305,7 @@ class PageCreateDialog
 
 	public function value(): array
 	{
-		return [
+		$value = [
 			'parent'   => $this->parentId,
 			'section'  => $this->sectionId,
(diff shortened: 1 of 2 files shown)


src/Panel/PageCreateDialog.php
```diff
@@ -301,13 +305,20 @@ public function validate(array $input, string $status = 'draft'): bool
...
 			'section'  => $this->sectionId,
 			'slug'     => $this->slug ?? '',
 			'template' => $this->template,
 			'title'    => $this->title ?? '',
 			'view'     => $this->viewId,
 		];
+
+		// add default values for custom fields
+		foreach ($this->customFields() as $name => $field) {
+			$value[$name] = $field['default'] ?? null;
```
bastianallgeier: I think this is risky because some fields don't really like null as value. I would probably create an if clause and only add it to the array if it exists. 

```suggestion
if (isset($field['default']) === true) {
  $value[$name] = $field['default'];
}
```

distantnative: Good idea

distantnative: Added the change

bastianallgeier (approved):
Works great! I've added examples for all page creation options to the sandbox, so we can better test them in the future.

Issue #5946, opened 2023-11-09
Cropping acts weird on cross-orientation crops with and without focus


nilsmielke:
## Description
I have some weird cropping issues in beta.3. Am cropping to a landscape format like so:
```
<img src="<?= $slide->bild()->toFile()->crop(1280, 320)->url() ?>" 
     alt=""
     srcset="<?= $slide->bild()->toFile()->srcset([
                    '400w'  => ['width' => 400, 'height' => 100, 'crop' => true,],
                    '800w'  => ['width' => 800, 'height' => 200, 'crop' => true,],
                    '1200w' => ['width' => 1280, 'height' => 320, 'crop' => true]
                 ]) ?>"
     sizes="(max-width: 1280px) 100vw, 1280px"
>
```

If the image chosen is a portrait format, that exceeds a certain size (tried with 4000 x 6000), the cropping goes bonkers, producing an extreme portrait format of 437 x 1748 px.
Then again I tried a 5462 × 8192 px image, which only acts weird, when a focus point is set, but generates perfectly without one. When focus is set, this image doesn’t seem to be cropped at all. It is named “pexels-cedric-fauntleroy-4266947-1280x320-crop-36-5-65-5.jpg” in the media folder, though.
 
I have deleted the media folder several times, to make sure. Doesn’t improve the situation.
Also tried `<?= $slide->bild()->toFile()->crop(1280, 320) ?>` instead of the above. Same result.

Landscape images seem to always work as expected. 

(This might be related to https://github.com/getkirby/kirby/issues/5855.)

**Expected behavior**  
All images should be cropped to the requested output dimensions.

## Setup
**Kirby Version**  
4.0.0-beta.3

**System**
 - Device: macBook Pro 13"
 - OS: macOS 13.5.2
 - Browser: Firefox
 - Version: 119.0

distantnative:
@nilsmielke could you please provide an example to work with? E.g. "use this picture with the following command"? That would help a lot

nilsmielke:
@distantnative Sure thing. In my case it is this image: https://www.pexels.com/de-de/foto/frauen-tablet-arzt-gerat-4266947/ (free download, not sure, if their license would allow for it to be uploaded here)
[...]

Issue #5855, opened 2023-10-23
[v4] Cropping a jpg with focus point results in wrong aspect ratio


mrflix:
## Description
So I'm cropping an image to **512x683**:

![in](https://github.com/getkirby/kirby/assets/60777/2b7e03d9-d08a-4839-9352-96673cd393bd)

```php
$image->crop(512, 683, ['crop' => '35.7% 48.1%']);
```

**Expected behavior**  
A 3/4 ratio image of the size 512x683:
![correct](https://github.com/getkirby/kirby/assets/60777/7f5e2449-ac2d-4358-b49c-c1e7500956b1)

**Actual behavior**
And thumbs says "best I can do is **732x683**":

![out](https://github.com/getkirby/kirby/assets/60777/41f8c725-893e-42de-8205-b65edca9bf63)

## To reproduce

1. it has to be a jpg image
2. there has to be a focus point set
3. you have to use ImageMagick
4. ...
5. fun

## Your setup
**Kirby Version**  
v4 beta 2

## Digging 
So I investigated a bit. This is the ImageMagick command that the php statement produces:
```
convert -limit thread 1 -define 'jpeg:size=512x683' 'in.jpg' -auto-orient -crop 1770x2362+379+0 -resize 512x683^ -quality '85' 'out.jpg'
```

Let's look at the relevant parts:
```
-define 'jpeg:size=512x683'
```
[jpeg:size](https://imagemagick.org/script/command-line-options.php?#define) is a performance optimization setting that gets added to jpg files. Based on it, the jpg loader will load the image smaller to save memory. It was introduced to Kirby with this [PR](https://github.com/getkirby/kirby/pull/4106).

```
-crop 1770x2362+379+0 -resize 512x683^
```
The image is 3543x2362. The setting correctly tells the process to cut out a 1770x2362 rectangle in the desired aspect ratio of 0.75.
`+379+0` offsets this rectangle based on the focus point of 35.7% 48.1% (it moves the frame a bit left from the center).
`-resize 512x683^` scales it down.

All looking good.

Now, if we remove the performance optimization `-define 'jpeg:size=512x683'`, the resulting image has the correct aspect ratio. I'm not 100% sure yet why but I guess it has to do with `jpeg:size` scaling the image down when loading it into memory and `-crop 1770x2362` working with the original image sizes.

For comparison a crop call without focus point:

```
[...]

PR #6159, opened 2024-01-16
Better `Image\Focus` class

distantnative:
## This PR …

Trying to do a better job than the `Focus` class. Logic to focus crop is as follows:
Screenshot 2024-01-16 at 13 40 29

### TODO
- Check to see if the following issues are solved:
  - [x] https://github.com/getkirby/kirby/issues/5855
  - [x] https://github.com/getkirby/kirby/issues/5946
- [x] Bring back `ratio` helper to avoid division by 0

## Ready?

### For review team


src/Image/Darkroom/GdLib.php
```diff
@@ -62,34 +64,12 @@ protected function resize(SimpleImage $image, array $options): SimpleImage
...
-			return $image->thumbnail($options['width'], $options['height']);
-		}
+		// first crop the image in the right ratio
+		$crop  = Focus::fromDarkroomOptions($options);
+		$image = $image->crop($crop->x1, $crop->y1, $crop->x2, $crop->y2);
 
-		// normal crop with crop anchor
-		return $image->thumbnail(
-			$options['width'],
-			$options['height'] ?? $options['width'],
```
distantnative, line 91: @bastianallgeier I am wondering about this old fallback... is it needed? Cause a few lines up we don|t use one for the no-crop variant. And `Darkroom::preprocess()`, which is providing the options, calls `Dimensions::height()` which always returns an integer.

bastianallgeier: The idea behind using the height as fallback is to create a square crop if no height is specified.

distantnative: But based on my description, do you see any possibility how that key could not exist? I think there is just no way

bastianallgeier: Sorry, now I understand what you mean. You are right. This is no longer needed here.

distantnative:
Currently, this is only making some code nicer but not fixing any of the mentioned issues. Closing this for now.

PR #6162, opened 2024-01-16
BlockTitle: use `label` as fallback for `name`

### LINKED ISSUE #6137: Blueprints: fix Block naming inconsistency
## Description
Currently in blueprints, every field is named by its label. Except blocks. Blocks are named by their title and the label value is used as a subtitle (TIL https://github.com/getkirby/kirby/issues/4194). This leads to developer mistakes when naming custom blocks, because one would expect that the label value would also name blocks.

**Expected behavior**  
I'd expect to be able to use label to name a custom block. For example by falling back to use label when a label and no title are defined.

## Your setup
**Kirby Version**  
4.0.2

distantnative:
### Enhancement
- When a block fieldset doesn't set a `name`, the `label` is now used as fallback
 #6137

### For review team

- [x] Add example to Lab
- [x] Add task to document `name` on website

### DIFF
--- a/panel/lab/components/block/4_title/index.vue
+++ b/panel/lab/components/block/4_title/index.vue
@@ -14,19 +14,19 @@
 				:fieldset="{
 					icon: 'image',
 					name: 'image',
-					label: 'This is an image block',
+					label: 'This is an image block'
 				}"
 			/>
 		</k-lab-example>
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
This is great.

Issue #6041, opened 2023-11-30
Block & layout field: help text centered


distantnative:
<img width="1115" alt="Screenshot 2023-11-30 at 22 42 07" src="https://github.com/getkirby/kirby/assets/3788865/c2a1320c-2654-4224-a22d-4ca6264bb5d2">
<img width="1117" alt="Screenshot 2023-11-30 at 22 42 18" src="https://github.com/getkirby/kirby/assets/3788865/8e8cd782-2317-4a12-aa2b-e672312c235c">

distantnative:
@bastianallgeier by now I am wondering what the "real" version actually should look like

PR #6166, opened 2024-01-17
Fix the help text position in block and layout fields

bastianallgeier:
## This PR …

The most sensibel fix is to move the button over to the right if there's help text – at least in my opinion :)

Screenshot 2024-01-17 at 16 22 55

### Fixes

- https://github.com/getkirby/kirby/issues/6041

### For review team

### DIFF
--- a/panel/src/components/Forms/Field/BlocksField.vue
+++ b/panel/src/components/Forms/Field/BlocksField.vue
@@ -30,15 +30,23 @@
 			v-on="$listeners"
 		/>
 
-		<footer v-if="!disabled && !isEmpty && !isFull && hasFieldsets">
-			<k-button
-				:title="$t('add')"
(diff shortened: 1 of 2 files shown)


distantnative:
Hmmm really not sure. I don't think we'll make a lot of friends with shoving the button all to the right from being in the center.

distantnative:
I'm gonna that the current version is probably the better one, in my opinion.

bastianallgeier:
Yeah, you might be right. It's one of those situations were both solutions kind of suck.

distantnative:
After all it was just me raising the question/issue about the current situation. Haven't heard much complaints from the community

bastianallgeier:
Ok, lets close the PR and issue and if it ever comes up again, we can think about it again.

PR #6170, opened 2024-01-18
Fix lab: dumber examples

distantnative:


panel/lab/components/fields/color/index.vue
```diff
@@ -1,72 +1,148 @@
...
+			<k-lab-example label="Autofocus">
+				<k-color-field
+					:autofocus="true"
+					label="color"
+					:value="value"
+					@input="value = $event"
+				/>
+			</k-lab-example>
+
+			<k-lab-example label="Required">
```
bastianallgeier, line 22: There's an issue with the required state. I think I added some CSS at some point to highlight invalid inputs and the wrapper component unset it properly. Now it's no longer unset. 

Screenshot 2024-01-22 at 11 02 39

panel/lab/components/fields/object/index.vue
```diff
@@ -3,24 +3,18 @@
...
 		<k-lab-example label="Default">
-			<k-object-field :fields="fields" label="Object" />
+			<k-object-field
+				:fields="[
+					{
+						label: 'Name',
+						name: 'name',
+						type: 'text'
+					}
+				]"
```
bastianallgeier: This was already wrong before. The fields need to be defined as an object.

panel/lab/components/fields/radio/index.vue
```diff
@@ -1,3 +1,90 @@
 <template>
-	<k-lab-options-field-examples type="radio" />
+	<k-lab-form>
```
bastianallgeier, line 2: Something is strange here and I don't know what it is, but the examples without columns don't have a proper width for each option.

panel/lab/components/inputs/timeoptions/index.vue
```diff
@@ -1,7 +1,60 @@
...
+	data() {
+		return {
+			value: null
+		};
+	}
+};
+</script>
+
 <style>
 .k-lab-input-examples .k-timeoptions-input {
```
bastianallgeier: Those styles are no longer applied

panel/lab/components/fields/radio/index.vue
distantnative: Is that different to the current live lab?

bastianallgeier: You are right, it actually isn't

panel/lab/components/fields/color/index.vue
bastianallgeier: I removed the outline. I think it's too global and not super helpful after all. When we look closer into validation in your PR, we might find better solutions to test this.

bastianallgeier (approved):

PR #6178, opened 2024-01-20
`file` KirbyTag: fix fallback for not found file

### LINKED ISSUE #6177: File kirbytag returns null and throws an error when no valid file and text attr present
## Description
When a file tag `(file: non-existing.jpg)` doesn't have a valid file object, the `text` attribute is returned, but this returns `null` instead of an empty string (` kirby/config/tags.php`, #64) when this attribute is not present and therefore throws an error down the line.

See also: https://forum.getkirby.com/t/bnomei-rss-feed-not-working-since-upgrade-to-kirby-4/30543/1

## Your setup
**Kirby Version**  
4.0.3

distantnative:
### Fixes
- `file` KirbyTag that references a non-existing file does not throw an exception anymore, when not `text` attribute passed
#6177 

## Ready?

### For review team

### DIFF
--- a/config/tags.php
+++ b/config/tags.php
@@ -62,7 +62,7 @@ return [
 		],
 		'html' => function (KirbyTag $tag): string {
 			if (!$file = $tag->file($tag->value)) {
-				return $tag->text;
+				return $tag->text ?? $tag->value;
 			}
 
 			// use filename if the text is empty and make sure to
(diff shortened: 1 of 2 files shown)


afbora (approved):

afbora:
This bug-fix could be move into 4.0.4 release.

bastianallgeier:
@afbora we merge 4.0.4 and 4.1.0 anyway. So I think it's fine.

PR #6180, opened 2024-01-21
Performance testing with `phpbench`

distantnative:
## This PR …

Actual tests to be added in future PRs (when improving parts or when ensuring a change doesn't worsen performance). So far only to run locally. Integration into CI makes sense once we have some tests and baselines, I'd say.

### Housekeeping
- Added setup to do performance tests via `phpbench`

### For review team

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,7 +11,7 @@ phpbench.json export-ignore
 phpmd.xml.dist export-ignore
 phpunit.xml.dist export-ignore
 psalm.xml.dist export-ignore
-performance/ export-ignore
+bench/ export-ignore
 tests/ export-ignore
 
 # panel
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,7 +6,6 @@
 # tests
 .codecov.yml export-ignore
 .php-cs-fixer.dist.php export-ignore
-.phpbench export-ignore
 phpbench.json export-ignore
 phpmd.xml.dist export-ignore
 phpunit.xml.dist export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,9 +6,12 @@
 # tests
 .codecov.yml export-ignore
 .php-cs-fixer.dist.php export-ignore
+.phpbench export-ignore
+phpbench.json export-ignore
 phpmd.xml.dist export-ignore
 phpunit.xml.dist export-ignore
 psalm.xml.dist export-ignore
+performance/ export-ignore
 tests/ export-ignore
 
 # panel
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 /.idea
 
 # tests
+.phpbench
 .phpunit.cache
 /tests/coverage
 /tests/tmp
--- a/bench/bootstrap.php
+++ b/bench/bootstrap.php
@@ -1,0 +1,3 @@
+<?php
+
+require_once __DIR__ . '/../vendor/autoload.php';
--- a/composer.json
+++ b/composer.json
@@ -90,7 +90,6 @@
 		]
 	},
 	"scripts": {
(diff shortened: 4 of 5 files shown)


bastianallgeier (dismissed):

bastianallgeier:
I think I would wait for @lukasbestle to give his comments as well.

lukasbestle:
Very much in favor for this. I had looked into PHPBench a few years ago, but back then we didn't really have the capacity to work on test cases and now I already forgot all the config stuff. Would it be OK to move this to v4.2 given that there are no tests yet anyway?

distantnative:
I have a whole branch using it already, but wanted to isolate this first so that the branch/PR for the other part would be cleaner. So ideally I'd like to merge it after 4.1 as is and then rebase the other branch having a PR with first tests.

.gitattributes
```diff
@@ -6,9 +6,12 @@
 # tests
 .codecov.yml export-ignore
 .php-cs-fixer.dist.php export-ignore
+.phpbench export-ignore
```
lukasbestle: As this is already in `.gitignore`, can there be a case where this file is in the repo and would be exported? I would think that we don't need to add it again here:

```suggestion
```

.gitattributes
```diff
@@ -6,9 +6,12 @@
...
+phpbench.json export-ignore
 phpmd.xml.dist export-ignore
 phpunit.xml.dist export-ignore
 psalm.xml.dist export-ignore
+performance/ export-ignore
```
lukasbestle: Shouldn't this be:

```suggestion
bench/ export-ignore
```

lukasbestle (commented):
In general it looks good to me. I like the idea to separate the benchmarks into a separate `bench` folder.

Config is certainly opinionated and we'll have to see if the defaults we set will fit us. It feels a bit like a process of "you can't believe statistics that you haven't fudged yourself". 🤣 But we can change the config at any time if we see a reason for it, so it's totally fine.

IMO we can still merge this into 4.1.0 as it doesn't affect the code.

.gitattributes
distantnative: Ah it's in there because I think eventually it shouldn't be in there `.gitignore`.

If we start using a baseline to compare against, that baseline data should probably be committed as well. Or we find a way how only the CI caches a baseline.

lukasbestle: I feel like running this in CI will be difficult as GitHub Actions workers are shared infrastructure, so probably not stable enough to measure performance consistently. I have no experience with performance regression testing though, so no idea how other projects do it.

lukasbestle: About the baseline: Doesn't it depend on the system a lot? If I have an environment that's twice as fast as another, the baseline cannot be compared. So wouldn't we need to run both runs (before and after) on the same machine right after another? Which would mean that committing the baseline wouldn't help us.

distantnative: True. Could also make the CI runs really long if we add many performance tests. I guess the question then is whether we'd want to commit (and update) a baseline as part of the repo. Or ignore that folder and everyone creates their own baseline with the main branch and compares their changed code against that baseline locally.

distantnative: Valid point about the system.

lukasbestle: Think our comments have crossed paths.

About the workflow: Maybe we incorporate it into the release workflow/script and create an issue for the next release if we find regressions? Or we run it on GitHub Actions but not in PRs, instead on regular schedule and then notify us if one of the `develop` branches has regressions. Would be helpful to see how other projects do it.

distantnative: Indeed. Let's try that once we have the first tests added.

I can already see that we'll have quite a few trial - learn more - refactor phases until we arrive at a state with tests that really makes sense.

lukasbestle (approved):

Issue #6169, opened 2024-01-18
Link field `id` missing


distantnative:
> Property or method "id" is not defined on the instance but referenced during render

PR #6182, opened 2024-01-22
Add link field id and other default input props

bastianallgeier:
### Fixes

- Add missing link field id https://github.com/getkirby/kirby/issues/6169

### Breaking changes

None

### For review team


distantnative:
I think by doing this it's loosing the `invalid` prop. I think to be consistent with other field, it would need to import the mixing as well as the prop son Input.vue

bastianallgeier:
Good point! I've added the input props back in. I think it's showing once more that we should really start working on the inputboxes refactoring to get all of this cleaned up.

distantnative (approved):

PR #6195, opened 2024-01-23
ColorField: proper HTML/XSS handling

### LINKED ISSUE #6183: Remaining inconsistencies of the color field
I have retested the state of `develop-minor` after #6152 with the HTML examples in the XSS lab, specifically the test cases I wrote down in the issue here: https://github.com/getkirby/kirby/pull/6056#issuecomment-1867946384

The new implementation unfortunately still suffers from HTML rendering issues:

- [x] All examples except the "Color (query, defaults)" example seem to double-escape the label. So we can render the label as HTML in Vue.
- [x] The "Color (query, defaults)" example does not seem to double-escape, which means that the default query does not escape on the backend. If I remember correctly with the other options fields, the query mode needs to escape on the backend so that the frontend can safely render as HTML.

To reproduce:
- Pull the latest commits of the Sandbox
- (Re)Load the Lab environment
- Go to the XSS page and look at the 8 color fields
  - Everything labeled "This is OK" should be rendered as HTML and displayed in green
  - The "API (with HTML)" field should also render the bold part in the label
  - No red "Malicious string" should be rendered in red

distantnative:
### Fixes
- https://github.com/getkirby/kirby/issues/6183

### For review team

### DIFF
--- a/config/fields/color.php
+++ b/config/fields/color.php
@@ -3,6 +3,7 @@
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Field\FieldOptions;
 use Kirby\Toolkit\A;
+use Kirby\Toolkit\Escape;
 use Kirby\Toolkit\Str;
 
 return [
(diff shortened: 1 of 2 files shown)


bastianallgeier:
I reviewed it and all the expectations from @lukasbestle' issue are met in my opinion. But I would leave the final approval to Lukas.

lukasbestle (approved):
Tested it again with the XSS sandbox and works great now 👍

Issue #5328, opened 2023-06-27
IM driver crashing php process when creating many thumbs for the first time


iskrisis:
## Description
I can't pin point out why or when but somewhere around v3.8 something about thumb creation changed. I know it from upgrading some image heavy sites to 3.8+. I have 4gig ram VPS and use `im` driver for most stuff to create avif/webp/jpg thumbs. It takes some time first time but it never was an issue. But with recent versions one can crash PHP process. Happened on both `php-fpm` and also clients Apache. Firefox starts to throw `NS_BINDING_ABORTED` erros for the images which i suspect is because error messages get served instead of image.

After asking @bastianallgeier he suspects it's 
https://github.com/getkirby/kirby/blob/6b5dda600472b59a3a779f7e6b8e9e130cc414a1/src/Cms/Media.php#L56C4-L56C4  change that loads/serves original files through PHP. 

I've tried to setup `im` security policies really thoroughly and i don't think it makes much difference. The `im` processes from logs seems to go fine and it never crashes so i don't think it's a problem with `im` driver. 

Curiously it doesn't seem to be a problem with `gd` but maybe thats just because `gd` can't create `.avif` which is the format that takes a long time to create.

I am not sure why this change has been made but i've managed to crash two live sites already when they launched. It's a bit nerve wrecking and really hard to track down. It would be really great if it worked as it did in the past.

iskrisis:
Can i help this somehow? Like would it help if i created test repo with example that was crashing?

distantnative:
@iskrisis that would help indeed

tobimori:
I think I had experienced similiar issues in the past, but I couldn't pinpoint it (as they're always gone after the first load)

bastianallgeier:
I agree. A test repo would be great. So far, I didn't run into this issue with our sandbox environments.

iskrisis:
Playing with it whole afternoon. I am starting to think this is not fixable unless some architectural change would happen.

## What i think is happening
When you have list page with enough images each of the thumb requests will spawn `convert` process
![iTerm2 - ploi@helpful-farm- ~-kirbyimagetest floriankarsten dev-media 2023-07-06 at 4 06 06 PM](https://github.com/getkirby/kirby/assets/4954323/f959ae3f-6b98-40ba-97fa-eec9637dbe1b)

[...]

PR #6205, opened 2024-01-25
ImageMagick: remove optimization that breaks focus

distantnative:
## This PR …

The optimization step only loads a minimized version of the image into memory which is why cropping (esp. focus cropping) fails as it is done in reference to the original sized image.

Without an apparent fix, I think it would be good to remove this optimize step for now.

### Fixes
- https://github.com/getkirby/kirby/issues/5855

### DIFF
--- a/src/Image/Darkroom/ImageMagick.php
+++ b/src/Image/Darkroom/ImageMagick.php
@@ -65,14 +65,6 @@ class ImageMagick extends Darkroom
 		// limit to single-threading to keep CPU usage sane
 		$command .= ' -limit thread 1';
 
-		// add JPEG size hint to optimize CPU and memory usage
-		if (F::mime($file) === 'image/jpeg') {
-			// add hint only when downscaling
-			if ($options['scaleWidth'] < 1 && $options['scaleHeight'] < 1) {
-				$command .= ' -define ' . escapeshellarg(sprintf('jpeg:size=%dx%d', $options['width'], $options['height']));
-			}
-		}
-
 		// append input file
 		return $command . ' ' . escapeshellarg($file);
 	}


lukasbestle:
I'm confused: #5946 says the issue appears in both GD and IM, but in your last comment you couldn't reproduce it. Does this instead fix #5855?

What I'm wondering though: Could it worsen the thumb performance and make issues like #5328 come up more? I don't have a good solution either, however I feel like we would need to trade one bug for another here.

distantnative:
Sorry, yes, wrong issue linked. Corrected it.

Certainly it could. And I don't like removing this either. But as I see it we have one bug the certainly breaks the thumb generation (right now) vs. one that causes problem with system performance limits. If we don't find a fix keeping the optimize flag - and I haven't had any success, I think the latter issue is the lesser evil.

lukasbestle (approved):
Yes, that's right.

I'm fine with this change, let's see what the real world implications are. We can still add it back if necessary. I first thought whether to comment out the optimization code instead, but then we'll have dead code lying around for the foreseeable future. So removing it completely for now is probably best.

bastianallgeier:
Two thoughts: 

1) is it good to add this to 4.1.0 shortly before the release? 
2) maybe we need to have a closer look at the imagick extension for a - hopefully - more stable alternative to use imagemagick. We could still keep the im driver for whenever the imagick extension isn't there though. But that's of course beyond the scope of this PR.

distantnative:
@bastianallgeier to my understanding thumbs are mostly broken for im users right now... so that's why I thought maybe it's not the best to have it broken still until 4.2

bastianallgeier:
@distantnative ok, that sounds like a very reasonable argument 🙈

bastianallgeier:
Focus definitely works well for me with this PR. I would say we give it a go. Maybe we should also have a close look at the suggested threads PR soon.

PR #6207, opened 2024-01-25
Improve input disabled styles

### LINKED ISSUE #6163: Date field: don't show placeholder when disabled
Potentially also relevant for other fields that have default placeholders...

I think when a (date) field is disabled, there should be no placeholder shown, but empty fields:

Screenshot 2024-01-16 at 21 46 12

distantnative:
### Fixes
- Disabled inputs don't show placeholders anymore
- Choice inputs have similar disabled styling as other fields
- #6163

### For review team

### DIFF
--- a/panel/src/components/Forms/Field/LinkField.vue
+++ b/panel/src/components/Forms/Field/LinkField.vue
@@ -29,7 +29,7 @@
 						:value="value"
 						@remove="removeModel"
 					>
-						<template v-if="!disabled" #placeholder>
+						<template #placeholder>
 							<k-button class="k-link-input-model-placeholder">
 								{{ currentType.placeholder }}
 							</k-button>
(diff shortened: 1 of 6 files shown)


bastianallgeier:
Instead of adding logic to the component, could we simply use CSS to hide the placeholder?

distantnative:
@bastianallgeier added a version with CSS.

bastianallgeier:
I think the CSS solution is a lot better. What do you think? I was wondering though if this is so global that we could actually put it in the reset.css for all inputs or would that be too aggressive? I'm not 100% sure.

distantnative:
@bastianallgeier that could lead to unwanted consequences for plugins, I fear

bastianallgeier:
You are right!

bastianallgeier (approved):

Issue #6095, opened 2023-12-28
Cannot create file with unknown mime type (pass null to string argument)


rasteiner:
## Description
Creating a file with unknown mime type via `$page->createFile(...)` fails with the error:
```
Kirby\Filesystem\Mime::matches(): Argument #1 ($test) must be of type string, null given, 
called in /var/www/html/kirby/src/Filesystem/File.php on line 284
```

**Expected behavior**  

The file should be created anyway, the mime type could be reported as `application/octet-stream`, for example, or `Mime::matches` should take `?string` as first argument and handle `null` somehow. 

## To reproduce
1. Create a file with unknown mime type (for example an EMF file) and put it into the path `$src`
2. Import it into Kirby: 
    ```php
    $page->createFile(['source' => $src,
                'filename' => 'test.emf',
                'parent' => $page,
                'content' => [
                    'uuid' => 'test'
                ]]);
    ```
3. See error

## Your setup
**Kirby Version**  
4.0.2

**Additional context**  
Related: https://github.com/getkirby/kirby/issues/5881

The problem is here:
https://github.com/getkirby/kirby/blob/4.0.2/src/Filesystem/File.php#L278-L286

```php
$mime = $this->mime();          // returns ?string
Mime::matches($mime, $pattern); // expects string
```

lukasbestle:
I agree it's a bug that we pass a nullable value into a method that cannot accept `null`. I have prepared a fix on the branch https://github.com/getkirby/kirby/tree/fix/6095-file-match-mime.

However it won't result in the behavior you wrote down as expected. The code in question can only be called if the file blueprint contains an `accept` rule/allowlist for the MIME type. If we cannot detect the file's MIME type but only specific MIME types are allowed, we cannot let the file through IMO.

How does your file blueprint look?

rasteiner:
> How does your file blueprint look?

I don't specify a file template in the createFile call, so I don't know where that `mime` rule comes from. I would have guessed the default file blueprint would be used. Does the default file blueprint expect a specific mine type? 
Or does the createFile call search for file templates in the parent page blueprint?

lukasbestle:
Sorry for the delay, will get back to you.

rasteiner:
I've worked around the issue in my project and honestly forgot about it, no hurry for me :)

lukasbestle:
It's indeed coming from the default in this line:

[...]

PR #6208, opened 2024-01-25
`File::match()`: Handle missing MIME types

lukasbestle:
### Fixes

- Proper error message when a MIME type of a file is being validated but could not be determined from the file #6095

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -277,6 +277,15 @@ class File
 		if (is_array($rules['mime'] ?? null) === true) {
 			$mime = $this->mime();
 
+			// the MIME type could not be determined, but matching
+			// to it was requested explicitly
+			if ($mime === null) {
+				throw new Exception([
+					'key'  => 'file.mime.missing',
+					'data' => ['filename' => $this->filename()]
+				]);
+			}
+
 			// determine if any pattern matches the MIME type;
 			// once any pattern matches, `$carry` is `true` and the rest is skipped
 			$matches = array_reduce(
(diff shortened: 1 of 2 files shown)


src/Filesystem/File.php
```diff
@@ -277,6 +277,15 @@ public function match(array $rules): bool
 		if (is_array($rules['mime'] ?? null) === true) {
 			$mime = $this->mime();
 
+			// the MIME type could not be determined, but matching
+			// to it was requested explicitly
+			if ($mime === null) {
+				throw new Exception([
```
bastianallgeier, line 283: Could this be a NotFoundException for a bit more context?

lukasbestle: I copied over the pattern of the existing exceptions. Just doing it here would be inconsistent IMO and changing it everywhere could be a breaking change.

distantnative: If we want to use the more specific exception, would it be an option to use it already here and add todos for v5 to change the others? Then we reconcile consistency vs. breaking change

lukasbestle: Maybe it would make sense to introduce more specific exceptions that better fit the type of error instead of the current broad "catch-all" ones. For this error I can't find one that fully fits. `NotFoundException` would mean that the file itself is not found (which is one possible cause, but could also be that the file exists but has no detectable MIME type). `LogicException` would fit but is *very* broad.

bastianallgeier: You are right. We could totally use better exceptions for various cases. But I guess this is way too much for this PR. So I would probably just stick with the Exception so far.

bastianallgeier (approved):

PR #6212, opened 2024-01-26
feat: `language:variables` root

### LINKED ISSUE #6176: External language variables setup no longer works since K4's language views
## Description

In the past we could require language variables from an externalized file like this: 

```php
// in /site/languages/en.php
'translations' => Yaml::decode(F::read(kirby()->root('storage').'/language-vars/en.yml'))
```

This worked really well with the philosophy of the [custom folder setup](https://getkirby.com/docs/guide/configuration#custom-folder-setup). Also, it had the added benefit you could setup your git repo in a good way to separate code/content (I'ld consider language variables now "content" in the new K4 setup with the new language view as they are now de facto under control by the panel editors).

As per [Sonja's suggesion](https://forum.getkirby.com/t/kirby-4-and-languages/30516/5), I've tried to set this up with a regular php array instead of yaml, like this:

```php
// in /site/languages/en.php
'translations' => require(kirby()->root('storage') . '/language_variables/en.php')
```

This works great when *reading* the variables. As soon as an editor creates a new variable, or changes the value of an existing one, the whole array is saved into `/site/languages/en.php` instead of in `/storage/language_variables/en.php`. 

With Kirby 3 I've never experienced this issue, but there we didn't have the possibility where panel editors could manage those language variables. Would you consider this a regression?

Also, a bit more background about how I usually setup my multilingual Kirby websites: languages can not be added/deleted/edited by editors. It's a dev only thing in my setups, but if we had more fine-grained permissions, this could also be done by a panel admin role. The reason why it's admin/dev only is it could lead to unexpected breakage of things in the frontend (or e.g. when pulling it other content over API's, ...), and I always like to have a developer/admin assess the requested language change. It's also something I'ld add to my git repo and consider "non user manipulated content" (this is also why I externalize language variables to another location).
Since K4, we can let panel editors manage language variables via the new language views. I really, really like that feature. But as you can see, it's not blending in well with the way I used to setup sites. I kindof miss more granular permissions: only allow admins to create/delete/update languages and language settings, but do allow editors to manage the language variables (ideally only "edit" and not "create"/"delete")

**Expected behavior**  
No changes in `/site/languages/en.php`, but in `/storage/language_variables/en.php`.

## To reproduce
Setup a language in `/site/languages/en.php` and a php file `/storage/language_variables/en.php` that returns an array with some translations in `'translations' => require(kirby()->root('storage') . '/language_variables/en.php')`.

In `/site/languages/en.php`, use following to require the external file:
```php
'translations' => require(kirby()->root('storage') . '/language_variables/en.php')
```

Go to the panel and add/edit a variable of the English language.

Check the `/site/languages/en.php` file. The require is now gone and the translations are now moved into this file.

## Your setup
**Kirby Version**  
4.0.3

afbora:
## This PR …
This was a more complex problem than I expected (always 🙈). Because the system should be able to read, update and delete language variables both from the language file and from a custom root. It involves a breakage change, but I wanted to remove the code complexity by creating a separate `LanguageVariables` object.

### Todos/To discuss
- [ ] Inconsistencies: Sometimes we call them variables, sometimes translations (e.g. in the `k-language-view` and the API endpoints). Do we fully want to switch to `variables`? Then we should make the changes everywhere (with legacy fallbacks for now).
- [ ] Do we need to pass `$props['translation']` to the `LanguageVariables` object? Or would it be better to keep them separate, so that `LanguageVariables` only takes care about the file and its value. And `Language` has the job to merge both arrays (`$props['translation']`) and `$this->variables->toArray()`) together. As well as filtering all keys from  `$props['translation']` to only write everything remaining to the variables file.
- [ ] Update `@since` tags
- [ ] Add unit tests for new/altered code
- [ ] Passing unit tests

### Fixes
- External language variables setup no longer works since K4's language views #6176

### Features
- New `language:variables` root

````php
$kirby = new Kirby([
    'roots' => [
        'language:variables' => __DIR__ . '/site/variables'
    ]
]);
````

### Breaking changes
- `Language::translations()` deprecated, use `Language::variables()` instead
- `$language->translations()` returns `LanguageVariables` object instead `array`. Use `$language->translations()->toArray()` or `$language->variables()->toArray()`.

## Ready?

### For review team

### DIFF
--- a/src/Cms/Core.php
+++ b/src/Cms/Core.php
@@ -357,6 +357,7 @@ class Core
 			'commands'          => fn (array $roots) => $roots['site'] . '/commands',
 			'config'            => fn (array $roots) => $roots['site'] . '/config',
 			'controllers'       => fn (array $roots) => $roots['site'] . '/controllers',
+			'language:variables' => null,
 			'languages'         => fn (array $roots) => $roots['site'] . '/languages',
 			'licenses'          => fn (array $roots) => $roots['site'] . '/licenses',
 			'license'           => fn (array $roots) => $roots['config'] . '/.license',
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -10,6 +10,7 @@ use Kirby\Filesystem\F;
 use Kirby\Toolkit\Locale;
 use Kirby\Toolkit\Str;
 use Stringable;
+use Throwable;
 
 /**
  * The `$language` object represents
@@ -51,8 +52,8 @@ class Language implements Stringable
 	protected bool $single;
 	protected array $slugs;
 	protected array $smartypants;
-	protected array $translations;
 	protected string|null $url;
+	protected LanguageVariables $variables;
 
 	/**
 	 * Creates a new language object
@@ -73,7 +74,6 @@ class Language implements Stringable
 		$this->single       = $props['single'] ?? false;
 		$this->slugs        = $props['slugs'] ?? [];
 		$this->smartypants  = $props['smartypants'] ?? [];
-		$this->translations = $props['translations'] ?? [];
 		$this->url          = $props['url'] ?? null;
 
 		if ($locale = $props['locale'] ?? null) {
@@ -81,6 +81,8 @@ class Language implements Stringable
 		} else {
 			$this->locale = [LC_ALL => $this->code];
 		}
+
+		$this->variables = new LanguageVariables($this, $props['variables'] ?? $props['translations'] ?? []);
 	}
 
 	/**
@@ -135,7 +137,7 @@ class Language implements Stringable
 			'name'         => $this->name,
 			'slugs'        => $this->slugs,
 			'smartypants'  => $this->smartypants,
-			'translations' => $this->translations,
+			'variables' => $this->variables->toArray(),
 			'url'          => $this->url,
 		], $props));
 	}
@@ -225,6 +227,7 @@ class Language implements Stringable
 		LanguageRules::delete($this);
 
 		// apply before hook
+		/** @var \Kirby\Cms\Language $language */
 		$language = $kirby->apply(
 			'language.delete:before',
 			['language' => $this]
@@ -236,6 +239,9 @@ class Language implements Stringable
 		if (F::remove($language->root()) !== true) {
 			throw new Exception(message: 'The language could not be deleted');
 		}
+
+		// delete custom translations file if defined
+		$language->variables()->delete();
 
 		// if needed, convert content storage to single lang
 		foreach ($kirby->models() as $model) {
@@ -483,7 +489,11 @@ class Language implements Stringable
 	 */
 	public function save(): static
 	{
-		$existingData = Data::read($this->root(), fail: false);
+		try {
+			$existingData = Data::read($this->root(), fail: false);
+		} catch (Throwable) {
+			$existingData = [];
+		}
 
 		$data = [
 			...$existingData,
@@ -492,11 +502,18 @@ class Language implements Stringable
 			'direction'    => $this->direction(),
 			'locale'       => Locale::export($this->locale()),
 			'name'         => $this->name(),
-			'translations' => $this->translations(),
+			'variables' => $this->variables()->toArray(),
 			'url'          => $this->url,
 		];
 
 		ksort($data);
+
+		// save translations to the custom root and remove translations
+		// to prevent duplication write into the language file
+		if ($this->variables()->root() !== null) {
+			$this->variables()->save($data['translations'] ?? []);
+			$data['translations'] = [];
+		}
 
 		Data::write($this->root(), $data);
 
@@ -558,11 +575,14 @@ class Language implements Stringable
 	}
 
 	/**
-	 * Returns the translation strings for this language
+	 * Alias for Language::variables()
+	 *
+	 * @deprecated 5.0.0 Use `::variables()` instead
+	 * @todo 7.0.0 Remove the method
 	 */
-	public function translations(): array
+	public function translations(): LanguageVariables
 	{
-		return $this->translations;
+		return $this->variables();
 	}
 
 	/**
@@ -589,6 +609,7 @@ class Language implements Stringable
 		$props['slug'] = Str::slug($props['slug'] ?? null);
 
 		// trigger before hook
+		/** @var \Kirby\Cms\Language $language */
 		$language = $kirby->apply(
 			'language.update:before',
 			[
@@ -601,7 +622,7 @@ class Language implements Stringable
 		$language = $language->clone($props);
 
 		if (isset($props['translations']) === true) {
-			$language->translations = $props['translations'];
+			$language->variables = $language->variables->update($props['translations'] ?? null);
 		}
 
 		// validate the language rules after before hook was applied
@@ -654,5 +675,13 @@ class Language implements Stringable
 			language: $this,
 			key:      $key
 		);
+	}
+
+	/**
+	 * Returns the language variables object for this language
+	 */
+	public function variables(): LanguageVariables
+	{
+		return $this->variables;
 	}
 }
--- a/src/Cms/AppTranslations.php
+++ b/src/Cms/AppTranslations.php
@@ -31,7 +31,10 @@ trait AppTranslations
 				$this->multilang() === true &&
 				$language = $this->languages()->find($locale)
 			) {
-				$data = [...$data, ...$language->translations()];
+				$data = [
+					...$data,
+					...$language->variables()->toArray()
+				];
 			}
 
 
@@ -139,7 +142,10 @@ trait AppTranslations
 
 		// inject current language translations
 		if ($language = $this->language($locale)) {
-			$inject = [...$inject, ...$language->translations()];
+			$inject = [
+				...$inject,
+				...$language->variables()->toArray()
+			];
 		}
 
 		// load from disk instead
@@ -164,14 +170,14 @@ trait AppTranslations
 		// injects languages translations
 		if ($languages = $this->languages()) {
 			foreach ($languages as $language) {
-				$languageCode         = $language->code();
-				$languageTranslations = $language->translations();
+				$languageCode      = $language->code();
+				$languageVariables = $language->variables()->toArray();
 
 				// merges language translations with extensions translations
-				if (empty($languageTranslations) === false) {
+				if (empty($languageVariables) === false) {
 					$translations[$languageCode] = [
 						...$translations[$languageCode] ?? [],
-						...$languageTranslations
+						...$languageVariables
 					];
 				}
 			}
--- a/src/Cms/LanguageVariable.php
+++ b/src/Cms/LanguageVariable.php
@@ -48,12 +48,12 @@ class LanguageVariable
 			);
 		}
 
-		$kirby        = App::instance();
-		$language     = $kirby->defaultLanguage();
-		$translations = $language->translations();
+		$kirby     = App::instance();
+		$language  = $kirby->defaultLanguage();
+		$variables = $language->variables()->toArray();
 
 		if ($kirby->translation()->get($key) !== null) {
-			if (isset($translations[$key]) === true) {
+			if (isset($variables[$key]) === true) {
 				throw new DuplicateException(
 					message: 'The variable already exists'
 				);
@@ -64,9 +64,9 @@ class LanguageVariable
 			);
 		}
 
-		$translations[$key] = $value ?? '';
+		$variables[$key] = $value ?? '';
 
-		$language = $language->update(['translations' => $translations]);
+		$language = $language->update(['variables' => $variables]);
 
 		return $language->variable($key);
 	}
@@ -80,11 +80,9 @@ class LanguageVariable
 	{
 		// go through all languages and remove the variable
 		foreach ($this->kirby->languages() as $language) {
-			$variables = $language->translations();
-
-			unset($variables[$this->key]);
-
-			$language->update(['translations' => $variables]);
+			$variables = $language->variables();
+			$variables->remove($this->key);
+			$language->update(['variables' => $variables->toArray()]);
 		}
 
 		return true;
@@ -96,7 +94,7 @@ class LanguageVariable
 	public function exists(): bool
 	{
 		$language = $this->kirby->defaultLanguage();
-		return isset($language->translations()[$this->key]) === true;
+		return $language->variables()->get($this->key) !== null;
 	}
 
 	/**
@@ -130,19 +128,19 @@ class LanguageVariable
 	 */
 	public function update(string|array|null $value = null): static
 	{
-		$translations             = $this->language->translations();
-		$translations[$this->key] = $value ?? '';
+		$variables = $this->language->variables();
+		$variables->set($this->key, $value);
 
-		$language = $this->language->update(['translations' => $translations]);
+		$language = $this->language->update(['variables' => $variables->toArray()]);
 
 		return $language->variable($this->key);
 	}
 
 	/**
-	 * Returns the value if the variable has been translated.
+	 * Returns the value if the variable has been translated
 	 */
 	public function value(): string|array|null
 	{
-		return $this->language->translations()[$this->key] ?? null;
+		return $this->language->variables()->get($this->key);
 	}
 }
--- a/src/Cms/LanguageVariables.php
+++ b/src/Cms/LanguageVariables.php
@@ -1,0 +1,133 @@
+<?php
+
+namespace Kirby\Cms;
+
+use Kirby\Data\Data;
+use Kirby\Exception\Exception;
+use Kirby\Filesystem\F;
+use Throwable;
+
+/**
+ * Manages the variables string for a language,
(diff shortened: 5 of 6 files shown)

### CODE REVIEW

File: src/Cms/Language.php [hunk 1]
```diff
@@ -474,9 +483,9 @@ public function toArray(): array
 	}
 
 	/**
-	 * Returns the translation strings for this language
+	 * Returns the language translations object for this language
 	 */
-	public function translations(): array
```
lukasbestle, line 569: Maybe we could keep returning an array here for backwards-compatibility and add an internal `translationsObject` method. This would allow shipping this feature in 4.x already as it seems to be a kind-of regression in 4.0. We could then change it to return the object in v5 with the breaking change and remove the internal method again.

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,141 @@
...
+class LanguageTranslations
+{
+	protected App   $kirby;
+	protected array $data;
+
+	public function __construct(
+		protected Language $language,
+		self|array $translations = []
+	) {
+		$this->kirby = App::instance();
```
lukasbestle: Do we need to have the app object in an instance prop as it's the global one anyway?

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,141 @@
...
+	/**
+	 * Saves the language translations in the custom root
+	 * @internal
+	 *
+	 * @return $this
+	 */
+	public function save(array $translations = []): static
+	{
+		if ($root = $this->root()) {
+			$this->data = $translations;
```
lukasbestle: I think this can call `$this->setTranslations($translations)` and be run even if there is no root for behavior consistency.

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	/**
+	 * Returns custom translations root path if defined
+	 */
+	public function root(): string|null
+	{
+		$kirby = App::instance();
+		$root  = $kirby->root('translations');
+		$file  = $root . '/' . $this->language->code() . '.php';
```
lukasbestle: Won't this explode when `$root` is `null` because of string concatenation with `null`?

lukasbestle (commented):
Great solution to move the logic to a separate class. 👍
A few things I noticed while skimming through the changes:

src/Cms/Language.php
afbora: I thought about the new `->translationsObject()` method but I thought it would cause more confusion. I will try to implement it in a separate PR.

afbora: Here K4 version: https://github.com/getkirby/kirby/pull/6216

tobimori:
Excited for this, thanks! Will this include the possibility to add translations (e.g. for the panel) even if the language does not have a content language assigned to it? Some of my pages are German only, but should still have English translations for the panel.

afbora:
@tobimori In your case; multilingual is not enabled for the frontend, but you want to add translations (for example for blueprint labels) for users in different languages using the panel. Do I understand correctly?

tobimori:
> @tobimori In your case; multilingual is not enabled for the frontend, but you want to add translations (for example for blueprint labels) for users in different languages using the panel. Do I understand correctly?

Multilingual is enabled, but the languages in the panel might differ from the languages in the frontend.
Let's say I have `de` and `en` in the frontend, but I want to add a french translation for my field labels etc. as well.

Currently, I have a translations directory and load yaml files from a plugin. (https://github.com/tobimori/kirby-baukasten/blob/main/site/plugins/project-extended/index.php#L20)

afbora:
If there is no FR language (so `fr.php` language file), unfortunately it will not load custom FR translations. It will only load the translation files for existing languages from a custom root.

tobimori:
😦

src/Cms/Language.php
```diff
@@ -461,6 +462,7 @@ public function save(): static
 	{
 		try {
 			$existingData = Data::read($this->root());
+			$data['translations'] = $this->translations()->load($existingData['translations'] ?? []);
```
distantnative: Where is `$data` coming from? Does this need to be `$existingData['translations']`?

src/Cms/Language.php
```diff
@@ -472,12 +474,19 @@ public function save(): static
...
 		];
 
 		ksort($data);
 
+		// save translations to the custom root and remove translations
+		// to prevent duplication write into the language file
+		if ($this->translations()->root() !== null) {
+			$this->translations()->save($data['translations'] ?? []);
+			$data['translations'] = [];
+		}
```
distantnative, line 516: How about instead of unsetting it here, we do an if-else clause: The if part writes it to the dedicated root, the else part adds it to the `$data` array?

src/Cms/Language.php
```diff
@@ -571,7 +580,7 @@ public function update(array|null $props = null): static
 		$updated = $this->clone($props);
 
 		if (isset($props['translations']) === true) {
-			$updated->translations = $props['translations'];
+			$updated->translations = new LanguageTranslations($updated, $props['translations']);
```
distantnative: Does this work? I thought `translations` property is protected.

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,140 @@
...
+
+namespace Kirby\Cms;
+
+use Exception;
+use Kirby\Data\Data;
+use Kirby\Filesystem\F;
+
+/**
+ * With helper methods provides get language translations or
+ * loads from custom `translations` root
```
distantnative:
```suggestion
 * Manages the translations string for a language,
 * either from the language file or `translations` root 
```

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,140 @@
...
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ */
+class LanguageTranslations
+{
+	protected array $data;
+
+	public function __construct(
+		protected Language $language,
+		self|array $translations = []
```
distantnative:
```suggestion
		LanguageTranslations|array $translations = []
```

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,140 @@
...
+	) {
+		$this->setTranslations($translations);
```
distantnative: I'd be happy if we can move away from these kind of set methods. If there is a nicer way.

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,140 @@
...
+		return $default;
+	}
+
+	/**
+	 * Saves the language translations in the custom root
+	 * @internal
+	 *
+	 * @return $this
+	 */
+	public function save(array $translations = []): static
```
distantnative:
```suggestion
	public function update(array $translations = []): static
```

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,140 @@
...
+		$file  = ($root ?? '') . '/' . $this->language->code() . '.php';
+
+		if (
+			$root !== null &&
+			F::exists($file) === true
+		) {
+			return $file;
+		}
+
+		return null;
```
distantnative:
```suggestion
		if ($root  = $kirby->root('translations')) {
				return  $root . '/' . $this->language->code() . '.php';
		}
		
		return null;
```

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,140 @@
...
+	 *
+	 * @return $this
+	 */
+	public function save(array $translations = []): static
+	{
+		$this->setTranslations($translations);
+
+		if ($root = $this->root()) {
+			Data::write($root, $translations);
+		}
```
distantnative:
```suggestion
		$this->data = [...$this->data, $translations];

		if ($root = $this->root()) {
			Data::write($root, $this->data);
		}
```

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,140 @@
...
+	{
+		if (empty($translations) === false) {
+			if ($translations instanceof self) {
+				$this->data = $translations->toArray();
+			} else {
+				$this->data = $translations;
+			}
+		} else {
+			$this->data = static::load();
+		}
```
distantnative: Why/where do we need to support that `$translations` would itself be a `LanguageTranslations` object? Would be a lot easier if we could rely on array only:

```suggestion
		$this->data = [
				...$this->load(),
				...$translations
		];
```

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,140 @@
...
+	{
+		return $this->data[$key] ?? $default;
+	}
+
+	/**
+	 * Loads the language translations based on custom roots for provided language code
+	 */
+	public function load(array $default = []): array
+	{
+		if ($file = static::root()) {
```
distantnative:
```suggestion
		if ($file = $this->root()) {
```

distantnative (commented):
Some first comments. Let me know if you think I misunderstood parts of the code :)

src/Cms/LanguageTranslations.php
afbora: I think it should be save to be consistence with other models. This method writes data.

src/Cms/LanguageTranslations.php
afbora: `::setTranslations()` is custom method that checks the argument whether is array or self object.

src/Cms/Language.php
afbora: If custom root is already set (`root !== null`), it empties the data.translations array and writes to the custom root. If custom root is not set, it saves the data array normally as before.

src/Cms/Language.php
afbora: It was already protected: `protected array $translations`?

afbora:
@distantnative I did some changes based on your reviews and done some improvements. Would be nice if you take a look when you're available before writing unit tests and last touch 🙌

src/Cms/Core.php
```diff
@@ -331,33 +331,34 @@ public function load(): Loader
...
+			'languages'    => fn (array $roots) => $roots['site'] . '/languages',
+			'license'      => fn (array $roots) => $roots['config'] . '/.license',
+			'logs'         => fn (array $roots) => $roots['site'] . '/logs',
+			'models'       => fn (array $roots) => $roots['site'] . '/models',
+			'plugins'      => fn (array $roots) => $roots['site'] . '/plugins',
+			'roles'        => fn (array $roots) => $roots['blueprints'] . '/users',
+			'sessions'     => fn (array $roots) => $roots['site'] . '/sessions',
+			'snippets'     => fn (array $roots) => $roots['site'] . '/snippets',
+			'templates'    => fn (array $roots) => $roots['site'] . '/templates',
+			'translations' => null,
```
distantnative: `Question:` I don't have a good idea yet, but wondering if it's too confusing to have `i18n:translations` and `translations`? Maybe `languages:translations`?

src/Cms/Language.php
distantnative: What I meant:

```php
		$data = [
			...$existingData,
			'code'         => $this->code(),
			'default'      => $this->isDefault(),
			'direction'    => $this->direction(),
			'locale'       => Locale::export($this->locale()),
			'name'         => $this->name(),
			'url'          => $this->url,
		];

		ksort($data);

		// save translations to the custom root and remove translations
		// to prevent duplication write into the language file
		if ($this->translations()->root() !== null) {
			$this->translations()->save();
		} else {
			$data['translations'] = $this->translations()->toArray();
		}
```

(not passing anything to `$this->translations()->save()` here based on the other suggestion I make below)

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	/**
+	 * Saves the language translations in the custom root
+	 * @return $this
+	 * @internal
+	 */
+	public function save(array $translations = []): static
+	{
+		$this->data = $translations;
```
distantnative:
```suggestion
	public function save(array|null $translations = null): static
	{
		if ($translations !== null) {
			$this->update($translations);
		}	
```

src/Cms/Language.php
```diff
@@ -548,9 +559,9 @@ public function toArray(): array
 	}
 
 	/**
-	 * Returns the translation strings for this language
+	 * Returns the language translations object for this language
 	 */
-	public function translations(): array
+	public function translations(): LanguageTranslations
```
distantnative, line 564: `Suggestion:` What if we keep returning an array here? Then we would not need to change the `AppTranslations.php` and externally it would work as before.

src/Cms/Language.php
afbora: We need to empty to `$data['translations']` index if custom root defined. Because after few lines later will run `Data::write()` and same translation strings to be written to language translations index. `$data['translations'] = []` prevents it.

src/Cms/Core.php
afbora: @bastianallgeier What do you think?

bastianallgeier: What was our last decision for the language variable permissions? I wonder if we maybe should go the same route here with naming as for permissions?

afbora: For now to be `languageVariables.create|update|delete`.

bastianallgeier: We could go for `language:variables` maybe

I know that it totally clashes with the translations naming everywhere. But I think it makes sense to fully commit to one naming convention here and language variables are at least not giving us any more headaches with the entire content translations vs. ui translations monster.

afbora: In the long run, what if we also changed the `translations` index in the language file to `variables`? For ex: we can deprecate `translations` index v5-v6 and we can remove v7. How does that sound?

bastianallgeier: Yes, we should definitely do that. Maybe it would be good to already use LanguageVariables for the class name in this PR and maybe even refactor the property name?

afbora: Done for this PR. I'll deprecate the `translations` index in another PR.

distantnative: Maybe even `i18n:variables` then?

src/Cms/Language.php
distantnative: But if we don't add it to `$data` in the first place, we wouldn't need to unset it, right? Or could `$existingData` also have a `translations` key?

src/Cms/Core.php
afbora: Not sure for this PR, @bastianallgeier ?

src/Cms/Language.php
afbora: Yes, `$existingData` mean `/site/languages/*.php` data and currently languages have `translations` key.

src/Cms/Core.php
bastianallgeier: I would say that it belongs to the language and thus the language namespace feels better to me.

distantnative: My point there is that so far we don't have a language namespace anyways, we are creating a new one. And what if language is the i18n namespace?

src/Cms/Language.php
```diff
@@ -68,14 +68,15 @@ public function __construct(array $props)
...
-		$this->translations = $props['translations'] ?? [];
 		$this->url          = $props['url'] ?? null;
 
 		if ($locale = $props['locale'] ?? null) {
 			$this->locale = Locale::normalize($locale);
 		} else {
 			$this->locale = [LC_ALL => $this->code];
 		}
+
+		$this->variables = new LanguageVariables($this, $props['translations'] ?? []);
```
bastianallgeier: I think we could already support a new variables prop here as well and fallback to translations to keep them available. 

```suggestion
		$this->variables = new LanguageVariables($this, $props['variables'] ?? $props['translations'] ?? []);
```

src/Cms/Language.php
```diff
@@ -130,7 +131,7 @@ public function clone(array $props = []): static
 			'name'         => $this->name,
 			'slugs'        => $this->slugs,
 			'smartypants'  => $this->smartypants,
-			'translations' => $this->translations,
+			'translations' => $this->variables->toArray(),
```
bastianallgeier:
```suggestion
			'variables' => $this->variables->toArray(),
```

src/Cms/Language.php
```diff
@@ -482,12 +487,19 @@ public function save(): static
 			'direction'    => $this->direction(),
 			'locale'       => Locale::export($this->locale()),
 			'name'         => $this->name(),
-			'translations' => $this->translations(),
+			'translations' => $this->variables()->toArray(),
```
bastianallgeier:
```suggestion
			'variables' => $this->variables()->toArray(),
```

src/Cms/Core.php
bastianallgeier: This is still about the root and I think we are mixing them up here. The i18n roots are all in our kirby folder to store interface translations and slug rules. The language:variables root would be a custom project root inside of site.

distantnative:
Rebased this but moved it to v6 due to the breaking change.

PR #6213, opened 2024-01-26
Fix namespaces for Options, OptionsApi and Query alias

bezin:
While testing https://github.com/lukaskleinschmidt/kirby-types I ran into issues with aliases. However the error deviated from the Kirby core aliases. I debugged this and found wrong aliases in `config/aliases.php`. 

The namespace is `Kirby\Option\...` rather than `Kirby\Options\...`.


distantnative:
Urgs what an embarrassing mistake 🙈

distantnative (approved):

PR #6216, opened 2024-01-27
[v4] New `translations` root

afbora:
## This PR …

This is K4 version of https://github.com/getkirby/kirby/pull/6212. Doesn't include a breaking change.

### Fixes
- External language variables setup no longer works since K4's language views #6176

### Features
- New `translations` root

````php
$kirby = new Kirby([
    'roots' => [
        'translations' => __DIR__ . '/site/translations',
    ],
]);
````

### Breaking changes
None

## Ready?

### For review team

### CODE REVIEW

File: src/Cms/Language.php [hunk 1]
```diff
@@ -409,6 +410,13 @@ public function save(): static
 	{
 		try {
 			$existingData = Data::read($this->root());
+
+			// inject translations from custom root
+			// returns existing translations
+			// if custom root is not defined as fallback
+			$data['translations'] = $this
```
distantnative: I think this will fail `$data` array doesn't exist yet.

src/Cms/Language.php
```diff
@@ -427,11 +435,31 @@ public function save(): static
...
 
 		return $this;
 	}
 
+	/**
+	 * Sets the translations data
+	 *
+	 * @return $this
+	 */
+	protected function setTranslations(array $translations = []): static
```
distantnative, line 454: Since everything else happens directly in the constructor, I wonder if we'd better off also putting this code directly in the constructor instead of adding a `setTranslations` method.

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,140 @@
...
+		if (empty($translations) === false) {
+			if ($translations instanceof self) {
+				$this->data = $translations->toArray();
+			} else {
+				$this->data = $translations;
+			}
+		} else {
+			$this->data = static::load();
+		}
+
```
distantnative, line 125: I think we could condense these nested if/else conditions into

```php
$this->data = match (true) {
    empty($translations) === true => static::load(),
    $translations instanceof self => $translations->toArray(),
    default                       => $translations
};
```

distantnative: I'd also consider putting it directly into the constructor.

afbora: Also `save` method calls the `setTranslations()` method.

src/Cms/Language.php
```diff
@@ -409,6 +409,13 @@ public function save(): static
 	{
 		try {
 			$existingData = Data::read($this->root());
+
+			// inject translations from custom root
+			// returns existing translations
+			// if custom root is not defined as fallback
+			$existingData['translations'] = $this
+				->translationsObject()
+				->load($existingData['translations'] ?? []);
```
afbora, line 418: Btw should it be able to pull language variables and merge from both main root (/site/languages/en.php) and custom root (/site/translations/en.php)? 

Or should it stay as it is now, if custom root is defined, it pulls from there, otherwise it pulls from the main language file.

bastianallgeier: Sorry, I'm late here. But I would say if the translations root is defined, the translations array in the language file should be ignored.

afbora: So the PR behavior is correct.

src/Cms/Language.php
```diff
@@ -427,11 +434,31 @@ public function save(): static
 
 		ksort($data);
 
+		// save translations to the custom root and remove translations
+		// to prevent duplication write into the language file
+		if ($this->translationsObject()->root() !== null) {
+			$this->translationsObject()
+				->save($data['translations'] ?? []);
+			$data['translations'] = [];
```
bastianallgeier, line 441: Wouldn't it make sense to even unset it here?

src/Cms/LanguageTranslations.php
```diff
@@ -0,0 +1,136 @@
...
+	{
+		return $this->data[$key] ?? $default;
+	}
+
+	/**
+	 * Loads the language translations based on custom roots for provided language code
+	 */
+	public function load(array $default = []): array
+	{
+		if ($file = static::root()) {
```
bastianallgeier: This does not seem to be correct. The root method isn't static.

src/Cms/Language.php
afbora: Seems setting empty array is correct.

With set empty array:
````php
<?php

return [
    'code' => 'en',
    'default' => true,
    'direction' => 'ltr',
    'locale' => [
        'LC_ALL' => 'en_US'
    ],
    'name' => 'EN',
    'translations' => [

    ],
    'url' => NULL
];
````

With unseting:

````php
<?php

return [
    'code' => 'en',
    'default' => true,
    'direction' => 'ltr',
    'locale' => [
        'LC_ALL' => 'en_US'
    ],
    'name' => 'EN',
    'url' => NULL
];
````

afbora:
I didn't understand the unit test coverage. 

For ex: I added `LanguageTranslations::get` tests with `@covers` notation but shows still uncovered.

Changes:
https://github.com/getkirby/kirby/pull/6216/files#diff-4e3a14d5dc988b6b8165c6ab49bbe0a32cf21c4c141054da017fe0730e83bf5aR54-R82

Codecov:
https://app.codecov.io/gh/getkirby/kirby/pull/6216/blob/src/Cms/LanguageTranslations.php#L34

distantnative:
@lukasbestle @afbora To be honest, I find the v4 version very confusing with the duality of the `translation` prop and `translationObject` etc. Since this is an area already hard to overlook and understand (even just with all the language, translation etc. names that are so similar), I would vote for letting go of a v4 version but really just focus on finishing the v5 version (which looks so much nicer already due to the cleaner syntax).

PR #6217, opened 2024-01-27
Fix type hint on `Panel::area()`

distantnative:
I say this isn't a breaking change but just bug fix as the code inside the method would crash if a string was passed.

### DIFF
--- a/src/Panel/Panel.php
+++ b/src/Panel/Panel.php
@@ -35,7 +35,7 @@ class Panel
 	/**
 	 * Normalize a panel area
 	 */
-	public static function area(string $id, array|string $area): array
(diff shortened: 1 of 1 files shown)


afbora (approved):
This was correct up to 3.8. In 3.8 we had applied type hints in docblocks and docblock hints were wrong for the method.

https://github.com/getkirby/kirby/commit/283d00c8358d9bee1dda5d081ee74d49d7e483b2#diff-33e7d22c225cf6c17ede4dec69c897ac102c42d013a388bd18b462caaac579d8

bastianallgeier (approved):

PR #6218, opened 2024-01-28
Slight performance improvements for `Dir::inventory()`

### LINKED ISSUE #5126: Extremely slow site due to the use of models
## Description
This is how Kirby determines the model for a given page:

https://github.com/getkirby/kirby/blob/9ecd11cd5b7d4875d4212f767a2b3ad3e52b3528/src/Filesystem/Dir.php#L346-L353

It means that for each child page, it checks **every possible model** for its existence. If I have 10 models, that means 10 `file_exists()` checks. Of course, if your model starts with "a", it will alphabetically be first and you might get away with just one iteration, then the loop breaks. But that's not guaranteed.

In our site, we have about 900 pages and 7 models. That's 6300 filesystem calls to index all pages. Adding just one more model and we're up to over 7000…

Our panel currently takes ~2.5 seconds to load. If I delete our models, that time drops down to **~0.4 seconds**. In other words, **the use of models makes our panel 6 times slower**. Outside of the panel, our home page is also **about 3 times slower**.

## Possible Solution
You wouldn't have to make all those filesystem calls if the filename is predetermined. All TXT files of pages could just be `page.txt` (the same way `site.txt` works) and you could store the model type inside the page, like so:

```
Model: article

----

Title: Hello World
```

distantnative:
## This PR …

Tried to fix #5126 - while the issue itself remains without a good solution, some improvements to the methods structure and performance could be made. Added those in this PR.

### Refactored
- Slight performance improvements for `Dir::inventory()`

### Housekeeping
- Upgraded Psalm
- Added performance tests for `Dir::inventory()` with various number of page models

## Ready?

### For review team


src/Filesystem/Dir.php
```diff
@@ -220,143 +220,145 @@ public static function inventory(
...
 		}
 
-		foreach ($content as $contentName) {
-			// could be a meta file. i.e. cover.jpg
-			if (isset($inventory['files'][$contentName]) === true) {
-				continue;
+		// determine the model
+		if (empty(Page::$models) === false) {
+			if ($multilang === true) {
+				$code = App::instance()->defaultLanguage()->code();
```
bastianallgeier, line 319: Wouldn't it help to pass this once in multi lang mode from inventory?

src/Filesystem/Dir.php
```diff
@@ -220,143 +220,145 @@ public static function inventory(
...
+				$contentExtension = $code . '.' . $contentExtension;
 			}
 
-			// it's most likely the template
-			$inventory['template'] = $contentName;
+			// look if a content file can be found
+			// for any of the available models
+			foreach (Page::$models as $modelName => $modelClass) {
+				if (file_exists($root . '/' . $modelName . '.' . $contentExtension)) {
```
bastianallgeier:
```suggestion
				if (file_exists($root . '/' . $modelName . '.' . $contentExtension) === true) {
```

bastianallgeier:
Are there any before/after numbers to compare this or is there a good way for me to compare locally?

src/Filesystem/Dir.php
distantnative: @bastianallgeier Like this https://github.com/getkirby/kirby/pull/6218/commits/8e92a46bffdf03384fa10a045b70d26f3d32a7e0 ?

distantnative:
@bastianallgeier 

To run it yourself, you need to copy the added performance tests from `bench/Filesystem` over to the `develop/minor` state. There run `composer run-script bench:baseline`. Then you can switch back to the branch and run `composer run-script bench`. That way, performance of `develop/minor` is compared against this PR.

However, for me the result seems to be worse now after https://github.com/getkirby/kirby/pull/6218/commits/7d97311131b5209f5feb93211a6768aac9e8414d

bastianallgeier:
That's so weird. Do you have any idea why this could perform worse?

distantnative:
@bastianallgeier I have reverted https://github.com/getkirby/kirby/pull/6218/commits/c903e6b192617645f2ffc9cdc3bf243a392821b0 again.

With it, the performance was worse than `develop-minor`, particularly with no or one model:
Screenshot 2024-02-17 at 23 53 54

After reverting that change, this PR shows slight improvements over `develop-minor`:
Screenshot 2024-02-17 at 23 53 46

My guess would be that for multilang installations, the difference isn't bad. But in the bench tests with single lang, calling the app instance even with `$multilang = false` adds quite some overhead.

src/Filesystem/Dir.php
```diff
@@ -220,143 +220,145 @@ public static function inventory(
...
+				$code = App::instance()->defaultLanguage()->code();
+				$contentExtension = $code . '.' . $contentExtension;
 			}
 
-			// it's most likely the template
-			$inventory['template'] = $contentName;
+			// look if a content file can be found
+			// for any of the available models
+			foreach (Page::$models as $modelName => $modelClass) {
+				if (file_exists($root . '/' . $modelName . '.' . $contentExtension) === true) {
```
lukasbestle: I think we should use `is_file()` instead, otherwise it will also be true for folders with that name.

lukasbestle:
> My guess would be that for multilang installations, the difference isn't bad. But in the bench tests with single lang, calling the app instance even with $multilang = false adds quite some overhead.

Can't we also use `if ($multilang === true)` when fetching the language code? If it's false, we don't even need to load the app object. `$languageCode` can just always be `null` then.

src/Filesystem/Dir.php
```diff
@@ -220,143 +220,145 @@ public static function inventory(
...
+			'root'    => $root,
+			'slug'    => $slug ?? $item,
+		];
 	}
 
 	/**
-	 * Go through all inventory children
-	 * and inject a model for each
+	 * Determines the main template for the inventory
+	 * from all collected content files, ignory file meta files
```
lukasbestle:
```suggestion
	 * from all collected content files, ignore file meta files
```

lukasbestle (commented):
Love the refactoring. Besides performance, the new structure makes much more sense and is easier to read and understand IMO.

distantnative:
> Can't we also use if ($multilang === true) when fetching the language code? If it's false, we don't even need to load the app object. $languageCode can just always be null then.

I think it could still have performance downsides for multilang inventories without children (currently app instance wouldn't be called then at all, after the change it would be called).

lukasbestle (approved):

PR #6219, opened 2024-01-28
Fix `A::prepend()`

distantnative:
## This PR …

With the previous code, non-associative arrays would not be prepended.

### Fixes
- `A::prepend()` now behaves the same as `A::append()` (just opposite side) for non-associative arrays

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -588,7 +588,7 @@ class A
 	 */
 	public static function prepend(array $array, array $prepend): array
 	{
-		return $prepend + $array;
+		return static::merge($prepend, $array, A::MERGE_APPEND);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
LGTM

src/Toolkit/A.php
```diff
@@ -588,7 +588,7 @@ public static function pluck(array $array, string $key): array
 	 */
 	public static function prepend(array $array, array $prepend): array
 	{
-		return $prepend + $array;
+		return static::merge($prepend, $array, A::MERGE_APPEND);
```
distantnative, line 591: Thinking about it, @bastianallgeier - maybe even better to reuse the append one, writing: 

```php
return state::append($prepend, $array);
```

PR #6220, opened 2024-01-28
Some unit test improvements

distantnative:
### Fixes
- Fixed fallback `null` value for `Text\KirbyTag::parent()`

### Housekeeping
- Added more explicit coverage annotations to `Options`, `Query`, `Text` and partially `Toolkit` packages tests, including additional tests for uncovered code


lukasbestle:
About the files in the `config` dir: Unfortunately no, such a feature is missing in PHPUnit, which is the reason why we don't use `@covers` annotations for those methods at the moment: https://github.com/sebastianbergmann/phpunit/issues/3794

About coverage in general going down: I'd say if the new annotation is correct (in that it declares the method(s) that are actually being tested with assertions), then it's fine if coverage goes down as it helps us to see areas of the code that are not properly tested. Even better would of course be to add tests for those now uncovered methods. I think we need to keep in mind that the coverage percentage is also relevant for marketing (e.g. we use it on the "for developers" page). What I *wouldn't* do is to add `@coversNothing` to the config tests before PHPUnit supports declaring coverage for those. Otherwise we will no longer know which parts of the config dir we test at all.

distantnative:
The problem is that often those test for config or other non-class code also touch a dozen other classes etc - so without `coversNothing` we also have spillover into many other classes.

lukasbestle:
That is true. But I feel we need a solution in PHPUnit then to be able to do it properly. Before we have that, I'd rather exclude the whole config dir from the coverage allowlist. But that's also not great as a lot of our logic happens there that we need to test.

So TBH I feel like keeping the status quo is overall the best option.

lukasbestle:
A hack could be to put all these config tests into a separate `@group` so we can test them separately without destroying the coverage stats of the strict tests.

lukasbestle (approved):
Have not reviewed all changes in detail, but skimmed over the general setup changes. Good idea to split up the KirbyTags tests into separate files. Also great that many of the previously uncovered methods now have tests. :+1:

Could you please add/update the first post with the final contents of this PR?

Issue #6230, opened 2024-01-31
"Invalid Date" parsing dates in pages section with layout:table


sciloqi:
## Description
In the pages section, when using `layout:table` with custom columns, the date parsing shows "Invalid Date" for days > 12 and don't format as expected for days <= 12. Using date handler `intl`.

**Expected behavior**  
`Date: 2023-11-24 16:00:00`.toDate('dd.MM.Y') should be 24.11.2023 (is "Invalid Date")
`Date: 2021-10-05 00:00:00`.toDate('dd.MM.Y') should be 05.10.2021 (is "2021-05-10")

**Screenshots**  
![Bildschirmfoto 2024-01-31 um 18 12 57](https://github.com/getkirby/kirby/assets/23317040/44232804-e944-4c5e-8423-4a33c057abf9)

## To reproduce

1. Fresh starter-kit
2. add `intl` date handler to config.php
3. add `layout:table` with `columns: / date: / value: "{{ page.date.toDate('dd.MM.Y') }}` to blueprints/sections/notes.yml
4. See panel "site"

## Your setup
Kirby 4.1.0 (no problems with Kirby 4.0.3)

afbora:
I can reproduce the issue on 4.1.0 either. Works great for 4.0.3

distantnative:
I fear this is originating from https://github.com/getkirby/kirby/pull/6147

Before the PR, this would render as a text preview - now it's trying to render as a date preview and tripping up.

@sciloqi as quick fix adding `type: text` to your column definition should probably solve it for you

afbora:
@distantnative Yes `type: text` works great.

afbora:
I think also following usage give outputs same with `{{ page.date.toDate('dd.MM.Y') }}`

````yaml
columns:
  date:
    display: DD.MM.YYYY
````

janstuemmel:
Bug still exists (kirby@4.2.0) when day ist < 10 in `2024-09-05` for example. 

![Bildschirmfoto_2024-04-16_19-23-27](https://github.com/getkirby/kirby/assets/7893682/f2918a4d-b85a-4b3b-9a7a-74bbb675def8)

For the first cell "Date Start" data is `Datestart: 2024-04-09`, kriby will print `2024-09-04 `. So i think there is any confusion with day and month. This happens only in table columns i think. 

However, it is still fixable with `type: text`.

Please reopen @distantnative

distantnative:
@janstuemmel What's your blueprint?

[...]

PR #6234, opened 2024-02-01
Fix invalid date in table layout

afbora:
## This PR …

I've found a simple solution that shows custom value prop if defined. Otherwise date preview handles rest. I hope it's correct solution.

### Fixes
- "Invalid Date" parsing dates in pages section with layout:table #6230

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/lab/components/fieldpreviews/date/index.vue
+++ b/panel/lab/components/fieldpreviews/date/index.vue
@@ -19,6 +19,14 @@
 			</k-lab-table-cell>
 		</k-lab-example>
 
+		<k-lab-example label="Invalid date">
+			<k-lab-table-cell>
+				<!-- @code -->
+				<k-date-field-preview value="Not a date" />
+				<!-- @code-end -->
+			</k-lab-table-cell>
+		</k-lab-example>
+
 		<k-lab-example label="Time">
 			<k-lab-table-cell>
 				<!-- @code -->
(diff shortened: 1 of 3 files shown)


bastianallgeier:
I think we can keep it even simpler by checking for a valid parsed date and fall back to the raw value. This would still allow date parsing for the custom column value. I also added lab examples. Could you give it another quick look if you agree with my solution?

afbora:
@bastianallgeier It works but this creates inconsistency.

screen-capture-1829-Notes - Mægazine-localhost

bastianallgeier:
What are the values for those rows?

afbora:
For first one: 2023-11-24 16:00:00
The last one: 2020-09-05 00:00:00

You can test on notes pages in starterkit with:

````php
return [
    'date.handler' => 'intl',
];
````

````yaml
sections:
  listed:
    extends: sections/notes
    label: Published
    status: listed
    layout: table
    columns:
      date:
        value: "{{ page.date.toDate('dd.MM.Y') }}"
````

bastianallgeier:
Ah, now I understand. The custom value is leading to dates, which cannot always be parsed. Some of them work. Some of them don't. I would say that we should accept this as a limitation, because you can still work around it by switching the column type to text. 

@distantnative do you have an opinion on this?

distantnative:
I agree, I think after all the `value: "{{ page.date.toDate('dd.MM.Y') }}"` is using it incorrectly.

Then again, that's mainly because I don't think we have column types documented in a good way (which would be then the proper way to get the right preview).

bastianallgeier (approved):

bastianallgeier:
I would merge this fix for the next patch and then we should also take care about the column type docs.

Issue #6244, opened 2024-02-05
API endpoints for sections


distantnative:
As easily definable as we have them for fields already.

PR #6246, opened 2024-02-05
Section api

bastianallgeier:
### Features

- The new section api lets you define API endpoints for your panel sections. Just like for fields, sections can now have their own routes. https://github.com/getkirby/kirby/issues/6244

```php 
// my section
return [
  'api' => function() {
    return [
      [ 
        'pattern' => '/say-hello', 
        'action' => function () {
          return [
            'message' => 'Hello, World!'
          ];
        }
      ]
    ];
  }
];
```

With the route defined on the backend, you can now access it in the panel with the API library. 

```js
const response = await this.$api.get('/pages/blog+article-a/sections/sectionName/say-hello');

alert(response.message);
```

### Enhancements

- Use the new section API in the files section to sort files.

## Ready?

### For review team

### DIFF
--- a/config/api/routes/files.php
+++ b/config/api/routes/files.php
@@ -26,6 +26,15 @@ return [
 		}
 	],
 	[
+		'pattern' => $filePattern . '/fields/(:any)/(:all?)',
+		'method'  => 'ALL',
+		'action'  => function (string $parent, string $filename, string $sectionName, string $path = null) {
+			if ($file = $this->file($parent, $filename)) {
+				return $this->sectionApi($file, $sectionName, $path);
+			}
+		}
+	],
+	[
 		'pattern' => $parentPattern,
 		'method'  => 'GET',
 		'action'  => function (string $path) {
--- a/config/api/routes/files.php
+++ b/config/api/routes/files.php
@@ -8,7 +8,15 @@ $parentPattern = '(account|pages/[^/]+|site|users/[^/]+)/files';
  * Files Routes
  */
 return [
-
+	[
+		'pattern' => $filePattern . '/fields/(:any)/(:all?)',
+		'method'  => 'ALL',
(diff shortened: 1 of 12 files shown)


config/api/routes/files.php
```diff
@@ -25,6 +25,15 @@
 			}
 		}
 	],
+	[
+		'pattern' => $filePattern . '/fields/(:any)/(:all?)',
```
distantnative: Is this right? Not `/sections/`?

src/Cms/Api.php
```diff
@@ -185,6 +185,28 @@ public function searchPages(string|null $parent = null): Pages
...
+	 */
+	public function sectionApi(
+		ModelWithContent $model,
+		string $name,
+		string|null $path = null
+	): mixed {
+		$section = $model->blueprint()?->section($name);
+
+		$sectionApi = $this->clone([
+			'data'   => array_merge($this->data(), ['section' => $section]),
```
distantnative:
```suggestion
			'data'   => [...$this->data(), 'section' => $section],
```

distantnative (dismissed):

distantnative:
I think this is a good case to see that having our API routes in the code coverage included is not very useful IMO

bastianallgeier:
Quite happy with the state of unit tests now for this.

distantnative:
I think we should really think of adding `codeCoverageIgnore` to some parts, in general, where we feel that tests are just busy work but don't add real benefit (e.g. one liner API routes etc.)

Cause it's bad if such a PR drags down our coverage - while I agree that I don't think adding more unit tests really adds any benefit here.

distantnative (dismissed):

distantnative (approved):

johannschopplich:
This is incredible! No more pollution of the `api` namespace for my plugins. Nice work. 👏

PR #6262, opened 2024-02-12
Fix to handle invalid PHP version strings

SeriousKen:
## This PR …

Some versions of PHP have extra information in the version string that does not conform to semantic versioning and this causes a problem matching messages in UpdateStatus.

### Fixes
This fix strips off any extra info from the version string with a regex. The regex is taken from Composer where it is used to handle the same situation.

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/System/UpdateStatus.php
+++ b/src/Cms/System/UpdateStatus.php
@@ -158,7 +158,9 @@ class UpdateStatus
 		// collect all matching custom messages
 		$filters = [
 			'kirby' => $this->app->version(),
-			'php'   => phpversion()
+			// Some PHP version strings contain extra info that makes them
+			// invalid so we need to strip it off.
+			'php'   => preg_replace('#^([^~+-]+).*$#', '$1', phpversion())
 		];
 
 		if ($type === 'plugin') {
--- a/src/Cms/System/UpdateStatus.php
+++ b/src/Cms/System/UpdateStatus.php
@@ -158,9 +158,9 @@ class UpdateStatus
 		// collect all matching custom messages
 		$filters = [
 			'kirby' => $this->app->version(),
-			// Some PHP version strings contain extra info that makes them
-			// invalid so we need to strip it off.
-			'php'   => preg_replace('#^([^~+-]+).*$#', '$1', phpversion())
+			// some PHP version strings contain extra info that makes them
+			// invalid so we need to strip it off
+			'php'   => preg_replace('/^([^~+-]+).*$/', '$1', phpversion())
 		];
 
 		if ($type === 'plugin') {


afbora:
I don't know why, but I don't feel comfortable with this solution. What about using predefined constants?

````php
// PHP 8.3.2-1+ubuntu20.04.1+deb.sury.org+1
PHP_MAJOR_VERSION   // 8
PHP_MINOR_VERSION   // 3
PHP_RELEASE_VERSION // 2
````

**Comment by SeriousKen** [comment]
@afbora I thought about using the constants and just joining them at first too, but then there are cases where you may have 8.4.0RC1 for example which this solution keeps as a valid semantic version. This is the way composer itself deals with this issue - I just lifted the regex from the source.

**Comment by afbora** [comment]
Oh, I understand. So what about using `Composer\Semver\VersionParser` to normalize? I know that Semver also does it manually like you did, but I think this ensures stability for future versions.

**Comment by SeriousKen** [comment]
@afbora `Composer\Semver\VersionParser::normalize` is used internally in `Composer\Semver\Semver::satisfies` which is called in `Kirby\Cms\System\UpdateStatus::checkConstraint` and this is what is throwing the error because the PHP string is not a valid semantic version. The explanation on Composer's end is that the semantic version tools provided only work with valid semantic version strings - which makes sense. We have to have a valid semantic version by removing anything that is invalid in the PHP version string before calling any of Composer's methods.

There's a discussion about this issue here -> https://github.com/composer/semver/issues/125

**Comment by afbora** [comment]
Thanks @SeriousKen for all details. I think you're right and we have no choice but to use regex. 

Just one open case: I just installed PHP version 8.3.0RC5 and the predefined constants output correctly as 8.3.0. Seems works correctly for RC versions. We don't need to use `PHP_EXTRA_VERSION`. I think we can use predefined constants for the issue.

```php
// PHP 8.3.0RC5
PHP_MAJOR_VERSION   // 8
PHP_MINOR_VERSION   // 3
PHP_RELEASE_VERSION // 0
PHP_EXTRA_VERSION   // RC5
```

afbora:
One more thing. Actually, I think it makes more sense and important to show `PHP 8.3.0RC5` instead of `PHP 8.3.0` to the user/developer in the panel. The user/developer who sees `PHP 8.3.0` may think that the site is working on a stable PHP version. Maybe we should split the php version information into two part. 

- Base version: PHP 8.3.0
- Real version: PHP 8.3.0RC5

SeriousKen:
The constants will work, however, this is about matching updates to a PHP version, in which case 8.3.0 and 8.3.0RC1 are two different versions which is not indicated when using the constants, you need PHP_EXTRA_VERSION and that would still need sanitising to remove the part of the string which does not conform to semantic versioning.On 14 Feb 2024 10:58, Ahmet Bora ***@***.***> wrote:
Thanks @SeriousKen for all details. I think you're right and we have no choice but to use regex.
Just one open case: I just installed PHP version 8.3.0RC5 and the predefined constants output correctly as 8.3.0. Seems works correctly for RC versions.
// PHP 8.3.0RC5
PHP_MAJOR_VERSION   // 8
PHP_MINOR_VERSION   // 3
PHP_RELEASE_VERSION // 0

SeriousKen:
This part of the code doesn't display that to the user, so any call to phpversion elsewhere is left untouched and would use the full version string. This part of the code is only performing a version comparison against a constraint, such as a wildcard same as used in a composer.json file.On 14 Feb 2024 11:07, Ahmet Bora ***@***.***> wrote:
One more thing. Actually, I think it makes more sense and important to show PHP 8.3.0RC5 instead of PHP 8.3.0 to the user/developer in the panel. The user/developer who sees PHP 8.3.0 may think that the site is working on a stable PHP version. Maybe we should split the php version information into two part.
Base version: PHP 8.3.0Real version: PHP 8.3.0RC5

afbora:
> This part of the code doesn't display that to the user,

Yes, I noticed and I had already deleted the comment 🙈 

After the comments above, I'm pinging @lukasbestle and @distantnative to make a final decision or offer a different suggestion.

distantnative:
So if I understand it, we could also to

```php
PHP_MAJOR_VERSION  . '.' . PHP_MINOR_VERSION . '.' . PHP_RELEASE_VERSION
```

Could be slightly more performant than a regex. But it probably doesn't matter and either way would be fine.

SeriousKen:
> So if I understand it, we could also to
> 
> ```
> PHP_MAJOR_VERSION  . '.' . PHP_MINOR_VERSION . '.' . PHP_RELEASE_VERSION
> ```
> 
> Could be slightly more performant than a regex. But it probably doesn't matter and either way would be fine.

As I've said above, this method would result in PHP `8.4.0RC1-1+ubuntu20.04.1+deb.sury.org+1` becoming 8.4.0 whereas the regex would give 8.4.0RC1 and would then mean you couldn't specifically target that version - it is this distinction between the two methods that I think the regex is the safer option to go with.

lukasbestle (approved):
Makes sense to me. I think the regex is fine as it matches everything we are interested in from the beginning of the string.

PR #6263, opened 2024-02-12
Allow unknown file extensions to be selected in file open dialog

rasteiner:
## This PR …

Resumes the work of #4344, but without merge conflicts. 

It:
 - Implements the `acceptAttribute` method in the `FileBlueprint` class. 
 - deprecates the `acceptMime` method
 - adds a `$matchWildcard` attribute to the `Kirby\Filesystem\Mime::toTemplates()` method. 
 - changes the `upload` field mixin and the `files` section config to use `acceptAttribute` instead of `acceptMime`. 

The new `acceptAttribute` function returns a list of extensions, instead of the list of mime types. 

The primary merit is that this allows files to be uploaded which are unknown to Kirby, but without forwading an all allowing `*` to the browser file open dialog. Because Kirby and the browser no longer need to know a mime type for those files. 

The modified `Mime::toTemplates` method can now produce an array of extensions also for a mime type like `image/*`. 

### Breaking changes
I'm not sure if the deprecation is a breaking change, in which case we could just not deprecate the no longer used `toMime()` method. 

## Docs

acceptAttribute also has a clearer docblock which mentions how the list of file extensions is actually built: 
 - If there is a mime type, the "extension" and "type" options are ignored. 
 - If there is no mime type, only the intersection between the "extension" and "type" is returned (not their union). 
The reproduces how acceptMime worked. 

## Ready?
I think so :)

- [X] Tests and checks all pass

### For review team

### DIFF
--- a/src/Cms/FileBlueprint.php
+++ b/src/Cms/FileBlueprint.php
@@ -57,6 +57,8 @@ class FileBlueprint extends Blueprint
 	/**
 	 * Returns the list of all accepted MIME types for
 	 * file upload or `*` if all MIME types are allowed
+	 *
+	 * @deprecated 4.2.0 Use `acceptAttribute` instead
 	 */
 	public function acceptMime(): string
 	{
@@ -113,6 +115,82 @@ class FileBlueprint extends Blueprint
 		}
 
 		// no restrictions, accept everything
+		return '*';
+	}
+
+
+	/**
+	 * Returns the list of all accepted file extensions
+	 * for file upload or `*` if all extensions are allowed
+	 *
+	 * If a mime type is specified in the blueprint, the extensions and types options are ignored for the browser.
+	 * Extensions and Types, however, are still used to validate an uploaded file on the server.
+	 * This behavior might change in the future to better represent what file extensions are actually allowed.
+	 *
+	 * If no mime type is specified, the intersection between manually defined extensions and the Kirby "file types" is returned.
+	 * If the intersection is empty, an empty string is returned.
+	 * This behavior might change in the future to instead return the union of mime, extensions and types.
+	 *
+	 * @since 4.2.0
+	 * @return string
+	 */
+	public function acceptAttribute(): string
+	{
+		// don't disclose the specific default types
+		if ($this->defaultTypes === true) {
+			return '*';
+		}
+
+		$accept = $this->accept();
+
+		// get extensions from "mime" option
+		if (is_array($accept['mime']) === true) {
+			// determine the extensions for each MIME type
+			$extensions = array_map(
+				fn($pattern) => Mime::toExtensions($pattern, true),
+				$accept['mime']
+			);
+
+			$fromMime = array_unique(array_merge(...$extensions));
+
+			// return early to ignore the other options
+			return implode(',', array_map(fn($ext) => ".$ext", $fromMime));
+		}
+
+		$restrictions = [];
+
+		// get extensions from "type" option
+		if (is_array($accept['type']) === true) {
+			$extensions = array_map(
+				F::typeToExtensions(...),
+				$accept['type']
+			);
+
+			// F::typeToExtensions might return null instead of empty arrays,
+			// we need to filter those out
+			$fromType = array_merge(...array_filter($extensions));
+			$restrictions[] = $fromType;
+		}
+
+		// get extensions from "extension" option
+		if (is_array($accept['extension']) === true) {
+			$fromExtension = $accept['extension'];
+			$restrictions[] = $fromExtension;
+		}
+
+		// intersect all restrictions
+		if (count($restrictions) > 1) {
+			$list = array_intersect(...$restrictions);
+		} else {
+			$list = $restrictions[0];
+		}
+
+		$list = array_unique($list);
+
+		// format the list to include a leading dot on each extension
+		return implode(',', array_map(fn($ext) => ".$ext", $list));
+
+		// unknown restrictions, accept everything
 		return '*';
 	}
 
--- a/src/Cms/FileBlueprint.php
+++ b/src/Cms/FileBlueprint.php
@@ -59,6 +59,7 @@ class FileBlueprint extends Blueprint
 	 * file upload or `*` if all MIME types are allowed
 	 *
 	 * @deprecated 4.2.0 Use `acceptAttribute` instead
+	 * @todo 5.0.0 Remove method
 	 */
 	public function acceptMime(): string
 	{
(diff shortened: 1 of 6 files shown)


src/Cms/FileBlueprint.php
```diff
@@ -116,6 +118,79 @@ public function acceptMime(): string
...
+	 * If a mime type is specified in the blueprint, the extensions and types options are ignored for the browser.
+	 * Extensions and Types, however, are still used to validate an uploaded file on the server.
+	 * This behavior might change in the future to better represent what file extensions are actually allowed.
+	 *
+	 * If no mime type is specified, the intersection between manually defined extensions and the Kirby "file types" is returned.
+	 * If the intersection is empty, an empty string is returned.
+	 * This behavior might change in the future to instead return the union of mime, extensions and types.
+	 *
+	 * @since 4.2.0
+	 * @return string
```
distantnative:
```suggestion
```

src/Cms/FileBlueprint.php
```diff
@@ -116,6 +118,79 @@ public function acceptMime(): string
...
+
+		// get extensions from "type" option
+		if (is_array($accept['type']) === true) {
+			$extensions = array_map(
+				F::typeToExtensions(...),
+				$accept['type']
+			);
+
+			// F::typeToExtensions might return null instead of empty arrays,
+			// we need to filter those out
```
distantnative: Wouldn't it be easier to do the following above then

```php
fn ($type) => F::typeToExtensions($type) ?? []
```

src/Cms/FileBlueprint.php
```diff
@@ -116,6 +118,79 @@ public function acceptMime(): string
...
+			$fromType = array_merge(...array_values(array_filter($extensions)));
+			$restrictions[] = $fromType;
+		}
+
+		// get extensions from "extension" option
+		if (is_array($accept['extension']) === true) {
+			$fromExtension = $accept['extension'];
+			$restrictions[] = $fromExtension;
```
distantnative:
```php
$restrictions[] = $accept['extension'];
```

src/Cms/FileBlueprint.php
```diff
@@ -116,6 +118,79 @@ public function acceptMime(): string
...
+		}
+
+		// intersect all restrictions
+		if (count($restrictions) > 1) {
+			$list = array_intersect(...$restrictions);
+		} else {
+			$list = $restrictions[0];
+		}
```
distantnative: Wouldn't this fail for an empty restrictions array? We can also avoid the if/else statement by using `match`

```php
$list = match (count($restroctions)) {
    0 => [],
    1 => $restrictions[0],
    default => array_intersect(...$restrictions)
};
```

src/Filesystem/Mime.php
```diff
@@ -268,17 +269,31 @@ public static function toExtension(string $mime = null): string|false
...
 			}
 
-			if ($value === $mime) {
+			if ($isArray === true && in_array($mime, $value) === true) {
+				$extensions[] = $key;
+			}
+
+			if ($isArray === false && $value === $mime) {
 				$extensions[] = $key;
 			}
```
distantnative, line 298: Since every if statement checks for array, I feel like this should be a larger wrapping if statement. Then we also wouldn't need `$isArray`.

tests/Cms/Blueprints/FileBlueprintTest.php
```diff
@@ -0,0 +1,119 @@
...
+				'notExpected' => ['.gif', '.js', '.pdf', '.docx', '.zip', '.foo', '.bar']
+			],
+		];
+
+		// set up the blueprint files
+		foreach ($this->acceptCases as $name => $case) {
+			Blueprint::$loaded['files/' . $name] = [
+				'accept' => $case['accept']
+			];
+		}
```
distantnative: Could this rather be in a static data provider method?

src/Cms/FileBlueprint.php
rasteiner: In theory the empty array case never happens: if there are no mime types, no extensions and no types, the default types are used and the function returns early with "*". But you're right, this for sure makes it more resilient. :)

distantnative (commented):
Thanks for adding the changes. I think @lukasbestle should have the final review.

src/Cms/FileBlueprint.php
```diff
@@ -57,6 +57,8 @@ public function accept(): array
 	/**
 	 * Returns the list of all accepted MIME types for
 	 * file upload or `*` if all MIME types are allowed
+	 *
+	 * @deprecated 4.2.0 Use `acceptAttribute` instead
```
lukasbestle, line 61: I think the deprecation in 4.2 is fine, however I'd already add a comment for removal so that we don't forget about it:

```suggestion
	 * @deprecated 4.2.0 Use `acceptAttribute` instead
	 * @todo 5.0.0 Remove method
```

I'd say removing in v5 without prior deprecation warning is fine as there was no real use case for external use of this method. What do you think @distantnative?

src/Cms/FileBlueprint.php
```diff
@@ -116,6 +118,75 @@ public function acceptMime(): string
...
+	 * Returns the list of all accepted file extensions
+	 * for file upload or `*` if all extensions are allowed
+	 *
+	 * If a mime type is specified in the blueprint, the extensions and types options are ignored for the browser.
+	 * Extensions and Types, however, are still used to validate an uploaded file on the server.
+	 * This behavior might change in the future to better represent what file extensions are actually allowed.
+	 *
+	 * If no mime type is specified, the intersection between manually defined extensions and the Kirby "file types" is returned.
+	 * If the intersection is empty, an empty string is returned.
+	 * This behavior might change in the future to instead return the union of mime, extensions and types.
```
lukasbestle:
```suggestion
	/**
	 * Returns the list of all accepted file extensions
	 * for file upload or `*` if all extensions are allowed
	 *
	 * If a MIME type is specified in the blueprint, the `extension` and `type` options are ignored for the browser.
	 * Extensions and types, however, are still used to validate an uploaded file on the server.
	 * This behavior might change in the future to better represent which file extensions are actually allowed.
	 *
	 * If no MIME type is specified, the intersection between manually defined extensions and the Kirby "file types" is returned.
	 * If the intersection is empty, an empty string is returned.
	 * This behavior might change in the future to instead return the union of `mime`, `extension` and `type`.
```

tests/Cms/Blueprints/FileBlueprintTest.php
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * @coversDefaultClass \Kirby\Cms\FileBlueprint
+ */
+class FileBlueprintTest extends TestCase
+{
+	public static function acceptAttributeProvider()
+	{
+		return [
+			[
+				'wildcard', // case name
```
lukasbestle: If we pass it as array key in the returned provider array instead of as an argument, PHPUnit will automatically use it as case name for error messages.

tests/Cms/Blueprints/FileBlueprintTest.php
```diff
@@ -0,0 +1,95 @@
...
+
+		foreach ($expected as $extension) {
+			$this->assertStringContainsString($extension, $acceptAttribute, "Case $name: $extension should be accepted");
+		}
+
+		foreach ($notExpected as $extension) {
+			$this->assertStringNotContainsString($extension, $acceptAttribute, "Case $name: $extension should not be accepted");
+		}
+
+		unset(Blueprint::$loaded['files/' . $name]);
```
lukasbestle: This should happen in a `tearDown()` method so it can work even if the test fails (then this line would never be executed). Putting it in `tearDown()` also allows us to use the same file blueprint name for all cases as we can be sure that the global state will always be cleaned up in between.

lukasbestle (commented):
I think the logic is awesome as a first step without breaking changes. Thanks! 👍

lukasbestle (approved):

bastianallgeier:
I'm really glad we were able to finally merge this. What a marathon :) Thanks for your patience and your contribution, Roman!

PR #6292, opened 2024-02-24
Fix support for BETWEEN predicate in database class

SeriousKen:
### Fixes
Using BETWEEN in a where was not supported in the 3 argument form (eg. Db::table('products')->where('quantity', 'BETWEEN', [10, 20])) this PR fixes this.

### Breaking changes
None

## Docs

You can now use `Db::table('products')->where('quantity', 'BETWEEN', [$min, $max])` to add a BETWEEN clause to a query. 

## Ready?

### For review team


src/Database/Query.php
```diff
@@ -903,15 +904,17 @@ protected function filterQuery(array $args, $current, string $mode = 'AND')
 							$values[] = $valueBinding;
 						}
 
-						// add that to the where clause in parenthesis
-						$result = $key . ' ' . $predicate . ' (' . implode(', ', $values) . ')';
+						// add that to the where clause in parenthesis or seperated by AND
+						$result = $key . ' ' . $predicate . (in_array($predicate, ['IN', 'NOT IN'])
+							? ' (' . implode(', ', $values) . ')' // IN predicate
+							: ' ' . $values[0] . ' AND ' . $values[1] // BETWEEN predicate
+						);
```
distantnative: This got even harder to read now. Maybe

```php
$values = match ($predicate) {
    'IN', 
    'NOT IN'      => '(' . implode(', ', $values) . ')',
    'BETWEEN', 
    'NOT BETWEEN' => $values[0] . ' AND ' . $values[1]
};
$result = $key . ' ' . $predicate . ' ' . $values;

distantnative:
@SeriousKen don't worry about the failing code styling test. We changed something on our CI yesterday and the reviewdog action seems to be broken right now.

src/Database/Query.php
SeriousKen: Yes, that looks much better, thanks.

distantnative (commented):
Looks good to me (thanks!). Leaving the final review to @lukasbestle as he knows those SQL classes much better.

afbora:
BETWEEN operator also supports text values like `BETWEEN 'Foo' AND 'Bar'`. I just wonder that is this PR also supports strings?

SeriousKen:
> BETWEEN operator also supports text values like `BETWEEN 'Foo' AND 'Bar'`. I just wonder that is this PR also supports strings?

I can't see any reason why it would not support strings as it ultimately gets translated into to a prepared statement with parameters.

afbora:
Great! So could you add unit test for strings please?

SeriousKen:
> Great! So could you add unit test for strings please?

No problem, test added

lukasbestle (approved):
Awesome, thanks!

PR #6294, opened 2024-02-24
Fix `Toolkit\Dom::body()` for false query

distantnative:
### Fixes
- `Toolkit\Dom::body()`: fixed for non-existent fallback

### DIFF
--- a/src/Toolkit/Dom.php
+++ b/src/Toolkit/Dom.php
@@ -125,7 +125,11 @@ class Dom
 	 */
 	public function body(): DOMElement|null
 	{
-		return $this->body ??= $this->query('/html/body')[0] ?? null;
+		if ($this->body === null && $query = $this->query('/html/body')) {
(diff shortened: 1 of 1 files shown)


lukasbestle (commented):
Hm, I don't really understand which case this change will protect against. The following still works for me:

```php
$var = false;
var_dump($var[0] ?? null);
```

distantnative:
Was a result of checking more code quality with Psalm. Got flagged there as improper.

lukasbestle:
Can you reproduce that on https://psalm.dev? If not, what is the full warning/error you get from Psalm?

distantnative:
It's from an older branch of a lot of small changes I collected. But honestly not worth it spending the time again trying to find out what Psalm setting I changed to see the error. Probably shouldn't look at Psalm much anyways as it keeps also throwing false positives. Closing this.

PR #6295, opened 2024-02-24
Lab: add GitHub source links

distantnative:
### Housekeeping
- Lab: added GitHub source links for all helpers, libraries and CSS examples
- Removed `Lab\Example::github()` linking to lab example file


bastianallgeier (approved):
This is really nice!

PR #6297, opened 2024-02-25
Improved parameter binding for inserts and selects

SeriousKen:
## This PR …

Improved parameter binding for database inserts and selects.

### Fixes
Fixes a problem where binding a boolean false would result in an attempt to insert a blank string into the table rather than 0, This would insert the value with Sqlite but silently fail on MySQL unless exceptions are turned on. Also fixes a similar issue when selecting against boolean columns.

### Breaking changes
None

## Docs

You can now insert a `false` value and it gets correctly bound to the right database type (eg. `Db::insert('users', ['username' => 'paul', 'active' => true])`). You can also check for equality against a boolean value (eg. `Db::table('users')->where('active', '=', true)` )

## Ready?

### For review team

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -305,7 +305,23 @@ class Database
 		// try to prepare and execute the sql
 		try {
 			$this->statement = $this->connection->prepare($query);
-			$this->statement->execute($bindings);
+			// bind parameters to statement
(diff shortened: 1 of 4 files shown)


lukasbestle (approved):
Thanks, looks perfect to me.

PR #6299, opened 2024-02-25
Panel theme

distantnative:
### TODO
- [x] use `localStorage` instead of user content field

### Feature
- Dark theme for the Panel

### Enhancements
- New `--color-l-min` and `--color-l-max` CSS properties

### Deprecated
- `--color-backdrop` CSS property has been deprecated. Use `--color-light` instead.

## Ready?

### For review team

### DIFF
--- a/config/areas/account/dialogs.php
+++ b/config/areas/account/dialogs.php
@@ -1,11 +1,12 @@
 <?php
 
+use Kirby\Cms\App;
+use Kirby\Panel\Panel;
 use Kirby\Panel\UserTotpEnableDialog;
 
 $dialogs = require __DIR__ . '/../users/dialogs.php';
 
 return [
-
 	// change email
 	'account.changeEmail' => [
 		'pattern' => '(account)/changeEmail',
@@ -88,6 +89,16 @@ return [
 		'pattern' => '(account)/files/(:any)/fields/(:any)/(:all?)',
 		'load'    => $dialogs['user.file.fields']['load'],
 		'submit'  => $dialogs['user.file.fields']['submit']
+	],
+
+	// change Panel theme
+	'account.theme' => [
+		'pattern' => 'account/theme/(:any)',
+		'load' => function (string $theme) {
+			$user = App::instance()->user();
+			$user->update(['theme' => $theme]);
+			Panel::go($user->panel()->url());
+		}
 	],
 
 	// account enable TOTP
--- a/config/areas/account/dialogs.php
+++ b/config/areas/account/dialogs.php
@@ -1,7 +1,5 @@
 <?php
 
-use Kirby\Cms\App;
-use Kirby\Panel\Panel;
 use Kirby\Panel\UserTotpEnableDialog;
 
 $dialogs = require __DIR__ . '/../users/dialogs.php';
@@ -89,16 +87,6 @@ return [
 		'pattern' => '(account)/files/(:any)/fields/(:any)/(:all?)',
 		'load'    => $dialogs['user.file.fields']['load'],
 		'submit'  => $dialogs['user.file.fields']['submit']
-	],
-
-	// change Panel theme
-	'account.theme' => [
-		'pattern' => 'account/theme/(:any)',
-		'load' => function (string $theme) {
-			$user = App::instance()->user();
-			$user->update(['theme' => $theme]);
-			Panel::go($user->panel()->url());
-		}
 	],
 
 	// account enable TOTP
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -27,6 +27,9 @@
 	"date": "Date",
 	"date.select": "Select a date",
 
+	"dark.on": "Lights off",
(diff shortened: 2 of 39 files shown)


afbora:
I don't understand why we save the dark mode setting in the user file contents. I think it would make more sense to save it in `localStorage` like most systems do.

Also, if we use `localStorage`, I would prefer to have 3 modes: Auto, light, dark.

distantnative:
Do most systems use localStorage? I thought it would be a good user preference to be available across devices, browsers etc. and not bound to one browser only.

I think we should keep it simple: it's auto by default unless the user sets it to dark or light explicitly.

afbora:
IMHO, with `localStorage` this is much simpler. It only deals with the frontend part and does not involve any breaking changes.

> I thought it would be a good user preference to be available across devices, browsers etc. and not bound to one browser only.

This makes sense, but if I use my mobile or other devices in dark mode, it will be applied automatically anyway.

afbora:
screen-capture-1837-Log in - Mægazine-localhost

When auto-filled.

distantnative:
> This makes sense, but if I use my mobile or other devices in dark mode, it will be applied automatically anyway.

Yes as long as you haven't set a setting/you are on auto when your phone is in dark mode. But there might be users who want light mode even when they use OS on dark mode.

distantnative:
@afbora fixed the auto-filled color

bastianallgeier:
I didn't have a chance to test and comment on this yet. I think it already looks absolutely amazing! I was wondering if our variable system needs more work. Did you run into more places where you felt like the css variables don't really fit? I was thinking if more of them should be more universal instead of describing a color.

distantnative:
Semantically a lot of the variable names don't make sense: `---color-white` is literally black in dark mode. However, I think it would bee quite a big effort the refactor it and add an intermediary layer of Moree abstract CSS properties in between. We might want to do that in places over time, but honestly I think reassigning colours based on the theme will always be a huge effort considering my attempt is just more or less flipping the lightness values and maybe a little more than a handful specific rules only.

afbora:
@distantnative I just wonder that why you switched to localStorage way? 🙂

distantnative:
@afbora some days of thinking and your argument convinced me :)

bastianallgeier:
Does anything speak against merging this already into the alpha branch?

distantnative:
@bastianallgeier not from my side

afbora:
I would prefer dark/light mode toggle move to global position like sidebar. For example case: what if a user doesn't have `account` permission?

afbora:
Except toggle position, seems everything fine for me. I loved it 😍

afbora:
Also just an idea; since there will be custom plugins and fields in the panel, some users may want to disable it in case of incompatibilities with the plugins. In this case, we can enable this by default and the user can disable this setting in config if they want.

distantnative:
Too many options/settings in my pinion. It makes it much more complex. Given the nature how we do our dark mode, most plugins should work without many tweaks needed. And they have 9 months or so to make those adaptations.

bastianallgeier:
I think any further decisions about the toggle position and other parts are totally something that we can solve in the next months.

bastianallgeier (approved):

WebMechanic:
**Just a heads up:** We had several Sentry issues popping up concerning `localStorage` in **Chrome**. 
There happens to be a "don't save absolutely no website data" privacy settings that also prevents access to the Storage interfaces and errors with "access denied".
I'm not sure if this is a very recent thing, Chrome and/or Edge/Brave and whether its `localStorage` (persistent) only 'cos we never saw anything on `sessionStorage` which we use at the same time. 
I currently don't have access to these browsers to tell where to find such setting, just be aware that there might be issues.

I was under the impression Storage access follows the user's cookie settings - it does so in Firefox iirc.

distantnative:
@WebMechanic currently Kirby also stores all unsaved changes in `localStorage`. From your report, that'd be all broken too on your setups?

WebMechanic:
> From your report, that'd be all broken too on your setups?

not our (or my) setup, but apparently some of our users are xtra paranoid :-) 🤷‍♂️ 
This happens in a Vue3/VueUse project "at work" which is unrelated to Kirby.
Sentry simply reports "Chrome" as the only browser to be "affected". We never had SF or FF triggering this (or Edge iirc). It's also **_not thousands of errors a day_**, just enough to raise an eyebrow. It's when we learned about this "don't save **_any_** site data" thing.

I check with my colleagues, but we can't exactly pinpoint a version. It started a few weeks back. Might as well be related to Google's "Cookie Phaseout" or "Cookie Deprecation Trial" and people browsing with some obscure "Web Experiment" enabled.

I use Kirby privately and ended here via the v5 announcement and though I share this.
We probably have a very different demographic and user base so this might never show up for you.

distantnative:
@WebMechanic Please let us know if you receive more information from them. No localStorage would render the Panel useless currently. Not because of the dark theme, but because content changes would not work either then.

Issue #6291, opened 2024-02-24
Inconsistent behaviour with where clause in database when parameter is integer


SeriousKen:
## Description

Using the database abstraction the following should be equivalent:

`Db::table('products')->where('quantity < ?', 10)->all()`

`Db::table('products')->where('quantity', '<', 10)->all()`

**Expected behavior**  

Both should return all products which have a quantity less than 10. However, the first returns all products - the query silently ignores the _where_ part of the query when the parameter is an integer.

`var_dump(Db::table('products')->where('quantity < ?', 10)->build('select'));` gives the following output:

```
array(2) {
  ["query"]=>
  string(24) "SELECT * FROM `products`"
  ["bindings"]=>
  array(0) {
  }
}
```

## Your setup
**Kirby Version**  

4.1.0

**Your system**
 - Device: Dell Inspiron Laptop
 - OS: Ubuntu 20.04 on WSL2 on Windows11

SeriousKen:
If I put the parameter in an array (eg. `Db::table('products')->where('quantity < ?', [10])->all();`) it works as expected:

lukasbestle:
The bug is that in the mode of two arguments, only arrays and strings are supported.

I think the following line needs to be modified:

https://github.com/getkirby/kirby/blob/2965c3124e3b141072a2d46c798a327dda710060/src/Database/Query.php#L872

The second check should be `is_scalar($args[1])` instead of `is_string($args[1])`. However I'm not sure if `bool`s (which are scalar) will work. Maybe it instead needs to be `is_string() || is_int() || is_float()` then.

@afbora Could you work on a fix?

SeriousKen:
What about `is_scalar()` instead? Same thing but easier to read.

SeriousKen:
Bools do seem to work as they are converted to 1 or 0 when they are bound to parameters, so is_scalar should be OK.

Edit: Actually just checked, PDO seems to be inconsistent with treatment of boolean parameters - it works as expected with MySQL but false does not get converted to int with Sqlite.

SeriousKen:
If you change the code here

https://github.com/getkirby/kirby/blob/2965c3124e3b141072a2d46c798a327dda710060/src/Database/Database.php#L307-L308

to explicitly bind with the correct types rather than rely on the execute method of PDOStatement then things work consistently with booleans.

```php
$this->statement = $this->connection->prepare($query);
        foreach ($bindings as $binding => $value) {
        $type = match (gettype($value)) {
[...]

PR #6300, opened 2024-02-25
Support non string scalar values in where clause builder for 2 arguments

SeriousKen:
## This PR …

Adds support for non-string scalar values in where clause builder when using 2 arguments. Previously non-string scalar values would be silently ignored.

### Fixes
https://github.com/getkirby/kirby/issues/6291

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -869,7 +869,7 @@ class Query
 					$this->bindings($args[1]);
 
 				// ->where('username like ?', 'myuser')
-				} elseif (is_string($args[0]) === true && is_string($args[1]) === true) {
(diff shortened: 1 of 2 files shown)


distantnative:
@SeriousKen I think your PR is based on develop-minor but you're pointing at develop-patch (that's why it has so many commits shown).

SeriousKen:
Oops, yes it should be develop-minor. I think I've fixed that.

bastianallgeier (approved):

bastianallgeier:
Thank you for the PR!!

PR #6310, opened 2024-02-29
Fix commands for custom textarea buttons

### LINKED ISSUE #6039: Custom textareaButtons: crash when multiple times rendered
## Description
Using a plugin to add a custom textareaButton (as described in the docs and [this recent issue](https://github.com/getkirby/kirby/issues/6028)) fails to work reliably, can cause entire page to timeout/crash.

Creating a custom button and registering a click event for it will only works in one single instance of the textarea. That's to say, in one field on one page. Navigating the panel to another page to use the custom button again fails. What's worse, setting up a blueprint with two unique textarea fields that both use the custom button will grind the page to a halt.

**Expected behavior**  
1. Button is added to one textarea or several textareas via the `buttons:` property in a blueprint.
2. `click: function() { // do something }` should trigger the appropriate action.
3. Navigate to a new page or the next textarea, the button should continue to trigger the appropriate action

## To reproduce
Create a new plugin using the sample code from getkirby.com/releases/4.0 for a new textareaButton in an index.js file.
Use a long-form function rather than arrow function as [uncovered here](https://github.com/getkirby/kirby/issues/6028)
```
panel.plugin("getkirby/custom-textarea-buttons", {
  textareaButtons: {
    highlight: {
      label: "Highlight",
      icon: "wand",
      click: function() {
        this.command("toggle", "<mark>", "</mark>")
      },
      shortcut: "m",
    },
  }
});
```
**Test 1: Sibling Failure**
Add the highlight button to a textarea field in a blueprint.
Load the panel and edit the page.
Click on the custom button. It should work. 
Navigate to a sibling page and click the custom button. Nothing will happen in the textarea -- no markup is inserted whether text is selected or not.
**Test 2: Multiple textareas crash page**
Add the highlight button to two different textarea fields in a blueprint.
Load the panel and edit the page.
The page times out / crashes.

## Your setup
**Kirby Version**  
 4.0.0

 - Device: Apple iMac
 - OS: MacOS 12.7
 - Browser: Firefox 120.0 and Safari 17.0
 - PHP 8.2

bastianallgeier:
### Fixes

- Fix commands for custom textarea buttons #6039 

### Breaking changes

None

### For review team

### DIFF
--- a/panel/src/components/Forms/Toolbar/TextareaToolbar.vue
+++ b/panel/src/components/Forms/Toolbar/TextareaToolbar.vue
@@ -153,10 +153,13 @@ export default {
 				if (button === "|") {
 					layout.push("|");
 				} else if (available[button]) {
-					// ensure that click handler is bound to this component
-					// so that plugins can use `this.command` etc.
-					const command = available[button];
-					command.click = command.click?.bind(this);
+					const command = {
+						...available[button],
+						click: () => {
+							available[button].click?.call(this);
+						}
+					};
+
 					layout.push(command);
 				}
 			}


bastianallgeier:
@afbora you are the expert in testing stuff like this. Could you give it a go and check out if the textarea buttons now work for you? I've added an example to the textarea field setup in the sandbox (lab environment)

afbora (approved):
Works great for me. I just wonder that what's the difference?

````js
const command = {
  ...available[button],
  click: () => {
    available[button].click?.call(this);
  }
};

const command = {
  ...available[button],
  click: available[button].click?.bind(this)
};
````

bastianallgeier:
@afbora for some reason, `this` is not correctly bound like that and I'm honest with you, I have no idea why.

Issue #5551, opened 2023-09-01
[V4] UUID permalinks always redirect to default language page


dviate:
## Description
When the languages are enabled, when you add a link to another page, the `@/page/uuid`-link is the same for every language.

**Expected behavior**  
When I'm on a page in any non-default language, for instance `site.com/es/sobre-nosotros/`, where the default page link will be `site.com/about-us/`, I want to open the page in that language and not the default language.

## To reproduce
1. Enable languages in your `config.php`

```
/*
* https://getkirby.com/docs/reference/system/options/languages
*/
'languages' => [
	'detect' => false
],
```

2. Make your own language your default language and add another non-default one, for instance Spanish
3. Create a few page, type some stuff and add on one of those pages a link to another one

In your default language it works, when you open the Spanish version of the page and you click the link, it redirects you to the linked page in the default language. Editing the link on the Spanish version of the page doesn't help either.

## Your setup
**Kirby Version**  
4.0.0-alpha.7

**Language files in `site/languages/`**
NL (for me)
```
<?php
    return [
        'code' => 'nl',
        'default' => true,
        'direction' => 'ltr',
        'locale' => 'nl_NL',
        'name' => 'Nederlands',
        'url' => '/',
        'translations' => [ *etc*
```
ES (test)
```
<?php
    return [
        'code' => 'es',
        'default' => false,
        'direction' => 'ltr',
        'locale' => 'es_ES',
        'name' => 'Español',
        'url' => '/es',
        'translations' => [ *etc*
```

distantnative:
Permalinks could gain that we optionally add the language code at the end `@/page/dojfppdz/de` and then redirect to the right language.

And then the link field would need to support that in multilingual setups.

distantnative:
@lukasbestle What do you think of this idea to add language support to permalinks? Or do you have another idea?

lukasbestle:
In theory, a UUID could contain a slash if a custom generator is used. This would then break the language code separated by a slash. But I think the UUID cache already doesn't work with slashes in UUIDs, right? If that's the case, I agree with your suggestion. It's functional but also clean.

distantnative:
Could also be `?lang=de` but I'd think that could have the same problem.

lukasbestle:
That would already not have worked as the question mark would have ended the path parsed by PHP/our router.

[...]

Issue #5181, opened 2023-05-23
File Permalinks throw 404 if UUID cache is empty


tobimori:
## Description
A client of mine uses the newly added UUID Permalinks for sending file links to customers. When a new deployment of the site happens, we flush the complete cache, including the UUID cache.  (essentially `rm -rf` the folder) 

Now, the non-presence of the UUID folder will make all Permalinks fail as 404, until they are created again. As the media folder URLs are prone to change, we need to use the permalinks.

**Expected behavior**  
Permalinks need to also work when the UUID cache hasn't been generated yet, which means that any access of a permalink route should also generate the UUID cache.

## To reproduce
1. Delete the UUID cache.
2. Try to open a file via its Permalink.
3. 404
4. Trigger anything that creates the UUID cache (e.g. store and save a file)
5. Permalink works again

I have not tested this with Pages, but I suspect the behavior to be the same.

## Your setup
**Kirby Version** 3.9.4

distantnative:
So far this is intentional, we can debate though if it's the right call we made:

Since looking up UUIDs without the cache means traversing through the whole site index, especially for large sites allowing to trigger this action via a permalink request seemed to be an easy vector for DDOS attacks. Which is why we made the lookup for permalink lazy, only from the cache.

tobimori:
I understand the decision behind that. Given that, it should probably still be resolved in some way when e.g. the entire cache is empty. Right now the behavior doesn't work for any site where the cache folder is non-persistent, which will probably be the case for most "modern" deploy setups. If it's actually a problem, make it a config option so large sites can turn it off if needed.

bastianallgeier:
As a quick fix, I would suggest that you repopulate the UUID cache after deployment.

bnomei:
just my 2c. for autoid i was only running the indexing if the current index count was 0.

lukasbestle:
What Bruno wrote could be an option. Then the DoS/DDoS potential is very low. But we might need a mechanism to ensure that only one PHP process can start the time-intensive indexing process at a time.

distantnative:
@lukasbestle is there a good way to check how many entries one of our caches has?

lukasbestle:
Not a way that works with all cache drivers. But we could do it something like this:

- When storing any value in the UUID cache, call `$cache->getOrSet('status', fn () => false)`. This will ensure that a cache key `status` exists (but without writing the same value all the time, reducing wear impact on flash storage).
[...]

PR #6312, opened 2024-02-29
Add language specific UUID URLs

bastianallgeier:
## This PR …

This is an attempt to finally get parts of the multi-language UUID issues solved. (https://github.com/getkirby/kirby/issues/5551)

This adds a new route for language-specific UUID urls: 

Examples: 
```
/en/@/page/1234
/en/@/file/1234
```

## Ready?

### For review team

### DIFF
--- a/src/Cms/LanguageRouter.php
+++ b/src/Cms/LanguageRouter.php
@@ -7,6 +7,7 @@ use Kirby\Exception\NotFoundException;
 use Kirby\Http\Router;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\Str;
+use Kirby\Uuid\Uuid;
 
 /**
  * The language router is used internally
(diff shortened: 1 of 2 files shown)


distantnative:
Either we should add unit tests or we should be more explicit what we want to ignore for coverage.

bastianallgeier:
@distantnative I'm trying to create tests for the UUID route but I fail to setup a test environment with working UUIDs. Could you maybe have a look at the test? Why can the model not be found like that?

distantnative:
Had you tried manually adding a uuid in the app setup `['content' => ['uuid' => 'my-uuid']]`? That's how many of the other tests do it.

bastianallgeier:
Yep, I tried that too but also without luck.

distantnative:
@bastianallgeier Ok I finally understood where it fails:
- The uuid route looks up the model lazily, so only from cache. Since the cache is not populated, it fails
- When I was trying to manually set a UUID, I did it with `content`. But we are in multilang territory, so we have to set the `translations` array instead

Alternatively, solving https://github.com/getkirby/kirby/issues/5181 too.

distantnative:
@bastianallgeier Apparently, sometimes it takes months to realize what one is overlooking. I even wrote

> Since the cache is not populated, it fails.

But only today it clicked that, well yea then we need to add a `$uuids->populate()` line to add it to the cache.

distantnative (dismissed):
@bastianallgeier Please check if the unit test is testing the right things from your perspective.

bastianallgeier:
@distantnative yes, looks good to me! I can't believe that we might actually be able to merge this :)

distantnative (approved):

PR #6321, opened 2024-03-02
`k-input-validator`

distantnative:
### Enhancements
- Added `k-input-validator` helper element to provide native form validation for complex/group inputs and fields

### For review team

### DIFF
--- a/panel/lab/components/fields/blocks/index.vue
+++ b/panel/lab/components/fields/blocks/index.vue
@@ -1,11 +1,20 @@
 <template>
-	<k-lab-examples>
+	<k-lab-examples class="k-lab-field-examples">
 		<k-box theme="notice" icon="alert">
 			The blocks field does not work without the backend yet
 		</k-box>
 
 		<k-lab-example label="Default">
 			<k-blocks-field label="Blocks" :value="value" @input="value = $event" />
+		</k-lab-example>
+
+		<k-lab-example label="Required">
+			<k-blocks-field
+				label="Blocks"
+				:required="true"
+				:value="value"
+				@input="value = $event"
+			/>
 		</k-lab-example>
 	</k-lab-examples>
 </template>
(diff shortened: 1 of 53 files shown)


panel/lab/components/inputs/list/index.vue
```diff
@@ -33,6 +33,14 @@
...
 				<k-list-input :disabled="true" :value="value" @input="value = $event" />
 			</k-lab-example>
+
+			<k-lab-example label="Minlength: 5">
+				<k-list-input :minlength="5" :value="value" @input="value = $event" />
+			</k-lab-example>
+
+			<k-lab-example label="Maxlength: 10">
+				<k-list-input :maxlength="10" :value="value" @input="value = $event" />
+			</k-lab-example>
```
bastianallgeier: They both don't really seem to do anything.

bastianallgeier (commented):
I really like the idea of the input validator. I think we need to add an option though to provide an element that the validity check can focus on. Right now, there are still console issues because of it and no message can be displayed on submit.

distantnative:
> provide an element that the validity check can focus on

You mean a normal, non-input element to be focussed?

bastianallgeier:
@distantnative I haven't tried if that actually works. But e.g. for the checkboxes we could focus the first checkbox at least. I'm not sure how we should properly handle it when there's actually no valid form element to focus on – e.g. for blocks.

distantnative:
@bastianallgeier I fear we cannot target any other element that isn't a descendant of `k-input-validator`:

```
Failed to execute 'setValidity' on 'ElementInternals': The Element argument should be a shadow-including descendant of the target element.
```

So we would have to wrap the whole component in it... which makes it a little nasty/ugly.

However, according to the docs, any HTMLElement can be the anchor then.

distantnative:
But this seems to be fake news. HTMLElement as anchor still causes console error for me.

distantnative:
@bastianallgeier I think I found a way

bastianallgeier (approved):
Awesome. It works really well like this 💛

bastianallgeier:
Could you take care of the conflicts after the class attributes? I think I just broke something while trying for the checkboxes input.

bastianallgeier (approved):

PR #6323, opened 2024-03-03
Fix WriterField counter

distantnative:
### Fixes
- Writer field counter correctly strips HTML again

### For review team


bastianallgeier (commented):
Something seems to be off. When I open the writer field lab example, the counter is at 4 at first when the field is still empty.

distantnative:
Ok sometimes JavaScript is just.... why is `String(null)` not `""` but `"null"`?

bastianallgeier (approved):

PR #6326, opened 2024-03-04
Toolkit: `@covers` annotations

afbora:
## This PR …

I've refactored all namespaces except the `\Kirby\Cms` in my local branch. But I'll create PRs for each namespace separately after getting your reviews. This is only for `\Kirby\Toolkit` one.

### Refactor
- Toolkit: `@covers` annotations

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/tests/Toolkit/CollectionGetterTest.php
+++ b/tests/Toolkit/CollectionGetterTest.php
@@ -2,8 +2,14 @@
 
 namespace Kirby\Toolkit;
 
+/**
+ * @coversDefaultClass \Kirby\Toolkit\Collection
+ */
 class CollectionGetterTest extends TestCase
 {
+	/**
+	 * @covers ::__call
+	 */
 	public function testGetMagic()
 	{
 		$collection = new Collection([
@@ -16,6 +22,9 @@ class CollectionGetterTest extends TestCase
 		$this->assertNull($collection->three);
 	}
 
+	/**
+	 * @covers ::get
+	 */
 	public function testGet()
 	{
 		$collection = new Collection([
@@ -28,6 +37,9 @@ class CollectionGetterTest extends TestCase
 		$this->assertSame('default', $collection->get('three', 'default'));
 	}
 
+	/**
+	 * @covers ::__call
+	 */
 	public function testMagicMethods()
 	{
 		$collection = new Collection([
@@ -40,6 +52,9 @@ class CollectionGetterTest extends TestCase
 		$this->assertNull($collection->three());
 	}
 
+	/**
+	 * @covers ::toArray
+	 */
 	public function testGetAttribute()
 	{
 		$collection = new Collection([
--- a/tests/Toolkit/PaginationTest.php
+++ b/tests/Toolkit/PaginationTest.php
@@ -6,6 +6,9 @@ use Kirby\Exception\ErrorPageException;
 use Kirby\Exception\Exception;
 use Kirby\TestCase;
 
+/**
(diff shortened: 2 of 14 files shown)


tests/Toolkit/CollectionGetterTest.php
```diff
@@ -2,8 +2,14 @@
 
 namespace Kirby\Toolkit;
 
+/**
+ * @coversDefaultClass \Kirby\Toolkit\Collection
+ */
 class CollectionGetterTest extends TestCase
 {
+	/**
+	 * @covers ::__call
```
lukasbestle, line 11:
```suggestion
	 * @covers ::__get
```

tests/Toolkit/CollectionGetterTest.php
```diff
@@ -40,6 +52,9 @@ public function testMagicMethods()
 		$this->assertNull($collection->three());
 	}
 
+	/**
+	 * @covers ::toArray
```
lukasbestle, line 56:
```suggestion
	 * @covers ::getAttribute
```

tests/Toolkit/VTest.php
```diff
@@ -68,6 +68,9 @@ public function testCallInvalidMethod()
 		V::fool('me');
 	}
 
+	/**
+	 * @covers ::__callStatic
```
lukasbestle, line 72: Not sure about those TBH. I'd expect a test for `__callStatic()` to test the logic inside the method (calling a validator that exists, one that exists but with different case, one that doesn't exist as well as passing arguments correctly to the validator). The actual validators are IMO more a case for `@coversNothing`.

tests/Toolkit/PaginationTest.php
```diff
@@ -340,6 +435,8 @@ public static function rangeProvider(): array
 	}
 
 	/**
+	 * @covers ::__construct
+	 * @covers ::range
```
lukasbestle, line 439:
```suggestion
	 * @covers ::range
	 * @covers ::rangeStart
	 * @covers ::rangeEnd
```

tests/Toolkit/PaginationTest.php
```diff
@@ -459,6 +591,13 @@ public function testToArray()
 		}
 	}
 
+	/**
+	 * @covers ::for
+	 * @covers ::page
+	 * @covers ::pages
+	 * @covers ::limit
+	 * @covers ::total
```
lukasbestle, line 599: These are not really tested here IMO, same for the following tests.

lukasbestle (commented):
This is great. 💛

A general note: The added annotations make our total coverage go down a lot. I wonder what the best way forward is. IMO it is better to have a realistic view on the code parts that have or need tests than to have a nice high number. But the real coverage percentage is somewhere in between as we do test parts of underlying (not explicitly covered) logic in tests all around our codebase, we just haven't gotten around to add explicit tests for each relevant method. So if this reduces our coverage percentage by a lot, it could be seen as negative on the outside even though our test quality actually improved. Not sure...

bastianallgeier:
I realized that the latest PHPUnit version is switching to PHP attributes for all of this 🙈 I just want to keep this in mind when we spend so much time on it. I'm also with Lukas. We need to be careful about the way we deal with coverage. To what extend are we making the machines happy and to what extend are we actually improving our tests. My bigger issue with our tests is the rather chaotic way that older tests are written to be honest.

afbora:
Can you be a bit more specific? 🙈 How should we proceed as a result?

PR #6332, opened 2024-03-06
Explicitly passing `class` attribute

distantnative:
for components that not inherit attributes

### Refactor
- Explicitly added `$attrs.class` to components that disable inheriting attributes

### DIFF
--- a/panel/src/components/Forms/Blocks/BlockPasteboard.vue
+++ b/panel/src/components/Forms/Blocks/BlockPasteboard.vue
@@ -1,11 +1,11 @@
 <template>
 	<k-dialog
 		ref="dialog"
+		:class="['k-block-importer', $attrs.class]"
 		:cancel-button="false"
 		:submit-button="false"
 		:visible="true"
 		size="large"
-		class="k-block-importer"
 		@cancel="$emit('cancel')"
 		@submit="$emit('submit')"
 	>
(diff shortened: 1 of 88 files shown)


panel/src/components/Forms/Blocks/BlockPasteboard.vue
```diff
@@ -1,11 +1,11 @@
 <template>
 	<k-dialog
 		ref="dialog"
+		:class="['k-block-importer', $attrs.class]"
```
bastianallgeier, line 4: Something seems to be broken with this one. Maybe because of the way that dialogs pass along the class attribute?

bastianallgeier: It's weird because everything else seems to work just as before.

distantnative: Yes, because we only used the static class in `k-dialog` and now it's assigned dynamically. Changed now to `k-dialog` using whichever all.

distantnative: The upside: in Vue 3 we can get rid of the `$vnode.data.` hack as `$attrs.class` will work there as well.

bastianallgeier (approved):

Issue #6328, opened 2024-03-05
[Panel] Keep title when switching template in create dialog


mrflix:
When changing the template after typing in the title in the page creation dialog, the title doesn't persist.

**Expected behavior**  
I'd expect the title to stay there. I know that the creation dialog is configurable. I'd expect all fields that are the same to keep their values while switching the template.

**Screenrecording**  

https://github.com/getkirby/kirby/assets/60777/bbf8ab2a-a94e-4011-beaa-5247341c8154

## Your setup
**Kirby Version**  
4.1.1

PR #6334, opened 2024-03-07
Keep title when switching template in create dialog 

afbora:
## This PR …

I'm not sure the solution but it seems works for me great. Also I couldn't write unit test for getting title and value prop while changing template.

### Fixes
- Keep title when switching template in create dialog #6328

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Panel/PageCreateDialog.php
+++ b/src/Panel/PageCreateDialog.php
@@ -369,9 +369,9 @@ class PageCreateDialog
 		$value = [
 			'parent'   => $this->parentId,
 			'section'  => $this->sectionId,
-			'slug'     => '',
+			'slug'     => $this->slug ?? '',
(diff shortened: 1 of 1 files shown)


distantnative:
The problem is that any other (custom) fields would still be lost.

afbora:
Yes. I see. I'll check out. But if you think this (getting values of custom fields) is complex one, I would merge the PR to fix the regression.

afbora:
@distantnative By the way, since custom fields will usually be different (of course they can be the same) based on template, it makes sense to keep only the title and slug common to all templates for now.

distantnative:
Fair point

bastianallgeier (approved):

bastianallgeier:
I agree, let's keep this simple first.

PR #6337, opened 2024-03-07
Plugin registration: named arguments

distantnative:
## This PR …

Background: https://github.com/getkirby/kirby/pull/5764

### Enhancement
- Plugin registration with named arguments
- Allow passing plugin info, e.g. version, as argument that gets merged with the info from composer.json
```php
Kirby::plugin(
    name: 'my/plugin',
    extends: [],
    info: [
        'homepage' => 'https://my-plugin.example.com'
    ],
    version: '1.0.0'
);
```

### Deprecated
- Plugin registration: Passing `info` or `root` in plugin extends array. Instead pass these as standalone named arguments.

## Docs

- Update examples of how to register a plugin throughout getkirby.com to use named arguments

## Ready?

### For review team

### DIFF
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -714,18 +714,17 @@ trait AppPlugins
 	public static function plugin(
 		string $name,
 		array $extends = null,
-		string|null $version = null
+		string|null $version = null,
+		string|null $root = null
 	): PLugin|null {
 		if ($extends === null) {
 			return static::$plugins[$name] ?? null;
 		}
 
-		// get the correct root for the plugin
-		$extends['root'] ??= dirname(debug_backtrace()[0]['file']);
-
 		$plugin = new Plugin(
 			name:    $name,
 			extends: $extends,
+			root:    $root,
 			version: $version
 		);
 
@@ -798,7 +797,11 @@ trait AppPlugins
 				// register as anonymous plugin (without actual extensions)
 				// to be picked up by the Panel\Document class when
 				// rendering the Panel view
-				static::plugin('plugins/' . $dirname, ['root' => $dir]);
+				static::plugin(
+					name: 'plugins/' . $dirname,
+					extends: [],
+					root: $dir
+				);
 			} else {
 				continue;
 			}
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -714,7 +714,7 @@ trait AppPlugins
 	public static function plugin(
 		string $name,
 		array $extends = null,
-		string|null $version = null,
+		array $info = [],
 		string|null $root = null
 	): PLugin|null {
 		if ($extends === null) {
@@ -724,8 +724,8 @@ trait AppPlugins
 		$plugin = new Plugin(
 			name:    $name,
 			extends: $extends,
+			info:    $info,
 			root:    $root ?? dirname(debug_backtrace()[0]['file']),
-			version: $version
 		);
 
 		$name = $plugin->name();
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -715,7 +715,8 @@ trait AppPlugins
 		string $name,
 		array $extends = null,
 		array $info = [],
-		string|null $root = null
+		string|null $root = null,
+		string|null $version = null
 	): PLugin|null {
 		if ($extends === null) {
 			return static::$plugins[$name] ?? null;
@@ -725,7 +726,9 @@ trait AppPlugins
 			name:    $name,
 			extends: $extends,
 			info:    $info,
-			root:    $root ?? dirname(debug_backtrace()[0]['file']),
+			// TODO: Remove fallback to $extends in v5
+			root:    $root ?? $extends['root'] ?? dirname(debug_backtrace()[0]['file']),
+			version: $version
 		);
 
 		$name = $plugin->name();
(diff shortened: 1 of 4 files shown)


bastianallgeier:
A couple first thoughts and steps: 

- I think we should not trigger deprecation warnings in v4.2 This would feel a lot like a breaking change. I already removed them. It would be better to keep track of this properly and remove it in v5 as a breaking change
- I've refactored the plugin tests and also fixed an issue after the new order how to recognise the version
- I have no idea so far why the assets test fails and how this could be related.

distantnative:
@bastianallgeier we should keep the deprecation messages, but we can silence them first via https://github.com/getkirby/kirby/blob/main/src/Cms/Helpers.php#L31

bastianallgeier:
@distantnative that's clever. I have to say that I never fully explored the options of the Helpers class. I just brought them back.

distantnative:
@bastianallgeier I moved the `version` from its own named argument into the new `info` named argument. Felt more consistent as otherwise we might would want to add all info entries as their own named arguments. Which could be an option as well.

lukasbestle:
One argument for a separate `version:` named argument could be that is easier to match with a regex like `/^(\s*version:\s*["'])([^"']+)(["'])/m`.

Inside an array, it could use `"` or `'` as keys, value indentation etc., which makes it more complex. But it wouldn't be impossible either. So fine with both.

distantnative:
Yea I really get that. I opted against it in the PR because I think it makes the whole setup a bit easier/consistent if not one of the info parts get it's extra treatment than others. But open to change this.

lukasbestle:
Yeah, makes sense. If we want to make it possible to register plugins without a `composer.json` at all, that definitely fits better.

bastianallgeier:
TBH, I would vote for separate named arguments if they are really relevant. In addition to that there could still be a meta or info pool. But the arrays of schema-less info have gotten us into so many difficult situations, so I would avoid them wherever possible.

distantnative:
So all info values from composer json also as named arguments?

bastianallgeier:
I would really keep it simple for now and solve the version problem with this PR and then maybe add more later or think some more about an optional info array

distantnative:
Ok. The optional info array we need anyways as this has been supported so far already, just as subentry of the `extends` array. So it's more about whether to have all the info in that array or have version (and/or others) also as another additional named argument on their own (but still with necessary support to also allow it in the info array).

distantnative:
@lukasbestle @bastianallgeier now with additional `version` arg

afbora:
I'm thinking loud. Why don't we use the `Plugin` class constructor like other model class:

````php
class Plugin
{
  public function __construct(array $props) {
    // check and set props
  }
}
````
This covers props to be added later/future and ensures consistency with other models.

**Comment by lukasbestle** [comment]
@afbora The `array $props` pattern is pretty old and we used it before named arguments were a thing. Those are much more solid because of their type guarantees and also because they allow IDE hints, which a plain array can't without a lot of boilerplate comment blocks.

File: src/Cms/AppPlugins.php [hunk 1]
```diff
@@ -713,17 +713,22 @@ public function nativeComponent(string $component): Closure|false
...
 			return static::$plugins[$name] ?? null;
 		}
 
-		// get the correct root for the plugin
-		$extends['root'] ??= dirname(debug_backtrace()[0]['file']);
+		$plugin = new Plugin(
+			name:    $name,
+			extends: $extends,
+			info:    $info,
+			root:    $root ?? dirname(debug_backtrace()[0]['file']),
```
lukasbestle: We need to still support the old `$extends` way, otherwise it would be an immediate breaking change:

```suggestion
			// TODO: Remove fallback to $extends in v5
			root:    $root ?? $extends['root'] ?? dirname(debug_backtrace()[0]['file']),
```

src/Cms/Plugin.php
```diff
@@ -40,16 +34,44 @@ class Plugin
...
+		// TODO: Remove in v5
+		if ($root = $extends['root'] ?? null) {
+			Helpers::deprecated('Plugin "' . $name . '": Passing the `root` inside the `extends` array has been deprecated. Pass it directly as named argument `root`.', 'plugin-extends-root');
+			$this->root ??= $root;
+			unset($this->extends['root']);
+		}
+
+		$this->root ??= dirname(debug_backtrace()[0]['file']);
+
+		// TODO: Remove info property in v5
```
lukasbestle: I think you meant the fallback, not the property:

```suggestion
		// TODO: Remove in v5
```

src/Cms/AppPlugins.php
```diff
@@ -713,17 +713,22 @@ public function nativeComponent(string $component): Closure|false
 	 */
 	public static function plugin(
 		string $name,
-		array $extends = null
+		array $extends = null,
+		array $info = [],
+		string|null $root = null
```
lukasbestle: The separate `$version` argument makes sense here as well IMO. It's true we need the general `$info` array anyway, but IMO every info prop Kirby needs should be provided with a typed named argument like Bastian wrote. With a separate argument, we can advertise the use of the named argument in the docs from the start.

distantnative: As I wrote, what's every info prop? Are we adding named arguments for every possible composer entry? Which is why Bastian suggested to still keep it simple with just the version named argument.

lukasbestle: Exactly, but it is only in the `Plugin` constructor, not the `App::plugin()` method at the moment. Therefore my suggestion to add it there as well.

What I mean with "every info prop Kirby needs": Every info prop that is accessed by Kirby directly. This could also be the info for the plugin table in the system view, but there I think it's less of a problem to have it untyped as it's just for display.

lukasbestle (approved):

PR #6340, opened 2024-03-09
Pagination for the file browser component

### LINKED ISSUE #6288: Kirby 4 New Link-Field not showing all files from a page
## Description
I have a download-archive with over 200 pdf-files for a website. The goal was to have a single page, listing all pdfs from the whole system. But those files are also used in other subpages, where i want to link them inside a button.
When using the new Linkfield it only shows me the first 100 files of the page. There is no pagination or a search function.

**Expected behavior**  
The Files-List in the link-field should have a pagination, search function or just showing all files

**Screenshots**  
files
files2

## Your setup
**Kirby Version**  
4.1.0

afbora:
## This PR …

I'm not sure how logical it is to add `limit` prop only for file-browser but I'm waiting for your reviews and thoughts.

### Fixes
- Kirby 4 New Link-Field not showing all files from a page #6288

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/fields/link.php
+++ b/config/fields/link.php
@@ -13,6 +13,13 @@ return [
 		'placeholder' => null,
 
 		/**
+		 * Sets the limit to listing files for a page
+		 */
+		'limit' => function (int $limit = null) {
+			return $limit;
+		},
+
+		/**
 		 * @values 'anchor', 'url, 'page, 'file', 'email', 'tel', 'custom'
 		 */
 		'options' => function (array|null $options = null): array {
--- a/config/fields/link.php
+++ b/config/fields/link.php
@@ -13,13 +13,6 @@ return [
 		'placeholder' => null,
 
 		/**
-		 * Sets the limit to listing files for a page
-		 */
-		'limit' => function (int $limit = null) {
-			return $limit;
-		},
-
-		/**
 		 * @values 'anchor', 'url, 'page, 'file', 'email', 'tel', 'custom'
 		 */
 		'options' => function (array|null $options = null): array {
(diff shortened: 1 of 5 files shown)


distantnative:
To be discussed: I would not add it as new blueprint option but just choose a sensible default.

afbora:
I'm fine with the default limit 👍

frankziski:
maybe this helps for clarification:
the bug mentioned is not about the files browser when you have a blueprint field type "files". Its about the field type "link" when you change to link to a file. Then there is no pagination at all and you only see 100 files, regardless on how many files are on the selected page. So it would be great to add the pagination here like it is in the file browser.

afbora:
@frankziski Yes, this PR fixes the issue you mentioned.

bastianallgeier:
@afbora I've adjusted the styling a bit and added a lab example. Could you double-check if everything is still as you intended?

distantnative:
@bastianallgeier ehat do you think of removing the new option and having a fixed limit instead?

And there was a console.log added now.

bastianallgeier:
@distantnative ah, I missed the option discussion. I agree that we should not add another option to the blueprint, but I would keep the limit prop for the file browser component

The console.log is intended. I think it's easier to debug that way what is being emitted by the select event without adding a huge chunk of json to the lab output.

bastianallgeier:
I've cleaned up the limit prop from the link field. It's now only available for the file browser component.

afbora:
Thank you @bastianallgeier Now better with your touches 👍 

What do you think reducing the default limit to 25 or 50? Really long list with the limit 100.

bastianallgeier:
@afbora 50 sounds good to me.

afbora:
Now more looks well and fast loading when paginate 👍 Now ready to merge.

bastianallgeier (approved):

PR #6345, opened 2024-03-14
Inherit `style` attribute

distantnative:
Everywhere where we set `inheritAttrs: false`, we need to ensure to manually inherit `class` and `style` attributes then.


distantnative:
Don't merge while we explore option to remove `inheritAttrs: false` completely

distantnative:
We will keep `inheritAttrs: false`, so this can be reviewed again.

panel/src/components/Forms/Blocks/Block.vue
```diff
@@ -13,8 +14,8 @@
 		:data-last-selected="isLastSelected"
 		:data-selected="isSelected"
 		:data-translate="fieldset.translate"
-		class="k-block-container"
 		:tabindex="isDisabled ? null : 0"
+		:style="$attrs.style"
```
bastianallgeier:
```suggestion
		:style="$attrs.style"
		:tabindex="isDisabled ? null : 0"
```

bastianallgeier (changes requested):
Just a small change.

bastianallgeier (approved):

bastianallgeier:
I think we need to move forward after v5 by actively describing, which components should be able to set the style attribute at all. It's a great comporomise like this and I'm happy that we can get rid of inheritAttributes, but it does not make sense either to simply allow it everywhere. I think that's the strength of this approach. We have more control and transparency of what is actually possible.

distantnative:
@bastianallgeier just for clarification, we can't get rid of inheritAttrs

bastianallgeier:
I put it the wrong way. I mean we get rid of inheritAttributes: true and avoid injecting unwanted non-valid attributes.

PR #6354, opened 2024-03-23
Include drafts in finding pages key recursively

### LINKED ISSUE #6339: $site->findPageOrDraft() does not work for nested drafts
## Description

- page-1 -> listed
  - page-2 > draft
     - page-3 > draft
-> will fail

- page-1 -> draft
  - page-2 > listed
     - page-3 > listed
-> will work

- page-1 -> listed
  - page-2 > draft
     - page-3 > listed
-> will fail

Using `$site->findPageOrDraft("page-1/page-2/page-3")`

**Expected behavior**  
I expect the method to recursively find draft pages as well.

## Your setup
**Kirby Version**  4.1.2

distantnative:
## This PR …

Let's think hard if this could cause any regressions or is a breaking change (not just fixing a bug).

### Fixes
- Pages collection now include drafts when trying to recursively find a key,
#6339 

## Ready?

### For review team

### DIFF
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -256,7 +256,7 @@ class Pages extends Collection
 		$query      = $startAt;
 
 		foreach ($path as $key) {
-			$collection = $item?->children() ?? $this;
+			$collection = $item?->childrenAndDrafts() ?? $this;
 			$query      = ltrim($query . '/' . $key, '/');
 			$item       = $collection->get($query) ?? null;
 
--- a/tests/Cms/Pages/PagesTest.php
+++ b/tests/Cms/Pages/PagesTest.php
@@ -630,6 +630,50 @@ class PagesTest extends TestCase
 		$this->assertIsPage('zzz/yyy', $pages->find('zzz/yyy'));
 	}
 
+	public function testFindChildrenDrafts()
+	{
+		$app = new App([
+			'roots' => [
+				'index' => '/dev/null'
+			],
+			'site' => [
+				'children' => [
+					[
+						'slug' => 'grand',
+						'children' => [
+							[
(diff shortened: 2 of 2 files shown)


lukasbestle (commented):
Hm... There would suddenly no longer be a difference between `HasChildren::find()` and `HasChildren::findPageOrDraft()` anymore, right? I feel like there could be use cases where one would want to explicitly only find published pages. So this would be a breaking change and also one that wouldn't be intended I think.

Maybe we need to introduce a separate recursive finding mode that uses `childrenAndDrafts()`. That could then be specifically used by `HasChildren::findPageOrDraft()`, also allowing us to get rid of the separate search in both collections.

One way could be to have a new marker prop for `Pages` that keeps track of whether the collection is `children`, `drafts` or `childrenAndDrafts`. The same collection type would then be used in the `findByKeyRecursive()` method.

distantnative:
@lukasbestle Revisiting this, I am wondering if we need to just adapt `HasChildren::find()` to filter the result of drafts.

Because in general, I think this PR would be the more correct behavior: Drafts and pages can only have unique IDs. So when trying to access a page with a key-path, it's currently weird that we completely ignore a page (draft) that actually matches the specified path. And this is just because of its page status. This creates a greater separation between pages and drafts than there should be. Drafts shouldn't be automatically included when returning collections, e.g. to avoid putting them in templates while they're still drafts. But when we explicitly look them up with a key, it's weird that we ignore them so much and say "oh, well, you didn't explicitly ask for drafts" (as if drafts aren't pages, too). 😅

I'm totally happy to move this to v5 as breaking change. But I think it would be good to change this.

distantnative:
Closing this for now as no progress has been made.

PR #6355, opened 2024-03-23
`(image:)` escape alt text for smartypants

### LINKED ISSUE #6344: Image tag breaks with smartypants enabled
## Description
I like using the [smartypants](https://getkirby.com/docs/reference/system/options/smartypants) option to have typographically correct quotes. This can however break the usage of the `(image: )` kirbytag.

**Output**  

```html
<img alt="Test with "quotes"" src="(…)/test.png">
```

**Output with more/less quotes** (now it gets weird)
```html
<img alt=&#8221;Test with &#8220;quotes&#8221; src=&#8221;(…)/test.png&#8221;>
```

**Expected Output**

```html
<img alt="Test with &quot;quotes&quot;" src="(…)test.png">
```

## To reproduce

1. Upload an image
2. Use the `(image: )` tag to insert this image in a textarea
3. Add an alt text in the file content or image tag (it doesn't matter)
4. Turn the textarea into kirbytext with `->kirbytext()`
5. Enable smartypants in the config
6. See error

## Your setup
**Kirby Version**  
4.1.2

distantnative:
### Fixes
- `(image: )` KirbyTag: escape `alt` text for proper value, e.g. when smartypants option is activated
#6344

## Ready?

### For review team

### DIFF
--- a/config/tags.php
+++ b/config/tags.php
@@ -154,7 +154,7 @@ return [
 				'height' => $tag->height,
 				'class'  => $tag->imgclass,
 				'title'  => $tag->title,
-				'alt'    => $tag->alt ?? ''
+				'alt'    => Str::esc($tag->alt ?? '')
 			]);
 
 			if ($tag->kirby()->option('kirbytext.image.figure', true) === false) {


lukasbestle (commented):
Very much looks like a hack to me TBH. `Html::img()` (or to be more specific, `Xml::attr()`) already uses `Xml::encode()` to make the value string safe. We should investigate why that doesn't bring the intended effect.

distantnative:
@lukasbestle You're right, I think the problem starts because smartypants is applied after resolving all kirbytags https://github.com/getkirby/kirby/blob/main/src/Cms/App.php#L872-L874 - so actually applied on the HTML.

The SmartyPants library seems like it should actually respect HTML code and not create problems. But my guess is that something there fails.

distantnative:
Closing this for now as the proposed solution isn't quite right.

PR #6356, opened 2024-03-23
Table layout: fix width for image + flag columns

### LINKED ISSUE #6228: Incorrect display of preview images (icons) with 'layout: table' responsive window size.
**Kirby 4.1.0**

**Scenario: Listing of pages as table layout**
When changing the browser window width, the preview images are displayed too wide. 
Triggered by `width: auto!important` of this css-class:

```
@container (max-width: 40rem)
style.min.css:1
.k-table :where(th,td):not(.k-table-index-column):not(.k-table-options-column) {
    width: auto!important;
}
```

https://github.com/getkirby/kirby/assets/140849227/965d7d59-12da-456d-bdbf-47a9ec6c41c6

My blueprint:
```
  changedpages:
    label: "Chronologische Ansicht der zuletzt bearbeiteten Seiten:"
    type: pages
    query: site.index(true).sortBy('modified', 'desc')
    create: false
    layout: table
    limit: 15
    search: true
    columns:
      parentpage:
        label: "Übergeordnete Seite"
        value: "{{ page.parent ? page.parent.title : '–' }}"
        width: 1/6
      day:
        label: Datum
        value: "{{ page.modified('Y-m-d') }}"
        width: 1/6
      time:
        label: Uhrzeit
        value: "{{ page.modified('H:i:s') }}"
        after: " Uhr"
        width: 1/6
    image:
      back: "#000000"
      color: "#ffffff"
      query: false
```

distantnative:
## This PR …

New `static` option for columns that avoids applying auto width to it on narrower containers.

To be honest, I am not in love with the name yet.

### Fixes
- #6228

### For review team

### DIFF
--- a/config/sections/mixins/layout.php
+++ b/config/sections/mixins/layout.php
@@ -40,7 +40,8 @@ return [
 					'label'  => ' ',
 					'mobile' => true,
 					'type'   => 'image',
-					'width'  => 'var(--table-row-height)'
+					'width'  => 'var(--table-row-height)',
+					'static' => true
 				];
 			}
 
@@ -88,6 +89,7 @@ return [
 					'mobile' => true,
 					'type'   => 'flag',
 					'width'  => 'var(--table-row-height)',
+					'static' => true
 				];
 			}
 
--- a/panel/src/components/Layout/Table.vue
+++ b/panel/src/components/Layout/Table.vue
@@ -17,6 +17,7 @@
 						:key="columnIndex + '-header'"
 						:data-align="column.align"
 						:data-mobile="column.mobile"
+						:data-static="column.static"
 						:style="{ width: width(column.width) }"
 						class="k-table-column"
 						@click="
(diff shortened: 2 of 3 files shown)


bastianallgeier (commented):
I was wondering if we should add the column name to each td (e.g. as data attribute instead.) Then we could use that as a styling hook to fix the width of the image column. This would have the advantage of more CSS control but without introducing yet another option.

distantnative:
I think it boils down to whether we want only to support/fix it for our default columns or support it also for e.g. custom image columns or similar.

bastianallgeier:
Yep, I think the css classes would be a good first option to fix it. But the main question remains: how do we create a proper system to work with responsive tables. I feel like the mobile option that we are offering right now is pretty limited and has lots of potential problems. That's why I'm not sure of the future here to be fully honest.

distantnative:
@bastianallgeier that would be then like this: https://github.com/getkirby/kirby/pull/6368

bastianallgeier:
I like it! I would suggest to move this into a lab branch and talk some more about options.

distantnative:
Yes, will close it for now and keep it locally.

PR #6359, opened 2024-03-23
fix not creating PDF as different format

### LINKED ISSUE #6268: File blueprint: error when create/format option is set and document is uploaded
## Description
When a file blueprint has the format option set so that images get converted to another format, uploading a PDF results in an error `imagecreatefromstring(): Data is not in a recognized format `. The uploaded file is corrupt due to the conversion process.

**Expected behavior**  
Images should be converted to the format defined in the blueprint. Any other file should be uploaded as is.

**Screenshots**  
image
image

## To reproduce
1. Create a new file blueprint or use default file blueprint.
2. Add the following properties to the blueprint: 
```
create:
  format: webp
```
3. Upload a PDF using that blueprint

## Your setup
**Kirby 4.0.1 up to 4.1**  

**Console output**  
image
First error while uploading. Second error while trying to get a thumbnail for the uploaded PDF that doesn't exist.

 - Device: Macbook
 - OS: macOS
 - Browser: Firefox
 - Version: 122.01

**Additional context**

distantnative:
## This PR …

If we change the extension first, `$file->isResizable()` will test for the new/target format, not the original one (e.g. PDF).

### Fixes
- File `create` `format` option is not applied to non-images anymore 
#6268

## Ready?

### For review team

### DIFF
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -139,8 +139,8 @@ trait FileActions
 
 			// rename and/or resize the file if configured by new blueprint
 			$create = $file->blueprint()->create();
+			$file = $file->manipulate($create);
 			$file = $file->changeExtension($file, $create['format'] ?? null);
-			$file->manipulate($create);
 
 			return $file;
 		});
@@ -266,7 +266,6 @@ trait FileActions
 		// we need to already rename it so that the correct file rules
 		// are applied
 		$create = $file->blueprint()->create();
-		$file = $file->changeExtension($file, $create['format'] ?? null);
 
 		// run the hook
 		$arguments = compact('file', 'upload');
@@ -284,6 +283,7 @@ trait FileActions
 
 			// resize the file on upload if configured
 			$file = $file->manipulate($create);
+			$file = $file->changeExtension($file, $create['format'] ?? null);
 
 			// store the content if necessary
 			// (always create files in the default language)
@@ -384,8 +384,8 @@ trait FileActions
 
 			// apply the resizing/crop options from the blueprint
 			$create = $file->blueprint()->create();
-			$file   = $file->changeExtension($file, $create['format'] ?? null);
 			$file   = $file->manipulate($create);
+			$file   = $file->changeExtension($file, $create['format'] ?? null);
 
 			// return a fresh clone
 			return $file->clone();


lukasbestle (commented):
Should it be intended behavior that non-images are allowed through? TBH I would rather expect that every file uploaded with that file blueprint will have the configured target format. This would mean that any upload of a file that cannot be converted would fail.

distantnative:
@lukasbestle I think this has been the intended behavior: https://github.com/getkirby/kirby/blob/main/src/Cms/FileActions.php#L327 - only that this failed as the extension already had been changed and thus comparing by the extension would create false results for the check. I would keep the logic as is and only fix the bug.

lukasbestle:
That link you point to concerns resizing though, not file format conversion. I'd say that it's expected that a file that cannot be resized won't be resized. But if I say that I want format X when format Y is uploaded, isn't it unexpected to get format Z instead? TBH I never thought about that before this issue.

distantnative:
@lukasbestle you sure? I thought the thumbs component also takes care of the format conversion

distantnative:
It's only that the check back then was named `isResizable()` - but it's actually more a check for "is darkroom processable" as it also covers the format or alterations like b/w etc.

lukasbestle:
Ah, right. To me that looks even more like we simply didn't consider this case before.

First we had resizing (which of course doesn't apply to PDFs or other file types). Now we have the option to convert the file to a different format entirely (which can lead to the expectation that all files will have the configured format).

Maybe let's get the author of the issue @JerikoOne involved. You wrote in the expected behavior section that other files should be uploaded as is. I'm curious: Is it really expected for you in your use case to have a single file template that can have files with completely different types?

distantnative:
I think this is something we should leave in the hand of the developers. Kirby is best at where it allows some flexibility like that instead of being too restrictive that people have to jump around hurdles if they still want to allow that.

So I would really want to merge this PR to fix the existing bug and not dive too much into the general debate whether we should reconsider the mechanism in general.

lukasbestle:
You are right, makes sense. If we'd block it, there would be no way for our users to change this behavior. So I agree, let's fix the bug.

I think we should document this though with an info block. If the format should be enforced, the file type needs to be restricted separately in the blueprint.

bastianallgeier (approved):
LGTM

PR #6361, opened 2024-03-23
Object field: no inline edit when field disabled

### LINKED ISSUE #6346: Disabled toggle field in object preview can still be changed
## Description
When you have a toggle field in an object field with `disabled: true`, you can still change it from the object preview.

**Expected behavior**  
That this field would've been disabled there too.

## Your setup
**Kirby Version**  
4.1.1

 - Device: Macbook Pro
 - OS: macOS
 - Browser: N/A

distantnative:
### Fixes
- Object field: inline previews are not editable anymore when the subfield is disabled
#6346

### For review team

### DIFF
--- a/panel/lab/components/fieldpreviews/toggle/index.vue
+++ b/panel/lab/components/fieldpreviews/toggle/index.vue
@@ -27,6 +27,14 @@
 				<!-- @code-end -->
 			</k-lab-table-cell>
 		</k-lab-example>
+
+		<k-lab-example label="Disabled">
+			<k-lab-table-cell>
+				<!-- @code -->
+				<k-toggle-field-preview :field="{ disabled: true }" />
+				<!-- @code-end -->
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):
I've added a lab example and moved the isEditable computed prop to the toggle component. We will need to find a consistent solution for all field previews and I think it's better to keep the decision more open here.

distantnative:
@bastianallgeier Fine with it. Although I added it to the mixing cause I wanted to lay the groundwork for something general. And I think if any field is disabled, its preview should not be editable.

bastianallgeier:
@distantnative yes, absolutely! But I wonder how we want to call it and I think we need to define this more properly first before we move forward with all previews. Do we introduce an additional disabled prop? Shouldn't it be isDisabled? Questions like that.

PR #6364, opened 2024-03-24
Allow to prefill the login form

lukasbestle:
Main use case for me is to make it easier to log in to the Demokit. We could also use it in the Sandbox.

---

### Enhancements

- The login form can now be prefilled for testing use cases with the new `value` prop of the `k-login-view` component

### Breaking changes

None

## Docs

Probably too obscure to document

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Login.vue
+++ b/panel/src/components/Forms/Login.vue
@@ -45,7 +45,8 @@
 <script>
 export default {
 	props: {
-		methods: Array
+		methods: Array,
+		value: Object
 	},
 	emits: ["error"],
 	data() {
@@ -53,8 +54,8 @@ export default {
 			currentForm: null,
 			isLoading: false,
 			user: {
-				email: "",
-				password: "",
+				email: this.value.email || "",
+				password: this.value.password || "",
 				remember: false
 			}
 		};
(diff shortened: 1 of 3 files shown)


distantnative:
I'd suggest to just calling them email and password.

And let's avoid v-bind=$propd - that'll cause issues down the road with Vue 3. already looking into how to reduce this in our existing code (so totally get why you added it as we currently use it a lot).

lukasbestle:
Got rid of `v-bind`.

Naming the props just `email` and `password` would be super confusing IMO because the data is also called `email` and `password`. "Prefilled" makes clear that it's just the initial value.

distantnative:
The data object is only internally and is `user.email` and `user.password`. 

Alternatively, the prop would be `value` with an object as well to be more in line with our other components.

lukasbestle:
Fine for me. Went with a `value` object so we can also prefill the code if needed.

distantnative (dismissed):
@lukasbestle looks good to me, added some inline docs for the props when we're already touching this. Also switched to inherit props to make them more consistent - we're trying to do it like that in more places recently.

lukasbestle:
Looks good, thanks for the changes.

distantnative (approved):

PR #6365, opened 2024-03-24
Login lab examples

distantnative:
- Added lab examples for `k-login` and `k-login-code`
- Clean up of components code

### DIFF
--- a/panel/src/components/Forms/Login.vue
+++ b/panel/src/components/Forms/Login.vue
@@ -1,32 +1,31 @@
 <template>
 	<form class="k-login-form" @submit.prevent="login">
-		<div class="k-login-fields">
-			<button
-				v-if="canToggle === true"
-				class="k-login-toggler"
-				type="button"
-				@click="toggleForm"
-			>
-				{{ toggleText }}
-			</button>
+		<button
+			v-if="canToggle === true"
+			class="k-login-toggler"
+			type="button"
+			@click="toggle"
+		>
+			{{ toggleText }}
+		</button>
 
-			<k-fieldset
-				ref="fieldset"
-				:novalidate="true"
-				:fields="fields"
-				:value="user"
-				@input="user = $event"
+		<k-fieldset
+			ref="fieldset"
+			:novalidate="true"
+			:fields="fields"
+			:value="user"
+			@input="user = $event"
+		/>
+
+		<footer class="k-login-buttons">
+			<k-checkbox-input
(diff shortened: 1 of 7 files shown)


panel/src/components/Forms/Login.vue
```diff
@@ -1,32 +1,31 @@
 <template>
 	<form class="k-login-form" @submit.prevent="login">
-		<div class="k-login-fields">
```
bastianallgeier: I guess this would technically count as a breaking change for those who have custom styles for their login

distantnative: 👍

panel/src/components/Forms/LoginCode.vue
```diff
@@ -37,9 +37,9 @@
 				variant="filled"
 			>
 				{{ $t("login" + (mode === "password-reset" ? ".reset" : "")) }}
-				<template v-if="isLoadingLogin"> … </template>
+				<template v-if="isLoading"> … </template>
```
bastianallgeier: I think it would be nice to have the same submitText method here to keep it in line with the Login.vue component and simplify the template.

panel/src/components/Forms/Login.vue
```diff
@@ -127,28 +147,30 @@ export default {
...
-				return "email-password";
-			}
-		},
 		async login() {
 			this.$emit("error", null);
 			this.isLoading = true;
 
 			// clear field data that is not needed for login
-			let user = Object.assign({}, this.user);
+			const user = Object.assign({}, this.user);
```
bastianallgeier: I'm not entirely sure about it, but I think this should work as well and would feel closer to what we normally do …

```suggestion
			const user = {
			  ...this.user
			};
```

panel/src/components/Forms/LoginCode.vue
```diff
@@ -17,15 +17,15 @@
...
 
-		<div class="k-login-buttons">
+		<footer class="k-login-buttons">
 			<k-button
 				class="k-login-button k-login-back-button"
 				icon="angle-left"
 				size="lg"
 				variant="filled"
-				@click="back"
+				@click="$go('/logout')"
```
bastianallgeier: I think we can do …

```suggestion
				link="/logout"
```

bastianallgeier (approved):
Nice! I would change the release to 4.3 just to be a bit safer.

PR #6369, opened 2024-03-25
Fix deprecation warning in router

### LINKED ISSUE #5895: Deprecation Warning: urldecode()
## Description
I found this by mistake when I've added two slashes or more after localhost:8000 and followed by a page (ex. localhost:8000//panel)

When two slashes are added I get this deprecated warning:
`Deprecated: urldecode(): Passing null to parameter #1 ($string) of type string is deprecated in C:\Users\bogda\OneDrive\Projects\htdocs\double-check\vendor\getkirby\cms\router.php on line 3`

image

When more than two slashes are added I get this deprecated notice:
`Deprecated: Automatic conversion of false to array is deprecated in C:\Users\bogda\OneDrive\Projects\htdocs\double-check\vendor\getkirby\cms\src\Http\Uri.php on line 114`

image

**Expected behavior**  
I gues it should not show any errors or correct the url maybe??

## To reproduce

1. Go to localhost:port and place two or more slashes after it followed by panel or anything else
2. You should see a deprecated warning with two slashes and a deprecated notice when more than two slashes placed

## Your setup
**Kirby Version**  
3.9.7

**Console output**  
nothing relevant

 - OS: Windows 11
 - Browser: Edge
 - Version: 118

**Additional context**  
I've tested on other projects and the errors are to be reproduced only on local dev and not on production.

Edit: I've tested this on a starterkit with v4 and it still happens

bastianallgeier:
### Fixes

- https://github.com/getkirby/kirby/issues/5895

### For review team

### DIFF
--- a/router.php
+++ b/router.php
@@ -1,7 +1,7 @@
 <?php
 
 $uri = urldecode(
-	parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH)
+	parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH) ?? '/'
 );
 
 // Emulate Apache's `mod_rewrite` functionality
--- a/router.php
+++ b/router.php
@@ -1,11 +1,10 @@
 <?php
 
-$uri = urldecode(
-	parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH) ?? '/'
-);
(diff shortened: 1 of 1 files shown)


lukasbestle:
Shouldn't we rather trim off additional slashes at the beginning to make routing work? Something along the lines of `'/' . ltrim($path, '/')`.

bastianallgeier:
@lukasbestle I committed another try with the full-url trick. I think this should be more stable.

lukasbestle (dismissed):
Brilliant! Added a comment so we later remember what it does and why.

PR #6375, opened 2024-03-30
Require PHP 8.2

distantnative:
## This PR …

I think it's quite settled that we will release Kirby 5 somewhere around the end of future support of PHP 8.1 and we can consider raising the PHP requirement.

Main question: Do we think webhosters will keep up the tempo or could this be a problem?

Opening this PR already as I think, it's good to merge it early - only once we are certain we want to do it - so we can develop all of v5 already with the new baseline of PHP 8.2.

### Breaking changes
- Kirby requires at least PHP 8.2

## Ready?

### For review team

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -259,7 +259,7 @@ jobs:
         with:
           php-version: ${{ env.php }}
           coverage: none
-          tools: php-cs-fixer:3.13.0
+          tools: php-cs-fixer:3.52.1
 
       - name: Cache analysis data
         id: finishPrepare
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -57,7 +57,7 @@ jobs:
     timeout-minutes: 5
     strategy:
       matrix:
-        php: [ "8.1", "8.2", "8.3" ]
+        php: ["8.2", "8.3"]
     env:
       extensions: mbstring, ctype, curl, gd, apcu, memcached
       ini: apc.enabled=1, apc.enable_cli=1, pcov.directory=., "pcov.exclude=\"~(vendor|tests)~\""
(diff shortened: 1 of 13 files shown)


lukasbestle (commented):
Dropping 8.0 in v4 didn't seem to be a problem, so I think the same can be said about 8.1 in v5. Keeping the yearly schedule seems sensible to me.

Could you take a look at the coding style changes? I think something broke again with the indentation. We had that before and I don't remember what fixed it.

distantnative:
@lukasbestle The indentation is basically dependent on the version of php-cs-fixer. But with 8.2 as baseline we have to move to a higher version which brings this new formatting.

(When we had this before we just moved back to the older version)

distantnative:
@lukasbestle managed to find a config setting that preserves our old comment before elseif indentation. But haven't found a way yet to have the multiline @param descriptions render right.

lukasbestle:
That's indeed really weird. They even have a unit test for this that passes:
https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/9782155324211694dd6593e2faf7d4458440e45b/tests/Fixer/Phpdoc/PhpdocAlignFixerTest.php#L200
But even if I disable all custom rules, I still get the wrong output.

distantnative:
If I disable all, also the `@PSR12` ruleset, it doesn't get "fixed". But har dot know what in the rules causes it.

distantnative:
@lukasbestle you think we could move it still then with those few adapted comments?

lukasbestle (approved):
Sure. It's only a cosmetic issue. Maybe it will fix itself at some point.

lukasbestle:
OMG, it seems to have been caused by the missing param types. Without the type the `@param` rule is not valid, so the PHP-CS-Fixer devs probably didn't test that.

distantnative:
Oh man, great find!

PR #6377, opened 2024-04-01
Extend PHP code styling rules

distantnative:
### Housekeeping
- Extended rule list for `php-cs-fixer`

### For review team


lukasbestle (commented):
I agree on all changes to the ruleset, they improve a lot of parts in the code. However I'm worried about two things:

1. In some places, PHP-CS-Fixer went rogue and butchered comments or formatting (e.g. the removed superfluous PHPDoc tags include ones where we typoed the param name; also the QrCode class used manual formatting that is now butchered). We'd need to go over them manually before merging this.
2. We already have some changes in the v5 branch and #6391 and I'm afraid there will be a lot of merge conflicts if we merge this to 4.3.0 and then try to merge it back into v5. Maybe it would be safer to only copy over the ruleset changes to a v5 PR and then run PHP-CS-Fixer there again with a commit for each rule so we can properly review.

PS: We could also remove the commented out `phpdoc_add_missing_param_annotation` rule as is no longer necessary with `no_superfluous_phpdoc_tags`. Same with `class_keyword_remove` as it is deprecated anyway.

distantnative:
https://github.com/getkirby/kirby/pull/6398

PR #6379, opened 2024-04-03
Fix table column regressions

distantnative:
- [ ] Double check if the current state is actually intended, see issue reported for docs: https://github.com/getkirby/getkirby.com/issues/2291

Two fixes for 4.2.0 regressions:

- properly allowing to override column values for default ids (e.g. `text`) when explicit `value` option is set
- also apply `data-column-id` to td in table body row


bastianallgeier (approved):

distantnative:
@bastianallgeier sandbox examples look good and functional to me. Also showing that we could close https://github.com/getkirby/getkirby.com/issues/2291 then

PR #6382, opened 2024-04-06
Writer: support for node and mark views

distantnative:
## This PR …

Adds new `view` extensions to writer mark and nodes.
https://feedback.getkirby.com/633

```js
writerMarks: {
    highlight: {
        get button() {
            return {
                icon: "palette",
                label: window.panel.$t("color"),
            };
        },

        commands() {
            return () => this.toggle();
        },

        get name() {
            return "highlight";
        },

        get schema() {
            return {
                parseDOM: [{ tag: "mark" }],
                toDOM: () => ["mark", 0],
            };
        },

        view(mark, view, inline) {
            return new HighlightView(mark, view, inline);
        },
    },
},
```

```js
class HighlightView {
    constructor(mark, view, inline) {
        const el = document.createElement("mark");
        const before = document.createElement("span");
        before.textContent = "→ ";
        const after = document.createElement("span");
        after.textContent = " ←";

        this.contentDOM = document.createElement("span");

        el.appendChild(before);
        el.appendChild(this.contentDOM);
        el.appendChild(after);
        this.dom = el;
    }
}
```
Screenshot_2024-04-06_at_13 56 52

## Docs

I would suggest to leave this undocumented for now as it is quite the advanced feature of ProseMirror and we are not sure yet if there aren't additional hurdles in the way. I would want to add these extensions so that experienced developers can start working on implementations and explore if there are additional hurdles, without steering less experienced developers in this difficult (and itself not well documented) ProseMirror context

### For review team

### DIFF
--- a/panel/public/js/plugins.js
+++ b/panel/public/js/plugins.js
@@ -9,9 +9,7 @@ window.panel.plugins = {
 	use: [],
 	views: {},
 	writerMarks: {},
-	writerMarkViews: {},
-	writerNodes: {},
-	writerNodeViews: {}
+	writerNodes: {}
 };
 
 window.panel.plugin = function (plugin, extensions) {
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):
I've tested it with new examples in the lab and it works great. I just went a bit nuts because it didn't work at first until I realised that your example code was missing the commands method :) I've added it to make sure we don't stumble over it again. 

I like the idea to keep this undocumented for now.

PR #6384, opened 2024-04-06
Writer toolbar: support inline node buttons

distantnative:
## This PR …

TBD: Is it too confusing that in the blueprint option, it would then have to be listed under `marks` instead of `nodes` even though it's technically a custom node?

- [ ] requires bode to be added to both, marks and nodes option, currently 

### Enhancement
- Custom writer nodes: button can now set an `inline: true` option to not be listed in the toolbar dropdown but inline alongside marks

## Docs

```js
window.panel.plugin("your/plugin", {
  writerNodes: {
    placeholder: {
      get button() {
        return {
          icon: "wand",
          inline: true
        }
      },
 ```

### For review team

### DIFF
--- a/panel/src/components/Forms/Writer/Toolbar.vue
+++ b/panel/src/components/Forms/Writer/Toolbar.vue
@@ -158,6 +158,15 @@ export default {
 		markButtons() {
 			const available = this.editor.buttons("mark");
 
+			// add node buttons that are meant to be displayed inline
+			const nodes = this.editor.buttons("node");
+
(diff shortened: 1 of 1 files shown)


distantnative:
@bastianallgeier alternatively, we could scrap the inline option again and just make all nodes and marks buttons available for either. And as soon as a mark or node name is mentioned in `nodes` for the dropdown or `marks` for the inline buttons, the mark/node button is shown in that position. No matter if it really is a mark or a node.

PR #6385, opened 2024-04-06
Fix `nodeInputRule`

distantnative:
## This PR …

For the purpose of the hr node, `nodeInputRule` also deleted the previous character before the match (to remove the extra leftover empty line). This behavior doesn't make sense for other nodes, in particular inline nodes. This PR fixes the behavior of `nodeInputRule` and adds the custom behavior directly to the hr node.

### Fixes
- Editor: `nodeInputRule` only overwrites matched selection

### For review team


bastianallgeier (approved):

bastianallgeier:
I've added a hr example to the writer field page blueprint.

PR #6386, opened 2024-04-06
Writer: fix nodes for `inline: true`

distantnative:
### Fixes
- Writer: inline nodes are included when `inline: true` mode is active

### For review team


bastianallgeier (approved):

bastianallgeier:
I've added an inline example to the writer field page blueprint

PR #6391, opened 2024-04-11
Type hinting for collections

distantnative:
## TBD
- [ ] How can we resolve the type templates in our docs?

### Enhancement
- Improved support for IDE autocompletion and type hints for collection items ($pages, $files...)

### Breaking change
- PHP (return) type hints have been added to many collection methods. If you are extending any collection classes, you might need to add the same to your methods.

## Ready?

### For review team

### DIFF
--- a/src/Cms/Collection.php
+++ b/src/Cms/Collection.php
@@ -22,26 +22,24 @@ use Kirby\Uuid\Uuid;
  * @link      https://getkirby.com
  * @copyright Bastian Allgeier
  * @license   https://getkirby.com/license
+ *
+ * @template TValue
+ * @template-extends \Kirby\Toolkit\Collection<TValue>
  */
 class Collection extends BaseCollection
 {
 	use HasMethods;
 
 	/**
-	 * Stores the parent object, which is needed
-	 * in some collections to get the finder methods right.
-	 *
-	 * @var object
+	 * @var \Kirby\Cms\Pagination|null
 	 */
-	protected $parent;
+	protected $pagination;
 
 	/**
-	 * Magic getter function
-	 *
-	 * @param string $key
-	 * @param mixed $arguments
-	 * @return mixed
+	 * Stores the parent object, which is needed
+	 * in some collections to get the finder methods right.
 	 */
+
 	public function __call(string $key, $arguments)
 	{
 		// collection methods
@@ -53,13 +51,14 @@ class Collection extends BaseCollection
 	/**
 	 * Creates a new Collection with the given objects
 	 *
-	 * @param array $objects
-	 * @param object|null $parent
+	 * @param object|null $parent Stores the parent object,
+	 *                            which is needed in some collections
+	 *                            to get the finder methods right
 	 */
-	public function __construct($objects = [], $parent = null)
-	{
-		$this->parent = $parent;
-
+	public function __construct(
+		iterable $objects = [],
+		protected object|null $parent = null
+	) {
 		foreach ($objects as $object) {
 			$this->add($object);
 		}
@@ -72,7 +71,7 @@ class Collection extends BaseCollection
 	 * child classes can override it again to add validation
 	 * and custom behavior depending on the object type
 	 *
-	 * @param object $object
+	 * @param TValue $object
 	 */
 	public function __set(string $id, $object): void
 	{
@@ -84,7 +83,7 @@ class Collection extends BaseCollection
 	 * override from the Toolkit Collection is needed to
 	 * make the CMS collections case-sensitive
 	 */
-	public function __unset($id)
+	public function __unset(string $id)
 	{
 		unset($this->data[$id]);
 	}
@@ -94,9 +93,10 @@ class Collection extends BaseCollection
 	 * an entire second collection to the
 	 * current collection
 	 *
-	 * @param mixed $object
+	 * @param TValue|self|array $object
+	 * @return $this
 	 */
-	public function add($object)
+	public function add($object): static
 	{
 		if ($object instanceof self) {
 			$this->data = [...$this->data, ...$object->data];
@@ -115,12 +115,18 @@ class Collection extends BaseCollection
 	/**
 	 * Appends an element to the data array
 	 *
-	 * @param mixed ...$args
-	 * @param mixed $key Optional collection key, will be determined from the item if not given
-	 * @param mixed $item
-	 * @return \Kirby\Cms\Collection
+	 * ```php
+	 * $collection->append($object);
+	 * $collection->append('key', $object);
+	 * $collection->append([$object1, $object2]);
+	 * ```
+	 *
+	 * If not collection key is passed, it will be determined from the item (id)
+	 *
+	 * @param array|string|TValue ...$args
+	 * @return $this
 	 */
-	public function append(...$args)
+	public function append(...$args): static
 	{
 		if (count($args) === 1) {
 			// try to determine the key from the provided item
@@ -141,7 +147,7 @@ class Collection extends BaseCollection
 	 * Find a single element by an attribute and its value
 	 *
 	 * @param mixed $value
-	 * @return mixed|null
+	 * @return TValue|null
 	 */
 	public function findBy(string $attribute, $value)
 	{
@@ -158,13 +164,15 @@ class Collection extends BaseCollection
 	 * Groups the items by a given field or callback. Returns a collection
 	 * with an item for each group and a collection for each group.
 	 *
-	 * @param string|Closure $field
+	 * @param string|\Closure $field
 	 * @param bool $caseInsensitive Ignore upper/lowercase for group names
-	 * @return \Kirby\Cms\Collection
+	 * @return self
 	 * @throws \Kirby\Exception\Exception
 	 */
-	public function group($field, bool $caseInsensitive = true)
-	{
+	public function group(
+		$field,
+		bool $caseInsensitive = true
+	): self {
 		if (is_string($field) === true) {
 			$groups = new Collection([], $this->parent());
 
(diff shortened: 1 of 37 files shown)


src/Cms/File.php
```diff
@@ -35,6 +35,9 @@ class File extends ModelWithContent
 	use FileActions;
 	use FileModifications;
 	use HasMethods;
+	/**
+	 * @use HasSiblings<\Kirby\Cms\Files>
```
lukasbestle: Shouldn't those also be fully-qualified:

```suggestion
	 * @use \Kirby\Cms\HasSiblings<\Kirby\Cms\Files>
```

src/Cms/Collection.php
```diff
@@ -94,9 +88,10 @@ public function __unset($id)
 	 * an entire second collection to the
 	 * current collection
 	 *
-	 * @param mixed $object
+	 * @param TValue|self|array $object
```
lukasbestle: I think instances of child classes would still satisfy `instanceof self` as a child class instance is automatically also an instance of the parent class. So IMO this type hint should be:

```suggestion
	 * @param TValue|static|array $object
```

src/Cms/Collection.php
```diff
@@ -158,13 +159,15 @@ public function findBy(string $attribute, $value)
...
-	 * @return \Kirby\Cms\Collection
+	 * @return self
 	 * @throws \Kirby\Exception\Exception
 	 */
-	public function group($field, bool $caseInsensitive = true)
-	{
+	public function group(
+		$field,
+		bool $caseInsensitive = true
+	): self {
```
lukasbestle, line 166: I think the the `return parent::group()` even returns a Toolkit collection, so that case does not satisfy the `self` type hint.

src/Cms/Item.php
```diff
@@ -23,6 +23,9 @@
  */
 class Item
 {
+	/**
+	 * @use HasSiblings<self::ITEMS_CLASS>
```
lukasbestle: Does this dynamic definition work? And what's its purpose when `self::ITEMS_CLASS` will always be `\Kirby\Cms\Items`?

src/Cms/PageSiblings.php
```diff
@@ -16,101 +16,75 @@ trait PageSiblings
 	/**
 	 * Checks if there's a next listed
 	 * page in the siblings collection
-	 *
-	 * @param \Kirby\Cms\Collection|null $collection
 	 */
-	public function hasNextListed($collection = null): bool
+	public function hasNextListed(Collection $collection = null): bool
```
lukasbestle, line 20: I think this and the following should be:

```suggestion
	public function hasNextListed(Collection|null $collection = null): bool
```

src/Cms/Pages.php
```diff
@@ -295,11 +296,9 @@ public function findOpen(): Page|null
 	 * Custom getter that is able to find
 	 * extension pages
 	 *
-	 * @param string $key
 	 * @param mixed $default
-	 * @return \Kirby\Cms\Page|null
 	 */
-	public function get($key, $default = null)
+	public function get(string $key, $default = null): Page|null
```
lukasbestle: We could get rid of the remaining docblock type as well (supported since PHP 8.0):

```suggestion
	public function get(string $key, mixed $default = null): Page|null
```

src/Cms/User.php
```diff
@@ -664,6 +667,7 @@ protected function sessionFromOptions(Session|array|null $session): Session
 
 	/**
 	 * Returns the parent Users collection
+	 * @internal
```
lukasbestle: Not necessary as the method is `protected` anyway.

src/Cms/Language.php
```diff
@@ -436,8 +439,9 @@ public function save(): static
 
 	/**
 	 * Private siblings collector
+	 * @internal
```
lukasbestle: Same here.

src/Toolkit/Iterator.php
```diff
@@ -20,9 +20,16 @@
  * 										in this class would require
  * 										implementing them throughout
  * 										the code base: https://github.com/getkirby/kirby/pull/4886#pullrequestreview-1203577545
```
lukasbestle:
```suggestion
 *                                      in this class would require
 *                                      implementing them throughout
 *                                      the code base: https://github.com/getkirby/kirby/pull/4886#pullrequestreview-1203577545
```

src/Toolkit/Collection.php
```diff
@@ -135,12 +123,15 @@ public function __unset($key)
 	/**
 	 * Appends an element
 	 *
-	 * @param mixed $key
-	 * @param mixed $item
-	 * @param mixed ...$args
+	 * ```php
+	 * $collection->append('key', $value);
+	 * $collection->append([$value1, $value2]);
```
lukasbestle: Does passing an array actually work? I'd expect that to run into the `count($args) === 1` case and append the array to the data array instead of appending the array *values*.

src/Toolkit/Collection.php
```diff
@@ -463,11 +432,10 @@ public function flip()
 	/**
 	 * Getter
 	 *
-	 * @param mixed $key
 	 * @param mixed $default
 	 * @return mixed
```
lukasbestle:
```suggestion
	 * @template TDefault
	 * @param TDefault $default
	 * @return TValue|TDefault|null
```

src/Toolkit/Collection.php
```diff
@@ -1193,27 +1114,25 @@ public function when($condition, Closure $callback, Closure $fallback = null)
 	}
 
 	/**
-	 * Alias for $this->not()
+	 * @see \Kirby\Toolkit\Collection::not()
 	 *
 	 * @param string ...$keys any number of keys, passed as individual arguments
```
lukasbestle: Covered by the `@see` annotation:

```suggestion
```

src/Cms/Item.php
distantnative: It somehow seemed to work in some cases. but then it also doesn't. Psalm seems to be buggy there. I added an issue with them as well. There also only is `self` but not `static` but `self` seems to act more like `static`. But you're right, we probably should just replace it with hardcoded  values.... though I haven't found a good way for a default template value:

How can I tell Psalm: use `Items` when no template is passed, but use the passed template when a child class extends this class?

src/Cms/PageSiblings.php
distantnative: I only didn't do it cause https://github.com/getkirby/kirby/pull/6377 will add those all automatically everywhere needed.

src/Toolkit/Collection.php
distantnative: This is a nice one. I am just afraid how we will handle these in the docs.

lukasbestle: I think the docs should treat any such template types as `mixed` except if they were defined with `@template ... of ...`.

src/Cms/Item.php
lukasbestle: I don't think a child class can override the template type without also importing the trait a second time. Where would we put the `@use` rule in the child class otherwise?

If `self` really means `static` here, then keeping it dynamic would of course be even better. It's just quite confusing, so we might need an extra comment line explaining that.

src/Toolkit/Collection.php
distantnative: True that should work for method templates. For the class templates that get inherited/extended we probably really need to add some logic to getirby.com to resolve them across inheritance.

src/Cms/Item.php
distantnative: I think `Item` would need to add a new class template which it passes to HasSiblings. And then the classes extending `Item` need to pass the actual type via `@extends Item<Blocks>` (for example). This would work well except that for the `Item` class itself the template would be `mixed` as I don't think one can set a default template value of `Items` (as in "this template is Items as long as any child doesn't pass it's more specific type for the template").

src/Cms/Collection.php
distantnative: Should `Kirby\Toolkit\Collection::group()` really use `self` then or should that be switched to `static`?

src/Cms/Item.php
distantnative: I added explicit extends to all classes inheriting from `Item` - that seems to work.

src/Cms/Collection.php
lukasbestle: `Kirby\Toolkit\Collection::group()` currently uses `new Collection()`. I think the idea was that we e.g. wouldn't want to have a `Pages` object with the groups as each group is not a page but another collection. But it would indeed make sense to return a `Cms\Collection` at least from `Cms\Collection::group()`. That will be difficult to build though without adding a new argument for the class name to the `Toolkit\Collection::group()` method.

src/Cms/Item.php
lukasbestle: Don't think templates can have default values. But would have been great, I agree.

src/Cms/Collection.php
distantnative: Ok understand now the use case of not having it `static` all the way down to e.g. `Pages`.

`Kirby\Cms\Collection::group()` could return `Collection|BaseCollection`. Or we unwrap the BaseCollection from the `parent::group()` call and move the data into a CmsCollection. So it would always be `Kirby\Cms\Collection::group(): Kirby\Cms\Collection`

lukasbestle: Unwrapping sounds sensible. Then the return value will be consistent.

distantnative: I added a version, what do you think?

lukasbestle (approved):
Looks great 👍

PR #6393, opened 2024-04-11
Custom Panel view buttons

distantnative:
### Feature
- Panel view buttons extension
https://feedback.getkirby.com/411

```js
panel.plugin("getkirby/custom-view-buttons", {
  viewButtons: {
    applause: {
      template: `<k-button icon="heart" variant="filled" theme="love" size="sm" @click="applause">Applause</k-button>`,
      methods: {
        applause() {
           alert("👏");
         },
       },
    },
  },
});
```

Global config options:

```php
return [
  'panel' => [
    'viewButtons' => [
      'page' => ['preview', 'settings', 'applause', 'status']
    ]
  ]
}
  ```

Blueprint:

```yml
buttons:
  - preview
  - settings
  - applause
  - status
  ```
  
Screenshot 2024-04-25 at 13 06 19

### Deprecated
- `<k-languages-dropdown>` has been renamed to `<k-view-languages-button>`. `k-languages-dropdown` alias will be removed in a future version

### Breaking changes
- When overwriting/extending any model view, ensure to include new `<k-view-buttons :buttons="buttons">` component

### For review team

- [x] Add lab and/or sandbox examples (wherever helpful): https://github.com/getkirby/sandbox/pull/5

### DIFF
--- a/panel/public/js/plugins.js
+++ b/panel/public/js/plugins.js
@@ -7,6 +7,7 @@ window.panel.plugins = {
 	textareaButtons: {},
 	thirdParty: {},
 	use: [],
+	viewButtons: {},
 	views: {},
 	writerMarks: {},
 	writerNodes: {}
@@ -46,6 +47,11 @@ window.panel.plugin = function (plugin, extensions) {
 			...options,
 			mixins: ["section", ...(options.mixins ?? [])]
 		};
+	});
+
+	// View Buttons
+	resolve(extensions, "viewButtons", (name, options) => {
+		window.panel.plugins.components[`k-view-${name}-button`] = options;
 	});
 
 	// `Vue.use`
--- a/panel/src/components/View/Buttons/AddUsersButton.vue
+++ b/panel/src/components/View/Buttons/AddUsersButton.vue
@@ -1,0 +1,16 @@
+<template>
+	<k-button
+		:disabled="!$panel.permissions.users.create"
+		:text="$t('user.create')"
+		icon="add"
+		size="sm"
+		variant="filled"
+		@click="
+			$dialog('users/create', {
+				query: {
+					role: $panel.view.props.role?.id
+				}
+			})
+		"
+	/>
+</template>
--- a/panel/src/components/View/Buttons/AddUsersButton.vue
+++ b/panel/src/components/View/Buttons/AddUsersButton.vue
@@ -1,5 +1,6 @@
 <template>
 	<k-button
+		v-if="$panel.view.component === 'k-users-view'"
 		:disabled="!$panel.permissions.users.create"
 		:text="$t('user.create')"
 		icon="add"
--- a/panel/src/components/View/Buttons/RemoveLanguageButton.vue
+++ b/panel/src/components/View/Buttons/RemoveLanguageButton.vue
@@ -1,0 +1,10 @@
+<template>
+	<k-button
+		v-if="$panel.view.props.deletable"
+		:title="$t('delete')"
+		icon="trash"
+		size="sm"
+		variant="filled"
+		@click="$dialog(`languages/${$panel.view.props.id}/delete`)"
+	/>
+</template>
--- a/panel/src/components/View/Buttons/RemoveLanguageButton.vue
+++ b/panel/src/components/View/Buttons/RemoveLanguageButton.vue
@@ -1,6 +1,8 @@
 <template>
 	<k-button
-		v-if="$panel.view.props.deletable"
+		v-if="
+			$panel.view.component === 'k-language-view' && $panel.view.props.deletable
+		"
 		:title="$t('delete')"
 		icon="trash"
 		size="sm"
--- a/panel/src/components/View/Buttons/ThemeButton.vue
+++ b/panel/src/components/View/Buttons/ThemeButton.vue
@@ -1,14 +1,20 @@
 <template>
 	<k-button
+		v-if="$panel.view.id === 'account'"
 		:icon="icon"
 		:text="text"
 		size="sm"
 		variant="filled"
+		class="k-view-theme-button"
 		@click="$panel.theme.toggle()"
 	/>
 </template>
 
 <script>
+/**
+ * View header button to toggle the Panel theme
+ * @since 5.0.0
+ */
 export default {
 	computed: {
 		icon() {
--- a/panel/src/components/View/Buttons/index.js
+++ b/panel/src/components/View/Buttons/index.js
@@ -1,0 +1,32 @@
+import Add from "./AddButton.vue";
+import AddLanguages from "./AddLanguagesButton.vue";
+import AddUsers from "./AddUsersButton.vue";
+import Languages from "./LanguagesButton.vue";
+import Preview from "./PreviewButton.vue";
+import Remove from "./RemoveButton.vue";
+import RemoveLanguage from "./RemoveLanguageButton.vue";
+import Settings from "./SettingsButton.vue";
+import Status from "./StatusButton.vue";
+import Theme from "./ThemeButton.vue";
+
+import Buttons from "./Buttons.vue";
+
+export default {
+	install(app) {
+		app.component("k-view-add-button", Add);
+		app.component("k-view-add-languages-button", AddLanguages);
+		app.component("k-view-add-users-button", AddUsers);
+		app.component("k-view-languages-button", Languages);
+		app.component("k-view-preview-button", Preview);
+		app.component("k-view-remove-button", Remove);
+		app.component("k-view-remove-language-button", RemoveLanguage);
+		app.component("k-view-settings-button", Settings);
+		app.component("k-view-status-button", Status);
+		app.component("k-view-theme-button", Theme);
+
+		app.component("k-view-buttons", Buttons);
+
+		// @deprecated
+		app.component("k-languages-dropdown", Languages);
+	}
+};
--- a/panel/src/components/Views/Users/UserView.vue
+++ b/panel/src/components/Views/Users/UserView.vue
@@ -26,27 +26,7 @@
 			</template>
 
 			<template #buttons>
-				<k-account-theme-button v-if="$panel.view.id === 'account'" />
-
-				<k-button-group>
-					<k-button
-						:disabled="isLocked"
-						:dropdown="true"
-						:title="$t('settings')"
-						icon="cog"
-						size="sm"
-						variant="filled"
-						class="k-user-view-options"
-						@click="$refs.settings.toggle()"
-					/>
-					<k-dropdown-content
-						ref="settings"
-						align-x="end"
-						:options="$dropdown(id)"
-					/>
-					<k-languages-dropdown />
-				</k-button-group>
-
+				<k-view-buttons :buttons="buttons" />
 				<k-form-buttons :lock="lock" />
 			</template>
 		</k-header>
--- a/src/Panel/File.php
+++ b/src/Panel/File.php
@@ -65,6 +65,21 @@ class File extends Model
 	}
 
 	/**
+	 * Returns header button names which should be displayed
+	 * on the file view
+	 */
+	public function buttons(): array
+	{
+		return
+			$this->model->blueprint()->buttons() ??
+			$this->model->kirby()->option('panel.viewButtons.file', [
+				'preview',
+				'settings',
+				'languages'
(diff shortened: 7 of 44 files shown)

### CODE REVIEW

**Comment by lukasbestle** [comment]
I like the blueprint syntax of a simple list that also allows to reorder or hide the default buttons. I'd name the extension and blueprint option `headerButtons` though, otherwise it would be confusing next to the existing `textareaButtons` extension. The props passed to the extension components make sense IMO.

File: src/Panel/File.php [hunk 1]
```diff
@@ -64,6 +64,19 @@ public function breadcrumb(): array
...
+	 * Returns array of names for which
+	 * header buttons should be displayed
+	 */
+	public function buttons(): array
+	{
+		return $this->model->blueprint()->buttons() ?? [
+			"preview",
+			"settings",
+			"languages"
+		];
```
distantnative: @bastianallgeier or do you think the defaults should rather go into` PageBlueprint`/`FileBlueprint`... ?

tobimori:
Is it possible for a plugin to globally register a button on all pages/files? This would be helpful for https://github.com/tobimori/kirby-seo, so people don't have to add the robots indicator to each page. (of course they should be able to override it, as soon as `buttons` is specified on a blueprint)

distantnative:
@tobimori Extended the PR to allow setting the defaults via e.g. `Kirby\Panel\Page::$buttons[] = 'seo'`. It has the little downside that either you just add your plugin button name to the array and it will be added to the end. Or you overwrite the whole array with adding yours in the position you want this at. But if multiple plugins do that they'll overwrite each other's changes.

EDIT: well, if plugin developers act responsibly, they could still do

```php
array_splice(Kirby\Panel\Page::$buttons, 2, 0, 'seo');
```

Or something similar. Avoiding overwriting each other's additions.

distantnative:
@lukasbestle I adapted the JS entry to `headerButtons` for blueprint options I would really like to avoid camel-casing/too long option names. There I thought as it's on the top level of the blueprint `buttons` could be fine and still would be my favourite. One could also do an option `header` with a sub-option `buttons` but I feel then a lot more, e.g. `title` shouldn't be on the blueprint's top level but inside the `header` option.

@bastianallgeier @texnixe would be great to hear your opinions on naming etc. too, so I can finalise and polish this PR. Thanks!

panel/src/components/Views/Users/UserView.vue
```diff
@@ -26,27 +26,7 @@
...
-					/>
-					<k-dropdown-content
-						ref="settings"
-						align-x="end"
-						:options="$dropdown(id)"
-					/>
-					<k-languages-dropdown />
-				</k-button-group>
-
+				<k-header-buttons :buttons="buttons" :props="{ id, isLocked }" />
```
distantnative: Why am I not passing model and permissions here as well as props?

bastianallgeier:
I like pretty much everything about it. I just wonder if the buttons shouldn't just get all the same props that the view gets or if they should rather use panel.view instead to access everything they need to know.

distantnative:
I like the idea of using `$panel.view`. We're loosing the convenience of `isLocked` as that's not a prop but a computed. But I think that'd be ok. I am not sure yet if I would let each button access `$panel.view.props` itself. Or if `k-header-buttons` internally does `v-bind="$panel.view.props"` on each button component.

distantnative:
@bastianallgeier went with the "don't pass any props" approach. 

The CI only fails because an rpm bug, I think we can ignore it for this PR.

bastianallgeier:
There's an issue in the Header.vue where the slot is being wrapped with a k-header-buttons div but then the component also has the k-header-buttons class and the margin is applied twice: 

```html
<div v-if="$slots.buttons" class="k-header-buttons">
	<!-- @slot Position for optional buttons opposite the headline -->
	<slot name="buttons" />
</div>
```

I'm not sure how we can correctly fix it. If we remove the wrapping div, we turn this into a breaking change. But removing the class from the component would also be weird.

distantnative:
@bastianallgeier You are right. I think the new `View/Header/Buttons.vue` needs a more specific name. How about `k-model-header-buttons`?

bastianallgeier:
I keep coming back to the word "action" something like `k-view-action-buttons` maybe?

distantnative:
What about just `k-view-buttons`?

bastianallgeier:
I'm happy with that

panel/public/js/plugins.js
```diff
@@ -2,6 +2,7 @@ window.panel = window.panel ?? {};
 window.panel.plugins = {
 	components: {},
 	created: [],
+	headerButtons: {},
```
bastianallgeier: It would probably make sense to rename to viewButtons here as well, right?

distantnative: I thought about it but is it too confusing that the panel extension is `viewButtons` but the config option is `headerButtons`? Cause I don't think view buttons makes sense for the config option as this is not for all views but just those 4 model views.

bastianallgeier: I haven't thought about it before, but do we really want to limit them to model views only? Couldn't they be useful for other views as well?

distantnative: Hmm I think it's already a bit on the line: each button could technically be used in any model view, but not all buttons will work in all views. And we want to keep the names simple. So we want `preview` and not `preview-page`, `preview-file`, `preview-system`.... but opening this to more views, e.g. users and languages with their add buttons, this will get really difficult with overlaps in names, functionality etc.

bastianallgeier: You are right. Can we find anything else then that we could use in all places. It feels a bit weird to deal with different names.

distantnative: My mind still jumps to `model` then. But it feels a bit less connected to where they show up.

distantnative: Or we indeed open it up for all views (I mean I could see cases where it's nice to add a button on the `users` view via the config). But then our individual buttons have to get quite smart/encompass a bunch of logic. E.g. one `k-view-add-button` that includes the logic of the add buttons for users view as well as languages view. With a generic text for when used in another view context.

distantnative: The idea is growing on me. It'll take some work to transform the button components but I think in the end it could be a nice solution and the most flexible one. And then this would totally make sense term wise:

```php

return [
  'panel' => [
    'viewButtons' => [
      'page' => ['preview', 'settings', 'applause', 'status'],
      'users' => ['add', 'import']
    ]
  ]
];
```

distantnative: A `k-view-add-button` would look something like this:

```html
<template>
	<k-button
		:disabled="disabled"
		:text="text"
		icon="add"
		size="sm"
		variant="filled"
		@click="onClick"
	/>
</template>

<script>
export default {
	computed: {
		disabled() {
			if (this.$panel.view.component === "k-users-view") {
				return !this.$panel.permissions.users.create;
			}

			return false;
		},
		text() {
			if (this.$panel.view.component === "k-languages-view") {
				return this.$t("language.create");
			}

			if (this.$panel.view.component === "k-users-view") {
				return this.$t("user.create");
			}

			return this.$t("add");
		}
	},
	methods: {
		onClick() {
			if (this.$panel.view.component === "k-languages-view") {
				return this.$dialog("languages/create");
			}

			if (this.$panel.view.component === "k-users-view") {
				return this.$dialog("users/create", {
					query: {
						role: this.$panel.view.props.role?.id
					}
				});
			}
		}
	}
};
</script>
```

distantnative: @bastianallgeier this is how it could look https://github.com/getkirby/kirby/commit/f5cf0fadab5f71c7264ca0761277c68ce3d78088

I really love the flexibility of also being able to add buttons to those other views. I am not sure if I am happy enough though with the multiple add button components etc.

bastianallgeier: I also really like it and I think it is the cleanest option in terms of naming. I would avoid confusion about the add button by simply introducing multiple buttons "add-user", "add-page", etc.

distantnative: @bastianallgeier I have those in the other branch, but I thought they need a wrapping component so that users in their blueprints etc. really just reference `add`, `preview`, `setttings`etc.

I think blueprints/configs where they always have to write `preview-page`, `settings-model` etc. aren't so pretty/easy to use.

bastianallgeier: You are right. I didn't think about the blueprints.

panel/src/components/View/Buttons/ThemeButton.vue
```diff
@@ -1,14 +1,20 @@
 <template>
 	<k-button
+		v-if="$panel.view.id === 'account'"
```
bastianallgeier, line 3: Could there be any other view where the users might want to include the theme button?

panel/src/components/View/Buttons/index.js
```diff
@@ -0,0 +1,32 @@
...
+import Status from "./StatusButton.vue";
+import Theme from "./ThemeButton.vue";
+
+import Buttons from "./Buttons.vue";
+
+export default {
+	install(app) {
+		app.component("k-view-add-button", Add);
+		app.component("k-view-add-languages-button", AddLanguages);
+		app.component("k-view-add-users-button", AddUsers);
```
bastianallgeier, line 18: I know that this is useful to match the view name with the right button, but the plural feels off here to me.

panel/src/components/View/Buttons/RemoveLanguageButton.vue
```diff
@@ -0,0 +1,10 @@
+<template>
+	<k-button
+		v-if="$panel.view.props.deletable"
```
bastianallgeier: What happens if the button is used in a view that uses the deletable prop for something completely different?

panel/src/components/View/Buttons/AddUsersButton.vue
```diff
@@ -0,0 +1,16 @@
...
+	<k-button
+		:disabled="!$panel.permissions.users.create"
+		:text="$t('user.create')"
+		icon="add"
+		size="sm"
+		variant="filled"
+		@click="
+			$dialog('users/create', {
+				query: {
+					role: $panel.view.props.role?.id
```
bastianallgeier, line 12: After seeing a few of such dependency on view props, I wonder if we might need a system that a view can provide configuration for potential buttons and then the buttons can specifically check for those.

panel/src/components/View/Buttons/RemoveLanguageButton.vue
distantnative: We probably should prepend another if condition that only shows this button on the languages view for now.

panel/src/components/View/Buttons/ThemeButton.vue
distantnative: In theory this button could live on any view as it does not depend on any props. But since our user and account view are kind of the same, then it would by default show up on any user view but always only relate to the current user, which could be quite confusing.

panel/src/components/View/Buttons/index.js
distantnative: Do you have a suggestion how to do the matching instead?

bastianallgeier (approved):

panel/src/components/View/Buttons/index.js
bastianallgeier: I made up my mind. Let's keep it like it is. We are overcomplicating this PR and I really think that it is at a point where it is more than clean enough to be in the alpha.

philippoehrlein:
I just enabled this feature in my plugin, and here are some thoughts about it.

**From a plugin developer’s perspective:**
This is a very useful feature, though it does take away a bit of the satisfaction of having achieved this functionality independently for my plugin! :)

**From a web developer’s perspective:**
Not sure yet, but I do wonder if the current approach could use some refinement. Since developers need to set up Kirby’s default buttons themselves, it introduces some risk for errors. You might consider adding safeguards for some of the core buttons.

For example:
	•	If the status button is configured, it appears at the specified position.
	•	If it’s not included, it defaults to the end of the button group.

Just my two cents. In any case, this is a great update for Kirby!

distantnative:
> For example:
• If the status button is configured, it appears at the specified position.
• If it’s not included, it defaults to the end of the button group.

We actually want to give developers the flexibility to also exclude certain core buttons, if they prefer that.

PR #6394, opened 2024-04-11
`k-button`: New `theme` notations w/ `-icon` suffix for only colored icon

distantnative:
### Enhancement
- `<k-button>`: `theme` supports values suffixed with `-icon` now, e.g. `positive-icon` for filled buttons with only the icon colored
https://feedback.getkirby.com/635

## Docs

```html
<k-button icon="heart" theme="negative-icon" variant="filled">Love</k-button>
```

### For review team

### DIFF
--- a/panel/src/components/Navigation/Button.vue
+++ b/panel/src/components/Navigation/Button.vue
@@ -312,21 +312,21 @@ export default {
 }
 
 /** Filled Buttons **/
-.k-button:where([data-variant="filled"], [data-variant="filled-icon"]) {
+.k-button:where([data-variant="filled"]) {
 	--button-color-back: var(--color-gray-300);
 }
 .k-button:where([data-variant="filled"]):not([aria-disabled]):hover {
 	filter: brightness(97%);
 }
-.k-button:where([data-theme][data-variant="filled"]) {
+.k-button:where([data-theme$="-icon"][data-variant="filled"]) {
+	--button-color-icon: var(--theme-color-600);
+}
+.k-button:where([data-theme][data-variant="filled"]):not(
+		[data-theme$="-icon"]
+	) {
 	--button-color-icon: var(--theme-color-700);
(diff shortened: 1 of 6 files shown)


panel/src/components/Navigation/Button.vue
```diff
@@ -307,7 +307,6 @@ export default {
 	--button-color-text: var(--button-color-dimmed-off);
 }
 .k-button:where([data-theme][data-variant="dimmed"]) {
-	--button-color-icon: var(--theme-color-icon);
```
bastianallgeier, line 310: Removing this line breaks the colors for the dimmed buttons. 
Screenshot 2024-04-23 at 12 19 00

distantnative: Ahh got it, we can out it back in.

bastianallgeier:
Something about the logic is still off. The text shouldn't be colored and the empty theme also does not seem to work properly. 

Screenshot 2024-04-24 at 15 44 33

bastianallgeier:
My last commit is just a suggestion, but I think this PR really uncovers some issues: 

- theme-color-text-dimmed is indeed pretty useless and too bright. I tried to fix that
- the new -icon styles make the button CSS a lot more complex. I tried to simplify/clarify the rules. It is a bit more redundant but I think it's often easier to understand when we go from less specific to more specific from top to bottom. 
- the icon color on the gray button background is too weak in contrast for many colors (green, yellow, aqua) but if we bump it too much the status buttons look shit. I tried to find a good compromise. 
- The empty theme does not fit into the other color themes at all and actually makes things weird. We should not try to mix it with colors and icons and stuff. It would probably be good to turn it into a variant instead. But that's not that easy. I kept it seperately in the examples
- I improved the examples quite a bit more to really see what we are doing. 

Screenshot 2024-04-25 at 12 26 54

distantnative:
@bastianallgeier makes sense to me

PR #6398, opened 2024-04-15
Extend PHP code styling rules

distantnative:
## This PR …

@lukasbestle we can't exclude QrCode from just some rules. So I think it's better to loose the pretty formatting of numbers that we won't understand nevertheless but have the rest of the file in conformity Thant to exclude it fully.

### Housekeeping
- Extended rule list for `php-cs-fixer`

### For review team


lukasbestle (dismissed):
I read through all changes and fixed those places where PHP-CS-Fixer alone couldn't fully do the changes.

About the `QrCode` class: It's totally fine for me that the manual indentation is gone. In the last version it somehow got butchered in a way that kept some of the indentation and that didn't really make sense. Now it's great.

lukasbestle:
Bildschirmfoto 2024-04-15 um 22 10 53

distantnative:
I guess there was a reason after all, why I didn't include 'nullable_type_declaration_for_default_null_value`... 

I think we have to remove that for now. These are not easy fixes but changes we should tackle separately (but probably need to do sooner than later: https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated

lukasbestle:
I think it's a rather simple fix by removing the no longer needed and wrong `= null` default values in the affected places.

But feel free to merge this PR without the last commit. I still have that locally and will prepare another PR with the fixed changes.

distantnative:
👍 just think it's cleaner/easier to see later on if we do it in separate steps

lukasbestle:
I agree, also makes sense to not let this PR lay around for long to avoid merge conflicts. Thanks for taking care of it 👍

PR #6407, opened 2024-04-23
CSS `data-theme` with color names

distantnative:
I think it could make sense also to support the literal color names for `data-theme`


bastianallgeier:
I like the idea, but I think this should belong to 4.3 and not into the patch release.

distantnative:
Sure

bastianallgeier:
I've added examples to the lab and also added gray as alias for passive.

bastianallgeier (approved):

PR #6411, opened 2024-04-26
DropdownContent: support global events for `click`

distantnative:
## This PR …

- [x] Should the event should also be emitted globally when `click` is a string or only when an object?

### Enhancement
- Dropdown options can emit global Panel events on click now, which can be used e.g. when extending setting dropdowns in a plugin to react in JS with an action when option is clicked
https://feedback.getkirby.com/638

## Docs

Dropdown options `click` handler:

- `click` is a function, will be called (only works when options get passed in JS)
- `click` is a string: emit to parent as action event
- `click` is an object: 
  - If `name` key is set, emit to parent with optional `payload`
  - If `global` key is set, emit globally with optional `payload`

Should be mentioned around here: https://getkirby.com/docs/reference/plugins/extensions/panel-dropdowns#extending-dropdowns

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Dropdowns/DropdownContent.vue
+++ b/panel/src/components/Dropdowns/DropdownContent.vue
@@ -187,10 +187,24 @@ export default {
 		onOptionClick(option) {
 			this.close();
 
+			// click handler is a callback function
+			// which is executed
 			if (typeof option.click === "function") {
-				option.click.call(this);
-			} else if (option.click) {
+				return option.click.call(this);
+			}
+
+			// click handler is an action string
+			// which is emitted to the parent as well as globally
+			if (typeof option.click === "string") {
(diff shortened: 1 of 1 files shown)


bastianallgeier (dismissed):

bastianallgeier:
To answer your question: I think I would leave it like this and not emit a global event. That could get pretty messy.

distantnative:
@bastianallgeier that's confusing 😆 cause like this right now it does also submit an event globally when the `click` option is just a string. So "leave it like this" or "don't submit it globally"?

bastianallgeier:
@distantnative sorry for the confusion. The discussion continues on Nolt :)

bastianallgeier (approved):

PR #6414, opened 2024-04-27
Page section: show Add based on `create` status

distantnative:
### Enhancement
- Page section: Show Add button also in sections only showing listed/unlisted pages when all allowed blueprints create new child pages directly as listed/unlisted https://feedback.getkirby.com/181

## Ready?

### For review team

- [x] Add lab and/or sandbox examples (wherever helpful) https://github.com/getkirby/sandbox/pull/6

### DIFF
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -236,9 +236,18 @@ return [
 				return false;
 			}
 
-			// ensure that the add button is only shown when
-			// the section is displaying pages with the same status
-			// as the status that is used for page creation
+			// form here on, we need to check with which status
+			// the pages are created and if the section can show
+			// these newly created pages
+
+			// if the section shows pages no matter what status they have,
+			// we can always show the add button
+			if ($this->status === 'all') {
+				return true;
+			}
+
+			// collect all statuses of the blueprints
+			// that are allowed to be created
 			$statuses = [];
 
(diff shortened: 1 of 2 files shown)


config/sections/pages.php
```diff
@@ -232,11 +232,30 @@
...
+
+			$statuses = array_unique($statuses);
+
+			if (count($statuses) > 1 && $this->status !== 'all') {
+				return false;
+			}
+
+			if (in_array($this->status, [...$statuses, 'all']) === false) {
 				return false;
 			}
```
lukasbestle: Since `$this->status === 'all'` always allows the add button, we could have this as a shortcut above the new logic:

```php
if ($this->status === 'all') {
	return true;
}
```

These checks could then be:

```suggestion
			if (count($statuses) > 1 || $this->status !== $statuses[0]) {
				return false;
			}
```

distantnative: Good call.

I think the last check has to be a bit different as the array can also be empty.

lukasbestle: I guess it would be `count($statuses) !== 1` then, right?

lukasbestle (approved):
The code looks good, the test is super hard to review with the combination of of reordering and new code. It's probably fine in this case, but in general I'd really prefer to keep refactoring and code changes separate.

bastianallgeier (approved):

PR #6415, opened 2024-04-27
Remove server allowlist

distantnative:
## This PR …

- [ ] Do we want to display the full server name in the Panel now or do we still want to clean it up?
Screenshot 2024-04-27 at 12 47 00

### Enhancement
- Remove hard-coded restriction of supported servers https://feedback.getkirby.com/567

## Docs

- Remove hard server requirements in favor of something like
  
  > Apache works out of the box, nginx and Caddy work with our cookbook recipes and other servers can be used by advanced users
- Remove `servers` option

## Ready?

### For review team

### DIFF
--- a/tests/Cms/Api/routes/SystemRoutesTest.php
+++ b/tests/Cms/Api/routes/SystemRoutesTest.php
@@ -19,19 +19,6 @@ class SystemRoutesTest extends TestCase
 		]);
 	}
 
-	public function testGetWithValidServerSoftware()
-	{
-		$app = $this->app->clone([
-			'server' => [
-				'SERVER_SOFTWARE' => 'apache'
-			]
-		]);
-
-		$response = $app->api()->call('system', 'GET');
-
-		$this->assertTrue($response['data']['isOk']);
-	}
-
 	public function testGetWithoutUser()
 	{
 		$response = $this->app->api()->call('system', 'GET');
--- a/tests/Cms/Api/routes/SystemRoutesTest.php
+++ b/tests/Cms/Api/routes/SystemRoutesTest.php
@@ -19,21 +19,6 @@ class SystemRoutesTest extends TestCase
 		]);
 	}
 
-	public function testGetWithInvalidServerSoftware()
-	{
-		// set invalid server software
-		$app = $this->app->clone([
-			'server' => [
-				'SERVER_SOFTWARE' => 'invalid'
-			]
-		]);
-
-		$response = $app->api()->call('system', 'GET');
-
-		$this->assertFalse($response['data']['isOk']);
-		$this->assertFalse($response['data']['requirements']['server']);
-	}
-
 	public function testGetWithValidServerSoftware()
 	{
 		$app = $this->app->clone([
--- a/tests/Cms/System/SystemTest.php
+++ b/tests/Cms/System/SystemTest.php
@@ -135,8 +135,9 @@ class SystemTest extends TestCase
 			['Nginx', true],
 			['caddy', true],
 			['Caddy', true],
-			['iis', false],
-			['something', false],
+			['iis', true],
+			['something', true],
+			[null, false]
 		];
 	}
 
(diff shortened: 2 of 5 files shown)


tests/Cms/Api/routes/SystemRoutesTest.php
```diff
@@ -19,21 +19,6 @@ public function setUp(): void
...
-			]
-		]);
-
-		$response = $app->api()->call('system', 'GET');
-
-		$this->assertFalse($response['data']['isOk']);
-		$this->assertFalse($response['data']['requirements']['server']);
-	}
-
 	public function testGetWithValidServerSoftware()
```
lukasbestle: This test can be removed as well.

tests/Panel/Areas/InstallationTest.php
```diff
@@ -21,20 +21,18 @@ public function testInstallation()
 		$this->assertSame('Installation', $view['title']);
 		$this->assertSame('k-installation-view', $view['component']);
 
-		// cli is not a valid server, therefor installation fails
 		$this->assertFalse($view['props']['isInstallable']);
 		$this->assertFalse($view['props']['isInstalled']);
-		$this->assertFalse($view['props']['isOk']);
+		$this->assertTrue($view['props']['isOk']);
```
lukasbestle, line 26: With this change, the test does not actually test anymore whether the `isOk` prop switches to `false` on issues. We could instead e.g. make the `content` folder non-writable in the test.

tests/Cms/System/SystemTest.php
```diff
@@ -619,11 +620,10 @@ public function testPlugins()
 	}
 
 	/**
-	 * @covers ::server
 	 * @covers ::serverSoftware
 	 * @dataProvider providerForServerSoftware
 	 */
-	public function testServer($software, $expected)
+	public function testServerSoftware($software, $expected)
```
lukasbestle: As the new code just passes through the environment variable 1:1, there is IMO no need for the data provider anymore. It's enough to test this without any value and with a single valid value.

lukasbestle (commented):
> Do we want to display the full server name in the Panel now or do we still want to clean it up?

Since `SERVER_SOFTWARE` is an environment variable, it can be set to whatever value. I briefly looked for standards, but it seems like there is none. So we cannot expect the value to be set in a uniform way. For debugging it may actually be useful to have access to the full string including version, so I think we could keep it like this.

tests/Panel/Areas/InstallationTest.php
distantnative: I don't think the test was really meant to test the false `isOk` – I think that was just an unavoidable circumstance back then when testing in CLI would not allow the server. But if I look at the rest of the test, I think this is meant to test a positive installation state.

We are testing whether `isOk` works or not in its regular tests in `SystemTest`. If we want to add more test cases there, we can, but I would keep this separate from this PR.

lukasbestle: I think we should absolutely have a test for that. Otherwise we could easily break the installation test by accident and we would only realize that after having debugged support requests for ages. IMO it's a regression to remove the test in this PR. A test for the "normal sunshine case" is not useful as it only provides code coverage without actually testing the behavior.

But it's fine for me to merge it first if you add a test separately, directly to the develop branch without another PR would be fine as well.

lukasbestle (approved):

tests/Panel/Areas/InstallationTest.php
distantnative: But this test isn't testing `$system->isOk()`, it's testing that the installation view receives props. Testing `$system->isOk()` is important, I agree, but it should take place in `SystemTest`. I think we should avoid this kind of cross-testing where we can and rather have actual tests for these parts.

distantnative: A change that could make sense here to really test the installation props functionality is to write

```php
$this->assertSame($system->isOk(), $view['props']['isOk']);
```

It seems a bit redundant, but that's the only thing we should test in this test case: is the prop providing the state our system checks return.

lukasbestle: Ah, right. I missed that this is a test for the area. I agree.

PR #6416, opened 2024-04-27
System view: copy info

distantnative:
### Feature
- System view: added button to copy relevant system information

### For review team

### DIFF
--- a/panel/src/components/Views/System/SystemView.vue
+++ b/panel/src/components/Views/System/SystemView.vue
@@ -38,6 +38,7 @@ export default {
 	props: {
 		environment: Array,
 		exceptions: Array,
+		info: Object,
 		plugins: Array,
 		security: Array,
 		urls: Object
@@ -57,10 +58,7 @@ export default {
 		copy() {
 			const info = JSON.stringify(
 				{
-					version: this.environment[1].value,
-					license: this.environment[0].value,
-					php: this.environment[2].value,
-					server: this.environment[3].value,
+					info: this.info,
 					security: this.security.map((issue) => issue.text),
 					plugins: this.plugins.map((plugin) => ({
 						name: plugin.name.text,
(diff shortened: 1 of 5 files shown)


panel/src/components/Views/System/SystemView.vue
```diff
@@ -42,6 +52,30 @@ export default {
...
 		}
+	},
+	methods: {
+		copy() {
+			const info = JSON.stringify(
+				{
+					version: this.environment[1].value,
+					license: this.environment[0].value,
+					php: this.environment[2].value,
+					server: this.environment[3].value,
```
lukasbestle: This is certainly the best way to avoid sending the data to the frontend twice, but it could easily break if we reorganize the system view. Maybe we could instead send those values as a separate `info` prop from the backend and then add the security and plugin info dynamically in the frontend.

tests/Cms/System/SystemTest.php
```diff
@@ -339,6 +339,16 @@ public function testInitPermission($root)
 		new System($app);
 	}
 
+	/**
+	 * @covers ::info
+	 */
+	public function testIinfo()
```
lukasbestle:
```suggestion
	public function testInfo()
```

tests/Cms/System/SystemTest.php
```diff
@@ -339,6 +339,16 @@ public function testInitPermission($root)
...
+	{
+		$app = $this->app->clone([]);
+		$system = new System($app);
+		$this->assertIsArray($system->info());
```
lukasbestle: With type hinting, the only thing this assertion does is to check whether the method explodes if it's being called. For that I'd rather use `$system->info(); $this->expectNotToPerformAssertions();`. But the usefulness of the test is still almost zero.

I get that the Kirby and PHP versions cannot be tested well and the server software and license label are unlikely to break, but maybe we could at least test the languages array as it is built using actual logic.

lukasbestle (approved):

gbdesign2023:
The new info field in version 4.3 for the server configuration should certainly not look like this, right?

b

I liked the display more in version 4.2. 

a

distantnative:
@gbdesign2023 Please don't comment on closed PRs but open an issue. It's hard to keep track of this otherwise.

PR #6419, opened 2024-04-28
Unit test for `$system->isOk()`

distantnative:


tests/Cms/System/SystemTest.php
```diff
@@ -507,6 +502,22 @@ public function testIsOk()
...
+	/**
+	 * @covers ::isOk
+	 */
+	public function testIsOkContentMissingPermissions()
+	{
+		$app    = $this->app->clone([]);
+		$system = new System($app);
+
+		$before = fileperms($app->root('content'));
+		chmod($app->root('content'), 0000);
```
lukasbestle:
```suggestion
		chmod($app->root('content'), 0o000);
```

tests/Cms/System/SystemTest.php
```diff
@@ -507,6 +502,22 @@ public function testIsOk()
...
+
+		$this->assertFalse($system->isOk());
+
+		chmod($app->root('content'), $before);
```
lukasbestle: This should happen in a `tearDown()` method, otherwise the test suite will error out on cleanup if the assertion fails and this is not run.

bastianallgeier: Do we even need to reset it at all? The entire test installation is killed after the test anyway.

distantnative: Without it a number of tests started failing.

distantnative: @lukasbestle could you please give it a try to move it to the `tearDown()` method? My attempts just end in phpunit completely blowing up with errors :/

lukasbestle: > Do we even need to reset it at all? The entire test installation is killed after the test anyway.

We need to reset it because a folder without write permissions can only be deleted by `root`. That's why subsequent tests would fail if PHPUnit cannot clean up the test dir.

@distantnative I can take a look in the next days.

bastianallgeier: @lukasbestle of course 💡 That explanation makes total sense.

lukasbestle (approved):
Fixed it by hooking into the existing `tearDown` logic. I also added another test for `status` with the same setup.

PR #6420, opened 2024-04-28
Panel search: abort previous requests

### LINKED ISSUE #6409: Search in panel shows old results when response times from server vary
## Description
The search in the panel can show partial/old results when the response time from the server can vary. Example: You want to search for "test". You type slowly, so a first query is launched for "te", then another one when you finished typing "test". If the request for "test" is faster to load, it shows those results. If afterwards the request for "te" completes, it overwrites the results shown. So you seem to see the results for "test" but you actually see the results for "te".

**Expected behavior**  
When a new search query is launched, the old running queries should be canceled so that this behaviour can't happen.

## To reproduce
Easiest way might be to overwrite the search method to include a random delay:
```php
Kirby::plugin('searchtest', [
    // This is overwriting the search used in the panel
    'areas' => [
        'site' => function ($kirby) {
            return [
                'searches' => [
                    'pages' => [
                        'query' => function (string $query = null, int $limit, int $page) use ($kirby) {
                            sleep(rand(0, 3));
                            // Simply pass on to the core search
                            return $kirby->core()->area('site')['searches']['pages']['query']($query, $limit, $page);
                        }
                    ]
                ]
            ];
        },
    ]
]);
```
Then go to the panel and quickly type part of the query, wait half a second, then type the rest. You might have to try a few times so that the first query actually takes longer than the second one.

## Your setup
**Kirby Version**  
4.2.0

**Additional context**  
The behaviour might not happen with just the default setup and few results. But if you overwrite the panel search with a custom search that e.g. fetches data from a big DB or another API, this happens and there's no way to fix it within the custom search.

distantnative:
### Fixes
- Panel search: previous search requests are aborted when new ones are sent to the backend
https://github.com/getkirby/kirby/issues/6409

### For review team

### DIFF
--- a/panel/src/components/Dialogs/SearchDialog.vue
+++ b/panel/src/components/Dialogs/SearchDialog.vue
@@ -40,7 +40,7 @@
 			/>
 		</div>
 
-		<div v-if="query?.length > 1" class="k-search-dialog-results">
+		<div v-if="query?.length > 1 && !isLoading" class="k-search-dialog-results">
 			<!-- Results -->
 			<k-collection
 				v-if="items.length"
--- a/panel/src/panel/panel.js
+++ b/panel/src/panel/panel.js
@@ -9,6 +9,7 @@ import Notification from "./notification.js";
 import Language from "./language.js";
 import Plugins from "./plugins.js";
 import Menu from "./menu.js";
+import Search from "./search.js";
 import System from "./system.js";
 import Translation from "./translation.js";
 import { buildUrl, isUrl } from "@/helpers/url.js";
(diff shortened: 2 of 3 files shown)


bastianallgeier:
There are now quite a few situations where the search is showing "No results" while there's still a search going on in the background that will eventually deliver results. It feels a bit confusing.

distantnative:
@bastianallgeier tricky one. I think we cannot work with the current boolean `isLoading` then but rather need a counter incrementing when a request starts, decrementing when one is finished/aborted. What do you think?

bastianallgeier:
Yep, makes sense. I also saw yesterday that we need to improve the "no results" part. It should only be shown if nothing is loading and maybe show a "Searching …" placeholder instead if there are no results in the list but loading is still happening.

distantnative:
@bastianallgeier I tried to change it, but I am missing something, it still happens for me. Mostly when deleting parts of the query with backspace.

distantnative:
Starting over with this

PR #6421, opened 2024-04-28
Chunked file uploads

distantnative:
### Summary of changes
- New `Api\Upload` class
  - Takes over all the nested code from `Api\Api::upload()` to make it more easily to read and test
  - Adds support for uploads in chunks

### Reasoning
- https://feedback.getkirby.com/256
- Cleaner class structure and better to test

### Features
- Panel uploads can now exceed the `upload_max_filesize` limit as they're uploaded in chunks. If you want to restrict the upload size, please use the file blueprint `accept` `maxsize` option: https://getkirby.com/docs/reference/panel/blueprints/file#accept

### Enhancements
- New `uploadAsChunks` JS helper function

### Fixed
- Canceling the file upload dialog now also cancels ongoing uploads

### Refactored
- New `Kirby\\Api\Upload` class to handle file uploads via the REST API

## Ready?

### For review team


afbora:
I've tried to upload a EXE file. Kirby 4.2.0 throws a error `The extension "exe" is not allowed` but nothing happens for this PR. Seems something broken.

afbora:
Also I realized that file is still remains when I cancel the upload. Should be deleted. But should be careful when file replacing.

distantnative:
@afbora We cannot delete the tmp file on cancel. But we clean up the files with a garbage collection.

config/api/routes/files.php
```diff
@@ -47,17 +51,28 @@
 			// move_uploaded_file() not working with unit test
 			// @codeCoverageIgnoreStart
 			return $this->upload(function ($source, $filename) use ($path) {
-				$props = [
+				// handle chunked uploads
+				$source = Upload::chunk($this, $source, $filename);
+
+				// if file is not yet complete, return
+				if ($source === null) {
+					return;
```
distantnative: Works already, but wondering if for these non-last chunk responses a different HTTP response code or so should be sent?

panel/src/helpers/upload.js
```diff
@@ -1,3 +1,45 @@
...
+		if (parts > 1) {
+			params.headers["X-Upload-Chunk"] = end === file.size ? "last" : "chunk";
+		}
+
+		response = await upload(chunk, {
+			...params,
+			headers: {
+				...params.headers,
+				"X-Unique-Upload-Id": id,
+				"Content-Range": `bytes ${start}-${end - 1}/${file.size}`
```
distantnative: I think, `- 1` actually isn't needed here as `end` is already zero-indexed

src/Panel/Panel.php
```diff
@@ -137,6 +138,16 @@ public static function firewall(
 		return true;
 	}
 
+	/**
+	 * @since 4.3.0
+	 */
+	protected static function garbage(): void
+	{
+		// run garbage collection with a chance of 20%;
+		if (mt_rand(1, 10000) <= 0.2 * 10000) {
```
distantnative: Is 20% probability a good value for the Panel garbage collection?

src/Api/Upload.php
```diff
@@ -0,0 +1,103 @@
...
+		return (int)floor(min($max) * 0.95);
+	}
+
+	/**
+	 * Returns root of directory used for
+	 * temporarily storing (incomplete) uploads
+	 */
+	public static function tmp(): string
+	{
+		return App::instance()->root('media') . '/uploads';
```
distantnative: Is media/uploads an issue as tmp directory as it could be accessible?
Is it especially problematic, that e.g. half a PHP/.exe file could be uploaded to the tmp directory? Rules only kick in after last chunk has been uploaded.

config/api/routes/files.php
bastianallgeier: This one sounds good but is unfortunately depreacted: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/102

206? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206 

They both seem to only fit partially.

src/Api/Upload.php
bastianallgeier: I think it could be risky, but @lukasbestle should probably have a look at it.

config/api/routes/files.php
distantnative: I think 206 is for the reverse scenario: where I send a request to receive a range of content and the response only contains that range. So wouldn't touch that. Maybe as is is fine

src/Api/Upload.php
distantnative: Alternatively, do you have ideas for alternatives. The sys tmp folders (as initially suggested on Notion) don't work as they don't persist across the multiple requests.

lukasbestle: Maybe we could upload to an invisible folder in `content` like `/content/.uploads`. That folder is already blocked in web access by our suggested default server configs. Also if someone has set up the content folder on a different filesystem, we'd avoid moving the file between different filesystems after the last chunk.

lukasbestle: In addition, it would be great to run at least basic security checks if not all file rules. Having a PHP or Phar file on the server would be game over. One can easily upload a full file and pretend that another chunk follows afterwards, but never send another chunk.

src/Panel/Panel.php
lukasbestle: Maybe a bit too much. Using the Panel sends a lot of requests. I'd say we should aim for in average one execution for each short editing session (like open the Panel, navigate to a page, fix a typo). Based on that, I think 10% is probably more than enough.

config/api/routes/files.php
lukasbestle: I agree, 200 seems fine.

panel/src/helpers/upload.js
lukasbestle: Yes, but I wouldn't use the `Content-Range` header. In requests it is [only allowed for partial `PUT` requests, with other HTTP verbs it "MUST" be ignored by the server](https://datatracker.ietf.org/doc/html/rfc9110#section-14.4-4). So we could run into situations where a web application firewall filters it out. I think that's the reason why the tus protocol uses two custom headers (see my separate review comment on the backend code).

config/api/routes/files.php
```diff
@@ -1,6 +1,10 @@
 <?php
 
 // routing pattern to match all models with files
+
+use Kirby\Api\Upload;
+use Kirby\Filesystem\F;
```
lukasbestle: These should be above the comment.

panel/src/helpers/upload.js
```diff
@@ -1,3 +1,45 @@
...
+ */
+export async function chunked(file, params, size = 5242880) {
+	const parts = Math.ceil(file.size / size);
+	const id = random(4).toLowerCase();
+	let response;
+
+	for (let i = 0; i < parts; i++) {
+		const start = i * size;
+		const end = Math.min(start + size, file.size);
+		const chunk = file.slice(start, end, file.type);
```
lukasbestle: How does this slicing work? Does the browser stream the file to the server or does it need to read the chunk into client memory?

src/Api/Upload.php
```diff
@@ -0,0 +1,103 @@
...
+		string $source,
+		string $filename
+	): string|null {
+		// if the file is uploaded in chunks
+		if ($isChunk = $api->requestHeaders('X-Upload-Chunk')) {
+			// ensure the directory exists
+			Dir::make($dir = static::tmp());
+
+			// append chunk to tmp file with unique filename (id suffix)
+			$id     = $api->requestHeaders('X-Unique-Upload-Id');
```
lukasbestle: Needs to be filtered for safe characters, otherwise we would have a major directory traversal attack vector.

src/Api/Upload.php
```diff
@@ -0,0 +1,103 @@
...
+			$chunk  = fopen($source, 'r');
+			$source = $dir . '/' . $filename;
```
lukasbestle: - Filename also needs to be filtered against directory traversal, e.g. with `basename($filename)`.
- Can we avoid overwriting the `$source` variable? IMO that makes the code quite confusing.

src/Api/Upload.php
```diff
@@ -0,0 +1,103 @@
...
+			rename($tmp, $source);
+		}
+
+		return $source;
+	}
+
+	/**
+	 * Clean up tmp directory of stale files
+	 */
+	public static function clear(): void
```
lukasbestle: Shouldn't this rather be `clean`? With `clear` I'd think that the folder is completely emptied.

src/Api/Upload.php
```diff
@@ -0,0 +1,103 @@
...
+	 * in the tmp directory and only returning the new
+	 * $source path to the tmp file once complete
+	 */
+	public static function chunk(
+		Api $api,
+		string $source,
+		string $filename
+	): string|null {
+		// if the file is uploaded in chunks
+		if ($isChunk = $api->requestHeaders('X-Upload-Chunk')) {
```
lukasbestle: The `X-` prefix for custom headers is [deprecated](https://datatracker.ietf.org/doc/html/rfc6648). I went looking for how the tus protocol does it and we could use it as inspiration:

- `Upload-Length` would carry the total file size. If present, we know that we need to engage chunk mode.
- `Upload-Offset` would carry the byte offset where the current chunk begins. We could use this for additional protections:
  - If the offset is 0 but the file with the provided unique ID already exists, fail (could occur if two clients generate the same random ID for whatever reason).
  - If the offset is non-0, compare it with the current file size. If they differ or if the file doesn't exist at all yet, we got the requests out of order, so fail.
- `Upload-Id` would carry the unique ID from the client.

I know this increases complexity, but I think it could well be worth it for added robustness.

config/api/routes/files.php
```diff
@@ -47,17 +51,28 @@
...
 					'source'   => $source,
 					'template' => $this->requestBody('template'),
 					'filename' => $filename
-				];
+				], true);
 
-				// move the source file from the temp dir
-				return $this->parent($path)->createFile($props, true);
+				// clean up tmp
+				F::remove($source);
```
lukasbestle: As the file is moved above, is there a case where the source file still exists at this point?

lukasbestle:
Adding the "two-person review" label as I realized we need to be super careful about the security implications here. There's a lot that could go horribly wrong if we miss anything. I think we are heading in a great direction though 👍

distantnative:
@lukasbestle makes sense. And as written in the initial post, I think we should have a few review rounds as well. Now getting the concept right, later adding unit tests etc. before final review.

src/Api/Upload.php
distantnative: I would like to away from the `content` folder. Cause our setup is that basically everything in `content` means something. We'd have to add in a few places code to ignore that directory etc. And I don't like to spread the feature code even more around the system. Could we rather use `site/cache` or so? `site` subfolders are also protected.

Not sure how to go about the file rules yet as they depend a lot on the filename, extension and already creating a File object.

config/api/routes/files.php
distantnative: Need to double check, but I think in case e.g. FileRules refused it. As the move/copy part only takes place after running the rules.

panel/src/helpers/upload.js
distantnative: Using https://developer.mozilla.org/en-US/docs/Web/API/Blob/slice

src/Api/Upload.php
lukasbestle: What about `site/temp`? It could also be useful for other things in the future and also devs could easily customize its path by overriding the root.

File rules: You are right, probably too complex. But then we should at least validate the full temp name (filename + unique ID) against our security blocklists.

src/Api/Upload.php
distantnative: How do we know which is the last chunk? Do we then have to measure the size of each chunk on the backend again?

src/Api/Upload.php
distantnative: We have to overwrite it at least at the end of the if block since we need to return `$source` either unaltered if it wasn't a chunk or altered after the last chunk.

src/Api/Upload.php
distantnative: `Str::slug($id, '', 'a-z')`?

src/Api/Upload.php
lukasbestle: If `Upload-Length === Upload-Offset + size of current chunk`, the file is complete. We could implement this by first appending the current chunk and then comparing the file size on disk with `Upload-Length`. If they are the same, we are done.

src/Api/Upload.php
lukasbestle: The end of the if block could IMO have its own return statement, it doesn't need to fall through.

afbora:
> We cannot delete the tmp file on cancel. But we clean up the files with a garbage collection.

@distantnative What if we upload to the temp folder, does it make sense? If upload successful, we can move the file to the model root, if not garbage collector take care.

distantnative:
@afbora That's what we are doing here. Just not the system temp folder as that doesn't persist across requests. So the first parts aren't there anymore when the last chunk is received.

src/Api/Upload.php
bastianallgeier: I know that quite some people started setting up a custom storage folder for all those "secret" writable files: 

- sessions
- cache
- accounts

I agree that the tmp folder should be in the site folder as well to keep it save. But then you would need to take care of file permissions in yet another spot. If we had such a single, writable storage folder inside of site, it would be a lot easier. Otherwise we would also need another check in the installer and the system view for the new tmp folder. 

But I guess moving to a unified storage folder is making this PR even more complex than it already is.

distantnative: @bastianallgeier that really sounds like it's own project – also big breaking change

distantnative: Or we just use a subdirectory in `site/cache` for the tmp chunk uploads.

src/Api/Upload.php
distantnative: Somehow, the sizes don't add up for me at the end:
987,268,918 is the `Upload-Length`, but the last chunk also sends an offset of 980,208,771 and has a length of 7,060,476 which is 987,269,247 which is larger than the whole file should be.

distantnative: https://github.com/getkirby/kirby/pull/6421/commits/bebeea708c95ac25b39decd10755b9398112f5b4#diff-3cd311828c9ff4560c3674cf73afee2c3e0d7d16a866e80630cde300c47a9aecR58-R60 also still fails for the 2nd chunk as `F::size()` returns 0.

src/Api/Upload.php
bastianallgeier: Oh, I quite like the idea.

tobimori:
Is it theoretically possible to add a config option that will limit the file size?

afbora:
I've found few issues, could you confirm that?

- File not uploaded without error or log (dialog closed and no file uploaded,) with big files. For me, ~100 MB success, ~300 MB fails.
- I'm getting `A file with the name "filename.zip" already exists` error when big files upload from textarea field

distantnative:
@afbora the PR currently is not functional as it's in the middle of implementing changes based on Lukas' feedback and some troubles I encountered (see our comments discussion). Sorry that you tried, but I think any upload fails currently that needs more than one chunk.

afbora:
@distantnative No problem. I'll test again after PR is ready.

src/Api/Upload.php
lukasbestle: Using the cache directory as an intermediate solution makes sense. Maybe we could then implement a new `temp` root separately for v5. That would then contain `cache`, `sessions` and `uploads`. Wouldn't put `accounts` there though as they are not ephemeral unlike the others.

panel/src/helpers/upload.js
lukasbestle: OK, so that reads the chunk into client memory. There seems to be a streaming API for files and for `fetch`, but in a quick search I couldn't find a example to plug them together.

Maybe we should limit the chunk size to something like 5 MB in the meantime in the `Upload::chunkSize()` method to avoid overloading the browser. Especially mobile devices may not always have enough memory to handle 100 MB or even larger requests.

src/Api/Upload.php
lukasbestle: Maybe we need `clearstatcache()` before checking the file sizes, especially when something was written to the file directly before. If it still doesn't line up, something is wrong either client-side or server-side, potentially producing corrupt files.

lukasbestle:
> Is it theoretically possible to add a config option that will limit the file size?

Good point. I think it would make sense to make this an option for the file blueprint.

distantnative:
@lukasbestle @tobimori are you thinking of something different than the existing accept maxsize option?

tobimori:
I didn't know there was an option already. Consider my request as solved 😀

panel/src/helpers/upload.js
distantnative: I've looks around and all implementations of chunks uploads, incl. tus (https://github.com/tus/tus-js-client/blob/main/lib/browser/sources/FileSource.js#L19), are using it the same way. I don't think we need to worry here and shouldn't overengineer it.

src/Api/Upload.php
distantnative: Referenced the wrong variable in a path 🙈

distantnative: I think I covered all suggestions here @lukasbestle, but please have a look yourself and mark as resolved if you agree.

lukasbestle:
Also forgot about `maxsize`. 😆

panel/src/helpers/upload.js
lukasbestle: That's great then!

distantnative:
@lukasbestle added some more validations – skipping MIME types (will still checked for the full file), but at least basic ones as allowed extensions.

src/Api/Upload.php
```diff
@@ -0,0 +1,170 @@
...
+				throw new Exception('Chunk offset ' . $offset . ' does not match the existing tmp upload file size of ' . F::size($tmp));
+			}
+		} elseif ($offset > 0) {
+			throw new NotFoundException('Chunk offset ' . $offset . ' for non-existing tmp file "' . $filename . '"');
+		}
+
+		// validate file
+		$parent = new Page(['slug' => 'tmp']);
+		$file   = new File(['filename' => $filename, 'parent' => $parent]);
+		FileRules::validFile($file, false);
```
distantnative: @lukasbestle Does it make sense to run these on every chunk request or should we limit it to only the first (file does not exist, offset === 0)?

src/Api/Upload.php
```diff
@@ -0,0 +1,170 @@
...
+			// get id of file upload and ensure it doesn't
+			// contain forbidden characters
+			$id = $api->requestHeaders('Upload-Id');
+			$id = Str::slug($id, '', 'a-z0-9');
+
+			// create path for file in tmp upload directory;
+			// append id while the file isn't completely uploaded yet
+			$filename  = basename($filename);
+			$tmp       = $dir . '/' . $filename;
+			$tmpWithId = $tmp . '.' . $id;
```
lukasbestle: Maybe we could **prepend** the random ID to make it less likely that it's a forbidden extension (by random accident). Also it will allow QuickLook on the tmp file as it will already be the correct extension.

src/Api/Upload.php
```diff
@@ -0,0 +1,170 @@
...
+			if ($offset !== F::size($tmp)) {
+				throw new Exception('Chunk offset ' . $offset . ' does not match the existing tmp upload file size of ' . F::size($tmp));
+			}
+		} elseif ($offset > 0) {
+			throw new NotFoundException('Chunk offset ' . $offset . ' for non-existing tmp file "' . $filename . '"');
+		}
+
+		// validate file
+		$parent = new Page(['slug' => 'tmp']);
+		$file   = new File(['filename' => $filename, 'parent' => $parent]);
```
lukasbestle: This should operate on the actual filename as it's primarily a security check.

```suggestion
		$file   = new File(['filename' => basename($tmp), 'parent' => $parent]);
```

panel/src/helpers/upload.js
lukasbestle: Maybe we could skip the slicing if it's just one part anyway. Could increase performance for the most common case of small files.

panel/src/helpers/upload.js
```diff
@@ -1,3 +1,45 @@
...
+				"Upload-Id": id
+			};
+		}
+
+		response = await upload(chunk, {
+			...params,
+			// calculate the total progress based on chunk progress
+			progress: (xhr, chunk, percent) => {
+				const total = (i + percent / 100) / parts;
+				params.progress(xhr, file, Math.ceil(total * 100));
```
lukasbestle: With large files, `Math.ceil` will immediately start with `1%` even though it may stay at 0.x% for quite a while. What about `Math.round()`?

src/Api/Upload.php
lukasbestle: I think that would be fine. As we only validate the filename and extension, we can assume that this check already happened for files that exist at the `$tmp` path.

lukasbestle:
About `maxsize`: We need to ensure that we check for that in three ways:

- The frontend should check for it before even uploading the first chunk (avoids user frustration). Maybe this already happens.
- The backend should also ensure it by checking the `Upload-Length` against it.
- Also there should be a validation for the actual uploaded chunk size. So the size of the current tmp file + the size of the temporary chunk file provided by PHP should be compared to `maxsize` before streaming over the chunk to our tmp file. Otherwise one could provide an `Upload-Length` with an allowed size (e.g. 5 MB) but send a chunk with 100 MB. The file would then already take up double that size on the server before Kirby finally validates the size in `$page->createFile()`.

src/Api/Upload.php
distantnative: I see your point. However, I thought it might actually be also beneficial to not keep the original extension for the tmp file but have our random ID. Wouldn't this minimize also some attack vectors where malicious files would need their file extension?

distantnative:
@lukasbestle I am not sure we can follow your suggestions for `maxsize`. This would require us parsing the blueprints on each of those requests (Panel view, chunk upload). And given that our blueprints currently are always fully resolved (sadly still), this could create real performance issues. Because of that, I think until we resolve blueprints more lazily, we have to stick to the (not ideal) behavior that the `accept` rules only get evaluated at the end of the full upload.

src/Api/Upload.php
distantnative: @lukasbestle do you have an idea why this change causes the following failures?

```
There was 1 failure:

1) Kirby\Api\UploadTest::testValidateChunkInvalidExtension
Failed asserting that exception of type "TypeError" matches expected exception "Kirby\Exception\InvalidArgumentException". Message was: "stream_copy_to_stream(): Argument #1 ($from) must be of type resource, bool given" at
/home/runner/work/kirby/kirby/src/Api/Upload.php:70
/home/runner/work/kirby/kirby/tests/Api/UploadTest.php:346
.

--

1 test triggered 1 PHP warning:

Error: 1) /home/runner/work/kirby/kirby/src/Api/Upload.php:69
fopen(/home/runner/work/kirby/kirby/tests/tmp/Api.Upload/a.php): Failed to open stream: No such file or directory

Triggered by:

* Kirby\Api\UploadTest::testValidateChunkInvalidExtension
  /home/runner/work/kirby/kirby/tests/Api/UploadTest.php:331
  ```

File: src/Api/Upload.php (hunk 13 above)
Comment (lukasbestle) [reply src/Api/Upload.php]: The issue is that the random ID is generated on the frontend and thus entirely controlled by a potential attacker anyway.

File: src/Api/Upload.php (hunk 14 above)
Comment (lukasbestle) [reply src/Api/Upload.php]: Does this change alone cause those failures? That's really weird as we are only doing a validation in this method and don't overwrite the value of `$tmp` that the calling method works with afterwards.

File: src/Api/Upload.php (hunk 13 above)
Comment (distantnative) [reply src/Api/Upload.php]: But we validate it before doing anything. So I think the attacker part here isn't the issue. How about we require the id to be a certain size (5-6 characters) which couldn't act as harmful file extension?

File: src/Api/Upload.php (hunk 14 above)
Comment (distantnative) [reply src/Api/Upload.php]: Yes, it did for me. Just this change. I can test it again later this week if you say it doesn't happen for you when just changing this.

**Comment by lukasbestle** [comment]
About `maxsize`: You are right that it won't work out to do the first check in the frontend as we would need to provide all possible maximum sizes for all allowed file blueprints before we even know if a file will be uploaded at all. However during an active upload, this is different: There I feel like the performance impact of checking the `maxsize` is negligible compared to the duration of the upload (unless the uploader is somewhere in Switzerland with a symmetric 1 Gbit/s connection 😆).

However with these checks we could prevent attackers from stuffing our custom temp dir with insanely large partially uploaded files that are so far prevented by the blueprint. At the moment, PHP will accept the file first, store it to the system temp folder and then Kirby decides not to copy it elsewhere. So the damage is limited to the `upload_max_filesize` for a brief moment in time (PHP auto-cleans uploaded files that were not copied during the request). With chunked uploads, an attacker could however initiate an upload of a 1 GB file and upload 999 MB of that. Those 999 MB will reside on the server until cleanup. This could lead to denial of service if done until the disk on the server or quota is full.

File: src/Api/Upload.php (hunk 14 above)
Comment (lukasbestle) [reply src/Api/Upload.php]: Can't test it right now, but I can debug this further.

File: src/Api/Upload.php (hunk 13 above)
Comment (lukasbestle) [reply src/Api/Upload.php]: > But we validate it before doing anything. So I think the attacker part here isn't the issue.

Yes, but that means that appending the random ID doesn't minimize an attack vector either.

So the order isn't about security, but increases reliability and also adds the benefit of better overview of the temp folder (also think file type icons and such).

> How about we require the id to be a certain size (5-6 characters) which couldn't act as harmful file extension?

With 5-6 chars, `.phtml` and `.phtml5` would still cause a failure when randomly generated (which is probably like winning in the lottery, but not impossible).

All in all, prepending the ID feels more robust to me. Let me know if I missed a reason for appending.

File: src/Api/Upload.php (hunk 14 above)
Comment (distantnative) [reply src/Api/Upload.php]: Worked on the maxsize and shuffled things around. Now it uses `basename($tmp)` and works.

**Comment by distantnative** [comment]
@lukasbestle I added a check on the backend for `maxsize`: both, checking the total length from the header as well as existing tmp plus chunk size.

File: src/Api/Upload.php [hunk 16]
```diff
@@ -0,0 +1,207 @@
...
+		} elseif ($offset > 0) {
+			// sent chunk is not expected to be the first part,
+			// but no tmp file to continue exists yet
+			throw new NotFoundException('Chunk offset ' . $offset . ' for non-existing tmp file: ' . $name);
+		} elseif ($offset === 0) {
+			// validate file (extension, name) for first chunk;
+			// will also be validate again by `$model->createFile()`
+			// when completely uploaded
+			FileRules::validFile($file, false);
+		}
```
lukasbestle: To get rid of the `elseif`s:

```suggestion

			// existing file is valid, done validating
			return;
		}

		if ($offset === 0) {
			// validate file (extension, name) for first chunk;
			// will also be validate again by `$model->createFile()`
			// when completely uploaded
			FileRules::validFile($file, false);
		}

		throw new NotFoundException('Chunk offset ' . $offset . ' for non-existing tmp file: ' . $name);
```

config/api/routes/files.php
```diff
@@ -47,17 +49,26 @@
 			// move_uploaded_file() not working with unit test
 			// @codeCoverageIgnoreStart
 			return $this->upload(function ($source, $filename) use ($path) {
-				$props = [
+				// check if upload is  sent in chunks and handle them
```
lukasbestle:
```suggestion
				// check if upload is sent in chunks and handle them
```

src/Api/Upload.php
```diff
@@ -0,0 +1,207 @@
...
+			// prefix with id while file isn't completely uploaded yet
+			$filename  = basename($filename);
+			$tmp       = $dir . '/' . $filename;
+			$tmpWithId = $dir . '/' . $id . '-' . $filename;
+
+			// get byte offset of current chunk for total file length
+			$offset = (int)$api->requestHeaders('Upload-Offset');
+
+			// validate various aspects of the request
+			// to ensure the chunk isn't try to do malicious actions
```
lukasbestle:
```suggestion
			// to ensure the chunk isn't trying to do malicious actions
```

tests/Api/UploadTest.php
```diff
@@ -0,0 +1,418 @@
...
+		$_SERVER['HTTP_CF_CONNECTING_IP'] = '1.1.1.1';
+
+		$this->assertSame(
+			(int)floor(100 * 1024 * 1024 * 0.95),
+			Upload::chunkSize()
+		);
+
+		ini_restore('upload_max_filesize');
+		ini_restore('post_max_size');
+		unset($_SERVER['HTTP_CF_CONNECTING_IP']);
```
lukasbestle: Those should be in `tearDown()`, otherwise they won't be reset if the test fails, thus changing the state for following tests.

lukasbestle (commented):
Awesome, I feel like the logic is really solid now. My remaining comments are just minor fixes and refactorings.

@afbora I think now is the time when you can test it in detail as behavior changes are not too be expected anymore.

src/Api/Upload.php
distantnative: Ended up fixing it differently as this solution would always throw the last exception for valid chunks. But managed to untangle the if-elseifs a bit.

lukasbestle (dismissed):
Oh yes, that's much better. 👍

Code-wise I think this is great now. To be on the safe side, it would be good to get another code review by someone uninvolved (@bastianallgeier if you have the time) and also @afbora's testing skills will help us a lot.

distantnative:
Would be great indeed. I've rebased and squashed the commits already.

afbora:
@distantnative I'm getting `A file with the name "filename.zip" already exists` error when big files upload from textarea field. Same file uploads without error from files section.

distantnative:
@afbora good catch. Can reproduce myself. Somehow already after the first chunk they get moved to the content folder. But need a fresher mind to investigate why.

lukasbestle:
Maybe a different route processes these so the `Upload-Length` header is ignored and the new chunk logic on the backend doesn't get executed?

distantnative:
Indeed it is. So I am thinking to actually move the `Upload::chunk()` call here: https://github.com/getkirby/kirby/blob/main/src/Api/Api.php#L697

So any place using `$api->upload()` would work with it.

Furthermore, I think I would like to move most of the Code from `Api::upload()` to the new `Upload` class. I think that could als help us to simplify that code into several single-purpose methods, e.g. `Upload::response()`, `Upload::error()`, `Upload::filename()` ...

lukasbestle:
Makes sense to me. The `Api::upload()` method is indeed a spaghetti monster at the moment.

Behavior would be that `Api::upload()` would check for chunks in the place you linked to and only call the callback on the last chunk when the upload is complete, right?

distantnative:
Yes, exactly. Have a version working locally. It really helps breaking this up but also having it all in the one class then.

Now only need to add unit tests to all new parts of the `Upload` class. Bit of work but also is nice cause we'll cover most of this with proper unit tests now.

src/Api/Upload.php
```diff
@@ -0,0 +1,428 @@
...
+ * The Upload class handles file uploads in the
+ * context of the API. It adds support for chunked
+ * uploads.
+ *
+ * @package   Kirby Api
+ * @author    Nico Hoffmann <nico@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @since     4.3.0
```
lukasbestle:
```suggestion
 * @since     4.4.0
```

src/Panel/Panel.php
```diff
@@ -137,6 +138,21 @@ public static function firewall(
 		return true;
 	}
 
+	/**
+	 * Garbage collection which runs with a probability
+	 * of 10% on each Panel request
+	 *
+	 * @since 4.3.0
```
lukasbestle:
```suggestion
	 * @since 4.4.0
```

lukasbestle (dismissed):
Code-wise this is absolutely brilliant now. I love how clean the new `Upload` class is. 👍

This makes the PR ready for another round of manual testing. 🤞

src/Api/Upload.php
```diff
@@ -0,0 +1,428 @@
...
+		$file = new File([
+			'parent'   => new Page(['slug' => 'tmp']),
+			'filename' => $filename = basename($tmp),
+			'template' => $template
+		]);
+
+		// if the blueprint `maxsize` option is set,
+		// ensure that the total size communicated in the header
+		// as well as the current tmp size after adding this chunk
+		// does not exceed the max limit
```
distantnative:
```suggestion
		// do not exceed the max limit
```

panel/src/helpers/upload.js
```diff
@@ -96,4 +98,46 @@ export default async (file, params) => {
...
+				"Upload-Offset": start,
+				"Upload-Id": id
+			};
+		}
+
+		response = await upload(chunk, {
+			...params,
+			// calculate the total progress based on chunk progress
+			progress: (xhr, chunk, percent) => {
+				const total = (i + percent / 100) / parts;
```
lukasbestle: This calculation causes the progress to jump quickly at the end if the last chunk is smaller than the rest. E.g. file is 91 MB and chunk size is 10 MB. Then chunk 10/10 is only 1 MB, so the last 10% of progress pass 10 times faster.

Maybe we can implement it as transmitted size (start offset + current chunk size * percent) compared to the total file size.

afbora:
Works great for me. I couldn't see any issue 🎉

panel/src/helpers/upload.js
distantnative: @lukasbestle 

```suggestion
				const offset = i * size;
				const current = size * (percent / 100);
				const total = (offset + current) / file.size;
```

?

distantnative:
I think I have encountered an issue: Uploading a large file and canceling while it's uploading, it still shows up in the file list after a reload.

panel/src/helpers/upload.js
lukasbestle: I think instead of `size` (which again does not respect that the last chunk may be smaller) it should be `chunk.size`. And instead of `offset` I think it should be `start`. Rest looks great.

distantnative:
Turns out: When one cancels the upload dialog, all running uploads still run and finish in the background. This is not new in this PR but a lot more visible/likely to happen with a different expectation by the user.

distantnative:
@lukasbestle @afbora Added a way to abort ongoing upload requests. Could you please review the last commit and test it out?

afbora:
@distantnative Still works great for me 👍

panel/src/helpers/upload.js
```diff
@@ -94,6 +97,61 @@ export default async (file, params) => {
...
+				"Upload-Offset": start,
+				"Upload-Id": id
+			};
+		}
+
+		response = await upload(chunk, {
+			...params,
+			// calculate the total progress based on chunk progress
+			progress: (xhr, chunk, percent) => {
+				const start = i * size;
```
lukasbestle: Is this line needed? We already set `start` to the same value in line 127 and this closure should be able to access that, no?

distantnative: Good catch

lukasbestle (approved):
Works great for me as well 👍

PR #6430, opened 2024-05-01
Image blocks: add pattern background

### LINKED ISSUE #6378: White PNG/SVG images are not visible in the panel
- Add an `image` with the block or layout editor
- Upload a PNG/SVG with white fill and transparent background
- Only a white area is visible, which is irritating because I can't see if/which image is hidden behind it.

image-1

Wouldn't it be great to display the pattern in the background by default, as adapted by the cards/cardlets? 

It also would be great to be able to change the background using the same principle in the blueprint:
https://getkirby.com/docs/reference/panel/sections/files#preview-images__back

I solved it with a simple CSS class in `custom-panel.css`:

```
.k-block-type-image-auto {
    background: var(--pattern);
}
```

image-2

I am aware that I can customize the block type. But there will be many (new) users who are happy about such useful little details if they are already part of the core functions of Kirby. 

It is not a bug, but a cosmetic improvement for the panel. 

**Kirby 4.1.2**

distantnative:
### Summary of changes
- Image and gallery block now feature a little background color selector dropdown (selection gets stored in localStorage)
- `k-block-figure`: New `back` option to set the background (pass pure CSS value)
- New `k-block-figure-caption` that adapts its color based on the background (to also work on dark backgrounds)
- New `k-block-background-dropdown`
- `k-color-frame` can now also handle patterns as pseudo-color

### Enhancements
- Image/Gallery blocks: added selector to change background

https://github.com/user-attachments/assets/b355918e-9d34-42a5-8184-a749723cfa3c

### For review team


tobimori:
Maybe it'd be possible to toggle this via a blueprint option? I think it's a good thing, especially with the dark mode coming up, but most of my blocks don't need it as you figured.

bastianallgeier:
I also think that this could/should be a toggle. Maybe to even toggle between white/black/pattern.

distantnative:
@bastianallgeier like an actual toggle field? The issue right now is that I wanted to rather add it as a fieldset blueprint option. However, I realized that fieldsets do no allow passing any custom options: https://github.com/getkirby/kirby/blob/main/src/Cms/Fieldset.php#L204-L219

bastianallgeier:
@distantnative I was thinking of a toggle in the block preview component. I would not connect it to the blueprint

distantnative:
@bastianallgeier but if it's not a fixed state, one would have to change the toggle e.g. for a white transparent graphic to pattern each time loading the view. Or would we try to store that in localStorage?

bastianallgeier:
Do you think it's going to get too messy to put it in localStorage? I think that would be quite nice as a solution.

distantnative:
I think the localStorage part could work. Just thinking of the UI of this toggle, won't be so easy as it would have to float on top of the image, or what do you have in mind?


https://github.com/getkirby/kirby/assets/3788865/a027bded-950c-4001-a44c-8a5745b9f9f6

bastianallgeier:
I think it really looks good already. I wonder if we should use the background on the entire width or really only for the image. We could maybe reuse https://lab.getkirby.com/public/lab/components/frames/4_color to create a little square button in the bottom right corner instead of the text button. Maybe together with a dropdown arrow.

tobimori:
I would go with the entire width. Looks better.

distantnative:
@bastianallgeier Implemented a visual selector as suggested.

Also coloring now the full block which I think looks best.

panel/src/components/Forms/Blocks/Elements/BlockBackgroundDropdown.vue
```diff
@@ -12,13 +12,12 @@
...
 			:options="[
-				{ text: 'White', value: 'white', click: 'var(--color-white)' },
+				{ text: 'White', click: 'var(--color-white)' },
 				{
 					text: 'Pattern (light)',
-					value: 'pattern',
 					click: 'var(--pattern-light)'
 				},
-				{ text: 'Pattern (dark)', value: 'pattern', click: 'var(--pattern)' }
+				{ text: 'Pattern (dark)', click: 'var(--pattern)' }
```
bastianallgeier: I guess we still need translations here, right?

bastianallgeier:
I've pushed a suggestion for a slight design change for the button …

Screenshot 2024-07-25 at 16 23 40

panel/src/components/Forms/Blocks/Elements/BlockBackgroundDropdown.vue
distantnative: Added

bastianallgeier (approved):
It's ready to go in my opinion.

PR #6432, opened 2024-05-02
Make `contentToken` independent from `num` in root

### LINKED ISSUE #5102: Reordering subpages generates duplicate thumbnails
## Description

Reordering subpages in the panel is resulting in regeneration of all thumbnails of all pages, that received a new sorting number.

## To reproduce

1. Download starterkit. 
2. Visit all the Photography subpages in a browser
3. Move one of the Photography pages to the first position (in the panel)
4. Re-visit all the Photography pages
5. Now all the thumbnails are re-generated with a new folder name.

**Expected behavior**  

The thumbnails should not be regenerated, when reordering subpages.

**Screenshots**  

Here is a screenshot of the waterfall page. The thumbnails were generated once at 10:41 during the first visit. And after the reordering at 10:44 during the second visit.

Thumbnails

## Your setup

**Kirby Version**  

Tested in 3.9.1 and 3.9.2. Happened also in previous versions.

**Your system**

Tested on local server and productive debian systems. Happens with GD or ImageMagick. Happens with PHP 8.1.0, PHP 8.1.13, and also 7.4.33.

**Additional context**

For multiple Kirby sites we have running, this issue repeatedly brings the according server to its knees. Especially instances, where the template shows a gallery of images of all the subpages. With GD thumbnail creation, the Debian server would regulary be unresponsive. With ImageMagick, the server does not crash, but CPU and Memory usage go to 100% for a few minutes.
Needless to say, for these usecases we have, this bug is very severe.

Also, each time this happens, the used storage space is increased dramatically. From the viewpoint of the first new visitor of a page, the page load time is very bad.

We think, that many users might be affected, especially those, where blog-style subpages change frequently.

distantnative:
## This PR …

See discussion at #5102.
Please have a close look at it. I think I implemented it as discussed, but e.g. not sure where best to add a unit test.

### Enhancements
- Thumbnails don't need to be regenerated when page sorting changes

### Breaking changes
- All thumbnails will regenerate

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -435,10 +435,11 @@ class App
 	 */
 	public function contentToken(mixed $model, string $value): string
 	{
-		$default = match (method_exists($model, 'root')) {
-			true  => $model->root(),
-			false => $this->root('content')
-		};
+		$default = $this->root('content');
+
+		if (method_exists($model, 'id') === true) {
+			$default .= '/' . $model->id();
+		}
 
 		$salt = $this->option('content.salt', $default);
 
--- a/tests/Cms/App/AppTest.php
+++ b/tests/Cms/App/AppTest.php
@@ -274,7 +274,7 @@ class AppTest extends TestCase
 			]
 		]);
 		$this->assertSame(hash_hmac('sha1', 'test', '/dev/null/content'), $app->contentToken('model', 'test'));
-		$this->assertSame(hash_hmac('sha1', 'test', '/dev/null'), $app->contentToken($app, 'test'));
+		$this->assertSame(hash_hmac('sha1', 'test', '/dev/null/content'), $app->contentToken($app, 'test'));
 
 		// with custom static salt
 		$app = new App([
--- a/tests/Cms/Pages/PageTest.php
+++ b/tests/Cms/Pages/PageTest.php
@@ -636,7 +636,11 @@ class PageTest extends TestCase
 		]);
 
 		if ($draft === true && $expected !== null) {
-			$expected = str_replace('{token}', 'token=' . hash_hmac('sha1', $page->id() . $page->template(), $page->root()), $expected);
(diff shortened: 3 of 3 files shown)


bastianallgeier:
It looks good to me. I guess the impact is that all thumbnails need to be rerendered though. That's always a bit of a pain in the ass because it can affect big installations quite heavily.

distantnative:
If we want to solve https://github.com/getkirby/kirby/issues/5102 I don't think there is away around it. Any solution not having the num considered will result in different hashes.

But we could consider it as somewhat breaking change and move it to v5.

bastianallgeier:
I feel like we have to move it to v5, yes. It's too drastic for a ~patch~  minor release.

lukasbestle:
I agree. Change looks good, but would be safer in 5.0.

distantnative:
I don't mind at all moving it 👍

(Though 4.3 isn't a patch release 😜)

bastianallgeier:
@distantnative I don't know what you are talking about 🙈🤣

lukasbestle:
It's the natural evolution of our versioning scheme: Patch is the second digit, minor the first and the major version is invisible. Who needs major versions anyway. 😅

bastianallgeier (approved):
LGTM! I think we can ignore the CI issue. The frontend stuff is acting up somehow. But it's not related to the PR.

bastianallgeier:
@distantnative I just saw the needs:test label after my approval. Why do you think we will need additional tests?

distantnative:
@bastianallgeier thought we might add a test for the reported issue: that thumbs don't regenerate just because the page got sorted. Though will be hard to write that test I fear.

bastianallgeier:
Hm, from my perspective this is the obvious result of no longer including the number. I'm really not sure if we need that at all.

PR #6435, opened 2024-05-03
Remove null headers from API calls

bastianallgeier:
### Fixes

- Headers with null as value are no longer added to API requests. For some weird reason, JS will turn them into 'NULL' strings which is pretty annoying to work with in the backend. 

### Breaking changes

None

### For review team

This is another prep PR for changes. I'm trying to extract all the no-brainers that we can already add to v4

### DIFF
--- a/panel/src/api/request.js
+++ b/panel/src/api/request.js
@@ -29,11 +29,11 @@ export default (api) => {
 		}
 
 		// remove null headers
-		Object.keys(options.headers).forEach((key) => {
+		for (const key in options.headers) {
 			if (options.headers[key] === null) {
 				delete options.headers[key];
 			}
-		});
+		}
 
 		// build the request URL
 		options.url = rtrim(api.endpoint, "/") + "/" + ltrim(path, "/");
--- a/panel/src/api/request.js
+++ b/panel/src/api/request.js
@@ -28,6 +28,13 @@ export default (api) => {
 			options.method = "POST";
 		}
 
+		// remove null headers
+		Object.keys(options.headers).forEach((key) => {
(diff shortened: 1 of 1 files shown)


distantnative:
Could this be a breaking change for people actually checking for those `NULl` headers on the backend?

bastianallgeier:
You mean if someone tried to work around broken values in API calls? Hm, maybe but I wonder if this would still count as a breaking change. 

Because the backend literally receives `"NULL"` So you have to check for something like 

```php
if ($languageHeader === 'NULL') {

}
```

distantnative:
Yes, cause if before the headers would not actually be received as `null` but `"NULL"`, people might have adapted to it to check if a header was set or not (if their JS part always adds the header, just sets it as `null` when empty).

And do we maybe have the same problem with `undefined`?

bastianallgeier:
It seems like undefined is properly ignored. I also don't mind to put this into the alpha

distantnative:
I have no set opinion: technically it could be a breaking change, practically it probably doesn't affect any user.

panel/src/api/request.js
```diff
@@ -28,6 +28,13 @@ export default (api) => {
 			options.method = "POST";
 		}
 
+		// remove null headers
+		Object.keys(options.headers).forEach((key) => {
+			if (options.headers[key] === null) {
+				delete options.headers[key];
+			}
+		});
```
distantnative: Can I suggest

```js
for (const key in options.headers)
```
?

bastianallgeier:
I changed the base to add this to the alpha, just to be sure.

distantnative (approved):

PR #6442, opened 2024-05-06
Changes 3: New Version class

bastianallgeier:
## This PR …

- [x] 🚨 Merge first: #6436
- [x] 🚨 Merge first: #6439

### Refactoring

- New `Version` class which inherits most of the logic from the `ContentStorage` handler and will eventually replace it. 

### Outlook

- This PR will be followed by a later PR to improve the `Version` class and add additional unit tests (https://github.com/getkirby/kirby/pull/6450) The steps are separated to make sure that all other unit tests always keep passing. It's sometimes a bit of a chicken and egg situation here.
- Once the `ContentStorageHandler` methods accept a `Language` object instead of a string (https://github.com/getkirby/kirby/pull/6448) the `Version` class also needs to pass a `Language` object instead of language codes. It will therefor get a new protected `Version::language()` method that is mostly adapted from `ContentStorage::language()` but will always return a `Language` object - even in a single-language installation. There's a new `Language::single()` method coming up in https://github.com/getkirby/kirby/pull/6448 for that. 
- The `Version` class will also inherit a few more specific logic methods from `ContentStorage` in preparation of removing `ContentStorage` entirely 
  - `::deleteLanguage`
  - `::touchLanguage`
  - `::contentFile`
- `Version::modified` will later return null if the version does not exist.
- `Version::ensure` will later use `ContentStorageHandler::ensure` https://github.com/getkirby/kirby/pull/6457

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -65,12 +65,12 @@ class Version
 
 		// delete a single language
 		$this->model->storage()->delete($this->id, $language);
-		return;
+
 	}
 
 	/**
 	 * Ensure that the version exists and otherwise
-     * throw an exception
+	 * throw an exception
 	 *
 	 * @throws \Kirby\Exception\NotFoundException if the version does not exist
 	 */
@@ -130,7 +130,7 @@ class Version
 
 	/**
 	 * Returns the stored content fields
-     *
+	 *
 	 * @param string $lang Code `'default'` in a single-lang installation
 	 * @return array<string, string>
 	 */
@@ -172,8 +172,8 @@ class Version
 	/**
 	 * Updates the content fields of an existing version
 	 *
-	 * @param string $lang Code `'default'` in a single-lang installation
 	 * @param array<string, string> $fields Content fields
+	 * @param string $lang Code `'default'` in a single-lang installation
 	 *
 	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
 	 */
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -3,6 +3,7 @@
 namespace Kirby\Content;
 
 use Kirby\Cms\ModelWithContent;
+use Kirby\Exception\NotFoundException;
 
 /**
  * The Version class handles all actions for a single
@@ -39,7 +40,7 @@ class Version
 	/**
 	 * Creates a new version for the given language
 	 */
-	public function create(string $language, array $fields): void
+	public function create(array $fields, string $language = 'default'): void
 	{
 		$this->model->storage()->create($this->id, $language, $fields);
 	}
@@ -49,15 +50,35 @@ class Version
 	 */
 	public function delete(string|null $language = null): void
 	{
-		// delete a single language
-		if ($language !== null) {
-			$this->model->storage()->delete($this->id, $language);
+		// delete all languages
+		if ($language === null) {
+			foreach ($this->model->kirby()->languages() as $language) {
+				$this->model->storage()->delete($this->id, $language->code());
+			}
+		}
+
+		// delete the default language in single-language mode
+		if ($this->model->kirby()->multilang() === false) {
+			$this->model->storage()->delete($this->id, 'default');
 			return;
 		}
 
-		// delete all languages
-		foreach ($this->model->kirby()->languages() as $language) {
-			$this->model->storage()->delete($this->id, $language->code());
+		// delete a single language
+		$this->model->storage()->delete($this->id, $language);
+		return;
+	}
+
+	/**
+	 * Ensure that the version exists and otherwise
+     * throw an exception
+	 *
+	 * @throws \Kirby\Exception\NotFoundException if the version does not exist
+	 */
+	public function ensure(
+		string $language = 'default'
+	): void {
+		if ($this->exists($language) !== true) {
+			throw new NotFoundException('Version "' . $this->id . ' (' . $language . ')" does not already exist');
 		}
 	}
 
@@ -85,8 +106,25 @@ class Version
 		return $this->model;
 	}
 
+	/**
+	 * Returns the modification timestamp of a version
+	 * if it exists
+	 *
+	 * @param string $lang Code `'default'` in a single-lang installation
+	 */
+	public function modified(
+		string $language = 'default'
+	): int|null {
+		$this->ensure($language);
+		return $this->model->storage()->modified($this->id, $language);
+	}
+
+	/**
+	 * Moves the version to a new language and/or version
+	 */
 	public function move(string $fromLanguage, VersionId $toVersionId, string $toLanguage): void
 	{
+		$this->ensure($fromLanguage);
 		$this->model->storage()->move($this->id, $fromLanguage, $toVersionId, $toLanguage);
 	}
 
@@ -98,6 +136,7 @@ class Version
 	 */
 	public function read(string $language = 'default'): array
 	{
+		$this->ensure($language);
 		return $this->model->storage()->read($this->id, $language);
 	}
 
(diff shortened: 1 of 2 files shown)


src/Content/Version.php
```diff
@@ -0,0 +1,185 @@
...
+		protected VersionId $id
+	) {
+	}
+
+	/**
+	 * Returns a Content object for the given language
+	 */
+	public function content(string $language = 'default'): Content
+	{
+		return new Content(
```
lukasbestle, line 34: I assume `Version` will later cache the content object for each language, right?

src/Content/Version.php
```diff
@@ -0,0 +1,185 @@
...
+			parent: $this->model,
+			data:   $this->model->storage()->read($this->id, $language),
+		);
+	}
+
+	/**
+	 * Creates a new version for the given language
+	 */
+	public function create(array $fields, string $language = 'default'): void
```
lukasbestle, line 43: With our current translation architecture (automatic field fallbacks to the default language) we can only lock all language versions together to the same user when editing.

So e.g. when creating a `changes` version from an existing `published` version, the `changes` version must always contain all languages that the `published` version has.

I don't fully understand how this `$version->create()` instance method is going to be used. To ensure that the version is complete and consistent, I feel like creation needs to happen atomically, e.g. with a static `Version::create(ModelWithContent $model, VersionId $id)` method that then takes care to take the content fields from the right source (changes takes content from published, drafts start with empty content, published cannot be created standalone) and to ensure all languages.

src/Content/Version.php
```diff
@@ -0,0 +1,185 @@
...
+		}
+
+		// delete the default language in single-language mode
+		if ($this->model->kirby()->multilang() === false) {
+			$this->model->storage()->delete($this->id, 'default');
+			return;
+		}
+
+		// delete a single language
+		$this->model->storage()->delete($this->id, $language);
```
lukasbestle, line 67: What is the use case for deleting a single language? Do we already support that in v4?

src/Content/Version.php
bastianallgeier: Yes, exactly. But this isn't implemented in the other PRs so far. It's only on the todo list in Notion

src/Content/Version.php
bastianallgeier: You are right. This was implemented before we talked about returning null for non-existing versions. That would have meant that the version instance could be initalized and the creat method could have been called like that before it exists.

src/Content/Version.php
bastianallgeier: Tbh I mostly just moved over what we had in ContentStorage here. I think you are right and we don't need it so far. This would also clean up the method some more. I'm not super happy about the ambiguous usage depending on what's passed.

src/Content/Version.php
lukasbestle: How do we proceed here? I guess we will produce a lot of merge conflicts if we change it here. So maybe let's continue and add it to the todo list?

src/Content/Version.php
lukasbestle: I'll check the code and concept history again. If there is no reason, I'll remove it. We can still add it back if we ever need it.

src/Content/Version.php
bastianallgeier: Yes, sounds good to me. We already have a todo list item in Notion. I think it makes sense to fix it in a separate PR when we get to that point.

src/Content/Version.php
bastianallgeier: Did you have time to check it already or should I have another look at it?

src/Content/Version.php
lukasbestle: TBH I can't find a todo list item for this. Maybe I missed it though. :)

src/Content/Version.php
lukasbestle: Checked the concept and code again. What I could find out:

- In the concept, we only had the handlers, not `ContentStorage`. There we only intended to support deleting a specific language. A `null` param to the `$handler->delete()` method for looping over all languages was not supported.
- Reasoning as far as I remember was again to prepare for a future where languages are independent. So you may then want to delete a single language.
- The `ContentStorage` class introduced the nullable `$lang` param. But if I understand the `ContentStorage` logic correctly, it does not mean "all languages" but instead "default language". The `null` is passed to the `$contentStorage->language()` method and that passes it to `$app->language(null)` in multi-lang or just returns `default` in single-lang.

**TL;DR:** If I understand it correctly, the nullable param as you have implemented it here is a new concept we didn't have in the current code base.

`Suggestion`: As the `Version` class is internal anyway, we only need to support what we need in the core. And so far that's only "all languages" (correct?). "Specific language" starts getting relevant with the new translation concept only.

src/Content/Version.php
bastianallgeier: Screenshot 2024-06-04 at 09 15 47

src/Content/Version.php
bastianallgeier: I think your suggestion makes a lot of sense. Can we also handle it in a way that we combine it with a later refactoring after all the other PRs and add a todo for it to avoid a lot of rebases?

src/Content/Version.php
lukasbestle: Added a link to this discussion to the todo item so we can come back to it later.

src/Content/Version.php
lukasbestle: Added it to the todo list

tests/Content/VersionTest.php
```diff
@@ -0,0 +1,180 @@
...
+
+		$version->create([
+			'title' => 'Test'
+		]);
+
+		$this->assertSame('Test', $version->read()['title']);
+
+		$version->create([
+			'title' => 'Updated Title'
+		]);
```
lukasbestle: Shouldn't this test use the `update()` method?

tests/Content/VersionTest.php
```diff
@@ -0,0 +1,180 @@
...
+	 * @covers ::create
+	 */
+	public function testCreate(): void
+	{
+		$version = new Version(
+			model: $this->model,
+			id: VersionId::published()
+		);
+
+		$this->assertFalse($version->exists());
```
lukasbestle, line 47: These kinds of checks make the tests depend on methods of the class that aren't tested. This could result in false-positive test results if e.g. the `exists()` method uses an incorrect data source that happens to lead to the correct results in our test cases.

With the current simple implementation (mostly pass-throughs to the storage handler), a real-world bug is pretty unlikely. However it could make the tests more meaningful to use more low-level checks e.g. with `assertFileExists()`/`assertFileNotExists()` that actually verify what happens.

For the `testExists()` that applies the other way around as well.

lukasbestle (commented):
Added more context and links to our todo list for later. I agree it makes sense to merge this as is to avoid merge conflicts and endless rebasing. The class is not fully there yet, but we can and will improve it later in the process.

Added some comments on the tests. Same applies here: You don't have to change anything right now. I'd also be totally fine with merging the PR and adding those to the todo list for later.

tests/Content/VersionTest.php
bastianallgeier: Ok, that's really stupid. I just fixed it.

tests/Content/VersionTest.php
bastianallgeier: You are totally right about this. I would love to fix this in the version-class-improvements branch though because I've refactored the unit tests there to properly test single and multi language installations side by side. 

I've tried to already refactor the unit tests in this branch to check for existing files, but there's no proper way to do that for the multi-lang checks without those changes from the later branch.

lukasbestle (approved):
Makes sense. Could you add it to the todo list please?

tests/Content/VersionTest.php
bastianallgeier: Added it to the todo list.

PR #6447, opened 2024-05-12
Item: improve title and info wrapping

distantnative:
### Enhancements
- Sections: improved title and info wrapping

### For review team

### DIFF
--- a/panel/lab/components/item/1_list/index.vue
+++ b/panel/lab/components/item/1_list/index.vue
@@ -77,5 +77,20 @@
 				theme="disabled"
 			/>
 		</k-lab-example>
+		<k-lab-example label="Wrap for long title/info">
+			<k-item
+				:buttons="[
+					{
+						icon: 'edit'
+					},
+					{
+						icon: 'trash'
+					}
+				]"
+				:image="{ src: 'https://picsum.photos/800/600' }"
+				info="With some more information than usual"
+				text="This is a nice item with a rather long title that should wrap"
+			/>
+		</k-lab-example>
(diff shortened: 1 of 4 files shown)


bastianallgeier:
Just a few thoughts: 

- How does it look like when the image is not cropped? 
- How does it look if one item in the list is wrapping, but the others are not? Is it weird if the height is different? 
- The info text does not seem to wrap in the cardlets component

distantnative:
> How does it look like when the image is not cropped?

I think fine.

Screenshot 2024-05-17 at 16 52 55
Screenshot 2024-05-17 at 16 53 28

> How does it look if one item in the list is wrapping, but the others are not? Is it weird if the height is different?

Not super pretty, but I don't think there is a way that looks super pretty and still allows the same utility of being able to read all the information for long titles/info. I think it's a very good compromise that looks ok even in those extreme cases. 

Screenshot 2024-05-17 at 16 54 57

> The info text does not seem to wrap in the cardlets component

Those are still like before. The title wraps, the info not. If you think it should here as well, we can change that, sure.

bastianallgeier:
Uhhh, I just thought that we could maybe use grid-auto-row for the list items. I'm not entirely sure if it really works, but normally it takes care of always setting all items to the same height. That could clean up the list quite nicely. 

The images look great indeed.

distantnative:
@bastianallgeier I looked into `grid-auto-rows` but I don't understand (or could make it happen when trying) how this should work.

bastianallgeier (approved):

PR #6448, opened 2024-05-17
Changes 4: Refactor PlainTextContentStorageHandler to use Language object

bastianallgeier:
## This PR …

- [x] 🚨 Merge first: #6436
- [x] 🚨 Merge first: #6439
- [x] 🚨 Merge first: #6442

### Refactoring

- New `Language::single()` method to create a Language placeholder object in single language installations
- Use full language objects in `ContentStorageHandler` and `PlainTextContentStorageHandler` methods
- Refactor `ContentStorage` to always provide a full language object to the storage methods. This is only temporary to keep all the unit tests working. `ContentStorage` will be removed later, as planned.

### Outlook

- The `PlainTextContentStorageHandler` class will no longer implement the interface but extend the abstract `ContentStorageHandler` class. This change will also remove the `__construct` method, which will then be implemented by the abstract. https://github.com/getkirby/kirby/pull/6449
- The new `::write` method will later be moved to the `ContentStorageHandler` class as an abstract method. `ContentStorageHandler` will then also get a non-abstract update and create method, which will use the new write method by default. This will add the option for storage handlers to only modify the write method if no further logic is needed for create and update. https://github.com/getkirby/kirby/pull/6457

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -297,6 +297,7 @@ class Language implements Stringable
 
 	/**
 	 * Checks if this is the single language object
+	 * @internal
 	 */
 	public function isSingle(): bool
 	{
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -45,6 +45,7 @@ class Language implements Stringable
 	protected string $direction;
 	protected array $locale;
 	protected string $name;
+	protected bool $single;
 	protected array $slugs;
 	protected array $smartypants;
 	protected array $translations;
@@ -64,6 +65,7 @@ class Language implements Stringable
 		$this->default      = ($props['default'] ?? false) === true;
 		$this->direction    = ($props['direction'] ?? null) === 'rtl' ? 'rtl' : 'ltr';
 		$this->name         = trim($props['name'] ?? $this->code);
+		$this->single       = $props['single'] ?? false;
 		$this->slugs        = $props['slugs'] ?? [];
 		$this->smartypants  = $props['smartypants'] ?? [];
 		$this->translations = $props['translations'] ?? [];
@@ -177,8 +179,8 @@ class Language implements Stringable
 		if ($languages->count() === 0) {
 			foreach ($kirby->models() as $model) {
 				$model->storage()->convertLanguage(
-					'default',
-					$language->code()
+					Language::single(),
+					$language
 				);
 			}
 		}
@@ -225,7 +227,7 @@ class Language implements Stringable
 
 		foreach ($kirby->models() as $model) {
 			if ($this->isLast() === true) {
-				$model->storage()->convertLanguage($code, 'default');
+				$model->storage()->convertLanguage($this, Language::single());
 			} else {
 				$model->storage()->deleteLanguage($code);
 			}
@@ -273,7 +275,7 @@ class Language implements Stringable
 	public function isDeletable(): bool
 	{
 		// the default language can only be deleted if it's the last
-		if ($this->isDefault() === true && $this->isLast() === false) {
+		if ($this->isSingle() === true && $this->isDefault() === true && $this->isLast() === false) {
 			return false;
 		}
 
@@ -286,6 +288,14 @@ class Language implements Stringable
 	public function isLast(): bool
 	{
 		return App::instance()->languages()->count() === 1;
+	}
+
+	/**
+	 * Checks if this is the single language object
+	 */
+	public function isSingle(): bool
+	{
+		return $this->single;
 	}
 
 	/**
@@ -455,7 +465,8 @@ class Language implements Stringable
 		return new static([
 			'code'    => 'en',
 			'default' => true,
-			'locale'  => App::instance()->option('locale', 'en_US.utf-8')
+			'locale'  => App::instance()->option('locale', 'en_US.utf-8'),
+			'single'  => true
 		]);
 	}
 
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -447,6 +447,19 @@ class Language implements Stringable
 	}
 
 	/**
+	 * Create a placeholder language object in a
+	 * single-language installation
+	 */
+	public static function single(): static
+	{
+		return new static([
+			'code'    => 'en',
+			'default' => true,
+			'locale'  => App::instance()->option('locale', 'en_US.utf-8')
+		]);
+	}
+
+	/**
 	 * Returns the custom slug rules for this language
 	 */
 	public function slugs(): array
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -274,8 +274,13 @@ class Language implements Stringable
 	 */
 	public function isDeletable(): bool
 	{
+		// a single-language object cannot be deleted
+		if ($this->isSingle() === true) {
+			return false;
+		}
+
 		// the default language can only be deleted if it's the last
-		if ($this->isSingle() === true && $this->isDefault() === true && $this->isLast() === false) {
+		if ($this->isDefault() === true && $this->isLast() === false) {
 			return false;
 		}
 
--- a/src/Content/ContentStorage.php
+++ b/src/Content/ContentStorage.php
@@ -83,11 +83,8 @@ class ContentStorage
 	 * Adapts all versions when converting languages
 	 * @internal
 	 */
-	public function convertLanguage(string $from, string $to): void
+	public function convertLanguage(Language $from, Language $to): void
 	{
-		$from = $this->language($from, true);
-		$to   = $this->language($to, true);
-
 		foreach ($this->dynamicVersions() as $versionId) {
 			$this->handler->move($versionId, $from, $versionId, $to);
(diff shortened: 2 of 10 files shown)


src/Cms/Language.php
```diff
@@ -446,6 +446,19 @@ protected function siblingsCollection(): Languages
...
 	}
 
+	/**
+	 * Create a placeholder language object in a
+	 * single-language installation
+	 */
+	public static function single(): static
+	{
+		return new static([
+			'code'    => 'en',
```
lukasbestle, line 472: Just a `Thought` I had: In the mid-term this setup with the language object for single-lang could help us to implement https://kirby.nolt.io/246 so that we don't need to hardcode `en` everywhere. But I think for now it is consistent and safe to use `en` here.

src/Content/ContentStorage.php
```diff
@@ -262,32 +267,21 @@ protected function ensureExistingVersion(
...
 	 * @param bool $force If set to `true`, the language code is not validated
-	 * @return string Language code
 	 */
 	protected function language(
 		string|null $languageCode = null,
-		bool $force = false
-	): string {
-		// in force mode, use the provided language code even in single-lang for
-		// compatibility with the previous behavior in `$model->contentFile()`
-		if ($force === true) {
```
lukasbestle, line 273: `Question/Possible Issue`: Do we lose any backwards-compatibility by removing this?

If I remember correctly, we needed the force mode to process the automatic conversion from single-lang to multi-lang when creating the first language in the Panel. At this point, the site is still single-lang but Kirby needs to build storage paths for multi-lang to rename the existing content files.

src/Content/PlainTextContentStorageHandler.php
```diff
@@ -279,27 +255,26 @@ public function touch(VersionId $versionId, string $lang): void
...
-		$this->write($versionId, $lang, $fields);
+		$this->write($versionId, $language, $fields);
 	}
 
 	/**
 	 * Writes the content fields of an existing version
 	 *
-	 * @param string $lang Code `'default'` in a single-lang installation
 	 * @param array<string, string> $fields Content fields
+	 * @param string $lang Code `'default'` in a single-lang installation
```
lukasbestle: `Todo`: This hint can be completely removed as we now use the `$language` object:

```suggestion
```

tests/Content/ContentTranslationTest.php
```diff
@@ -109,9 +109,18 @@ public function testContentMerged()
 	public function testContentFile()
 	{
 		$app = new App([
+			'languages' => [
```
lukasbestle, line 112: `Question`: What was the reasoning to remove those props (making the app object single-lang in the test)?

tests/Content/PlainTextContentStorageHandlerTest.php
```diff
@@ -47,308 +93,372 @@ public function tearDown(): void
 	 */
 	public function testCreateChangesMultiLang()
 	{
+		$this->setUpMultiLanguage();
```
lukasbestle, line 96: `Feedback`: I like this kind of test-specific setup, good idea.

tests/Content/ContentTranslationTest.php
bastianallgeier: Is this the correct line? I'm converting the App object to a multi-language installation here.

src/Content/ContentStorage.php
bastianallgeier: ~~We use this internally for the conversion only. That's why I wasn't worried about backwards compatibility. We have to pass this force parameter down a lot of methods so far just to manipulate the filename. That seemed like a very rough and dirty approach to me. But you might be right that the conversion could actually break but we don't catch it in the unit tests so far. I haven't double-checked that.~~

~~In that case, I would suggest to handle this differently. I remember that we talked about it at btconf. The Language object could have a hint if it needs the language code prefix or not. We pass the Language object all the way down already - which makes a lot of sense and the content file methods should react to that and not to the $force argument.~~

bastianallgeier: ~~Sorry, I had a look at this again and it's starting to break my brain once more.~~

~~The main reason why I had to remove the force option here is obviously the ContentStorage::language method that has to return a Language object in order to pass it down to the ContentStorageHandler. We can't simply "force" an invalid language code anymore at this point.~~

~~For some reason, this does not break any unit tests, which is kind of shit and I basically relied on that.~~

~~The content storage class will be removed entirely in one of the next PRs. That's why we are in a state right now where we would need to find a temporary fix for this for this particularly PR while it will change later again.~~

~~I absolutely hate those situations. We are obviously lacking a proper unit test for single-language to multi-language conversion. I guess if we had one of those, it would probably fail with this PR. But what do we do?~~

~~1) ignore the issue for now and add a todo for it?~~
~~2) write a unit test in this PR and then create a temporary fix somehow only to replace it later again?~~

~~I'm really not sure how to proceed.~~

bastianallgeier: Ok, I sat down and I think I have a proper solution without delaying it further. 

1. I created a dedicated LanguageConversionTest class to make sure that we properly test single-to-multi-to-single language conversion
2. I've introduced a new single option for the Language model together with a new Language::isSingle method.  This is now used in Language::single to return a Language object with that state. 
3. ContentStorage::convertLanguage will now either be called with Language::single as first argument to convert form single to multi language, or as second argument to convert from multi language to single language. 
4. The PlainTextContentStorageHandler uses the Language::isSingle method to reliably determine if it should add the language code or not. 

I feel like this is a really clean way to handle this chaos. Another option we might want to consider is to use a SingleLanguage class, which extends the Language class instead of the ::isSingle state. This would maybe be a bit more predictable. What would you prefer?

tests/Content/ContentTranslationTest.php
lukasbestle: Dafuq. Either I was super confused or it showed up red before in the iOS app.

src/Cms/Language.php
```diff
@@ -288,6 +295,14 @@ public function isLast(): bool
 		return App::instance()->languages()->count() === 1;
 	}
 
+	/**
+	 * Checks if this is the single language object
```
lukasbestle, line 299: As this will change with the new always-multilang concept, I think we should make this internal:

```suggestion
	 * Checks if this is the single language object
	 * @internal
```

src/Content/PlainTextContentStorageHandler.php
```diff
@@ -99,42 +94,36 @@ protected function contentFileForPage(Page $model, VersionId $versionId, string
...
 	}
 
 	/**
 	 * Creates a filename with extension and optional language code
 	 * in a multi-language installation
-	 *
-	 * @param string $lang Code `'default'` in a single-lang installation
 	 */
-	protected function contentFilename(string $name, string $lang): string
+	public function contentFilename(string $name, Language $language): string
```
lukasbestle: Looks like this became public again by rebase or something. If we need it, it should be internal like `contentFile()`.

```suggestion
	protected function contentFilename(string $name, Language $language): string
```

tests/Cms/Languages/LanguageConversionTest.php
```diff
@@ -0,0 +1,128 @@
...
+		$this->assertCount(1, $app->languages());
+		$this->assertTrue($app->multilang());
+
+		// the content file should now have been moved to .en
+		$this->assertFileExists($page->root() . '/test.en.txt');
+		$this->assertFileDoesNotExist($page->root() . '/test.txt');
+
+		// the content should still be loadable for the page
+		$this->assertSame('Title', $page->content()->title()->value());
+		$this->assertSame('Title', $page->content('en')->title()->value());
```
lukasbestle, line 63: To verify that the existing content actually got moved correctly and isn't only cached in memory somewhere, I suggest we check the file contents as well.

src/Content/ContentStorage.php
lukasbestle: The language objects instead of strings with force are really great and are indeed cleaning it all up in a very good way.

About `isSingle` vs. `SingleLanguage`: Both have their advantages. `SingleLanguage` would show up more clearly when debugging, but since the single language will go with the new always-multilang concept, I don't think it's worth it.

bastianallgeier: Good point. Then I would leave it as it is

src/Content/PlainTextContentStorageHandler.php
bastianallgeier: Yep, it slipped through when I rebased the PRs

tests/Cms/Languages/LanguageConversionTest.php
bastianallgeier: Good idea

lukasbestle (approved):
That all looks very sensible now. 👍
The `ContentStorage` class is in a weird limbo now, but as it will get replaced anyway quite soon, it doesn't really matter.

PR #6450, opened 2024-05-17
Changes 6: Version class improvements and unit tests

bastianallgeier:
## This PR …

- [x] 🚨 Merge first: #6436
- [x] 🚨 Merge first: #6439
- [x] 🚨 Merge first: #6442
- [x] 🚨 Merge first: #6448
- [x] 🚨 Merge first: #6449

### Refactoring

- Move more logic from `ContentStorage` to `Version` where it makes sense
  - `Version::contentFile`
  - Modified `Version::delete` to always delete all languages 
  - `Version::touchLanguage`
- Pass `Language` objects to storage methods
- Return null in `::modified` if a version does not exist.
- Add unit tests

### Outlook

- The new `VersionTest::setUpSingleLanguage` and `VersionTest::setUpMultiLanguage` methods turn out to be super useful. https://github.com/getkirby/kirby/pull/6454 will implement a new `Kirby\Content\TestCase` class with those setup methods and the `PlainTextContentStorageHandlerTest` and `VersionTest` classes will both extend them, which makes the tests a lot easier. We will also be able to use this for the `MemoryContentStorageHandlerTest` class later https://github.com/getkirby/kirby/pull/6457

### Breaking changes

None

## Ready?

### For review team

- [x] ~Add changes & docs to release notes draft in Notion~ (not relevant as the `Version` class is added in v5 alpha 1)

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -194,7 +194,7 @@ class Version
 	/**
 	 * Updates the modification timestamp of an existing version
 	 *
-	 * @param string|null $language If null, all available languages will be touched
+	 * @param Language|string|null $language If null, all available languages will be touched
 	 *
 	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
 	 */
@@ -223,7 +223,7 @@ class Version
 	 *
 	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
 	 */
-	public function touchLanguage(Language|string $language = 'default'): void
+	protected function touchLanguage(Language|string $language = 'default'): void
 	{
 		// make sure the version exists
 		$this->ensure($language);
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -165,6 +165,8 @@ class Version
 
 	/**
 	 * Moves the version to a new language and/or version
+	 *
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
 	 */
 	public function move(
 		Language|string $fromLanguage,
@@ -184,6 +186,8 @@ class Version
 	 * Returns the stored content fields
 	 *
 	 * @return array<string, string>
+	 *
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
 	 */
 	public function read(Language|string $language = 'default'): array
 	{
@@ -194,38 +198,10 @@ class Version
 	/**
 	 * Updates the modification timestamp of an existing version
 	 *
-	 * @param Language|string|null $language If null, all available languages will be touched
-	 *
 	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
 	 */
-	public function touch(Language|string|null $language = null): void
+	public function touch(Language|string $language = 'default'): void
 	{
-		// touch a single language
-		if ($this->model->kirby()->multilang() === false) {
-			$this->touchLanguage('default');
-			return;
-		}
-
-		// touch a specific language
-		if ($language !== null) {
-			$this->touchLanguage($language);
-			return;
-		}
-
-		// touch all languages
-		foreach ($this->model->kirby()->languages() as $language) {
-			$this->touchLanguage($language);
-		}
-	}
-
-	/**
-	 * Updates the modification timestamp of a specific language
-	 *
-	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
-	 */
-	protected function touchLanguage(Language|string $language = 'default'): void
-	{
-		// make sure the version exists
 		$this->ensure($language);
 		$this->model->storage()->touch($this->id, $this->language($language));
 	}
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -154,8 +154,11 @@ class Version
 	public function modified(
 		string $language = 'default'
 	): int|null {
-		$this->ensure($language);
-		return $this->model->storage()->modified($this->id, $this->language($language));
+		if ($this->exists($language) === true) {
+			return $this->model->storage()->modified($this->id, $this->language($language));
+		}
+
+		return null;
 	}
 
 	/**
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -70,6 +70,7 @@ class Version
 		// delete the default language in single-language mode
 		if ($this->model->kirby()->multilang() === false) {
 			$this->model->storage()->delete($this->id, $this->language('default'));
+			return;
 		}
 
 		// delete all languages
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -86,16 +86,16 @@ class Version
 	public function ensure(
 		Language|string $language = 'default'
 	): bool {
-		if ($this->exists($language) !== true) {
-			$message = match($this->model->kirby()->multilang()) {
-				true  => 'Version "' . $this->id . ' (' . $language . ')" does not already exist',
-				false => 'Version "' . $this->id . '" does not already exist',
-			};
-
-			throw new NotFoundException($message);
+		if ($this->exists($language) === true) {
+			return true;
 		}
 
-		return true;
+		$message = match($this->model->kirby()->multilang()) {
+			true  => 'Version "' . $this->id . ' (' . $language . ')" does not already exist',
+			false => 'Version "' . $this->id . '" does not already exist',
+		};
+
+		throw new NotFoundException($message);
 	}
 
 	/**
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -62,35 +62,19 @@ class Version
 	}
 
 	/**
-	 * Deletes a version by language or for any language
-	 *
-	 * @param string|null $language If null, all available languages will be deleted
+	 * Deletes a version with all its languages
 	 */
-	public function delete(string|null $language = null): void
+	public function delete(): void
 	{
-		// delete a single language
+		// delete the default language in single-language mode
 		if ($this->model->kirby()->multilang() === false) {
-			$this->deleteLanguage('default');
-		}
-
-		// delete a specific language
-		if ($language !== null) {
-			$this->deleteLanguage($language);
-			return;
+			$this->model->storage()->delete($this->id, $this->language('default'));
 		}
 
 		// delete all languages
 		foreach ($this->model->kirby()->languages() as $language) {
-			$this->deleteLanguage($language);
+			$this->model->storage()->delete($this->id, $language);
 		}
-	}
-
-	/**
-	 * Deletes a version by a specific language
-	 */
-	public function deleteLanguage(string $language = 'default'): void
-	{
-		$this->model->storage()->delete($this->id, $this->language($language));
 	}
 
 	/**
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -31,7 +31,7 @@ class Version
 	/**
 	 * Returns a Content object for the given language
 	 */
-	public function content(string $language = 'default'): Content
+	public function content(Language|string $language = 'default'): Content
 	{
 		return new Content(
 			parent: $this->model,
@@ -46,18 +46,17 @@ class Version
 	 *
 	 * @internal
 	 */
-	public function contentFile(string $language = 'default'): string
+	public function contentFile(Language|string $language = 'default'): string
 	{
 		return $this->model->storage()->contentFile($this->id, $this->language($language));
 	}
 
 	/**
 	 * Creates a new version for the given language
-	 * @todo Convert to a static method that creates the version initially with all relevant languages
 	 *
 	 * @param array<string, string> $fields Content fields
 	 */
-	public function create(array $fields, string $language = 'default'): void
+	public function create(array $fields, Language|string $language = 'default'): void
 	{
 		$this->model->storage()->create($this->id, $this->language($language), $fields);
 	}
@@ -70,7 +69,6 @@ class Version
 		// delete the default language in single-language mode
 		if ($this->model->kirby()->multilang() === false) {
 			$this->model->storage()->delete($this->id, $this->language('default'));
-			return;
 		}
 
 		// delete all languages
@@ -86,7 +84,7 @@ class Version
 	 * @throws \Kirby\Exception\NotFoundException if the version does not exist
 	 */
 	public function ensure(
-		string $language = 'default'
+		Language|string $language = 'default'
 	): bool {
 		if ($this->exists($language) !== true) {
 			$message = match($this->model->kirby()->multilang()) {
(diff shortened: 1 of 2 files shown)


src/Content/Version.php
```diff
@@ -98,6 +114,27 @@ public function id(): VersionId
 		return $this->id;
 	}
 
+	/**
+	 * Converts a "user-facing" language code to a `Language` object
+	 * to be used in storage methods
+	 */
+	protected function language(
+		string|null $languageCode = null,
```
lukasbestle: `Suggestion`: Maybe we could make this support `Language` objects as well (similar to how you implement `VersionId::from()` in PR 8). Then the method signature for all methods could be `Language|string $language = 'default'`. So both we internally as well as users could pass whatever is most convenient in a particular situation.

src/Content/Version.php
```diff
@@ -33,39 +35,46 @@ public function content(string $language = 'default'): Content
 	{
 		return new Content(
 			parent: $this->model,
-			data:   $this->model->storage()->read($this->id, $language),
+			data:   $this->model->storage()->read($this->id, $this->language($language)),
```
lukasbestle, line 38: `Question`: If I understand it correctly, all those calls assume that `$model->storage()` returns the handler instance and not `ContentStorage`, right? As `ContentStorage` only supports language strings as arguments.

Wouldn't that mean that PR 8 breaks the setup because it will pass language objects while `ContentStorage` is still in place (to be removed in PR 9)?

src/Content/Version.php
```diff
@@ -33,39 +35,46 @@ public function content(string $language = 'default'): Content
...
+	 * @internal
+	 */
+	public function contentFile(string $language = 'default'): string
+	{
+		return $this->model->storage()->contentFile($this->id, $this->language($language));
+	}
+
 	/**
 	 * Creates a new version for the given language
+	 *
```
lukasbestle, line 57: For reference later:

```suggestion
	 * @todo Convert to a static method that creates the version initially with all relevant languages
	 *
```

src/Content/Version.php
```diff
@@ -33,39 +35,46 @@ public function content(string $language = 'default'): Content
...
-			foreach ($this->model->kirby()->languages() as $language) {
-				$this->model->storage()->delete($this->id, $language->code());
-			}
-		}
-
 		// delete the default language in single-language mode
 		if ($this->model->kirby()->multilang() === false) {
-			$this->model->storage()->delete($this->id, 'default');
-			return;
+			$this->model->storage()->delete($this->id, $this->language('default'));
```
lukasbestle, line 72: `Suggestion`: The `return` statement doesn't do much, but it could improve clarity and also slightly increase performance in single-lang:

```suggestion
			$this->model->storage()->delete($this->id, $this->language('default'));
			return;
```

src/Content/Version.php
```diff
@@ -76,18 +85,25 @@ public function delete(string|null $language = null): void
...
-			throw new NotFoundException('Version "' . $this->id . ' (' . $language . ')" does not already exist');
+			$message = match($this->model->kirby()->multilang()) {
+				true  => 'Version "' . $this->id . ' (' . $language . ')" does not already exist',
+				false => 'Version "' . $this->id . '" does not already exist',
+			};
+
+			throw new NotFoundException($message);
 		}
+
+		return true;
```
lukasbestle: `Question`: What is the plan for this return value? In its current form the method only ever returns `true` or throws an exception, so the return value becomes meaningless. Are there plans to add a `false` return later?

src/Content/Version.php
```diff
@@ -125,61 +163,74 @@ public function modified(
...
+		if ($language !== null) {
+			$this->touchLanguage($language);
 			return;
 		}
 
-		// touch a single language
-		$this->model->storage()->touch($this->id, $language);
+		// touch all languages
+		foreach ($this->model->kirby()->languages() as $language) {
+			$this->touchLanguage($language);
```
lukasbestle: `Note`: This call especially benefits from allowing to pass `Language` objects to `Version` methods instead of the language strings (see above). Otherwise the object would have to be converted to a string and back to an object for each call here.

src/Content/Version.php
```diff
@@ -125,61 +163,74 @@ public function modified(
...
+		}
+	}
+
+	/**
+	 * Updates the modification timestamp of a specific language
+	 *
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
+	 */
+	public function touchLanguage(string $language = 'default'): void
```
lukasbestle: `Question`: What were your thoughts about the separate `touchLanguage` method? From what I can tell when reading the code, the logic here could be integrated into `touch()` directly, but I'm curious about your plans and thoughts.

tests/Content/VersionTest.php
```diff
@@ -14,18 +15,78 @@ class VersionTest extends TestCase
 {
```
lukasbestle, line 17: The new testing setup is really great and very clean 👍

`Todo`: Could you please add tests for the following that are currently not covered:
- Different ways of retrieving language objects (suggestion: I'd add `@covers ::language` as a second covered method to relevant tests that test the different ways of passing/not passing a language).
- Test(s) for the `move()` behavior
- Touching all languages at once

src/Content/Version.php
bastianallgeier: I really like that idea. I stumbled across it a couple times later where it would have been nice to simply pass the Language object. But I did not implement it because it would not have fitted in any other PR. I think it's a good idea to already take care of it in this one.

src/Content/Version.php
bastianallgeier: This really looks like an oversight on my end. The return is definitely needed.

src/Content/Version.php
bastianallgeier: No, it could really also be marked as void and return nothing instead. It's more of a personal preference. I read it like "Everything fine down here" but from a usage perspective, it does not add any value.

src/Content/Version.php
bastianallgeier: `Explanation`: It's an artefact from when we still had `::deleteLanguage` to keep both methods (delete and touch) consistent. When I removed `::deleteLanguage` I was wondering if I should delete this method as well and the decided against it for some reason that I cannot remember. I think it really isn't needed at all.

src/Content/Version.php
bastianallgeier: ~~`Explanation:` It's all a bit of a chicken and egg situation in between the different PRs. It would have been correct to still pass a string argument here, but since `Version::content` isn't used in our own code so far, it won't break anything.~~

~~If we want to keep it temporarily correct, this would need to be fixed and then later reverted. I guess it's more a question if we are able to keep track of it until PR 9. I basically used the CMS unit tests as a safety net here.~~

src/Content/Version.php
bastianallgeier: ✅

src/Content/Version.php
bastianallgeier: `Correction`: The Language object is stringable. That's why it all still keeps working with the outdated `ContentStorage` class.

src/Content/Version.php
bastianallgeier: `Correction`: I've looked at ways to remove `::touchLanguage` and it would turn `::touch` into a lot more Spaghetti code to remove it. I decided to keep it but turned it into a protected method to be used as an internal helper for us. 

`Question`: I think the bigger decision here is if we want to keep `null` as option to touch all versions. You had that question before and I'm still not 100% sure about it. 

`Suggestion:` I would keep it as it is for now. But maybe we can turn it into a todo/question for later in Notion.

src/Content/Version.php
bastianallgeier: `Suggestion`: I switched to an early return in the method. For me, this is an even better argument to return true here, for readability reasons. Returning void would still work as well though. I think both ways are fine for me.

tests/Content/VersionTest.php
bastianallgeier: ✅

src/Content/Version.php
lukasbestle: I'd prefer `void`. `bool` suggests that there could be `false` as well, which isn't the case here.

src/Content/Version.php
lukasbestle: **About the separate `touchLanguage` method:** I think if we
- change the first if to override the `$language` variable and fall through,
- put the logic of `touchLanguage` into the second if and
- change the recursive call in the loop to just `touch()`

we can get rid of the separate method without too much spaghetti.

**About touching all languages:** AFAIK, we only need touching to update the modification timestamp for the locking feature. At the moment we lock all languages together, which would mean it is enough to lock the default language file. In the future we will lock each language separately. So touching individual languages is definitely needed, while touching all languages can be done by touching the default language. Unless we have other use cases for touching, I think we could remove the `null` option.

bastianallgeier: You are right. I removed touchLanguage and simplified touch by removing the null option.

src/Content/Version.php
bastianallgeier: Ok, I've refactored the method to use void as return type.

src/Content/Version.php
```diff
@@ -109,77 +152,70 @@ public function model(): ModelWithContent
...
 	/**
 	 * Updates the content fields of an existing version
 	 *
 	 * @param array<string, string> $fields Content fields
-	 * @param string $lang Code `'default'` in a single-lang installation
 	 *
 	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
 	 */
-	public function update(array $fields, string $language = 'default'): void
+	public function update(array $fields, Language|string $language = 'default'): void
```
lukasbestle, line 218: `Suggestion`: For consistency with the storage classes, I wonder whether we should flip the argument order in `create()` and `update()`. Would also make the language the first argument for all version methods.

src/Content/Version.php
```diff
@@ -109,77 +152,70 @@ public function model(): ModelWithContent
...
 	 * Moves the version to a new language and/or version
+	 *
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
 	 */
-	public function move(string $fromLanguage, VersionId $toVersionId, string $toLanguage): void
-	{
+	public function move(
+		Language|string $fromLanguage,
+		VersionId $toVersionId,
+		Language|string $toLanguage
```
lukasbestle, line 176: `Idea`: The `toVersionId` and `toLanguage` arguments could be nullable to make them optional. So one could e.g. do:

```php
$version->move('de', toLanguage: 'en');
$version->move('de', toVersionId: VersionId::published());
```

The argument that is left out would default to its "current" value. What we would need though is an exception if neither the target version nor target language are passed. So I'm not sure if the added code is worth the usage sugar.

src/Content/Version.php
```diff
@@ -109,77 +152,70 @@ public function model(): ModelWithContent
...
 	 *
-	 * @param string $lang Code `'default'` in a single-lang installation
 	 * @return array<string, string>
+	 *
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
 	 */
-	public function read(string $language = 'default'): array
+	public function read(Language|string $language = 'default'): array
 	{
 		$this->ensure($language);
```
lukasbestle, line 196: `Idea`: I wonder whether we should replace this `ensure` call with a nullable return value.

Because with the current logic, there could be race conditions. E.g. the calling code calls `exists()` and on success it calls `read()`. But if the version/language is deleted or moved in between those calls, the `read()` call will throw an unexpected exception. If `read` would return `null` in this case, the calling code could handle that much more gracefully without having to wrap everything in a `try catch`.

But then, we have the same potential issue in `move()`, `touch()` and `update()` and I'm not sure how we would solve it there.

tests/Content/VersionTest.php
```diff
@@ -35,79 +106,326 @@ public function tearDown(): void
...
+			'title' => 'Test'
+		]);
 
-		$version->create([
+		Data::write($this->contentFile('de'), [
+			'title' => 'Töst'
+		]);
+
+		$this->assertSame('Test', $version->content('en')->get('title')->value());
+		$this->assertSame('Test', $version->content($this->app->language('en'))->get('title')->value());
```
lukasbestle, line 130: Added test for the default language:

```suggestion
		$this->assertSame('Test', $version->content($this->app->language('en'))->get('title')->value());
		$this->assertSame('Test', $version->content()->get('title')->value());
```

However I'm actually wondering if the default should actually be the current version for consistency with `$modelWithContent->content()`.

src/Content/Version.php
distantnative: Chiming in here from the side, but I'd have required parameters before optional ones. I think newer PHP even requires it.

lukasbestle: Right, that's a good point.

src/Content/Version.php
bastianallgeier: We hardly ever use `::move` so far in our code. I would not optimize this to be honest, or rather optimize it later if we ever find a part where we need to work with it more often and find it annoying.

src/Content/Version.php
bastianallgeier: I chose consistency with the ModelWithContent::update and ModelWithContent::save here instead. It's a bit weird because it's a different order than in the storage handlers, but I think it's consistent in both places now and Version is acting some sort of a translator.

src/Content/Version.php
bastianallgeier: I haven't added it to the outlook section so far because it's in the steps after PR 10, but I ran into more situations where `::read` actually should return an empty array (or maybe null) if the version does not exist or cannot be read. 

I would vote to solve this in a later step when we see why it needs to be different.

tests/Content/VersionTest.php
bastianallgeier: We have some really nasty situations where we mix 'default' and 'current' as default language codes in model methods. I totally get why it feels like it should be 'current' here, but I also really cringe at the thought to mix 'current' and 'default' in multiple Version methods. 

I think we are having better chances to make a good decision when we arrive at the translation vs content nightmare.

src/Content/Version.php
lukasbestle: Empty array is confusing/inconsistent IMO as it could well be that the version exists but does not have fields. But `null` could work.

Fine with doing it later. Could you please add it as a todo?

tests/Content/VersionTest.php
lukasbestle: I agree. For now it's fine as it's internal anyway. And maybe we can write our own code in a way that is always explicit. I.e. the calling code never relies on the default. Then we'll have an easier time to refactor this later if needed.

bastianallgeier:
@lukasbestle I think we can ignore the coverage. Codecov is considering the Language::single line uncovered although it clearly is covered by multiple tests.

lukasbestle:
It was covered, but not marked as covered 😉

lukasbestle (approved):

PR #6453, opened 2024-05-20
Fix non existing file links

### LINKED ISSUE #6446: Non existing files throw error 500: the thumbnail configuration could not be found
## Description
Non-existing files trigger an error 500. I can reproduce this in any installation, also on getkirby.com:

https://getkirby.com/media/site/hello/error.pdf
https://getkirby.com/media/pages/features/hello/error.jpeg

Enabling debug modes shows this error message: The thumbnail configuration could not be found.

**Expected behavior**  
A 404 should be returned instead.

## Your setup
**Kirby Version**  
4.2.0

afbora:
### Fixes
- Non existing media files no longer throw error 500: the thumbnail configuration could not be found #6446

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Media.php
+++ b/src/Cms/Media.php
@@ -60,7 +60,7 @@ class Media
 		try {
 			return static::thumb($model, $hash, $filename);
 		} catch (NotFoundException) {
-			// render the error page if some stuff for thumb generating not found
+			// render the error page if there is no job for this filename
 			return false;
 		}
 	}
--- a/src/Cms/Media.php
+++ b/src/Cms/Media.php
@@ -57,7 +57,12 @@ class Media
 		}
 
 		// try to generate a thumb for the file
-		return static::thumb($model, $hash, $filename);
+		try {
+			return static::thumb($model, $hash, $filename);
+		} catch (NotFoundException) {
+			// render the error page if some stuff for thumb generating not found
+			return false;
+		}
 	}
 
 	/**
--- a/tests/Cms/Media/MediaTest.php
+++ b/tests/Cms/Media/MediaTest.php
@@ -57,6 +57,11 @@ class MediaTest extends TestCase
 		$this->assertFalse(Media::link(null, 'hash', 'filename.jpg'));
 	}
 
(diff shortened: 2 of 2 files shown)


src/Cms/Media.php
```diff
@@ -57,7 +57,17 @@ public static function link(
 		}
 
 		// try to generate a thumb for the file
-		return static::thumb($model, $hash, $filename);
+		try {
+			return static::thumb($model, $hash, $filename);
+		} catch (Throwable $e) {
```
lukasbestle: Not sure, but maybe we could restrict this to just the `NotFoundException`. Then all actual errors will still produce HTTP 500 and the fatal page, while the non-existing thumbs will return the error page with HTTP 404. With this change, we also wouldn't need to differentiate between debug and non-debug.

Maybe I'm missing something though, so feel free to correct me :)

bastianallgeier: I like the idea to really look for the specific NotFoundException

afbora: @lukasbestle I'm a bit confused. Actually `static::thumb` already throws `NotFoundException` but page status code is 500 🤷‍♂️ 
https://github.com/getkirby/kirby/blob/fix/6446-non-existing-file-links/src/Cms/Media.php#L122-L123

Can you explain what you have in mind, please?

afbora: Nvm, I get it. I'll implement.

afbora:
I'll squash commits if it's OK.

src/Cms/Media.php
```diff
@@ -57,7 +57,12 @@ public static function link(
 		}
 
 		// try to generate a thumb for the file
-		return static::thumb($model, $hash, $filename);
+		try {
+			return static::thumb($model, $hash, $filename);
+		} catch (NotFoundException) {
+			// render the error page if some stuff for thumb generating not found
```
lukasbestle:
```suggestion
			// render the error page if there is no job for this filename
```

lukasbestle (approved):

PR #6454, opened 2024-05-21
Changes 7: Move ContentStorage methods to ContentStorageHandler

bastianallgeier:
## This PR …

- [x] 🚨 Merge first: #6436
- [x] 🚨 Merge first: #6439
- [x] 🚨 Merge first: #6442
- [x] 🚨 Merge first: #6448
- [x] 🚨 Merge first: #6449
- [x] 🚨 Merge first: #6450

### Refactoring

- Add methods from ContentStorage to ContentStorageHandler to prepare the removal of the ContentStorage class
  - `ContentStorageHandler::all`
  - `ContentStorageHandler::deleteLanguage`
  - `ContentStorageHandler::dynamicVersions`
  - `ContentStorageHandler::moveLanguage` (formerly `ContentStorage::convertLanguage`)
  - `ContentStorageHandler::touchLanguage`
- Added unit tests for …
  - `ContentStorageHandler::all`
  - `ContentStorageHandler::dynamicVersions`

The other methods are hard to test for the abstract class. We should probably add tests for the PlainTextContentStorageHandler class but so far all those tests have been missing anyway.

### Outlook 

- The `ContentStorage` class cannot be removed yet. That step would be too much. First, the new `ModelWithContent::version` method needs to be implemented, which happens here: https://github.com/getkirby/kirby/pull/6455 This PR also takes care of using `::version` instead of direct access to `$this->storage()` in all model action methods whereever possible. 
- Once the new `ModelWithContent::version()` method is implemented, https://github.com/getkirby/kirby/pull/6456 will remove the ContentStorage class. This will also require a few unit tests to be refactored and the `Language` class needs adjustments. 

### Breaking changes

None

## Ready?

### For review team

- [x] ~Add changes & docs to release notes draft in Notion~ (not relevant as the changes are purely internal and temporary)

### DIFF
--- a/tests/Content/TestCase.php
+++ b/tests/Content/TestCase.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Kirby\Content;
+
+use Kirby\Cms\App;
+use Kirby\Filesystem\Dir;
+use Kirby\TestCase as BaseTestCase;
+
+class TestCase extends BaseTestCase
+{
+	public const TMP = KIRBY_TMP_DIR;
+
+	protected $app;
+	protected $model;
+
+	public function setUp(): void
+	{
+		Dir::make(static::TMP);
+	}
+
+	public function setUpMultiLanguage(): void
+	{
+		$this->app = new App([
+			'languages' => [
+				[
+					'code'    => 'en',
+					'default' => true
+				],
+				[
+					'code' => 'de'
+				]
+			],
+			'roots' => [
+				'index' => static::TMP
+			],
+			'site' => [
+				'children' => [
+					[
+						'slug'     => 'a-page',
+						'template' => 'article',
+					]
+				]
+			]
+		]);
+
+		$this->model = $this->app->page('a-page');
+
+		Dir::make($this->model->root());
+	}
+
+	public function setUpSingleLanguage(): void
+	{
+		$this->app = new App([
+			'roots' => [
+				'index' => static::TMP
+			],
+			'site' => [
+				'children' => [
+					[
+						'slug'     => 'a-page',
+						'template' => 'article'
+					]
+				]
+			]
+		]);
+
+		$this->model = $this->app->page('a-page');
+
+		Dir::make($this->model->root());
+	}
+
+	public function tearDown(): void
+	{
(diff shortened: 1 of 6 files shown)


tests/Content/TestCase.php
```diff
@@ -0,0 +1,77 @@
...
+
+use Kirby\Cms\App;
+use Kirby\Filesystem\Dir;
+use Kirby\TestCase as BaseTestCase;
+
+class TestCase extends BaseTestCase
+{
+	public const TMP = KIRBY_TMP_DIR;
+
+	protected $app;
```
lukasbestle: `Todo`: Could you please rebase the PR against `v5/develop`? Merging #6474 broke test classes like this one because of the incompatible app property definition. Fixed that on `v5/develop` for the previous PRs we merged, but this PR would reintroduce the issue.

lukasbestle (commented):
I'm having trouble understanding why we need each method and why the methods are implemented in the "handler sphere".

So could you please share some reasoning for the methods themselves and their place in `ContentStorageHandler`?

bastianallgeier:
@lukasbestle the only purpose of this PR is to prepare getting rid of the ContentStorage class. I'm not super happy with the individual method names, but I decided to port them over 1:1 from the ContentStorage class (except of ::convertLanguage) I really did not want to look into better names here at this point. 

I think eventually those methods could be in other places. 

`:all` and `::dynamicVersions` seem to me like they could be in a new Versions class. But I'm not 100% sure if that's too much. Do we have any other usage for such a Versions class at this point? 

`::deleteLanguage`, `::moveLanugage` and `::touchLanguage` are only needed in the `Language` class and maybe we should move them there instead. 🤷‍♂️

But those decisions personally really led to decision-paralysis when all I wanted was to get rid of ContentStorage. So I decided to keep everything as it is. It's also easier to make sure that unit tests still pass and we don't need tons of new ones. 

I think it's important to keep in mind with all those PRs that they help to form a decision-making process for us. I totally feel like the individual PRs bring clarity to where things really should belong. But some of the decisions only fall into place two, three or maybe even five steps in the future.

PR #6455, opened 2024-05-21
Changes 8: New ModelWithContent::version method

bastianallgeier:
## This PR …

- [x] 🚨 Merge first: #6436
- [x] 🚨 Merge first: #6439
- [x] 🚨 Merge first: #6442
- [x] 🚨 Merge first: #6448
- [x] 🚨 Merge first: #6449
- [x] 🚨 Merge first: #6450
- [x] 🚨 Merge first: #6454

### Features

- New `ModelWithContent::version()` method

### Refactoring

- Use new version method wherever it makes sense. 
  - `ModelWithContent::convertTo`
  - `ModelWithContent::readContent`
  - `ModelWithContent::writeContent`
  - `File::modifiedContent`
  - `Page::modified`
  - `System::exposedFileUrl`
  - `User::exists`
  - `User::modified`
  - `ContentTranslation::contentFile`
  - `FileActionsTest`
  - `PageActionsTest`

The other methods are hard to test for the abstract class. We should probably add tests for the PlainTextContentStorageHandler class but so far all those tests have been missing anyway.

### Breaking changes

- `ModelWithContent::version()` is now a reserved keyword. If you've used a field with this name, you need to use `$model->content()->get('version')` instead to work with the field.
- `Page::modified()`, `File::modifiedContent()` and `User::modified()` now use the current language instead of the default language.

## Ready?

### For review team

### DIFF
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -4,7 +4,6 @@ namespace Kirby\Cms;
 
 use Closure;
 use Kirby\Content\Field;
-use Kirby\Content\VersionId;
 use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Exception\NotFoundException;
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -830,7 +830,7 @@ class Page extends ModelWithContent
 		string|null $handler = null,
 		string|null $languageCode = null
 	): int|string|false|null {
-		$modified = $this->version(VersionId::default($this))->modified(
+		$modified = $this->version()->modified(
 			$languageCode ?? 'current'
 		);
 
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -830,9 +830,8 @@ class Page extends ModelWithContent
 		string|null $handler = null,
 		string|null $languageCode = null
 	): int|string|false|null {
-		$modified = $this->storage()->modified(
-			VersionId::default($this),
-			$languageCode
+		$modified = $this->version(VersionId::default($this))->modified(
+			$languageCode ?? 'current'
 		);
 
 		if ($modified === null) {
--- a/src/Cms/User.php
+++ b/src/Cms/User.php
@@ -208,10 +208,7 @@ class User extends ModelWithContent
 	 */
 	public function exists(): bool
 	{
-		return $this->storage()->exists(
-			VersionId::published(),
-			'default'
-		);
+		return $this->version(VersionId::published())->exists('default');
 	}
 
 	/**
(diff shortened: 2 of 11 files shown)


src/Content/VersionId.php
```diff
@@ -79,8 +79,12 @@ public static function default(ModelWithContent $model): static
 	/**
 	 * Creates a VersionId instance from a simple string value
 	 */
-	public static function from(string $value): static
+	public static function from(self|string $value): static
```
distantnative: I think, we usually rather use the classname for these cases than `self`

bastianallgeier: static does not work. I've switched to full class names.

src/Cms/Page.php
```diff
@@ -830,9 +829,8 @@ public function modified(
 		string|null $handler = null,
 		string|null $languageCode = null
 	): int|string|false|null {
-		$modified = $this->storage()->modified(
-			VersionId::default($this),
-			$languageCode
+		$modified = $this->version()->modified(
+			$languageCode ?? 'current'
```
lukasbestle, line 833: `Thought`: v4 seems to be using the `default` language, so this would technically be a breaking change. However I think it makes sense to go with `current` here. But then we need to document it in the release notes and also think about consistent behavior in `$file->modifiedContent()`.

src/Cms/User.php
```diff
@@ -473,7 +470,7 @@ public function modified(
 		string|null $handler = null,
 		string|null $languageCode = null
 	): int|string|false {
-		$modifiedContent = $this->storage()->modified(VersionId::published(), $languageCode);
+		$modifiedContent = $this->version(VersionId::published())->modified($languageCode ?? 'current');
```
lukasbestle, line 473: Same here

tests/Cms/Models/ModelWithContentTest.php
```diff
@@ -443,4 +444,26 @@ public function testUuid()
...
+
+	public function testVersion()
+	{
+		$model = new Site();
+		$this->assertInstanceOf(Version::class, $model->version('published'));
+		$this->assertSame('published', $model->version()->id()->value());
+
+		$model = new Page(['slug' => 'foo']);
+		$this->assertInstanceOf(Version::class, $model->version('published'));
+		$this->assertSame('published', $model->version()->id()->value());
```
lukasbestle: `Todo`: Since the fallback is tested in a separate test:

```suggestion
		$this->assertSame('published', $model->version('published')->id()->value());

		$model = new Page(['slug' => 'foo']);
		$this->assertInstanceOf(Version::class, $model->version('published'));
		$this->assertSame('published', $model->version('published')->id()->value());
```

We may also want to test passing a `VersionId` instance.

src/Cms/Page.php
bastianallgeier: I think we are really cleaning things up here by following the scheme to always use `current` for all model methods. I agree that we need to follow through here and do the same in modifiedContent then. I will change it and add the release notes and breaking changes.

PR #6456, opened 2024-05-21
Changes 9: Remove content storage class

bastianallgeier:
## This PR …

- [x] 🚨 Merge first: #6436
- [x] 🚨 Merge first: #6439
- [x] 🚨 Merge first: #6442
- [x] 🚨 Merge first: #6448
- [x] 🚨 Merge first: #6449
- [x] 🚨 Merge first: #6450
- [x] 🚨 Merge first: #6454
- [x] 🚨 Merge first: #6455

### Refactoring

- `ModelWithContent` and all it's children are now using the `PlainTextContentStorage` class without the wrapping `ContentStorage` class
- `ContentStorage` has been removed
- `Language::create` and `Language::delete` have been fixed 

### Breaking changes

- `ContentStorage` does no longer exist. Since it has been marked as internal, this should not be an issue. 
  
## Ready?

### For review team


src/Cms/Language.php
```diff
@@ -176,18 +176,18 @@ public static function create(array $props): static
 
 		$language->save();
 
+		// update the main languages collection in the app instance
+		$kirby->languages(false)->append($language->code(), $language);
```
lukasbestle: `Question`: What's the reasoning behind moving this line to the top?

bastianallgeier: That change wasn't necessary. I had a different version of the code before which needed to make sure that the language collection is updated first. But I just reverted it to avoid unnecessary changes and confusion here.

lukasbestle (approved):

PR #6457, opened 2024-05-22
Changes 10: New MemoryContentStorageHandler

bastianallgeier:
## This PR …

- [x] 🚨 Merge first: #6436
- [x] 🚨 Merge first: #6439
- [x] 🚨 Merge first: #6442
- [x] 🚨 Merge first: #6448
- [x] 🚨 Merge first: #6449
- [x] 🚨 Merge first: #6450
- [x] 🚨 Merge first: #6454
- [x] 🚨 Merge first: #6455
- [x] 🚨 Merge first: #6456

### Features

- New `MemoryContentStorageHandler` class (see outlook for details what this class is supposed to do)

### Refactor

- Move `::write` and `::ensure` into `ContentStorageHandler` class and make them abstract
- Turn `ContentStorageHandler::update` and `ContentStorageHandler::create` into non-abstract methods that use `::write` under the hood. 
- Use `ContentStorageHandler::ensure` in `Version::ensure`

### Outlook

- The new `MemoryContentStorageHandler` class will be used in `ModelWithContent::setContent` and `ModelWithContent::setTranslation`. It will replace the `PlainTextContentStorageHandler` instance with an in-memory version of the content or translation that is passed via array. This will (hopefully) keep our virtual pages, files and users working, while still making use of the underlying storage implementation. It should also help to untie the content vs. translations knot. 

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Content/ContentStorageHandler.php
+++ b/src/Content/ContentStorageHandler.php
@@ -101,18 +101,18 @@ abstract class ContentStorageHandler
 	 *
 	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
 	 */
-	public function ensure(VersionId $versionId, Language $language): bool
+	public function ensure(VersionId $versionId, Language $language): void
 	{
-		if ($this->exists($versionId, $language) !== true) {
-			$message = match($this->model->kirby()->multilang()) {
-				true  => 'Version "' . $versionId . ' (' . $language->code() . ')" does not already exist',
-				false => 'Version "' . $versionId . '" does not already exist',
-			};
-
-			throw new NotFoundException($message);
+		if ($this->exists($versionId, $language) === true) {
+			return;
 		}
 
-		return true;
+		$message = match($this->model->kirby()->multilang()) {
+			true  => 'Version "' . $versionId . ' (' . $language->code() . ')" does not already exist',
+			false => 'Version "' . $versionId . '" does not already exist',
+		};
+
+		throw new NotFoundException($message);
 	}
 
 	/**
--- a/src/Content/ContentStorageHandler.php
+++ b/src/Content/ContentStorageHandler.php
@@ -6,6 +6,7 @@ use Generator;
 use Kirby\Cms\Language;
 use Kirby\Cms\ModelWithContent;
 use Kirby\Cms\Page;
+use Kirby\Exception\NotFoundException;
 
 /**
  * Abstract for content storage handlers;
@@ -52,7 +53,10 @@ abstract class ContentStorageHandler
 	 *
 	 * @param array<string, string> $fields Content fields
 	 */
-	abstract public function create(VersionId $versionId, Language $language, array $fields): void;
+	public function create(VersionId $versionId, Language $language, array $fields): void
+	{
+		$this->write($versionId, $language, $fields);
+	}
 
 	/**
 	 * Deletes an existing version in an idempotent way if it was already deleted
@@ -92,6 +96,26 @@ abstract class ContentStorageHandler
 	}
 
 	/**
+	 * Checks if a version/language combination exists and otherwise
+	 * will throw a NotFoundException
+	 *
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
+	 */
+	public function ensure(VersionId $versionId, Language $language): bool
+	{
+		if ($this->exists($versionId, $language) !== true) {
+			$message = match($this->model->kirby()->multilang()) {
+				true  => 'Version "' . $versionId . ' (' . $language->code() . ')" does not already exist',
+				false => 'Version "' . $versionId . '" does not already exist',
+			};
+
+			throw new NotFoundException($message);
+		}
+
+		return true;
+	}
+
+	/**
 	 * Checks if a version exists
 	 */
 	abstract public function exists(VersionId $versionId, Language $language): bool;
@@ -104,12 +128,21 @@ abstract class ContentStorageHandler
 	/**
 	 * Moves content from one version-language combination to another
 	 */
-	abstract public function move(
+	public function move(
 		VersionId $fromVersionId,
 		Language $fromLanguage,
 		VersionId $toVersionId,
 		Language $toLanguage
-	): void;
+	): void {
+		// read the existing fields
+		$fields = $this->read($fromVersionId, $fromLanguage);
+
+		// create the new version
+		$this->create($toVersionId, $toLanguage, $fields);
+
+		// clean up the old version
+		$this->delete($fromVersionId, $fromLanguage);
+	}
 
 	/**
 	 * Adapts all versions when converting languages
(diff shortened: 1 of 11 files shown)


distantnative:
@bastianallgeier have you already an idea how to solve the cloning of model objects with these? I remember last year this was one of the hard walls we hit with this idea

bastianallgeier:
@distantnative tbh not yet. But I think it's the only way to move forward and we really need to tackle it again when we get there. I feel that we have more headspace this year, so I'm confident that we can get it done. I would try to work on the clone method some more to get it ported over to the clone

src/Content/ContentStorageHandler.php
```diff
@@ -6,6 +6,7 @@
 use Kirby\Cms\Language;
 use Kirby\Cms\ModelWithContent;
 use Kirby\Cms\Page;
+use Kirby\Exception\NotFoundException;
 
 /**
  * Abstract for content storage handlers;
```
lukasbestle, line 12: `General thought/idea`: The `ContentStorageHandler` abstract keeps getting more complex, which could make it harder to implement custom handlers. Once we have the methods cleaned up and moved to the right places, it might make sense to add an interface again. We would then have `abstract class ContentStorageHandler implements ContentStorageHandlerInterface` and type-hint the interface where we currently type-hint the abstract. Plugin devs who want to use our helpers in the abstract can extend the abstract and those who want to start fresh can implement the interface directly.

src/Content/ContentStorageHandler.php
```diff
@@ -91,6 +95,26 @@ public function dynamicVersions(): array
 		return $versions;
 	}
 
+	/**
+	 * Checks if a version/language combination exists and otherwise
+	 * will throw a NotFoundException
```
lukasbestle:
```suggestion
	 * will throw a `NotFoundException`
```

src/Content/ContentStorageHandler.php
```diff
@@ -158,7 +191,20 @@ public function touchLanguage(Language $language): void
...
+
+	/**
+	 * Writes the content fields of an existing version
+	 *
+	 * @param array<string, string> $fields Content fields
+	 *
+	 * @throws \Kirby\Exception\Exception If the content cannot be written
 	 */
-	abstract public function update(VersionId $versionId, Language $language, array $fields): void;
+	abstract protected function write(VersionId $versionId, Language $language, array $fields): void;
```
lukasbestle, line 209: This is another reason for a separate interface. A handler implementation that has separate `create()` and `update()` methods does not need `write()` but still needs to implement it. With a separate interface, this becomes a non-issue as the interface only defines the public methods.

src/Content/MemoryContentStorageHandler.php
```diff
@@ -0,0 +1,130 @@
...
+	/**
+	 * Returns the modification timestamp of a version if it exists
+	 */
+	public function modified(VersionId $versionId, Language $language): int|null
+	{
+		if ($this->exists($versionId, $language) === false) {
+			return null;
+		}
+
+		return $this->cache->get($this->cacheId($versionId, $language))['modified'];
```
lukasbestle: `Question`: Couldn't we use `$cache->modified()` for this?

src/Content/MemoryContentStorageHandler.php
```diff
@@ -0,0 +1,130 @@
...
+
+	/**
+	 * Updates the modification timestamp of an existing version
+	 *
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
+	 */
+	public function touch(VersionId $versionId, Language $language): void
+	{
+		$fields = $this->read($versionId, $language);
+		$this->write($versionId, $language, $fields);
```
lukasbestle, line 90: `Suggestion`: For this we could add `$cache->touch()` that updates the created/modified time of the cache `Value`. I think that could be universally useful.

tests/Content/MemoryContentStorageHandlerTest.php
```diff
@@ -0,0 +1,353 @@
...
+class MemoryContentStorageHandlerTest extends TestCase
+{
+	protected $storage;
+
+	/**
+	 * @covers ::create
+	 * @covers ::delete
+	 * @covers ::exists
+	 * @covers ::write
+	 */
```
lukasbestle: `Question`: Do `@covers` annotations take effect on such custom assertion methods? I thought PHPUnit only respects the test method itself and doesn't track other called methods in the test class.

src/Content/ContentStorageHandler.php
bastianallgeier: Sounds good to me. I still think that we can clean up the abstract class quite a bit later. But I like the idea of the interface as a baseline.

tests/Content/MemoryContentStorageHandlerTest.php
bastianallgeier: TBH, I 'm not entirely sure. I thought it would work this way. I've removed them. The test cases should already be enough to cover the class.

src/Content/MemoryContentStorageHandler.php
bastianallgeier: I like the idea, but I shy away from implementing it for all cache handlers in this PR.

src/Content/MemoryContentStorageHandler.php
bastianallgeier: The problem seems to be that the MemoryCache class does not properly update the modification date or at least I get wrong results from calling `MemoryCache::modified` directly.

src/Content/ContentStorageHandler.php
bastianallgeier: I've added a todo to Notion.

src/Content/MemoryContentStorageHandler.php
lukasbestle: You are right, I think it's still fine the way you've done it.

src/Content/MemoryContentStorageHandler.php
lukasbestle: Hm, I wonder why. `MemoryCache::set()` creates a fresh `Cache\Value` instance each time and the `Value` constructor sets `$this->created` to `time()`. So `$cache->modified()` (being an alias for `$cache->created()`) should return the timestamp when the cache value was created.

If we can get it to work with the cache modified timestamp, we could clean up the handler a lot as it would then only need to store the fields array.

bastianallgeier: I will try again today. Maybe I did something wrong in my tests or the implementation.

bastianallgeier: ✅

src/Content/MemoryContentStorageHandler.php
```diff
@@ -0,0 +1,115 @@
...
+	 * @param array<string, string> $fields Content fields
+	 *
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
+	 */
+	public function update(VersionId $versionId, Language $language, array $fields): void
+	{
+		$this->ensure($versionId, $language);
+		$this->write($versionId, $language, $fields);
+	}
+
```
lukasbestle: `Todo`: Unless I'm missing something, this looks to be the exact same implementation as in the `ContentStorage` abstract and can be removed.

```suggestion
```

bastianallgeier: You are totally right. I just removed it.

lukasbestle (approved):

PR #6459, opened 2024-05-25
Select field: remove undocumented empty prop

distantnative:
## This PR …

Revives https://github.com/getkirby/kirby/pull/3373. Presented breaking case inside structure fields works by now.

By now the empty prop is redundant. Everything can be replicated by required, default and placeholder props.

### Enhancements
- `default` prop supports Kirby queries in radio and select field

### Breaking changes
- Select field: `empty` prop was removed. Use combination of `required`, `placeholder` and `default` to replicate functionality.

### For review team

### DIFF
--- a/panel/src/components/Forms/Input/SelectInput.vue
+++ b/panel/src/components/Forms/Input/SelectInput.vue
@@ -13,13 +13,13 @@
 			:disabled="disabled"
 			:name="name"
 			:required="required"
-			:value="selected"
+			:value="value"
 			class="k-select-input-native"
-			@change="onInput($event.target.value)"
+			@change="$emit('input', $event.target.value)"
 			@click="onClick"
 		>
 			<option v-if="hasEmptyOption" :disabled="required" value="">
-				{{ emptyOption }}
+				{{ empty }}
 			</option>
 			<option
 				v-for="option in options"
@@ -42,7 +42,6 @@ export const props = {
 	mixins: [InputProps, options, placeholder],
 	props: {
 		ariaLabel: String,
-		default: String,
 		value: {
 			type: [String, Number, Boolean],
 			default: ""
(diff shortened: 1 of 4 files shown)


panel/src/components/Forms/Input/SelectInput.vue
```diff
@@ -70,10 +63,6 @@ export default {
 			return this.placeholder ?? "—";
 		},
 		hasEmptyOption() {
-			if (this.empty === false) {
-				return false;
-			}
-
 			return !(this.required && this.default);
```
lukasbestle: To make it easier to understand what this does, we could split this up and add a comment:

```suggestion
			// empty option is only hidden if the field is both required and has a default
			return !(this.required && this.default);
```

lukasbestle (commented):
Beautifully simple. After wrapping my head around the logic, it makes a lot of sense to me.

panel/src/components/Forms/Input/SelectInput.vue
distantnative: I like it, but I think now the comment is confusing as it doesn't quite describe what the code below expresses (even if the result is the same). Maybe keep the old notation with the comment?

lukasbestle: You are right. I first changed the logic because it twisted my head and then wrote the comment, but the comment is much clearer with the original syntax. Changed it above.

lukasbestle (approved):
Gets my approval backend-wise 👍

distantnative:
@lukasbestle I spotted that it can be simplified even further: `default` in the Vue component had no real own functionality. It wouldn't actually set a default value. As other inputs, the default functionality depends on the backend writing the default to `value`. 

So I removed the `default` prop from the Vue component and now the empty option is shown if the input is not required. Or if it's required, then it's only shown as long as the value is empty. Once it has a value, the empty option disappears.

Also allows us to get rid of the local `selected` state and just use `value`.

lukasbestle (commented):
I had written a comment asking about how the `default` prop works at all, but then realized the backend does it, so I answered my own question. But I didn't realize that required fields only need the empty state before they are first set. That's a good point and simultaneously clears up that confusion I had for the future.

TBH I don't fully get why we needed that `selected` attribute at all, but you know better why we don't need it anymore. If you are sure about the implementation, I think nothing speaks against merging the PR, otherwise someone else should review the frontend code.

distantnative:
@afbora would be great to get your review and tests here as well. Always good to be double safe with such changes.

afbora (approved):
LGTM and works great 👍

PR #6462, opened 2024-05-26
Upload dialog: fix allowed characters for filename

### LINKED ISSUE #6461: Filename sanitization rules are inconsistent
## Description
There’s an inconsistency when changing the filename in the upload dialog compared to changing it afterwards. During upload `a-z0-9` is used as allowed character set ([source](https://github.com/getkirby/kirby/blob/c1e6ff77b056774543f201035f5dab6a02bfdb3b/config/areas/files/dialogs.php#L29)), while `changeName` uses  `a-z0-9@._-` ([source](https://github.com/getkirby/kirby/blob/c1e6ff77b056774543f201035f5dab6a02bfdb3b/panel/src/helpers/string.js#L179)).

`F::safeName()` allows underscores as well `a-z0-9@._-` ([source](https://github.com/getkirby/kirby/blob/c1e6ff77b056774543f201035f5dab6a02bfdb3b/src/Filesystem/F.php#L762C32-L762C43)), so they stay in place.

**Expected behavior**  
Same sanitization rules in all places.

## To reproduce
1. Upload a file in the panel and try to add an underscore in the filename
2. Change the filename afterwards and try to add an underscore in the filename

## Your setup
**Kirby Version**  
4.2.0

 - Device: MacBook Pro
 - OS: macOS 14.5
 - Browser: Safari
 - Version: 17.5

distantnative:
## This PR …

Consistent with change filename dialog: https://github.com/getkirby/kirby/blob/c1e6ff77b056774543f201035f5dab6a02bfdb3b/config/areas/files/dialogs.php#L29

### Fixes
- Upload dialog: allows `@`, `.` and `_` in filenames to be consistent with other filename rules
https://github.com/getkirby/kirby/issues/6461
- `$file->changeName()` now sanitizes new filenames with`.` included correctly

### For review team

### DIFF
--- a/panel/src/components/Uploads/UploadItem.vue
+++ b/panel/src/components/Uploads/UploadItem.vue
@@ -13,6 +13,7 @@
 			:novalidate="true"
 			:required="true"
 			:value="name"
+			allow="a-z0-9@._-"
 			class="k-upload-item-input"
 			type="slug"
 			@input="$emit('rename', $event)"
--- a/panel/src/helpers/string.js
+++ b/panel/src/helpers/string.js
@@ -210,8 +210,8 @@ export function slug(string, rules = [], allowed = "", separator = "-") {
 	string = string.replace("/", separator);
 
 	// trim leading and trailing non-word-chars
-	string = string.replace(new RegExp("^[^" + allowed + "]+", "g"), "");
-	string = string.replace(new RegExp("[^" + allowed + "]+$", "g"), "");
+	string = string.replace(new RegExp("^[^a-z0-9]+", "g"), "");
+	string = string.replace(new RegExp("[^a-z0-9]+$", "g"), "");
 
 	return string;
 }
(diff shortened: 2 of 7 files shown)


afbora:
@distantnative I found something . Could you check whether it's bug or not? I think there is an inconsistency about `@` char.

1. Select a file from files section
2. Set filename as `file.a@b_c-d` in dialog before upload and upload it
3. `file.a@b_c-d.jpg` files will be uploaded successfully and you'll see in section
4. Try to rename the file with `file.a@b_c-de` (append just `e`)
5. You will see that `@` is gone as `file.a-b-c-de.jpg`

distantnative:
@afbora Good catch. This was actually its own bug where `$file->changeName()` would apply `F::safeName()` even though passing the name string without extension. So when that name contained a `.`, `F::safeName()` would consider that the separator between base name and extension. Now using `F::safeBasename()`.

afbora (approved):

PR #6463, opened 2024-05-30
Move all login components into same location

distantnative:
## This PR …

Moving the closely related Vue components to the same folder makes it easier to keep an overview and work on them going forward.

### Refactor
- Moved all login-related Vue components into same location
- Aligned some of the component names

### Deprecated
- `<k-login>` is deprecated. Use `<k-login-form>` instead.
- `<k-login-code>` is deprecated. Use `<k-login-code-form>` instead.

## Ready?

### For review team


panel/src/components/Views/Login/index.js
```diff
@@ -0,0 +1,28 @@
+import LoginView from "./LoginView.vue";
+
+import LoginForm from "./LoginForm.vue";
+import LoginAlert from "./LoginAlert.vue";
+import LoginCodeForm from "./LoginCodeForm.vue";
+
+import InstallationView from "./InstallationView.vue";
+import ResetPasswordView from "./ResetPasswordView.vue";
```
bastianallgeier: I love that the we reorganize the login components like this. But I don't really think of the Installation View as a part of this to be honest. I'm not entirely sure about the ResetPasswordView. It's a weird hybrid somewhere between account and login.

panel/src/components/Views/Login/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+import UserInfo from "./UserInfo.vue";
```
bastianallgeier: Would it make sense to rename this into LoginUserInfo as well while we are already at it?

panel/src/components/Views/Login/index.js
distantnative: Moved InstallationView to its own folder and ResetPasswordView to users. It's a hybrid, but in the end maybe more account related as you need a user first with a password to be able to reset it.

panel/src/components/Views/Login/index.js
distantnative: Sine ResetPasswordView is also using it and it could be used in other user-related contexts as well, I moved it to the Users folder and kept the name as is.

bastianallgeier (approved):

Issue #6451, opened 2024-05-18
`k-table`: index column too narrow for high indexes


distantnative:
4 digit indexes cannot be displayed:

<img width="739" alt="Screenshot 2024-05-18 at 19 11 07" src="https://github.com/getkirby/kirby/assets/3788865/ce25b319-0961-4412-81ec-676e40530a42">

First index is actually `1201`

afbora:
@distantnative I think if we remove the `padding-inline` for first index, the issue will be solved.

https://github.com/getkirby/kirby/blob/c1e6ff77b056774543f201035f5dab6a02bfdb3b/panel/src/components/Layout/Table.vue#L386

distantnative:
@afbora I think the culprit is rather 

https://github.com/getkirby/kirby/blob/c1e6ff77b056774543f201035f5dab6a02bfdb3b/panel/src/components/Layout/Table.vue#L480

PR #6467, opened 2024-06-03
Fix table index column for high indexes

afbora:
### Fixes
- k-table: index column too narrow for high indexes #6451

### Breaking changes
None

## Docs

None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Layout/Table.vue
+++ b/panel/src/components/Layout/Table.vue
@@ -479,6 +479,7 @@ export default {
 .k-table .k-table-index-column {
 	width: var(--table-row-height);
 	text-align: center;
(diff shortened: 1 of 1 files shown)


distantnative:
@afbora Does this really fix it even for indexes with more than 4 digits?

afbora:
@distantnative No, max 5 digits 🙂 Needs to be dynamic width for more than 4 digits.

distantnative:
Yes, so I don't think this is a suitable fix. We need to find one that works with any number of digits. I know it's hard and I don't think fixing this needs a high priority. But wouldn't go for a fix for just 4 digits, but not more.

PR #6479, opened 2024-06-09
Revert: deprecating `this.$helper.object.clone`

distantnative:
### Summary of changes
- Undeprecate `$helper.object.clone`
- Use `$helper.object.clone` instead of `structuredClone`

### Reasoning
`structuredClone` cannot handle the automatic reactivity layer that Vue 3 adds on most/all objects. So we need a beefier version of `$helper.object.clone` to solve this. But therefore, we should stick to using the helper everywhere.

### Fix
- `$helper.object.clone` is no longer deprecated. Please use it instead of `structuredClone` as this might cause issues down the road with Vue 3.

## Ready?

### DIFF
--- a/panel/lab/components/stats/1_stats/index.vue
+++ b/panel/lab/components/stats/1_stats/index.vue
@@ -38,11 +38,11 @@ export default {
 			];
 		},
 		reportsWithoutTheme() {
-			return structuredClone(this.reports).map((report) => {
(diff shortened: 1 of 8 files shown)


bastianallgeier (approved):
Sometimes I really just hate all the reactivity parts for exactly those kind of quirks.

PR #6483, opened 2024-06-11
Changes 11: Language::ensure & Languages::ensure

bastianallgeier:
## This PR …

- [x] 🚨 Merge first: #6436
- [x] 🚨 Merge first: #6439
- [x] 🚨 Merge first: #6442
- [x] 🚨 Merge first: #6448
- [x] 🚨 Merge first: #6449
- [x] 🚨 Merge first: #6450
- [x] 🚨 Merge first: #6454
- [x] 🚨 Merge first: #6455
- [x] 🚨 Merge first: #6456
- [x] 🚨 Merge first: #6457

### Features

- New `Language::ensure` method to always return a Language object - even in single-language mode
- New `Languages::ensure` method to always return a Languages collection - even in single-language mode

### Refactor

- Use new methods in ModelWithContent, ContentStorageHandler & Version classes

### Breaking changes

`ModelWithContent::translation` throws a NotFoundException if the requested language does not exist. 

## Ready?

- [x] In-code documentation (wherever needed)
- [x] Unit tests for fixed bug/feature
- [x] Tests and checks all pass

### For review team

- [x] Add changes & docs to release notes draft in Notion

PR #6490, opened 2024-06-14
Changes 12: More Version class improvements

bastianallgeier:
## This PR …

- [x] 🚨 Merge first: #6436
- [x] 🚨 Merge first: #6439
- [x] 🚨 Merge first: #6442
- [x] 🚨 Merge first: #6448
- [x] 🚨 Merge first: #6449
- [x] 🚨 Merge first: #6450
- [x] 🚨 Merge first: #6454
- [x] 🚨 Merge first: #6455
- [x] 🚨 Merge first: #6456
- [x] 🚨 Merge first: #6457
- [x] 🚨 Merge first: #6483

### Reasoning

While working on the next steps (following PRs) I realised that the Version class still needs additional adjustments to help us tackle the problems in the ModelWithContent class later. 

One of the most important changes is to move field normalization and field removal to the Version class. If we don't do that, developers could call e.g. `$model->version()->update()` with fields that should actually not be stored at all. Moving this part of the logic to the ContentStorageHandler classes would be too much and Version seems like the perfect place for it. It will also reduce the logic in the Model classes. 

With this change, we can also simplify the Content class later. It is no longer responsible to normalize field names afterwards. 

### Features

- New `Version::replace` method as a dedicated way to replace all existing fields in a version with the given array. This helps to clean up the `$overwrite` logic in `ModelWithContent::save` and `ModelWithContent::update`
- New `Version::save` method to provide a reliable way to always save the version, no matter if it already exists or not. This is a wrapper around `Version::create`, `Version::update` and `Version::replace` and prepares to deprecate
  - `ModelWithContent::save` 
  - `ModelWithContent::saveContent`
  - `ModelWithContent::saveTranslation`
  - `ModelWithContent::writeContent`
- New protected `Version::convertFieldNamesToLowerCase` method to normalize field names on read and write. 
- New protected `Version::prepareFieldsBeforeWrite` method to remove untranslatable fields and the uuid field for non-default languages.
- New protected `Version::prepareFieldsAfterRead` to clean field names after fields have been read from storage.

### Refactor

- `Version::read` returns null if the version does not exist and throws an exception if the language does not exist.
- `Version::content` will now use `Version::read` to always return a Content object, even if a version does not exist. It also implements the content fallback merge with the default language for non-default languages. 
- `Version::update` will now read the last state and merge fields with the given array to make sure that the end result will always be a full set with all fields and no fields get lost.
- The protected `Version::language` method can now be removed since `Language::ensure` is available.

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -86,10 +86,11 @@ class Version
 	{
 		$language = Language::ensure($language);
 
-		$fields = $this->convertFieldNamesToLowerCase($fields);
-		$fields = $this->removeUnwantedFields($fields, $language);
-
-		$this->model->storage()->create($this->id, $language, $fields);
+		$this->model->storage()->create(
+			versionId: $this->id,
+			language: $language,
+			fields: $this->prepareFieldsBeforeWrite($fields, $language)
+		);
 	}
 
 	/**
@@ -172,28 +173,17 @@ class Version
 	}
 
 	/**
-	 * Returns the stored content fields
-	 *
-	 * @return array<string, string>
+	 * Prepare fields to be written by removing unwanted fields
+	 * depending on the language or model and by cleaning the field names
 	 */
-	public function read(Language|string $language = 'default'): array
+	protected function prepareFieldsBeforeWrite(array $fields, Language $language): array
 	{
-		$language = Language::ensure($language);
+		// convert all field names to lower case
+		$fields = $this->convertFieldNamesToLowerCase($fields);
 
-		try {
-			$fields = $this->model->storage()->read($this->id, $language);
-			$fields = $this->convertFieldNamesToLowerCase($fields);
-			return $fields;
-		} catch (Throwable) {
-			return [];
-		}
-	}
+		// make sure to store the right fields for the model
+		$fields = $this->model->contentFileData($fields, $language);
 
-	/**
-	 * Remove fields that should not be stored for the given version and language
-	 */
-	protected function removeUnwantedFields(array $fields, Language $language): array
-	{
 		// the default language stores all fields
 		if ($language->isDefault() === true) {
 			return $fields;
@@ -213,6 +203,33 @@ class Version
 	}
 
 	/**
+	 * Make sure that reading from storage will always
+	 * return a usable set of fields with clean field names
+	 */
+	protected function prepareFieldsAfterRead(array $fields, Language $language): array
+	{
+		return $this->convertFieldNamesToLowerCase($fields);
+	}
+
+	/**
+	 * Returns the stored content fields
+	 *
+	 * @return array<string, string>
+	 */
+	public function read(Language|string $language = 'default'): array
+	{
+		$language = Language::ensure($language);
+
+		try {
+			$fields = $this->model->storage()->read($this->id, $language);
+			$fields = $this->prepareFieldsAfterRead($fields, $language);
+			return $fields;
+		} catch (Throwable) {
+			return [];
+		}
+	}
+
+	/**
 	 * Replaces the content of the current version with the given fields
 	 *
 	 * @param array<string, string> $fields Content fields
@@ -225,10 +242,11 @@ class Version
 
 		$language = Language::ensure($language);
 
-		$fields = $this->convertFieldNamesToLowerCase($fields);
-		$fields = $this->removeUnwantedFields($fields, $language);
-
-		$this->model->storage()->update($this->id, $language, $fields);
+		$this->model->storage()->update(
+			versionId: $this->id,
+			language: $language,
+			fields: $this->prepareFieldsBeforeWrite($fields, $language)
+		);
 	}
 
 	/**
@@ -280,9 +298,10 @@ class Version
 			...$this->convertFieldNamesToLowerCase($fields)
 		];
 
-		// make sure to not store unnecessary fields for the version & language
-		$fields = $this->removeUnwantedFields($fields, $language);
-
-		$this->model->storage()->update($this->id, $language, $fields);
+		$this->model->storage()->update(
+			versionId: $this->id,
+			language: $language,
+			fields: $this->prepareFieldsBeforeWrite($fields, $language)
+		);
 	}
 }
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -48,7 +48,7 @@ class Version
 	 */
 	public function contentFile(Language|string $language = 'default'): string
 	{
-		return $this->model->storage()->contentFile($this->id, $this->language($language));
+		return $this->model->storage()->contentFile($this->id, Language::ensure($language));
 	}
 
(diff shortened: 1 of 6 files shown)


src/Content/Version.php
```diff
@@ -146,23 +165,88 @@ public function move(
...
+	public function read(Language|string $language = 'default'): array
+	{
+		$language = Language::ensure($language);
+
+		try {
+			$fields = $this->model->storage()->read($this->id, $language);
+			$fields = $this->convertFieldNamesToLowerCase($fields);
+			return $fields;
+		} catch (Throwable) {
+			return [];
```
lukasbestle: `Question`: Do we need to return an empty array here? As I wrote before, I feel like this could fall on our feet as it becomes impossible to differentiate between content that does not exist at all and empty content. If possible, I'd prefer `null` for clarity.

src/Content/Version.php
```diff
@@ -146,23 +165,88 @@ public function move(
...
 	 *
 	 * @return array<string, string>
+	 */
+	public function read(Language|string $language = 'default'): array
+	{
+		$language = Language::ensure($language);
+
+		try {
+			$fields = $this->model->storage()->read($this->id, $language);
+			$fields = $this->convertFieldNamesToLowerCase($fields);
```
lukasbestle: `Thought`: TBH I'm not sure about converting to lowercase on read (on write is totally correct of course). At the moment, the `Data\Txt` class already handles normalization. And not just lowercasing but also normalization of dashes, underscore and spaces.

On the other hand, if we remove it here, we would run into issues with underlying storage handlers that don't normalize. Newly stored content should always be fine as it's also normalized on write, but existing content with different field name formatting would break merging of field arrays etc.

So maybe we do need this line here. I'm just not sure how far it should go, e.g. if we should also normalize the underscore/dash/space situation.

src/Content/Version.php
bastianallgeier: My main reason to go for the array was to simplify the usage of ::read in other places. We already have `Version::exists` to check if the version is actually there. That's why I thought we already offer ways to detect this properly.

src/Content/Version.php
bastianallgeier: So far, we do this in the Content class, which is a total mess and I never understand in which situations we actually convert the keys and when not. Putting it in the Version class felt like it's this single place where we take care of cleaning what's going into storage and what's coming out of storage and that felt nice. I have one more step on my list which is still missing here, but could maybe help with the decision. We also still have `ModelWithContent::contentFileData` which we also need to run here before storing content. Once that's implemented in the Version class as well, all the field normalization/sanitization/etc. is done here.

bastianallgeier: Yesterday, while working on moving ::contentFieldData here, I thought that the method names could be better though. Maybe something like `::prepareFieldsToWrite` and `::prepareFieldsToRead`

src/Content/Version.php
lukasbestle: About simplification of calling code: With `null`, we could use `?? []` where we need it. This would also make the behavior more explicit and visible at the first glance.

An argument why `exists()` is not enough: We could run into race conditions where a version first exists when `exists()` is called but is then deleted before `read()` is called. So the calling code would get an empty array and think it's the actual content, even though the version was deleted. If we directly call `read()` and check on `null`, we can be sure that the data is consistent. The race condition will be a rare edge case, but it will be even harder to debug if it ever happens. So IMO it would be good to avoid it from the start.

So TL;DR: With `null` returned on non-existing versions, we can improve reliability and increase flexibility for the calling code while making the calling code only slightly longer.

src/Content/Version.php
lukasbestle: Good points. 👍

I fully agree it makes sense to have all the normalization in a single place. And `Version` is a great place for it.

I was thinking about the performance optimizations we currently have e.g. in `ModelWithContent::content()` to skip a second normalization because `Data\Txt::decode()` already does it. But maybe they are not needed as lowercasing array keys is probably not that intensive. And maybe PHP is even optimized to skip processing of array keys that are already lowercase.

src/Content/Version.php
```diff
@@ -186,6 +288,20 @@ public function touch(Language|string $language = 'default'): void
...
 		$this->ensure($language);
-		$this->model->storage()->update($this->id, $this->language($language), $fields);
+
+		$language = Language::ensure($language);
+
+		// merge the previous state with the new state to always
+		// update to a complete version
+		$fields = [
+			...$this->read($language),
+			...$this->convertFieldNamesToLowerCase($fields)
```
lukasbestle: `Suggestion`: This extra lowercasing is not needed as we call `prepareFieldsBeforeWrite()` below anyway. If `array_change_key_case()` gets an array with multiple keys with the same string but different case, it automatically chooses the last one. So we can safely skip this to slightly improve performance:

```suggestion
			...$fields
```

src/Content/Version.php
bastianallgeier: I think we really need to look into caching the version in memory after reading it anyway. Once we move `ModelWithContent::content` and `ModelWithContent::translation` to Version, we no longer have the $content property in the `ModelWithContent` class to take care of that. But that step is not yet ready.

src/Content/Version.php
bastianallgeier: Oh, wow. That's super handy and good to know.

src/Content/Version.php
bastianallgeier: Ok, convinced :) I will try to adapt the code in those places where the array is expected.

bastianallgeier: I think I got a good solution now. The storage handler can use `::ensure` to throw a NotFoundException on `::read` and we can catch that in `Version::read` to return null. This will make sure that other read errors still get thrown correctly, so we don't accidentally bury them and make real reading issues harder to debug.

src/Content/Version.php
lukasbestle: I think having a guaranteed normalization and reduced code complexity is worth more than a slight (probably not even noticeable) performance improvement.

src/Content/Version.php
```diff
@@ -146,23 +166,105 @@ public function move(
...
+		bool $overwrite = false
+	): void {
+		match (true) {
+			$this->exists($language) === false
+				=> $this->create($fields, $language),
+			$overwrite === true
+				=> $this->replace($fields, $language),
+			default
+			=> $this->update($fields, $language)
+		};
```
lukasbestle: `Suggestion` to increase readability:

```suggestion
		if ($this->exists($language) === false) {
			$this->create($fields, $language);
			return;
		}

		if ($overwrite === true) {
			$this->replace($fields, $language);
			return;
		}

		$this->update($fields, $language);
```

lukasbestle (approved):

PR #6491, opened 2024-06-14
Changes 13: New translation classes

bastianallgeier:
## This PR …

- [x] 🚨 Merge first: #6436
- [x] 🚨 Merge first: #6439
- [x] 🚨 Merge first: #6442
- [x] 🚨 Merge first: #6448
- [x] 🚨 Merge first: #6449
- [x] 🚨 Merge first: #6450
- [x] 🚨 Merge first: #6454
- [x] 🚨 Merge first: #6455
- [x] 🚨 Merge first: #6456
- [x] 🚨 Merge first: #6457
- [x] 🚨 Merge first: #6483
- [x] 🚨 Merge first: #6490

### Reasoning

The `ContentTranslation` class is one of our internal classes that still uses array props in the constructor and also has some logic that shouldn't be in there. To untangle the translation knot, it really helps to refactor this class to named properties and turn most of its methods into proxies to access the underlying version or language objects. 

In order to turn this into a smooth transition, I've decided to extend the `ContentTranslation` class for now. This way, we can use it as a replacement for `ContentTranslation` in many parts already before we replace it later. We can also change methods more freely without affecting the CMS Models and their unit tests too drastically for now. 

### Features

- New `Kirby\Content\Translation` class (to eventually replace `Kirby\Content\ContentTranslation`)
- New `Kirby\Content\Translations` class to help in `ModelWithContent::translations` later and to provide a more type-safe collection for all translations. 

The following changes to `Translation` will be new features as soon as it replaces the `ContentTranslation` class. 

- New `Translation::version` method to return the parent `Version` instance
- New `Translation::create` method to setup a new translation from an array of fields.

### Deprecated

- `Translation::code` Use `::language()->code()` instead
- `Translation::content` Use `::version()->content()->toArray()` instead
- `Translation::contentFile` Use `::version()->contentFile()` instead
- `Translation::exists` Use `::version()->exists()` instead
- `Translation::isDefault` Use `::language()->isDefault()` instead

### Breaking changes

The following changes to the `Translation` class will be breaking changes as soon as it will replace the `ContentTranslation` class. 

- `Translation::parent` has been replace with `Translation::model`
- `Translation::update` has been removed

### Code coverage

No matter what I do, `Translations::load` is simply not covered and I don't know why. 

## Ready?

### For review team

### DIFF
--- a/src/Cms/Helpers.php
+++ b/src/Cms/Helpers.php
@@ -38,6 +38,13 @@ class Helpers
 		// TODO: switch to true in v6
 		'plugin-extends-root' => false,
 
+		// The `Content\Translation` class keeps a set of methods from the old
+		// `ContentTranslation` class for compatibility that should no longer be used.
+		// Some of them can be replaced by using `Version` class methods instead
+		// (see method comments). `Content\Translation::contentFile` should be avoided
+		//  entirely and has no recommended replacement.
+		'translation-methods' => false,
+
 		// Passing a single space as value to `Xml::attr()` has been
 		// deprecated. In a future version, passing a single space won't
 		// render an empty value anymore but a single space.
--- a/src/Cms/Helpers.php
+++ b/src/Cms/Helpers.php
@@ -43,7 +43,7 @@ class Helpers
 		// Some of them can be replaced by using `Version` class methods instead
 		// (see method comments). `Content\Translation::contentFile` should be avoided
 		//  entirely and has no recommended replacement.
-		'translation-methods' => false,
+		'translation-methods' => true,
 
 		// Passing a single space as value to `Xml::attr()` has been
 		// deprecated. In a future version, passing a single space won't
--- a/src/Content/Translation.php
+++ b/src/Content/Translation.php
@@ -16,7 +16,7 @@ use Kirby\Cms\ModelWithContent;
  * @copyright Bastian Allgeier
  * @license   https://getkirby.com/license
  */
-class Translation extends ContentTranslation
+class Translation
 {
 	/**
 	 * Creates a new translation object
@@ -40,6 +40,8 @@ class Translation extends ContentTranslation
 	/**
 	 * Returns the language code of the
 	 * translation
+	 *
+	 * @deprecated since 5.0.0 Use `::language()->code()` instead
 	 */
 	public function code(): string
 	{
@@ -49,14 +51,18 @@ class Translation extends ContentTranslation
 	/**
 	 * Returns the translation content
 	 * as plain array
+	 *
+	 * @deprecated since 5.0.0 Use `::version()->content()` instead
 	 */
-	public function content(): array
+	public function content(): Content
 	{
-		return $this->version->read($this->language);
+		return $this->version->content($this->language);
 	}
 
 	/**
 	 * Absolute path to the translation content file
+	 *
+	 * @deprecated since 5.0.0 Use `::version()->contentFile()` instead
 	 */
 	public function contentFile(): string
 	{
@@ -89,6 +95,8 @@ class Translation extends ContentTranslation
 
 	/**
 	 * Checks if the translation file exists
+	 *
+	 * @deprecated since 5.0.0 Use `::version()->exists()` instead
 	 */
 	public function exists(): bool
 	{
@@ -100,12 +108,14 @@ class Translation extends ContentTranslation
 	 */
 	public function id(): string
 	{
-		return $this->code();
+		return $this->language->code();
 	}
 
 	/**
 	 * Checks if the this is the default translation
 	 * of the model
+	 *
+	 * @deprecated since 5.0.0 Use `::language()->isDefault()` instead
 	 */
 	public function isDefault(): bool
 	{
@@ -133,22 +143,7 @@ class Translation extends ContentTranslation
 	 */
 	public function slug(): string|null
 	{
-		return $this->content()['slug'] ?? null;
-	}
-
-	/**
-	 * Merge the old and new data
-	 */
-	public function update(array|null $data = null, bool $overwrite = false): static
-	{
-		$data = array_change_key_case((array)$data);
-
-		$this->content = match ($overwrite) {
-			true    => $data,
-			default => [...$this->content(), ...$data]
-		};
-
-		return $this;
+		return $this->version->read()['slug'] ?? null;
 	}
 
 	/**
@@ -158,9 +153,9 @@ class Translation extends ContentTranslation
 	public function toArray(): array
 	{
 		return [
-			'code'    => $this->code(),
-			'content' => $this->content(),
-			'exists'  => $this->exists(),
+			'code'    => $this->language->code(),
+			'content' => $this->version->content($this->language)->toArray(),
+			'exists'  => $this->version->exists($this->language),
 			'slug'    => $this->slug(),
 		];
 	}
--- a/src/Content/Translation.php
+++ b/src/Content/Translation.php
@@ -58,7 +58,7 @@ class Translation extends ContentTranslation
 	 */
 	public function content(): array
 	{
-		Helpers::deprecated('`$translation->content()` has been deprecated. Use `$translation->version()->content()` instead.', 'translation-methods');
+		Helpers::deprecated('`$translation->content()->toArray()` has been deprecated. Use `$translation->version()->content()` instead.', 'translation-methods');
 		return $this->version->content($this->language)->toArray();
 	}
 
@@ -69,7 +69,7 @@ class Translation extends ContentTranslation
 	 */
 	public function contentFile(): string
 	{
-		Helpers::deprecated('`$translation->contentFile()` has been deprecated. Use `$translation->version()->contentFile()` instead.', 'translation-methods');
+		Helpers::deprecated('Please let us know if you have a use case for a replacement.', 'translation-methods');
 		return $this->version->contentFile($this->language);
 	}
 
--- a/src/Content/Translation.php
+++ b/src/Content/Translation.php
@@ -1,0 +1,175 @@
+<?php
+
+namespace Kirby\Content;
+
+use Kirby\Cms\Language;
+use Kirby\Cms\ModelWithContent;
+
+/**
+ * Each page, file or site can have multiple
+ * translated versions of their content,
+ * represented by this class
+ *
+ * @package   Kirby Content
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ */
+class Translation extends ContentTranslation
+{
+	/**
+	 * Creates a new translation object
+	 */
+	public function __construct(
+		protected ModelWithContent $model,
+		protected Version $version,
+		protected Language $language
+	) {
+	}
+
+	/**
+	 * Improve `var_dump` output
+	 * @codeCoverageIgnore
+	 */
+	public function __debugInfo(): array
+	{
+		return $this->toArray();
+	}
+
+	/**
+	 * Returns the language code of the
+	 * translation
+	 */
+	public function code(): string
+	{
+		return $this->language->code();
+	}
+
+	/**
+	 * Returns the translation content
+	 * as plain array
+	 */
+	public function content(): array
+	{
+		return $this->version->read($this->language);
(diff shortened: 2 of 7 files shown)


distantnative:
@bastianallgeier regarding your code coverage: none of your test methods says that it covers the `load` method. So even if the code does, it won't be reported as the test is limit to only cover the methods mentioned (and none mentions load).

bastianallgeier:
@distantnative wait, am I missing something? https://github.com/getkirby/kirby/pull/6491/files#diff-be3909337c2195af14e3332b8945acaa88ae6801f3b6f92f4fa40343240c0747R111 I really think I'm going crazy here. I even tried to re-type that @covers part in case I had an invisible character in it or some other typo that I just don't see.

distantnative:
@bastianallgeier sorry, my bad, mistook `Translations::load()` for `Translation::load()`

distantnative:
@bastianallgeier it seems tied to the method name. When I rename the method to `loadd`, it shows up fully covered.

`Question:` What is the reason for using the deprecated methods in the tests? https://github.com/getkirby/kirby/pull/6491/files#diff-ee064be1fccc6c3b78a97760793cd868737cdcf4273b70b1b9e61e189975b45aR63 failed for me when I replaced all deprecated methods with their suggested alternatives.

tests/Content/TranslationTest.php
```diff
@@ -0,0 +1,361 @@
+<?php
+
+namespace Kirby\Content;
+
+use Kirby\Cms\Language;
+use Kirby\Data\Data;
+
+/**
+ * @coversDefaultClass Kirby\Content\Translation
```
distantnative:
```suggestion
 * @coversDefaultClass \Kirby\Content\Translation
```

tests/Content/TranslationsTest.php
```diff
@@ -0,0 +1,143 @@
+<?php
+
+namespace Kirby\Content;
+
+/**
+ * @coversDefaultClass Kirby\Content\Translations
```
distantnative:
```suggestion
 * @coversDefaultClass \Kirby\Content\Translations
```

bastianallgeier:
So weird about the name! We could try what happens when we add a coverage ignore rule. 

About the deprecated methods. We don't need to be that strict. My thought behind is that the translation class really acts more like glue for us between language, version and content. To repeat a lot of the version methods would be quite redundant. That's why I think we should encourage to prefer going through the versions. But that's totally up for discussions. 

Did you specify the language when you used the alternative?

distantnative:
I'm for the depreciations. But we should adapt the tests then, shouldn't we? Currently the CI throws all the deprecation warnings.

bastianallgeier:
I have to be honest, I don't really know how to treat the depredations in unit tests so far

distantnative:
@bastianallgeier I am talking about replacing the deprecated code. Currently the test still use all the deprecated methods, e.g.

```php
$this->assertSame('en', $translationEN->code());
```

should be changed to

```php
$this->assertSame('en', $translationEN->language()->code());
````

etc.

**Comment by bastianallgeier** [comment]
@distantnative I've replaced all deprecated methods in the unit tests.

File: src/Content/Translation.php [hunk 3]
```diff
@@ -0,0 +1,176 @@
...
+	public function __debugInfo(): array
+	{
+		return $this->toArray();
+	}
+
+	/**
+	 * Returns the language code of the
+	 * translation
+	 *
+	 * @deprecated since 5.0.0 Use `::language()->code()` instead
```
lukasbestle: `Todo`: Consistent syntax with existing uses:

```suggestion
	 * @deprecated 5.0.0 Use `$translation->language()->code()` instead
```

src/Content/Translation.php
```diff
@@ -0,0 +1,176 @@
...
+	{
+		Helpers::deprecated('`$translation->code()` has been deprecated. Use `$translation->language()->code()` instead.', 'translation-code');
+		return $this->language->code();
+	}
+
+	/**
+	 * Returns the translation content
+	 * as plain array
+	 *
+	 * @deprecated since 5.0.0 Use `::version()->content()->toArray()` instead
```
lukasbestle:
```suggestion
	 * @deprecated 5.0.0 Use `$translation->version()->content()->toArray()` instead
```

src/Content/Translation.php
```diff
@@ -0,0 +1,176 @@
...
+	public function content(): array
+	{
+		Helpers::deprecated('`$translation->content()` has been deprecated. Use `$translation->version()->content()` instead.', 'translation-content');
+		return $this->version->content($this->language)->toArray();
+	}
+
+	/**
+	 * Absolute path to the translation content file
+	 *
+	 * @deprecated since 5.0.0 Use `::version()->contentFile()` instead
```
lukasbestle: This one could be entirely deprecated as future storage implementations won't always have a content file.

src/Content/Translation.php
```diff
@@ -0,0 +1,176 @@
...
+			model: $model,
+			version: $version,
+			language: $language,
+		);
+	}
+
+	/**
+	 * Checks if the translation file exists
+	 *
+	 * @deprecated since 5.0.0 Use `::version()->exists()` instead
```
lukasbestle:
```suggestion
	 * @deprecated 5.0.0 Use `$translation->version()->exists()` instead
```

src/Content/Translation.php
```diff
@@ -0,0 +1,176 @@
...
+	public function id(): string
+	{
+		return $this->language->code();
+	}
+
+	/**
+	 * Checks if the this is the default translation
+	 * of the model
+	 *
+	 * @deprecated since 5.0.0 Use `::language()->isDefault()` instead
```
lukasbestle:
```suggestion
	 * @deprecated 5.0.0 Use `$translation->language()->isDefault()` instead
```

src/Content/Translation.php
```diff
@@ -0,0 +1,176 @@
...
+
+	/**
+	 * Returns the language code of the
+	 * translation
+	 *
+	 * @deprecated since 5.0.0 Use `::language()->code()` instead
+	 */
+	public function code(): string
+	{
+		Helpers::deprecated('`$translation->code()` has been deprecated. Use `$translation->language()->code()` instead.', 'translation-code');
```
lukasbestle: The deprecations should be added to the `Helpers::$deprecations` array. I think it would be enough to have one key for all of them like `translation-methods`.

src/Content/Translations.php
```diff
@@ -0,0 +1,78 @@
...
+ */
+class Translations extends Collection
+{
+	/**
+	 * Creates a new Translations collection from
+	 * an array of translations properties. This is
+	 * used in LabPage::setTranslations to properly
+	 * normalize an array definition.
+	 */
+	public static function create(
```
lukasbestle, line 30: `Question`: How is this method (and `Translation::create()` intended to be used? I think the current arguments don't fit our plans to make `Version::create()` a static method. One cannot pass a version instance if it doesn't exist yet.

src/Content/Translations.php
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	/**
+	 * Loads all available translations for a given model
+	 */
+	public static function load(
+		ModelWithContent $model,
+		Version $version
+	): static {
+		$translations = new static();
```
lukasbestle: `Suggestion`: To avoid having to manipulate the internal `$data` array of the collection after creation, we could start with a plain array, add the translations to it and then pass that array to the constructor at the end of the method.

bastianallgeier: Good idea ✅ That's maybe a pattern that we should use in more places.

src/Content/Translation.php
```diff
@@ -0,0 +1,176 @@
...
+
+	/**
+	 * Returns the translation content
+	 * as plain array
+	 *
+	 * @deprecated 5.0.0 Use `::version()->content()->toArray()` instead
+	 */
+	public function content(): array
+	{
+		Helpers::deprecated('`$translation->content()` has been deprecated. Use `$translation->version()->content()` instead.', 'translation-methods');
```
lukasbestle:
```suggestion
		Helpers::deprecated('`$translation->content()` has been deprecated. Use `$translation->version()->content()->toArray()` instead.', 'translation-methods');
```

src/Content/Translation.php
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	/**
+	 * Absolute path to the translation content file
+	 *
+	 * @deprecated 5.0.0
+	 */
+	public function contentFile(): string
+	{
+		Helpers::deprecated('`$translation->contentFile()` has been deprecated. Use `$translation->version()->contentFile()` instead.', 'translation-methods');
```
lukasbestle:
```suggestion
		Helpers::deprecated('`$translation->contentFile()` has been deprecated. Please let us know if you have a use case for a replacement.', 'translation-methods');
```

src/Content/Translation.php
```diff
@@ -0,0 +1,176 @@
...
+	{
+		return $this->model;
+	}
+
+	/**
+	 * Returns the custom translation slug
+	 */
+	public function slug(): string|null
+	{
+		return $this->version->read($this->language)['slug'] ?? null;
```
lukasbestle: `Suggestion`: As we want to cache the `Content` object in `Version`, we can increase performance by using it here:

```suggestion
		return $this->version->content($this->language)->data()['slug'] ?? null;
```

src/Cms/Helpers.php
```diff
@@ -38,6 +38,13 @@ class Helpers
 		// TODO: switch to true in v6
 		'plugin-extends-root' => false,
 
+		// The `Content\Translation` class keeps a set of methods from the old
+		// `ContentTranslation` class for compatibility that should no longer be used.
+		// Some of them can be replaced by using `Version` class methods instead
+		// (see method comments). `Content\Translation::contentFile` should be avoided
+		//  entirely and has no recommended replacement.
+		'translation-methods' => false,
```
lukasbestle: `Todo`: I think we should be throwing deprecation warnings by default for this one as we intend to remove the methods in v6:

```suggestion
		'translation-methods' => true,
```

tests/Content/TranslationTest.php
```diff
@@ -0,0 +1,361 @@
...
+
+		$translationDE = new Translation(
+			model: $this->model,
+			version: $this->model->version(),
+			language: Language::ensure('de')
+		);
+
+		Data::write($this->contentFile('en'), [
+			'title' => 'Test',
+			'slug'  => 'english-slug'
```
lukasbestle: `Todo`: Getting the slug of a non-default language is not something we do in the core if I remember correctly as we dropped support for storing the slug field in the default language content for performance reasons. So I wouldn't test it either (allows us to change that later without breaking the test that tests what we don't actually need).

tests/Content/TranslationTest.php
```diff
@@ -0,0 +1,361 @@
...
+	{
+		$this->setUpMultiLanguage();
+
+		$translation = new Translation(
+			model: $this->model,
+			version: $this->model->version(),
+			language: Language::ensure('en')
+		);
+
+		Data::write($this->contentFile('en'), [
```
lukasbestle: This should then also test the secondary language for consistency:

```suggestion
			language: Language::ensure('de')
		);

		Data::write($this->contentFile('de'), [
```

src/Content/Translations.php
bastianallgeier: They are indeed only useful this way as long as we still have the setup with Version instances that may not exist yet. We discussed this yesterday. To convert everything to `ModelWithContent::version(): Version|null`  is quite a step at the moment and I'm not fully able to complete/plan it yet. 

I think we should leave it out of this PR and go with the flow here, as long as we agree that those two methods are going to get refactored again and only help temporarily to clean things up.

lukasbestle: Sure. Could you add `@todo` comments to the code as well?

bastianallgeier: ✅

lukasbestle (approved):
Two methods of `ContentTranslation` that `Translation` does not implement anymore (and doesn't need to) are `parent()` and `update()`. In order to make the transition smoother, we need to find bugs early where we maybe still rely on these methods. So I suggest to add overrides for these two methods to the `Translation` class that throw an exception.

With that I think we can finally merge this.

PR #6492, opened 2024-06-14
Changes 14: LabPage class

bastianallgeier:
## This PR …

- [x] 🚨 Merge first: #6436
- [x] 🚨 Merge first: #6439
- [x] 🚨 Merge first: #6442
- [x] 🚨 Merge first: #6448
- [x] 🚨 Merge first: #6449
- [x] 🚨 Merge first: #6450
- [x] 🚨 Merge first: #6454
- [x] 🚨 Merge first: #6455
- [x] 🚨 Merge first: #6456
- [ ] 🚨 Merge first: #6457
- [ ] 🚨 Merge first: #6483
- [ ] 🚨 Merge first: #6490
- [ ] 🚨 Merge first: #6491

### Reasoning

The LabPage class extends the Page class and serves as a playground for all the methods that need to be changed in order to get Versions properly implemented. It really helped to untangle the content vs. translation mess in a more isolated state without all the other methods. Afterwards I also took care of save/update, etc. It starts to come along really nicely, but there are still some elements missing. 

### Deprecated

- `::readContent` Use `::version()->read()` instead
- `::save` Use `::version()->save()` instead
- `::saveContent` Use `::version()->save()` instead
- `::saveTranslation` Use `::version()->save()` instead
- `::writeContent` Use `::version()->save()` instead

### Breaking changes

- `::translations` will now always return a `Translations` collection with at least the `Language::single` object in single language mode.

## Ready?

### For review team

### DIFF
--- a/src/Content/LabPage.php
+++ b/src/Content/LabPage.php
@@ -46,6 +46,10 @@ class LabPage extends Page
 		return $this->version()->read($languageCode ?? 'current');
 	}
 
+	/**
+	 * Stores the content on disk
+	 * @internal
+	 */
 	public function save(
 		array|null $data = null,
 		string|null $languageCode = null,
--- a/src/Content/LabPage.php
+++ b/src/Content/LabPage.php
@@ -21,6 +21,10 @@ class LabPage extends Page
 	 */
 	public function clone(array $props = []): static
 	{
+		// keep the current state of the content in memory storage
+		// before creating the clone. The storage of the clone can
+		// afterwards be mutated without affecting the content in the
+		// original instance.
 		$this->storage = MemoryContentStorageHandler::from($this->storage);
 		return parent::clone($props);
 	}
--- a/src/Content/LabPage.php
+++ b/src/Content/LabPage.php
@@ -144,6 +144,22 @@ class LabPage extends Page
 	}
 
 	/**
+	 * Returns the slug of the page
+	 */
+	public function slug(string|null $languageCode = null): string
+	{
+		if ($this->kirby()->multilang() === true) {
+			$language = Language::ensure($languageCode ?? 'current');
+
+			if ($language->isDefault() !== true) {
+				return $this->translation($language)->slug() ?? $this->slug;
+			}
+		}
+
+		return $this->slug;
+	}
+
+	/**
 	 * Returns a single translation by language code
 	 * If no code is specified the current translation is returned
 	 */
--- a/src/Content/LabPage.php
+++ b/src/Content/LabPage.php
@@ -1,0 +1,162 @@
+<?php
+
+namespace Kirby\Content;
+
+use Kirby\Cms\Collection;
+use Kirby\Cms\Language;
+use Kirby\Cms\Page;
+
+/**
+ * Test Model to prototype the content and translation
+ * mechanics
+ */
+class LabPage extends Page
+{
+	/**
+	 * Creates a new instance with the same
+	 * initial properties
+	 *
+	 * @todo eventually refactor without need of propertyData
+	 */
+	public function clone(array $props = []): static
+	{
+		$this->storage = MemoryContentStorageHandler::from($this->storage);
+		return parent::clone($props);
+	}
+
+	/**
+	 * Returns the content for the default version and given language code
+	 */
+	public function content(string|null $languageCode = null): Content
+	{
+		return $this->version()->content($languageCode ?? 'current');
+	}
+
+	/**
+	 * @deprecated since 5.0.0 Use `::version()->read()` instead
+	 */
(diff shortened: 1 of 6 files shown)


src/Content/LabPage.php
```diff
@@ -0,0 +1,193 @@
...
+	 *
+	 * @todo eventually refactor without need of propertyData
+	 */
+	public function clone(array $props = []): static
+	{
+		// keep the current state of the content in memory storage
+		// before creating the clone. The storage of the clone can
+		// afterwards be mutated without affecting the content in the
+		// original instance.
+		$this->storage = MemoryContentStorageHandler::from($this->storage);
```
distantnative: `Question:` How will this work when I am cloning a page object that is working with a content file storage and afterwards I update the cloned page? I think one would expect that still the content file gets updated, but with this, I fear it would only update the new memory storage so the changes would disappear after the request. Or am I overlooking something?

bastianallgeier: I think this is actually avoiding an accidental overwrite of content. Let me try to explain what happens here in code:

```php
// Assuming the title of the page on disk is just "Version 1"
$a = page('some-page');

echo $page->title() // "Version 1"

// We overwrite the title on disk
$b = $a->update([
  'title' => 'Version 2'
]);

echo $a->title() // "Version 1"
echo $b->title() // "Version 2"

// At this point, $a is in memory and $b is on disk

// calling update again on $a will create another clone in memory but not make any changes on disk
$c = $a->update([
  'title' => 'Version 1.1'
]);

echo $a->title() // "Version 1" -> in memory
echo $b->title() // "Version 2" -> on disk
echo $c->title() // "Version 1.1" -> in memory

// calling update again on $b will store the changes on disk again and keep $b in memory while returning another clone $d
$d = $b->update([
  'title' => 'Version 3'
]);

echo $a->title() // "Version 1" -> in memory
echo $b->title() // "Version 2" -> in memory
echo $c->title() // "Version 1.1" -> in memory
echo $d->title() // "Version 3" -> on disk
```

I really hope this helps to understand the mechanics behind it. 

With our old setup, you could totally kill your own changes. I.e. 

```php
$a = page('some-page');

echo $a->title() // "Version 1" 

$b = $a->update([
  'title' => 'Version 2'
]);

echo $a->title() // "Version 1" 
echo $b->title() // "Version 2" 

// calling update on the old/outdated $a version would overwrite $b on disk again, which can easily lead to unwanted results. 
$a->update([
  'title' => 'Version 3'
]);

```

distantnative: Your examples make sense that they'd lead to unwanted behaviour in the past. But maybe I am understanding this wrong: If we replace it on clone with a memory storage, won't $b also have the memory storage and no connection to the content file?

I think it's definitely a big breaking change that needs to be listed, if objects loose their connection to the content files once they get cloned. This might be the better/safer way forward indeed based on your examples, but I would bet a number of users runs into it where they relied upon any of the objects being able to write to the content file.

bastianallgeier: The `::clone` method is pretty unintuitive to read, but it basically translates to this: 

```php
$this->storage = MemoryContentStorageHandler::from($this->storage);
return new static(array_replace_recursive($this->propertyData, $props));
```

This means that a completely new Page object is returned with the same properties and that new instance starts with a fresh PlainTextContentStorageHandler instance, which provides the connection to the content file. 

The main problem here really is that we introduced the idea of immutable objects in a very half-baked way. As long as cloning a Page object leads to a state where mutating the clone also mutates the origin, it's not really immutable at all. 

I would almost say that this change is fixing our terrible implementation. But of course, this could lead to breaking changes where people rely on our terrible implementation :)

bastianallgeier: I think that we really only have two choices here: 

1) we continue with this idea of immutable objects and use the storage handlers to fix it (accepting that there will be some breaking changes)
2) move away from immutable objects entirely, which means a lot more breaking changes.

bastianallgeier: Two more ideas: 

1. Maybe we need another method that switches to memory storage and will be used in update/save, etc. that is not ::clone. Then ::clone could still be used to clone the instance without detaching the model from disk automatically. 
2. We could also think about creating another MemoryStorageHandler just for those detached model states. I.e. DetachedStorageHandler. This handler could then throw exceptions in all mutating methods to let devs know that they are trying to mutate the in-memory version of a detached model state and should rather work with the cloned instance.

distantnative: First impulse is that I like the latter. Would work like the MemoryStorageHandler but make it clear that it's for an object that usually has a content file storage (or any non-just-in-memory storage).

PR #6493, opened 2024-06-15
`Filesystem\Mime`: more readable if logic

distantnative:
### Summary of changes
- `Filesystem\Mime::toExtensions()`: refactored nested if-else into one-level if

### Reasoning
The nested if-else logic is super hard to read and keep track of the conditions for each result, which also does the same. Changing it to a one-level if is a lot easier to read, which various conditions lead to adding an extension, even if we have to list some conditions multiple times then.

## Ready?

### DIFF
--- a/src/Filesystem/Mime.php
+++ b/src/Filesystem/Mime.php
@@ -270,32 +270,37 @@ class Mime
 	/**
 	 * Returns all available extensions for a given MIME type
 	 */
-	public static function toExtensions(string|null $mime = null, bool $matchWildcards = false): array
-	{
+	public static function toExtensions(
+		string|null $mime = null,
+		bool $matchWildcards = false
+	): array {
 		$extensions = [];
-		$testMime = fn (string $v) => static::matches($v, $mime);
+		$testMime   = fn (string $v): bool => static::matches($v, $mime);
 
 		foreach (static::$types as $key => $value) {
-			if (is_array($value) === true) {
-				if ($matchWildcards === true) {
-					if (A::some($value, $testMime)) {
-						$extensions[] = $key;
-					}
-				} else {
-					if (in_array($mime, $value) === true) {
-						$extensions[] = $key;
-					}
-				}
-			} else {
-				if ($matchWildcards === true) {
-					if ($testMime($value) === true) {
-						$extensions[] = $key;
-					}
-				} else {
-					if ($value === $mime) {
-						$extensions[] = $key;
-					}
-				}
+			if (
+				(
+					is_array($value) &&
+					$matchWildcards &&
+					A::some($value, $testMime)
+				) ||
+				(
+					is_array($value) &&
+					$matchWildcards === false &&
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
TBH I think this is much harder to read, especially because the first two conditions in each group look the same at first glance and only differ by `=== false` being present or not.

If we boil it down, it's basically a matrix of the `$value` type and the requested `$matchWildcard` mode. Maybe we could simplify it by using `A::wrap()` to get rid of half of the cases first.

We could further simplify it by using `array_filter()`. Its callback can then use early returns of `true/false` so that we can use one level of separate ifs.

distantnative:
I disagree that it's harder to read. It takes such a mental load down in the second else clause to keep track of which conditions actually apply there. With that, this new explicitness would help a lot.

I agree that this still isn't "the yellow from the egg". I will have a stab at your suggestions, would be indeed great to improve this more.

src/Filesystem/Mime.php
```diff
@@ -270,36 +270,34 @@ public static function toExtension(string|null $mime = null): string|false
...
+		$extensions = array_keys(static::$types);
+
+		// filter extensions for given MIME type
+		$extensions = A::filter(
+			$extensions,
+			function ($extension) use ($mime, $matchWildcards) {
+				// get corresponding MIME types as array
+				$mimes = A::wrap(static::$types[$extension]);
+
+				if ($matchWildcards) {
```
lukasbestle:
```suggestion
				if ($matchWildcards === true) {
```

src/Filesystem/Mime.php
```diff
@@ -270,36 +270,34 @@ public static function toExtension(string|null $mime = null): string|false
...
 				}
+
+				// check if at least one MIME type matches exactly
+				return in_array($mime, $mimes);
 			}
-		}
+		);
 
-		return $extensions;
+		return array_values($extensions);
```
lukasbestle, line 301:
```suggestion
		// renumber array with consecutive keys
		return array_values($extensions);
```

lukasbestle (approved):
Much better 👍

src/Filesystem/Mime.php
distantnative: I was thinking that with PHP type hints added, we. could start to drop some of those explicit comparisons (which we once in K2 -> K3 started to use as ambigious values in K2 caused problems). I wouldn't do it everywhere, but maybe in places where the variable/method return is ensured to be boolean and also from reading the code still very clear what's happening.

What do you think?

lukasbestle: Interesting idea. I agree that with a `bool` type the same behavior is guaranteed with or without an explicit comparison, so it's not needed. It's probably more a case of getting used to it.

distantnative: I'll add it here to be consistent with the status quo. But maybe something to briefly discuss in our next call and agree if we would like to do this more going forward.

PR #6494, opened 2024-06-15
More type hints

distantnative:
### Summary of changes
- Added more type hints (mainly for class properties)

### Reasoning
Keeps our code safer, also our tests, by hardcoding our implicit assumptions.

### Additional context
I know these PRs are always huge and thus hard to review. I don't think breaking them down into separate PRs would help much here. The changes are simple, it's just a lot of them. So having dozens of PRs instead, doesn't really help to reduce the review load.

We could contemplate if we are ok with merging these kind of PRs during alpha without full review (similarly how we push some changes already directly), as the alpha and RC phase in combination with existing unit tests could be a sufficient baseline. And the cost-benefit ratio of a full review might be not that great compared to merging without full review and occasionally running into a small regression we can fix then.

### Breaking changes
- Increased PHP type hinting: when you're extending core classes, you might have to add the same type hints to your code as added to our core classes.

## Ready?

### For review team


src/Cache/Value.php
```diff
@@ -20,7 +20,7 @@ class Value
 	/**
 	 * Cached value
 	 */
-	protected $value;
+	protected mixed $value;
```
bastianallgeier, line 23: I'm not entirely sure anymore but I believe that we had issues with mixed at some point and then decided to leave it untyped if the type is mixed.

bastianallgeier (commented):
My only concern would be the mixed type and to see if we can merge it properly into our changes monster PR chain. But otherwise I think this is great.

src/Cache/Value.php
distantnative: Earlier we didn't add it as it forces to also add it without much (any) typesafety earned. However, @lukasbestle brought up a while ago that we might still want to start adding it slowly for consistency and clarity (this really should be any type, not just we didn't bother to think through what types).

distantnative:
I explicitly stayed away from content classes for this reason.

src/Cache/Value.php
lukasbestle: I agree. In our first try, we went too far and added it to a lot of places at once. But I think e.g. in this place it's completely accurate as Nico wrote.

lukasbestle (approved):
I agree with your reasoning on how to manage such PRs. Especially because most of the changes are in test files and I think it's mostly about "do the tests still pass and is there no change that could cause a false-positive test success".

However I think it would be good to limit refactoring PRs or at least commits inside refactoring PRs to one type of refactoring. For this PR it's fine, but if we take it as an example for future PRs: one commit for added type hints, one for reordered methods, one for constructor property promotion, one for cleaned up test case classes...
I know it's sometimes hard/a lot of work to split it up like that after the fact. I have experimented with keeping a separate todo list for stuff I notice while I'm at a different kind of refactoring. This allows to first do the first type of refactoring and then revisit other types directly afterwards.

distantnative:
@lukasbestle i think your comment/suggestion makes a lot of sense. And as you say it's more a matter of finding a good way to split these up again. Do you have a good way for keeping track of such todo notes?

lukasbestle:
I just use a blank text document in `TextEdit`. It's just temporary anyway. Or if it's about a specific line, I add a TODO comment that I will then see when going through the changes in Tower to select them for staging. Basically skipping over that TODO comment then and first committing everything else.

PR #6501, opened 2024-06-23
DateInput: fix `Cmd+S` in nested structures

### LINKED ISSUE #6390: Nested drawer: Ctrl + S with date fields clears content
## Description
I have a custom block with a structure field, that has several fields (text, color, list, two date fields)
When I edit a structure item and hit CTRL+S, the structure form is closed and I see the editing mask for the block again, but all content is gone – at least visually. When I close the drawer and save the changes, everything works as expected (the content is still there incl. my intentionally made changens). 

This does not happen, when I exclude the date fields. However, the changes I made were not in the date fields, but in a text field.

**Expected behavior**  
I expect CTRL+S to close the structure field edit mask and return to the block edit mask without loosing the content. 

## Your setup
Kirby 4.2.0 on PHP 8.3, running on Ubuntu in WSL (Windows 11) with Symfony Webserver (in Brave)
Also tested with Kirby 4.3.0 on PHP 8.3, running on MacOS 14.5 with Symfony Webserver (in Firefox)

distantnative:
### Summary of changes
- `DateInput.vue` does not use anymore the global `keydown.cmd.s` event to commit unfinished input, but rather listens to the event locally, preventing propagation and only emits a `submit` event once the input value has been committed/transofrmed to a proper date time object and emitted/updated.

### Reasoning
- Using the global `keydown.cmd.s` event has been triggering the `onBlur()` method no matter whether the date input was actually being edited or just present in the current view. We only need to trigger it when the input is currently being edited (as `@blur` will take care of committing the value when the focus changes away from the input already).
- To avoid the previous race conditions, we listen locally for the event (prevent default and stop propagation), commit then the value to a real datetime object and emit the `submit` event. We do not use `this.$events.emit("keydown.cmd.s")` as this triggers a race condition, where the `input` event is still propagating up the chain when the save would also happen. Emitting `submit` ensures that this event has to travel up the same chain as the `input` event and thus they arrive in the right order.

### Fixes
- Using `Cmd + S`/`Ctrl + S` in nested structures/objects with date fields does not corrupt data anymore
#6390

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Input/DateInput.vue
+++ b/panel/src/components/Forms/Input/DateInput.vue
@@ -17,6 +17,8 @@
 		@keydown.down.stop.prevent="onArrowDown"
 		@keydown.up.stop.prevent="onArrowUp"
 		@keydown.enter.stop.prevent="onEnter"
+		@keydown.meta.s.stop.prevent="onEnter"
+		@keydown.ctrl.s.stop.prevent="onEnter"
 		@keydown.tab="onTab"
 	/>
 </template>
@@ -137,13 +139,6 @@ export default {
 			immediate: true
 		}
 	},
-	mounted() {
-		// make sure to commit input value when Cmd+S is hit
-		this.$events.on("keydown.cmd.s", this.onBlur);
-	},
-	destroyed() {
-		this.$events.off("keydown.cmd.s", this.onBlur);
-	},
 	methods: {
 		/**
 		 * Increment/decrement the current dayjs object based on the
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):

bezin:
Fantastic, thanks for this fix, much appreciated! Shouldn't this be target at 4.3.1, though?

distantnative:
@bezin I only like to include fixes that are very simple in patch releases. When we're super sure of no regressions. As they don't get any release candidate.

I'm not that 100% sure with this one. So I rather want to include it in 4.4

PR #6512, opened 2024-06-27
Changes A: `k-form-controls`

distantnative:
### Summary of changes
- New `<k-form-controls>` component that takes care of the "dumb" UI for save/publish buttons and dropdown based on different props

### Reasoning
- No breaking changes with `k-form-buttons` for now
- Move logic outside of the UI component (subsequent steps: `panel.content` module and `<k-model-form-controls>` wrapper component that connects the `panel.content` modele getters and methods with the UI)
- See `/panel/lab/components/formcontrols` for UI state (interactive) tests

### Additional context
- I will use alphabetical (A, B, C...) numbering for frontend changes PRs

### For review team

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -372,6 +372,8 @@
 	"form.discard": "Discard changes",
 	"form.discard.confirm": "Do you really want to <strong>delete all unpublished changes</strong>?",
 	"form.locked": "This content is disabled for you as it is currently edited by another user",
+	"form.preview": "Preview changes",
+	"form.preview.draft": "Preview draft",
 	"form.publish": "Publish",
 	"form.published": "Published",
 	"form.save": "Save",
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -369,6 +369,14 @@
 
 	"filter": "Filter",
 
+	"form.discard": "Discard changes",
+	"form.discard.confirm": "Do you really want to <strong>delete all unpublished changes</strong>?",
+	"form.locked": "This content is disabled for you as it is currently edited by another user",
+	"form.publish": "Publish",
+	"form.published": "Published",
+	"form.save": "Save",
+	"form.saved": "Saved",
+
 	"hide": "Hide",
 	"hour": "Hour",
 	"hue": "Hue",
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -374,10 +374,6 @@
 	"form.locked": "This content is disabled for you as it is currently edited by another user",
 	"form.preview": "Preview changes",
 	"form.preview.draft": "Preview draft",
-	"form.publish": "Publish",
-	"form.published": "Published",
-	"form.save": "Save",
-	"form.saved": "Saved",
 
 	"hide": "Hide",
 	"hour": "Hour",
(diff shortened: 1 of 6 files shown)


i18n/translations/en.json
```diff
@@ -369,6 +369,16 @@
...
 
+	"form.discard": "Discard changes",
+	"form.discard.confirm": "Do you really want to <strong>delete all unpublished changes</strong>?",
+	"form.locked": "This content is disabled for you as it is currently edited by another user",
+	"form.preview": "Preview changes",
+	"form.preview.draft": "Preview draft",
+	"form.publish": "Publish",
+	"form.published": "Published",
+	"form.save": "Save",
+	"form.saved": "Saved",
```
bastianallgeier: `Thought` I'm always torn with translation variables. `save` and `saved`, `publish` and `published` are universally useful and we already have `save` and `saved` at least. Do we prefer more of those global translation strings or do we rather clarify the usage here by prefixing it with form? I see pros and cons in both ways.

panel/src/components/Forms/Controls/Controls.vue
```diff
@@ -0,0 +1,151 @@
...
+				}
+			});
+		},
+		locked() {
+			this.$panel.notification.open({
+				icon: "lock",
+				theme: "negative",
+				message: this.$t("form.locked")
+			});
+		}
```
bastianallgeier, line 148: What do you think of keeping this even dumber and emitting @discard and @locked events instead, or do we make our lives more complicated than they should be?

panel/src/components/Forms/index.js
```diff
@@ -3,6 +3,7 @@ import Autocomplete from "./Autocomplete.vue";
 import Counter from "./Counter.vue";
 import Form from "./Form.vue";
 import FormButtons from "./FormButtons.vue";
+import Controls from "./Controls/Controls.vue";
```
bastianallgeier, line 6: I suppose you have plans to add more Controls components to that folder, right?

distantnative: Yes, at least `<k-model-form-controls>`.

panel/src/components/Forms/Controls/Controls.vue
distantnative: I had that before. It felt worse to me, cluttering up the `panel.content` module. I perceive this still as dumb UI. There should be no discard event triggering anything before discard wasn't confirmed.  So I liked having these UI parts still in this component, so that the module really can focus more on the logic.

i18n/translations/en.json
distantnative: I can see the argument for save/saved/publish/published. I think the others are too specific to the context and we shouldn't try to use them as generics.

panel/src/components/Forms/Controls/Controls.vue
bastianallgeier: Ok, makes sense to me.

i18n/translations/en.json
bastianallgeier: I agree. Maybe we can create `publish` and `published` as generic ones in this pr, reuse the generic `save` and `saved` and move the rest to `form.` (as you already have)?

bastianallgeier (approved):

PR #6513, opened 2024-06-27
Changes B: `panel.content`

distantnative:
### Summary of changes
- New `panel.content` module aimed at taking over the intermediary role between UI and API (when removing Vuex store)
- Currently serves as interface to the Vuex store
- Started to use `this.$panel.content.` methods instead of `this.$store.dispatch("content/...")`

### Reasoning
- Allows us to think about the ideal JS interface we would want for the content
- Already starts separating the Vuex store from the rest of the code, allowing us later to more easily replace it with backend API

### Enhancements
- New `panel.content`/`this.$content` module to access and alter the current model content (unsaved changes)

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/FormButtons.vue
+++ b/panel/src/components/Forms/FormButtons.vue
@@ -186,7 +186,7 @@ export default {
 		},
 		download() {
 			let content = "";
-			const changes = this.$panel.content.changed;
+			const changes = this.$panel.content.changes;
 
 			for (const key in changes) {
 				const change = changes[key];
--- a/panel/src/components/Forms/FormButtons.vue
+++ b/panel/src/components/Forms/FormButtons.vue
@@ -51,12 +51,12 @@ export default {
 			if (this.mode === "lock") {
 				return [
 					{
-						icon: this.$content.lock.unlockable ? "unlock" : "loader",
+						icon: this.$panel.content.lock.unlockable ? "unlock" : "loader",
 						text: this.$t("lock.isLocked", {
-							email: this.$esc(this.$content.lock.email)
+							email: this.$esc(this.$panel.content.lock.email)
 						}),
 						title: this.$t("lock.unlock"),
-						disabled: !this.$content.lock.unlockable,
+						disabled: !this.$panel.content.lock.unlockable,
 						click: () => this.unlock()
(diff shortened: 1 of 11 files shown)


bastianallgeier:
Based on the other discussion about `$notification`, I think we really need to decide first if we want more or less shortcuts.

distantnative:
@bastianallgeier I think we need to decide that before merging, yes. But you can review it already without that being decided, it's such a minor detail of the syntax we can clean up at the end, if needed.

bastianallgeier:
Yes, definitely!

panel/src/panel/content.js
```diff
@@ -0,0 +1,118 @@
...
+ */
+export default (panel) => {
+	return {
+		/**
+		 * Returns all fields and their values that
+		 * have been changed but not yet published
+		 *
+		 * @returns {Object}
+		 */
+		get changed() {
```
bastianallgeier: Reading through the other changed files, I lean towards `this.$content.changes` for this. I assume that you thought about `changed` & `published` and I really like this as well. But we also use `changes` as version id, so I think it would be more consistent.

panel/src/components/Forms/FormButtons.vue
```diff
@@ -121,10 +118,10 @@ export default {
 			return null;
 		},
 		lockState() {
-			return this.supportsLocking && this.lock ? this.lock.state : null;
+			return this.supportsLocking && this.$content.lock?.state;
 		},
 		supportsLocking() {
-			return this.lock !== false;
+			return this.$content.lock !== false;
```
bastianallgeier: `Suggestion`: It might be cool to have `this.$content.supportsLocking` as well.

distantnative: I left this out as I right now I am doubtful we will need it in the future. My thoughts right now are that locking would completely be dealt with on the backend (if it's not supported, not done). And the Panel receives just the info whether something is locked. If locking isn't supported, that info will just always be empty.

bastianallgeier: Good point. We can already hand it over as a view prop

bastianallgeier (approved):

PR #6514, opened 2024-06-27
`this.$notification`

distantnative:
### Summary of changes
- New `this.$notification` shortcut for Vue components

### Enhancements
- New `this.$notification` shortcut for Vue components

```js
this.$notification.success("Yay!");
```

### For review team

### DIFF
--- a/panel/lab/components/boxes/6_buttons/index.vue
+++ b/panel/lab/components/boxes/6_buttons/index.vue
@@ -6,7 +6,7 @@
 				icon="add"
 				theme="empty"
 				text="Click to add something …"
-				@click.native="$panel.notification.success('Yay, you clicked the box')"
+				@click.native="$notification.success('Yay, you clicked the box')"
 			/>
 		</k-lab-example>
 	</k-lab-examples>
@@ -15,7 +15,7 @@
 <script>
 export default {
 	props: {
-		themes: Array,
-	},
+		themes: Array
+	}
 };
 </script>
--- a/panel/lab/internals/panel/notification/index.vue
+++ b/panel/lab/internals/panel/notification/index.vue
@@ -7,11 +7,11 @@
 					<code>window.panel.notification.context</code> returns context where
 					the notification should be displayed (view, dialog, drawer).
 				</p>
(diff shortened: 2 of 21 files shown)


bastianallgeier:
As far as I remember, we had $notification in 3.x and then removed it in 4.x. Could you explain why we should bring it back?

distantnative:
@bastianallgeier Oh did we? I was writing another notification call yesterday and was wondering why we make it so "wordy" and don't have a shortcut for it if it is such a common thing (also a thing we want plugin developers to use).

Do you remember why we removed it?

Considering that is in legacy.js, was our idea to remove all those shortcuts? I'm not sure that would be so great. But I also don't remember what the intention behind it was.

bastianallgeier:
Yes, everything in legacy.js was intended to be removed at some point. I really think it was all about using $panel as a single hub and to remove the shortcuts. With shortcuts, there's always the question about the right way. Should I use the long version or the shortcut? Do we mix them? What should plugin developers do? Then I kind of remember that we also wanted to clean up for Vue 3, but I don't know if this was really something that needs to be done at all. 

TBH I'm not sure about all of this. I agree that some of the long $panel versions are quite annoying.

distantnative:
I guess our thinking then was maybe also that having only `this.$panel` would make it very easy to replace it, if needed, later on with `window.panel`. But it's also not that hard to replace `this.$notification` by search and replace with `window.panel.notification`.

I do like the look of `this.$content.set()` more than `this.$panel.content.set()` but I wouldn't die on that hill. Especially if there are practical advantages of not having the shortcuts. Though we do use other shortcuts like `this.$dialog` still a lot (and push it to the developers in that way I think). So if not doing shortcuts, we should go the verbose way of `this.$panel.dialog` etc. for all parts.

bastianallgeier:
I think you are right about `window.panel` and `$panel`. But as you write, we are really not very consistent with all of this so far. I would suggest that we don't introduce new shortcuts for now until we've made up our minds. If we decide to go without shortcuts entirely, we should really start refactoring them in all other places as well. But this should not affect the chain of changes PRs.

PR #6523, opened 2024-07-04
Fix language permissions

afbora:
## Description

Fixed the issue with adding new checks to `Language::create` and `Language::delete` methods.

### Additional context

Currently users can create any language with `Language::create()` method without authenticate, so no needed to impersonate. I kept this behavior with following syntax. But I'm not sure this is correct behavior? For panel, users have to permission to create languages, for frontend users don't have to any permission. Please direct me.

````php
if ($kirby->user()?->role()->permissions()->for('languages', 'create') === false) {
  throw new PermissionException('You are not allowed to create a language');
}
````

### Fixes
- Language permissions not working getkirby/backlog#89

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -6,6 +6,7 @@ use Kirby\Data\Data;
 use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Exception\LogicException;
+use Kirby\Exception\PermissionException;
 use Kirby\Filesystem\F;
 use Kirby\Toolkit\Locale;
 use Kirby\Toolkit\Str;
@@ -145,8 +146,13 @@ class Language
 	 */
 	public static function create(array $props): static
 	{
+		$kirby = App::instance();
+
+		if ($kirby->user()?->role()->permissions()->for('languages', 'create') === false) {
+			throw new PermissionException('You are not allowed to create a language');
+		}
+
 		$props['code'] = Str::slug($props['code'] ?? null);
-		$kirby         = App::instance();
 		$languages     = $kirby->languages();
 
 		// make the first language the default language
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/Cms/Language.php [hunk 1]
```diff
@@ -145,8 +146,13 @@ public function code(): string
 	 */
 	public static function create(array $props): static
 	{
+		$kirby = App::instance();
+
+		if ($kirby->user()?->role()->permissions()->for('languages', 'create') === false) {
```
bastianallgeier, line 151: Is this correct if there's no authenticated user? I would assume that `$kirby->user()?` returns null here, which would then not match `false` and could bypass the permssion check.

src/Cms/Language.php
```diff
@@ -206,6 +212,10 @@ public function delete(): bool
 		$kirby = App::instance();
 		$code  = $this->code();
 
+		if ($kirby->user()?->role()->permissions()->for('languages', 'delete') === false) {
```
bastianallgeier, line 215: Same here

src/Cms/Language.php
afbora: That's right @bastianallgeier . That's why I wrote for this case in `Additional context` part of the PR description :)

Btw currently there is already no check for the creating and deleting languages. I kept this behavior and that's why I need your suggestion. I I remove the `null` check, that means breaking change. 

So what do you think?

bastianallgeier: Sorry for completely missing this in your summary. You explained it perfectly there. I really think that this is something we need to fix in this case more strictly and definitely ask for an authenticated user. It's indeed a breaking change, but one that I think is necessary. Let's take the discussion about it to Discord.

PR #6527, opened 2024-07-13
Remove container query polyfill

distantnative:
### Summary of changes
- Remove container query polyfill 
- Raise necessary browser requirements

### Reasoning
I think it's time that we can raise those browser requirements and get rid of the polyfill to reduce bundle size.

### Breaking changes
- Kirby requires the following browser versions to use the Panel (other browser requirements remain unchanged)
  - Safari 16+
  - Mobile Safari 16+
  - Android Browser 126+
  - Chrome for Android 126+

### For review team

### DIFF
--- a/panel/package-lock.json
+++ b/panel/package-lock.json
@@ -7,7 +7,6 @@
 			"name": "kirby",
 			"dependencies": {
 				"autosize": "^6.0.1",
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):
Yep, v5 is definitely the right release for this.

PR #6528, opened 2024-07-13
Panel theme: allow resetting to system default

distantnative:
### Summary of changes
- Theme button: now as dropdown with additional option to reset to using system default 

Screenshot 2024-07-13 at 12 37 23

### Reasoning
Before this, there is no way to use the system default again, once set a theme (besides going into the dev tools and removing localStorage).

### Changes compared to alpha.1
- Theme button is now a dropdown and allows to reset to match system default color scheme

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -27,9 +27,6 @@
 	"date": "Date",
 	"date.select": "Select a date",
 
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):
This is awesome! I also really love the theme module for the panel.

PR #6529, opened 2024-07-13
Turn `panel` JS modules fully reactive

distantnative:
### Summary of changes
- Wrap all relevant `panel` JS modules in Vue's `reactive()`

### Reasoning
While in Vue 2 as it seems no problem, in Vue 3 the modules aren't fully reactive without these. In particular, when not accessed through `this.$panel` which already receives `reactive()` from the `return reactive(this)` in panel.js. When internally accessing the modules in the panel JS code, they are not going through the proxies created from `reative()` and thus Vue 3 doesn't register them as altered then.

I tired to apply it less broadly, e.g. just on `state.js` but that didn't work.

### DIFF
--- a/panel/src/panel/activiation.js
+++ b/panel/src/panel/activiation.js
@@ -1,18 +1,29 @@
+import { reactive } from "vue";
+import State from "./state.js";
+
+export const defaults = () => {
+	return {
+		isOpen: sessionStorage.getItem("kirby$activation$card") !== "true"
+	};
+};
+
 /**
  * @since 4.0.0
  */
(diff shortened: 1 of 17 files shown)


bastianallgeier:
Is there a reason why this step needs to go into alpha.3?

distantnative:
@bastianallgeier no, more a signal that there's less priority for this as this is mainly needed for Vue 3. and depending if we'd want to release alpha.2 soon, this could wait.

bastianallgeier (approved):

PR #6530, opened 2024-07-13
Panel search improvements

### LINKED ISSUE #6409: Search in panel shows old results when response times from server vary

distantnative:
### Summary of changes
- New `k-search-bar` component to separate UI logic from API interaction
- New `panel.searcher` module
- Ongoing Panel search request get aborted when new one is started
- Search view: add UI to differentiate searching states

### Reasoning
- A "dumber" component that only takes care of the UI logic makes it easier to test the different states with Lab examples.

### Additional context
For testing `k-search`: https://sandbox.test/panel/lab/components/search

### Fixes
- Fixed parallel Panel search requests
https://github.com/getkirby/kirby/issues/6409

### For review team


panel/src/panel/panel.js
```diff
@@ -72,6 +73,7 @@ export default {
 		this.activation = Activation(this);
 		this.drag = Drag(this);
 		this.events = Events(this);
+		this.searcher = Search(this);
```
distantnative, line 76: `Ideas?` for a better name than `panel.searcher`? `panel.search` is blocked by the method name

panel/src/components/Dialogs/SearchDialog.vue
```diff
@@ -182,52 +65,8 @@ export default {
 .k-search-dialog {
 	--dialog-padding: 0;
 	--dialog-rounded: var(--rounded);
-	overflow: visible;
```
distantnative: `Question:` Anyone recall what this was for? With it the border-radius has no effect. Removing it, I din't spot any problems.

panel/src/panel/panel.js
afbora: finder?

panel/src/components/Dialogs/SearchDialog.vue
bastianallgeier: I think it was about the outline for the input/dropdown button.

bastianallgeier (commented):
I really like the abstraction. I would vote for either choosing k-search and k-search-bar though. Right now, the component is called k-search but the class is k-search-bar. I think I'm leaning more towards k-search-bar, but I'm open for both.

distantnative:
@bastianallgeier I guess that's a leftover from when I had it as `k-search-bar`. I am quite indifferent, so let me change it back to `k-search-bar`.

bastianallgeier (approved):
I renamed the lab example page as well and fixed a small issue with the lab example CSS. But other than that it's a great update.

PR #6531, opened 2024-07-13
View buttons: backend support

distantnative:
### Summary of changes
- New `Panel\Ui` namespace with basic `Panel\Ui\Component` and `Panel\Ui\Button` classes
  - Each component renders as an array with a `component`, unique `key` and `props` entries
- New `Panel\Ui\Buttons\ViewButtons` and `Panel\Ui\Buttons\ViewButton` classes responsible for gathering and transforming view buttons configured in blueprints, config files and/or Panel areas into PHP array definitions that can be handed to the dedicated Vue frontend components
  - Dedicated view button PHP classes for some core buttons (page status, settings, preview, languages)
- New `k-view-button` component as building block for the backend-provided data
  - Lab examples: https://sandbox.test/panel/lab/components/buttons/1_variants
- Support for Panel area `buttons` extension
- Exclude Panel area buttons config files as well as the whole lab area config from code coverage
- Using backend-defined view buttons for all core views

### Reasoning
The more we can enable from the PHP backend instead of Vue, the better.

### Feature
- Define custom view buttons in your blueprints or config
- Custom Panel are buttons extension

### Fixed
- `k-header`: better wrapping for many buttons in narrow views

## Docs

As with the previous PR, view buttons can be defined by names in a model blueprint or for all views in the config. 

### Define custom button in config

However, one can now even define whole buttons by setting an array instead of a name string. The array consist of a `component` key (optional, if omitted `k-view-button` will be used) and a `props` array.

Via config:
```php
'panel' => [
	'viewButtons' => [
		'site' => [
			'preview',
			'a' => [
				'props' => [
					'icon'   => 'heart',
					'text'   => 'Kosmos',
					'theme'  => 'purple-icon',
					'target' => '_blank',
					'link'   => 'https://getkirby.com'
				]
			],
			'b-dropdown' => [
				'props' => [
					'icon'     => 'info',
					'text'     => 'Info',
					'dropdown' => 'my/dropdown/route'
				]
			],
			'c-component' => [
				'component' => 'my-custom-component',
				'props' => [
					'foo' => 'bar'
				]
			]
		]
	]
]
```

#### Shorthand for only props

If you don't pass a component-props definition, you can also just pass the props on their own, being then applied to `k-view-button`:

```php
'panel' => [
	'viewButtons' => [
		'site' => [
			'a' => [
				'icon'   => 'heart',
				'text'   => 'Kosmos',
				'theme'  => 'purple-icon',
				'target' => '_blank',
				'link'   => 'https://getkirby.com'
			]
		]
	]
]
```

### Define custom button in a blueprint

This also works in blueprints:

```yml
buttons:
  - preview
  - settings
  foo:
      text: Foo
```

### Reusable buttons: Panel area buttons extension

If you do not want to define just one button via the config, but reuse it (or even ship it as part of your plugin), you have to add them to the Panel area extension:

```php
Kirby::plugin('custom/buttons', [
	'areas' => [
		'todos' => function () {
			return [
				'buttons' => [
					'todos.add' => function () {
						return [
							'props' => [
								'icon'   => 'add',
								'dialog' => 'todos/create'
							]
						];
					}
				]
			];
		}
	]
]);
```

You have the same options for your return value: full component-props array or just props on top-level. In addition, you can also return directly a `Kirby\Panel\Ui\Button\ViewButton` object. 

See `Kirby\Panel\Ui\Button\ViewButton` properties for available `props` options (also for array notation).

## Ready?

### For review team

  - https://sandbox.test/panel/lab/components/buttons/1_variants
  - https://github.com/getkirby/sandbox/pull/8


config/areas/files/buttons.php
```diff
@@ -0,0 +1,26 @@
...
+	},
+	'file.settings' => function (File $file) {
+		if ($file->permissions()->can('preview') === false) {
+			return false;
+		}
+
+		return [
+			'component' => 'k-view-settings-button',
+			'props'     => [
+				'dropdown' => $file->previewUrl(),
```
distantnative: `Bug` Wrong copy pasted

panel/src/components/View/Buttons/Button.vue
```diff
@@ -0,0 +1,62 @@
...
+	<div class="k-view-button">
+		<k-button v-bind="$props" :dropdown="hasDropdown" @click="onClick" />
+		<k-dropdown-content
+			v-if="hasDropdown"
+			ref="dropdown"
+			:options="options"
+			align-x="end"
+			@action="$emit('action', $event)"
+		/>
+	</div>
```
distantnative, line 15: `Comment:` It's a big ugly for now that we have to add a `<div>` for each button. With Vue 3 this will improve as we will be able to define components that have more than one root element in `<template>`. Then the button and dropdown can live side by side without a wrapper element.

panel/src/components/View/Buttons/LanguagesButton.vue
```diff
@@ -1,58 +1,35 @@
...
+				if (option === "-") {
+					return option;
+				}
 
-			return options;
+				return {
+					...option,
+					click: () => this.change(option)
+				};
+			});
```
distantnative, line 32: `Idea?` The languages button can almost be fully rendered by the backend. Just this click handler calling `this.$reload` is why we need a custom component. But I don't see a way to pass something from the backend that achieves the same.

panel/src/components/View/Buttons/SettingsButton.vue
```diff
@@ -1,49 +1,24 @@
...
-			v-if="dropdown"
-			ref="dropdown"
-			:options="$dropdown(dropdown)"
-			align-x="end"
-			@action="$emit('action', $event)"
-		/>
-	</div>
+	<k-view-button
+		v-bind="$props"
+		:disabled="$panel.content.isLocked"
```
distantnative, line 4: `Comment:` This part is currently still defined on the frontend as I think `isLocked` can dynamically change without the top-level `buttons` prop being reloaded as well. And if we then would check `isLocked` on the backend, it could become outdated. I would hope we can improve this after implementing the Changes feature and move the whole implementation to the backend, without the need for a custom Vue component.

panel/src/components/View/Buttons/StatusButton.vue
```diff
@@ -1,40 +1,22 @@
...
-		v-bind="button"
-		:responsive="true"
-		:text="status.label"
-		size="sm"
-		variant="filled"
-		class="k-view-status-button k-page-status-button"
-		@click="$dialog(model.link + '/changeStatus')"
+	<k-view-button
+		v-bind="$props"
+		:disabled="disabled || $panel.content.isLocked"
```
distantnative, line 4: `Comment:` same consideration as with the setting button.

panel/src/components/Views/Languages/LanguageView.vue
```diff
@@ -91,15 +85,6 @@ export default {
...
 		},
-		update(focus) {
-			this.$dialog(`languages/${this.id}/update`, {
-				on: {
-					ready: () => {
-						this.$panel.dialog.focus(focus);
-					}
-				}
-			});
-		},
```
distantnative, line 102: We can remove this and just use `$dialog(`languages/${this.id}/update`)` as the `focus` parameter was never passed.

src/Panel/Ui/Buttons/Button.php
```diff
@@ -0,0 +1,99 @@
...
+		Closure $button,
+		array $args = []
+	): array|null {
+		$kirby      = App::instance();
+		$controller = new Controller($button);
+
+		return $controller->call(data: [
+			'kirby' => $kirby,
+			'site'  => $kirby->site(),
+			'user'  => $kirby->user(),
```
distantnative: `Question`: what do you think about these default parameters/arguments to always pass to the Closure?

src/Panel/Ui/Buttons/Buttons.php
```diff
@@ -0,0 +1,70 @@
...
+	/**
+	 * Sets the default buttons
+	 *
+	 * @return $this
+	 */
+	public function defaults(string ...$defaults): static
+	{
+		$this->buttons ??= $defaults;
+		return $this;
+	}
```
distantnative: `Question`: What do you think about this syntax to pass defaults to the object? Allowing this:

```php
Buttons::view('test')->defaults('a', 'b', 'c')->render();
```

I think it makes the syntax attractive, but one might see it as a bit uncommon for Kirby.

distantnative (commented):
Would love to get some first conceptual reviews 🙏 I assume those will lead to further refactoring so that the final review would take another round at least.

src/Panel/Ui/Buttons/Button.php
afbora: If this usage risky for the performance, what about using `LazyValue`?

afbora:
## First review

1. Console error when no buttons set

> TypeError: Cannot read properties of undefined (reading 'map')

2. Handling mobile view

Needs to be customizable or good standart for mobile view. For ex currently broken when few buttons available for mobile view.

I couldn't find any issue for now. Looks great 🎉 I'll re-test again when you move forward.

distantnative:
1. @afbora I don't see any console errors with 

```php
'viewButtons' => [
	'site' => []
]
```

or what do you mean with `when no buttons set`?

2. They behave already just like our core buttons. You can set the `responsive` prop to customize their behavior. What do you think is broken?

afbora:
1. @distantnative Could you try to remove entire `panel.viewButtons` option?
2. Looks like following on mobile:

_removed large screenshot after issue has been fixed_

distantnative:
@afbora no error without the config option either.

Mobile issue, I see now. This is only partially related to this PR. The problem already seems to exists.

afbora:
@distantnative This issue is easy to reproduce for me. Starterkit with this branch codes. Just enable the `debug` mode.

> TypeError: Cannot read properties of undefined (reading 'map')

Related lines:
- https://github.com/getkirby/kirby/blob/v5/feature/view-buttons-php/panel/src/components/View/Buttons/LanguagesButton.vue#L2
- https://github.com/getkirby/kirby/blob/v5/feature/view-buttons-php/panel/src/components/View/Buttons/LanguagesButton.vue#L23


Also getting `undefined` in `LanguagesButton::languages()`

````js
console.log(this.options); // undefined
````

I think no options set for this vue component on single language setup.

src/Panel/Ui/Buttons/Button.php
distantnative: @afbora I checked that but we also only use it as wrapper in `App::collection` for `pages` and `users`, not for these more basic objects.

distantnative:
Actually, let's use the Changes PRs as examples and break this up in a few separate steps.

PR #6532, opened 2024-07-14
`View::data()` use `Language::toArray()`

distantnative:
### Summary of changes
- In `Panel\View::data()` use `$language->toArray()` instead of creating array manually again 
- Simplify closure by using short-arrow functions and `match()`

### Reasoning
Keep code DRY.

## Ready?

### DIFF
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -522,8 +522,7 @@ class Language implements Stringable
 	}
 
 	/**
-	 * Returns the most important
-	 * properties as array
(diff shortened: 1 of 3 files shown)


lukasbestle (approved):
Means that the responses get slightly larger with the added keys, but worth it for keeping it DRY IMO.

distantnative:
Yes, I was mostly worried about rules being the biggest. But those we include already anyways. URL and locales are probably fine for size.

PR #6538, opened 2024-07-19
Model actions: apply hooks instead of only triggering

### LINKED ISSUE #2828: multiple updates in hooks override one another
**Describe the bug**  
When having multiple plugins using the same hook to `update()` the `$file` or `$page` object, only one change actually comes through, the other gets lost or is overwritten.

**To Reproduce**  
1. Install 2 plugins that both use the `file.create:after` hook (or similar) to initially fill one field. e.g.:
```php
Kirby::plugin('kirby/plugin-1', [
  'hooks' => [
    'file.create:after' => function ($file) {
      $file->update([
        'title' => 'Title'
      ]);
    },
  ]
]);
Kirby::plugin('kirby/plugin-2', [
  'hooks' => [
    'file.create:after' => function ($file) {
      $file->update([
        'caption' => 'Caption'
      ]);
    }
  ]
]);
```
2. Upload an image
3. Check for the content file, only one of both fields got filled in

**Expected behavior**  
Would be nice, if both fields would be filled in.

**Kirby Version**  
3.4.3

**Forum discussion**
https://forum.getkirby.com/t/how-to-trigger-file-update-in-two-plugins-that-use-the-same-hook/19072/5

### LINKED ISSUE #6460: `file.create:before` hook: allow to modify `$file` object
## Description
I want to add additional data to an uploaded file in the `file.create:before` hook. (In my case: the original filename before it gets sanitized.) The hook is called but the data is not added to the file metadata.

```php
Kirby::plugin('sl/sl-plugin', [
  'hooks' => [
    'file.create:before' => function (
      Kirby\Cms\File $file,
      Kirby\Filesystem\File $upload
    ) {
      $file->update(['test' => 'test']);
    }
  ]
]);
```

**Expected behavior**  
The data to be updated and saved.

## To reproduce
1. Add a `file.create:before` hook
2. Update the `$file` in the hook
3. Upload a file
4. Check the file’s metadata

## Your setup
**Kirby Version**  
4.2.0

 - Device: MacBook Pro
 - OS: macOS 14.5
 - Browser: Safari
 - Version: 17.5

distantnative:
### Summary of changes
- Apply (not trigger) model action hooks
  - for `before` hooks the first argument will be modified (the original model), altering it will change the model being passed into the action (or subsequent hooks)
  - for `after` hooks the first after argument will be modified (most often the $newModel), altering it will change the result of `::commit()` or passed to a  subsequent hook
- Model rules run on the model after the before hooks have been applied, ensuring that even after those all rules are ensured

### Reasoning
This way multiple hooks won't overwrite anymore each other changes.

### Fixes
- Model hooks for actions: multiple hooks don't overwrite each others' changes to the model
- https://github.com/getkirby/kirby/issues/6460
- https://github.com/getkirby/kirby/issues/2828

### Breaking changes
- Model action `before` hooks: rules get applied after the hook runs

## Docs

- Need to modify reference docs info
- Add info box to encourage returning the model, if hook does alter it

## Ready?

### For review team

### DIFF
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -162,24 +162,41 @@ trait FileActions
 		array $arguments,
 		Closure $callback
 	): mixed {
-		$old            = $this->hardcopy();
-		$kirby          = $this->kirby();
-		$argumentValues = array_values($arguments);
+		$kirby = $this->kirby();
+		$old   = $this->hardcopy();
+		$apply = array_key_first($arguments);
 
-		$this->rules()->$action(...$argumentValues);
-		$kirby->trigger('file.' . $action . ':before', $arguments);
+		// run before hook and apply result to first argument
+		$arguments[$apply] = $kirby->apply(
+			'file.' . $action . ':before',
+			$arguments,
+			$apply
+		);
 
-		$result = $callback(...$argumentValues);
+		$values = array_values($arguments);
 
-		$argumentsAfter = match ($action) {
+		// check page rules
+		$this->rules()->$action(...$values);
+
+		// run closure
+		$result = $callback(...$values);
+
+		// determine arguments for after hook depending on the action
+		$after = match ($action) {
 			'create' => ['file' => $result],
 			'delete' => ['status' => $result, 'file' => $old],
 			default  => ['newFile' => $result, 'oldFile' => $old]
 		};
 
-		$kirby->trigger('file.' . $action . ':after', $argumentsAfter);
+		// run after hook and apply return as result
+		$result = $kirby->apply(
+			'file.' . $action . ':after',
+			$after,
+			array_key_first($after)
+		);
 
 		$kirby->cache('pages')->flush();
+
 		return $result;
 	}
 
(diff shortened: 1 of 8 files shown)


src/Cms/FileActions.php
```diff
@@ -151,35 +151,61 @@ public function changeTemplate(string|null $template): static
...
+			$appliedTo
+		);
+
+		// get just the values of argument, so that spreading them
+		// as arguments in a method call gets only assigned by order
+		// not by key name
 		$argumentValues = array_values($arguments);
 
+		// check file rules
 		$this->rules()->$action(...$argumentValues);
```
distantnative: `Question:` Should the rules run twice? Before and after the `before` hook?

bastianallgeier: Hm, very good question. I think the hook should really only run if the model passes examination. That's why I would say yes. But I'm not sure if running the rules twice could get us into trouble somehow. On the other hand, the hook really has the potential of changing the model in a way that would needs to be checked and maybe rejected again. I think I stick to yes: before and after the hook is probably needed.

bastianallgeier (approved):

PR #6548, opened 2024-07-21
`Permissions::for()`: new `$default` argument

distantnative:
### Summary of changes
- `Permissions::for()` has a new `$default` parameter specifying what is returned when the permission category or action is not found
- Implemented `$default` also for `ModelPermissions::can()` and `::cannot()`

### Reasoning
Currently, if one wants to check for a permission but by default wants to grant it, there is no easy way. Instead one has to go through `->toArray()` which isn't ideal. Allowing to also use `->for()` improves this.

### Enhancements
- `$permissions->for()`, `$modelpermissions->can()` and `$modelpermissions->cannot()` accept a new `$default` parameter

## Ready?

### For review team

### DIFF
--- a/src/Cms/ModelPermissions.php
+++ b/src/Cms/ModelPermissions.php
@@ -43,6 +43,12 @@ abstract class ModelPermissions
 		return $this->toArray();
 	}
 
+	/**
+	 * Returns whether the current user is allowed to do
+	 * a certain action on the model
+	 *
+	 * @param bool $default Will be returned if $action does not exist
+	 */
 	public function can(
 		string $action,
 		bool $default = false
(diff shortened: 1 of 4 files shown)


src/Cms/ModelPermissions.php
```diff
@@ -95,12 +97,14 @@ public function can(string $action): bool
...
-		return $this->permissions->for($this->category, $action);
+		return $this->permissions->for($this->category, $action, $default);
 	}
 
-	public function cannot(string $action): bool
-	{
-		return $this->can($action) === false;
+	public function cannot(
+		string $action,
+		bool $default = false
```
bastianallgeier: This is creating a knot in my brain. I understand why you chose false here, but is it still logical? Wouldn't you read it like true === cannot & false === can?

distantnative: I see your point. I think we could do for the parameter `bool $default = true` and then the code reads something like this?

```php
return $this->can($action, !$default) === false;
```

bastianallgeier: Yes, I really just wonder which way is better. It's one of those situations where both solutions are correct and logically explainable and no matter how we solve it, it will always be one of those arguments where you need to look up how it was meant to be used afterwards.

distantnative: I would say `bool $default = true` and `$this->can($action, !$default) === false` makes most sense. And then a DocBlock comment:

```
@param bool $default Will be returned if $action does not exist
```

distantnative: Changed it.

bastianallgeier (approved):

PR #6555, opened 2024-07-22
`A::random()`: exception if $count exceeds array

### LINKED ISSUE #5934: A::random (Error when the limit is bigger than the array)
## Description

Error:

```
array_rand(): Argument #2 ($num) must be between 1 and the number of elements in argument #1 ($array)
```

Source
```
	public static function random(
		array $array,
		int $count = 1,
		bool $shuffle = false
	): array {
		if ($shuffle === true) {
			return array_slice(self::shuffle($array), 0, $count);
		}

		if ($count === 1) {
			$key = array_rand($array);
			return [$key => $array[$key]];
		}

		return self::get($array, array_rand($array, $count));
	}
```

fix would be:

```
	public static function random(
		array $array,
		int $count = 1,
		bool $shuffle = false
	): array {
		if ($shuffle === true) {
			return array_slice(self::shuffle($array), 0, $count);
		}

                $arrayCount = count($array);

                // Return the entire array when count is greater or equal to the array size
                if ($count >= $arrayCount) {
                    return $array; 
                }

		if ($count === 1) {
			$key = array_rand($array);
			return [$key => $array[$key]];
		}

		return self::get($array, array_rand($array, $count));
	}
```

## Your setup
kirby cms 4.beta1

distantnative:
### Fixes
- `A::random()` throws exception instead of error when `$count` is higher than array length

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -738,12 +738,18 @@ class A
 	/**
 	 * Returns a number of random elements from an array,
 	 * either in original or shuffled order
+	 *
+	 * @throws \Exception When $count is larger than array length
 	 */
 	public static function random(
 		array $array,
 		int $count = 1,
 		bool $shuffle = false
 	): array {
+		if ($count > count($array)) {
+			throw new Exception('$count is larger than available array items');
+		}
+
 		if ($shuffle === true) {
 			return array_slice(self::shuffle($array), 0, $count);
 		}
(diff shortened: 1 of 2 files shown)


src/Toolkit/A.php
```diff
@@ -738,12 +738,18 @@ public static function nestByKeys($value, array $keys)
...
+	 *
+	 * @throws \Exception When $count is larger than array length
 	 */
 	public static function random(
 		array $array,
 		int $count = 1,
 		bool $shuffle = false
 	): array {
+		if ($count > count($array)) {
+			throw new Exception('$count is larger than available array items');
```
bastianallgeier: I think it would be nice to throw an InvalidArgumentException here.

bastianallgeier (approved):

PR #6556, opened 2024-07-22
`k-button`: add `aria-label`

### LINKED ISSUE #5899: [v4] Improve accessibility for icon-only buttons
## Description
Icon-only buttons get a `title` attribute, which is helpful for mouse interaction. But they ~~also~~ rather need an `aria-label` for better accessibility.

It is also not possible to add an `aria-label` to a `<k-button>` in a custom panel plugin. Probably because this attribute is not passed to the [`<component>`](https://github.com/getkirby/kirby/blob/v4/main/panel/src/components/Navigation/Button.vue#L2)?  

## Your setup
Kirby 4 beta 2

distantnative:
### Summary of changes
- Uses the title or text of a button to set `aria-label`

### Reasoning
https://github.com/getkirby/kirby/issues/5899

### Enhancements
- `k-button`: use title or text of a button to set `aria-label`

## Ready?

- [x] ˜Unit tests for fixed bug/feature˜

### For review team

### DIFF
--- a/panel/src/components/Navigation/Button.vue
+++ b/panel/src/components/Navigation/Button.vue
@@ -138,7 +138,7 @@ export default {
 			const attrs = {
 				"aria-current": this.current,
 				"aria-disabled": this.disabled,
-				"aria-label": this.title ?? this.text,
+				"aria-label": this.text ?? this.title,
 				"aria-selected": this.selected,
 				"data-responsive": this.responsive,
 				"data-size": this.size,
--- a/panel/src/components/Navigation/Button.vue
+++ b/panel/src/components/Navigation/Button.vue
@@ -138,6 +138,7 @@ export default {
 			const attrs = {
 				"aria-current": this.current,
 				"aria-disabled": this.disabled,
+				"aria-label": this.title ?? this.text,
 				"aria-selected": this.selected,
 				"data-responsive": this.responsive,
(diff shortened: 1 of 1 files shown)


bastianallgeier:
This would also add the button text as aria label when there's actually text for the button. Would screen readers read the text twice in this case?

distantnative:
My understanding from https://www.sarasoueidan.com/blog/accessible-icon-buttons/#technique-%233%3A-using-aria-label would be that it overwrites it. But sadly not sufficiently knowledge about a11y myself that I would guarantee it.

bastianallgeier:
Ok, that sounded good at first, but I think the second part could be a problem for us when we auto-assign the title attribute because it could totally happen that the title is different from the text. 

Screenshot 2024-07-24 at 10 10 44

I wonder if we could fix this by switching the order. 

We take the button text as label if it exists. Otherwise we take the title because then it means that there's no button text at all and probably just an icon.

distantnative:
This leads to almost all buttons having an `aria-label` but I guess that's fine.

bastianallgeier (approved):

bastianallgeier:
I think we could probably still make the logic behind it more complex. But that's also something that we can solve in further steps. Maybe we should also have a new "label" prop or a dedicated aria-label prop or something like that.

PR #6558, opened 2024-07-22
Fix `focus` helper for non-native inputs

### LINKED ISSUE #6347: Autofocus in modal should also focus non-native input elements
## Description
I have a structure with two tags fields at the beginning:

```yml
credits:
  type: structure
  fields:
    role:
      type: tags
    name:
      type: tags
    website:
      type: url
```

Whenever I add a new entry and the fields slide in from the side, the website field gets focused.

**Expected behavior**  
The first field – the role field – should get focused.

**Screenrecording**  

https://github.com/getkirby/kirby/assets/60777/7cd6fe2e-d40f-4fc1-b7db-ef2bd9b37577

## Your setup
**Kirby Version**  
4.1.2

distantnative:
### Summary of changes
- `focus` helper treats certain groups of CSS selectors now as equals
- set `input-focus` class on non-native inputs to receive equal focus importance as input, textarea etc.

### Reasoning
In dialogs etc., the focus helper should try helping select the first input. However, if it's a non-native input (e.g. tags input), the focus selects the second field instead. Treating the set `:where(input, textarea, select, [contenteditable=true], .input-focus)` as equals, the helper should be able to always focus on the first field.

### Fixes
- #6347

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Field/BlocksField.vue
+++ b/panel/src/components/Forms/Field/BlocksField.vue
@@ -10,6 +10,7 @@
 					icon="add"
 					variant="filled"
 					size="xs"
+					class="input-focus"
 					@click="$refs.blocks.choose(value.length)"
 				/>
 				<k-button
--- a/panel/src/components/Forms/Field/LayoutField.vue
+++ b/panel/src/components/Forms/Field/LayoutField.vue
@@ -8,6 +8,7 @@
 					icon="add"
 					variant="filled"
 					size="xs"
+					class="input-focus"
 					@click="$refs.layouts.select(0)"
 				/>
 				<k-button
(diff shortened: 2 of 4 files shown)


bastianallgeier:
`Question`: I wonder if it is a good idea to use a css class for this or if we should rather use a data attribute.

distantnative:
I"m fine with either, so if you prefer the data attribute, I'll adapt it.

distantnative:
Looks like data attributes aren't as easily passed on when applied to a custom component :/

bastianallgeier:
Ah, that makes sense. I'm fine with the class in this case. A bit of a `nitpick`: what do you think about `focusable` instead to make it a bit more descriptive?

distantnative:
To me `focusable` would be a bit too generic as items lower in the `focus.js` stack are also focusable. But here it should be more specific about being focusable as an input.

bastianallgeier:
Good point! ✅

bastianallgeier (approved):

PR #6560, opened 2024-07-22
Fix `new Uri()` for relative URLs with param colon

### LINKED ISSUE #6331: can't `parse_url()` for creating URIs that use pagination logic
## Description
I've been following the [cookbook to setup 'Load More' Functionality with Kirby Pagination](https://getkirby.com/docs/cookbook/templating/ajax-load-more).

The model named 'Rewards' is using virtual pages, to fetch the children of the model from an API.

I'm using [page cache](https://getkirby.com/docs/guide/cache) to cache the content of the page. 
During the `function children()`, when I call the children of the model over API, I clear the page cache by using `kirby()->cache('pages')->remove('rewards.html')` in certain conditions.

This accessing the page cache is not working once I click the 'Load More' Button, that calls the paginated children from the `rewards.json.php`.

**Expected behavior**  

The 'Load More' Button should fetch the children in the same way and display them, as an initial page load should be able to as well.

**Actual behaviour**

When I use the AJAX 'Load More' Button as outlined in the cookbook, the request fails upon the moment of accessing the page cache.

I've traced it down to `kirby/src/Http/Uri.php:91`, during accessing the page cache, it tries to build a URI, but somehow the `parse_url()` function from php can't build a url from the pagination url `"/rewards.json/page:2"` that is being sent with the call.
I guess because of the colon `:`. 

## Stack trace

```bash
"NOTICE: PHP message: Whoops\Exception\ErrorException: Automatic conversion of false to array is deprecated in /var/www/html/kirby/src/Http/Uri.php:92"
"Stack trace:"
"#0 /var/www/html/kirby/src/Http/Uri.php(92): Whoops\Run->handleError(8192, 'Automatic conve...', '/var/www/html/k...', 92)"
"#1 /var/www/html/kirby/src/Http/Uri.php(214): Kirby\Http\Uri->__construct(Array, Array)"
"#2 /var/www/html/kirby/src/Cms/App.php(1662): Kirby\Http\Uri::current(Array)"
"#3 /var/www/html/kirby/src/Cms/System.php(159): Kirby\Cms\App->url('index', true)"
"#4 /var/www/html/kirby/src/Cms/AppCaches.php(80): Kirby\Cms\System->indexUrl()"
"#5 /var/www/html/kirby/src/Cms/AppCaches.php(32): Kirby\Cms\App->cacheOptions('pages')"
"#6 /var/www/html/site/models/rewards.php(29): Kirby\Cms\App->cache('pages')"
"#7 /var/www/html/site/controllers/rewards.json.php(6): RewardsPage->children()"
"#8 [internal function]: Kirby\Cms\App->{closure}(Object(RewardsPage))"
"#9 /var/www/html/kirby/src/Toolkit/Controller.php(60): Closure->call(Object(Kirby\Cms\App), page..."
```

## To reproduce

1. try `parse_url('/rewards.json/page:2')` that is used in the mentioned file
2. it will give you false, because the URL is malformed.

_-> Is there something wrong with this URI Constructor?_

## Your setup
**Kirby Version**  
getkirby/cms (4.1.1) composer install

distantnative:
## Description

With this change an existing unit test for `Environment` fails. I think the test has been wrong:

`detectRequestUri()` receives `index.php?foo=bar` as input and expects

```php
'path'  => '',
'query' => 'foo=bar'
```

as result. However, before this PR `detectRequestUri` would concat the string with `https://getkirby.com` without trailing slash.

So it would actually examine `https://getkirby.comindex.php?foo=bar` for the test in question. Which seems wrong. When actually running it on `https://getkirby.com/index.php?foo=bar` as with this PR now, the result is `'path'  => 'index.php',`.

@lukasbestle @bastianallgeier Is this unexpected/a problem if we change this and adapt the test?

### Fixes
- Fixed `new Http\Uri()` for relative URLs with a colon inside
#6331

## Ready?

### For review team

### DIFF
--- a/src/Http/Environment.php
+++ b/src/Http/Environment.php
@@ -602,13 +602,7 @@ class Environment
 	 */
 	protected function detectRequestUri(string|null $requestUri = null): Uri
 	{
-		// make sure the URL parser works properly when there's a
-		// colon in the request URI but the URI is relative
-		if (Url::isAbsolute($requestUri) === false) {
-			$requestUri = 'https://getkirby.com' . $requestUri;
-		}
-
-		$uri = new Uri($requestUri);
+		$uri = new Uri($requestUri ?? '');
 
 		// create the URI object as a combination of base uri parts
 		// and the parts from REQUEST_URI
--- a/src/Http/Uri.php
+++ b/src/Http/Uri.php
@@ -88,7 +88,16 @@ class Uri
 	public function __construct(array|string $props = [], array $inject = [])
 	{
 		if (is_string($props) === true) {
-			$props = parse_url($props);
+			// make sure the URL parser works properly when there's a
+			// colon in the string but the string is a relative URL
+			if (Url::isAbsolute($props) === false) {
+				$props = 'https://getkirby.com/' . $props;
+				$props = parse_url($props);
+				unset($props['scheme'], $props['host']);
+			} else {
+				$props = parse_url($props);
+			}
+
 			$props['username'] = $props['user'] ?? null;
 			$props['password'] = $props['pass'] ?? null;
 
--- a/tests/Http/EnvironmentTest.php
+++ b/tests/Http/EnvironmentTest.php
@@ -1407,7 +1407,7 @@ class EnvironmentTest extends TestCase
 			[
 				'index.php?foo=bar',
 				[
-					'path'  => '',
+					'path'  => 'index.php',
 					'query' => 'foo=bar'
 				]
 			],
--- a/tests/Http/UriTest.php
+++ b/tests/Http/UriTest.php
@@ -325,6 +325,15 @@ class UriTest extends TestCase
 				'/search/page:2?q=something'
 			],
 
+			// relative path with colon + adding query
+			[
+				'/search/page:2',
+				[
+					'query' => ['q' => 'something']
+				],
+				'/search/page:2?q=something'
+			],
+
 			// path + adding params + query
 			[
 				'https://getkirby.com/search',


bastianallgeier:
The URI class used to remove index.php from the path. I think it was mostly about making sure that `https://domain.com/something` and `https://domain.com/something/index.php` lead to the same thing. I remember that it was important for some reason further down under the hood. I suppose it was about auto URL detection. But I'm not entirely sure anymore. The concatenated path without slash is definitely wrong though.

distantnative:
@bastianallgeier I cannot reproduce that on `develop-minor`, `index.php` was already present there:

```php
$uri = new Uri('index.php?foo=bar');
$uri->toString(); // /index.php?foo=bar
$uri->path()->toString(); // index.php

$uri = new Uri('https://getkirby.com/index.php?foo=bar');
$uri->toString(); // https://getkirby.com/index.php?foo=bar
$uri->path()->toString(); // index.php
```

So what you describe doesn't seem to hold true for v4.3 at least.

This PR only exposes the bug from `Environment::detectRequestUri()` resulting from the wrongly concatenated path. The question is whether the Environment somehow depends on the false result or if changing it is just fine and only exposes the false negative unit test now, which we could simply change.

bastianallgeier:
Ok, I think my outdated assumptions are a good thing. I would vote to fix the environment unit test in this case. I double-checked the results and it is indeed more consistent and correct this way. If we really run into a regression because of it (which would surprise me) we can still consider what to do. But I guess it would be more useful to work around having index.php in the path in a different place, if that is what we want to achieve.

bastianallgeier (approved):

PR #6561, opened 2024-07-22
Changes 17: Routes

bastianallgeier:
## Description

This PR implements the changes routes for all models and their files.

### Summary of changes

- New request routes for Site, User, Page and File models
- New `Kirby\Panel\Controller\Changes` class to handle the request logic
- The `::setUpSingleLanguage` and `::setUpMultiLanguage` helper test methods have been moved to the global `TestCase` class.

### Reasoning

The routes can be implemented on the backend without any side effects already. We can fine-tune them in the next steps and then integrate them properly with the frontend without having to finalize all other content classes.

## Ready?

### For review team

### DIFF
--- a/src/Panel/Controller/Changes.php
+++ b/src/Panel/Controller/Changes.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace Kirby\Panel\Controller;
+
+use Kirby\Cms\ModelWithContent;
+use Kirby\Content\VersionId;
+use Kirby\Exception\Exception;
+use Kirby\Form\Form;
+
+/**
+ * The Changes controller takes care of the request logic
+ * to save, discard and publish changes, as well as unlocking.
+ *
+ * @package   Kirby Panel
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ */
+class Changes
+{
+	/**
+	 * Discards unpublished changes by deleting the version
+	 */
+	public static function discard(ModelWithContent $model): array
+	{
+		$model->version(VersionId::changes())->delete();
+
+		return [
+			'status' => 'ok'
+		];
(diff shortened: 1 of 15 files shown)


src/Panel/Controller/Changes.php
```diff
@@ -0,0 +1,97 @@
...
+	 */
+	public static function save(ModelWithContent $model, array $input): array
+	{
+		// we need to run the input through the form
+		// class to get a set of storable field values
+		// that we can send to the content storage handler
+		$form = Form::for($model, [
+			'ignoreDisabled' => true,
+			'input'          => $input,
+		]);
```
distantnative, line 69: `Question:` Isn't this something the `Version` class needs to handle as it would also apply to programmatic updates not through this Panel controller class?

tests/TestCase.php
```diff
@@ -92,6 +94,66 @@ protected function hasTmp(): bool
...
+				'children' => [
+					[
+						'slug'     => 'a-page',
+						'template' => 'article',
+					]
+				]
+			]
+		]);
+
+		$this->model = $this->app->page('a-page');
```
distantnative: `Concern:` I think this model part could make it hard to reuse in other tests that need e.g. more than one page in their setup.

src/Panel/Controller/Changes.php
bastianallgeier: Yes, we are getting to the point where this needs to be in the Version class, but I honestly want to spend more time on it and it should be in a separate PR in my opinion.

tests/TestCase.php
bastianallgeier: You are right. I've removed it from the global TestCase class and made the setUp methods more universally useful. I think we can add more arguments for other options like blueprints or whatever we need. But I would do this step by step.

distantnative (approved):
With recent discussions around simplifying unlock, we might not need that route after all. Let's keep it as is and change/remove later.

PR #6562, opened 2024-07-22
Writer field: fix content with non-breaking space

### LINKED ISSUE #6285: Writer: non-breaking spaces cause endless "Revert" state
## Description

The exact issue I'm seeing is described here (in the issue it was suggested to the OP to create a GH issue but it seems he never did): 

https://forum.getkirby.com/t/orange-bar-keeps-re-appearing-on-reverting-changes-in-panel/24271

I experienced the bug in Kirby v3 and am still experiencing it in Kirby v4.

It does not always happen. I have not yet been able to isolate what's needed to recreate it.

It happens often enough to be annoying and potentially quite confusing for users (e.g. resulting in not knowing if their changes were actually reverted).

**Expected behavior**  

When clicking "revert" the "Revert" and "Save" buttons should disappear. 

## Your setup
**Kirby Version**  
4.1.0

 - Device: Mac Book
 - OS: macOS
 - Browser: Brave

distantnative:
### Summary of changes
Enforcing conversion of a non-breaking space to `&nbsp;` as this is what ProseMirror internally will do automatically. And if the backend delivers the actual non-breaking space character but ProseMirror converts it to `&nbsp:` we have an unreasonable state where the current input value can never be the original value (and thus always shows as changed).

### Fixes
- Writer field: content with non-breaking changes doesn't anymore show up as changed always
#6285

## Ready?

### For review team

### DIFF
--- a/config/fields/writer.php
+++ b/config/fields/writer.php
@@ -63,7 +63,14 @@ return [
 	'computed' => [
 		'value' => function () {
 			$value = trim($this->value ?? '');
-			return Sane::sanitize($value, 'html');
+			$value = Sane::sanitize($value, 'html');
+
+			// convert non-breaking spaces to HTML entity
+			// as that's how ProseMirror handles it internally;
+			// will allow comparing saved and current content
+			$value = str_replace(' ', '&nbsp;', $value);
+
+			return $value;
 		}
 	],
 	'validations' => [


bastianallgeier:
Is there a good process to reproduce this issue? I'm trying to add non-breaking spaces, but I never really run into the issue myself on develop-minor. The fix looks good to me. I just really would love to see the difference IRL first :)

distantnative:
@bastianallgeier this https://github.com/getkirby/kirby/issues/6285#issuecomment-1959047594 worked reliably for me. I added a text with a `&nbsp:` to the default template and copied it from there.

bastianallgeier (approved):
Thanks for pointing me to that comment. I was also able to reproduce it and the fix works great.

PR #6564, opened 2024-07-23
`content.uuid.index` option: prevent index lookup

distantnative:
### Enhancement
- `content.uuid.index` option to prevent index lookup. Will throw an exception if a UUID model cannot be looked up from the cache alone. This requires you to keep a full UUID cache at all times but can be helpful for very large sites where any index lookup would run into memory limits.
  - Use `content.uuid.format`  to specify `uuid-v4` as format.

## Ready?

### For review team


bastianallgeier (approved):

bastianallgeier (approved):

bastianallgeier:
I had a second look at it and I would suggest to merge this. Since it's only activated by option, I don't see any risk here to get this into 4.4

PR #6567, opened 2024-07-23
Duplicating page: replace UUIDs with new ones

### LINKED ISSUE #4844: Duplicating pages without files does not remove files in panel
(I've already written a bit about this in a different bug #4831, posting here for visibility)
## Description
While duplicating a page, setting 'copy files?' to false used to generate a new page with only text content and no images. Now, it still copies the file-UUIDs from the original content file, and keeps the connection to the images from the original page.
This can be confusing for the user, because it is not obvious where those images are coming from.
I think that's bad UX: 
Some clients might keep using these images, then at some point delete the original page and will be suprised that all the images are gone.

**Expected behavior**  
Duplicating a page without files should remove links to file UUIDs not belonging to the new page.

I realise this is more complicated now with UUIDs.
With Kirby <= 3.7, the content.txt was simply copied, and no images, pointing to an image that didn't exist anymore in that folder:
```
Cover: - monster-trees-in-the-fog.jpg
```
The panel handled that invalid link nicely by just not showing any image.
Now with UUIDs, the panel still can find that image from the original page and displays it.

Just an idea: Since the file links have a very distinctive pattern, wouldn't it be possible to use regex
(maybe `file:\/\/[\w]{16}`) to invalidate file links during duplication?

```
Cover: - file://soTV5M2PG5DJfuHw
```
becomes
```
Cover: - file://
```

Or is that replacement operation too risky? 

I would just be nice to be able to use file UUIDs while still having the old, more natural duplication behaviour:
When duplication without images -> no images on new, duplicated page.

## To reproduce
- Create a page with images
- Duplicate that page, "copy files?" -> no
- The panel still shows linked images

## Your setup
**Kirby Version**  
3.8.1.1

### LINKED ISSUE #4867: Duplicating pages does regenerate file UUIDs, but does not update content file with new UUIDs
## Description
Related to #4831. 
Duplicating pages with files now regenerates file UUIDs, but the files linked in the content.txt still point towards the original page (the duplication source).

**Expected behavior**  
Duplicated files should be linked correctly in the content.txt.

## To reproduce
1. Create a page with images, with UUIDs active
2. Duplicate that page, with files
3. File UUIDs get regenerated, but linked are still the parent files, not the newly generated ones.

## Your setup
**Kirby Version**  
3.8.2

**Additional context**  
See also dicussion here #4844

distantnative:
### Summary of changes
- New `PageCopy` class to handle all adaptations for a copied page, its files and children
- Tracking a map of all `oldUUID => newUUID` relations when replacing a UUID on newly generated models
- Replacing all UUID mappings in all new content files
- Removing file UUIDs (replacing with empty string) if files aren't copied
- Removing children UUIDs (replacing with empty string) if children aren't copied
- New `ContenStorageHandler::replaceStrings()`

### Reasoning
- Removing UUIDs for files/children when not copying files/children: Users obligation to fix empty.invalid KirbyTags or other afterwards when not copying files. But in the end it did feel better in the UX to handle it this way, even though I initially thought differently.

### Additional context
- We will have to refactor the content storage changes directly for v5. But I think this way is the right way to implement it for v4.

### Enhancements
- Duplicating pages: All UUIDs within the copied page, its files and children that point to an object that got copied as well will be replaced with the UUID of the copied version. If no copying files, any UUID pointing to a file within the copied page gets removed.

## Docs

- See where to mention this behavior.

## Ready?

### For review team

### DIFF
--- a/src/Cms/PageCopy.php
+++ b/src/Cms/PageCopy.php
@@ -2,8 +2,6 @@
 
 namespace Kirby\Cms;
 
-use Closure;
-use Kirby\Uuid\Identifiable;
 use Kirby\Uuid\Uuid;
 use Kirby\Uuid\Uuids;
 
@@ -20,7 +18,7 @@ use Kirby\Uuid\Uuids;
  * @since     5.0.0
  * @internal
  */
-class PageCopy extends Blueprint
+class PageCopy
 {
 	public function __construct(
 		public Page $copy,
@@ -32,22 +30,149 @@ class PageCopy extends Blueprint
 	}
 
 	/**
-	 * Applies various adaptions to the copied pages, files and children
+	 * Returns the pages collection for children to adapt
 	 */
-	public function adapt(): void
+	public function children(): Pages
 	{
-		foreach ($this->languages() as $language) {
-			$this->adaptSlug($language);
-			$this->adaptUuids($language);
+		return match($this->children) {
+			true  => $this->copy->index(drafts: true),
+			false => new Pages()
+		};
+	}
+
+	/**
+	 * Converts UUIDs for copied pages,
+	 * replacing the old UUID with a newly generated one
+	 * for all newly generated pages and files
+	 */
+	public function convertUuids(Language|null $language): void {
+		if (Uuids::enabled() === false) {
+			return;
+		}
+
+		if ($language instanceof Language && $language->isDefault() === false) {
+			return;
+		}
+
+		// store old UUID
+		$old = $this->copy->uuid()->toString();
+
+		// re-generate UUID for the page
+		$this->copy = $this->copy->save(
+			['uuid' => Uuid::generate()],
+			$language?->code()
+		);
+
+		// track UUID change
+		$this->uuids[$old] = $this->copy->uuid()->toString();
+
+
+		$this->convertFileUuids($language);
+		$this->convertChildrenUuids($language);
+	}
+
+	/**
+	 * Re-generate UUIDs for each child recursively
+	 * and merge with the tracked changed UUIDs
+	 */
+	protected function convertChildrenUuids(Language|null $language): void
+	{
+		foreach ($this->children() as $child) {
+			// always adapt files of subpages as they are
+			// currently always copied; but don't adapt
+			// children because we already operate on the index
+			$child = new PageCopy($child, files: true, uuids: $this->uuids);
+			$child->convertUuids($language);
+			$this->uuids = [...$this->uuids, ...$child->uuids];
 		}
 	}
 
 	/**
-	 * Adapts slug for copied pages
+	 * Re-generate UUID for each file and track the change
 	 */
-	public function adaptSlug(
-		Language|null $language
-	): void {
+	protected function convertFileUuids(Language|null $language): void
+	{
+		// if files have been copied,
+		// re-generate UUIDs and track changes
+		foreach ($this->files() as $file) {
+			// store old file UUID
+			$old = $file->uuid()->toString();
+
+			// re-generate UUID for the file
+			$file = $file->save(
+				['uuid' => Uuid::generate()],
+				$language?->code()
+			);
+
+			// track UUID change
+			$this->uuids[$old] = $file->uuid()->toString();
+		}
+
+		// if files have not been copied over,
+		// track file UUIDs from original page to
+		// remove/replace with empty string
+		if ($this->files === false) {
+			foreach ($this->original?->files() ?? [] as $file) {
+				$this->uuids[$file->uuid()->toString()] = '';
+			}
+		}
+	}
+
+	/**
+	 * Returns the files collection for files to adapt
+	 */
+	public function files(): Files
+	{
+		return match($this->files) {
+			true  => $this->copy->files(),
+			false => new Files()
+		};
+	}
+
+	/**
+	 * Returns all languages to adapt
+	 *
+	 * @todo Refactor once singe-lang mode also works with a language object
+	 */
+	public function languages(): Languages|iterable
+	{
+		$kirby = App::instance();
+
+		return match ($kirby->multilang()) {
+			true  => $kirby->languages(),
+			false => [null]
+		};
+	}
+
+	/**
+	 * Processes the copy with all necessary adaptations.
+	 * Main method to use if not familiar with individual steps.
+	 */
+	public static function process(
+		Page $copy,
+		Page $original,
+		bool $files = false,
+		bool $children = false
+	): Page {
+		$converter = new static($copy, $original, $files, $children);
+
+		// loop through all languages to remove slug from non-default
+		// languages and re-generate UUIDs (and track changes)
+		foreach ($converter->languages() as $language) {
+			$converter->removeSlug($language);
+			$converter->convertUuids($language);
+		}
+
+		// apply all tracked UUID changes at once
+		$converter->replaceUuids();
+
+		return $converter->copy;
+	}
+
+	/**
+	 * Removes translated slug for copied page
+	 */
+	public function removeSlug(Language|null $language): void {
 		// single lang setup
 		if ($language === null) {
 			return;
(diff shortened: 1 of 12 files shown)


src/Cms/PageCopy.php
```diff
@@ -0,0 +1,229 @@
...
+
+	/**
+	 * Returns the pages collection for children to adapt
+	 */
+	public function children(): Pages
+	{
+		return match ($this->children) {
+			true  => $this->copy->index(drafts: true),
+			false => new Pages()
+		};
```
bastianallgeier:
```suggestion
		if ($this->children === true) {
			return $this->copy->index(drafts: true);
		}

		return new Pages();
```

I can't help but feel uneasy with using match everywhere to be honest. It's one of those decisions that we would need to move along a lot more in so many other parts to turn it into a consistent coding style and I'm really not sure if it is helping. I don't want to block PRs because of this. I really just think that we should decide at which point it makes sense to use match and when there are just two cases like here, I would vote against it.

src/Cms/PageCopy.php
```diff
@@ -0,0 +1,229 @@
...
+			// children because we already operate on the index
+			$child = new PageCopy($child, files: true, uuids: $this->uuids);
+			$child->convertUuids($language);
+			$this->uuids = [...$this->uuids, ...$child->uuids];
+		}
+
+		// if children have not been copied over,
+		// track all children UUIDs from original page to
+		// remove/replace with empty string
+		if ($this->children === false) {
```
bastianallgeier: I stumbled across the property name here while reading, because I thought it was somehow related to the children method. This is of course only happening because we use the children property in the Page method as a cache for the children method. But maybe it would help to rename the children and file properties to something like includeChildren and includeFiles or copyChildren and copyFiles. It's just a thought though after stumbling across it. It's not super important.

src/Cms/PageCopy.php
```diff
@@ -0,0 +1,229 @@
...
+
+	/**
+	 * Returns the files collection for files to adapt
+	 */
+	public function files(): Files
+	{
+		return match ($this->files) {
+			true  => $this->copy->files(),
+			false => new Files()
+		};
```
bastianallgeier: See match comment above.

src/Cms/PageCopy.php
```diff
@@ -0,0 +1,229 @@
...
+	 *
+	 * @todo Refactor once singe-lang mode also works with a language object
+	 */
+	public function languages(): Languages|iterable
+	{
+		$kirby = App::instance();
+
+		return match ($kirby->multilang()) {
+			true  => $kirby->languages(),
+			false => [null]
```
bastianallgeier: Same here (match)

src/Cms/PageCopy.php
```diff
@@ -0,0 +1,229 @@
...
+		}
+
+		// apply all tracked UUID changes at once
+		$converter->replaceUuids();
+
+		return $converter->copy;
+	}
+
+	/**
+	 * Removes translated slug for copied page
```
bastianallgeier: I forgot again why we need this and had to look through the code first to understand it again. I think it would be helpful to add something like …
```suggestion
	 * Removes translated slug for copied page. This is needed to avoid translated slug collisions with the original page.
```

src/Content/ContentStorage.php
```diff
@@ -217,6 +217,18 @@ public function read(
...
+	 *
+	 * @param array $map search-replace mapping where all keys are replaced by their values
+	 */
+	public function replace(array $map): void
+	{
+		foreach ($this->all() as $version => $lang) {
+			$this->handler->replace($version, $lang, $map);
+		}
+	}
+
```
bastianallgeier: The ContentStorage class no longer exists in the alpha. Is there any chance that we can implement this without relying on the wrapper class? I guess we will otherwise create some nasty merge conflicts later.

src/Content/ContentStorageHandler.php
```diff
@@ -75,6 +75,20 @@ public function move(
...
 
+	/**
+	 * Searches and replaces one or multiple strings in all versions
+	 *
+	 * @param string $lang Code `'default'` in a single-lang installation
+	 * @param array $map search-replace mapping where all keys are replaced by their values
+	 *
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
+	 */
+	public function replace(
```
bastianallgeier: In the context of the content storage handler classes, I would think that "replace" means that the content is replaced/overwritten. We use Version::replace like that. Maybe we could use something like replaceStrings?

src/Content/PlainTextContentStorageHandler.php
```diff
@@ -128,6 +128,24 @@ public function read(string $version, string $lang): array
...
 
+	/**
+	 * Searches and replaces one or multiple strings in all versions
+	 *
+	 * @param string $lang Code `'default'` in a single-lang installation
+	 * @param array $map search-replace mapping where all keys are replaced by their values
+	 *
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
+	 */
+	public function replace(
```
bastianallgeier: See comment above

src/Filesystem/F.php
```diff
@@ -727,6 +727,25 @@ public static function remove(string $file): bool
 		return static::unlink($file);
 	}
 
+	/**
+	 * Searches and replaces one or multiple strings in a file
+	 */
+	public static function replace(
```
bastianallgeier: I would probably also vote for replaceStrings here.

src/Content/ContentStorage.php
distantnative: Maybe just adding a comment? Cause I am not sure if trying to work around it wouldn't make quite the mess for 4.5 – or we directly target v5 with this from the start?

bastianallgeier: Seeing all the work ahead of us for the changes, I would probably prefer to put it in v5 and not cause additional conflicts with it tbh.

src/Cms/PageCopy.php
distantnative: I went with `withChildren` and `withFiles`.

src/Content/MemoryContentStorageHandler.php
```diff
@@ -79,6 +80,28 @@ public function read(VersionId $versionId, Language $language): array
...
+		$fields = $this->read($versionId, $language);
+		$fields = A::map(
+			$fields,
+			fn ($field) => str_replace(
+				array_keys($map),
+				array_values($map),
+				$field
+			)
+		);
+		$this->update($versionId, $language, $fields);
```
bastianallgeier: Couldn't this part be in the ContentStorageHandler class as a default instead of declaring it as abstract method?

distantnative: Good idea!

src/Content/PlainTextContentStorageHandler.php
```diff
@@ -280,6 +280,20 @@ public function read(VersionId $versionId, Language $language): array
...
+	 * @throws \Kirby\Exception\NotFoundException If the version does not exist
+	 */
+	public function replaceStrings(
+		VersionId $versionId,
+		Language $language,
+		array $map
+	): void {
+		$file = $this->contentFile($versionId, $language);
+		F::replaceStrings($file, $map);
+	}
```
bastianallgeier: One last thought here. Do we really gain so much by using the new `F::replaceStrings` method? In theory, this could also mean that a field name is accidentally replaced, right? If we just keep the default behaviour from ContentStorageHandler::replaceStrings for text files, we would probably not hit huge performance issues, but at the same time we don't need the new F:: method and we also don't risk overwriting field names by accident.

bastianallgeier (approved):
Super nice!

silllli:
Just copied a page and noticed that all referenced files, although they were copied as well, still pointed to the originals … great that a fix is already in place, thank you! @distantnative

PR #6569, opened 2024-07-24
Replace bubbles components with tags

distantnative:
### Reasoning
Time to fade out those bubble components. We don't need two type of components that basically do the same.

### Enhancements
- `<k-tag>`: new `element` and `theme` props
- `<k-tags>`: new `element`, `element-tag` and `theme` props
- New `<k-tags-field-preview>` component

### Deprecated
- `<k-bubble>`, `<k-bubbles>` and `<k-bubbles-field-preview>`. Use `<k-tag>`, `<k-tags>` and `<k-tag-field-preview>` instead.

## Ready?

### For review team

### DIFF
--- a/panel/lab/components/fieldpreviews/tags/index.php
+++ b/panel/lab/components/fieldpreviews/tags/index.php
@@ -1,0 +1,5 @@
+<?php
+
+return [
+	'docs' => 'k-tags-field-preview',
+];
--- a/panel/lab/components/fieldpreviews/tags/index.vue
+++ b/panel/lab/components/fieldpreviews/tags/index.vue
@@ -1,0 +1,45 @@
+<template>
+	<k-lab-examples>
+		<k-lab-example label="Default">
+			<k-lab-table-cell>
+				<!-- @code -->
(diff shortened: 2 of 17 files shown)


bastianallgeier:
There are two issues I could spot: 

- The disabled state in light mode looks too dark and kind of more active than the regular one
- The tag field preview and all its children is still sortable 

In general, I wonder if we make the tags too complex and just blow the component up by using it in more places. But I also feel that the distinction between tag and bubble was never really strong enough for two separate components.

distantnative:
@bastianallgeier 
- Disabled dragging.
- Added a new suggestion for the light disabled state.
- Added a lab example for tags field preview.
- I understand the concern, I just think the current bubbles-tag duality is worse, especially since we use them so inconsistently. I was thinking if we could refactor it further: have `k-bubble` as dumb component and `k-tag` extends it, adding tag-specific stuff. But looking at `k-tag` currently, I don't think there is much to split up. What do you think?

bastianallgeier (approved):

PR #6572, opened 2024-07-25
Support role permissions for  `preview` option

### LINKED ISSUE #6400: Adding role-specific `preview` permission to blueprint results in Panel error
## Description
By adding this to my page blueprint

```yaml
options:
  preview:
    ncuser: true
```
I get the following error in the panel:
```
Kirby\Cms\PageBlueprint::preview(): Return value must be of type string|bool, array returned
```
Any other permissions work fine, e.g. if I add
```
options:
  update:
    ncuser: true
```
everything works fine. It's just the `preview` option that triggers this error.

**Expected behavior**  

Panel should load just fine.

**Screenshots**  

## To reproduce
1. Download plainkit
2. Create any user blueprint (e.g. `users/ncuser.yml`)
3. Add the code above to any page blueprint
4. See error in panel

## Your setup
**Kirby Version**  
4.1.2

distantnative:
### Summary of changes
- Blueprint classes for `::preview()` are now using the permissions to actually resolve what to return
- This allows not just to set true/false or the string, but an array with role-based permissions

### Fixes
- `preview` blueprint option now supports setting role-based permissions
#6400

### Breaking changes
- `$file->previewUrl()` returns `null` if the user has no permission for the file's parent (page/site) preview

## Ready?

### For review team

### DIFF
--- a/tests/Cms/Files/FileTest.php
+++ b/tests/Cms/Files/FileTest.php
@@ -725,6 +725,25 @@ class FileTest extends TestCase
 
 	public function testPreviewUrl()
 	{
+		$app = $this->app->clone([
+			'users' => [
+				[
+					'id'    => 'test',
+					'email' => 'test@getkirby.com',
+					'role'  => 'editor'
+				]
+			],
+			'roles' => [
+				[
+					'id'    => 'editor',
+					'name'  => 'editor',
+				]
+			]
+		]);
+
+		// authenticate
+		$app->impersonate('test@getkirby.com');
+
 		$page = new Page([
 			'slug'  => 'test',
 			'files' => [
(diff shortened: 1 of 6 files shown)


tests/Cms/Files/FileTest.php
```diff
@@ -738,8 +757,42 @@ public function testPreviewUrl()
...
+			'slug'  => 'test',
+			'files' => [
+				[
+					'filename' => 'test.pdf'
+				]
+			]
+		]);
+
+		$file = $page->file('test.pdf');
+		$this->assertSame('/media/pages/test/' . $file->mediaHash() . '/test.pdf', $file->previewUrl());
```
bastianallgeier: Shouldn't this return null in case of an unauthenticated user?

distantnative: @bastianallgeier I don't think so. If you see https://github.com/getkirby/kirby/blob/main/src/Cms/File.php#L636-L641 - that's pretty much the case covered here.

distantnative: Or do you think this current state  (also before PR) is actually a bug?

bastianallgeier: I saw this here for the Page and I think it actually makes sense that the preview URL is null when the user is not authenticated. https://github.com/getkirby/kirby/blob/b4c0dfdd95fcc8556963d9fc2dbcb3cbd72a2b98/tests/Cms/Pages/PageTest.php#L607

distantnative: Changed it, but now it's a breaking change, so we would have to move this to v5?

bastianallgeier: I would argue that this has been a pretty clear bug before. I'd say we could still put it in 4.4 but if you feel better to move it to 5, I'm fine with that as well.

distantnative: Ok, then ready for review/approval.

bastianallgeier (approved):

PR #6574, opened 2024-07-25
Previews 1 - `filePreviews` extension

distantnative:
### Summary of changes
- Adding a new `filePreviews` extension and core handlers

### Rationale
- We add all file preview handlers as class string to the `filePreviews` extension
- To identify which handler to use, we can loop through those and call on each a static method (`::accepts()`) to see if a handler should be used for this file
- We're adding custom file preview handlers before the core ones, so that when we loop through the handlers more specific custom handlers (e.g. for a specific extension) get checked before our broader default handlers.

### Additional context
Really not exciting yet but it's helpful to have this first in place before the other steps start. Tests will adapted along the way to be more meaningful

## Changelog

_Will include this in the final changelog in a subsequent PR._

## Ready?

### DIFF
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -57,6 +57,7 @@ trait AppPlugins
 		'collectionMethods' => [],
 		'fieldMethods' => [],
 		'fileMethods' => [],
+		'filePreviews' => [],
 		'fileTypes' => [],
 		'filesMethods' => [],
 		'fields' => [],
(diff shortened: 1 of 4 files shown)


bastianallgeier:
Do we need some kind of check at this point that the file preview classes extend a certain base class?

distantnative:
@bastianallgeier True, probably good to enforce them being any `instanceof \Kirby\Panel\Ui\FilePreview`. Should we add this in https://github.com/getkirby/kirby/pull/6575/files as only that PR adds this class?

distantnative:
@bastianallgeier added it on the other PR, I think the `::factory()` method is the best place to check for this (instead on the extension registration). https://github.com/getkirby/kirby/pull/6575/commits/4ad17167c9c2dd8cd35a0f5afd6598c52ecbdc41

bastianallgeier (approved):

PR #6575, opened 2024-07-25
Previews 2 -  UI classes

distantnative:
## Description

- [x] Merge https://github.com/getkirby/kirby/pull/6574

### Summary of changes
- Abstract `FilePreview` UI class
  - Serves as basic for any file preview component that renders as our other UI components
  - `::factory()` method I used to match a file to the correct file preview class
- `FileDefaultPreview` class as default/fallback handler if no more specific can be matched

### Additional context
- Currently not yet connected to any code that actually impacts the file preview
- Code coverage complains about `FileDefaultPreview::image()` not tested but the props test does cover it. I have no clue.

## Ready?


src/Panel/Ui/FilePreviews/FileDefaultPreview.php
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * Icon or image to display as thumbnail
+	 */
+	public function thumb(): array|null
+	{
+		return $this->file->panel()->image([
+			'back'  => 'transparent',
+			'ratio' => '1/1'
+		], 'cards');
+	}
```
distantnative, line 40: `Question:` What do you think about the name? So far, we have called it `image`. But I was wondering if we can find a way to differentiate between the image shown on items in overviews and this preview image shown on the file view. Or maybe there is no difference and both should be called `image` here.

src/Panel/Ui/FilePreview.php
```diff
@@ -0,0 +1,88 @@
...
+				'title' => I18n::translate('url'),
+				'text'  => $this->file->id(),
+				'link'  => $this->file->previewUrl()
+			],
+			[
+				'title' => I18n::translate('size'),
+				'text'  => $this->file->niceSize()
+			],
+		];
+	}
```
distantnative: `Comment`: I think these details are useful for any file preview made, so I put them in the abstract class.

src/Panel/Ui/FilePreview.php
```diff
@@ -0,0 +1,88 @@
...
+ * @author    Nico Hoffmann <nico@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @since     5.0.0
+ * @internal
+ */
+abstract class FilePreview extends Component
+{
+	public string $component = 'k-file-default-preview';
```
distantnative: `Comment:` For these, it's actually a lot easier to just define it on each class on top than handing it through every constructor.

src/Panel/Ui/FilePreviews/FileDefaultPreview.php
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * Icon or image to display as thumbnail
+	 */
+	public function image(): array|null
+	{
+		return $this->file->panel()->image([
+			'back'  => 'transparent',
+			'ratio' => '1/1'
+		], 'cards');
+	}
```
distantnative: `Question`: or should this be in the abstract class for all previews?

src/Panel/Ui/FilePreview.php
bastianallgeier: Would that be something thaat we should also apply to the Buttons?

src/Panel/Ui/FilePreviews/FileDefaultPreview.php
bastianallgeier: It looks to me like it could also be in the abstract class.

bastianallgeier:
Looking at it more closely, I think I would turn FilePreview into a non-abstract class and just use it as the default. It looks to me like having an additional FileDefaultPreview class would not really be necessary.

src/Panel/Ui/FilePreview.php
```diff
@@ -0,0 +1,109 @@
...
+
+	/**
+	 * Returns true if this class should
+	 * handle the preview of this file
+	 *
+	 * @codeCoverageIgnore
+	 */
+	public static function accepts(File $file): bool
+	{
+		throw new LogicException('Filew preview classes must define the static ::accepts() method');
```
bastianallgeier: I think this went wrong with the merge …
```suggestion
		return true;
```

distantnative: Mmm need to have a look why. I wanted to avoid your solution as any customer file preview not adding their own accepts method will claim any file. And they shouldn't. So i would really like to enforce them writing the accepts method (which is why I went abstract class in the first place).

bastianallgeier: Ah, now I get it. That explains the abstract class quite well. I'm not too sure though that we can enforce it like that. I think the risk still remains that people would just extend some other existing FilePreview class or the default class instead and the same problem would happen there as well. It's probably more about documenting the extension process correctly.

distantnative: I agree on having the one class. What I didn't understand why you'd prefer to not throw an exception when no custom accepts method has been defined.

bastianallgeier: Maybe I'm misunderstanding the concept, but wouldn't our default file preview class also need to define which types of files it can handle with the accept method?

src/Panel/Ui/FilePreview.php
```diff
@@ -0,0 +1,109 @@
...
+			if (is_subclass_of($handler, static::class) === false) {
+				throw new InvalidArgumentException('File preview handler "' . $handler . '" must extend ' . static::class);
+			}
+
+			if ($handler::accepts($file) === true) {
+				return new $handler($file);
+			}
+		}
+
+		return new static($file);
```
distantnative: `Todo:` should be `self`

src/Panel/Ui/FilePreview.php
distantnative: @bastianallgeier My idea was different:
- `FilePreview::factory()` loops through all file preview extensions and calls their `::accepts()` method
- The first returning true gets chosen, a new object created and returned
- `FilePreview` itself isn't in that list of file preview extensions.
- Instead if the loop runs through without finding any match, we create an object with the base class and return it.

https://github.com/getkirby/kirby/pull/6575/files#diff-bfc2ed602ed817e9a56aa6e78e991c493524a5948405519d87d8d9c20fc224b7R77-R87

This is why the base class would not need to define its own accepts, as it is used always as fallback if no other accepts matches. But I think it should ideally force all child classes to define their own accepts. And not to inherit an accepts that matches on every file (which could be easily overlooked).

distantnative:
Contrary to what the coverage tests say, the `::image()` method has unit tests.

src/Panel/Ui/FilePreview.php
bastianallgeier: I feel really bad about it, but after understanding it better, I think your initial setup with the abstract indeed made more sense 🙈 sorry for letting you jump through those hoops.

distantnative: Changed back 😅

bastianallgeier:
I really like it a lot. I hope you don't kill me, but we maybe should have one last conversation about the naming convention. 

Do we go with 

- ``File${type}Preview`` or 
- ``${type}FilePreview``?

With our components, we mostly follow something like `RemoveDialog`, `TextField`, `FormDrawer` etc.

bastianallgeier (approved):
Love it! I think the naming is a lot better this way.

PR #6577, opened 2024-07-25
Previews 4 - Connect FilePreview classes to core

distantnative:
## Description

- [x] Merge https://github.com/getkirby/kirby/pull/6574
- [x] Merge https://github.com/getkirby/kirby/pull/6575
- [x] Merge https://github.com/getkirby/kirby/pull/6576

### Summary of changes
- Start using the new file preview classes to provide the data for the frontend in `Panel\File::props()`

## Ready?

- [x] In-code documentation (wherever needed)
- [x] Unit tests for fixed bug/feature
- [ ] Tests and CI checks all pass

PR #6578, opened 2024-07-25
Previews 5 - Rework Vue components

distantnative:
## Description

- [x] Merge https://github.com/getkirby/kirby/pull/6574
- [x] Merge https://github.com/getkirby/kirby/pull/6575
- [x] Merge https://github.com/getkirby/kirby/pull/6576
- [x] Merge https://github.com/getkirby/kirby/pull/6577
- [x] Lab examples
  - [x] `k-file-preview-details`
  - [x] `k-file-preview-thumb`

### Summary of changes
- `k-file-preview` is now only a mere wrapper to receive `{component, props}` data and render it. It also applies the basic class and some basic styling connected to it.
- `k-default-file-preview` serves all files that don't have any specific preview (currently all but images). Icon in the thumb column and details listed as we're used to.
- `k-image-file-preview` serves images and is similar to the default preview but displays the actual image and adds the focus feature. It integrates directly the focus button, so the standalone component for that got removed.
- `k-file-preview-thumb` is a wrapper component which shows the slot content on a pattern background. It adds an optional dropdown menu.
- `k-file-preview-details` is a component that takes over rendering the details list. Added it as standalone as I suspect it will be used in many, but maybe not all previews. And maybe also in different ways.

### Reasoning
I expect this one to be the most discussed PR in this series as it really redefines how we restructure the Vue components around file previews, what components to have and what we expect previews to generally look like (what's shared code/styling, what's custom). Happy to spend some time and thought together on this one to be sure we're happy with it.

### Additional context
This is the PR that finishes the feature. I will have one more PR to suggest an audio and a video preview. But those are just extra. The feature is complete with this PR.

### Features
- Custom file previews

### Breaking changes
- `<k-file-preview>` got fully refactored. If you were replacing or extending it, your code likely will break. Check out the new custom file preview feature if you want to provide previews for specific files.

## Docs

Custom file previews consist of two parts: 
- a PHP class defining what data is passed to the preview and what files this preview is used for
- a Vue component to render the preview

### Custom FieldPreview class

For a custom file preview, you need a new class that extends `Kirby\Panel\Ui\FieldPreview`. In this example, we're building a custom preview for `.glb` files that are used to display 3D data:

```php
class GlbFilePreview extends FilePreview
{
	public function __construct(
		public File $file,
		public string $component = 'k-glb-file-preview'
	) {
	}

	public static function accepts(File $file): bool
	{
		return $file->extension() === 'glb';
	}
}
```

This is the minimal setup for a custom class. What it does:
- It defines which Vue component to use, in this case `k-glb-file-preview` which we will create later. If no component is specified, `k-default-file-preview` will be used.
- Something you must define is the `::accepts($file)` method. It is used to determine which preview class is used for a certain file. Kirby will loop through all available preview classes and use the first one that accepts the file.

### Register your preview class

To use your custom file preview class, you need to register it as a plugin extension:

```php
	'filePreviews' => [
		GlbFilePreview::class
	]
]);
```

### Vue component

Finally, we need to create the Vue component that is used to render the preview. As above listed, we have named it `k-glb-file-preview`.

By default, file preview components will receive three default props:
- `url` to the file
- `details` array with basic file information (ideally used for `<k-file-preview-details>`)
- `image` object with the thumb icon/image also used in sections etc. (we don't use it for this preview)

We also use the `<k-file-preview-thumb>` component as wrapper for our actual 3D model, which adds a pattern background.

```js
panel.plugin("getkirby/custom-file-preview", {
	components: {
		"k-glb-file-preview": {
			template: `
				<figure class="k-default-file-preview k-glb-file-preview">
					<k-file-preview-thumb>
						<model-viewer :src="url" />
					</k-file-preview-thumb>

					<figcaption>
						<k-file-preview-details :details="details" />
					</figcaption>
				</figure>
			`,
			props: {
				details: Array,
				url: String,
			},
		},
	}
});
```

The `<model-viewer>` element is [from a library](https://modelviewer.dev) that helps to actually display the GLB file. I won't go into the details how to use it exactly (loading external script etc.). You can try the full custom view: https://github.com/getkirby/sandbox/pull/7

https://github.com/user-attachments/assets/3ef71d15-2d70-4a4d-a1d9-1e16cb4b13b0

## Ready?

### For review team

  - https://github.com/getkirby/sandbox/pull/7 


panel/lab/components/filepreviews/2_image/index.vue
```diff
@@ -1,61 +1,52 @@
...
 		<k-lab-example label="Landscape">
-			<k-file-preview
-				:focus="focus"
-				:focusable="true"
+			<k-image-file-preview
 				:details="details"
 				:image="image"
 				url="https://getkirby.com"
-				@focus="focus = $event"
+				class="k-file-preview"
```
bastianallgeier, line 8: Wouldn't this class be something that the file preview components should already have applied?

bastianallgeier: Ah, sorry. I saw further down in the code that it is applied if you use the k-file-preview wrapper

panel/src/components/Views/Files/ImageFilePreview.vue
```diff
@@ -0,0 +1,147 @@
...
+			const focus = this.$panel.content.values["focus"];
+
+			if (!focus) {
+				return;
+			}
+
+			const [x, y] = focus.replaceAll("%", "").split(" ");
+
+			return { x: parseFloat(x), y: parseFloat(y) };
+		},
```
bastianallgeier, line 72: Could focus be a prop instead, which is provided by the preview class?

bastianallgeier (approved):
It's great to see how all of this is coming together. I think it's already super clean. I was wondering about passing focus as a prop instead of computing it, but that's a decision we can always delay and merge this first.

panel/lab/components/filepreviews/2_image/index.vue
distantnative: Yes, I was myself torn gif this is the best way, but couldn't make up my mind.

panel/src/components/Views/Files/ImageFilePreview.vue
distantnative: I don't think yet, only once we have changes on the backend (as this also needs to reflect unsaved changes). And I moved it inside the image file preview to not apply it in JS to every file preview otherwise.

distantnative:
@bastianallgeier great to hear you have such a positive view of it. My own biggest concerns for criticism were the division of CSS rules between the components: which component adds what as baseline, how much/little do custom components get by default etc. – does this work for you as is?

distantnative:
Merging it, we can still tweak it during the alpha if necessary

PR #6583, opened 2024-07-29
`rawvalues` option for pages section tables

### LINKED ISSUE #6404: Page tables load very slow in 4.2
## Description
After updating to 4.2, some page sections with `layout: table` load extremely slow. 100 pages ~10 seconds, 30 pages ~6 seconds. The issue goes away when changing layout to list, cardlets or cards. Confirmed by other users too.

In the forum it was suggested by @rasteiner  that the toggle field preview fix in this commit https://github.com/getkirby/kirby/commit/75eb89cc27d35c0cd0da06680077462119bd2e88 can be the cause:
"It seems to introduce a new instantiation (new Form($model)) which is run for every row in the table." 

## Your setup
**Kirby Version**  
Kirby 4.2

**Additional context**  
More details in the forum post https://forum.getkirby.com/t/page-sections-load-slow-kirby-4-2-issue/31373/3

distantnative:
### Summary of changes
- `rawvalue` option to restore pre-v4.2 more performant behavior

### Reasoning
Full explanation: https://github.com/getkirby/kirby/issues/6404#issuecomment-2249886213

### Enhancement
- Pages section: if you are experiencing performance issues with a pages section and the table layout, you can try to use `rawvalues: true` to improve performance. This will use the unprocessed values from the content file, which is a lot more performant, but can result in some broken column previews (that would require process values).

## Ready?

### For review team

### DIFF
--- a/config/sections/mixins/layout.php
+++ b/config/sections/mixins/layout.php
@@ -143,6 +143,7 @@ return [
 
 			// if forcing raw values, get those directly from content file
 			// TODO: remove once Form classes have been refactored
+			// @codeCoverageIgnoreStart
 			if ($this->rawvalues === true) {
 				foreach ($this->columns as $columnName => $column) {
 					$item[$columnName] = match (empty($column['value'])) {
@@ -156,6 +157,7 @@ return [
 
 				return $item;
 			}
+			// @codeCoverageIgnoreEnd
 
 			// Use form to get the proper values for the columns
 			$form = Form::for($model)->values();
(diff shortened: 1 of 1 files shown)


config/sections/mixins/layout.php
```diff
@@ -133,8 +141,10 @@
 				$item['info'] = $model->toString($this->info);
 			}
 
-			// Use form to get the proper values for the columns
-			$form = Form::for($model)->values();
+			if ($this->rawvalues !== true) {
```
bastianallgeier: I was wondering if we could turn this into two separate code blocks for the two variants for better clarity. The combination of the conditional $form variable with the match clause in the foreach loop is a bit hard to parse.

bastianallgeier (approved):

PR #6589, opened 2024-07-30
`files.sort` permission

### LINKED ISSUE #1969: `$file->changeSort()` doesn't have permission setting

distantnative:
### Enhancement
- New `files.sort` permission
#1969

### Breaking changes
- If file sorting was previously disabled via the `files.update` permission, the new `file.sort` permission has to be configured accordingly.

## Docs

- Add new `files.sort` permission

## Ready?

### For review team

### DIFF
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -110,7 +110,8 @@ return [
 			$dragTextAbsolute = $this->model->is($this->parent) === false;
 
 			foreach ($this->models as $file) {
-				$panel = $file->panel();
+				$panel       = $file->panel();
+				$permissions = $file->permissions();
 
 				$item = [
 					'dragText'  => $panel->dragText('auto', $dragTextAbsolute),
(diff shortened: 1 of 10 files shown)


bastianallgeier (approved):

afbora:
@distantnative @bastianallgeier I remembered https://github.com/getkirby/kirby/pull/2043#issuecomment-528304516 comment when I saw this PR. As far as I understand, this PR solves the performance issue or not valid for v5, right?

distantnative:
@afbora I didn't experience at least locally when testing. But another reason why this is good to have in v5 so we have plenty of test time to see if it plops up as issue.

afbora:
It is necessary to test with a large number of files to catch the issues. A similar situation happened with https://github.com/getkirby/kirby/issues/6404 issue. The issue was revealed by a user with a large number of pages.

PR #6590, opened 2024-07-30
Unlink tmp file for failed uploads

### LINKED ISSUE #2476: Doesn't clear `tmp` when uploading files
**Describe the bug**  
When you upload a file through PHP, it is usually temporarily stored in the `/tmp` folder until the script is finished. Since Kirby's upload method renames each file on upload, however, files pile up in the `/tmp` folder until manually removed. This can easily end up in people no longer being able to upload files due to space issues.

**To Reproduce**  
1. Upload a file via the Panel.
2. Open the `/tmp` folder.
3. See how the uploaded file ends up in the `/tmp` folder and is not removed when the upload is finished.

**Expected behavior**  
Files should either not be renamed on upload or removed again afterward.

**Kirby Version**  
Tested with 3.3.4

**Additional context**  
https://forum.getkirby.com/t/uploading-images-keep-a-copy-of-the-original-in-phps-tmp-folder/17469/2

distantnative:
### Summary of changes
File uploads that get rejected, e.g. from FileRules, remain in the tmp dir as the native PHP garbage collection doesn't pick them up. It doesn't, because we rename them in place for MIME detection. But afterwards we have to take care of removing them ourselves. As the discussion in https://github.com/getkirby/kirby/issues/2476 shows, solving this directly is rather complex and complicated. 

But I thought that we probably can cover most cases (e.g. as FileRules rejections), if we unlink the tmp file in the try-catch block of the `Upload::process` method.

### Additional context
It's targeted or v5 as I wanted to work with the new `Upload` class and not touch the old upload spaghetti code.

### Fixes
- Kirby is better at removing failed file uploads from the server tmp directory
https://github.com/getkirby/kirby/issues/2476

## Ready?

### For review team

### DIFF
--- a/src/Api/Upload.php
+++ b/src/Api/Upload.php
@@ -189,6 +189,9 @@ readonly class Upload
 				};
 			} catch (Exception $e) {
 				$errors[$upload['name']] = $e->getMessage();
+
+				// clean up file from system tmp directory
+				F::unlink($upload['tmp_name']);
 			}
 
 			if ($this->single === true) {
--- a/tests/Api/UploadTest.php
+++ b/tests/Api/UploadTest.php
@@ -176,7 +176,7 @@ class UploadTest extends TestCase
 				'files' => [
 					[
 						'name'     => 'test.txt',
-						'tmp_name' => KIRBY_TMP_DIR . '/api.upload/abc',
+						'tmp_name' => static::TMP . '/abc',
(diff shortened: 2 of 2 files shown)


bastianallgeier (approved):
I love how practical this quick fix is. I agree that it's likely not going to fix all cases, but it should definitely cover most of them.

PR #6593, opened 2024-07-31
Changes 21: Render Mode

bastianallgeier:
### Summary of changes

- New `VersionId::$render` property
- Use new render mode in `ModelWithContent::content` method
- Use `changes` Version in `Panel\Model::content()` if it exists, to show changes in the panel.

## Ready?

### For review team

### DIFF
--- a/src/Cms/ModelWithContent.php
+++ b/src/Cms/ModelWithContent.php
@@ -136,7 +136,7 @@ abstract class ModelWithContent implements Identifiable, Stringable
 		// @todo this entire block can be radically simplified as soon
 		// as the models use the versions exclusively.
 		if (VersionId::$render ?? null) {
-			$version = $this->version(VersionId::render($this));
+			$version = $this->version(VersionId::$render);
 
 			if ($version->exists($language) === true) {
 				return $version->content($language);
--- a/src/Cms/ModelWithContent.php
+++ b/src/Cms/ModelWithContent.php
@@ -129,13 +129,24 @@ abstract class ModelWithContent implements Identifiable, Stringable
 	 */
 	public function content(string|null $languageCode = null): Content
 	{
-		// single language support
-		if ($this->kirby()->multilang() === false) {
-			return $this->content ??= $this->version()->content('default');
-		}
-
 		// get the targeted language
 		$language = Language::ensure($languageCode);
+
+		// fetch a specific version in preview render mode
+		// @todo this entire block can be radically simplified as soon
+		// as the models use the versions exclusively.
+		if (VersionId::$render ?? null) {
+			$version = $this->version(VersionId::render($this));
+
+			if ($version->exists($language) === true) {
+				return $version->content($language);
+			}
+		}
+
+		// single language support
+		if ($this->kirby()->multilang() === false) {
+			return $this->content ??= $this->version()->content($language);
+		}
 
 		// only fetch from cache for the current language
 		if ($languageCode === null && $this->content instanceof Content) {
--- a/src/Panel/Model.php
+++ b/src/Panel/Model.php
@@ -43,8 +43,9 @@ abstract class Model
 		if ($version->exists() === true) {
 			$changes = $version->content()->toArray();
 		}
-// create a form which will collect the published values for the model,
-// but also pass along unpublished changes as overwrites
+
+		// create a form which will collect the published values for the model,
+		// but also pass along unpublished changes as overwrites
 		return Form::for(
 			model: $this->model,
 			props: [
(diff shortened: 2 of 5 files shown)


src/Cms/ModelWithContent.php
```diff
@@ -129,14 +129,25 @@ abstract protected function commit(
...
+
+		// fetch a specific version in preview render mode
+		// @todo this entire block can be radically simplified as soon
+		// as the models use the versions exclusively.
+		if (VersionId::$render ?? null) {
+			$version = $this->version(VersionId::render($this));
+
+			if ($version->exists($language) === true) {
+				return $version->content($language);
+			}
```
distantnative, line 143: `Understanding`: What happens if the render mode is set to changes, but the currently requested language doesn't exist? Shouldn't it still show changes, but in the default language? I think based on this code it would fall back to the published version instead.

bastianallgeier: Uff, it's creating another brain knot for me. Let me try to break it down. 

- The English version has changes
- I'm switching to German 
- German already has published translations for all fields 
- I want to make some additional changes there as well. 
- The changes version wouldn't exist at this point. 

But wouldn't it be completely misleading or even wrong to fall back to the changes in English here? That's at least how I think about it.

distantnative: Is it the case at the moment that there can be a changes version in English but non in German? I thought currently it's as soon as any language has changes, the full set of languages is created as changes version.

If we treat the completely independent, then the chain how you describe it makes more sense indeed. But then, wouldn't our hierarchy rather be translations have versions instead of versions have translations?

distantnative: I think it's actually fine how it is.

src/Panel/Page.php
```diff
@@ -373,11 +374,23 @@ public function view(): array
...
 			'breadcrumb' => $page->panel()->breadcrumb(),
 			'component'  => 'k-page-view',
 			'props'      => $this->props(),
 			'title'      => $page->title()->toString(),
 		];
+
+		// The render mode needs to be undone here to keep unit tests working
+		// That's the downside of such static properties. They are likely to
+		// cause side-effects
+		VersionId::$render = null;
```
distantnative: `Concern:` This feels like a hack to me I am not sure it's good to introduce, even temporarily. The data we pass to the Panel should not rely on the render mode, don't you think? Should we really establish this connection? To me, I'd think that the Panel data should be separated from the render mode, shouldn't it?

bastianallgeier: You are right. We could probably handle this in the `Panel\Model::content` method instead and check for a changes version and get the content directly from there. It feels cleaner indeed. I will change it.

bastianallgeier: I've moved the logic into `Model::content` This works really well and is definitely cleaner. It's still a bit weird in combination with the Form class, but I think that's fine for now.

src/Cms/ModelWithContent.php
```diff
@@ -129,14 +129,25 @@ abstract protected function commit(
...
 	public function content(string|null $languageCode = null): Content
 	{
+		// get the targeted language
+		$language = Language::ensure($languageCode);
+
+		// fetch a specific version in preview render mode
+		// @todo this entire block can be radically simplified as soon
+		// as the models use the versions exclusively.
+		if (VersionId::$render ?? null) {
+			$version = $this->version(VersionId::render($this));
```
distantnative: `Observation:` No idea if significant, a problem or just fine. But what is the benefit of calling `VersionId::render($this)` when we check in the line before that `VersionId::$render` is set and thus the method will 100% just return that. Couldn't we do then just

```php
if ($versionId = VersionId::$render ?? null) {
    $version = $this->version($versionId);
```

src/Panel/Model.php
```diff
@@ -37,7 +37,19 @@ abstract public function buttons(): array;
...
 	public function content(): array
 	{
-		return Form::for($this->model)->values();
+		$version = $this->model->version('changes');
+		$changes = [];
+
+		if ($version->exists() === true) {
+			$changes = $version->content()->toArray();
+		}
+
```
distantnative, line 46:
```suggestion
// create a form which will collect the published values for the model,
// but also pass along unpublished changes as overwrites
```

src/Cms/ModelWithContent.php
bastianallgeier: You are completely right. It's not needed here at all.

distantnative (approved):

PR #6596, opened 2024-08-02
Changes 22: Provide access to pages rendered with changes via URL

bastianallgeier:
## Description

> [!CAUTION]
> This PR leads to a big TODO to get the final token verification for preview URLs right later.

### Summary of changes

This is a very rough and insecure first implementation to get preview URLs for changes. After a huge App class rabbit hole, I decided to go for a very simple first "better done than perfect" approach to finally be able to move on with the panel integration. 

Add `?_version=changes` to the URL of any page to see changes if any are available

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -4,6 +4,7 @@ namespace Kirby\Cms;
 
 use Closure;
 use Generator;
+use Kirby\Content\VersionId;
 use Kirby\Data\Data;
 use Kirby\Email\Email as BaseEmail;
 use Kirby\Exception\ErrorPageException;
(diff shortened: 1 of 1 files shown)


src/Cms/Page.php
```diff
@@ -769,18 +769,32 @@ public function isUnlisted(): bool
...
+	 * Validates the given token
+	 */
+	public function isVerifiedByToken(string|null $token): bool
+	{
+		if ($token !== null && $this->token() === $token) {
+			return true;
 		}
 
-		return $this->token() === $token;
+		return false;
```
distantnative: `Question`: What's the purpose of having `::isVerified($token)` as well as `::isVerifiedByToken()`? If both are essential separately, can we find more distinct names that explain their individual purpose?

bastianallgeier: I'm not too happy with the method names myself yet. ::isVerified already existed and it's already not a really good name. It should probably be something like `::isViewable` We could then maybe use `::hasToken` as second method to just check if the token matches.

distantnative: How about `::isPreviewable($token)` and `::isValidPreviewToken()`?

bastianallgeier:
Ugh, this is leading to a new rabbit hole in the App::resolve method that we need to tackle first.

distantnative (approved):
Important TODO indeed, but if it helps to move forward for the moment.

PR #6601, opened 2024-08-11
Redirect language URLs with non-translated slugs

### LINKED ISSUE #3550: Redirect requests to default language URI in secondary languages to the custom secondary language slug
## Basic example

**URLs:**

```
/page/subpage
/de/seite/unterseite
```

**Request mapping right now:**

```
/page/subpage ✅
/de/seite/unterseite ✅

/de/page/subpage    -> serves the page /de/seite/unterseite
/de/seite/subpage   -> same here
/de/page/unterseite -> same here
```

**Expected behavior:**

The last three examples should be `HTTP 302` redirects to the canonical URL `/de/seite/unterseite`.

## Complex example

This works at the moment and there needs to be a unit test to verify that we don't break this behavior.

**URLs:**

```
/page1
/de/seite1

/page2
/de/page1 (translation of page2!)
```

**Expected routing:**

```
/page1 ✅
/de/seite1 ✅

/page2 ✅
/de/page2 -> 302 to /de/page1
/de/page1 ✅ (should *not* redirect to /de/seite1, which is a different page)
```

## Background context

See #2257.

distantnative:
### Summary of changes
- If a secondary-language page gets accessed with a URL that contains untranslated slugs (for which a translation exists), the request is redirected to the fully translated URL (302 redirect)

### Reasoning
To make those pages only available under the main (full- translated) URL and not several ones (with untranslated slug parts).

### Additional context
In https://github.com/getkirby/kirby/issues/3550#issue-947976809 under "Complex example", @lukasbestle asks for an additional unit test where the translated slug of a page is the same as the untranslated slug of another page. This unit test is not included here as it would fail. However, this behavior is also different before this PR: if a translated slug is the same as the untranslated slug of a different page, our router will return the different page. Moreover,  our page rules also prevent creating a translated slug that matches an existing untranslated slug. This is why I didn't include this aspect in this PR. To me this seems as a much larger change to be discussed, while this PR implements the solution for #3550 based on the existing rules.

### Fixes
- #3550

## Ready?

### For review team

### DIFF
--- a/src/Cms/LanguageRoutes.php
+++ b/src/Cms/LanguageRoutes.php
@@ -3,6 +3,7 @@
 namespace Kirby\Cms;
 
 use Kirby\Filesystem\F;
+use Kirby\Toolkit\Str;
 
 class LanguageRoutes
 {
@@ -29,8 +30,25 @@ class LanguageRoutes
 				'pattern' => $language->pattern(),
 				'method'  => 'ALL',
 				'env'     => 'site',
-				'action'  => function ($path = null) use ($language) {
+				'action'  => function ($path = null) use ($kirby, $language) {
 					$result = $language->router()->call($path);
+
+					// redirect secondary-language pages that have
+					// been accessed with non-translated slugs in their path
+					// to their fully translated URL
+					if ($path !== null && $result instanceof Page) {
+						if (Str::endsWith($result->url(), $path) === false) {
+							$url   = $result->url();
+							$query = $kirby->request()->query()->toString();
+
+							// preserve query across redirect
+							if (empty($query) === false) {
+								$url .= '?' . $query;
+							}
+
+							return Response::redirect($url);
+						}
+					}
 
 					// explicitly test for null as $result can
 					// contain falsy values that should still be returned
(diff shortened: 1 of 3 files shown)


src/Cms/LanguageRoutes.php
```diff
@@ -29,9 +30,26 @@ public static function create(App $kirby): array
...
+							$url   = $result->url();
+							$query = $kirby->request()->query()->toString();
+
+							// preserve query across redirect
+							if (empty($query) === false) {
+								$url .= '?' . $query;
+							}
+
+							return Response::redirect($url);
+						}
```
distantnative, line 50: `Question`: @bastianallgeier https://github.com/getkirby/kirby/pull/6601/files#diff-141beee66f7ef4aacf2cf10c723ab70676499ba12e2b2639fd82d9c4ce65e3ffL69-R114 tries to solve something similarly. Can you explain what? I don't think though we could move this new code to `::fallback` as `$result` actually is a Page object found. That's why I solved it here. (but maybe we could align the code style, e.g. `Response::redirect()` vs `$kirby->response()->redirect()`)

bastianallgeier: That part tries to redirect `/de/something` to an absolute language URL by detecting the code. I'm not entirely sure anymore why we needed this. I think it's code that supported legacy behavior. 

I agree that we can't implement both it in the same place. I think the `$kirby->response()->redirect()` solution is a bit nicer when we already have access to the kirby instance.

bastianallgeier (approved):

PR #6602, opened 2024-08-11
`(image: )` tag: `width: auto`

### LINKED ISSUE #5064: Missing width/height attribute for images when added via (image: ) / kirbytext
## Description
When Kirby renders a field containing the _kirbytext_ flavor of markdown into HTML, it does not add `width` and `height` attributes to the `` element. Besides their obvious functionality, these two fields inform the browser about an image‘s aspect ratio. Adding them will improve the experience for site visitors on bad/slow connections, and users of the `kirbytext()`-method can currently not add this themselves easily. I did not test the `markdown()`-method. If the lack of `width`/`height` attributes is considered a bug, the `markdown()`-method may need to be updated as well to be consistent.

I understand this might constitute a breaking change for some themes, as there will be the need to update some theme‘s CSS to include a `height`-attribute. Still, I think the improvement is worth considering. There is information about this in this older video from 2019: https://youtu.be/4-d_SoCHeWE

**Expected behavior**  
Any `(image: )` element contained in a Field should be rendered into HTML that includes the `width` and `height` attributes for that image. Considering images currently render a `figure`-element containing an `img`-element, I would expect this result:

```HTML
<figure>
  <img alt="…" src="…" width="…" height="…">
</figure>
```

**Alternative 1**
I considered suggesting this alternative: Add `style: "aspect-ratio: [width] / [height]"` to all `img`-elements. Example:

```HTML
<figure>
  <img alt="…" src="…" style="aspect-ratio: [width] / [height]">
</figure>
```

However, this would give the aspect-ratio the [highest specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#inline_styles) and likely break more than adding the `width`/`height` attributes.

**Alternative 2**
Do nothing. I considered if adding `width`/`height` attributes would do more harm than good. I do not know for sure. As adding them is a best-practice to avoid reflows after the initial render of the page, I think the fact that they are missing is a bug.

## To reproduce
1. Add an `(image: )` to a _kirbytext_-field in a page‘s TXT file

example.txt
```TXT
title: Example Page

----

text:

Nulla vitae elit libero, a pharetra augue. Nullam quis risus eget urna mollis ornare vel eu leo.

(image: image.png alt: An image)

Maecenas sed diam eget risus varius blandit sit amet non magna. Duis mollis, est non commodo luctus, nisi erat porttitor ligula, eget lacinia odio sem nec elit.

```

## Your setup
**Kirby Version**  
3.9.1

 - Device: MacBook Pro
 - OS: macOS 13.2.1 (22D68)
 - Browser: Safari
 - Version: Version 16.3 (18614.4.6.1.6)

distantnative:
### Summary of changes
- Image KirbyTag: support `auto` as value for `width` and `height` attributes, which will use the actual image dimensions for local files

### Reasoning
Based on the outline https://github.com/getkirby/kirby/issues/5064#issuecomment-1498060867
- I went for `auto` instead of `intrinsic` as it felt more common in the Kirby language to me
- I only implemented the optional, additive enhancement as there has been no consensus in the issue if we would want to migrate to make this the default behavior. I think it's a good enhancement as option nevertheless.

### Fixes
- `(image: )` KirbyTag: for local files, add `width: auto` and/or `height: auto` to include the actual image dimensions as attributes 
#5064

## Docs

```
(image: cat.jpg width: auto height: auto)
```

```php
// config.php
'kirbytext' => [
    'image' => [
        'width' => 'auto',
        'height' => 'auto',
    ]
]
```

## Ready?

### For review team


bastianallgeier (approved):
Nice!

PR #6606, opened 2024-08-11
Filename sanitization: consider user language

### LINKED ISSUE #4972: File upload: Take ASCII conversion rules of current language and/or user language into account
## Description

Kirby only uses the default conversion rules when converting filenames on upload. This breaks when files with names in another language are uploaded.

**Expected behavior**  

Kirby should take the current language into account. We need to determine whether this should be the current content language, the current user language or both.

## To reproduce

1. Name a file `日本語.pdf`
2. Upload the file to a files section
3. Error "You are not allowed to upload invisible files" because all characters are stripped by the default ASCII rules

## Additional context

https://github.com/getkirby/getkirby.com/issues/1817

distantnative:
### Summary of changes
- `Filesystem\Filename::sanitizeName()` considers the current user language for filename sanitization

### Reasoning
The assumption is that a file name might be in the same language as the user is using the Panel with. Avoids more cases where just the ASCII rules would leave an empty string (e.g. with filenames such as `안녕하세요.pdf`).

### Enhancement
- Filename sanitization considers user language for better results
#4972

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/Filename.php
+++ b/src/Filesystem/Filename.php
@@ -2,6 +2,8 @@
 
 namespace Kirby\Filesystem;
 
+use Kirby\Cms\App;
+use Kirby\Cms\Language;
 use Kirby\Toolkit\Str;
 
 /**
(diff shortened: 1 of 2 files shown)


bastianallgeier (commented):
If I'm honest, I prefer the separation into sanitization methods instead of caching the results in the getter methods. The cached properties often lead to problems or at least a bit more brain power when reading the code.

distantnative:
@bastianallgeier changed it back to keep the methods

bastianallgeier (approved):

PR #6608, opened 2024-08-11
fix: Translated error pages in multilang

### LINKED ISSUE #4834: Error page always rendered in default language for some languages
### Scenario to reproduce:

Have 2 languages:
- French (default), with `"url" => "/"`
- English (secondary), with `"url" => "en"`

If you visit `/en/asdf`, the url is right, but the content is not translated (appears in default language).
If you visite `/asdf`, the url is right and the content is translated.

Fix:
If I set the french `url => "fr"`, it works as intended. Both "/" and "/en" are displaying the right locale.
If I set the english as default, with `url => "/"`, it works as intended.

This means I cannot have French as default language with `url => "/"`.

Same issue as this user: https://forum.getkirby.com/t/error-page-is-shown-in-wrong-language/18084/9

_Originally posted by @juliencharette in https://github.com/getkirby/kirby/issues/2609#issuecomment-1306071201_

distantnative:
### Summary of changes
- If a language has no specific URL path, its pattern doesn't match anymore `(:all)`. Instead it only matches all that does not start with the URL path prefix of any other page.
- Because of that, a path without any matching page now renders as error page in the specific language. While before it was always considered as a path of the default language/language without any URL prefix.

### Reasoning
- To me it makes sense with specific prefixes for each language to treat anything within these namespace as only belonging to that particular language. Any other language, also if not defining itself any prefix, should not interfere here.
- Error pages are correctly translated now.
- As this is only used in the catch-all route for each language (https://github.com/getkirby/kirby/blob/main/src/Cms/LanguageRoutes.php#L28-L47), the effect is scoped to only protect the languages from each other, while custom routes etc. are not affected (or only those that apply a language scope - and there I'd say protecting those from each other make sense).

### Additional context
Touching those routing basics etc. is always a delicate matter. I'd be fine with moving this to 5.0, if preferred.

### Fixes
- Error pages are rendered in the current, not always the default language
#4834

## Ready?

### For review team


distantnative (commented):
This probably could use another unit tests, that really checks the behavior in matching full paths in the router. Any `Idea` where to add this unit test?

bastianallgeier (dismissed):

bastianallgeier:
@distantnative I think it's already covered in the `LanguageRoutesTest::testFallback`. To be sure, I added @afbora's test case from the issue. It all seems to be fine to me.

distantnative:
@bastianallgeier I can't follow quite. You just replaced the languages with different ones in the existing test, or? But the test is the same as before. Which then would mean that it doesn't test what this PR fixes, cause otherwise it couldn't have passed the checks before this PR.

bastianallgeier:
I thought that we had some weird edge case when en is the default language. That's why I changed the tests.

distantnative:
My point is: if those tests passed before this PR, they aren't testing the fix this PR provides.

bastianallgeier:
As far as I understood @afbora's comment, it worked for some reason with English before, but not with other languages. But I did not double-check if the new test case would have failed before the PR.

distantnative:
It doesn't fail before this PR either. Will need to look into how a test that really covers the issue needs to look like.

afbora:
I can able to test the PR tonight.

distantnative:
I hate this, cause I cannot find a unit test. Also it's no issue if English is the default language and I don't know why because with the fix in this PR it should also apply to English beforehand. Moving this to 4.5 so we can really be sure that this is the right fix.

afbora:
@distantnative The PR fixes the issue. But I'm not sure about that. Something smells not good. I found a clue while trying to write a unit test for this case For ex:

Following both codes doesn't give same output (current state, not for this PR). I think loading language ways are different.

### Set languages from root argument as default

````php
$app = new Kirby([
    'roots' => [
        'index'     => __DIR__,
        'languages' => __DIR__ . '/site/languages',
        'templates' => __DIR__ . '/site/templates'
    ],
    'site' => [
        'children' => [
            [
                'slug'     	   => 'error',
                'template' 	   => 'error',
                'translations' => [
                    [
                        'code'    => 'fr',
                        'content' => [
                            'title' => 'Erreur'
                        ]
                    ],
                    [
                        'code'    => 'en',
                        'content' => [
                            'title' => 'Error'
                        ]
                    ]
                ]
            ]
        ]
    ]
]);

echo $app->render('en/not-exist');
````

### Set languages from languages argument (for ex: for unit tests)

````php
$app = new Kirby([
    'roots' => [
        'index' => __DIR__,
        'templates' => __DIR__ . '/site/templates'
    ],
    'languages' => [
        [
            'code'     => 'fr',
            'name'    => 'French',
            'default'  => true,
            'url'         => '/'
        ],
        [
            'code'   => 'en',
            'name'  => 'English',
            'url'       => null
        ],
    ],
    'site' => [
        'children' => [
            [
                'slug'     	   => 'error',
                'template' 	   => 'error',
                'translations' => [
                    [
                        'code'    => 'fr',
                        'content' => [
                            'title' => 'Erreur'
                        ]
                    ],
                    [
                        'code'    => 'en',
                        'content' => [
                            'title' => 'Error'
                        ]
                    ]
                ]
            ]
        ]
    ]
]);

echo $app->render('en/not-exist');
````

afbora:
@distantnative Here you can able to test via https://github.com/getkirby/kirby/pull/6608/commits/20c1a1b14d1ef9429af6fc3403655d0ab667fd88 unit test.

distantnative:
@afbora thank you! I will have a closer look on the weekend. I agree that doesn't smell right at first at least.

distantnative:
Closing this for now. We will need to start fresh here.

PR #6610, opened 2024-08-11
User view: always show prev-next

### LINKED ISSUE #4225: Logical inconsistency while navigating through panel users (prev/next buttons disappear on own account)

distantnative:
### Summary of changes
- Prev/next arrows are shown on any user view, also the account view
- `$user->next()`, `$user->prev()` are based on a siblings collection that now is sorted by username (not id)

### Reasoning
- Current state has been too confusing. Maybe the new solution isn't the best one can imagine, but I'd argue better and a lot less confusing.
- Prev/next arrows should move through the same list as shown on users view. That's why the sorting needs to be added to `User::siblingsCollection()` too.

### Additional context
- I think we can improve this a lot more easily once we have view UI classes. That will allow us e.g. to have specific routes for the user view but when filtered by a specific role, where the prev-next arrows could then loop only through the users with that role.

### Enhancements
- User view: show prev/next buttons also on account view

### For review team

### DIFF
--- a/src/Cms/User.php
+++ b/src/Cms/User.php
@@ -666,7 +666,7 @@ class User extends ModelWithContent
 	 */
 	protected function siblingsCollection(): Users
 	{
-		return $this->kirby()->users();
+		return $this->kirby()->users()->sortBy('username', 'asc');
 	}
 
 	/**
--- a/src/Panel/User.php
+++ b/src/Panel/User.php
@@ -223,7 +223,7 @@ class User extends Model
 
 		return array_merge(
 			parent::props(),
-			$account ? [] : $this->prevNext(),
+			$this->prevNext(),
 			[
 				'blueprint' => $this->model->role()->name(),
 				'model' => [


bastianallgeier:
I think the most confusing part is that we jump to the account view for the current user. We should probably separate this and have an account view and a user view for the current user.

distantnative:
@bastianallgeier I think I found a relatively simple way that seems to work

bastianallgeier:
The option idea is great. But unfortunately, we can't simply change the path method like that. It's needed in other places as well. I guess it's best to untangle that as well in this PR.

distantnative:
@bastianallgeier Do you have an example? I've looked and found other places, but all would work also with the normal user path instead of account as far as I can tell.

distantnative:
@bastianallgeier removed that part again. I still think that this PR improves the situation, even if it's not the perfect result.

afbora:
I like this solution. It's easy and simple 👍 

If we're still not sure and it confusing, I have an another idea that remove current user from users collection only for the pagination. Maybe it's stupid idea 🙈

PR #6612, opened 2024-08-11
Fix user `create` & `changeRole` permissions/options handling in Panel UI

### LINKED ISSUE #5146: user permissions: changeRole has no effect
**hey there lovely kirby team 🌻**
i am struggling with two things related to user permissions. i don’t know if they are related.
[this](https://github.com/getkirby/kirby/issues/5147) is the other issue. feel free to merge if needed.

## description
i would like to control what roles users with a certain role are allowed to change. for example a `manager` should be able to change roles for other users. but an `assistant` shouldn’t be able to do that. when i set the `changeRole` permissions to `true` or `false` in my user blueprints it doesn’t seem to have any effect.

also it seems like there is a difference between [this](https://getkirby.com/docs/reference/panel/blueprints/user) doc and [this](https://getkirby.com/docs/guide/users/permissions) doc. in the first one it seems like `changeRole` is only available for `user`, not `users`. in the second one it seems like it’s available for both.

Screenshot 2023-04-13 at 18 40 22

Screenshot 2023-04-13 at 18 38 58

## to reproduce
start with a clean starterkit. add a new user blueprint with the following permissions.

```yml
permissions:
    user:
        delete: false # works, this user can’t delete itself
        changeRole: true # doesn’t work, this user can’t change its role
    users:
        changeRole: true # doesn’t work, this user can’t change other user’s roles
```

Screenshot 2023-04-13 at 18 27 13

## your setup
kirby 3.9.3

thank you for your help!
🤍 ✨

### LINKED ISSUE #5147: user permissions: support for `options` broken
**hey there lovely kirby team 🌻**
i am struggling with two things related to user permissions. i don’t know if they are related.
[this](https://github.com/getkirby/kirby/issues/5146) is the other issue. feel free to merge if needed.

## description
i would like users with some roles to be able to create only users with certain roles. for example a `manager` should be able to add an `assistant`. but an `assistant` shouldn’t be able to add a `manager`. both roles however should be able to add a `customer`. to me this sounds like a very common use-case.

i found [this](https://forum.getkirby.com/t/setting-default-user-role-on-creating-in-panel/14927/4) post by @texnixe from some time ago but the code doesn’t seem to work anymore. don’t know if it ever did, but seems like it. also seems like `options` have been removed from user blueprints, and now there’s only `permissions`. at least this is what the latest docs say. anyway, something like the code down below also doesn’t seem to work.

could you please let me know if this is a bug? or a feature request? i guess i could build a workaround with hooks. so when a user tries to create another user with a certain role, an error is being shown. but i would like if some users can’t even see all roles in the "add a new user" modal.

## to reproduce
```yml
permissions:
    users:
        create:
            manager: false
            assistant: true
            customer: true
```

## your setup
kirby 3.9.3

thank you for your help!
🤍 ✨

distantnative:
## Description

⚠️ Trying to split this into smaller, more modular PRs.
- [x] https://github.com/getkirby/kirby/pull/6653
- [ ] https://github.com/getkirby/kirby/pull/6654
- [x] https://github.com/getkirby/kirby/pull/6655
- [x] https://github.com/getkirby/kirby/pull/6657
- [ ] https://github.com/getkirby/kirby/pull/6658
- [ ] https://github.com/getkirby/kirby/pull/6694

### Summary of changes
- `Cms\User::roles()`:
  - So far, this only handled admin vs. anybody else: admins would get all roles, everyone else just their own role. This is not what the method promises to return.
  - Changed it to return the actual list of available roles for any users.
  - Also checks permissions and user options: `User::roles($context)` and `App::roles($context)` which apply `->canBeChanged()` or `->canBeCreated()` accordingly to the context.
- Panel dialogs: `user.create` and `user.changeRole`
  - Pass `$roles` to `Panel\Field::role()` using the right `$context` to retrieve the roles depending on their action
  - Show a radio input, even if only one option is available (though to make it transparent what gets created)
- Resorted UserPermissions and UserRules
  - `UserPermssions::canChangeRole()` can be true even if only one role is available - as this isn't a permissions restriction. Instead this check is performed separately  where the UI needs it.

### Reasoning
The main issue of #5146 why the permission wasn't reflected in the UI, was `UserPermissions::canChangeRole` only checking if `User::roles()` returns more than one role. However, `User::roles()` would always only return the current role for non-admins. That's why the permission itself never was acknowledged. Fixing these now really relies on the permission.

Moreover, in the refactoring to Panel backend dialogs etc. it got lost to filter available roles further by applying `->canBeChanged()` or `->canBeCreated()`.

### Additional context
As `UserRules::changeRole()` was fully restrictive in place, this is no security fix - the permission was always fully enforced. Even more, the Panel UI was falsely more restrictive than it should've been based on the permissions.

### Fixes
- Panel UI fixed for `create` and `changeRole` permission and user options
#5146

### Enhancements
- Role always shown when creating a new user, even if only one role available

## Docs

- [ ] Consistent listing of `user.changeRole` AND `users.changeRole`

## Ready?

### For review team

### DIFF
--- a/config/api/routes/roles.php
+++ b/config/api/routes/roles.php
@@ -8,13 +8,9 @@ return [
 		'pattern' => 'roles',
 		'method'  => 'GET',
 		'action'  => function () {
-			$kirby = $this->kirby();
-
-			return match ($kirby->request()->get('canBe')) {
-				'changed' => $kirby->roles()->canBeChanged(),
-				'created' => $kirby->roles()->canBeCreated(),
-				default   => $kirby->roles()
-			};
+			$kirby   = $this->kirby();
+			$context = $kirby->request()->get('canBe');
+			return $kirby->roles($context);
 		}
 	],
 	[
--- a/config/areas/users/dialogs.php
+++ b/config/areas/users/dialogs.php
@@ -21,9 +21,14 @@ return [
 		'load' => function () {
 			$kirby = App::instance();
 
+			// get role field definition, incl available role options
+			$roles = Field::role(context: 'create', props: [
+				'required' => true
+			]);
+
 			// get default value for role
 			if ($role = $kirby->request()->get('role')) {
-				$role = $kirby->roles()->find($role)?->id();
+				$role = $kirby->roles()->canBeCreated()->find($role)?->id();
 			}
 
 			return [
@@ -39,9 +44,7 @@ return [
 						'translation'  => Field::translation([
 							'required' => true
 						]),
-						'role' => Field::role(props: [
-							'required' => true
-						])
+						'role' => $roles
 					],
 					'submitButton' => I18n::translate('create'),
 					'value' => [
@@ -49,7 +52,7 @@ return [
 						'email'       => '',
 						'password'    => '',
 						'translation' => $kirby->panelLanguage(),
-						'role'        => $role ?? $kirby->user()->role()->name()
+						'role'        => $role ?? $roles['options'][0]['value'] ?? null
 					]
 				]
 			];
@@ -228,7 +231,7 @@ return [
 				'component' => 'k-form-dialog',
 				'props' => [
 					'fields' => [
-						'role' => Field::role($user, [
+						'role' => Field::role($user, 'change', [
 							'label'    => I18n::translate('user.changeRole.select'),
 							'required' => true,
 						])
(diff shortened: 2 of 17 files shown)


distantnative:
Closing this as we now have all parts from this PR as smaller PRs ready.

PR #6617, opened 2024-08-15
`Url::slug()` that obeys `maxlength` option

### LINKED ISSUE #6526: Use `slugs.maxlength` when generating slugs
Currently, we often ignore the `slugs.maxlength` option when generating slugs but rather use the default 128 from `Str::slug()`.

I see two options:

1. Integrate `App(null, true)?->option('slugs.maxlength')` into `Str::slug()` to use as default when set. Could this be problematic if someone limits it to a very short length, as we also often use `Str::slug()` for non-slugs?

2. Create a new, e.g. `App::slug()` that takes the option and passes it with all the other parameters to `Str::slug()`. Search through the code where we should use `App::slug()` instead of `Str::slug()`. (or `Page::slug()`, `Router::slug()` or so).

@lukasbestle @bastianallgeier Which do you think would be better?

distantnative:
### Summary of changes
- New `Url::slug()` method that obeys `slugs.maxlength` option
- Using `Url::slug()` instead of `Str::slug()` everywhere that is about page slugs
- `Str::slug()` supports `false` as maxlength to disable shortening

### Reasoning
Documented option https://getkirby.com/docs/reference/system/options/slugs#set-allowed-max-length-of-slugs has no effect

### Fixes
- `slugs.maxlength` option works now 
https://github.com/getkirby/kirby/issues/6526

## Ready?

### For review team

### DIFF
--- a/config/areas/site/dialogs.php
+++ b/config/areas/site/dialogs.php
@@ -431,7 +431,7 @@ return [
 				];
 			}
 
-			$slugAppendix  = Str::slug(I18n::translate('page.duplicate.appendix'));
+			$slugAppendix  = App::slug(I18n::translate('page.duplicate.appendix'));
 			$titleAppendix = I18n::translate('page.duplicate.appendix');
 
 			// if the item to be duplicated already exists
--- a/config/areas/site/dialogs.php
+++ b/config/areas/site/dialogs.php
@@ -3,6 +3,7 @@
 use Kirby\Cms\App;
 use Kirby\Cms\Find;
 use Kirby\Cms\PageRules;
+use Kirby\Cms\Url;
 use Kirby\Exception\Exception;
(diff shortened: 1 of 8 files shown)


bastianallgeier (commented):
I wonder if App is a good place for this. Maybe use the URL class instead? I probably wouldn't even mind to look for the app option in the `Str::slug` method to be honest.

distantnative:
@bastianallgeier I don't think we can add it to the `Str` method. Not because of using an app option, but because we use `Str::slug()` in many contexts that are not related at all to the option. And we should not affect those.

bastianallgeier (approved):
The placement is still a bit weird after looking at it again. But I really can't think of any other place and I guess we don't have to be too nitpicky about it.

PR #6618, opened 2024-08-15
Exceptions: support named arguments

distantnative:
### Summary of changes
- Exception classes support named arguments
- Doesn't check anymore for `App` class existence - we don't want to support using the package outside of Kirby

### Reasoning
- Modernizing code
- Will allow for easier additions, e.g. `log` argument

### Additional context
- Duplicated unit tests, so they cover array as well as named arguments for the same tests.

### Refactored
- Exception classes support named arguments

### Deprecated
- Exception classes:  passing `$arg` array will be removed in a future version. Use named arguments.

## Ready?

### For review team

### DIFF
--- a/src/Exception/Exception.php
+++ b/src/Exception/Exception.php
@@ -102,14 +102,12 @@ class Exception extends \Exception
 
 		if ($message !== null) {
 			$this->isTranslated = false;
-			return parent::__construct($message);
+			parent::__construct($message);
+			return;
 		}
 
 		// define whether message can/should be translated
 		$translate = $args['translate'] ?? $translate;
-
-		// fallback waterfall for message string
-		$message = null;
 
 		// a. translation for provided key in current language
 		// b. translation for provided key in default language
--- a/src/Exception/Exception.php
+++ b/src/Exception/Exception.php
@@ -2,10 +2,10 @@
 
 namespace Kirby\Exception;
 
-use Kirby\Cms\App;
 use Kirby\Http\Environment;
 use Kirby\Toolkit\I18n;
 use Kirby\Toolkit\Str;
+use Throwable;
 
 /**
  * Exception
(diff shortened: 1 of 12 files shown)


bastianallgeier (commented):
That's such a great idea to deprecate the props/args arrays 😍

I wasn't able to fully test the PR before leaving. But it looks good on first sight and I really love the direction. Could be in alpha-3

distantnative:
@bastianallgeier I already have a local branch that switches (almost) all our uses over to named arguments. It's really nice but probably will be hard/impossible to review.

src/Exception/Exception.php
```diff
@@ -56,78 +58,90 @@ class Exception extends \Exception
...
+			return parent::__construct($message);
 		}
 
-		// set the Exception code to the key
-		$this->code = $key;
+		// define whether message can/should be translated
+		$translate = $args['translate'] ?? $translate;
+
+		// fallback waterfall for message string
+		$message = null;
```
afbora: I think `$message` is always `null` for this line.

src/Exception/Exception.php
```diff
@@ -56,78 +58,90 @@ class Exception extends \Exception
...
+		if (Str::startsWith($this->code, self::$prefix . '.') === false) {
+			$this->code = self::$prefix . '.' . $this->code;
 		}
 
 		if (is_string($args) === true) {
+			$message ??= $args;
+		}
+
+		if ($message !== null) {
 			$this->isTranslated = false;
```
afbora, line 104: Still needed? Because returns new object in following line.

distantnative: Doesn't really return a new object as `parent::contract()` still modifies the current object. Could also be written as

```php
parent::__construct();
return;
```

src/Exception/Exception.php
distantnative: This is to ensure that `$message` exists in any case as it then will be filled inside some if conditions. Otherwise it would not exist for cases that match no if statement.

afbora: I think it always exists since we have `$message` named argument and it's `null` since we have `$message !== null` check in above line.

distantnative: 👍 now understood it, yes

bastianallgeier (approved):

PR #6622, opened 2024-08-16
Fix updating new language variables for secondary languages

afbora:
## Description

If you try to edit (without any update, open dialog and try to submit) new language variables from secondary language, you'll get following error in dialog.

> Kirby\Cms\LanguageVariable::update(): Argument # 1 ($value) must be of type string, null given, called in \kirby\config\areas\languages\dialogs.php on line 258

### Fixes

- Fixed updating new language variables for secondary languages

### Breaking changes
None 

## Ready?

### For review team

### DIFF
--- a/config/areas/languages/dialogs.php
+++ b/config/areas/languages/dialogs.php
@@ -256,7 +256,7 @@ return [
 		},
 		'submit' => function (string $languageCode, string $translationKey) {
 			Find::language($languageCode)->variable($translationKey, true)->update(
-				App::instance()->request()->get('value')
+				App::instance()->request()->get('value', '')
 			);
 
 			return true;
--- a/src/Cms/LanguageVariable.php
+++ b/src/Cms/LanguageVariable.php
@@ -56,7 +56,7 @@ class LanguageVariable
 			throw new DuplicateException('The variable is part of the core translation and cannot be overwritten');
 		}
 
-		$translations[$key] = trim($value ?? '');
+		$translations[$key] = $value ?? '';
 
 		$language->update(['translations' => $translations]);
 
@@ -102,10 +102,10 @@ class LanguageVariable
 	/**
 	 * Sets a new value for the language variable
 	 */
-	public function update(string $value): static
+	public function update(string|null $value = null): static
 	{
-		$translations = $this->language->translations();
-		$translations[$this->key] = $value;
+		$translations             = $this->language->translations();
+		$translations[$this->key] = $value ?? '';
 
 		$language = $this->language->update(['translations' => $translations]);
 


config/areas/languages/dialogs.php
```diff
@@ -249,7 +249,7 @@
 					'size'   => 'large',
 					'value'  => [
 						'key'   => $variable->key(),
-						'value' => $variable->value()
+						'value' => $variable->value() ?? ''
```
distantnative: Wondering if it should rather be fixed on line 259, where the error message also occurs. Something like

```php
App::instance()->request()->get('value', '')
````

Comment (afbora) [reply config/areas/languages/dialogs.php]: My first commit was that but I prefer like in PR. Both solution are good for me.

Comment (afbora) [reply config/areas/languages/dialogs.php]: I've pushed both changes. What do you think?

Comment (distantnative) [reply config/areas/languages/dialogs.php]: Sorry for turning in circles, but I think maybe the actual problem is in the `LanguageVariable` class. We do allow creating one with `value = null`, so maybe it would rather be more consistent, that also `LanguageVariable::update()` allows passing `null`.

File: src/Cms/LanguageVariable.php [hunk 2]
```diff
@@ -102,10 +102,10 @@ public function key(): string
 	/**
 	 * Sets a new value for the language variable
 	 */
-	public function update(string $value): static
+	public function update(string|null $value = null): static
 	{
-		$translations = $this->language->translations();
-		$translations[$this->key] = $value;
+		$translations             = $this->language->translations();
+		$translations[$this->key] = $value ?? '';
```
distantnative, line 108: `Question:` Do we need the fallback to an empty string? What happens when it just stays null?

`Todo:` We would need to add a unit test for passing null to `::update()`

afbora: If stays `null`, language variables comes `null`. Do you think it's ok? I don't think so.

````php
'translations' => [
    'test-variable' => NULL
],
````

distantnative: So how does that array look before when you got the error? I'd think it's also `=> NULL` then that let to the error, isn't it?

afbora: No. When adding a variable (without a value) for the default language, an empty string is already added:
https://github.com/getkirby/kirby/blob/fix/updating-null-language-variables/config/areas/languages/dialogs.php#L195

However, since that variable does not exist in the secondary language yet, it returns `null` and if no update is made in the dialog, it tries to save `null` and throws an exception.

`create` solved this (with empty string fallback) in the dialog process, we can solve it the same way for `update`.

distantnative: Let me have a closer look during the week. I think we have a problem if the dialogs behave differently than calling the methods programmatically. That's why I think it makes sense to spend some time here finding the ideal solution - all your solutions are functional patches for sure.

distantnative: @afbora finally had a closer look: I agree with you

- `language.translation.update` dialog: in the `submit` handler, we should probably add an empty string as fallback `->get('value', '')` to match the behavior of the create dialoh
- `LanguageVariable::update()` should indeed add a fallback to an empty string as you did.
- Should it also add `trim()` for the value as `::create()` does?

afbora: Actually we should remove the `trim` function for create method. I think user is responsible for the spaces. May be user needs spaces start/end of string. What do you think? @distantnative

distantnative: True, could be a valid use case. So aas you said, maybe remove it from create.

afbora: What do you think? @bastianallgeier

bastianallgeier: I think Nico's summary sounds good but I agree with you that we should remove trim entirely.

bastianallgeier (approved):

PR #6625, opened 2024-08-17
Audio & video file preview

distantnative:
### Summary of changes
- Added file previews for audio and video
- Improved pattern background in darkmode

### Reasoning
I kept it as basic as possible so far with the native audio and video tags. For me testing it with Firefox, Safari and Chrome looked fine and worked well.

### Features
- New file previews for audio and video files
Screenshot 2024-08-17 at 11 30 19 AM
Screenshot 2024-08-17 at 11 30 27 AM

## Ready?

### For review team


panel/src/components/Views/Files/VideoFilePreview.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+	<div class="k-video-file-preview">
+		<k-file-preview-thumb :options="options">
```
bastianallgeier, line 3: `Suggestion`: Looking at this in the new context, I was wondering if we should maybe use `k-file-preview-frame` instead. It could maybe be a little more universal instead of thumb.

bastianallgeier:
I've added lab examples. I think we can delay the thumbs decision and merge this PR. The new previews are awesome.

bastianallgeier (approved):

panel/src/components/Views/Files/VideoFilePreview.vue
distantnative: I agree, will commit the change directly to `v5/develop`

PR #6626, opened 2024-08-17
Improve code style `Collection::sort()`

distantnative:
### Summary of changes
- Cleaned up `Collection::sort()` using some more modern PHP code, e.g `array_key_last()`

### Reasoning
- Moving our code base along bit by bit

## Ready?

### DIFF
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -965,10 +965,7 @@ class Collection extends Iterator implements Stringable
 			$field = array_key_last($fields);
 
 			// normalize $arg
-			$arg = match (true) {
-				is_string($arg) => strtolower($arg),
-				default         => $arg
-			};
+			$arg = is_string($arg) === true ? strtolower($arg) : $arg;
 
 			// $arg defines sorting direction
 			if (
@@ -991,10 +988,7 @@ class Collection extends Iterator implements Stringable
 						// make sure that we return something sortable
 						// but don't convert other scalars (especially numbers)
 						// to strings!
-						return match (true) {
-							is_scalar($value) => $value,
-							default           => (string)$value
-						};
+						return is_scalar($value) === true ? $value : (string)$value;
 					})
 				];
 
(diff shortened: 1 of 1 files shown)


src/Toolkit/Collection.php
```diff
@@ -961,49 +961,63 @@ public function sort(...$args): static
 		$fields = [];
 
 		foreach ($args as $arg) {
-			// get the index of the latest field array inside the $fields array
-			$currentField = $fields ? count($fields) - 1 : 0;
+			// get the index of the latest field array inside $fields
+			$field = array_key_last($fields);
```
bastianallgeier, line 965: Uhhh, I didn't know about this one. That's nice 💛

src/Toolkit/Collection.php
```diff
@@ -961,49 +961,63 @@ public function sort(...$args): static
...
 
-			// detect the type of argument
-			// sorting direction
-			$argLower = is_string($arg) ? strtolower($arg) : null;
+			// normalize $arg
+			$arg = match (true) {
+				is_string($arg) => strtolower($arg),
+				default         => $arg
+			};
```
bastianallgeier: As much as I love the match clauses, I'm not sure I prefer it here over the old version to be honest. I found that line to be very readable. I would always use match when the optional lines get too long and complicated to read, but here it's really short and quite easy to scan.

src/Toolkit/Collection.php
```diff
@@ -961,49 +961,63 @@ public function sort(...$args): static
...
+					'values' => A::map($array, function ($value) use ($collection, $arg) {
+						$value = $collection->getAttribute($value, $arg);
+
+						// make sure that we return something sortable
+						// but don't convert other scalars (especially numbers)
+						// to strings!
+						return match (true) {
+							is_scalar($value) => $value,
+							default           => (string)$value
+						};
```
bastianallgeier: It's the same for me here. I can definitely live with the match clause. I'm really just not sure if it is worth it to refactor such short lines.

src/Toolkit/Collection.php
distantnative: For me personally it's easier to read logic vertically, I think. I always have troubles with the inline syntax to spot where is the condition where does the true-value start, where the false-value. But it's not a big issue. If you and/or others prefer it, I'll turn it back. 👍

bastianallgeier: I've just converted the three lines back.

bastianallgeier (approved):

PR #6628, opened 2024-08-17
Rector code quality improvements

distantnative:
### Summary of changes
See individual commits. Each corresponds to one Rector rule that made sense to me for our project.

Best to review them commit by commit.

### Reasoning
- I think the most unfamiliar/controversial could be to compare directly to an empty array (`$a === []`/`$a !== []`) instead of using `empty()` or `count()`. Allegedly, it is faster as `count()` -  for me it feels very unfamiliar, but I do like the brevity and explicitness.

## Ready?

### DIFF
--- a/src/Api/Upload.php
+++ b/src/Api/Upload.php
@@ -271,7 +271,7 @@ readonly class Upload
 		array $errors
 	): array {
 		if (count($uploads) + count($errors) <= 1) {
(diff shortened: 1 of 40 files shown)


afbora (approved):
LGTM 👍 

An article about using `empty` function:
https://dev.to/klnjmm/never-use-empty-function-in-php-4pb0

PR #6629, opened 2024-08-18
Email 1 - `Address` class

distantnative:
### Summary of changes
- Added new `Email\Address` class that handles all address parsing and resolving for from, to, bcc, cc, replyTo
- Removed deprecated `::clone()` method

### Reasoning
- First step in trying to untangle and modernize the email classes
- Trying to keep it without breaking changes yet

### Removed deprecated code
- `Email\Email::clone()`

## Ready?

### For review team

### DIFF
--- a/src/Email/Address.php
+++ b/src/Email/Address.php
@@ -1,0 +1,116 @@
+<?php
+
+namespace Kirby\Email;
+
+use Kirby\Cms\User;
+use Kirby\Cms\Users;
+use Kirby\Exception\InvalidArgumentException;
+use Kirby\Toolkit\A;
+use Kirby\Toolkit\V;
+
+/**
+ * An email address with optional name
+ *
+ * @package   Kirby Email
+ * @author    Nico Hoffmann <nico@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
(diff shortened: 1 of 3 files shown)


src/Email/Address.php
```diff
@@ -0,0 +1,116 @@
...
+	public function name(): string|null
+	{
+		return $this->name;
+	}
+
+	/**
+	 * Returns one or multiple addresses as array
+	 * where the emails are the keys and the names the values
+	 */
+	public static function resolve(array|Address $address): array
```
afbora, line 92: I think array usage that doesn't includes `Address` object is invalid:

````php
// single
Address::resolve(['homer@simpson.com']);

// multiple
Address::resolve([
  'homer@simpson.com' => null,
  'lisa@simpson.com'  => 'Lisa Simpson'
]);
````

If array items are should to be `Address` objects, it would be nice to have a type check.

distantnative: I'll make the method description more explicit. It is only meant to resolve Address objects and arrays of Address objects. There should be no need to support other formats.

afbora: Conversation is resolved but I couldn't see any changes?

distantnative:
Closing this for a moment, so we can better focus on some bugs

PR #6635, opened 2024-08-20
Calendar input: first day of week based on locale

distantnative:
### Summary of changes
- New `Date::firstWeekday()` method that receives a locale and returns `0`-`6`
- New `date.weekday` config option as override 
- New `$translation.weekday` fiber datapoint
- Calendar input calculates calendar grid based on first weekday defined by `$translation.weekday `

### Additional context
Great to test by opening a calendar dropdown and changing the config value via the browser console

```js
window.panel.translation.weekday = 0 // Sunday
window.panel.translation.weekday = 1  // Monday
```

### Enhancements
- Calendar input: first day of the week is chosen based on user language or `date.weekday` config option  (`0` for Sunday ... `6` for Saturday)
https://feedback.getkirby.com/416

### Breaking changes
- Calendar dropdown input will show Sunday now as first day of the week (depending on the user's language). If you want to enforce Monday as first day of the week, you can set the `date.weekday` option to `1`.

## Ready?

### For review team


panel/src/components/Forms/Input/CalendarInput.vue
```diff
@@ -119,21 +121,21 @@ export default {
...
 		/**
-		 * Translated weekday names
+		 * Ordered weekday names abbreviations
 		 * @returns {array}
 		 */
 		weekdays() {
-			return ["mon", "tue", "wed", "thu", "fri", "sat", "sun"].map((day) =>
-				this.$t("days." + day)
-			);
+			const first = this.$panel.config.weekday - 1;
```
distantnative: `Discuss`: should this be a prop to easily have Lab examples? Or is that overdoing it?

afbora:
I would expect to set based on user language or locale option. It works only with `date.weekday` option for me. Am I missing something? Because PR title is `first day of week based on locale` 🙂

distantnative:
@afbora hmmm seems to work for me:

### English
Screenshot 2024-08-21 at 5 45 24 PM
Screenshot 2024-08-21 at 5 44 38 PM

### German
Screenshot 2024-08-21 at 5 45 31 PM
Screenshot 2024-08-21 at 5 44 54 PM

~Though I had to do a hard reload before it showed, which shouldn't be the case.~ (fixed that)

afbora:
@distantnative Sorry, I tested in single language setup. It doesn't work for user panel language (not content language) and `locale` option from config file.

distantnative:
@afbora good catch - I think I'll actually have to move it over to the `$translation` object to work right for single lang. Looking at the `locale` option though, not sure this would be work as override. As it is based on the user language and the user always has a language, it will load the i18n file and use that locale. I could add support for the override, but I don't think it would be consistent. To my knowledge, the `locale` option doesn't affect the Panel UI language in any way so far, does it?

src/Cms/Translation.php
```diff
@@ -83,6 +84,26 @@ public function direction(): string
...
+		$weekday = App::instance()->option('date.weekday');
+
+		if (is_int($weekday) === true) {
+			return $weekday;
+		}
+
+		$locale   = $this->locale();
+		$calendar = IntlCalendar::createInstance(null, $locale);
+
+		return $calendar->getFirstDayOfWeek() - 1;
```
afbora: This enhancement requires `intl` php extension even `date.handler` haven't set `intl`. I think this is a breaking change for users who not installed `intl` php extension. 

I have a solution for this issue and will be only `date.weekday` option available for this idea. Roughly we can check the date handler or check the `intl` extension. I'm not sure which one is best way but checking `date.handler` seems more consist.

```suggestion
		$kirby   = App::instance();
		$weekday = $kirby->option('date.weekday');

		if (is_int($weekday) === true) {
			return $weekday;
		}
		
		// returns sunday as default first day of week if date handler is not `intl`
		if ($kirby->option('date.handler') !== 'intl') {
			return 0;
		}

		$locale   = $this->locale();
		$calendar = IntlCalendar::createInstance(null, $locale);

		return $calendar->getFirstDayOfWeek() - 1;
```

afbora: What would happen if `$locale` is incorrect, invalid or not found by server? `::getFirstDayOfWeek` returns `int|false`. Are we sure that returns always int for this line?

src/Cms/Translation.php
```diff
@@ -84,6 +85,39 @@ public function direction(): string
...
+		$weekday = $kirby->option('date.weekday');
+
+		if (is_int($weekday) === true) {
+			return $weekday;
+		}
+
+		// returns sunday as default first day of week
+		// if date handler is not `intl`
+		if ($kirby->option('date.handler') !== 'intl') {
+			return 0;
```
distantnative: `Discuss`: Should this rather be `1` for Monday as default to have a smaller breaking change?

afbora: > According to ISO 8601:2004, the Gregorian calendar consists of 52 or 53 seven-day weeks, and Monday is designated as the first day of the week.

I think you're right.

afbora:
@distantnative I just wonder why you moved this enhancement to v5? Currently doesn't includes a breaking change since now we have `date.handler !== intl` check.

distantnative:
@afbora because currently without `intl` handler, Sunday would be the first day of the week. And so far (Kirby 4.3.1) Monday is the default. So for anyone without `intl` (I guess quite a few) this would be a big UX breaking change.

afbora:
I've wrote a comment about first day of week. https://github.com/getkirby/kirby/pull/6635#discussion_r1740079709

distantnative:
👍 yes I saw. This will reduce the number of breaking changes.

Still thinking whether this should be better in v5 as it could still be a breaking change for those with `intl` that prefer Monday to be the first day. And so far it has been Monday. But with this change it could show up as Sunday for them (if they have a Sunday-first locale).

afbora (approved):
Great 💛

PR #6638, opened 2024-08-23
Improve `$field->isEmpty()`

### LINKED ISSUE #6637: Symfony YAML handler stores non-empty empty array value
## Description
When using the Symfony YAML handler – `'yaml.handler' => 'symfony'` – Kirby will create non-empty values for empty content. Given a files field, removing all previously selected files will be stored as `[]` in the content file. Using the default YAML handler will result in an empty string.

**Result with Symfony YAML handler:**

```yaml
Files: []
```

**Result with default YAML handler:**

```yaml
Files:
```

This divergent behaviour will cause issues when using `isEmpty` or methods relying on this method (like `isNotEmpty` or `or`) because `isEmpty` checks on the plain field value. So even if `isNotEmpty` will return `true` on an object, as consecutive call to `toFile()` will still result in an empty object. This is unexpected and does break templating logic.

**Expected behavior**  
Empty values should be identified correctly independent of the chosen YAML handler.

**Scope**
This will affect any field storing collection-like content.

## To reproduce
1. Create a files field.
2. Select a file, save.
3. Unselect that file, save.
4. Check the stored value in the content file.
5. Repeat with different YAML formatter.

## Your setup
**Kirby Version**  
4.4.0 RC1 but it should be the same issue ever since YAML handlers have been introduced.
**PHP Version**
In case this matters, it happens on PHP 8.2 for us.

distantnative:
### Summary of changes
- Removed `empty()` check from `$field->isEmpty()` as it checks for too many false positives
- Instead checking all relevant cases explicitly
- Will also consider `'[]'` as empty (Symfony parser stores empty arrays like that)
- WIll also consider empty strings (just spaces) as empty

### Fixes
- Fixed `$field->isEmpty()` for some empty arrays
#6637

## Ready?

### For review team

### DIFF
--- a/src/Content/Field.php
+++ b/src/Content/Field.php
@@ -124,9 +124,17 @@ class Field
 	 */
 	public function isEmpty(): bool
 	{
+		$value = $this->value;
+
+		if (is_string($value) === true) {
+			$value = trim($value);
+		}
+
 		return
-			empty($this->value) === true &&
-			in_array($this->value, [0, '0', false], true) === false;
+			$value === null ||
+			$value === '' ||
+			$value === [] ||
+			$value === '[]';
 	}
 
 	/**
--- a/tests/Content/FieldTest.php
+++ b/tests/Content/FieldTest.php
@@ -168,6 +168,11 @@ class FieldTest extends TestCase
 			['false', false],
 			[null, true],
 			['', true],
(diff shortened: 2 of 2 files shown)


bastianallgeier (approved):
Much better!

Issue #6640, opened 2024-08-26
Same Uuid is re-used every time when creating a child page


ovenum:
## Description

When a parent page contains a child page with slug _new_ Kirby will re-use that child pages Uuid when a new child page is created. 

**Expected behavior**  

The Uuid to be unique : )

## To reproduce

Using a fresh plainkit installation:

1. Create a new Page from the panel named Some Page (name does not matter, it’s used just for reference)
3. _Manually_ create an unlisted child page with slug new and _manually_ add a Uuid field 
```yaml
# /content/some-page/new/default.txt
Uuid: 7XTwwjop6ogKKCEW
```
4. Using the Panel create another child page of Some Page 
5. Compare the Uuid of the child pages created via the panel to the Uuid of some-page/new.
They are the same

**Note**
When the Uuid of the draft child page created via the panel is deleted manually, Kirby will correctly assign a new, truly unique Uuid to that page when it’s requested by the panel. 

## Your setup
**Kirby Version**  
4.3.0

distantnative:
I can reproduce the issue with the steps you've outlined. The `new` name seems to be crucial - renaming that sibling page to `neww` and a new UUID is correctly generated when adding a sibling.

distantnative:
Guilty lines:
https://github.com/getkirby/kirby/blob/main/src/Panel/PageCreateDialog.php#L251
https://github.com/getkirby/kirby/blob/main/src/Panel/PageCreateDialog.php#L270

ovenum:
Thanks for looking into this @distantnative.
Is the temporary page created by https://github.com/getkirby/kirby/blob/main/src/Panel/PageCreateDialog.php#L251 created in the filesystem and if so how is it removed, renamed? 

I'm quite confident that a page with the slug `new` was not created by a user logged in to the panel. Is it possible that it’s an artifact of the method mentioned above that was not cleaned up properly?
The site in question had a few issues with oom-kill shutting down php-fpm, so this could have been caused by a very rare edge case. 🤷‍♂️

distantnative:
> I'm quite confident that a page with the slug new was not created by a user logged in to the panel.

But isn't that your use case you outlined in the top post to reproduce this issue?

[...]

PR #6643, opened 2024-08-27
PageCreateDialog: prevent tmp using actual page

distantnative:
## Description

@afbora Would appreciate your testing skills for this.

### Summary of changes
- Page create dialog's temporary page object would use actual page if called `new`
- Not a perfect solution (ideal will be real in-memory virtual pages), but a lot less likely for a real-world page `__new__`

### Fixes
- Fixed issue where the page create dialog would use an existing page `new` instead just creating a temporary object
 #6640 

## Ready?

### For review team

### DIFF
--- a/src/Panel/PageCreateDialog.php
+++ b/src/Panel/PageCreateDialog.php
@@ -249,7 +249,7 @@ class PageCreateDialog
 	public function model(): Page
 	{
 		return $this->model ??= Page::factory([
-			'slug'     => 'new',
+			'slug'     => '__new__',
 			'template' => $this->template,
 			'model'    => $this->template,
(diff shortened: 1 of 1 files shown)


afbora (dismissed):
I'm not sure what risks there might be in cloning an existing page rather than creating a new `tmp` page, but the issue fixed and everything seems to work fine.

LGTM 👍

distantnative:
@afbora as @ovenum reports that the `new` page actually gets created but Kirby itself, this isn't ready as we probably haven't fully caught the bug.

distantnative:
Closing this as we found out that a `->uuid()` call is the root of the actual issue. We'll be able to fix this better in Kirby 5 using truly in-memory pages.

afbora (approved):

PR #6646, opened 2024-08-29
Language view buttons (incl. permission check)

distantnative:
### Summary of changes
- Renamed `k-languages-view-button` to `k-content-translations-view-button`
- Added dedicated `LanguageAddButton`, `LanguageDeleteButton` and `LanguageSettingsButton` classes
- Incorporating permissions check for disabled state 
- Fixing LanguageConversionTest after merging 4.3.1 with stricter language permission checks

## Ready?


config/areas/languages/buttons.php
```diff
@@ -1,35 +1,21 @@
...
+	'language.settings' => fn (Language $language) =>
+		new LanguageSettingsButton($language),
 	'language.remove' => function (Language $language) {
 		if ($language->isDeletable() === true) {
-			return new ViewButton(
-				dialog: 'languages/' . $language->id() . '/delete',
-				icon: 'trash',
-				title: I18n::translate('delete'),
-			);
+			return new LanguageDeleteButton($language);
```
bastianallgeier, line 18: `Reasoning`: Why did you go for `LanguageDeleteButton` instead of `LanguageRemoveButton`.

config/areas/site/buttons.php
```diff
@@ -25,7 +25,7 @@
 	// `languages` button needs to be in site area, as languages area itself
 	// is only loaded when in multilang setup
 	'languages' => function () {
-		return new LanguagesButton();
+		return new ContentTranslationsButton();
```
bastianallgeier: `Reasoning` Could you explain the change here as well?

bastianallgeier (commented):
I really love the button classes. Don't get my inline comments wrong. I really see that we often are not super consistent between model method names and UI routes and components. I just really want to understand why you made the naming decisions.

config/areas/site/buttons.php
distantnative: Mainly cause I thought with the classes we could be a but more explicit, while I get that we use very short strings for referencing them in blueprints etc. 

I was talking to Lukas that merging the v4.3.1 changes into v5 would require adding proper languages button classes. And he thought the `LanguagesButton` class would already handle those given the name. So with that background and then adding `LanguagesAddButton` etc., the class name `LanguagesButton` felt a little out of place and non-descriptive.

config/areas/languages/buttons.php
distantnative: I guess to be more in line with the dialog name and i18n string. But could be either way.

bastianallgeier: Oh you are right. I thought the dialog was also called language.remove. Maybe we can update the button keys instead to be in line with the dialog route keys.

config/areas/site/buttons.php
bastianallgeier: What about LanguagesDropdown or LanguagesToggle instead? I agree that we can/should be more explicit, but as long as we don't use ContentTranslations in more places, I would not personally go this way.

distantnative: Went for LanguagesDropdown

config/areas/languages/buttons.php
distantnative: Renamed it and did the same then with `users.create` instead `users.add`

bastianallgeier: I just renamed the LanguageAddButton into LanguageCreateButton as well.

config/areas/site/buttons.php
bastianallgeier: Sorry, one more round 🙈 I hope I'm not too much a PITA at this point. Should it be LanguagesDropdownButton instead?

PR #6650, opened 2024-08-31
`api.methodOverwrite` config option and check

### LINKED ISSUE #6636: Config option: disable `PATCH` methodOverwrite
https://forum.getkirby.com/t/kirby-panel-not-editable-with-bunny-cdn/32215

distantnative:
### Summary of changes
- Switching default to use pure `PATCH` requests etc. without `x-http-method-override` header
- New `api.methodOverwrite` config option to activate `x-http-method-override` header
- New check in system settings that shows error if `PATCH` requests don't work

### Reasoning
https://forum.getkirby.com/t/kirby-panel-not-editable-with-bunny-cdn/32215

### Enhancements
- `api.methodOverwrite` config option to activate pure `PATCH` requests

## Docs

- Add `api.methodOverwrite` config option docs

## Ready?

### For review team

### DIFF
--- a/src/Panel/View.php
+++ b/src/Panel/View.php
@@ -275,6 +275,9 @@ class View
 
 		return [
 			'$config' => fn () => [
+				'api'         => [
+					'methodOverwrite' => $kirby->option('api.methodOverwrite', false)
+				],
 				'debug'       => $kirby->option('debug', false),
 				'kirbytext'   => $kirby->option('panel.kirbytext', true),
 				'translation' => $kirby->option('panel.language', 'en'),
--- a/src/Panel/View.php
+++ b/src/Panel/View.php
@@ -276,7 +276,7 @@ class View
 		return [
 			'$config' => fn () => [
 				'api'         => [
-					'methodOverwrite' => $kirby->option('api.methodOverwrite', false)
+					'methodOverwrite' => $kirby->option('api.methodOverwrite', true)
 				],
 				'debug'       => $kirby->option('debug', false),
 				'kirbytext'   => $kirby->option('panel.kirbytext', true),
--- a/config/api/routes/system.php
+++ b/config/api/routes/system.php
@@ -32,6 +32,18 @@ return [
 		}
 	],
 	[
+		'pattern' => 'system/method-test',
+		'method'  => 'PATCH',
+		'action'  => function () {
+			return [
+				'status' => match ($this->kirby()->request()->method()) {
+					'PATCH' => 'ok',
+					default => 'fail'
+				}
+			];
+		}
+	],
+	[
 		'pattern' => 'system/register',
 		'method'  => 'POST',
 		'action'  => function () {
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -629,6 +629,7 @@
 
 	"system.info.copy": "Copy info",
 	"system.info.copied": "System info copied",
+	"system.issues.api.methods": "Your server does not support PATCH requests",
 	"system.issues.content": "The content folder seems to be exposed",
 	"system.issues.eol.kirby": "Your installed Kirby version has reached end-of-life and will not receive further security updates",
 	"system.issues.eol.plugin": "Your installed version of the { plugin } plugin is has reached end-of-life and will not receive further security updates",
--- a/panel/src/api/index.js
+++ b/panel/src/api/index.js
@@ -23,7 +23,7 @@ export default (panel) => {
 	const api = {
 		csrf: panel.system.csrf,
 		endpoint: rtrim(panel.urls.api, "/"),
-		methodOverwrite: true,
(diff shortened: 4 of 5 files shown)


src/Panel/View.php
```diff
@@ -275,6 +275,9 @@ public static function globals(): array
 
 		return [
 			'$config' => fn () => [
+				'api'         => [
+					'methodOverwrite' => $kirby->option('api.methodOverwrite', false)
```
bastianallgeier: I remember that quite a lot of shared hosting environments disable patch requests for some reason. I think we should keep them off by default tbh.

bastianallgeier (commented):
I really like this new option and it would be great if we could get this PR done.

src/Panel/View.php
distantnative: I really don't mind either or. I changed it because @lukasbestle said the opposite of you in the forum 😄 no idea myself who is right

bastianallgeier (approved):

beromir:
Thank you @distantnative,  works great!
But shouldn't the default for the option be `false`? In my case the `method-test` in the system view returned a 404 error, because it was using the method override header. After disabling it, it worked.

distantnative:
@beromir that would be a breaking change for all setups that don't support pure PATCH requests etc. which we could not introduce in 4.5. This is why you actively have to set the config option, if you want the new behavior.

beromir:
Yeah, makes sense. Thank you.

lukasbestle:
@distantnative @bastianallgeier Super late to the party. Can we change the default option value to `false` for v5? The check for support of the `PATCH` method doesn't really make sense if the default is not to use it anyway. Also what Henrik wrote: If the default is `true`, the check won't even go through on servers that are affected because the method is already overridden.

Maybe we could also use the chance to rename the option to `methodOverride` (`overwrite` was surely a typo when we originally introduced it in the Panel code).

distantnative:
@lukasbestle we switched it because of https://github.com/getkirby/kirby/pull/6650#discussion_r1782400245 - are you sure that different defaults won't harm more setups in the end?

lukasbestle:
I read that, yes. I am not sure at all, but I think we should try it. If we then see that there are really too many providers that still block it, we can revert it but then also throw out the check that was added in this PR and serves no purpose with the `true` default.

Being standards-compliant whenever possible is worth a lot. Our hack with the override is not and this can break assumptions of standards-compliant proxies or reverse proxies.

@bastianallgeier Do you know of specific providers that are affected by this? Or was it just a feeling?

distantnative:
But wouldn't it be a breaking change too? I think we could only set a different default in v5 (and maybe that was actually the idea). And in v4 the check is only there for if someone sets the config to false but the server isn't supporting it to warn them.

bastianallgeier:
I don't know the specific providers. I just remember that it happened multiple times and the main problem was that it was super hard to debug. The requests just didn't go through and it took a while to figure out what was going on. When we switch it back on by default, we should probably already be prepared and know exactly how it looks like in the Panel when it happens. Otherwise we might be having a hard time again with debugging.

lukasbestle:
@distantnative Yes, it's a breaking change. That's why I think we should change the default in v5.

@bastianallgeier To make debugging easier, this PR introduced an automatic check and warning. This should cover it IMO.

bastianallgeier:
I found out in the meantime, that Mode Security with the OWASP CRS seems to be the cause why those requests are sometimes blocked.

lukasbestle:
> I found out in the meantime, that Mode Security with the OWASP CRS seems to be the cause why those requests are sometimes blocked.

Ah, right. Good point. I looked into ModSecurity and the OWASP CRS. It looks like the default setup of CRS is to [block everything except `GET HEAD POST OPTIONS`](https://github.com/coreruleset/coreruleset/blob/e54c4632de56d34cb9ce56a4d92fb488839b3ee2/crs-setup.conf.example#L457). This makes sense from the perspective of a general-use WAF. It advises to customize this for the application and there are also default "exclusion packages" that provide optimized default for different web apps and systems, but those again need to be installed manually as CRS plugins. Not something we can expect in shared hosting.

Unfortunately this matches with my previous experience with WAFs. It's great that they exist and make it simple to protect against a lot of different attacks, but there are always false positives. Be it that the methods are blocked or that the circumvention is blocked (like @beromir experienced it). This is why we can't have nice things. 😫

Looks like it is indeed safer to keep the default at `true` (= set the override header by default). But then I think we can throw out the automated check in the system view. Running it in its current version only makes sense if the option is set to `false`. But if a user does that manually, they know they need it and don't need the automated check. This was only intended to uncover issues with a default of `false` (= when a WAF like ModSecurity blocks native REST methods like `PUT`, `PATCH` and `DELETE`).

Also I still think the option should be named `api.methodOverride`. v5 could be a last chance to rename this with a minor breaking change until the name is in the core for too long.

PR #6653, opened 2024-09-07
Fix `$user->roles($purpose)`

distantnative:
### Summary of changes
-  `User::roles()` has a new optional `$purpose` parameter to pass a user action (`create` or `change`). If one is passed, the roles collection is filtered via `$roles->canBeChanged()`/`$roles->canBeCreated()` which checks if the current user has the permissions to apply the action to each role.
- `::roles()` doesn't return only the current role for non-admin users but all available roles

### Reasoning
Available roles depend on the context as they depend on whether the current user has permissions to e.g. create a user with a certain role or change the role to a certain role.

### Additional context
We need to pass the `$purpose` parameters, e.g. in user dialogs, in a subsequent PR.

## Ready?

### DIFF
--- a/config/api/routes/users.php
+++ b/config/api/routes/users.php
@@ -203,7 +203,9 @@ return [
 			'users/(:any)/roles',
 		],
 		'action'  => function (string $id) {
-			return $this->user($id)->roles();
+			$kirby   = $this->kirby();
+			$purpose = $kirby->request()->get('purpose');
+			return $this->user($id)->roles($purpose);
 		}
 	],
 	[
--- a/src/Cms/User.php
+++ b/src/Cms/User.php
@@ -574,33 +574,36 @@ class User extends ModelWithContent
 	}
 
 	/**
-	 * Returns all available roles
-	 * for this user, that can be selected
-	 * by the authenticated user
+	 * Returns all available roles for this user,
+	 * that can be selected by the authenticated user
+	 *
+	 * @param string|null $purpose User action for which the roles are used (create, change)
 	 */
-	public function roles(): Roles
+	public function roles(string|null $purpose = null): Roles
 	{
 		$kirby = $this->kirby();
 		$roles = $kirby->roles();
 
-		// a collection with just the one role of the user
-		$myRole = $roles->filter('id', $this->role()->id());
-
-		// if there's an authenticated user …
-		// admin users can select pretty much any role
-		if ($kirby->user()?->isAdmin() === true) {
-			// except if the user is the last admin
-			if ($this->isLastAdmin() === true) {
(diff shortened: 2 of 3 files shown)


config/api/routes/users.php
```diff
@@ -203,7 +203,9 @@ function ($source, $filename) use ($id) {
 			'users/(:any)/roles',
 		],
 		'action'  => function (string $id) {
-			return $this->user($id)->roles();
+			$kirby   = $this->kirby();
+			$purpose = $kirby->request()->get('purpose');
+			return $this->user($id)->roles($purpose);
```
afbora, line 208: I would prefer such a simple usage instead of making the `User::roles()` method more complicated.

```suggestion
            $roles = $this->user($id)->roles();

            return match ($purpose) {
			    'create' => $roles->canBeCreated(),
			    'change' => $roles->canBeChanged(),
			    default  => $roles
			};
```

bastianallgeier: I think I agree with @afbora. Alternatively, I would probably vote to have two methods with a single purpose each, instead of the additional argument.

distantnative: But we will need this in a couple of more places where we have so far forgotten to apply the filters. I think we should do it in one central place otherwise the danger is that we keep forgetting it when we change something

afbora: Understood. Then it makes more sense to apply Bastian's idea of ​​a secondary separated method, for example: 

````php
Roles::canBe($action)
````

distantnative: But we need it on the user object, not the roles object.

afbora: Sorry, my mistake. `User::canBe()` or something like that.

bastianallgeier: I was really thinking about two individual methods. Something like User::canCreateRoles and User::canChangeRoles but after trying to find some really good method names, I'm wondering if we are overthinking it and the argument really is the more practical option. Naming is hard :)

distantnative: I think the argument is for me the one to not as easily forget about. If I know `->roles()`, I will spot the optional argument via the IDE when using it. Remembering that there are also `->canCreateRoles()` etc. might be harder.

distantnative: If you feel very strongly about not touching `->roles()`, I think the alternative method should be named very closely, e.g. `->rolesByPurpose($purpose)`, `->rolesFor($purpose)`, `->rolesTo($action)` ...

bastianallgeier (approved):
I've come to the point where I really see the argument as a valid solution and would not overthink this anymore tbh.

PR #6659, opened 2024-09-08
Refactor `LanguageRules`

distantnative:
### Summary of changes
- Moved logic from language methods to `LanguageRules` methods
- Marked all language rules methods void
- Added tests for all cases for `LanguageRules` class
- Switched language hooks to apply
- Apply language rules before and after applying before hook

### Reasoning
Trying to create consistency for all `*Rules` classes

## Ready?

- [x] In-code documentation (wherever needed)
- [x] Unit tests for fixed bug/feature
- [x] Tests and CI checks all pass

PR #6660, opened 2024-09-08
Mark all `*Rules` class methods void

distantnative:
### Summary of changes
- Switching all rules class methods to void

### Reasoning
- Consistency
- As we work with throwing exceptions, I think the main factor to check for is whether an exception is thrown to not. And not checking for a `true` returned.

### Additional context
Language rules are covered in https://github.com/getkirby/kirby/pull/6659

## Ready?

### DIFF
--- a/src/Cms/FileRules.php
+++ b/src/Cms/FileRules.php
@@ -27,7 +27,7 @@ class FileRules
 	 * @throws \Kirby\Exception\DuplicateException If a file with this name exists
 	 * @throws \Kirby\Exception\PermissionException If the user is not allowed to rename the file
 	 */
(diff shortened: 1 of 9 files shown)


afbora:
I think this PR includes a breaking change since `*Rules` methods are `public static`. I don't think there is such a use, but in rare cases someone may have used it. I just mentioned to update the PR description.

bastianallgeier (approved):
I agree with @afbora that we should mention it as a breaking change, even though it will probably not affect anyone.

distantnative:
You are right. That's why I actually put it for v5, but forgot when writing the notes.

PR #6675, opened 2024-09-14
Use exceptions with named arguments

distantnative:
### Summary of changes
- Using named arguments for exception constructor
- Convention is: 
  - Only message string, pass directly without `message: `
  - All others as named arguments

### Reasoning
- Move to more modern code style

## Ready?

### DIFF
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -234,7 +234,9 @@ class Api
 		array|BaseCollection|null $collection = null
 	): Collection {
 		if (isset($this->collections[$name]) === false) {
-			throw new NotFoundException(sprintf('The collection "%s" does not exist', $name));
+			throw new NotFoundException(
+				message: sprintf('The collection "%s" does not exist', $name)
+			);
 		}
 
 		return new Collection($this, $collection, $this->collections[$name]);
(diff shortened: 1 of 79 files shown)


src/Api/Api.php
```diff
@@ -234,7 +234,9 @@ public function collection(
 		array|BaseCollection|null $collection = null
 	): Collection {
 		if (isset($this->collections[$name]) === false) {
-			throw new NotFoundException(sprintf('The collection "%s" does not exist', $name));
+			throw new NotFoundException(
+				message: sprintf('The collection "%s" does not exist', $name)
```
bastianallgeier, line 238: Should we do this in all cases where we only pass a message? I've seen various refactored Exception calls with messages in previous files, where you only changed the indentation. I personally really like the more explicit version with the named argument.

bastianallgeier (approved):
I've used the named message parameter in more places and converted the line breaks in a few more palces for more consistency. I'm happy with the PR, but how do we handle the coverage changes? It's all just newly detected stuff as far as I can see.

distantnative:
@bastianallgeier Ok, as I wrote above I thought for just `message:` we would drop the named argument, but also fine with it. There are a lot more though of these still without `message:` - should we then convert all?

Code coverage: Yes, I think these are all parts that have been uncovered all along but only because we now touched them. I think it's ok.

bastianallgeier:
Sorry, I obviously didn't read the summary carefully enough. 

I think the main argument for using `message` more explicitly is the argument chaos in our exception classes. I looked at it and instantly thought that the named arguments will be a great way for us to refactor that chaos later and with the named arguments we no longer need to worry about the order of it. I don't think we really need to change it everywhere all at once, but I wanted to change it in the files that already had changes. I think we just have to define a clear path ahead.

PR #6677, opened 2024-09-16
Changes 23: Simplify Draft Handling

bastianallgeier:
### Summary of changes

- Draft pages are no longer treated differently than other pages and can have their own _changes version
- `VersionId::published()` is now also used as default version for draft pages
- Removed `VersionId::default()`
- New `VersionId::all()` method to get an array with all available Version ids. 

### Reasoning

The reason to use `changes` as the default version for drafts seemed to make a lot of sense at first. But when implementing the panel integration, it turns out that this is adding a lot of unnecessary complexity. 

With our old approach of storing changes in local storage, we basically already had unsaved changes for drafts and it made a lot of sense in the interface and never felt weird. We can think the same way about how changes for drafts are handled. A draft is a page that has not been published yet. That unpublished page can still have two versions, or even multiple revisions later. 

By removing that distinction, we can avoid complexity in the backend and in the frontend. 

## Ready?

### DIFF
--- a/src/Content/ContentStorageHandler.php
+++ b/src/Content/ContentStorageHandler.php
@@ -38,7 +38,7 @@ abstract class ContentStorageHandler
 	public function all(): Generator
 	{
 		foreach (Languages::ensure() as $language) {
-			foreach ([VersionId::published(), VersionId::changes()] as $versionId) {
+			foreach (VersionId::list() as $versionId) {
 				if ($this->exists($versionId, $language) === true) {
 					yield $versionId => $language;
 				}
@@ -68,7 +68,7 @@ abstract class ContentStorageHandler
 	 */
 	public function deleteLanguage(Language $language): void
 	{
-		foreach ([VersionId::published(), VersionId::changes()] as $versionId) {
+		foreach (VersionId::list() as $versionId) {
 			$this->delete($versionId, $language);
 		}
 	}
@@ -156,7 +156,7 @@ abstract class ContentStorageHandler
 	 */
 	public function moveLanguage(Language $fromLanguage, Language $toLanguage): void
 	{
-		foreach ([VersionId::published(), VersionId::changes()] as $versionId) {
+		foreach (VersionId::list() as $versionId) {
 			if ($this->exists($versionId, $fromLanguage) === true) {
 				$this->move($versionId, $fromLanguage, $versionId, $toLanguage);
 			}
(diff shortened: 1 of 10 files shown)


src/Content/ContentStorageHandler.php
```diff
@@ -38,7 +37,7 @@ public function __construct(protected ModelWithContent $model)
 	public function all(): Generator
 	{
 		foreach (Languages::ensure() as $language) {
-			foreach ($this->dynamicVersions() as $versionId) {
+			foreach ([VersionId::published(), VersionId::changes()] as $versionId) {
```
distantnative: `Suggestion`: I think we should still keep this DRY and have something like `VersionId::defaults()`, `ContentStorageHandler::defaultVersions()` instead.

bastianallgeier: I went for `VersionId::list()` Not 100% convinced. But maybe we should just mark it as internal and not worry too much about it. We will need something else sooner or later when we will have revisions anyway.

src/Content/VersionId.php
```diff
@@ -68,18 +66,14 @@ public static function changes(): static
 	}
 
 	/**
-	 * Returns the default version id for the model
+	 * List of available version ids
 	 */
-	public static function default(ModelWithContent $model): static
+	public static function list(): array
```
distantnative: `Suggestion:`

```suggestion
	public static function all(): array
```

src/Content/VersionId.php
```diff
@@ -68,18 +66,14 @@ public static function changes(): static
...
-		if (
-			$model instanceof Page === true &&
-			$model->isDraft() === true
-		) {
-			return VersionId::changes();
-		}
-
-		return VersionId::published();
+		return [
+			static::published(),
```
distantnative, line 66: `Thought:` I'm totally on board with the general change of this PR. But now `VersionId::published()` feels not right semantically for drafts as well - as their saved state surely isn't published.

bastianallgeier: You are right. Maybe the best would be `VersionId::latest()` This would also still be pretty useful as soon as we have multiple versions. But I would love to not handle that in this PR to be honest. It would get quite massive.

src/Content/VersionId.php
bastianallgeier: All is better!

src/Content/VersionId.php
distantnative: I'm ok with solving this later. Can we add it to the Todo list?

distantnative (approved):

src/Content/VersionId.php
bastianallgeier: I've just added a todo in Notion.

PR #6680, opened 2024-09-17
Changes 25: New Version::diff method

bastianallgeier:
### Reasoning

- `Version::diff` will be used later to check if a changed version really has any differences. This will help to avoid tracking unnecessary changes. 

### Enhancements

- New `Version::diff` method to compare two versions and get an array of changed fields

## Ready?

### For review team

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -105,6 +105,23 @@ class Version
 	}
 
 	/**
+	 * Returns the changed fields, compared to the given version
+	 */
+	public function diff(VersionId|string $versionId, Language|string $language = 'default'): array
+	{
+		$versionId = VersionId::from($versionId);
+
+		if ($versionId->is($this->id) === true) {
+			return [];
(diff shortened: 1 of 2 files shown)


src/Content/Version.php
```diff
@@ -104,6 +104,23 @@ public function delete(): void
...
 
+	/**
+	 * Returns the changed fields, compared to the given version
+	 */
+	public function diff(VersionId|string $versionId, Language|string $language = 'default'): array
+	{
+		$versionId = VersionId::from($versionId);
+
+		if ($versionId->is($this->id) === true) {
+			return [];
```
distantnative, line 115: Unit test missing for this, but other than that I think it makes sense to have that method and implementation looks good.

bastianallgeier:
@distantnative just added a unit test to cover that if clause

distantnative (approved):

PR #6681, opened 2024-09-17
Changes 26: New Changes class

bastianallgeier:
### Summary of changes

- New `Kirby\Content\Changes` class to track changed models in the `site.txt`
- The Changes class is used in the Changes controller to track and untrack changed versions.

### Enhancements

- New `Kirby\Content\Changes` class to track changed models in the `site.txt`

## Ready?

### For review team

### DIFF
--- a/src/Content/Changes.php
+++ b/src/Content/Changes.php
@@ -1,0 +1,110 @@
+<?php
+
+namespace Kirby\Content;
+
+use Kirby\Cms\App;
+use Kirby\Cms\Files;
+use Kirby\Cms\ModelWithContent;
+use Kirby\Cms\Pages;
+use Kirby\Cms\Site;
+use Kirby\Cms\Users;
+
+/**
+ * The Changes class tracks changed models
+ * in the Site's changes field.
+ *
+ * @package   Kirby Content
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ */
+class Changes
+{
+	/**
+	 * Access helper for the field, in which changes are stored
+	 */
+	public function field(): Field
+	{
+		return $this
+			->site()
+			->version(VersionId::published())
+			->content()
+			->get('changes');
+	}
+
+	/**
+	 * Return all files with unsaved changes
+	 */
+	public function files(): Files
+	{
+		return $this->field()->toFiles();
+	}
+
+	/**
+	 * Return all pages with unsaved changes
+	 */
+	public function pages(): Pages
+	{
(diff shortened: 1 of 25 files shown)


src/Content/Changes.php
```diff
@@ -0,0 +1,110 @@
...
+	/**
+	 * Access helper for the field, in which changes are stored
+	 */
+	public function field(): Field
+	{
+		return $this
+			->site()
+			->version(VersionId::published())
+			->content()
+			->get('changes');
```
distantnative, line 33: `Suggestion:` To avoid more protected filenames, how about using `_changes` or `__changes`?

src/Content/Changes.php
```diff
@@ -0,0 +1,110 @@
...
+	 */
+	public function untrack(ModelWithContent $model): void
+	{
+		$changes = $this->field()->yaml();
+		$uuid    = (string)$model->uuid();
+		$index   = array_search($uuid, $changes);
+
+		if ($index !== false) {
+			unset($changes[$index]);
+		}
```
distantnative, line 82:
```suggestion
		$changes = A::filter(
			$this->field()->yaml(),
			fn ($uuid) => $uuid !== (string)$model->uuid()
		);
```

Maybe a little easier? Untested, just an idea.

src/Content/Changes.php
```diff
@@ -0,0 +1,110 @@
...
+	public function update(array $changes): void
+	{
+		$changes = array_unique($changes);
+		$changes = array_values($changes);
+
+		$this
+			->site()
+			->version(VersionId::published())
+			->update([
+				'changes' => $changes
```
distantnative, line 99: See above re: field name.

distantnative (commented):
`Thought:` What happens in this scenario, if I make changes to a page, it autosaves, adds page to be tracked as unsaved change. Then I change the field value back to the original, autosaves... should it still be in unsaved changes (as I think it would be with this PR)?

src/Content/Changes.php
bastianallgeier: I'm really torn with the underscore fieldnames. Same for the _lock field. The `Site::changes()` method is already kind of protecting the field by avoiding direct access. On the other hand, it's definitely more likely that a changes field is already used. It also feels a bit like we should have made that decision in other places too and used it for sort or template or other such system fields. I guess I'm fine with both.

bastianallgeier:
@distantnative about autosaving and then undoing that change. This will be really hard to track. I built the Version::diff method partly for that but if there's a race like this, we can't really say anymore what was the original version that we should compare to. I guess we can really only find a solution to this by experimenting around with such scenarios in the panel and playing with various options to avoid such conflicts.

PR #6689, opened 2024-09-19
Changes 24: New Content Flow

bastianallgeier:
## Description

**I messed up the previous PR. This is replicating https://github.com/getkirby/kirby/pull/6678. See https://github.com/getkirby/kirby/pull/6678#issuecomment-2354688899 for a more detailed explanation why I think this is a good and important step**

### Summary of changes

- `content` is passed as prop to the `k-sections` component, which will then pass it on to all child section components. 
- `input` and `submit` events are bubbling up from the fields section -> to the sections component -> to the model view components. 
- The `k-form-button` component is simplified and no longer handles `submit` and `discard` events itself. It sends the events up to the model view components to be handled there.
- The file preview components now also receive the content as property and can send input and submit events up to the model view. 
- The `ModelView` component class takes over handling submit, input and discard events. It also listens to the `view.save` event and handles that. With this change, each view component is the single place of truth to pass content down to child components and handle input, submit and discard events. This drastically reduces the mental load to parse which component is responsible for what and the child components can be a lot more simple. 

### Reasoning

Simplifying the content flow will help a lot with the next steps when changes are coming from the backend. We can pass them as view property and the view can then pass it further down. This will avoid having to run some weird content module init code like we currently do in the Vuex module. It will also no longer tightly couple module logic with the components. 

### Next steps

- The `k-form-buttons` component is about to be removed and replaced by the new `k-form-controls` component. I wanted to refactor it in this PR anyway, to already get prepared for this step. The form controls component will also send the events up to the model view so it makes sense to already do that and reduce the dependency on the old form buttons component. 
- The content module will likely need a few more changes. I've prepared a couple things on the next-steps branch, but didn't want to blow this PR up even more.

### DIFF
--- a/panel/src/components/Views/Files/FileView.vue
+++ b/panel/src/components/Views/Files/FileView.vue
@@ -19,7 +19,7 @@
 
 			<template #buttons>
 				<k-view-buttons :buttons="buttons" @action="onAction" />
-				<k-form-buttons />
+				<k-form-buttons @discard="onDiscard" @submit="onSubmit" />
 			</template>
 		</k-header>
 
@@ -29,10 +29,13 @@
 
 		<k-sections
 			:blueprint="blueprint"
+			:content="content"
 			:empty="$t('file.blueprint', { blueprint: $esc(blueprint) })"
 			:lock="lock"
 			:parent="id"
 			:tab="tab"
+			@input="onInput"
+			@submit="onSubmit"
 		/>
 	</k-panel-inside>
 </template>
--- a/panel/src/components/Views/Files/FileView.vue
+++ b/panel/src/components/Views/Files/FileView.vue
@@ -23,7 +23,12 @@
 			</template>
 		</k-header>
 
-		<k-file-preview v-bind="preview" />
+		<k-file-preview
+			v-bind="preview"
+			:content="content"
+			@input="onInput"
+			@submit="onSubmit"
+		/>
 
 		<k-model-tabs :tab="tab.name" :tabs="tabs" />
 
--- a/panel/src/components/Views/ModelView.vue
+++ b/panel/src/components/Views/ModelView.vue
@@ -1,4 +1,6 @@
 <script>
+import debounce from "@/helpers/debounce.js";
+
 /**
  * @internal
  */
@@ -33,6 +35,9 @@ export default {
 		}
 	},
 	computed: {
+		content() {
+			return this.$panel.content.values;
+		},
 		id() {
 			return this.model.link;
 		},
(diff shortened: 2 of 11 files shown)


panel/src/components/Views/Files/FileView.vue
```diff
@@ -19,20 +19,28 @@
...
+				<k-form-buttons @discard="onDiscard" @submit="onSubmit" />
 			</template>
 		</k-header>
 
-		<k-file-preview v-bind="preview" />
+		<k-file-preview
+			v-bind="preview"
+			:content="content"
+			@input="onInput"
+			@submit="onSubmit"
```
distantnative, line 30: `Question:` Should a preview component really submit?

panel/src/components/Views/ModelView.vue
```diff
@@ -57,16 +62,34 @@ export default {
 		}
 	},
 	mounted() {
+		this.onInput = debounce(this.onInput, 50);
```
distantnative, line 65: `Thought:` Instead of debouncing here the `onInput` method, I think we should only debounce the API request. That way we could avoid issues we currently have (e.g. the not-so-responsive range slider due to debouncing updating the values) and still debounce the API calls. Though, this might be hard to realize when the content comes only from the backend through the props (might be further delayed) and not directly updated via un-debounced input event.

This might be actually a **bigger concern**: How much delay do we add by having the updated values coming in via props from the backend? Most of our inputs don't work with a local state but rely on that the new value they emit via `input` event, immediately loops in via the `value` prop again.

bastianallgeier: Debouncing the API request is happening on the next branch in one of the next steps and it works really well. 

Working on this, I realized that we only need the debouncing here to avoid multiple JSON.stringify comparisons between original and changed fields. This has been the big performance hit. It really is all about showing or hiding the form buttons. 

So I think it might be worth it to separate the "hasChanges" check into its own method and really only debounce that one as well. 

Inputs could then fire right away. Autosaving via API requests would be debounced and the changes checks would also be debounced. By separating those into two debounced methods, we can have more control over ideal debounce intervals to get the feeling right. It seems like it's totally enough to send API requests every 200 ms or something. But we really might want to have the update checks more frequently.

panel/src/components/Views/Files/FileView.vue
bastianallgeier: It's not needed so far and we could probably not implement it yet and only add it if it's really needed. But I thought that the previews really just could act like another section and have all the capabilities to interact with content if needed. So this change was more about consistency.

panel/src/components/Views/ModelView.vue
distantnative: Sounds good. 

How do we avoid the delay for the updated values coming in via the props though?

bastianallgeier: I think we really can't. This is a problem that we haven't solved at all so far and will be a new "feature" anyway. In the next branch, I'm instantly overwriting the view props on input and before submit. When the API requests finish and the backend returns different values, we could probably compare the differences first and do nothing if there are no changes. If there are changes, we might actually think about triggering a dialog or a little notification or something, but we really need to try this out.

distantnative (approved):

PR #6690, opened 2024-09-19
Changes 25: New Changes Class

bastianallgeier:
## Description

**This is replicating https://github.com/getkirby/kirby/pull/6681**

### Summary of changes

- New `Kirby\Content\Changes` class to track changed models in the `site.txt`
- The Changes class is used in the Changes controller to track and untrack changed versions.

### Enhancements

- New `Kirby\Content\Changes` class to track changed models in the `site.txt`

## Ready?

### For review team

### DIFF
--- a/src/Content/Changes.php
+++ b/src/Content/Changes.php
@@ -8,6 +8,7 @@ use Kirby\Cms\ModelWithContent;
 use Kirby\Cms\Pages;
 use Kirby\Cms\Site;
 use Kirby\Cms\Users;
+use Kirby\Toolkit\A;
 
 /**
  * The Changes class tracks changed models
@@ -73,13 +74,10 @@ class Changes
 	 */
 	public function untrack(ModelWithContent $model): void
 	{
-		$changes = $this->field()->yaml();
-		$uuid    = (string)$model->uuid();
-		$index   = array_search($uuid, $changes);
-
-		if ($index !== false) {
-			unset($changes[$index]);
-		}
+		$changes = A::filter(
+			$this->field()->yaml(),
+			fn ($uuid) => $uuid !== (string)$model->uuid()
+		);
 
 		$this->update($changes);
 	}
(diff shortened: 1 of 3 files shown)


distantnative:
Another thought: we could also stores this in a dedicated file in the `site/log/` root instead of the site content file.

bastianallgeier:
I thought about this as well. Storing it in the site.txt really isn't that great after all. I was also thinking about a new /site/accounts/users.txt 

I like site/log, but my only concern is that it would be a new directory that needs write permissions.

distantnative:
I do like `/site/accounts/users.txt` - could be an issue if the directory is excluded from uploads/git... etc. though?

bastianallgeier:
That's basically why I still went for the site.txt after all. Each alternative solution has potential pitfalls that we need to think through. That's why I thought I really want to get the functionality done first and we can still change the location later. 

I know that there are a lot of "let's think about this later" points at the moment, but I honestly feel like we won't ever get this done if we don't approach it more pragmatically right now.

distantnative:
Ok, let's leave it with the `site.txt`.

src/Content/Changes.php
```diff
@@ -0,0 +1,110 @@
...
+	 */
+	public function untrack(ModelWithContent $model): void
+	{
+		$changes = $this->field()->yaml();
+		$uuid    = (string)$model->uuid();
+		$index   = array_search($uuid, $changes);
+
+		if ($index !== false) {
+			unset($changes[$index]);
+		}
```
distantnative: What do you think https://github.com/getkirby/kirby/pull/6681#discussion_r1766708369 ?

bastianallgeier: ✅

bastianallgeier:
The test coverage report is again completely wrong. I have no idea why this is failing so often to even recognize the @covers annotations. So annoying.

distantnative (approved):
Me neither. There are factual unit tests covering the methods marked as uncovered.

PR #6694, opened 2024-09-22
`UserRules::create()` validate roles

distantnative:
## Description

- [x] https://github.com/getkirby/kirby/pull/6693
- [x] https://github.com/getkirby/kirby/pull/6658

### Summary of changes
- `UserRules::create()` also validates if role is a valid option

### Reasoning
Consistency and the rules class should double check this as well.

## Ready?

### DIFF
--- a/src/Cms/UserRules.php
+++ b/src/Cms/UserRules.php
@@ -198,10 +198,22 @@ class UserRules
 		// check user permissions (if not on install)
 		if (
 			$user->kirby()->users()->count() > 0 &&
-			$user->permissions()->can('create') !== true
+			$user->permissions()->create() !== true
 		) {
 			throw new PermissionException([
 				'key' => 'user.create.permission'
+			]);
+		}
+
+		$role = $props['role'] ?? null;
+
+		// prevent creating a role that is not available for user
+		if (
+			in_array($role, [null, 'default', 'nobody'], true) === false &&
+			$user->kirby()->roles('create')->find($role) instanceof Role === false
+		) {
+			throw new InvalidArgumentException([
+				'key' => 'user.role.invalid',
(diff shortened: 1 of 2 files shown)


src/Cms/UserRules.php
```diff
@@ -198,13 +198,25 @@ public static function create(User $user, array $props = []): bool
...
 			throw new PermissionException([
 				'key' => 'user.create.permission'
 			]);
 		}
 
+		$role = $props['role'] ?? null;
+
+		// prevent creating a role that is not available for user
+		if (
+			in_array($role, [null, 'default', 'nobody'], true) === false &&
```
bastianallgeier, line 212: I'm not sure I fully understand the first line. It's allow-listing some roles here. But do we really want to make them creatable like this at all? Or am I reading it wrong?

distantnative: Answering for each case separately:
- `null`: Feels like the stupid one, but so far we haven't enforced any user role in `UserRules::create`, so it would be a breaking change if we do now.
- `default`: This will be the value when using `User::create()` without passing any role, but it won't be included in the roles collection to validate against. That's why we're skipping the check here. Not skipping will cause at least `SystemTest::testIsInstalled()` to fail.
- `nobody` isn't part of any unit test, but serves as a valid role inside Kirby. That's why I wanted to cover it here as well. Although probably creating `nobody` isn't a real use case. We would need to add additional logic then, as not skipping the collection lookup will fail otherwise.

bastianallgeier: Ok, now I get it. Makes sense! Thanks for the explanation :)

bastianallgeier (approved):

PR #6701, opened 2024-09-24
Better content storage `::move()` + new `::copy()`

distantnative:
### Summary of changes
- `ContentStorageHandler::move()` treats `$toVersionId` and `$toLanguage` optionally and falls back to the same version Id/language as `$from` (avoids having to repeat arguments and allows for better usage with named arguments)
- `ContentStorageHandler::move()` allows specifying a `$toStorage` to move content to another storage 
- New `ContentStorageHandler::copy()` analogous to `::move()`
- New `ContentStorageHandler::moveAll(to: $storage)` and `ContentStorageHandler::copyAll(to: $storage)`
- Removed custom `PlainTextContentStorageHandler::move()`

### Reasoning
- Allows us to move storage logic out of the `FileActions` class into the storage handler classes
- Allows to keep those method class DRY (e.g. when moving between the same version ID, where only the language changes)
- While the `PlainTextContentStorageHandler ` implementation might have had a performance advantage, I think it's better to reuse the main implementation to keep the code complexity smaller.

## Ready?

### DIFF
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -85,11 +85,7 @@ trait FileActions
 			F::move($oldFile->root(), $newFile->root());
 
 			// move the content storage versions
-			foreach ($oldFile->storage()->all() as $version => $lang) {
-				$content = $oldFile->storage()->read($version, $lang);
-				$oldFile->storage()->delete($version, $lang);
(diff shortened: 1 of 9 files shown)


bastianallgeier (approved):
I like this a lot!

PR #6703, opened 2024-09-25
Changes: New Changes Dialog

bastianallgeier:
### Summary of changes

Screenshot 2024-09-25 at 12 53 27

- The `k-changes-dialog` component is now refactored to receive a list of changed pages, files and accounts directly from the backend. This simplifies the component drastically. 
- The `/panel/dialogs/changes` route now only has a load handler. The submit handler is no longer needed
- The `Kirby\Panel\ChangesDialog` class can be a lot simpler as well and now pulls changes directly via the new `Kirby\Content\Changes` class. 
- The en.json has new strings for the dialog. 

### Temporarily broken

- The changes in local storage are completely ignored and no longer shown in the dialog. 

### How to test the new dialog

As long as changes are not yet stored and tracked in the panel, you have to emulate changed pages, files and accounts by adding a list of UUIDs to the `Changes` field in the site.txt

### Additional ideas

- We could extend the dialog to save or discard all current changes
- Each item in the dialog could contain the last edited time and editor.

### DIFF
--- a/config/areas/site/dialogs.php
+++ b/config/areas/site/dialogs.php
@@ -641,13 +641,7 @@ return [
 	'changes' => [
 		'pattern' => 'changes',
 		'load'    => function () {
-			$dialog = new ChangesDialog();
-			return $dialog->load();
+			return (new ChangesDialog())->load();
 		},
-		'submit' => function () {
-			$dialog = new ChangesDialog();
-			$ids    = App::instance()->request()->get('ids');
-			return $dialog->submit($ids);
-		}
 	],
 ];
--- a/src/Panel/ChangesDialog.php
+++ b/src/Panel/ChangesDialog.php
@@ -2,70 +2,53 @@
 
 namespace Kirby\Panel;
 
-use Kirby\Cms\App;
-use Kirby\Cms\Find;
-use Kirby\Http\Uri;
-use Kirby\Toolkit\Escape;
-use Throwable;
+use Kirby\Cms\Collection;
+use Kirby\Content\Changes;
 
 class ChangesDialog
 {
-	public function changes(array $ids = []): array
+	protected Changes $changes;
+
+	public function __construct()
 	{
-		$kirby     = App::instance();
-		$multilang = $kirby->multilang();
-		$changes   = [];
+		$this->changes = new Changes();
+	}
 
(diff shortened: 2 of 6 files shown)


config/areas/site/dialogs.php
```diff
@@ -641,13 +641,7 @@
 	'changes' => [
 		'pattern' => 'changes',
 		'load'    => function () {
-			$dialog = new ChangesDialog();
-			return $dialog->load();
+			return (new ChangesDialog())->load();
```
distantnative, line 644: `Comment:` Just a future idea to share, not at all related to this PR: I think it would be cool if at one point we have standardized dialog classes and then it would be only 

```php
'changes' => ChangesDialog::class
```

or so - and our Fiber dialog classes would in that case take care of calling the standardized `::pattern()`, `->load()`, `->submit()` etc.

src/Panel/ChangesDialog.php
```diff
@@ -2,70 +2,53 @@
...
-			} catch (Throwable) {
-				continue;
-			}
+		foreach ($models as $model) {
+			$items[] = $model->panel()->dropdownOption();
 		}
 
-		return $changes;
+		return $items;
 	}
```
distantnative, line 44:
```php
public function items(Collection $models): array
{
	return $models->values(
		fn ($model) => $model->panel()->dropdownOption()
	);
}	
```

Though maybe it doesn't need the `::items()` method even, if we can do simply the following?


```php
public function pages(): array
{
	return $this->changes->pages()->values(
		fn ($page) => $page->panel()->dropdownOption()
	);
}	
```

config/areas/site/dialogs.php
bastianallgeier: I love the idea!

src/Panel/ChangesDialog.php
bastianallgeier: This would be totally enough for now. But we would still need the items method later if we want to add the feature ideas from above. We could go for the simpler version first though. What do you prefer?

distantnative: Fine with me, then let's keep the `::items()`method but with `->values()`?

distantnative (approved):

PR #6704, opened 2024-09-25
Changes: Content Flow

bastianallgeier:
## Merge first

- [x] https://github.com/getkirby/kirby/pull/6702

## Known problems

- [x] `Invalid prop: type check failed for prop "content". Expected Object, got Array ` on AccountView (probably when `content` etc. are empty as then PHP empty array !== empty object in JS anymore)

### Refactored `panel.content` module
- removed `content.hasUnsavedChanges`
- removed `content.hasUnublishedChanges`
- removed `content.isDraft`
- removed `content.isPublishing`
- removed `content.isSaving`
- simplified `content.isLocked` and temporarliy removed `content.lock` and `content.unlock` for now to start the lock concept from scratch as soon as it exists in the backend.
- Changed `content.set` to `content.update`

### Refactored Panel Model classes
- New `Model::originals()` method 
- Refactored `::props` methods

### Refactored all Model Views 
- The model child object is gone and required props are directly registered for each model view
- `k-user-profile` and `k-user-avatar` have been refactored to register props directly instead of working with a model object property

## Questions 

Is the View refactoring too much? We could do it without it, but I think it's a great chance to clean up the model views and make them less smart.

### DIFF
--- a/panel/src/components/Views/ModelView.vue
+++ b/panel/src/components/Views/ModelView.vue
@@ -6,18 +6,19 @@ import debounce from "@/helpers/debounce.js";
  */
 export default {
 	props: {
+		api: String,
 		blueprint: String,
 		buttons: Array,
-		next: Object,
-		prev: Object,
-		permissions: {
-			type: Object,
-			default: () => ({})
-		},
+		content: Object,
+		id: String,
+		link: String,
 		lock: {
 			type: [Boolean, Object]
 		},
-		model: {
+		next: Object,
+		originals: Object,
+		prev: Object,
+		permissions: {
 			type: Object,
 			default: () => ({})
 		},
@@ -32,38 +33,24 @@ export default {
 		tabs: {
 			type: Array,
 			default: () => []
-		}
+		},
+		uuid: String
 	},
 	computed: {
 		changes() {
 			return this.$panel.content.changes;
 		},
-		content() {
-			return this.$panel.content.values;
-		},
-		id() {
-			return this.model.link;
+		hasTabs() {
+			return this.tabs.length > 1;
 		},
 		isLocked() {
 			return false;
 		},
 		isUnsaved() {
-			return this.$helper.object.length(this.changes) > 0;
+			return this.$panel.content.hasChanges;
 		},
 		protectedFields() {
 			return [];
-		}
-	},
-	watch: {
-		"$panel.view.timestamp": {
-			handler() {
-				// this is a temporary emulation of what should be coming
-				// directly from the backend.
-				this.$panel.view.props.originals = this.model.content;
-				this.$panel.view.props.content = this.model.content;
-				this.$panel.view.props.api = this.id;
-			},
-			immediate: true
 		}
 	},
 	mounted() {
--- a/panel/src/components/Views/ModelView.vue
+++ b/panel/src/components/Views/ModelView.vue
@@ -35,6 +35,9 @@ export default {
 		}
 	},
 	computed: {
+		changes() {
+			return this.$panel.content.changes;
+		},
 		content() {
 			return this.$panel.content.values;
 		},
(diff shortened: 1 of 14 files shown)


distantnative:
@bastianallgeier re: view refactoring: just the changes in templates might break custom implementations that overwrote the view components. But I think we'll have plenty of changes there anyways, that we cannot avoid breaking them. However, this PR is also stripping quite a bunch of information (e.g. page view: status etc.) that custom implementations might have used and won't be able going forward, even when they fix the template part. 

EDIT: After a night of sleep - I would still go ahead. We can see during the beta if we want to add back a few more props.

bastianallgeier:
I had a look at the model items for each model and we have a lot of stuff in those objects that seems outdated to me or at least we don't use it ourselves. That's why I decided to remove those items. I think it would be best to have a clear plan how to move forward with backend classes for views that make it easy to extend those props if you build a custom view implementation. Right now we have a weird assumption-based solution where we offer some of the model info but not all of it. So some plugins might be happy with what we have in the model object and for some they still would need a custom solution to get more props. 

To provide a more elegant upgrade path, we could also think about providing the same model object for now but deprecate it.

bastianallgeier:
I've reverted all the breaking model object removals. Some props are now redundant and also available on the top-level but I think that's fine for now. I've marked the model prop as deprecated in the backend classes.

panel/src/panel/content.js
```diff
@@ -1,130 +1,151 @@
...
+			const changes = {};
+
+			for (const field in panel.view.props.content) {
+				const changed = JSON.stringify(panel.view.props.content[field]);
+				const original = JSON.stringify(panel.view.props.originals[field]);
+
+				if (changed !== original) {
+					changes[field] = panel.view.props.content[field];
+				}
+			}
```
distantnative, line 32: Wouldn't this be right too?

```suggestion
			for (const field in this.values) {
				const changed = JSON.stringify(this.values[field]);
				const original = JSON.stringify(this.originals[field]);

				if (changed !== original) {
					changes[field] = this.values[field];
				}
			}
```

panel/src/panel/content.js
```diff
@@ -1,130 +1,151 @@
...
 		 * Returns all fields and values incl. changes
 		 *
 		 * @returns {Object}
 		 */
 		get values() {
-			return {
-				...this.published,
-				...this.changes
-			};
+			return panel.view.props.content;
```
distantnative, line 148: How about adding also 

```
set values(values) {
  panel.view.props.content = values
}
```

So that we can stay more inside the module logic with less references directly to view props?

panel/src/panel/content.js
bastianallgeier: I don't fully trust the getters in combination with reactivity tbh.

panel/src/panel/content.js
bastianallgeier: Same as above. If we test it and it properly works, I'm fine with it. But I had issues with getters and setters before. Unfortunately, not in a way that I could fully reproduce the issues or that I know what's really happening.

distantnative: I get what you mean, let's keep it, not an important change request.

panel/src/components/Views/ModelView.vue
```diff
@@ -32,40 +34,28 @@ export default {
...
-					content: this.model.content,
-					ignore: this.protectedFields
-				});
-			},
-			immediate: true
-		}
-	},
 	mounted() {
-		this.onInput = debounce(this.onInput, 50);
+		this.autosave = debounce(this.autosave, 200);
```
distantnative: `Thought:` Is pure debouncing the best way? If I keep continuously writing to a field, could be that there is no save for minutes, right? Do we need a "debounce but ensure at least every x seconds"?

bastianallgeier: ~~I think we need to tackle this one separately. We would probably need to start a timeout on every autosave call and then trigger another call a 1 minute later or so. This way, even if you don't type anything, you would still keep on firing new autosave events at least every minute.~~

~~Or we really just keep the regular calls, like we did for locking so far. They should still be debounced together with the ones from the input event when we use the same autosave method.~~

Wait, I think I'm thinking the wrong way here. The nice part about our current solution is, that it gives a realistic "last update" timestamp for the changes. We use this in one of the next PRs for the form controls dropdown when the page is locked. We would totally lose that if we auto save with fixed intervals.

PR #6706, opened 2024-09-27
Changes: Locking

bastianallgeier:
### Merge first 

- [x] https://github.com/getkirby/kirby/pull/6705

### Summary of changes

- Keep track of the editing user in the `Version` class.
- New `Content\Lock` class to eventually replace `Cms\ContentLock`
- Use the new Lock class to return an updated lock prop for all panel model views.
- Pass new lock info to form controls.

### Additional context

The best way to test the locking state manually … 

1. create a second user
2. make some changes to a page without saving
3. go to the _changes folder and the content text file and change the lock id to second user's id
4. Reload the panel view to see the lock state. 

## Ready?

### For review team

### DIFF
--- a/panel/src/libraries/dayjs.js
+++ b/panel/src/libraries/dayjs.js
@@ -1,6 +1,5 @@
 import dayjs from "dayjs";
 import customParseFormat from "dayjs/plugin/customParseFormat";
-import relativeTime from "dayjs/plugin/relativeTime";
 import interpret from "./dayjs-interpret.js";
 import iso from "./dayjs-iso.js";
 import merge from "./dayjs-merge.js";
@@ -13,7 +12,6 @@ dayjs.extend(interpret);
 dayjs.extend(iso);
 dayjs.extend(merge);
 dayjs.extend(pattern);
-dayjs.extend(relativeTime);
 dayjs.extend(round);
 dayjs.extend(validate);
 
--- a/panel/src/libraries/dayjs.js
+++ b/panel/src/libraries/dayjs.js
@@ -1,5 +1,6 @@
 import dayjs from "dayjs";
 import customParseFormat from "dayjs/plugin/customParseFormat";
+import relativeTime from "dayjs/plugin/relativeTime";
 import interpret from "./dayjs-interpret.js";
 import iso from "./dayjs-iso.js";
 import merge from "./dayjs-merge.js";
@@ -12,6 +13,7 @@ dayjs.extend(interpret);
 dayjs.extend(iso);
 dayjs.extend(merge);
 dayjs.extend(pattern);
+dayjs.extend(relativeTime);
 dayjs.extend(round);
 dayjs.extend(validate);
 
--- a/src/Content/Lock.php
+++ b/src/Content/Lock.php
@@ -3,9 +3,7 @@
 namespace Kirby\Content;
 
 use Kirby\Cms\App;
-use Kirby\Cms\ModelWithContent;
 use Kirby\Cms\User;
-use Kirby\Exception\LogicException;
 use Kirby\Toolkit\Str;
 
 /**
--- a/src/Content/Lock.php
+++ b/src/Content/Lock.php
@@ -44,15 +44,13 @@ class Lock
 			);
 		}
 
-		$user = null;
-
 		// Read the locked user id from the version
 		if ($userId = ($version->read('default')['lock'] ?? null)) {
 			$user = App::instance()->user($userId);
 		}
 
 		return new static(
-			user: $user,
+			user: $user ?? null,
 			modified: $version->modified()
 		);
 	}
--- a/src/Content/Lock.php
+++ b/src/Content/Lock.php
@@ -66,6 +66,14 @@ class Lock
 	}
 
 	/**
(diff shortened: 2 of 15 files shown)


src/Content/Lock.php
```diff
@@ -0,0 +1,125 @@
...
+		if ($version->exists() === false) {
+			// create an open lock for the current user
+			return new static(
+				user: App::instance()->user(),
+				modified: 0
+			);
+		}
+
+		// Read the locked user id from the version
+		$userId = $version->read()['lock'] ?? null;
```
distantnative:
```suggestion
		$userId = $version->read('default')['lock'] ?? null;
```

I think it would be good to make it explicit here that we want the default language. Just not sure if passing as string or actual language object.

src/Content/Version.php
```diff
@@ -214,6 +214,11 @@ protected function prepareFieldsBeforeWrite(array $fields, Language $language):
 		// make sure to store the right fields for the model
 		$fields = $this->model->contentFileData($fields, $language);
 
+		// add the editing user
+		if ($this->id->is(VersionId::changes()) === true) {
+			$fields['lock'] = $this->model->kirby()->user()?->id();
+		}
```
distantnative: `Question:` Shouldn't this only happen for the default language?

src/Content/Lock.php
bastianallgeier: I think this really depends on how we want to move forward. We could still think about locking individual languages. But I guess this way would be easiest for now.

distantnative: Once we support it, sure. But for now I would still make it explicit so it reads more explicitly without looking up what the default for the parameter is.

distantnative:
Two thoughts:
- Should we include the config option in this PR?
- Should we remove the old locking classes and method from ModelWithContent?

src/Content/Lock.php
```diff
@@ -0,0 +1,133 @@
...
+				user: App::instance()->user(),
+				modified: null
+			);
+		}
+
+		// Read the locked user id from the version
+		$userId = $version->read('default')['lock'] ?? null;
+
+		// No user? Create the lock for the currently authenticated user
+		$user = App::instance()->user($userId) ?? App::instance()->user();
```
distantnative: `Question:` I don't quite understand why we are not passing null in this case but the current user. Shouldn't there be a difference between "lock is set but user not found/deleted" and "no lock is set"?

bastianallgeier: Good catch. In my first iteration, it wasn't possible to create a lock instance without user. That's why I had that fallback. Then I changed it and made the user optional, but I didn't refactor that line. It should now make more sense.

bastianallgeier:
About your thoughts: I think I would like to handle both in a separate PR.

distantnative:
Ok but we really need to make sure to keep track of these todos. It's super easy otherwise we forget about this config option down the road.

panel/src/libraries/dayjs.js
```diff
@@ -1,5 +1,6 @@
 import dayjs from "dayjs";
 import customParseFormat from "dayjs/plugin/customParseFormat";
+import relativeTime from "dayjs/plugin/relativeTime";
```
distantnative: `Concern:` We won't have any language support for this, will we? Cause in that case we will get reports from users right away. Not sure it's worth it then.

src/Content/Lock.php
```diff
@@ -0,0 +1,134 @@
...
+	 */
+	public static function for(
+		Version $version
+	): static {
+		// if the version does not exist, it cannot be locked
+		if ($version->exists() === false) {
+			// create an open lock for the current user
+			return new static(
+				user: App::instance()->user(),
+				modified: null
```
distantnative:
```suggestion
```

panel/src/libraries/dayjs.js
bastianallgeier: You are right! Maybe let's go with a simple date and time solution first.

src/Content/Lock.php
```diff
@@ -0,0 +1,134 @@
...
+			);
+		}
+
+		$user = null;
+
```
distantnative:
```suggestion
```

src/Content/Lock.php
```diff
@@ -0,0 +1,134 @@
...
+		// Read the locked user id from the version
+		if ($userId = ($version->read('default')['lock'] ?? null)) {
+			$user = App::instance()->user($userId);
+		}
+
+		return new static(
+			user: $user,
```
distantnative:
```suggestion
			user: $user ?? null,
```

src/Content/Lock.php
distantnative: If we use `?? null` as suggested below.

distantnative (approved):

bastianallgeier:
> Ok but we really need to make sure to keep track of these todos. It's super easy otherwise we forget about this config option down the road.

I think you are right that we might loose track. I've decided to already implement the config option. I would still remove the outdated code in a second PR, because we probably need to adjust quite some unit tests.

distantnative (approved):

PR #6710, opened 2024-09-28
`Search` Panel controller class

distantnative:
### Summary of changes
- Moved Panel search code from area configs to `Search` controller class

### Fixes
- Panel search now also considers site files

## Ready?

### For review team

### DIFF
--- a/src/Panel/Controller/Search.php
+++ b/src/Panel/Controller/Search.php
@@ -19,7 +19,7 @@ use Kirby\Toolkit\Escape;
 class Search
 {
 	public static function files(
-		string $query,
+		string|null $query = null,
 		int|null $limit = null,
 		int $page = 1
 	): array {
@@ -52,7 +52,7 @@ class Search
 	}
 
 	public static function pages(
-		string $query,
+		string|null $query = null,
 		int|null $limit = null,
 		int $page = 1
 	): array {
@@ -79,7 +79,7 @@ class Search
 	}
 
 	public static function users(
-		string $query,
+		string|null $query = null,
 		int|null $limit = null,
 		int $page = 1
 	): array {
--- a/src/Panel/Controller/Search.php
+++ b/src/Panel/Controller/Search.php
@@ -1,0 +1,104 @@
+<?php
+
+namespace Kirby\Panel\Controller;
+
+use Kirby\Cms\App;
+use Kirby\Toolkit\Escape;
+
+/**
+ * The Search controller takes care of the logic
+ * for delivering Panel search results
+ *
+ * @package   Kirby Panel
+ * @author    Nico Hoffmann <nico@getkirby.com>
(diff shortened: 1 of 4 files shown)


src/Panel/Controller/Search.php
```diff
@@ -0,0 +1,104 @@
...
+ * @author    Nico Hoffmann <nico@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @internal
+ */
+class Search
+{
+	public static function files(
+		string $query,
```
afbora: Actually  `$query` argument could be nullable but not for this method. This is valid issue for all methods for this class.

src/Panel/Controller/Search.php
```diff
@@ -0,0 +1,104 @@
...
+		int|null $limit = null,
+		int $page = 1
+	): array {
+		$kirby = App::instance();
+		$files = $kirby->site()
+			->index(true)
+			->filter('isListable', true)
```
afbora, line 29:
```suggestion
```
I think this filter is not needed since we'll filter again after few lines later.

src/Panel/Controller/Search.php
```diff
@@ -0,0 +1,104 @@
...
+			->files();
+
+		// add site files which aren't considered by the index
+		$files = $files->add($kirby->site()->files());
+
+		// filter and search among those files
+		$files = $files->filter('isListable', true)->search($query);
```
afbora, line 36: Maybe we can write like that

```suggestion
		$files = $kirby->site()
			->index(true)
			->files()
			//add site files which aren't considered by the index
			->add($kirby->site()->files())
			->filter('isListable', true)
			->search($query);
```

src/Panel/Controller/Search.php
distantnative: This would be a change to how the code was before. I think it's really meant to only use files from listable pages and then filter for listable files. One filter is applied to pages, the other to the files.

afbora: Oh that's right. I missed the pages and files collection there.

bastianallgeier (approved):
It's great to see how much controllers help to clean up the code. Do you think we could already add this to 4.5?

afbora (approved):
As Bastian said, it could be part of 4.5 since you just moved area codes to standalone class.

distantnative:
Would there be any advantage of having these pure refactoring change in a 4.x release?

bastianallgeier:
I would say that the advantage is better test coverage and maybe a few less changes in v5. But I'm fine to move it to v5 as well.

bastianallgeier:
I thought about it some more. Let's keep the refactoring jobs for v5. It's not too far away and it will help us by not introducing too many changes in stable releases in the meantime.

PR #6711, opened 2024-09-30
Remove Vuex

bastianallgeier:
### Merge first

- [x] https://github.com/getkirby/kirby/pull/6706

### Summary of changes

- Remove Vuex and the content module


distantnative (approved):
Removed some more leftover code. I think the dialogs should still work without the dispatch event now, but please have a check as well.

PR #6712, opened 2024-10-01
Remove field signature

bastianallgeier:
## Description

The Field class added an md5 signature for very field by json_encoding all its props. We added this ages ago to provide unique field keys for Vue components. This is no longer needed as far as I can see and it's a potential performance bottleneck. 

### Summary of changes

- Remove signature key in `\Kirby\Form\Field::toArray()`
- Use field name as key for Vue field components in the `k-fieldset` component
- Remove the unit test for the signature 

### Refactoring

- Remove unnecessary generic signature key for each form field.  

### Breaking changes

- The signature is longer available as field property when loading fields from the panel backend. 

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Fieldset.vue
+++ b/panel/src/components/Forms/Fieldset.vue
@@ -4,7 +4,7 @@
 			<template v-for="(field, fieldName) in fields">
 				<k-column
 					v-if="$helper.field.isVisible(field, value)"
-					:key="field.signature"
+					:key="field.name"
 					:width="field.width"
 				>
 					<!-- @event input Triggered whenever any field value changes -->
--- a/panel/src/components/Forms/Fieldset.vue
+++ b/panel/src/components/Forms/Fieldset.vue
@@ -1,7 +1,7 @@
 <template>
 	<div class="k-fieldset">
 		<k-grid variant="fields">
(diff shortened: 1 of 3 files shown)


panel/src/components/Forms/Fieldset.vue
```diff
@@ -4,7 +4,7 @@
 			<template v-for="(field, fieldName) in fields">
 				<k-column
 					v-if="$helper.field.isVisible(field, value)"
-					:key="field.signature"
+					:key="field.name"
```
distantnative: `Question:` Reading this, is there a difference between `fieldName` and `field.name` here?

bastianallgeier: Nope. Could totally go with fieldName instead.

distantnative: I'd probably do it the other way around and maybe shift to `field.name` instead of the key all together. But doesn't need to be in this PR.

distantnative (approved):

panel/src/components/Forms/Fieldset.vue
bastianallgeier: I've fixed it.

Issue #6716, opened 2024-10-02
Page `move` rules: allow any template if no pages section in parent


distantnative:
Currently, the rules check all pages sections to gather what templates a parent is allowed to contain. However, if no pages section is in the blueprint (or one could also say no allowed templates are gathered), any page is rejected to be moved into this parent.

I think, it would make more sense to allow any template in this case (as the parent is not restricting its children to any specific templates).

PR #6717, opened 2024-10-02
Move page: allow parent without any pages sections

distantnative:
## Description

- [x] Unit test

### Summary of changes
- Allow any templates as children for a page that does not define any pages sections or only sections with no template restrictions

### Reasoning
It's confusing why currently no template is accepted when moving, while no template restrictions exists for the page.

### Additional context
Could be seen a breaking change but I think the current behavior is more a unexpected bug.

### Fixes
- https://github.com/getkirby/kirby/issues/6716

## Ready?

### For review team

### DIFF
--- a/src/Cms/PageRules.php
+++ b/src/Cms/PageRules.php
@@ -385,7 +385,10 @@ class PageRules
 		}
 
 		// check if the template of this page is allowed as subpage type
-		if (in_array($page->intendedTemplate()->name(), $allowed) === false) {
+		if (
+			$allowed !== [] &&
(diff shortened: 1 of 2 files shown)


bastianallgeier:
I remember that I thought that a page should only really be allowed to have children if this is defined with page sections. But I think I agree now after a while that this is too strict. If you move a page into another page that does not have any page sections, nothing bad will truly happen. The only problem might be that you are no longer able to find that page later, except by searching it.

bastianallgeier (approved):

PR #6719, opened 2024-10-03
Remove old Lock classes

distantnative:
### Summary of changes
- Removed old `ContentLocks` and `ContentLock` classes
- Moved `::lock()` method to `ModelWithContent` class and returns `Lock` object instead of array
- Removed all `lock` REST API routes
- Fixed outdated `$model->isLocked()` call
- Adapted unit tests 

## Ready?


src/Cms/ModelWithContent.php
```diff
@@ -304,15 +304,6 @@ public function increment(
 		return $this->update([$field => $value]);
 	}
 
-	/**
-	 * Checks if the model is locked for the current user
-	 */
-	public function isLocked(): bool
-	{
-		$lock = $this->lock();
-		return $lock && $lock->isLocked() === true;
```
distantnative, line 313: `Comment`: Breaking change but I don't think many would use this at all. If we think this is an issue, we can add it as deprecated alias method pointing to `$this->panel()->lock()->isLocked()`.

src/Panel/Model.php
```diff
@@ -332,7 +332,7 @@ public function options(array $unlock = []): array
 	{
 		$options = $this->model->permissions()->toArray();
 
-		if ($this->model->isLocked()) {
+		if ($this->lock()['isLocked'] === true) {
```
distantnative: `Question:` @bastianallgeier What do you think about having `$this->lock()` actually return the `Lock` object and not the array? It would make these occasions a lot cleaner.

bastianallgeier: Yes, I would definitely change that!

bastianallgeier (commented):
I'm a bit undecided if we should really remove `ModelWithContent::lock` and `::isLocked`. Both methods could still be useful. On the other hand, it's also good to clean up the models some more and you can still use `Lock::for($model)`

bastianallgeier (approved):

PR #6727, opened 2024-10-08
New dropdown with preview button 

bastianallgeier:
### Summary of changes

- The form controls now always show the dropdown (not just for the locked state)
- Page and Site views pass a new changesUrl prop to the form controls to show the preview option in the dropdown 
- The content module updates the lock modification timestamp

https://github.com/user-attachments/assets/2cc51aba-5d9c-4dcc-bf39-f55bd2c1eaf3

### DIFF
--- a/panel/src/components/Forms/FormControls.vue
+++ b/panel/src/components/Forms/FormControls.vue
@@ -1,16 +1,14 @@
 <template>
-	<k-button-group
-		v-if="buttons.length"
-		layout="collapsed"
-		class="k-form-controls"
-	>
-		<k-button
-			v-for="button in buttons"
-			:key="button.text"
-			v-bind="button"
-			size="sm"
-			variant="filled"
-		/>
+	<div v-if="buttons.length" class="k-form-controls">
+		<k-button-group layout="collapsed">
+			<k-button
+				v-for="button in buttons"
+				:key="button.text"
+				v-bind="button"
+				size="sm"
+				variant="filled"
+			/>
+		</k-button-group>
 		<k-dropdown-content
 			ref="dropdown"
 			align-x="end"
@@ -48,7 +46,7 @@
 				</k-dropdown-item>
 			</template>
 		</k-dropdown-content>
-	</k-button-group>
+	</div>
 </template>
 
 <script>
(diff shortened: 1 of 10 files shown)


panel/src/components/Forms/FormControls.vue
```diff
@@ -1,25 +1,29 @@
...
-		class="k-form-controls"
-	>
-		<k-button
-			v-for="button in buttons"
-			:key="button.text"
-			v-bind="button"
-			size="sm"
-			variant="filled"
-		/>
+	<div v-if="buttons.length" class="k-form-controls">
```
distantnative, line 2: `Question:` Why isn't the `k-button-group` an option as wrapper element anymore? Would save a div level.

panel/src/components/Views/ModelView.vue
```diff
@@ -9,6 +9,7 @@ export default {
 		api: String,
 		blueprint: String,
 		buttons: Array,
+		changesUrl: String,
```
distantnative, line 12: `Comment:` Not super happy with the name. It doesn't really explain what the URL is for: is that where I send the changes to store (I know it's not). Is something like `changesPreviewUrl` to verbose?

I am wondering if the FormControls could be mostly a `Panel\Ui\Button` like class as well, so that most of it actually gets assigned on the backend. But that's surely not for this PR.

panel/src/panel/content.js
```diff
@@ -133,6 +133,9 @@ export default (panel) => {
 				return;
 			}
 
+			// update the last modification timestamp
+			panel.view.props.lock.modified = new Date();
```
distantnative: `Todo:` I think this is in the wrong place. This needs to go into `save()` in the `finally` section doesn't it?

panel/src/components/Views/ModelView.vue
bastianallgeier: I'm not super happy with it either. We still pass the preview URL (for the published version) as part of the model object and I was first tempted to simply use "preview" as a new prop, but then decided against it to avoid confusion. 

I was also thinking about implementing the dropdown on the backend and load it asynchronously. The model view wouldn't need to pass the preview URL in that case at all. 

Maybe we can keep changesURL as a temporary solution, since we both agree that the final implementation for the form controls needs to be different anyway.

panel/src/components/Forms/FormControls.vue
bastianallgeier: When the dropdown is the last element, the collapsed button group creates a weird display CSS glitch. The rounded corners on the last element vanish when you open the dropdown.

distantnative (approved):

PR #6732, opened 2024-10-09
Multi-language fixes

bastianallgeier:
### Summary of changes

- `Version::exists` has a new wildcard option to check if a version exists in any language. E.g. `$page->version('changes')->exists('*')` We had this in an earlier draft where you could pass null to achieve this. But needing this again, I thought that the wildcard is helping to avoid confusion. It could be indeed an option for more places like this later. 
- I made sure to always get the `current` language in more places to fix bugs when changing the language. 
- Tracking changes has been improved and the wildcard above is used to untrack model changes properly if there are no more changes in any other language. 
- The lock field is now stored in every language file. This will already help to simplify the logic and it already enables the option to work on translations side by side. This has been possible before as well, when we stored stuff in local storage. 
- The lock field is now properly removed from the published version.

### DIFF
--- a/src/Panel/Controller/Changes.php
+++ b/src/Panel/Controller/Changes.php
@@ -24,10 +24,7 @@ class Changes
 	 */
 	public static function discard(ModelWithContent $model): array
 	{
-		$model->version(VersionId::changes())->delete();
-
-		// remove the model from the user's list of unsaved changes
-		App::instance()->site()->changes()->untrack($model);
+		$model->version(VersionId::changes())->delete('current');
 
 		return [
 			'status' => 'ok'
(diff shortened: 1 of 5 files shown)


src/Panel/Controller/Changes.php
```diff
@@ -84,9 +81,6 @@ public static function save(ModelWithContent $model, array $input): array
 			language: 'current'
 		);
 
-		// add the model to the user's list of unsaved changes
-		App::instance()->site()->changes()->track($model);
```
distantnative, line 88: `Question:` ~~Where is the tracking happening now?~~ In `Version::create()`.

And is the untacking still needed in `Controller\Changes::publish()` as it would also be taken care by `Version::delete()` now?

bastianallgeier: Damn, I totally missed that one. I just removed it.

distantnative (approved):

PR #6733, opened 2024-10-10
Use new changes cache instead of the changes field in the site.txt

bastianallgeier:
### Summary of changes

The Changes class now uses the new changes cache to store the model ids instead of putting them into the site.txt

### DIFF
--- a/src/Content/Changes.php
+++ b/src/Content/Changes.php
@@ -2,11 +2,15 @@
 
 namespace Kirby\Content;
 
+use Kirby\Cache\Cache;
 use Kirby\Cms\App;
+use Kirby\Cms\File;
 use Kirby\Cms\Files;
 use Kirby\Cms\ModelWithContent;
+use Kirby\Cms\Page;
 use Kirby\Cms\Pages;
 use Kirby\Cms\Site;
+use Kirby\Cms\User;
 use Kirby\Cms\Users;
 use Kirby\Toolkit\A;
 
@@ -23,15 +27,23 @@ use Kirby\Toolkit\A;
 class Changes
 {
 	/**
-	 * Access helper for the field, in which changes are stored
+	 * Access helper for the cache, in which changes are stored
 	 */
-	public function field(): Field
+	public function cache(): Cache
 	{
-		return $this
-			->site()
-			->version(VersionId::published())
-			->content()
-			->get('changes');
+		return App::instance()->cache('changes');
+	}
+
+	/**
+	 * Returns the cache key for a given model
+	 */
+	public function cacheKey(ModelWithContent $model): string
+	{
+		return match(true) {
+			$model instanceof File => 'files',
+			$model instanceof Page => 'pages',
+			$model instanceof User => 'users'
+		};
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


distantnative (commented):
Do we need to add a method to rebuild the cache? That actually walks through the index and checks. Or are we skipping this for now?

bastianallgeier:
I've skipped it for now, but I think it would be good to have this in the end.

src/Content/Changes.php
```diff
@@ -23,86 +27,112 @@
...
+	/**
+	 * Returns the cache key for a given model
+	 */
+	public function cacheKey(ModelWithContent $model): string
+	{
+		return match(true) {
+			$model instanceof File => 'files',
+			$model instanceof Page => 'pages',
+			$model instanceof User => 'users'
+		};
```
distantnative: `Suggestion:`

```suggestion
		return $model::CLASS_ALIAS;
```

or if you want it grammatically correct

```suggestion
		return $model::CLASS_ALIAS . 's';
```

src/Content/Changes.php
```diff
@@ -23,86 +27,112 @@
...
 	 */
-	public function field(): Field
+	public function cache(): Cache
 	{
-		return $this
-			->site()
-			->version(VersionId::published())
-			->content()
-			->get('changes');
+		return App::instance()->cache('changes');
```
distantnative: `Suggestion:` Since we keep using `App::instance()` around the class, why not define this once in the constructor?

```suggestion
		return $this->kirby->cache('changes');
```

src/Content/Changes.php
```diff
@@ -23,86 +27,112 @@
...
 
 	/**
 	 * Return all pages with unsaved changes
 	 */
 	public function pages(): Pages
 	{
-		return $this->field()->toPages();
+		return App::instance()->site()->find(
+			false,
+			false,
```
distantnative, line 72: `Question:` What are these for? `HasChildren::find()` does only accept strings and arrays. And `Collection::find()` also considers all arguments as keys, so why passing twice a `false` key?

src/Content/Changes.php
```diff
@@ -23,86 +27,112 @@
...
 
 	/**
 	 * Return all users with unsaved changes
 	 */
 	public function users(): Users
 	{
-		return $this->field()->toUsers();
+		return App::instance()->users()->find(
+			false,
+			false,
```
distantnative, line 133: `Question`: same as above

src/Content/Changes.php
bastianallgeier: It might sound a bit stupid, but I really just copied that code over from the field methods. I think it was actually about making sure that we return a collection and not just a single item.

src/Content/Changes.php
bastianallgeier: See above

src/Content/Changes.php
distantnative: :O I never spotted that :D And I guess `false` is casted to and empty string and thus will never actually return a real model.

distantnative (approved):

PR #6736, opened 2024-10-11
New `redis` cache driver

afbora:
## Description
Redis is one of the most used cache types and I thought it should be in the core. Since I don't use Redis cache much, PR may have some shortcomings. Redis has many features but I wanted to add its basic features.

````php
<?php

return [
    'cache' => [
        'pages' => [
            'active'    => true,
            'type'      => 'redis',
            'host'      => '127.0.0.1', // default host address
            'port'      => 6379, // default port number
            'auth'      => ['user', 'password'], // optional
        ]
    ]
];

````

### Featues
- New `redis` cache driver

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cache/RedisCache.php
+++ b/src/Cache/RedisCache.php
@@ -1,0 +1,160 @@
+<?php
+
+namespace Kirby\Cache;
+
+use Kirby\Cms\Helpers;
+use Redis;
+use Throwable;
+
+/**
+ * Redis Cache Driver
+ *
+ * @package   Kirby Cache
+ * @author    Ahmet Bora <ahmet@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://opensource.org/licenses/MIT
+ */
+class RedisCache extends Cache
+{
+	/**
+	 * Store for the redis connection
+	 */
+	protected Redis $connection;
+
+	/**
+	 * Sets all parameters which are needed to connect to Redis
+	 *
+	 * @param array $options 'host'   (default: 127.0.0.1)
+	 *                       'port'   (default: 6379)
+	 */
+	public function __construct(array $options = [])
+	{
+		$options = [
+			'host' => '127.0.0.1',
+			'port' => 6379,
+			...$options
+		];
+
+		parent::__construct($options);
+
+		// available options for the redis driver
+		$allowed = [
+			'host',
+			'port',
+			'readTimeout',
+			'connectTimeout',
+			'persistent',
+			'auth',
+			'ssl',
+			'retryInterval',
+			'backoff'
+		];
+
+		// filters only redis supported keys
+		$redisOptions = array_intersect_key($options, array_flip($allowed));
+
+		// creates redis connection
+		$this->connection = new Redis($redisOptions);
+
+		// sets the prefix if defined
+		if ($prefix = $options['prefix'] ?? null) {
+			$this->connection->setOption(Redis::OPT_PREFIX, rtrim($prefix, '/') . '/');
+		}
+
+		// selects the database if defined
+		$database = $options['database'] ?? null;
+		if ($database !== null) {
+			$this->connection->select($database);
+		}
+	}
+
+	/**
+	 * Returns the database number
+	 */
+	public function databaseNum(): int
+	{
+		return $this->connection->getDbNum();
+	}
+
+	/**
+	 * Returns whether the cache is ready to store values
+	 */
+	public function enabled(): bool
+	{
+		try {
+			return Helpers::handleErrors(
+				fn () => $this->connection->ping(),
+				fn (int $errno, string $errstr) => true,
+				fn () => false
+			);
+		} catch (Throwable) {
+			return false;
+		}
+	}
+
+	/**
+	 * Determines if an item exists in the cache
+	 */
+	public function exists(string $key): bool
+	{
+		return $this->connection->exists($this->key($key)) !== 0;
+	}
+
+	/**
+	 * Removes keys from the database
+	 * and returns whether the operation was successful
+	 */
+	public function flush(): bool
+	{
+		return $this->connection->flushDB();
+	}
+
+	/**
+	 * The key is not modified, because the prefix is added by the redis driver itself
+	 */
+	protected function key(string $key): string
+	{
+		return $key;
+	}
+
+	/**
+	 * Removes an item from the cache
+	 * and returns whether the operation was successful
+	 */
+	public function remove(string $key): bool
+	{
+		return $this->connection->del($this->key($key));
+	}
+
+	/**
+	 * Internal method to retrieve the raw cache value;
+	 * needs to return a Value object or null if not found
+	 */
+	public function retrieve(string $key): Value|null
+	{
+		$value = $this->connection->get($this->key($key));
+		return Value::fromJson($value);
+	}
+
+	/**
+	 * Writes an item to the cache for a given number of minutes
+	 * and returns whether the operation was successful
+	 *
+	 * <code>
+	 *   // put an item in the cache for 15 minutes
+	 *   $cache->set('value', 'my value', 15);
+	 * </code>
+	 */
+	public function set(string $key, $value, int $minutes = 0): bool
+	{
+		$key   = $this->key($key);
+		$value = (new Value($value, $minutes))->toJson();
+
+		if ($minutes > 0) {
+			return $this->connection->setex($key, $minutes * 60, $value);
+		}
+
+		return $this->connection->set($key, $value);
+	}
+}
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -59,7 +59,7 @@ jobs:
       matrix:
         php: ["8.2", "8.3"]
     env:
-      extensions: mbstring, ctype, curl, gd, apcu, memcached
+      extensions: mbstring, ctype, curl, gd, apcu, memcached, redis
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by bastianallgeier** [comment]
Nice! I really like the idea to add a redis cache driver by default. Could you take care of adding unit tests as well?

**Comment by afbora** [comment]
Thank you. I'll add. I didn't want to work on much more. Because I thought maybe the pr is not acceptable for you. If the PR for @lukasbestle and @distantnative is good overall I will continue with unit testing.

File: src/Cache/Redis.php [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Redis Cache Driver
+ *
+ * @package   Kirby Cache
+ * @author    Ahmet Bora <ahmet@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://opensource.org/licenses/MIT
+ */
+class Redis extends Cache
```
afbora: I think the class name should be `RedisCache`, right?

afbora: Also I would like to use separated namespace for drivers like that:

````php
Kirby\Cache\Cache
Kirby\Cache\Value
Kirby\Cache\Drivers\Apcu
Kirby\Cache\Drivers\File
Kirby\Cache\Drivers\MemCached
Kirby\Cache\Drivers\Memory
Kirby\Cache\Drivers\Null
Kirby\Cache\Drivers\Redis
````

What do you think?

Comment (distantnative) [reply src/Cache/Redis.php]: Yes it should be `RedisCache`, but why would it need a separate namespace? I think in the end, using something like `new FileCache()` or `new NullCache()` in your code is a little easier than `new File()`/`new Null()`

File: src/Cache/Redis.php [hunk 2]
```diff
@@ -0,0 +1,161 @@
...
+		$this->connection = new RedisClient($redisOptions);
+
+		// set the prefix if defined
+		if ($prefix = $options['prefix'] ?? null) {
+			$this->connection->setOption(RedisClient::OPT_PREFIX, $prefix);
+		}
+
+		// select the database if defined
+		$database = $options['database'] ?? null;
+		if ($database !== null) {
```
distantnative:
```suggestion
		if ($database = $options['database'] ?? null) {
```

afbora: I was expecting this suggestion :) But it will not work since database can be integer. So if database index number is `0`, if condition won't work. `if (0) { }`. Any idea?

src/Cache/Redis.php
afbora: Yes, maybe we will continue to use as `FileCache` like `use Kirby\Cache\Drivers\File as FileCache`. But directory structure will be more readable. Currently mixed drivers and main classes. Not important, it just the idea.

github com_getkirby_kirby_tree_feature_redis-cache_src_Cache

src/Cache/Redis.php
distantnative: Ah I see, no all good then as is

afbora:
Anyone have an idea about the failed tests? I've tried some changes but still failed.

afbora:
I've fixed some unit test except one: `RedisCacheTest::testConstructServer()`

https://github.com/getkirby/kirby/blob/feature/redis-cache/tests/Cache/RedisCacheTest.php#L105-L108
> RedisException: Redis server 127.0.0.1:1234 went away

If the redis server is invalid or inactive, the cache methods (`set`, `get`, `exists`) throw an above error instead of returning `null`. What do you think we should do? How about putting each method in a try catch block?

afbora:
I've put the methods into try catch block but now getting a warning when host is invalid:

> Redis::del(): php_network_getaddresses: getaddrinfo for invalid.host failed: Name or service not known

So this error can be `RedisException` or a warning. I can't decide how to handle this. Which way is the most stable and correct?

distantnative:
@afbora looking at the existing other cache drivers, they don't take care of this again in the get/set methods but leave it to the developer for checking `::enabled()`. I think we should then do it the same here and assume in the methods that a valid connection exists and not wrap them in try-catch as well.

So for testing the invalid states in https://github.com/getkirby/kirby/pull/6736/files#diff-505d24331add9cae9e36caf292b1b664163d8248226df1254c19cf1ceb34b573R103-R120, I wouldn't test the set/remove methods, but rather only test if `::enabled()` returns the right value.

afbora:
> looking at the existing other cache drivers, they don't take care of this again in the get/set methods but leave it to the developer for checking ::enabled()

Exactly! That's why I'm confused. 

For ex: memcached returns `null` when you try to access a key while invalid memcached setup. But redis will not respond same behavior. Do you think is that a problem?

afbora:
@distantnative I think we should also implement to the models. For ex: `Page::isCacheable()` method also should check the `$cache->enabled()` method. What do you think? Maybe this change can be part of v5 to prevent possible breaks.

afbora:
Now getting imagemagick errors in unit tests 🤷‍♂️

distantnative:
@afbora The imagick errors seem to be a general problem: https://github.com/getkirby/kirby/actions/runs/11325231985/job/31491697204?pr=6739

Will try to see how to fix them on `develop-minor` first.

afbora:
@distantnative @bastianallgeier I've tracked the issue and found a clue about imagemagick issue. I can't write from Discord that's why I'm writing here, sorry.

Last success workflow run on `Ubuntu 22.04.5`
https://github.com/getkirby/kirby/actions/runs/11306649866/job/31447482868

But failed workflows run on `Ubuntu 24.04.1`
https://github.com/getkirby/kirby/actions/runs/11329888134/job/31506434810

That's the difference. But I've still no idea about the solution.

Some references:
- https://github.com/actions/runner-images/issues/10636
- https://github.blog/changelog/2024-09-25-actions-new-images-and-ubuntu-latest-changes/

distantnative:
I have been getting deprecation notices locally for the convert command. So my guess is, that's using one higher imagick version where this is not supported at all anymore. But it's a little trick for us to know which command to run as we need to support different imagick versions for our users

afbora:
Please check out https://github.com/ImageMagick/ImageMagick/discussions/7366

distantnative:
Yes, but how do we support just `convert` for IM 5 and just `magick` for IM 7 at the same time?

afbora:
Not sure. Via `Imagick::getVersion()`? Or running `convert -version` in try catch? (If not catched that's mean version is 5, otherwise 7)

distantnative:
Haven't been able to catch a failing exec so far.

afbora:
I think migration will be a little complicated. So I think we should use `runs-on: ubuntu-22.04` for github actions until we fixed the issue.

afbora:
I've tested and it's working with: `runs-on: ubuntu-22.04`
https://github.com/getkirby/kirby/actions/runs/11341791723

Here the PR: https://github.com/getkirby/kirby/pull/6743

bastianallgeier:
I think it's a good temporary fix to get on with the PRs.

src/Cache/RedisCache.php
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	/**
+	 * Removes all keys from all databases
+	 * and returns whether the operation was successful
+	 */
+	public function flushAll(): bool
+	{
+		return $this->connection->flushAll();
+	}
```
afbora: Do you think `::flushAll()` is needed? Since only one active redis DB will be used at all times, I think `::flushDb()` is sufficient. Do you think multiple redis DBs can be used at the same time? Also, this requires a complex config setting.

bastianallgeier: I would keep it as simple as possible

afbora:
Ready to review 👍

bastianallgeier:
I hope that my fix is correct, but looking at the entries, the prefix was added twice. By Redis::OPT_PREFIX and the key method. I now simply overwrote the key method to always return the plain key. We could also decide to not use the key method at all.

afbora:
Thank you Bastian. I suspected this but I couldn't check it because I don't have a setup where I can test the entries 🙈 

I think we can use the `key` property directly. We can also use `@codeCoverageIgnore` for the `key` method.

bastianallgeier:
I think we can also leave it like it is. Maybe the method could be useful later. Especially if you extend the RedisCache driver.

bastianallgeier (dismissed):
I've tested the regular operations with the starterkit and it all seems to work fine. I just wonder if we want to put this into 4.5 or into 5.0. @distantnative @lukasbestle what do you think?

distantnative:
Maybe v5 based on the recent partners call – e.g. if someone created their own RedisDriver so far this could clash. Or is too cautious?

bastianallgeier:
Sounds good to me. @afbora could you base it on the v5/develop branch?

afbora:
> if someone created their own RedisDriver so far this could clash

@distantnative But this new driver will be enabled when only `caches.pages.type === redis`. Isn't it? Do you mean for existing apps whose already owning `Kirby\Cache\RedisCache` cache class?

bastianallgeier:
@afbora yes, I think this is what Nico meant. We already have a Redis cache driver plugin by Bruno. It has a different class namespace and should not collide there, but of course the configs would collide. His plugin has slightly different cache options.

distantnative:
> Do you mean for existing apps whose already owning Kirby\Cache\RedisCache cache class?

Yes, that's what I meant. Not sure if it's too overcautious.

bastianallgeier:
@distantnative I think it's a good call and the new driver will also be a nice additional feature for v5

PR #6737, opened 2024-10-11
New Kirby\Plugin namespace

bastianallgeier:
## Description

This PR moves all CMS plugin classes into their own namespace and adds aliases for the old class names. This PR is in preparation for a new plugin license class to provide a way for plugin developers to handle custom licensing. The PR for the second step will follow …

### Deprecated

- `Kirby\Cms\Plugin` is deprecated. Use `Kirby\Plugin\Plugin` instead
- `Kirby\Cms\PluginAsset` is deprecated. Use `Kirby\Plugin\Asset` instead
- `Kirby\Cms\PluginAssets` is deprecated. Use `Kirby\Plugin\Assets` instead

## Ready?

- [x] In-code documentation (wherever needed)
- [x] Unit tests for fixed bug/feature
- [x] Tests and CI checks all pass

### For review team

- [x] Add changes & docs to release notes draft in Notion

PR #6738, opened 2024-10-11
New plugin license classes

bastianallgeier:
## Merge first 

- [x] https://github.com/getkirby/kirby/pull/6737

## Description

The new `Kirby\Plugin\License` and `Kirby\Plugin\LicenseStatus` classes add the possibility for plugin developers to create their own custom licenses and set their status and link to a potential license shop. 

### Features

- New Plugin License handling
  - New `Kirby\Plugin\License` class
  - New `Kirby\Plugin\LicenseStatus` class
  - New `k-table-license-cell` component in the system view
  - Licenses are now shown in the plugins table with an additional icon and theme to indicate their status. 
  - Licenses can now have an additional link to their license document or a custom license shop. 

Screenshot 2024-10-11 at 14 59 46

### Breaking changes

- `Plugin::license()` now returns the License object instead of a simple string
- `Plugin::toArray()` includes the full license info as child array instead of a string

## Docs

Defining a plugin license as simple array

```php
App::plugin(
  name: 'my/plugin', 
  extends: [...],
  license: [
    'name' => 'Custom license', 
    'link' => 'https://mylicenseshop.com', 
    'status' => 'missing',
  ]
);
```

Defining a plugin license with the new license class

```php
use Kirby\Plugin\License;

App::plugin(
  name: 'my/plugin', 
  extends: [...],
  license: new License(
    name: 'Custom license',
    link: 'https://mylicenseshop.com', 
  )
);
```

You can also extend the license class to customize its behavior

```php
use Kirby\Plugin\License;
use Kirby\Plugin\LicenseStatus;

class MyCustomLicense extends License 
{
  public function __construct()
  {
    $this->name = 'My Custom License';
    $this->link = 'https://mylicenseshop.com';
    $this->status = LicenseStatus(
      value: 'missing', 
      theme: 'negative', 
      label: 'Get a license please',
      icon: 'alert'
    )
  }
}

App::plugin(
  name: 'my/plugin', 
  extends: [...],
  license: new MyCustomLicense()
);
```

## Ready?

### For review team


bastianallgeier:
@distantnative how did you fix the CI frontend issue in the changes branches? I think it's the same thing here.

src/Plugin/License.php
```diff
@@ -0,0 +1,103 @@
...
+	 */
+	public function __toString(): string
+	{
+		return $this->name();
+	}
+
+	/**
+	 * Creates a license instance from a given value
+	 */
+	public static function from(self|array|string|null $license): static
```
distantnative:
```suggestion
	public static function from(License|array|string|null $license): static
```

More common in the rest of our code

src/Plugin/License.php
```diff
@@ -0,0 +1,103 @@
...
+	{
+		if ($license instanceof self) {
```
distantnative:
```suggestion
		if ($license instanceof License) {
```

src/Plugin/LicenseStatus.php
```diff
@@ -0,0 +1,140 @@
...
+				label: 'Unknown license',
+				theme: 'passive'
+			),
+		];
+	}
+
+	/**
+	 * Returns a status by its name
+	 */
+	public static function from(self|string|null $status): static
```
distantnative:
```suggestion
	public static function from(LicenseStatus|string|null $status): static
```

src/Plugin/LicenseStatus.php
```diff
@@ -0,0 +1,140 @@
...
+	{
+		if ($status instanceof self) {
```
distantnative:
```suggestion
		if ($status instanceof LicenseStatus) {
```

src/Plugin/LicenseStatus.php
```diff
@@ -0,0 +1,140 @@
...
+				theme: SystemLicenseStatus::Missing->theme()
+			),
+			'unknown' => new static(
+				value: 'unknown',
+				icon: 'question',
+				label: 'Unknown license',
+				theme: 'passive'
+			),
+		];
+	}
```
distantnative: `Suggestion`:  This doesn't look very DRY to me, so I had a though:
- Get rid of `::defaults()`
- Add `unkown` as case to the system `LicenseStatus`
- And then in `LicenseStatus::from()` we could to

```php
$status   = SystemLicenseStatus::fromString($status ?? 'unknown'):
$status ??= SystemLicenseStatus::unknown;
return new static(
    value: $status->value,
    icon: $status->icon(),
    label: $status->label(),
    theme: $status->theme()
)
```

distantnative:
@bastianallgeier it's annoying:
- Delete `package-lock.json` and the `panel/node_modules` folder
- Run `npm i`
- Commit the updated `package-lock.json`

That has worked for me a few times when this error pops up.

distantnative:
And trying them out, I was thinking that we might want to tweak the icons and colors from `Cms\LicenseStatus` a little.

distantnative:
Sorry, another thought: maybe somehow allowing to use the button as button would be nice for developers who want to implement their own registration dialog.

src/Plugin/LicenseStatus.php
bastianallgeier: Love this! I just implemented it.

bastianallgeier:
I agree about tweaking the icons and themes, but would do that in a separate step. I also think that an option to configure the buttons and even introduce custom dialogs would be awesome. But I would also plan this as some of the next steps.

distantnative (approved):

nilshoerrmann:
I checked this out today: How would you prefer to get feedback on this topic? 
Here in this closed PR? In a new issue? On Discord?

bastianallgeier:
@nilshoerrmann feel free to post it here and we can take optional changes over to another PR.

nilshoerrmann:
Okay, I will post notes here, no specific order:

1. If the license status requires a user action, it is currently not clear that the license status is interactive and can be clicked. I know that this is technically a link but it would be great to display this as button if the user is supposed to do something. E. g. the current design is just fine for active licenses but not for missing or invalid licenses that require a call to action.
2. If the license status is `missing` the current design doesn't highlight the fact that the current plugin state is invalid. The pink key icon looks nice but does not imply that anything is wrong. I think that states that require interaction should also have a written label, `missing` in this case.
3. There should be a status of `invalid` for cases of, well, invalid licenses (incorrect data, invented keys).
4. This is more about documentation: What does `inactive` as status mean? Is that an invalid or a valid state? E. g. it could mean "expired but valid" (no updates any more) or it could mean "deactivated and no longer valid".
5. In general, as a plugin developer, it would be great to know which statuses Kirby defines as valid and which ones are considered invalid.
6. Why not extend the simple array syntax to allow additional status settings, e. g.:
```php
App::plugin(
  name: 'my/plugin', 
  extends: [...],
  license: [
    'name' => 'Custom license', 
    'link' => 'https://mylicenseshop.com', 
    'status' => [
      'value' => 'missing', 
      'theme' => 'negative', 
      'label' => 'Get a license please',
      'icon' => 'alert'
    ]
  ]
);
```

distantnative:
@nilshoerrmann the most common use case for plugin developers will be a custom license class (to check, validate etc. license). In this, you can already set the theme, label etc.

What the different license states mean: https://github.com/getkirby/kirby/pull/6738/files#diff-64a97de7bd826afd0db41e35cac44f1d90b4807c206d38a1648c5330781cac40L18-R50

nilshoerrmann:
> [T]he most common use case for plugin developers will be a custom license class …

I'm not entirely sure about this, because – thinking of our own work – we also create project/client related plugins that are very basic (no Composer, no submodule) which would benefit from custom license display as well. The easier the better. So this is not just about plugins that are sold in a custom marketplace.

nilshoerrmann:
One question regarding plugin implementation:
We have plugins that work across different major Kirby versions. How can we add this license information for K5 while still being able to use the plugin in K4? (K4 not knowing the license classes yet of course.)

bastianallgeier:
In general, we would love to keep a certain pre-defined set of license statuses that are consistent to what we are doing. But adding the option to set the license status via array would be an easy change: https://github.com/getkirby/kirby/pull/6745

To make sure that the plugins still work in older installations, you could simply check if the `Kirby\Plugin\License` class exists.

distantnative:
And probably a notation for supporting older versions something a long the lines of


```php
$extends = [...];

if (class_exists('\Kirby\Plugin\License')) {
  $license = [...];

  App::plugin($name, $extends, $license);
} else {
  App::plugin($name, $extends);
}
```

PR #6742, opened 2024-10-14
Refactor Form Classes

bastianallgeier:
## Description

This is a first step to clean up the form and field classes. We need this for our changes to make sure that we only run through field property evaluation when really necessary. 

### Summary of changes

- New `Kirby\Toolkit\Component::applyProp` method to refactor prop application and make it more accessible for individual props
- Refactored `Kirby\Form\Field` class
  - New `Kirby\Form\Field::fill` method which overwrites the value and reevaluates the computed properties. 
  - New `Kirby\Form\Field::isDisabled` method to get the old class closer to the new `FieldClass`
- Refactored `Kirby\Form\Fields` class
  - Improved `Kirby\Form\Fields::toArray` to accept a map function 
  - New `Kirby\Form\Fields::defaults` method to return default values for all fields in the collection
  - New `Kirby\Form\Fields::errors` method to return all validation errors for all fields in the collection
  - New `Kirby\Form\Fields::fill` method to add new values to matching fields
- Remove the outdated md5 signature from the `FieldClass`. I missed this the last time when I removed the signature from the old Field class. 
- Refactored `Kirby\Form\Form` class
  - Simplified `Kirby\Form\Form::toArray` method
  - Use the new `Kirby\Form\Fields::errors` method for `Kirby\Form\Form::errors`

## Ready?

### For review team

### DIFF
--- a/src/Form/Fields.php
+++ b/src/Form/Fields.php
@@ -3,6 +3,7 @@
 namespace Kirby\Form;
 
 use Closure;
+use Kirby\Toolkit\A;
 use Kirby\Toolkit\Collection;
 
 /**
@@ -36,6 +37,14 @@ class Fields extends Collection
 		parent::__set($field->name(), $field);
 	}
 
+	/**
+	 * Returns an array with the default value of each field
+	 */
+	public function defaults(): array
+	{
+		return $this->toArray(fn($field) => $field->default());
+	}
+	
 	/**
 	 * Converts the fields collection to an
 	 * array and also does that for every
--- a/src/Form/Fields.php
+++ b/src/Form/Fields.php
@@ -20,6 +20,13 @@ use Kirby\Toolkit\Collection;
 class Fields extends Collection
 {
 	/**
+	 * Cache for the errors array
+	 *
+	 * @var array<string, array<string, string>>|null
+	 */
+	protected array|null $errors = null;
+
+	/**
 	 * Internal setter for each object in the Collection.
 	 * This takes care of validation and of setting
 	 * the collection prop on each object correctly.
@@ -35,6 +42,9 @@ class Fields extends Collection
 		}
 
 		parent::__set($field->name(), $field);
+
+		// reset the errors cache if new fields are added
+		$this->errors = null;
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


bastianallgeier:
Oh nice. The imagemagick issue is now also hitting us here.

src/Form/Fields.php
```diff
@@ -36,19 +37,52 @@ public function __set(string $name, $field): void
...
+
+	/**
+	 * An array of all found in all fields errors
+	 */
+	public function errors(): array
+	{
+		$errors = [];
+
+		foreach ($this->data as $name => $field) {
+			if (empty($field->errors()) === false) {
```
distantnative:
```suggestion
			if ($field->errors() !== []) {
```

Or using `count()`. But we should get away from using `empty()`.

src/Form/Fields.php
```diff
@@ -36,19 +37,52 @@ public function __set(string $name, $field): void
...
+	 */
+	public function defaults(): array
+	{
+		return $this->toArray(fn ($field) => $field->default());
+	}
+
+	/**
+	 * An array of all found in all fields errors
+	 */
+	public function errors(): array
```
distantnative, line 61: We have been caching the errors in the Form class. Do you think it's ok regarding performance to not cache them here?

bastianallgeier: You are right. We should still cache them. I just need to make sure that the cache gets flushed when the fill method is used.

bastianallgeier: ☑️

src/Form/Fields.php
```diff
@@ -50,18 +60,22 @@ public function defaults(): array
...
+		}
+
+		$this->errors = [];
 
 		foreach ($this->data as $name => $field) {
 			if ($field->errors() !== []) {
-				$errors[$name] = [
+				$this->errors[$name] = [
 					'label'   => $field->label(),
 					'message' => $field->errors()
```
distantnative: `Suggestion`: Instead of running `$field->errors()` twice, write it first to a variable, then check the variable for `!== []` and use it in the array.

bastianallgeier: I'm sorry, I don't think I can follow :)

distantnative: You are currently calling `$field->errors()` in the if statement and then in the array assignment again.

I think it would be better to be

```php
$errors = $field->errors():
if ($errors !== []) {
    $this->errors[$name] = [
        'label' => $field->label(),
        'errors' => $errors
    ]
}
```

bastianallgeier: Oh, now I got you. Yes, that definitely makes sense.

distantnative (approved):

PR #6743, opened 2024-10-15
Use `runs-on: ubuntu-22.04`

afbora:
Use `runs-on: ubuntu-22.04` instead `ubuntu-latest` until we fixed the issue about imagemagick and ubuntu 24.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -53,7 +53,7 @@ jobs:
         (github.repository == 'getkirby/kirby' || github.repository_owner != 'getkirby')
       )
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):

distantnative:
@bastianallgeier @afbora i think this is nice to temporarily fix our CI, but we shouldn't create a false feeling of being fine: the problem that this revealed will soon hit our users in their projects too. And I haven't found any good path to a solution yet.

bastianallgeier:
@distantnative oh, absolutely! We should turn this into an issue and try to find a fix asap.

afbora:
@distantnative Exactly! We only updated the workflow temporary until we fixed the issue: https://github.com/getkirby/kirby/issues/6744

PR #6746, opened 2024-10-15
Discard changes when no diff with published content

distantnative:
### Summary of changes
- New `Version::prepareFieldsForContent()` where we e.g. unset fields (currently, `lock` and `slug`) that will be returned from `::read()` (and cannot be filtered by `::prepareFieldsAfterRead()`, otherwise we cannot read them for legitimate use cases) but should not be part of the `Content` object
- Use `::content()` instead of `::read()` for `Version::diff()`
- Allow passing a `Version` object directly to `Version::diff()`
- Delete changes version when `Panel\Controller\Changes::save()` produces a result with no diff between published and changes (basically when the user manually modifies the content back to the state of published)

### Reasoning
Prevent the from controls from becoming a ghost (appearing even though there are no real changes).

## Ready?

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -2,9 +2,11 @@
 
 namespace Kirby\Content;
 
+use Kirby\Cms\File;
 use Kirby\Cms\Language;
 use Kirby\Cms\Languages;
 use Kirby\Cms\ModelWithContent;
+use Kirby\Cms\Page;
 use Kirby\Exception\LogicException;
 use Kirby\Exception\NotFoundException;
 
@@ -301,7 +303,16 @@ class Version
 		array $fields,
 		Language $language
 	): array {
-		unset($fields['lock'], $fields['slug']);
+		unset($fields['lock']);
+
+		if ($this->model instanceof Page) {
+			unset($fields['slug']);
+		}
+
+		if ($this->model instanceof File) {
+			unset($fields['template']);
+		}
+
 		return $fields;
 	}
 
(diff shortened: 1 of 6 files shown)


src/Content/Version.php
```diff
@@ -283,6 +293,18 @@ protected function prepareFieldsAfterRead(array $fields, Language $language): ar
...
 
+	/**
+	 * Make sure that the Content object receives the right set of fields
+	 * filtering fields used for lower logic (e.g. lock)
+	 */
+	protected function prepareFieldsForContent(
+		array $fields,
+		Language $language
+	): array {
+		unset($fields['lock'], $fields['slug']);
```
bastianallgeier: Should we make this dependent on the model and filter out sort and template here for files instead? Slug also only needs to be filtered out for pages.

distantnative: Yes, changing it.

Only that `sort` currently is really treated as content field in the way how we use it. So removing that breaks a few things and would likely be a breaking change for users, which we could do for v5. What do you think? (We would need to create a dedicated `File::sort(): Field` method to keep the breaking change small).

bastianallgeier (approved):

src/Content/Version.php
bastianallgeier: Ok, let's take this to another PR and make sure we get it right there. I think we should really start treating such values as system values and not mix them up with fields. The compatibility method sounds like a good plan, but I would probably immediately deprecate it.

distantnative: Added a "nice to have" todo in the Notion Changes todo list.

PR #6748, opened 2024-10-15
Store changes before exit view

distantnative:
### Summary of changes
- Call `panel.content.save()` when changing views and not all unsaved changes have been sent to backend yet
- Show warning when user tries to close tab and not all unsaved changes have been sent to backend yet

## Ready?

### For review team

### DIFF
--- a/panel/src/panel/content.js
+++ b/panel/src/panel/content.js
@@ -5,7 +5,7 @@ import { reactive } from "vue";
  * @since 5.0.0
  */
 export default (panel) => {
-	return reactive({
+	const content = reactive({
 		/**
 		 * API endpoint to handle content changes
 		 */
@@ -69,6 +69,13 @@ export default (panel) => {
 		isProcessing: false,
 
 		/**
+		 * Whether all content updates have been successfully sent to the backend
+		 *
+		 * @returns {Boolean}
+		 */
+		isSaved: true,
+
+		/**
 		 * The last published state
 		 *
 		 * @returns {Object}
@@ -124,6 +131,7 @@ export default (panel) => {
 				// update the last modification timestamp
 				panel.view.props.lock.modified = new Date();
 
+				this.isSaved = true;
 				this.isProcessing = false;
 			} catch (error) {
 				// silent aborted requests, but throw all other errors
@@ -154,6 +162,8 @@ export default (panel) => {
 				...panel.view.props.originals,
 				...values
 			};
+
+			this.isSaved = false;
 		},
 
 		/**
@@ -163,6 +173,30 @@ export default (panel) => {
 		 */
 		get values() {
 			return panel.view.props.content;
+		},
+
+		rescue() {
+			if (this.isSaved === false) {
+				const data = new FormData();
+
+				for (const field in panel.view.props.content) {
+					data.append(field, panel.view.props.content[field]);
+				}
+
+				navigator.sendBeacon(panel.url(this.api + "/changes/save"), data);
+
+				// update the last modification timestamp
+				panel.view.props.lock.modified = new Date();
+				this.isSaved = true;
+			}
 		}
 	});
+
+	document.addEventListener("visibilitychange", () => {
+		if (document.visibilityState === "hidden") {
+			content.rescue();
+		}
+	});
+
+	return content;
 };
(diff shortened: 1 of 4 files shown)


panel/src/panel/content.js
```diff
@@ -149,6 +159,30 @@ export default (panel) => {
...
+
+		rescue() {
+			if (this.isSaved === false) {
+				const data = new FormData();
+
+				for (const field in panel.view.props.content) {
+					data.append(field, panel.view.props.content[field]);
+				}
+
+				navigator.sendBeacon(panel.url(this.api + "/changes/save"), data);
```
bastianallgeier: Did you know how much data we can send via sendBeacon? Does it have any limits? I've never worked with it before.

panel/src/panel/content.js
```diff
@@ -149,6 +159,30 @@ export default (panel) => {
...
+
+				// update the last modification timestamp
+				panel.view.props.lock.modified = new Date();
+				this.isSaved = true;
+			}
 		}
 	});
+
+	document.addEventListener("visibilitychange", () => {
```
bastianallgeier: `Suggestion:` We could maybe try to move this to the events.js to make sure that we have a better overview of global event handlers.

panel/src/panel/view.js
```diff
@@ -23,6 +23,14 @@ export default (panel) => {
 	return reactive({
 		...parent,
 
+		async open(feature, options = {}) {
+			if (panel.content.isSaved === false) {
+				await panel.content.save();
```
bastianallgeier: At first sight, I'm not entirely sure what could be a better place. But this seems to me like we need to have at least some sort of toggle to know if the previous view was actually a model view. We probably don't really want this to run on every view change. I.e. you come from the system view and you navigate to the users view.

panel/src/panel/content.js
distantnative: From various discussions on the web, it's depending on the browser but seems to be limited in Chrome and Safari to 64 KB. Same for `fetch` with `keepalive`, which though isn't supported by Firefox yet.

Alternatively, the preventing `beforeunload` event (which opens this "you really want to close, there might be unsaved suer data" popup) would be nicer in the way that it jus buys enough time to send the last regular throttled request. But the message is not configurable, so it can be confusing to the user, e.g. thinking that they need to really save their unsaved changes and not just wait a second for storing unsaved changes.

We could still use this as the safer alternative in terms of data processing, but maybe reduce the throttle delay to decrease the likelihood of that opoup needing to show up ever.

bastianallgeier: Is it complicated to implement the popup? If it's easy to do, I would probably really go this way as a first solution and then see how often we run into it and how confusing it really is.

distantnative: It's super easy:

```js
window.addEventListener('beforeunload', function (e) {
    e.preventDefault();
    e.returnValue = '';
});
```

Or for our case then inside wrapped with `if (this.isSaved === false)`.

panel/src/panel/view.js
distantnative: @bastianallgeier What if `content.isSaved` would be always true for non-model views?

bastianallgeier: That's a good idea.

panel/src/panel/content.js
bastianallgeier: Then I would really go this way - at least for the alphas and betas.

panel/src/panel/view.js
distantnative: I used a watcher for the `api` inside `content.js` now. This will trigger also on view changes of non-model views. However, for those `isSaved` should never be changed to `false` (or if, then because those views are actually using `$panel.content.update()` - so with `isSaved = true` the saving will not be pursued.

panel/src/panel/panel.js
```diff
@@ -91,6 +90,7 @@ export default {
 		// features
 		this.dropdown = Dropdown(this);
 		this.view = View(this);
+		this.content = Content(this);
```
distantnative, line 93: `Comment:` without this I sometimes got console errors on first load for not finding `panel.view.props` etc.

bastianallgeier: I'm ok with that.

bastianallgeier (approved):
Works great, but I had to try really hard to type and then hit cmd+w fast enough to close the window to trigger the dialog 😆

distantnative:
@bastianallgeier 😆 I usually increase the throttle delay in ModelView and add a `sleep(1)` to `Panel\Controller\Changes::save()` to give me a little more time

PR #6749, opened 2024-10-16
Abort `save` requests to prevent race conditions

distantnative:
### Summary of changes
- Abort previous requests in `$panel.content.save()` to avoid that a long-running previous request with old data is processed after the new request with the newer data
- Ensure that the last request really is being sent

### DIFF
--- a/panel/src/panel/content.js
+++ b/panel/src/panel/content.js
@@ -110,21 +110,31 @@ export default (panel) => {
 		 * Saves any changes
 		 */
 		async save() {
-			if (this.isProcessing === true) {
-				return true;
-			}
-
 			this.isProcessing = true;
 
+			// ensure to abort unfinished previous save request
+			// to avoid race conditions with older content
+			this.saveAbortController?.abort();
+			this.saveAbortController = new AbortController();
+
 			try {
-				await panel.post(this.api + "/changes/save", panel.view.props.content);
-			} finally {
+				await panel.post(this.api + "/changes/save", panel.view.props.content, {
+					signal: this.saveAbortController.signal
+				});
+
 				this.isProcessing = false;
 
 				// update the last modification timestamp
 				panel.view.props.lock.modified = new Date();
+			} catch (error) {
+				// silent aborted requests, but throw all other errors
+				if (error.name !== "AbortError") {
+					throw error;
+				}
 			}
 		},
+
+		saveAbortController: null,
 
 		/**
 		 * Updates the values of fields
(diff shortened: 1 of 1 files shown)


panel/src/panel/content.js
```diff
@@ -110,22 +110,32 @@ export default (panel) => {
 		 * Saves any changes
 		 */
 		async save() {
-			if (this.isProcessing === true) {
-				return true;
-			}
```
distantnative, line 115: If we simply ignore requests while `isProcessing`, it could be that that last save request at the end of input is ignored if the backend is really slow (as previous requests won't have completed). We need to remove this to ensure that the latest request gets processed.

panel/src/panel/content.js
```diff
@@ -110,22 +110,32 @@ export default (panel) => {
...
-				await panel.post(this.api + "/changes/save", panel.view.props.content);
-			} finally {
+				await panel.post(this.api + "/changes/save", panel.view.props.content, {
+					signal: this.saveAbortController.signal
+				});
+
 				this.isProcessing = false;
 
 				// update the last modification timestamp
 				panel.view.props.lock.modified = new Date();
```
distantnative: I moved these inside the `try` part as these shouldn't happen for failed requests or aborted requests. Maybe `isProcessing = false` should also happen for failed requests but in that case, we would rather need a counter how many requests have been started and `isProcessing` is a computed prop with `request !== 0` logic.

bastianallgeier: What happens on abort though? Could we reset isProcessing on abortion and then always have a correct state without a counter?

distantnative: Reset to what? `false`? But when a request is aborted, it means that another request was just started.

bastianallgeier: Sorry, I missed the point here. The second request would have already set isProcessing to true and if we set it to false it would be totally wrong. Then we probably really need a counter.

distantnative: @bastianallgeier What about this?

```js
try {
	await panel.post(...);
	
	// update the last modification timestamp
	panel.view.props.lock.modified = new Date();

	this.isProcessing = false;
} catch (error) {
	if (error.name !== "AbortError") {
		this.isProcessing = false;
		throw error;
	}
}
```

bastianallgeier: Oh, if that works, that'd be great!

bastianallgeier (approved):

PR #6750, opened 2024-10-16
Ensure changes exists after reading from cache

distantnative:
## Description

See discussion of https://github.com/getkirby/kirby/pull/6751 that might interplay with this PR.
 
### Summary of changes
- New `Content\Changes::ensure()` method that double-checks for existing changes and untracks model if none exist anymore
- `Changes::files()`, `Changes::pages()` and `Changes::users()` have new `$ensure = true` parameter that controls whether the collection retrieved from the cache should be double-checked

### Reasoning
This really enforces that those changes for the model really exists. However, it makes the `Changes` class a little less flexible (and probably performant due to the extra checks), especially when testing just the cache.

## Ready?

### DIFF
--- a/src/Content/Changes.php
+++ b/src/Content/Changes.php
@@ -7,7 +7,6 @@ use Kirby\Cms\App;
 use Kirby\Cms\Files;
 use Kirby\Cms\ModelWithContent;
 use Kirby\Cms\Pages;
-use Kirby\Cms\Site;
 use Kirby\Cms\Users;
 use Kirby\Toolkit\A;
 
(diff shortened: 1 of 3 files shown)


bastianallgeier:
Is there ever a situation where we want to set the $ensure argument for one of the collection methods to false? I was wondering if we could just leave this out there and always run the ensure method first.

distantnative:
I mainly added it to not complicate the unit tests that currently just set the cache, but not actually create the changes versions.

bastianallgeier:
I get kind of conservative with method arguments these days. We've always thrown them around like they cost nothing, but they always cause pain later when we no longer need them or they should be changed somehow. That's why I would probably rather change the unit tests.

bastianallgeier (approved):

PR #6751, opened 2024-10-16
Changes dialog: entry per language

distantnative:
### To be discussed

There is a certain overlap with https://github.com/getkirby/kirby/pull/6750 where we would now be checking twice if a version actually exists. Has me wondering if this logic should rather be part of `Content\Changes` but I don't have a good idea for syntax/method name or so.

### Summary of changes
- Shows a changes dialog entry per version language that has changes
- Shows language code as entry info in multilang

### Reasoning
- Users should be able to quickly get to the unsaved changes from the dialog. Before this PR it's difficult for them if the changes aren't in the default language.
- We only track the UUID for changes not the language. This is why it needs extra logic for the dialog, checking which language has the change. Alternatively, we would need to consider how to track and retrieve the language alongside uuid.

## Ready?

### DIFF
--- a/src/Panel/ChangesDialog.php
+++ b/src/Panel/ChangesDialog.php
@@ -3,7 +3,10 @@
 namespace Kirby\Panel;
 
 use Kirby\Cms\Collection;
+use Kirby\Cms\Languages;
 use Kirby\Content\Changes;
+use Kirby\Content\VersionId;
+use Kirby\Toolkit\Str;
 
 /**
  * Manages the Panel dialog for content changes in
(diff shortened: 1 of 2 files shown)


bastianallgeier:
I was thinking about changing this behavior. Especially since we are now listing all changes for all users, the changes dialog might potentially have to list a lot of changed models. If we then also show an item for each language, it could be a problem. 

My idea would be to show only if a model has changed and then improve the language dropdown in a model view. Right now, it's pretty confusing when you are in the English translation for example and there are no changes, but the German translation might have some. The language dropdown has no indication that other translations could have changes. We could solve this with a badge and then in the dropdown, we could also add indicators to each translation with changes.

distantnative:
Ok, maybe we could think about highlighting changed fields as well.

PR #6753, opened 2024-10-16
Changes dialog: rebuild cache if not exists

distantnative:
### Summary of changes
- `Content\Changes::cacheExists()` to check if the cache exists or was flushed
- `Content\Changes::generateCache()` to loop through models and update the cache
- Changes dialog: when opening and the changes cache doesn't exist, generate it from site index
  - _Any way we can make this performant for large indexes?_
- Adapted `Content\Changes` so it also works when UUIDs have been disabled

## Ready?

### DIFF
--- a/src/Content/Changes.php
+++ b/src/Content/Changes.php
@@ -38,6 +38,14 @@ class Changes
 	}
 
 	/**
+	 * Returns whether the cache has been populated
+	 */
+	public function cacheExists(): bool
+	{
+		return $this->cache()->get('__updated__') !== null;
+	}
+
+	/**
 	 * Returns the cache key for a given model
 	 */
 	public function cacheKey(ModelWithContent $model): string
@@ -157,6 +165,7 @@ class Changes
 		$changes = array_values($changes);
 
 		$this->cache()->set($key, $changes);
+		$this->cache()->set('__updated__', time());
 	}
 
 	/**
--- a/src/Content/Changes.php
+++ b/src/Content/Changes.php
@@ -94,10 +94,20 @@ class Changes
 	 */
 	public function generateCache(): void
 	{
+		$models = [
+			'files' => [],
+			'pages' => [],
+			'users' => []
+		];
+
 		foreach ($this->kirby->models() as $model) {
 			if ($model->version(VersionId::changes())->exists('*') === true) {
-				$this->track($model);
+				$models[$this->cacheKey($model)][] = (string)($model->uuid() ?? $model->id());
 			}
+		}
+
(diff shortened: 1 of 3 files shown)


src/Content/Changes.php
```diff
@@ -62,6 +69,18 @@ public function files(): Files
 		return $files;
 	}
 
+	/**
+	 * Rebuilds the cache by finding all models with changes version
+	 */
+	public function generateCache(): void
+	{
+		foreach ($this->kirby->models() as $model) {
+			if ($model->version(VersionId::changes())->exists('*') === true) {
```
distantnative, line 104: `Question`: btw do we prefer this that much over writing `$model->version('changes')`?

Especially since our code does immediately check if the provided string is a valid `VersionId` value.

src/Panel/ChangesDialog.php
```diff
@@ -18,11 +18,9 @@
  */
 class ChangesDialog
 {
-	protected Changes $changes;
-
-	public function __construct()
-	{
-		$this->changes = new Changes();
+	public function __construct(
+		protected Changes $changes = new Changes()
```
distantnative, line 22: This works for PHP 8.1+ so I thought it's a neat place to give it a first try.

bastianallgeier: OMG 😍 I don't know how many times I wanted that.

src/Content/Changes.php
bastianallgeier: I had the same thought a couple times. The strings feel "dirty" with all our shiny new type checks. But on the other hand, such lines get incredibly difficult to read. I think we only need to be consistent, but I'm up for the simpler version.

bastianallgeier (approved):
Looks good! If you want to simply that version param, feel free to add it before merging.

src/Content/Changes.php
distantnative: I'll leave it for now as it's the style used mostly at the moment. But maybe something to revisit (but then consistently everywhere).

src/Content/Changes.php
```diff
@@ -62,6 +69,18 @@ public function files(): Files
...
+				$this->track($model);
```
distantnative: @bastianallgeier Looking at this once more, this might be quite inefficient to call `::track()` for each model. Maybe this method should rather collect three arrays (files, pages, users) and then write them each with `::update()` once?

bastianallgeier: Yeah, that makes sense.

bastianallgeier (approved):
Nice!

PR #6758, opened 2024-10-17
Form Cleanup (Volume 2)

bastianallgeier:
### Summary of changes

- `Form::$fields` is no longer optional. 
- Replace `empty()` calls
- Field Siblings refactoring
  - Rename `Field::$formFields` to `Field::$siblings` for more consistency 
  - Added a new `Field::siblings()` method and deprecated `Field::formFields()`
  - Added the `HasSiblings` trait to align `Field` with `FieldClass` This also needed a new protected `Field::siblingsCollection` method
- Sorted all methods in FieldTest alphabetically and added missing test methods
- Renamed `Field::save()` to `Field::isSaveable()` to align it with the `FieldClass` 
- New `Field::isEmptyValue()` method to align it with `FieldClass` Passing an argument to `Field::isEmpty()` is now deprecated. 
- New Trait classes to remove redundancies between `Field` and `FieldClass`
  - `Kirby\Form\Mixin\Api`
  - `Kirby\Form\Mixin\Model`
  - `Kirby\Form\Mixin\Validation`
  - `Kirby\Form\Mixin\When`
- New `$model` argument for the `Kirby\Form\Fields` constructor. This will automatically inject the model into all fields.

### Reasoning

The final goal of this cleanup is to get to a point where we can construct a form and fill it with values afterwards: 

```php
$form = new Form([
  'fields' => []
]);

$form->fill([
  'a' => 'A',
  'b' => 'B',
]);
```

Why is this needed? For our models, we want to be able to set the original values for the form fields and then overwrite changed fields in a second fill call. 

```php
$form = new Form([
  'fields' => [], 
  'values' => [
    // original values
  ]
]);

$form->fill([
  // changed values
]);
```

This should not cause the entire form and every field to be reconstructed and every property to be reevaluated. Otherwise we will create massive performance issues. 

But our Form class is full of spaghetti code. We also have a lot of inconsistencies between `Field` and `FieldClass`. The combination of those "dirty" parts lead to a situation where it is basically impossible to create such a `Form::fill` method. 

### Next steps

This cleanup is not done yet. In a next step, the Form class needs to be refactored more to get rid of a very big foreach loop in the constructor, which handles all kind of weird edge cases for fields and values. We also need to build the final ::fill method and make sure that all our tests still pass and we don't introduce any form regressions. 

### Deprecated

- `Field::save()`: use `Field::isSaveable()` instead
- `Field::formFields()`: use `Field::siblings()` instead
- Passing a value to `Field::isEmpty($value)` is deprecated. Use `Field::isEmptyValue($value)` instead to check if a value qualifies as empty for this field type. `Field::isEmpty()` should be reserved to check if the current Field value is empty. 

### Breaking changes

- `Kirby\Form\Field::$formFields` has been renamed to `Kirby\Form\Field::$siblings`
- The `$formField` argument in `Kirby\Form\Field::factory()` has been renamed to `$siblings`
- `Kirby\Form\Form::fields()` does no longer return null but an empty Fields collection if there are no fields. 
- `Kirby\Form\Field::validate()` returns an array of errors instead of void. The validate method is protected and this should not really qualify as a public breaking change, but it's still worth noting it here. 

## Ready?

### For review team

### DIFF
--- a/src/Form/Field.php
+++ b/src/Form/Field.php
@@ -26,12 +26,8 @@ use Kirby\Toolkit\V;
 class Field extends Component
 {
 	use HasSiblings;
+	use HasValidation;
 	use HasWhenQuery;
-
-	/**
-	 * An array of all found errors
-	 */
-	protected array|null $errors = null;
 
 	/**
 	 * Parent collection with all fields of the current form
@@ -290,14 +286,6 @@ class Field extends Component
 	}
 
 	/**
-	 * Validates when run for the first time and returns any errors
-	 */
-	public function errors(): array
-	{
-		return $this->errors ??= $this->validate();
-	}
-
-	/**
 	 * Creates a new field instance
 	 */
 	public static function factory(
@@ -386,14 +374,6 @@ class Field extends Component
 	}
 
 	/**
-	 * Checks if the field is invalid
-	 */
-	public function isInvalid(): bool
-	{
-		return $this->isValid() === false;
-	}
-
-	/**
 	 * Checks if the field is required
 	 */
 	public function isRequired(): bool
@@ -407,14 +387,6 @@ class Field extends Component
 	public function isSaveable(): bool
 	{
 		return ($this->options['save'] ?? true) !== false;
-	}
-
-	/**
-	 * Checks if the field is valid
-	 */
-	public function isValid(): bool
-	{
-		return $this->errors() === [];
 	}
 
 	/**
@@ -477,56 +449,7 @@ class Field extends Component
 			fn ($item) => $item !== null && is_object($item) === false
 		);
 	}
-
-	/**
-	 * Runs the validations defined for the field
-	 */
-	protected function validate(): array
-	{
-		$validations = $this->validations();
-		$value       = $this->value();
-		$errors      = [];
-
-		// validate required values
-		if ($this->needsValue() === true) {
-			$errors['required'] = I18n::translate('error.validation.required');
-		}
-
-		foreach ($validations as $key => $validation) {
-			if (is_int($key) === true) {
-				// predefined validation
-				try {
-					Validations::$validation($this, $value);
-				} catch (Exception $e) {
-					$errors[$validation] = $e->getMessage();
-				}
-				continue;
-			}
-
-			if ($validation instanceof Closure) {
-				try {
-					$validation->call($this, $value);
-				} catch (Exception $e) {
-					$errors[$key] = $e->getMessage();
-				}
-			}
-		}
-
-		if (
-			empty($this->validate) === false &&
-			($this->isEmpty() === false || $this->isRequired() === true)
-		) {
-			$rules = A::wrap($this->validate);
-
-			$errors = [
-				...$errors,
-				...V::errors($value, $rules)
-			];
-		}
-
-		return $errors;
-	}
-
+	
 	/**
 	 * Defines all validation rules
 	 */
--- a/src/Form/Field.php
+++ b/src/Form/Field.php
@@ -3,7 +3,6 @@
 namespace Kirby\Form;
 
 use Closure;
-use Kirby\Cms\App;
 use Kirby\Cms\HasSiblings;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Toolkit\A;
--- a/src/Form/Field.php
+++ b/src/Form/Field.php
@@ -24,6 +24,7 @@ use Kirby\Toolkit\I18n;
 class Field extends Component
 {
 	use HasSiblings;
+	use Mixin\Api;
 	use Mixin\Validation;
 	use Mixin\When;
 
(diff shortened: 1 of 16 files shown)


bastianallgeier:
This is a big one. @getkirby/backend let me know if this is still ok to review, if you need anything else from me in the description or if I should break it down into smaller steps. I tried to do that with the commits and keep each one mostly focused on one particular task. I hope it helps.

src/Form/Field.php
```diff
@@ -50,7 +52,7 @@ class Field extends Component
 	public function __construct(
 		string $type,
 		array $attrs = [],
-		Fields|null $formFields = null
+		Fields|null $siblings = null
```
distantnative, line 55: `Todo:` We need to list the argument name change as breaking change.

src/Form/Field.php
```diff
@@ -290,16 +274,16 @@ public function drawers(): array
 	public static function factory(
 		string $type,
 		array $attrs = [],
-		Fields|null $formFields = null
+		Fields|null $siblings = null
```
distantnative, line 277: `Todo:` List as breaking change.

src/Form/Field.php
```diff
@@ -323,23 +307,11 @@ public function fill(mixed $value): static
 	}
 
 	/**
-	 * Parent collection with all fields of the current form
+	 * @deprecated Use `::siblings() instead
```
distantnative:
```suggestion
	 * @deprecated  5.0.0 Use `::siblings() instead
```

src/Form/Field.php
```diff
@@ -352,6 +324,7 @@ public function isDisabled(): bool
 
 	/**
 	 * Checks if the field is empty
+	 * @deprecated Passing arguments is deprecated. Use `::isEmptyValue()` instead to check for
```
distantnative:
```suggestion
	 * @deprecated 5.0.0 Passing arguments is deprecated. Use `::isEmptyValue()` instead.
```

src/Form/Field.php
```diff
@@ -360,6 +333,14 @@ public function isEmpty(mixed ...$args): bool
 			default => $args[0]
 		};
 
+		return $this->isEmptyValue($value);
+	}
+
+	/**
+	 * Checks if the given value is considered empty
+	 */
```
distantnative, line 343:
```suggestion
   * @since 5.0.0
	 */
```

src/Form/Field.php
```diff
@@ -392,77 +365,51 @@ public function isRequired(): bool
...
+			return $this->options['api']->call($this);
+		}
+
+		return [];
 	}
 
 	/**
-	 * Returns the parent model
+	 * Checks if the field is saveable
+	 * @deprecated Use `::isSaveable()` instead
```
distantnative:
```suggestion
	 * @deprecated 5.0.0 Use `::isSaveable()` instead
```

src/Form/Mixin/When.php
```diff
@@ -0,0 +1,80 @@
...
+			$this->isSaveable() === false ||
+			$this->isRequired() === false ||
+			$this->isEmpty() === false ||
+			$this->isActive() === false
+		) {
+			return false;
+		}
+
+		return true;
+	}
```
distantnative, line 63: Really unsure if this shouldn't be in the validation mixin. It's somewhat part of both.

bastianallgeier: I'm working on additional steps with a new Value mixin. I think it should actually go in there later. Should I move it back to the classes for this PR or do you think it's fine to move it to a different Mixin later?

distantnative: It's fine to have it in the mixing and move it later.

distantnative:
@bastianallgeier something broke but it seems like already on `v5/changes/develop` - I still think we need to fix it here/first as it makes testing this PR hard:

1. Check out lab
2. Edit `site/blueprints/pages/text-field.yml`
3. Add when conditions to the two validation fields:

```yml
     required:
        type: text
        required: true
        when:
          text: "foo"
      minlength:
        type: text
        minlength: 10
        when:
          text: "foo"
```

4. Go the page in channel. Fields will be hidden as long as the `text` field value isn't `foo` ☑️
5. Then you can also type in anything but `foo` and save, it gets saved as the validations of the hidden fields are rightfully ignored ☑️
6. Type in `foo` to the text field
7. The fields in the validations tab show up  ☑️
8. Try to save
9. No error message is shown instead of an error message that the `required` field has no value but is required ❌

bastianallgeier:
@distantnative I started by adding a new when query test setup to the sandbox. https://github.com/getkirby/sandbox/commit/76bd69b848b9a63473d39995967714c4a1ed7f7c

bastianallgeier:
Turns out: we don't run any form of validation on v5/changes/develop so far 🙈

It's really not related to this PR though.

distantnative:
@bastianallgeier 🙈

But I think we need to fix it first or else we won't know if this PR leaves everything else functioning. I know you don't want this to become stuck but it's hard to test without.

bastianallgeier:
I agree. I will create a new PR to fix this.

distantnative (approved):

PR #6759, opened 2024-10-17
New blueprint cache

afbora:
## Description

Implements https://feedback.getkirby.com/328

I created the PR without any concept because I thought it was a simple development. If it is not as simple as I thought or does not fit into our future plans (especially regarding the blueprint editor feature), feel free to close the PR 👍 I'll add unit tests if PR is acceptable.

````php
<?php

return [
    'cache' => [
        'blueprints' => [
            'active' => true,
            'ignore' => ['site', 'pages/notes']
        ]
    ],
];
````

````yaml
title: Simple Page

# cache duration as minutes and default is 0
# never expires if blueprint cache is enabled with no option set
# so developer should be delete manually the blueprint caches
cache: 10
````

### Features

- New blueprint cache

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Blueprint.php
+++ b/src/Cms/Blueprint.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Cms;
 
+use Closure;
 use Exception;
 use Kirby\Data\Data;
 use Kirby\Exception\InvalidArgumentException;
@@ -55,19 +56,36 @@ class Blueprint
 			);
 		}
 
+		$cache       = null;
+		$kirby       = App::instance();
 		$this->model = $props['model'];
 
 		// the model should not be included in the props array
 		unset($props['model']);
+
+		// normalize the name
+		$props['name'] ??= 'default';
+
+		// try to get the blueprint from cache
+		if (static::isCacheable($props['name']) === true) {
+			$cache = $kirby->cache('blueprints');
+
+			if ($blueprint = $cache->get($props['name'])) {
+				$this->tabs     = $blueprint['tabs'];
+				$this->sections = $blueprint['sections'];
+				$this->fields   = $blueprint['fields'];
+				unset($blueprint['fields'], $blueprint['sections']);
+
+				$this->props = $blueprint;
+				return;
+			}
+		}
 
 		// extend the blueprint in general
 		$props = static::extend($props);
 
 		// apply any blueprint preset
 		$props = $this->preset($props);
-
-		// normalize the name
-		$props['name'] ??= 'default';
 
 		// normalize and translate the title
 		$props['title'] ??= ucfirst($props['name']);
@@ -82,6 +100,17 @@ class Blueprint
 		$props['tabs'] = $this->normalizeTabs($props['tabs'] ?? []);
 
 		$this->props = $props;
+
+		// cache the blueprint
+		if ($cache !== null) {
+			// store fields, sections, and tabs in the cache
+			// to set these properties when the cached blueprint is loaded
+			$props['fields']   = $this->fields;
+			$props['sections'] = $this->sections;
+			$props['tabs']     = $this->tabs;
+
+			$cache->set($props['name'], $props, $props['cache'] ?? 0);
+		}
 	}
 
 	/**
@@ -420,6 +449,34 @@ class Blueprint
 	protected function i18n(mixed $value, mixed $fallback = null): mixed
 	{
 		return I18n::translate($value, $fallback) ?? $value;
+	}
+
+	/**
+	 * Checks if the blueprint can be cached in the blueprints cache.
+	 * This will also check if one of the ignore rules from the config kick in.
+	 */
+	public static function isCacheable($name): bool
+	{
+		$kirby   = App::instance();
+		$cache   = $kirby->cache('blueprints');
+		$options = $cache->options();
+		$ignore  = $options['ignore'] ?? null;
+
+		// check for a custom ignore rule
+		if ($ignore instanceof Closure) {
+			if ($ignore($name) === true) {
+				return false;
+			}
+		}
+
+		// ignore blueprints by name
+		if (is_array($ignore) === true) {
+			if (in_array($name, $ignore) === true) {
+				return false;
+			}
+		}
+
+		return true;
 	}
 
 	/**

### CODE REVIEW

**Comment by bastianallgeier** [comment]
I love how simple the implementation is. Did you see big differences in the performance afterwards? I did a few tests and I wasn't 100% able to tell if it's that much faster to be honest. I wonder if our main bottle neck is really in the evaluation of the blueprints or actually in the architecture of the forms. But I think the biggest difference is probably when you have a lot of blueprint extensions. I haven't tested that yet.

**Comment by afbora** [comment]
Yes, this needs complex blueprints setup to test it. I haven't see %100 performance yet either. I've asked the setup from @bnomei and @mynameisfreedom

**Comment by distantnative** [comment]
What if the cache would also store the modified time? And instead of an option to define how long valid, we could test the cached modified time against the actual modified time and discard it if the time differs.

Or maybe the modified check takes up too much performance itself?

**Comment by afbora** [comment]
@distantnative For example, for the project.yml blueprint; a block blueprint inside a block field inside a tab was updated. In order to check whether project.yml has been updated, it is necessary to scan all blueprints and compare their modification dates. IMHO yes, this would create a performance issue.

This new caching system caches blueprints that have been already parsed, prevents the need to scan and parse all blueprints over and over again.

**Comment by distantnative** [comment]
Got you!

**Comment by afbora** [comment]
This clearing caches process will be easier when the blueprint editor feature comes. When a blueprint is edited/updated from the panel blueprint editor, we can easily clear the cache 👍

**Comment by bastianallgeier** [comment]
I think we can compare caching blueprints a lot to caching pages. This would be something that you do in production, but not in development. I was wondering if we could even tell people that the cache needs to be flushed manually on deployment.

**Comment by afbora** [comment]
This makes sense 👍 Blueprints are typically updated during the development phase. They can also be disabled the blueprints cache during the development phase or in the environment.

Another idea is maybe we can set up a manual version for developers:

````php
<?php

return [
    'cache' => [
        'blueprints' => [
            'active' => true,
            
            // can be a date
            'lastModified' => '2024-10-18 12:44',
            
            // or can be a version
            'changesVersion' => 27            
        ]
    ],
];
````

**Comment by bastianallgeier** [comment]
Uh, I like that. We could even think about adding this option to all cache drivers. Could be quite useful in other places as well.

**Comment by mynameisfreedom** [comment]
Hi @afbora! 

Did you find a complex blueprint example? 
If you give me your email I will send you the Zero One theme, which has a very complex Layout Builder. So you can test right on it. 

P.S. Just for the info... In the new Zero One version (which should be released next month) I accomplished a speed-up of around 30% by rewriting the Layout Builder blueprint and Subgrid block blueprint in PHP. But if the blueprint cache works it would be tremendous! 🚀

**Comment by bnomei** [comment]
@afbora my sakila-kit will enable you to setup blueprints that query a lot of data.

1) Caching API response used in a blueprint statically. like an aggregated string: https://github.com/bnomei/kirby-sakila-kit/blob/4ed80a0846e965601766356ff62a9cc7a1839211/site/blueprints/pages/actor.yml#L8

```yml
blueprint: page

fields:
  films:
    type: info
    label: Films
    text: >
      {< page.filmsByActor() >}
```

2) Cache some options from an API (if the current implementation caches that)

```yml
blueprint: page

fields:
  films:
    type: select
    label: Films
    options:
      type: query
      query: site.find('film').children
```

3) Cache of "unrolled" fields. While the blueprint might look simple the actual data that kirby stores for a blueprint is much more detailed (like default values and other props). so "unrolling" a large blueprint to these takes time. this also related to blueprint that have lots of albeit simple blocks... all bigger themes (like zero-one).

here is a large blueprint: https://gist.github.com/bnomei/a1365c804482a8ea517d9ce2d6451d07

afbora:
Thanks for the setup @bnomei But it doesn't matter how many fields there are. It matters how many extends there are. The goal is to cache one blueprint and read only that instead of loading dozens of different blueprints.

Remember, this cache system does not cache queries or dynamic values ​​in the blueprint.

@mynameisfreedom That would be great. You can send it to ahmet@getkirby.com or support@getkirby.com 👍

mynameisfreedom:
Hi @afbora! Just sent on [ahmet@getkirby.com](mailto:ahmet@getkirby.com)!

afbora:
I tested it on @mynameisfreedom's theme (thanks for that 🙌) but I didn't see any visible performance increase. I think the real problem is not the blueprint caches but the API queries in blueprints (dynamic values, options, loading sections, form fields, nested loads, etc..) and our form class. Also we already knew about the big performance problem especially for nested layout and blocks fields. 

Sorry for the lack of PR :(

mynameisfreedom:
Ah, sorry to hear that @afbora 🙁 

I remember one thing, maybe helps, because Zero One page builder was initially created with an old [Kirby Builder](https://github.com/TimOetting/kirby-builder) plugin, when migrated to Kirby blocks it was 2-3 times slower. Something was different. 

Also, I saw Python folks complain about yaml performance. Many migrated to JSON. 

There is definitely performance improvements when I rewrite block blueprints to PHP, although I rewrote only the two most problematic ones. When I find time, as a temporary performance solution, I will rewrite all block blueprints to PHP, and get back with performance tests. If it means anything to you...

mynameisfreedom:
> It matters how many extends there are.

Are you saying here that if I don't extend/separate blueprint into multiple blueprints, but keep it in one, it will be faster?

mynameisfreedom:
> Thanks for the setup @bnomei But it doesn't matter how many fields there are. It matters how many extends there are. The goal is to cache one blueprint and read only that instead of loading dozens of different blueprints.


Are you saying here that if I don't extend blueprints in many files, but keep them as one, it will be faster?

afbora:
> Are you saying here that if I don't extend blueprints in many files, but keep them as one, it will be faster?

Yes and no. What I mean is that it will need to make 101 queries for a blueprint that contains 100 separate blueprints. The blueprints cache did not do anything significant. The reason for the slowness in your theme is not the blueprints, but the nested layout and block fields. I'll see what I can do about it.

mynameisfreedom:
Ah, I think I understand... If I add all extends into one blueprint it will improve performance but not too much. Great, feel free to test anything on Zero One. Any panel performance improvement or a hint for different structuring is highly appreciated, thank you 🙏

PR #6765, opened 2024-10-19
Languages dropdown: badge for translation changes

distantnative:
### Summary of changes
- Shows badge on languages dropdown button whether any translations have unsaved changes

### Enhancement
- Panel: Languages dropdown indicates whether any translations have unsaved changes

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Layout/Tabs.vue
+++ b/panel/src/components/Layout/Tabs.vue
@@ -183,6 +183,6 @@ export default {
 
 .k-tab-button .k-button-badge {
 	top: 3px;
-	inset-inline-end: calc(var(--button-padding) / 2);
+	inset-inline-end: calc(var(--button-padding) / 4);
 }
 </style>
--- a/panel/src/components/Navigation/Button.vue
+++ b/panel/src/components/Navigation/Button.vue
@@ -381,11 +381,13 @@ export default {
 	position: absolute;
 	top: 0;
 	inset-inline-end: 0;
-	transform: translate(50%, -20%);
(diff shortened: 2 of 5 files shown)


bastianallgeier:
This is really cool. For some reason, I don't get it to show the real count of changed languages though.

distantnative:
@bastianallgeier it doesn't count the current viewed into. Only how many other languages have changed. Does that explain what you see?

bastianallgeier:
Oh, yes. That explains it. But for my taste, it would need to include the current one as well. Especially when we show the changes in the dropdown later too. I think if you open the dropdown and two languages show changes in there, but the badge only says 1 it would feel wrong.

distantnative:
I didn't include it in the beginning as new changes to the current view wouldn't be reflected from what was loaded from the backend. But I think hat we could solve by sending the same number as now from the backend and adding `+1` if the current view has `$panel.content.hasChanges`.

It's still a bit distracting visually that so many things suddenly change when there are no changes at all and one adds the first change to the current view, but I understand that the PR as is right now is also confusing.

bastianallgeier:
Without wanting to create extra work, maybe we should take a step back here. What happens for example when you are in a view with tabs and the tabs each also have badges with counters? I was just wondering if we could use a badge without showing a number at all. Maybe just the badge or a badge with a tiny edit icon or something. Omitting the numbers could help to simply indicate "Something has changed here" and let the dropdown then have more information.

bastianallgeier (approved):
I really like it with the minimal style!

PR #6768, opened 2024-10-22
Add the plugin as required argument for the license object

bastianallgeier:
## Description

@nilshoerrmann reported that the plugin instance might often be a very important dependency that's needed in a custom License class to return the proper directory for the license file for example. 

This PR adds the Plugin instance as required first argument to the License object and also changes the way that license instances can be added to a plugin. 

### Summary of changes

A custom license class has to have the plugin instance as first argument of the constructor. 

```php
use Kirby\Plugin\Plugin;
use Kirby\Plugin\License;

class MyLicense extends License
{
  public function __construct(
    protected Plugin $plugin
  ) {
  }  
}
```

A license can no longer be passed directly as instance, but has to be wrapped in a closure to inject the plugin instance. 

```php 
Kirby::plugin(
  name: 'my/plugin', 
  extends: [...]
  license: function ($plugin) {
    return new MyLicense(
      plugin: $plugin
    );
  }
)
```

Nothing changed about passing a license as string or array.

## Ready?

### For review team

### DIFF
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -870,7 +870,7 @@ trait AppPlugins
 		array $info = [],
 		string|null $root = null,
 		string|null $version = null,
-		License|string|array|null $license = null,
+		Closure|string|array|null $license = null,
 	): Plugin|null {
 		if ($extends === null) {
 			return static::$plugins[$name] ?? null;
--- a/src/Plugin/License.php
+++ b/src/Plugin/License.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Plugin;
 
+use Closure;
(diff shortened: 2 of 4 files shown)


distantnative (approved):

nilshoerrmann:
Trying this out, I get the following error:

> `Kirby\Cms\App::plugin()`: Argument #4 (`$root`) must be of type `?string`, `Closure` given

grafik

bastianallgeier:
Damn, I missed the type hinting there. I just fixed it. Can you give it another try?

nilshoerrmann:
I think it's working now (encountered another error – but that one's on our end) :+1:

PR #6770, opened 2024-10-22
New Changes Comparison View

bastianallgeier:
### Summary of changes

- New `k-preview-view` component
- New `/pages/(:any)/preview` route/view
- New `/site/preview` route/view
- New icons
  - `asterisk`
  - `layout-columns`
  - `layout-right`
  - `layout-left`
  - `expand-horizontal`
  - `collapse-horizontal`
  - `window`
- Responsive form control buttons
- New `PreviewDropdownButton` class
- new `copyToClipboard` event handler in the event.js which can be triggered like this: 
```js 
panel.events.emit('copyToClipboard', 'Copy this text to the clipboard');
```

### Features

- New changes view to compare your edits with the latest saved version before you save them

Screenshot 2024-11-04 at 16 55 13 
Screenshot 2024-11-04 at 17 01 48
Screenshot 2024-11-04 at 17 01 33

## Ready?

### For review team


panel/src/components/Views/Pages/PageChangesView.vue
```diff
@@ -0,0 +1,180 @@
...
+					icon: "layout-left",
+					current: this.mode === "changes",
+					click: () => (this.mode = "changes")
+				},
+				compare: {
+					label: this.$t("version.compare"),
+					icon: "layout-columns",
+					current: this.mode === "compare",
+					click: () => (this.mode = "compare")
+				}
```
distantnative: `Suggestion:` I'd flip the order, to have the compare mode on top as the default one and then the more specific ones further below.

i18n/translations/en.json
```diff
@@ -748,10 +748,13 @@
 	"users": "Users",
 
 	"version": "Version",
+	"version.changes": "Changed version",
+	"version.compare": "Compare versions",
```
distantnative, line 754: `Suggestion:`

```suggestion
	"version.compare": "Comparing versions",
```

panel/src/components/Views/Pages/PageChangesView.vue
```diff
@@ -0,0 +1,180 @@
...
+					:text="$t('back')"
+					icon="angle-left"
+					size="sm"
+					variant="filled"
+				/>
+				<k-button
+					:icon="modes[mode].icon"
+					:dropdown="true"
+					:responsive="true"
+					:text="$t('view')"
```
distantnative: `Suggestion:` I find "View" very undescriptive and would prefer to show the currently selected mode here.

panel/src/components/Views/Pages/PageChangesView.vue
```diff
@@ -0,0 +1,180 @@
...
+					size="sm"
+					variant="filled"
+					@click="$refs.layout.toggle()"
+				/>
```
distantnative: `Suggestion:` Center this button to make room for title on the left.

panel/src/components/Views/Pages/PageChangesView.vue
```diff
@@ -0,0 +1,180 @@
...
+		<header class="k-page-changes-header">
+			<k-button-group>
+				<k-button
+					:link="link"
+					:responsive="true"
+					:text="$t('back')"
+					icon="angle-left"
+					size="sm"
+					variant="filled"
+				/>
```
distantnative: `Suggestion:` When the frontend page itself isn't great at indicating which one it is etc., it's a little hard to know where you're at. I think we could draw from the previous preview mode experiment and display the title of the current page here, maybe even with the dropdown tree navigator from that experiment. Even without it, I think this would help to give more orientation and structure to this view.

config/areas/site/views.php
```diff
@@ -8,6 +8,25 @@
 		'pattern' => 'pages/(:any)',
 		'action'  => fn (string $path) => Find::page($path)->panel()->view()
 	],
+	'page.changes' => [
+		'pattern' => 'pages/(:any)/changes',
+		'action'  => function (string $path) {
+			$page = Find::page($path);
+			$view = Find::page($path)->panel()->view();
```
distantnative:
```suggestion
			$view = $page->panel()->view();
```

panel/src/components/Views/Pages/PageChangesView.vue
```diff
@@ -0,0 +1,180 @@
...
+		}
+	},
+	methods: {
+		async onDiscard() {
+			if (this.isLocked === true) {
+				return false;
+			}
+
+			await this.$panel.content.discard();
+			await this.$panel.view.open(this.link);
```
distantnative: `Thoughts:` I didn't find it the most pleasant UX to be thrown out of the preview just because of saving/discarding changes. Was a bit like "woa, what's happening, where am I?". So I was thinking if that's necessary or we could just remain on the preview view and letting the user decide what to do next? On the other hand there is not much to preview without any changes, so I also do understand the thought of adding it. What's your feeling with it? Does it feel good that with the button press also the view dissolves?

panel/src/components/Views/Pages/index.js
```diff
@@ -1,9 +1,11 @@
 import PageView from "./PageView.vue";
+import PageChangesView from "./PageChangesView.vue";
 import SiteView from "./SiteView.vue";
 
 export default {
 	install(app) {
 		app.component("k-page-view", PageView);
+		app.component("k-page-changes-view", PageChangesView);
```
distantnative: `Todo:` We also need it for the site view.

panel/src/components/Views/Pages/PageChangesView.vue
bastianallgeier: I've tried that and found it confusing with the headlines below basically repeating the same. I also tried hiding the headlines and only showing them in the side-by-side view, but that also felt weird.

panel/src/components/Views/Pages/PageChangesView.vue
bastianallgeier: I totally forgot about that. That's a great idea.

panel/src/components/Views/Pages/PageChangesView.vue
bastianallgeier: Hm, I really never thought about keeping people in the view after saving/discarding. It would be indeed a pretty useless place to stay to be honest.

panel/src/components/Views/Pages/index.js
bastianallgeier: I was wondering what we do with that and also for files. Would it be confusing if we show the home page in site view and the parent page for files? Or is it more confusing if we don't have the comparison view for those at all.

distantnative: I'd say for site, yes the home page - as where the normal preview button also sends the user. I think for files and users it's ok to not have one as it's difficult to pick one that really makes sense in most setups.

panel/src/components/Views/Pages/PageChangesView.vue
distantnative: Maybe, just a thought, taking inspiration from the previous experiment would work also here. If we move this button to the center to make space for the page title, we could try if instead of button with dropdown, we just have the three icons as a toggle, similar to the view size toggle we had in the other experiment:

Screenshot 2024-10-22 at 4 13 21 PM

I could imagine that it doesn't actually need the text/headlines/labels at all and just the three icons next to each other make it quite clear what they actually do. Especially since then each section is labeled and self-explaining, as you say.

bastianallgeier:
@distantnative I played around a lot after your suggestions and came up with a mixture of the old live preview view, your suggestions and my own version. 

Screenshot 2024-11-04 at 16 55 13

- The title is back and I think it's really important
- I moved the form controls above the changes iframe and I think it really works great to highlight the changes even more and give a better connection to the related actions a user can take.  
- I've added expand/collapse toggles for each version. In my opinion, they work great as a shortcut to switch between the different views. I decided to stick with the additional dropdown in the top right corner instead of using the toggles input. My main reason would be that we can add more options to it later (mostly live editing) and the labels in the dropdown really help to clarify the views some more. 
 
Screenshot 2024-11-04 at 17 01 48
Screenshot 2024-11-04 at 17 01 33

distantnative:
@bastianallgeier What do you think about including the page tree dropdown for the title as in the old experiment? Or too much for this?

bastianallgeier:
I did that at first and then took it out again. It's awesome, but I think we need to get the basics right first.

afbora:
I loved the feature, almost perfect 😍 

Some points for first test:
- Preview changes doesn't work for draft pages
- New keyboard shortcut would be great for preview action
- You'll see following console error after open preview changes and back to panel page
> TypeError: Cannot read properties of undefined (reading 'toggle')

bastianallgeier:
@afbora thanks for testing. I just fixed the toggle issue. I cannot really reproduce the draft problem though. What kind of error do you see? Shortcuts would be nice indeed.

afbora:
@bastianallgeier Thank you. Please follow the steps:

- Update a draft page without save
- Preview the changes
- See that both side is same

distantnative:
I can reproduce the draft problem. Trying to find out where it's going wrong.

EDIT: found it. The URLs created for the preview are `?token=XXX?_version=changes`

So double `?` instead of a connecting `&` for the two query parameters.

distantnative:
We will need a pretty version of

```php
$changes = match (parse_url($preview, PHP_URL_QUERY)) {
	null    => $preview . '?_version=changes',
	default => $preview . '&_version=changes'
};
```

config/areas/site/views.php
```diff
@@ -14,6 +14,28 @@
...
+			$preview = $page->previewUrl();
+
+			return [
+				'component' => 'k-preview-view',
+				'props'     => [
+					...$view['props'],
+					'back' => $view['props']['link'],
+					'mode' => $mode,
+					'src'  => [
+						'changes' => $preview . '?_version=changes',
```
distantnative: `Todo:` Doesn't work for draft URLs that already include the `?token=` query. See general discussion.

config/areas/site/views.php
```diff
@@ -14,6 +14,28 @@
...
+						'latest'  => $preview,
+					]
+				],
+				'title' => $view['props']['title'] . ' | ' . I18n::translate('changes'),
```
distantnative, line 43: `Todo:` Add `use Kirby\Toolkit\I18n` line

config/areas/site/views.php
```diff
@@ -24,4 +46,26 @@
...
+			$preview = $site->previewUrl();
+
+			return [
+				'component' => 'k-preview-view',
+				'props'     => [
+					...$view['props'],
+					'back' => $view['props']['link'],
+					'mode' => $mode,
+					'src'  => [
+						'changes' => $preview . '?_version=changes',
```
distantnative: `Todo:` same here for draft pages.

i18n/translations/en.json
```diff
@@ -18,9 +18,11 @@
 	"collapse.all": "Collapse All",
 	"color": "Color",
 	"coordinates": "Coordinates",
+	"copied": "Copied",
```
distantnative: `Problem:` I think the naming is not so great if we also have `copy.success` with very similar content. In theory, the new line should be `copy.success` and the old `copy.succcesses` or so.

panel/src/components/Views/PreviewView.vue
```diff
@@ -0,0 +1,227 @@
...
+				<k-button
+					:link="back"
+					:responsive="true"
+					:title="$t('back')"
+					icon="angle-left"
+					size="sm"
+					variant="filled"
+				>
+				</k-button>
+				<k-button icon="title" element="span">
```
distantnative, line 14: `Question:` We are using `k-button` here for the styling mainly? (as forcing to span kinda remove the functionality purpose, I'd think)

panel/src/panel/events.js
```diff
@@ -33,6 +33,15 @@ export default (panel) => {
...
 	emitter.on("keydown.cmd./", () => panel.search());
 
+	/**
+	 * Custom copy to clipboard event
+	 * @since 5.0.0
+	 */
+	emitter.on("copyToClipboard", async (e) => {
+		navigator.clipboard.writeText(e);
+		panel.notification.success(panel.t("copied") + "!");
+	});
```
distantnative, line 44: `Discuss:` I'm wondering if this is the best as global event or if rather DropdownItem/DropdownContent should handle this? 

Also, shouldn't we use `$helper.clipboard.write()` then? Or adapt the helper method to this code?

i18n/translations/en.json
bastianallgeier: I've refactored it to use `copy.success` and `copy.success.multiple`

panel/src/components/Views/PreviewView.vue
bastianallgeier: It's the easiest solution to get the icon and I was also thinking that it is a good preparation to turn it into the tree dropdown.

panel/src/panel/events.js
bastianallgeier: Woah, I completely forgot about the clipboard helpers 🙈 I just replaced that part. I'm also not entirely sure about the event. We could maybe provide some handy action shortcuts and copy could be one of them. But we also have to be careful that we don't make this all too complicated.

config/areas/site/views.php
bastianallgeier: I've fixed it by adding a new VersionId option to the `Site::previewUrl` and `Page::previewUrl` methods.

src/Cms/Page.php
```diff
@@ -929,26 +930,31 @@ public function permissions(): PagePermissions
...
 
-		$url = match ($preview) {
-			true    => $this->url(),
-			default => $preview
+		$url = match ($url) {
+			true, null => $this->url(),
+			default    => $url
 		};
 
+		$uri = new Uri($url ?? $this->url());
```
distantnative: I don't think it needs the fallback as after the match statement, there is no way `$url` is `null`, right?

panel/src/panel/events.js
distantnative: Thinking about it some more, the global event might be good. Can then also be used by e.g. dialog success events. But how about simplifying the name to just `clipboard`?

afbora:
Only I can found:

- `*/preview/compare` still accessible even if `preview: false`

### 💡  Ideas:

- Mobile preview feature (maybe via simple toggle) also would be great (not for now, for future)
- I think new button would be great for `Changed version` side. Something like `Continue to editing` or `ESC` shortcut. Otherwise I need to click each time to back arrow in top left of the window and it's so far away 😅

panel/src/panel/events.js
bastianallgeier: Maybe we could do `clipboard.write` and then there's still an open slot for `clipboard.read` later?

bastianallgeier:
@afbora I've fixed the permission issue and added a new ESC shortcut to leave the preview. 

The idea for mobile test buttons is definitely cool. Especially as soon as we implement live-editing.

panel/src/panel/events.js
bastianallgeier: I've already renamed it to clipboard.write. If you don't like it, we can still change it.

afbora:
If the `preview` option is an absolute url and iframe is blocked (content security policy stuff) for the absolute url, then preview doesn't work, right? If yes, so the documentation may provide a warning for this case and a hint on what to do as a solution.

bastianallgeier:
@afbora yes, I think that's still a problem for now. I wonder if we can catch this somehow and maybe show a notification directly in the view.

afbora:
````js
fetch(this.src.latest)
    .then(resp => {
        const csp = resp.headers.get('Content-Security-Policy');
        console.log(csp);
    });
````

Maybe we can check via something like above codes. But I'm not sure if it's worth it. I just meant to provide a warning and solution in the our documentation pages.

bastianallgeier:
@afbora I thought about it some more and I would suggest that we rather leave it as it is and then react to edge cases when they happen.

afbora (approved):
Looks great for me 👍  Lets release in alpha release to allow developers to test it and get it feedback from them.

PR #6772, opened 2024-10-22
Changes: fix locking

distantnative:
### Summary of changes
- `Lock` class does not just check default translation, but looks in all translations' changes for `lock` field

## Ready?


bastianallgeier:
I think this is no longer needed if we move forward with the VersionRules PR.

PR #6774, opened 2024-10-23
Version Rules

bastianallgeier:
### Summary of changes

- New `VersionRules` class with checks for all modification actions in the `Version` class
- New `Version::lock` and `Version::isLocked` methods
- Move `Storage::ensure` and `Version::ensure` to `VersionRules::ensure`
- Adjusted unit tests after first set of stricter rules

## Ready?

### For review team


src/Content/Version.php
```diff
@@ -195,6 +202,41 @@ public function id(): VersionId
...
+
+		return $this->lock($language)->isLocked();
+	}
+
+	/**
+	 * Returns the lock object for the version
+	 */
+	public function lock(Language|string $language = 'default'): Lock
+	{
+		return Lock::for($this, Language::ensure($language));
```
distantnative: `Lock::for()` currently doesn't take a second parameter.

src/Content/Version.php
```diff
@@ -319,17 +376,14 @@ protected function prepareFieldsForContent(
...
-		if ($this->id->is(VersionId::latest()) === true) {
-			throw new LogicException(
-				message: 'This version is already published'
-			);
-		}
-
 		$language = Language::ensure($language);
 
 		// the version needs to exist
 		$this->ensure($language);
```
distantnative: `Question:` Shouldn't these also belong into the rules class?

bastianallgeier: I thought about this as well. I really think they could and maybe even should.

src/Content/Version.php
bastianallgeier: ☑️

src/Content/Version.php
bastianallgeier: I moved it over from the Version and Storage classes and I think it really cleans up the storage classes a lot more.

src/Content/Lock.php
```diff
@@ -34,8 +35,11 @@ public function __construct(
...
 	 */
 	public static function for(
-		Version $version
+		Version $version,
+		Language|string $language = 'default'
 	): static {
+		$language = Language::ensure($language);
+
 		// if the version does not exist, it cannot be locked
 		if ($version->exists() === false) {
```
distantnative: Wouldn't this also need to receive the language?

distantnative: Nevermind next commit added it.

tests/Content/PlainTextStorageTest.php
```diff
@@ -378,19 +377,6 @@ public function testReadLatestSingleLang()
...
-	 */
-	public function testReadWhenMissing()
-	{
-		$this->setUpSingleLanguage();
-
-		$this->expectException(NotFoundException::class);
-		$this->expectExceptionMessage('Version "changes" does not already exist');
-
-		$this->storage->read(VersionId::changes(), Language::single());
-	}
```
distantnative, line 392: `Question:` What's now returned when calling `$storage->read()` when non existent?

bastianallgeier: The storage can throw its own exception in this case. But since we wrap the storage methods anyway and you never interact with them directly, the same exception is simply thrown one level up before ->storage()->read() is even called.

distantnative: I understand how it is working in our use case with version. But I think we need to define what's the expected case for the storage handler method here (return null, exception, ?) and then have a test for it. Cause with removing the test completely, we leave it unsure what's the behavior of a storage handler for this case.

src/Content/Version.php
```diff
@@ -195,6 +187,41 @@ public function id(): VersionId
 		return $this->id;
 	}
 
+	/**
+	 * Checks if the version is the latest version
+	 */
+	public function isLatest(): bool
+	{
+		return $this->id->is(VersionId::latest());
```
distantnative: `Suggestion`: Since `VersionId::is()` is casting it to a string anyways

```suggestion
		return $this->id->is('latest');
```

tests/Content/PlainTextStorageTest.php
bastianallgeier: I think that the type hinting in the read method is also indicating that the storage handler should always return an array. If it can't provide that array, it can throw an exception to indicate that something went wrong.

distantnative: > I think that the type hinting in the read method is also indicating that the storage handler should always return an array. 

Exactly, but ours currently don't, e.g. `MemoryStorage::read()` runs into an error as it returns `null` but type hints for an array, while `PlainTextStorage::read()` returns an empty array for the same situation. I don't think we can leave it like this but need to ensure a consistent behavior and also have tests for that.

bastianallgeier: Ok, I went over the code again and removed the Exception annotations and decided to keep MemoryStorage and PlainTextStorage consistent by always returning an array. 

We have `exists` here to run additional checks if necessary and otherwise it's probably the easiest and cleanest option to just always expect an array

distantnative (approved):

PR #6776, opened 2024-10-24
Form Cleanup Vol. 3

bastianallgeier:
### Summary of changes

- The Form class now passes the model to the Fields collection
- The Form class now uses `Field::isSaveable` instead of `Field::save` consistently. 
- New `Fields::findByKey` and `Fields::findByKeyRecursive` to find fields within fields (e.g. a field in a structure field) This has been handled by the `Form::field` method before, but it's the wrong place to do that. With the new collection methods, this is handled in a very similar fashion as we do for pages. 
- Removed an unnecessary check for a blueprint method in the `Form::for` method. ModelWithContent has a blueprint method by default and we already check for ModelWithContent child classes with the argument type. 
- A new `Value` trait class to handle all value-related tasks. This brings `Field` and `FieldClass` even closer together.
  - `$field::data()` (deprecated)
  - `$field::default()`
  - `$field::isEmpty()`
  - `$field::isEmptyValue()`
  - `$field::needsValue()`
  - `$field::toFormValue()`
  - `$field::toStoredValue()`
  - `$field::value()`
  - `$field::valueFromJson()`
  - `$field::valueFromYaml`
  - `$field::valueToJson`
  - `$field::valueToYaml`
  - Fixed query support for default values: https://github.com/getkirby/kirby/issues/2814
- New consistent concept for `$field::toStoredValue()` and `$field::toFormValue()` We have this difference between the values that we need to pass forward to the vue components and the values that we want to store in the text files. So far, there was a lot of chaos around those methods and their names weren't clear. `Field`, `FieldClass`, `Fields` and `Form` now implement `::toStoredValue(s)()` and `::toFormValue(s)()` methods to get rid of the chaos. Old `::data()` and `::value(s)()` methods are either already replaced and deprecated or will be in the next steps. 
- Improved prop unsetting in the `Kirby\Toolkit\Component` class. When we overwrite an existing prop in a child component, the class property wasn't unset properly. This only started to throw issues with better tests and type hinting after the last refactoring jobs. 
- New `Translatable` trait class to handle the `'translate'` state for fields consistently. This is supposed to help us further in the next steps when we can find and filter translatable or untranslatable fields more efficiently in multi-language setups. So far, disabling untranslatable fields is still handled by the `Form` class, which is mostly spaghetti code. 
- Use `$fail` parameter also for `Data::read()`

## Ready?

### For review team


src/Form/Field.php
```diff
@@ -28,8 +28,12 @@ class Field extends Component
 	use HasSiblings;
 	use Mixin\Api;
 	use Mixin\Model;
+	use Mixin\Translatable;
 	use Mixin\Validation;
 	use Mixin\When;
+	use Mixin\Value {
+		isEmptyValue as protected isEmptyValueFromTrait;
```
distantnative: `Question:` What's the purpose of this?

bastianallgeier: It's outdated. I just removed it.

src/Form/Field.php
```diff
@@ -76,30 +78,6 @@ public function __construct(
 		$this->siblings = $siblings ?? new Fields([$this]);
 	}
 
-	/**
-	 * Returns field data
-	 */
-	public function data(bool $default = false): mixed
```
distantnative, line 82: `Question:` Should we first deprecate this (and inside the method call `::toStoredValue()`)?

src/Form/Field/BlocksField.php
```diff
@@ -238,19 +239,6 @@ public function routes(): array
 		];
 	}
 
-	public function store(mixed $value): mixed
```
distantnative, line 241: `Question:` Deprecate before removal?

src/Form/Field/LayoutField.php
```diff
@@ -267,8 +269,9 @@ public function settings(): Fieldset|null
 		return $this->settings;
 	}
 
-	public function store(mixed $value): mixed
```
distantnative, line 270: `Question:` Keep deprecated `::store()`?

src/Form/Fields.php
```diff
@@ -106,6 +107,52 @@ public function fill(array $input): static
...
+						return null;
+					}
+
+					$fields = $form->fields();
+				}
+
+				continue;
+			}
+
+			return null;
```
distantnative: `Suggestion:` for readability

```suggestion
			$field = $fields->get($name);

			if ($field === null) {
				return null;
			}

			if ($index < $count) {
				$form = $field->form();

				if ($form instanceof Form === false) {
					return null;
				}

				$fields = $form->fields();
			}
```

I think this should do the same, but maybe I also didn't understand the code fully?

src/Form/Form.php
```diff
@@ -119,17 +122,17 @@ public function data($defaults = false, bool $includeNulls = true): array
...
 			} else {
-				$data[$field->name()] = $field->data($defaults);
+				$data[$field->name()] = $field->toStoredValue($defaults);
 			}
 		}
 
+		if ($includeNulls === false) {
+			$data = array_filter($data, fn ($value) => $value !== null);
+		}
+
```
distantnative: `Question:` Not sure about this, but I think before with `$field->save() === false` and `$includeNulls === true` the field would end up as `null` in the array. But with the new code. I think it's not anymore.

src/Form/Form.php
```diff
@@ -301,7 +276,7 @@ protected static function prepareFieldsForLanguage(
 	/**
 	 * Converts the data of fields to strings
 	 *
-	 * @param false $defaults
+	 * @deprecated 5.0.0 Use `::toStoredValues` instead
 	 */
 	public function strings($defaults = false): array
```
distantnative, line 279: `Question:` If we are deprecating this with this PR, should this PR also replace our own using of `::strings()` throughout the core?

src/Form/Mixin/Value.php
```diff
@@ -0,0 +1,108 @@
...
+trait Value
+{
+	/**
+	 * @deprecated 3.5.0 Use `::toStoredValue()` instead
+	 * @todo remove when the general field class setup has been refactored
+	 */
+	public function data(bool $default = false): mixed
+	{
+		return $this->toStoredValue($default);
+	}
```
distantnative, line 20: `Question:` If we had deprecated this since 3.5 - maybe time to drop it now?

src/Form/Mixin/Value.php
```diff
@@ -0,0 +1,108 @@
...
+	{
+		return $this->isEmptyValue($this->toFormValue());
+	}
+
+	/**
+	 * Checks if the given value is considered empty
+	 */
+	public function isEmptyValue(mixed $value = null): bool
+	{
+		return in_array($value, [null, '', []], true);
```
distantnative, line 47: `Question:` Wondering if this actually would need to mirror `Content\Field::isEmpty()`?

src/Form/Field.php
bastianallgeier: This is already happening. The method is now located in the Value mixin and marked as deprecated there.

src/Form/Field/BlocksField.php
bastianallgeier: If we want to keep this around and deprecate it, we would need to add the string conversion code back in there and use it in the `::toStoredValue` method or add the same code twice.

src/Form/Field/LayoutField.php
bastianallgeier: See above

src/Form/Form.php
bastianallgeier: I think you are right. This is too soon to deprecate it until toStoredValues is a full replacement. I've just reverted that.

src/Form/Mixin/Value.php
bastianallgeier: Hm, that's a very good question. I'm not sure if we break something unexpected when we start trimming strings before checking here.

src/Form/Mixin/Value.php
bastianallgeier: The deprecation warning was wrong. We added that in the FieldClass but not in Field back then. I changed it to 5.0.0

src/Form/Fields.php
bastianallgeier: It does the same. I just implemented it that way and added a couple inline comments.

src/Form/Form.php
bastianallgeier: I've reverted that part. I don't think it's necessary to make the PR more complex here than it already is.

src/Form/Mixin/Value.php
distantnative: Maybe not trimming, but check for `'[]'` as well?

distantnative: But we can also leave it as is for now, so this is not the only thing left blocking merging.

bastianallgeier: I thought about it some more and I actually wonder if this is a good idea at all. Even in the Content/Field class. I mean it's super unlikely, but what if anyone would ever want to type in "[]" in a text field? Why do we even check for that? But I'm getting side tracked. I think we can definitely care about this later.

distantnative: We just recently added this because of https://github.com/getkirby/kirby/issues/6637 - that's how the Symphony YAML parser stores empty arrays.

bastianallgeier: Ah, ok. I wasn't aware. But my feeling would be that the form classes shouldn't handle that at all. I would think that this should be handled by the storage classes, not even the Content Field class.

distantnative: Not sure I would go there, transforming each field now in storage first would be a big change. But I'll mark this as resolved for this PR then.

distantnative (approved):

PR #6781, opened 2024-11-05
`Version::isIdentical()`

distantnative:
### Summary of changes
- Added `Version::isIdentical()`
- Removed `Version::diff()`
- Using `Version::isIdentical()` to check if a changes version is still needed or identical with latest (in which case changes gets removed)


bastianallgeier (approved):
I really like the changes.

PR #6782, opened 2024-11-05
`Form\Value`: Move data methods to `Data` classes

distantnative:
- `Data::decode()` has new `$fail` parameter. When set to false, method will return empty array instead of throwing exception.
- `Json::encode()` has new `$pretty` parameter
- Removed all `Form\Mixin\Value` methods that just encode/decode data


src/Data/Data.php
```diff
@@ -80,9 +81,20 @@ public static function handler(string $type): Handler
 	/**
 	 * Decodes data with the specified handler
 	 */
-	public static function decode($string, string $type): array
-	{
-		return static::handler($type)->decode($string);
+	public static function decode(
+		$string,
+		string $type,
+		bool $exceptions = true
```
distantnative: `Questions`: 
- What do you think of the parameter name?
- Is this ok for just the `Data` class or would we need to implement it for each handler `::decode()` too?

bastianallgeier: - Another option could be `$fail`. I think we have that somewhere else, but I don't remember where. But `$exceptions` works for me as well.  
- We could indeed think of the Data class as a wrapper, which would be nice because the handlers don't have to deal with it. Would also probably be a lot more past-proof for other plugin handlers.

distantnative: Renamed it to `$fail`

bastianallgeier (approved):
I like it :)

PR #6785, opened 2024-11-09
`Data::read(fail: false)`

distantnative:
- Use `$fail` parameter also for `Data::read()`


bastianallgeier (approved):
Nice! This feels really good.

PR #6788, opened 2024-11-12
feat: AST-based query runner

rasteiner:
## Description
This PR Introduces a new query classes based on parsing queries as Abstract Syntax Tree. Based on initial tests, this new runner seems to be about 38% faster. 

### Summary of changes
- Many new classes in `Kirby\Query`:
  - `Kirby\Query\Parser`: take care of splitting a string query first into tokens of specific types and then parsing these into an AST nodes tree
  - `Kirby\Query\AST`: all kind of nodes a query can consist of and the logic how they can be resolved
  - `Kirby\Query\Visitor`: classes that help resolving (visit) an AST and turn nodes either directly into processed results (`DefaultVisitor`), but plugins could build also different visitor e.g. producing PHP code representations (transpired)
  - `Kirby\Query\Runner`: classes that take a query and with the help of the above either directly process the result (`DefaultRunner`, with in-memory cache)

- The general process is:
  1. Query string is split into a flat sequence of tokens (`Kirby\Query\Parser\Tokenizer`).
  2. Tokens are parsed into a recursive abstract syntax tree (AST) (`Kirby\Query\Parser\Parser`).
  3. AST is then [visited](https://en.wikipedia.org/wiki/Visitor_pattern) by an `Interpreter` that directly evaluates the query, or a `Transpiler` that transpiles it into PHP code.
  4. The whole process and the caching is handled by the "runner" class: `DefaultRunner`
- The original `Kirby\Query\Query` class remains in place and chooses which runner to use based on the `query.runner` option.
- All previous Query classes have been deprecated and will be removed in v7. Until then setting `query.runner` to `legacy` can still make use of them.

### Reasoning
Separating the parsing and execution steps allows for more flexibility and better performance, since it allows us to cache the AST, either in memory during a request or (via a plugin) as transpired and stored PHP code.

By adding the new runners as optional and leaving the old parser as default legacy in place, this minimizes the risk of issues and hopefully allows us to add this still to v5, only changing the default in v6.

### Additional context
The parser is a predictive recursive descent parser, making the parsing step `O(n)`. As such it can't support ambigous code, which leads to the following breaking change.

#### Query language
Kirby 5.1 adds a number of improvements to the Kirby Query language. These are based on a new AST-based parser and runner that has been added in addition to the previous (legacy) parser. The legacy parser which will remain the default parser for now but we aim to replace it with the new runner in Kirby 6. Consider the new runner in beta state during Kirby 5.x releases.

The new AST-based runner (thx to @rasteiner for starting off the work on this) runs not only more reliably and performant, but also allow us to add more functionalities. Those are not supported by the legacy runner.

- Support for more operators:
  - Logic (`AND`/`&&`/`OR`/`||`)
  - Compare (`==`, `!=`, `>`, `<, `<=`, `>=`)
  - Math (`+`, `-`, `*`, `/`, `%`)
- Closures in queries can now accept arguments
- Queries support subscript access notation, e.g. `page[site.pageMethodName]`
- `query.runner` option to switch to `Kirby\Query\Runners\DefaultRunner::class` to activate the new AST-based runner

### Breaking changes (only when manually switching to new runner)
The new query runner is stricter in parsing queries. Ambiguous terms aren't supported anymore. In some edge cases, this will require you to access an array/object via subscript notation - in rare cases on the top level via the new `this` keyword, e.g. `this["indentifier.with dots and spaces"]`. 

#### `query.runner` config option

Kirby offers a new runner for its query syntax based on parsing the query as abstract syntax tree. This runner is more reliable and efficient than the rudimentary legacy runner (which in Kirby 5 remains the default).  The new runner is currently in its beta phase. You can activate it by setting the `query.runner` config option:

  ```php
  <?php 

use Kirby\Query\Runners\DefaultRunner;
  
  return [
    'query.runner' => DefaultRunner::class
  ];
  ```
  
  #### Query syntax
  
  When using the new query runner, you can also use logical, comparison and math operators in your queries as well as receive arguments in closures:
  ```php
  // new operators
  $query = new Query('age > 25 && name != "John"');
  $query = new Query('price + vat');
  
  // arguments in query closures
  $query = new Query('(foo) => foo.homer');
  $data  = [];
  
  $bar = $query->resolve($data);
  $bar = $bar(['homer' => 'simpson']);
  $this->assertSame('simpson', $bar);
  ```

Queries will still be tried to resolve first directly from the data context, e.g. when your query is `null` and your data `['null' => 'foo']` the result will directly be `'foo'`. Same for `user.username` and `['user.username' => 'foo']`.

## Ready?
 

### For review team

### DIFF
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -9,34 +9,28 @@ use Kirby\Cms\File;
 use Kirby\Cms\Page;
 use Kirby\Cms\Site;
 use Kirby\Cms\User;
+use Kirby\Exception\InvalidArgumentException;
 use Kirby\Image\QrCode;
+use Kirby\Query\Runners\Runner;
 use Kirby\Toolkit\I18n;
 
 /**
- * The Query class can be used to query arrays and objects,
- * including their methods with a very simple string-based syntax.
- *
- * Namespace structure - what handles what:
- * - Query			Main interface, direct entries
- * - Expression		Simple comparisons (`a ? b :c`)
- * - Segments		Chain of method calls (`site.find('notes').url`)
- * - Segment		Single method call (`find('notes')`)
- * - Arguments		Method call parameters (`'template', '!=', 'note'`)
- * - Argument		Single parameter, resolving into actual types
+ * The Query class can be used to run expressions on arrays and objects,
+ * including their methods with a very simple string-based syntax
  *
  * @package   Kirby Query
  * @author    Bastian Allgeier <bastian@getkirby.com>,
- * 			  Nico Hoffmann <nico@getkirby.com>
+ *            Nico Hoffmann <nico@getkirby.com>
  * @link      https://getkirby.com
  * @copyright Bastian Allgeier
  * @license   https://opensource.org/licenses/MIT
  */
 class Query
 {
-	/**
-	 * Default data entries
-	 */
+	public static array $cache = [];
 	public static array $entries = [];
+
+	public Runner|string $runner;
 
 	/**
 	 * Creates a new Query object
@@ -46,6 +40,17 @@ class Query
 	) {
 		if ($query !== null) {
 			$this->query = trim($query);
+		}
+
+		$this->runner = App::instance()->option('query.runner', 'legacy');
+
+		if ($this->runner !== 'legacy') {
+
+			if (is_subclass_of($this->runner, Runner::class) === false) {
+				throw new InvalidArgumentException("Query runner $this->runner must extend " . Runner::class);
+			}
+
+			$this->runner = $this->runner::for($this);
 		}
 	}
 
@@ -71,6 +76,7 @@ class Query
 	 * can be found, otherwise returns null
 	 *
 	 * @throws \Kirby\Exception\BadMethodCallException If an invalid method is accessed by the query
+	 * @throws \Kirby\Exception\InvalidArgumentException If an invalid query runner is set in the config option
 	 */
 	public function resolve(array|object $data = []): mixed
 	{
@@ -78,6 +84,24 @@ class Query
 			return $data;
 		}
 
+		// TODO: switch to 'interpreted' as default in v6
+		// TODO: remove in v7
+		// @codeCoverageIgnoreStart
+
+		if ($this->runner === 'legacy') {
+			return $this->resolveLegacy($data);
+		}
+		// @codeCoverageIgnoreEnd
+
+		return $this->runner->run($this->query, (array)$data);
+	}
+
+	/**
+	 * @deprecated 5.1.0
+	 * @codeCoverageIgnore
+	 */
+	private function resolveLegacy(array|object $data = []): mixed
+	{
 		// merge data with default entries
 		if (is_array($data) === true) {
 			$data = [...static::$entries, ...$data];
--- a/src/Query/AST/ArgumentListNode.php
+++ b/src/Query/AST/ArgumentListNode.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Kirby\Query\AST;
+
+use Kirby\Query\Visitors\Visitor;
+
+/**
+ * Represents a list of (method) arguments
+ *
+ * @package   Kirby Query
+ * @author    Roman Steiner <roman@toastlab.ch>,
+ *            Nico Hoffmann <nico@getkirby.com>
+ * @link      https://getkirby.com
+ * @license   https://opensource.org/licenses/MIT
+ * @since     5.1.0
+ */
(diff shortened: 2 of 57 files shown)


rasteiner:
Updates
----- 

 - the syntax `user\.username` has been implemented for now. The corresponding test has been updated to use that instead of the ambiguous `user.username`. 
   Similarly, the following syntax is supported too: `user\\\.username`, for when the actual array key is really `user\.username`. (The first backslash escapes the second, the third escapes the dot. 
 - intercept has been implemented. 

Discussion
-----
In regards to intercept, all evaluated objects and variables are now passed through intercept. Even stuff like literal strings or numbers. It could be worth discussing if this is actually useful or just slow. One idea could be to only intercept objects on which a member field or method is accessed. Like:

 - `page('foo')` would stay `page('foo')`
 - `page('foo').title` would become `intercept(page('foo')).title`.

distantnative:
> One idea could be to only intercept objects on which a member field or method is accessed.
This would be an option if it also includes objects down the chain: E.g. `page.files.first` would also need the option to intercept the `Files` and the `File` object.

rasteiner:
> > One idea could be to only intercept objects on which a member field or method is accessed.

> This would be an option if it also includes objects down the chain: E.g. `page.files.first` would also need the option to intercept the `Files` and the `File` object.

Sure, `page.files.first` would be: 
```js
intercept(
  intercept(page).files
).first
```
`page.files.first.delete` would be: 
```js
intercept(
  intercept(
    intercept(page).files
  ).first
).delete
```

also `foo.callMeBack(() => page('bar')).delete` would be:

```js
intercept(
  intercept(foo).callMeBack(() => page('bar'))
).delete
```

rasteiner:
Having implemented the "`user\.username`" syntax, I've now realized that, while passing the unit test, I didn't really solve the underlying problem.
By allowing us to escape the dot, there's now an exception explicitly for the dot, but every other special symbol (even like whitespace) would still break the parsing:

```php
$query = new Query('Shipping Address'); 
$this->assertSame('742 Evergreen Terrace', $query->resolve(['Shipping Address' => '742 Evergreen Terrace']));
```

I see 4 options:
1. We ignore this: the only special character allowed in array keys remains the dot 
2. We escape literally everything, kinda like in regex
3. We introduce an index operator (i.e. `foo["Shipping Address"]` or `{"Shipping Address"}`)
4. We introduce a new kind of string literal which is interpreted as identifier, like in MySQL queries the back ticks for complex column names `` `Shipping Address` ``

The index operator obviously gives the most bang for the buck as it allows any kind of expression as index (not just strings), but it either requires an unconventional syntax (the `{}` brackets in the example above) or, for the special case of wanting to access a "root level variable", we need to introduce a reserved keyword for the global context (`foo` in example)

rasteiner:
I've implememted the intercept logic like described above (see test [here](https://github.com/rasteiner/kirby/blob/002664224ed2a6d6fcbbe3b9da051a925b1c4e56/tests/Toolkit/Query/RunnerTest.php#L172-L234))

Other notable change would be moving the "resolver cache" (the cache that maps query strings to closures) to the `Kirby\Query\Query` class. 
Before this change the cache was hidden away in the Runner classes, but we want that cache to be both static (so that one can simply create a new `Query` instance and benefit from the cache) as well as under the control of `Query` (should the "Query::entries" change during a request).

distantnative:
@rasteiner we are currently trying to wrap some things up for the v5 beta, that's why I'm not active here much these days, but will get back to it probably next week! :)

src/Query/Query.php
```diff
@@ -78,6 +76,27 @@ public function resolve(array|object $data = []): mixed
...
+		$runner = $this->runner();
+
+		return $runner->run($this->query, (array)$data);
+	}
+
+	/**
+	 * @deprecated 6.0.0
+	 * @codeCoverageIgnore
+	 */
+	private function resolve_legacy(array|object $data = []): mixed
```
bastianallgeier:
```suggestion
	private function resolveLegacy(array|object $data = []): mixed
```

rasteiner:
😯

src/Query/Query.php
```diff
@@ -47,6 +41,17 @@ public function __construct(
 		if ($query !== null) {
 			$this->query = trim($query);
 		}
+
+		$this->runner = App::instance()->option('query.runner', 'legacy');
```
distantnative, line 45: `Discuss:` Should this be a config option or a core component?

bastianallgeier: It feels more like a potential core component to me.

distantnative (approved):

PR #6790, opened 2024-11-12
Update EULA to 2024-11-12

lukasbestle:
PR to `main` as this is independent of a release. We could also merge it to the 4.5 release branch if needed.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,7 +1,7 @@
 # Kirby License Agreement
 
-Published: November 28, 2023  
-Source: https://getkirby.com/license/2023-11-28
+Published: November 12, 2024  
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):
I'm fine with merging it now.

lukasbestle:
Ah, doesn't work because this doesn't come from a release branch, so the release CI workflows are not executed. I think we need to merge this into the 4.5 release branch first. Speaking of which, is there one? Can't find one.

distantnative:
@lukasbestle I don't think it exists but could be created from the 4.5.0-rc.1 tag

lukasbestle:
Now we just need to remember to use the release branch for the final 4.5.0 release :)

PR #6791, opened 2024-11-13
Add the latest version of a translation automatically

bastianallgeier:
## Description

If the latest version of a translation does not exist yet when creating changes, the latest version of the default language is automatically copied over. 

## Ready?

### For review team

### DIFF
--- a/tests/Content/VersionTest.php
+++ b/tests/Content/VersionTest.php
@@ -191,6 +191,44 @@ class VersionTest extends TestCase
 	/**
 	 * @covers ::create
 	 */
+	public function testCreateMultiLanguageWhenLatestTranslationIsMissing(): void
+	{
+		$this->setUpMultiLanguage();
+
+		$latest = new Version(
+			model: $this->model,
+			id: VersionId::latest()
+		);
+
+		$changes = new Version(
+			model: $this->model,
+			id: VersionId::changes()
+		);
+
+		$this->assertContentFileDoesNotExist('en', $latest->id());
+		$this->assertContentFileDoesNotExist('en', $changes->id());
+		$this->assertContentFileDoesNotExist('de', $latest->id());
+		$this->assertContentFileDoesNotExist('de', $changes->id());
+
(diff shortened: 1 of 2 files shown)


tests/Content/VersionTest.php
```diff
@@ -188,6 +188,44 @@ public function testCreateMultiLanguage(): void
...
+
+		// create a changes version in the other language
+		$changes->save([
+			'title' => 'Translated Test',
+		], $this->app->language('de'));
+
+		$this->assertContentFileExists('en', $latest->id());
+		$this->assertContentFileDoesNotExist('en', $changes->id());
+		$this->assertContentFileExists('de', $latest->id());
+		$this->assertContentFileExists('de', $changes->id());
```
distantnative, line 226: `Comment:` I don't want to block this PR because of it, but we should really try to get away from unit tests checking for the content file going forward as those are exactly the ones really biting us when we try to implement `ImmutableMemoryStorage` etc.

distantnative (approved):

tests/Content/VersionTest.php
bastianallgeier: That's the opposite of what @lukasbestle suggested. His thoughts were that we try to keep the dependencies in tests as low-level as possible. Checking for changes in the file system directly will make sure that we don't use version code or storage code in tests to prepare or check version and storage code.

distantnative: I can just tell you that so many unit tests issues were because of this when I tried to make the LabPage branch work. Because in many cases the storage type could change and then pinning unit tests to checking for a content file that is only used by one storage type blows up the setup.

bastianallgeier: Yeah, I get it. I feel a bit lost here to be honest. I don't know a good way forward. Maybe some form of abstraction?

distantnative: I don't think we need to solve this here. Only midterm. But I would've thought that storage is exactly that abstraction.

PR #6792, opened 2024-11-13
Move lock wildcard into the Lock::for method

bastianallgeier:
## Description

`Lock::for` now accepts the Language wildcard to search for a lock in any translation. 

## Ready?

- [x] In-code documentation (wherever needed)
- [x] Unit tests for fixed bug/feature
- [x] Tests and CI checks all pass

### For review team

- [ ] Add lab and/or sandbox examples (wherever helpful)
- [ ] Add changes & docs to release notes draft in Notion

PR #6793, opened 2024-11-13
Languages dropdown with changes/lock indicator

distantnative:
## Description

- [x] https://github.com/getkirby/kirby/pull/6792
- [x] @bastianallgeier Do you want to include the mockup Lab example or rather not?

### Summary of changes
- Languages dropdown indicates default language
- Languages dropdown indicates when has changes/locked
- `k-dropdown-content` exposes `items` as scoped default slot
- `k-dropdown-content` has new `item` scoped slot


afbora:
I've found a lot of issues:
- Wrong dot indicator
- Wrong text indicator
- Cannot discard some changes

I can reproduce with starterkit notes pages but not with default pages. Just edit some blocks and switch the languages. You'll see the all issues.

distantnative:
@afbora Could you please be more specific which steps lead to which problem? And also, is that an issue of this PR or on `v5/changes/develop` already?

afbora:
@distantnative Yes wrong dot indicator and discarding changes issues still valid for `v5/changes/develop` branch.

### Wrong dot & text  indicator
 - Edit default language blocks without save
 - Switch to the secondary language and edit some blocks without save
 - Try to discard for secondary language
 - You'll see text and dot indicator is wrong
 
 Another steps:
 - Switch to secondary language
 - Edit some blocks and discard it
 - Still showing indicator for secondary language

### Discarding changes
- Edit default language blocks without save
- Switch to secondary language than go back to default language again without any edit
- You'll see default language changes are gone.

### Cannot discard some changes
It happened once or twice but I couldn't find the exact steps right now. If it happens again, I'll write it.

bastianallgeier:
So this seems to be a particular issue of the blocks field? Have you experienced the same with other fields?

bastianallgeier:
@distantnative I wouldn't add the old lab example to be honest.

afbora:
@bastianallgeier I think you're right. Seems some locking process broken with blocks field.

afbora:
I've done few more tests. I guess it's related with `writer` field not blocks field. Blocks field also have writer field.

bastianallgeier:
@afbora that makes sense. We had so many issues with the writer before. If I remember correctly, it's the way that the input event is triggered in the writer field. It doesn't only happen when you type something, it also happens when the value is updated via JS from the backend. That basically creates a wrong additional event.

PR #6794, opened 2024-11-14
Version::isIdentical now uses form values to check for identical versions

bastianallgeier:
## Description

Using the Form class to create form values is unfortunately the only reliable method at the moment to compare two versions with each other. 

## Ready?

### For review team

### DIFF
--- a/panel/src/panel/content.js
+++ b/panel/src/panel/content.js
@@ -38,11 +38,6 @@ export default (panel) => {
 		 * Removes all unpublished changes
 		 */
 		async discard() {
-			// avoid requests if there is no API defined
-			if (!this.api) {
-				return;
-			}
-
 			if (this.isProcessing === true) {
 				return;
 			}
@@ -91,11 +86,6 @@ export default (panel) => {
 		 * Publishes any changes
 		 */
 		async publish() {
-			// avoid requests if there is no API defined
-			if (!this.api) {
-				return;
-			}
-
 			if (this.isProcessing === true) {
 				return;
 			}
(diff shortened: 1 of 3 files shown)


panel/src/panel/content.js
```diff
@@ -38,6 +38,11 @@ export default (panel) => {
 		 * Removes all unpublished changes
 		 */
 		async discard() {
+			// avoid requests if there is no API defined
+			if (!this.api) {
+				return;
```
distantnative: `Questions:` What were the cases where this was happening?

`Suggestion:` Shouldn't this throw a console error at least as something is misconfigured then?

bastianallgeier: The last input happened and the save call is scheduled but the user already navigated quickly enough to a new view without an API endpoint. I ran into that. 

Thinking about it while typing, I realize that there's even a worse option. If you switch fast enough to a different view with a different API endpoint, the call would go to that one instead. This means we really need to cancel the last call when moving away.

distantnative: But if we cancel it, the content gets lost. We rather need to ensure that the request succeeds to the right endpoint.

distantnative: I think the issue lies somewhere here: https://github.com/getkirby/kirby/blob/v5/changes/develop/panel/src/panel/content.js#L179-L189 

We watch the endpoint for a change to send a last save request if any new input came in after the last request. But since we call it when `content.api` changes, the call made then won't have the old API endpoint anymore. I think we would need to pass the old API endpoint string there somehow to the save method.

distantnative: Also seeing that the watcher fires even on the initial load which isn't great.

PR #6799, opened 2024-11-18
Improvements for the content.js module

bastianallgeier:
### Summary of changes

- New option to pass the silent argument to api requests in the options object. This will simplify the API calls quite a bit. Before: `this.$panel.api.post('/path', data, options, 'POST', silent = true)` after: `this.$panel.api.post('/path', data, { silent: true})` This is now used in `this.$panel.content.save()` to keep safe calls silent.
- New `changes` property for the `k-model-tabs` component to pass an object of changes directly. This will make the model tabs more independent from the content module and we can reuse the computed changes object that we already define in the model views.
- All content module methods with a dependency on the api endpoint now have a second optional api endpoint argument. This argument will use the view's api prop by default. 
  - `this.$panel.content.discard(api)`
  - `this.$panel.content.isLocked(api)`
  - `this.$panel.content.lock(api)`
  - `this.$panel.content.publish(values, api)`
  - `this.$panel.content.save(values, api)`
  - `this.$panel.content.saveLazy(values, api)`
  - `this.$panel.content.update(values, api)`  
- All content module methods will throw an Error if they cannot be used on a different view for a different API endpoint
- New content module methods: 
  - `this.$panel.content.isLocked()`
  - `this.$panel.content.lock()`
  - `this.$panel.content.changes()`
  - `this.$panel.content.isCurrent()`
  - `this.$panel.content.saveLazy()` 
- Removed content module props/getters
  - `this.$panel.content.isSaved`
  - `this.$panel.content.originals`
- `this.$panel.content.saveLazy` is now used in the model view `onInput` event instead of throttling it directly in the view. This means that devs can now choose between calling `this.$panel.content.save` or `this.$panel.content.saveLazy` if the action should be delayed.
- The ModelView `onSave` method has been renamed to `onViewSave` This matches the custom event in the events.js, which will trigger this and causes less confusion. This event is only fired on `keydown.cmd.s` and will in fact submit the form and not simply save the content. That's why I believe we should rename this right away.
- The ModelView `isUnsaved` computed property has been renamed to `hasChanges`. This is also supposed to clean up possible confusions, because it's computing the number of items in the `this.$panel.content.changes()` object. 
- All actions in the content module now fire custom events after the requests went through: 
  - `content.save`
  - `content.discard`
  - `content.publish`
- The new `content.save` event is used to update the new ModelView `isSaved` data property. By moving it to the ModelView and using the custom event to keep it updated, we can make sure that it is always correct and bound to the right api endpoint. Otherwise, we would need to keep track of multiple `isSaved` states in the content module. 
- Redundant `on` methods have been removed from the PreviewView.
- The outdated `protectedFields` property has been removed from ModelView

## Outlook

Methods that depend on data from the current view could later be refactored to make API calls instead whenever they are used with different endpoints. 

I.e. right now `this.$panel.content.lock()` wiil only work for the current view and throw an error if you try to pass a different endpoint. But it's totally possible to simply check for the endpoint and then make an async call to that view instead and get the lock info from there. The same would work for `this.$panel.content.changes()` for example. But I didn't want to blow this up even more. I just think that this is a nice way forward that could be really useful. 

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Views/Files/FileView.vue
+++ b/panel/src/components/Views/Files/FileView.vue
@@ -22,7 +22,7 @@
 				<k-form-controls
 					:editor="editor"
 					:is-locked="isLocked"
-					:is-unsaved="isUnsaved"
+					:is-unsaved="hasChanges"
 					:modified="modified"
 					@discard="onDiscard"
 					@submit="onSubmit"
--- a/panel/src/components/Views/Files/FileView.vue
+++ b/panel/src/components/Views/Files/FileView.vue
@@ -21,8 +21,8 @@
 				<k-view-buttons :buttons="buttons" @action="onAction" />
 				<k-form-controls
 					:editor="editor"
+					:has-changes="hasChanges"
 					:is-locked="isLocked"
-					:is-unsaved="hasChanges"
 					:modified="modified"
 					@discard="onDiscard"
 					@submit="onSubmit"
--- a/panel/src/components/Views/Files/FileView.vue
+++ b/panel/src/components/Views/Files/FileView.vue
@@ -38,7 +38,7 @@
 			@submit="onSubmit"
 		/>
 
-		<k-model-tabs :tab="tab.name" :tabs="tabs" />
+		<k-model-tabs :changes="changes" :tab="tab.name" :tabs="tabs" />
 
 		<k-sections
 			:blueprint="blueprint"
--- a/panel/src/components/Views/ModelView.vue
+++ b/panel/src/components/Views/ModelView.vue
@@ -80,7 +80,7 @@ export default {
 	mounted() {
 		// create a delayed version of save
 		// that we can use in the input event
-		this.autosave = throttle(this.save, 1000, {
+		this.save = throttle(this.save, 1000, {
 			leading: true,
 			trailing: true
 		});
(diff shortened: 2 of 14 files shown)


panel/src/components/Views/Files/FileView.vue
```diff
@@ -22,7 +22,7 @@
 				<k-form-controls
 					:editor="editor"
 					:is-locked="isLocked"
-					:is-unsaved="isUnsaved"
+					:is-unsaved="hasChanges"
```
distantnative: `Question:` Should we rename the `k-form-controls` prop as well?

panel/src/components/Views/ModelView.vue
```diff
@@ -37,92 +37,93 @@ export default {
...
 		},
 		onInput(values) {
-			if (this.isLocked === true) {
-				return false;
-			}
+			// update the content for the current view
+			// this will also refresh the content prop
+			this.$panel.content.update(values, this.api);
+			// trigger a throttled save call with the updated content
+			this.$panel.content.saveLazy(this.content, this.api);
```
distantnative: `Discussion:` This seems a bit unintuitive: Why do we need to call `saveLazy` separately? Shouldn't any call of `update` also trigger a lazy save? Which situation to we have where only an update but no save should happen? I would think that this is actually a situation that we really want to avoid. Then the content module should take care of it itself and not put that responsibility out in the calling code.

panel/src/panel/content.js
```diff
@@ -1,25 +1,23 @@
...
-		/**
-		 * Returns all fields and their values that
-		 * have been changed but not yet saved
-		 *
+		 * Returns an object with all changed fields
+		 * @param {String} api
 		 * @returns {Object}
 		 */
-		get changes() {
+		changes(api = panel.view.props.api) {
```
distantnative: `Question:` Not sure about this but I somehow remember that the default will not be applied when the method is called but when the module code is parsed the first time. Do you know if I am wrong or right with this?

panel/src/panel/content.js
```diff
@@ -1,25 +1,23 @@
...
+			// changes can only be computed for the current view
+			if (this.isCurrent(api) === false) {
+				throw new Error("Cannot get changes from another view");
```
```suggestion
				throw new Error("Cannot get changes for another view");
```

panel/src/panel/content.js
```diff
@@ -156,52 +174,29 @@ export default (panel) => {
...
 			}
 
+			if (this.isCurrent(api) === false) {
+				throw new Error("The content in another view cannot be updated");
+			}
+
 			panel.view.props.content = {
 				...panel.view.props.originals,
 				...values
 			};
```
distantnative, line 199: `Suggestion:` Continuing my thought from above, wouldn't it make sense to call here `saveLazy(api, panel.view.props.content)`?

panel/src/components/Views/Files/FileView.vue
bastianallgeier: Yes, I would love to do that. Just didn't dare to put even more changes in there :)

panel/src/components/Views/ModelView.vue
bastianallgeier: You might be right that a separate call is not needed at all. We had the seperate update call for a while. That's mainly why I kept it that way. But I also don't think that there are a lot of good reasons for it.

panel/src/panel/content.js
bastianallgeier: I'm not sure I understand what you mean.

panel/src/components/Views/Files/FileView.vue
distantnative: 😆 I think you can. This won't make much of a difference.

panel/src/panel/content.js
distantnative: Whether the default of `api` will be `panel.view.props.api` from the moment when the method gets called or whatever `panel.view.props.api` contains at the moment the content module gets initialized for the first time.

bastianallgeier: Oh, now I get it. That's interesting. That would mean that we should rather leave it undefined and then set the default in each method call, right? Right now, it's not super important, because the view will set the api endpoint explicitely on all calls, but in other places it might lead to unwanted problems indeed.

panel/src/components/Views/Files/FileView.vue
bastianallgeier: Thank you for the permission. I needed that. 😂

bastianallgeier: ☑️

panel/src/panel/content.js
bastianallgeier: I've moved the default api endpoint assignment into the method calls.

panel/src/components/Views/ModelView.vue
bastianallgeier: I remember again why we need this. We want the content prop to be updated immediately, while saving will happen only a second later. When we put the update part into the save method, it will also be delayed and create weird side-effects.

distantnative: But we wouldn't put the update part into the save method, but rather make the update method do the update logic and then call itself the `saveLazy` method, ensuring that this update eventually will get saved.

bastianallgeier: Ah, right. But would we then also need a update and updateLazy method or would you handle that with a method argument?

distantnative: Why would you want to offer the option? I think every update should trigger saveLazy.

bastianallgeier: All of our api calls are asynchronous and I would call the default expectation for the api calls in the content method as well. We only need the throttled one for our text inputs. But let's say you want to modify a toggle, that toggle action could totally be sent directly instead of waiting. That's why I thought it would be great to have both. The async one and a lazy one

distantnative: I don't know, to me it would feel easier to have just one method (`update`) that developers use and we take care of all of it internally in the module. Not having them think about "do I need update or updateLazy". And I think always lazy would be just fine. Yes modifying a toggle wouldn't need it, but it also doesn't hurt, the delay is a mere second. And we don't mind that delay in our normal form with toggles either, there the delay is also applied no matter if it's text or a toggle.

bastianallgeier: Ok, convinced. I implemented it like that.

distantnative (approved):

PR #6401, opened 2024-04-16
Syntax preparations for PHP 8.4

lukasbestle:
### Refactoring

- Use more modern PHP syntax and PHPUnit assertions where applicable

### Breaking changes

- `Str::camel()`, `Str::camelToKebab()`, `Str::float()`, `Str::kebab()`, `Str::kebabToCamel()`, `Str::length()`, `Str::lower()`, `Str::safeTemplate()`, `Str::short()`, `Str::slug()`, `Str::snake()`, `Str::studly()`, `Str::substr()`, `Str::template()`, `Str::ucfirst()`, `Str::ucwords()`, `Str::unhtml()`, `Str::upper()` and `Str::widont()` can no longer be called without a value argument (passing a `null` value still works)

## For review team

- [x] Add changes & docs to release notes draft in Notion

PR #6811, opened 2024-11-23
[v4] PHP 8.4 support

distantnative:
- Skips Psalm analysis for PHP 8.4 as Psalm doesn't support it yet.

### Enhancements
- Support for PHP 8.4

### Refactoring
- Explicitly type-hint nullable parameters with `|null`
- Removed `= null` default for implicitly required parameters

## Ready?

### For review team


lukasbestle (commented):
I think most of the CS fixes will create merge conflicts with v5. E.g. in v5 we have replaced the `something|null $param = null` params with `something|null $param` and made some fixes to docblocks where CS fixer now again wreaks havoc.

If we merge this as is to ensure PHP 8.4 compatibility in v4, we should probably not continue to merge back v4 changes into the v5 branch, otherwise it will be a merge conflict hell. Alternative could be to backport the CS changes manually from v5.

distantnative:
@lukasbestle We made nullable parameters non-optional?

This got created as Basti mentioned the user demand for 8.4 support in v4 as well. I think we'll need to tackle the merge conflict once. I think it'll be not too bad (as we've merged regularly, merging into `v5/develop` after merging this PR we can go with the `v5/develop` version in pretty much all conflicts). But while we probably won't have a lot more to merge from v4 to v5 I wouldn't want to rule it out in general.

lukasbestle:
Not in general, but in some `Str` methods (breaking change) and also in those cases where the parameter is not in the last position, which are those lines changed by this PR. Those parameters are never optional, which is the reason why PHP 8.4 deprecates the `type $param = null` syntax in favor or `type|null`. We had discussed that in https://github.com/getkirby/kirby/pull/6398 and implemented it with https://github.com/getkirby/kirby/pull/6401.

Looking back, it would have been better to make those changes in v4 and not v5. But now it's like it is.

distantnative:
If we merge `develop-minor` into `v5/develop` right before and right after merging this PR and then go with the `v5/develop` version for any merge conflict, we should be fine, in terms of merge conflicts.

In terms of making those changes for v4 as well of course not. But maybe that's ok that v4 doesn't get those enhancements but just most basic 8.4 support.

lukasbestle (approved):

PR #6816, opened 2024-11-26
Support for old locks

bastianallgeier:
### Summary of changes

- New `Lock::legacy()` method to return a lock instance based on an old .lock file
- New `Lock::legacyFile()` method to return an absolute path to an old lock file based on the model
- The PreviewView now shows a custom message for old lock states because it cannot display the current changes from other users. 
- New cleanup method in the Changes controller to remove old lock files as soon as new changes are stored or discarded

### DIFF
--- a/src/Api/Controller/Changes.php
+++ b/src/Api/Controller/Changes.php
@@ -3,7 +3,9 @@
 namespace Kirby\Api\Controller;
 
 use Kirby\Cms\ModelWithContent;
+use Kirby\Content\Lock;
 use Kirby\Content\VersionId;
+use Kirby\Filesystem\F;
 use Kirby\Form\Form;
 
 /**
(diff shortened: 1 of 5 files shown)


src/Api/Controller/Changes.php
```diff
@@ -77,6 +91,8 @@ public static function save(ModelWithContent $model, array $input): array
 		$changes = $model->version(VersionId::changes());
 		$latest  = $model->version(VersionId::latest());
 
+		static::cleanup($model);
```
distantnative, line 106: `Todo:` Can we add comments here to clarify as I was writing a long question only to realize at the end why this works? Something like "it's save to remove the legacy lock file at any call to this method as all invalid calls, e.g. from another users, will be blocked through the existing legacy lock. The only actions that can call this method can origin from the lock owner which then should clean up the legacy lock". Maybe better formulated, but to make this easier to understand later on.

Plus, we should add `@todo` marks to remove this in v6, v7 or so.

src/Content/Lock.php
```diff
@@ -124,6 +140,49 @@ public function isLocked(): bool
...
+			legacy: true
+		);
+	}
+
+	/**
+	 * Returns the absolute path to a legacy lock file
+	 */
+	public static function legacyFile(ModelWithContent $model): string
+	{
+		$root = $model::CLASS_ALIAS === 'file' ? dirname($model->root()) : $model->root();
```
distantnative:
```suggestion
		$root = match ($model::CLASS_ALIAS) {
			'file'  => dirname($model->root()),
			default => $model->root()
		};
```

distantnative (approved):

PR #6821, opened 2024-11-29
Preview tokens 1: Groundwork A (Testing)

lukasbestle:
### Summary of changes

Ensure that all features of `$page->render()` are tested in one place

### Reasoning

Provides a baseline to reduce the risk of regressions

### Housekeeping

- Refactor page render tests out into separate test class and extend them to cover more cases

## Ready?

### For review team


distantnative:
`Suggestion:` Could we add a comment to the PageCacheTest class that now the tests for the caching in the render method can be found in the PageRenderTest class? Might help us down the road to remember/understand where to find what.

tests/Cms/Pages/PageRenderTest.php
```diff
@@ -0,0 +1,401 @@
...
+
+namespace Kirby\Cms;
+
+use Kirby\Cache\Value;
+use Kirby\Exception\NotFoundException;
+use Kirby\Filesystem\Dir;
+use Kirby\TestCase;
+
+/**
+ * @covers \Kirby\Cms\Page::cacheId
```
distantnative: `Todo:` This one doesn't make that much sense for me. If we want to improve the separation what tests do what, this method should be covered in the PageCacheTest class (if needed with a new test), but not by PageRenderClass.

lukasbestle: `cacheId` is only ever called and used by `render()`.

distantnative: Currently yes, but we should test it then separately in `PageCacheTest`. For me, understanding our tests will be harder if we have a method like `::cacheId()` but its test coverage is not in `PageCacheTest` (the main class testing the caching), but some other class.

lukasbestle: The only thing that still remains in `PageCacheTest` are tests for the `isCacheable` method. Should we rather move those back in to `PageRenderTest` then? Because I think it would be even worse to start adding additional render tests to `PageCacheTest`.

distantnative: If you prefer that, I'd see no downside.

distantnative (approved):

bastianallgeier (approved):
The test improvements are really great.

PR #6822, opened 2024-11-29
Preview tokens 2: Groundwork B (Core)

lukasbestle:
## 🚨 Merge first

- #6821

### Summary of changes

Two smaller changes that don't fit into the following PRs

### Additional context

The preview tokens will no longer depend on the model in PR 8 because we will need to be able to generate tokens for arbitrary URIs if a custom preview URL was set in the blueprint.

No unit test for `$page->isVerified()` as this method will be removed in PR 7.

### Enhancements

- Harden preview token verification to mitigate guessing the preview token with a timing attack

### Bug fixes

- Fix PHP error when `null` is passed as model to `$kirby->contentToken()`

### Breaking changes

- `$kirby->contentToken()` only accepts `object|null` for the `$model` parameter instead of `mixed`. In the Kirby core, we currently only ever pass `File` objects or `null`.

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -431,14 +431,14 @@ class App
 	 * Generates a non-guessable token based on model
 	 * data and a configured salt
 	 *
-	 * @param mixed $model Object to pass to the salt callback if configured
+	 * @param object|null $model Object to pass to the salt callback if configured
 	 * @param string $value Model data to include in the generated token
 	 */
-	public function contentToken(mixed $model, string $value): string
+	public function contentToken(object|null $model, string $value): string
 	{
 		$default = $this->root('content');
 
-		if (is_object($model) === true && method_exists($model, 'id') === true) {
+		if ($model !== null && method_exists($model, 'id') === true) {
 			$default .= '/' . $model->id();
 		}
 
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -438,7 +438,7 @@ class App
 	{
 		$default = $this->root('content');
 
-		if (method_exists($model, 'id') === true) {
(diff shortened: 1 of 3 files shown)


distantnative:
`Suggestion:` would be great if we could make the type hinting of `App::contentToken($model)` more specific. Whether in this PR or a later one.

lukasbestle:
What would be the reasoning? We don't know if it's being used in plugins, so it would be a breaking change.

distantnative:
But wouldn't be that a good breaking change to better specify what makes sense to pass into this method? `mixed` is like "just try and see where it fails". The doc blocks say it should be an object, if I understand your description here correctly it might be a string as well in the future. But then we should type hint that IMO and not allow a float or int or array that will not lead to anything.

lukasbestle:
In the core we would only need `File|null` now. Files for the media tokens, `null` for the preview tokens. The unit tests have already tested strings before, but I don't think that there was ever a use case for that.

But I really wonder if and what plugin devs might need. So maybe we can go with `object|null`? This would at least be consistent with the param description we already had.

distantnative:
`object|null` sounds good to me. Not that anyone probably really is using it with a custom object, but it would work in theory if that custom object had the method. So no harm I think to allow it.

lukasbestle:
@distantnative Implemented as discussed

distantnative:
@lukasbestle You want to do anything about the coverage test? Otherwise looks good 👍

lukasbestle:
See PR description:

> No unit test for `$page->isVerified()` as this method will be removed in PR 7.

distantnative (approved):

bastianallgeier (approved):

PR #6824, opened 2024-12-01
Preview tokens 4: `VersionId::render()`

lukasbestle:
## 🚨 Merge first

- #6821
- #6822
- #6823

### Summary of changes

Helper method `VersionId::render()` to set `VersionId::$render` only during rendering

### Reasoning

- Atomically set the value (always ensure to restore it afterwards using `try finally`)
- Avoids having to mess with the static `$render` property from other classes (clean code); we could set the visibility of that prop asymmetrically once we drop support for PHP 8.3

### Additional context

Will be used in PR 5

### Features

- New `Kirby\Content\VersionId::render()` method to atomically set the render version during a render operation

## Docs

*None*

## Ready?

- [x] In-code documentation (wherever needed)
- [x] Unit tests for fixed bug/feature
- [x] Tests and CI checks all pass

### For review team

- [x] ~Add lab and/or sandbox examples (wherever helpful)~
- [x] Add changes & docs to release notes draft in Notion

PR #6823, opened 2024-12-01
Preview tokens 3: Preview URL refactoring

lukasbestle:
## 🚨 Merge first

- #6821
- #6822

### Summary of changes

A PR with three related refactoring changes to the existing preview URL logic.

### Reasoning

- Duplicated code in `$page->previewUrl()` and `$site->previewUrl()` would become even more in following PRs if we wouldn't DRY it
- v5 is a good opportunity to rename the `token` query param in draft preview URLs to `_token` as the generation logic of those tokens will change anyway (in PR 8)

### Additional context

`$page->previewUrl()` and `$site->previewUrl()` can stay as aliases/shortcuts as they do no harm and might be useful to devs as well.

### Housekeeping

- Consolidate duplicated logic from `$page->previewUrl()` and `$site->previewUrl()` in `$version->url()` (with aliases to the existing methods)

### Breaking changes

- Rename `token` query param for preview URLs to `_token` for consistency with the new `_version` param

### Breaking changes (since alphas)

- Rename parameter name `$page->previewUrl($version)` to `$page->previewUrl($versionId)` for consistency with other methods

## Ready?

- [x] In-code documentation (wherever needed)
- [x] Unit tests for fixed bug/feature
- [x] Tests and CI checks all pass

### For review team

- [x] ~Add lab and/or sandbox examples (wherever helpful)~
- [x] Add changes & docs to release notes draft in Notion

PR #6825, opened 2024-12-01
Preview tokens 5: Version-based rendering

lukasbestle:
## 🚨 Merge first

- #6821
- #6822
- #6823
- #6824

### Summary of changes

`$page->render()` is now responsible to choose the correct version to render.

Users can at any time pass a custom version ID that will always be used if passed. Otherwise the version ID is auto-detected.

If a render version is already set globally, this is the first detection priority. This can be relevant if a template calls `$somePage->render()` of another page. By using the global render version ID, that rendered version will match the one of the "outer" page.

Otherwise the version ID will be taken from the request. In this PR this only re-implements the draft tokens.

### Reasoning

- Usage flexibility
- DRY code

### Additional context

- Logic is moved over 1:1 for now. Following PRs 7-9 will implement the actual version tokens.
- Tests need to disable caching in this PR because the current caching logic would make those tests fail otherwise. PR 6 will fix this and delete the temporary `TODO` additions to the tests.

### Enhancements

- It is now possible to override the version ID in `$page->render()` with the new `$versionId` parameter.

### Refactoring (since alphas)

- Move request-based version detection from the `App` class closer to the action in `$page->render()`

## Docs

*We could document the new `$versionId` argument to `$page->render()`. But I don't think we already have a place in the guide or cookbook that uses `$page->render()`, at least I couldn't find one.*

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -4,7 +4,6 @@ namespace Kirby\Cms;
 
 use Closure;
 use Generator;
-use Kirby\Content\VersionId;
 use Kirby\Data\Data;
 use Kirby\Email\Email as BaseEmail;
 use Kirby\Exception\ErrorPageException;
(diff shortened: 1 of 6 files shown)


bastianallgeier (approved):
It all looks very clean to me. But a second pair of eyes from @distantnative would be good :)

distantnative (approved):

bastianallgeier:
We mention the render method in the routing guide: https://getkirby.com/docs/guide/routing I'm not really sure if it actually fits there to document the $versionId argument. 

But we should definitely add an example here: https://getkirby.com/docs/reference/objects/cms/page/render

PR #6828, opened 2024-12-02
Preview tokens 6a: Fix preview token of the site model

lukasbestle:
## 🚨 Merge first

- #6827

### Summary of changes

In PR 3 I introduced the new `$version->previewToken()` method. While writing tests for PR 7 I realized that giving the site model its own preview token was wrong. The site preview renders the home page by default, so it also needs to use the home page's preview token by default.

### Additional context

This is an "intermediary" PR that gets its name 6a so that we can continue with PR 7 with its original intended contents I already announced.

## Changelog

*None* (internal change before any release)

## Docs

*None*

## Ready?

### For review team

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -398,15 +398,24 @@ class Version
 	 */
 	public function previewToken(): string
 	{
-		$id = match (true) {
-			$this->model instanceof Site => '',
-			$this->model instanceof Page => $this->model->id() . $this->model->template(),
-			default                      => throw new LogicException('Invalid model type')
-		};
+		if ($this->model instanceof Site) {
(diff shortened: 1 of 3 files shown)


bastianallgeier (commented):
`Question`: What happens when the site or a page sets a completely different preview URL in its blueprint? Does this matter here at all?

lukasbestle:
Very good question. But also simple to answer:

Those custom blueprint-defined previews will follow in the last PR 9 and will be implemented by constructing the token from the raw URI. Model-based tokens will still be relevant for the default case as they are more robust and reliable if no URI-based token is needed.

bastianallgeier (approved):

PR #6829, opened 2024-12-02
Preview tokens 7: Main feature

lukasbestle:
## 🚨 Merge first

- #6827
- #6828

### Summary of changes

It's finally time for the real thing and main attraction for this PR series. 🎉

This actually:

- adds that the token is added to all changes preview URLs,
- validates the token when auto-detecting the changes version from the request and
- uses the same implementation in `$page->renderVersionFromRequest()` for the draft detection in `$app->resolve()`

### Reasoning

- General thought behind this feature: Prevent externals from accessing changes previews unless they received the full preview URL including token
- Avoid duplicated code by making the draft resolver hook into the same logic as well

### Additional context

This is the first PR where it really makes sense to test it by hand as well. All "default" previews should now work. Only custom blueprint `preview` options that point to other pages don't work yet, those will be added in the last (smaller) PRs 8 and 9.

So for this PR, we can already test the expected behavior for:

- Site preview URL includes token and renders the homepage correctly
- Changing or omitting that token makes Kirby render the latest version instead
- Same for any page previews
- Draft previews should still work as well (with token the draft gets resolved, without token the draft 404s completely)

### Enhancements (since alphas)

- Preview URLs for changes versions now include a `_token` query param like draft previews. This protects those previews from access by external visitors.

### Breaking changes

- Removed internal `$page->isVerified()` method in favor of internal `$page->renderVersionFromRequest()`

## Docs

*None*

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1277,7 +1277,7 @@ class App
 		if (!$page && $draft = $site->draft($path)) {
 			if (
 				$this->user() ||
-				$draft->isVerified($this->request()->get('_token'))
+				$draft->renderVersionFromRequest() !== null
 			) {
 				$page = $draft;
 			}
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -777,27 +777,6 @@ class Page extends ModelWithContent
 	}
 
 	/**
-	 * Checks if the page access is verified.
-	 * This is only used for drafts so far.
-	 * @internal
-	 */
-	public function isVerified(string|null $token = null): bool
-	{
-		if (
-			$this->isPublished() === true &&
-			$this->parents()->findBy('status', 'draft') === null
-		) {
-			return true;
-		}
-
-		if ($token === null) {
-			return false;
-		}
-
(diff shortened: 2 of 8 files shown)


bastianallgeier (commented):
I haven't done any manual tests yet, but I just wanted to leave a comment that the code looks really good to me. I love the idea for the new renderVersionFromRequest method to handle the verification. I was thinking if there could be an even more obvious name for it, but I cannot come up with something which would be more convincing :)

distantnative:
Had the same thoughts but also couldn't come up with a great name. Maybe `requestedRenderVersion`, but it's not really much different.

bastianallgeier:
It's an internal method, so we should probably really not waste too much energy on finding a better name when the current one is already perfectly ok.

bastianallgeier:
Ok, I just went through the suggested test scenarios and they all work as expected. I would merge this and then fix issues if we still encounter some.

bastianallgeier (approved):

PR #6833, opened 2024-12-05
Retry failed save and publish requests

bastianallgeier:
## Description

The model views will now show a "Retry" dialog when a save or publish request failed. 

Screenshot 2024-12-05 at 10 21 32
Screenshot 2024-12-05 at 10 21 48

The methods will be executed again after clicking the "Try again" button. 

### Summary of changes

- New `panel.content.closeRetryDialog()` method
- New `panel.content.retry()` method

### Additional context

The easiest way to test this is with the following steps. 

1. Add this to the `Kirby\Api\Controller\Changes` class at the beginning of the `save` method:
```php
if ($input === ['error' => true]) {
	throw new \Exception('Something went wrong');
}
```
2. Open the browser console and check both dialog versions like this: 
```js
// to show the retry dialog for failed save requests
panel.content.save({error: true});
// to show the retry dialog for failed publish requests
panel.content.publish({error: true});
```

### Fixes

- Fixed footer margin in dialogs if the `cancelButton` property is a string or object instead of a boolean.

### Enhancements

- New retry dialog if changes cannot be saved or published.

### For review team

### DIFF
--- a/panel/src/panel/content.js
+++ b/panel/src/panel/content.js
@@ -34,17 +34,7 @@ export default (panel) => {
 			return changes;
 		},
 
-		/**
-		 * Closes the retry dialog if it is still open
-		 */
-		closeRetryDialog() {
-			if (
-				panel.dialog.isOpen &&
-				panel.dialog.props.id === "content-retry-dialog"
-			) {
-				panel.dialog.close();
-			}
-		},
+		dialog: null,
 
 		/**
 		 * Removes all unpublished changes
@@ -158,7 +148,7 @@ export default (panel) => {
 				await panel.api.post(api + "/changes/publish", values);
 
 				// close the retry dialog if it is still open
-				this.closeRetryDialog();
+				this.dialog?.close();
 
 				// update the props for the current view
 				if (this.isCurrent(api)) {
@@ -182,8 +172,11 @@ export default (panel) => {
 			// easier to debug the issue
 			console.error(error);
 
+			// set the dialog instance
+			this.dialog = panel.dialog;
+
 			// show a dialog to the user to try again
-			panel.dialog.open({
+			this.dialog.open({
 				component: "k-text-dialog",
 				props: {
 					id: "content-retry-dialog",
@@ -195,13 +188,17 @@ export default (panel) => {
 					}
 				},
 				on: {
+					close: () => {
+						this.dialog = null;
+					},
 					submit: async () => {
-						panel.dialog.isLoading = true;
+						this.dialog.isLoading = true;
 
 						// try again with the latest state in the props
 						await this.save(panel.view.props.content, api);
 
-						panel.dialog.isLoading = false;
+						// make sure the dialog is closed if the request was successful
+						this.dialog?.close();
 
 						// give a more reassuring longer success notification
 						panel.notification.success(panel.t(`form.${method}.success`));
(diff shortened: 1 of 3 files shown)


afbora:
Works great for me 👍 Also I couldn't see a discard changes issue for this PR.

panel/src/panel/content.js
```diff
@@ -145,17 +157,59 @@ export default (panel) => {
...
+					}
+				},
+				on: {
+					submit: async () => {
+						panel.dialog.isLoading = true;
+
+						// try again with the latest state in the props
+						await this.save(panel.view.props.content, api);
+
+						panel.dialog.isLoading = true;
```
distantnative: Rather `false` here?

panel/src/panel/content.js
```diff
@@ -191,7 +248,7 @@ export default (panel) => {
 				// silent aborted requests, but throw all other errors
 				if (error.name !== "AbortError") {
 					this.isProcessing = false;
-					throw error;
+					this.retry("save", error, panel.view.props.content, api);
```
distantnative, line 247: `Question:` are we adding then in a separate PR to check the type of error for "has been locked by another user" and not retry but reload the view?

panel/src/panel/content.js
```diff
@@ -34,6 +34,18 @@ export default (panel) => {
...
+		 * Closes the retry dialog if it is still open
+		 */
+		closeRetryDialog() {
+			if (
+				panel.dialog.isOpen &&
+				panel.dialog.props.id === "content-retry-dialog"
+			) {
+				panel.dialog.close();
+			}
+		},
```
distantnative: `Thought`: I found the method not super elegant (without a openRetyDialog e.g.). But this is really nitpick and can be ignored if you disagree. Was thinking of instead adding a `dialog` property in the content module that is `null` by default but set to `panel.dialog` when the dialog was opened. Cause then one could always just to `panel.content.dialog?.close()`.

panel/src/panel/content.js
bastianallgeier: 🙈 yes of course.

panel/src/panel/content.js
bastianallgeier: Yes, we need another one where we check for the type of exception that's coming from the backend. I have a couple ideas, but haven't started yet.

panel/src/panel/content.js
bastianallgeier: I really liked the idea of the dialog prop. Already implemented :)

panel/src/panel/content.js
```diff
@@ -145,17 +147,65 @@ export default (panel) => {
...
+				},
+				on: {
+					close: () => {
+						this.dialog = null;
+					},
+					submit: async () => {
+						this.dialog.isLoading = true;
+
+						// try again with the latest state in the props
+						await this.save(panel.view.props.content, api);
```
```suggestion
						await this[method](panel.view.props.content, api);
```

Otherwise when retrying publish, it's calling save... and I don't think it should or am I missing something?

panel/src/panel/content.js
bastianallgeier: https://github.com/getkirby/kirby/pull/6835

panel/src/panel/content.js
bastianallgeier: Good catch! I just fixed it.

distantnative (approved):

PR #6835, opened 2024-12-05
New LockException and Lock Dialog

bastianallgeier:
### Merge first

- https://github.com/getkirby/kirby/pull/6833

### Summary of changes

- New `Kirby\Content\LockException`
- New `k-lock-dialog`
- New `panel.content.lockDialog` method

### Reasoning

When an editor starts editing, but someone else made changes at the same time, there can be a race condition where the editor can no longer save or publish the changes because the other one was faster. This is now handled with a new dialog to inform the editor in a good way and reload the view afterwards. 

The new LockException is used to throw a custom error with all the lock details. This is used in the Panel to fill the dialog with the necessary information.

Screenshot 2024-12-05 at 16 10 36

### Testing

The easiest way to test this is to make sure that you have two users. Let's assume, the second one has the id "editor"

1. Go to any page and start editing
2. Navigate to the changes text file in the filesystem and change the Lock: field from your user id to the "editor" user id.
3. Reload the page in the panel. The page should now be locked
4. Open the browser console and try to still save or publish changes with the following commands: 
```js
panel.content.save({subheading: 'Test'});
panel.content.publish({subheading: 'Test'});
```
The new dialog should appear and give you the right information about the lock. Closing the dialog should reload the view.

### DIFF
--- a/panel/src/components/Dialogs/index.js
+++ b/panel/src/components/Dialogs/index.js
@@ -12,6 +12,7 @@ import FilesDialog from "./FilesDialog.vue";
 import FormDialog from "./FormDialog.vue";
 import LanguageDialog from "./LanguageDialog.vue";
 import LicenseDialog from "./LicenseDialog.vue";
+import LockDialog from "./LockDialog.vue";
 import LinkDialog from "./LinkDialog.vue";
 import ModelsDialog from "./ModelsDialog.vue";
 import PageCreateDialog from "./PageCreateDialog.vue";
@@ -38,6 +39,7 @@ export default {
 		app.component("k-form-dialog", FormDialog);
 		app.component("k-license-dialog", LicenseDialog);
 		app.component("k-link-dialog", LinkDialog);
+		app.component("k-lock-dialog", LockDialog);
 		app.component("k-language-dialog", LanguageDialog);
 		app.component("k-models-dialog", ModelsDialog);
 		app.component("k-page-create-dialog", PageCreateDialog);
--- a/panel/src/components/Dialogs/index.js
+++ b/panel/src/components/Dialogs/index.js
@@ -12,7 +12,7 @@ import FilesDialog from "./FilesDialog.vue";
 import FormDialog from "./FormDialog.vue";
 import LanguageDialog from "./LanguageDialog.vue";
 import LicenseDialog from "./LicenseDialog.vue";
-import LockDialog from "./LockDialog.vue";
+import LockAlertDialog from "./LockAlertDialog.vue";
 import LinkDialog from "./LinkDialog.vue";
 import ModelsDialog from "./ModelsDialog.vue";
 import PageCreateDialog from "./PageCreateDialog.vue";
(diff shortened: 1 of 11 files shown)


panel/src/panel/content.js
```diff
@@ -212,9 +239,9 @@ export default (panel) => {
 		async save(values, api) {
 			api ??= panel.view.props.api;
 
-			if (this.isCurrent(api) === true && this.isLocked(api) === true) {
-				throw new Error("Cannot save locked changes");
-			}
+			// if (this.isCurrent(api) === true && this.isLocked(api) === true) {
+			// 	throw new Error("Cannot save locked changes");
+			// }
```
distantnative: `Todo:` above we removed this code, so remove it here as well?

src/Content/LockException.php
```diff
@@ -0,0 +1,31 @@
...
+use Kirby\Exception\LogicException;
+
+/**
+ * @package   Kirby Content
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ */
+class LockException extends LogicException
```
distantnative: `Suggestion:` rename to `LockedContentException` to fit more to our other exception classes and be a bit more descriptive.

src/Content/LockException.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Kirby\Content;
```
distantnative: `Question:` Should this rather go into `Kirby\Exception`? We do have a namespace for exceptions, but we have started also to include some exceptions in other places.

src/Content/VersionRules.php
```diff
@@ -68,7 +68,8 @@ public static function delete(
 		Language $language
 	): void {
 		if ($version->isLocked('*') === true) {
-			throw new LogicException(
+			throw new LockException(
+				lock: $version->lock('*'),
 				message: 'The version is locked and cannot be deleted'
```
distantnative: `Question:` Should we add i18n strings for these. And then we could actually use the exception message in the lock dialog.

panel/src/components/Dialogs/index.js
```diff
@@ -38,6 +39,7 @@ export default {
 		app.component("k-form-dialog", FormDialog);
 		app.component("k-license-dialog", LicenseDialog);
 		app.component("k-link-dialog", LinkDialog);
+		app.component("k-lock-dialog", LockDialog);
```
distantnative: `Suggestion:` As this sounds ambitious (could be a dialog to lock a page etc.) how about `LockedContentDialog` or `LockAlertDialog`?

src/Content/LockException.php
bastianallgeier: I thought about this as well and I cannot decide which way I like better.

panel/src/components/Dialogs/index.js
bastianallgeier: I like LockAlertDialog

src/Content/LockException.php
bastianallgeier: I'm fine with that.

panel/src/panel/content.js
bastianallgeier: I'm still not entirely sure about it to be honest. We could still keep the client-side checks in all methods and pass the view.props.lock object to the lockDialog method here. On the other hand, the result from the request will always be a bit more up to date.

bastianallgeier:
I guess it makes sense to merge this first: https://github.com/getkirby/kirby/pull/6839 and then continue with this

distantnative:
@bastianallgeier I added the suggested changes, please have a look if it's fine for you.

bastianallgeier:
@distantnative looks great!

PR #6837, opened 2024-12-05
Preview tokens 9: Blueprint-based custom preview URLs

lukasbestle:
## 🚨 Merge first

- #6836

### Summary of changes

Properly handle string values for the `preview` blueprint option by extracting the URI from the custom string and generating a token for it; external URLs (that don't start with the base URL of the current site) won't get a token as we cannot generate one for external environments

### Reasoning

Backwards-compatibility & Customization

Extra logic is needed only now in v5 because for drafts it wasn't such a big deal whereas custom version preview URLs would be completely broken without it.

### Additional context

This is the final PR of this series 🎉 

## Changelog

*None (already covered by PR 7)*

## Docs

*None*

## Ready?

### For review team


src/Content/Version.php
```diff
@@ -431,8 +432,13 @@ protected function previewTokenFromUrl(string $url): string|null
 			return null;
 		}
 
+		// get rid of all modifiers after the path
+		$uri = new Uri($url);
+		$uri->fragment = null;
+		$uri->query    = new Query([]);
```
bastianallgeier: Do we also need to get rid of our custom params here?

src/Content/Version.php
```diff
@@ -590,24 +596,56 @@ public function url(): string|null
...
 
+		// fall back to the URL as defined in the blueprint
+		return $url;
+	}
+
+	/**
+	 * Assembles the preview URL with the added `_token` and `_version`
+	 * query params, no matter if the base URL already contains query params
+	 */
+	protected function urlParams(string $baseUrl, string $token): string
```
bastianallgeier: `Nitpick`: We could call this `urlWithQueryParams` for a bit more clarity. I think it also fits `urlFromOption` a bit better.

PR #6840, opened 2024-12-06
Fix sqlite driver using double quotes for string literals

### LINKED ISSUE #6769: SQLite 3.41.0 double-quoted string literals are disabled by default
Hi all!

I ran into a nasty bug, where any SQLite Queries failed with `Invalid table` and it took a long time debugging it ;)
After talking to the hoster, I learned that as of SQLite 3.41.0 double-quoted string literals are [disabled by default](https://www.sqlite.org/quirks.html#double_quoted_string_literals_are_accepted) and therefore Kirby's [check](https://github.com/getkirby/kirby/blob/eb47158e91e74184ab291ede020664f1d2829f74/src/Database/Sql/Sqlite.php#L133) if the table exists fails.

Currently I don't have a good suggestion to you as to how to verify this, but I guess "just" changing it to single-quoted literals might fix this?

**Kirby Version**  
afaik any, as the [current implementation in this repo](https://github.com/getkirby/kirby/blob/eb47158e91e74184ab291ede020664f1d2829f74/src/Database/Sql/Sqlite.php#L133) still has double-quoted  string literals

Best
Malte

rasteiner:
## Description

Fix the table discovery query in the sqlite class using double quotes for string literals. 

### Summary of changes
 - Using single quotes in sqlite table discovery query fixes #6769

### Reasoning
Double quotes to enclose string literals are *sometimes* supported by sqlite configurations (mainly for backwards compatibility).
Single quotes are *always* supported. 

### Additional context
https://www.sqlite.org/quirks.html#double_quoted_string_literals_are_accepted

## Changelog
 - Using single quotes in sqlite table discovery query fixes #6769

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Database/Sql/Sqlite.php
+++ b/src/Database/Sql/Sqlite.php
@@ -130,7 +130,7 @@ class Sqlite extends Sql
 	public function tables(): array
 	{
 		return [
-			'query'    => 'SELECT name FROM sqlite_master WHERE type = "table" OR type = "view"',
+			'query'    => 'SELECT name FROM sqlite_master WHERE type = \'table\' OR type = \'view\'',
 			'bindings' => []
 		];
 	}
--- a/tests/Database/Sql/SqliteTest.php
+++ b/tests/Database/Sql/SqliteTest.php
@@ -160,7 +160,7 @@ class SqliteTest extends TestCase
 	public function testTables()
 	{
 		$result = $this->sql->tables();
-		$this->assertSame('SELECT name FROM sqlite_master WHERE type = "table" OR type = "view"', $result['query']);
+		$this->assertSame('SELECT name FROM sqlite_master WHERE type = \'table\' OR type = \'view\'', $result['query']);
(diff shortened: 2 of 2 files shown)


rasteiner:
PS: the tests actually pass, I think they timed out here :/

afbora:
@rasteiner Thank you for the PR. I've re-run the tests and all passed 🎉 

I'm just curious about the [Sqlite::quoteIdentifier()](https://github.com/getkirby/kirby/blob/4.5.0/src/Database/Sql/Sqlite.php#L112-L124) method. This method also uses double-quotes. Does this method also need to be updated?

rasteiner:
The quoteIdentifier function is fine, as it correctly puts the quotes around identifiers, such as table or column names and not around string literals.

bastianallgeier (approved):

bastianallgeier:
Thanks for the PR!

PR #6842, opened 2024-12-07
Batch editing

bastianallgeier:
### Features

- New option for pages and files sections, to delete multiple files or pages at once https://kirby.nolt.io/38
batch-delete

### Enhancements

- New `Kirby\Cms\Pages::delete(array $ids)` and `Kirby\Cms\Files::delete(array $ids)` methods to handle batch delete for multiple files in a collection. 
- New batch mixin for model sections (pages & files), which introduces the batch option and a computed property to check for supported layouts. 
- New `selectable` property for the `<k-item>` and `<k-items>` components. When set to true, the items show checkboxes and emit an select event, when the checkbox is clicked. 
- The `files` and `pages` sections introduce new `delete` API endpoints. Those endpoints can take a set of page or file ids and will delete each one. Errors will be caught and sent back in the details array of the final thrown exception. 

### Breaking changes

- none

## Docs

- New `batch` option for pages and files sections
```yaml
projects: 
  type: pages
  batch: true
```

## Ready?

### For review team

### DIFF
--- a/config/sections/mixins/batch.php
+++ b/config/sections/mixins/batch.php
@@ -1,5 +1,9 @@
 <?php
 
+use Kirby\Exception\Exception;
+use Kirby\Exception\PermissionException;
+use Kirby\Toolkit\I18n;
+
 return [
 	'props' => [
 		/**
@@ -9,4 +13,33 @@ return [
 			return $batch;
 		},
 	],
+	'methods' => [
+		'deleteSelected' => function (array $ids): bool {
+			if ($ids === []) {
+				return true;
+			}
+
+			// check if batch deletion is allowed
+			if ($this->batch() === false) {
+				throw new PermissionException(
+					message: 'The section does not support batch actions'
+				);
+			}
+
+			$min = $this->min();
+
+			// check if the section has enough items after the deletion
+			if ($this->total() - count($ids) < $min) {
+				throw new Exception(
+					message: I18n::template('error.section.' . $this->type() . '.min.' . I18n::form($min), [
+						'min'     => $min,
+						'section' => $this->headline()
+					])
+				);
+			}
+
+			$this->models()->delete($ids);
+			return true;
+		}
+	]
 ];
--- a/config/sections/mixins/batch.php
+++ b/config/sections/mixins/batch.php
@@ -9,9 +9,4 @@ return [
 			return $batch;
 		},
 	],
-	'computed' => [
-		'batch' => function () {
-			return $this->layout === 'table' ? false : $this->batch;
-		}
-	],
 ];
--- a/config/sections/mixins/batch.php
+++ b/config/sections/mixins/batch.php
@@ -1,0 +1,17 @@
+<?php
+
+return [
+	'props' => [
+		/**
+		 * Activates the batch delete option for the section
+		 */
+		'batch' => function (bool $batch = false) {
+			return $batch;
+		},
+	],
+	'computed' => [
+		'batch' => function () {
+			return $this->layout === 'table' ? false : $this->batch;
+		}
+	],
+];
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -4,9 +4,7 @@ use Kirby\Cms\Blueprint;
 use Kirby\Cms\Page;
 use Kirby\Cms\Pages;
 use Kirby\Cms\Site;
-use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
-use Kirby\Exception\PermissionException;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\I18n;
 
@@ -325,30 +323,9 @@ return [
 				'pattern' => 'delete',
 				'method'  => 'DELETE',
 				'action'  => function () {
-					$section = $this->section();
-
-					// check if batch deletion is allowed
-					if ($section->batch() === false) {
-						throw new PermissionException(
-							message: 'The section does not support batch actions'
-						);
-					}
-
-					$ids = $this->requestBody('ids');
-					$min = $section->min();
-
-					// check if the section has enough pages after the deletion
-					if ($section->total() - count($ids) < $min) {
-						throw new Exception(
-							message: I18n::template('error.section.pages.min.' . I18n::form($min), [
-								'min'     => $min,
-								'section' => $section->headline()
-							])
-						);
-					}
-
-					$section->models()->delete($ids);
-					return true;
+					return $this->section()->deleteSelected(
+						ids: $this->requestBody('ids'),
+					);
 				}
 			]
 		];
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -335,11 +335,15 @@ return [
 					}
 
 					$ids = $this->requestBody('ids');
+					$min = $section->min();
 
 					// check if the section has enough pages after the deletion
-					if ($section->total() - count($ids) < $section->min()) {
+					if ($section->total() - count($ids) < $min) {
 						throw new Exception(
-							message: 'The section needs to have at least ' . $section->min() . ' pages',
+							message: I18n::template('error.section.pages.min.' . I18n::form($min), [
+								'min'     => $min,
+								'section' => $section->headline()
+							])
 						);
 					}
 
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -184,10 +184,11 @@ return [
 					'link'        => $panel->url(true),
 					'parent'      => $page->parentId(),
 					'permissions' => [
-						'sort'         => $permissions->can('sort'),
+						'delete'       => $permissions->can('delete'),
 						'changeSlug'   => $permissions->can('changeSlug'),
 						'changeStatus' => $permissions->can('changeStatus'),
 						'changeTitle'  => $permissions->can('changeTitle'),
+						'sort'         => $permissions->can('sort'),
 					],
 					'status'      => $page->status(),
 					'template'    => $page->intendedTemplate()->name(),
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -4,8 +4,9 @@ use Kirby\Cms\Blueprint;
 use Kirby\Cms\Page;
 use Kirby\Cms\Pages;
 use Kirby\Cms\Site;
-use Kirby\Exception\InvalidArgumentException;
 use Kirby\Exception\Exception;
+use Kirby\Exception\InvalidArgumentException;
+use Kirby\Exception\PermissionException;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\I18n;
 
@@ -325,8 +326,16 @@ return [
 				'method'  => 'DELETE',
 				'action'  => function () {
 					$section = $this->section();
-					$ids     = $this->requestBody('ids');
-					$errors  = [];
+
+					// check if batch deletion is allowed
+					if ($section->batch() === false) {
+						throw new PermissionException(
+							message: 'The section does not support batch actions'
+						);
+					}
+
+					$ids    = $this->requestBody('ids');
+					$errors = [];
 
 					// check if the section has enough pages after the deletion
 					if ($section->total() - count($ids) < $section->min()) {
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -5,6 +5,7 @@ use Kirby\Cms\Page;
 use Kirby\Cms\Pages;
 use Kirby\Cms\Site;
 use Kirby\Exception\InvalidArgumentException;
+use Kirby\Exception\Exception;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\I18n;
 
@@ -322,9 +323,26 @@ return [
 				'pattern' => 'delete',
 				'method'  => 'DELETE',
 				'action'  => function () {
+					$errors = [];
+
 					foreach ($this->requestBody('ids') as $id) {
-						$this->section()->kirby()->page($id)?->delete();
+						try {
+							$this->section()->kirby()->page($id)?->delete();
+						} catch (Throwable $e) {
+							$errors[] = [
+								'label'  => $id, 
+								'message' => $e->getMessage()
+							];
+						}
 					}
+
+					if ($errors !== []) {
+						throw new Exception(
+							message: 'Not all pages could be deleted',
+							details: $errors
+						);
+					}
+
 					return true;
 				}
 			]
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -323,11 +323,21 @@ return [
 				'pattern' => 'delete',
 				'method'  => 'DELETE',
 				'action'  => function () {
-					$errors = [];
+					$section = $this->section();
+					$ids     = $this->requestBody('ids');
+					$errors  = [];
 
-					foreach ($this->requestBody('ids') as $id) {
+					// check if the section has enough pages after the deletion
+					if ($section->total() - count($ids) < $section->min()) {
+						throw new Exception(
+							message: 'The section needs to have at least ' . $section->min() . ' pages',
+						);
+					}
+
+					// delete all pages and collect errors
+					foreach ($ids as $id) {
 						try {
-							$this->section()->kirby()->page($id)?->delete();
+							$section->kirby()->page($id)?->delete();
 						} catch (Throwable $e) {
 							$errors[] = [
 								'label'  => $id, 
@@ -336,6 +346,7 @@ return [
 						}
 					}
 
+					// throw an error if not all pages could be deleted
 					if ($errors !== []) {
 						throw new Exception(
 							message: 'Not all pages could be deleted',
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -150,25 +150,26 @@ return [
 				$pages = $pages->flip();
 			}
 
+			return $pages;
+		},
+		'modelsPaginated' => function () {
 			// pagination
-			$pages = $pages->paginate([
+			return $this->models()->paginate([
 				'page'   => $this->page,
 				'limit'  => $this->limit,
 				'method' => 'none' // the page is manually provided
 			]);
-
-			return $pages;
 		},
 		'pages' => function () {
 			return $this->models;
 		},
 		'total' => function () {
-			return $this->models->pagination()->total();
+			return $this->models()->count();
 		},
 		'data' => function () {
 			$data = [];
 
-			foreach ($this->models as $page) {
+			foreach ($this->modelsPaginated() as $page) {
 				$panel       = $page->panel();
 				$permissions = $page->permissions();
 
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -11,6 +11,7 @@ use Kirby\Toolkit\I18n;
 
 return [
 	'mixins' => [
+		'batch',
 		'details',
 		'empty',
 		'headline',
@@ -340,7 +341,7 @@ return [
 							$section->kirby()->page($id)?->delete();
 						} catch (Throwable $e) {
 							$errors[] = [
-								'label'  => $id, 
+								'label'  => $id,
 								'message' => $e->getMessage()
 							];
 						}
@@ -366,6 +367,7 @@ return [
 			'errors'  => $this->errors,
 			'options' => [
 				'add'      => $this->add,
+				'batch'    => $this->batch,
 				'columns'  => $this->columnsWithTypes(),
 				'empty'    => $this->empty,
 				'headline' => $this->headline,
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -315,6 +315,22 @@ return [
 			return $blueprints;
 		},
 	],
+	// @codeCoverageIgnoreStart
+	'api' => function () {
+		return [
+			[
+				'pattern' => 'delete',
+				'method'  => 'DELETE',
+				'action'  => function () {
+					foreach ($this->requestBody('ids') as $id) {
+						$this->section()->kirby()->page($id)?->delete();
+					}
+					return true;
+				}
+			]
+		];
+	},
+	// @codeCoverageIgnoreEnd
 	'toArray' => function () {
 		return [
 			'data'    => $this->data,
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -334,8 +334,7 @@ return [
 						);
 					}
 
-					$ids    = $this->requestBody('ids');
-					$errors = [];
+					$ids = $this->requestBody('ids');
 
 					// check if the section has enough pages after the deletion
 					if ($section->total() - count($ids) < $section->min()) {
@@ -344,26 +343,7 @@ return [
 						);
 					}
 
-					// delete all pages and collect errors
-					foreach ($ids as $id) {
-						try {
-							$section->kirby()->page($id)?->delete();
-						} catch (Throwable $e) {
-							$errors[] = [
-								'label'  => $id,
-								'message' => $e->getMessage()
-							];
-						}
-					}
-
-					// throw an error if not all pages could be deleted
-					if ($errors !== []) {
-						throw new Exception(
-							message: 'Not all pages could be deleted',
-							details: $errors
-						);
-					}
-
+					$section->models()->delete($ids);
 					return true;
 				}
 			]
--- a/panel/src/components/Sections/ModelsSection.vue
+++ b/panel/src/components/Sections/ModelsSection.vue
@@ -105,7 +105,7 @@ export default {
 						this.$panel.dialog.open({
 							component: "k-remove-dialog",
 							props: {
-								text: `Do you really want to delete ${this.selected.length} items at once? This action cannot be undone.`
+								text: this.confirmDeleteSelectedMessage
 							},
 							on: {
 								submit: () => {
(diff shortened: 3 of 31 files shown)


config/sections/mixins/batch.php
```diff
@@ -0,0 +1,17 @@
...
+		/**
+		 * Activates the batch delete option for the section
+		 */
+		'batch' => function (bool $batch = false) {
+			return $batch;
+		},
+	],
+	'computed' => [
+		'batch' => function () {
+			return $this->layout === 'table' ? false : $this->batch;
```
distantnative: `Question:` Did you not allow this for table layout just because there is no selectable UI for the table layout yet? Or other reasons as well? 

If the former, I think the table layout is actually the one where people request most batch actions, so I would think we should then rather spend the time and implement selections for the table layout as well.

config/sections/pages.php
```diff
@@ -315,12 +318,49 @@
...
+						throw new Exception(
+							message: I18n::template('error.section.pages.min.' . I18n::form($min), [
+								'min'     => $min,
+								'section' => $section->headline()
+							])
+						);
+					}
+
+					$section->models()->delete($ids);
+					return true;
```
distantnative: `Todo:` This code is almost the same as in the files section. I think we should DRY this with a method included in the `batch` mixin.

panel/src/components/Sections/ModelsSection.vue
```diff
@@ -91,6 +95,46 @@ export default {
...
+				} else {
+					buttons.push({
+						icon: "trash",
+						text: this.$t("delete") + ` (${this.selected.length})`,
+						theme: "negative",
+						click: () => {
+							this.$panel.dialog.open({
+								component: "k-remove-dialog",
+								props: {
+									text: `Do you really want to delete ${this.selected.length} items at once? This action cannot be undone.`
```
distantnative: `Todo:` Turn into i18n string.

panel/src/components/Sections/ModelsSection.vue
```diff
@@ -120,6 +172,9 @@ export default {
 		canSearch() {
 			return this.options.search;
 		},
+		canSelect() {
+			return this.options.batch && this.items.length > 0;
```
distantnative, line 172: `Question:` Should we offer selection mode for sections with a single item?

panel/src/components/Sections/ModelsSection.vue
```diff
@@ -231,6 +309,24 @@ export default {
 			this.searching = !this.searching;
 			this.searchterm = null;
 		},
+		onSelect(event, item) {
+			if (event.target.checked) {
```
distantnative: `Todo:` Making this depend on the event (and more specifically a checkbox event) that many layers up from the actual input seems not ideal to me. Can this just accept the item as parameter and decide what to do based on whether the item is already in the selected array or not?

panel/src/components/Sections/ModelsSection.vue
```diff
@@ -231,6 +309,24 @@ export default {
...
+				);
+			}
+		},
+		onSelectToggle() {
+			if (this.isSelecting) {
+				this.selected = [];
+				this.isSelecting = false;
+			} else {
+				this.isSelecting = true;
+			}
```
distantnative: `Suggestion:`
```suggestion
			this.isSelecting = !this.isSelecting;
			
			// start new selection with a fresh list
			if (this.isSelecting) {
				this.selected = [];
			}
```

src/Cms/Files.php
```diff
@@ -93,6 +96,40 @@ public function changeSort(array $files, int $offset = 0): static
...
+				$exceptions[$id] = $e;
+			}
+		}
+
+		if ($exceptions !== []) {
+			throw new Exception(
+				key: 'file.delete.multiple',
+			);
+		}
+	}
```
distantnative, line 132: `Discussion:` I of course understand why it is implemented like this currently, but for a core method in this class, it seems a bit unusual to swallow the actual exceptions and just throw this vague one. Not sure there is a better way.

config/sections/mixins/batch.php
bastianallgeier: Yes, the main reason is that I haven't found the time yet to implement it for table rows. It should not be too difficult though.

panel/src/components/Sections/ModelsSection.vue
bastianallgeier: Probably not. It feels pretty weird.

config/sections/mixins/batch.php
distantnative: I also just had a look at the table component and it shouldn't be that hard. As we could reuse the index column for the selection input.

panel/src/components/Sections/ModelsSection.vue
lukasbestle: Does it do harm? I think for consistency it would still be useful. E.g. a user wants to delete all items and is used to using the batch editing mode for this. They can still use it with just one item. If we just hide the button, the user could be confused.

afbora:
I just wondering if it wouldn't be better if `batch` mode was enabled by default. I don't think it would be considered a breaking change to enable this feature. After all, it's a nice and useful feature.

afbora:
Getting `Not all pages could be deleted` dialog error when I select a non-deletable pages. I have to reload the page after the error to see remains.

Ideal solution: shouldn't be able to select non-deletable models.

afbora:
When batch is active, the status icon is still clickable. If you click the status icon, both the status dialog opens and batch selection works. A user may accidentally delete the model while trying to change the status while batch mode is active.

Ideal solutions:
1. When the status icon is clicked, the status dialog should not open/work.
2. If the status icon should work when batch mode is active, it should not trigger the batch mode checkbox.

bastianallgeier:
@afbora that's wrong. Reloading should not be necessary. Already blocking the checkbox would be ideal indeed, but also a lot harder to implement. That's why I went for a simpler first solution that just skips the ones that cannot be deleted during the request.

afbora:
@bastianallgeier I've tested again and page is not reloading after the error.

bastianallgeier:
I've added table selection and it's working great

https://github.com/user-attachments/assets/3754f99e-33d3-4a0a-bbe8-e16f7087397f

afbora:
@bastianallgeier Is it testable for bug-fixes?

bastianallgeier:
@afbora no, I still need to look into the exception issue.

bastianallgeier:
@afbora I've fixed it now and it's working fine for me.

config/sections/pages.php
bastianallgeier: I've moved it all into the batch mixin. Works great and fits very well.

panel/src/components/Sections/ModelsSection.vue
bastianallgeier: I've thought about it again and with the opt-in character of the feature I start to think that we should just keep it, even if there's only one option.

afbora:
@bastianallgeier Deletion doesn't work with paginated sections.

````yaml
pages:
  type: pages
  batch: true
  limit: 2
````

**Comment by pReya** [comment]
> I've added table selection and it's working great
> CleanShot.2024-12-10.at.12.19.59.mp4

Sorry for a slightly Off Topic remark: But I see you're touching the "ID" column of table views and conditionally using them as a checkbox for multi-select. I know of multiple projects that are hiding the ID column (because it's [the only way](https://forum.getkirby.com/t/hide-number-on-structure-table/26872/3) to get rid of this column as it's not configurable). Is there any way to hide the column and still be able to use this new multi-select possibility?

**Comment by bastianallgeier** [comment]
@afbora can you give more context. What happens in that case?

**Comment by bastianallgeier** [comment]
@pReya if the id column is disabled, it should still be visible when the select mode is on.

**Comment by pReya** [comment]
> @pReya if the id column is disabled, it should still be visible when the select mode is on.

AFAIK the ID column can't be disabled. It needs to be hidden via CSS (see linked forum post). So what I'm trying to say is: While you're working on this – is there any chance, you can allow disabling/hiding it while still allowing multi-select? See also https://kirby.nolt.io/460

**Comment by afbora** [comment]
@bastianallgeier Nothing happened. Just getting `Not all pages could be deleted` error and no pages are deleted. Even if they are deletable pages.

**Comment by nilshoerrmann** [comment]
Two questions regarding the selection layout:

1. Looking at the cards layout, I was wondering if using colours to highlight the selection would be a good idea. E. g. adding a blue outline or even changing the card colour (thinking of how selections work in Finder or Explorer). Right now, because the checkboxes are not that visually attached to the cards, it's hard to see the current selection easily. The checkbox' active state seen in the screenshot makes this very clear: it looks like the checkbox is active, not the card – but the user is interacting with the full card and not this single checkbox.
2. And regarding the table layout – I'm sure I already reported that a few years ago – I am always wondering: Why are  interactions like dragging and now selecting attached to the first _cell_? Having the action inside a cell makes it look like a data entry actually. But again, we are interacting with the entire _row_ and for the list layout the drag icon is visually prefixing the item already, hovering on the grey background. It would be great to have a consistent behaviour across layout variants here in my eyes.

A quick visualisation:

kirby-table

I'm really fond of the introduction of batch editing. But I'd personally prefer a bolder design for destructive interactions like this to clearify the impact of the action. That's why I'd like to suggest the inverted colours here.

grafik

**Comment by distantnative** [comment]
@nilshoerrmann if you have some good tips how to implement drag handles outside of a `<table>` element, please let us know. We tried https://github.com/getkirby/kirby/issues/4386 but so far without success.

**Comment by nilshoerrmann** [comment]
@distantnative I would put the handle inside the first cell (which ever that is in the current table) and move it outside using `translate`. I haven't tested this right now but I think I have done that this way before.

**Comment by nilshoerrmann** [comment]
I've just tried in the lab, quickly altering the CSS:

grafik

The only problem might be that you have to remove the overflow handling:

grafik

Not sure why it's there to be honest. So that would be interesting to know, if that really has to stay.

**Comment by nilshoerrmann** [comment]
Oh, the `overflow: hidden` is for text shortening with an ellipsis of course (sorry, I didn't notice before because we usually remove that behaviour with custom CSS). In that case an additional wrapper would be needed inside the cell or it would be possible to to create an "invisible" cell at the start of the table (zero width, no background and borders) to place the drag handle inside and allow overflow for that cell specifically.

**Comment by nilshoerrmann** [comment]
By the way, I've read in the linked issue about the issues with hover interactions: as long as you don't remove the handle from the content flow with `display: none` this works perfectly. Setting `opacity: 0` by default, switching to `opacity: 1` on row hover makes this visually consistent with the list layouts.

**Comment by nilshoerrmann** [comment]
@distantnative If you apply the following CSS to https://lab.getkirby.com/public/lab/components/items/4_table the drag handle will be outside of the table with sorting functionality intact, showing on hover:

```css
.k-table .k-table-index-column {
  /* ↑ this would need a dedicated first column with a different name, e. g. .k-table-sorting-column */
  overflow: auto;
  position: relative;
  width: 0;
  padding: 0;
}

.k-table .k-table-index-column .k-sort-handle {
  /* display: none; ← remove this */
  position: absolute;
  translate: -2rem;
  left: 0;
  top: 0;
  opacity: 0;
  width: 2rem;
}

.k-table thead th:first-child {
  position: initial;
}

.k-table .k-table-index {
  display: none; /* ← this removes the index number which would have to move to the second column instead */
}

.k-table .k-table-sortable-row:hover .k-sort-handle {
  opacity: 1;
}
```

In this case the first index column will be visually hidden. So it's a proof of concept that this works with a "ghost" column. Tested in Firefox and Safari.

There is a display error with images inside the table when sorting (outline cut off by image) but that's unrelated to the above styles.

johannahoerrmann:
Another issue I see with placing the select in the ID cell (and thus hiding the number) is – numbers are very useful to see during batch edit if they are something editors actually work with. The number is the clearest and easiest indicator for specific rows, so I do see the use case of eg. jotting down or agreeing on numbers of the items to delete (or otherwise edit, once that's possible). Pretty much the exact opposite use case of what @pReya wrote, but the same issue (showing select in the space reserved for numbers).

bastianallgeier:
@afbora the problem comes from the way that batch deletion works. It only deletes items from a given collection. When we apply pagination, the collection does no longer contain all models. I've fixed it by introducing models and modelsPaginated in files and pages sections. This way, we can access either the full collection or the paginated one by demand.

panel/src/components/Sections/ModelsSection.vue
bastianallgeier: You are right. I've refactored it to only send the item and item index.

bastianallgeier:
@nilshoerrmann @johannahoerrmann it's a first step and I didn't go as dark as you suggested, but it's already making the select state a lot clearer. 

Screenshot 2024-12-13 at 14 04 18
Screenshot 2024-12-13 at 12 44 13
Screenshot 2024-12-13 at 12 44 21
Screenshot 2024-12-13 at 14 04 08

panel/src/components/Sections/ModelsSection.vue
bastianallgeier: ☑️

nilshoerrmann:
@bastianallgeier Thanks for the added colour highlighting, that's much clearer. 

I still find the checkbox placement very confusing because I think this should be a prefix. But having them inside and either left _or_ right in the interface doesn't make sense to me at all. But I must say that this never worked for me in page or file selection dialogs as well where the checkboxes or radios are right aligned as well. I guess it's because of the image previews that you moved those right in the list layout but it feels quite unorthodox.

bastianallgeier:
@nilshoerrmann it's more of a practical decision at this point. I don't find it ideal either. I think your table solution is much better. But the batch mode is already a bit challenging and sometimes it's really just about keeping additional challenges for later. So it's more like a "good enough" solution. Not a perfect one.

src/Cms/Files.php
bastianallgeier: All exceptions are now collected and displayed properly in the error dialog …

Screenshot 2024-12-13 at 17 22 35

distantnative:
@nilshoerrmann I tried your suggestion https://github.com/getkirby/kirby/tree/v5/enhancement/table-sort-column - but while visually mostly right (besides the outline), it renders the sorting functionality broken. The sorting sometimes happens not at all or entries jump around instead of where being dragged. That's what I experienced in the linked issue before. (thanks a lot to trying to help with all your efforts!!! it just seems to be a lot harder than this)

nilshoerrmann:
I can take a look into this over the winter holidays. This should be fixable.

bastianallgeier:
From my point of view, it would be good to keep the sort handle part out of this PR. I see that as an enhancement but not a critical part why this feature should be blocked. The rest of the PR is ready from my perspective.

afbora:
@bastianallgeier There is still pagination issue:

- Enable pagination
- For ex: go to page 2
- Select a page to delete
- Go to page 1 and go to page 2 again
- Selection is gone

Deletion still works great just checkbox selection is missing while switching pages.

afbora:
Also why don't we integrate `deletable` property for each item to the panel. So we don't have to select the pages that cannot be deleted. I implemented it in a simple way, it works fine. Or will it cause a performance problem?

### My test steps:

- Add `delete` permission here:
````php
'delete' => $permissions->can('delete'),
````
https://github.com/getkirby/kirby/blob/v5/batch-editing/config/sections/pages.php#L186-L191

- Modify this like that: (I know this is section `deleteable` permission. But it also sends to pages. We need to separate that.)
````js
const deletable = page.permissions.delete && this.data.length > this.options.min;
````
https://github.com/getkirby/kirby/blob/v5/batch-editing/panel/src/components/Sections/PagesSection.vue#L14

- Finally add `deletable` boolean prop and add `deletable` variable to `<label v-if>` 
````js
<label v-if="selectable && deletable" ..>
````
https://github.com/getkirby/kirby/blob/v5/batch-editing/panel/src/components/Collection/Item.vue#L56

I did this experiment just to test if I can move the checkbox of the page's `delete` permission. I know it is not the right implementation. But it will give you an idea.

### Current view:

v5-develop test_panel_site (1)

**Comment by bastianallgeier** [comment]
I've added test setups to the sandbox https://github.com/getkirby/kirby/pull/6842

File: panel/src/components/Sections/ModelsSection.vue [hunk 8]
```diff
@@ -185,11 +240,40 @@ export default {
...
+						ids: this.selected.map((item) => item.id)
+					}
+				);
+			} catch (error) {
+				this.$panel.notification.error(error);
+			} finally {
+				this.reload();
+				this.isSelecting = false;
+				this.isProcessing = false;
+				this.selected = [];
```
distantnative: `Todo:` We need to emit a global event (so far `model.update` but maybe we add a more specific one), so that also other files sections reload. Currently, with the new sandbox examples, you can delete files and you will see that the other sections on the same page still show the deleted files.

bastianallgeier: Sending the model.update event is a nice way to also remove the reload method because that basically already happens then anyway.

bastianallgeier:
@afbora sorry, I completely missed your comment. The solution is great and I've now fully implemented it. But I've also changed the selection state for items and table rows again. There's now a `selecting` property to indicate that the checkbox should be shown, but there's also the `selectable` prop to tell if the item can actually be deleted. This ensures that the checkbox is still there, but deactivated when an item is not selectable. I think this is much clearer for the UX. Looks like this …

Screenshot 2025-01-21 at 18 07 36

bastianallgeier:
I've added better examples with undeletable files and pages to the sandbox https://github.com/getkirby/kirby/pull/6842

afbora:
1. For dark mode; non-deletable pages (about and home pages for the example ss) color is more obvious (white rounded) more than deletable pages (gray). Looks great for light mode.
v5-develop test_panel_site (3)

2. After deleting a page, deleted item is still there. AFAIR you was fixed but still broken. It will be gone after refresh the page. 

3. I've tried to delete the `notes` page (have children), getting `The page has subpages and cannot be deleted`. So why I can able select the notes page? I think I shouldn't be able to select any pages that have children. So `deletable` also should means have children.

4. With pagination: checkbox state broken. Select any item to batch deleting. Change the pagination page and go back to previous page. See the issue that the item is no longer selected.

bastianallgeier:
@afbora 

1. the disabled state in dark mode definitely needs fixing
2. that's so so confusing. I don't have that issue at all anymore
3. I don't think we can check for all subpages as well in addition to the permissions. That could lead to a big bottleneck. We could either add a checkbox to force delete all pages with subpages or we keep ignoring them here as it already is. 
4. I think that we should move out of selection mode on pagination to be honest. To remember the selected IDs and keep them when you are navigating back seems way too complex for this first version.

jirichlebus:
Hi there, I apologize for not having a direct contribution to the topic, but I wanted to ask if I can also test the batch select somehow? I have Kirby 5 beta.2 version.

Batch select is mentioned to me very often by clients and I'm stoked that it's coming.

distantnative:
@jirichlebus You would have to check out the branch of this PR and get Panel running in dev setup :D Maybe a bit too much. But this is still in review and not part of any beta yet. Let's see, we would love to finalize it for beta.3.

distantnative:
@bastianallgeier disabled state in dark mode to be fixed by https://github.com/getkirby/kirby/pull/6947

I have the same problem as Ahmet that section isn't refreshed after deletion. Just taking your sandbox branch, load it in lab and delete any page on https://sandbox.test/panel/pages/sections+pages via batch delete.

EDIT: found the issue - only files section was listening for `model.update`. I fear Basti was testing with files sections while Ahmet and I with pages sections.

distantnative:
Suddenly I also get this:

Screenshot 2025-01-24 at 21 01 53

And I don't know why.

panel/src/components/Sections/ModelsSection.vue
```diff
@@ -91,6 +95,39 @@ export default {
...
+				buttons.push({
+					disabled: this.selected.length === 0,
+					icon: "trash",
+					text: this.$t("delete") + ` (${this.selected.length})`,
+					theme: "negative",
+					click: () => {
+						this.$panel.dialog.open({
+							component: "k-remove-dialog",
+							props: {
+								text: this.confirmDeleteSelectedMessage
```
distantnative:
```suggestion
								text: this.$t(`${this.type}.delete.confirm.selected`, {
										count: this.selected.length
								});
```

The computed prop is only ever used here. I'd prefer to keep it here and not add an additional computed prop.

panel/src/components/Sections/ModelsSection.vue
```diff
@@ -129,10 +178,16 @@ export default {
...
+				selecting: !this.isProcessing && this.isSelecting,
 				sortable: !this.isProcessing && this.options.sortable,
 				size: this.options.size
 			};
 		},
+		confirmDeleteSelectedMessage() {
+			return this.$t(`${this.type}.delete.confirm.selected`, {
+				count: this.selected.length
+			});
+		},
```
distantnative:
```suggestion
```

See above.

panel/src/components/Sections/ModelsSection.vue
```diff
@@ -185,11 +240,42 @@ export default {
 			this.reload();
 		}
 	},
+	created() {
+		this.$events.on("model.update", this.reload);
+		this.$events.on("selecting", this.stopSelectingCollision);
```
distantnative: `Suggestion:` This feels like a hack we won't like down the road. What are you trying to achieve here? That I cannot enter select mode in two sections in parallel, select the same pages, then delete them in one and in therefore the other section's state getting out of sync? If so, I would suggest to rather include in the `reload()` method to reset the selecting state (so `this.stopSelecting()`). Since this will be triggered by the `model.update` event, even if I select too sections in parallel, once I delete the models in one section, the other section gets reset and reloaded.

panel/src/components/Sections/ModelsSection.vue
```diff
@@ -185,11 +240,42 @@ export default {
...
+					{
+						ids: this.selected.map((item) => item.id)
+					}
+				);
+			} catch (error) {
+				this.$panel.notification.error(error);
+			} finally {
+				this.$panel.events.emit("model.update");
+				this.selected = [];
+				this.isSelecting = false;
```
distantnative:
```suggestion
				this.stopSelecting();
```

panel/src/components/Sections/ModelsSection.vue
```diff
@@ -223,8 +309,12 @@ export default {
 		onAdd() {},
 		onChange() {},
 		onDrop() {},
-		onSort() {},
 		onPaginate(pagination) {
+			// reset batch mode
+			this.isSelecting = false;
+			this.selected = [];
```
distantnative:
```suggestion
			this.stopSelecting();
```

panel/src/components/Sections/ModelsSection.vue
```diff
@@ -233,6 +323,33 @@ export default {
...
+			if (this.selected.includes(item)) {
+				this.selected = this.selected.filter(
+					(selected) => selected.id !== item.id
+				);
+			} else {
+				this.selected.push(item);
+			}
+		},
+		onSelectToggle() {
+			this.isSelecting ? this.stopSelecting() : this.startSelecting();
```
distantnative, line 327: Maybe we could spare ourselves also those two methods and just do

```js
this.selected = [];
this.isSelecting = !this.isSelecting;
```

panel/src/components/Sections/ModelsSection.vue
bastianallgeier: This is based on feedback from one of the partner office hours. When you are able to switch two sections into select mode, it seems to the user that they can also select multiple items from multiple sections and they will all be deleted. It's breaking the expectations when only models from one section are being deleted afterwards. That's why I think it's clearer to simply switch off select mode whenever you enter another one.

panel/src/components/Sections/ModelsSection.vue
bastianallgeier: Since we already have the stopSelecting method, I find the code above more readable.

bastianallgeier: This really depends on your comment and the decision for the created method. When we still need the stopSelecting method there, we should also keep using it in the toggle method.

panel/src/components/Sections/ModelsSection.vue
distantnative: Ok, I didn't remember that feedback and feels odd to me (based on behaviour of other system), but ok if that creates problem with a big part of the audience. But then I think this might be rather a global state in our Panel module? Something just feels very duck-tape about this global event listener. Maybe also that it's just called "selecting" which could mean anything in a number of different contexts.

distantnative: How about we at least make it more specific, e.g. calling the global event `section.selecting`?

afbora:
> I don't think we can check for all subpages as well in addition to the permissions. That could lead to a big bottleneck. We could either add a checkbox to force delete all pages with subpages or we keep ignoring them here as it already is.

@bastianallgeier I mean, we can easily solve it by adding `'hasChildren' => $page->hasChildren(),` data to the pages section. Is this exactly the bottleneck you're talking about? Are the child pages being processed when `->hasChildren()` is called?

If this will cause a performance problem, the user can solve this by adding the `delete: false` option to pages that have children.

panel/src/components/Sections/ModelsSection.vue
bastianallgeier: We do something similar for dropdowns, so I don't think it's that much of a hack to be honest. But I agree that a more specific event name is good here. I just changed that.

distantnative (approved):

PR #6844, opened 2024-12-07
Make use of CSS `:has()`

distantnative:
### Reasoning
We already have started again to use `:has()` in many places. It might be time to bring all these back as well. I couldn't notice any performance issues in Firefox this time.

### Breaking changes
- `<k-item>`: Removed `data-only-option` attributes
- `<k-dialog>`: Removed `data-has-footer` attribute
- `<k-toggles-input>`: Removed `data-disabled` attribute
- `<k-writer-input>`: Removed `data-toolbar-inline` attribute
- `<k-bubble>`: Removed `data-has-text` attribute
- `<k-header>`: Removed `data-has-buttons` attribute
- `<k-tag>`: Removed `data-has-image` and `data-has-toggle` attributes
- `<k-tree>`: Removed `data-has-subtree` attribute

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Collection/Item.vue
+++ b/panel/src/components/Collection/Item.vue
@@ -2,6 +2,7 @@
 	<div
 		v-bind="data"
 		:class="['k-item', `k-${layout}-item`, $attrs.class]"
+		:data-has-image="hasFigure"
 		:data-layout="layout"
 		:data-theme="theme"
 		:style="$attrs.style"
@@ -210,7 +211,7 @@ export default {
 	align-items: center;
 	grid-template-columns: 1fr auto;
 }
-.k-item[data-layout="list"]:has(.k-item-image) {
+.k-item[data-layout="list"][data-has-image="true"] {
 	grid-template-columns: var(--item-height) 1fr auto;
 }
 .k-item[data-layout="list"] .k-frame {
@@ -269,7 +270,7 @@ export default {
 	grid-template-columns: 1fr;
 	grid-template-rows: 1fr var(--height-md);
 }
-.k-item[data-layout="cardlets"]:has(.k-item-image) {
+.k-item[data-layout="cardlets"][data-has-image="true"] {
 	grid-template-areas:
 		"image content"
 		"image options";
--- a/panel/src/components/Collection/Item.vue
+++ b/panel/src/components/Collection/Item.vue
@@ -2,7 +2,6 @@
 	<div
 		v-bind="data"
 		:class="['k-item', `k-${layout}-item`, $attrs.class]"
-		:data-has-image="hasFigure"
 		:data-layout="layout"
 		:data-theme="theme"
 		:style="$attrs.style"
@@ -39,7 +38,6 @@
 		<div
 			v-if="buttons?.length || options || $slots.options"
 			class="k-item-options"
-			:data-only-option="!buttons?.length || (!options && !$slots.options)"
 		>
 			<!-- Buttons -->
 			<k-button
@@ -156,12 +154,6 @@ export default {
 .k-item:has(a:focus) {
 	outline: 2px solid var(--color-focus);
 }
-/** TODO: remove when firefox supports :has() */
-@supports not selector(:has(*)) {
-	.k-item:focus-within {
-		outline: 2px solid var(--color-focus);
-	}
-}
 
 .k-item .k-icon-frame {
 	--back: var(--color-gray-300);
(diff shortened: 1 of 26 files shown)


bastianallgeier:
I wonder if we should switch the css rules, but still keep the data attributes a little longer. They migth still be useful as hooks for custom styles.

distantnative:
I'd love to steer people away from these data attributes as soon as possible.

nilshoerrmann:
Removing the data attributes will be huge breaking change for many projects. Make sure to communicate this loudly if you proceed in that direction. It's a good idea in general :+1: but it will be disruptive for existing projects 🚨

distantnative:
We of course need to communicate it properly. That's why they're all listed as breaking changes in the initial post. 

I am still wondering if these specific data attributes are really used in "many projects". What are the use cases in projects for these?

panel/src/components/Collection/Item.vue
bastianallgeier, line 1: I would keep the data-has-image attribute here

panel/src/components/Forms/Field/BlocksField.vue
bastianallgeier, line 1: I would keep the margin-top solution here.

panel/src/components/Forms/Field/DateField.vue
bastianallgeier, line 1: I think the data-has-time attribute adds some additional semantic value here. I would keep it.

panel/src/components/Forms/Field/LayoutField.vue
bastianallgeier, line 1: I would stick with the margin-top solution

panel/src/components/Forms/Field/StructureField.vue
bastianallgeier, line 1: I would stick with the margin-top solution

panel/src/components/Lab/PlaygroundView.vue
```diff
@@ -106,7 +103,7 @@ export default {
 </script>
 
 <style>
-.k-lab-playground-view[data-has-tabs="true"] .k-lab-playground-header {
+.k-lab-playground-view:has(.k-tabs) .k-lab-playground-header {
```
bastianallgeier: I would say that this is more risky because it could get triggered by an unexpected .k-tabs component somewhere (i.e. in a nested component). I would keep the data attribute for more control.

panel/src/components/Layout/Header.vue
bastianallgeier, line 1: I'm torn here. I like to get rid of the slot logic in the data-has-buttons attribute, but I also feel like it adds some semantic value and the :has selector is quite complex.

panel/src/components/Navigation/Button.vue
bastianallgeier, line 1: I guess that this is the most critical part to remove the attributes, because buttons are used so often. Not sure if it is worth it to be honest. At least at this point. Maybe let's wait one more major release?

panel/src/components/Views/Files/FileView.vue
bastianallgeier, line 1: This is another place where I feel like the data attribute is giving more control and selector "collision safety"

panel/src/components/Views/Pages/PageView.vue
bastianallgeier, line 1: See comment above

panel/src/components/Views/Pages/SiteView.vue
bastianallgeier, line 1: See comment above.

panel/src/components/Views/Users/UserView.vue
bastianallgeier, line 1: See comment above.

bastianallgeier (commented):
I think it's good to clean up, but there are parts that are easier to read with the data attributes (in my opinion) or provide a bit more selector safety. I've tried to go through all the different changes and comment accordingly. I hope it helps to get this done.

panel/src/components/Lab/PlaygroundView.vue
distantnative: How about `:has(> .k-tabs)`?

panel/src/components/Layout/Header.vue
distantnative: @bastianallgeier You really think it adds more value that `:has(.k-header-buttons)` doesn't solve in the same way? And while I understood other comments, I don't feel that selector is very complex now.

bastianallgeier (approved):

PR #6846, opened 2024-12-07
Writer toolbar: hide when no buttons

### LINKED ISSUE #6804: Writer Fields Non-inline toolbar should not show when no marks or nodes are available
## Description

**Expected behavior**  

**Screenshots**  

## Your setup
**Kirby Version**  

**Console output**  

 - Device: 
 - OS: 
 - Browser: 
 - Version: 

**Additional context**

distantnative:
### Fixes
- Writer field: inline toolbar is hidden when no nodes and marks are available #6804

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Toolbar/Toolbar.vue
+++ b/panel/src/components/Forms/Toolbar/Toolbar.vue
@@ -1,5 +1,5 @@
 <template>
-	<nav class="k-toolbar" :data-theme="theme">
+	<nav v-if="buttons.length" class="k-toolbar" :data-theme="theme">
 		<template v-for="(button, index) in buttons">
 			<hr v-if="button === '|'" :key="index" />
 
--- a/panel/src/components/Forms/Writer/Toolbar.vue
+++ b/panel/src/components/Forms/Writer/Toolbar.vue
@@ -338,8 +338,7 @@ export default {
 </script>
 
 <style>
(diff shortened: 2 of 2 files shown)


afbora (approved):
Shouldn't the target branch be `develop-minor` for 4.6.0?

distantnative:
@afbora not necessarily. We put everything that could be released in a patch release on `develop-patch` - in that case that we would want to release a 4.5.1, we could do that right away with the necessary changes already on `develop-patch`. Since we continuously merge `develop-patch` into `develop-minor`, if we decide against any patch release, they're automatically included in the next minor release.

bastianallgeier:
I'm totally fine to include it in the patch release, but I think we should remove the :has change in that case, right?

distantnative:
You're right, we only started adding `:has()` back with v5.... mm but without that change it's still broken/needs additional changes in other components.

distantnative:
@bastianallgeier I'd suggest then to move it to v5. I think it's rather complicated to fix this without `:has()` as the style is applied to `.k-writer` not the toolbar itself. I though about introducing another data attribute, but then we would have to duplicate the whole `buttons` logic from `k-writer-toolbar` in `k-writer` to make it consistent.

bastianallgeier:
@distantnative can you fix the conflicts?

distantnative:
@bastianallgeier done

PR #6855, opened 2024-12-09
Fix alt attribute for `FileVersion` instances

### LINKED ISSUE #6852: Alt attribute lost after resize/crop
## Description
Returning an image file also handles populating the `alt` attribute with the `alt` field. After resizing or cropping the image, this alt text is however lost, returning an empty `alt` attribute.

### Without resizing
```php
echo $page->image();
```
```html
<img src="image.jpg" alt="My alt text">
```

### Same image after resizing
```php
echo $page->image()->resize(600);
```
```html
<img src="image.jpg" alt="">
```

By simply returning an image, Kirby calls `html()`, as defined in its `__toString()` function. This method can be found here: https://github.com/getkirby/kirby/blob/94cc37ee7c3004ebb4950a53f14e1329ed4d28d3/src/Image/Image.php#L116-L134

**Expected behavior**  
The alt tag should be returned after resizing/cropping the image.

## To reproduce

1. Make sure your page has an image
2. Return something like `$page->image()` and see the alt text
2. Add `->resize(600)` or `->crop(600, 400)` to it and see the alt text disappear

## Your setup
**Kirby Version**  
Tested in 4.4.1 and 4.5.0

afbora:
## Description

If a file is cropped/resized, the FileVersion instance is used. We can get the original model from `FileVersion::original()` method.

### Fixes
- #6852

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/src/Image/Image.php
+++ b/src/Image/Image.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Image;
 
+use Kirby\Cms\FileVersion;
 use Kirby\Content\Content;
 use Kirby\Exception\LogicException;
 use Kirby\Filesystem\File;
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
Nice!

PR #6856, opened 2024-12-09
Fix for homepage preview URLs on proper domains

lukasbestle:
### Summary of changes

- Clean up today's preview token quickfix
- Add unit tests to prevent this bug from reappearing

### Reasoning

Sites on real domains have their homepage on `https://example.com` and not `https://example.com/`. The unit tests did not catch this because they all used `/` as site index URL, which naturally does not have this trailing slash problem.

### Additional context

Today's quickfix by Basti passes all those tests as well. 🎉

So I think releasing a beta 2 with this PR isn't urgent at all. I decided to change the code mainly for readability and future robustness (Basti's fix only works because the `Uri` class normalizes the trailing slash as well, but we can't rely on that here).

### Refactoring (since beta 1, not for final release notes)

- Update preview token logic to reliably fix tokens for the homepage

## Docs

*None*

## Ready?

### For review team

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -427,9 +427,13 @@ class Version
 	{
 		$localPrefix = $this->model->kirby()->url('base') . '/';
 
-		// Todo: this is only a quick fix to get home page previews working again,
-		// we need to double-check if this is still correct
-		if (Str::startsWith($url, $localPrefix) === false && $url . '/' !== $localPrefix) {
+		// normalize homepage URLs to have a trailing slash
+		// to make the following logic work with those as well
+		if ($url . '/' === $localPrefix) {
+			$url .= '/';
+		}
+
+		if (Str::startsWith($url, $localPrefix) === false) {
 			return null;
 		}
 
--- a/tests/Content/VersionTest.php
+++ b/tests/Content/VersionTest.php
@@ -774,12 +774,33 @@ class VersionTest extends TestCase
 		$this->assertSame($content, Data::read($fileENLatest));
 	}
 
+	public static function previewTokenIndexUrlProvider()
+	{
+		return [
+			['/'],
+			['/subfolder'],
+			['/subfolder/'],
+			['https://example.com'],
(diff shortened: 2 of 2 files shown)


src/Content/Version.php
```diff
@@ -427,9 +427,13 @@ protected function previewTokenFromUrl(string $url): string|null
...
-		if (Str::startsWith($url, $localPrefix) === false && $url . '/' !== $localPrefix) {
+		// normalize homepage URLs to have a trailing slash
+		// to make the following logic work with those as well
+		if ($url . '/' === $localPrefix) {
+			$url .= '/';
+		}
+
+		if (Str::startsWith($url, $localPrefix) === false) {
 			return null;
 		}
```
afbora, line 438: What do you think about following usage? I think it's more simply and readable.

````php
$baseUrl     = $this->model->kirby()->url('base');
$localPrefix = $baseUrl . '/';

if (Str::startsWith($url, $baseUrl) === false) {
    return null;
}
````

Rather than adding `/` to two different variables and checking them, it seemed more logical to me to check their raw states.

lukasbestle: I had that first, but it would break for subfolder installations. E.g. the current site is at `https://example.com/subfolder` and you have a preview link set to `https://example.com/subfolder-with-some-extra`. This URL starts with the index URL but points to a different folder and is not part of the same Kirby site.

afbora: I guess I didn't explain it. You check by adding `/` to the `base` url and `$url`. I say let's check without adding `/` to both. 

What I mean is why don't we check without adding `/` at the end? Does the `/` at the end have anything to do with the different subfolder setup?

lukasbestle: I think so, because the `Str::startsWith()` check will return `true` for both even though one is considered external. With the trailing slash we know for sure that the URL belongs to the same site.

afbora (approved):

PR #6874, opened 2024-12-17
Access permissions 1: Refactor access to roles

lukasbestle:
### Summary of changes

`$app->role()` method is mainly a shortcut to `$app->user()?->role()` for us, but can also be used to directly access a specific role object.

### Reasoning

Following PRs would add significantly more uses of `$app->user()?->role()` that can be shortened and DRYed by this.

### Features

- New `$app->role()` method that works like `$app->user()` to return a specific role or the role of the current user.

### Housekeeping

- Move `$app->roles()` method to the `AppUsers` trait to increase the overview

### Breaking changes

None

## Docs

None (automatically documented via reference)

## Ready?

### For review team


bastianallgeier (approved):
Good cleanup. Do we worry about the coverage? I would merge it otherwise.

lukasbestle:
The entire view action for the language view is not covered at the moment, this includes the only uncovered line in this changeset. I think at this point it's more important to continue with the permission additions than coverage.

PR #6879, opened 2024-12-23
Access permissions 3: Smaller refactorings

lukasbestle:
### Summary of changes

Smaller refactorings

### Reasoning

Consistency and cleanup

### Housekeeping

- Clean up permission-related code for consistency

### Breaking changes

None

## Docs

None

## Ready?

### For review team

### DIFF
--- a/config/sections/fields.php
+++ b/config/sections/fields.php
@@ -13,7 +13,7 @@ return [
 	'computed' => [
 		'form' => function () {
 			$fields   = $this->fields;
-			$disabled = $this->model->permissions()->update() === false;
+			$disabled = $this->model->permissions()->cannot('update');
 			$lang     = $this->model->kirby()->languageCode();
 			$content  = $this->model->content($lang)->toArray();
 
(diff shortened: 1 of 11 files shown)


config/areas/site/buttons.php
```diff
@@ -17,7 +17,7 @@
 		);
 	},
 	'page.preview' => function (Page $page) {
-		if ($page->permissions()->can('preview') === true) {
+		if ($page->permissions()->preview() === true) {
```
distantnative: When working on the user roles, I ran into some problems where using the magic caller methods was inferior to directly calling the `->can()` method. And tried to start moving away from using the magic methods. I can't remember right now what was the exact problem (maybe related to testing) - I'll try to remember.

bastianallgeier: I think I would also go for the can method here. We should get away from the magic callers wherever possible.

lukasbestle: You are right. Flipped the commit around to rather change the existing uses of the magic method to the proper `can()` method. Now everything is consistent again.

bastianallgeier (approved):

PR #6880, opened 2024-12-23
Access permissions 4: Fix user caching

lukasbestle:
## 🚨 Merge first

- #6879 

### Summary of changes

- Prepare static permission caching for PR 5
- No longer cache the user object in each `ModelPermissions` instance

### Reasoning

Prevents the permission checks from returning stale results if the user changed during the request and the permissions object is kept in memory (currently not the case in the core, but this refactoring is a security precaution against future regressions).

### Fixes

- Permissions are now always correctly determined if the active user changes during the request

### Breaking changes

- Classes extending `ModelPermissions` need to set their category with the `const CATEGORY` or `static function category()` instead of `protected string $category` and use the static `user()` method instead of `$user` and `$permissions` properties

## Docs

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/FilePermissions.php
+++ b/src/Cms/FilePermissions.php
@@ -13,7 +13,7 @@ namespace Kirby\Cms;
  */
 class FilePermissions extends ModelPermissions
 {
-	protected string $category = 'files';
+	protected const CATEGORY = 'files';
 
 	protected function canChangeTemplate(): bool
 	{
--- a/src/Cms/LanguagePermissions.php
+++ b/src/Cms/LanguagePermissions.php
@@ -13,7 +13,7 @@ namespace Kirby\Cms;
  */
 class LanguagePermissions extends ModelPermissions
 {
-	protected string $category = 'languages';
+	protected const CATEGORY = 'languages';
 
 	protected function canDelete(): bool
 	{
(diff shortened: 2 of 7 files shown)


distantnative:
@lukasbestle Do you have an idea how much of a performance impact this has? Since if we don't cache the `ModelPermission` objects, each time they now have to be created the blueprint is also read (which has been a performance bottleneck for us).

lukasbestle:
There are two components to this:

- At the moment we don't cache the `ModelPermissions` objects at all. E.g. `$page->permissions()` creates an instance each time. Which means that the user object is retrieved each time a permission is checked (before: in the constructor of `ModelPermissions`; after: when actually checking the permission). So in practice this PR alone does not create any difference. Only if we ever reuse an existing `ModelPermissions` object for additional checks, this becomes a difference. But there I would say security > performance. Viewed from the opposite perspective: With this PR, it becomes practical for the first time to cache the `ModelPermissions` object without security impact, thus allowing us to read the blueprint only once (per model).
- We have another layer of caching for `access`/`list` permissions, which are those that are commonly checked more than once per request. All the other permissions are only checked for a particular action, of which not many happen per request.

bastianallgeier (approved):

PR #6881, opened 2024-12-23
Access permissions 5: DRY permission caching

lukasbestle:
## 🚨 Merge first

- #6879
- #6880 

### Summary of changes

Move caching of `access` and `list` permissions to `ModelPermissions`

### Reasoning

DRY code, especially now that we would add five more of these methods to the `User`, `Site` and `Language` classes

### Additional context

We could cache all other permissions as well, but I've wrestled with unit tests for too long before Christmas. The write permissions would also benefit less from the caching as there is usually only one write operation per request (in contrast to potentially hundreds of list/access permission checks).

Another advantage of the separate `canFromCache()` method is that it makes explicit which checks come from cache and which are dynamic. If we ever use `canFromCache()` in a place that needs dynamic checking, we will immediately be able to tell because of the `LogicException` that is thrown then.

The new `UserPermissions::cacheKey()` is currently uncovered as the `User` class does not have `access`/`list` permissions at the moment. This will be fixed in the PR 7.

### Refactoring

- Globally cache `access` and `list` permissions per permission category, model type and user role to reduce code duplication

### Breaking changes

None

## Docs

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -318,12 +318,7 @@ class File extends ModelWithContent
 			return false;
 		}
 
-		static $accessible   = [];
-		$role                = $this->kirby()->role()?->id() ?? '__none__';
-		$template            = $this->template() ?? '__none__';
-		$accessible[$role] ??= [];
-
-		return $accessible[$role][$template] ??= $this->permissions()->can('access');
+		return FilePermissions::canFromCache($this, 'access');
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


lukasbestle:
~I think this PR really does have a performance impact, see https://github.com/getkirby/kirby/pull/6880#issuecomment-2559657289. With these changes, we need to create a `ModelPermissions` instance for each model and not just for each combination of template and user role.~

~I wonder if there is a better way to DRY the logic without this performance regression. Or otherwise we should probably discard this and add the same caching logic to the new `isAccessible()/isListable()` methods for site/user/language in upcoming PR 7.~

The performance impact should now be fixed with a different approach: I moved the static caching logic to a static method inside `ModelPermissions`. This effectively does the same as before, just in one place to avoid the code duplication.

bastianallgeier (approved):

PR #6894, opened 2025-01-11
Stucture field: fix preview input when paginated

### LINKED ISSUE #6892: [v4] Paginated structure display bug
## Description
On a paginated structure with having a column with a toggle field, subsequent pages are deleted when changing the toggle switch state.

**Expected behavior**  
Changing the state of a toggle of an entry in the structure should just change the state. No entries should be deletd.

## To reproduce

1. Create a structure field in a page blueprint
2. Add a toggle field to the blueprint fields definition and columns definition
3. Set the limit setting to create a paginated structure entries list
4. In the panel switch the toggle state of an entry on page 1 for example
5. Save and refresh the page: subsequent pages are gone

## Your setup
**Kirby 4.5**  

 - Device: MacBook Pro (2016)
 - OS: macOS, iOS 15.2
 - Browser: Firefox
 - Version: 133

distantnative:
### Summary of changes
- When a structure field is paginated, `k-table` will only receive the slice of values for the current page.
- Consequently, it will only emit those rows again when a row is updated via a field preview.
- Thus, we cannot simply save what we receive from he table input event, but need to splice those rows into the full unpaginated array and save that one.

### Fixes
- Structure field: fixed losing rows when changes were made through a field preview in a paginated table

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Field/StructureField.vue
+++ b/panel/src/components/Forms/Field/StructureField.vue
@@ -64,7 +64,7 @@
 					:sortable="isSortable"
 					:data-invalid="isInvalid"
 					@cell="open($event.row, $event.columnIndex)"
-					@input="save"
+					@input="onTableInput"
 					@option="option"
 					@paginate="paginate"
 				/>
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):
TIL about toSpliced 🤯 Nice!

PR #6898, opened 2025-01-11
Range/Number field: fix `isEmpty()` usage

### LINKED ISSUE #6871: [v5] `number`/`range` field in page create dialog: `step` option is ignored
## Description
I have a simple range slider in my blueprint with a `step` value set to `0.5`

```
rating:
    label : Rating
    type  : range
    step  : 0.5
    min   : 0
    max   : 6
```

The field is the displayed inside the page creation modal.

Screenshot 2024-12-17 at 9 20 03 AM

The `step` value seems to be missing from the input inside the modal and as a result the range slider only goes up and down by 1

Screenshot 2024-12-17 at 9 20 11 AM

Everything works perfectly fine inside the single page, bug only happens inside the page creation modal.

**Expected behavior**  
The `step` value set in the blueprint should be applied to the input inside the modal.

## Your setup
**Kirby Version**  

**Console output**  

 - Device: Mac
 - OS: 15.1
 - Browser: Safari 18.1

**Additional context**  
Everything works fine using latest K4. Bug was present in latest Alpha and it's still there in the current Beta.

distantnative:
### Summary of changes
- After restructuring the form classes, `Form\Value::isEmpty()` (and by that `$field->isEmpty()` cannot be called anymore with a value passed
- Number, range and toggle field therefore must use `Value::isEmptyValue($value)` instead

## Ready?

### For review team

### DIFF
--- a/config/fields/number.php
+++ b/config/fields/number.php
@@ -34,7 +34,7 @@ return [
 	],
 	'methods' => [
 		'toNumber' => function ($value) {
-			if ($this->isEmpty($value) === true) {
+			if ($this->isEmptyValue($value) === true) {
 				return null;
 			}
 
--- a/config/fields/toggle.php
+++ b/config/fields/toggle.php
@@ -67,7 +67,7 @@ return [
 		'required' => function ($value) {
 			if (
 				$this->isRequired() &&
-				($value === false || $this->isEmpty($value))
+				($value === false || $this->isEmptyValue($value))
 			) {
 				throw new InvalidArgumentException(
 					message: I18n::translate('field.required')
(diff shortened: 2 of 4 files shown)


bastianallgeier:
I think we should still support the isEmpty option in an overwritten `Field::isEmptyValue` method. I would say that this is an oversight on my end.

distantnative:
@bastianallgeier wasn't sure if deliberate or oversight, so I wanted to at least mention it as breaking change.

bastianallgeier:
I've brought it back. @distantnative are you fine to add it to the PR like that? The rest looks good to me.

bastianallgeier (approved):

distantnative:
Yes, looks good

PR #6907, opened 2025-01-16
Fix link color in themed boxes

### LINKED ISSUE #6906: [v5] Link in info box invisible in dark mode
CleanShot 2025-01-16 at 09 41 00

CleanShot 2025-01-16 at 09 42 12

Kirby 5.0.0-beta.1

bastianallgeier:
### Fixes

- Fixed link color in dark and light mode in all themed boxes https://github.com/getkirby/kirby/issues/6906

### For review team

### DIFF
--- a/panel/lab/components/boxes/1_themes/index.vue
+++ b/panel/lab/components/boxes/1_themes/index.vue
@@ -24,13 +24,25 @@
 				/>
 			</k-grid>
 		</k-lab-example>
+		<k-lab-example label="Text, Icon & Link">
+			<k-grid style="--columns: 1; gap: 0.5rem">
+				<k-box
+					v-for="theme in themes"
+					:key="theme"
+					:html="true"
+					:theme="theme"
+					text='Text with <a href="/">URL</a>'
+					icon="box"
+				/>
+			</k-grid>
+		</k-lab-example>
 	</k-lab-examples>
 </template>
 
 <script>
 export default {
 	props: {
-		themes: Array,
-	},
+		themes: Array
+	}
 };
 </script>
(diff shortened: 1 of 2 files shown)


distantnative (commented):
The `text` theme used to have a nice dark blue link with lighter blue hover state. This is broken with this PR.

I tried but indeed getting a good color tint is tricky without adding a whole range of more theme properties. Probably not worth it. But hover state might be nice.

Maybe we could use

```css
--link-color-hover: var(--theme-color-text-dimmed);
```

bastianallgeier:
@distantnative I've tried it and it does not look good in most themes to be honest. I would leave it as it is.

distantnative:
@bastianallgeier I get that this isn't maybe the best solution, but I feel leaving it as it is without any color variation from the normal text, without any hover, also isn't right.

bastianallgeier:
Could we maybe work with the brightness filter for the hover effect instead? filter: brightness(110%) or 90% depending on the theme? 

On the other hand, I feel like this is one of the really minor issues and I would rather vote for the simple but less good version instead of spending too much time on it to be honest.

distantnative:
I think I would go with 

```
--link-color-hover: var(--theme-color-850);
```

That seems to work well, only the code theme looks a bit drastic with it, but I'd say there is few code with links in it.

bastianallgeier:
But that would only be useful for one mode and depending on the box theme it might also be wrong. I.e. when the box is black.

distantnative (approved):
Ok I give in :D

bastianallgeier:
I really don't want you to give in. I would prefer a better solution myself. But looking at this, it's quite complex to solve the more you think about it …

Screenshot 2025-01-21 at 21 43 23

PR #6920, opened 2025-01-18
Plugin update status: passive theme for unkown

distantnative:
### Summary of changes
- System view: update status dropdown uses passive theme when update status is unknown 
Screenshot 2025-01-18 at 11 17 49

### Reasoning
The `notice` seem always has implied some urgency to act for the user. However, the user/developer cannot do much about this actually. Only the plugin developer can. I feel like the passive theme is more fitting for this unknown status

## Ready?


afbora (approved):
Your reason is totally make sense 👍

bastianallgeier (approved):

PR #6922, opened 2025-01-18
Page duplicate dialog: don't require toggles

### LINKED ISSUE #6921: [v5] Duplicate action is forcing to copy files
> Glad to hear this will be part of Kirby 5. While trying this behaviour using beta-1, I had a different issue where leaving the toggle set to false would not let me duplicate the page at all:
> 
> image 

 _Originally posted by @linusrogge in [#6885](https://github.com/getkirby/kirby/issues/6885#issuecomment-2599682949)_

distantnative:
Fixes #6921

### Summary of changes
- Page duplicate dialog: don't require toggles

### Reasoning
- With `required` toggles, in v5 one cannot submit the form without activating the toggle, so false is no option.
- Apparently, in v4 toggles could be on or off even when required.
- I was wondering if we should change the general behavior for toggles (I assume also checkbox) to be the same as v4, but then `required: true` actually has no effect. So I do like that one can actually require a toggle/checkbox to be set. We just have to remove the required option from use cases like here where it's not right.

## Changelog

Not from this PR, but so far missed...

### Breaking changes
- `required: true` on checkbox and toggle fields now enforces that these fields need to be checked/toggles (active state)

## Ready?

### For review team

### DIFF
--- a/config/areas/site/dialogs.php
+++ b/config/areas/site/dialogs.php
@@ -410,19 +410,17 @@ return [
 
 			if ($hasFiles === true) {
 				$fields['files'] = [
-					'label'    => I18n::translate('page.duplicate.files'),
-					'type'     => 'toggle',
-					'required' => true,
-					'width'    => $toggleWidth
+					'label' => I18n::translate('page.duplicate.files'),
+					'type'  => 'toggle',
+					'width' => $toggleWidth
 				];
 			}
 
 			if ($hasChildren === true) {
 				$fields['children'] = [
-					'label'    => I18n::translate('page.duplicate.pages'),
-					'type'     => 'toggle',
-					'required' => true,
-					'width'    => $toggleWidth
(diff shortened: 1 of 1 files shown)


afbora:
IMHO v5 behavior is correct. Otherwise required attribute would useless. But this is also a breaking change. If we continue with that we need to add to breaking changes in release notes.

Another way is deprecate this usage in v5 and remove it in v6.

Also PR is looks good to me and work great! 👍

bastianallgeier:
I agree that the new behavior makes a lot more sense. There's an indeterminate state for checkboxes: https://css-tricks.com/indeterminate-checkboxes/ but we don't support that yet at all. I think this would be the only way to have toggles and checkboxes that are not set yet. But with our current state, the required attribute does not make any sense in v4

bastianallgeier (approved):

PR #6931, opened 2025-01-22
Clean up ZIP exports

lukasbestle:
### Summary of changes

Updated which files are needed in the ZIP export

### Reasoning

Cleanup

### Additional context

The `package-lock.json` was added in c840949299f7416cb67e374cdacd1a935c228ab8  and looks like an accidental addition to me.

### Enhancements

- Update the list of export ignores to keep unneeded files out of the Kirby ZIP downloads

## Docs

*None*

### For review team

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -15,9 +15,10 @@ tests/ export-ignore
 
 # panel
 panel/.env.example export-ignore
-panel/.eslintrc.js export-ignore
+panel/.prettierignore export-ignore
 panel/.prettierrc.json export-ignore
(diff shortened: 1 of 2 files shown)


distantnative (approved):

distantnative:
Guess I ran `npm i` in the wrong directory and didn't notice. Should we add it in the gitignore?

lukasbestle:
Won't harm I guess. But also not critical IMO. Normally we should catch something like this in review, as it can happen with other file types and locations as well.

PR #6935, opened 2025-01-23
Support editing arrays for language variables

### LINKED ISSUE #6930: Language view: support for i18n string as arrays (those using `translateCount`
## Description

Using `I18n::translateCount()` or its alias `tc()` requires the use of arrays in language variables:

```php
'translations' => [
    'en' => [
        'errors'  => ['There are no errors', 'There is { count } error.', 'There are { count } errors.'],
    ],
    // …
],
```

The panel languages view correctly renders the array:

Image

…but does not allow edition of the entries:
```
Kirby\Cms\LanguageVariable::value(): Return value must be of type ?string, array returned
```

## Your setup

**Kirby 4.5.0**

afbora:
## Description
develop test_panel_languages_en

### Summary of changes
- Now language variables also support arrays (string, array and null)

### Reasoning
We talked with Nico about how we should support array language variables.

I tried structure field but gave up because it was opened with drawer (not easy to edit each entry), it had deletion issues and it did not work fully compatible with dialog.

Nico's suggestion was to show as many inputs as there are variables in the array. I implemented this. I did not implement this in the creation part because it is not clear how many variables the user will enter, so I did not do this feature for now.

My alternative suggestion is a new field suggestion for such needs for K5 or K6; `entries`. It is similar to structure field but much simpler. You can think of it as a text field field that can be added more than once. If we have entries field in K5 or K6, we can use `entries` field in the language variables part.

I did not want to add unit tests without getting a full response from you. I will add it last.

### Fixes
- #6930

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/areas/languages/dialogs.php
+++ b/config/areas/languages/dialogs.php
@@ -240,24 +240,55 @@ return [
 
 			$fields = $translationDialogFields;
 			$fields['key']['disabled'] = true;
-			$fields['value']['autofocus'] = true;
+
+			$value = ['key' => $variable->key()];
+
+			// if language variable is array,
+			// set text fields as per array size
+			// otherwise it will show only one textarea
+			if (is_array($variable->value()) === true) {
+				unset($fields['value']);
+
+				foreach ($variable->value() as $index => $val) {
+					$fields['value_' . $index] = [
+						'autofocus' => $index === 0,
+						'counter'   => false,
+						'label'     => I18n::translate('language.variable.value') . ' ' . ($index + 1),
+						'type'      => 'text',
+					];
+					$value['value_' . $index]  = $val;
+				}
+			} else {
+				$fields['value']['autofocus'] = true;
+				$value['value'] = $variable->value();
+			}
 
 			return [
 				'component' => 'k-form-dialog',
 				'props' => [
 					'fields' => $fields,
 					'size'   => 'large',
-					'value'  => [
-						'key'   => $variable->key(),
-						'value' => $variable->value()
-					]
+					'value'  => $value
 				],
 			];
 		},
 		'submit' => function (string $languageCode, string $translationKey) {
-			Find::language($languageCode)->variable($translationKey, true)->update(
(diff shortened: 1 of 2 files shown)


bastianallgeier:
I was thinking about an "entries" field for a while now too and I agree that this would be a great chance to use it here. 

But since we are working with a very defined set of options here, I think we could also add a toggle to switch between single value and pluralized array values. 

The fields could then also get better labels. I.e.

- "Zero"  
- "One"
- "Many"

distantnative:
@bastianallgeier The tricky part is that our setup actually allows more then 3 options. A i18n translation might be defined for e.g. cases of 0, 1, 2, and more. Which needs 4 fields then. That's why Ahmet creates them dynamically from the number of array entries.

bastianallgeier:
Ok, but that makes sense. Then I would really just improve the labels

afbora:
How would you like to improve labels?

bastianallgeier:
Ah, I've just seen in the translateCount method that it's not super easy. Can we maybe explain it somehow in an info box above and then do labels like this: 

Count: 0
Count: 1
Count: 2

?

distantnative:
Or we do dynamic help texts as well?

This string will be used if `count` is 0.
This string will be used if `count` is 1.
This string will be used if `count` is 2 or more.

afbora:
So what about labels? 
Also what if 1, 2 or 4 entries have? Should be like that?

This string will be used if count is 0.
This string will be used if count is 1 or more.

This string will be used if count is 0.
This string will be used if count is 1.
This string will be used if count is 2 or more.

This string will be used if count is 0.
This string will be used if count is 1.
This string will be used if count is 2.
This string will be used if count is 3 or more.

distantnative:
Yep, I think that would be the logic.

For labels it could be similar:

Translation for 0 count
Translation for 1 count
Translation for 2+ count

afbora:
Closing favor of #6963

PR #6941, opened 2025-01-23
PDF file preview

distantnative:
## Description

I think this could be a neat addition for the core. With fallback in case inline PDF viewing isn't supported.

Or if not core, we could release it as small plugin.

### Feature
- PDF file preview

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Views/Files/DefaultFilePreview.vue
+++ b/panel/src/components/Views/Files/DefaultFilePreview.vue
@@ -29,7 +29,7 @@ export default {
 </script>
 
 <style>
-.k-default-file-preview .k-file-preview-frame > .k-icon {
+.k-default-file-preview .k-file-preview-frame .k-item-icon {
 	--icon-size: 3rem;
 }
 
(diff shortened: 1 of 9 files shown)


bastianallgeier:
This is great. The fallback is super elegant. I would only be worried if loading PDFs could lead to any new JS security issues. I've read somewhere that PDFs have optional JS support, which is pretty crazy. But I'm not aware if this could be exploited or not. Maybe @lukasbestle has some additional thoughts?

distantnative:
mdn has a section on security concerns with frames, but uses object as example for PDFs without any further security mentions: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/General_embedding_technologies#the_embed_and_object_elements

lukasbestle:
PDFs do have JS support, but I would assume that when embedding a PDF in `<object>`, browsers sandbox the embed similar to when using the `` element with an SVG. So the JS code won't have access to the running Panel instance. However I don't have experience with the `<object>` element in a security context.

lukasbestle:
Actually not a very good comparison as SVGs in `` just don't execute JS at all, but what I wanted to say is that embeds with `<object>` should not be able to interact with the page.

bastianallgeier (approved):
I've just added a subtle border to separate the preview a bit better. But I'm super happy with the rest. It's nice to have this in the core.

PR #6948, opened 2025-01-24
Dragtext: always use UUID

### LINKED ISSUE #6943: Kirbytag uses filename instead of UUID when file is dragged in a texture field
https://github.com/user-attachments/assets/ac8f4f27-6059-4fc0-9854-656d1422ddaa

Kirbytag uses filename instead of UUID when file is dragged in a text field

**Expected behavior**  
I would like to use the UUID instead of the filename to reference a file in a kirbytag for consistency, and so filename changes don’t mess up my kirbytags.

distantnative:
### Summary of changes
- Always use UUID for file drag text (when enabled); no more check whether the file is part of the same model

### Reasoning
Even if the file is in the same directory, if it gets moved around or filename changed, the reference is lost without use of UUID. It makes sense to always use the UUID, if enabled.

### Breaking changes
- File drag texts in the Panel always use the file UUID, if UUIDs are not disabled

## Ready?

  - fully covered for what changed
### For review team


bastianallgeier:
I did not have time to test it manually. I think the solution looks good and it's indeed so much better when we can actually use UUIDs here. But it also means that we are more dependent on the UUIDs. Adding the URLs as fallback will make the fallback worse than what we had so far. If you have an absolute URL setup, you will have absolute URLs in your content.

distantnative:
So maybe then
- UUIDs active: always use UUID/permalink
- UUIDs disabled: use relative URL where possible, otherwise absolute URL
?

bastianallgeier:
I think that would indeed be the best cascade, yes.

bastianallgeier (approved):

PR #6950, opened 2025-01-24
Support shared `site.contentType.php` controller

distantnative:
### Summary of changes
- Allows content representation-specific site controllers, e.g. `site.json.php`
- Either `site.php` or the more specific representation site controller get merged into any other controller, but never both

### Reasoning
Allows to globally provide data to all controllers/templates that only a specific content type needs.

### Additional context
- Please review the unit tests closely if each configuration ends in the results you would expect as well.

### Enhancement
- Support for content representation specific site controllers, e.g. `site.rss.php`

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -464,27 +464,24 @@ class App
 		$data = [];
 
 		// always use the site controller as defaults, if available
-		if ($controller = $this->controllerLookup('site')) {
-			$data = (array)$controller->call($this, $arguments);
+		$site   = $this->controllerLookup('site', $contentType);
+		$site ??= $this->controllerLookup('site');
+
+		if ($site !== null) {
+			$data = (array)$site->call($this, $arguments);
 		}
 
 		// try to find a specific representation controller
-		if ($controller = $this->controllerLookup($name, $contentType)) {
+		$controller   = $this->controllerLookup($name, $contentType);
+		// no luck for a specific representation controller?
+		// let's try the html controller instead
+		$controller ??= $this->controllerLookup($name);
+
+		if ($controller !== null) {
 			return [
 				...$data,
 				...(array)$controller->call($this, $arguments)
 			];
-		}
-
-		if ($contentType !== 'html') {
-			// no luck for a specific representation controller?
-			// let's try the html controller instead
-			if ($controller = $this->controllerLookup($name)) {
-				return [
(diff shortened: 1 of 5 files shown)


src/Cms/App.php
```diff
@@ -464,29 +464,26 @@ public function controller(
 		$data = [];
 
 		// always use the site controller as defaults, if available
-		if ($controller = $this->controllerLookup('site')) {
-			$data = (array)$controller->call($this, $arguments);
+		$site   = $this->controllerLookup('site', $contentType);
+		$site ??= $this->controllerLookup('site');
```
distantnative, line 468: `Discuss:` Is it ok that we use either or? Or should both site controllers be used, if available?

tests/Cms/App/AppTest.php
```diff
@@ -1369,22 +1372,21 @@ public function testControllerRepresentation()
 		]);
 
 		Page::factory([
-			'slug' => 'test',
+			'slug'     => 'test',
 			'template' => 'another'
 		]);
 
-		ob_start();
-		$app->controller('another.json', [], 'json');
```
distantnative, line 1377: `Question:` Was it ever right to have this test calling with `another.json` as name and `json` as contentType instead of `another` and `json`?

tests/Cms/App/fixtures/controllers/another.json.php
```diff
@@ -1,3 +1,5 @@
 <?php
 
-echo json_encode(['foo' => 'bar']);
```
distantnative, line 3: `Question:` Why would the controller echo the JSON? Isn't that the job of the template in a content representation?

src/Cms/App.php
bastianallgeier: We used either/or so far and I think we should stick with that.

tests/Cms/App/AppTest.php
bastianallgeier: No, that looks wrong.

tests/Cms/App/fixtures/controllers/another.json.php
bastianallgeier: I have no idea. The controller does not make any sense in that form.

bastianallgeier (approved):

PR #6951, opened 2025-01-25
Blocks field: fix empty display

### LINKED ISSUE #5290: Drag'n'drop between block fields broken

distantnative:
### Summary of changes
The blocks field is currently glitchy when showing the empty placeholder. As it so far relies on `data-empty` attribute, it still gets shown even when a block from another blocks field is moved over (as it technically is only the sortable ghost and not present yet in the blocks data array).

This PR tries to solve this rather by utilizing CSS `:empty` and `:has()` to decide when the `.-k-blocks-list` is truly empty of blocks but also sortable ghosts and spending on that show/hide the empty placeholder.

With this change (and the previous sortable change), the ancient https://github.com/getkirby/kirby/issues/5290 seems to be solved.

### Additional context
I didn't run into any errors with this last state. But would feel better if @afbora hits it hard with your bug finding skills.

### Fixes
- Blocks field: drag & drop between different blocks fields now functional

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Blocks/Blocks.vue
+++ b/panel/src/components/Forms/Blocks/Blocks.vue
@@ -54,18 +54,12 @@
 					@split="split(block, index, $event)"
 					@update="update(block, $event)"
 				/>
-
-				<!-- No blocks -->
-				<template v-if="blocks.length === 0" #footer>
-					<k-empty
-						class="k-blocks-empty"
-						icon="box"
-						@click="choose(blocks.length)"
-					>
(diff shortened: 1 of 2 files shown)


afbora (approved):
Yes, looks like works great 🎉

PR #6953, opened 2025-01-25
Redirect `html` representation URLs

### LINKED ISSUE #6509: Duplicate content when appending `.html` to page URLs
## Description

Kirby renders the same page if `.html` is appended to the page URL as if no content representation is appended at all.

**Expected behavior**  

`.html` should behave like `.somethinginvalid` and render a 404 unless a template `.html.php` actually exists.

## To reproduce

Visit https://getkirby.com/docs/guide/quickstart and https://getkirby.com/docs/guide/quickstart.html

## Your setup
**Kirby Version**  

4.3.0

**Additional context**  

I believe this is caused by `$page->render(array $data = [], $contentType = 'html')`. Because `$contentType` defaults to `html` if no content representation was called, it cannot differentiate between `page.html` and `page`. In both cases it gets `$contentType = 'html'`.

So to fix this, we would need to change the default to `string|null $contentType = null`. Only if `empty($contentType) === true`, the main template would be loaded. `html` would then load a `.html.php` template if it exists.

This would be a breaking change, so v5 would be a good candidate.

distantnative:
### Summary of changes
- Requests for a page's `html` representation get redirected to the normal page URL without extension

### Reasoning
Avoiding duplicate content, e.g. at `https://getkirby.com/docs` as well as `https://getkirby.com/docs.html`.

I have explained in https://github.com/getkirby/kirby/issues/6509#issuecomment-2350977276 why I think making Kirby stricter here and just delivering content if a `docs.html.php` template exists would be the wrong direction IMO, considering that within the Kirby system the one without any extension is directly tied to `html` as type. Having the option to have a `docs.php` alongside `docs.html.php` would be a contradiction.

### Fixes
- Requests for a page's `html` representation get redirected to the normal page URL without extension

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1214,8 +1214,10 @@ class App
 	 * @internal
 	 * @throws \Kirby\Exception\NotFoundException if the home page cannot be found
 	 */
-	public function resolve(string|null $path = null, string|null $language = null): mixed
-	{
+	public function resolve(
+		string|null $path = null,
+		string|null $language = null
+	): mixed {
 		// set the current translation
 		$this->setCurrentTranslation($language);
 
@@ -1263,6 +1265,12 @@ class App
 		// only try to return a representation
 		// when the page has been found
 		if ($page) {
+			// if extension is the default content type,
+			// redirect to page URL without extension
+			if ($extension === 'html') {
+				return Response::redirect($page->url(), 301);
+			}
+
 			try {
 				$response = $this->response();
 				$output   = $page->render([], $extension);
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
I agree with the reasoning and I really like how simple the solution is. @lukasbestle what do you think?

bastianallgeier:
I'm going to merge it. I think this is definitely the better solution compared to the status quo

lukasbestle:
I only now saw the notification. Weird that GH no longer seems to send push notifications for direct mentions?

I agree that this is already a great step. Thinking about my reasoning in the linked issue again, I think I forgot about the `Content-Type` response header: With HTML being the default response type, we can also default to `text/html` as the `Content-Type`. Without any default, Kirby wouldn't know what to do. So there's probably no way around a default.

PR #6956, opened 2025-01-25
Update README.md

distantnative:


bastianallgeier:
I think we should keep those blocks consistent (see CLI readme)

distantnative:
The CLI is the only one diverting 😆 Can we change that one? I would prefer not to go through a dozen+ repos.

bastianallgeier:
Oh I did not see that you changed it everywhere else already. Of course, then let's fit in in the CLI repo

bastianallgeier (approved):

bastianallgeier:
The stuck tests are so annoying for this simple change :)

distantnative:
No idea how to get them unstuck

bastianallgeier:
I've made the change now on the release/4.6.0 branch to fix this

PR #6958, opened 2025-01-28
Option to switch off the Vue template compiler

bastianallgeier:
## Summary of changes

- A new vue.runtime.js is added to the dist folder and loaded on demand when the compiler is disabled
- A new config option (see below) is used in the Assets class to load the right file
- An info box is shown in lab examples if the compiler is disabled. Lab examples can no longer be displayed
- A new security warning has been added to the system view when the compiler is active. This might need to be less aggressive. 

## Todos

- [ ] The security warning leads to https://getkirby.com/security/vue-compiler which does not exist yet
- [x] The text for the security warning is not yet translated.
- [ ] Add unit test for `Assets::vue`

### Enhancements

- The Vue template compiler can be switched off in the config. This will load the vue runtime version. Plugins and the lab views that rely on the template compiler will no longer work. 
```php
// `/site/config/config.php`
return [
  'panel' => [
    'vue' => [
      'compiler' => false
    ]
  ]
]
```


config/areas/system/views.php
```diff
@@ -86,6 +86,14 @@
 				];
 			}
 
+			if ($kirby->option('panel.vue.compiler', true) === true) {
```
distantnative: `Suggestion:` I think this should be only visible during debug mode.

panel/src/components/Lab/PlaygroundView.vue
```diff
@@ -9,9 +9,16 @@
 		</k-header>
 		<k-tabs :tab="tab" :tabs="tabs" />
 
-		<component :is="component" v-if="component" v-bind="props" />
-		<!-- eslint-disable-next-line vue/no-v-html, vue/no-v-text-v-html-on-component -->
-		<component :is="'style'" v-if="styles" v-html="styles" />
+		<template v-if="compiler === false">
```
distantnative: `Question:` Is there an extra benefit of the template tags over putting the `v-if` directly on the `k-box`?

bastianallgeier: No, there isn't. I don't know why I prefer it sometimes. It just happens :)

distantnative: 😄 I'll allow it 😆 was just curious

bastianallgeier:
Closed in favor of https://github.com/getkirby/kirby/pull/6965

PR #6961, opened 2025-01-29
Core component `sessions` for custom session store

distantnative:
### Feature
- New `session::store` core component: return your custom `SessionStore` handler 

## Ready?

### For review team

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -266,7 +266,7 @@ return [
 	/**
 	 * Add your own session store
 	 */
-	'sessions' => function (App $kirby): string|SessionStore {
+	'session::store' => function (App $kirby): string|SessionStore {
 		return $kirby->root('sessions');
 	},
 
--- a/config/components.php
+++ b/config/components.php
@@ -14,6 +14,7 @@ use Kirby\Filesystem\Filename;
 use Kirby\Http\Uri;
 use Kirby\Http\Url;
 use Kirby\Image\Darkroom;
+use Kirby\Session\SessionStore;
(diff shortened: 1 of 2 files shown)


bastianallgeier (dismissed):
I love the idea. @lukasbestle should have a final look at it, but I think it's great to move more parts to core components.

lukasbestle:
For v5 we added a `storage` component. I think the naming is quite confusing.

Maybe we should rename `storage` to `contentStorage` and the new `sessions` to `sessionStorage`. Should there be other storage types in the future, we would already have a clear and consistent pattern.

distantnative:
Would these be anything else around `sessions`? Otherwise I'd stick with that name, the storage suffix doesn't make it easier to understand IMO.

lukasbestle:
I don't think we need any other session-related component at this time. We might in the future, but that's not the point I'm trying to make.

I think it would increase consistency by using the same suffix for all storage-related components. Also all existing components (except `smartypants`, which is a given name) are singular. By using `sessionStorage` or `session::storage`, we could keep this pattern.

distantnative:
`session::storage` seems good to me - @bastianallgeier also ok?

lukasbestle:
I think we should also use the opportunity and rename `storage` to `content::storage` in the `v5/develop` branch before the RC.

bastianallgeier:
I like the idea to reuse the :: naming pattern and I agree that we should use it for content as well.

lukasbestle (approved):

PR #6963, opened 2025-01-30
Show info text when variable is an array 

afbora:
## Description

Currently when variable is an array, error thrown. For now shows info text when variable is an array. We'll fix the issue permanently in K5.

### Fixes
- #6930

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/config/areas/languages/dialogs.php
+++ b/config/areas/languages/dialogs.php
@@ -244,7 +244,9 @@ return [
 
 			// shows info text when variable is an array
 			// TODO: 5.0: use entries field instead showing info text
-			if (is_array($variable->value()) === true) {
+			$isVariableArray = is_array($variable->value()) === true;
+
+			if ($isVariableArray === true) {
 				$fields['value'] = [
 					'label' => I18n::translate('info'),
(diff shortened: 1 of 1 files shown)


afbora:
Also what do you think about hiding submit button when variable is an array?

develop test_panel_languages_en (1)

bastianallgeier:
I really like the simplicity of this solution. I also think the info text is great and I would agree that the submit button should be removed. I think we can remove the cancel button as well. You can simply click outside the dialog to close it and it's not that necessary here.

afbora:
@bastianallgeier Thank you. I'll do. Any idea about the text? Is that good?

> You are using an array variable for this key. Please modify it in the language file in /site/languages

Do you want to improve it?

bastianallgeier:
@afbora no, I think the text is perfectly fine.

bastianallgeier (commented):
Do we want to translate the message or is it so temporary that we don't really need to?

afbora:
I think no need to translate since we'll remove the changes in K5.

bastianallgeier (approved):

PR #6965, opened 2025-01-30
New option to disable the vue template compiler

bastianallgeier:
## Summary of changes

- A new vue.runtime.js is added to the dist folder and loaded on demand when the compiler is disabled
- A new config option (see below) is used in the Assets class to load the right file
- An info box is shown in lab examples if the compiler is disabled. Lab examples can no longer be displayed
- A new security warning has been added to the system view when the compiler is active. This might need to be less aggressive. 

## Todos

- [ ] The security warning leads to https://getkirby.com/security/vue-compiler which does not exist yet
- [x] The text for the security warning is not yet translated.
- [x] Add unit test for `Assets::vue`

### Enhancements

- The Vue template compiler can be switched off in the config. This will load the vue runtime version. Plugins and the lab views that rely on the template compiler will no longer work. 
```php
// `/site/config/config.php`
return [
  'panel' => [
    'vue' => [
      'compiler' => false
    ]
  ]
]
```


config/areas/system/views.php
```diff
@@ -80,11 +81,19 @@
 				];
 			}
 
+			if ($debugMode === true && $kirby->option('panel.vue.compiler', true) === true) {
```
lukasbestle: I get why this security warning depends on the debug mode: If we would always display it, we would have a security warning for the default configuration of the `panel.vue.compiler` option, which is weird. But IMO making it dependent on the debug mode is also weird – it's not an information one needs during debugging, but especially in production.

So I wonder if we can find a different solution:

- Maybe we can find a different way or place to display this warning in a context-sensitive way.
- Maybe it is possible to auto-detect whether any installed plugin needs the template compiler and control the default value of the option accordingly? Then we could display this warning even without debug mode.
- ...

distantnative: I wouldn't display it without debug mode or some other restriction (admins only?) cause the choice to have it might be done very deliberately by the developer, but then the users/customers freak out due to this warning if it's shown to them in production.

lukasbestle: That's true. I like the idea of admin-only.

distantnative: Or we only show it as long as the config option isn't set (so instead of passing a fallback/default we first check for null).

bastianallgeier: I love that last idea.

bastianallgeier: I've implemented it like that.

bastianallgeier:
I've had a closer look at the coverage and I really think it's another issue from Codecov. The lines are all covered.

config/areas/system/views.php
distantnative: @bastianallgeier Now he message is shown when set to `true` explicitly. My suggestion was rather to check for null, show it as long as no config option is set.

bastianallgeier: 🙈 of course! I fixed it.

distantnative: @bastianallgeier sorry but I think it's still not correct: It should **not** be shown when no config option is set, the if check needs to be 

```php
if ($compiler === true)
```

Not with the `|| $compiler === null` part.

bastianallgeier: Unbelievable. I think my brain got fried by the writer issue this morning.

bastianallgeier: Wait, but then my previous one was already correct, right?

distantnative: Was I the confused one? I thought you were still checking for a true that came as default. Sorry if my brain was the fried one.

bastianallgeier: Doesn't matter. I think it's correct now :)

distantnative (dismissed):

config/areas/system/views.php
lukasbestle: Either my brain is also fried or it's still not correct.

The current implementation:

```php
if ($kirby->option('panel.vue.compiler', null) === true) {
```

uses `null` as the default and checks for an explicit value of `true`. So the warning is displayed if and only if the option is manually set to `true`. Which would be exactly the opposite of what Nico proposed.

For that it would need to be `, null) === null`. Then the warning would be displayed if the option is not set at all. In this case the default of `true` in `Assets::vue()` will enable the compiler by default but the warning would tell users to check if they need it for their project so they can either set `true` if they do (and don't need protection against the template compiler vulnerabilities) or set it to `false` to protect their sites. Both would then disable the warning. At least that's how I understood @distantnative's idea?

config/areas/system/views.php
```diff
@@ -80,11 +81,19 @@
 				];
 			}
 
+			if ($kirby->option('panel.vue.compiler', null) === true) {
+				$security[] = [
+					'id'   => 'compiler',
```
lukasbestle: `Suggestion`: For clarity and unambiguity:

```suggestion
					'id'   => 'vue-compiler',
```

i18n/translations/en.json
```diff
@@ -641,6 +641,7 @@
 	"system.issues.https": "We recommend HTTPS for all your sites",
 	"system.issues.kirby": "The kirby folder seems to be exposed",
 	"system.issues.site": "The site folder seems to be exposed",
+	"system.issues.vue.compiler": "The Vue template compiler is enabled",
```
lukasbestle, line 644: `Suggestion`: To make the warning more actionable:

```suggestion
	"system.issues.vue.compiler": "The Vue template compiler is enabled. Please follow this link and review if it can be disabled for your site to improve security.",
```

(Maybe too long, feel free to change it.)

We could also think about setting the `back` and/or `icon` properties to make it more a warning than an urgent issue.

lukasbestle (commented):
About coverage: Probably rather than a Codecov issue, this is another PHPUnit issue where the method name cannot be marked as covered. We had that a lot, so I wouldn't be surprised if the name `js` is also affected. Nothing to worry about IMO.

config/areas/system/views.php
distantnative: Yep! See my brain also got fully twisted at the end. But you are right, the warning should be shown when no config option is set - because in the assets part we do load it automatically when no config option is set.

distantnative:
It seems almost as if short methods are the problem (js, vue, url...). And if we say Codecov we mean PHPunit... just the coverage part of our CI overall getting it wrong.

lukasbestle:
Short names could indeed be a connection!

i18n/translations/en.json
distantnative: It's a bit long and isn't the second part true for any of the security warnings we list?

Implemented a different theme 👍

distantnative (approved):

bastianallgeier:
@distantnative the themes are great!

PR #6969, opened 2025-02-03
New Area and MenuItem classes

bastianallgeier:
## Summary 

- New `Kirby\Panel\Ui\MenuItem` class
- New `Kirby\Panel\Area` class
- Refactored Areas and Menu classes
- The Fiber class no longer takes a wildcard options array, but has new dedicated props for areas and area
- Other panel classes have been refactored to work with the new Area instances
- Unit tests have been adapted and new unit tests have been added.

### DIFF
--- a/src/Panel/Router.php
+++ b/src/Panel/Router.php
@@ -36,7 +36,7 @@ class Router
 		$this->garbage();
 
 		// collect areas
-		$areas  = $this->panel->areas()->toArray();
+		$areas = $this->panel->areas()->toArray();
 
 		// create a micro-router for the Panel
 		return BaseRouter::execute(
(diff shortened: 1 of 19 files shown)


distantnative:
I think it would be great if `Ui\Menu` and `Ui\MenuItem` would follow our UI component classes logic, e.g. extending `Ui\Component` (or `Ui\Button` for the menu items) but at least working in the same `::render()` with `component`/`props` logic. What do you think?

src/Panel/Router.php
```diff
@@ -159,15 +159,15 @@ public static function routes(array $areas): array
...
-				...static::routesForDropdowns($id, $area),
-				...static::routesForRequests($id, $area),
+				...static::routesForViews($area),
+				...static::routesForSearches($area),
+				...static::routesForDialogs($area),
+				...static::routesForDrawers($area),
+				...static::routesForDropdowns($area),
+				...static::routesForRequests($area),
 			];
 		}
```
distantnative, line 172: `Nitpick:` Not necessarily in this PR, but looking at this it just screams to me for 

```php
...$area->routes()
```

distantnative:
Looks great! I'd suggest to merge https://github.com/getkirby/kirby/pull/6970 first in here and then together in the other PR.

PR #6970, opened 2025-02-03
Menu/MenuItem as Ui Components

distantnative:
- `Menu` and `MenuItem` are `Ui\Component` classes
- Move all panel module specific code out of `k-panel-menu`
- Allow config options that include a `Ui\Component` object directly
  - Not really meant for anything now, but I could see this a way for customization for developers to include something else than menu buttons in the menu (e.g. headlines for button groups)


bastianallgeier (approved):
This is great. I've added a couple unit tests and I think it's ready.

PR #6971, opened 2025-02-03
Require current password when user changes own pwd

distantnative:
### Enhancements
- Panel: user needs to confirm current password when changing their own password

## Ready?

### For review team

### DIFF
--- a/config/areas/users/dialogs.php
+++ b/config/areas/users/dialogs.php
@@ -3,6 +3,7 @@
 use Kirby\Cms\App;
 use Kirby\Cms\Find;
 use Kirby\Cms\UserRules;
+use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Panel\Field;
 use Kirby\Panel\Panel;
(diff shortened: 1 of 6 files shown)


config/areas/users/dialogs.php
```diff
@@ -180,31 +181,58 @@
...
+			];
+
+			if ($user->is($user->kirby()->user()) === true) {
+				$fields = [
+					'currentPassword' => Field::password([
+						'label' => I18n::translate('user.changePassword.current'),
+					]),
+					...$fields
+				];
+			}
```
bastianallgeier: This means that the current user has a harder process than any other user which might be allowed to change passwords. Is this the right way to solve this? Should we maybe always ask for the password of the authenticated user - even when they want to change the password of another account?

distantnative: That makes sense. I was just thinking of the current password of the affected user - and of course no other user could enter it. But having to always enter one's own password sounds right. What should the label be then? "Your current password"?

bastianallgeier (approved):
Works like a charm

guidoferreyra:
Can this be disabled? As admin I need to reset password for users and I logically I don’t know their passwords

distantnative:
@guidoferreyra You don't need to know their passwords. You have to enter your current one and then their new one.

guidoferreyra:
ha! thanks

PR #6973, opened 2025-02-05
🎉 New `entries` field

afbora:
## Description

v5-develop test_panel_pages_entries-field_tab=types (1)

### Blueprint setup

````yaml
fields:
  entries:
    type: entries
````

#### Extended usage
You can use `field` prop to provide different field types.

````yaml
fields:
  select:
    type: entries
    min: 2
    max: 4
    required: true
    field:
      type: select
      options:
        design: Design
        architecture: Architecture
        photography: Photography
        3d: 3D
        web: Web
````

### Content storage
````
----

Text:

- Some text
- New text
- Another text

----
````

### Usage in template
````php
<ul>
    <?php foreach($page->entries()->yaml() as $entry): ?>
        <li><?= $entry ?></li>
    <?php endforeach ?>
</ul>
````

Currently only `color`, `date`, `email`, `number`, `select`, `slug`, `tel`, `text`, `time`, `url` fields supported.

## Ready?

### For review team

### CODE REVIEW

File: panel/src/components/Forms/Field/EntriesField.vue [hunk 1]
```diff
@@ -0,0 +1,438 @@
...
+		/**
+		 * The text, that is shown when the field has no entries.
+		 */
+		empty: String,
+		/**
+		 * Field attrs for the form
+		 */
+		field: {
+			type: Object,
+			default: () => {}
```
distantnative:
```suggestion
			default: () => ({})
```

panel/src/components/Forms/Field/EntriesField.vue
```diff
@@ -0,0 +1,438 @@
...
+				if (entries === this.values) {
+					return;
+				}
+
+				this.entries = entries.map((value) => {
+					return {
+						id: this.$helper.uuid(),
+						value
+					};
+				});
```
distantnative:
```suggestion
				this.entries = entries.map((value) => ({
					id: this.$helper.uuid(),
					value
				}));
```

panel/src/components/Forms/Field/EntriesField.vue
```diff
@@ -0,0 +1,438 @@
...
+				id: this.$helper.uuid(),
+				value: value ?? ""
+			};
+
+			if (index !== null) {
+				this.entries.splice(index, 0, entry);
+			} else {
+				this.entries.push(entry);
+				index = this.entries.length - 1;
+			}
```
distantnative: Not sure if mine is really better

```suggestion
			index ??=  this.entries.length - 1;
			this.entries.splice(index, 0, entry);
```

panel/src/components/Forms/Field/EntriesField.vue
```diff
@@ -0,0 +1,438 @@
...
+
+			this.save();
+
+			this.$nextTick(() => {
+				this.focus(index);
+			});
```
distantnative:
```suggestion
			await this.$nextTick();
			this.focus(index);
```

(but then the method needs to become `async` too).

panel/src/components/Forms/Field/EntriesField.vue
```diff
@@ -0,0 +1,438 @@
...
+			if (input instanceof ClipboardEvent) {
+				input = this.$helper.clipboard.read(input, true);
+			}
+
+			const entries = input
+				.split(/\r?\n/)
+				.filter(entry => entry.trim().length > 0)
+				.map(entry => entry.startsWith("- ") ? entry.slice(2) : entry);
+
+			entries.forEach(entry => this.add(null, entry));
```
distantnative:
```suggestion
			for (const entry of entries) { 
				this.add(null, entry);
			}
```

afbora: Sometimes I write the short version, thinking @distantnative would write shorter, and then we move on to the long version. When I write the long version, and then we move on to the short version. What's the logic behind this? Can someone explain it to me? 🙈😄

distantnative: :D Sorry, it might be confusing as it's not always about shorter/longer.

In this case, we have moved away from using too many methods where the following code is nested in a callback. Which is why we dropped most uses of `.forEach()` completely and went with `for ()` loops instead. But it's a total `Nitpick` - you don't need to accept it :)

bastianallgeier (approved):
I'm happy with this first version. Maybe @distantnative wants to have a second look though.

distantnative:
Just one small thing: in other fields (structure etc.) one has to confirm the deletion also of a single entry via a remove dialog. I'd do that also for the entries field.

bastianallgeier:
@distantnative you are probably right. We should keep it consistent. I haven't thought about it.

afbora:
I don't quite agree. There can be many fields/data in a row of structure field . But there is only 1 data in a row of entries. Entries should be easy to add and remove. It may be difficult for users to confirm to delete each row one by one. We already have a confirmation for delete all.

bastianallgeier:
Ok, let's leave it like it is and maybe get a better feeling for it during the remaining beta. We can still add a dialog or batch delete or whatever later.

PR #6979, opened 2025-02-07
Fix entries field overlook issues

afbora:
- [x] Fixed discarding issue via storing field value correctly
- [x] Run `toFormValue`/`toStoredValue` methods for each value

### DIFF
--- a/src/Form/Field/EntriesField.php
+++ b/src/Form/Field/EntriesField.php
@@ -9,6 +9,7 @@ use Kirby\Form\Form;
 use Kirby\Form\Mixin\EmptyState;
 use Kirby\Form\Mixin\Max;
 use Kirby\Form\Mixin\Min;
+use Kirby\Toolkit\A;
 use Kirby\Toolkit\Str;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Form/Field/EntriesField.php
```diff
@@ -126,21 +123,25 @@ public function supports(): array
...
-
-		if ($value === null) {
-			return null;
+		$value = [];
+		foreach (parent::toFormValue($default) ?? [] as $val) {
+			$value[] = $this->form([$val])
+							->fields()
+							->first()
+							->toFormValue();
 		}
```
distantnative:
```suggestion
		$value = A::map(
				$parent::toFormValue($default),
				fn ($value) => $this->form([$value])->fields()->first()->toFormValue()
		);
```

src/Form/Field/EntriesField.php
```diff
@@ -126,21 +123,25 @@ public function supports(): array
...
-
-		if ($value === null) {
-			return '';
+		$value = [];
+		foreach ($this->toFormValue($default) as $val) {
+			$value[] = $this->form([$val])
+							->fields()
+							->first()
+							->toStoredValue();
 		}
```
distantnative:
```suggestion
		$value = A::map(
				parent::toStoredValue($default),
				fn ($value) => $this->form([$value])->fields()->first()->toStoredValue()
		);
```

src/Form/Field/EntriesField.php
afbora: I swear. I wrote the short version like yours but made it longer to make it readable 🤣 Btw feel free to push to the branch your changes.

distantnative: 😅👍

distantnative:
Great fix to make it functioning properly.

I would still love to move the UUID injection to the backend, but that's its own project and not necessary to have the field function - it does with this PR.

PR #6980, opened 2025-02-07
Entries field: markup improvements

distantnative:
- Simplifies DOM (I think we could get rid of some wrapper DIVs or am I overlooking something?)
- Some CSS simplification

### DIFF
--- a/panel/src/components/Forms/Field/EntriesField.vue
+++ b/panel/src/components/Forms/Field/EntriesField.vue
@@ -18,7 +18,7 @@
 					variant="filled"
 					size="xs"
 					class="input-focus"
-					@click="add"
+					@click="add()"
 				/>
 				<k-button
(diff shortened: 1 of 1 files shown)


panel/src/components/Forms/Field/EntriesField.vue
```diff
@@ -101,7 +18,7 @@
 					variant="filled"
 					size="xs"
 					class="input-focus"
-					@click="add()"
+					@click="add"
```
afbora: Actually I was wrote like in initial version that but doesn't work properly. Idk but works when `add()`. Probably passing event as first argument or something like that.

distantnative: Ahh yes, of course! I'll revert my change.

afbora (approved):

PR #6981, opened 2025-02-07
Fix entries field issues

afbora:
### Fixes

- [x] Min/max validation
- [x] Default value for the field form
- [x] Disabled state
- [X] Input validator when empty state
- [x] Consistency with min trait and array

### Breaking changes
- `isRequired()` method of blocks/layout fields that use the `min` prop now returns `true` for empty state and `min` returns `1` for fields that no defined `min` prop and `requried: true`.


bastianallgeier (approved):

bastianallgeier:
I've added one last change and removed the add button in the header when more === false

bastianallgeier (approved):

PR #6985, opened 2025-02-09
PHPUnit replace covers annotations with attributes

distantnative:
## Description

- [x] https://github.com/getkirby/kirby/pull/6986
- [ ] Figure out why we're losing coverage %

### Summary of changes
- Replaces PHPUnit `@covers`/`@coversDefaultClass` and `@dataProvider` annotations with PHP attributes

### Reasoning
PHPUnit will remove docblock annotation support in v12, already throwing deprecation notices in v11.

PHPUnit does offer a `CoversMethod` attribute. However, this can only be applied per class, not to a specific test method. As such, I don't think there is much value in it. That's why I followed a only `CoversClass` approach (which is also the preferred one by the maintainer of PHPUnit).

### Additional context
Doesn't remove all PHPUnit deprecation errors but the biggest chunk.

### Housekeeping
- PHPUnit: use attributes for coverage and data providers

## Ready?

### For review team

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -264,7 +264,7 @@ jobs:
         with:
           php-version: ${{ env.php }}
           coverage: none
-          tools: php-cs-fixer:3.52.1
+          tools: php-cs-fixer:3.65.0
 
       - name: Cache analysis data
         id: finishPrepare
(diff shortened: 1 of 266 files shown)


distantnative:
Closing this - will do smaller PRs per package or so later on.

PR #6986, opened 2025-02-09
Tests: fix `Form\Field` namespace

distantnative:
Tests and actual code should be in the same namespace (`Kirby\Form\Field`).

----

Not quite understanding why the test I didn't touch now throwing errors.


afbora (approved):
I've fixed the broken unit test. Don't ask me why it was failing and how I fixed. I don't know 😅 (Probably this was about `FieldTest` tests that run before `FieldsTest`). I've done dozen tests and I've found the correct setup for the unit test setup.

distantnative:
Ahhh perfect, thanks! And indeed, probably because of the different order the test classes now ran.

PR #6988, opened 2025-02-10
Fix `api.methodOverride` option

### LINKED ISSUE #6984: Reconsider `api.methodOverwrite` option
See https://github.com/getkirby/kirby/pull/6650#issuecomment-2634970023:

- [x] Remove automated check as it doesn't work with the default of `true`
- [x] Rename option to `api.methodOverride` (overwrite was a typo when we originally introduced it in the Panel code)

distantnative:
### Summary of changes
- Corrected option name
- Removed JS automatic system check
- Fixed console error when local (no `urls`)

### Breaking changes
- `api.methodOverwrite` option has been renamed to `api.methodOverride`

## Ready?

### For review team

### DIFF
--- a/src/Panel/View.php
+++ b/src/Panel/View.php
@@ -262,7 +262,7 @@ class View
 		return [
 			'$config' => fn () => [
 				'api'         => [
-					'methodOverwrite' => $kirby->option('api.methodOverwrite', true)
+					'methodOverride' => $kirby->option('api.methodOverride', true)
 				],
 				'debug'       => $kirby->option('debug', false),
 				'kirbytext'   => $kirby->option('panel.kirbytext', true),
--- a/config/api/routes/system.php
+++ b/config/api/routes/system.php
@@ -32,18 +32,6 @@ return [
 		}
 	],
 	[
-		'pattern' => 'system/method-test',
-		'method'  => 'PATCH',
-		'action'  => function () {
-			return [
-				'status' => match ($this->kirby()->request()->method()) {
-					'PATCH' => 'ok',
-					default => 'fail'
-				}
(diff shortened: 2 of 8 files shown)


bastianallgeier (approved):

src/Panel/View.php
```diff
@@ -262,7 +262,7 @@ public static function globals(): array
 		return [
 			'$config' => fn () => [
 				'api'         => [
-					'methodOverwrite' => $kirby->option('api.methodOverwrite', true)
+					'methodOverride' => $kirby->option('api.methodOverride', true)
```
lukasbestle, line 265: `Suggestion` to keep backwards-compatibility:

```suggestion
					// TODO: Remove fallback to `api.methodOverwrite` in v6
					'methodOverride' => $kirby->option('api.methodOverride', $kirby->option('api.methodOverwrite', true))
```

We could add "Deprecated: Option `api.methodOverwrite`: Use `api.methodOverride` instead" into the release notes. I'd also be fine with the breaking change though as it's unlikely many users would be affected.

lukasbestle (commented):
Thanks for taking care of it. I have removed the leftover translation that is no longer used and have one minor suggestion:

src/Panel/View.php
distantnative: I'll let Basti decide on it. Totally valid suggestion. My take here was that it's gonna affect so few to none that we might just get it over right with v5.

bastianallgeier: I would agree with @distantnative here. It seems like one of the really minor breaking changes to me for v5 and in return we get a cleaner setup.

PR #6989, opened 2025-02-10
Version Cache

bastianallgeier:
### Todos

- [x] Unit tests for the VersionCache class

### Summary of changes

- New Kirby\Content\VersionCache class
- The cache class is used in the Version class to avoid multiple reads
- The VersionCache is cleared whenever the App constructor is called

### Enhancements

- Multiple storage reads are avoided with a new VersionCache class.

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -5,6 +5,7 @@ namespace Kirby\Cms;
 use Closure;
 use Generator;
 use Kirby\Content\Storage;
+use Kirby\Content\VersionCache;
 use Kirby\Data\Data;
 use Kirby\Email\Email as BaseEmail;
 use Kirby\Exception\ErrorPageException;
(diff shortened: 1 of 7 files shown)


distantnative:
No full review but to not forget this:
We need to ensure that the key never points to another model. Which right now could happen if in the same request two pages change their slug/id and page B ends up with the cache key of page A.

I wanted to suggest to use it in `::purge()` but I realized that `::purge()` isn't really used in our real code but mainly by our plugins (which sheds a whole new light on all the unit tests failing because of purge with LabPage if that's really almost just a tests thing).

bastianallgeier:
@distantnative it's a problem that not all actions go through the Version and Storage classes yet. This is quite a good example for that. I think we really need to go through all the model actions again and check for parts where this could happen.

distantnative (commented):
I'm a bit torn whether I like this global cache instead of model-specific. And sometimes we try to remove/add specific models, sometimes we just hard reset. Doesn't feel fully elegant. But maybe it's an easier compromise to do it that broad.

bastianallgeier:
To be honest, this has been the approach that felt most straight forward to me. I agree that it is not super elegant though. It is 100% internal though. If you want to give it a go to build a model-based version, that'd be great. But otherwise I think the main objective is to get it done for v5.

distantnative (approved):
Probably indeed the most pragmatic approach right now.

PR #6990, opened 2025-02-11
Remove paste action from entries field

afbora:
Paste action removed


bastianallgeier:
While the fix looks ok on first sight, I'm tempted to say we should remove the paste feature for beta 3. Just to keep the feature set as stable as possible. This seems like it could hit us in other ways later.

afbora:
@bastianallgeier I've removed.

bastianallgeier (approved):

PR #6996, opened 2025-02-12
 New Languages::migrateToSingleLang & Languages::migrateToMultiLang methods

bastianallgeier:
## Description

This PR is trying to clean up the last Storage and Version parts that we weren't super happy with so far. 

### Summary of changes

- New `Languages::migrateToSingleLang` method
- New `Languages::migrateToMultiLang` method
- Removed `Storage::deleteLanguage` method
- Removed `Storage::moveLanguage` method
- Removed `Storage::touchLanguage` method

### Reasoning

Keeping the logic parts in the Storage classes low means that they are easier to extend and adapt for multiple storage types. 

## Ready?

### For review team

### DIFF
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -2,6 +2,7 @@
 
 namespace Kirby\Cms;
 
+use Kirby\Content\VersionId;
 use Kirby\Data\Data;
 use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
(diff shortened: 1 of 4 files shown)


bastianallgeier:
@distantnative @lukasbestle to be honest, I'm not sure about this on the day after. Is it worth to actually tackle this at this point? Is the Languages class really a better place for it?

distantnative:
@bastianallgeier in general I think it does make sense. Maybe I"d change a few things style-wise, but it makes more sense to me to have these in the language classes (some as we have many of the model actions in the model classes and not in version/storage). But if you aren't sure, IMO this doesn't need to happen still for v5 but could also happen later.

bastianallgeier:
@distantnative I'm closing this for now. The implementation is not 100% clear and we really don't need it for v5

Issue #6959, opened 2025-01-28
[v5] Block fields preview: `content.save` event is triggered although content hasn’t changed


SebastianEberlein-JUNO:
## Description
The `content.save` event is triggered in a block `fields` preview although content hasn’t changed, see video below.

**Expected behavior**  
If possible, only trigger the event when there are actual content changes. 
(We are trying to use this event reload the preview page automatically, i.e. extending our [reload on save plugin](https://github.com/junohamburg/kirby-reload-on-save).)

**Video**  
https://github.com/user-attachments/assets/e205bdf6-3e79-444d-b598-eeb17c10fa5a

## To reproduce
1. In the Starterkit, add this plugin and find a page with blocks.
2. Add text to a block, then just click around in the text without changing the text. The console should still log the event.

```js
panel.plugin("plugin/test", {
  created() {
    window.panel.events.on("content.save", () => {
      console.log("content.save", new Date().toLocaleString());
    });
  },
});
```

## Your setup
**Kirby Version**  
5.0.0 beta 2

**Your system**
Chrome 132, Safari 18.3, Firefox 134 on macOS 15.3

afbora:
@distantnative I've an idea about the solution. Compare the changes for each update event and only trigger the save event when blocks have any changes. Roughly something like following. What do you think?


````js
update(block,content ) {
	let hasChanges = false;
	const index = this.findIndex(block.id);

	if (index !== -1) {
		for (const key in content) {
			if (this.blocks[index].content[key] !== content[key]) {
				hasChanges = true;
				set(this.blocks[index].content, key, content[key]);
			}
		}
	}

	if (hasChanges) {
		this.save();
	}
}
````

https://github.com/getkirby/kirby/blob/main/panel/src/components/Forms/Blocks/Blocks.vue#L709-L717


More simpler version:

````js
update(block, content) {
	const index = this.findIndex(block.id);
	if (index === -1) return;

	const targetBlock = this.blocks[index].content;

	// compare the content changes
	const hasChanges = Object.entries(content).some(([key, value]) => {
		if (targetBlock[key] !== value) {
			set(targetBlock, key, value);
			return true;
		}
		return false;
	});

	// only trigger save event when has changes
	if (hasChanges) {
		this.save();
	}
}
````

bastianallgeier:
@afbora I think this is another writer field issue

bastianallgeier:
[...]

PR #7001, opened 2025-02-13
Fix writer update issue

bastianallgeier:
## Description

The writer input records transactions for actual state updates but also for selections. So far, we don't check before we emit an input event. This leads to a quite nasty situation where selections or even just a focus even also emit an input event at the same time. With the new changes setup, this means that every selection change or focus event will send an update to the backend. This PR fixes this by checking for the involved steps in the transaction object. If there are no steps, there's no need to emit something. 

### Fixes

- https://github.com/getkirby/kirby/issues/6959

### Breaking changes

None

### For review team

### DIFF
--- a/panel/src/components/Forms/Input/WriterInput.vue
+++ b/panel/src/components/Forms/Input/WriterInput.vue
@@ -201,16 +201,6 @@ export default {
 						return;
 					}
 
-					// if no steps are recorded, this is a selection change
-					// and we don't need to update the content
-					if (payload.transaction.steps.length === 0) {
-						return;
-					}
-
-					if (payload.transaction.meta.preventUpdate === true) {
-						return;
-					}
-
 					// compare documents to avoid minor HTML differences
 					// to cause unwanted updates
 					const jsonNew = JSON.stringify(this.editor.getJSON());
--- a/panel/src/components/Forms/Input/WriterInput.vue
+++ b/panel/src/components/Forms/Input/WriterInput.vue
@@ -207,6 +207,10 @@ export default {
 						return;
 					}
 
+					if (payload.transaction.meta.preventUpdate === true) {
+						return;
+					}
+
 					// compare documents to avoid minor HTML differences
 					// to cause unwanted updates
 					const jsonNew = JSON.stringify(this.editor.getJSON());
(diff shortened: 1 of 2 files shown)


afbora (approved):
Brilliant!

bastianallgeier:
@afbora I've just made another change. We never really watch the preventUpdate meta data entry in the update method. But this is set on purpose in the value watcher to avoid sending additional input events when the value changed from the backend. 

Before, hitting the discard button would always lead to another input event being triggered by the writer. After this change, the additional input events are gone.

afbora:
@bastianallgeier Actually we watch it. Is this same?
https://github.com/getkirby/kirby/blob/v5/fix/6959-writer-update-issue/panel/src/components/Forms/Writer/Editor.js#L288

bastianallgeier:
Oh, this is super weird. You are right. But it seems that it somehow still propagates to the update event. Then I have to track down why it is not being detected there.

afbora:
OK. Also we can move the `payload.transaction.steps.length` check into `Editor.js::dispatchTransaction()` method to consistency.

bastianallgeier:
The update event seems to be triggered from somewhere else and I can't find it right now. It isn't enough to make that check in the dispatchTransaction method. Update still gets called. 

I've also made another change. We trigger an input event when the editor is being initialized. This seems very wrong to me. I think we did this earlier when the HTML from the backend would be different from what the writer would create as document. In that case the values between backend and changes in local storage would not be exactly the same and the changes button would appear. But I think we don't really need this anymore with the new setup. But we have to test it some more.

bastianallgeier:
I just tested it with invalid HTML on the backend and it works flawlessly. I really think that the combination of cleaned up events and the new changes architecture will finally help us tame the Writer monster.

afbora:
@bastianallgeier I think if condition is wrong, it should be like that: use `and` operator instead `or`. Just following changes works for me.

````js
// Only emit an update if the doc has changed and
// an update has not been actively prevented
if (
  transaction.docChanged &&
  !transaction.getMeta("preventUpdate") &&
  transaction.steps.length > 0
) {
  this.emit("update", payload);
}
````
https://github.com/getkirby/kirby/blob/v5/fix/6959-writer-update-issue/panel/src/components/Forms/Writer/Editor.js#L288

bastianallgeier:
This is embarrasing. You are totally right. The `||` broke the logic there 🙈 Works like a charm now.

afbora (approved):

PR #7002, opened 2025-02-13
New ExceptionField class

bastianallgeier:
## Description

This PR picks up the work on the Form classes again with a first small improvement to trim the logic in the Form class.

### Summary of changes

- New `KIrby\Form\Field\ExceptionField` to handle field errors
- Simplified logic in the Form class. 

### Breaking changes

- Removed `Kirby\Form\Form::exceptionField()`

## Ready?

### For review team

### DIFF
--- a/tests/Form/Field/ExceptionFieldTest.php
+++ b/tests/Form/Field/ExceptionFieldTest.php
@@ -15,6 +15,12 @@ class ExceptionFieldTest extends TestCase
 		$this->app = new App();
 	}
 
+	public function testIsSaveable(): void
+	{
+		$field = new ExceptionField('test', new Exception());
+		$this->assertTrue($field->isSaveable());
+	}
+
 	public function testLabel(): void
 	{
 		$field = new ExceptionField('test', new Exception());
--- a/tests/Form/Field/ExceptionFieldTest.php
+++ b/tests/Form/Field/ExceptionFieldTest.php
@@ -1,0 +1,99 @@
+<?php
+
+namespace Kirby\Form\Field;
+
+use Exception;
+use Kirby\Cms\App;
+use Kirby\TestCase;
+
+/**
+ * @coversDefaultClass \Kirby\Form\Field\ExceptionField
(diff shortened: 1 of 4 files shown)


afbora:
I think we should also mark as not saveable field like following:

````php
public function isSaveable(): bool
{
  return false;
}
````

File: tests/Form/Field/ExceptionFieldTest.php [hunk 1]
```diff
@@ -0,0 +1,99 @@
+<?php
+
+namespace Kirby\Form\Field;
+
+use Exception;
+use Kirby\Cms\App;
+use Kirby\TestCase;
+
+/**
+ * @coversDefaultClass \Kirby\Form\Field\ExceptionField
```
distantnative: `Suggestion:` I think we could start using PHPUnit attributes to shift away from the deprecated docblock annotations.

bastianallgeier: I just had a look at your PR to add annotations and it looks like there are some remaining issues there. At least we probably would already need to update CS fixer. Do you think it makes sense to combine it with this PR?

bastianallgeier: I tried it and it seems to work great without additional changes.

distantnative:
I think @afbora is right, that's what is missing from before when using the info field.

bastianallgeier:
Yep, I just added it.

distantnative (approved):

afbora:
@distantnative @bastianallgeier Am I missing something? `isSaveable` shouldn't return `false`? If it should return `true`, this method is unnecessary since parent method is same (`FieldClass::isSaveable()`).

distantnative:
@afbora ~~Maybe I am thinking  about it wrong, but why should the exception field be savable?~~ You are right it should be `false`.

bastianallgeier:
Sorry, this was completely wrong. Of course it should return false. I fixed it on the develop branch.

PR #7004, opened 2025-02-14
Add missing `reset` option for radio field

gitlabisbetterthangithub:
Should fix https://github.com/getkirby/getkirby.com/issues/2410

## Description
This is my attempt at fixing https://github.com/getkirby/getkirby.com/issues/2410. This might not be correct as I might have misunderstood how the field properties table is generated. Hopefully it's correct.

### DIFF
--- a/config/fields/radio.php
+++ b/config/fields/radio.php
@@ -17,6 +17,12 @@ return [
 		'columns' => function (int $columns = 1) {
 			return $columns;
 		},
+		/**
+		 * A radio button can be deactivated on click. If reset is `false` deactivating a radio button is no longer possible.
+		 */
+		'reset' => function (bool $reset = true) {
+			return $reset;
+		}
 	],
 	'computed' => [
 		'default' => function () {


afbora:
You're right. But AFAIK we also need to unset the `reset` prop from `select` field since `select` field inherit from `radio` props.

distantnative:
I am wondering about two things:
- Should a radio button be resettable? The native HTML element isn't.
- If it is resettable, should the default be `true` - I have doubts as this could be a larger breaking change for setups that have been build with them not been resettable.

bastianallgeier:
@distantnative I would agree with you here. We don't want them to be resettable in the first place in my opinion.

distantnative:
@bastianallgeier That said, it's quite explicitly implemented in `k-radio-input`. Do we want to remove it? I'd think it would pose a breaking change.

distantnative:
Closing in favor of https://github.com/getkirby/kirby/pull/7385

PR #7011, opened 2025-02-15
Support creating/editing array language variables with using new `entries` field

afbora:
## Description

v5-develop test_panel_languages_en

### Open question

- [ ] Decide Kirby release, 5.0? 5.1?
- [ ] Need help to improve labels and help texts (Would be great dynamic placeholder, but AFAIK not available for now such a feature)

### Features

- Now array language variables can be created/edited from panel

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -449,7 +449,10 @@
 	"language.variables.empty": "No translations yet",
 
 	"language.variable.delete.confirm": "Do you really want to delete the variable for {key}?",
+	"language.variable.entries": "Entries",
+	"language.variable.entries.help": "This strings will be used for translations count. Each entry corresponds to a different case counter. For example: 0, 1, 2 and more",
 	"language.variable.key": "Key",
+	"language.variable.multiple": "Multiple",
 	"language.variable.notFound": "The variable could not be found",
 	"language.variable.value": "Value",
 
--- a/panel/src/components/Dialogs/Elements/Fields.vue
+++ b/panel/src/components/Dialogs/Elements/Fields.vue
@@ -2,7 +2,7 @@
 	<k-fieldset
 		v-if="hasFields"
 		:fields="fields"
-		:value="value"
+		:value="formValue"
 		class="k-dialog-fields"
 		@input="$emit('input', $event)"
 		@submit="$emit('submit', $event)"
@@ -45,6 +45,12 @@ export default {
 	mixins: [props],
 	emits: ["input", "submit"],
 	computed: {
+		formValue() {
+			return {
+				...this.$helper.field.form(this.fields),
+				...this.value
+			};
+		},
 		hasFields() {
 			return this.$helper.object.length(this.fields) > 0;
 		}
--- a/config/areas/languages/dialogs.php
+++ b/config/areas/languages/dialogs.php
@@ -49,11 +49,29 @@ $translationDialogFields = [
 		'label'   => I18n::translate('language.variable.key'),
 		'type'    => 'text'
 	],
+	'multiple' => [
+		'label'   => I18n::translate('language.variable.multiple'),
+		'type'    => 'toggle',
+		'default' => false
+	],
 	'value' => [
 		'buttons' => false,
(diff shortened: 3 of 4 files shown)


i18n/translations/en.json
```diff
@@ -449,7 +449,10 @@
 	"language.variables.empty": "No translations yet",
 
 	"language.variable.delete.confirm": "Do you really want to delete the variable for {key}?",
+	"language.variable.entries": "Entries",
+	"language.variable.entries.help": "This strings will be used for translations count. Each entry corresponds to a different case counter. For example: 0, 1, 2 and more",
```
distantnative, line 453: I think this would be more fitting with the existing `language.variable.value`:

```suggestion
	"language.variable.values": "Values",
	"language.variable.values.help": "These values will be used according to the translation count. Each entry corresponds to a different case, e.g. for 0, for 1, for 2 and more.",
```

i18n/translations/en.json
```diff
@@ -449,7 +449,10 @@
...
 	"language.variable.key": "Key",
+	"language.variable.multiple": "Multiple",
```
distantnative, line 455:
```suggestion
	"language.variable.multiple": "Different strings based on count",
```

Not so happy with this either. But I think the `multiple` field needs more explanatory label and maybe also on/off texts.

panel/src/components/Dialogs/Elements/Fields.vue
```diff
@@ -45,6 +45,12 @@ export default {
 	mixins: [props],
 	emits: ["input", "submit"],
 	computed: {
+		formValue() {
+			return {
+				...this.$helper.field.form(this.fields),
```
distantnative, line 50: `Question:` Why is this needed?

afbora: Good question.

For this line:
https://github.com/getkirby/kirby/blob/lab/afbora/support-array-language-variables/config/areas/languages/dialogs.php#L55

Otherwise default value of `multiple` field is `null`. We have two condition fields `value` and `entries`. When `multiple` disabled `value` field visible and when enabled `entries` field visible. So when `multiple` value is `null`, no fields visible.

Not sure this is a correct solution.

i18n/translations/en.json
afbora: Absolutely. I'm not happy with the labels and help texts either. That's why I mentioned it in the PR description. It needs to be more descriptive and clear.

panel/src/components/Dialogs/Elements/Fields.vue
distantnative: What if we don't use `default` there but rather add here https://github.com/getkirby/kirby/blob/lab/afbora/support-array-language-variables/config/areas/languages/dialogs.php#L202-L205 something like

```php
'value' => [
  'multiple' => false
]
```

afbora: Yes, it seems to be working. 

What I'm wondering is; does this change I made really solve the problem of default values ​​in dialogs? Maybe we can use it in other dialogs or in the future.

If my solution is correct I would prefer the method where the fields work exactly as intended. Otherwise I'm fine with your solution 👍

afbora: @distantnative What do you think?

distantnative: To be honest, I am not sure if your change is what is expected for dialogs - or rather the previous behavior. To get the entries field enhancement in v5.0 I would vote for reverting the `this.$helper.field.form()` part. We can do this in another PR later as a general improvement. But to get this PR merged, it would help if we don't reach into other areas that broadly.

distantnative: @afbora I have forked your PR, implemented the initial value via the `value` prop and make some tweaks to i18n texts in this new PR: https://github.com/getkirby/kirby/pull/7173

Issue #7022, opened 2025-02-20
[v5] Image focus can't be removed


medienbaecker:
## Description
After setting a focus point to an image, you're no longer able to remove it by clicking the × button. The Discard/Save buttons appear, clicking save does however not save the change.

**Expected behavior**  
The focus should be empty again.

**Screenshots**  

![Image](https://github.com/user-attachments/assets/677744f0-887f-4fe0-bfe1-f715e006d836)

## To reproduce

1. Upload an image
2. Set a focus point
3. Save
4. Try to remove the focus point

## Your setup
**Kirby Version**  
5.0.0-beta.3

**Your system (please complete the following information)**
 - Browser: Chrome
 - Version: 133.0.6943.99

bastianallgeier:
We finally have a solution 🎉

PR #7028, opened 2025-02-24
Fix image focus removing

afbora:
### Fixes
- Image focus can't be removed #7022

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Views/Files/ImageFilePreview.vue
+++ b/panel/src/components/Views/Files/ImageFilePreview.vue
@@ -108,7 +108,7 @@ export default {
 			}
 
 			if (!focus) {
-				focus = null;
+				focus = "";
 			} else if (this.$helper.object.isObject(focus) === true) {
 				focus = `${focus.x.toFixed(1)}% ${focus.y.toFixed(1)}%`;
 			}


distantnative:
I also saw that this would fix this specific issue.

But I am wondering If this rather points to a general regression, where null/undefined values don't get updated anymore via our new content module. Where in the past they did. @bastianallgeier

bastianallgeier:
We should test this in the new holy merge branch

bastianallgeier:
I've added unit tests to the "Holy merge" branch to check if we can still remove fields properly by passing null and that works on the backend. If I remember correctly here, the problem was that null values are not properly sent to the server in the first place. I think we should look into that some more.

bastianallgeier:
Ok, I know now where this happens. When a field is unset in the changes version by passing null, the Form class merges it with the latest version and the missing field gets overwritten with the value from the latest version: 

https://github.com/getkirby/kirby/blob/6c4c378d90e57b7c91e8e05f222212ef560bd2de/src/Panel/Model.php#L49-L54

bastianallgeier:
I tracked it down further and it seems like the main issue is the Form::for part in ModelWithContent::update because that's another part where we merge the input with the latest version.

bastianallgeier:
I have a replacement for this PR which does also fix the underlying regression that null values can no longer remove fields: https://github.com/getkirby/kirby/pull/7116

Issue #7010, opened 2025-02-15
[v5] virtual content ignored when actual content file exists


distantnative:
Or even if just the folder exists (as Kirby will treat the folder alone as a content file with just the slug as title)

PR #7029, opened 2025-02-24
Holy merge

bastianallgeier:
### Enhancements

- `App::setSite()` is now public and can be used to overwrite the current site instance
- New `Page::normalizeProps()`
- New `File::noramlizeProps()`
- New `User::normalizeProps()`
- New `ModelWithContent::changeStorage()` method. 
- The MemoryStorage class uses the object id for guaranteed uniqueness
- A couple methods in Kirby\Content\Translation have been undeprecated because they turn out to be useful 
  - `::code()`
  - `::content()`
  - `::exists()`
- New virtual page tests
- Unit tests for Models have been reorganized and cleaned up

### Fixes

- https://github.com/getkirby/kirby/issues/7010

### Breaking changes

- When working with old model instances in hooks, the models can no longer overwrite their content. This will now throw an exception if you try. Use the new model instances instead.
- `$page::translation()` returns a single translation in single language mode. This is also included in the `$page::toArray()` method
- `$page::translation()` returns a new `Kirby\Content\Translation` object instead of the old `Kirby\Content\ContentTranslation` object. 
- The `Kirby\Content\ContentTranslation` class has been removed and repalced with `Kirby\Content\Translation`
- `$page::translations()` returns a new `Kirby\Content\Translations` collection instead of a generic `Kirby\Content\Collection` It will always include at least one language - even in single language mode.
- `Kirby\Content\Content::update()` is actively deprecated and must no longer be used. It will throw an exception to find usage early on.
- The `ModelWithContent::$content` and `ModelWithContent::$translations` cache props have been removed.
- `User::create()` does no longer accept null as argument. 
- Setting a name field in `User::update` does no longer have any effect. Use `User::changeName()` instead.


lukasbestle (commented):
As I said in today's meeting, the PR is super hard to review. Mainly because of its size of course (but not really avoidable), but also because the initial comment doesn't really explain *why* the changes were made and omits the blocks we normally have in the PR template. But since I cannot possibly review this in detail anyway (just give some input), it's fine for this. Just as a feedback for the future. :)

I started adding comments from what I noticed while reading through the diffs of the `src` files. I managed to read until half of `ModelWithContent` today. Note that I wasn't able to follow all changes and their purpose (especially model creation), so I only comment on what I can actually comment on.

Even Safari on my 24 GB M2 Mac mini completely hangs in the files view, so I'm adding my comments here as a list:

- Changing the order of `App` initialization is risky. We had multiple regressions in the past when we reordered it. But I don't think we can prepare for that as those regressions have always turned up in real world edge cases in the past. So I guess the solution is to keep this change in mind if any weird issues pop up.
- `$file->changeTemplate()`: Avoiding to call `update()` another time after the fields have been converted to the new blueprint makes sense. But calling `$version->save()` also feels like a hack to me. Maybe we could integrate this logic into `$file->convertTo()` so there is just one step in the first place?
- `$model->changeStorage()`: I would feel better if string values were checked before creating the instance. Perfect would be a check using reflection if the class actually exists and implements the `Storage` abstract, but I'm not sure if that's viable regarding performance.
- `changeStorage()` also says in the docblock that it "moves" the storage, but it actually copies, doesn't it?

bastianallgeier:
@lukasbestle 

- I've fixed the docblock for changeStorage 
- I had a look if we might be able to add checks for the storage class. As far as I can see, type hinting and PHP is helping us out here already. If the class does not exist, the instance can't be created. If the class is not an extension of Storage, the type check in the copyAll method will throw an error. I think both errors are helpful enough in a debugging scenario
- I've updated the Content::convertTo logic to also overwrite the template field. I then improved the File::contentFileData logic to make sure that we never send the default template value to the storage handlers. The additional save call in File::changeTemplate is now gone and I also added more unit tests for those cases.

lukasbestle:
- changeStorage: TBH I wonder if that's the right way around or if the method should really move the content. For memory handlers it doesn't really matter, but let's say I have a model in the filesystem and want to change its storage to database (not in v5 of course, but in the future). If we copy, the content data will exist twice afterwards.
- On the checks for instantiation: You are right that an invalid class name will blow up in one way or another and won't be unnoticed. I'm more concerned about attacks that could potentially cause remote code execution or similar vulnerabilities. Because the class is first instantiated and *then* checked for validity by PHP, one could execute the constructor of an arbitrary class. By that point, damage might already be done. Not easily exploitable, but still a risk.

bastianallgeier:
@lukasbestle 

- changeStorage only copies the content because we realized that we often really need to keep storage in two places for all our clones: In memory for the outdated instance and on disk for the fresh clone. But we might really want to make this clearer by changing the method name or introducing a flag which can switch between copying or moving.

bastianallgeier:
@lukasbestle I've just found out that the check for the correct class does not need Reflections at all: 

```php
if (is_subclass_of($toStorage, Storage::class) === false) {
    throw new InvalidArgumentException('Invalid storage class');
}
```

… works like a charm. I've added it together with a new unit test.

lukasbestle:
The `is_subclass_of()` solution is great!

About copy/move: Maybe we could have an enum like `StorageChangeMode` that would get values of `COPY` and `MOVE`. This could then optionally be passed in a second `$mode` argument. Maybe this is shooting a gun on sparrows, but it would certainly be quite clean.

PR #7035, opened 2025-02-26
feat: Plugin extension autoloader

plain-solutions-gmbh:
## Description

I have been working on a adequate solution for an plugin autoloader.

Here is an example:

```php
App::plugin(
    'vendor/plugin', autoloader: [
    	'cache' => true,
    	'classes' => [
    		'namespace' => 'Vendor\\Namespace'
    	],
    	'config' => [
    		'path' => './extends'
    	],
    	'blueprints', 'snippets', 'fields', 'translations',
    	'custom' => function($autoloader) {
	        $autoloader->merge([
	            "options" => [
	                "my_option" => 'foo'
	            ]
	        ]);
	    }
    ]
);

```

If true is passed to the autoloader, all tasks defined in the [autoloader](https://github.com/plain-solutions-gmbh/kirby/blob/511fbf0746921b96f7b7f61265d76469da63a338/src/Plugin/Autoloader.php#L35) are executed.

### Summary of changes
- Add Autoloader class
- implement to AppPlugin (not plugin class - cause info and root needs to be defined)
- Add Method to AppPlugin to get allowed extensions keys

### Reasoning
Plugins (if the developer use the autoloader) can be loaded from cache.
Extensions could be easly integrating into the plugin.

### Additional context
I create it because of this [rejected pull request](https://github.com/getkirby/kirby/pull/6714) 
It would be nice if it's goes into V5. Cause Plugin already takes a new parameter `license`

## Changelog
- The autoloader executed if the parameter 'autoload' is either 'true' or an array. (Disabled by default)
- The array contains tasks that are processed in sequence
- Tasks can be modify, enable or extend. (Cache is enabled by default)
- The result is cached and stored by default under `./site/cache/autoloder/{vendor}/{plugin_name}\{timestamp}.php`. (Can be deactivated)
- The plugin folder is modified. The cache is renewed.
- Loads the classes (by default from classes folder) automatically
- Fields, sections, blueprints, snippets, templates and translations are loaded from the root by default.
- For other extensions is the subfolder `./config` in charge.

## Docs
I could write a cookbook if the feature will be applied

## Ready?

### For review team


src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,566 @@
...
+        //Caching disabled
+        if ($this->cache === false) {
+            return;
+        }
+
+        //Clean from old caches
+        $cache_folder = pathinfo($this->cache_file, PATHINFO_DIRNAME);
+        try {
+            Dir::remove($cache_folder);
+        } catch (\Throwable $th) {};
```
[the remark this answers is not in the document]
plain-solutions-gmbh: Fixed in commit `Fix empty catch`.

bastianallgeier:
This looks very interesting. Give us a bit of time to review this, while we still struggle with the last v5 steps.

plain-solutions-gmbh:
Todo:
- use F::loadClasses to load classes.
- remove cache task and move them to tasks parameter

distantnative:
@plain-solutions-gmbh I finally had a chance to take a closer look again. Unfortunately, it's still quite far from being mergable. That's not to say it isn't functional, but the approach and code style differ greatly from other parts of Kirby. I'd be concerned about maintainability, etc.

It would need to be much less smart and automatic. For example, it shouldn't be a list of tasks that works with reflections. Rather, it should be a very explicit list of dedicated methods that deal with autoloading the different extensions. It shouldn't manipulate variables passed as arguments via references, but rather return values from methods and set them in the main context.

It wouldn't make sense to send you in circles working on this. That might end up being a frustrating experience, and we don't want that to be the result of such a well-intentioned code contribution. However, we also need to ensure the quality and style of the code we add to the core. We might eventually get to it ourselves, but it's not our highest priority at the moment.

plain-solutions-gmbh:
That’s sad to hear.

For over three years, I’ve been working exclusively with the Kirby CMS codebase, and I thought I had largely adapted its style of writing code.

I have to admit that I demanded a lot of flexibility from this feature and, as you mentioned, ended up building in a bit too much automation. As Bastian emphasized several times during the introduction of K5, a new feature often sparks ideas for further improvements — and narrowing those down is truly a challenge.

If you could spare a little time, I’d really appreciate it if you could leave me a few line-comments on this pr. That way, I can see where I’ve strayed furthest from Kirby’s code style and work on improving the quality of my code.

Best regard Roman

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,568 @@
...
+            'path'      => './templates',
+            'rootkey'   => 'templates',
+            'read'      => true
+        ],
+        'translations' => [
+            'method'    => 'flatWalker',
+            'path'      => './i18n',
+            'rootkey'   => 'translations'
+        ]
+    ];
```
distantnative: I think this is too abstract and cryptic - I understand the aim for flexibility, but I think in the long run this will work better if it is more verbose. Having actual methods for each of the extensions that runs its necessary logic. And one main method that calls each of these methods one by one as `$this->loadXYZ()` without the magic, without for loops etc.

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,568 @@
...
+use Kirby\Exception\Exception;
+use Kirby\Exception\InvalidArgumentException;
+use Kirby\Filesystem\F;
+use Kirby\Toolkit\Str;
+use ReflectionMethod;
+
+class Autoloader
+{
+
+    private ?string $cache_file;
```
distantnative: `Style:` We always write `|null` instead of `?` for nullable types.

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,568 @@
...
+     * @param string $root The root path of the plugin
+     * @param array $data Predefined extension data to extend
+     * @param array|bool $tasks A list of tasks to autoload (see $tasks property)
+     * @return void|$this 
+     */
+    public function __construct(
+        private string $name,
+        private string $root,
+        private array $data = [],
+        array|bool $tasks = true
```
distantnative: I am wondering if it would suit Kirby better if not each plugin defines their own setup and where to load what from but rather uses a very fixed setup that the autoloader enforces when using autoloading.

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,568 @@
...
+            $method = $task['method'];
+
+            //Call given user function
+            if ($method instanceof Closure) {
+                $method($this);
+                continue;
+            }            
+
+
+            $reflection = new ReflectionMethod($this, $method);
```
distantnative: For performance, it would be good if we can avoid reflections - rather keep it more explicit (even if verbose) by defining what methods to call.

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,568 @@
...
+        //Load resistant cache items to data
+        if ($this->cache_resistance = $cache['resistance'] ?? null) {
+            foreach ($this->cache_resistance as $file => $keychain) {
+                $this->setValueFromKeyChain($this->data, $keychain, Data::read($file));
+            };
+        };
+
+        //Indicate that cache is loadet
+        $this->cache = true;
+    }
```
distantnative: Could we not use any of the existing `Kirby\Cache` options available? Overall the caching logic really complicates this class currently. it would be good if we can reuse existing solutions -- and if it still requires a lot code to handle the caching logic, maybe this should be moved to its own class.

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,568 @@
...
+     * @param bool $read
+     * @return void 
+     */
+    public function filesWalker(
+        ?string $path = null,
+        ?Closure $callback = null,
+        bool $read = true
+    ): void {
+
+        $callback ??= fn($file) => $this->setValue(pathinfo($file, PATHINFO_FILENAME), $file, $read);
```
distantnative: Having these helper methods setting the global state can be hard to follow and understand as the flow of data doesn't loop back to the main logic route. It'd be better if helper methods returned the (processed) data but then the main method adds it to the cache or so.

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,568 @@
...
+    /**
+     * Walk throw the given directory recursively and pass the filename to the callback
+     * Check if files doesn't starts with '_'
+     * 
+     * @param null|string $path 
+     * @param null|Closure $callback 
+     * @param bool $read
+     * @return void 
+     */
+    public function filesWalker(
```
distantnative: These types are methods are maybe very flexible and smart, but they are difficult to understand and therefore hard to maintain in the long run with Kirby where our attention constantly needs to shift around the system and so 2-3 years down the road, we'll never encountering this class again as "newbies". I would really suggest more explicit, less magic methods, e.g. `loadBlueprints` that run their logic rather straightforward. There might be the use case for some helper methods surely, but what they do should be easy to understand from the calling method but their name and context - and make it easy to understand the flow of the code.

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,568 @@
...
+    private function registerClasses(?array $classes = null): void
+    {
+        if (is_array($classes)) {
+            $this->classes = $classes ?? [];
+        }
+
+        if (count($this->classes) > 0) {
+            spl_autoload_register([$this, 'autoload']);
+        };
+    }
```
distantnative: Most of this logic could be replaces by `F::loadClasses()`/ the `load()` helper.

src/Plugin/Plugin.php
```diff
@@ -44,6 +44,7 @@ public function __construct(
 		Closure|string|array|null $license = null,
 		protected string|null $root = null,
 		protected string|null $version = null,
+		protected bool|array $autoloader = false,
```
distantnative: I cannot fully overlook this now - but if possible, it would be great if we rather just have a `autoload` boolean flag. That would be very convenient. But there might be the need for some configs after all, so I cannot assess that yet. In general, I think some conventions and standards often go further than too many config options.

src/Cms/AppPlugins.php
```diff
@@ -925,6 +927,11 @@ public function plugins(array|null $plugins = null): array
 		return static::$plugins;
 	}
 
+	public function allowedExtensionsKeys(): array
+	{
+		return array_keys($this->extensions);
+	}
+
```
distantnative, line 949: We don't need to base the logic on the available extensions. It might seem like a good idea to make it dynamic based on this, but some verbosity will go a long way here for understanding and maintaining this (even with a bit more effort). Look at classes like `Kirby\Cms\Core`, `Kirby\Cms\AppPlugins` - writing code for each extension manually helps, even if that means we have to change the code in multiple classes when we add/remove an extension.

distantnative (commented):
I left some in-code comments. I think you are right that the biggest "problem" currently is that the code tries to be too flexible, too magic, too smart. We have all been there ourselves (and will be again now and then), but we have learned that this usually backfires down the road. Refactoring those tasks/usertasks and magic walker methods into very concrete methods for each extension would be an important step.

plain-solutions-gmbh:
Thank you for your detailed explanation.

I took the liberty of redesigning the autoloader based on your input. This time, I removed everything that made it confusing and slow. This means that caching is no longer necessary.

Now you can simply pass an extended autoloader class to the plugin and you're already extremely flexible.

src/Cms/AppPlugins.php
```diff
@@ -871,18 +872,35 @@ public static function plugin(
...
+		if ($extends === null && $autoloader === false) {
 			return static::$plugins[$name] ?? null;
 		}
 
+		$root ??= $extends['root'] ?? dirname(debug_backtrace()[0]['file']);
+
+		if ($autoloader) {
+
+			//Allow to apply custom Autoloader
+			$autolader_class = is_bool($autoloader) ? Autoloader::class : $autoloader;
```
distantnative, line 888: `Question:` What for would one use a custom autoloader class?

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,219 @@
...
+        public string $name,
+        public string $root,
+        public array $data = []
+    ) {
+
+        //Load classes before everything else
+        $classfolder = $root . '/classes/';
+        if (Dir::exists($classfolder)) {
+            $this->_classes($classfolder);
+        }
```
distantnative: I would move this to a `loadClasses()` method.

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,219 @@
...
+
+        foreach (Dir::dirs($this->root) as $dir) {
+            if (method_exists($this, $dir)) {
+                $this->$dir($this->root . '/' . $dir . '/');
+            }
+        }
```
distantnative: This should be less magic. We can simply list all potential extensions:

```php
$this->loadBlueprints();
$this->loadFields();
....
```

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,219 @@
...
+            }
+
+            $file = $root . $path;
+            if (F::exists($file)) {
+                $dirname = F::dirname($path);
+                $dirname = $dirname === '.' ? '' : $dirname . '/';
+                $fnc($dirname . F::name($path), $file);
+            }
+        }
+    }
```
distantnative: I think a less flexible and more explicit approach would be better. Reading the code, the `_dirWalker` method comes across quite cryptic and hard to understand. I also don't understand why many extensions would need to run the index vs. just `Dir::files()`. That's why IMO it would be better to leave it to each dedicated `load*()` method to use the appropriate code and not this meta method.

src/Cms/AppPlugins.php
```diff
@@ -871,18 +872,35 @@ public static function plugin(
...
+
+			$extends = $autolader_class::load(
+				name: $name,
+				root: $root,
+				data: $extends ?? []
```
distantnative: `Suggestion:` Instead of passing `$extends` to the class and merging it in the class's `::toArray()` method, the class could be responsible to only gather and return autoloader extensions. And then we rather merge the explicitly passed `$extends` and the result of `Autoloader::toArray()` here.

src/Plugin/Autoloader.php
```diff
@@ -0,0 +1,219 @@
...
+     */
+    public function blueprints(string $root)
+    {
+        $this->_dirWalker($root, function ($path, $file) {
+
+            $name = F::relativepath($file);
+            //Set YAML-File or they content
+            $this->extends['blueprints'][$path] = (F::extension($file) === 'yml') ? $file : $this->_read($file);
+        });
+    }
```
distantnative: As an example, I think this could be refactored more in this direction:

```suggestion
    public function loadBlueprints(): void
    {
        $root = $this->root . '/blueprints';
        
        foreach (Dir::index($root, true) as $blueprint) {
            $file = $root . '/' . $blueprint;
            
            if (F::exists($file) === false) {
                continue;
            }
            
            $this->extends['blueprints'][$blueprint] =  $$file;
        }
    }
```

(Just writing this on GitHub so code is probably not fully right, but to communicate the general idea)

distantnative (commented):
I think this was a good step in the right direction!

I would encourage you to even less think of flexibility. Simplicity might be the more important first goal. Once we have a simple, solid, working solution, we could still see where to make it more flexible. 

One thought that came up while reviewing and I just want to share it, might not work in the end. But instead of defining a separate autoload option in the plugin registration call, couldn't we use the logic of e.g. for blueprints "if the blueprint extend is an array, use it as before. If it is a string, assume it's a directory and automatically autoload it"?

src/Cms/AppPlugins.php
plain-solutions-gmbh: For example: In one of my cases i used it to extend the `$this->extend()` with data from a Class (Which is loading in the autoloader). So i made a custom Autoloader class and modified the `toArray()` method.
An other case could be, that you need to load snippets from another folder than snippets.

src/Plugin/Autoloader.php
plain-solutions-gmbh: That was my initial plan. However, this makes the entire file very repetitive. I use index so that I can include all subfolders at once.

src/Plugin/Autoloader.php
plain-solutions-gmbh: A good point. Would that mean that an array containing all specific folders with methods? For example, with:

```
$folders = [
    'snippets' => 'loadSnippets'
    ...
];

foreach (Dir::dirs($this->root) as $dir) {
    if (array_key_exists($dir, $folders) === false) {
        $method = $folders[$dir];
        $this->$method($this->root . '/' . $dir . '/');
    }
}
´´´

distantnative: I would not try to make it that smart/magic. I know it seems very verbose and repetitive, but from our experience "boring" code like that has often paid off down the road.

plain-solutions-gmbh:
Alright. Looks nicer now.
Can you tell me why the tests where failing? Is there an report to check?

plain-solutions-gmbh:
> One thought that came up while reviewing and I just want to share it, might not work in the end. But instead of defining a separate autoload option in the plugin registration call, couldn't we use the logic of e.g. for blueprints "if the blueprint extend is an array, use it as before. If it is a string, assume it's a directory and automatically autoload it"?

That thought occurred to me as well. Developers typically select the 'Blueprint' folder to store blueprints, etc.
If only one definition is affected, that makes sense. However, when it comes to outsourcing api and areas, etc., writing a definition for each of them would make the autoloader obsolete.

For me, the development of the autoloader is also about making it easier for developers to get started with plugin development. Writing definitions that are self-explanatory would be an unnecessary barrier.

bastianallgeier:
I think it's only fair to close this PR as we won't be able to include it like it is. But I would still like to thank you for your time and ideas!!

PR #7036, opened 2025-02-26
Fix picker list layout style when long info text available

### LINKED ISSUE #6904: [v5] Ensure same listing style (breaking or not) in list view
CleanShot 2025-01-15 at 00 08 43@2x

Might be hard to solve, but this looks really off to me and makes it hard to read.

v5 beta.1

afbora:
## Description

In small dialogs (max 30rem width), if there is info text, it always shows as two lines. Thanks to @bastianallgeier giving solutions and ideas.

v5-develop test_panel_pages_notes+exploring-the-universe (1)

### Fixes
- Ensure same listing style (breaking or not) in list view  #6904

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Collection/Item.vue
+++ b/panel/src/components/Collection/Item.vue
@@ -267,8 +267,8 @@ export default {
 	text-overflow: ellipsis;
 }
 /** Provides a consistent look when texts are long in small dialogs */
-@container (max-width: 30rem) {
-	.k-dialog .k-item[data-layout="list"] .k-item-content:has(.k-item-info) {
+@container (max-width: 25rem) {
+	.k-item[data-layout="list"] .k-item-content:has(.k-item-info) {
 		flex-direction: column;
 	}
 }
--- a/panel/src/components/Collection/Item.vue
+++ b/panel/src/components/Collection/Item.vue
@@ -266,7 +266,12 @@ export default {
 	white-space: nowrap;
 	text-overflow: ellipsis;
 }
-
+/** Provides a consistent look when texts are long in small dialogs */
+@container (max-width: 30rem) {
+	.k-dialog .k-item[data-layout="list"] .k-item-content:has(.k-item-info) {
+		flex-direction: column;
+	}
+}
 .k-item[data-layout="list"] .k-sort-button {
 	--button-width: calc(1.5rem + var(--spacing-1));
 	--button-height: var(--item-height);


distantnative (changes requested):
With this change, the title and info texts don't wrap nicely anymore, unfortunately. Even outside picker dialogs. It reverts the improvements we've made from v4 to v5.

You can test it in the lab, e.g. https://sandbox.test/panel/lab/components/item/1_list the last example and compare the `v5/develop` behavior and this PR.

afbora:
@distantnative I understood. So we can solve this checking `.k-dialog .k-item ... {}` and move this logic to `PagesDialog.vue`. Make sense for you?

afbora:
I added a check whether it is in the dialog or not. What do you think?

bastianallgeier:
I talked to Ahmet about it and the main objective for this PR is to create consistency and avoid the following scenario …

Screenshot 2025-02-27 at 10 19 09

As far as we can see, there's no chance to find out if one of the items needs to wrap and then wrap all the others too. 

This solution is basically a compromise. It only applies when there's an info line and it only applies when the container width is already pretty narrow. I think the solution is pretty good and I would also vote to not just use it in the dialogs. 

Screenshot 2025-02-27 at 10 19 25

distantnative:
> This solution is basically a compromise. 

It's not really, in my view. Texts now don't wrap anymore and we're back at the v4 situation. While I totally agree that the v5 version doesn't look great in these cases where not all items do the same, I think it's overall still better to have a design problem than a problem with functionality (long texts overly often being truncated and not being visible).

Limiting it to dialogs could make sense to me. But as you point out will not fix all the cases with this design problem. But applying it everywhere seems like a step back to me, not an improvement.

Edit: To illustrate my argument, I think we shouldn't go back to this just because it overall look prettier:
Screenshot 2025-02-27 at 10 34 15

bastianallgeier:
I totally understand your point. I would counter with this by a famous IT brand from California 😂 …

Screenshot 2025-02-27 at 10 45 38

But of course that's not a good argument to make here. I think you are right about form follows function in this case. But then I would really vote for increasing the container width to a point where truncation becomes an edge case and that edge case is then still covered by our hover titles. 

Screenshot 2025-02-27 at 10 47 18

The differently wrapping items are in my opinion also a functional issue because they make scanning the list a lot harder. I think the two-line list looks pretty nice and if there's info text we could really just say that it needs to wrap quite early.

bastianallgeier:
Or another idea. We have size for cards and cardlets. Maybe the list layout needs a wrapping option. 

```
pages: 
  layout: list
  info: Some text
  wrap: true  
```

bastianallgeier:
@distantnative we need a decision about this for the final release. I still feel like the solution from the PR is maybe not 100% great, but better than the inconsistent version.

bastianallgeier (approved):

PR #7037, opened 2025-02-27
Cancel running requests when discarding or publishing content

bastianallgeier:
## Description

This PR makes sure that discard and publish requests cancel ongoing or schedule save requests properly, to avoid race conditions.

### Summary of changes

- The method to cancel throttled functions
- New `panel.content.cancelSaving()` method
- `panel.content.discard()` and `panel.content.publish()` cancel any ongoing or scheduled save requests. They also no longer can be executed for other views than the current one. 
- `panel.content.save()` no longer sets the `isProcessing` flag. Otherwise, discard and publish will be blocked. 

## Ready?

### For review team

### DIFF
--- a/panel/src/helpers/throttle.js
+++ b/panel/src/helpers/throttle.js
@@ -15,7 +15,7 @@ export default (
 	let last = null;
 	let trailing = null;
 
-	return function (...args) {
+	function throttled(...args) {
 		if (timer) {
 			last = this;
 			trailing = args;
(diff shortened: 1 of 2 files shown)


afbora:
If I'm missing nothing, works great for me 🎉

bastianallgeier:
I think we can never know 100% if we miss something until more real-life tests are made during the beta. But it's great to hear that you also don't run into further issues.

PR #7040, opened 2025-03-02
Hide entry label for entries field as default

### LINKED ISSUE #7039: [v5] Entries Field breaks when field has label
## Description

When adding a `label` to the `field` option of the new `type: entries` field, the UI breaks and shows the label, leading to overflowing rows.

Image

```yaml
services:
    type: entries
    label: Am Haus
    width: 1/2
    field:
      type: text
      label: Service
      placeholder: z.B. Tapezieren
```

## To reproduce

1. Create an entries field like above with a label
2. Add one or more entities

## Your setup
**Kirby Version**  

5.0.0-beta.3

**Console output**  

 - Device:  MacBook Pro
 - OS:  MacOS 15.2
 - Browser:  Arc (Chrome)
 - Version:  1.83.0

afbora:
### Fixes
- Entries Field breaks when field has label #7039

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Field/EntriesField.vue
+++ b/panel/src/components/Forms/Field/EntriesField.vue
@@ -64,6 +64,7 @@
 						:is="`k-${field.type}-field`"
 						:ref="'entry-' + index + '-input'"
 						:disabled="disabled"
+						:label="false"
 						:value="entry.value"
 						v-bind="field"
 						class="k-entries-field-item-input"
--- a/panel/src/components/Forms/Field/EntriesField.vue
+++ b/panel/src/components/Forms/Field/EntriesField.vue
@@ -64,9 +64,9 @@
 						:is="`k-${field.type}-field`"
 						:ref="'entry-' + index + '-input'"
 						:disabled="disabled"
-						:label="false"
 						:value="entry.value"
 						v-bind="field"
+						:label="false"
 						class="k-entries-field-item-input"
 						@input="onInput(index, $event)"
 					/>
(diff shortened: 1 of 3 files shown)


distantnative:
Hi @afbora, I have three two thoughts:
- Maybe it would be even better to already prevent the field props that we don't want to support on the backend, e.g. I'd think in `EntriesField::fieldProps()`: https://github.com/getkirby/kirby/blob/v5/develop/src/Form/Field/EntriesField.php#L49 (or in addition to blocking them in the Vue file too)
- Vue 3 has a breaking change where the position of `v-bind` will start to matter. Any prop that we want to overwrite what's applied by `v-bind` should follow after `v-bind`, not before.

afbora:
Thank you @distantnative for the reviews. How about now? https://github.com/getkirby/kirby/pull/7040/commits/d61514db2fbc8572a1a077ce2ec5a9c192a6941f

distantnative:
Looks good. Are there other options we should block as well? Maybe `counter`?

afbora:
Like that https://github.com/getkirby/kirby/pull/7040/commits/0b0bc45f9c6fb6c121b3beffd4620dbb01e70b4c?

distantnative:
I think this is good like this. I am only wondering if there are more options we should prevent. But also would be ok with merging this like it is now.

bastianallgeier (approved):
For the field types we support so far, this looks like the perfect solution. We need to be careful when we support more types later. Fields that have a button right next to the label could become another problem. Maybe we also want to hide the field header entirely. But I think for now, this is great.

PR #7050, opened 2025-03-11
Holy merge cleanup

bastianallgeier:
## Description

This is cleaning up some of the points that @lukasbestle mentioned in Notion. The changes are minor and mostly affect the docblocks. That's why I kept the PR description simple.

### DIFF
--- a/src/Content/Content.php
+++ b/src/Content/Content.php
@@ -237,10 +237,10 @@ class Content
 	}
 
 	/**
-	 * Updates the content and returns
-	 * a cloned object
-	 *
-	 * @deprecated 5.0.0 Use $model->version()->update() instead.
+	 * Removed in 5.0.0. We only keep it to simplify debugging, 
+	 * when this method is still in use in the wild.
+	 * 
+	 * Use $model->version()->update() instead.
 	 */
 	public function update(
 		array|null $content = null,
--- a/src/Content/Content.php
+++ b/src/Content/Content.php
@@ -246,6 +246,6 @@ class Content
 		array|null $content = null,
 		bool $overwrite = false
 	): static {
-		throw new Exception('This is no longer functional. Please use `$model->version()->update()` instead');
+		throw new Exception('`$content->update()` is no longer functional. Please use `$model->version()->update()` instead');
 	}
 }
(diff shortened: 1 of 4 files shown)


src/Content/Content.php
```diff
@@ -237,15 +237,15 @@ public function toArray(): array
...
 	 *
-	 * @deprecated 5.0.0 Use $model->version()->update() instead.
+	 * Use $model->version()->update() instead.
 	 */
 	public function update(
 		array|null $content = null,
 		bool $overwrite = false
 	): static {
-		throw new Exception('`$content->update()` has been deprecated. Please use `$model->version()->update()` instead');
+		throw new Exception('This is no longer functional. Please use `$model->version()->update()` instead');
```
distantnative: If it's not functional anymore anyways, could we just remove it?

bastianallgeier: The problem is our magic caller. When we remove the method, it would silently fail by returning an empty "update" field.

lukasbestle: To ease debugging:

```suggestion
		throw new Exception('`$content->update()` is no longer functional. Please use `$model->version()->update()` instead');
```

src/Content/Content.php
```diff
@@ -237,15 +237,15 @@ public function toArray(): array
 	}
 
 	/**
-	 * Updates the content and returns
-	 * a cloned object
+	 * Removed in 5.0.0. We only keep it to simplify debugging,
```
lukasbestle, line 240: We could still add a `@todo` comment to remove this in v7.

tests/Content/ContentTest.php
```diff
@@ -198,7 +198,7 @@ public function testUpdate()
 		]);
 
 		$this->expectException(Exception::class);
-		$this->expectExceptionMessage('`$content->update()` has been deprecated. Please use `$model->version()->update()` instead');
+		$this->expectExceptionMessage('This is no longer functional. Please use `$model->version()->update()` instead');
```
lukasbestle:
```suggestion
		$this->expectExceptionMessage('`$content->update()` is no longer functional. Please use `$model->version()->update()` instead');
```

lukasbestle (approved):

PR #7057, opened 2025-03-11
Clean up delete actions

bastianallgeier:
### Merge first

- [x] https://github.com/getkirby/kirby/pull/7038

### Summary of changes

- New wildcard option for `Version::delete()`
- Move the folder clean up logic into the PlainTextStorage handler
- New unit tests 
- Refactor`User::delete()`, `Page::delete()` and `File::delete()` to delete the Versions properly and let the PlainTextStorage handler take care of cleaning up the disk. 

### Reasoning

- The delete actions were completely buggy so far for virtual pages and it makes sense in my opinion to also refactor the other model delete actions at the same time. 

### Fixes

- `User::delete()` will now properly call `file.delete` hooks for deleted user files. 
- Versions will be properly deleted for all storage handlers. 

## Ready?

### For review team

### DIFF
--- a/src/Cms/FileActions.php
+++ b/src/Cms/FileActions.php
@@ -285,13 +285,16 @@ trait FileActions
 	public function delete(): bool
 	{
 		return $this->commit('delete', ['file' => $this], function ($file) {
-			// remove all public versions, lock and clear UUID cache
+			// remove all public versions and clear the UUID cache
 			$file->unpublish();
 
-			foreach ($file->storage()->all() as $version => $lang) {
-				$file->storage()->delete($version, $lang);
-			}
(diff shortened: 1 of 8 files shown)


distantnative:
It would be great if we had something like `$model->version('*')` too then. Not having to hardcode in so many places the two specific versions when it should actually always apply to all versions.

```php
foreach ($model->version('*') as $version) {
    $version->delete('*');
}
```

distantnative (approved):

bastianallgeier:
@distantnative I think it should be `$model->versions()` It does not feel good to me to return either a Version or a collection of Versions from the same method. But we can definitely take care of that in another PR.

PR #7059, opened 2025-03-11
`$model->versions()`

distantnative:
### Summary of changes
- New `$model->versions()` method and `Versions` collection class
- Using them in model actions to delete all versions
- Also using them in `Storage::all()`
- Removed `VersionId::all()` in turn

### Reasoning
Allows us to write code that is agnostic of the number of versions/the actual specific versions when we just mean "do this with all versions".

### Additional context
`Storage:all()` remains not great but I didn't want to follow down that rabbit hole.

### Enhancements
- New `$model->versions()` method returning a versions collection (currently changes and latest)

### Breaking

- `ModelWithContent::versions()` is no longer available as direct field accessor. You need to use `$model->content()->versions()` instead if you are using a field with the name `versions`.

## Ready?

### For review team


src/Cms/FileActions.php
```diff
@@ -288,11 +288,10 @@ public function delete(): bool
...
 
-			// delete all changes first
-			$file->version('changes')->delete('*');
-
-			// delete all latest versions as last step.
-			$file->version('latest')->delete('*');
+			// delete all versions
+			foreach ($file->versions() as $version) {
+				$version->delete('*');
+			}
```
lukasbestle: `Issue`: These changes in the delete actions make the version order implicit. Nothing would prevent us from breaking this order by changing it in `Versions::load()` by accident for some other feature or fix.

If the order is indeed required as stated in the previous code comments, here are some ideas:

- Add a param for `$model->versions()` that controls the order
- Add a code comment to `Versions::load()` and an unit test that both ensure that `changes` is always first, pointing to the `delete` actions that need it

distantnative: I think I would go for a comment in the versions class. I don't think it's something specific of the delete method that would be better dealt with a parameter. It's always the case that the latest version is the umbrella one so to say and should go last. Likely also in a future where there might be revisions too.

bastianallgeier: So far, the unit tests should also protect us, right?

src/Cms/PageActions.php
```diff
@@ -561,13 +561,12 @@ public function delete(bool $force = false): bool
...
-
-			// delete all latest versions as last step.
+			// delete all versions,
 			// the plain text storage handler will then clean
-			// up the directory if it's empty.
-			$page->version('latest')->delete('*');
+			// up the directory if it's empty
+			foreach ($page->versions() as $version) {
+				$version->delete('*');
+			}
```
bastianallgeier: Since Versions is already a custom collection, it would be quite cool to implement: 

`$page->versions()->delete()`

src/Cms/FileActions.php
distantnative: Yes. I guess adding a comment both on the method and in the unit test to explain that the order is important should be sufficient.

bastianallgeier: ☑️

src/Cms/PageActions.php
bastianallgeier: ☑️

bastianallgeier (approved):

PR #7060, opened 2025-03-12
Moving Storage instead of copying, whenever possible

bastianallgeier:
## Description

So far, we copy the storage when we create clones. But this creates redundant storage instances that are not useful. (see @lukasbestle's comment on Notion) 

The `ModelWithContent::changeStorage()` method does now move storage by default, but has a new flag to copy instead if needed. We still need to copy storage when we create detached in-memory clones for our hooks. This happens in the `ModelWithContent::save()` method and in the `ModelWithContent:convertTo()` method. 

When changing this at first, a couple tests broke until I realized that we had quite a big logical issue in our `Storage::move()` and `Storage::copy()` methods. Both methods did move or copy, even if the storage location actually did not change. This is super bad for memory consumption but also for the disk usage. We basically deleted files and recreated them or added new in-memory cache entries every time. But there was an even bigger issue in the PlainTextStorage handler. Consider this code: 

```php

$language = Language::ensure('en');
$version = VersionId::latest();

$handler1 = new PlainTextStorage($page);
$handler2 = new PlainTextStorage($page);

$handler1->move(
  fromVersionId: $version,
  fromLanguage: $language,
  toVersionId: $version,
  toLanguage: $language,
  toStorage: $handler2
);
```

The move logic looks like this: 

```php
// copy content to new version
$this->copy(
  $fromVersionId,
  $fromLanguage,
  $toVersionId,
  $toLanguage,
  $toStorage
);

// clean up the old version
$this->delete($fromVersionId, $fromLanguage);
```

This makes perfect sense when we have unique locations for both handlers. In memory, we have unique cache instances for example. But in PlainTextStorage both handlers point to the exact same location on disk. E.g. `some-page/article.en.txt`

This means that we would copy the content from the same file to the same file and then afterwards delete that same file 🙈

I think I have a really stable fix for this now. 

### Summary of changes

- New `Language::is()` method to compare language objects. This is pretty much idential to what we do in `Page::is()` and useful in other places
- New `Storage::isSameStorageLocation()`, which is now used in `Storage::move()` and `Storage::copy()` to return early if the storage location did not change. 
- New `PlainTextStorage::isSameStorageLocation()` implementation, which compares text file locations for more accurate results. 
- `ModelWithContent::changeStorage()` has a new second, boolean `$copy` argument, which is false by default. Storage is thus moved by default, but can be copied on demand.
- `ModelWithContent::save()` and `ModelWithContent::convertTo()` use that new `$copy` flag to create a fully in-memory copy for the old instance. 

## Ready?

### For review team

### DIFF
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -308,6 +308,7 @@ class Language implements Stringable
 	/**
 	 * Checks if the language is the same
 	 * as the given language or language code
+	 * @since 5.0.0
 	 */
 	public function is(self|string $language): bool
 	{
--- a/src/Cms/Language.php
+++ b/src/Cms/Language.php
@@ -306,6 +306,15 @@ class Language implements Stringable
 	}
 
 	/**
+	 * Checks if the language is the same
+	 * as the given language or language code
+	 */
+	public function is(self|string $language): bool
+	{
+		return $this->code() === static::ensure($language)->code();
+	}
+
+	/**
 	 * Checks if this is the default language
 	 * for the site.
 	 */
--- a/src/Cms/ModelWithContent.php
+++ b/src/Cms/ModelWithContent.php
@@ -104,11 +104,11 @@ abstract class ModelWithContent implements Identifiable, Stringable
 	}
 
 	/**
-	 * Copies the model to a new storage instance/type
+	 * Moves or copies the model to a new storage instance/type
 	 * @since 5.0.0
 	 * @internal
 	 */
-	public function changeStorage(Storage|string $toStorage): static
+	public function changeStorage(Storage|string $toStorage, bool $copy = false): static
 	{
 		if (is_string($toStorage) === true) {
 			if (is_subclass_of($toStorage, Storage::class) === false) {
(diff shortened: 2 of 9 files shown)


src/Cms/Language.php
```diff
@@ -305,6 +305,15 @@ public function exists(): bool
 		return file_exists($this->root());
 	}
 
+	/**
+	 * Checks if the language is the same
+	 * as the given language or language code
```
distantnative, line 310:
```suggestion
	 * as the given language or language code
	 * @since 5.0.0
```

src/Cms/ModelWithContent.php
```diff
@@ -443,7 +445,10 @@ public function save(
 		$clone = $this->clone();
 
 		// move the old model into memory
-		$this->changeStorage(ImmutableMemoryStorage::class);
+		$this->changeStorage(
+			toStorage: ImmutableMemoryStorage::class,
+			copy: true
+		);
```
distantnative, line 451:
```suggestion
		$this->changeStorage(ImmutableMemoryStorage::class, copy: true);
```

bastianallgeier: I would ignore this one. It's going to change in the next PR anyway when we need to pass the clone. :)

distantnative (approved):

PR #7067, opened 2025-03-15
Rename & refactor `Cms\Hooks` as `Cms\Events`

distantnative:
### Summary of changes
- Renamed `Cms\Hooks` to `Cms\Events` to match `Cms\Event`
- Refactored
  - If `hooks` isn't passed to `Events` but `bind` it uses this to get the hooks extensions
  - DRY-ed `Events::apply()` and `Events::trigger()` by using shared `Events::process`() 
  - Simplified `Events::hooks()`
- Removed custom event parameter from `App::apply()` and `App::trigger()`

### Reasoning
- In a few places we currently distinguishing between event and hook in the code, renaming `Cms\Event` to `Cms\Hook` made that code a bit messy, so I went with renaming it to `Cms\Events`
- Removed the custom events parameter as this was for internal use only and we only used it to re-call these methods for name wildcards.


lukasbestle (approved):
Looks great, I think this is now as clean as it gets given the constraints of the loop protection. 👍

PR #7082, opened 2025-03-20
refact!: Move field methods to `Content\Field`

### LINKED ISSUE #7078: Code completion for field methods
## The problem

The `Kirby\Content\Field` class is a type returned by common templating methods such as `$page->content()->get('field_name')`.

This object supports [56 methods documented here](https://getkirby.com/docs/reference/templates/field-methods).

But in an IDE, only a few methods (10) are visible, because only 10 methods are statistically declared by the `Field` class ([generated docs](https://getkirby.com/docs/reference/objects/content/field)). The other 46 are created here instead:
https://github.com/getkirby/kirby/blob/5.0.0-beta.4/config/methods.php

This means that most helper methods one might want to use with fields show an error in an IDE with any PHP language server. For instance, in Zed with the [phpactor](https://github.com/phpactor/phpactor) plugin:

Screenshot of the Zed editor with a $field variable set to a Kirby\Content\Field object, then a couple method calls like exists() and isNotEmpty() showing no error, and a couple others like toBlocks() and kirbytext() being underlined in red. An error tooltip says: phpactor: Method 'toBlocks' does not exist on class 'Kirby\Content\Field'

This makes it needlessly hard to work with content fields in PHP code, since you have to rely on the documentation exclusively.

## Expected outcome

When working with a `Kirby\Content\Field` instance, all utility methods listed in the docs should be visible in the IDE's code completion.

Out of scope: field methods [provided by extensions](https://getkirby.com/docs/reference/plugins/extensions/field-methods) would not be listed.

## Possible solutions

1. Refactor `config/methods.php` to move all the built-in methods directly to the `Kirby\Content\Field` class.
2. Add 40+ [PHPDoc `@method` tags](https://docs.phpdoc.org/guide/references/phpdoc/tags/method.html) to the `Kirby\Content\Field` class.
3. Declare all the built-in methods as public methods of `Kirby\Content\Field`, but call the current implementation:

```php
// src/Content/Field.php

class Field implements Stringable
{
	// ...

	/**
	 * Splits the field content into an array
	 */
	public function split(string $separator = ','): array
	{
		return $this->__call('split', [$separator]);
	}

}
```

The first option would be ideal, but is a bigger refactoring which might have some technical challenges and more technical risk.

The second option has no risk of breakage, but risks getting out of sync with the actual implementation. Because it uses the PHPDoc `@method` tag, it's a bit limited in what information it can convey (vs what a dedicated PHPDoc comment could contain).

The third option is a variant of the second one, and has much of the same tradeoffs. It's a bit nicer to author, and the IDE's jump-to-implementation is just a tad nicer.

distantnative:
## Description

- Moved all field methods from `config/methods.php` into `Kirby\Content\FieldMethods` trait
- Removed `Kirby\Content\Field::$aliases` array and instead added aliases as actual methods that call the main methods
- Callback passed to `Field::value($value)` is now bound to the cloned new field object
- Merged unit test classes and sorted them alphabetically 
- Removed  `Core::fieldMethods()` and `Core::fieldMethodsAliases()`
- Calling an unknown field method now throws an exception

Allows IDEs etc. to actually "understand" field methods, their parameter and return type hints etc.

### Refactored
- Moved default field methods into new `Kirby\Content\FieldMethods` trait use by `Kirby\Content\Field`

### Breaking changes
- Calling non-existing field methods is throwing a `Kirby\Exception\BadMethodCallException`
- Removed `Kirby\Content\Field::$aliases`
- Removed `Kirby\Cms\Core::fieldMethods()` and `Kirby\Cms\Core::fieldMethodsAliases()`

## Docs

[Cookbook article ](https://getkirby.com/docs/cookbook/content-structure/use-placeholders#custom-field-method) defines custom `replace` method even though this is in the core. Change?

## Ready?

### For review team


fvsch:
<details>
<summary>Not relevant, comment was based on incorrect assumption</summary>

One behavior change not listed in the TODO section for this PR:

Currently, method access with `__call` is case-insensitive:

1. All `'fieldMethods'` are registered with their name converted to lowercase (including the built-in ones, so `'isValid'` is documented as `$field->isValid(…)` but registered with the name `'isvalid'`; and an extension could override it by defining `'isValid'`, `'isvalid'`, `'ISVALID'`, etc.).
2. Then they are accessed through `__call` with the method name also converted to lowercase.

This means that currently it's possible to write `$field->kirbytext()` or `$field->kirbyText()` (or `$field->KirbyTEXT()`, etc.) and get the same result, but with this change users will have to make sure to use `$field->kirbytext()`, which is a breaking change.

A way to keep this backwards-compatible would be to have a mapping of public methods, then in `__call` to check if one of those needs to be called (NB: pseudoish code, I don't write a lot of PHP so some syntax may be off):

```php
class Field
{
	private static $builtinMethods = [
		'isvalid' => 'isValid',
		'kirbytext' => 'kirbytext',
		// …
	];

	public function __call(string $method, array $arguments = []): mixed
	{
		$method = strtolower($method);

		if ($this->hasMethod($method) === true) {
			return $this->callMethod($method, [clone $this, ...$arguments]);
		} else if (isset(static::$builtinMethods[$method])) {
			return $this->{static::$builtinMethods[$method]}(...$arguments);
		} else {
			// TODO: throw deprecation, then exception
			// when unknown method is called
		}

		return $this;
	}
}
```

(Alternatively, some of that logic — or something a bit more elaborate — could move to the `HasMethods` trait, if it's used in other classes which would need similar capabilities.)

</details>

bastianallgeier:
@fvsch this is not needed. PHP Methods are case insensitive by default.

fvsch:
@bastianallgeier Ha! I did not know that. 😅

tests/Toolkit/VTest.php
```diff
@@ -600,18 +600,18 @@ public function testSize(): void
...
-		$this->assertTrue(V::size($field->foo(), 2));
-		$this->assertTrue(V::size($field->foo(), 3, '<'));
-		$this->assertTrue(V::size($field->foo(), 1, '>'));
-		$this->assertFalse(V::size($field->foo(), 1, '<'));
-		$this->assertFalse(V::size($field->foo(), 3, '>'));
+		$this->assertTrue(V::size($field, 2));
+		$this->assertTrue(V::size($field, 3, '<'));
+		$this->assertTrue(V::size($field, 1, '>'));
+		$this->assertFalse(V::size($field, 1, '<'));
+		$this->assertFalse(V::size($field, 3, '>'));
```
distantnative, line 607: `Explanation:` I corrected these, as calling the non-existent field method `$field->foo()` would just return `$field` anyways. `V::size()`  itself takes care of applying the `->value()` field method.

config/methods.php
```diff
@@ -1,635 +0,0 @@
...
-	return [
-
-		// states
-
-		/**
-		 * Converts the field value into a proper boolean and inverts it
-		 */
-		'isFalse' => function (Field $field): bool {
-			return $field->toBool() === false;
-		},
```
distantnative, line 43: `Comment:` Alternatively to deleting the file we could add some legacy code to ease breaking changes where users require the config file to extend certain field methods. But not sure if this really helps or just delays the pain.

```php
'isFalse' => fn (Field $field) => $field->isFalse()
```

bastianallgeier:
Finally coming back to this and my first thought was that we could move all the field methods into a FieldMethods trait or maybe into multiple ones by group. I think having them all in a single file is quite overwhelming when we have to deal with core field logic.

distantnative:
@bastianallgeier Moved them into a `FieldMethod` trait.

src/Content/Field.php
```diff
@@ -138,31 +81,20 @@ public function key(): string
 	}
 
 	/**
-	 * @see self::parent()
+	 * Returns the Kirby instance
+	 * @since 5.1.0
```
bastianallgeier:
```suggestion
	 * @since 6.0.0
```

src/Content/Field.php
```diff
@@ -200,11 +132,12 @@ public function value(string|Closure|null $value = null): mixed
 			return $this->value;
 		}
 
+		$clone = clone $this;
```
bastianallgeier, line 135: What's the reason for cloning here? I think this could be a breaking change.

distantnative: Within `::value()` we are cloning the field object anyways (also before this PR), but I saw cases where within the closure `$this->value` was altered directly and not just a value returned. But this would then alter the value of the old field object, not the cloned new one (before this PR), which is why I think that's a bug and cloning needs to happen before calling the closure.

bastianallgeier: I completely missed the clone line below.

bastianallgeier:
I keep on thinking about how we could potentially have custom methods also in a trait or custom class, but I guess there really is no good solution for this.

config/methods.php
bastianallgeier: When we no longer allow to overwrite field methods, extending should not be an issue, right? It feels pretty dirty to keep all the old code around for a very unlikely scenario. If we see that a lot of people need it during the alpha or beta, we could still do that.

distantnative:
@bastianallgeier I think there would be a way but not sure that's actually more helpful for developers: `HasMethods` could support class names for its static `$methods` prop as well. And in `::getMethods()` loop through the registered classes to find the method in one of them. But where would you see the advantage?

bastianallgeier:
There's none if we don't manage to get IDE support through that.

src/Cms/Collection.php
```diff
@@ -126,7 +127,13 @@ public function append(...$args): static
...
 				is_callable([$args[0], 'id']) === true
 			) {
-				return parent::append($args[0]->id(), $args[0]);
+				try {
+					return parent::append($args[0]->id(), $args[0]);
+				} catch (Throwable) {
+					// is_callable might be true when the object implements
+					// the magic __call method, but __call can still throw
+					// an exception
+				}
```
lukasbestle: `Questions`: How is this related to the field method rewrite? Do we really need to swallow all throwables here?

If it's just about `->id()` potentially failing and there is no other fix for that, I think it would be safer to swallow throwables only for that call, not for the whole `parent::append()` call.

lukasbestle:
I think everything we would do at runtime would have a really (really) hard time of being picked up by IDEs. And since plugins need to register their extensions, the only way to pick those up would be a custom Kirby extension for IDE language servers. By that point, it would not really matter how we implement it in Kirby. Runtime stays runtime. So I think there's nothing we can really do there.

About overwriting core methods: I would assume that some devs have done that despite the docs. But I guess we will see during the pre-release phase if it will be a notable problem. We could still go with Florens' suggested solution for specific tags should it become relevant.

src/Cms/Collection.php
distantnative: Yes, it is about `is_callable([$args[0], 'id'])` being true due to a  `__call` method in the `$args[0]` class. But if that actually doesn't resolve the call but throws (as this PR introduces for `Content\Field::__call()` ), we ran into walls.

Not sure how we could only swallow it for `$args[0]->id()` cause it still would also need to wrap the `return` so that if something is thrown, we do not just stop there but continue with the `return parent::append($args[0])` line.

Do you have an idea @lukasbestle?

distantnative: Maybe:

```php
// try to determine the key from the provided item
if (
	is_object($args[0]) === true &&
	is_callable([$args[0], 'id']) === true
) {
	try {
		$id = $args[0]->id();
	} catch (Throwable) {
		// is_callable might be true when the object implements
		// the magic __call method, but __call can still throw
		// an exception
		return parent::append($args[0]);
	}

	return parent::append($id, $args[0]);
}

return parent::append($args[0]);
```

lukasbestle: Would it be an option to switch from `is_callable()` to `method_exists()`? This would then only cover objects that actually have a real `id()` method.

distantnative: I was a bit afraid whether this would be a significant breaking change for such an allrounder class as `Collection`, maybe working on some objects that use `__call` to provide method-like access to an `id` property. But could be worth it to do that breaking change and really require a dedicated `id()` method.

distantnative: But of course we are running ourselves into this breaking change in at least the `Panel\Area` class. And have a unit test failing. Do we want to pursue this?

EDIT: have added a version with the necessary changes.

lukasbestle: Hm, you are right. Thinking about it again, I realized we have another potentially common case with the `Obj` class that also doesn't define an `id()` method but is likely used together with `Collection`s in the wild.

Your suggestion in https://github.com/getkirby/kirby/pull/7082#discussion_r2462859943 would work. Not super clean, but it definitely avoids breaking changes that could bite us.

distantnative: I have added that.

tests/Content/FieldMethodsTest.php
```diff
@@ -1,1110 +0,0 @@
...
-use Kirby\Cms\Users;
-use Kirby\Data\Data;
-use Kirby\Data\Json;
-use Kirby\Data\Yaml;
-use Kirby\Exception\InvalidArgumentException;
-use Kirby\Filesystem\Dir;
-use Kirby\Image\QrCode;
-use Kirby\TestCase;
-
-class FieldMethodsTest extends TestCase
```
lukasbestle, line 22: `Suggestion`: I actually think having the field methods tests in a separate test class still makes sense, especially given that the methods are defined in a trait. Would be fine for me to keep it like this though.

src/Panel/Ui/FilePreview/ImageFilePreview.php
```diff
@@ -30,17 +31,22 @@ public static function accepts(File $file): bool
 
 	public function details(): array
 	{
-		return [
-			...parent::details(),
-			[
+		$details    = parent::details();
+		$dimensions = $this->file->dimensions();
+
+		if ($dimensions instanceof Dimensions) {
```
lukasbestle: `Question`: How can `$file->dimensions()` ever not return a `Dimensions` object? As far as I can tell, `File` proxies the call to `Image`, which has a `dimensions()` method that is type-hinted to always return `Dimensions`.

Also I don't really understand how this change relates to the field methods refactoring.

lukasbestle (approved):
Apart from these minor things, I think the PR is ready to merge.

src/Panel/Ui/FilePreview/ImageFilePreview.php
distantnative: I was thinking that this is needed for all non-image files because then `$file->dimensions()` would return a field instance and with his PR non existent field method calls throw an exception now.

But then again, you are right, this is the image file preview. The file should always be an image.

I realized that `ImageFilePreviewTest::testProps()` seems to be wrongly implemented and threw me off in that direction.

lukasbestle: Ah, that makes sense. Thanks for the explanation :)

tests/Content/FieldMethodsTest.php
distantnative: I've moved them to their own test file.

src/Panel/Ui/FilePreview/ImageFilePreview.php
distantnative: I have reverted it and fixed the unit test instead.

bastianallgeier (approved):

PR #7090, opened 2025-03-26
Fix block selector not closing after pasting block

### LINKED ISSUE #7087: Block Selector not closing after pasting block
## Description
When the block selector is open and I paste a block, the block selector stays opens. The dev tools console shows this error: `Uncaught TypeError: this.close is not a function`.

Image

**Expected behavior**  
Pasting should close the block selector, according to this line:
https://github.com/getkirby/kirby/blob/main/panel/src/components/Forms/Blocks/BlockSelector.vue#L124

## To reproduce
1. In the starterkit, open a blog / notes article.
2. Copy a block.
3. Open the block selector and paste the block.

## Your setup
**Kirby Version**  
4.6.1 and 5.0.0-beta.4

 - OS: macOS 15.3.2
 - Browser: Chrome 134

fnwbr:
Fixes #7087

* Call the correct `$emit("close")` function after pasting a block to then close the block selector

### DIFF
--- a/panel/src/components/Forms/Blocks/BlockSelector.vue
+++ b/panel/src/components/Forms/Blocks/BlockSelector.vue
@@ -122,7 +122,6 @@ export default {
 		paste(e) {
 			this.$emit("paste", e);
 			this.close();
-			this.close();
 		},
 		close() {
 			this.$emit("cancel");
--- a/panel/src/components/Forms/Blocks/BlockSelector.vue
+++ b/panel/src/components/Forms/Blocks/BlockSelector.vue
@@ -66,7 +66,7 @@ export default {
 			type: String
 		}
 	},
-	emits: ["cancel", "input", "paste", "submit"],
+	emits: ["cancel", "close", "input", "paste", "submit"],
 	data() {
 		return {
 			selected: null
@@ -121,10 +121,7 @@ export default {
 	methods: {
 		paste(e) {
 			this.$emit("paste", e);
-			this.close();
-		},
-		close() {
-			this.$emit("cancel");
+			this.$emit("close");
 		}
 	}
 };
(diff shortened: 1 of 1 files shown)


fnwbr:
Full disclosure, I've used this GitHub AI stuff called "Copilot Workspace" for the first time ever and it did everything. I'm baffled...

But no worries, we've locally verified the change: the dialog will now close after pasting. I assume that this is the intended behavior, even though pasting blocks was working perfectly fine beforehand.

distantnative:
Please don't point a PR to the `main` branch but to `develop-patch` or `develop-minor`.

panel/src/components/Forms/Blocks/BlockSelector.vue
```diff
@@ -122,6 +122,9 @@ export default {
 		paste(e) {
 			this.$emit("paste", e);
 			this.close();
+		},
+		close() {
+			this.$emit("cancel");
```
distantnative: `Nitpick:` emitting `close` could be semantically more correct, functionally I think both work

fnwbr: No yeah, I think you're right. Wasn't sure if this was going to work... changing.

fnwbr:
Updated; this should also be more in line with how you do it in https://github.com/getkirby/kirby/blob/main/panel/src/components/Forms/Blocks/BlockPasteboard.vue#L33

_EDIT:_ Also apologies for not following the PR template, which I only realized after talking to @aofn.

distantnative:
No need to apologize. It's great that you're helping us with fixing a bug. Thanks for that. We just need to make sure it ends up in the right places :)

bastianallgeier (approved):

bastianallgeier:
Thanks for the fix. It works great.

PR #7098, opened 2025-03-28
Use type hinting in A::get instead of is_array

### LINKED ISSUE #7096: Replace is_array in A::get with a type
## Description
The A::get helper is build in a way to return non array back as result. Why? Where is this useful?
In my opinion feeding a non array into the the helper should yield an error not silently continue.

Also when swapping the check with a type for the param the `is_array` call can be avoided and since Kirby internally uses this helper gazillion of times the performance impact is **2% of total execution time** (meassured with SPX profile) and even bigger in setups with lots of content (because of reading more content fields via that helper).

```php
public static function get(
		$array,
		string|int|array|null $key,
		mixed $default = null
	) {
		if (is_array($array) === false) {
			return $array;
		}

```

```php
public static function get(
		array $array,
		string|int|array|null $key,
		mixed $default = null
	) {
                /* REMOVE THESE LINES
		if (is_array($array) === false) {
			return $array;
		}
                */

```

https://github.com/getkirby/kirby/blob/a0e64e87743af69a3798bfc78e4e578b674a1173/src/Toolkit/A.php#L269

bastianallgeier:
### Enhancements

- Use type hinting in `A::get()` instead of the is_array check. Results in a 2% improvement of total execution time according to @bnomei https://github.com/getkirby/kirby/issues/7096

### Breaking changes

- `A::get()` no longer accepts a first argument that is not an array

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/A.php
+++ b/src/Toolkit/A.php
@@ -255,21 +255,17 @@ class A
 	 * // result: ['cat' => 'miao', 'dog' => 'wuff'];
 	 * ```
 	 *
-	 * @param mixed $array The source array
+	 * @param array $array The source array
 	 * @param string|int|array|null $key The key to look for
 	 * @param mixed $default Optional default value, which
 	 *                       should be returned if no element
 	 *                       has been found
 	 */
 	public static function get(
-		$array,
+		array $array,
 		string|int|array|null $key,
 		mixed $default = null
 	) {
-		if (is_array($array) === false) {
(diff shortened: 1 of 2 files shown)


distantnative (approved):

distantnative:
Went through Kirby and couldn't find any place where this would seem to be a problem (e.g. anywhere potentially passing `null`).

PR #7099, opened 2025-03-28
Improve `F::read()` performance

### LINKED ISSUE #7095: Avoid is_readable check when possible
## Description

The `is_readable` function seems innocent but since it checks the file on the disk it has a noticable impact on performance when execute thousands of times. My suggestion would be to delay it and just try and fail. This will improve the read function speed about **30-40%** based on my tests – which is huge imho.

**current**
```php
public static function read(string $file): string|false
	{
		if (
			is_readable($file) !== true &&
			Str::startsWith($file, 'https://') !== true &&
			Str::startsWith($file, 'http://') !== true
		) {
			return false;
		}

		return file_get_contents($file);
	}
```

**proposed**
```php
public static function read(string $file): string|false
	{
		if (
			Str::startsWith($file, 'https://') !== true &&
			Str::startsWith($file, 'http://') !== true
		) {
			return false;
		}

		try { 
                    return file_get_contents($file);
                } catch (Exception $ex) {
                    return false;
                }
	}
```

https://github.com/getkirby/kirby/blob/a0e64e87743af69a3798bfc78e4e578b674a1173/src/Filesystem/F.php#L612

bastianallgeier:
### Enhancements

- `F::read()` now uses the @ error control operator instead of calling is_readable first. This should lead to a read function speed improvement of about 30-40% according to @bnomei https://github.com/getkirby/kirby/issues/7095

### Breaking changes

- `F::read()` no longer supports requesting URLs with protocols such as `http://` and `https://`. Please use the `Remote` class for this purpose.
- `F::read()` no longer silently returns `false` when the file exists but is not readable. Instead it triggers a PHP "Permission denied" warning.

## Ready?

### For review team

### DIFF
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -611,11 +611,19 @@ class F
 			return false;
 		}
 
-		try {
-			return @file_get_contents($file);
-		} catch (Throwable) {
+		// exit early on empty paths that would trigger a PHP `ValueError`
+		if ($file === '') {
 			return false;
 		}
+
+		// to increase performance, directly try to load the file without checking
+		// if it exists; fall back to a `false` return value if it doesn't exist
+		// or is not readable while letting other warnings through
+		return Helpers::handleErrors(
+			fn (): string|false => file_get_contents($file),
+			fn (int $errno, string $errstr): bool => str_contains($errstr, 'No such file') || str_contains($errstr, 'Permission denied'),
+			false
+		);
 	}
 
 	/**
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -607,7 +607,11 @@ class F
 	 */
 	public static function read(string $file): string|false
 	{
-		return @file_get_contents($file);
+		try {
+			return @file_get_contents($file);
+		} catch (Throwable) {
+			return false;
+		}
 	}
 
 	/**
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -607,15 +607,7 @@ class F
 	 */
 	public static function read(string $file): string|false
 	{
-		if (
-			is_readable($file) !== true &&
-			Str::startsWith($file, 'https://') !== true &&
-			Str::startsWith($file, 'http://') !== true
-		) {
-			return false;
-		}
-
-		return file_get_contents($file);
+		return @file_get_contents($file);
 	}
 
 	/**
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -618,10 +618,10 @@ class F
 
 		// to increase performance, directly try to load the file without checking
 		// if it exists; fall back to a `false` return value if it doesn't exist
-		// or is not readable while letting other warnings through
+		// while letting other warnings through
 		return Helpers::handleErrors(
 			fn (): string|false => file_get_contents($file),
-			fn (int $errno, string $errstr): bool => str_contains($errstr, 'No such file') || str_contains($errstr, 'Permission denied'),
+			fn (int $errno, string $errstr): bool => str_contains($errstr, 'No such file'),
 			false
 		);
 	}
--- a/src/Filesystem/F.php
+++ b/src/Filesystem/F.php
@@ -607,6 +607,10 @@ class F
 	 */
 	public static function read(string $file): string|false
 	{
+		if (str_contains($file, '://') === true) {
+			return false;
+		}
+
 		try {
 			return @file_get_contents($file);
 		} catch (Throwable) {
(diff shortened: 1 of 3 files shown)


bastianallgeier:
@lukasbestle I think you should have a look at this. Are we ok with the error surpressing operator here?

distantnative:
@bastianallgeier Isn't the `@` actually preventing that anything is thrown that can be caught?

bnomei:
the `@` is not supression exceptions. thats why i had the try-catch block initally. see test https://3v4l.org/ltg7U

distantnative:
But see https://3v4l.org/IH1V6 - without `@` it throws a warning when the file doesn't exist, with `@` the catch block isn't reached either. So the catch code isn't doing anything?

src/Filesystem/F.php
```diff
@@ -598,15 +598,11 @@ public static function niceSize(
 	 */
 	public static function read(string $file): string|false
 	{
-		if (
-			is_readable($file) !== true &&
-			Str::startsWith($file, 'https://') !== true &&
-			Str::startsWith($file, 'http://') !== true
```
lukasbestle, line 613: `Issue`: We cannot remove these checks. Without them, it will be possible to do remote reads (which could be a vulnerability in internal networks). As Bruno had it in his original code example, these should be moved to the top of the method and return early.

src/Filesystem/F.php
```diff
@@ -598,15 +598,11 @@ public static function niceSize(
...
-		) {
+		try {
+			return @file_get_contents($file);
+		} catch (Throwable) {
 			return false;
 		}
```
lukasbestle: `Issue`: I don't think we can use `try catch` here. This only works in the CMS context because Whoops converts PHP warnings (such as "No such file or directory") into exceptions. But as soon as Whoops is disabled, there won't be an exception, so the `catch` won't execute and the method will return `void`, which will then be a fatal error because of the type hinting and might also break expectations in other parts of the system.

Instead of `try catch` with error suppression, it would be a lot cleaner to use `Helpers::handleErrors()`. The question is: Does that have a significant performance impact? It at least does not do I/O, so I guess it won't, but we should verify that.

src/Filesystem/F.php
bastianallgeier: Your comment makes sense, but our old code does not really match that: 

```php
if (
	is_readable($file) !== true &&
	Str::startsWith($file, 'https://') !== true &&
	Str::startsWith($file, 'http://') !== true
) {
	return false;
}
```

Wouldn't it be correct to do this instead? 

```php
if (str_starts_with($file, 'https://') || str_starts_with($file, 'http://')) {
    return false;
}
```

Our old code does basically the opposite.

src/Filesystem/F.php
bastianallgeier: I've tried wrapping it in handleErrors, but the warnings are still created without the error controller.

src/Filesystem/F.php
lukasbestle: Oh, you are right. I totally missed that the logic works in the opposite way. I looked through all uses of `F::read()` in our current (4.7.0) codebase and we never use the HTTP requests anywhere with `F::read()`.

Removing support would be a breaking change, but we could do so in v5. It would certainly be a lot more secure than having this unexpected remote request behavior in a file system class. We could even do this IMO:

```php
if (str_contains($file, '://') === true) {
	return false;
}

src/Filesystem/F.php
lukasbestle: Could you push your try? If we suppress the "No such file or directory" specifically in the handler callback but let all other warnings and errors be thrown, I think that would be exactly what we need.

src/Filesystem/F.php
bastianallgeier: I think it's totally worth to introduce this as a breaking change. I've rebased the PR to v5/develop. We already have the Remote class as a far better alternative to do remote requests and if you really want to keep it super basic, you can still use file_get_contents. I've added the check and also improved the tests.

src/Filesystem/F.php
bastianallgeier: The thing here really is that we cannot surpress the warnings in any other way as far as I can see. The controller is the best bet and it still keeps the results in line with what we had so far. We return false if the file cannot be read and the content if it can.

lukasbestle: Pushed a working version, see Discord

lukasbestle:
@bnomei We have updated the code to use PHP error handling. Could you please check how that compares performance-wise to the original version and the one where all warnings are suppressed with the `@` operator?

lukasbestle (approved):

PR #7101, opened 2025-03-29
`novalidate` custom fields in page create dialog

### LINKED ISSUE #7079: Page Create Dialog: no validation for drafts
Currently, the backend part of the dialog only validates the field if a published page not a draft will be created (correct). However, the native browser validation still kicks in and prevents the dialog to be submitted even for drafts.

We should consider if for drafts the native browser validation can be prevented. Maybe a minimal solution could be at least to prevent the required attribute.

distantnative:
### Summary of changes
- All fields have received a `novalidate` prop which sets the `formnovalidate` attribute
- Custom fields for the page create dialog are set to `novalidate` when creating a draft

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Field.vue
+++ b/panel/src/components/Forms/Field.vue
@@ -58,10 +58,18 @@
 </template>
 
 <script>
-import { disabled, help, id, label, name, required } from "@/mixins/props.js";
+import {
+	disabled,
+	help,
+	id,
+	label,
+	name,
+	novalidate,
+	required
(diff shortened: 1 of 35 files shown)


bastianallgeier:
I'm a bit confused about the formnovalidate attribute. MDN says that it's only valid for the submit and image input https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#formnovalidate

distantnative:
No, you are right, I missed that. So we need another solution.

PR #7104, opened 2025-03-29
feat!: Vue 3

distantnative:
## Description

> [!IMPORTANT]
> This PR will need discussion and testing by more eyes/hands than mine before it should be merged.

### Summary of changes
- Upgrade Vue to version 3.5
  - Removed all `.native` modifiers
  - Renamed `destroyed` to `unmounted`
  - Removed `vue-portal` library and using new native `Teleport`
  - Removed `set` and `del` from Vue - we can use pure javascript now
  - A number of changes to deal with Vue 3's stronger/stricter reactivity
  - `Vue.prototype.` is now `app.config.globalProperties.`
 - `vue` is externalized and can be loaded by Kirby as well as plugins via JS import map

### Reasoning
Vue 2 has reached EOL.

### Enhancements
- Panel: Migrated from Vue 2 to Vue 3

### Breaking changes
- All plugins that have compiled Vue Single File Components with Vue 2 have to recompile their SFC for Vue 3 to work with Kirby 6. Consider also the [Vue 3 migration guide](https://v3-migration.vuejs.org/breaking-changes/).
- All plugin JS files get loaded as module.

## Ready?

### For review team

- [ ] Squash PR


panel/src/components/Collection/Item.vue
```diff
@@ -48,8 +48,8 @@
 		>
 			<!-- Buttons -->
 			<k-button
-				v-for="(button, buttonIndex) in buttons"
-				:key="'button-' + buttonIndex"
+				v-for="button in buttons"
+				:key="JSON.stringify(button)"
```
distantnative, line 52: `Todo:` We should solve this already in v5 or on `develop/v6` before the Vue 3 merge. Ideally also not with `JSON.stringify`

panel/src/components/Drawers/FormDrawer.vue
```diff
@@ -8,7 +8,10 @@
 		@submit="$emit('submit', value)"
 		@tab="$emit('tab', $event)"
 	>
-		<slot slot="options" name="options" />
+		<template #options>
+			<slot name="options" />
+		</template>
```
distantnative: `Todo:` I don't think this is a Vue 3 change. Try to already change this in v5.

panel/src/components/Drawers/TextDrawer.vue
```diff
@@ -8,7 +8,10 @@
 		@submit="$emit('submit', value)"
 		@tab="$emit('tab', $event)"
 	>
-		<slot slot="options" name="options" />
+		<template #options>
+			<slot name="options" />
+		</template>
```
distantnative: `Todo:` I don't think this is a Vue 3 change. Try to already change this in v5.

panel/src/components/Forms/Blocks/Block.vue
```diff
@@ -106,6 +106,7 @@ export default {
 		"paste",
 		"prepend",
 		"remove",
+		"removeSelected",
```
distantnative: `Todo:` Add emit already in v5.

panel/src/components/Forms/Toolbar/Toolbar.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<nav v-if="buttons.length" class="k-toolbar" :data-theme="theme">
 		<template v-for="(button, index) in buttons">
-			<hr v-if="button === '|'" :key="index" />
+			<hr v-if="button === '|'" :key="'separator-' + index" />
```
distantnative: `Todo:` Check if we could already do this in v5.

panel/src/helpers/useUid.js
```diff
@@ -0,0 +1,5 @@
+let id = 0;
+
+export function useUid() {
+	return String(id++);
+}
```
distantnative: `Todo:` Check if newer Vue 3 versions don't include something like this natively.

panel/src/mixins/preserveListeners.js
```diff
@@ -0,0 +1,43 @@
...
+						});
+					}
+				}
+			}
+		}
+	},
+	destroyed() {
+		this.__listenersController.abort();
+	}
+};
```
distantnative, line 43: `Discussion:` Unsure if this could have an impact on performance. But so far haven't found another way to still use listeners even when using `inheritAttrs: false`.

views/panel.php
```diff
@@ -20,6 +20,12 @@
 
   <title>Kirby Panel</title>
 
+  <script type="importmap">
+	{
+	  "imports": <?= json_encode($assets['import-maps'] ?? []) ?>
+	}
+	</script>
```
distantnative: `Todo:` check if Kirby itself could start using Vue via an import map in v5 already.

distantnative: https://github.com/getkirby/kirby/pull/7107

panel/src/helpers/object.js
```diff
@@ -9,7 +11,27 @@ export function clone(value) {
...
+			return Object.keys(input).reduce((acc, key) => {
+				acc[key] = unwrapper(input[key]);
+				return acc;
+			}, {});
+		}
+
+		return input;
+	};
+
+	return unwrapper(value);
```
distantnative, line 34: `Investigate:` could maybe also solved by `JSON.parse(JSON.stringify(input))`

views/panel.php
```diff
@@ -45,7 +45,7 @@
   <?php endforeach ?>
 
   <?php foreach ($assets['js'] as $js): ?>
-  <?php if (($js['type'] ?? null) === 'module'): ?>
+  <?php if (($js['defer'] ?? null) !== true): ?>
```
distantnative, line 48: Is this right?

panel/src/components/Lab/PlaygroundView.vue
```diff
@@ -90,7 +92,7 @@ export default {
 			component.template = this.template;
 
 			// unwrap to be recognized as new component
-			this.component = { ...component };
+			this.component = markRaw({ ...component });
```
distantnative, line 99: Maybe doesn't need both, `markRaw` and `...`?

panel/src/mixins/preserveListeners.js
distantnative: @johannschopplich @bastianallgeier This is one part of the Vue 3 PR where additional brainpower (and experience) could be helpful. This seems a bit ugly but I did not fine any other solution, maybe you know one. The issue is that in Vue 3 when we set `inheritAttrs: false` also listeners get ignored and are not attached to the root element of the component. This mixin that gets applied to every component adds them manually, which seems to work but also seems like a hack.

panel/src/helpers/object.js
distantnative: @johannschopplich @bastianallgeier Also here your help would be much appreciated. I was facing really weird reactive behaviors with Vue 3 that led me to this solution how to clone a nested object and removing the reactivity. But I really wonder that there isn't a better way to do this in Vue 3?

johannschopplich: It may feel odd and counterintuitive at first, but this is actually a cleaner solution than Nuxt's approach (imho). A reactive object may be nested, so you have to "break" its reactivity deeply in specific scenarios. Objects like props are all reactive, so we often don't want to pass their reactive proxy objects, but only their underlying plain value instead.

Nuxt typically accomplishes this by using [`klona` across their codebase](https://grep.app/search?f.repo=nuxt%2Fnuxt&f.repo.pattern=nuxt%2Fnuxt&q=klona) to create a non-reactive clone from reactive proxy objects. One of the core composables `useRuntimeConfig`, for example:
https://github.com/nuxt/nuxt/blob/3e7608920b0f3c50b638f48a233375f2e18bd73d/packages/kit/src/runtime-config.ts#L17

Without introducing another dependency, such as `klona`, your verbose approach is cleaner and even more readable. Kudos!

panel/src/mixins/preserveListeners.js
johannschopplich: I haven't had this issue with Vue before. May feel hacky, but yeah, pretty solid solution.

johannschopplich: Side note: Isn't `destroyed` replaced by `unmounted` in Vue 3?

PR #7107, opened 2025-03-29
Load Vue via import map

distantnative:
### Summary of changes
- Make `Vue` bundle accessible as JS Import map

### Reasoning
Prepares for Vue 3 usage.

### Additional context
As we have to build a minified ESM runtime version of Vue ourselves, the file grows from 76 to 99 KB.

## Ready?

### For review team

### DIFF
--- a/src/Panel/Assets.php
+++ b/src/Panel/Assets.php
@@ -27,7 +27,7 @@ use Kirby\Toolkit\A;
  */
 class Assets
 {
-	protected bool $dev;
+	protected bool $isDev;
 	protected App $kirby;
 	protected string $nonce;
 	protected Plugins $plugins;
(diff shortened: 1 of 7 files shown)


src/Panel/Assets.php
```diff
@@ -331,14 +337,17 @@ public function url(): string
 	 * Get the correct Vue script URL depending on dev mode
 	 * and the enabled/disabled template compiler
 	 */
-	public function vue(bool $production = true): string
+	public function vue(): string
 	{
-		$script = $this->kirby->option('panel.vue.compiler', true) === true ? 'vue' : 'vue.runtime';
+		// During dev mode, load the dev version of Vue
+		if ($this->isDev === true) {
+			return $this->url . '/node_modules/vue/dist/vue.esm.browser.js';
```
distantnative, line 346: In dev mode, we now always load the full Vue bundle. Before, if the `panel.vue.compiler` was set to `false` it would load a runtime-only bundle. But I think dev mode should always override this config option.

src/Panel/Assets.php
```diff
@@ -203,61 +206,63 @@ public function favicons(): array
...
+				'type' => 'module'
 			],
 			'plugins' => [
 				'nonce' => $this->nonce,
 				'src'   => $this->plugins->url('js'),
-				'defer' => true
+				'defer' => true,
 			],
 			...A::map($this->custom('panel.js'), fn ($src) => [
-				'nonce' => $this->nonce,
```
distantnative, line 238: Nonce missing is a problem when using Html::tag

src/Panel/Assets.php
bastianallgeier: I'm not sure about this to be honest. When you intentionally need to switch off the compiler in a project, but it still works in dev environments, it might cause quite some confusion.

distantnative: But who actually runs develop mode in a project? This doesn't mean debug mode, it's `panel.dev` config active AND vite running in `npm run dev` (not even for plugins, for the core)

bastianallgeier: 🙈 totally mixed it up. You are right. In this case, it makes perfect sense.

bastianallgeier (approved):

PR #7108, opened 2025-03-30
`k-panel-menu`: simplify gap

distantnative:
### Summary of changes
- `margin: auto` is a good way in a flex context to create a space, instead of growing an item.
- `.k-panel-menu-buttons:last-child` had no effect because of the activation button menu

### DIFF
--- a/panel/src/components/View/Menu.vue
+++ b/panel/src/components/View/Menu.vue
@@ -154,13 +154,8 @@ export default {
 }
 /* Keep the remaining space between 2nd last and last button group */
 .k-panel-menu-buttons[data-second-last="true"] {
-	flex-grow: 1;
+	margin-bottom: auto;
 }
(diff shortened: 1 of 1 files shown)


bastianallgeier (dismissed):

bastianallgeier:
As soon as the license is activated, this happens …

Screenshot 2025-03-31 at 11 43 59

distantnative:
Totally didn't think of that 🙈 Do you have a good CSS idea? Or going back to the data attribute?

bastianallgeier:
I think we could only work with a :has query to check for the Activation button and I'm not sure if it is worth it tbh.

distantnative:
@bastianallgeier Kept the data attribute around now, switching to the `margin: auto` solution and removed the dead code.

bastianallgeier (approved):

PR #7116, opened 2025-04-03
Fix unsetting field values in changes 

### LINKED ISSUE #7022: [v5] Image focus can't be removed

bastianallgeier:
## Description

When passing null as field value when saving changes, the value from the latest version would be magically merged back in. This PR fixes this regression. 

### Fixes

- The focus point can be removed again, once set. https://github.com/getkirby/kirby/issues/7022
- Passing `null` as field value will remove the field correctly from the latest version, when publishing changes.
- `Version::isIdentical()` is now comparing two versions more reliably, by not merging original values.

### Enhancements

- New `merge: true` argument for `Kirby\Form\Form::for()` which will disable merging original values. Merging is activated by default. 

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -234,19 +234,23 @@ class Version
 
 		$a = Form::for(
 			model: $this->model,
+			merge: false,
 			props: [
 				'language' => $language->code(),
 				'values'   => $a,
-			]
+			],
 		)->values();
 
 		$b = Form::for(
 			model: $this->model,
+			merge: false,
 			props: [
 				'language' => $language->code(),
 				'values'   => $b
 			]
 		)->values();
+
+		dump($a, $b);
 
 		ksort($a);
 		ksort($b);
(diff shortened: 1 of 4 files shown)


distantnative:
@bastianallgeier It appears to solve part of the problem (image focus can be unset, saved and on reload it remains without focus). However, I am still running in one related problem:

- Set focus on an image
- Save
- Unset the focus (do not save)
- Reload view
- Focus is back

I think the problem originates in `Panel\Model::content()` where `Form::for()` also merges in the latest content, not just returns the changes content. However, adding `merge: false` alone doesn't solve the problem as then the form controls stop showing up in the Panel alltogether. `Panel\Model::content()` probably needs to take care of the situation as well then when the changes version does not exist (as latest isn't anymore merged into it automatically).

bastianallgeier:
@distantnative I've tracked it down yesterday and it's getting more complicated again. 

There are three parts to this. 

Kirby\Panel\Model::content() should look something like this: 
```php
public function content(): array
{
	$version = $this->model->version('changes');

	if ($version->exists('current') === false) {
		$version = $this->model->version('latest');
	}

	// create a form which will collect the latest values for the model,
	// but also pass along unpublished changes as overwrites
	return Form::for(
		model: $this->model,
		props: [
			'values' => $version->content('current')->toArray()
		],
		merge: false
	)->values();
}
```

This would be very clean, but the problem here is that there might be extra fields in "latest" which will no longer be in the final version. I.e. uuid, slug or template for files. Selectively merging them seems very complicated. 

Then we need to work on the changes method in our content.js module. We simply ignore removed fields in there. One option could be another loop: 

```js
changes(env = {}) {
    // changes can only be computed for the current view
    if (this.isCurrent(env) === false) {
        throw new Error("Cannot get changes for another view");
    }

    const changes = {};

    for (const field in panel.view.props.content) {
        const changed = JSON.stringify(panel.view.props.content[field]);
        const original = JSON.stringify(panel.view.props.originals[field]);

        if (changed !== original) {
            changes[field] = panel.view.props.content[field];
        }
    }

    for (const field in panel.view.props.originals) {
        if (changes[field] === undefined) {
            changes[field] = null;
        }
    }

    return changes;
},
```

Then there's still a problematic part in Controller\Changes::save where we use the form again and it will merge the submitted content with the original version once more. I tried to fix it there as well, but failed so far. 

It really is still quite a complicated setup.

Issue #6301, opened 2024-02-25
`$panel.dialog.open()`: reactive props


distantnative:
Our dialog components do not respect updates from reactive props. They are turned static.

Consider this modified example of the text dialog lab:

```html
<template>
	<k-lab-examples>
		<k-lab-example :flex="true" label="default">
			<k-button
				icon="open"
				variant="filled"
				@click="
					$panel.dialog.open({
						component: 'k-text-dialog',
						props: {
							text: text
						}
					})
				"
			>
				Open
			</k-button>
		</k-lab-example>
	</k-lab-examples>
</template>

<script>
export default {
	data() {
		return {
			time: null
		};
	},
	computed: {
		text() {
			return "This is a text dialog: " + this.time;
		}
	},
	mounted() {
		setInterval(() => {
			this.time = new Date().toLocaleTimeString();
		}, 1000);
	}
};
</script>
```

The text isn't updating while the dialog is opened.

PR #7118, opened 2025-04-04
Reactive blocks drawer hidden icon

### LINKED ISSUE #7105: Blocks drawer: hidden icon in drawer options not reactive
Click on the button does unhide the block, but the icon remains in the block drawer options instead of disappearing.

Branch: v5/develop

afbora:
### Fixes
- Blocks drawer: hidden icon in drawer options not reactive #7105

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Drawers/BlockDrawer.vue
+++ b/panel/src/components/Drawers/BlockDrawer.vue
@@ -14,7 +14,7 @@
 				v-if="hidden"
 				class="k-drawer-option"
 				icon="hidden"
-				@click="$emit('show')"
+				@click="show"
 			/>
 			<k-button
 				:disabled="!prev"
@@ -63,6 +63,12 @@ export default {
 		"show",
 		"submit",
 		"tab"
-	]
+	],
+	methods: {
+		show() {
+			this.hidden = false;
+			this.$emit("show");
+		}
+	}
 };
 </script>
--- a/panel/src/components/Drawers/BlockDrawer.vue
+++ b/panel/src/components/Drawers/BlockDrawer.vue
@@ -11,7 +11,7 @@
 	>
 		<template #options>
 			<k-button
-				v-if="hidden"
+				v-if="isHidden"
 				class="k-drawer-option"
 				icon="hidden"
 				@click="show"
@@ -64,9 +64,14 @@ export default {
 		"submit",
 		"tab"
 	],
+	data() {
+		return {
+			isHidden: this.hidden
+		};
+	},
 	methods: {
 		show() {
-			this.hidden = false;
+			this.isHidden = false;
 			this.$emit("show");
 		}
 	}


panel/src/components/Drawers/BlockDrawer.vue
```diff
@@ -63,6 +63,12 @@ export default {
 		"show",
 		"submit",
 		"tab"
-	]
+	],
+	methods: {
+		show() {
+			this.hidden = false;
```
distantnative: Don't you get console errors with this? Usually Vue is strictly against directly mutating props.

afbora: Hmm, no 🤷‍♂️

afbora: I'm not expert about that but I think you're right. I have updated as your idea anyway.

bastianallgeier: I think we should update the drawer props instead: 

```js
this.$panel.drawer.props.hidden = false;
```

afbora: I think my last change that Nico's idea is better :)

distantnative: @afbora I wouldn't create a new data entry. It introduces just another place this needs to be updated, another place that could be false later. Better to have a single place of truth (the prop).

@bastianallgeier I also don't think directly updating the drawer prop is the best idea. We should find out why emitting the event isn't working - or why it is working but the drawer props aren't reactively updating. I think that's the actual bug.

As I think about it, it's probably the same issue as https://github.com/getkirby/kirby/issues/6301

afbora: @distantnative I looked into it a bit. The problem seemed simple but I guess it's more difficult than I expected. Feel free to close the PR

afbora:
After digging into Vue’s reactivity system, I realized that props are one-way bindings. `block.isHidden` was already being updated correctly (`set(block, "isHidden", false)`), so the underlying data was never the issue. I think the problem is that the `BlockDrawer` component only received the **initial value** of `block.isHidden` as a plain prop, and since props are not reactive to changes in their original source unless explicitly watched, the component had **no idea** the value had changed. 

So far IMHO this PR changes seems make sense :)

afbora:
We can also add a watch to the hidden prop in `BlockDrawer` to make sure the local value updates if the parent changes it later.

````js
watch: {
  hidden(value) {
    this.isHidden = value;
  }
}

distantnative:
Have you tried if that works?

My understanding is rather that Vue props indeed are reactive. But the problem is that we lose reactivity when opening a dialog or drawer down in `State.set()`. I've tried a few things but so far not found a way how to keep reactivity alive.

I really think it's the same problem as https://github.com/getkirby/kirby/issues/6301

panel/src/components/Drawers/BlockDrawer.vue
distantnative: I think I would indeed rather keep this small bug around until we have a proper fix of the reactivity. Instead of patch-fixing only this one occurrence. What do you think @bastianallgeier ?

PR #7121, opened 2025-04-06
Fix `App::$events` initialization

distantnative:
### Summary of changes
- `App::$events` is initialized before error handler and extension, so that any error thrown e.g. by a plugin and handled by the error handler can actually deal with the `system.exception` event
- `Events` class doesn't accept a custom `$hooks` argument anymore
- `Events` class requires `$app` argument now
- `Events` class now loads hooks from extensions each time a hook runs. This could be a performance issue. However, caching the hooks array could lead to problems when the first event is already triggered/applied before all extensions (and thus all hooks have been registered).

## Ready?

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -98,7 +98,8 @@ class App
 	 */
 	public function __construct(array $props = [], bool $setInstance = true)
 	{
-		$this->core = new Core($this);
+		$this->core   = new Core($this);
+		$this->events = new Events($this);
 
 		// start with a fresh version cache
 		VersionCache::$cache = [];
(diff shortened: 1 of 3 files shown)


bastianallgeier:
As long as the hooks are mostly used for mutations, I'm not super worried about the performance part. If we would start using them for reading models, it could indeed be a problem. 

Can we add a unit test that checks for the issue that we had before the refactoring, to make sure that we definitely fixed it and it won't happen again later?

distantnative:
@bastianallgeier It shouldn't be running the hooks anyways, soo not important what happens inside the hooks. Only the `$this->app->extensions('hooks')` call happening now multiple times and not just once. But I guess that's ok.

bastianallgeier:
Just looked into an option to write a unit test for it, but it's going to be ugly to get that one right. I think we are fine without it.

bastianallgeier (approved):

PR #7130, opened 2025-04-09
New `::submit` method for fields

bastianallgeier:
### Enhancements

- New `::submit` for fields to be used when a form is submitted in a request. `::fill()` is now only responsible for providing the values that the form should have on render.

### Breaking changes

None

## Docs

Full docs for all Form changes: https://www.notion.so/getkirby/Fields-docs-1c359ef0a5cb805189b5ffdc7fa54e0b

## Ready?

### For review team

### DIFF
--- a/src/Form/Mixin/Value.php
+++ b/src/Form/Mixin/Value.php
@@ -137,6 +137,20 @@ trait Value
 	}
 
 	/**
+	 * Submits a new value for the field.
+	 * Fields can overwrite this method to provide custom
+	 * submit logic. This is useful if the field component
+	 * sends data that needs to be processed before being
+	 * stored.
+	 *
+	 * @since 5.0.0
+	 */
+	public function submit(mixed $value): static
+	{
+		return $this->fill($value);
+	}
+
+	/**
 	 * Returns the value of the field in a format to be used in forms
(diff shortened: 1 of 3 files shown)


src/Form/Mixin/Value.php
```diff
@@ -136,6 +136,20 @@ protected function setDefault(mixed $default = null): void
...
+	 * Fields can overwrite this method to provide custom
+	 * submit logic. This is useful if the field component
+	 * sends data that needs to be processed before being
+	 * stored.
+	 *
+	 * @since 5.0.0
+	 */
+	public function submit(mixed $value): static
+	{
+		return $this->fill($value);
```
distantnative, line 150: `Understanding:` E.g. with the picker fields, this would be the place where we could transform the big data array to simple list of (UU)IDs?

distantnative (approved):

src/Form/Mixin/Value.php
bastianallgeier: Yes, exactly. We basically have two ways now: 

**::fill:**
values come from storage, will be used for field components and might need to be transformed

**::submit** 
values come from form submissions and will eventually go into storage and they might need to be transformed.

PR #7134, opened 2025-04-10
Remove $defaults argument from form methods wherever possible

bastianallgeier:
### Refactoring from previous betas

- Remove `$defaults` argument from form and field methods wherever possible, to clean up default value handling
  - Removed from `Field::toFormValue`
  - Removed from `Field::toStoredValue`
  - Removed from `FieldClass::toFormValue`
  - Removed from `FieldClass::toStoredValue`
  - Removed from `Fields::toFormValues`  
  - Removed from `Fields::toStoredValues`  
  - Removed from `Form::toFormValues`  
  - Removed from `Form::toStoredValues`  

### Breaking changes

None. The `toForm...` and `toStored...` methods didn't exist in v4.

## Reasoning

Removing the $defaults argument significantly simplifies the methods and makes them a lot easier to extend. Especially for custom fields. They can now focus on how values need to be transformed on fill or submit. It's also clearer to use the `$fields->defaults()` method in form setups and removes some of the magic behind filling in defaults. 

### Filling fields with defaults

```php
$fields = new Fields(
  fields: [
    'a' => [
      'default' => 'Some default value',
      'type'    => 'text'
    ], 
    'b' => [
	    'type' => 'text',
    ]
  ],
  model: $page  
);

$fields->fill($fields->defaults());

// fill in additional values afterwards to overwrite defaults
$fields->fill([
  'b' => 'A value for the field without default'
]);
```

## Ready?

### For review team

### DIFF
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -137,7 +137,7 @@ class Form
 					unset($data[$field->name()]);
 				}
 			} else {
-				$data[$field->name()] = $field->toStoredValue($defaults);
+				$data[$field->name()] = $field->data($defaults);
 			}
 		}
 
@@ -284,18 +284,18 @@ class Form
 	 * Returns an array with the form value of each field
 	 * (e.g. used as data for Panel Vue components)
 	 */
-	public function toFormValues(bool $defaults = false): array
+	public function toFormValues(): array
 	{
-		return $this->fields->toFormValues($defaults);
+		return $this->fields->toFormValues();
 	}
 
 	/**
 	 * Returns an array with the stored value of each field
 	 * (e.g. used for saving to content storage)
 	 */
-	public function toStoredValues(bool $defaults = false): array
+	public function toStoredValues(): array
 	{
-		return $this->fields->toStoredValues($defaults);
+		return $this->fields->toStoredValues();
 	}
 
 	/**
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -137,7 +137,11 @@ class Form
 					unset($data[$field->name()]);
 				}
 			} else {
-				$data[$field->name()] = $field->data($defaults);
+				if ($defaults === true && $field->isEmpty() === true) {
+					$field->fill($field->default());
+				}
+
+				$data[$field->name()] = $field->toStoredValue();
 			}
 		}
 
--- a/src/Form/Mixin/Value.php
+++ b/src/Form/Mixin/Value.php
@@ -17,7 +17,11 @@ trait Value
 	protected mixed $value = null;
 
 	/**
-	 * @deprecated 5.0.0 Use `::toStoredValue()` instead
+	 * @deprecated 5.0.0 Use `::toStoredValue()` instead to receive
+	 * the value in the format that will be needed for content files.
+	 *
+	 * If you need to get the value with the default as fallback, you should use
+	 * the fill method first `$field->fill($field->default())->toStoredValue()`
 	 */
 	public function data(bool $default = false): mixed
 	{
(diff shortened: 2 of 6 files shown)


src/Form/Form.php
```diff
@@ -137,7 +137,7 @@ public function data($defaults = false, bool $includeNulls = true): array
 					unset($data[$field->name()]);
 				}
 			} else {
-				$data[$field->name()] = $field->toStoredValue($defaults);
+				$data[$field->name()] = $field->data($defaults);
```
distantnative: `Question:` Haven't we deprecate `::data()` with the recommendation to use `::toStoredValue()` instead?

src/Form/Mixin/Value.php
```diff
@@ -21,7 +21,11 @@ trait Value
 	 */
 	public function data(bool $default = false): mixed
 	{
-		return $this->toStoredValue($default);
+		if ($default === true && $this->isEmpty() === true) {
+			$this->fill($this->default());
+		}
+
+		return $this->toStoredValue();
```
distantnative, line 32: `Todo:` We shouldn't add functionality on which remaining code relies to a method that we have already deprecated. While we recommend to use `::toStoredValue()` this will then function differently.

src/Form/Mixin/Value.php
```diff
@@ -181,37 +185,37 @@ public function submit(mixed $value): static
...
+		return $this->toFormValue();
 	}
 
 	/**
 	 * Returns the value of the field if it has a value
 	 * otherwise it returns null
 	 *
 	 * @see `self::toFormValue()`
-	 * @todo might get deprecated or reused later
+	 * @todo might get deprecated or reused later. Use `self::toFormValue()` instead.
```
distantnative, line 215: But `:: toFormValue()` doesn't support the default behavior. So it won't be a real replacement then.

tests/Form/FieldTest.php
```diff
@@ -1126,7 +1126,6 @@ public function testToFormValue()
 		$this->assertNull($field->value());
 
 		$field = new Field('test', ['default' => 'Default value']);
-		$this->assertSame('Default value', $field->toFormValue(true));
 		$this->assertSame('Default value', $field->value(true));
```
distantnative, line 1129: `Question:` What will be the ideal way to get the default value of a single field in the future? `$field->default()`? Then we should also stop testing `$field->value(true)` here?

bastianallgeier: Depending on what you need, the best way to get the default field value would be: 

```php
$field->fill($field->default())->toFormValue();
$field->fill($field->default())->toStoredValue();
```

src/Form/Mixin/Value.php
bastianallgeier: Would it be enough in your opinion, to extend the explanation in the deprecated warning and add a full example how to handle default values?

src/Form/Form.php
bastianallgeier: You are right. I was lazy here. I've just replaced it with the full code, which will also probably make it easier to understand and get rid of it later.

src/Form/Mixin/Value.php
bastianallgeier: Same here. Do you think a better explanation is enough?

src/Form/Mixin/Value.php
distantnative: I think. I understand it that we want in the future that all code that used `$field->data(true)` needs to handle this itself (instead of having a class method that provides this functionality). We need to explain that with the deprecation message then.

distantnative (approved):

PR #7135, opened 2025-04-10
Passthrough for fields

bastianallgeier:
## Merge first

- [ ] https://github.com/getkirby/kirby/pull/7134

## Context

I came up with the concept of UnknownFields in the original PR: https://github.com/getkirby/kirby/pull/7081 but this is actually way over the top. Inspired by Paddle's passthrough argument for their checkout API, we can simply pass through values in a way leaner way, by adding them to a new $passthrough property. The PHP API is IMHO very readable and with the option to chain methods, it becomes pretty self-explainatory to use. 

### Enhancements

- New `Fields::passthrough($values)` method

### Refactoring

- Switch to PHPUnit Attributes in the FieldsTest class

### Breaking changes

None

### Filling unknown fields

If you want to pass through values for undefined fields (e.g. slug, uuid, etc. for models) you can use the passthrough method. Be careful though because this can be abused to inject unwanted values. 

```php
$fields = new Fields(
  fields: [
    'text' => [
      'type' => 'text'
    ]
  ],
  model: $page  
);

$input = [
    'text'        => 'Some text',
    'description' => 'Some description',
];

$fields->passthrough($input)->fill($input);

dump($fields->toFormValues());
dump($fields->toStoredValues());
// will both include the passed through description field 
```

You can remove passthrough values again by calling passthrough without argument or with an empty array: 

```php
$fields->passthrough();

dump($fields->toFormValues());
dump($fields->toStoredValues());
// will both no longer include the passed through description field 
```

## Ready?

### For review team

### DIFF
--- a/src/Form/Fields.php
+++ b/src/Form/Fields.php
@@ -167,8 +167,10 @@ class Fields extends Collection
 	 */
 	public function passthrough(array $values = []): static
 	{
+		// always start with a fresh set of passthrough values
+		$this->passthrough = [];
+
 		if ($values === []) {
-			$this->passthrough = [];
 			return $this;
 		}
 
--- a/src/Form/Fields.php
+++ b/src/Form/Fields.php
@@ -26,7 +26,7 @@ use Kirby\Toolkit\Str;
 class Fields extends Collection
 {
 	protected Language $language;
-	protected array $passthrough = []; 
+	protected array $passthrough = [];
 
 	public function __construct(
 		array $fields = [],
@@ -178,8 +178,8 @@ class Fields extends Collection
 			if ($this->get(strtolower($key)) !== null) {
 				continue;
 			}
-			
-			$this->passthrough[$key] = $value;			
+
+			$this->passthrough[$key] = $value;
 		}
 
 		return $this;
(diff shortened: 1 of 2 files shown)


src/Form/Fields.php
```diff
@@ -173,7 +201,7 @@ public function toArray(Closure|null $map = null): array
 	 */
 	public function toFormValues(): array
 	{
-		return $this->toArray(fn ($field) => $field->toFormValue());
+		return $this->toValues('toFormValue');
```
distantnative: `Suggestion:` I think passing an actual Closure instead of a string name. I think it should work as `$this->toValues($this->toFormValue(...))`.

Allows us more easily to not overlook this e.g. when the method name changes.

bastianallgeier: I have not internalized this new PHP awesomeness yet :)

bastianallgeier: Ah, too bad. It does not seem to work. I guess because of $this not pointing to the field object, but the collection. Or maybe I'm doing something wrong.

distantnative: Ah true, sorry did read it as `$this->toFormValue()` not `$field->toFormValue()`.

I would still suggest to rather keep passing the existing closure to the new `::toValues()` instead just the string. It just feels a bit more reliable than the method name as string.

bastianallgeier: I've just changed it.

distantnative (approved):

PR #7138, opened 2025-04-11
New Fields::reset method

bastianallgeier:
## Merge first

- [x] https://github.com/getkirby/kirby/pull/7134
- [x]  https://github.com/getkirby/kirby/pull/7135
- [x] https://github.com/getkirby/kirby/pull/7136

## Context

We need to be able to reset/clear the values of all fields later when we want to use the same Field instance multiple times. E.g. when comparing versions. 

Let's have a look at this code: 

```php
$fields = new Fields(
  fields: $page->blueprint()->fields()
  model: $page
);

$latest = $page->version('latest')->content()->toArray();
$changes = $page->version('latest')->content()->toArray();

$a = $fields->reset()->fill($latest)->toFormValues();
$b = $fields->reset()->fiil($changes)->toFormValues();
```

If the reset method would not exist, some fields could already contain values from previous fill calls and we would have to create two independent instances of the Fields object to make a valid comparison. When we call reset first, we can make sure that we reuse the same fields, but we always start from scratch with fresh values.  

### Enhancements

- New `Fields::reset()` method, which will set the value of each field to null. 

### Breaking changes

None

### Resetting field values

```php
$fields = new Fields(
  fields: [
    'text' => [
      'type' => 'text'
    ]
  ],
  model: $page  
);

$fields->fill([
  'text' => 'Some text',
]);

$fields->get('text')->toFormValue()
// will return 'Some text'

$fields->reset();

$fields->get('text')->toFormValue()
// will return ''
```

## Ready?

### For review team

### DIFF
--- a/src/Form/Fields.php
+++ b/src/Form/Fields.php
@@ -202,6 +202,18 @@ class Fields extends Collection
 	}
 
 	/**
+	 * Resets the value of each field
+	 */
+	public function reset(): static
+	{
+		foreach ($this->data as $field) {
+			$field->fill(null);
+		}
+
+		return $this;
+	}
+
+	/**
 	 * Sets the value for each field with a matching key in the input array
 	 * but only if the field is not disabled
 	 * @since 5.0.0
--- a/src/Form/Fields.php
+++ b/src/Form/Fields.php
@@ -206,6 +206,10 @@ class Fields extends Collection
 	 */
 	public function reset(): static
 	{
+		// reset the passthrough values
+		$this->passthrough = [];
+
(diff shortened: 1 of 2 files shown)


distantnative (commented):
`Question:` should fields after resetting still be part of the output (e.g toStoredValues) or rather be absent completely from the values array?

bastianallgeier:
When a field is defined in v4 but has no value, the field is still stored but empty. I think this is what still needs to happen here.

distantnative (approved):

PR #7143, opened 2025-04-13
Reset Vite dev mode after PHPUnit tests

lukasbestle:
### Summary of changes

Clean up `.vite-running` file that is created during PHPUnit tests

### Reasoning

I often do my backend work with production assets and kept wondering why a `.vite-running` file was appearing out of nowhere.

### DIFF
--- a/tests/Panel/AssetsTest.php
+++ b/tests/Panel/AssetsTest.php
@@ -14,9 +14,11 @@ use Kirby\Toolkit\Str;
  */
 class AssetsTest extends TestCase
 {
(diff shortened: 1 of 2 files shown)


bastianallgeier:
But this would mean that the dev mode breaks when it is actually running and you run unit tests in the meantime, right?

lukasbestle:
You are right. When I am at my desk, I will fix this by remembering the original state in `setUp` and restoring that state (whatever it was) in `tearDown`.

lukasbestle:
Done :)

bastianallgeier (approved):

PR #7148, opened 2025-04-16
New Fields::for method

bastianallgeier:
## Merge first

- [x] https://github.com/getkirby/kirby/pull/7147

## Context

The `Fields:for()` method can be used in multiple places later to load the field objects for a model and work with them. We don't need the full Form class with its content magic for a lot of cases (API, dialogs, versions). This method will also help to get rid of `Form::for()` in general. 

### Enhancements

- New `Fields::for()` method, which can be used to get a collection of all field objects for a model

### Breaking changes

None

## Docs

```php
$fields = Fields::for($page);

$fields->submit([
  'text' => 'Some nice text'
]);

$fields->toStoredValues();
```

### Get field definitions in a different language

```php
$fields = Fields::for($page, 'de')->toProps();
```

## Ready?

### For review team

### DIFF
--- a/src/Form/Fields.php
+++ b/src/Form/Fields.php
@@ -185,6 +185,20 @@ class Fields extends Collection
 	}
 
 	/**
+	 * Creates a new Fields instance for the given model and language
+	 */
+	public static function for(
+		ModelWithContent $model,
+		Language|string $language = 'default'
+	): static {
+		return new static(
+			fields: $model->blueprint()->fields(),
+			model: $model,
+			language: Language::ensure($language),
+		);
+	}
+
+	/**
 	 * Returns the language of the fields
 	 */
 	public function language(): Language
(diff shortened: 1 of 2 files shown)


src/Form/Fields.php
```diff
@@ -184,6 +184,20 @@ public function findByKeyRecursive(string $key): Field|FieldClass|null
 		return $field;
 	}
 
+	/**
+	 * Creates a new Fields instance for the given model and language
+	 */
+	public static function for(
+		ModelWithContent $model,
+		Language|string $language = 'default'
```
distantnative, line 192: `Question:` is `current` maybe the more common use case?

bastianallgeier: A bit torn here. We mix this. In the version classes, we use default and in form and content method, we often use current.

distantnative (approved):

PR #7150, opened 2025-04-16
Use the Fields class to compare versions

bastianallgeier:
## Merge first

- [x] https://github.com/getkirby/kirby/pull/7148
- [x] https://github.com/getkirby/kirby/pull/7149

## Context

This is finally one of those big steps that took 18 gazillion smaller PRs to get there. 🎉

### Refactoring

- Use the Fields class to improve Version comparison. This will especially help to improve the performance of comparing two versions, as the field instances can be reused. 

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -9,7 +9,7 @@ use Kirby\Cms\Page;
 use Kirby\Cms\Site;
 use Kirby\Exception\LogicException;
 use Kirby\Exception\NotFoundException;
-use Kirby\Form\Form;
+use Kirby\Form\Fields;
 use Kirby\Http\Uri;
 use Kirby\Toolkit\Str;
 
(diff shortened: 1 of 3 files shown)


src/Content/Version.php
```diff
@@ -232,21 +233,8 @@ public function isIdentical(
...
-
-		$b = Form::for(
-			model: $this->model,
-			props: [
-				'language' => $language->code(),
-				'values'   => $b
-			]
-		)->values();
+		$a = $fields->reset()->passthrough($a)->fill($a)->toFormValues();
+		$b = $fields->reset()->passthrough($b)->fill($b)->toFormValues();
```
distantnative: Reading this I wonder if instead the passthrough method the fill method should get a `strict: false` argument

bastianallgeier: I thought about it as well and for some reason that I cannot replicate anymore, I decided against it :) But the code in this case would be a lot nicer indeed.

bastianallgeier: I've added it but called the argument passthrough, which is false by default.

distantnative (approved):

PR #7153, opened 2025-04-17
New Model::versions method

bastianallgeier:
## Merge first

- [x] https://github.com/getkirby/kirby/pull/7148
- [x] https://github.com/getkirby/kirby/pull/7149
- [x] https://github.com/getkirby/kirby/pull/7150

## Context

This is another big milestone, where we finally can use all the Fields improvements to only create a single Fields instance for both versions. This should eliminate any prior performance issues and also give us a better boilerplate to work with content consistently between versions. 

### Enhancement

- New `Model::versions()` method, which returns an array with the two versions `latest` and `changes`. If there's no changes version, the `latest` will be used. 

### Deprecated 

- `Model::content()` and `Model::originals()` are deprecated. Use `Model::versions()` instead. 

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Panel/Model.php
+++ b/src/Panel/Model.php
@@ -36,7 +36,7 @@ abstract class Model
 	/**
 	 * Get the content values for the model
 	 *
-	 * @deprecated Use `versions()` instead
+	 * @deprecated 5.0.0 Use `self::versions()` instead
 	 */
 	public function content(): array
 	{
--- a/src/Panel/Model.php
+++ b/src/Panel/Model.php
@@ -332,16 +332,6 @@ abstract class Model
 	}
 
 	/**
-	 * Get the original content values for the model
-	 *
-	 * @deprecated Use `versions()` instead
-	 */
-	public function originals(): array
-	{
-		return $this->versions()['latest'];
-	}
-
-	/**
 	 * Returns the full path without leading slash
 	 */
 	abstract public function path(): string;
--- a/src/Panel/Model.php
+++ b/src/Panel/Model.php
@@ -4,9 +4,10 @@ namespace Kirby\Panel;
 
 use Closure;
 use Kirby\Cms\File as CmsFile;
+use Kirby\Cms\Language;
 use Kirby\Cms\ModelWithContent;
 use Kirby\Filesystem\Asset;
-use Kirby\Form\Form;
+use Kirby\Form\Fields;
 use Kirby\Http\Uri;
 use Kirby\Toolkit\A;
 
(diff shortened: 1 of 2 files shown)


src/Panel/Model.php
```diff
@@ -34,24 +35,12 @@ abstract public function buttons(): array;
 
 	/**
 	 * Get the content values for the model
+	 *
+	 * @deprecated Use `versions()` instead
```
distantnative:
```suggestion
	 * @deprecated 5.0.0 Use `self::versions()` instead
```

src/Panel/Model.php
```diff
@@ -344,10 +333,12 @@ public function options(array $unlock = []): array
...
 	/**
 	 * Get the original content values for the model
+	 *
+	 * @deprecated Use `versions()` instead
 	 */
 	public function originals(): array
 	{
-		return Form::for(model: $this->model)->values();
+		return $this->versions()['latest'];
 	}
```
distantnative: `Todo:` Can be removed completely as it didn't exist prior to v5.

src/Panel/Model.php
```diff
@@ -467,6 +459,36 @@ public function url(bool $relative = false): string
...
+	public function versions(): array
+	{
+		$language = Language::ensure('current');
+		$fields   = Fields::for($this->model, $language);
+
+		$latestVersion  = $this->model->version('latest');
+		$changesVersion = $this->model->version('changes');
+
+		$latestContent  = $latestVersion->content($language)->toArray();
+		$changesContent = $latestContent;
```
distantnative, line 470: `Discuss:` it feels wrong that we send a changes content along even when there is none. Shouldn't we rather send `null` along in this case?

bastianallgeier: Then the frontend would need to take care of the fallback. I'm fine with that too. It just felt a bit more pragmatic - and maybe a bit more lazy :)

distantnative: I guess it has been the same before this PR, right? Cause `Form::for()` merged in the latest version even for `$model->changes()`.

bastianallgeier: Yes, but we no longer want the merging part. That's the big difference now. We can finally prevent the merging logic at this point. 

Sending two objects, even when it's not needed could maybe be a performance issue. That would definitely speak for sending null instead. But thinking of the reactivity issues in Vue, it might also make sense to always have the full objects there from the beginning. Especially when we need to work with view.props in some cases.

distantnative (approved):

PR #7155, opened 2025-04-23
Move validation logic to the Version class

bastianallgeier:
### Reasoning

We want to be able to validate individual versions before publishing them. 

- New `Version::errors()` method
- New `Version::isValid()` method, which is now used in `ModelWithContent::isValid()` instead of the Form class. 

### Refactoring

- Switch to PHPUnit Attributes in `VersionTest` and `ModelWithContentTest`
- New `ModelWithContent::isValid` unit tests

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/ModelWithContent.php
+++ b/src/Cms/ModelWithContent.php
@@ -353,7 +353,7 @@ abstract class ModelWithContent implements Identifiable, Stringable
 	 */
 	public function isValid(): bool
 	{
-		return Form::for($this)->isValid() === true;
+		return $this->version('latest')->errors('current') === [];
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


distantnative (commented):
I would suggest to also add `Version::isValid($language)`.

bastianallgeier:
@distantnative 👍 while refactoring this, I realized that we have two different ways to validate between `ModelWithContent::isValid()` and `ModelWithContent::errors()` 🙈 In ::errors we go through all sections and also collect errors from sections without fields. (e.g. for pages with a max or min) and in ::isValid we only validate the content. Not sure what to do with that and if we should already tackle it here as well.

distantnative (approved):
@bastianallgeier This is indeed a weird split between `ModelWithContent::errors()` and `ModelWithContent::isValid()`. But it has been weird like this before, so no need to fix that right now with this PR.

PR #7156, opened 2025-04-23
Use Fields class in Changes controller

bastianallgeier:
## Merge first

- [x] https://github.com/getkirby/kirby/pull/7155

## Context

This is another big milestone. We can finally get rid of content merges in the Changes controller and remove the Form class there. 

### Refactoring from previous betas

- The Fields class is now used instead of Form in the Changes controller to save changes. 

### Breaking changes

- None

## Ready?

### For review team

### DIFF
--- a/src/Api/Controller/Changes.php
+++ b/src/Api/Controller/Changes.php
@@ -2,10 +2,12 @@
 
 namespace Kirby\Api\Controller;
 
+use Kirby\Cms\Language;
 use Kirby\Cms\ModelWithContent;
 use Kirby\Content\Lock;
 use Kirby\Content\VersionId;
 use Kirby\Filesystem\F;
+use Kirby\Form\Fields;
 use Kirby\Form\Form;
 
 /**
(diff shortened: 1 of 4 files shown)


src/Api/Controller/Changes.php
```diff
@@ -90,34 +92,45 @@ public static function publish(ModelWithContent $model, array $input): array
...
+
+		// get the changes and latest version for the model
 		$changes = $model->version(VersionId::changes());
 		$latest  = $model->version(VersionId::latest());
 
-		// Removes the old .lock file when it is no longer needed
-		// @todo Remove in 6.0.0
-		static::cleanup($model);
+		// get the source version for the existing content
+		$source  = $changes->exists($language) === true ? $changes : $latest;
```
distantnative, line 110: As it doesn't look that we use `$latest` otherwise, would maybe be better to only initialise it when changes does not exist.

bastianallgeier: We use it once more further down for the isIdentical check

distantnative (approved):

PR #7158, opened 2025-04-27
Number input: support `step: any`

### LINKED ISSUE #7110: Number input: support for `step: any`
Not sure if this is a bug or a feature (I tend towards "bug"). Anway: Currently there seems to be no way to allow/validate an input of arbitrary precision for a `number` field.

E.g. I want to have a number field that allows/successfully validates the values `55`, `55.1` and `55.123456` (basically any positive number).

If I set this up naively like this in my blueprint:

```
          lat:
            type: number
            label: Geografische Breite
            required: true
            min: 0
```

This will only validate successfully `55`. Both `55.1` and `55.123456` will always show the red "invalid" indicator next to the field. (But: I can still press "save" – I'm not getting an error, even though the field is not valid and it is required -> that's why I think this is definitely a bug)

I can add the `step` prop to validate a number with a fixed precision, but this will then stop me from entering a number with a higher precision.

This was tested on `5.0.0-beta.4`.

distantnative:
## Description

@afbora Would be great to get your testing skills on this 🙏

### Summary of changes
- Number input: `step: any` is now supported, allowing for any amount of decimals (e.g. great for lat/lng coordinates) 

### Fixes
- #7110

## Ready?

### For review team

### DIFF
--- a/config/fields/number.php
+++ b/config/fields/number.php
@@ -24,16 +24,20 @@ return [
 		},
 		/**
 		 * Allowed incremental steps between numbers (i.e `0.5`)
+		 * Use `any` to allow any decimal value.
 		 */
-		'step' => function ($step = null) {
-			return $this->toNumber($step) ?? '';
+		'step' => function ($step = null): float|string {
+			return match ($step) {
+				'any'   => 'any',
+				default => $this->toNumber($step) ?? ''
+			};
 		},
 		'value' => function ($value = null) {
 			return $this->toNumber($value) ?? '';
 		}
 	],
 	'methods' => [
-		'toNumber' => function ($value) {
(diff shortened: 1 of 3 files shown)


afbora:
@distantnative Doesn't work correctly. We need to some changes from php side:
https://github.com/getkirby/kirby/blob/fix/7110-number-input-step-any/config/fields/number.php#L29

distantnative:
@afbora oh my of course! I was so stuck in the frontend world that I fully relied on the lab example to test.

distantnative:
@afbora ok another attempt at getting it right this time

afbora (approved):
Looks and works great 👍

PR #7164, opened 2025-04-29
Fields Class Improvements

bastianallgeier:
## Description

The next form steps brought up some more Fields class improvements, which can be extracted into their own PR. 

- The Fields class constructor now accepts a string value for the language parameter
- The Fields class will fall back to the site if no model is given.
- `Fields::passthrough()` will resolve closure values (just like fill and submit)
- Add missing since tags to methods. 

### Fixes from previous betas

- The `Fields::passthrough()` method supported passing an empty array to reset all passthrough values. This seemed like a good idea, but can lead to quite unexpected results, especially when using the fill and submit methods and the input to those is potentially empty (e.g. an empty post request). This could erase passthrough values even if not wanted. This PR removes the option to pass an empty array to reset passthrough values. We can still use the `::reset` method to achieve the same.

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Form/Fields.php
+++ b/src/Form/Fields.php
@@ -65,6 +65,8 @@ class Fields extends Collection
 
 	/**
 	 * Returns an array with the default value of each field
+	 *
+	 * @since 5.0.0
 	 */
 	public function defaults(): array
 	{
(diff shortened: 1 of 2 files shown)


src/Form/Fields.php
```diff
@@ -206,6 +216,8 @@ public static function for(
```
distantnative, line 216: `Suggestion:` For simplicity, maybe pass here `$language` simply to the constructor which now already takes care of ensuring the language object.

distantnative (approved):

PR #7179, opened 2025-04-30
Content module improvements

bastianallgeier:

- `panel.content.changes()` has been renamed to `panel.content.diff()`
- New `panel.content.hasDiff()` method.
- New `panel.content.version(versionId)` method
- New `panel.content.versions()` method
- The `content`and `original` props have been removed from all model views.
- A new `versions` prop has been added instead with `latest` and `changes` as child objects.
- The `changes` property in the `<k-model-tabs>` component has been renamed to `diff`.
- hasChanges has been renamed to hasDiff in all components. 

### Breaking changes

- The `content` prop has been removed from all model views.

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Views/ModelView.vue
+++ b/panel/src/components/Views/ModelView.vue
@@ -9,7 +9,6 @@ export default {
 		api: String,
 		blueprint: String,
 		buttons: Array,
-		content: Object,
 		id: String,
 		link: String,
 		lock: {
@@ -17,7 +16,6 @@ export default {
 		},
 		model: Object,
 		next: Object,
-		originals: Object,
 		prev: Object,
 		permissions: {
 			type: Object,
@@ -35,7 +33,8 @@ export default {
 			type: Array,
 			default: () => []
 		},
-		uuid: String
+		uuid: String,
+		versions: Object
 	},
 	data() {
 		return {
@@ -43,8 +42,11 @@ export default {
 		};
 	},
 	computed: {
-		changes() {
-			return this.$panel.content.changes({
+		content() {
+			return this.versions.changes;
+		},
+		diff() {
+			return this.$panel.content.diff({
 				api: this.api,
 				language: this.$panel.language.code
 			});
@@ -53,7 +55,7 @@ export default {
 			return this.lock.user.email;
 		},
 		hasChanges() {
-			return length(this.changes) > 0;
+			return length(this.diff) > 0;
 		},
 		isLocked() {
 			return this.lock.isLocked;
--- a/panel/src/components/Views/ModelView.vue
+++ b/panel/src/components/Views/ModelView.vue
@@ -1,6 +1,4 @@
 <script>
-import { length } from "@/helpers/object";
-
 /**
  * @internal
  */
@@ -46,16 +44,13 @@ export default {
 			return this.versions.changes;
 		},
 		diff() {
-			return this.$panel.content.diff({
-				api: this.api,
-				language: this.$panel.language.code
-			});
+			return this.$panel.content.diff();
 		},
 		editor() {
 			return this.lock.user.email;
 		},
-		hasChanges() {
-			return length(this.diff) > 0;
+		hasDiff() {
+			return this.$panel.content.hasDiff();
 		},
 		isLocked() {
 			return this.lock.isLocked;
--- a/panel/src/panel/content.js
+++ b/panel/src/panel/content.js
@@ -19,26 +19,28 @@ export default (panel) => {
 			this.saveAbortController?.abort();
 		},
 
+		dialog: null,
+
 		/**
 		 * Returns an object with all changed fields
 		 * @param {Object} env
 		 * @returns {Object}
 		 */
-		changes(env = {}) {
+		diff(env = {}) {
 			// changes can only be computed for the current view
 			if (this.isCurrent(env) === false) {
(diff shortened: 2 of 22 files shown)


panel/src/panel/content.js
```diff
@@ -84,7 +85,7 @@ export default (panel) => {
 				await this.request("discard", {}, env);
 
 				// update the props for the current view
-				panel.view.props.content = panel.view.props.originals;
+				this.versions().changes = this.versions().latest;
```
distantnative: This feels like relying a lot on reactivity and looks like it could break easily (that when assigning a new value to the result of a method this is still actually assigned to the original object in the method).

panel/src/panel/content.js
```diff
@@ -324,6 +328,23 @@ export default (panel) => {
...
+		 */
+		version(versionId) {
+			return this.versions()[versionId];
+		},
+
+		/**
+		 * Returns all versions of the content
+		 * @returns {Object}
+		 */
+		versions() {
```
distantnative, line 349: `Question:` Would this also work with a getter instead of a method?

panel/src/panel/content.js
bastianallgeier: You are right. I wanted to avoid going through the view props all the time, but I guess it's really the more stable solution here.

panel/src/panel/content.js
bastianallgeier: I looked again at all the env usage and I think I want to keep it. There is definitely some potential future benefit to it and we definitely need it for the requests. But that means that we hardly ever can "just" have a getter or we have to mix parts that honor the env and parts that don't. I haven't added env support to the versions method yet, but when we declare it as a method, we keep that option at least to also have an env argument for it later.

distantnative: Ok, I think I would add explicit `changes()` and `latest()` methods in addition then. For developers calling `panel.content.versions().changes` might feel a little double instead of just `panel.content.changes()`/`panel.content.latest()`

bastianallgeier: That's what I created `panel.content.version("latest")` for

bastianallgeier: I don't think we should have direct methods for versions. If we think ahead and there are more potential versions than just latest and changes, we can no longer follow that pattern.

panel/src/components/Views/ModelView.vue
```diff
@@ -53,7 +55,7 @@ export default {
 			return this.lock.user.email;
 		},
 		hasChanges() {
```
distantnative: `Suggestion:` maybe should be called `hasDiff` for consistency

panel/src/components/Views/ModelView.vue
```diff
@@ -35,16 +33,20 @@ export default {
...
 	data() {
 		return {
 			isSaved: true
 		};
 	},
 	computed: {
-		changes() {
-			return this.$panel.content.changes({
+		content() {
+			return this.versions.changes;
```
distantnative, line 44: `Suggestion:` Since we are getting rid of the `content` prop and call our whole module `content` (with versions changes and latest), maybe we don't add this computed prop but use `versions.changes` in the templates instead to make it more explicit what version is used.

panel/src/panel/content.js
```diff
@@ -84,7 +85,7 @@ export default (panel) => {
 				await this.request("discard", {}, env);
 
 				// update the props for the current view
-				panel.view.props.content = panel.view.props.originals;
+				panel.view.props.versions.changes = this.versions().latest;
```
```suggestion
				panel.view.props.versions.changes = this.version("latest");
```

panel/src/panel/content.js
```diff
@@ -193,12 +195,12 @@ export default (panel) => {
 				values = {};
 			}
 
-			panel.view.props.content = {
-				...panel.view.props.content,
+			panel.view.props.versions.changes = {
+				...this.versions().changes,
```
```suggestion
			const changes = panel.view.props.versions.changes = {
				...this.version("changes"),
```

panel/src/panel/content.js
```diff
@@ -193,12 +195,12 @@ export default (panel) => {
...
 				...values
 			};
 
-			return panel.view.props.content;
+			return panel.view.props.versions.changes;
```
```suggestion
			return changes;
```

panel/src/panel/content.js
```diff
@@ -229,7 +231,7 @@ export default (panel) => {
 				this.dialog?.close();
 
 				// update the props for the current view
-				panel.view.props.originals = panel.view.props.content;
+				panel.view.props.versions.latest = this.versions().changes;
```
```suggestion
				panel.view.props.versions.latest = this.version("changes");
```

panel/src/components/Views/ModelView.vue
bastianallgeier: That was all about backward compatibility, but I think you are right that we should rather switch here as well.

bastianallgeier: I looked at it again and we loop the content object through k-sections to each k-section. I think it makes sense to keep the wording here for the sections.

distantnative (approved):

panel/src/components/Views/ModelView.vue
bastianallgeier: I've refactored it, which involved quite some steps. Could you take a closer look at it and check if I didn't miss anything? Do you think it's better this way?

distantnative (approved):

PR #7181, opened 2025-05-01
Consistent VersionId code style

distantnative:
### Summary of changes
- Use simple strings (`'latest'`/`'changes'`) for `$model->version($versionId)` and similar
- Use constant for `$versionId->is($versionID)`, e.g. `$versionId->is(VersionId::LATEST)`
- Use static methods `VersionId::latest()` and `VersionId::changes()` whenever a `VersionId` object is directly required or the method doesn't ensure a proper object by using `VersionId::from()`

### Reasoning
Over time, our code started to be a bit inconsistent how we use version ID references when. This tries to clean it up.

## Ready?

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -101,7 +101,7 @@ class Version
 		Language|string $language = 'default'
 	): void {
 		$language = Language::ensure($language);
-		$latest   = $this->sibling(VersionId::latest());
+		$latest   = $this->sibling('latest');
 
 		// if the latest version of the translation does not exist yet,
 		// we have to copy over the content from the default language first.
@@ -120,7 +120,7 @@ class Version
 		VersionRules::create($this, $fields, $language);
 
 		// track the changes
-		if ($this->id->is(VersionId::changes()) === true) {
+		if ($this->id->is(VersionId::CHANGES) === true) {
 			(new Changes())->track($this->model);
 		}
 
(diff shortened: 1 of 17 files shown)


src/Content/Version.php
```diff
@@ -120,7 +120,7 @@ public function create(
 		VersionRules::create($this, $fields, $language);
 
 		// track the changes
-		if ($this->id->is(VersionId::changes()) === true) {
+		if ($this->id->is(VersionId::CHANGES) === true) {
```
bastianallgeier, line 123: Why not …

```suggestion
		if ($this->id->is('changes') === true) {
```

?

src/Content/Version.php
```diff
@@ -156,7 +156,10 @@ public function delete(Language|string $language = 'default'): void
 
 		// untrack the changes if the version does no longer exist
 		// in any of the available languages
-		if ($this->id->is(VersionId::changes()) === true && $this->exists('*') === false) {
+		if (
+			$this->id->is(VersionId::CHANGES) === true &&
```
bastianallgeier, line 160:
```suggestion
			$this->id->is('changes') === true &&
```

src/Content/Version.php
bastianallgeier: Why do we use the constants here, but not for the sibling calls? I think if we use the strings everywhere now, we should also use them here – or we switch to the constants everywhere.

distantnative: I used the constants here as this was used in most other `$versionId::is()` calls - and as comparing it there to the constant felt the proper thing to do in context of the VersionId class, where no `VersionId::from()` call ensures proper translation.

`Version::sibling()` and `ModelWithContent::version()` receive strings as they ensure a proper handling by calling `VersionId:;from()` internally.

bastianallgeier: I wonder if we need to clean this up in the is method though. It feels pretty weird from just reading the code.

distantnative: Reworked `VersionId::is()` a little bit as suggested.

bastianallgeier (approved):
Nice!

PR #7186, opened 2025-05-05
Only display reset password view when available

lukasbestle:
### Summary of changes

- Session receives a new `kirby.resetPassword => true` key when login occurred via the password reset mode
- The reset password view is loaded with an extra "current password" field (unless in password reset mode)
- Setting a password via the API requires knowledge of the acting user's password (unless in password reset mode)

### Reasoning

- Enforce the password change requirements throughout the system

### Enhancements

- Resetting the password without knowledge of the current password is now only possible after logging in in password reset mode

### Breaking changes

- The `PATCH account/password` and `PATCH users/$user/password` API routes require an additional `currentPassword` parameter with the password of the acting user unless in password reset mode of the current user
- The `panel.api.users.changePassword()` method now receives three arguments for the user ID, the current password of the acting user (new) and the new password to set

## Docs

None

## Ready?

### For review team

### DIFF
--- a/config/api/routes/users.php
+++ b/config/api/routes/users.php
@@ -194,8 +194,9 @@ return [
 
 			$result = $user->changePassword($this->requestBody('password'));
 
-			// only allow to reset the password once, now it is known again
+			// if we changed the password of the current user…
 			if ($user->isLoggedIn() === true) {
+				// …don't allow additional resets (now the password is known again)
 				$this->session()->remove('kirby.resetPassword');
 			}
 
--- a/config/api/routes/users.php
+++ b/config/api/routes/users.php
@@ -184,7 +184,22 @@ return [
 		],
 		'method'  => 'PATCH',
 		'action'  => function (string $id) {
-			return $this->user($id)->changePassword($this->requestBody('password'));
+			$user = $this->user($id);
+
+			// validate password of acting user unless they have logged in to reset it;
+			// always validate password of acting user when changing password of other users
+			if ($this->session()->get('kirby.resetPassword') !== true || $user->isLoggedIn() !== true) {
+				$this->user()->validatePassword($this->requestBody('currentPassword'));
+			}
+
+			$result = $user->changePassword($this->requestBody('password'));
+
+			// only allow to reset the password once, now it is known again
+			if ($user->isLoggedIn() === true) {
+				$this->session()->remove('kirby.resetPassword');
+			}
+
+			return $result;
 		}
 	],
 	[
--- a/config/api/routes/users.php
+++ b/config/api/routes/users.php
@@ -188,7 +188,7 @@ return [
 
 			// validate password of acting user unless they have logged in to reset it;
 			// always validate password of acting user when changing password of other users
-			if ($this->session()->get('kirby.resetPassword') !== true || $user->isLoggedIn() !== true) {
+			if ($this->session()->get('kirby.resetPassword') !== true || $this->user()->is($user) !== true) {
 				$this->user()->validatePassword($this->requestBody('currentPassword'));
 			}
 
--- a/panel/src/api/users.js
+++ b/panel/src/api/users.js
@@ -16,10 +16,10 @@ export default (api) => ({
 	async changeName(id, name) {
 		return api.patch("users/" + id + "/name", { name: name });
 	},
-	async changePassword(id, currentPassword, password) {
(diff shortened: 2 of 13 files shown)


config/api/routes/users.php
```diff
@@ -184,7 +184,22 @@ function ($source, $filename) use ($id) {
 		],
 		'method'  => 'PATCH',
 		'action'  => function (string $id) {
-			return $this->user($id)->changePassword($this->requestBody('password'));
+			$user = $this->user($id);
+
+			// validate password of acting user unless they have logged in to reset it;
+			// always validate password of acting user when changing password of other users
+			if ($this->session()->get('kirby.resetPassword') !== true || $user->isLoggedIn() !== true) {
```
distantnative: `Suggestion:` My brain had troubles comprehend this. Mainly the second part. Is that to check whether the current user is `$user`? If so maybe `$this->user()->is($user)` could be the more easily parsable alternative?

config/api/routes/users.php
```diff
@@ -184,7 +184,22 @@ function ($source, $filename) use ($id) {
...
+				$this->user()->validatePassword($this->requestBody('currentPassword'));
+			}
+
+			$result = $user->changePassword($this->requestBody('password'));
+
+			// only allow to reset the password once, now it is known again
+			if ($user->isLoggedIn() === true) {
```
distantnative:
```suggestion
			// if we changed the password of the current user…
			if ($user->isLoggedIn() === true) {
				// …don't allow additional resets (now the password is known again)
```

panel/src/api/users.js
```diff
@@ -16,8 +16,11 @@ export default (api) => ({
 	async changeName(id, name) {
 		return api.patch("users/" + id + "/name", { name: name });
 	},
-	async changePassword(id, password) {
-		return api.patch("users/" + id + "/password", { password: password });
+	async changePassword(id, currentPassword, password) {
```
distantnative: `Suggestion:` Add the new parameter last so that old code does not pass new passwords now as `currentPassword`.

panel/src/components/Views/Users/ResetPasswordView.vue
```diff
@@ -27,20 +27,49 @@
  * @internal
  */
 export default {
+	props: {
+		/**
+		 * Enable the field for the current password of the acting user
+		 */
+		currentPassword: {
```
distantnative: `Suggestion:` How about naming this `requirePassword` instead? Distinguishes it more from `this.values.currentPassword`.

panel/src/api/users.js
lukasbestle: I thought about it, but decided against it because with order-based arguments, we would never be able to flip the order later without making the mess even bigger. So we would forever have that order that doesn't match the "intuitive" order.

Given that there is only one place in the entire Panel codebase that called this method and uses in plugins are unlikely/not super common, I think documenting it as a breaking change is a more sensible solution overall.

distantnative: Fair. One more thought though: is this really the intuitive order? I'd think that id and new password are more linked as they both belong to the same user (the one updated) while current password can belong to that user but also another (current) user. That's why my thought would be the other two go first together, and only then current password. If you still think the suggested order is the more intuitive, feel free to resolve this though. It's just a suggestion.

lukasbestle: Hm, good thought. I think either view makes sense. Yours has the added benefit that the current password as last argument can be completely omitted when just resetting a password (as it's not needed there).

bastianallgeier (approved):
It all looks good for me now. @lukasbestle you decide if you want to wait for another look from @distantnative

panel/src/api/users.js
```diff
@@ -16,8 +16,11 @@ export default (api) => ({
 	async changeName(id, name) {
 		return api.patch("users/" + id + "/name", { name: name });
 	},
-	async changePassword(id, password) {
-		return api.patch("users/" + id + "/password", { password: password });
+	async changePassword(id, password, currentPassword) {
+		return api.patch("users/" + id + "/password", {
+			password: password,
+			currentPassword: currentPassword
```
distantnative: `Nitpick:` JS allows us to just reference the keys here (when key name and const/variable name are the same) and we usually do that when the whole object can be referenced by keys.
```suggestion
			password,
			currentPassword
```

distantnative (approved):

PR #7191, opened 2025-05-10
`Version::create()`: don't always create latest

### LINKED ISSUE #7184: [v5] creating file content translation throws exception
For non-default languages

### LINKED ISSUE #7185: [v5]`latest` content file for translations created too early
Gets already created for unsaved changes instead of only when publishing

distantnative:
### Summary of changes
- `Version::create()` does not automatically create a latest version a language anymore when creating a changes version for the language. This allows for the following scenario
  - Page has a latest version of the default language, not translated in any secondary language
  - Start drafting the version for. a secondary language (as changes version)
  - Only have a latest version once the changes version for the secondary language is published

### Reasoning
We do not want to create latest/published versions for a secondary language immediately when a Panel user starts editing/creating it.

### Fixes from earlier pre-releases
- https://github.com/getkirby/kirby/issues/7184
- https://github.com/getkirby/kirby/issues/7185

## Ready?

### For review team

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -101,20 +101,6 @@ class Version
 		Language|string $language = 'default'
 	): void {
 		$language = Language::ensure($language);
-		$latest   = $this->sibling('latest');
-
-		// if the latest version of the translation does not exist yet,
-		// we have to copy over the content from the default language first.
-		if (
-			$this->isLatest() === false &&
-			$language->isDefault() === false &&
-			$latest->exists($language) === false
-		) {
-			$latest->create(
-				fields: $latest->read(Language::ensure('default')),
-				language: $language
-			);
-		}
 
 		// check if creating is allowed
 		VersionRules::create($this, $fields, $language);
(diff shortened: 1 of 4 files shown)


src/Content/Version.php
```diff
@@ -502,8 +488,8 @@ public function publish(Language|string $language = 'default'): void
 		// check if publishing is allowed
 		VersionRules::publish($this, $language);
 
-		$latest  = $this->sibling('latest')->read($language);
-		$changes = $this->read($language);
+		$latest  = $this->sibling('latest')->read($language) ?? [];
+		$changes = $this->read($language) ?? [];
```
distantnative, line 492: @bastianallgeier this change is necessary as we get otherwise an `foreach argument is null` error. But it makes me wonder whether it is the best decision that `::read()` returns `null`.

bastianallgeier: I wonder about that as well. We still have the ::exists method in addition if you really need to find out if content exists or not.

bastianallgeier: I looked at it again and I think it adds some value and clarity to return null here when the version does not exist, even if it is a bit more effort to write the fallbacks. I would leave it as it is.

bastianallgeier (approved):

PR #7192, opened 2025-05-10
Keep translated slug for non-default languages

### LINKED ISSUE #7183: Custom slug of secondary language page is renamed to slug of default language on save
## Description
The custom slug of a secondary language page is changed to the slug of the default language when content changes are saved via the panel. This then results in a 404 error when trying to open the page with the intended custom slug. It seems this was introduced with beta.4 as everything works as expected up until beta3.

e.g. 
default slug: /kontakt
secondary slug: /en/contact → /en/kontakt

**Expected behavior**  
The slug in secondary language is not automatically changed to default slug on save.

## To reproduce

1. Set up two languages.
2. Rename slug of a page in the secondary language.
3. Save content changes.
4. Slug is changed to default slug.

## Your setup
**Kirby Version**  
5.0.0-beta.4 and up

distantnative:
### Summary of changes
- Removed `Version::prepareForContent()` - keeps `slug` as part of the content, removing `lock` still but now directly in `Version::content()`

### Reasoning
We ended up removing the slug from content for non-default languages but also not adding it back when saving. So saving would eliminate the custom slug for a secondary language.

### Fixed regressions from previous pre-releases
- https://github.com/getkirby/kirby/issues/7183

### Breaking changes from previous pre-releases
- Removed `Kirby\Content\Version::prepareForContent()`

## Ready?

### For review team

### DIFF
--- a/src/Content/Version.php
+++ b/src/Content/Version.php
@@ -56,8 +56,8 @@ class Version
 			];
 		}
 
-		// prepare raw content file fields as fields for Content object
-		$fields = $this->prepareFieldsForContent($fields, $language);
+		// remove fields that should not be used for the Content object
+		unset($fields['lock']);
 
 		return new Content(
 			parent: $this->model,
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):
Works great and it's good to simplify this some more

PR #7195, opened 2025-05-10
Media firewall 3: Introduce media subhashes

lukasbestle:
### Summary of changes

- Inside the media hash directory (`/media/site/abcdefghij-1234567890`), there is now another directory layer (e.g. `/media/site/abcdefghij-1234567890/bcdefghija`) to separate all individual files (the original and each thumb)
- Transitional code is added so that all old media URLs stay valid until v6 (these are automatically redirected to the new URLs)
- Existing media files and thumbs are automatically migrated to their new locations to avoid having to regenerate them

### Reasoning

- Prevent access to file versions (especially originals but also different thumb variants) by guessing the media URL (so far if you e.g. had `/media/site/abcdefghij-1234567890/file-120x.jpg` you could change the filename to `file.jpg` to access the original, which may not be intended by devs).
- Using another directory layer makes sure that the filenames stay clean and don't need to include generated hashes.

### Additional context

With the transitional and migration code, there shouldn't be a breaking impact to sites. But since there is already a lot going on in v5, I suggest we include this PR in 5.1 or 5.x. This ensures that users upgrading to v5 with a lot of media files and thumbs don't immediately run into performance problems when all those files are migrated.

### Enhancements

- It is no longer possible to guess other URLs to file versions (uploaded file originals and their thumbnails) from a thumbnail URL.

### Deprecated

- The URL structure of media files (file URLs and thumb URLs) has changed. The old URL structure is still supported for now, but will be dropped in Kirby 6.

### Breaking changes

None

## Docs

None

## Ready?

- [ ] More robust migration code (currently not safe against race conditions when multiple requests migrate the same files at the same time)

### For review team

### DIFF
--- a/config/components.php
+++ b/config/components.php
@@ -76,7 +76,7 @@ return [
 		}
 
 		// create url and root
-		$mediaRoot = dirname($file->mediaRoot());
+		$mediaRoot = $file->mediaRoot();
 		$template  = $mediaRoot . '/{{ name }}{{ attributes }}.{{ extension }}';
 		$thumbRoot = (new Filename($file->root(), $template, $options))->toString();
 		$thumbName = basename($thumbRoot);
@@ -102,7 +102,7 @@ return [
 			'modifications' => $options,
 			'original'      => $file,
 			'root'          => $thumbRoot,
-			'url'           => dirname($file->mediaUrl()) . '/' . $thumbName,
+			'url'           => $file->mediaUrl($thumbName),
 		]);
 	},
 
--- a/config/components.php
+++ b/config/components.php
@@ -75,16 +75,17 @@ return [
 			return $file;
 		}
 
-		// create url and root
-		$mediaRoot = $file->mediaRoot();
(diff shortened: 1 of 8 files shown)


distantnative:
@lukasbestle Can I help getting this ready for 5.1?

lukasbestle:
I think we should move it to 5.2 to be honest. I need to take another look at it.

ovenum:
It’s a bit offtopic here, but when you change the generated media url paths in an upcoming release, would you also consider to allow UUID based media root paths instead of the current slug based paths? 

If it’s considered, this looks to be the right moment to avoid two media url path migrations in the future. 

Can create an Issue for this. Did  share this idea on [Discord](https://discord.com/channels/525634039965679616/525641819854471168/1372923203701440663) and briefly talked about this with @distantnative at the Berlin Kirby meetup

lukasbestle:
@ovenum Not off-topic at all. In fact I agree it's a great idea and I have already added it to the internal project todo list for further research.

lukasbestle:
Moving this to the lab for now. We will pick this up again at a later date.

PR #7196, opened 2025-05-10
Media firewall 2: Differentiate between `mediaRoot`/`mediaDir`

lukasbestle:
### Summary of changes

- New custom `$filename` param for `$file->mediaRoot()` and `$file->mediaUrl()`
- New `$file->mediaDir()` method (= directory where media of that file model is stored)
- Rename `mediaRoot()` to `mediaDir()` for other model classes

### Reasoning

- With the new media path structure introduced by #7195, methods throughout the core need access to both the "topmost" media root and the individual file paths.
- Being able to pass a custom filename to `mediaRoot()` and `mediaUrl()` allows to clean up thumb-related code and is also needed for the media firewall PR 3 #7195.

### Enhancements

- New `$file->mediaDir()` method that returns the directories where media of that file model is stored. Matching method aliases for the `Page`, `Site` and `User` classes are provided.
- The `$file->mediaRoot()` and `$file->mediaUrl()` methods now accept a new optional `$filename` argument so they can be used to construct roots and URLs to thumbs.

### Breaking changes

None

## Docs

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -374,6 +374,7 @@ class File extends ModelWithContent
 	/**
 	 * Returns the absolute path to the media folder for the file and its versions
 	 * @internal
+	 * @since 5.0.0
 	 */
 	public function mediaDir(): string
 	{
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -372,6 +372,15 @@ class File extends ModelWithContent
 	}
 
 	/**
+	 * Returns the absolute path to the media folder for the file and its versions
+	 * @internal
+	 */
+	public function mediaDir(): string
+	{
+		return $this->parent()->mediaDir() . '/' . $this->mediaHash();
+	}
+
+	/**
 	 * Creates a unique media hash
 	 * @internal
 	 */
@@ -386,20 +395,11 @@ class File extends ModelWithContent
 	 *
 	 * @param string|null $filename Optional override for the filename
 	 */
-	public function mediaPath(string|null $filename = null): string
+	public function mediaRoot(string|null $filename = null): string
 	{
 		$filename ??= $this->filename();
 
-		return $this->mediaRoot() . '/' . $filename;
-	}
-
-	/**
-	 * Returns the absolute path to the media folder for the file and its versions
-	 * @internal
-	 */
-	public function mediaRoot(): string
-	{
-		return $this->parent()->mediaRoot() . '/' . $this->mediaHash();
+		return $this->mediaDir() . '/' . $filename;
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


src/Cms/File.php
```diff
@@ -383,10 +383,23 @@ public function mediaHash(): string
...
 	 * Returns the absolute path to the file in the public media folder
 	 * @internal
+	 *
+	 * @param string|null $thumbName Optional thumb filename
+	 */
+	public function mediaPath(string|null $thumbName = null): string
+	{
+		$thumbName ??= $this->filename();
+
+		return $this->mediaRoot() . '/' . $thumbName;
```
distantnative: `Suggestion:` I would go with `$filename` as this relates not just to images/thumbs but all files. Same for `::mediaUrl()`.

src/Cms/File.php
```diff
@@ -402,10 +415,15 @@ public function mediaToken(): string
...
+	 * @param string|null $thumbName Optional thumb filename
 	 */
-	public function mediaUrl(): string
+	public function mediaUrl(string|null $thumbName = null): string
 	{
-		return $this->parent()->mediaUrl() . '/' . $this->mediaHash() . '/' . $this->filename();
+		$rootUrl     = $this->parent()->mediaUrl() . '/' . $this->mediaHash();
+		$thumbName ??= $this->filename();
+
+		return $rootUrl . '/' . $thumbName;
```
distantnative: `Nitpick`: combined with suggestion from above

```suggestion
		$url        = $this->parent()->mediaUrl() . '/' . $this->mediaHash();
		$filename ??= $this->filename();

		return $url . '/' . $filename;
```

distantnative (dismissed):
To me the breaking change is worth it as the naming scheme would make a lot of sense this way.

Would be good if @bastianallgeier also has a look at this.

bastianallgeier:
## Changes

- I've created a new `::mediaDir` method for all models. 
- `::mediaRoot` is now an alias for `::mediaDir` in the Page, Site and User classes
- `::mediaRoot` has a new optional $filename param in the File class
- `::mediaPath` has been removed again in favor of the existing `::mediaRoot`
- All unit tests have been adapted.

@lukasbestle please check if you are happy with those changes.

lukasbestle:
Thank you for adapting it, looks good to me. Unfortunately I cannot approve my own PR :)

One thing I thought about is whether we want to deprecate `$page/site/user->mediaRoot()` with a deprecation warning. Just to clean it all up and avoid accumulating aliases that aren't really needed. Especially with our plan to introduce media subclasses (which mean we need to deprecate all those media methods in the model classes anyway).

lukasbestle:
Another thought on the deprecation: We could actually introduce the media subclasses in v5.x and deprecate all old methods in one go then. This avoids having to migrate calls to the new methods twice.

Which means we can merge this as is. If we want, we could already add `@deprecated` tags to the old methods without deprecation warnings, but it’s not required IMO.

bastianallgeier:
@lukasbestle sounds good to me!

PR #7197, opened 2025-05-11
test: Ensure XML parse entities exceeds memory

distantnative:
### Summary of changes
- Increase the billion laughs attack for `Toolkit\XmlTest::testParseRecursiveEntities()`

### Reasoning
Finally found the source of why this unit test kept failing for me locally. The previous malicious billion laughs XML wasn't malicious enough for my local setup and would actually complete instead of running into a memory limit (and thus return actually something instead of the expected `null`). Extending the malicious XML ensures that `simplexml_load_string` fails again.

## Ready?

### DIFF
--- a/src/Toolkit/Xml.php
+++ b/src/Toolkit/Xml.php
@@ -280,7 +280,7 @@ class Xml
 	{
 		$xml = @simplexml_load_string($xml);
 
-		if (is_object($xml) !== true) {
+		if (is_object($xml) === false) {
 			return null;
 		}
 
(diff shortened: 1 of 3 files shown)


lukasbestle (approved):
I've tried something along the same lines but didn't go as far in my tests (maybe until level 6 or 7, without success).

What I'm a bit concerned about is that we are now dealing with 6^8 = 1,679,616 entities here. Should we ever run our tests on a system that again doesn't spot this billion laughs attack, this could break the tests completely (not sure how much memory those entities use in SimpleXML).

But not sure if it's really ever going to be a problem. So I'm fine with merging it. It's great that it fixes the local issue (I'm having the same one on my machine with Herd).

PR #7209, opened 2025-05-12
fix: set locale to `C` while handling errors

### LINKED ISSUE #7086: `F::unlink()` error handling fails when server language not English
## Description

I'm currently working on a new website that was running just fine on **5.0.0-beta.3**, but now has some [serious issues](https://github.com/getkirby/kirby/issues/7084) in the panel after updating to the latest beta. I could track down the one linked above (failing page creation), but am pretty stuck with another one:

**Deleting pages and files fails in all cases and always returns this kind of error:**

- ``unlink(/var/www/clients/clientXX/webXX/web/content/2_about/_changes/ocean.jpg.de.txt): Datei oder Verzeichnis nicht gefunden `` 
- ``unlink(/var/www/clients/clientXX/webXX/web/content/2_about/testkapitel/_changes/kapitel.de.txt): Datei oder Verzeichnis nicht gefunden `` 

I guess it has to do something with the new versions-feature and the "`_changes`"-folder not existing, but I'm not sure, as I'm not able to reproduce this behaviour myself.

I tried to track it down by rebuilding the whole site step by step under a different domain up to the moment, where I had a complete clone of the website running there (including the complete content), but everything kept working fine, including the deleting of files and pages.

I then copied my working clone from Domain B and replaced my original code of Domain A (fixing the configs/urls for sure) only to face the error again… no deleting of pages or files possible.

Both domains run on the same server (Timme/Nginx) with exactly the same configuration.

Pretty strange, but even if I can't reproduce it and it might be somehow releated to my specific domain/server-setup I thought I'd post a bug report nonetheless, as this errors are a direct result of updating to the latest beta and we didn't face them before (with the same server/domain-setup).

## Screenshot

Image

## Your setup

**Kirby Version:**  5.0.0.beta4

bastianallgeier:
### Enhancements

- Always set the locale to `'C'` while handling errors with`Helpers::handleError()` to avoid invalid system error message checks

### Fixed regression from previous betas

- Deleting pages and files no longer results in a PHP warning when PHP is compiled with localized warning messages https://github.com/getkirby/kirby/issues/7086

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/Helpers.php
+++ b/src/Cms/Helpers.php
@@ -120,7 +120,8 @@ class Helpers
 		// backup current locale
 		$locale = setlocale(LC_MESSAGES, 0);
 
-		// set locale to C to avoid locale errors
+		// set locale to C so that errors and warning messages are
+		// printed in English for robust comparisons in the handler
 		setlocale(LC_MESSAGES, 'C');
 
 		/**
--- a/src/Cms/Helpers.php
+++ b/src/Cms/Helpers.php
@@ -159,10 +159,10 @@ class Helpers
 			// action or the standard error handler threw an
 			// exception; this avoids modifying global state
 			restore_error_handler();
-		}
 
-		// reset to original locale
-		setlocale(LC_MESSAGES, $locale);
+			// reset to original locale
+			setlocale(LC_MESSAGES, $locale);
+		}
 
 		return $override ?? $result;
 	}
--- a/src/Cms/Helpers.php
+++ b/src/Cms/Helpers.php
@@ -117,6 +117,12 @@ class Helpers
 	) {
 		$override = null;
 
+		// backup current locale
+		$locale = setlocale(LC_MESSAGES, 0);
+
+		// set locale to C to avoid locale errors
+		setlocale(LC_MESSAGES, 'C');
+
 		/**
 		 * @psalm-suppress UndefinedVariable
 		 */
@@ -153,6 +159,9 @@ class Helpers
 			// exception; this avoids modifying global state
 			restore_error_handler();
 		}
+
+		// reset to original locale
+		setlocale(LC_MESSAGES, $locale);
 
 		return $override ?? $result;
 	}
--- a/tests/Cms/Helpers/HelpersTest.php
+++ b/tests/Cms/Helpers/HelpersTest.php
@@ -5,19 +5,20 @@ namespace Kirby\Cms;
 use Kirby\Exception\Exception;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Toolkit\Obj;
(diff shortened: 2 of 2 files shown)


src/Cms/Helpers.php
```diff
@@ -154,6 +160,9 @@ public static function handleErrors(
 			restore_error_handler();
 		}
 
+		// reset to original locale
+		setlocale(LC_MESSAGES, $locale);
```
lukasbestle: `Issue`: We should do this in the `finally` block so it is executed even if an exception is thrown.

It would be good to test this in the unit tests for `Helpers::handleErrors()` (start the test with a custom locale, check for the overridden `C` locale in the action, throw an exception in the action, catch the exception in the test and afterwards check that the locale is back at the original custom locale).

src/Cms/Helpers.php
```diff
@@ -117,6 +117,12 @@ public static function handleErrors(
 	) {
 		$override = null;
 
+		// backup current locale
+		$locale = setlocale(LC_MESSAGES, 0);
+
+		// set locale to C to avoid locale errors
```
lukasbestle: `Suggestion` to make clearer why all of this is needed:

```suggestion
		// set locale to C so that errors and warning messages are
		// printed in English for robust comparisons in the handler
```

lukasbestle:
I've updated the changelog snippet in the first post to separate it into the change to `Helpers::handleErrors()` and the regression fix.

src/Cms/Helpers.php
bastianallgeier: I've added the unit test and moved the setlocale call to the finally block.

tests/Cms/Helpers/HelpersTest.php
```diff
@@ -144,6 +144,27 @@ public function testHandleErrorsException()
 		);
 	}
 
+	/**
+	 * @covers ::handleErrors
+	 */
+	public function testHandleErrorsLocaleReset()
+	{
+		setlocale(LC_ALL, 'de_DE.UTF-8');
```
lukasbestle: `Suggestions`:
- As this changes global state, we should backup the original locale in `setUp()` and restore it in `tearDown()` (otherwise other tests will have a different baseline depending on whether the entire tests or just a single test class are executed)
- I think it's enough to set `LC_MESSAGES` to keep the change to global state as minimal as possible

lukasbestle (commented):
Apart from the minor suggestion, this is great now.

tests/Cms/Helpers/HelpersTest.php
distantnative: Added these.

lukasbestle (approved):

PR #7212, opened 2025-05-12
feat: `content.fileRedirects` option

lukasbestle:
### Summary of changes

- New `content.fileRedirects` option that controls whether "clean" file routes should be accessible
- Can be set to `true` (default), `false` or a closure to control dynamically per file

### Reasoning

- Allows to limit access to sensitive file originals

### Additional context

Follow-up will be two PRs for v5 (change default to `true` and fix lookup of site files)

### Features

- URLs of the form `https://example.com/some/page/file.pdf` or `https://example.com/some-site-file.pdf` can be disabled or dynamically controlled with a new `content.fileRedirects` option to protect against unauthorized file downloads.

### Breaking changes

None

### Config reference for `content` option: https://getkirby.com/docs/reference/system/options/content

    <since v="4.8.0">
    ## Access to file originals by URL

    Uploaded files can be accessed with URLs of the form `https://example.com/some/page/file.pdf` or `https://example.com/some-site-file.pdf`. This can be useful to share clean links to files that keep working even if the media URL changes, e.g. when moving the site to a different server.

    Depending on the nature of your files, you can control this behavior either globally or dynamically per file by setting the `content.fileRedirects` option.

    Possible values: `true`| `false` | `Closure` (default: `true`)

    ```php
    return [
      'content' => [
        'fileRedirects' => false
      ]
    ];
    ```

    You can also dynamically control this behavior, e.g. based on a file field or other custom logic:

    ```php
    return [
      'content' => [
        'fileRedirects' => fn (File $file): bool => $file->public()->toBool()
      ]
    ];
    ```
    </since>

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1292,11 +1292,35 @@ class App
 
 		// try to resolve image urls for pages and drafts
 		if ($page = $site->findPageOrDraft($id)) {
-			return $page->file($filename);
+			return $this->resolveFile($page->file($filename));
 		}
 
 		// try to resolve site files at least
-		return $site->file($filename);
+		return $this->resolveFile($site->file($filename));
+	}
+
+	/**
+	 * Filters a resolved file object using the configuration
+	 */
+	protected function resolveFile(File|null $file): File|null
+	{
+		// shortcut for files that don't exist
+		if ($file === null) {
+			return null;
+		}
+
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):
I love how simple and effective this is

PR #7214, opened 2025-05-13
fix: Regression in `Media` path traversal hardening

### LINKED ISSUE #7218: Thumbnail error in 4.7.1
## Description
Since `Kirby 4.7.1` a bug was introduced for the creation of thumbnails.
If a configuration of `roots` is created with different paths,  the new code to prevent path traversal, will never guess the realpath

```
$root = F::realpath($source, $index);
```

The parent of the `content` and the `index` will never be the same, and always will be throw an error:
```
The file is not within the parent directory
```

**Expected behavior**  
Create a proper `thumbnail`

## To reproduce

Create a new `roots` structure:

```php
$kirby = new Kirby([
    'roots' => [
        'index' => __DIR__,
        'base' => $base = dirname(__DIR__),
        'content' => $base.'/content',
        'site' => $base.'/site',
    ],
]);

echo $kirby->render();
```

It should throw an error when a thumbnail with attributes like:

```
$file->thumb(['width' => 200])->url(),
```

## Setup

Kirby 4.7.1
PHP 8.4.7

lukasbestle:
### Summary of changes

- Only use the realpath check in `Media` when it is needed.

### Reasoning

The `F::realpath()` check was only intended for the case where the path is constructed from an asset path (passed as string). The other cases don’t need it and broke when the content folder was outside the `index` root.

### Additional context

As noted by @jensscherbl in https://github.com/getkirby/kirby/commit/1dbc4365f746b92afefbddce0ec49014f57afd0c#r156932648.

I suggest we publish this in a patch release 4.7.2.

### Fixes

- Fix regression in 4.7.1 that broke the creation of thumbs when the `content` folder was outside of the `index` root (e.g. in `public` folder setups) #7218

### Breaking changes

None

## Docs

None

### For review team

### DIFF
--- a/src/Cms/Media.php
+++ b/src/Cms/Media.php
@@ -132,18 +132,18 @@ class Media
 			// this adds support for custom assets
 			$source = match (true) {
 				is_string($model) === true
-					=> $index . '/' . $model . '/' . $options['filename'],
+					=> F::realpath(
+						$index . '/' . $model . '/' . $options['filename'],
+						$index
+					),
 				$model instanceof File
 					=> $model->root(),
 				default
 				=> $model->file($options['filename'])->root()
 			};
 
-			// prevent path traversal
-			$root = F::realpath($source, $index);
-
 			// generate the thumbnail and save it in the media folder
-			$kirby->thumb($root, $thumb, $options);
+			$kirby->thumb($source, $thumb, $options);
 
 			// remove the job file once the thumbnail has been created
 			F::remove($job);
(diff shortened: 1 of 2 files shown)


afbora:
@lukasbestle We don't need to change it for the directory either, right? I mean following line:
https://github.com/getkirby/kirby/blob/fix/media-realpath-regression/src/Cms/Media.php#L101-L115

lukasbestle:
@afbora That one checks for the file being inside `$media`, which is the case for all three variants.

bastianallgeier:
I've just tested it with the new composerkit and it works great. I had tested original images yesterday with it and couldn't reproduce it, but of course it was also affected for thumb files. Now it's all working as epxected again.

bastianallgeier (dismissed):

bastianallgeier (approved):

PR #7220, opened 2025-05-14
Enhanced Preview Buttons and Preview View

bastianallgeier:
## Description

This PR refactors the new Preview View in order to clean up the UX and simplify how to preview pages.

### Enhancement

- New `git-branch` icon

- The PreviewView component and backend code has been radically simplified and its UX has been improved. You can now switch between versions and languages easily. 
- The old open button is back in each model view and a new preview button has been added. 

### Breaking changes from previous betas

- The PreviewButton class now uses `preview` as icon and `Preview` as text. 
- All model views now use the new OpenButton instead of the PreviewButton and the refactored PreviewButton in addition to that. 
- The PreviewDropdownButton has been removed

## Ready?

### For review team


lukasbestle (commented):
**What I like:**

- Displaying the form buttons depending on the active preview makes a lot of sense.
- It's definitely simpler than the previous UI.

**What I don't quite like:**

- Having the version selector as a dropdown feels a bit cumbersome. Two clicks are needed to switch between changes and latest.
- Somehow I miss the side-by-side view as I wrote on Discord already.

**Suggestion:**

Maybe the version selector could be a button group with:

> Latest | *Layout columns icon* | Changes

Then one could easily switch between the versions and with the middle button use the side-by-side view. Then we would no longer need all those previous buttons to shrink/expand the individual preview panes but the side-by-side preview would still survive. This of course only works as long as there are only ever two versions (latest and changes), but we will need and want to redesign the view later again anyway. This simple control would IMO make the new changes feature easier to understand and use in 5.0.

config/areas/site/buttons.php
```diff
@@ -3,30 +3,57 @@
...
+use Kirby\Panel\Ui\Buttons\OpenButton;
 use Kirby\Panel\Ui\Buttons\PageStatusButton;
-use Kirby\Panel\Ui\Buttons\PreviewDropdownButton;
+use Kirby\Panel\Ui\Buttons\PreviewButton;
 use Kirby\Panel\Ui\Buttons\SettingsButton;
+use Kirby\Panel\Ui\Buttons\VersionsButton;
 
 return [
+	'site.open' => function (Site $site, string $versionId = 'latest') {
+		if ($site->previewUrl() !== null) {
```
distantnative: `Question:` Shouldn't we pass the version ID here as well?

```php
		if ($previewUrl = $site->previewUrl($versionId)) {
```

bastianallgeier: I refactored it some more.

distantnative (approved):

PR #7229, opened 2025-05-22
fix: Save title and slug in changes if they exist

### LINKED ISSUE #7227: [v5] Changed title is reverted back to previous version after saving
## Description
If I have unsaved changes and then update the title, the "Changed version" still shows the old title. The "Latest version" shows the new title. After saving the changes, the old title reappears.

See videos:
https://github.com/user-attachments/assets/faa45d01-09e7-4f38-a90e-8c84e49f0884
https://github.com/user-attachments/assets/036871a0-21ec-49ac-ab1b-d40e2c5890d6

## To reproduce
In the starterkit:
1. Open notes article
2. Change text
3. Change title
4. Preview changes (optionally)
5. Save changes

## Your setup
**Kirby Version**  
5.0.0 rc1

**Your system**
Chrome macOS 15.5

bastianallgeier:
### Enhancements

- `page.changeTitle` and `page.changeSlug` hooks receive a new optional `$language` parameter, which passes a full Language object. 

### Fixes

- When changing the title or slug (in a secondary language), the `::changeTitle()` and `::changeSlug()` methods will check for an existing `changes` version and store them there as well. Otherwise, the changed title or slug would get reverted as soon as the `changes` version is published.

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/PageActions.php
+++ b/src/Cms/PageActions.php
@@ -10,7 +10,6 @@ use Kirby\Content\VersionId;
 use Kirby\Exception\DuplicateException;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Exception\LogicException;
-use Kirby\Exception\NotFoundException;
 use Kirby\Filesystem\Dir;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\I18n;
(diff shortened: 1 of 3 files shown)


lukasbestle (approved):
At first I thought the `changeSlug` behavior was inconsistent between default and secondary languages because one still updates the file system and the other updates the content files. But it makes sense: These actions take immediate effect on the latest version and are not only performed when the version is published. So for the default language we don't need to do anything special with the changes content.

`Feedback`: I really like the refactorings to use `Language::ensure()`. Would have been great to put these into a separate first commit though for easier review and better overview in the Git history later.

`Todo`: I think we need to extend the changelog snippet with an enhancement that the `changeSlug` and `changeTitle` actions now receive an additional `$language` argument.

bastianallgeier:
@lukasbestle about your feedback. You are totally right about the separate commit. I sometimes still struggle to clean up the Git history properly. Still need to learn some more tricks here.

PR #7239, opened 2025-05-23
fix: Store site title updates in the `changes` version if the version exists. 

bastianallgeier:
## Sandbox

There's now a test setup for site content changes in the sandbox. Pull the latest changes, switch to the Lab environment and visit https://sandbox.test/panel/site?tab=content

### Enhancements

- New optional `$language` arg is passed to the `site.changeTitle` hooks

### Fixes

- fix: Store site title updates in the `changes` version if the version exists. #7227 

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Cms/SiteActions.php
+++ b/src/Cms/SiteActions.php
@@ -50,12 +50,13 @@ trait SiteActions
 			'site'         => $this,
 			'title'        => trim($title),
 			'languageCode' => $languageCode,
+			'language'     => $language
 		];
 
 		return $this->commit(
 			'changeTitle',
 			$arguments,
-			fn ($site, $title, $languageCode) => $site->save(['title' => $title], $language->code())
+			fn ($site, $title, $languageCode, $language) => $site->save(['title' => $title], $language->code())
 		);
 	}
 
--- a/src/Cms/SiteActions.php
+++ b/src/Cms/SiteActions.php
@@ -44,23 +44,18 @@ trait SiteActions
 		string $title,
 		string|null $languageCode = null
 	): static {
-		// if the `$languageCode` argument is not set and is not the default language
-		// the `$languageCode` argument is sent as the current language
-		if (
-			$languageCode === null &&
-			$language = $this->kirby()->language()
-		) {
-			if ($language->isDefault() === false) {
-				$languageCode = $language->code();
-			}
-		}
+		$language = Language::ensure($languageCode ?? 'current');
 
-		$arguments = ['site' => $this, 'title' => trim($title), 'languageCode' => $languageCode];
+		$arguments = [
+			'site'         => $this,
+			'title'        => trim($title),
+			'languageCode' => $languageCode,
+		];
 
 		return $this->commit(
 			'changeTitle',
 			$arguments,
-			fn ($site, $title, $languageCode) => $site->save(['title' => $title], $languageCode)
+			fn ($site, $title, $languageCode) => $site->save(['title' => $title], $language->code())
 		);
 	}
 
--- a/src/Cms/SiteActions.php
+++ b/src/Cms/SiteActions.php
@@ -53,11 +53,16 @@ trait SiteActions
 			'language'     => $language
 		];
 
-		return $this->commit(
-			'changeTitle',
-			$arguments,
-			fn ($site, $title, $languageCode, $language) => $site->save(['title' => $title], $language->code())
-		);
+		return $this->commit('changeTitle', $arguments, function ($site, $title, $languageCode, $language) {
+
+			// make sure to update the title in the changes version as well
+			// otherwise the new title would be lost as soon as the changes are saved
+			if ($site->version('changes')->exists($language) === true) {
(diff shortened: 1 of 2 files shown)


src/Cms/SiteActions.php
```diff
@@ -44,24 +44,25 @@ public function changeTitle(
...
-			if ($language->isDefault() === false) {
-				$languageCode = $language->code();
-			}
-		}
+		$language = Language::ensure($languageCode ?? 'current');
 
-		$arguments = ['site' => $this, 'title' => trim($title), 'languageCode' => $languageCode];
+		$arguments = [
+			'site'         => $this,
+			'title'        => $title,
```
lukasbestle: Why did we trim the title before here but not in `$page->changeTitle()`? I think we should not make this change without checking the possible impacts, putting it into a separate commit and mentioning it in the changelog.

You added this trimming for Fiber in 3.6.0 in this commit: https://github.com/getkirby/kirby/commit/9c329fc580fb4b949188af77e1d2c561f855032f
Unfortunately I can't tell the reason. Do you remember?

bastianallgeier: Ugh, good catch. TBH Cursor just replaced that block because it's basically the same and that was one of those moments where the AI autocomplete felt pretty magical. I didn't even see the trim method. Not good.

But it's still weird that we have it in the site dialog but not for pages. It's also not great that removing the trim method didn't cause any trouble in the tests. We should definitely clean all of this up. I just wonder if we shouldn't add trim to the page title dialog as well.

lukasbestle: First we need to know why we added it in 3.6. :)

lukasbestle:
`General feedback`: I think we shouldn't split a change up into an implementation commit (`feat`/`fix`) and a separate `test` commit. If we do that, the tests for the implementation commit will fail as they are not adapted yet. But for any commit the tests should ideally pass (to enable things like `git bisect` scripting). Also by separating the commits, we will lose the connection in the Git history of why the test changes were necessary.

TL;DR: I think the `test` commit type is only meant for isolated testing fixes and enhancements that are not connected to a code change.

src/Cms/SiteActions.php
bastianallgeier: Ha, turns out we don't need this at all. We already trim the value in `Kirby\Data\Txt::encodeResult()` There's no difference if we add trim here. The only breaking change is in the hooks, when you expect the value to be trimmed, because the trimmed title was passed as argument. I've added the trim method again to keep the old behavior. I think we need to clean this up more carefully in v6.

bastianallgeier:
@lukasbestle good points! I've refactored the commit history and squashed the commits that belong together.

src/Cms/SiteActions.php
distantnative: Didn't we add it because the trimming in data only happens after validations/rules checking for e.g. title length?

bastianallgeier: That might indeed be a good reason. Also another reason why we might already want to add it Page::changeTitle as well in another PR. 

@distantnative it's a bit misleading because the diff still shows it to be removed, but it's back.

src/Cms/SiteActions.php
```diff
@@ -44,24 +44,25 @@ public function changeTitle(
...
-				$languageCode = $language->code();
-			}
-		}
+		$language = Language::ensure($languageCode ?? 'current');
 
-		$arguments = ['site' => $this, 'title' => trim($title), 'languageCode' => $languageCode];
+		$arguments = [
+			'site'         => $this,
+			'title'        => trim($title),
+			'languageCode' => $languageCode,
```
lukasbestle, line 52: `Issue`: In the previous code, we have overwritten the `$languageCode` variable if the language object was not the default language. Now we leave it unchanged, which passes `$languageCode = null` as argument to `commit`, which was previously only case for the default language. Not sure how we can even replicate this behavior with the new language setup.

Just realized we have the same regression in the page PR #7229 as well.

lukasbestle (commented):
About the `trim()` call: You are both right. At least for the `PlainTextStorage`, the only relevant real-world effects are the `before` hook and the `SiteRules` validation.

For other storage implementations this is not necessarily the case though.

But this leads me to another conclusion: Why do we do it for the title but not for any other field? Does the title have any special properties that other fields don't have in regards to whitespace in its value? If not, we could also think about trimming all fields in `$version->prepareFieldsBeforeWrite()`.

I think after v5 we should get back to the drawing board on this topic. It's not urgent enough of a topic to still change it in 5.0.

src/Cms/SiteActions.php
bastianallgeier: Good catch. TBH I would refactor this and simply use `$language->code()` This is by far the leanest and most stable solution to always pass a reliable language code. Otherwise, those methos need to introduce language logic again, which we really don't want and which is now so nicely solved by `Language::ensure()`. But maybe we can do this in another PR that solves it for both methods?

PR #7252, opened 2025-05-28
fix: Wait for upload and refresh the view afterwards

### LINKED ISSUE #7249: [v5] Uploading a file by dragging it over the textarea field doesn't work
## To reproduce

1. Go to panel 
2. Edit any page that have textarea field
3. Drag and drop a file into textarea field from your device to upload
4. See that file drag text will be disappear

## Related issue: 

Fabian's markdown editor has this issue:
https://github.com/fabianmichael/kirby-markdown-field/issues/196

## Your setup
**Kirby Version**  
5.0.0-rc.2

bastianallgeier:
### Enhancements

- New async `TextareaInput.restoreSelection()` method 
- New `panel.upload.announce()` method, which sends the success notification and `model.update` event after uploads have successfully completed. 
- The upload module now sends a `file.upload.error` event for every file that could not be uploaded.
- Various TextareaInput methods have been turned into async methods for more stable async/await control throughout all TextareaInput methods.
  - `insert`
  - `insertFile`
  - `insertUpload`
  - `prepend`
  - `toggle`
  - `wrap`

### Fixes

- `panel.content.save()` now calls `panel.content.cancelSaving()` to make sure that any old scheduled save requests are properly cancelled. This could have lead to potential race conditions before.
- The `FilesField` no longer sends unnecessary `file.upload` and `model.update` emitters after upload. Those lead to duplicate section reload requests so far. After uploads, the `panel.upload.done()` already takes care of reloading the current view and all its sections. 
- The `TextareaInput.insertUpload()` method now uses `await panel.content.update()` to push text value changes directly to the changes API and to reload the view correctly. This fixes https://github.com/getkirby/kirby/issues/7249
- The upload module sends the `file.upload` event directly when a file has been uploaded and passes the file object correctly. Before this change, the file.upload event was only called in the FilesField and FilesSection components when all uploads finished. 
- Both the `upload.done()` and `upload.cancel()` handlers now properly emit the `complete` event if there are any completed uploads. 

### Refactoring

- `TextareaInput.selectionRange` and `TextareaInput.restoreSelection()` are used to replace the old `restoreSelectionCallback` method in dialog events for better async control.
- Removed unnecessary complete handler in `FilesSection`. The upload module already takes care of the notifications and events. 

### Deprecated

- `TextareaInput.restoreSelectionCallback()` Use `TextareaInput.restoreSelection()`.

### Breaking changes

None 

## Sandbox

There's a new "Uploads" tab in the Textarea setup in the sandbox. 

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Input/TextareaInput.vue
+++ b/panel/src/components/Forms/Input/TextareaInput.vue
@@ -346,8 +346,8 @@ export default {
 		upload() {
 			this.$panel.upload.pick(this.uploadOptions);
 		},
-		wrap(before, after) {
-			this.insert(before + this.selection() + (after ?? before));
+		async wrap(before, after) {
+			return this.insert(before + this.selection() + (after ?? before));
 		}
 	}
 };
--- a/panel/src/components/Forms/Input/TextareaInput.vue
+++ b/panel/src/components/Forms/Input/TextareaInput.vue
@@ -103,7 +103,7 @@ export default {
 					cancel: async () => await this.setSelectionRange(selectionRange),
 					done: async (files) => {
 						await this.setSelectionRange(selectionRange);
-						this.insertUpload(files);
+						await this.insertUpload(files);
 					}
 				}
 			};
@@ -200,8 +200,8 @@ export default {
 				await this.insert(files.map((file) => file.dragText).join("\n\n"));
 			}
 		},
-		insertUpload(files) {
-			this.insertFile(files);
+		async insertUpload(files) {
+			await this.insertFile(files);
 			this.$events.emit("model.update");
 		},
 		onCommand(command, ...args) {
--- a/panel/src/components/Forms/Input/TextareaInput.vue
+++ b/panel/src/components/Forms/Input/TextareaInput.vue
@@ -329,7 +329,7 @@ export default {
 			this.$refs.input.setSelectionRange(start, end);
 			return this.$nextTick();
 		},
-		toggle(before, after) {
+		async toggle(before, after) {
 			after = after ?? before;
 			const selection = this.selection();
 
--- a/panel/src/components/Forms/Input/TextareaInput.vue
+++ b/panel/src/components/Forms/Input/TextareaInput.vue
@@ -160,7 +160,7 @@ export default {
 					submit: async (file) => {
 						this.$panel.dialog.close();
 						await this.setSelectionRange(selectionRange);
-						this.insertFile(file);
+						await this.insertFile(file);
 					}
 				}
 			});
@@ -195,9 +195,9 @@ export default {
 
 			return input.value;
 		},
-		insertFile(files) {
+		async insertFile(files) {
 			if (files?.length > 0) {
-				this.insert(files.map((file) => file.dragText).join("\n\n"));
+				await this.insert(files.map((file) => file.dragText).join("\n\n"));
 			}
 		},
 		insertUpload(files) {
--- a/panel/src/components/Forms/Input/TextareaInput.vue
+++ b/panel/src/components/Forms/Input/TextareaInput.vue
@@ -41,6 +41,7 @@
 				@dragover="onOver"
 				@dragleave="onOut"
 				@drop="onDrop"
+				@selectionchange="onSelectionChange"
 			/>
 		</div>
 	</div>
@@ -89,20 +90,19 @@ export default {
 	emits: ["focus", "input", "submit"],
 	data() {
 		return {
-			over: false
+			over: false,
+			selectionRange: null
 		};
 	},
 	computed: {
 		uploadOptions() {
-			const selectionRange = this.selectionRange();
-
 			return {
 				url: this.$panel.urls.api + "/" + this.endpoints.field + "/upload",
 				multiple: false,
 				on: {
-					cancel: async () => await this.setSelectionRange(selectionRange),
+					cancel: async () => await this.restoreSelection(),
 					done: async (files) => {
-						await this.setSelectionRange(selectionRange);
+						await this.restoreSelection();
 						await this.insertUpload(files);
 					}
 				}
@@ -129,26 +129,22 @@ export default {
 	},
 	methods: {
 		dialog(dialog) {
-			const selectionRange = this.selectionRange();
-
 			this.$panel.dialog.open({
 				component: "k-toolbar-" + dialog + "-dialog",
 				props: {
 					value: this.parseSelection()
 				},
 				on: {
-					cancel: async () => await this.setSelectionRange(selectionRange),
+					cancel: async () => await this.restoreSelection(),
 					submit: async (text) => {
 						this.$panel.dialog.close();
-						await this.setSelectionRange(selectionRange);
+						await this.restoreSelection();
 						await this.insert(text);
 					}
 				}
 			});
 		},
 		file() {
-			const selectionRange = this.selectionRange();
-
 			this.$panel.dialog.open({
 				component: "k-files-dialog",
 				props: {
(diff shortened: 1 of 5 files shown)


afbora:
@bastianallgeier Still doesn't work :(

bastianallgeier:
That's very weird. Works perfectly for me now. Ok, then I need to double-check again.

distantnative:
It does work for me. @afbora sure there aren't any old files cached, e.g. in `media/panel` if you're running it via `npm run build` not `npm run dev`?

afbora:
@distantnative I'm testing always via `build`. The issue still valid for me. Sometimes works sometime not. With this PR better but it's not completely fixed.

distantnative:
@afbora Very weird. Any insights on the preconditions to have it succeed vs. when it fails?

afbora:
@distantnative I can reproduce it with every request via browser's slow 4G or 3G network throttling feature. 

Not sure but I think this content lost happens when pages/field/files request done after save request.

distantnative:
@afbora You mention "save request" - so it's not just uploading what you are trying to do (as described in the original issue) but some combination with uploading and directly saving (and if you say slow connection, maybe you are trying to super quickly save before the upload request is actually done)?

But then I would think this fix might still be right, but we have in addition still problems with a race condition between `content.save` and `content.publish` (or similar).

afbora:
I mean  `/save` request after uploading done (automaticly and after modal closed) with "save request" not page save request via manual clicking.

distantnative:
Mmm ok. Couldn't reproduce it. Even when throttling speed.

One thought could be if `insertFile` takes longer than the update call. There is a setTimeout involved. So maybe that.

afbora:
@distantnative Here the screencast:

https://github.com/user-attachments/assets/92da57b9-fdaf-41b3-8909-ad5f3736b790

bastianallgeier:
I think our main issue really is the view reloading in the done event of the uploader. I have an idea.

bastianallgeier:
@distantnative @afbora I've refactored the code quite a bit and I hope this PR is not getting too big. 

- We had a couple places in the TextareaInput component that made race conditions possible. I've refactored them with async/await to be more in control of what we actually do
- The upload module had a few weird inconsistencies with event emitting. I've cleaned it up to better see what we are doing at which point
- We only need to emit the `model.update` event once all file uploads are finished. This will trigger all sections to be reloaded. A full `panel.view.reload()` is not necessary here and will always lead to issues with open drawers. 

## Sandbox

I've added a new "Others > Uploads" examples page where we can test textarea, structure, blocks and files fields with uploads and we can also see if a files section updates accordingly.

afbora:
@bastianallgeier I've tested now. There is restore selection issue at first glance. 

Click/select to random place of the paragraph text and try to drag/drop items to selected area. You can repeat this with selecting random places of the text with drag/drop an item. You'll see the issue that dragtexts are inserted always same places instead of selected area.

panel/src/components/Forms/Input/TextareaInput.vue
```diff
@@ -154,51 +156,57 @@ export default {
...
 		focus() {
 			this.$refs.input.focus();
 		},
-		insert(text) {
+		async insert(text) {
 			const input = this.$refs.input;
 			const current = input.value;
 
 			if (typeof text === "function") {
 				text = text(this.$refs.input, this.selection());
```
distantnative:
```suggestion
				text = text(input, this.selection());
```

panel/src/components/Forms/Input/TextareaInput.vue
```diff
@@ -291,36 +299,44 @@ export default {
...
+		},
+		selectionRange() {
+			return {
+				start: this.$refs.input.selectionStart,
+				end: this.$refs.input.selectionEnd
+			};
+		},
+		async setSelectionRange({ start, end }) {
+			this.$refs.input.setSelectionRange(start, end);
+			return this.$nextTick();
```
distantnative:
```suggestion
			await this.$nextTick();
```

`Suggestion:` I feel we do not actually want to return the promise - this method probably should run as void, shouldn't it? But await the next tick before the method resolves.

panel/src/panel/upload.js
```diff
@@ -74,12 +78,9 @@ export default (panel) => {
 			panel.dialog.close();
 
 			if (this.completed.length > 0) {
+				await this.emit("complete", this.completed);
 				await this.emit("done", this.completed);
-
-				if (panel.drawer.isOpen === false) {
```
distantnative, line 79: `Test:` Adding this also as a todo for myself to test if it still works properly inside drawers now this guard has been removed.

panel/src/components/Forms/Input/TextareaInput.vue
```diff
@@ -291,36 +299,44 @@ export default {
...
 		},
-		toggle(before, after) {
+		async toggle(before, after) {
 			after = after ?? before;
```
distantnative:
```suggestion
			after ??= before;
```

panel/src/components/Forms/Input/TextareaInput.vue
```diff
@@ -337,8 +353,8 @@ export default {
 		upload() {
 			this.$panel.upload.pick(this.uploadOptions);
 		},
-		wrap(before, after) {
-			this.insert(before + this.selection() + (after ?? before));
+		async wrap(before, after) {
+			return this.insert(before + this.selection() + (after ?? before));
```
distantnative:
```suggestion
		async wrap(before, after) {
			after ??= before;		
			await this.insert(before + this.selection() + after);
```

panel/src/components/Forms/Input/TextareaInput.vue
```diff
@@ -291,36 +299,44 @@ export default {
...
+			return async (callback) => {
+				await this.setSelectionRange(selectionRange);
 
-					if (callback) {
-						callback();
-					}
-				});
+				if (callback) {
+					callback();
+				}
```
distantnative, line 327: `Question:` Isn't this losing now the implicit next tick that `setTimeout` would add before?

distantnative (commented):
I also experienced what @afbora writes. Not always consistently but multiple times the selection would e.g. not be restored when canceling an upload or other dialog.

I think the issue is that the `selectionRange` gets stored inside the computed property. But the computed property actually does not get refreshed each time the selection changes. So it ends up restoring always the first state when the computed property got generated.

panel/src/panel/upload.js
bastianallgeier: You can use https://sandbox.test/panel/pages/other+uploads?tab=blocks to test this.

panel/src/components/Forms/Input/TextareaInput.vue
bastianallgeier: The nextTick is now handled in restoreSelection

bastianallgeier:
Your tests and comments made me realize that I had a logical issue in the way the selectionRange was stored but not updated when dialogs are opened. This makes no sense and leads to the issues that you describe. Then I realized that we can use the `onselectionchange` event of the input perfectly to always store the latest selection range and restore that one if it exists. This works really well and cleans up the code some more. 

I've tested everything in Firefox, Chrome and Safari and it seems to run smoothly now, but I would want to wait for another round of tests from your side.

afbora:
@bastianallgeier I tried hundreds of times in different ways, but on one of my first attempts, the file disappeared once. Then I couldn't reproduce the problem again. I think it was cache related. So everything seems to be working fine 🎉

bastianallgeier:
@afbora that's great. I think we should wait for another look from @distantnative before we give it a go. I guess it's ok if we put it into rc.4

distantnative:
Haven't had much chance to test it myself, but no further comments on the code and totally fine for me to have it merged if you two tested it successfully 👍

PR #7259, opened 2025-06-02
fix: Add missing submit handler

### LINKED ISSUE #7257: [v5] Save shortcut does not work when focus is in date field
## Description
When changing the date field by keyboard, the `CMD + S` shortcut does not work.

Image

## Expected behaviour
In Kirby 4, I can change a date field by keyboard and then use `CMD + S` to save.

## To reproduce
In the starterkit:

1. Go to a *Notes* article 
2. Click on *Date* field (either date or time input)
3. Change date or time by keyboard, for example: Delete day and type in a new day.
4. Press CMD + S to save

## Your setup
**Kirby Version**  
5.0.0-rc2

Chrome 137 on macOS 15.5 (I was only able to test macOS.)

bastianallgeier:
## Description

For some reason, we removed the submit handler in the Input component, which leads to a broken enter key in the date field. It's probably going to cause similar issues in other fields, but we haven't noticed yet. The fix is easy though :) 

### Fixed regression

- <kbd>Enter</kbd> works again as shortcut in the date field https://github.com/getkirby/kirby/issues/7257

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Input.vue
+++ b/panel/src/components/Forms/Input.vue
@@ -20,6 +20,7 @@
 					v-bind="inputProps"
 					:value="value"
 					@input="$emit('input', $event)"
+					@submit="$emit('submit', $event)"
 				/>
 			</slot>
 		</span>
@@ -53,7 +54,7 @@ export const props = {
 			default: null
 		}
 	},
-	emits: ["input"]
+	emits: ["input", "submit"]
 };
 
 export default {


distantnative:
@bastianallgeier have you already spotted where we removed it? Otherwise I'll search it. Just to be double sure that we didn't remove it for a reason (and then to fix that reason in another way).

bastianallgeier:
@distantnative I didn't check yet.

bastianallgeier:
https://github.com/getkirby/kirby/commit/e33f335a#diff-4d41cd4c1e66e7b266c144306c1acb06952c780eeaac30533b2ba26144d15136L68

by removing $listeners here, we lost the submit handler.

distantnative (approved):

PR #7269, opened 2025-06-05
fix: Add missing array check and fix invalid label value in the entries field

### LINKED ISSUE #7265: [v5] Console error for language translation dialog with entries field
## To reproduce

1. Add two languges,
2. Add translation key with  entries field from primary language,
3. Go to seconary language and edit the that key,
4. See console error: TypeError: Cannot read properties of null (reading 'map')

## Your setup
**Kirby Version**  
5.0.0-rc.3

**Console output**  

> TypeError: Cannot read properties of null (reading 'map')

bastianallgeier:
### Fixes

- Passing `false` as label value to `<k-field>` will no longer throw a console error. We use this in the new `<k-entries-field>` to disable labels for nested fields.  
- The new `<k-entries-field>` checks for a valid array as given value and will no longer throw a console error if `null` is passed. https://github.com/getkirby/kirby/issues/7265

### Breaking changes

None

### For review team

### DIFF
--- a/panel/src/components/Forms/Field/EntriesField.vue
+++ b/panel/src/components/Forms/Field/EntriesField.vue
@@ -229,10 +229,7 @@ export default {
 	watch: {
 		value: {
 			handler(entries) {
-				if (Array.isArray(entries) === false) {
-					this.entries = [];
-					return;
-				}
+				entries ??= [];
 
 				// no need to add ids again if the values are the same
 				if (entries === this.values) {
--- a/panel/src/components/Forms/Field/EntriesField.vue
+++ b/panel/src/components/Forms/Field/EntriesField.vue
@@ -229,6 +229,11 @@ export default {
 	watch: {
 		value: {
 			handler(entries) {
+				if (Array.isArray(entries) === false) {
+					this.entries = [];
+					return;
+				}
+
 				// no need to add ids again if the values are the same
 				if (entries === this.values) {
 					return;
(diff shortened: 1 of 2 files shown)


panel/src/components/Forms/Field/EntriesField.vue
```diff
@@ -230,7 +230,7 @@ export default {
 		value: {
 			handler(entries) {
 				// no need to add ids again if the values are the same
-				if (entries === this.values) {
+				if (entries === this.values || Array.isArray(entries) === false) {
```
distantnative: `Question:` Based on the `value` prop type this should only be an array. So I would think the only other value but an array is undefined/null. Which then we say defaults to an empty array in the prop definition. Which is then why I am wondering if instead of ignoring this new input here, we should rather set `this.entries = []`?

bastianallgeier: I've changed it.

panel/src/components/Forms/Field/EntriesField.vue
```diff
@@ -229,6 +229,11 @@ export default {
 	watch: {
 		value: {
 			handler(entries) {
+				if (Array.isArray(entries) === false) {
+					this.entries = [];
+					return;
+				}
+
```
distantnative:
```suggestion
				entries ??= [];

```

`Suggestion:` Shouldn't this be enough? Then we could still benefit from the `entries === this.values` check?

distantnative (approved):

PR #7272, opened 2025-06-07
fix: Move trait template annotations to class doc

distantnative:
### Summary of changes
- Move `@use MyTrait<MyTemplateType>` to class docblocks.

### Reasoning
IDEs recognize it in both positions, but our Reference code cannot access the annotation on the `use Trait` statement. That's why we move it to the class docblock.

## Ready?

### DIFF
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -34,6 +34,9 @@
 
 		<!-- Don't warn about duplicate array keys as Psalm will throw false positive when unpacking arrays -->
 		<DuplicateArrayKey errorLevel="suppress" />
+
+		<!-- Don't warn about missing template params as Psalm does not recognize this correctly when using `@uses` for traits in class docblock -->
+		<MissingTemplateParam errorLevel="suppress" />
 	</issueHandlers>
 
 	<plugins>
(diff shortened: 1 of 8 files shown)


bastianallgeier:
I'm not very involved in the reflection stuff for the reference. Just a throught. Could they check for Trait classes and get annotations from there with something like https://www.php.net/manual/en/reflectionclass.gettraits.php ?

distantnative:
@bastianallgeier Not really. That gets use the `ReflectionClass` objects of the traits themselves. However, the data we provide here is not about the trait themself, but how the implementing class implements it (kinda). E.g. within the `File` class specifying that the collections used inside the `HasSbilings` trait are of the type `Files`.

```
@use \Kirby\Cms\HasSiblings<\Kirby\Cms\Files>
```

That we cannot annotate on the trait itself, because that depends (and varies) on the class that uses the trait.

The Type in the `<>` is interpreted for the `TCollection` template in the trait (https://github.com/getkirby/kirby/blob/v5/develop/src/Cms/HasSiblings.php#L17). Which we use in the trait to more specifically type parameters or return (e.g. https://github.com/getkirby/kirby/blob/v5/develop/src/Cms/HasSiblings.php#L102-L103).

The combination of this results in IDEs understanding that e.g. `File::nextAll()` accepts not any Collection, but a Files collection as argument and returns not any Collection but a Files collection.

bastianallgeier (approved):

PR #7279, opened 2025-06-10
Collapsed button group: remove coloured borders on buttons

adamkiss:
### Summary of changes

Removes border colour on collapsed buttons, regardless of theme (`*`, `*-icon`).

### Reasoning

I think it looks better.

CleanShot 2025-06-10 at 16 04 38@2x
CleanShot 2025-06-10 at 16 03 40@2x

### Additional context

I'm super cool if you decide against this.

### Fixes

Weirdly coloured lines that look out of place.

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Navigation/ButtonGroup.vue
+++ b/panel/src/components/Navigation/ButtonGroup.vue
@@ -84,9 +84,10 @@ export default {
 	border-end-end-radius: 0;
 }
 
-.k-button-group[data-layout="collapsed"] > .k-button:not([data-theme]) {
+.k-button-group[data-layout="collapsed"] > .k-button {
 	--theme-color-border: var(--panel-color-back);
 }
+
 .k-button-group[data-layout="collapsed"]
 	> .k-button[data-variant="filled"]:not(:first-child) {
 	border-start-start-radius: 0;


distantnative:
First of all, I agree with you that it looks bad. Especially with `*-icon` theme. 

I am wondering how it looks after your change of all buttons in the group have the same theme. Do we then have light stripes in between?

If so, maybe we need to solve it differently. Only have it colored when full theme, not icon theme.

adamkiss:
@distantnative 
CleanShot 2025-06-10 at 16 29 14
CleanShot 2025-06-10 at 16 28 46

## "looks better" (opinions ahead)
- `*` same color - **coloured border** (no border is acceptable)
- `*` diff colors - **no border** (coloured border is not good)
- `*-icon` same color - **no border** (coloured border is not good)
- `*-icon` diff colors - **no border** (coloured border is not good)

I think if you look at all the states, no border is the best all around IMO. I can imagine people having large amount of buttons grouped by color perhaps, but I like to do separate colors = separate function, and I'd bet I'm not alone.

distantnative:
@adamkiss Fair point. I am wondering if there would be some elaborate way in CSS to achieve a rule that only targets those themed elements (`*`) that are following a themed element with the same theme. If we cannot find that, I probably would agree with you to go without.

But also curious to hear what @bastianallgeier thinks about it.

bastianallgeier:
I like that the change is so minimal and the result does indeed look a lot better. It's a bit of a shame when the buttons have a coloured backround and all are the same. I liked our coloured borders there as well, but I think it's a small trade-off to get rid of them there as well in comparison to the massive enhancement in all other cases.

adamkiss:
> I liked our coloured borders there as well

I've been looking at it since I PRed it and I like non-coloured buttons more and more, and I'm starting to think, that the original coloured borders were a reaction to (you, or us designers generally) missing colours in modern ui 😂

bastianallgeier:
I've added a lab example on `pullrequests/adamkiss/v5/collapsed-button-group-border-color` and I have to say that I still miss the colored borders a tiny bit for our form buttons, but everywhere else I prefer the ones without the colored borders. Especially in dark mode …

Screenshot 2025-06-13 at 10 28 20

bastianallgeier:
@distantnative what are your thoughts?

distantnative:
I am fine with the suggestion. 
I guess creating CSS rules to add a colored border between two buttons of the same theme could get a bit complex.
But I didn't see your lab example on the PR?

distantnative:
Or maybe instead of no border (same color as background) one does for filled buttons a slightly darker gray than the background, lighter than the normal filled buttons?

bastianallgeier:
@distantnative I can't contribute to the PR directly. The added example is here: https://github.com/getkirby/kirby/blob/pullrequests/adamkiss/v5/collapsed-button-group-border-color/panel/lab/components/buttons/5_groups/index.vue

Gray often collides with colors if the contrast isn't tuned perfectly. That's why it looks slightly shittier in light mode than in dark mode.

distantnative:
Ok fine for me 👍

PR #7283, opened 2025-06-12
fix: Allow submitting disabled fields

### LINKED ISSUE #7262: [v5] Field called "UUID" nested in structure field does not return value
## Description

After upgrading from Kirby `5.0.0-beta.4` to `5.0.0-rc.2`, it's not possible anymore to get the value of text fields called `uuid` nested inside structure fields when trying to access them inside `page.update:after` hook via `$newPage`.

_I couldn't find anything about this in the v5 changelog and hope I haven't overlooked any breaking change that I should've taken into account. If that's the case, feel free to close issue._

**Expected behavior**  
- It should be possible to access the value of field called `uuid` nested inside structure field.

## To reproduce

The fact that this does not seem possible anymore becomes an issue when e.g. trying to programmatically generate UUIDs for structure field  entries in a setup like this:

Blueprint:

```yaml
categories:
  type: structure
  fields:
    title:
      type: text
    uuid:
      type: text
      disabled: true
```

Hook:

```php
<?php

use Kirby\Data\Data;
use Kirby\Uuid\Uuid;

return [

		$data = $newPage->categories()->yaml();

		$data = array_map(function ($category) {
			if (empty($category['uuid'])) {
				$category['uuid'] = Uuid::generate();
			}
			return $category;
		}, $data);

		$newPage->save(['categories' => Data::encode($data, 'yaml')]);
		break;
	},
];
```

In this example `$category['uuid']` will always return an empty string and therefore always causes the UUIDs of all categories to be regenerated on every save.

## Your setup
**Kirby Version** 5.0.0-rc.2
**PHP Version** 8.4.7

**Console output**  

 - Device: MacBook Pro
 - OS: macOS 15.4.1 
 - Browser: Firefox 
 - Version: 138.0.4

**Additional context**

### LINKED ISSUE #7268: [v5] `$model->update()` has no effect on disabled fields
## Description
Calling $newPage->update() in page.update:after hook doesn't seem to update the page.

## To reproduce

// ... in config hooks ...

    'page.update:after' => function($newPage, $oldPage){

      $newPage->update([
            'test' => "Something"
          ]);
    },

## Your setup
**Kirby Version**  
5.0 rc-3

bastianallgeier:
### Enhancements

- New `$force` argument for `Kirby\Form\Fields::submit()` and `Kirby\Form\Form::submit()` to submit any value, even if the field is disabled, inactive or not translatable.

### Fixed regressions from previous pre-releases

- `Kirby\Cms\ModelWithContent::update()` will accept disabled fields again, unless `$validate` is switched on. https://github.com/getkirby/kirby/issues/7268

### Breaking changes

None

### Submitting any value 

Sometimes, you need to make sure that all values are submitted, even if the field is disabled or not translatable. The `$force` parameter for the submit method can help here. 

```php
use Kirby\Form\Form;

$form = new Form(
  fields: [
    'text' => [
      'type' => 'text',
      'disabled' => true
    ]
  ]
);

$form->submit(
  input: [
    'text' => 'This will be submitted'
  ],
  force: true
);
```

## Ready?

### For review team

### DIFF
--- a/src/Form/Fields.php
+++ b/src/Form/Fields.php
@@ -277,7 +277,10 @@ class Fields extends Collection
 	/**
 	 * Sets the value for each field with a matching key in the input array
 	 * but only if the field is not disabled
+	 *
 	 * @since 5.0.0
+	 * @param bool $passthrough If true, values for undefined fields will be submitted
+	 * @param bool $force If true, values for fields that cannot be submitted (e.g. disabled or untranslatable fields) will be submitted
 	 */
 	public function submit(
 		array $input,
--- a/src/Form/Fields.php
+++ b/src/Form/Fields.php
@@ -281,7 +281,8 @@ class Fields extends Collection
 	 */
 	public function submit(
 		array $input,
-		bool $passthrough = true
+		bool $passthrough = true,
+		bool $force = false
 	): static {
 		$language = $this->language();
 
@@ -294,8 +295,13 @@ class Fields extends Collection
 				continue;
 			}
 
-			// don't change the value of non-submittable fields
-			if ($field->isSubmittable($language) === false) {
+			// don't submit fields without a value
+			if ($force === true && $field->hasValue() === false) {
+				continue;
+			}
+
+			// don't submit fields that are not submittable
+			if ($force === false && $field->isSubmittable($language) === false) {
 				continue;
 			}
 
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -318,11 +318,13 @@ class Form
 	 */
 	public function submit(
 		array $input,
-		bool $passthrough = true
+		bool $passthrough = true,
+		bool $force = false
 	): static {
 		$this->fields->submit(
-			input:       $input,
-			passthrough: $passthrough
+			input: $input,
+			passthrough: $passthrough,
+			force: $force
 		);
 		return $this;
 	}
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -315,6 +315,8 @@ class Form
 	 * but only if the field is not disabled
 	 *
 	 * @since 5.0.0
+	 * @param bool $passthrough If true, values for undefined fields will be submitted
+	 * @param bool $force If true, values for fields that cannot be submitted (e.g. disabled or untranslatable fields) will be submitted
 	 */
 	public function submit(
 		array $input,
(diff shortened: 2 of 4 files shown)


distantnative:
`Thought:` Are we getting lost in complexity down this route? We already added `passthrough`, now `dirty`. Should we rather remove the isDisabled safeguard when so far this was more a UI option and not enforced when updating the objects? This feels like it moves us a bit away from simple form classes when we add these workarounds - maybe we just cannot enforce strict disabled field handling yet?

bastianallgeier:
@distantnative you are not wrong, but I would argue that passthrough and dirty are both form handling tools that help to solve different situations in ways that are well-known from other form libraries. Both help us to get to a point where we can get more strict. When we simply remove the disabled field check, we are basically getting less strict than in v4 and we would need to add that part again in some other way in our validation logic.

distantnative:
@bastianallgeier Honestly, I don't event understand what's exactly the difference between `passthrough` and `dirty`. And the naming with `any` and `passthrough` probably makes it even harder. I don't know that these are well-known concepts from other form libraries. I think we should find a way that this is easily understandable without having pre-knowledge from other form libraries.

bastianallgeier:
@distantnative ok, let's take more time for this and move it to rc.5

src/Form/Fields.php
```diff
@@ -281,7 +281,8 @@ public function reset(): static
 	 */
 	public function submit(
 		array $input,
-		bool $passthrough = true
+		bool $passthrough = true,
+		bool $force = false
```
distantnative, line 288: `Todo:` @bastianallgeier could you please DocBlock param descriptions for these two to explain (also for the docs) what they do and especially what they do differently?

bastianallgeier: @distantnative could you check if you find my comments understandable?

distantnative (approved):

src/Form/Form.php
```diff
@@ -318,11 +318,13 @@ public function strings($defaults = false): array
 	 */
```
lukasbestle, line 320: `Suggestion`: As a user of `$form->submit()` may not be aware about `$fields->submit()`:

```suggestion
	 *
	 * @param bool $passthrough If true, values for undefined fields will be submitted
	 * @param bool $force If true, values for fields that cannot be submitted (e.g. disabled or untranslatable fields) will be submitted
	 */
```

lukasbestle (commented):
I've added another unit test for the `date` field use case I explained yesterday. Just to make sure that we don't accidentally optimize this behavior away in a later release.

distantnative (approved):

PR #7288, opened 2025-06-13
fix: Bring back the old `Content::update()`

### LINKED ISSUE #7250: [v5] Assigning a value to a field no longer working
## Description
Assigning a value to a field no longer working, in PHP that is.

**Expected behavior**  
The assigned value should appear when used in a template.

## To reproduce

1. Start from starterkit, paste the code below into template home.php
2. Open page in browser
3. See output that says "title: 'Home'" instead of 'XXX'

```php
$page->title()->value = 'XXX';
$value = $page->title()->value();
echo "title: '$value'<br>\n";

$page->title()->value('XXX');
$value = $page->title()->value();
echo "title: '$value'<br>\n";
```
Two variants, both working in V4. The second variant is the prefered one according to https://getkirby.com/docs/reference/templates/field-methods/value

## Your setup

Starterkit with Kirby 5.0.0-rc.2

### LINKED ISSUE #7285: [v5] Problems with content()->update() deprecation
Hi, 
I'd like to share some consideration about the deprecation of these features in the new release. 
In your deprecation method you wrote: 

> $content->update()` is no longer functional. Please use `$model->version()->update()` instead

The thing is the previous method and the new one you are suggesting as alternative do two very different things. 
The old content update() allowed to modify the data only in memory, in the specific object instance that was modified, without storing anything on the disk, while the new suggested method actually store changes. There was already a different dedicated way to store changes, through page->update() method.

In our company we relied on the possibility to edit the in-memory content on the fly a lot, to implement several interesting mechanisms. We have conditional layouts and blocks, we inject custom blocks before rendering in certain situations, we inject missing default field values (extracted from blueprints) inside blocks (if you add a field in a later moment, currently existing content is not updated, making old and new content being inconsistent, and also default values are not always stored by kirby, it works only for certain field types), and so on. 

Now, I guess you could argue that modify objects in memory before using them was never encouraged or an intended feature by yourself, but we are talking about methods available in the public interface and nicely documented by you, we were not inventing strange workarounds to do this. Our assumption was that public and documented methods are there for a reason and reasonably safe to use. 

Don't get me wrong, I totally understand that during major versions releases the public interface can be broken, but there should be a very strong reason to do that and currently I'm struggling to see it in this case, since the two features (change an array of data only in memory and have dedicated methods to store the changes on disk) worked well together before and it seems to me they could continue to work well now even after adding the new versioning mechanism (I would just update in memory the current version content object), but of course I don't have a complete view on your processes, so maybe there is a reason that you see as valid enough to break things on user land.

I would just take the occasion to signal our discomfort. I had to subclass lot of stuff and look for workarounds to make our mechanisms work again, but I also wonder how long that could last until the next breaking change.

bastianallgeier:
## Description

We have a couple issues related to no longer being able to manipulate the content object in memory. One step to solve this, could be to bring back the `Content::update()` method. There's no harm to allow modifying the object in memory. But it still needs to be used differently after this change, so it would only a part of the solution. 

### Before

```php
$page->title();
// Foo

$page->content()->update([
  'title' => 'Bar'
]);

$page->title(); 
// Bar
```

### After 

```php
$page->title();
// Foo

$content = $page->content()->update([
  'title' => 'Bar'
]);

$page->title(); 
// Foo

$content->title();
// Bar
```

### Fixes

- Unbreak `Content::update()`

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Content/Content.php
+++ b/src/Content/Content.php
@@ -5,7 +5,6 @@ namespace Kirby\Content;
 use Kirby\Cms\Blueprint;
 use Kirby\Cms\File;
 use Kirby\Cms\ModelWithContent;
-use Kirby\Exception\Exception;
 use Kirby\Form\Form;
 
 /**
@@ -237,17 +236,18 @@ class Content
 	}
 
 	/**
-	 * Removed in 5.0.0. We only keep it to simplify debugging,
-	 * when this method is still in use in the wild.
-	 *
-	 * Use $model->version()->update() instead.
-	 *
-	 * @todo Remove in 7.0.0
+	 * Updates the content in memory.
 	 */
 	public function update(
 		array|null $content = null,
 		bool $overwrite = false
 	): static {
-		throw new Exception('`$content->update()` is no longer functional. Please use `$model->version()->update()` instead');
+		$content = array_change_key_case((array)$content, CASE_LOWER);
+		$this->data = $overwrite === true ? $content : array_merge($this->data, $content);
+
+		// clear cache of Field objects
+		$this->fields = [];
+
+		return $this;
 	}
 }
--- a/tests/Content/ContentTest.php
+++ b/tests/Content/ContentTest.php
@@ -3,7 +3,6 @@
 namespace Kirby\Content;
 
 use Kirby\Cms\Page;
-use Kirby\Exception\Exception;
 use PHPUnit\Framework\Attributes\CoversClass;
 
 #[CoversClass(Content::class)]
@@ -197,9 +196,33 @@ class ContentTest extends TestCase
 			'b' => 'B'
 		]);
 
-		$this->expectException(Exception::class);
-		$this->expectExceptionMessage('`$content->update()` is no longer functional. Please use `$model->version()->update()` instead');
+		$content->update([
+			'a' => 'aaa'
+		]);
 
-		$content->update();
+		$this->assertSame('aaa', $content->get('a')->value());
+
+		$content->update([
+			'miXED' => 'mixed!'
+		]);
+
+		$this->assertSame('mixed!', $content->get('mixed')->value());
+
+		// Field objects should be cleared on update
+		$content->update([
+			'a' => 'aaaaaa'
+		]);
+
+		$this->assertSame('aaaaaa', $content->get('a')->value());
+
+		$content->update($expected = [
+			'TEST' => 'TEST'
+		], true);
+
+		$this->assertSame(['test' => 'TEST'], $content->data());
+
+		$content->update(null, true);
+
+		$this->assertSame([], $content->data());
 	}
 }


distantnative:
I guess this alone could be fine - it allows updating in-memory cached content on object like blocks. At the same time, since models do not cache the content object anymore, updating that will immediately vanish again. Which is I think appropriate (as we probably don' want separate in-memory content vs. stored content anymore) but could of course also be a bit confusing.

bastianallgeier:
I agree that it is a little less practical from a user perspective for models in general, but could still be useful. If you use it in a controller for example, you could still pass a updated content object to the template. 


```php
<?php 

return function ($page) {
  
  $content = $page->content()->update([
    'text' => 'Updated text'
  ]);

  return [
    'content' => $content
  ];
};
```


```php
<?= $content->text()->kt() ?>
```

It would also still be quite useful in an extended model though. 

```php
use Kirby\Cms\Page;
use Kirby\Content\Content;

class MyPage extends Page 
{
  public function content(string|null $languageCode = null): Content
  {
    $content = parent::content($languageCode);

    return $content->update([
      'injectedField' => 'Some additional in-memory stuff'
    ]);
  }
}
```

I think at this point it's probably more about providing good docs.

lukasbestle:
I agree that supporting this makes sense. I think your last code examples are quite practical and don't conflict with other goals.

But maybe we could use the opportunity to rename the method. `update()` suggests it would update the real content (which it doesn't). A different name like `override()` could make the purpose and behavior clearer.

Since the need to return the updated content object is a breaking change anyway, changing the name is not such big of a deal. Quite the contrary, the rename will make uses of the old method clearly visible.

nilshoerrmann:
> But maybe we could use the opportunity to rename the method. update() suggests it would update the real content (which it doesn't). A different name like override() could make the purpose and behavior clearer.

In other places Kirby uses the word `virtual` for content that is not stored on disk.

bastianallgeier:
@lukasbestle I think bringing this back with the old method name would just reduce the upgrade pain for probably quite some people - even with the breaking change that you have to keep using the content object when working with models. All those who use it for structures and blocks would not have to change anything. We could maybe deprecate it in favor of another method immediately though.

distantnative (approved):

SkidX:
First of all thanks a lot for the effort and for listening about our use case scenarios.
As mentioned also in the other thread (by me or by @distantnative) there are a lot of different ways to work around the limitations in order to achieve similar results, so maybe here it would make more sense to talk about the API itself, more than our own solutions on top of that. 

From a user perspective, personally (I guess it could be different for other users) the only two things that justified to access directly the Content objects were the has() method and the update() method, but I find myself not using has() anymore since $page->fieldName()->isNotEmpty() is usually enough and more practical, so basically the only reason I still need to access the Content object directly is only when I want to modify the data in it (in memory). 

So, in this regard, I second what said by @distantnative above about having the object editable but getting a new copy everytime you access it: it's not only confusing, it would make it very impractical for me, because the whole point of using the internal Content object (and not inventing another layer of data in the page models subclassing the default Page) is to make the (eventual) modification completely transparent to snippets down in the line: I want my snippets at any level doing $page->text() to be completely unaware if that text string is coming from the disk or has been created / modified just an instant before in another part of the code. 
So, in order to make this whole in-memory editing usable for me, I would continue to cache the Content object on the page model similar to what @bastianallgeier showed above, I would just keep a local copy like this:
```php
class DefaultPage extends Page
{
	protected array $editableContents = [];

	public function content(string|null $languageCode = null): Content
	{
		// get the targeted language
		$language  = Language::ensure($languageCode ?? 'current');
		$languageCode = $language->code();

		if (!isset($this->editableContents[$languageCode])) {
			$this->editableContents[$languageCode] = parent::content($languageCode);
		}
		return $this->editableContents[$languageCode];
	}
}
```
I can imagine the reasons you wanted to get rid of keeping the local Content copy as before (cache invalidation is notoriously difficult and I can surely relate with your struggles there), but at the same time, for me the only practical way to modify that data on the fly implies to transparently access always the same copy of that data. 
So, I'm totally fine with doing the Content object memoization on my Page subclasses, the only thing, like @bastianallgeier pointed out, would be to make it very clear in the documentation that the default content() method has a different behavior from before, giving now always a freshly created object and not the same copy.

Thanks again for the great job you are doing!

lukasbestle:
@SkidX Your code example for an override of `content()` will certainly work, but please keep in mind that you are then also responsible for cache invalidation as you already write. As soon as the underlying content store changes (via `$page->update()` for example), your cached content object will be outdated.

We get the use case and wish we could provide a core solution that "just works", but we have really considered different options and cannot solve this in a robust way that covers all use cases.

Our official recommendation is to return either your custom content object or just the variables you need from the controller. Yes, you cannot access them "magically" via the `$page` object, but this approach has the added benefits that they add clarity to your setup and make unambiguous where a piece of content comes from (content file or dynamic controller code).

SkidX:
@lukasbestle, of course, in general, caching comes with that risk and implication. 

In the specific scenario you described though, that in which the underlying content store get changed, I think (but please correct me if I'm wrong) it does not imply any issue, since all those methods ($page->update() and so on), as explained by your docs, follow an "immutable" approach and return a new copy of the model object. If I access the new copy, the content is not cached yet (the first time) so I don't get access to outdated values. 
This code currently works as expected with my page model being modified to cache the content objects:
```php
echo $page->description()->value(); // whatever old value
$page->content()->edit(['description' => 'temp value']); // edit() could be update() after rc.5
echo $page->description()->value(); // 'temp value'
$newPage = $page->update(['description' => 'new value']);
echo $newPage->description()->value(); // new value, as expected
```
Of course, accessing the old $page object after calling update() on it would give me outdated values, but I think that was always the case even on versions 3, 4, and so on, since the old object does not get modified, by your design. 

So, I guess if you are not going to abandon your immutable approach on stored content changes, it should work as expected. 
I also don't think I would ever be in that specific scenario anyway, since I never mix page renderings (which is where I modify their in-memory content on the fly) with page on-disk modifications (Pages rendering happen on HTTP GET requests, it would really be an overall terrible idea to modify pages file content there).

PR #7295, opened 2025-06-15
Lab docs improvements

distantnative:
## Description

- Refactored the classes responsible for the Lab component docs into smaller single-focus classes
- Hide internal components from Lab docs
- Mark unstable components in Lab docs

Screenshot 2025-06-15 at 18 41 42

Screenshot 2025-06-15 at 18 41 24

### Enhancements
- Lab: internal components are hidden from docs; unstable components are marked as such

PR #7296, opened 2025-06-15
refact: Move since tag to own section in Lab docs

distantnative:
## Description

- [x] https://github.com/getkirby/kirby/pull/7295

### Summary
Moves the since tag information into its own simple section. This avoids displaying an empty description box if the component has no description but a since tag.

Screenshot 2025-07-08 at 22 09 31

### DIFF
--- a/config/areas/lab/drawers.php
+++ b/config/areas/lab/drawers.php
@@ -16,12 +16,14 @@ return [
 				];
 			}
 
+			$doc = Doc::factory($component);
+
 			return [
 				'component' => 'k-lab-docs-drawer',
 				'props' => [
 					'icon' => 'book',
 					'title' => $component,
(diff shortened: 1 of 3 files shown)


bastianallgeier (commented):
The positioning and sice of the since tag is still a bit off and it also shows up for components that don't have info about it

distantnative:
@bastianallgeier Not too happy with it either today... do you have any idea where to put it better?

bastianallgeier:
@distantnative maybe just a simple additional headline with the info below if since is defined? I don't think we need to be super fancy about it for now.

distantnative:
@bastianallgeier Made it a very simple section

bastianallgeier (approved):

Issue #7305, opened 2025-06-19
Writer field without nodes and marks still shows toolbar


SebastianEberlein-JUNO:
## Description
The writer field without nodes and marks still shows the toolbar, just with the paragraph button.
```yml
writer:
  type: writer
  marks: false
  nodes: false
```

<img width="604" alt="Image" src="https://github.com/user-attachments/assets/d5324855-ff2a-4605-81b9-c2a6144b2e9a" />

<img width="615" alt="Image" src="https://github.com/user-attachments/assets/cd337588-3a14-4cce-ae4e-26c49fd1670d" />

**Expected behavior**  
In Kirby 4, no toolbar is visible.

## To reproduce
Add the field shown above in the starterkit.

## Your setup
**Kirby Version**  
5.0.0-rc5

**Your system (please complete the following information)**
Chrome 137 on macOS 15.5

afbora:
@bastianallgeier Not sure but I think the https://github.com/getkirby/kirby/pull/7242 PR seems to fix one thing while breaking another. Toolbar is visible only with `inline: true`.

PR #7310, opened 2025-06-20
fix: Hides nodes in toolbar when nodes disabled

afbora:
## Description

I fixed the issue via solving passing nodes and headings props to toolbar component. I think these params were missing.

### Fixes
- [v5] Writer field without nodes and marks still shows toolbar #7305

### Breaking changes
None

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Input/WriterInput.vue
+++ b/panel/src/components/Forms/Input/WriterInput.vue
@@ -141,9 +141,14 @@ export default {
 			return {
 				// if custom set of marks is enabled, use as toolbar default as well
 				marks:
-					Array.isArray(this.marks) || this.marks === false
+					(Array.isArray(this.marks) || this.marks === false)
 						? this.marks
 						: undefined,
+				nodes:
+					(Array.isArray(this.nodes) || this.nodes === false)
+						? this.nodes
+						: undefined,
+				headings: this.headings,
 				...this.toolbar,
 				editor: this.editor
 			};
--- a/panel/src/components/Forms/Input/WriterInput.vue
+++ b/panel/src/components/Forms/Input/WriterInput.vue
@@ -93,7 +93,7 @@ export const props = {
 		},
 		nodes: {
 			type: [Array, Boolean],
-			default: () => ["heading", "bulletList", "orderedList"]
+			default: () => ["paragraph", "heading", "bulletList", "orderedList"]
 		},
 		paste: {
 			type: Function,
--- a/panel/src/components/Forms/Writer/Toolbar.vue
+++ b/panel/src/components/Forms/Writer/Toolbar.vue
@@ -26,6 +26,13 @@ export default {
 			type: Object
 		},
 		/**
+		 * Which headings to show in the toolbar
+		 */
+		headings: {
+			default: () => [1, 2, 3, 4, 5, 6],
+			type: [Array, Boolean]
+		},
+		/**
 		 * Whether the toolbar is displayed inline or as
 		 * a floating toolbar near the selection
 		 */
@@ -155,10 +162,10 @@ export default {
 				return available;
 			}
 
-			// get requested nodes from available entries
-			// if they are available
+			const expandedNodes = this.expandNodeTypes(this.nodes);
+
 			return Object.fromEntries(
-				this.nodes
+				expandedNodes
 					.filter((node) => available[node])
 					.map((node) => [node, available[node]])
 			);
@@ -296,6 +303,25 @@ export default {
 				label: entry.label,
 				click: () => this.command(entry.command ?? type)
 			};
+		},
+		/**
+		 * Helper method to expand node types
+		 * @returns {Array}
+		 */
+		expandNodeTypes(nodes) {
+			const expanded = [];
+
+			nodes.forEach((node) => {
+				if (node === 'heading') {
+					this.headings.forEach((level) => {
+						expanded.push(`h${level}`);
+					});
+				} else {
+					expanded.push(node);
+				}
+			});
+
+			return expanded;
 		},
 		/**
 		 * Creates an inline button object


bastianallgeier:
@distantnative I think you know more about the writer toolbar at this point to give a usable opinion on the solution.

panel/src/components/Forms/Writer/Toolbar.vue
```diff
@@ -297,6 +304,25 @@ export default {
...
+					this.headings.forEach((level) => {
+						expanded.push(`h${level}`);
+					});
+				} else {
+					expanded.push(node);
+				}
+			});
+
+			return expanded;
+		},
```
distantnative, line 325: `Todo:` I do not think we need to expand the headings here. The `Heading` node itself is registering multiple buttons depending on the heading level. Did this not work for you or why did you add this new helper method?

distantnative (commented):
I think we need to move this to 5.0.1 or so - there are some changes needed. See one comment I made, but also:

`Todo:` With this PR the paragraph node is gone from the dropdown when not having any `nodes` option set. I don't think this is intended or is it?

afbora:
> With this PR the paragraph node is gone from the dropdown when not having any nodes option set. I don't think this is intended or is it?

@distantnative Yes I know, because default nodes don't include the paragraph node:
https://github.com/getkirby/kirby/blob/5.0.0-rc.6/panel/src/components/Forms/Input/WriterInput.vue#L96

Maybe we can add it as default.

panel/src/components/Forms/Writer/Toolbar.vue
afbora: It is set as `heading` index in the `nodes` array. But in the `available` array, it is passed as "h1-h6" indexes instead of `heading`. So I had to convert `heading` to `h*` indexdes.

afbora:
I also agreed that this is regression but can be doable in next milestone since this is minor issue.

panel/src/components/Forms/Writer/Toolbar.vue
distantnative: But something is off here, the headings node entries show up in different order before/after the PR. I am not sure I can fully follow yet why this change is needed.

bastianallgeier:
Can we pick this up in a good way?

distantnative:
Maybe these two screenshots show what I meant with that something is off with the headings.

Default inline toolbar, no custom marks or nodes options set in the blueprint.

**Before this PR**
Screenshot 2025-07-14 at 20 30 05

**With this PR**
Screenshot 2025-07-14 at 20 29 46

panel/src/components/Forms/Input/WriterInput.vue
```diff
@@ -141,9 +141,14 @@ export default {
...
 				marks:
-					Array.isArray(this.marks) || this.marks === false
+					(Array.isArray(this.marks) || this.marks === false)
 						? this.marks
 						: undefined,
+				nodes:
+					(Array.isArray(this.nodes) || this.nodes === false)
+						? this.nodes
+						: undefined,
+				headings: this.headings,
```
distantnative, line 151: I do not think we need to pass `headings` to the toolbar. We pass it to the `Heading` node which then takes care itself which heading levels to show as node buttons.

afbora:
I couldn't explain the situation well. Let me try:

1. The behavior in the screenshots is different because [we weren't previously sending nodes to the toolbar](https://github.com/getkirby/kirby/blob/5.0.2/panel/src/components/Forms/Writer/Toolbar.vue#L142-L144). This is why `nodes: false` wasn't working.

2. If we pass nodes to the toolbar, the default nodes are used, and the paragraph isn't [here](https://github.com/getkirby/kirby/blob/5.0.2/panel/src/components/Forms/Input/WriterInput.vue#L94-L97).

3. I passed the headings because the [nodesForBlock prop](https://github.com/getkirby/kirby/blob/5.0.2/panel/src/components/Forms/Writer/Toolbar.vue#L231C3-L236) is already converted. It doesn't contain the word "heading"; contains as h1-6. Therefore, to match the headings [in the dropdown](https://github.com/getkirby/kirby/blob/5.0.2/panel/src/components/Forms/Writer/Toolbar.vue#L141-L165), we needed to determine which headings were enabled or disabled at the blueprint level.

distantnative:
@afbora But then I think we need to include the paragraph node in the default. We cannot have this PR simply dropping it. 

I still cannot follow you regarding no 3. Which heading levels should included as buttons should be handled by the `Heading` node itself. We should not pass the `heading` prop to the toolbar to try letting the toolbar take over that logic, in my opinion.

Right now also the order of nodes changes due to this PR.

afbora:
> But then I think we need to include the paragraph node in the default. We cannot have this PR simply dropping it.

Yes, I agree we should add to the defualts and done in https://github.com/getkirby/kirby/pull/7310/commits/79b2213ce8b180afd3172fe6210e700b16e71dc0

An example about headings issue:

````yaml
fields:
  writer:
    type: writer
    nodes:
      - heading
      - bulletList
      - orderedList
````

image

As you see heading node is not visible because `heading` is converted to `headings`.

After my PR, it will work great:

````yaml
fields:
  writer:
    type: writer
    nodes:
      - paragraph
      - heading
      - bulletList
      - orderedList
    headings:
      - 2
      - 3
      - 4
````

image

afbora:
@distantnative I'm not an expert in Vue. If you have a more correct and stable solution for the heading issue, feel free to commit it 🙂

distantnative:
@afbora I have come up with an alternative PR: https://github.com/getkirby/kirby/pull/7542

I don't think we need to pass the nodes to the toolbar at all after all. Whether they are listed or not depends on their availability in the editor in general. And the nodes themselves take care of if they want to show all their buttons or just some (e.g. the Heading node).

PR #7312, opened 2025-06-22
fix: Slug input with custom `allow`

### LINKED ISSUE #7309: `allow` option renders slug field unusable
## Description
If I provide an `allow` option in a slug field blueprint to allow additional characters (like suggested in the Docs), for example: `allow: _` I cannot type anything at all into the slug field anymore.

Still, the "Save | Discard Changes" buttons pop up and when I check the content file, the *last character that I typed blindly into the field* is saved to the content file (no matter if it is a legal slug character or not), yet nothing shows up in the field UI, even after reloading the page.

(btw. if I provide already legal characters in the allow option, like `allow: abc` I can then type those three characters, although if I type to fast it keeps deleting the last few characters)

**Expected behavior**  
I would expect the field to accept `strings_like_this` if I set it to `allow: _` as suggested in the Docs and certainly I would expect other letters to still show up in the input field and the field to save what I am actually seeing instead of the last letter only.

**Screenshots**  

## To reproduce
Starterkit, set up a slug field, set it to

```yml
testslug:
  type: slug
  allow: _
```
and try typing into the slug field. Nothing will show up, the last typed letter will be saved to the content file.

## Your setup
**Kirby Version**  
v5rc6

**Additional context**  
I did not encounter that specific bug in v4, but in v4 the slug field seems to be broken in other major ways that I do not have time to figure out right now.

distantnative:
## Description

- `$helper.string.slug()` would trim all trailing/leading characters that aren't `a-z0-9`. When a custom `allow` argument was passed, this prohibited any other characters that actually should be allowed.
- Slug input would emit unsluggified and sluggified value on input simultaneously. Refactored it to not use `this.slug` as local state but all use `onInput` when any input occurs.

### Fixes
- Fixed slug field with custom `allow` character set 
#7309 
- `$helper.string.slug()` respects `allow` argument

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Forms/Input/SlugInput.vue
+++ b/panel/src/components/Forms/Input/SlugInput.vue
@@ -2,10 +2,10 @@
 	<k-string-input
 		v-bind="$props"
 		:spellcheck="false"
-		:value="slug"
+		:value="value"
 		autocomplete="off"
 		class="k-slug-input"
-		@input="$emit('input', $event)"
+		@input="onInput"
 	/>
 </template>
 
@@ -50,7 +50,6 @@ export default {
 	mixins: [props],
 	data() {
 		return {
-			slug: this.sluggify(this.value),
 			slugs: this.$panel.language.rules ?? this.$panel.system.slugs,
 			syncValue: null
 		};
@@ -71,19 +70,14 @@ export default {
 				}
 
 				this.syncValue = newValue[this.sync];
-				this.onInput(this.sluggify(this.syncValue));
+				this.onInput(this.syncValue);
 			},
 			deep: true,
 			immediate: true
-		},
-		value(newValue) {
-			newValue = this.sluggify(newValue);
-
-			if (newValue !== this.slug) {
-				this.slug = newValue;
-				this.$emit("input", this.slug);
-			}
 		}
+	},
+	mounted() {
+		this.onInput(this.value);
 	},
 	methods: {
 		sluggify(value) {
@@ -94,8 +88,7 @@ export default {
 			);
 		},
 		onInput(value) {
-			this.slug = this.sluggify(value);
-			this.$emit("input", this.slug);
+			this.$emit("input", this.sluggify(value));
 		}
 	}
 };
--- a/panel/src/helpers/string.js
+++ b/panel/src/helpers/string.js
@@ -184,11 +184,11 @@ export function slug(string, rules = [], allowed = "", separator = "-") {
 	// replace according to language and ascii rules
 	for (const ruleset of rules) {
 		for (const rule in ruleset) {
-			const isTrimmed = rule.slice(0, 1) !== "/";
-			const trimmed = rule.slice(1, rule.length - 1);
-			const regex = isTrimmed ? rule : trimmed;
+			const isRegex = rule.startsWith("/") && rule.endsWith("/");
+			const pattern = isRegex ? rule.slice(1, -1) : rule;
+
 			string = string.replace(
-				new RegExp(RegExp.escape(regex), "g"),
+				new RegExp(RegExp.escape(pattern), "g"),
 				ruleset[rule]
 			);
 		}
@@ -200,18 +200,18 @@ export function slug(string, rules = [], allowed = "", separator = "-") {
 	// replace non-allowed characters (e.g. spaces) with separator
 	string = string.replace(new RegExp("[^" + allowed + "]", "ig"), separator);
 
-	// remove double separators
+	// replace slashes with dashes
+	string = string.replace("/", separator);
+
+	// remove groups of multiple separators
 	string = string.replace(
 		new RegExp("[" + RegExp.escape(separator) + "]{2,}", "g"),
 		separator
 	);
 
-	// replace slashes with dashes
-	string = string.replace("/", separator);
-
 	// trim leading and trailing non-word-chars
-	string = string.replace(new RegExp("^[^a-z0-9]+", "g"), "");
-	string = string.replace(new RegExp("[^a-z0-9]+$", "g"), "");
+	string = string.replace(/^\W+/g, "");
+	string = string.replace(/\W+$/g, "");
 
 	return string;
 }
--- a/panel/src/helpers/string.js
+++ b/panel/src/helpers/string.js
@@ -201,7 +201,7 @@ export function slug(string, rules = [], allowed = "", separator = "-") {
 	string = string.replace(new RegExp("[^" + allowed + "]", "ig"), separator);
 
 	// replace slashes with dashes
-	string = string.replace("/", separator);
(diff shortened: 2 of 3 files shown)


panel/src/components/Forms/Input/SlugInput.vue
```diff
@@ -71,20 +70,15 @@ export default {
...
-
-			if (newValue !== this.slug) {
-				this.slug = newValue;
-				this.$emit("input", this.slug);
-			}
 		}
 	},
+	mounted() {
+		this.onInput(this.value);
+	},
```
distantnative, line 81: `Comment:` In case the field receives an initial value that isn't sluggified, sluggify and emit immediately.

panel/src/helpers/string.js
```diff
@@ -184,11 +184,11 @@ export function slug(string, rules = [], allowed = "", separator = "-") {
...
 		for (const rule in ruleset) {
-			const isTrimmed = rule.slice(0, 1) !== "/";
-			const trimmed = rule.slice(1, rule.length - 1);
-			const regex = isTrimmed ? rule : trimmed;
+			const isRegex = rule.startsWith("/") && rule.endsWith("/");
+			const pattern = isRegex ? rule.slice(1, -1) : rule;
+
 			string = string.replace(
-				new RegExp(RegExp.escape(regex), "g"),
+				new RegExp(RegExp.escape(pattern), "g"),
```
distantnative, line 191: `Comment:` mixed in a code cleanup bit, sorry.

panel/src/helpers/string.js
```diff
@@ -200,18 +200,18 @@ export function slug(string, rules = [], allowed = "", separator = "-") {
 	// replace non-allowed characters (e.g. spaces) with separator
 	string = string.replace(new RegExp("[^" + allowed + "]", "ig"), separator);
 
-	// remove double separators
+	// replace slashes with dashes
+	string = string.replace("/", separator);
```
distantnative: `Comment`: should go first, so that multiple separators resulting from this replacement get reduced to a single separator in the next step.

panel/src/helpers/string.js
```diff
@@ -200,18 +200,18 @@ export function slug(string, rules = [], allowed = "", separator = "-") {
...
 	);
 
-	// replace slashes with dashes
-	string = string.replace("/", separator);
-
 	// trim leading and trailing non-word-chars
-	string = string.replace(new RegExp("^[^a-z0-9]+", "g"), "");
-	string = string.replace(new RegExp("[^a-z0-9]+$", "g"), "");
+	string = string.replace(/^\W+/g, "");
+	string = string.replace(/\W+$/g, "");
```
distantnative, line 214: `Though:` Could also be based on `allowed`?

panel/src/components/Forms/Input/SlugInput.vue
bastianallgeier: This could probably only lead to a bit of confusion because the form buttons would appear immediately, right?

distantnative: Yes, but I think that's the same behaviour as before when a value that doesn't meet the slug standard arrives in the input.

panel/src/helpers/string.js
bastianallgeier: This would only replace one slash though. Is that intended?

panel/src/helpers/string.js
bastianallgeier: I'm not sure I understand what you mean here.

panel/src/helpers/string.js
distantnative: I copied this one from line 210. But indeed, should probably be `replaceAll`?

panel/src/helpers/string.js
distantnative: This is kind of the last stance stripping all leftover unwanted characters from beginning and end. So far we did `a-z0-9` which leads to problems if someone wants to allow e.g. `_` because it will be stripped from the end even while typing. I added here to strip all non-word characters instead, but also thought that maybe here one should also just strip/trim again any non-`allowed` characters.

bastianallgeier: Sounds definitely logical to go for the non-allowed characters. I would definitely give that a try.

panel/src/helpers/string.slug.test.js
```diff
@@ -71,8 +71,8 @@ describe.concurrent("$helper.string.slug()", () => {
 	});
 
 	it("should produces safe filenames", () => {
-		const result = slug("-what a view@2x.png_", [], "a-z0-9@._-");
```
bastianallgeier, line 74: Wouldn't the underscore at the end of the string be another good test case here? Was there a particular reason why you removed it?

panel/src/helpers/string.js
bastianallgeier: ☑️

panel/src/helpers/string.slug.test.js
distantnative: I guess that's actually the explanation why we added this very harsh trimming in the first place. Cause `.png_` is not a valid filename. But maybe we have to split this up: `slug` helper can still produce strings with all the allowed characters also at beginning/end. But maybe it needs a new filename helper which uses `slug` but is also more restrictive on the extension?

panel/src/helpers/string.js
bastianallgeier: I tried it and I think the non-word characters solution leads to better results. For good slugs, IMHO you don't want them to start with -, _ or any other special characters. If we use "allowed" here, those characters wouldn't be trimmed if they are included in the set.

distantnative:
@bastianallgeier I think we have a conflict by principle here:
- We want to be able to define a custom set of allowed characters.
- We want to trim though all non-word characters, even if they re present in the allowed set.
- We use it to sluggify input while the use is typing. By definition, characters will appear temporarily at the end that later on would be in the middle of the string. Our code then strips non-word characters even if they would have been in the allowed characters set and thus would've been fine when later in the middle of the string.

This is difficult to resolve. We could either not allow a custom set anymore. Or we are fine with all allowed characters at the end of the string (and then e.g. would have to remove them for file names separately). Or we create a solution where while typing we only strip non-allowed characters and run the stricter mode (via a flag) only on blur/submit (which could have the effect though that the user first sees something and on blur it suddenly changes again).

lukasbestle:
Another alternative: Have a stricter mode via flag but activate that not only on blur/submit but also on a debounce of 500-1000 ms. Then if one continues typing, the non-word character is accepted. But if one stops typing, it is removed. Not sure how the UX of that could work in practice, but we could try it.

distantnative:
Closing this to keep our PR list manageable. We will have to take a new stab at this eventually.

PR #7335, opened 2025-06-25
feat: Navigate Preview view with a page tree

distantnative:
## Description

Screenshot 2025-06-25 at 20-26-33 A night in the forest Preview Lab

### ✨ Enhancements

- Added page tree navigation to Preview view

## Ready?

### For review team

### DIFF
--- a/panel/src/components/Views/Preview/PreviewView.vue
+++ b/panel/src/components/Views/Preview/PreviewView.vue
@@ -19,11 +19,7 @@
 				>
 					{{ title }}
 				</k-button>
-				<k-dropdown-content
-					ref="tree"
-					theme="light"
-					class="k-preview-view-tree"
-				>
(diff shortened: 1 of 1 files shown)


bastianallgeier:
This is really cool. First impressions: 

- I think we need to set a fixed width for the dropdown to avoid the layout shift. 
- When you navigate to a slower loading page, we need some sort of loading indication. Otherwise it seems like nothing is happening for a split second
- It would look amazing to have a black tree dropdown, but the tree component isn't styled for that yet. 

Otherwise it works really well already.

distantnative:
I stayed away from a fixed width as that then cuts off Long page titles again. Not sure if the visual there should overrule the usability. But we could maybe add a min-width and max-width?

distantnative:
@bastianallgeier I have added your suggestions

bastianallgeier (approved):

PR #7340, opened 2025-06-26
Add $collection->join()

adamkiss:
## Description
Adds a method to join collection into a string, supporting mapping and custom glue.

Helps to simplify mapping and joining a collection:

```php
$x = $page->children()->map(fn($p)=>snippet('some', ['page'=>$page], return: true);
$x = $x->toArray();
$x = A::join($x, '');
```

```php
$x = $page->children()->join(as: fn($p)=>snippet('some', ['page'=>$page], return: true);
```

### Summary of changes
- new Collection method

### Reasoning
- collection joining is the second most used function after foreach, IMO. This even replaces some foreach calls

## Ready?

### For review team

### DIFF
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -626,6 +626,14 @@ class Collection extends Iterator implements Stringable
 	}
 
 	/**
+	 * Joins the collection elements into a string, optionally running a closure over elements
+	 */
+	public function join(string $glue = '', Closure|null $as = null): string {
+		$values = $this->toArray($as);
+		return implode($glue, $values);
+	}
+
+	/**
 	 * Returns the last element
 	 *
 	 * @return TValue
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -628,7 +628,8 @@ class Collection extends Iterator implements Stringable
 	/**
 	 * Joins the collection elements into a string, optionally running a closure over elements
 	 */
-	public function join(string $glue = '', Closure|null $as = null): string {
+	public function join(string $glue = '', Closure|null $as = null): string
+	{
 		$values = $this->toArray($as);
 		return implode($glue, $values);
 	}
--- a/tests/Toolkit/CollectionTest.php
+++ b/tests/Toolkit/CollectionTest.php
@@ -489,6 +489,27 @@ class CollectionTest extends TestCase
 		$this->assertFalse(isset($this->collection->super));
 	}
 
+	public function testJoin(): void
+	{
+		$collection = new Collection([
+			'a' => 'd',
+			'b' => 'e',
+			'c' => 'f',
+		]);
+
+		$this->assertSame('def', $collection->join());
+		$this->assertSame('d-e-f', $collection->join('-'));
+
+		$collection = new Collection([
+			'a' => ['d'],
+			'b' => ['e'],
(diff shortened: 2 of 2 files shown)


distantnative:
I would expect that a collection method called `join` joins two collections into one. Not turning them into a string. I am a bit unsure that this is a common use case but happy to get to know if I'm wrong.

adamkiss:
@distantnative 

- PHP alts `join` for `implode`
- there's `A::join` already
- [Laravel also has one](https://laravel.com/docs/12.x/collections)
- there's already `add` for adding, I could also PR merge  as an alternative

afbora:
I think it makes sense since we already have `A::join` that does same thing.

distantnative:
Fair. @adamkiss Would you be able to fix the two code style errors?

adamkiss:
@distantnative yeah

distantnative (approved):

afbora:
Sorry for delay but I just wonder why we used empty separator for for collection join but we used comma for arrays?

````php
public static function join(
  array|string $value,
  string $separator = ', '
): string {
  if (is_string($value) === true) {
    return $value;
  }

  return implode($separator, $value);
}
````
https://github.com/getkirby/kirby/blob/main/src/Toolkit/A.php#L401-L410

**Comment by distantnative** [comment]
@afbora I think you are right and it makes sense to stay consistent there. Probably should also stick to `$separator` as parameter name. I can fix it directly on the branch.

**Comment by adamkiss** [comment]
@afbora @distantnative Because while I understand why you thought joining array items vas imagined with some kind of tags or something, where commas might be preferable, I think major usage of collection (pages, users) join is html output, where no extraneous elements are wanted. 

I think it's better to keep it empty, but settable, rather than default to comma and force every html output usage to unset it

**Comment by distantnative** [comment]
@adamkiss I disagree, I think you are coming from a very specific use case in mind. But joining a collection doesn't have to be about HTML output (in fact I think your example use case with the snippet is more an outlier), e.g.

```php
$venues = $events->join(as: fn ($event) => $event->venue());
```

I think it does make sense to keep `A::join()` and `Collection::join()` consistent here.

afbora:
I don't think it's a big deal. Because both spaces and commas can be used in different use cases. But it would be nice to have consistency.

adamkiss:
@distantnative I'm not saying it's impossible to come up with a different use case, but I wholeheartedly think that

1. There are more uses cases where the output is HTML in my opinion, especially given the major use of collection and inheriting classes we work with are Pages
2. it's no harder to add then unset

Thus consistency in the direction of default value is IMO detrimental to the major workflow

@afbora the default I set isn't a space, it's nothing, just putting the outputs together

distantnative:
@adamkiss Agree to disagree here. Consistency has a super high value with a framework tool like Kirby - we can tell how it bites us every time we aren't consistent as it defies user expectations. The Collection class is such a general tool, to me it doesn't make sense to add a narrower use-case of HTML (this is where we will disagree) as default, trading off with consistency.

afbora:
> the default I set isn't a space, it's nothing, just putting the outputs together

What I meant by "space" was the spaceless 🙂 

Beyond consistency, if I wanted to join a collection or array to a string, I would think it would need a separator to make it readable. So, unlike you, I personally think users would prefer commas.

PR #7346, opened 2025-06-30
Fix: Slow structures

bastianallgeier:
## Description

We’ve refactored the structure field to be faster and unfortunately missed one important part. The form method is called for every row and will recreate the entire form instead of reusing a cached variant of the form. This destroys all the hard work that we put in the form refactoring. 

We should finally move the structure field to an extended FieldClass, but we can't do this without keeping the old code around. Otherwise, all structure field extensions would break. 

### Fixes 

- Fixed performance bottleneck in structure fields 

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/config/fields/structure.php
+++ b/config/fields/structure.php
@@ -159,7 +159,6 @@ return [
 	'methods' => [
 		'rows' => function ($value) {
 			$rows  = Data::decode($value, 'yaml');
-			$form  = $this->form();
 			$value = [];
 
 			foreach ($rows as $index => $row) {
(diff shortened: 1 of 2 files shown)


config/fields/structure.php
```diff
@@ -156,29 +156,31 @@
 			return $columns;
 		}
 	],
+	'ready' => function () {
+		$this->form = new Form(
+			fields: $this->attrs['fields'] ?? [],
+			model: $this->model,
+			language: 'current'
+		);
+	},
```
afbora: I love the new `ready` prop but I've a question. Can be doable within methods? What's the difference with following?

````php
'methods' => [
	'form' => function () {
		$this->form ??= new Form(
			fields: $this->attrs['fields'] ?? [],
			model: $this->model,
			language: 'current'
		);

		return $this->form->reset();
	}
],
````

bastianallgeier: 🙈 I feel quite stupid now. There's no difference at all. I just fixed it.

afbora (approved):

PR #7354, opened 2025-07-06
chore: Remove vite's `splitVendorChunkPlugin `

distantnative:
## Description

- Remove the deprecated `splitVendorChunkPlugin` from site.config
- `manualChunks` option should produce the same split as when using the deprecated plugin.
- Vite 7 removes the deprecated plugin which is why we should loose it.


bastianallgeier (approved):
I only ran a quick build test but it all seems to work fine afterwards.

PR #7361, opened 2025-07-11
fix!: Place sort handle inside `k-item` boundaries

### LINKED ISSUE #7349: Problem changing order of page list items in panel
## Description
When I try to change the order of the **pages** in the panel, it does not change.

````yaml
# site.yml
sections:
  pages:
    type: pages
    image: icon
    search: true
    max: 20
    create:
      - default
````

**Screenshots**  
https://github.com/user-attachments/assets/662b2e27-f0bd-4fde-9999-fa4c345d34e1

## Your setup
Kirby 5.0.0

distantnative:
## Description

- [x] https://github.com/getkirby/kirby/pull/7390
- [x] https://github.com/getkirby/kirby/pull/7516

This PR improves drag'n'drop sorting of `k-items` for list layout: Instead of positioning the sort handle absolute outside of the `k-items` boundaries, it is now positioned relatively inside the container. This is necessary as SortableJS gets tripped up by the absolute positioning outside. To achieve the same look/style as before, a new `.k-item-wrapper` div was added and the main styles applied to this. 

### Enhancements
- Better drag sorting (e.g. pages and files sections) in list layout
#7349 

### Breaking changes
- DOM structure of `k-item` has been changes. An additional wrapper `.k-item-box` div was added.

### For review team

### DIFF
--- a/panel/src/components/Collection/Item.vue
+++ b/panel/src/components/Collection/Item.vue
@@ -212,6 +212,7 @@ export default {
 .k-item-box {
 	position: relative;
 	width: 100%;
+	height: 100%;
 	background: var(--item-color-back);
 	box-shadow: var(--item-shadow);
 	border-radius: var(--rounded);
@@ -276,7 +277,7 @@ export default {
 .k-item[data-layout="list"][data-has-image="true"] .k-item-box {
 	grid-template-columns: var(--item-height) 1fr auto;
 }
-.k-item[data-layout="list"] .k-frame {
+.k-item[data-layout="list"] .k-item-image {
 	--ratio: 1/1;
 	border-start-start-radius: var(--rounded);
 	border-end-start-radius: var(--rounded);
@@ -346,7 +347,7 @@ export default {
 		"image options";
 	grid-template-columns: minmax(0, var(--item-height)) 1fr;
 }
-.k-item[data-layout="cardlets"] .k-frame {
+.k-item[data-layout="cardlets"] .k-item-image {
 	grid-area: image;
 	border-start-start-radius: var(--rounded);
 	border-end-start-radius: var(--rounded);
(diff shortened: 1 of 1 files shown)


bastianallgeier:
This is the kind of change where I wonder if we would actually need to call it "breaking". At least in terms of the markup and CSS for items. Unfortunately the items are so dominant in the panel that I can see how this could be an issue.

distantnative:
I totally understand that point. At the same time I am not sure we can leave sorting pretty broken in the list layout for all of v5.

bastianallgeier:
I tried a couple more things to get this done via a CSS-only solution, but it all feels like a hack. I think your solution really is the most stable one. It's nitpicky, but maybe we can find a better name for the wrapper? Something like k-item-box maybe?

distantnative:
I just saw a comment on the sortablejs repo that position absolute would work as long as the same z-index. Maybe we could try that.

Otherwise, open to any name for the new div .

distantnative:
Nevermind, that was also only true when the handle is placed absolute but inside the parent. So I think we really have to restructure it like in this PR. But not sure if in v5.x or v6.

bastianallgeier:
Can we maybe split the PR? All your non-destructive changes to the items component could be immediately merged and I think they already help to fix a few things. With the other parts, we might think again if there's a CSS-only solution for v5 and then a more drastic solution for v6?

bastianallgeier (approved):

bastianallgeier:
@distantnative I think you can resolve the conflicts better than me, but the PR definitely looks good.

distantnative:
@bastianallgeier I will wait for the multi-select PR first, to not have to do it twice.

distantnative:
@bastianallgeier Bringing this back for v6. The tab focus currently grabs the whole `.k-item`, but now it would be better to grab just `.k-item-box`, could you help with that?

bastianallgeier:
I just had a brief first look at it and realized that we have another layout issue besides the focus problem. In grids, the height of the card is no longer adapted automatically because of the wrapping element. 

Screenshot 2025-12-22 at 13 17 09

distantnative:
@bastianallgeier I think I have found solutions for both problems.

bastianallgeier (approved):
Works great!!

PR #7368, opened 2025-07-13
refact: Allow folders inside `site/controllers`

distantnative:
## Description

This PR changes `App::controller()` to also allow controllers noted in subfolders inside `site/controllers`. This change enables us to add snippet and block controllers (https://feedback.getkirby.com/213) via a custom plugin then:

```php
<?php

use Kirby\Cms\App;
use Kirby\Template\Snippet;
use Kirby\Toolkit\Str;

class SnippetWithController extends Snippet
{
  protected static $controllers = [];

  public static function controller(
    string|null $file,
    array $data = []
  ): array {
    if (
      $file === null ||
      str_starts_with($file, static::root()) === false
    ) {
      return $data;
    }

    if (isset(static::$controllers[$file])) {
      return array_replace_recursive(
        $data,
        static::$controllers[$file]
      );
    }

    $name = ltrim(Str::before(Str::after($file, static::root()), '.php'), '/');

    // if the snippet has a name, we can load the controller
    // and merge the data with the controller's data
    if ($name !== null) {
      $data = array_replace_recursive(
        $data,
        static::$controllers[$file] = App::instance()->controller('snippets/' . $name, $data)
      );
    }

    return $data;
  }

    string|array|null $name,
    array $data = [],
    bool $slots = false
  ): static|string {
    $file = $name !== null ? static::file($name) : null;
    $data = static::controller($file, $data);
    return parent::factory($name, $data, $slots);
  }

  public function render(array $data = [], array $slots = []): string
  {
    $data = array_replace_recursive(
      static::controller($this->file, $data),
      $data
    );

    return parent::render($data, $slots);
  }
}

App::plugin('getkirby/snippet-controllers', [
  'components' => [
    'snippet' => function (
      App $kirby,
      string|array|null $name,
      array $data = [],
      bool $slots = false
    ): Snippet|string {
      return SnippetWithController::factory($name, $data, $slots);
    },
  ]
]);
```

With this plugin place, one can add a controller for e.g. `site/snippets/header.php` to `site/controllers/snippets/header.php`. And since blocks are also just snippets: `site/controllers/snippets/blocks/video.php` (though this currently just works if a custom blocks snippet has been added, not for the default ones).

Choosing to go down the plugin path as it can have quite the performance impact if every snippet checks for the existence of a controller. Maybe you have also ideas how to improve that performance.

### Refactored
- Support `App::controller()` to return controllers nested in subfolders inside `site/controllers`

## Ready?

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -443,15 +443,15 @@ class App
 		array $arguments = [],
 		string $contentType = 'html'
 	): array {
-		$name = basename(strtolower($name));
+		$name = strtolower($name);
 		$data = [];
 
 		// always use the site controller as defaults, if available
-		$site   = $this->controllerLookup('site', $contentType);
-		$site ??= $this->controllerLookup('site');
-
-		if ($site !== null) {
-			$data = (array)$site->call($this, $arguments);
+		// (unless the controller is a snippet controller)
+		if (strpos($name, '/') === false) {
+			$site   = $this->controllerLookup('site', $contentType);
+			$site ??= $this->controllerLookup('site');
+			$data   = (array)$site?->call($this, $arguments) ?? [];
 		}
 
 		// try to find a specific representation controller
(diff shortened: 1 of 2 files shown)


bastianallgeier (approved):

jensscherbl:
Nice! I ended up not using `App::controller()` and instead using `Controller::load` directly  because of this limitation when playing with the idea on my own [a while back][1].

[1]: https://github.com/kenshodigital/kirby-snippet

PR #7369, opened 2025-07-14
perf: Cache snippets

distantnative:
## Description

Avoid repeated `F::exists()` and extension calls when using the same snippets within the request

### Performance
- Snippets get cached for the duration of the request

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -102,7 +102,8 @@ class App
 		$this->core   = new Core($this);
 		$this->events = new Events($this);
 
-		// start with a fresh version cache
+		// start with a fresh snippet and version cache
+		Snippet::$cache = [];
 		VersionCache::reset();
 
 		// register all roots to be able to load stuff afterwards
--- a/src/Template/Snippet.php
+++ b/src/Template/Snippet.php
@@ -24,6 +24,11 @@ use Kirby\Toolkit\Tpl;
 class Snippet extends Tpl
 {
 	/**
+	 * Cache for the snippet files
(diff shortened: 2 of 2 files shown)


src/Cms/App.php
```diff
@@ -102,7 +102,8 @@ public function __construct(array $props = [], bool $setInstance = true)
 		$this->core   = new Core($this);
 		$this->events = new Events($this);
 
-		// start with a fresh version cache
+		// start with a fresh snippet and version cache
+		Snippet::resetCache();
```
distantnative: This will not work when someone replaces the snippet component with their own class. Have to think of an alternative.

distantnative: @bastianallgeier Do you see any way with the current `snippet` core component implementation to find out from `Kirby\Cms\App` what class is used for snippets?

bastianallgeier:
I have no immediate solution yet, but to move this further, I've setup a benchmark scenario in the sandbox to see the difference. 

With 1000 pages rendering 1000 snippets each, the difference is roughly 2 seconds on my M2. I think it's worth to keep working on this.

distantnative:
@bastianallgeier Maybe we should have a general way to hook into the app setup method and then use it for our `Snippet` and `VersionCache` reset calls. I think normal hooks though probably don't work because plugins/options are set up themselves only later (or maybe that's fine cause we do not necessarily need it on any first call but mainly on every `$app->clone()` calls.

bastianallgeier (approved):

PR #7371, opened 2025-07-14
fix: Allow submitting inactive fields

### LINKED ISSUE #7333: Conditional field in structure looses content when condition is not default
## Description
If there is a field with a condition in a structure field, the content doesn't get stored if that condition isn't the default value of the field that triggers it (e.g. a radiofield). It shows up short when the structure is closed, but as soon as you save the page the field is empty again.

In the contentfolder the `_changes`-folder gets not created and even existing content from other structureentries gets deleted.

**Expected behavior**
The content should be stored and saved.

**Screenshots**
In the first video there is a simple condition with radiobuttons. `Image` is the default value in this case. All fields that have this condition can be saved. The other fields with `Video` as condition loose there content.

https://github.com/user-attachments/assets/4bb25655-b2e8-4985-9f9d-36824a5f1447

In the second video the default value is switched, then you can save the content of fields with the `Video` condition. But all other content is lost.

https://github.com/user-attachments/assets/252a8757-b441-487c-9691-dc7307da20a9

## To reproduce

1. Install a fresh plainkit
2. Add structurefield to the blueprint
3. In the structurefield add some conditional fields and a radiofield to trigger the conditions
4. Try to add content and save the page

## Your setup
**Kirby Version**
Kirby 5.0.0

**Console output**
No error in the console

 - Device: Mac
 - OS: macOS
 - Browser: Firefox Developer
 - Version: 140

### LINKED ISSUE #7370: Data loss in structure fields with toggle field and when option
If a structure field contains a `toggle field` and a `when option`, changing the status of the `toggle field` will result in data loss. This will become apparent after reloading the page or, at the latest, after saving.

To make the data loss visible, I added the `when option` afterwards.

Tested with a fresh start kit and Kirby 5.0.1.

https://github.com/user-attachments/assets/4136f1ab-875b-4b57-a5e2-b2811791d5ad

```
sections:

  content:
    type: fields
    fields:

      addresses:
        label: Addresses
        type: structure
        fields:
          toggle:
            label: Include in menu?
            type: toggle
            text:
              - "no"
              - "yes"
          street:
            label: Street
            type: text
            when:
              toggle: true
          zip:
            label: ZIP
            type: text
            when:
              toggle: true
          city:
            label: City
            type: text
            when:
              toggle: true
          text:
            type: blocks
```

bastianallgeier:
## Description

Our `$field->isSubmittable()` check was too strict and did not allow to submit inactive when query fields that could still contain data though. This could lead to data loss. 

### Fixes

- Inactive fiels (due to non-matching when queries) can be submitted again https://github.com/getkirby/kirby/issues/7333 https://github.com/getkirby/kirby/issues/7370

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Form/Mixin/Value.php
+++ b/src/Form/Mixin/Value.php
@@ -120,10 +120,6 @@ trait Value
 			return false;
 		}
 
-		if ($this->isActive() === false) {
-			return false;
-		}
-
 		return true;
 	}
 
--- a/tests/Form/FieldClassTest.php
+++ b/tests/Form/FieldClassTest.php
@@ -417,7 +417,45 @@ class FieldClassTest extends TestCase
 			],
 		]);
 
-		$this->assertFalse($field->isSubmittable($language));
+		$this->assertTrue($field->isSubmittable($language));
+	}
+
+	public function testIsSubmittableWithWhenQueryAndTwoFields()
+	{
+		$language = Language::ensure('current');
+
+		$fields = new Fields([
+			new TestField([
+				'name'  => 'a',
+				'value' => 'a',
+				'when'  => [
+					'b' => 'b'
+				]
+			]),
+			new TestField([
+				'name'  => 'b',
+				'value' => 'b',
+				'when'  => [
+					'a' => 'a'
+				]
+			]),
+		]);
+
+		$a = $fields->get('a');
+		$b = $fields->get('b');
+
+		$this->assertTrue($a->isSubmittable($language));
+		$this->assertTrue($b->isSubmittable($language));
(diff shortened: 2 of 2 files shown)


distantnative (commented):
@bastianallgeier shoule we maybe remove the inActive method since it's also broken?

bastianallgeier:
No, it's still needed in the needsValue method and has the same logic as we used in v4 afaik. As soon as the fields are submitted, the logic in that method should still be correct.

distantnative (approved):

PR #7372, opened 2025-07-14
refact!: Drop PHP 8.2 support

distantnative:
### Breaking changes
- Removed support for PHP 8.2. Use PHP 8.3 or 8.4 instead.

### For review team


bastianallgeier:
If we are realistically looking at a possible launch for v6, it won't be before Mid 2026. We should not do less than a year between major releases. By that time, I think it's more than fair to drop 8.2 support. PHP drops support for 8.2 in December 2025.

bastianallgeier:
Quite some changes are related to the protected URI properties. It makes me wonder if we actually want to make them public instead. It's a bit of a general style question. But in those mutable classes, the right property types should normally already protect us from invalid values, right?

distantnative:
@bastianallgeier I stayed away from it as I wasn't sure if e.g. making the `password` property public would be great security-wise.

bastianallgeier:
You are right, we should probably wait for 8.4 to finally move to setters and getters for props and better be safe until then.

bastianallgeier (approved):

PR #7373, opened 2025-07-14
feat: `Kirby\Panel\Ui\Stat` backend component

bastianallgeier:
## Description

The new class is used in the stats section to clean up the code and can later be used in our new stats field as well to simplify migration between sections and fields. 

This is a first PR in a series of steps to get rid of sections. 

### Enhancements

- New `Kirby\Panel\Ui\Stat` backend component, which is used to create the correct information for the `<k-stat>` Vue component. 
### Refactoring

- The `stats` field uses the new `Kirby\Panel\Ui\Stat` class to clean up the props generation for the frontend.

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/src/Panel/Ui/Stat.php
+++ b/src/Panel/Ui/Stat.php
@@ -69,27 +69,27 @@ class Stat extends Component
 
 	public function icon(): string|null
 	{
-		return $this->query($this->icon);
+		return $this->stringTemplate($this->icon);
 	}
 
 	public function info(): string|null
 	{
-		return $this->query(
-			$this->translate($this->info)
+		return $this->stringTemplate(
+			$this->i18n($this->info)
 		);
 	}
 
 	public function label(): string
 	{
-		return $this->query(
-			$this->translate($this->label)
+		return $this->stringTemplate(
+			$this->i18n($this->label)
 		);
 	}
 
 	public function link(): string|null
 	{
-		return $this->query(
-			$this->translate($this->link)
+		return $this->stringTemplate(
+			$this->i18n($this->link)
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


src/Panel/Ui/Stat.php
```diff
@@ -0,0 +1,133 @@
...
+				query: $input
+			);
+		}
+
+		return new static(...[
+			...$input,
+			'model' => $model,
+			'label' => $input['label'],
+			'value' => $input['value'],
+		]);
```
distantnative: `Suggestion:` Have you already tried if this works?

```suggestion
		$input['model'] = $model;
		return new static(...$input);
```

src/Panel/Ui/Stat.php
```diff
@@ -0,0 +1,133 @@
...
+		public array|string $value,
+		public string $component = 'k-stat',
+		public string|null $icon = null,
+		public array|string|null $info = null,
+		public array|string|null $link = null,
+		public string|null $theme = null,
+	) {
+	}
+
+	public static function from(
```
distantnative, line 34: `Nitpick:` Wondering if `factory` would be more fitting, but no strong opinion here.

src/Panel/Ui/Stat.php
```diff
@@ -0,0 +1,133 @@
...
+			'link'  => $this->link(),
+			'theme' => $this->theme(),
+			'value' => $this->value(),
+		];
+	}
+
+	protected function query(string|null $query): string|null
+	{
+		return $query === null ? null : $this->model->toString($query);
+	}
```
distantnative: With models, the difference has been that `query` returns all kind of data while `toString` always returns only strings. Mixing up the terminology here might not be ideal.

src/Panel/Ui/Stat.php
bastianallgeier: I just tried and Psalm isn't happy about it. It feels a bit weird indeed to list them here and potentially have notice errors. Maybe defaulting to null could be another option and then the class constructor will throw a usable error?

distantnative: If it's just psalm, we might also want to suppress the psalm error, if we actually prefer our syntax there.

src/Panel/Ui/Stat.php
bastianallgeier: I guess we need a clearer concept for when to use factory or from :) I'm not sure about it either and I also don't have a strong opinion.

distantnative: Ignore it. We have `from` and `for` also in many places.

src/Panel/Ui/Stat.php
bastianallgeier: You are completely right. I just remembered that we do exactly the same in the FieldClass with the i18n and stringTemplate methods. I just copied them over to have them consistent between FieldClass and Stat. I think it would be good to turn that into a standard or a trait somehow at a later point.

src/Panel/Ui/Stat.php
bastianallgeier: you are right. It's weird to make the machine happy and us sad 😆

distantnative (approved):

PR #7374, opened 2025-07-14
feat: `Kirby\Panel\Ui\Stats` backend component

bastianallgeier:
## Merge first

- [x] https://github.com/getkirby/kirby/pull/7373

## Description

The new class is used in the stats section to clean up the code and can later be used in our new stats field as well to simplify migration between sections and fields. 

This is the second PR in a series of steps to get rid of sections. 

### Enhancements

- New `Kirby\Panel\Ui\Stats` backend component, which is used to create the correct information for the `<k-stats>` Vue component. 

### Refactoring

- The `stats` field uses the new `Kirby\Panel\Ui\Stats` class to clean up the props generation for the frontend.

### Breaking changes

None

## Ready?

### For review team

### DIFF
--- a/config/sections/stats.php
+++ b/config/sections/stats.php
@@ -1,7 +1,6 @@
 <?php
 
-use Kirby\Exception\InvalidArgumentException;
-use Kirby\Panel\Ui\Stat;
+use Kirby\Panel\Ui\Stats;
 
 return [
 	'mixins' => [
@@ -11,20 +10,8 @@ return [
 		/**
 		 * Array or query string for reports. Each report needs a `label` and `value` and can have additional `info`, `link`, `icon` and `theme` settings.
 		 */
-		'reports' => function ($reports = null) {
-			if ($reports === null) {
-				return [];
-			}
-
-			if (is_string($reports) === true) {
-				$reports = $this->model()->query($reports);
-			}
-
-			if (is_array($reports) === false) {
-				return [];
-			}
-
-			return $reports;
+		'reports' => function (array|string|null $reports = null) {
+			return $reports ?? [];
 		},
 		/**
 		 * The size of the report cards. Available sizes: `tiny`, `small`, `medium`, `large`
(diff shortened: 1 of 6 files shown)


config/sections/stats.php
```diff
@@ -34,24 +21,18 @@
...
-				} catch (InvalidArgumentException) {
-					continue;
-				}
-
-				$reports[] = $stat->props();
-			}
-
-			return $reports;
+		'stats' => function (): Stats {
+			return $this->stats = Stats::from(
```
distantnative: Isn't the computed entry already assigning it to `$this->stats` anyways?

bastianallgeier: In theory yes, but the obscure state of the Component.php class has some weird surprises. While refactoring the structure field recently, it did not work as cache properly.

distantnative: And should it be `??=` then?

src/Panel/Ui/Stats.php
```diff
@@ -0,0 +1,90 @@
...
+		ModelWithContent $model,
+		string $query,
+		string $size = 'large'
+	): static {
+		return new static(
+			model: $model,
+			reports: $model->query($query),
+			size: $size
+		);
+	}
```
distantnative: `Suggestion:` Is there a benefit of having the extra method?

```suggestion
	public static function from(
		ModelWithContent $model,
		array|string $reports,
		string $size = 'large'
	): static {
		if (is_string($reports) === true) {
			$reports = $model->query($reports);
		}

		return new static(
			model: $model,
			reports: $reports,
			size: $size
		);
	}
```

config/sections/stats.php
bastianallgeier: Yes, of course 🙈

src/Panel/Ui/Stats.php
bastianallgeier: ☑️

distantnative (approved):

PR #7380, opened 2025-07-15
feat: Backend classes for dialogs and drawers

distantnative:
## Description

Adds the basic PHP backend classes that we can use to build classes for the specific dialogs and drawers (to get rid of the code in `config/areas/`).

### Enhancements
- Backend classes for dialogs and drawer (in `Kirby\Panel\Controller` and `Kirby\Panel\Ui`)

### Breaking changes
- Renamed `Kirby\Panel\Controller\PageTree` to `Kirby\Panel\Controller\PageTreeController`
- Renamed `Kirby\Panel\Controller\Search` to `Kirby\Panel\Controller\SearchController`

### For review team

### DIFF
--- a/config/areas/site/requests.php
+++ b/config/areas/site/requests.php
@@ -1,13 +1,13 @@
 <?php
 
 use Kirby\Cms\App;
-use Kirby\Panel\Controller\PageTree;
+use Kirby\Panel\Controller\PageTreeController;
 
 return [
 	'tree' => [
 		'pattern' => 'site/tree',
 		'action'  => function () {
-			return (new PageTree())->children(
+			return (new PageTreeController())->children(
 				parent: App::instance()->request()->get('parent'),
 				moving: App::instance()->request()->get('move')
 			);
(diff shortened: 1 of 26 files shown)


bastianallgeier:
I like this direction a lot. My first thought was if we maybe want to add Controller to the classnames. DialogController and DrawerController might help to differentiate them better. For the components, that additional class name appendix feels weird. But for the controllers it makes sense to me. What do you think?

distantnative:
@bastianallgeier We would also rename `PageTreeController` and `SearchController` then?

bastianallgeier:
@distantnative if we agree here, yes.

distantnative:
@bastianallgeier I don't think I have an opinion here 😄, so I'll just change it.

afbora:
I am also in favor of adding suffix 👍

distantnative:
@bastianallgeier @afbora Having added the suffix, I feel like this makes the code a bit clunky. I think without it, it felt more elegant. And with the namespaces it would still be very clear which is which.

PR #7381, opened 2025-07-15
refact!: Move `$panel` logic out of `k-panel-menu`

distantnative:
BREAKING CHANGE: `k-menu-panel` requires new props

## Description

Making the `k-panel-menu` component a bit dumber itself.

### Breaking changes
- `<k-panel-menu>` now requires to pass props explicitly instead of using `$panel.menu` itself.

### For review team

### DIFF
--- a/panel/src/components/View/Menu.vue
+++ b/panel/src/components/View/Menu.vue
@@ -2,9 +2,9 @@
 	<nav
 		class="k-panel-menu"
 		:aria-label="$t('menu')"
-		:data-hover="$panel.menu.hover"
-		@mouseenter="$panel.menu.hover = true"
-		@mouseleave="$panel.menu.hover = false"
+		:data-hover="isHovered"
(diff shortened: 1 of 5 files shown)


panel/src/components/View/Menu.vue
```diff
@@ -40,42 +40,57 @@
...
+		isOpen: Boolean,
+		items: {
+			type: Array,
+			default: () => []
+		},
+		license: String,
+		searches: {
+			type: Object,
+			default: () => ({})
+		}
```
bastianallgeier: I wonder if this should be even more "stupid" and just be a boolean "search" prop. We don't do anything else with the searches object and only check its length. That could also be done outside of the menu.

distantnative: Have added it.

bastianallgeier (approved):

PR #7382, opened 2025-07-15
feat: Allow class controller for dialog/drawer routes

distantnative:
## Description

If we want to use backend classes for dialogs and drawers (https://github.com/getkirby/kirby/pull/7380), we need an easy way to use them inside the dialog/drawer routes config. 

### Enhancements
- Dialog and drawer routes: support passing a `controller` class name or closure hat returns an object with `::load()` and `::submit()` methods that will be used then as `load`/`submit` event actions. Ideally used with returning a `Kirby\Panel\Controller\Dialog`/`Kirby\Panel\Controller\Drawer` object. If the named class also defines a static `::for()` method, it will be used to create an instance of the class.

## Docs

Instead of a `load` and `submit` callback, you can also define a class of  `Kirby\Panel\Controller\Dialog`/`Kirby\Panel\Controller\Drawer` as `controller` for the route:

```php
[
	'pattern' => 'my/drawer',
	'controller' => fn () => new MyDrawer()
],
[
	'pattern'    => ''another/drawer',
	'controller' => MyDrawer::class
]
```

### For review team

### DIFF
--- a/src/Panel/Dialog.php
+++ b/src/Panel/Dialog.php
@@ -50,6 +50,20 @@ class Dialog extends Json
 		$pattern = trim($prefix . '/' . ($options['pattern'] ?? $id), '/');
 		$type    = str_replace('$', '', static::$key);
 
+		// create load/submit events from controller class
+		if ($controller = $options['controller'] ?? null) {
+			if (is_string($controller) === true) {
+				if (method_exists($controller, 'for') === true) {
+					$controller = $controller::for(...);
+				} else {
+					$controller = fn (...$args) => new $controller(...$args);
+				}
+			}
+
+			$options['load']   ??= fn (...$args) => $controller(...$args)->load();
+			$options['submit'] ??= fn (...$args) => $controller(...$args)->submit();
+		}
+
 		// load event
 		$routes[] = [
 			'pattern' => $pattern,
--- a/tests/Panel/DialogTest.php
+++ b/tests/Panel/DialogTest.php
@@ -9,6 +9,29 @@ use Kirby\Filesystem\Dir;
 use Kirby\TestCase;
 use PHPUnit\Framework\Attributes\CoversClass;
 
+class TestDialog
+{
+	public function load()
+	{
+		return [
+			'component' => 'k-test-dialog',
+		];
+	}
+
+	public function submit()
+	{
+		return 'success';
+	}
+}
+
+class TestDialogWithFor extends TestDialog
+{
+	public static function for(): static
+	{
+		return new static();
+	}
+}
+
 #[CoversClass(Dialog::class)]
 class DialogTest extends TestCase
 {
@@ -177,12 +200,6 @@ class DialogTest extends TestCase
 
 	public function testRoutesWithoutHandlers(): void
 	{
-		$area = [
-			'dialogs' => [
-				'test' => []
-			]
-		];
-
 		$routes = Dialog::routes(
 			id: 'test',
 			areaId: 'test',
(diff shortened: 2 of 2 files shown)


src/Panel/Dialog.php
```diff
@@ -50,6 +50,12 @@ public static function routes(
 		$pattern = trim($prefix . '/' . ($options['pattern'] ?? $id), '/');
 		$type    = str_replace('$', '', static::$key);
 
+		// create load/submit events from class handler
+		if ($handler = $options['handler'] ?? null) {
```
distantnative: `Discuss:` We weren't sure about the term `handler`.

bastianallgeier: First thought while looking at it again, do we want to use Controller here?

distantnative: I am not sure if I can follow your thought. A controller in Kirby's context is something quite different, isn't it? (a closure that provides an array of data). Whereas this is either directly a class name or returns a class instance of a suitable dialog/drawer class.

bastianallgeier: Yes, that was my original thought as well, but we also started using the same term for the API (see Changes) and the Panel (see Search) with classes.

distantnative: Hmm yes, with Search is true. But would it be weird that then some controllers are located in `Panel\Controllers`, others though we pull in from `Panel\Ui\Dialogs`?

bastianallgeier: I think we just really need to go towards more consistency in the long run here. I quite like the differenciation between Ui backend components and Controller classes for the routes, which return them. I think I would not mind if they all end up in `Panel\Controllers` in the end.

src/Panel/Dialog.php
```diff
@@ -50,6 +50,20 @@ public static function routes(
...
+				if (method_exists($handler, 'for') === true) {
+					$handler = $handler::for(...);
+				} else {
+					$handler = fn (...$args) => new $handler(...$args);
+				}
+			}
+
+			$options['load']   ??= fn (...$args) => $handler(...$args)->load();
+			$options['submit'] ??= fn (...$args) => $handler(...$args)->submit();
+		}
```
bastianallgeier:
```suggestion
		if ($controller = $options['controller'] ?? null) {
			if (is_string($controller) === true) {
				if (method_exists($controller, 'for') === true) {
					$controller = $controller::for(...);
				} else {
					$controller = fn (...$args) => new $controller(...$args);
				}
			}

			$options['load']   ??= fn (...$args) => $controller(...$args)->load();
			$options['submit'] ??= fn (...$args) => $controller(...$args)->submit();
		}
```

src/Panel/Dialog.php
```diff
@@ -50,6 +50,20 @@ public static function routes(
...
+				if (method_exists($controller, 'for') === true) {
+					$controller = $controller::for(...);
+				} else {
+					$controller = fn (...$args) => new $controller(...$args);
+				}
+			}
+
+			$options['load']   ??= fn (...$args) => $controller(...$args)->load();
+			$options['submit'] ??= fn (...$args) => $controller(...$args)->submit();
+		}
```
afbora, line 65: IMHO we should leave using anonymous classes for such as extendings. Using abstract classes or interfaces would be extremely useful for usability, control, stability and documentation. So, Dialog controllers should always extend something like `Kirby\Panel\DialogController`

Something like that:

````php
namespace Kirby\Panel;

interface DialogControllerInterface
{
	public function load(): array;
	public function submit(): mixed;
}
````

````php
namespace Kirby\Panel;

abstract class DialogController implements DialogControllerInterface
{
	public static function for(): static
	{
		return new static();
	}
}
````

distantnative: @afbora Anonymous classes are hard to test, I don't think I'd like them for this. And yes, we will have an abstract class as basis: https://github.com/getkirby/kirby/pull/7380/files#diff-19095a59579a137667409ec6836a4d9c0d527cd85ed4447647aa2583dd14c774

afbora: Wow, that's great. That's what I meant 👏

bastianallgeier (approved):

PR #7389, opened 2025-07-16
feat: New `<k-stats-field>` component

bastianallgeier:
## Description

The new `<k-stats-field>` component is basically rebuilding what happened in the `<k-stats-section>` but without the API props loading. We can plug it in as soon as the backend code for the field is also there (coming in the next PR) 

### ✨ Enhancements

- New `<k-stats-field>` component 
- The `<k-label>` component now also accepts `false` for the input prop. This will switch the label element to an h2 and also remove the `<abbr>` for the invalid state. All fields that don't store any values will need this. We can also use it for the info field, which did still create its own headline.

### 🐛 Bug fixes

- Removed destructive CSS rules in the lab for `h2` that easily bleed into component rules.

### For review team

### DIFF
--- a/panel/src/components/Forms/Field/StatsField.vue
+++ b/panel/src/components/Forms/Field/StatsField.vue
@@ -1,0 +1,24 @@
+<template>
+	<k-field
+		:id="id"
+		:help="help"
+		:input="false"
+		:label="label"
+		:name="name"
+		type="stats"
+	>
+		<k-stats :reports="reports" :size="size" />
+	</k-field>
+</template>
+
+<script>
+import { help, id, label, name } from "@/mixins/props.js";
+import { props as statsProps } from "@/components/Layout/Stats.vue";
+
+/**
+ * @example <k-stats-field label="Stats" :reports="reports" />
+ */
+export default {
+	mixins: [help, id, label, name, statsProps]
+};
+</script>
--- a/panel/src/components/Lab/PlaygroundView.vue
+++ b/panel/src/components/Lab/PlaygroundView.vue
@@ -113,10 +113,10 @@ export default {
 </script>
 
 <style>
-.k-lab-examples h2 {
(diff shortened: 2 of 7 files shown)


panel/src/components/Lab/PlaygroundView.vue
```diff
@@ -113,13 +113,6 @@ export default {
 </script>
 
 <style>
-.k-lab-examples h2 {
-	margin-bottom: var(--spacing-6);
-}
-.k-lab-examples * + h2 {
-	margin-top: var(--spacing-12);
-}
```
distantnative: `Todo:` This removal breaks parts of the Lab, e.g. https://sandbox.test/panel/lab/internals/panel/api and other of its tabs.

panel/src/components/Forms/Field/StatsField.vue
```diff
@@ -0,0 +1,24 @@
...
+
+<script>
+import { help, id, label, name } from "@/mixins/props.js";
+import { props as statsProps } from "@/components/Layout/Stats.vue";
+
+/**
+ * @example <k-stats-field label="Stats" :reports="reports" />
+ */
+export default {
+	mixins: [help, id, label, name, statsProps]
```
distantnative, line 22:
```suggestion
import { props as StatsProps } from "@/components/Layout/Stats.vue";

/**
 * @example <k-stats-field label="Stats" :reports="reports" />
 */
export default {
	mixins: [help, id, label, name, StatsProps]
```

panel/src/components/Lab/PlaygroundView.vue
bastianallgeier: Ah, good that you spotted the issue. I was not able to find where this rule applied. I will fix it.

distantnative (approved):

PR #7391, opened 2025-07-16
refact!: New `Kirby\Panel\Areas` class

distantnative:
## Description

One more step for the Panel namespace refactoring. I think the `Area` class itself needs more discussion how to do best. So this PR is just the `Areas` class alone for now.

### ♻️ Refactored

- New `Kirby\Panel\Areas` class

### 🚨 Breaking changes

- `KIrby\Panel\Panel` returns `Kirby\Panel\Areas` object instead of an array.
- Moved `KIrby\Panel\Panel::area()` to Kirby\Panel\Areas::area()`
- Moved `KIrby\Panel\Panel::buttons()` to Kirby\Panel\Areas::buttons()`

### For review team


bastianallgeier:
Do we want to introduce an Area object here as well or would that be a later step?

distantnative:
Read my first sentence in the first post 😂😂😂

Yes, but I think the one we created in Januar is weird. I think once I figured out better versions of the menu classes, it will become clearer how the Area class should look like.

bastianallgeier:
Tell me you didn't read the description, without telling me you didn't read the description. 🙈 Sorry about that! Totally makes sense.

bastianallgeier (approved):

PR #7392, opened 2025-07-16
feat: Stats Field

bastianallgeier:
### 🎉 Features

- We’ve turned the stats section into an additional new field. The stats section keeps existing, but with the new stats field there are more layout possibilities and options for simpler blueprint setups. The field has the exact same options as the stats section. 

### For review team

- [x] Add lab and/or sandbox examples (wherever helpful) (There's a new stats-field branch in the sandbox)

### DIFF
--- a/src/Form/Field/StatsField.php
+++ b/src/Form/Field/StatsField.php
@@ -1,0 +1,74 @@
+<?php
+
+namespace Kirby\Form\Field;
+
+use Kirby\Form\FieldClass;
+use Kirby\Panel\Ui\Stats;
+
+/**
+ * Stats field
+ *
+ * @package   Kirby Field
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @since     5.1.0
+ */
+class StatsField extends FieldClass
+{
+	/**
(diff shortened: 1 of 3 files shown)


src/Form/Field/StatsField.php
```diff
@@ -0,0 +1,74 @@
...
+	 */
+	protected array|string $reports;
+
+	/**
+	 * The size of the report cards. Available sizes: `tiny`, `small`, `medium`, `large`
+	 */
+	protected string $size;
+
+	/**
+	 * Cache for the stats object
```
distantnative:
```suggestion
	 * Cache for the Stats UI component
```

src/Form/Field/StatsField.php
```diff
@@ -0,0 +1,74 @@
...
+
+	public function reports(): array
+	{
+		return $this->stats()->reports();
+	}
+
+	public function size(): string
+	{
+		return $this->stats()->size();
+	}
```
distantnative, line 56: `Nitpick:` Aren't we just taking a round trip here and `return $this->size` would do the same?

bastianallgeier: Right now, it's indeed just a roundtrip, but the Stats class might further compute this later. I really just wanted to think of it as a proxy here and let the Stats class do the work.

distantnative (approved):

PR #7403, opened 2025-07-19
feat: Load panel translation from query parameter

### LINKED ISSUE #6616: Multilingual registration auth code template is not translated
## Description
Hi, I have a multilingual site. 

I register users on the frontend. 
I followed the docs:  https://getkirby.com/docs/cookbook/forms/user-registration#frontend-login-with-code

It all works great. My default language is "de". I am trying to get everything translated to "fr" and "it". 

Within the controller I set the language dynamically based on where the user registers: 
`'language' => kirby()->language()->code(),`

The emails get send in the corresponding language.

There's one template I am not able to serve in the correct language.
It's immediately displayed after sending the registration form: 
/panel/login

Bildschirmfoto 2024-08-14 um 15 07 10

It only displays in the main language (de). 

In my opinion this template should be served in the email's associated panel language. Like the emails. 

I've checked the users and all of them have the correct panel language set. 
I hope somebody can help me on the topic. 

The screen that follows is half in the main language, and half in the language of the panel user. 
So maybe there's something a little of? 

Bildschirmfoto 2024-08-14 um 16 00 35

**Expected behavior**  
In my opinion this template should be served in the email's associated panel language.

**This is the support forum Link**
https://forum.getkirby.com/t/multilingual-registration-auth-code-template-is-not-translated/32252/1

**This is what Sonja sais**
> I can reproduce that, even the weird mixed language interface. Seems to be a bug. Could you create an issue on GitHub, please?

## Your setup
**Kirby Version**  
4.3.0

bastianallgeier:
## Description

It sometimes might be really helpful to override the panel interface translation with a query parameter. Especially if you want to send users to a translated login (see: https://github.com/getkirby/kirby/issues/6616)

### ✨ Enhancements

- You can now set the Panel interface translation on demand with the new `translation` query parameter. E.g. `https://example.com/panel/login?translation=it` https://github.com/getkirby/kirby/issues/6616

### For review team

### DIFF
--- a/src/Cms/AppTranslations.php
+++ b/src/Cms/AppTranslations.php
@@ -88,6 +88,12 @@ trait AppTranslations
 	 */
 	public function panelLanguage(): string
 	{
+		$translation = $this->request()->get('translation');
+
+		if ($translation !== null && $this->translations()->find($translation)) {
+			return $translation;
+		}
+
 		if ($this->multilang() === true) {
 			$defaultCode = $this->defaultLanguage()->code();
 
--- a/tests/Cms/App/AppTranslationsTest.php
+++ b/tests/Cms/App/AppTranslationsTest.php
@@ -493,5 +493,33 @@ class AppTranslationsTest extends TestCase
 			]
 		]);
 		$this->assertSame('it', $app->panelLanguage());
+
+		// override with the translation query parameter
+		$app = new App([
+			'roots' => [
+				'index' => '/dev/null'
+			],
+			'request' => [
+				'query' => [
+					'translation' => 'it'
+				]
+			]
+		]);
+
+		$this->assertSame('it', $app->panelLanguage());
+
+		// override with invalid translation query parameter
+		$app = new App([
+			'roots' => [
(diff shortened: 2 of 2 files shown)


src/Cms/AppTranslations.php
```diff
@@ -88,6 +88,12 @@ protected function i18n(): void
 	 */
 	public function panelLanguage(): string
 	{
+		$translation = $this->request()->get('translation');
+
+		if ($translation && Translation::exists($translation)) {
```
distantnative: `Suggestion:` I like when we are really explicit what we check

```suggestion
		if ($translation !== null && Translation::exists($translation)) {
```

distantnative: `Question:` Why do we not use `$app->translations()` to check for the existence of the translation in the actual app context (instead of the new method that checks if a file is there?

bastianallgeier: The answer is something about trees and the wood. 🙈 I've just fixed it.

distantnative (approved):

PR #7409, opened 2025-07-20
refact!: Non-static `Kirby\Panel\Panel`

distantnative:
## Description

I didn't touch the methods yet that other PRs will change anyways (e.g. response/routes/router).

### ✨ Enhancements

- New `$kirby->panel()` method to access `Kirby\Panel\Panel` object

### 🚨 Breaking changes

- Converted methods in `Kirby\Panel\Panel` from static to non-static

### For review team

### DIFF
--- a/src/Panel/Home.php
+++ b/src/Panel/Home.php
@@ -31,12 +31,11 @@ use Throwable;
 class Home
 {
 	protected App $kirby;
-	protected User|null $user;
 
-	public function __construct()
-	{
-		$this->kirby = App::instance();
-		$this->user  = $this->kirby->user();
+	public function __construct(
+		protected Panel $panel
+	) {
+		$this->kirby = $this->panel->kirby();
 	}
 
 	/**
@@ -49,7 +48,7 @@ class Home
 	 */
 	public function alternative(): string
 	{
-		$permissions = $this->user->role()->permissions();
+		$permissions = $this->user()->role()->permissions();
 
 		// no access to the panel? The only good alternative is the main url
 		if ($permissions->for('access', 'panel') === false) {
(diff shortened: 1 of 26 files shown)


src/Panel/Home.php
```diff
@@ -239,10 +238,19 @@ public function url(): string
...
 	}
+
+	/**
+	 * Returns the current user
+	 * @since 6.0.0
+	 */
+	public function user(): User|null
+	{
+		return $this->kirby->user();
+	}
```
bastianallgeier, line 255: Wouldn't it be enough to get that once in the methods above and store it in a var instead of adding another method?

src/Panel/Panel.php
```diff
@@ -119,7 +126,7 @@ protected static function garbage(): void
 	 */
 	public static function go(string|null $url = null, int $code = 302): void
 	{
-		throw new Redirect(static::url($url), $code);
+		throw new Redirect(App::instance()->panel()->url($url), $code);
 	}
```
bastianallgeier, line 130: Do you think it would break too much if we convert this to non-static as well?

bastianallgeier (commented):
Already feels so much better 💛

src/Panel/Home.php
distantnative: I had this in the constructor, but the other can change during a request (e.g. impersonation) and then the troubles start (quite common in our tests) as the instance is also cached in the Panel class and the Panel instance is cached in the App class.

src/Panel/Panel.php
distantnative: I thought about it and am on the fence about it, but I think the static syntax might be easier for the occasions where users use it?

bastianallgeier: If we start to use Controllers in all places, the base controller class for all panel routes could already have the panel instance. Then it would only be a `$this->panel()->go()` call. But for all the closure actions, you would of course still be more work. Although, we could reuse our `App::instance()` standard and have `Panel::instance()` for such cases.

bastianallgeier: But maybe we are having this discussion too early on. We can still decide and change this later.

PR #7411, opened 2025-07-20
refactor: Convert array-based fields to class-based component

afbora:
This is just first step of the migration of the array-based config fields to backward compatible class-based components. This only includes text field for POC. 

@getkirby/backend I need your first glance reviews to continue or not 🙂 

### My personal plan in my mind
- 6.0 Introduce the new class-based fields
- 7.0 Deprecate the extends/mixins usage for array-based fields
- 8.0 Remove array-based fields entirely

### Todos

- [x] Migrate text field
- [x] Migrate mixins
- [x] Use named arguments for class-based fields
- [ ] Write unit tests


src/Form/Field/TextField.php
```diff
@@ -0,0 +1,198 @@
...
+			return $value;
+		}
+
+		$converter = $this->converters()[$this->converter()];
+
+		if (is_array($value) === true) {
+			return array_map($converter, $value);
+		}
+
+		return call_user_func($converter, trim($value ?? ''));
```
bastianallgeier:
```suggestion
		return $converter(trim($value ?? '')));
```

bastianallgeier:
The class looks solid. That's a great start. 

The idea with the config path is nice too. I wonder if we should move that to a different PR though. To keep the old configs working will be quite tricky indeed when you need to extend a component and we maybe need more experiments with that. As long as we keep in mind that we have to get this done, I would not mind to put that a bit further back and focus on the class migrations first. 

I would also start turning repeatable things into mixins: 

- counter
- converter
- font
- pattern
- spellcheck
- minlength
- maxlength

Those will help us a lot when we need to compose more field classes later.

afbora:
@bastianallgeier I'm glad you liked it. I don't quite understand why we're moving the `getConfigFilePath()` change to a different PR. Will we be following a different path for backward compatibility? Could you elaborate on that a bit?

bastianallgeier:
The PRs will be a lot easier to merge when we keep the focus on just one thing. The class code is almost mergable in my opinion. With your getConfigFilePath solution, I'm not 100% sure yet.

afbora:
I see. You're saying we can evaluate different solutions after moving it to classes. OK.

afbora:
@bastianallgeier I've cleaned up the codes and moved field methods to mixins. Should I continue to other fields as well?

distantnative:
If we are doing this transition (which I think is great!), any way we could have named arguments instead of the $params array?

afbora:
1. Named arguments are nice, but I don't like them when there are too many arguments, to be honest.
2. I would prefer to use the same constructor for each field rather than a different one.

@distantnative  In short, it's possible, but I'm not sure.

distantnative:
It should definitely be the same for all fields the same in the end. I think there are a lot of benefits with named arguments instead, if documentation alone. And that's why we should make the decision and find a solution early on before we migrate all the fields - now it's only 3-4 to change.

bastianallgeier:
I would definitely also vote for named parameters. We already head in a similar direction with other refactoring parts. 

We can still always work with an additional `::factory` method which will accept an array and then pass that array to the constructor.

afbora:
- `::construct`- accepts named arguments
- `::factory` - accepts array

Correct?

bastianallgeier:
Quick example: 

```php
$field = new TextField(
  name: 'test',
  required: true
);

$field = TextField::factory([
  'name' => 'test',
  'required' => true
]);

// under the hood
class TextField {
  public function __construct(
    protected string $name,
    protected bool $required = false,
  ) { 
  }

  public static function factory(array $props): static
  {
    return new static(...$props);
  }
}
```

afbora:
I've pushed the changes with https://github.com/getkirby/kirby/pull/7411/commits/697ec68ebff06768df28319a802683d8689bd423

What about using named arguments in FieldClass, I guess that will be breaking change, right?

src/Form/Field/TextField.php
```diff
@@ -31,24 +31,37 @@ class TextField extends FieldClass
...
-		parent::__construct($params);
+	public function __construct(
+		mixed $converter = null,
+		bool $counter = true,
+		string|null $font = null,
+		int|null $maxlength = null,
+		int|null $minlength = null,
+		string|null $pattern = null,
+		bool $spellcheck = false,
+		...$parentParams
```
distantnative: This negates the benefits of named parameters. It is a bit tedious, but we should repeat all parameters that can be passed so that IDEs, our docs etc. will be able to pick them up.

src/Form/Field/TextField.php
```diff
@@ -0,0 +1,117 @@
...
+		string|null $name = null,
+		string|null $pattern = null,
+		array|string|null $placeholder = null,
+		bool $required = false,
+		Fields|null $siblings = null,
+		bool $spellcheck = false,
+		bool $translate = true,
+		array|null $when = null,
+		string|null $width = null,
+		mixed $value = null
```
afbora: Now, the `value` argument will return `null` as a fallback. The `FieldClass::__construct` only fills if the value parameter is passed. This seems like it might cause a break.

https://github.com/getkirby/kirby/blob/3407c4f70b2df37085dd025adf6c0a097b3189f7/src/Form/FieldClass.php#L63-L65

src/Form/Mixin/Converter.php
```diff
@@ -0,0 +1,67 @@
...
+			'ucfirst' => function ($value) {
+				return Str::ucfirst($value);
+			},
+			'upper'   => function ($value) {
+				return Str::upper($value);
+			},
+		];
+	}
+
+	public function convert(mixed $value): mixed
```
bastianallgeier, line 38: I would list all allowed types here. 

```suggestion
	public function convert(string|array|null $value): mixed
```

src/Form/Field/TextField.php
bastianallgeier: I'm not sure I fully understand. You mean we should also check for a null value here?

src/Form/Field/TextField.php
```diff
@@ -0,0 +1,117 @@
...
+	use Maxlength;
+	use Minlength;
+	use Pattern;
+	use Spellcheck;
+
+	public function __construct(
+		string|null $after = null,
+		bool $autofocus = false,
+		string|null $before = null,
+		mixed $converter = null,
```
bastianallgeier:
```suggestion
		string|null $converter = null,
```

src/Form/Field/TextField.php
afbora: No. Before we weren't doing a null check but key exists check. So if `$params` doesn't have a value prop, it doesn't touch the value. But with this refactor, `$value` default value is `null`. The class will set the value always `null` if it doesn't passed.

afbora:
If I convert all the fields in this PR, this PR will grow significantly. Currently, only the textfield and mixins are ready (I'll write unit test later)

I think we should first discuss how to integrate backward compatibility and finalize this PR. Then, we can convert the other fields in another PR. 

What do you think?

bastianallgeier:
@afbora I think I would do all the refactoring and unit tests for the field classes step by step. Maybe a PR for every field? Backward compatibility could be a last step afterwards. To go back into all field classes and add another method should be no problem at that point.

afbora:
@bastianallgeier OK. So I'll continue with unit tests for just this PR changes.

src/Form/Field/TextField.php
afbora: @bastianallgeier @distantnative Do you think that's not an issue?

afbora:
@distantnative @bastianallgeier Current `Field::factory()` class trying to create instance from `array`. But our new constructor is using named arguments.

````php
	string $type,
	array $attrs = [],
	Fields|null $siblings = null
): static|FieldClass {
	$field = static::$types[$type] ?? null;

	if (is_string($field) && class_exists($field) === true) {
		$attrs['siblings'] = $siblings;
		return new $field($attrs);
	}

	return new static($type, $attrs, $siblings);
}
````
https://github.com/getkirby/kirby/blob/lab/afbora/refactor-array-based-config-fields/src/Form/Field.php#L264

So I think we should switch named-arguments for current class-based fields (blocks-layout-entries). Is there anyone against it? Or any idea about solving this issue?

New behavior:

````php
	string $type,
	array $attrs = [],
	Fields|null $siblings = null
): static|FieldClass {
	$field = static::$types[$type] ?? null;

	if (is_string($field) && class_exists($field) === true) {
		$attrs['siblings'] = $siblings;
		unset($attrs['type']);
		return new $field(...$attrs);
	}

	return new static($type, $attrs, $siblings);
}
````

afbora:
Maybe we can check first argument whether array or not but this seems like a hack.

distantnative:
@afbora I think it is right that we should change the existing class-based fields also to use named parameters for their constructor.

bastianallgeier:
We could add a factory method to the FieldClass and use that in Field::factory instead. It would always receive an array. The old Field classes could then just pass on the array to their constructor as long as they are not refactored yet and the new ones could already use the spread operator to support the named arguments.

afbora:
> I think I would do all the refactoring and unit tests for the field classes step by step. Maybe a PR for every field? Backward compatibility could be a last step afterwards. To go back into all field classes and add another method should be no problem at that point.

@bastianallgeier I've only converted the text field, but it's failing the unit tests due to a lack of backward compatibility. Since the unit tests on other fields also failed, moving forward seems risky.

> We could add a factory method to the FieldClass and use that in Field::factory instead. It would always receive an array. The old Field classes could then just pass on the array to their constructor as long as they are not refactored yet and the new ones could already use the spread operator to support the named arguments.

I added `FieldClass::factory,` but this time, the `Field::factory` class is affected. There's no inheritance between them, but the IDE is warning me. Any ideas?

image
image
image

@bastianallgeier I'm a little lost. Could you give me some guidance?

distantnative:
I have created a new attempt at this: https://github.com/getkirby/kirby/pull/7608 - first just taking care of `FieldClass` and the existing field classes, before we start converting the array-based fields.

bastianallgeier:
We started a new approach to tackle this.

Issue #3138, opened 2021-02-03
Writer field: Option to disable markdown input


mlthrk:
**Describe the bug**  
When writing in or pasting to a Writer field, text is parsed for the mark characters of the field’s active marks and automatically converted. Thus it is not possible to use these characters for other purposes, e.g. writing the “gender star”. Escaping the mark characters with a backslash seems to be ignored.

**To Reproduce**  
1. Add a Writer field to an existing blueprint
2. Write or paste text containing more than one asterisk

**Expected behavior**  
1. Add option to disable markdown input
2. Escape markdown characters

**Screenshots**  
![Gender*](https://user-images.githubusercontent.com/1251280/106764952-98a3b780-6638-11eb-96b0-b56883d80993.gif)

**Kirby Version**  
3.5+

afbora:
1. You can disable marks with `marks: false` witn new 3.5.2 release (still RC, will release Feb 9)
2. When you disabled `marks`, you'll still see marks on preview editor, this issue will be fixed 3.5.x.

mlthrk:
Thanks for looking into this. Not sure if we are talking about the same issue?

Disabling marks completely with `marks: false` already works for me with the current 3.5.1 release. I just tried out 3.5.2-rc.1 and didn’t notice much of a difference.

In [Writer/Editor.js](https://github.com/getkirby/kirby/blob/0d2751af7cb478df14f30fac47fe9a2cd3b38114/panel/src/components/Writer/Editor.js) there are two default options that seem to do what I’m after, though. Setting both `disableInputRules` and `disablePasteRules` to `true`, markdown parsing is disabled but it seems still possible to use the marks formatting of the Writer field. If this was an appropriate way to do this, it’d be great to be able to set these options in a blueprint.

davidjablonski:
I hope it’s okay for me to add to this – not sure whether this qualifies as an additional issue. 
I’m running into the same issue as described above with the default text block in `fields:block`. I haven’t found a way to disable marks for this block type or am I missing something?
Thanks!

moritzebeling:
Since we have text formatting via the tooltip, is the markdown-thing still a feature that is needed in the writer field? Maybe the md parsing can be removed all together?

At least in German the gender-star problem makes it really hard to use the writer field. Thanks for looking into it!

lukasbestle:
[...]

PR #7412, opened 2025-07-21
fix: Better writer markdown input rules for bold and italic

bastianallgeier:
### 🐛 Bug fixes

- Better writer markdown input rules for bold and italic https://github.com/getkirby/kirby/issues/3138 

### For review team


distantnative:
@bastianallgeier What's the case that this fixes?

bastianallgeier:
Pasting this for example: 

> Schüler*innen und Student*innen

It also works more relibaly for me if I start a new line and add real italic text afterwards.

distantnative (approved):
Seems working well to me. But @afbora usually has a good hand at figuring out if there is any regression. Maybe he should also have a look at it.

PR #7414, opened 2025-07-21
feat: Add support for refresh-based redirects in Panel

### LINKED ISSUE #7274: `Invalid CSRF token` error when logging in after panel session expired
## Description
When a panel session expires (which happens often during development), I'm redirected to the login view. When I try to log in I get a `Invalid CSRF token` error, so I manually need to reload the page.

**Expected behavior**  
Being able to log in without reloading the page first. Or: not being redirected to the login view, so that it's clear I need to navigate/reload myself.

## To reproduce
1. Open the panel and log in
2. Let the session expire
3. Try to log in again

## Your setup
**Kirby Version**  
5.0.0-rc.3

**Console output**  
?

 - Device: MacBook Pro 16" (Intel)
 - OS: macOS 15.5 (24F74)
 - Browser: Safari
 - Version: Version 18.5 (20621.2.5.11.8)

afbora:
## Description

Introduces the ability to perform refresh-based redirects by extending the `Panel::go()` and `Redirect` classes to accept a refresh parameter. The View class now returns a refresh response when appropriate and `Response::refresh` has been added to generate the correct HTTP headers. Updates login area config to use the new refresh redirect.

~I'll add unit tests after the review is OK.~

### ✨ Enhancements
- Add support for refresh-based redirects in Panel
- New `Response::refresh()` method added

### 🐛 Bug fixes
- Invalid CSRF token error when logging in after panel session expired #7274

### For review team


bastianallgeier:
@afbora do you have a good way to reproduce the initial problem? I've been trying a couple things but I cannot reproduce it, which makes it really hard to test the before and after state.

afbora:
@bastianallgeier Ok so easy. Pelase follow the steps:

1. Login to panel
2. Open console and delete the session cookie from application tab
3. Try to enter any panel page via click (do not refresh)
4. See you'll be redirected to login page
5. Try to login again
6. You'll see invalid csrf message

With this PR, the login page will be refreshed instantly and you can able to login without error.

bastianallgeier:
Now I can replicate it and your fix works great. I just changed the argument order and accepted any Throwable in the exception. I think with unit tests we are ready to merge this.

distantnative:
What about rather using a boolean for our method parameter and turn that into an integer inside the method?

afbora:
@distantnative You mean "`Panel::go()` takes bool and sends 0 to refresh", right? But it's like users can say "Go to this page of the panel, wait 3-5 seconds and then refresh." It seems possible, isn't it?

distantnative:
Ok I see why it is an integer. Not sure about the use case though. Is a time-based reload delay not very unreliable if the delay is meant to ensure some other actions to be performed first?

afbora:
That's right. Both ways are fine for me.

bastianallgeier:
I also think that it's a pretty limited usecase, but it's at least in line with the HTTP response standard. Even if we only use it without a delay, it makes sense for me to follow that.

distantnative:
Ok then I would only opt for `int|false` instead of `int|null`.

afbora:
@bastianallgeier @distantnative Tests added and ready to review now.

afbora:
We can merge this feature into 5.0.3 or 5.1.0. Feel free to merge it or move next milestone 👍

bastianallgeier (dismissed):
I think it would be great to add this to 5.0.3. But I cannot change the base here on Github without messing up the commits. Could you try to create a version that is based on develop-patch?

PR #7419, opened 2025-07-25
lab: Getting rid of sections

bastianallgeier:
This is a lab branch with a set of ideas to get rid of the files and pages sections and migrate them over to fields. But there are a lot of steps in between that are needed to refactor the code in a way that we can still keep the sections around. Here's a summary of the ideas that I had to tackle this. 

## Individual steps 

- [x] https://github.com/getkirby/kirby/pull/7433

## New UI Classes

New UI Classes help to create the correct props for our `<k-collection>` component, which we use to display pages, files and users. 

- `Kirby\Ui\ModelsCollection`
- `Kirby\Ui\PagesCollection`
- `Kirby\Ui\FilesCollection`
- `Kirby\Ui\UsersCollection`

Those classes take over a lot of the work that is normally be done in the pages and files section, but in a lot more controllable way. They should purely take care of creating the correct props. No additional logic should be solved here. 

Ideally, those classes are so universally usable, that each new backend route could later use them to create a list of models with the correct layout. 

The most complex part here is the table logic. I've outsourced it into two additional classes: 

- `Kirby\Ui\ModelsTable`
- `Kirby\Ui\PagesTable`

Again, those classes are here to mostly prepare all the needed props for the tables. They need a lot more logic though to fetch the values for the rows. It would maybe make sense to add another class one level below, which is purely just a Table UI Class without any or hardly any logic.

## New Collector classes

This is a brand new concept. I like it so far, but I'm not 100% sure about the naming yet. The idea is to outsource the complex data fetching logic that we need for pages, files and users in the Panel - including sorting, filtering and searching. 

Those classes can serve as data source for multiple use cases. Not just for sections. But right now they neatly take care of a huge chunk of the section logic. We can also write tests for this in a much more elegant way than we did for the sections. (still needs to be done) I really think that this separation between fetching data and displaying data is pretty useful. If you come up with a better name though, I'm totally up for it. 

## Next steps

The more logic we've separated from the sections' array config, the easier it will be to have new Field classes for them, but still keep the old array configuration around. I think there are a couple more parts that could be improved / outsourced: 

- Min/Max validation
- The `add` logic for pages
- The page status ('all', 'draft', 'published', 'listed', 'unlisted') could maybe become an Enum
- The `upload` logic in the files section

## Failing tests

There are only a few minor failing unit tests. I think we can debug this more easily as soon as the Table classes have their own unit tests.

### DIFF
--- a/config/areas/users/views.php
+++ b/config/areas/users/views.php
@@ -2,6 +2,7 @@
 
 use Kirby\Cms\App;
 use Kirby\Cms\Find;
+use Kirby\Panel\Collector\UsersCollector;
 use Kirby\Panel\Ui\Buttons\ViewButtons;
 use Kirby\Toolkit\Escape;
 
@@ -31,23 +32,16 @@ return [
 					},
 					'roles' => array_values($roles),
 					'users' => function () use ($kirby, $role) {
-						$users = $kirby->users();
 
-						if (empty($role) === false) {
-							$users = $users->role($role);
-						}
-
-						// sort users alphabetically
-						$users = $users->sortBy('username', 'asc');
-
-						// paginate
-						$users = $users->paginate([
-							'limit' => 20,
-							'page'  => $kirby->request()->get('page')
-						]);
+						$users = new UsersCollector(
+							role: $role,
+							sortBy: 'username asc',
+							page: $kirby->request()->get('page', 1),
+							limit: 20
+						);
 
 						return [
-							'data' => $users->values(fn ($user) => [
+							'data' => $users->paginated()->values(fn ($user) => [
 								'id'    => $user->id(),
 								'image' => $user->panel()->image(),
 								'info'  => Escape::html($user->role()->title()),
--- a/config/areas/users/views.php
+++ b/config/areas/users/views.php
@@ -4,6 +4,7 @@ use Kirby\Cms\App;
 use Kirby\Cms\Find;
 use Kirby\Panel\Collector\UsersCollector;
 use Kirby\Panel\Ui\Buttons\ViewButtons;
+use Kirby\Panel\Ui\UsersCollection;
 use Kirby\Toolkit\Escape;
 
 return [
@@ -32,7 +33,6 @@ return [
 					},
 					'roles' => array_values($roles),
 					'users' => function () use ($kirby, $role) {
-
 						$users = new UsersCollector(
 							role: $role,
 							sortBy: 'username asc',
@@ -40,15 +40,13 @@ return [
 							limit: 20
 						);
 
+						$collection = new UsersCollection(
+							users: $users->all()
+						);
+
 						return [
-							'data' => $users->paginated()->values(fn ($user) => [
-								'id'    => $user->id(),
-								'image' => $user->panel()->image(),
-								'info'  => Escape::html($user->role()->title()),
-								'link'  => $user->panel()->url(true),
-								'text'  => Escape::html($user->username())
-							]),
-							'pagination' => $users->pagination()->toArray()
+							'data'       => $collection->items(),
+							'pagination' => $collection->pagination()
 						];
 					},
 				]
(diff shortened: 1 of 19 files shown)


config/areas/users/views.php
```diff
@@ -31,30 +33,20 @@
...
-						$users = $users->sortBy('username', 'asc');
-
-						// paginate
-						$users = $users->paginate([
-							'limit' => 20,
-							'page'  => $kirby->request()->get('page')
-						]);
+						$collection = new UsersCollection(
+							users: $users->all()
+						);
```
distantnative: `Suggestion:` Might make it easier to understand what is what if we rather use `$component` here

```suggestion
						$component = new UsersCollection(
							users: $users->all()
						);
```

config/sections/mixins/pagination.php
```diff
@@ -15,7 +15,7 @@
 		 * Sets the default page for the pagination.
 		 */
 		'page' => function (int|null $page = null) {
-			return App::instance()->request()->get('page', $page);
+			return App::instance()->request()->get('page', $page ?? 1);
```
distantnative: `Suggestion:` Rather directly set as parameter default?

src/Form/Field/StatsField.php
```diff
@@ -0,0 +1,90 @@
...
+			label: $label,
+			model: $model,
+			name: $name,
+			siblings: $siblings,
+			when: $when,
+			width: $width
+		);
+
+		$this->reports = $reports;
+		$this->size    = $size;
```
distantnative: `Suggestion:` I would write these as promoted properties directly above.

src/Form/FieldAbstract/BaseField.php
```diff
@@ -0,0 +1,196 @@
+<?php
+
+namespace Kirby\Form\FieldAbstract;
```
distantnative: `Question:` Do we need this extra namespace? Couldn't this live also in `Kirby\Form\Field`?

src/Panel/Collector/FilesCollector.php
```diff
@@ -0,0 +1,61 @@
...
+
+	protected function collectByParent(): Files
+	{
+		return $this->parent->files();
+	}
+
+	protected function collectByQuery(): Files
+	{
+		return $this->parent->query($this->query, Files::class) ?? new Files([]);
+	}
```
distantnative, line 34: `Discussion:` Wondering if we would rather define the Model class (here `Files::class`) as e.g. a class constant and then the individual model collector classes wouldn't have to repeat this method with almost the same code.

src/Form/Field/StatsField.php
bastianallgeier: That was a mistake. The branch wasn't based on the right one. The additional stats field changes and abstracts should not be in there. Sorry for the confusion.

src/Panel/Collector/FilesCollector.php
bastianallgeier: I've tried it but since the UsersCollector needs a adjusted version anyway, it feels a bit too abstracted for two classes.

lukasbestle (commented):
Thanks for the summary, that was quite helpful for understanding this project.

My general thoughts:
- The `Ui` and `Collector` building blocks are great. I love how they already make the hard to test `config` spaghetti so much shorter.
- The `Ui` classes are so far a lot of boilerplate. But I see the potential. We could also do a lot of prop validation in them so that invalid configurations/prop combinations of components already explode on the backend and not in Vue where they are harder to debug.
- At first from your description I was wondering if `Collector` really belongs in the `Panel` namespace. But looking at the code it makes sense. It's quite specific to the Panel and really doesn't do anything generally useful that would need to be in lower code levels.
- The `Collector` name fits well IMO.
- I think we need to work on the interfaces some more. E.g. why does `Collector` have both `collect()` and `all()` as public methods? Th `Ui` classes have loads of public methods where I would mainly just expect `props()` as an exporter for Fiber to the frontend.
- Another thing we might want to think about is whether there is a cleaner way to pass all those props to the building block classes. Individual arguments instead of an array are great but very verbose. Maybe we could later refactor the section/field classes to have ways to get a `Collector` or component with all the props. But without being too magical of course. If not possible, individual props are still good and easy to understand at least.

PR #7420, opened 2025-07-26
refact: `Panel\Ui\Stat` optional `$model` property

distantnative:
## Description

@bastianallgeier I think the component would be more useful if a model isn't required. IMO it's only required when one wants to use the query functionality, but as UI element, a stat could also be used completely without that smart logic but by providing the final values directly.

### DIFF
--- a/src/Panel/Ui/Stat.php
+++ b/src/Panel/Ui/Stat.php
@@ -17,13 +17,13 @@ use Kirby\Toolkit\I18n;
 class Stat extends Component
 {
 	public function __construct(
-		public ModelWithContent $model,
 		public array|string $label,
 		public array|string $value,
 		public string $component = 'k-stat',
 		public string|null $icon = null,
 		public array|string|null $info = null,
 		public array|string|null $link = null,
+		public ModelWithContent|null $model = null,
 		public string|null $theme = null,
 	) {
 	}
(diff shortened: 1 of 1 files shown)


src/Panel/Ui/Stat.php
```diff
@@ -90,6 +90,10 @@ public function props(): array
 
 	protected function stringTemplate(string|null $string = null): string|null
 	{
+		if ($this->model === null) {
+			return $string;
+		}
+
 		if ($string !== null) {
 			return $this->model->toString($string);
```
afbora, line 98: Maybe simpler and with a single line?

```suggestion
			return $this->model?->toString($string) ?? $string;
```

distantnative: I first had it like that but then felt it was easier to read with its own if statement on top.

afbora: OK fine 👍

bastianallgeier (approved):

bastianallgeier:
Yikes. I merged it too soon. This could have totally gone into develop-minor. We don't have a good way to fix this in hindsight, right?

distantnative:
@bastianallgeier hust cherry pick the commit and add it to develop-minor?

PR #7421, opened 2025-07-26
fix: Update toolbar sticky selector for focus states

### LINKED ISSUE #7364: Clicking a button in the Writer toolbar has no effect when it is sticky at the top.
## Description
Clicking buttons in the Writer Field toolbar has no effect when setting the `toolbar` option to `inline: false` and the writer field content area is scrolled so that the toolbar is sticky at the top. 

This can be reproduced with Chrome Version 138.0.7204.101 (Official Build) (arm64) on macOS.
With FireFox (140.0.2 (aarch64), also macOS) clicking a button will execute the expected action but also change the scroll position of the content area. It jumps up.

**Screenshots**  
Image

Clicking on the _italic_ button will not apply the format to the selected text. Instead the toolbar disappears and reappears. I noticed the `ProseMirror-focused` css class being toggled on the `contenteditable="true"` div container of the Writer field when the italic button is clicked.

## Your setup
**Kirby Version**  

Kirby v5.0.1

afbora:
I've applied my solution that I've mentioned [in here](https://github.com/getkirby/kirby/issues/7364#issuecomment-3075833987).

Modified the CSS selector to make the toolbar sticky when any following sibling or itself is focused.

### 🐛 Bug fixes

- Clicking a button in the Writer toolbar has no effect when it is sticky at the top. #7364

### For review team

### DIFF
--- a/panel/src/components/Forms/Toolbar/Toolbar.vue
+++ b/panel/src/components/Forms/Toolbar/Toolbar.vue
@@ -134,7 +134,7 @@ export default {
 	--toolbar-text: var(--color-gray-400);
 	--toolbar-border: var(--panel-color-back);
 }
-.k-toolbar:not([data-inline="true"]):has(~ :focus-within) {
+.k-toolbar:not([data-inline="true"]):has(~ :focus-within, :focus) {
 	position: sticky;
 	top: var(--header-sticky-offset);
 	inset-inline: 0;


distantnative (commented):
Couldn't really reproduce the issue in the first place, but did not spot any regressions for me with this change (so if it fixes the original issue for you, seems to be good).

One thing I noted but also present on `develop-minor`: When clicking the toolbar button, the browser will jump to the top of the writer field.

afbora:
I can still reproduce the issue easily in develop-minor branch and this PR fixes the issue.

But I can't produce the your jump issue on develop-minor branch. What's your blueprint setup? Maybe browser issue?

bastianallgeier (dismissed):
Works great! Could you again change the base to develop-patch, so that we can include it in 5.0.3?

bastianallgeier (approved):

PR #7422, opened 2025-07-27
feat: Panel routes controllers

distantnative:
## Description

Moved away from the `controller` key from the previous PR to use the `action` directly and consistently across all Panel routes.

### ✨ Enhancements

- New `Kirby\Panel\Controller\Controller` classes that can be passed by their name as `action` argument of a Panel route (dialog, drawer, dropdown, request, search and view) and then handle the route action. This helps to move away from the anonymous closures in `config/areas/` and instead build handler classes that can be properly tested etc.

### ♻️ Refactored

- New `Kirby\Panel\Routes\Routes::controller()` helper method 
- New `Kirby\Panel\Routes\Routes::params()` helper method to use consistently across all Panel route types
- `Kirby\Panel\Routes\Routes::$type` is now empty by default

I didn't mark these as breaking changes as they're only breaking towards previous unreleased v6 changes.

## Docs

For all Panel route based extensions, we have to add documentation showcasing how to build a respective controller (e.g. DialogController) and pass it as `'action' => MyDialogController::class` in the route config.

### For review team

### DIFF
--- a/src/Panel/Controller/Controller.php
+++ b/src/Panel/Controller/Controller.php
@@ -27,12 +27,12 @@ abstract class Controller
 		$this->request = $this->kirby->request();
 	}
 
-	abstract public function load();
+	abstract public function load(): mixed;
 
 	/**
 	 * Submit successfully by default to allow for submit-less controllers
 	 */
-	public function submit()
+	public function submit(): mixed
 	{
 		return true;
 	}
--- a/src/Panel/Controller/Controller.php
+++ b/src/Panel/Controller/Controller.php
@@ -26,4 +26,14 @@ abstract class Controller
 		$this->kirby   = App::instance();
 		$this->request = $this->kirby->request();
 	}
+
+	abstract public function load();
+
+	/**
+	 * Submit successfully by default to allow for submit-less controllers
+	 */
+	public function submit()
+	{
+		return true;
+	}
 }
--- a/src/Panel/Controller/Controller.php
+++ b/src/Panel/Controller/Controller.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Kirby\Panel\Controller;
+
+use Kirby\Cms\App;
+use Kirby\Http\Request;
+
+/**
+ * @package   Kirby Panel
+ * @author    Nico Hoffmann <nico@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @since     6.0.0
+ * @unstable
+ *
+ * @codeCoverageIgnore
+ */
+abstract class Controller
+{
+	protected App $kirby;
+	protected Request $request;
+
+	public function __construct()
+	{
+		$this->kirby   = App::instance();
+		$this->request = $this->kirby->request();
+	}
+}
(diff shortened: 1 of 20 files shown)


bastianallgeier:
I think this is 99% there. I wonder if we should move to `load` as standard for all controllers instead of `options`, `view`, or `data` to add some more consistency.

distantnative:
@bastianallgeier I can understand the consistency argument. I think e.g. `ViewController::view()` or `RequestController::data()` not the strongest names. On the other hand, I think `DropdownController::options()` is a pretty good name and with these `::load()` might be a consistent name across controllers but also quite opaque and not as good to understand what it does/what's expected e.g. in the case of a dropdown etc.

bastianallgeier:
I agree that options is by far the best of those inconsistent method names. I always thought that we might some day also have a submit case for dropdowns. Think of a dropdown with a simple form for smaller tasks. If you take a look at the changes dropdown for the form controls or the plugin dropdown, I really believe that dropdowns could be way more interesting and useful as soon as we go beyond just options. Ideally, they could also have custom components at some point.

distantnative:
@bastianallgeier I went a bit further and actually moved  `::load()` and `::submit()` to the main Controller class, which also allowed me to move all of the logic to the `Routes` class - makes the individual Routes classes simpler.

src/Panel/Routes/DropdownRoutes.php
```diff
@@ -17,34 +15,17 @@ class DropdownRoutes extends Routes
...
 
 		foreach ($this->routes as $name => $params) {
-			$params  = $this->params($params);
+			$params  = $this->params($params, 'options');
 			$pattern = $this->pattern($params['pattern'] ?? $name);
 
 			$routes[] = $this->route(
 				pattern: $pattern,
-				action:  $params['action'],
+				action:  $params['action'] ?? fn () => 'The dropdown action handler is missing',
```
distantnative: One could also think whether here we should use `$params['load']` then. But then we have to add some legacy mapping from `action` to `load` in `Routes::params()`.

src/Panel/Controller/Controller.php
```diff
@@ -0,0 +1,39 @@
...
+	protected App $kirby;
+	protected Request $request;
+
+	public function __construct()
+	{
+		$this->kirby   = App::instance();
+		$this->request = $this->kirby->request();
+	}
+
+	abstract public function load();
```
distantnative: What are actually the allowed return types here? Array? True? String for error?

bastianallgeier: It depends on whatever we accept in our response handlers for each context. There's also a difference between load and submit. 

In case of the load event, true as response does not make sense. In case of the submit event, we've mapped it to a success response. We definitely also want to support a Response object as type for both.

Maybe we should make a documented list for this.

src/Panel/Routes/DropdownRoutes.php
distantnative: @bastianallgeier Implemented it.

src/Panel/Controller/Controller.php
distantnative: @bastianallgeier I realized that if we add the return type hints in the Controller class, all child classes need to implement it exactly the same. Which might become a bit wild if they all have to add `: Response|array|true` or so, even when just returning always an array in their case. So maybe it is better for now without any type hint in the parent class.

bastianallgeier: We need to test it again, but for return types, afaik the child classes can be stricter than the parent class. It just does not work for method arguments that way (which I personally find weird)

bastianallgeier (approved):

PR #7423, opened 2025-07-27
refact!: Panel search controllers

distantnative:
## Description

- [x] https://github.com/getkirby/kirby/pull/7422

I was thinking about https://getkirby.com/docs/reference/plugins/extensions/panel-search#extending-default-search-types__reusing-core-code - do you think in the future it. would be the best way for extending core search to tell users extending our search controller class and just setting that as new controller class?

### ♻️ Refactored

- Refactored `Kirby\Pane\Controller\Search` methods as individual `Kirby\Panel\Controller\SearchController` classes

### 🚨 Breaking changes

- Removed `Kirby\Pane\Controller\Search`

### For review team


bastianallgeier:
@distantnative yes, I actually think that's the way nicer and more reliable default way to extend the search.

bastianallgeier (approved):

Issue #7405, opened 2025-07-19
UUID for new page is the same as the latest page UUID


maxhoffmann:
## Description

After upgrading from version `4.8.0` to version `5.0.2`, creating a new page with these create dialog settings throws an error, that the page with this slug already exists:

```
create:
  title: "{{ page.uuid.id }}"
  slug: "{{ page.uuid.id }}"
```

I’ve checked the file inside the `__new__` folder and can confirm that it includes the same UUID as the page created before.

**Expected behavior**  

The newly created page should have a newly generated UUID as it used to in version 4.

## To reproduce

1. Create a blueprint with these settings:
```
create:
  title: "{{ page.uuid.id }}"
  slug: "{{ page.uuid.id }}"
```

2. Click on 'Create a new page' in the panel
3. When creating a second page, the error is displayed

## Your setup
**Kirby Version**: 5.0.2

afbora:
I can't reproduce the issue that you mentioned (throwing an error). I can able to create many pages. 
But I encountered another issue that generated content uuid and page title/slug is different for me.

bastianallgeier:
@afbora I think your issues are both closely related. The generated UUID is not cached at this moment and that's why you get two when you ask for it twice in those queries.

maxhoffmann:
FYI: deleting the `__new__` folder fixed the problem for me

afbora:
@bastianallgeier @distantnative When using `Page::factory` in the `PageCreateDialog::model` method, the `uuid` was coming up empty. This resulted in two different uuids. Sending the `content.uuid` to `Page::factory` fixed the issue. 

What do you think about the solution? Do you think this is the right place to implement it?

````php
public function model(): Page
{
	if (isset($this->model)) {
		return $this->model;
	}

	$props = [
		'slug'     => '__new__',
		'template' => $this->template,
		'model'    => $this->template,
		'parent'   => $this->parent instanceof Page ? $this->parent : null
	];

	// make sure that a UUID gets generated
	// and added to content right away
	if (Uuids::enabled() === true) {
		$props['content']['uuid'] = Uuid::generate();
	}

	return $this->model = Page::factory($props);
}
````

https://github.com/getkirby/kirby/blob/5.0.2/src/Panel/PageCreateDialog.php#L256-L264

distantnative:
@afbora Could be a solution as far as I can oversee it.

[...]

PR #7424, opened 2025-07-27
fix: Ensure temporary page models use memory storage and UUIDs

afbora:
## Description

AFAIK content uuid is generated when the page is created (`PageActions::create`). Uuid is not generated when a new page object is created with the `Page::factory()` method. Therefore, I solved the problem by generating and sending the uuid when the page is factory-generated.

Furthermore, following @distantnative suggestion, we ensured that this temporary model is never written to disk using `MemoryStorage`.

### 🐛 Bug fixes

- UUID for new page is the same as the latest page UUID #7405

### 🚨 Breaking changes

AFAIK none.

### For review team

### DIFF
--- a/config/areas/site/dialogs.php
+++ b/config/areas/site/dialogs.php
@@ -287,6 +287,7 @@ return [
 				slug: $request->get('slug'),
 				template: $request->get('template'),
 				title: $request->get('title'),
+				uuid: $request->get('uuid'),
 				viewId: $request->get('view'),
 			);
 
@@ -300,6 +301,7 @@ return [
 				slug: $request->get('slug'),
 				template: $request->get('template'),
 				title: $request->get('title'),
+				uuid: $request->get('uuid'),
 				viewId: $request->get('view'),
 			);
 
--- a/src/Panel/PageCreateDialog.php
+++ b/src/Panel/PageCreateDialog.php
@@ -142,6 +142,13 @@ class PageCreateDialog
 			]);
 		}
 
+		// pass uuid field to the dialog if uuids are enabled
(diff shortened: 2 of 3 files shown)


afbora:
What do you think about moving `Uuid::generate()` process into `Page::factory()` method instead in `PageCreateDialog::model()` for consistency for all calls from anywhere?

bastianallgeier:
I looked at it more closely and I think your solution is great for now. The Page::create method is still a bit messed up after the refactoring. As soon as you pass a content array to the Page constructor, the Page will be switched to the MemoryStorage handler. That's why we can't really add the UUID in the factory method because that would create a breaking change for some cases. 

I think what we need in the long term is a way to pass a custom UUID, but not as part of the content array. We need to separate those concepts some more. I had similar issues when working on the Database Storage plugin. But that needs more thought and discussions. That's why I would keep it simple for this fix and fix it in the way that you did.

afbora:
I agree. I think uuid should be passed as a top-level prop like template or model, not in content as you said..

bastianallgeier:
@afbora do you see a chance to test that last uncovered line? Otherwise, I'm happy with the PR.

bastianallgeier (approved):
I've added an example setup to the sandbox to https://sandbox.test/panel/pages/other+page-creation (Create with UUIDs in custom title and slug) This PR fixes two things: a __new__ folder is no longer created and the UUIDs are consistent between auto-created slug, title and fields. That's great.

@distantnative do you want to give it another try as well?

distantnative:
I think the class could bee refactored a bit but I would do that in our dialog controller refactoring later. 

It's great that these two issues are fixed with this!

PR #7425, opened 2025-07-27
refact!: Panel setting dropdown controllers

distantnative:
## Description

- [x] https://github.com/getkirby/kirby/pull/7422
- [x] https://github.com/getkirby/kirby/pull/7427
- [x] Complete unit tests

### ♻️ Refactored

- Added `Kirby\Panel\Controller\DropdownController` classes to handle setting dropdowns

### Deprecated
- `Kirby\Panel\File::dropdown()`, `Kirby\Panel\Page::dropdown()` and `Kirby\Panel\User::dropdown()` have been deprecated and will be removed in a future Kirby version. Use the respective `Kirby\Panel\Controller\DropdownController` instead.

### Breaking changes
- Removed `Kirby\Panel\Model::isDisabledDropdownOption()`

### For review team


src/Panel/File.php
```diff
@@ -126,69 +126,9 @@ public function dragText(
 	/**
 	 * Provides options for the file dropdown
 	 */
-	public function dropdown(array $options = []): array
+	public function dropdown(): array
```
distantnative, line 130: `Comment:` I did not migrate the feature of passing a `$options` argument to the new controller classes. I could not find any use case for passing extra options.

bastianallgeier: Totally possible that we never used it at all.

bastianallgeier (commented):
I think I would deprecate all `Kirby\Panel\Model::dropdown()` methods with this PR

bastianallgeier (approved):

PR #7436, opened 2025-07-30
feat: New Upload class to handle upload settings

bastianallgeier:
### ✨ Enhancements

- New `Kirby\Panel\Ui\Upload` class to handle upload settings in sections and potentially also in fields

### For review team

### DIFF
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -1,8 +1,8 @@
 <?php
 
 use Kirby\Cms\File;
-use Kirby\Cms\Files;
 use Kirby\Panel\Collector\FilesCollector;
+use Kirby\Panel\Ui\Upload;
 use Kirby\Toolkit\I18n;
 
 return [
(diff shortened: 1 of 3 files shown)


config/sections/files.php
```diff
@@ -185,26 +186,17 @@
...
-			];
+			$settings = new Upload(
+				api: $this->parent->apiUrl(true) . '/files',
+				accept: $this->accept,
+				multiple: true,
+				max: $this->max ? $this->max - $this->total : null,
+				preview: $this->image,
+				sort: $this->sortable === true ? $this->total + 1 : null,
+				template: $this->template,
+			);
```
distantnative, line 169: `Suggestion:` Maybe we could simplify/contain this even better if we add a `Upload::factory(Section $section)` method that takes care of assigning the properties to the constructor.

config/sections/files.php
```diff
@@ -185,26 +186,17 @@
...
-					//       at the same time and upload a file,
-					//       uploaded files have the same sort number
-					'sort'     => $this->sortable === true ? $this->total + 1 : null,
-					'template' => $template
-				]
-			];
+			$settings = new Upload(
+				api: $this->parent->apiUrl(true) . '/files',
+				accept: $this->accept,
+				multiple: true,
```
distantnative: `Suggestion:` Remove to use it as the default of the class

```suggestion
```

config/sections/files.php
bastianallgeier: Since we want to get rid of the sections, I would probably delay this until we have the field class ready.

distantnative (dismissed):

PR #7441, opened 2025-07-30
feat: Drawers UI classes

distantnative:
### ✨ Enhancements

- New `Kirby\Panel\Ui\Drawer`, `Kirby\Panel\Ui\Drawers\TextDrawer` and `Kirby\Panel\Ui\Drawers\FormDrawer` classes

### For review team

### DIFF
--- a/src/Panel/Ui/Drawer.php
+++ b/src/Panel/Ui/Drawer.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Kirby\Panel\Ui;
+
+/**
+ * @package   Kirby Panel
+ * @author    Nico Hoffmann <nico@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @since     6.0.0
+ * @unstable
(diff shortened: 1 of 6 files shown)


src/Panel/Ui/Drawer.php
```diff
@@ -0,0 +1,42 @@
...
+ * @unstable
+ */
+class Drawer extends Component
+{
+	public function __construct(
+		string $component = 'k-drawer',
+		string|null $class = null,
+		public string|null $icon = null,
+		public array|null $options = null,
+		string|null $style = null,
```
bastianallgeier, line 21: `Question` Why did you add the class and style args here again? They would already be included in the $args list and passed on properly otherwise, right?

distantnative: For me the `...$args` list really is only the support for unknown properties. Any known ones from the parent class that we would also allow as parameters for the child class and pass on I would explicitly list still. Really just a style matter, I guess technically it would also work with having them just included in `...$args`.

bastianallgeier (approved):

PR #7442, opened 2025-07-30
feat: Allow UI component in `JsonResponse::from()`

distantnative:
### ✨ Enhancements

- Resolve a Panel UI component object in `JsonResponse::from()` but calling its `::render()` method. This allows us to return UI component instances from a Panel controller's `::load()` method.

### For review team


bastianallgeier (approved):
Uhhhhh, I love this!

PR #7449, opened 2025-07-31
refact: Use list settings for table layout

bastianallgeier:
## Description

This is just a simple enhancement for the `$panel->image()` method to always use the list settings for table layouts. 

### ♻️ Refactored

- Always use the `list` settings for table layouts in `Kirby\Panel\Model::image()`

### 🚨 Breaking changes

None

### For review team

### DIFF
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -94,10 +94,7 @@ return [
 					'extension' => $file->extension(),
 					'filename'  => $file->filename(),
 					'id'        => $file->id(),
-					'image'     => $panel->image(
-						$this->image,
-						$this->layout === 'table' ? 'list' : $this->layout
-					),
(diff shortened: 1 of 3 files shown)


src/Panel/Model.php
```diff
@@ -110,6 +110,11 @@ public function image(
 			return null;
 		}
 
+		// use the list settings for the table layout
+		if ($layout === 'table') {
+			$layout = 'list';
+		}
```
distantnative: `Suggestion:` Wouldn't this belong in the `::imageSource()` method? Although I am not so sure why we even added it (in the sections) cause the `::imageSource()` method seems to fall back to the list dimensions by default anyways.

bastianallgeier: You are completely right. It's not needed at all.

distantnative (approved):

PR #7454, opened 2025-07-31
refact: Field dialogs + drawers as controller

distantnative:
### ♻️ Refactored

- Refactored field dialogs and drawers as dialog controller and drawer controller (using a shared `FieldController` trait)

### 🚨 Breaking changes

- Removed `Kirby\Panel\Field::dialog()`. Use `Kirby\Panel\Controller\Dialog\FieldDialogController` instead.
- Removed `Kirby\Panel\Field::drawer()`. Use `Kirby\Panel\Controller\Drawer\FieldDrawerController` instead.

### For review team


config/areas/site/dialogs.php
```diff
@@ -471,8 +471,8 @@
 	],
 
 	'page.fields' => [
-		...$fields['model'],
 		'pattern' => '(pages/.*?)/fields/(:any)/(:all?)',
```
distantnative, line 477: `Question:` Shouldn't the last `(:all?)` be more specific to only match field routes that start with `dialogs/`? Not new with this PR but I think all of those field routes should maybe be more specific to really just match their use cases?

bastianallgeier: I think you are right. Could have been pure laziness to be honest, because the Filed router would then take care. But it makes sense to separate this more.

distantnative:
@bastianallgeier Added a fix and also an example to the sandbox for testing.

bastianallgeier:
Just tested your sandbox example and the dialog and drawer can both not be submitted. There's a return type issue.

bastianallgeier (approved):

PR #7456, opened 2025-08-01
feat: New Model Item classes

bastianallgeier:
### ✨ Enhancements

- New Model Item Classes to create all the necessary props for a `<k-item>` component for a model, such as text, info, image, link, etc.
  - `Kirby\Panel\Ui\Item\FileItem`
  - `Kirby\Panel\Ui\Item\ModelItem`
  - `Kirby\Panel\Ui\Item\PageItem`
  - `Kirby\Panel\Ui\Item\UserItem`

### ♻️ Refactored

- Use new Model Item classes in the files and pages sections and the users view to clean up the code.

### For review team

### DIFF
--- a/config/areas/users/views.php
+++ b/config/areas/users/views.php
@@ -4,7 +4,7 @@ use Kirby\Cms\App;
 use Kirby\Cms\Find;
 use Kirby\Panel\Collector\UsersCollector;
 use Kirby\Panel\Ui\Buttons\ViewButtons;
-use Kirby\Toolkit\Escape;
+use Kirby\Panel\Ui\Item\UserItem;
 
 return [
 	'users' => [
@@ -42,13 +42,7 @@ return [
 						$users = $collector->models(paginated: true);
 
 						return [
-							'data' => $users->values(fn ($user) => [
-								'id'    => $user->id(),
-								'image' => $user->panel()->image(),
-								'info'  => Escape::html($user->role()->title()),
-								'link'  => $user->panel()->url(true),
-								'text'  => Escape::html($user->username())
-							]),
+							'data'       => $users->values(fn ($user) => (new UserItem(user: $user))->props()),
 							'pagination' => $users->pagination()->toArray()
 						];
 					},
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -6,6 +6,7 @@ use Kirby\Cms\Pages;
 use Kirby\Cms\Site;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Panel\Collector\PagesCollector;
+use Kirby\Panel\Ui\Item\PageItem;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\I18n;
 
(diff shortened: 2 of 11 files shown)


config/areas/users/views.php
```diff
@@ -4,7 +4,7 @@
 use Kirby\Cms\Find;
 use Kirby\Panel\Collector\UsersCollector;
 use Kirby\Panel\Ui\Buttons\ViewButtons;
-use Kirby\Toolkit\Escape;
+use Kirby\Panel\Ui\UserItem;
```
distantnative: `Suggestion:` Since we are starting to create more and more UI classes, we should think whether some subfolders could be useful here as well? E.g. `Layout`?

src/Panel/Ui/FileItem.php
```diff
@@ -0,0 +1,74 @@
...
+			info: $info,
+			layout: $layout,
+			text: $text,
+		);
+	}
+
+	protected function dragText(): string
+	{
+		return $this->panel->dragText(absolute: $this->dragTextIsAbsolute);
+	}
```
distantnative: `Suggestion:` I think we should consider whether such methods fully migrate over from the panel model class to these new classes.

src/Panel/Ui/ModelItem.php
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	protected function info(): string|null
+	{
+		return $this->model->toSafeString($this->info ?? false);
+	}
+
+	protected function image(): array|null
+	{
+		return $this->panel->image($this->image, $this->layout);
```
distantnative: `Suggestion:` Same here - aren't the `image` methods from the panel model classes really just serving the previews in the item sections? If so, would make sense to move them over to this class. I think with a lot of our current refactoring (also dropdown controllers, view controllers) the panel model classes in the end should be relatively lean.

config/areas/users/views.php
bastianallgeier: I hate our Component subfolders in Vue. Not because of the subfolders itself, but because of the inconsistent and sometimes pretty random naming. If we want to go for subfolders here, we would really need a consistent subfolder scheme for our frontend components first (imho)

src/Panel/Ui/FileItem.php
bastianallgeier: Yes, I would be very much in favor of that. But maybe in a separate PR.

config/areas/users/views.php
distantnative: I agree that it's not always consistent. But imagine all our Vue components in just one folder - that would be horrible as well. So I think some structure is needed.

src/Panel/Ui/ModelItem.php
bastianallgeier: Yep, fully on board. I was also thinking about adding new ImageSettings classes (name is WIP) later that could be passed as a value …

```php
$item = new PageItem(
  page: $page,
  image: new PageImageSettings(
    page: $page,
    back: 'black'        
  )
);
```

config/sections/pages.php
```diff
@@ -116,27 +117,13 @@
...
-					'template'    => $page->intendedTemplate()->name(),
-					'text'        => $page->toSafeString($this->text),
-				];
+				$item = (new PageItem(
+					page: $page,
+					image: $this->image,
+					layout: $this->layout,
+					info: $this->info,
+					text: $this->text,
+				))->props();
```
distantnative, line 126: `Suggestion:` If we put the object instantiating into a separate method (`$this->item($page): PageItem`), it would allow for developers who extend the section (or later field) more easily to extend the `PageItem` class and overwrite the method to return their new extended version. Without having to rewrite the code before and after.

config/areas/users/views.php
bastianallgeier: I'm really not sure about it. I often thought about trying to put them all just in one directory or maybe just keep a very few for the different types (fields, inputs, blocks, etc.)

config/sections/pages.php
bastianallgeier: Yes, that's something I'm doing on the "bigger picture branch". Definitely planned as one of the next steps.

config/areas/users/views.php
bastianallgeier: Thinking some more about it, maybe that could already be the rule. If there's more than one type of a component, it needs a subfolder. In this case …

Kirby\Panel\Ui\ModelItem
Kirby\Panel\Ui\ModelItems\FileItem
Kirby\Panel\Ui\ModelItems\PageItem
Kirby\Panel\Ui\ModelItems\UserItem

That would be very consistent with the Buttons and FilePreviews.

distantnative (approved):

PR #7457, opened 2025-08-01
refact: Use model items classes in more places

bastianallgeier:
### ♻️ Refactored

- Use new Model Item classes in `::pickerData()` methods and in the `Kirby\Panel\Controller\Search` controller to clean up all places where such item props are created. 

### For review team

### DIFF
--- a/src/Panel/File.php
+++ b/src/Panel/File.php
@@ -7,6 +7,7 @@ use Kirby\Cms\ModelWithContent;
 use Kirby\Filesystem\Asset;
 use Kirby\Panel\Ui\Buttons\ViewButtons;
 use Kirby\Panel\Ui\FilePreview;
+use Kirby\Panel\Ui\Item\FileItem;
 use Kirby\Toolkit\I18n;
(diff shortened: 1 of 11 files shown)


src/Panel/File.php
```diff
@@ -374,13 +376,19 @@ public function pickerData(array $params = []): array
 			};
 		}
 
-		$params['text'] ??= '{{ file.filename }}';
+		$item = new FileItem(
+			file: $this->model,
+			dragTextIsAbsolute: $absolute,
```
distantnative, line 381:
```suggestion
			dragTextIsAbsolute: $absolute ?? false,
```

`Suggestion:` I often prefer adding those fallback values as actual fallbacks instead of adding the variable on top and having to keep track of the initial value as fallback (mentally).

bastianallgeier: I agree, but in this case the default should be "true" in my opinion and it's more of a special here that it is false.

distantnative (approved):

PR #7463, opened 2025-08-01
refact!: Remove `model` prop from model views

distantnative:
## Description

It would help a lot with the view controllers and UI classes if we can loose those redundant nested data inside the `model` prop.

### 🚨 Breaking changes

- Removed derprevcatd `model` prop from model views. Use the top-level props instead.

### For review team


bastianallgeier (approved):
Yes, I was really looking forward to remove this!!

PR #7465, opened 2025-08-02
feat: `Toolkit\HasI18n` trait

distantnative:
### Enhancements
- New `Kirby\Toolkit\Has18n` trait that adds a `::i18n()` helper method to the class which can be used to translate and/or template i18n strings

### ♻️ Refactored

- Replaced `I18n::` calls in panel controllers with `Kirby\Toolkit\HasI18n::i18n()` helper

### For review team

### DIFF
--- a/src/Panel/Controller/Controller.php
+++ b/src/Panel/Controller/Controller.php
@@ -5,7 +5,7 @@ namespace Kirby\Panel\Controller;
 use Kirby\Cms\App;
 use Kirby\Cms\Site;
 use Kirby\Http\Request;
-use Kirby\Toolkit\I18n;
+use Kirby\Toolkit\HasI18n;
 
 /**
  * @package   Kirby Panel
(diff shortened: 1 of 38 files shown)


src/Panel/Controller/Controller.php
```diff
@@ -30,6 +31,22 @@ public function __construct()
...
+			return null;
+		}
+
+		if ($data === null) {
+			return I18n::translate($key, $key);
+		}
+
+		return I18n::template($key, $data);
+	}
+
```
bastianallgeier: We keep implementing the same method in various places. Components, Fields and now the Controllers. I wonder if we should add a I18n trait to the Toolkit for this

distantnative: Great idea! I have added it.

bastianallgeier (approved):

PR #7466, opened 2025-08-02
refact!: Page create dialog as section dialog

distantnative:
### Todo

- [x] Add back query parameters that are necessary for the deprecated global route
- [x] PR notes
- [x] Unit tests

### Context for reviewers

The current `page.create` dialog route is a bit weird as it is global but actually belongs to a specific view and section. To achieve this with a global route a lot of request query parameters are passed around. This PR rather adds the dialog controller to its dedicated section (future perspective: field).

### Enhancements
- New `create` dialog endpoint for pages sections

### Breaking changes
- `Kirby\Panel\Controller\Dialog\PageCreateDialogController` has been fully refactored and now is initiated with a parent model (page/site) as well as a section (name) 

### Refactored
- New abstract `Kirby\Panel\Controller\Dialog\ModelCreateDialogController` class that takes over some basic logic from the `PageCreateDialogController` class

### Deprecated
- The global `page/create` dialog endpoint has been deprecated. Use the specific dialog endpoint of a pages section instead.
- `Kirby\Panel\Controller\Dialog\PageCreateDialogController::factory()`

### DIFF
--- a/src/Panel/Controller/Dialog/PageCreateDialogController.php
+++ b/src/Panel/Controller/Dialog/PageCreateDialogController.php
@@ -3,17 +3,16 @@
 namespace Kirby\Panel\Controller\Dialog;
 
 use Kirby\Cms\App;
-use Kirby\Cms\File;
 use Kirby\Cms\Find;
+use Kirby\Cms\ModelWithContent;
 use Kirby\Cms\Page;
-use Kirby\Cms\PageBlueprint;
 use Kirby\Cms\PageRules;
+use Kirby\Cms\Section;
 use Kirby\Cms\Site;
 use Kirby\Cms\User;
 use Kirby\Content\MemoryStorage;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Form\Form;
-use Kirby\Panel\Controller\DialogController;
 use Kirby\Panel\Field;
 use Kirby\Panel\Panel;
 use Kirby\Panel\Ui\Dialog;
@@ -23,7 +22,7 @@ use Kirby\Uuid\Uuid;
 use Kirby\Uuid\Uuids;
 
 /**
- * Manages the Panel dialog to create new pages
+ * Controls the Panel dialog to create a new page
  *
  * @package   Kirby Panel
  * @author    Bastian Allgeier <bastian@getkirby.com>
@@ -32,13 +31,8 @@ use Kirby\Uuid\Uuids;
  * @license   https://getkirby.com/license
  * @since     6.0.0
  */
-class PageCreateDialogController extends DialogController
+class PageCreateDialogController extends ModelCreateDialogController
 {
-	public PageBlueprint $blueprint;
-	public Page $model;
-	public Page|Site $parent;
-	public Page|Site|User|File $view;
-
 	public static array $fieldTypes = [
 		'checkboxes',
 		'date',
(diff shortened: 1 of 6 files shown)


src/Panel/Controller/Dialog/PageCreateDialogController.php
```diff
@@ -147,85 +139,33 @@ public function coreFields(): array
 
 		return [
 			...$fields,
-			'parent'   => Field::hidden(),
-			'section'  => Field::hidden(),
 			'template' => Field::hidden(),
-			'view'     => Field::hidden(),
```
distantnative, line 153: `Todo:` We likely need to keep these for now as long as we keep the deprecated global route that relies on those query parameters. If we remove them here, the dialog's submit route would not be able to create the factory, I fear.

src/Panel/Controller/Dialog/PageCreateDialogController.php
```diff
@@ -415,23 +331,12 @@ public function validate(array $input, string $status = 'draft'): bool
...
-			'template' => $this->template,
-			'title'    => $this->title ?? '',
-			'uuid'     => $this->uuid,
-			'view'     => $this->viewId,
+		return [
+			...parent::value(),
+			'slug'     => $this->request->get('slug', ''),
+			'template' => $this->template(),
+			'title'    => $this->request->get('title', ''),
+			'uuid'     => $this->model()->uuid()->toString(),
```
distantnative, line 346: `Todo:` Add back in additional request query parameters used by the deprecated global route for the factory method.

bastianallgeier:
Finally coming back to this. I think we really need to look ahead here. What do we want to achieve and what would be most convenient for other developers? 

If you are a plugin dev and you want to create pages from a plugin, there should be an easy and well documented way to open the dialog. That would be an argument for me to not bind it too closely to the sections. 

We discussed a couple times that we need to get rid of the sections and especially the custom constraints in sections anyway and move to more global page rules for subpages. 

Just thinking out loud here. I imagine to have something like this in the blueprint: 

```yaml
children: 
  max: 20
  templates: 
    - article
    - link
    - video
```

Each section or page field would need to stick to those main constraints and could maybe only add additional one on top of it. 

In this case, the create dialog would also get its options from there and could receive additional parameters via get or maybe path elements.  

It could look like this for example …

```js
// opens with the template picker
this.$panel.dialog('pages/some+page/create');

// opens with a particular template (if it is allowed)
this.$panel.dialog('pages/some+page/create/article');
this.$panel.dialog('pages/some+page/create/video');
```

distantnative:
> If you are a plugin dev and you want to create pages from a plugin, there should be an easy and well documented way to open the dialog. That would be an argument for me to not bind it too closely to the sections.

I see the point, but I think this already doesn't work well without a section even with the global endpoint. Right now, we need a section and then which one is just passed via query parameters - that's odd. I feel it's a lot better that as long as the dialog depends on a section, it should be actually belong to it as this PR achieves.

> We discussed a couple times that we need to get rid of the sections 

Yes, but this will as much apply when we turn them into fields. Then this will be a field dialog, not a section dialog - migrating it will actually be easier IMO after this PR.

> Each section or page field would need to stick to those main constraints and could maybe only add additional one on top of it.

I get your vision, but this feels to me like so many steps into the future. Many breaking changes along the way, I don't know it it makes sense to base our current decisions on such an idea to fully reimagine our blueprint options.

bastianallgeier:
You are right. The argument convinced me that this will help us to move away from sections. It is definitely a good step in between.

distantnative:
@afbora Your testing skills here would be appreciated as well :)

bastianallgeier:
I've ran a first set of tests in our sandbox with the page creation test scenarios and it all looks good. But I would also appreciate another look from @afbora. Is there anything we can do to get the coverage to 100%?

afbora:
@distantnative Your PR works almost perfect. Also I've realized that new dialog open faster than old 🎉 

But I've found small error. Maybe I'm wrong:

> Kirby\Panel\Controller\Dialog\SectionDialogController::__construct(): Argument #\1 ($section) must be of type Kirby\Cms\Section, null given, called in \kirby\src\Panel\Controller\SectionController.php on line 40

````yaml
title: User

sections:
  drafts:
    type: pages
    parent: site
````

To reproduce the issue, use above blueprint in a user blueprint and try to click `Add` button. You'll see the error dialog. I only reproduce the issue with specific section names: `drafts`, `listed` and `unlisted`. When you change the section name, error is gone. Weird!

So do you think this is a bug or I'm missing something?

distantnative:
Sounds like a bug, but interesting that it's based on the section name. Need to look into that.

distantnative:
@afbora I think I fixed it. For me, it wasn't related to the section name but rather `parent: site` not being properly supported.

bastianallgeier (approved):

PR #7470, opened 2025-08-02
fix: Panel view state do not replace with `null`

distantnative:
## Description

I am not sure if this is the right solution, but here is the problem:
If we have a key in `$this->view` that is `null` by default, it will overwrite the defaults/area data. Most times this isn't intended, because `null` here just means no actual value has been provided. This PR would fix this.

However, it also then removes the possibility to actually unset an existing value by setting the key to `null`. Alternatively though, we would have to wrap every `View::render()` call with the filter function.

What do you think?

### DIFF
--- a/src/Panel/State.php
+++ b/src/Panel/State.php
@@ -350,8 +350,8 @@ class State
 
 		$view = array_replace_recursive(
 			$defaults,
-			$this->area?->view() ?? [],
-			$this->view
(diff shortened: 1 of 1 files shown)


distantnative:
Solved it selectively only for the keys in question: https://github.com/getkirby/kirby/pull/7469/commits/a6e2c1bdd9b5dd8747efebf274729311d18daca4

PR #7471, opened 2025-08-02
refact: Language views as controllers

distantnative:
## Description

- [x] https://github.com/getkirby/kirby/pull/7469

### ✨ Enhancements

- New `Kirby\Panel\Ui\Item` class that also builds the foundation for `ModelItem` etc
- New `Kirby\Panel\Ui\LanguageItem` class 

### ♻️ Refactored

- Refactored languages view and language view as controllers

### For review team


src/Panel/Ui/Item/LanguageItem.php
```diff
@@ -0,0 +1,42 @@
...
+			]
+		);
+	}
+
+	public function props(): array
+	{
+		return [
+			...parent::props(),
+			'default'   => $this->language->isDefault(),
+			'deletable' => $this->language->isDeletable(),
```
bastianallgeier, line 38: `Suggestion` we could already make it more consistent here by using the permissions array. But this would mean that we also need to refactor the frontend part a bit more.

distantnative: Let's do that in a separate step. We could add a todo comment.

bastianallgeier: I realized only after my comment that this is not based on ModelItem of course, because the Language object is quite different from other models. I don't think it makes sense at the moment to refactor all of that. Could be done at a later point. Let's leave it as it is.

bastianallgeier (approved):

PR #7474, opened 2025-08-03
refact: System view as controller

distantnative:
## Description

- [x] https://github.com/getkirby/kirby/pull/7472 and merged into `v6/develop`
- [x] https://github.com/getkirby/kirby/pull/7471

### ♻️ Refactored

- Refactored system view as controller

### For review team


src/Panel/Controller/View/SystemViewController.php
```diff
@@ -0,0 +1,201 @@
...
+			return [];
+		}
+
+		// Call `::plugins()` to ensure they inject their exceptions
+		$this->plugins();
+
+		return $this->exceptions;
+	}
+
+	protected function isDebug(): bool
```
bastianallgeier: `Suggestion`: This didn't feel natural when I came across it. I would suggest to use `isDebugging` or `isInDebugMode`.

src/Panel/Controller/View/SystemViewController.php
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+		if ($this->isDebug() === true) {
+			$security[] = [
+				'id'    => 'debug',
+				'icon'  => $this->isLocal() ? 'info' : 'alert',
+				'theme' => $this->isLocal() ? 'info' : 'negative',
+				'text'  => $this->i18n('system.issues.debug'),
+				'link'  => 'https://getkirby.com/security/debug'
+			];
```
bastianallgeier, line 130: It would be really cool if we could use our Item class here as well. But that's also maybe something for a later PR.

distantnative: True, but I would agree not in this PR, because then I think it would be even better to have a `SystemSecurityItems` class not just a `SystemSecurityItem` class - so we can move the whole logic from the method there.

bastianallgeier (approved):

PR #7478, opened 2025-08-04
fix: Merging Responder headers into Response

distantnative:
### 🐛 Bug fixes

- Fixed a case where a subclass of `Kirby\Http\Response` would falsely converted to a `Kirby\Http\Response` object

### ♻️ Refactored

- Support passing a `Kirby\Http\Response` object to `Kirby\Cms\Responder::send()`

### For review team

### DIFF
--- a/src/Cms/Responder.php
+++ b/src/Cms/Responder.php
@@ -4,6 +4,7 @@ namespace Kirby\Cms;
 
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Filesystem\Mime;
+use Kirby\Http\Response as HttpResponse;
 use Kirby\Toolkit\Str;
 use Stringable;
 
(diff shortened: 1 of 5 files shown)


src/Cms/Responder.php
```diff
@@ -337,8 +338,15 @@ public function redirect(
 	/**
 	 * Creates and returns the response object from the config
 	 */
-	public function send(string|null $body = null): Response
+	public function send(HttpResponse|string|null $body = null): HttpResponse
 	{
+		if ($body instanceof HttpResponse) {
+			// inject headers from the responser into the response
```
bastianallgeier:
```suggestion
			// inject headers from the responder into the response
```

bastianallgeier (approved):

PR #7480, opened 2025-08-04
refact!: Model views as controllers

distantnative:
## Description

- [x] Complete unit tests
- [x] Review breaking changes

### ♻️ Refactored

- Refactored site, page, file, user and preview views as controllers

### 🚨 Breaking changes

- `Kirby\Panel\Ui\Button\ViewButtons::view()` now accepts a `Kirby\Panel\Controller\View\ModelViewController` as `$view` argument, instead of a `Kirby\Panel\Model`
- Removed `:: toPrevNextLink()` and deprecated `::content()` method from `Kirby\Panel\Model`, `Kirby\Panel\Site`, `Kirby\Panel\Page`, `Kirby\Panel\File` and `Kirby\Panel\User`.

### Deprecated
- `::breadcrumb()`, `::buttons()`, `::prevNext()`, `::props()`, `::versions()` and  `::view()` methods of `Kirby\Panel\Model`, `Kirby\Panel\Site`, `Kirby\Panel\Page`, `Kirby\Panel\File` and `Kirby\Panel\User`. Use the respective `Kirby\Panel\Controller\View\` classes instead.

### For review team


src/Panel/Controller/View/PageViewController.php
```diff
@@ -0,0 +1,138 @@
...
+
+	public function props(): array
+	{
+		return [
+			...parent::props(),
+			'blueprint' => $this->model->intendedTemplate()->name(),
+		];
+	}
+
+	public function siblings(string $direction): Pages
```
bastianallgeier, line 86: The logic in this entire method almost looks like it could go into a collector class.

src/Panel/Controller/View/FileViewController.php
```diff
@@ -0,0 +1,143 @@
...
+				break;
+			case 'page':
+				/** @var \Kirby\Cms\Page $parent */
+				$breadcrumb = $this->model->parents()->flip()->values(
+					fn ($parent) => [
+						'label' => $parent->title()->toString(),
+						'link'  => $parent->panel()->url(true),
+					]
+				);
+		}
```
bastianallgeier: This part makes me wonder if we should have a UserFileViewController, SiteFileViewController and a PageFileViewController to handle different breadcrumbs and maybe potentially more differences later.

bastianallgeier (commented):
I love how this is drastically trimming down the logic in the Panel Model classes.

src/Panel/Controller/View/PageViewController.php
distantnative: It would work as 

```php
return $this->siblings ??= (new FilesCollector(
	template: $this->model->template(),
	sortBy:   'sort asc filename asc'
))->models();
```

But I am wondering if that really makes this method better than the simple method call?

src/Panel/Controller/View/FileViewController.php
distantnative: Added it

bastianallgeier (approved):
I love the changes. The additional classes feel a bit much at first, but it's so much cleaner and I think that we can avoid quite some unwanted side effects in the future by letting such classes only be responsible for one thing only.

src/Panel/Controller/View/PageViewController.php
bastianallgeier: Let's keep it as it is. Maybe this needs additional collector methods or collectors for all the stuff from the blueprint to really make a difference, but it does not need to happen in this pr.

PR #7484, opened 2025-08-05
feat: `Uuid::toUrl()` + `PageUuid/FileUuid::toPermalink()`

### LINKED ISSUE #7477: allow anchors in kirby link tag for uuid based links
## Description
I am not 100% sure if this is a bug report or feature request. For our use-case it is a blocking issue in using uuid based links and an inconsistency in link handling that i'd love to discuss and fix. We are currently migrating our full site from hardcoded urls to page:// based links. This works fine most of the time and we are super happy to make page renames easier in the future. One thing that is blocking us from generally using this method is, that we are loosing all anchors from  all of our links, which breaks a lot of content on our site.

Before we used:

> (link: /content/#myanchor text: Test)

Now we would like to migrate to:

> (link: page://EWQEPJBB2Xyrrrj6#myanchor text: Test)

But the output on the rendered site will just eat up the anchor. The `link` tag also does not allow an isolated definition of anchors. Therefore i think this is inconsistent and feels more like a subtle bug.

**Expected behavior**  
Even if using uuid based links we should be allowed to use anchors (and maybe even query parameters if we want some javascript magic to happen)

**Proposed patch (Our quick fix for now)**
```
diff --git a/kirby/config/tags.php b/kirby/config/tags.php
index 9f98127..9619c53 100644
--- a/kirby/config/tags.php
+++ b/kirby/config/tags.php
@@ -210,7 +210,11 @@ return [
                                Uuid::is($tag->value, 'page') === true ||
                                Uuid::is($tag->value, 'file') === true
                        ) {
-                               $tag->value = Uuid::for($tag->value)->model()?->url();
+                               $anchor = null;
+                               if (preg_match('/(#[^#]*)$/', $tag->value, $matches)) {
+                                       $anchor = $matches[1];
+                               }
+                               $tag->value = Uuid::for($tag->value)->model()?->url() . $anchor;
                        }
 

```

If this patch is good enough i can submit a PR. Feel free to just use the proposed solution. No credit required.

## To reproduce
Just add a kirby tag with an anchor 

> (link: page://EWQEPJBB2Xyrrrj6#myanchor text: Test)

Render the page and the anchor will be lost.

## Your setup
**Kirby Version**  
Current version from starterkit (8419d9e37360c12029375ef44cb9499323d4098a)

distantnative:
### ✨ Enhancements

- New `Uuid::toUrl()` method that returns the model's absolute URL and includes query and/or fragment that was included in the UUID uri
- Better IDE support for the return type of `Page::uuid()`, `Site::uuid()`, `File::uuid()` and `User::uuid()`

### 🐛 Bug fixes

- Support query and fragments in `link` KirbyTag when using a UUID 
#7477

### Deprecated
- The `::url()` method of `PageUuid` and `FileUUid` has been renamed to `::toPermalink()` and will be removed in a future release.

### For review team

### DIFF
--- a/config/tags.php
+++ b/config/tags.php
@@ -210,7 +210,7 @@ return [
 				Uuid::is($tag->value, 'page') === true ||
 				Uuid::is($tag->value, 'file') === true
 			) {
-				$tag->value = Uuid::for($tag->value)?->modelUrl();
+				$tag->value = Uuid::for($tag->value)?->toUrl();
 			}
 
 			// if url is empty, throw exception or link to the error page
--- a/config/tags.php
+++ b/config/tags.php
@@ -210,7 +210,7 @@ return [
 				Uuid::is($tag->value, 'page') === true ||
 				Uuid::is($tag->value, 'file') === true
 			) {
-				$tag->value = Uuid::for($tag->value)->model()?->url();
+				$tag->value = Uuid::for($tag->value)?->modelUrl();
 			}
 
 			// if url is empty, throw exception or link to the error page
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -514,7 +514,7 @@ class File extends ModelWithContent
 	 */
 	public function permalink(): string|null
 	{
-		return $this->uuid()?->url();
+		return $this->uuid()?->toPermalink();
 	}
 
 	/**
--- a/src/Cms/File.php
+++ b/src/Cms/File.php
@@ -31,6 +31,7 @@ use Kirby\Toolkit\Str;
  * @license   https://getkirby.com/license
  *
  * @use \Kirby\Cms\HasSiblings<\Kirby\Cms\Files>
+ * @method \Kirby\Uuid\FileUuid uuid()
  */
 class File extends ModelWithContent
 {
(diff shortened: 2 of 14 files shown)


distantnative:
Code coverage not very helpful in telling me what's missing:

Screenshot 2025-08-05 at 15 52 43

bastianallgeier:
`Suggestion:` I somehow stumbled over the method name and couldn't really figure out first what it does. Maybe ::toUrl or ::toModelUrl could work better to convey the idea of the method. 

```
page://something#foo -> https://example.com/something#foo
```

distantnative:
@bastianallgeier I didn't go for something with just `url` because `PageUuid::url()` already exists, which returns the permalink. I think the method names aren't always great here. So maybe we could do it like this:

- Change `url()` to `toPermalink()`
- Add a deprecated `url()` method that is just an alias for `toPermalink()`
- Add this new method as `toUrl()` (and not `modelUrl()`)

What do you think?

bastianallgeier:
Sounds great!! I think that the "to" prefix makes a lot of sense for the permalink method as well.

distantnative:
@bastianallgeier I have implemented it

bastianallgeier (approved):

Issue #7417, opened 2025-07-23
Range field: Label width changes while dragging, changing the size of <input> while in use


stairjoke:
## Description
The label displaying the current value of a `field: type: range` changes width and subsequently the `<input>` changes width while in use.

**Expected behavior**  
`<input>` is fixed in width while handle is being dragged.

**Screenshots**  
![Image](https://github.com/user-attachments/assets/ff958ae3-fa19-41f1-8e7a-2fddc139c50c)

## To reproduce
1. Add `range`-field with min-max values that have differing numbers of digits, like "1" and "10".
2. drag slider

## Your setup
**Kirby Version**  
5.0.2

**Console output**  
none

**Your system (please complete the following information)**
 - Device: MacBook
 - OS: macOS 15
 - Browser: Safari
 - Version: 18.5

**Additional context**  
n/A

stairjoke:
💡Maybe using a monospaced font for the label and giving it a min-width in `ch`-units based on the length of the max-value could be a quick fix?

PR #7486, opened 2025-08-05
fix: Stable tooltip width for range input

distantnative:
### 🐛 Bug fixes

- Range input: Tooltip keeps a stable width based on the `max` value
#7417

### For review team


bastianallgeier (approved):
Works great

PR #7489, opened 2025-08-05
Toggle fixes

distantnative:
## Description

- Toggles field: a bit stronger focus color (blue-300)
- Fixed ToggleInput where the styling would act up with Vue 3 (different way of passing down classes). Now the field uses `k-toggle-input` instead `k-input :type="toggle"` similarly to the radio and checkboxes field.

### DIFF
--- a/panel/lab/components/formcontrols/index.vue
+++ b/panel/lab/components/formcontrols/index.vue
@@ -17,7 +17,7 @@
 						/>
 
 						<k-form-controls
-							:has-changes="hasChanges"
+							:has-diff="hasDiff"
 							:is-locked="isLocked"
 							editor="editor@getkirby.com"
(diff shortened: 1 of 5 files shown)


bastianallgeier (approved):
The changes are great. One thing that I stumbled over with the toggles field is the tabindex on the input-validator element. This is not really related to this PR at all, but I was wondering if we should fix it here as well. We basically have two focus steps here. One for the validator and one for the actual inputs. I would love to get rid of the validator focus, but I'm not sure if the validator breaks if we set the tabindex to -1 here.

bastianallgeier:
You know what: Let's keep the tabindex change for the input validator out of this. This is already good on itself. I will also cherrypick the focus color change into develop-minor.

Issue #6582, opened 2024-07-29
Table layout pages section: interactive previews are not updating content


twiro:
### Description:

If a **toggle field** is included as a **column** in a **list of pages in table layout** the preview of the field behaves buggy/irritating:

### Blueprint:

```
example_a:
  label: Table
  type: pages
  layout: table
  columns:
    toggle:
      label: Toggle Field Table Preview
```
### Screenshot of the resulting table:

![Bildschirmfoto 2024-07-29 um 14 26 24](https://github.com/user-attachments/assets/e8bb48f5-7737-40a3-84e7-9a4538a758bd)

### Expected behavior / Problem:

With that preview given _(especially as the UI feels fully functional)_ **I'd expect the toggle value to get saved whenever I use a toggle in the table** – only, it doesn't!

### More Context:

Looks like this preview was originally [implementend for the structure field](https://github.com/getkirby/kirby/pull/2141) and works there. _(Haven't tested that actually)_

## B) Example B

In the forum I ran across the idea to use `type: text` for the preview of a toggle field in table layout, but that doesn't result in a satisfying experience either:

### Blueprint:

```
example_b:
  label: Table
  type: pages
  layout: table
  columns:
    toggle:
      label: Toggle Field Table Preview
      type: text
```
### Screenshot of the resulting table:

<img width="743" alt="Bildschirmfoto 2024-07-29 um 14 27 01" src="https://github.com/user-attachments/assets/dc50a37d-902a-46e1-9fde-5b7c86475932">

### Expected behavior / Problem:

If changing the preview of a toggle field to text in a table column is actually possible, I'd expect to see the toggle text values I defined in my blueprint (**`Ja`**, **`Nein`**) or at least the default text values for the given language (**`an`**, **`aus`**) – but all i get is **`true`** and **`false`**.

## C) Example C

Thinking about the text values of the toggle field preview a little more neither **`true`** and **`false`** nor the default text values of the given language (**`an`** and **`aus`**) do feel really convincing here  – if the values can be customized in the blueprint, the only really satisfying solution for the field preview would be to have it customazible there too. If not automatically, then at least by redefining the values in the columns-settings of the table:

### Blueprint:

```
example_c:
  label: Table
  type: pages
  layout: table
  columns:
    toggle:
      label: Toggle Field Table Preview
      type: toggle
      text:
        - "Nein"
        - "Ja"
```
### Screenshot of the resulting table:

[...]

PR #7490, opened 2025-08-05
fix: Disable field previews in sections

distantnative:
## Description

We already expected to receive `fields` in `ModelsSection.vue`. However, it was never passed. This PR starts passing them and setting on all fields `disabled: true`. This also solves editable field previews (e.g. toggle) that would not update the content nevertheless. The toggle preview now also shows texts from the blueprint.

It makes the `layout` mixin not less complex, rather more, but I don't think we can solve this with this PR. And I think it would be good to fix the bug.

### ✨ Enhancements

- Toggle field preview shows defined texts from blueprint in pages and files sections

### 🐛 Bug fixes

- Editable field previews are disabled in  pages and files sections (instead of intractable but non-functional)
#6582

### For review team

### DIFF
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -190,6 +190,7 @@ return [
 				'batch'    => $this->batch,
 				'columns'  => $this->columnsWithTypes(),
 				'empty'    => $this->empty,
+				'fields'   => (object)$this->fields(),
 				'headline' => $this->headline,
 				'help'     => $this->help,
 				'layout'   => $this->layout,
--- a/config/sections/mixins/layout.php
+++ b/config/sections/mixins/layout.php
@@ -2,6 +2,7 @@
 
 use Kirby\Cms\ModelWithContent;
 use Kirby\Form\Form;
+use Kirby\Toolkit\A;
 use Kirby\Toolkit\I18n;
 use Kirby\Toolkit\Str;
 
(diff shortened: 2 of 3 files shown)


bastianallgeier:
So far the `fields` prop was only passed on the client side with the fields that we already have for the model. While I like this in general, I'm a bit worried that we will run into more performance issues with the table layout again. A simple first patch could be to disable the toggle field in JS in the ModelsView. It's not super elegant, but could have less side effects.

distantnative:
We could send it only for `layout: table` but IMO it's the same code we already run for columns, so it shouldn't impact performance worse than before.

bastianallgeier:
But only as long as rawvalues isn't activated, right?

bastianallgeier:
I've created a quick proof of concept how actually activating the toggle could look like. It's technically not super hard and would be an amazing feature. The only real issue is a conceptual one. Do we update the latest version or the changes version? 

In both cases, the state can easily get out of sync between the table and the form. 

https://github.com/getkirby/kirby/pull/7553

bastianallgeier:
I've also created an alternative fix to disable the fields in the frontend: https://github.com/getkirby/kirby/pull/7554 It doesn't solve the texts from blueprints part, but has less potential performance impact

Issue #7491, opened 2025-08-06
Panel: unsaved changes UI pops up when opening structure field side panel


rasteiner:
## Description
When a structure field contains specific fields (link, color, probably others) which already contain a value, merely opening its side pane triggers the "unsaved changes" UI. Even when no changes were made. 

The changes aren't actually persisted on the backend (no `_changes` folder), and refreshing the page makes the buttons go away. 

**Expected behavior**  
The "Discard" & "Save" buttons should only appear if the data has been changed. 

**Screenshots**  
From the demo online:

https://github.com/user-attachments/assets/fe88c322-11e5-4164-a312-fc22a2928f9c

## To reproduce

1. Go to https://getkirby.com/try and start a demo instance
2. Log into the panel and navigate to a shop product
3. Under "Available colors" click anything, but then make sure the color field in the side pane receives focus at some point
4. close the side pane and notice the "Save" button

The link field in the same situation doesn't need focus, but it needs to point to a Page (not just any url). 

## Your setup
**Kirby Version**  5.0.3 (also 5.0.2, haven't tried other)

**Console output**  
```js
// BEFORE Save Button appears
> console.log(JSON.stringify(panel.content.version('latest').colors, null, 2))
< [
  {
    "color": "#005fa9",
    "name": "Blue"
  }
]

// AFTER Save Button appears
> console.log(JSON.stringify(panel.content.version('latest').colors, null, 2))
< [
  {
    "color": "#005fa9",
    "name": "Blue"
  }
]

> console.log(JSON.stringify(panel.content.diff(), null, 2))
< {
  "colors": [
    {
      "_id": "7235159d-afa9-4953-97fa-32fa8ba35622",
      "color": "#005fa9",
      "name": "Blue"
    }
  ]
}
```

**Your system (please complete the following information)**
 - Device: Desktop PC
 - OS: Windows 11
 - Browser: Chrome
 - Version: latest

**Additional context**  
https://discord.com/channels/525634039965679616/525641819854471168/1402552743029243934

PR #7492, opened 2025-08-06
fix: Stucture field `_id` handling

distantnative:
## Description

This seems to be a feasible quick fix for #7491: Adding `_id` in the structure field will trigger the form controls every time when included in the emitted value as it will always be different to what's coming from the backend. This PR strips `_id` again before emitting the structure field value.

In the larger picture, I think our handling of `_id` here is a bit weird: I was first thinking "but StructureObject does already have an ID" - but then realized that in the structure field handling itself the `Structure`\`StructureObject` classes aren't involved at all. I think we should change this for a future refactoring of the field:
- Parse the content value as yaml
- Pass it to the structure classes
- Retrieve the data (including its UUID) from the classes and pass it to the frontend
- Only for newly added rows, add a UUID in the frontend

### 🐛 Bug fixes

- Structure field does not trigger the form controls for no reason 
#7491

### For review team

### DIFF
--- a/config/fields/structure.php
+++ b/config/fields/structure.php
@@ -198,9 +198,6 @@ return [
 				)
 				->toStoredValues();
 
-			// remove frontend helper id
-			unset($row['_id']);
-
 			$data[] = $row;
 		}
 
--- a/panel/src/components/Forms/Field/StructureField.vue
+++ b/panel/src/components/Forms/Field/StructureField.vue
@@ -517,7 +517,12 @@ export default {
 		 * @param {array} values
 		 */
 		save(values = this.items) {
-			this.$emit("input", values);
+			this.$emit(
+				"input",
+				// strip _id from values
+				// eslint-disable-next-line no-unused-vars
+				values.map(({ _id, ...value }) => value)
+			);
 		},
 		/**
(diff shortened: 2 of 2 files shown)


bastianallgeier:
This won't work unfortunately. As soon as we remove the ID on save, the rows are no longer connected properly. You can see this if you edit a row in the drawer. With your PR, the values are no longer properly updated.

distantnative:
@bastianallgeier What would be the alternative? We cannot add randomly generated data to the value and emit that. This will always kick in our form controls as the values get changed every time.

bastianallgeier:
TBH, I don't really think there's an alternative to actually finally adding IDs to structures. The way we handle this right now is just not good and I don't see how we could fix it at all.

distantnative:
I agree, but this will take a longer while - we do need a patch fix for this now though. Do you know why emitting the `_id` is crucial and now it stops working? I do not quite understand why it is needed to emit it.

bastianallgeier:
Yes, the problem is that emitting will immeditately also update the value and then the rows no longer have the id to be properly connected.

distantnative:
Closing this as we need a different solution.

PR #7496, opened 2025-08-07
fix: Only redirect to valid URLs after Panel login

### LINKED ISSUE #6682: Panel not opened after login when page doesn’t exist
## Description
This might be a bit of an edge case, but when logging in to the Panel using a route that points to a specific page (eg. `http://localhost:8000/panel/pages/example`) which **doesn’t** exist in the current installation (but was suggested by the browser from another project for example), it will show a modal saying this page doesn’t exist but will **stay on the login screen**. 

A simple reload or signing in again will do, but it’s just an extra step that can be spared.

Screenshot 2024-09-17 at 4 22 10 PM

**Expected behavior**  

Kirby should redirect either to the Panel’s “Page not found” page or simply to `/panel/site`.

## To reproduce

1. Start a local development server and navigate to a non-existent Panel route (eg. `/panel/pages/idontexist`)
2. Kirby redirects to login: enter your credentials
3. Hit enter/click “Sign in”
4. Modal should appear

## Your setup
**Kirby Version**  
v4.4.0

**Console output**  
``` 
GET http://localhost:8000/panel/pages/schule+digitalisierung+luinet-einfuehrung 404 (Not Found)
gu @ index.min.js:2
request @ index.min.js:2
get @ index.min.js:2
open @ index.min.js:2
load @ index.min.js:2
open @ index.min.js:2
reload @ index.min.js:2
login @ index.min.js:2
await in login
submit @ index.min.js:2
_n @ vue.min.js:11
r @ vue.min.js:11
i._wrapper @ vue.min.js:11
index.min.js:2 pu: The request to http://localhost:8000/panel/pages/schule+digitalisierung+luinet-einfuehrung failed
    at ku (http://localhost:8000/media/panel/9b4bffa460105781f82b1d463bde8200/js/index.min.js:2:380455)
    at async gu (http://localhost:8000/media/panel/9b4bffa460105781f82b1d463bde8200/js/index.min.js:2:380038)
    at async Object.get (http://localhost:8000/media/panel/9b4bffa460105781f82b1d463bde8200/js/index.min.js:2:403857)
    at async Object.open (http://localhost:8000/media/panel/9b4bffa460105781f82b1d463bde8200/js/index.min.js:2:404014)
    at async Object.load (http://localhost:8000/media/panel/9b4bffa460105781f82b1d463bde8200/js/index.min.js:2:384911)
```

 - Device: MacBook Pro
 - OS: MacOS 15.0
 - Browser: Any (Tested in Chrome, Safari)

distantnative:
### How to test
- Log out of the Panel
- Try access https://sandbox.test/panel/pages/foo+bar
- Being redirected to login
- Log in
- Being redirected to site view with this PR; before this PR error dialog

### 🐛 Bug fixes

- Panel will not redirect anymore to non-existing Panel views if these were attempted to access prior to login
#6682

### For review team

### DIFF
--- a/src/Panel/Home.php
+++ b/src/Panel/Home.php
@@ -118,13 +118,21 @@ class Home
 					return false;
 				}
 
-				// if auth is not required the redirect is allowed
-				if ($auth === false) {
-					return true;
+				// check the firewall, if auth is required
+				if (
+					$auth !== false &&
+					Panel::hasAccess($user, $areaId) === false
+				) {
+					return false;
 				}
 
-				// check the firewall
-				return Panel::hasAccess($user, $areaId);
+				// check if the route yields a valid result
+				$result = $route->action()->call(
+					$route,
+					...$route->arguments()
+				);
+
+				return $result !== false;
 			});
 		} catch (Throwable) {
 			return false;
--- a/tests/Panel/HomeTest.php
+++ b/tests/Panel/HomeTest.php
@@ -179,6 +179,11 @@ class HomeTest extends TestCase
 			],
 			'users' => [
 				['email' => 'test@getkirby.com', 'role' => 'admin']
+			],
+			'options' => [
+				'api' => [
+					'allowImpersonation' => true
+				]
 			]
 		]);
 
@@ -220,6 +225,11 @@ class HomeTest extends TestCase
 			],
 			'users' => [
 				['email' => 'test@getkirby.com', 'role' => 'editor']
+			],
+			'options' => [
+				'api' => [
+					'allowImpersonation' => true
+				]
 			]
 		]);
 
@@ -229,6 +239,28 @@ class HomeTest extends TestCase
 		$this->assertFalse(Home::hasAccess($user, 'pages/test'));
 		$this->assertTrue(Home::hasAccess($user, 'users/test@getkirby.com'));
 		$this->assertTrue(Home::hasAccess($user, 'account'));
+	}
+
+	public function testHasAccessWithInvalidPath(): void
+	{
+		$this->app = $this->app->clone([
+			'site' => [
+				'children' => [
+					['slug' => 'test']
+				]
+			],
+			'users' => [
+				['email' => 'test@getkirby.com', 'role' => 'admin']
+			],
+			'options' => [
+				'api' => [
+					'allowImpersonation' => true
+				]
+			]
+		]);
+
+		$user = $this->app->impersonate('test@getkirby.com');
+		$this->assertFalse(Home::hasAccess($user, 'pages/foo+bar'));
 	}
 
 	public function testHasValidDomain(): void


bastianallgeier (commented):
Since this is calling the route already, couldn't this cause some side effects if the route would actually execute something? It could also have quite some performance hit for more complex routes. Checking for false is probably also not enough when routes could return error response objects instead. I'm not sure if we could solve it differently, but I'm a bit worried about the implications.

distantnative:
@bastianallgeier I get the concerns, let's try to dissect it a little:
- This is only checking view routes. Yes, technically they could also execute code on load but I would say the scenario where code is executed which is harmful when executed by this check vs. any other visit is low. Though not impossible.
- it surely has a performance implication. But only when this check is actually run. Which is only after login. So not on the regular Panel navigation. We could consider whether we want to differentiate between remembered paths which should be checked and maybe blueprint-defined paths (where we might skip the check and if it results in errors, the developer can easily adjust the blueprint). Ideally we would check both, but if this is a concern.
- We should indeed consider some other return values and whether we then consider the route as valid redirect target or not.

distantnative:
We could also skip calling the route for common routes we know will work by this point (e.g. site).

lukasbestle (commented):
I agree with Basti that it's risky to call route code in a method called `hasAccess`. Even if our views don't have side effects, we cannot guarantee it in the future and also for plugin views, so this is a risk that could become a security vulnerability later.

Trying to understand why the old code didn't already cover this: There must be a fallback route for non-existing views that returns `false`, right? Couldn't we add a flag to it that can be easily detected without calling the route action?

bastianallgeier:
I like @lukasbestle' idea. Maybe we could make sure to always throw a NotFoundException in our error routes and catch that?

distantnative:
The case here is that we want to specifically handle the remembered redirect after the login.

The suggestion would mean we just always redirect to site if a Panel route is not found without an error message. Not just after login.

bastianallgeier:
Forget my comment. We would still need to run the route action for what I suggested. I mix things up in my head here.

lukasbestle:
Trying to approach this from a different perspective:

Do we even want to silently redirect to the site view on a "page not found" error?

If I'm already logged in and try to visit a page view of a non-existing page, I get a proper view with the error message. From there I can navigate to whatever other view I want, so I'm not lost.

Could we do the same with the redirect after login? So instead of displaying the error in a dialog on the login view, we would perform the redirect to the provided URL and display the `k-error-view` there as normal.

bastianallgeier:
@lukasbestle good point. The error dialog is definitely weird here. I had that in different scenarios where it really bothered me. So this is something that we should definitely fix - independently from the login redirect issue. If a redirect leads to an error view, the redirect should go through and not lead to the error dialog.

distantnative:
I think the login redirect is special here as running into the error is not like usually intentionally trying a panel route that doesn't exist, but just trying to access a page before logging in that then got renamed. I don't think redirecting the user to a big error view instead of an error dialog is the solution here.

lukasbestle:
I really think it is. Let's say the following happens:

1. User A shares a direct Panel page link with a colleague B.
2. Before user B can access the link, user A renames the page.
3. User B clicks the link.
  a) They are already logged in and are presented with the error view. They will know what happened and can ask their colleague if the link is broken.
  b) They are not already logged in. With this PR, they will land on the site view and be confused.

What I'm asking is: Why should case 3b) be different from case 3a) just because the user was not already logged in?

bastianallgeier:
I think we have a couple problems here that add up: 

- The error dialogs for redirects are bad. We should show the error view instead.
- But our error pages in the panel are also pretty bad. If I end up on a page route for a missing page, we could massively improve the experience by having a custom error view that says something like … "This page seems to have been moved or no longer exists. Go to the parent page, Go the site overview, Search for the page, etc."

distantnative:
Closing this as it seems to be a dead end based on the feedback

PR #7499, opened 2025-08-07
fix: Abort previous Panel view requests

### LINKED ISSUE #6536: Slower navigation requests still go through even if already navigated elsewhere

distantnative:
## Description

I tried to catch the abort request error right in the view code, but only could manage to silence it in the error handler.

### 🐛 Bug fixes

- Subsequent Panel view request cancel out ongoing (slow) previous view request to ensure navigation always lands at the latest requested view
#6536

### For review team

### DIFF
--- a/panel/src/panel/view.js
+++ b/panel/src/panel/view.js
@@ -24,6 +24,21 @@ export default (panel) => {
 		...parent,
 
 		/**
+		 * Load a view from the server and
+		 * cancel any previous request
+		 *
+		 * @param {String|URL} url
+		 * @param {Object|Function} options
+		 * @returns {Object} Returns the current state
+		 */
+		async load(url, options = {}) {
+			// cancel any previous request
+			this.abortController?.abort();
+
+			return parent.load.call(this, url, options);
+		},
+
+		/**
 		 * Setting the active view state
 		 * will also change the document title
 		 * and the browser URL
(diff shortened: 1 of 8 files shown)


panel/src/panel/view.js
```diff
@@ -23,6 +23,21 @@ export default (panel) => {
...
+		 * Load a view from the server and
+		 * cancel any previous request
+		 *
+		 * @param {String|URL} url
+		 * @param {Object|Function} options
+		 * @returns {Object} Returns the current state
+		 */
+		async load(url, options = {}) {
+			// cancel any previous request
+			this.abortController?.abort();
```
bastianallgeier, line 36: `Question:` Can you help me understand why this line is needed when the load method in feature.js already aborts the controller?

distantnative: But the load method in `feature.js` isn't aborting the controller. It's just replacing it with a new abort controller, but leaving the old untouched. I didn't implement the abort call in `feature.js` as I was wondering if in other features (e.g. dialogs, drawers) it could lead to problems aborting an ongoing request. Whereas with view navigation, I was more confident that aborting the request is the right thing to do.

bastianallgeier: Oh, I completely missed that. Makes perfect sense now.

bastianallgeier (approved):
Just double-checked with a super slow loading system view, by injecting a sleep call and it works great.

PR #7504, opened 2025-08-08
refact: Layout mixin refactoring

bastianallgeier:
## Description

The table layout refactoring is turning into a larger project and this PR is extracting one of the quick wins that should be rather easy to review. 

### ♻️ Refactored

- Refactor table layout mixin

### For review team

### DIFF
--- a/config/sections/mixins/layout.php
+++ b/config/sections/mixins/layout.php
@@ -145,30 +145,25 @@ return [
 			// TODO: remove once Form classes have been refactored
 			// @codeCoverageIgnoreStart
 			if ($this->rawvalues === true) {
-				foreach ($this->columns as $columnName => $column) {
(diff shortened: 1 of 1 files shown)


config/sections/mixins/layout.php
```diff
@@ -145,30 +145,25 @@
...
-				return $item;
+				$row = [
+					...$model->content()->toArray(),
+					...$item,
+				];
+			} else {
+				$row = [
+					...Form::for($model)->toFormValues(),
+					...$item,
+				];
```
distantnative, line 156: `Question:` Reading the code, I would think we have a breaking change here where the assignment from column id to content field name no longer works (before it was either the key of the column array or the `id` option inside the array). Do you know if this has any implications?

PR #7505, opened 2025-08-08
Add collapse/expand all for Field-Blocks in Blocks Menu

dennisbaum:
## Description

This PR adds two new buttons to the `blocks` menu of the panel to allow editors to expand or collapse all blocks with `preview: fields` at once.
The goal is to improve usability for content-heavy pages by reducing repetitive clicks and making it faster to manage large sets of blocks. The feature should be fully backwards-compatible and does not affect other block types, so the buttons wont show in the menu, if not at least one field-block is in the fieldsets. No need for config or field params atm.

https://github.com/user-attachments/assets/ef1887e6-d2da-4ca9-92fd-9f8e3bb018cd

### 🎉 Features

* Added "Expand all" and "Collapse all" buttons to the `blocks` menu,  for blocks with `preview: fields`.
* Buttons use already existing Translations and Kirby Icons
* Buttons wont show, if not at least 1 block with preview: fields is in the fieldsets
* Buttons are disabled, if blocks is empty
 
### ✨ Enhancements

* Improves workflow for editors managing large sets of block content.

## Docs

* Include a screenshot or short GIF to demonstrate the feature.

### For review team

* [ ] Add lab and/or sandbox examples (wherever helpful)
* [ ] Add changes & docs to release notes draft in Notion

### DIFF
--- a/panel/src/components/Forms/Blocks/Block.vue
+++ b/panel/src/components/Forms/Blocks/Block.vue
@@ -158,9 +158,6 @@ export default {
 
 			return "k-block-type-default";
 		},
-		isCollapsable() {
-			return this.fieldset.preview === "fields";
-		},
 		isDisabled() {
 			return this.disabled === true || this.fieldset.disabled === true;
 		},
@@ -250,14 +247,10 @@ export default {
 			this.$panel.drawer.close(this.id);
 		},
 		collapse() {
-			if (this.isCollapsable) {
-				this.$refs.editor?.collapse();
-			}
+			this.$refs.editor?.collapse?.();
 		},
 		expand() {
-			if (this.isCollapsable) {
-				this.$refs.editor?.expand();
-			}
+			this.$refs.editor?.expand?.();
 		},
 		focus() {
 			if (typeof this.$refs.editor?.focus === "function") {
--- a/panel/src/components/Forms/Blocks/Block.vue
+++ b/panel/src/components/Forms/Blocks/Block.vue
@@ -266,7 +266,6 @@ export default {
 				this.$refs.container?.focus();
 			}
 		},
-
 		goTo(block) {
 			if (block) {
 				block.$refs.container?.focus();
--- a/panel/src/components/Forms/Blocks/Block.vue
+++ b/panel/src/components/Forms/Blocks/Block.vue
@@ -265,6 +265,12 @@ export default {
 				block.open(null, true);
 			}
 		},
+		isCollapsible() {
+			if (this.$refs.editor) {
+				return typeof this.$refs.editor.collapse === "function";
+			}
+			return false;
+		},
 		isSplitable() {
 			if (this.isFull === true) {
 				return false;
--- a/panel/src/components/Forms/Blocks/Block.vue
+++ b/panel/src/components/Forms/Blocks/Block.vue
@@ -158,6 +158,9 @@ export default {
 
 			return "k-block-type-default";
 		},
+		isCollapsable() {
+			return this.fieldset.preview === "fields";
+		},
 		isDisabled() {
 			return this.disabled === true || this.fieldset.disabled === true;
 		},
@@ -246,6 +249,16 @@ export default {
 		close() {
 			this.$panel.drawer.close(this.id);
 		},
+		collapse() {
+			if (this.isCollapsable) {
+				this.$refs.editor?.collapse();
+			}
+		},
+		expand() {
+			if (this.isCollapsable) {
+				this.$refs.editor?.expand();
+			}
+		},
 		focus() {
 			if (typeof this.$refs.editor?.focus === "function") {
 				this.$refs.editor.focus();
@@ -253,6 +266,7 @@ export default {
 				this.$refs.container?.focus();
 			}
 		},
+
 		goTo(block) {
 			if (block) {
 				block.$refs.container?.focus();
--- a/panel/src/components/Forms/Field/BlocksField.vue
+++ b/panel/src/components/Forms/Field/BlocksField.vue
@@ -75,7 +75,7 @@ export default {
 			return this.max && this.value.length >= this.max;
 		},
 		isCollapsable() {
-			var hasPreviewFields = false;
+			let hasPreviewFields = false;
 			this.value.forEach((item) => {
 				if (this.fieldsets[item.type]?.preview === "fields") {
 					hasPreviewFields = true;
--- a/panel/src/components/Forms/Field/BlocksField.vue
+++ b/panel/src/components/Forms/Field/BlocksField.vue
@@ -37,6 +37,7 @@
 				@close="opened = $event"
 				@open="opened = $event"
 				@input="$emit('input', $event)"
+				@collapsible-change="onCollapsibleChange"
 			/>
 		</k-input-validator>
 
@@ -61,7 +62,8 @@ export default {
 	inheritAttrs: false,
 	data() {
 		return {
-			opened: []
+			opened: [],
+			isCollapsible: false
 		};
 	},
 	computed: {
@@ -74,12 +76,6 @@ export default {
 		isFull() {
 			return this.max && this.value.length >= this.max;
 		},
-		isCollapsable() {
-			return this.value.some(
-				(item) => this.fieldsets[item.type]?.preview === "fields"
-			);
-		},
-
 		options() {
 			return [
 				{
@@ -94,7 +90,7 @@ export default {
 					icon: "download",
 					text: this.$t("paste")
 				},
-				...(this.isCollapsable
+				...(this.isCollapsible
 					? [
 							"-",
 							{
@@ -124,6 +120,9 @@ export default {
 	methods: {
 		focus() {
 			this.$refs.blocks.focus();
+		},
+		onCollapsibleChange(value) {
+			this.isCollapsible = value;
 		}
 	}
 };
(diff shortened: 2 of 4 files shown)


panel/src/components/Forms/Blocks/Block.vue
```diff
@@ -158,6 +158,9 @@ export default {
 
 			return "k-block-type-default";
 		},
+		isCollapsable() {
+			return this.fieldset.preview === "fields";
```
distantnative: `Suggestion:` Instead of this check, I would make the two methods dependent on the presence of a collapse/expand function:

```js
collapse() {
  this.$refs.editor?.collapse?.();
}
```

panel/src/components/Forms/Field/BlocksField.vue
```diff
@@ -88,6 +93,23 @@ export default {
...
+								text: this.$t("collapse.all")
+							},
+							{
+								click: () => this.$refs.blocks.expandAll(),
+								disabled: this.isEmpty,
+								icon: "expand",
+								text: this.$t("expand.all")
+							}
+						]
+					: []),
```
distantnative, line 109: `Discussion:` I think it'll be a bit confusing that these show up even if only one is collapse/expendable.

dennisbaum: Do you think the label `expand` vs. `expand.all` should depend on the amount of field-blocks? one vs. many?
I don’t think that’s really necessary. There ist also `delete.all` and one could even be all ;)

dennisbaum: But because it's possible to have also non fields-blocks added _only_, it could indeed be a bit confusing to have an action where nothing happens. So here you're definitly right.
In that case would you prefer the buttons to be `disabled` or `hidden`, if only  non-fields-block are set, but field-blocks are in the fieldsets?
How can i detect, if there is at least one fields-block in the current blocks object? 
i wasn't sure about the data-structure and just chose the fieldsets-object for the sake of simplicity.

`this.$refs.blocks.blocks` doesn't have a property preview.
Do i have to go with `this.$refs.blocks.$refs` ??

Any suggestion would be nice here! Tnx

bastianallgeier: The isCollapsable method is already checking for each added block if its fieldset has the preview: fields option. The state is definitely correct. If we only want to show the buttons if there's more than one block, we'd just need to add a counter here. I would hide the buttons entirely as you already do.

dennisbaum: I just did an update to `isCollapsable` to compare the current blocks with the available fieldsets, and now i enable collapsibility only if at least one block with preview=fields is set.
I can't detect a better solution, while a block in `this.value` does not have the  property `preview` but `type`, right?
As i said earlier: In my opinion, `collapse.all` is similar to `copy.all` and `delete.all` - both actions work also, if only one block is set. furthermore editors getting used to that feature, would also use that with only one block i guess.
But if you think that would be better - with the new logic in `isCollapsable`, a counter would also be possible...

https://github.com/user-attachments/assets/2200f79c-b1af-42d3-a5bf-4afce82940d4

bastianallgeier: TBH, I'm fine if it also shows up for just one block. As long as it is not confusing by not doing anything, I'm happy :)

dennisbaum: If no counter is needed, i would optimize isCollapsable to:

```ts
isCollapsable() {
    return this.value.some(
        (item) => this.fieldsets[item.type]?.preview === "fields"
    );
},
```

dennisbaum: what do you think @distantnative ?

distantnative: I would be fine with them showing up even if there is just one collapsable block, I guess. I don't like so much that we hardcode it to preview fields. I would rather tie it wether the block has a collapse method or not.

bastianallgeier: It would indeed be cool if the block components could determine this for themselves. But going through the blocks refs always feels a bit nastier than using the state in my opinion. So far, we really only have the ones with the field preview that could be collapsable at all, so I would say that we could also safely delay this until a later point when other block types become collapsable as well - if there's ever a use case for it.

distantnative: @bastianallgeier I think the use case is not so much our core blocks but custom blocks from plugins.

dennisbaum: i worked out a solution to dynamically detect if the block's editor has a function `collapse`, but tbh it feels a bit ugly compared to the previous solution and i am not sure if this would be the best way - but i can understand why it would be nice to enable collapsibility not only by `preview: fields` if there will be other types of collapsable blocks.

So as far as i understood the structure, i used `this.$refs.blocks.$refs` and compare the `IDs` from this.value, to check for the editors collapse function.

`this.$refs?.blocks?.$refs` is not available on first use of computed, but it is on mounted. unfortunately it wont reactivate computed again, so i added a const `canCheckCollapsibility` and changes it on mounted. i don't know why the computed does not take action, after `this.$refs?.blocks?.$refs` is available . you guys?

```ts
isCollapsableByCollapseFunction() {
  if (!this.canCheckCollapsibility) {
    return false;
  }

  return this.value.some((block) => {
    const blocksRef = this.$refs?.blocks?.$refs;
    const editor = blocksRef[`block-${block.id}`][0]?.$refs?.editor;
    return editor && typeof editor?.collapse === "function";
  });
 }
 
mounted() {
  this.canCheckCollapsibility = true;
},
```
 
What do you think about that solution?
I really don't like that too much. But maybe there is another concept i did not get yet?

Interesting: When i add a FieldsBlock with a required field, the editor is `undefined` until that field is filled, but collapse by mouse works already 🤔

bastianallgeier: Also not super excited about this version tbh.

distantnative: If you both think we should limit it to fields preview, ok. I'd like it to be not as limited. The long ref chains look off-putting, but the solution there would be to actually break it down and have each component do part of the check so it's not one long chain but each component handling just their own ref.

dennisbaum: i like the idea not being limited to `preview: fields`, but i don't like the above implementation that much.
I can have a further look, but the blocks-system is quite complex to me and maybe i would need some kind of advice, but i know you have by far other things to do. so if we limit the collapsibility to `preview: false` for now, the feature is nearly ready to ship imo.

I was wondering if it theoretically would be possible to add a property isCollapsable to `this.value`. 
Would be really easy to consume in the menu: 
 
image

dennisbaum:
I did a general refactoring and enable `isCollapsible` in the `BlocksField` only if at least 1 block with a collapse function exists in the current blocks-list.

- Because of the async state of `this.$refs` i added an async method `checkCollapsibility` in `Blocks` and watch the state `isCollapsible` to emit `collapsible-change` on change. `BlocksField` is listening and change the menu options
- I also added a deep check for the `collapse` method in `Block` by the existing `ref` method in `Blocks`.
- In `Block` I did the guard for `editor` in `isCollapsible` similar to the one in `isSplitting` (TBD)

Please have a look and tell me, what you think about that!

panel/src/components/Forms/Field/BlocksField.vue
distantnative: I tried an alternative solution here: https://github.com/getkirby/kirby/commit/4f7c6dd6e4dd1abb126d6138babf68167fd18f0a

I tried to avoid introducing extra state (via data and events). When we move the `options` of `BlocksField` from a computed prop to a method, we can also move a `isCollapsibleExpandable` method in place that checks the method of `Blocks` etc. - and down on the lowest level at `Block` we simply check if the block has a `collapse` and an `expand` method - if so, we can consider it as capable for this functionality.

Have a look and let me know what you think.

dennisbaum: i think moving the `options` of `BlocksField` from computed to a method and create the options `onClick` first is a really good idea to get rid of all the async probs, events and extra watcher. keeps the diff small and elegant. I really like it! Thank you @distantnative

dennisbaum: With this nice basement, i would like to go one step further and regarding to the disabled state of some buttons, i.e. `delete.all` if `this.empty`, i also would like to update the disabled state of the expand/collapse-buttons. In that case, we can show the buttons, if at least one block with collapse/expand is in the blocks, but set their disable state exactly to the collapsed-state of the block. So there would also not be an action, where nothing happens. What do you think about that improvement? 

Empty:   
image

Fake:   
image

distantnative: @dennisbaum I like the idea: https://github.com/getkirby/kirby/commit/13cc9bb6aeb88bd9f124193a3ab03e46c380dda4

dennisbaum: Top!! Are we done then? @distantnative

distantnative:
I have created a new PR with the refined version https://github.com/getkirby/kirby/pull/7555

PR #7508, opened 2025-08-08
fix: Pass missing input prop in complex fields 

### LINKED ISSUE #7506: [v5.1] Validation asterisk missing for certain fields
Blocks, pages, files, users, entries and others - all those that do not pass a `input` prop to `k-field`

distantnative:
## Description

Fixes unreleased regression on `develop-minor`: As `k-label` now hides the asterisk when no `input` is specified, it became apparent that some fields are not passing any `input` to `k-label`.

This PR passes `input` for those missing fields. To do that it also adds `id` attribute support to `k-input-validator`.

### For review team

### DIFF
--- a/panel/src/components/Forms/Field/BlocksField.vue
+++ b/panel/src/components/Forms/Field/BlocksField.vue
@@ -1,6 +1,7 @@
 <template>
 	<k-field
 		v-bind="$props"
+		:input="id"
 		:class="['k-blocks-field', $attrs.class]"
 		:style="$attrs.style"
 	>
@@ -28,7 +29,7 @@
 		</template>
 
 		<k-input-validator
-			v-bind="{ min, max, required }"
+			v-bind="{ id, min, max, required }"
 			:value="JSON.stringify(value)"
 		>
 			<k-blocks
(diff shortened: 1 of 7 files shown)


bastianallgeier:
To avoid a breaking change here, we could also only ever hide the asterisk if input is set to false. This could be changed later in v6.

distantnative:
@bastianallgeier sounds good to me

bastianallgeier:
@distantnative I've created a new PR for it: https://github.com/getkirby/kirby/pull/7524

distantnative:
@bastianallgeier I have approved the other PR. But I don't think it should fully replace this one to be honest. The complex fields that do not pass an `input` value so far still remain not fully functional. Clicking the label doesn't focus the relevant input part as with this PR, I would say.

bastianallgeier:
TBH, I don't really know if it is technically correct to pass any id for any element to the for attribute or if it has to be an input element. Feels at least weird to me to set the focus on a table.

distantnative:
Is it setting it on the table and not the button? It should follow the same logic our input validator element already handles navigation/focus setting.

bastianallgeier:
It's only setting it on the table for the object field, the rest seems fine.

distantnative:
Oh true, I think removing the `id` from the table and just having it on the add button could be an option then.

distantnative:
@bastianallgeier I have removed the ID from the object field table.

bastianallgeier (approved):

PR #7516, opened 2025-08-09
feat: `selected` items & radio choice

distantnative:
## Description

This PR makes the batch selection feature of `k-items` more useful for other scenarios:
- Which items are selected can be determined by the parent component via a new prop
- New event that receives the updated selected list to not repeat the logic of add or remove

Different states can be inspected: https://sandbox.test/panel/lab/components/item/1_list
And selecting can be tested: https://sandbox.test/panel/lab/components/items/1_list

With these changes in place, we could use this selecting items feature e.g. for the models picker dialogs.

### ✨ Enhancements

- `k-item` has a new `selected` prop to control it's selection checked state and `seelctmode` prop to define whether to select single or multiple items
- `k-collection`/`k-items` can receive a list of `selected` items to control selection status from parent

### 🐛 Bug fixes

- `k-item` only hides default options button when in selecting mode, but not custom used options slot
- `k-item` is selectable by default unless explicitly defined otherwise

### Breaking changes
- `k-collection` and `k-items`: `@select` events now passes an array of selected IDs, not a single item object

### For review team

### DIFF
--- a/panel/src/components/Collection/Item.vue
+++ b/panel/src/components/Collection/Item.vue
@@ -53,16 +53,12 @@
 				v-bind="button"
 			/>
 
-			<label
-				v-if="selecting !== false"
-				class="k-item-options-checkbox"
-				@click.stop
-			>
+			<label v-if="selecting" class="k-item-options-checkbox" @click.stop>
 				<input
 					ref="selector"
 					:checked="selected"
 					:disabled="!selectable"
-					:type="selecting === 'single' ? 'radio' : 'checkbox'"
+					:type="selectmode === 'single' ? 'radio' : 'checkbox'"
 					@change="$emit('select', $event)"
 				/>
 			</label>
@@ -70,7 +66,7 @@
 			<!-- Options -->
 			<slot name="options">
 				<k-options-dropdown
-					v-if="options && selecting === false"
+					v-if="options && !selecting"
 					:options="options"
 					class="k-item-options-dropdown"
 					@option="onOption"
@@ -118,15 +114,6 @@ export default {
 			type: [Array, Function, String]
 		},
 		/**
-		 * If `true`, the item will be selectable via a checkbox
-		 * @since 5.0.0
-		 * @values "single", "multiple", false
-		 */
-		selecting: {
-			type: [Boolean, String],
-			default: false
-		},
-		/**
 		 * If `false`, the select checkbox will be disabled
 		 * @since 5.0.0
 		 */
@@ -135,10 +122,24 @@ export default {
 			default: true
 		},
 		/**
+		 * If `true`, the item's checkbox or radio button
+		 * will be shown to select the item
+		 * @since 5.0.0
+		 */
+		selecting: Boolean,
+		/**
 		 * Whether the item is selected
 		 * @since 6.0.0
 		 */
 		selected: Boolean,
+		/**
+		 * @since 6.0.0
+		 * @values "single", "multiple"
+		 */
+		selectmode: {
+			type: String,
+			default: "multiple"
+		},
 		/**
 		 * If `true`, the sort handle will be shown on hover
 		 */
--- a/panel/src/components/Collection/Item.vue
+++ b/panel/src/components/Collection/Item.vue
@@ -53,19 +53,24 @@
 				v-bind="button"
 			/>
 
-			<label v-if="selecting" class="k-item-options-checkbox" @click.stop>
+			<label
+				v-if="selecting !== false"
+				class="k-item-options-checkbox"
+				@click.stop
+			>
 				<input
 					ref="selector"
-					type="checkbox"
+					:checked="selected"
 					:disabled="!selectable"
+					:type="selecting === 'single' ? 'radio' : 'checkbox'"
 					@change="$emit('select', $event)"
 				/>
 			</label>
 
 			<!-- Options -->
-			<slot v-else name="options">
+			<slot name="options">
 				<k-options-dropdown
-					v-if="options"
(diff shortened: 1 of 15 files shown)


panel/src/components/Collection/Item.vue
```diff
@@ -53,19 +53,24 @@
...
 					@change="$emit('select', $event)"
 				/>
 			</label>
 
 			<!-- Options -->
-			<slot v-else name="options">
+			<slot name="options">
 				<k-options-dropdown
-					v-if="options"
+					v-if="options && selecting === false"
```
distantnative: `Explanation:` I think it makes sense to still allow showing a custom options slot even when selecting. E.g. when using this in the pages picker dialog, we would want to show the checkbox alongside a navigate into the page button next to it at the same time. 

Alternatively, we could keep it the old way but add another slot for UI that should still be shown when in selecting mode.

panel/src/components/Collection/Item.vue
```diff
@@ -115,13 +120,25 @@ export default {
 		/**
 		 * If `true`, the item will be selectable via a checkbox
 		 * @since 5.0.0
+		 * @values "single", "multiple", false
 		 */
-		selecting: Boolean,
+		selecting: {
+			type: [Boolean, String],
+			default: false
+		},
```
distantnative: `Explanation:` I wanted to avoid to add another prop, but if mixing this feels too weird, we could add a `selectingMultiple: true` prop or so.

panel/src/components/Collection/Items.vue
```diff
@@ -38,7 +38,8 @@
 					:layout="layout"
 					:link="link ? item.link : false"
 					:selecting="selecting"
-					:selectable="item.selectable"
+					:selectable="item.selectable !== false"
```
distantnative, line 40: `Explanation:` I think items should be selectable by default and only disabled, when explicitly marked as not selectable.

panel/src/components/Collection/Items.vue
```diff
@@ -59,6 +60,29 @@
 
 <script>
 import { layout } from "@/mixins/props.js";
+
+export function findSelectedIndex(items, item) {
+	return items.findIndex((selected) =>
+		Object.hasOwn(selected, "_id")
+			? selected._id === item._id
+			: selected.id === item.id
+	);
```
distantnative: `Explanation:` To avoid the extra `batchEditingIdentifier` and having something similar for other use cases of `k-items`, I thought we could just check either `_id` or `id`.

Alternatively, one would have to pass the key as additional parameter to `findSelectedIndex` and then within `batchEditing.js` we would pass `batchEditingIdentifier`, in `k-items` we would use `id` in the `onSelected` method (and since batch editing keeps relying on just the `onSelect` method that would not conflict) .

panel/src/components/Forms/Field/StructureField.vue
```diff
@@ -415,6 +412,8 @@ export default {
 				return false;
 			}
 
+			this.stopSelecting();
```
distantnative: `Explanation:` Would make sense to me not having the selecting UI being active while overlayed by the drawer.

panel/src/components/Collection/Items.vue
```diff
@@ -156,19 +193,30 @@ export default {
...
-			this.$emit("select", event, item, itemIndex);
+		onSelect(item, itemIndex) {
+			this.$emit("select", item, itemIndex);
+			this.onSelected(item);
+		},
+		onSelected(item) {
+			this.$emit(
+				"selected",
+				updateSelection(this.selected, item, this.selecting)
+			);
```
distantnative: `Discussion:` While `onSelect` moves around the whole item object, I am still wondering if it would be better to have `onSelected` only emit the IDs of selected items.

panel/src/components/Forms/Field/StructureField.vue
bastianallgeier: I think we should add this as a fix to develop-minor even develop-patch

distantnative: I can backport that, yes.

distantnative: Added it to `develop-minor`. Doesn't exist yet on `develop-patch`.

panel/src/components/Collection/Item.vue
bastianallgeier: I think I would prefer a separate prop for it. Maybe we could call it `selectmode`?

panel/src/components/Collection/Items.vue
bastianallgeier: From a consistency standpoint it feels a lot cleaner to emit a single item and multiple items in those events. Getting the ids from the items array is still very easy, so I would vote to keep it as it is.

bastianallgeier (commented):
This is a really nice update 💛

panel/src/components/Collection/Items.vue
distantnative: I think the downside of that is that it then would make more sense to also pass full item objects to `:selected` - which in some cases like the picker fields aren't always available and passing just IDs would be a lot simpler. But let's leave it like this.

bastianallgeier: Oh, that's indeed something I didn't think about. "selected" should really just be a list of ids and then you are of course right that the matching event should again also be just ids. Quite challenging to get all the details right :)

distantnative: Haha great, now I settled for your opinion and you switched to mine 😆

bastianallgeier (approved):

PR #7519, opened 2025-08-10
feat: Nested dialogs via history module

distantnative:
## Changelog 

Bringing the nested drawer support also to dialogs. 
I think the new `closed` listener could be great for nested dialogs if a dialog opens another dialog to define in that `closed` listener logic that needs to run when the 2nd dialog is closed and the first dialog is resumed. The `close` listener does not always work for that as the state will still be of the 2nd dialog - so if it's any logic that needs to interact with the state of the 1st/original dialog, this needs to happen after the 2nd dialog has been closed.

### ✨ Enhancements

- History support for `panel.dialog` for nested dialogs
- New `closed` listener for Panel dialogs and drawers that gets called after the dialog has been closed

### 🐛 Bug fixes
- Fixed dialog button layout when there's only one button

### For review team


bastianallgeier:
I've added more lab examples to test the nested dialogs and realized that we need to fix the backdrop click for the dialogs. 

1. Go to https://sandbox.test/panel/lab/components/dialogs/1_dialog
2. Scroll down to the nested example
3. Open the first dialog and submit it to get to the second
4. Click on the backdrop to close the dialog

Instead of arriving back at the first dialog, both dialogs are closed. When you try to open them again with the open button, nothing happens.

distantnative:
@bastianallgeier is it the same for Drawers already?

bastianallgeier:
Could be that it happens there as well. Let me setup a test example in the lab as well.

bastianallgeier:
I've added an example and drawers don't have the same problem. Clicking on the backdrop correctly closes the currently open drawer, while keeping open parents.

bastianallgeier:
Ahhh, I found the missing piece. The dialog overlay needs the nested state in Panel.vue. From my perspective, everything works really well now. We could consider adding a breadcrumb to nested dialogs at a later point as well. But this is already so cool and will make quite a few missing features possible!!

bastianallgeier (approved):

PR #7520, opened 2025-08-10
feat: Dialog + drawer dropzone

distantnative:
### ✨ Enhancements

- Added a dropzone to dialogs and drawers that becomes active when a `@drop` listener is added to the component

### For review team

### DIFF
--- a/panel/lab/components/dropzones/index.vue
+++ b/panel/lab/components/dropzones/index.vue
@@ -1,6 +1,6 @@
 <template>
 	<k-lab-examples>
-		<k-lab-example label="dropzone">
+		<k-lab-example label="Dropzone">
 			<k-dropzone @drop="onDrop">
 				<k-box
 					theme="empty"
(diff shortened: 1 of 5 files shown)


bastianallgeier:
I like this, but I also wonder if we actually need this that often that it should be built into the dialogs and drawers at all instead of defining custom components like we do for the uploader?

distantnative:
@bastianallgeier If we do it like in the uploader, you cannot make the full dialog the dropzone. Only some of the connect (e.g. just the collection), but not all of it incl. the buttons etc. - which I though think makes for a good UX in many cases when you have a larger area to just drop the file. But if we decide to merge this feature, I think we could refactor the uploader to use this as well.

bastianallgeier (approved):
I just tested the PR properly for the first time and the effect of dragging files into the dialogs and drawers is really nice. i'm happy with this. We might need to check during the alpha/beta if the additional wrapper element causes any CSS issues, but otherwise we can merge this.

distantnative:
@bastianallgeier I fixed the outline (had to add a negative offset) and directly made use of it in the upload dialog.

PR #7521, opened 2025-08-10
feat: `k-item` skeleton theme

distantnative:
## Description

The idea is to have an `k-item` theme that can be used while an item is still being loaded as placeholder.

I don't think this is necessarily the best design, but I could not come up with something better and am counting a little bit on @bastianallgeier to maybe have a better idea how it should look like.

Lab: https://sandbox.test/panel/lab/components/item/4_skeletons

### ✨ Enhancements

-`sekeleton` theme for `k-item`

### For review team

### DIFF
--- a/panel/lab/components/item/4_skeletons/index.vue
+++ b/panel/lab/components/item/4_skeletons/index.vue
@@ -1,0 +1,13 @@
+<template>
+	<k-lab-examples>
+		<k-lab-example label="List item">
+			<k-item :image="{ icon: 'loader' }" layout="list" theme="skeleton" />
+		</k-lab-example>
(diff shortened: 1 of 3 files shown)


bastianallgeier:
I like the skeletons a lot. I replaced the fallback dash for missing titles with a &nbsp; The dash does not really add any value and the skeletons look better when they are completely empty. We need the &nbsp; though to keep the min height of the title element.

distantnative:
@bastianallgeier Fien to me :) We should just remove the one CSS rule then. Opacity not needed for the space.

bastianallgeier (approved):

PR #7522, opened 2025-08-10
refact!: Models fields use models picker dialogs

distantnative:
## Description

- [ ] https://github.com/getkirby/kirby/pull/7723
- [ ] https://github.com/getkirby/kirby/pull/7726
- [ ] https://github.com/getkirby/kirby/pull/7727

### ♻️ Refactored

- Pages, files and users field use new picker dialog components
- Pages, files and users field dynamically fetch item data from new `items` field API endpoint

### ☠️ Deprecated

- `k-models-dialog`, `k-pages-dialog`, `k-files-dialog` and `k-users-dialog` have been deprecated. Use `k-models-picker-dialog`, `k-pages-picker-dialog`, `k-files-picker-dialog`, `k-users-picker-dialog` instead.
- `Kirby\Cms\Picker`, `Kirby\Cms\PagePicker`, `Kirby\Cms\FilePicker` and `Kirby\Cms\UserPicker` have been deprecated. Use `Kirby\Panel\Controller\Dialog\ModelsPickerDialogController`, `Kirby\Panel\Controller\Dialog\PagesPickerDialogController`, `Kirby\Panel\Controller\Dialog\FilesPickerDialogController` and `Kirby\Panel\Controller\Dialog\UsersPickerDialogController` instead.

### 🚨 Breaking changes

- Pages, files and users field: 
  - Backend sends only UUIDs/IDs to these fields and the fields only send UUIDs/IDs back (instead of full item objects with display data etc.)
  - Same communication happens between the field and picker dialog
  - Removed `pagepicker`, `filepicker` and `userpicker` field mixins

### For review team


config/fields/files.php
```diff
@@ -48,9 +49,6 @@
 		'parent' => function () {
 			return $this->parentModel->apiUrl(true);
 		},
-		'query' => function () {
-			return $this->query ?? $this->parentModel::CLASS_ALIAS . '.files';
-		},
```
distantnative, line 53: `Explanation:` Query default is handled by new `FilePickerDialogController` class.

config/fields/files.php
```diff
@@ -60,51 +58,48 @@
...
-						'limit'  => $field->limit(),
-						'page'   => $this->requestQuery('page'),
-						'query'  => $field->query(),
-						'search' => $this->requestQuery('search'),
-						'text'   => $field->text()
-					]);
-				}
+				'pattern' => 'items',
+				'method'  => 'GET',
+				'action'  => fn () => $this->field()->itemsFromRequest()
```
distantnative, line 83: I didn't found a better way as `api` routes cannot be defined by a mixin as it seems.

config/fields/mixins/picker.php
```diff
@@ -98,20 +107,27 @@
...
 				}
 
 				if ($id !== null && ($model = $this->toModel($id))) {
-					$items[] = $this->toItem($model);
+					$ids[] = $model->id();
 				}
 			}
 
-			return $items;
+			return $ids;
```
distantnative, line 133: `Review:` Does it make sense that we always use IDs here and not UUIDs? Since we use `->toModel()->{$store)` in the `save` callback, it should work consistently to always just communicate with the IDs. But wondering if conceptually we should also use `$uuid?? $id` here.

distantnative:
@afbora Could you help me to test this in the Panel (I know a few unit tests are still not passing)? Basically figure out where the files, pages and users field have issues with this PR? I assume we will still have a few regressions, so it would be great to find these together.

afbora:
@distantnative I've found only one issue (two sub issues, I think both are related) for files field so far:

Files field doesn't work correctly:
1. There is a selected file but file details not visible and loading spinner visible
2. When you try to upload new one from field, getting: `Kirby\Form\Field::{closure}(): Argument #1 ($model) must be of type Kirby\Cms\ModelWithContent, null given`

distantnative:
@afbora I think I found a fix for that problem

config/fields/files.php
distantnative: We can do this better in the next step when turning fields into classes.

distantnative:
Closing in favor of https://github.com/getkirby/kirby/pull/7731 which is built on the class-based picker fields

PR #7510, opened 2025-08-08
fix: `::pickerData()` for models

distantnative:
## Description

We missed some props in the `ModelItem` refactoring PR that we need in `::pickerData()` - especially for the User object.

This PR fixes this unreleased regression on `develop-minor`.

- [x] https://github.com/getkirby/kirby/pull/7509

### For review team

- [x] Add lab and/or sandbox examples (wherever helpful)
- [x] Add changes & docs to release notes draft in Notion

PR #7527, opened 2025-08-13
feat: Models picker dialogs classes

distantnative:
## Description

https://github.com/getkirby/kirby/pull/7522 for seeing the larger context how they should be used.

### Merge first

- [x] https://github.com/getkirby/kirby/pull/7510

### 🎉 Features

- New `Kirby\Panel\Controller\Dialog\ModelsPickerDialogController` classes (`Kirby\Panel\Controller\Dialog\PagesPickerDialogController`, `Kirby\Panel\Controller\Dialog\FilesPickerDialogController`, `Kirby\Panel\Controller\Dialog\UsersPickerDialogController`)

### For review team

### DIFF
--- a/src/Panel/Controller/Dialog/FilesPickerDialogController.php
+++ b/src/Panel/Controller/Dialog/FilesPickerDialogController.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace Kirby\Panel\Controller\Dialog;
+
+use Kirby\Cms\File;
+use Kirby\Cms\ModelWithContent;
+use Kirby\Panel\Collector\FilesCollector;
+
+/**
+ * Controls the Panel dialog for selecting files
+ *
+ * @package   Kirby Panel
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @since     6.0.0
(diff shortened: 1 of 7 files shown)


bastianallgeier (commented):
I really like the refactoring. The props in the constructor make me wonder though if this is something that we should do for other controllers as well, where we often go directly through request params instead.

distantnative:
@bastianallgeier I can't quite follow your thought - what is it what you are suggesting? 😅

bastianallgeier:
@distantnative 😆 sorry for the confusion. Let's take the PageChangeTitleDialogController. It maybe not be the best example, but the first one I found where we use the request in the load method to get the 'preselect' prop. We normally mostly just pass the Model to the constructor and search for the model in the factory method of the controller. Additional options are then coming from the request object. Would it be better to set those options properties in the constructor as well and then hand them over from the request in the factory method instead?

bastianallgeier (approved):

bastianallgeier:
But this discussion is basically not something that we have to decide for this PR. It's more a general conceptual decision that we can still leave open or change later.

distantnative:
Now understood you. Yes, maybe in the long term they should already be passed to the constructor - I think we need to test out what works best for the most cases once we have done a bit more refactoring (e.g. https://github.com/getkirby/kirby/pull/7466).

PR #7530, opened 2025-08-13
refact!: Make Symfony default YAML handler

distantnative:
### ☠️ Deprecated

- The Spyc YAML handler has been deprecated and will be removed in a future release.

### 🚨 Breaking changes

- Changed the default YAML handler to Symfony YAML which sometimes enforces a stricter syntax than our previous Spyc handler. You can switch back to Spyc with the config option `'yaml.handler' => 'spyc'`

### For review team

### DIFF
--- a/src/Data/Yaml.php
+++ b/src/Data/Yaml.php
@@ -22,8 +22,8 @@ class Yaml extends Handler
 	public static function encode($data): string
 	{
 		return match (static::handler()) {
-			'symfony' => YamlSymfony::encode($data),
-			default   => YamlSpyc::encode($data),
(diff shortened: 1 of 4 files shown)


bastianallgeier:
This one could indeed have the biggest noticeable negative impact for users. But on the other hand, we have to go there at some point. Maybe we could already implement a cli command to convert all yaml files from spyc to symfony. (Read with spyc -> write with symfony) That should be pretty straight forward.

distantnative:
I agree. This could really create some issues for some sites - but as you said, we have to do it eventually. CLI command sounds good.

bastianallgeier (approved):

PR #7531, opened 2025-08-13
fix: add "audio/wav" mime type

pReya:
## Description
`.wav` files come with either one of two mime types: Either `audio/wav` or `audio/x-wav`. The first one (which seems to be more popular) was missing.

### DIFF
--- a/src/Filesystem/Mime.php
+++ b/src/Filesystem/Mime.php
@@ -98,7 +98,7 @@ class Mime
 		'tgz'   => ['application/x-tar', 'application/x-gzip-compressed'],
 		'tif'   => 'image/tiff',
 		'tiff'  => 'image/tiff',
-		'wav'   => ['audio/wav', 'audio/x-wav'],
+		'wav'   => ['audio/wav', 'audio/x-wav', 'audio/vnd.wave', 'audio/wave'],
 		'wbxml' => 'application/wbxml',
 		'webm'  => 'video/webm',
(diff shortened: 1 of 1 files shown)


src/Filesystem/Mime.php
```diff
@@ -98,7 +98,7 @@ class Mime
 		'tgz'   => ['application/x-tar', 'application/x-gzip-compressed'],
 		'tif'   => 'image/tiff',
 		'tiff'  => 'image/tiff',
-		'wav'   => 'audio/x-wav',
+		'wav'   => ['audio/x-wav', 'audio/wav'],
```
bastianallgeier:
```suggestion
		'wav'   => ['audio/wav', 'audio/x-wav'],
```

bastianallgeier: When audio/wav is the more popular one - which totally makes sense - I would set it as first option in the array.

bastianallgeier (dismissed):

afbora:
Maybe we should support more mime types for wav files according the followings:

- https://mimetype.io/audio/vnd.wav
- https://en.wikipedia.org/wiki/WAV

Especially we can add `audio/vnd.wav` mime type.

afbora:
Great! Could you fix the unit tests, please?

distantnative:
I added it manually to `develop-minor` with unit tests. Thanks!

PR #7540, opened 2025-08-15
feat: Section dialogs + drawers

distantnative:
### ✨ Enhancements

- Sections can define their own dialogs and drawers on the backend now, as fields have already been able to

## Docs

```php
[
    'props' => [...],
    'dialogs' => function () {
        return [
            'my-dialog' => [
                'load' => function () {},
                'submit' => function () {},
            ]
        ];
    },
    'drawers' => function () {
        return [
            'my-drawer' => [
                'load' => function () {},
                'submit' => function () {},
            ]
        ];
    }
]
```

### For review team

  - https://github.com/getkirby/sandbox/pull/20

### DIFF
--- a/config/areas/account/dialogs.php
+++ b/config/areas/account/dialogs.php
@@ -33,6 +33,10 @@ return [
 		...$dialogs['user.fields'],
 		'pattern' => '(account)/fields/(:any)/(:all?)',
 	],
+	'account.sections' => [
+		...$dialogs['user.sections'],
+		'pattern' => '(account)/sections/(:any)/(:all?)',
+	],
 	'account.file.changeName' => [
 		...$dialogs['user.file.changeName'],
 		'pattern' => '(account)/files/(:any)/changeName',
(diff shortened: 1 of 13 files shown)


lukasbestle:
I know there are uses for this as we discussed, but looking at the amount of added code I wonder if we really *need* it before sections are going to be deprecated. Especially since this targets 6.0.0.

distantnative:
@lukasbestle I think we do if we want to make progress with the page create dialog, upload dialogs... I don't think it's that much added code tbh, most of it is tests. And the benefits outweigh this in my opinion. If the migration away from sections happens quickly (before release of v6), we can throw it out again before ever releasing. But if it takes longer, we don't stall other dependencies because that project takes longer. I think that's ehe way for us going forward: trying to see where we can split up different projects into smaller steps and how we can make them less dependent on each other (even if it means to add a bit more code for a version or two). Instead of building long, single-thread dependency chains.

lukasbestle:
Right, that makes sense.

bastianallgeier (approved):

PR #7541, opened 2025-08-15
fix: Page create ignore all core fields

distantnative:
## Description

This PR uses all fields that the `::coreFields()` method defines and ignores them instead of a separately defined list of fields.

### For review team

### DIFF
--- a/src/Panel/Controller/Dialog/PageCreateDialogController.php
+++ b/src/Panel/Controller/Dialog/PageCreateDialogController.php
@@ -159,13 +159,14 @@ class PageCreateDialogController extends DialogController
 	 */
 	public function customFields(): array
 	{
-		$custom    = [];
-		$ignore    = ['title', 'slug', 'parent', 'template', 'uuid'];
-		$blueprint = $this->blueprint();
-		$fields    = $blueprint->fields();
+		$custom = [];
+		$fields = $this->blueprint()->fields();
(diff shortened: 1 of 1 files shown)


afbora:
Even though it is not a correct usage, I can bypass your control mechanism with the blueprint below:

````yaml
title: Simple Page

create:
  title: "{{ site.time }}"
  slug: "{{ site.time }}"
  fields:
    - slug
    - text

fields:
  slug:
    type: text
  text:
    type: text
````

Because if I pass title and slug to create prop, coreFields array will not contain title and slug indexes.

https://github.com/getkirby/kirby/blob/v6/fix/panel-create-dialog-ignore-all-core-fields/src/Panel/Controller/Dialog/PageCreateDialogController.php#L122-L139

Although my example is not a realistic usage, I expect this control mechanism to be consistent.

**Comment by distantnative** [comment]
@afbora That's a good point I didn't think about. We probably need to hardcode these two then as they are special:

```php
$ignore = [...array_keys($this->coreFields()), 'title', 'slug'];
```

What do you think?

afbora:
This solution will work 👍 

Here's another idea: What if we stored the field names in a constant like `RESERVED_FIELDS`?

distantnative:
Then we would have to alter both the `::coreFields()` method as well as the constant when we extend this class. I think it makes sense to base it on the `::coreFields()` class.

bastianallgeier (approved):

PR #7544, opened 2025-08-15
refact!: Split `Uuid::for()` and `Uuid::from()`

distantnative:
## Description

- [x] https://github.com/getkirby/kirby/pull/7543
- [x] https://github.com/getkirby/kirby/pull/7545

The motivation here is to break up some of the magic monster methods into separate, dedicated methods.

### ♻️ Refactored

- New `Kirby\Uuid\Uuid::from(string $uuid)` method for creating an Uuid object from a UUID string. `Kirby\Uuid\Uuid::for()` remains to create a Uuid object for a model object.

### 🚨 Breaking changes

- `Kirby\Uuid\Uuid::for()`  cannot be called any longer with a string. Use `Kirby\Uuid\Uuid::from(string $uuid)` or `Kirby\Uuid\Permalink::from(string $permalink)` instead.

### For review team

### DIFF
--- a/config/tags.php
+++ b/config/tags.php
@@ -206,8 +206,8 @@ return [
 
 			// if value is a UUID, resolve to page/file model
 			// and use the URL as value
-			if (Uuid::is($tag->value, ['page', 'file']) === true) {
-				$tag->value = Uuid::for($tag->value)?->toUrl();
+			if ($uuid = Uuid::from($tag->value, ['page', 'file'])) {
+				$tag->value = $uuid->toUrl();
 			}
 
 			// if url is empty, throw exception or link to the error page
--- a/config/tags.php
+++ b/config/tags.php
@@ -206,8 +206,8 @@ return [
 
 			// if value is a UUID, resolve to page/file model
 			// and use the URL as value
-			if ($uuid = Uuid::from($tag->value, ['page', 'file'])) {
-				$tag->value = $uuid->toUrl();
+			if (Uuid::is($tag->value, ['page', 'file']) === true) {
+				$tag->value = Uuid::from($tag->value)?->toUrl();
 			}
 
 			// if url is empty, throw exception or link to the error page
--- a/src/Cms/HasFiles.php
+++ b/src/Cms/HasFiles.php
@@ -72,8 +72,8 @@ trait HasFiles
 		}
 
 		// find by global UUID
-		if (Uuid::is($filename, 'file') === true) {
-			return Uuid::for($filename, $this->$in())->model();
+		if ($uuid = Uuid::from($filename, 'file', $this->$in())) {
+			return $uuid->model();
 		}
 
 		if (str_contains($filename, '/') === true) {
--- a/src/Cms/HasFiles.php
+++ b/src/Cms/HasFiles.php
@@ -72,8 +72,8 @@ trait HasFiles
 		}
 
 		// find by global UUID
-		if ($uuid = Uuid::from($filename, 'file', $this->$in())) {
-			return $uuid->model();
+		if (Uuid::is($filename, 'file') === true) {
+			return Uuid::from($filename, context: $this->$in())->model();
 		}
 
 		if (str_contains($filename, '/') === true) {
--- a/src/Text/KirbyTag.php
+++ b/src/Text/KirbyTag.php
@@ -116,28 +116,31 @@ class KirbyTag
 	 */
 	public function file(string $path): File|null
 	{
-		if ($parent = $this->parent()) {
-			// check first for UUID
-			if (method_exists($parent, 'files') === true) {
+		$parent = $this->parent();
+
+		// check first for UUID
+		if (Uuid::is($path, 'file') === true) {
+			if ($parent && method_exists($parent, 'files') === true) {
 				$context = $parent->files();
 			}
 
-			if ($uuid = Uuid::from($path, 'file', $context ?? null)) {
-				return $uuid->model();
-			}
-
-			if (method_exists($parent, 'file') === true) {
-				return $parent->file($path);
-			}
-
-			if (
-				$parent instanceof File &&
-				$file = $parent->page()?->file($path)
-			) {
-				return $file;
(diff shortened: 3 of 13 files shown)


src/Uuid/FieldUuid.php
```diff
@@ -108,13 +110,16 @@ public static function index(): Generator
...
-	 * constisting of three parts:
+	 * consisting of three parts:
 	 * - parent UUID including scheme
 	 * - field name
 	 * - UUID id string for model
 	 */
 	public function value(): array
 	{
+		/**
+		 * @var \Kirby\Cms\Site|\Kirby\Cms\Page|\Kirby\Cms\User $model
```
lukasbestle: `Question`: Couldn't it also be a file?

src/Uuid/Uuid.php
```diff
@@ -214,6 +189,43 @@ final public static function for(
...
+			return match ($uri) {
+				'page'   => new PageUuid(uuid: $uuid, context: $context),
+				'file'   => new FileUuid(uuid: $uuid, context: $context),
+				'site'   => new SiteUuid(uuid: $uuid, context: $context),
+				'user'   => new UserUuid(uuid: $uuid, context: $context),
+				// TODO: activate for uuid-block-structure-support
+				// 'block'  => new BlockUuid(uuid: $seed, context: $context),
+				// 'struct' => new StructureUuid(uuid: $seed, context: $context),
+				default  => throw new InvalidArgumentException(
+					message: 'Invalid UUID URI "' . $uri . '" in ' . $uuid
```
lukasbestle:
```suggestion
					message: 'Invalid UUID type "' . $uri . '" in ' . $uuid
```

Probably makes sense to rename the variable as well. Not sure why we used `$uri` here.

src/Uuid/Uuid.php
```diff
@@ -214,6 +189,43 @@ final public static function for(
 		};
 	}
 
+	/**
+	 * Creates a UUID object from a UUID string
+	 * @since 6.0.0
+	 */
+	final public static function from(
+		string $uuid,
+		string|array|null $scheme = null,
```
lukasbestle: `Suggestion`: `Uuid::is()` uses the name `$type`, I'd use the same here for consistency:

```suggestion
		string|array|null $type = null,
```

config/tags.php
```diff
@@ -206,8 +206,8 @@
 
 			// if value is a UUID, resolve to page/file model
 			// and use the URL as value
-			if (Uuid::is($tag->value, ['page', 'file']) === true) {
-				$tag->value = Uuid::for($tag->value)?->toUrl();
+			if ($uuid = Uuid::from($tag->value, ['page', 'file'])) {
+				$tag->value = $uuid->toUrl();
```
lukasbestle: `Issue`: This changes behavior. Before, a valid but non-existent UUID would lead to an empty value, which would be handled below. Now, the original UUID URI would stay in the value unconverted.

src/Cms/HasFiles.php
```diff
@@ -72,8 +72,8 @@ public function file(
 		}
 
 		// find by global UUID
-		if (Uuid::is($filename, 'file') === true) {
-			return Uuid::for($filename, $this->$in())->model();
+		if ($uuid = Uuid::from($filename, 'file', $this->$in())) {
+			return $uuid->model();
```
lukasbestle: `Issue`: Same here, but not as critical. Here, the valid but non-existent UUID will make Kirby execute all of the other checks instead of exiting early (but before with an error like `Call to function model() on null`).

src/Text/KirbyTag.php
```diff
@@ -116,35 +116,29 @@ public static function factory(...$arguments): string
...
 
-		if (
-			is_object($parent) === true &&
-			method_exists($parent, 'file') === true &&
-			$file = $parent->file($path)
-		) {
-			return $file;
-		}
+			if (method_exists($parent, 'file') === true) {
+				return $parent->file($path);
```
lukasbestle: `Issue`: If called on `$parent = Page`, this will return `null` if that page doesn't have the file `$path`. The old code continued to the very last line of the method and checked for site files of that name.

lukasbestle (commented):
The `for` and `from` names are great here!

src/Uuid/FieldUuid.php
distantnative: You are right. It seems I was rather thinking about the result of `$model->parent()` here.

config/tags.php
distantnative: I think here we could solve it by just turning it into a one-liner:

```php
$tag->value = Uuid::from($tag->value, ['page', 'file'])?->toUrl();
```

as this would also lead to `$tag->value` being `null` afterwards if it isn't a valid UUID.

src/Cms/HasFiles.php
distantnative: That raises maybe a question: If something looks like a UUID but cannot be resolved to a model, would we rather have it throwing some form of error or just ignore it and try to resolve it with the other checks?

config/tags.php
lukasbestle: Not quite. Then the value will also be null for all non-UUID links (links that don't match the `page://something` or `file://something` patterns).

src/Cms/HasFiles.php
lukasbestle: I think we should throw an error. The UUID pattern is very unique (`file://something`) and it is quite unlikely that something else was meant. Especially not a filename as filenames cannot contain two slashes sequentially.

config/tags.php
distantnative: Reverted it

src/Text/KirbyTag.php
distantnative: Reverted it

src/Cms/HasFiles.php
distantnative: Reverted it

bastianallgeier (approved):

PR #7545, opened 2025-08-15
refact: `Uuid\Permalink` class

distantnative:
### ♻️ Refactored

- Moved permalink logic to new `Kirby\Uuid\Permalink` class

### 🚨 Breaking changes

- `Kirby\Uuid\Uuid::for()` does not resolve any permalinks anymore. Use `Kirby\Uuid\Permalink::from()` instead.

### For review team


src/Uuid/Permalink.php
```diff
@@ -0,0 +1,79 @@
...
+ * @since     6.0.0
+ */
+class Permalink
+{
+	public function __construct(
+		public FileUuid|PageUuid $uuid
+	) {
+	}
+
+	public static function from(string $uuid): static|null
```
lukasbestle: `Suggestion`: As there are two factory methods that take different input, bringing their names in line could increase the overview:

```php
Permalink::fromUuid(string $uuid)
Permalink::fromUrl(string $url)
```

This would also be consistent with the `uuid()` and `url()` methods that convert in the opposite direction.

distantnative: What would you think about `::for(string $uuid)` and `::from(string $url)`? Kinda similar how the following PR then does it for the main UUID class as well with the uuid/model cases (where here we have uuid/url).

lukasbestle: Right, that could work!

bastianallgeier (approved):

bastianallgeier:
The changes look good to me. It would be great if we could get the coverage up to 100% for the Permalink class though.

PR #7547, opened 2025-08-16
test: Add coverage attributes for `Kirby\Http`

distantnative:
## Description

This PR
- Adds coverage attributes to `Kirby\Http` (we missed adding them before)
- Updates the code style of the classes to modernize the code a bit

### Housekeeping

- Clean up code style of the `Http` classes

### Deprecated

- `Kirby\Http\Router::find()`: Passing `$ignore = null` has been deprecated (pass an empty array or omit the argument instead)

### DIFF
--- a/src/Http/Params.php
+++ b/src/Http/Params.php
@@ -38,7 +38,7 @@ class Params extends Obj implements Stringable
 	 */
 	public static function extract(string|array|null $path = null): array
 	{
-		if (empty($path) === true) {
+		if ($path === null || $path === '' || $path === []) {
 			return [
 				'path'   => null,
 				'params' => null,
@@ -62,12 +62,16 @@ class Params extends Obj implements Stringable
 					continue;
 				}
 
-				$paramParts = Str::split($p, $separator);
-				$paramKey   = $paramParts[0] ?? null;
-				$paramValue = $paramParts[1] ?? null;
+				$parts = Str::split($p, $separator);
 
-				if ($paramKey !== null) {
-					$params[rawurldecode($paramKey)] = $paramValue !== null ? rawurldecode($paramValue) : null;
+				if ($key = $parts[0] ?? null) {
+					$key = rawurldecode($key);
+
+					if ($value = $parts[1] ?? null) {
+						$value = rawurldecode($value);
+					}
+
+					$params[$key] = $value;
 				}
 
 				unset($path[$index]);
@@ -89,7 +93,7 @@ class Params extends Obj implements Stringable
 
 	public function isEmpty(): bool
 	{
-		return empty((array)$this) === true;
+		return (array)$this === [];
 	}
 
 	public function isNotEmpty(): bool
@@ -106,15 +110,7 @@ class Params extends Obj implements Stringable
 	 */
 	public static function separator(): string
 	{
-		if (static::$separator !== null) {
-			return static::$separator;
-		}
-
-		if (DIRECTORY_SEPARATOR === '/') {
-			return static::$separator = ':';
-		}
-
-		return static::$separator = ';';
+		return static::$separator ??= DIRECTORY_SEPARATOR === '/' ? ':' : ';';
 	}
 
 	/**
@@ -134,7 +130,9 @@ class Params extends Obj implements Stringable
 
 		foreach ($this as $key => $value) {
 			if ($value !== null && $value !== '') {
-				$params[] = rawurlencode($key) . $separator . rawurlencode($value);
+				$key      = rawurlencode($key);
+				$value    = rawurlencode($value);
+				$params[] = $key . $separator . $value;
 			}
 		}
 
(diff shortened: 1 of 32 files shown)


src/Http/Params.php
```diff
@@ -38,7 +38,7 @@ public function __construct(array|string|null $params)
 	 */
 	public static function extract(string|array|null $path = null): array
 	{
-		if (empty($path) === true) {
+		if ($path === null || $path === '' || $path === []) {
```
afbora, line 41: I'm sorry, but I don't like this syntax 😅 It may use for a single type as you did, but writing it separately for 3 different types is not readable at all. IMHO `empty($path) === true` is simple and clear 🙂

distantnative: The problem with `empty()` is that it is very opaque and checks a lot more things than one often considers. See
https://www.beberlei.de/post/when_to_use_empty_in_php_i_say_never
https://dev.to/klnjmm/never-use-empty-function-in-php-4pb0

That's why I think we should stay away from using `empty()` wherever we can and rather spell out the individual conditions - even if this can look a little verbose at times.

afbora: I know this article. Since there is an extra 0.01 seconds of speed in 1 million iterations, I don't think it makes sense to reduce readability too much. You can test benchmark with following:

````php
<?php

function benchmark($name, $iterations, $action)
{
	$time = microtime(true);
	for ($i = 0; $i <= $iterations; ++$i) {
		$action();
	}
	echo $name . ' ' . round(microtime(true) - $time, 6) . "\n";
}

$iterations = 1000000;
$x = array();
$actions = array(
	"empty()"   => function () use ($x) {
		if (empty($x)) {}
	},
	"type2type" => function () use ($x) {
		if ($x === null || $x === [] || $x === '') {}
	}
);

foreach ($actions as $name => $action) {
	benchmark($name, $iterations, $action);
}
````

Comment (distantnative) [reply src/Http/Params.php:41]: It's not about speed/performance and it's the contrary in terms of readability: `empty()` has worse readability as it obfuscates what it does. As the article says:

> - it tests both for the existance and the non-falsy value and as a reader the writers intent is not clear.
>- it does not communicate what type a variable has, seeing empty($value) does not narrow down what value is enough.
>- It hides errors caused by typos in the variable name passed to empty.

That's why many PHP developers try to ban `empty()` from there code. And I think we should try as much as we can when revising and writing new code. This should not be about taste - I agree the short `empty()` looks better - but what helps us in the long term to write more reliable, less error prone code.

File: src/Http/Request.php [hunk 2]
```diff
@@ -411,14 +418,16 @@ protected function authString(): string|null
 		// because empty strings are treated as invalid
 		// but the `??` operator wouldn't do the fallback
 
-		$option = $this->options['auth'] ?? null;
-		if (empty($option) === false) {
-			return $option;
+		if ($option = $this->options['auth'] ?? null) {
+			if ($option !== '') {
+				return $option;
+			}
```
lukasbestle: `Comment`: Not a huge fan of the nested ifs TBH. I think a separate assignment followed by `if (is_string($option) === true && $option !== '')` would be cleaner.

src/Http/Request.php
```diff
@@ -411,14 +418,16 @@ protected function authString(): string|null
...
 		}
 
-		$header = $this->header('authorization');
-		if (empty($header) === false) {
-			return $header;
+		if ($header = $this->header('authorization')) {
+			if ($header !== '') {
+				return $header;
+			}
```
lukasbestle: Same here.

src/Http/Router.php
```diff
@@ -172,16 +172,14 @@ public function find(
 		}
 
 		// remove leading and trailing slashes
-		$path = trim($path, '/');
+		$path     = trim($path, '/');
+		$ignore ??= [];
```
lukasbestle, line 179: `Suggestion`: We could deprecate being able to pass `null` for `$ignore` to increase cleanliness.

src/Http/Uri.php
```diff
@@ -503,8 +504,9 @@ public function toString(): string
 	 */
 	public function unIdn(): static
 	{
-		if (empty($this->host) === false) {
-			$this->setHost(Idn::encode($this->host));
+		if ($this->host !== null && $this->host !== '') {
```
lukasbestle:
```suggestion
		if ($this->isAbsolute() === true) {
```

src/Http/Uri.php
```diff
@@ -281,8 +281,9 @@ public function https(): bool
 	 */
 	public function idn(): static
 	{
-		if (empty($this->host) === false) {
-			$this->setHost(Idn::decode($this->host));
+		if ($this->host !== null && $this->host !== '') {
```
lukasbestle:
```suggestion
		if ($this->isAbsolute() === true) {
```

distantnative:
I love how suddenly the coverage tests kicked in and are unhappy 😆 Will check it.

lukasbestle (approved):

PR #7548, opened 2025-08-16
refact!: Clean up `VersionsButton`

distantnative:
## Description

Splitting up the code in `VersionsButton` into smaller methods - makes it easier to customize it a bit further down the road in a v6 PR. But I think these changes could already go into v5.1.

### Breaking change
- `$versionId` parameter for `Kirby\Panel\Ui\Button\VersionsButton::__construct()` has been renamed to `$mode`

### DIFF
--- a/src/Panel/Ui/Button/VersionsButton.php
+++ b/src/Panel/Ui/Button/VersionsButton.php
@@ -21,37 +21,94 @@ class VersionsButton extends ViewButton
 {
 	public function __construct(
 		ModelWithContent $model,
-		VersionId|string $versionId = 'latest'
+		public string $mode = 'latest'
 	) {
-		$versionId = $versionId === 'compare' ? 'compare' : VersionId::from($versionId)->value();
-		$viewUrl   = $model->panel()->url(true) . '/preview';
-
 		parent::__construct(
+			model: $model,
 			class: 'k-versions-view-button',
-			icon: $versionId === 'compare' ? 'layout-columns' : 'git-branch',
-			options: [
-				[
-					'label'   => I18n::translate('version.latest'),
-					'icon'    => 'git-branch',
-					'link'    => $viewUrl . '/latest',
-					'current' => $versionId === 'latest'
-				],
-				[
-					'label'   => I18n::translate('version.changes'),
-					'icon'    => 'git-branch',
-					'link'    => $viewUrl . '/changes',
-					'current' => $versionId === 'changes'
-				],
-				'-',
-				[
-					'label'   => I18n::translate('version.compare'),
-					'icon'    => 'layout-columns',
-					'link'    => $viewUrl . '/compare',
-					'current' => $versionId === 'compare'
-				],
-
-			],
-			text: I18n::translate('version.' . $versionId),
+			icon: $this->icon(),
+			text: $this->i18n('version.' . $this->mode()),
 		);
+	}
+
+	/**
+	 * Returns the button icon based on the view's mode
+	 * @since 5.1.0
+	 */
+	public function icon(): string
+	{
+		return match ($this->mode) {
+			'compare' => 'layout-columns',
+			default   => 'git-branch',
+		};
+	}
+
+	/**
+	 * Whether the given mode is the current mode
+	 * @since 5.1.0
+	 */
+	public function isCurrent(string $mode): bool
+	{
+		return $this->mode() === $mode;
+	}
+
+	/**
+	 * Returns the view's mode using the proper
+	 * values for version IDs
+	 * @since 5.1.0
+	 */
+	public function mode(): string
+	{
+		return match ($this->mode) {
+			'compare' => 'compare',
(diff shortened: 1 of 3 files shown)


src/Panel/Ui/Buttons/VersionsButton.php
```diff
@@ -21,37 +21,93 @@ class VersionsButton extends ViewButton
 {
 	public function __construct(
 		ModelWithContent $model,
-		VersionId|string $versionId = 'latest'
+		public VersionId|string $versionId = 'latest'
```
lukasbestle: Having an instance prop that can carry both types is confusing and error-prone IMO.

`Suggestions` for possible solutions:

- We could create a new `PreviewVersionId` class that supports the `compare` case. However it probably shouldn't `extend VersionId` because of [type-safety](https://www.php.net/manual/en/language.enumerations.object-differences.inheritance.php). So not really clean either.
- We could always convert the passed value to a string in the constructor as we did before. Then the instance prop can always be a string.

distantnative: You are right, but given the use case and rest of the code, I think this should always be a string then (maybe also different name as it already can also be "compare" and could be something like "edit" in the future). I will change it.

distantnative: @lukasbestle Actually, I think we have to leave it like this for 5.1.0 to not introduce that breaking change. We can adapt it directly on `v6/develop` after merging. What do you think?

lukasbestle: Which breaking change? Before this PR, we didn't have an instance prop at all. The constructor argument can stay the same and will be normalized in the constructor.

distantnative: But the constructor parameter existed already before this PR and we would also rename it then, so that would be a breaking change.

lukasbestle: Ah, the renaming. Yes, that is a good point. Then I agree to move the breaking change to a separate PR for v6.

distantnative: Actually, we do not need to have this in v5.1 in general. I moved it to 5.1 as we try to back port as much as possible. But in this case, I think it then doesn't make sense to split this up further but rather do the whole PR for v6.

bastianallgeier: Yes, let's move it to v6

lukasbestle (commented):
Don't we also need to update `buttons.php` in the site area to use the new `$mode` argument?

distantnative:
Sorry, my bad. Let's pause this actually.

src/Panel/Ui/Button/VersionsButton.php
```diff
@@ -21,37 +21,94 @@ class VersionsButton extends ViewButton
...
-					'current' => $versionId === 'compare'
-				],
+			icon: $this->icon(),
+			text: $this->i18n('version.' . $this->mode()),
+		);
+	}
+
+	/**
+	 * Returns the button icon based on the view's mode
+	 * @since 5.1.0
```
lukasbestle: `Todo`: These versions need to be updated to 6.0.0.

src/Panel/Ui/Button/VersionsButton.php
```diff
@@ -21,37 +21,94 @@ class VersionsButton extends ViewButton
...
+	 */
+	public function icon(): string
```
lukasbestle: `Question`: Should this methods that is called in the constructor actually be `public`? Public methods overshadow `Component::__call()`, so `$versionsButton->icon()` for example will always return the dynamic value based on the mode and not the prop value. As the `$icon` property is `public` as well, one could set the property to a different value on the instance, creating a discrepancy between the property and return value of the method.

In this case, this likely won't cause any problems, I'm asking more about the general coding pattern for such component classes.

distantnative: @lukasbestle I think in this case (but also in a looot of other places in Kirby), we use public methods to be able to test them more easily where based on the general usage they could be protected instead. Maybe if we could find an easy/lean unit test style to keep the protected but still test the method in isolation. I roughly know that one could turn a protected method public via reflections in the unit tests, but IMO that has quite a bit of boilerplate code attached to it, which is why I think in many places we have stayed away from it (and rather made the method public).

lukasbestle: Yes, that is definitely true. I have used this pattern as well and generally it has worked quite well.

I think in this case however, we don't need to make the method public to test it. If we make it `protected`, we can still call `$versionsButton->icon()`, but PHP will then automatically fall back to the magic `__call()` method, thus accessing the `$icon` property that we want to test.

distantnative: @lukasbestle I have changed it to protected.

bastianallgeier (approved):

PR #7549, opened 2025-08-16
feat: New `session.cookieDomain` option

lukasbestle:
## Description

Implements <https://feedback.getkirby.com/628>.

### 🎉 Features

- New `session.cookieDomain` option that allows to share the session cookie across subdomains with a shared `sessions` folder <https://feedback.getkirby.com/628>

## Docs

For <https://getkirby.com/docs/reference/system/options/session>:

```md
return [
    'session' => [
        'durationNormal' => 7200,            // default: 2 hours
        'durationLong'   => 1209600,         // default: 2 weeks
        'timeout'        => 1800,            // default: half an hour
        'cookieDomain'   => 'getkirby.com',  // default: automatic browser behavior
        'cookieName'     => 'kirby_session',
        'gcInterval'     => 100              // default: cleanup every ~100 requests
    ]
];

...

<since v="5.1.0">
## `cookieDomain`

The topmost-level domain on which the session cookie should be sent.

<info>
You can read more about the behavior of the `domain` attribute for cookies in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#define_where_cookies_are_sent).

Note that setting the `cookieDomain` option disables Kirby setting the `path` attribute to the cookie (relevant in subfolder setups).
</info>
</since>
```

### For review team

### DIFF
--- a/src/Session/Session.php
+++ b/src/Session/Session.php
@@ -465,9 +465,12 @@ class Session
 
 		// (re)transmit session token
 		if ($this->mode === 'cookie') {
+			$cookieDomain = $this->sessions->cookieDomain();
+
 			Cookie::set($this->sessions->cookieName(), $this->token(), [
 				'lifetime' => $this->tokenExpiry,
-				'path'     => Url::index(['host' => null, 'trailingSlash' => true]),
+				'path'     => $cookieDomain ? '/' : Url::index(['host' => null, 'trailingSlash' => true]),
+				'domain'   => $cookieDomain,
 				'secure'   => Url::scheme() === 'https',
 				'httpOnly' => true,
 				'sameSite' => 'Lax'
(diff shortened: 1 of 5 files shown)


distantnative (approved):

src/Session/Session.php
```diff
@@ -465,9 +465,12 @@ public function regenerateToken(): void
 
 		// (re)transmit session token
 		if ($this->mode === 'cookie') {
+			$cookieDomain = $this->sessions->cookieDomain();
+
 			Cookie::set($this->sessions->cookieName(), $this->token(), [
 				'lifetime' => $this->tokenExpiry,
-				'path'     => Url::index(['host' => null, 'trailingSlash' => true]),
+				'path'     => $cookieDomain ? '/' : Url::index(['host' => null, 'trailingSlash' => true]),
+				'domain'   => $cookieDomain,
```
afbora, line 473: I just wonder that `setcookie()` method accepts only string for `domain` argument, but we pass `string|null`. There seems to be an inconsistency.

lukasbestle: `Cookie::set()` already defaults to `null` if the domain option is not passed. I think PHP silently treats that the same as an empty string. On that higher level, I think `null` makes more sense and is cleaner as it means "omit the attribute" in contrast to "set it to an empty string".

afbora: OK. I just wanted to make sure it wasn't overlooked 👍

afbora (approved):

PR #7546, opened 2025-08-16
feat: `Query::merge()` + `Params::merge()`

distantnative:
## Description

- [x] https://github.com/getkirby/kirby/pull/7547

### ✨ Enhancements

- New `Kirby\Http\Query::merge()` and `Kirby\Http\Params::merge()` methods

### For review team

- [x] Add changes & docs to release notes draft in Notion

PR #7550, opened 2025-08-16
feat: Follow navigation in preview view

distantnative:
## Description

Feature for the new Panel preview view that allows users to navigate in the frontend iframe and the Panel following their navigation (syncing to the newly page they navigated to):

https://github.com/user-attachments/assets/134b31d0-dba0-42d9-88e3-ac494199ee7e

### Merge first
- [x] https://github.com/getkirby/kirby/pull/7546

### 🎉 Features

- Preview view: When you navigate through your frontend, the Panel syncs to the new location

### For review team

### DIFF
--- a/panel/src/components/Views/Preview/PreviewBrowser.vue
+++ b/panel/src/components/Views/Preview/PreviewBrowser.vue
@@ -67,6 +67,7 @@ export default {
 				return;
 			}
 
+			// open external links and Panel links in new tab
 			if (
 				link.href.startsWith(location.origin) === false ||
 				link.href.startsWith(this.$panel.urls.panel) === true
@@ -75,6 +76,7 @@ export default {
 				return true;
 			}
 
+			// catch internal links and emit navigate event
 			e.preventDefault(e);
 			this.$emit("navigate", link.href);
 		},
--- a/panel/src/components/Views/Preview/PreviewBrowser.vue
+++ b/panel/src/components/Views/Preview/PreviewBrowser.vue
@@ -56,21 +56,30 @@ export default {
 		this.$events.off("content.publish", this.reload);
 	},
 	methods: {
+		onClick(e) {
+			const link = e.target.closest("a");
+
+			if (!link) {
+				return;
+			}
+
+			if (!link.href || link.onclick) {
+				return;
+			}
+
+			e.preventDefault();
+
+			if (link.href.startsWith(location.origin) === false) {
+				return window.open(link.href, "_blank");
+			}
+
+			this.$emit("navigate", link.href);
+		},
 		onLoad() {
+			this.window.document.addEventListener("click", this.onClick);
+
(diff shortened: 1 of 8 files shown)


panel/src/components/Views/Preview/PreviewBrowser.vue
```diff
@@ -56,8 +55,25 @@ export default {
...
+
+				if (link.href.startsWith(location.origin) === false) {
+					link.target = "_blank";
+					continue;
+				}
+
+				link.addEventListener("click", (e) => {
+					e.preventDefault();
+					this.$emit("navigate", link.href);
+				});
```
bastianallgeier: I know that you had issues with event delegation, but I just realized that this will only work for elements that are present on load. For links that get injected after ajax calls or other more reactive JS parts, this will not be working.

distantnative: Hmm that's true, but as shown by our Demokit, event delegation is even worse as it gets easily suppressed by a script of the page (and I think these kind of scripts are even more likely/common).

distantnative: @bastianallgeier Maybe our best chance is to act on both? Pushed a version.

bastianallgeier:
There are two things that I found when testing it with the demokit. 

1. When you click on a link that has target _blank, the onClick handler would still try to open a new window for it and Firefox supresses that with a warning. I think we should check what the target for the link is. 

2. When I click on one of the "View source" links in the footer, the link is internal, but there's no matching model and a warning will pop up. Could we check for that error and still open the link in a new window?

distantnative:
@bastianallgeier I have dived into this again

1. I hope this has been solved with https://github.com/getkirby/kirby/pull/7550/changes/2169a5aaa7b8bd755e1ea76483bb5184ae6a4064

2. Checking for that error is pretty hard as our Panel setup just grabs it as RequestError with the actual error message ending up deep in the view object. And event if I try to grab that at the request module, I still have trouble routing it back to the original view that called `$panel.view.reload()`. What does work is not throwing a `LogicException` but a `Redirect` though this does not open a new window but redirects the current one. Would this be an acceptable middle ground? Or do you see any way to catch the error inside `k-preview-view`?

distantnative:
@bastianallgeier I added a checkbox to enable/disable the behavior. We probably want a better UI for that. But I realized working on the live edit mode that there sometimes following navigation (and pulling up the right model/fields) is great, but sometimes you might rather want to look at e.g. the blog overview page while editing a specific blog post (and stay on the blog post fields no matter how navigating around in the iframe).

bastianallgeier:
@distantnative maybe it's a good idea to work on this in combination with the live editing feature for v6 then?

distantnative:
We could move it to v6 as well, would be fine with me.

distantnative:
Added it to https://github.com/getkirby/kirby/pull/7780

PR #7553, opened 2025-08-18
lab: Active toggle in pages table

bastianallgeier:
You can try the effects of this lab PR in https://sandbox.test/panel/pages/other+page-table

### DIFF
--- a/config/sections/pages.php
+++ b/config/sections/pages.php
@@ -1,5 +1,6 @@
 <?php
 
+use Kirby\Cms\App;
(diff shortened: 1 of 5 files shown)


distantnative:
As you wrote, the biggest problem here is which version to update. It should be changes in one way but the section doesn't display the values of changes but latest. Probably better to stick to disabling for now.

distantnative:
@bastianallgeier I would vote to close this for now. As we both concluded, in its current state it would cause a lot of confusion around what version actually gets updated.

PR #7556, opened 2025-08-19
Remove unnecessary PHP code

distantnative:
## Description

Those functions will throw deprecation errors in PHP 8.5 - we can already remove them as they have no function anymore. See individual commit messages for what exactly got removed.

### 🧹 Housekeeping

- Removed old PHP functions that will get deprecated

### For review team

### DIFF
--- a/src/Filesystem/Mime.php
+++ b/src/Filesystem/Mime.php
@@ -188,7 +188,6 @@ class Mime
 		) {
 			$finfo = finfo_open(FILEINFO_MIME_TYPE);
 			$mime  = finfo_file($finfo, $file);
-			finfo_close($finfo);
 			return $mime;
 		}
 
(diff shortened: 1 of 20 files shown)


bastianallgeier (approved):

lukasbestle:
@distantnative IMO it would be good to backport this to v5 already. As the removed code only affects PHP versions that we already don't support anymore, there are no breaking changes. But we need to remove those calls in 5.x as soon as we want to add support for PHP 8.5 to v5 – which will be relevant because it will be published long before Kirby 6.0.

distantnative:
You're right. I thought it was just relevant for v6, but true the timelines will make it relevant for v5 as well. Will try to cherrypick it directly on `develop-minor`.

distantnative:
@lukasbestle Actually, I will cherry-pick them after 5.1 release. I think `develop-minor` currently still needs to be protected for 5.1 final release.

PR #7558, opened 2025-08-19
lab: New Model classes and storage ideas

bastianallgeier:
## Moved into Notion
https://www.notion.so/getkirby/New-Model-classes-and-storage-ideas-2ae59ef0a5cb80209661f2244d878c6a?v=2ae59ef0a5cb81fb9a98000c648995ef&source=copy_link

### DIFF
--- a/src/Model/FileParent.php
+++ b/src/Model/FileParent.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Kirby\Model;
+
+class FileParent
+{
+	protected Model $parent;
+
+	public function __construct(
+		protected Model|string $pointer,
+	) {
+	}
+
+	public static function from(Model|self|string $parent): static
+	{
+		if ($parent instanceof self) {
+			return $parent;
+		}
+
+		return new static($parent);
+	}
+
+	public function load(): Model
+	{
+		return $this->parent ??= match (true) {
+			$this->pointer === null
+				=> null,
+			$this->pointer instanceof Model
+				=> $this->pointer,
+			default
+				=> Page::findByIdentifier($this->pointer),
+		};
+	}
+}
--- a/src/Model/FileStorage.php
+++ b/src/Model/FileStorage.php
@@ -1,0 +1,123 @@
+<?php
+
+namespace Kirby\Model;
+
+use Kirby\Data\Data;
+use Kirby\Exception\DuplicateException;
+use Kirby\Filesystem\F;
+use Kirby\Uuid\Uuid;
+use Throwable;
+
+class FileStorage extends Storage
+{
+	public function changeFilename(FileMeta $oldMeta, FileMeta $newMeta): void
+	{
+		$name      = F::name($newMeta->filename);
+		$extension = F::extension($newMeta->filename);
+		$directory = dirname($newMeta->identifier);
+
+		$newMeta->identifier = $directory . '/' . $name . '.' . $extension;
+
+		$this->moveIdentifier($oldMeta, $newMeta);
+	}
+
+	public function changeNum(FileMeta $oldMeta, FileMeta $newMeta): void
+	{
+		$this->update([
+			'num' => $newMeta->num,
+		]);
+	}
+
+	public function changeTemplate(FileMeta $oldMeta, FileMeta $newMeta): void
+	{
+		$this->update([
+			'template' => $newMeta->template,
+		]);
+	}
+
+	protected static function contentFile($meta): string
+	{
+		return $meta->identifier . '.txt';
+	}
+
+	public static function create(FileMeta $meta): File
+	{
+		// create the new identifier
+		$meta->identifier = static::createIdentifier($meta);
+
+		// generate the new UUID if it is not set yet
+		$meta->uuid ??= Uuid::generate();
+
+		// use the default template if it is not set yet
+		$meta->template ??= 'default';
+
+		if (is_file($meta->identifier) === true) {
+			throw new DuplicateException('The file already exists');
+		}
+
+		// copy the source file to the new identifier
+		F::copy($meta->source, $meta->identifier);
+
+		// create the new content file
+		Data::write(static::contentFile($meta), [
+			'template' => $meta->template,
+			'num'      => $meta->num,
+			'uuid'     => $meta->uuid,
+		]);
+
+		return static::find(File::class, $meta->identifier);
+	}
+
+	protected static function createIdentifier(FileMeta $meta): string
+	{
+		// create the new identifier
+		$identifier = $meta->parent->load()->meta()->identifier;
+
+		// add the filename
+		$identifier .= '/' . $meta->filename;
+
+		return $identifier;
+	}
+
+	public static function find(string $class, string $identifier): Model|null
+	{
+		if (is_file($identifier) === false) {
+			return null;
+		}
+
+		$meta = new FileMeta(
+			filename: basename($identifier),
+			identifier: $identifier,
+			parent: FileParent::from(dirname($identifier)),
+			source: $identifier,
+		);
+
+		try {
+			$content	    = Data::read(static::contentFile($meta));
+			$meta->num      = $content['num']      ?? null;
+			$meta->template = $content['template'] ?? 'default';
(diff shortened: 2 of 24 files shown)


src/Model/UserMeta.php
```diff
@@ -0,0 +1,26 @@
...
+class UserMeta extends ModelMeta
+{
+	public function __construct(
+		public string $identifier,
+		public string|null $email = null,
+		int $created = 0,
+		public string|null $language = null,
+		int $modified = 0,
+		public string|null $name = null,
+		public string|null $password = null,
```
lukasbestle, line 14: `Todo`: We should go for `$secrets` here as the `.htpasswd` file can also contain a TOTP secret and later other secrets (like the WebAuthn key). The `password` methods are only still present in `File` for compatibility. We shouldn't use this term going forward.

src/Model/Storage.php
```diff
@@ -0,0 +1,67 @@
...
+	public function read(): array
+	{
+		try {
+			return Data::read($this->contentFile($this->model->meta()));
+		} catch (Throwable $e) {
+			return [];
+		}
+	}
+
+	public function changeMeta(array $meta): ModelMeta
```
lukasbestle, line 38: `Suggestion`: We could already pass `ModelMeta $meta` here so that the storage handler has a cleaner interface. It shouldn't be the handler's obligation to construct the meta object. Also it would allow the handler to return `void` when changing meta (consistent with the existing methods for updating properties).

src/Model/Storage.php
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		foreach (array_keys($meta) as $key) {
+			$this->{'change' . ucfirst($key)}($this->model->meta(), $newMeta);
+		}
+
+		return $newMeta;
+	}
+
+	public function update(array $data): void
```
lukasbestle, line 53: `Suggestion`: We could use the opportunity to bring the two update methods in line: Maybe `updateContent` and `updateMeta`? We have the same issue with many other methods like `read` and `write`. Maybe we need a totally different solution to keep the meta and content spheres separate.

src/Model/Storage.php
```diff
@@ -0,0 +1,67 @@
...
+	public function changeMeta(array $meta): ModelMeta
+	{
+		$newMeta = clone $this->model->meta();
+
+		foreach ($meta as $key => $value) {
+			$newMeta->$key = $value;
+		}
+
+		foreach (array_keys($meta) as $key) {
+			$this->{'change' . ucfirst($key)}($this->model->meta(), $newMeta);
```
lukasbestle, line 47: `Todo`: We need to detect if each meta prop actually changed. Otherwise we will run a lot of operations that aren't needed. We could do it in each individual action, but that's a lot of duplicated code.

src/Model/FileStorage.php
```diff
@@ -0,0 +1,123 @@
...
+		return $meta->identifier . '.txt';
+	}
+
+	public static function create(FileMeta $meta): File
+	{
+		// create the new identifier
+		$meta->identifier = static::createIdentifier($meta);
+
+		// generate the new UUID if it is not set yet
+		$meta->uuid ??= Uuid::generate();
```
lukasbestle, line 49: `Thought`: We need to find a good place for such code that isn't actually dependent on the storage handler. A database handler shouldn't have to reimplement things like UUID generation.

src/Model/FileStorage.php
```diff
@@ -0,0 +1,123 @@
...
+
+namespace Kirby\Model;
+
+use Kirby\Data\Data;
+use Kirby\Exception\DuplicateException;
+use Kirby\Filesystem\F;
+use Kirby\Uuid\Uuid;
+use Throwable;
+
+class FileStorage extends Storage
```
lukasbestle, line 11: `Todo`: Reading the content for a file will need to filter out the `num` and `template` fields as the file handler is the only one that needs to store them in the content file.

src/Model/HasUuid.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Kirby\Model;
+
+trait HasUuid
+{
+	public function uuid(): string|null
+	{
+		return $this->meta->uuid ??= $this->storage()->read()['uuid'] ?? null;
```
lukasbestle, line 9: `Question`: Won't this try to read the UUID from content? It should be in meta.

src/Model/Page.php
```diff
@@ -0,0 +1,83 @@
...
+			status: $status,
+			template: $template,
+			uuid: $uuid,
+		);
+	}
+
+	public function children(): Collection
+	{
+		$children = $this->storage()->children();
+		$children = array_map(fn($child) => Page::findByIdentifier($child), $children);
```
lukasbestle, line 46: `Idea`: We could use the opportunity to look into lazy collections some more for children and files. Adding support for lazy lookup later would require significant changes to the storage logic, so best to plan ahead in this area. Lazy collections would be a massive performance win.

As far as I see it, lazy collections are well doable with PHP iterators, but not with our existing `Collection` base class as that heavily assumes that all collection items are already there. But maybe it can still work: We could create the collection with the placeholders and let it hydrate the individual objects when they are needed. That's maybe not 100% optimized (as we still need to fetch all identifiers from storage), but still a good improvement. Or we go further and add logic to the storage so that we can also just create a collection with a number of items but only hydrate the individual identifiers if they are needed (allowing a database handler to paginate queries in a huge data set where even getting all identifiers at once would be too much). What we also need to think about is if and how we can implement more complex query logic in the storage so that something like `$pages->filterBy('template', 'something')` will not need to go over all model objects but be passed on to the storage for a direct query.

TL;DR: That can all get incredibly complex, but I think it's worth it to at least play around with these ideas before we commit on a storage setup. This part will be super hard to change later.

src/Model/FileParent.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Kirby\Model;
+
+class FileParent
```
lukasbestle, line 5: `Suggestion`: We can use the same placeholder classes in lazy children and files collections. So maybe we could give them a more general name like `FilePlaceholder` or `FileStub`.

lukasbestle (commented):
As I said in our call, I love the general concept.

I'm already adding comments on things I noticed (even though I know things are very much in flux at this point and not all comments will end up being relevant – just so that I don't forget my first thoughts and so that we can see in which general direction we want to go in).

> We should start working towards normalizing `num` and `sort` for Pages and Files. We should only end up with one and I would probably vote for `num`.

Another vote for `num`.

src/Model/Storage.php
bastianallgeier: Yes, that would indeed be much better.

src/Model/Storage.php
bastianallgeier: You are totally right. I didn't want to throw all our storage stuff over board already and tried to keep close to what we have, but it does not feel ideal indeed. We could also really think about a separation between Content Storage, Meta Storage and for files there could even be a Source Storage. But we also have to be careful that it does not get too crazy complex.

src/Model/HasUuid.php
bastianallgeier: You are totally right. This should probably go through a readMeta method or something similar.

src/Model/FileParent.php
bastianallgeier: Good point! More name ideas: 

- AsyncFile
- FileFetcher
- FileLoader

But the Placeholder idea is already pretty great.

src/Model/Storage.php
lukasbestle: The challenge is that these different types of data are so deeply interconnected. E.g. the file system stores the translated slugs, file sorting numbers and UUIDs in the content, the storage location of the content and file source depends on the metadata etc. Based on this, my first assumption is that we can't easily separate these parts into separate storage handlers. But if we'd find a good and robust way, that would certainly be the cleanest possible setup.

About breaking changes: I don't think we need to be too careful here given that the storage handlers have been clearly communicated as internal in 5.0. It's worth it to find a clean setup for the future.

PR #7559, opened 2025-08-19
lab: New permission classes

bastianallgeier:
## Moved to notion
https://www.notion.so/getkirby/Permission-Classes-2ae59ef0a5cb80ad9e8fe4c490ecc326?v=2ae59ef0a5cb81fb9a98000c648995ef&source=copy_link


src/Toolkit/Reflection.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+namespace Kirby\Toolkit;
+
+use ReflectionMethod;
+use ReflectionParameter;
+
+class Reflection
```
lukasbestle: `Thought`: This class feels like a revival of our `Properties` class. A bit like magic. We need to reflect if we feel comfortable with it or if it may cause similar problems as `Properties` did.

src/Toolkit/Reflection.php
```diff
@@ -0,0 +1,69 @@
...
+	 * Returns all values in the $props array that are
+	 * defined as constructor arguments.
+	 */
+	public static function extractProps(array $props, string $class): array
+	{
+		$propsNames = static::paramsNames($class);
+
+		foreach ($props as $key => $value) {
+			if (in_array($key, $propsNames) === false) {
+				unset($props[$key]);
```
lukasbestle: `Todo`: Removing items during a loop can cause issues with the loop execution. I think we should use `array_filter()` instead.

src/Toolkit/Reflection.php
```diff
@@ -0,0 +1,69 @@
...
+	 * defined as constructor arguments.
+	 */
+	public static function extractAttrs(array $props, string $class): array
+	{
+		$propsNames = static::paramsNames($class);
+		$attrs      = [];
+
+		foreach ($props as $key => $value) {
+			if (in_array($key, $propsNames) === false) {
+				$attrs[$key] = $value;
```
lukasbestle: `Suggestion`: This then could use `array_diff(static::extractProps())`

src/Permissions/PanelPermissions.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Kirby\Permissions;
+
+class PanelPermissions extends ModelPermissions
+{
+	public function __construct(
+		public bool|null $access = null,
```
lukasbestle, line 8: `Todo`: We also need a way to define area permissions. As areas are dynamic, we won't be able to easily declare them using parameters/properties.

src/Permissions/LanguagePermissions.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Kirby\Permissions;
+
+class LanguagePermissions extends ModelPermissions
```
lukasbestle: `Todo`: We also had the idea to add `LanguageVariablesPermissions` in our permission improvement project in Notion. Might be a good opportunity to do it in this refactoring.

src/Permissions/ModelPermissions.php
```diff
@@ -0,0 +1,89 @@
...
+	{
+		if (is_bool($matrix) === true) {
+			return $matrix;
+		}
+
+		if (isset($matrix[$role]) === true) {
+			return $matrix[$role];
+		}
+
+		return $matrix['*'] ?? false;
```
lukasbestle: `Question`: Wasn't the default a permission of `true` if the blueprints and options don't define any restriction?

lukasbestle (commented):
It's great to think about the permission setup and to improve the structure of the classes. The current state is indeed very confusing and inflexible.

Some questions:

- The difference between the singular and plural options isn't entirely clear to me. E.g. there is both `file.create` and `files.create`. What is each permission supposed to do?
- Your example for `User::can()` only checks for model permissions (= ignores role permissions) when a model is passed. This would mean that the models themselves need to handle merging/overriding of the general role permissions, right? I see you added `ModelPermissions::merge()`. But where would we call this from?

General thought: Maybe we should complete the access permission feature (`lab/lukas/access-permissions` branch) together before we start the implementation of this refactoring. This would avoid loads of merge conflicts.

src/Toolkit/Reflection.php
distantnative: I had the same thought seeing this. Would be worth it to throw our combined ideas onto this how we could maybe do it without this class. Or: make it a much more specific class. As such a general toolkit class, it brings up flashbacks from Properties.

bastianallgeier:
@distantnative @lukasbestle instead of the "magic" Reflection toolkit class, I've added a way more specific constructor reflection class: https://github.com/getkirby/kirby/pull/7620

src/Permissions/LanguagePermissions.php
bastianallgeier: ☑️

distantnative:
@bastianallgeier what I cannot quite overlook yet but I think would be important to consider for us: does this help us to allow for more powerful permission setups (even if in a later step)? Or would this make our life harder to get there?

Think defining permissions a specific user for specific pages. Dynamic permissions depending on a field value or external API...

bastianallgeier:
@distantnative my first goal was to create really clean classes. I'm convinced that the promoted props in the constructors are perfect for permission definitions for any use case. I already have an idea how to solve the dynamic area permissions and this could also work for fields and other parts.

lukasbestle:
As I wrote in https://github.com/getkirby/kirby/pull/7620, I'm still torn if using reflections is a good solution in general, especially for such a security-relevant part as permissions. While the `Reflection\Constructor` class itself is quite simple, the logic that follows in the abstracts in this PR does more heavy lifting, defining the behavior at runtime. As everything is based on static method reflections, I guess we won't have to expect surprising behavior in the wild, however it still makes the behavior and logic complex and hard to understand, which can easily cause security issues.

I do get the advantages and that it allows to deduplicate the logic (only defining the arguments and not having to define the same list elsewhere). That does make it quite clean. I'm just not sure if it's the best way or if it will fall on our feet. Maybe there's a more standardized pattern we are not aware of.

lukasbestle:
@bastianallgeier My general questions in https://github.com/getkirby/kirby/pull/7559#pullrequestreview-3148813394 seem to have been lost in the PR history during your vacation.

bastianallgeier:
The main reason why we need the reflection and the logic in the from methods is a translation layer between our classes and yaml in blueprints. The second reason is dynamic permissions on runtime for custom areas and maybe later fields.

It's a plausible option to move entirely to class based permission settings instead of blueprints. That would give us and developers more control and type safety. But as long as we want to keep the blueprint options, I don't see a way around reflections. 

The reflections give us a couple of nice options:

A) we simply ignore all arguments that don't have a matching parameter (current solution) Example: typos in blueprints or hallucinated options.

B) we log invalid arguments to improve debugging 

C) we throw errors in debug mode and keep ignoring them in production. 

D) we use the unknown arguments as dynamic permissions (not yet implemented)

---

Another idea to keep that translation layer more visible for us could be to extract the additional logic in the from methods and move it to a dedicated BlueprintTranslator class (Name to be discussed :)) I partly did that with the RulesMatrix class already. This could help to keep the permission classes super clean and would also make it quite easy to rip out that layer later.

---

@lukasbestle I have your questions in my mental backlog :) 

When I started working on the Access permissions for Languages, Users, etc. I realized that it makes more sense semantically to also move the create permission there. I kept it in the model classes for now, because I'm not 100% yet how and if that transition will fully work. We still need to cover a case where we want to have different create permissions based on a page type for example. There could be a layered approach. General create permissions on the plural level + specific create permissions on the singular level. But that could also be quite confusing. Having the create permission on the model level would mean that the permissions need to be declared statically and that could also be a problem in the implementation. That's why I wanted to wait with the decision until I can start using the classes in real world scenarios.

The user class examples are very preliminary so far. 

In general, it would follow an onion approach. The default permissions come from the user role and get merged with more specific permissions from each model that needs to be worked on.

That's why we need the null (undecided) state. 

The permissions defined by the model have priority when not null, otherwise the default permissions from the role are used. 

Specific permissions for a model can be set for individual roles as well. Those would be evaluated for the current user as soon as permissions for a model are requested for a particular action. 

I hope I explained it well enough. Could be a good case for a flow chart :)

lukasbestle:
> It's a plausible option to move entirely to class based permission settings instead of blueprints. That would give us and developers more control and type safety.

Hm, TBH that confuses me. Even if we have permissions as class properties, don't we need a way for devs to set them? How would that work if not via role blueprints and options in model blueprints?

> The reflections give us a couple of nice options:
> A) we simply ignore all arguments that don't have a matching parameter (current solution) Example: typos in blueprints or hallucinated options.

With the current implementation in this PR, we only need reflections because the permissions are set on an existing instance. If we were to only use the constructor, we could just pass a splat array with whatever permissions in it and the constructor would automatically filter the non-existing arguments. So I think the reflections only become required for the other options.

> General create permissions on the plural level + specific create permissions on the singular level.

What I really don't understand is why there is a difference between singular and plural in general. How can a collection of models have a permission attached to it?

I think we should be careful to avoid overcomplicating it. I know there is always the balance between complexity and customizability/powerfulness, but I'd rather tend towards simplicity with everything security-related.

> The default permissions come from the user role and get merged with more specific permissions from each model that needs to be worked on. That's why we need the null (undecided) state. The permissions defined by the model have priority when not null, otherwise the default permissions from the role are used.

That makes sense. I also like the idea to have a single flow for this evaluation that brings it all together. I'm just not sure where in the code we would place it.

bastianallgeier:
Now I'm confused. Since when is it possible to use a splat operator and the unknown args get filtered out? I'm either completely wrong here and make my life a lot harder than it actually needs to be or as I remember it a class will always throw an error if you pass unknown args via splat operator.

bastianallgeier:
Ok, I just did another quick test and I'm right. You can't use the splat operator with unknown parameters. 

Screenshot 2025-10-12 at 16 18 45

bastianallgeier:
One of my goals/side-effects of with this class refactoring is that a Model can declare its permissions more flexibly in code: 

```php
class Page
{
  public function permissions(User $user): PagePermissions 
  {
    if ($user->id() === 'peter') {
      return new PagePermissions(
        changeTitle: true,
        delete: false
      );
    }

    return parent::permissions($user);
  }
}
```

We could totally think about having custom Role model classes as well, which then define their default permissions in code.

lukasbestle:
> You can't use the splat operator with unknown parameters.

Ah, I somehow got confused. Yes, then the only alternative to reflections is a static array definition that would duplicate the supported properties – hardly any kind of alternative.

PR #7563, opened 2025-08-20
fix: respect case preferences when grouping by callable

### LINKED ISSUE #7562: Case preferences are ignored when grouping collections by callable
## Description
I was trying to group my page collection by a custom callable:

```php
$kirby->collection('artists')->group(function (ArtistPage $artist) {
  return match ($artist->category()->value()) {
    'musician' => $artist->instrument()->value(),
    default => $artist->category()->value(),
  };
}, caseInsensitive: false);
```
Unfortunatly the value of the caseInsensitive argument is ignore whenever a callable is provided. As a consequence the name of the categories/instruments are lower case.

**Expected behavior**  
Keep the original case of the key used for each group intact, if `caseInsensitive` is set to `false`.

## To reproduce
1. Create a collection
2. Group the collection by a callable and set `caseInsensitive` to `false`

## Your setup
**Kirby Version**  
5.0.3

dweidner:
### 🐛 Bug fixes
Keep original case intact when grouping a collection by callable. See issue #7562.

### DIFF
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -553,7 +553,7 @@ class Collection extends Iterator implements Stringable
 				}
 			}
 
-			return new self($groups);
+			return new self($groups, !$caseInsensitive);
 		}
 
 		throw new Exception(
--- a/tests/Toolkit/CollectionTest.php
+++ b/tests/Toolkit/CollectionTest.php
@@ -276,7 +276,7 @@ class CollectionTest extends TestCase
 		$this->expectException(Exception::class);
 		$this->expectExceptionMessage('Invalid grouping value for key: a');
 
-		$collection->group(fn($item) => false);
+		$collection->group(fn ($item) => false);
 	}
 
 	public function testGroupArray(): void
@@ -286,7 +286,7 @@ class CollectionTest extends TestCase
 		$this->expectException(Exception::class);
 		$this->expectExceptionMessage('You cannot group by arrays or objects');
 
-		$collection->group(fn($item) => ['a' => 'b']);
+		$collection->group(fn ($item) => ['a' => 'b']);
 	}
 
 	public function testGroupObject(): void
@@ -296,7 +296,7 @@ class CollectionTest extends TestCase
 		$this->expectException(Exception::class);
 		$this->expectExceptionMessage('You cannot group by arrays or objects');
 
-		$collection->group(fn($item) => new Obj(['a' => 'b']));
+		$collection->group(fn ($item) => new Obj(['a' => 'b']));
 	}
 
 	public function testGroupStringObject(): void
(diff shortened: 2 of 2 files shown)


src/Toolkit/Collection.php
```diff
@@ -553,7 +553,7 @@ public function group(
 				}
 			}
 
-			return new self($groups);
+			return new self($groups, !$caseInsensitive);
```
afbora, line 556:
```suggestion
			return new self($groups, $caseInsensitive);
```

afbora: It should be without `!` to respect parent `$caseInsensitive`.

distantnative: @afbora no it's correct, the method and class use opposites: caseSensitive vs caseInsensitive

afbora: Oh sorry, I thought both are same variable names 😅

afbora:
@dweidner Can you add a unit test for this improvement, please? I can take care if you're not available.

dweidner:
Sorry, it took me a while to find the time. I will add some tests today. Thank you for considering my pull request.

dweidner:
Sorry for the noise. Was struggeling with the prettier configuration in VSCode. Should be ready now.

afbora (dismissed):
Great 👍 

_Optional note: IMHO It's not a big deal, but in unit tests, counting the results of `->group()` to check how many results the grouping explicitly returns increases the accuracy of the unit test._

dweidner:
> _Optional note: IMHO It's not a big deal, but in unit tests, counting the results of `->group()` to check how many results the grouping explicitly returns increases the accuracy of the unit test._

Was considering this as well, but was not sure whether this should be part of this particular unit test.

afbora:
What I mean is that the group size changes with sensitive and insensitive grouping. You do this with the key control. We can also make the unit test more meaningful by checking the group size. Something like that:

````php
public function testGroupByCallableCaseSensitive(): void
{
	$collection = new Collection();

	$collection->taylor = [
		'name' => 'Taylor',
		'genre' => 'Pop',
	];

	$collection->justin = [
		'name' => 'Justin',
		'genre' => 'Pop',
	];

	$collection->aubrey = [
		'name' => 'Aubrey',
		'genre' => 'Hip-Hop',
	];

	$collection->kayne = [
		'name' => 'kayne',
		'genre' => 'hip-hop',
	];

	$groupsCaseInsensitive = $collection->group(fn (array $item) => $item['genre'], true);

	$this->assertCount(2, $groupsCaseInsensitive);
	$this->assertTrue($groupsCaseInsensitive->has('pop'));
	$this->assertTrue($groupsCaseInsensitive->has('hip-hop'));


	$groupsCaseSensitive = $collection->group(fn (array $item) => $item['genre'], false);

	$this->assertCount(3, $groupsCaseSensitive);
	$this->assertTrue($groupsCaseSensitive->has('Pop'));
	$this->assertTrue($groupsCaseSensitive->has('Hip-Hop'));
}
````

dweidner:
That makes perfect sense. Thank you for clarifying.

afbora (approved):

PR #7565, opened 2025-08-23
refactor: Add #[\Override] attribute

afbora:
### For review team

### DIFF
--- a/src/Cache/ApcuCache.php
+++ b/src/Cache/ApcuCache.php
@@ -3,6 +3,7 @@
 namespace Kirby\Cache;
 
 use APCUIterator;
+use Override;
 
 /**
  * APCu Cache Driver
@@ -18,6 +19,7 @@ class ApcuCache extends Cache
 	/**
 	 * Returns whether the cache is ready to store values
 	 */
+	#[Override]
 	public function enabled(): bool
 	{
 		return apcu_enabled();
@@ -26,6 +28,7 @@ class ApcuCache extends Cache
 	/**
 	 * Determines if an item exists in the cache
 	 */
+	#[Override]
 	public function exists(string $key): bool
 	{
 		return apcu_exists($this->key($key));
@@ -35,6 +38,7 @@ class ApcuCache extends Cache
 	 * Flushes the entire cache and returns
 	 * whether the operation was successful
 	 */
+	#[Override]
 	public function flush(): bool
 	{
 		if (empty($this->options['prefix']) === false) {
@@ -48,6 +52,7 @@ class ApcuCache extends Cache
 	 * Removes an item from the cache and returns
 	 * whether the operation was successful
 	 */
+	#[Override]
 	public function remove(string $key): bool
 	{
 		return apcu_delete($this->key($key));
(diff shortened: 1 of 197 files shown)


distantnative:
@afbora I considered this as well when reviewing the PHP release notes. Honestly, I think we should not do it. It just bloats our code with too little benefit, in my opinion. It only becomes useful if we think a method would override an original method which actually isn't the case - but I think we'll run way too rarely into that situation that it justifies the amount of times it gets added here.

afbora:
@distantnative Thanks for the review.

On the contrary, I thought, and still do, that this would be very beneficial for documentation, IDEs, and code readability.

I think seeing which methods should be explicitly overridden will be very helpful when fixing bugs and developing code (both for the team and for developers who aren't familiar with the code). This shouldn't be viewed as bloat, but rather as a way to increase code readability and refactor safety.

distantnative:
> seeing which methods should be explicitly overridden

The attribute does the opposite: it doesn't indicate what methods should be overridden, but that the annotated method is supposed to override the method of a parent class.

This is why I think the usefulness is quite limited (not none). I can't quite see how it helps IDEs to be honest, neither users of Kirby. Where it has a use case is that if we alter the core code and remove a method in a parent class that we would not forget to remove it in a child class as well. 

In my opinion, that's too little for having to add an attribute to the majority of all our methods. But let's see what the others think - if the majority finds them useful, I won't try to veto this 😄

afbora:
No, AFAIK it does not specify which methods should be overridden, but it clearly indicates to the developer which methods are overridden. Or am I wrong about `#Override` attribute? 🙂

distantnative:
I can't quite follow you how it directly indicates which methods are overridden. As I wrote it's more the other way around. As per the PHP docs:

> This attribute used to indicate that a method is intended to override a method of a parent class

https://www.php.net/manual/en/class.override.php

As you did it with the changes: it goes on the method of the child class, not on the original method of the parent class.

lukasbestle:
I'm torn on this. It does make sense in cases where we absolutely need to make sure that the overridden implementation is not silently ignored (e.g. when we refactor the parent class and remove a method that was previously called for some processing). I'm sure we could find such cases and add the attribute there specifically. But I also wouldn't add it everywhere:

- We would need to remember to add it to all new code as well (which we will certainly forget sometimes).
- It would make refactorings both easier and more complex at the same time. Easier because we will immediately see what needs to be adapted, but in places where the override is not a necessity (and the child class can independently use the method), it would be a nuisance.
- I see the bloat argument. If it was useful everywhere, I think we could accept the added code, but with its limited effect the bloat becomes an issue.

To summarize: Maybe we can pinpoint the places where it really protects us and becomes useful. And only add it there.

afbora:
@lukasbestle @distantnative You're right, maybe using it everywhere is too much. So let's do this: Let's make this PR more smaller and specific. Let's focus on really critical methods:

- Authentication and security related methods
- Critical overrides for core functionality
- Places that definitely need to be protected during refactoring

This way we can see the benefits and also no "bloat" problem. If this approach works good, we can expand to other places later.
What do you think which methods/classes are in critical category? I can make a list and update the PR according to that.

distantnative:
If we could limit it to places where "this could be harmful if left forgotten when the parent method it overrides gets removed" - I think this would be a sweet spot then. 👍

afbora:
So I'm closing the PR to create new one in next days 👍

PR #7571, opened 2025-08-29
fix: layout snippet inside of snippet (#7567)

### LINKED ISSUE #7567: Layout snippets do not work from within other layout snippets
## Description

When I use a [layout snippet](https://getkirby.com/docs/guide/templates/snippets#passing-slots-to-snippets__layout-snippets) within another layout snipped, the content is discarded without any error reported.

**Expected behavior**  

I would expect to see the content of the included snippet.

**Note**

Since I found a workaround, the issue is not really important to me. However, I was a bit lost initially when I got an empty page without any error. I thought I should let you know anyway.

## To reproduce

1. Create a snippet
2. Create another snippet which uses the previous snipped as a layout snippet
3. Create a template which uses the snipped from step 2 as a layout snippet
4. **Content of the snipped from step 1 is missing**

**Example**

In the following example, I get a HTTP 200 response, but the page is completely empty.

https://github.com/JojOatXGME/plainkit/commit/5798fca5aa919f5160ed57964a22069f852542d0

<details>
<summary>Alternative: patch on top of plainkit (based on commit 3d651f772485fa7ec836238bde103d4368e9fdcb)</summary>

```diff
$ git diff --staged
diff --git a/site/snippets/base.php b/site/snippets/base.php
new file mode 100644
index 0000000..74e38ae
--- /dev/null
+++ b/site/snippets/base.php
@@ -0,0 +1 @@
+<h1><?= $page->title() ?></h1>
diff --git a/site/snippets/subsnippet.php b/site/snippets/subsnippet.php
new file mode 100644
index 0000000..b95c82b
--- /dev/null
+++ b/site/snippets/subsnippet.php
@@ -0,0 +1 @@
+<?php snippet('base', slots: true);
diff --git a/site/templates/default.php b/site/templates/default.php
index 74e38ae..9ca23b8 100644
--- a/site/templates/default.php
+++ b/site/templates/default.php
@@ -1 +1 @@
-<h1><?= $page->title() ?></h1>
+<?php snippet('subsnippet', slots: true);
```

</details>

## Your setup
**Kirby Version**  
4.8.0

**Console output**  

```
> Composer\Config::disableProcessTimeout
> @php -S localhost:8000 kirby/router.php
[Mon Aug 25 01:15:56 2025] PHP 8.4.11 Development Server (http://localhost:8000) started
[Mon Aug 25 01:16:02 2025] [::1]:54971 Accepted
[Mon Aug 25 01:16:02 2025] [::1]:54971 Closing
```

 - Device: Desktop
 - OS: Windows 10
 - Browser: Firefox
 - Version: 142.0

**Workaround**  

Call `endsnippet()` at the end of either the snippet of step 2, or the template of step 3.

JojOatXGME:
## Description

Fixes #7567. Implement handling of [layout snippets](https://getkirby.com/docs/guide/templates/snippets#passing-slots-to-snippets__layout-snippets) within snippets.

Extract implementation from `Template::render` into `Snippet::load`. `Snippet::load` overrides `Tpl::load`. The calls to `static::load` inside `Snippet::render` and `Snippet::factory` are now referencing `Snippet::load` instead of `Tpl::load`. As a result, these two methods are now correctly handling layout snippets.

### 🐛 Bug fixes

- Fix layout snippet used within another snippet (#7567)

## Docs

- https://getkirby.com/docs/guide/templates/snippets#passing-slots-to-snippets__layout-snippets

### For review team

### DIFF
--- a/src/Template/Snippet.php
+++ b/src/Template/Snippet.php
@@ -135,15 +135,30 @@ class Snippet extends Tpl
 	}
 
 	/**
-	 * Closes any potentially open layout snippet.
+	 * Renders the given template.
+	 * This method supports the use of layout snippets inside the template.
 	 *
-	 * @param Snippet|null $outsideSnippet The expected value for `Snippet::$current`.
-	 * @param string|null $template The content of the currently processed template or snippet.
-	 * @param array $data Additional data provided to any potential layout snippet.
-	 * @return string|null The new adjusted content of the template if `$template` was given.
+	 * @param string|null $file Path to the template file.
+	 * @param array $data Data available inside the template.
+	 * @param array $layoutData Data available inside any potential layout snippet.
+	 * @return string The rendered content of the given template file.
+	 * @throws Throwable
+	 *
+	 * @see Tpl::load() Base implementation without support for layout snippets.
 	 */
-	public static function endlayout(Snippet|null $outsideSnippet, string|null $template = null, array $data = []): string|null
-	{
+	public static function load(
+		string|null $file = null,
+		array $data = [],
+		array $layoutData = [],
+	): string {
+		// if the template is rendered inside a snippet,
+		// we need to keep the "outside" snippet object
+		// to compare it later
+		$outsideSnippet = Snippet::$current;
+
+		// load the template
+		$template = parent::load($file, $data);
+
 		// if last `endsnippet()` inside the current template
 		// has been omitted (= snippet was used as layout snippet),
 		// `Snippet::$current` will point to a snippet that was
@@ -159,21 +174,16 @@ class Snippet extends Tpl
 			return $template;
 		}
 
-		if (!isset($template)) {
-			// let the snippet close and render natively
-			echo static::$current?->render($data);
-			return null;
-		}
 		if (Snippet::$current->slots()->count() === 0) {
 			// no slots have been defined, but the template code
 			// should be used as default slot
-			return Snippet::$current->render($data, [
+			return Snippet::$current->render($layoutData, [
 				'default' => $template
 			]);
 		}
 		// swallow any "unslotted" content
 		// between start and end
-		return Snippet::$current->render($data);
+		return Snippet::$current->render($layoutData);
 	}
 
 	/**
@@ -289,16 +299,8 @@ class Snippet extends Tpl
 		// custom data overrides for the data that was passed to the snippet instance
 		$data = array_replace_recursive($this->data, $data);
 
-		// if the template is rendered inside a snippet,
-		// we need to keep the "outside" snippet object
-		// to compare it later
-		$snippet = Snippet::$current;
-
 		// load the template representing this snippet
-		$template = static::load($this->file, static::scope($data, $this->slots()));
-
-		// handle any potentially open layout snippet
-		return static::endlayout($snippet, $template);
+		return static::load($this->file, static::scope($data, $this->slots()));
 	}
 
 	/**
--- a/src/Template/Snippet.php
+++ b/src/Template/Snippet.php
@@ -135,6 +135,48 @@ class Snippet extends Tpl
 	}
 
 	/**
+	 * Closes any potentially open layout snippet.
+	 *
+	 * @param Snippet|null $outsideSnippet The expected value for `Snippet::$current`.
+	 * @param string|null $template The content of the currently processed template or snippet.
+	 * @param array $data Additional data provided to any potential layout snippet.
+	 * @return string|null The new adjusted content of the template if `$template` was given.
+	 */
+	public static function endlayout(Snippet|null $outsideSnippet, string|null $template = null, array $data = []): string|null
+	{
+		// if last `endsnippet()` inside the current template
+		// has been omitted (= snippet was used as layout snippet),
+		// `Snippet::$current` will point to a snippet that was
+		// opened inside the template; if that snippet is the direct
+		// child of the snippet that was open before the template was
+		// rendered (which could be `null` if no snippet was open),
+		// take the buffer output from the template as default slot
+		// and render the snippet as final template output
+		if (
+			Snippet::$current === null ||
+			Snippet::$current->parent() !== $outsideSnippet
+		) {
+			return $template;
+		}
+
+		if (!isset($template)) {
+			// let the snippet close and render natively
+			echo static::$current?->render($data);
+			return null;
+		}
+		if (Snippet::$current->slots()->count() === 0) {
+			// no slots have been defined, but the template code
+			// should be used as default slot
+			return Snippet::$current->render($data, [
+				'default' => $template
+			]);
+		}
+		// swallow any "unslotted" content
+		// between start and end
+		return Snippet::$current->render($data);
+	}
+
+	/**
 	 * Closes the last openend slot
 	 */
 	public function endslot(): void
@@ -247,7 +289,16 @@ class Snippet extends Tpl
 		// custom data overrides for the data that was passed to the snippet instance
 		$data = array_replace_recursive($this->data, $data);
 
-		return static::load($this->file, static::scope($data, $this->slots()));
+		// if the template is rendered inside a snippet,
+		// we need to keep the "outside" snippet object
+		// to compare it later
+		$snippet = Snippet::$current;
+
+		// load the template representing this snippet
+		$template = static::load($this->file, static::scope($data, $this->slots()));
+
+		// handle any potentially open layout snippet
+		return static::endlayout($snippet, $template);
 	}
 
 	/**
--- a/src/Template/Template.php
+++ b/src/Template/Template.php
@@ -5,7 +5,6 @@ namespace Kirby\Template;
 use Exception;
 use Kirby\Cms\App;
 use Kirby\Filesystem\F;
-use Kirby\Toolkit\Tpl;
 use Stringable;
 
 /**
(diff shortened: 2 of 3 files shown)


src/Template/Template.php
```diff
@@ -163,31 +163,8 @@ public function render(array $data = []): string
...
-		if (Snippet::$current->slots()->count() === 0) {
-			return Snippet::$current->render($data, [
-				'default' => $template
-			]);
-		}
-
-		// let the snippet close and render natively
-		return Snippet::$current->render($data);
+		// handle any potentially open layout snippet
+		return Snippet::endlayout($snippet, $template, $data);
```
JojOatXGME: Passing the `$data` from the template to the layout snippet is unexpected for me. This special case for layout snippets (in contrast to regular snippets) does not seem to be documented. Is this a bug? I think the `$data` from the template even overwrites data explicitly passed to the snippet.

I kept this behavior for backward compatibility. If this is a bug, we might consider removing or deprecating this behavior. Otherwise, I should probably forward `$data` in `Snippet::render` as well.

**EDIT:** I think forwarding data and slots to layout snippets might actually be a nice feature. This would match the behavior of *extending templates* I have seen in other template engines. I actually thought about creating a feature request for that a few days ago. However, in this case, we should also forward slots, not just data.

src/Template/Snippet.php
```diff
@@ -134,6 +134,48 @@ public static function end(): void
...
 
+	/**
+	 * Closes any potentially open layout snippet.
+	 *
+	 * @param Snippet|null $outsideSnippet The expected value for `Snippet::$current`.
+	 * @param string|null $template The content of the currently processed template or snippet.
+	 * @param array $data Additional data provided to any potential layout snippet.
+	 * @return string|null The new adjusted content of the template if `$template` was given.
+	 */
+	public static function endlayout(Snippet|null $outsideSnippet, string|null $template = null, array $data = []): string|null
```
JojOatXGME: Note that this is a new public method exposed to all users. Not sure if we should hide it somehow.

src/Template/Template.php
JojOatXGME: Note that any output from before opening the layout snippet gets discarded. This was already the case before. I think ideally, we should call `Snippet::endlayout` before `Tpl::load` calls `ob_get_contents()`. However, I avoided to call `Snippet::endlayout` from within `Tpl::load` because I assumed dependencies from `\Kirby\Toolkit` to `\Kirby\Template` are to be avoided. Alternatively, it might make sense to trigger an exception if there is any output before opening a layout snippet.

src/Template/Snippet.php
distantnative: I find the naming a bit add as we do not have any other layout-related method. There is no corresponding `beginLayout` or so which makes understanding the role of this method in the general picture a bit hard IMO. Maybe we can find a name that explains it a bit better?

distantnative:
@JojOatXGME Could you please provide the setup that this PR tries to solve. I have to admit I couldn't quite follow the description in the issue and therefore aren't sure if this is fixing an unwanted behavior or enabling an unwanted usage 😅

src/Template/Snippet.php
JojOatXGME: Not sure. I also thought about wrapping `Tpl::load`.

```php
/**
 * Load template with support for layout snippets.
 * This method mostly equivalent to `Tpl::load`, except that it supports the use of layout snippets.
 */
public static function loadTemplate(string|null $file = null, array $templateData = [], array $snippetData = []): string|null
{
	...
}
```

We could then also simplify the call side. The following code inside `Template::render` and `Snippet::render`

```php
// if the template is rendered inside a snippet,
// we need to keep the "outside" snippet object
// to compare it later
$snippet = Snippet::$current;

// load the template
$template = Tpl::load($this->file(), $data);

// handle any potentially open layout snippet
return Snippet::endlayout($snippet, $template, $data);
```

could then be replaced by this:

```php
$template = Snippet::loadTemplate($this->file(), $data, $data);
```

JojOatXGME:
> Could you please provide the setup that this PR tries to solve.

@distantnative I just created a fork of `plainkit` and made some adjustment to reproduce the issue. Does this help?

https://github.com/JojOatXGME/plainkit/commit/5798fca5aa919f5160ed57964a22069f852542d0

If you check out commit `5798fca5aa919f5160ed57964a22069f852542d0` and run the project, the resulting page will be empty. However, I would expect to see the content in the `default` template.

src/Template/Snippet.php
JojOatXGME: I update the implementation in fd22bcb4d94df716fc3fd826b1382129161f5018. I am now overriding (and wrapping) `Tpl::load` with `Snippet::load`. This also fixed a bug since I previously forgot to adjust `Snippet::factory`.

afbora:
@JojOatXGME I've been trying to understand the PR for 15 minutes, but I don't get it, sorry 🙂 I just see that the Template::render method has been moved to Snippet::load. Am I missing something? Does this PR actually solve the problem?

distantnative:
I think this might be the better/right direction, having the handling in `Snippet::load()` after every time a snippet file was actually loaded. I'll try to give it a test myself tomorrow, we might have to adapt some of the comments to make sense now.

The Template render call to snippet load reads a little weird - but might be full functional. Just semantically made me stumble.

JojOatXGME:
> I just see that the `Template::render` method has been moved to `Snippet::load`.

@afbora `Snippet::render` and `Snippet::factory` use `static::load`. They have previously used `Tpl::load`, but are now using `Snippet::load`. As a result, these two methods are now correctly handling layout snippets.

**EDIT:** I also now updated the description in the PR accordingly.

src/Template/Snippet.php
```diff
@@ -134,6 +134,58 @@ public static function end(): void
...
+	 * @param array $layoutData Data available inside any potential layout snippet.
+	 * @return string The rendered content of the given template file.
+	 * @throws Throwable
+	 *
+	 * @see Tpl::load() Base implementation without support for layout snippets.
+	 */
+	public static function load(
+		string|null $file = null,
+		array $data = [],
+		array $layoutData = [],
```
distantnative, line 152: `Question:` Where do you see the need to split this up into two parameters? Instead just having one `$data` parameter.

distantnative (commented):
We will need very thorough testing. These nested snippets and slots pose a lot of edge cases that I am afraid our unit tests do not fully cover yet. Before we merge, we should be very certain this doesn't add a regression for other use cases.

We also need to update the DocBlock and comments of the `Snippet::load()` method. It's misleading that we talk about templates win the context of snippets. I would also love if we could remove the term layout snippet (I know, not introduced but this PR but let's take the chance). Especially in this use case of nesting, the idea of layout doesn't make a lot of sense to me. I think in the end it's about snippets with slots where the closing `endsnippet()` call has been omitted and Kirby should end it implicitly at the end of a snippet file.

src/Template/Snippet.php
JojOatXGME: Currently, `Template::render` sets `$layoutData` to the same value as `$data`, but `Snippet::render` keeps the parameter empty. I only set `$layoutData` to the same value as `$data` inside `Template::render` for backward compatibility, but it looks to me like it actually might have been a bug. (At least the resulting deviation from normal snippets seems not to be documented.)

Within `Snippet::render`, I cannot set `$layoutData` to the same value as `$data` because this would lead to a runtime exception. The array for `$layoutData` currently must not contain `'slot'` nor `'slots'`. However, the `$data` given by `Snippet::render` does contain `'slot'` and `'slots'`.

I feel like the currently intended behavior is to always keep `$layoutData` empty. In other words, I think forwarding `$data` from the template to the snippet was actually another minor bug in the existing code, which I just preserved. If my understanding is correct, I could remove this parameter and always use an empty array instead.

However, if I look at the behavior not from the perspective of the bug fix, I would actually prefer if layout snippets would forward the data from the parent. I actually considered creating a feature request for that one or two weeks ago. This would match the behavior of *“Template Inheritance”* in other popular template engines like [Twig](https://twig.symfony.com/doc/3.x/templates.html#template-inheritance), [Laravel's Blade Templates](https://laravel.com/docs/12.x/blade#layouts-using-template-inheritance), [Jinja](https://jinja.palletsprojects.com/en/stable/templates/#template-inheritance), and [Django's template language](https://docs.djangoproject.com/en/5.2/ref/templates/language/#template-inheritance). However, I would consider this a new feature and therefore did not implement this behavior. To make this feature work, I would always set `$layoutData` to the same value as `$data`, implement proper merging of `slot` data, and tweak the precedence of data coming from this parameter vs the data explicitly provided from within the template. (The explicit content from the template should maybe take precedence for data, but not for slots.) While I would be happy to implement this behavior, I did not consider it part of this ticket. Also note that full “Template Inheritance” usually also includes a feature to call `super()` or `@@parent` to preserve the content from the extended template. This functionality would still not be covered by the implementation I have just described.

distantnative:
@JojOatXGME To give you an update where we are standing: The snippet nesting code is rather complex to track mentally with the levels it opens up and thus potential edge cases. Which is why we are a bit worried to simply merge this. I have started to set up more test scenarios (e.g. https://github.com/getkirby/sandbox/pull/21) to see if all other use cases are not broken accidentally by this PR. I will also try to add similar ones as unit tests.

Also changed the base for this PR as we will rather include this in a minor, but not patch release.

distantnative:
Had to recreate it as PR in our own repo: https://github.com/getkirby/kirby/pull/7599

PR #7580, opened 2025-09-04
fix: `Exif::iso()` supports array

### LINKED ISSUE #7569: ISOSpeedRatings in Kirby\Image\Exif can be an array
When setting the $iso property in Kirby\Image\Exif that data can be an array in some instances, not a string. This throws an error.

I ran into this problem today batch importing images. It was [posted previously on the forum](https://forum.getkirby.com/t/after-adding-multiple-images-kirby-throws-error-in-exif-php/34032). The proposed solution works.

Using Kirby 5.0.4

distantnative:
### 🐛 Bug fixes

- `Exif` class now supports arrays for `ISOSpeedRatings `
#7569

### For review team

### DIFF
--- a/src/Image/Exif.php
+++ b/src/Image/Exif.php
@@ -25,7 +25,7 @@ class Exif
 	protected string|null $exposure = null;
 	protected string|null $focalLength = null;
 	protected bool|null $isColor = null;
-	protected string|null $iso = null;
+	protected string|array|null $iso = null;
 	protected Location|null $location = null;
 	protected string|null $timestamp = null;
 	protected int $orientation;
@@ -94,7 +94,7 @@ class Exif
 	/**
 	 * Returns the iso value
 	 */
-	public function iso(): string|null
+	public function iso(): string|array|null
 	{
 		return $this->iso;
 	}


afbora (approved):
I think we should mention this because it involves a breaking change, although it is rare.

distantnative:
I don't think it is a breaking change technically. As those cases when the code now would return an array would fully crash beforehand - so there shouldn't be any code/setup that worked before but now breaks due to this change.

afbora:
@distantnative I know, so what happens if there are developers who use `->iso()` in their application and expects a `string`?

distantnative:
@afbora It would run into an error if `ISOSpeedRatings` is an array. But it would already run into an error before this PR as well. Just a different one. That's why I don't think this is a breaking change - it doesn't break any existing setups that aren't already broken in the first place.

afbora:
@distantnative That's right 👍 🙂

lukasbestle:
Not sure TBH. With the old code, an image with an ISO array triggered a PHP fatal error because of the wrong return type. Now it suddenly depends on what the calling code does with the return value. It could fail much more subtly and in a way that is very hard to spot.

Idea:

- Given that the feature does not seem to be used often, we could implement an intermediary solution for v5 that returns the first array item, unbreaking the current fatal error.
- In v6 we could then change it to the implementation in this PR, thus supporting the full feature set.

afbora:
If the developer runs it inside a try-and-catch and it behaves differently inside the catch, that's also a breaking change. But I think this is a one-in-a-million case 🙂 I agree with Nico and would prefer it be considered a fixed bug.

lukasbestle:
But what speaks against the two-step fix to be on the safe side?

afbora:
@lukasbestle Although it is very rare, we can fix it in just two steps :) I am still OK with the two step solution to avoid taking risks 👍

bastianallgeier:
For the two-step fix, we could change the base for this PR to v6/develop and create another one to return the first array item.

bastianallgeier:
I've created a second PR for the non-breaking fix: https://github.com/getkirby/kirby/pull/7585

PR #7581, opened 2025-09-05
feat: Replace `pointer-events: none` with logic

### LINKED ISSUE #7493: `pointer-events: none` on disabled fields/inputs
We are using this to prevent some interaction with UI elements. However, it also prevents desirable interactions (e.g. being able to select the value of a disabled field).

I think we should remove the general `pointer-events: none`  on all disabled fields/inputs but rather define very precisely if interaction with certain field elements needs to be prevented.

distantnative:
## Description

Replacing the general `pointer-events: none` for disabled fields/inputs with very specific rules per field and adding the necessary logic to continue not allowing certain actions when field is disabled.

### ✨ Enhancements

- Values of disabled Panel fields can be selected 
#7493

### For review team

### DIFF
--- a/panel/src/components/Forms/Blocks/Blocks.vue
+++ b/panel/src/components/Forms/Blocks/Blocks.vue
@@ -57,7 +57,11 @@
 			</k-draggable>
 
 			<!-- No blocks -->
-			<k-empty class="k-blocks-empty" icon="box" @click="choose(blocks.length)">
+			<k-empty
+				class="k-blocks-empty"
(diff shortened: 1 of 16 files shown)


bastianallgeier (commented):
I went through all the fields in the lab and the disabled states worked great. There's one tiny exception in the tags field. The plus button still shows up …

Image

bastianallgeier:
I've fixed the appearing button

bastianallgeier (approved):

Issue #7243, opened 2025-05-25
System view: Plugin link to plugins.getkirby.com


distantnative:
If listed in plugins.getkirby.com, it could be nice to have the plugin link point there.

afbora:
<img width="1551" height="175" alt="Image" src="https://github.com/user-attachments/assets/49d02b59-9b23-46fa-b455-44bc206720d5" />

Technically, this is a very simple development, but the most challenging part of this development is the question of where to provide this link 🙂 

Possible ideas:

- A new column: Contains the link text and link. It can provide a link for each plugin (View in plugins directory) or be blank.
- A new column: Plugins directory status. For each plugin, it will indicate whether it is exists in the plugins directory. If so, a link will be provided. For ex: Not exits string with cross icon in red color and Exists with check icon in green color.
- An additional link text in the Version dropdown: View in plugins directory
- A link can be provided by placing an icon in the Name cell at the far right.

What kind of UI do you think it should be?

distantnative:
I don't think we should add another UI element. I think that'd be too much - then having two links to two different places. My idea was to use it instead of the current link that goes e.g. to GitHub. 

But I am not fully sure whether we should do this. Or if the current links are the better option.

afbora:
> My idea was to use it instead of the current link that goes e.g. to GitHub.

It makes sense to use the link on the name. If it's listed in the directory, we can provide a directory link on plugin name, if not, provide a GitHub link.

bastianallgeier:
I think it's a good idea to prioritize the plugin directory here. This also makes us more independent from Github in the long run.

distantnative:
Do we have a good way to know in the backend what plugin is listed in the plugin directory? I was thinking maybe with he update status, but that can also be disabled via config option etc.

I guess what we could do is
- Add a route to the plugin directory that captures the plugin namespace and name
- Try to find the matching plugin in the directory, if successful redirect to that page
- If not, we also send `$plugin->link()` (what we currently use) as query parameter and the directory redirects to that instead

`Plugin::link()` could be rewritten as

```php
public function link(): string|null
{
	$info   = $this->info();
	$link   = $info['homepage'] ?? null;
	$link ??= $info['support']['docs'] ?? null;
	$link ??= $info['support']['source'] ?? null;

	return 'https://plugins.getkirby.com/find/' . $this->name() . '?link=' . (V::url($link) ? $link : null);
}
````

(+ some url encoding I guess)

What do you think?
[...]

PR #7582, opened 2025-09-05
feat: Plugin links point to plugins.getkirby.com

distantnative:
### ✨ Enhancements

- Plugin links point to `plugins.getkirby.com` directory if plugin is listed there
#7243

### For review team

### DIFF
--- a/src/Panel/Controller/View/SystemViewController.php
+++ b/src/Panel/Controller/View/SystemViewController.php
@@ -84,9 +84,9 @@ class SystemViewController extends ViewController
 	public function plugins(): array
 	{
 		return $this->plugins ??= $this->system->plugins()->values(function (Plugin $plugin) {
-			$authors   = $plugin->authorsNames();
-			$update    = $plugin->updateStatus();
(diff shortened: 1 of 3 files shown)


afbora (approved):
Great! So why don't we move it in 5.1 or 5.1.x?

distantnative:
@afbora I think we shouldn't add larger things to 5.1.0 anymore as it is already in RC phase, I would only fix regressions or small patch fixes. Same for 5.1.x releases: they're supposed to be patch releases, so in my view only small direct fixes should go in there. And enhancements like this rather should go at least into minor releases.

Why I went here for v6 is mainly to avoid merging conflicts due to the view controller refactoring in v6. One could do the extra work to backport it to 5.2 and then really pay attention when merging to resolve the conflict properly. I didn't see the benefit of this PR as big enough to do this.

distantnative:
I realized that in the end the code in the view controllers was irrelevant. Created a back ported PR for v5.2 https://github.com/getkirby/kirby/pull/7584

PR #7589, opened 2025-09-13
fix: Disable form controls while processing

### LINKED ISSUE #7586: Save button UX
Hello Kirby team,

In Kirby 5 save button is clickable during save process, which is pretty irritating on larger pages which takes more time to save, because user don't know what is going on, is it saving or not. The only signal that something is happening is a small circle instead of home icon at site title, which is not visible if user is lower in the page. So most users keep clicking the save button thinking it will speed up the save process. 

In Kirby 4 save button was nicely disabled during save process, so we knew save process is happening. 

Can you do something to improve the save process UX, please?

distantnative:
### 🐛 Bug fixes

- Panel form controls are disabled again while saving
#7586

### For review team

### DIFF
--- a/panel/lab/components/formcontrols/index.vue
+++ b/panel/lab/components/formcontrols/index.vue
@@ -58,6 +58,9 @@
 				@submit="log('submit')"
 			/>
 		</k-lab-example>
+		<k-lab-example label="Saving">
+			<k-form-controls :has-diff="true" :is-processing="true" />
+		</k-lab-example>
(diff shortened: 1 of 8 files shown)


bastianallgeier (commented):
We kept the rest of the form controls stupid and I think I would do that for the processing state as well. It's a bit more work, but I would add an is-processing prop and provide the state from outside.

bastianallgeier (approved):

PR #7590, opened 2025-09-13
refact: Support button props for `k-dropdown-item`

distantnative:
## Description

`k-dropdown-item` so far has not supported `dialog`/`drawer` props. Instead, one had to take the route via the `click` handler. However, in many backend-driven scenarios (e.g. view buttons) this is not an option.

### ✨ Enhancements

- `k-dropdown-item` supports all button props (esp. `dialog` and `drawer`)

### ♻️ Refactored

- `k-button` supports dialog/drawer objects as props
- `$panel.dialog.open()`/`$panel.drawer.open()` accept a single object argument if it has a `url` key

### For review team

### DIFF
--- a/panel/src/panel/dialog.js
+++ b/panel/src/panel/dialog.js
@@ -1,5 +1,6 @@
-import { reactive } from "vue";
 import Modal, { defaults as modalDefaults } from "./modal.js";
+import { isObject } from "@/helpers/object.js";
+import { reactive } from "vue";
 
 export const defaults = () => {
 	return {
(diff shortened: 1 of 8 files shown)


panel/src/panel/dialog.js
```diff
@@ -88,6 +89,13 @@ export default (panel) => {
 				return this.openComponent(dialog);
 			}
 
+			// handle drawer object with url property
+			if (isObject(dialog) && dialog.url) {
+				options = dialog;
+				dialog = dialog.url;
+				delete options.url;
+			}
```
distantnative, line 97: `Explanation:` This sadly isn't fully DRY but repeated for the drawer as well. However, we can not simply move it to `modal.open()` as then the URL prefixing code below, which is different for dialogs and drawers, would not be taken into account.

bastianallgeier (approved):
Even though that one part might not be super dry, the rest is so much cleaner now. I love it.

PR #7592, opened 2025-09-14
fix: Add CSS class for page status button

distantnative:
## Description

Working on the Hub, I realized that this button is missing a specific class, making it hard to e.g. hide it via CSS.

### ✨ Enhancements

- Page status button in sections received a `.k-page-status-icon-option` CSS class

### For review team

### DIFF
--- a/panel/src/components/Sections/PagesSection.vue
+++ b/panel/src/components/Sections/PagesSection.vue
@@ -19,6 +19,7 @@ export default {
 						page.status,
 						page.permissions.changeStatus === false
 					),
+					class: "k-page-status-button",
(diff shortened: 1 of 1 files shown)


bastianallgeier:
Could that collide with the k-page-status-button class in the view button?

distantnative:
Maybe `.k-page-status-icon-option` instead?

bastianallgeier:
It's mostly internal so far, but we could also reuse the "flag" term here. `.k-page-status-flag`

distantnative:
We don't use the flag term anywhere else and actually started migration away form it in the JS code, rather treating it as any other item buttons.

bastianallgeier (approved):

PR #7599, opened 2025-09-21
fix: Unclosed nested snippets

### LINKED ISSUE #7567: Layout snippets do not work from within other layout snippets

distantnative:
## Description

Original PR: https://github.com/getkirby/kirby/pull/7571

- [x] Unit tests

### 🐛 Bug fixes

- Fixed auto-closing open snippets at the end of a nested snippet (thx to @JojOatXGME)
https://github.com/getkirby/kirby/issues/7567

### Breaking changes
- Template data must not include variables named `$slot` or `$slots`

### For review team

  - [x] https://github.com/getkirby/sandbox/pull/21


tests/Template/TemplateTest.php
```diff
@@ -102,8 +102,8 @@ public function testRender(): void
 			]
 		]);
 
-		$template = new Template('simple');
-		$this->assertSame('Test', $template->render(['slot' => 'Test']));
+		$template = new Template('template');
+		$this->assertSame('Test', $template->render(['foo' => 'Test']));
```
distantnative, line 106: `Discuss:` As we reuse `Snippet::load()` for `Template::render()`, templates now would also not allow to include `$slot` and `$slots` in the data array passed to the template. This is a breaking change, but maybe more consistent?

src/Template/Snippet.php
```diff
@@ -128,6 +128,57 @@ public static function end(): void
...
+			// no slots have been defined, but the snippet code
+			// should be used as default slot
+			return Snippet::$current->render($data, [
+				'default' => $snippet
+			]);
+		}
+
+		// swallow any "unslotted" content
+		// between start and end
+		return Snippet::$current->render($data);
```
distantnative, line 179: `Help:` I cannot find a unit test to cover this line. I am wondering if there is a logical mistake and this line is never reachable.

bastianallgeier: I'm not 100% sure, but I think this was about the following case: 

The layout snippet has a head and default slot: 

```php
<head>
<?= $slots->head() ?>
</head>

<body>
<?= $slot ?>
</body>
```

And the snippet is used like this: 

```php
<?php snippet('layout', slots: true) ?>

<?php slot('head') ?>
This goes into the head
<?php endslot() ?>

This content is swallowed. As soon as a named slot is used, the default slot also has to be added explicitely.

<?php endsnippet() ?>
```

tests/Template/TemplateTest.php
bastianallgeier: I'm in general fine with this breaking change. Moving it to v6 does not sound like a bad idea to me anyway. It's quite a bit of a change in the template engine.

src/Template/Snippet.php
bastianallgeier: I added unit tests for it and they run through, but the line is still not marked as covered. If I remove the line from the class, my two new tests fail though. I'm pretty sure this is a Codecov issue.

distantnative: OK thanks, then let's ignore it. But move this to v6 as discussed in the other comment.

bastianallgeier (approved):

PR #7600, opened 2025-09-21
fix: Add Kirby query `users` entry point

### LINKED ISSUE #7597: query: users is broken
## Description
I'm in the process of upgrading a website from Kirby 3 to Kirby 5. I encountered a `Invalid query result data: NULL` error message and when digging into it i found out that the origin is `query: users` in a select field:

```yml
editor:
  label: Editor
  type: select
  options:
    type: query
    query: users
```

What does work is `query: kirby.users`. I tested `query: users` in recent kirby versions and it turns out that it broke a long time ago. 5.0.0, 4.8.0, 4.6.0, 4.3.0, 4.1.1, 3.10.0.1, 3.9.6 – this query breaks all of them. So the regression must have been introduced somewhere in Kirby 3.x. I wrote that query on a cold 9th of November in 2020. I was using Kirby 3.4.4 at that time.

**Expected behavior**  
The select should let me pick a panel user [as its shown in the docs](https://getkirby.com/docs/reference/panel/fields/select#dynamic-options):

Image

By the way – respect for your hard work on the Kirby issues!

distantnative:
### 🐛 Bug fixes

- Added back `users` as entry point for Kirby queries
#7597

### For review team

### DIFF
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -9,6 +9,7 @@ use Kirby\Cms\File;
 use Kirby\Cms\Page;
 use Kirby\Cms\Site;
 use Kirby\Cms\User;
+use Kirby\Cms\Users;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Image\QrCode;
 use Kirby\Query\Runners\Runner;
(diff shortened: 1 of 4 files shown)


afbora (approved):
@distantnative Sorry, I couldn't add related unit tests. I didn't understand why the coverage still not passed 🙈 🙂

bastianallgeier:
Maybe our static definition of query entries is messing up code coverage here. I would ignore it.

PR #7608, opened 2025-09-28
(Fields PR 2) refact!: `FieldClass` with named props

distantnative:
### Todo

- [x] Merge first https://github.com/getkirby/kirby/pull/7679
- [x] Changelog

### Next

- [ ] List all constructor parameters instead `...$props`
  - [ ] https://github.com/getkirby/kirby/pull/7684
  - [ ] https://github.com/getkirby/kirby/pull/7685

### ✨ Enhancements

- New `FieldClass::validate()` method that runs all validations. 
- New `FieldClass::factory()` method, which is used in Fields to create instances from arrays
- New `Kirby\Form\Mixin\Siblings` trait

### ♻️ Refactored

- Use named props instead of arrays in all Field classes

### 🚨 Breaking changes

- `FieldClass::setSiblings()` is now public and siblings are injected after the construction. Setters can no longer rely on the siblings collection. Use getters or overwrite the `::setModel` method if necessary. 
- `FieldClass::setModel()` is now public and the model is injected after the construction. Setters can no longer rely on the model. Use getters or overwrite the `::setModel` method if necessary. 
- Field classes no longer define the value in the constructor. Use `FIeldClass::fill()` instead to provide an initial value. 

### For review team

### DIFF
--- a/src/Form/Field/StatsField.php
+++ b/src/Form/Field/StatsField.php
@@ -20,7 +20,7 @@ class StatsField extends FieldClass
 	/**
 	 * Array or query string for reports. Each report needs a `label` and `value` and can have additional `info`, `link`, `icon` and `theme` settings.
 	 */
-	protected array|string $reports;
+	protected array|string|null $reports;
 
 	/**
 	 * The size of the report cards. Available sizes: `tiny`, `small`, `medium`, `large`
@@ -32,12 +32,15 @@ class StatsField extends FieldClass
 	 */
 	protected Stats $stats;
 
-	public function __construct(array $params)
-	{
-		parent::__construct($params);
+	public function __construct(
+		array|string|null $reports = null,
+		string|null $size = null,
+		...$props
+	) {
+		parent::__construct(...$props);
 
-		$this->reports = $params['reports'] ?? [];
-		$this->size    = $params['size']    ?? null;
+		$this->reports = $reports;
+		$this->size    = $size;
 	}
 
 	public function hasValue(): bool
@@ -59,7 +62,7 @@ class StatsField extends FieldClass
 	{
 		return $this->stats ??= Stats::from(
 			model:   $this->model,
-			reports: $this->reports,
+			reports: $this->reports ?? [],
 			size:    $this->size ?? 'large'
 		);
 	}
--- a/src/Form/FieldClass.php
+++ b/src/Form/FieldClass.php
@@ -2,7 +2,6 @@
 
 namespace Kirby\Form;
 
-use Kirby\Cms\HasSiblings;
 use Kirby\Cms\ModelWithContent;
 use Kirby\Exception\NotFoundException;
 use Kirby\Toolkit\HasI18n;
@@ -23,7 +22,6 @@ use Kirby\Toolkit\HasI18n;
 abstract class FieldClass
 {
 	use HasI18n;
-	use HasSiblings;
 	use Mixin\After;
 	use Mixin\Api;
 	use Mixin\Autofocus;
@@ -37,6 +35,7 @@ abstract class FieldClass
 	use Mixin\Name;
 	use Mixin\Placeholder;
 	use Mixin\Required;
+	use Mixin\Siblings;
 	use Mixin\Translatable;
 	use Mixin\Validation;
 	use Mixin\Value;
@@ -56,7 +55,6 @@ abstract class FieldClass
 		string|null $name = null,
 		array|string|null $placeholder = null,
 		bool|null $required = null,
-		protected Fields|null $siblings = null,
 		bool|null $translate = null,
 		array|null $when = null,
 		string|null $width = null
@@ -73,7 +71,6 @@ abstract class FieldClass
 		$this->setName($name);
 		$this->setPlaceholder($placeholder);
 		$this->setRequired($required);
-		$this->setSiblings($siblings);
 		$this->setTranslate($translate);
 		$this->setWhen($when);
 		$this->setWidth($width);
@@ -174,16 +171,6 @@ abstract class FieldClass
 	{
 		$this->value = $this->emptyValue();
 		return $this;
-	}
-
-	protected function setSiblings(Fields|null $siblings = null): void
-	{
-		$this->siblings = $siblings ?? new Fields([$this]);
-	}
-
-	protected function siblingsCollection(): Fields
-	{
-		return $this->siblings;
 	}
 
 	/**
--- a/src/Form/FieldClass.php
+++ b/src/Form/FieldClass.php
@@ -109,6 +109,19 @@ abstract class FieldClass
 		return [];
 	}
 
+	/**
+	 * Creates a new field instance from a $props array
+	 * @since 6.0.0
+	 */
+	public static function factory(
+		array $props,
+		Fields|null $siblings = null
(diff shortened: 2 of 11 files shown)


src/Form/Field/StatsField.php
```diff
@@ -17,27 +17,26 @@
...
-		parent::__construct($params);
-
-		$this->reports = $params['reports'] ?? [];
-		$this->size    = $params['size']    ?? 'large';
+	public function __construct(
+		/**
+		 * Array or query string for reports.
+		 * Each report needs a `label` and `value` and can have
+		 * additional `info`, `link`, `icon` and `theme` settings.
+		 */
```
distantnative: Not sure actually if comments like these per argument work and can be parsed by our docs reflections. For reading the code, I think the would be superior to putting them all together in the method docblock.

src/Form/FieldClass.php
```diff
@@ -37,42 +40,54 @@ abstract class FieldClass
...
+		$this->setModel($model);
+		$this->setName($name);
+		$this->setPlaceholder($placeholder);
+		$this->setRequired($required);
+		$this->setSiblings($siblings);
+		$this->setTranslate($translate);
+		$this->setWhen($when);
+		$this->setWidth($width);
+
+		if ($value !== null) {
```
distantnative: The old version check for array key exists... that could also be `null`. How do we handle that?

afbora: What I mean that in https://github.com/getkirby/kirby/pull/7411#discussion_r2231965380

distantnative: True. I think the way here might still make the most sense:
- On object initialization, we will the value with anything but null
- The default for the value can be set via the property default in the class
- I don't see cases where the property default is different than `null` but on initialization we want to explicitly set it to `null` (but this is the case that would be tricky if I am wrong)

src/Form/Field/StatsField.php
lukasbestle: I agree. Would also allow us to make these descriptions a lot more detailed and add additional metadata.

I don't fully understand our reference code, but I assume we would need to get the doc comment from the PHP reflection, right? My first instinct was a method like `ReflectionParameter::getDocComment()`, but that does not seem to exist :(

src/Form/FieldClass.php
lukasbestle: I think we need a difference between "empty existing field" and "non-existing field". I.e. with fields that don't store anything like `line`. With our goal to eventually get rid of sections, those will also be candidates for such "fields that don't store anything".

So I would expect a field value of `null` to mean "filter this out of the content data" and `""` to mean "this field exists but is currently empty". Not sure if we already handle it this way.

This would mean that we need a way to set the value to `null`, however I think the current implementation looks good then. If the field does not store anything, it doesn't need to be filled.

src/Form/Field/StatsField.php
distantnative: I tried a few things but it really seems like those direct argument comments do not get picked up by the DocBlock library we use. Which is a pity cause I think putting it all as `@param` in the method DocBlock would get really messy and hard to maintain for us.

distantnative: Gotta correct myself:

```php
new ReflectionProperty()->getDocComment()
```

seems to be able to access it, even if the comment is within the parameters/promoted named arguments.

So I guess one could create some logic in our docs classes that combines the information: If `ReflectionParameter::isPromoted()`, also take `ReflectionProperty::getDocComment()` into consideration.

src/Form/FieldClass.php
distantnative: @lukasbestle Good point. I think @bastianallgeier had in his experiments a new `FieldWithValue` class (probably named differently) that would extend the basic class, so that the basic class would not introduce any value related behavior. I think that would make a lot of sense (also probably for other parts where we ship defaults for all fields but maybe the basic field class needs to be more basic and we should move those to some other abstract field class types). 

But maybe we do that as separate steps in following PRs? I am trying to balance to keep the PRs rather small in their scope, but also to keep the code at every stage functional and making sense.

src/Form/FieldClass.php
```diff
@@ -37,42 +40,54 @@ abstract class FieldClass
...
 
 	public function __call(string $param, array $args): mixed
 	{
-		if (isset($this->$param) === true) {
+		if (property_exists($this, $param) === true) {
 			return $this->$param;
 		}
 
-		return $this->params[$param] ?? null;
+		throw new NotFoundException(message: 'Method or option "' . $param . '" does not exist for field type "' . $this->type() . '"');
```
distantnative, line 82: @bastianallgeier I was asking the other day about the magic `validate` prop that's not really defined anywhere. I guess this change would cause a problem with our magic usage there.

src/Form/Field/StatsField.php
distantnative: Thinking more about it, I think something like this needs to be the way anyway as a bunch of the properties will be defined in mixins. So `ReflectionProperty::getDocComment()` would be the right way as well there.

But that leaves us with the questions from my initial post:
- How do we know if a `__construct()` parameter is supposed otherwise to be a blueprint option? Existence of a property with the same name/parameter is promoted?
- What's our rule to pass through parameters of parent field classes? `...$args`? And how do we loop through parent classes to gather the whole list of blueprint options?
- How do we remove an option from the list if the specific field isn't using it but a parent class defines it?

lukasbestle: I don't quite understand the first question, could you rephrase that?

The second and third seem to be related. I think if we want to have a way to remove arguments/properties from parent classes, we need to define all individually. We can only use the splat if we want to inherit all. And by that point, I wonder if we should rather always define them individually for better IDE support. We can still use `@inherit` in the doc comments.

bastianallgeier: I think Nico means that a parameter could be needed internally (e.g. $model) but should not be included in the docs later. 

We could look into Attributes for that. For a potential blueprint editor, this would be a great way for additional configuration options anyway. Such as a field type or additional validation options.

src/Form/FieldClass.php
bastianallgeier: Yes, I still think that this would be the cleanest approach. So far my structure would have been something like this (names are completely temporary for now): 

- FoundationFieldClass (all the basic features of a field)
  - InputFieldClass (all features for fields with values)
  - DisplayFieldClass (all features for display-only fields. E.g. info, headline, etc.)

src/Form/FieldClass.php
bastianallgeier: If we want to bring back this undocumented option for input fields, we should really do that in a way that sets a validate property and gets rid of the magic anyway.

src/Form/Field/StatsField.php
distantnative: That's a great idea! I forgot about attributes being available not only for classes and methods.

src/Form/FieldClass.php
bastianallgeier: I think we can safely discard this. When we get to a field that needs custom validators, we can solve it there.

bastianallgeier:
@distantnative it's your PR, but I think you should still have another look after my changes.

PR #7609, opened 2025-09-29
fix: Prevent page create dialog from nesting when changing template

distantnative:
### 🐛 Bug fixes

- Reloading a dialog/drawer will no longer nest the new dialog/drawer but replace the previous one

### For review team


bastianallgeier:
Do you think we can add the change to the reload method to the patch release?

distantnative:
@bastianallgeier I think so. It's basically a regression in 5.1.0 currently that `reload()` starts nesting dialogs (and probably long existing bugs for drawers).

bastianallgeier (approved):

PR #7611, opened 2025-09-29
feat: `Uri::inherit()`

distantnative:
## Description

New method used to preserve/copy query, params and fragment from an old URL onto a new URL for a redirect.

### ✨ Enhancements

- New `Kirby\Http\Uri::inherit()` methods that copies over query, params and fragment from another URL

### For review team

### DIFF
--- a/src/Http/Uri.php
+++ b/src/Http/Uri.php
@@ -304,6 +304,27 @@ class Uri implements Stringable
 	}
 
 	/**
+	 * Inherit query, params and fragment from a parent Uri
+	 * @since 5.2.0
+	 * @return $this
+	 */
+	public function inherit(Uri|string $parent): static
+	{
+		if (is_string($parent) === true) {
+			$parent = new static($parent);
+		}
+
+		$this->query->merge($parent->query());
+		$this->params->merge($parent->params());
+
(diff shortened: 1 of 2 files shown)


src/Http/Uri.php
```diff
@@ -303,6 +303,27 @@ public static function index(array $props = []): static
 		return new static($url, $props);
 	}
 
+	/**
+	 * Inherit query, params and fragment from a parent Uri
+	 * @since 5.2.0
+	 * @return $this
+	 */
+	public function inherit(Uri|string $parent): static
```
distantnative, line 311: If someone has a better suggestion for the name, I am not too sure about `inherit`.

lukasbestle: Some ideas in a rough brainstorming: `integrate`, `succeed`, `combine`, `coalesce`

distantnative: `adopt`

bastianallgeier: Maybe `merge`?

bastianallgeier: You know what. While reading through the code, I really started to like inherit. Let's keep it.

bastianallgeier (approved):

PR #7620, opened 2025-10-09
feat: New Kirby\Reflection\Constructor class

bastianallgeier:
## Description

This is a small step towards an easier transition to classes with named parameters. By being able to reflect which arguments are accepted or ignored in a class constructor, we can pass arrays of arguments (e.g. from blueprints) more reliably and without risiking exceptions because of unknown arguments. 

### 🎉 Enhancements

- New `Kirby\Reflection\Constructor` class

## Docs

```php
use Kirby\Reflection\Constructor;

class Reflectable
{
  public function __construct( 
      protected $a,
      protected $b
  ) {
  }
}

$constructor = new Constructor(Reflectable::class);

$constructor->getAcceptedArguments([
  'a' => 'Test A',
  'b' => 'Test B',
  'c' => 'Test C'
]);
// will keep 'a' and 'b'

$constructor->getIgnoredArguments([
  'a' => 'Test A',
  'b' => 'Test B',
  'c' => 'Test C'
]);
// will keep 'c'
```

### For review team

### DIFF
--- a/src/Reflection/Constructor.php
+++ b/src/Reflection/Constructor.php
@@ -14,6 +14,7 @@ use ReflectionParameter;
  * @link      https://getkirby.com
  * @copyright Bastian Allgeier
  * @license   https://opensource.org/licenses/MIT
+ * @since     5.2.0
  */
 class Constructor extends ReflectionMethod
 {
--- a/src/Reflection/Constructor.php
+++ b/src/Reflection/Constructor.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Kirby\Reflection;
+
+use ReflectionMethod;
+use ReflectionParameter;
+
+/**
+ * Specialized Reflection Method class to
+ * inspect a class constructor
+ *
+ * @package   Kirby Reflection
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://opensource.org/licenses/MIT
+ */
+class Constructor extends ReflectionMethod
+{
+	public function __construct(object|string $objectOrClass)
+	{
+		parent::__construct($objectOrClass, '__construct');
+	}
+
+	/**
+	 * Group arguments into 'accepted' and 'ignored' arrays
+	 */
+	public function classifyArguments(array $arguments): array
+	{
+		$parameterNames = $this->getParameterNames();
+
+		$accepted = [];
+		$ignored  = [];
+
+		foreach ($arguments as $argumentName => $argumentValue) {
+			if (in_array($argumentName, $parameterNames) === true) {
+				$accepted[$argumentName] = $argumentValue;
+			} else {
+				$ignored[$argumentName] = $argumentValue;
+			}
+		}
+
+		return [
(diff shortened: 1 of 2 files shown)


lukasbestle (dismissed):
Code-wise this looks good. I also like that it's a lot more focused and specific.

But still torn if using reflections is a good solution in general, especially for such a security-relevant part as permissions. But let's continue that discussion in #7559 and keep this waiting until then.

bastianallgeier:
@distantnative @lukasbestle because we had this conversation in the other PR yesterday, I want to emphasise here again that this is the only viable solution in my opinion to go all in with named parameters in constructors, but still have a pathway for passing arrays of arguments to a class. We will almost exclusively need this in static `from` helper methods. 

We will need this not just for Permissions but for any class that could get its options from a blueprint or a PHP array (e.g. from config, cached settings, etc.)

We will also need this for any class that should have a toArray option and where we don't want to or cannot redefine all props for the output array manually.

src/Reflection/Constructor.php
```diff
@@ -0,0 +1,72 @@
...
+	 */
+	public function classifyArguments(array $arguments): array
+	{
+		$parameterNames = $this->getParameterNames();
+
+		$accepted = [];
+		$ignored  = [];
+
+		foreach ($arguments as $argumentName => $argumentValue) {
+			if (in_array($argumentName, $parameterNames) === true) {
```
distantnative: We should use the strict parameter for in_array (just a best practice we're trying to achieve in as many places as possible).

tests/Reflection/ConstructorTest.php
```diff
@@ -0,0 +1,105 @@
...
+		$result = $this->reflection()->getAcceptedArguments([
+			...$accepted,
+			...$ignored
+		]);
+
+		$this->assertSame($accepted, $result);
+
+		$result = $this->reflection()->getAcceptedArguments([]);
+
+		$this->assertSame([], $result);
```
distantnative, line 69: Should have a test with passing just ignored and resulting in an empty array.

tests/Reflection/ConstructorTest.php
```diff
@@ -0,0 +1,105 @@
...
+		$result = $this->reflection()->getIgnoredArguments([
+			...$accepted,
+			...$ignored
+		]);
+
+		$this->assertSame($ignored, $result);
+
+		$result = $this->reflection()->getIgnoredArguments([]);
+
+		$this->assertSame([], $result);
```
distantnative, line 100: Should have a test with just passing accepted resulting in an empty array.

distantnative:
Ah one last comment: I would not list this in the changelog under "Features". I understand that in the classic feature/fix/refact this would be considered a feature as new code/new class. But I think we have to look at this more in the product perspective and where this change would benefit users most being listed. And there I think it's more in the refactoring area... maybe enhancement. But not leading as a feature, IMO.

bastianallgeier:
@distantnative you are totally right. I still think too much in "thing has been added" "thing has been fixed" "thing is now better"

distantnative (dismissed):

distantnative:
Ahh one more thing: `@since` tag is missing

PR #7622, opened 2025-10-10
Underscore-prefixed folders

medienbaecker:
## Description

Building upon the underscore concept introduced with `_drafts` and `_changes`, this PR adds a generic system to register underscore-prefixed folders like `_modules`, `_media`, etc.

I really like this "readable/understandable on the filesystem level" way of going about non-pages and as a plugin developer I'd love to use it myself. I tried to keep it as simple and unopinionated as possible but I know there's always a more elegant solution, no matter how long I work on something. Consider this a first draft.

### 🎉 Features

New `folders` extension for registering handlers for underscore-prefixed folders. These folders are discovered in content but excluded from regular page children, allowing plugins to handle them with custom logic. You could display them in the Panel or not, but this way they definitely get excluded from things like the sitemap because they are not "pages" in the first place.

Example use case for the modules plugin. There are simpler examples but this is why I started working on this PR, so I still wanted to use it:

```php
Kirby::plugin('medienbaecker/modules', [
    'folders' => [
        'modules' => function ($parent) {
            $path = $parent->inventory()['folders']['modules'] ?? null;

            return new class($parent, $path) {
                private $parent;
                private $path;

                public function __construct($parent, $path) {
                    $this->parent = $parent;
                    $this->path = $path;
                }

                public function pages(): Pages {
                    if (!$this->path) {
                        return new Pages([], $this->parent);
                    }

                    $inventory = Dir::inventory($this->path, ...);
                    return Pages::factory($inventory['children'], $this->parent);
                }
            };
        }
    ]
]);
```

Usage in templates:
```php
$modules = $page->folder('modules')->pages();
```

Usage in blueprints ("pages" section):
```yml
parent: page.folder("modules").pages()
```

Things I learned along the way:

1. First I had a simpler, more direct wrapper but decided to go with storing only the path to not have a performance downgrade. That's why I have to go to the inventory first.
2. First I returned the pages collection directly but found it to be more in line with other Kirby functionality to use a class with a `pages()` method (or `files()` in other cases). This also allows for additional methods like `exists()` or `create()`.
3. I've changed the name of this functionality a dozen times but ended up with "folders" in the end.

### ♻️ Refactored
`_drafts` uses the same folder handler system instead of being hardcoded. I've also included a placeholder for `_changes` so it can not be overwritten.

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -1658,6 +1658,14 @@ class App
 	}
 
 	/**
+	 * Returns a registered folder handler
+	 */
+	public function folder(string $name): Closure|null
+	{
+		return $this->extensions['folders'][$name] ?? null;
+	}
+
+	/**
 	 * Returns a system url
 	 *
 	 * @param bool $object If set to `true`, the URL is converted to an object
--- a/src/Cms/AppPlugins.php
+++ b/src/Cms/AppPlugins.php
@@ -82,6 +82,7 @@ trait AppPlugins
 		'templates' => [],
 		'thirdParty' => [],
 		'translations' => [],
+		'folders' => [],
 		'userMethods' => [],
 		'userModels' => [],
 		'usersMethods' => [],
@@ -665,6 +666,20 @@ trait AppPlugins
 	}
 
 	/**
+	 * Registers folder handlers
+	 */
+	protected function extendFolders(array $handlers): array
+	{
+		// prevent overriding existing handlers
+		$handlers = array_diff_key($handlers, $this->extensions['folders']);
+
+		return $this->extensions['folders'] = [
+			...$this->extensions['folders'],
+			...$handlers
+		];
+	}
+
+	/**
 	 * Registers additional user methods
 	 */
 	protected function extendUserMethods(array $methods): array
@@ -825,6 +840,30 @@ trait AppPlugins
 		$this->extendSnippets($this->core->snippets());
 		$this->extendTags($this->core->kirbyTags());
 		$this->extendTemplates($this->core->templates());
+		$this->extendFolders([
+			'drafts' => function ($parent) {
+				$draftsPath = $parent->inventory()['folders']['drafts'] ?? null;
+
+				if ($draftsPath === null) {
+					return new Pages([], $parent);
+				}
+
+				$kirby = $parent->kirby();
+
+				$inventory = Dir::inventory(
+					$draftsPath,
+					$kirby->contentExtension(),
+					$kirby->contentIgnore(),
+					$kirby->multilang()
+				);
+
+				return Pages::factory($inventory['children'], $parent, true);
+			},
+			'changes' => function ($parent) {
+				// reserved for content versioning
+				return null;
+			}
+		]);
 	}
 
 	/**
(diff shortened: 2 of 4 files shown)


distantnative:
Thanks for giving this a stab!

I think I like the general idea to provide an abstract way of defining such underscore-prefixed folders inside a page. The problem I see here is that with v5 we started moving towards adding different storage providers that could be different from filesystems. That's why e.g. `folders` is maybe not the best name for such an abstract feature and only the filesystem storage provider (currently `PlaintextStorage`) would implement this feature via underscored folders. 

So I think if we want to add this as a general feature, we need to spent some more thought on how to make it more abstract and fall in line with the ideas and plans to rely on the storage classes for the concrete implementation.

medienbaecker:
Thanks for the quick reply, Nico!

I understand your concerns about the different storage providers. While I do know that's something you've started working on for v5, I haven't fully grasped the whole thing. 

I see that `_drafts` currently use the same filesystem-specific approach via `Dir::inventory()`. I also noticed your comment in `Storage.php` mentioning that "the CMS core relies on the filesystem." So the question is, should this wait until storage abstraction is complete? Is there value in having this PR then for discussions about naming or so?

bastianallgeier:
I really like how simple it is to get this implemented. 

I think we can split this mentally into two parts. 

1.) Dir::inventory() changes

I don't see any negative side effects in changing the inventory method like you did here and register all those underscore folders. This would already be super handy. 

2.) "Folder" 

I agree with @distantnative that we need a more universal term for it if we want to move forward with more universal storage options. But when we agree on a better name, I don't think that the implementation needs to be changed drastically. Let's just say we call it `$page->fragments()` for the sake of this conversation. I would not really care if such a fragment is coming from the filesystem with our old way of directly reading from it, or if it would be implemented in a more abstracted way. This could easily be decided on a fragment basis. Our drafts and changes would use our storage handlers and plugin devs could decide to do this too, but they could also just call a rest api or read directly from disk. Do you see any problems there? 

More ideas for names: 

- `$page->partial('drafts')`
- `$page->segment('drafts')`
- `$page->unit('drafts')`

We just have to be aware that those will replace content fields again. I guess it's more unlikely that a field is called fragment or partial than unit for example. 

It also depends a bit on what is going to be stored in there. Thinking in terms of drafts, changes or modules, it's all content-related and we could prefix it like that. `$page->contentFragments()` This would most definitely avoid collissions. 

But if this concept should also work for media or other file types, we might prefer something like `$page->storageContainer()` or `$page->storageRepository()`

medienbaecker:
Thanks for the feedback, Bastian. I'm excited you guys see value in this.

Reserving/replacing content fields is definitely a concern. Something like `folder` (aside from the strong filesystem vibes) could definitely be a breaking change for installations that use `folder` fields.

I actually like `fragment` (or `contentFragment`) as it's weird enough for people to not have a ton of `fragment` fields already and it's not saying anything _about_ the content. It's probably always "content" in the broader sense of the word. It's also always related to the page it's in. An alternative to `fragment` would be `resource`. it's a bit more generic but potentially breaking existing resource fields.

Two other ideas that I had along the way (and scrapped), I'm just going to drop them here for the discussion:

1. Using underscores in the method itself like `$page->_modules()`. It's a bit confusing because of private methods in PHP.
2. Using the [`$page->content()`](https://getkirby.com/docs/reference/objects/cms/page/content) method itself and adding an attribute. As it already has the optional language code (and that only for multi-language sites?) it might be confusing, too.

Issue #7623, opened 2025-10-14
Incomplete deletion of pages with files when using batch delete


benwest:
## Description

When batch deleting pages which have files, I'm finding that the content files and file-files are not deleted completely, and are leaving scraps behind.

## To reproduce

1. Run Kirby Plainkit (5.1.2, PHP 8.4 development server on MacOS 15.6, but I also see this behaviour on my client's Apache server with PHP 8.2).
2. Add `batch: true` to the site.yml blueprint.
3. Create 2 new pages, make them public, and add a file to each.
4. Batch delete the 2 pages from the Panel.
5. One page's folder and files are still there.

<img width="190" height="203" alt="Image" src="https://github.com/user-attachments/assets/e32acb8e-110b-4414-aa5a-21e97ef3a66e" />

Before - default.txts contain uuid, title and text

<img width="213" height="114" alt="Image" src="https://github.com/user-attachments/assets/533c4b6b-8f23-4871-8a35-b3893bffa421" />

After - default.txt contains uuid only - not the same as the uuid before deletion. j6tcu.png.txt is not changed.

I tried to step through the deletion with xdebug, but then it worked fine and deleted everything - which makes me wonder if it's a race condition with the file system?

Deleting the pages one at a time always works. So does batch-deleting pages without files.

Hope you can help, kirby is great 😘

Issue #7615, opened 2025-10-06
Strange behavior when deleting pages with children


texnixe:
## Description
When deleting a parent page with children, in some situations the page is not deleted, but the page content file renamed to `default.txt`, while the previously default numbered children are all prepended with 1_.

**Expected behavior**  
Page should be deleted

**Screenshots**  

## To reproduce

1. In a 5.1.2 Starterkit, open the Panel
2. From the Dashboard, open the `Photography` page, then from the Settings menu, select `Delete` 
3. Enter the page title and confirm with the `Delete` button
4. You are redirected to the Dashboard, but you will see that the page is still there.
5. Inspect the filesystem to the effects described above.
6. Note that if you delete the Notes page afterwards, this does not happen. Also, when you delete the `Photography` page again now, it will be gone.

## Your setup
**Kirby Version**  
Fresh 5.1.2 Starterkit on MacOS with Herd as local dev server

**Console output**  

**Your system (please complete the following information)**
 - Device: 
 - OS: 
 - Browser: 
 - Version: 

**Additional context**  

See also: https://forum.getkirby.com/t/publish-pages-in-batch/34785/

afbora:
I can confirm that this issue is valid for 5.0.0+ and works as expected in 4.8.0.

afbora:
### First sight of the issue

I've found related line where issue starts. The source of the problem is the `->resortSiblingsAfterUnlisting()` call in `PageActions::delete()` method. My guess is that this method may still be storing old siblings when it runs.

https://github.com/getkirby/kirby/blob/0f916becac8a2bc824a72cdb70066ed56b50c0f4/src/Cms/PageActions.php#L592-L597

Note: The reason it works on the notes page is because notes use a num `date`. This only affects those with the `default` num.

### Possible solution

I think we should delete the actually files and folders from the disc before re-sorting.

````php
// delete all public media files
Dir::remove($old->mediaRoot());

// delete the content folder for this page
Dir::remove($old->root());

// re-sorting stuff
````

distantnative:
@afbora but that's what supposedly the line above (`$old->versions()->delete()`) should do. So something might be broken there?

Cause we cannot add those file system method to page actions as with v5 now the page might use a different storage than plaintext and then it would crash.

afbora:
@distantnative Yes, it's possible. As I said, if the `resortSiblingsAfterUnlisting` method isn't called, there's no problem. Not sure but I guess this means there's still old data in the `$old` object's cache.

[...]

PR #7625, opened 2025-10-16
fix: Fix recursive and batch delete issues

bastianallgeier:
### 🐛 Bug fixes

- Deleting sorted pages recursively or in a batch could lead to unwanted side effects and left-over folders and files. This fix will now make sure that all pages are properly deleted. https://github.com/getkirby/kirby/issues/7623 & https://github.com/getkirby/kirby/issues/7615

### For review team

### DIFF
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -144,6 +144,7 @@ class Pages extends Collection
 				}
 
 				$page->delete();
+				$this->remove($id);
 			} catch (Throwable $e) {
 				$exceptions[$id] = $e;
 			}
--- a/src/Cms/Pages.php
+++ b/src/Cms/Pages.php
@@ -132,19 +132,18 @@ class Pages extends Collection
 	public function delete(array $ids): void
 	{
 		$exceptions = [];
+		$kirby      = App::instance();
 
 		// delete all pages and collect errors
 		foreach ($ids as $id) {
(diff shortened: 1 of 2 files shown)


afbora (dismissed):

afbora (dismissed):

src/Cms/Pages.php
```diff
@@ -132,19 +132,19 @@ public function code(): Files
...
 		$exceptions = [];
+		$kirby      = App::instance();
 
 		// delete all pages and collect errors
 		foreach ($ids as $id) {
 			try {
-				$model = $this->get($id);
-
-				if ($model instanceof Page === false) {
+				if ($this->get($id) instanceof Page === false || !$page = $kirby->page($id)) {
```
distantnative:
```suggestion
				// Explanation: We get the page object from the global context
				// as the objects in the pages collection itself could have rendered
				// outdated from a sibling delete action in this loop (e.g. resorting
				// after deleting a sibling page and leaving the object in this collection
				// with an old root path).
				//
				// TODO: We can remove this part as soon
				// as we move away from our immutable object architecture.
				$page = $kirby->page($id);
				
				if ($page === null || $this->get($id) instanceof Page === false) {
```

distantnative (approved):

afbora (approved):

PR #7630, opened 2025-10-17
feat: CORS handling across Kirby routes

johannschopplich:
## Description

This PR adds centralized CORS (Cross-Origin Resource Sharing) support to Kirby across all application routes. The implementation is opt-in and backward compatible – CORS is disabled by default and requires explicit configuration.

### 🎉 Features
- New CORS configuration system for cross-origin resource sharing support with sensible defaults
- Automatic CORS preflight request handling for all routes
- Support for all standard CORS headers: `Access-Control-Allow-Origin`, `Access-Control-Allow-Methods`, `Access-Control-Allow-Headers`, `Access-Control-Max-Age`, `Access-Control-Allow-Credentials`, and `Access-Control-Expose-Headers`
- Dynamic CORS configuration via closures for request-specific settings

### ✨ Enhancements
- CORS headers are automatically injected into all responses (API, Panel, site routes) when enabled
- New `$kirby->isCorsEnabled()` method to check if CORS is enabled
- Wildcard origin (`*`) with credentials throws a validation error for security
- Smart caching with automatic `Vary` header management

### Configuration

Enable CORS with minimal configuration in your `site/config/config.php`:

```php
return [
    'cors' => true
];
```

This applies sensible defaults that work for most use cases. All defaults can be customized:

```php
return [
    'cors' => [
        'allowOrigin'      => '*',                                    // Default: '*'
        'allowMethods'     => ['GET', 'POST', 'PUT', 'DELETE'],       // Default: ['GET', 'HEAD', 'PUT', 'POST', 'DELETE', 'PATCH']
        'allowHeaders'     => true,                                   // Default: [] (do not allow request headers)
        'maxAge'           => 3600,                                   // Default: null (browser default)
        'allowCredentials' => true,                                   // Default: false
        'exposeHeaders'    => ['X-Custom-Header']                     // Default: [] (no response headers are exposed)
    ]
];
```

### Available Options

| Option | Type | Default | Description |
|--------|------|---------|-------------|
| `allowOrigin` | `string\|array` | `'*'` | Allowed origins (e.g., `'*'`, `'https://example.com'`, or `['https://app1.com', 'https://app2.com']` for multiple origins) |
| `allowMethods` | `string\|array` | `['GET', 'HEAD', 'PUT', 'POST', 'DELETE', 'PATCH']` | Allowed HTTP methods for preflight requests |
| `allowHeaders` | `string\|array\|bool` | `[]` | Allowed request headers. `[]` (default) suppresses `Access-Control-Allow-Headers`; `true` reflects headers from `Access-Control-Request-Headers`; strings/arrays allowlist specific headers explicitly |
| `maxAge` | `int` | `null` | Preflight cache duration in seconds. `null` uses the browser default (typically 5 seconds) |
| `allowCredentials` | `bool` | `false` | Allow requests with credentials (cookies, auth). Cannot be `true` with wildcard origin |
| `exposeHeaders` | `string\|array` | `[]` | Response headers exposed to the browser |

### Enabling CORS

CORS can be enabled in three ways:

#### 1. Boolean (uses all defaults)

```php
return [
    'cors' => true
];
```

#### 2. Array (custom configuration)

```php
return [
    'cors' => [
        'allowOrigin' => 'https://example.com',
        'allowCredentials' => true
    ]
];
```

#### 3. Closure (dynamic/request-based)

```php
return [
    'cors' => function ($kirby) {
        $origin = $kirby->request()->header('Origin');
        
        // Allow specific origins with credentials
        if (in_array($origin, ['https://app1.com', 'https://app2.com'])) {
            return [
                'allowOrigin' => $origin,
                'allowCredentials' => true,
                'allowMethods' => ['GET', 'POST']
            ];
        }
        
        // Fallback to wildcard for other origins
        return ['allowOrigin' => '*'];
    }
];
```

**Note:** Setting `'cors' => []` (empty array) is equivalent to `'cors' => true` and enables CORS with defaults. To disable CORS, use `'cors' => false` or omit the option entirely.

### Array Support

The `allowOrigin`, `allowMethods`, `allowHeaders`, and `exposeHeaders` options support both string and array formats:

```php
return [
    'cors' => [
        'allowOrigin'   => ['https://app1.com', 'https://app2.com'],   // Multiple origins
        'allowMethods'  => ['GET', 'POST', 'PUT', 'DELETE'],           // Array format
        'allowHeaders'  => ['Content-Type', 'Authorization'],          // Explicit allowlist
        'exposeHeaders' => ['X-Total-Count', 'X-Page-Number']          // Array format
    ]
];
```

#### Minimal Setup (Public API)

```php
return [
    'cors' => true
];
```

This uses the defaults: wildcard origin, standard method allowlist, no custom headers allowed, no credentials.

#### Headless CMS with Authentication

```php
return [
    'cors' => [
        'allowOrigin'      => 'https://example.com',
        'allowMethods'     => ['GET', 'POST', 'PATCH', 'DELETE'],
        'allowHeaders'     => true,
        'allowCredentials' => true
    ]
];
```

#### Multiple Frontend Apps

```php
return [
    'cors' => [
        'allowOrigin'      => ['https://app.example.com', 'https://admin.example.com'],
        'allowHeaders'     => ['Authorization', 'Content-Type'],
        'allowCredentials' => true
    ]
];
```

### Security Considerations

- Enabling CORS allows external origins to interact with your Kirby site. This means that sites hosted on external origins (domains) can access or control your Kirby site via the browser of their visitors. Only enable CORS when needed, and restrict the configuration (origins, methods, headers) to the minimum required for your use case.
- `'allowCredentials' => true` lets browsers include cookies and HTTP authentication with cross-origin requests, effectively granting the external origin the same permissions as the logged-in user. Only enable credentials when the requesting origin is fully trusted and under your control.

### API Methods

```php
$kirby = kirby();

// Check if CORS is enabled
if ($kirby->isCorsEnabled()) {
    // CORS is active
}

// Get CORS headers (static method)
$headers = Kirby\Cms\Cors::headers();
// Returns array with CORS headers for normal requests

// Get preflight headers
$preflightHeaders = Kirby\Cms\Cors::headers(preflight: true);
// Returns array with CORS headers including preflight-specific headers
```

### How It Works

1. **Simple enablement**: Set `'cors' => true`, `'cors' => []`, or `'cors' => [...]` to enable CORS. CORS is only disabled when the option is `false` or not set.
2. **Automatic preflight handling**: When CORS is enabled, all `OPTIONS` CORS preflight requests automatically receive a `204 No Content` response with appropriate CORS headers.
3. **Sensible defaults**: When CORS is enabled without specific configuration, sensible defaults are applied automatically:
   - `allowOrigin`: `'*'` (public access)
   - `allowMethods`: All common HTTP methods including `OPTIONS`
   - `allowHeaders`: `[]` (no `Access-Control-Allow-Headers` is sent)
   - `maxAge`: `null` (uses browser default, typically 5 seconds)
4. **Dynamic configuration**: Use a closure for request-specific CORS settings, allowing different rules based on the origin, authentication state, or other request properties.
5. **Header injection**: CORS headers are lazily injected into the `Responder::headers()` method for all responses. Custom headers set by plugins or user code are never overridden.
6. **Vary header management** ([Hono](https://hono.dev/)-inspired):
   - **Wildcard origins (`*`)**: No `Vary: Origin` header is added (cache-efficient since response is identical for all origins)
   - **Specific origins**: Automatic `Vary: Origin` header ensures different origins are cached separately
   - **Header reflection**: Automatic `Vary: Access-Control-Request-Headers` for preflight requests when reflection is enabled
   - **Auth/Cookie tracking**: Automatic `Vary: Authorization, Cookie` when response uses authentication or cookies
   - **Smart merging**: All Vary values are combined intelligently without duplication
7. **Header reflection opt-in**: With `'allowHeaders' => true`, preflight requests mirror the headers requested by the client while `Vary: Access-Control-Request-Headers` ensures correct caching. The default `[]` remains secure-by-default by omitting the header.

---

### For review team


config/routes.php
```diff
@@ -3,6 +3,8 @@
 use Kirby\Cms\App;
 use Kirby\Cms\LanguageRoutes;
 use Kirby\Cms\Media;
+use Kirby\Cms\Responder;
+use Kirby\Http\Response;
```
johannschopplich, line 7: `use Kirby\Cms\Response` extends `use Kirby\Http\Response`. Should I use the base class or rather in the CMS context the `use Kirby\Cms\Response` instead?

distantnative: I think both are fine.

src/Cms/App.php
```diff
@@ -543,6 +543,15 @@ public function csrf(string|null $check = null): string|bool
 		return false;
 	}
 
+	/**
+	 * Checks if CORS support is enabled
+	 * @since 5.2.0
+	 */
+	public function cors(): bool
```
distantnative: `Discuss:` Just a thought, but not sure myself, the short name also seems fine, but could also have a name that directly signifies returning a boolean, e.g. `::isCors()`

distantnative (commented):
Code reads well 👍 - I have too little knowledge of the CORS headers to comment on whether they're used correctly

src/Cms/App.php
johannschopplich: I’ve considered this as well. In other classes, Kirby generally follows the `isXyz` naming convention, whereas app-level boolean functions omit the is prefix, such as `multilang`. 🤔

bastianallgeier: We normally try to use is method names consistently and when we don't it's often either very old code or the is version sounds super weird grammatically. Multilang is very old. isMultilang would actually work really well. 

I would vote for isCors here. I think it works grammatically well enough.

config/routes.php
bastianallgeier: We started to merge more and more of our classes and avoid unnecessary extensions. With some parts of our framework, we once thought about making it public for people outside the Kirby ecosystem and that's why we tried to keep them independent from App logic, but we found this to be way too exhausting for us as a small team. That's why I think it's actually better to use Kirby\Http\Response here, because that class could be another good case for such a merge at some point.

johannschopplich: Thanks for the explanation! I'll keep `Kirby\Http\Response` and consider this resolved then.

src/Cms/App.php
bastianallgeier: Second thought: The option is called `cors.enabled` Maybe we could go for `isCorsEnabled`? A bit more verbose, but granmatically a lot nicer.

bastianallgeier:
Pinging @lukasbestle here. I think it would be important to have a second look at this from a security perspective.

src/Cms/App.php
johannschopplich: Good to know! My 2 cents:
- `isCors` – CORS is a protocol/feature, not a state
- `isCorsEnabled` – might be better

johannschopplich: Haha, same thought! Agreed. 🤝

lukasbestle (commented):
I can look into it more closely in the next days, but here are already some notes from my first look:

- I think it's a very useful feature overall.
- Thanks for your detailed tests and documentation. 💛 
- The `Responder::corsHeaders()` method is very long. Maybe we could turn it into its own class so the logic can be more modular.
- I have two suggestions for the config:


1. For most other (newer) core options that can be enabled but also customized (e.g. `api`, `content.uuid`, `languages`, `panel`), we use the pattern of a `bool|array` option. Defining a custom array implies that the feature is enabled. So for example:

```php
// enabled with default config
'cors' => true,

// the same
'cors' => [],

// custom overrides
'cors' => [
  'allowOrigin'      => 'https://example.com',
  'allowMethods'     => ['GET', 'POST', 'PATCH', 'DELETE'],
  'allowCredentials' => true
]
```

2. Maybe we could even support a closure for the `cors` option (like with the `content.fileRedirects` option). This would then be able to return custom config for the current request. Not sure though if we can make it work to generate consistent results for preflight and real requests.

johannschopplich:
@lukasbestle Thanks for the feedback! Love it. Feels much more Kirby-esque. I have moved the `Responder::corsHeader` method into a dedicated `Cors` class with composable methods.

The `bool|array` pattern is also lovely. I have implemented that approach and dropped the `enabled` key.

src/Cms/App.php
```diff
@@ -543,6 +543,15 @@ public function csrf(string|null $check = null): string|bool
 		return false;
 	}
 
+	/**
+	 * Checks if CORS support is enabled
+	 * @since 5.2.0
+	 */
+	public function isCorsEnabled(): bool
```
lukasbestle, line 550: `Thought`: The `App` class already covers a lot of features and therefore has quite many public methods. This makes the class hard to maintain and to keep an overview of all functionality.

`Suggestion`: With the new `Cors` class, we could move this method to `Cors::isEnabled()`.

config/routes.php
```diff
@@ -29,6 +31,24 @@
...
+					return null;
+				}
+
+				$headers = Cors::headers(preflight: true);
+
+				if ($headers === []) {
+					return null;
+				}
+
+				return new Response('', null, 204, $headers);
```
lukasbestle, line 60: `Question`: I wonder if we should rather use `HTTP 200` here. I stumbled across this on MDN, but not sure if it's (still) relevant.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/OPTIONS#:~:text=204%20No%20Content%20applies%20to%20the%20resource

config/routes.php
```diff
@@ -29,6 +31,24 @@
 	 * plugins.
 	 */
 	$before = [
+		// handle CORS preflight requests
+		[
+			'pattern' => '(:all)',
+			'method'  => 'OPTIONS',
+			'action'  => function () use ($kirby) {
+				if ($kirby->isCorsEnabled() === false) {
+					return null;
```
lukasbestle: `Question`: Do we need to support "regular" `OPTIONS` requests as well? I.e. do we need to return an `Allow` header with all HTTP methods that fit the requested route?

This would pretty much be an independent feature from CORS support and I wonder if this is useful in the wild outside of APIs. But I also wonder if this new core route would make such custom `OPTIONS` responses harder.

lukasbestle (commented):
The `Cors` class looks great in its general structure, I love how this cleans it up so much.

Because of the size of the PR, I am kind of reviewing "in layers". Hope it's OK if I already share another part of my comments without having looked at everything already.

src/Cms/App.php
bastianallgeier: I think this is something that we should agree upon first as a team because @distantnative and I already let @johannschopplich refactor this part before with a new method name :) I would not mind another method in the App class, but the `Cors::isEnabled()` pattern would also replicate what we do with `UUIDs` so I definitely see the benefits of that as well.

config/routes.php
johannschopplich: Good catch. From my understanding, `204 No Content` is more semantically correct for a CORS preflight because the server is only confirming permissions via headers and not returning a representation. Some old or nonstandard clients mishandle `204`, that's why Mozilla notes it.

The modern frameworks I know of use `204` across the board.

config/routes.php
johannschopplich: Afaik, browsers never issue a "regular" OPTIONS outside of CORS preflight (due to API tooling). Spec-wise, a generic `OPTIONS` request should always return `204`/`200` with an `Allow` header (see [RFC 9110 §9.3.7](https://www.rfc-editor.org/rfc/rfc9110.html#name-options)). 

You're right, with the new before-route we already intercept every OPTIONS hit... If CORS is enabled we unconditionally answer `204` and short‑circuit the rest of the routing stack, so plugins or custom routes can no longer provide their own `Allow` header.

💡  Idea:
- Keep routing flexible by limiting the core handler to "real" preflight requests: check for `Access-Control-Request-Method` and only return the CORS response when present. Otherwise fall back to normal routing. What do you say?

config/routes.php
lukasbestle: Thanks for your insights. I agree that 204 is the semantically correct way. So let's keep that.

config/routes.php
lukasbestle: That's an awesome idea! Calling `$this->next()` should probably work here to skip the route.

config/routes.php
johannschopplich: Great! 🤝

config/routes.php
johannschopplich: I've implemented support for fall-through OPTIONS requests: https://github.com/getkirby/kirby/pull/7630/commits/9fa18cc0514e89de8cd4b2f4c8b29796455c4b01

If `cors` is disabled, we now use `$this->next()` as well – now the CORS feature is a non-breaking change if someone ever uses custom OPTIONS handlers.

src/Cms/App.php
lukasbestle: We decided to keep it like this as it's overall simpler due to the access to the `App` object.

distantnative (dismissed):

src/Cms/Cors.php
```diff
@@ -0,0 +1,236 @@
...
+	 * Adds the Vary header for cache control
+	 *
+	 * @param array $headers Headers array (passed by reference)
+	 * @param string $allowOrigin Allowed origin value
+	 */
+	protected function addVaryHeader(array &$headers, string $allowOrigin): void
+	{
+		// response varies by origin for non-wildcard origins
+		if ($allowOrigin !== '*') {
+			$currentVary = $this->kirby->request()->header('Vary');
```
lukasbestle: `Question`: What would be the use case for a `Vary` request header? AFAIK, `Vary` is only standardized as response header.

src/Cms/Cors.php
```diff
@@ -0,0 +1,236 @@
...
+	/**
+	 * Adds headers to expose custom headers to the client
+	 *
+	 * @param array $headers Headers array (passed by reference)
+	 */
+	protected function addExposeHeaders(array &$headers): void
+	{
+		$exposeHeaders = $this->config['exposeHeaders'] ?? [];
+
+		if (empty($exposeHeaders) === false) {
```
lukasbestle: `Todo`: We are trying to get away from the `empty()` function (see [this post](https://dev.to/klnjmm/never-use-empty-function-in-php-4pb0)).

As all four cases in this file check whether the input is either a non-empty string or a non-empty array, we could add a `protected function configIsEmpty(mixed $input)` that does:

```php
return (
	(is_array($input) === true && $input !== []) ||
	(is_string($input) === true && $input !== '')
);
```

This would harden the logic against invalid configuration.

src/Cms/Cors.php
```diff
@@ -0,0 +1,236 @@
...
+	{
+		$allowHeaders = $this->config['allowHeaders'] ?? [];
+
+		// reflect request headers if not explicitly configured
+		if (empty($allowHeaders) === true) {
+			$requestHeaders = $this->kirby->request()->header('Access-Control-Request-Headers');
+			if ($requestHeaders !== null) {
+				$allowHeaders = Str::split($requestHeaders, ',');
+			}
+		}
```
lukasbestle: `Comment`: This default to reflect the request headers when an empty array was configured for `allowHeaders` surprises me. This would mean that there is no way to prevent custom request headers entirely.

`Suggestion`: A new option value `true` could handle this reflection mode. Setting the option to `[]` however would disable sending `Access-Control-Allow-Headers` entirely.

Also not sure about the default TBH. I get that reflecting the request increases flexibility and avoids having to allowlist all headers (breaking the app if one was missed). However I'd tend to go for "secure by default" with the option to switch the flexibility mode on at dev's own risk.

src/Cms/Cors.php
```diff
@@ -0,0 +1,236 @@
...
+ * headers for HTTP responses
+ *
+ * @package   Kirby Cms
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @since     5.2.0
+ */
+class Cors
```
lukasbestle, line 19: `General docs comment` (placing this here to have a separate discussion thread):

Your documentation snippet is great and shows real-world config examples. What I'm missing is some warnings on the security impact:
- One general warning that enabling CORS can open the site to malicious requests from external sites (and should only be done if specifically needed and with a configuration that opens up as little as needed).
- A specific warning on the `allowCredentials` option (allows to execute actions with the permissions of the logged in user; option should only be enabled if the external origin is run under the dev's control and can fully be trusted).

lukasbestle (commented):
I'm done with my full review now, having reviewed all code.

General feedback: Amazing work! Your code and documentation is super clean, very easy to understand/follow and matches our coding style and structure really well! 💛

src/Cms/Cors.php
johannschopplich: Yes, you're right! I'm unsure where I got the idea to check on the request header. Fixed in https://github.com/getkirby/kirby/pull/7630/commits/73aa7d039e531c78ab04c8ca91f04b5fb5e1c19a.

src/Cms/Cors.php
johannschopplich: Good to know! I've set up a new protected function `hasConfigValue` – I felt the naming was more semantic. What do you say?

Related: https://github.com/getkirby/kirby/pull/7630/commits/1ae1bf7d8ea41a080ec490b14b992694b85512b7

lukasbestle: That's perfect, thanks! 💛

src/Cms/Cors.php
johannschopplich: Thanks for catching that edge case! I'm leaning towards "secure by default" as well. Especially since the whole feature is opt-in. I've adjusted the implementation as you suggested: https://github.com/getkirby/kirby/pull/7630/commits/a1ec3d1b090e7065d6fb77922fc9c3da4ee2376a.

With this change:
- `allowHeaders => true` now opts into reflecting `Access-Control-Request-Headers`.
- `allowHeaders => []` (or omitting the option) suppresses `Access-Control-Allow-Headers`, so you can explicitly disallow custom headers and stay secure by default.

New tests cover all three modes (explicit list, reflection opt-in, fully disabled).

src/Cms/Cors.php
johannschopplich: Thanks for the feedback. I've added a **Security Considerations** heading to track these concers – Do you think they're enough?

lukasbestle: Thanks, I think that's great. I've added two more sentences to make it even more clear for those without prior CORS experience.

lukasbestle (approved):

lukasbestle:
Thanks again for your amazing work and productive collaboration. I think this is now really a very useful and flexible feature.

PR #7632, opened 2025-10-19
fix: Show placeholder when site title is missing

### LINKED ISSUE #7628: Missing placeholder text on site
## Description
The placeholder text for the site title is missing in site.yml. The pencil icon only becomes visible when you hover over the white space. On the user pages, a placeholder text is displayed in the same place, which improves usability.

**Expected behavior**  
A placeholder `Site ...` should be visible, e.g.
```
<span class="k-header-title-text">
    <span class="k-site-name-placeholder"> Site … </span>
</span>
```

## Your setup
**Kirby Version**  
All versions

distantnative:
Fixes #7628

## Description

- Added back `data-editable` to `k-header` as the user view already had a CSS rule for that. Seems like we removed it overeagerly.
- While the site change title dialog does not allow to remove the site title completely, it is possible via the content file. Without any backfill like e.g. for pages. I think a placeholder like in the user view makes sense here.

### 🐛 Bug fixes

- Shows placeholder in Panel view when site title is missing
#7628
- Fix placeholder color in dark mode in the user view header. 

### For review team

### DIFF
--- a/panel/src/components/Views/Pages/SiteView.vue
+++ b/panel/src/components/Views/Pages/SiteView.vue
@@ -12,7 +12,7 @@
 		>
 			<span
 				v-if="!title || title.length === 0"
-				class="k-site-title-placeholder"
+				class="k-header-title-placeholder"
 			>
 				{{ $t("view.site") }} …
 			</span>
@@ -62,13 +62,6 @@ export default {
 </script>
 
 <style>
-.k-site-title-placeholder {
-	color: var(--color-gray-500);
-	transition: color 0.3s;
-}
-.k-site-view-header[data-editable="true"] .k-site-title-placeholder:hover {
-	color: var(--color-gray-900);
-}
 .k-site-view[data-has-tabs="true"] .k-site-view-header {
 	margin-bottom: 0;
 }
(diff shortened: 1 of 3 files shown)


panel/src/components/Views/Pages/SiteView.vue
```diff
@@ -54,6 +62,13 @@ export default {
 </script>
 
 <style>
+.k-site-title-placeholder {
+	color: var(--color-gray-500);
+	transition: color 0.3s;
+}
+.k-site-view-header[data-editable="true"] .k-site-title-placeholder:hover {
+	color: var(--color-gray-900);
+}
```
bastianallgeier: Since we use it in two places (UserView and SiteView) we might reconsider and add this to the Header component instead. I also wonder if those color values actually work in dark mode. I'm not sure if I ever checked.

bastianallgeier:
@distantnative I've moved the CSS rules for the placeholder into the header component. I decided to keep the old class name for the user name placeholder to avoid any breaking changes though.

bastianallgeier (approved):

distantnative:
👍

Issue #7621, opened 2025-10-09
Confusing password change dialog


tillsanders:
## Description
When trying to change the password of another user through the respective menu option in the user's details view, the dialog asks me to also enter the current password of the user. Needless to say, I do not know ;)

**Expected behavior**  
As an admin, I should be able to change another user's password simply by entering it twice.
(Actually, a password reset feature would be nice, but I'm sure there are many tickets for that idea)

**Screenshots**  

<img width="736" height="588" alt="Image" src="https://github.com/user-attachments/assets/bfe7fe23-978b-4fc5-873d-5468d78adb4a" />
<img width="951" height="751" alt="Image" src="https://github.com/user-attachments/assets/83e7494f-5fda-48f3-baf4-4ce2d90be040" />

## To reproduce

1. Log into the panel as an admin user.
2. Go to the users page.
3. Open another user.
4. Open the menu behind the cog icon and start the password change dialog.

## Your setup
**Kirby Version**  
Observed in 5.0.4 and 5.1.2.
The behaviour in Kirby v4 is as expected.

tillsanders:
After opening this ticket I learned that the dialog is expecting MY password as a security mechanism. Honestly, I didn't get that 😄 From other software, I would expect to only be asked to verify the request with my own password in a separate dialog. Maybe the field could be renamed and moved downwards with a slight visual distinction. I think that would have helped me.

distantnative:
We just discussed the same this week on Discord that apparently users get mistaken by this even though the wording is actually correct - we will look into changing the wording to make it clear right away.

The password reset feature already exists: https://getkirby.com/docs/guide/authentication/password-reset-form

tillsanders:
I agree, the wording is actually correct. I assume the confusion arises because the form currently looks very similar to the universal process of *changing* passwords. If you're only going to change the wording and nothing else, I suggest "Dein eigenes Passwort" and maybe a little help text, too.

bastianallgeier:
Maybe something like this …

<img width="551" height="533" alt="Image" src="https://github.com/user-attachments/assets/5e3cdc85-b113-434d-b20d-c5cf830461f5" />

tillsanders:
In my opinion, this is a lot clearer, thank you! Especially the help text is great. I'm still stumbling a little over the word "current". It suggests that this is the password I'm changing. Maybe it could simply be left out? "Your password" or "Your own password" or "Password for admin@getkirby.com".

distantnative:
[...]

PR #7636, opened 2025-10-20
feat: Clarify change password dialog

bastianallgeier:
### ✨ Enhancements

- Improve the wording and layout of the "Change Password" dialog. https://github.com/getkirby/kirby/issues/7621
Screenshot 2025-10-20 at 13 47 28

### For review team


bastianallgeier:
I'm not entirely sure if we should release this in a patch. The translations change slightly, which could be considered a bigger change as long as translations in other languages are not updated.

i18n/translations/en.json
```diff
@@ -768,7 +769,7 @@
 	"user.changeLanguage": "Change language",
 	"user.changeName": "Rename this user",
 	"user.changePassword": "Change password",
-	"user.changePassword.current": "Your current password",
+	"user.changePassword.current": "Your own password",
```
distantnative: I think we need two different variants here depending on whether editing one own password or another user. For editing another user, this change is great. But when chasing your own password, "current" is actually the better phrasing.

bastianallgeier:
@distantnative it's a bit more effort, but you are right and it would also even help to not turn this into a quasi-breaking-change

distantnative:
@bastianallgeier I have pushed a version, what do you think?

I haven't added a decimated unit test as the test setup in this version, but would add one when merging into `v6/develop` when implementing this into the dialog controller. The dialog controller tests will make this easier.

i18n/translations/en.json
```diff
@@ -769,7 +769,8 @@
 	"user.changeLanguage": "Change language",
 	"user.changeName": "Rename this user",
 	"user.changePassword": "Change password",
-	"user.changePassword.current": "Your own password",
+	"user.changePassword.admin": "Your own password",
```
bastianallgeier: Suggestion: Couldn't we just call it `user.changePassword.own` to not bind it to a role? There could be roles later which are also allowed to change passwords.

distantnative: Changed it

bastianallgeier: Awesome! Thanks :) I'm happy now.

distantnative (approved):

PR #7644, opened 2025-10-25
fix: Dropdown content items with options prop

distantnative:
## Description

No changelog as this fixes a regression/bug in the unreleased `develop-minotr`.

### For review team

### DIFF
--- a/panel/src/components/Dropdowns/DropdownContent.vue
+++ b/panel/src/components/Dropdowns/DropdownContent.vue
@@ -26,7 +26,7 @@
 						<k-dropdown-item
 							:key="'item-' + index"
 							v-bind="option"
-							@click="onOptionClick(option)"
+							:click="() => onOptionClick(option)"
 						>
 							{{ option.label ?? option.text }}
 						</k-dropdown-item>


panel/src/components/Dropdowns/DropdownContent.vue
```diff
@@ -26,7 +26,7 @@
 						<k-dropdown-item
 							:key="'item-' + index"
 							v-bind="option"
-							@click="onOptionClick(option)"
+							:click="() => onOptionClick(option)"
```
distantnative, line 29: `Explanation:` The problem/regression was that the `option` object includes a `click` key itself which we bin to dropdown item and on `develop-minor` pass through to `k-button`. While `k-button` supports a `click` function, in the context of `k-dropdown-content` `option.click` though can also be a string or an object, which is handled by `onOptionClick` in `k-dropdown-content`. But this is why we need to avoid passing it through to `k-button`.

With this change, we overwrite any bound `click` with a call to `onOptionClick`.

bastianallgeier (approved):

PR #7646, opened 2025-10-25
fix: PagePicker parent handling when same as start model

### LINKED ISSUE #7624: Pages field dialog lists unwanted pages after navigating back from subpages
## Description
The pages field dialog ignores the 'filterBy' defined in the query when navigating back from the subpages.

**Expected behavior**  
When navigating forward and backward in the pages field dialog (page <-> subpage), the pages visible should stay the same.

**Content folder**
Image

**1. Dialog not opened yet**
Image

**2. Dialog opened with pages properly filtered**
Image

**3. Dialog when navigating to subpages**
Image

**4. Dialog when navigating back**
Image

## To reproduce on a fresh Plainkit install (5.1.2)

1. Clone the `default.yml` blueprint twice and call it `bug.yml` and `subpage.yml`
2. Edit the default blueprint so its pages section accepts the new `subpage` template
3. Create a new `test.yml` blueprint with a pages sections accepting `default` and `bug` templates
4. Create the `test` page untitled _Test_, then create children with the `default` and `bug` templates  
5. In the new `default` pages, create some children with the `subpage` template
6. Edit the `site.yml` blueprint to:
```
title: Site

sections:
  pages:
    type: pages
  bug:
    type: fields
    fields:
      pages:
        label: Pages
        type: pages
        query: page('test').children.filterBy('intendedTemplate', 'in', ['default', 'subpage'])
```
7. In Site, open the pages field dialog. You should see in the list pages with the `default` template only.
8. Click on a right arrow to navigate to subpages, then click the back arrow.
9. You'll notice now in the list that the page with the `bug` template is now appearing.

afbora:
## Description

Prevents the parent option from being used in PagePicker if it is the same as the start/top-most model. Adds a test to verify correct behavior when query and parent refer to the same page.

### 🐛 Bug fixes

- Pages field dialog lists unwanted pages after navigating back from subpages #7624

### For review team

### DIFF
--- a/src/Cms/PagePicker.php
+++ b/src/Cms/PagePicker.php
@@ -130,7 +130,12 @@ class PagePicker extends Picker
 		// when subpage navigation is enabled, a parent
 		// might be passed in addition to the query.
 		// The parent then takes priority.
-		} elseif ($this->options['subpages'] === true && empty($this->options['parent']) === false) {
+		// Don't use the parent if it's the same as the start/top-most model.
+		} elseif (
+			$this->options['subpages'] === true &&
+			empty($this->options['parent']) === false &&
+			$this->model()->id() !== $this->start()->id()
+		) {
 			$items = $this->itemsForParent();
 
 		// search by query
--- a/tests/Cms/Page/PagePickerTest.php
+++ b/tests/Cms/Page/PagePickerTest.php
@@ -22,9 +22,9 @@ class PagePickerTest extends ModelTestCase
 							[
 								'slug' => 'mother',
 								'children' => [
-									['slug' => 'child-a'],
-									['slug' => 'child-b'],
-									['slug' => 'child-c']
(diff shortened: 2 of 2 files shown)


bastianallgeier:
It looks good to me. Is there a reason why this should not be in 5.1.4?

afbora:
No reason, feel free to move into 5.1.4. But I think 5.2.0 is fine since it's not a critical bug.

bastianallgeier (dismissed):

bastianallgeier:
After testing it some more I moved it to 5.1.4. I think it can actually be quite critical when you use the query to filter pages that the user should not see.

afbora:
> you use the query to filter pages that the user should not see.

I hadn't thought of it that way, you're right 👍Thanks 🙌

bastianallgeier (approved):

PR #7648, opened 2025-10-29
feat: Constant to disable the PHP version check

lukasbestle:
## Description

As discussed, we want to allow developers to temporarily disable the PHP version check e.g. during an upgrade process or for debugging.

I've added the lengthy warning both as a disclaimer and also as a real warning because of the risks of unsupported PHP versions (too old = does not support our use of new PHP features; too new = possible breaking changes in PHP that are not accounted for).

Because of these risks, we will not document the constant but keep it just documented in the code for those who are running into PHP version related issues and have so far needed to comment out the check in the core.

### ✨ Enhancements

- The PHP version check can now temporarily be disabled with a new `KIRBY_PHP_VERSION_CHECK` constant. This is for very specific cases and comes with risks, please note the code comment in `bootstrap.php`.

## Docs

None

### For review team

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -3,10 +3,23 @@
 /**
  * Validate the PHP version to already
  * stop at older or too recent versions
+ *
+ * This check can be disabled by setting:
+ * define('KIRBY_PHP_VERSION_CHECK', false);
+ * in your `index.php`.
+ *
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):
Nice!

PR #7649, opened 2025-10-31
feat: New create option for the files section

bastianallgeier:
### ✨ Enhancements

- A new `create` option for the files section lets you disable the upload button. This is useful if you are using queries in files sections and the upload would not actually show up in the list. It is comparable to the `create` option in the pages section. 
```yaml
sections: 
  images: 
    type: files
    query: page.images.filter('customField', 'customFilter')
    create: false
```

### For review team

### DIFF
--- a/config/sections/files.php
+++ b/config/sections/files.php
@@ -23,6 +23,12 @@ return [
 	],
 	'props' => [
 		/**
+		 * Option to switch off the upload button
+		 */
+		'create' => function (bool $create = true) {
+			return $create;
+		},
+		/**
 		 * Filters pages by a query. Sorting will be disabled
 		 */
 		'query' => function (string|null $query = null) {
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
I wonder if this should also be called `create`. Two reasons:
- Consistency with the pages section. Makes it easier to keep the option names in mind. Also, should we ever want to merge or restructure these sections (especially in the "sections as fields" context), having one naming scheme would reduce the overhead.
- Maybe there could be use cases for creating files directly in the Panel without uploading them. This could then be covered by this option as well.

I know all of these thoughts are looking far ahead, but IMO there is no harm to the `create` name, only possible advantages.

bastianallgeier:
I was thinking about it as well. I'm fine with create. I think the second argument is especially convincing. Let's rename it

distantnative (approved):

PR #7650, opened 2025-10-31
refact: Use item classes for the dropdownOption method

bastianallgeier:
### ♻️ Refactored

- Use the new UI Item classes to create the props for model's dropdown options. 

### For review team

### DIFF
--- a/src/Panel/ChangesDialog.php
+++ b/src/Panel/ChangesDialog.php
@@ -3,7 +3,13 @@
 namespace Kirby\Panel;
 
 use Kirby\Cms\Collection;
+use Kirby\Cms\File;
+use Kirby\Cms\Page;
+use Kirby\Cms\User;
 use Kirby\Content\Changes;
+use Kirby\Panel\Ui\Item\FileItem;
+use Kirby\Panel\Ui\Item\PageItem;
+use Kirby\Panel\Ui\Item\UserItem;
 
 /**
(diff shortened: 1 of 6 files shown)


distantnative:
@bastianallgeier since the changes dialog is the only code using this method, I was wondering if we should rather implement this method in the changes dialog and start depreciating it in the panel model classes. Would be a good chance to do this here.

bastianallgeier:
@distantnative I've added a new ::item method to the ChangesDialog class and deprecated all the dropdownOptions methods.

src/Panel/ChangesDialog.php
```diff
@@ -31,13 +37,27 @@ public function files(): array
...
+
 	/**
 	 * Helper method to return item props for the given models
 	 */
 	public function items(Collection $models): array
 	{
 		return $models->values(
-			fn ($model) => $model->panel()->dropdownOption()
+			fn ($model) => $this->item($model)
 		);
```
distantnative: `Nitpick/Suggestion:` I don't know if you would like that, but this could be also written as

```php
return $models->values($this->item(...));
```

I think.

distantnative (dismissed):

src/Panel/ChangesDialog.php
bastianallgeier: Yep, that works.

PR #7655, opened 2025-11-11
feat: New Str::label method

bastianallgeier:
## Description

This is the first step to improve automatic labels in all relevant parts in blueprints (fields, sections, etc.) when the keys are written in camelCase or kebab_case. It's based on a suggestion from one of the meetup attendees at btconf. 

In the next step, the method needs to be used in various places, but I wanted to separate this to simplify the PR reviews.

### ✨ Enhancements

- New `Str::label($string)` method to convert keys or ids into human-readable labels by normalizing punctuation, splitting camel- or kebab-case, and title-casing the result. It’s very helpful to turn a blueprint field key like `workEmailAddress` into `Work email address`

### For review team

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -601,6 +601,40 @@ class Str
 	}
 
 	/**
+	 * Converts keys or ids into human-readable labels by
+	 * normalizing punctuation, splitting camel- or kebab-case,
+	 * and title-casing the result.
+	 *
+	 * Example: `workEmailAddress` will turn into `Work email address`
+	 *
+	 * @since 5.2.0
+	 */
+	public static function label(string $value): string
+	{
+		// replace punctuation with spaces
+		$value = str_replace(['_', '-', '.'], ' ', $value);
+
+		// add a space before every uppercase character by matching
+		// all characters that are not Unicode lowercase or numbers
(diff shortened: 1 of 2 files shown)


src/Toolkit/Str.php
```diff
@@ -600,6 +600,31 @@ public static function kebabToCamel(string|null $value): string
 		));
 	}
 
+
+	public static function label(string $value): string
+	{
+		// replace punctuation with spaces
+		$value = str_replace(['_', '-', '.'], ' ', $value);
+
+		// add a space before every uppercase character
```
lukasbestle: `Suggestion`: I first had to wrap my head around this RegEx, so for future reference:

```suggestion
		// add a space before every uppercase character by matching
		// all characters that are not Unicode lowercase or numbers
```

src/Toolkit/Str.php
```diff
@@ -600,6 +600,31 @@ public static function kebabToCamel(string|null $value): string
 		));
 	}
 
+
```
lukasbestle: `Suggestion`: I think a docblock here would be good to explain what this method does, especially given its specific steps that work for text with underscores, dashes and dots, but may not work with other kind of input.

lukasbestle (commented):
Very useful indeed!

src/Toolkit/Str.php
bastianallgeier: I've just added one.

distantnative (approved):

PR #7656, opened 2025-11-11
feat: Use Str::label in various places throughout the system

bastianallgeier:
## Merge first

- https://github.com/getkirby/kirby/pull/7655

### ✨ Enhancements

- The new `Str::label()` method is used to create better fallback labels, titles and more in various places of the system. This will often save you from defining additional label/title/name options. This can now help you in the following places: 
  - Blueprint title 
  - Blueprint tab labels
  - Section headlines
  - Form field labels
  - Block field: fieldset names
  - Block field: fieldset tab names
  - Block field: fieldset group names
  - Role titles
  - Custom panel search labels

### For review team

I've split the PR into multiple commits to help with the review process. The field class needed a new `nameUnmodified` property to store the name before lower-casing, otherwise `Str::label()` would have no effect on the lowercase name. This could also help us in the future if we want to improve the mapping of field names with the names in storage. 

### DIFF
--- a/config/sections/mixins/headline.php
+++ b/config/sections/mixins/headline.php
@@ -1,6 +1,7 @@
 <?php
 
 use Kirby\Toolkit\I18n;
+use Kirby\Toolkit\Str;
 
 return [
 	'props' => [
@@ -30,7 +31,7 @@ return [
 				return $this->model()->toString($this->headline);
 			}
 
-			return ucfirst($this->name);
+			return Str::label($this->name);
 		}
 	]
 ];
--- a/src/Cms/Blueprint.php
+++ b/src/Cms/Blueprint.php
@@ -10,6 +10,7 @@ use Kirby\Filesystem\F;
 use Kirby\Form\Field;
 use Kirby\Toolkit\A;
 use Kirby\Toolkit\I18n;
+use Kirby\Toolkit\Str;
 use Throwable;
 
 /**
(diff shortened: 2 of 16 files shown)


src/Form/Field/EntriesField.php
```diff
@@ -198,7 +198,7 @@ public function validations(): array
 							throw new InvalidArgumentException(
 								key: 'entries.validation',
 								data: [
-									'field' => $this->label() ?? Str::ucfirst($this->name()),
+									'field' => $this->label() ?? Str::label($this->name()),
```
distantnative: Does it make sense to transform the field name here? I think for the error message the actual blueprint key might be more helpful, no? I think even the ucfirst before was maybe not needed?

src/Form/FieldClass.php
```diff
@@ -168,7 +169,8 @@ protected function setDisabled(bool $disabled = false): void
 
 	protected function setName(string|null $name = null): void
 	{
-		$this->name = strtolower($name ?? $this->type());
+		$this->nameUnmodified = $name ?? $this->type();
+		$this->name           = strtolower($this->nameUnmodified);
```
distantnative: `Question:` Would it make more sense to keep the unmodified one as `name` and only apply `strtolower` in the actual places where we need it lowercased?

src/Form/Field/EntriesField.php
bastianallgeier: Good point. I removed it.

src/Form/FieldClass.php
bastianallgeier: I've removed this as well. In hindsight, it's quite a big change that needs more planning and we should probably think more about it while working on the fields anyway.

distantnative (approved):

Issue #7641, opened 2025-10-22
Field options query executes 3 to 5 times on panel page load, making it very slow


SkidX:
## Description
When there are fields with options (tags, select...) that use a dynamic query, the query is executed multiple times (3 to 5, see below) when the page loads in the panel, making the panel page load potentially very slow.

I'm aware each fields' section gets also loaded dynamically with an ajax call, once the main page has rendered, but that's not what I'm referring to. The actual main page preparation calls the query multiple times for each field.

**How I found out the issue**
On a client project I have an actual panel page taking ~ 5 secs. to load. Initially I suspected it was related to content blocks, but after removing all of them the page was still very slow, above 3 - 4 seconds. I isolated the issue as related to 4 tags fields that have dynamic options (a kirby query that takes data from a structure field on a different content page). Since they all use the same query, I replaced the pure query with a call to a custom method that use memoization in order to run the query only on the first call and return the cached values on the next calls. This allowed to save ~2 seconds but the page was still slow. I put some logging in the method and found out it was called 20 times! I reduced the amount of tag fields from 4 to 1 and verified the method was still called 5 times. 

**Expected behavior**  
I would expect each options query to be called only once in the page preparation, in order to not slow down the page load with extra not needed work. 

## To reproduce
Here I'm making a simplified demo in order to reproduce the issue.

A test page blueprint (test.yaml) :
```yaml
title: Test page
fields:
    tags:
        type: tags
        label: Tags
        accept: options
        translate: false
        options:
            type: query
            query: page.getTagsValues
```

The custom page model test.php
```php 
class TestPage extends Page
{
	protected static bool $first = true;

	public function getTagsValues()
	{
		// I want to increment the counter only for calls happening during the main page rendering,
		// not in the section ajax calls
		$sectionCall = Url::current() !== $this->panel()->url();
		if (!$sectionCall) {
			// main page rendering
			if (static::$first) {
				static::$first = false;
				// on the first call we initialize the counter and store it in a log file
				file_put_contents($this->root() . '/counter.log', '1');
				$counter = 1;
			} else {
				// on next calls we read the counter from the log file and increment it
				$counter = (int) file_get_contents($this->root() . '/counter.log');
				$counter++;
				file_put_contents($this->root() . '/counter.log', (string) $counter);
			}
		} else {
			// on section's ajax call we just read the current value of the counter from the log file
			$counter = (int) file_get_contents($this->root() . '/counter.log');
		}
		return [
[...]

PR #7658, opened 2025-11-12
fix: Form performance improvements

bastianallgeier:
## Chunks: 

- [ ] https://github.com/getkirby/kirby/pull/7663
- [ ] https://github.com/getkirby/kirby/pull/7664
- [ ] https://github.com/getkirby/kirby/pull/7665

## Description

After looking closer at this issue and @afbora research https://github.com/getkirby/kirby/issues/7641 I found multiple ways how we can improve/fix this. We have to decide if we do all the steps at once, or if we want to break them down, but I'm using this PR as a big sweep first to give you the bigger picture. 

## Benchmark results

To not step in the dark with premature improvements, I've created a little benchmark setup for the sandbox with a new CLI command that can help us test our changes more easily. 

The results of this PR are impressive. I'm benchmarking two scenarios. 

### `Form::reset()` 

To reset all form fields took forever before the PR because the old Field class needed to rerun all prop and computed prop methods to make sure that there's not an invalid state for one of them which depends on the value. 

I'm running this 10 times in a loop. 

```php
foreach (page('fields')->children() as $fieldPage) {
  Form::for($fieldPage)->reset();
}
```

**Before**
```
Bench: Time: 00:11.542, Memory: 24.00 MB
```

**After (2 times faster)**
```
Bench: Time: 00:05.704, Memory: 24.00 MB
```

### `$page->panel()->view()`

Creating all the props for the view needs to create two forms: one for the latest version and one for the changed version (if it exists) This makes the reset problem from above even worse because we need to reset the form when we want to reuse it for two versions and this could also add up even more if there are nested forms (structures, blocks, etc.)

Again, I'm running this 10 times in a loop. 

```php
foreach (page('fields')->children() as $fieldPage) {
  $fieldPage->panel()->view();
}
```

**Before**
```
Bench: Time: 00:16.887, Memory: 24.00 MB
```

**After (3,2 times faster)**
```
Bench: Time: 00:05.832, Memory: 24.00 MB
```

This clearly shows how the problem escalates with more form setup and reset tasks. 

This should have positive performance side-effects to pretty much any part of the panel. 

## Running benchmarks

To test the differences, you can use the Kirby CLI in the sandbox root directory. Make sure to pull the latest sandbox changes. There's a new way to set the sandbox URL with a new custom sandbox config. In the root directory, you'll find a new sandbox.config.sample.php. Rename it to sandbox.config.php and add your custom URL in there. All options in the file will be merged with the Kirby config of the sandbox. The sandbox.config.php is automatically ignored. 

Once this is done, you can run

```
kirby sandbox:bench:fields
```

… and …

``` 
kirby sandbox:bench:views
```

### ✨ Enhancements

- Cache options in the options mixin for fields.
- Cache block forms by type in the blocks field. 
- Cache columns in the structure field
- Cache color options in the color field
- Cleaned up object field with the new form methods and caching, similar to the structure field. 
- Improve type hinting in the options mixin for fields
- New `::fillWithEmptyValue` method for Field and FieldClass. This is used in `Fields::reset` to reset the value to a correct empty one. The naming is not great, but we cannot use reset, as it exists as option in the toggles field and would collide. It's at least clear enough. This will make sure that resetting a field will be as fast as possible without reevaluating props and computed props in the old rusty Field component. This is one big source for performance hits. 
- New `Field::handlerExists` and `Field::handlerCall` methods. Those are not directly related to this PR, but simply help to remove some redundant code. We can totally do this in a separate PR. 

### ♻️ Refactored

- Use Form instead of the Fields class in Kirby\Panel\Model
- Remove an unused class declaration in the Form class

### For review team

### DIFF
--- a/config/fields/number.php
+++ b/config/fields/number.php
@@ -48,7 +48,7 @@ return [
 			return $this->toNumber($default) ?? '';
 		}
 	],
-	'fillWithEmptyValue' => function (): static {
+	'fillWithEmptyValue' => function () {
 		$this->value = '';
 		return $this;
 	},
--- a/config/fields/number.php
+++ b/config/fields/number.php
@@ -48,6 +48,10 @@ return [
 			return $this->toNumber($default) ?? '';
 		}
 	],
+	'fillWithEmptyValue' => function (): static {
+		$this->value = '';
+		return $this;
+	},
 	'methods' => [
 		'toNumber' => function ($value): float|null {
 			if ($this->isEmptyValue($value) === true) {
--- a/config/fields/number.php
+++ b/config/fields/number.php
@@ -30,11 +30,11 @@ return [
 		'step' => function ($step = null): float|string {
 			return match ($step) {
 				'any'   => 'any',
-				default => $this->toNumber($step) ?? ''
+				default => $this->toNumber($step) ?? $this->emptyValue()
 			};
 		},
 		'value' => function ($value = null) {
-			return $this->toNumber($value) ?? '';
+			return $this->toNumber($value) ?? $this->emptyValue();
 		}
 	],
 	'computed' => [
@@ -45,12 +45,11 @@ return [
 				$default = $this->model()->toString($default);
 			}
 
-			return $this->toNumber($default) ?? '';
+			return $this->toNumber($default) ?? $this->emptyValue();
 		}
 	],
-	'fillWithEmptyValue' => function () {
-		$this->value = '';
-		return $this;
+	'emptyValue' => function () {
+		return '';
 	},
 	'methods' => [
 		'toNumber' => function ($value): float|null {
--- a/src/Form/Field.php
+++ b/src/Form/Field.php
@@ -297,6 +297,21 @@ class Field extends Component
 	}
 
 	/**
+	 * Preferred name would be `::reset` but this is
+	 * taken by options in other fields.
+	 *
+	 * @since 5.2.0
+	 */
+	public function fillWithEmptyValue(): static
+	{
+		if ($this->handlerExists('fillWithEmptyValue') === true) {
+			return $this->handlerCall('fillWithEmptyValue');
+		}
+
+		return $this->fill(null);
+	}
+
+	/**
 	 * @deprecated 5.0.0 Use `::siblings() instead
 	 */
 	public function formFields(): Fields
--- a/src/Form/Field.php
+++ b/src/Form/Field.php
@@ -243,6 +243,15 @@ class Field extends Component
 		return [];
 	}
 
+	public function emptyValue(): mixed
+	{
+		if ($this->handlerExists('emptyValue') === true) {
+			return $this->handlerCall('emptyValue');
+		}
+
+		return null;
+	}
+
 	/**
 	 * Creates a new field instance
 	 */
@@ -302,7 +311,8 @@ class Field extends Component
 			return $this->handlerCall('fillWithEmptyValue');
 		}
 
-		return $this->fill(null);
+		$this->value = $this->emptyValue();
+		return $this;
 	}
 
 	/**
--- a/src/Form/Field.php
+++ b/src/Form/Field.php
@@ -224,11 +224,8 @@ class Field extends Component
 	 */
 	public function dialogs(): array
 	{
-		if (
-			isset($this->options['dialogs']) === true &&
-			$this->options['dialogs'] instanceof Closure
-		) {
-			return $this->options['dialogs']->call($this);
+		if ($this->handlerExists('dialogs') === true) {
+			return $this->handlerCall('dialogs');
 		}
 
 		return [];
@@ -239,11 +236,8 @@ class Field extends Component
 	 */
 	public function drawers(): array
 	{
-		if (
-			isset($this->options['drawers']) === true &&
-			$this->options['drawers'] instanceof Closure
-		) {
-			return $this->options['drawers']->call($this);
+		if ($this->handlerExists('drawers') === true) {
+			return $this->handlerCall('drawers');
 		}
 
 		return [];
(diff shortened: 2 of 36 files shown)


config/fields/number.php
```diff
@@ -48,6 +48,10 @@
 			return $this->toNumber($default) ?? '';
 		}
 	],
+	'fillWithEmptyValue' => function () {
+		$this->value = '';
+		return $this;
+	},
```
afbora: Regarding the new fillWithEmptyValue callback:

We also use the default empty value for fields in different places in the code. For Source of Truth, a more systematic improvement would be nice, both for this improvement and for maintaining the default empty value for fields.

For example for number field; there are four times to set empty string. Also same behavior for other fields:

- https://github.com/getkirby/kirby/blob/fix/form-performance/config/fields/number.php#L33
- https://github.com/getkirby/kirby/blob/fix/form-performance/config/fields/number.php#L37
- https://github.com/getkirby/kirby/blob/fix/form-performance/config/fields/number.php#L48
- https://github.com/getkirby/kirby/blob/fix/form-performance/config/fields/number.php#L52


What do you think?

bastianallgeier: We could indeed also think about having a method that's called `emptyValue` which returns the empty state. The fillWithEmptyValue method could then use that as well as all those other places you mention. I will give that a try.

bastianallgeier: Ok, I've implemented a first version of that. Unfortunately, our old fields are so bad that we cannot use it everywhere. All fields that use the options mixin cannot access the emptyValue method properly because of prop colissions. This is a seperarate problem though.

distantnative:
@bastianallgeier I am having a bit of a hard time following all the different changes in this PR, so apologies if I haven't quite understood all the aspects. But if I look at the changes necessary in all the different options fields, isn't this then a breaking change for all plugin fields using the options mixin etc.? Or custom fields now would need to define an `emptyValue`?

src/Form/Field.php
```diff
@@ -296,6 +299,22 @@ public function fill(mixed $value): static
...
+	 * Preferred name would be `::reset` but this is
+	 * taken by options in other fields.
+	 *
+	 * @since 5.2.0
+	 */
+	public function fillWithEmptyValue(): static
+	{
+		if ($this->handlerExists('fillWithEmptyValue') === true) {
+			return $this->handlerCall('fillWithEmptyValue');
+		}
```
distantnative, line 312: `Question`: Where do we ever have this? Or is this method now unnecessary and should be replaced with a simple `$field->fill($field->emptyValue())`?

src/Form/Field.php
```diff
@@ -304,6 +323,16 @@ public function formFields(): Fields
...
 
+	protected function handlerCall(string $handler, ...$args): mixed
+	{
+		return $this->options[$handler]->call($this, ...$args);
+	}
+
+	protected function handlerExists(string $handler): bool
+	{
+		return isset($this->options[$handler]) === true && $this->options[$handler] instanceof Closure;
+	}
```
distantnative, line 334: `Feedback:` Personally, to me this optimization makes the code harder to read and understand while not keeping it that much more DRY. For me, it would be easier to read if we leave the actual checks and calls in the methods, but maybe it's different for you and/or others.

bastianallgeier:
It's quite much. I can really break it down more. But I kind of also wanted to use this PR to verify that the performance gains are actually there and as big as in my tests. 

The fields don't have to implement the emptyValue method. It's a performance enhancement that is helping to make things faster, but not necessary for custom fields. When your custom field does not use a lot of computed props, you won't see major performance benefits anyway. The worst offenders are fields that compute heavy stuff like the options. 

You also have to keep in mind that this issue is only related to the Field class. Anything that's built on top of the FieldClass class, does not have those performance issues. This is where we are heading anyway.

src/Form/Field/BlocksField.php
```diff
@@ -77,6 +72,11 @@ public function blocksToValues(
 		return $result;
 	}
 
+	public function emptyValue(): mixed
+	{
+		return [];
+	}
```
distantnative, line 78: `Suggestion:` I do understand that we need to add `emptyValue` as attr to the array-notation fields, what feels a bit like a hack but probably no better way for these. Though here in the class-based fields, I don't know... doesn't come across as a super satisfying way of doing this. Also, if we look ahead to named-parameter constructors, wouldn't it be better/cleaner to define the empty value via the default value of the `value` property?

src/Form/Field/BlocksField.php
```diff
@@ -110,6 +115,10 @@ public function fieldsetGroups(): array|null
 	 */
 	public function fill(mixed $value): static
 	{
+		if ($this->isEmptyValue($value) === true) {
+			return $this->fillWithEmptyValue();
```
distantnative, line 119:
```suggestion
			return $this->fill($value);
```

distantnative:
@bastianallgeier But if a custom field uses the `options` mixin and has a `options` prop defined, it will not reset the `optionsCache` as this PR does for the core fields (and thus seems required for a field to properly work with these changes). No?

src/Form/Fields.php
```diff
@@ -268,7 +268,7 @@ public function reset(): static
 
 		// reset the values of each field
 		foreach ($this->data as $field) {
-			$field->fill(null);
+			$field->fillWithEmptyValue();
```
distantnative, line 271: `Suggestion:` Maybe instead of `$field->fillWithEmptyValue()` this should rather be `$field->reset()`.

src/Form/Field.php
bastianallgeier: I will leave this out in the process of creating the individual chunks.

src/Form/Field.php
bastianallgeier: See above. I will also remove this again in the individual chunks. You are right that it is not necessary.

src/Form/Field/BlocksField.php
bastianallgeier: We could only handle this with a reflection if we want to get default value of the value prop. I'm not sure if that is worth it to be honest.

src/Form/Field/BlocksField.php
bastianallgeier: This would lead to an endless loop.

src/Form/Fields.php
bastianallgeier: See my comments in the #7663 description. Reset is unfortunately still not an option because of the toggles field :(

PR #7659, opened 2025-11-12
feat: New batch mode for the checkboxes field

bastianallgeier:
#### New batch select mode for the checkboxes field

You can now switch on the `select/deselect all` toggle with the new `batch` option in the checkboxes field to select or deselect multiple checkboxes at once. This is helpful when a checkboxes field has a lot of options https://feedback.getkirby.com/81 

```yaml
fields: 
  countries: 
    type: checkboxes
    options: 
      # lots of countries 
    batch: true
```

https://github.com/user-attachments/assets/dacd55db-0496-4e53-a903-38d67ebf4c70

- New `select-all` and `deselect-all` icons
- New `select.all`, `deselect.all` and `deselect` translation strings

### For review team

- [x] Add lab and/or sandbox examples (wherever helpful) https://github.com/getkirby/sandbox/pull/22


distantnative:
@bastianallgeier have you checked if this will also work with v6/develop given that we refactored the batch feature internally quite a bit there?

bastianallgeier:
@distantnative it's only introducing the same batch prop. The logic is quite different and a lot more simple because we don't have to switch between modes.

nilshoerrmann:
Thanks for this fast feature addition! ❤️

Just a quick feedback on the UI:

I looked at the still image before playing the video and neither did I understand that the box in the upper right corner was a checkbox nor did I expect what it was related to. I actually was confused that it was releated to this feature at all. Yes, this is a common interface pattern for interactive tables but usually all checkboxes are aligned in a column in that case. In general, I think this needs labeled buttons and it needs separate buttons for both selecting and deselecting: with a few items selected, I might want to select or deselect all depending on the context.

bastianallgeier:
@nilshoerrmann you are not wrong. Here's another version. 


https://github.com/user-attachments/assets/c581c242-a872-4546-b0ab-94413bcc27d5

nilshoerrmann:
That looks very good. Thanks Bastian!

panel/src/components/Forms/Field/CheckboxesField.vue
```diff
@@ -6,6 +6,38 @@
...
+					:responsive="true"
+					icon="deselect-all"
+					size="xs"
+					variant="filled"
+					@click="deselectAll"
+				>
+					{{ $t("deselect.all") }}
+				</k-button>
+				<k-button
+					:disabled="value.length === options.length"
```
distantnative:
```suggestion
					:disabled="value.length >= options.length"
```

I know not a likely scenario, but maybe if the content file has still 4 values, but in the blueprint all but 2 options are removed.

distantnative (dismissed):
Just one small change I think would be safer.

distantnative (approved):

afbora:
Sorry for late joining to the party but I love it ❤️ Looks and works great 🎉

PR #7663, opened 2025-11-14
refact: New ::emptyValue and ::fillWithEmptyValue methods for Field and FieldClass

bastianallgeier:
## Description

This PR extracts a big portion from https://github.com/getkirby/kirby/pull/7658. It seems that this alone has already major positive performance implications. 

### Before & After

```
kirby sandbox:bench:fields - 00:06.198 -> 00:03.855
kirby sandbox:bench:views - 00:19.390 -> 00:11.257
```

## Context for reviewers

- I still had to stick with `::fillWithEmptyValue` even for the `FieldClass`. We use the Value mixin in the old Field class as well, so using `::reset` would already lead to the collision with the toggles field. 
- I moved `emptyValue` into the methods array for field components, which is lot nicer than defining the method outside of it. (see @distantnative's comments in the review of the original PR)

### ✨ Enhancements

- New `::emptyValue()` method for `Kirby\Form\Field` and `Kirby\Form\FieldClass`, which can be overwritten to define the preferred empty value when the field is being reset
- New `::fillWithEmptyValue()` method for `Kirby\Form\Field` and `Kirby\Form\FieldClass`, which is used in `Kirby\Form\Fields` to reset field values without evaluating computed props again if not necessary. This is the part that leads to the performance improvements. 

### For review team

### DIFF
--- a/src/Form/Field/BlocksField.php
+++ b/src/Form/Field/BlocksField.php
@@ -77,6 +77,11 @@ class BlocksField extends FieldClass
 		return $result;
 	}
 
+	public function emptyValue(): mixed
+	{
+		return [];
+	}
+
 	public function fields(string $type): array
 	{
 		return $this->fieldset($type)->fields();
--- a/src/Form/Field/BlocksField.php
+++ b/src/Form/Field/BlocksField.php
@@ -77,11 +77,6 @@ class BlocksField extends FieldClass
 		return $result;
 	}
 
-	public function emptyValue(): mixed
-	{
-		return [];
-	}
-
 	public function fields(string $type): array
 	{
 		return $this->fieldset($type)->fields();
--- a/src/Form/Field/EntriesField.php
+++ b/src/Form/Field/EntriesField.php
@@ -43,6 +43,11 @@ class EntriesField extends FieldClass
 		$this->setSortable($params['sortable'] ?? true);
 	}
 
+	public function emptyValue(): mixed
+	{
+		return [];
+	}
+
 	public function field(): array
 	{
 		return $this->field;
@@ -133,7 +138,7 @@ class EntriesField extends FieldClass
 	public function toFormValue(): mixed
 	{
 		$form  = $this->form();
-		$value = parent::toFormValue() ?? [];
+		$value = parent::toFormValue() ?? $this->emptyValue();
 
 		return A::map(
 			$value,
--- a/src/Form/Field/EntriesField.php
+++ b/src/Form/Field/EntriesField.php
@@ -31,6 +31,7 @@ class EntriesField extends FieldClass
 	protected array $field;
 	protected Form $form;
 	protected bool  $sortable = true;
+	protected mixed $value = [];
 
 	public function __construct(array $params = [])
 	{
@@ -41,11 +42,6 @@ class EntriesField extends FieldClass
 		$this->setMax($params['max'] ?? null);
 		$this->setMin($params['min'] ?? null);
 		$this->setSortable($params['sortable'] ?? true);
-	}
-
-	public function emptyValue(): mixed
-	{
-		return [];
 	}
 
 	public function field(): array
--- a/src/Form/Fields.php
+++ b/src/Form/Fields.php
@@ -268,7 +268,11 @@ class Fields extends Collection
 
 		// reset the values of each field
 		foreach ($this->data as $field) {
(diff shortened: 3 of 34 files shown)


src/Form/Field/BlocksField.php
```diff
@@ -77,6 +77,11 @@ public function blocksToValues(
 		return $result;
 	}
 
+	public function emptyValue(): mixed
+	{
+		return [];
+	}
+
```
distantnative: `Suggestion:` Instead of defining these in each field class in two ways (as property default and as emptyValue method), we could add just one method to the `FieldClass` class (or Value mixin). Something along the lines of

```php
public function emptyValue(): mixed
{
    return (new ReflectionProperty('value'))->getDefaultValue();
}
```

We need to test what performance effect this has due to the reflection. But if the performance impact is small, I think this would be the better way instead of having to maintain the empty value of a field in two different places within the class.

src/Form/Fields.php
```diff
@@ -268,7 +268,7 @@ public function reset(): static
 
 		// reset the values of each field
 		foreach ($this->data as $field) {
-			$field->fill(null);
+			$field->fillWithEmptyValue();
```
distantnative: `Suggestion:`

```suggestion
			if ($field instance of Field) {
						$field->fillWithEmptyValue();
			} else {
						$field->reset();
			}
```

src/Form/Mixin/Value.php
```diff
@@ -60,6 +68,18 @@ public function fill(mixed $value): static
...
+	 * taken by options in other fields.
+	 *
+	 * @since 5.2.0
+	 */
+	public function fillWithEmptyValue(): static
+	{
+		$this->value = $this->emptyValue();
+		return $this;
+	}
+
```
distantnative: `Feedback:` I think it is more important to look forward what method signatures/names etc. we ideally would like to use in the future for our classes. IMO `::reset()` would be the much better option. `::fillWithEmptyValue()` is a bit quirky, but I understand why it makes a lot of sense for the array-based fields. You mention in the description that we cannot use `::reset()` because the Value mixing is used for both classes. IMO we should not sacrifice then just for the "uniform-ness" of using the mixing for both classes to be stuck with a non-ideal name in the future. If both field classes need to diverge because we need to treat them differently (`FieldClass` how we want it to be in the future, `Field` how it needs to be to work around the array-based weirdness), then let's add `::fillWithEmptyValue()` to `Field` and `::reset()` to `FieldClass` and once the old fields are gone, we can move `::reset()` to the Value mixing.

src/Form/Fields.php
bastianallgeier: I've added some more context about that in the description. We can unfortunately not use reset in the FieldClass either.

src/Form/Field/EntriesField.php
```diff
@@ -133,7 +138,7 @@ public function supports(): array
 	public function toFormValue(): mixed
 	{
 		$form  = $this->form();
-		$value = parent::toFormValue() ?? [];
+		$value = parent::toFormValue() ?? $this->emptyValue();
```
distantnative, line 137: `Question:` Seeing this, I was wondering if `::toFormValue()` should not return `null` but `::emptyValue()` in general?

src/Form/Field/BlocksField.php
distantnative: See https://github.com/getkirby/kirby/pull/7666/commits/ed64ba40d7c956484cc4733a4573c74f2fcf6e05 - seems to work for me without performance issues.

src/Form/Fields.php
distantnative: See https://github.com/getkirby/kirby/pull/7666/commits/69db4e3ecdf2ad1a1cdbebc22cca1ad880110779

src/Form/Mixin/Value.php
distantnative: See https://github.com/getkirby/kirby/pull/7666/commits/69db4e3ecdf2ad1a1cdbebc22cca1ad880110779

src/Form/Field/EntriesField.php
bastianallgeier: Sounds right, but my gut feeling is that it could lead to some side-effects with our forms that are hard to see. We should try to get there, but with more time and tests.

distantnative (approved):

PR #7666, opened 2025-11-16
Forms/empty value nico

distantnative:


src/Form/Mixin/Value.php
```diff
@@ -56,7 +57,7 @@ public function default(): mixed
 	 */
 	public function emptyValue(): mixed
 	{
-		return null;
+		return (new ReflectionProperty($this, 'value'))->getDefaultValue();
```
distantnative, line 60: I had this cached in a `emptyValue` prop as well but showed no difference in the sandbox benchmarks than returning it uncached.

bastianallgeier (approved):

PR #7667, opened 2025-11-16
test: Remove PHPBench

distantnative:
## Description

I think the Kirby CLI bench setup in the sandbox will serve us better currently for our purposes. Suggesting to remove PHPBench from the core repo therefore.

### 🧹 Housekeeping

- Removed PHPBench tests in favor of Sandbox CLI bench setup

### For review team


bastianallgeier:
Should we create the same or a similar setup in the sandbox then?

distantnative:
@bastianallgeier Eventually probably yes, but I thought makes probably more sense when we are actually trying to tackle the performance of that Dir method again (and then knowing better what specific test makes most sense).

bastianallgeier (approved):

PR #7670, opened 2025-11-19
feat: Error lab examples & fixes

bastianallgeier:
## Description

We have a couple of inconsistencies in the way that errors are handled in various parts of the panel. In order to tackle this, I started adding a first set of lab examples and already found the first issues and fixes. 

### ✨ Enhancements

- New error handling examples in the lab (/panel/lab/internals/errors) for various scenarios (opening views, dialogs, drawers, sending requests, etc.)

### 🐛 Bug fixes

- The RequestError class is polyfilling error messages stored in the error prop instead of the message prop to create more consistency. 
- The RequestError class is now responsible for all server errors to add another layer of consistency there. 
- The `details` array from custom Kirby Exceptions is now always properly passed forward in all error responses. 
- The `panel.notification.error()` method is now also passing details forward to the error object. 

### For review team

### DIFF
--- a/config/areas/lab/dialogs.php
+++ b/config/areas/lab/dialogs.php
@@ -1,0 +1,10 @@
+<?php
+
+use Kirby\Panel\Lab\Responses;
+
+return [
+	'lab.errors' => [
+		'pattern' => 'lab/errors/(:any?)',
+		'load'    => fn (string|null $type = null) => Responses::errorResponseByType($type)
+	],
+];
--- a/config/areas/lab.php
+++ b/config/areas/lab.php
@@ -2,10 +2,12 @@
 
 return function () {
 	return [
-		'icon'    => 'lab',
-		'label'   => 'Lab',
-		'menu'    => false,
-		'drawers' => require __DIR__ . '/lab/drawers.php',
-		'views'   => require __DIR__ . '/lab/views.php'
+		'icon'     => 'lab',
+		'label'    => 'Lab',
+		'menu'     => false,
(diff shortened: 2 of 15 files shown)


config/areas/lab/dialogs.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+use Kirby\Panel\Lab\Responses;
+
+return [
+	'lab.errors' => [
+		'pattern' => 'lab/errors/(:any?)',
+		'load'    => fn (string|null $type = null) => Responses::errorResponseByType($type)
```
distantnative, line 8: `Nitpick:`

Only if you like that style too

```suggestion
		'load'    => Responses::errorResponseByType(...)
```

bastianallgeier: I tried that but it doesn't work. It seems to be trying to bind $this for some which leads to an error.

distantnative: @bastianallgeier Indeed getting the same error. I have trouble debugging it as our Panel does swallow the PHP details/stack trace of the exception/error completely. This might not be caused by this PR, but since this PR kind of addresses these problems around errors in the Panel, could we improve this as well? Those PHP error messages are often worthless without more details.

bastianallgeier: As a next step, I really want to build a better error dialog with the option to debug errors with more details. But I think we should get this done first and move on from here.

distantnative: Ok

distantnative (approved):

PR #7674, opened 2025-11-20
refact!: Validators as regular `V` methods

distantnative:
### ♻️ Refactored

- Implemented default validators as regular class methods of `Kirby\Toolkit\V`

### 🚨 Breaking changes

- Custom validators cannot overwrite default validators from the `Kirby\Toolkit\V` class any longer.

### For review team


bastianallgeier (approved):
I love this!

PR #7679, opened 2025-11-21
(Fields PR 1) refact: New FieldClass mixins

bastianallgeier:
## Description

This PR extracts more properties from field classes and moves them into mixins. This is a first step to create an extensive mixin library, which will be useful for all fields. 

## Changelog 

- Removed all default values for mixin properties 
- Allowed null for all bool mixin properties 

### ♻️ Refactored

- New Kirby\Form\Mixin\DefaultValue mixin 
- New Kirby\Form\Mixin\Disabled trait
- New Kirby\Form\Mixin\Name trait
- New Kirby\Form\Mixin\Pretty trait
- New Kirby\Form\Mixin\Required mixin
- New Kirby\Form\Mixin\Sortable trait

### 🚨 Breaking changes

- All mixin props no longer define a default value. The default value is now defined by the getter when the property value is null. 
- Setter methods of FieldClass mixins no longer have a default value. The default value is worthless in the constructor, as we never call a setter without argument. It just increases the redundancy in the code. The following `set` methods are affected.
  - `::setAfter()`
  - `::setAutofocus()`
  - `::setBefore()`
  - `::setDefault()`
  - `::setDisabled()`
  - `::setEmpty()`
  - `::setHelp()`
  - `::setIcon()`
  - `::setLabel()`
  - `::setMax()`
  - `::setMin()`
  - `::setName()`
  - `::setPlaceholder()`
  - `::setPretty()`
  - `::setRequired()`
  - `::setSortable()`
  - `::setTranslate()`
  - `::setWhen()`
  - `::setWidth()`
- All translatable properties in FieldClass mixins are no longer translating the value in the setter but in the getter. This is a cleaner approach, where we store the property as untouched as possible and then compute any changes in the getter if needed. This way, we can always access the originally passed value for the property and we can also reduce the amount of computing that is done when the instance is created. The following props are affected:
  - `after`
  - `before`
  - `empty`
  - `help`
  - `label`
  - `placeholder`
- Lower-casing of the name property of the FieldClass has been moved from the setter to the getter. 

### For review team

### DIFF
--- a/src/Form/Mixin/Help.php
+++ b/src/Form/Mixin/Help.php
@@ -11,7 +11,7 @@ trait Help
 
 	public function help(): string|null
 	{
-		if ($this->help !== null && $this->help !== []) {
+		if ($this->help !== null && $this->help !== [] && $this->help !== '') {
 			$help = $this->i18n($this->help);
 			$help = $this->stringTemplate($help);
 			$help = $this->kirby()->kirbytext($help);
--- a/src/Form/Mixin/Help.php
+++ b/src/Form/Mixin/Help.php
@@ -7,7 +7,7 @@ trait Help
 	/**
 	 * Optional help text below the field
 	 */
-	protected array|string|null $help = null;
+	protected array|string|null $help;
 
 	public function help(): string|null
 	{
--- a/src/Form/Mixin/Help.php
+++ b/src/Form/Mixin/Help.php
@@ -14,12 +14,13 @@ trait Help
 	/**
 	 * Optional help text below the field
 	 */
-	protected string|null $help;
+	protected array|string|null $help = null;
 
 	public function help(): string|null
 	{
-		if (empty($this->help) === false) {
-			$help = $this->stringTemplate($this->help);
+		if ($this->help !== null && $this->help !== []) {
+			$help = $this->i18n($this->help);
+			$help = $this->stringTemplate($help);
 			$help = $this->kirby()->kirbytext($help);
 			return $help;
 		}
@@ -27,8 +28,8 @@ trait Help
 		return null;
 	}
 
-	protected function setHelp(array|string|null $help = null): void
+	protected function setHelp(array|string|null $help): void
 	{
-		$this->help = $this->i18n($help);
+		$this->help = $help;
 	}
 }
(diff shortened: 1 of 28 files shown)


src/Form/Mixin/Help.php
```diff
@@ -2,33 +2,27 @@
...
 	 * Optional help text below the field
 	 */
-	protected string|null $help;
+	protected array|string|null $help = null;
 
 	public function help(): string|null
 	{
-		if (empty($this->help) === false) {
-			$help = $this->stringTemplate($this->help);
+		if ($this->help !== null && $this->help !== []) {
```
distantnative: `Question:` What about an empty string?

src/Form/Mixin/Label.php
```diff
@@ -4,29 +4,24 @@
...
-	protected string|null $label;
+	protected array|string|null $label = null;
 
 	public function label(): string|null
 	{
-		return $this->stringTemplate(
-			$this->label ?? Str::ucfirst($this->name())
-		);
+		if ($this->label === null || $this->label === []) {
+			return Str::ucfirst($this->name());
```
distantnative, line 17: `Question:` Why not use `Str::label()` here?

distantnative:
@bastianallgeier Many of your commit messages contain the `!` for breaking changes, but the changelog above doesn't list any?

src/Form/Mixin/Label.php
bastianallgeier: We haven't solved the problem of lowercase names yet. That's why Str::label would not be very effective so far.

distantnative:
@bastianallgeier 
> Default values in FieldClass mixins are now set for all properties. We used to set a default value in some, but not all which is not very consistent. We could think about removing any kind of default value here as well, but then we would need to make sure that all properties are always initiated or have isset checks in their getters.

Not sure I understand you here. It seems still the case that some properties have a default (mainly the boolean ones, e.g. `required`) while others are `null` by default and receive their default fallback in their getter method (e.g. `::width()`).

bastianallgeier:
@distantnative I mean that they have any value, including null. We had quite a couple properties which have not been set at all before. For width, I would indeed argue that we might actually want to change it to never accept null at all and set "1/1" as default.

distantnative (approved):
I think this is good to go. Just wondering if we should merge it into `v6/develop` directly or keep a new base branch for all field improvements `v6/fields`.

bastianallgeier:
I like the idea to keep it in a fields branch for a little longer.

PR #7680, opened 2025-11-21
(Field PR) feat: More mixins 

bastianallgeier:
## Merge first

## Description

After the mixin refactoring in https://github.com/getkirby/kirby/pull/7679, this PR creates more mixins from existing array-definition fields. 

## Code Coverage

We did not test mixins in particular so far. The coverage always came from usage in Fields. Since we don't use those new mixins in fields yet, the coverage is at 0% at the moment. I don't think we can solve this in this PR. 

## New mixins

- autocomplete
- batch
- counter
- font
- maxlength
- minlength
- pattern
- sortable
- spellcheck

### DIFF
--- a/src/Form/Mixin/Font.php
+++ b/src/Form/Mixin/Font.php
@@ -11,7 +11,10 @@ trait Font
 
 	public function font(): string|null
 	{
-		return $this->font === 'monospace' ? 'monospace' : 'sans-serif';
+		return match ($this->font) {
+			'monospace', 'mono' => 'monospace',
+			default             => 'sans-serif'
+		};
 	}
 
 	protected function setFont(string|null $font): void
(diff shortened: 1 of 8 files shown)


src/Form/Mixin/Font.php
```diff
@@ -0,0 +1,21 @@
...
+trait Font
+{
+	/**
+	 * Sets the font family (sans or monospace)
+	 */
+	protected string|null $font;
+
+	public function font(): string|null
+	{
+		return $this->font === 'monospace' ? 'monospace' : 'sans-serif';
```
distantnative: `Suggestion:` Maybe we can directly improve its functionality a tidbit.

```suggestion
		return match ($this->font) {
				'monospace', 'mono' => 'monospace',
				default             => 'sans-serif'
		};
```

distantnative: @bastianallgeier what do you think about this?

bastianallgeier: I had already merged this, but it seems like I've wrecked it with a force push after rebase. It's now fixed.

distantnative (approved):

PR #7684, opened 2025-11-24
(Fields PR 3) refact: Block field improvements

bastianallgeier:
## Merge first

- [x] https://github.com/getkirby/kirby/pull/7679

### DIFF
--- a/src/Form/Field/BlocksField.php
+++ b/src/Form/Field/BlocksField.php
@@ -68,15 +68,15 @@ class BlocksField extends FieldClass
 	) {
 		parent::__construct(
 			autofocus: $autofocus,
-			default: $default,
-			disabled: $disabled,
-			help: $help,
-			label: $label,
-			name: $name,
-			required: $required,
+			default:   $default,
+			disabled:  $disabled,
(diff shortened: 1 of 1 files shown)


src/Form/Field/BlocksField.php
```diff
@@ -259,17 +294,9 @@ protected function setDefault(mixed $default = null): void
 	}
 
 	protected function setFieldsets(
-		string|array|null $fieldsets,
-		ModelWithContent $model
+		array|null $fieldsets,
```
distantnative: `Todo:` In the constructor this can also be a string

distantnative (changes requested):

src/Form/Field/BlocksField.php
bastianallgeier: We pass this to Fieldsets::factory where we don't really accept a string. I think the type hinting here was wrong.

distantnative: Then the constructor parameter type hint needs to change 👍

distantnative (approved):

PR #7689, opened 2025-11-25
refact!: HeadlineField class

distantnative:
### ♻️ Refactored

- Refactored `headline` field as class

### ☠️ Deprecated

- `legacy-headline` field will be removed in an upcoming major version. Please move to class-based fields instead and extend the `HeadlineField` class.

### 🚨 Breaking changes

- The `headline` field is now implemented as class. When extending in an array-based field, either switch your field to a class as well or extend the deprecated `legacy-headline` field for the moment.

### For review team

  - [x] Need to remove unsupported `numbered` example 


src/Form/Field/DisplayField.php
```diff
@@ -0,0 +1,23 @@
...
+ * @license   https://getkirby.com/license
+ * @since     6.0.0
+ */
+abstract class DisplayField extends FieldClass
+{
+	public function hasValue(): bool
+	{
+		return false;
+	}
+}
```
bastianallgeier, line 23: I think we should really recreate the Field class here with less mixins that are not relevant for a DisplayField. E.g. Placeholder, Value, Validations, etc. But this could/should also happen in a separate PR.

distantnative: Yes, I think let's do that in a following PR. Then also creating an abstract `InputField` class - then it will become a lot clearer what should go where.

bastianallgeier (approved):

PR #7691, opened 2025-11-25
(Fields PR) refact: New BaseField class

bastianallgeier:
### ✨ Enhancements

- New `Kirby\Form\Field\BaseClass`

### ♻️ Refactored

- Only pass accepted arguments to the constructor in `BaseField::factory()`
- `Kirby\Form\Field` now also accepts BaseClass extensions in the factory method. 
- `Kirby\Form\FieldClass` & `Kirby\Form\Field\DisplayField` now extend `Kirby\Form\Field\BaseClass`

### For review team

### DIFF
--- a/src/Form/Field/BaseField.php
+++ b/src/Form/Field/BaseField.php
@@ -1,0 +1,156 @@
+<?php
+
+namespace Kirby\Form\Field;
+
+use Kirby\Form\Fields;
+use Kirby\Form\Mixin;
+use Kirby\Toolkit\HasI18n;
+
+/**
+ * Base class for any field type
+ *
+ * @package   Kirby Field
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @since     6.0.0
+ */
+abstract class BaseField
+{
+	use HasI18n;
+	use Mixin\Api;
+	use Mixin\DefaultValue;
+	use Mixin\Model;
+	use Mixin\Name;
+	use Mixin\Siblings;
+	use Mixin\Translatable;
+	use Mixin\Value;
+	use Mixin\When;
+
+	public function __construct(
+		string|null $name = null,
+		array|null $when = null
+	) {
+		$this->setName($name);
+		$this->setWhen($when);
+	}
+
+	/**
+	 * Returns optional dialog routes for the field
+	 */
+	public function dialogs(): array
+	{
+		return [];
+	}
+
+	/**
+	 * Returns optional drawer routes for the field
+	 */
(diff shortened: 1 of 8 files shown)


src/Form/Field/BaseField.php
```diff
@@ -0,0 +1,152 @@
...
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @since     6.0.0
+ */
+abstract class BaseField
+{
+	use HasI18n;
+	use Mixin\Api;
+	use Mixin\DefaultValue;
```
distantnative, line 24: Why does the BaseField have a default value if we want to split next between InputField and DisplayField (the latter without any value)?

src/Form/Field/BaseField.php
```diff
@@ -0,0 +1,152 @@
...
+	use Mixin\Model;
+	use Mixin\Name;
+	use Mixin\Siblings;
+	use Mixin\Translatable;
```
distantnative, line 28: DisplayField won't need this, will they? Should go into InputField otherwise.

src/Form/Field/BaseField.php
```diff
@@ -0,0 +1,152 @@
...
+	use Mixin\Value;
```
distantnative, line 29: Same as above.

src/Form/Field/BaseField.php
```diff
@@ -0,0 +1,152 @@
...
+		$field = new static(...$args);
+		$field->setSiblings($siblings);
+
+		if (array_key_exists('model', $props) === true) {
+			$field->setModel($props['model']);
+		}
+
+		if (array_key_exists('value', $props) === true) {
+			$field->fill($props['value']);
+		}
```
distantnative, line 76: Also this value handling should probably be confined to the InputField

src/Form/Field/BaseField.php
```diff
@@ -0,0 +1,152 @@
...
+
+		return $field;
+	}
+
+	public function hasValue(): bool
```
distantnative, line 81: Maybe better to make this an abstract method and let each child class choose?

src/Form/Field/BaseField.php
```diff
@@ -0,0 +1,152 @@
...
+
+	/**
+	 * @since 5.2.0
+	 * @todo Move to `Value` mixin once array-based fields are unsupported
+	 */
+	public function reset(): static
+	{
+		$this->value = $this->emptyValue();
+		return $this;
+	}
```
distantnative, line 119: As above, seems to belong to InputField.

distantnative (approved):
Let's keep it moving, we can still improve it before merging `v6/fields` into `v6/develop`.

PR #7694, opened 2025-11-25
(Fields PR) refact: New GapField class

bastianallgeier:
## Merge first

- [x] https://github.com/getkirby/kirby/pull/7691
- [x] https://github.com/getkirby/kirby/pull/7692
- [x] https://github.com/getkirby/kirby/pull/7693

### ♻️ Refactored

- Refactored `gap` field as class

### ☠️ Deprecated

- `legacy-gap` field will be removed in an upcoming major version. Please move to class-based fields instead and extend the `GapField` class.

### 🚨 Breaking changes

- The `gap` field is now implemented as class. When extending in an array-based field, either switch your field to a class as well or extend the deprecated `legacy-gap` field for the moment.

### DIFF
--- a/src/Form/Field/GapField.php
+++ b/src/Form/Field/GapField.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Kirby\Form\Field;
+
+/**
+ * Gap field
+ *
+ * @package   Kirby Field
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @since     6.0.0
+ */
+class GapField extends LineField
+{
+}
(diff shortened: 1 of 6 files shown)


src/Form/Field/GapField.php
```diff
@@ -0,0 +1,17 @@
...
+ * Gap field
+ *
+ * @package   Kirby Field
+ * @author    Bastian Allgeier <bastian@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ * @since     6.0.0
+ */
+class GapField extends LineField
```
distantnative: Not sure if this dependency makes sense semantically. If we ever change something to the line field (e.g. line color) would we want that for the gap field?

bastianallgeier: You could be right. Do we want to create another abstract class (something like DecorationField) or do we simply duplicate the code between both fields?

distantnative: I would duplicate it. But also was wondering if width could be inside BaseField.

bastianallgeier: I've moved width into the base field. Makes it all nice and tidy.

distantnative (approved):

PR #7706, opened 2025-11-27
(Fields PR) refact!: New NumberField class

bastianallgeier:
### ♻️ Refactored

- New `Kirby\Form\Field\NumberField` class

### 🚨 Breaking changes

- `NumberField::toNumber()` will now always return null instead of an empty string in case of empty values. 

### For review team

### DIFF
--- a/src/Form/Field/NumberField.php
+++ b/src/Form/Field/NumberField.php
@@ -59,15 +59,15 @@ class NumberField extends InputField
 	) {
 		parent::__construct(
 			autofocus: $autofocus,
-			default: $default,
-			disabled: $disabled,
-			help: $help,
-			label: $label,
-			name: $name,
-			required: $required,
+			default:   $default,
+			disabled:  $disabled,
+			help:      $help,
+			label:     $label,
+			name:      $name,
+			required:  $required,
 			translate: $translate,
-			when: $when,
-			width: $width,
+			when:      $when,
+			width:     $width,
 		);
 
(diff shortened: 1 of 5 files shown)


src/Form/Field/NumberField.php
```diff
@@ -0,0 +1,153 @@
...
+
+	public function default(): float|null
+	{
+		$default = $this->default;
+
+		if (is_string($default) === true) {
+			$default = $this->stringTemplate($default);
+		}
+
+		return static::toNumber($default);
```
distantnative:
```suggestion
		return static::toNumber(parent::default());
```

Would this work?

bastianallgeier: Yes, it should. But I just realized that we don't use the stringTemplate method in the DefaultValue mixin. We should change that as well.

distantnative: I think because it's not defined by the mixin but the class.

bastianallgeier: We use class methods in other mixins as well though. ::model is also not defined in the mixin.

distantnative: True

bastianallgeier: Ah, it throws a couple issues if I try to use it. Let's skip that for this PR.

bastianallgeier: Just added your suggestion though.

PR #7714, opened 2025-11-27
(Fields PR) refact!: New EmailField class

bastianallgeier:
## Merge first

- [x] https://github.com/getkirby/kirby/pull/7712

### ♻️ Refactored

- New `Kirby\Form\Field\EmailField` class
- `spellcheck` is now set to null by default for all `StringField` extensions. 

### For review team


distantnative:
I kinda understood that spellcheck your be on by default for the text field. But shouldn't it then be disabled by default at least for the email field?

bastianallgeier:
@distantnative I think the default should probably be in all cases that we just omit it by default.

bastianallgeier:
@distantnative I've changed it to fall back to null by default. This way, we don't add the spellcheck attribute at all and the browser defaults decide.

distantnative (approved):

PR #7716, opened 2025-11-28
refact!: New UrlField class

bastianallgeier:
### ♻️ Refactored

- New `Kirby\Form\Field\UrlField` class

### For review team

### DIFF
--- a/tests/Form/Field/ObjectFieldTest.php
+++ b/tests/Form/Field/ObjectFieldTest.php
@@ -102,7 +102,7 @@ class ObjectFieldTest extends TestCase
 
 		$expected = [
 			'object' =>
-				'There’s an error in the "url" field:' . "\n" .
+				'There’s an error in the "Url" field:' . "\n" .
 				'Please enter a longer value. (min. 20 characters)' . "\n" .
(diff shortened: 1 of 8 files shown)


tests/Form/Field/ObjectFieldTest.php
```diff
@@ -102,7 +102,7 @@ public function testErrors(): void
 
 		$expected = [
 			'object' =>
-				'There’s an error in the "url" field:' . "\n" .
+				'There’s an error in the "Url" field:' . "\n" .
```
distantnative, line 105: `Comment:` This looks like a side-effect we might not want? Wouldn't it be better to have the actual name in the error message how it is written in blueprints etc., instead of the human-readable label version?

I know this isn't directly caused by this PR, but something to consider in general.

bastianallgeier: We have a couple of those places and they are all pretty inconsistent. This is a user-facing error though and the name of the field might not actually be the same as the label. The user would only know the label though.

distantnative (approved):

PR #7717, opened 2025-11-28
bugfix: use same method signature as I18n::translate

nerdcel:
### Description
This pull request makes a small update to the t helper function in kirby/config/helpers.php, adjusting its parameter types and default values for improved flexibility and type safety.

### Summary of changes
The $key and $fallback parameters of the t function now accept both array and string types, and default to null if not provided.…back

### Reasoning
Since I18n::translate supports arrays in the $fallback variable, the helper should also support arrays. Otherwise the application throws an error.


distantnative (approved):

distantnative:
Thank you!

PR #7723, opened 2025-11-28
feat: Async models field previews

distantnative:
### ✨ Enhancements

- New `items/files`, `items/pages` and `item/users` Panel requests endpoints to turn IDs into item data. This is supported by the new `Kirby\Panel\Controller\Request\FileItemsRequestController`, `Kirby\Panel\Controller\Request\PageItemsRequestController` and `Kirby\Panel\Controller\Request\UserItemsRequestController` classes

### ♻️ Refactored

- Files, pages and users field previews now support IDs alongside item objects as value and will fetch the item data for these IDs automatically.

### For review team

### DIFF
--- a/config/areas/site/requests.php
+++ b/config/areas/site/requests.php
@@ -1,9 +1,19 @@
 <?php
 
+use Kirby\Panel\Controller\Request\FileItemsRequestController;
+use Kirby\Panel\Controller\Request\PageItemsRequestController;
 use Kirby\Panel\Controller\Request\PageTreeParentsRequestController;
 use Kirby\Panel\Controller\Request\PageTreeRequestController;
 
 return [
+	'items.files' => [
+		'pattern' => 'items/files',
+		'action'  => FileItemsRequestController::class
+	],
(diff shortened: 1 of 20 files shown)


config/areas/site/requests.php
```diff
@@ -1,9 +1,19 @@
...
 
 return [
+	'items.files' => [
+		'pattern' => 'items/files',
+		'action'  => FileItemsRequestController::class
+	],
+	'items.pages' => [
+		'pattern' => 'items/pages',
+		'action'  => PageItemsRequestController::class
+	],
```
distantnative, line 16: `Discussion:` Not sure if these are the best names for those request endpoints. Happy for other suggestions.

bastianallgeier: I thought about it and I can't think of a much better name to be honest. "Preview" would be too close to our page previews. Item is nice and close to what we do in the front and backend. I'm fine with it.

bastianallgeier (approved):

PR #7724, opened 2025-11-28
fix: Handle malformed image files gracefully

afbora:
Handle cases where files with valid image extensions pass mime type validation but fail getimagesize() checks. Previously this caused a TypeError when rendering pages.

Changes:
- Image::imagesize() now returns array|false
- Dimensions::forImage() returns 0x0 for invalid images
- Image::isResizable() checks for zero dimensions early

### For review team


src/Image/Image.php
```diff
@@ -144,9 +144,9 @@ public function html(array $attr = []): string
 	/**
 	 * Returns the PHP imagesize array
 	 */
-	public function imagesize(): array
+	public function imagesize(): array|false
 	{
-		return getimagesize($this->root);
+		return @getimagesize($this->root);
```
distantnative, line 149: `Question`: @afbora Do we need the `@` to suppress errors? Or would it work without as well. The docs say `On failure, false is returned.`

src/Image/Image.php
```diff
@@ -178,7 +178,19 @@ public function isSquare(): bool
...
+		if (in_array($this->extension(), static::$resizableTypes, true) === false) {
+			return false;
+		}
+
+		$dimensions = $this->dimensions();
+
+		// invalid or corrupted images return 0x0 dimensions
+		if ($dimensions->width() === 0 || $dimensions->height() === 0) {
+			return false;
+		}
```
distantnative, line 191: `Thought:` Wondering if for performance reasons, we should do this rather inside the Darkroom drivers where we already create a dimensions object (to not run that twice). But might be good to leave it in this PR as is and then tackle it in the darkroom drivers later.

src/Image/Image.php
afbora: > Errors/Exceptions[ ¶](https://www.php.net/manual/en/function.getimagesize.php#refsect1-function.getimagesize-errors)
> If accessing the filename image is impossible getimagesize() will generate an error of level [E_WARNING](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-warning). On read error, getimagesize() will generate an error of level [E_NOTICE](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-notice).

It says it will throw an error or exception if it can't read the file. If we've already done this check elsewhere, we can remove it here, as you mentioned.

However, if this line still works when there's an error in the file (which it appears to), I think using `@` would be the correct option. Or, we could put it in a try-catch block and return false.

bastianallgeier: I would prefer the try/catch block

src/Image/Image.php
bastianallgeier: @distantnative I first thought that the performance implications could be really be a problem, but I just checked and I can only find isResizable in places where we need the dimensions of the image anyway. There was a part where we also check if an image should get an image file preview, but I can't find it anymore. Do you know if it could be an issue there?

src/Image/Image.php
lukasbestle: We cannot try/catch PHP warnings and notices (that only works because Whoops converts them to exception, but will not work if Whoops is disabled). But we could use `Helpers::handleErrors()`.

However I'm not sure if it's overkill. What would we do in the handler aside from returning `false` (which the current implementation already does)?

src/Image/Image.php
lukasbestle: We call `isResizable()` externally in some places before Darkroom is even fired up. So I think we should definitely have it here. We can additionally add the check in Darkroom if it makes sense.

distantnative: @lukasbestle The concern was not additionally adding it to Darkroom but having it in `isResizable` being a performance issue as indeed it is called in several places. But those other places probably do not need the extra check introduced with this PR, but we only need to harden the usage where indeed the Dimensions class will be used, no? And that would be e.g. Darkroom.

afbora: Which part of the performance are you worried about? Does the system read the file every time an instance of `Dimensions(0, 0)` is created? If we still have concerns, the control here was extra, we can remove it.

src/Image/Image.php
afbora: That's correct, we need to catch it with `::handleErrors()`. But as @lukasbestle said, we'll return false after catching it. If we're going to do the same thing with `@`, I don't think there's any need to overcomplicate the code.

src/Image/Image.php
lukasbestle: Ah, I see.

If we move it from here to Darkroom, we need to check each calling case whether any code run before Darkroom will explode with invalid image files.

bastianallgeier (approved):

PR #7727, opened 2025-11-28
refact!: New classes for picker fields

distantnative:
### ♻️ Refactored

- New abstract `Kirby\Form\Field\ModelPickerField` class
- New `Kirby\Form\Field\FilePickerField` class
- New `Kirby\Form\Field\PagePickerField` class
- New `Kirby\Form\Field\UserPickerField` class
- New `Kirby\Form\Mixin\Upload` trait

### ☠️ Deprecated

- `pages`, `files`, `users` array.-based..

### 🚨 Breaking changes

- `legacy-pages`, `legacy-files`, `legacy-users` ...

### For review team

- [x] Add lab and/or sandbox examples (wherever helpful)
- [ ] Add changes & docs to release notes draft in Notion

PR #7731, opened 2025-11-30
refact!: Model picker fields to use picker dialogs

distantnative:
## Description

Replaces https://github.com/getkirby/kirby/pull/7522

This actually removes the `k-file-picker-dialog` and `k-user-picker-dialog` components again. Turns out, they do not need custom components. They can just use. `k-model-picker-dialog`.

### Merge first
- [x] https://github.com/getkirby/kirby/pull/7727
- [x] https://github.com/getkirby/kirby/pull/7728
- [x] https://github.com/getkirby/kirby/pull/7723
- [x] https://github.com/getkirby/kirby/pull/7750
- [x] https://github.com/getkirby/kirby/pull/7751
- [x] https://github.com/getkirby/kirby/pull/7752
- [x] https://github.com/getkirby/kirby/pull/7753
- [x] https://github.com/getkirby/kirby/pull/7754
- [x] https://github.com/getkirby/kirby/pull/7755
- [x] https://github.com/getkirby/kirby/pull/7756

### Enhancements
- Better picker dialogs that use batch selection UI 
- New `picker` option for `files`, `pages` and `users` field to customize the picker dialogs (e.g. `size`, `layout`)
- Page picker dialog shows badge for selected children

### ♻️ Refactored

- `files`, `pages` and `users` field use new picker dialogs
- `files`, `pages` and `users` fields dynamically fetch item data from new `items` field API endpoint

### 🚨 Breaking changes

- `files`, `pages` and `users` field: 
  - Backend sends only UUIDs/IDs to these fields and the fields only send UUIDs/IDs back (instead of full item objects with display data etc.)
  - Same communication happens between the field and picker dialog
- `image`, `gallery` and `video` block previews receive file UUID instead of full file item data

### For review team


src/Panel/Controller/Dialog/FilePickerDialogController.php
```diff
@@ -65,6 +66,27 @@ public function find(string $id): File|null
...
+			image:              $this->image,
+			info:               $this->info,
+			layout:             $this->layout,
+			text:               $this->text
+		))->props();
+
+		return [
+			...$props,
+			'id' => $props['uuid'] ?? $props['id'],
+		];
```
distantnative: This is quite ugly, but I didn't know a better solution within scope. The problem is that at the latest batch mode in the collection will pass around only the `id`. So we need to have the UUID if available in the `id` field.

One alternative (untested) could be that we remove UUIDs completely from the communication between backend and frontend here, but always use the regular IDs. And only when storing to content we use the UUID.

bastianallgeier (approved):

PR #7732, opened 2025-11-30
fix: Cropping with Imagick

### LINKED ISSUE #7657: Imagick: Wrong cropping without focal point
## Description  
We’ve noticed that when using the **Imagick thumb driver** and **no focus point** is selected, cropping doesn’t work correctly for certain image dimensions.  
In those cases, the image is only **resized** but not **cropped**.  

**Expected behavior**  
The image should be cropped to the specified dimensions, not just resized.

---

## To Reproduce  

1. Set up the Starterkit  
2. Set the thumb driver to **Imagick**  
3. Go to `site/templates/home.php`  
4. Change line 48 from `resize` to `crop`  
5. Clear the `media` folder  
6. Check the images on the home page (e.g. *mountains* is `1024x683` instead of `1024x1024`)  

---

## Your Setup  

**Kirby Version**  
`5.1.3`

**PHP Version**  
`8.4.14`

**System Information**  
- **Device:** Desktop  
- **OS:** Windows 11  
- **Browser:** Chrome  
- **Version:** 142.0.7444.60 (Official Build) (64-bit)

---

## Additional Context  

We’ve reviewed the `Imagick.php` class and tried to fix the issue ourselves.  
We came up with the following possible solution:  

🔗 [Imagick.php – relevant lines](https://github.com/getkirby/kirby/blob/main/src/Image/Darkroom/Imagick.php#L216-L225)

```php
//$landscape = $options['width'] >= $options['height'];

// $image->thumbnailImage(
//     $landscape ? $options['width'] : $image->getImageWidth(),
//     $landscape ? $image->getImageHeight() : $options['height'],
//     true
// );

$sourceWidth = $image->getImageWidth();
$sourceHeight = $image->getImageHeight();

$sourceRatio = Focus::ratio(
    $sourceWidth,
    $sourceHeight
);
$targetRatio = Focus::ratio(
    $options['width'],
    $options['height']
);

$resizeWidth = $options['width'];
$resizeHeight = $options['height'];
if ($sourceRatio > $targetRatio) {
    $resizeWidth = (int)round($resizeHeight * $sourceRatio);
} else {
    $resizeHeight = (int)round($resizeWidth / $sourceRatio);
}

$image->thumbnailImage(
    $resizeWidth,
    $resizeHeight,
    true
);

//$image->setGravity($gravity);

$image->cropImage(
    $options['width'],
    $options['height'],
    abs($resizeWidth - $options['width']) / 2,
    abs($resizeHeight - $options['height']) / 2
);

```

According to our findings, the gravity also needs to be calculated manually, since setGravity in Imagick does not behave the same way as the --gravity flag in ImageMagick.

The calculation of the crop start points in the example above is only implemented for the center position and was fixed for testing purposes.
The remaining gravity options would need to be implemented accordingly.

distantnative:
## Description

I tried a version where I refactored a lot of code to include the focus crop logic also in the Dimensions class. I still think this should be our goal but it got very messy and I realized that it would be too big of a change for v5.2.

Maybe I can give it another try for v6.

However, this smaller fix should align the cropping across all three darkroom drivers.

### ✨ Enhancements

- New `Kirby\Image\Gravity` enum

### 🐛 Bug fixes

- Fixed cropping with `imagick` thumb driver 

### For review team

### DIFF
--- a/src/Image/Darkroom/GdLib.php
+++ b/src/Image/Darkroom/GdLib.php
@@ -45,17 +45,9 @@ class GdLib extends Darkroom
 	 */
 	protected function resize(SimpleImage $image, array $options): SimpleImage
 	{
-		// just resize, no crop
-		if ($options['crop'] === false) {
-			return $image->resize($options['width'], $options['height']);
-		}
-
-		// crop based on focus point
-		if ($options['crop'] !== null) {
-			// get crop coords for focal point:
-			// if image needs to be cropped, crop before resizing
+		if ($crop = $options['crop'] ?? null) {
 			if ($focus = Focus::coords(
-				$options['crop'],
+				$crop,
 				$options['sourceWidth'],
 				$options['sourceHeight'],
 				$options['width'],
@@ -72,7 +64,8 @@ class GdLib extends Darkroom
 			return $image->thumbnail($options['width'], $options['height']);
 		}
 
-		return $image;
+
+		return $image->resize($options['width'], $options['height']);
 	}
 
 	/**
--- a/src/Image/Darkroom/GdLib.php
+++ b/src/Image/Darkroom/GdLib.php
@@ -51,7 +51,7 @@ class GdLib extends Darkroom
 		}
 
 		// crop based on focus point
-		if (Focus::isFocalPoint($options['crop']) === true) {
+		if ($options['crop'] !== null) {
 			// get crop coords for focal point:
 			// if image needs to be cropped, crop before resizing
 			if ($focus = Focus::coords(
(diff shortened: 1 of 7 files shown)


bastianallgeier (commented):
It's quite a big change, but looks good to me from what I can see in the code. I think it would be really useful though if we had a test setup in the lab where we can compare all the gravity positions. I really like that this is simplifying our gravity and focus point code quite a bit.

distantnative:
It's still a small change compared to what would be a clean and good solution. But that would be way too big for a minor release. So I went with something smaller - still seems to be big :D

bastianallgeier:
Big is relative here. It's big because the Image classes are such a mess and every change feels hard to review. I think my main stomache ache is coming from the changes for the imagemagick command because it caused so many issues before.

bastianallgeier (approved):

PR #7734, opened 2025-12-01
fix: `Str::short()` to consider appendix length

### LINKED ISSUE #7722: Str::Short returned string too long
## Description
I am using the Str class to shorten text. I am handing in a string variable and a max length. But the appendix is added to the shortened string at the end. So when my string has 11 chars and I shorten it to 10, I get a result with 11 chars.

**Expected behavior**  
I expected the final length of the string to be precisely the length I handed into the method. That's also what your example in the docs says, where the text is cut after 9 chars and then the appendix is added.

If this is expected behavior, the example should be changed.

## To reproduce

```php
<?php
$example = '12345678901';
echo strlen($example); // echoes '11'
$result = Str::Short($example, 10, '-');
echo strlen($result); // echoes '11'
echo $result; // echoes '1234567890-'
```

This happens here: https://github.com/getkirby/kirby/blob/2278dae6b41879bd1a8ac70ae62a6ea781fc629a/src/Toolkit/Str.php#L1062C3-L1062C58

## Your setup
**Kirby Version**  
5.1.4

distantnative:
@getkirby/backend Not sure if this is the "right" fix or if we just should adapt documentation.

Having less of the original string based on the length of the appendix also felt a little counterintuitive. At the same time, this fix does ensure that the string really ends up at the requested length.

### 🐛 Bug fixes

- `Str::short()` now considers length of appendix to always produce a string of the provided length
https://github.com/getkirby/kirby/issues/7722

### For review team

### DIFF
--- a/src/Toolkit/Str.php
+++ b/src/Toolkit/Str.php
@@ -1070,8 +1070,8 @@ class Str
 	 * ```
 	 *
 	 * @param string $string The string to be shortened
-	 * @param int $length The final number of characters the
-	 *                    string should have
+	 * @param int $length Final number of characters
+	 *                    the string (excl. appendix) should have
 	 * @param string $appendix The element, which should be added if the
 	 *                         string is too long. Ellipsis is the default.
 	 * @return string The shortened string


bastianallgeier:
To be honest, I would probably adjust the documentation. I feel that counting the chars in the actual string makes more sense semantically. I would count the appendix as added decoration on top of that.

distantnative:
@bastianallgeier Makes sense. I changed it to just adjusting the docblock.

bastianallgeier (approved):

PR #7745, opened 2025-12-05
fix: exclude CORS headers from pages cache

johannschopplich:
@lukasbestle A follow-up to https://github.com/getkirby/kirby/pull/7630.

## Description

Kirby's page cache currently persists the complete response snapshot (headers included). With request-dependent CORS settings (multiple `allowOrigin` entries, reflected request headers, credentials), the first visitor's `Access-Control-*` headers are frozen into the cache entry and leaked to every later visitor using Kirby 5.2.0 RC. That makes per-request CORS unusable whenever the page cache is active.

Rather than trying different ways to skip caching, the responder now keeps doing what it always did – inject request-aware headers when `headers()` is called. But it also records which headers are volatile. Right before a page response is written to cache, `Responder::cacheHeaders()` removes those volatile entries (or trims only the relevant `Vary` parts). When the cached response is reused, the responder re-runs the standard injection and therefore produces the correct CORS headers for the current request. There’s effectively no runtime overhead because the cache layer simply skips the extra array merge when no volatile headers were registered.

(I've added regression coverage for the multi-origin cache scenario and dedicated responder tests to focus on the new "volatile header" semantics.)

## Notes

- By trimming headers before serialization, we keep a single code path: every response (cached or not) goes through `Responder::headers()` to assemble the final header set. Cached entries simply store less data (no volatile headers), so when they're rehydrated, the exact same header injection logic runs again for the current request.
- Yes, it adds bookkeeping (tracking volatile headers, normalization helpers), but that's the minimum needed to keep caching generic. I haven't found a simpler way to manage this, maybe @lukasbestle you have another idea? 
- We no longer encode CORS-specific behavior into the cache layer; if another feature adds dynamic headers later, it only needs to call `markRequestDependentHeader()` and trimming continues to work automatically.

### ✨ Enhancements
- Track request-dependent (volatile) response headers in `Responder`, enabling Kirby to regenerate them per request without special-case cache handling.

### 🐛 Bug fixes (from 5.2.0-RC1)
- Prevent cached pages from leaking `Access-Control-*` headers between different origins; the HTML stays cached, while CORS headers are recalculated on every request.

### For review team

### DIFF
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -986,12 +986,9 @@ class Page extends ModelWithContent
 			// cache the result
 			$response = $kirby->response();
 			if ($cache !== null && $response->cache() === true) {
-				$responseData = $response->toArray();
-				$responseData['headers'] = $response->cacheHeaders($responseData['headers']);
-
 				$cache->set($cacheId, [
 					'html'        => $html,
-					'response'    => $responseData,
+					'response'    => $response->toCacheArray(),
 					'usesAuth'    => $response->usesAuth(),
 					'usesCookies' => $response->usesCookies(),
 				], $response->expires() ?? 0);
--- a/src/Cms/Page.php
+++ b/src/Cms/Page.php
@@ -986,9 +986,12 @@ class Page extends ModelWithContent
 			// cache the result
 			$response = $kirby->response();
 			if ($cache !== null && $response->cache() === true) {
+				$responseData = $response->toArray();
+				$responseData['headers'] = $response->cacheHeaders($responseData['headers']);
+
 				$cache->set($cacheId, [
 					'html'        => $html,
-					'response'    => $response->toArray(),
+					'response'    => $responseData,
 					'usesAuth'    => $response->usesAuth(),
 					'usesCookies' => $response->usesCookies(),
 				], $response->expires() ?? 0);
--- a/src/Cms/Responder.php
+++ b/src/Cms/Responder.php
@@ -407,9 +407,15 @@ class Responder implements Stringable
 
 	/**
 	 * Strips request-dependent headers for safe caching
+	 *
+	 * @since 5.2.0
 	 */
 	public function cacheHeaders(array $headers): array
 	{
+		if ($this->requestDependentHeaders === []) {
+			return $headers;
+		}
+
 		foreach ($this->requestDependentHeaders as $name => $values) {
 			if ($name === 'Vary' && is_array($values) === true) {
 				if (isset($headers['Vary']) === false) {
@@ -482,8 +488,10 @@ class Responder implements Stringable
 	/**
 	 * Marks headers (or header parts) as request-dependent, so they
 	 * can be subtracted before caching a response snapshot
+	 *
+	 * @since 5.2.0
 	 */
-	protected function markRequestDependentHeader(string $name, array|null $values = null): void
+	public function markRequestDependentHeader(string $name, array|null $values = null): void
 	{
 		if ($values === null) {
 			$this->requestDependentHeaders[$name] = null;
--- a/src/Cms/Responder.php
+++ b/src/Cms/Responder.php
@@ -69,7 +69,7 @@ class Responder implements Stringable
 	 * Tracks headers that depend on the request
 	 * and must not be persisted in the cache
 	 */
-	protected array $requestDependentHeaders = [];
+	protected array $volatileHeaders = [];
 
 	/**
 	 * Creates and sends the response
@@ -236,7 +236,7 @@ class Responder implements Stringable
 	 */
 	public function fromArray(array $response): void
 	{
-		$this->requestDependentHeaders = [];
+		$this->volatileHeaders = [];
 		$this->body($response['body'] ?? null);
 		$this->cache($response['cache'] ?? null);
 		$this->code($response['code'] ?? null);
@@ -334,7 +334,7 @@ class Responder implements Stringable
 		}
 
 		$this->headers = $headers;
-		$this->requestDependentHeaders = [];
+		$this->volatileHeaders = [];
 		return $this;
 	}
 
@@ -412,11 +412,11 @@ class Responder implements Stringable
 	 */
 	public function cacheHeaders(array $headers): array
 	{
-		if ($this->requestDependentHeaders === []) {
+		if ($this->volatileHeaders === []) {
 			return $headers;
 		}
 
-		foreach ($this->requestDependentHeaders as $name => $values) {
+		foreach ($this->volatileHeaders as $name => $values) {
 			if ($name === 'Vary' && is_array($values) === true) {
 				if (isset($headers['Vary']) === false) {
 					continue;
@@ -494,11 +494,11 @@ class Responder implements Stringable
 	public function markRequestDependentHeader(string $name, array|null $values = null): void
 	{
 		if ($values === null) {
-			$this->requestDependentHeaders[$name] = null;
+			$this->volatileHeaders[$name] = null;
 			return;
 		}
 
-		if (array_key_exists($name, $this->requestDependentHeaders) === true && $this->requestDependentHeaders[$name] === null) {
+		if (array_key_exists($name, $this->volatileHeaders) === true && $this->volatileHeaders[$name] === null) {
 			return;
 		}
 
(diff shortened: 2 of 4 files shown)


src/Cms/Page.php
```diff
@@ -986,9 +986,12 @@ public function render(
 			// cache the result
 			$response = $kirby->response();
 			if ($cache !== null && $response->cache() === true) {
+				$responseData = $response->toArray();
+				$responseData['headers'] = $response->cacheHeaders($responseData['headers']);
```
lukasbestle: `Suggestion`: This could be a `Responder::toCacheArray()` method.

`Reasoning`: The `cacheHeaders()` method is very internal and its interface (having to pass in an input array) is only useful for this one purpose. By having `toCacheArray()` next to `toArray()`, we can make this purpose more explicit and make the internal `cacheHeaders()` method `protected`. Or maybe we could also directly integrate its logic into `toCacheArray()` to keep the method list slimmer.

src/Cms/Responder.php
lukasbestle: For completeness:

```suggestion
// the `cache`, `expires`, `usesAuth`, `usesCookies` and
// `volatileHeaders` values are explicitly *not* serialized
// as they are volatile and not to be exported
```

src/Cms/Responder.php
```diff
@@ -229,6 +236,7 @@ public function code(int|null $code = null)
 	 */
 	public function fromArray(array $response): void
 	{
+		$this->volatileHeaders = [];
```
lukasbestle: `Question`: Why is this not set from the input array like the other props?

For consistency with the other properties, I think this should also be able to be set externally. Not needed for the rehydration from cache, but there could be other cases where someone wants to turn an array into a response object. We would just need to set that at the bottom (otherwise `headers()` will unset it again).

src/Cms/Responder.php
```diff
@@ -284,6 +292,15 @@ public function headers(array|null $headers = null)
...
 			$corsHeaders = Cors::headers();
 			if ($corsHeaders !== []) {
+				foreach ($corsHeaders as $name => $value) {
+					if ($name === 'Vary') {
+						$corsVaryValues = array_map('trim', explode(',', $value));
+						$this->markRequestDependentHeader('Vary', $corsVaryValues);
+					} else {
+						$this->markRequestDependentHeader($name);
+					}
+				}
```
lukasbestle: `Thought`: This causes side effects in a getter by changing the volatile header list whenever the headers are retrieved. This could lead to unexpected behavior.

`Suggestion`: We could instead dynamically inject the CORS headers into the `volatileHeaders` list in the `cacheHeaders()`/`toCacheArray()` method just like we dynamically inject them into the header list here. Then both getters do their thing without influencing each other. Disadvantage would be that the CORS headers need to be collected twice. Maybe you know a way to avoid that, but I would say the performance impact should still be pretty low unless the user defined a custom `cors` callback option that does expensive processing.

tests/Cms/CorsTest.php
```diff
@@ -527,6 +517,8 @@ public function testHeadersWithCacheBehavior(): void
 
 		$this->assertSame('https://example.com', $headers['Access-Control-Allow-Origin']);
 		$this->assertSame('no-store, private', $headers['Cache-Control']);
+
+		unset($_COOKIE['foo']);
```
lukasbestle: `Todo`: Cleaning up at the end of a test should be avoided. If any assertion in the test method fails, the cleanup code will not be executed. `tearDown()` will however always be executed. If you only want to execute it for this single test, you could use the [`After` attribute](https://docs.phpunit.de/en/12.5/attributes.html#after). But it's mostly not necessary as nothing will break and the performance impact is negligible.

tests/Cms/CorsTest.php
```diff
@@ -584,4 +576,58 @@ public function testEmptyArrayEquivalentToTrue(): void
...
+			'server' => [
+				'HTTP_ORIGIN' => 'https://app2.com'
+			]
+		]);
+
+		$html2 = $app->page('test')->render();
+		$this->assertSame($html1, $html2);
+		$this->assertSame('https://app2.com', $app->response()->headers()['Access-Control-Allow-Origin']);
+
+		Dir::remove(static::TMP);
```
lukasbestle: Same here.

src/Cms/Responder.php
```diff
@@ -431,4 +484,57 @@ public static function isPrivate(bool $usesAuth, array $usesCookies): bool
...
 		return false;
 	}
+
+	/**
+	 * Marks headers (or header parts) as request-dependent, so they
+	 * can be subtracted before caching a response snapshot
+	 *
+	 * @since 5.2.0
+	 */
+	public function markRequestDependentHeader(string $name, array|null $values = null): void
```
lukasbestle: `Todo`: Probably a leftover from the previous name:

```suggestion
	public function markVolatileHeader(string $name, array|null $values = null): void
```

Maybe the name could also be just `volatileHeader()` for consistency with `usesCookie()` etc.

tests/Cms/ResponderTest.php
```diff
@@ -228,6 +228,163 @@ public function testHeadersCacheBehavior(): void
...
+		$responder->headers();
+
+		$responder->headers(['X-Custom' => 'value']);
+
+		$headers   = $responder->toArray()['headers'];
+		$cacheable = $responder->cacheHeaders($headers);
+
+		$this->assertSame(['X-Custom' => 'value'], $cacheable);
+	}
+
```
lukasbestle, line 390: `Nitpick`: Could we also have a test for the combination with `markRequestDependentHeader()`/`volatileHeader()` (custom volatile header)?

lukasbestle (commented):
The approach with the volatile headers is great. Yes, it's additional bookkeeping, but still much cleaner and simpler than to modify the caching logic in other places. Also it will then be usable for other use cases as well without any extra logic.

I had a quick thought that we now have somewhat overlapping logic with the `usesAuth` and `usesCookies` features (both control caching of headers), but those are entirely different features with a different behavior. So I think we don't need to touch these.

lukasbestle:
Regarding timing: We want to release 5.2.0 this week (ideally Tuesday), but I don't think this needs to block the release. The CORS feature does not introduce a regression, just a bug when it is used with custom config. So we could well fix this with a 5.2.1 shortly after whenever this is ready. So no rush.

tests/Cms/ResponderTest.php
johannschopplich: For sure, I've added a unit test that calls `$responder->markVolatileHeader()`, sets headers, and verifies `toCacheArray()` removes that header while others remain.

src/Cms/Responder.php
johannschopplich: I agree. Yes, a leftover from the renaming.

tests/Cms/CorsTest.php
johannschopplich: Good catch, thanks. I've moved the cleanups to `tearDown()`.

src/Cms/Responder.php
johannschopplich: No clear intention here, I agree! Done.

src/Cms/Responder.php
johannschopplich: ✅

src/Cms/Responder.php
johannschopplich: Good catch! Now, `headers()` no longer mutates the volatile tracker at all. Instead, `toCacheArray()` builds the sanitized payload and internally calls `collectVolatileHeaders()`, which re-derives the CORS headers (if any). So the getter stays side-effect free, while the caching path still knows which headers to drop. The extra CORS lookup happens only during cache serialization, which should keep runtime overhead minimal.

src/Cms/Page.php
johannschopplich: Done, `Responder::toCacheArray()` now sits next to `toArray()` and returns the serialized snapshot with volatile headers stripped.

The old `cacheHeaders()` logic was inlined into internal helpers (`collectVolatileHeaders()` and `stripVolatileHeaders()`), so there’s no longer a public method that requires passing an array back in. In `Page::render()` we now just call `$response->toCacheArray()` before persisting.

johannschopplich:
@lukasbestle I've updated the PR – feels like a great addition to have in 5.2.0. But yeah, no rush! 🙂

lukasbestle (approved):
That's very clean now, thank you! 💛

@bastianallgeier I couldn't find a release branch for 5.2.0. If there is isn't one, could you please create one and make sure to include this PR in the final release? Thanks :)

distantnative (commented):
@lukasbestle @johannschopplich Seeing many new methods around headers and adding, removing etc them in the Responder class, I was wondering if it would be cleaner to move that logic and handling into its own class. What do you think?

bastianallgeier:
@lukasbestle ✅

@distantnative sounds good to me. I would tackle that for 5.3

PR #7750, opened 2025-12-06
fix: Dialogs/Drawers routes for block subfields

distantnative:
## Description

Adds dialog/drawer sub routes for blocks field subfields.
Was overlooked when adding earlier  FieldDialogController/FieldDrawerController to `v6/develop`.

@bastianallgeier If you have an idea how to write tests for this. I tried but failed.


bastianallgeier (approved):

bastianallgeier:
Codecov is acting up and does not show any results. I think we can ignore this safely at this point.

PR #7753, opened 2025-12-06
refact: Use `ModelItem` classes for picker dialogs

distantnative:
## Description

Goal would be to maybe eventually get rid of `Panel\Model::pickerData()`.

### Merge first

- [x] https://github.com/getkirby/kirby/pull/7752

### DIFF
--- a/src/Panel/Ui/Item/PageItem.php
+++ b/src/Panel/Ui/Item/PageItem.php
@@ -65,6 +65,7 @@ class PageItem extends ModelItem
 		return [
 			...parent::props(),
 			'dragText'    => $this->dragText(),
+			'hasChildren' => $this->model->hasChildren(),
 			'parent'      => $this->model->parentId(),
 			'status'      => $this->model->status(),
 			'template'    => $this->model->intendedTemplate()->name(),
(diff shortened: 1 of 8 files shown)


src/Panel/Ui/Item/PageItem.php
```diff
@@ -65,6 +65,7 @@ public function props(): array
 		return [
 			...parent::props(),
 			'dragText'    => $this->dragText(),
+			'hasChildren' => $this->model->hasChildren(),
```
bastianallgeier: I don't think I would add this to every page item, to be honest. We only need it in the pickers.

distantnative: Hmmm then maybe we need to create a new child class PagePickerItem?

bastianallgeier: Do you think this would be too ugly: 


```php
public function item(ModelWithContent $model): array
{
	return [
		...(new PageItem(
			page: $model,
			image: $this->image,
			info: $this->info,
			layout: $this->layout,
			text: $this->text
		))->props(),
		'hasChildren' => $model->hasChildren(),
	];
}
```

distantnative: A bit yes, but will settle for it if you think that's the good way.

bastianallgeier: I feel like an extra class just for this is a bit excessive at this point. Maybe if we find out later that we can do more things with the pickers, it makes more sense.

bastianallgeier (approved):

PR #7756, opened 2025-12-06
feat: `k-image-frame` to support UUID as src

distantnative:
### ✨ Enhancements

- `image` and `gallery` block previews now load properly sized thumbs that also respect the `cover` and `ratio` settings
- `k-image-frame` can receive a file ID/UUID via new `file` prop

### For review team

### DIFF
--- a/panel/src/components/Forms/Blocks/Types/Image.vue
+++ b/panel/src/components/Forms/Blocks/Types/Image.vue
@@ -10,23 +10,8 @@
 		@open="open"
 		@update="update"
 	>
-		<template v-if="src">
-			<k-image-frame
-				v-if="ratio"
-				:ratio="ratio"
-				:cover="crop"
-				:alt="content.alt"
-				:src="src"
-			/>
-			<img
-				v-else
-				:alt="content.alt"
-				:src="src"
-				class="k-block-type-image-auto"
-			/>
-
-			<k-block-background-dropdown :value="back" @input="onBack" />
-		</template>
+		<k-image-frame :alt="content.alt" :cover="crop" :ratio="ratio" :src="src" />
+		<k-block-background-dropdown :value="back" @input="onBack" />
 	</k-block-figure>
 </template>
 
@@ -55,14 +40,10 @@ export default {
 				return this.content.src;
 			}
 
-			if (this.content.image?.[0]?.url) {
-				return this.content.image[0].url;
-			}
-
-			return false;
+			return this.content.image?.[0]?.uuid ?? this.content.image?.[0]?.id;
 		},
 		ratio() {
-			return this.content.ratio ?? false;
+			return this.content.ratio ?? "";
 		}
 	},
 	methods: {
@@ -96,11 +77,13 @@ export default {
 	padding: var(--spacing-3);
 }
 
-.k-block-type-image-auto {
+/* Ratio: auto */
+.k-block-type-image .k-block-figure-container > img {
 	max-width: 100%;
 	max-height: 30rem;
 	margin-inline: auto;
 }
+
 .k-block-type-image .k-background-dropdown {
 	position: absolute;
 	inset-inline-end: var(--spacing-3);
--- a/panel/src/components/Forms/Blocks/Types/Image.vue
+++ b/panel/src/components/Forms/Blocks/Types/Image.vue
@@ -43,7 +43,7 @@ export default {
 			return this.content.image?.[0]?.uuid ?? this.content.image?.[0]?.id;
 		},
 		ratio() {
-			return this.content.ratio ?? "";
+			return !this.content.ratio ? "auto" : this.content.ratio;
 		}
 	},
 	methods: {
@@ -77,8 +77,7 @@ export default {
 	padding: var(--spacing-3);
 }
 
-/* Ratio: auto */
-.k-block-type-image .k-block-figure-container > img {
+.k-block-type-image .k-image-frame[data-ratio="auto"] img {
 	max-width: 100%;
 	max-height: 30rem;
 	margin-inline: auto;
--- a/panel/src/components/Forms/Blocks/Types/Image.vue
+++ b/panel/src/components/Forms/Blocks/Types/Image.vue
@@ -62,7 +62,7 @@ export default {
 			return undefined;
 		},
 		ratio() {
-			return !this.content.ratio ? "auto" : this.content.ratio;
+			return !this.content.ratio ? false : this.content.ratio;
 		}
 	},
 	methods: {
@@ -96,7 +96,7 @@ export default {
 	padding: var(--spacing-3);
 }
 
-.k-block-type-image .k-image-frame[data-ratio="auto"] img {
+.k-block-type-image .k-image-frame[data-ratio="false"] img {
 	max-width: 100%;
 	max-height: 30rem;
 	margin-inline: auto;
--- a/panel/src/components/Forms/Blocks/Types/Image.vue
+++ b/panel/src/components/Forms/Blocks/Types/Image.vue
@@ -5,12 +5,18 @@
 		:caption-marks="captionMarks"
 		:empty-text="$t('field.blocks.image.placeholder') + ' …'"
 		:disabled="disabled"
-		:is-empty="!src"
+		:is-empty="isEmpty"
 		empty-icon="image"
 		@open="open"
 		@update="update"
 	>
-		<k-image-frame :alt="content.alt" :cover="crop" :ratio="ratio" :src="src" />
+		<k-image-frame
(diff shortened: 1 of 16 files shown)


panel/src/components/Layout/Frame/ImageFrame.vue
```diff
@@ -43,14 +45,59 @@ export const props = {
...
+				this.resolvedSrc = this.src;
+				this.resolvedSrcset = this.srcset;
+				return;
+			}
+
+			// if internal file, load data for file UUID from request endpoint
+			const data = await await this.$panel.get("items/files", {
+				query: {
+					items: this.src,
+					layout: "cards",
```
distantnative: @bastianallgeier Should we add another "type" in `Panel\Model::imageSrcset()` for this case with more/different sizes?

panel/src/components/Layout/Frame/Frame.vue
```diff
@@ -12,6 +13,7 @@
 	>
 		<slot />
 	</component>
+	<slot v-else />
```
distantnative: This could be controversial.

panel/src/components/Forms/Blocks/Types/Image.vue
```diff
@@ -55,14 +40,10 @@ export default {
...
-			if (this.content.image?.[0]?.url) {
-				return this.content.image[0].url;
-			}
-
-			return false;
+			return this.content.image?.[0]?.uuid ?? this.content.image?.[0]?.id;
 		},
 		ratio() {
-			return this.content.ratio ?? false;
+			return this.content.ratio ?? "";
```
distantnative: Looking at other code parts (e.g. `k-item-image`), we should decide if the fallback/unset for ratio should be `auto` or empty.

panel/src/components/Layout/Frame/Frame.vue
distantnative: Image

Not just controversial, it also breaks stuff.

panel/src/components/Layout/Frame/ImageFrame.vue
```diff
@@ -43,14 +45,59 @@ export const props = {
...
+	data() {
+		return {
+			resolvedAlt: null,
+			resolvedSrc: null,
+			resolvedSrcset: null
+		};
+	},
+	computed: {
+		isInternal() {
+			return this.src.startsWith("file://") === true;
```
bastianallgeier: This isn't really handling the situation when we pass a simple id, right? So this would be another argument to have a better uuid fallback.

panel/src/components/Layout/Frame/ImageFrame.vue
bastianallgeier: Maybe we could not pass a layout at all and in this case we provide a full range srcset? This would mean that the fallback is no longer the list layout though.

distantnative: And we would need to support this through the Item classes down to the `Panel\Model::image*()` methods then. I also don't like too much the frontend choosing the specific secret sizes. Feels like that encourages all kind of components just asking for different sizes and fully overwhelming the media folder then. I think it would be good to settle on a couple of defined srcset ranges.

panel/src/components/Layout/Frame/ImageFrame.vue
distantnative: Oh true, good catch. I had before comparing it with http... not sure why I stopped doing that :D

panel/src/components/Layout/Frame/ImageFrame.vue
bastianallgeier: Stupid question. Do we still need the different srcsets for different layouts at all anymore when the image frame sets the width? Isn't just one enough?

distantnative: I think we are thinking about this in different ways: my thought is we provide a few standard sizes as srcset and by telling the browser the current exact size in sizes, it can choose the best.

I understand you like you would like to pass the current size to the backend (or the nearest x size) and then the backend provides only the best fitting image.

I think the latter has the danger of creating many different thumb sized. And when the browser window gets resized, the backend might have to create a new thumb and update the src with it. While in the first option, the standard srcset is there already and providing the exact size just helps the browser to know which one to pick.

bastianallgeier: No, I'm with you. But so far we have different srcsets for different layouts, right? But if we define one good srcset, the layout option would no longer be needed.

panel/src/components/Layout/Frame/ImageFrame.vue
bastianallgeier: But when we check for http, we can't have absolute relative URLs with a leading / anymore. I think we should really pass the id in the UUID format as well.

panel/src/components/Layout/Frame/ImageFrame.vue
distantnative: Ok, could be. But probably a bit wasteful to send all the larger srcset sizes also to a list item, no?

panel/src/components/Layout/Frame/ImageFrame.vue
distantnative: I know that you want that UUID/ID refactoring a lot :D , but I fear this will need to be its own massive project. I really would love not to make this small PR dependent on that.

panel/src/components/Layout/Frame/ImageFrame.vue
bastianallgeier: Ah, right. The response would get bigger. Ok, so let's call the default srcset simply `default` and return that whenever there's no layout specified and otherwise we stick to the srcsets that we already have.

panel/src/components/Layout/Frame/ImageFrame.vue
bastianallgeier: Any other possible prefix? fileId:// or something else?

bastianallgeier (approved):

src/Panel/Model.php
```diff
@@ -233,59 +234,60 @@ protected function imageSrcset(
 		// depending on layout type, set different sizes
 		// to have multiple options for the srcset attribute
 		$sizes = match ($layout) {
-			'cards'    => [352, 864, 1408],
+			'auto'     => [100, 300, 800, 1500, 2500],
+			'cards'    => [300, 800, 1500],
 			'cardlets' => [96, 192],
-			default    => [38, 76]
+			default    => [36, 72],
```
distantnative: @bastianallgeier I checked, I think nowadays the actual list size is 36px not 38px. Do you see the same?

src/Panel/Model.php
```diff
@@ -233,59 +234,60 @@ protected function imageSrcset(
 		// depending on layout type, set different sizes
 		// to have multiple options for the srcset attribute
 		$sizes = match ($layout) {
-			'cards'    => [352, 864, 1408],
+			'auto'     => [100, 300, 800, 1500, 2500],
+			'cards'    => [300, 800, 1500],
```
distantnative: @bastianallgeier Do you think this is a sensible list for `auto`? And thought we might use the same steps for `cards` to not create very close but different thumbs

src/Panel/Model.php
bastianallgeier: Yes, I see the same. 38px seems outdated.

src/Panel/Model.php
bastianallgeier: I think it would be ideal if we could always use the same sizes for all layouts. Maybe something like this: 

```php
'auto' => [50, 100, 200, 400, 800, 1200, 1600, 2400],
'cards' => [400, 800, 1600],
'cardlets' => [100, 200],
default => [50, 100]
```

distantnative: I get the idea, but would it make sense to send srcsets to cardlets and lists that always have to be scaled - we do know the fixed sizes there, so shouldn't we deliver directly the best thumbs for these?

distantnative: I already adapted the first two lines at least.

bastianallgeier: I think it's quite some balance between the image processing on the server, the file size of the cached image files and the quality. I would say that downscaling a tiny bit in the browser is a good compromise. But you might be right that we could also set the lower sizes to the defined one we already know.

distantnative: Let's do actually that: let's use the same size points for all, but for the lower ones we use the crude exact sizes of list and cardlets - looks a bit weird but probably doesn't matter for e.g. cards if the smallest is 36 or 50, but then we deliver directly the perfect ones for list and cardlets

bastianallgeier (approved):

PR #7757, opened 2025-12-06
Auto sizes from observer for image frame

distantnative:
### Merge first

- [x] https://github.com/getkirby/kirby/pull/7756

### ✨ Enhancements

- Image frames without `sizes` will now provide their actual width as `sizes` attribute to support the browser in picking the best match from `srcset`

### For review team


distantnative:
@bastianallgeier I think quite a downside that this currently has is that `sizes` will only be measured/calculated once the image has been loaded... which means the browser will load first fully `src` and then the matching secret element. So images twice at times.

distantnative:
Closing this for now until we have a better solution.

PR #7758, opened 2025-12-06
fix: Proper fallback for `HasI18n` with data

distantnative:
## Description

- Fix: Pass key also as fallback in `HasI18n:: i18n()`
- Make `HasI18nTest ` independent of the unit tests preloading i18n strings

### DIFF
--- a/tests/Toolkit/HasI18nTest.php
+++ b/tests/Toolkit/HasI18nTest.php
@@ -6,6 +6,11 @@ use Kirby\TestCase;
 
 class HasI18nTest extends TestCase
 {
+	public function setUp(): void
+	{
+		$this->setUpI18nLoader();
+	}
+
 	protected function object(): object
 	{
 		return new class () {
--- a/tests/Toolkit/HasI18nTest.php
+++ b/tests/Toolkit/HasI18nTest.php
@@ -2,18 +2,32 @@
 
 namespace Kirby\Toolkit;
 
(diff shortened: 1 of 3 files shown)


tests/Toolkit/HasI18nTest.php
lukasbestle: `Suggestion`: I think it would be good to also have assertions for the fallback, i.e. requesting a translation key that doesn't exist, both with and without data. Expected behavior would then be that the key is returned in both cases.

tests/Toolkit/HasI18nTest.php
```diff
@@ -6,6 +6,11 @@
 
 class HasI18nTest extends TestCase
 {
+	public function setUp(): void
+	{
+		$this->setUpI18nLoader();
```
lukasbestle: `Thought`: Wondering whether we should actually test with Kirby core translations. Those can change, breaking tests even without any code changes. Maybe we should rather populate `I18n::$translations[$locale] = [<some test strings>]` and unset the array again in `tearDown()`. This would isolate the tests as much as possible and also make it clearer what data we test on.

tests/Toolkit/HasI18nTest.php
distantnative: Only key is already lines 31 etc. but adding one example with non-existing key and data as well.

bastianallgeier (approved):

PR #7760, opened 2025-12-07
refact: `$value` prop handling in field classes 

distantnative:
## Description

- Do not define the property in `Mixin\Value` but the field classes so we can type hint it more strictl
- Move `Mixin\Value` to `InputField` as just the base field class should not yet have a value
- For `::hasValue()` check if `$value` property exists

### DIFF
--- a/src/Form/Field/BaseField.php
+++ b/src/Form/Field/BaseField.php
@@ -26,7 +26,6 @@ abstract class BaseField
 	use Mixin\Name;
 	use Mixin\Siblings;
 	use Mixin\Translatable;
-	use Mixin\Value;
 	use Mixin\When;
 	use Mixin\Width;
 
@@ -74,16 +73,15 @@ abstract class BaseField
 			$field->setModel($props['model']);
 		}
 
-		if (array_key_exists('value', $props) === true) {
-			$field->fill($props['value']);
-		}
-
 		return $field;
 	}
 
+	/**
+	 * Checks if the field has a value
+	 */
 	public function hasValue(): bool
 	{
-		return false;
+		return property_exists($this, 'value') === true;
 	}
 
 	public function id(): string
--- a/src/Form/Field/BaseField.php
+++ b/src/Form/Field/BaseField.php
@@ -113,16 +113,6 @@ abstract class BaseField
 	}
 
 	/**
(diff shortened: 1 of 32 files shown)


src/Form/Field/BaseField.php
```diff
@@ -74,16 +73,22 @@ public static function factory(
 			$field->setModel($props['model']);
 		}
 
-		if (array_key_exists('value', $props) === true) {
+		if (
+			array_key_exists('value', $props) === true &&
+			method_exists(static::class, 'fill') === true
+		) {
```
distantnative, line 79: Was considering moving this to a more specific `InputField::factory()` instead. However, then we need to implement this for all other fields that use `Mixin\Value` but not extend `InputField`, e.g. `HiddenField`.

src/Form/Fields.php
```diff
@@ -277,7 +278,7 @@ public function reset(): static
 		foreach ($this->data as $field) {
 			if ($field instanceof Field) {
 				$field->fillWithEmptyValue();
-			} else {
+			} elseif ($field instanceof InputField) {
 				$field->reset();
```
distantnative: Problem: This now leaves the HiddenField without being reset. Maybe check for `method_exists` instead?

bastianallgeier: Hm, it makes me wonder if we should simply have fill and reset for all fields and the field can decide if it uses it or not.

bastianallgeier: Or we always use the hasValue check

bastianallgeier:
General thoughts: This is a great step. The type hinting feels so much better and cleaner now. We could really think about introducing more foundational field types (ArrayField, BooleanField)

src/Form/Fields.php
distantnative: Maybe hasValue check makes sense. That's then required for all fields to have, but only those that return true need to implement/use the Value mixin methods.

bastianallgeier (approved):
I think the changes look really good now. I wonder where the coverage issues are coming from. From what I can see, it's mostly stuff where it suddenly decided that already existing code isn't correctly covered.

distantnative:
Even the responder class that isn't touched in this PR

PR #7770, opened 2025-12-11
feat: Responsive mode for the table

bastianallgeier:
### ✨ Enhancements

- The `<k-table>` component has a new `responsive` prop, which is set to `true` by default. Switching the responsive mode off will no longer hide columns that are not marked with `data-mobile="true"`. This gives more control over custom responsive behaviors for tables and also helps to improve the usability in narrow widths. 

### For review team

### DIFF
--- a/panel/lab/components/tables/4_component/index.vue
+++ b/panel/lab/components/tables/4_component/index.vue
@@ -3,6 +3,14 @@
 		<k-lab-example label="Default">
 			<k-table :columns="columns" :options="options" :rows="rows"></k-table>
 		</k-lab-example>
+		<k-lab-example label="Responsive: false">
+			<k-table
+				:columns="columns"
+				:options="options"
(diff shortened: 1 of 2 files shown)


bastianallgeier:
`Discussion:` When creating a custom table without the component, but raw HTML, devs would now need to set the `data-responsive="true"` attribute to make sure that the old responsive behavior still works. This is technically a breaking change, but I would almost consider this a bugfix. So far, you really had to think about adding the data-mobile attributes to a last one column. Otherwise the table would be gone in narrow containers. If we consider this a real breaking change though, we should move it to 6.0.0

distantnative:
@bastianallgeier I agree that it's the right thing to do. But since we don't have any urgency on this (do we? nothing that needs to go into v5.x depends on this, right?), we might honor the breaking change and move this to v6.

PR #7774, opened 2025-12-12
feat: New k-error-trace component

bastianallgeier:
### 🎉 Features

- New `<k-error-trace>` component to display PHP stack traces. We will use it for PHP error stacks in error dialogs, but it's universal enough to also be used in other places or for JS traces. 
Screenshot 2025-12-12 at 11 29 24

### For review team

### DIFF
--- a/panel/src/components/Errors/ErrorTrace.vue
+++ b/panel/src/components/Errors/ErrorTrace.vue
@@ -1,0 +1,72 @@
+<template>
+	<div class="k-error-trace">
+		<ol class="k-stack">
+			<li v-for="(item, index) in trace" :key="'trace-' + index">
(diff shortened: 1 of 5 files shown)


panel/src/components/Errors/ErrorTrace.vue
```diff
@@ -0,0 +1,72 @@
...
+						<span>()</span>
+					</p>
+				</a>
+			</li>
+		</ol>
+	</div>
+</template>
+
+<script>
+export default {
```
distantnative, line 40:
```suggestion
/**
 * @since 6.0.0
 */
export default {
```

distantnative (approved):

PR #7777, opened 2025-12-12
fix: Handle race conditions in `Dir::remove()`

### LINKED ISSUE #6266: Error deleting cache on high traffic site
## Description
On our customers high traffic website we get sometime this error messages in the Panel:
rmdir(cachefolderpath) Directory is not empty

**Expected behavior**  
Deleting cache without a Error or prevent deleting of not empty folders

**Screenshots**  
grafik
or
grafik

## To reproduce
We where not able to reproduce the error in a local or staging environment by ourselfs.
i think the problem is caused by the high traffic. When we remove the cache for 20k+ pages and meanwhile new pages have been cached the folder is not empty and can not be deleted in /vendor/getkirby/cms/src/Filesystem/Dir.php line 553 and causes the error shown above.
grafik

## Your setup
**Kirby Version**  
4.1.0

afbora:
## Description

Wrap rmdir() call in Helpers::handleErrors() to gracefully handle race conditions where files are written to a directory between the scandir() and rmdir() calls. This can occur when concurrent processes (e.g., cron jobs and web requests) operate on the same cache directory simultaneously.

`FileCache::flush()` now catches the exception and returns false instead of failing fatally.

I've been trying to write a unit test about race conditions but I haven't succeeded. I'd appreciate any ideas you have.

### 🐛 Bug fixes

- Flush cache generates a "Directory not empty" error
https://forum.getkirby.com/t/flush-cache-generates-a-directory-not-empty-error/35063/3
#6266 

### For review team

### DIFF
--- a/src/Cache/FileCache.php
+++ b/src/Cache/FileCache.php
@@ -6,6 +6,7 @@ use Kirby\Exception\Exception;
 use Kirby\Filesystem\Dir;
 use Kirby\Filesystem\F;
 use Kirby\Toolkit\Str;
+use Throwable;
 
 /**
  * File System Cache Driver
@@ -214,13 +215,10 @@ class FileCache extends Cache
 	 */
 	public function flush(): bool
 	{
-		if (
-			Dir::remove($this->root) === true &&
-			Dir::make($this->root) === true
-		) {
-			return true;
+		try {
+			return Dir::remove($this->root) && Dir::make($this->root);
+		} catch (Throwable) {
+			return false;
 		}
-
-		return false; // @codeCoverageIgnore
 	}
 }
--- a/src/Filesystem/Dir.php
+++ b/src/Filesystem/Dir.php
@@ -588,7 +588,11 @@ class Dir
 			}
 		}
 
-		return rmdir($dir);
+		return Helpers::handleErrors(
+			fn (): bool => rmdir($dir),
+			fn (int $errno, string $errstr) => true,
+			fn (): never => throw new Exception('The directory could not be deleted'),
+		);
 	}
 
 	/**


src/Cache/FileCache.php
```diff
@@ -214,13 +215,10 @@ protected function removeEmptyDirectories(string $dir): void
...
 	{
-		if (
-			Dir::remove($this->root) === true &&
-			Dir::make($this->root) === true
-		) {
-			return true;
+		try {
+			return Dir::remove($this->root) && Dir::make($this->root);
+		} catch (Throwable) {
+			return false;
```
afbora, line 221: Do you think we should return true or false when an error occurs?

src/Filesystem/Dir.php
```diff
@@ -588,7 +588,11 @@ public static function remove(string $dir): bool
 			}
 		}
 
-		return rmdir($dir);
+		return Helpers::handleErrors(
+			fn (): bool => rmdir($dir),
+			fn (int $errno, string $errstr) => true,
+			fn (): never => throw new Exception('The directory could not be deleted'),
+		);
```
distantnative, line 595: I read a few posts on how to deal with such race conditions. Some recommendations were:
- Rename the directory first as this is atomic. Then delete the renamed folder. This way any new writes to the same location of the folder that come in after `Dir::remove()` started can still be created, but will not create problems deleting the files captured at the start of the remove call.
- Instead of just giving up and throwing an exception, there are a few posts that recommend to keep retrying the `rmdir` call a few times if an error occurs. That increases the chances of one being completed. However, in a high traffic caching environment, there is still the chance that none will actually succeed.

What do you think @lukasbestle  @bastianallgeier?

bastianallgeier: I like the idea to think outside the box for more ways to prevent this. Retrying seems worth a shot. Moving the folder sounds like a good idea at first too, but we could end up with a lot of unremoved clones if we don't get it right.

afbora: I really like the idea. The `rename` method will return either true or false. If it returns true, I see no reason why `Dir::remove` shouldn't work. If `rename` returns false, the `Dir::remove` process won't even start. Thus, no clone will be created.

If we want to be absolutely sure, we can add an extra clone check and delete handler inside the `FileCache::flush()` method. This way, if a cloned folder remains somehow, it will always be cleaned up before flushing.

lukasbestle: Making the deletion atomic via a prior renaming operation would be perfect indeed. But it needs to be reliable (no left-over random directories on failure) and it also needs to be transparent to calling code (there should be no need to do anything special in `FileCache::flush()` for example).

With a naive approach of just renaming and then deleting contents as normal, we could run into issues e.g. when any file is not writable by the PHP user. The deletion will then fail in the middle of the operation and the renamed directory will still be there.

What we would need is logic along these lines:

```php
rename($currentName, $newName = $currentName . $randomSuffix);

try {
  // current `scandir()` loop
} catch (Throwable $e) {
  Helpers::handleErrors(
    fn (): bool => rename($newName, $currentName),
    function () {
      // someone has already created the original directory again;
      // instead move all contents of the renamed directory to the new directory
    }
  );
}
```

But that's very complex and would require us to convert all PHP warnings that can occur in the process to proper exceptions. Which would be nice anyway, but a project on its own.

---

Another thought: `FileCache::flush()` doesn't really need to delete the directories (neither the top-level one nor any subfolders). These can still be deleted on the next flush operation without any impact to the cache behavior. So we could add a `bool $lenient = false` mode to `Dir::remove()` that uses `@rmdir()` instead of `rmdir()`. This way, any issues with file deletions will still trigger a warning, but directory removals will no longer make the process fail.

distantnative: Just another thought on the renaming: could we rename it to somewhere where it doesn't harm if by whatever reason it remains as leftover? And then e.g. via our Panel garbage collection just regularly delete any potential leftovers in that location? (As anything in that location would have been meant to be deleted anyways)

bastianallgeier: If the tmp folder would be cleaned up correctly by the system, it would be ideal to move it there if possible. But we have a similar situation with uploads already. It's not being cleaned up as regular as we wish and as long as we don't have some built in queue system, I'm not sure what would trigger such clean ups on our end.

distantnative: Panel garbage collection which runs in 10% of any Panel requests. The same we use for uploads.

src/Cache/FileCache.php
bastianallgeier: Definitely false.

src/Filesystem/Dir.php
bastianallgeier: Ah, that's a good idea indeed. Maybe we should separate this. I think @afbora's PR is already a great first step to improve error handling. Adding a second layer to solve the flush problem would still make sense to me, but could be another PR. What do you think?

distantnative: Can be two separate PRs, yes. But wouldn't use the rename or other solutions just for flushing but `Dir::remove()` in general as well.

lukasbestle: I agree on the two-step approach.

However I think this PR can be considered a breaking change. Before, any failure during the flushing operation will trigger a warning and therefore an exception if Whoops is enabled. Now, there will be only a silent `false` return value that so far no one was checking. So if the caller needs to rely on the cache being flushed for following actions, such use cases will now silently break if anything goes wrong during the flushing operation.

IMO to fix this in a way that still keeps reliability, we need to differentiate between errors during file deletion and during directory deletion (the latter can be ignored).

bastianallgeier: I think we need to view this from the opposite angle to be honest. Right now, you can no longer save pages or anything else on a high-traffic website when the folder cannot be properly removed, because it will always cause an error. This is pretty bad if you ask me. I still agree that we cannot simply swallow the issue. We should maybe think about writing it to the php error log instead though.

distantnative: I think both of you are right: We need to fix this, but maybe the fix needs to be more specific about which errors get swallows (and maybe written to the php error log) and which ones it should not swallow.

Cause this isn't just about flushing the chance, right. Any change will affect also all other use cases of `Dir::remove()`.

bastianallgeier: That's true. Let's maybe break it down. 

I think it's a good idea to improve the error handling in Dir::remove. This will be helpful for all situations in which it is being used. 

We are not swallowing anything here, we only make it easier to catch because it will result in an exception and not a warning. 

Swallowing it in the flush method could indeed be the wrong approach at this point. Maybe we shouldn't be doing this in this PR at all and only focus on the error handling in the Dir method and go from here in the next step?

afbora:
So what's the last status of the PR? 🙂

src/Filesystem/Dir.php
lukasbestle: Coming back to the lenient mode I proposed above: I think this discussion has two components that we mixed and that are hard to solve together:

- We want cache flushing to continue even if a directory gets left over because a new cache file was created in the meantime.
  - For this, catching warnings and converting them to exceptions will not work either, because this would cancel the loop and exit early. Following directories and their files will not be deleted then.
  - We need to skip over directory deletion errors but continue with the loop.
- We want to improve error handling in general.
  - This has other goals: We want to make errors more visible and use standardized exceptions rather than PHP warnings.
  - But this won't help us with the cache flushing.

So I think if we want to solve the cache flushing issue, we rather need a solution that uses `@rmdir()` instead of `rmdir()`. Then all files that were found will be deleted, but some directories may be left over (which is not a problem in the specific case of the file cache as directories themselves don't carry any meaning here, just files).

The `@rmdir()` solution will however not work for all uses of `Dir::remove()`, so it either needs to be a separate method, a custom implementation right in `FileCache` or a flag for `Dir::remove()`.

distantnative: I searched around. The "rename then remove" strategy is often quoted as the way to go as it is atomic. Then again, I couldn't find it in large PHP projects in action. I still think this would be a good approach to enhance `Dir::remove()` for contexts where the to be removed directory is constantly changing. Rename it to move it out of the way (the changes can start to create a new directory/files in the old place), then delete it without filesystem writes still trying to happen on it.

lukasbestle: If we find a way to combine it with background garbage collection in case the removing of the renamed directory fails, I think it's the best solution.

distantnative: Proposal:
- `Dir::remove()` first renames/moves the directory to a tmp folder, e.g. `site/cache/.tmp`
- If rename fails, just try to remove in place
- if that fails, do some bounded retries (e.g. try 5 times with a delay of some ms), after that throw error/return false, whatever we feel works best then
- If rename succeeded, try to remove recursively inside tmp folder
- if that fails, just return true and let the rest of the clean-up be handled by garbage collection
- For garbage collection of tmp folder use `Panel::garbage()` similarly to how we use it already for `Upload::cleanTmpDir()`

bvdputte:
Hi all, any update on this? Thx!

bastianallgeier:
I've created a new attempt to move forward with this https://github.com/getkirby/kirby/pull/8125

PR #7789, opened 2025-12-18
refact: Remove `default` from field props

distantnative:


src/Cms/Fieldset.php
```diff
@@ -77,12 +77,18 @@ public function __construct(array $params = [])
...
+		foreach ($fields->keys() as $field) {
+			$props[$field]['default'] = $fields->field($field)->default();
+		}
 
 		// collect all fields
-		$this->fields = [...$this->fields, ...$fields];
+		$this->fields = [...$this->fields, ...$props];
 
-		return $fields;
+		return $props;
```
distantnative: `Discussion:` To keep this DRY-er, we would consider moving this logic to `Form\Fields::toProps()`, maybe via a parameter flag, `$fields->toProps(defaults: true)`.

bastianallgeier (approved):

PR #7791, opened 2025-12-19
feat: Remove `legacy` query runner as default

distantnative:
### ☠️ Deprecated

- `legacy` query runner has been deprecated and removed as default. Use `Kirby\Query\Runners\DefaultRunner` instead.
- `Kirby\Query\Argument`, `Kirby\Query\Arguments`, `Kirby\Query\Expression`, `Kirby\Query\Segment`, `Kirby\Query\Segments` have been deprecated and will be removed in Kirby 7.

### For review team

### DIFF
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -12,6 +12,7 @@ use Kirby\Cms\User;
 use Kirby\Cms\Users;
 use Kirby\Exception\InvalidArgumentException;
 use Kirby\Image\QrCode;
+use Kirby\Query\Runners\DefaultRunner;
 use Kirby\Query\Runners\Runner;
 use Kirby\Toolkit\I18n;
 
@@ -30,8 +31,7 @@ class Query
 {
 	public static array $cache = [];
 	public static array $entries = [];
-
-	public Runner|string $runner;
+	public static Runner|string|null $runner = null;
 
 	/**
 	 * Creates a new Query object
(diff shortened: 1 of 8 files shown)


src/Query/Query.php
```diff
@@ -43,15 +43,16 @@ public function __construct(
 			$this->query = trim($query);
 		}
 
-		$this->runner = App::instance()->option('query.runner', 'legacy');
-
-		if ($this->runner !== 'legacy') {
+		static::$runner ??= App::instance()->option('query.runner', DefaultRunner::class);
```
distantnative, line 46: `Discuss:` Moving this to a static property allows us to prevent calling `$kirby->option()` for every query run in the same request. Which seems like quite the performance downside - one we're trying to get rid of for UUIDs. However, with this change, queries would no pick up any changed `query.runner` option via e.g. `$app->clone()`. We should decide if in general we want to always support such runtime config changes or not.

distantnative: Alternatively, we could add specialized methods to `Kiry\Cms\App`, e.g. `::queryRunner()`, `::hasUuids()`, and cache those inside the App class, where we then could reset the caches in `::clone()`, `::destroy()` etc.

bastianallgeier: We have a couple other static public props that we handle in the App class. It always feels a bit icky, but I think instead of introducing yet another method, I would simply reset the new static method directly in clone and destroy. It helps us at least to keep track of it all.

distantnative: Added it this way 👍

bastianallgeier (approved):

PR #7793, opened 2025-12-20
Fix focus loss in entries field

afbora:
## Description

Every time you type a character:
1. The watcher sees a new array reference coming in
2. It recreates all entries with brand new IDs
3. Vue sees different IDs and remounts everything
4. Focus is lost

The problem was this check wasn't working:
```js
if (entries === this.values) {
  return;
}
```

Because `this.values` is a computed property that returns a new array each time, so `===` always returns false.

Added a proper value comparison using `JSON.stringify()` to check if the actual content has changed before recreating entries. Not the most elegant solution but it works and is simple enough.

### 🐛 Bug fixes

- Entries field: can only type one letter before field looses focus #7779

### For review team

### DIFF
--- a/panel/src/components/Forms/Field/EntriesField.vue
+++ b/panel/src/components/Forms/Field/EntriesField.vue
@@ -237,8 +237,9 @@ export default {
 			handler(entries) {
 				entries ??= [];
 
-				// no need to add ids again if the values are the same
-				if (entries === this.values) {
+				// Prevent unnecessary re-creation of entries that causes focus loss
+				// when typing in input fields by comparing actual values
+				if (JSON.stringify(entries) === JSON.stringify(this.values)) {
 					return;
 				}
 


afbora:
Actually, the logic in the entries field hasn't changed in v6, but I think this problem might be due to some changes that came with Vue 3.

distantnative:
@afbora @bastianallgeier I am wondering if it would make sense to add the IDs rather in the PHP class instead?

bastianallgeier:
I like the idea to add unique IDS on the backend. We do the same in the Structure field and with the new classes it should be easier to handle them correctly.

afbora:
In the structure field, we're only removing the `_id` index on the PHP side. In the entries `field, should we generate the ID on the PHP side and store it in the content file? If not, how does generating a unique ID on the PHP side change (compared to doing it on the front end)?

If you provide a little more detail, I can make the necessary improvements 🙂

bastianallgeier:
My bad, you are completely right. We actually attach them in the frontend in the structure field so far. We plan to do it on the backend though. 🙈

Let's keep the JSON comparison for now until we have a clean solution for the UUIDs in the structure field that we can use in the same way in the entries field. Or what do you think @distantnative?

distantnative:
Also a wrong thought on my end: I don't think we can store them in the content file yet, but I thought they would be stable coming from the backend and then we would have not to worry so much in the Vue watch handler about them.

But you're right @afbora that even if we add them on the backend, as long as we do not store them in content, they would constantly be different.

I am fine with this PR.

distantnative (approved):

PR #7796, opened 2025-12-20
fix: Writer toolbar position without Panel menu

distantnative:
### 🐛 Bug fixes

- Writer field: fixed inline toolbar position on views without Panel menu

### For review team


bastianallgeier:
This is something that we should really fix in 5.x

distantnative:
@bastianallgeier We could but we also do not have any view without the Panel menu that could host a writer field. That only becomes an issue with the new preview view form.

PR #7798, opened 2025-12-20
refact: Model tabs reload view instead link

distantnative:
## Description

I think it would be advantageous for the model tabs to not use the `link` of the tab, but rather reload the view with the new tab name. I don't think for model tabs there is any other use case for the link but to point to the same URL with a different tab. By ignoring the link, this will allow us to use model tabs in other views, e.g. preview view, where the URL would not match but reloading with new tab name still works.

### Breaking changes

- `k-model-tabs` ignores `tab.link` but reloads the view with the tab as query

### For review team

### DIFF
--- a/panel/src/components/Navigation/ModelTabs.vue
+++ b/panel/src/components/Navigation/ModelTabs.vue
@@ -8,7 +8,10 @@
  */
 export default {
 	props: {
-		diff: Object,
+		diff: {
+			type: Object,
+			default: () => ({})
+		},
 		tab: String,
 		tabs: {
 			type: Array,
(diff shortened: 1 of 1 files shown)


bastianallgeier:
But we would lose the direct links for tabs that way, right? We shouldn't do that. We had tabs without links in the past and I remember that it was an actual feature wish back then to add links

distantnative:
Was that for model tabs or just `k-tabs`?

bastianallgeier:
It was specifically about having direct links to the different tabs for models. I.e. when you want to send an editor directly to the settings tab of a page or something.

PR #7803, opened 2025-12-21
test: Use ParaTest to run PHP unit tests

distantnative:
## Description

- Make sure to install paratest 7.16.1 before testing locally

### 🧹 Housekeeping

- Using ParaTest to run PHPUnit tests in parallel

### For review team

### DIFF
--- a/tests/Toolkit/VTest.php
+++ b/tests/Toolkit/VTest.php
@@ -33,8 +33,8 @@ class VTest extends TestCase
 {
 	public function tearDown(): void
 	{
+		parent::tearDown();
 		V::$validators = [];
-		App::destroy();
 	}
 
 	public function testCallCustomValidator(): void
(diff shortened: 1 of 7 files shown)


tests/Toolkit/VTest.php
```diff
@@ -31,10 +32,22 @@ public function count(): int
...
+		I18n::$locale = 'foo';
+		I18n::$translations['foo'] = [
+			'error.validation.same' => 'Error message for same: {other}'
+		];
+	}
+
 	public function tearDown(): void
 	{
+		unset(I18n::$translations['foo']);
+		I18n::$locale = $this->localeBackup;
```
distantnative: If someone has a good idea how to maybe wrap this in a helper method or so... it's not really that great having this code non-DRY here, in `HasI18nTest` and potentially more tests in the future.

distantnative: That also seems to be not a bulletproof solution. Occasionally I see ParaTest still fail for HasI18nTest with missing i18n strings...

bastianallgeier (approved):
The performance is incredible. It's all fine for me locally except the missing memcache extension. But that's the same as before. So awesome!

PR #7807, opened 2025-12-21
test: Reset $_SERVER manipulation

distantnative:


bastianallgeier (approved):

tests/Cms/App/AppLanguagesTest.php
```diff
@@ -102,6 +103,9 @@ public function testDetectedLanguage($accept, $expected): void
 		]);
 
 		$this->assertSame($expected, $app->detectedLanguage()->code());
+
+		// reset the $_SERVER value
+		$_SERVER['HTTP_ACCEPT_LANGUAGE'] = $acceptLanguage;
```
lukasbestle, line 108: These won't run if the assertion above fails. Is there a reason why these can't be in `tearDown()`?

distantnative: I put them there as these were just relevant for 2 of the tests. But you might be right that it'd be safer in setUp/tearDown.

PR #7808, opened 2025-12-21
perf: Cache `Uuids::enabled()` result

### LINKED ISSUE #7094: Add a static cache for the UUID enabled check
## Description
The enabled check for UUIDs is call on every creation of every UUID object. On setups with many content pages that thousands of roundtrips to the app instance can be avoided.

Adding a simple static cache here will vastly improve that.

**current**
```php
public static function enabled(): bool
	{
		return App::instance()->option('content.uuid') !== false;
	}
```

**proposed**
```php
private static ?bool $is_enabled = null;
public static function enabled(): bool
	{
                if (static::$is_enabled !== null) {
                      return static::$is_enabled;
                }
		return static::$is_enabled = App::instance()->option('content.uuid') !== false;
	}
```

https://github.com/getkirby/kirby/blob/a0e64e87743af69a3798bfc78e4e578b674a1173/src/Uuid/Uuids.php#L83

distantnative:
### ⚡ Performance

- Improved performance of UUID enabled check
https://github.com/getkirby/kirby/issues/7094

### For review team

### DIFF
--- a/src/Cms/App.php
+++ b/src/Cms/App.php
@@ -36,6 +36,7 @@ use Kirby\Toolkit\LazyValue;
 use Kirby\Toolkit\Locale;
 use Kirby\Toolkit\Str;
 use Kirby\Uuid\Uuid;
+use Kirby\Uuid\Uuids;
 use Throwable;
 
 /**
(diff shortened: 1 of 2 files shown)


bastianallgeier:
This is great. I just wonder if we should put this in 5.3 instead. It does not feel like a part of a patch release.

bastianallgeier (dismissed):

distantnative:
Fine for me

bastianallgeier (approved):

PR #7810, opened 2025-12-21
lab: `css('@snippets')`

distantnative:
## Description

I need your opinions on this one:
It has been requested a few times to have something like `css('@auto')` but for snippets, not just templates (https://feedback.getkirby.com/535).

This PR kind of achieves this by introducing a new `css('@snippets')`. And the first draft seems to work relatively well without too much code overhead. 

However, it only works when placed at the bottom of the template. Or to phrase it differently: It can only automatically include assets for a snippet that has already been called before the `css('@snippets')` call. It isn't aware of snippets that follow after it. Which for JS is less relevant as that most often gets placed right before the `</body>` tag, but for CSS it would mean that a second call (aside from the usual one inside `<head>`) would have to be placed there as well if one wants to use this feature.

Given the, I am not so sure about the usefulness of this feature. It's a little quirky. Then again, especially for block snippets that could be quite relevant.

### DIFF
--- a/src/Cms/Html.php
+++ b/src/Cms/Html.php
@@ -31,29 +31,8 @@ class Html extends \Kirby\Toolkit\Html
 		string|array|Plugin|Assets $url,
 		string|array|null $options = null
 	): string|null {
-		if ($url instanceof Plugin) {
-			$url = $url->assets();
-		}
-
-		if ($url instanceof Assets) {
-			$url = $url->css()->values(fn ($asset) => $asset->url());
-		}
-
-		if (is_array($url) === true) {
-			$links = A::map($url, fn ($url) => static::css($url, $options));
-			return implode(PHP_EOL, $links);
-		}
-
 		if (is_string($options) === true) {
 			$options = ['media' => $options];
-		}
-
-		$kirby = App::instance();
-
-		if ($url === '@auto') {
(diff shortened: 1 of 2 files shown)


bastianallgeier:
@afbora had a suggestion for adding stacks once, which is inspired by Laravel Blade Stacks: https://laravel.com/docs/12.x/blade#stacks It feels to me that this would be a more generally useful approach to handle additional CSS or JS for snippets or other parts in the template.

distantnative:
@bastianallgeier But stacks Is the developer stacking scripts or similar things in snippets/templates and then rendering them somewhere else. This still wouldn't solve plugins being able to automatically add their scripts/styles to templates, or in what way do you see that?

bastianallgeier:
No, that part wouldn't be solved. But plugin snippets could of course still use stacks to push scripts and styles and the template developer would then have to use such a stack instead of the css helper. I don't think css('@snippets') vs. stack('styles') is so much different.

afbora:
If you missed; I've a plugin for it: https://plugins.getkirby.com/owebstudio/template-stacks Not sure maybe we can move into core with enhanced version.

distantnative:
@afbora How does your plugin compare to this implementation? https://github.com/getkirby/kirby/commit/8f32d543d33d40563e0cf2cc6bbf4bdfe36c4aeb

PR #7812, opened 2025-12-21
feat: Warnings for all required PHP extensions

distantnative:
## Description

This one of these things that has been in my lab folder for years. I refreshed it for v6. I think it would be great to make a decision if we want this, so we can either merge it or I can throw it our of the window (my lab folder) :D

I can see it going either way, but want to put it up for a decision with this PR.

### ✨ Enhancements

- On installation, all required PHP extensions are checked and installation is blocked if one is missing
- System view shows an alert if a required PHP extension is missing

### For review team

### DIFF
--- a/i18n/translations/en.json
+++ b/i18n/translations/en.json
@@ -427,11 +427,10 @@
 	"installation.disabled": "The panel installer is disabled on public servers by default. Please run the installer on a local machine or enable it with the <code>panel.install</code> option.",
 	"installation.issues.accounts": "The <code>/site/accounts</code> folder does not exist or is not writable",
 	"installation.issues.content": "The <code>/content</code> folder does not exist or is not writable",
-	"installation.issues.curl": "The <code>CURL</code> extension is required",
+	"installation.issues.extension": "The <code>{extension}</code> PHP extension is required",
 	"installation.issues.headline": "The panel cannot be installed",
-	"installation.issues.mbstring": "The <code>MB String</code> extension is required",
 	"installation.issues.media": "The <code>/media</code> folder does not exist or is not writable",
-	"installation.issues.php": "Make sure to use <code>PHP 8+</code>",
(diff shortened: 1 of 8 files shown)


bastianallgeier (commented):
I love this. We should maybe do a quick benchmark comparison. The isOk check is running before any Panel route and I'm not sure how long it takes to check for all extensions.

bastianallgeier (approved):

i18n/translations/en.json
```diff
@@ -427,11 +427,10 @@
 	"installation.disabled": "The panel installer is disabled on public servers by default. Please run the installer on a local machine or enable it with the <code>panel.install</code> option.",
 	"installation.issues.accounts": "The <code>/site/accounts</code> folder does not exist or is not writable",
 	"installation.issues.content": "The <code>/content</code> folder does not exist or is not writable",
-	"installation.issues.curl": "The <code>CURL</code> extension is required",
+	"installation.issues.extension": "The <code>{extension}</code> PHP extension is required",
 	"installation.issues.headline": "The panel cannot be installed",
-	"installation.issues.mbstring": "The <code>MB String</code> extension is required",
 	"installation.issues.media": "The <code>/media</code> folder does not exist or is not writable",
-	"installation.issues.php": "Make sure to use <code>PHP 8+</code>",
+	"installation.issues.php": "Make sure to use <code>PHP 8.3+</code>",
```
lukasbestle, line 433: I have the feeling we will forget to update this. Also I think it would be good to include the max version (`PHP 8.3-8.5`) as well.

Thinking whether we should make the version a placeholder and find a global place where we can define the supported versions (min and max) once.

On the other hand, the only case this message will ever be printed is if the `KIRBY_PHP_VERSION_CHECK` constant was set to `false`. By default, the Panel won't even be rendered on an incompatible version. So maybe we could just throw this translation key and its matching logic out completely?

distantnative: I will create a PR to remove `System::php()` etc.

PR #7819, opened 2025-12-25
[License 1] feat: `LicenseStatus` + `LicenseType` cases

distantnative:
- `LicenseType::Free`
- `LicenseStatus::Acknowledged`

### DIFF
--- a/src/Cms/LicenseStatus.php
+++ b/src/Cms/LicenseStatus.php
@@ -21,6 +21,13 @@ enum LicenseStatus: string
 	case Active = 'active';
 
 	/**
+	 * Compliance with the conditional license
+	 * requirements has been confirmed
+	 * @since 5.3.0
+	 */
+	case Acknowledged = 'acknowledged';
+
+	/**
 	 * Only used for the demo instance
 	 */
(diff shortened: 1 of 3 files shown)


src/Cms/LicenseStatus.php
```diff
@@ -20,6 +20,12 @@ enum LicenseStatus: string
 	 */
 	case Active = 'active';
 
+	/**
+	 * Compliance with the conditional license
+	 * requirements has been confirmed
+	 */
+	case Acknowledged = 'acknowledged';
```
distantnative, line 28: `Discuss:` Not really happy about this term. If someone has a better idea what to call this in the context of the other license status cases.

bastianallgeier: I'm totally fine with it. It could just be "confirmed" instead, but I don't think it's better.

bastianallgeier (approved):

src/Cms/LicenseStatus.php
distantnative: Then maybe week keep it ✅

PR #7820, opened 2025-12-25
[License 3] feat: License activate dialog for free licenses

distantnative:
## Description

First step for this feature, just some first UI changes:

Screenshot 2025-12-30 at 15 15 20
Screenshot 2025-12-30 at 15 15 26

Dialog still requires email and code for develop/private and registers it as before, no matter which toggle you chooses. This will change in a future PR. But this PR can be used to figure out the UI already, discuss the labels etc.


bastianallgeier:
I would maybe remove the "personal use" part for now and only focus on development licenses first. This would give us a great starting point to test this all without making it seem too open.

distantnative:
@bastianallgeier Makes sense to me to keep it more confined for the start. I changed the strings.

bastianallgeier:
I like the dialog changes in general, but we no longer show a warning in presumably local environments that a regular license should probably not be registered here.

distantnative:
@bastianallgeier Added back an info field

bastianallgeier (approved):

bastianallgeier (approved):

PR #7821, opened 2025-12-25
[License 2] feat: `License::register()` for free licenses

distantnative:
## Description

- Create unsigned local license file if free license and system is local (do not require a signature in this case)
- Get payload from Hub for free license otherwise (depends https://github.com/getkirby/hub.getkirby.com/pull/113 on the Hub end - but that depends on the other PR we merged, we should consider if we want to patch in the `LicenseStatus` and `LicenseType` changes to the Hub already ahead of 5.3 release to be able to merge the Hub PRs)
- Send the Kirby version to the Hub with the register request

### Merge first

- [x] https://github.com/getkirby/kirby/pull/7819

### DIFF
--- a/src/Cms/License.php
+++ b/src/Cms/License.php
@@ -164,7 +164,7 @@ class License
 			$this->domain !== null &&
 			$this->email !== null &&
 			$this->order !== null &&
-			($this->signature !== null || $this->isFreeAndLocal()) &&
+			$this->signature !== null &&
 			$this->hasValidEmailAddress() === true &&
 			$this->type() !== LicenseType::Invalid
 		) {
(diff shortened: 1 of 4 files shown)


src/Cms/License.php
```diff
@@ -53,6 +53,10 @@ public function __construct(
 			$this->email = $this->normalizeEmail($email);
 		}
 
+		if ($code === LicenseType::Free->prefix()) {
+			$this->email ??= 'mail@getkirby.com';
```
bastianallgeier: Nitpick: We could use our licensing@getkirby.com email here.

src/Cms/License.php
```diff
@@ -372,8 +404,19 @@ public function register(): static
...
 		}
 
+		if ($this->isFreeAndLocal() === true) {
+			$response = $this->selfsign([
+				'activation' => date('Y-m-d H:i:s'),
+				'code'       => $this->code,
+				'date'       => date('Y-m-d H:i:s'),
+				'domain'     => $this->domain,
+				'email'      => $this->email,
+				'order'      => '12345678',
```
bastianallgeier, line 414: Nitpick: I would use something like 'free' or 'develop' or maybe a random hash. It's super theoretical, but I don't think we can guarantee that this would never collide with a paddle order id.

distantnative: We can change it. But we should not worry about collisions. This never sees the light of the Hub - it stays only on the site itself. And the license codes will also be always `FREE`  - so all the same. But also doesn't matter as this here is only for local license file, never shared with anywhere else. If you would like the random hash, we should also do that for the code, no? (but then the Enum logic needs to change).

bastianallgeier: Let's leave it as it is.

bastianallgeier (approved):

PR #7823, opened 2025-12-25
[License 6] feat: License dialog improvements

distantnative:
### ✨ Enhancements

- License dialog: Show registered domain and link to license terms

### For review team

- [x] Add changes & docs to main license improvements PR

### DIFF
--- a/panel/src/components/Dialogs/LicenseDialog.vue
+++ b/panel/src/components/Dialogs/LicenseDialog.vue
@@ -6,45 +6,54 @@
 		@cancel="$emit('cancel')"
 		@submit="$emit('submit')"
 	>
-		<k-bar style="margin-bottom: var(--spacing-2)">
-			<h2 class="k-headline">
-				{{ $t("license") }}
-			</h2>
+		<k-stack>
+			<k-bar>
+				<h2 class="k-headline">
+					{{ $t("license") }}
+				</h2>
 
-			<k-button
-				:text="$t('remove')"
-				icon="trash"
-				size="xs"
-				variant="filled"
-				dialog="license/remove"
-			/>
-		</k-bar>
+				<k-button
+					:text="$t('remove')"
+					icon="trash"
+					size="xs"
+					variant="filled"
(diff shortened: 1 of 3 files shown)


panel/src/components/Dialogs/LicenseDialog.vue
```diff
@@ -6,45 +6,54 @@
...
-				</tbody>
-			</table>
-		</div>
+			<k-definitions>
+				<k-definition :term="$t('type')">
+					<p class="k-license-dialog-type">
+						{{ license.type }}
+						<k-button
+							icon="info"
+							link="https://getkirby.com/license"
```
distantnative, line 30: @lukasbestle I had the idea if we could add a feature to getkirby.com where we could resolve an activation date or similar to a license terms version, e.g. `https://getkirby.com/license?on=2025-10-15` redirecting to the license terms that were the current ones at that time point in time.

panel/src/components/Layout/Definition.vue
```diff
@@ -36,8 +36,7 @@ export default {
 .k-definition dd {
 	padding-block: var(--spacing-2);
 	padding-inline: var(--table-cell-padding);
-	height: var(--definition-height);
-	min-height: var(--table-row-height);
+	min-height: var(--definition-height);
```
distantnative, line 39: @bastianallgeier The old code prevented line breaks in the definition description.

panel/src/components/Dialogs/LicenseDialog.vue
bastianallgeier: I really like that. Could we already add that to the PR and then take care of it on the other side later?

bastianallgeier (approved):

panel/src/components/Dialogs/LicenseDialog.vue
distantnative: Thinking again about this, modifications to the license terms are binding also for previous purchases. So I think we should always link to the most recent terms.

PR #7824, opened 2025-12-25
[License 5] feat: License file expires + automatic reissue

distantnative:
### Merge first

- [x] https://github.com/getkirby/kirby/pull/7821

### 🧹 Housekeeping

- License files can have an `expires` date and get automatically reissued or removed once expired

### For review team

### DIFF
--- a/src/Cms/License.php
+++ b/src/Cms/License.php
@@ -44,6 +44,7 @@ class License
 		protected string|null $order = null,
 		protected string|null $date = null,
 		protected string|null $signature = null,
+		protected string|null $expires = null
 	) {
 		if ($code !== null) {
 			$this->code = trim($code);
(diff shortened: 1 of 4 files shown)


src/Cms/License.php
```diff
@@ -191,7 +207,8 @@ public function isFreeAndLocal(): bool
 	{
 		return
 			$this->isFree() === true &&
-			$this->kirby->system()->isLocal() === true;
+			$this->kirby->system()->isLocal() === true &&
+			$this->expires !== null;
```
distantnative: Lukas brought up that those local free licenses do not need to expire. Let's discuss that in the team.

bastianallgeier:
It looks like something's wrong with the code ignore comment in the System.php. We could also test the signatureData method, since it is public.

bastianallgeier (approved):

PR #7826, opened 2025-12-25
Preview view tweaks

distantnative:
## Description

- View size buttons also for compare mode
- Disable view size button if browser width not sufficient to be used
- Position view buttons dead center


bastianallgeier:
Removing the view size buttons on smaller widths also seems to break the flex layout …

Screenshot 2025-12-30 at 11 25 53

distantnative:
@bastianallgeier Could you please try again?

bastianallgeier:
Works great now!

PR #7827, opened 2025-12-26
fix(slug-field): Apply language rules on initial value

### LINKED ISSUE #7662: Filename field bug in upload dialog
## Description
When uploading a file with German umlauts, the filename displayed in the upload field may differ from the name under which the file is actually saved.

- Umlauts are replaced as follows: ä → a, ö → o, ü → u, or when the slug option is set to `de`, ä → ae, ö → oe, ü → ue.
- However, the `e` may appear as a `-` in the filename field of the upload dialog.
- If you do not change the filename in the upload dialog, the file will be saved correctly with the replaced characters. If you do modify the filename, the file will be saved based on the wrong name.

**Expected behavior**  
The filename should match the saved filename.

**Screenshots**
Upload dialog:
Image
Saved file:
Image

## To reproduce
1. Upload a file with German umlauts
2. Check the filename in the upload dialog
3. Check the actually saved filename

## Your setup
**Kirby Version**  
5.1.3

**Console output**  
None

 - Device: MacBook Pro
 - OS: macOS 26.1
 - Browser: Safari
 - Version: 26.1

afbora:
## Description

The slug field wasn't converting special characters (ä→ae, ü→ue) when initialized with a default value. It only worked when typing manually.

The issue was that sluggify() was called in data() before this.slugs was available. The value watcher now runs immediately on component mount to ensure slugs are generated with proper language rules.

### 🐛 Bug fixes

- Filename field bug in upload dialog #7662

### For review team

### DIFF
--- a/panel/src/components/Forms/Input/SlugInput.vue
+++ b/panel/src/components/Forms/Input/SlugInput.vue
@@ -76,13 +76,16 @@ export default {
 			deep: true,
 			immediate: true
 		},
-		value(newValue) {
-			newValue = this.sluggify(newValue);
+		value: {
+			handler(newValue) {
+				newValue = this.sluggify(newValue);
 
-			if (newValue !== this.slug) {
-				this.slug = newValue;
-				this.$emit("input", this.slug);
-			}
+				if (newValue !== this.slug) {
+					this.slug = newValue;
+					this.$emit("input", this.slug);
+				}
+			},
+			immediate: true
 		}
 	},
 	methods: {


distantnative:
Wondering if it also work not to call it inside `data()` at all but instead add `immediate: true` to the value watcher? If that would work, it would be even cleaner.

afbora:
@distantnative That's the correct and nice solution. Thanks. I've pushed the related changes.

distantnative:
@afbora Could you please base this on `develop-patch` if it shall be merged for 5.2.2? Thanks!

afbora:
@distantnative Done.

bastianallgeier (approved):

PR #7830, opened 2025-12-29
fix: Accept empty string as valid group key

### LINKED ISSUE #7733: `$collection->group(callable)` should accept empty string result as key
## Description

When using `$collection->group('some_field')`, an empty field value is allowed as key in the resulting data structure. Similarly, when returning a Field from `$collection->group(fn($item) => $item->some_field())`, an empty field value is allowed.

However, returning the empty string directly throws an error: 

```php
$collection->group(
    fn($item) => $item->some_field()->value()
)
```

Offending line: https://github.com/getkirby/kirby/blob/2278dae6b41879bd1a8ac70ae62a6ea781fc629a/src/Toolkit/Collection.php#L523C5-L528C6

**Expected behavior**  

An empty string should be an acceptable return value from the `$collection->group()` callback.

## Your setup
**Kirby Version**  5.1.4.

afbora:
## Description

The group() method was throwing an exception when a callable returned an empty string, even though empty strings are valid array keys in PHP. This was inconsistent with grouping by field name, which already accepted empty strings.

Changed the validation to only reject null values, allowing empty strings, zero, and other falsy but valid values to be used as group keys.

Fixes #7733

### 🐛 Bug fixes

- `$collection->group(callable)` should accept empty string result as key #7733

### For review team

### DIFF
--- a/src/Toolkit/Collection.php
+++ b/src/Toolkit/Collection.php
@@ -521,7 +521,7 @@ class Collection extends Iterator implements Stringable
 				$value = $field($item);
 
 				// make sure that there's always a proper value to group by
-				if (!$value) {
+				if ($value === null || $value === false) {
 					throw new Exception(
 						message: 'Invalid grouping value for key: ' . $key
 					);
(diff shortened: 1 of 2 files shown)


distantnative (approved):

distantnative:
@afbora Only realized after merging that you assigned this for 5.2.2 (which is fine) but pointed it at `develop-minor`. Will cherry-pick the commit for `develop-patch`.

PR #7837, opened 2025-12-30
[User performance 1] perf: Only load user credentials file once

lukasbestle:
## Performance

> All tested on my Mac mini M2 with PHP 8.4
> Timing values are TTFB measured using `h2load`:
> 296.49ms / 3.42s means "296 ms mean TTFB for 10 sequential requests, 3.42 s mean TTFB under load with 20 requests in parallel, 100 requests total" (numbers under load are better to compare because they fluctuate less during test executions)

### Frontend render using `$kirby->user()`

```
Baseline with 1 user:
23 ms

With 20000 users:
296.49ms / 3.42s @ Kirby 5.2.1
293.58ms / 3.10s @ PR 1 (Only load user credentials once)
```

### Starterkit site view

```
Baseline with 1 user:
75 ms

With 20000 users:
255.04ms / 2.79s @ Kirby 5.2.1
237.86ms / 2.73s @ PR 1 (Only load user credentials once)
```

### Users view

```
Baseline with 1 user:
65 ms

With 20000 users:
435.88ms / 4.88s @ Kirby 5.2.1
312.80ms / 3.30s @ PR 1 (Only load user credentials once)
```

### Account view (admin)

```
Baseline with 1 user:
70 ms

With 20000 users:
459.64ms / 5.38s @ Kirby 5.2.1
363.35ms / 4.34s @ PR 1 (Only load user credentials once)
```

### Account view (non-admin)

```
Baseline with 2 users:
75 ms

With 20000 users:
392.59ms / 4.79s @ Kirby 5.2.1
324.36ms / 3.67s @ PR 1 (Only load user credentials once)
```

### ⚡Performance

- User credentials (main data from `index.php`) are only loaded once when the user object is first created, improving system performance with many user accounts.

## Docs

None

### For review team

- [x] Add changes & docs to release notes draft in Notion

PR #7838, opened 2025-12-30
[User performance 2] refact: Refactor `Iterator` class

lukasbestle:
## Merge first

- [x] #7837 

## Description

This PR combines a few minor refactorings:

- Widening the return value for `getIterator()` is needed as `LazyCollection` (PR 3) needs to return a more general `Iterator`, not an `ArrayIterator`; this change has no effect on code behavior
- Switching from `isset()` to `array_key_exists()` is needed as `LazyCollection` will have elements that are `null` because they are not yet loaded. But they are still valid elements of the collection. IMO it was a bug to use `isset()` here.
- The old iterator methods are a leftover from #2216. Before that PR, we implemented the `Iterator` interface, which requires all these separate methods. We switched to the `IteratorAggregate` interface, which uses `getIterator()` instead. The old methods are not used anywhere and are not useful either because they all depend on the internal array cursor (which is only useful during iteration). So deprecating these cleans up the code.
- Also added another test for iteration while I was at it.

### ✨ Enhancements

- Child classes of `Iterator` (mostly collections) can now implement a custom iterator and are not pinned to `ArrayIterator`

### 🐛 Bug fixes

- `Iterator::has()` and the `has()` method of collections correctly treats an existing but `null` value as a valid collection element

### ☠️ Deprecated

- The following cursor-dependent methods of `Iterator` and collection classes have been deprecated and will be removed in Kirby 6: `key`, `current`, `next`, `prev`, `rewind` and `valid`

### 🧹 Housekeeping

- Improve `Iterator` PHPUnit tests

### For review team

### DIFF
--- a/src/Toolkit/Iterator.php
+++ b/src/Toolkit/Iterator.php
@@ -45,6 +45,8 @@ class Iterator implements Countable, IteratorAggregate
 
 	/**
 	 * Returns the current key
+	 * @deprecated
+	 * @todo Remove in v6
 	 */
 	public function key(): int|string|null
 	{
@@ -61,6 +63,8 @@ class Iterator implements Countable, IteratorAggregate
 
 	/**
 	 * Returns the current element
+	 * @deprecated
+	 * @todo Remove in v6
 	 * @return TValue
 	 */
 	public function current(): mixed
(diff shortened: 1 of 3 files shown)


distantnative (commented):
Should we already remove the tests for the deprecated methods or rather wait until those methods get removed? That's the only point I was wondering about.

lukasbestle:
TBH I think these methods are so irrelevant and unused that the risk is super low in the time between 5.3 and 6.0.

distantnative (approved):

PR #7839, opened 2025-12-30
[User performance 3] feat: New `LazyCollection` class

lukasbestle:
## Merge first

- [x] #7837 
- [x] #7838 

## Description

The `LazyCollection` works by introducing a `null` value for collection elements. If a data key is set to `null`, this means "element exists but was not loaded yet". The loading then happens dynamically via a hydration method. This `hydrateElement` method is defined in the child collection and handles the actual loading and instance creation for the element object. This is the reason why `LazyCollection` is an `abstract` class.

What's important is that all collection methods need to hook into this dynamic behavior. I have overridden all core collection methods from parent classes that perform low-level actions directly on the data array. Then there are some methods that use `foreach`. For these, it's enough to replace `foreach ($this->data)` with `foreach ($this)` (making the loops use the custom iterator and therefore dynamically hydrate just as many elements as are needed until the loop is complete). Custom collection methods are currently made backwards-compatible by hydrating everything first. In v6, collection methods should call the hydration method themselves if they need to perform low-level actions.

Inherited `Iterator` methods (see PR 2) are implemented for now for backwards-compatibility but already marked as `@deprecated`. This allows us to clean them up in v6 as well.

This PR is the first step that only lazily hydrates the objects. Meaning that in this step, the collection needs to be initialized with all possible collection elements (each with a `null` value). This already brings performance improvements (see following PR 4). PR 5 will then also make the initialization lazy, which is useful if one only wants to retrieve a few specific collection elements by their ID.

### 🎉 Features

- New `LazyCollection` class that initializes and loads its data and elements only when they are first needed

### ☠️ Deprecated

- Starting in Kirby 6, collection methods can no longer rely on the low-level `$collection->data` property to be populated. Collection methods should use `foreach ($collection)` instead of `foreach ($collection->data)`. If direct access to the data array is required, call `$collection->initialize()` (to initialize all keys) or `$collection->hydrate()` (to initialize all keys and values). During Kirby 5, collection methods still receive a fully hydrated collection for backwards-compatibility.

### For review team

### DIFF
--- a/src/Cms/LazyCollection.php
+++ b/src/Cms/LazyCollection.php
@@ -1,0 +1,324 @@
+<?php
+
+namespace Kirby\Cms;
+
+use Closure;
+use Iterator;
+
+/**
+ * The LazyCollection class is a variant of the CMS
+ * Collection that is only initialized with keys for
+ * each collection element. Collection values (= objects)
+ * are loaded and initialized lazily when they are
+ * first used.
+ *
+ * @package   Kirby Cms
+ * @author    Lukas Bestle <lukas@getkirby.com>
+ * @link      https://getkirby.com
+ * @copyright Bastian Allgeier
+ * @license   https://getkirby.com/license
+ *
+ * @template TValue
+ * @extends \Kirby\Cms\Collection<TValue>
+ */
+abstract class LazyCollection extends Collection
+{
+	/**
+	 * Flag that tells whether hydration has been
+	 * completed for all collection elements;
+	 * this is used to increase performance
+	 */
+	protected bool $hydrated = false;
+
+	/**
+	 * Temporary auto-hydration whenever a collection
+	 * method is called; some methods may not need raw
+	 * access to all collection data, so performance
+	 * will be improved if methods call hydration
+	 * themselves only if they need it
+	 * @deprecated
+	 * @todo Remove this in v6
+	 */
+	public function __call(string $key, $arguments)
+	{
+		$this->hydrateAll();
+		return parent::__call($key, $arguments);
+	}
+
+	/**
+	 * Low-level getter for elements
+	 *
+	 * @return TValue|null
+	 */
+	public function __get(string $key)
+	{
+		$element = parent::__get($key);
+
+		// `$element === null` could mean "element does not exist"
+		// or "element found but not hydrated"
+		if ($element === null && array_key_exists($key, $this->data)) {
+			return $this->hydrateElement($key);
+		}
+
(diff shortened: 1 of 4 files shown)


src/Cms/LazyCollection.php
```diff
@@ -0,0 +1,324 @@
...
+	 *
+	 * @return TValue|null
+	 */
+	public function __get(string $key)
+	{
+		$element = parent::__get($key);
+
+		// `$element === null` could mean "element does not exist"
+		// or "element found but not hydrated"
+		if ($element === null && array_key_exists($key, $this->data)) {
```
distantnative, line 59: `Suggestion:` Working with `null` as placeholder probably works but maybe making it more explicit could have an advantage/less risk? E.g. would it be an option to use e.g. a new enum case `CollectionElement::Latent`?

src/Cms/LazyCollection.php
```diff
@@ -0,0 +1,324 @@
...
+			yield $key => $value;
+		}
+	}
+
+	/**
+	 * Ensures that all collection elements are loaded,
+	 * essentially converting the lazy collection into a
+	 * normal collection
+	 */
+	public function hydrateAll(): void
```
distantnative, line 137: `Suggestion:` hydrating the collection would imply all elements

```suggestion
	public function hydrate(): void
```

lukasbestle: I like that. To avoid merge conflicts (especially because PR 5 also introduces `initializeAll()` that shares the same naming change), I'd implement this in a PR 6. I have added it to my local todo list.

src/Cms/LazyCollection.php
lukasbestle: The parent class `Cms\Collection` is only used in the CMS context and type-hinted for objects as values. It does support non-objects as values, but AFAIK we never use it like this anywhere. `null` is even more unlikely as collection element.

I chose `null` because we are potentially dealing with a lot of these latent elements in a large collection and `null` is the most simple value there can be. I haven't done research on the way PHP represents enums internally, but what we would definitely not want is that each of these enums takes up additional memory. So not sure about the performance implications. But I'm open for enums if they are performance-neutral. Would certainly make it extra safe.

src/Cms/LazyCollection.php
lukasbestle: #7843

bastianallgeier (approved):

distantnative (approved):

PR #7840, opened 2025-12-30
refact!: Login view as stack

distantnative:
## Description

Without touching the existing methods/challenges backend logic, let's already move the login view in the direction we want. We also keep the `k-login-form` + `k-code-login-form` setup for now. But move the toggling up to the view component:

Screenshot 2025-12-30 at 16 38 50
Screenshot 2025-12-30 at 16 38 59
Screenshot 2025-12-30 at 16 39 03

### ♻️ Refactored

- New `k-login-back-button`

### 🚨 Breaking changes

- `k-login-view` and components have been refactored with breaking changes

### For review team


bastianallgeier (approved):
Nice!

PR #7842, opened 2025-12-30
[User performance 5] perf: Lazy initialization

lukasbestle:
## Merge first

- [x] #7837 
- [x] #7838 
- [x] #7839 
- [x] #7841

## Description

This PR extends the `LazyCollection` concept by integrating an optional lazy initialization mode. Lazy initialization means that the keys of the collection elements also only need to be loaded when they are needed. A `Dir::read()` is already pretty expensive when there are thousands of children in a directory. If only specific children are needed (e.g. just the currently logged in user), this operation can be skipped. This is the reason why this PR mainly improves performance for the frontend render case (and that one dramatically).

Child collections can decide if they just want to hydrate lazily or also want to initialize lazily. Default is just the lazy hydration mode.

Similar to PR 3, there are collection methods that perform low-level operations on the data array and that depend on all keys being there. These methods then dynamically trigger the initialization.

## Performance

> All tested on my Mac mini M2 with PHP 8.4
> Timing values are TTFB measured using `h2load`:
> 296.49ms / 3.42s means "296 ms mean TTFB for 10 sequential requests, 3.42 s mean TTFB under load with 20 requests in parallel, 100 requests total" (numbers under load are better to compare because they fluctuate less during test executions)

### Frontend render using `$kirby->user()`

```
Baseline with 1 user:
23 ms

With 20000 users:
296.49ms / 3.42s @ Kirby 5.2.1
293.58ms / 3.10s @ PR 1 (Only load user credentials once)
127.20ms / 1.38s @ PR 4 (Lazy hydration)
 20.02ms / 0.24s @ PR 5 (Lazy initialization)
```

### Starterkit site view

```
Baseline with 1 user:
75 ms

With 20000 users:
255.04ms / 2.79s @ Kirby 5.2.1
237.86ms / 2.73s @ PR 1 (Only load user credentials once)
114.46ms / 1.36s @ PR 4 (Lazy hydration)
112.67ms / 1.16s @ PR 5 (Lazy initialization)
```

### Users view

```
Baseline with 1 user:
65 ms

With 20000 users:
435.88ms / 4.88s @ Kirby 5.2.1
312.80ms / 3.30s @ PR 1 (Only load user credentials once)
308.57ms / 3.60s @ PR 4 (Lazy hydration)
305.56ms / 3.54s @ PR 5 (Lazy initialization)
```

### Account view (admin)

```
Baseline with 1 user:
70 ms

With 20000 users:
459.64ms / 5.38s @ Kirby 5.2.1
363.35ms / 4.34s @ PR 1 (Only load user credentials once)
371.18ms / 4.28s @ PR 4 (Lazy hydration)
368.34ms / 3.80s @ PR 5 (Lazy initialization)
```

### Account view (non-admin)

```
Baseline with 2 users:
75 ms

With 20000 users:
392.59ms / 4.79s @ Kirby 5.2.1
324.36ms / 3.67s @ PR 1 (Only load user credentials once)
327.61ms / 3.53s @ PR 4 (Lazy hydration)
290.66ms / 3.47s @ PR 5 (Lazy initialization)
```

## Changelog 

*No additional release notes (covered by PRs 3 and 4)*

### DIFF
--- a/src/Cms/LazyCollection.php
+++ b/src/Cms/LazyCollection.php
@@ -261,7 +261,7 @@ abstract class LazyCollection extends Collection
 			return;
 		}
 
-		throw new LogicException('Lazy initialization is not implemented for ' . static::class); // @codeCoverageIgnore
+		throw new LogicException(static::class . ' class does not implement `initialize()` method that is required for lazy initialization'); // @codeCoverageIgnore
 	}
 
 	/**
--- a/src/Cms/LazyCollection.php
+++ b/src/Cms/LazyCollection.php
@@ -4,13 +4,24 @@ namespace Kirby\Cms;
 
 use Closure;
 use Iterator;
+use Kirby\Exception\LogicException;
 
 /**
  * The LazyCollection class is a variant of the CMS
  * Collection that is only initialized with keys for
- * each collection element. Collection values (= objects)
+ * each collection element or without any data.
+ * Collection elements and their values (= objects)
  * are loaded and initialized lazily when they are
  * first used.
+ *
+ * You can use LazyCollection in two ways:
+ * 1. Initialize with keys only (values are `null`),
+ *    define `hydrateElement` method that initializes
+ *    an element dynamically.
+ * 2. Option 1, but also don't initialize any keys,
+ *    set `$initialized` prop to `false` and define
+ *    `initializeAll` method that defines which keys
+ *    are available.
  *
  * @package   Kirby Cms
  * @author    Lukas Bestle <lukas@getkirby.com>
@@ -31,11 +42,18 @@ abstract class LazyCollection extends Collection
 	protected bool $hydrated = false;
 
 	/**
+	 * Flag that tells whether all possible collection
+	 * items have been loaded (only relevant in lazy
+	 * initialization mode)
+	 */
+	protected bool $initialized = true;
+
+	/**
 	 * Temporary auto-hydration whenever a collection
 	 * method is called; some methods may not need raw
 	 * access to all collection data, so performance
-	 * will be improved if methods call hydration
-	 * themselves only if they need it
+	 * will be improved if methods call initialization
+	 * or hydration themselves only if they need it
 	 * @deprecated
 	 * @todo Remove this in v6
 	 */
@@ -56,11 +74,52 @@ abstract class LazyCollection extends Collection
 
 		// `$element === null` could mean "element does not exist"
 		// or "element found but not hydrated"
-		if ($element === null && array_key_exists($key, $this->data)) {
+		if (
+			$element === null &&
+			(array_key_exists($key, $this->data) || $this->initialized === false)
+		) {
 			return $this->hydrateElement($key);
 		}
 
 		return $element;
+	}
+
+	/**
+	 * Low-level element remover
+	 */
+	public function __unset(string $key)
+	{
+		// first initialize, otherwise a later initialization
+		// might bring back the element that was unset
+		$this->initializeAll();
+
+		return parent::__unset($key);
+	}
+
+	/**
+	 * Creates chunks of the same size.
+	 * The last chunk may be smaller
+	 *
+	 * @param int $size Number of elements per chunk
+	 * @return static A new collection with an element for each chunk and
+	 *                a sub collection in each chunk
+	 */
+	public function chunk(int $size): static
+	{
+		// chunking at least requires the collection structure
+		$this->initializeAll();
+
+		return parent::chunk($size);
+	}
+
+	/**
+	 * Counts all elements
+	 */
+	public function count(): int
+	{
+		$this->initializeAll();
+
+		return parent::count();
 	}
 
 	/**
@@ -84,6 +143,17 @@ abstract class LazyCollection extends Collection
 	}
 
 	/**
+	 * Returns the elements in reverse order
+	 */
+	public function flip(): static
+	{
+		// flipping at least requires the collection structure
+		$this->initializeAll();
+
+		return parent::flip();
+	}
+
+	/**
 	 * Filters elements by one of the
 	 * predefined filter methods, by a
 	 * custom filter function or an array of filters
@@ -103,6 +173,9 @@ abstract class LazyCollection extends Collection
 	 */
 	public function first()
 	{
+		// returning a specific offset requires the collection structure
+		$this->initializeAll();
+
 		$first = parent::first();
 
 		// `$first === null` could mean "empty collection"
@@ -120,6 +193,9 @@ abstract class LazyCollection extends Collection
 	 */
 	public function getIterator(): Iterator
 	{
+		// ensure we are looping over all possible elements
+		$this->initializeAll();
+
 		foreach ($this->data as $key => $value) {
 			if ($value === null) {
 				$value = $this->hydrateElement($key);
@@ -130,12 +206,26 @@ abstract class LazyCollection extends Collection
 	}
 
 	/**
+	 * Checks by key if an element is included
+	 * @param TKey $key
+	 */
+	public function has(mixed $key): bool
+	{
+		$this->initializeAll();
+
+		return parent::has($key);
+	}
+
+	/**
 	 * Ensures that all collection elements are loaded,
 	 * essentially converting the lazy collection into a
 	 * normal collection
 	 */
 	public function hydrateAll(): void
 	{
+		// first ensure all keys are initialized
+		$this->initializeAll();
+
 		// skip another hydration loop if no longer needed
 		if ($this->hydrated === true) {
 			return;
@@ -152,10 +242,38 @@ abstract class LazyCollection extends Collection
 
 	/**
 	 * Loads a collection element, sets it in `$this->data[$key]`
-	 * and returns the hydrated object value; to be implemented
-	 * in each specific collection
+	 * and returns the hydrated object value (or `null` if the
+	 * element does not exist in the collection); to be
+	 * implemented in each specific collection
 	 */
-	abstract protected function hydrateElement(string $key): object;
+	abstract protected function hydrateElement(string $key): object|null;
+
+	/**
+	 * Ensures that the keys for all valid collection elements
+	 * are loaded in the `$data` array and sets `$initialized`
+	 * to `true` afterwards; to be implemented in each collection
+	 * that wants to use lazy initialization; be sure to keep
+	 * existing `$data` values and not overwrite the entire array
+	 */
+	public function initializeAll(): void
+	{
+		if ($this->initialized === true) {
+			return;
+		}
+
+		throw new LogicException('Lazy initialization is not implemented for ' . static::class); // @codeCoverageIgnore
+	}
+
+	/**
+	 * Returns an array of all keys
+	 */
+	public function keys(): array
+	{
+		// ensure we are returning all possible keys
+		$this->initializeAll();
+
+		return parent::keys();
+	}
 
 	/**
 	 * Tries to find the key for the given element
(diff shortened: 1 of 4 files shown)


src/Cms/LazyCollection.php
```diff
@@ -30,12 +41,19 @@ abstract class LazyCollection extends Collection
 	 */
 	protected bool $hydrated = false;
 
+	/**
+	 * Flag that tells whether all possible collection
+	 * items have been loaded (only relevant in lazy
+	 * initialization mode)
+	 */
+	protected bool $initialized = true;
```
distantnative, line 49: `Question:` Would this not depend on whether any objects were passed to the constructor or not?

src/Cms/LazyCollection.php
```diff
@@ -56,13 +74,54 @@ public function __get(string $key)
 
 		// `$element === null` could mean "element does not exist"
 		// or "element found but not hydrated"
-		if ($element === null && array_key_exists($key, $this->data)) {
+		if (
+			$element === null &&
+			(array_key_exists($key, $this->data) || $this->initialized === false)
```
distantnative, line 79: `Question:` What is the situation where the collection was not yet initialized but we want to hydrate an element already?

src/Cms/LazyCollection.php
```diff
@@ -152,10 +242,38 @@ public function hydrateAll(): void
...
+	 * existing `$data` values and not overwrite the entire array
 	 */
-	abstract protected function hydrateElement(string $key): object;
+	public function initializeAll(): void
+	{
+		if ($this->initialized === true) {
+			return;
+		}
+
+		throw new LogicException('Lazy initialization is not implemented for ' . static::class); // @codeCoverageIgnore
```
distantnative: `Suggestion:` Make clear in the error message that the error is thrown because this specific method is missing.

src/Cms/Users.php
```diff
@@ -147,14 +147,22 @@ public function findByKey(string $key): User|null
...
 	{
 		if ($this->root === null) {
 			throw new LogicException('Cannot hydrate user "' . $key . '" with missing root'); // @codeCoverageIgnore
 		}
 
+		// check if the user directory exists if not all keys have been
+		// populated in the collection, otherwise we can assume that
+		// this method will only be called on "unhydrated" user IDs
+		$root = $this->root . '/' . $key;
+		if ($this->initialized === false && is_dir($root) === false) {
```
distantnative, line 160: I keep rereading this comment, but cannot follow. If the collection is not initialized yet, shouldn't we throw an error in general if someone wants to hydrate it before initialization?

src/Cms/Users.php
```diff
@@ -170,20 +178,48 @@ protected function hydrateElement(string $key): User
...
 	/**
-	 * Loads a user from disk by passing the absolute path (root)
+	 * Ensures that the IDs for all valid users are loaded in the
+	 * `$data` array and sets `$initialized` to `true` afterwards
 	 */
-	public static function load(string $root, array $inject = []): static
+	public function initializeAll(): void
 	{
-		$users = new static(root: $root, inject: $inject);
+		// skip another initialization if no longer needed
```
distantnative: "if no longer needed"? What does it refer to?

src/Cms/Users.php
```diff
@@ -170,20 +178,48 @@ protected function hydrateElement(string $key): User
...
 				continue;
 			}
 
-			$users->set($userDirectory, null);
+			$this->data[$userDirectory] = null;
 		}
 
+		foreach ($existing as $id => $user) {
+			$this->data[$id] = $user;
+		}
```
distantnative:
```suggestion
		$this->data = [...$this->data, ...$existing];
```

src/Cms/Users.php
```diff
@@ -170,20 +178,48 @@ protected function hydrateElement(string $key): User
...
+
+		$this->initialized = true;
+	}
+
+	/**
+	 * Loads a user from disk by passing the absolute path (root)
```
distantnative:
```suggestion
	 * Loads users from disk by passing the absolute directory path (root)
```

I know this description was there before already, but I think it doesn't make sense if phrased about a singular user, no?

src/Cms/LazyCollection.php
lukasbestle: `get()` is actually the only method that benefits from lazy initialization.

It allows to retrieve single objects without first reading the entire directory or database and fetching all possible IDs of collection entries.

Basically it works like this: If not initialized, we try to hydrate the queried element. `hydrateElement()` then checks if this element could be valid. If yes, hydrate and return it, otherwise return `null` from the getter.

src/Cms/Users.php
lukasbestle: See above: This is needed for the getter.

src/Cms/Users.php
lukasbestle: If `initialize()` is called on an already initialized collection, this returns early to improve performance.

src/Cms/LazyCollection.php
lukasbestle: There are some cases where collections are constructed empty and then immediately the `$data` prop gets set with the initial data. So the constructor cannot yet know if the collection will start out with data.

The `$initialized` prop also controls whether the class implementation uses lazy initialization at all. It should only be set to `false` if the collection class implements `initialize()`.

src/Cms/LazyCollection.php
distantnative: I still cannot really follow - I guess these will need some longer comments block for us (others than you) to understand this down the road still :D But here where I stumble: If the key doesn't exist, why should we still hydrate when it hasn't been initialized? Not initialized should mean there are no elements, no? Shouldn't it first need to be initialized (= adding empty keys to `data`) to then be able to get hydrated (= replacing `null` as value with the actual object)?

src/Cms/Users.php
distantnative: `Suggestion:` "skip another initialization if it already has been initialized"

src/Cms/LazyCollection.php
distantnative: Shouldn't then the default/start value still be `false`?

src/Cms/LazyCollection.php
distantnative: Phrasing it differently: likely I just haven't understood the concept pf initializing yet 😅😆

src/Cms/LazyCollection.php
lukasbestle: When implementing the `Users` improvements it was pretty great to be able to do it in small steps. `LazyCollection` can be used for lazy hydration and lazy initialization. Lazy initialization basically becomes progressive enhancement. If we would set the default for `$initialized` to `false`, this would mean that each child class needs to implement the `initialize()` method, breaking the "small progressive enhancement" approach.

src/Cms/LazyCollection.php
lukasbestle: With large collections, initialization is already an expensive process (just the `Dir::read()` on 20000 subfolders and the conversion into an array with `null` keys takes roughly 70 ms on my machine). So if we don't need to know all collection elements, we want to avoid initialization entirely. Almost all collection methods need to know all elements (simple example: `count()` needs to know how many elements there are, but doesn't care if they are hydrated). But `get()` is the only exception that doesn't care about any of that. It only needs to know if the requested collection element exists and it then needs the object for that element. So it can get away with just initializing and hydrating the single requested element, saving these 70 ms of directory reading if just the current user object is needed for example.

About implementation:

I first had separate `initializeAll()` and `initializeElement()` methods but I then simplified it to just `initializeAll()` (which becomes `initialize()` in PR 6). `initializeElement()` merged into `hydrateElement()`. The separate methods were easier to understand semantically, but have a major flexibility disadvantage for non-filesystem backends. In the filesystem, `initializeElement()` would do the `is_dir()` check and `hydrateElement()` would construct the `User` object from the `index.php` contents. But e.g. in a database, the existence and metadata could be queried in a single SQL query. So the separate methods would require to artificially split the single query into two, reducing performance.

So the result is:

- `initializeAll()` (= `initialize()`) that just focuses on the entire collection structure. This is needed for most collection methods that don't care about data of the elements but just their existence.
- `hydrateElement()` that takes two roles:
  - In an initialized collection, it replaces the `null` value with an object for the element.
  - In a non-initialized collection, it additionally validates the element key and skips hydration on keys that are not valid collection elements. This basically means that `hydrateElement()` also does partial initialization of a single collection element unless the entire collection is already initialized.

bastianallgeier:
Sorry for being late to this conversation. 

1. @lukasbestle do you already have a rough concept how this could work with database queries? I think it would be really great if we could already think ahead here a bit more and it would maybe also help to get a better understanding how this works in other scenarios. 
2. I also wonder if we could work on the naming of `initialize` vs. `hydrateElement` to make sure that we always come back to this anytime later and know exactly what we meant. `hydrateElement` works for me. Here are some ideas for `initialize`: `prefetchElements`, `discoverElements`, `buildIndex` - or more focused on keys: `prefetchKeys`, `discoverKeys`

lukasbestle:
> do you already have a rough concept how this could work with database queries? I think it would be really great if we could already think ahead here a bit more and it would maybe also help to get a better understanding how this works in other scenarios.

Thinking in terms of the database storage plugin:

- `DatabasePages::$initialized` would be set to `false` so that we can implement lazy initialization.
- `DatabasePages::fromRows()` would only set the arguments as props in the collection object but return an empty collection. The props can then later be used in `initialize()` and `hydrateElement()`.
- `DatabasePages::initialize()` would only do `$table->select('slug')`.
- `DatabasePages::hydrateElement()` would do something like `$table->where(['slug' => $key])`. With a database there should not even be a need to differentiate between the collection states "initialized" and "not yet initialized". The WHERE query for the slug will just return an empty result for elements that don't exist, which will then lead to a `null` return value. Any valid result can be set in the data array and returned.

But maybe it would help to build a prototype actually to test whether this works in practice. TBH I haven't really worked with databases for a while, maybe @bastianallgeier you could try it with the experience from the database storage plugin?

> I also wonder if we could work on the naming of initialize vs. hydrateElement to make sure that we always come back to this anytime later and know exactly what we meant. hydrateElement works for me. Here are some ideas for initialize: prefetchElements, discoverElements, buildIndex - or more focused on keys: prefetchKeys, discoverKeys

The advantage of `initialize()` and `hydrate()` is that they consistently apply to the entire collection (`$collection->initialize()` and `$collection->hydrate()`), see https://github.com/getkirby/kirby/pull/7839#discussion_r2653288717 and PR 6.

I like `discoverKeys()` though. It's easy to understand. Maybe we need another term for `hydrate()` though (referring to the entire collection) to make it consistent again.

bastianallgeier:
After reading your thoughts on database handling (which seems to be a great solution) I feel like maybe we are really nitpicking here with the naming and maybe it is all about having more real-world scenarios and proper docs that do the trick.

lukasbestle:
TBH I'm not even sure `LazyCollection` should be much of a user-facing feature. The database storage plugin implements custom model and collection classes because there are currently no proper hooks for plugin extensions. This will still be the case for the time being.

But once our content storage infrastructure is at the point where storage providers can be implemented fully (= once no core code contains hardcoded access to filesystem functions but only accesses the storage), one should not need to override the collection class just to be able to get models from another storage handler.

So this is again one of these intermediary internal features that will take a lot of time to "click in" fully.

So when thinking about docs, I agree that using this class only really makes sense with docs (extending from what I wrote in the PR descriptions, but with example code). But I am not sure if using this class makes sense at all for anyone but us.

bastianallgeier:
@distantnative I would suggest to move forward with it as it is, unless you still have thoughts.

bastianallgeier (approved):

